PR #5007, opened 2016-12-12
throws in fn decl: AST and parser

psahabu:
- adds `_throwsError`, `throwsError()`, and `throwsErrorInit()` to FnSymbol
- modifies `buildFunctionDecl` to take a new arg, `optThrowsError`
- updates the parser to accept `throws` in a function declaration
- updates AstDump to print that a function `throws`
- passes full linux64+flat testing, `--verify` testing on `release/examples`


compiler/include/symbol.h
```diff
@@ -478,6 +478,7 @@ class FnSymbol : public Symbol {
   virtual void    printDocs(std::ostream *file, unsigned int tabs);
 
   void            setThrowsError();
+  bool            getThrowsError();
```
mppf: I'd consider naming this method throwsError() (and the private field would need a different name in that case).

PR #4931, opened 2016-12-01
Figure out when we underestimate the heap page size

mppf:
Addresses a problem with the muxed tasking layer when using GASNet on a Cray XC system but without a hugepages module loaded.

In that configuration even Hello World would crash.  The problem was that in that configuration, GASNet arranges for the heap to be on huge pages.  Then muxed tries to create guard pages, but the mprotect call to do so creates a guard hugepage (much too large) instead of a guard page. The result is a core dump when the following code tries to use adjacent memory.

This patch adds logic to chplsys.c to try calling mprotect and accessing memory in sequence in a way that experimentally determines the page size of the Chapel heap. This new logic is run from chpl_mem_init.

Verified a simple Chapel program running getHeapPageSize() returns the correct value in these configurations:
 * a linux 4.4 system (using a custom comm-none that allocates huge pages).
 * Mac OS X 10.10.5 (using a custom comm-none that allocates huge pages and using SIGBUS instead of SIGSEGV)
 * a Cray XC with comm=gasnet tasks=fifo and with CHPL_RT_CALL_STACK_SIZE=128k or 2MB (this configuration did not work before)
 * a Cray XC with comm=gasnet tasks=muxed without a hugepages module loaded
 * a Cray XC with comm=ugni tasks=muxed with or without a hugepages module loaded

Verified that Hello World works in these configurations:
 * a Cray XC with GASNet or UGNI and tasks=muxed with or without a hugepages module
 * a 64-bit Cygwin system
 * a FreeBSD 10 system
 * an omnios (solaris clone) system

Passed full local testing.

Reviewed by @gbtitus - thanks!

PR #5008, opened 2016-12-12
Fix hugepages

mppf:
Fixes a problem with chpl_computeHeapPageSize that @benharsh noticed with CHPL_COMM=ugni that was introduced by PR #4931.

computeHeapPageSizeByGuessing seems to be causing some sort of problem. It was added for the case that CHPL_COMM=gasnet provides the heap on huge pages, but huge pages aren't configured. Now CHPL_COMM=ugni is running into problems when huge pages are requested. So, this PR simply changes to trusting the user if a huge page size is set (which we used to do).

- [ ] get Greg's feedback
- [ ] test with CHPL_COMM=ugni and no huge page module loaded
- [ ] test with CHPL_COMM=ugni and a huge page module loaded
- [ ] test with CHPL_COMM=gasnet and a huge page module loaded
- [ ] test with CHPL_COMM=gasnet and no huge page module loaded

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -311,7 +311,11 @@ void chpl_computeHeapPageSize(void) {
 #endif
 
   // note: sets heapPageSize
-  computeHeapPageSizeByGuessing(pageSize);
+  if (pageSize == chpl_getSysPageSize()) {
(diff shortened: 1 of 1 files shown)


mppf:
Closing in favor of PR #5009.

PR #5009, opened 2016-12-12
Move test page to the end of the heap

mppf:
This PR resolves a problem with CHPL_COMM=ugni where allocating and using a large amount of memory could result in some PUTs to fail. This issue was introduced in PR #4931.

In order to be sure we correctly know the page size of the Chapel heap, we experimentally verify the page size with mprotect in computeHeapPageSizeByGuessing. Before this PR, those experiments used memory in the middle of the heap. That has the unfortunate side effect of splitting the heap in the middle as far as the linux kernel is concerned - which complicates memory registration.

This PR changes the logic in computeHeapPageSizeByGuessing to use memory at the end of the heap instead. This memory is less likely to be used in practice. (However if it is used, there is still something to fix within CHPL_COMM=ugni).

While working on this PR, I noticed that the experiment with mprotect has an impact on the "first-touch" algorithm for moving memory between NUMA nodes. That is because it sets one byte of memory in addition to mprotecting. This effect too should be minimized by working with the end of the heap.

Tested with @benharsh's reproducer and the hellos on:
- CHPL_COMM=ugni and no huge page module loaded
- CHPL_COMM=ugni and a huge page module loaded
- CHPL_COMM=gasnet and a huge page module loaded
- CHPL_COMM=gasnet and no huge page module loaded

Tested hellos on:
- Tested hellos on Mac OS X with GASNet segment=fast.
- CHPL_COMM=gasnet and CHPL_TASKS=muxed and a huge page module loaded
- CHPL_COMM=gasnet and CHPL_TASKS=muxed and no huge page module loaded

(note that @benharsh's reproducer runs out of memory allocating array elements in gasnet+muxed configurations, but that also happens with master before PR #4931 so it must be an unrelated issue)

Reviewed by @gbtitus.

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -135,9 +135,9 @@ int check_page_size(size_t page_size_guess, size_t max_page_size,
   const int debug = 0; // set to 1 for debug prints
 
   if (debug)
-    printf("check_page_size(%li, %li, %p, %p, %li)\n",
+    printf("check_page_size(%li, %lx, %p, %p-%p)\n",
            (long) page_size_guess, (long) max_page_size,
-           ptr, heap_start, (long) heap_size);
+           ptr, heap_start, ((unsigned char*) heap_start) + heap_size);
 
   // set signal handler variables
   check_page_size_ptr = ptr + page_size_guess;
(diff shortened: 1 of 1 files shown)


gbtitus:
Looks good!

PR #5012, opened 2016-12-13
Error handling futures, CHIP 8 improvements

psahabu:
- updated the CHIP to include auto-propagation in Default mode
- added four futures to `test/errhandling`:
  - basic `try-catch`
  - matching errors to a `catch`
  - propagating an error with `try`
  - single-statement `try!`

### DIFF
--- a/doc/developer/chips/8.rst
+++ b/doc/developer/chips/8.rst
@@ -211,9 +211,13 @@ the compiler will run checks in one of two modes, based on a flag.
 
 If a call to a throwing procedure is not enclosed by a ``try`` or ``try!``: 
 
-1. **Default.**: the program will compile with a warning. An error will
-   be propagated by the calling procedure if it ``throws``, and will
-   ``halt()`` if it does not throw.
+1. **Default.**: 
+
+   a. if the call is contained in a procedure labelled with ``throws``, any
+      error encountered will be propagated.
+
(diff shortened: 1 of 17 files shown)


mppf:
Re. the tests, we'll need more tests for these cases:
 - [ ] make sure Error objects are freed appropriately
 - [ ] make sure other variables  in scope (e.g. arrays) are freed appropriately
 - [ ] exercise default and strict mode

but those additional tests don't have to be in this PR.

doc/developer/chips/8.rst
```diff
@@ -208,14 +209,15 @@ Default vs strict mode
...
-1. **Default.** If a call to a throwing procedure is not enclosed by a ``try``
-   or ``try!``, the program will compile with a warning and the call will
-   ``halt()`` on an error.
+If a call to a throwing procedure is not enclosed by a ``try`` or ``try!``: 
 
-2. **Strict.** If a call to a throwing procedure is not enclosed by a ``try``
-   or ``try!``, the compiler will raise an error.
+1. **Default.**: the program will compile with a warning. An error will
+   be propagated by the calling procedure if it ``throws``, and will
+   ``halt()`` if it does not throw.
```
psahabu: That's a fair point, if the developer wants warnings then they should compile in Strict mode. I'll update this to reflect that.

PR #4489, opened 2016-09-09
Add initial support for using native qthread sync variables

ronawho:
Add support to map int/uint/bool sync vars down to native qthreads aligned_t
sync vars. It's not enabled by default yet, because the current qthreads
release is missing some FEB operations. I contributed them upstream, and they
will be in the next release, at which point these sync vars can be enabled by
default.

For "simple" sync var types, we can directly use the aligned_t qthreads FEB
routines instead of using the heavyweight routines that we've traditionally
used. This support is loosely modeled after the fast syncvars that we used for
MTA. That code was added with 8503f6e removed with d7707439.

#3192 has more details on the history of sync vars and why the implementation
has traditionally been so heavyweight. Short version is that we used to allow
arbitrary types to be sync, but we're moving towards just primitives being
allowed.

Native sync vars should drastically improve the performance of highly contended
sync variables. e.g. threadring is over 6 times faster with native qthread
sync vars. The main reason for this is because qthreads will actually
de-schedule tasks that are blocked on a native sync var. Our other
implementation has to lock before any operation, and that lock is implemented
with a spin-wait. This often led to thrashing and bogging down the scheduler.

Future work:
 - extend support to single vars
 - extend support for more base types
   - currently integral and bool only, expand to all primitives <= 64-bit
 - better yielding policy for readXX
 - use native sync vars on 32-bit platforms for <= 32-bit base types
   - currently all aligned_t sync vars are broken on 32-bit platforms (qthreads bug)

See JIRA 249 for more info: https://chapel.atlassian.net/browse/CHAPEL-249

PR #5013, opened 2016-12-13
Remove old native qthreads sync var prototype code

ronawho:
Remove some old (and so far as I can tell never used) code in our qthreads shim
that attempted to map simple sync/single types down to native qthread sync
vars.

It looks like this was originally added to qthreads in https://github.com/Qthreads/qthreads/commit/c97ab1f.

It made use of the old "fast" sync var machinery that MTA used (https://github.com/chapel-lang/chapel/commit/8503f6e) but so
far as I can tell it was never enabled.

https://github.com/chapel-lang/chapel/commit/d770743 removed all the old mta code and https://github.com/chapel-lang/chapel/pull/4489 added true support native 
qthread sync vars, so this code is no longer needed.

### DIFF
--- a/runtime/include/tasks/qthreads/tasks-qthreads.h
+++ b/runtime/include/tasks/qthreads/tasks-qthreads.h
@@ -64,51 +64,6 @@ typedef struct {
     syncvar_t signal_empty;
 } chpl_sync_aux_t;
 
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus and maybe @bradcray mind verifying my thinking that this code is old and not used?

gbtitus:
I agree that it's unused -- I did a full grep for all of these and saw only the `#define` instances that you're removing.    👍

PR #5015, opened 2016-12-13
Clean up runtime include paths

sungeunchoi:
- Fix runtime include path for chpl-env-gen.h
- Removed bogus paths

@mppf 

Tested hello*.chpl with CHPL_COMM=none and gasnet.


ronawho:
👍   :)

PR #5016, opened 2016-12-13
Make libnuma available on Crays when needed

dmk42:
The memkind library requires libnuma, and there is no static version of either one of them.

On Crays, when the cray-memkind module is loaded (which also sets the default link mode to dynamic), we get libnuma whether we want it or not, so hwloc should take advantage of it.  This change does so for that case.

On Crays, when a non-flat locale model is used, we need hwloc to have access to libnuma, otherwise we will be unable to specify from which numa node a memory allocation should occur.  This change enables libnuma for that case as well.

This was tested on Mac, Linux, and Crays, including KNL with and without memkind, and with and without a flat locale model.

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -676,7 +676,11 @@ def set_up_environment():
         args.performance = True
         args.gen_graphs = True
         args.compopts += " --fast"
-        if not tgt_platform == "darwin":
+        # Force static linking unless the target requires dynamic.
+        if (tgt_platform != 'darwin' and
+            not (chpl_compiler.get('target').startswith("cray-prgenv") and
+                 (os.getenv("CRAY_MEMKIND", "") != "" or
+                  chpl_locale_model.get() != 'flat'))):
             args.compopts += " --static"
 
     if args.performance_configs:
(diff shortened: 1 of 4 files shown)


ronawho:
I'd actually prefer to let somebody else review this to see what think of this approach

gbtitus:
I will look at it.

util/start_test
```diff
@@ -676,7 +676,11 @@ def set_up_environment():
         args.performance = True
         args.gen_graphs = True
         args.compopts += " --fast"
-        if not tgt_platform == "darwin":
+        # Force static linking unless the target requires dynamic.
```
gbtitus: Reversing the condition to check it (and using liberal pseudocode): we won't force static linking iff:
  platform == darwin || (cray-prgenv && (isSet(CRAY_MEMKIND) || locModel!=flat))

That seems okay, but what about locModel!=flat when using something other than a cray-prgenv compiler?  Wouldn't we also want to use libnuma and thus dynamic linking if we were testing locModel=numa with hwloc on non-Cray linux64 systems?  To achieve this, should we change make/compiler/Makefile.<several-other-compilers> similarly to how we changed make/compiler/Makefile.cray-prgenv?  (But maybe this is a longer-term question and we should tackle it as numa matures?)

gbtitus:
I think this is right and good as it stands.  I made a line comment pertaining to non-Cray platforms but I don't think it's something that needs to be dealt with right now.  This seems like it will get the Cray case right.

util/start_test
dmk42: Long term, I agree that we will want libnuma other than on Crays.  Elliot was nervous about doing that, so I limited myself to Crays for this PR.

PR #5018, opened 2016-12-13
Improve task creation speed for "bounded" coforalls

ronawho:
Improve task creation speed for "bounded" local coforalls by limiting how many
times we touch the EndCount and runningTasks atomics.

Previously coforalls were turned into something roughly like:

```chapel
coforall i in 1..10 { body(); }
```

```chapel
for i in 1..10 {
  _upEndCount(_coforallCount);
  chpl_taskListAddCoStmt(body());
}
_waitEndCount(_coforallCount);
```

where `_upEndCount` bumped the atomics here.runningTasks, EndCount.i, and the
non atomic EndCount.taskCnt. Modifying the atomics for each iteration is
actually quite expensive so this changes the code to do a single upEndCount for
a coforall when we can determine the number of tasks the coforall will create
before executing it. Currently that check is if the coforall is over a
boundedRange, a domain, or an array.

We now create something like:

```chapel
var tmpIter = 1..10;
if isBoundRange(tmpIter) || isDomain(tmpIter) || isArray(tmpIter) {
  var numTasks = tmpIter.size;
  _upEndCount(_coforallCount, numTasks);
  for i in tmpIter {
    chpl_taskListAddCoStmt(body());
  }
  _waitEndCount(_coforallCount, numTasks);
} else  {
  // old style coforall
}
```

which allows us to only bump the EndCount and here.runningTaskCount once, and
allows us to avoid using taskCnt since we're now passing the number of tasks
created directly to waitEndCount.

Note that this does effectively disable the anonymous range/domain optimization
for coforalls since we're de-anonymizing the iterator. That's unfortunate, but
the performance benefit of not manipulating atomics far outweighs the cost of
constructing a range/domain.

This has a pretty significant performance improvement for several benchmarks:

 - 50% speedup for forall/coforall task microbenchmarks (within 20% of gomp now)
 - 25-35% speedup for short parallel LCALS
 - 10% speedup for long LCALS, lulesh, miniMD, NPB FT, and a few others

Some future follow-ons to this include:
 - apply this optimization to coforall+on stmts
 - come up with a better way to detect "bounded" coforalls instead of just
   checking for boundRange/domain/array
 - figure out a way to still apply the anonymous range/domain optimization
 - have "bounded" coforalls call into an optimized runtime routine that spawns
   multiple tasks at once instead of calling chpl_taskListAddCoStmt once per
   iteration.
 - cleanup the code for this optimization (this adds a some duplicate code/logic)

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -1581,19 +1581,63 @@ BlockStmt* buildCoforallLoopStmt(Expr* indices,
     onBlock->insertAtTail(new CallExpr("_downEndCount", coforallCount));
     return block;
   } else {
-    VarSymbol* coforallCount = newTemp("_coforallCount");
-    BlockStmt* beginBlk = new BlockStmt();
-    beginBlk->blockInfoSet(new CallExpr(PRIM_BLOCK_COFORALL));
-    addByrefVars(beginBlk, byref_vars);
-    beginBlk->insertAtHead(body);
-    beginBlk->insertAtTail(new CallExpr("_downEndCount", coforallCount));
-    BlockStmt* block = ForLoop::buildForLoop(indices, iterator, beginBlk, true, zippered);
-    block->insertAtHead(new CallExpr(PRIM_MOVE, coforallCount, new CallExpr("_endCountAlloc", /* forceLocalTypes= */gTrue)));
-    block->insertAtHead(new DefExpr(coforallCount));
-    beginBlk->insertBefore(new CallExpr("_upEndCount", coforallCount));
-    block->insertAtTail(new CallExpr("_waitEndCount", coforallCount));
-    block->insertAtTail(new CallExpr("_endCountFree", coforallCount));
-    return block;
+
+    BlockStmt* coforallBlk = new BlockStmt();
+
+    BlockStmt* vectorCoforallBlk = new BlockStmt();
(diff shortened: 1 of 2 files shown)


ronawho:
Link to improvements in perf playground: http://chapel.sourceforge.net/perf/chapcs/vectorCoforall/?startdate=2016/12/08&enddate=2016/12/12&graphs=lcalsrawompshort,lcalsrawompmedium,lcalsrawomplong,luleshrelease,luleshbradcstudyversion,minimdljsize10time,blackscholesparsecbenchmark,nasparallelbenchmarksfttimingssizea,emptytaskspawntimings500000xmaxtaskpar,emptyserializedtaskspawntimings500000xmaxtaskpar,reductionstimesec,timetobuildtrees

ronawho:
I recommend ignoring the whitespace when reviewing.

ronawho:
@benharsh mind reviewing?

benharsh:
Looks good to me! Those perf changes are great!

bradcray:
This desire to be able to generally identify bounded iterator cases and their trip count seems like it comes up reasonably often -- e.g., we could do better bounds checking for zippered foralls than (the none) we do today in the event that the things being zippered could be queried and identified as being bounded.  Similarly, the check for zippered for loops could be done prior to the loop executing rather than afterwards, which might be desirable.

PR #5019, opened 2016-12-13
Moved lowering of PRIM_DELETE to resolve();

vasslitvinov:
Moved lowering of PRIM_DELETE to resolve();
replaced compiler-generated AST with a Chapel function, chpl__delete().

This is motivated by the upcoming ForallStmt node that will be lowered
during resolve(). It requires "delete" of iteration classes, so
inability of generate PRIM_DELETE during resolution would be a hindrance.

* Renamed "~chpl_destroy" to "chpl__deinit". This is so that it can be
invoked from Chapel code, specifically from chpl__delete().

* PRIM_DELETE now lowers directly to a call to chpl__delete() in preFold().
  Alternatively we could generate a call to chpl__delete() right in the parser
and so dispense with PRIM_DELETE. I did not do that so that 'delete'
remains as a primitive operation, in case we want to do something about it
before resolution.

* Some destructors are written by users without parens. We decided to allow
that - see the discussion in this PR #5019.
  There is also the thread "parens required for deinitializers"
on the internal list.

* For that, have the compiler add parens internally if the user didn't.
This is implemented by removing FLAG_NO_PARENS.
  This loses one bit of information about the user program, which
we may want to preserve in the future.
  I did it this way so that the call arg.chpl__deinit() in chpl__delete()
resolves.  Since the call is written with parens, it requires
chpl__deinit() to be written with parens as well.
  To allow paren-less chpl__deinit, we can either use reflection to try
paren-less call first, then fall back to paren-ful call in chpl__delete().
Or we can add some logic to the compiler/resolution.

* Reduced comm counts (gets) in a couple of tests.

* Potential future improvement - better error message when 'delete' is
invoked on a non-class.
  Right now the cases of records and _ddata are caught explicitly.
However, the cases like delete of an integer result in a "unresolved
call to chpl__deinit()" error, with a ton of chpl__deinit candidates
printed out.
  If seems easy to just test for isClass(arg). However that does not
work because isClass() returns false for "extern" classes, whereas we
DO delete those in our tests.
  This all is unaffected by this PR.

* Move the INT_ASSERT check that a yield is always within an iterator
from normalize() to check_checkParsed(). Note that if the user made
an error, checkParsed issues USR_FATAL_CONT, so the check is violated
and the compiler has not aborted yet. Force exitIfFatalErrorsEncountered()
before running that check.

Testing:
+ linux64 full suite
+ valgrind full suite (slightly earlier version)
+ gasnet on multilocale tests
+ memleaks on test/release/examples


ronawho:
Note that shootouts in the "submitted" directory should not be modified. They are intended to let us know if changes will break the entries we've submitted to the CLBG site.

vasslitvinov:
@ronawho Thank you.

In general, what do we do with the "submitted" directory when we make breaking changes in the language?

ronawho:
In the past we have notested them and you definitely need to make @chapel-lang/perf-team aware so we know that we need to resubmit when the CLBG site upgrades versions

It's probably worth opening an issue to track what benchmarks will have to be resubmitted

compiler/passes/normalize.cpp
```diff
@@ -77,33 +77,13 @@ void normalize() {
   transformLogicalShortCircuit();
   lowerReduceAssign();
 
-  // tag iterators and replace delete statements with calls to ~chpl_destroy
+  // tag iterators and replace delete statements with calls to chpl__deinit
```
benharsh: Is this comment still accurate? It looks like the replacing part was removed.

benharsh:
Have you done any memleaks testing?

Otherwise, this looks fine to me.

bradcray:
[tagging @noakesmichael and @mppf in this discussion due to their involvement in initializers design and implementation]  

Maybe I'm just forgetting, but what is the plan for destructors in the new initializers world?  (checking the CHIP, I see it's deinit() though I can't remember if this ever came up in group discussion... it seems unfamiliar).

From a language flexibility perspective, it seems like it _could_ be nice to permit destructors to either take parens or not, at the user's choice, if they take no arguments (and presumably most won't in the common case?).  Rather than requiring this change, e.g., could the build.cpp routine that creates methods detect that it's a destructor method, detect that it has no parens, and add the parens for the user rather than flagging an error?  This might only make sense in the current/old world (where destructors can't be called manually, right?) than in the new (where presumably they can be and should therefore be called with / without parens as required by the declaration).

This wasn't called out explicitly in our 1.15 priorities due to an oversight on my part, but hopefully was implicit:  It'd be ideal to have 1.15 go out with new destructors as well as initializers.  Who should own that?  (Vass, do you have cycles given that you're currently working in this area and also have an initializer-related priority in the release?)  W.r.t. the need to update a shootout, it'd be nice to go directly to the new destructor syntax rather than add parenthesis in one release and then switch to new destructors in the next.

Perf team: Anyone remember if this destructor is formally required or if we could just leave it out altogether?  Memory leaks testing wouldn't be happy as we continue to drive it to zero, but if we had a mechanism for permitting exceptions (which we might want/need for other tests that intentionally check leaks and such...) maybe that's OK.

lydia-duncan:
I believe the thinking on deinit() was that it would be identical to ~className() but would following a naming style that matched init().  With that in mind, it would be extremely easy to add support for this.  I just haven't prioritized it.

lydia-duncan:
And it is something that anyone could do.  There are some futures for it in the test/classes/initializers/ directory.

bradcray:
In some instances, users have wanted the ability to call destructors manually, like any other procedure, along with having them called automatically by the compiler on delete.  My guess was that in making the destructor name more of a normal identifier, this also would permit users to call them if they wanted to.  I don't think users can call our current destructors because of the syntax ambiguity with the ~ operator in the calling context (but could be wrong about that?).

lydia-duncan:
I don't have a problem with deinit supporting that use case, personally

bradcray:
Thinking ahead to the 1.15 release, if there were a sneaky way to keep paren-less destructors working without source changes (like the build.cpp manipulation I suggested above), that seems more attractive than both restricting what's accepted by the compiler while deprecating the feature...

vasslitvinov:
Thanks @bradcray - I will allow paren-less destructors by having the compiler add parens post-parser. That way the tests themselves can stay unchanged.

With this PR, it should be trivial to allow user args on delete. There is ambiguity whether "delete x(args)" means "delete x" with (args) or it means "temp = x.this(args); delete temp".

Renaming of ~className to deinit seems straightforward. Allowing users to invoke deinit() directly - just like invoking init() directly - is a bigger design discussion.

bradcray:
In my proposal that deinit should be permitted to take args, I'm not sure I think that a user should necessarily be able to supply those args on a delete call.  Rather, what I was imagining was that a user could write:

    class C {
      proc deinit(verbose=false, debug=false) {
         // frees some resources or something
      }
    }

and that calls to delete would still have to look like:

    delete myC;

but that a user could also write any of the following (prior to deleting, of course), as a means of having the code in deinit() run without deallocating the object / its memory:

    myC.deinit();
    myC.deinit(true, true);
    myC.deinit(debug=true);

(and ditto for records except that the call to delete would be taken care of by the compiler in the normal place).

Maybe there's a precedent for calling delete on an object with args that I'm not aware of though...?

lydia-duncan:
If we supported an argument version for records, we'd also want to ensure that calling deinit(<args>) on a record early prevents the compiler from inserting delete on it later, or you'd get a double delete.

bradcray:
No, the idea is that calling deinit() executes the code within deinit() but doesn't actually deallocate the memory/object.  So for a class, the delete would still need to do so; for a record, the compiler would still free it as normal.

bradcray:
(Expounding on my previous response: As I understand it, this can be useful when one's destructor frees some resources and you want to release those resources without freeing the object).  The alternative requires either some code duplication:

    proc freeResources() {
      // free some resources
    }
    proc ~C {
      // free the same resources
    }

or some somewhat convoluted code patterns:

    proc freeResources() {
      // free some resources
    }
    proc ~C {
      freeResources();
    }

where having a callable destructor supports both cases with less typing:

    proc deinit() {
      // free some resources
    }

(though clearly you could still do either of the previous patterns as well...)

noakesmichael:
I think it’s unfortunate that there are so many situations in which the compiler alters
what the user actually said and therefore I think it’s unfortunate to be adding another
instance of this.


On Dec 13, 2016, at 3:41 PM, Vassily Litvinov <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks @bradcray<https://github.com/bradcray> - I will allow paren-less destructors by having the compiler add parens post-parser. That way the tests themselves can stay unchanged.

With this PR, it should be trivial to allow user args on delete. There is ambiguity whether "delete x(args)" means "delete x" with (args) or it means "temp = x.this(args); delete temp".

Renaming of ~className to deinit seems straightforward. Allowing users to invoke deinit() directly - just like invoking init() directly - is a bigger design discussion.

bradcray:
I suspected you'd feel that way and would feel more bad about it if it weren't a feature that we were planning to deprecate in this release and remove in the next one...

PR #5024, opened 2016-12-14
Upgrade release fasta ; reduce diffs between versions

bradcray:
This commit upgrades the release version of fasta to incorporate elements from the fasta-blc.chpl version (notably the parallelism) along with several stylistic changes made to prepare it for submission.  It also updates the fasta-blc[-numa].chpl versions to reduce the diff with the release version.

As part of this change, I switched the arrays of random numbers over to use 'randType' which is an alias for a 'uint(32)' and this had a positive performance benefit of reducing the execution time from ~2.7s on the shootout box to ~2.0s.  However, in further simplifying/clarifying the loop structure of the main 'for i in..." work chunking loop, I switched it to use zero-based indexing which eliminated the need to initialize the atomic arrays to 1 for the initial condition.  For reasons I can't explain, this bumped the execution time up to ~2.2s on the shootout machine.  The only explanation I can come up with is that that parallel array assignment somehow warms up the tasking interface...or something.  In the study versions, I left this initialization in (re-initializing the atomics to 0) to capture the difference on the shootout box and see how it behaved on other systems.

One other change I made, but which didn't work, was to set up a 'config type' for 'char' aliasing it to 'int(8)'.  It seems that having two config types in a single statement causes an internal error in the compiler, so I backed away from this change and filed a bug and issue #5039 against it instead.


bradcray:
@chapel-lang/perf-team : I'd like to update the release version of fasta to this one in preparation for submission to the CLBG site.  Who feels invested enough in fasta/shootouts to want to review?

bradcray:
P.S. -- We could also review in the weekly meeting, but we've got a couple other things on the agenda, and I'd ideally like to submit it before then in hopes of having my desk clear by the end of the week.

test/release/examples/benchmarks/shootout/fasta.chpl
```diff
@@ -96,47 +101,93 @@ proc repeatMake(desc, alu, n) {
...
+  var randGo, outGo: [0..#numTasks] atomic int;
+  randGo.write(1);
+  outGo.write(1);
+
+  coforall tid in 0..#numTasks {
+    const chunkSize = lineLength*blockSize,
+          nextTask = (tid + 1) % numTasks;
+
+    var myBuff: [0..(lineLength+1)*blockSize-1] int(8),
+        myRands: [0..chunkSize] int/*(32)*/;
```
benharsh: Two minor things:
1) I think you could use the "#" operator in ``myBuff``'s domain.
2) Do you want that ``/*(32)*/`` to be there?

test/release/examples/benchmarks/shootout/fasta.chpl
```diff
@@ -96,47 +101,93 @@ proc repeatMake(desc, alu, n) {
...
+
+    for i in 1..n by chunkSize*numTasks align 1+tid*chunkSize {
```
benharsh: I think this would mark the first use of an 'align' in the shootouts. Is it worth writing in a different way, or adding a comment to explain it better?

test/release/examples/benchmarks/shootout/fasta.chpl
```diff
@@ -96,47 +101,93 @@ proc repeatMake(desc, alu, n) {
...
+        off += 1;
+      }
+
+      // Write the output in a coordinated manner
+      wait(outGo);
+      stdout.write(myBuff[0..#off]);
+      signal(outGo);
+
+      // Helper routines for taking turns with shared resources
+      inline proc wait(guard) {
```
benharsh: It might be interesting to try out the ``waitFor`` method to see if performance is different, but this is fine by me for this PR.

test/release/examples/benchmarks/shootout/fasta.chpl
```diff
@@ -96,47 +101,93 @@ proc repeatMake(desc, alu, n) {
 }
 
 //
-// Output a random sequence of length 'n' using distribution a
+// Output a random sequence of length 'n' using distribution 'a'
```
benharsh: Just to sanity-check this comment: What does "distribution 'a'" refer to? I don't see a variable named 'a' anywhere.

benharsh:
Otherwise, it looks good to me!

test/release/examples/benchmarks/shootout/fasta.chpl
bradcray: 1. Good call.  It took me awhile to get that bound right, and by the time I did...
2. Jessh, no.  Worse, I think I want randType to save space...  I wonder if that's any part of the perf difference.  Generally, I haven't seen much effect in this benchmark when switching between 64- and 32-bit ints, but there are a lot of moving parts...
Thanks!

test/release/examples/benchmarks/shootout/fasta.chpl
bradcray: I was mentally wrestling with this loop yesterday as well.  What I like about it is that it essentially says "We're going from 1..n" what I don't like about it is that the pattern is so carefully orchestrated but unclear.  I found myself wishing for an iterator from the standard modules that would help me with this.  Something like 'for i in roundRobin(1..n, chunkSize, tid, numTasks).'  This relates a bit to the RangeChunking library yet is different (I believe).  That said, it's also intertwined with the semi-magic numbers that are stored into the atomics to pass the token which seems similarly unfortunate...

All that said, without a library routine, the other way I was thinking of writing it yesterday was:

    for 1+tid*chunkSize..n  by chunkSize*numTasks

if you think that's any better.  I'm not sure it is so maybe the comment is the way to go for now.

benharsh: I guess I prefer the alternate form over the align version. I think using align in this case would be confusing, and it seems like a comment explaining the align would be somewhat long. The alternate form seems clearer to me.

If you prefer the align, I won't hold up the PR over it.

test/release/examples/benchmarks/shootout/fasta.chpl
bradcray: Interestingly, this PR already had changed an array (cumulProb) from ints to randTypes and that resulted in a 25% performance improvement over my previous timings which I hadn't anticipated.  Sadly, changing myRands[] to randType only improved it another 3% (but is still the "right" thing to do in terms of the types we want to be storing).

test/release/examples/benchmarks/shootout/fasta.chpl
bradcray: OK, given that I was stewing on this yesterday myself, I'm making the change.  I'm also finding that switching this loop to 0-based indexing cleans up some things like removing the need for initializing the arrays of guards and some +/-1 style math.

test/release/examples/benchmarks/shootout/fasta.chpl
bradcray: Yeesh, no idea.  Maybe it used to be called 'a' and I improved it?  Where's my semantic comment checker?

benharsh:
Looks good, thanks!

PR #5027, opened 2016-12-14
Pre-exec files should check CHPL_TEST_ROOT_DIR before CHPL_HOME

awallace-cray:
11 pre-exec files are changed to check env var CHPL_TEST_ROOT_DIR
before CHPL_HOME, when constructing an absolute path to a test
file.

This change is needed to permit Cray module tests to work without
copying the entire test tree into the installed Chapel module.


benharsh:
Looking at it again, I think it needs a minor tweak. We usually set CHPL_TEST_ROOT_DIR to point to the actual test/ dir. That means the preexec will fail when trying to find $CHPL_TEST_ROOT_DIR/test/..., because test/test/ doesn't exist.

awallace-cray:
FYI, There are no other preexec's that used CHPL_HOME. These are the only ones.

benharsh:
My local testing came back clean, so feel free to merge 👍

PR #5033, opened 2016-12-14
Verify less

mppf:
* Move verification of Symbol's list of SymExpr to checks.cpp
* Enable --verify --verify to do extra verification after every pass

Now the slow verification only occurs after functionResolution with --verify,
but if --verify --verify is used, it occurs after every pass.


vasslitvinov:
How about a different way to implement this check:

* A bit set of one bit per SymExpr in gSymExprs, say "seen".

* for each symbol "sym" do for_SymbolSymExprs(se, sym) do check:

(a) se->symbol() == sym

(b) the bit in "seen" for "se" is not set yet, then set that bit.

* Ensure that all bits in "seen" are set.

Separately, I am OK with doing this check just once upon --verify.

mppf:
Closing in favor of PR #5040.

PR #5040, opened 2016-12-15
Faster SymExpr::verify

mppf:
Verify Symbol's SymExpr list using local operations

 * each Symbol checks that its first and last list elements are in the tree and have the same Symbol
 * each SymExpr checks that its prev and next elements, if any, are in the tree and have the same Symbol. If there is no prev/next element, check that this SymExpr is the head/tail of the appropriate Symbol's list.

I've checked that these rules catch likely invalid possibilites, such as:
 - setting SymExpr::var without using setSymbol()
 - inappropriately calling Symbol::removeSymExpr()
 - inappropriately calling Symbol::removeSymExpr() and then
   Symbol::addSymExpr() with the wrong Symbol.

These rules produce better checking than we had before. Moreover, I think that one could prove that these rules ensure the following properties:

 - every live SymExpr is in the appropriate Symbol's list
 - no Symbol's list contains a SymExpr with the with a different symbol()

Passed full local --verify testing.

### DIFF
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -513,18 +513,32 @@ void SymExpr::verify() {
   if (var != NULL && var->defPoint != NULL && var->defPoint->parentSymbol == NULL)
     INT_FATAL(this, "SymExpr::verify %12d:  var->defPoint is not in AST", id);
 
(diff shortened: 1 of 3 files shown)


benharsh:
Looks good to me! 👍

PR #5045, opened 2016-12-15
Add some annotations for recent events

bradcray:
I focused on those that affected fasta and LCALS for the sake of
tweeting and things I improved.

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -207,6 +207,8 @@ all:
     - Stack allocate on and task function argument bundles (#4074)
   12/08/16:
     - Re-enable denormalization (#4547)
+  12/12/16:
+    - Reduce task creation time for "bounded" coforalls (#5018)
 
 
 AllCompTime:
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho : Would you review?

test/ANNOTATIONS.yaml
```diff
@@ -487,6 +495,8 @@ LCALSMain:
   10/20/16:
     - Clean up to HYDRO_2D kernel (#4757)
     - update to TRAP_INT serial kernel (#4756)
+  12/09/16:
```
ronawho: should be 12/10/16

ronawho:
Think you have the "commit" dates for most things instead of the "affected" dates: https://github.com/bradcray/chapel/blob/8f730c2c3b0e01c29b4a7101a45f4538a5fa7fce/test/ANNOTATIONS.yaml#L26-L29

test/ANNOTATIONS.yaml
```diff
@@ -207,6 +207,8 @@ all:
     - Stack allocate on and task function argument bundles (#4074)
   12/08/16:
     - Re-enable denormalization (#4547)
+  12/12/16:
```
lydia-duncan: I think this one should be 12/13/16 instead.  Multilocale testing worked that night and so the change turns up there, and there was a different set of tests impacted by commits from the 13th.

At least, that's only if I'm right in remember that we enabled annotations on days there isn't data (which I thought I saw go by a while ago, but I may be misremembering)

lydia-duncan: Oh, nevermind about the 13th, there's a different commit for those (I just verified that the coforall one went in on the 13th)

test/ANNOTATIONS.yaml
bradcray: Sigh...

PR #5047, opened 2016-12-15
If we might have sub-locales, use multiple ddata blocks for arrays.

gbtitus:
Add support for multiple ddata chunks when the locale model may have
sublocales.  When sublocales may be present, choose one of the array's
dimensions to chunk the storage on, split that dimension's values more
or less evenly into #sublocales chunks, and associate the chunks of
index values with chunks of array ddata storage, through a reference
vector.  This will be a building block for NUMA-aware array storage,
since it will allow for explicit affinity between index values (thus
execution locality) and NUMA array storage locality, as opposed to the
implicit affinity we get largely by luck, from first-touch and other
techniques.  At the same time, make careful use of params to separate
the simple-ddata and multi-ddata cases, because a major goal here is to
leave performance unchanged for locModel=flat (the simple-ddata case).
Adjust bulk transfer, I/O, distributions, and test results as needed.


gbtitus:
The multi-ddata PR is up!  @benharsh, at your convenience could you start reviewing, and @ronawho , could you arrange to perf-test the corresponding branch in the sandbox, using the locModel=flat, comm=none configuration? (We're primarily concerned initially that it not hurt the existing performance for flat.)

This has a couple of conflicts on test results; do you need me to resolve those before you do anything, or can that stay for now?  (I'm working on figuring out the 90 or so "stress" test failures.)

ronawho:
It's a lot easier for me if the branch merges cleanly with master

modules/dists/BlockDist.chpl
```diff
@@ -1568,6 +1613,60 @@ proc BlockArr.doiCanBulkTransferStride() param {
   return useBulkTransferDist;
 }
 
+proc BlockArr.oneDData {
```
benharsh: What do you think of adding a quick check against ``defRectSimpleDData`` here, in case someone calls this function without check that param first?

modules/internal/DefaultRectangular.chpl
```diff
@@ -809,6 +1013,241 @@ module DefaultRectangular {
...
+              yield dd(i);
+            }
+          } else if dom.dsiNumIndices > 0 {
+            const (chunkLo, _) = getDataIndex(dom.dsiLow);
+            const (chunkHi, _) = getDataIndex(dom.dsiHigh);
+
+            // first chunk
+            {
+              const dd = theDataChunk(chunkLo);
+              const (_, hi) = mdChunk2Ind(chunkLo);
```
benharsh: I take it that mdChunk2Ind doesn't return dom.dsiLow/dsiHigh correctly? Is that something that could be done at a later time, or is it technically not feasible?

modules/internal/DefaultRectangular.chpl
```diff
@@ -809,6 +1013,241 @@ module DefaultRectangular {
...
+      }
+    }
+
+    iter these(param tag: iterKind,
+               tasksPerLocale = dataParTasksPerLocale,
+               ignoreRunning = dataParIgnoreRunningTasks,
+               minIndicesPerTask = dataParMinGranularity)
+      ref where tag == iterKind.standalone && !defRectSimpleDData {
+      if debugDefaultDist {
+        chpl_debug_writeln("*** In array standalone code");
```
benharsh: It might be good to make this message more specific and state that it's in the multi-ddata path. Though I suppose if you're debugging DefaultRectangular you probably know... oh well, consider it optional.

modules/internal/DefaultRectangular.chpl
```diff
@@ -848,32 +1294,127 @@ module DefaultRectangular {
...
+          } else {
+            mdParDim = parDim;
+            mdNumChunks = numChunks;
+          }
+        }
+        mdRLo = dom.dsiDim(mdParDim).alignedLow;
+        mdRHi = dom.dsiDim(mdParDim).alignedHigh;
+        mdRLen = dom.dsiDim(mdParDim).length;
+        dataVec = _ddata_allocate(_ddata(eltType), mdNumChunks);
+        for i in 0..#mdNumChunks {
```
benharsh: Can we add a comment or TODO here describing the future intention of making sure we allocate in a locality-aware fashion?

modules/internal/DefaultRectangular.chpl
```diff
@@ -848,32 +1294,127 @@ module DefaultRectangular {
...
+          return (0, sum);
+        } else {
+          const mdpd = mdParDim;
+          const mdnc = mdNumChunks;
+          const mdrlo = mdRLo;
+          const mdrhi = mdRHi;
+          const mdrlen = mdRLen;
+          const chunk = mdInd2Chunk(ind(mdpd), mdnc, mdrlo, mdrlen, str);
+          const (lo, _) = mdChunk2Ind(chunk, mdnc, mdrlo, mdrhi, mdrlen);
+          return (chunk, sum - (lo - mdrlo) * blk(mdpd) / str);
```
benharsh: It seems a little verbose to copy these members into consts like this.

benharsh:
That's my first pass of the core module changes. Overall, it looks good to me. It's a lot cleaner than I expected given that everything needs to be stuffed into DefaultRectangular classes. While I think there are some potential code cleanup opportunities in the future, I think it would be fine to leave that as a TODO.

Nothing jumped out at me as wrong, but I didn't evaluate the slicing/reindexing/rank-change logic too carefully. I may revisit if you find some issues while looking into the remaining problematic tests.

Also, thanks very much for splitting things up into logical commits!

gbtitus:
I rebased to master and the branch merges cleanly now.

modules/dists/BlockDist.chpl
gbtitus: Good idea on the quick-check -- I'll add that.
Thought about computing this at setup time.  At the time I was thinking it might be more coding, but now that I've seen the version I did write I'm realizing I might be wrong about that.  For now I'll probably just modify the TODOs to note this alternative.

modules/internal/DefaultRectangular.chpl
gbtitus: I was actually trying to save a few computations -- thinking that since we have dom.dsiLow already and lo==dom.dsiLow in this (rank==1) case we don't need to compute lo.  And the same for the last case, and dom.dsiHigh.  But now that I look at the code it's unnecessarily complicated and the savings, if any is probably miniscule.  I'll make a note to review this section, though I'll probably delay actually doing that until later.

modules/internal/DefaultRectangular.chpl
```diff
@@ -809,6 +1013,241 @@ module DefaultRectangular {
...
+          followMe(mdParDim) = lo..hi;
+          const (numChunks2, parDim2) = _computeChunkStuff(numSublocTasks,
+                                                           ignoreRunning,
+                                                           minIndicesPerTask,
+                                                           followMe);
+          if debugDataParMultiDData then
+            chpl_debug_writeln("### multiDD: chunk ", chunk,
+                               ", followMe ", followMe,
+                               ", numChunks2 ", numChunks2);
+          coforall chunk2 in 0..#numChunks2 do local on here {
```
gbtitus: I think the 'local' is just saying that the on-stmt won't do anything across the network.  I don't believe it precludes referencing other sublocales, though.

modules/internal/DefaultRectangular.chpl
gbtitus: I originally did this to reduce the number of GETs for references to non-distributed remote arrays.  Based on a little bit of new testing, it now looks like this is unnecessary, so I'll undo this.

mppf:
Just wondering (I didn't look at the code changes). We have some support for passing arrays to extern functions that works by assuming that the array is stored contiguously in memory. E.g. see https://github.com/chapel-lang/chapel/blob/master/modules/internal/CPtr.chpl#L225

What impact will this PR have on code that uses that feature? If it makes it no longer work in NUMA configs, I think that should be noted in the PR message at least.

gbtitus:
@benharsh I've put in a number of fixes and rebased. I think at least the code is ready to look at as a "final" candidate now.  The changes I made will require some corresponding changes in test .good files where for example we print or count comms on the array meta-data.  However, I don't expect the code to change a lot from here on out.  (There are a few failing tests outstanding, which reflect problems, but I'm hoping none of them will affect nightly testing.)  I've started the first of the runs to check on the state of the .good files for all the common configurations.

benharsh:
These recent changes look good to me! Provided that testing comes back sufficiently-clean, I think this can be merged at your discretion.

PR #5048, opened 2016-12-15
New ISx study version

benharsh:
This is a study version that should demonstrate better multi-locale performance when compared to the release version. Here are some of the modifications:
- use local blocks to squash wide pointer overhead
- serialize potentially-parallel things
- use local ranges instead of remote domains to avoid on-statements
- disables lazy RAD init for Block, which reduces comm counts


benharsh:
@bradcray : Interested in reviewing?

The --no-bounds-checks flag is thrown to avoid local-block errors due to bounds checking with a remote domain.

bradcray:
Interested, but realistically I shouldn't before the new year...  Maybe an opportunity to get someone else ramped up on ISx and its optimizations?

benharsh:
Anyone on @chapel-lang/perf-team interested?

lydia-duncan:
If this is still open on Tuesday, I will take a look, but I'm pretty swamped today

lydia-duncan:
:+1:

PR #5050, opened 2016-12-16
Separate the CG and MG largest sizes run into their own .graph files

lydia-duncan:
This will enable us to selectively include only the largest size in our
performance tracking suite, since we don't need the repetition of seeing all
the sizes when we spot check.

Verified that --generate-graphs still works with these changes


lydia-duncan:
@bradcray, care to review?  I don't think I'll have time to merge tonight, so no rush.

bradcray:
Assuming it still builds and runs, this looks right to me.

lydia-duncan:
Double sanity checked a fresh checkout built, ran, and still had the new graphs in the NPB suite.

PR #5057, opened 2016-12-16
Overload specific

mppf:
This PR includes compiler and parser changes to make a program like the following compile and run:

```
record R {
  var x;
}

proc R.foo() {
  writeln("in generic R.foo");
}

proc (R(int)).foo() {
  writeln("in specific R(int).foo");
}

var x = new R(1.0);
x.foo(); // prints "in generic R.foo"

var z = new R(2);
z.foo(); // prints "in specific R(int).foo
```

Note the parentheses required around (R(int)) in the method declaration - those prevent parser conflicts (because without them, the parser would have to consider an unbounded number of tokens to know the difference between a procedure declaration or a specific method:
  * specific method: e.g.  `proc R( ... )` in `proc R( ... ) . methodname()`
  * procedure definition: e.g. `proc R( ...)`

Passed quickstart testing.
Passed full local testing.
Passed new tests and hellos with --verify.

Future work:
 * Allow local field/method access in the specific method without the `this` keyword. The `this` keyword is currently required in this case, because scopeResolve adds `this` but runs before the type of the receiver is known in this case. This could be fixed by moving the logic adding `this` into functionResolution.
 * Get a type declaration working as an alternative. It does not work now. I think the language specification is fairly clear in section 7.6 that `type A = B` creates a type alias (vs creating a new type).  However, it does describe the issue as an *Open issue*. Below is an example. My personal opinion is that the delegation feature that I'm working toward addresses the desire to have `type A = B` where `A` can have new methods that aren't in `B`.
```
record R {
  var x;
}
proc R.foo() {
  writeln("in generic R.foo");
}
type R_int = R(int);
proc R_int.foo() {
  writeln("in specific R(int).foo");
}
var x = new R(1.0);
x.foo(); // prints "in generic R.foo"
var z = new R(2);
z.foo(); // prints "in specific R(int).foo
```


lydia-duncan:
This looks good to me, assuming it passes testing, too!  I don't think more than the normal full paratest is required unless you feel so motivated

PR #5058, opened 2016-12-16
Forwarding feature

mppf:
Add a method forwarding feature to Chapel. I started implementing this as a first step towards constructing wrapper records for reference counting or other strategies of managing class lifetime.

Enables code like the following:

``` chapel
record Wrapper {
  var instance; // e.g. some class
  forwarding instance;
  proc foo() { writeln("in Wrapper.foo()"); }
}

class C {
  var field:int;
  proc foo() { writeln("in C.foo()"); }
  proc bar() { writeln("in C.bar()"); }
  proc baz() { field = 1; }
}

var r = new Wrapper(new C());
r.foo(); // direct method shadows delegated method
r.bar(); // same as r.instance.bar(), prints "in C.foo()"
```

Note that in addition to this simple example, it supports:
 * filtering of forwarded methods with `only` and `except`
 * multiple forwarding declarations
 * `forwarding var instance` short-form declaration.

Conflicts with forwarded methods are handled like this:
 * regular (non-forwarded) methods take priority
 * forwarded methods will otherwise participate in normal overload resolution
 * ambiguous forwarded methods will cause a resolution error at the call site

This PR goes about these changes in the following manner.

First, it adds a ForwardingStmt to the AST. This exists during parsing and through resolution. After resolution, it no longer exists.

In build.cpp, add support for saving the forwarded-to expression in a function. The DefExpr for this function hangs off of the ForwardingStmt AST node and the name of it is stored there too.

In functionResolution.cpp, in resolveNormalCall, types subject to forwarding run populateForwardingMethods for the particular call being resolved. That in turn will create wrapper methods for that call name based upon the available method for the type being forwarded to. Then, resolveNormalCall tries some parts computing visible functions and candidates again since we've added methods that are relevant. Then, the new function will be available for the usual candidate resolution and disambiguation process.

In normalize.cpp, in insertRetMove, I needed to check for FLAG_MAYBE_REF instead of FLAG_IF_EXPR_FN so that the functions returning the forwarded-to expression can return a ref if the forwarded-to expression is a ref (see d-record-to-record-nocopy.chpl for a test exercising this).

Note that currently, forwarding wrapper functions are marked with FLAG_COMPILER_GENERATED in order to give preference to normal methods.

Future Work:
 * document in a technote - issue #5682 notes requirement to document in a technote
 * consider supporting `only` and `except` on the short `forwarding var` form.

Testing
 * hellos and test/classes/ferguson/forwarding with:
    - GASNet
    - --baseline
    - --verify
    - --baseline --verify
    - valgrind
 * full local testing


mppf:
C++ accomplishes some of these things with user-defined coercions (aka implicit conversions in C++ lingo). But, I'd consider these implicit conversions an experts-only feature. In particular, it's easy to imagine being surprised by the way a variable of one type suddenly turns into another type. The type system keeps it straight, but I don't expect most C++ programmers actually understand *why* they type system has the rules it does and how they prevent confusion in the face of implicit conversions.

In contrast, a delegation feature matches something a programmer could do explicitly - just with less typing (and doing it in a less brittle / more generic way). Additionally, the delegation feature allows selective delegation. In contrast, if we had ContainsAnA(A) coercible to A, then *all* methods of A are available on ContainsAnA (at least as long as coercions occur on the receiver of methods - which Chapel does now and so does Scala, but not C++ (and I think C++ doesn't do it because 'this' is a pointer rather than a ref) ).

Lastly, I've found some examples describing delegation in Ruby that show how using delegation enables better object-oriented programs. Such patterns aren't possible with implicit conversions alone. (But in C++ they might be accomplished with multiple inheritance).

mppf:
Delegation" as a term in the PL community refers to a variant of this mechanism that includes virtual dispatch - that is, the receiver of the delegated method can call back to one of the methods in the original object. That's not possible with this particular implementation. So, perhaps the keyword should be 'forward'.

I think it would be possible to add a language feature for generally forwarding one function to another as well. I think that would amount to a generalization of the feature.

References
"On the notion of inheritance", Antero Taivalsaari , CSUR 1996
https://en.wikipedia.org/wiki/Delegation_(object-oriented_programming)

mppf:
We probably shouldn't support coercion on method receivers that take in their argument by ref, but it's OK to support delegation in that context.

For this reason delegation can be used to wrap an array and still make the methods on it available.

mppf:
The forwarding of methods in the Array View implementation could be simplified with a strategy such as this one.

compiler/passes/normalize.cpp
```diff
@@ -779,7 +779,7 @@ static void insertRetMove(FnSymbol* fn, VarSymbol* retval, CallExpr* ret) {
                       new CallExpr(PRIM_COERCE, ret_expr,
                         fn->retExprType->body.tail->copy())));
   }
-  else if (fn->hasFlag(FLAG_IF_EXPR_FN))
+  else if (fn->hasFlag(FLAG_MAYBE_REF))
   {
```
vasslitvinov: Could you double-check that this change is intented?

compiler/include/stmt.h
```diff
@@ -307,6 +308,46 @@ class ExternBlockStmt : public Stmt {
...
+  virtual Expr*       getFirstChild();
+  virtual Expr*       getFirstExpr();
+
+  // used during parsing
+  DefExpr*            toFnDef;
+  // used before, during resolution
+  const char*         fnReturningForwarding;
+  // used during resolution
+  Type*               type;
+
```
vasslitvinov: Please add a comment for the 'type' field explaining what it is for.

compiler/AST/build.cpp
```diff
@@ -2525,6 +2525,111 @@ void applyPrivateToBlock(BlockStmt* block) {
   }
 }
 
+static
+DefExpr* buildForwardingExprFnDef(Expr* expr) {
+  // Instead of just storing expr directly, put it into a method
+  // and store a call to that method.
```
vasslitvinov: I think "store a call to that method" is incorrect. If so, please fix.

compiler/passes/normalize.cpp
mppf: Yes it is. Without this change, test/classes/ferguson/forwarding/d-record-to-record-nocopy.chpl does not pass, because the ForwardingStmt::toFnDef will always return a value. In that test, we're returning a field accessor expression, which is naturally a ref. We want it to stay a ref if it starts out a ref.

PR #5063, opened 2016-12-17
Refactor locale models

dmk42:
#4999 focused on fixing a race condition, and began refactoring the locale models only in the context of that fix.  This change finishes the refactoring to prepare for the new KNL locale model.

The layout within modules/internal is as follows.

localeModels/{flat,numa}/LocaleModel.chpl -- the main definitions of the locale models
LocaleModelHelp{Flat,NUMA}.chpl -- helpers and peripheral definitions
LocaleModelHelpSetup.chpl -- common definitions across locale models for setup
LocaleModelHelpMem.chpl -- common definitions across locale models for memory management
LocaleModelHelpRuntime.chpl -- common definitions across locale models for runtime interface

This has been tested on Mac, Linux, and Crays, using the flat and numa locale models, and various comm options.


modules/internal/LocaleModelHelpSetup.chpl
```diff
@@ -35,6 +35,11 @@ module LocaleModelHelpSetup {
 
   config param debugLocaleModel = false;
 
+  // We would really like a class-static storage class.(C++ nomenclature)
```
mppf: I thought that existed? Or do we just have 'type methods' but not 'type variables' ?

modules/internal/localeModels/flat/LocaleModel.chpl
```diff
@@ -230,7 +145,7 @@ module LocaleModel {
       return chpl_buildLocaleID(numLocales:chpl_nodeID_t, c_sublocid_none);
     }
     proc chpl_name() return local_name();
-    proc local_name() return "rootLocale";
+    proc local_name() return "rootLocale":string;
```
mppf: That's odd - is the `:string` really necessary here or a leftover from an earlier time?

dmk42: I was just making the two locale models identical.  I can delete the \:string from both.

modules/internal/LocaleModelHelpSetup.chpl
dmk42: I can probably delete that variable, because it is never set to anything but false.  I think I will leave that hook there in case it is needed later, though, and just delete the comment.  It works just as well as a static class member would, so the comment is not needed.

dmk42:
Before merging, I noticed that I had forgotten to break out the memory management functions into a separate file, to make it easier to implement them differently for KNL.  I created a new LocaleModelHelpMem.chpl for this purpose and retested everything.

@mppf : I'll wait until late Wednesday morning to merge, just in case you have any objections.

mppf:
@dmk42 - still looks good to me. I'd sanity test hello.chpl on flat and numa - if you havn't already - for the final version before merging.

dmk42:
Thanks.  I actually did a full retest on Mac, Linux, and Crays, on both locale models, so it looks like I'm good to go.  I'll go ahead and merge.

PR #5064, opened 2016-12-17
fasta: Changed from busy-wait to waitFor()

bradcray:
In conversations with the Bens, particularly BHarsh, I took an IOU to
see how much switching from a busy-wait to a more principled waitFor()
hurts performance.  It's reasonably minimal, so I made the switch.  In
addition to using supplied features and avoiding the taboo of
busy-waiting, this also reduces the need for the helper functions which
I eliminated.

While here, I renamed nextTask to nextTid to improve the symmetry
between the waitFor() and the write().

### DIFF
--- a/test/release/examples/benchmarks/shootout/fasta.chpl
+++ b/test/release/examples/benchmarks/shootout/fasta.chpl
@@ -121,7 +121,7 @@ proc randomMake(desc, nuclInfo, n) {
   // create tasks to pipeline the RNG, computation, and output
   coforall tid in 0..#numTasks {
     const chunkSize = lineLength*blockSize,
-          nextTask = (tid + 1) % numTasks;
+          nextTid = (tid + 1) % numTasks;
 
     var myBuff: [0..#(lineLength+1)*blockSize] int(8),
         myRands: [0..chunkSize] randType;
@@ -131,9 +131,9 @@ proc randomMake(desc, nuclInfo, n) {
       const bytes = min(chunkSize, n-i);
 
       // Get 'bytes' random numbers in a coordinated manner
-      wait(randGo);
+      randGo[tid].waitFor(i);
       getRands(bytes, myRands);
-      signal(randGo);
+      randGo[nextTid].write(i+chunkSize);
 
       // Compute 'bytes' nucleotides and store in 'myBuff'
       var col = 0,
@@ -162,18 +162,9 @@ proc randomMake(desc, nuclInfo, n) {
       }
 
       // Write the output in a coordinated manner
-      wait(outGo);
+      outGo[tid].waitFor(i);
       stdout.write(myBuff[0..#off]);
-      signal(outGo);
-
-      // Helper routines for taking turns with shared resources
-      inline proc wait(guard) {
-        while guard[tid].read() != i do ;
-      }
-
-      inline proc signal(guard) {
-        guard[nextTask].write(i+chunkSize);
-      }
+      outGo[nextTid].write(i+chunkSize);
     }
   }
 }
--- a/test/studies/shootout/fasta/bradc/fasta-blc-numa.chpl
+++ b/test/studies/shootout/fasta/bradc/fasta-blc-numa.chpl
@@ -158,7 +158,7 @@ proc randomMake(desc, nuclInfo, n) {
     if itid%div == 0 {
     const tid = itid / div;
     const chunkSize = lineLength*blockSize,
-          nextTask = (tid + 1) % numTasks;
+          nextTid = (tid + 1) % numTasks;
 
     var myBuff: [0..#(lineLength+1)*blockSize] int(8),
         myRands: [0..chunkSize] randType;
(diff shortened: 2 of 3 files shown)


bradcray:
@chapel-lang/perf-team : Note that I made the change from busy-waiting to waitFor() as it didn't hurt performance much, and it came up a lot in conversations.  I plan to merge this later today but am leaving it open for a bit in case anyone happens to be online and wants to take a peek.

bradcray:
Note: Tested 100 perf trials on shootout machine to look for problems and found none.  Tested under bencher script against 1.14 and everything seemed fine.

ben-albrecht:
👍 I like this a lot better.

ronawho:
yay!

You can make [numTasks](https://github.com/chapel-lang/chapel/blob/master/test/release/examples/benchmarks/shootout/fasta.chpl#L13) be `here.maxTaskPar` now, right?

Seems to improve timings slightly (timed with the release version): 
 - 2.2 ->1.9s on shootout box
 - 1.3 -> 1.2s on chap04
 - seems to stabilize chapcs numbers to 1.3s but I didn't do that many trials so no promises

bradcray:
If anything, I think we'd now want to make 'numTasks' simply be '3' -- that's the degree of parallelism that the computation supports and some parameter sweeps suggest that using more or less doesn't help and often hurts.  But I kept the here.numTasks value in because I think if you only had 1 or 2 cores (as on u32), you'd really only want 1 or 2 tasks to avoid oversubscription.

I didn't do any timings with the 1.14 release other than on sealcs01 which is a pretty noisy machine, and only saw things stay the same or get incrementally worse.  nightly graphs seem to be saying that as well (though I can't recall the magic incantation to bring up the submitted shootout graphs as timed against 1.14).

bradcray:
Oh wait, I'm realizing that I think I mis-parsed your comment.  You were saying changing to here.maxTaskPar had those effects?  I thought you were saying that my change had those effects.  Last I checked, increasing the number of tasks only hurt performance on the shootout box, but it's been awhile -- I'll double check.  It may be that '4' rather than here.maxTaskPar would be a sweet spot for chapcs in that it's an even number (would divide across numa evenly) and I can't think there's any advantage to adding more tasks (though I'll check that as well).

bradcray:
Here's what I'm seeing, sweeping the number of tasks from 1 to here.maxTaskPar:

* on shootout box, slight improvement for numTasks = 4 over 3, as you note (nice!  wasn't true earlier, but timings have changed a lot since then; makes sense to have retimed).  Varying blocksize across powers of 2, 1024 seems to be the sweet spot.

* on chapcs07, best timings are 1.5sec - 1.6 sec for numTasks = 4, 2, and 6 in roughly that order.  1 task is slower, as you'd expect.  For odd numbers of tasks, things are noisiest.  12 tasks vary from 2 to 10 seconds; 24 tasks vary from 5 to 7 secs

* on chap04, things continue to be more stable and I get best numbers at numTasks = 5, then 6, then 4, 7, 8 are all about the same (and all of these differences are probably in the noise).

Based on these observations, my thinking is to change the numTasks to min(4, here.maxTaskPar) since using more than 4 tasks doesn't seem to result in much of a positive improvement.

I wanted to correct one thing in my previous thinking:  I was thinking that because there are three stages in the pipeline that using more than three tasks was unlikely to ever be useful.  However, what I forgot to take into account is that only 2 of the stages require a lock and that the third, which doesn't, is the longest running.  That means that having more tasks could take up some of the slack in that task.  But it isn't longer than the sum of the other two by much, which is why I think we don't see great improvements after 4 tasks.

Anyway, curious for your thoughts.  I'd like to make one more change to this tonight and submit tomorrow.

bradcray:
Did a second run on chapcs07 because I didn't see anything like your 1.3s using 24 tasks and saw the same thing: best time at 4 tasks (1.1sec) and 24 tasks is in the 6.5-13.6 range (3 trials).  Can you reproduce that 1.3sec number?  Do you get better times at 4 tasks also?

ronawho:
I was on chapcs00 (chapcs07 was still running nightly testing when I was trying it out.) Let me rerun now

ronawho:
Still seeing 1.3 for chapcs

```
tail perfdat/chapcs07/fasta-release.dat
#
# master (numTasks=3)
12/18/16 	7.954
12/18/16 	10.21
12/18/16 	5.809

# numTasks=here.maxTaskPar
12/18/16 	1.305
12/18/16 	1.366
12/18/16 	1.314

```

ronawho:
Oh, just did 10 trials, and i'm starting to see the fluctuation. Guess I just got really "lucky" for 6 runs before:

```
chapcs07 (master) $ tail fasta-release.dat
# Date	real
12/18/16 	9.456
12/18/16 	1.377
12/18/16 	2.005
12/18/16 	1.387
12/18/16 	2.594
12/18/16 	11.62
```

bradcray:
Weird.  And that was on chapcs07?  I didn't get anything close to that.

Are you OK with the min(4, here.maxTaskPar) going onto master tonight?  (and likely into the site tomorrow barring surprises in the nightlies)?

ronawho:
yeah, no problem with `min(4, here.maxTaskPar)`

ronawho:
and yeah, I was using /tmp on chapcs07

bradcray:
Oh, that's interesting... I was thinking I was on a local share, but actually I was only on a geographically local share, not system-local.  I'm glad to hear you still saw variance in that mode, though.

PR #5065, opened 2016-12-17
Users guide: promotion

bradcray:
This is a first draft of a promotion section for the Chapel Users Guide with accompanying test from which code excerpts are pulled.

TODO:
- [X] write draft of section
- [x] learn how to run Paul's spell checker and use it on this section


cassella:
fortytwo@magrathea:~/src/chapel (users-guide-promotion)$ chplspell --report-only doc/users-guide test/release/examples/users-guide
    doc/users-guide/datapar/promotion.rst:281: 'evaluted' not found in dictionary (from token 'evaluted')

doc/users-guide/datapar/promotion.rst
```diff
@@ -0,0 +1,324 @@
...
+.. _ug-promotion:
+
+Promotion: implicit data parallelism
+====================================
+
+In Chapel, *function promotion* (or simply *promotion*) refers to an
+implicit form of data parallelism that is triggered by passing a
+collection of values to an argument that expects a single value.  More
+precisely, when an array of type *t* (or a range or domain whose index
+type is *t*) is passed to a function argument of type *t*, that
```
mppf: Is promotion of iterable-expressions (e.g. a leader/follower iterator) too low-level for this user's guide? I'd probably say "when an iterable expression yielding type *t* is passed to a function argument of type *t*, that function will be called in a data-paralle manner for all values yielded by the iterator."

One of the things that confused me the most when starting to learn Chapel is that a 'forall' expression (e.g. with the `[x in A]` syntax) actually creates an iterator rather than an array. While it's true that arrays are the common and easier-to-understand case here, it seems that being clear about how it works with iterators would be nice.

doc/users-guide/datapar/promotion.rst
```diff
@@ -0,0 +1,324 @@
...
+function which assigns its *y* argument into its *x* argument only if
+the third argument, *b*, is true:
+
+.. literalinclude:: examples/users-guide/datapar/promotion.chpl
+  :language: chapel
+  :lines: 51-54
+
+The following four calls demonstrate that various combinations of
+arguments may be promoted, where *Mask* is an array of ``bool``
+values:
```
mppf: Maybe it is worth showing an example definition for `Mask` ?

doc/users-guide/datapar/promotion.rst
```diff
@@ -0,0 +1,324 @@
...
+.. literalinclude:: examples/users-guide/datapar/promotion.chpl
+  :language: chapel
+  :lines: 63-65
+
+Then *A* will store the following values afterward:
+
+.. literalinclude:: examples/users-guide/datapar/promotion.good
+  :language: text
+  :lines: 23,25,27,29
+
```
mppf: I wonder if this section is trying too hard with the different variations of maybeCopy. Perhaps we only need one example of a non-iterable argument that is provided to each of the promoted calls?

doc/users-guide/datapar/promotion.rst
```diff
@@ -0,0 +1,324 @@
...
+as expected:
+
+.. literalinclude:: examples/users-guide/datapar/promotion.good
+  :language: text
+  :lines: 34
+
+Note that Chapel does not support an explicit ``+`` operator for
+arrays of reals.  Rather, Chapel implements this operation by
+promoting the standard ``+`` operator for scalar values of type
+``real``.  For this reason, applying ``*`` to Chapel arrays results in
```
mppf: I was initially confused by the switch from ``+`` to ``*``. Maybe instead of "For this reason" it should say "Similarly, ``*`` on arrays is implemented with promotion. For this reason, ..."

doc/users-guide/datapar/promotion.rst
```diff
@@ -0,0 +1,324 @@
...
+  :language: chapel
+  :lines: 123-124,130-131
+
+
+Promotion vs. Whole-Array Operations
+------------------------------------
+
+A key distinction should be made between Chapel's use of promotion
+versus whole-array operations as supported in traditional array
+languages.  To see the difference, consider the following statement:
```
mppf: When I initially read this paragraph, I thought the following would expect me to know something about "traditional array languages". It does not - in fact it explains that context very well. However there might be something that can be done here to assure the reader that this is still introductory material...

doc/users-guide/datapar/promotion.rst
bradcray: Oh, this was a misunderstanding on my part.  I knew that we didn't support promotion by arbitrary iterable expressions at present.  For example, if you define a class with all the normal these() iterators, we don't permit that class to promote functions... (note that I think this is what we ultimately want... it's just not what we did in the early days when we didn't yet have a parallel iterator story in place).

Instead, the current Chapel source code trigger for something to enable promotion is the presence of a member field called 'promotionType'.  This mechanism was not intended for end-users and is currently only used by ranges, domains, and arrays -- so I thought I was being precise in the above (even though that's not where we ultimately want to be).  What I failed to account for in my 'git grep' was that the compiler propagates this information for iteration expressions like the one you note... oops.

It's pretty fragile in when it does or does not do the propagation, though, so I don't know how broadly we should take the statement above at present.  I could try to be precise (which I think is "ranges, domains, arrays, or forall expressions over these types"), or I could stick with the present wording to be conservative and consider the lack of anonymous forall expressions a sin of omission, or I could say "all iterable expressions" and let someone call it a (big) bug when they run afoul of the current limitation (really more of an unimplemented feature that we've never formally discussed).

On the day that we do support promotion by iteration, I think the approach you suggest is definitely right.  I've been trying to write the users guide more to describe what we support than to what we hope to support in the future which is why I took this approach (and would be open to trying to describe what we have today more precisely if that's what you think is best... I'm just not entirely sure where the line between works and doesn't falls).

doc/users-guide/datapar/promotion.rst
bradcray: I believe I did this in part to set up for the following section which essentially demonstrates that if either of the latter two arguments are promoted, the first must be as well due to default forall intents.  Does your hesitancy extend even with that explanation?  If so, which variations would you keep and which would you drop?  (I'm about to push a commit that tries to simplify things slightly by moving the definitions of A, B, and Mask before the calls to use three code blocks rather than 4 which I think also helps a little).

doc/users-guide/datapar/promotion.rst
bradcray: Sounds good.

doc/users-guide/datapar/promotion.rst
bradcray: I like the idea, but used a different wording (which I'll commit in a moment)

bradcray:
Thanks for the feedback Michael!  See what you think of the proposed changes...

doc/users-guide/datapar/promotion.rst
mppf: I think it would be great to just say "ranges, domains, arrays, or forall expressions over these types". I think it would be good to include forall expressions in this document in some capacity too (I can't remember right now if you already did).

I also think we should make an issue about how more general promotion should probably be supported (I don't think I was aware of the missing functionality).

doc/users-guide/datapar/promotion.rst
mppf: Right now, it looks like there are 4 samples showing promoted calls to maybeCopy and the equivalent forall loop. I'd personally probably just make one. I think two would be OK. I think 4 is overkill. I think the first two examples would be sufficient. (I think *any two* of your examples would be sufficient, actually).

doc/users-guide/datapar/promotion.rst
mppf: Yes, but it would also be fine with me to not even bother to have 'Mask' as a way of reducing the number of examples here.

doc/users-guide/datapar/promotion.rst
bradcray: I added issue #5114 for this and filed some tests against.   In doing so, I learned that the fragility I referred to yesterday was a general fragility in resolving parallel loops and nothing specific to promotion.  It also turned out that all forall expressions promote, not just those over ranges, domains, and arrays (because I had a bug in my user-defined iterators, I was running into the general forall loop fragility and not anything specific to promotion).

I've also reworded (and reformatted) that sentence to try and list these three cases explicitly.

I agree that there should be a section on forall expressions in the users guide (along with all other language features).  I'm trying to prioritize based on importance / uniqueness to Chapel.  I've added an item to my TODO list to add forall expressions between the existing two data parallel sections.

mppf:
@bradcray the description of when promotion applies is much improved. I think this is ready to go in when you are ready to merge it.

bradcray:
@mppf: I reduced the number of partially promoted calls from 4 to 2 as you suggested and also added an example of promoting using a forall expression in the ranges/domains promotion section since we called that out as an additional non-array case.  Also did some inevitable wordsmithing just due to looking at it again with fresh eyes.  Let me know whether you want to review the latest commit or are ready to be done with it.

mppf:
I'm ready to be done with it - so as far as I'm concerned, merge away.

PR #5069, opened 2016-12-17
Support params without initializers

bradcray:
While it's not very interesting or useful, it seems that we should
permit 'param' declarations without initialization expressions for
orthogonality (where the 'param's value would be its default value).
It turns out that it was a five-minute fix to enable this and that
only one test seemed to rely on the old "error message" behavior,
so I made the fix and propose we accept it.  While normal params
without initializers are not that interesting (because they can only
hold their type's default values), in the 'config param' case, it's
slightly more useful/interesting (though arguably, supplying an
initializer might still be considered better form).

I checked the spec to see if it needed updating if we accept this
change, but didn't see that params said anything about requiring
initializers. However, consts did say they required initializers even
though the compiler permits them to be elided (as I believe it should),
so I updated that wording since it is thematically related to this PR.

TODO
- [X] implement a nice 5-minute fix (TM)
- [x] run testing
- [x] make sure spec still builds
- [x] read spec pdf to make sure it rendered OK

### DIFF
--- a/compiler/passes/checkParsed.cpp
+++ b/compiler/passes/checkParsed.cpp
@@ -161,12 +161,6 @@ static void checkPrivateDecls(DefExpr* def) {
 
 static void
 checkParsedVar(VarSymbol* var) {
-  if (var->isParameter() && !var->immediate)
-    if (!var->defPoint->init &&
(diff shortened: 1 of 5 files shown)


lydia-duncan:
This looks good to me!

PR #5070, opened 2016-12-17
Enable multiple bad param errors

bradcray:
This change falls in the category of "making the compiler better at
reporting multiple errors."  Previously, when we hit a case where a
'param' was initialized with a non-param expression, we generated
one error and exited.  Here, I switch a USR_FATAL() into a 
USR_FATAL_CONT() in combination with marking the earlier param
symbol as not being a param (since, in reality, it isn't) as a means to
generating multiple errors in such cases.  Failure to de-mark the
symbol it like this resulted in other errors that added confusion
because we'd end up calling 'param' versions of functions
with seeming-param that the compiler couldn't interpret.  While this
removal of the flag is a little destructive in nature (throwing away
information), since we're about to error-out anyway, it seems justifiable
for the sake of giving users more error messages (which is a
common complaint against the Chapel compiler).

TODO:
- [X] implement a nice 5-minute fix (TM)
- [x] testing

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -7309,8 +7309,11 @@ postFold(Expr* expr) {
               }
             }
           }
-          if (!set && lhs->symbol()->isParameter())
-            USR_FATAL(call, "Initializing parameter '%s' to value not known at compile time", lhs->symbol()->name);
+          if (!set && lhs->symbol()->isParameter() && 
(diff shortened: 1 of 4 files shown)


lydia-duncan:
Minor nit: your PR message calls it a INT_FATAL_CONT, when it is a USR_FATAL_CONT

lydia-duncan:
Other than that, this looks good!

PR #5071, opened 2016-12-17
Move adjustArgSymbolTypesForIntent function

benharsh:
I believe this was placed in IWR to make progress during the initial QualifiedType effort. Now I think it can safely be moved into codegen.

Testing:
- [x] full local, no-local
- [x] full gasnet, dist-cyclic, dist-replicated
- [x] release examples llvm, llvm-no-local


benharsh:
Yup, it just moves the code. Nothing specific suggested that it could be moved, but I thought I'd try since it seems unnecessary for IWR.

PR #5072, opened 2016-12-17
Flag illegal param types

bradcray:
While writing up my users guide section on params, I realized that we don't really do any proactive errors to indicate that a param is not of a legal type, which can be confusing to users.  Instead, we just say that its initializer expression is not a param, which is vague.  This PR attempts to tighten that up by explicitly detecting whether the type is a legal param type or nor (bool, int, uint, real, imag, enum, or string).  I also let dtUnknown go through because the user can't declare variables of this type, and otherwise we get errors (though perhaps that can be improved in the future).

While working on this I (re-)remembered that we don't support 32-bit float params because our defaultOf() function isn't a param because we don't have any good way to create a 32-bit floating point value at present.  Rather than calling this case illegal in my new check, I decided to let it fall through and filed a future against the 32-bit float case instead.

TODO:
- [X] write a five-minute fix to try and make things better
- [X] do initial testing with manually merged #5069, which this PR depends upon
- [x] merge PR #5069 to master
- [x] do full testing
- [x] see if any existing futures exhibit this same 32-bit float issue and unify with them
- [x] file an issue against this new future and any found in the previous step
- [x] see if this breaks any futures


mppf:
Looks good to me, assuming the next steps go OK. One note: I tend to prefer FnSymbol::getReturnSymbol to checking for FLAG_RVV, but that is a minor thing.

bradcray:
@mppf: To make sure I'm understanding, are you suggesting replacing "lhsSym->hasFlag(FLAG_RVV)" with "lhsSym == fn->getReturnSymbol()"?  Thanks for clarifying/verifying.

mppf:
that was what I was suggesting. I don't know why it didn't work...

bradcray:
It turned out that I needed to ensure that 'fn' was defined / non-NULL.  I tried doing this by setting it to fn->callResolved(), but didn't get as accurate error messages anymore (e.g., for test/users/weili/metaprog2.chpl).  Since you considered this a minor thing and I don't really understand the relationship between getReturnSymbol() and FLAG_RVV well enough, I'd like to merge it as-is for the sake of expediency unless you object.

mppf:
oh yeah, that's fine with me.

PR #5073, opened 2016-12-18
Add param versions of min max for ints

bradcray:
It seemed like an oversight and a non-orthogonality with other ChapelBase
operators that we don't support min() and max() on param ints.  This adds
them along with a test that exercises them.

We might want to think about supporting param min max for other types as well,
but I was nervous about doing it for floating point variables given that we typically
haven't supported operators on floating point params and supporting it on bools
and enums seemed a little "out there" without further conversation or thought.

TODO:
- [X] complete 5-minute fix (TM)
- [x] run full testing


mppf:
Looks good to me.

ben-albrecht:
Looks good to me.

bradcray:
Oops, I mis-credited the review on the merge message... should've been @mppf and ben... Sorry Michael.

PR #5074, opened 2016-12-18
First draft of a users guide section on constants (consts and params)

bradcray:
This is my first draft at a users guide section on consts and params.
It includes a test that contains code excerpted into the users guide.
It's dependent on several other PRs that I've opened but not yet merged:
#5073, #5072, #5070, #5069 and therefore should not be merged until
those are.  If those PRs don't get merged, it just means that the test
and .good file likely need edits (and perhaps the supporting text would
as well...

TODO:
- [x] write section
- [x] merge the depenent PRs mentioned above
- [x] learn how to use Paul's spell checker and run it

PR #3406, opened 2016-03-04
Fix type alias for generic classes

mppf:
A future work item from the addition of PCG to modules/standard/Random.chpl is to adjust RandomStream to be NBPRandomStream and then (in some future version) become PCGRandomStream. The obvious way to do that is with a type alias e.g.:

```
type RandomStream = PCGRandomStream;
```

However, since RandomStream is a generic class, this did not work as expected.

This patch adjusts a few lines in normalize, resolution, and code generation, in order to allow such type aliases to work. It removes the chpl__typeAliasInit call which was previously added in normalize, in favor of expecting function resolution to detect the error when a type is initialized with a value.

For history, the chpl__typeAliasInit call was added in 68d1ad7 with the explanation "main reason for creating and resolving this expression is to issue an error if an attempt is made to initialize a type variable using a value that known only at run-time."

Adds several tests that show the class type alias working in the case needed for Random.chpl.

Passed full local testing.
Passed hellos with --baseline and/or --verify.

Reviewed by @lydia-duncan - thanks!

PR #5075, opened 2016-12-18
Users guide: type aliases

bradcray:
Add a first draft of a users guide section on type aliases and a
supporting test.

TODO:
- [x] write section
- [x] commit consts section in PR #5074 which this one cross-references
- [x] learn how to use Paul's spell checker and run it


cassella:
fortytwo@magrathea:~/src/chapel (users-guide-types)$ chplspell --report-only doc/users-guide test/release/examples/users-guide
    test/release/examples/users-guide/base/typeAliases.chpl:24: 'bigvec' not found in dictionary (from token 'bigvec')
    test/release/examples/users-guide/base/typeAliases.chpl:27: 'bigvec' not found in dictionary (from token 'bigvec')

Since this is a deliberate new word, you can add it to the dictionary
interactively.  For example, to add it to a new file-specific
dictionary for this file, at the prompts eter "a" and "N" to `(a)dd to
dictionary` then `(N)ew file-specific dictionary`:

    fortytwo@magrathea:~/src/chapel (users-guide-types)$ chplspell doc/users-guide test/release/examples/users-guide
    test/release/examples/users-guide/base/typeAliases.chpl:24: Unmatched 'bigvec' --> {bigvec}
       (i)gnore, (I)gnore all, (r)eplace, (R)eplace all, (a)dd to dictionary, or
       show (c)ontext? [i]
          Subtoken 'bigvec':
             (b)ack, (i)gnore, add to (n)atural language dictionary, add to
             (p)rogramming language dictionary, add to
             (N)ew file-specific dictionary [i]
    New file ID c265c700-c675-11e6-ae77-10ddb1d4c3d5 for test/release/examples/users-guide/base/typeAliases.chpl

Then `git add .scspell` and commit the dictionary update.

doc/users-guide/base/typeAliases.rst
```diff
@@ -0,0 +1,67 @@
...
+  ``uint(8)`` type would.
+
+* **minimizing the number of code changes required to represent a
+  certain class of values:** If we decided that we wanted to represent
+  other plants and animals in addition to people, we'd probably want
+  to redefine *age* to be a larger integer size in order to store the
+  ages of long-lived organisms like giant redwoods.  It's easier to do
+  this by changing a single type alias's definition than by finding
+  all references to ``uint(8)`` and determining whether or not they
+  represent ages.
```
mppf: thanks for mentioning giant redwoods :)

doc/users-guide/base/typeAliases.rst
```diff
@@ -0,0 +1,67 @@
...
+
+.. _ug-typeAliases:
+
+Type Aliases
+============
+
+In addition to :ref:`declaring variables <ug-vars>` using ``var`` and
+:ref:`constants <ug-constParam>` using ``const`` and ``param``, Chapel
+supports the declaration of *type aliases* using the ``type`` keyword.
+As the term suggests, type aliases create symbolic names for types.
```
mppf: I think it would be good to go further to clarify  "Is it another name for the same type, or a new type?" - e.g. by adding "The type alias only creates a new name for an existing type. It does not create a new type".

Additionally, it's not exactly the job of this PR, but the language spec currently describes that as an *open issue*. I think it needs to be another name for an existing type (rather than creating a new type) but I'm not sure everyone will think the issue is settled without discussion. But, I'd like to see that settled in the spec, too.

Also see:
 * PR #5057, specifically the "Future Work" section of the PR message
 * PR #3406, which enables type aliases for generic types

mppf:
I've looked over all of this (sans spelling and rendering) and it looks reasonable to me.

doc/users-guide/base/typeAliases.rst
bradcray: This sounds like a good clarification to make.  

FWIW, my current thinking on the topic is that while I'd ultimately like a means of creating new types, at this point, we're far enough into things that if we were to do that, we should probably use a different keyword than simply 'type'.  E.g.,

    typedef age: int;

as a means of saying "define a new type named 'age' that is a subtype of 'int'".

Unlike 'type', I imagine that 'typedef' would not be a legal argument intent.  By extension if a class/record contained a 'typedef' declaration, I wouldn't expect any compiler-generated initializers to support an override via argument (i.e., the presence of the typedef wouldn't make the class/record generic... only 'type's would do that).

mppf: I'm not so sure I like the exact 'typedef' proposal, but I *do* think we'd want different syntax for getting a "new type" and that it makes sense for the current usage to be an "alias type". We should probably move this discussion to some issue, but my counter-proposal would be:
```
 new type age = int;
```
or possibly
```
type age = new type int;
```

I like your reasoning about typedef/whatever not applying to argument intent/generic fields.

bradcray: I worry that these counterproposals overload the 'type' keyword too much, but I think you're right that we should move this discussion to its own issue rather than belaboring it here.  That said, I can't resist saying that part of the reason I like using the ':' for such cases is that it suggests a "subtype" relationship while also providing a nice way to define default values for new types, if we wanted to support it:

    typedef age: int = 21;

PR #5079, opened 2016-12-19
Users guide: configs section

bradcray:
This is a configs section for the users guide and supporting tests.  One thing that was a little unfortunate about these tests is that I wanted to show full command lines in the documentation and not have test-system-specific comments (like '# compareToThis.good') appear.  For this reason, I created .execopts-fake and .compopts-fake files that nearly mirror the .exceopts and .compopts files modulo those changes.  To ensure that these don't get out of sync with one another, it would be nice to add meta-tests to the test system that would strip out the extra characters and diff them against one another, but I didn't have the scripting skills to get that done in the time that I had on this pass.  My current thinking is to see how hard the maintenance of these files is over time and add such checks if/when it seems like it's a problem.

TODO:
- [x] write section
- [x] learn how to use Paul's spell checker and run it


cassella:
fortytwo@magrathea:~/src/chapel (users-guide-configs)$ chplspell --report-only doc/users-guide test/release/examples/users-guide
    doc/users-guide/base/configs.rst:179: 'followng' not found in dictionary (from token 'followng')

test/release/examples/users-guide/base/configs3-setops.good
```diff
@@ -0,0 +1,2 @@
+10
+1e-09
```
cassella: I don't think this file is referenced anywhere.  Is it just an extra copy of configs3-seteps.good?

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+then *epsilon* will be computed in terms of it, resulting in the value:
+
+.. literalinclude:: examples/users-guide/base/configs3-n1000.good
+  :language: text
+  :lines: 2
+
+Note that the compiler's command-line override can still be
+re-overridden on the executable's command-line as usual:
+
+.. literalinclude:: examples/users-guide/base/configs2.compopts-fake
```
cassella: The preceding text suggests we're going to see an example of overriding config vars on the executable's command line  But it includes only a line from a compopts file, showing an invocation of `chpl`.

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+which their symbols' initial values as specified in the source code
+can be overridden on the command-line.  This supports the ability to
+specify and override default values, and is achieved by prefixing the
+statement with the keyword ``config``.
+
+``config var`` and ``config const``
+-----------------------------------
+
+Configuration variables are variables whose default values can be
+overridden the executable's command-line.  Here's a simple example:
```
cassella: "can be overridden on"?

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+initializers for *bitMask*, *size*, and *age*:
+
+.. literalinclude:: examples/users-guide/base/configs2.compopts-fake
+  :language: text
+  :lines: 3
+
+One of the nice things about setting ``config`` declarations in the
+compiler is that the specified value can take the form of reasonably
+arbitrary Chapel code since the compiler is involved in interpreting
+it.  Thus, we can use more interesting expressions when setting
```
cassella: Do you want to mention that the expression for a ``config param`` has to be a param expression?  The examples in this section are two ``config param``s and a ``config type``.

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
+:title: Users Guide: Configuration Declarations
+
+.. _ug-typeAliases:
+
+**config** Declarations: command-line overrides
+===============================================
+
+The four main declaration keywords that have been introduced so far —
+``var``, ``const``, ``param``, and ``type`` — each support a mode in
+which their symbols' initial values as specified in the source code
```
vasslitvinov: I suggest dropping "as" - it will be simpler that way.

vasslitvinov: How about symbol's instead of symbols' ?

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+can be overridden on the command-line.  This supports the ability to
+specify and override default values, and is achieved by prefixing the
+statement with the keyword ``config``.
```
vasslitvinov: "declaration" instead of "statement" ?

test/release/examples/users-guide/base/configs.execopts
```diff
@@ -0,0 +1,4 @@
+./a.out --n 1000
+./a.out --n=1000
+./a.out --epsilon=0.000001 --debug=true  # configs-eps+debug.good
+./a.out --epsilon=0.000001 --debug       # configs-eps+debug.good
```
vasslitvinov: Hm does it work to have a.out in an .execopts file?
If it does, then OK.

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+  :lines: 1
+
+As with the generated executable, boolean values can also be set on
+the compiler's command line simply using the name of the flag.  Thus,
+the following would be equivalent to the previous line:
+
+.. literalinclude:: examples/users-guide/base/configs2.compopts-fake
+  :language: text
+  :lines: 2
+
```
vasslitvinov: Would be nice to show also that a space after "-s" is allowed, e.g. "chpl -s size=8" .

test/release/examples/users-guide/base/configs2.compopts-fake
```diff
@@ -0,0 +1,8 @@
+chpl -sverbose=true configs2.chpl
+chpl -sverbose configs2.chpl
+chpl -sbitMask=0b00001111 -ssize=16 -sage=uint(16) configs2.chpl
+chpl -sage=real(32) configs2.chpl
```
vasslitvinov: I think we should also use quotes around command-line flags that include parens, e.g. -sage=uint(16)

Otherwise some readers (like myself) will ponder how this is supposed to work with bash/tcsh/...

test/release/examples/users-guide/base/configs2.compopts-fake
```diff
@@ -0,0 +1,8 @@
...
+chpl -sbitMask='1 << 10' -ssize=numBits(bitMask.type)/2 configs2.chpl
```
vasslitvinov: Quotes around parens here, too.

doc/users-guide/base/configs.rst
vasslitvinov: Right. Maybe say that "the compiler must be able to compute these expressions at compile time".

vasslitvinov: While "reasonably arbitrary" sounds cool, I would prefer to see something that's reasonably readable, e.g.

"the specified value can be an arbitrary Chapel expression, provided the compiler can evaluate it during compilation"

With this, "since the compiler is involved in interpreting it" is unnecessary and can be removed.

doc/users-guide/base/configs.rst
```diff
@@ -0,0 +1,220 @@
...
+and types, but with complex expressions:
+
+.. literalinclude:: examples/users-guide/base/configs2.compopts-fake
+  :language: text
+  :lines: 5-6
+
+Note the use of quotes in this example to preserve whitespace in the
+initial value.  This relates to having the command shell pass the
+entire expression to Chapel as a single argument rather than being
+a Chapel-specific syntax.
```
vasslitvinov: Add: "Likewise, parentheses, asterisks, etc. need to be quoted as well to escape their special meaning to the command shell."

vasslitvinov:
Other than the comments from Paul and myself, this is good to do.

For a bonus point, could say that strings must be double-quoted at chpl command line, for bash/tcsh like so:

  chpl -smyStringConfig='"init override"'

and not on executable line:

  a.out --myStringConfig=simpleString

Likewise enum constants must be preceeded by the enum name when compiling, not necessarily when running the compiled executable, e.g.:

  chpl -s distType=DistType.block
  a.out --distType=DistType.block
  a.out --distType=block

test/release/examples/users-guide/base/configs3-setops.good
bradcray: Whoops, sure looks like it... thanks!

doc/users-guide/base/configs.rst
bradcray: The three comments on this paragraph are good.  I incorporated them and edited further, ending up with:

    The four main declaration keywords that have been introduced so far —
    ``var``, ``const``, ``param``, and ``type`` — support a mode in which
    their initializing expressions in the source code can be overridden on
    the command-line.  This	is achieved by prefixing the declaration with
    the keyword ``config``.  Chapel's ``config`` feature provides a simple
    means of specifying and	overriding default values.

Thoughts?

doc/users-guide/base/configs.rst
bradcray: I think I must've been tired by the time I got to this section...

doc/users-guide/base/configs.rst
vasslitvinov: Sounds good!

doc/users-guide/base/configs.rst
bradcray: The comments on this paragraph are good.  I went with the following, if you guys want to give it another look:

    One of the nice things about setting ``config`` declarations on the
    compiler's command-line is that the specified value can be an
    arbitrary Chapel expression, provided that the compiler can evaluate
    it in place of the source-based initialization	expression.

Since we go on to say that config consts and vars can be set on the compiler line as well, I tried to stay away from anything as specific as saying that the compiler had to be able to evaluate the expressions and instead focused on the expression needing to make sense in the context of the initializer (which arguably subsumes the compile-time condition for params and types while also saying something stronger about what is necessary for the expression to be legal).

Does this work for you guys?

doc/users-guide/base/configs.rst
bradcray: Good catch, thanks!

test/release/examples/users-guide/base/configs.execopts
bradcray: I cheated a little in the file by having main() take additional arguments and asserting that the argument was ./a.out... :)

But then I had to create the -fake version of the file anyway because of the need for the '# overrides for lines 3 and 4... :'(

An open question is how I make these -fake files maintainable over time...  sigh...

test/release/examples/users-guide/base/configs2.compopts-fake
bradcray: Great point.  It hadn't occurred to me that the use of the .execopts file was protecting me from issues like this...

doc/users-guide/base/configs.rst
vasslitvinov: Sure, sounds good.

test/release/examples/users-guide/base/configs.execopts
vasslitvinov: All that's needed is for literalinclude to strip the # comments when reading .compopts/.execopts. Right?

bradcray: Yeah, but we don't control literalinclude -- that's part of Sphinx's functionality (and I don't think we want to modify Sphinx...  And I'd still need to have -fake files for the compopts files...

I think the "right" thing to do is to have the test system do some stripping of prefixes/suffixes and diff that the fake and real files match... but I didn't have the scripting skills to do that in the time I've had so far...

bradcray:
@vasslitvinov : Your bonus ideas are good, but I'm going to pass on them for now.  I put in some text about how "different shells will need things guarded in different ways" which I hope is enough of a cover for the first issue... I don't think of the users guide as needing to be exhaustive as long as it allows users to draw the right conclusions.  We can add additional special cases as users express confusion is my current thinking...

The points about enums are good ones... I'm going to put them off (and am noting them in the TODO file) until I actually introduce enums, and will probably cover them there rather than here.  Also, note that the compile-time -s doesn't *necessarily* require the enum to be named, if the scope has a 'use  DistType' already...  I.e., I think the previous "if the expression can appear in the source here, it's OK" guards this case.  So really it's primarily the ability to leave the enum off on the executable line that's probably worth calling out as a special case (?).

PR #5080, opened 2016-12-19
Fix typos in ssca2

cassella:
Fix typos in test/releas/examples/benchmarks/ssca2.  While I'm at it,
spot-fix those same typos where they appear in test/studies/ssca2,
without attempting a full spellchecking of those versions.


cassella:
The only non-comment change is weightd->weight.  That string does not appear in any .good or .bad files.

testing: start_test test/release/examples/benchmarks/ssca2 passes

I didn't test the other variants.

psahabu:
Looks good to me.

PR #5084, opened 2016-12-19
compiler pass: lower error handling 

psahabu:
Note: These are the first steps toward a fully functioning pass that lowers error handling AST. Strict mode, `catch`, and the `Error` type are completely non-functional here.

- creates the pass `lowerErrorHandling`, just before `callDestructors`
- creates checks `check_lowerErrorHandling`, `check_afterLowerErrorHandling`
- draft implementation of `throws`, `try!`
- passes linux64+flat testing

@ronawho: This is the PR that adds a compiler pass.

### DIFF
--- a/compiler/include/checks.h
+++ b/compiler/include/checks.h
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2016 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
  * The entirety of this work is licensed under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License.
- * 
+ *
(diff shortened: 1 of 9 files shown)


mppf:
@psahabu - I would expect to see some tests that change behavior based on this work. Is there some reason we can't add a test exercising the new code? Also, you should run a -futures test over test/errhandling since this will almost certainly change a bunch of .bad files for your own error handling tests.

ronawho:
@psahabu just ping me whenever this is committed. If it goes in today, sometime before 10 EST would be preferable

psahabu:
@mppf and I have created a dramatically different version of this pass, and so believe that this PR should be closed. I'll create a separate PR that adds the pass without any functionality.

PR #5085, opened 2016-12-19
Perf graph filter

benharsh:
Add a text box to the perf graphs page that, when text is entered, will only show graphs whose titles (in lowercase) match the text box contents.

If a user filters graphs and hits "select all", only visible graphs will have their checkboxes ticked.

I currently disable the filter box for typing while json requests are in flight because I saw some odd behavior otherwise.

### DIFF
--- a/util/test/perf/index.html
+++ b/util/test/perf/index.html
@@ -48,6 +48,7 @@
             value="Select All">
             <input type="button" onclick="unselectAllGraphs()"
             value="Unselect All">
+            <p>Filter: <input type="text" name="filterBox"></p>
             <div id="graphlist" class="graphList"></div>
           <td valign="top">
             <div id="graphdisplay" class=graphdisplay></div>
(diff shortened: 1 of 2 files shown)


benharsh:
@chapel-lang/perf-team : Here's this feature I've talked about for a while now. I can provide you Crayons with an internal demo link to try it out.

bradcray:
Ooooooh, nice...!

Vaguely related: what's the level of effort required in adding an "invert selection" button next to the select/deselect all?

benharsh:
That should be fairly straightforward. I can do that in a separate PR.

lydia-duncan:
I'm not sure I like that it adjusts the displayed graphs based on the filter.  Personally, I would prefer that to only impact the selection boxes.  I can see it being confusing/frustrating to have selected a couple of graphs, filter to select the last graph you want, and then hit display and only have the last graph turn up, even if the rest turn up as you delete from the filter box.

I do like that select all and unselect all only affect the currently filtered graphs

ronawho:
I love that the filter adjusts the displayed graphs, don't listen to lydia!

PR #5086, opened 2016-12-19
Annotate last night's max tasks change for fasta

bradcray:
Also fix a case where I forgot to change a 3 to a 4 in the numa
study version as pointed out by @cassella.  Named it symbolically
to avoid such issues in the future.

THinking about the numa version in the waitFor() world, I'm also
realizing that we could now simplify a bunch of the complicated
logic I introduced to avoid starving tasks when we were using
busy waits.  Good post-vacation activity.


bradcray:
@cassella: Want to give this a quick sanity check for me since you caught the original error?

cassella:
The change makes sense to me.

PR #5090, opened 2016-12-20
strengthen verification of Symbols

vasslitvinov:
* Ensure that the only symbol that can have null defPoint is rootModule.

* Ensure that the symbol referenced by a live SymExpr is itself live aka "in AST".

While there, removed special treatment of symbols with FLAG_GLOBAL_TYPE_SYMBOL
w.r.t. isAlive() check.

Why is it OK to remove the FLAG_GLOBAL_TYPE_SYMBOL check?
Suppose that it matters, i.e. there is a symbol S for which
isAlive returned true before this change and now returns false.
In that case, S will not undergo verification, which is an omission.
However, S will now be cleaned up i.e. deleted at the end of
the pass. If S were actually needed, then its deletion will cause
memory corruption, very likely. That should be detected by valgrind
testing and/or cause downstream failures.

(This was motivated by the verification check not catching a bug
I'd introduced in my local tree, which resulted in accessing
clobbered memory.)

Testing:
+ linux64 --verify on full suite
+ gasnet --verify on hello4
+ gasnet --verify on multilocale tests, for an earlier version

### DIFF
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -510,8 +510,13 @@ void SymExpr::verify() {
   if (var == NULL)
     INT_FATAL(this, "SymExpr::verify %12d: var is NULL", id);
 
-  if (var != NULL && var->defPoint != NULL && var->defPoint->parentSymbol == NULL)
-    INT_FATAL(this, "SymExpr::verify %12d:  var->defPoint is not in AST", id);
+  if (var->defPoint) {
+    if (var->defPoint->parentSymbol == NULL)
+      INT_FATAL(this, "SymExpr::verify %12d:  var->defPoint is not in AST", id);
+  } else {
+    if (var != rootModule)
+      INT_FATAL(this, "SymExpr::var is a symbol without a defPoint");
+  }
 
   /* Check that:
       - every live SymExpr is in a Symbol's list
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -144,10 +144,15 @@ static inline void verifyInTree(BaseAST* ast, const char* msg) {
 }
 
 void Symbol::verify() {
-  if (defPoint && !defPoint->parentSymbol && !toModuleSymbol(this))
-    INT_FATAL(this, "Symbol::defPoint is not in AST");
-  if (defPoint && this != defPoint->sym)
-    INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym");
+  if (defPoint) {
+    if (!defPoint->parentSymbol && this != rootModule)
+      INT_FATAL(this, "Symbol::defPoint is not in AST");
+    if (this != defPoint->sym)
+      INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym");
+  } else {
+    if (this != rootModule)
+      INT_FATAL(this, "Symbol without a defPoint");
+  }
   verifyInTree(type, "Symbol::type");
 
   if (symExprsHead) {
--- a/compiler/include/expr.h
+++ b/compiler/include/expr.h
@@ -381,13 +381,11 @@ static inline bool isAliveQuick(Symbol* symbol) {
 }
 
 static inline bool isAlive(Symbol* symbol) {
-  if (symbol->hasFlag(FLAG_GLOBAL_TYPE_SYMBOL)) return true;
-  if (! symbol->defPoint) return false;
-  return isAliveQuick(symbol);
+  return symbol->defPoint && isAlive(symbol->defPoint);
 }
 
 static inline bool isAlive(Type* type) {
-  return isAliveQuick(type->symbol);
+  return isAlive(type->symbol->defPoint);
 }
 
 #define isRootModule(ast)  \


compiler/AST/symbol.cpp
```diff
@@ -144,10 +144,15 @@ static inline void verifyInTree(BaseAST* ast, const char* msg) {
...
-    INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym");
+  if (defPoint) {
+    if (!defPoint->parentSymbol && this != rootModule)
+      INT_FATAL(this, "Symbol::defPoint is not in AST");
+    if (this != defPoint->sym)
+      INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym");
+  } else {
+    if (this != rootModule)
+      INT_FATAL(this, "Symbol without a defPoint");
+  }
```
mppf: I think this section might be clearer if you separated out the this != rootModule check.

```
if ( this == rootModule ) {
  if (defPoint != NULL && this != defPoint->sym)
    INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym");
} else {
  if (defPoint == NULL)
    INT_FATAL(this, "Symbol without a defPoint");
  if (defPoint->parentSymbol == NULL)
    INT_FATAL(this, "Symbol::defPoint is not in AST");
  if (this != defPoint->sym)
    INT_FATAL(this, "Symbol::defPoint != Sym::defPoint->sym"); 
}
```

But even then I am a bit confused - does rootModule have DefPoint or not? (I would expect it not to have a defPoint). In that case, you could simply skip all of this error checking if this == rootModule.

compiler/include/expr.h
```diff
@@ -381,13 +381,11 @@ static inline bool isAliveQuick(Symbol* symbol) {
 }
 
 static inline bool isAlive(Symbol* symbol) {
-  if (symbol->hasFlag(FLAG_GLOBAL_TYPE_SYMBOL)) return true;
-  if (! symbol->defPoint) return false;
-  return isAliveQuick(symbol);
+  return symbol->defPoint && isAlive(symbol->defPoint);
```
mppf: Do you know why the case for FLAG_GLOBAL_TYPE_SYMBOL was added? Do you know that we can remove it? Certainly good testing results is one indication.
Also, can we now remove isAliveQuick ? It does not seem to be any "quicker" than isAlive at this point.

mppf:
@vasslitvinov - generally, looks good. I think the structure of the conditionals could be improved. I'm happy you were able to add the check & that all the testing passed. You could take my advice or merge it the way it is. But it would be good to at least answer the questions I asked in the PR itself.

compiler/AST/expr.cpp
```diff
@@ -510,8 +510,13 @@ void SymExpr::verify() {
...
 
-  if (var != NULL && var->defPoint != NULL && var->defPoint->parentSymbol == NULL)
-    INT_FATAL(this, "SymExpr::verify %12d:  var->defPoint is not in AST", id);
+  if (var->defPoint) {
+    if (var->defPoint->parentSymbol == NULL)
+      INT_FATAL(this, "SymExpr::verify %12d:  var->defPoint is not in AST", id);
+  } else {
+    if (var != rootModule)
+      INT_FATAL(this, "SymExpr::var is a symbol without a defPoint");
+  }
```
vasslitvinov: See my response below.

compiler/AST/symbol.cpp
vasslitvinov: I agree this version is clearer. However it adds a branch + memory access to execute "this == rootModule" for each symbol, whereas my version executes it only for the rootModule symbol.

You are correct, rootModule->defPoint is NULL. So indeed there is no checking to be performed in that case... although I would throw in the check that it is actually NULL - to catch an unlikely bug when it is not.

compiler/include/expr.h
vasslitvinov: The case for FLAG_GLOBAL_TYPE_SYMBOL was added in cd44f189 without an explanation in the commit message. 
I will extend the PR message with an explanation of why I think it is OK to remove it.

isAliveQuick contains one extra branch compared to isAlive.

vasslitvinov:
@mppf - I am going ahead with this PR as-is. However, if you think that saving the extra branches, as per my explanations, are not worth the added code complexity or decreased clarity, I will put together a separate PR. Although w.r.t. clarity one can argue either way...

mppf:
@vasslitvinov - I do not think that saving branches/memory access in the compiler is a worthwhile goal generally. I would feel differently if we were talking about a region of code we knew was responsible for most of the compile time. In this case, we are talking about code that only runs on --verify and in that context I would definitely value simplicity/clarity over speed. Nonetheless I'm OK with this as-is.

noakesmichael:
I am 100% with Michael on this theme.  Unless one is discussing code that
one can prove has a non-trivial impact on application-level performance, one
should prefer simplicity/clarity.

Clarity is a subjective matter, to some degree, but it should not be disregarded
based on an argument about number of branches/memory-ops in the C source
code.



On Dec 21, 2016, at 5:24 AM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@vasslitvinov<https://github.com/vasslitvinov> - I do not think that saving branches/memory access in the compiler is a worthwhile goal generally. I would feel differently if we were talking about a region of code we knew was responsible for most of the compile time. In this case, we are talking about code that only runs on --verify and in that context I would definitely value simplicity/clarity over speed. Nonetheless I'm OK with this as-is.

PR #5094, opened 2016-12-21
fix bugs in CSRDom.dsiPartialThese

vasslitvinov:
Since our compiler silently reverts to using leader-follower for
a forall loop when the standalone iterator does not compile,
these bugs originally went unnoticed. They were exposed and fixed
by commenting out the leader-follower iterator.

Contributed by Engin.


bradcray:
:'(  

(and further evidence that we should completely retire the 'try token' mechanism in favor of the Reflection module?)

PR #5104, opened 2016-12-23
Fix sloppiness in bulk transfer.

gbtitus:
The transfer code was making assumptions that weren't warranted, given
what doiUseBulkTransfer() said could be handled.  In the particular case
of optimizations/bulkcomm/alberto/Block/3dStrideTest, the affected chunk
was different for the A and B arrays, but doiBulkTransfer() wasn't coded
to handle that.  Fix doiUseBulkTransfer() to be more precise about what
we can deal with, and doiBulkTransfer() to actually handle everything we
can.

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -2189,13 +2189,33 @@ module DefaultRectangular {
     if debugDefaultDistBulkTransfer then chpl_debug_writeln("In DefaultRectangularArr.doiUseBulkTransfer()");
 
     //
-    // With multi-ddata, at least for now if the arrays aren't chunked
-    // exactly the same way we don't do direct bulk transfer.
+    // With multi-ddata, at least for now either the entire transfer has
+    // to involve just one chunk of each array, or the arrays have to be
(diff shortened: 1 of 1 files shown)


gbtitus:
@benharsh , would you take a look at this one?  No rush -- this is not causing any failures in nightly testing.  I've marked it work-in-progress because without it being a rush, it might well be better to go ahead and do the complete implementation that would allow bulk transfer between any two contiguous arrays no matter whether or how they were chunked.

benharsh:
Looks good to me, sorry for taking so long!

gbtitus:
We're currently reviewing multi-ddata to decide if we want to continue with it in the face of performance challenges.  If we do continue with it I'll probably instead do the more-complete solution I mentioned earlier, but I'm leaving this PR here as a placeholder for now.

gbtitus:
I've gone back and tested optimizations/bulkcomm/alberto/Block/3dStrideTest on master with locModel=numa, and it passes.  I've also taken another look at the code on master and believe that it should work in all cases.  (I had fixed it for some other tests that were failing.)  So I believe this PR is now moot, and I'm going to close it.

PR #5109, opened 2017-01-03
Cray Copyright year changed from 2016 to 2017

awallace-cray:
Cray Copyright year changed from 2016 to 2017 in source files (not including third-party).


bradcray:
Because I often refer to these PRs the year following the PR itself, I'm noting here that I don't think we should feel the need to update copyright years in the test/ directory because (a) we intentionally don't bother putting copyrights in every test and (b) those that are here seem to be either unnecessary or else cases of making a snapshot of some code that had the copyright.  But given the nature of the snapshot, it need not be preserved.

Put another way, any script that's used to check for these things should feel free to skip over the test/ directory, in my opinion.  It could arguably scan the test/release/examples/ directory (since we release that).

PR #4074, opened 2016-06-24
Stack allocate on and task function argument bundles

mppf:
We've been heap allocating argument bundles for task bodies and for on statements.
This may have once been necessary but it does not seem to be necessary any longer.

For on statements:
- GASNet already has a model, for AMRequestMedium (which is what we eventually call to do an on statement), in which the passed buffer can be reused immediately after that GASNet function returns. Additionally, the buffer need not be in communication registered memory.
- Generally, we are trying to get to task stacks being allocated from communication-registered memory anyway. Additionally, the communication layer is free to copy into a heap allocated area if necessary. Lastly, on "on" statement causes the arguments to be copied over the network, so it is a fiction that by heap allocating them we preserve any sense of identity.

For tasks:
- tasks-qthreads.c already calls qthread_fork_copyargs and more generally qthread_spawn copies arguments to a thread (and heap allocates in some case if they are too big).
- for FIFO, the tasking layer was already calling the allocator for each spawned task in order to create a new task pool entry. This allocation can easily be extended to include space for the argument bundle.
- Any tasking layer that does not have the capability to copy some variable-length arguments to the task can heap-allocate and copy the arguments.

One issue with no longer heap-allocating the argument bundles is that the tasking and communication layer need to pass some additional information to the remote locale / executing task. For example, comm-gasnet.c needs to pass the requesting nodeID in order to handle execute_on_nb. Since the network communication will copy a single buffer, if the size of this comm-layer information was private, the comm layer would have to heap or stack allocate a new the bundle (that it includes the extra information) and copy in the old bundle's information. At the same time, since the Chapel comm layer interface is for the code generated by the compiler (vs a user-facing API), there is not much encapsulation benefit to making the size of the comm-layer information private in this way. Therefore this PR uses the alternative strategy, which is to allocate all argument bundles with a runtime header field already included. The runtime uses this space as needed in order to pass data (like the requesting nodeID). This extra field is included as the first field in the argument bundle class. The entire argument bundle is passed to on/task wrapper functions and this extra field is ignored at that point.

Note that in some cases a communication layer might prefer to copy the argument bundle into a payload of a different sort of message with a different sort of header. In such cases, the original argument bundle can be formed again in a stack variable on the other side. I expect that such an approach is fine for small message sizes.

[...]

PR #5110, opened 2017-01-04
massivethreads works with stack-allocated task bundle

taura:
with this patch, massivethreads tasking layer passes  nightly -examples with no errors on single locale.
I will do fuller tests later.

### DIFF
--- a/runtime/include/tasks/massivethreads/tasks-massivethreads.h
+++ b/runtime/include/tasks/massivethreads/tasks-massivethreads.h
@@ -58,7 +58,8 @@ typedef struct {
   c_sublocid_t requestedSubloc;
   chpl_fn_int_t requested_fid;
   chpl_fn_p requested_fn;
-  chpl_task_prvData_t prv;
+  chpl_taskID_t id;
+  //chpl_task_prvData_t prv;
 } chpl_task_bundle_t;
 
 #endif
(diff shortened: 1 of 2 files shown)


runtime/src/tasks/massivethreads/tasks-massivethreads.c
```diff
@@ -699,13 +683,15 @@ void chpl_task_setSerial(chpl_bool state) {
 
 // Get pointer to task private data.
 #ifndef CHPL_TASK_GET_PRVDATA_IMPL_DECL
+#if 0
```
mppf: The commented-out code here looks wrong in any case; it should be getting task-private information. See

https://github.com/chapel-lang/chapel/pull/4074/commits/b80da99f4441ebea473b85e6df398771a74aa4d8#diff-d428c9d1e3eb148d6b27e05c03657e5cR72

for one way to do that.

runtime/include/tasks/massivethreads/tasks-massivethreads.h
```diff
@@ -58,7 +58,8 @@ typedef struct {
   c_sublocid_t requestedSubloc;
   chpl_fn_int_t requested_fid;
   chpl_fn_p requested_fn;
-  chpl_task_prvData_t prv;
+  chpl_taskID_t id;
+  //chpl_task_prvData_t prv;
```
mppf: I'd expect it to include chpl_task_prvData here so you can return it in chpl_task_getPrvData.

PR #4922, opened 2016-11-30
Update bugs.rst for GitHub Issues + new privatebugs.rst

ben-albrecht:
- Point users to GitHub Issues from our bug reporting page, now named *Reporting Chapel Issues*
    - Note: link to issues redirects to pull requests until we enable issues
- Remove mentions of JIRA
- Reduce the content in `bugs.rst`, since instructions are in `ISSUE_TEMPLATE.md` and `privatebugs.rst`
- Create a `privatebugs.rst` for instructions on submitting private issues
    - I chose to put the requested information into a code block so that users reporting private bugs could easily copy/paste the requested information into an email body, preserving spacing/formatting. If there is opposition, we can put this back into rendered markup.
    - Open question: Should we adopt and document a security policy similar to [Go's security policy](https://golang.org/security) ?
- New custom sphinx extension that adds a `:disguise:` role for disguising email addresses with HTML comments

I think we can open up GitHub Issues in conjunction with merging this PR

PR #5113, opened 2017-01-04
Utilize disguise role to obfuscate email addresses in docs

ben-albrecht:
The disguise role was added back in #4922 to disguise an email address. This PR uses the role to disguise other email addresses in the docs.


ben-albrecht:
**Open Question:** Does this disguise role actually help, since the plain text email address is still in the source file, which is linked from each web page? Also, the undisguised addresses will continue to exist in the older versions of docs.

Issue #5114, opened 2017-01-04
Presence of [parallel?] these() enables promotion


bradcray:
Today, promotion is enabled by the presence of a _promotionType field within a class/record, and is not a mechanism intended for end-users.  This property is also propagated to certain iterators created by the compiler to represent forall expressions, albeit in a somewhat fragile way.

My proposal is that promotion should be enabled not via this special field, but by the presence of default iterators ('these()' iterators).  The ability to iterate over a type seems like a natural means of saying that the type should enable promotion.  This would remove what's arguably a roundabout way of indicating that something can be promoted.

We usually think of promotion as generating data parallel execution, so there's a question in my mind as to whether a serial these() iterator should be sufficient to enable (serial) promotion or whether a parallel iterator (standalone or leader-follower) should be required.

TODO:
[] file futures for this issue

bradcray:
Tagging @mppf due to his request for this issue.  Tagging @vasslitvinov and @daviditen because I think you two are most likely to have worked in the areas that this would require improving recently (though others may have as well).  This is not at all high priority, but I consider it a wart in the language and if it were an easy fix...  (assuming others didn't object to the proposal).

vasslitvinov:
To the question whether serial these() should suffice for promotion, I suggest that the answer be the same as whether a forall-loop is legal when the iterable expression has only serial these() and does not have any parallel variants.

The latter has a .future here:

    test/functions/iterators/vass/forall-without-leaderfollower.future

benharsh:
I'm working on replacing this field with a method, ``chpl__promotionType()``. This method will return a type, the same type as the current ``_promotionType`` field.

I'm running into an issue with the order in which we resolve initializers and fields. I think what we need is to set the scalar promotion type **after** phase 1 is resolved but **before** phase 2 is resolved. We should be able to resolve methods and initializers (eventually) with only phase 1 resolved. Phase 2 can contain arbitrary code, meaning it may include promotion over the type being resolved.

I'm not sure how to solve this. Perhaps some special call/primitive inserted right after phase 1?

benharsh:
Alternatively we could disallow promotion over the initialized type during phase 2.

vasslitvinov:
One part of this issue is that when we promote something that has a standalone iterator, we should use that one. Currently promotion does not look at standalone iterators, choosing the serial iterator instead when there is no leader.

bradcray:
[...]

PR #5115, opened 2017-01-04
Add tests of promotion using forall expressions and user-defined types

bradcray:
This adds a handful of tests that exercise promotion to clarify some corner cases that I (and perhaps @mppf ) misunderstood while working on the users guide.  Specifically:

* forallPromotes*.chpl demonstrate that forall expressions generating type 't' result in promotion to functions accepting type 't' regardless of what's driving the forall expression.  This contradicted my original assumption that they wouldn't cause promotion and my intermediate belief that it only worked when iterating over types that resulted in promotion.

* theseShouldPromote*.chpl propose that it should be the presence of a these() iterator that results in promotion rather than the presence of a "_promotionType" field as in the current implementation.  Specifically version 1 relies on the presence of a standalone these() iterator; version 2 asks whether a serial version should be sufficient; and version 3 relies on the presence of leader/follower iterators.

These tests are related to issue #5114.


mppf:
thanks!

PR #5116, opened 2017-01-05
chplspell: replace use of shopt -u globstar

cassella:
bash prior to version 4.0 doesn't support the globstar option.  Macs
only have bash 3.2 installed by default.  So replace the use of
globstar globbing with a $(find) command.

The file lists produced by find are sorted by directory, so we don't
need to sort them any more.

The sorting was the only place the globs in DEFAULTFILES[] were
resolved, so those need to be explicitly resolved now, after cd'ing to
$CHPL_HOME.

### DIFF
--- a/util/chplspell
+++ b/util/chplspell
@@ -43,12 +43,20 @@ if ! $RIVENV $SCSPELL --help > /dev/null 2> /dev/null; then
     exit 1
 fi
 
-# The "globstar" option below lets these **/ patterns recurse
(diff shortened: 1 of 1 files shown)


cassella:
@bradcray I don't have a bash 3 to test with.  Do you want to give this a try with yours?

bradcray:
There is no longer a warning and I was able to run it in "point to a directory" mode and have it find an intentional misspelling that I injected for the sake of testing.

PR #5117, opened 2017-01-05
Add some .futures about type methods

daviditen:
Type methods currently accept any type as the receiver, so defining a type
method with the same name in two different classes is ambiguous instead of
using the type of the receiver to disambiguate.

Also, a type method defined in class "C" can be called on any type, for example
"int".

Both of these can be worked around with simple where clauses.

### DIFF
--- a/test/classes/diten/typeMethodOverload.bad
+++ b/test/classes/diten/typeMethodOverload.bad
@@ -1,0 +1,3 @@
+typeMethodOverload.chpl:13: error: ambiguous call 'type C1.typeMethod()'
+typeMethodOverload.chpl:2: note: candidates are: _any.typeMethod()
+typeMethodOverload.chpl:8: note:                 _any.typeMethod()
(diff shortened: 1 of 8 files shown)


bradcray:
Geesh, this is embarrassing...  David, is the simple fix here to have the compiler introduce the where clauses itself when it's parsing/creating a type method?  Are you looking into that at all?  Thanks.

daviditen:
I haven't looked into fixing it yet, but I think adding the where clauses during parsing would seem to be the easiest and fastest way to do it.  Another approach would be to add something like disambiguate_by_type_receiver() to resolution.  The resolution approach feels more principled to me, but possibly higher effort.

bradcray:
I was thinking of the parser-based approach as well.  It doesn't seem *un*principled...  After all, essentially that is what the user has requested and whatever expression falls before the '.' is the one the where clause should eventually match against.  I do wonder whether it would break down in the face of generics, but generics are apparently not working very well with type methods right now anyway (I think Michael filed some futures against them?)

daviditen:
PR #5153 fixes these.  I started out with the where clause approach, but then realized I could just change the type of the receiver from _any to the type of the class/record defining the method.

PR #5120, opened 2017-01-06
Re-arrange performance tracking suite and graphs

bradcray:
This change was motivated by a longstanding desire in the perf team to update the list of graphs in the "perf tracking" suite (the ones we triage weekly) to get rid of some cases that didn't feel useful and add other cases that we felt we were lacking.  While here, I also did some cleanup.

In more detail:
* added/removed graphs for the perf tracking suite as generally discussed with the perf team
* tried to reorganize the perf tracking suite a bit to be more logical, though this is thwarted in the web view by our heuristic for how to order graphs
* moved fasta-redux into the shootout-extras category because the CLBG doesn't track it anymore
* alphabetized shootouts when they appear
* added some of the missing shootouts to the release examples suite
* split some of the graphs in test/performance/sungeun with wildly varying performance into pairs of graphs in order to make the results easier to track


bradcray:
@chapel-lang/perf-team : Here's my proposed refactoring of the perf tracking suite.  I'd suggest reviewing by creating a copy of the perf graphs webpage using this PR and browsing it there rather than reading the sources themselves (and then using the "invert selection" button :) to see what is being omitted).

@ronawho: Do you remember the arcane heuristic we use to determine what order to render the graphs on the main page?  I tried to rearrange the graphs to make them appear in a more logical order, but failed.

ronawho:
I'm thought it's the order they're listed in, and we just don't repeat them if they're listed in multiple suites

lydia-duncan:
I believe we've set it up so that the performance tracking suite doesn't affect the order the graphs are listed in, but that every other suite does.

bradcray:
@ronawho: I thought that that was the order for the sidebar (and that maybe there was some sort of "bottom-up" ordering to focus on the primary suites over the meta-suites?).

@lydia-duncan: I'd forgotten about any sort of special treatment for the performance tracking suite, if there is any

bradcray:
(In any case, don't let the question of graph ordering get in the way of reviewing the PR... it was just a side question)

lydia-duncan:
I'm pretty sure there is, because otherwise the performance tracking suite being listed so early would make it so that graphs like EP size B would not be listed right next to the other EP sizes which aren't in the performance tracking suite, for instance.

lydia-duncan:
(Not that you're doubting me, just providing evidence :) )

bradcray:
@lydia-duncan : My memory was that we sorted graphs (on the sidebar at least) by reading up the GRAPHFILES list in reverse order so that all the EPs would appear next to one another since that would be the first place they'd be encountered.  That wouldn't require special treatment of that suite.  But I could definitely be misremembering -- maybe we do that as well so that all the release examples appear in the same order?  In part what I was wondering is whether we could use that heuristic for the sidebar (which makes sense) while ordering graphs in the  main pane in the order of the suite itself.  But thinking about it a bit more, I suppose the two are linked -- the sidebar order dictates the main order, and the suites just say which checkboxes on the sidebar are checked.

lydia-duncan:
This looks fine to me

ronawho:
@bradcray you're right, we effectively read the list in reverse order since our "meta suites" appeared towards the top: https://github.com/chapel-lang/chapel/commit/88d5f4f50bb6f15ad1282ea8904d474b685c2e94

Issue #5097, opened 2016-12-22
BlockCycDist.chpl:1115: error: halt reached - array index out of bound


cassella:
### Summary of Problem

The program below hits an error,

    $CHPL_HOME/modules/dists/BlockCycDist.chpl:1115: error: halt reached - array index out of bounds: (288)

while processing `A_bs2`.  There are no problems with the very similar `A_bs`.

It hits the error for index (7,7,1) when run with -nl 1, and for index
(5,7,1) when run with -nl 8.  It runs to completion with -nl 4.

### Steps to Reproduce

**Source Code:**

```chapel
use BlockCycDist;

// works
var Space = { 1..8, 1..4, 1..9};
var D_bs = Space dmapped BlockCyclic(startIdx=Space.low, blocksize=(2,2,3));

// doesn't work with -nl 1, -nl 2, -nl 8.  works with -nl 4.
var Space2 = { 1..8, 1..9, 1..4 };
var D_bs2 = Space2 dmapped BlockCyclic(startIdx=Space2.low, blocksize=(2,3,2));


proc show_array_locality(ref A: [?D]) {
  var L: [D] int;
  for ij in D {
      L[ij] = A[ij].locale.id;
  }
  writeln(L);
}

writeln("A_bs");
var A_bs: [D_bs] int;
show_array_locality(A_bs);

writeln("A_bs2");
var A_bs2: [D_bs2] int;
show_array_locality(A_bs2);
```

**Compile command:**
`chpl --devel blockcycind.chpl -o blockcycind`

**Execution command:**
`./blockcycind -nl 1`

In `BlockCycDist.chpl`, I uncommented the writeln() in
`LocBlockCyclicArr.this(i)`, and enabled (and enhanced) the debugging
for the problematic indices in `LocBlockCyclicArr.mdInd2FlatInd(i: ?t)`

The `-nl 1` run ends with

```code
...
(7, 6, 3)->268
(7, 6, 3)->268
(7, 6, 4)->269
(7, 6, 4)->269
0:d = 1
0:blksize = 2
0:ind0 = 6
0:blkNum = 3
0:blkDimOff = 0
0:
0:d = 2
0:blksize = 3
0:ind0 = 6
0:blkNum = 2
0:blkDimOff = 0
0:
0:d = 3
0:blksize = 2
0:ind0 = 0
0:blkNum = 0
0:blkDimOff = 0
0:
0:blockmults = 12 6 2
0:numblocks = 4 3 2
[...]

PR #5121, opened 2017-01-06
test case showing BlockCycDist problem

cassella:
test case showing BlockCycDist problem


cassella:
This is the future requested in issue #5097.  I have a few reservations about it:

1. I don't know what's going on well enough to write a useful description in the future.
2. Especially since it's only a few combinations of parameters that hit the problem.
3. Once the bug is fixed, this combination probably won't be particularly interesting.
4. And the filename will be wrong.
5. If the combination is still interesting, there will probably be lots of other interesting combinations.
6. I'd submit other combinations now too, but I'd like to have the test working in order to generate the .good files for me.  :)
7. I don't know how best to integrate the multi-locale combinations with the test system wrt `.skipifs` vs. `.execopts` vs. `.numlocales`, etc.

benharsh:
I'm looking into this bug and will try to create a more general test that captures the issue. See #5097 for further discussion.

PR #5124, opened 2017-01-07
Make array init method selectable

ronawho:
Allow us to specify noinit, serial init, parallel init, or heuristic init for
arrays. This is intended to help test out the performance implications of
serial vs parallel init for LCALS and other benchmarks where we want to make
sure we get first-touch right. Note that this is a hacky/not user-facing
solution meant to let us test things out until we have a real
initialization story.

We can now do something like:

```chapel
chpl_arrayInitMethod = ArrayInit.parallelInit;
var A: [1..100] int;
chpl_arrayInitMethod = ArrayInit.noInit;
var B: [1..10000000] int;
chpl_arrayInitMethod = ArrayInit. heuristicInit;
```

The default can also be set with ` -schpl_defaultArrayInitMethod=ArrayInit.serialInit`

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -604,11 +604,11 @@ module ChapelBase {
   }
 
   enum ArrayInit {heuristicInit, noInit, serialInit, parallelInit};
-  config param defaultArrayInitMethod = ArrayInit.heuristicInit;
-  var arrayInitMethod = defaultArrayInitMethod;
+  config param chpl_defaultArrayInitMethod = ArrayInit.heuristicInit;
+  var chpl_arrayInitMethod = chpl_defaultArrayInitMethod;
 
   proc init_elts(x, s, type t) : void {
-    var initMethod = arrayInitMethod;
+    var initMethod = chpl_arrayInitMethod;
 
     // for uints, check that s > 0, so the `s-1` below doesn't overflow
     if isUint(s) && s == 0 {
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -603,9 +603,60 @@ module ChapelBase {
     __primitive("chpl_exit_any", status);
   }
 
-  config param parallelInitElts=true;
+  enum ArrayInit {defaultInit, noInit, serialInit, parallelInit};
+  config var arrayInitMethod = ArrayInit.defaultInit;
+
   proc init_elts(x, s, type t) : void {
-    //
+    var initMethod = arrayInitMethod;
+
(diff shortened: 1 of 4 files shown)


bradcray:
Given the concerns voiced at the meeting today about race conditions around global state, how about if we at least rename this PR to "programmer selectable" since we don't really want end-users to use it...  :)

ronawho:
ok

ronawho:
Blegh, either need to avoid using a config var, or will need to update the following tests to expect `arrayInitMethod: ArrayInit` in the --help output:

```
[Error matching program output for execflags/ferguson/help2]
[Error matching program output for execflags/shannon/help]
[Error matching program output for execflags/shannon/configs/help/configVar-Dash]
[Error matching program output for execflags/shannon/configs/help/configVarHelp]
[Error matching program output for execflags/shannon/configs/help/configVarModStrings1]
[Error matching program output for execflags/shannon/configs/help/configVarModStrings2]
[Error matching program output for execflags/shannon/configs/help/configVarSetOver]
[Error matching program output for execflags/shannon/configs/help/configVarSetTwoTypes]
[Error matching program output for execflags/shannon/configs/help/configVarTwoModules]
[Error matching program output for execflags/shannon/configs/help/varNameEnumQM]
[Error matching program output for execflags/shannon/configs/help/varNameQMark]
```

bradcray:
Can config vars be declared 'private'?  If so, maybe we could squash such config vars from the help message?  (I'm half joking, but half not)

ronawho:
already tried :(

You can make config vars private, but it doesn't help. The printing code is all in the runtime, so there's no easy way prevent one from being printed out

mppf:
yeah but the compiler is making the list of config vars for the runtime. I'm not saying making them "private" is or is not a good idea - just that I wouldn't expect it to be a big compiler mod. But in the meantime, if this config var is holding this PR up, could we just change it to a config `param` which doesn't (yet) show up in any help output?

ronawho:
@mppf the code that sets a config var from the command line is in the runtime too, so I don't think there's a clean way to modify things so that one config var isn't printed in the --help, but have it still work normally otherwise.

The plan for now is to just make a config param for defaultArrayInitMethod and change arrayInitMethod over to a regular (non config) var. Nothings really holding up this PR other than me actually making that change.

ronawho:
@benharsh changed to a config param for the defaultArrayInitMethod. Also changed defaultInit to heuristicInit. Mind re-reviewing?

benharsh:
Looks good to me. Only feedback (which you can consider optional) would be to add a "chpl__" prefix to the global variable to keep the namespace cleaner.

PR #5126, opened 2017-01-07
Adding API for Task ID comparison and conversion to string

srirajpaul:
API is required for comparing task IDs that are not integers,
since other types such as struct won't work with binary operators.
Also adding a 'to string' API to help with the printing of task IDs.

### DIFF
--- a/runtime/include/chpl-tasks.h
+++ b/runtime/include/chpl-tasks.h
@@ -228,6 +228,20 @@ c_sublocid_t chpl_task_getRequestedSubloc(void);
 chpl_taskID_t chpl_task_getId(void);
 
 //
+// Checks whether two task IDs are the same
+//
+chpl_bool chpl_task_idEquals(chpl_taskID_t, chpl_taskID_t);
+
+//
+// Returns the string representation of task ID
+// The string returned is the same buffer passed as argument
+//
+char* chpl_task_idTostring(
+                           char *,         //buffer on which ID is written
+			   size_t,         //length of the buffer in bytes
+			   chpl_taskID_t); //Task ID
+
+//
 // Yield.
 //
 void chpl_task_yield(void);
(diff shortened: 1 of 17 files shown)


runtime/include/chpl-tasks.h
```diff
@@ -228,6 +228,20 @@ c_sublocid_t chpl_task_getRequestedSubloc(void);
 chpl_taskID_t chpl_task_getId(void);
 
 //
+// Checks whether two task IDs are the same
+//
+chpl_bool chpl_task_idEquals(chpl_taskID_t, chpl_taskID_t);
+
+//
+// Returns the string representation of task ID
+// The string returned is the same buffer passed as argument
```
mppf: I'd make this function return void, personally.

mppf: I take that back. I think it should return whether or not the operation was successful. E.g. return 0 if the task ID was encoded in the string, or 1 if there was not enough space.

srirajpaul: I returned the buffer, so that the tostring() function can directly be used for printing. Somewhat similar to how tostring() is done in java and c#. 
For example, 
printf(“task id %s used %d cycles\n”, chpl_task_idToString(buff, 10, ID), num_cycle);

Otherwise, we will have to invoke the tostring() function in a separate call and then use the buffer in printf.

One solution is that the implementation of chpl_task_idToString should check whether the buffer was big enough and the conversion is successful. If not, print a warning (chapel_warning()) and proceed. I don't think any error in toString() should cause the program to exit.

srirajpaul: As Greg suggested, I will return NULL in case of any error. Thus if the callers want to identify errors, they will be able to do it.

mppf: OK. Let's update the comment here and also check the return value of the snprintfs.

mppf:
I don't know how you merged with master, but something is not right. The Diff page on the GitHub PR should show only your changes.

srirajpaul:
I did some mistake with merging. I will close this pull request and create a new one.

PR #5134, opened 2017-01-10
Change array blank intent, fix up ref-pairs

### LINKED ISSUE #5217: default intent for arrays, const, and ref-pairs
This issue is a place to discuss an issue with ref-pairs and blank array intents. Currently, the language specification describes blank argument intent for arrays as `ref`. That has some possibly unintended consequences and this issue brings up motivation for considering refinements.

# Starting Questions

Probably the quickest way to bring up the question is to ask what the following two programs should do:

## Array blank intent with ref-pair

```
var globalArray : [1..4] int;

proc getter() ref {
  writeln("REF");
  return globalArray;
}

proc getter() const ref {
  writeln("CONST REF");
  return globalArray;
}

proc writeit(x) {
  writeln(x);
}

writeit(getter()); // Should this print "REF" or "CONST REF" ?
```

It currently prints "REF" but I think that's surprising.

## Array blank intent with const variable

```
const A = [1,2,3];
foo(A);
proc foo(data) {
  data[1] = 100;
}
```

This program should almost certainly result in compilation error but it does not currently.

# Motivation

First, why do we have a ref-pair feature and why do we have default intent for arrays described as `ref` in the spec?

## Why do we have a ref-pair feature?

The ref-pair feature (formerly `setter`) enables things like sparse arrays. In the case of sparse arrays, an element that is in the rectangular domain but has a zero value can be read or written, but the implementation can have different behavior based upon whether the element was read or written.

For a sparse array, reads can return the `zero` value (called `IRV`). But for writes, it is an error if the element isn't already in the sparse domain's set of nonzero indices.

This comes up too for associative domains, where an associative array which does not share its domain will automatically add to the domain when an array element is set; but it will raise an index-out-of-bounds error if it is read.

We can proceed with either case as an example but the point is that there will be fundamentally different behavior depending on whether an array access is a `read` or a `write`.

## Why is the default intent for arrays ref?

Now, about the default intent for arrays. I think it is 'ref' for programmer convenience. As in less typing. The common case that we'd like to not have to annotate is that the array elements are modified.

I've argued in the past that the default intent should be 'const ref' for arrays. Brad responded with two principles:
* most programs should be able to be written using only blank intents
* blank intents should minimize the surprises you get if you don't quite understand what they do.

(see [1] if you want to see the entire discussion).

## What is going wrong

OK, so perhaps now we understand the motivation/purpose of the ref-pair feature and the arrays-blank-intent-is-ref feature.  Now what is going wrong?

## Array behavior depends on element type

The program below demonstrates how an associative array behaves differently depending on whether or not it is storing integer or array elements.

```
// First, let's try some of these things with an
// associative array of ints:
{
  var D:domain(int);
  var A:[D] int;

  // This adds index 1 to the domain, implicitly
  A[1] = 10;
  writeln(D.member(1)); // outputs `true`

  // This will halt, because index 2 is not in the domain
  //var tmp = A[2];

  // This will also halt, for the same reason
  //writeln(A[3]);
}

// Now, let's try the same things with an array of arrays:
{
  var D:domain(int);
  var AA:[D] [1..4] int;
  var value:[1..4] int = [10,20,30,40];

  // This adds index 4 to the domain, implicitly
  AA[4] = value;
  writeln(D.member(4)); // outputs `true`

  // This will halt, because index 5 is not in the domain
  //var tmp = AA[5];

  // It seems that this *should* halt, but it does not now
  // Instead, it adds index 6 to the domain
  writeln(AA[6]);
  writeln(D.member(6)); // outputs `true` !
}
```

Now, you might say "clearly the author of writeln needed to use the `const` argument intent". That brings me to my next point.

## Generic functions have to think about this to be correct

Since Chapel is a language supporting generic functions like writeln - I think it is arguably a language design problem if writeln appears to work correctly everywhere except for arrays-of-arrays. That's the kind of latent bug we'd like our compilers to find or our language design to prevent.

Normally it is not a big issue if generic code has some capability that it does not use - for example the ability to modify an argument passed by blank intent.  However, this 'ref' default has an impact on generic code when it interacts with the ref-pair feature.

For example, what if I wanted to write a nice little generic function to print out all of the array elements in a particular domain.

```
proc printArrayElementsInDomain(A:[], D:domain) {
  for i in D {
     writeln("A[", i, "] = ", A[i]);
  }
}
```

So far so good. Let's run this generic function over an array of integers:

```
{
  var A:[1..10] int;

  printArrayElementsInDomain(A, {1..2});
}
```

That outputs
```
A[1] = 0
A[2] = 0
```

as expected, so far so good. But now what if we use a sparse array of arrays?

```
{
  var dense = {1..10};
  var sps : sparse subdomain(dense);
  var A : [sps] [1..5] int;

  printArrayElementsInDomain(A, {1..2});
}
```

Uh-oh! on master this causes a halt like this:
aa3.chpl:3: error: halt reached - attempting to assign a 'zero' value in a
sparse array: (1)

I think that is surprising, since it wouldn't halt for a less-complicated array type. As I said, it's also likely to be a latent bug. And I'm not convinced that fixing 'writeln' is the answer, since printArrayElementsInDomain could have easily called some other generic function accepting an element by default intent.

I don't think that this behavior meets the goals described above for blank intents.

Let's consider how lvalue checking might also be a hazard for authors of generic code.

## Inconsistency in lvalue checking

Let's consider the following program:

```
  proc producer(x) { return x; }
  proc consumer(x) { writeln(x); }
  proc ref_consumer(ref x) { writeln(x); }
```

First, let's try an example with an integer:

```
  consumer(producer(1));
```

OK, consumer gets the returned value from producer by blank intent (which is const in for integers).

```
  ref_consumer(producer(1));
```

This one results in an error:

```
aa4.chpl:6: error: non-lvalue actual is passed to 'ref' formal 'x' of ref_consumer()
```

Which is expected. The ref_consumer function has a ref formal, so we need to pass in something other than a compiler-introduced temporary to it. These l-value errors prevent programmers from inadvertently modifying a temporary. I think that's a useful language feature.

What if we do the same things with arrays?

```
  consumer(producer( [1,2] ));
```

This one almost has to work according to the intents design goal of "most programs should be able to be written using only blank intents". And in fact it does work. But, the default intent for arrays is `ref`. Shouldn't this be an error?

Let's consider the more explicit example:

```
  ref_consumer(producer( [1,2] ));
```

This one also works but almost certainly should be an error.  Now we have a problem with generalizing generic code - if the code was only ever tested for arrays, it's easy to imagine it wouldn't even compile for integers.

# Related Problems with `const`

## `const` arrays can be modified

While discussing this issue, we noticed that `const` for array variables is not being respected. In particular, the following program:

```
const A = [1,2,3];
foo(A);
proc foo(data) {
  data[1] = 100;
}
```

compiles and runs now but should be some sort of compilation error, since the const array `A` is modified. A related issue is that an array returned with `const ref` return intent should not be modifyiable. This is the case one would see when passing a sparse array's `IRV` to a function by blank intent.

```
  var globalIRV:[1..10] int;
  proc getIRV() const ref {
    return globalIRV;
  }

  proc mutate(A) {
    A[1] = 1;
  }
  mutate(getIRV()); // should be a compilation error
```

## blank intent, const, and task intents

Note that blank intent currently drives task and forall intents. This is particularly important for promotion, where there isn't currently a syntax to override the blank intent.

For example,

```
var D = {1..10};
var A : [D] int;
forall i in D {
  A[i] = 3;
}
```

or related:

```
	A[I] = 3;  // where I is an array of index(A.domain)
```

## Chaining ref-pairs for arrays-of-arrays

It seems that the below strategy ought to work to enable arrays-of-arrays that create elements on access to work in a reasonable manner.

```
  record R {
    proc ref getOrCreate() ref {
      // return existing element, or create & return if it doesn't exist
    }
    proc const getOrCreate() const ref {
      // return existing element, error if it doesn't exist...
    }
  }

  // suppose outer's element is another record R containing an int element
  outer.getOrCreate().getOrCreate() = 1;
```

# References

[1] See also thread from 2012 on chapel-developers "argument intents".

https://sourceforge.net/p/chapel/mailman/message/28928289/

I was proposing that the default argument intent always be
`const`. In that thread, Brad said:

> For #1, I'm a bit more sympathetic to the argument ("wouldn't it
> be nice if blank always meant the same thing?").  But I'm also not
> convinced that it's been that much of a point of confusion for
> users.  And I think your  proposal adds burden in the common case
> to users of sync vars, domains, and arrays, which I would argue
> are very important common cases to worry about keeping simple.
> Put another way, our philosophy has traditionally been "most
> programs should be able to be written using only blank intents."
> And along with that, "blank intents should minimize the surprises
> you get if you don't quite understand what they do."  I'm not
> convinced that the consistency gained by making blank mean the
> same thing for all types is worth the burden to the programmer
> (read: number of keywords I'd have to type) for a
> typical Chapel program using domains, arrays, and sync vars.

### LINKED ISSUE #5219: approach for #5217: ref-if-modified
This issue describes a possible approach to solve the problems in #5217.

The particular approach under consideration here is that blank intent be `ref` or `const ref` for arrays, depending on whether or not the array formal was set or used in a setting context in the body of a function.

I.e. the compiler would choose what blank intent means based on how the formal is used rather than based upon the type.

In some ways the analysis to compute blank intent would be similar to choosing one or the other branch of a ref-pair.  I think it could proceed in a bottom-up manner. E.g. the following things would cause the formal intent to become ref:
- 'ref' variable declared pointing to formal
- formal passed to another function by 'ref', 'inout, or 'out' intent (including functions we recursively decide have blank intent meaning 'ref').
- formal assigned to (but that's arguably passing to proc = with ref intent).

For arrays, modifying an element of an array formal would cause the concrete intent to be 'ref'. (I *think* this will "fall out" with minor adjustments to the ref-pair 'this' methods. We might need to add support for the idea of ref-pair on iterators though).

For task/forall intents, the same rule would apply to the body of the task/loop as would to the body of a function with an array formal. If the array is modified, it should be "passed" to the task with `ref` intent.

Pros:
* fits with current language - programs that work now still work
* can apply to task/forall intents as well as argument intents
* arguably meets blank-intent design goals:
    * most programs should be able to be written using only blank intents
    * blank intents should minimize the surprises you get if you don't quite understand what they do.
* integrates with analysis to handle ref-pairs - both are backward
* conceptually matches with the idea that blank intent is a shorthand

Cons:
* more complicated in concept than some alternatives
* definitely more complicated to implement than resolveIntents is now
* it might require explicit programmer help for recursive functions (but maybe not)
* it might be surprising how changing one function can change the behaviors of callers to that function
* Almost certainly will need to explicitly choose `const` or `ref` for array formals in extern procs.

mppf:
Closes #5217. Closes #5219.

This PR changes the blank intent for arrays to be "ref-if-modified" (see #5219). Additionally, it adjusts behavior of the ref-pair feature when arrays, domains, and distributions are returned. Additionally it adjusts the implementation to better handle _this arguments in method calls.

This change is use-visible. In particular, the following program has different output. Before this change, it prints out REF, but after, it prints out CONST REF:

``` chapel
var globalArray : [1..4] int;

proc getter() ref {
  writeln("REF");
  return globalArray;
}

proc getter() const ref {
  writeln("CONST REF");
  return globalArray;
}

proc writeit(x) {
  writeln(x);
}

writeit(getter()); // Should this print "REF" or "CONST REF" ?
```

This language change was discussed in great detail in issues #5217 and #5219.

The overall approach here is:
 * add a flag, FLAG_DEFAULT_INTENT_IS_REF_MAYBE_CONST, to mark a type as being subject to blank-intent-is-ref-if-formal-is-modified behavior. Add INTENT_REF_MAYBE_CONST to reflect this intent.
 * significantly rework cullOverReferences to handle ref-if-modified and to properly chain ref-pair calls.
 * fix up some bugs in other areas of the implementation
 * add (currently disabled) late const checking, so that we can detect cases like this:
``` chapel
const constArray : [1..4] int;

proc setElement(x) {
  x[1] = 1;
}
```

Here is an overview of what changed. In most cases, the commit messages have more detail.

headers:
 * add FLAG_DEFAULT_INTENT_IS_REF_MAYBE_CONST and intent INTENT_REF_MAYBE_CONST
 * add QualifiedType.isConst, QualifiedType.toConst

astutil:
 * adjust isDefAndOrUse so that actuals passed to a formal accepting a record-wrapped type (array/domain/distribution) use the normal intent machinery instead of always considering the actual to be defined.
 * added, in commented-out form, a fix to make isDefAndOrUse properly treat PRIM_SET_MEMBER and PRIM_SET_SVEC_MEMBER. These are commented out because of issues with copyPropagation.

deadCodeElimination
 * use getSettingPrimitiveDstSrc so that this code will be ready for future improvements to isDefAndOrUse

buildDefaultFunctions
 * mark "setter" accessor functions for records as taking in _this by REF.
 * make initCopy take in an INTENT_CONST argument

checkResolved:
 * include INTENT_REF_MAYBE_CONST in checking for returning ref argument by ref

flattenFunctions:
 * use INTENT_REF_MAYBE_CONST for formals for captured variables

resolveIntents:
 * array wrapper records use INTENT_REF_MAYBE_CONST
 * updated blankIntetnForThisArg to better handle domain/distribution and to take in record _this argument by const ref. (But this isn't the whole story).
 * created constIntentForThisArg and use it in concreteIntentForArg.

cullOverReferences:
 * adds gatherLoopDetails which I'd like to move to ForLoop.cpp once it's correct. This collects information about a ForLoop, including what is iterated over and what the index variables are. This is complicated by the current AST structure for these loops and amounts to AST pattern matching (but Elliot tells me that's the best we can hope for right now).
 * refNecessary in cullOverReferences no longer contains a special "heuristic" to always pass arrays by ref.
 * refNecessary no longer always passes _this arguments by ref.
 * refactored and added to significantly in order to support ref-if-modified blank intent for arrays. This analysis is inter procedural.

functionResolution:
 * ignore const-ness errors for functions marked FLAG_REF_TO_CONST_WHEN_CONST_THIS since these need to be handed after the analysis in cullOverReferences.
 * in handleTaskIntentArgs, clear INTENT_FLAG_MAYBE_CONST when setting INTENT_FLAG_CONST
 * in lvalueCheck, ignore INTENT_REF_MAYBE_CONST

implementForallIntents:
 * handle INTENT_REF_MAYBE_CONST

modules:
 * tidy up const-ness
 * use const 'this' argument on array accessors returning const. This allows ref-pair analysis to propagate.
 * mark _array with "default intent is ref if modified"

primitive:
 * added some comments to indicate strucutre of some primitives
 * added getSettingPrimitiveDstSrc utility function

@benharsh reported that a multilocale miniMD run goes from 60s to 9s with this change.

Passed full local testing (8df75360460a7c1b5271a7b094b5a274e761b3e8, 9220a376729d4e37743b817be45bcf02699d88de, f968f7057e188c35839729141b4891b530b7f22d, 0c729a3e989eaee9bbfd2acd9f6c3485a6b9ba01, 0b88aaf46f37587c13224a49c4b9a980dee1f94a).
Passed --no-local testing.

Future Work:

* revisit fix to isDefAndOrUse if we remove copy propagation
* check `const` this intent is const-checked
* sort out _this intent for records (see issue #5266)
* update language specification for ref-if-modified (PR #5444)
* get the compiler accurate enough with const propagation to re-enable the late const checking
   - wrapper functions for promotion, default arguments are the main known challenge here
* improve cullOverReferences to better handle tuples; it should make a const-or-not-const determination for each tuple element of ref-if-modified-type in tuple arguments.
* move gatherLoopDetails into ForLoop.cpp and check that it is correct in a --verify step 
* rename cullOverReferences to something indicating what it does


mppf:
const-checking support will need to be updated to get this PR to pass testing.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -230,7 +230,7 @@ static void createShadowVars(DefExpr* defChplIter, SymbolMap& uses,
       // See if we want to prune it.
       bool pruneit = false;
 
-      if (tiIntent == INTENT_REF) {
+      if (tiIntent == INTENT_REF || tiIntent == INTENT_REF_MAYBE_CONST) {
```
mppf: @vasslitvinov - see this change. It sounds like changing forall intents to pass by `ref` will be hard until we can remove `_toLeader` `_toStandalone` because those use the default intent.

mppf:
This change triggers a bug in copy propagation for test/functions/deitz/nested/test_nested_var_iterator2.chpl

mppf:
I believe 35dd9fc4622113f1f30ff1cc7dd57bb0602b7d9d will pass local testing.

mppf:
8df75360460a7c1b5271a7b094b5a274e761b3e8 passed full local testing.

mppf:
With version 66f613a, there are about 20 tests that fail for the reason of incorrect const checking (i.e. a problem within the compiler).

```
users/shetag/unboundslice2.chpl
test/release/examples/users-guide/datapar/promotion.chpl
test/performance/compiler/bradc/cg-sparse-timecomp.chpl
npb/cg/bradc/cg-dense.chpl
npb/cg/bradc/cg-sparse.chpl

 -> seems to be a problem with a forall expr

studies/shootout/submitted/knucleotide.chpl
release/examples/benchmarks/shootout/knucleotide.chpl

 -> 'in' intent for arrays causes it to be considerd CONST

studies/shootout/submitted/fasta.chpl

 -> actual const checking errors, but I don't want to change the submitted version
 
studies/madness/aniruddha/madchap/test_diff.chpl
studies/madness/aniruddha/madchap/test_gaxpy.chpl
studies/madness/aniruddha/madchap/test_likepy.chpl
studies/madness/aniruddha/madchap/test_mul.chpl
studies/madness/aniruddha/madchap/test_showboxes.chpl

 -> sets const field in fn called from constructor
 
test/studies/hpcc/PTRANS/jglewis/ptrans_2011-blkcyc.chpl
studies/cholesky/jglewis/version2/elemental/block_distribution/test_elemental_cholesky_block.chpl
studies/cholesky/jglewis/version2/elemental/fully_blocked/test_fully_blocked_elemental_cholesky.chpl
studies/cholesky/jglewis/version2/elemental/strided/test_elemental_cholesky_explicitly_symmetric_strided.chpl
studies/cholesky/jglewis/version2/elemental/test_elemental_cholesky.chpl
studies/cholesky/jglewis/version2/elemental/unsymmetric_indices/test_elemental_cholesky_unsymmetric_ranges.chpl

 -> ? problem with getting locale -> ref ?
 
test/distributions/robust/associative/basic/zipper.chpl
```

modules/standard/IO.chpl
```diff
@@ -4557,7 +4556,7 @@ proc channel.read(type t ...?numTypes) where numTypes > 1 {
 
 // documented in style= error= version
 pragma "no doc"
-inline proc channel.write(args ...?k, out error:syserr):bool {
+inline proc channel.write(const args ...?k, out error:syserr):bool {
```
benharsh: Are the added formal intents a way to get things working more quickly, or do you seem them as a long-term necessity?

For example, the 'const' intent on channel.write's args seems like it should be unnecessary.

test/domains/bradc/modifyDomainArg.chpl
```diff
@@ -8,7 +8,7 @@ writeln("A is:\n", A);
 
 foo(D);
 
-proc foo(dom) {
+proc foo(ref dom) {
```
benharsh: Were ref intents added on domains because const-checking is better?

modules/standard/IO.chpl
mppf: Right now, we're not tracking const-ness within tuple elements. So you can possibly do things like capture a ref to a const array in a tuple formal argument and then modify it.
I think it's reasonable for documentation to use the `const` intent on these `write` functions but it shouldn't need to be there, long-term.

test/domains/bradc/modifyDomainArg.chpl
mppf: Yes. The new cullOverReferences.cpp code includes late const checking that is currently off by default. However, I've run testing with it on and tried to correct as many cases as I could where the tests were wrong. I don't think I'm very far from having the late const checking working properly (and passing testing) but I don't want the other elements of this PR to be held up by it.

mppf:
@benharsh - thanks for the review! I've already made the minor adjustments you suggested. Re  test/functions/bharshbarg - these all pass now.

TODO: 
- [ ] full testing with -futures

Issue #5081, opened 2016-12-19
futures in Chapel


mppf:
This issue is for tracking the feature request for 'futures' in Chapel.

Some work towards adding futures was completed around 2013. See slides summarizing that work here:

http://chapel.cray.com/presentations/SC13/03-futures-imam.pdf

Here is an email thread about that work:

https://sourceforge.net/p/chapel/mailman/message/30815892/

Here is the branch containing that work:

http://svn.code.sf.net/p/chapel/code/branches/collaborations/futures

Here is a discussion on chapel-users about retiring 'single' and replacing it with 'future'.
https://sourceforge.net/p/chapel/mailman/chapel-users/thread/alpine.LNX.2.00.1309131418050.327%40bradc-lnx.us.cray.com/#msg31409942

Here are the tasks that were outlined to complete 'futures' support in the language:
 - [x] implement future as a library using sync variables
 - [x] add future types and implement begin expressions (ie compiler support for begin expressions)
 - [ ] add automatic coercions from future(T) to T. Compiler adds type inference support and generates calls to read. E.g. `const x = begin foo() + begin bar()`.
 - [ ] add some form of statement-block expressions. The goal here is to enable putting multiple statements inside of begin expressions.
 - [ ] compiler/runtime optimizations. Avoid task creation for short-lived computations.

bradcray:
Two quick notes that are quite likely mentioned in the threads Michael points to above, but which always stick in my mind on this topic:

* In the expression context, 'begin <expr-of-type-T>' should be inferred by the compiler to be of type 'future T'.  Michael's examples above imply this, but I wanted to make it concrete.

* I think there's a semantic/optimization question regarding the interpretation of:

      var x = begin foo() + begin bar() + begin baz();

  Naively, this seems to be saying "create three tasks to compute foo(), bar(), and baz() simultaneously and add their results when they're done."  But given current Chapel semantics (and/or implementation), it is probably more correct to think of it as:

      var t: future int = begin foo() + begin bar()
      var x = t + begin baz();

  which suggests that only two tasks would be running at a time, not three.  To get the original effect, you might need to write:

      var t = begin baz();
      var x = begin foo() + begin bar() + t;

  or maybe:

      var x = begin (begin foo() + begin bar()) + begin baz();

  ??  I think this is acceptable, just counterintuitive.  At the same time, I'm not crazy about going down the road of "Chapel doesn't guarantee the evaluation order of operators like +"...

mppf:
[...]

Issue #5096, opened 2016-12-21
Using classes without needing to delete


mppf:
It would be nice to be able to write Chapel programs that used clasess without ever needing to manually write `delete` statements.

The existing automatic memory management for arrays, domains, and distributions offers some evidence that this idea is possible for arbitrary user-defined classes.

For inspiration, look to C++ shared_ptr and unique_ptr and the Box type in Rust. We should expect that users will be able to get bare pointers out of these types. (Note it's important to be able to call methods or functions working with the bare pointer type somehow - possibly by explicitly getting the bare pointer - because otherwise one wouldn't be able to do much with the type managing the memory).

Initial Design direction:

 - equivalents to C++ shared_ptr and unique_ptr are implemented as records containing a field of class type that points to the managed memory.
 - these records will use existing (but not yet specified) language features of copy-initializer, assignment operator, record destructor to manage the memory
 - these records need to be able to *forward* method invocations to the class

Open Design Questions:

 - If class A inherits from class B, A is a subtype of B. Should e.g. RefCounted(A) be a subtype of RefCounted(B) ?
 - Should a variable of type RefCounted(A) be passable to a function expecting a bare A? (or, would the code making such a call need to use a method to get the bare pointer out of RefCounted(A) ?)
 - Should operators on A be available for RefCounted(A) ?
 - Should methods on A be available for RefCounted(A) ?
 - Should fields in A be available for RefCounted(A) ?

Language Strategy Questions:

 - What language features do we need to make this feature satisfying? The features might include:
    - user-defined coercions
    - direct support for method forwarding
    - `move` function in library akin to C++ std::move
    - more direct compiler support
    - ability to specify operators as methods

mppf:
PR #5058 implements compiler support for a delegation feature, but it doesn't enable the subtyping (should RefCounted(A) be a subtype of RefCounted(B) if A is a subtype of B).

mppf:
C++ accomplishes the forwarding with implicit conversions. Implicit conversions are by definition hidden and can be confusing. For example

```
proc test(a: ?t, b: t) {
  return a + b;
}

var a:uint(8)  = 1;
var b:uint(32) = 2;

var c = test(a,b);

writeln(c);
```

fails to compile, because generic instantiation assumes that no conversions can be applied - it instantiates for an exact type. (This example is inspired by the Stack Overflow question "C++ implicit type conversion with template").

[...]

PR #5135, opened 2017-01-10
Add Futures package module

nspark:
This PR adds a simple package module for creating and accessing futures. See also #5081 for future directions for futures. One big limitation of this module is that declaration and initialization of futures cannot currently be separated.

Poking @mppf @bradcray


nspark:
I decided to give this module an overhaul to address the declaration/initialization issue I raised above. The interface will also be a bit simpler, which I think is good. I'll push some commits in a day or so, so whoever is interested may want to hold off on reviewing the code I currently have out on this PR.

nspark:
I'm a lot happier with the latest set of changes. I think the Futures API is cleaner and more flexible now.

modules/packages/Futures.chpl
```diff
@@ -0,0 +1,240 @@
...
+  pragma "no doc"
+  pragma "auto copy fn"
+  proc chpl__autoCopy(x: Future) {
+    x.acquire();
+    return x;
+  }
+
+  pragma "no doc"
+  proc =(ref lhs: Future, rhs: Future) {
+  if lhs.classRef != nil then
```
mppf: The normal pattern for reference counting is to acquire first and then release, because the lhs and rhs could point to the same thing.
Also I'm not sure you've actually run this method - you'd have to write something like
```
   var future : Future( int );
   future = async(lambda() { return 2; });
```
but then you'd run into the possibility of changing a future while you are waiting for it (possibly leading to deadlock).
IIRC, the Rice group thought that you shouldn't be able to assign futures. I believe you could just declare proc = and make it call compilerError("future assignment not supported") and that's probably better for now anyway.

Note that
```
   var future = async(lambda() { return 2; });
```
does not call `proc =` - it calls chpl__initCopy instead (which we're working on exposing as a `proc init(copyFrom)` initializer.

modules/packages/Futures.chpl
```diff
@@ -0,0 +1,240 @@
...
+
+ */
+
+module Futures {
+
+  use Reflection;
+  use RefCount;
+
+  pragma "no doc"
+  class FutureClass: RefCountBase {
```
mppf: Can you explain why a Future is reference-counted? I suspect there's some simple example you'd like to work but I havn't come up with it myself.

modules/packages/RefCount.chpl
```diff
@@ -0,0 +1,40 @@
...
+
+  pragma "no doc"
+  class RefCountBase {
+    var refcnt: atomic int;
+
+    proc incRefCount() {
+      refcnt.add(1);
+    }
+
+    proc decRefCount() {
```
mppf: Need some comments in this module, at least on `decRefCount`. You have to say it returns 1 if the value has no more references - i.e. that decRefCount returns the count before the decrement operation.

mppf: Also, I'd probably name this class "IntrusiveRefCount" or something like that.

mppf:
This looks pretty good to me. Besides what I've already commented on, I wanted to point out that you can create a record or a class with a 'proc this' and use it instead of a lambda. That might be important if a user of Futures.chpl runs into any of the open problems with lambdas today.

modules/packages/Futures.chpl
nspark: > The normal pattern for reference counting...

Good point about the release-acquire order here. (If it's a bug here, which I agree it is, then it's also a bug in the ZeroMQ module.)

> ...you'd have to write something like...

I have intentionally written tests like that. I think of the declaration-without-initialization pattern as something like an "incomplete future" (I need a better term here). In your example, calling `future.get()` (before the `async()` call) would block forever.

I can add some runtime logic that notes whether a future is incomplete the sense here and calls `halt()` if `get()` is called on an incomplete future. (I'd update this to use Chapel's future error throwing/handling when that's available.) I think the C++ analog to "incomplete future" as I'm using it here is whether the future refers to any shared state, which can be checked via [`std::future::valid()`](http://en.cppreference.com/w/cpp/thread/future/valid).

> IIRC, the Rice group thought that you shouldn't be able to assign futures. [...]

One thing 2f1b857 (vs 1fa8ffe) was meant to change is to explicitly allow assignment of futures, particularly for the declaration-without-initialization pattern. Otherwise, one cannot create an array of (incomplete) futures and create acyclic dependencies (e.g., for a 2D wavefront propagation). If I add runtime checks for `get()` on a incomplete futures, I could add similar logic for assignment. I think this also relates to whether the future is a record-wrapped, reference-counted class or just a plain record.

modules/packages/Futures.chpl
nspark: I thought of this Chapel `Future` object as something closer to C++'s `std::shared_future` than `std::future`. I'm not crazy about the read-once behavior of `std::future`, even when the future is not shared between threads. In particular, I thought it seemed most intuitive for the future's state to be sharable between tasks and be readable for the lifetime of those references. Maybe that's overkill.

modules/packages/Futures.chpl
mppf: Yes, but isn't it the case that if you can do things like make the array of futures and fill it in later, you could make a *cyclic* dependency among the futures?

Either way I think it'd be sufficient for progress (ie merging this PR) to make assigning a future an error for now. But the code you had here also didn't actually move any fields from rhs to lhs...

modules/packages/RefCount.chpl
nspark: Is "IntrusiveRefCount" named as such to contrast with something more automatic, as in #5096?

nspark:
I think the `proc this()` suggestion would be a good example and is something I can add (along with a declaration-without-initialization example) as both a test and documentation.

modules/packages/Futures.chpl
mppf: I'm not really familiar with C++'s shared_future and future. Do you think we'd want to have both of these eventually? Certainly if `std::future` doesn't need to do reference counting it could be more performant... If that's the case, maybe we should name the class accordingly?

modules/packages/Futures.chpl
mppf: I take that last part back... I see now that lhs.acquire(value) assigns the field.  Perhaps that is sufficient to guarantee no deadlock cycles, but I'd be interested to know if you think that's the case. Something like a deadlock like this (where it probably has syntax errors):
```
a: future;
b: future;
a = async( lamda() { b.get(); } );
b = async( lambda() { a.get(); } );
```

Is it the case that such examples don't deadlock because the *class instance* itself is immutable, and we're just changing what class instance a or b point to?

nspark: I don't think you could create a cyclic dependence, although I need to prove it. I think that you just lose the "handle" to the shared state of the original future, but that its lifetime persists until it is consumed by a dependent future.

What I'm trying to solve by allowing assignment is to support programmatic construction of a dependence pattern, unlike [this Rice example](http://svn.code.sf.net/p/chapel/code/branches/collaborations/futures/test/futures/futures_diamond_dependency.chpl) where `f1` ... `f4` are all explicitly named.

nspark: > Is it the case that such examples don't deadlock because the _class instance_ itself is immutable, and we're just changing what class instance a or b point to?

I think so; that was my rationale.

mppf: Works for me, thanks for talking me through it.

modules/packages/Futures.chpl
nspark: I think Chapel has a much higher likelihood for code to be implicitly parallel than does C++. Assuming that, I think shared-by-default is helpful and more intuitive to users. If the future isn't shared between tasks, I wouldn't expect reference counting to incur too much overhead, since I doubt users would create many handles to the same future.

modules/packages/RefCount.chpl
mppf: Yes. I think RefCountBase is a reasonable name for your new class but I might rather not name the module RefCount. I'm expecting to name the record-wrapper adding general reference counting RefCounted or Shared and havn't decided on a module name yet, but I currently named it ReferenceCounting...

mppf:
@nspark - as far as I know, we're waiting for you to put the finishing touches on this PR. Once you do that, I plan to run testing & merge if it is clean.

nspark:
@mppf I think this PR is ready for testing now.

mppf:
Passed full local testing.

PR #5136, opened 2017-01-10
Spec changes specialize generic method

mppf:
This is a continuation of PR #5057.

*  Add a new test case for specific-instantiation 
*  Spec changes describing overload-specific method

reviewed by @vasslitvinov - thanks!

### DIFF
--- a/spec/Classes.tex
+++ b/spec/Classes.tex
@@ -228,10 +228,11 @@ secondary methods can be defined not only for classes, records, and
 unions, but also for any other type (e.g., integers, reals, strings).
 
 Secondary methods can be declared with a type expression instead of a
-type identifier. In particular, if the \sntx{type-binding} includes a
+type identifier. In particular, if the \sntx{type-binding} is a
 parenthesized expression, the compiler will evaluate that expression to
-find the receiver type for the method. See also
-\rsec{Declaring_Specific_Instantiations}.
+find the receiver type for the method. In that case, the method applies
+only to receivers of that type. See also
+\rsec{General_Functions_and_Special_Versions}.
 
 Method calls are described in \rsec{Class_Method_Calls}.
 
--- a/spec/Classes.tex
+++ b/spec/Classes.tex
@@ -214,16 +214,24 @@ this-intent:
 
 type-binding:
   identifier .
+  `(' expr `)' .
+
 \end{syntax}
 Methods defined within the lexical scope of a class, record, or union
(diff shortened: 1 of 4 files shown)


spec/Classes.tex
```diff
@@ -214,17 +214,25 @@ \subsection{Class Methods}
...
 Methods defined outside of such scopes are known as \emph{secondary
 methods} and must have a \sntx{type-binding} (otherwise, they would
 simply be standalone functions rather than methods).  Note that
 secondary methods can be defined not only for classes, records, and
 unions, but also for any other type (e.g., integers, reals, strings).
 
+Secondary methods can be declared with a type expression instead of a
+type identifier. In particular, if the \sntx{type-binding} includes a
+parenthesized expression, the compiler will evaluate that expression to
+find the receiver type for the method. See also
```
vasslitvinov: "includes" -> "is" a parenthesized expression

Before "See also", I would add a sentence like "If so, the method applies only to receivers of this type."

spec/Generics.tex
```diff
@@ -868,6 +868,81 @@ \section{User-Defined Compiler Diagnostics}
...
 
+\section{Declaring Specific Instantiations}
+\label{Declaring_Specific_Instantiations}
+\index{specific instantiations}
+\index{generic specialization}
+
+The Chapel language facility supports three mechanisms for creating generic
+functions along with concrete functions that work with specific types.
+One can use these mechanisms to create specialized versions of generic
+functions.
```
vasslitvinov: ... three mechanisms for *using* generic functions ...

I"d drop "that work with specific types".

"to create specialized versions" -- I think you mean "to choose between" generic and/or concrete functions. The last sentence above does not seem right to me as-is. Maybe blend it with the first sentence in the paragraph by using "choose between" alongside with "use" verbs somehow.

I suggest replacing, throughout this section, "specialization" with "instantiation" to be consistent with other parts of the spec.

mppf:
@vasslitvinov - thanks for your feedback. Do you think "specific instantiation" is a good term for this ? I think C++ would call it "template specialization".

mppf:
@vasslitvinov - I've adjusted based upon your feedback. Now the main thing I'm not quite happy with is the title for the new section in Generics.tex. Let me know if you have any suggestions there.

mppf:
@vasslitvinov - I think this is good to go now - do you agree?

vasslitvinov:
@mppf yes, sounds good. thank you!

PR #5139, opened 2017-01-10
Replace temporary array in BlockCycDist with tuple

benharsh:
This temporary array was resulting in possibly millions of array
creations in HPCC PTRANS. Given that it's an array of length ``rank``
whose size does not change, I think a tuple makes more sense here.

I also removed what appears to be some old debugging code.


benharsh:
Testing:
- [ ] full local
- [x] full gasnet, dist-cyclic, dist-replicated

vasslitvinov:
Good to go.

I'll just mention that these fields of LocBlockCyclicArr: blocksize, locsize, numblocks - can also benefit from array-to-tuple conversion.

PR #5144, opened 2017-01-11
Remove the old flag FLAG_OMIT_FROM_CONSTRUCTOR

lydia-duncan:
I ran across this flag when trying to understand type constructors.  It appears
to be checked against but not attached to anything in either the compiler or
the modules.  It also appears to have been around for a really long time.  I
suspect that the last set of it was removed some time after 2010 (the latest
time the flags were reshuffled), but I don't know when that was removed.  In
any case, we don't use it now, so we might as well remove it.

In addition, at Elliot's suggestion, I removed a reference to the flag from the
old constructor proposal.

- [x] complete full paratest, for a sanity check.  I couldn't find a use of it with
grepcomp, grepmod or greptests, so this should just pass


ronawho:
Just FYI: https://github.com/lydia-duncan/chapel/blob/76d7d6d93359c7b4bad866ec443989ef59f6168b/spec/proposals/constructors/Examples.tex#L531

lydia-duncan:
Oh, good point!  I guess it happened sooner than Tom was anticipating?

PR #5152, opened 2017-01-11
Adding API for Task ID comparison and conversion to string

srirajpaul:
API is required for comparing task IDs that are not integers,
since other types such as struct wont work with binary operators.
Also adding a 'to string' API to help with printing of task IDs.
Adding CHPL_TASK_ID_STRING_MAX_LEN to define the
maximum length of string buffer to use.

### DIFF
--- a/runtime/include/tasks/fifo/tasks-fifo.h
+++ b/runtime/include/tasks/fifo/tasks-fifo.h
@@ -51,7 +51,7 @@ void chpl_task_stdModulesInitialized(void);
 typedef uint64_t chpl_taskID_t;
 #define chpl_nullTaskID 0
 #ifndef CHPL_TASK_ID_STRING_MAX_LEN
-#define CHPL_TASK_ID_STRING_MAX_LEN 20
+#define CHPL_TASK_ID_STRING_MAX_LEN 21
 #endif
 
 //
--- a/runtime/include/tasks/fifo/tasks-fifo.h
+++ b/runtime/include/tasks/fifo/tasks-fifo.h
@@ -50,6 +50,9 @@ void chpl_task_stdModulesInitialized(void);
 //
 typedef uint64_t chpl_taskID_t;
 #define chpl_nullTaskID 0
+#ifndef CHPL_TASK_ID_STRING_MAX_LEN
+#define CHPL_TASK_ID_STRING_MAX_LEN 20
+#endif
 
 //
 // Task layer private area argument bundle header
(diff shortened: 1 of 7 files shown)


runtime/include/tasks/fifo/tasks-fifo.h
```diff
@@ -50,6 +50,9 @@ void chpl_task_stdModulesInitialized(void);
 //
 typedef uint64_t chpl_taskID_t;
 #define chpl_nullTaskID 0
+#ifndef CHPL_TASK_ID_STRING_MAX_LEN
+#define CHPL_TASK_ID_STRING_MAX_LEN 20
```
gbtitus: Should be 21, including trailing '\0'.

runtime/include/tasks/massivethreads/tasks-massivethreads.h
```diff
@@ -45,6 +45,9 @@ typedef struct {
 // between C code and Chapel code in the runtime.
 typedef intptr_t chpl_taskID_t;
 #define chpl_nullTaskID 0
+#ifndef CHPL_TASK_ID_STRING_MAX_LEN
+#define CHPL_TASK_ID_STRING_MAX_LEN 20
```
gbtitus: Probably 11 (8 hex digits plus leading "0x" and a trailing '\0')?

runtime/include/tasks/qthreads/tasks-qthreads.h
```diff
@@ -51,6 +51,9 @@ void chpl_task_yield(void);
 //
 typedef unsigned int chpl_taskID_t;
 #define chpl_nullTaskID QTHREAD_NULL_TASK_ID
+#ifndef CHPL_TASK_ID_STRING_MAX_LEN
+#define CHPL_TASK_ID_STRING_MAX_LEN 20
```
gbtitus: For safety probably should be 21, appropriate for a 64-bit uint plus a trailing '\0'.  (There are C compilers, typically HPC-oriented, that can make the size of a default int 64 bits.)

gbtitus:
This looks good other than the specific line comments I made.  Those should be addressed, but once they are this can be merged.

bradcray:
@srirajpaul: In addition to the ICLA you've signed, I suspect we need you to be added to Rice's CCLA, assuming you're working on this as part of school-funded work.

runtime/include/tasks/fifo/tasks-fifo.h
srirajpaul: Done

runtime/include/tasks/massivethreads/tasks-massivethreads.h
srirajpaul: Making it to 21 since the format string used is %PRIiPTR which prints it like normal integer.

runtime/include/tasks/qthreads/tasks-qthreads.h
srirajpaul: done

gbtitus:
The PR now looks fine to me technically.  We're waiting on the CCLA, as I understand it.

PR #5153, opened 2017-01-11
Change the type of the receiver in type methods from _any to the defining type

daviditen:
Change the type of the receiver for type methods from _any to the type of the
class/record/union that defined the method.  This prevents calling type
methods on types other than the one that defined them and allows overloading
type methods.

Removed two .futures.

### DIFF
--- a/compiler/passes/cleanup.cpp
+++ b/compiler/passes/cleanup.cpp
@@ -206,6 +206,20 @@ static void add_parens_to_deinit_fns(FnSymbol* fn) {
 }
 
 
+/* Change the type of the receiver in type methods from "_any" to
+   the type that defined the method.
+*/
+static void set_receiver_type_in_type_methods(FnSymbol* fn) {
+  if (fn->hasFlag(FLAG_METHOD) && fn->thisTag == INTENT_TYPE) {
+    if (TypeSymbol* ts = toTypeSymbol(fn->defPoint->parentSymbol)) {
+      if (fn->_this->type == dtAny) {
+        fn->_this->type = ts->type;
+      }
+    }
+  }
+}
+
+
 void cleanup(void) {
   std::vector<BaseAST*> asts;
   collect_asts(rootModule, asts);
@@ -227,6 +241,7 @@ void cleanup(void) {
         destructureTupleAssignment(call);
     } else if (DefExpr* def = toDefExpr(ast1)) {
       if (FnSymbol* fn = toFnSymbol(def->sym)) {
+        set_receiver_type_in_type_methods(fn);
         flatten_primary_methods(fn);
         change_cast_in_where(fn);
         add_parens_to_deinit_fns(fn);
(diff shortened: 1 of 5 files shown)


compiler/passes/cleanup.cpp
```diff
@@ -206,6 +206,20 @@ static void add_parens_to_deinit_fns(FnSymbol* fn) {
...
 
+/* Change the type of the receiver in type methods from "_any" to
+   the type that defined the method.
+*/
+static void set_receiver_type_in_type_methods(FnSymbol* fn) {
+  if (fn->hasFlag(FLAG_METHOD) && fn->thisTag == INTENT_TYPE) {
+    if (TypeSymbol* ts = toTypeSymbol(fn->defPoint->parentSymbol)) {
+      if (fn->_this->type == dtAny) {
+        fn->_this->type = ts->type;
+      }
```
lydia-duncan: Does this code assume the type method is a primary method?  What happens if the type method is a secondary method?

lydia-duncan: As Mike and Michael will attest, I've been on a tear about secondary methods recently.  If it doesn't work, feel free to just add a future/issue for the secondary method case and move on.

lydia-duncan:
Looks good to me, other than the comment.  Thanks!

compiler/passes/cleanup.cpp
daviditen: Secondary methods fall through here, and basically the same transformation happens to them in normalization, though it wasn't super obvious exactly where in normalization it happens.

lydia-duncan: Cool!  Can you add a comment to that effect?

daviditen: Will do.  Thank you!

noakesmichael:
I also worried about secondary methods and so I did a test on that case
and determined that they work fine on master.

Currently secondary methods have a typeExpr field inserted in the
ArgSymbol for fn->_this and, as David indicates, this is used to fix-up
the type of “this” somewhere in normalize().

Longer term it seems like a bug to me that primary methods and
second methods rely on different mechanisms to determine the
type of “this”.  I mean this in the general case and not merely for
type methods.

Mike




On Jan 11, 2017, at 2:24 PM, David Iten <notifications@github.com<mailto:notifications@github.com>> wrote:


@daviditen commented on this pull request.

In compiler/passes/cleanup.cpp<https://github.com/chapel-lang/chapel/pull/5153>:

> @@ -206,6 +206,20 @@ static void add_parens_to_deinit_fns(FnSymbol* fn) {
 }


+/* Change the type of the receiver in type methods from "_any" to
+   the type that defined the method.
+*/
+static void set_receiver_type_in_type_methods(FnSymbol* fn) {
+  if (fn->hasFlag(FLAG_METHOD) && fn->thisTag == INTENT_TYPE) {
+    if (TypeSymbol* ts = toTypeSymbol(fn->defPoint->parentSymbol)) {
+      if (fn->_this->type == dtAny) {
+        fn->_this->type = ts->type;
+      }


Secondary methods fall through here, and basically the same transformation happens to them in normalization, though it wasn't super obvious exactly where in normalization it happens.

PR #5155, opened 2017-01-12
Trivial cleanup for flattenClasses

noakesmichael:
I happened to glance at flattenClasses.cpp and noted a trivial use of Vec;
   1) Collect a subset of TypeSymbols in to a Vec
   2) Process the Vec

I thought I would spend 60 seconds converting the Vec to a std::vector but 30 seconds
in to the effort I realized that the real "problem" is that there's no need to run two loops
at all.

Reduced to a single loop

Full paratest with --verify on linux64-single-locale and linux64-gasnet

### DIFF
--- a/compiler/passes/flattenClasses.cpp
+++ b/compiler/passes/flattenClasses.cpp
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2017 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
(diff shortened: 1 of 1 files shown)


benharsh:
Looks good to me 👍

noakesmichael:
Thanks Ben.

Mike




On Jan 11, 2017, at 5:51 PM, Ben Harshbarger <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks good to me 👍

PR #5156, opened 2017-01-12
Revised solution to PR 5153

noakesmichael:
Parsing a primary type method corrupted the type of the "this" ArgSymbol.
This resulted in a resolution failure for certain simple programs.  The first
solution for this error added code to the end of parsing to undo the
corruption.

The first commit reverts the changes to cleanup and fixes the root cause.

The second commit does some "whitespace" cleanup to the neighboring
code in type.cpp

Full paratest with --verify on linux64

### DIFF
--- a/compiler/AST/type.cpp
+++ b/compiler/AST/type.cpp
@@ -554,37 +554,58 @@ AggregateType::copyInner(SymbolMap* map) {
 }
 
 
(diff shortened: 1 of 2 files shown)


daviditen:
Thanks for the improvement.  This looks good.

PR #5157, opened 2017-01-12
Refactor rectangular I/O into a helper routine

bradcray:
This has been a longstanding change that I've carried around in
the various array view branches and which has been one of the
hardest to maintain.  Here, I'm pulling it out into its own PR
to see if I can get it onto master sooner.

The idea here is that a lot of printing of rectangular arrays
is reasonably independent of whether it's a default rectangular
array, a slice, a reindexed array, etc.  So to avoid repeating
that logic for each domain map, create a helper routine that
takes in the input/output stream, the array, and the domain
representing what should be printed and then just do what we
normally do.


mppf:
Looks good to me.

PR #5131, opened 2017-01-09
Use comm primitive in ISx study version

benharsh:
Optimize ISx communication by using a chpl_comm_array_put primitive instead of sliced assignment.

PR #5159, opened 2017-01-12
Perf annotations & label corrections

ben-albrecht:
## Annotations

* Replace temporary array in BlockCycDist with tuple (#5139)
* Move the multi-ddata metadata fields, checking cache effects for 'flat'. (#5145)
* Use comm primitive in ISx study version (#5131)

## Corrections

- Add ylabel `Time (seconds)` for NPBs
- Correct Spiral FFT ylabel to time instead of MFLOP/s

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -216,6 +216,8 @@ all:
       config: 16 node XC
   12/22/16:
     - multi-ddata (#5047)
+  01/11/17:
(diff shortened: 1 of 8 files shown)


benharsh:
👍

bradcray:
@ben-albrecht : My impression was that fft was actually measuring microseconds and was suggesting the label should be "Time (ms)"... did you find otherwise?

ben-albrecht:
@bradcray -- thanks for double checking. I did not notice that originally. It should actually be "Time (us)" for microseconds. I'll correct that.

PR #5160, opened 2017-01-12
epilogue label interface

psahabu:
This PR cleans up the handling of function epilogues in the Chapel AST. Function epilogues start with an epilogue label and end with a return. A normalized function contains gotos to the epilogue label instead of returns, so that cleanup code can be added to the epilogue.

In the process of lowering error handling, we discovered several problems with the way function epilogues were handled currently in the compiler. These issues are important for lowering error handling because lowering error handling adds control flow, unlike most other passes.

### What's wrong with function epilogues now?

- `FnSymbol::insertBeforeReturn` either inserted before the return or before the immediately preceding label. Unfortunately that means that `insertBeforeReturn` does not work correctly if the epilogue has more than one statement.

- `FnSymbol::insertBeforeReturnAfterLabel` does not create an epilogue label, so there's no guarantee that the code will actually be after an epilogue label if one is added later. Also, using this function could potentially break `insertBeforeReturn` as described above.

- There was no way to get an `FnSymbol`'s epilogue label, which is needed to implement error handling.

- There was also no way to distinguish an epilogue label from an ordinary label.

### Inserting around the epilogue label

- Two new functions have replaced the two broken ones:
  - `insertBeforeReturn` --> `insertBeforeEpilogue`
  - `insertBeforeReturnAfterLabel` --> `insertIntoEpilogue`.

- `insertBeforeEpilogue` searches for the epilogue label and will insert before it, but if one does not exist it will just insert before the return. If an epilogue label is created later it will be placed after the inserted node, so the inserted node will still be before the epilogue label.

- `insertIntoEpilogue` appends to the end of the epilogue block, just before the return. If will create an epilogue label if one does not exist, ensuring correct behavior.

### Getting the epilogue label

- `FLAG_EPILOGUE_LABEL` is now added to epilogue labels.

- `getEpilogueLabel` returns the `FnSymbol`'s epilogue label, or `NULL` if it doesn't exist.
The current implementation linearly scans the function from the end to find a label with `FLAG_EPILOGUE_LABEL`. This is unsatisfying, but creating an epilogue label field within `FnSymbol` has complications when the body of a function is replaced.

- `getOrCreateEpilogueLabel` returns the `FnSymbol`'s epilogue label, and creates one if it doesn't exist, marking it with `FLAG_EPILOGUE_LABEL`.

### Epilogue label protection and removal

- Epilogue labels are specially protected from deletion in `removeUnnecessaryGotos`. That function now includes a boolean option to force the deletion of unused epilogue labels. Otherwise, code that was put into the epilogue might not stay in the epilogue if control flow is added.

- Unused epilogue labels are removed in `denormalize` with a call to `removeUnnecessaryGotos` with the boolean option described above set. This should result in exactly the same emitted C code as before this PR.

- `inlineFunctions` removes `FLAG_EPILOGUE_LABEL` from any label it inlines, ensuring there is only one such label per `FnSymbol`.

### Notes

- needed for Error propagation
- cleaned up trailing whitespace
- passes linux64 + flat testing


vasslitvinov:
Just reading the commit message, linear search for the epilogue label makes me concerned about performance. I am open to merging this PR with linear search as-is then seeing the effect on compilation time. I have some suggestions on how to deal with issues arising when storing the epilogue label in FnSymbol.

Separately, should each function have an epilogue label *always* ? Perhaps created in the parser?

Also, do we expect cleanup code to be needed only at the function level? I can imagine that each scope aka BlockStmt potentially needing to do cleanup when exiting from the middle of it.

mppf:
@vasslitvinov - a few things in response to your questions.

First, I think it's acceptable now to do a linear search for the epilogue label because this only comes up in a relatively small amount of code that appends to the end of functions (instead of, say, appending to the function's block or replacing it's block with another block). Also, if the linear search is an issue, one way to solve it would be to make every function have an epilogue label (instead of just those with multiple returns). That would limit the linear search to just the epilogue which I think is almost certainly OK. I can think of other ways to make the search non-linear but am uncertain if they are worth the added complexity.

If we decided to make each function have an epilogue label always, the place to do that would be in `normalize` because having epilogue labels is something that normalized code already does. It's just that normalize includes the "optimization" that it doesn't add epilogue labels if there is only one return path.

Lastly, for the purposes of the current compiler and the current effort to implement error handling, we only need per-function "epilogues". We can add other cleanup before the `goto epilogue-label` if needed. I don't think enabling every BlockStmt to have an epilogue would be a real improvement right now. It seems speculative and something that would impact a lot of the compiler.

The review says "changes requested". Are these answers to your questions sufficient? If not, what in particular are you hoping we fix?

Thanks for the review!

noakesmichael:
I could not have said this better, or nearly as well for that matter.
I fully support Michael’s point of view on this aspect of the PR.



On Jan 19, 2017, at 5:29 AM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@vasslitvinov<https://github.com/vasslitvinov> - a few things in response to your questions.

First, I think it's acceptable now to do a linear search for the epilogue label because this only comes up in a relatively small amount of code that appends to the end of functions (instead of, say, appending to the function's block or replacing it's block with another block). Also, if the linear search is an issue, one way to solve it would be to make every function have an epilogue label (instead of just those with multiple returns). That would limit the linear search to just the epilogue which I think is almost certainly OK. I can think of other ways to make the search non-linear but am uncertain if they are worth the added complexity.

If we decided to make each function have an epilogue label always, the place to do that would be in normalize because having epilogue labels is something that normalized code already does. It's just that normalize includes the "optimization" that it doesn't add epilogue labels if there is only one return path.

Lastly, for the purposes of the current compiler and the current effort to implement error handling, we only need per-function "epilogues". We can add other cleanup before the goto epilogue-label if needed. I don't think enabling every BlockStmt to have an epilogue would be a real improvement right now. It seems speculative and something that would impact a lot of the compiler.

The review says "changes requested". Are these answers to your questions sufficient? If not, what in particular are you hoping we fix?

Thanks for the review!

vasslitvinov:
@mppf Yes, I would like to either eliminate linear search as I discussed with Preston, or as you suggest require every function to have the epilogue label by eliminating the optimization in normalize.

psahabu:
In the interest of getting this PR in for error handling, I will merge it now and observe the compiler performance impacts. If the impact is large, I will try to remove the epilogue label optimization in `normalize`.

The eventual plan for error handling is to have a cleanup construct `defer` which runs when a scope is exited. In that sense it would be useful to have epilogues on every `BlockStmt`, but that's not a pressing issue at the moment.

mppf:
@psahabu - I just noticed that this PR should have removed declarations for insertBeforeReturn insertBeforeReturnAfterLabel from symbol.h but did not. Could you have a look? Thanks.

psahabu:
Shoot, that must've happened while I was rebasing. I'll clear that out.

PR #5162, opened 2017-01-13
move Error to SysError; whitespace

psahabu:
In anticipation of the coming `Error` class, the module `Error` had to be moved to `SysError`. This is in line with the current naming scheme of `SysBasic` and `Sys`.

- while there, fixed trailing whitespace
- passes linux64+flat testing

### DIFF
--- a/modules/packages/HDFS.chpl
+++ b/modules/packages/HDFS.chpl
@@ -117,7 +117,7 @@ HDFS Support Types and Functions
  */
 module HDFS {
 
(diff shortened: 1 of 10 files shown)


mppf:
Looks OK to me. I'm worried that we'll run into problems with the compilerWarning, but let's try to keep it working if it's not too hard.

bradcray:
@psahabu: I'm noting that I think with this change, SysError doesn't seem to be included in the online documentation anymore.

PR #5164, opened 2017-01-13
Enable records without copy-initalizers

mppf:
Types that cannot be copied are important for implementing an equivalent to C++'s std::unique_ptr (see issue #5096). However such support is not as trivial to add as it seems.

This PR works around the issue that the compiler adds and resolves calls to initCopy before these calls might be eliminated.  That way, types that cannot be copied should still work in case the initCopy calls are removed after resolution. In particular, callDestructors removes extra initCopy calls for certain common patterns such as capturing the returned value from a function into a variable.

For example, the following code snippet adds an `initCopy` for `R` that is eliminated in callDestructors:
```
record R { }
proc returnR() { return new R(); }
var x = returnR();  // normalize adds initCopy here, but it's later removed
```

In more detail, here are the steps this PR takes:
 * adjusted generic instantiation to call fixupDefaultInitCopy when instantiating a compiler-generated initCopy function. Adjusted fixupTupleFunctions to return a bool so that tuple init copy can be only handled in fixupTupleFunctions and not in fixupDefaultInitCopy.
 * implemented fixupDefaultInitCopy to explicitly handle the creation of the initCopy function for types with initializers defined.  In particular, if a copy-initializer is not available, create a "dummy" initCopy marked with the flag FLAG_ERRONEOUS_INITCOPY. Otherwise, create an initCopy that runs the copy-initializer.
 * at the end of callDestructors, add code to check that no calls to initCopy functions marked with FLAG_ERRONEOUS_INITCOPY remain. If any do, raise a user-facing compilation error.
 * Adds a `checkonly` argument to resolveUninsertedCall. This was useful in earlier versions of this PR and has a straightforward implementation. It is not strictly necessary.

In order to get all of that to work, I also needed to adjust buildDefaultFunctions to avoid generating a compiler-default initCopy for a record that defines an initializer.

While there (and because an earlier version of this patch needed to do so), refactored the direct uses of autoCopyMap outside of functionResolution to instead call getAutoCopy, so that autoCopyMap can be private to resolution. At the same time, made autoCopyMap into a std::set instead of a Vec-that-is-a-set.

Lastly, this PR includes new tests and improvements to old tests to do with initializers.

Passed full local testing.

A longer-term-solution:
This PR adds dummy initCopy calls. A better long-term solution would be for the compiler to be structured in such a way that calls to a function (such as a copy initializer) can be added and resolved after the main body of resolution is completed. That would remove the need for autoCopyMap. Instead, code needing to add a copy would simply add a call to the copy initializer and resolve it.


vasslitvinov:
I am OK with the workaround of creating the "erroneous" initCopies then removing them. However I would like to see in the PR message some thoughts on a potential long-term solution.

Also, I would like to see in the PR message examples of situations when calls to initCopy are inserted and resolved then removed later, which makes us create dummy initCopies.

compiler/resolution/functionResolution.cpp
```diff
@@ -469,15 +471,19 @@ resolveUninsertedCall(BlockStmt* insideBlock,
   INT_ASSERT(block->parentSymbol);
 
   block->insertAtHead(call);
-  resolveCall(call);
+  if (checkonly && !call->primitive) {
+    resolveNormalCall(call, checkonly);
+  } else {
+    resolveCall(call);
```
vasslitvinov: Should we add INT_ASSERT(!checkonly) in this case, i.e. disallow calling this function on a primitive call with checkonly==true ?

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+      // If the user has defined any initializer,
+      // initCopy function should call the copy-initializer.
+      //
+      // If no copy-initializer exists, we should generate a call
+      // to a dummy init-copy function that generates an error
+      // if it remains in the AST after callDestructors
+      // (since callDestructors can remove some initCopy calls,
+      //  we'd like types that cannot be copied to survive
+      //  compilation until callDestructors has a chance to
+      //  remove those calls)).
```
vasslitvinov: Let's turn this parenthetical remark into a separate sentence. This will fix the double-close-parens too.

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+
+  if (AggregateType* ct = toAggregateType(arg->type)) {
+    if (isUserDefinedRecord(ct) &&
+        ct->initializerStyle == DEFINES_INITIALIZER) {
+      // If the user has defined any initializer,
+      // initCopy function should call the copy-initializer.
+      //
+      // If no copy-initializer exists, we should generate a call
+      // to a dummy init-copy function that generates an error
+      // if it remains in the AST after callDestructors
```
vasslitvinov: Let's edit the above sentence to remove "generate a call", if I get this right:

"If no copy-initializer exists, we should make this initCopy a dummy function that generates an error [then as before]"

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+              INT_ASSERT(se);
+              if (se->symbol() == retSym)
+                stmt->remove();
+            }
+        }
+
+        // Set the RVV to the copy
+        newFn->insertBeforeReturn(new CallExpr(PRIM_MOVE, retSym, memeTmp));
+
+        gdbShouldBreakHere();
```
vasslitvinov: Please remove gdbShouldBreakHere() and empty line.

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+      // to a dummy init-copy function that generates an error
+      // if it remains in the AST after callDestructors
+      // (since callDestructors can remove some initCopy calls,
+      //  we'd like types that cannot be copied to survive
+      //  compilation until callDestructors has a chance to
+      //  remove those calls)).
+
+      // Go ahead and instantiate the body now so we can fix
+      // it up completely...
+      instantiateBody(newFn);
```
vasslitvinov: I don't see that instantiateBody(newFn) is invoked in the situation when this function returns false. Are you sure it still works in that case?

As far as I can tell, the boolean result of fixupDefaultInitCopy is unused. Should it have void return type instead?

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+        // Remove other PRIM_MOVEs to the RVV
+        for_alist(stmt, newFn->body->body) {
+          if (CallExpr* callStmt = toCallExpr(stmt))
+            if (callStmt->isPrimitive(PRIM_MOVE)) {
+              SymExpr* se = toSymExpr(callStmt->get(1));
+              INT_ASSERT(se);
+              if (se->symbol() == retSym)
+                stmt->remove();
+            }
+        }
```
vasslitvinov: Assuming that 'newFn' is always compiler-generated, the above loop is probably OK. I would still suggest replacing it with for_SymbolSymExprs(se, retSym) { if (CallExpr* callStmt = toCallExpr(se->parentExpr)) ... }

compiler/include/resolution.h
```diff
@@ -29,7 +29,9 @@ class CallInfo;
 
 extern SymbolMap paramMap;
 extern Vec<CallExpr*> callStack;
-extern Map<Type*,FnSymbol*> autoCopyMap; // type to chpl__autoCopy function
+bool hasAutoCopyForType(Type* type);
```
vasslitvinov: If we replace hasAutoCopyForType() with getAutoCopyOrNullForType(), which returns straight autoCopyMap[type], we can replace two calls hasAutoCopyForType+getAutoCopyForType with just a call to getAutoCopyOrNullForType.

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+  if (!hasAutoCopyForType(type)) {
+
+    const char* copyFnToCall = "chpl__autoCopy";
+
+    if (isUserDefinedRecord(type) &&
+        !type->symbol->hasFlag(FLAG_TUPLE) &&
+        !isRecordWrappedType(type) &&
+        !isSyncType(type) &
+        !isSingleType(type) &&
+        !type->symbol->hasFlag(FLAG_EXTERN)) {
```
vasslitvinov: Sounds like a case to introduce a function isUserDefinedRecordReallyIMeanIt :)

compiler/resolution/functionResolution.cpp
```diff
@@ -588,20 +594,128 @@ hasUserAssign(Type* type) {
...
+      // Just use 'chpl__initCopy' instead of 'chpl__autoCopy'
+      // for user-defined records. This way, if the type does not
+      // support copying, the autoCopyMap will store a function
+      // marked with FLAG_ERRONEOUS_INITCOPY. Additionally, user-defined
+      // records shouldn't be defining chpl__initCopy or chpl__autoCopy
+      // and certainly shouldn't rely on the differences between the two.
```
vasslitvinov: The discussion of FLAG_ERRONEOUS_INITCOPY feels like a distraction here.

Is the motivation simply that we want user records to have nothing to do with autoCopy() ?

compiler/resolution/functionResolution.cpp
mppf: yeah. We should probably rename 'isUserDefinedRecord' to 'isNormalRecord' (or something) and then create 'isUserRecord' (or something) for isUserDefinedRecordReallyIMeanIt. I just don't have the appetite to do that now.

compiler/resolution/functionResolution.cpp
mppf: By removing the default auto-copy (that just calls init-copy), we enable the FLAG_ERRONEOUS_INIT_COPY strategy to work if an auto-copy is added later for a type that does not support copying. By putting something there (instead of NULL), we can let code used to working with autoCopyMap still function, but also get an appropriate error.

*at the same time* it's not appropriate for user records to have both autoCopy and initCopy. They just have copy initializers and the compiler currently uses autoCopy and initCopy as ways to call the copy initializer. I don't think user records should have either autoCopy or initCopy. autoCopy/initCopy are merely implementation concerns.

compiler/resolution/functionResolution.cpp
mppf: the boolean result is unused but the way it called in generic instantiation might depend on the boolean in the future. In particular, I just added such a boolean for fixupTupleFunctions.

The call to instantiateBody will happen later if this function returns false. It is added explicitly here so that the function is instantiated enough that we can "fix" it. Otherwise, it won't have a body with types resolved, etc.

compiler/resolution/functionResolution.cpp
mppf: newFn is always compiler generated and in fact only consist of this body:

  return arg;

I think it's unnecessarily complex to use for_SymbolSymExprs here. So, I'd rather not make that change, since we'd have to test again. Unless you want to argue more for it.

compiler/include/resolution.h
mppf: Or we could just put autoCopyMap to be an exported symbol again.

The point of these changes was to simplify the common case for getting the auto copy function from the map. In the common case, if it's NULL, it should be an error. So getAutoCopyForType fails with an error if it's about to return NULL.

As a separate matter, adding these "accessor functions" enables future modifications to do with the auto-copy map.

I think it's clearer to use hasAutoCopyForType to get the uncommon case - where not having an autoCopy is tolerable - to work. (And as an aside, I think tolerating a NULL auto-copy-map entry is a workaround anyway).

compiler/resolution/functionResolution.cpp
lydia-duncan: I don't see any reason why we couldn't have checkonly apply to primitives, it just seemed like an unnecessary amount of work to thread checkonly through all the other functions called by resolveCall, so we didn't do it.

Is your point that a caller of resolveUninsertedCall might miss that checkonly was being ignored in the case of primitives?  I think that is a fair concern, and so would be open to an INT_ASSERT with a comment about it being extensible or an INT_FATAL with that message.  @mppf - I suspect I don't have permission to push to your branch, how do you want to handle this?

compiler/include/resolution.h
lydia-duncan: I agree with Michael.  The NULL indicator is poor coding practice in my mind on both our homegrown Map and the C++ map class.  The boolean indicator is a much more principled strategy and one that is used in higher-level languages.

compiler/resolution/functionResolution.cpp
lydia-duncan: Commit with this change is here: https://github.com/lydia-duncan/chapel/commit/d1f1a46948765d13f3670cbe3a944bea575703f4

mppf: FYI I've cherry-picked that change into this PR and marked the PR as needing testing. I'm not going to start testing until we're done the review though.

vasslitvinov:
OK, sounds good.

PR #5170, opened 2017-01-16
All shootouts use high prec timer

bradcray:
Every  once in awhile, we run into a surprise that is caused by some of the shootout directories using the high precision timer and others the default timer.  At the very least, they all ought to use the same timer mechanism, especially when comparing different versions of the same benchmark.  This change converts them all over to us the high precision timer for the following reasons:

* It seems to more accurately resemble what is used on the CLBG site itself
* It returns more accurate results
* It sometimes returns better results, particularly when a program ends with file I/O (we saw this with fasta, for example)

I've started to make this change several times, then never quite finish it or balk, then run into yet another case where the difference between the two timers is noticeable.  The latest is mandelbrot which sees a ~0.2 sec  (or ~4%) difference between the two timers.


bradcray:
For consideration by @chapel-lang/perf-team , particularly @ronawho.

ronawho:
👍

PR #5173, opened 2017-01-16
Sparse block domain clear should reset size

bradcray:
This commit adds a test submitted by Ariful Azad that demonstrates that clearing a sparse block domain did not reset its size.  It also adds a five-minute fix (TM) that seems to fix the issue.


bradcray:
@e-kayrakli and/or @mppf, could you review this?

This issue is making me try to remember whether we have a sparse domain robustness suite, similar to what we support for other distributions in which we can run a bunch of standard sparse computational idioms across various layouts and distributions.  I feel like I did something like this once for LayoutCSR to make sure it was as good as DefaultSparse, but have a feeling that I just hacked it up by hand for a one-time test rather than making it part of the test system...  Do either of you know otherwise?  Thanks.

mppf:
Looks good to me. I don't know the answer about sparse domain robustness tests.

bradcray:
Also, do either of you know which directories contain code that exercise the distributed sparse block domains/arrays?  I'd like to only test those directories with gasnet enabled, but they're hard to find since there's no specific "sparse block" module used...

mppf:
test/distributions/engin/SparseBlock/ test/distributions/bradc/spsBlk/

e-kayrakli:
Fix looks good.

Other than what Michael pointed out for tests: I have test/users/engin/sparse_bulk and test/users/engin/sparse_bulk_dist that specifically tests bulk addition to local and distributed sparse domains, respectively.

PR #4405, opened 2016-08-25
Skip stencilSerial and mandelbrot-unrolled under --fast and 32-bit configs

ronawho:
Due to minor floating point variations across configurations, stencilSerial and
mandelbrot-unrolled can get slightly different binary output. A reference c
mandelbrot version also fails with the equivalent of --fast, so it seems like
it's not a bug on our end but is just a precision difference. This just skips
these tests for 32 bit platforms and when --fast is thrown.

See JIRA 244 for more info: https://chapel.atlassian.net/browse/CHAPEL-244

PR #5175, opened 2017-01-16
Skip new mandelbrot version for linux32

bradcray:
This change follows the lead of PR #4405 which disabled some versions
of mandelbrot for linux32 and --fast.

Over the weekend, testing of my new mandelbrot version failed on
linux32 due to the absence of a single bit on the uppermost border
of the mandelbrot set, which can be seen by zooming into the
expected solution on the CLBG site:

  http://benchmarksgame.alioth.debian.org/download/mandelbrot200.png

It's a little odd that this bit is "on" given how far it is outside
of the mandelbrot set, and on linux32, it's computed as being "off".
But only when --denormalize is on (as it is by default, currently).

As noted on the associated JIRA issue
(https://chapel.atlassian.net/browse/CHAPEL-244), some of the
reference versions of mandelbrot also seem to fail when compiled
with optimizations on (and looking at the website, some of the
leading versions fail or don't seem to be run in the 32-bit
configuration, also seeming to underscore this point).

We didn't seem to get failures for cygwin32, so I made my .skipif more
specialized for now (and Elliot confirms that he made the previous
skipif skip 32-bit just as a precaution).  If the behavior is unstable over
time, it can be expanded as necessary.

### DIFF
--- a/test/studies/shootout/mandelbrot/bradc/mandelbrot-blc.skipif
+++ b/test/studies/shootout/mandelbrot/bradc/mandelbrot-blc.skipif
@@ -2,4 +2,6 @@
 #
 # See JIRA 244 for more info: https://chapel.atlassian.net/browse/CHAPEL-244
 
+# the issue with --fast occurs on chapcs with CHPL_TARGET_ARCH=native
+COMPOPTS <= --fast
 CHPL_TARGET_PLATFORM == linux32
--- a/test/studies/shootout/mandelbrot/bradc/mandelbrot-blc.skipif
+++ b/test/studies/shootout/mandelbrot/bradc/mandelbrot-blc.skipif
@@ -1,0 +1,5 @@
+# Floating point precision variations in some configs cause binary mismatch
+#
+# See JIRA 244 for more info: https://chapel.atlassian.net/browse/CHAPEL-244
+
+CHPL_TARGET_PLATFORM == linux32


bradcray:
@benharsh and @ronawho may be interested in looking at this one.  Ben, it worries me a little that turning off denormalization causes this to pass...  How deeply did you look into the behavior to determine that it's a floating point issue and not a denormalization issue?

benharsh:
I believe we came to that conclusion by observing that a C reference of mandelbrot would also fail on 32-bit (when compiled with the equivalent of --fast). I don't think we looked any deeper than that.

benharsh:
More specifically, I think it was march=native that was causing problems.

bradcray:
OK, thanks.  Even though I'd just read it, I forgot that you'd run reference versions (I'd only checked his website to see that some of them had failed or were missing).  I updated my comment to better reflect this.  You guys OK if I merge?

benharsh:
Fine by me!

ronawho:
Fine by me too. Note that the other tests that we skipped weren't failing under cygwin32 either, I just did a general 32 bit skipif because I didn't want to have to update the skipif if we added new 32 bit platforms or something.

bradcray:
Elliot pointed out that --fast on chapcs is a problem if CHPL_TARGET_ARCH is set to native (I'd set it to none).  Adding that to the skipif as well to be safe.

PR #5177, opened 2017-01-16
Improve chpl_task_idTostring() name, add tests.

gbtitus:
The runtime function chpl_task_idTostring() recently added by #5152
really should have been called chpl_task_idToString() (capitalizing the
'String' part).  And we should have at least a sanity tests for that new
function as well as chpl_task_idEquals() which was added in the same PR.

I'm making these changes myself because I really should have caught the
need for them in my review of #5152.


gbtitus:
@ben-albrecht  would you take a look at this one?  It barely needs a review, just a glance would do.

PR #4710, opened 2016-09-29
Chapel Docs Reorg

ben-albrecht:
As Chapel documentation continues to grow, this growth demands greater efficiencies in all areas of its operation. One of the ways we've stepped up to meet these demands, is by merging the Chapel documentation hierarchy into a single directory `doc/`.

Please be aware of these changes, as they may impact you upon merging.
- `doc/sphinx` is now merged with `doc/release`, which both now live under `doc/`
  - There is no longer a distinction between the paths for docs in the release tarball and paths for the git repository. The `gen_release` script strips out the irrelevant sphinx-configuration / index files from `doc/`
  - There is now less of a distinction between the paths for docs in the `doc/` directory and the paths on the website, [chapel.cray.com/docs/master](chapel.cray.com/docs/master)
  - We no longer symlink files between these directories
  - All references to `doc/sphinx` and `doc/release` have been updated to `doc/`
  - Any skeleton directories that contain primarily generated documentation, e.g. `doc/modules`, will live in `meta/` to be copied in upon build, e.g. `cp -f .modules/* modules/`
    - This is to avoid clutter for the user browsing documentation in from the command line
- `doc/sphinx/build` is now generated as `build/doc`
- scripts for building docs now live in `doc/util`
- Makefiles for building docs now live in `doc/`
- developer docs continue to live in `doc/developer` and can be easily included into the built html documentation, once formatted over to restructured text

View the new docs layout [here](https://github.com/ben-albrecht/chapel/tree/docs-reorg/doc)

_Testing_

-  [X] `make docs`
- [x] `doc/ make docs`
- [ ] `doc/ make man-chapel`
- [x] `gen_release` tarball
- [x] test on system that requires the venv for sphinx

PR #5183, opened 2017-01-18
Build html docs into /doc/html

ben-albrecht:
**Motivation:**

Build docs hierarchy to mirror that of the release version.

**What this PR does:**

- Adds `make html-release` target which simply builds the html docs and moves them from `/build/doc/html` to `/doc/html`.
- Switches `make docs` to call `make html-release` instead of `make html`
- Cleans up the `html/` directory in the `make clean-build` target
- Moves `run-in-venv.bash` from `doc/util` to `/util`
    - it is generally useful outside of docs, and is already being used elsewhere (see chplspell)

- Following the Chapel convention that `make clean`  cleans up intermediate source files: a `make clean` at both top-level and  doc-level will only remove any generated .rst files within `doc/`

- `make clobber` (top-level and doc-level) removes everything, both source and build files, that were generated from `make docs`

- Updates `gen_release` to not doc-level clobber after `make docs`

**How this change affects the release:**

The release will now include all rst files generated during `make docs` that make up the content of the html docs. This means the `doc/` directory in the release will now contain:

- the rst generated from `chpl2rst.py`: `examples`, `primers`
- the rst generated from `chpldoc`: `builtins`, `modules`

**How this change affects most users:**

When invoking `make docs`, the final step will print out:

```
Build finished. The HTML pages are in $CHPL_HOME/doc/html.
```

and you will find your docs built into that path instead of `$CHPL_HOME/build/doc/html`.

**Testing:**

* [x] Build docs on a linux box
* [x] Build tarball with `gen_release`


ben-albrecht:
@bradcray -- as the requester of said change, are you interested in reviewing?

bradcray:
Yes

mppf:
@ben-albrecht - can you verify that a top-level 'make clobber' will remove the built docs as well? I'm guessing (from looking at the diff) that a top-level 'make clean' will not - which is fine but you should probably say clearly what happens there in the PR message.

bradcray:
I think (agree?) that 'make clean' should not remove the built docs.  The general principle we've taken for 'clean' has been that it doesn't break the things you've built, it just cleans up things that aren't strictly necessary to use those things.

ben-albrecht:
@mppf - that is correct. I have updated the PR message to clarify this. Let me know if I left anything ambiguous.

I am making edits to `gen_release` now, after my testing...

An open question: should the release contain the generated rst files (e.g. chpldoc-generated module documentation) that are used to build the full html documentation? 

We currently remove them for the release, but we leave the static rst files with the intention of having a way for people to read the documentation with a text-editor. I'm not sure what our original reason for removing the generated rst files was -- maybe because they aren't hand-written and aren't as text-readable?

doc/Makefile.sphinx
```diff
@@ -61,23 +66,32 @@ check-sphinxbuild:
 	@echo "Confirming that sphinx-build is available..."
 	@if [ $(SPHINXTEST) -eq 1 ]; then echo $(SPHINXERROR); fi
 
-html: check-sphinxbuild source clean-build
+html-release: check-sphinxbuild source clean-build
 	@echo
 	@echo "Building HTML documentation"
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@chmod -R ugo+rX $(BUILDDIR)
+	mv $(BUILDDIR)/html .
```
bradcray: Should this rm -r html before doing the mv?  Or can we be guaranteed that something else has already done that?

doc/Makefile.sphinx
```diff
@@ -61,23 +66,32 @@ check-sphinxbuild:
...
-	@echo "Build finished. The HTML pages are in "'$(BUILDPATH)'"/html."
+	@echo "Build finished. The HTML pages are in "'$$CHPL_HOME'"/doc/html."
 
 html-dev:
 	@echo
 	@echo "Building HTML documentation with no prerequisites (dev testing)"
 	@echo "Warnings not treated as errors"
 	$(SPHINXBUILD) -b html -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SOURCEDIR) $(BUILDDIR)/html
 	@chmod -R ugo+rX $(BUILDDIR)
+	mv $(BUILDDIR)/html .
```
bradcray: Maybe "ditto" but maybe less important if this is for devs...

bradcray:
I'd forgotten about the TODOs, so was about to note that gen_release needs to be updated.

> I'm not sure what our original reason for removing the generated rst files was -- maybe because they aren't hand-written and aren't as text-readable?

I'm not sure either.  I don't think I was around when this was being hashed out.  If they aren't reasonably readable, I probably wouldn't include them (i.e., focus on human-curated files designed to be read as text).  But if they are reasonably readable, I don't know of a strong reason not to include them.

ben-albrecht:
As an example, here's what the release looks like with the clobber (current approach):

```
language   platforms     technotes
tools      users-guide   usingchapel
html
```

Without the clobber (proposed new approach):

```
builtins    language    primers     users-guide
examples    modules     technotes   usingchapel
html        platforms   tools
```

The `primers` and `examples` are very readable, although they are just a rst-ified version of the primers/hellos code. The `modules` and `builtins` contain a lot of rst markup syntax, which makes them a bit harder to read, but still readable. 

I would opt for including them just for completion, i.e. all the of html content can be found in rst form in the `doc/` directory.

doc/Makefile.sphinx
ben-albrecht: This target depends on `clean-build` which does `rm -rf html`

doc/Makefile.sphinx
ben-albrecht: This target does not depend on `clean-build` (intentionally, to save time on rebuilds). I think I need to add a manual `rm -rf html` here and change the `mv` to a `cp -r`.

bradcray:
It appears to me that the doc/ directory in the release tarball built by gen_release by this PR differs very significantly from the 1.14.0 release.  Is this intentional or a surprise?  Or am I doing something wrong (we don't do any steps to create the tarball after gen_release do we?)

1.14.0:

    README.md		debugging.rst		prereqs.rst
    bugs.rst		executing.rst		quickReference.pdf
    building.rst		html/			tasks.rst
    chapelLanguageSpec.pdf	launcher.rst		technotes/
    chplenv.rst		multilocale.rst
    compiling.rst		platforms/

this branch:

    builtins/	language/	platforms/	technotes/	users-guide/
    examples/	modules/	primers/	tools/		usingchapel/

ben-albrecht:
>  Is this intentional or a surprise?

Intentional. My [diff above](https://github.com/chapel-lang/chapel/pull/5183#issuecomment-273625775) is comparing current master's release to this branch's release. The current branch's release is different from the 1.14 release already, as a result of changes in #4710, specifically this bullet:

> There is no longer a distinction between the paths for docs in the release tarball and paths for the git repository. The gen_release script strips out the irrelevant sphinx-configuration / index files from doc/

bradcray:
Huh... I guess I hadn't realized that #4710 had made the doc/ directory look so different in the release -- I thought it was more about changing how things were built than reorganizing the release directory (and re-reading the PR doesn't make me feel like I missed much).  I'd assumed that any aspect that was unifying the directory structure was striving to make the GitHub directory structure more like the release structure rather than vice-versa.  Some things about the current layout seem a little weird to me like not having a top-level README, having two copies of the language specification, having html and rst documentation be neither symmetric nor sufficiently asymmetric that it's not clear where to start reading the rst documentation (this is related to the lack of the README).  I don't think these comments should hold up this PR, but it suggests to me that we need to do more doc/ organization work before the 1.15 release.

I don't find the sphinx-generated .rst to be particularly unreadable, so would currently err on the side of including it for symmetry; though I could also see choosing not to include it because it's quite a bit of documentation and it's hard to imagine someone preferring to read it from the command-line over the html files.

ben-albrecht:
>  I don't think these comments should hold up this PR

Alright, I am going to merge this as it is now, which includes all the rst files. We can address some of these other concerns in a future pre-1.15 code/doc freeze PR. Summarizing the concerns you brought up before, issues to address include:

- Adding a doc-level README back in
    -  Addressing where one should begin reading the rst docs
- Looking into ways of having only 1 copy of spec in the release
    - This can be achieved with a spec removal & symlink in `gen_release` script
- Ensuring symmetry between source rst docs and html docs
    - Not too much to do here, after this PR

bradcray:
My comment about symmetry between source rst docs and html docs may not have been clear.  It seems to me that in the release, we should either have:

  README.md
  rst/
  html/

or something that treats rst and html more symmetrically.  Put another way, html/ isn't a sibling of all the other directories in doc/ currently.

But the other thing I was saying was that if we believe that one of them is the "primary" documentation that we want to promote, then perhaps the lack of symmetry is OK as long as there is good explanation for it.  For example, historically, txt files (which became rst files) were the default, in which case it seemed reasonable to have .txt files that helped you get up and running at the top-level, along with key documents like the spec.  I'm not sure whether we still have a notion of "primary documentation" anymore, though...

In any event, I think this warrants more discussion.

Issue #5017, opened 2016-12-13
CHPL_UNWIND=libunwind testing


mppf:
CHPL_UNWIND=libunwind should be tested nightly.

Note that `util/test/prediff-for-stacktrace` is supposed to help with the various forms of mismatch between a .good file and a .chpl file, in case one has a stacktrace and the other does not.

Also note that stack traces might include call sites on linux64, but generally won't include call sites on Mac OS X. And also note that stack traces *do* currently include line numbers from internal modules.

mppf:
We should probably decide, as a matter of policy, if stack traces are expected in .good files, or if the prediff should handle the case that the .good file has no stack trace. If the stack traces are included in .good files, many more internal/standard module line numbers can be exposed, so we presumably need a way to handle that intelligently. It would not be an improvement to have to update every test that halts when certain internal module code is moved to different lines.

PR #5184, opened 2017-01-18
Disable CHPL_UNWIND during start_test runs

bradcray:
This commit disables CHPL_UNWIND during start_test runs, since the
output can be dramatically different (e.g., if an expected error
message is called, which calls halt, a stack trace is printed).

It's an open question as to how we'd do testing in a "CHPL_UNWIND on
by default" world, but in the meantime, I'm trying to work with
CHPL_UNWIND on by default to see how well it works.  In the meantime,
this mod seems like the nice thing to do for developers who use it.

In the conversation on the PR, Michael points out that there is a prediff
script that can be manually invoked to filter out CHPL_UNWIND output
and that there are tests that will be skipped if CHPL_UNWIND is unset
(which now it always will be).  But, we're currently never running with
CHPL_UNWIND set in our nightly testing, so these tests aren't being run
anyway and Michael reports that their .good files are stale anyway.
Meanwhile, Michael has opened issue #5017 to further discuss how
CHPL_UNWIND testing should ultimately work.

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -560,6 +560,9 @@ def check_environment():
     if "CHPL_DEVELOPER" in os.environ: # unset CHPL_DEVELOPER
         del os.environ["CHPL_DEVELOPER"]
 
+    if "CHPL_UNWIND" in os.environ:    # unset CHPL_UNWIND
+        del os.environ["CHPL_UNWIND"]
+
     # Check for $CHPL_HOME
     global home
     if "CHPL_HOME" in os.environ:


bradcray:
Attn: @mppf @panzone

mppf:
it's fine with me to make this change for now but we do have a prediff that is supposed to address the mismatches with unwind / no unwind testing. I had been viewing libunwind-testing as something that is awaiting development/testing rather than something we need to discuss/agree upon... although if I recall correctly we have a difference of opinion about whether programs that halt should have the stack traces in the .good files or not. Let's talk about that in the unwind-testing issue, #5017, where I tried to describe the issue in a comment.

bradcray:
Are you suggesting that the prediff you mention should fire automatically when CHPL_UNWIND is set (in which case something would be wrong in the prediff if I'm seeing a failure due to having CHPL_UNWIND being set when I run start_test) or that I should be specifying that the prediff should be used if I run start_test with CHPL_UNWIND set?  (i.e., user error).

mppf:
It's not run automatically, but I think it should be. You could use the prediff now manually and I'd hope that would work.

bradcray:
In the current world, what is the observable difference between unsetting it in start_test vs. using the prediff?  (other than O(1) vs. O(n) execution time?)

mppf:
I'm not sure what you meant by execution time. But the difference would be that if you use the prediff, the .good file *should* be able to have a stack trace in it. That trace would be ignored by the prediff if CHPL_UNWIND is off.

Either approach lets us test with CHPL_UNWIND on and .good files that don't have stack traces.

bradcray:
By execution time I meant that unsetting an environment variable is something that is done once and takes a short amount of time whereas prediff requires running post-processing scripts over every test's output (right?).  Do we have .good files that contain stack trace data at present?  If we do, wouldn't this PR break them? (since no stack trace would be generated)

mppf:
Yes, the tests in `test/runtime/panzone/` have stack traces in them. But, these are skip-if'd if CHPL_UNWIND is not set... and so they never run in nightly testing. They also have stale line numbers....

So, I don't think this PR causes a problem in that regard.

Just to be sure I'm clear: I'm not arguing against this PR as an initial step. It's trivial enough that if we get the prediff online we can just un-do it, and I think it helps for now.

bradcray:
OK, that's what I was missing...  It will cause some tests to always be skipped that aren't being run in nightly testing anyway, and which are currently don't pass even if they were run.  I'll merge for now for convenience then.

(I knew you weren't arguing against it, I was just trying to understand things well enough to see if I was still for it).

PR #5189, opened 2017-01-18
Enable inlining the multi-ddata array follower iterator.

gbtitus:
The follower iterator for the multi-ddata case could not be inlined
because it had two separate iteration loops with an execution-time test
to choose which one to use.  I believe that the test actually checked a
condition that is always true, so here I'm removing it in order to get
the iterator inlined.  I am leaving behind a debug-protected assert()
test to check the condition, however.

For a number of tests, such as stream-ep, this gets performance back to
the expected range after the original multi-ddata slowed them way down.
It does not fix the largest losses, though, such as for '1D Array
Parallel Iteration'.  That has a different cause and will be treated
separately.

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -920,7 +920,7 @@ module DefaultRectangular {
                minIndicesPerTask:int = dataParMinGranularity)
       ref where defRectSimpleDData {
       if debugDefaultDist {
-        chpl_debug_writeln("*** In defRectArr simple-dd fallback iterator");
+        chpl_debug_writeln("*** In defRectArr simple-dd serial iterator");
       }
       if rank == 1 {
         // This is specialized to avoid overheads of calling dsiAccess()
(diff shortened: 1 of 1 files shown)


gbtitus:
@benharsh , would you review this one, please?

modules/internal/DefaultRectangular.chpl
```diff
@@ -1170,64 +1170,53 @@ module DefaultRectangular {
...
         const mdPDLow = dom.dsiDim(mdParDim).low;
         const chunk = mdInd2Chunk(mdPDLow + followThis(mdParDim).low);
-        if mdPDLow + followThis(mdParDim).high <= mData(chunk).pdr.high {
-          //
-          // entire followThis is in one multi-data chunk
-          // TODO: is it ever otherwise?
-          //
+        if debugDataParMultiDData {
+          // this code assumes followThis spans but a single chunk
+          assert(mdPDLow + followThis(mdParDim).high <= mData(chunk).pdr.high);
```
benharsh: Would something like this be better protected by the ``boundsChecking`` param so that we can have frequent testing of this condition?

benharsh:
Looks good otherwise! 👍

PR #5193, opened 2017-01-19
Create same number of mem descs w/ and w/o --devel

vasslitvinov:
MOTIVATION
----------

I found the number of AST nodes created with and without --devel
to differ. Here is what I believe was the cause.

callChplHereAlloc() invoked newMemDesc() passing it the name of
the type being allocated. newMemDesc() created a new memory descriptor
for each new name that it got. That involved new_IntSymbol() and so
a new VarSymbol + new DefExpr.

When compiling with --no-devel, type names for domains/arrays are
made nicer, e.g.:

  [CSRDom(2,int(64),domain(2,int(64),false))] int(64)

Cf. with --devel they are more verbose, e.g.:

  BaseSparseArr(int(64),2,int(64),CSRDom(2,int(64),_domain(DefaultRectangularDom(2,int(64),false))))
  BaseSparseArrImpl(int(64),2,int(64),CSRDom(2,int(64),_domain(DefaultRectangularDom(2,int(64),false))))
  CSRArr(int(64),2,int(64),CSRDom(2,int(64),_domain(DefaultRectangularDom(2,int(64),false))))

BTW they come from the following locations, respectively:

  ChapelDistribution.chpl:689
  ChapelDistribution.chpl:711
  LayoutCSR.chpl:425

when compiling $CHPL_HOME/test/users/engin/sparse_bulk_dist/sparseBulk.chpl -ssparseLayoutType=CSR

In a couple of cases, such as this, different types have the same
(nicer) name with --no-devel whereas different names with --devel.
So the latter caused extra new_IntSymbol() and so extra two nodes
compared to the former.

WITH THIS CHANGE
----------------

newMemDesc now accepts and hashes on the Type* to be allocated,
so distinct types always get distinct memory descriptors
regardless of --devel or not.

As a result, in the above compilation the number of AST nodes
matches between --devel and --no-devel until insertLineNumbers().
insertLineNumbers and codegen still add more nodes with --no-devel
than with --devel. For the above compilation I got ~13k extra in
insertLineNumbers and another ~1.1k in codegen.
(I did not investigate these.)

Interestingly, all calls to newMemDesc resulting in new descriptors
come from resolution - except for one. It happens during scopeResolve
calling callChplHereAlloc(fn->_this->type) (was callChplHereAlloc(fn->_this))
from within build_constructor(AggregateType* ct)
invoked on the generic _ref type.
The corresponding descriptor was associated with "this", now with "_ref".
I did not investigate what becomes of this allocation
when the constructor for _ref is instantiated, if it ever is.
(I did not see any "this" descriptors from nightly memleaks testing.)
This seems to be the first call to newMemDesc, so the memory descriptor is 0.

IMPLEMENTATION
--------------

As suggested by Michael, I made newMemDesc() create separate
descriptor numbers based on the type, rather than its name.

* This required a deviation from name-centric memory descriptors.
In a few cases newMemDesc() is invoked on char* literals e.g.
"global heap-converted data", "bundled args", etc.
Prior to this change, there was a 1:1 between the name and
the descriptor number. Now the same type name can have several
descriptors if multiple Types have the same name. See the above
example.

newMemDesc() still has an implementation that accepts strings.
It uses a separate hash table and issues distinct descriptors
from those on Types.

Michael and I discussed merging the two hash tables into one,
however did not find an easy solution, and it did not seem
worth it to me to try to merge them.

* To implement that, I created a separate newMemDesc() and memDescsMap
that operate on Types. Note that memDescsNodeMap hashes on Type ID,
not the Type* pointer. This is to guard against the case where a Type
is removed from tree and its pointer is reused. Even though such
a case is unlikely.

Michael has reservations about it, perhaps based on the rest
of the compiler using node pointer rather than node ID for hashing
in cases like this. I have been bitten once by pointer-based hashing
when the node used as the key got deallocated then another node
got allocated at the same address. The Map gave me an incorrect
node as a result. Given that I am unaware of the policy that Type
nodes are never deleted (even though in reality that may well be the
case), I am being extra cautious and using IDs for hashing.
However, we can revert to hashing on pointers if desired.

MORE MEMORY DESCRIPTORS NOW - esp. for _ic_ types
-------------------------------------------------

The names associated with memory descriptors in the compiled program's
output - those are still always char* - are visible e.g. when running
with --memLeaks.

This is most prominent for iterator classes, they are named _ic_....

Now we create separate descriptors for ICs corresponding to serial,
leader, follower, and standalone iterators (when applicable),
whereas before they were collapsed into one. I left it this way
encouraged by Brad. We can still collapse them by dispatching calls
on IC types to the string-based implementation of newMemDesc.
Of we can be fancy and extend the names for memory descriptors
of IC types with "(leader)", "(follower)", etc.

So now we may be observing multiple lines with the same descriptor name
instead of a single line with combined counts.

I did not come up with an example illustrating that. The test
discussed above leaks those arrays, however it does not illustrate.
We might see something in the nightly memleak testing where we do leak
_ic_ things occasionally.

WHILE THERE
-----------

* I modified callChplHereAlloc() to always accept Type*
instead of Symbol*, thus avoiding Type*->Symbol*->Type* conversion.

The only place where this was non-trivial was replacement:
  callChplHereAlloc(fn->_this) --> callChplHereAlloc(fn->_this->type)
in
  build_constructor(AggregateType* ct)
(See above for a remark on this.)

* Removed a couple of unnecessary lines from callChplHereAlloc().

* Removed an unnecessary SET_LINENO() call from scopeResolve().

* I considered removing the optional 'md' formal in callChplHereAlloc()
because it is currently unused. I chose to leave it there just in case
and for symmetry with insertChplHereAlloc.

### DIFF
--- a/compiler/AST/primitive.cpp
+++ b/compiler/AST/primitive.cpp
@@ -702,16 +702,26 @@ initPrimitive() {
   prim_def(PRIM_REQUIRE, "require", returnInfoVoid, false, false);
 }
 
-Map<const char*, VarSymbol*> memDescsMap;
+static Map<const char*, VarSymbol*> memDescsMap;
+static Map<int, VarSymbol*> memDescsNodeMap;
 Vec<const char*> memDescsVec;
+static int64_t memDescInt = 0;
 
 VarSymbol* newMemDesc(const char* str) {
-  static int64_t memDescInt = 0;
   const char* s = astr(str);
   if (VarSymbol* v = memDescsMap.get(s))
     return v;
   VarSymbol* memDescVar = new_IntSymbol(memDescInt++, INT_SIZE_16);
   memDescsMap.put(s, memDescVar);
   memDescsVec.add(s);
+  return memDescVar;
+}
+
+VarSymbol* newMemDesc(Symbol* sym) {
+  if (VarSymbol* v = memDescsNodeMap.get(sym->id))
+    return v;
+  VarSymbol* memDescVar = new_IntSymbol(memDescInt++, INT_SIZE_16);
+  memDescsNodeMap.put(sym->id, memDescVar);
+  memDescsVec.add(sym->name);
   return memDescVar;
 }
--- a/compiler/AST/primitive.cpp
+++ b/compiler/AST/primitive.cpp
@@ -703,7 +703,7 @@ initPrimitive() {
 }
 
 static Map<const char*, VarSymbol*> memDescsMap;
-static Map<int, VarSymbol*> memDescsNodeMap;
+static Map<int, VarSymbol*> memDescsNodeMap;  // key is the Type node's ID
 Vec<const char*> memDescsVec;
 static int64_t memDescInt = 0;
 
@@ -717,11 +717,11 @@ VarSymbol* newMemDesc(const char* str) {
   return memDescVar;
 }
 
-VarSymbol* newMemDesc(Symbol* sym) {
-  if (VarSymbol* v = memDescsNodeMap.get(sym->id))
+VarSymbol* newMemDesc(Type* type) {
+  if (VarSymbol* v = memDescsNodeMap.get(type->id))
     return v;
   VarSymbol* memDescVar = new_IntSymbol(memDescInt++, INT_SIZE_16);
-  memDescsNodeMap.put(sym->id, memDescVar);
-  memDescsVec.add(sym->name);
+  memDescsNodeMap.put(type->id, memDescVar);
+  memDescsVec.add(type->symbol->name);
   return memDescVar;
 }
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -1623,7 +1623,7 @@ CallExpr* callChplHereAlloc(Symbol *s, VarSymbol* md) {
   // Since the type is not necessarily known, resolution will fix up
   // this sizeof() call to take the resolved type of s as an argument
   sizeExpr = new CallExpr(PRIM_SIZEOF, new SymExpr(s));
-  mdExpr = (md != NULL) ? md : newMemDesc(s->name);
+  mdExpr = (md != NULL) ? md : newMemDesc(s);
   CallExpr* allocExpr = new CallExpr("chpl_here_alloc", sizeExpr, mdExpr);
   // Again, as we don't know the type yet, we leave it to resolution
   // to put in the cast to the proper type
@@ -1643,7 +1643,7 @@ void insertChplHereAlloc(Expr *call, bool insertAfter, Symbol *sym,
                                     new CallExpr(PRIM_SIZEOF,
                                                  (ct != NULL) ?
                                                  ct->symbol : t->symbol));
-  VarSymbol* mdExpr = (md != NULL) ? md : newMemDesc(t->symbol->name);
+  VarSymbol* mdExpr = (md != NULL) ? md : newMemDesc(t->symbol);
   Symbol *allocTmp = newTemp("chpl_here_alloc_tmp", dtCVoidPtr);
   CallExpr* allocExpr = new CallExpr(PRIM_MOVE, allocTmp,
                                      new CallExpr(gChplHereAlloc,
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -1623,7 +1623,7 @@ CallExpr* callChplHereAlloc(Symbol *s, VarSymbol* md) {
   // Since the type is not necessarily known, resolution will fix up
   // this sizeof() call to take the resolved type of s as an argument
   sizeExpr = new CallExpr(PRIM_SIZEOF, new SymExpr(s));
-  mdExpr = (md != NULL) ? md : newMemDesc(s);
+  mdExpr = (md != NULL) ? md : newMemDesc(s->type);
   CallExpr* allocExpr = new CallExpr("chpl_here_alloc", sizeExpr, mdExpr);
   // Again, as we don't know the type yet, we leave it to resolution
   // to put in the cast to the proper type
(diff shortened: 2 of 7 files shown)


compiler/AST/primitive.cpp
```diff
@@ -702,11 +702,12 @@ initPrimitive() {
   prim_def(PRIM_REQUIRE, "require", returnInfoVoid, false, false);
 }
 
-Map<const char*, VarSymbol*> memDescsMap;
+static Map<const char*, VarSymbol*> memDescsMap;
```
mppf: I think I'd expect you to remove memDescMap and replace newMemDesc with something like this:

```
 VarSymbol* newMemDesc(const char* str) {
  newMemDesc(new_StringSymbol(str));
}
```

Is there some reason that can't work?

compiler/AST/primitive.cpp
```diff
@@ -702,11 +702,12 @@ initPrimitive() {
...
+static Map<int, VarSymbol*> memDescsNodeMap;
```
mppf: I'm not crazy about the 'int' key here but I understand why it might make it easier. Do we use AST ids in this way in other places?

bradcray:
@vasslitvinov : Is this the PR that resulted in the change in how memory leaks are displayed that you were telling me about by the water cooler yesterday?  Or did you patch things up so that there wouldn't be a delta anymore?  If there is a change, could you describe what the old vs. new behavior is, maybe using output from a memleaks run?

compiler/AST/primitive.cpp
vasslitvinov: This particular code does not work because, I believe, new_StringSymbol(str)->name is something like _literal_3093, and we do not want these things to show in the memory leaks table.

compiler/AST/primitive.cpp
vasslitvinov: I tripped over the issue a couple of months ago, do not remember off hand where.

Other than that, I am not aware of other places where we use ints as keys.

compiler/AST/primitive.cpp
mppf: The simple solution to that is to give newMemDesc a special case for string symbols. If you think that is better than having 2 maps (I do, but it's not a sure thing) then it's worth trying.

compiler/AST/primitive.cpp
mppf: Isn't the argument key of this map really a TypeSymbol* ?
I don't think we ever remove TypeSymbols. Am I wrong?

compiler/AST/primitive.cpp
```diff
@@ -715,3 +716,12 @@ VarSymbol* newMemDesc(const char* str) {
   memDescsVec.add(s);
   return memDescVar;
 }
+
+VarSymbol* newMemDesc(Symbol* sym) {
```
mppf: Let's make this take in a TypeSymbol*

vasslitvinov: When newMemDesc is invoked from callChplHereAlloc(), it is passed a Symbol, not a TypeSymbol. I doubt it is worth converting Symbol to TypeSymbol in that case.

compiler/AST/primitive.cpp
vasslitvinov: Most likely. However, are you willing to bet developers' time that this will never happen?

compiler/AST/primitive.cpp
mppf: But it doesn't make sense to have memDescs for different variables. MemDescs are for types, right? It would be trivial to pass s->typeInfo()->symbol in callChplHereAlloc.

compiler/AST/primitive.cpp
mppf: I think it's more likely that somebody will be "suprised" by this use of AST ids than it is that they'll be surprised that we can't delete TypeSymbols. If I am correct that TypeSymbols are never dead-code eliminated, this code would be less unusual if it just used TypeSymbol* as the key to the map. (Or a Type* for that matter).

compiler/AST/primitive.cpp
mppf: The argument could be a Type* too and that would be just as good as a TypeSymbol* in my opinion.

vasslitvinov: Good idea. Would you be OK with s->type->symbol instead?

vasslitvinov: Or maybe we make newMemDesc accept Type* instead of TypeSymbol* ?

mppf: I think it's probably the most sane to make newMemDesc accept a Type*

compiler/AST/primitive.cpp
vasslitvinov: If we use new_StringSymbol(str), then we cannot hash on TypeSymbols, so our other discussion thread won't work.

mppf: I'd like the interface for the rest of the compiler to take in Type* or const char *.
You can write a do_newMemDescInternal or whatever you choose to name it that handles the details to get both to be a Symbol*. I'd rather get the interface right than reduce the number of hashtables from 2 to 1.

compiler/AST/primitive.cpp
vasslitvinov: So here is the trade-off:

* hash on IDs => somebody is surprised => they read the comment on memDescsNodeMap (which I will add) and figure it out

* hash on Type* => a minuscule chance of a bug when we possibly delete a Type* possibly in the future => misleading information in memLeaks output => either doesn't matter or takes hours? days? to figure out

Your call.

mppf: I don't have a strong opinion about this. However I think we should be clear with ourselves whether or not Type* can be deleted. E.g. in LLVM they cannot. I sense you prefer hashing on integer IDs which is OK for now.

mppf:
This is OK with me, provided testing is clean.

PR #5195, opened 2017-01-19
Miscellaneous QualifiedType improvements

benharsh:
- Improve AstDump::writeSymbol to display wide-ref qualifier
- Simplify returnInfoVal
- Correct some usage of codegenValue
- Add constructor: QualifiedType(Qualifier, Type*)
- Add methods toRef and toVal on QualifiedTypes
- Fix def/use bug in parallel.cpp when trying to insert broadcasts
- The 'concreteIntent' function should return INTENT_REF if the given
  type has FLAG_REF
- Use INTENT_REF in ReturnByRef::addFormal when creating a _retArg
- insertLocalsForRefs should look for a single PRIM_MOVE

Testing:
- [x] full local, no-local
- [x] full gasnet
- [x] llvm/llvm-no-local on examples/

### DIFF
--- a/compiler/passes/resolveIntents.cpp
+++ b/compiler/passes/resolveIntents.cpp
@@ -97,7 +97,11 @@ IntentTag blankIntentForType(Type* t) {
 
 IntentTag concreteIntent(IntentTag existingIntent, Type* t) {
   if (existingIntent == INTENT_BLANK) {
-    return blankIntentForType(t);
+    if (t->symbol->hasFlag(FLAG_REF)) {
+      return INTENT_REF;
+    } else {
+      return blankIntentForType(t);
+    }
   } else if (existingIntent == INTENT_CONST) {
     return constIntentForType(t);
   } else {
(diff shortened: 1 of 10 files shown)


compiler/passes/resolveIntents.cpp
```diff
@@ -97,7 +97,11 @@ IntentTag blankIntentForType(Type* t) {
 
 IntentTag concreteIntent(IntentTag existingIntent, Type* t) {
   if (existingIntent == INTENT_BLANK) {
-    return blankIntentForType(t);
+    if (t->symbol->hasFlag(FLAG_REF)) {
+      return INTENT_REF;
+    } else {
+      return blankIntentForType(t);
+    }
```
vasslitvinov: Please add a comment explaining why returning INTENT_REF for ref types. Is this a convention? When do we call concreteIntent() on ref types?

vasslitvinov:
Other than my comment, this is good to go.

PR #5197, opened 2017-01-19
Recode the multi-ddata serial iterator so that it will inline.

gbtitus:
As it was coded the multi-ddata serial iterator would not inline.  Both
the execution-time test against mdNumChunks and the doubly-nested loop
around the yield-stmt in the mdNumChunks!=1 case broke the limitations
required for inlining to occur.  Here, recode so that we can inline this
iterator.  This restores the pre-multi-ddata locModel=numa performance
for studies/prk/stencil/stencil and some others.


gbtitus:
@benharsh , would you review this one also, please?

benharsh:
Looks good to me! 👍

PR #5200, opened 2017-01-20
First draft of deinit()

bradcray:
This patch is a first draft of adding deinit() to the compiler as the next-generation alternative to destructors.  In this patch, either destructors or deinit can be used.  The main difference between the two is that deinit() can be called by a user and can optionally take arguments (as long as there is still a zero-argument version available).  The rationale for these differences is that users sometimes wish to call destructors/deinitializers on their types in order to free up resources without freeing up the object (via delete or letting it go out of scope).  Of course, a well-written class/record would support multiple such calls without causing problems.

I undertook this mod after realizing that we'd failed to sign up anyone to own it in the 1.15 planning, and after being assured that it shouldn't be hard.  To the extent that my attempt is correct-ish, it wasn't -- a 30-minute fix (TM).  I essentially:

* had the parser flag functions named deinit() as it did functions starting with ~
* loosened some requirements on deinit() to permit what I believe the expected semantics should be
* renamed chpl__deinit() to deinit()

Here I file tests that
* define deinit() and have it be called by delete
* show what happens if you define both deinit() and a destructor (ambiguity)
* called deinit()
* called a version of deinit() that takes an optional argument
* show that deinit() can be overloaded, that the compiler will call the right one, and that the user can call either
* retired some futures that anticipated deinit()
* updated another test whose output would've had to change to not name chpl__deinit() to use deinit() instead

Note that so far I've done nothing to test deinit() on records which may present additional challenges (or perhaps simply extraneous calls to deinit()?)

### DIFF
--- a/test/classes/deinitializers/deinitAndDestruct.good
+++ b/test/classes/deinitializers/deinitAndDestruct.good
@@ -1,0 +1,3 @@
+deinitAndDestruct.chpl:11: error: ambiguous call 'C.chpl__deinit()'
+deinitAndDestruct.chpl:2: note: candidates are: C.~C()
+deinitAndDestruct.chpl:6: note:                 C.deinit()
--- a/test/classes/deinitializers/deinitAndDestruct.good
+++ b/test/classes/deinitializers/deinitAndDestruct.good
@@ -1,3 +1,3 @@
-deinitAndDestruct.chpl:11: error: ambiguous call 'C.chpl__deinit()'
+deinitAndDestruct.chpl:11: error: ambiguous call 'C.deinit()'
 deinitAndDestruct.chpl:2: note: candidates are: C.~C()
 deinitAndDestruct.chpl:6: note:                 C.deinit()
--- a/test/classes/diten/multipledestructor.chpl
+++ b/test/classes/diten/multipledestructor.chpl
@@ -1,7 +1,7 @@
 class C {
-  proc ~C() { writeln("In ~C #1"); }
+  proc deinit() { writeln("In deinit #1"); }
 }
-proc C.~C() { writeln("In ~C #2"); }
+proc C.deinit() { writeln("In deinit #2"); }
 
 proc main {
   var c = new C();
--- a/test/classes/initializers/error-destructor-return-type.good
+++ b/test/classes/initializers/error-destructor-return-type.good
@@ -1,1 +1,1 @@
-error-destructor-return-type.chpl:2: error: destructors may not declare a return type
+error-destructor-return-type.chpl:4: error: destructors may not declare a return type
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -2394,7 +2394,8 @@ buildFunctionSymbol(FnSymbol*   fn,
   fn->cname   = fn->name = astr(name);
   fn->thisTag = thisTag;
 
-  if (fn->name[0] == '~' && fn->name[1] != '\0')
+  if ((fn->name[0] == '~' && fn->name[1] != '\0') ||
+      (strcmp(fn->name, "deinit") == 0))
     fn->addFlag(FLAG_DESTRUCTOR);
 
   if (receiver)
--- a/compiler/passes/buildDefaultFunctions.cpp
+++ b/compiler/passes/buildDefaultFunctions.cpp
@@ -1439,12 +1439,12 @@ static void buildStringCastFunction(EnumType* et) {
 
 
 void buildDefaultDestructor(AggregateType* ct) {
-  if (function_exists("chpl__deinit", 2, dtMethodToken, ct))
+  if (function_exists("deinit", 2, dtMethodToken, ct))
     return;
 
   SET_LINENO(ct->symbol);
 
-  FnSymbol* fn = new FnSymbol("chpl__deinit");
+  FnSymbol* fn = new FnSymbol("deinit");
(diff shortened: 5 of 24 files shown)


bradcray:
I opened an issue regarding deinit() design questions in issue #5201

lydia-duncan:
I'm going to review the tests now and wait to review the implementation until after design discussion has completed.

test/classes/deinitializers/deinitAndDestruct.good
```diff
@@ -0,0 +1,3 @@
+deinitAndDestruct.chpl:11: error: ambiguous call 'C.deinit()'
+deinitAndDestruct.chpl:2: note: candidates are: C.~C()
+deinitAndDestruct.chpl:6: note:                 C.deinit()
```
lydia-duncan: I think there should be an error if a user defines both a destructor and a deinitializer, even if resolving the call would not be ambiguous.

If we decide to not permit deinitializers to take additional arguments, then the normal resolution error is probably sufficient, but it seems friendlier to explicitly call out to the user that we don't permit both styles to be defined.

test/classes/diten/multipledestructor.chpl
```diff
@@ -1,7 +1,7 @@
 class C {
-  proc ~C() { writeln("In ~C #1"); }
+  proc deinit() { writeln("In deinit #1"); }
 }
-proc C.~C() { writeln("In ~C #2"); }
+proc C.deinit() { writeln("In deinit #2"); }
```
lydia-duncan: For my own sanity, I would prefer a copy of this test be made in either the test/classes/deinitializers or test/classes/initializers/ directories, rather than overwriting a test for destructors.  It allows me to limit the spot check testing I do to just those directories rather than running over the entire test/classes directory structure (which gets covered by my full standard run before committing anything)

test/classes/initializers/error-destructor-return-type.good
```diff
@@ -1 +1 @@
-error-destructor-return-type.chpl:2: error: destructors may not declare a return type
+error-destructor-return-type.chpl:4: error: destructors may not declare a return type
```
lydia-duncan: Whoops, thanks for catching that!

test/classes/deinitializers/deinitAndDestruct.good
lydia-duncan: Do you anticipate that being a huge implementation effort?  The check was relatively straight-forward for initializers

test/classes/diten/multipledestructor.chpl
lydia-duncan: It involves deinitializers now, which is directly related to initializers.  If my work happens to modify deinitializers, I would discover this through running both the aforementioned directories and if necessary would update those tests.  Having this test outside of those directories would mean that I wouldn't necessarily remember this one and so wouldn't include it in the commit with the others and wouldn't discover the change until the full standard run.  We should as much as possible avoid spreading tests related to a feature out in general for this very reason.  (btw, I'm stating my supporting viewpoints so you understand where I'm coming from, not because I necessarily think you disagree with them)

test/classes/deinitializers/deinitAndDestruct.good
lydia-duncan: Note: if we follow Mike's approach and merge this, you don't need to address this now, as we'll continue discussing it)

test/classes/diten/multipledestructor.chpl
lydia-duncan: Note: if we take Mike's approach and merge this PR now, I would still prefer this change occur today.  I can also do it myself, if you don't want to but don't otherwise object

test/classes/deinitializers/deinitAndDestruct.good
bradcray: I'm not seeing the value in adding it.  It's not clear to me when a user would write this code and find the current message confusing, especially given that deconstructors aren't long for this world (i.e., should soon give a deprecation warning).

lydia-duncan: Sorry, didn't mean that as a "we'll continue discussing it here".  I'm curious what other members of the initializers team think but am not convinced they've seen this question.  And if we're going to continue discussing and adding support for deinitializers anyways, this wouldn't land in your court even if the rest of the initializers team agreed with me

test/classes/diten/multipledestructor.chpl
bradcray: What you're describing is black-box testing (i.e., arguing from a position of having no knowledge of how the compiler works).  I assert (from a white-box viewpoint) that the implementation of deinit is so identical to deconstructors that if you broke deinit, you'd almost certainly break destructors as well, so would need to test everything that used a destructor (which is essentially _everything_) to be sure things are OK (and you'd need to do this to merge your code anyway).  Moreover, I think you'd be hard-pressed to break deinit in such a way that the deinit tests in test/initializers/ didn't break but this one did.  I also don't see how deinitializers are directly related to initializers.

I also don't think it's reasonable to assume that everything that uses init() and deinit() live under test/initializers.  As new features like these become available and the preferred way of doing things, they're going to crop up in arbitrary places (e.g., we didn't move binary-trees to test/initializers once it used init()).  Once deinit() is on master, I would hope none of us would write another destructor again (and to lazily convert ones we come across to deinit to prepare for the future).

PR #5202, opened 2017-01-20
Add annotations for this week

lydia-duncan:
Covers AST size changes, improvements to the mandelbrot benchmark, and the
addition of new tests which contain unsurprising memory leaks, as well as a
note on what is likely due to cache effects

Verified that -generate-graphs works as expected, and that an intentional error
in the ANNOTATIONS file led to an expected failure (to ensure I would detect if
something went wrong)


lydia-duncan:
I don't want to break performance testing over the weekend, so I will merge this on Monday

Issue #5141, opened 2017-01-10
array fields with domain queries (?)


bradcray:
## Semantic / Language Design Question

Should we support the ability to declare array fields that use domain queries (e.g., 'X: [?D] real;')?  I don't believe we ever intended to support domain queries here, but is there any reason not to?  Would 'D' be a local snapshot of the originating array's domain, or would it be a reference to it?

```chapel
// Excerpted from test/classes/generic/domainQueryField-feature.chpl:
class C {
  var X: [?D] real;
}

var A: [1..1000] real;
var myC = new C(A);
```

bradcray:
Speaking with @noakesmichael today, we came up with a variation on this which also seems like it would be compelling [Mike, it turns out I was misremembering precisely what this existing issue had been trying to do]:

````chapel
class C {
  var x: ?t;
  var y: t;
  var z: [1..n] t;
}
````

as a means of having a class be generic.  Without the query syntax, there are other ways to write this, though more verbose:

````chapel
class C {
  var x;
  var y: x.type;
  var z: [1..n] x.type;
}
````

or, in a future with private class members:

````chapel
class C {
  var x;
  private type t = x.type;
  var y: t;
  var z: [1..n] x.type;
````

of these, the version with the query field seems most palatable.  It also seems conceivable that a good job of introducing compiler-defined initializers for generics could lean on existing mechanisms for type query fields in argument lists...  (or maybe more accurately, that the two scenarios could use the same mechanisms even if it's different from what we do today).

PR #5203, opened 2017-01-20
Generate an error for domain queries in field

### LINKED ISSUE #5140: domain query fields cause internal errors
### Summary of Problem

When declaring array fields that use domain queries (e.g., 'X: [?D] real;'), the compiler generates an internal error.  Internal errors are always indicative of bugs.  I don't believe we ever intended to support domain queries here, so it should generate a helpful user-focused error message rather than an internal error.  (See also related issue #5141)

### Steps to Reproduce

**Source Code:**

```chapel
// Excerpted from test/classes/generic/domainQueryField.chpl:
class C {
  var X: [?D] real;
}

var A: [1..1000] real;
var myC = new C(A);
```

**Compile command:**
chpl domainQueryField.chpl

**Execution command:**
N/A

### Configuration Information

- Output of `chpl --version`: 1.14.0.117304e
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

bradcray:
This PR fixes #5140 in which a user reported that the use of domain queries in array field types resulted in an internal error.  This is something we've never intended to support, but have not generated a good error message for either.  Issue #5141 asks whether it's something we might want to support down the line.

The approach taken here is to add a check to the checkParsed pass that looks for DefExprs whose types are arrays and whose queries themselves contain a DefExpr.  I borrowed this logic from normalize.cpp which processes such domain queries when they do occur in formal argument types.

### DIFF
--- a/compiler/passes/checkParsed.cpp
+++ b/compiler/passes/checkParsed.cpp
@@ -55,7 +55,7 @@ checkParsed() {
   }
 
   forv_Vec(DefExpr, def, gDefExprs) {
-    if (toVarSymbol(def->sym))
+    if (toVarSymbol(def->sym)) {
       // The test for FLAG_TEMP allows compiler-generated (temporary) variables
       // to be declared without an explicit type or initializer expression.
       if ((!def->init || def->init->isNoInitExpr())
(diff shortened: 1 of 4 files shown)


compiler/passes/checkParsed.cpp
```diff
@@ -66,6 +66,25 @@ checkParsed() {
               USR_FATAL_CONT(def->sym,
                              "Variable '%s' is not initialized or has no type",
                              def->sym->name);
+    }
+
+    //
+    // This test checks to see if query domains (e.g., '[?D]') are
+    // used in places other than formal argument type specifiers.
+    //
```
noakesmichael: Largely FYI and not a request to edit.  It is increasingly the convention to prefer
isFnSymbo(), which returns a bool, rather than toFnSymbol(), which returns NULL
or a FnSymbol* in this context.

PR #5205, opened 2017-01-20
Fix problems with extern blocks, extern types

### LINKED ISSUE #5166: Example of extern blocks with GSL (in test/users/npadmana) no longer compiles
## Bug Report

The test of extern block support using GSL test/users/npadmana/gsl_demonstration no longer compiles (I've looked back as far as 1.14 here). This test completely relies on chapel to parse the GSL header files, and exercises the extern block support.

### Summary of Problem

I believe there are two separate issues here.

* On an Ubuntu 16.04 machine, with the stock GCC (GCC 5.4) and GSL packages, chpl immediately complains of missing symbols (eg. "gsl_sf_result"), even those these are correctly defined in the included header files. If I drop down to Ubuntu 14.04 (with GSL 1.16), it correctly gets the definitions from the header files (but see below). I believe (but will verify) that the GSL headers are unchanged from the 16.04 and 14.04 cases, so I think this is an issue with the compiler interacting with the chapel-packaged version of clang. [EDIT : I did verify that this failure seems to be associated with the GCC version, and not GSL].

* Even in this case, the "gsl_spline" case fails with the following error --

error: unresolved call 'gsl_interp_type.initmeme=gsl_interp_type'
note: candidates are: gsl_interp_type.init(name: c_string, min_size: c_uint, alloc: c_fn_ptr
, init: c_fn_ptr, eval: c_fn_ptr, eval_deriv: c_fn_ptr, eval_deriv2: c_fn_ptr, eval_integ: c_fn_ptr, free: c_fn_ptr)

I think this is coming from trying to pass in an external const pointer to a C-struct into a function. I believe this previously worked because the structure was being treated as opaque.

### Steps to Reproduce

The test case in the repository should work, on a system with GSL installed and $GSL_DIR set. The current compopts file assumes that the gsl header and library files are in $GSL_DIR/include and $GSL_DIR/lib respectively (if not in some standard system location.

If it is useful, I could package the two cases up in docker files to reproduce.....

### Configuration Information

Tested on Chapel 1.14, built with CHPL_LLVM=llvm.

mppf:
This is a bug-fix PR - fixes #5166.

* Fixes a bug with .c_str() in externCResolve.cpp 
* Fixes an issue where, in a PRIM_MOVE between type variables, a RHS that is a type alias to an extern type might not be resolved yet.
* Don't try to call `init` on extern types in the generated `initCopy`. Instead, generate a trivial `initCopy` for extern types.

This gets test/users/npadmana/gsl_demonstration/gsl_demo.chpl working for me on an Ubuntu 16 system.

Passed test/extern/ferguson/ test/extern/jjueckstock/ test/release/examples/primers/ test/users/npadmana/ with and without --llvm in a compiler built with CHPL_LLVM=llvm.
Passed full local testing with CHPL_LLVM=none.

### DIFF
--- a/compiler/passes/buildDefaultFunctions.cpp
+++ b/compiler/passes/buildDefaultFunctions.cpp
@@ -1078,14 +1078,20 @@ static void build_record_copy_function(AggregateType* ct) {
 
   fn->insertFormalAtTail(arg);
 
-  CallExpr* call = NULL;
+  Expr* toReturn = NULL;
 
   if (foundUserDefinedCopy) {
-    call = new CallExpr(PRIM_NEW, ct->symbol, new SymExpr(arg));
+    toReturn = new CallExpr(PRIM_NEW, ct->symbol, new SymExpr(arg));
     // If it has a user-defined copy initializer, it's not POD
     ct->symbol->addFlag(FLAG_NOT_POD);
+  } else if (ct->symbol->hasFlag(FLAG_EXTERN)) {
+    // Extern records/classes should only get trivial initCopy fns
+    // (at least if no other init method was defined).
+    toReturn = new SymExpr(arg);
   } else {
+    CallExpr* call = NULL;
     // generate the default copy initializer in chpl__initCopy for now
+    // which is currently implemented to call the compiler-generated initializer
 
     // MPF 2016-11-03: It would be better to move all of the logic below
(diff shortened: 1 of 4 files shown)


npadmana:
For the record -- I just tried this as well, worked perfectly! Thanks!

vasslitvinov:
Looks good to me.

PR #5206, opened 2017-01-20
Using API to compare and print Task ID

srirajpaul:
Follow up on Pull requests #5152 and #5177.
The task IDs are used like integers, for example
compare using binary operators and print using %d
format specifier. This will not work when task ID
is of non integer type such as struct. Therefore
use API to do comparison and conversion to string.

### DIFF
--- a/runtime/src/chpl-visual-debug.c
+++ b/runtime/src/chpl-visual-debug.c
@@ -64,6 +64,9 @@ static void cb_comm_executeOn_fast(const chpl_comm_cb_info_t *info);
 int chpl_vdebug_fd = -1;
 int chpl_vdebug = 0;
 
+static char buff_tid[CHPL_TASK_ID_STRING_MAX_LEN];
+#define TID_STRING(tid) (chpl_task_idToString(buff_tid, CHPL_TASK_ID_STRING_MAX_LEN, tid))
+
 int chpl_dprintf (int fd, const char * format, ...) {
   char buffer[2048]; 
   va_list ap;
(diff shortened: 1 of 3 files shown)


runtime/src/chpl-visual-debug.c
```diff
@@ -64,6 +64,9 @@ static void cb_comm_executeOn_fast(const chpl_comm_cb_info_t *info);
 int chpl_vdebug_fd = -1;
 int chpl_vdebug = 0;
 
+static char buff_tid[CHPL_TASK_ID_STRING_MAX_LEN];
```
gbtitus: This can't be static.  If multiple Chapel tasks do TID_STRING() invocations at once chaos will ensue due to the buffer sharing.  The buffer needs to be function-local, and TID_STRING changed to accept both the buffer and the tid as arguments.

runtime/src/qio/qio.c
```diff
@@ -95,12 +95,12 @@ bool qio_allow_default_mmap = true;
 #ifdef _chplrt_H_
 qioerr qio_lock(qio_lock_t* x) {
   // recursive mutex based on glibc pthreads implementation
-  int64_t id = chpl_task_getId();
+  chpl_taskID_t id = chpl_task_getId();
 
   assert( id != NULL_OWNER );
```
gbtitus: This has to be a chpl_task_idEquals() comparison also, doesn't it?

srirajpaul: Oops. compiled without CHPL_DEVELOPER and it disabled the asserts.

gbtitus:
Is this ready for re-review?  It looks like it but I didn't know if there were any other changes on the way.

srirajpaul:
Yes. This is ready for re-review.

gbtitus:
Apologies for the delay!  This looks good and can be merged.

PR #5207, opened 2017-01-20
Add future field to template file

ben-albrecht:
Add optional `Associated Future Test(s)` field to template to include path and PR # for a test demonstrating a bug.

Also in these changes:
- Remove `# Bug Report` header (can be confused for `## Summary of Problem` field)
- Make execution example more examplary, using numlocales arg: `./foo -nl 4`

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,3 @@
-## Bug Report
 <!--
 If you are filing an issue that is not a bug report, please feel free to erase
 this template and describe the issue as clearly as possible.
@@ -28,7 +27,7 @@ What behavior did you expect to observe?
 <!-- e.g. `chpl foo.chpl -o foo` -->
 
 **Execution command:**
-<!-- e.g. `./foo 42`. If an input file is required, include it as well. -->
+<!-- e.g. `./foo -nl 4`. If an input file is required, include it as well. -->
 
 **Associated Future Test(s):**
 <!-- Are there any tests in Chapel's test system that demonstrate this issue?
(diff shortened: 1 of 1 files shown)


bradcray:
*This* is what I've been missing!

.github/ISSUE_TEMPLATE.md
```diff
@@ -30,6 +30,9 @@ What behavior did you expect to observe?
 **Execution command:**
 <!-- e.g. `./foo 42`. If an input file is required, include it as well. -->
 
+**Associated Future Test(s):**
+<!-- Are there any tests in Chapel's test system that demonstrate this issue?
+     e.g. test/path/to/foo.future #1234 -->
```
bradcray: One thing that occurs to me (though I see the tension) is that this is normally the thing I'd want to see just before or after the source code.  Downside: If it's a field that only makes sense to developers it's an early distraction / confusion for users.

ben-albrecht: > If it's a field that only makes sense to developers it's an early distraction / confusion for users.

This was my thinking for the placement, but I'm not attached to the decision. If we move it up, maybe it ought to include the prefix: "Optionally,"

mppf: I agree about the tradeoff here.. I wonder if we might put Compile Command and Execution Command first, followed by Source Code and Associated Future Tests. (if we're not too attached to the current order).

bradcray: I thought about that too, but feel like the important stuff should come first, like a good news article, so am reluctant to move the source code down.  In particular, as I've filed these, I usually find the compile and execution commands to be pretty tedious (the bugs I've been filing aren't particularly sensitive to those commands, so I find myself always saying 'chpl foo.chpl' and 'a.out' (when I even get to the execution stage)).

bradcray:
I was curious if anyone else would comment on this after the brief discussion in yesterday's meeting, but for myself, I'm good with this as-is.

PR #5212, opened 2017-01-23
Auto-detect GSL in gsl_demonstration SKIPIF

mppf:
For now, this just follows the pattern from the ZMQ SKIPIF which is to use Python's ctypes find_library. That finds the .so file which isn't quite what we need but is a good start.

For now, I think it's a reasonable target that all of our nightly testing systems either:
 - have ZMQ/GSL .so .a and .h files available, OR
 - do not have ZMQ/GSL .so installed

Doing a better job at this is something we absolutely ought to do but it's not clear to me how to do it easily and correctly. (One idea is to use pkg-config but I don't know to what extent we can rely on that being installed. We could also try to find .a files and .h in common places).

This should be merged after PR #5205.
That's a first step towards getting gsl_demonstration tested nightly. The next step is to enable some testing that includes test/users/npadmana/ and sets CHPL_LLVM=llvm (but doesn't necessarily use --llvm).

Tested on chapcs11. Ran CHPL_LLVM=llvm full testing there with this and #5202 using the C backend. Didn't see any new failures.

Reviewed by @awallace-cray - thanks!

### DIFF
--- a/test/users/npadmana/gsl_demonstration/SKIPIF
+++ b/test/users/npadmana/gsl_demonstration/SKIPIF
@@ -1,8 +1,25 @@
 #!/usr/bin/env python
 
 #
-# skip this test when FFTW_DIR is not set
+# This test requires:
+#  * CHPL_LLVM!=none (since the test uses extern blocks)
+# and a GSL installation detected like so:
+#  * GSL_DIR environment variable is set, OR
+#  * GSL dynamic library is detected
 #
+# Note that if the dynamic library is found, this
+# test assumes that the header and static library are available.
 
-import os
-print('GSL_DIR' not in os.environ)
+from __future__ import print_function
+from ctypes.util import find_library
+from os import environ
+
+if environ['CHPL_LLVM'] == 'none':
+  print (True) # Skip if we don't have extern block support
+else:
+  # OK, we have extern block support.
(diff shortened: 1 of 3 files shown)


test/users/npadmana/gsl_demonstration/SKIPIF
```diff
@@ -1,8 +1,25 @@
...
+# Note that if the dynamic library is found, this
+# test assumes that the header and static library are available.
 
-import os
-print('GSL_DIR' not in os.environ)
+from __future__ import print_function
+from ctypes.util import find_library
+from os import environ
+
+if environ['CHPL_LLVM'] == 'none':
```
awallace-cray: This could throw a KeyError if GSL_DIR is not in the environment at all.

mppf: CHPL_LLVM is guaranteed to be in the environment by our testing system, as far as I know. In particular:
 *  `util/test/testEnv` runs skipif files
 * `testEnv` sets the environment based upon the output of `util/printchplenv --sh`
 * `util/printchplenv` always includes CHPL_LLVM

awallace-cray: okay.

PR #5215, opened 2017-01-23
Filing a bug against recursive implicit calls to .these()

bradcray:
The test recursive-iter.chpl shows that the compiler has trouble resolving
recursive iterators that implicitly call these().  If the call to these() is
made explicit, as shown in recursive-iter-works.chpl, then things work
as expected.


bradcray:
Issue #5216 corresponds to this bug

PR #5222, opened 2017-01-23
Add bug highlighting reduction bug with foralls inside of leaders

benharsh:


benharsh:
@vasslitvinov : enough info there for you?

vasslitvinov:
@benharsh - yes, thank you.

How about using dom.these() instead of dom._value.these(), to make it not hacky?

benharsh:
The 'hacky' part to me is more about specifying the tag, not calling _value (though that too is hacky). In my original BlockCyclic issue, I do need to call _value.

Issue #5223, opened 2017-01-23
+ reduce not working when a normal forall loop is used inside the leader


benharsh:
### Summary of Problem
When composing the leader iterator of a forall loop, I used a normal forall loop over a domain to yield indices. The follower iterator simply yielded what it received (an index in the domain). When calling ``+ reduce`` over this iterator, I was surprised to get a result of ``0``.

Manually invoking the leader and follower separately generates the correct result.

### Steps to Reproduce

See #5222 , which adds a .future to the testing suite.

benharsh:
@vasslitvinov , for your consideration.

benharsh:
@vasslitvinov : just to keep you in the loop, we ran into another instance of this in the rank-change domain view work.

PR #5224, opened 2017-01-23
More parallelism for BlockCyclic iteration

benharsh:
Prior to this change, BlockCyclic would only create one task per locale
when iterating over the domain (or array) in parallel. Iterating in
parallel over ``myStarts`` yields positive results for ptrans on
16-node-xc, and seems like a good starting point.

In support of this change, the ``tasksPerLocale`` argument to the
Blockcyclic constructor has been renamed to ``dataParTasksPerLocale`` to
match other distributions.

The parallel iterator bodies for BlockCyclicDom and BlockCyclicArr were
also cleaned up by:
- removing old commented-out debugging writelns
- caching common expressions to reduce verbosity

Testing:
- [x] full local
- [x] full gasnet

### DIFF
--- a/modules/dists/BlockCycDist.chpl
+++ b/modules/dists/BlockCycDist.chpl
@@ -506,7 +506,8 @@ iter BlockCyclicDom.these(param tag: iterKind) where tag == iterKind.leader {
         const dim    = whole.dim(j);
         const dimLow = dim.low;
 
-        var temp = max(lo, dimLow)..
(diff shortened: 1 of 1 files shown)


benharsh:
This branch results in a 4x speedup for ptrans on 16-node-xc.

benharsh:
See #5223 for more information on the TODO regarding the reduction bug.

PR #5225, opened 2017-01-24
Update fasta w.r.t. submitted versions

bradcray:
Belatedly, this adds the fasta#2 version that we submitted to the CLBG in December to the studies/shootout/submitted/ directory.  It also removes fastaredux from that directory and from the release, since the CLBG isn't tracking fastaredux any longer.  In support of this, it:

* updates the submitted fasta.graph file
* removes fastaredux from the release README and Makefile
* updates the shootout/submitted/README.md to include both fastas and both pidigits and no fastaredux


bradcray:
@ronawho, would you mind reviewing?

test/studies/shootout/submitted/fasta2.execopts
```diff
@@ -0,0 +1 @@
+--blockSize=4
```
ronawho: Do we need/want this for the submitted version?

bradcray: Without it, the correctness test is serialized which sometimes makes correctness errors in the parallel strategy not get caught until performance testing (if they're caught there).  So to the extent that the two versions might diverge from one another over time, I think it makes sense to correctness test it in this way, yes.  Let me know if you disagree.

ronawho: ok, no objections

PR #5226, opened 2017-01-24
Param guard reads/writes of chpl_arrayInitMethod

ronawho:
chpl_arrayInitMethod is a global variable. Reading it required a get, which had
the nasty side effect of increased comm counts and not being able to declare
arrays inside of a local-block.

This adds a param guard around it, which avoids the gets, but makes changing
the array init method at runtime more cumbersome.

Now instead of just doing

```chapel
chpl_arrayInitMethod = ArrayInit.parallelInit;
var A: [1..100] int;
```

you now have to compile with `-schpl_arrayInitMethodRuntimeSelectable` and do

```chapel
chpl_setArrayInitMethod(ArrayInit.parallelInit);
var A: [1..100] int;
```

which is annoying, but fine since this feature is just meant to help us explore
initialization strategies effect on first-touch.

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -605,10 +605,30 @@ module ChapelBase {
 
   enum ArrayInit {heuristicInit, noInit, serialInit, parallelInit};
   config param chpl_defaultArrayInitMethod = ArrayInit.heuristicInit;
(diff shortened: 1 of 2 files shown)


ronawho:
@benharsh still have to test, but thoughts?

benharsh:
Looks good to me 👍

PR #5229, opened 2017-01-24
Fix bug in BlockCyclic indexing

benharsh:
In #5097, @cassella reported an OOB error with BlockCyclic when using a 3-dimensional domain. I'm not sure what it was about that specific domain shape that caused the error. I wrote a more general test that ran across multiple dimensions and blocksizes (that fit evenly/oddly). This test exposed an OOB error in a 3-dimensional domain as well.

The LocBlockCyclicArr.mdInd2FlatInd method appeared to be simply incorrect. I believe the new logic is more conceptually consistent with both BlockCyclic and our existing translation of dimensional indices to flat indices.

While there I also converted some arrays into tuples, which is cleaner and appears to improve performance. Helper functions were added to aid in computing the tuple values.

Testing:
- [x] full local
- [x] full gasnet


vasslitvinov:
Looks good to me. Thank you Ben!

PR #5240, opened 2017-01-25
lower error handling constructs

psahabu:
This is the next step for the error handling feature. Passes full linux64+flat testing, in addition to tests added in this PR.

### What syntax is being lowered?
- `throws`: added to a function signature to designate that it can throw an error
- `throw`: a control flow structure to actually throw an error
- `try`: a control flow structure that propagates errors either to another `try` or out of the function
- `try!`: a control flow structure that halts on errors
- Strict mode isn't implemented yet.

### Lowering example
```
try {
  a(); // throws
  b(); // does not throw
  c(); // throws
} catch e: SubError {
  f();
} catch e: AnotherSubError {
  g();
} // propagates

// suppose this is in a throwing fn
// out _e_out: Error
{
  var _e: Error;
  a(_e);
  if _e then
    goto catches;
  b();
  c(_e);
  if _e then
    goto catches;

  label catches:
  if _e {
    if _e: SubError {
      f();
    } else if _e: AnotherSubError {
      g();
    }
    // set and return (fallback)
    {
      _e_out = _e;
      goto epilogue_label;
    }
  }
}
```

### Implementation of the above

When an error is handled by a `try` block, we would like to jump to the `catches` label using a `goto`. However, `GOTO_NORMAL` is asserted to be invalid in other parts of the compiler, so `GOTO_ERROR_HANDLING` was created.

`try` blocks can change their control flow behavior depending on their context, hence the use of the `ErrorHandlingVisitor` and its stack of `TryInfo`, each representing a `try` block.

The implementation of `try`, `try!`, and `throws` has been refactored into `haltOrPropagateError` because they each assign an object to the function's `out` variable.

Additionally, `ifErrorExists` creates AST nodes that check if an error has been set by a function call.

### Additional changes
Extends the following to work with AList for convenient insertion of multiple nodes:
- Expr::insertBefore()
- Expr::insertAfter()
- BlockStmt::insertAtHead()
- BlockStmt::insertAtTail()


compiler/passes/errorHandling.cpp
```diff
@@ -17,8 +17,185 @@
  * limitations under the License.
  */
 
+#include "AstVisitorTraverse.h"
```
noakesmichael: #include "foo.h" should always be the upper most #include in foo.cpp (to avoid a situation in which
foo.h has a hidden dependency on some other .h file

PR #5241, opened 2017-01-25
Fix a bug with array-as-vec pop_front

daviditen:
@ronawho discovered that array.pop_front() was increasing the size of the
array instead of decreasing it.

When trying to shrink an array-as-vec for the pop_front function, the order
of the operands of a subtraction were reversed, causing the array to grow
instead of shrinking.  Switch them so the array shrinks as was intended.

Passed full testing on standard linux64.  Also ran
arrays/diten/array-as-vec-start-idx.chpl 25,000 times on our 32 bit testing VM.


ronawho:
👍

Issue #5198, opened 2017-01-20
Implement initializers, based on CHIP 10


lydia-duncan:
For reference, see:
- [CHIP 10](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/10.rst)
- [Release notes](http://chapel.cray.com/releaseNotes.html)
  - for Chapel 1.14, see the Language and Compiler Improvements deck, slides 28-40
  - for Chapel 1.13, see the Ongoing Efforts deck, slides 5-23

### Summary
This work seeks to improve the support of classes and records, by replacing our old-style constructors with a more principled strategy of initialization, dubbed "initializers".  When initializers are viewed as sufficiently stable, constructors will begin generating a deprecation message, and will be removed the following release.  Details on what initializers entail may be found in the references listed above.

Tests of this feature live in the directory test/classes/initializers.  Some of those tests pass today and some do not - some tests that pass today are hiding portions that have not been implemented for initializers (for instance, those that rely on generating proper default initializers work, but default initializers are not implemented).  Some of the tests that fail today are not necessarily expected to be fixed by the initializers work, but needed to be migrated to the initializers directory instead of left forgotten in the constructors directory.  Before this feature is complete, we will revisit these tests and take a full appraisal.

Currently, developer resources are being spent on generic initializers and fixing known bugs.  Once generic initializers are complete, it is expected that resources can be placed on generating initializers by default, migrating our library types to use initializers, and implementing more of the checks.  The first two rely on the completion of generic initializers before reasonable progress can be made.

# Remaining TODOs

- [x] generic initializers
  - [x] with type fields
  - [x] with param fields
  - [x] with generic var/const fields
  - [x] for records
  - [x] instance declarations
- [ ] generate initializers by default, instead of constructors
  - potential issues here are runtime types and array alias fields
- [ ] move all library types to using initializers instead of constructors
- [ ] noinit
- [ ] finish implementing Phase 1 and Phase 2 checks, including:
  - [ ] detecting the presence of multiple `.init()` calls in a single control flow path
  - [x] recognizing and generating an error when a field is accessed before it is initialized
  - [x] detecting `const` modification in Phase 2
  - [x] generating an error when field initializations occur before a `this.init()` call
  - [x] recognizing and generating an error when a field initialization or `.init()` calls are in parallel statements
  - [x] detecting and generating an error when `type` and `param` fields are updated in Phase 2
- [ ] fix known bugs, including:
[...]

PR #5244, opened 2017-01-25
Fix issue where copies of AggregateTypes lost their initializerStyle

lydia-duncan:
Really trivial fix, just update AggregateType::copyInner to copy that
information as well - this was causing problems for me in my generics branch.

TODO: 
- [x] full paratest (in progress)


lydia-duncan:
Work in support of #5198 (not very major, but still)

PR #4574, opened 2016-09-19
reduce intents: different input/state/output types

vasslitvinov:
When specifying a reduce intent, allow specifying the input type
for the reduction operation. The output type is the type of the
outer variable. The shadow variable has the type of the accumulation
state, which is determined implicitly by the 'identity' method
of the ReduceScanOp subclass being used with the reduce intent.

See #4565 for detailed discussion.

The input type is specified as an argument to the reduction class
in the reduce intent, e.g.:

    forall ... with (MyReductionOp(int) reduce x) ....

I implemented this via explicit support in setupOneReduceIntent().
It builds a call to create a new instance of a reduction op class,
like MyReductionOp above. If the code passes an argument to the class,
then its eltType is taken to be that argument. Otherwise its eltType
is taken to be the type of the outer variable, as before.
If <1 or >1 argument, I report an error.

Also, I now avoid setting the shadow variable's type explicitly in
implementForallIntents.cpp / createShadowVars(). That way its type
is determined by the reduce op class and can differ from the outer
variable's type.

Added tests.

Added an example to reduceIntents.rst.
While there, fixed an omission of 'ref' arg intent in reduceIntents.rst.

PR #5249, opened 2017-01-26
introduce ForallStmt node

vasslitvinov:
This PR introduces the ForallStmt node - a subclass of Stmt.
These nodes are created in the parser and are lowered away in resolve().
This message is geared towards design review and omits minor details.

RATIONALE
---------

We used to lower forall statements immediately in the parser.
I believe the motivation for doing it that way originally was
to minimize the set of AST constructs for the compiler to deal wtih.

Recently we have been doing more things with forall loops,
such as forall intents. Having forall loops lowered before anything else
forced us to do pattern matching against the AST to access the information
about the forall loops that we need. This was costly in development time
and resulted in fragile compiler code.

Representing a forall loop with a ForallStmt node is a more direct
representation of the program and lets us access more information
about the forall loop easily.

IMPLEMENTATION HIGHLIGHTS - WORKFLOW

The handling of a forall loop still follows most of the ingredients
of the old buildForallLoopStmt. The modifications are:

* During parsing, very minimal lowering for iterators and
forall intents is made.

* Dispatching to standalone vs. leader/follower is now done
in the compiler during resolution, by resolving direct calls
to iteratorName(tag=standalone) or, if that does not resolve, to
iteratorName(tag=leader). See findStandaloneOrLeader().
(The remainder is done in lowerForallStmts(), see below.)

  Once done with that, I establish the type yielded by the iterator.

  This replaces using tryToken; _toStandalone/_toLeader functions in
internal modules; PRIM_TO_STANDALONE/PRIM_TO_LEADER in the compiler.

  This brought up bugs in some standalone iterators, which used to be
masked by tryToken. In those cases leader/follower were silently chosen
instead of the intended standalone iterators. See below for details.

* Given the traversal order of for_exprs_postorder() / resolveBlockStmt(),
the above resolution of the parallel iterator is done when traversing
the iterable-expression of the forall loop, or the first of them
for a zippered forall. For that, I augmented resolveExpr() with calls to
resolveParallelIteratorAndForallIntents().

* At start of resolveParallelIteratorAndForallIntents() the iterable
expression(s) are (almost) what the user actually wrote. Specifically,
if it is a call, then only the subexpressions are normalized; the call
itself is preserved as-is. For that I added a test to insert_call_temps.

This gives us the opportunity to see at resolution what the user wrote.
In particular, if it is a call to an iterator, we can modify it to a call
to the corresponding standalone or leader. Otherwise we normalize i.e.
insert a call temp for that expression at this time and invoke these() on it.

This replaces the _getIterator functions in internal modules.
(The functions are still there because they are used in other contexts.)
I also needed to have the compiler redirect to _value when
the iterable expression is an array/domain. Before we used to do so
using these methods in _array/_domain:

    inline proc these() {
      return _value.these();
    }

* I also needed to prevent normalization of the reduce-intent
specification for reduce intents, see restoreRiSpecCall().
This is to support reduce intents of the form ReduceOp(InputType)
for specifying the input type of the reduction explicitly, see #4574.

* The follower loops are unaffected by this PR, allowing it to
be smaller in size and scope. Thankfully the follower loops are
orthogonal and can be dealt with separately. The key complexity
factor is the static+dynamic selection between fast and regular
followers. For that, we still have _toFollwer/_toFastFollower
and PRIM_TO_FOLLOWER (which can also do "to fast follower",
depending on the context). Zippering needs to be dealt with, too.

* The remainder of lowering of ForallStmt is done in lowerForallStmts().
This is in resolve() after most code has been "resolved"
i.e. after resolveUses(mainModule) and resolveFns(chpl_gen_main).
This sets up most of the iterator handling that used to be done
in buildForallLoopStmt during parsing.

  I lower ForallStmt completely and remove them from AST so that
downstream passes are unaffected. It makes sense to lower them later
in the compiler when we switch to a more high-level representation of
iterators in general, through resolution.

* resolveParallelIteratorAndForallIntents() invokes
implementForallIntentsNew(), which is a combination of
implementForallIntents1() and implementForallIntents2()
with simplifications afforded by a more direct representation.

IMPLEMENTATION HIGHLIGHTS - DATA STRUCTURES
-------------------------------------------

The ForallStmt node has fields, most notably, for:
- index variable(s) (fVar),
- iterable expression(s) (fIter),
- loop body (fBody).

* A big design question is where to place the DefExpr(s) for
the loop index variable(s). It would be great to store
them in a dedicated list, analogously to FnSymbol::formals.
I did not take it on. Instead I went for a simpler solution.

  I place them in a BlockStmt referenced by fVar. In that same
BlockStmt after those DefExprs I also place the BlockStmt with the
body of the original forall loop, referenced by fBody. That way
scoping works out.

  Currently there is always just one index variable whose DefExpr goes
into fVar initially. Having an enclosing BlockStmt ends up being
convenient because I can place scaffolding related to forall intents
into that same BlockStmt between the index var's DefExpr and the fBody.

  If the user program has a tuple of index variables,
those are detupled within fBody from the master index var.

  Todo: given the current design, it does not make sense to have
a separate field for fBody. I will probably replace that field
with an accessor method that fetches the last stmt of fVar.
Probably rename fVar as well - what would be a good name for it?

* For now ForallStmt reuses the existing ForallIntents class
to represent forall intents. (It is not an AST node.)
This is to reduce the scope of this change. Ideally
we want to switch to a new AST node. I list this
under Future Work below.

MODULE CHANGES
--------------

Fixed a couple of standalone iterators. Those errors used to be
masked due to tryToken as discussed above.

I had to use __primitive("method call resolves") in one standalone
iterator that redirects to another one so it is available only when
the redirectee is available. Currently canResolve() does not work
there because it loses param-ness of the 'tag' argument.

P.S. The change to modules/minimal/internal/ChapelBase.chpl
is not needed and is here unintentionally, I will revert it.

FUTURE WORK
-----------

* Convert the remaining cases where forall loops are created
to ForallStmt or other dedicated representation, e.g.:

  - forall expressions
  - promoted expressions
  - reduce expressions

* In propagateExtraLeaderArgs(), use a dedicated data structure
to indicate intents on individual variables. This is different
than the following item because of its localized use and
the need to carry more information.

* Switch the representation of forall intents to a new AST node.
This will blend better with our current representation and
traversals.

* When the parallel iterator is specialized for particular use/
forall intents, it is extended to yield shadow variables. We
need to switch to yielding those variables that correspond to
in intents by ref. Currently they are yielded by value, making
them iteration-private rather than task-private how they are
supposed to be. My only concern for this change is whether
performance will be affected.

* Work out user-defined reduction interface, as it will affect
how reduce intents are lowered.

* Implement partial reductions.

### DIFF
--- a/compiler/AST/AstLogger.cpp
+++ b/compiler/AST/AstLogger.cpp
@@ -130,6 +130,13 @@ void AstLogger::exitBlockStmt(BlockStmt* node) {
 void AstLogger::visitForallIntents(ForallIntents* clause) {
 }
 
+bool AstLogger::enterForallStmt(ForallStmt* node) {
+  return true;
+}
+
+void AstLogger::exitForallStmt(ForallStmt* node) {
+}
+
 bool AstLogger::enterWhileDoStmt(WhileDoStmt* node) {
   return true;
 }
--- a/compiler/AST/AstPrintDocs.cpp
+++ b/compiler/AST/AstPrintDocs.cpp
@@ -175,6 +175,12 @@ bool AstPrintDocs::enterBlockStmt(BlockStmt* node) {
   return node->parentExpr == NULL;
 }
 
+
+bool AstPrintDocs::enterForallStmt(ForallStmt* node) {
+  return false;
+}
+
+
 bool AstPrintDocs::enterWhileDoStmt(WhileDoStmt* node) {
   return false;
 }
--- a/compiler/AST/AstVisitorTraverse.cpp
+++ b/compiler/AST/AstVisitorTraverse.cpp
@@ -178,6 +178,16 @@ void AstVisitorTraverse::visitForallIntents(ForallIntents* clause)
 
 }
 
+bool AstVisitorTraverse::enterForallStmt(ForallStmt* node)
+{
+  return true;
+}
+
+void AstVisitorTraverse::exitForallStmt(ForallStmt* node)
+{
+
+}
+
 bool AstVisitorTraverse::enterWhileDoStmt(WhileDoStmt* node)
 {
   return true;
--- a/compiler/AST/CollapseBlocks.cpp
+++ b/compiler/AST/CollapseBlocks.cpp
@@ -140,6 +140,10 @@ void CollapseBlocks::visitForallIntents(ForallIntents* clause) {
   INT_ASSERT(false);
 }
 
+bool CollapseBlocks::enterForallStmt(ForallStmt* node) {
+  return enterBlockStmt(node->fVar);
+}
+
 
 // The c for loop primitive is of the form:
 //   __primitive("C for loop", {inits}, {test}, {incrs})
@@ -333,6 +337,11 @@ void CollapseBlocks::visitUseStmt(UseStmt* node)
 }
 
 void CollapseBlocks::exitBlockStmt(BlockStmt* node)
+{
+
+}
+
+void CollapseBlocks::exitForallStmt(ForallStmt* node)
 {
 
 }
--- a/compiler/AST/baseAST.cpp
+++ b/compiler/AST/baseAST.cpp
@@ -76,8 +76,8 @@ void printStatistics(const char* pass) {
 
   foreach_ast(decl_counters);
 
-  int nStmt = nCondStmt + nBlockStmt + nGotoStmt + nUseStmt + nTryStmt;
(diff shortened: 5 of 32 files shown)


mppf:
I think the iterable expression in the forallstmt could be normalized and you'd have to only make minor modifications to how you handle it in resolution. I'm not sure about the value of sometimes not normalizing certain things - I might prefer to move normalization as a separate effort.

PR #5257, opened 2017-01-26
Fix parse slowdown from #5160

mppf:
PR #5160 added a linear traversal looking for the epilogue label when insertBeforeEpilogue is used. It turns out that this was being used during parse-time when adding string literals to the initStringLiterals function. However that case is easy to address by computing the epilogue label for initStringLiterals only once and storing it into a global variable.

If we run into more difficult sources of slow-down in the compiler as a result of the linear traversal, we should investigate making the epilogue label symbol a field in FnSymbol and that would allow this change to be removed.

Passed full local testing. Passed hellos with GASNet.


vasslitvinov:
Looks fine to me.

PR #5258, opened 2017-01-26
Don't change "this" into a reference if it is a type variable

daviditen:
Avoid turning the "this" argument in methods into a reference if it is a
type variable.

This fixes the problem in classes/diten/typeMethodDefaultArg.future.


bradcray:
Hi @daviditen : This looks good to me, thanks for catching and fixing it!

bradcray:
(I'm assuming you've done a full linux64 test run.  Might also be interesting to do a --futures-only run to see if there are other futures already filed against this.  You might be able to focus such a run on the directories where you, me, and Michael have filed bugs against type methods).

daviditen:
OK, I just merged master into this branch and will do a new full test run as well as the --futures-only run.

daviditen:
Linux64 paratest is finished; no errors.
For the --futures-only run I didn't find any newly passing .futures.  I'm getting a .bad mismatch for distributions/bradc/blkCyc/blkCycSliceOOB, but I also get it with this change backed out.

daviditen:
The .bad mismatch is related to PR #5229 .

bradcray:
Sounds good, thanks David!

PR #5262, opened 2017-01-27
Introduce a _new "type method" to wrap every class init method

noakesmichael:
Currently

var x  = new MyClass(10, 20, 30);

is converted to inline code allocate an instance of MyClass followed by a call to the appropriate
init method.  This method receives the new instance and also returns it.

This PR

1) creates a new "type method" for every init method of a class.  The type method allocates an
instance and then invokes the init method on the fresh memory.

2) converts the init method to have a void return

3) Adds the inline flag to the init method so that it is used efficiently by the only method that
could invoke it.

4) Updates resolution for new expressions for classes that have init methods to use the
new type method.

Ran paratest on linux64-single-locale, linux64-gasnet.

Also ran with -futures for classes/initializers

### DIFF
--- a/test/classes/initializers/noReturnInInit.bad
+++ b/test/classes/initializers/noReturnInInit.bad
@@ -1,6 +1,2 @@
-stretch tree of depth 11	 check: -1
-2048	 trees of depth 4	 check: -2048
-512	 trees of depth 6	 check: -512
-128	 trees of depth 8	 check: -128
-32	 trees of depth 10	 check: -32
-long lived tree of depth 10	 check: -1
+noReturnInInit.chpl:82: In initializer:
+noReturnInInit.chpl:84: error: illegal cast from Tree to void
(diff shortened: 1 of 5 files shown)


test/classes/initializers/noReturnInInit.bad
```diff
@@ -1,6 +1,2 @@
-stretch tree of depth 11	 check: -1
-2048	 trees of depth 4	 check: -2048
-512	 trees of depth 6	 check: -512
-128	 trees of depth 8	 check: -128
-32	 trees of depth 10	 check: -32
-long lived tree of depth 10	 check: -1
+noReturnInInit.chpl:82: In initializer:
+noReturnInInit.chpl:84: error: illegal cast from Tree to void
```
lydia-duncan: But doesn't have to happen this PR

lydia-duncan:
Looks good!

PR #5264, opened 2017-01-27
 Update BlockCyclic .bad file

benharsh:
This .bad file's output didn't really make sense. The test writes to an index out of bounds, yet the written value (1.0) shows up when printing the whole array. After #5229, the indexing logic is correct and the output is all zeroes.


benharsh:
@daviditen , @bradcray : here you go!

daviditen:
That looks fine to me.  Thanks!

Issue #5238, opened 2017-01-25
where clauses are ignored on non-generic functions


bradcray:
# Bug Report

### Summary of Problem

At present, where clauses are completely ignored on non-generic functions.  I think the original reason for this is that we thought that there was no reason to support them on non-generic functions since the single function signature didn't result in multiple versions of the function being instantiated.  But this is odd for a few reasons:

1) If you have a where clause that's completely nonsensical (e.g., calls a function that does not exist), the compiler completely ignores it.

2) Where clauses on non-generic functions could be useful as a means of completely squashing an overload of a function for a given compile -- e.g., by basing the where clause on something other than the arguments like a config param or type.

For these reasons, I'm suggesting that we evaluate where clauses on non-generic functions.

### Steps to Reproduce

**Source Code:**

case 1 (bar() should not resolve b/c it doesn't exist):

```chapel
proc foo(x: int) where bar() { // note that there is no bar()
  writeln("In foo()");
}

foo(1);
```

case 2 (foo() should not resolve because permitfoo is false):

```chapel
config param permitfoo = false;

proc foo() where permitfoo {
  writeln("In foo");
}

foo();
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
`./a.out`

### Configuration Information

- Output of `chpl --version`: 1.14.0.a771ec7
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

Issue #5237, opened 2017-01-25
where clauses on generic functions are evaluated too early


bradcray:
### Summary of Problem

The compiler currently evaluates where clauses on generic functions too early.  Specifically, it evaluates a generic function's where clause even if the function is not a candidate that can be called.  This results in errors that should never occur and is one of the root causes of issue #5236 .  It seems that a function's where clause should only be evaluated if the actual arguments would work for the function in question.

### Steps to Reproduce

**Source Code:**

```chapel
proc foo(param x: uint(8)) where bar() {
  writeln("In foo()");
}

proc bar() {
  compilerError("should not evaluate this");
}

foo(1.2);
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
`./a.out`

### Configuration Information

- Output of `chpl --version`: 1.14.0.a771ec7
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

Issue #5236, opened 2017-01-25
internal error: CAL0057


bradcray:
### Summary of Problem

In evaluating where clauses, we sometimes insert a cast optimistically, hoping that it will be folded back out again.  However, if it isn't folded out, we get an assertion/internal error in callInfo.cpp.

### Steps to Reproduce

**Source Code:**

```chapel
proc foo(param x: uint(8)) where x == 0 {
  writeln("In foo()");
}

foo(1.2);
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
`./a.out`

### Configuration Information

- Output of `chpl --version`: 1.14.0.a771ec7
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

Issue #5235, opened 2017-01-25
comparisons between uint and 0 aren't folded by the compiler


bradcray:
### Summary of Problem

Given a comparison between a uint and 0 that is known to always be true by definition such as 'a > 0' or 'a >= 0', I believe the compiler should param fold this conditional as it does with other obvious facts.  It doesn't currently at present because the overload that handles this case returns 'true' along one branch (when the other arg is a param 0) and calls the execution time comparison in the other branch; so the routine can't be a param routine and can't be folded.

A potential fix is to break the routine into two using a where clause to disambiguate and making the obvious case a param function and the other case not.  But this runs into bugs in our handling of where clauses which is another issue.

### Steps to Reproduce

**Source Code:**

```chapel
config var x: uint;

if (x < 0) then
  compilerError("This should never happen");
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
`./a.out`

### Configuration Information

- Output of `chpl --version`: 1.14.0.a771ec7
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

bradcray:
PR #5277 makes an attempt to fix this issue using the new where clause fixes introduced in PR #5275 but is currently failing with three tests.

mppf:
PR #11570 has some related changes. See https://github.com/chapel-lang/chapel/pull/11570#discussion_r230961816

mppf:
Here is a link to the test we'd like to get to pass for this issue to close: https://github.com/chapel-lang/chapel/pull/5277/files#diff-c60cb09d88210e77db2f5c34662d065d

vasslitvinov:
#25529 implements this specifically for `<`.

I am leaving this issue open as a request for similar changes to other operators: `<=` `>=` `>` `==` `!=`.

bradcray:
I hate to look a gift horse in the mouth (particularly one that I requested), but I wonder whether it's worse to support thse foldings only for `<` without adding the symmetric support to `>`, `<`, `>=`, and `>=`...  (just from a non-orthogonality perspective)

vasslitvinov:
@bradcray I do not have an answer to the "is it worse" question. My thinking is (a) if we like what I did in #25529 then we can do the same for the other operators, and (b) feel free to back out #25529 pending (a), if desired.

[...]

PR #5275, opened 2017-01-28
Resolve wheres later for generics and for instantiated generics

bradcray:
This change fixes a few issues with where clauses and files a number of tests (some working, some futures) against where-clause-related issues.

A few things that we have been doing with where clauses that do not seem quite right:

1) We evaluate where clauses for generic functions prior to determining whether or not that generic function is actual a match for the callsite.  This sometimes results in generating errors for a where clause (due to, say, type mismatches in where clauses that were never intended to fire for that routine) that it shouldn't (issue #5237).

2) We insert casts into where clauses a bit haphazardly, expecting them to be folded away, but when they can't be, we hit internal errors later in the compiler (issue #5236).

3) We silently ignore where clauses for concrete functions (issue #5238).

This change fixes the first two issues and almost fixes the third, but doesn't quite pass testing (ongoing work).  The theory is that when we stop ignoring where clauses for concrete functions, we run into some concrete (compiler-generated?) functions that have bogus where clauses that mess things up.

The approach taken here is to evaluate where clauses along both the generic and concrete resolution paths, but to have the generic path conservatively not generate errors, since we don't yet know whether or not the function will be a candidate.  Once generic functions are instantiated, they are re-resolved on the concrete path, at which point we have a second opportunity to inspect their where clauses.  To that end, we resolve the where clause on the concrete path after we've filtered the candidates down to those that are definitely callable.  This prevents us from triggering the style of issue listed in item 1 below.  On the concrete path, we currently only look at where clauses for instantiated generic functions to avoid hitting the "bogus where clause" case in item 3 above.  However, once such bogus where clauses are removed, all functions could have their where clauses evaluated along this path.

For item 2, the change was to insert the cast as before, and then try to fold the where clause.  In the event that the where clause did not fold, we remove the cast, and process the where clause as we would along the traditional path.  This got rid of the internal error by no longer leaving a non-folded cast call in the where clause where it was tripping up downstream code.

A number of the tests added here did not work on the master branch, but work now.  The primary category of test that continues not to work are concrete functions with where clauses.  In addition, there are some futures in which even the early evaluation of the where clause is more than we'd like to do because it could lead to triggering compiler errors, say.  Unfortunately, my attempts to completely eliminate where clause evaluation from the generic path were not successful.

This PR should also permit commit 4b451db77fb36cdee6f9d414c699883585fd3220 to be merged which will (a) fold uint/0 comparisons at compile-time (issue #5235) and (b) support a more principled solution to the uint **= uint bug (issue #5234).


bradcray:
Put this up for joint work with @ronawho -- closing for now til we land it better.

bradcray:
@ronawho: Just pushed the "evaluate where clauses for instantiated concrete functions only" conditional that I just mailed you about.  Firing off testing now.

bradcray:
[Summary: #Successes = 6898 | #Failures = 3 | #Futures = 0]

The three failures are three new futures that rely on resolving where clauses for non-instantiated concrete functions, so this is a clean run -- just have to re-futurize those until we determine which concrete functions' where clauses are blowing up.

bradcray:
As a note to posterity: @ronawho has begun the process of looking into concrete functions that have where clauses and found some cases in compiler-created promotion iterators that do.  This correlates strongly to the 200-some failures I saw when evaluating where clauses to all concrete functions -- lots of failures in tests using promotion.

PR #5276, opened 2017-01-28
Fix enum param compare

bradcray:
While looking at enum param comparisons with @ronawho this afternoon, we noticed that it seems to be comparing enums based on their integer values rather than their symbols.  This means, for example that "blue == monday" could be true where blue and monday are two completely different enum types.  Should it be?  This PR assumed the answer was "no" and tried to fix it....

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -7227,14 +7227,19 @@ static void foldEnumOp(int op, EnumSymbol *e1, EnumSymbol *e2, Immediate *imm) {
   type2 = toEnumType(e2->type);
   INT_ASSERT(type1 && type2);
 
-  // If our two enum types are different and we're doing == or !=, the
-  // answer should always be false, not based on the two enums'
-  // integer values.
+  // If our two enum types are different and we're doing == (or !=),
+  // the answer should always be false (or true), not based on the two
+  // enums' integer values.  If the two enum types are the same, we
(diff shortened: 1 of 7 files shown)


cassella:
Just to kibbitz through,

Why do you want a comparison between enums of different types to
compile in the first place if you don't want to compare their integral
values?

Or, why do you want to compare them numerically with ``<`` and ``>`` if you
don't want to compare them with ``==``?

Or from another angle, does it make sense to have ``(a <= b) && (a >= b)``
but at the same time not have ``(a == b)``?


I'd think that having ``(blue <= monday) && !(blue < monday)`` and not
having ``(blue == monday)`` would imply some kind of new order.

bradcray:
Yeah, that's where I ended up as well which is why I marked this as "language" and "design" (and should've marked it as "work in progress."  I thought I'd indicated better than I don't feel convinced this should be merged, but was working on getting back to the real task at hand (this was a sidebar that we ran into along the way), so left things dangling a bit...  I didn't have time to file an issue against it and was worried if I didn't file the PR I'd completely forget about it.

A compile error for comparing enums of different types with == or != makes sense to me as well -- then a user could insert casts if they really wanted to compare them.  This would still support the more common case of enum == int because the enum would coerce to int.

Mostly, I wanted to go back and see what C and other languages do while taking more time to think about it.

(Good new order joke... I'm shellshocked that I inadvertantly set you up for it like that)

PR #5277, opened 2017-01-28
Param fold uint comparisons with 0

bradcray:
It turns out that some clearly 'param' comparisons on uints weren't
folded at compile-time because the other branch of the conditional
required execution-time computation.  By splitting these param
cases out into their own param overloads, the compiler can fold
things better.  The "else" cases that were there before can go away
since the non-param overloads will now catch them and they can't
be evaluated at compile-time anyway.   This PR also adds a test to
ensure that these conditionals are folded as expected.

This PR is dependent on #5275 because without it, we try to evaluate
the where clauses on floats even though the function should not be
a candidate for such cases.  It also turned out to be dependent on
#5292 and/or #5345 to avoid evaluating where clauses incorrectly.


bradcray:
When merged with master, this currently fails for three test cases:

[Error matching compiler output for types/enum/diten/paramenums]
[Error matching compiler output for types/enum/sungeun/test_foldEnumOp]
[Error matching compiler output for types/string/diten/foldStringCompare]

ronawho:
@bradcray your regressions should have been resolved with #5345. This PR looks good to me, will make one minor comment inline

modules/internal/ChapelBase.chpl
```diff
@@ -1496,8 +1496,11 @@ module ChapelBase {
   inline proc >(a: uint(64), param b: uint(64)) {
     return __primitive(">", a, b);
   }
+  inline proc >(param a: uint(?w), b: uint(w)) param where a == 0 {
```
ronawho: Seems a little inconsistent  that you use `uint(?w)` while the other versions use `uint(64)` and rely on coercion. Is there any reason to not keep the signatures consistent?

bradcray: I did this somewhat conservatively, concerned that if we were comparing 8-bit values, the compiler might dispatch to the non-param versions and not fold the conditionals.  But granted, that was essentially paranoia.

Looking at the diff above with fresh eyes, it looks to me like we should remove the >(non-param, param) overload that doesn't fold since it doesn't make use of the param-ness of b and is presumably redundant with the >(non-param, non-param version).  Do you see any reason to keep it?

ronawho: I don't see any reason to keep it

bradcray: I did try changing to uint(64)s on these overloads and, as feared, my new test didn't pass -- I believe because comparisons on smaller uint types are caught by some of the smaller int overloads first.

test/expressions/comparisons/paramUintCompares.chpl
```diff
@@ -0,0 +1,45 @@
...
+testit(uint);
+testit(uint(64));
+testit(uint(32));
+testit(uint(16));
+testit(uint(8));
+
+//
+// This isn't actually an error, but avoids paying any execution time
+//
+compilerError("Everything's OK!");
```
ronawho: Seems weird to try to avoid execution time cost, when it'll be far, far faster than compilation times. If you want to skip execution, maybe use a .noexec instead?

bradcray:
The requirement for the explicit cast in Math.chpl is annoying to me, so I'm going to continue working on a more general version of this that folds all comparators (==, !=, <, <=, >, >=) for all param/param and param/non-param cases that can be using a smaller number of overloads.  I've got it planned out in my head, but am going to continue to prioritize array maps during the workday.

bradcray:
Closing this for the time being, having captured it in issue #5936.

Issue #5178, opened 2017-01-17
domain.shape && array.shape


ben-albrecht:
I propose to add a parenthesis-less `.shape` method to domains and arrays, which returns a tuple of the number of elements per dimension. This is particularly useful in linear algebra codes where checking against matrix dimensions is a common test.

The feature is inspired by NumPy's [ndarray.shape](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.shape.html#numpy.ndarray.shape) member.

An implementation and usage might look like the following:

```chapel

  /* Return shape of domain as a tuple */
  proc _domain.shape {
    var shape: this.rank*(int);
    for (i, r) in zip(1..shape.size, this.dims()) do
      shape(i) = r.size;
    return shape;
  }

  /* Return shape of array as a tuple */
  proc _array.shape {
    return this.domain.shape;
  }


// Usage example:

var A41: [{1..4, 1..1}] real,
    A14: [{1..1, 1..4}] real,
    A1:  [{1..1}] real,
    A44: [{1..4, 1..4}] real;

writeln(A41.shape); // outputs: (4, 1)
writeln(A14.shape); // outputs: (1, 4)
writeln(A1.shape);  // outputs: (1)
writeln(A44.shape); // outputs: (4, 4)
writeln(A41.domain.shape) // outputs: (4, 1)
writeln(A14.domain.shape); // outputs: (1, 4)
writeln(A1.domain.shape);  // outputs: (1)
writeln(A44.domain.shape); // outputs: (4, 4)
```

### Details
(some discussed in comments below)

- Sparse arrays and domains return the shape of their parent domain
- `.shape` should always refer to the entire array/domain (locale-independent)
- Associative arrays and domains behave in a similar way as 1D arrays
- To be implemented using `where` clauses to designate array and domain variations 
  - e.g. `where isAssociativeArr(this)`
  - a `where`-less  method will return a compiler error to catch any unimplemented cases.

benharsh:
Would these only be supported on rectangular arrays (e.g., default, block, sparse)? If not, what would they mean for something like an associative array?

bradcray:
Can _array.shape simply 'return this.domain.shape' to reduce future code maintenance?

ben-albrecht:
@benharsh -- It should have the same behavior for sparse arrays and all distributions. I don't think associative arrays have a well-defined shape and should just throw a compiler error.

ben-albrecht:
[...]

PR #5284, opened 2017-01-30
domain.shape && array.shape

ben-albrecht:
This PR adds the `domain.shape` and `array.shape` methods that return that shape of an array or domain in the form of a tuple, originally proposed and discussed in #5178. See the issue for further details.

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -1078,6 +1078,37 @@ module ChapelArray {
       for i in _value.dimIter(d, ind) do yield i;
     }
 
+   /* Returns a tuple of integers describing the size of each dimension.
+      For a sparse domain, returns the shape of the parent domain.*/
+    proc shape where isRectangularDom(this) {
+      var shape: rank*(int);
+      for (i, r) in zip(1..shape.size, dims()) do
+        shape(i) = r.size;
+      return shape;
+    }
+
+    pragma "no doc"
+    proc shape where isSparseDom(this) {
+      var shape: this._value.parentDom.rank*(int);
+      for (i, r) in zip(1..shape.size, this._value.parentDom.dims()) do
+        shape(i) = r.size;
+      return shape;
+    }
+
+    pragma "no doc"
+    proc shape where isOpaqueDom(this) || isAssociativeDom(this) { 
+      // Opaque domains assumed 1D
+      var shape: (int,);
+      shape[1] = size;
+      return shape;
+    }
+
+    pragma "no doc"
+    /* Unsupported case */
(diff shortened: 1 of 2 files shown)


modules/internal/ChapelArray.chpl
```diff
@@ -1078,6 +1078,37 @@ module ChapelArray {
       for i in _value.dimIter(d, ind) do yield i;
     }
 
+   /* Returns a tuple of integers describing the size of each dimension.
+      For a sparse domain, returns the shape of the parent domain.*/
+    proc shape where isRectangularDom(this) {
+      var shape: rank*(int);
```
mppf: As a rule, I wouldn't personally declare a variable with the same name as the function it's enclosed in. Nonetheless if it works that's just a style issue.

modules/internal/ChapelArray.chpl
```diff
@@ -1078,6 +1078,37 @@ module ChapelArray {
...
+      for (i, r) in zip(1..shape.size, dims()) do
+        shape(i) = r.size;
+      return shape;
+    }
+
+    pragma "no doc"
+    proc shape where isSparseDom(this) {
+      var shape: this._value.parentDom.rank*(int);
+      for (i, r) in zip(1..shape.size, this._value.parentDom.dims()) do
+        shape(i) = r.size;
```
mppf: I think the sparse version can be the same as the Rectangular one since dims() for a sparse domain returns the dimensions of the parent domain.

modules/internal/ChapelArray.chpl
```diff
@@ -1078,6 +1078,37 @@ module ChapelArray {
...
+      return shape;
+    }
+
+    pragma "no doc"
+    proc shape where isAssociativeDom(this) || isOpaqueDom(this) {
+      // Opaque domains assumed 1D
```
mppf: comment should say Opaque and Associative domains are assumed to be 1-D

mppf:
I left minor comments, but otherwise it looks good. I'm going to assume from the issue and discussion there that the answer to the question *should we add this?* is *yes* and you're just asking me to review the implementation.

modules/internal/ChapelArray.chpl
ben-albrecht: Ah good catch. I did not realize this.

ben-albrecht:
@mppf - thanks for review. Changes made.

>I'm going to assume from the issue and discussion there that the answer to the question should we add this? is yes and you're just asking me to review the implementation.

Correct. I gathered  positive feedback on the idea prior to opening #5178, and there appears to be no push back on the feature there either.

PR #4992, opened 2016-12-08
Next step towards using SymbolSymExprs more

mppf:
This PR adjusts bulidDefUseMaps and compute_call_sites to use SymbolSymExprs. It is an intermediate step. The next step is to entirely remove the vectors/sets associated with these functions and instead compute them on-the-fly. (Along with adding some helper functions to make that easy).

Passed full local testing.
Passed quickstart+verify testing.

Reviewed by @lydia-duncan - thanks!

PR #5285, opened 2017-01-31
Loop over Defs and Uses with SymbolSymExprs

mppf:
Continuing PR #4992, this PR uses SymbolSymExprs in more places in the compiler.  In particular, it adds loop macros for_SymbolDefs and for_SymbolUses which can be used to replace for_defs and for_uses. Additionally, it provides some convenient methods on the Symbol class: countDefs, countUses, isUsed, isDefined, getSingleUse, getSingleDef.

Then, this PR updates 7 files to use the new functionality and removes 11 calls to buildDefUseMaps in the process. The new code has lower complexity and some basic experimentation shows about a 5% compile time improvement for hello4-datapar-dist.chpl on my system.

About 25 uses of  buildDefUseMaps remain. The main tricky thing to watch out for when replacing the old pattern with the new one is in loops that add AST while iterating over uses or defs. With for_defs/for_uses, the new AST will not be looped over, since the def/use lists and maps have already been computed. With for_SymbolDefs or for_SymbolUses, the new AST can be looped over.

I personally think that a compiler transformations of the style
 * figure out what to modify and add it to a work queue / list / vector
 * do the modification where appropriate

are generally easier to reason about. Such transformations can work with for_SymbolDefs/Uses or for_defs/uses without any issue.

Passed full local testing.
Passed hellos with --baseline --verify;  with --no-local; and with GASNet.

### DIFF
--- a/compiler/AST/astutil.cpp
+++ b/compiler/AST/astutil.cpp
@@ -359,6 +359,9 @@ bool isRelationalOperator(CallExpr* call) {
 // return & 1 is true if se is a def
 // return & 2 is true if se is a use
 //
+// Note that a DefExpr is where we hang the variable declaration, but after
+// normalize, a DefExpr itself does not set a variable, and so it does not count
+// as a Def.
 int isDefAndOrUse(SymExpr* se) {
   if (CallExpr* call = toCallExpr(se->parentExpr)) {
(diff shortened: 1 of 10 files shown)


benharsh:
Looks fine to me. I think we should keep an eye on compile-time performance graphs once this goes in (hopefully it's all beneficial).

PR #5288, opened 2017-02-01
Fix problem with addAllToVirtualMaps

mppf:
The new test, generic-inherit-bug1, shows a case where the virtual dispatch tables were not being correctly set up. I encountered this issue when working on improving domain assignment for irregular domains.

In particular, the problem came up for generic classes where there is a class with no declarations between the base class defining some methods and a child class overriding it.

e.g.
```
  class GrandParent {
    type t;
    proc foo() {
      writeln("GrandParent.foo()");
    }
  }

  class Parent : GrandParent {
  }

  class Child : Parent {
    proc foo() {
      writeln("Child.foo()");
    }
  }

  var x:Parent(int);
  x = new Child(int);
  x.foo();
```
should output Child.foo() but before this change outputs GrandParent.foo().

This commit solves the problem by removing a conditional in addAllToVirtualMaps. I cannot explain why that conditional should be there now. It was added in SVC commit 7117 (git id c8befa71e8).

Passed quickstart testing.
Passed full local testing with --futures.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -8875,10 +8875,8 @@ addAllToVirtualMaps(FnSymbol* fn, AggregateType* pct) {
          ct->defaultTypeConstructor->isResolved()))
       addToVirtualMaps(fn, ct);
 
-    // if sub-class is not a generic instantiation
-    if (!ct->instantiatedFrom)
-      // add to maps transitively
-      addAllToVirtualMaps(fn, ct);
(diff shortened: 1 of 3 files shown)


lydia-duncan:
Have you done additional testing with futures?  I think Engin ran into something that this also might fix, and I hope I made him file a bug on it

mppf:
I'll add a TO-DO to the PR to do futures testing in addition to qthreads testing.

lydia-duncan:
Awesome, thanks!  This looks good to go to me, otherwise :)

PR #5291, opened 2017-02-02
Delete free tests

mppf:
For now, record-wrappers for not needing to delete don't explicitly provide methods/coerce to the base pointer type. You have to use `get()` for that and we're discussing whether or not we want to enable coercions.

Reviewer Questions:

 * What should the names of the modules be?
 * What should the names of the records be?


mppf:
@vasslitvinov and/or @daviditen - could you have a look at this? Once we agree on names, I'll move the reference counting and owned wrapper records to modules/packages.

At the moment, I'm thinking of using these names:
 * module SharedObject, record Shared
 * module OwnedObject, record Owned

Other variants that come immediately to mind are SharedPtr, SharedPointer, SharedInstance. I'm leaning away from including the mechanism (ReferenceCounting) in the names.

vasslitvinov:
I am comfortable with `Shared` and `Owned`.

Shared/OwnedObject sound OK for module names.

daviditen:
`Shared` and `Owned` work for me too.  I don't like having Ptr/Pointer in the names - if anything it should be Instance or Class.  But `Shared` and `Owned` alone sounds better to me.

mppf:
@daviditen - how do you feel about SharedObject.chpl and record Shared ?

daviditen:
@mppf SharedObject.chpl sounds fine for the module/filename that `Shared` would be in.

mppf:
Closing this in favor of PR #5683 which adds them to modules/packages with the appropriate names. I will work on refining the interface there.

PR #5292, opened 2017-02-02
Stop ignoring where-clauses for non-generic functions

### LINKED ISSUE #5238: where clauses are ignored on non-generic functions

ronawho:
Evaluate and respect where-clauses for non-generic functions. This is a
follow-on to #5275, and closes #5238.

Previously the language definition stated that "The optional where-clause is
only applicable if the function is generic". The compiler didn't error if a
where-clause was attached to a non-generic function, it just never evaluated
it, effectively pretending the where-clause wasn't there.

You could write something like:

```chapel
proc foo(x: int) where bar() { // note that there is no bar()
  writeln("In foo()");
}
foo(1);
```

and the compiler would ignore the bogus `where bar()` and happily resolve foo()

Originally, we didn't see a use-case for where-clauses on non-generic
functions, which is why the spec disallowed it. We imagined that where-clauses
would only compute on aspects of the functions arguments, so if the function
was non-generic, there'd be no need for a where-clause. We've since come to
realize that it's more orthogonal to support where-clauses on non-generics and
that there are real use-cases for it such as selecting a function
implementation based off a config param:

```chapel
config param matrixLayout = row_major;

proc matrixOp(A, B) where matrixLayout == row_major { /* row-major impl */ }
proc matrixOp(A, B) where matrixLayout == col_major { /* column-major impl */ }
```

This fixes a promotion wrapper bug triggered by non-generic where-clauses,
updates resolution to evaluate where-clauses on non-generic functions, fixes
some tests with bogus where-clauses, retires some futures, adds some new tests,
and updates the spec to indicate that where-clauses are supported on
non-generic functions.


ronawho:
@bradcray feel like reviewing? 

The code changes are simple enough, I think I'm mostly looking for a review of the motivation and somebody to tell me what else I need to update in the spec. I think 22.4 -- "Where Expressions" might need some updates, or might need to be placed somewhere other than the generic section, but I wasn't really sure what to do there.

bradcray:
On it.

compiler/resolution/functionResolution.cpp
```diff
@@ -2584,18 +2584,11 @@ filterConcreteCandidate(Vec<ResolutionCandidate*>& candidates,
...
   //
-  // If the candidate is an instantiation of a generic function,
-  // re-evaluate the where clause here in order to make sure it
-  // evaluates to true and to catch errors in it.  Ultimately, we'd
-  // like to do this for all concrete functions (instantiated or not),
-  // but at present, enabling it for them causes problems, perhaps
-  // because we're generating bogus where clauses on concrete
-  // functions today and getting away with it since they're ignored.
+  // Evaluate non-generic where clauses (also re-evaluates where cluases for
+  // instantiated generic functions.)
```
bradcray: At one point, it sounded like you had a branch that could nearly remove the where clause evaluation from the generic branch, which continues to seem attractive to me.  Are you leaving that for a separate branch or dropping the effort on the floor?

ronawho: I'm still working on that, but it'll be in a separate PR

spec/Functions.tex
```diff
@@ -221,8 +221,7 @@ \section{Procedure Definitions}
 The \sntx{return-type} is optional and is discussed in~\rsec{Return_Types}.
 A type function may not specify a return type.
 
-The optional \sntx{where-clause} is only applicable if the function is
-generic.  It is discussed in~\rsec{Where_Expressions}.
+The optional \sntx{where-clause} is discussed in~\rsec{Where_Expressions}.
```
bradcray: I hate to say this since (a) it involves the spec and (b) you've taken this effort on from me, but I think that we need to move the current Where Expressions section from the generics chapter up to the Procedures chapter; or perhaps move most of it up to Procedures and have a section in Generics refining the concrete explanation.  You can push this back on me if you want.

bradcray:
Other than the comment on the spec change, this looks great to me -- thanks!

spec/Functions.tex
ronawho: I don't want to push back, this was the feedback I was looking for with regard to the spec. I was pretty sure we wanted to move `22.4 -- "Where Expressions"` somewhere else, but I wasn't familiar enough with the spec to know where.

bradcray: I'd move it into Functions.tex (the "Procedures" chapter).  It looks to me like the current organization _roughly_ follows the order of the syntactic elements in declaring a procedure, which would suggest putting the where clause section after the "Return Types" section and before the "Nested Functions" section (but double-check me).

bradcray:
Belatedly: test/functions/diten/whereClausesIgnored.chpl should be able to be de-futurized.  It was anticipating that this would be an error which is why it didn't pass under --futures testing.

ronawho:
Retired @daviditen's future, which led me to notice that we didn't have any tests with legal/valid where-clauses on non-generic functions so I added a couple of tests.

I moved the "Where Expressions" section after the return-type section since that's the same order things were listed in `13.2 Procedure Definitions` -- does that sound right?

I still have to update the actual text of the "Where Expressions" section to not be so specific to generic functions. Despite saying "I don't want to push it back to you", I have very little experience writing spec text, and wouldn't mind if you wanted to take that over. I'm also happy to give it a shot and let you/somebody else have a pass if you'd prefer.

vasslitvinov:
While I agree we want where clauses on non-generic functions, the atomic += example in the PR message does not give a good motivation. I can see that we can make += on atomics legal or not, I do not see the benefit of that choice - why would we ever want to disallow it?

Maybe put in the PR message another example along these lines:

    proc myfun() where implementationChoice == 1 { ... first implementation ... }
    proc myfun() where implementationChoice == 2 { ... second implementation ... }

ronawho:
@vasslitvinov fair point about the example not being great. I wanted an example that was at least a little concrete -- what do you think of:


```chapel
config param matrixStorage = row-major;

proc matrixOp where matrixStorage == row-major { /* row-major implementation */ }
proc matrixOp where matrixStorage == col-major { /* column-major implementation */ }
```

vasslitvinov:
@ronawho works for me! Please use underscore instead of minus in *-major.

ronawho:
@bradcray draft spec update added with https://github.com/chapel-lang/chapel/pull/5292/commits/c65193bdbbdd10c6f58d2d83993751606616431f

Thoughts?

spec/Functions.tex
```diff
@@ -1047,14 +1047,13 @@ \section{Where Expressions}
...
-instantiated, the expression in the where clause must be a parameter
-expression and must evaluate to either \chpl{true} or \chpl{false}.
-If it evaluates to \chpl{false}, the instantiation is rejected and the
-function is not a possible candidate for function resolution.
-Otherwise, the function is instantiated.
+The list of function candidates can be constrained by {\em where clauses}.  A
+where clause is specified in the definition of a
+function~(\rsec{Function_Definitions}).  When a function is resolved, the
+expression in the where clause must be a parameter expression and must
+evaluate to either \chpl{true} or \chpl{false}.  If it evaluates to
```
bradcray: wordsmithing, I might say either "a parameter function that evaluates to either true or false" or maybe simply "a boolean parameter expression".

spec/Functions.tex
```diff
@@ -1041,6 +1041,37 @@ \subsection{Implicit Return Types}
...
+\chpl{false}, the function is rejected and thus is not a possible candidate
+for function resolution.
+\begin{chapelexample}{whereClause.chpl}
+Given two overloaded function definitions
+\begin{chapel}
+proc foo(x) where x.type == int { writeln("int"); }
+proc foo(x) where x.type == real { writeln("real"); }
+\end{chapel}
```
bradcray: I haven't looked at the context enough to have a strong opinion on this:  is it weird to use a generic function example in this section (i.e., before the "Generics" chapter?)  I don't find the presence of it here inherently off-putting, but I have a pretty low bar when it comes to such things.  If it does seem weird, I might use a concrete function overload here -- say one based on whether or not a config param is true v. false and move this example back to the Generics chapter with a perfunctory section that essentially says something like "Where clauses are particularly useful in the context of generic functions since they can be used to reason about the argument types."  But again, I don't feel insistent on this if the current approach seems OK.

bradcray:
No other comments at this time other than to make sure to build the spec and read the sections that were modified in PDF form to make sure there aren't any surprises (LaTeX being such a surprise-free medium... :) ).

spec/Functions.tex
ronawho: I personally don't think it's weird, and there's plenty of other places where we use generic examples or talk about generics before formally introducing them in section 22.

bradcray: OK, just wanted to make sure this wasn't the only such case.  If we already do this, I like this approach better than the alternative I proposed and think this could go in either as-is, or taking the wordsmithing suggestion above.

spec/Functions.tex
ronawho: `When a function is resolved, the expression in the where clause must be a parameter expression and must evaluate to either \chpl{true} or \chpl{false}.`

=> 

`The where clause must contain a boolean parameter expression.`

or 

`When a function is resolved, the expression in the where clause must be a boolean parameter expression`

?

ronawho: Going with:

`The expression in the where clause must be a boolean parameter expression that evaluates to either "true" or "false"`

PR #5294, opened 2017-02-02
Add new files initializerResolution.h and .cpp and support them

lydia-duncan:
This involved moving some contents from initializerRules into
initializerResolution.  It also moved a lot of class and function declarations
into resolution.h, as my support for resolving initializers relies on them.

I suspect this may impact compilation time in the resolve pass, as this removes
the static declaration from roughly 15-20 functions.  We shall see.

### DIFF
--- a/compiler/include/initializerResolution.h
+++ b/compiler/include/initializerResolution.h
@@ -17,13 +17,23 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


lydia-duncan:
Passed std/ testing.  Going to merge without review, it is mostly reshuffling

Issue #5255, opened 2017-01-26
iterator forwarding


vasslitvinov:
We would like to have support for iterator forwarding so that

    for idx in bar(args...) do ....
    // or even
    forall idx in bar(args...) do ....

actually runs some other iterator foo(args...).

One idea is `yieldall`, e.g.

    iter bar() {
      yieldall foo();
    }

An open question with `yieldall` is: how to redirect parallel iterators?

Another option is to use delegation, esp. for an iterator method to forward to the same iterator in the class/record's field.

Ultimately we also want to return an iterator from a function as a first-class function, perhaps like this:

    iter bar() {
      return <capture the foo iterator>;
    }

Here the open questions include: how to capture foo() while passing it some arguments? how to specify the return type of bar()?

Note: currently our implementation relies on the following "feature" for: redirecting array/domain "these" iterators to "these" in their domain maps; for wrapper functions around iterators; in the Random module:

    // some iterator
    iter foo() { ....... }
    // bar() acts as an iterator that executes foo()
    proc bar() return foo();

We have also a few cases where we simply write for/forall loop to express the desired semantics directly. It works, however it feels a bit too verbose.

See also the email thread "iterator forwarding in Chapel? available!" on the internal list in Dec 2016.

vasslitvinov:
One aspect highlighted in #5295 is that we may forward an iterator -- or a procedure as well -- only when the target (forwardee) is available. Otherwise the forwarding function should be treated as non-existing.

A simple example is that the _domain/_array standalone iterator forwards to the standalone iterator on the underlying _value. If _value does not support the standalone iterator, then the compiler should act as if the standalone iterator on the _domain/_array does not exist and so try leader/follower instead. #5295 writes this pattern as follows:

```chapel
record _domain {
  ...
   inline proc these(param tag) where
        (tag == iterKind.leader || tag == iterKind.standalone) &&
        __primitive("method call resolves", _value, "these", tag=tag)
      return _value.these(tag=tag);
```

(This forwards both the standalone and the leader, as long as the forwardee is resolves OK.)

mppf:
I wonder if the `forwarding` syntax can do this?

vasslitvinov:
Consider, for example, this code:

```
test/users/engin/partial_reduction_support/modules/dsiMethods.chpl
```

It defines several sets of iterators (serial, leader, follower, standalone), most of which follow the same pattern:

[...]

PR #5295, opened 2017-02-02
introduce ForallStmt node

vasslitvinov:
This PR introduces the `ForallStmt` node - a subclass of `Stmt`,
to represent Chapel forall statements. See Rationale below.

Future work: #6797

STATUS
------

* Currently, ForallStmt nodes are created in the parser and are lowered away in `resolve`.

* This is the first step towards the target representation discussed in #6514.

* This PR replaces and extends #5249.

* This PR futurizes one test - see below.

* New features - see Bonus below.

Testing:
* linux64 --verify
* gasnet
* valgrind
* memleaks - only minor differences from master

This is a rebase of: 8f80a01..98b34d9 and 6f8e963
on Vass's fork.

OVERVIEW OF CHANGES
-------------------

* moved things around in ForallStmt.cpp

* ForallStmt::build() - for the parser to create forall loops
   (in ForallStmt.cpp)

* introduced parIdxVar(), parIdxCopyVar(), userLoop()
    they are used when lowering
    and earlier in resolution for forall intents
  addParIdxVarsAndRestruct() restructures things to help lowering

* lowerForallStmts() (in foralls.cpp)
    invoked mid-way during resolution
    removes ForallStmt nodes entirely

* chapel.ypp - uses ForallStmt::build() instead of buildForallLoopStmt()
  parser-generated files are currently not shown in this PR

* adjustments to normalize
    do not normalize riSpec
    also handleReduceAssign

* scopeResolve - how to resolve ForallStmt body

* cullOverReferences
    adjustments for the new representation

* resolve()
    call resolveParallelIteratorAndForallIntents()
      when resolving the first iterable
    call lowerForallStmts()
    preFold PRIM_REDUCE_ASSIGN

* implementForallIntentsNew()
    adjusted version of implementForallIntents1(),implementForallIntents2()
      and helper functions
    includes catering to iterator forwarders e.g.
      proc iterate(...) { ... return PCGRandomPrivate_iterate(...); }
      needed for standalone and leader iterators

* Chapel code
    add where-clauses to iterator forwarders
      so standalone forwarder is available only when forwardee is
    added standalone+leader iterator forwarders for _domain/_array.these()

RATIONALE
---------

We used to lower forall statements immediately in the parser.
I believe the original motivation for it was
to minimize the set of AST constructs for the compiler to deal wtih.

Recently we have been doing more things with forall loops,
such as forall intents. Having forall loops lowered before anything else
forced us to do pattern matching against the AST to access the information
about the forall loops that we need. This was costly in development time
and resulted in fragile compiler code.

Representing a forall loop with a ForallStmt node is a more direct
representation of the program and lets us access more information
about the forall loop easily.

IMPLEMENTATION HIGHLIGHTS - WORKFLOW

The handling of a forall loop still follows most of the ingredients
of the old buildForallLoopStmt. The modifications are:

* During parsing, very minimal lowering for iterators and
forall intents is made.

* There is always one index variable per iterable expression
in the zippered case. If the user did not write it that way
ex. forall i in zip(...) or forall ((a,b),(c,d)) in zip(...),
then the compiler will create them. If this is what the user
wrote, ex. forall (a,b) in zip(c,d), then the user variables
will be used.

In the non-zippered case, there has always been a compiler-
introduced single index variable. With this change, if there is
a user index variable, it will be used directly.

* Dispatching to standalone vs. leader/follower is now done
in the compiler during resolution, by resolving direct calls
to iteratorName(tag=standalone) or, if that does not resolve, to
iteratorName(tag=leader). See findStandaloneOrLeader().
(The remainder is done in lowerForallStmts(), see below.)

  Once done with that, I establish the type yielded by the iterator.

  This replaces using tryToken; _toStandalone/_toLeader functions in
internal modules; PRIM_TO_STANDALONE/PRIM_TO_LEADER in the compiler.
(tryToken is still used because implicit forall loops go through
the old implementation.)

  This had brought up bugs in some standalone iterators, which used to be
masked by tryToken. In those cases leader/follower were silently chosen
instead of the intended standalone iterators. See below for details.

* Given the traversal order of for_exprs_postorder() / resolveBlockStmt(),
the above resolution of the parallel iterator is done when traversing
the iterable-expression of the forall loop, or the first of them
for a zippered forall. For that, I augmented resolveExpr() a with call to
resolveParallelIteratorAndForallIntents().

* At start of resolveParallelIteratorAndForallIntents(), if the iterable
expression(s) are iterator calls, we get to the original iterator
call (the first of them if zippered) using findDefOfTemp().

  Cf. in #5249 I instead had an add-on to normalize/insert_call_temps()
that prevented these from being normalized. That way at resolution time
they remained (almost) what the user actually wrote. Specifically,
if it is a call, then only the subexpressions are normalized; the call
itself is preserved as-is. Michael suggested not doing that, leaving
our dealings with normalization as a separate matter.

  One way or another, we get hold on the iterator call as written by user.
So we can modify it to call the corresponding standalone or leader.

* I also needed to prevent normalization of the reduce-intent
specification for reduce intents, see restoreRiSpecCall().
This is to support reduce intents of the form ReduceOp(InputType)
for specifying the input type of the reduction explicitly, see #4574.

* The follower loops are unaffected by this PR, allowing the PR to
be smaller in size and scope. Thankfully the follower loops are
orthogonal from the parallel machinery and can be dealt with separately.
The key complexity factor with follower loops is the static+dynamic
selection between fast and regular followers. For that, we still have
_toFollwer/_toFastFollower and PRIM_TO_FOLLOWER (which can also do "to
fast follower", depending on the context). Zippering needs to be dealt
with, too.

* The remainder of lowering of ForallStmt is done in lowerForallStmts().
This is in resolve() after most code has been "resolved"
i.e. after resolveUses(mainModule) and resolveFns(chpl_gen_main).
This sets up most of the iterator handling that used to be done
in buildForallLoopStmt during parsing.

  I lower ForallStmt completely and remove them from AST so that
downstream passes are unaffected. We plan to push lowering of ForallStmt
to a later point in compilation to preserve valuable information for
later passes.

* resolveParallelIteratorAndForallIntents() invokes
implementForallIntentsNew(), which is a combination of
implementForallIntents1() and implementForallIntents2()
with simplifications afforded by a more direct representation.

KEY IMPLEMENTATION CHOICES
--------------------------

* DefExprs for index variables are placed on the AList
ForallStmt::inductionVariables(). My concern was that
they are outside the normal BlockStmt scoping structure.
Surprisingly, to get it to work, I needed only a minor adjustment
in scopeResolve.

  Unfortunately I needed to move those DefExprs into the user loop body
while resolving the ForallStmt, which is earlier than desirable.
That's when the user loop body gets duplicated and dissolves into
the fast-vs.-normal follower machinery and it would be hard to do so
after that.

* For now ForallStmt reuses the existing ForallIntents class
to represent forall intents. (Which is not an AST node.)
This is to reduce the scope of this change. We will switch
representing forall intents with AST nodes.

* Given that ForallStmt is a kind of loop, we may want to call it
ForallLoop and/or have it inherit from LoopStmt.

  Right now ForallStmt cannot inherit from LoopStmt because
that would make it inherit transitively from BlockStmt,
and that is not the right relationship.

MODULE CHANGES
--------------

Testing exposed several bugs in standalone iterators that
resulted in resolution failures - which would be masked by
"if tryToken" used to choose standalone vs. leader -
which would result in silently choosing the leader iterator
over standalone. Those bugs I believe got fixed on master
prior to this change.

I added the check __primitive("method call resolves") to a few standalone
iterators that redirect to others. That way a given standalone iterator
is available only when the redirectee is available. Without this check
there would be a resolution error, when all that's needed is to use
leader, not standalone.
Currently I have to use __primitive - can't use canResolve()
because it loses param-ness of the 'tag' argument.

BTW there is an inconsistency in these cases - sometimes the 'tag'
argument is passed as a named actual, sometimes not.
(This had been so prior to this change.)

BONUS FEATURES
--------------

The user has the option of passing an instance of a ReduceOp class
into a reduce intent, rather than specifying just the name
of the ReduceOp class. See

    test/parallel/forall/vass/ri-6-user-instance.chpl

This is not yet available for `reduce` expressions because
those are still handled the same way as before.

Also, a parallel iterator can now recurse via a forall loop. See

    test/functions/iterators/bradc/recursiveStandalone.chpl
    test/functions/iterators/diten/recursiveLeader.chpl
    test/functions/iterators/tzakian/par_iter_recursive.chpl

FUTURIZED TEST
---------------

This PR futurizes this tests:

    test/parallel/forall/in-intents/both-arr-dom-const-const.chpl

This PR makes this test not capture an in-intented array,
passing it by reference instead.
It worked before this PR "for a wrong reason". That's because
the array is passed by default intent to `_toStandalone()`, which
stashes it into an iterator record then invokes chpl__autoCopy()
on the _ir_. So the array gets deep-copied as a field or this _ir_.
To fix this, I would like to rework how we pass to the parallel iterator
the arguments that correspond to forall-intent variables.
This is a fairly independent effort, so I would like
to address it separately from this PR.


compiler/include/stmt.h
```diff
@@ -275,6 +276,46 @@ class GotoStmt : public Stmt {
...
+  // All pointer fields are always non-null.
+  BlockStmt*     fsBlock;   // contains defs for index variable(s) and fBody
+  AList          fIter;     // iterable expression(s)
+  bool           fZippered;
+  ForallIntents* fWith;     // todo: replace with AList of per-variable "clauses"
+
+  // derived from the above
+  VarSymbol* fIdxVar()     const;
+  VarSymbol* fIdxCopyVar() const;
+  BlockStmt* fBody()       const; // forall loop body
```
mppf: For other work I am doing (ref-if-modified), it would be really useful to be able to go from index variable to iterator variable. I wonder if there should be a method to access that information?

I also find it a little odd that these methods start with 'f'. I'd expect the ForallStmt to have fields internal to the implementation, such as _block, _iter, _zippered, _with; and then to have methods with names like indexVariable() etc. to access them.

Also, it seems to me that we'd like to have ways of accessing particular index variables or iterators when doing zippered iteration. I wonder if we should have:

 * get number of zippered iterables
 * get index variable i
 * get iterable expression i

mppf: Also, I'd like this information to still work after the Forall statements are "lowered" in resolution. But that is something we can discuss if it's difficult.

mppf:
Hi @vasslitvinov - I'm experimenting with this branch with a zippered forall over 1-D arrays.

It seems odd to me that in the AST dump I'm seeing variables like chpl__saIter for the Leader iterator ForLoop. Is that just a matter of confusing variable names, or is there a deeper problem here?

vasslitvinov:
@mppf Good point - it's just variable naming. I copied that code from build.cpp so wanted the new code to be as close to the original as possible.

Do you think that such close-ness is less valuable than clear variable names? I will probably call them chpl_parIter/iterPAR etc.

mppf:
Yes I think it'd be better to have clear variable names. Also, the pain I'm personally suffering with Forall is all after resolution. Do you think it'd be possible to keep around the ForallStmt AST nodes as a map to the implementation, even after it's lowered? E.g. so it might be easy to:
 * find the Follower ForLoop
 * identify what is iterated over (ie what went inside zip() )
 * identify loop variables

It seems that ForLoop is mostly just a BlockStmt that offers additional functionality if you know what you're doing with it. I wonder if ForallStmt could be kept around in a similar way. (Or perhaps what I'd need is really just a more full-featured ForLoop).

compiler/AST/foralls.cpp
```diff
@@ -230,3 +234,763 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+    // Todo: why?
+    pfs->parIdxCopyVar()->removeFlag(FLAG_INSERT_AUTO_DESTROY);
+    // NB former buildForallLoopStmt() set this flag on saIdx, not on leadIdx.
+    // We better set it here - it's needed in setConstFlagsAndCheckUponMove().
+    pfs->parIdxVar()->addFlag(FLAG_INDEX_OF_INTEREST);
+    
+  } else    // vass: just curious; remove this "else" clause after testing
+    // NB former buildForallLoopStmt() set this flag on saIdx, not on leadIdx.
+    INT_ASSERT(!pfs->parIdxVar()->hasFlag(FLAG_INDEX_OF_INTEREST));
+
```
vasslitvinov: I will remove the above assertion in the final version - it does not feel useful.

benharsh:
Could you indicate somewhere the testing you have done? I'd also recommend testing these configurations:
- baseline
- verify
- either linux32 or cygwin* (maybe just release/examples)
- valgrind or memleaks

I recognize that's a non-trivial amount of testing, but things have been noisy lately and we don't have many nights to spare before release freezes.

benharsh:
I agree with Michael that the variable names related to parallel loops could be clearer.

It would be useful to have some kind of abstraction to aid us in finding some of the important parts of lowered loops. I'm not sure how difficult that would be to implement though...

mppf:
@vasslitvinov - re the `proc these` for _array and _domain - if I recall correctly you pointed out to me why this was necessary to handle in the compiler? Why couldn't we just have for-loops for serial/standalone/leader/follower that yield appropriately?

compiler/AST/build.cpp
```diff
@@ -1360,7 +1360,7 @@ buildStandaloneForallLoopStmt(Expr* indices,
  *   }
  */
 BlockStmt*
-buildForallLoopStmt(Expr*      indices,
+buildForallLoopStmtOld(Expr*   indices,
```
mppf: I think you said this is still used for forall expressions? Could you add a comment indicating what causes it to be called and what related code would need to change in order to remove it? Could you add those topics as future work in the PR message?

compiler/include/baseAST.h
```diff
@@ -543,6 +547,17 @@ static inline const CallExpr* toConstCallExpr(const BaseAST* a)
...
     break;                                                              \
+  case E_ForallStmt: {                                                  \
+    AST_CALL_LIST (_a, ForallStmt,    iterExprs(),  call, __VA_ARGS__); \
+    ForallIntents* fi = ((ForallStmt*)_a)->withClause();                \
+    AST_CALL_STDVEC(fi->fiVars,  Expr,              call, __VA_ARGS__); \
+    AST_CALL_STDVEC(fi->riSpecs, Expr,              call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, iterRec,      call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, leadIdx,      call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, leadIdxCopy,  call, __VA_ARGS__); \
+    AST_CALL_CHILD(_a, ForallStmt,    masterBlock(),call, __VA_ARGS__); \
```
mppf: I think it is deceptive to use methods like iterExprs() or masterBlock() here. Basically, this section of code needs to track any modifications to the internal representation of ForallStmt. So, I'd personally make those fields public and just have private-looking names (e.g. _masterBlock). But that is up to you.

Arguably we all should just know that this code is sensitive to representational changes...

compiler/passes/normalize.cpp
```diff
@@ -978,6 +944,43 @@ static TypeSymbol* resolveTypeAlias(SymExpr* se)
   return NULL;
 }
 
+//
+// These RiSpec helpers implement reduce intents of the form
```
mppf: You know, I have no idea what RiSpec means. Maybe describe that in the comment and/or choose a better function name?

compiler/passes/normalize.cpp
```diff
@@ -978,6 +944,43 @@ static TypeSymbol* resolveTypeAlias(SymExpr* se)
...
+    SymExpr* riSpecMaybe = dest->firstSymExpr();
+    if (ForallStmt* fs = toForallStmt(riSpecMaybe->parentExpr)) {
+      for_vector(Expr, riSpec, fs->withClause()->riSpecs)
+        if (riSpecMaybe == riSpec)
+          return riSpecMaybe;
+    }
+  }
+  return NULL;
+}
+static void restoreRiSpecCall(SymExpr* riSpec, CallExpr* call) {
```
mppf: I think it'd be nice to have a comment here pointing up and saying that this function partially un-does the work of call_constructor_for_class so that reduce intents don't get messed up.

compiler/AST/foralls.cpp
```diff
@@ -230,3 +234,763 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+  followBlock->insertAtTail("{TYPE 'move'(%S, iteratorIndex(%S)) }", followIdx, followIter);
+
+  followBlock->insertAtTail(followBody);
+  followBlock->insertAtTail(new CallExpr("_freeIterator", followIter));
+
+  return followBlock;
+}
+
+static SymExpr* normalizeFSIterExpr(ForallStmt* pfs, SymExpr* origSE)
```
mppf: I'd prefer not to call this "normalize" since that word means something else in our compiler.
Also, it appears that this is only called from buildForallParIterCall and that is called from resolveParallelIteratorAndForallIntents. These last two can't possibly have reasonable method names, because we have to redirect to _value for for loops in addition to forall loops over _domain/_array. (Or else I'm missing something).

compiler/AST/foralls.cpp
```diff
@@ -230,3 +234,763 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+  return iterSE;
+}
+
+// Return NULL if the def is not found or is uncertain.
+static Expr* findDefOfTemp(SymExpr* origSE)
+{
+  Symbol* origSym = origSE->symbol();
+  if (!origSym->hasFlag(FLAG_TEMP)) return NULL;  // only temps
+
+  SymExpr* otherSE = NULL;
```
mppf: Why doesn't this just use getSingleDef() ?

compiler/AST/foralls.cpp
```diff
@@ -230,3 +234,763 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+    resolveBlockStmt(PARBlock);
+    move_freeIterator_call(PARBlock);
+    PARBlock->flattenAndRemove();
+
+    fs->loopBody()->flattenAndRemove(); // into fs->masterBlock()
+
+    PARBody->insertAtTail(fs->masterBlock()); // m.Block is already resolved
+    fs->masterBlock()->flattenAndRemove();    // into PARBody
+  }
+}
```
mppf: I'd like to see lowerForallStmts and its supporting functions moved to its own file in resolution/.

mppf:
Does lowerForallStmts have to happen during function resolution proper? Could it happen for example after cullOverReferences?

vasslitvinov:
@mppf I am sure lowerForallStmts must currently happen around the spot where I put it. If we move it down a few lines, I strongly suspect we will get failures. (I think I tried and there were reasons not to.)

In the future it does not need to be that way, we can work on pushing lowering until later in compilation.

modules/internal/ArrayViewSlice.chpl
```diff
@@ -90,7 +90,8 @@ module ArrayViewSlice {
     }
 
     iter these(param tag: iterKind) ref
-      where tag == iterKind.standalone && !localeModelHasSublocales {
+      where tag == iterKind.standalone && !localeModelHasSublocales &&
+           __primitive("method call resolves", privDom, "these", tag) {
```
benharsh: I could live with this for now, but how do we want users to write these kinds of iterators in the future?

vasslitvinov: This is an open question. Let us discuss it in #5255 .

mppf:
I've asked @noakesmichael to give us a 2nd opinion on the header file changes (especially the AST representation). Then I'm going to be the main reviewer for the remainder.

mppf:
@vasslitvinov - can you explain why each Forall loop needs two index variables? Why would it need "index" and "index copy"?

compiler/include/baseAST.h
mppf: Why can't this be something like
```
AST_CALL_CHILD(_a, BlockStmt, fMBlock, call, __VA_ARGS__);
AST_CALL_LIST(_a, ?, fIter, call, __VA_ARGS__);
AST_CALL_CHILD(_a, ForallIntents, fWith, call, __VA_ARGS__);
AST_CALL_CHILD(fWith);
```

?

I'm worried there will be bugs if everything in the pointed-to AST node is not visited here...

compiler/resolution/cullOverReferences.cpp
```diff
@@ -1309,6 +1336,13 @@ void cullOverReferences() {
                   FnSymbol* iteratorFn  = detailsVector[i].iterator;
                   SymExpr* iterableSe = toSymExpr(iterable);
 
+                  // Also check if we are iterating using these() method
+                  // ex. functions/ferguson/ref-pair/const-error-iterated*
+                  if (!iterableSe)
+                    if (CallExpr* iterableCall = toCallExpr(iterable))
+                      if (iterableCall->isNamed("these"))
+                        iterableSe = toSymExpr(iterableCall->get(1));
+
```
mppf: Is this a workaround or something we'll need long-term?

mppf:
Can you add completely removing the try token as future work?

compiler/include/expr.h
```diff
@@ -40,6 +40,8 @@ class Expr : public BaseAST {
   virtual bool    isStmt()                                           const;
   virtual QualifiedType qualType();
   virtual void    verify();
+          void    verify(AstTag expectedTag);
+          void    verifyParent(const Expr* child, const char* field);
```
mppf: Could you put this verify/verifyParent change (also in expr.cpp) in a separate PR please?

compiler/include/stmt.h
```diff
@@ -113,6 +113,7 @@ class BlockStmt : public Stmt {
   virtual bool        deadBlockCleanup();
 
   void                appendChapelStmt(BlockStmt* stmt);
+  void                flattenAndRemove();
```
mppf: Could you put the addition of flattenAndRemove in a separate PR please?

modules/standard/FileSystem.chpl
```diff
@@ -777,7 +777,7 @@ iter glob(pattern: string = "*", param tag: iterKind): string
   use chpl_glob_c_interface;
   var glb : glob_t;
 
-  const err = chpl_glob(pattern:c_string, 0, glb);
+  const err = chpl_glob(pattern.localize().c_str(), 0, glb);
```
mppf: This seems unrelated. Could this go in a separate PR please?

mppf:
@vasslitvinov - I think it would be really much nicer if the new AST elements had their own file (like the new TryStmt.{.h,.cpp}). Also this PR is really quite big and it's hard to follow all the changes.

Since I think the AST elements themselves should get a little more scrutiny than all of the lowering code, could you put the addition of the new AST elements in one or more new PRs?

 * add ForallStmt.{.h, .cpp}
 * (possibly in a separate PR) extend AstVisitor and subclasses
 * (possibly in a separate PR) additions to view.cpp

Note that ForallStmt is never actually generated by anything at the time these PR merge. It's simply a way to focus on those parts.

Once those & the other changes I asked to be separated are on master, we can resume the review process for the lowering implementation (which is where most of the code is).

It's harder for me to know how to break apart the changes to build, foralls.cpp, implementForallIntents.cpp, but speak up if you have any ideas there. Otherwise I'll have to screen share with you for a day to understand the changes...
Thanks!

noakesmichael:
I fully support the direction Michael is proposing here.


I share his view that this PR is too large to process thoughtfully.
That’s an easy trap to fall in to.

It’s not clear how many PRs there should be but, when in doubt,
we should be inclined to lean towards more rather than fewer.


I like Michael’s proposal to start with a PR that defines
include/ForallStmt.h and AST/ForallStmt.cpp based on the
declarations and definitions that are in include/stmt.h and
AST/stmt.cpp for this PR.

Then follow up with a few more simple PRs for the relatively trivial
support changes.

These PRs ought to go pretty smoothly and help to clear the deck
so that we can focus on the meat of this PR.

Ideally we’ll find ways to partition the remaining code in to a sequence
of manageable PRs terminating with the PR that actually updates the
Parser to emit the new form of ForallStmt and deals with any final
issues.


Mike





On Jun 13, 2017, at 12:45 PM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@vasslitvinov<https://github.com/vasslitvinov> - I think it would be really much nicer if the new AST elements had their own file (like the new TryStmt.{.h,.cpp}). Also this PR is really quite big and it's hard to follow all the changes.

Since I think the AST elements themselves should get a little more scrutiny than all of the lowering code, could you put the addition of the new AST elements in one or more new PRs?

  *   add ForallStmt.{.h, .cpp}
  *   (possibly in a separate PR) extend AstVisitor and subclasses
  *   (possibly in a separate PR) additions to view.cpp

Note that ForallStmt is never actually generated by anything at the time these PR merge. It's simply a way to focus on those parts.

Once those & the other changes I asked to be separated are on master, we can resume the review process for the lowering implementation (which is where most of the code is).

It's harder for me to know how to break apart the changes to build, foralls.cpp, implementForallIntents.cpp, but speak up if you have any ideas there. Otherwise I'll have to screen share with you for a day to understand the changes...
Thanks!

compiler/AST/foralls.cpp
```diff
@@ -230,3 +235,495 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+    INT_ASSERT(origIterCall);
+
+    FnSymbol* origTarget = origIterCall->resolvedFunction();
+    INT_ASSERT(origTarget);
+
+    const char* targetName = origTarget->name;
+    if (!strncmp(targetName, "_parloopexpr", 12))
+      // a forall loop over a (possibly zippered) forall expression, ex.:
+      //  test/reductions/deitz/test_maxloc_reduce_wmikanik_bug2.chpl
+      targetName = astr("_iterator_for_loopexpr", targetName + 12);
```
mppf: Am I understanding that this is an artifact of not using ForallStmt everywhere we could?

compiler/AST/foralls.cpp
```diff
@@ -230,3 +235,495 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+
+  if (!gotParallel) {
+    // Cannot USR_FATAL_CONT in general: e.g. if these() is not found,
+    // we do not know the type of the index variable.
+    // Without which we cannot typecheck the loop body.
+    if (iterCall->isNamed("these") && isTypeExpr(iterCall->get(2)))
+      USR_FATAL(iterCall, "unable to iterate over type '%s'", toString(iterCall->get(2)->getValType()));
+    else
+      USR_FATAL(iterCall, "A%s leader iterator is not found for the iterable expression in this forall loop", pfs->zippered() ? "" : " standalone or");
+  }
```
mppf: What happens if a leader iterator is defined but a follower isn't ? Where does the error for that come abaout? In some ways I'd expect to see this code resolving both the leader *and* the follower for the leader/follower case.

compiler/AST/foralls.cpp
```diff
@@ -230,3 +235,495 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+
+  followIdx->addFlag(FLAG_INDEX_OF_INTEREST);
+
+  resultBlock->insertAtTail(new DefExpr(iterRec));
+  resultBlock->insertAtTail(new CallExpr(PRIM_MOVE, iterRec, iterExpr));
+  Expr* toNormalize = resultBlock->body.tail;
+
+  // loop body to refer to followIdx instead of leadIdxCopy
+  for_SymbolSymExprs(se, leadIdxCopy)
+    if (se->parentExpr->parentExpr != pfs->loopBody()) // heuristic
```
mppf: I've learned to dread comments like that. If it's a *heuristic*, what you're actually saying is that this code is a bug. Why did you put that comment? Why isn't it a complete solution? Should you be checking for arbitrary levels of nesting with parentExpr here, e.g.?

compiler/AST/foralls.cpp
```diff
@@ -230,3 +235,495 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
...
+    if (zippered == false) {
+      leadForLoop->insertAtTail("'move'(%S, chpl__staticFastFollowCheck(%S))",    T1, iterRec);
+      leadForLoop->insertAtTail(new CondStmt(new SymExpr(T1),
+                                          new_Expr("'move'(%S, chpl__dynamicFastFollowCheck(%S))",    T2, iterRec),
+                                          new_Expr("'move'(%S, %S)", T2, gFalse)));
+    } else {
+      leadForLoop->insertAtTail("'move'(%S, chpl__staticFastFollowCheckZip(%S))", T1, iterRec);
+      leadForLoop->insertAtTail(new CondStmt(new SymExpr(T1),
+                                          new_Expr("'move'(%S, chpl__dynamicFastFollowCheckZip(%S))", T2, iterRec),
+                                          new_Expr("'move'(%S, %S)", T2, gFalse)));
```
mppf: I'd probably write this code as setting two `const char*` variables based upon `zippered` to name the function to call, since the rest is identical. I don't particularly care for the `new_Expr` thing but I don't object to you using it. In this case though, it might be interfering with writing it the way I would have.

compiler/main/runpasses.cpp
```diff
@@ -165,6 +166,11 @@ void runPasses(PhaseTracker& tracker, bool isChpldoc) {
   size_t passListSize = sizeof(sPassList) / sizeof(sPassList[0]);
 
   setupLogfiles();
+  //vass - here because I am patching this file already
+  AstDumpToNode::compact = true;
+  AstDumpToNode::delimitEnter = "(";
+  AstDumpToNode::delimitExit  = ")";
```
mppf: Huh? Maybe you should just delete this change? (Or, at least, I don't think it belongs in this PR)

compiler/passes/normalize.cpp
```diff
@@ -386,65 +388,27 @@ static void transformLogicalShortCircuit()
...
 //
-// Parser represents "x reduce= y" in source code as PRIM_REDUCE_ASSIGN(x,y).
-// lowerReduceAssign() converts it into globalOp.accumulateOntoState(x,y)
-// where globalOp comes from the reduce intent for x.
-// This includes checking that there is, indeed, a reduce intent for x:
-// without it, there is no globalOp.
+// handleReduceAssign(): check+process the reduce= calls
 //
-static void lowerReduceAssign() {
+static void handleReduceAssign() {
```
mppf: I thought that the new ForallStmt is only used in some cases, and in particular not for forall expressions. Did you change reduce intents for both cases? Or can you not use reduce= for forall expressions?

I see below the error saying it must be in a "forall loop". Does that mean not a forall expression? Because that is also a kind of loop, isn't it?

Should the error say "must be in a forall statement"? And do you expect `reduce=` will ever be supported in forall expressions?

compiler/passes/normalize.cpp
```diff
@@ -567,6 +531,15 @@ static void checkUseBeforeDefs() {
...
       collect_asts_postorder(fn, asts);
+#if 0 //vass
+if (!strcmp(fn->name,"main")) { //or ->id == breakOnResolveID
+      void debugSum(int id);void debugSum(BaseAST* ast);
+      printf("\n" "main: "); debugSum(fn);
+      for_vector(BaseAST, ast, asts)
+        debugSum(ast);
+      printf("done\n\n");
+}
+#endif
```
mppf: remove debugging code

compiler/passes/normalize.cpp
```diff
@@ -1066,6 +1041,51 @@ static TypeSymbol* resolveTypeAlias(SymExpr* se)
...
+      for_vector(Expr, riSpec, fs->withClause()->riSpecs)
+        if (riSpecMaybe == riSpec)
+          return riSpecMaybe;
+    }
+  }
+  return NULL;
+}
+//
+// This function partially un-does the work of call_constructor_for_class
+// so that reduce intents specs (see above) don't get messed up.
```
mppf: Would it be simpler to make call_constructor_for_class not transform reduce intents, instead of transforming them and then undoing it? Is there some reason that would be harder to implement?

compiler/passes/scopeResolve.cpp
```diff
@@ -427,10 +422,22 @@ static void scopeResolve(const AList& alist, ResolveScope* scope) {
...
       }
 
+    } else if (ForallStmt* forallStmt = toForallStmt(stmt)) {
+      BlockStmt* fBody = forallStmt->loopBody();
+      // or, we could construct ResolveScope specifically for forallStmt
+      ResolveScope* bodyScope = new ResolveScope(fBody, scope);
+      // cf. scopeResolve(FnSymbol*,parent)
+      for_alist(ivdef, forallStmt->inductionVariables()) {
+        Symbol* sym = toDefExpr(ivdef)->sym;
+        if (strcmp(sym->name, "chpl__tuple_blank"))
```
mppf: Can you explain what's going on with this strcmp? Seems like some sort of weird special case/workaround.
Better would be to find another way, but at a minimum we need a comment talking about the special case it's trying to handle.

compiler/passes/scopeResolve.cpp
```diff
@@ -427,10 +422,22 @@ static void scopeResolve(const AList& alist, ResolveScope* scope) {
...
+          bodyScope->extend(sym);
+      }
+      scopeResolve(fBody->body, bodyScope);
+      // Anything else needs to be scopeResolve-d here?
```
mppf: Probably this comment just needs removal

compiler/resolution/functionResolution.cpp
```diff
@@ -5637,6 +5640,10 @@ resolveExpr(Expr* expr) {
       }
     }
 
+    if (!tryFailure)
+      callStack.pop();
```
mppf: I'm having trouble understanding why callStack.pop() needs to move. Is this change necessary for ForallStmts? Is this just trying to make this code clearer? If so it should not be in this PR.

compiler/resolution/functionResolution.cpp
```diff
@@ -5667,9 +5674,8 @@ resolveExpr(Expr* expr) {
       }
 
     }
-
-    callStack.pop();
   }
+  INT_ASSERT(expr); // ensure no surprises
```
mppf: I don't think that comment adds any value.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -694,6 +695,23 @@ static void verifyOuterVars(BlockStmt* body2,
 //
 void implementForallIntents1(DefExpr* defChplIter)
 {
+  // Sometimes defChplIter is introduced for ForallStmt. If so, ignore.
+  bool oldStyle = false;
+  if (DefExpr* next = toDefExpr(defChplIter->next))
+    if (!strcmp(next->sym->name, "chpl__leadIter") ||
+        !strcmp(next->sym->name, "chpl__saIter"))
```
mppf: It's a shame to use strcmps for this. Seems like a flag would be better. However I'm not objecting too strongly if you expect to remove "old style" foralls in short order.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1699,3 +1717,735 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+
+  detupleLeadIdxNew(leadIdx, leadIdxCopy, fbody, numLeaderActuals,
+                    outerVars, shadowVars, reduceGVars);
+}
+
+static void insertGenerateAfter(Expr* ref, Symbol* fiVarSym, Symbol* globalOp) {
+  ref = ref->next; // nicer ordering of the following insertions
+  VarSymbol* genTemp = newTemp("chpl_gentemp");
+  ref->insertBefore(new DefExpr(genTemp));
+  ref->insertBefore("'move'(%S, generate(%S,%S))",
```
mppf: what is "generate" referring to?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1699,3 +1717,735 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+
+  if (numLeaderActuals == 0)
+    fbody->insertBefore("'move'(%S, %S)", leadIdxCopy, leadIdx);
+
+  detupleLeadIdxNew(leadIdx, leadIdxCopy, fbody, numLeaderActuals,
+                    outerVars, shadowVars, reduceGVars);
+}
+
+static void insertGenerateAfter(Expr* ref, Symbol* fiVarSym, Symbol* globalOp) {
+  ref = ref->next; // nicer ordering of the following insertions
```
mppf: Let's make a new variable `next` and INT_ASSERT that it isn't NULL.

compiler/resolution/preFold.cpp
```diff
@@ -883,6 +882,23 @@ Expr* preFold(CallExpr* call) {
         }
       }
     }
+  } else if (call->isPrimitive(PRIM_REDUCE_ASSIGN)) {
```
mppf: It'd be nice to put a comment here like "Translate PRIM_REDUCE_ASSIGN into a call to accumulateOntoState"

modules/dists/SparseBlockDist.chpl
```diff
@@ -229,7 +229,13 @@ class SparseBlockDom: BaseSparseDomImpl {
...
   }
 
-  iter these(param tag: iterKind) where tag == iterKind.standalone {
+  iter these(param tag: iterKind) where tag == iterKind.standalone &&
+    // Ensure it is legal to invoke the standalone iterator
+    // on locDom.mySparseBlock below.
+    __primitive("method call resolves",
+                locDoms[createTuple(rank,int,0)].mySparseBlock._value,
+                "these", tag)
+  {
```
mppf: I understand why this logic might be worthwhile for array views but I'm scratching my head a bit about SparseBlock. Shouldn't all of the ._values in SparseBlock just always have serial/standalone/leader/follower iteratiors? Is there some reason that is not a good idea? Aren't they all controlled by our implementation of SparseBlock?

In other words, what is it about SparseBlock that required this, separately from Block or Cyclic or ... ?

modules/dists/SparseBlockDist.chpl
```diff
@@ -367,7 +373,13 @@ class SparseBlockArr: BaseSparseArr {
...
   }
 
-  iter these(param tag: iterKind) ref where tag == iterKind.standalone {
+  iter these(param tag: iterKind) ref where tag == iterKind.standalone &&
+    // Ensure it is legal to invoke the standalone iterator
+    // on locA.myElems below.
+    __primitive("method call resolves",
+                locArr[locArrDom.low].myElems._value,
+                "these", tag)
+   {
```
mppf: Same comment here as above.

mppf:
@vasslitvinov  - I just noticed that this PR added a "Forward the leader iterator as well" to NPBRandomStream but not to PCGRandomStream. Why did we need to add the leader forwarder? Assuming there was a test failure, can we add a test for the equivalent functionality for PCGRandomStream?

mppf:
https://github.com/chapel-lang/chapel/blame/master/modules/standard/Random.chpl#L2194

vasslitvinov:
@mppf I am sure your diagnosis is correct. I will look into adding a new test.

PR #5296, opened 2017-02-02
First steps towards better irregular domain assignment

mppf:
Before this change, domain assignment was implemented in ChapelArray and included basically two branches: one branch for regular domains that amounted to calling `setIndices(rhs.getIndices())`and a second branch for irregular domains that worked in serial and used `dsiIndsIterSafeForRemoving`.

That implementation approach presents problems for things like distributed associative or distributed sparse, where it's important that those implementations have an opportunity to optimize. In particular, assigning one distributed associative domain to another, where they share the same distribution, is an important case to optimize for. In that case, the domain assignment should be possible to implement with no communication (assuming privitaization and no arrays over the lhs-domain). Since such implementations depend on the details of the implementation, it makes sense to have each domain map define how domain assignment is to be accomplished.

This strategy is complicated by the fact that domain assignment is sometimes possible between domains of different types. A typical example might be assignment between a distributed domain and a non-distributed one. Of course, one could argue that such things should be compilation errors... but the pattern itself is useful and currently supported directly in domain assignment.

One piece of important background is worth mentioning: domain assignment causes arrays over the updated domain to be resized.

The strategy to allow irregular domains to implement domain assignment is as follows:
 * create a new `dsiAssignDomains` as required part of the DSI interface
 * implement the old behavior with generic functions `assignDomainWithGetSetIndices` (for rectangular domains) and `assignDomainWithIndsIterSafeForRemoving` (for irregular domains)
 * clean up the type-system issues with `assignDomainWithGetSetIndices`.
   * Since `assignDomainWithGetSetIndices` iterates over the domain's arrays and calls `dsiReallocate` on them, it would run into issues (described below) with overriding and overloading at once if dsiReallocate continued to accept a generic domain and be declared in BaseArr
   * To resolve the problem, factor `rank`, `idxType`, and `stridable` out of individual rectangular arrays into BaseArrOverRectangularDom; and while there, factor out eltType into BaseRectangularArr.
   * Now, `assignDomainWithGetSetIndices`  can cast the BaseArr from the domain's list of arrays into the appropriate `BaseArrOverRectangularDom`
   * `BaseArrOverRectangularDom` contains a default implementation of `dsiReallocate` that is focused on only the appropriate rank and index type.
   * All `dsiReallocate` functions now accept only a tuple of ranges, since it is only used (and only makes sense) in the rectangular array case.

### dsiAssignDomains signature

The new method is something like this:
``` chapel
  proc SomeDom.dsiAssignDomain(rhs: domain, lhsPrivate:bool) { }
```
The first argument is generic over all domains.  The second argument is used to optimize initCopy-like cases when assigning domains. In particular, in the case of initCopy, there is no need to worry about arrays over the new domain or old elements in the new domain (since we are constructing the new domain).

### The override-overload issue

Now, note that the DSI interface is partially specified in the methods in the base class BaseDom. It would be nice to put a dsiAssignDomain there, but doing so encounters a problem with how generic functions interact with virtual methods.

When a generic function is overridden (such as with one taking a `domain` argument like dsiAssignDomain), the compiler resolves versions of that function for each instantiation up and down the inheritance tree.

That means that if we have
``` chapel
  proc BaseDom.dsiAssignDomain(rhs:domain)
```
and somewhere there is
``` chapel
  proc DomainA.dsiAssignDomain(rhs:domain)
  proc DomainB.dsiAssignDomain(rhs:domain)
```
and then there is a call like
``` chapel
  someDomainA.dsiAssignDomain( some2DDomain )
```
then function resolution will instantiate BaseDom.dsiAssignDomain for a that particular 2D domain argument. Then, it will instantiate DomainB.dsiAssignDomain for that particular 2D domain argument. But DomainB isn't necessarily one that can be set to a 2D domain - it might be 1D.

I think what the compiler is doing here is reasonable because if BaseDom includes dsiAssignDomain, one could presumably call it with a 2D domain type and expect virtual dispatch to handle it.

To do better at describing and enforcing the DSI "interface" at compile time, I think we need interfaces/concepts. For now, I'll seek to document this required method in the DSI technote.

This PR doesn't actually change the way domain assignment works for irregular domains yet.

### Testing

Passed quickstart testing.

- [ ] figure out what's going wrong with distributions/robust/arithmetic/resizing/resizeFromOtherLocale
- [ ] qthreads testing
- [ ] GASNet testing  (perhaps with Cyclic/Replicated as well)
- [ ] perf testing/ perf playground
- [ ] check for interference with associative array for Replicated's targetLocales vs array views


modules/internal/ChapelArray.chpl
```diff
@@ -1767,6 +1767,7 @@ module ChapelArray {
                                                          isAssociativeDom(d2)) {
     if d1._value == d2._value then return true;
     if d1.numIndices != d2.numIndices then return false;
+    // TODO -- should this be a forall?
```
bradcray: You can't return within a forall, so there would need to be some sort of reduction, atomic, or eureka here to make that work...

modules/internal/ChapelDistribution.chpl
```diff
@@ -816,4 +890,71 @@ module ChapelDistribution {
     // runs the array destructor
     delete arr;
   }
+
+  // domain assigment helpers
+
```
bradcray: I wondered a bit about putting these routines into a module other than this one, possibly a new one, as this one has seemed poorly named and/or overloaded for too long now.  Ultimately, I think we should be splitting ChapelDistribution and ChapelArray into approximately a module per major class/record.  But it'd be helpful to not dive into that before we land array views so maybe we can park these here until then.

Should their names be chpl_ protected?  Or if they're intended for end-users, should we chpldoc them?

modules/internal/ChapelRange.chpl
```diff
@@ -2092,13 +2092,15 @@ proc _cast(type t, r: range(?)) where isRangeType(t) {
...
   // source: Knuth Volume 2 --- Section 4.5.2
   //
-  proc chpl__extendedEuclidHelper(u, v)
+  proc chpl__extendedEuclid(u, v)
+    where isIntType(u.type) && isIntType(v.type)
   {
-    var zero: u.type = 0;
-    var one: u.type = 1;
+    // figure out the bigger type (in case u is int(64) v is int(32) e.g.)
+    var zero = 0:u.type + 0:v.type;
```
bradcray: Will this get us into trouble (or perhaps "more trouble than we were already in"?) if the types are int and uint?

bradcray: I'm also curious what motivated these changes and whether there's a test that will exhibit the problem if it were reverted...

bradcray:
I reviewed this from a design perspective, and also looked through the code at a very high-level.  The design changes all seem to me to be somewhere between "looks good!" and "whatever" (maybe just because I didn't think about the issue hard enough), so I'm good with giving this PR a more thorough code review and merging it when it's ready.  I think it would be worthwhile to run this in a perf playground if @ronawho has one available just because it doesn't take much to throw off array performance.

If he has time, I think @benharsh would be an ideal reviewer for it given his familiarity both with domain maps and the associative domain/array code.  He would also probably be able to anticipate potential problematic aspects of merging with the array views branch and perhaps be able to make the call as to whether one or the other should go in first.

benharsh:
FYI, using an associative array for Replicated's targetLocales has been implemented/tested during the array-views work.

doc/technotes/dsi.rst
```diff
@@ -236,6 +236,20 @@ class ``GlobalDomain``
   ``dsiNewRectangularDom()`` to the index set of the corresponding Chapel
   domain value.
 
+.. method:: proc GlobalDomain.dsiAssignDomain(rhs: domain, lhsPrivate:bool) {
```
benharsh: Should ``rhs`` have the ``const`` intent?

modules/dists/PrivateDist.chpl
```diff
@@ -116,6 +113,10 @@ class PrivateDom: BaseRectangularDom {
   proc dsiSetIndices(x: domain) { halt("cannot reassign private domain"); }
   proc dsiGetIndices() { return {0..numLocales-1}; }
 
+  proc dsiAssignDomain(rhs: domain, lhsPrivate:bool) {
+    halt("cannot reassign private domain");
```
benharsh: Is it possible for this to be a compile-time error?

bradcray:
> FYI, using an associative array for Replicated's targetLocales has been implemented/tested during the array-views work.

Looking at the array views branch this morning, I was just wondering whether this piece of the array views work should be split off and committed separately to master (if that wasn't too much effort).

modules/internal/ChapelDistribution.chpl
benharsh: I think these should at least be prefixed with ``chpl_``.

modules/internal/ChapelRange.chpl
mppf: The test I was having trouble with is test/types/range/diten/rangeSlice.chpl. It appears that reverting these changes allows that test to pass, so assuming no testing surprises, I'll remove the changes to extendedEuclid.

doc/technotes/dsi.rst
mppf: default intent for domains is already `const ref`.

modules/dists/PrivateDist.chpl
mppf: Not without some significant effort. autoCopy/initCopy for domains are resolved during resolution so they can be called later in callDestructors. If there were a compilation error here, any program using PrivateDist would fail with a resolution error.

modules/internal/ChapelArray.chpl
```diff
@@ -3065,9 +3065,8 @@ module ChapelArray {
 
     a._instance.dsiAssignDomain(b, lhsPrivate=false);
 
-    if _isPrivatized(a._instance) {
-      _reprivatize(a._value);
-    }
+    // MPF note: should this call _reprivatize(a._value) at this point?
+    // I think so, but I see errors if I enable it.
```
mppf: @benharsh - this seems strange to me. It probably makes sense to leave it off for now, but I wanted to check if it's obvious to you what the problem is.

mppf: Hmm... it gets resizeFromOtherLocale to work to leave it out, but other tests, including a new one I wrote to check arrays-resize-on-domain-assignment do not.

modules/dists/PrivateDist.chpl
benharsh: OK, fair enough.

modules/internal/ChapelArray.chpl
mppf: Other tests that fail with this change:
distributions/robust/arithmetic/basics/test_array_iteration1.chpl
distributions/robust/arithmetic/basics/test_array_resize_from_domain_assignment.chpl
distributions/robust/arithmetic/basics/test_array_swap.chpl
distributions/robust/arithmetic/basics/test_domain_iteration1.chpl
distributions/robust/arithmetic/basics/test_zipper_default.chpl
distributions/robust/arithmetic/composite/test_composite1.chpl
distributions/robust/arithmetic/composite/test_composite2.chpl
distributions/robust/arithmetic/kernels/hpl.chpl
distributions/robust/arithmetic/modules/test_module_Norm.chpl
distributions/robust/arithmetic/modules/test_module_Random.chpl
distributions/robust/arithmetic/modules/test_module_Search.chpl
distributions/robust/arithmetic/reindexing/test_array_alias4.chpl
distributions/robust/arithmetic/reindexing/test_reindexing4.chpl
distributions/robust/arithmetic/reindexing/test_strided_reindex1.chpl
distributions/robust/arithmetic/slicing/test_strided_slice1.chpl
distributions/robust/arithmetic/slicing/test_strided_slice2.chpl

mppf:
Closing this for now since we'll need to revisit it after the release.

PR #5298, opened 2017-02-03
Edits to CHIP 10 on deinit()

vasslitvinov:
Clarified that a compiler-generated deinit() is equivalent
to an empty user-defined deinit().

Added the rationale for current rules.

Minor style changes.

More changes needed
-------------------

While reading the CHIP, I realized that

* We need to define deinitialization at least for arrays.

* Deinitialization makes sense also for domains,
given that the domain map author may want to perform some actions
(like freeing resources) when the domain is deleted.

* To dot all the Is, we should at some point look through
all Chapel types and see which ones need to have deinitialization
defined or be not a NOP.

  As an intermediate step we could say things like "defining the actions
that happen when a domain is deinitialized is future work."

* The current CHIP imposes deinitialization order on record fields
and hints on that requirement for array/domain fields. Meanwhile
Brad in his email on 1/26 suggested, and we all agreed to,
reverse-declaration order of deinitialization for **all** fields.
Should we change the CHIP accordingly?

* [ ] change single backticks to double backticks

### DIFF
--- a/doc/developer/chips/10.rst
+++ b/doc/developer/chips/10.rst
@@ -533,7 +533,7 @@ Deinitialization
 
 Deinitialization is performed in these cases:
 
- * For a ``var`` or ``const`` variable - when it does out of scope.
+ * For a ``var`` or ``const`` variable - when it goes out of scope.
  * For a ``[const] in`` formal - before its parent function returns;
    after the return value has been computed.
    There may be exceptions when a record or an array is returned.
@@ -551,17 +551,16 @@ Deinitialization is performed in these cases:
 Deinitialization consists of the following actions,
 depending on the type of the variable being deinitialized:
 
- * Primitive, enumerated, range, class, synchronization types -
+ * Primitive, enumerated, range, class, synchronization type -
    nothing happens. Implementation-level actions,
    ex. deallocation of a string or a sync variable,
    can be performed without affecting the observable behavior directly.
+ * Tuple - its components are deinitialized.
  * A record variable; a class instance (upon ``delete``) - see below.
- * Unions - t.b.d.
- * Arrays, domains, distributions - defined by their domain map
-   in a way t.b.d.
- * Arrays, tuples - their elements/components are deinitialized.
- * For an array, its elements are deinitialized before
-   its domain map-defined actions are performed.
+ * Union - t.b.d.
+ * Array - its elements are deinitialized. Then, additional actions
+   are performed as defined by its domain map, if any.
+ * Domain, distribution - defined by their domain map in a way t.b.d.
  * Alias types - at present, the action is determined by the aliasee type.
 
 Note that we distinguish between:
@@ -573,9 +572,9 @@ Note that we distinguish between:
 The remainder of this section is concerned with deinitialization of
 record variables, or "records", and class instances, or "classes".
  
-To specify actions that occur upon deinitialization, specify a `proc
-deinit()` method. This method must have zero arguments and cannot be called
-directly. If a type does not specify a `proc deinit()` method, the compiler
+To specify actions that occur upon deinitialization, specify a ``proc
+deinit()`` method. This method must have zero arguments and cannot be called
+directly. If a type does not specify a ``proc deinit()`` method, the compiler
 will generate one.
 It will be equivalent to a user-written ``deinit`` with an empty body.
 
@@ -585,10 +584,10 @@ compiler-generated or user-defined, in the process of deinitialization:
  * parent ``deinit`` methods are called in the case of inherited classes, and
  * all fields of record type are deinitialized.
 
-The expected use-case for the `deinit` method is to call `delete` on
+The expected use-case for the ``deinit`` method is to call ``delete`` on
 fields of class type, when appropriate. It is the responsibility of the
-class designer to `delete` fields when appropriate. The compiler will not
-automatically add `delete` calls for fields of class type.
+class designer to ``delete`` fields when appropriate. The compiler will not
+automatically add ``delete`` calls for fields of class type.
 Conversely, the compiler adds ``deinit`` calls on fields of record type.
 These calls cannot be added by the user.
 
@@ -618,13 +617,16 @@ in reverse declaration order.
 Rationale
 *********
 
-The goal of the above specifications is support type safety, specifically
+Our goal is to support type safety, specifically
 to prevent accesses to fields after they have been deinitialized.
 This includes preventing access to an array after its domain
 has been deinitialized.
-
-The above specifications one way to achieve this goal.
+The above rules specify one way to achieve this goal.
 They are a work in progress.
+
+For example, disallowing the use of ``this`` as a whole within ``deinit``
+is intended merely as the minimum functionality currently required,
+so that we can refine it in the future without breaking code.
 
 
 Summary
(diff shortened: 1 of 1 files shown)


doc/developer/chips/10.rst
```diff
@@ -531,24 +531,76 @@ For more details on when the copy initializer would be called, please refer to
...
 
 The expected use-case for the `deinit` method is to call `delete` on
 fields of class type, when appropriate. It is the responsibility of the
 class designer to `delete` fields when appropriate. The compiler will not
 automatically add `delete` calls for fields of class type.
+Conversely, the compiler adds `deinit` calls on fields of record type.
+These calls cannot be added by the user.
+
+Ensuring type safety
+********************
```
mppf: I think this section is a good discussion of our rationale, but I never intended to include such a detailed rationale in this CHIP. If we decide to include the Rationale here, we should put it after the description of the language rules. I think it's confusing here because a user might interpret "could potentially be accessed" as saying that it is legal Chapel to write code matching one of these sections.

In particular, it was our intention to prevent a user from using the whole object `this` because dynamic dispatch could occur. We relaxed that slightly to allow field access (because these are technically methods now) but that was more an accommodation to the current compiler implementation than anything else. I'm not sure we agreed that you could pass `this` to a function that never does dynamic dispatch on it.

Lastly, in the case of section (d), we can make "dependencies" between fields in any order we please with mutable fields. I havn't personally been very satisfied with the term "dependency" in this context and I'm not sure what it means. Consider this:

``` chapel
class MyObject { }
record Owned {
 var instance;
 proc deinit() { delete instance; }
}
class Parent {
  var myInstance: MyObject;
  proc init() {
    myInstance = new MyObject(); 
  }
  proc deinit() {
    writeln(myInstance);
  }
}
class Child {
  var pointsToParent: Owned(MyObject);
  proc init() {
    pointsToParent = new Owned(nil);
    super.init();
    pointsToParent.instance = this.myInstance;
    // now pointsToParent refers to the object in Parent.myInstance
  }
}
var x = new Child();
delete x;
```

Would you call pointsToParent a "dependency" ?
Uh-oh! The deinitialization process on `x` will cause Parent.myInstance to be deleted before it is used in Parent.deinit!

Or consider this example, which has "dependent" fields in reverse declaration order:
``` chapel
record ObjectPair {
  var a: MyObject;
  var b: MyObject;
  proc deinit() {
    writeln(a);
    writeln(b);
  }
}
class Test {
  var recPair: ObjectPair;
  var objA:Owned(MyObject);
  var objB:Owned(MyObject);
  proc init() {
    super.init();
    objA = new Owned(new MyObject());
    objB = new Owned(new MyObject());
    recPair = new ObjectPair( objA, objB );
  }
}
var y = new Test();
delete y;
```

Would you call recPair's use of objA and objB a "dependency"?
Uh-oh! The deinitialization process on `y` will deinitialize objA and objB  before they are used in `ObjectPair.deinit`.

Bottom line: I don't think we can ever really "ensure type safety" as long as `delete` is subject to user management (and not checked to a great degree by the compiler). I think the rules we had in the CHIP already are fine (along with your minor changes to them) but I don't agree with this new section.

lydia-duncan: Haven't gone through the PR yet, but I wanted to respond to Michael's comment first.

> I'm not sure we agreed that you could pass this to a function that never does dynamic dispatch on it.

I'm fairly certain we agreed that you could never pass this to a function during deinit.  I'm not sure at what point this became more specific.  The rationale still holds on this - if we were to allow operations on the "this" during any deinit, because we execute all deinits from child to base type, the object won't be in a valid state for any deinit method except for the first one that is called (because we decided that updating the type as we completed each deinitializer was probably a bad idea and at least unnecessary for now).

doc/developer/chips/10.rst
```diff
@@ -531,24 +531,76 @@ For more details on when the copy initializer would be called, please refer to
...
-Records and classes can specify actions to take when a variable of that type is
-deinitialized. Record variables are deinitialized when they go out of scope and
-class variables are deinitialized when `delete` is called on them.
+Deinitialization is performed in these cases:
 
-To specify additional actions that occur upon deinitialization, specify a `proc
+ * Record variables are deinitialized when they go out of scope.
+ * Class variables are deinitialized when `delete` is called on them.
+
+To specify actions that occur upon deinitialization, user can specify a `proc
```
lydia-duncan: This sentence is a fragment now that you've changed the second clause to no longer be in the imperative tone.  I believe this would be rectified by changing it to "the user can specify"

doc/developer/chips/10.rst
```diff
@@ -531,24 +531,76 @@ For more details on when the copy initializer would be called, please refer to
...
 deinit()` method. This method must have zero arguments and cannot be called
 directly. If a type does not specify a `proc deinit()` method, the compiler
 will generate one.
+It will be equivalent to a user-written `deinit` with an empty body.
```
lydia-duncan: Is this true?  Or were we expecting the compiler generated deinit to call delete on class fields?

lydia-duncan: I see that this is mentioned in a later (older) section.  I must have misremembered

doc/developer/chips/10.rst
```diff
@@ -531,24 +531,76 @@ For more details on when the copy initializer would be called, please refer to
...
+
+ (b) `deinit()` in children of `RC`, if any, are invoked
+     before `RC.deinit()` and therefore before `F` is deinitialized.
+
+ (c) Dynamic dispatch on `this` is disallowed in `deinit` methods.
+
+ (d) Dependent fields within `RC` are deinitialized
+     in reverse declaration order.
+
+This approach is detailed below. It is a work in progress.
```
lydia-duncan: I agree with Michael that this section is too detailed.  Do we not already mention all of these requirements?  Why not put the short rationale for why each limitation is necessary closer to the introduction of the limitation?

vasslitvinov:
Here is my commentary for 84b16df, copied here from the commit message:

* Define deinitialization as applicable to all types.
  This lets me talk about when deinitialization occurs etc.
  in a simple, uniform manner.

* Distinguish between deinitialization of a class variable vs.
  of a class instance. Same motivation.

* State simply: "deinitialize fields in reverse declaration order".
  Given that we agree on this requirement,
  I do not see why we need to say any more.

* Our meeting notes say explicitly "Disallow accessing 'this' as a
  whole object in the body of deinit()". I believe we wrote this
  sentence together. So this is what the new revision says.

* Do you think that we can allow methods on 'this' when it is a record?
  Or we should still avoid it because in the future we may introduce
  inheritance on records and again get in trouble for accessing
  child fields too late?

* I moved and reduced the Rationale section, by popular request.

* I have an idea of what a "dependence between fields" is.

  It is when the value of one field, say F1, references
  another field, say F2, or a part of F2.

  In the absence of ref fields, this can only happen when
  F1 is/contains an array and F2 is/contains that array's domain.

  It is significant because:

   + This dependence comes from field declarations.
   + It is established during Phase 1 of initialization.
   + Users cannot affect it.

  It would be the same for reference fields, if we had them.
  Can you see other scenarios where these properties hold?

  For these kinds of fields, it is our job to ensure type safety,
  including during deinitialization.

* By contrast, we cannot even get close to ensuring type safety
  for **class** fields, alas. No matter what we do, it will be
  possible - even easy - to write code that will access
  class instances after they are delete()-ed.

  So I argue that we do not need to go out of our way to prevent that.

* How to reflect these thoughts in the spec is t.b.d.

PR #5299, opened 2017-02-03
Enable user-defined deinit() methods

vasslitvinov:
This is an elaboration on #5200 that adjusts the code and tests
in accordance with the current CHIP 10. In particular,

* Disallow deinit() with arguments.
* Disallow user calls to deinit().
* Rename, internally to the compiler ~classname methods to deinit().
* Added tests.
* While there, tidied up some tests in classes/figueroa

Next steps:

* Convert our Chapel code base to the new name.
* Live in the new world for a while to make sure everything works.
* Deprecate ~classname naming.

Tested: linux64 on full suite; gasnet on multilocale tests.

### DIFF
--- a/compiler/passes/checkParsed.cpp
+++ b/compiler/passes/checkParsed.cpp
@@ -24,16 +24,19 @@
 #include "stmt.h"
 #include "expr.h"
 #include "astutil.h"
+#include "stringutil.h"
 #include "stlUtil.h"
 #include "docsDriver.h"
 
 
 static void checkNamedArguments(CallExpr* call);
+static void checkExplicitDeinitCalls(CallExpr* call);
 static void checkPrivateDecls(DefExpr* def);
 static void checkParsedVar(VarSymbol* var);
 static void checkFunction(FnSymbol* fn);
 static void checkExportedNames();
 static void checkModule(ModuleSymbol* mod);
+static void setupForCheckExplicitDeinitCalls();
 
 void
 checkParsed() {
@@ -50,8 +53,11 @@ checkParsed() {
     return;
   }
 
+  setupForCheckExplicitDeinitCalls();
+
   forv_Vec(CallExpr, call, gCallExprs) {
     checkNamedArguments(call);
+    checkExplicitDeinitCalls(call);
   }
 
   forv_Vec(DefExpr, def, gDefExprs) {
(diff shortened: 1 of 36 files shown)


compiler/passes/checkParsed.cpp
```diff
@@ -124,6 +130,25 @@ checkNamedArguments(CallExpr* call) {
...
+static const char* dotAstr;
+static VarSymbol* deinitStrLiteral;
+
+static void setupForCheckExplicitDeinitCalls() {
+  dotAstr = astr(".");
+  deinitStrLiteral = new_CStringSymbol("deinit");
+}
+
+static void checkExplicitDeinitCalls(CallExpr* call) {
+  if (call->id == breakOnResolveID) gdbShouldBreakHere(); //vass
```
mppf: Looks like debug code to remove to me

compiler/passes/checkParsed.cpp
```diff
@@ -124,6 +130,25 @@ checkNamedArguments(CallExpr* call) {
...
+  if (UnresolvedSymExpr* target = toUnresolvedSymExpr(call->baseExpr))
+    if (target->unresolved == dotAstr)
+      if (SymExpr* arg2 = toSymExpr(call->get(2)))
+        if (arg2->symbol() == deinitStrLiteral)
+          // OK to invoke explicitly from internal modules
+          if (strcmp(call->parentSymbol->name, "chpl__delete"))
```
mppf: The comment above and the actual code don't match. It says OK to invoke from internal modules; so I'd expect you to be checking if it's in an internal module. Instead you check if it's in chpl__delete. At least make the comment and the code consistent.

compiler/passes/normalize.cpp
```diff
@@ -173,13 +173,14 @@ void normalize() {
...
 
-        INT_ASSERT(fn->name[0] == '~' && thisDef);
+        INT_ASSERT(thisDef);
 
-        // make sure the name of the destructor matches the name of the class
-        if (ct && strcmp(fn->name + 1, ct->symbol->name) != 0) {
-          USR_FATAL(fn, "destructor name must match class name");
+        // verify the name of the destructor
+        if (ct && (strcmp(fn->name + 1, ct->symbol->name) &&
+                   strcmp(fn->name, "deinit"))) {
```
mppf: I'd prefer this conditional to be of a simpler style. (e.g. bool isDestructor = .... ; bool isDeinit = ...; if (!isDestructor && !isDeinit);).
Otherwise, it's hard for me to follow the logic.
In any case, I think it should check that the first character is `~` in the destructor case.

mppf:
Other than the comments, I wanted to mention I would've preferred that this PR include Brad's commits rather than just re-applying the same changes in a new commit. As your reviewer, I had already seen Brad's changes and could have reviewed it more quickly knowing just which parts you changed.

Once you clean up the minor issues, I think this is ready to merge.

test/classes/deinitializers/deinitAndDestruct.good
```diff
@@ -0,0 +1,3 @@
+deinitAndDestruct.chpl:11: error: ambiguous call 'C.deinit()'
+deinitAndDestruct.chpl:2: note: candidates are: C.~C()
+deinitAndDestruct.chpl:6: note:                 C.deinit()
```
lydia-duncan: I am curious what your thoughts were on whether we should always error if both a destructor and a deinitializer are present on the same type.  I'm asking because I figured you saw the discussion on the other PR, but no one else commented on it so I wasn't sure what position to assume everyone took.

mppf: You're probably asking Vass, but my opinion is that we should take the path of least resistance in this case.

test/classes/diten/multipledestructor.chpl
```diff
@@ -1,7 +1,7 @@
 class C {
-  proc ~C() { writeln("In ~C #1"); }
+  proc deinit() { writeln("In deinit #1"); }
 }
-proc C.~C() { writeln("In ~C #2"); }
+proc C.deinit() { writeln("In deinit #2"); }
```
lydia-duncan: Should we update the name of this test, to reflect that it exercises deinitializers now instead of destructors?

PR #5300, opened 2017-02-03
Update initializer phase1 so that omitted fields implement init rather than assign

noakesmichael:
Brad was playing with a class that included a field of array type.  He tried to provide
a user defined init() method where his code ran in phase 2.  The code compiled but
failed with a "dereference nil" bug on master.

The compiler had inserted code to initialize the "omitted" field for phase 1 but
accidentally generated an assign operation on the field rather than implementing
low-level initialization (there isn't a fully principled way for it to do real initialization
yet).

This trivial PR converts the use of the assignment call with an implementation
for low-level init.

Ran on test/classes/initializers on darwin.  Ran full paratest on linux64.

### DIFF
--- a/compiler/passes/initializerRules.cpp
+++ b/compiler/passes/initializerRules.cpp
@@ -459,39 +459,52 @@ bool isParentField(AggregateType* t, const char *name) {
   }
 }
 
-static
-void insertOmittedField(Expr* next, DefExpr* field, AggregateType* t) {
+static void insertOmittedField(Expr* next, DefExpr* field, AggregateType* t) {
+  SET_LINENO(next);
+
   const char* nextField = field->sym->name;
 
-  SET_LINENO(next);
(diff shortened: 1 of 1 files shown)


bradcray:
We should be sure to file a test that locks in the new behavior like the original test.  I can check in the one I had yesterday if you like, but if you developed anything more sophisticated, let's check that in.

noakesmichael:
I used your test to develop the “5 minute fix”.


On Feb 3, 2017, at 3:57 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


We should be sure to file a test that locks in the new behavior like the original test. I can check in the one I had yesterday if you like, but if you developed anything more sophisticated, let's check that in.

bradcray:
OK, I'll file it then.

Issue #5281, opened 2017-01-30
PCGRandomStream does not default the seed to the current time


nspark:
### Summary of Problem
The `RandomStream` class in the `PCGRandom` module does not appear to set the default seed to the current time, as documented. The output of pseudo-random values is identical from run-to-run, even across different systems. I expected the pseudo-random values to vary from run-to-run, as one would see with `srandom(time(NULL))` and `random()` in C.

### Steps to Reproduce

**Source Code:**

```chapel
use Random.PCGRandom;

var R = new RandomStream(uint);
for i in 1..10 do
  writeln(R.getNext() % 100);
```

**Compile command:**
`chpl test-pcg-random.chpl -o test-pcg-random`

**Execution command:**
`./test-pcg-random`

**Output:**
```
0
0
94
62
28
10
41
53
39
79
```

### Configuration Information

- Output of `chpl --version`: 1.14.0
- Environments:
  - Cray chapel/1.14.0 module
  - Source build with `CHPL_UNWIND=libunwind`, `CHPL_GMP=gmp`, and `CHPL_REGEXP=re2` set

nspark:
I noticed that `PCGRandom.RandomStream` behaves as expected with the following change:

```chapel
use Random.PCGRandom;

var R = new RandomStream(eltType=uint); // Added "eltType="
for i in 1..10 do
  writeln(R.getNext() % 100);
```

This is a bit surprising, and I think the case identified in the bug report should be supported as valid behavior.

bradcray:
@mppf: Can you take a look at this one?

mppf:
Another work-around is to use makeRandomStream(eltType=uint). The problem is that the compiler-generated constructor is being called in your first version:
```chapel
var R = new RandomStream(uint);
```

Really there are two problems in the implementation:
 1) The constructor in RandomStream does not prevent the compiler-generated constructor from being made
 2) Function resolution doesn't match actual arguments to formal arguments based upon type/param-ness. For example:
    ```chapel
    proc f(x:int = 1, type t = real) {
      writeln("in f(", x, ", ", t:string);
    }

    f(uint);
    ```

    does not compile.

I intend to look and the second issue but would prefer to wait for generic new-style initializers to be complete to solve the first issue.

bradcray:
I don't think that the example in item 2 above should compile.  By analogy, the following is not intended to compile:

    proc f(x: int = 1, y: string = "hello") {
      writeln("in f(", x, ", ", y);
    }

[...]

PR #5303, opened 2017-02-03
 make eltType mandatory first arg in RandomStream ctors

mppf:
Fixed issue #5281 by adjusting RandomStream constructor arguments. Before this PR, they are
``` chapel
      proc RandomStream(seed: int(64) = SeedGenerator.currentTime,
                        param parSafe: bool = true,
                        type eltType = real(64) )
```
and this PR changes them to
``` chapel
      proc RandomStream(type eltType,
                        seed: int(64) = SeedGenerator.currentTime,
                        param parSafe: bool = true)
```

since:

1) the type is more important, important enough to
    be the first argument and important enough to be required
2) the default of `real` only really made sense when NPB
    was the only RNG and it couldn't do integers.

(See the issue for details for why this change resolves the issue).

This PR additionally uses a function of last resort (with the "compiler generated" pragma) to create a deprecation error in cases that would require code changes.

Passed full local testing.
Passed quickstart+GASNet testing.

### DIFF
--- a/modules/standard/Random.chpl
+++ b/modules/standard/Random.chpl
@@ -185,6 +185,9 @@ module Random {
       The :mod:`NPBRandom` RNG will halt if provided an even seed.
       :mod:`PCGRandom` has no restrictions on the provided seed value.
 
+    :arg eltType: The element type to be generated.
+    :type eltType: `type`
+
     :arg seed: The seed to use for the PRNG.  Defaults to `oddCurrentTime` from
      :type:`RandomSupport.SeedGenerator`.
     :type seed: `int(64)`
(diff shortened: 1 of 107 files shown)


modules/standard/Random.chpl
```diff
@@ -2029,19 +2057,54 @@ module Random {
...
         :type parSafe: `bool`
 
-        :arg eltType: The element type to be generated.  Defaults to `real(64)`.
-        :type eltType: `type`
       */
+      proc NPBRandomStream(type eltType,
+                           seed: int(64) = SeedGenerator.oddCurrentTime,
+                           param parSafe: bool = true) {
+
+        // The mod operation is written in these steps in order
```
ben-albrecht: Is this a fix unrelated to the constructor argument re-ordering?

modules/standard/Random.chpl
```diff
@@ -210,6 +210,17 @@ module Random {
...
   }
 
+  // remove this deprecation version when appropriate
+  pragma "no doc"
+  pragma "compiler generated"
+  proc makeRandomStream(seed: int(64) = SeedGenerator.oddCurrentTime,
+                        param parSafe: bool = true,
+                        type eltType = real(64),
+                        param algorithm = defaultRNG) {
+    compilerWarning("makeRandomStream now requires eltType as first argument");
```
ben-albrecht: Really minor, but removing the word 'now' would make all of the warnings consistent.

ben-albrecht:
Aside from a really minor wording suggestion, it looks good. Thanks for the fix.

bradcray:
Sorry that I didn't get a chance to look at this sooner.  I think that the warning that's printed when the old constructors/routines are used should use the term "deprecated" to give a better sense that the user is using a routine that works today but is expected to go away in the near future.  The current wording suggests that you did something wrong, but then the code still works which seems surprising/confusing.

Thanks for remembering to add something to test/deprecated.

mppf:
Hi @bradcray - not a problem. Please see #5329 which tries to improve the wording. Let me know if you are happy with that wording or if you have other suggestions.

PR #5304, opened 2017-02-03
Add support for passing arbitrary env vars to paratest clients

ronawho:
Add the ability to do something like:

    $CHPL_HOME/util/test/paratest.server -env QT_AFFINITY=no

so that you can send arbitrary env vars to paratest clients. We already forward
the chplenv automatically, but we don't have any way (automatic or manual) to
set third-party or other arbitrary env vars.

To speed up testing and limit timeouts I like to run paratest with:

    QT_AFFINITY=no
    CHPL_TEST_LIMIT_RUNNING_EXECUTABLES=yes

I always set `CHPL_TEST_LIMIT_RUNNING_EXECUTABLES` in my bashrc, but I can't just
do that for `QT_AFFINITY` since that could affect performance runs, so I have to
remember to set it only for paratesting, and then remember to remove it when
I'm done. This gives me a way to only set these env vars for paratesting, and
it's a feature I've heard others want for similar reasons (setting gasnet
vars, etc.)

### DIFF
--- a/util/test/paratest.server
+++ b/util/test/paratest.server
@@ -495,10 +495,11 @@ sub find_files {
 
 
 sub print_help {
-    print "Usage: paratest.server [-compopts s] [-dirfile d] [-dirs d] [-execopts s] [-filedist] [-futures] [-logfile l] [-memleaks f] [-nodefile n] [-nodepara m] [-valgrind[exe]] [-help|-h] [-timeout t]\n";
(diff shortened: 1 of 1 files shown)


ronawho:
That's embarrassing.. I had intended to alphabetize it, but clearly my alphabet skills aren't what they used to be.

ronawho:
Alphabetization should be fixed now, thanks for catching that

Issue #5253, opened 2017-01-26
Improve detection of when types are used in the compiler


lydia-duncan:
### Summary of Problem
In my initializers work, I ran into a problem because our check on whether a class
is used or not depends on if the default constructor or type constructor is used.
In the case of initializers, these fields would be NULL on the AggregateType even
if the type was in use.  For my purposes, I am working around the problem, but
ideally this would be solved by adding actual analysis and tracking for type uses.

This issue is tied to the function `isUnusedClass`, currently defined in
functionResolution.cpp

### Configuration Information

- Output of `chpl --version`:
1.14.0

PR #5307, opened 2017-02-04
Run unusedExternRecordInProc.chpl with --verify

vasslitvinov:
Without --verify, compiler performs invalid memory accesses
when compiling this test. Adding --verify to ensure the failure
for nightly testing.

From .future:

* proc foo() is removed by `fn->defPoint->remove()` in removeUnusedFunctions()

* The type `timeval` gets removed in the process

* The type `_ref(timeval)`, however, sticks around. Its `_val` field still has the type `timeval`.

* `timeval` is still accessed as `_val` field's type during resolveIntents() - after it is cleaned up at end of resolve(). valgrind flags this access.


lydia-duncan:
Ooh, looks like another instance of #5253

PR #5309, opened 2017-02-05
Skip tests -nl>4 when running multinode multilocale on chapcs cluster

awallace-cray:
A few tests in release/examples want > 4 compute nodes (in multilocale mode), 
but Chapel's nightly build schedule as of 02/2017 only leaves 4 nodes available 
on the chapcs cluster. 
This change skips those tests just for the one multinode multilocale test config
which runs on chapcs cluster.

### DIFF
--- a/test/release/examples/primers/chplvis/chplvis1.skipif
+++ b/test/release/examples/primers/chplvis/chplvis1.skipif
@@ -1,0 +1,6 @@
+# Skip test "chplvis1" when running multinode multilocale on chapcs cluster
+
+# The test wants 5 compute nodes, but Chapel's nightly build schedule as of 02/2017
+# only leaves 4 nodes available on the chapcs cluster
+
+CHPL_NIGHTLY_TEST_CONFIG_NAME <= ^slurm-gasnet-ibv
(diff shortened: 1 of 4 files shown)


awallace-cray:
This applies the same skipif I added in Nov 2016 for chplvis3, to the other three tests with -nl > 4.

ronawho:
I'd prefer to keep using the more "automatic" `CHPL_TEST_NUM_LOCALES_AVAILABLE`. That way if somebody adds a new test that uses more than 4 locales we don't have to hope that we remember to add a skipif. It also prevents more clutter from being added to a repo.

We can downgrade the warning that `CHPL_TEST_NUM_LOCALES_AVAILABLE` so that it doesn't crop up in nightly mails or cause jenkins to report errors.

awallace-cray:
Yes lets do keep CHPL_TEST_NUM_LOCALES_AVAILABLE in the global env, because it helps let us know if some new test shows up w nl > 4 and we are skipping that test on chapcs. 

If we change sub_test so it never says WARNING when nl > CHPL_TEST_NUM_LOCALES_AVAILABLE, then we would likely never know if any more tests get skipped for that reason. Until we noticed it for some other reason.  I like that it throws an error.

awallace-cray:
See PR #5312

PR #5312, opened 2017-02-06
Use CHPL_TEST_NUM_LOCALES_AVAILABLE to limit demand for compute nodes…

awallace-cray:
… on chapcs cluster

For scheduling reasons, we now use global env variable CHPL_TEST_NUM_LOCALES_AVAILABLE
to limit the max number of compute nodes that multi-node multi-locale tests can
request, on the chapcs cluster.

This change:
- Removes the word "Warning" from the log message issued when a test is skipped
  for that reason. The word "Warning" causes the test harness to mark the test as
  an "error"; instead, we just want to silently skip the test.
- Deletes a skipif file that was inserted a few months ago, which is no longer needed
  now that we are using CHPL_TEST_NUM_LOCALES_AVAILABLE.


awallace-cray:
Take your pick.

PR #5316, opened 2017-02-07
Add a simple slurm-aware paratest wrapper for chapcs

ronawho:
Add a basic wrapper for paratest that does an salloc and runs paratest. It
essentially does:

    salloc --nodes=${num_free_nodes} --partition=chapel --share paratest.server

You use it just like you do paratest.server, except that you don't set
-nodefile or -nodepara. The wrapper automatically determines how many nodes to
use by calculating how many nodes are not reserved exclusively. It also does
some things so make paratests play nicer with other paratests in order to avoid
timeouts.

This should allow us to run a std paratest in ~20 minutes, without interfering
with nightly testing, and without causing timeouts for ourselves, or other
developers. It should also make it easy for devs to grab a node exclusively
during the day to run performance tests. They'll have to wait for existing
paratests to finish, but paratests started after an exclusive reservation will
leave nodes open for that job.

With regular paratest I often see timeouts even if just one other person is
running paratest. With this, I was able to run 6 concurrent paratests (max
allowed by slurm) without getting any timeouts. I tested with both gasnet and
std configuration paratests.

Some details about the wrapper:
-------------------------------

To calculate the number of nodes to use we: use `sinfo` to determine how many
nodes are online, and `squeue` to determine how many nodes are reserved by
non-shared jobs. This allows us to run on all nodes not being used exclusively
(by nightly testing, or by a developer wanting to do performance testing or
something.)

The wrapper does a few other things:
 - automatically determines a "good" nodepara so that testing runs faster
 - throws `--share --nice` to `salloc` to share slurm resources
 - sets `CHPL_TEST_LIMIT_RUNNING_EXECUTABLES=yes`, `QT_AFFINITY=no`, and
   `QT_SPINCOUNT=300` to limit timeouts

### DIFF
--- a/util/test/paratest.chapcs
+++ b/util/test/paratest.chapcs
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+
+#
+# Simple slurm-aware paratest wrapper for testing on the chapcs cluster. This
+# script aims to run parallel testing as quickly as possible without
+# interfering with nightly testing or other exclusive reservations (e.g.
+# developers running performance experiments.) It also tries to interfere as
+# little as possible with other paratests in order to avoid timeouts.
+#
+
+# Get the number of nodes reserved exclusively on the chapel partition
+get_num_exclusive_nodes() {
+  local delim=","
+  # Grab the "SHARED,NODES" info for all jobs. For each exclusive (SHARED=no)
+  # job add up the number of nodes it's using
+  local squeue_output=$(squeue --partition=chapel --noheader --format="%h${delim}%D")
+  local num_exclusive_nodes=0
+  for job_info in ${squeue_output}; do
+    IFS=${delim} read -r -a split_job_info <<< "${job_info}"
+    local shared=${split_job_info[0]}
+    local num_nodes=${split_job_info[1]}
+    if [ "${shared}" == "no" ]; then
+      num_exclusive_nodes=$((${num_exclusive_nodes} + ${num_nodes}))
+    fi
+  done
+  echo ${num_exclusive_nodes}
+}
+
+# Get the number of nodes available for testing on the chapel partition (total - exclusive)
+get_num_non_exclusive_nodes() {
+  local num_online_nodes=$(sinfo --partition=chapel --noheader --responding --format="%D")
+  local num_exclusive_nodes=$(get_num_exclusive_nodes)
+  local num_non_exclusive_nodes=$((${num_online_nodes} - ${num_exclusive_nodes}))
+  echo ${num_non_exclusive_nodes}
+}
+
+# Get the number of shared jobs running on the chapel partition
+get_num_shared_jobs_running() {
+  local num_jobs=$(squeue --partition=chapel --noheader --format="%h" | grep "yes" | wc -l)
+  echo ${num_jobs}
+}
+
+# Get a "good" nodepara value: use up to 3 for comm=none testing, but limit to
+# 1 for comm!=none since that's already oversubscribed. 
+#
+# TODO: Consider increasing nodepara if no other shared jobs are running. Note
+# that we should wait to do this until everybody is using slurm on chapcs.
+get_good_nodepara() {
+  local nodepara=3
+  if [[ -n ${CHPL_COMM} && "${CHPL_COMM}" != "none" ]]; then
+    nodepara=1
+  fi
+  #if [[ $(get_num_shared_jobs_running) -eq 0 ]]; then
+  #  nodepara=$((${nodepara} + 1))
+  #fi
+  echo ${nodepara}
+}
+
+# Run paratest inside an salloc using all nodes that are not reserved
+# exclusively on the chapel partition. Throw `--share --nice` and turn off
+# affinity and limit how many executables can run at once so we play nice with
+# other testing going on.
+run_paratest() {
+  local nodepara=$(get_good_nodepara)
+  local num_free_nodes=$(get_num_non_exclusive_nodes)
+  local para_env="-env CHPL_TEST_LIMIT_RUNNING_EXECUTABLES=yes"
+  para_env="${para_env} -env QT_AFFINITY=no"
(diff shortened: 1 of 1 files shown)


ronawho:
@mppf mind reviewing?

@ben-albrecht feel like reviewing with an eye on bash syntax/style?

mppf:
I think the main question I have is about the use of --immediate. In particular, I'd expect that we're willing to tolerate some overlap of paratest runs (although I think I've seen timeouts with just 2 developers running them at once). But we almost certainly don't want e.g. 12 jobs running on all the nodes at once.

That leads me to think:
 * there's some number of paratests we can tolerate at once
 * if there are too many, it might be nice for the script to wait for me (so I can run it in screen, go to lunch, etc)

I can't remember if with --share SLURM already has this concept of maximum over-subscription.

ronawho:
Our slurm partition allows up to 6 --share'd jobs to run the same time, so 6 different paratests can run at the same time if they use this script. Setting `QT_AFFINITY=no` and `CHPL_TEST_LIMIT_RUNNING_EXECUTABLES=yes` is meant to alleviate timeouts, and I believe with those set we can have 6 std paratests running concurrently without timeouts

ronawho:
And `--immediate` probably isn't needed anymore. I originally added it when I was experimenting, and I didn't have as good of a way to determine how many nodes were being used exclusively. We _could_ probably remove it, though I'd prefer to leave it at least initially so that people can report any problems back to me sooner

ronawho:
Ran an experiment to see how many paratests can actually run concurrently. I'm basically just creating X copies of the repo and modifying the execution limiter to include the slurm job_id so that I can simulate running as multiple people myself, and then running X paratests concurrently by doing `cd $CHPL_HOME/test && $CHPL_HOME/util/test/paratest.chapcs` Note that I'm using the std configuration, so comm=none, tasks=qthreads, etc.

With this script a single paratest finishes in ~20 minutes. Running 3 paratests concurrently finishes in around ~28 minutes with no timeouts. Running 6 paratests concurrently finishes in ~42 minutes and 2 instances had a single timeout for parallel/forall/vass/ri-3-stress.not-fast

So I think it's definetly achieving it's goal of running paratests quickly, without interfering too much with other peoples concurrent paratests. 

On an empty machine, start_test test/parallel/forall/vass/ri-3-stress.not-fast.chpl takes 120s, so it seems like we could dial down the problem size for that particular test if we wanted to.

ronawho:
We could also lower `QT_SPINCOUNT` for paratests, and that would further improve how well qthreads plays with other qthreads instances. 

Lowering it to 300 (from default of 300,000) allows 6 concurrent paratests to finish in ~33 minutes with no timeouts, so that's probably worth setting

`QT_SPINCOUNT` just takes a numeric value, which isn't ideal. I'm hoping that  eventually https://github.com/Qthreads/qthreads/issues/38 adds a `QT_WAIT_POLICY` var where we can set something like `QT_WAIT_POLICY=PASSIVE` to reduce the amount of spin-waiting done while paratesting.

ronawho:
Single gasnet paratest finishes in ~65 min. I was able to run 6 concurrent gasnet paratests with no timeouts in ~92 minutes. That's a lot better than I was expecting, I thought for sure there'd be some timeouts.

ronawho:
And mostly so I can find it later, here's the script I'm using to test multiple paratest instances:

```bash
#/bin/bash

NUM_MACHINES=6
runOnChapcs() {
  local num_machines_used=0
  echo "running ${@}"
  for i in 0{0..9} {10..11} ; do
    ssh -q "chapcs$i" "${@}" &
    num_machines_used=$((num_machines_used + 1))
    if [[ ${num_machines_used} -eq ${NUM_MACHINES} ]]; then
      break
    fi
  done
  wait
  echo "DONE running ${@}"
}

repoNameCmd="chapel-\$(hostname)"
branchName="test-slurm-aware-paratest"

cloneCmd="git clone https://github.com/ronawho/chapel.git ${repoNameCmd} && cd ${repoNameCmd} && git checkout test-slurm-aware-paratest"
updateCmd="cd ${repoNameCmd} && git checkout master && git branch -D ${branchName} && git pull && git checkout ${branchName}"
cleanRepoCmd="cd ${repoNameCmd} && git clean -fdx ."
cleanTestCmd="cd ${repoNameCmd} && git clean -fdx test/"
rmRepoCmd="rm -rf ${repoNameCmd}"
makeCmd="cd ${repoNameCmd} && source util/setchplenv.bash && make -j"
paratestCmd="cd ${repoNameCmd} && source util/setchplenv.bash && cd test && \${CHPL_HOME}/util/test/paratest.chapcs"
reportResultsCmd="cd ${repoNameCmd}/test && cat Logs/\$(whoami).linux64.log.summary"

runOnChapcs "${cloneCmd}"
#runOnChapcs "${updateCmd}"
#runOnChapcs "${cleanRepoCmd}"
#runOnChapcs "${cleanTestCmd}"
#runOnChapcs "${rmRepoCmd}"
runOnChapcs "${makeCmd}"
runOnChapcs "${paratestCmd}"
runOnChapcs "${reportResultsCmd}"
```

mppf:
OK, I'm happy with this in concept. I still think we should remove --immediate, but we can do that later.

I didn't nit-pick on the bash script implementation itself, since I'm not a great basher, but I'm hoping that @ben-albrecht's look at it will cover that angle. (I didn't see anything wrong with it either, though, just things I'm unfamiliar with).

ben-albrecht:
I am happy with the bash code as is. My only nit-pick was that this wasn't written in python, but it's understandable given that the script is a fairly thin wrapper around some slurm commands.

ronawho:
This was mostly cobbled together from functions I had my bashrc. Agreed that python would be nice, but I think I'll wait to make that transition until we need to do something more complicated (like parsing args or do more complicated things with parsing slurm commands.)

PR #5318, opened 2017-02-07
Improve how 'CHPL_UNWIND' output is squashed in testing

bradcray:
An unanticipated impact of my change to unset CHPL_UNWIND in
start_test (#5184) is that if you've got CHPL_UNWIND set, build the
runtime, and then run start_test, you may not have the correct runtime
built and up-to-date since we specialize the runtime builds based on
the CHPL_UNWIND setting.  Thus, you could get an error saying that the
proper runtime configuration isn't built (even though you thought you
just built it), or you could inadvertantly test a stale runtime that
you had lying around.

This change backs out the unsetting of CHPL_UNWIND and switches
to the undocumented CHPL_RT_UNWIND variable that @ronawho 
noticed which serves as a means to squash CHPL_UNWIND output on
a run-by-run basis.

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -560,6 +560,9 @@ def check_environment():
     if "CHPL_DEVELOPER" in os.environ: # unset CHPL_DEVELOPER
         del os.environ["CHPL_DEVELOPER"]
 
+    if "CHPL_UNWIND" in os.environ:    # squash CHPL_UNWIND output
+        os.environ["CHPL_RT_UNWIND"] = "0"
+
     # Check for $CHPL_HOME
     global home
     if "CHPL_HOME" in os.environ:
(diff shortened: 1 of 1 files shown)


bradcray:
@mppf: Would you mind reviewing this?

ronawho:
Instead of just unsetting `CHPL_UNWIND`, I think we can set `CHPL_RT_UNWIND=0` to prevent stack unwinding at runtime: https://github.com/chapel-lang/chapel/blob/master/runtime/src/error.c#L102

The unwind .skipifs already respect that env var too

mppf:
@ronawho - good point!

bradcray:
Is CHPL_RT_UNWIND intended to be a user-facing feature?  If so, could it be documented?  git grepping the repository, it only shows up in .skipifs and I didn't see a mention of it in the merge messages either.

If we already have the ability to disable unwindings dynamically, that suggests to me that the compiler could support a CHPL_UNWIND=none compilation without requiring the runtime to be rebuilt if it was built in a CHPL_UNWIND!=none mode.  Is that right?  (this ties into an email conversation we were having about the degree to which the whole runtime really needs to be rebuilt when CHPL_UNWIND changes).

mppf:
Re CHPL_RT_UNWIND, yes it is intended to be user-facing.

Re not recompiling when you change CHPL_UNWIND - I'm having trouble seeing how that would work and be good. Of course we could trivially make CHPL_UNWIND behave like CHPL_RT_UNWIND... but most of the details I can come up here fall into the "probably a bad idea" category (in my view, at least).

bradcray:
> Re CHPL_RT_UNWIND, yes it is intended to be user-facing.

Opened #5320 to not lose this.

bradcray:
Updated PR to use CHPL_RT_UNWIND.

mppf:
Looks good, except the PR message needs updating.

PR #5321, opened 2017-02-08
ZMQ weather test fix

nspark:
This PR fixes failures with a ZMQ module test.

The ZMQ weather example/test program uses a pub/sub communication model. A ZMQ pub/sub connection is inherently unreliable (like UDP, unlike TCP). This was resulting in some automated test failures when message(s) would get dropped. This test preserves the example, but doesn't check the output, which was pseudo-randomly generated and subject to packet losses as described here.

Poking @awallace-cray, who brought this to my attention.


awallace-cray:
Approved - thanks!

PR #5327, opened 2017-02-09
suppressif for error handling tests

psahabu:
When `CHPL_HOST_COMPILER=pgi`, there is an unexplained error in the lowering of `TryStmt`. @noakesmichael and I have a strong suspicion of the PGI compiler itself, so these tests will be suppressed for now.

### DIFF
--- a/test/errhandling/inprogress/empty-throws.suppressif
+++ b/test/errhandling/inprogress/empty-throws.suppressif
@@ -1,1 +1,7 @@
+# pgi
+#
+# When PGI is the host compiler, the lowering of TryStmt fails for no clear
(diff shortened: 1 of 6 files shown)


ronawho:
Think you need to say `CHPL_HOST_COMPILER==pgi` as they are I think they'll trigger for `CHPL_HOST_COMPILER=prgenv-cray-pgi`, which you don't want

psahabu:
Gotcha, will do.

ronawho:
Mind adding an issue to track this, just so it doesn't get lost?

psahabu:
Will do.

PR #5328, opened 2017-02-09
Switched to using deinit()

vasslitvinov:
This is a follow-up to #5299.
[do not merge: the warnings needs to be taken out first]

* Added a compiler warning that flags functions named "~classname".

* Switched from ~classname() to deinit() in the code: modules, tests, spec.
Left "~R" naming in printouts, to reduce the volume of changes.

* Added a test that specifically tests ~classname() and the warning.

Non-trivial pieces:

* Changed the spec text in that the destructor is called "deinit",
not "~classname".

* In learnChapelInYMinutes() also adjusted the comments on the constructor
and destructor. Switched to "initializer" and "deinitializer" words
proactively. Promised compiler-generated initializer according to the current
world for constructors, and warned that it is going away soon.

* Left the old "~classname" naming in these tests:

    // to test "destructor name must match class/record name" warning
    classes/figueroa/BogusDestructors5
    classes/figueroa/BogusDestructors7

    // these test scanning or pasing "~"
    classes/diten/deleteDollar
    classes/diten/test_destroy2

    // tests both new and old syntax
    classes/deinitializers/deinitAndDestruct

Tested: linux64 - full suite; gasnet - multilocale tests


spec/Classes.tex
```diff
@@ -1160,8 +1160,8 @@ \subsection{Class Destructor}
 \index{destructor!classes}
 
 A class author may specify additional actions to be performed before a class object is
-reclaimed, by defining a class destructor.  A class destructor is a method that has the
-same name as the class prefixed by a tilde.  A class destructor takes no arguments (aside from
+reclaimed, by defining a class destructor.  A class destructor is a method named
```
lydia-duncan: Want to rename "destructors" to "deinitializers" while you're at it?  Or wait until the more major spec change comes along?  If you want to wait, I would add a note to the spec that the rewrite is coming, just in case we forget between now and after doc freeze

spec/Records.tex
```diff
@@ -341,8 +341,8 @@ \subsection{Record Destructor}
 \index{destructor!records}
 
 A record author may specify additional actions to be performed before record storage is
-reclaimed by defining a record destructor.  A record destructor is a method that has the
-same name as the record prefixed by a tilde.  A record destructor takes no arguments
+reclaimed by defining a record destructor.  A record destructor is a method named
+\chpl{deinit}.  A record destructor takes no arguments
```
lydia-duncan: Same comment here

lydia-duncan:
Aside from the comment I made on the spec change, this seems good to go to me!

modules/internal/Atomics.chpl
```diff
@@ -308,7 +308,7 @@ module Atomics {
     var _v:atomic_bool = create_atomic_bool();
 
     pragma "no doc"
-    inline proc ~atomicbool() {
+    inline proc deinit() {
```
ronawho: Just curious -- does it make sense for deinitializers to be marked inline?

vasslitvinov: I imagine so. The compiler inserts calls to deinitializers, right? If so, it is meaningful to talk about inlining at these call sites.

PR #5329, opened 2017-02-09
Update deprecation message wording

mppf:
Brad suggested using the word "deprecate" in it somewhere, so as to be less confusing to users.

Verified that these still pass:
 * test/deprecated
 * test/modules/standard/Random

Suggested by and reviewed by @bradcray - thanks!


bradcray:
Looks much clearer, thanks!  Here's a reminder to re-test / update your tests of the warning.

PR #5334, opened 2017-02-10
Beef up arrays mem tests

bradcray:
On the array views branch, I noticed some cases that were leaking but which weren't
caught by current tests of memory leaks so far related to whether the array view was
computed within a function or passed into it.  Happily these cases work on master.
To that extent, I beefed up Sung's arrays memory test to cover additional patterns
and created a variation of it that does similar patterns for rank change.  These will
be good stress tests for memory management on the array views branch(es).


bradcray:
@benharsh: These were the tests I was referring to.  Would you review them before I merge?

benharsh:
Looks good!

Issue #5310, opened 2017-02-05
Redesign bulk-transfer interface


benharsh:
## Existing Bulk-Transfer Interface
Here is our existing interface:
```chpl
proc BaseArr.dsiSUpportsBulkTransfer() param : bool;
proc BaseArr.dsiSupportsBulkTransferInterface() param : bool;
proc BaseArr.doiCanBulkTransferStride(viewDom) param : bool;

proc BaseArr.doiCanBulkTransfer(viewDom) : bool;
proc BaseArr.doiUseBulkTransfer(B) : bool;
proc BaseArr.doiUseBulkTransferStride(B) : bool;

proc BaseArr.doiBulkTransferFrom(Barg, viewDom);
proc BaseArr.doiBulkTransferToDR(Barg, viewDom);
proc BaseArr.doiBulkTransferFromDR(Barg, viewDom);
proc BaseArr.doiBulkTransfer(B, viewDom);
proc BaseArr.doiBulkTransferStride(B, viewDom);
```
I think the existing bulk-transfer interface is unnecessarily complicated by the number of functions and separation between contiguous, strided, and distributed transfers. I think we can develop a simpler and more flexible interface.

## Proposed Interface
With this new interface my hope is that we can avoid embedding the following questions into the API:
- Does this transfer involve a DefaultRectangular?
- Is this a strided transfer
- Can/should I use a bulk transfer?

Here is the proposed interface:
```chpl
// srcClass and destClass inherit from BaseArr
proc srcClass.doiBulkTransferToKnown(srcDom:domain, destClass, destDom:domain) : bool {}
proc destClass.doiBulkTransferFromKnown(destDom:domain, srcClass, srcDom:domain) : bool {}

proc srcClass.doiBulkTransferToAny(srcDom:domain, destClass, destDom:domain) : bool {}
proc destClass.doiBulkTransferFromAny(destDom:domain, srcClass, srcDom:domain) : bool {}
```
The following guarantees are made:
- srcClass.rank == srcDom.rank
- destClass.rank == destDom.rank
- srcDom.size == destDom.size

Implementers should know that:
- it is valid for srcClass.rank != destClass.rank (for rank changes)
- it is valid for the domains to not be distributed, even if the classes implement distributed domains

[...]

Issue #5289, opened 2017-02-01
Compiler-generated constructors are surprisingly non-generic for array fields


bradcray:
The compiler is currently surprising in the following respect.  Given a class with an array field like:

```chapel
class C {
  var X: [1..3] real;
}
```

the compiler-generated constructor requires the actual array argument passed to it to be a default rectangular array.  In contrast, if the user adds a constructor that resembles the one they might expect the compiler to create:

```chapel
proc C(IX: [1..3] real = [0.0, 0.0, 0.0]) {
  X = IX;
}
```

This constructor permits arrays of various domain maps to be passed in, as in other function call contexts.  Arguably, the compiler constructor should do the same thing.

Apparently, the reason that it isn't at present is because we skip the processing of formal array arguments in normalize.cpp's fixup_array_formals() routine for compiler-provided constructors or type constructors.

In PR #5325 I filed a test demonstrating the distinct behavior between the user and compiler constructors.

I ran into this issue in the context of the array views work, where array slices and other operations may introduce different domain maps for an array's representation.  For example, the following call wouldn't work in an array view domain map world:

```chapel
var A: [1..10] real;
var myC = new C(A[1..3]);
```

for reasons similar to the Block distribution case above -- the compiler would start considering A[1..3] to have a different domain map than default rectangular.

bradcray:
@lydia-duncan and @noakesmichael: Do either of you have any insights off the tops of your heads as to how the compiler-generated constructor might differ from my user constructor above in a way that would cause it not to resolve?  I even tried my best to create the moral equivalent to the compiler-defined constructor, but this one works as well:

    proc C(IX: [1..3] real = _defaultOf([1..3] real)) {
      X = IX;
    }

lydia-duncan:
What's the precise error message that you get?  I know that the default constructor is a little wonky at times, but I wouldn't expect it to result in that different of behavior . . .

bradcray:
It's nothing very interesting:

    testit.chpl:12: error: unresolved call '_construct_C(_array(ArrayViewSliceArr(real(64),int(64),DefaultRectangularDom(1,int(64),false),int(64),DefaultRectangularArr(real(64),1,int(64),false,int(64)))))'
    testit.chpl:1: note: candidates are: _construct_C(X: _array(DefaultRectangularArr(real(64),1,int(64),false,int(64)))) [265530]
    testit.chpl:12: error: unresolved call had id 577356

[...]

PR #4510, opened 2016-09-13
Add an optimization that removes the inner multiply in dsiAccess

ronawho:
Add a compiler optimization (--optimize-array-indexing) that removes the inner
multiply for array indexing when possible. The basic idea of the optimization
is that if the blk field is never modified, we know that blk(rank) will always
be 1, so we can optimize away the last blk multiply in dsiAccess().

In order to determine if blk is modified, we explicitly mark all functions that
might modify it with a pragma. Then, if none of those functions are called we
know blk cannot be modified, so it's safe to remove the multiply. The functions
that modify blk are adjustBlkOffStrForNewDomain() (called for strided reindexing
and strided slicing) and rankChange().

This is kind of a lame optimization because if any array requires blk, all
arrays will suffer the performance penalty. i.e. the optimization just checks
if any function modifies block, it doesn't do any contextual or type based
analysis. Longer term we expect array views will provide a way to remove the
inner multiply in a cleaner and more principled manner, so this optimization is
really just a stopgap until that comes along.

Until then, this gets us the performance benefit for most cases, particularly
for benchmarks like the PRKs, array microbenchmarks, lcals, and some of the
shootout benchmarks. It also allows us to remove the hackier assertNoSlicing
param, because we now optimize almost all the cases where assertNoSlicing
wouldn't break things anyways.

The generated code improvement from this optimization is also pretty sweet:

```chapel
// stream like code
for i in 1..10 do A[i] = B[i] + C[i];
```

used to produce:

```c
for (i = INT64(1); i <= INT64(10); i += INT64(1)) {
  sum = INT64(0);                         // dead store
  sum += ((int64_t)((i * coerce_tmp)));   // extra mult
  call_tmp7 = (coerce_tmp2 + sum);
  sum2 = INT64(0);
  sum2 += ((int64_t)((i * coerce_tmp3)));
  call_tmp9 = (coerce_tmp4 + sum2);
  ret = *(call_tmp9);
  sum3 = INT64(0);
  sum3 += ((int64_t)((i * coerce_tmp5)));
  call_tmp11 = (coerce_tmp6 + sum3);
  *(call_tmp7) = ((int64_t)((ret + *(call_tmp11))));
}
```

and now generates: 

```c
for (i = INT64(1); i <= INT64(10); i += INT64(1)) {
  call_tmp7 = (coerce_tmp + i);
  call_tmp9 = (coerce_tmp2 + i);
  call_tmp11 = (coerce_tmp3 + i);
[...]

PR #5338, opened 2017-02-10
Array views: Slice, Rank change, and Reindex

bradcray:
This PR changes the implementation of array slices, rank change slices, and reindexing operations. Traditionally, we have expected each domain map to support a dsi call that implements each of these three operations.  In the new approach, we support a domain map per operation to represent the result of that operation.  As a simple example, the slice domain map is represented by storing a reference to the array being sliced and the domain representing the slice.  From there, operations can be implemented by forwarding methods to the referenced arrays/domains.  For example, on an array view representing a slice, "indexing/random access" (dsiAccess in the domain map standard interface) can be performed by bounds checking against the referenced domain and then  (if it's in bounds) forwarding the dsiAccess to the array.

In this PR, the term "array view" refers both to these operations (since they provide a "view" on the elements of an original array, aliasing them rather than copying them) and to the new domain maps that implement these operations.  Like other domain maps, array views are essentially an implementation of an array and, as a result, support all operations that a typical array would.

This PR also uses the term "closed form" to refer to the old world in which we would represent a slice of a DefaultRectangular array using another DefaultRectangular array; the slice of a Block array using another Block array; etc.  (There was, in fact, no firm requirement in the old world that applying these operations to an array would result in an array of the same type, but in practice this is what every domain map that we're aware of did, and it's hard to know what else they would do short of resorting to a more general array view domain map as in this effort).

Advantages of array views
===================

This new approach has a number of advantages relative to the status quo, including:

Reducing the size and complexity of the dsi interface
-----------------------------------------------------
Historically, the three routines corresponding to these operations have been among the most complex for domain map authors to write.  In many cases, they would either fail to get implemented, or else would be implemented incompletely, containing halt()s for cases that we did not know how to handle (or could not handle in traditional "closed form" in which the same representation would be used for the result of the operation as the input array).  A lot of code has been removed from each domain map as a result of retiring these routines and a number of test cases that would run into halts now work.

Retiring the need for the '=>' operator in constructor contexts
The '=>' operator was introduced as a means of representing slices of block arrays using the same classes as the original block array.  Specifically, each "local block array" class stores a local array to represent the block of the global array that is owned locally.  When we were creating a slice of a block array in closed form, we would need that slice to alias the original array.  This meant that the array field of the result array would need to alias the array field of the original array.  The use of the => operator in the constructor context was added to the language and compiler to give domain map authors a means of doing this.

However, this has always felt like a very special case, and also a fragile one.  It requires a fair amount of fiddly special-case code in the compiler that several of us have run afoul of.  It also is (at least in its present form) reasonably inflexible -- e.g., if the original array and the new array have different static types, the feature does not work (and this has at times prevented us from adding new generic fields to arrays because it prevented such array fields from aliasing arrays whose generic fields differed).

This PR does not actually deprecate the feature, though we believe it now could be and hope to do this for the 1.15 release (see post-commit TODO list below).

Providing a more precise way of optimizing away inner multiplies
----------------------------------------------------------------
At compile-time, given an array, we can make a fairly accurate determination as to whether slices, rank changes, or reindexings of that array will need a non-unit multiplier to advance from one element to the next in the innermost dimension (i.e., to determine whether or not blk(rank) == 1).  An early attempt to optimize away such unnecessary inner multiplies by adding a 'param' to each array saying whether or not it's necessary was thwarted due to its breaking the use of '=>' operators in contexts like Block distributions due to type differences (as referred to in the previous section).  This caused us to fall back to a much more conservative, global, and instable compiler-based analysis, implemented in PR #4510).  Specifically, the application of an operation like a rank change to one array across an entire program would de-optimize the accesses to all arrays whether or not they were directly affected.

With this new approach, we could introduce such a param.  But better, because we no longer represent such operations in closed-form using DefaultRectangular arrays themselves, we now can know that DefaultRectangular arrays will never need that inner multiply.  Moreover, by examining an array view and the array(s) that it refers to, we can determine whether or not the inner multiply is required at compile-time for any given array expression that's very specific to that expression.

Creating distinct types for array view operations
------------------------------------------------
At times, the compiler has wanted to be able to distinguish between an array that "owns" its own storage vs. one that aliases/views another array's storage.  In the old closed form world, this was not possible/easy since the two array expressions would have the same static type.  In the new world, a slice expression can be distinguished from an originating array by reasoning about its domain map.  This permits the compiler to distinguish between the array types and for the standard modules to overload operations relative to slices vs. original arrays (for example).  At times, we discussed giving a domain map the option to use closed form if it wished to, but the current thinking is that the utility of having distinct types is strong enough that we want to avoid doing so without good reason.

(Note: Our traditional "good reason" was expected to be performance, but in this effort, we've managed to get the performance close enough to closed-form (and in some cases beating it as a result of such optimizations) that we are now less inclined to permit closed form due to the type benefits in the "always use array views" approach).

Downsides to array views
===================

more indirection (potentially/likely)
-----------------------------------
At a high-level, the main downside to this approach is that it adds indirection where it didn't exist before.  E.g., an array view class refers to other domain map classes and therefore requires a pointer chase anytime we need to forward a call to the component domains/arrays.  By contrast, closed-form  represented everything in the same way as the original components and therefore added no overhead when using them.  However, there are also tradeoffs in terms of cost of setting up the descriptors (e.g., setting up an array view slice descriptor can be much simpler than a closed-form array slice since it's ostensibly just tucking away some pointers).  That said, in the case of array views on DefaultRectangular arrays, we've optimized the representation of the array views to permit direct accesses to the underlying DefaultRectangular arrays' ddata fields without indirection into the original array descriptor.  This both reduces indirection overhead and increases setup time, yet in a way that stays true to the array view philosophy and benefits.

Positive impacts of this PR
====================

Better performance / Reduced communication
-----------------------------------------------
Using array views has generally modestly reduced communication requirements for many idioms on distributed arrays (as evidenced by the test/distributions/robust/arithmetic/performance/ suite of tests that count communications).  In addition, new/simplified optimizations (such as bulk copies between distributed arrays) have been enabled by array views resulting in sometimes significant performance impacts (e.g., 3x improvement on miniMD, 2x improvement on ISx for 16 locales on ugni/qthreads).

Reduced memory leaks
------------------------
As a result of this effort, leaks due to domain maps are reduced and we are not aware of new leaks caused by array views.

Retired futures
---------------
A number of futures are able to be retired due to increased generality in the new solution.  See below for details.

Backslides/Downsides to this PR
========================
In order to land it in a timely manner (or at all), this PR also makes some compromises relative to what's on master today which we plan to address in future work.

Domains of distributed rank-changed or reindexed arrays are localized
-----------------------------------------------------------------------
(This will be tracked as issue #5473)

Previously, for a code like:

````chapel
const D = {1..n} dmapped Block(...);
var A: [D] real;
foo(A[2..n-1]);
proc foo(X: [?D] ) { forall i in D ... }
````

The domain 'D' and the iterations of the forall loop within foo() would be distributed in such a way as to respect the original Block distribution.  In this PR, this domain is completely localized to the domain on which the task making the slice was called.  In order to support the traditional (and intuitive) interpretation, we need to support domain and distribution views --- the analogue of array views but for the domain and distribution (domain map) classes.  This remains future work.  A new future, test/arrays/views/viewPreserveDist.chpl has been added to track this issue.  Somewhat urprisingly, no existing tests ran afoul of this change, nor did existing benchmarks take a performance hit as a result of it that we're aware of.

Initializing array fields using views via the compiler-provided constructor
-------------------------------------------------------------------------
Due to issue #5289 and the fact that array views now have different types, passing a view into a compiler-provided constructor for a (default rectangular) array field no longer works, as it has on master.  The anticipated fix here is to make compiler-provided initializers similarly generic to the equivalent that a user would get if they created a formal of default rectangular type.  It is expected that this work will take place as we switch from creating compiler-provided constructors to compiler-provided initializers.  The simple workaround is to provide a user-defined constructor that takes in an array and assigns it to the field.

A few tests had to be rewritten to work around this issue (BergerRigoutClustering.chpl, tensor-workaround.chpl) and new futures were added to track the issue (initArrFieldWithSlice.chpl, tensor.chpl).

Type queries of view arguments
Under closed form, a function signature of the form:

````chapel
proc foo(x: ?t, y: t) { ... }
````

could be passed in an array and a slice of an array due to the fact that the compiler would consider them the same type.  Now that they're different types, this no longer works.  We could either say that this makes sense in the new world or that such type queries should be flexible w.r.t. domain map when 't' is a default rectangular, as we do in other formal argument contexts.

However, more surprisingly, if two array views are passed into such a function, it will not resolve.  The reason for this seems related to the way the compiler sets up the 't' type alias.  It seems like a copy is introduced, and since a copy of an array view yields a new array, the compiler decides that 't' is a default rectangular, but then won't permit the second slice argument to match against it.  One pre-existing test ran into this and has been split into a workaround version of itself and a futurized version of the original test (see test/functions/diten/refIntents*.chpl).

Replicated Dist testing regressions
-----------------------------------
With these changes there will be roughly 18 failures in the dist-replicated
testing configuration with incorrect program output. They fall into the
following categories:
1) 'internal error: compiler generated error' - We haven't dug much into these
   yet, but past experience suggests this is likely an iterator bug in the compiler.
2) Incorrect values in the array - We're pretty sure the problem is that the
   ArrayView iterators iterate over 'privDom', which will not be distributed in
   the rank-change and reindex cases.  This is related to the failure to preserve
   locality on domains/distributions noted above.
3) Higher communication counts - the cause is unknown at this time, but we'd
   prefer to solve the correctness issues before tackling these. They may also be
   related to the non-distributed 'privDom'.

These cases are being tracked in issue #5472.

At present, we are proposing taking these failures on master for the time being.
The tests are difficult to futurize because they are tested across other distributions
as well.

Details of Our Approach
==================

## New Array view files / classes

A new internal module has been added for each of the three array view operations.  Briefly, they store fields as follows:
* ArrayViewSlice: stores a reference to the slicing domain and the sliced array
* ArrayViewRankChange: stores a reference to the slicing domain (with degenerate dimensions removed) and the sliced array; stores an tuple of booleans indicating which dimensions are collapsed and a tuple of indices indicating the indices to use in those dimensions when converting a low-D index of the slicing domain into a high-D index of the original array.
* ArrayViewReindex: stores a reference to the reindexing domain and the reindexed array.

Most of the operations on these array views are reasonably straightforward.  See comments in the source code for more details.

### chpl__unalias

The ``chpl__unalias`` function now returns something instead of
modifying the argument in-place. For ArrayViews the chpl__unalias
function will return a copy of the ArrayView, resulting in a non-view
type. Otherwise it will return a shallow copy of the _array record.

Other ``chpl__unalias`` functions were updated to reflect this new
behavior.

### ArrayViews Now Copy-Out When Returned

In GitHub issue #5341 @benharsh presented the following code and started
a discussion about what should happen when an ArrayView is returned:

```chpl
config var test = false;

var glob : [1..10] int = 1..10;

proc ret() {
  if test {
    return glob[1..5];
  } else {
    return glob;
  }
}

writeln(ret());
```
We all felt that this should be a valid Chapel program. The tricky
part is that, unless handled specially, the branches return a different
type. We eventually decided that we should pursue the direction of always
copying out when returning an array. There is an ongoing discussion of
how to return without copying in user-code.  For the time being,
we've worked around this by inserting pragmas into user code
(see the "Unfortunate test changes" category below for affected
files).

The major changes are as follows:
- Add a new flag, FLAG_NO_COPY_RETURN, to mark functions that
  should not copy-out arrays. This flag has been added to various
  internal functions.
- Implement ``chpl__unref`` functions for Arrays (shallow copy) and
  ArrayViews (deep copy). The compiler will insert calls to ``chpl__unref``
  in functionResolution prior to resolving a function's return type.

The impact is that ArrayViews do not leak, based on our testing.

### Privatization

ArrayViews will be privatized if the underlying array is also
privatized. Initial multilocale performance testing suggests
that this isn't big issue relative to our historical approach.
Future efforts may want to look into privatizing lazily or
ceasing to privatize in hopes of leaning on remote value
forwarding more (?).

## Bulk-Transfer changes

Most bulk-transfer API methods/functions now accept an additional
argument, named ``viewDom``, that represents the domain of the
ArrayView involved in the transfer. @benharsh hopes to solidify the
bulk-transfer interface at a later time as discussed in GitHub issue
#5310

Generally, ArrayViews have a positive impact on the performance of
sliced bulk-transfer assignment between arrays as the overhead for
creating a slice is much lower. This is a known benefit for the ISx
benchmark and likely helps the miniMD benchmark as well.

### Optimizing Indexing

We can re-use the RAD record to precompute the correct
off/blk/str/origin info such that we can index into a ddata pointer
without doing the index translation normally required by the reindex and
rank-change ArrayViews. The RAD record was chosen to avoid creating
another very similar record. Locality has nothing to do with this change.

The helper functions (toReindex, toRankChange, toSlice) are based off of
the old dsi* functions involving those views.

Moves the BlockDist RAD accessor method into DefaultRectangular for
code-reuse.

### Miscellaneous Optimization Notes
- Caching the member access of ``_instance`` in the _domain record
  improved performance. Shouldn't our compiler or the backend compiler
  be able to do that for us?

##ReplicatedDist Changes

Prior to this change the 'targetLocales' field was fully generic.
While investigating bugs in tests using this distribution, it
seemed easiest to decide that the 'targetLocales' array must
be associative because that satisfies the general case. See the
comments at the top of the module for further information on
the potential associative nature of the targetLocales argument
to the ReplicatedDist constructor. As a result the 'sorted'
iterator is used to achieve deterministic results.

## StencilDist Changes

As part of this effort, StencilDist was improved as follows:
* Fixed dsiMember to return 'true' for indices in the fluff region
* Modified check in chpl_replaceWithDeepCopy to check if _instance types
  match
* Added 'checkBounds' method to ArrayViewSlice and use it in dsiAccess
* Created the ref/const-ref/value variants of dsiAccess
* Threaded a param called 'ignoreFluff' through the Stencil classes in
  order to support a fluff-less view of the array.

Changes by Source Category and File
============================

documentation changes
---------------------

doc/technotes/dsi.rst:
* removed array-view dsi interfaces

internal/standard module changes

modules/internal/ChapelArray.chpl:
* rewrote implementations of slice, rank-change, and reindex functions to use array views
  - cleaned up error overload of reindex() for non-default rectangular arrays
* new utility functions for array views:
  - chpl__isArrayView()
  - chpl__getViewDom()
  - chpl__getActualArray()
* memory management-related improvements:
  - removed chpl_replaceWithDeepCopy() -- no longer needed
  - applied pragmas to internal routines
* added/tightened up early exit conditions for _array.equals()
  - related to use of associative array for Replicated Dist

modules/internal/ArrayView*.chpl:
* the new array view domain map classes

modules/internal/DefaultRectangular.chpl:
* removed array-view dsi interfaces
* folded out inner multiplies for default rectangular arrays
  - inner multiply is now always 1
* added helper routines to RAD data structure _remoteAccessData
  - now used to optimize views on default rectangular arrays
* refactored I/O routines to support re-use within array views and ReplicatedDist
  - hoisted consecutive binary I/O optimizations from individual dsiSerialRead/Write routines to shared routine
* updated bulk-transfer interface
* added missing dsi* calls in support of local subdomain queries to doms/arrs
* moved bulkCommConvertCoordinate from DefaultRectangular to DSIUtil.chpl

modules/dists/DSIUtil.chpl:
* moved bulkCommConvertCoordinate routine here from DefaultRectangular

modules/dists/BlockCycDist.chpl:
modules/dists/BlockDist.chpl:
modules/dists/CyclicDist.chpl:
modules/dists/DimensionalDist2D.chpl:
modules/dists/ReplicatedDist.chpl
modules/dists/SparseBlockDist.chpl:
modules/dists/StencilDist.chpl:
* removed array-view dsi interface
* updated bulk-transfer interface
* add "no copy return" to dsiLocalSlice routines
* distribution-specific changes:
  - changed targetLocales to associative domain (ReplicatedDist only)
  - improved StencilDist as noted above (StencilDist only)

modules/standard/UtilReplicatedVar.chpl:
* changes related to target locales being an associative domain now

modules/internal/ChapelDistribution.chpl:
* memory management-related improvements
* added default behavior for dsiDisplayRepresentation()
* base class implementations of is*ArrayView() functions
* updated bulk-transfer interface

Compiler changes
----------------

compiler/include/flags_list.h:
compiler/resolution/callDestructors.cpp:
compiler/resolution/functionResolution.cpp:
modules/internal/ChapelBase.chpl:
modules/minimal/internal/ChapelBase.chpl:
* memory management-related improvements

New tests
---------

test/arrays/views/assignReindex.*:
test/arrays/views/assignSlice.*:
test/arrays/views/initWithSlice.*:
test/arrays/views/passReindex.*:
test/arrays/views/passSlice.*:
test/arrays/views/passSlice2.*:
test/arrays/views/printSlice.*:
* simple tests used to get things up and running

test/optimizations/bulkcomm/bharshbarg/arrayViews.*:
* test of bulk communication with array views

Test changes related to backslides
----------------------------------

test/arrays/views/viewPreserveDist.*:
* new future showing backslide for distributed domains for rank-change/reindex

test/arrays/views/initArrFieldWithSlice.*:
* new future showing backslide in terms of default constructors + views

test/studies/amr/.../BergerRigoutsosClustering.chpl:
* added constructor to work around inability to pass views to comp. ctors

test/users/shetag/tensor*.chpl:
* futurized test and added workaround version for inability to pass views to ctors

test/functions/diten/refIntents*.chpl:
* created future + workaround versions of test due to ?t / :t backslide

Test changes indicating improvements

test/.../blkCycSliceOOB.*:
* retired future which now works

test/distributions/dm/s8.*:
test/distributions/dm/s9.*:
test/distributions/dm/sx8.*:
* retired sx8.future and updated s8.* and s9.* to reflect re-enabled feature

test/distributions/robust/arithmetic/reindexing/*:
* all distributions now get same result, so simplified test jig

test/memory/.../domainMaps*:
* updated files to reflect leak-free domain maps across configurations

test/optimizations/rafa/rankChange.future:
* retired future which now works

test/distributions/robust/arithmetic/performance/multilocale/*.good
* reduced comm counts caused by this work

test/distributions/robust/arithmetic/reindexing/test_strided_reindex2.*:
* cyclic case no longer hits a halt(); simplified test jig as a result

Unfortunate test changes
------------------------

test/optimizations/bulkcomm/asenjo/ptransDR/v1/ptrans.chpl:
test/studies/amr/lib/grid/GridVariable_def.chpl:
test/studies/amr/lib/level/LevelVariable_def.chpl:
* update tests wanting to return array alias views to use a pragma for the time being

Improvements to existing tests
------------------------------

test/.../loc-arr-slice.chpl:
* expand to cover additional cases

Removed tests
-------------
test/.../removeInnerMultOpt/*:
* removed these tests as they're no longer relevent/useful

test/.../semantic-examples/9*
* remove tests in preparation for deprecating '=>' in constructors

Test maintenance
----------------

test/.../test_array_alias_field.chpl:
* rewrote tests to avoid passing '=>' of array view in constructor context

test/.../semantic-examples/1.5*.good:
test/.../semantic-examples/4*.good:
* updated tests to reflect copy-out when returning arrays

test/*/chunking.chpl:
test/distributions/bharshbarg/stencil/rankChange.chpl:
test/functions/vass/ref-intent-bug-2big.chpl:
test/optimizations/bulkcomm/.../stencil.chpl:
test/studies/.../hpl.hpcc2012.chpl:
* rewrote test to reduce or work around reliance on _value assumptions within source code

test/.../CMO_array.chpl:
* added missing dsiDims() routine

test/distributions/bharshbarg/stencil/slice.chpl:
* updated to reflect StencilDist improvements

test/distributions/dm/n.chpl:
* retired dsiReindex() routine

test/.../printModStuff/foo.good:
* updated module print order to reflect earlier uses of 'Reflection' module (or did we?)

test/optimizations/bulkcomm/bharshbarg/remote.good.orig:
test/optimizations/bulkcomm/bharshbarg/util.chpl:
* update to reflect bulk comm interface changes

test/optimizations/deadCodeElimination/.../countDeadModules.good:
* updated dead module count to reflect new array view modules

test/distributions/dm/t5a.compopts:
* increase default instantiation limit

test/distributions/deitz/test_distribution_syntax2.bad
* update module line numbers

======

TODO to get ready for code review:
- [x] close memory leaks
- [ ] investigate (+ fix?) single-locale performance degradations
  - [x] array vect ops
  - [x] fasta
  - [x] LCALS
  - [ ] 2D matrix multiplication
  - [x] less optimal fannkuch reduxes
  - [ ] string temporary copies (?)
- [x] check multi-locale performance
  - hpl.hpcc2012 is currently worse due to rewrite
- [x] quiet or fix tests that break due to compiler constructor array field issue ( #5289)
- [x] quiet or fix tests that break due to change in ?t / t behavior
- [x] update comm counts
  - [x] for non-block configurations of distribution robustness as well...
- [x] update/fix tests that rely overmuch on direct access to _value:
  - [x] hpl.hpcc2012
  - [x] functions/vass/ref-intent-bug-2big and optimizations/bulkcomm/asenjo/stencilDR/v2/stencil
- [x] review and clean up delta w.r.t. master
  - [x] ~~is _arrAlias unused now?  Can it be removed?~~
    - almost, but StencilDist still uses it...
  - [x] do we have pointless initCopy and/or autoCopy overloads? (identical to "normal" arrays?)
  - [x] ~~remove newAlias()~~
  - [x] merge chpl__unalias() overloads on arrays
- [x] test full array interface on array views to make sure something reasonable happens
  - [ ] including slices of distributed arrays...

Ideally, before merge:
- [ ] test other configs:
  - [x] baseline?
  - [x] --verify?
  - [x] numa?
  - [ ] llvm?

[moved other post-merge TODOs to issue #5474]


bradcray:
only 14 failures on --local testing:

new memory leaks
----------------
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]
[Error matching program output for memory/sungeun/refCount/arrays]

unresolved dsiTargetLocales (no bulk transfer for reindex?)
-----------------------------------------------------------
[Error matching compiler output for studies/cholesky/jglewis/version2/elemental/unsymmetric_indices/test_elemental_cholesky_unsymmetric_ranges]

unresolved oneDData (no bulk transfer for reindex?)
---------------------------------------------------

Need to update .good / retire test
----------------------------------
[Error matching program output for arrays/elliot/removeInnerMultOpt/stridedReindexRequiresInnerMult]

compiler-defined constructors need to be more generic w.r.t. array views
------------------------------------------------------------------------
[Error matching compiler output for arrays/deitz/part4/test_array_alias_field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field2]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field3]
[Error matching compiler output for studies/amr/advection/amr/AMR_AdvectionCTU_driver]
[Error matching compiler output for users/shetag/tensor]

Need to make :?t / :t more generic or rewrite test
[Error matching compiler output for functions/diten/refIntents]

Test tries to do too much surgery on domain maps, needs updating
----------------------------------------------------------------
[Error matching compiler output for functions/vass/ref-intent-bug-2big]
[Error matching compiler output for optimizations/bulkcomm/asenjo/stencilDR/v2/stencil]

bradcray:
Maybe for Ben?
==============

output difference (maybe due to use of associative target locales)
------------------------------------------------------------------
[Error matching program output for distributions/vass/wrapper_nl4_testReplicatedDist1]
[Error matching program output for distributions/vass/wrapper_nl4_testReplicatedVar1]
[Error matching program output for distributions/vass/wrapper_nl4_testReplicatedVar2]
[Error matching program output for release/examples/primers/distributions]

associative array zip mismatch
------------------------------
[Error matching program output for distributions/bradc/distEquality]

bulk transfer tracing differently
[Error matching program output for optimizations/bulkcomm/bharshbarg/remote]


Brad issues
===========

shape mismatch
--------------
[Error matching program output for arrays/deitz/test_remote_block_slicing]
[Error matching program output for distributions/robust/arithmetic/slicing/test_array_slicing3]
[Error matching program output for optimizations/bulkcomm/alberto/Block/2dStrideBlock]
[Error matching program output for optimizations/bulkcomm/alberto/Block/3dStrideBlock]
[Error matching program output for optimizations/bulkcomm/alberto/Block/AllCasesPerformance]
[Error matching program output for optimizations/bulkcomm/alberto/Block/perfTest]
[Error matching program output for optimizations/bulkcomm/alberto/Block/rafatest2]
[Error matching program output for optimizations/bulkcomm/alberto/Block/rankChange]
[Error matching program output for optimizations/rafa/AgRend]
[Error matching program output for optimizations/rafa/AgTest]
[Error matching program output for optimizations/rafa/2dAgRend]
[Error matching program output for optimizations/rafa/2dAgTest]
[Error matching program output for optimizations/rafa/AgWCopyTest]

unresolved init()
-----------------
[Error matching compiler output for distributions/robust/arithmetic/composite/test_composite1]
[Error matching compiler output for distributions/robust/arithmetic/composite/test_composite2]
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Norm]
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Random]
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Sort]
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Search]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_array_alias1]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_array_alias2]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing1]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing2]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing5]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_strided_reindex1]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing3]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_strided_reindex2]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing4]
[Error matching compiler output for distributions/robust/arithmetic/reindexing/test_reindexing6]
[Error matching compiler output for studies/cholesky/jglewis/version2/elemental/unsymmetric_indices/test_elemental_cholesky_unsymmetric_ranges]

comm differences
----------------
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/alloc]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/alloc_all]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assign]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assignAlias]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assignSlice]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/reduceAlias]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/reduceSlice]
[Error matching program output for performance/ferguson/array-assign-block-get]
[Error matching program output for performance/ferguson/array-assign-block]
[Error matching program output for performance/ferguson/array-assign-get]
[Error matching program output for studies/ssca2/performance/SSCA2_commDiags]

unresolved theData call
-----------------------
[Error matching compiler output for optimizations/cache-remote/ferguson/reduces_comm/ra_prefetch]


Unsurprising
============

new memory leaks
----------------
[Error matching program output for memory/sungeun/refCount/domainMaps]

Same as in --local testing
--------------------------
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]
[Error matching program output for memory/sungeun/refCount/arrays]
[Error matching compiler output for studies/cholesky/jglewis/version2/elemental/strided/test_elemental_cholesky_explicitly_symmetric_strided]
[Error matching program output for arrays/elliot/removeInnerMultOpt/stridedReindexRequiresInnerMult]
[Error matching compiler output for arrays/deitz/part4/test_array_alias_field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field2]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field3]
[Error matching compiler output for studies/amr/advection/amr/AMR_AdvectionCTU_driver]
[Error matching compiler output for users/shetag/tensor]
[Error matching compiler output for functions/diten/refIntents]
[Error matching compiler output for functions/vass/ref-intent-bug-2big]
[Error matching compiler output for optimizations/bulkcomm/asenjo/stencilDR/v2/stencil]

bradcray:
Testing is getting close enough that I'm going to condense into a single list unifying local and gasnet:

compiler-defined constructors need to be more generic w.r.t. array views
------------------------------------------------------------------------
[Error matching compiler output for arrays/deitz/part4/test_array_alias_field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field2]
[Error matching compiler output for arrays/ferguson/semantic-examples/9-array-alias-field3]
[Error matching compiler output for studies/amr/advection/amr/AMR_AdvectionCTU_driver]
[Error matching compiler output for users/shetag/tensor]

view's notion of domain isn't distributed properly
-------------------------------------------------
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/alloc]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/alloc_all]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assign]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assignAlias]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/assignSlice]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/reduceAlias]
[Error matching program output for distributions/robust/arithmetic/performance/multilocale/reduceSlice]
[Error matching program output for performance/ferguson/array-assign-block-get]
[Error matching program output for performance/ferguson/array-assign-block]
[Error matching program output for performance/ferguson/array-assign-get]
[Error matching program output for studies/ssca2/performance/SSCA2_commDiags]

Need to make :?t / :t more generic or rewrite test
[Error matching compiler output for functions/diten/refIntents]

Test tries to do too much surgery on domain maps, needs updating
----------------------------------------------------------------
[Error matching compiler output for functions/vass/ref-intent-bug-2big]
[Error matching compiler output for optimizations/bulkcomm/asenjo/stencilDR/v2/stencil]

New leaks
----------
[Error matching program output for memory/sungeun/refCount/domainMaps]
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]
[Error matching program output for memory/sungeun/refCount/arrays]

unresolved theData call
-----------------------
[Error matching compiler output for optimizations/cache-remote/ferguson/reduces_comm/ra_prefetch]

core dump in single-locale / non-local access in multi-locale
-------------------------------------------------------------

bradcray:
@mppf and @ronawho : I think this is ready to review now.  My recommendation is to start with the PR message and then when it gets to the list of sources and their changes, review them in that order.  (This is not a PR to review by commit, at least at the outset...).  Thanks!

compiler/resolution/functionResolution.cpp
```diff
@@ -8253,6 +8258,73 @@ static bool isVecIterator(FnSymbol* fn) {
...
+            // Remove now-useless AST
+            tmp->defPoint->remove();
+            init_unref_tmp->remove();
+            INT_ASSERT(unrefCall->inTree() == false);
+          } else {
+            // The array is an ArrayView
+            //
+            // Used by callDestructors to catch assignment from a ref to
+            // 'tmp' when we know we don't want to copy.
+            tmp->addFlag(FLAG_NO_COPY);
```
mppf: I'm OK with this code as-is but it seems some minor adjustments to the structure around lines 8294 - 8320 would make it easier to follow.

Since tmp is removed in the other half of this conditional, why not set FLAG_NO_COPY on it unconditionally just after the newTemp call?

Would it be more reasonable to leave call->get(2) in the tree until we get to the retType != tmp->typeInfo condition?

mppf:
I'm OK with the compiler changes.

doc/technotes/dsi.rst
```diff
@@ -636,44 +636,6 @@ as an indication of what procedure(s) need to be defined.
...
-  .. [TODO: explain the arguments.]
-
-.. [TODO: the following seem to support dsiReindex and dsiRankChange:]
-   .. method:: proc GlobalDistribution.dsiCreateReindexDist(newSpace, oldSpace)
-   .. method:: proc GlobalDistribution.dsiCreateRankChangeDist(param newRank: int, args)
-.. [TODO: the following seems correct. The returned object can be an instance
-  of a different class than the receiver, at the implementor's discretion.
-  That class must satisfy the requirements on "GlobalDistribution"
-  defined in this document.]
-
```
mppf: `doiCanBulkTransferStride(viewDom) param`  is a new signature in this PR, right? Is that documented in the DSI interface? At least add a TODO if it's not?

modules/dists/DSIUtil.chpl
```diff
@@ -470,3 +470,35 @@ proc setupTargetLocalesArray(targetLocDom, targetLocArr, specifiedLocArr) {
...
+// in support of ArrayViews.
+//
+// TODO: assert that bView and aView are rectangular?
+// TODO: If we wanted to be more general, the domain args could turn into
+//       tuples of ranges
+//
+proc bulkCommConvertCoordinate(ind, bView:domain, aView:domain)
+{
+  assert(bView.rank == aView.rank);
+  const b = chpl__tuplify(ind);
```
mppf: Isn't this the same as (ind,) ?

modules/dists/DSIUtil.chpl
```diff
@@ -470,3 +470,35 @@ proc setupTargetLocalesArray(targetLocDom, targetLocArr, specifiedLocArr) {
...
+// In the SBAC'12 paper it is called m(). Later modified by Ben Harshbarger
+// in support of ArrayViews.
+//
+// TODO: assert that bView and aView are rectangular?
+// TODO: If we wanted to be more general, the domain args could turn into
+//       tuples of ranges
+//
+proc bulkCommConvertCoordinate(ind, bView:domain, aView:domain)
+{
+  assert(bView.rank == aView.rank);
```
mppf: Should/could this assert be a compilerError ?

modules/dists/DSIUtil.chpl
mppf: Oh I see, no, it's making a tuple from 1 thing or leaving a tuple a tuple.

mppf: In that case should we also compileError / assert that b.size == rank ?

modules/internal/DefaultRectangular.chpl
```diff
@@ -705,6 +716,164 @@ module DefaultRectangular {
...
+    if earlyShiftData && !stridable {
+      type idxSignedType = chpl__signedType(idxType);
+      const shiftDist = if isIntType(idxType) then origin - factoredOffs
+                        else origin:idxSignedType - factoredOffs:idxSignedType;
+      shiftedData = _ddata_shift(eltType, data, shiftDist);
+    }
+  }
+
+  proc _remoteAccessData.toSlice(newDom) {
+    compilerAssert(defRectSimpleDData && this.rank == newDom.rank);
```
mppf: Will this assert lead to NUMA failures? Or is this RAD stuff disabled there already?

modules/internal/DefaultRectangular.chpl
```diff
@@ -705,6 +716,164 @@ module DefaultRectangular {
...
+    rad.computeFactoredOffs();
+    rad.initShiftedData();
+
+    return rad;
+  }
+
+  proc _remoteAccessData.toRankChange(newDom, cd, idx) {
+    compilerAssert(defRectSimpleDData && this.rank == idx.size && this.rank != newDom.rank);
+    // TODO: If 'collapsedDims' were param, we would know if blk(rank) was 1 or not.
+    var rad : _remoteAccessData(eltType, newDom.rank, newDom.idxType, newDom.stridable, true);
```
mppf: is this the only place that sets blkChanged = true (vs forwarding it along) ? If so, it's probably worth a comment that mentions blkChanged.

modules/internal/DefaultRectangular.chpl
```diff
@@ -1186,25 +1355,14 @@ module DefaultRectangular {
           // If we detect that blk is never changed then then blk(rank) == 1.
           // Knowing this, we need not multiply the final ind(...) by anything.
           // This relies on us marking every function that modifies blk
-          if __primitive("optimize_array_blk_mult") {
```
mppf: I think the comment above is now out of date, right? Also, can we now remove the "modifies blk" flag?

modules/internal/DefaultRectangular.chpl
```diff
@@ -2179,10 +2070,16 @@ module DefaultRectangular {
...
-                && mdRLen == B._value.mdRLen));
+    if this.rank != B.rank {
+      return false;
+    } else if defRectSimpleDData {
+      return true;
+    } else {
+      const actual = chpl__getActualArray(B);
+      return mdParDim == actual.mdParDim
+      && mdNumChunks == actual.mdNumChunks
+      && mdRLen == actual.mdRLen;
```
mppf: Seems these && lines should be indented more.

mppf:
For the RAD optimization for DR array views - can we also have that optimization for plain old DR arrays?

DefaultRectangular: RAD stuff is a little opaque to me. Is most of this code moved from somewhere else? Bulk transfer changes are in the same category.

BlockCyclic, SparseBlockDist: OK
DSIUtil: OK

I definately still need to look at:
 * ChapelArray changes
 * BlockDist changes
 * new array view modules
 * test changes

modules/internal/DefaultRectangular.chpl
benharsh: No, blkChanged can also be set in ``toSlice`` if the new domain is stridable.

modules/internal/DefaultRectangular.chpl
benharsh: This RAD stuff is currently disabled for NUMA.

benharsh:
I'm not exactly sure what you mean by using the 'RAD' optimization for plain-old DR arrays. DefaultRectangular should now be able to avoid the unnecessary multiplication by ``blk(rank)``, so I'm not sure what else there is to do there. Unless you're talking about the actual 'remoteness' part? If it wasn't clear in the PR message, I'm simply re-using the RAD record to implement this 'index cache' optimization. It has nothing to do with optimizing remote accesses of ArrayViews.

Most of the RAD-related stuff in DefaultRectangular.chpl is copied from existing or removed functionality. I'll add comments explaining that.

Can you expand more on what is unclear about the bulk-transfer changes?

doc/technotes/dsi.rst
benharsh: On it.

modules/dists/DSIUtil.chpl
benharsh: I'll change this to a compilerAssert and add a check for ``b.size == rank``.

modules/internal/DefaultRectangular.chpl
benharsh: On it.

doc/technotes/dsi.rst
benharsh: Looks like we don't document the bulk-transfer interface anywhere. I'd be inclined to hold off until I implement the changes in my proposal described in #5310 (provided they pass design review).

modules/internal/DefaultRectangular.chpl
ronawho: I'll be removing the flag and any remnants of the old optimization in a separate PR.

compiler/resolution/functionResolution.cpp
benharsh: Doesn't ``unrefCall`` need to be in the tree for ``resolveNormalCall`` to work? That was my impression and it seemed reasonable to insert it where ``call->get(2)`` used to be.

I will move the ``addFlag``.

modules/internal/DefaultRectangular.chpl
mppf: I think it'd be nice to call out the places that set this in comments. When I was trying to understand that field, I searched for blkChanged, and this one at least isn't findable that way.

mppf:
re bulk transfer changes: I've never really followed the bulk transfer code generally, so I'm having some trouble following the changes. But for now let me focus on getting through the other changes.

doc/technotes/dsi.rst
mppf: Can you just add a TODO in the dsi.rst and in the PR message?

compiler/resolution/functionResolution.cpp
mppf: Yes that's right, but you don't have to remove call->get(2) in order to create unrefCall. You could copy it and then replace it in the "Array View" case branch.

doc/technotes/dsi.rst
benharsh: Yup, can-do.

compiler/resolution/functionResolution.cpp
mppf: But remember I'm OK with it as-is, it just seems like it could be simpler somehow...

modules/internal/ChapelBase.chpl
```diff
@@ -1032,20 +1032,23 @@ module ChapelBase {
     return if x != 0i then true else false;
 
   pragma "dont disable remote value forwarding"
+  pragma "no copy return"
```
mppf: I'm not very familiar with _createFieldDefault. Can you explain why it needs "no copy return" ?

It seems unlikely to me that the param and nilType versions could ever be returning an array view.

benharsh: It needs a "no copy return" because it may be used to initialize an array field (from what I could tell while debugging leaks/bugs).

You're probably right about the param and nilType versions, I likely just applied it to all of them without thinking much. I'll remove them and add a comment.

mppf: It's OK to leave it for now, but I'd expect this function to need neither "no copy return" nor "no auto destroy".

modules/internal/ChapelBase.chpl
```diff
@@ -1099,12 +1102,21 @@ module ChapelBase {
...
-  inline proc chpl__unalias(ref x) { }
+  inline proc chpl__unalias(x) {
+    pragma "no copy" var ret = x;
+    return ret;
+  }
 
   pragma "unalias fn"
-  inline proc chpl__unalias(x:_iteratorClass) { }
+  inline proc chpl__unalias(x:_iteratorClass) {
+    pragma "no copy" var ret = x;
```
mppf: could you add a comment here to indicate that this will return an array storing the result of the iterator?

modules/internal/ChapelBase.chpl
```diff
@@ -1099,12 +1102,21 @@ module ChapelBase {
...
+    return ret;
+  }
   pragma "unalias fn"
-  inline proc chpl__unalias(const ref x:_iteratorRecord) { }
+  inline proc chpl__unalias(const ref x:_iteratorRecord) {
+    pragma "no copy" var ret = x;
```
mppf: and here

modules/internal/ChapelArray.chpl
```diff
@@ -495,6 +500,7 @@ module ChapelArray {
   }
 
 
+  pragma "no copy return"
```
mppf: This "no copy return" seems to be not so obvious to me. I'm often confused by runtime types but if you know something about why it's necessary here I think a comment is called for.

modules/internal/ChapelArray.chpl
```diff
@@ -610,6 +616,46 @@ module ChapelArray {
...
   }
 
+  //
+  // Array-view utility functions
+  //
+  proc chpl__isArrayView(arr) param {
+    use Reflection;
+    param isSlice = if canResolveMethod(arr, "isSliceArrayView") then arr.isSliceArrayView() else false;
+    param isRankChange = if canResolveMethod(arr, "isRankChangeArrayView") then arr.isRankChangeArrayView() else false;
+    param isReindex = if canResolveMethod(arr, "isReindexArrayView") then arr.isReindexArrayView() else false;
```
mppf: Couldn't these be written like this:
``` chapel
  proc isSliceArrayView( x ) param return false;
  proc isSliceArrayView ( x: ArrayViewSliceArr ) param return true;
  param isSlice = isSliceArrayView(arr);
```

etc?

It seems unnecessary to use canResolve here.

modules/internal/ChapelArray.chpl
```diff
@@ -610,6 +616,46 @@ module ChapelArray {
...
+  // Returns a domain with a rank equivalent to chpl__getActualArray(arr).rank.
+  // This domain is no larger than the innermost array's domain. It represents
+  // the 'active' indices that the top-level ArrayView works with. For example:
+  //
+  //   var A : [1..10, 1..10];
+  //   var B => A[1, 1..10];
+  //   writeln(chpl__getViewDom(B)); // {1..1, 1..10}
+  //
+  proc chpl__getViewDom(arr: []) {
+    if chpl__isArrayView(arr._value) then return arr._value._getViewDom();
```
mppf: I wonder if this should be similiarly flexible to getActualArray so that it can accept an array class in addition to an array?

modules/internal/ChapelArray.chpl
benharsh: You're right, I think I used 'canResolve' during some intermediate state. I'll clean it up.

modules/dists/BlockDist.chpl
```diff
@@ -1641,90 +1494,76 @@ proc BlockArr.doiUseBulkTransferStride(B) {
...
 }
 
-proc BlockArr.doiBulkTransfer(B) {
+proc BlockArr.doiBulkTransfer(B, viewDom) {
   if debugBlockDistBulkTransfer then
     writeln("In BlockArr.doiBulkTransfer");
+  const actual = chpl__getActualArray(B);
+  // TODO: how to avoid a deep copy of the domain here? If it's distributed,
+  // it can cause a lot of comms...
+  pragma "no copy" const actDom = chpl__getViewDom(B);
```
mppf: I find this pragma "no copy" suspect and the comment next to it too.
The reason is that I wouldn't expect
  const actdom = chpl__getViewDom(B)
to create any copy at all in the calling code. If there is any copy made, it would be within chpl__getViewDom, in which case the pragma "no copy" would have no effect.

Does it work to simply remove this pragma?

modules/dists/ReplicatedDist.chpl
```diff
@@ -177,10 +177,10 @@ when the constructor encounters an error.
 
 */
 class ReplicatedDist : BaseDist {
+  var targetLocDom : domain(Locales.idxType);
```
mppf: Can you say more about what motivated this change to always associative targetLocDom ? I don't see anything about why it was done in the PR message.

modules/dists/StencilDist.chpl
```diff
@@ -1531,6 +1464,7 @@ iter StencilArr.dsiBoundaries(param tag : iterKind) where tag == iterKind.standa
 //
 // Copies the range slice function in ChapelArray.
 //
+pragma "no copy return"
```
mppf: Should noFluffView() return an actual Array View instead of a Stencil ?

mppf: (I mostly mean eventually, it probably doesn't make sense now b/c the view might not be distributed right)

modules/internal/ChapelDistribution.chpl
```diff
@@ -581,7 +587,19 @@ module ChapelDistribution {
...
+      return false;
+    }
+
+    proc isRankChangeArrayView() param {
+      return false;
+    }
   
+    proc isReindexArrayView() param {
+      return false;
+    }
```
mppf: I'm not sure these belong here. Two reasons:
 1) I think it could be a simpler query
 2) I thought we were checking for these with canResolve

modules/internal/ChapelDistribution.chpl
```diff
@@ -785,8 +803,11 @@ module ChapelDistribution {
     delete dist;
   }
 
-  proc _delete_dom(dom:BaseDom, param privatized:bool) {
-    dom.dsiDestroyDom();
+  proc _delete_dom(dom, param privatized:bool) {
+    use Reflection;
+    if canResolveMethod(dom, "dsiDestroyDom") {
+      dom.dsiDestroyDom();
```
mppf: This conditional seems odd to me. Wouldn't it be an error not to implement dsiDestroyDom?

modules/dists/ReplicatedDist.chpl
benharsh: I will update the PR.

modules/internal/ChapelDistribution.chpl
```diff
@@ -798,7 +819,10 @@ module ChapelDistribution {
   // that arr.eltType is meaningful.
   proc _delete_arr(arr, param privatized:bool) {
     // decide whether or not the array is an alias
-    var isalias = (arr._arrAlias != nil);
+    var isalias = ((arr._arrAlias != nil) || 
+                   arr.isSliceArrayView() || 
+                   arr.isRankChangeArrayView() ||
+                   arr.isReindexArrayView());
```
mppf: You couldn't call 'chpl__isArrayView' ? Is this an order-of-resolution issue or something?

modules/dists/BlockDist.chpl
benharsh: Unfortunately it seems that the pragma is useful in this case. Without it, the distributions/robust/arithmetic/performance/multilocale/assignAlias test sees a large increase in communication stats (gets, ons, etc.).

mppf:
ChapelArray: OK
CyclicDist: OK, but I havn't dug into the bulk transfer changes so much
BlockDist: OK, similarly cursory view of bulk transfer changes
Replicated: OK (but I wonder why we made targetLocDom always associative)
StencilDist: OK

I think I've looked over all the modules except for the array view .chpl files themselves. I need to look at those and the tests.

modules/dists/StencilDist.chpl
benharsh: Long-term, yes. For this effort I attempted to do so but observed an increase in leaks. I'd like attempt more, but would rather leave that as a post-merge item.

modules/internal/ChapelDistribution.chpl
benharsh: Regarding point two: It was my mistake to use canResolve, and I should have been using these methods to begin with.

modules/dists/BlockDist.chpl
mppf: That's still surprising to me. If it were an *array* it wouldn't be so surprising... but it's returning a domain, and we don't yet have any "domain view"s that would cause it to be a deep copy. Perhaps it's just a compiler TODO.

modules/internal/ChapelDistribution.chpl
benharsh: This is a (hopefully) temporary workaround for a bug with recursive iterators. I will add a comment clarifying that and pointing readers to a GitHub issue.

modules/dists/StencilDist.chpl
mppf: Let's just make sure it's a TODO comment in the source code then.

modules/internal/ChapelDistribution.chpl
mppf: right but can't this just be checking for : ArrayViewSliceArr ?

modules/internal/ArrayViewSlice.chpl
```diff
@@ -0,0 +1,391 @@
...
+
+
+    //
+    // introspection routine used elsewhere to filter array views
+    //
+    // TODO: Could this be replaced with more type-based introspection?
+    // I shied away from it since this is a generic class, but there
+    // must be (or should be) some way to do it without relying on
+    // methods like this...
+    //
```
mppf: This works for me on your branch:
``` chapel
var A : [1..10] int;

proc isArrayViewSliceArr( x ) param return false;
proc isArrayViewSliceArr( x : ArrayViewSliceArr ) param return true;

writeln( isArrayViewSliceArr(A._value) );
writeln( isArrayViewSliceArr(A[2..3]._value) );
```

If there is an expectation that there will be other array view types implemented, I think it makes sense to retain the isSliceArrayView etc methods. If not, switching to the above is better (assuming you don't run into some problem related to resolving it all in internal modules).

modules/internal/ArrayViewSlice.chpl
```diff
@@ -0,0 +1,391 @@
...
+    // accessors
+    //
+
+    inline proc dsiAccess(i: idxType ...rank) ref {
+      return dsiAccess(i);
+    }
+
+    inline proc dsiAccess(i: idxType ...rank)
+    where !shouldReturnRvalueByConstRef(eltType) {
```
mppf: I like to indent where clauses but you might prefer this style. Just making sure it's not an accidental.

modules/internal/ArrayViewSlice.chpl
```diff
@@ -0,0 +1,391 @@
...
+
+    proc dsiHasSingleLocalSubdomain() param
+      return privDom.dsiHasSingleLocalSubdomain();
+
+    proc dsiLocalSubdomain() {
+      return privDom.dsiLocalSubdomain();
+    }
+
+    proc dsiNoFluffView() {
+      if canResolveMethod(arr, "dsiNoFluffView") {
```
mppf: is this canResolveMethod here just to give a better error message? If that's the case, I'd like a comment to indicate that is the purpose (vs a workaround for a compiler bug).

modules/internal/ArrayViewSlice.chpl
```diff
@@ -0,0 +1,391 @@
...
+      arr.doiBulkTransferToDR(B, viewDom);
+    }
+
+    proc doiBulkTransferFromDR(B, viewDom) {
+      arr.doiBulkTransferFromDR(B, viewDom);
+    }
+
+    proc doiBulkTransferFrom(B, viewDom) {
+      arr.doiBulkTransferFrom(B, viewDom);
+    }
```
mppf: There are like 20 methods here that forward to arr. Is this a use case for delegation/forwarding?

mppf:
The array views modules (slice, rank change, reindex) look OK. I'm going to look at the tests now.

test/arrays/views/assignReindex.chpl
```diff
@@ -0,0 +1,21 @@
...
+  a = 1.0;
+
+writeln(A, "\n");
+
+ref B = A.reindex({2..10, 0..8});
+
+forall (i,j) in B.domain do
+  B[i,j] = i + j/10.0;
+
+writeln(A, "\n");
```
mppf: I think GitHub is telling me that this file has a zero byte or something at the end. Could you try removing the last several characters?

test/arrays/views/assignSlice.chpl
```diff
@@ -0,0 +1,19 @@
...
+A[5..6, 7..8] = 0.0;
+
+writeln(A, "\n");
+
+forall a in A[5..6, 7..8] do
+  a = 1.0;
+
+writeln(A, "\n");
+
+forall a in A[5..6, 7..8] do
```
mppf: Just curious if there's a reason this test doesn't use something like
```
ref B = A[5..6, 7..8];
forall a in B do a = 2.0;
```

ronawho:
I'm really happy with the single and multi-locale performance results. I'm also excited that this removes all sungeun/refCount/domainMaps*.good files.

The rad record and bulk comm changes looks good to me. I spent an hour or two looking at this last night, but I don't think I have any particularly helpful review feedback. I think this is good to merge once any outstanding feedback from Michael has been addressed.

doc/technotes/dsi.rst
ronawho: I tend to find TODOs that are only in the code almost never get fixed and are very rarely helpful for developers that come along later.

I would prefer that we open a new issue to track these sorts of TODOs. I'm ok with the TODOs being in the source too, I would just prefer that's not the only place they're tracked.

benharsh: I will start tracking this in #5310, if that works for you.

modules/internal/DefaultRectangular.chpl
bradcray: Good catch on the comment.  Now you know why I rarely write comments -- It seems I rarely read them either.  :)

Tacking onto Elliot's note:  While it was tempting to pull out all Elliot's inner-mult-optimization code in this PR, we decided to wait for a separate PR for that (I pulled the tests of his optimization out because they became harder to maintain than seemed worthwhile).

modules/dists/ReplicatedDist.chpl
bradcray: There's a description of the ReplicatedDist changes in the PR message now (provided by @benharsh)

modules/internal/ChapelDistribution.chpl
bradcray: I had (and I think still have?) a comment in the code defining these routines on array views asking that question but went the conceptually simpler (to me) route on day one to sprint to getting things working.  I can take a stab at removing these calls and relying on type information if you like.  I believe my hesitation was not knowing what I could expect from casting a concrete object by a generic class name (do you have experience with that in Chapel?)

modules/internal/ArrayViewSlice.chpl
bradcray: The downside of the above, to me, is creation of more global symbols.  Can you articulate what makes the above better for you?  (my guess: avoids relying on statically evaluated method overrides that won't actually dynamically dispatch, so can only be used in a generic context?).

I like the cast approach better, if that works (see my comment above for where my concern that it wouldn't came from -- though I didn't experiment with it to confirm my fears...  haven't had time yet).

modules/internal/ArrayViewSlice.chpl
bradcray: Seems like it.

test/arrays/views/assignReindex.chpl
bradcray: It's telling you it doesn't end with a linefeed.  I'll update it.

test/arrays/views/assignSlice.chpl
bradcray: No strong reason... it was just a simple stepping stone test to get things bootstrapped and up and running.  Once they were, I relied on pre-existing tests for more complete coverage.  Put another way, I wasn't trying to write tests that exercised every possible way of using views, hoping that we had enough coverage in the test system that I wouldn't need to put a lot of effort in here.  (That said, given that there are things I know are wrong that the test system didn't catch, that isn't the best assumption).

test/functions/vass/ref-intent-bug-2big.chpl
```diff
@@ -43,8 +43,8 @@ config const ldx=2, ldy=3;
 type elType = real;
 
 // for neighbor-cache pointers
-var auxArr: [1..1] elType;
-type cacheType = auxArr._value.type; // a class type, so it can be nil
+var auxArr: [1..1, 1..1] elType;
+type cacheType = auxArr[1,..]._value.type; // a class type, so it can be nil
```
bradcray: This test (and the bigger test that it stemmed from) is arguably a bit goofy, and both @benharsh and I were tempted to just retire it, but felt that keeping it running was the path of least resistance.  I believe what's happening here is that the test is saying "I want to store an array object field representing the result of a slice directly, but I don't know what to call that type, so I'll write a little piece of code that will resemble the operation closely enough that I can use Chapel's type query machinery to get its type.

Arguably, no "real" tests should be using _value at all (or we should improve the language), which makes it hard for me to take this test very seriously.

modules/internal/DefaultRectangular.chpl
ronawho: Just following up on this -- remnants of the optimization are removed with #5548

PR #5345, opened 2017-02-10
Only evaluate where-clauses for functions with matching signatures

ronawho:
Previously the compiler would evaluate where-clauses on generic functions too
early. Specifically it would evaluate where-clauses on functions even if their
signatures didn't match, which could lead to all sorts of errors. For example,
this program:

```chapel
proc foo(param x: int(64)) where (x % 2) == 0 { writeln(x, " is even");   }
proc foo(param x: int(64)) where (x % 2) == 1 { writeln(x, " is odd");    }
proc foo(param x: real(64))                   { writeln(x, " is a real"); }

foo(2.2);
```

would lead to errors about `%` not being defined for reals, even though it
never should have looked at the where clause for the int(64) overloads since
the signatures don't match.

This fixes things so that we only evaluate where-clauses on functions with
matching signatures. Basically, we non longer evaluate where-clauses while
we're instantiating generic functions. Instead we evaluate where-clauses for
non-virtual functions while we're filtering candidates (after we've filtered
out ones with non-matching signatures) and we evaluate where-clauses for
virtual functions after made sure the signature matches.


ronawho:
@bradcray this resolves the other futures you added in #5345 (and a related future that you added years ago) and should fix the current regressions in #5277

test/functions/whereClauses/whereClauseParamArgMismatch3.good
```diff
@@ -1,2 +1,2 @@
 whereClauseParamArgMismatch3.chpl:10: error: unresolved call 'foo(1.2)'
-whereClauseParamArgMismatch3.chpl:1: note: candidates are: foo(param x: uint(8))
+whereClauseParamArgMismatch3.chpl:1: note: candidates are: foo(param x: uint(64))
```
ronawho: @bradcray I chose to update the .good to match the .chpl file, but I could instead update the .chpl file to use uint(8) to match whereClauseParamArgMismatch2's signature if you'd prefer

compiler/include/resolution.h
```diff
@@ -248,6 +248,6 @@ bool fixupTupleFunctions(FnSymbol* fn, FnSymbol* newFn, CallExpr* call);
 AggregateType* computeNonRefTuple(Type* t);
 AggregateType* computeTupleWithIntent(IntentTag intent, Type* t);
 
-bool evaluateWhereClause(FnSymbol* fn, bool generic=true);
+bool evaluateWhereClause(FnSymbol* fn);
```
bradcray: Yay!

compiler/resolution/functionResolution.cpp
```diff
@@ -8657,7 +8656,7 @@ addToVirtualMaps(FnSymbol* pfn, AggregateType* ct) {
         }
         if (fn) {
           resolveFormals(fn);
-          if (signature_match(pfn, fn)) {
+          if (signature_match(pfn, fn) && evaluateWhereClause(fn)) {
```
bradcray: Yay!

compiler/resolution/generics.cpp
```diff
@@ -651,23 +651,6 @@ FnSymbol* instantiateSignature(FnSymbol*  fn,
...
-  if (newFn->hasFlag(FLAG_GENERIC) == false &&
-      evaluateWhereClause(newFn)   == false) {
-    //
-    // where clause evaluates to false so cache gVoid as a function
-    //
-    replaceCache(genericsCache, root, (FnSymbol*)gVoid, &all_subs);
-
-    return NULL;
-  }
-
```
bradcray: Yay!

compiler/resolution/generics.cpp
```diff
@@ -697,20 +680,7 @@ bool evaluateWhereClause(FnSymbol* fn, bool generic) {
...
-      // actual-formal matching.  For that reason, conservatively
-      // return 'true' in error cases.  We'll then re-evaluate the
-      // where clause on the concrete instantiation of the generic
-      // function and issue the error (if appropriate) there.
-      //
-      if (generic) {
-        return true;
-      } else {
-        USR_FATAL(fn->where, "invalid where clause");
-      }
```
bradcray: Yay!

test/functions/whereClauses/whereClauseParamArgMismatch3.good
bradcray: I think we could also remove this test.  I think when I was writing these, I was under the impression that non-default integer sizes caused distinct problems or went down distinct code paths, but I've since become convinced that I was high at the time.  I can clean this up after the merge if you don't want to mess further with it.

bradcray:
This is great, and I think is ready to be merged.

test/functions/whereClauses/whereClauseParamArgMismatch3.good
ronawho: I think you're right that it can go down distinct code paths.

For instance 

```chapel
proc foo(param x: int)     where (x % 2) == 1 { writeln(x, " is odd");    }
proc foo(param x: real(64))                   { writeln(x, " is a real"); }

foo(2.2)
```

compiles just fine on master today, but making x an` int(64)` in the first function causes it to break.

ronawho: But anyways, I'm happy to let you clean stuff after the fact if you want, but I have no problem leaving the tests as they are.

ronawho:
@mppf has given this a tentative "ok by me" but he was concerned that this might have some compiler performance regressions because this changes things WRT caching functions. I'm optimistic that it won't result in any regressions, but I'll be sure to check the nightly graphs.

test/functions/whereClauses/whereClauseParamArgMismatch3.good
bradcray: Always good to get evidence that I haven't secretly given up my straight-edge lifestyle! (particularly at work!)

PR #5352, opened 2017-02-13
Added future on passing subclass to superclass ref formal

vasslitvinov:
When a [const] ref formal has the type C, the actual also must
have the type C. When the program passes a D actual, where D
is a subclass of C, the compiler fails verification or generates
invalid C code.

This is based on a bug posted by a user to chapel-users list today.


test/functions/vass/subclass-to-ref-formal.chpl
```diff
@@ -0,0 +1,14 @@
+// Test that for a '[const] ref' formal, the type of the actual
+// must equal the type of the formal.
+// (Unless an implicit conversion comes to play.)
+
+class C    { }
+class D: C { }
+
+proc test(ref lhs: C) {}
```
daviditen: Is lhs supposed to be "const ref" here?  As it is written I think it should compile and execute without error.  Actually, even if it is const I don't see an obvious reason that passing a D shouldn't work.

vasslitvinov: Yeah, it is not intuitive. The reason it shouldn't work is because if we allowed it, the following program would compile yet produce undefined behavior at run time:

    class C    { }
    class D: C { proc print { writeln(); } }

    proc test(ref lhs: C) { lhs = new C(); }

    var d = new D();
    test(d);
    d.print;  // passes static checking, invokes non-existing C.print at runtime

PR #5360, opened 2017-02-15
Add NUMA-aware array CHIP.

gbtitus:
This is a draft CHIP On NUMA-aware array storage, access, and iteration.
It represents the work of a briefly-existing mini-team formed to do a
sanity check on our path forward for arrays with locModel=numa.

### DIFF
--- a/doc/developer/chips/18.rst
+++ b/doc/developer/chips/18.rst
@@ -245,7 +245,7 @@ partitioning of the array's domain among 4 tasks.  We will have poor
 affinity no matter what we do, and thus poor performance.
 
 DSIUtil chunking need not partition on the leftmost dimension of a
-multi-dimension domain.  It actually selects the leftmost dimension
+multi-dimensional domain.  It actually selects the leftmost dimension
 whose size is at least as large as the number of NUMA domains.  Consider
 this array:
 
@@ -396,10 +396,10 @@ This is a start, but note that if either the for-stmt here cannot be
 converted to a forall-stmt or ``complicatedExpensiveComputation()``
 really is as expensive as its name implies then the ``nextLayer``
 references in the loop kernel may not actually be particularly
-NUMA-sensitive, in that getting them perfectly right makes the loop run
-any faster overall.  But for now this is is all we have seen that might
-be sensitive to the cost of array-by-array registration as described
-here.
+NUMA-sensitive, in that the performance will be driven by other things
+and whether the NUMA affinity is right or wrong will not make much
+difference.  But for now this is all we have seen that might be affected
+by the cost of array-by-array registration as described here.
 
 If no such applications exist the performance concerns would be reduced
 but not eliminated, because there would still be the need for dynamic
@@ -525,10 +525,11 @@ optimizations that could be made in the access code.  Nevertheless while
 the multi-ddata access code could be faster than it is now, it will not
 ever be as simple or fast as plain base-plus-offset addressing.  At
 least with all the addressing exposed, getting within something like 3x
-of single-ddata for single-dimension arrays is probably the best we can
-hope for.  For multi-dimension arrays we can probably get closer just
-because multi-dimension addressing is more complicated and multi-ddata
-only adds a fixed cost, but the effect will always be noticeable.
+of single-ddata for single-dimensional arrays is probably the best we
+can hope for.  For multi-dimensional arrays we can probably get closer
+just because multi-dimensional addressing is more complicated and
+multi-ddata only adds a fixed cost, but the effect will always be
+noticeable.
 
 This does not mean that multi-ddata necessarily has bad performance
 overall.  Single-ddata only performs much better when the multi-ddata
--- a/doc/developer/chips/18.rst
+++ b/doc/developer/chips/18.rst
@@ -692,6 +692,17 @@ badly with any envisioned multi-ddata.
 
 ----
 
+.. code-block:: chapel
+
+   var A: [0..7] [0..#2**20] int;
+   forall i in A.domain do forall j in A(i).domain do f(A(i)(j));
+
+Arrays of arrays are a use case we have not put any thought into yet,
+but we will need to do so when thinking longer term about the full
+solution space.
+
+----
+
 *Add examples here.*
 
 
--- a/doc/developer/chips/18.rst
+++ b/doc/developer/chips/18.rst
@@ -4,11 +4,11 @@ NUMA-Aware Array Storage
 Status
   Draft
 
-Author
-  Greg Titus (lead)
-  Ben Harshbarger
-  Elliot Ronaghan
-  Michael Ferguson
+Authors
+  Greg Titus (lead),
+  Ben Harshbarger,
+  Elliot Ronaghan,
+  Michael Ferguson,
   Sung-Eun Choi
 
 Abstract
@@ -26,8 +26,8 @@ particular array elements need to run on CPUs close to the NUMA domain
 on which those array elements are stored.  With ``locModel=numa``, the
 ``DefaultRectangular`` domain has long had NUMA-aware parallel
 iteration, based on ``DSIUtil.chpl:_computeChunkStuff()`` and related
-functionality.  (From here on this will be referred to as *DSIUtil*
-chunking.)  DSIUtil chunking divides the domain as equally as possible
+functionality.  (From here on this will be referred to as *DSIUtil
+chunking*.)  DSIUtil chunking divides the domain as equally as possible
 into subdomains, one for each sublocale, and then creates a task on each
 sublocale to process each subdomain.  What was missing was the ability
 to localize array data in a matching NUMA-sensitive way.  Recent work
@@ -43,19 +43,28 @@ multi-chunk array addressing code is more complicated than that for
 single-chunk addressing because of the need to compute a chunk index
 from the array index.  Thus the dsiAccess() array access method runs
 about 8x slower for multi-ddata arrays than for single-ddata ones.  In
-some cases the extra computation can be in effect hoisted out of kernel
+some cases the extra computation in effect can be hoisted out of kernel
 loops, but not always.  At present any loop which does not use the
 ``DefaultRectangularArr`` serial, parallel standalone, or parallel
 follower iterators performs quite poorly, and really only the parallel
 iterators give good performance.  The serial one is distinctly slower
 than for single-ddata, just not as bad as some other cases.  As an
 example, the following loop written in a natural style performs quite
-badly.
+badly with the current multi-ddata implementation, compared to how it
+does with single-ddata.
 
 .. code-block:: chapel
 
   forall i in A.domain do
     A(i) = i;
+
+But the following fairly similar loop performs quite well, similarly to
+how it does with ``locModel=flat``.
+
+.. code-block:: chapel
+
+  forall (a,i) in zip(A,A.domain) do
+    a = i;
 
 This CHIP documents our mid-course exploration of solutions for this
 problem and some other related ones.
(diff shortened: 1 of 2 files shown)


doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+---------
+
+For architectures with NUMA characteristics, best performance depends on
+affinity between execution and data storage.  The tasks that operate on
+particular array elements need to run on CPUs close to the NUMA domain
+on which those array elements are stored.  With ``locModel=numa``, the
+``DefaultRectangular`` domain has long had NUMA-aware parallel
+iteration, based on ``DSIUtil.chpl:_computeChunkStuff()`` and related
+functionality.  (From here on this will be referred to as *DSIUtil*
+chunking.)  DSIUtil chunking divides the domain as equally as possible
```
sungeunchoi: Totally small nit:  Why isn't chunking italicized?

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+from the array index.  Thus the dsiAccess() array access method runs
+about 8x slower for multi-ddata arrays than for single-ddata ones.  In
+some cases the extra computation can be in effect hoisted out of kernel
+loops, but not always.  At present any loop which does not use the
+``DefaultRectangularArr`` serial, parallel standalone, or parallel
+follower iterators performs quite poorly, and really only the parallel
+iterators give good performance.  The serial one is distinctly slower
+than for single-ddata, just not as bad as some other cases.  As an
+example, the following loop written in a natural style performs quite
+badly.
```
sungeunchoi: Just for reference, I'd add the thing you're implicitly comparing it to.  That is:
```
forall a in A do
  a = ...;
```
And does this faster version typically run as fast as C+OpenMPI?

doc/developer/chips/18.rst
gbtitus: Ah, should be -- will fix.

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+  Some network interface chips (NICs), including Cray's Aries, require
+  that memory be "registered" with them in order to be referenced in
+  network operations.  At least for Aries and probably for some others,
+  this involves pinning virtual addresses to physical addresses to fix
+  their relationship.  Doing so necessarily also sets the pages' NUMA
+  locality, since NUMA locality is a physical characteristic.  If the
+  localization of a registered page is to be changed, then the page must
+  be un-registered first and re-registered afterward.  By itself this
+  may not be very costly, but broadcasting the new registration to the
+  other nodes which need it, and synchronizing that broadcast, may be.
```
sungeunchoi: I think it's worth mentioning libfabric here, because it requires all memory involved in data transfers to be registered.

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+    network cooperation         poor
+    ==========================  ====
+
+
+Description
+...........
+
+This is the simplest model: allocate one ddata for an array and then
+NUMA-localize that memory.  It is what Chapel with ``locModel=numa``
+used to do, and still does for arrays of fewer than 2\*\*20 elements,
```
sungeunchoi: I'd tighten up the wording here a bit.
> It is what Chapel with `localeModel=numa` does for arrays of fewer than ...

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+with all the remote nodes which know about the registration.  In
+particular, the need for synchronization could increase the cost of
+allocation.
+
+All of the single-ddata alternatives have a lower limit on the size of
+array they can be applied to without too much waste.  Since the minimum
+unit of NUMA localization is a page, if it is to be localized an array
+needs to occupy at least as many memory pages as there are NUMA domains
+to avoid wasting space.  This can be large if NIC memory registration
+requires the use of hugepages, as it does on Cray X\* systems.
```
sungeunchoi: To be more precise, you don't have to use hugepages for NIC registered memory on the Cray XC, but the way our comm layers are written requires the use of hugepages.

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+           var idx: int = complicatedExpensiveComputation(v);
+           nextLayer[idx] = someValue;
+       }
+    }
+
+Note that if either the for-stmt cannot be converted to a forall-stmt or
+``complicatedExpensiveComputation()`` really is expensive then the
+``nextLayer`` references in the kernel may not really be NUMA-sensitive.
+But for now this is is all we have seen that might be sensitive to the
+cost of array-by-array registration as described here.
```
sungeunchoi: I don't necessarily find this a compelling example

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+metadata fields used in those computations.  For one thing, computing a
+chunk index from an array index requires doing a divide, effectively.
+This could be sped up significantly by doing an integer multiply by the
+reciprocal instead, or even a right-shift for power-of-two divisors.
+The ddatas and other per-chunk information are themselves currently
+stored as a ddata of records, which could be a tuple of records at a
+fair saving at array creation time.  And the are some other small
+optimizations that could be made in the access code.  Nevertheless while
+the multi-ddata access code could be faster than it is now, it will not
+ever be as simple or fast as plain base-plus-offset addressing.  AT
```
sungeunchoi: AT -> At (right?)

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+least with all the addressing exposed, getting within something like 3x
+of single-ddata is probably the best we can hope for.
```
sungeunchoi: Is true also for multidimensional arrays?  It seems like it should be less.

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+
+This does not mean that multi-ddata necessarily has bad performance
+overall.  Single-ddata only performs much better in when the multi-ddata
```
sungeunchoi: 'much better in when' -> 'much better when'

doc/developer/chips/18.rst
```diff
@@ -0,0 +1,825 @@
...
+there are likely implicit interactions with, and/or subtle dependencies
+on, the current Chapel execution model that runs a program instance per
+compute node.  If so, implementing this idea might be harder than
+expected.
+
+As with "Locale per NUMA Domain" above, this seems likely to produce
+lower performance with arrays too large for a single NUMA domain but
+small enough for a single locale, because of the added involvement of
+the ``Block`` domain map where it is not currently used.
+
```
sungeunchoi: Would it even be possible to have arrays that are too large for a single NUMA domain?  In the case for comm ugni and gasnet/aries where the heap is preallocated, it doesn't seem possible.

sungeunchoi:
Any thoughts on arrays of arrays?  I can't think of how things would be chunked there or the best way to map that to NUMA domains (probably depends on the particular use case).

doc/developer/chips/18.rst
gbtitus: Yeah.  But that might get complicated depending on where (which cache) things were spatially cached in. If they're in an L1 cache at or near the processor that would hide the effect of getting the NUMAness wrong, but if they were in an L3 cache on the NUMA domain it wouldn't.  I'd prefer to avoid that level of detail.  I did do a lot of rephrasing on this part, so see how you like the next version.

doc/developer/chips/18.rst
gbtitus: I don't think it makes much difference, at least as things are coded now.  The chunking is applied after we convert down from the possibly multi-dimensional array domain-based index to a 1-dimensional data storage index.

doc/developer/chips/18.rst
gbtitus: This is specifically talking about single-locale arrays. That might not be clear, though I thought it should be because I talked about arrays smaller than a NUMA domain, those larger than a NUMA domain but smaller than a compute node, and those larger than a compute node (all in the previous section).  In this particular case, consider a single-locale program with an array that takes up 2/3 of compute node memory (and assume 2 NUMA domains).  With any of the single-ddata alternatives, or multi-ddata, that array doesn't need the Block distribution.  You can just declare it without a domain map, and the references go directly through DefaultRectangularArr mechanisms. But with the Distribution-over-Sublocales alternative you would domain map that array using Block.  So all your references would go through the Block mechanisms, which I'd expect to slow things down a fair amount.

gbtitus:
I think for arrays of arrays programmers would need to be able to localize the sub-arrays explicitly, say to put an entire sub-array on a particular sublocale (NUMA domain).  Maybe that's another use case in favor of eventually supporting explicit Contiguous and Chunked domain maps (layouts), though?  I could add an entry to the "Use Cases" section (the one before the Summary) about this if you think it's worthwhile.

doc/developer/chips/18.rst
sungeunchoi: Sounds good.

doc/developer/chips/18.rst
sungeunchoi: But aren't there more multiplies and adds to get from the user's index, say [i,j,k], than that of a one dimensional array?

gbtitus: Oh!  Sorry, misunderstood.  Yes, as the number of dimensions rises the linearization down to an offset in the data storage gets more expensive, and the cost of the chunk computation for multi-ddata is a smaller proportion of that because the chunking is on a specific single dimension and thus is a fixed cost.  That not actually an area we've investigated deeply at all, performance-wise.  The current divide is quite expensive compared even to multiplies.  But if we can implement the divide as a multiply-by-reciprocal then indeed multi-ddata should hurt less the more dimensions there are.  I think in the grand scheme of things this is kind of secondary (one more instance of "but in this somewhat-special case multi-ddata doesn't hurt as much"), but I'll add a sentence about it.

doc/developer/chips/18.rst
sungeunchoi: I was envisioning having this pre-allocated heaps for each numa domain (which is what I meant when I mentioned the comm layers), but I see that your wrote *first-touch* above, so never mind.

doc/developer/chips/18.rst
```diff
@@ -373,11 +392,18 @@ many NUMA-sensitive arrays turned up one that looks like this:
...
-Note that if either the for-stmt cannot be converted to a forall-stmt or
-``complicatedExpensiveComputation()`` really is expensive then the
-``nextLayer`` references in the kernel may not really be NUMA-sensitive.
-But for now this is is all we have seen that might be sensitive to the
-cost of array-by-array registration as described here.
+This is a start, but note that if either the for-stmt here cannot be
+converted to a forall-stmt or ``complicatedExpensiveComputation()``
+really is as expensive as its name implies then the ``nextLayer``
+references in the loop kernel may not actually be particularly
+NUMA-sensitive, in that getting them perfectly right makes the loop run
```
sungeunchoi: Do you mean 'in that getting them perfectly right does not make the loop run any faster overall'?

doc/developer/chips/18.rst
```diff
@@ -373,11 +392,18 @@ many NUMA-sensitive arrays turned up one that looks like this:
...
+any faster overall.  But for now this is is all we have seen that might
```
sungeunchoi: 'is is' -> 'is'

doc/developer/chips/18.rst
```diff
@@ -494,15 +520,18 @@ This could be sped up significantly by doing an integer multiply by the
...
 stored as a ddata of records, which could be a tuple of records at a
-fair saving at array creation time.  And the are some other small
+fair saving at array creation time.  And there are some other small
 optimizations that could be made in the access code.  Nevertheless while
 the multi-ddata access code could be faster than it is now, it will not
-ever be as simple or fast as plain base-plus-offset addressing.  AT
+ever be as simple or fast as plain base-plus-offset addressing.  At
 least with all the addressing exposed, getting within something like 3x
-of single-ddata is probably the best we can hope for.
+of single-ddata for single-dimension arrays is probably the best we can
```
sungeunchoi: 'single-dimension' -> 'single-dimensional' maybe?

doc/developer/chips/18.rst
```diff
@@ -494,15 +520,18 @@ This could be sped up significantly by doing an integer multiply by the
...
+hope for.  For multi-dimension arrays we can probably get closer just
```
sungeunchoi: Maybe multi-dimensional?  I'm not sure.

sungeunchoi:
Re: array of arrays.  Maybe put a sort of place holder in the Use Cases to document that that's another case that might need to be thought about.  I don't know how prevalent arrays of arrays are, but seem like currently not so much, so probably no need to put too much effort into it yet.

doc/developer/chips/18.rst
gbtitus: Hmm, yeah, let me work that over a bit.

doc/developer/chips/18.rst
gbtitus: d'oh!

doc/developer/chips/18.rst
gbtitus: yes, and caught all the "multi-dimension" instances also

PR #5361, opened 2017-02-15
Begin to revise normalize to support non-generic record initializers

noakesmichael:
This is a first non-functional update to a portion the normalization of DefExpr in preparation
for updates to improve the handling of initialization for non-generic records.

1) fix_def_expr() has been renamed normalizeVariableDefinition()

2) Trivial updates to clarify logic flow

Passed with local paratest on linux64


lydia-duncan:
Though the diff is tricky to understand, I think the code is much clearer after your change, and it seems to be maintaining the goal of the original.  I think this is good to go

PR #5362, opened 2017-02-15
Rearrange later portion of resolveNormalCall

lydia-duncan:
Some of my work in the generic initializer version of this function can
be shared between the two versions, for code maintainability purposes.
I created two new helper functions with this intention - this portion of
the code was trickier to rearrange than my last few merges, so I
realized that much of what I had copied is actually not necessary, beyond
the updates to make these functions.

Passed full std/ testing

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -4339,22 +4339,8 @@ FnSymbol* resolveNormalCall(CallExpr* call, bool checkonly) {
   Vec<ResolutionCandidate*> candidates;
   gatherCandidates(candidates, visibleFns, info);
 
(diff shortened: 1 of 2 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -4541,6 +4518,35 @@ FnSymbol* resolveNormalCall(CallExpr* call, bool checkonly) {
   return resolvedFn;
 }
 
+void explainGatherCandidate(Vec<ResolutionCandidate*>& candidates,
+                            CallInfo& info, CallExpr* call) {
+  if ((explainCallLine && explainCallMatch(info.call)) ||
+      call->id == explainCallID)
```
noakesmichael: Inconsistent brace style.

PR #5365, opened 2017-02-16
A second round of updates to the normalization of DefExpr for VarSymbol

noakesmichael:
Continue to straighten out the code to normalize DefExpr for VarSymbol in preparation for
changes to support the initialization of non-generic records.  No changes in behavior.

The overall diff might be hard to follow and so it is broken down in to a series of simple
commits with simple transformations.

Compiled with/without CHPL_DEVELOPER on clang/darwin and gcc/linux.
Passed full paratests

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -1127,6 +1127,11 @@ static void init_typed_var(VarSymbol* var,
                            Expr*      stmt,
                            VarSymbol* constTemp);
 
+static void init_typed_var(VarSymbol* var,
+                           Expr*      type,
+                           Expr*      stmt,
+                           VarSymbol* constTemp);
+
 static void updateVariableAutoDestroy(DefExpr* defExpr);
 
 static void normalizeVariableDefinition(DefExpr* defExpr) {
@@ -1178,6 +1183,9 @@ static void normalizeVariableDefinition(DefExpr* defExpr) {
 
       if (type == NULL) {
         init_untyped_var(var, init, stmt, constTemp);
+
+      } else if (init == NULL) {
+        init_typed_var(var, type,       stmt, constTemp);
 
       } else {
         init_typed_var(var, type, init, stmt, constTemp);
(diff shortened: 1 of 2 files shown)


compiler/passes/normalize.cpp
```diff
@@ -1179,6 +1184,9 @@ static void normalizeVariableDefinition(DefExpr* defExpr) {
       if (type == NULL) {
         init_untyped_var(var, init, stmt, constTemp);
 
+      } else if (init == NULL) {
+        init_typed_var(var, type,       stmt, constTemp);
```
lydia-duncan: I think I get what you're trying to do here, but the insertion of blank space will probably be confusing for a casual reader.  Instead, could you maybe have a small comment describing what the space represents, i.e.

`init_typed_var(var, type, /* init=NULL, */ stmt, constTemp); // skips the init argument`

Otherwise, I think a code reader at a later point will be confused why the space is there and just remove it.

noakesmichael:
Thanks.  In fact that space is gone in my current branch.

Mike


On Feb 16, 2017, at 8:00 AM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


In compiler/passes/normalize.cpp<https://github.com/chapel-lang/chapel/pull/5365#discussion_r101554009>:

> @@ -1179,6 +1184,9 @@ static void normalizeVariableDefinition(DefExpr* defExpr) {
       if (type == NULL) {
         init_untyped_var(var, init, stmt, constTemp);

+      } else if (init == NULL) {
+        init_typed_var(var, type,       stmt, constTemp);


I think I get what you're trying to do here, but the insertion of blank space will probably be confusing for a casual reader. Instead, could you maybe have a small comment describing what the space represents, i.e.

init_typed_var(var, type, /* init=NULL, */ stmt, constTemp); // skips the init argument

Otherwise, I think a code reader at a later point will be confused why the space is there and just remove it.

compiler/passes/normalize.cpp
```diff
@@ -1339,71 +1339,53 @@ static void init_typed_var(VarSymbol* var,
...
-  if (isNoinit && !fUseNoinit) {
-    // In the case where --no-use-noinit is thrown, we want to still use
-    // noinit in the module code (as the correct operation of complexes
-    // depends on it).
-
-    // Lydia note: The requirement for complexes is expected to go away when
-    // we transition to constructors for all types instead of the _defaultOf
-    // function
-    Symbol* moduleSource = var;
+  if (init->isNoInitExpr() == true) {
```
lydia-duncan: If Coverity were still useful, it would probably yell at you for accessing a method on something that could be NULL.  I'm not sure if you change the `hasFlag(FLAG_PARAM)` case above this code to no longer check that init is not NULL, but either way, I think it might be good to assert at the beginning of this function that init is not NULL (with maybe a comment pointing to the other version), in case someone comes along later and mistakenly tries to use this function instead of the one for init being NULL.  Doing that will ensure that your assumptions and preconditions will continue to be met going forward.

compiler/passes/normalize.cpp
```diff
@@ -1448,6 +1415,32 @@ static void init_typed_var(VarSymbol* var,
...
+           moduleSource->defPoint        != NULL) {
+      moduleSource = moduleSource->defPoint->parentSymbol;
+    }
+
+    if (ModuleSymbol* mod = toModuleSymbol(moduleSource)) {
+      if (moduleSource->defPoint != NULL) {
+        retval = mod->modTag == MOD_INTERNAL || mod->modTag == MOD_STANDARD;
+      }
+    }
+  }
```
lydia-duncan: Did you restore the explanatory comments to this function?

lydia-duncan:
I am not sure I follow the assertion for the second to last commit.  How can we be certain that the var cannot be extern in this case?  Is this a consequence of our rules for externs, or is there something which I missed in the immediate code that causes that to be true?  In either case, I think it would be good to add an assertion this is the case, so that we can know going forward that is true.

lydia-duncan:
I don't have any further comments on this branch.  I would like to see some assertions added so that your assumptions are not violated in the future, and it would be nice to have the comments restored that I pointed out.  Other than that, I think this is a good change

PR #5366, opened 2017-02-16
Fix testing problems with a new C atomics test

mppf:
I tried for a few minutes to run the test and link with the Chapel runtime but I couldn't quickly figure out how to do so. Therefore, for now, this test assumes that the relevant functionality is in header files, and it skips CHPL_ATOMICS which is the case where we know it is not.

Additionally, initialize a local variable so that this test runs cleanly under valgrind. (The error valgrind was complaining about is harmless and has to do with a compare-and-swap loop using memory that hasn't been written to yet, which is not necessarily wrong).

Verified test is skipped under CHPL_ATOMICS=locks and that the test passes with intrinsics or cstdlib atomics on my system.

### DIFF
--- a/test/runtime/ferguson/testatomics.test.c
+++ b/test/runtime/ferguson/testatomics.test.c
@@ -74,6 +74,8 @@ int main(int argc, char** argv)
 {
   atomic_bool flag;
 
+  memset(&flag, 0, sizeof(flag));
+
   atomic_store_bool(&flag, false);
   assert( false == atomic_exchange_bool(&flag, true) );
   assert( true == atomic_exchange_bool(&flag, true) );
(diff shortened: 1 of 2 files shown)


test/runtime/ferguson/testatomics.test.c
```diff
@@ -74,6 +74,8 @@ int main(int argc, char** argv)
 {
   atomic_bool flag;
 
+  memset(&flag, 0, sizeof(flag));
```
ronawho: Does atomic_init_bool not work?

mppf: Is it important to you that we test it? This whole business came from moving an old test that wasn't getting run at all... and I'm not that interested in making the test better as much as just getting it out of the include/ directory.

Anyway, I didn't remember there was such a thing when I was looking at the valgrind error.

ronawho: You need to make the test "better" to quiet it for valgrind, and if atomic_init_bool quiets valgrind, I can't imagine why you wouldn't want to use the runtime provided "initializer" over rolling your own with a memset.

Issue #5347, opened 2017-02-11
GASNet LD ignored / Makefiles set LD too much


bradcray:
When linking compiler-generated code, Chapel's Makefiles set LD too many times.  For example:

    runtime/etc/Makefile.comm-gasnet:LD = $(GASNET_LD)
    runtime/etc/Makefile.include:LD = $(CC)
    runtime/etc/Makefile.launcher:LD = $(CC)
    runtime/etc/Makefile.regexp-re2:LD=$(CXX)

As an impact of this, if a GASNet configuration requires using the linker that GASNet determined at its configuration time, we override it and use some other linker instead.

This was originally reported by Jintack Lin in July 2016 (see https://sourceforge.net/p/chapel/mailman/message/35219550/ ) and has since been reported by other users who report that GASNet wants to use mpicc as the linker, but then by simply using gcc/g++, we lose crucial libraries that mpicc would normally bring in like -lmpi.  

I started on a fix for this last summer in this branch https://github.com/bradcray/chapel/tree/too-many-LDs but did not get to a point where I felt confident it was working well.  Specifically, I was worried about how to resolve the contradiction between the regexp Makefile thinking it wants to use the C++ compiler to link vs. GASNet possibly wanting something else.  My intuition is that by link time, we should just be stitching object files together without worrying about whether they're from C or C++ code, but @mppf was concerned that the linker wouldn't link in required aspects of the C++ standard library if this was done.  Then I timed out before getting more confident in my solution.

mppf:
Some possible approaches:
 * adjust LD to use C++ version instead of overriding it (e.g. mpicc -> mpi++)
 * link with -lstdc++ when RE2 is used as described [on this stack overflow issue](http://stackoverflow.com/questions/1001535/linking-c-code-with-gcc-without-g ). I'm not sure how standard -lstdc++ is - is it just an GCC thing?

PR #5369, opened 2017-02-16
Unify how LD is set

bradcray:
It has been observed by a few users that the LD setting which GASNet wants us to use is not always (ever?) what we actually end up using.  For example, if GASNet wants to use mpicc as the linker but we fail to use it, we may not get the necessary mpi libraries linked into our binaries.  See issue #5347 which was recently filed retroactively against a user issue last summer after a second user ran into similar problems this past month.  Prior to this change, LD was set in a number of places in our Makefiles in a reasonably scattershot manner.  GASNet wanted one thing, re2 wanted another, etc.

In general, it's easy to see how different components of Chapel may want different linkers, and it's tough to see, in general, how to arbitrate between them.  At present, though, it seems like GASNet has a stronger and more important opinion about linkers than most other packages we use.  That said, it sometimes chooses a C-based linker, while re2 wants C++ to be linked in (and other user-provided code files may as well).  

Q: Why does the link step care whether it's using a C or C++ linker?
A: Because of the need to link in standard libraries for that language.

To that end, I've taken the following approach in this PR:

* For the main generated binary / _real executable, LD is now always set in runtime/etc/Makefile.comm-*:
  - for CHPL_COMM=none and ugni, we set it to CXX.
  - for CHPL_COMM=gasnet...
    - on Crays, where GASNet hard-codes the path more than we'd like, we set it to CXX as we have
    - when the host and target platforms don't match, we set it to CXX (I'm not entirely sure why, I just followed the historical pattern in this case)
    - otherwise, for typical gasnet cases:
      - if GASNET_LD is GASNET_CXX, we use it
      - if GASNET_LD is GASNET_CC, we use GASNET_CXX
      - if GASNET_LD is MPI_CC we use mpicxx (this seems to be the most portable C++ wrapper name across MPI implementations)
      - otherwise, we print a warning (because the GASNet team says it should always be one of the three cases above at present) and set it to CXX
* For the launcher, LD is set in Makefile.launcher and is set to CC on the argument that launcher code shouldn't require anything exotic.

I've tested this with re2 and gasnet enabled/disabled on:
* darwin using clang
* Crays using cray-prgenv-cray and cray-prgenv-intel
* linux64 using gnu
It could still blow up in other configurations, but I'm hoping to lean on smoke tests and nightlies to help with that unless anyone has additional favorite configurations they'd like me to spot-check.

TODO:
- [x] re-test on Crays after latest Makefile change


mppf:
Wondering if this works with
 * GASNet
 * not on a Cray
 * RE2 enabled

it seems that would lead to missing C++ symbols but I havn't tried it.

bradcray:
@mppf: Yeah, that's the main mode I've been testing -- regexp-requiring test with and without GASNet enabled.  From what I've seen, my impression is that GASNet also likes to choose C++-enabled linkers.  The main thing I'm uncertain about in this PR is that GASNet will always choose a C++-enabled linker in every setting.  But given that things are currently known to be broken for users and working for the test configurations I've looked at, I think this is the right step forward until we find counterexamples that break.

mppf:
Seems like a good question for Paul.  Anyway, if that's how you've been testing it - maybe update your testing paragraph to say "with and without GASNet but with RE2 enabled" e.g.

Thanks for fixing it. You can consider it reviewed by me.

bradcray:
So I "made the mistake" of checking with Paul and learned that I was just getting lucky by looking at GASNet configurations that use C++ as the linker.  Unfortunately, there's currently no good way of saying "hey gasnet, will you hand me a linker variable that's the C++ variant of what you wanted to set to?" but we did come up with a Makefile conditional that should cover all the currently possible cases so I'll amend my original solution to that more complex logic.

bradcray:
@mppf: This last commit was big enough that it probably deserves to be re-reviewed, if you wouldn't mind.

mppf:
Looks good, thanks.

PR #5370, opened 2017-02-16
Enable const/const ref this intent

mppf:
This PR adjusts the parser to support `const` and `const ref` this intents. This helps with PR #5134.

For example, the `ref` below is a `this` intent:

``` chapel
record R {
  proc ref f() { }
}
```

After this PR, the following `const` and `const ref` this intents will
also be valid:

``` chapel
record R {
  proc const g() { }
  proc const ref h() { }
}
```

Passed full local testing.


lydia-duncan:
I don't know enough about intents to have an opinion, but I ran afoul of refs and "this" during functionResolution.  You may want to verify that the output you see after that pass is what you would expect

mppf:
@lydia-duncan - PR #5134 includes tests that they are getting parsed & set up right. But other things in that PR are necessary for it to be reasonable to work with them.

bradcray:
I'm generally in favor of this change.

Q: Should 'const in' and/or 'in' be supported as options for 'this intents' as well?

It seems to me that the spec changes should include more than just the syntax diagram update -- i.e., narrative describing what the 'const ref' and 'const' intents mean in this position.  Taking a quick look, it seems like these might fit in the paragraph that starts "When this-intent is ref..." if we didn't want to rewrite the this intent description more significantly.

(Related: Not your problem per se, but it seems unfortunate that the introduction of 'this-intents' takes place in the Classes chapter given that we permit methods to be defined for arbitrary types.  Might be worth putting in spec/TODO (if it isn't already) or creating an issue for in order to not lose this note).

One other thing that occurs to me on this topic is that the original 'this-intents' use cases (and spec text, arguably) were done with the mindset that each method will only have one signature/overload, and therefore one 'this intent.'  But it seems as though some methods could be overloaded based on their this intents (e.g., param and type?) while others might lead to ambiguities/problems (e.g., an overload on 'const' vs. 'const ref' might be ambiguous; as might an overload on 'ref' vs 'default' in some cases; or 'ref' and 'in'?).

Maybe putting that last paragraph another way:  I think when we introduced the first 'this intent', we thought of it as requiring a particular characteristic of the receiver it corresponded to (e.g., "it must be a param value").  By contrast, the newer 'this intents' feel more like "here's how to pass the receiver into this method if it's the one you're invoking."  I don't think that's inconsistent with the original definition, but it seems like a different way of viewing/describing it.

These general observations (e.g., this intents are becoming more like argument intents for an implicit argument, they can be applied to non-classes, maybe all argument intents make sense as this intents) are what makes me think that the ideal spec update around this intents should perhaps take a new cut at describing them.

mppf:
@bradcray - I've tried to make some minor improvements. Do you think what I have here is acceptable for now?

I think that the entire "Methods" section needs to be pulled out of the "Classes" section and substantially re-worked. However, I'm having trouble believing that is what I personally need to do next...

Re. supporting `const in` or `in` for `this` intents - I don't have any conceptual reason why that should not be supported - I just don't currently have a use case I care about for it and so am hesitant to open what might be a can of worms in the implementation.

spec/Classes.tex
```diff
@@ -915,8 +915,13 @@ \subsection{The Method Receiver and the {\em this} Argument}
 
 When \sntx{this-intent} is \chpl{ref}, the receiver argument will be
 passed by reference, allowing modifications to \chpl{this}.  If
-no \sntx{this-intent} is specified, the receiver will be passed with
-the default intent as specified in \rsec{The_Default_Intent}.
+\sntx{this-intent} is \chpl{const ref}, the receiver argument is passed
+by reference but it cannot be modified inside the method. The
+\sntx{this-intent} can also describe an abstract intent as follows. If it is
+\chpl{const}, it the receiver argument will be passed with \chpl{const}
```
bradcray: I think there's a typo here in the second use of "it"

test/parsing/ferguson/this-intent.chpl
```diff
@@ -0,0 +1,22 @@
...
+  }
+  proc ref g() {
+    writeln(x);
+  }
+  proc const h() {
+    writeln(x);
+  }
+  proc const ref i() {
+    writeln(x);
+  }
```
bradcray: It appears as though there's nothing about this test that actually checks that the argument is sent in by 'ref' (e.g., can be modified and is an alias) or 'const ref' (e.g., can't be modified / isn't an alias).  I.e., it essentially seems to be a test that the parser accepts the keywords.  Could it be beefed up to actually test that the implementation is correct as well?

bradcray:
Thanks for adding the description to the spec which was the main thing I thought was missing in the previous PR.

> Re. supporting const in or in for this intents - I don't have any conceptual reason why that should not be supported - I just don't currently have a use case I care about for it and so am hesitant to open what might be a can of worms in the implementation.

It's fine with me to put this off.  The thing that caused me to mention it was that your email introducing this PR seemed to suggest "Huh, it looks like we inadvertently missed these cases" which made me think that you were trying to flesh out any remaining missing intents.  If you felt so inclined and agreed that 'in' and 'const in' were appropriate and missing, it might be nice to put some breadcrumbs in place in the form of comments in the code and/or spec, and/or an issue+future indicating that we believe that it ought to be done (and maybe mark it as 'easy' and point to this PR as a reference as to what parts of the code base would likely need to change?).

> I think that the entire "Methods" section needs to be pulled out of the "Classes" section and substantially re-worked. However, I'm having trouble believing that is what I personally need to do next...

I agree, which is why I suggested updating the TODO.  Thanks for doing that.

> Do you think what I have here is acceptable for now?

With the typo fix I'd consider it acceptable.  If the new test made sure that the intents were doing what they were supposed to, I'd be much more happy with it.  If breadcrumbs were filed against the remaining intents, I'd be very content.

test/parsing/ferguson/this-intent.chpl
mppf: No, because we don't check const-ness for such things right yet. This PR started as taking some separable things out of PR #5134. But the parent PR should enable the const-ness checking. I'll add a test there.

bradcray: Could the 'ref' case be checked to see if it does get a ref?  (is modifiable and the mod is reflected at the callsite?).  Would this be an appropriate time to add a future for the const checked case?

mppf: I take that back - it does work with 'const ref' but not 'const'. I'll file a test and a .future in this PR about it and make a note to myself to try to fix the 'const' case in the umbrella PR.

PR #5374, opened 2017-02-16
Fix a bad standalone iter in a vectorization test

ronawho:
Fix some invalid/bad syntax in the standalone iter for nonInlinableFollower
test. On master, our use of the tryToken to pick between standalone and
leader/follower hid this, but Vass discovered the bug in his forall-stmt work.

Correct the standalone iter, and update the test output and some comments to
accurately reflect that the for-loop in the standalone iter *should* be marked
as order-independent.


ronawho:
@vasslitvinov, quick review/sanity check?

vasslitvinov:
Looks good - thank you!

PR #5377, opened 2017-02-17
Declare ISx's arrays over an anonymous range instead of a remote domain

benharsh:
Declaring an array over a remote domain results in an on-statement onto
that domain's locale in order to increment the reference count and
manipulate the list of arrays declared over the domain. This can result
in a potentially large number of on-statements on Locale0, which can
hurt performance and hold up the exchange's barrier. Slicing a remote
domain also introduces some communication.

Long-term, we should try to optimize this pattern as it surely exists in other benchmarks and is easy to fall into.


benharsh:
@chapel-lang/perf-team : this is a relatively minor change that helps ISx performance. Some may find it less elegant, but I think it looks fine. Let me know if you object.

lydia-duncan:
While I do think that is less elegant, I think it is fine.  Have you opened an issue to track it?

benharsh:
Just opened #5378 , thanks for the reminder.

bradcray:
I'd probably go one step further and remove the named LocTaskSpace domain altogether for now.  It seems odd to introduce it only to distribute it on the next line.

I think it'd also be nice to insert a future with this PR that's linked to your issue that shows how this idiom (local domain on locale 0 used to declare arrays on locale 1 causes communication that we'd like to avoid.  While I can imagine variations on this theme that would be more difficult to address, it seems like this case should be fairly easy to optimize with some compiler / param support:

* const domain => there's no need to keep a list of arrays that we'll need to update when it grows/shrinks because it can't.
* module-scope domain => there's no need to keep a list of arrays / reference count to know when all its arrays are gone and it should be freed up.  The only time it goes out of scope is when the program is ending and it can either be freed or leaked at that point regardless of its arrays' status.

(scanning your issue quickly, I see that you're thinking along similar lines...)

benharsh:
How do things look now?

bradcray:
Looks great to me, thanks!

PR #5380, opened 2017-02-17
CHAPEL-295 MPI testing : add mpicc correctness test

awallace-cray:


util/cron/test-mpicc.bash
```diff
@@ -0,0 +1,26 @@
...
+module list -l 2>&1 | grep -E '\bmpi/mpich\b' || exit $?
+
+export MPICH_MAX_THREAD_SAFETY=multiple
+
+export CHPL_TARGET_COMPILER=mpi-gnu
+export CHPL_TASKS=fifo
+export CHPL_COMM=none
+export CHPL_LAUNCHER=mpirun
+
+export CHPL_NIGHTLY_TEST_DIRS="--no-recurse release/examples modules/packages/mpi/spmd/hello-chapel"
```
ben-albrecht: I believe we want to cover `modules/packages/mpi/spmd` *with recurse*. Are we able to get both?

util/cron/common.bash
```diff
@@ -110,18 +110,3 @@ fi
 if [ -z "$CHPL_TEST_COMP_PERF_DIR" ]; then
     export CHPL_TEST_COMP_PERF_DIR=$logdir_prefix/NightlyPerformance/$(hostname -s)
 fi
-
```
ben-albrecht: Is this removal related to this PR?

util/cron/test-mpicc.bash
```diff
@@ -0,0 +1,26 @@
...
+# Test CHPL_COMM=none && CHPL_TARGET_COMPILER=mpi-gnu for MPI module testing
+#
+CWD=$(cd $(dirname $0) ; pwd)
+source $CWD/common.bash
+
+echo >&2 module load mpi
+module load mpi
+
+set -x
+: confirm mpi module is loaded
```
ben-albrecht: Is this intended to be a comment, or a clever way to echo?

awallace-cray: ":" is a shell command that returns "true" and does nothing else
With set -x, ":" shows up on stderr in the same format as all the other shell commands echo'd  when set -x. 
(we want set -x from here on to see the shell commands - but not before here, because source $CWD/common.bash and any module load commands are ugly.)

util/cron/test-mpicc.bash
awallace-cray: Not in one pass of "nightly". At least, I don't know how to get --no-recurse for release/examples and (recurse) for anything else, all in the same pass.

util/cron/test-mpicc.bash
ben-albrecht: OK, works for me.

util/cron/common.bash
awallace-cray: Loosely related. While testing this (and other recent changes) on SLES 12 server, I found out that Cray I.T. has changed their recommended modules setup. I hope to gradually squeeze the module setups out of the Chapel test scripts.  Not the module loads; only "source <some/path>/modules/bash", "module use <some/path>", etc, which are host-specific.  I want those to come from some host-specific script maintained outside of Chapel. (.bashrc for example)

util/cron/test-mpicc.bash
ben-albrecht: Hm, bummer. Perhaps we could drop the `--no-recurse` and try:

`export CHPL_NIGHTLY_TEST_DIRS="release/examples/hello*.chpl modules/packages/mpi/spmd"`

PR #5382, opened 2017-02-17
default to --spmd 1 for mpirun launcher

ben-albrecht:
This PR changes gives the `--spmd` flag a default argument of `1` for `CHPL_LAUNCHER=mpirun`, enabling testing, without adding `execopts` files or special treatment of the `mpirun` launcher in `sub_test`.

- [X] Tested on linux64

### DIFF
--- a/runtime/src/launch/mpirun/launch-mpirun.c
+++ b/runtime/src/launch/mpirun/launch-mpirun.c
@@ -26,7 +26,7 @@
 
 #define CHPL_SPMD "--spmd"
 
-static char* mpi_num_ranks=NULL;
+static char* mpi_num_ranks="1";
 
 static char _nlbuf[16];
 static char** chpl_launch_create_argv(const char *launch_cmd,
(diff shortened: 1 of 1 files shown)


runtime/src/launch/mpirun/launch-mpirun.c
```diff
@@ -38,7 +38,7 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
   int numranks;
 
   // TODO When MPI Comm layer is supported, support -nl>1 && CHPL_COMM!=none
-  if (numLocales != 1 || mpi_num_ranks==NULL) {
+  if (numLocales != 1) {
     chpl_error("For mpirun, specify number of SPMD images via --spmd <#>", 0, 0);
   }
```
awallace-cray: This makes the launcher stop with the given message and an error, if nl > 1... is it clear enough to the user that mpirun's problem is nl > 1 ?

PR #5393, opened 2017-02-21
Fix minor logic error in new business logic to track generic AggregateTypes

noakesmichael:
Lydia noted that the new code I introduced to track generic class/record had a trivial
and embarrassing error.  The code path in question is not currently being stressed on
master and so didn't surface as a failure in behavior.

Compiled on clang/darwin
Tested on darwin for release/examples.

### DIFF
--- a/compiler/AST/type.cpp
+++ b/compiler/AST/type.cpp
@@ -633,7 +633,8 @@ void AggregateType::addDeclaration(DefExpr* defExpr) {
     } else if (var->hasFlag(FLAG_PARAM) == true) {
       mIsGeneric = true;
 
-    } else if (defExpr->exprType == NULL) {
+    } else if (defExpr->exprType == NULL &&
+               defExpr->init     == NULL) {
       mIsGeneric = true;
     }
 


bradcray:
Would it be possible/easy to add a test whose presence would've caught this mistake outright?

noakesmichael:
I gave that a few moments thought about that when Lydia pointed
out my oversight.




I am not as creative as some about such matters, I wouldn’t be a good
black-hat hacker, but I think the answer is “not right now but plausibly
in the future”.


This property isn’t being used to drive any business logic on master and
so there nothing to test against esp. at the Chapel level.



Even on my branch it’s currently only being used in a trivial way in one
place so that I can avoid triggering my new code along an exception path.
If one knows the details then one could write a test that determines whether
certain error cases continue to generate the existing error message or
generate a more laborious resolution error.


I expect to lean on it a bit more going forward esp. to reduce the likelihood
that I unintentionally kick Lydia in the shins while working on non-generic
records.   So then it would be easier to tell if it’s working reliably.


In the longer term this property might be used to simplify existing logic for
catching and reporting user level coding errors. If that were to happen then
it would be easier to confirm that it is being set appropriately.

Mike





On Feb 21, 2017, at 3:52 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


Would it be possible/easy to add a test whose presence would've caught this mistake outright?

bradcray:
OK, np.  I wasn't sure whether "not exercised on master" meant "missing an obvious test" or "the compiler simply doesn't exercise this code (much/easily)."  Sounds like it's the latter, which is fine with me.

PR #5395, opened 2017-02-22
Update logic for pruning unused module level variables

noakesmichael:
Work on normalization lead to a regression for 1 test due to the ad-hoc nature of some code at
the end of function-resolution that prunes unused module level variables.  This code has always
missed some cases and, on my branch, misses an important case that then leads to an INT_ASSERT.

Ripped out the ad-hoc logic and replaced it with a simple liveness test that uses the new-ish
infrastructure that Michael/Ben introduced for symbols.

On a simple test, akin to hello world, this pruned 22 module variables that were not being
pruned.  One of these was the final module level variable in CommDiagnostics.  This implies
that this PR removes 1 additional module for tests that don't use comm diagnostics.  Updated
two tests accordingly.

Passed paratest on linux64 for single-locale and gasnet

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -296,7 +296,7 @@ static void removeUnusedFunctions();
 static void removeUnusedTypes();
 static void buildRuntimeTypeInitFns();
 static void buildRuntimeTypeInitFn(FnSymbol* fn, Type* runtimeType);
-static void removeUnusedGlobals();
+static void removeUnusedModuleVariables();
 static void removeParamArgs();
 static void removeRandomPrimitives();
 static void removeActualNames();
@@ -9917,7 +9917,7 @@ pruneResolvedTree() {
   replaceTypeArgsWithFormalTypeTemps();
   removeParamArgs();
 
-  removeUnusedGlobals();
+  removeUnusedModuleVariables();
   removeUnusedTypes();
   removeActualNames();
   removeFormalTypeAndInitBlocks();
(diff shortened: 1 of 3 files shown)


ronawho:
👍  Looks great to me.

I haven't used the the new def/use API myself, so it might be worth getting a sanity check from @mppf or @benharsh

benharsh:
Looks good to me!

compiler/resolution/functionResolution.cpp
```diff
@@ -10049,17 +10049,19 @@ static void removeUnusedTypes() {
...
-            (def->sym->type->symbol->hasFlag(FLAG_GENERIC) &&
-             def->sym->hasFlag(FLAG_TYPE_VARIABLE)))
-          def->remove();
+// Remove module level variables if they are not defined or used
+// With the exception of variables that are defined in the rootModule
+static void removeUnusedModuleVariables() {
+  forv_Vec(DefExpr, def, gDefExprs) {
+    if (VarSymbol* var = toVarSymbol(def->sym)) {
+      if (ModuleSymbol* module = toModuleSymbol(def->parentSymbol)) {
+        if (var->isDefined() == false && var->isUsed() == false) {
```
mppf: I think this is a reasonable way to write it, but knowing what isDefined and isUsed actually do - you're actually just checking if the Symbol is referred to by any SymExprs that are in the tree. That's even simpler than isDefined/isUsed (which check for SymExprs in specific contexts - uses or defs). So it would seem that we could have a Symbol method that just returns true if for_SymbolSymExprs iterates over anything at all. The main problem I see there is that it's hard to come up with a good name for it.

Either way this is a minor issue & I'm glad that this code is simplified and more correct.

PR #5396, opened 2017-02-22
Update logic to normalize a subset of variable declarations

noakesmichael:
Expand the logic that normalizes variables with default initialization e..g

  var x : int;
  var  y : myRecord;

to simplify the AST emitted by normalize with a focus to avoiding
copying during the initialization of non-generic records.

Passed single-locale paratest on linux64


lydia-duncan:
This looks fine to me

PR #5398, opened 2017-02-22
Fix a copy propagation bug related to array alloc

ronawho:
Prevent propagating through an array alloc primitive. This fixes a bug
@gbtitus ran into while trying to edit our ddata_allocate function.


benharsh:
👍 👍

Issue #5357, opened 2017-02-14
Consider removing the inlineSingleYieldIterator() optimization and supporting code


ronawho:
`inlineSingleYieldIterator()` was originally added af0e77e1, a commit that added initial/prototype GPU support. 

Most of that code was later removed in 60c2580, though inlineSingleYieldIterator remained. I'd like to remove this code because I don't believe it's well understood by anybody currently on the team, and it doesn't seem like it has much value. I'd imagine it could have been useful for the initial GPU work, but I don't think it provides much functionality today.

As far as I can tell, the only thing it does is inline zippered iterators that yield a single value. It does not inline iterators that have loops, or ones that have more than one yield, so in effect it only inlines iterators that yield one, and only one value, and I don't think there's any compelling cases where we do that. i.e. it allows:

```chapel
iter myIter() { yield 1; }
for (i, j) in zip(myIter(), myIter()) do writeln(i, " ",  j);
```

to be generated as

```c
writeln_chpl2(INT64(1), INT64(1));
```

and without this optimization we generate some ugly code: 

```c
  _ic__chpl = chpl_here_alloc(sizeof(chpl__ic_myIter_chpl_object), INT16(31));
  (_ic__chpl)->more_chpl = INT64(1);
  _ic__chpl2 = chpl_here_alloc(sizeof(chpl__ic_myIter_chpl_object), INT16(31));
  (_ic__chpl2)->more_chpl = INT64(1);
  advance_chpl(_ic__chpl);
  advance_chpl(_ic__chpl2);
  for (; (tmp_chpl = (_ic__chpl)->more_chpl,tmp_chpl); ) {
    writeln_chpl2((_ic__chpl)->value_chpl, (_ic__chpl2)->value_chpl);
    advance_chpl(_ic__chpl);
    advance_chpl(_ic__chpl2);
  }
  chpl_here_free(_ic__chpl);
  chpl_here_free(_ic__chpl2);
```

While the optimization clearly allows that code to be generated much more elegantly/cheaply, this code pattern (zippered iterating over a single yield) is almost never used. There are only 3 tests that run into it: functions/iterators/bradc/zipIteratorsRef.chpl, statements/vass/index-variable-accesses.chpl, and studies/glob/test_glob.chpl. 

The only potentially interesting case that comes to mind is that it allows zippered followers that only yield a single value to be optimized, and that seems like a far more interesting case than zippered serial iterators, but I'm not sure it's worth keeping the extra code around for.

[...]

PR #5404, opened 2017-02-23
Remove the single-yield iterator inlining "optimization"

ronawho:
The inlineSingleYieldIterator optimization was originally added as part of some
prototype GPU code with af0e77e. Most of the GPU code was later removed in
60c2580, but inlineSingleYieldIterator remained.

The only thing this optimization does is permit zippered inlining of iters that
only yield a single value. i.e. iterators that do *not* have any loops and only
have a single top-level yield like `iter myIter() { yield 1; }`

This is not a common iterator idiom, and certainly not a performance critical
one. In fact there are only 3 tests that run into this, and none of them are
compelling.

This code isn't well understood by anybody on the team, and doesn't do anything
for performance, so it's being removed to simply iterator lowering.

This closes https://github.com/chapel-lang/chapel/issues/5357, which has some more info as well as examples.

### DIFF
--- a/compiler/resolution/lowerIterators.cpp
+++ b/compiler/resolution/lowerIterators.cpp
@@ -1411,69 +1411,6 @@ canInlineIterator(FnSymbol* iterator) {
   return (count == 1) ? true : false;
 }
 
-
-static void
-getRecursiveIterators(Symbol* iter, std::vector<Symbol*>& iterators) {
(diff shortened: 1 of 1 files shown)


ronawho:
@daviditen mind giving this a quick review?

daviditen:
Looks good.

PR #5406, opened 2017-02-23
Remove leak workaround in formalRequiresTemp

mppf:
This workaround interfered with other work I'm doing (to do with cleaning up const checking). Although I might use a different approach for the other work, I believe this workaround can be removed.

See commits 4e918181b9 and  2cd1856f5e4 for information about its origin.

I ran memory leaks testing with test/release/examples and observed no change in the number of tests with leaks (compared to nightly testing) and the same order of magnitude of leaked memory.

Passed full local testing.
Passed test/release/examples with --no-local.


mppf:
@bradcray - would you mind taking a quick look at this? Thanks.

bradcray:
Sweeeet!  Thanks!

PR #5411, opened 2017-02-23
Upgrade from hwloc 1.11.5 to 1.11.6

ronawho:
This release includes minor bug fixes and includes support for AMD's Zen
processors. It also speeds up topology discovery, which is what dominates our
startup time:
 - "Make the Linux discovery about twice faster, especially on the CPU side,
   by trying to avoid sysfs file accesses as much as possible."

On chapcs startup times go from .040 seconds to .032 seconds.

For a full changelog see:
  https://mail-archive.com/hwloc-announce@lists.open-mpi.org/msg00095.html


ronawho:
@gbtitus not sure there's actually anything to review since it's all third-party changes. Just giving you the traditional "hey, I'm upgrading a 3rd party package" ping.

gbtitus:
The Makefile update looks right to me.  :-)   👍

PR #5413, opened 2017-02-23
Date time module

daviditen:
Add a DateTime module to the standard modules, ported from the python datetime module.

There are four main records involved:
* date - a date that assumes the current Gregorian calendar has always applied
* time - a time of day
* datetime - a combined date and time
* timedelta - an amount of time

Operators on these types are supported such as adding or subtracting a datetime and timedelta, comparing (>, >=, <, <=), two datetimes, etc.

There is also an abstract base class 'TZInfo' for representing timezones.  Classes derived from and implementing this class can be used with time and datetime objects to represent times including time zone.

There are high level chpldoc comments, but it defers to the python documentation for more detail.

There is also a port of much of a large python unit test, currently passing.  These tests are not a complete port because most of the error cases are left out.  Python's errors are exceptions, so they can be caught and the test can continue.  We'd currently have to have a different test for each error condition, so I skipped those.


modules/standard/DateTime.chpl
```diff
@@ -0,0 +1,1571 @@
...
+module DateTime {
+  /* The minimum year allowed in date objects */
+  param MINYEAR = 1;
+  /* The maximum year allowed in date objects */
+  param MAXYEAR = 9999;
+
+  private const DAYS_IN_MONTH = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+  private const DAYS_BEFORE_MONTH = init_days_before_month();
+
+  /* The Unix Epoch date and time */
```
daviditen: I think this is a useful enough constant that it's worth exposing.

modules/standard/DateTime.chpl
```diff
@@ -0,0 +1,1571 @@
...
+  }
+
+  /* Methods on time values */
+
+  /* Replace the hour, minute, second, microsecond and tzinfo in a time
+     to create a new time. The tzinfo argument is required, the rest are
+     optional.
+   */
+  proc time.replace(hour=-1, minute=-1, second=-1, microsecond=-1,
+                    tzinfo:TZInfo) {
```
daviditen: Of course! I initially set it to `nil`, but then realized that there had to be a way to specify that you want to clear the `tzinfo`.  I didn't think of setting the default to `this.tzinfo`, so made it a required argument instead.

modules/standard/DateTime.chpl
```diff
@@ -0,0 +1,1571 @@
...
+        (t1.tzinfo == nil && t2.tzinfo != nil) {
+      halt("both datetimes must both be either naive or aware");
+    } else if t1.tzinfo == t2.tzinfo {
+      const sec1 = t1.hour*3600 + t1.minute*60 +
+                   t1.second + t1.microsecond/1000000.0;
+      const sec2 = t2.hour*3600 + t2.minute*60 +
+                   t2.second + t2.microsecond/1000000.0;
+      return sec1 < sec2;
+    } else {
+      // As far as I can tell, python's datetime.time() comparisons don't
```
daviditen: In this comment, I was looking at the constructor named `time`, not the getter named `time`.  It's a little confusing because so many things have the same name in the python module.  The module is named `datetime`, there's a `datetime` type, and a `datetime` constructor, then there's the `time` type, `time` constructor and `time` getter function.

modules/standard/DateTime.chpl
```diff
@@ -0,0 +1,1571 @@
...
+    const newMonth = if month > 0 then month else this.month;
+    const newDay = if day > 0 then day else this.day;
+    return new date(newYear, newMonth, newDay);
+  }
+
+  /* Return a filled record matching the C "struct tm" type for the given date */
+  proc date.timetuple() {
+    var timeStruct: tm;
+
+    // no 0 init for extern records?
```
daviditen: I think I may have been mistaken.  I tried a standalone test and it appears to zero the fields.  I think I'll just remove that comment.

modules/standard/DateTime.chpl
ben-albrecht: This seems like it could be a bug in python's implementation, unless we're both misunderstanding [the documentation for time comparisons](https://docs.python.org/2.7/library/datetime.html#datetime.time.tzinfo):

> comparison of time to time, where a is considered less than b when a precedes b in time. If one comparand is naive and the other is aware, TypeError is raised. If both comparands are aware, and have the same tzinfo attribute, the common tzinfo attribute is ignored and the base times are compared. **If both comparands are aware and have different tzinfo attributes, the comparands are first adjusted by subtracting their UTC offsets (obtained from self.utcoffset()).** In order to stop mixed-type comparisons from falling back to the default comparison by object address, when a time object is compared to an object of a different type, TypeError is raised unless the comparison is == or !=. The latter cases return False or True, respectively.

PR #5418, opened 2017-02-24
For locModel=numa, always use libnuma.

gbtitus:
The primary change here is to change third-party/hwloc/Makefile to
disable libnuma in hwloc configuration for the flat locale model, and
enable it for all others.  This allows hwloc to be used to set and query
NUMA memory locality, among other things.  Since we now configure hwloc
differently for different locale models, we also make a change here to
include the locale model in the hwloc unique config path.

This in turn allows simplifying make/compiler/Makefile.cray-prgenv,
which essentially made the identical decision for hwloc configurations
but only with Cray PrgEnv target compilers.

Finally, there's a similar adjustment to util/start_test, where it
decides whether having been given -performance should cause it to throw
--static to the compiler.  It used to do that for non-flat locale models
only if a Cray PrgEnv target compiler was being used.  Now the target
compiler doesn't matter.

### DIFF
--- a/make/compiler/Makefile.cray-prgenv
+++ b/make/compiler/Makefile.cray-prgenv
@@ -54,19 +54,15 @@ GEN_DYNAMIC_FLAG = -dynamic
 LIB_STATIC_FLAG =
 LIB_DYNAMIC_FLAG = -shared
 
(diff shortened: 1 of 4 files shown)


gbtitus:
@ronawho , @dmk42 , care to review?

PR #5419, opened 2017-02-24
Make Math.conjg more robust

ben-albrecht:
Due to coercion rules, `Math.conjg` (complex conjugate) accepts any numeric type and always returns a `complex(64)` or `complex(128)` type. This means programs that support many numeric types and use `conjg` must often cast the result of `conjg` back to the type of the input.

Consider a simple generic L-2 `norm()` implementation that doesn't rely on the `abs` function:

Today, this might look like:

```chapel
/* Take the L-2 vector norm of a vector with coordinates x and y. */
proc norm(x: ?t, y: ?t) where isNumeric(x) && isNumeric(y) {
    return sqrt((conjg(x): t)*x + (conjg(y): t)*y);
}
```

With this change, it could look like:
```chapel
/* Take the L-2 vector norm of a vector with coordinates x and y. */
proc norm(x, y) where isNumeric(x) && isNumeric(y) {
    return sqrt(conjg(x)*x + conjg(y)*y);
}
```

This PR ensures the `conjg` function always has the same return type as the argument type.

### Testing

- [x] linux64 testing *passed*

### DIFF
--- a/modules/standard/Math.chpl
+++ b/modules/standard/Math.chpl
@@ -371,10 +371,10 @@ module Math {
 
      :rtype: A complex number of the same type as `z`.
   */
-  inline proc conjg(z: ?t) where isComplex(z) {
+  inline proc conjg(z: complex(?w)) {
     extern proc conjf(z: complex(64)): complex(64);
     extern proc conj(z: complex(128)): complex(128);
-    if numBits(t) == 64 then
+    if w == 64 then
       return conjf(z);
     else
       return conj(z);
@@ -384,7 +384,7 @@ module Math {
 
      :rtype: An imaginary number of the same type as `z`.
   */
-  inline proc conjg(z) where isImag(z) {
+  inline proc conjg(z: imag(?w)) {
     return -z;
   }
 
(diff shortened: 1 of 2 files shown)


modules/standard/Math.chpl
```diff
@@ -380,6 +381,22 @@ module Math {
...
+  */
+  inline proc conjg(z: imag(?w)): imag(w) {
+    return -z;
+  }
+
+  /* Returns the argument `z`.
+
+     :rtype: A number that is not complex or imaginary of the same type as `z`.
+  */
+  inline proc conjg(z: ?t) where isNumericType(t) {
```
mppf: proc isNumericType(type t) param return
  isIntegralType(t) || isFloatType(t) || isComplexType(t);


-> this should say where isIntegralType(t) || isFloatType(t)

modules/standard/Math.chpl
```diff
@@ -367,11 +367,12 @@ module Math {
   }
 
 
-  /* Returns the complex conjugate of the argument `z`.
+  /* Returns the complex conjugate of the complex argument `z`.
 
      :rtype: A complex number of the same type as `z`.
   */
-  inline proc conjg(z: complex(?w)): complex(w) {
+  inline proc conjg(z: complex(?w)) where isComplexType(z.type): complex(w)
```
mppf: Surprised the 'where' clause is necessary here

mppf:
Looks OK other than my nit-picking about where clauses.

modules/standard/Math.chpl
ben-albrecht: I was too. It's necessary because of the explicit `: complex(?w)` type. Other numeric types get coerced to complex resulting in this overload taking priority over the other more general overload. I just made a commit that drops the explicit type so that all 3 overloads consistently use `where` clauses to resolve distinctly.

PR #5420, opened 2017-02-24
Add test of Chapel array API and improve API somewhat

bradcray:
In working on array views, I decided to write a test that tried
to exercise all standard user-facing APIs on arrays in order to
ensure that they work on array views as well.  In doing so, I
made some improvements to the array API to make some cases work
better:

* added an error overload to IRV so that when called on non-sparse
  arrays it would give a nice message

* changed the where clauses on the array-vector ops into param
  conditionals with compilerError() messages to give errors when
  applied to arrays other than 1D default rectangulars.

* fixed a bug in the find() routine in which it would only work for 1D
  arrays (when I think we want it to work for general arrays, and
  nothing about its signature or documentation suggested otherwise)

This test is independent of the array views work, as are the
improvements, so I'm filing it against master separately.

The next step here is to create versions of the test that pass in a slice,
a rank change slice, and a reindexed array to make sure they work as
expected on both master and the array views branch.


bradcray:
@daviditen : Would you have time to review this on Friday?  If not, let me know if the changes to the array-as-vec routines are OK with you and I'll ask someone else to review.

bradcray:
PS -- I'm firing off testing concurrently to making this request...

bradcray:
Linux64 testing came back clean.

daviditen:
@bradcray Sure, I can review it.

modules/internal/ChapelArray.chpl
```diff
@@ -2245,9 +2245,13 @@ module ChapelArray {
       _value.dsiSerialRead(f);
     }
 
+    proc IRV where !isSparseArr(this) {
```
daviditen: I'd suggest leaving off this where clause.  Since the two IRV functions cover all cases with where clauses, it leaves no room for other overloads, e.g. for a more specific error message.

test/arrays/userAPI/arrayAPItest.chpl
```diff
@@ -0,0 +1,152 @@
+config param testError = 0;
+
+proc readArray(X) {
+  open("realValues.txt", iomode.r).reader().read(X);
```
daviditen: I think you should close() the channel and file here.  We plan to reference count these, but I don't think we currently do.

test/arrays/userAPI/arrayAPItest.chpl
```diff
@@ -0,0 +1,152 @@
...
+  for locdom in X.localSubdomains() do
+    writeln(locdom);
+  writeln();
+
+  // Test collection interface
+  writeln("is empty: ", X.isEmpty());
+  writeln("head: ", X.head());
+  writeln("tail: ", X.tail());
+  writeln("find 2.8: ", X.find(X[X.domain.high]));
+  writeln("count 2.8: ", X.count(X[X.domain.high]));
```
daviditen: The value that find and count are looking for is 16.8, not 2.8.

daviditen:
I made a few fairly minor comments, the rest looks good!

bradcray:
@daviditen: Where are the comments, I don't see them?

daviditen:
Oops.  I forgot to click submit review.  They should be there now.

bradcray:
Good comments, thanks.  W.r.t. the IRV change, I'm planning to only test against tests that query the IRV directly and obvious sparse directories rather than re-testing everything -- is that OK with you?

Also, I'm not seeing any memory leaks on the test -- am I correct that in suggesting we close the file/channel it was leaks you were concerned about?  if so, I'll confirm on chat that it's necessary (not that adding closes would be an issue, except that it would throw off all the line numbers for my error messages...  This is also the reason that the creation of file and channel are so stacked -- I was reading from stdin until the very last mod...)

bradcray:
Oh, testing reminds me why I put the where clause on the error IRV call.  If I don't, the getter/setter pair resolution calls it on true sparse arrays in getter scenarios.  I suspect this could be worked around in other ways (providing additional overloads of IRV... maybe removing the 'ref' from the existing IRV call?), but these seemed more frightening to me than adding the where clause.  Unless you have strong objections, I'd prefer to keep the where clause on both cases today and wait until there's a more compelling reason to relax one or both of them, reconsider their return intents, etc.  (simply because my goal was to generate a useful error message where we don't get one today, not to make IRV more perfect than it is already).  Unless you have a compelling/realistic case in mind where we/a user would want to provide another IRV overload on the _array record (I interpreted your comment as more of a "maybe someday" thing...).

daviditen:
For the suggestion to close the channel and file, I was concerned with memory leaks and about leaking the open file descriptor.  The documentation on master says it's still necessary, but maybe that needs an update.

I'm OK with leaving the where clause on then.  Yes, it was a "maybe someday" thought.

bradcray:
For posterity: On chat, we clarified that the documentation refers to the time prior to record memory management being fixed/improved.  Now that it is, resources are reclaimed as desired.  For that reason (and because I don't want to have to update all the error message .good files), we're leaving the closes out.

PR #5421, opened 2017-02-24
Make the FFTW includes and libraries param strings

npadmana:
The PR has two changes to the FFTW module --

1. It removes the explicit require command for the FFTW libraries. This allows the user to link in different
versions of FFT libraries (which may, eg., be machine dependent), as long as they conform to the FFTW3 calling definitions. An example here is the Intel MKL, which provides FFTW3 compatible wrappers. 

A side benefit of this is that it will allow us to merge the FFTW and FFTW_MT modules. 

2. We've added in a switch for the Intel MKL wrappers, which includes the additional MKL header files. Compiling with `-sisFFTW_MKL` includes these headers.

------------------------------------------------------------------------------------------------------
(The original PR comment is below, for archival purposes)

The current FFTW module has hardcoded strings to specify its include and library dependencies. This makes using a different FFTW implementation (eg. the MKL version) harder, since the libraries are different and there are additional include files.

This pull request makes these param strings, defaulting to their original values. One can now change these at compile time. For the default values, this passes the tests test/release/examples/primers

Trying this on the MKL version, there were a few surprises here. I required the following compile command

```
 chpl -o FFTWlib FFTWlib.chpl  \
   -sFFTW_includes='"fftw3_mkl.h"' \
   -sFFTW_includes='"fftw3.h"' \
   -sFFTW_libraries='"-lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -lgomp -lpthread -lm -ldl"'  \
   -I$MKLROOT/include/fftw
```

1. Note that I needed to double-quote the filenames, otherwise I got an undefined symbol error.
2. I could not figure out how to put both includes in one string. Using 
   '"fftw3_mkl.h","fftw3.h"' resulted in a 
 Command-line arg (FFTW_includes): syntax error parsing '"fftw/fftw3.h","fftw/fftw3_mkl.h"'

  '"fftw3_mkl.h fftw3.h"' or '"fftw3_mkl.h, fftw3.h"' 
resulted in a file not found.  

Somewhat surprising to me was the fact that putting each in its own option worked; I'm guessing that the semantics for these parameters concatenates strings.

3. Interestingly, for the libraries, putting the full list in one string worked fine. 

In putting this together, I couldn't seem to find the syntax for require described anywhere... 

If we like this solution, we can update the multi-threaded version as well, as well as BLAS/LAPACK.

@bradcray, @ben-albrecht --- this might interest you.

### DIFF
--- a/modules/packages/FFTW.chpl
+++ b/modules/packages/FFTW.chpl
@@ -88,13 +88,15 @@
 
 module FFTW {
 
-  // Define includes and libraries
-  config param FFTW_includes : string ="fftw3.h";
-  config param FFTW_libraries : string ="-lfftw3";
+  // The MKL implementation of FFTW requires an additional
+  // include file
+  config param isFFTW_MKL=false;
 
   use SysCTypes;
-  require FFTW_includes, FFTW_libraries;
-
+  require "fftw3.h"; // This is common
+  if (isFFTW_MKL) {
+    require "fftw3_mkl.h";
+  }
 
   /*
     Controls execution-time array size checks in the FFTW
--- a/modules/packages/FFTW.chpl
+++ b/modules/packages/FFTW.chpl
@@ -36,16 +36,19 @@
 
   2. Add ``use FFTW;`` to your Chapel code.
 
-  3. Include the appropriate libraries in your compilation command. For
-     a standard FFTW installation, this may be as simple as including
-     ``-lfftw3`` when compiling. You may also need to use the ``-I`` and
-     ``-L`` flags to specify the locations of the header and library files.
+  3. Include the appropriate libraries in your compilation command.
 
-  .. note::
-   To use the MKL FFTW wrappers, compile with ``-sisFFTW_MKL`` to include
-   the ``fftw3_mkl.h`` header in addition to the usual ``fftw3.h`` header
-   file. You may also need to add ``-I${MKLROOT}/include/fftw`` to point the
-   compiler to the location of these header files.
+     a. For a standard FFTW installation, this may be as simple as including
+        ``-lfftw3`` when compiling. You may also need to use the ``-I`` and
+        ``-L`` flags to specify the locations of the header and library files
+        if these are in non-standard locations.
+
+     b. *Intel MKL* : To use the MKL FFTW wrappers, compile with ``-sisFFTW_MKL`` to include
+        the ``fftw3_mkl.h`` header in addition to the usual ``fftw3.h`` header
+        file. You may also need to add ``-I${MKLROOT}/include/fftw`` to point the
+        compiler to the location of these header files. Refer to the Intel MKL
+        documentation for the appropriate libraries to include.
+
 
   As in standard FFTW usage, the flow is to:
 
--- a/modules/packages/FFTW.chpl
+++ b/modules/packages/FFTW.chpl
@@ -29,17 +29,23 @@
 
   To use this module:
 
-  1. Ensure that FFTW (version 3) is installed on your system and that
-     the header and library files (e.g., fftw3.h, libfftw3.*) are
-     either installed in a standard system location or that your C
-     compiler's environment variables are set up to find them
-     (alternatively, the Chapel compiler's ``-I`` and ``-L`` flags can
-     be used to specify these locations).
+  1. Ensure that FFTW (version 3) is installed on your system. The
+     current version of the Chapel module only supports double precision ``real(64)``
+     transforms. We do support using the FFTW compatible wrappers provided
+     by the Intel Math Kernel Library (MKL) (see below for usage).
 
   2. Add ``use FFTW;`` to your Chapel code.
 
-  3. Compile and run your Chapel program as usual.
+  3. Include the appropriate libraries in your compilation command. For
+     a standard FFTW installation, this may be as simple as including
+     ``-lfftw3`` when compiling. You may also need to use the ``-I`` and
+     ``-L`` flags to specify the locations of the header and library files.
 
+  .. note::
+   To use the MKL FFTW wrappers, compile with ``-sisFFTW_MKL`` to include
+   the ``fftw3_mkl.h`` header in addition to the usual ``fftw3.h`` header
+   file. You may also need to add ``-I${MKLROOT}/include/fftw`` to point the
+   compiler to the location of these header files.
 
   As in standard FFTW usage, the flow is to:
 
@@ -88,8 +94,10 @@
 
 module FFTW {
 
-  // The MKL implementation of FFTW requires an additional
-  // include file
+  /*
+    Set this to `true` if you are using the Intel MKL FFTW
+    wrappers
+  */
   config param isFFTW_MKL=false;
 
   use SysCTypes;
--- a/modules/packages/FFTW.chpl
+++ b/modules/packages/FFTW.chpl
@@ -89,7 +89,7 @@
 module FFTW {
 
   // Define includes and libraries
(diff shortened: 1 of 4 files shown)


bradcray:
> Note that I needed to double-quote the filenames, otherwise I got an undefined symbol error.

Hi @npadmana : This is expected, and can be explained as follows:

* your shell gobbles up one set of quotes, so they never make it to the Chapel compiler
* the Chapel compiler needs the second set of quotes to determine that you're trying to supply a string literal rather than some other expression.  In particular, config params permit you to override their default value with some reasonably arbitrary Chapel code expression that would be the equivalent of dropping that expression into the program text.  Examples of this can be seen at http://chapel.cray.com/docs/master/users-guide/base/configs.html#config-param-and-config-type  An example specific to this context is that you could say something like -sFFTW_libraries='FFTW_includes+"goober"' to insert a string concatenation expression into the initializer for FFTW_libraries.

> Using [one string] resulted in a syntax error

This is because the string you supplied is dropped into the program text as the config param's initializer and a single variable can't have a comma-separated initializer.  I.e., the following is a syntax error in Chapel, so trying to supply that value for a config param's initializer is as well:

````chapel
config param foo = "fftw3_mkl.h","fftw3.h";
````

> [Using two strings] resulted in a file not found

This is because the code we generate would end up saying:

````
#include "foo.h, bar.h"
````

[of all these surprises, this is the one that's most likely negotiable -- perhaps Chapel should permit specifying multiple requirements in a single string and require the compiler to parse them apart... I'd prefer not to do that for simplicity, but maybe it would be convenient and more orthogonal to the libraries case below].

> Somewhat surprising to me was the fact that putting each in its own option worked; I'm guessing that the semantics for these parameters concatenates strings.

I think it's more likely that it's only necessary to #include one of the two header files to make MKL FFTW work (probably the second?).  I'd suggest printing out the config param within the program to verify my guess that only one of these "stuck."

> Interestingly, for the libraries, putting the full list in one string worked fine.

I think this is simply because the library requirements file gets dropped into the compiler's link command-line which accepts an arbitrary number of comma-separated arguments (where #include and Chapel initializers do not).

> In putting this together, I couldn't seem to find the syntax for require described anywhere...

It hasn't made it into the spec yet, but see:  http://chapel.cray.com/docs/master/technotes/extern.html?highlight=external#expressing-dependencies


Note that I believe another way we could do this (untested by me) would be:

````chapel
config param FFTW_MKL = false;

if (FFTW_MKL) {
  require "fftw/fftw3.h";
  require "fftw/fftw3_mkl.h";
  require "-lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -lgomp -lpthread -lm -ldl";
} else {
  require "fftw3.h";
  require "-lfftw3";
}
````

(This is also new since 1.14: Ability to have require statements appear at various scopes and to only consider those that don't get param-folded out by the compiler).

Upside: user only has to throw one flag to switch between implementations
Downside: code has to change in order to add additional flavors (e.g., bool could become an enum, but each new implementation would have to add a new value to the enum and a new branch to a conditional/select statement

File: modules/packages/FFTW.chpl [hunk 1]
```diff
@@ -88,8 +88,12 @@
 
 module FFTW {
 
+  // Define includes and libraries
+  config param FFTW_includes : string ="fftw/fftw3.h, fftw/fftw3_mkl.h";
```
bradcray: I'd expect that the line above would not work and would result in the similar "file not found" error as you reported on the command-line.  I'm guessing that maybe you were experimenting with different values in the source and committed this one instead of one that simply set this to "fftw3.h" (the non-MKL default you said you'd implemented)?

npadmana: @bradcray -- oops, that's embarassing. Fixed.

npadmana:
Hi @bradcray --

Thanks for the very detailed discussion --- this all makes sense to me. 

I think it's worth separating out the ``#include`` step from the link step. 

For the includes, I think your suggestion of a switch to pick between different implementations is probably the right one. It's quite unlikely that there will be a proliferation of cases, and the convenience of a switch over manually specifying header files is a win. I think this works for both BLAS as well as FFTW.

The linking case is more muddled, since it is compiler and implementation and likely hardware specific (the MKL link line I sent was for GCC on intel64 etc etc etc....). For BLAS, we took the position that specifying the link details is something that the user should handle and we didn't hardcode it into the module. 

I think that is the right solution; I  think it's usual for the user to need to think about what their linker flags are, and it means we don't tie the module to a particular case.

There is a nice side benefit to that for FFTW. We currently have a separate module to handle the multi-threaded case ``FFTW_MT``. However, the only dependencies here are at link time, so we could just merge the two into one.

So here is my suggestion on how to proceed
- [ ] We add in a switch for MKL. We could do an enum, but (a) that is the only case I know of, and (b) I can't think of a good name!
- [ ] We remove the link require, and let the user specify it by hand. We'll need to update the compopts file.
- [ ] Update the documentation...

In a separate pull, we could merge in the FFTW_MT module....

Thoughts?

ben-albrecht:
I like the approach you and Brad have proposed.

> The linking case is more muddled, since it is compiler and implementation and likely hardware specific (the MKL link line I sent was for GCC on intel64 etc etc etc....). For BLAS, we took the position that specifying the link details is something that the user should handle and we didn't hardcode it into the module.

We *could* infer some of this information by reading values from the [Chapel environment variables](http://chapel.cray.com/docs/master/builtins/internal/ChapelEnv.html), but it might get messy depending on the sensitivity of the naming schemes. I would be in favor of providing link flags for any dependency that we supply ourselves, e.g. FFTW, and possibly BLAS/LAPACK in the future, to minimize setup for these modules.

npadmana:
@ben-albrecht -- 

>  I would be in favor of providing link flags for any dependency that we supply ourselves, e.g. FFTW, and possibly BLAS/LAPACK in the future, to minimize setup for these modules.

I guess I was proposing the opposite --- that we don't provide link flags. Unless you were thinking of dependencies that are packaged with the Chapel source, in which case I agree with you. However, note that FFTW isn't supplied with Chapel, just the module wrapper...

ben-albrecht:
@npadmana --

> Unless you were thinking of dependencies that are packaged with the Chapel source, in which case I agree with you. 

Sorry for the confusion, this is what I meant.

> However, note that FFTW isn't supplied with Chapel, just the module wrapper...

Whoops, that would make FFTW a bad example. Maybe libhdfs3 is a better example.

bradcray:
@npadmana : I like your proposed next steps.

> There is a nice side benefit to that for FFTW. We currently have a separate module to handle the multi-threaded case FFTW_MT. However, the only dependencies here are at link time, so we could just merge the two into one.

Yeah, I meant to return to this case after making the require statements more flexible to see what could be done to clean it up, but failed to do so (and probably will continue to, at least for awhile).

npadmana:
@bradcray, @ben-albrecht -- I think this is ready for review....

bradcray:
I haven't done any testing yet (of the code or the generated documentation), but browsing the code, it looks good to me.  Would you update the PR message to reflect where the code ended up rather than where it started?  I'm going to cut and paste some of the text that I was replying to into my original response comment to not lose the context (this is a good reason to keep the PR comment focused on what the PR is and then move commentary into a first comment on the PR).

npadmana:
I've updated the PR comment. I haven't had a chance to check the generated documentation as yet....

npadmana:
@ben-albrecht -- 

Fixed. I word-smithed the docs some more, so you may want to quickly look over that.

I've updated the compopts file as well; both the tests that I missed now pass on my system.

PR #5422, opened 2017-02-24
Let hwloc auto-detect libnuma for non-flat locModels; don't force it.

gbtitus:
This fixes a flaw in #5418 (yesterday).  There I arranged to throw
--enable-libnuma to the hwloc configure step for locale models other
than 'flat'.  But when libnuma is not present on the system, doing so
will cause the configure step to produce an error.  On systems without
libnuma, we want to be able to use such locale models even though they
might be hobbled by the lack of NUMA info.  So instead of insisting that
hwloc use libnuma in this case, just let it auto-detect libnuma, and
we'll get by as best we can if it's not there.

### DIFF
--- a/third-party/hwloc/Makefile
+++ b/third-party/hwloc/Makefile
@@ -30,10 +30,10 @@ CHPL_HWLOC_CFG_OPTIONS += --enable-static \
                           --disable-opencl \
                           --disable-pci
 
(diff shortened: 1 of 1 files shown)


gbtitus:
I suspect this doesn't really need review, but @ronawho if you want to glance at it, that would be reassuring.

PR #5424, opened 2017-02-24
Implement support for initializers on types with generic type fields

lydia-duncan:
Prior to this change, initializers were not supported on generic types.  This was
because the strategy for creating a generic instantiation needed to be different
for initializers than the strategy currently used by constructors.  Initializers do not
rely on the generic fields being explicitly duplicated in their argument list to
determine the generic instantiation - rather, the generic fields are set during
Phase 1 of the initializer body, along with the value of any non-generic fields.
This gives more power to the writer of an initializer, allowing them to support
initializers that would not have been possible in the constructor world.  This
change adds the framework (but not the full implementation) for supporting all
possible generic types which would define initializers, and adds support for
generic types with type fields.  It is expected that the extension of this work to
support generic types with param or generic var/const fields will be a minor
extension of this work, but for sanity's sake we decided to merge the type field
case first.

This PR/merge can be broken down into three key components, described in
detail in their commits.  There are four minor clean up commits for some
mistakes I made when dividing the change into logical commits and one
commit for my testing changes.

The three components are: adding and utilizing a new primitive,
PRIM_INITIALIZER_SET_FIELD; modifying resolveNew to respond to
generic initializers; and related clean ups and modifications to checks
which were made necessary by this work.

Details:
---------
- Adding and utilizing the new primitive is described in detail in commit five,
"Add support for a new primitive...".  Commits nine, ten and eleven rectify some
oversights from this commit.
- Modifying resolveNew is described in commit six, "Modify resolveNew...".
Commit eight removes a comment I didn't mean to include in this commit. Commit
thirteen adds some additional modifications to support utilizing the generic type
multiple times.
- The related clean ups and modifications to checks are found in commits two,
three, and seven ("Clean up the init and exprType...", "Modify some checks...",
and "Avoid accidental clean up...").  Commit four amends commit two.

Testing changes: 13 new tests and 2 modifications
See description in the first commit ("Test updates - 8 new tests and 2
modifications") and the last two commits prior to the master merge and
clean ups.

Full standard testing passed


lydia-duncan:
This PR continues work on resolving issue #5198

compiler/resolution/functionResolution.cpp
```diff
@@ -4296,6 +4297,7 @@ resolveCall(CallExpr* call)
      case PRIM_TUPLE_EXPAND:        resolveTupleExpand(call);           break;
      case PRIM_SET_MEMBER:          resolveSetMember(call);             break;
      case PRIM_MOVE:                resolveMove(call);                  break;
+     case PRIM_INITIALIZER_SET_FIELD: resolveFieldInit(call);           break;
```
noakesmichael: It looks like the neighboring indentation needs to be updated here.

compiler/AST/symbol.cpp
```diff
@@ -1598,6 +1598,16 @@ int FnSymbol::hasGenericFormals() const {
   bool hasGenericDefaults =  true;
   int  retval             =     0;
 
+  bool inInitRes = false;
```
noakesmichael: What does "res" mean in inInitRes? Maybe there's a better name.  Also alignment does not
appear to match the nearby locals if github is rendering correctly.

compiler/AST/type.cpp
```diff
@@ -492,6 +492,7 @@ AggregateType::AggregateType(AggregateTag initTag)
 
   aggregateTag       = initTag;
   initializerStyle   = DEFINES_NONE_USE_DEFAULT;
+  initializerResolved = false;
```
noakesmichael: It appears that neighboring alignment needs to be updated to match the new variable.

compiler/resolution/functionResolution.cpp
lydia-duncan: I will update alignment in a separate commit, I didn't want to muddy the change with non-functional operations.

compiler/AST/symbol.cpp
lydia-duncan: "res" is a shortening of resolution because I didn't want to make the argument name that long

lydia-duncan: Would resolveInit be a more appealing name?

noakesmichael:
1) In general names should be long enough to be sufficiently descriptive but it’s easy
to agree that it’s possible to choose names that are unnecessarily long.

2) In this context it’s not an argument name.  It’s a local variable for
FnSymbol:hasGenericFormals() which doesn’t immediately conjure up
resolution.  So this name isn’t super successful in this context



On Feb 24, 2017, at 12:39 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


In compiler/AST/symbol.cpp<https://github.com/chapel-lang/chapel/pull/5424#discussion_r103028296>:

> @@ -1598,6 +1598,16 @@ int FnSymbol::hasGenericFormals() const {
   bool hasGenericDefaults =  true;
   int  retval             =     0;

+  bool inInitRes = false;


"res" is a shortening of resolution because I didn't want to make the argument name that long

lydia-duncan:
Oh duh, you're right.  I was thinking of a previous iteration of the code where that was an argument.  Does the suggested new name sound better?

compiler/resolution/initializerResolution.cpp
```diff
@@ -33,14 +33,523 @@
...
+static void
+filterInitCandidate(Vec<ResolutionCandidate*>& candidates,
+                ResolutionCandidate* currCandidate,
+                CallInfo& info);
+
+static void
+filterInitCandidate(Vec<ResolutionCandidate*>& candidates,
+                FnSymbol* fn,
+                CallInfo& info);
+
```
noakesmichael: I believe that instantiateInitSig() is static to this file

compiler/include/initializerResolution.h
```diff
@@ -27,13 +27,17 @@
 // handled, as we want to have the this argument for the type in the initializer
 // argument list, and want to utilize Phase 1 of the initializer body to
 // determine the generic instantiation of it.
+class AggregateType;
```
noakesmichael: I don't think callInfo.h needs to be included in this header file

I also don't think it needs BlockStmt, DefExpr, Expr, or Symbol.

lydia-duncan: Thanks for the reminder!

lydia-duncan:
Running a last sanity paratest after my clean ups, but I don't expect any problems, since I ran one on the commits from yesterday.  This will be ready to merge after that.

PR #5426, opened 2017-02-24
 Add iterator bug with multiple yields

benharsh:
Add future demonstrating bug with a reduction over a standalone iter with multiple yields.


benharsh:
@vasslitvinov : this is the bug we were talking about. Is this an example you can work with (eventually)?

vasslitvinov:
@benharsh yes, this is perfect.

PR #5427, opened 2017-02-24
Update docs regarding refcounting of files and channels

bradcray:
The previous text here was written when we weren't doing memory
management of records properly, leading to a lack of guarantees
that we were closing them appropriately.  This tries to improve
up on this.

One hesitation I have with this current wording (inherited from                 
the previous version) is that it's vague in terms of what a                     
"reference" is.  I.e., I don't think the following would keep                   
a channel or file from closing, would it?                                       

````chapel                                                                                
var infile = ...                                                              
ref secondRef = infile;                                                       
````
                                                                                
What about the following?

````chapel
proc foo(ref f) { ... }

foo(infile);
````
(in these examples, the	original variable wouldn't outlive the
references, but	if we started mixing in	some asynchrony...)

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -289,14 +289,11 @@ that has been closed.
 It is an error to close a file when it has channels that
 have not been closed.
 
-In the future we plan to implement reference counting for files
-and channels. Each file and channel will be closed automatically
-when no references remain to it. For example, if only a local
-variable refers to a channel, the channel will be closed
-when that variable goes out of scope.
-
-The ability for the program to close a file or a channel
-explicitly will remain available.
+Files and channels are reference counted. Each file and channel is
+closed automatically when no references to it remain. For example, if
+a local variable is the only reference to a channel, the channel will
+be closed when that variable goes out of scope.  Programs may also
+close a file or channel explicitly.
 
 .. _about-io-style:
 
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -2124,9 +2124,9 @@ proc file._style:iostyle {
    running out of storage space or power loss. See also
    :ref:`about-io-ensuring-successful-io`.
 
-   In the future, we hope to automatically close files when the file variable
-   goes out of scope and all channels using that file are closed. The ability
-   for a program to close a file will remain available.
+   Files are automatically closed when the file variable
+   goes out of scope and all channels using that file are closed. Programs
+   may also explicitly close a file using this method.
 
    :arg error: optional argument to capture an error code. If this argument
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bradcray** [comment]
@mppf , @daviditen : For your consideration...

File: modules/standard/IO.chpl [hunk 1]
```diff
@@ -289,14 +289,11 @@ that has been closed.
...
-In the future we plan to implement reference counting for files
-and channels. Each file and channel will be closed automatically
-when no references remain to it. For example, if only a local
-variable refers to a channel, the channel will be closed
-when that variable goes out of scope.
-
-The ability for the program to close a file or a channel
-explicitly will remain available.
+Files and channels are reference counted. Each file and channel is
+closed automatically when no references to it remain. For example, if
```
mppf: it could say "when no variable refers to it"

modules/standard/IO.chpl
```diff
@@ -289,14 +289,11 @@ that has been closed.
...
+a local variable is the only reference to a channel, the channel will
+be closed when that variable goes out of scope.  Programs may also
+close a file or channel explicitly.
```
mppf: ... and in some cases closing a file explicitly might be important to get an error-on-write correctly reported.

modules/standard/IO.chpl
bradcray: Or maybe "when no further copies of it remain"?  (to avoid the ambiguity over whether, given:

````chapel
ref x = ...
````

x is a variable or not?

modules/standard/IO.chpl
bradcray: Is that text that will similarly become outdated when error-handling is online and incorporated?

daviditen:
There's another similarly worded comment for the file.close() method that would be good to update in the same way.

modules/standard/IO.chpl
mppf: I'm not so sure because it has to do with what we do about errors in destructors. I wasn't saying you should add it, just noting it. It's already described elsewhere.

modules/standard/IO.chpl
mppf: It could, but "no further copies" seems confusing to me b/c we're not really copying the File or Channel... Maybe we don't want to remove "references" :)

bradcray:
Thanks @daviditen, I missed that and have updated it now.

Barring objections, I feel inclined to submit the current text as-is on the argument that it's strictly better / more accurate than what we had before and leave questions of wording to the next major draft to the text, or the first time someone is confused by it.  Is that OK?

daviditen:
That sounds fine to me.

PR #5429, opened 2017-02-24
Add annotation for last week's performance triage

lydia-duncan:
Covers turning hyperthreading on for one of our performance machines and a
slight change in performance due to the better cleaning up of unused globals.

Verified that `start_test --generate-graphs` worked with my change and broke if
I inserted an intentional error


lydia-duncan:
Going to merge on Monday so I don't break weekend performance testing

PR #5431, opened 2017-02-25
Test array API on views

bradcray:
The primary goal of this PR is to file a series of three tests which apply the same array API test from yesterday on (a) a slice of an array, (b) rank change slices of an array, (c) a reindexed array to make sure it works as expected.  This will then be taken to the array view branch to add confidence that no key array APIs are missing.

While getting the test working, I found a few bugs in arrays (and in my API test) that had easy fixes:

* head/tail (and my test yesterday) queried the domain's low and high bound rather than using alignedLow and alignedHigh, which made them brittle to strided domains in which the domain's extreme values weren't in the index set.  E.g., for 'A: [1..10 by 2] real', they would try to access A[10] rather than A[9] to get the high value.

* the existing reindex functions were insufficiently general in the presence of strided domains in the source and target index sets.  This was a bug I'd already addressed in the array views branch, but which none of our existing tests seem to run afoul of.

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -1013,7 +1013,8 @@ module ChapelArray {
     pragma "no doc"
     proc this(args ...rank) where _validRankChangeArgs(args, _value.idxType) {
       var ranges = _getRankChangeRanges(args);
-      param newRank = ranges.size, stridable = chpl__anyStridable(ranges);
+      param newRank = ranges.size,
+            stridable = chpl__anyStridable(ranges) || this.stridable;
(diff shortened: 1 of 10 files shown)


bradcray:
@daviditen: You should make sure to look at the fixes to the head() and tail() methods in case any of your other vector routines need a similar fix (I didn't check and suspect it depends heavily on whether or not we permit strided arrays to have vector operations or not).

daviditen:
@bradcray All of the vector routines currently require unstridable arrays, so I don't think they can run into this.  If we extend them to allow strided arrays, we'll need to keep this in mind.

bradcray:
Thanks for checking.  I thought that might be the case, but didn't take the minute or two to check...

PR #5439, opened 2017-02-27
Add 'stridable' query to the Replicated Array class

bradcray:
It seems that our Replicated array class hasn't supported 'stridable'
queries though ChapelArray.chpl assumes that all rectangular arrays will
and the Replicated domain and local array class support it.  Add it in the
trivial way by adding a method that defers to the domain's stridable
value.

This fixes the weekend's regressions due to my bug fix for reindexing
between stridable ranges/domains.


bradcray:
@vasslitvinov : This is arguably trivial, but would you give it a quick look and see if it raises any flags for you?

vasslitvinov:
Looks good to me.

Issue #5266, opened 2017-01-27
inconsistency for record blank intents, this, and const


mppf:
Right now, records have inconsistent blank intents. The `this` formal is passed by `ref` by default for record methods and const checking here is not done right.

Consider this code:

``` chapel
record R {
  var i:int;
  proc foo() {
    i = 3;
  }
}


const r = new R();
r.foo();
writeln(r);
```

This example compiles and runs now but should be a compilation error. Arguably, modifying record fields in methods is a convenience of a similar type to being able to modify array elements from a blank intent formal.

Here are the options we are considering:

1) Change default intent for 'this' records into 'const ref' always and require the user to create 'ref' methods for things that should actually modify the record.

 * Pro: makes default intents consistent for 'this' and non-'this' arguments
 * Pro: makes treatment of records more consistent with scalars (our other main value types)
 * Con: adds some syntactic/programmer overhead to the (arguably common) case of defining record methods that modify fields

2) Keep the current default intent for 'this' records as 'ref' but do a better job of const-checking such methods when called on const records.

* Pro: supports the common case of record methods that modify fields
* Con: makes default intents inconsistent between 'this' and non-'this' arguments [blc edit: This used to say "Pro", but should be "Con" right?]
* Con: adds some syntactic/programmer overhead to the (arguably common) case of supporting read-only record methods on constant records.
 * Con: 'this' default intent does not match default argument intent.
 * Con: ref-pair surprises with record methods

3) Make the default intent for 'this' records be 'const ref' or 'ref' based on whether the method is read-only or read-write for the record. (ref-if-modified on this formal only).

 * Pro: permits most record methods to not declare "this intents"
 * Con: 'this' default intent does not match default argument intent.

4) Make the default intent for all record arguments be 'ref' or 'const ref' based on whether the method is read-only or read-write for the argument (i.e. ref-if-modified #5219 for records in addition to arrays).

 * Pro: makes default intents consistent between 'this' and non-'this' arguments
 * Pro: arguably supports "my record should be array-like" cases well
 * Con: arguably doesn't support "my record should be int-like" cases well
 * Con: may feel surprising to C programmers

5) Provide some mechanism for saying what blank intent should mean for a record. Users can select between "ref-if-modified" and "const ref" for blank intent. The 'this' intent matches the blank intent.

 * Pro: makes default intents consistent between 'this' and non-'this' arguments
 * Pro: flexible -- supports array-like and int-like records
[...]

PR #5444, opened 2017-02-27
spec: describe array passing, returning, ref-if-modified

mppf:
This is the spec change to go along with PR #5134. See also issues #5217 and #5219.
While here, also include record  _this arguments, to go along with PR #5618 and issue #5266.

See also #5767 which changes evolution.rst.

### DIFF
--- a/spec/Functions.tex
+++ b/spec/Functions.tex
@@ -593,7 +593,7 @@ type & default intent \\
 \chpl{sync}    & \chpl{ref}   \\
 \chpl{single}  & \chpl{ref}   \\
 \chpl{atomic}  & \chpl{ref}   \\
-\chpl{record}  & \chpl{const} \\
+\chpl{record}  & \chpl{const} (but see below) \\
 \chpl{class}   & \chpl{const} \\
 \chpl{union}   & \chpl{const} \\
 \chpl{dmap}    & \chpl{const} \\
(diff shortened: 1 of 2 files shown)


benharsh:
This looks alright to me, but I've been heavily involved in the related discussions leading to these changes. I think it would be good to find someone less familiar with these topics to look this over.

spec/Functions.tex
```diff
@@ -593,16 +593,27 @@ \subsubsection{The Default Intent}
...
+The default intent for arrays is \chpl{ref} or \chpl{const ref} - meaning
+that the array is not copied by default. The compiler will choose
+\chpl{ref} or \chpl{const ref} based upon whether or not the formal
+argument is modified inside of the function. The choice between
+\chpl{ref} and \chpl{const ref} is similar to and interacts with the
+ref-pair feature (see \rsec{Ref_Pair}).
+
+The default intent for the implicit \chpl{this} argument for records is
+\chpl{ref} or \chpl{const ref} depending on whether the formal argument
+is modified inside of the function.
```
bradcray: While gathering links for the CHANGES file, I'm noting that this seems like it ended up in the wrong place and feels like somewhat of a non-sequitur here.  This section is about default intents for formal arguments.  While 'this' is a formal argument in the implementation, for users, the spec section dealing with how 'this' is passed is in Classes.tex "The Method Receiver and the this Argument" (search on 'this-intent')

mppf: See #5883

PR #5445, opened 2017-02-27
Refactor resolveNew()

noakesmichael:
I happened to notice that resolveNew() has grown a bit wholly over time and it's
now surprisingly hard to recognize that it's actually doing something quite easy.
Things have been getting even worse as Lydia and I have been adding more cases
for initializers.

This PR performs a number of simple transformations, all as separate commits, that
serve to completely rewrite resolveNew().  IMO the final version is much easier to follow
than before.  Additionally it is much more clear what path is followed by existing production
code and what paths are being followed for the new initializer work.

Passed single-locale paratest on linux64.
Also passes test/classes/initializers on darwin.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -5162,8 +5162,6 @@ static void resolveNew(CallExpr* call) {
             call->parentExpr->insertBefore(def);;
           }
 
-          resolveExpr(def);
-
           if (isClass(typeToNew) == true) {
             // Invoking a type  method
             call->insertAtHead(new SymExpr(typeToNew->symbol));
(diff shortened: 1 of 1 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -5155,6 +5151,11 @@ static void resolveNew(CallExpr* call) {
             }
 
             resolveExpr(call);
+
+          } else {
+            USR_FATAL(call,
+                      "Sorry, new style initializers don't work with "
+                      "generics yet.  Stay tuned!");
```
lydia-duncan: This change isn't really important, since it will change entirely on my branch, but there isn't much point to removing it from your branch either.

lydia-duncan:
This all looks in keeping with what we discussed, thanks!

PR #5462, opened 2017-02-28
Add simple numa localization

gbtitus:
This adds a primitive capability to set the NUMA locality of array
memory.  In the runtime we add a chpl-topo module that can set NUMA
locality for a block of memory, either localizing the whole thing to a
single sublocale (NUMA domain) or doing a blocked localization across
all the sublocales.  DefaultRectangularArr:initialize() specifies in its
calls to _ddata_allocate() how to set the locality for array memory when
there are multiple sublocales, and ChapelBase:_ddata_allocate() calls
the new chpl_topo runtime code to do the actual work of localizing the
memory.  The runtime topology also includes a function to retrieve the
locality of an address, and there is a new test to check that everything
works as expected.


runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+  // Limit discovery.
+  //
+  if (hwloc_topology_set_flags(topology, HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM)) {
+    report_error("hwloc_topology_set_flags()", errno);
+  }
+
+  //
+  // Perform the topology detection.
+  //
+  if (hwloc_topology_load(topology)) {
```
ronawho: Note that hwloc_topology_load is relatively expensive (the source of most of our startup time): https://github.com/Qthreads/qthreads/issues/41#issuecomment-267716248

My preference would be to only initialize for non-flat locale modules, but I haven't actually gone through the PR in detail to know if that makes sense yet.

And longer term I think we (chapel runtime) want to do the topology load and just pass the topology onto qthreads, but obviously that's for another time.

dmk42:
Greg, if you don't mind, I'm going to take a look at this too.  I need your hwloc topology initialization for KNL because there is an hwloc function to tell what kind of memory a particular address lives in.

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+  node = hwloc_bitmap_first(nodeset);
+
+  hwloc_bitmap_free(nodeset);
+
+  return node;
+}
+
+
+static
+void report_error(const char* what, int errnum) {
```
ronawho: Any reason to use report_error instead using chpl_internal_error directly? Doesn't seem like it's providing much abstraction/functionality IMO.

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+  // Allocate and initialize topology object.
+  //
+  if (hwloc_topology_init(&topology)) {
+    report_error("hwloc_topology_init()", errno);
+  }
+
+  //
+  // Limit discovery.
+  //
+  if (hwloc_topology_set_flags(topology, HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM)) {
```
ronawho: Is this needed/desired (only wondering since qthreads hwloc interface does not set it)

ronawho:
I'm timing out for the day, but so far I'm happy with this PR. I think we should talk about the extra cost of doing hwloc_topology_load, but that's pretty minor.

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+    for (level = 0, numaLevel = -1;
+         level < topoDepth && numaLevel == -1;
+         level++) {
+      if (hwloc_get_depth_type(topology, level) == HWLOC_OBJ_NUMANODE) {
+        numaLevel = level;
+      }
+    }
+  }
+
+  numNumaDomains = hwloc_get_nbobjs_by_depth(topology, numaLevel);
```
dmk42: Just a heads-up, this will give you the total number of NUMA nodes, including both compute nodes and memory-only nodes.  That's correct, just wanted to make sure you knew that.  It means that on KNL, it can return twice the number of compute nodes, so it can be different from chpl_task_getNumSublocales().

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+         (int) hwloc_bitmap_first(nodeset));
+  if (hwloc_set_area_membind_nodeset(topology, p, size,
+                                     nodeset, HWLOC_MEMBIND_BIND, flags)) {
+    report_error("hwloc_set_area_membind_nodeset()", errno);
+  }
+}
+
+
+c_sublocid_t chpl_topo_getMemLocality(void* p) {
+  const int flags = HWLOC_MEMBIND_STRICT | HWLOC_MEMBIND_BYNODESET;
```
dmk42: Because you're calling hwloc_get_area_membind_nodeset instead of just hwloc_get_area_membind, the HWLOC_MEMBIND_BYNODESET flag is not used and all you need is HWLOC_MEMBIND_STRICT.

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,249 @@
...
+#endif
+
+static hwloc_topology_t topology;
+
+static const struct hwloc_topology_support* topoSupport;
+
+static int topoDepth;
+
+static int numaLevel;
+static int numNumaDomains;
```
dmk42: I'd like to have a getter for numNumaDomains.  If you'd rather not do that now, I can add it as part of the KNL changes.

runtime/src/chpl-topo.c
gbtitus: I'll adjust this so that for now the hwloc topology is only loaded for locModel!=flat.  As you say, in the long run at least with tasks=qthreads we'll want to load it even for flat and pass the result over to qthreads so that doesn't load it again.  But we can take that step later.

runtime/src/chpl-topo.c
gbtitus: I went and read the documentation again and now I think we don't need this in our usage, so I'm going to remove it.

runtime/src/chpl-topo.c
gbtitus: Seeing the system errno string has been very useful to me during development, so I'd like to keep this.

ronawho: good enough for me

runtime/src/chpl-topo.c
dmk42: On second thought, I should probably stick with hbw_available() which I've been using, as it's a better indicator for my particular purpose.  No need to have a getter for numNumaDomains unless there is another use for it.

mppf:
@gbtitus : I havn't looked at this because I trust @dmk42 and @ronawho to handle it. Please let me know if you'd like me in particular to look at some portion.

gbtitus:
I'm holding this for now because in further testing I'm getting NIL check errors in `_ddata_localize()` in some tests.  I suspect I'll need to make some adjustments to that to fix those.  Also this is likely to conflict with the array views PR which is already up, so I don't think this one will go in for a few days.

@mppf , sounds good -- I think @ronawho and @dmk42 have things well in hand.

gbtitus:
After some difficulties arose in full testing yesterday afternoon I chatted a bit with @ronawho and we decided a reasonable response was to make the localization calls from the allocation functions instead of from the module code after the memory was allocated.  This is more like I think we'd envisioned things working anyway, long term.  So I've done one more commit (whose discussion describes the problems I ran into in testing), and would appreciate it if you'd take another look at your convenience.  I won't be trying to merge this until Monday at the earliest, so no rush.  This does pass full testing with locModel=numa, except for some things that didn't work before and still fail in the same way.

gbtitus:
And now I've added another small change, but it's just to the sanity test .prediff file.

runtime/src/chpl-topo.c
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  if ((nodeset = hwloc_bitmap_alloc()) == NULL) {
+    report_error("hwloc_bitmap_alloc()", errno);
+  }
+
+  if (hwloc_get_area_membind_nodeset(topology, p, 1,
+                                     nodeset, &policy, flags)) {
+    report_error("hwloc_get_area_membind_nodeset()", errno);
+  }
```
gbtitus: Is it necessary that getMemLocality() be able to handle a NULL pointer?  Could we simply avoid calling it on those?  (I'm wondering how this situation arises, basically.)

dmk42: The reason I found it was that I needed to free an object, and was testing whether or not it was in MCDRAM to determine which version of free() to call.  When trying to stuff a 1..4 array into a 1..3 array, somehow the code generated a null pointer and asked me to free it, before the error message could come out.  The free() operation is supposed to be a no-op on null pointers, so this is OK from that code's point of view.

You're right about my workaround.  I put in a check for NULL before calling chpl_topo_getMemLocality().  Once it becomes part of the API, though, there could be future instances of passing it a null pointer.  It just seems safest to check.

dmk42: An alternative would be to expand the error message for hwloc_get_area_membind_nodeset() failing, to print the value of the pointer passed to it.  Then it would at least be clear that the reason it failed was that it was passed a null pointer.

gbtitus:
I think I'm ready to go with this now.  Reviewers, feel free to take a look at the last two commits.  And especially @dmk42, would you check if the very last commit does what you wanted in your review feedback?

dmk42:
Looks good, ready to go.

PR #5469, opened 2017-03-01
Upgrade to jemalloc 4.5.0

ronawho:
jemalloc 4.5.0 includes a significant number of bug fixes, most of which are
the result of significantly improved CI testing. It fixes a number of bugs
introduced in the 4.X series. The one most important to us is:

 - "Restructure *CFLAGS configuration, so that CFLAGS behaves typically, and
   EXTRA_CFLAGS provides a way to specify e.g. -Werror during building, but not
   during configuration."

which will allow us to remove an osx+gnu workaround and simplify how we specify
CFLAGS to jemalloc.

I do not expect there to be any performance impacts from this upgrade.

For the full changelog see:
 - https://github.com/jemalloc/jemalloc/releases/4.5.0


ronawho:
@gbtitus mind reviewing? 1st commit upgrades jemalloc-src. 2nd one is the one to review. 

I probably won't merge this until the weekend since I don't want to chance messing master up for the array-views merge

PR #5465, opened 2017-03-01
Open source CHPL_COMM=ugni

ronawho:
CHPL_COMM=ugni is Chapel's communication layer over Cray's uGNI interface.
Historically it has been closed-sourced, but no more!

This comm layer was written and has been almost solely maintained by @gbtitus.
All credit goes to him, I'm just merging it into the public repository on his
behalf.

CHPL_COMM=ugni is a Cray XE/XC only communication layer, it does not work on
any other systems. It typically outperforms gasnet-aries by a significant
margin and is thus our recommended and default communication layer on Crays.

PR #5475, opened 2017-03-02
Get comm=ugni working with pgi

ronawho:
Get ugni working with pgi:
 - switch a struct initializer that pgi couldn't handle
 - use atomic_load instead of just casting an atomic for comm diags
 - throw -c11 to get __thread support
 - just use -lpthread instead of -pthread
 - default to CHPL_NETWORK_ATOMICS=none when CHPL_ATOMICS=locks

This was mostly so that I didn't have to special case the documentation for
pgi, but it resulted in some minor cleanup as well.

With this ugni works with all supported Cray backends, including llvm

### DIFF
--- a/runtime/etc/Makefile.comm-ugni
+++ b/runtime/etc/Makefile.comm-ugni
@@ -29,9 +29,4 @@ ifeq (,$(findstring HUGETLB,$(PE_PRODUCT_LIST)))
 GEN_LFLAGS += -lhugetlbfs
 endif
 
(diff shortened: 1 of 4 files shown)


ronawho:
@gbtitus if you get a second I'm looking for some initial feedback on these changes. It's based on #5465, so ignore the first 5 commits.

gbtitus:
All of those look good to me.  👍

PR #5478, opened 2017-03-02
Implement targetLocales for DefaultRectangularArr

benharsh:


benharsh:
@bradcray : is this what you had in mind?

modules/internal/DefaultRectangular.chpl
```diff
@@ -1769,7 +1769,7 @@ module DefaultRectangular {
     }
 
     proc dsiTargetLocales() {
-      compilerError("targetLocales is unsupported by default domains");
+      return [this.data.locale, ];
```
bradcray: Yes, this is what I was imagining.  I think it needs to be added to the the domain as well (?).

As extra credit, I also ultimately think that we should not require dsiTargetLocales on both array and domain objects, but should support it on domains only and have arrays answer the target locales query by passing it along to their domains...

test/arrays/userAPI/arrayOps2D-targetLocales.good
```diff
@@ -1,2 +1,105 @@
-./arrayAPItest.chpl:7: In function 'testArrayAPI2D':
-./arrayAPItest.chpl:92: error: targetLocales is unsupported by default domains
```
bradcray: Here, I was imagining that we'd actually take this out of the "error cases" category and make it part of the "normal" operation of the test, just to avoid maintenance costs and duplication in the .good files (though this'll mean updating most of the .goods in that directory).

benharsh:
How does that look? There aren't any tests for DefaultRectangularDom.dsiTargetLocales, but I figure that can wait until domain-views (or whatever).

bradcray:
Awesome, thanks!  Your thinking is the same as mine:  As I added domain views, I was planning on creating a similar userAPI test for them which would exercise this.

PR #5480, opened 2017-03-02
Remove unnecessary 'no copy return' pragma

benharsh:
It turns out that this 'no copy return' pragma is unnecessary in this case.


benharsh:
@bradcray , @mppf : Michael was correct in his suspicion that this pragma was unnecessary. This branch has passed a full local paratest.

PR #5481, opened 2017-03-02
Skip this test if we will link with libnuma.

gbtitus:
At least so far, libnuma is only available in the form of a shared
object for dynamic linking.  So, the compopts that direct this test
to do static linking will certainly fail if libnuma is referenced.
Therefore, skip the test if we're using it.


gbtitus:
@ronawho , would you mind reviewing?  Should be simple.

ronawho:
Sorry, missed this before. Looks good to me

PR #5483, opened 2017-03-02
Use Time.sleep so that these tests play nice with qthreads

benharsh:
Prior to #5446 , qthreads was intercepting the ``usleep`` call. When that PR went in we observed a "regression" in nightly testing for these tests. Using ``Time.sleep`` is the proper way to sleep and play nice with the tasking layer.

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -844,3 +844,33 @@ timeVectorArray:
     - increased array-as-vec problem size (#3422)
   05/28/16:
     - upgrade jemalloc to 4.2.0 (#3919)
+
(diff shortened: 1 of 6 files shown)


benharsh:
@chapel-lang/perf-team : An odd fix for an odd test. This resolve the "regression" discussed earlier. Thanks to @ronawho for the solution.

bradcray:
Would you be willing to throw an annotation onto these graphs in this PR as well?

benharsh:
Now that there are annotations, I think I'll wait to merge until tomorrow to avoid screwing up tonight's perf testing for ArrayViews.

PR #5485, opened 2017-03-03
Adjust STREAMS again for RandomStream interface change

mppf:
Adjusts .chpl files in test/studies/hpcc/STREAMS.

Use by-name argument passing to enable release-over-release testing.

Verified no compile failures when testing this directory with -performance under 1.14.
Verified directory passes local testing and local -performance testing.
Trivial, not reviewed.

### DIFF
--- a/test/studies/hpcc/STREAMS/bradc/stream-block1d-dom.chpl
+++ b/test/studies/hpcc/STREAMS/bradc/stream-block1d-dom.chpl
@@ -61,7 +61,7 @@ proc printConfiguration() {
 
 
 proc initVectors(B, C) {
-  var randlist = new NPBRandomStream(real, seed);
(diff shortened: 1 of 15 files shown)


bradcray:
Hi @mppf: Thanks for jumping onto this.  

Unfortunately, I haven't compiled a short list of tests that failed for this reason on release-over-release testing, but the easiest way to do this would be to search chapel-perf/chapcs/releaseOverRelease/LOG.gz for errors similar to:

````
unresolved call 'NPBRandomStream.init(type real(64), int(64))'
````

...searching from the bottom of the file upwards.  (If you're not familiar with the release-over-release logs, they run the tests through all versions of the compiler from 1.5 onward, so I'd limit the search to failures during the 1.14 testing since earlier ones are likely to be repeats or caused by other such changes).

PR #2127, opened 2015-06-29
Replace fast.suppress with individual .suppressif

CoryMcCartan:
Removes fast.suppress, which suppressed certain tests when the --fast compiler option is specified, and replaces it with a .suppressif file for each test.

PR #5493, opened 2017-03-03
Fix --fast failure by always throwing --bounds-checks for this test

bradcray:
This test was written to do bounds checks, so here I'm throwing the
flag to make sure they're enabled even if testing in configurations
like --fast that would normally disable them.

### DIFF
--- a/test/distributions/bradc/blkCyc/blkCycSliceOOB.compopts
+++ b/test/distributions/bradc/blkCyc/blkCycSliceOOB.compopts
@@ -1,0 +1,1 @@
+--bounds-checks


bradcray:
@benharsh: This is trivial, but would you mind giving it a quick look to make sure I got things right?

benharsh:
Looks good to me 👍

bradcray:
offline, @ronawho asked:

> For other cases, we've done suppressifs for tests that are explicitly
> checking bounds checking: https://github.com/chapel-lang/chapel/pull/2127
> 
> Any reason to do something different here?

I thought about it, but this seemed simpler.  I started out writing the suppressif 
(actually a skipif) first, but then started thinking about cases like:

* What if compopts contains --no-checks or --no-bounds-checks but not --fast?  (test would fail but suppressif wouldn't help us with it)
* What if it's throwing both --fast and --bounds-checks? (test would pass, but suppressif would expect it not to)
* ...and from there you can imagine any one of a number of complex command line or environment settings that would make the test pass or fail...

Writing a skip/suppressif that handled all these cases seemed daunting, so I was about to fall back on "well, we know we only run nightly testing with --fast, so only developers who explicitly test those other cases would suffer, so I'll just do the simple thing."  But then I thought "What I really want this test to do is to check that the OOB message appears as expected when we're bounds checking.  So given that that's an assumption of the test, why not just ensure that it's true by throwing the flag?"

I guess one downside to my approach is that it doesn't test that we do bounds checking on by default. But at present we know that to be true, and arguably it should be a test other than this one that locks that in...

Thoughts?

ronawho:
I think we've (we the group, not me and you explicitly) had this conversation a bunch of times, and we always go back and forth on it.

From that PR I mentioned

```
# --fast sets --no-checks and this test is explicitly checking that one
# of our runtime checks is working correctly. There's been debate as to
# whether tests like this should be suppressed or skipped. For now we
# suppress with the argument that we're testing that --fast actually
# turns our checks off. Note that we don't just turn checks on in the
# comopts since we wouldn't be testing the default behavior, and
# wouldn't be able to verify that checks are enabled by default. 
```

IMO we should just stick with that philosophy rather than change our minds on a case-by-case basis. You're right that this will fail if a dev tests with --no-checks or --no-bounds-checks, but that's true for all the other --fast suppressifs and in general for so many of our skipifs/suppressifs (e.g. skipifs for --baseline when it's really a copy prop or inlining bug)

And to be clear, I don't have a strong preference here. I liked the .suppressif as a policy for cases like this so that we wouldn't have to think/talk about something that doesn't really matter :)

ronawho:
And I didn't mean for my response to come off as snarky or anything. I'm ok leaving things as they are, I was just curious if there was a specific reason for not using a .supressif

bradcray:
(See, I told you I never read comments! -- I read the "code" in that PR but not the paragraph that preceded it :) )

My inclination would be to let the test author do what they think is best for a given test and then beat up on them if that decision causes undue hardship for others and wasn't rationalized well.  For example, your rationale above is oriented toward a "let's make sure --fast works as expected" perspective, whereas I'd argue that this test is oriented more toward "let's make sure OOB works as expected."  Arguably, a mix of both kinds of tests will cover both bases (fast turns off checks; when checks are on, they work).

In this case, I personally don't think there needs to be a global policy that we hold ourselves to uniformly for the sake of consistency.  Particularly since neither of us seems to think it really matters much for cases like this and I only have so many free neurons to remember such policies (e.g., I have no recollection of having established a policy even though I very well may have been involved in the original discussion).

ronawho:
ok, sounds good to me

PR #5495, opened 2017-03-04
Retire futures that started passing with array views

bradcray:
While I had the foresight to run futures testing prior to merging array
views, I forgot to look for newly passing futures... sigh.

* hplx was dying due to some simultaneous array views on an array which
  might be due to the old closed-form approach potentially modifying
  aspects of the array descriptor (?).  In any case, it seems to be
  working fairly regularly now.  Its .future said to remove the hpl.skipif
  in the robustness suite when it started working, so I've done that too.

* 4-consume-return-slice-local.* now works as expected, which I think
  makes sense given the new "copy out" semantics for views, so
  retiring its .future and .bad


bradcray:
@benharsh: Any interest in giving this a quick sanity-check look?

Issue #5488, opened 2017-03-03
Deprecate support for '=>' in passing arrays to constructors


bradcray:
For some time, Chapel has supported the ability to have an array field alias another array by using the '=>' operator in a constructor.  This issue proposes deprecating this feature for 1.15 and removing it for 1.16.  For example:

````chapel
class C {
  var X: [1..3] real;
}

var A: [1..3] real = ...;
var myC = new C(X=A);            // gets its own 3-element array 'X' that copies 'A'
var myAliasC = new C(X=>A);  // this object's 'X' field will alias 'A'
````

This feature has been problematic in at least a few ways:

1) It's a reasonably fragile feature.  The array passed in must be of precisely the same static type as the field array in the class, which makes some cases that seem like they should work not (e.g., if the actual is stridable but has stride 1 and the field is non-stridable, the alias won't work).
2) It adds a fair amount of complex special-case code to the compiler that several of us have run afoul of and not been able to fix or work around very easily.
3) It feels like a very "special"/"unique" feature that would be better handled by supporting 'ref' fields within classes/records or perhaps even fields that could be param-switched between 'ref' and 'var' in a generic class.

The original motivation for this feature was to support closed-form array views using distributed domain maps like Block.  In more detail:

* Block-distributed arrays support a global descriptor that stores no data.  It points to a per-locale local descriptor that stores a local array field representing that locale's portion of the global index set (similar to C above).
* When a Block-distributed array is sliced, in closed-form, we created a new global descriptor that pointed to new local descriptors with their own array fields.  But to get the aliasing semantics of slicing, we had to set these alias fields up to point to the original local descriptors' local arrays.  Thus, we introduced the '=>' operator in the constructor context to support this.
* While this approach "worked" it had similar fragilities in the Block distribution that made it difficult to change the DefaultRectangular class much for reasons similar to the fragilities above:  a small change to make it more generic, say, resulted in cases where the local slice Block descriptors could no longer alias the originals because some aspect of the type had shifted slightly.

Happily, this need for the '=>' goes away with the array views work, where all aliasing of arrays happens by referring to the array through its array class descriptor rather than trying to alias the ddata field itself.  For this reason, I'm proposing that we deprecate this feature in order to move away from the downsides listed above and remove the special-case code from the compiler (after the upcoming release).

[...]

PR #5497, opened 2017-03-04
Deprecate array aliasing in constructor calls

bradcray:
As mentioned in PR #5338, with array views, we no longer require supporting '=>' in constructor calls, which used to be required for the closed-form by distributions like Block.  See issue #5488 for more details.

I was planning on deprecating support for '=>' over the next week or so, but in testing array views had not realized that the one main test still exercising the feature, test/arrays/deitz/part4/test_array_alias_field.chpl, wasn't actually working correctly anymore.  Specifically, when freeing the class objects, the array fields were causing the original array to be freed even though it had not yet left scope.  This showed up in valgrind testing as well as (by chance?) the 32-bit platforms.

It turns out that the pointless slicing that I'd removed from this test had been added recently in commit a9a5e95 as part of the array memory management fixes and has the same valgrind issue at that time if left unchanged.  So the valgrind error is not a new problem, it's just one I unearthed by changing the test as part of the array views work (necessary because now that slices are a different type, the '=>' operator no longer permitted them to be passed as an alias to the field due to type mismatches -- again, see issue #5448.

For that reason, I'm deprecating '=>' earlier by adding a warning when it's used that it's deprecated and also indicating that previous uses of it are unlikely to work (either because they were aliases of views which will now not work because of the type difference, or because they were normal aliases in which they could run afoul of this use-after-free problem introduced in a9a5e95.

I moved that last test into test/deprecated as a placeholder for completely removing the warning.  In order to keep it working, I commented out its calls to 'free' (sorry "zero memory leak" fans -- but they're small).  I also found a future that used the feature and removed it.


daviditen:
I'd prefer to break the string for the warning across multiple lines in so it fits in an 80 character buffer.  Something like:

```
  USR_WARN(expr, "support for '=>' in constructor argument lists is deprecated"
                 " as of chpl version 1.15 and is unlikely to work as well as"
                 " it used to.  If you rely on this feature, please let the"
                 " Chapel team know.");
```

Otherwise it looks good to me.

bradcray:
Hmm, good point.  I've obviously been living in Chapel too long...  Which makes me think, maybe we should open up an "easy" issue asking someone to add support for multi-line strings to Chapel.  :)

bradcray:
Filed the issue marked it "easy" for GSoC-ers

daviditen:
I don't think the current Chapel way of breaking up strings across lines is particularly bad.  If the same call were written in Chapel, I would have suggested:

```
  USR_WARN(expr, "support for '=>' in constructor argument lists is" +
                 " deprecated as of chpl version 1.15 and is unlikely to" +
                 " work as well as it used to.  If you rely on this feature," +
                 " please let the Chapel team know.");
```

Sure, it doesn't become a single string until function resolution vs. in the lexer, but I don't think that's a big deal.

I don't have a problem with adding the C-style way, I just don't think we're in a bad place without it.

bradcray:
I agree that the current approach isn't awful, and almost mentioned it as a workaround in filing the issue, but it also seems like it carries needless complication and overhead through multiple passes of compilation while also requiring an unfamiliar approach for the C crowd whose typical approach would be easy and unambiguous to support.

That said, whoever takes this on should probably try to ensure there's consensus before getting too far into it.

PR #5056, opened 2016-12-16
Split the LCALS .graph file into six graph files

daviditen:
The LCALS .graph file contained setup for six different graphs.  Split them
apart into six files at @bradcray's request.

Tested with: start_test -performance LCALSMain.chpl
New data points were added to the right .dat file and graphs were generated.

PR #5499, opened 2017-03-04
Fix lcals annotations

ronawho:
Lcals annotations were applied to LCALSMain, but that was split out into 6
graphs with #5056.

This uses YAML anchors and the merge-key type to copy the annotations from one
graph to the other five


ronawho:
@chapel-lang/perf-team I think this yaml feature is super cool and I didn't know about it before. 

http://stackoverflow.com/questions/9254178/is-there-yaml-syntax-for-sharing-part-of-a-list-or-map has a better explanation than I can provide for what it does.

PR #5502, opened 2017-03-04
Update .good file to match recent deprecation of '=>' 

noakesmichael:
I happened to notice that arrays/deitz/part7/test_illegal_named_alias_argument.chpl
started failing due to Brad's recent update.  This updates the .good file to resolve this.


bradcray:
Shoot, thanks for the catch Mike.  I thought I remembered there being a case in part7 that I needed to update, but then (a) in searching for it I was grepping for tests that use 'new' and '=>' on the same line (and since this test was testing an error case, it didn't), and (b) the number "2" tests to update was stuck in my mind, so when I got the first arrays/deitz case and diten's future, I thought I was done.  In retrospect, this was the 2nd test (since diten's test was a future that I didn't run) and my grep was naive.

I'm actually going to just remove this test as originally intended since we have a placeholder for the deprecation warning and keeping this test alive doesn't add value.

PR #5506, opened 2017-03-04
Fixing broken Quickstart.rst link

TestSubjector:
Hello.
Clicking on the Quickstart link in README.rst gives a 404 error, as the Quickstart.rst file resides in docs/usingchapel directory (and not in the top level directory). 
PR to fix the same.


ben-albrecht:
Thanks for catching this and providing a fix!

PR #5508, opened 2017-03-04
Update to 2nd of implementation of return by ref

noakesmichael:
Summary:

Update callDestructors.cpp to modify the code that converts procedures that
return values with record-type to be an equivalent procedure that returns the
value, in principle, by out-intent.

This change altered AST in a way that impacted deadStringLiteralElimination()
and so this PR includes a second commit that a) makes this code a little more
transparent b) updates the code to handle the new form of AST.  This update
includes a pair of largely "ad hoc" test that are intended to catch situations in
which changes to previous passes impact this cleanup

Passes paratest for single-locale, baseline, and gasnet.

Details/background:

A recent change to normalization of DefExpr caused multiple failures for
many no-local configurations.  One of the factors for these errors is an
interaction between code and callDestructors.cpp and code in parallel.cpp.

The former code converts procedures that return values with record type to an
equivalent procedure that, in principle, returns the value for out intent

The latter code replicates const module level variables for multi-locale applications.

Issues with the representation of reference variables and def-use analysis lead
to problems that were being masked by choices made in normalization.

BenH proposed a change to the return-by-value transformation that he recognized
would better match the current def-use analysis and would be likely to help with
replicating const module level variables and might also help with wide-ref
issues he had investigated.

This PR implements Ben's proposal.  A subsequent commit was required to handle
an interaction between this transformation and deadStringLiteralElimination.

This PR passes the expected paratests.  Additionally I developed a separate branch
which layers the problematic normalization change on top of this PR.  As hoped this
resolved the regressions that the normalization change would have on master.

### DIFF
--- a/compiler/optimizations/deadCodeElimination.cpp
+++ b/compiler/optimizations/deadCodeElimination.cpp
@@ -305,6 +305,7 @@ void deadCodeElimination(FnSymbol* fn) {
 *      other portions of the compiler.                                        *
 *                                                                             *
 * The hardest part is determining when this code needs to be revisited.       *
+* 2017/03/04: Perform an "ad hoc" sanity check for now.                       *
 *                                                                             *
 ************************************** | *************************************/
 
@@ -329,14 +330,23 @@ static void deadStringLiteralElimination() {
     }
   }
 
-  // Confirm that the analysis found some dead literals
-  // Ignore this test for --minimal-modules
-  if (numStmt > 500) {
-    INT_ASSERT((1.0 * numDeadLiteral) / numStmt > 0.2);
+  //
+  // Noakes 2017/03/04
+  //
+  // There is not currently a principled way to determine if other
+  // paseses have changed in a way that would confuse this pass.
+  //
+  // A quick review of a portion of test/release/examles shows that
+  // this pass removes 85 - 95% of the string literals.  Signal an
+  // error if this pass doesn't reclaim at least 10% of all string
+  // literals unless this is minimal modules
+  //
+  if (fMinimalModules == false) {
+    INT_ASSERT((1.0 * numDeadLiteral) / numStmt > 0.10);
   }
 }
 
-// 2017/03/03: All literals have 1 def. Dead literals have 0 uses.
+// Noakes 2017/03/04: All literals have 1 def. Dead literals have 0 uses.
 static bool isDeadStringLiteral(VarSymbol* string) {
   bool retval = false;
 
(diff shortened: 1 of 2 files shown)


compiler/optimizations/deadCodeElimination.cpp
```diff
@@ -274,77 +274,139 @@ void deadCodeElimination(FnSymbol* fn) {
...
+          numDeadLiteral = numDeadLiteral + 1;
+        }
       }
     }
   }
+
+  // Confirm that the analysis found some dead literals
+  // Ignore this test for --minimal-modules
+  if (numStmt > 500) {
+    INT_ASSERT((1.0 * numDeadLiteral) / numStmt > 0.2);
```
benharsh: Shouldn't this test if the --minimal-modules flag is set?

Also, this INT_ASSERT seems a little arbitrary to me. If you feel it's useful, can you insert a comment describing how you arrived at this check? Also, based on the comment "Confirm that the analysis found some dead literals" I was expecting to see something like ```INT_ASSERT(numDeadLiteral >= 1);```

benharsh:
Otherwise this looks fine to me.

noakesmichael:
On Mar 4, 2017, at 1:35 PM, Ben Harshbarger <notifications@github.com<mailto:notifications@github.com>> wrote:


@benharsh commented on this pull request.

In compiler/optimizations/deadCodeElimination.cpp<https://github.com/chapel-lang/chapel/pull/5508#discussion_r104297795>:

>        }
     }
   }
+
+  // Confirm that the analysis found some dead literals
+  // Ignore this test for --minimal-modules
+  if (numStmt > 500) {
+    INT_ASSERT((1.0 * numDeadLiteral) / numStmt > 0.2);


Shouldn't this test if the --minimal-modules flag is set?


Duoh.  That was lame.


Also, this INT_ASSERT seems a little arbitrary to me. If you feel it's useful, can you insert a comment describing how you arrived at this check? Also, based on the comment "Confirm that the analysis found some dead literals" I was expecting to see something like INT_ASSERT(numDeadLiteral >= 1);


You’re right of course.  I was worried that a threshold of 1 was too limited. and so
this means that at least 20% of literals are swept.  The test I was playing with reaped
something like 90% of the literals.

I’ll take a crack at this.

Thanks,

Mike

compiler/optimizations/deadCodeElimination.cpp
ronawho: Instead of having the check in the compiler like this, what do you think about having a `--report-dead-string-literals` flag (like `--report-dead-modules`) and just creating a test like optimizations/deadCodeElimination/elliot/countDeadModules. The number of dead string literals will probably change pretty frequently, so it should probably have some sort of threshold.

Alternatively not eliminating dead strings really blows up the generated code size so we could just rely on the generated code size compiler graph to catch this after the fact. 

And also, sorry. This is definetly not code that I'm proud of having written.

PR #5511, opened 2017-03-04
Make LLVM back-end support multiple/space-separated ccflags (Fixes #5510)

### LINKED ISSUE #5510: Clang arguments passed in wrong way when using --ccflags
There is a problem associated with passing options to backend C compiler when it is clang (in case of gcc this works fine). Clang receives options as one string being concatenation of all options and doesn't parse this.

chpl --ccflags "-mllvm -use-cfl-aa" --llvm test.chpl
and
chpl --ccflags -mllvm --ccflags -use-cfl-aa --llvm test.chpl

Give:

error: unknown argument: '-mllvm -use-cfl-aa'

Using CHPL_CC_FLAGS also doesn't work.

coodie:
PR contributed by @coodie - thanks!

This fixes #5510. Not a big deal, simply split passed arguments by spaces.

Reviewed by @mppf.

Tested:
- hellos pass with CHPL_LLVM=none
- test/release/examples/primers and test/extern/ferguson with CHPL_LLVM=llvm
- test/release/examples/primers and test/extern/ferguson with CHPL_LLVM=llvm and --llvm

### DIFF
--- a/compiler/util/clangUtil.cpp
+++ b/compiler/util/clangUtil.cpp
@@ -1170,7 +1170,11 @@ void runClang(const char* just_parse_filename) {
   }
   clangCCArgs.push_back(std::string("-I") + getIntermediateDirName());
 
-  clangCCArgs.push_back(ccflags);
+  //split ccflags by spaces
+  std::stringstream ccArgsStream(ccflags);
+  std::string ccArg;
+  while(ccArgsStream >> ccArg)
+      clangCCArgs.push_back(ccArg);
 
   clangCCArgs.push_back("-pthread");
 
--- a/compiler/util/clangUtil.cpp
+++ b/compiler/util/clangUtil.cpp
@@ -28,6 +28,7 @@
 #include <cctype>
 #include <cstring>
 #include <cstdio>
+#include <sstream>
 
 #include "astutil.h"
 #include "driver.h"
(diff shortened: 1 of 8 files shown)


bradcray:
@mppf: You are the most obvious reviewer for this PR, if you have time to look at it.

@coodie: One other thing that this PR should include before being merged is a test that will lock in the working behavior to make sure we don't backslide in the future.  Let us know if you need pointers into the Chapel testing system for help on how to do this.

coodie:
I like how you put effort to testing even little changes like this :), that's really proffesional.

I have added locking tests. Please take a look.

bradcray:
Hi @coodie: If you were to check in this new test separately/first (say at the same time you filed the issue), you would need to include the .future and .bad files as you've prepared them here.  Since in this case, they're most likely to be merged at the same time as the fix itself, you can (and should) drop those files from the PR because otherwise our test system will just tell us that the future is passing (an indication that its .future and .bad files should be removed).  Put another way, since you'll be filing the fix, it ceases to be a test that will work in the "future" and should work now instead.

mppf:
@bradcray, @coodie - this looks fine to me. We just need somebody to shepherd it through testing.

bradcray:
Do you think it requires a full test run, or just for those tests that use ccflags and the llvm back-end?

mppf:
I think testing test/release/examples/primers and test/extern/ferguson with CHPL_LLVM=llvm with and without --llvm would be sufficient.

mppf:
I plan to merge this PR on Monday.

PR #5514, opened 2017-03-05
Implement the ArrayView index cache for NUMA

benharsh:
Once again this was accomplished primarily by copying the
old dsiSlice/RankChange/Reindex methods.

Testing:
- [x] local-flat
- [x] gasnet-flat
- [x] local-numa
- [x] gasnet-numa


benharsh:
@gbtitus : Could you review? This basically copies the old dsiSlice/RankChange/Reindex logic from the old-arrays-world. It also appears to resolve the LCALS performance regressions.

gbtitus:
This looks good.  I was afraid that in order to resolve the LCALS performance regressions we'd have to hoist the addressing messiness of multi-ddata into the array views code, so I'm a bit sad to be right about that.  But it's nice to have the performance back!

benharsh:
@gbtitus : Care to look over this any more? I've added some commits since your last comment. I plan on doing a final local paratest before merging (probably tomorrow or Thursday to keep noise down).

gbtitus:
I don't see any problems in the new commits -- good to go!

PR #5516, opened 2017-03-06
Harden the make check script and provide more output for errors.

awallace-cray:
Adds more error checking and verbose diagnostics to
util/test/checkChplInstall, the shell script that implements
"make check".
    
* Set env variable CHPL_CHECK_DEBUG=1 everywhere make check is
  called from the nightly build scripts, to activate more verbose
  error diagnostics.
* The new verbose diagnostic messages are marked [Debug], and include:
  * Msgs at beginning and end of the make check script, including
    the exit status returned by the script to "make".
  * Exit status returned to the script from both the Chpl compilation
    and the test run.
  * Full copies of reference and actual test outputs, not just "diff".
* Explicitly check ("cd" to) all important file system locations,
  to guard against unexpected file system problems.

### DIFF
--- a/util/test/checkChplInstall
+++ b/util/test/checkChplInstall
@@ -100,12 +100,12 @@ fi
 
 # Tmp directory for compile output
 TMP_TEST_DIR="$(mktemp -d ${CHPL_DIR}/chapel-test-XXXXX)"
-if [ -z "$TMP_TEST_DIR" ] ; then
+if [ -z "${TMP_TEST_DIR}" ] ; then
     log_error "failed mktemp -d ${CHPL_DIR}/chapel-test-XXXXX"
     exit 1
 fi
 log_info "Temporary test job directory: ${TMP_TEST_DIR}"
-( cd $TMP_TEST_DIR ) || { log_error "failed cd $TMP_TEST_DIR" ; exit 1 ; }
+( cd ${TMP_TEST_DIR} ) || { log_error "failed cd ${TMP_TEST_DIR}" ; exit 1 ; }
 
 # Cleanup the tmp directory whenever exit is invoked
 function cleanup_tmp_dir()
@@ -208,7 +208,7 @@ TEST_EXEC_OUT=${TEST_JOB}.exec.out
 
 # Run compiled binary with parsed execution options
 (
-    cd ${TMP_TEST_DIR} || { log_error "failed cd $TMP_TEST_DIR" ; exit 1 ; }
+    cd ${TMP_TEST_DIR} || { log_error "failed cd ${TMP_TEST_DIR}" ; exit 1 ; }
 
     EXEC_OPTS="$(cat ${TEST_DIR}/${TEST_JOB}.execopts)"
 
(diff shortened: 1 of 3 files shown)


util/test/checkChplInstall
```diff
@@ -95,10 +96,16 @@ else
...
+( cd "${CHPL_DIR}" ) || { log_error "failed cd $CHPL_DIR" ; exit 1 ; }
 
 # Tmp directory for compile output
 TMP_TEST_DIR="$(mktemp -d ${CHPL_DIR}/chapel-test-XXXXX)"
+if [ -z "$TMP_TEST_DIR" ] ; then
+    log_error "failed mktemp -d ${CHPL_DIR}/chapel-test-XXXXX"
+    exit 1
+fi
 log_info "Temporary test job directory: ${TMP_TEST_DIR}"
+( cd $TMP_TEST_DIR ) || { log_error "failed cd $TMP_TEST_DIR" ; exit 1 ; }
```
ben-albrecht: For consistency,  this should be `"${TMP_TEST_DIR}"`

util/test/checkChplInstall
```diff
@@ -95,10 +96,16 @@ else
     log_info "${CHPL_DIR} already exists. Using it."
     CHPL_DIR_REMOVE=0
 fi
+( cd "${CHPL_DIR}" ) || { log_error "failed cd $CHPL_DIR" ; exit 1 ; }
 
 # Tmp directory for compile output
 TMP_TEST_DIR="$(mktemp -d ${CHPL_DIR}/chapel-test-XXXXX)"
+if [ -z "$TMP_TEST_DIR" ] ; then
```
ben-albrecht: similarly, `"${TMP_TEST_DIR}"`

util/test/checkChplInstall
```diff
@@ -189,40 +200,63 @@ else
...
 # Run compiled binary with parsed execution options
 (
-    cd ${TMP_TEST_DIR}
+    cd ${TMP_TEST_DIR} || { log_error "failed cd $TMP_TEST_DIR" ; exit 1 ; }
 
     EXEC_OPTS="$(cat ${TEST_DIR}/${TEST_JOB}.execopts)"
 
     log_info "Running test job."
-    ./${TEST_JOB} -nl${NUMLOCALES} ${EXEC_OPTS} | sort > ${TEST_EXEC_OUT} 2>&1
+    ./${TEST_JOB} -nl${NUMLOCALES} ${EXEC_OPTS} | sort > ${TEST_EXEC_OUT}
```
ben-albrecht: Why drop stderr redirection? Is this more consistent with `sub_test`?

awallace-cray: Yes. Even if not, the only thing on stderr right there would be errors from "sort" itself. 
A better question is, why isn't the stderr from running the test job redirected into THAT stdout stream? It probably should be. But, that would NOT be consistent with sub_test. And, I'm not about to change sub_test.

ben-albrecht: OK makes sense

awallace-cray:
How's that?

PR #5519, opened 2017-03-06
Introduce PRIM_INIT_VAR

noakesmichael:
Summary:

This PR moves a small amount of resolution business logic from normalize.cpp
to functionResolution.cpp as a step towards improving the final version of
non-generic record initialization.

This first PR alters the AST emitted by normalize() for common uses of DefExpr
but has no impact on the AST emitted by resolve().  Hence there should be no
impact on application behavior.

Paratested with single-locale and gasnet on linux64.

Additional discussion.

Consider

  var x = buildMyRecord();

Before this PR, normalization converted this to (approximately)

  var tmp = buildMyRecord();
  CallExpr(PRIM_MOVE, x, CallExpr("initCopy", tmp));

This poses a challenge to initializing records in the desired manner.
This PR introduces a new primitive PRIM_INIT_VAR to avoid injecting the
initCopy prematurely.  Normalization becomes

  var tmp = buildMyRecord();
  CallExpr(PRIM_INIT_VAR, x, tmp);

and delays handling tmp until the type can be inspected i.e. in that
portion of resolution that is currently implemented in functionResolution.cpp.

However this first PR does not actually leverage the type information to emit
better AST.  This simply relocates existing business logic.

### DIFF
--- a/compiler/AST/primitive.cpp
+++ b/compiler/AST/primitive.cpp
@@ -449,13 +449,15 @@ void
 initPrimitive() {
   primitives[PRIM_UNKNOWN] = NULL;
 
-
   prim_def(PRIM_ACTUALS_LIST, "actuals list", returnInfoVoid);
   prim_def(PRIM_NOOP, "noop", returnInfoVoid);
   prim_def(PRIM_MOVE, "move", returnInfoVoid, false, true);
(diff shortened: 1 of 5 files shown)


lydia-duncan:
This looks reasonable to me!

PR #5521, opened 2017-03-06
Fix denormalize bug with PRIM_ARRAY_SHIFT_BASE_POINTER

benharsh:
Denormalize was modifying a PRIM_ARRAY_SHIFT_BASE_POINTER call such that
the return expr (the first expr in the alist) was an rvalue. More
specifically, it would transform AST like this:

  var ret = (cast ddata nil)
  shift_base_pointer(ret, ...)

into this:

  shift_base_pointer((cast ddata nil), ...)

Long-term, a better solution will be to have PRIM_ARRAY_SHIFT_BASE_POINTER return something instead of setting its first argument. We could then rely on PRIM_MOVE's logic.

Testing:
- [x] full local

### DIFF
--- a/compiler/passes/denormalize.cpp
+++ b/compiler/passes/denormalize.cpp
@@ -319,11 +319,16 @@ bool isDenormalizable(Symbol* sym,
                   // FnSymbol expects to return one symbol, so it's easier to
                   // just not denormalize the returned symbol.
                   //
-                  // shift-base-pointer: we do not want to denormalize this:
+                  // PRIM_ARRAY_SHIFT_BASE_POINTER sets its first argument, so
+                  // we should not denormalize if 'se' is the first actual in
+                  // case of AST like this:
                   //   var ret = (cast ddata(...) nil)
                   //   shift_base_pointer(ret, ...)
-                  // into this:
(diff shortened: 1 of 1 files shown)


mppf:
@benharsh - I don't have any problem with this PR itself, but I wonder if we should generally try to limit the number of primitives that have side effects? I.e. change PRIM_ARRAY_SHIFT_BASE_POINTER to a PRIM_MOVE + a new primitive that returns the new base pointer?

benharsh:
I think that's probably the right long-term fix. For now though I'm hesitant to make a change like that close to the release, and I also have a number of other nontrivial things I want to get in first.

I'll take this as a TODO for the next cycle.

mppf:
A related question is if PRIM_ARRAY_SHIFT_BASE_POINTER counts as a "def" of its argument that it changes. Either way, it can be a TODO.

mppf:
@benharsh - I've reviewed this, it's OK. Just please  adjust the new comment to indicate that PRIM_ARRAY_SHIFT_BASE_POINTER sets its first argument. Then please list the TODO we agreed upon in the PR message as the preferred solution.

PR #5524, opened 2017-03-06
Array view annotations

bradcray:
Add annotations for recent array views work:

* merge of array views
  - included separate 16-node entry for hpl.hpcc2012, rewritten as part of array views work
* merge of wide pointer optimization for arrays resulting from array views work

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -221,6 +221,10 @@ all:
   02/19/17:
     - text: Turned hyperthreading on for our testing machine
       config: chapcs
+  03/03/17:
+    - Array views (#5338)
+  03/05/17:
+    - Wide-pointer optimization for arrays (#5487)
 
 
 AllCompTime:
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho or @benharsh : Would one of you give this a quick review?  I've checked to make sure that it works for normal perf testing results, but don't feel confident about my 16-node annotation (nor am I sure how to test it easily).  Perhaps I'm overdoing it in trying to add that annotation, but I wanted to remember that the test got worse less due to array views (I think) and more because we rewrote aspects of the test.

ronawho:
Looks fine to me

ronawho:
I regenerated the 16 node graphs just to make sure my commits didn't screw anything up. Looks like the hpl.hpcc2012 annotation is missing. 

Think you need to specify the full basename (hpcc2012.ml-perf and hpcc2012.ml-time) instead of just trying to use hpl.hpcc2012. And since this is a multi-locale only graph, you can drop the config from it.

ronawho:
And Bens wide pointer patch went in on the 5th, but the annotation date should be for the 6th.

ronawho:
sorry, should have looked more carefully the first time

PR #5191, opened 2017-01-19
Add ML perf files

benharsh:
I think the main difference from the old approach is that SSCA2 size22 is now under studies/ssca2/bharshbarg, as to not clutter the release directory  even more.

PR #5525, opened 2017-03-06
Convert xc- annotations to ml- annotations

ronawho:
PR #5191 brought our multi-locale overlay into the public repo and renamed xc-*
files to ml-*. We missed updating the annotations at the time, so this makes
those changes.


ronawho:
@benharsh quick sanity check?

benharsh:
Seems sufficiently sane to me 👍

PR #5528, opened 2017-03-06
Update and fix multi-locale isx annotations

ronawho:
Remove annotations for dead graphs, add annotations for new graphs, and add an
annotation for isx problem size increase


ronawho:
@benharsh look ok?

benharsh:
Yes, thank you! 👍  👍

PR #5529, opened 2017-03-06
A new benchmark using primality test

krishnakeshav:
An implementation of a **prime number generator** using **primality test**. The performance has been better compared to it's implementation in **C++** .


test/prime generator/prime.chpl
```diff
@@ -0,0 +1,46 @@
+/* Primality Test */
```
lydia-duncan: The Wikipedia entry with the pseudocode for this algorithm seems to think it needs more citations than they currently provide, so I would suggest adding a comment with where you got the algorithm from, just to be nice :)

test/prime generator/prime.chpl
```diff
@@ -0,0 +1,46 @@
...
+
+use Time;
+proc isPrime(n:int): bool {
```
lydia-duncan: In case you missed it, you don't have to declare argument types or return types for functions in Chapel if you don't want to.  You are also under no obligation to remove them, either :)

test/prime generator/prime.chpl
```diff
@@ -0,0 +1,46 @@
...
+			if(isPrime(i) == true) then
+				writeln(i);
+		}
+		t+=1;
+		totalTime = totalTime + T.elapsed();
+		T.stop();
+	}
+
+	//writeln("Total Time elapsed ",totalTime);
+	writeln("End of program reached");
```
lydia-duncan: You mentioned that this performed better than a C++ implementation on your machine.  Would you like to include a sibling C/C++ implementation?  The "time_iterate.chpl" performance test I pointed you at earlier demonstrates using C code as part of a performance test so that both timings are present in our generated graphs; that's probably the simplest strategy within our testing framework.

test/prime generator/prime.stdin
```diff
@@ -0,0 +1,6 @@
+2
+2
+10
+5
+50
```
lydia-duncan: You might want to perform a run with an even larger set of primes to output and verify against, but that's up to you.

test/prime generator/prime.chpl
lydia-duncan: I believe the underscore is important to prevent at least one of our test systems from getting confused about the pathname (cygwin, iirc), but I could be misremembering.  Please use either an underscore or camelCase naming for the directory

test/patterns/prime generator/prime.perfcompopts
```diff
@@ -0,0 +1 @@
+-sexecTime=true
```
lydia-duncan: This and the next file should end with a newline.

Also, while this will generate performance timings for us nightly, it doesn't include any way for us to track what that looks like visually.  For that, you'll need a .graph file (and to add that .graph file to test/GRAPHFILES).  The details on how to do that should also be in the TestSystem document.

test/patterns/prime generator/prime.stdin
```diff
@@ -3,4 +3,3 @@
 10
 5
 50
```
lydia-duncan: Hmm.  After doing a performance run, I have changed my mind.  Would you mind using a larger problem size to verify that this is working?  This completes in 1.5e-5 on my machine and that is just too fast for us to get useful timing information out of it (as at that speed, most of the overhead will be task start up).

test/patterns/prime generator/prime.perfcompopts
krishnakeshav: I added .graph file and larger test.

krishnakeshav: 1. Will fix newline thing in follow up patch.
2. Also added .graph file

test/patterns/prime generator/prime.stdin
krishnakeshav: Updated .stdin with larger test.

test/GRAPHFILES
```diff
@@ -290,3 +290,5 @@ users/npadmana/twopt/twopt-paircount.graph
 studies/elegance/elegance.graph
 # suite: Samples (bogus)
 Samples/Performance/foo.graph
+#suite: Pimality test
```
lydia-duncan: Typo - this should say "Primality".  Would you move it to before the Samples suite, while you're at it?  If I have any other changes to it, I'll take them on myself (as they will likely be persnickety) :)

lydia-duncan:
This looks great, thanks!  Once you fix the typo, I will merge it :)

ben-albrecht:
As a helpful follow-up to this PR beyond the C++ implementation, consider changing the test to only print if the number provided from `stdin` is a prime or not, and update the `.good` file to be only 4 lines instead of 10,000 😄

krishnakeshav:
Changes made in - #5612 @ben-albrecht

PR #5532, opened 2017-03-06
Move user map assoc distribution to modules

mppf:
Passed full local testing.

- [x] GASNet testing in distributions/bradc/assoc

TODO after feature freeze:
 * add documentation

- [ ] draft chpldoc documentation and circulate it
- [ ] look at https://arxiv.org/abs/1406.2294


mppf:
@bradcray - this PR just moves UserMapAssoc to modules/dists. Could you review? I'm particularly interested to know if there are other steps you think are necessary, beyond my main TODO to write documentation.

bradcray:
I don't have time to review this in the near-term.  The main things I'm interested in seeing other than documentation are:

* a group review of the user-interface to avoid future API changes, as with anything that's put into modules/
* a test along the lines of test/arrays/userAPI/arrayAPITest.chpl + arrayOps2D.chpl that tries all of the public interfaces on associative domains in a simple way applied to the domain map so that a month from now if someone uses this we don't run into a "clear() was never implemented" type of problem.

mppf:
@bradcray - that's the kind of feedback I was looking for. I don't consider this low-hanging or priority enough to continue for the feature freeze.

mppf:
https://arxiv.org/abs/1406.2294 jump consistent hashing is interesting but only really relevant if it's important to make migrating a distributed associative thing between n locales and n+1 locales.

mppf:
On names:
```
use HashedDist;
D dmapped Hashed

use GranularDist;
D dmapped Granular

use IndexToLocaleDist;
D dmapped IndexToLocale

  -> worry that this means index 0 always goes to locale 0

  {1..10} dmapped IndexToLocale;
    -> implemented as associative array

```

modules/dists/HashedDist.chpl
```diff
@@ -42,15 +47,77 @@ record DefaultMapper {
 //
 // The distribution class
 //
-class UserMapAssoc : BaseDist {
+
+/*
+
+The Hashed distribution stores an associative domain or array across many
```
bradcray: "...maps an associative domain [and its arrays] to a set of target locales"?

modules/dists/HashedDist.chpl
```diff
@@ -42,15 +47,77 @@ record DefaultMapper {
...
+locales. Each index (or key) in the domain is stored in a locale based upon that
```
bradcray: "stored in a locale" -> "mapped to a locale"
"that index" -> "its value"

modules/dists/HashedDist.chpl
```diff
@@ -42,15 +47,77 @@ record DefaultMapper {
...
+index. When constructing a Hashed distribution, you can optionally provide
+provide a `mapper` function or function object to compute the destination
```
bradcray: "provide" appears twice in a row

modules/dists/HashedDist.chpl
```diff
@@ -42,15 +47,77 @@ record DefaultMapper {
...
+  record CustomMapper {
+    proc this(ind:string, targetLocs: [?D] locale) : D.idxType {
+      const numlocs = targetLocs.domain.size;
+      // use the first digit of the string to choose the destination locale
+      var byte: int = ascii(ind);
+      return byte % numlocs;
+    }
+  }
+
+If a custom mapper is not provided, a default mapper will be used. The default
```
bradcray: Is the default mapper the same as the local hash function that we use within a locale to map to the open addressing slots?  If so, I worry that this could result in weird interference effects in which since each locale is getting a specific fraction of the keys that the local hashing wouldn't populate the table evenly (I'm not sure I'm being clear).  Put another way, I wonder whether we should have two default hash functions for each type, one for inter-locale, one for intra-locale.  All that said, I wouldn't hold this PR up over this (if it's even true that we're using the same hash both times).

mppf: Actually there's some interesting work on hashing that points out that if you have 2 completely different hash functions, the combinations of those are almost as good as a 3rd completely different hash function. So, it'd be appealing, if we are concerned about such things, to provide 2 hash functions, and then to use a different combination of these in the mapper than we use in the hashtable itself. Let me know if you think I should make an issue proposing that.

(I was thinking about proposing that our hash functions be "keyed" by an integer argument; in which case if providing hash functions for user types becomes user facing, they'd only need to create one overload, but could use the "key" to choose totally different functions if desired. Point being, asking all types to have "2 hash functions" doesn't necessarily mean that there are actually 2 functions to implement).

ronawho:
@mppf do you mind removing uses of `_value.these()` from HashedDist (removed from UserMapAssoc in https://github.com/chapel-lang/chapel/pull/11559)?

modules/dists/HashedDist.chpl
```diff
@@ -20,13 +26,13 @@ config param debugUserMapAssoc = false;
 // b/c this matches best with the expected use and it is
 // easy to guarantee that the returned locale is in the target set.
 record AbstractMapper {
-  proc indexToLocaleIndex(ind, targetLocales: [] locale) : int {
+  proc this(ind, targetLocales: [?D] locale) : D.idxType {
```
benharsh: Should we use the 'const ref' intent explicitly here?

test/arrays/userAPI/assocDistributedHashed.chpl
```diff
@@ -0,0 +1,44 @@
...
+  D += "three";
+  D += "four";
+  D += "five";
+  D += "six";
+  D += "seven";
+  D += "eight";
+  D += "nine";
+  D += "ten";
+
+  testAssocArrayAPI(A);
```
benharsh: I think the following methods were missed in the associative API test:
- clear
- requestCapacity
- isSubset, isSuper

There are also the set operations, but I feel like there may have been discussion recently about retiring/changing those, so feel free to skip them for now.

vasslitvinov:
Looks good to me. I wish I could propose a more specific name than "Hashed".

I do suggest to find a more specific name than "DefaultMapper". How about:
* `DefaultHashedMapper`
* `DefaultLocaleMapper`

?

vasslitvinov:
Good point about DefaultMapper being an internal name. +1 to `DefaultHashToLocaleMapper`. Or we can just leave `DefaultMapper` for now and change it later if desired.

modules/dists/HashedDist.chpl
bradcray: I like the keyed hashing function concept.

PR #5534, opened 2017-03-06
Fix to this morning's 16-node XC perf annotation

bradcray:
I got the format wrong for this case.  Also had a typical off-by-one
error for the date of BHarsh's wide pointer optimization.

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -223,7 +223,7 @@ all:
       config: chapcs
   03/03/17:
     - Array views (#5338)
-  03/05/17:
+  03/06/17:
     - Wide-pointer optimization for arrays (#5487)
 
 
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho : Can you give this a third look?

ronawho:
👍

ronawho:
Ok, so this is getting embarrassing for one or both of us. You can't use "-text" without a config

lydia-duncan:
@bradcray - I tend to test for what @ronawho just found by:

1. sourcing `third-party/chpl-venv/install/<platform>/<python version>/chpl-virtualenv/bin/activate` so I am in a virtual environment with pyYAML
2. running `start_test --generate-graphs` and seeing if what I've written works
2a. if it does work, I make a change I know doesn't work and rerun to ensure I'm noticing properly

ronawho:
Would that catch ml- specific stuff?

lydia-duncan:
No, but it would catch "-text" without a config or if config was misspelled, as that gets parsed before we determine if the ml- stuff applies

bradcray:
I did run start_test --gen-graphs before submitting both PRs, didn't notice any errors, and got annotations on my local graphs.  I assumed that this was because I wasn't generating graphs for the 16-node results.

lydia-duncan:
Hmm, I'm probably wrong on the ml case, then

lydia-duncan:
I've definitely caught stuff with the 16 node XC config with that strategy, though

PR #5543, opened 2017-03-07
Add a workaround for an issue with RVF'ing a reduce intent shadow var

mppf:
After PR #5134, the _build_tuple_always_allow_ref function used in the
implementation of reduce intents no longer caused remote value forwarding
to decide references were mutable. The result was forwarding and
localizing a reference to a reduction shadow variable in this test

    parallel/forall/vass/binary-tree-spawn.no-recurse.chpl

This PR adds a simple workaround to resolveIntents to restore something
like the previous behavior. However, it would be better for forall/task
intents to clearly mark the arguments in a way that distinguishes these
cases for references:

 * not const
 * const
 * later analysis should decide if it's const

Passed full local testing.
Passed GASNet testing.
Trivial and not reviewed.

### DIFF
--- a/compiler/passes/resolveIntents.cpp
+++ b/compiler/passes/resolveIntents.cpp
@@ -167,9 +167,19 @@ IntentTag concreteIntentForArg(ArgSymbol* arg) {
   else if (arg->hasFlag(FLAG_ARG_THIS) && arg->intent == INTENT_CONST)
     return constIntentForThisArg(arg->type);
   else if (toFnSymbol(arg->defPoint->parentSymbol)->hasFlag(FLAG_EXTERN) &&
-           arg->intent == INTENT_BLANK) {
+           arg->intent == INTENT_BLANK)
     return INTENT_CONST_IN;
-  } else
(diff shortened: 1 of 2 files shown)


mppf:
@benharsh - could you help me understand whether the remote.chpl failure just means that its .good file needs updating? Or what other work is required?

Note that remote.chpl started failing after PR #5134. This PR doesn't fix it, but if there's a simple fix, I'd like to put it into this PR.

Thanks!

benharsh:
There wasn't anything missing from the .good file in the output, so I would guess that it just needs updating.

lydia-duncan:
Did this fix also resolve the issue with parallel/forall/vass/ri-iterator-with-on that turned up in nightly testing?

mppf:
@lydia-duncan - yes

mppf:
@benharsh - I accepted the current test output for remote.chpl as the good file. I'm going to let you be in charge of deciding if that's wrong for some reason... but it sounded like what you're suggesting. thanks.

PR #5546, opened 2017-03-07
Implement support for param fields and initializers

lydia-duncan:
Most of the support code added for type fields was sufficient, but some minor
tweaks were needed for param fields.

Details: In AggregateType::getInstantiation, the detection of prior instantiations
needed to be modified, the "isArgSymbol" branch was completely unnecessary,
the field's check against its type should have been against dtUnknown instead
of NULL, and another update to its type was missing.  In the insertion of omitted
initialization, we needed to flag the temporary we insert as param if the field it is
used for was param.

The remaining failures are shared across initializers and will be resolved separately:
- Currently, using a param field in the default initialization of another field does
  not work, due to the pre-existing bug with utilizing prior fields.
- declaring a variable of the instantiated class type without using new breaks for
  generic classes and records
- using a default value for one of the arguments in an initializer on a generic type
  still breaks

Add five new tests of this functionality, and remove the .future file on another:
- Validates using multiple instantiations for a particular initializer when the type
  defines a param field
- Validates using the same instantiation for a particular initializer multiple times
  when the type defines a param field
- Validates using the param field in the declared initialization for a concrete field
  - when the initialization in the initializer for the concrete field is omitted (future work)
  - when the initialization in the initializer for the concrete field explicitly uses a
    generic argument
  - when the initialization in the initializer for the concrete field explicitly uses
     the param field.
- Remove the .future file for omittedGenerics1.chpl, which validates the omitted
  initialization for param fields declared with only a type or only a default value.

Passed full paratest


lydia-duncan:
This PR works towards the completion of #5198

lydia-duncan:
Passed full std paratest

PR #5548, opened 2017-03-07
Remove remnants of the array indexing optimization

ronawho:
Remove a stop-gap optimization that eliminated the inner multiply in dsiAccess.
Happily, this optimization has been obsoleted by the array-views (#5338) work.
With array-views we're able to avoid the inner-multiply in more principled and
fine-grain manner (see #5338's PR message for more info.)

This just removes the remnants of the optimization that weren't already removed
with the array-views merge. i.e. it reverts the parts of #4510 that were still around.


ronawho:
@benharsh mind doing a sanity check?

benharsh:
Looks good!

PR #5549, opened 2017-03-07
Implement support for initializers on types with generic var fields

lydia-duncan:
Turns out everything I needed was already present!  No tweaks necessary.

Adds three new tests and removes one .future file:

The first verifies that we get the same type when we pass in the exact same
argument to the initializer in multiple calls.  The second verifies that we
get the same type when we pass in two different arguments of the same type.  The
third verifies that we get two different types when we pass in two arguments of
different types.

With the support for param and generic var/const fields, the future now works in
its entirety.

Running full paratest on clean branch (no surprises expected)


lydia-duncan:
This also is in support of #5198

PR #5550, opened 2017-03-07
Reduce memory leaks in AMR tests

bradcray:
This cleans up a bunch of user memory to dramatically reduce the amount
of memory leaked by the studies/amr tests.  While the original author did
a (reasonably) good job of creating destructors for the classes that cleaned
up after themselves, he did not do a good job of deleting the classes used
by the programs that exercised the framework (and in at least one major
important case, missed the class itself).

The few remaining leaks that I'm aware of are some compiler-leaked memory
related to first-class functions and some MDNodes that I wasn't easily able
to track down in AMR_AdvectionCTU_driver.chpl.

### DIFF
--- a/test/studies/amr/advection/amr/AMR_AdvectionCTU_driver.chpl
+++ b/test/studies/amr/advection/amr/AMR_AdvectionCTU_driver.chpl
@@ -195,7 +195,7 @@ proc main {
 
   //<=== Generate output <===
   
-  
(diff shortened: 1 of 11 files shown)


bradcray:
@noakesmichael will likely be interested in this one, at some level.

(I'm running a valgrind test on it as I type to make sure I didn't free anything unfairly).

Issue #5545, opened 2017-03-07
Internal compiler error using generic first class functions


npadmana:
### Summary of Problem
I'm running into an internal compiler error trying to compile the code below. Note that I wouldn't expect this code to compile, since I'm trying to capture a generic function (which is currently not supported). However, I did want to flag the fact that this triggered an internal error.

### Steps to Reproduce

**Source Code:**

```chapel
proc f1(x) {
  return x;
}

var func = f1;
```

**Compile command:**
`chpl buggy.chpl`

**Output:**
```
buggy.chpl:1: internal error: CHE0320 chpl Version 1.14.0.2a61e6e

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.
```

bradcray:
Anything that generates an internal error is definitely a problem on our part, so thanks for reporting this.  As you note, the problem here is the fact that the function is generic.  I've introduced an error message to catch this case in PR #5552 and will hope to merge today/tomorrow.

bradcray:
Fixed by PR #5552.  I didn't get the message right to have this close automatically.

PR #5552, opened 2017-03-08
Added a check that first-class functions aren't generic

bradcray:
We require first-class functions to be concrete.  This adds a quick
check that prints an error if they aren't.

I've also added an expanded version of the test that @npadmana 
included when filing issue #5545 that covers a few different
cases.


bradcray:
Passed linux64 testing and revealed no new futures

PR #5554, opened 2017-03-08
Do not insert an autoCopy for chpl__iterLF

benharsh:
ArrayViews exposed an issue in callDestructors where an autoCopy was
inserted for a chpl__iterLF variable with an ArrayView type. The
autoCopy results in a non-view type, leading to compilation issues
later on. Vass and I feel that this autoCopy should be unnecessary
given the usage of chpl__iterLF variables, so for now we will
test for FLAG_CHPL__ITER and avoid inserting an autoCopy.

Testing:
- [x] full local
- [x] full gasnet
- [x] numa release/examples


benharsh:
For those following along at home, this fixes the miniMD compilation failure on numa configurations.

benharsh:
@vasslitvinov : provided that testing passes, does this look alright to you?

vasslitvinov:
Looks great!

PR #5555, opened 2017-03-08
KNL Locale Model

dmk42:
This change implements a locale model for KNL.  It adds four member functions for public use.  These are implemented in all locale models, so that users can write portable code and yet use the advantages of specific types of machines.  Locale models that have only one type of memory stub these functions out to point to that memory.

locale.defaultMemory()  can get you back to normal memory if you are nested inside an on statement that uses a different kind of memory.  On KNL, this is normal off-chip DDR memory.

locale.largeMemory()  for whatever type of memory is the largest.  On KNL, this is DDR memory.

locale.lowLatencyMemory()  for low latency memory.  On KNL, this is DDR memory.

locale.highBandwidthMemory()  for high-bandwidth memory.  On KNL, this is the on-chip MCDRAM.

The KNL NUMA domains that contain both compute cores and access to off-chip memory are the normal sublocales.  For example, in KNL's SNC4 configuration, which has four compute-capable NUMA domains, these are numbered 0-3.

Internally, the KNL NUMA domains that contain only on-chip MCDRAM are numbered consecutively above the compute sublocales.  For example, in the SNC4 configuration, these are numbered 4-7.  Sublocale 4's memory is closest to sublocale zero's cores, 5 is closest to 1, etc.  If sublocale 4 is to be used, it means to compute in sublocale zero and store data in its closest MCDRAM bank.  Users will normally access these sublocales with the member functions rather than having to know their number.

If any sublocale can be used, and data is to be stored in normal off-chip memory, the sublocale ID is c_sublocid_any (which is negative 2).  If any sublocale can be used, and data is to be stored in on-chip MCDRAM, the sublocale ID is one greater than the highest MCDRAM bank (on SNC4, this is sublocale 8).

As an example, to specify locale zero, sublocale 2, MCDRAM, you can use the following.

Locales(0).getChild(2).streamingMemory()

The following specifies any sublocale, and MCDRAM.

Locales(0).streamingMemory()

You can also specify the same locale as an existing variable, and still control which type of memory is used.

var x;
on x.locale.streamingMemory() { ... }

If the KNL node you are using is booted in cache mode (where MCDRAM is not accessible except as a level 3 cache), the streamingMemory() function just gives you normal DDR memory.

The KNL locale model is chosen by setting CHPL_LOCALE_MODEL=knl .
The following libraries are required.
hwloc
memkind
libnuma

PR #5566 (already merged) is a prerequisite for this PR.


doc/technotes/localeModels.rst
```diff
@@ -21,13 +21,14 @@ multiple instances of memories and/or processors with differing
...
-There are currently two locale models available, flat and NUMA.  The
+There are currently three locale models available, flat, NUMA, and KNL.  The
 flat model is the default and maps closely to the view of locales
 implemented in the 1.7 release.  The NUMA locale model maps sublocales
 to NUMA domains.  The NUMA model is currently implemented at a prototype
 level. Performance has not yet been a focus in the NUMA locale model and
 will require additional effort in future releases.  We expect to add
-more locale models in future releases.
+more locale models in future releases.  The KNL locale model provides
+access to on-chip high-bandwidth DRAM.
```
gbtitus: I would swap these last two sentences, so that we end with the "and more are coming in the future" one.  Also, the KNL sentence should say that it's for compute nodes based on self-hosted Intel Xeon Phi (formerly Knights Landing) processors and includes provision for access to the tightly-coupled on-board high bandwidth DRAM memory.  (Strictly speaking the HBM isn't on-chip, it's on-package, but I think they're just referring to it using "on-board" and other general phrases like that so they have implementation flexibility.

modules/internal/ChapelLocale.chpl
```diff
@@ -247,6 +247,35 @@ module ChapelLocale {
...
+      _throwPVFCError();
+      return this;
+    }
+
+    pragma "no doc"
+    proc streamingMemory() : locale {
+      _throwPVFCError();
+      return this;
+    }
+
```
gbtitus: The `madvise(2)` differentiation between random and streaming access does make sense to me.  But that's intended to tell the kernel something about the program's behavior with respect to the given memory segment.  Locale models, on the other hand, are about architecture rather than program behavior and I think we should be describing the characteristics of the memory itself rather than the program access behavior that would match up with it.  So I'd prefer to see the random vs. streaming differentiation expressed as (optimized for) latency vs. bandwidth.  I'd be interested in what others think about this, though.  Maybe our notion of a "locale model" needs to be expanded to include how programs expect to interact with the architecture, rather than just being about the architecture as such.

Another topic: this model allows a program to express at most one memory characteristic and get either a matching memory sublocale or, failing that, a fallback sublocale chosen by the locale model.  Some time ago we had discussed, and I thought informally settled on, a model in which programmers would somehow be able to express an ordered set of characteristics and the returned sublocale would be the locale model's best attempt to match all of those.  For example you could say something like `selectMemory(large, latency)` to say that you wanted a memory that was both large and latency-optimized but that the size was more important to you than optimization for access pattern.  Or maybe you'd even be able to say `selectMemory(sizeMin=1**30, bandwidth)` where sizeMin was a required characteristic, to say that you wanted a bandwidth-optimized memory but it had to be at least 1gb in size.  This isn't actually so important when all we have to choose from is size and latency vs. bandwidth, but the day may not be far off when memories also have volatility and/or resilience characteristics, and it could be very important to be able to express that you want a large resilient memory for your long-lasting in-memory database. It looks hard to achieve this with characteristic-specific selector functions as you've implemented here.

modules/internal/localeModels/flat/LocaleModel.chpl
```diff
@@ -31,6 +31,24 @@ module LocaleModel {
   use LocaleModelHelpFlat;
   use LocaleModelHelpMem;
 
+  //
+  // The task layer calls these to convert between full sublocales and
+  // execution sublocales.  Full sublocales may contain more information
+  // in some locale models, but not in this one.
+  //
+  export
+  proc chpl_localeModel_sublocToExecutionSubloc(full_subloc:chpl_sublocID_t)
```
gbtitus: The `chpl_localeModel_` prefix here is probably not needed.  We need this sort of thing for runtime symbols because they're in C and the namespace is flat, but in Chapel code we could disambiguate using 'module.' syntax in the unlikely event `sublocToExecutionSubloc` was ambiguous.  And the shorter names are nice.

modules/internal/localeModels/flat/LocaleModel.chpl
```diff
@@ -31,6 +31,24 @@ module LocaleModel {
...
+  {
+    return full_subloc;  // execution sublocale is same as full sublocale
+  }
+
+  export
+  proc chpl_localeModel_sublocMerge(full_subloc:chpl_sublocID_t,
```
gbtitus: same for `chpl_localeModel_` prefix here

runtime/include/localeModels/flat/chpl-locale-model.h
```diff
@@ -66,4 +66,17 @@ c_sublocid_t chpl_rt_sublocFromLocaleID(chpl_localeID_t loc) {
   return c_sublocid_any;
 }
 
+//
+// These functions are exported from the locale model for use by
+// the tasking layer to convert between a full sublocale and an
+// execution sublocale.
+//
+extern
+c_sublocid_t chpl_localeModel_sublocToExecutionSubloc(
```
gbtitus: Ah.  Now I see why these have the `chpl_localeModel_` prefix.  I still think there's a simpler way.  Can the tasking layers just understand low-level sublocales, what I think you're calling "execution" sublocales, and the module code deal with any need to turn these into more-descriptive things?  I.e., make the sublocToExecutionSubloc() and sublocMerge() calls in the module code that interacts with the C runtime, rather than in the C runtime itself?

test/distributions/sungeun/dataPar2.prediff
```diff
@@ -2,7 +2,7 @@
 # sort the output to handle the numa case where iterator debug output is in a
 # non-deterministic order. Don't sort for non-numa
```
gbtitus: "Don't sort for flat"

test/performance/hilde/coforallOnByRef.skipif
```diff
@@ -1,2 +1,2 @@
 # numa is orthogonal to this test and changes comm counts, so skip
-CHPL_LOCALE_MODEL==numa
+CHPL_LOCALE_MODEL!=flat
```
gbtitus: fix comment above: "non-flat is orthogonal ..."

third-party/jemalloc/Makefile
```diff
@@ -15,7 +15,8 @@ CHPL_JEMALLOC_CFG_OPTIONS += --host=$(CHPL_MAKE_TARGET_PLATFORM)-unknown-linux-g
 endif
 
 CHPL_JEMALLOC_CFG_OPTIONS += --prefix=$(JEMALLOC_INSTALL_DIR) \
-			     --with-jemalloc-prefix=je_
+                             --with-jemalloc-prefix=je_ \
+                             --with-private-namespace=jeint_
```
gbtitus: Understood that the need for this comes from the current PR, but still it seems it might be better as a separate one done before the current one.  Perhaps with a mention that the current one is what creates the need.

Also, "jeint" seems vague and/or subject to collision with someone else doing the same thing.  Maybe "jechpl"?  (I could actually see using "jechpl" for both the public and private symbols, but realize "je" for the public ones predates this change.  @ronawho may have some input here also.)

gbtitus:
I made a bunch of line comments, some extraordinarily wordy :-) and I think there are some things we need to work through still on this.  But I also want to say that it's super to have this work arriving!

modules/internal/ChapelLocale.chpl
mppf: Adding on to what Greg is saying, I thought we were trying to make getChildCount / getChild take in an optional bitmask argument indicating the type of resources requested. I think we were hoping that large/random/streaming memory would be bitmask positions and that we'd have functions to gather up sublocales with a particular combination of properties.

I don't think what you have here is necessarily the wrong direction; I just think we'd like to possibly have a different user interface that's got more of a flavor of querying for the sublocale with the best match for some pattern of expected use.

doc/technotes/localeModels.rst
```diff
@@ -96,8 +97,8 @@ Qthreads thread scheduling
 
 When qthreads tasking is used, different Qthreads thread schedulers are
 selected depending upon the ``CHPL_LOCALE_MODEL`` setting.  For the flat
-locale model the "nemesis" thread scheduler is used, and for the NUMA
-locale model the "distrib" thread scheduler is used.  This selection is
+locale model the "nemesis" thread scheduler is used, and for the NUMA and KNL
+locale models the "distrib" thread scheduler is used.  This selection is
 done at the time the Qthreads third-party package is built, and cannot
 be adjusted later, either at user compile time or at execution time.
```
ronawho: Seems like we should add a section for the KNL locale model to this file. Doesn't have to be done at part of this commit, but I think we wants something in by doc freeze.

ronawho: Should also update chplenv.rst:CHPL_LOCALE_MODEL

modules/internal/DefaultRectangular.chpl
```diff
@@ -21,7 +21,7 @@
 //
 module DefaultRectangular {
   config const dataParTasksPerLocale = 0;
-  config const dataParIgnoreRunningTasks = if CHPL_LOCALE_MODEL=="numa" then true
+  config const dataParIgnoreRunningTasks = if CHPL_LOCALE_MODEL!="flat" then true
```
ronawho: Huh, we should revisit this logic. Doesn't have to be part of this commit, but with this set to false, we can end up heavily oversubscribing the system. See 720c769658358e406745ed8dc11160dd9e61e023 for more info. 

Maybe make an issue to track post-commit TODOs?

modules/internal/localeModels/flat/LocaleModel.chpl
ronawho: `chpl_localeModel_sublocToExecutionSubloc` is pretty long name. I'd prefer something shorter too.

runtime/include/localeModels/knl
```diff
@@ -0,0 +1 @@
+numa
```
ronawho: What is this doing?

runtime/src/tasks/qthreads/tasks-qthreads.c
```diff
@@ -892,18 +892,21 @@ void chpl_task_addToTaskList(chpl_fn_int_t       fid,
...
+    assert(full_subloc != c_sublocid_none);
 
     PROFILE_INCR(profile_task_addToTaskList,1);
 
     if (serial_state) {
         // call the function directly.
         requested_fn(arg);
     } else {
+        c_sublocid_t execution_subloc =
+          chpl_localeModel_sublocToExecutionSubloc(full_subloc);
```
ronawho: I'm curious if this is going to result in any noticeable increase in task-spawning times for flat. I doubt it, but it's something to keep an eye out for.

modules/internal/ChapelLocale.chpl
gbtitus: Just a follow-on comment re: what @mppf just said: I'm not that fond of the bitmask idea, because I don't see how it can express an ordering for the desired characteristics and because it seems more like a C-style solution to me than a Chapel-style one.  Isn't there a more Chapelrific way to do this?

doc/technotes/localeModels.rst
ronawho: There's also doc/platforms/knl.rst which deserves some attention

ronawho:
Overall notes: 

I'm Super excited about this! 

I didn't review a whole lot of the actual knl locale model, but I'm pretty happy with this overall

Do you set `CHPL_TARGET_ARCH=knl`? On cray's we get the target_arch based on the loaded targeting module. Currently it's `CHPL_TARGET_ARCH=mic-knl` from `craype-mic-knl` There are some .execenvs and some other place that refer to CHPL_TARGET_ARCH=mic-knl. I think something needs to change, here I'm just not sure what.

We should figure out what module build changes are required to incorporate this into the cray module release and figure out what sort of testing we want to do. This will probably be a thing for Tony

third-party/jemalloc/Makefile
ronawho: Agreed, this is small, but I think it should be split off into a separate PR. 

I'm also not a fan of jeint_, I think we should go with something with `chpl` init. My preference would be je_chpl_ or chpl_je_

ronawho:
And it looks like in a few places (mostly tests) you adjusted `==numa` to `!=flat`. I'm not sure this is needed for all of these tests, and I think going forward we should be more cognizant of future locale models, especially those being developed outside of Cray. (e.g. the current gpu work being done)

modules/internal/ChapelLocale.chpl
ronawho: This is something we discussed in a KNL meeting at some point. Notes can be found at chapel-docs/notes/topics/KNL/2016-04-21.txt

third-party/jemalloc/Makefile
dmk42: I put an example of how we might do this in #5566 .  If you like it, you can approve that PR, or we can discuss doing something else.

runtime/include/localeModels/knl
dmk42: Oh, github shows this in a strange way.  It's actually just a symbolic link from knl -> numa.  I didn't need to change anything from numa for this directory, but our build system picks which directory to use based on the locale model name, so I made a symbolic link.

doc/technotes/localeModels.rst
dmk42: I agree.  I'm about to push a change that addresses the documentation, which might hide this comment.

doc/technotes/localeModels.rst
dmk42: I've fixed chplenv.rst and am about to push that change.  I agree that localeModels.rst and knl.rst need additional work before the doc freeze.  I've created issue #5569 to track changes such as this that need to be made after the KNL locale model commit.

runtime/include/localeModels/knl
ronawho: Weird. Did you make the directory or the file a sym link? 

If dir, I think try making the dir a real dir, and just symlinking chpl-locale-model.h to the numa one. Or the file is small enough that I don't think duplicating would be terrible

runtime/include/localeModels/flat/chpl-locale-model.h
dmk42: The problem is that it's the tasking layer that caches the requested sublocale via chpl_task_setSubloc and retrieves it with chpl_task_getRequestedSubloc.  To also store it in the locale model, we would need to cache in two places, and it would be hard to ensure that no future changes called the wrong layer.

gbtitus: Got it.  Yes, if the "requested subloc" for a task needs to include the full information, then because they're responsible for storing them in task-private memory, the runtime tasking layers have to be able to assemble and disassemble them even though they don't need to know what the assembled form looks like.  Now that I get that, I'm okay with this at least for now.  I have in mind an inkling of what might be a better way to do module-oriented task-private data, not just for this but other things like serial state, but I'm not sure I can get that done for the release and definitely this PR shouldn't be held up for it.  I'll make a design issue for that.

gbtitus:
From my point of view everything is okay except that we need to reach agreement on what to do about the sublocale-selector function(s): (random vs. streaming) vs. (latency-optimized vs.bandwidth-optimized), and other topics like whether we need to be able to ask for multiple characteristics, ordered or not, etc.

mppf:
@gbtitus - even so, isn't this PR good enough as a start? I mean, I think we want to talk about the user-API for selecting sublocales, but it seems what is in this PR is a reasonable start. Can't we beef up that interface in a separate effort from merging this PR?

modules/internal/ChapelLocale.chpl
dmk42: Regarding the selection of multiple characteristics, sorry, I tried to take into account all the information from before I got involved, but I somehow missed that.  I am not sure how a hierarchical specification would work.  If someone asked for low latency first, and high bandwidth second, would we give them DDR because latency was first priority, or would we give them MCDRAM because it was the best compromise between the specified characteristics?  If the former, then the KNL locale model would only ever have to look at the first characteristic in order unless a specific size was specified first, and if the latter, then everything except a specific size could be specified with a bit vector rather than a hierarchy.

I think what we are missing most here is experience with a wide range of locale models.  As architectures evolve, it may become clearer what is needed.  I suspect it would be better to start with what we know we are going to need and then grow as requirements change.  I am open to continuing this discussion; the only question is whether it continues as part of this PR or separately.

I have no objection to renaming the characteristics to be about the hardware rather than about the program.

modules/internal/localeModels/flat/LocaleModel.chpl
dmk42: Given that we've resolved that there is currently a need for calling into the locale model from the tasking layer, the remaining question about the chpl_localeModel_* functions is their long names.  I could make it _lm_ instead of _localeModel_.  I just wanted to be sure that someone reading the tasking code would know exactly where to look when they came across those calls.  I'm not sure _lm_ is enough to do that, but I'm open to suggestions.

test/distributions/sungeun/dataPar2.prediff
dmk42: I'm about to push a change that fixes these comments in prediff and skipif files.  Pushing the change will eventually hide this github comment.

modules/internal/DefaultRectangular.chpl
dmk42: I've added this dataParIgnoreRunningTasks item to issue #5569 as a non-KNL-related issue that should be investigated at a later date.

runtime/include/localeModels/knl
dmk42: The symbolic link is consistent with runtime/include/cygwin{32,64} which I thought was a good way to show that this part of the code doesn't change for KNL -- i.e. file hierarchy as documentation.

dmk42:
I was pretty careful about what I changed from ==numa to != flat, and I left a lot of those occurrences unchanged.  I agree that as we get more and more locale models, we will need to take a closer look at what tests to run for each.

modules/internal/ChapelLocale.chpl
dmk42: If nobody objects, I will change randomMemory to lowLatencyMemory, and streamingMemory to highBandwidthMemory.  I won't push that change right away because it will hide this part of the github conversation.

gbtitus:
Replying to @mppf about 'isn't this good enough for a start'?  Yeah, but.  As you know I lean heavily, at least with regard to introducing public interfaces, toward getting them as product-like (vs. prototypical) as we can.  Of course we can merge this now.  It works, which is a great step forward.  But we know, or at least I think we know, that what we want for the style of this part of the interface is rather different than what is here.  If we could easily get closer to that style that would benefit the users and we'd get high-quality feedback (because it would be about something more like our intentions).  But obviously sure, we can release this style of interface now while reserving the possibility of introducing a better one later and deprecating this one at that time.

modules/internal/ChapelLocale.chpl
gbtitus: Answering "would we give them DDR because latency was first priority", it would depend on whether we decided the selector function should treat the characteristics as being ordered in terms of desirability.  If yes then yes, I think we'd want to give them DDR.  If no then definitely MCDRAM would be a possible answer (and if they wanted to skew us toward giving them DDR instead they could throw "large" into the set of desired characteristics).  In any case, I think this is something we could take offline from this PR.  A change to the style of selector I'm talking about seems too large for this PR now, both philosophically and in terms of getting it in well ahead of the release.  So let's go with what you have, modulo the change from program to hardware characteristics unless anyone objects to that.

modules/internal/localeModels/flat/LocaleModel.chpl
gbtitus: Let's leave them long for now.  They're entirely internal, and I have some hope we might not need them for terribly long anyway.  See https://github.com/chapel-lang/chapel/issues/5587.

ronawho: I think `lm` is fine. It's the same prefix we use for testing and in some other place. i.e. `lm-numa.good`

runtime/include/localeModels/knl
ronawho: Ah ok. Good enough for me, thanks for the explanation

dmk42:
I think I see a decent way to extend the interface.  I've added it to issue #5569 and would be interested in feedback there.

I will go ahead and rename the characteristics to be about the hardware, and redo a couple of paratests before merging this PR (so there is plenty of time to stop me if something else comes up).

PR #5558, opened 2017-03-08
Mark the Chapel test "failed" if .skipif or .suppressif returns error

awallace-cray:
As discussed with mppf. If the test harness runs a .skipif (or .suppressif) script and it emits anything to stderr, sub_test should flag the corresponding Chapel test as failed. 
This change does not apply to directory-wide SKIPIFs.

### DIFF
--- a/util/test/sub_test
+++ b/util/test/sub_test
@@ -408,6 +408,15 @@ def printTestVariation(compoptsnum, compoptslist,
     sys.stdout.write(')')
     return
 
+# print '[Elapsed time to compile and execute all versions of ...'
+#   in the format expected by convert_start_test_log_to_junit_xml.py.
+
+def printEndOfTestMsg(test_name, elapsedTime):
+    sys.stdout.flush()
+    print('[Elapsed time to compile and execute all versions of "{0}" - '
+        '{1:.3f} seconds]'.format(test_name, elapsedTime))
+
+
 # return true if string is an integer
 def IsInteger(str):
     try:
@@ -1095,8 +1104,7 @@ for testname in testsrc:
                 sys.stdout.write(str(e)+'\n')
                 sys.stdout.write('[Error processing .skipif file for %s]\n'
                     % os.path.join(localdir, test_filename))
-                sys.stdout.write('[Elapsed time to compile and execute all versions of "%s" - 0.0 seconds]\n'
-                    % os.path.join(localdir, test_filename))
+                printEndOfTestMsg(os.path.join(localdir, test_filename), 0.0)
                 do_not_test=True
             if do_not_test:
                 break
@@ -1121,8 +1129,7 @@ for testname in testsrc:
                 sys.stdout.write(str(e)+'\n')
                 sys.stdout.write('[Error processing .suppressif file for %s]\n'
                     % os.path.join(localdir, test_filename))
-                sys.stdout.write('[Elapsed time to compile and execute all versions of "%s" - 0.0 seconds]\n'
-                    % os.path.join(localdir, test_filename))
+                printEndOfTestMsg(os.path.join(localdir, test_filename), 0.0)
                 do_not_test=True
                 break
 
@@ -2175,8 +2182,7 @@ for testname in testsrc:
 
     elapsedCurFileTestTime = time.time() - curFileTestStart
     test_name = os.path.join(localdir, test_filename)
-    print('[Elapsed time to compile and execute all versions of "{0}" - '
-        '{1:.3f} seconds]'.format(test_name, elapsedCurFileTestTime))
+    printEndOfTestMsg(test_name, elapsedCurFileTestTime)
 
 
 sys.exit(0)
--- a/util/test/sub_test
+++ b/util/test/sub_test
@@ -507,14 +507,18 @@ def runSkipIf(skipifName):
     p = subprocess.Popen([utildir+'/test/testEnv', './'+skipifName], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     (stdout, stderr) = p.communicate()
     status = p.returncode
+
     stderr = stderr.strip()
+    errmsg = ""
+    if stderr:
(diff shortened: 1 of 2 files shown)


awallace-cray:
Both branches have been tested. If anything goes to stderr when the .skipif script is executed (from sub_test), then the corresponding Chapel test is marked as failed ("[Error ..." in the log). 
If stderr is null, then there is no change to current behaviour.

mppf:
@awallace-cray - this seems OK to me but it seems to me we should be checking the exit code from test/testEnv as well.

awallace-cray:
These changes are lightly tested, in the sense that the test harness responds correctly to skipif scripts returning "True" or "False", with and without "system errors".  However, I have not tested all existing Chapel test files in all test configurations.  Also, I have not tested other places (besides the Chapel test harness) where testEnv is used: devel/test/checkIgnore.sh, test/paratest.server.

util/test/testEnv
```diff
@@ -28,7 +28,8 @@ if (-x "$envfile") {
   $envfile = abs_path($envfile);
   $execskiptest = `$envfile`;
   print "$execskiptest";
-  exit("$?");
+# exit with the return code that the envfile gave us
+  exit($?>>8);
```
mppf: Why isn't this just exit($?) ?

awallace-cray: That doesn't work. http://perldoc.perl.org/perlvar.html#%24CHILD_ERROR

mppf: Roger that.

mppf:
This is fine with me. I think it'd be nice to do a full local test if you havn't yet, just to make sure it's not disruptive. Thanks!

util/test/sub_test
```diff
@@ -504,8 +504,18 @@ def get_exec_log_name(execname, comp_opts_count=None, exec_opts_count=None):
...
-    skiptest = subprocess.Popen([utildir+'/test/testEnv', './'+skipifName], stdout=subprocess.PIPE).communicate()[0]
-    return skiptest
+    p = subprocess.Popen([utildir+'/test/testEnv', './'+skipifName], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    (stdout, stderr) = p.communicate()
+    status = p.returncode
+    stderr = stderr.strip()
+    if status:
+        if stderr:
+            sys.stdout.write(stderr+'\n')
+        Fatal('%s: exit status %d' % (skipifName, status))
```
ronawho: `Fatal` seems extreme here since it kills the entire sub_test. If you have a .skipif for a single test that's bad, won't this kill testing for the entire directory?

awallace-cray: Yes. Commit message:
    This change also makes sub_test exit immediately upon detecting
    such an error. Any Chapel tests in the current test directory which
    haven't already been processed, will never show up in the test
    results at all.
    
    Sub_test already handles some other system-error conditions the
    same way. IMO, it would be too difficult / not worth the effort
    to make sub_test continue working the current test directory in
    these cases.

ronawho: Doesn't seem hard to just skip the test if runSkipIf detected an error. Off the top of my head you coudl have runSkipIf throw an exception and have the callsite catch it and do the normal "Error processing .skipif" 

util/test/testEnv can easily fail for non system-error reasons, so it seems excessive to fail fatally

awallace-cray: That's about what I thought, but when I tried it I never got the '[Error <something>' messages to be recognized by everything downstream. For example when I made the skipif throw something on stederr, the console log "saw" the test failure but the Junit xml didn't.

ronawho: Weird, the start_test to junit script just looks for `[Error`, like start_test and every other log summarizer, so I can't think of any reason it wouldn't catch `[Error processing .skipif...`

`error_pattern = re.compile('^\[Error.*$',`

awallace-cray: The start_test to JUnit script has different logic, and the attempts I made up to now were unsuccessful. The best I could do up to now was to get the failed skipif to show up but then the following test disappeared- the JUnit script ate that part of the log because the failed skipif did not emit exactly the right stuff.

awallace-cray:
I never did finish the full test suite, and I agree we should do one.

awallace-cray:
FInished a full test suite - test-gasnet.fast. 
```
2017-03-14 01:17:46,763 [INFO] Parsed 7731 test cases from "/tmp/day1-Mon-gasnet-fast.log.15756".
...
[Summary: #Successes = 7062 | #Failures = 0 | #Futures = 19 | #Warnings = 0 ]
[Summary: #Passing Suppressions = 0 | #Passing Futures = 0 ]
```

ronawho:
At a glance, this seems way more complicated than it needs to be to me. That said, I haven't spent much time thinking about it, and I probably won't have a chance to for a while.

What's the motivation for this? i.e. what was the inspiration for this PR? If it's not related to the release, can we just punt on this for now?

awallace-cray:
mppf requested it when he was triage, and some .skipif script failed when it ran, and the test harness didn't notice anything wrong. 
I don't mind if it misses this release. 
OTOH, this has been *extensively* tested, and it would work out fine.

awallace-cray:
Refactored three instances of an important print statement (2 of which added in this PR) into one.
Rebased to master for the first time in over a month. 
Will re-run tests.

ronawho:
It'll be a while until I'd get a chance to look at this, so I wouldn't wait on me to merge assuming @mppf is still happy with this.

awallace-cray:
The only thing that concerned mppf the last time I asked, was running a full test suite before merging this. Just to be sure this would not suddenly expose a bunch of broken or ill-behaved .skipifs or .suppressifs which already existed.  I am running another full suite test on the updated PR right now. 

I am not planning to repeat all the functional tests I did a month ago, because today's commit really wouldn't change those results.

mppf:
No concerns here.

Issue #5402, opened 2017-02-23
count() on arrays is serial


bradcray:
Browsing the sources today, I was noting that the count() method on arrays is serial, which seems surprising for a parallel language (and will likely die, performance-wise, on a distributed array).  While we might ultimately want some clever heuristics to decide how much parallelism to use in various cases, it seems like we should use parallelism (a reduction) by default, as a parallel language.

sagar-sam:
I would like to work on the issue if the issue is still open. I build the code and get familiarize myself with the chapel, I am new to chapel, please guide me solving the issue.
Thank you

bradcray:
Hi sagar-sam -- It is still open.  I'd suggest reading about Chapel's data parallel features (http://chapel.cray.com/docs/master/users-guide/index.html#data-parallelism) and looking at the definition of 'proc count()' in $CHPL_HOME/modules/internal/ChapelArray.chpl as starting points.

sagar-sam:
Hello @bradcray 
I converted total variable of type sync and used a forall loop instead of a for loop.
Is it beneficial??

bradcray:
Hi @sagar-sam --

Yes, you should definitely create a pull request to start discussing the code itself.  And you'll need to sign a contributor's agreement in order to have the PR be merged.  Using a sync var and forall is a legal way to solve this problem, but probably not the most efficient, since it will require every iteration to access the same variable which could cause a bottleneck when the array is distributed across a large number of locales.  To avoid this, you'd want to use Chapel's 'reduce' expression or a 'reduce' task intent on the forall loop.

Unfortunately, I'm seeing from @lydia-duncan's tag above that someone else has submitted a PR with a reduce-related solution...  But I think it'd still be worth comparing the performance of the two approaches as part of your continuing introduction to the language...

sagar-sam:
Thanks @bradcray

PR #5559, opened 2017-03-08
Make count() on arrays parallel by default (issue #5402)

krishnakeshav:
Parallelizing count() method on array. I read the the following - https://github.com/chapel-lang/chapel/commit/690b5e01fea87d6894489796b6fc73a055f061e0. Would like to know if it's useful.
Also from Ben's comment I interpreted(not sure yet) that using **reduction** construct isn't harmful.

PR #5560, opened 2017-03-08
performance of count in chapelArray increases

sagar-sam:
issue #5402 
Performance of count in chapelArray increases.

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -2702,8 +2702,8 @@ module ChapelArray {
 
     /* Return the number of times ``val`` occurs in the array. */
     proc count(val: this.eltType): int {
-      var total: int = 0;
-      for i in this do if i == val then total += 1;
+      var total: sync int = 0;
+      forall i in this do if i == val then total += 1;
       return total;
     }
 


bradcray:
This looks correct but is likely to be less efficient than using a reduction at scale, as proposed in #5559.  The reason is that reductions will use local, uncoordinated variables for each task when accumulating their own local results and then only use synchronization when combining between tasks.  In contrast, the approach here will write into the synchronized variable for every accumulation, adding extra overhead.  Moreover, if the array is distributed, tasks across the system will all be writing into a single synchronized variable on a single locale.  Whereas in the reduction case, cross-locale communications should only occur when tasks on a given locale complete.

bradcray:
(All that said, thanks very much for taking a stab at this, and for registering your interest in the issue prior to tackling it which is a good practice.  Take my feedback above as technical notes for future consideration, not criticism).

sagar-sam:
Thank you @bradcray for your feedback,
I will definitely take it as technical notes for future consideration

ronawho:
Closing this in favor of #5559, but thanks for your interest and willingness to contribute @sagar-sam

PR #3818, opened 2016-05-03
Fix failure to report ambigous value-version of a ref-pair

mppf:
The below code was not triggering an ambiguity error in function resolution due
to some problems with how the ref-pair support was initially implemented.

    var global = 0;
    proc access() ref { return global; }
    proc access() { return global; }
    proc access() const ref { return global; }

    var x = access();  // value version should be ambiguous
    writeln(x);

Instead, the ref version was being called (because disambiguateByMatch
was returning NULL for ambiguous but the code assumed it meant
no suitable function was found).

This PR fixes this problem. In order to do so, it separates the logic for
finding ref return and not-ref return candidates from disambiguateByMatch. The
problem with the disambiguateByMatch approach is that disambiguateByMatch
returns NULL for either

 * no matching candidates, or
 * ambiguity

Other function resolution code checks for ambiguity by inspecting the number
of candidates. So, this PR adds a function, splitCandidates, that separates the
candidates into ref and non-ref lists. Additionally, it carefully handles the
case when disambiguateByMatch returned NULL but either of these lists has > 0
elements.

Thanks to @cassella for pointing out this problem.

PR #5561, opened 2017-03-08
Three way refpair

### LINKED ISSUE #5468: ref-pair .locale problem
Right now, cullOverReferences includes a special case to make a .locale query use the "setter" version. This was added in commit d6bb7fe8.

This special case is problematic. I think it trades one bug for another.

Consider the following program:

``` chapel
const ParentDom = {0..10};
var SparseDom: sparse subdomain(ParentDom);
var A:[SparseDom] int;

proc f(const ref x)
{
  writeln(x.locale.id);
}

for i in ParentDom {
  f(A[i]); // appears to work
}

for i in ParentDom {
  writeln(A[i].locale.id);  // error: halt reached - attempting to assign a 'zero' value in a sparse array: (0)
}
```

Along with other work I'm doing in cullOverReferences, I'm able to remove this special case. However, that makes programs such as the following no longer work as expected:

``` chapel
use BlockDist;

var Space = {1..10};
var D = Space dmapped Block(Space);
var A:[D] int;

for i in D {
  writeln(A[i].locale.id); // prints only 0 if workaround in cullOverReferences is disabled
}
```

The reason this program no longer works is that the array accessor uses a value-return `proc this` for POD types (including integers which are used in the example). So without the workaround, the .locale query is getting the locale of a stack-local variable, which is always local.

However, the workaround in cullOverReferences doesn't extend to this program:

``` chapel
use BlockDist;

var Space = {1..10};
var D = Space dmapped Block(Space);
var A:[D] int;

proc f(const ref x)
{
  writeln(x.locale.id);
}

for i in D {
  f(A[i]); // prints only 0 with or without workaround in cullOverReferences
}
```

Are these programs erroneous? It seems to me that the first should run without error and the second should print out the locale on which A[i] is stored (and not just Locale 0).

mppf:
Fixes  #5468.

Before this PR, cullOverReferences included a workaround in which using a ref-pair, such as an array accessor, in the context of querying its locale will always use the "setter". This workaround is problematic for reasons pointed out in issue #5468 - such as when querying the locale of a sparse array element that isn't in the array. But, the workaround also has the problem that it is very fragile - passing an array element through a function call with a `const ref` formal will prevent the workaround from working. And, without the workaround, printing out the queried locale for a Block distributed array's elements will always print `0`. See the related issue for these examples.

This PR responds to that problem by adjusting the `ref-pair` feature in the language to have the ability to select between 3 things:
 * a ref-return version
 * a const-ref-return version
 * a value-return version

Then, querying the locale or passing to a `const ref` function should use the `const ref` return array accessor. This resolves the issue.

Outline of changes:
 * adjust ContextCallExpr to store 2 or 3 CallExprs
 * adjust resolveNormalCall to separately find a ref, const ref, value return versions. This was done in the most straightforward way and I'm certain that resolveNormalCall could be improved (see below).
 * in modules, adjust array accessor methods to use a new method, `shouldReturnRvalueByValue`, instead of `!shouldReturnRvalueByConstRef`.
 * make `shouldReturnRvalueByConstRef` always return `true`, so that the `const ref` version is always available. `shouldReturnRvalueByValue` only returns `true` for POD types.

Passed full local testing 9fd800f d8e1956
Passed gasnet testing.

Future Work:
 * clean up resolveNormalCall. I'm thinking that we should gather all of the candidates and filter them before choosing ref/non-ref candidates. Right now, resolveNormalCall doesn't correctly handle ambiguity in a ref-pair, since disambiguateByMatch returns NULL for both no matching candidates and also for ambiguity. Anyway, such an improvement has to be careful to handle differences, such as a where clause in a `const ref` return but not in the `ref` return, without always returning the `ref` version. PR #3818 was an initial attempt along these lines.
 * update spec description of `ref-pair` to describe new behavior.


compiler/AST/expr.cpp
```diff
@@ -1456,6 +1469,35 @@ CallExpr* ContextCallExpr::getRValueCall() {
...
+  } else if (options.length == 3) {
+    // Note: it would be nicer to check retTag to decide between
+    // ref / value versions. However, doing so is challenging because
+    // of the way that iterator functions no longer have the original
+    // retTag.
+    constRefCall = toCallExpr(options.get(1));
+    valueCall = toCallExpr(options.get(2));
+    refCall = toCallExpr(options.get(3));
+  } else {
+    INT_ASSERT(0 && "Bad ContextCallExpr options");
```
benharsh: Did you mean to use something else here, like INT_FATAL? Should the string be a comment instead?

compiler/include/resolution.h
```diff
@@ -146,7 +146,8 @@ void wrapAndCleanUpActuals(ResolutionCandidate* best, CallInfo& info,
 typedef enum {
   FIND_EITHER = 0,
   FIND_REF,
-  FIND_NOT_REF
+  FIND_CONST_REF,
+  FIND_NOT_REF_OR_CONST_REF,
```
benharsh: My brain doesn't parse this well. Is it ``!ref || const_ref`` or ``!(ref || const_ref)`` ? Would an identifier like ``FIND_VALUE`` be better, if that's what it's supposed to do?

test/functions/ferguson/ref-pair/sparse-array-query-irv-locale.chpl
```diff
@@ -0,0 +1,19 @@
+
```
benharsh: I'm not see the value in this test. Since there's no distribution, shouldn't it always print out zero? Doesn't this work on master already? Would a sparse Block array be better?

mppf: It would halt with "attempt to write to zero value" on master.

compiler/include/resolution.h
mppf: It's ``!(ref || const_ref)``. But FIND_VALUE isn't good because there is also TYPE/PARAM ret tags. I'm happy to change the name if you have a suggestion.

compiler/AST/expr.cpp
mppf: I'll change it to INT_FATAL.

compiler/include/resolution.h
benharsh: Ah, OK, makes sense. I don't think I have a better name. Maybe just a comment with that logic, but that's optional.

PR #5564, opened 2017-03-08
Date time fixes for 32 bit systems and valgrind

daviditen:
Fix compile time problem in DateTime module for 32 bit systems.  A 64 bit 'int' was assigned to a 'c_long', which is a 32 bit value on our 32 bit systems.  Add a cast to 'c_long'.  This gets the testTime.chpl test to compile on these systems, but it is still tripping an assert in the test due to an apparent denormalize issue.

Initialize all fields of an extern struct before passing it to an extern function. Valgrind was giving errors about uninitialized values otherwise.

### DIFF
--- a/modules/standard/DateTime.chpl
+++ b/modules/standard/DateTime.chpl
@@ -618,7 +618,7 @@ module DateTime {
     timeStruct.tm_mon = 1;
 
     if tzinfo != nil {
-      timeStruct.tm_gmtoff = abs(utcoffset()).seconds;
+      timeStruct.tm_gmtoff = abs(utcoffset()).seconds: c_long;
       timeStruct.tm_zone = __primitive("cast", c_ptr(c_char), tzname().c_str());
       timeStruct.tm_isdst = dst().seconds: int(32);
     } else {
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Do you have more info on the denormalize issue, i.e. is it documented elsewhere?

daviditen:
I don't have more info about it. I just discovered that the failures would come and go after small changes like adding writeln calls to try to look at the values the asserts fail for, so tried --no-denormalize and found it worked.  I still need to investigate that further.

modules/standard/DateTime.chpl
```diff
@@ -613,12 +620,15 @@ module DateTime {
     timeStruct.tm_sec = second: int(32);
     timeStruct.tm_min = minute: int(32);
     timeStruct.tm_hour = hour: int(32);
-    timeStruct.tm_year = 1900;
+    timeStruct.tm_year = 0;
     timeStruct.tm_mday = 1;
     timeStruct.tm_mon = 1;
 
+    timeStruct.tm_wday = ((new date(1900, 1, 1)).weekday():int(32) + 1) % 7;
```
ben-albrecht: Was this previously occurring automatically as a result of `timeStruct.tm_year = 1900` ?

daviditen: No, it was uninitialized before.  timeStruct.tm_year = 1900 was a mistake - the year is supposed to be 1900 based, so 0 is 1900.

daviditen:
Thanks @ben-albrecht !

PR #5571, opened 2017-03-08
Fix support for default arguments for initializers on generic classes

lydia-duncan:
The issue was that we were creating a wrapper to make the call for the
omitted arguments with the default value for them, but at the time we created
the wrapper, we still didn't know the generic instantiation (since this operation
happens when we find the best candidate, before we resolve Phase 1 of the
body).  So when the default wrapper got resolved, it would complain about the
"this" argument being generic.  But, in the cases where we figure out the generic
instantiation, the call is instead wrapped in a call to "_new" by the end of it, which
has all the same arguments (including their default values), thus ensuring that
the wrappers directly on the initializer never get called, only the wrappers created
for "_new".  So creating the wrappers was unnecessary.

Note that a different strategy will be needed for generic records, but that generic
records are currently prevented by the "super.init()" bug on records and so code
in support of them will not be reachable just yet.

Adds three tests to verify this works, one each for a type, param, and generic
var field.  These all failed on master, but will pass with this branch.

Passed full std/ paratest


lydia-duncan:
Works towards a bug listed in #5198

PR #5572, opened 2017-03-08
Re-enable binary IO optimization on NUMA

benharsh:
By constraining the mData properties by the given domain's
properties we can print out the correct results.

Passed a full local paratest.


benharsh:
@gbtitus : Can you review this one as well? It's much simpler. I believe this will resolve the fasta and reverse-complement performance regressions, though I still need to confirm on the appropriate machine.

gbtitus:
That looks good to me.   👍

PR #5573, opened 2017-03-08
Fix MPI module dereference nil bug

ben-albrecht:
When running a Chapel program with the MPI module in multilocale mode with the gasnet+mpi configuration (more details below), it appears that `CHPL_COMM_WORLD_REPLICATED(1)` becomes freed before reaching the `_initMPI.deinit()` call. As a result, we get the following error `numLocales`-times at the end of the program execution:

```
$CHPL_HOME/modules/packages/MPI.chpl:220: error: attempt to dereference nil
```

Until we can correctly identify the condition in which `CHPL_COMM_WORLD_REPLICATED(1)` needs to be freed or not, this PR introduces a simple fix that omits the `MPI_Comm_Free` call altogether.

The result is a leak size on the order of `numLocales` at the end of MPI multilocale programs for some configurations. Obviously, a leak-free solution would be preferable in the long-term, but this patch reduces the run-time error to a negligible leak for now.

Configuration that encountered the issue:

```bash
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: mpi-gnu *
CHPL_TARGET_ARCH: unknown
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: mpi *
CHPL_TASKS: fifo *
CHPL_LAUNCHER: gasnetrun_mpi
CHPL_MEM: jemalloc

# Using MPICH 12.1.0
MPICH_MAX_THREAD_SAFETY=multiple
AMMPI_MPI_THREAD=multiple
```

Also, the `coforall` over locales will effectively do the same thing as the
else-block for `numLocales == 1`, so this if/else block was dropped.

### DIFF
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -215,12 +215,10 @@ module MPI {
     pragma "no doc"
     proc deinit() {
       if freeChplComm {
-        if numLocales > 1 {
-          coforall loc in Locales do on loc {
-            C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
-          }
-        } else {
-          C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
+        coforall loc in Locales do on loc {
+          // TODO : Don't leak this. This results in dereference nil error for
+          //        some cases, e.g. linux64 multilocale mode with gasnet+mpi
+          //C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
         }
       }
       if doinit {


ben-albrecht:
I'd like to get @npadmana's thoughts on this issue.

npadmana:
@ben-albrecht --- that's interesting. I'm pretty sure I've run cases with a mixed configuration without an issue, so let me take a look. This might have been a bug that crept in at some point in development....

npadmana:
@ben-albrecht --- what determines the order in which a module-scope variable gets freed?

npadmana:
@ben-albrecht --- BTW, I'm "fine" with the idea of leaking this, but I'm wondering if there is something more interesting going on here, where a record accessing a module-level variable has its destructor called after the module-level variable is freed.....

It might be worth seeing if we can create a similar error all of the MPI stuff.....

ben-albrecht:
@npadmana - Out of curiosity, were you able to reproduce the problem, and did you draw the same conclusions I have?

I don't think the record deconstructor getting called early is an issue (double-checked this with some debug prints), but rather, `CHPL_COMM_WORLD_REPLICATED(1)` (which stores `MPI_COMM_WORLD`) is getting freed early.

After testing some different configurations & platforms, I've deduced that the gasnet comm layer is the likely cause of this error (as opposed to any of the other configurations). I suspect gasnet is doing something funny with MPI calls...

npadmana:
@ben-albrecht ---

Yes, I was able to reproduce it. 

Here is a simpler case, that has nothing to do with MPI that fails in exactly the same way ---

```
use UtilReplicatedVar;

var x : [rcDomain] int;
rcReplicate(x, 10);

record Test {
  var y : int;

  proc ~Test() {
    ref y = x(1);
    writeln(y);
  }
}

var t : Test;
writeln("Hello World!");
```

npadmana:
In fact, here's a cleaner version of this ---

```
use UtilReplicatedVar;

var x : [rcDomain] int;
rcReplicate(x, 10);

record Test {
  proc ~Test() {
    writeln(x(1));
  }
}

var t : Test;
```

ben-albrecht:
That simplifies life quite a bit. So it's a `UtilReplicatedVar` + gasnet bug then...

ben-albrecht:
Since this bug has nothing to do with MPI, I'm going to close this issue, and re-open a bug report for distributed arrays & destructors.

Here's another test that gets a little closer to the core of the issue (accessing distributed arrays in destructors):

```chapel

/* foo.chpl

 > chpl --comm=gasnet foo.chpl && ./a.out -nl 1

  $CHPL_HOME/modules/internal/ChapelArray.chpl:2350: error: attempt to dereference nil

 > chpl --comm=none foo.chpl && ./a.out

  segmentation fault

*/

use ReplicatedDist;
use BlockDist;

const rcDomain     = {1..1} dmapped ReplicatedDist();
var x : [rcDomain] int;

const blockDomain = {1..1} dmapped Block(boundingBox={1..1});
var y : [blockDomain] int;

const Domain = {1..1};
var z : [Domain] int;


record Test {
  proc deinit() {
    // x & y are nil by the time we call this deconstructor
    writeln(x);
    writeln(y);
    // z still exists (tested by commenting out other writeln()'s)
    writeln(z);
  }
}

var t : Test;

```

Issue #5341, opened 2017-02-10
What should the semantics be when returning an ArrayView?


benharsh:
## Background

In the process of implementing ArrayViews, we encountered problems when returning an ArrayView (slice, rank-change, or reindex). The following code would either leak or segfault:
```chpl
proc retOrig(A: []) {
  return A;
}
var X : [1..10] int;
writeln(retOrig(X[1..5]));
```

We then realized that we had written ``chpl__autoCopy`` for ArrayViews incorrectly (they in fact did not perform a deep copy). If a deep-copy is performed, we end up with a non-ArrayView type. This causes issues with resolution in our current compiler. The return-type is inferred to be the same type as ``A``, because at the time we haven't decided to insert an autoCopy yet. In discussing with other team members, we realized that we may want to revisit/finalize semantics for returning an aliasing array.

## Should ArrayViews copy-out?

One possible approach to this problem is that ArrayViews should always be copied when returned. This would allow us to more easily determine the return type of the ``retOrig`` example above, and is a consistent rule for users to remember. When considering this option, also consider the following code:
```chpl
config var test = false;

var glob : [1..10] int = 1..10;

proc ret() {
  if test {
    return glob[1..5];
  } else {
    return glob;
  }
}

writeln(ret());
```

My gut reaction is that this should be a valid Chapel program. If we did not copy-out here, then there would be a type mismatch due to the ArrayView implementation.

## Should we allow arrays/ArrayViews to be returned by ``ref`` ?

ArrayViews development progressed very far without running into issues related to returning an ArrayView, suggesting that this isn't a commonly used pattern in our testing suite. That said, could/should we support arrays and ArrayViews to be returned if the ``ref`` return-intent is used?

If one modified the function ``ret`` in the previous code sample to have the ``ref`` return-intent, I believe that the program would no longer be valid. In such a case, I think we should have a better error message than just complaining about a type mismatch.

mppf:
> That said, could/should we support arrays and ArrayViews to be returned if the ref return-intent is used?

We already have code that returns arrays by `ref` or `const ref` and that works OK. But returning a local slice expression (such as `glob[1..5]`) by `ref` or `const ref` seems wrong to me and requires a fair amount of acrobatics in the compiler to support (see `fixupNewAlias` for some of it).

I'm tentatively in favor of the `copy-out` approach generally for ArrayViews.

bradcray:
[...]

PR #5576, opened 2017-03-09
Deprecate array alias

bradcray:
Work in progress...

TODO:
- [ ] resolve remaining failures
- [ ] valgrind testing
- [x] update spec


bradcray:
Here's an enumeration of the failures I'm currently seeing:

C codegen type mismatch
-----------------------
[Error matching compiler output for arrays/ferguson/global-returns]
[Error matching compiler output for arrays/ferguson/semantic-examples/1.5-return-alias-slice-global]

fewer error messages than before
[Error matching compiler output for arrays/ferguson/loc-arr-slice]

non-const ref to a const variable (blank intent not firing correctly?)
-------------------------------------------------------------------
[Error matching compiler output for arrays/ferguson/choose-array/choose-array5]
[Error matching compiler output for arrays/ferguson/choose-array/choose-array6]
[Error matching compiler output for arrays/ferguson/choose-array/choose-array7]
[Error matching compiler output for arrays/vass/const-checking.aliases]
[Error matching compiler output for memory/sungeun/refCount/arrays]
[Error matching compiler output for memory/sungeun/refCount/arraysAndDomains]

segfault
--------
[Error matching program output for arrays/ferguson/semantic-examples/4-global-slice-alias]
[Error matching program output for release/examples/benchmarks/miniMD/miniMD (compopts: 1)]

wrong result
------------
[Error matching program output for arrays/ferguson/semantic-examples/4-return-alias-global]
[Error matching program output for distributions/robust/arithmetic/reindexing/test_array_alias2]

attempt to dereference 'nil'
----------------------------
[Error matching program output for release/examples/benchmarks/miniMD/miniMD (compopts: 2)]

spec needs updating
-------------------
[Error matching program output for release/examples/spec/Arrays/reindexing]

bradcray:
@benharsh determined that the causes of at least some of the failures above are due to a similar issue to the one we've been discussing on #5341.  Namely, when using '=>', the compiler inserts a newAlias() call which does a shallow copy of the record descriptor off of the stack, preserving it even if the original calling context goes away.  By contrast 'ref' does not.  Arguably, this is a bug in the compiler given that we want things like:

````chapel
ref A = B[1..n, 1..n];
````

to work relative to the lifetime of B, not the array view descriptor that describes the slice.  But there is effort to get that working.  We might think of that effort as being similar to inserting newAlias() calls in cases like the above, but a challenge to doing so is that newAlias() is currently inserted in normalize and it's likely we wouldn't be able to detect the cases where it should be inserted until after function resolution.

Tagging @mppf on this due to his interest in issue #5341 and the fact that I think he implemented the newAlias logic (?).

W.r.t. this PR, I feel tempted to merge all of the test changes that pass to master, just to reduce our reliance on '=>' (and because I don't think we want to recreate this effort later).  @benharsh and I were discussing whether or not we should also commit the deprecation warning just to give users a heads-up that it was going away.  I'd originally been thinking I'd improve the deprecation warning so that it would direct users to convert:

````chapel
var A => B;    /* into */  ref A = B;
var A: [D] => B;  /* into */ ref A = B.reindex(D);
````

A challenge with the ref bug noted above is that it makes it hard to tell users when they should expect it to work or not.  Ben's opinion seemed to be that a small enough number of our tests ran afoul of this bug that we shouldn't worry about that and wait for users to hit and report the bug rather than trying to steer them away from it (filing a new issue to track the bug, presumably).

**Comment by mppf** [comment]
Can you show a complete example of the issue? Is refs-to-slices the only issue?

It might not be so hard to have the compiler adjust for slicing, if that's the main issue. See for example fixupNewAlias in callDestructors. Right now, it only works when creating an alias of a slice using the => syntax. I'd presume, but am not certain without seeing more examples, that all we need to do is to adjust that code to recognize and transform

``` chapel
  ref A = B[1..n, 1..n];
```

in a similar way to how it currently handles
``` chapel
  var A => B[1..n, 1..n];
```

bradcray:
In asking for a concrete example, are you wanting to see an excerpt from one of the failing test cases listed above inlined here in the discussion thread?

mppf:
Yes. Just a really short test program demonstrating the problem. I think it would help discussion. Thanks.

bradcray:
I think the following is the one that @benharsh was walking me through (a modified version of arrays/ferguson/semantic-examples/4-global-slice-alias.chpl):

````chapel
module bar {
  var dom = {1..10, 1..10};
  var A : [dom] int;

  ref R = A[1..4, 1..4]; // call_tmp = slice                                   
  // slice freed here                                                           
}

use bar;

proc main() {
  writeln("Setting A[1,1] to 1");
  A[1,1] = 1;
  writeln("R[1,1] is now     ", R[1,1]);

  for r in R do write(r);
  writeln();
}
````

As I understand it, the record representing the A[1..4, 1..4] slice is deallocated at the end of bar's init() function even though A and R are declared at module scope.  To make this work, we'd need to do a copy-out of that record as we do when the declaration is 'var R => A[1..4, 1..4]' via the newAlias() call.

mppf:
Right, fixupNewAlias handled the corresponding cases for `=>` and it took specific work to make that functional. Is the problem only coming up with global variables (i.e. it has to do with global vs local in the init fn)?

mppf:
@bradcray, @benharsh - PR #5692 fixed the segfaults. PR #5738 will fix the errors from the C compiler. I'm testing the second of these now. Assuming testing comes back OK, I think we could try to get this PR in today.

bradcray:
I think that would be great!

PR #5579, opened 2017-03-09
Improve support for paratesting without building our venvs

ronawho:
I install all the python dependencies required for test-venv and chpldoc-venv
locally and set CHPL_TEST_VENV_DIR=none. Since I'm often making third-party
changes I usually clobber of git clean, and installing the venv requirements
locally halves the build time.

test-venv and chpldoc-venv are dependencies for things like man-chpldoc, which
is required for clean paratesting. This just makes it possible to build
everything if CHPL_TEST_VENV_DIR is set. This way test/man/ won't fail.

This also allows our docs to use a locally installed sphinx-build instead of
always using the one in chpl-venv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,10 +104,14 @@ third-party-try-gmp: FORCE
 	fi
 
 third-party-test-venv: FORCE
-	cd third-party && $(MAKE) test-venv
+	-@if [ -z "$$CHPL_TEST_VENV_DIR" ]; then \
+	cd third-party && $(MAKE) test-venv; \
+	fi
 
 third-party-chpldoc-venv: FORCE
-	cd third-party && $(MAKE) chpldoc-venv
+	-@if [ -z "$$CHPL_TEST_VENV_DIR" ]; then \
+	cd third-party && $(MAKE) chpldoc-venv; \
+	fi
 
 test-venv: third-party-test-venv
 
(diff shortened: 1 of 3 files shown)


ronawho:
@ben-albrecht mind reviewing?

compiler/passes/docs.cpp
```diff
@@ -309,7 +309,7 @@ void generateSphinxOutput(std::string sphinxDir, std::string outputDir) {
     CHPL_HOME, "/third-party/chpl-venv/install/", CHPL_HOST_PLATFORM,
     "/py", getChplPythonVersion().c_str(), "/chpl-virtualenv");
   const char * venvBinDir = astr(venvDir, "/bin");
-  const char * sphinxBuild = astr(venvBinDir, "/sphinx-build");
+  const char * sphinxBuild = astr("sphinx-build");
```
ben-albrecht: Question: If user has sphinx installed globally, and also has the chpldoc-venv built, will this resolve to the chpldoc-venv version?

ronawho: chpl-venv version since the path is prepended with the venv bin dir

ben-albrecht: cool, just double checking.

Makefile
```diff
@@ -104,10 +104,14 @@ third-party-try-gmp: FORCE
 	fi
 
 third-party-test-venv: FORCE
-	cd third-party && $(MAKE) test-venv
+	-@if [ -z "$$CHPL_TEST_VENV_DIR" ]; then \
```
bradcray: @ronawho / @ben-albrecht : Why do we use a `@` here? (which causes failures to be ignored)  In nightly testing, this seems to sweep real problems under the rug temporarily until we actually try to run the test system later (e.g., see issue #395).  If there's no good reason for it, I'm inclined to remove it.

ronawho: I'm not sure, and I don't remember Makefile changes well enough to have an answer. Usually I'm copying from some other makefile. Feel free to change it if you think it'd be valuable.

bradcray: Oops, I got my Makefile symbols backwards.  The `-` ignores errors, the `@` squashes printing the command to stdout.

bradcray: Opened PR #13698 to address this.

PR #5580, opened 2017-03-09
Permit Block distributions to defer the specification of their bounding box

bradcray:
This change permits Block distributions to defer the specification of their bounding boxes, in which case the first domain declared over the block distribution will form the bounding box.  This permits declarations like the following:

````chapel
const D = {1..n, 1..n} dmapped Block({1..n, 1..n});
````

to be written as:

````chapel
const D = {1..n, 1..n} dmapped Block(rank=2);
````

The rank still has to be specified since the Block distribution will need to set up its storage and that of the domains declared over it based on that value.

TODO:
- [ ] update documentation as necessary
- [ ] linux64 testing
- [ ] gasnet testing


bradcray:
Note that this pattern could be cleaned up further by introducing a factory utility as follows:

````chapel
const D = newBlockDomain({1..n, 1..n});
````

but I haven't gone down that path here, in part because of concern that supplying such a routine will make it far less likely that users will create domains with shared distributions which will ultimately be useful for compiler alignment-based optimizations.

bradcray:
Closing this for the time being, having captured it in issue #5938.

bradcray:
Tried testing this today and got many failures including surprisingly simple cases like stream and RA.  Closing until I have time to look into it more.

PR #5583, opened 2017-03-09
Add support for divide by zero checks

### LINKED ISSUE #5476: param-guarded divide-by-zero test?
Twice in the past year, I've had users get reasonably confused/frustrated by hitting divide-by-zero errors in their Chapel programs (in both cases, in code that they did not write) where they showed up as exceptions.  Given that in our default 'checks' mode we check for seg faults and OOB errors to head off similar exceptional conditions, should we put a check for zero divisor in our / definitions in ChapelBase that are disabled with a --no-div-zero-checks flag (controlled by --no-checks and therefore --fast)?

bradcray:
Twice in the past year, users have run into a certain amount of confusion due to getting execution-time integer divide-by-zero errors which we don't handle gracefully.  This change adds an execution time check for divide by zeroes that can be disabled via a new --no-div-by-zero-checks flag (itself controlled by the --no-checks flag).  It also unifies the error messages for param divide-by-zero checks, updates the man page, and adds a test to lock in the check.  As a clean-up, it also re-alphabetizes the --no*checks flags on the --help message and man page, which had been lost when --no-cast-checks was added.

This resolves #5476 

TODO:
- [x] update man page
- [x] run testing


test/compflags/checks/checkDivByZero.execopts
```diff
@@ -1,2 +1,4 @@
 --testExecCases=0  # checkDivByZero-intvar.good
 --testExecCases=1  # checkDivByZero-uintvar.good
+--testExecCases=2  # checkDivByZero-intDivEquals.good
```
noakesmichael: I think GIT has noted that this execopts file does not have a carriage return

PR #5586, opened 2017-03-09
Update time comparison operators to avoid floating point

daviditen:
Using floating point in the time comparison operators led to some problems on
linux32, so switch to using integers only.  This should fix the testTime failure on linux32.

While there, noticed some unnecessary nil checks, so removed them.


gbtitus:
I believe this is proper -- go for it.   👍

daviditen:
Thank you @gbtitus !

PR #5591, opened 2017-03-09
Move new primality graph to its own suite

lydia-duncan:
It didn't make sense to force the contributor of the primality test to
worry so much about which suite the graph should go in, so I
resorted it myself after the merge.

Tested correctness and performance graph generation

### DIFF
--- a/test/GRAPHFILES
+++ b/test/GRAPHFILES
@@ -288,6 +288,7 @@ users/npadmana/twopt/twopt-buildtrees.graph
 users/npadmana/twopt/twopt-paircount.graph
 # suite: Elegance
 studies/elegance/elegance.graph
+# suite: Math operations
+patterns/prime_generator/prime.graph
 # suite: Samples (bogus)
 Samples/Performance/foo.graph
-patterns/prime_generator/prime.graph


bradcray:
It seems unfortunate to create a suite for a singleton test.  Maybe 'misc' would be a better category for this test?  (at least until it has a sufficient number of siblings that one would want to dial up that suite?)

lydia-duncan:
We already have suites for singleton tests, though.  There's an example right above it

bradcray:
Perhaps, but I think we should stop doing it.  I think we've already got way too many suites that are not used/useful in practice and should generally be reducing the number of them, not increasing them (at least, until the current suites get too full).

lydia-duncan:
That seems reasonable to me.  Alright if I wait to do this until after the performance team meeting?  I was going to bring up this new test there anyways

bradcray:
Yeah, no rush.  This philosophy question would be a good topic.

lydia-duncan:
For those following along at home, I'm going to make a "scalar performance" suite for this test and some of our other tests that fall under that category.

PR #5600, opened 2017-03-09
Make paratest.server work within salloc on Cray systems

dmk42:
The paratest.server script is designed to ssh into the worker nodes when inside an salloc.  This works for a cluster, but on a Cray X-series system, it tries to ssh into the compute nodes.  Turning off the slurm detection when on a Cray X-series machine allows it to work under salloc there.  This is a two-line change to do that.

@mppf and @ronawho are likely to be interested in looking at this.

### DIFF
--- a/util/test/paratest.server
+++ b/util/test/paratest.server
@@ -682,7 +682,8 @@ sub main {
             }
         }
     } else { # else, just current node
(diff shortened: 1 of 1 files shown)


dmk42:
Based on approval from @ronawho and successful tests, I'll go ahead and merge.

PR #5601, opened 2017-03-09
Reorganize performance suites to avoid latest singleton

lydia-duncan:
This
- removes the top-level suites "Chapel versus C comparisons", "Statement
  Comparisons" and "Math operations"
- creates a new top-level suite "Serial Performance" and two sub-suites,
  "Chapel versus C Comparisons" and "Statment Comparisons"
  - moves entirety of old top level suite "Chapel versus C comparisons" into
    suite of same name, and moves entirety of "Math operations" into the same
    suite
- moves some of "Statement Comparisons" into a new suite, "Parallel Statement
  Comparisons"
  - moves the rest of the old "Statement Comparisons" suite into the new
    sub-suite of the same name.

Verified output with --generate-graphs and that no .graph was missing after the
commit.

### DIFF
--- a/test/GRAPHFILES
+++ b/test/GRAPHFILES
@@ -274,7 +274,7 @@ users/npadmana/twopt/twopt-buildtrees.graph
 users/npadmana/twopt/twopt-paircount.graph
 # suite: Elegance
 studies/elegance/elegance.graph
-# suite: Scalar Performance
+# suite: Serial Performance
 patterns/prime_generator/prime.graph
 statements/lydia/forCompare.graph
 statements/lydia/whileCompare.graph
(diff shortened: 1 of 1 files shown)


test/GRAPHFILES
```diff
@@ -288,7 +274,37 @@ users/npadmana/twopt/twopt-buildtrees.graph
 users/npadmana/twopt/twopt-paircount.graph
 # suite: Elegance
 studies/elegance/elegance.graph
-# suite: Math operations
+# suite: Scalar Performance
```
ben-albrecht: - Would 'serial performance' be a more fitting title?

lydia-duncan: Ugh, we totally said that and for some reason I wrote down "scalar" . . .

ben-albrecht: You had it right: "scalar" was said in the meeting, but I think "serial" is a better title.

ben-albrecht:
Oh - be sure to update PR description

PR #5604, opened 2017-03-10
Improve LICM effectiveness for iterator classes

benharsh:
LICM currently checks if any symbols used in the loop
are formals with a ref intent. If that's the case, then it will
not hoist them out of the loop. We observed a performance
regression in the views-forall mini-benchmark for rank-change
arrays, and tracked the cause down to a lack of hoisting due
to the way in which an iterator class was used.

@ronawho and I feel that we can safely side-step this formal
check in the case that the formal is an iterator class. We think
that this is a safe approach because in order to see changes
synchronization would be required to comply with our memory
model, and LICM does not attempt hoisting if such constructs
are present in the loop.

While other tests also generated less-than-ideal loops, the
rank-change loop involved more work (reading blk(rank)).

TODO: Long-term we'd like to revisit this ArgSymbol check
in its entirety, but we're close to the release and this seems
like a fairly targeted change.

Testing:
- [x] full local
- [x] full gasnet

### DIFF
--- a/compiler/optimizations/loopInvariantCodeMotion.cpp
+++ b/compiler/optimizations/loopInvariantCodeMotion.cpp
@@ -858,19 +858,22 @@ static void computeLoopInvariants(std::vector<SymExpr*>& loopInvariants, Loop*
     // Note that not all things that are passed by ref will have the ref intent
     // flag, and may just be ref variables. This is a known bug, see comments
     // in addVarsToFormals(): flattenFunctions.cpp.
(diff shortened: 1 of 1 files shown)


ronawho:
Think the PR message could use a little more info about "we think it's safe because syntonization with an atomic/sync would be required and licm doesn't run if synchronization constructs are detected" -- or something to that effect

Looks good code wise to me though

PR #5609, opened 2017-03-10
Overlay nightly graphs onto releaseOverRelease graphs

ronawho:
When looking at the releaseOverRelease graphs we often open up the nightly
graphs at the same time so that we can see if regressions are within the
nightly noise or if it's something real to investigate. This allows us to view
releaseOverRelease and the nightly results on the same graph to simplify
releaseOverRelease performance triage.


ronawho:
@bradcray Mind giving this a quick look if you're still around? Don't think the perl changes need much review, but feel free to take a look at http://chapel.sourceforge.net/perf/chapcs/releaseOverRelease/

It's basically the same as the preview I sent out before, but the nightly numbers are hidden by default, and the annotations should work.

PR #5613, opened 2017-03-10
Remove bashism from clean/cleanall/clobber targets

mppf:
I've been working on an Ubuntu 17.04 system and there, Makefile commands
run with /bin/sh. A result of that is that an if statement like

     if [ -a doc/Makefile ]

results in a syntax error, since -a is a bash-ism.

This commit simply changes it to use -e which as far as I can tell does
exactly the same thing (does the file exist?) and is compatible with
/bin/sh.

Passed full local testing.
Reviewed by @ben-albrecht.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ clean: FORCE
 	cd modules && $(MAKE) clean
 	cd runtime && $(MAKE) clean
 	cd third-party && $(MAKE) clean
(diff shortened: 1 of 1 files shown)


mppf:
@ben-albrecht  might be interested as the original author of these lines.

@awallace-cray, could you double-check my work here?

Issue #5614, opened 2017-03-10
ref-pair precedence question


mppf:
### Summary of Problem

Currently, ref-pairs are resolved regardless of other precedence choices. This might be confusing.

After PR #5561, the behavior of this program changed, because const ref vs value return is now included in ref-pair calculations in resolveNormalCall. However, the issue started with the first versions of ref-pair support in the compiler.

The question is - if normal resolution rules would select the ref/const ref/value version in preference to the other version, should that always be used? Should ref-pair behavior only be triggered if there would otherwise be an ambiguity? It's not currently the case as this test demonstrates.

One reason the answer to this question isn't entirely obvious is that generic functions with a matching `where` clause are normally selected in favor of generic functions without a `where` clause. That pattern comes up in the array accessor methods, where the `ref` return version does not have a `where` clause but the value or const ref return versions do.

### Steps to Reproduce

In the below program, one might expect the child name functions to always return "Child" but they do not currently.

**Source Code:**

```chapel
var parentName = "Parent";
var childName = "Child";

class Parent {
  // ref + const ref
  proc name const ref return parentName;
  proc name2 ref return parentName;

  // const ref + value
  proc name3 const ref return parentName;
  proc name4 return parentName;

  // ref + value
  proc name5 return parentName;
  proc name6 ref return parentName;
}


class Child : Parent {
  // ref + const ref
  proc name ref return childName;
  proc name2 const ref return childName;

  // const ref + value
  proc name3 return childName;
  proc name4 const ref return childName;

  // ref + value
  proc name5 ref return childName;
  proc name6 return childName;
}


writeln("Calling child name functions");
var c = new Child();
writeln(c.name);
writeln(c.name2);
writeln(c.name3);
writeln(c.name4);
writeln(c.name5);
writeln(c.name6);

writeln("Calling parent name functions");
var d = c:Parent;
writeln(d.name);
writeln(d.name2);
writeln(d.name3);
writeln(d.name4);
writeln(d.name5);
writeln(d.name6);
```

**Associated Future Test(s):**
test/functions/ferguson/ref-pair/across-scopes.chpl

benharsh:
Would this case work correctly if we required overriding functions to have the same return-intent(s)?

mppf:
Parentheses-less methods aren't overridden or subject to virtual dispatch. So this case is more about coecing from Child to Parent.

benharsh:
[...]

PR #5616, opened 2017-03-10
Rename name field from NumaDomain and add tests

mppf:
The name field is confusingly named because the parent class locale
already has a parentheses-less method named `proc name`.

The recent PR #5561 changed function resolution in a way that revealed
this issue.

I'm not sure the language should support for such overloading, but even
if it does, I don't think we should use it in NumaDomain because we'll
just confuse ourselves.

So, this PR simply renames the name field to ndName.

It adds two futures. One future tracks the exact issue we were running
into with NumaDomain and is related to const-checking in a constructor.
The second future asks about how ref-pairs should be resolved when one of
the elements of the ref-pair would normally have lower resolution
precedence (e.g. requires coercion, in a different scope, etc).

Trivial, not reviewed.

### DIFF
--- a/modules/internal/localeModels/numa/LocaleModel.chpl
+++ b/modules/internal/localeModels/numa/LocaleModel.chpl
@@ -54,27 +54,27 @@ module LocaleModel {
   //
   class NumaDomain : AbstractLocaleModel {
     const sid: chpl_sublocID_t;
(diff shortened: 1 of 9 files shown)


mppf:
This relates to isssue #5614.

PR #5617, opened 2017-03-10
Fix future (Thanks Ben H!)

mppf:
Ben noticed a trivial problem in the test itself, so this fixes it.

Verified test/functions/ferguson/ref-pair/across-scopes.chpl has clean match against .bad file.


mppf:
Relates to issue #5614.

PR #5618, opened 2017-03-10
Make _this formals of record type be ref-if-modified

### LINKED ISSUE #5266: inconsistency for record blank intents, this, and const

mppf:
Closes #5266.

* Adjusts blankIntentForThisArg in resolveIntents to treat
  records similarly to arrays

* Adds two tests. One is a future because late const checking is
  currently disabled.

Passed full local testing.

### DIFF
--- a/compiler/passes/resolveIntents.cpp
+++ b/compiler/passes/resolveIntents.cpp
@@ -144,8 +144,14 @@ static IntentTag constIntentForThisArg(Type* t) {
 static IntentTag blankIntentForThisArg(Type* t) {
   // todo: be honest when 't' is an array or domain
 
+  Type* valType = t->getValType();
+
+  // For user records or types with FLAG_DEFAULT_INTENT_IS_REF_MAYBE_CONST,
+  // the intent for this is INTENT_REF_MAYBE_CONST
+  //
   // This applies to both arguments of type _ref(t) and t
-  if (t->getValType()->symbol->hasFlag(FLAG_DEFAULT_INTENT_IS_REF_MAYBE_CONST))
+  if (isRecord(valType) ||
+      valType->symbol->hasFlag(FLAG_DEFAULT_INTENT_IS_REF_MAYBE_CONST))
     return INTENT_REF_MAYBE_CONST;
 
   if (isRecordWrappedType(t))  // domain / distribution
--- a/test/functions/ferguson/ref-pair/record-this-intent-const-error.chpl
+++ b/test/functions/ferguson/ref-pair/record-this-intent-const-error.chpl
@@ -1,0 +1,15 @@
+
+
+record R {
+  var field:int;
+
+  proc setsField() {
+    field = 1;
+  }
+}
+
+const globalR = new R();
+
+globalR.setsField();
+
+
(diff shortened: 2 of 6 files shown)


benharsh:
Looks OK to me. Could/should the 'default intent is ref if modified' pragma be removed from the _array record?

mppf:
No, because that applies to all array formals (not just _this formals).

PR #5619, opened 2017-03-10
Fix ref-pair bug

### LINKED ISSUE #5614: ref-pair precedence question

mppf:
Closes #5614.

PR #5561 introduced a bug. When working with ref-pairs, resolveNormalCall
includes a series of calls to isBetterMatch that select ref/value
versions in cases where one is a better match than the other.

PR #5561 extended the ref-pair logic to treat const ref and value return
intents differently, but in doing so, it didn't extend this logic to call
isBetterMatch to choose between ref and non-ref versions.

Passed full local testing.


benharsh:
Looks fine to me. Does this fix any existing futures?

mppf:
Yeah, the one I just added ;P

PR #5620, opened 2017-03-10
Change a record destructor into a deinit

daviditen:
At @lydia-duncan's request, changed an old style destructor into a new
style deinit() in the DateTime module.


lydia-duncan:
:+1: Thanks!

PR #5624, opened 2017-03-10
Improve cullOverReferences and enable late const checking

mppf:
This PR improves const-checking in cullOverReferences and enables error messages for const errors from the lateConstChecking function. In order to do so, it enables the analysis in cullOverReferences to work by tuple elements instead of just by Symbols. That means that, blank intent for tuples of array elements will include whether or not that tuple element was modified. For example,
``` chapel
proc f(tup) {
  A[1] = 1;
  writeln(B[1]);
}
var A:[1..4] int;
const B:[1..4] int;
f( (A,B) ); // ok
f( (B,A) ); // error
```

In order to facilitate this analysis, and because I believe that ref-to-const is a useful property for passes after cullOverReferences, I added a field called fieldQualifiers to Symbol. This tracks the Qualifier for reference fields in a tuple.

I tightened up `in` vs `const in` for integers.

I removed an old workaround from functionResolution's formalRequiresTemp.

I made many bug fixes to the analysis in cullOverReferences.

Test updates:
 *  statements/vass/index-variable-const-errors.chpl  appears to rely on incomplete tuple const-ness analysis in functionResolution in one case. I disabled this incomplete analysis to get other programs working. I moved the case that no longer works to a new future.
 * performance/vectorization/vectorizeOnly/constChecking/standaloneConstChecking.chpl  had its .bad file updated for the same reason.
 * studies/madness/aniruddha/madchap/MRA.chpl sets const fields in a function other than the constructor. That's not expected to be OK with initializers. I made these fields `var` since I didn't see any downside to doing so.
 * studies/shootout/submitted/fasta.chpl is now no-tested because it modifies a const variable by passing it to sumProbs.
 * some clean-ups in test/functions/ferguson/ref-pair

Future Work:

 * Remove the workaround code in lateConstCheck for `this` arguments, since record _this is now ref-if-modified.
 * Extend the analysis in cullOverReferences to apply to all tuple variables or do other analysis to mark tuple variables as elements const / not const.
 * Remove the workaround code in  lateConstCheck for tuple construction, once the analysis handles all tuples.
 * move gatherLoopDetails to another file and make sure it works by adding --verify checks.

Review & Testing:

 * full local testing
 * futures testing
 * valgrind examples/primers testing
 * baseline examples/primers testing
 * GASNet examples/primers testing
 * hellos with --baseline --verify
 * hellos with --verify
 * quickstart, -futures testing


mppf:
@benharsh - could you look at this? I think it's ready for review & improves on other PRs you reviewed.

Thanks!

compiler/resolution/cullOverReferences.cpp
```diff
@@ -1003,17 +1336,20 @@ void cullOverReferences() {
...
 
-                    collectedSymbols.push_back(index);
+                    GraphNode srcNode = makeNode(index, indexTupleElement);
+                    collectedSymbols.push_back(srcNode);
                     revisit = true;
-                    revisitGraph[index].push_back(sym);
+                    addDependency(revisitGraph, srcNode, node);
                     handled = true;
-                    break;
+                    //break;
```
benharsh: Is it useful to keep this ``break`` in a comment?

compiler/resolution/functionResolution.cpp
```diff
@@ -5554,6 +5579,7 @@ resolveCoerce(CallExpr* call) {
 //
 static bool backendRequiresCopyForIn(Type* t) {
   return (isRecord(t) ||
+          isUnion(t) ||
```
benharsh: More of an observation than a critique of this PR: I wonder if we should revisit this logic as Qualifiers become more widely used. I'm guessing the result of this function may be different in the _ref type world.

mppf:
I'm planning to merge this on Wednesday.

PR #5635, opened 2017-03-11
 Fix compile-time numa issue

benharsh:
Fix compile-time failure where I incorrectly passed a 2*int to a max function


ronawho:
@benharsh just FYI: I was doing numa testing for the qt-use-chpl-alloc stuff, and I'm still seeing failures for test/arrays/bradc/idxTypes/diffIndexTypes.chpl and  test/arrays/elliot/emptyArray.chpl

benharsh:
Darn, I'll take a look.

PR #5637, opened 2017-03-11
Have qthreads use Chapel's allocator

ronawho:
Use the new qthreads 1.12 `--with-alloc=chapel` feature to have qthreads
allocate memory with Chapel's allocator instead of the system allocator.

This should result in some nice performance improvements for multi-locale when
a registered heap is being used (ugni, gasnet-{fast,large}). With this, qthread
stacks will be in the registered heap, so we won't have to trampoline through a
registered segment.

I don't expect this to have a major performance impact for single-locale.
Qthreads uses memory pooling for stacks (and most things) so using jemalloc
over the system allocator shouldn't have too much of an impact.

Implementation notes:
---------------------
To have qthreads use Chapel's allocator, we need to ensure that jemalloc is
built before qthreads. We also need to add --with-alloc=chapel to qthread's
configure step, and make sure we provide the right include and link args so
that qthreads can find our allocation routines and link against jemalloc.

Correctness testing:
--------------------
I've done pretty extensive correctness testing:
 - linux64: full std, CHPL_MEM=cstdlib, and numa
 - gasnet : full gasnet-mpi, gasnet-fast, and gasnet-numa
 - cray-xc: release/examples for ugni-qthreads and gn-aries-qthreads
 - extra  : full 300 trial run of std and gasnet-fast

I've also used `nm -A` to look through libqthread.a to ensure there's no calls
to the system allocator for flat(nemesis) and numa(distrib). I didn't find any
calls to malloc, calloc, valloc, realloc, free, aligned_alloc, posix_memalign,
or memalign so I'm pretty confident that we're in good shape as far as not
mixing calls between our allocator and the system allocator.

Performance results:
--------------------
I've also looked at some preliminary performance results for ugni-qthreads:

This change results in improved multi-locale performance by avoiding
trampolining for memory in task stacks. This had a larger impact before the
array-views and ref-if-modified work, both of which reduced communication in
some benchmarks that really benefited from this. Currently I expect this to
result in a 10% improvement for HPL, MiniMD, and fft. Previously this PR
brought ugni-qthreads in parity with ugni-muxed, but that's not quite the case
now (fft and HPL are still lagging by ~20%.) An upcoming PR that increases
QT_SPINCOUNT on Crays should result in ugni-qthreads beating ugni-muxed for fft
and HPL (the last benchmarks where muxed was winning.)


ronawho:
@gbtitus I'm going to merge this today since performance/correctness testing has been pretty quiet, but I think this needs a careful review. Mind taking a look?

gbtitus:
This looks good to me.  👍

PR #5639, opened 2017-03-11
Fix type error in ChapelRange

benharsh:
For some reason the binary IO optimization for DefaultRectangular exposed an issue with a type conflict due to ``chpl__extendedEuclid`` returning ``int(32)`` into an ``int(8)``. The strided type can be a variety of integer widths, so for now safeCast to ``strType`` to avoid the compile-time error.

Testing:
- [x] full numa
- [x] full local


benharsh:
@ronawho : This should fix the errors you pointed out in the other PR.

ronawho:
👍 thanks

PR #5646, opened 2017-03-13
Increase QT_SPINCOUNT when running on a Cray

ronawho:
If you're using a Cray it's likely that your application is going to be pretty
parallel. This increases the amount of spinning we do on crays before putting a
worker to sleep.

With this ugni-qthreads should now beat ugni-muxed for all remaining benchmarks:
 - fft gets ~.090 GFlops (50% faster than ugni-muxed)
 - hpl-release gets .0028 GFlops (15% faster than ugni-muxed)

### DIFF
--- a/runtime/src/tasks/qthreads/tasks-qthreads.c
+++ b/runtime/src/tasks/qthreads/tasks-qthreads.c
@@ -685,6 +685,13 @@ static void setupWorkStealing(void) {
     chpl_qt_setenv("STEAL_RATIO", "0", 0);
 }
 
+static void setupSpinWaiting(void) {
+  const char *crayPlatform = "cray-x";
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus I'm doing some additional perf runs to see if I want to go with 3,000,000 vs 30,000,000. Wanted to see if you're ok with this idea overall in the meantime though.

Ideally I'd like to set `QT_WAIT_POLICY=active`, but that's not implemented yet: https://github.com/Qthreads/qthreads/issues/38

ronawho:
@gbtitus Going to stick with 3,000,000. Ok to merge?

ronawho:
And longer term we probably want to do the same thing for `CHPL_COMM!=none` assuming we're not oversubscribed or doing local spawning.

gbtitus:
Just finished looking at it -- yes, okay to merge.  👍

PR #5648, opened 2017-03-13
Remove DR array standalone iterator for !defRectSimpleDData

vasslitvinov:
Before this change we had DefaultRectangularArr standalone iterator
applicable when !defRectSimpleDData, which redirected to the
DefaultRectangularDom standalone iterator. However the latter
has a where clause restricting it to defRectSimpleDData==true.
So the former failed to compile and the leader/follower iterator
was silently chosen instead.

This PR removes the array standalone for !defRectSimpleDData.
Once we write the domain standalone for !defRectSimpleDData,
we can put the array standalone back in, or perhaps just remove
the defRectSimpleDData==true restriction on the existing one.

The goal of treating the !defRectSimpleDData case separately is to
avoid the multi-ddata overhead from being seen in the single-ddata
case (flat locale model).

While there, I changed the where clause on the domain standalone
iterator from !localeModelHasSublocales to defRectSimpleDData.
These are identical, however the former is consistent with
the other uses throughout the file.

This change does not affect any behavior at the moment and
I am not adding any tests for it. It will affect behavior and be tested
by the upcoming ForallStmt PR.

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -183,7 +183,7 @@ module DefaultRectangular {
                ignoreRunning = dataParIgnoreRunningTasks,
                minIndicesPerTask = dataParMinGranularity,
                offset=createTuple(rank, idxType, 0:idxType))
-      where tag == iterKind.standalone && !localeModelHasSublocales {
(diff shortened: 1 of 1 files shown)


benharsh:
Looks fine to me. Hopefully we can work on the numa standalone case post-1.15.

PR #5652, opened 2017-03-14
Stop limiting numtrials for multi-locale MiniMD, SSCA2, and hpl.hpcc

ronawho:
Stop limiting the number of trials for MiniMD, SSCA2, and hpl.hccc multi-locale
performance runs:

 - MiniMD is way faster than it used to be, so stop limiting to 1 trials.
 - SSCA2 no longer takes over 5 minutes for ugni+muxed, remove timeout
 - hpl.hpcc2012 is fast for gn-aries since the jemalloc switch, remove timeout


ronawho:
@benharsh any objections?

benharsh:
Looks good 👍

PR #5653, opened 2017-03-14
Stop limiting the number of ugni communication domains under slurm

ronawho:
When "native" slurm was first added, we couldn't get exclusive access to all
the network resources, so we had to limit how much memory we registered with
the NIC. This is no longer an issue, so we don't have to limit the number of
communication domains under slurm.

Note that we're still limiting the default heap size under slurm to 16GB
instead of 2/3 available memory. This could be changed too, but the smaller
size seems to be working for us, and it improves startup time. This is
something to revisit at another time.

Increasing the number of communication domains results in some pretty dramatic
performance improvements for some latency bound benchmarks. I'm seeing a 25%
speedup for SSCA, 50% speedup for Lulesh, and a 75% improvement for RA:

| Benchmark | Old Perf     | New Perf     |
| --------- | ------------ | ------------ |
| SSCA#2    | 111 sec      | 85 sec       |
| Lulesh    | 4.7 sec      | 3.3 sec      |
| RA-atomic | .075 GUPS    | .12 GUPS     |
| RA-rmo    | .040 GUPS    | .075 GUPS    |

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -366,13 +366,6 @@ typedef struct mem_map_t_struct {
 static mem_map_t* mem_map_map;
 
 //
-// See chpl_comm_desired_shared_heap() and compute_comm_dom_cnt().
-// This is a conservative estimate of the total amount of memory we
-// can register with the NIC when running under slurm.
-//
-#define MAX_MEM_REG_UNDER_SLURM  (((size_t) 240) << 30)
-
-//
 // This is the memory region for the guaranteed NIC-registered memory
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus this needs cleanup and more testing, but I wanted to give you an early ping so you know it's coming.

gbtitus:
This looks fine and the performance rebound is nice.  👍 

Either the task-qthreads change should be in a different PR or the title and message on this one should be adjusted, since that's not just a "while I was here" change.

ronawho:
This is currently based on #5646 (which adds the qt spincont stuff), once that's merged I'll rebase.

gbtitus:
Looks good -- okay to merge.

PR #5654, opened 2017-03-14
Plug leaks caused by arrays-of-arrays

bradcray:
While puzzling over leaked listNode() classes I started digging into some of the primer examples that were leaking arrays and domains and put together the following fix which reduces, for example, test/release/examples/primers/arrays.chpl to zero leaks without introducing valgrind errors.  It similarly takes the test/users/bugzilla/bug794133/ directory down to zero leaks with no valgrind errors.  I still need to do full testing on it, but so far it looks promising!

The approach taken here is: When we're processing elements of arrays-of-arrays w.r.t. memory processing, when we're releasing the elements, to mark their domains as being able to be freed once all their arrays are freed.


bradcray:
Shoot, valgrind testing turned up 9 failures + 3 timeouts:

````
[Error matching program output for classes/bradc/arrayInClass/genericArrayInClass-arith]
[Error matching program output for classes/bradc/arrayInClass/genericArrayInClass-otharrs]
[Error matching program output for distributions/robust/arithmetic/basics/test_array_of_distributed_array1]
[Error matching program output for distributions/robust/arithmetic/basics/test_array_of_distributed_array2]
[Error matching program output for distributions/robust/arithmetic/basics/test_distributed_array_of_array2]
[Error matching program output for distributions/robust/arithmetic/basics/test_distributed_array_of_record2]
[Error: Timed out executing program parallel/bundles/bundles]
[Error: Timed out executing program reductions/vass/reductions-stress-mf]
[Error matching program output for release/examples/benchmarks/miniMD/explicit/miniMD]
[Error matching program output for studies/colostate/bradc/cfd-mini-serial-blc]
[Error: Timed out executing program users/npadmana/twopt/do_smu_soa]
````

**Comment by bradcray** [comment]
The thing that the errors above seem to have in common is the following pattern:

````chapel
const D = {1..3};

class C {
  var A: [1..10] [D] real;
}

var myC = new C();

delete myC;
````

That is, when we delete myC and see that there are no more arrays based on D, we delete D even though it's still a live variable.

I'm now convinced that if I could adjust master to get three simple cases to be leak-free and valgrind-error-free then this would be mergeable and a step forward.  The above is case 1.  Case 2 is the same but with a local/anonymous domain:

````chapel
class C {
  var A: [1..10] [1..3] real;
}

var myC = new C();

delete myC;
````

And the third is:

````chapel
var A: [1..10] [1..3, 1..3] real;
````

**Comment by bradcray** [comment]
With  more investigation, I think that difference between case 1 and case 2 above is that the compiler seemingly never inserts an autodestroy for the anonymous {1..3} domain used to declare the inner array dimension, and that if it did so, the code would be leak-free without my playing the module code games that I was attempting to in this PR.  This makes me wonder whether having normalize hoist that anonymous domain into its own temp variable would solve the problem...

I think the current code base assumes that the module code would free the domain in this case, but the trouble is that the module code can't distinguish between the named and anonymous cases (that information is lost), so either leaks in the anonymous case (what's happening on master today) or prematurely frees the named case (what's happening on my branch which, in retrospect, was naive... it was effectively saying "this domain is out of scope now" in the "used as an inner array's domain context" without knowing anything about the domain).

**Comment by bradcray** [comment]
Thinking about this a bit more, I find myself wondering what it is about the outer anonymous domain in:

````chpl
var A: [1..3, 1..3] [1..10] real;
````

that makes it get freed while the inner domain doesn't?  Does it get an autodestroy / get moved into a temp?  Or does it get flagged in some other way as being anonymous and able to be freed when the array is?  How is this case distinguished from:

````chapel
var A: [D] [1..10] real;
````

where D presumably would not be freed before it left scope?

**Comment by mppf** [comment]
I'm investigating these cases.

For 
``` chapel
var A: [1..10] [1..3, 1..3] real;
```

I can see it leaks memory, but it does not leak memory if I put that declaration into a function. AFAIK it can only leak memory when it's for a global variable. I havn't investigated why it leaks in this case but I might've added the leak myself to work around other issues. Either way, leaking the global variable's domain seems low-priority to me.

``` chapel
class C {
  var A: [1..10] [1..3] real;
}
var myC = new C();
delete myC;
```

literally (as a test) does not leak memory for me, and neither does

``` chapel
const D = {1..3};
class C {
  var A: [1..10] [D] real;
}
var myC = new C();
delete myC;
```

I'm much more concerned about other memory-management errors than these ones (such as the leak from the compiler-generated default constructor for generic `var` or `const` fields).

Issue #4977, opened 2016-12-07
Implement an error handling model


psahabu:
### awaiting implementation
- [ ] standard library
  - [ ] investigate/remove `try!` and `halt()` from module code (#8452)
  - [ ] deprecate out error= functions
    - dependent on defer/deinit strategy, error memory management to store errors in vars
    - strategy:
      - rename and privatize them
      - change internal calls to the new name
      - make a wrapper with the old name, calling the privatized version
      - insert a compiler warning into the wrapper
    - [ ] update #7281
  - [ ] `try! this.lock();` -> `this.lock(); defer this.unlock();`
- [ ] add error handling to mason
- [ ] throwing from non-inlined iterators (#7134) 
  - MF note: fixing this will require possibly significant development effort (at least 1 day, maybe a week, maybe worse.) Also, one of my recent PRs added it-throws-inline.chpl and it-throws-noinline.chpl as tests.
  - [ ] ItemReader.these and channel.matches should be `throws` (turned off to work around failures in --baseline)
  - potential cause for memory leaks in the following tests:
    - `errhandling/parallel/forall-range-throws`
    - `errhandling/parallel/forall-throws-subtasks`
    - `errhandling/parallel/forall-throws`
- [ ] incorrect line number from uncaught error due to inlining (#7297)
- [ ] ensure `catch` clause lists go from specific to general
- [ ] initializer error handling (#6145)
  - should be possible to clean up partially initialized state in phase 1
  - could either call deinit or manually clean up initialized state in phase 2

### pending design
- [x] runtime errors
  - working conclusion: orthogonal to system resilience, waiting for use cases
- [ ] make `writeThis()` a throwing method?
  - MF: More difficult than initially thought, I/O code uses deferred syserr, would need defer.
- [ ] error memory management (#6428)
  - [ ] future: utilize Owned/Shared
  - note: Owned is a useful pattern for class cleanup on error
- [ ] should errors interact with stack tracing?
  - idea: `halt` raises an uncatchable error, which stores the stack trace

### testing and bugs
- [ ] add Curl testing (#7570)
- [ ] investigate failures on pgi (#5331)

### documentation
- [ ] chpldoc and prototype modules (#7489)
- [ ] update doc/rst/developer/bestPractices/ErrorWarningMessaging.txt
- [ ] reflect changes in CHIP 8
- [ ] language spec changes

# Overview

[...]

PR #5655, opened 2017-03-14
class Error, catch blocks, and strict mode

psahabu:
This is the next step for the error handling feature. linux64+flat tests pending.

### What syntax is being lowered?
- catch: a control flow structure that handles errors when paired with try or try!
  - catch (catches all errors, without access to the error)
  - catch err (catches all errors, with named access to the error)
  - catch err: ErrType (catches errors of a given subtype, with named access)
- Note that errors are not automatically deallocated when consumed by a catch.
  - This will be in a future PR and is tracked in #4977.

### Lowering example
```
// given code
proc propagate() throws {
  try {
    a(); // throws
    b(); // does not throw
    c(); // throws
  } catch e: SubError {
    f();
  } catch e: AnotherSubError {
    g();
  }
}

// after this pass
proc propagate(out error_out: Error) {
  var error: Error;
  a(error);
  if error then
    goto handler;
  b();
  c(error);
  if error then
    goto handler;

  label handler:
  if error {
    var e = error: SubError;
    if _cast {
      f();
    } else {
      var e = error: AnotherSubError;
      if e {
        g();
      } else {
        // set and return
        error_out = error;
        goto epilogue_label;
      }
    }
  }
}
```

### `Error` implementation
`module ChapelError` has been included in `ChapelStandard`. It contains `class Error`, which the error handling system uses as its base class. This is done through `dtError`, which is used in `passes/errorHandling.cpp`.

### `catch` implementation

When a `try` block handles an error, we jump to the `handler` label using a `goto`. Inside that label, we ensure that the error is set and then iterate through the `catches` list using `PRIM_DYNAMIC_CAST` and conditionals. If a catch-all block is not present, `try` will attempt to propagate the error while `try!` will halt. The appropriate policy will be inserted as a catch-all.

While implementing this feature, it was found that codegen improperly generated dynamic casts because ternary operators do not have sufficient precedence. @mppf fixed this by adding those parentheses during codegen.

### strict mode implementation
When the `--strict-errors` flag is passed to the compiler, all throwing calls must be marked with `try` or `try!`. This is implemented by a simple compiler flag that is used in `passes/errorHandling.cpp`.

### Additional changes
- Updated parser for `catch` syntax
- Added `CatchStmt`
- Update `TryStmt` to include `CatchStmts`
- De-futurized error handling tests
- Added tests for error cases
- Cleaned up testing in general


compiler/AST/CatchStmt.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CatchStmt.h"
+#include "AstVisitor.h"
+
+// TODO: do we need this build?
```
mppf: Let's either remove the comment or call new CatchStmt from the parser. I think it's fine to call a ::build routine from the parser on principle.

compiler/include/CatchStmt.h
```diff
@@ -0,0 +1,56 @@
...
+
+  void                accept(AstVisitor* visitor);
+  void                replaceChild(Expr* old_ast, Expr* new_ast);
+  Expr*               getFirstChild();
+  Expr*               getFirstExpr();
+  GenRet              codegen();
+  DECLARE_COPY(CatchStmt);
+
+  // this BlockStmt contains two parts:
+  // 1. an (optional) filter expression.
```
mppf: The filter expression is optional in the parser but it is not optional in the AST. Let's remove the word "(optional)" here.

compiler/AST/CatchStmt.cpp
```diff
@@ -0,0 +1,82 @@
...
+  return toDefExpr(_body->body.first());
+}
+
+BlockStmt* CatchStmt::body() const {
+  return toBlockStmt(_body->body.last());
+}
+
+void CatchStmt::accept(AstVisitor* visitor) {
+  if (visitor->enterCatchStmt(this)) {
+    _body->accept(visitor);
```
mppf: I would expect this to visit the expr() and then the body() rather than visiting _body.
Could you try that and keep it if it works?

compiler/AST/TryStmt.cpp
```diff
@@ -18,20 +18,38 @@
...
+BlockStmt* TryStmt::build(bool tryBang, BlockStmt* body, BlockStmt* catches) {
+  return buildChplStmt(new TryStmt(tryBang, body, catches));
 }
 
 BlockStmt* TryStmt::buildChplStmt(Expr* expr) {
   return new BlockStmt(expr, BLOCK_SCOPELESS);
 }
 
-TryStmt::TryStmt(bool tryBang, BlockStmt* body) : Stmt(E_TryStmt) {
+TryStmt::TryStmt(bool tryBang, BlockStmt* body, BlockStmt* catches) : Stmt(E_TryStmt) {
```
mppf: It would be nice to have a comment here to indicate that catches are stored in a BlockStmt to make parsing more convenient.

compiler/AST/TryStmt.cpp
```diff
@@ -52,30 +70,61 @@ void TryStmt::accept(AstVisitor* visitor) {
...
-Expr* TryStmt::copy(SymbolMap* map, bool internal) {
-  return NULL;
-}
-
-Expr* TryStmt::copyInner(SymbolMap* map) {
-  return NULL;
+TryStmt* TryStmt::copyInner(SymbolMap* map) {
+  TryStmt* copy = new TryStmt(_tryBang, COPY_INT(_body), NULL);
+  for_alist(c, _catches) {
+    copy->_catches.insertAtTail(c->copy());
```
mppf: I believe this should use `COPY_INT(c)` rather than `c->copy()`. Could you try it & accept it if your tests pass?

compiler/AST/TryStmt.cpp
```diff
@@ -52,30 +70,61 @@ void TryStmt::accept(AstVisitor* visitor) {
...
+  }
+  return copy;
 }
 
 void TryStmt::replaceChild(Expr* old_ast, Expr* new_ast) {
-
+  if (old_ast == _body) {
+    _body = toBlockStmt(new_ast);
+  } else if (old_ast->list == &_catches) {
```
mppf: I don't believe this old_ast->list condition is necessary although I can't explain why. None of the other replaceChild implementations with ALists take any action for members of the AList. E.g. CallExpr::replaceChild and FnSymbol::replaceChild.

Could you try removing it and accept the change if everything looks OK?

compiler/include/TryStmt.h
```diff
@@ -26,25 +26,31 @@ class TryStmt : public Stmt
...
-  Expr*               copy(SymbolMap* map = NULL, bool internal = false);
-  Expr*               copyInner(SymbolMap* map);
   void                replaceChild(Expr* old_ast, Expr* new_ast);
   Expr*               getFirstChild();
   Expr*               getFirstExpr();
   Expr*               getNextExpr(Expr* expr);
   GenRet              codegen();
+  DECLARE_COPY(TryStmt);
+
+  BlockStmt*          _body;
```
mppf: I'd add a comment next to _body here that it is intended to be private but it needs to be accessed by astutil.h macros.

compiler/parser/chapel.ypp
```diff
@@ -704,8 +708,29 @@ if_stmt:
...
+                           { $$ = buildChapelStmt(); }
+| catch_stmt_ls catch_stmt { $1->insertAtTail($2); }
+;
+
+catch_stmt:
+  TCATCH catch_expr block_stmt { $$ = CatchStmt::build($2, $3); }
+;
+
+
+/* TODO: would be nice to parenthesize a catch */
```
mppf: you mean like
```
try {
  ...
} catch (e:Error) {
 ...
}
```

Did we run into a problem enabling that, once we resolved the other parser issues?
(Not a big priority for this PR, I just couldn't remember if we tried it once we had other stuff working).

compiler/passes/errorHandling.cpp
```diff
@@ -20,14 +20,55 @@
...
+#include "CatchStmt.h"
 #include "stmt.h"
 #include "symbol.h"
 #include "TryStmt.h"
-#include "view.h"
 
 #include <stack>
 #include <vector>
 
+/*
```
mppf: I like to see such listings showing what the pass is supposed to do. However, this comment needs a sentence or two of introduction since it might seem out of the blue to a reader. It also needs to mark what the input and output of the pass are.

compiler/passes/errorHandling.cpp
```diff
@@ -65,7 +107,12 @@ ErrorHandlingVisitor::ErrorHandlingVisitor(ArgSymbol*   _outError,
 bool ErrorHandlingVisitor::enterTryStmt(TryStmt* node) {
   SET_LINENO(node);
 
-  VarSymbol*   errorVar     = newTemp("error", dtObject);
+  // TODO: nested try
+  if (!tryStack.empty()) {
+    USR_FATAL("nested try is not yet supported");
+  }
```
mppf: I am curious, do you think you can get nested try working soon? I think we have the right framework here... consider it a challenge :)

mppf:
@psahabu - I think this is looking good and I think it's reasonable to consider me the reviewer.
I'd like to see the following happen before merge:
 - [ ] make the minor changes requested
 - [ ] run `start_test test/errhandling -valgrind` and `start_test test/errhandling -compopts --baseline`
 - [ ] run full local testing

compiler/AST/CatchStmt.cpp
```diff
@@ -0,0 +1,82 @@
...
+    visitor->exitCatchStmt(this);
+  }
+}
+
+CatchStmt* CatchStmt::copyInner(SymbolMap* map) {
+  DefExpr* expr = toDefExpr(_body->body.first());
+  BlockStmt* body = toBlockStmt(_body->body.last());
+  return new CatchStmt(COPY_INT(expr), COPY_INT(body));
+}
```
vasslitvinov: Let's instead use expr() and body() in ::copyInner().

compiler/AST/TryStmt.cpp
vasslitvinov: replaceChild on an AList member works automatically. It uses the member's prev/next/parent pointers. The member's/AList's enclosing class should not do anything about its ALists in its replaceChild.
So, what Michael said.

compiler/AST/TryStmt.cpp
```diff
@@ -52,30 +70,61 @@ void TryStmt::accept(AstVisitor* visitor) {
...
+  }
+
+  return retVal;
+}
+
+GenRet TryStmt::codegen() {
+  codegenStmt(this);
+
+  return _body->codegen();
 }
```
vasslitvinov: Please add a comment explaining how _cathces are codegen'ed.

compiler/AST/view.cpp
```diff
@@ -26,6 +26,7 @@
 #include "AstDump.h"
```
vasslitvinov: Why did view.cpp need any of these changes?

If they are not needed, please revert view.cpp,
to reduce the diffs.

compiler/codegen/expr.cpp
```diff
@@ -1810,7 +1810,7 @@ GenRet codegenTernary(GenRet cond, GenRet ifTrue, GenRet ifFalse)
 #endif
 
   if( info->cfile ) {
-    ret.c = "(" + cond.c + ")?(" + ifTrue.c + "):(" + ifFalse.c + ")";
+    ret.c = "( (" + cond.c + ")?(" + ifTrue.c + "):(" + ifFalse.c + ") )";
   } else {
```
vasslitvinov: Please add a "while there" note to the PR message about this fix.

compiler/include/TryStmt.h
vasslitvinov: I am dealing with the same issue in my ForallStmt, which Michael commented on. To me it is a bummer to let AST_CHILDREN_CALL prevent a field from being private. Given that we are already using accessor methods instead of fields for BlockStmt (see ex. initBlockGet() et al.), I favor doing that for TryStmt as well, keeping its _body private.
Feel free to leave things as-is, if you feel strongly.

vasslitvinov: Oh I see Preston just made the opposite change. In this case, I will let you decide.

Just keep in mind that "intended to be private" != "private". :)

compiler/parser/chapel.ypp
vasslitvinov: Given that Chapel does not require parens ex. in if/while conditions, I would like to allow e:Error without parens.

Separately, again to keep with Chapel spirit, I'd like to propose "do STMT;" as an option instead of block_stmt for the catch body. (Surely not in this PR.)

compiler/passes/errorHandling.cpp
vasslitvinov: Also I think the top "try {" line needs to be added.

compiler/passes/errorHandling.cpp
```diff
@@ -194,16 +292,18 @@ CallExpr* ErrorHandlingVisitor::haltExpr() {
   return new CallExpr(PRIM_RT_ERROR, new_CStringSymbol("uncaught error"));
 }
 
-// TODO: dtObject should be dtError, but we don't have that right now
 void lowerErrorHandling() {
+  INT_ASSERT(dtError->inTree());
+  INT_ASSERT(dtError->getRefType());
```
vasslitvinov: I suggest removing the getRefType() assertion. In the future 'ref' formals will not require it. In the present, let the code fail instead at the point where the ref type is asked for.

compiler/resolution/functionResolution.cpp
```diff
@@ -30,6 +30,7 @@
 #include "build.h"
 #include "caches.h"
 #include "callInfo.h"
+#include "CatchStmt.h"
```
vasslitvinov: If functionResolution compiles/works without CatchSTmt.h, let's remove this #include.

compiler/AST/TryStmt.cpp
mppf: Actually it's probably better to make TryStmt::codegen() just run INT_FATAL. These should be lowered and removed before codegen.

compiler/parser/chapel.ypp
mppf: @vasslitvinov - actually only the no-parens version parses right now; the TODO is to allow it with parens as I showed above.
``` chapel
try {
  ...
} catch e:Error {
  ...
}
```

parses now.

compiler/passes/errorHandling.cpp
mppf: Fine with me.

compiler/AST/CatchStmt.cpp
psahabu: I vaguely remember trying this and it not working, but I'll give it another shot.

mppf: If it does not work, just but a comment to say you were having trouble with that... but that it is what we would expect it to do.

psahabu: Okay, it seems to work over `test/errhandling`. I'll run it in flat testing and see how it goes.

compiler/AST/TryStmt.cpp
psahabu: This seems to work, I've put it in.

compiler/include/TryStmt.h
psahabu: Yeah, this is disappointing, but piping AST_CHILDREN_CALL through accessor methods was brittle and we'd like to stay consistent. We can fix it in v2. :)

compiler/AST/TryStmt.cpp
psahabu: This works, I've put it into the change.

compiler/parser/chapel.ypp
psahabu: @mppf We never tried it. I'll look into it after nested tries.

compiler/passes/errorHandling.cpp
psahabu: In addition to the intro, I also expanded the example to a full-blown function.

compiler/passes/errorHandling.cpp
psahabu: Yep, I've got a branch going on this. Shooting for the end of the week.

compiler/AST/CatchStmt.cpp
psahabu: Ah, this is leftover from when I removed `expr()` and `body()`. That was a bad time.

compiler/AST/TryStmt.cpp
psahabu: I went with `INT_FATAL` and had it return `GenRet(-1)`. Also applied it to `CatchStmt`.

compiler/AST/view.cpp
psahabu: Ah, I have my vim config set up to clean up trailing whitespace in files that I touch. Since I needed to add the include, it cleaned up the whitespace as well.

Is that alright or would you still prefer that I revert?

compiler/codegen/expr.cpp
psahabu: I included this under the `catch` implementation section, since it actually inhibited the proper function of `catch`.

compiler/passes/errorHandling.cpp
psahabu: Done.

compiler/resolution/functionResolution.cpp
psahabu: We need it because of `AST_CHILDREN_CALL`.

psahabu:
Alright, I've rebased onto master and am running the full flat paratest now. Ran into issues with valgrind that I'm looking into.

compiler/AST/view.cpp
vasslitvinov: If CatchStmt.h is indeed needed, then it's fine to remove whitespace as well.

psahabu: It's needed for `AST_CHILDREN_CALL` and other macros.

compiler/AST/TryStmt.cpp
mppf: I have no idea what GenRet(-1) is. I'd normally just return an empty one, like this
GenRet ret;
return ret;

PR #5659, opened 2017-03-14
Accept libhugetlbfs page size setting; compute heap page size as and when needed.

gbtitus:
There are two primary changes here.  The first is that when computing
the heap page size, if we find the HUGETLB_DEFAULT_PAGE_SIZE environment
variable set then we simply accept that as the page size, rather than
doing any more work to see if a larger size seems to work.  This saves a
bit of time on Cray X* systems but much more importantly, it avoids the
page touching needed in order to explore for larger page sizes.  That
page touching could cause undesirable NUMA localization of some of the
memory in comm layer desired shared heaps.

The second change is to compute the heap page size the first time it is
requested, rather than doing it in an explicit initialization step ahead
of time.  This relieves us needing to move the init call around to make
sure it stays ahead of the first call to get the heap size it computed
as the runtime evolves.  (This was going to be an issue with an upcoming
change to NUMA-localize comm layer desired shared heaps.)  As a side
effect of this change I reworked how we decide whether to bring in the
pthreads.h #include file.

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -349,19 +349,12 @@ size_t chpl_getHeapPageSize(void) {
     if (pthread_once(&onceCtl, computeHeapPageSize) != 0) {
       chpl_internal_error("pthread_once() failed");
     }
-  }
-#elif defined __CYGWIN__
-  return chpl_getSysPageSize();
-#else
-  //
-  // Note: no single-threaded initialization for non-POSIX OS.
-  //
-  if (heapPageSize == 0) {
-    computeHeapPageSize();
-  }
-#endif
 
-  return heapPageSize;
+    return heapPageSize;
+  }
+#else
+  return chpl_getSysPageSize();
+#endif
 }
 
 
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -35,10 +35,20 @@
 #include "error.h"
 
 // System headers
+
+// We need this first so we can then decide based on the existence and
+// perhaps the value of _POSIX_VERSION what else to #include below.
+#ifdef __unix__
+#include <unistd.h>
+#endif
+
 #include <errno.h>
 
-#ifdef __linux__
+#ifdef _POSIX_VERSION
 #include <pthread.h>
+#endif
+
+#ifdef __linux__
 #include <sched.h>
 #endif
 
@@ -87,11 +97,8 @@ size_t chpl_getSysPageSize(void) {
   return pageSize;
 }
 
-static size_t heapPageSize = 0;
 
-size_t chpl_getHeapPageSize(void) {
-  return heapPageSize;
-}
+static size_t heapPageSize = 0;
 
 static volatile unsigned char* check_page_size_ptr;
 static unsigned char* check_page_size_base;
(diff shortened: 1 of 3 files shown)


gbtitus:
@mppf, would you review this please?  The original heap page size computing code was yours.  @ronawho, you may want to take a look as well.

runtime/src/chplsys.c
```diff
@@ -299,40 +306,60 @@ static void computeHeapPageSizeByGuessing(size_t page_size_in)
...
+
+    if (pthread_once(&onceCtl, computeHeapPageSize) != 0) {
+      chpl_internal_error("pthread_once() failed");
+    }
   }
 #else
-  pageSize = chpl_getSysPageSize();
+  //
+  // Note: no single-threaded initialization for non-POSIX OS.
+  //
```
mppf: I wonder if we should print a warning in this case? Or use an atomic or something? It seems if we end up in this case, it could have disasterous performance consequences.

mppf:
@gbtitus - this is fine with me, other than my question, and supposing you have / will test appropriately before merging. I agree with the strategy here as a fix for the NUMA localization issue.  Thanks!

runtime/src/chplsys.c
gbtitus: Are you thinking that disastrous performance consequences could occur if, say, somehow two threads called for the heap page size at the same time and then interfered with each other's computations of that?  I don't think that can happen: the calls to get the heap page size only occur during runtime initialization when we are still running in the process (no concurrency). If it really is a worry, though I could just recode to return `chpl_getSysPageSize()` for the non-POSIX case, which I think is only cygwin.
(We can't use atomics because we'd have to protect ourselves against someone using the "locks" implementation of atomics which is based on sync vars, since this code runs way before the tasking layer and its sync var support is initialized.)

mppf: If it's just cygwin, I'd prefer we disable the guessing behavior entirely by returning chpl_getSysPageSize().

I'm worried about performance consequences of doing the mmap game in a call we expect (generally) to basically amount to a load (of the page size).

gbtitus: Okay, I'll change the non-POSIX (cygwin) side of this to just return the system pagesize.  Where we'll still do the mmap game is when there is a comm layer desired shared heap but no hugepage module and thus the `HUGETLB_DEFAULT_PAGE_SIZE` env var is not set.  So for example: the gasnet+aries case with tasks=muxed you originally wrote the "guessing" code to handle.

gbtitus:
The testing has been lightweight but I think the coverage is sufficient, because this is the sort of thing that creates disaster quickly if it's functionally broken.  This only really affects the "comm layer has desired shared heap" situation, so on vanilla linux64 I just ran the hellos.  On Cray XC I ran hello.chpl in all configurations combining (comm=ugni, gasnet over mpi, and gasnet over aries) with (tasks=qthreads and muxed), and for the comm=ugni ones I ran with and without a hugepage module loaded.  For the muxed runs I stuffed a debug print into muxed tasking to see the heap page size it got back from chplsys, and for the qthreads runs I just set QT_INFO=2 so it would tell me whether it was making guard pages, since that is predicated on having small (non-huge) heap pages.  I have run on cygwin yet, but I will before merging.

Issue #5657, opened 2017-03-14
inadvertently using a non-distributed domain


mppf:
### Summary of Problem

If the type is specified for a domain and it is initialized by an expression with a dmapped clause, it is surprising that the resulting domain is not distributed. The program below compiles and runs without error but does not give any indication that the line containing `dmapped` is problematic. In particular, that line is assigning a distributed domain to a non-distributed one. Perhaps a warning could be generated in this case.

### Steps to Reproduce

**Source Code:**

```chapel
  use BlockDist;
  config const n = 100;
  var myDomain: domain(2) = {1..n, 1..n} dmapped Block({1..n, 1..n});
  for d in myDomain {
    writeln(d.locale.id);
  }
```

bradcray:
I don't agree that the line above is nonsensical, though I agree that it likely reflects a misunderstanding on the user's part, and one that real users have run afoul of at times.

Why I don't think it's nonsensical:  We permit domain assignment between domains of distinct distributions, and the definition is that the result is a copy of the index set, not of the implementation (its domain map / distribution).  So the statement above is legal and sensical even though it may not reflect the programmer's intention.  This property is why it's legal to do:

````chapel
var myDomain: domain(2) dmapped Block(...) = {1..n, 1..n};
````

even though the two domain maps don't match.

A language change we could consider would be to change the interpretation of:

````chapel
var myDomain: domain(2) = ...;
````

from:

````chapel
var myDomain: domain(2) dmapped DefaultDomainMap = ...
````

to:

````chapel
var myDomain: domain(2) dmapped [InheritDistFromInitExpr] = ...
````

which would be somewhat analogous to our treatment of formal array types like:

````chapel
proc foo(X: [1..n] real) { ... }
````

where we interpret this not as "the actual must be a default rectangular" but rather "the actual must have this index set and we'll inherit its domain map" (making the function generic w.r.t. domain map).

Alternatively, I feel like if we were to generate a warning in this context as you suggest ("you're initializing a non-distributed domain using a distributed domain -- do you mean to do that?  If not, add a dmapped clause to your type expression or remove the type expression altogether."), that would probably be sufficient as well.

bradcray:
[...]

PR #5661, opened 2017-03-14
Add warning for initializing local dom from distributed one

mppf:
Fixes issue #5657.

* adds flags FLAG_DSI_DISTRIBUTED and FLAG_DSI_NOT_DISTRIBUTED and mark modules/dists and internal DSI implementations (like DefaultRectangular) with the appropriate one
* add code in checkResolved to find the problematic case and raise a warning on it


mppf:
Closing until I complete more testing...

bradcray:
I'm not wild about requiring a pragma to be placed on every distributed domain map to make this work.  It seems like we already have a way to distinguish between distributed- and non-distributed domain maps, though I need to think to remember what it is.

I'd also expect this warning to be able to be implemented strictly with module-level code (and prefer it if so) -- for example, a param check in the array-to-array assignment routine (or initialization routine if/when we have one?).

mppf:
@bradcray - I agree it would be nicer if we could do it in domain assignment / array assignment routines. I'd need some help knowing how we know if a domain is distributed though. Or maybe we need to add `dsiIsLocalOrDistributed` or something that returns a param to the DSI interface...

mppf:
Oh, and to be clear, distributed domain maps that don't have the pragma (in the approach in this PR) would still function - they just would not get this warning.

mppf:
Maybe `isDefaultRectangular` can help here.

bradcray:
That could be a good start today, though there may be other layouts over time (and if you count things like sparse, there already are).  I'd prefer to see this go in the direction of having domain maps support an 'isDistribution' vs. 'isLayout' style of query which seems like a useful property to know in general.

PR #5662, opened 2017-03-14
Add specialized overload for c_ptrTo for DefaultRectangular

benharsh:
ArrayViews over a DefaultRectangular array store enough information in
their indexCache to determine whether the array can be passed as a
single _ddata. The logic for getting that pointer remains the same, but
the runtime check will utilize the indexCache when possible.

Also fixes a compile-timer error in the general c_ptrTo([]) case for
ArrayViews over other distributions (e.g., BlockDist).

Testing (thanks @ben-albrecht !):
- [x] full local
- [x] full numa
- [x] gasnet release-examples


modules/internal/CPtr.chpl
```diff
@@ -223,8 +223,26 @@ module CPtr {
...
-    if !arr._value.oneDData then halt("error: c_ptrTo(multi_ddata array");
+  inline proc c_ptrTo(arr: []) where isRectangularArr(arr) && !arr._value.isDefaultRectangular() {
+    if !chpl__getActualArray(arr).oneDData then halt("error: c_ptrTo(multi_ddata array");
+    return c_pointer_return(arr[arr.domain.low]);
+  }
+
+  pragma "no doc"
+  inline proc c_ptrTo(arr: []) where arr._value.isDefaultRectangular() {
+    const val = arr._value;
+    if chpl__isArrayView(val) {
```
bradcray: If I'm reading the above right, array views on DRs return 'true' for 'isDefaultRectangular()' which seems surprising to me (but is also a pre-existing condition before this PR if I'm not mistaken).  I'd think that it should return false for array views and that the filter here should be more like "isDR || isArrayViewThatWrapsDR".

W.r.t. that note and your TODO in the code below, it seems to me that the better way to implement c_ptrTo() for arrays would be to have its body call a method on the array itself that would be implemented on _array in terms of a dsi routine that each domain map could them specialize.  That would put this logic where it belongs (in the array implementation) rather than way out here in CPtr.  It would also avoid resorting to a ._value access outside of the domain map code.

benharsh: Both of your points/suggestions seem reasonable to me. Would you like either of them to be implemented before this PR goes in?

bradcray: For the CPtr -> domain map refactoring, it doesn't matter to me deeply whether that's part of this PR or a follow-on.

For making isDefaultRectangular not return true for views, I'd do that as a separate PR (though whether it goes in before or after this one doesn't matter to me).  I'm thinking I probably should've noticed this issue earlier, but somehow didn't...

benharsh: OK, I think I'll work on those after this PR goes in.

Either of you have any other feedback, or can I merge?

bradcray: I'm good with it.

benharsh: Were you thinking that each domain map would implement the portion that returns, or the part that does the runtime check?

bradcray: Both, I think (though I may be misunderstanding what checks you're referring to).  I was imagining something like:

````chapel
proc c_ptrTo(arr: []) {
  return _array.cPtrTo();
}

proc _array.cPtrTo() {
  return _value.dsiCPtrTo();
}

// and then a per domain-map dsiCPtrTo() routine that would either generate an error or return the pointer
````

The idea being to get rid of a lot of where clauses and cases in this c_ptrTo() routine and have each domain map answer the question itself.

Comment (benharsh) [inline modules/internal/CPtr.chpl]: Currently we really only need to specialize when checking if the array can be expressed as a single _ddata. They can all return the right thing by doing this:
```chpl
return c_pointer_return(arr[arr.domain.low]);
```

It may also be somewhat inconvenient to implement this as a method. For example, for DefaultRectangular the ArrayView already stores the necessary info to check whether it can be expressed as a single ddata. If we wrote this as a method, we'd have to do something like:
```chpl
chpl__getActualArray(arr).dsiToCPtr(chpl__getViewDom(arr));
```
And potentially recompute all the metadata within that method. What do you think?

benharsh: Offline we felt that this could wait until after 1.15. The discussion continues in #5667.

PR #5666, opened 2017-03-14
Suppress threadring failures on bradc-lnx

ronawho:
Performance runs of threadring use ~5GB ram ((503 tasks*8MB stacks)+misc)

Bradc-lnx only has 6 GB total (2GB physical, 4GB swap) so we're hitting OOMs
with threadring since we started having qthreads use our allocator with #5637.

When qthreads was using the system allocator the pages weren't being faulted
in, so we were only using virtual memory, but with jemalloc it seems were
actually using the full amount of physical memory. I still need to dig into
this more, but for now suppress the failures.


lydia-duncan:
Looks good, assuming you've checked it works!

ronawho:
Lame, apparently we don't support directory wide SUPPRESSIFs. Switching to indvididual .suppressifs

PR #5668, opened 2017-03-14
Optimize serial iteration for ArrayViews

benharsh:
When ArrayViews were merged we were no longer able to take
advantage of an optimization for serial iteration over a multi-ddata
array. This resulted in poor performance for STREAM.

This PR implements a serial iterator, named ``chpl__serialViewIter``, that
domain map authors can overload to optimize serial iteration over
ArrayViews for their fancy array.

The DefaultRectangular overload is very similar to the old serial iterator
implemented on DefaultRectangularArr. A main difference is that it will
utilize the indexCache from an ArrayView. This is possible due to the
similarities between _remoteAccessData and DefaultRectangularArr's
fields and methods.

Testing:
- [x] full numa
- [x] full local
- [x] full gasnet-flat
- [x] full gasnet-numa

### DIFF
--- a/modules/dists/BlockDist.chpl
+++ b/modules/dists/BlockDist.chpl
@@ -1122,7 +1122,7 @@ proc BlockArr.nonLocalAccess(i: rank*idxType) ref {
       pragma "no copy" pragma "no auto destroy" var myLocRAD = myLocArr.locRAD;
       pragma "no copy" pragma "no auto destroy" var radata = myLocRAD.RAD;
       if radata(rlocIdx).shiftedDataChunk(0) != nil {
-        var dataIdx = radata(rlocIdx).getRADDataIndex(myLocArr.stridable, i);
+        var dataIdx = radata(rlocIdx).getDataIndex(i);
         return radata(rlocIdx).getDataElem(dataIdx);
       }
     }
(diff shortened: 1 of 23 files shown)


benharsh:
@gbtitus and @bradcray : this PR restores the optimized serial iterator under NUMA. It resolves the block-stream regression. The matrix-multiply tests are helped a bit, but are still lagging behind for unknown reasons.

gbtitus:
Nice, @benharsh.  This looks good to me.  👍

benharsh:
@gbtitus : I pushed a commit that updates some improved comm counts. I'm running a gn-numa paratest to catch other comm count changes.

@bradcray : any objections to this PR? OK to merge today?

benharsh:
I think the 'convert index' helpers could definitely be abstracted away more cleanly.

I'm not really bothered by most of the other uses, since this is an iterator designed to optimize ArrayView iteration and such iterators will need to reason about ArrayView functionality.

bradcray:
Oh, I probably didn't make it clear that when I used the term 'view' in this context, I wasn't thinking of it as being specific to the ArrayView* (TM) classes.

For example, I could imagine that perhaps in a follower iterator on a distributed array that was (necessarily) iterating over a sub-chunk what it owned, it might also be able to call into these helper routines, passing along its own subdomain or view argument to say which indices it needed to access, and avoiding a general dsiAccess() call (not that I was expecting anything like that to happen in this PR... more thinking of setting up iterator helpers that took arrays and domains in separately and knew how to iterate over the subarray efficiently without having to create a slice or use dsiAccess).  Of course, Block arguably already does this by using the slice and then following your new code path, but my guess (maybe incorrect) is that other contexts may want to (or be able to) do this without creating an explicit slice.

I'm maybe getting pretty abstract here, though (and again, I'm not suggesting this is a reason to hold up the PR in any way).

PR #5392, opened 2017-02-21
BLAS level 2

ben-albrecht:
This PR adds BLAS 2 support to the Chapel BLAS module.

**Progress**

- Implementation
    - [X] BLAS 2 routines
    - [X] Additional routine assertions, e.g. square, symmetric, Hermitian matrices...
- Tests
    - [X] Simple tests (1 per routine)
    - [X] Full testing coverage (op, uplo, diag, lda, ..)
    - [X] Dense (general) matrix routines
    - [X] Symmetric matrix routines
    - [X] Hermitian matrix routines
    - [X] Triangular matrix routines
- Documentation
    - [X] Simple documentation (name, equations)
    - [X] MKL URLs instead of netlib (for all BLAS routines)
- Testing
    - [X] Passes OS X + netlib BLAS testing
    - [X] Passes Cray + CrayBLAS testing

**Slated for a future PR**

- Tests
    - Band matrix routines
    - Packed matrix routines
    - Cleaner error messages replacing where-clauses
    - Tests for error-cases (halts/compiler-errors)

PR #5669, opened 2017-03-15
Deprecate ldA argument in BLAS module

ben-albrecht:
With the addition of array views, we can now compute the `ldA` values automatically in all cases, rendering the argument obsolete. 

In this PR:
- modify and overload `getLeadingDim()` to handle array views
- remove `ld*` as an argument in the BLAS routines and tests
- document this change in the Chapel BLAS documentation.

**Deprecation warning** -- I have opted to *not* provide a deprecation warning for the following reasons:
- BLAS was added in 1.14.0 as a WIP package module (BLAS 3 only)
- The `ldA` argument was already optional and would only be necessary if users were passing an array slice to a BLAS 3 routine.

I am open to including a deprecation warning if my reviewer feels strongly about it.

Some less noteworthy changes in this PR:
- Add `config param isBLAS_MKL`, so that users don't have to symlink their MKL BLAS header when using BLAS MKL (yuck!)
- Slightly modified the `gemv` array slice test to check that passing an array sliced on both dimensions works (made possible by #5662 -- thanks @benharsh !)
- Remove `conjg` BLAS helper routine, since `Math.conjg` has been updated to always return the same type it receives.
- Add next steps in a module-level TODO comment (requested in #5392 as a follow-up)

- [X] OS X + netlib
- [x] Cray XC + CrayBLAS

### DIFF
--- a/modules/packages/BLAS.chpl
+++ b/modules/packages/BLAS.chpl
@@ -28,6 +28,9 @@ reference version of BLAS, this documentation refers to the
 `MKL BLAS <https://software.intel.com/en-us/node/520725>`_ documentation, due
 to quality and interface similarities.
 
+This module is intended to work with non-distributed dense rectangular arrays.
+
+
 Compiling with BLAS
 -------------------
 
@@ -105,6 +108,10 @@ dimensions of the matrices from the arrays that are passed in, even when one is
 passing in a sub-array such that the array elements are not contiguously stored
 in memory.
 
+.. warning::
+
+  The ``CHPL_LOCALE_MODEL=numa`` configuration is not supported by this module.
+
 .. MKL Documentation References
 
 .. _GEMM:   https://software.intel.com/en-us/node/ae8380b9-cac8-4c57-9af3-2eaac6acfc1b
@@ -163,6 +170,7 @@ in memory.
   - Clearer compiler errors instead of using where-clauses
   - Modular implementations using config param-wrapped require statements
   - More consistent documentation
+  - Support banded/packed matrix routines
 
 */
 module BLAS {
--- a/modules/packages/BLAS.chpl
+++ b/modules/packages/BLAS.chpl
@@ -163,10 +163,9 @@ The native BLAS interface can still be accessed by calling routines from the
 .. _CABS1:  https://software.intel.com/en-us/node/64961e94-92d0-4671-90e6-86995e259a85
 
 .. BLAS Module TODO:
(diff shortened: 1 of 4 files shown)


ben-albrecht:
May be of interest to @npadmana

modules/packages/BLAS.chpl
```diff
@@ -162,6 +159,11 @@ The native BLAS interface can still be accessed by calling routines from the
 .. _AMIN:   https://software.intel.com/en-us/node/50dceaa5-3463-402f-8065-a48bc68e0888
 .. _CABS1:  https://software.intel.com/en-us/node/64961e94-92d0-4671-90e6-86995e259a85
 
+.. BLAS Module TODO:
+  - Clearer compiler errors instead of using where-clauses
+  - Modular implementations using config param-wrapped require statements
+  - More consistent documentation
```
benharsh: I'd add a note somewhere in this top-level documentation (unless I've missed one) that this module is only intended to work with non-distributed dense rectangular arrays.

benharsh: And maybe a ``warning::`` regarding NUMA?

ben-albrecht: Sure. Should we provide an umbrella warning such as "`CHPL_LOCALE_MODEL=numa` is not supported by this module", or explicitly state what might be problematic (which I think is array slicing?).

benharsh: I think it's multi-ddata in general that is problematic, at least for medium/largish arrays. I think what you have is fine for the release. We can see how multi-ddata/numa evolves over the 1.16 cycle and reevaluate later.

npadmana:
@ben-albrecht -- 

1. I agree that we don't really need a DeprecationWarning.....
2. Not really the point of this PR, but do we want a todo to add in a switch for MKL (instead of our current comment about renaming .h files)?

ben-albrecht:
@npadmana - thanks for the feedback.

>2.  Not really the point of this PR, but do we want a todo to add in a switch for MKL (instead of our current comment about renaming .h files)?

This TODO was meant to cover this:

`  - Modular implementations using config param-wrapped require statements`

However, after your comment, I wondered "why not just add the config param in this PR?", and so I did, and removed the header-renaming business from the documentation. Also, clarified the TODO I referenced.

modules/packages/BLAS.chpl
ben-albrecht: OK thanks.

PR #5672, opened 2017-03-15
Future for ArrayView of ArrayView bug

benharsh:
Adds a future for use-after-free bug when returning an ArrayView of an ArrayView.


benharsh:
@bradcray : here's a future for the problem we discussed yesterday. It is tracked in #5673

PR #5674, opened 2017-03-15
Fix chpl_launchcmd.py under slurm when --nostdinredirect is used

ronawho:
Under slurm, chpl_launchcmd tries to take stdin and write it to a file for
slurm to use. However, if stdin is empty (like when --nostdinredirect is
thrown) then chpl_launchcmd just hangs waiting for an EOF. Fix that by checking
if stdin has any input.

### DIFF
--- a/util/test/chpl_launchcmd.py
+++ b/util/test/chpl_launchcmd.py
@@ -34,6 +34,7 @@ import logging
 import os
 import os.path
 import re
+import select
 import shlex
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


ronawho:
@benharsh mind reviewing? @gbtitus ran into these even not under start_test (just doing `$CHPL_HOME/util/test/chpl_launchcmd.py ./a.out -nl1`) which is why I didn't just check if `CHPL_NO_STDIN_REDIRECT` is set.

Check is taken from http://stackoverflow.com/questions/3762881/how-do-i-check-if-stdin-has-some-data

benharsh:
Awesome, I think I've ran into this in the past myself.

Looks good 👍

PR #5681, opened 2017-03-15
Allow fields and vars to be 'void', make range.stride void if not stridable

daviditen:
Remove the code that issues the error "illegal use of function that does not
return a value".  That is legal now as long as the 'void' result is only
assigned to a 'void' variable.

Allow '_void' to be returned from functions as a param.

At the end of function resolution, clean up by removing 'void' variables and
fields.  Remove PRIM_DEREF, PRIM_GET_MEMBER, PRIM_SET_MEMBER, etc. involving
'void'. Remove 'void' actuals and formals.  Change
'PRIM_MOVE(voidVar, aFunction())' into just a call to 'aFunction()'.

Add assignment operators that issue a compilerError for 'void = val' and
'val = void' assignments. Add a 'void = void' assignment that does nothing.

Add a blank 'void.writeThis(f)' method and make 'void' not be an 'isIoField' in
order to make it not print anything in writeln() and similar functions.

Add a compilerError in 'buildArray' for arrays of 'void'.

Add 'void' to the list of types resolveIntents can handle.

Allow 'autoCopy' and 'initCopy' for 'void' to pass through
'removeUnnecessaryAutoCopyCalls' even though they don't have an argument
anymore.

Updated some .good files for tests that were expecting the "illegal use of
function that does not return a value" error.  Some of these tests now compile
and run, others get different, but reasonable compile time errors. Made one
into a .future because it is trying to cast something to the void value (not type).

Removed .future files for tests of 'void' variables and fields.  Update two of
the 'void' test .good files to reflect current expected behavior.

As a bigger test of void fields, changed range._stride a void field if the
range is not stridable.  This required some cleanup like using a 'stride'
accessor function in several places instead of accessing the '_stride' field
directly.


daviditen:
Passed single-locale paratest on linux64
Passed --no-local paratest on linux64

test/functions/bradc/forgotReturn.good
```diff
@@ -1 +1 @@
-forgotReturn.chpl:5: error: illegal use of function that does not return a value: 'foo'
+forgotReturn.chpl:9: error: unresolved call '<<(2, void)'
```
bradcray: I heard that @ronawho already brought this up on chat, but find myself wondering whether we should/could treat both "returns void expression" and "contains no returns" functions as having "void" return type (from a user perspective), yet having the compiler flag the latter case in order to generate error messages more like the old one here (my motivation being that I think it might catch an error of misunderstanding or mis-typing earlier in the code).

Without having put too much thought into it, it seems to me like an advantage would be that error messages might appear earlier and be clearer to understand (e.g., on line 5 rather than line 9); and that a disadvantage would be that we'd essentially have two flavors of void functions distinguished only by the presence/absence of return statements (which seems easy for the compiler to keep straight, via a flag, say -- but may be weird to purist users, esp. if they think they want/need to be able to declare both flavors of void explicitly, which I wouldn't want to support...).

test/functions/iterators/diten/iteratorNoYield.good
```diff
@@ -1 +1 @@
-iteratorNoYield.chpl:9: error: illegal use of function that does not return a value: 'itr'
+iteratorNoYield.chpl:9: error: cannot iterate over values of type void
```
bradcray: This is an interesting one.  Should an iterator with no yields be treated as yielding voids, or should it result in the loop collapsing into a no-op?  I might be inclined to do the latter, as it seems like a potentially powerful feature (e.g., if a param conditional folded all the actual yields out of the body...?)

test/functions/nspark/fcf-rettype-void.good
```diff
@@ -1 +1 @@
-fcf-rettype-void.chpl:3: error: illegal use of function that does not return a value
+void
```
bradcray: This seems like a nice improvement.

test/trivial/shannon/condReturn.good
```diff
@@ -1 +1,4 @@
-condReturn.chpl:19: error: illegal use of function that does not return a value: 'foo'
+In foo()
+{m = 2.3}
+In foo()
+condReturn.chpl:14: error: halt reached - not enough args
```
bradcray: This is a nice improvement!

test/users/jglewis/wrong_error_message_april_26_2010.future
```diff
@@ -0,0 +1,9 @@
+bug: cast to non-type is not flagged as an error
+
+The cast on lines 18-20 is casting to the void value, not type.  It currently
+compiles and runs, but should be an error.
+
+This is the same basic problem as shown by the tests:
+  expressions/vass/cast-to-non-type-1.future
+  expressions/vass/cast-to-non-type-2.future
+
```
bradcray: This makes sense to me.  I'm always surprised that our compiler still lets this case slip through... it seems like it should be a pretty easy hole to fix.  Would you file an issue against it assuming we don't have one already? (and add it to the 1.15.0 milestone in hopes that we'll finally plug the hole?  :) ).

test/types/void/void-variable-use-3.good
```diff
@@ -1 +1 @@
-void-variable-use-3.chpl:5: error: cannot use the value of a void variable
+
```
bradcray: I was wondering what happened if a void is written out.  Looks like nothing?  Does this seem more correct than writing 'void' or something like that?  (I'm honestly not sure... can see downsides to both approaches, though potential power in what you've done here...).

bradcray:
I made some comments on the tests, but I wouldn't hold up the PR over any of them -- instead, I'd file design issues against them and then try to have some discussion on them as time permits (online or offline).  Of the ones I made, I think the "what should a yield-less iterator do?" is the one that seems most compelling to me.

I didn't review the code changes themselves.

I was surprised that you didn't add any new tests because it seemed likely that you would've written some good small ones as you were working on this to try and work through issues in a controlled setting...?

Thanks for taking this on -- I'm really happy to have this feature coming online.

modules/internal/ChapelRange.chpl
```diff
@@ -144,7 +144,8 @@ module ChapelRange {
   
     var _low: idxType = 1;                         // lower bound
     var _high: idxType = 0;                        // upper bound
-    var _stride: strType = 1;                      // signed stride of range
+    var _stride: if stridable then strType else void =
+                   if stridable then 1 else _void; // signed stride of range
```
bradcray: This might read more cleanly as:

````chapel
var _stride = if stridable then 1:strType else _void;
````

This is also the first time I noticed that there's a '_void' value.  I'd like to come up with another notation for this that doesn't use an underscore...  (particularly if we expect users to use it).

File: modules/internal/ChapelRange.chpl [hunk 8]
```diff
@@ -144,7 +144,8 @@ module ChapelRange {
...
     var _alignment: idxType = 0;                   // alignment
     var _aligned : bool = false;
```
bradcray: Ultimately, I'd really like alignment and aligned to also be void in the event that stridable is false.  But I think that can/should wait for another PR (arguably perhaps even the stride change should be, but I'm not going to object to it being here as well...).

modules/internal/ChapelBase.chpl
```diff
@@ -103,7 +103,25 @@ module ChapelBase {
...
+  inline proc =(ref a: void, b: ?t) where t != void {
+    compilerError("a void variable cannot be assigned");
+  }
+
+  inline proc =(ref a: ?t, b: void) where t != void {
+    compilerError("cannot assign void to a variable of non-void type");
+  }
 
+  inline proc =(ref a: void, b: void) param {
+    return _void;
```
bradcray: Most =() overloads don't return anything -- does this one need to return something for some reason?  I was expecting an empty body here.

modules/internal/ChapelArray.chpl
```diff
@@ -1156,6 +1156,9 @@ module ChapelArray {
     pragma "no doc"
     pragma "no copy return"
     proc buildArray(type eltType) {
+      if eltType == void {
+        compilerError("array element type cannot be void");
+      }
```
bradcray: Perhaps another semantic issue: Should arrays of void simply become void variables themselves?  (or arrays of void variables?)

modules/internal/ChapelRange.chpl
ronawho: I was able to just use `void` in https://github.com/chapel-lang/chapel/commit/0b6db1aaa7c11be4b0df9cb55486b4f79da77a6a#diff-86f9faa2b8181d9c22f3b157cd767397R51

modules/internal/ChapelBase.chpl
```diff
@@ -103,7 +103,25 @@ module ChapelBase {
...
+  }
+
+/*
+  // TMacD suggested possibly wanting to be able to cast a non-void to void
+  // This should do that.
+  inline proc _cast(type t, a) where t == void {
+    return _void;
+  }
+*/
```
bradcray: Hmmm... I don't think I'd disable this by default, at least without a compelling example.  If we did want to support it, I'd put it in a little module of its own rather than ChapelBase (unless the compelling example convinced me).

modules/internal/ChapelRange.chpl
ronawho: Oh wait, nevermind, my thing was for types, not values

ronawho: If the void cast was supported could you just do `0:void` or something?

modules/internal/ChapelRange.chpl
```diff
@@ -861,7 +862,12 @@ proc _cast(type t, r: range(?)) where isRangeType(t) {
 
     r1._low = r2._low;
     r1._high = r2._high;
-    r1._stride = r2._stride;
+
+    if s1 && s2 then
+      r1._stride = r2.stride;
+    else if s1 then
+      r1._stride = 1;
```
bradcray: Nice!  Integrating the !s1 && s2 compilerError above with this control flow might help readers like me who found themselves asking "What about the missing s2 && !s1 case...?" before noticing it's handled separately and earlier.

bradcray:
I noticed belatedly that the range changes were on this PR as well, so looked quickly at them as well as the other module code changes.  So maybe now the thing that still needs review is the compiler changes (though I wouldn't mind another set of eyes on the things I looked at because I was admittedly reading fast).

modules/internal/ChapelRange.chpl
bradcray: I think that would work, but I'm not sure if that's better than having a named void value or a built-in function that generates void values...  I worry that "0:void" won't be very meaningful to most readers (or we could do things like "100432:void" just to be even more confusing.  :)

vasslitvinov:
Code changes are good to go.
I suggest paratest for linux64 --verify, if you haven't already.

test/functions/iterators/diten/iteratorNoYield.good
bradcray: I was thinking about this case a bit more this morning, and it increasingly seems to me that this might be the compelling case for distinguishing between "actually yields voids" vs. "contains no yields."  Consider:

````chapel
iter yieldTenVoids() {
  for i in 1..10 do
    yield _void;
}

iter yieldAVoid() {
  yield _void;
}

iter yieldNothing() {
  writeln("In my iterator!")
}
````

I strongly suspect that it would be surprising if the following loops did not behave as indicated in the comments:

````chapel
for v in yieldTenVoids() do
  writeln("Got a void!");    // should print this message ten times

for v in yieldAVoid() do
  writeln("Got a void!");    // should print this message one time

for v in yieldNothing() do
  writeln("Got a void!");    // should not print the loop body's message at all, but should still print the iterator's message
````

Does it seem correct that that would be the right behavior?  Like the "return nothing" error message change comments above, it seems plausible that we could implement this by having the compiler attach a flag to the iterator to distinguish between "void because it returns/yields a void" vs. "void because it returns/yields nothing."

As with my other comments, I wouldn't hold the PR up over this, but at present, this seems like the logical way to take these cases in the future.

**Comment by daviditen** [comment]
Passed --verify paratest on linux64

File: test/users/jglewis/wrong_error_message_april_26_2010.future (hunk 5 above)
Comment (daviditen) [inline test/users/jglewis/wrong_error_message_april_26_2010.future]: You've already spotted it, but I filed issue #5691

File: test/types/void/void-variable-use-3.good (hunk 6 above)
Comment (daviditen) [inline test/types/void/void-variable-use-3.good]: Right, it prints nothing as implemented.  I took the approach that a void variable or field is something you wanted to be removed from the code, and something that is not there shouldn't cause anything to print.  I agree that it could also be nice to go the other way and print 'void'.  It's not obvious to me which is better.

File: modules/internal/ChapelBase.chpl (hunk 9 above)
Comment (daviditen) [inline modules/internal/ChapelBase.chpl]: I think you're right - it shouldn't need to return.

File: modules/internal/ChapelRange.chpl (hunk 8 above)
Comment (daviditen) [inline modules/internal/ChapelRange.chpl]: Currently an unstridable range can have alignment - though I'm not sure it's possible to make it do anything.  I guessed that it would affect the new range that gets created `by` the by operator, but it doesn't appear to.

```
var r = 1..10 align 2;
writeln(r.aligned); // true
writeln(r.alignment); // 2
writeln(r.stridable); // false
writeln((r by 2).alignment); // 1
```

Making them `void` and disallowing this seems fine to me.

bradcray: This is a case I was discussing with Nikhil on email recently (and we talked about generating a warning for applying the align operator to an unstrided range indicating that it was a no-op, but I never implemented that).  What's happening is that the align operator normalizes the alignment to a value in 0..stride-1.  So if stride == 1, all indices are equivalent mod 1 and therefore we always store the alignment as 0 (so it's a no-op and we could warn against it... and ultimately should).

modules/internal/ChapelBase.chpl
daviditen: I removed this (already commented out) _cast function.  If we enabled it, it would cause any value to be able to be cast and then assigned to a void variable.  Not very useful, but I wanted to at least consider the idea.

daviditen:
Thank you @bradcray, @vasslitvinov, @ronawho for looking at and reviewing this.  I think the main TODOs I have left are to write some issues:

- [ ] Semantic issue about functions not returning vs. returning a void value
- [ ] Semantic issue about Iterator not yielding vs. yielding void value
- [ ] Semantic issue about arrays of void
- [ ] Issue asking for _void to be renamed

bradcray:
Sounds good, thanks David!  I think you could potentially combine the first two semantic issues into a single item (though I'm seeing you already filed one of them).  I think you can also merge prior to filing the semantic issues if you like.

PR #5683, opened 2017-03-15
Draft of delete-free records

mppf:
Adds records Owned and Shared.

Shared is relatively easy in terms of language integration.
Owned is harder, because the natural implementation makes copy-initialization and assignment transfer ownership. That means that the RHS of those operations is modified (to contain 'nil' instead of whatever else was stored there).
This PR makes an Owned that allows copy-initialization and assignment (in a way that modifies the RHS of those operations - it "empties" the RHS). However, the Owned in this PR can only transfer ownership on variable initialization, assignment, or `in` intent. Other `autoCopy` calls added implicitly by the compiler will cause a compilation error.

This PR includes some compiler changes to enable Owned:
 * adds flags FLAG_COPY_MUTATES and FLAG_ERRONEOUS_AUTOCOPY. These currently mark Owned type and auto copy function. In the future, FLAG_COPY_MUTATES could be applied to any record with a copy-initializer accepting its argument by `ref`.
 * this PR adds two workarounds to copyPropagation.cpp to disable propagation for Owned. I'm sure there is a better way to solve the problem in copy propagation.
 * adjusts resolveArgIntent in resolveIntents to match what resolution does. In particular, for a type marked with FLAG_COPY_MUTATES, the `in` intent changes to a `ref` intent rather than a `const ref` intent.
 * in checkForErroneousInitCopies, check for calls to any function marked with FLAG_ERRONEOUS_AUTOCOPY.
 * in functionResolution, in resolveAutoCopyEtc use the autoCopy function for a type marked with FLAG_COPY_MUTATES. This is a workaround to allow Owned to have an autoCopy with FLAG_ERRONEOUS_AUTOCOPY - normally a record can only have initCopy.
 * in functionResolution, in lvalueCheck, check in and const in intents for types with FLAG_COPY_MUTATES. For such types, passing to `in` or `const in` intent will mutate the actual (and so the actual can't be `const`).

Passed full local testing.


mppf:
@daviditen, @vasslitvinov - please have a look at this draft of the wrapper records for delete-free. I even attempted to do some documentation.

modules/packages/OwnedObject.chpl
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+
+/*
+
+   :record:`Owned` (along with :record:`Shared`) manage the deallocation
+   of a class instance. :record:`Owned` is meant to be used when only
+   one reference to an object needs to manage that object's storage.
+
+   To use :record:`Shared`, allocate a class instance following this
```
cassella: I think you mean "To use :record:`Owned`" here.

test/classes/ferguson/delete-free/rc1.chpl
```diff
@@ -0,0 +1,20 @@
+use SharedObject;
+
+class Impl {
+  var field: int;
+  proc init(arg:int) {
+    field = arg;
+    writeln("in Impl.init");
+  }
+
+  proc ~Impl() {
```
daviditen: The tests should all switch to deinit() too.

modules/packages/SharedObject.chpl
```diff
@@ -0,0 +1,221 @@
...
+       globalSharedObject = mySharedObject;
+       // now mySharedObject is deinitialized, but the MyClass
+       // instance is not deleted until globalSharedObject is deinitialized.
+     }
+
+ */
+module SharedObject {
+
+  pragma "no doc"
+  class ReferenceCount {
```
daviditen: Do you anticipate this class being used for anything other than `Shared`?  What do you think about nesting the `ReferenceCount` class in the `Shared` record?

daviditen:
I think this is looking really nice.  I left some inline comments, but I think they're fairly minor.  Nice work!

test/classes/ferguson/delete-free/rc2.chpl
```diff
@@ -0,0 +1,25 @@
...
+  proc ~Impl() {
+    writeln("in Impl.~Impl");
+  }
+}
+
+proc run() {
+  var x = new Shared(new Impl(1));
+
+  // check initCopy
+  var y = x;
```
vasslitvinov: How about add x.p.field += 1
to ensure both x and y refer to the same Impl?

test/classes/ferguson/delete-free/rc4.chpl
```diff
@@ -0,0 +1,23 @@
...
+
+var globalSharedObject:Shared(Impl);
+
+proc makeGlobalSharedObject() {
+  var x = new Shared(new Impl(1));
+
+  globalSharedObject = x;
+}
+
+makeGlobalSharedObject();
```
vasslitvinov: let's add writeln(globalSharedObject.p.field)
this will for example show that deinit happens later

modules/packages/SharedObject.chpl
vasslitvinov: ReferenceCount may be useful on its own, so I can see providing it to users. If so:

* Rename to ReferenceCounter
* Optionally, place it in its own module and file, like ReferenceCounterObject, with SharedObject publicly using it.

OTOH I am OK going in the other direction. If so, I prefer not nesting it inside `Shared` - just make it `private` to the SharedObject module.

modules/packages/SharedObject.chpl
```diff
@@ -0,0 +1,221 @@
...
+     that supports multiple owners of the class instance.
+
+     This is currently implemented with task-safe reference counting.
+
+   */
+  record Shared {
+    pragma "no doc"
+    var p;                 // contained pointer (class type)
+    pragma "no doc"
+    var pn:ReferenceCount; // reference counter
```
vasslitvinov: I suggest making both `p` and `pn` be `private`.
We don't want users to access them, right?

Correspondingly, change tests to use `.borrow()` instead of `.p`.
We should be testing as if we are users, right? :)

modules/packages/OwnedObject.chpl
```diff
@@ -0,0 +1,130 @@
...
+       instance in manages when the :record:`Owned` goes out of scope.
+     */
+    proc ~Owned() {
+      if p then
+        delete p;
+    }
+
+    /*
+       Change the instance managed by this class to `newPtr`.
+       If this class was already managing a non-nil instance,
```
vasslitvinov: class --> record?

modules/packages/SharedObject.chpl
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    pragma "no doc"
+    proc retain() {
+      pn.retain();
+    }
+
+    /*
+       Change the instance managed by this class to `newPtr`.
+       If this class was the last :record:`Shared` managing a
```
vasslitvinov: class --> record here, too?

modules/packages/SharedObject.chpl
vasslitvinov: For now, tagging it with "no doc" is just as good as making it "private".
I.e. I am OK leaving it as-is.

modules/packages/OwnedObject.chpl
```diff
@@ -0,0 +1,130 @@
...
+       When this :record:`Owned` goes out of scope, it will
+       delete whatever class instance it is storing.
+
+       It is an error to directly delete the class instance
+       while it is managed by a :record:`Owned`.
+
+       :arg p: the class instance to manage. Must be of class type.
+     */
+    proc Owned(p) {
+      if !isClass(p.type) then
```
vasslitvinov: let's do simply !isClass(p) above

Next line: how about "only works" --> "works only" ?

modules/packages/SharedObject.chpl
mppf: I'd prefer to consider it "private" for now but AFAIK we can't have private classes. It might be useful in the future but I'd prefer to wait until we have a 2nd use to do that. It might turn out that it's trivial enough that it's not worth sharing.

mppf:
@daviditen - I don't anticipate any testing problems, but testing is ongoing. This is ready for your final review.

modules/packages/SharedObject.chpl
nspark: This seems like a spiritual duplicate to the [RefCountBase](https://github.com/chapel-lang/chapel/blob/master/modules/packages/ExplicitRefCount.chpl#L23) class. Does it make sense to de-duplicate?

mppf: RefCountedBase is meant to be for intrusive reference counting while this one is for a separate reference count. Nonetheless I think we could de-duplicate them... I just don't want to try to do that today (which means not doing it for 1.15).

mppf: The other difference - which might seem trivial - is that RefCountBase starts with a 0 count but ReferenceCount starts with a 1 count. I'm just trying to point out that it will take us effort to de-duplicate them.

modules/packages/OwnedObject.chpl
```diff
@@ -0,0 +1,187 @@
...
+       that takes over ownership from `src`. `src` will
+       refer to `nil` after this call.
+     */
+    proc Owned(ref src:Owned, type t=src.t) {
+      this.p = src.release();
+    }
+
+    /*
+       The deinitializer for :record:`Owned` will destroy the class
+       instance in manages when the :record:`Owned` goes out of scope.
```
daviditen: typo: instance _in_ manages

modules/packages/SharedObject.chpl
```diff
@@ -0,0 +1,235 @@
...
+       globalSharedObject = mySharedObject;
+       // now mySharedObject is deinitialized, but the MyClass
+       // instance is not deleted until globalSharedObject is deinitialized.
+     }
+
+ */
+module SharedObject {
+
+  pragma "no doc"
+  class ReferenceCount {
```
daviditen: Maybe add a comment here saying that it would be nice to unify this with the other ReferenceCount type that @nspark mentioned.

daviditen:
I noted some typos and suggested one more comment, but looks good to me.

PR #5684, opened 2017-03-15
Update normalization of new-expressions for non-generic records with initializers

noakesmichael:
Before this PR statements such as

   return new MyRec(10, 20);

or

  foo(new MyRec(10, 20), new  MyRec(30, 40));

where MyRec is a non-generic record with an initializer failed to resolve.

The problem was that the compiler was still trying to "construct" the call_tmp
instead of "initializing" it.  A relatively simple change to insertCallTemps()
addressed this oversight; as least for the common cases of new expressions.

start_test classes/initializers -futures on darwin
single-locale paratest on linux64


lydia-duncan:
This seems reasonable!

PR #5685, opened 2017-03-15
Add annotation for "ListDest" array-vector perf regression

ronawho:
Turns out the regression was caused by the jemalloc 4.4.0 upgrade (which also
hurt binary-trees a little.) #5278 has more details on the regression, and it's
not one I think we should worry about. Also note that overall this test has
benefited significantly from jemalloc, so I'm not worried about a minor
regression anyways.)


ronawho:
@bradcray just fyi since this is one of the regressions we saw in rel-over-rel that you wanted investigated.

PR #5689, opened 2017-03-16
NUMA-localize comm=ugni desired shared heap.

gbtitus:
With comm=ugni as it's normally used, we have a fixed-size heap that is
registered with the NIC.  If the heap were not registered with the NIC
then we could (and indeed, do) re-localize allocated memory however we'd
like.  But in order for memory gotten from the registered uGNI/Aries
heap to have a specific locality it must already have that locality when
it is allocated.  The locality cannot be changed post-allocation because
NIC registration fixes the memory page virtual-to-physical relationship
and thus the NUMA locality.  The new multi-ddata feature for arrays gets
us partway to a solution, because it allows the memory for an array to
consist of separately allocated subchunks, one for each NUMA domain
sublocale.  That's necessary but not sufficient.  This is the rest of
the solution: here we split the heap itself into per-sublocale
subchunks, bind those chunks to their NUMA domains, and arrange for
tasks to allocate from the NUMA domain they are running in at the time
they call the allocator.  The calls to allocate multi-ddata chunks are
already being made from the sublocales on which we want the memory to
reside -- that code was included in the original multi-ddata work.  With
this change, the allocated memory will actually have that locality.

Here, in the jemalloc memory layer, we block-localize the subchunks of
the comm layer desired shared heap across the NUMA sublocales and then
manage each subchunk as a separate sub-heap.  Jemalloc maintains a fixed
number of arenas (a small multiple of the number of processor cores).
Each pthread has a particular arena bound to it.  With a tasking layer
such as qthreads which has one worker pthread per core or hyperthread
(depending on circumstances) and thus fewer pthreads than arenas, each
pthread will essentially end up with a private arena.  When we extend an
arena we get the memory for that from the sub-heap the thread is running
on, so a thread's arena will consist entirely of memory localized to the
NUMA domain Qthreads has bound the thread to, and tasks running on that
worker will always allocate memory bound to that domain.

For free() we don't have to do anything special, since jemalloc
internally arranges to free memory back to the same arena from which it
was allocated.

Most of the work here is in mem-jemalloc and most of that is to support          
multiple heaps (at most 4, currently) instead of just 1.  Once we know           
we have a desired shared heap from the comm layer, if we have multiple           
NUMA domains (it's too early to ask the tasking layer about this so we           
ask the topology layer), we call the topology layer to block-localize            
that heap, and then manage the localized blocks as our arenas.  (For             
reasons that are not entirely clear at this point we still have to touch         
the pages of these blocks while running on the appropriate NUMA domains          
to get the actual memory localized, so that is also done at this time.           
We'll look further into why this is needed later.)                               
                                                                                 
In the topology layer, there are new functions to get and set execution
locality.  And, what used to be a single function to localize either the
whole of an area of memory of its subchunks has been turned into two
functions that do specifically those tasks, with common code factored
out into helpers.  There are also some changes for the sake of clarity,
such as moving flag settings closer to where they are used.

Finally, at the top surface of the memory layer interface in chpl-mem.h,
we now localize memory after allocation only if the allocator does not
return already-localized memory and the allocation is large enough that
it seems like localizing would help performance.  This threshold is
currently 2 pages times the number of sublocales.  For now, jemalloc is
the only memory layer that can allocate localized memory directly, and
then only when the comm layer provides the desired shared heap.

### DIFF
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -234,7 +234,7 @@ void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
 // where chpl_mem_doLocalization is true, above.)
 //
 static inline
-chpl_bool chpl_mem_allocLocalizes(void) {
+bool chpl_mem_allocLocalizes(void) {
 #ifdef CHPL_MEM_IMPL_ALLOCLOCALIZES
     return CHPL_MEM_IMPL_ALLOCLOCALIZES;
 #else
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -29,6 +29,7 @@
 #include "arg.h"
 #include "chpl-comm.h"
 #include "chpl-mem-hook.h"
+#include "chplsys.h"
 #include "chpl-topo.h"
 #include "chpltypes.h"
 #include "chpl-tasks.h"
@@ -59,16 +60,15 @@ int chpl_posix_memalign_check_valid(size_t alignment);
 // and no additional error checking.
 #include "chpl-mem-impl.h"
 
-size_t chpl_mem_localizationThreshold;
-
 void chpl_mem_init(void);
 void chpl_mem_exit(void);
 
 int chpl_mem_inited(void);
 
-// predeclared here because we need it below; actual definition is
-// near the end
-static chpl_bool chpl_mem_allocLocalizes(void);
+// predeclared here because we need them below; actual definitions
+// are near the end
+static chpl_bool chpl_mem_alloc_localizes(void);
+static size_t chpl_mem_localizationThreshold(void);
 
 static inline
 void* chpl_mem_allocMany(size_t number, size_t size,
@@ -139,14 +139,14 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize,
   void* p = chpl_mem_allocMany(nmemb, eltSize, CHPL_RT_MD_ARRAY_ELEMENTS,
                                lineno, filename);
   if (isActualSublocID(subloc)) {
-    if (!chpl_mem_allocLocalizes()
-        && nmemb * eltSize >= chpl_mem_localizationThreshold) {
+    if (!chpl_mem_alloc_localizes()
+        && nmemb * eltSize >= chpl_mem_localizationThreshold()) {
       chpl_topo_setMemLocality(p, nmemb * eltSize, true, subloc);
     }
   }
   else if (localizeSubchunks) {
-    if (!chpl_mem_allocLocalizes()
-        && nmemb * eltSize >= chpl_mem_localizationThreshold) {
+    if (!chpl_mem_alloc_localizes()
+        && nmemb * eltSize >= chpl_mem_localizationThreshold()) {
       chpl_topo_setMemSubchunkLocality(p, nmemb * eltSize, true, NULL);
     }
   }
@@ -233,13 +233,18 @@ void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
 // for allocations large enough that we'll try to force localization
 // where chpl_mem_doLocalization is true, above.)
 //
-static inline
-bool chpl_mem_allocLocalizes(void) {
-#ifdef CHPL_MEM_IMPL_ALLOCLOCALIZES
-    return CHPL_MEM_IMPL_ALLOCLOCALIZES;
-#else
-    return false;
+#ifndef CHPL_MEM_IMPL_ALLOC_LOCALIZES
+  #define CHPL_MEM_IMPL_ALLOC_LOCALIZES() false
 #endif
+
+static inline
+bool chpl_mem_alloc_localizes(void) {
+  return CHPL_MEM_IMPL_ALLOC_LOCALIZES();
+}
+
+static inline
+size_t chpl_mem_localizationThreshold(void) {
+  return chpl_topo_getNumNumaDomains() * 2 * chpl_getHeapPageSize();
 }
 
 #else // LAUNCHER
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -59,11 +59,16 @@ int chpl_posix_memalign_check_valid(size_t alignment);
 // and no additional error checking.
 #include "chpl-mem-impl.h"
 
+size_t chpl_mem_localizationThreshold;
+
 void chpl_mem_init(void);
 void chpl_mem_exit(void);
 
 int chpl_mem_inited(void);
 
+// predeclared here because we need it below; actual definition is
+// near the end
+static chpl_bool chpl_mem_allocLocalizes(void);
 
 static inline
 void* chpl_mem_allocMany(size_t number, size_t size,
@@ -133,9 +138,17 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize,
                            int32_t lineno, int32_t filename) {
   void* p = chpl_mem_allocMany(nmemb, eltSize, CHPL_RT_MD_ARRAY_ELEMENTS,
                                lineno, filename);
-  if (localizeSubchunks || isActualSublocID(subloc)) {
-    chpl_topo_setMemLocality(p, nmemb * eltSize,
-                             true, localizeSubchunks, subloc);
+  if (isActualSublocID(subloc)) {
+    if (!chpl_mem_allocLocalizes()
+        && nmemb * eltSize >= chpl_mem_localizationThreshold) {
+      chpl_topo_setMemLocality(p, nmemb * eltSize, true, subloc);
+    }
+  }
+  else if (localizeSubchunks) {
+    if (!chpl_mem_allocLocalizes()
+        && nmemb * eltSize >= chpl_mem_localizationThreshold) {
+      chpl_topo_setMemSubchunkLocality(p, nmemb * eltSize, true, NULL);
+    }
   }
   return p;
 }
@@ -210,6 +223,24 @@ void chpl_mem_layerExit(void);
 void* chpl_mem_layerAlloc(size_t, int32_t lineno, int32_t filename);
 void* chpl_mem_layerRealloc(void*, size_t, int32_t lineno, int32_t filename);
 void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
+
+//
+// Does the implementation provide allocated memory that is already
+// localized to the calling sublocale?  That is, when the locale model
+// has sublocales and an allocation is done while running on one, will
+// the allocated memory be localized to that sublocale?  (Note that the
+// answer doesn't have to be completely truthful; it really only matters
+// for allocations large enough that we'll try to force localization
+// where chpl_mem_doLocalization is true, above.)
+//
+static inline
+chpl_bool chpl_mem_allocLocalizes(void) {
+#ifdef CHPL_MEM_IMPL_ALLOCLOCALIZES
+    return CHPL_MEM_IMPL_ALLOCLOCALIZES;
+#else
+    return false;
+#endif
+}
 
 #else // LAUNCHER
 
--- a/runtime/include/mem/jemalloc/chpl-mem-impl.h
+++ b/runtime/include/mem/jemalloc/chpl-mem-impl.h
@@ -21,8 +21,6 @@
 #ifndef _chpl_mem_impl_H_
 #define _chpl_mem_impl_H_
 
-#include "chpl-comm.h"
-
 // jemalloc.h references the token "malloc" (but not the actual function) and
 // our warning macros mess up jemalloc's use of it.
 #include "chpl-mem-no-warning-macros.h"
@@ -32,7 +30,7 @@
 #define MALLOCX_NO_FLAGS 0
 
 
-extern chpl_bool chpl_mem_impl_allocLocalizes;
+extern bool chpl_mem_impl_allocLocalizes;
 #define CHPL_MEM_IMPL_ALLOCLOCALIZES chpl_mem_impl_allocLocalizes
 
 
--- a/runtime/include/mem/jemalloc/chpl-mem-impl.h
+++ b/runtime/include/mem/jemalloc/chpl-mem-impl.h
@@ -21,6 +21,8 @@
 #ifndef _chpl_mem_impl_H_
 #define _chpl_mem_impl_H_
 
+#include "chpl-comm.h"
+
 // jemalloc.h references the token "malloc" (but not the actual function) and
 // our warning macros mess up jemalloc's use of it.
 #include "chpl-mem-no-warning-macros.h"
@@ -28,6 +30,11 @@
 #include "chpl-mem-warning-macros.h"
 
 #define MALLOCX_NO_FLAGS 0
+
+
+extern chpl_bool chpl_mem_impl_allocLocalizes;
+#define CHPL_MEM_IMPL_ALLOCLOCALIZES chpl_mem_impl_allocLocalizes
+
 
 static inline void* chpl_calloc(size_t n, size_t size) {
   return chpl_je_calloc(n,size);
--- a/runtime/include/mem/jemalloc/chpl-mem-impl.h
+++ b/runtime/include/mem/jemalloc/chpl-mem-impl.h
@@ -30,10 +30,6 @@
 #define MALLOCX_NO_FLAGS 0
 
 
-extern bool chpl_mem_impl_allocLocalizes;
-#define CHPL_MEM_IMPL_ALLOCLOCALIZES chpl_mem_impl_allocLocalizes
-
-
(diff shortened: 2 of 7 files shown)


gbtitus:
@ronawho, can you review?  I'm especially interested to see if I've made any mistakes in interpreting the jemalloc behavior!

ronawho:
Yup, will start reviewing when I get into the office. Should be in ~45 min.

ronawho:
In the PR message -- 

```
But in order for memory gotten from the registered uGNI/Aries
heap to have a specific locality it must already have that locality when
it is allocated. The locality cannot be changed post-allocation
```

Shouldn't this be talking about locality when it's registered, not when it's allocated?

runtime/include/chpl-mem.h
```diff
@@ -59,11 +59,16 @@ int chpl_posix_memalign_check_valid(size_t alignment);
 // and no additional error checking.
 #include "chpl-mem-impl.h"
 
+size_t chpl_mem_localizationThreshold;
```
ronawho: Should this be an extern?

runtime/include/chpl-mem.h
```diff
@@ -211,6 +224,24 @@ void* chpl_mem_layerAlloc(size_t, int32_t lineno, int32_t filename);
...
+// has sublocales and an allocation is done while running on one, will
+// the allocated memory be localized to that sublocale?  (Note that the
+// answer doesn't have to be completely truthful; it really only matters
+// for allocations large enough that we'll try to force localization
+// where chpl_mem_doLocalization is true, above.)
+//
+static inline
+bool chpl_mem_allocLocalizes(void) {
+#ifdef CHPL_MEM_IMPL_ALLOCLOCALIZES
+    return CHPL_MEM_IMPL_ALLOCLOCALIZES;
```
ronawho: I have a hard time parsing `ALLOCLOCALIZES`. Mind going with `ALLOC_LOCALIZES`

runtime/include/chpl-mem.h
ronawho: And actually, instead of using a global, can you just make this a function?

runtime/include/mem/jemalloc/chpl-mem-impl.h
```diff
@@ -29,6 +29,11 @@
 
 #define MALLOCX_NO_FLAGS 0
 
+
+extern bool chpl_mem_impl_allocLocalizes;
+#define CHPL_MEM_IMPL_ALLOCLOCALIZES chpl_mem_impl_allocLocalizes
```
ronawho: This is really confusing to me. Can we just have a chpl_mem_impl_allocLocalizes() routines that each mem layer must implement. And for jemalloc have it be `strcmp(CHPL_COMM, "ugni") != 0 || (num_heaps = chpl_topo_getNumNumaDomains()) <= 1` and then reuse that code in chpl_mem_layerInit?

runtime/src/chpl-topo.c
```diff
@@ -30,6 +30,7 @@
 #include "error.h"
 
 #include <errno.h>
+#include <pthread.h>
```
ronawho: Not jumping out to me, what requires this?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -57,39 +71,51 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
 // mmap/sbrk.) Grab memory out of the shared heap and give it to jemalloc.
 static void* chunk_alloc(void *chunk, size_t size, size_t alignment, bool *zero, bool *commit, unsigned arena_ind) {
 
+  int hpi;
   void* cur_chunk_base = NULL;
   size_t cur_heap_size;
 
+  // which heap?
+  if (num_heaps == 1) {
```
ronawho: Can you stuff this into a helper function?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -28,15 +28,29 @@
...
+typedef struct shared_heap {
   void* base;
   size_t size;
   size_t cur_offset;
   pthread_mutex_t alloc_lock;
-} heap; // static, will be "zero" initialized automatically
+} shared_heap_t;
+
+static shared_heap_t heaps[MAX_HEAPS];
+static int num_heaps;
```
ronawho: I have a small preference for using a function to get at num_heaps, instead of a global static

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -310,12 +343,69 @@ void chpl_mem_layerInit(void) {
   //   jemalloc 4.5.0 man: "Once, when the first call is made to one of the
   //   memory allocation routines, the allocator initializes its internals"
   if (heap_base != NULL) {
-    heap.base = heap_base;
-    heap.size = heap_size;
-    heap.cur_offset = 0;
-    if (pthread_mutex_init(&heap.alloc_lock, NULL) != 0) {
-      chpl_internal_error("cannot init chunk_alloc lock");
+    int hpi;
```
ronawho: I think this logic should be stuffed into a helper routine that's called from initializeSharedHeap()

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -310,12 +343,69 @@ void chpl_mem_layerInit(void) {
...
+
+    //
+    // For now, we only NUMA-localize the comm layer desired shared heap
+    // if we're using the ugni comm layer.  ugni is the simpler case
+    // because it's only used on Cray X* compute blades which are a
+    // real-memory architecture.  We'll tackle comm=gasnet later; that's
+    // more complicated because GASNet is used on all sorts of systems,
+    // some virtual-memory, and we can't just go touching all the pages.
+    //
+    if (strcmp(CHPL_COMM, "ugni") != 0
```
ronawho: Mentioned in another comment, but I think this conditional should be turned into a helper function that also replaces chpl_mem_impl_allocLocalizes()

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -28,15 +28,29 @@
...
 #include "chpltypes.h"
 #include "error.h"
 
-static struct shared_heap {
+
+//
+// Information about shared heaps gotten from the comm layer.
+// MAX_HEAPS must be >= the number of sublocales.
+//
+#define MAX_HEAPS 4
```
ronawho: Is there any reason to set the limit at compile time? Couldn't we just switch to initializing that shared_heap struct dynamically?

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: I would prefer to explicit exclude flat here. I think we'll end up excluding it currently because chpl_topo_getNumNumaDomains() will be zero under flat currently, but that might not always be the case

ronawho:
I left a lot of style/cleanup comments (I really dislike our liberal use of globals in a lot of the runtime code), but this looks good from a correctness standpoint to me. 

My preference would be to merge this today to get more testing on it, and do the cleanup on Monday or something.

@dmk42 might be interested in reviewing the topo changes, I did not review them too carefully.

gbtitus:
> In the PR message --
> 
> But in order for memory gotten from the registered uGNI/Aries
> heap to have a specific locality it must already have that locality when
> it is allocated. The locality cannot be changed post-allocation
> 
> Shouldn't this be talking about locality when it's registered, not when it's allocated?

What I was trying to say there was that if the caller wants locality for memory within such a registered heap, the allocator has to return memory already localized as the caller desired, because the fact that it's registered precludes the caller re-localizing it after allocating it.

runtime/include/mem/jemalloc/chpl-mem-impl.h
gbtitus: I'd prefer not to have to mention it at all in the layers that can't do it.  We have a long-standing technique for doing optional stuff in layer implementations; see the discussion of `CHPL_TASK_func_IMPL_DECL` in chpl-tasks.h.  But that's come to seem clunky to me because the #ifdef controls whether the base .h file or the one from the implementation provides both the declaration and the definition.  So here what I was trying to do was to have the base .h file always provide the declaration, with the implementation optionally providing the implementation, effectively.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: We can.  I'd need to use malloc(), of course, which would mean that we could never set up our jemalloc to intercept the C library allocator symbols, since here we'd need to do that malloc before we're set up.  Do we feel confident that we'll never want to do that?

runtime/src/chpl-topo.c
gbtitus: There's a `pthread_self()` call in `chpl_topo_getThreadLocality()`.

runtime/include/chpl-mem.h
gbtitus: Not a bit, and this may also change depending on where we end up with the optional-implementation question.

runtime/include/chpl-mem.h
gbtitus: Yes, will do (static inline function).

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: Basically just a getter for a file static var?  Will do.

dmk42:
I took a brief look at the topo changes.  I'll read them in more detail later for my own education, but I won't hold up the review for that.  What I've seen so far looks good to me.

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: Could still use jemalloc, right? It just wouldn't be part of the registered heap, which seems fine.

gbtitus: Hmm, hadn't thought about that option.  Would `useUpMemNotInHeap()` then have to do more work, since we'd initialized jemalloc?  Or is that not really a concern?

gbtitus:
I've pushed a commit that I believe covers the feedback so far.  @ronawho, could you take another look?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -328,8 +439,27 @@ void chpl_mem_layerInit(void) {
...
+  for (hpi = 0; hpi < get_num_heaps(); hpi++) {
+    if (heaps[hpi].base != NULL) {
+      // ignore errors, we're exiting anyways
+      (void) pthread_mutex_destroy(&heaps[hpi].alloc_lock);
+    }
   }
 }
+
+
+bool chpl_mem_impl_alloc_localizes(void) {
```
ronawho: bool or chpl_bool?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -310,12 +426,7 @@ void chpl_mem_layerInit(void) {
   //   jemalloc 4.5.0 man: "Once, when the first call is made to one of the
   //   memory allocation routines, the allocator initializes its internals"
   if (heap_base != NULL) {
-    heap.base = heap_base;
-    heap.size = heap_size;
-    heap.cur_offset = 0;
-    if (pthread_mutex_init(&heap.alloc_lock, NULL) != 0) {
-      chpl_internal_error("cannot init chunk_alloc lock");
-    }
+    layerInitHelper(heap_base, heap_size);
```
ronawho: Think something like setupLocalization() or initializeLocalization() and move the call into initializeSharedHeap()

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -294,6 +330,86 @@ static void initializeSharedHeap(void) {
...
+    subchunk_base += heaps[hpi].size;
+
+    //
+    // The chpl_topo_setMemSubchunkLocality() call above isn't
+    // having the effect we desire, for reasons that are mostly but
+    // not completely understood.  So for now, to get the locality
+    // we want, touch the pages into existence while executing on
+    // the desired NUMA domains here.
+    //
+    fflush(stdout);
```
ronawho: Is this leftover?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -28,15 +28,40 @@
...
   size_t cur_offset;
   pthread_mutex_t alloc_lock;
-} heap; // static, will be "zero" initialized automatically
+} shared_heap_t;
+
+static shared_heap_t* heaps;
+static int num_heaps;
+
+static inline int get_num_heaps(void) {
+  return num_heaps;
```
ronawho: Can you make num_heaps a local static, and just initialize it in this routine?

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: I don't see any particular reason to make it chpl_bool since it isn't shared across the interface with Chapel (and I think in too many places in the runtime we use chpl_bool where we don't need it).  That said, I'm not averse to changing it if you'd prefer that.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: d'oh!

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: Oh, I didn't think we could rely on bool existing, so we used chpl_bool

ronawho: Never mind, we use bool all over the place -- thinko on my part

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: I could.  Note that that'll necessitate an "is this set?" check on every access and also either an implicit assertion that the first access precedes all parallelism or a parallel-unsafe initialization at least for non-POSIX systems where we don't have pthread_once().  (I'm sensitive to this issue because I just went round a bit with another reviewer a few days ago about a variable initialization I'd done under pthread_once() as a future-proofing step, even though I knew that in current code the initialization would definitely occur before any concurrency.)

ronawho: Ok. We could also just have get_num_heaps return chpl_topo_getNumNumaDomains(). get_num_heaps and get_heaps are't going to be used in performance critical areas. It's called when we're doing a chunk_allocation, but that doesn't get called very often.

ronawho: And my motivation here is that I have such a hard time following when a global has been initialized, and when it's safe to use. I much prefer avoiding that unless there's a big performance reason to do so.

Also even if get_num_heaps was called concurrently, I don't think that'd be a problem. You might end up initializing concurrently, but I don't think that'd hurt anything.

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -57,39 +168,44 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
 // mmap/sbrk.) Grab memory out of the shared heap and give it to jemalloc.
 static void* chunk_alloc(void *chunk, size_t size, size_t alignment, bool *zero, bool *commit, unsigned arena_ind) {
 
+  int hpi;
   void* cur_chunk_base = NULL;
   size_t cur_heap_size;
 
+  // which heap?
```
ronawho: I'm ok leaving it in, but I don't think this comment adds much

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -28,15 +28,126 @@
...
+      } else if (!chpl_mem_impl_alloc_localizes()) {
+        num_heaps = 1;
+      }
+    }
+  }
+
+  return num_heaps;
+}
+
+static inline int get_heap(void) {
```
ronawho: Maybe rename to get_nearest_heap.. that's a bad name, but something that indicates it's getting the heap that you have affinity to?

get_heap_with_thread_locality?

get_nearest_heap

get_heap_sorry_elliot_sucks_at_naming_things

Eh.. maybe just leave the name alone (unless you have a really good one) and add a short comment noting it'll get the heap with thread locality

ronawho: oh, maybe get_localized_heap (which kind of matches setupLocalizedHeap)

ronawho:
👍 Loving the cleanup. Left a minor note about get_heap's name, but I'm happy with this!

PR #5694, opened 2017-03-16
Rank change dom dist view

bradcray:
Overall, this PR accomplishes two things, one of which is core, one of
which is cleanup:

1) It implements rank-change views on domains and distributions in
   support of having locality be preserved for domains of rank-change
   array slices.  This can perhaps be most easily understood by
   comparing the before and after .bad entries for rank-change slices
   in test/arrays/viewsviewPreserveDist.chpl.

2) It cleans up some aspects of the dsi interface.

In more detail
==========

1) Generally speaking, the implementation of rank-change views on
   domains and distributions is done by creating distribution and
   domain "view classes" that are similar to the rank-change array
   view classes, in that they can translate between lower-dimensional
   logical views of the distribution/domain to the higher-dimensional
   physical views that implement the representation.  This is
   particularly important for distributed domains in which the
   rank-change domain/array must be aligned with a specific
   higher-dimensional domain/array.

   For example, if one were to query a rank-change array's
   distribution to say "which locale owns index i live?", or its
   domain to say "is index i a member?"  these indices must be
   translated into their higher-dimensional equivalents before passing
   them down to the underlying domain or distribution that implements
   them.  In essence this avoids requiring a 3D block distribution to
   know how to create a 1D or 2D version of itself just as we didn't
   want to teach every 3D array how to store 1D or 2D slices of
   itself, so created views instead.

   For the most part, these classes work by simply implementing the
   dsi interface and translating indices, tuples, and domains up and
   down between the user's low-dimensional view and the underlying
   higher-dimensional view.

2) W.r.t. dsi cleanup, it removes [domain].dsiBuildRectangularDom() on the
   argument that domains should not be creating domains, but rather
   distributions should be.  It also adds an 'indices' argument to
   GlobalDistribution.dsiBuildRectangularDom() (and an optional one to
   newRectangularDom()) in order to (i) optimize a common pattern in
   which the indices are known at domain creation time; and in doing
   so (ii) restore the performance that was being obtained by the
   traditional [domain].dsiBuildRectangularDom() call (which did take
   indices as arguments).

   It also removes the dsiCreateRankChangeDist() routine which was the
   old way of having distributions create lower-D views of themselves
   (although, as the diff will show, in practice, many of them did not).

Changes by file
===============

test/arrays/views/viewPreserveDist*
-----------------------------------
* improved formatting of output and updated improved results

modules/internal/ArrayViewRankChange.chpl
-----------------------------------------

* added ArrayViewRankChangeDist and ArrayViewRankChangeDom classes
  which implement the rank-change views of distributions (really
  "domain maps") and domains respectively

* moved several fields that used to be stored in
  ArrayViewRankChangeArr into the domain class where they more
  naturally belong.

* moved some of the helper routines which had previously been methods
  of the ArrayViewRankChangeArr class out into standalone helper
  functions in order to call them from ArrayViewRankChangeDom as well

* removed a warning about dsiTargetLocales() that seemed too
  conservative and noisy at this point in time

modules/internal/ChapelArray.chpl

* expanded the set of helper functions used to reason about views
  to include domains

* created a new version of the 'newRectangularDom' routine that takes
  initial indices to pass along to dsiNewRectangularDom().  The old
  routine now builds a set of empty ranges and passes them into the
  new routine in order to preserve the old behavior.

* replaced a number of idioms of the form:

     var d = _value.dsiBuildRectangularDom(rank, _value.idxType, stridable, r);
     // Since we've created a new domain, the distribution needs to
     // live at least as long as this new domain.
     if d.linksDistribution() then
       d.dist.add_dom(d);
     return _newDomain(d);

  with the simpler:

     return _newDomain(dist.newRectangularDom(rank, _value.idxType, stridable, \
r));

* re-implemented rank-change slices on domains to use distribution and
  domain view classes.

* re-implemented rank-change slices on arrays to first create a
  rank-change domain using the logic above and then to create the
  array.

* moved an error for all-int accesses of domains into a new this()
  overload

modules/dists/BlockCycDist.chpl
modules/dists/BlockDist.chpl
modules/dists/CyclicDist.chpl
modules/dists/DimensionalDist2D.chpl
modules/dists/PrivateDist.chpl
modules/dists/ReplicatedDist.chpl
modules/dists/SparseBlockDist.chpl
modules/dists/StencilDist.chpl
modules/dists/dims/*.chpl
test/arrays/marybeth/CMO_array.chpl
test/distributions/*
test/users/vass/crash*
-------------------------------------
* updated dsiNewRectangularDom() interface and implementation
* removed [domain].dsiBuildRectangularDom()
* removed dsiCreateRankChange() when present
* updated newRectangularDom() to support new interface (StencilDist only)

modules/internal/DefaultRectangular.chpl
----------------------------------------
* same as the above, plus:

* changed implementation of dsiLocalSubdomain*() to return aliases rather
  than new arrays

* updated chpl__serialViewIter() to not assume default rectangular
  domains and to factor out a common subexpression that could be
  nontrivial to compute in a rank-change array view world

doc/technotes/dsi.rst
---------------------
* updated documentation for dsiNewRectangularDom to accept initial indices
* removed documentation for [domain].dsiBuildRectangularDom

compiler/functionResolution.cpp
-------------------------------
* updated logic for differentiating view arrays from non- w.r.t.
  how to return arrays

test/domains/userAPI/rankChangeFollow.chpl
----------------------------------------------
* a new test spot-checking that followers work as expected

test/arrays/tmacd/promotionLosesArrayShape.*
* updated domain type to reflect domain view

test/arrays/userAPI/rankChangeOps3to2D.good
-------------------------------------------
* updated to remove warning

test/studies/hpcc/common/bradc/unit/multipleDomains2.*
----------------------------------------------------------
* updated expected output and stopped testing numa which added maintenance without much value

TODO:
- [x] One more round of valgrind testing
- [x] One more round of gasnet testing (?)
- [ ] Test robust/arithmetic/ for non-Block distributions? (or maybe just performance/multilocale/?)
- [x] Check playground performance once more
- [ ] ~~Check 16-node XC performance once more?~~
- [x] FIle test against followers to make sure 0..0 is OK.
- [x] Fix local query for default rectangular or file an issue against it and remove TODO

post-merge TODOs:
- [ ] file future against claim that inline field type specification causes leak
- [ ] File test against dsiHasLocalSubdomain to make sure it's not wrong (it looks wrong).
- [ ] Can the index/tuple/etc. routines be refactored to call one another?
- [ ] Can collectRanges be rewritten in a non-recursive manner if it uses ranges only?


bradcray:
@benharsh: Here's the branch with the work on locality-preserving rank-change domains/dists.  Some of the outlier cases that I mentioned are:

incorrect output
----------------
[Error matching program output for studies/sudoku/dinan/search/sudoku-reduce]
[Error matching program output for studies/sudoku/dinan/search/sudoku-reduce3]

stray debugging output?
----------------------
[Error matching program output for studies/hpcc/common/bradc/unit/multipleDomains2]

core dump
---------
[Error matching program output for users/npadmana/twopt/do_smu_soa]

They could be really obvious... I've been tackling the more core cases that kill a dozen tests first...

bradcray:
Oh yeah, one other thing that you'd be ideal for helping with:  I have a merge conflict with master right now which I believe is due to your optimized array view iterators from the other day (?).  Haven't looked at that enough to know whether it's an easy or hard merge yet...

benharsh:
Looking at things more closely, I'm pretty sure we're running into a reduction bug in the sudoku cases. That is, a bug in the compiler. As a simpler reproducer, this program:
```chpl
proc main() {
  var A : [1..10, 1..10] int;
  A[1, ..] = 1..10;

  const sum = + reduce A[1, ..];
  writeln(sum);
}
```

prints "0", where we would expect "55".

benharsh:
It seems we can work around this by modifying the ArrayViewRankChangeDom standalone iter from this:
```chpl
forall i in downdom do
  yield downIdxToUpIdx(i);
```

to this:
```chpl
for i in downdom.these(tag) do
  yield downIdxToUpIdx(i);
```

This is similar to what I had to do in the original ArrayViews effort to work around a very simliar issue.

benharsh:
It looks like you added debugging output in Block.dsiNewRectangularDom resulting in a failure for multipleDomains2.

benharsh:
Initial tests also suggest that the reduction workaround fixes the core dump as well.

bradcray:
Thanks for looking into these...  I don't think I was aware of the "using foralls to forward standalone iterators" bug and would've taken a long time to find it...  That seems worthy of an issue if we don't have one already.

benharsh:
I believe it is covered in #5223

bradcray:
Great, thanks for the pointer.

bradcray:
Here's where things stand at present for --local testing:

invalid argument types for domain slicing
-----------------------------------------
[Error matching compiler output for arrays/deitz/part3/test_rank_change3]
[Error matching compiler output for distributions/bharshbarg/stencil/rankChange]
[Error matching compiler output for distributions/robust/arithmetic/collapsing/test_domain_rank_change1]
[Error matching compiler output for distributions/robust/arithmetic/collapsing/test_domain_rank_change_OOB]
[Error matching compiler output for domains/bradc/rankChange]
[Error matching compiler output for domains/bradc/slices/sliceAllDegenerate]
[Error matching compiler output for domains/deitz/test_domain_rankchange]
[Error matching compiler output for users/aroonsharma/correlation]
[Error matching compiler output for users/aroonsharma/covariance]
[Error matching compiler output for users/aroonsharma/LinearAlgebra/mvt]
[Error matching compiler output for users/aroonsharma/LinearAlgebra/cholesky]

OOB
---
[Error matching program output for arrays/userAPI/rankChangeOps3to2D]

can't re-assign index set
-------------------------
[Error matching program output for distributions/robust/arithmetic/composite/test_composite1]
[Error matching program output for distributions/robust/arithmetic/composite/test_composite2]
[Error matching program output for distributions/robust/arithmetic/modules/test_module_Sort]

type mismatch in assignment
---------------------------
[Error matching compiler output for distributions/robust/arithmetic/kernels/hpl]
[Error matching compiler output for studies/hpcc/HPL/bradc/hpl-blc]
[Error matching compiler output for studies/hpcc/HPL/stonea/errors/writelnCausesError]
[Error matching compiler output for studies/hpcc/HPL/stonea/serial/hplExample5]

unable to resolve return type
-----------------------------
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Search]

new leaks
---------
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]

unresolved access
-----------------

modules/internal/DefaultRectangular.chpl
```diff
@@ -1926,12 +1926,13 @@ module DefaultRectangular {
...
-        const stride = viewDom.ranges(1).stride: viewDom.idxType,
-              start  = viewDom.ranges(1).first,
+        const viewDomDim = viewDom.dsiDim(1),
+              stride = viewDomDim.stride: viewDom.idxType,
+              start  = viewDomDim.first,
               first  = info.getDataIndex(start),
               second = info.getDataIndex(start + stride),
               step   = (second-first):chpl__signedType(viewDom.idxType),
-              last   = first + (viewDom.ranges(1).length-1) * step:viewDom.idxType;
+              last   = first + (viewDomDim.length-1) * step:viewDom.idxType;
```
bradcray: @benharsh: Would you review that this is correct (the logical change of converting .ranges(1) calls over to .dsiDim(1) calls...  I can't figure out how this worked before, or maybe it was just never called?  My change is obfuscated somewhat by my pulling the dsiDim query into a temp, knowing that the call isn't dirt cheap in rank-change land.

bradcray:
If I didn't mess up, there should be even fewer failures now, but we'll have to wait until morning to see, when the test machines are free and clear...

modules/internal/DefaultRectangular.chpl
benharsh: Looks fine to me. I think I had been relying on 'viewDom' being a DefaultRectangularDom, so ``.ranges`` would be a valid member.

benharsh:
From some manual testing of the errors listed, it looks like the memory leak and 'invalid argument types' are the only remaining errors of that list.

bradcray:
That's what I was hoping -- I'm doing a full test this morning to double-check that I didn't break anything new.  I'm saving the memory leak for last because I'm still shuffling around where and when up and down domains are created and tucked away.

Playground perf testing spiked for several tests... :(  My expectation was that setting up a rank change will be more expensive than it is currently on master, but that accesses/operations through rank change slices shouldn't be much more expensive.  So I'm curious whether the tests that got worse create lots of rank changes in an inner loop, or whether I've somehow made the operations themselves more expensive (and if so, which one).

I'm going to start off this morning looking at the invalid argument type cases, though.

benharsh:
It looks like iteration is more expensive, at least in the standalone case, because of the calls to ``downIdxToUpIdx`` . This is invoked because the array standalone does something like ```for i in privDom.these(tag) ...```

bradcray:
Full run came back as expected:

invalid argument types
----------------------
[Error matching compiler output for arrays/deitz/part3/test_rank_change3]
[Error matching compiler output for distributions/bharshbarg/stencil/rankChange]
[Error matching compiler output for distributions/robust/arithmetic/collapsing/test_domain_rank_change1]
[Error matching compiler output for distributions/robust/arithmetic/collapsing/test_domain_rank_change_OOB]
[Error matching compiler output for domains/bradc/slices/sliceAllDegenerate]
[Error matching compiler output for domains/deitz/test_domain_rankchange]
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]
[Error matching compiler output for users/aroonsharma/correlation]
[Error matching compiler output for users/aroonsharma/covariance]
[Error matching compiler output for users/aroonsharma/LinearAlgebra/mvt]
[Error matching compiler output for users/aroonsharma/LinearAlgebra/cholesky]

new leaks
---------
[Error matching compiler output for domains/bradc/rankChange]

bradcray:
Two failures remaining:

new leaks
----------
[Error matching program output for memory/sungeun/refCount/arrays-rankchange]

nil dereference
---------------
[Error matching program output for users/aroonsharma/LinearAlgebra/mvt]

compiler/resolution/functionResolution.cpp
```diff
@@ -8928,10 +8928,9 @@ static void insertUnrefForArrayReturn(FnSymbol* fn) {
           call->insertAtTail(unrefCall);
           FnSymbol* unrefFn = resolveNormalCall(unrefCall);
           resolveFns(unrefFn);
-
-          if (unrefFn->retType == tmp->typeInfo()) {
-            // If the types are equal, we must be dealing with a non-view
-            // array, so we can remove the useless unref call.
+          if (unrefFn->where == NULL) {
```
bradcray: @benharsh: Would using a pragma/flag be a more principled way of distinguishing the array view overload of chpl__unref() from the others?  (assuming I'm understanding this properly).  It seems we could have other chpl__unref()s down the line that will also have where clauses over time?

bradcray:
@mppf: I think this is ready for the early-morning review you offered.
@vasslitvinov: You're more than welcome to review it as well, though I didn't get it done during the weekend as expected (well, 20 minutes to spare...)
Note that the "changes by file" section in the PR message is also probably the best order in which to review the files.

test/arrays/views/viewPreserveDist.bad
```diff
@@ -1,4 +1,20 @@
...
+Y is: 0.0 0.0 0.0 0.0 0.2 0.2 0.2 0.2
+
+D is: {1..8}
+X is: 0.2 0.2 0.2 0.2 0.3 0.3 0.3 0.3
+Y is: 0.2 0.2 0.2 0.2 0.3 0.3 0.3 0.3
+
+D is: {0..15 by 2}
+X is: 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
+Y is: 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
+
```
mppf: I think I see what's going on here. The rank-change cases are now working, but the reindex case is not. I think it would make sense to split this test into a separate reindex-preserves-distribution future and let the rank-change cases be a normal test (that now passes, thanks to this PR).

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
   //
+  // This class represents a distribution that knows how to create
+  // rank-change domains and arrays similar to the one that caused
+  // it to be created.  It stores a pointer down to the distribution
+  // that it creates lower-dimensional views of, and two tuples
+  // indicating which dimensions were collapsed by the rank-change
+  // and what the indices of those dimensions are.
+  //
+  class ArrayViewRankChangeDist: BaseDist {
+    const downdist;
```
mppf: The comment on the class does a good job of explaining this, but I think it'd be nice to have above-line or end-of-line comments on these variables to say what they are expected to be.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+  //
+ class ArrayViewRankChangeDom: BaseRectangularDom {
+    param rank;
+    type idxType;
+    param stridable;
+
+    // the lower-dimensional index set that we represent upwards
+    var updom: DefaultRectangularDom(rank, idxType, stridable);
+
+    // the collapsed dimensions and indices in those dimensions
```
mppf: It's not a big deal either way, but is there a reason that these variables are repeated here, instead of using e.g. dist.collapsedDim every time? If there is, it'd be nice to say so in a comment. If not, that'd be nice to comment too (e.g. "these variables could be removed and uses point to dist.idx etc").

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+    const collapsedDim;
+    const idx;
+
+    const dist;  // a reference back to our ArrayViewRankChangeDist
+
+    // the higher-dimensional domain that we're equivalent to
+    var downdomPid:int;
```
mppf: I'm not sure it's worth changing before merging, but I would have expected it to be upDom and downDomPid etc. It's a little hard for me to read downdom.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+// while 'up*' refers to the lower-dimensional, user-facing view.
+//
+// TODOs for this file:
+//
+// * the dist, dom, and arr classes store a certain amount of
+//   duplicated information -- should these be focused in the higher
+//   classes (dist/dom) and referenced by lower ones (dom/dist)?
+//
+// * I don't think I've done anything to privatize the dist or dom
+//   classes... should I?
```
mppf: There are several things about this file that assume that the domain to be rank-changed is rectangular. That's not strictly necessary (nor is it for slicing?). It'd be nice to either clearly say that currently the language has no plans to support rank-change views of non-rectangular arrays/domains or to say that supporting such things is a TODO.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+
+    const dist;  // a reference back to our ArrayViewRankChangeDist
+
+    // the higher-dimensional domain that we're equivalent to
+    var downdomPid:int;
+    var downdomInst: downdomtype(downrank, idxType, stridable);
+
+    //
+    // TODO: If we put this expression into the variable declaration
+    // above, we get a memory leak.  File a future against this?
```
mppf: Now would be a good time to create such a future and remove the comment about filing one.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+            return downdom.dsiDim(d);
+        }
+      }
+      halt("Fell out of loop in dsiDim():", (upDim, downDim));
+    }
+
+    proc dsiDims() {
+      var upDims: rank*range(idxType, stridable=stridable);
+      var upDim = 1;
+      for param downDim in 1..downrank {
```
mppf: I'm curious if you know why this loops is a `for param` while the loop in `dsiDim()` is a `for`. It seems likely to me we'd want them to be the same in this regard. I'd probably remove the `param` from this one.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+        }
+      }
+      return upDims;
+    }
+
+    proc dsiGetIndices() {
+      return updom.dsiGetIndices();
+    }
+
+    proc dsiSetIndices(inds) {
```
mppf: I'm probably missing something, but it seems to me that assigning to the indices of a rank-change slice (or regular slice) of a domain should probably be a user error. If not... I'm not sure what it's supposed to do and so have trouble evaluating this routine.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+                                                       collapsedDim=collapsedDim,
+                                                       idx=idx);
+  }
+
+  //
+  // This class represents the domain of a rank-change slice of an
+  // array.  Like other domain class implementations, it supports the
+  // standard dsi interface.  Note that rank changes only make sense
+  // for rectangular domains so this is a subclass of
+  // BaseRectangularDom.
```
mppf: It seems to me that rank-change *could* make sense for an associative domain. Is it just that associative domains are always 1-D as far as the current implementation is concerned? Is that fundamental, or is it a quirk of how it is built?

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+        }
+      }
+    }
+
+    proc chpl_rankChangeConvertLoDTupleToHiD(tup) {
+      var tupHiD: downrank*tup(1).type;
+      var i = 1;
+      for param d in 1..downrank do
+        if collapsedDim(d) then
+          tupHiD(d) = 0..0;  // TODO: This seems bogus/arbitrary... or is it?
```
mppf: Shouldn't this be `tupHiD(d) = idx(d)` ?

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -22,9 +22,337 @@
...
+      return dist;
+    }
+
+    proc dsiTargetLocales() {
+      //
+      // BLC: To tighten this up, we'd need to query the distribution to
+      // see what subset of target locales the rank-change slice hit vs.
+      // not.
+      //
+      //      compilerWarning("Calls to .targetLocales() on rank-change slices may currently return a superset of the locales targeted.");
```
mppf: It's not clear to me why this compilerWarning is disabled.

modules/internal/ArrayViewRankChange.chpl
mppf: Is it basically just part of domain construction?

mppf: Ah, I see dsiNewRectangularDom in ArrayViewRankChangeDist calls it. Wouldn't it be better to name this method something else, so we can (potentially at least) have different behavior for assigning to a rank-change domain? Or at least to reduce confusion? E.g. finishInitializingWithIndices(inds).

Alternatively - maybe this code should be folded in to an alternative ArrayViewRankChangeDist constructor that accepts an 'inds' argument. But that might not go well until constructors/initializers are fixed.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -280,15 +571,15 @@ module ArrayViewRankChange {
       // see what subset of target locales the rank-change slice hit vs.
       // not.
       //
-      compilerWarning("Calls to .targetLocales() on rank-change slices may currently return a superset of the locales targeted.");
+      //      compilerWarning("Calls to .targetLocales() on rank-change slices may currently return a superset of the locales targeted.");
```
mppf: Not seeing why .targetLocales() warning went away here either.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -280,15 +571,15 @@ module ArrayViewRankChange {
...
       return arr.dsiTargetLocales();
     }
 
     proc dsiHasSingleLocalSubdomain() param
-      return privDom.dsiHasSingleLocalSubdomain();
+      return privDom.updom.dsiHasSingleLocalSubdomain();
```
mppf: What is here may well be correct but I'm missing something. I'm wondering in particular why ArrayViewRankChangeDom.dsiHasSingleLocalSubdomain doesn't return the property on updom, since updom is the "user-facing" domain we are working with. Is it because updom isn't distributed, but perhaps the rest is? Then shouldn't this routine on the array return privDom.dsiHasSingleLocalSubdomain() i.e. consider what the original rank change is over instead of always returning `true`  (or whatever DefaultRectangularDom does with this method) ?

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -280,15 +571,15 @@ module ArrayViewRankChange {
...
 
     proc dsiLocalSubdomain() {
-      return privDom.dsiLocalSubdomain();
+      return privDom.updom.dsiLocalSubdomain();
```
mppf: Same concern as in dsiHasLSingleLocalSuddomain.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -470,6 +762,75 @@ module ArrayViewRankChange {
...
+    param downrank = collapsedDim.size;
+    var ind = idx;
+    var j = 1;
+    for param d in 1..downrank {
+      if !collapsedDim(d) {
+        ind(d) = i;
+        j += 1;
+      }
+    }
+    assert (j == 2);
```
mppf: There seem to be too many of these methods. Could they call each other to make the relationship clearer? E.g. is this the i-isnt-a-tuple version? When is that called? If i isn't a tuple, is the number of dimensions necessarily 2?

And then, shouldn't the methods around line 260 such as chpl_rankChangeConvertLoDTupleToHiD, chpl_rankChangeConvertHiDTupleToLoD, downIdxToUpIdx just call methods down here?

modules/internal/ArrayViewRankChange.chpl
mppf: Or maybe it is `tupHiD(d) = idx(d) .. idx(d);`

modules/internal/ChapelArray.chpl
```diff
@@ -1066,59 +1102,66 @@ module ChapelArray {
...
-        // Otherwise, we don't have a way for the var d below
-        // to extend the lifetime of the distribution...
-        halt("Distribution must use trackDomains or be singleton");
-      }
-      var j = 1;
-      var makeEmpty = false;
+      //
+      // Compute which dimensions are collapsed and what the index
+      // (idx) is in the event that it is.  These will be stored in
+      // the array view to convert from lower-D indices to higher-.
```
mppf: Maybe the comment should end with "higher-D"

modules/internal/ChapelArray.chpl
```diff
@@ -1066,59 +1102,66 @@ module ChapelArray {
...
+
+      // Create distribution, domain, and array objects representing
+      // the array view
+      var upranges = _getRankChangeRanges(args, this._value);
+      const emptyrange: upranges(1).type;
+      param uprank = upranges.size;
+      //
+      // If idx isn't in the original domain, we need to generate an
+      // empty upward facing domain (intersection is empty)
+      //
```
mppf: Shouldn't that case be an error, when checks are on?

mppf: Also just calling .member doesn't seem right to me to do the check. Consider
``` chapel
  var D = {1..10, 1..10};
  var X = D[ 2, 0..50 };
```

shouldn't that effectively create the rank change domain with `idx = (2, 1)` ? Or in other words do the same thing as:
``` chapel
  var D = {1..10, 1..10};
  var X = D[ 2, 1..10 };
```

?

modules/internal/ChapelArray.chpl
```diff
@@ -1066,59 +1102,66 @@ module ChapelArray {
...
+          upranges(d) = emptyrange;
       }
-      var d = {(...newRanges)} dmapped newDist;
-      return d;
+
+      const rcdist = new ArrayViewRankChangeDist(downdist = dist,
+                                                 collapsedDim=collapsedDim,
+                                                 idx = idx);
+      const rcdistRec = _newDistribution(rcdist);
+      //      rcdist._free_when_no_doms = true;
```
mppf: I think we can remove this commented-out code, right?

modules/internal/ChapelArray.chpl
```diff
@@ -1066,59 +1102,66 @@ module ChapelArray {
...
+    proc this(i: integral ... rank) {
+      compilerError("domain slice requires a range in at least one dimension");
     }
 
     // anything that is not covered by the above
     pragma "no doc"
     proc this(args ...?numArgs) {
       if numArgs == rank {
         // Doing this just to get a better compiler error
-        var ranges = _getRankChangeRanges(args);
```
mppf: Let's remove the comment above it too.

modules/internal/ChapelArray.chpl
```diff
@@ -2136,50 +2142,28 @@ module ChapelArray {
...
-        if (isRange(args(i))) {
-          collapsedDim(i) = false;
-        } else {
-          collapsedDim(i) = true;
-          idx(i) = args(i);
-        }
-      }
+      pragma "no auto destroy"
+      const rcdom = this.domain[(...args)];
+      rcdom._value._free_when_no_arrs = true;
```
mppf: It seems likely that removing pragma "no auto destroy" on rcdom and removing the _free_when_no_arrs would simplify things and still be correct. The rcdom will be destroyed at the end of this fn, but because add_arr was called, it will end up not being destroyed yet and will set _free_when_no_arrs = true.

modules/internal/ChapelArray.chpl
mppf: I see now that upranges will include the intersection with the original domain. I also see that _getRankChangeRanges seems somehow too complicated. Wouldn't it be simpler to just create a tuple of ranges (with e.g. 1..1 for the collapsed one) and then intersect that with the original domain's ranges? And then collapse that tuple by leaving out the collapsed dimensions (similarly to chpl_rankChangeConvertIdxHiDToLoD etc) ?

Either way, I really don't like _getRankChangeRanges. Isn't there some way we can do the equivalent work in a for loop?

modules/internal/ChapelArray.chpl
```diff
@@ -3163,36 +3147,36 @@ module ChapelArray {
...
-  proc _getRankChangeRanges(args) {
-    proc _tupleize(x) {
-      var y: 1*x.type;
-      y(1) = x;
-      return y;
-    }
+  proc _getRankChangeRanges(args, downdom) {
+    return collectRanges(1);
+
     proc collectRanges(param dim: int) {
```
mppf: See my note around 1137. I think collectRanges is unnecessarily complicated and could be much simpler in a for-loop form and in a form that separated the matter of intersecting from the matter of leaving out the collapsed dimensions. Moreover, since this is only ever called from the domain rank-change `this`, it seems odd to have it all the way down here. Could it be a utility function in ArrayViewRankChange.chpl ? Or an inner function in the `this` method? Or just a series of loops that is easier to follow?

modules/internal/ChapelDistribution.chpl
```diff
@@ -90,6 +90,10 @@ module ChapelDistribution {
       return (count==0);
     }
 
+    //
+    // TODO: Are there opportunities to optimize out the locking here,
+    // as on the add_arr() case?
```
mppf: I'll both answer the question and request you add the answer to the comment.
We could optimize for example the construction of a domain rank change because that operation creates a new distribution. It's unclear how important this optimization is, though, because domain rank change is much less common than array slicing. However, adding the optimization would amount to adding a default argument e.g. `lockingDist=true` to `newRectangularDom`; then passing that to `add_dom`; and finally updating domain rank change `this` method to pass `lockingDist=false` to `newRectangularDom`.

AFAIK this domain rank change is the only case we could improve by doing this.

modules/internal/DefaultRectangular.chpl
```diff
@@ -627,11 +621,13 @@ module DefaultRectangular {
     proc dsiHasSingleLocalSubdomain() param return true;
 
     proc dsiLocalSubdomain() {
-      return _newDomain(this);
+      // TODO: this should only return this way if this.locale == here
```
mppf: Do you intend to fix this before merging? What is the impact of having it wrong as it is now?
It sounds trivial to fix. Is there some reason to hesitate in adding the comparison?

compiler/resolution/functionResolution.cpp
benharsh: I was hoping to find a cleaner way than yet another super-special flag, but at this point a flag may be best.

mppf: We could teach the compiler to identify an array view _array, but that seems like more work.
If we need to distinguish the array-view unref from the array one, I think a flag would be a reasonable way to do so, for now at least.

But, I'm not yet understanding why this change was necessary. Is there some pattern where we can create an array over a sliced domain e.g. where you wanted to change this behavior? Checking for the `where` clause doesn't seem right. Can you explain the problematic case that caused you to make this change?

mppf: re.  *I ran into odd iterator issues if I removed the chpl__unref for normal arrays* - that is because of the way tuples need chpl__unref - consider a function accepting a tuple argument containing two arrays. The arrays will be passed by reference. If it returns the tuple, the returned tuple needs to store the arrays by value.

mppf: Re. the rationale here, I see that Ben wrote
>   chpl__unref, given a rank-changed array, will now return an array
>   declared over a rank-changed domain. This violates some compiler logic
>   that inserts the chpl__unref. If the types were the same, then the
>   chpl__unref would be removed.

So, I think reasonable solutiions here are:

 * teach the compiler to identify an array-view _array, and filter on that. An advantage there is that we could know if we were adding a `chpl__unref` call immediately and wouldn't need to resolve it first.
 * create a different function, e.g. `chpl__arrayview_unref` that does the same thing as the current chpl__unref with the where clause. Then call that function here, but only if it resolves. Leave the other `chpl__unref` there to correctly handle passing an array view into a function in a tuple argument and then returning the tuple.

mppf:
I've completed my initial review. I'd like to see the feedback addressed before merging / in the next few days. Please ask if it's not clear which of my requests I'd view as requiring attention in the near term.

benharsh:
The StencilDist changes look fine to me.

test/arrays/views/viewPreserveDist.bad
bradcray: Yes, you're interpreting that correctly.  I could split it out, though the .bad should preserve the rank-change case working until reindex works (which should be any day now...).

mppf: If reindex is about to start working any day now then there's less relative value in splitting it apart. I was assuming it would not be working for the release.

modules/internal/ArrayViewRankChange.chpl
bradcray: I had a general TODO to this effect at the top of the file but have added an additional one here.  The main reason I didn't do it is because, by the time I got everything working, leak-free, and performing well, I was too scared to change anything else for fear of breaking something -- most likely performance (this is a pretty nervous reason, though... keep in mind that this PR was finally "ready" at 11:40pm last night... :)  ).

modules/internal/ArrayViewRankChange.chpl
bradcray: I kept typing it the wrong way too, so I'll change it.  Not sure why I set off down this path.

modules/internal/ArrayViewRankChange.chpl
bradcray: A comment higher in the function (on the Dom class) asserts that this only makes sense on rectangular.  I don't think rank change makes sense for non-rectangular domains/arrays because associative and opaque are arguably rank-1 (so rank-change could only reduce them to rank-0).

I think reindexing similarly doesn't make sense for non-rectangular because it would essentially require O(numIndices) space to store the map from old to new indices.

As you note, slicing makes sense for all three and I believe the slicing implementation doesn't do anything to prevent slicing a dense with a sparse, an associative with another associative, etc. (though I haven't tested many of those cases yet; though it's on my TODO list and the testing system suggests we at least haven't broken anything we'd verified as working).

mppf: OK, what would help is to put "because they are rank-1" next to the assertion earlier in the file that this only matters for rectangular domains. thanks!

modules/internal/ArrayViewRankChange.chpl
bradcray: I'm having trouble creating one in short order, but have added it to the TODO list above as desirable.

modules/internal/ArrayViewRankChange.chpl
bradcray: No strong reason I think.  I tend to use 'param' loops for ones over params that are expected to be small (like the # of dimensions in a domain/array), but apparently not consistently.

I think I can make this even simpler, though.  Re-reviewing this last night while working on reindexing, I found myself wondering why I didn't just pass the dsiDIm*() calls to updom where, arguably, all this work has already been done.  My best guess is that I wrote this routine before introducing updom and never thought to reconsider it.

modules/internal/ArrayViewRankChange.chpl
bradcray: You're correct that we haven't traditionally distinguished between the call that initializes a domain's indices and the one that re-assigns it -- both use dsiSetIndices() today.  I think that distinguishing those cases would be an interesting change to make (and one that we could make today), but don't think I should make it on this PR (even if time weren't a factor).  There's also a little bit of a question about what we would do differently in the two routines.

Note that I don't think dsiSetIndices() could be used to change the domain of a rank-change view array created by slicing an array because the only way that domain could be fetched is by applying a .domain to the array which returns a const ref to the domain.  But dsiSetIndices() could still be used for either initialization or assignment scenarios on a rank-change domain slice as follows:

````chapel
var D = {1..n, 1..n} dmapped Block(...);  // create a 2D domain
var A: [D] real;  // create an array over it
var X = A[3, ..];  // create X, a 1D view of 2D A

// create a new domain that has the same distribution as the rank-change view X.
// Note that this requires asking D's 2D block distribution to create a new 2D domain
// and for us to create a rank-change 1D view of it (because D's domain  map only knows
// how to create 2D things):
var D2 = {3..5} dmapped X.domain.dist;   // this will use dsiSetIndices() to establish D's indices
D2 = {9..11};  // D2 can be reassigned which will also call dsiSetIndices()
````

modules/internal/ArrayViewRankChange.chpl
bradcray: I can't figure out any way that rank-change slicing would make sense for associative because it doesn't have a notion of dimensions that can be collapsed out (essentially they're 1D as you guess).  If you have a vision for rank-change for associative, you'll need to clarify that for me.

In any case, I don't believe I've removed any capability that previously existed on master in this effort.

modules/internal/ArrayViewRankChange.chpl
bradcray: I've been meaning to write a test to see if this is buggy or not.  I'll add it to the TODO list.  I was guessing it needed to be something like that too, put this in as a placeholder in my first draft of the code for reasons I don't remember and was surprised it never failed.  I haven't thought through it enough to determine whether that means we're not testing this routine (likely... I'm working on a userAPI test like the one I wrote for arrays to make sure we exercise all routines in a simple way) or whether the fact that the dimension is getting compressed out anyway means that the index doesn't matter.

modules/internal/ArrayViewRankChange.chpl
bradcray: I added it in the original array view work, but commented it out here because on reflection I didn't think it was all that helpful.  Slicing may similarly restrict which locales own a piece of the view, but doesn't have a similar warning.  In the case of privatizing a distributed array, I also wasn't certain whether the answer to this query should be "everyone who owns a privatized copy of the array's class" or "only those locales that own non-zero data."  For complex distributions, it also seemed as though it might be challenging/expensive to determine exactly which target locales own a slice.

Mostly, I felt dissatisfied with the definition of targetLocales and didn't feel confident that preserving the warning here had value.

bradcray: Oh, I mis-spoke... the warning was added in the array class in the original array views work.  Then I copied and pasted its definition into the domain class, then I commented both out.

modules/internal/ArrayViewRankChange.chpl
bradcray: Hmm, good catch.  This looks like a bug to me and I'm guessing reveals another routine that isn't tested by the test system at present (but which my userAPI test would catch.  Finishing that test was going to be code freeze week work).  Let me write a test of this specific case before merging.

modules/internal/ArrayViewRankChange.chpl
bradcray: Yes, this is the i-isn't a tuple case and I believe it's called for 1D cases in which idxType is a scalar rather than a 1-tuple.

I can look into refactoring the conversion routines to make better use of each other.

modules/internal/ChapelArray.chpl
bradcray: > Shouldn't that case be an error, when checks are on?

No, domain slicing (like range slicing, but unlike array slicing) computes intersection and doesn't involve a bounds check, so if {1..10, 1..10, 1..10} is sliced with {3, .., 0} the right answer needs to be an empty domain rather than the {1..10} we'd get if we sliced with, say, {3, .., 3}.  This was a bug that was revealed at some point late in the original rank change support and which almost was re-enabled in my original draft of this PR except that the single test guarding against it failed.

> Either way, I really don't like _getRankChangeRanges. Isn't there some way we can do the 
> equivalent work in a for loop?

I don't believe it can use a for loop if it continues to mix scalars and singletons (at least, not without a similarly recursive thing to determine what the resulting tuple type would be).  If it returned ranges in all positions as you suggest then we may not need the recursion, which I agree would be attractive.  This is pretty old code that I've adapted, so I'll look into whether there's any way it can't just use ranges + collapsedDim-style info.

modules/internal/ChapelArray.chpl
bradcray: Yeah, I left it in because I remain somewhat snowblind about when #pragma no destroy should/shouldn't be used in these types of routines and when the free_when_no_doms flag should be used and remained a little surprised that I shouldn't apply them here.  A lot of my work to close memory leaks amounted to randomly applying one, the other or both, to variables until the leaks went away... :S

modules/internal/ChapelDistribution.chpl
bradcray: Note that array rank change slicing also requires a domain rank change slice, so this optimization would help any rank change slice, not just those on arrays.  Arguably rank-change slicing is less common than normal slicing, though (and more expensive in the distributed setting), so the "less common" comment may still apply.  I've added a modified comment to that effect.

modules/internal/DefaultRectangular.chpl
bradcray: I agree it's easy but I didn't want to add it to this PR.  I'll try to open up a new PR or issue for it in order to remove this TODO.

modules/internal/ArrayViewRankChange.chpl
mppf: I'd view the result of our discussion as "It's fundamental to associative domains that they don't support rank change because we only ever plan to have 1-D associative domains".

modules/internal/ArrayViewRankChange.chpl
mppf: OK, that's fine. So maybe we should remove the commented-out code, and perhaps put a sentence or two about how it's actually not necessarily reasonable to have a warning in this case?

modules/internal/ChapelArray.chpl
mppf: It doesn't look to me like _getRankChangeRanges actually returns tuples mixing ranges and integers. It looks like it just returns a tuple of ranges (in a complicated way). Anyway, I'll be much happier if you can clean it up somehow.

modules/internal/ArrayViewRankChange.chpl
mppf: sounds good.

modules/internal/ArrayViewRankChange.chpl
mppf: At the very least, I expect if this code is actually run it could cause type check errors if idxType is not int.

bradcray: What I'm finding is:
* if idxType is int, the follower works as expected
* if idxType is not int, lots of other things fall apart before ever getting here

For now, I'm going to add a TODO to try and fix bugs related to idxType != int and remove the comment since I'm more confident that it's correct.  Keep in mind that this is in the context of a leader/follower iterator's 'followThis' argument and that we 0-shift and densify the index sets in such cases so that arrays/domains of the same shape/size but different index sets can be zipped.  So for a singleton dimension, I believe that 0..0 is correct.  Though I don't recall offhand, I'd guess that we also have the 'followThis' argument be represented using 'int' ranges regardless of idxType so that arrays/domains of different index types can be zipped together (and arguably we could/should use 'uint' since we're not using negative numbers in this context and _may_ want the full range of uints available to us.  But that's independent of this PR.

modules/internal/ChapelArray.chpl
bradcray: You're right, I was mis-remembering.  The thing it's doing that warranted recursion was creating a tuple of ranges that eliminates any collapsed dimensions.  I think that could be replaced by a recursive function that determines how many dimensions are collapsed as a param and then declaring a tuple of that size (at least, I can't think of another way to do it offhand since params aren't re-assignable).  This approach could also lead to the possibility of storing collapsedDim as a compute-on-demand param.

test/arrays/views/viewPreserveDist.bad
bradcray: See https://github.com/chapel-lang/chapel/pull/5757

PR #5696, opened 2017-03-16
Fix BLAS.trsv test 

ben-albrecht:
The test for `BLAS.trsv`, which solves `A*x = b` for `x`, did not check non-singularity on the randomly generated input matrix, resulting in sporadic failures whenever near-singular matrices were generated. 

This PR resolves this issue by generating a random invertible matrix with the following formula:

```
(1/n**2)*A**2 + I
```
where `A` is a randomly generated matrix, `n` is dimension of `A`, and `I` is the identity matrix.

I've also returned the `errorThresholdSingle` to it's original value of `1.e-5`, since we originally bumped it to `1.e-4` to overcome this specific sporadic failure.

- [X] Tested on OS X + netlib
  - Ran `test_blas*` 1000 times, to confirm sporadic failure is eliminated.

### DIFF
--- a/test/modules/packages/blas/blas_helpers.chpl
+++ b/test/modules/packages/blas/blas_helpers.chpl
@@ -158,9 +158,7 @@ proc bandArrayTriangular(A: [?Dom] ?eltType, k, uplo=Uplo.Upper, order=Order.Row
   where A.rank == 2 {
 
   // Must be square matrix
-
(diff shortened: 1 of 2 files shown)


ben-albrecht:
A community physicist has suggested a more elegant solution, which I have just committed.

PR message updated and re-tested.

PR #5701, opened 2017-03-17
Switch array view perf playground over to my new branch

bradcray:
This will see whether the new locality-preserving rank change domains
will hurt performance for local testing (they're definitely higher
overhead to set up).


bradcray:
@ronawho, can you double-check me on this?

ronawho:
👍 Note that this will keep the historical data too (I'm assuming that's what you want, but just double checking)

PR #5702, opened 2017-03-17
Whitespace changes to initializer rules

noakesmichael:
This is a simple merge of a few "whitespace" changes I have collected as I begin to work on the
issues with initializing class/record fields within initializer methods.

Passed start_test on classes/initializers -futures and a single-locale paratest

### DIFF
--- a/compiler/passes/initializerRules.cpp
+++ b/compiler/passes/initializerRules.cpp
@@ -45,25 +45,27 @@ static void     phase1Analysis(BlockStmt* body, AggregateType* t);
 *                                                                             *
 ************************************** | *************************************/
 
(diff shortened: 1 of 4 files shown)


compiler/passes/initializerRules.cpp
```diff
@@ -295,29 +306,32 @@ void phase1Analysis(BlockStmt* body, AggregateType* t) {
...
           }
           curExpr = getNextStmt(curExpr, body, false);
           isInit = getInitCall(curExpr);
           continue;
         }
-      }/* else if (CondStmt* cond = toCondStmt(curExpr)) {
-        // Special handling for conditionals.
-        }*/
-
+      }
```
lydia-duncan: I would like this to stay until we have the conditional statement support, please

PR #5704, opened 2017-03-17
Fixed a bug in string.split()

vasslitvinov:
split() used to drop the last word in a string
if the word was just a single character.
This changes fixes that.

While there, I converted some vars and the loop bound to consts,
for enhanced optimization opportunities.

Testing: linux64 --verify


psahabu:
I wrote this code initially. This looks good to me and can be merged now, thanks.

Issue #5675, opened 2017-03-15
Add some missing testing for the Cray module


ronawho:
Some test configurations that are missing for the cray module:

 - [ ] Add minimal registration testing for ugni
 - [ ] Beef up llvm testing
 - [ ] Add --dynamic correctness testing

ronawho:
We should also add some testing with `--dynamic` to ensure https://github.com/chapel-lang/chapel/pull/3541 doesn't break. (Currently it's broken for gasnet-aries because we're not brining in udreg)

bradcray:
@ronawho: Can you give me a sense of where this one stands / how important you think it is?  (I'm having trouble even parsing it).  Thanks.

ronawho:
I updated the summary -- we still don't have testing for these configurations, but I guess you can close it if you want.

More llvm testing will fallout when we work towards making it the default, and minimal NIC registration isn't terribly important. `--dynamic` testing would be nice to make sure we're not breaking things, but `--static` is the default at most sites, so we could continue to punt on it

bradcray:
I'm not looking to close it per se, more like trying to understand how important we consider it (e.g., should it be in the backlog (do sooner) or icebox (do later).

ronawho:
Ah ok, I had misinterpreted the accidental closing of https://github.com/chapel-lang/chapel/issues/5457 to mean "we're trying to close unimportant issues" -- icebox for sure

ronawho:
I'm going to close this -- we have a few tests that run without hugepages, and I think that's enough of a sanity check fo that configuration. Dynamic linking is now the default on crays, so I think that's covered too. 

We have other efforts to beef up llvm testing and make it the default so I don't think we need to track that here.

PR #3921, opened 2016-05-25
 Upgrade the GASNet package from 1.26.0 to 1.26.3.

gbtitus:
This contains a variety of updates and bugfixes in the udp, ibv, and
gemini/aries conduits in 1.26.2, and some additional fixes for udp
conduit bugs in 1.26.3.  (None of these bugs were affecting us as
far as we know.)  See gasnet-src/ChangeLog for details.

While here, I also moved a slightly modified copy of the upgrade
instructions from $CHPL_HOME/third-party/README.devel into
$CHPL_HOME/hhird-party/gasnet/README, as we expect it might
be a bit easier to find in the latter.

PR #3541, opened 2016-03-22
Fix dynamic linking with cray PrgEnvs

ronawho:
On crays we had issues with dynamic linking (i.e. CRAYPE_LINK_TYPE=dynamic)
that resulted in undefined references to PMI and GNI routines. CrayPE/2.X
requires that dynamic libraries know their dependencies, but we weren't
specifying that Chapel requires pmi and ugni.

With static linking, libraries must specify the link args that they provide and
the ones that they require. We happened to get lucky because cray-mpich
satisfied our pmi and ugni dependencies.

If you tried to compile any chapel program with CRAYPE_LINK_TYPE=dynamic or
with cray-mpich unloaded, you'd get link time errors about undefined references
to PMI and GNI functions.

To resolve this, we now specify the PKGCONFIG_LIBS that are required by ugni
and gasnet-{gemini, aries} in our cray-prgenv makefile

PR #5705, opened 2017-03-17
Fix dynamic linking for gasnet-aries

ronawho:
Fix dynamic linking for gasnet-aries by linking in udreg. #3541 fixed dynamic
linking for cray PrgEnvs by telling the cc driver that comm gasnet-aries and
ugni require cray-pmi and cray-ugni. This commit is similar, but now we also
require cray-udreg for gasnet-aries. GASNet 1.26.2 added support for using
udreg, so we need to make sure we link it in.

From the 1.26.2 release notes:
     "Use Cray's UDREG library to ammortize cost of dynamic registration"

This has been a bug for almost a year (#3921), but since we don't have
--dynamic testing on crays, we missed this. We should really add some testing
for this configuration: https://github.com/chapel-lang/chapel/issues/5675

@dmk42 discovered this while trying to test the module with KNL while memkind
was loaded (memkind forces dynamic linking.) Note that this only seems to be an
issue under CLE6, when chapel was built on CLE5 (like with the module.) I can
reproduce this by building on CLE5 and compiling on CLE6.

### DIFF
--- a/make/compiler/Makefile.cray-prgenv
+++ b/make/compiler/Makefile.cray-prgenv
@@ -81,7 +81,7 @@ export PE_CHAPEL_PKGCONFIG_LIBS := cray-pmi:cray-ugni:$(PE_CHAPEL_PKGCONFIG_LIBS
 endif
 ifeq ($(CHPL_MAKE_COMM), gasnet)
 ifneq (,$(filter $(CHPL_MAKE_COMM_SUBSTRATE),gemini aries))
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus and @dmk42 just fyi. I don't think this needs a review, and I'll be merging shortly.

Tested by building on a CLE 5 machine, and compiling on a CLE 6 machine. w/o this --dynamic only failed on the CLE 6 machine. 

With this I verified that --static and --dynamic work on both CLE 5 and CLE 6 machines.

PR #5710, opened 2017-03-17
Improve inferConstRefs' canRHSBeConstRef helper function

benharsh:
This helper function would look at a certain set of primitives on the
RHS of a move/assign and determine if the LHS was const-ref. If so, then
the RHS use could also be const-ref.

However, we were only returning in this case if we knew the LHS was not
const-ref, then dropping into the default case where we would check if
the RHS was used in a 'safe' primitive. Generally, though, the first set
of primitives are not considered 'safe', and we would return false even
though the LHS was indeed const-ref.

To fix this, always return the result of inferConstRef(LHS) for the
initial set of primitives.

Testing:
- [x] full local
- [x] full gasnet
- [x] dist-cyclic
- [x] dist-replicated for distributions/robust/arithmetic/performance/multilocale/assign


mppf:
Looks good, thanks!

PR #5712, opened 2017-03-17
fix --strict-error docs

psahabu:
Fixes the man page issue created by #5655.


ronawho:
This looks fine from a quieting regressions perspective, but @bradcray _might_ care about the ordering?

bradcray:
I think it probably makes better sense as a misc. flag than a run-time semantic check flag.  Ultimately, we may want to make a new "compile-time semantic check" section that this could go in, but let's wait to get more experience with it (or, really, end up using code-based strictness cues) before doing that.

Note that the flags are currently alphabetized (ignoring any [no] prefixes, I think), so it'd be preferable to maintain that order in both the help output and man page.

compiler/main/driver.cpp
```diff
@@ -723,6 +722,7 @@ static ArgumentDescription arg_desc[] = {
  {"print-callgraph", ' ', NULL, "Print a representation of the callgraph for the program", "N", &fPrintCallGraph, "CHPL_PRINT_CALLGRAPH", NULL},
  {"print-callstack-on-error", ' ', NULL, "print the Chapel call stack leading to each error or warning", "N", &fPrintCallStackOnError, "CHPL_PRINT_CALLSTACK_ON_ERROR", NULL},
  {"set", 's', "<name>[=<value>]", "Set config param value", "S", NULL, NULL, readConfig},
+ {"strict-errors", ' ', NULL, "Enable strict mode for error handling", "F", &fStrictErrorHandling, NULL, NULL},
```
bradcray: I wouldn't change this today, but it seems inconsistent that this flag doesn't have a [no]- version such that one could override a default set  by an environment variable, an earlier flag, etc.

Issue #5068, opened 2016-12-17
avoid term 'lvalue' in error messages


bradcray:
### Summary of Problem

The compiler currently uses the term 'lvalue' in its error messages.  I think we should stop doing this.

### Steps to Reproduce

**Source Code:**

From test/variables/constants/assignConstError.chpl added in PR #5067 :

```chapel
const two = 2;
two = 3;
```

**Compile command:**
`chpl assignConstError.chpl`

**Execution command:**
N/A

### Configuration Information

- Output of `chpl --version`: 1.14.0.7d05050
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: N/A
- (For Cray systems only) Output of `module list`: N/A

lydia-duncan:
I wholeheartedly support moving away from "lvalue" in error messages

bradcray:
I threw the "easy" tag on here because I think it is (modulo agreeing on what the new error messages should say... but lots of things would be better than what we have today).

dmk42:
I agree with the concept for general users, though seeing the term lvalue in an error message saved me some work the other day, since it pinpointed exactly what was wrong.  It will be difficult to come up with another term that works as well for that purpose.

ben-albrecht:
I don't think this should have an `[easy]` label until we agree upon a replacement, which doesn't seem to be happening anytime soon.

bradcray:
I think the person who owns this issue will need to sort through the error messages and propose alternatives based on the context (e.g., things like 'const', 'param', or 'read-only' may be appropriate depending on the context).  I still assert that this is an easy task (including the proposal) but I'm trying to hold restrain myself from getting caught up into a label wrestling war.

PR #5714, opened 2017-03-17
alternative of 'lvalue' in error messages #5068

krishnakeshav:
If finding the alternative for original 'lvalue' statement was the only purpose of #5068 , I guess we need to come up with more relevant names.


ben-albrecht:
Hi @krishnakeshav - I think the team needs to agree upon a replacement for this before proceeding. As @daviditen points out in #5068, the existence of 'lvalue' in the error message is still useful today. 
Sorry if the premature `[easy]` label misled you.

PR #5717, opened 2017-03-17
Linear Algebra module

ben-albrecht:
# Linear Algebra module

The small feature set in this package mainly consists of:
- matrix/vector convenience initializers
    - e.g. `var A = Matrix(4, 4);` , `var v = Vector(3);`
- matrix structure functions
    -  e.g.`var d = triu(A)`
- BLAS-supported matrix|vector operations
    - e.g. `var AB = dot(A, B);`

The feature coverage is fairly minimal, as I have dropped LAPACK support for this release cycle, which is required for a large number of routines. As a result, this is still considered a prototype package module.

Some points of discussion with reviewer(s):

- Indexing
   - module creates new arrays with 0-based indexing unless a user specifies otherwise (with a domain of their own)
- Naming schemes
    - e.g. casing on array initializer wrappers: `Matrix()` and `Vector()`.
- Are the matrix operations that purely exist to create temp arrays and override promotion flattening valuable enough to include?
   - `matPlus`, `matMinus`, ...
   - See the documentation on promotion flattening if this doesn't make sense...
- Implementations
   - Are we OK with naive implementation for some routines, or would we prefer to not support them yet at all?
- Array methods
    - Are we OK with a package module adding base array methods as syntactic sugar?
        - This PR only introduced `_array.T`, but we could support many others down the road.
        - See the [numpy array documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.html) to get an idea of what kind of array methods we could utilize.

 
**Remaining TODOs**
- [x] Finish shape-assertions in multiplication operations
- [x] Finish Matrix structure tests
- [x] Testing
    - [x] OS X + netlib
    - [x] Cray + CrayBLAS

### DIFF
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -22,7 +22,8 @@
 A high-level interface to linear algebra operations and procedures.
 
 .. note::
-  This is
+  This is an early prototype package module. As a result, interfaces may change
+  over the next release.
 
 Compiling with Linear Algebra
 -----------------------------
@@ -32,8 +33,6 @@ many routines. In order to compile a Chapel program with this module, be sure
 to have BLAS and LAPACK implementations available on your system. See the BLAS
 and LAPACK documentation for further details.
 
-.. comment:
-  - native implementations some day
 
 Linear Algebra API
 ------------------
@@ -81,37 +80,62 @@ An outer product can be computed with the :proc:`outer` function.
 
 **Promotion flattening**
 
-``TODO``
+Promotion flattening is an unintended consequence of Chapel's
+:ref:`promotion feature <ug-promotion>`, when a multi-dimensional array assignment
+gets type-inferred as a 1-dimensional array of the same size, effectively
+flattening the array dimensionality. This can result in unexpected behavior in
+programs such as this:
 
-Discuss promotion flattening, and how to get around it, providing examples
-and helper functions like :proc:`matPlus`.
+.. code-block:: chapel
 
-.. note::
-    Promotion flattening is not an intended feature, and should be resolved in
-    future releases of Chapel.
+  var A = Matrix(4, 4),
+      B = Matrix(4, 4);
+  // A + B is a promoted operation, and C becomes a 1D array
+  var C = A + B;
+  // This code would then result in an error due to rank mismatch:
+  C += A;
 
-.. Examples:
+To get avoid this, you can avoid relying on inferred-types for new arrays:
 
-.. Features:
+.. code-block:: chapel
 
-.. Future directions:
+  var A = Matrix(4, 4),
+      B = Matrix(4, 4);
+  // C's type is not inferred and promotion flattening is not a problem
+  var C: [A.domain] A.eltType = A + B;
+  // Works as expected
+  C += A;
+
+Alternatively, you can use the LinearAlgebra helper routines, which create
+and return a new array:
+
+.. code-block:: chapel
+
+  var A = Matrix(4, 4),
+      B = Matrix(4, 4);
+  // matPlus will create a new array with an explicit type
+  var C = matPlus(A, B);
+  // Works as expected
+  C += A;
+
+Promotion flattening is not expected to be an issue in future releases.
+
+
+.. LinearAlgebra Module Future TODOs:
+  - Support non-matching domains and eltTypes in matOperations
+    - check that eltTypes are coercible
+    - check domain shapes are equal
+  - Add LinearAlgebra primer
+  - Add feature table, comparing to numpy and matlab
+  - performance testing
   - Domain maps
     - Distributed array support
-      - PBLAS
     - Sparse support
-      - suiteSparse
     - GPU support
-      - CuBLAS
   - Support fully native implementations
     - Only use the BLAS/LAPACK version when the libraries are installed
   - Provide MKL BLAS/LAPACK with Chapel installation
     - install with something like: CHPL_EXTRAS=MKL
-
-.. LinearAlgebra Module TODOs
-  - Support non-matching domains and eltTypes in matOperations
-    - check that eltTypes are coercible
-    - check domain shapes are equal
-  - performance testing
 
 */
 
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -388,20 +388,16 @@ proc dot(a, B: []) where isNumeric(a) {
 }
 
 
-pragma "no doc"
 /* Explicit matrix-(matrix|vector) multiplication */
 private proc matMult(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
   // matrix-vector
   if Adom.rank == 2 && Bdom.rank == 1 then
-    // TODO -- assert shapes are correct
     return _matvecMult(A, B);
   // vector-matrix
   else if Adom.rank == 1 && Bdom.rank == 2 then
-    // TODO -- assert shapes are correct
-    return _matvecMult(B, A);
+    return _matvecMult(B, A, trans=true);
   // matrix-matrix
   else if Adom.rank == 2 && Bdom.rank == 2 then
-    // TODO -- assert shapes are correct
     return _matmatMult(A, B);
   else
     compilerError("Rank sizes are not 1 or 2");
@@ -410,14 +406,19 @@ private proc matMult(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
 
 pragma "no doc"
 /* matrix-vector multiplication */
-private proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
+private proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType, trans=false)
   where isBLASType(eltType)
 {
   if Adom.rank != 2 || Xdom.rank != 1 then
     compilerError("Rank sizes are not 2 and 1");
+  if Adom.shape(2) != Xdom.shape(1) then
+    halt("Mismatched shape in matrix-vector multiplication");
+
+  var op = if trans then BLAS.Op.T
+           else BLAS.Op.N;
 
   var Y: [Xdom] eltType;
-  gemv(A, X, Y, 1:eltType, 0:eltType);
+  gemv(A, X, Y, 1:eltType, 0:eltType, opA=op);
   return Y;
 }
 
@@ -429,6 +430,8 @@ private proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
 {
   if Adom.rank != 2 || Bdom.rank != 2 then
     compilerError("Rank sizes are not 2");
+  if Adom.shape(2) != Bdom.shape(1) then
+    halt("Mismatched shape in matrix-matrix multiplication");
 
   var C: [Adom.dim(1), Bdom.dim(2)] eltType;
   gemm(A, B, C, 1:eltType, 0:eltType);
@@ -440,6 +443,8 @@ private proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
 proc inner(A: [?Adom], B: [?Bdom]) {
   if Adom.rank != 1 || Bdom.rank != 1 then
     compilerError("Rank sizes are not 1");
+  if Adom.size != Bdom.size then
+    halt("Mismatched size in vector-vector multiplication");
 
   return + reduce(A[..]*B[..]);
 }
@@ -474,6 +479,7 @@ proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
   return C;
 }
 
+
 pragma "no doc"
 /* Generic matrix-matrix multiplication */
 proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
@@ -495,6 +501,7 @@ proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
 /* Return the matrix ``A`` to the ``bth`` power, where ``b`` is a positive
    integral type. */
 proc matPow(A: [], b) where isNumeric(b) {
+  // TODO -- flatten recursion into while-loop
   if !isIntegral(b) then
     // TODO -- support all reals with Sylvester's formula
     compilerError("matPow only support powers of integers");
@@ -524,6 +531,7 @@ proc matPow(a, B: [?Bdom] ?eltType) where isNumeric(a) {
 }
 
 
+// TODO -- matrix-exponential
 /* Return element-wise exponential: ``e**A[i,j]`` */
 proc matExp(A: [?Dom] ?eltType) {
   var B: [Dom] eltType = e**A;
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -531,22 +531,6 @@ proc matPow(a, B: [?Bdom] ?eltType) where isNumeric(a) {
 }
 
 
-// TODO -- matrix-exponential
-/* Return element-wise exponential: ``e**A[i,j]`` */
-proc matExp(A: [?Dom] ?eltType) {
-  var B: [Dom] eltType = e**A;
-  return B;
-}
-
-
-pragma "no doc"
-/* Integer overload */
-proc matExp(A: [?Dom] ?eltType) where isIntegralType(eltType) {
-  var B: [Dom] real = e**A;
-  return B;
-}
-
-
 /* Return cross-product of 3-element vectors ``A`` and ``B`` with domain of
   ``A`` */
 proc cross(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -28,10 +28,10 @@ A high-level interface to linear algebra operations and procedures.
 Compiling with Linear Algebra
 -----------------------------
 
-The linear algebra module uses both `BLAS` and `LAPACK` for
-many routines. In order to compile a Chapel program with this module, be sure
-to have BLAS and LAPACK implementations available on your system. See the BLAS
-and LAPACK documentation for further details.
+The linear algebra module uses the :mod:`BLAS` module.
+In order to compile a Chapel program with this module, be sure
+to have a BLAS implementation available on your system. See the :mod:`BLAS`
+documentation for further details.
 
 
 Linear Algebra API
@@ -42,7 +42,7 @@ Linear Algebra API
 Matrices and vectors are represented as Chapel arrays, which means the
 convenience constructors provided in this module are not necessary to use this
 module's functions. This also means that matrices (2D arrays) and vectors (1D
-arrays) will work with any other Chapel library works with arrays.
+arrays) will work with any other Chapel library that works with arrays.
 
 .. note::
 
@@ -56,7 +56,7 @@ unless otherwise specified.
 
 **Matrix multiplication**
 
-``dot()`` is the general matrix multiplication function provided in this module.
+:proc:`dot` is the general matrix multiplication function provided in this module.
 This function supports any combination of scalars, vectors (1D arrays), and
 matrices (2D arrays). See the :proc:`dot` documentation for more information.
 
@@ -65,9 +65,6 @@ underlying functions: :proc:`matMult` and `inner`.
 
 The :proc:`dot` function, along with others may be given a matrix-specific
 operator in future releases.
-
-.. comment:
-  It is also less verbose than matrixMultiply, matMult, or mMult.
 
 **Row vs Column vectors**
 
@@ -272,6 +269,16 @@ proc eye(Dom: domain(2), type eltType=real) {
 //
 
 pragma "no doc"
+inline proc transpose(D: domain(1)) {
+  return D;
+}
+
+inline proc transpose(D: domain(2)) {
+  return {D.dim(2), D.dim(1)};
+}
+
+
+pragma "no doc"
 /* Transpose vector (no-op) */
 inline proc transpose(A: [?Dom]) where Dom.rank == 1 {
   return A;
@@ -295,12 +302,12 @@ proc transpose(A: [?Dom] ?eltType) where Dom.rank == 2 {
   if !isBLASType(eltType) then
     return _transpose(A);
   else if Dom.shape(1) == 1 then
-    return reshape(A, {Dom.dim(2), Dom.dim(1)});
+    return reshape(A, transpose(Dom));
   else if Dom.shape(2) == 1 then
-    return reshape(A, {Dom.dim(2), Dom.dim(1)});
+    return reshape(A, transpose(Dom));
   else {
     var B: [Dom] eltType = eye(Dom, eltType=eltType);
-    var C: [{Dom.dim(2), Dom.dim(1)}] eltType;
+    var C: [transpose(Dom)] eltType;
     // TODO -- use native algorithm (BLAS not necessary or memory-efficient)
     gemm(A, B, C, 1:eltType, 0:eltType, opA=Op.T);
     return C;
@@ -315,7 +322,7 @@ proc _array.T where this.domain.rank == 2 { return transpose(this); }
 /* Transpose matrix with native implementation */
 pragma "no doc"
 private proc _transpose(A: [?Dom] ?eltType) {
-  var C: [{Dom.dim(2), Dom.dim(1)}] eltType;
+  var C: [transpose(Dom)] eltType;
 
   // naive algorithm
   forall (i, j) in Dom do
@@ -327,7 +334,8 @@ private proc _transpose(A: [?Dom] ?eltType) {
 
 /* Add matrices, maintaining dimensions */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
-  if Adom.size != Bdom.size then halt('Unmatched array size');
+  if Adom.rank != Bdom.rank then compilerError("Ranks to not match");
+  if Adom.size != Bdom.size then halt("Unmatched array size");
   var C: [Adom] eltType = A + B;
   return C;
 }
@@ -335,7 +343,8 @@ proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
 
 /* Subtract matrices, maintaining dimensions */
 proc matMinus(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
-  if Adom.size != Bdom.size then halt('Unmatched array size');
+  if Adom.rank != Bdom.rank then compilerError("Ranks to not match");
+  if Adom.size != Bdom.size then halt("Unmatched array size");
   var C: [Adom] eltType = A - B;
   return C;
 }
@@ -391,7 +400,7 @@ proc matMult(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
 
 pragma "no doc"
 /* matrix-vector multiplication */
-proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
+private proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
   where isBLASType(eltType)
 {
   if Adom.rank != 2 || Xdom.rank != 1 then
@@ -405,7 +414,7 @@ proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
 
 pragma "no doc"
 /* matrix-matrix multiplication */
-proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
+private proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType)
   where isBLASType(eltType)
 {
   if Adom.rank != 2 || Bdom.rank != 2 then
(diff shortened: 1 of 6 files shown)


modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+pragma "no doc"
+/* Transpose vector (no-op) */
+inline proc transpose(A: [?Dom]) where Dom.rank == 1 {
+  return A;
+}
+
+
+pragma "no doc"
+/* Syntactic sugar for transpose(Vector) */
+proc _array.T where this.domain.rank == 1 { return transpose(this); }
```
benharsh: I'm a little wary of this addition to the array interface. I'm not necessary opposed, but I don't think it's been discussed amongst the team yet.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+proc transpose(A: [?Dom] ?eltType) where Dom.rank == 2 {
+  if !isBLASType(eltType) then
+    return _transpose(A);
+  else if Dom.shape(1) == 1 then
+    return reshape(A, {Dom.dim(2), Dom.dim(1)});
+  else if Dom.shape(2) == 1 then
+    return reshape(A, {Dom.dim(2), Dom.dim(1)});
+  else {
+    var B: [Dom] eltType = eye(Dom, eltType=eltType);
+    var C: [{Dom.dim(2), Dom.dim(1)}] eltType;
```
benharsh: Eventually it may be worth adding a function or method that transposes a domain as well.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+  // naive algorithm
+  forall (i, j) in Dom do
+    C[j, i] = A[i, j];
+
+  return C;
+}
+
+
+/* Add matrices, maintaining dimensions */
+proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
```
benharsh: I think these (matPlus, matMinus, etc.) should have where-clauses or compiler asserts ensuring that the rank is two (or maybe one as well).

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+    // TODO -- assert shapes are correct
+    return _matmatMult(A, B);
+  else
+    compilerError("Rank sizes are not 1 or 2");
+}
+
+
+pragma "no doc"
+/* matrix-vector multiplication */
+proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
```
benharsh: Should these functions with leading underscores be private?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+/* Return element-wise power: ``a**B`` */
+proc matPow(a, B: [?Bdom] ?eltType) where isNumeric(a) {
+  var C: [Bdom] eltType = a**B;
+  return C;
+}
+
+
+/* Return element-wise exponential: ``e**A[i,j]`` */
+proc matExp(A: [?Dom] ?eltType) {
+  var B: [Dom] eltType = e**A;
```
benharsh: I might be missing it, but where is ``e`` declared?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+.. comment:
+  It is also less verbose than matrixMultiply, matMult, or mMult.
+
+**Row vs Column vectors**
+
+Row and column vectors are both represented as 1D arrays and are
+indistinguishable in Chapel. In the :proc:`dot` function, matrix-vector
+multiplication assumes a column vector, vector-matrix multiplcation assumes a
+row vector, vector-vector multiplication is always treated as an inner-product,
+as the function name implies.
```
ben-albrecht: I meant `dot` as in `dot product` would imply an `inner product` between 2 vectors.

benharsh:
> Indexing
> module creates new arrays with 0-based indexing unless a user specifies otherwise (with a domain of their own)
fine by me!

> Naming schemes
> e.g. casing on array initializer wrappers: Matrix() and Vector().
Also fine. We can reevaluate once we know more about our c++-like-vector story

> Are the matrix operations that purely exist to create temp arrays and override promotion flattening > valuable enough to include?
> matPlus, matMinus, ...
I think as long as we have this ambiguity with promotion flattening, these helpers are fine.

> Implementations
> Are we OK with naive implementation for some routines, or would we prefer to not support them yet > at all?
Is performance your concern? As long as the interface is about right, I'd prefer we support the functionality and deal with performance later.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+  var M: [{dim1, dim2}] eltType;
+
+  for i in dim1 do
+    M[i, ..] = Arrays(i+1)[..]: eltType;
+
+  return M;
+}
+
+
+/* Return an identity matrix over domain ``{0..#m, 0..#n}`` */
```
ben-albrecht: While the Identity matrix is defined as square, most linear algebra libraries support generating non-square identity matrices.

modules/packages/LinearAlgebra.chpl
ben-albrecht: That seems logical. That operation is used quite a bit throughout.

modules/packages/LinearAlgebra.chpl
ben-albrecht: yes

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+  This is an early prototype package module. As a result, interfaces may change
+  over the next release.
+
+Compiling with Linear Algebra
+-----------------------------
+
+The linear algebra module uses both `BLAS` and `LAPACK` for
+many routines. In order to compile a Chapel program with this module, be sure
+to have BLAS and LAPACK implementations available on your system. See the BLAS
+and LAPACK documentation for further details.
```
benharsh: A link to the BLAS and LAPACK module documentation could be useful here.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I considered that, but then realized that their implementation supports any rank size, so if a user wants to do that, I'd let them. Do you think that should be restricted?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+
+Linear Algebra API
+------------------
+
+**Matrix and Vector representation**
+
+Matrices and vectors are represented as Chapel arrays, which means the
+convenience constructors provided in this module are not necessary to use this
+module's functions. This also means that matrices (2D arrays) and vectors (1D
+arrays) will work with any other Chapel library works with arrays.
```
benharsh: grammar: "will work with any other Chapel library that works with arrays"

modules/packages/LinearAlgebra.chpl
ben-albrecht: Already added in my next commit!

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+  `vector` to refer to `1D arrays`.
+
+**Indexing**
+
+All functions that return new arrays will return arrays over 0-based indices,
+unless otherwise specified.
+
+**Matrix multiplication**
+
+``dot()`` is the general matrix multiplication function provided in this module.
```
benharsh: You link to the ``dot`` function below, should you do so here as well?

modules/packages/LinearAlgebra.chpl
ben-albrecht: It's [Math.e](http://chapel.cray.com/docs/master/modules/standard/Math.html#Math.e)

modules/packages/LinearAlgebra.chpl
ben-albrecht: Probably.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+matrices (2D arrays). See the :proc:`dot` documentation for more information.
+
+If you find this convenient notation confusing, you can opt to calling the
+underlying functions: :proc:`matMult` and `inner`.
+
+The :proc:`dot` function, along with others may be given a matrix-specific
+operator in future releases.
+
+.. comment:
+  It is also less verbose than matrixMultiply, matMult, or mMult.
```
ben-albrecht: I was on the border of uncommenting this to describe why `dot` was chosen as the general matrix multiplication function as opposed to other names... but I also don't want the documentation to sound defensive.

modules/packages/LinearAlgebra.chpl
benharsh: Should the ranks of ``A`` and ``B`` match?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+  if Adom.rank != 1 || Bdom.rank != 1 then
+    compilerError("Rank sizes are not 1");
+  if Adom.size != 3 || Bdom.size != 3 then
+    halt("cross() expects arrays of 3 elements");
+
+  var C = Vector(Adom.size, eltType);
+
+  C[0] = A[1]*B[2] - A[2]*B[1];
+  C[1] = A[2]*B[0] - A[0]*B[2];
+  C[2] = A[0]*B[1] - A[1]*B[0];
```
benharsh: I don't think we should assume ``A`` and ``B`` are 0-based or unstrided. You may want to use ``orderToIndex`` or something here.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I've discussed this with @bradcray && @mppf . However, since dropping LAPACK support, transpose is the only Linear Algebra specific array method. If it's considered controversial, we could drop it for now (as it's just a syntactic sugar anyway), and have more thorough discussion before 1.16.

Anyway, I'll add this to the list of interesting topics in the PR description.

benharsh: If those they are fine with it, so am I.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,709 @@
...
+
+pragma "no doc"
+proc Matrix(Arrays...?n) {
+  type eltType = Arrays(1).eltType;
+  return Matrix((...Arrays), eltType=eltType);
+}
+
+
+/*
+    Return a matrix (2D array), given 2 or more vectors. Vectors are
```
ben-albrecht: That's actually fine. While the library represents both row and column vectors as 1D arrays, a user can still differentiate the two by using an `Nx1` or `1xN `matrix to explicitly represent row and column matrices, respectively.

A user might apply this to the `dot` function, to get an outer product instead of the default behavior of treating vector-vector multiplication as an inner product:

```chapel

var v = Matrix(1, 3);  // 1x3 "matrix" (column vector)
var vT = transpose(v); // 3x1 "matrix" (row vector)

var outerProduct = dot(vT, v);
writeln(outerProduct);
/* 
   0.0 0.0 0.0
   0.0 0.0 0.0
   0.0 0.0 0.0 
*/
```

modules/packages/LinearAlgebra.chpl
npadmana: Isn't the requirement here stronger --- that the number of indices in each dimension are the same? If I understand the logic below, you are ensuring the rank is the same, and the total number of indices is the same -- but that is not quite the same.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+}
+
+
+pragma "no doc"
+/* Element-wise scalar multiplication */
+proc dot(a, B: []) where isNumeric(a) {
+  return dot(B, a);
+}
+
+
```
npadmana: I found the switch between ``dot`` and ``matMult`` somewhat confusing. In usual usage, ``dot`` does mean matrix multiplication, so why not use it throughout?

Also, FWIW, both Mathematica and numpy use ``dot`` for matrix multiplication....

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+  if Adom.rank != 2 || Bdom.rank != 2 then
+    compilerError("Rank sizes are not 2");
+
+  var C: [Adom.dim(1), Bdom.dim(2)] eltType;
+  gemm(A, B, C, 1:eltType, 0:eltType);
+  return C;
+}
+
+
+/* Inner product of 2 vectors */
```
npadmana: Going back to my previous comment on ``dot``, I think inner could just be ``dot``. Is there a reason to pull it out separately?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+
+
+/* Return element-wise power: ``a**B`` */
+proc matPow(a, B: [?Bdom] ?eltType) where isNumeric(a) {
+  var C: [Bdom] eltType = a**B;
+  return C;
+}
+
+
+/* Return element-wise exponential: ``e**A[i,j]`` */
```
npadmana: There appears to be a mild inconsistency here -- matPow is (I'd argue correctly) the matrix power (and not the elementwise power), which matExp is elementwise, and not the matrix exponential...

npadmana:
@ben-albrecht, @benharsh ---  a different perspective on the naive implementations. I think the issue here is less one of speed, but of numerical stability; some naive algorithms are notoriously unstable (for instance, I don't actually know how the det algorithm as written performs on moderate sized matrices -- eg 100x100).

Given that we already have an interface to BLAS/LAPACK (so a user can directly call those if necessary), I might be tempted to be conservative and leave things unimplemented until one has the appropriate implementation in.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+  V[0] = x: eltType;
+
+  for i in 1..n do V[i] = Scalars[i]: eltType;
+
+  return V;
+}
+
+
+/* Return a matrix (2D array) over domain ``{0..#rows, 0..#cols}``*/
+proc Matrix(rows, cols, type eltType=real) where isIntegral(rows) && isIntegral(cols) {
```
npadmana: Should this have a check to make sure rows, cols >0?

npadmana: Also, I wonder if there should be a form of Matrix that just takes rows and returns a square matrix? That might be a trivial change to the above....

modules/packages/LinearAlgebra.chpl
npadmana: FWIW, here are some thoughts on the a general transpose for arrays (which I would love to see happen!). 

1.  I think it's more general that just matrices; one can imagine an arbitrary permutation of indices. Going beyond just matrices is useful for eg. in FFTs. It could also be a way of shuffling data for better locality....
2. You can imagine a view here, where the data aren't really transposed. 
3. You could do an in-place transpose. Doing this right always struck me as somewhat magical, and so I'd love for it to be baked into the language.
4. You could produce a new array.... 

And of course, now extend this to distributed arrays!

All of these strike me as tricky enough to do **right**, that it would be great if these were built into the language. But they also go beyond just the linear algebra transpose.....

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+  if !isBLASType(eltType) then
+    return _transpose(A);
+  else if Dom.shape(1) == 1 then
+    return reshape(A, transpose(Dom));
+  else if Dom.shape(2) == 1 then
+    return reshape(A, transpose(Dom));
+  else {
+    var B: [Dom] eltType = eye(Dom, eltType=eltType);
+    var C: [transpose(Dom)] eltType;
+    // TODO -- use native algorithm (BLAS not necessary or memory-efficient)
```
npadmana: I agree, this is complete overkill..... any reason not to just swap this now?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+    return C;
+  }
+}
+
+
+/* Transpose vector or matrix */
+proc _array.T where this.domain.rank == 2 { return transpose(this); }
+
+
+/* Transpose matrix with native implementation */
```
npadmana: Really a TODO for a later time, but I think if you do some amount of tiling, you can improve cache performance here.  Why this isn't a part of BLAS is a mystery to me; I believe MKL has a routine that does do this (so you don't have to reinvent the wheel). Of course, if Chapel implements an array-level transpose as I suggested above, this would be moot.

npadmana:
@ben-albrecht -- some suggestions for other _matrix structure_ routines (just to get them down on paper)...

- extract matrix diagonal (or off-diagonal) and store it to a vector.
- convert vector into diagonal matrix
- might be nice to have iterators for  (off diagonal) and upper/lower triangular matrices. I could imagine something that just returned indices, and something that iterated over the elements....  (maybe you don't need the latter).

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+  if !isSquare(A) then
+    halt("Array not square");
+
+  return _expBySquaring(A, b);
+}
+
+
+pragma "no doc"
+/* Exponentiate by squaring recursively */
+private proc _expBySquaring(x: ?t, n): t {
```
npadmana: A question --- while the recursion here makes the algorithm clear, I'm not sure what the memory usage is --- does it end up keeping copies of all the intermediate steps (especially when n is odd, where this doesn't look like it's tail-recursive).

I think you could do this with a while loop and only two matrices -- let me define the triplet (result, tmp, n).

- Initialize (R,T,i) = (1, A, n)
- If n even, (R,T,i) -> (R, T.T, n/2)
- If n odd (R,T,i) -> (R.T, T.T, (n-1)/2)
- repeat until n=1, at which point, return R*T

I think I got this right.... trying with --
* A^7 -- (1,A,7) -> (A, A^2, 3) -> (A^3, A^4, 1) -> A^7 
* A^4 -- (1, A, 4) -> (1, A^2, 2) -> (1, A^4, 1) -> A^4
* A^6 -- (1, A, 6) -> (1, A^2, 3) -> (A^2, A^4, 1) -> A^6

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+    // TODO -- _expBySquaring(pinv(x), -n);
+    if n < 0  then halt("Negative powers not yet supported");
+    else if n == 0  then return eye(x.domain);
+    else if n == 1  then return x;
+    else if n%2 == 0  then return _expBySquaring(dot(x, x), n / 2);
+    else return dot(x, _expBySquaring(dot(x, x), (n - 1) / 2));
+}
+
+
+/* Return element-wise power: ``a**B`` */
```
npadmana: I actually think my comment on matExp actually applies here as well. I guess you are including these because of this flattening problem, but we probably should be careful about mixing elementwise and matrix operations (not a problem for matPlus and matMinus, since they're the same).

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+      a[K,L] = A[k,l];
+    }
+  }
+  return a;
+}
+
+
+/* Return the trace of ``A`` */
+proc trace(A: [?D] ?eltType) {
+  if D.rank == 1 then return + reduce A[..];
```
npadmana: Hm.... the rank-1 behavior isn't consistent with the rank-2 behavior. 

- I think the rank-2 case only works for a square matrix (I'm fine with that), but it's probably worth an explicit check, as opposed to letting ``zip`` throw an error.
- In that case, the rank-1 case should only be allowed for a single element vector....

I'd be fine dropping the rank 1 case altogether.

modules/packages/LinearAlgebra.chpl
ben-albrecht: Hi @npadmana -- I'm happy that you commented on this:

To clarify, I *intend* for `dot` to be the general matrix-matrix multiplication routine, which  wraps `inner` and `matMult`. We leave `inner` available as part of the public interface in case users want to explicitly distinguish `inner` vs `outer` products of vectors (since row vs. column vector is indistinguishable.)

I also left `matMult` as part of the public interface trying to mirror what I did with `inner`, but it's not really useful aside from being an alternative syntax. As you point out, it's a potential source of confusion, so I'm inclined to make it private for this first version of the library.

modules/packages/LinearAlgebra.chpl
ben-albrecht: It is redundant with `dot`, but I believe that argument here (following numpy's interface), is that it can be useful to explcitly distinguish `inner` from `outer` products. We could also later extend `inner` to support higher-dimensional arrays, like numpy.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I'll try to get the more efficient implementation in for code/doc freeze.

modules/packages/LinearAlgebra.chpl
ben-albrecht: done

modules/packages/LinearAlgebra.chpl
ben-albrecht: done

ben-albrecht: added!

modules/packages/LinearAlgebra.chpl
ben-albrecht: Hm, good point. I think we ought to either have the name scheme reflect element-wise operations, and apply them to all of the element-wise ops, like `matPlus` and `matMinus`, or just drop the element-wise support for `matExp` and `matPow` to improve interface clarity. I'm leaning towards the latter, given that promotion flattening is a temporary challenge in the language.

modules/packages/LinearAlgebra.chpl
```diff
@@ -717,17 +727,15 @@ private proc _minor(A: [?Adom] ?eltType, j) {
 }
 
 
-/* Return the trace of ``A`` */
+/* Return the trace of matrix square matrix ``A`` */
```
npadmana: Do you have a stray "matrix" here?

modules/packages/LinearAlgebra.chpl
npadmana: I agree. Ideally, we should not have any elementwise operations defined as functions, since Chapel does this correctly. I hope this "flattening" feature goes away soon... (although it will cause code to generally break spectacularly, instead of silently).

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,733 @@
...
+proc dot(a, B: []) where isNumeric(a) {
+  return dot(B, a);
+}
+
+
+pragma "no doc"
+/* Explicit matrix-(matrix|vector) multiplication */
+proc matMult(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
+  // matrix-vector
+  if Adom.rank == 2 && Bdom.rank == 1 then
```
ben-albrecht: You're right. I've added the `gemv` arg, `opA=BLAS.Op.T`, for the `M.v(col)` case in my next commit.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I should have commented that this was a known naive implementation, which implies a TODO to come back for an improved implementation.

I'll plan to implement the non-recursive approach for code/doc freeze.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I've dropped the `matExp` routine for now, until it is implemented correctly as [matrix exponential](https://en.wikipedia.org/wiki/Matrix_exponential#Computing_the_matrix_exponential)

ben-albrecht:
I ended up dropping `proc det` for reasons @npadmana pointed out. I think it makes a lot more sense to hold off on implementing this, until LAPACK support is available.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+
+  for i in dim1 do
+    M[i, ..] = Arrays(i+1)[..]: eltType;
+
+  return M;
+}
+
+
+/* Return an identity matrix over domain ``{0..#m, 0..#n}`` */
+proc eye(m, n, type eltType=real) {
```
npadmana: Do we like the numpy ``eye`` notation? An alternative would be ``identity``, which is more characters, but has less of a feel of an inside joke. :-)

Also, ``n`` could be optional here, defaulting to a square matrix.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+proc matMinus(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
+  if Adom.rank != Bdom.rank then compilerError("Unmatched ranks");
+  if Adom.shape != Bdom.shape then halt("Unmatched shapes");
+  var C: [Adom] eltType = A - B;
+  return C;
+}
+
+/*
+    Generic matrix multiplication, ``A`` and ``B`` can be a matrix, vector, or
+    scalar
```
npadmana: I suggest adding in a note here that if the user calls this with a v.M, it actually computes M^t.v, which may not be as efficient as if the user could manually swap orders.

In particular, doing a.M.b might be better implemented as dot(dot(M,b),a). This gain is only true for vectors where rows and columns are swappable.

Now that I think about it, that would be a cool helper function, since that's a common operation, and I've always found the nested dots ugly!

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+  return C;
+}
+
+
+/* Inner product of 2 vectors */
+proc inner(A: [?Adom], B: [?Bdom]) {
+  if Adom.rank != 1 || Bdom.rank != 1 then
+    compilerError("Rank sizes are not 1");
+  if Adom.size != Bdom.size then
+    halt("Mismatched size in vector-vector multiplication");
```
npadmana: vector-vector multiplication ---> inner product?

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+  var C: [{Adom.dim(1), Bdom.dim(1)}] eltType;
+  forall (i,j) in C.domain do
+    C[i, j] = A[i]*B[j];
+  return C;
+}
+
+
+pragma "no doc"
+/* Generic matrix-vector multiplication */
+proc _matvecMult(A: [?Adom] ?eltType, X: [?Xdom] eltType)
```
npadmana: I think this needs a trans option to be consistent with the BLAS matrix-vector piece.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+    if n < 0  then halt("Negative powers not yet supported");
+    else if n == 0  then return eye(x.domain);
+    else if n == 1  then return x;
+    else if n%2 == 0  then return _expBySquaring(dot(x, x), n / 2);
+    else return dot(x, _expBySquaring(dot(x, x), (n - 1) / 2));
+}
+
+
+/* Return element-wise power: ``a**B`` */
+proc matPow(a, B: [?Bdom] ?eltType) where isNumeric(a) {
```
npadmana: Just not to lose this -- I think this scalar form needs to be removed, as we did for matExp...

modules/packages/LinearAlgebra.chpl
cassella: In point 1, do you mean "more general than just transposes" and "Going beyond just transposes"?

npadmana: @cassella -- I wasn't clear here. I just meant that the idea of transposes would go beyond the rank-2 case that we're considering here, but could be generalized to a rank-n rectangular domain/array. And that you can consider an consider an arbitrary permutation of indices, not just a two-index swap. And that this would be nice to have built in.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+private inline proc _raw(D: domain(1), i) {
+  return D.dim(1).orderToIndex(i);
+}
+
+
+//
+// Matrix Structure
+//
+
+/* Return lower triangular part of matrix, below the diagonal + ``k`` */
```
npadmana: I'd suggest being more explicit about the documentation here and in the next routine. In particular, I'm perpetually confused in functions like this about whether ``k=0`` includes the diagonal or not.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+  }
+  return true;
+}
+
+
+/* Return `true` if matrix is Hermitian */
+proc isHermitian(A: [?D]) {
+  if D.rank != 2 then
+    compilerError("Rank size is not 2");
+  for (i, j) in D {
```
npadmana: Just a comment here --- code like this is why it would be nice to have an iterator for upper/lower triangular.

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+    if i > j {
+      if A[i, j] != A[j, i] then return false;
+    }
+  }
+  return true;
+}
+
+
+/* Return `true` if matrix is lower triangular below the diagonal + ``k`` */
```
npadmana: I would just cut and paste the additional documentation from ``tril`` and ``triu`` here (explaining ``k`` explicitly), or link these to those routines...

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+  if D.rank != 2 then
+    compilerError("Rank size is not 2");
+  const (M, N) = A.shape;
+  return M == N;
+}
+
+
+pragma "no doc"
+/* Helper function for determinant */
+private proc _minor(A: [?Adom] ?eltType, j) {
```
npadmana: Given that ``det`` is removed, should we remove this?

test/modules/packages/linearalgebra/correctness/correctness.chpl
```diff
@@ -0,0 +1,435 @@
...
+    var I = eye(M.domain, eltType=t);
+
+    var v = Vector(1,2,3, eltType=t);
+
+    // matrix-matrix
+    var MI = dot(M, I);
+    assertEqual(MI, M, "dot(M, I)");
+
+    // vector-matrix && matrix-vector
+    var Iv = dot(I, v);
```
npadmana: I think we should do this test with M and not I. The identity is a little too special when it comes to linear algebra (eg. it would not have caught the bug in the gemv transpose).

test/modules/packages/linearalgebra/correctness/correctness.chpl
```diff
@@ -0,0 +1,435 @@
...
+  var M = Matrix(3, 3);
+  var S = Matrix(1, 1);
+
+  v31 = 1;
+  v13 = 1;
+  M = 1;
+
+  S[0, 0] =  inner(v13[0, ..], v31[.., 0]);
+
+  // outer-product
```
npadmana: Not sure I understand this comment....

modules/packages/LinearAlgebra.chpl
```diff
@@ -0,0 +1,701 @@
...
+pragma "no doc"
+proc Matrix(Arrays...?n) {
+  type eltType = Arrays(1).eltType;
+  return Matrix((...Arrays), eltType=eltType);
+}
+
+
+/*
+    Return a matrix (2D array), given 2 or more vectors. Vectors are
+    concatenated such that the ``ith`` vector corresponds to the matrix slice:
```
npadmana: A different way of saying this is that vectors form the rows of the matrix.... it might be good to have both statements. 

At a later point, an example here would possibly be helpful.

test/modules/packages/linearalgebra/correctness/correctness.chpl
```diff
@@ -0,0 +1,435 @@
...
+  var M = Matrix(3, 3);
+
+  v = 1;
+  M = 1;
+
+  // outer-product
+  assertEqual(outer(v, v), M, "outer(Vector(3), Vector(3))");
+}
+
+/* matPow scalar-matrix */
```
npadmana: I think this test goes the way of matExp, the dinosaur and the dodo.

modules/packages/LinearAlgebra.chpl
ben-albrecht: This library was primarily based on the linear algebra support in numpy and matlab. I'm open to alternatives to `eye`, but let's leave it as is for now, and discuss it further before code/doc.

test/modules/packages/linearalgebra/correctness/correctness.chpl
ben-albrecht: I think it was left-over from the TODO-block... removed!

PR #5718, opened 2017-03-17
delete errors, nested try

psahabu:
This PR inserts `chpl_delete_error` to automatically delete `Error`s as they are consumed. This strategy was chosen because `PRIM_DELETE` is lowered before the `lowerErrorHandling` pass, so it was more straightforward to create a routine to delete `Error`, include that routine in `sWellKnownFns`, then call it from the error handling lowering pass. `chpl_delete_error` also ensures that the `Error` module is properly resolved and not hoisted, allowing us to remove the `ensureErrorTypeRemains` dummy.

This PR also allows `try` statements to propagate to an enclosing `try` statement.

flat+linux64 testing is clean.

### Example: inserted deletes
```
// given code
proc propagate() throws {
  try {
    a(); // throws
    b(); // does not throw
    c(); // throws
  } catch e: SubError {
    f();
  } catch e: AnotherSubError {
    g();
  }
}

// after this pass
proc propagate(out error_out: Error) {
  var error: Error;
  a(error);
  if error then
    goto handler;
  b();
  c(error);
  if error then
    goto handler;

  label handler:
  if error {
    var e = error: SubError;
    if _cast {
      f();
      delete e;
    } else {
      var e = error: AnotherSubError;
      if e {
        g();
        delete e;
      } else {
        // set and return
        error_out = error;
        goto epilogue_label;
      }
    }
  }
}
```

### Example: nested try
```
try {
  try {
    throwingCall();
  } catch e: SpecificError {
    handleGracefully();
  }
  otherThrowingCall();
} catch {
  handleSomehow();
}

{
  var _e1: Error;
  {
    var _e2: Error;
    throwingCall(_e2);
    if _e2 then
      goto handler2;

    label handler2:
    if _e2 {
      var _cast = _e2: SpecificError;
      if _cast {
        handleGracefully();
        delete _cast;
      } else {
        _e1 = _e2;
        goto handler1;
      }
    }
  }
  otherThrowingCall(_e1);
  if _e1 then
    goto handler1;

  label handler1:
  if _e1 {
    handleSomehow();
    delete _e1;
  }
}
```

### In list form
- creates and uses `gChplDeleteError` for `chpl_delete_error`
- inserts moves between `error` variables of nested `try` statements
- refactoring to support these changes

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -93,7 +93,7 @@ try {
   handleSomehow();
 }
 
-{
+{FFFFF
   var _e1: Error;
   {
     var _e2: Error;
@@ -147,11 +147,12 @@ private:
   LabelSymbol*        epilogue;
   bool                insideCatch;
 
-  AList     tryHandler         (TryStmt*   tryStmt,  VarSymbol* errorVar, TryInfo* outerTry);
-  AList     setOutGotoEpilogue (VarSymbol* error);
-  AList     errorCond          (VarSymbol* errorVar, BlockStmt* thenBlock,
-                                BlockStmt* elseBlock = NULL);
-  CallExpr* haltExpr           ();
+  AList     handler           (TryStmt*   tryStmt,  VarSymbol* errorVar,
+                               TryInfo* outerTry);
+  AList     setOutGotoEpilogue(VarSymbol* error);
+  AList     errorCond         (VarSymbol* errorVar, BlockStmt* thenBlock,
+                               BlockStmt* elseBlock = NULL);
+  CallExpr* haltExpr          ();
 
   ErrorHandlingVisitor();
 };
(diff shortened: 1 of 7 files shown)


mppf:
Could you say more in the PR message about chpl_delete_error ? In particular, you might say something like "Because PRIM_DELETE is lowered by this point in compilation, it was most convenient to create a routine to delete Error, to include that routine in sWellKnownFns, and then to call that from the error handling lowering pass".

compiler/passes/errorHandling.cpp
```diff
@@ -136,77 +177,85 @@ bool ErrorHandlingVisitor::enterTryStmt(TryStmt* node) {
...
+  tryBlock->insertAtTail(handler(node, info.errorVar, outerTry));
 
   tryBlock->remove();
   node    ->replace(tryBlock);
-  tryStack.pop();
 }
 
-AList ErrorHandlingVisitor::tryHandler(TryStmt* tryStmt, VarSymbol* errorVar) {
+AList ErrorHandlingVisitor::handler(TryStmt* tryStmt, VarSymbol* errorVar,
+                                       TryInfo* outerTry) {
```
mppf: A nit: 'tryHandler' and 'handler' are both maybe not as good names as something like "lowerCatches" or "handleCatches" or "handleCatchesAndPropagation" etc. Change it if you like, or not...

mppf:
While this can go in now, it's a good idea to either be available tomorrow to fix regressions or wait to merge until Monday.

psahabu:
I'm going to be checking in tomorrow to make sure I fixed last night's regressions, so I'll merge this in and deal with the potential fallout tomorrow.

PR #5719, opened 2017-03-17
Avoid localizing the heap for CHPL_LOCALE_MODEL=knl

ronawho:
Only localize the heap for CHPL_LOCALE_MODEL=numa. #5689 added support to
NUMA-localize the heap under comm=ugni and loc-mod!=flat. Unfortunately this
doesn't seem to be working under knl right now, so disable it until Greg has a
chance to look into it.

Current failure is pretty generic:
  "internal error: hwloc_set_cpubind(): Invalid argument"


ronawho:
@gbtitus and @dmk42 i'm going to merge this to resolve the knl issues, but we should chat about this on monday

PR #5721, opened 2017-03-18
enable parentheses for catch

psahabu:
Enables `catch` filter expressions to be enclosed by parentheses.

### example
```
try! {
  throwOtherError();
} catch (e: OtherError) {
  writeln("specified catch");
}
```


test/errhandling/psahabu/catch-parens.chpl
```diff
@@ -0,0 +1,19 @@
+use ThrowError;
+
+try {
+  throwAnError();
+} catch () {
```
mppf: This case seems pretty weird to me. I'm not sure why anybody, even a C/C++ programmer, would write that...

mppf: That said, I don't have a problem with it being legal for now. But I'd probably rule it out in the parser. To do that, you could split catch_expr into catch_expr_nonempty and catch_expr_empty and just have the parenthesis version use catch_expr_nonempty. Up to you.

mppf:
Looks good other than the catch () { } weirdness, which I'm happy with you disallowing or leaving for future work. Make sure to run full local testing on it.

test/errhandling/psahabu/catch-parens.chpl
psahabu: Yeah this is weird, I'll disallow it. Splitting up empty and non-empty is also appealing because `newTemp(dtVoid)` doesn't survive until `errorHandling` due to the void type feature, so I can just create a separate case for that.

PR #5723, opened 2017-03-18
Make the multi-ddata specific fields in DR arrays 'void' when not using them

daviditen:
The changes in DefaultRectangular are mostly copied from a branch Elliot made.
It makes the md* fields 'void' when not doing multi-ddata, and their normal
types otherwise.  Then they can be removed and not cause overhead.

While testing this, I realized I made a mistake in cleaning up my 'void'
fields branch and made an =(void, void) operator non-param.  I was depending
on it being param so it gets removed after it is resolved.  Otherwise,
"void = void" assignments were getting left behind and causing problems.

Passed default linux64 paratest
Passed --no-local linux64 paratest
Passed --verify linux64 paratest

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -111,7 +111,8 @@ module ChapelBase {
     compilerError("cannot assign void to a variable of non-void type");
   }
 
-  inline proc =(ref a: void, b: void) { }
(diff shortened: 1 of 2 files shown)


ronawho:
Eventually we'll want to make the single-ddata fields void under mulit-ddata, but this is probably a good enough start.

daviditen:
Thanks @ronawho .

PR #5725, opened 2017-03-18
Make Statement Level Symbols into noops

bradcray:
In looking at generated code for standalone variable references like:

````chapel
A
````

I was surprised at the amount of code we generated, particularly for complex types like arrays.  It seems that the reason is due to the use of the _statementLevelSymbol() routine which returns the argument that was passed in rather than just dropping it on the floor.  Here, I change the behavior for value arguments to drop it on the floor.  I had to preserve the type and param versions as they were because of how they are apparently used in where clauses...

In looking at these cases, I found myself thinking that it seems a little silly that we're introducing these statementLevelSymbols rather than either

* switching to having the compiler special-case standalone sync reads/writes (as we already do in other non-statement contexts that reference such variables), or
* adding the ability to have a 0-argument 'this' accessor for types that would permit the sync class to say what its default read/write behavior is (and would give a similar capability to user types).

But that's going to have to be future work.

### CODE REVIEW

**Comment by bradcray** [comment]
@daviditen : Would you check out the first commit on this PR and try testing against test/release/examples/benchmarks/fft.chpl?  I was getting C compile-time issues due to the generation of '_void' symbols.  Apart from wondering why we were generating _void, I also wondered whether defining a global identifier named '_void' in the runtime headers that would just have the value of 0 (or something) would cause such errors to become innocuous.

**Comment by bradcray** [comment]
This got two failures on a test run:

[Error matching compiler output for functions/diten/whereClauseIgnored]
[Error matching compiler output for trivial/deitz/type-param-mod-config/type_expr_stmt]

The second has an easy and obvious fix.  The first needs a little more investigation.

**Comment by daviditen** [comment]
The ``_void`` is coming from a ``boundedLow`` range iterator that does:

```
    var i: idxType;
    const start = this.first;
    while __primitive("C for loop",
                      __primitive( "=", i, start),
                      true,
                      __primitive("+=", i, stride: idxType)) {
      yield i;
    }
```

The stopping condition above is ``true`` so it will run forever.  ``true`` is getting passed to ``_statementLevelSymbol()`` which is empty in your first commit (currently meaning returns ``_void``).  So that's where the _void is coming from, and I guess this is a case that I didn't catch to remove.

Unless the way we handle the code above is changed, I don't think you can just return ``true`` in ``_statementLevelSymbol()`` because if ``false`` was the stopping condition above it would turn into ``true`` after calling ``_statementLevelSymbol()``.  It appears we currently require it to return its argument.

**Comment by bradcray** [comment]
...except that we seem to insert statementLevelSymbols into where clauses... grumble...

**Comment by bradcray** [comment]
David: Thanks for the analysis.  I'm surprised that we insert statementLevelSymbols in that case -- i.e., I don't think we insert them for arbitrary variable references.  But I'm feeling less and less sure about where we do insert them.

**Comment by daviditen** [comment]
The "C for loop" translates into a BlockStmt for each argument, so it ends up with a block with a statement level symbol:

```
{
  true;
}
```

bradcray:
I see... thanks.  And then it presumably goes in and tries to do something with that later?

Mostly, this exploration is making me think that the whole statementLevelSymbol() concept should be removed from the compiler and that we should either:

* switch to always using the compiler to special-case sync reads/writes (as we already do in other non-statement contexts), or:
* add the ability to have a 0-argument 'this' accessor for types that would permit the sync class to say what its default read/write behavior is (and would give a similar capability to user types).

In any case, this it's becoming clear that this is a post-release effort, and not the simple cleanup I had hoped for.  Thanks for looking into the _void question quickly!

bradcray:
Filed issue #5725 to capture this thought process.

bradcray:
This latest commit passes testing... but by luck or skill?

bradcray:
This latest version passes testing as well.  The failure noted in the previous commit message was a problem on master, fixed by #5731 .

bradcray:
@vasslitvinov : Would you have time to, and be willing to, review this PR?  It's a minor cleanup on some code that I believe you've been involved with in the past.

vasslitvinov:
Assuming it works, this is great - good to merge.

bradcray:
I've only tested on linux64 -- does that seem sufficient to you?

vasslitvinov:
Yes, linux64 testing is enough to merge.

bradcray:
Thanks!

PR #5736, opened 2017-03-20
Improve const inference for chpl__iterLF

benharsh:
Improve the 'canRHSBeConstRef' function in inferConstRefs.cpp by handling the get-member-value primitives in the case that the LHS is neither a class nor a ref. This means that the 'use' can remain const.

"But why not handle classes?" you may ask. Consider the following pseudo-AST:
```
(move LHS (get-member-value base field))
(set-member LHS x 1)
```

If ``LHS`` is a record, I feel comfortable saying that ``base`` can be const because ``LHS`` will be copied in some way (e.g., bit copied). If ``LHS`` is a class, then we start getting into queries about aliasing and how "deep" const implies in this context (which we haven't really finalized as far as I know). So, for now, I believe the safest approach is to throw up our hands when we see a class and say 'not const'.

This reduces comm counts for stream-promoted.

Testing:
- [x] full local
- [x] full no-local
- [x] full gasnet
- [x] dist-cyclic


benharsh:
@ronawho : This should resolve the [16-node-xc regression](http://chapel.sourceforge.net/perf/16-node-xc/?startdate=2017/02/09&enddate=2017/03/19&configs=gnugasnetmpi&graphs=hpccpromotedstreamperfgbsn5723827200) for promoted STREAM.

mppf:
@benharsh - can you explain why this needs to check for LHS not-a-class? I understand why the new code should rule out the `ref` case but I think I'm missing something about why it should rule out classes.

PR #5738, opened 2017-03-20
Mark array view chpl__unref with FLAG_UNREF_FN

mppf:
callDestructors includes special logic to omit copies if it finds a call
to a function marked with FLAG_UNREF_FN.  This should also apply to the
copy generated in chpl__unref for returning array view slices (which
copies them into non-array-view arrays).

Resolves an error in the generated C code with this pattern:

``` chapel
  proc returnAliasSliceGlobal() {
    ref B = A[1..5];
    return B;
  }
```

which came up in PR #5576.

While there, adds a test of this code & a future of something related that I noticed is not working.

Passed quickstart testing.
Passed full local testing.

Passed hellos and test/arrays/ferguson/global-returns-ref.chpl  with:
 * --baseline
 *  GASNet
 * --verify
 * --baseline --verify


benharsh:
Looks reasonable to me 👍

bradcray:
@benharsh: Any guess as to whether this is going to cause problems for the chpl__unref-related mods on the rank-change branch?  (that code remains a complete black box to me...)

benharsh:
I wouldn't think so, but I haven't verified.

Michael was thinking that we could use this flag instead of the 'where == NULL' check currently on the branch.

bradcray:
I'd be a fan of that... I'm not sure whether you saw my question along those lines over the weekend, but I'd already been wishing for something nicer there.  (I'm writing this having not yet waded through Michael's comments on the rank-change PR).

benharsh:
I just tried a combination of this patch and replacing the 'where is null' check with a flag check, and it appears to work (at least for the sungeun/memory/refCount tests).

bradcray:
Cool, thanks @benharsh -- those definitely are the ones that are most likely to fail if something went wrong (I didn't tell you that I broke, and then re-fixed, those again sometime late yesterday... I'm still not entirely sure what I did to break them though I suspect it was due to the fragility of the test).  Can you put together a PR for that to be merged to the rank-change branch if/when this one is?

test/arrays/ferguson/tuples/return-tuple-slices.future
```diff
@@ -0,0 +1,4 @@
+bug: returning tuple of slices results in compilation error
+
+It should return a tuple of copies of the slices.
```
bradcray: @mppf: Is this a pre-existing condition or a new failure mode added by this PR?

mppf: Pre-existing condition. The problem is that computeNonRefTuple isn't taking in to account that autoCopy for slices returns a different type. I was just thinking about the related code and figured we probably had an error here...

PR #5740, opened 2017-03-20
Only include objects with CPUs in the NUMA node count.

gbtitus:
The topology layer's NUMA node count was simply the number of objects at
the topology's NUMA level.  This isn't right when some of the objects
there don't actually have CPUs, as for example the object for Intel Xeon
Phi (AKA KNL)  onboard memory (AKA HBM, MCDRAM, etc.).  From now on,
only include in the count objects that have CPUs associated with them.

Having done this, re-enable registered heap localization for KNL.

While here, I also fixed a broken debug output request in the topology
layer.

### DIFF
--- a/runtime/src/chpl-topo.c
+++ b/runtime/src/chpl-topo.c
@@ -148,12 +148,30 @@ void chpl_topo_init(void) {
   }
 
   //
-  // Note: This will find not only traditional CPU+memory NUMA nodes
-  //       but also memory-only nodes, so for example on KNL it may
-  //       not match chpl_task_getNumSublocales().  We will need to
-  //       revisit this in the future.
+  // Find the NUMA nodes, that is, the objects at numaLevel that also
+  // have CPUs.  This is as opposed to things like Xeon Phi HBM, which
+  // is memory-only, no CPUs.
   //
-  numNumaDomains = hwloc_get_nbobjs_by_depth(topology, numaLevel);
+  {
+    hwloc_cpuset_t cpuset;
+    hwloc_obj_t obj;
+
+    if ((cpuset = hwloc_bitmap_alloc()) == NULL) {
+      report_error("hwloc_bitmap_alloc()", errno);
+    }
+
+    numNumaDomains = 0;
+    for (obj = hwloc_get_next_obj_by_depth(topology, numaLevel, NULL);
+         obj != NULL;
+         obj = hwloc_get_next_obj_by_depth(topology, numaLevel, obj)) {
+      hwloc_bitmap_and(cpuset, obj->online_cpuset, obj->allowed_cpuset);
+      if (!hwloc_bitmap_iszero(cpuset)) {
+        numNumaDomains++;
+      }
+    }
+
+    hwloc_bitmap_free(cpuset);
+  }
 }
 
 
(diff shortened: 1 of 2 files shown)


gbtitus:
@ronawho and @dmk42, have a look at this at your convenience.  It's a possible solution to the problem where we were getting internal errors during locModel=knl testing, when trying to do NUMA localization.

ronawho:
Is there any benefit to localizing on knl if we're not using multi-ddata?

ronawho:
Code changes look good to me, though I wonder if there's any performance benefit to localizing under CHPL_LOCALE_MODEL=numa if we're not using multi-data there.

dmk42:
Yes, there is a benefit to localizing on KNL.  It isn't huge but it's noticeable.  For quad, there is only one numa node with cores, but for snc2 and snc4, things are definitely faster if they can be done locally.

runtime/src/chpl-topo.c
```diff
@@ -148,12 +148,30 @@ void chpl_topo_init(void) {
...
+         obj != NULL;
+         obj = hwloc_get_next_obj_by_depth(topology, numaLevel, obj)) {
+      hwloc_bitmap_and(cpuset, obj->online_cpuset, obj->allowed_cpuset);
+      if (!hwloc_bitmap_iszero(cpuset)) {
+        numNumaDomains++;
+      }
+    }
+
+    hwloc_bitmap_free(cpuset);
+  }
```
dmk42: The latest hwloc release that we picked up has a working hwloc_get_nbobjs_inside_cpuset_by_depth that could be used instead of this code.  In that past, that function was buggy, but the release we have now fixes it.

ronawho:
And there's a benefit, even if we're not using multi-ddata? If so, I don't understand how

dmk42:
@ronawho I see what you're saying.  If you localize a page but the allocator doesn't know to take advantage of that, there is no gain.  OK, that makes sense.  It shouldn't hurt KNL but it probably doesn't help without multi-ddata.

gbtitus:
I'm not following "And there's a benefit, even if we're not using multi-ddata?".  We get the multi-ddata support for any target including KNL if the locale model says (at compile time via `param localeModelHasSublocales)` that sublocales may exist, and if so then we get multiple ddata chunks at execution time if there are multiple sublocales (and the array is big enough).  `localeModelHasSublocales` is true for locModel=knl, so we should have multi-ddata support there.

runtime/src/chpl-topo.c
gbtitus: Cool, I'll give it a try.

ronawho:
Ok, I had been under the impression that multi-ddata was only being used for numa, but if that's not the case and it's being used for knl we should localize there too

gbtitus:
Yeah, we had wanted multi-ddata at least available for locModel=knl because in some modes (which are selected so late, at user program execution time!) it does have NUMA characteristics.

runtime/src/chpl-topo.c
gbtitus: Done, nice change and does work - thanks.

dmk42:
Looks good!

PR #5741, opened 2017-03-20
Add a .future about returning a tuple-type in a conditional expression

daviditen:
This gets an error about illegal value where type was expected even though
both sides of the conditional expression are types.


mppf:
Oddly enough, this works:
``` chapel
param useInt = false;

proc getType(type t) type {
  return if useInt then int else 2*int;
}

var a: getType(2*int);
writeln(a.type:string);
```

PR #5742, opened 2017-03-20
Work around a bug involving tuple types in conditional expressions

daviditen:
The void-multi-ddata fields change ran into a bug (see PR #5741) that caused
errors about using a value where a type was expected. Use a conditional
statement with two returns instead of a conditional expression to work
around that bug.

I tested that all of the failing tests now compile for CHPL_LOCALE_MODEL=numa
And did a plain linux64 paratest


ronawho:
Looks ok to me

daviditen:
Thanks!

PR #5744, opened 2017-03-20
Module deinit fns and deinitialization of globals

### LINKED ISSUE #5627: Distributed domain access from record destructor seg faults
### Summary of Problem

Accessing a distributed domain (or any array declared over that domain) from within a record destructor results in a segfault for `CHPL_COMM=none` and a `dereference to nil` error for `CHPL_COMM=gasnet`.

This error did not occur in version 1.14.0.

### Steps to Reproduce

**Source Code:**

```chapel
// Error occurs for any distribution, just picking Block as a simple example
use BlockDist;

const blockDomain = {1..1} dmapped Block(boundingBox={1..1});

const Domain = {1..1};

record Test {
  proc ~Test() {
    // nil by the time we call this deconstructor
    writeln(blockDomain);
    // still exists
    writeln(Domain);
  }
}

var t : Test;
```

**Compilation & Execution:**

 On master today (from `a9a5e95814a6` / #4762 and on):
```bash
 > chpl --comm=gasnet foo.chpl && ./a.out -nl 1

  $CHPL_HOME/modules/internal/ChapelArray.chpl:2350: error: attempt to dereference nil

 > chpl --comm=none foo.chpl && ./a.out

  Segmentation Fault: 11
```

On Chapel 1.14.0 (up to `d25e9279a127`)

```bash
 > chpl --comm=gasnet foo.chpl && ./a.out -nl 1

  0
  0

 > chpl --comm=none foo.chpl && ./a.out

  0
  0
```

*TODO*

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.14.0.ee1d3fd` (HEAD of master)

vasslitvinov:
This fixes #5627 and introduces user-writable module deinit functions.

MODULE DEINIT FUNCTIONS

* Allow users to declare 'proc deinit()' at the module level.
These are "module deinitialization" functions.

Their codegen names are: chpl__deinit_ + module name

They are executed after the program's main function completes
in reverse order of module initialization functions chpl__init_

* To ensure proper deinitialization order, each module init fn
places the corresponding deinit fn (if present) into a linked list.

The linked list is implemented in module code.
Deinit fns are dropped on the floor in --minimal-modules mode.

The linked list is traversed and the deinit fns are invoked
through chpl_deinitModules(). The modules being deinit'ed
are shown upon --printModuleDeinitOrder

Interestingly, to pass an address of a function represented
by a FnSymbol into Chapel code, it is sufficient to wrap
that FnSymbol into a SymExpr, giving us a c_fn_ptr.
(Required a fix to codegen.)

To execute a function through a pointer, I added, in runtime,
chpl_execute_module_deinit(). Despite its name, it is general-purpose
for invoking a 0-arg function through a c_fn_ptr.
We can rename it to be general-purpose if desired.

CALL TREE

The generated C code looks like this now:

```
chpl_executable_init <-- invoked as chpl_task_callMain(chpl_executable_init)

   chpl_std_module_init

      chpl__initStringLiterals  // set up string literals on each locale
      chpl__heapAllocateGlobals // allocate global vars on heap for multilocale

      if (chpl_nodeID == 0)

         chpl__init_preInit
            chpl__init_MODULE_p = false   // all MODULEs

         chpl__init_PrintModuleInitOrder  // no dependents

         chpl__init_ChapelStandard
            chpl__init_MODULE          // all MODULEs included for hello-world
                                       // i.e. used from ChapelStandard

   chpl_gen_main

      allocate+init an _EndCount (into which global?)
      chpl_rt_preUserCodeHook

      chpl__init_UM1
      chpl_user_main

      chpl_rt_postUserCodeHook
     _waitEndCount

     chpl_deinitModules
```

DEINITIALIZATION OF GLOBALS

We used to deinitialize globals whole-sale in chpl__autoDestroyGlobals,
in the order those globals are parsed. This test shows why this is
problematic:  modules/vass/deinit-order-globals

Now we are deinitializing them in the reverse order of initialization.
This is done by placing deinits of globals of each module at the end of
that module's deinit fn, in reverse order of declaration.
If the user hasn't written a deinit fn, the compiler creates one.
If there are no globals to deinitialize in a given module
and the user has not written a module deinit, then there won't be one.

TEST CHANGES

distributions/dm/t9.chpl used to wrap *twice* the same DimensionalDist2D
instance in dmap/_distribution wrappers. Now this would result in double-
deletion of that instance. So I removed the second wrap.

WHILE THERE

Removed a couple of comments that are now obsolete.

Removed a verify() check on initFn that is tautological.
Strengthened FnSymbol::verify().

OPEN QUESTIONS / FUTURE WORK

* It is an open question what is the best place for the deinitFn support
that I added to modules. See below...

* Can't put "var chpl_moduleDeinitFuns ..." in ChapelUtil.chpl because:
If so, it would gets assigned 'nil' when ChapelUtil module's initFn
is run. This happens after several other module initFns have been run
and this variable has already accumulated a list of their deinitFns.
When it is assigned 'nil', that accumulated contents is lost
so those deinitFns do not get executed at the end of the program.

* At the same time can't put "config var printModuleDeinitOrder ..."
in ChapelBase because config machinery requires parsing and ChapelRange
which are not yet available.

* I would love for printModuleDeinitOrder to be next to printModuleInitOrder
in a.out --help. However, it wasn't straightforward because
the PrintModuleInitOrder is special. We can work on it later.

* As the call stack above shows, module initialization is split
between standard modules and user modules (plus user-invoked modules):
there is a chpl_rt_preUserCodeHook between those.
This is not - and has not been - the case for deinitialization.
We may want to reciprocate.

* Future TODO on --printModuleInitOrder mode: Say there are modules User
and Helper. Module User uses Helper. Then the following happens:

  a line for User is printed
  a line for Helper is printed (indented vs. the previous line)
  Helper's module init is executed
  User's module init is executed

That is, there is a subtle difference between the order in which modules
are printed and the order in which their init functions are executed.
We may want to clean that.

* We may want to generate the sequence of deinit calls statically
instead of using a linked list. What about the same for init calls?
See #5746.

*  Is it OK to invoke resolveFns during callDestructors?
resolveFns() is probably not needed for a no-arg empty function
that is being created there - just set FLAG_RESOLVED?

Testing:
- [x] linux64 -verify
- [x] gasnet on multilocale tests
- [x] numa
- [x] valgrind


compiler/resolution/callDestructors.cpp
```diff
@@ -1004,34 +1004,52 @@ fixupDestructors() {
...
-
-  const char* name = "chpl__autoDestroyGlobals";
-  FnSymbol*   fn   = new FnSymbol(name);
+  SET_LINENO(mod);
+  FnSymbol* deinitFn = mod->deinitFn;
+  if (!deinitFn) {
+    deinitFn = new FnSymbol(astr("chpl__deinit_", mod->name));
+    mod->block->insertAtTail(new DefExpr(deinitFn));
+    normalize(deinitFn);
+    resolveFns(deinitFn);
```
benharsh: Seeing a ``resolveFns`` call in callDestructors concerns me a bit. Will 'chpl__deinit' have already been resolved in functionResolution? If not, what if it contains a loop?

vasslitvinov: This is a newly-created function - in case we need it and the user did not write a deinit().
It has empty body, so resolveFns() should do very little.

bradcray:
Yay!  Thanks Vass!

One question that was not clear to me in the PR message (though maybe I skimmed too quickly):  If a user creates a module deinit() function, I'm guessing the compiler still creates its own deinit which takes care of things a user can't/shouldn't deinit and that this is invoked after the user's deinit function?

bradcray:
Also, @nspark will be happy to hear that this feature is finally seeing the light of day...

vasslitvinov:
@bradcray I haven't thought about having the compiler create its own deinit() function. We may want to do so to isolate compiler stuff from user stuff.

If we do that, I would imagine that the compiler-created function would invoke the user function before deinit of globals.

Right now deinit of globals is simply added at the end of the user function. If deinits are needed and there is no user function, the compiler creates one.

PR #5745, opened 2017-03-20
Attempt to fix locality queries for default rectangular stuff

bradcray:
This is a five-minute fix (TM) to the 'localSubdomain*' queries for
DefaultRectangular domains and arrays.  Previously, these always
returned the complete domain's index set regardless of whether
or not the domain/array was stored on a given locale or not.  Now,
I only return those indices if the object is stored on the locale in
question.

This raises a bit of a design question:  For arrays, I believe it's
appropriate to return the empty domain if none of the array's
elements are owned by a given locale.  But given that a default
rectangular domain may live on a different locale than its array,
is it appropriate to say that the local indices of a default
rectangular domain are empty anywhere?  For now, I've said
"yes", but I could imagine this being the wrong decision, depending
on what these were used for.


bradcray:
@benharsh and @mppf : You two have done the most with this interface so far so are the appropriate reviewers for this change.  Note that there's a semantic question about whether or not I've done the right thing for domains above (or, arguably, for any of it).  We should probably air this in front of the group as a whole, but I thought I should start with the two of you since you've spent the most time with this interface so far.

modules/internal/DefaultRectangular.chpl
```diff
@@ -630,11 +630,21 @@ module DefaultRectangular {
     proc dsiHasSingleLocalSubdomain() param return true;
 
     proc dsiLocalSubdomain() {
-      return _getDomain(this);
+      if (this.locale == here) {
+        return _getDomain(this);
+      } else {
+        var a: domain(rank, idxType);
```
benharsh: I think this may need to specify 'stridable' as well.

bradcray: Is your thinking that we'll get a type mismatch if the domain/array itself is stridable?  In retrospect, I think you're right.

PR #1649, opened 2015-03-17
Bug fix for limiting the max size of qthreads memory pools

ronawho:
In cee911bf5a2e we added code to limit the max size of qthreads memory pools.
However, this code only limited the pool size when qthreads was adjusting the
initial alloc_size up if it thought it wasn't allocating enough bundles for
pooling to be worthwhile. It did not handle if the initial alloc_size was too
large.

Now if the initial alloc_size is too large, we round item_size up so it's a
multiple of pagesize (since qthreads expects allocations to be a multiple of
pagesize) and we find the max number items of that size that will fit within
max_alloc_size.

This bug was discovered while trying to figure out why ugni+qthreads seemed to
use so much more memory. The reason we noticed this bug there is because we ran
on some resource limited machines and guard pages don't work when ugni is being
used as the comm layer. In reality this was broken anytime we turned off guard
pages we just didn't notice it before.

With guard pages on, item_size just happened to work out to be a multiple of
pagesize so we got lucky and successfully limited the pool size. With guard
pages off stack size wasn't a multiple of pagesize, and the initial alloc_size
was over 1GB.

PR #5748, opened 2017-03-20
Limit the maximum size of qthreads memory pools to 65MB

ronawho:
Further limit the max pool size for qthreads now that we know there's a stack
pool per worker instead of a global pool.

In #257 (and #1649) we added some code to limit the size of qthreads memory
pools. Qthreads uses memory pools for things like task stacks, thread private
data, and runtime data. qthreads assumed that the item size for things in the
pool is small, however our default stack size is quite large (8MB.)

I did something like `QT_MAX_POOL_ALLOC_SIZE=2 * hwpar * callStackSize` with a
cap of 513MB (68 default-sized task stacks) to limit how much memory we'd use.
At the time I believed that there was a single global pool for task stacks,
so I wanted to make sure the pool would hold at least maxTaskPar (hwpar)
stacks. It turns out that there is actually a pool per worker.

On KNL, this meant we were trying to create 68 pools that each consumed 513MB of
memory. This ended up being ~34GB just for task stacks, which is outrageously
large. For now, just limit the max pool size to 65MB instead of 513MB (room of
8 default-size task stacks)

Note that this _could_ hurt performance for benchmarks that create a lot of
nested parallelism. It doesn't seem to hurt any of our existing benchmarks, and
it shouldn't hurt long lived tests because the allocation cost should be
amortized over the length of the program but it could hurt micro-benchmarks
that test nested parallelism.

### DIFF
--- a/runtime/src/tasks/qthreads/tasks-qthreads.c
+++ b/runtime/src/tasks/qthreads/tasks-qthreads.c
@@ -646,8 +646,8 @@ static void setupCallStacks(int32_t hwpar) {
         // pages, so we thrown an extra MB.
         if (hwpar > 0) {
             const size_t oneMB = 1024 * 1024;
-            const size_t allocSizeLowerBound =  33 * oneMB;
-            const size_t allocSizeUpperBound = 513 * oneMB;
+            const size_t allocSizeLowerBound = 33 * oneMB;
+            const size_t allocSizeUpperBound = 65 * oneMB;
             size_t maxPoolAllocSize;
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus any objections?

ronawho:
If you're gone for the night, I'll probably merge since the code change is simple.

gbtitus:
This seems fine to me.  I'll be surprised if a situation arises where we can improve performance by increasing the pool size per worker beyond this, at least in a real application or mini-app or other program not designed specifically to test or stress overheads.

PR #5750, opened 2017-03-21
Build amudprun for the host

dmk42:
By default GASNet builds the amudprun launcher for the target machine (compute nodes).  It should be built for the host (login node).  However, the library on which its protocol is based is still needed for both the host and the target.

This change leaves the GASNet build untouched, allowing it to build the target version of the library, and adds a second build of amudprun, which runs on the host.  This second build will only occur if CHPL_LAUNCHER=amudprun .

Tested on 64-bit ARM, Mac, Linux, and Cray XC.


ronawho:
Just FYI amudprun is not the reason we build the cray module for the lowest common denominator :)

ronawho:
I think third-party/amudprun deserves a small README indicating what it's for

ronawho:
Wonder if we should update third-party/README or LICENSE at all

ronawho:
Code changes look good to me, and updates to READMEs can happen post feature freeze. 

@gbtitus will probably want to review too, but that can be done post-commit.

dmk42:
Thanks.  I added the README info and updated the PR comment in response to the review.

gbtitus:
The need for this makes total sense to me and I'm a little surprised we never ran into a situation where we needed it before.  Three comments:

- I'm a little surprised to see this being placed into a top-level subdir under third-party.  Was a subdir or tree of subdirs under third-party/gasnet/install considered and rejected?  That seems preferable from an organizational point of view.

- It seems like cleaning or clobbering or building third-party/gasnet should also do the same for the launcher derived from the same sources.  In other words, if I do `make -C third-party/gasnet clean all` it should clean and rebuild the host-configured amudprun as well as the target-configured libgasnet etc.

- Are there other GASNet launchers for which we should do the same thing now, such as `gasnetrun_mpi`?  Or is the plan to take these on as needed, case by case?  (I'd be okay with the latter.)

dmk42:
Thanks, Greg.  I have no objection to moving this at some point.  One reason the top level was chosen this time was because the GASNet Makefile sets CHPL_MAKE_HOST_TARGET=\-\-target, and I needed to set it to \-\-host.  It should be possible to override this in a subdirectory, though.  Another reason was so that it would automatically be built when CHPL_LAUNCHER=amudprun .  I can see the disadvantage with respect to cleaning or clobbering, though.

Some of the other GASNet launchers are scripts, so they don't need a separate host instance.  The MPI-related launchers that are alongside amudp in gasnet-src/other are compiled, though, so it would be reasonable to make those available for the host as well.

I've been trying to convince the GASNet folks that not building these launchers for the host is a bug, and they have said that they might want to look at that someday, if it bubbles up the priority list.  If they take care of it, we can delete this directory altogether.  If not, a reasonable way to reorganize this would be to have a third-party/gasnet/host subdirectory and build all the non-script launchers in there.

PR #5752, opened 2017-03-21
Deprecate array alias

benharsh:
The ``=>`` operator has always been a bit too special/weird and with the advent of ``ref`` variables, we believe it can be done away with. This PR adds a deprecation warning for the ``=>`` operator and updates our test suite to use ref-vars instead.

Credit to @mppf and @bradcray for doing the bulk of the work related to this PR.

Testing:
- [x] full local
- [x] local futures
- [x] valgrind
- [x] subset of gasnet
  - [x] memory/sungeun/refCount/
  - [x] primers


benharsh:
So, I got a little too hung up on trying to fix some of the remaining errors in #5576, when I should have just accepted the differences and moved on.

HPL 2012 will continue to use the ``=>`` operator, but that test is already sufficiently weird that I'm fine living with it for this release.

Would either of you like me to do any testing aside from what's listed above?

test/arrays/ferguson/loc-arr-slice.good
```diff
@@ -2,9 +2,5 @@ loc-arr-slice.chpl:1: In function 'badReturnRefLocal':
 loc-arr-slice.chpl:3: error: illegal expression to return by ref
 loc-arr-slice.chpl:6: In function 'badReturnRefConstant':
 loc-arr-slice.chpl:7: error: function cannot return constant by ref
-loc-arr-slice.chpl:19: In function 'badReturnAliasLocal':
-loc-arr-slice.chpl:21: error: illegal return of array aliasing a local array
-loc-arr-slice.chpl:31: In function 'badReturnAliasAliasLocal':
-loc-arr-slice.chpl:33: error: illegal return of array aliasing a local array
```
bradcray: It seems like, ideally, we'd like to preserve these warnings, right?  To that end, could we create a version of this test that had the old .good as a future, with a "retire this test once it's working and the original's .good is updated/restored" type of note in its .future?

test/arrays/vass/const-checking.aliases.good
```diff
@@ -1,9 +1,2 @@
-const-checking.aliases.chpl:8: In function 'main':
-const-checking.aliases.chpl:10: error: illegal lvalue in assignment
-const-checking.aliases.chpl:12: error: illegal lvalue in assignment
-const-checking.aliases.chpl:14: error: creating a non-const alias 'aa2' of a const array or domain
-const-checking.aliases.chpl:18: error: creating a non-const alias 'dd2' of a const array or domain
-const-checking.aliases.chpl:23: error: illegal lvalue in assignment
-const-checking.aliases.chpl:24: error: illegal lvalue in assignment
-const-checking.aliases.chpl:27: error: illegal lvalue in assignment
-const-checking.aliases.chpl:28: error: illegal lvalue in assignment
```
bradcray: This is another one where it seems like ensuring that these error messages still occur is important.  Originally, I created a clone of this test that commented out lines 14 and 18 on the hypothesis that different passes created the errors and that I might still get the errors on the original lines once I got past the first few.  But I can't remember what happened in that experiment (only that I've been tripping over a "missing .good" error message ever since...).  Probably I timed out before completing it...  If creating such a variation of the test did not generate warnings for the other lines, that seems like another case to file a future against?

test/arrays/ferguson/loc-arr-slice.good
benharsh: The difference is that now this will copy-out, so I'm not sure the errors entirely make sense anymore

test/arrays/vass/const-checking.aliases.good
benharsh: If you turn the bad 'ref' into a 'const ref', you should then see the "illegal lvalue in assignment" failures (I will verify this on my end, but I'm pretty sure I tried this before).

test/arrays/ferguson/loc-arr-slice.good
bradcray: Oh yes, you're right.  Cool!

test/arrays/vass/const-checking.aliases.good
bradcray: I think we're saying similar things: that because the compiler exits prior to hitting the pass that prints the errors, we don't hit them anymore.  This is what I was suggesting in cloning the test with those lines commented out:  We get checks that both kinds of conditions generate errors, whereas by accepting this .good without the clone, we lose coverage of certain error cases that this test was locking in before (which may also be covered by other tests, but who knows...?).

benharsh: Ah, I see. I will prepare a variant that looks at assignment to elements.

bradcray:
I'd definitely do valgrind testing.  GASNet seems less obviously important to me, though maybe that's just laziness speaking.  IIRC, some of the updates were to .future files, which made me think I should make sure to run them, ensure they were still relevant, add .bad files if they predated that feature, etc.

It seems as though HPL 2012 should have a .good update for the deprecation warning, shouldn't it?

Ultimately, we should also add a test to test/deprecated to help us remember to remove this support in the next release (at which point we'll have to do something about HPL 2012 if nobody's put it out of its misery before then).

benharsh:
Ah, I may have neglected to rebuild my parser... I'll get on that tomorrow.

I'll also add a test to deprecated/

bradcray:
In converting the holdout hpl.hpcc2012.chpl over to avoid the use of '=>', I realized for the first time that a difference between `ref` and `=>` is that, for a privatized array, `=>` would create a privatized reference to the array, permitting it to be accessed within a `local` block from anywhere.  In contrast, `ref` creates only a local reference to the array which therefore can't be accessed from within a remote local block.  This doesn't change my opinion about deprecating `=>` (particularly since my hope is to also deprecate the `local` block), but it seemed worth noting.  We discussed some options for optimizing such cases, such as remote value forwarding references and, when possible, the things they refer to.  When the thing being referred to is a privatized array, it seems we could r.v.f. the privatization ID to get a similar effect as before.

bradcray:
In working on removing the deprecated '=>' feature, @noakesmichael found the following cases still use it, besides hpl.hpcc2012.chpl, which was already known above and which I'm actively working on rewriting:

* studies/cholesky/jglewis/version2/elemental/unsymmetric_indices/test_elemental_cholesky_unsymmetric_ranges
* test/users/jglewis/array_reindexing
* studies/ssca2/rachels/SSCA2_test
* arrays/deitz/part1/test_array_arg6_auto
* arrays/deitz/part1/test_array_arg7_auto
* arrays/deitz/part1/test_array_arg8_auto
* arrays/deitz/part1/test_array_arg9_auto
* arrays/deitz/part3/test_slice_stride2_auto.chpl]
* arrays/deitz/part3/test_slice_stride3_auto.chpl]
* arrays/deitz/part3/test_slice_stride4_auto.chpl]

PR #5757, opened 2017-03-21
Have reindex domains/distributions preserve locality

bradcray:
This implements reindexing array views such that locality is preserved for their domains and distributions such that any operations or new domains created from a reindex view have the same distribution as the original array being reindexed.  Notably, it permits the viewPreserveDist.chpl test to be unfuturized, locking in all array views to be locality-preserving.

The main part of the work here was to implement the domain and distribution view classes for
reindexing operations.  The nice side effect was the ability to retire all remaining per-distribution
support for reindexing domains and distributions.  Most of this was straightforward.


bradcray:
One failure for:

benharsh:
A couple of observations from local spelunking:

I don't think you should be passing ``updom`` to the reindex domain constructor in ``dsiNewRectangularDom``. The original ``updom`` could be unstrided but the new
``inds`` may be strided. For example:
```chpl
var A : [0..9] int;
ref B = A.reindex({1..10});
ref C = A[1..10 by 2]; // should be a strided reindexed domain, I think
```

I'm not sure about the order-to-index stuff yet.

benharsh:
Just ran a gasnet paratest. It came back with one failure for the reindex API test due to a mismatch in the local subdomain output (that I mentioned in the PR on your repo).

bradcray:
Closing temporarily to clean up the PRs area on request. Added issue #6434 as a reminder to get back to these.

bradcray:
Currently failing two tests for GASNet:

```
[Error matching program output for arrays/userAPI/unsignedOps]
[Error matching program output for arrays/userAPI/reindexOps2D]
```

bradcray:
@benharsh: I'm running a final gasnet test, but believe this is ready for review if you've still got some energy for array view reviews.  If not, let me know and I'll find another ~~victim~~ volunteer.

modules/internal/ArrayViewReindex.chpl
```diff
@@ -26,6 +26,268 @@ module ArrayViewReindex {
   use ChapelStandard;
 
   //
+  // This class represents a distribution that knows how to create
+  // reindexed domains and arrays similar to the one that caused
+  // it to be created.  It stores a pointer down to the distribution
+  // that it creates lower-dimensional views of, and two tuples
+  // indicating which dimensions were collapsed by the reindexed
+  // and what the indices of those dimensions are.
```
benharsh: This comment mentions collapsed dimensions, was this mistakenly copy/pasted from the rank-change array view?

PR #5758, opened 2017-03-21
Remove -q from diff command in checkChplInstall

mppf:
I noticed that diff -q is not portable (at least on OmniOS it is an error). At the same time, on Linux and Mac OS X, whet 'make check' works correctly, diff does not seem to output anything anyway.

The -q was added along with the diff to this script in c2f09379750f3e068def8cb0346758c9f078c415.

Verified make check looks OK on Mac OS X and on Linux.

### DIFF
--- a/util/test/checkChplInstall
+++ b/util/test/checkChplInstall
@@ -217,7 +217,7 @@ TEST_EXEC_OUT=${TEST_JOB}.exec.out
     log_info "Test job complete."
 
     # Check result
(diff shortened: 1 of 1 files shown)


ben-albrecht:
I see no problem with dropping the `-q`. 

@awallace-cray has led many of the recent changes to this script -- maybe he can doubly confirm this is OK.

awallace-cray:
Looks good to me.

ben-albrecht:
Double checked the output of `make check` on OS X. Looks good!

Issue #5536, opened 2017-03-06
Multilocale test failures when using gasnetrun_ibv as launcher.


epaulson10:
### Summary of Problem

I have been making some contributions to GASNet's ofi-conduit and I have been trying to test Chapel's multilocale functionality with my changes. I am seeing odd errors related to process spawning when using gasnetrun_ibv as the CHPL_LAUNCHER value.  I will use test/multilocale/deitz/needMultiLocales/dist/test_private_space.chpl as an example in this bug. 

What I am seeing with this test is the data that is owned by the first node is all zeros. For example:
$ ./test_private_space -nl 2
0 1
diagnostics = (get = 0, get_nb = 0, put = 0, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 0, execute
_on_fast = 0, execute_on_nb = 1) (get = 0, get_nb = 0, put = 0, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, exec
ute_on = 0, execute_on_fast = 1, execute_on_nb = 0)
0 0 0 0 5 6 7 8
0 26
diagnostics = (get = 0, get_nb = 0, put = 0, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 0, execute
_on_fast = 0, execute_on_nb = 2) (get = 0, get_nb = 0, put = 0, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, exec
ute_on = 0, execute_on_fast = 2, execute_on_nb = 0)
0: 0
1: 26

Here the 0 0 0 0 5 6 7 8 should be 1 2 3 4 5 6 7 8, showing that the first node is having some error.

This always happens if I use the SSH-based spawner (meaning that you pass -spawner=ssh to gasnetrun_ibv or set GASNET_IBV_SPAWNER=ssh). If I use MPI as GASNet's spawning mechanism, I can only get correct results on this test if gasnetrun_ibv spawns more processes than I have nodes listed in my hostfile. So for example, if I have 3 available nodes and I spawn the program with "gasnetrun_ibv -n 3 -spawner=mpi ./test_private_space_real -nl 3", I get the following incorrect output (diagnostics removed):
0 1 2
0 0 0 4 5 6 7
0 15 15
0: 0
1: 15
2: 15

If I change "-n 3" in the command above to "-n 4", then I get output that I expect:
0 1 2
1 2 3 4 5 6 7 8
[...]

PR #5759, opened 2017-03-21
Don't ever report a number of CPUs less than 1 (one).

gbtitus:
A user issue (https://github.com/chapel-lang/chapel/issues/5536) reports
test failures that ultimately are caused by chpl_getNumPhysicalCpus()
returning 0 (zero).  This change ensures that both of the functions that
return CPU counts always return 1 (one) or higher.

In the next release much of this code should be reworked in favor of
getting the value from the topology layer, but that's too big a change
for code freeze week.


gbtitus:
@ronawho, would you review this one, please?

ronawho:
👍

PR #5761, opened 2017-03-21
Use hw.ncpu to compute # cpus on FreeBSD

mppf:
Avoids a compilation warning with FreeBSD 11 in quickstart mode. The warning we would see is:

```
chplsys.c:568:2: warning: "Target architecture is not yet supported."
      [-W#warnings]
#warning "Target architecture is not yet supported."
 ^
chplsys.c:639:2: warning: "Target architecture is not yet supported."
      [-W#warnings]
#warning "Target architecture is not yet supported."
 ^
2 warnings generated.
```

* full local testing
* verified clean build on FreeBSD 11
* verified here.numPUs reports a reasonable value on FreeBSD 11

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -58,7 +58,7 @@
 #include <sys/mman.h>
 #include <sys/utsname.h>
 
-#if defined(__APPLE__) || defined(__NetBSD__)
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


mppf:
@gbtitus - mind having a look at this change to chpl_getNumPhysicalCpus / chpl_getNumLogicalCpus enabling basic functionality on FreeBSD? I was doing some pre-release portability testing to shake out any portability bugs that have cropped up recently and noticed C compilation warnings on FreeBSD. Admittedly FreeBSD isn't our main platform, but I think it's nice to have a clean quickstart / make check there if it's do-able. Thanks.

gbtitus:
It's nice to be able to cover more OS cases, yeah.  Okay to merge.

One question: unless I missed something, both the new blocks of `#if'd` code are identical to cases that were already there before: `__CYGWIN__` for the first one, and `__APPLE__` for the second one.  Was there a particular functional reason not to combine the new FreeBSD case with these, such as a likelihood of future divergence, or is the separation more for clarity/appearance?

mppf:
@gbtitus - I don't know but I figured Cygwin was likely to diverge from FreeBSD in the future. I think it's clearer this way though, anyway... I could combine them if you think it's better to do so - one of those arbitrary decisions IMO.

gbtitus:
Nah, leave it as is.  I think there's some possibility of divergence also.  Plus with any luck we'll be able to replace much of this with hwloc calls sometime soon, and it won't matter anyway.

PR #5763, opened 2017-03-21
Switch array view playground over to new reindex work

bradcray:
This switches the array view playground over to my new "preserve locality for domains/distributions of reindex views branch"


ronawho:
👍

Issue #5751, opened 2017-03-21
FFTW and MPI should cleanup using module-level deinit


npadmana:
Looking ahead to a post-#5744 world with module level deinits, both FFTW and MPI should cleanup using this functionality.

- [ ] FFTW -- currently doesn't automatically cleanup at the end, but relies on the user to do so. The module should call ``cleanup()``at the end.
- [ ] FFTW_MT -- again, ``fftw_cleanup_threads`` should be called at module deinit.
- [ ] MPI -- The MPI module currently wraps its initialization in a record, with an explicit ``deinit`` routine. All of the setup/teardown logic can now be moved to the module level.

Not quite a module-level issue, but --
- [ ] FFTW plans should be wrapped in a ``record`` with an explicit ``deinit`` to clean up when it goes out of scope. (A subtlety here is if the record is somehow copied, in which case the underlying plan may be deallocated. We could reference count, but a warning that plans are not meant to be copied may be sufficient).

vasslitvinov:
#5744 has been merged.

bradcray:
Good proposal, thanks Nikhil!

PR #5769, opened 2017-03-21
Cleanup MPI with module deinit

npadmana:
The MPI module previously used a module-level record to
handle cleanup of the MPI module. With the introduction of
module level deinits, this is no longer necessary, and so we
remove the record, and use a module deinit.

Note that the issue of CHPL_COMM_WORLD being deallocated too 
early has now been resolved upstream (thanks!).

Tested on Linux --
 
 CHPL_COMM=none 
    (modules/packages/mpi/spmd/hello-chapel)
 CHPL_COMM=gasnet CHPL_COMM_SUBSTRATE=mpi 
   (modules/packages/mpi/multilocale/hello-chapel)

This partially addresses #5751 (point 3).

### DIFF
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -213,13 +213,9 @@ module MPI {
   // Module level deinit
   proc deinit() {
     if _freeChplComm {
-      if numLocales > 1 {
-        coforall loc in Locales do on loc {
-            C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
-          }
-      } else {
-        C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
-      }
+      coforall loc in Locales do on loc {
+          C_MPI.MPI_Comm_free(CHPL_COMM_WORLD_REPLICATED(1));
+        }
     }
     if _doinit {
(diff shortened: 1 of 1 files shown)


modules/packages/MPI.chpl
```diff
@@ -208,42 +208,37 @@ module MPI {
...
-    proc deinit() {
-      if freeChplComm {
-        if numLocales > 1 {
-          coforall loc in Locales do on loc {
+  var _doinit : bool = false;
+  var _freeChplComm : bool = false;
+  // Module level deinit
+  proc deinit() {
+    if _freeChplComm {
+      if numLocales > 1 {
```
ben-albrecht: While we're here, differentiating `numLocales > 1` vs. `numLocales == 1` shouldn't be necessary here. The `coforall` should cover both cases.

ben-albrecht:
Changes look good. My comment was an additional request.  I just need to test it before merging:

- [ ] Test on linux64
- [ ] Test on Cray

modules/packages/MPI.chpl
npadmana: Fixed.

npadmana:
Re-tested the two configurations described in the PR.

PR #5771, opened 2017-03-21
Inline DefaultRectangular 1D strided iter

benharsh:
As far as I can tell, DefaultRectangular 1D strided iterators have never inlined (either directly or when zippered). We recently noticed a regression in a performance test that tracks 1D strided iteration performance. I'm not entirely sure of the cause but I believe b8fe0872 , which switched a call from 'ranges' to 'dsiDim', is involved. My guess is that creating that range every iteration was slightly more costly.

The simplest solution seems to be to rewrite this code such that it will inline, primarily by only having a single yield (as opposed to two across runtime-conditional branches).

This change not only resolves the regression, but brings the 1D strided performance even with 1D dense performance.

Testing:
- [x] full local

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -1944,9 +1944,9 @@ module DefaultRectangular {
               start  = viewDomDim.first,
               second = info.getDataIndex(start + stride);
 
-        var   first  = info.getDataIndex(start),
-              step   = (second-first):chpl__signedType(viewDom.idxType),
-              last   = first + (viewDomDim.length-1) * step:viewDom.idxType;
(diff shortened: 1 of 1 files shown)


benharsh:
@bradcray , @ronawho : This PR resolves the regression we were discussing:

http://chapel.sourceforge.net/perf/chapcs/?startdate=2017/02/25&enddate=2017/03/21&graphs=serial1darrayperformance

ronawho:
👍 👍

bradcray:
[Wow, three thumbs up?  He really is your favorite, huh?]

Looks good to me, thanks for taking care of this, Ben!

PR #5772, opened 2017-03-21
Remove the rule to generate STATUS from STATUS.devel

daviditen:
Remove the STATUS file completely.  Remove the rule to generate it and remove
it from the .gitignore file.

Since we only intend to use the STATUS information internally, STATUS.devel is
more useful anyway.  It includes pointers to the tests in question.


bradcray:
Looks good to me, thanks David!

PR #5773, opened 2017-03-21
Handle .bad mismatch for configHelp.chpl

vasslitvinov:
#5744 caused .bad mismatch on configHelp.chpl

Instead of adjusting .bad, I am adding a .prediff
that greps out most of the normal help message.

That way comparing against .bad will be insensitive to most of the
specifics of the help message that may again change in the future.
(Yes, I adjusted .bad for this prediff.)

I am preventing the .prediff from messing with the output
if it already matches .good, just because I can.

### DIFF
--- a/test/execflags/bradc/configHelp.prediff
+++ b/test/execflags/bradc/configHelp.prediff
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# Make the comparison against .bad insensitive
+# to most of the specifics of the help message.
+
+# ... except if the output already matches the .good.
+cmp -s $1.good $2 && exit
+
+set -x
+grep -E 'CONFIG VARS|numLocales:|help|custom' $2 > $2.tmp
+mv $2.tmp $2
--- a/test/execflags/bradc/configHelp.bad
+++ b/test/execflags/bradc/configHelp.bad
@@ -1,39 +1,4 @@
-
-FLAGS:
-======
   -h, --help            : print this message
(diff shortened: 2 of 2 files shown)


test/execflags/bradc/configHelp.prediff
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Make the comparison against .bad insensitive
+# to most of the specifics of the help message.
+
+# ... except if the output already matches the .good.
+cmp -s $1.good $2 && exit
+
+set -x
+grep -E 'CONFIG VARS|numLocales:|help|custom' $2 > $2.tmp
```
daviditen: What is "custom" for in the pattern?

daviditen:
I left an inline comment about the pattern, but looks good to me otherwise.

vasslitvinov:
@daviditen 'custom' allows the good input to go through.

I left it in there in case my "compare against .good and quit if match" line above does not fire for some reason. (Although I tested it and it *should* fire.)

bradcray:
I don't object to taking this approach, but does this mean that we're not changing the new flag to a config param as proposed?

vasslitvinov:
@bradcray It allows testing to pass independently of what we do with the new flag.

bradcray:
Agreed, I just wasn't sure what happened to that conversation and suspect that the level of effort to change the flag would've been lower.

PR #5774, opened 2017-03-21
Be sure the NUMA objects we refer to have CPUs.

gbtitus:
To get a NUMA sublocale object we were calling hwloc_get_obj_by_depth()
to get the Nth object at the topology level corresponding to NUMA
domains, where N corresponded to the Chapel sublocale ID.  This was
coded under the assumption that the things at the level were all
traditional CPU+memory NUMA domains.  That's not true in some of the KNL
configurations.  We fixed one problem in #5740, to avoid counting NUMA
domain objects not containing CPUs when computing our own NUMA sublocale
count when using Cray KNL nodes in the "quad" mode.  However, it wasn't
enough.  In the "snc4" mode we get 4 groups of pairs of NUMA objects,
with the first object in each pair containing 1/4 of the CPUs and 1/4 of
the regular memory, and the second containing 1/4 of the MCDRAM.  The
objects in order at the NUMA level in the topology are these alternating
CPU+DDR and MCDRAM ones, and using our sublocale IDs to index them gets
us the wrong (MCDRAM, not CPU+DDR) object for all odd sublocale index
values.  So here, switch to using hwloc_get_obj_inside_cpuset_by_depth()
instead, using the full cpuset from the root object to limit retrieved
objects to those containing CPUs in that set.

### DIFF
--- a/runtime/src/chpl-topo.c
+++ b/runtime/src/chpl-topo.c
@@ -68,10 +68,11 @@ static int numaLevel;
 static int numNumaDomains;
 
 
+static hwloc_obj_t getNumaObj(c_sublocid_t);
 static void alignAddrSize(void*, size_t, chpl_bool,
                           size_t*, unsigned char**, size_t*);
 static void chpl_topo_setMemLocalityByPages(unsigned char*, size_t,
-                                            hwloc_obj_t numaObj);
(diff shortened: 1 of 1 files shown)


gbtitus:
@dmk42, would you mind giving this a quick look?

PR #5776, opened 2017-03-21
Document that CHPL_COMM owns setting of LD

bradcray:
@sungeunchoi pointed out that when establishing this convention, I'd failed to recall that these text files would be a good place to document it.


bradcray:
@sungeunchoi : Would you review?

PR #5780, opened 2017-03-22
Add system config defines for amudprun when appropriate

dmk42:
In #5750 I left out two defines that were important for the case of running with GASNET_WORKERIP set.  When amudprun is built for the target (the wrong machine), these defines are set by the configure script.  I could not take advantage of the target configuration for the host, so in this change, the Makefile does a simple test to determine whether to set them.

### DIFF
--- a/third-party/amudprun/Makefile
+++ b/third-party/amudprun/Makefile
@@ -5,6 +5,11 @@ endif
 CHPL_MAKE_HOST_TARGET = --host
 include $(CHPL_MAKE_HOME)/make/Makefile.base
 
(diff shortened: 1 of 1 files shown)


gbtitus:
This seems fragile, though I could easily understand it being the only practical way to do it.  Maybe add a comment to the Makefile expressing a wish for a better way?  If you do that I don't need to look at it again, and I'm also okay with you merging even if you don't do it.

dmk42:
I added a comment explaining how we'd really rather this be done differently (by the GASNet distribution itself).

gbtitus:
That's a superb comment.  :-)  👍

PR #5782, opened 2017-03-22
Fix handful of standalone iter NUMA regressions

benharsh:
We do not currently have a standalone iterator for DefaultRectangularDom under NUMA, so avoid this case entirely for now.


bradcray:
Looks fine to me.

benharsh:
Oh, @gbtitus and @vasslitvinov may be interested in another instance of this.

PR #5784, opened 2017-03-22
Tidy up --llvm testing

mppf:
* add missing header to chpl-tasks-callbacks.h
* improve a prediff to work with --llvm
* skip several tests under --llvm that don't make sense there
  -  test/compflags/albrecht/chplenv/chplenv would be hard to get the test right, since --llvm builds change CHPL_TARGET_COMPILER to clang-includes
   -  test/compflags/bradc/longLineCppLines, test/compflags/codegen/cppLines test something that only makes sense for the C backend
    - test/link/sungeun/static_dynamic.skipif doesn't work with --llvm and I think it's a test issue but am not sure yet
    - test/trivial/bradc/savecInNonWriteableDir.skipif has trivially different output with --llvm

With these changes, full --llvm testing is closer to coming back clean - the remaining issue has to do with different behavior for ~ on bools, but we're not yet sure if we want to allow ~ on bools (see #5787).

Passed full local testing (C backend).

The runtime change was reviewed by @gbtitus. Other skipif-etc changes don't need review.

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -422,7 +422,7 @@ module ChapelBase {
   //
   // bitwise operations on primitive types
   //
-  inline proc ~(a: bool) return __primitive("u~", a);
+  inline proc ~(a: bool) return __primitive("!", a);
   inline proc ~(a: int(?w)) return __primitive("u~", a);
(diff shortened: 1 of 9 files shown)


mppf:
Note, ~ changes were reverted in order to have more discussion. See #5787.

PR #5788, opened 2017-03-22
Improve loop vectorization for llvm back-end

coodie:
This is "work in progress" PR, where I'd like to keep my work associated with adding loop vectorization hints for llvm backend.

Here are subtasks to be done:
- Add `llvm.loop.vectorize.enable` metadata for every loop, and make sure they are passed properly. **DONE**

I consider this subtask to be done because in case of tests where loop could be vectorized but wasn't following error was printed:

```warning: <unknown>:0:0: loop not vectorized: failed explicitly specified loop vectorization```

I have also compared generated llvm code with clang generated code, where I have explicitly passed loop vectorization hints for selected loop using `#pragma clang loop vectorize(enable)` and metadata was passed in exactly same what as stated in http://llvm.org/docs/LangRef.html#llvm-loop-vectorize-and-llvm-loop-interleave and in clang generated llvm code.

The main difficulty was lack of existing examples on the internet and quite messy documentation. 

- Make sure that no `loop not vectorized` llvm errors are printed if loop wasn't vectorized.
- Add `llvm.mem.parallel_loop_access` for memory access instructions every loop.

- Investigate if loop vectorization can be improved even more like: give more hints, improve loop generation algorithm to be more adjusted to llvm vectorization algorithm. Some starting points might be here:
http://llvm.org/docs/Vectorizers.html
http://llvm.org/docs/doxygen/html/LoopVectorize_8cpp_source.html (contains references to papers)
http://llvm.org/docs/LangRef.html#llvm-loop-vectorize-width-metadata

- Finish the job: add tests and refactor.

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, true);
+        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, a_signed or b_signed);
       }
       ret.isUnsigned = !values.isSigned;
     }
@@ -1542,7 +1542,7 @@ GenRet codegenSub(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFSub(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateSub(values.a, values.b, "", false, true);
+        ret.val = info->builder->CreateSub(values.a, values.b);
       }
       ret.isUnsigned = !values.isSigned;
     }
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, a_signed or b_signed);
+        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, values.isSigned);
       }
       ret.isUnsigned = !values.isSigned;
     }
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b);
+        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, true);
       }
       ret.isUnsigned = !values.isSigned;
     }
@@ -1542,7 +1542,7 @@ GenRet codegenSub(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFSub(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateSub(values.a, values.b);
+        ret.val = info->builder->CreateSub(values.a, values.b, "", false, true);
       }
       ret.isUnsigned = !values.isSigned;
     }
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -30,7 +30,7 @@
 #ifdef HAVE_LLVM
 namespace
 {
-  void addLoopVectorizationHint(llvm::Instruction* instruction)
+  llvm::MDNode* addLoopVectorizationHint(llvm::Instruction* instruction)
   {
     GenInfo* info                  = gGenInfo;
     auto &C                        = info->module->getContext();
@@ -44,6 +44,7 @@ namespace
 
     dummy->replaceAllUsesWith(llvmLoopMetadata);
     instruction->setMetadata(llvm::StringRef("llvm.loop"), llvmLoopMetadata);
+    return llvmLoopMetadata;
   }
 }
 #endif
@@ -179,7 +180,14 @@ GenRet CForLoop::codegen()
 
     if(fNoVectorize == false && isOrderIndependent())
     {
-        addLoopVectorizationHint(endLoopBranch);
+        llvm::MDNode* llvmLoopMetadata = addLoopVectorizationHint(endLoopBranch);
+        for(auto it = blockStmtBody->begin(); it != blockStmtBody->end(); it++)
+        {
+            if(it->mayReadOrWriteMemory())
+            {
+                it->setMetadata(llvm::StringRef("llvm.mem.parallel_loop_access"), llvmLoopMetadata);
+            }
+        }
     }
 
     func->getBasicBlockList().push_back(blockStmtEnd);
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -27,6 +27,23 @@
 
 #include <algorithm>
 
+namespace
+{
+  void addLoopVectorizationHint(llvm::Instruction* instruction)
+  {
+    GenInfo* info    = gGenInfo;
+    auto &C = info->module->getContext();
+    llvm::ArrayRef<llvm::Metadata*> loopVectorizeEnable({llvm::MDString::get(C, "loop.vectorize.enable"),
+                                      llvm::ValueAsMetadata::get(llvm::ConstantInt::getTrue(C))}) ;
+    llvm::MDTuple* loopVectorizeMetadata = llvm::MDNode::get(C, loopVectorizeEnable);
+
+    llvm::ArrayRef<llvm::Metadata*> MDs({loopVectorizeMetadata });
+    llvm::MDTuple* llvmLoopMetadata = llvm::MDNode::getDistinct(C, MDs);
+
+    instruction->setMetadata(llvm::StringRef("llvm.loop"), llvmLoopMetadata);
+  }
+}
+
 /************************************ | *************************************
 *                                                                           *
 * Instance methods                                                          *
@@ -129,8 +146,10 @@ GenRet CForLoop::codegen()
                                                llvm::ConstantInt::get(condValue0->getType(), 0),
                                                FNAME("condition"));
 
+
     // Create the conditional branch
     info->builder->CreateCondBr(condValue0, blockStmtBody, blockStmtEnd);
+
 
     // Now add the body.
     func->getBasicBlockList().push_back(blockStmtBody);
@@ -154,7 +173,12 @@ GenRet CForLoop::codegen()
                                                FNAME("condition"));
 
     // Create the conditional branch
-    info->builder->CreateCondBr(condValue1, blockStmtBody, blockStmtEnd);
+    llvm::Instruction* endLoopBranch = info->builder->CreateCondBr(condValue1, blockStmtBody, blockStmtEnd);
+
+    if(fNoVectorize == false && isOrderIndependent())
+    {
+        addLoopVectorizationHint(endLoopBranch);
+    }
 
     func->getBasicBlockList().push_back(blockStmtEnd);
 
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -27,6 +27,7 @@
 
 #include <algorithm>
 
+#ifdef HAVE_LLVM
 namespace
 {
   void addLoopVectorizationHint(llvm::Instruction* instruction)
@@ -45,6 +46,7 @@ namespace
     instruction->setMetadata(llvm::StringRef("llvm.loop"), llvmLoopMetadata);
   }
 }
+#endif
 
 /************************************ | *************************************
 *                                                                           *
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -33,12 +33,17 @@ namespace
   {
     GenInfo* info    = gGenInfo;
     auto &C = info->module->getContext();
-    llvm::ArrayRef<llvm::Metadata*> loopVectorizeEnable({llvm::MDString::get(C, "loop.vectorize.enable"),
+    llvm::ArrayRef<llvm::Metadata*> loopVectorizeEnable({llvm::MDString::get(C, "llvm.loop.vectorize.enable"),
                                       llvm::ValueAsMetadata::get(llvm::ConstantInt::getTrue(C))}) ;
-    llvm::MDTuple* loopVectorizeMetadata = llvm::MDNode::get(C, loopVectorizeEnable);
+    llvm::MDTuple* loopVectorizeEnableMetadata = llvm::MDNode::get(C, loopVectorizeEnable);
 
-    llvm::ArrayRef<llvm::Metadata*> MDs({loopVectorizeMetadata });
-    llvm::MDTuple* llvmLoopMetadata = llvm::MDNode::getDistinct(C, MDs);
+    //llvm::MDNode* nullNode = nullptr;
+    llvm::TempMDTuple dummy = llvm::MDNode::getTemporary(C, {});
+    llvm::ArrayRef<llvm::Metadata*> MDs({dummy.get(), loopVectorizeEnableMetadata});
+    llvm::MDNode* llvmLoopMetadata = llvm::MDNode::get(C, MDs);
+
+    dummy->replaceAllUsesWith(llvmLoopMetadata);
+    //llvm::MDNode::deleteTemporary(dummy.get());
 
     instruction->setMetadata(llvm::StringRef("llvm.loop"), llvmLoopMetadata);
   }
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -31,20 +31,17 @@ namespace
 {
   void addLoopVectorizationHint(llvm::Instruction* instruction)
   {
-    GenInfo* info    = gGenInfo;
-    auto &C = info->module->getContext();
-    llvm::ArrayRef<llvm::Metadata*> loopVectorizeEnable({llvm::MDString::get(C, "llvm.loop.vectorize.enable"),
-                                      llvm::ValueAsMetadata::get(llvm::ConstantInt::getTrue(C))}) ;
-    llvm::MDTuple* loopVectorizeEnableMetadata = llvm::MDNode::get(C, loopVectorizeEnable);
+    GenInfo* info                  = gGenInfo;
+    auto &C                        = info->module->getContext();
+    llvm::ArrayRef<llvm::Metadata*> loopVectorizeEnable({llvm::MDString::get(C, "llvm.loop.vectorize.enable"), llvm::ValueAsMetadata::get(llvm::ConstantInt::getTrue(C))});
+    llvm::MDTuple*                  loopVectorizeEnableMetadata = llvm::MDNode::get(C, loopVectorizeEnable);
 
-    //llvm::MDNode* nullNode = nullptr;
-    llvm::TempMDTuple dummy = llvm::MDNode::getTemporary(C, {});
+    //Try to make cyclic reference to itself, since that's how llvm.loop parameters are supposed to be passed
+    llvm::TempMDTuple dummy        = llvm::MDNode::getTemporary(C, {});
     llvm::ArrayRef<llvm::Metadata*> MDs({dummy.get(), loopVectorizeEnableMetadata});
     llvm::MDNode* llvmLoopMetadata = llvm::MDNode::get(C, MDs);
 
     dummy->replaceAllUsesWith(llvmLoopMetadata);
-    //llvm::MDNode::deleteTemporary(dummy.get());
-
     instruction->setMetadata(llvm::StringRef("llvm.loop"), llvmLoopMetadata);
   }
 }
@@ -151,10 +148,8 @@ GenRet CForLoop::codegen()
                                                llvm::ConstantInt::get(condValue0->getType(), 0),
                                                FNAME("condition"));
 
-
     // Create the conditional branch
     info->builder->CreateCondBr(condValue0, blockStmtBody, blockStmtEnd);
-
 
     // Now add the body.
     func->getBasicBlockList().push_back(blockStmtBody);

### CODE REVIEW

**Comment by coodie** [comment]
I've managed to add llvm.mem.parallel_loop_access to code generation, I'm not sure If I did this properly. The best testcase would be to find a loop which wouldn't be vectorized without this metadata, but would be with them, however I'm struggling to find a case where a compiler-generated loop would be vectorized at all. There are few problems with this:

- Loop-vectorization pass depends on other passes. I think I managed to find reasonably small subset of passess that need to be given before the vectorization pass is done. I've created simple testcase in clang using `#pragma clang loop vectorize(enable)` with simple loop:

```
    #pragma clang loop vectorize(enable)(4)
    for(int i = 0; i < n; i++)
    {
        A[i] = a*B[i]+C[i];
    }
```

Which was vectorized, then I extracted commands given by clang to the llvm using "-mllvm -debug-pass=Arguments". And here is subset of commands which can be given to `opt` which vectorizes the loop:

`-licm -adce -simplifycfg -domtree -instcombine -barrier -float2int -domtree -loops -loop-simplify -lcssa -loop-rotate -branch-prob -block-freq -scalar-evolution -loop-accesses -loop-vectorize`. There is also -slp-vectorizer, but the given commands were enough to vectorize the loop, therefore I decided not to include it to avoid pollution.

These commands in case of chapel generated code, gave the warning:
```remark: <unknown>:0:0: loop not vectorized: could not determine number of loop iterations```

I looked at how clang compiles loops and how chpl compiles loops, and they seem to be different in this case. My current hypothesis is that loops cannot be vectorized because llvm can't see that loop is indeed a loop.

**Comment by coodie** [comment]
Here is example of IR llvm code generated by clang (the one in comment above):

```  
  store i32 0, i32* %i, align 4
  br label %15

; <label>:15:                                     ; preds = %34, %0
  %16 = load i32, i32* %i, align 4
  %17 = load i32, i32* %n, align 4
  %18 = icmp slt i32 %16, %17
  br i1 %18, label %19, label %37

; <label>:19:                                     ; preds = %15
  %20 = load i32, i32* %a, align 4
  %21 = load i32, i32* %i, align 4
  %22 = sext i32 %21 to i64
  %23 = getelementptr inbounds i32, i32* %11, i64 %22
  %24 = load i32, i32* %23, align 4
  %25 = mul nsw i32 %20, %24
  %26 = load i32, i32* %i, align 4
  %27 = sext i32 %26 to i64
  %28 = getelementptr inbounds i32, i32* %14, i64 %27
  %29 = load i32, i32* %28, align 4
  %30 = add nsw i32 %25, %29
  %31 = load i32, i32* %i, align 4
  %32 = sext i32 %31 to i64
  %33 = getelementptr inbounds i32, i32* %8, i64 %32
  store i32 %30, i32* %33, align 4
  br label %34

; <label>:34:                                     ; preds = %19
  %35 = load i32, i32* %i, align 4
  %36 = add nsw i32 %35, 1
  store i32 %36, i32* %i, align 4
  br label %15, !llvm.loop !1
...
  !1 = distinct !{!1, !2}
  !2 = !{!"llvm.loop.vectorize.enable", i1 true}
```

Here is the same code but ran with opt command with flags in above comment. It is easily noticeable that loop was vectorized:
```
  %11 = load i32, i32* %n, align 4
  %12 = load i32, i32* %a, align 4
  %13 = icmp slt i32 0, %11
  br i1 %13, label %.lr.ph3, label %59

.lr.ph3:                                          ; preds = %0
  %min.iters.check = icmp ult i32 %11, 8
  br i1 %min.iters.check, label %scalar.ph, label %min.iters.checked

min.iters.checked:                                ; preds = %.lr.ph3
  %n.mod.vf = urem i32 %11, 8
  %n.vec = sub i32 %11, %n.mod.vf
  %cmp.zero = icmp eq i32 %n.vec, 0
  br i1 %cmp.zero, label %scalar.ph, label %vector.ph

vector.ph:                                        ; preds = %min.iters.checked
  %broadcast.splatinsert7 = insertelement <4 x i32> undef, i32 %12, i32 0
  %broadcast.splat8 = shufflevector <4 x i32> %broadcast.splatinsert7, <4 x i32> undef, <4 x i32> zeroinitializer
  br label %vector.body

vector.body:                                      ; preds = %vector.body, %vector.ph
  %index = phi i32 [ 0, %vector.ph ], [ %index.next, %vector.body ]
  %broadcast.splatinsert = insertelement <4 x i32> undef, i32 %index, i32 0
  %broadcast.splat = shufflevector <4 x i32> %broadcast.splatinsert, <4 x i32> undef, <4 x i32> zeroinitializer
  %induction = add <4 x i32> %broadcast.splat, <i32 0, i32 1, i32 2, i32 3>
  %induction5 = add <4 x i32> %broadcast.splat, <i32 4, i32 5, i32 6, i32 7>
  %14 = add i32 %index, 0
  %15 = add i32 %index, 4
  %16 = sext i32 %14 to i64
  %17 = sext i32 %15 to i64
  %18 = getelementptr inbounds i32, i32* %8, i64 %16
  %19 = getelementptr inbounds i32, i32* %8, i64 %17
  %20 = getelementptr i32, i32* %18, i32 0
  %21 = bitcast i32* %20 to <4 x i32>*
  %wide.load = load <4 x i32>, <4 x i32>* %21, align 4
  %22 = getelementptr i32, i32* %18, i32 4
  %23 = bitcast i32* %22 to <4 x i32>*
  %wide.load6 = load <4 x i32>, <4 x i32>* %23, align 4
  %24 = mul nsw <4 x i32> %broadcast.splat8, %wide.load
  %25 = mul nsw <4 x i32> %broadcast.splat8, %wide.load6
  %26 = sext i32 %14 to i64
  %27 = sext i32 %15 to i64
  %28 = getelementptr inbounds i32, i32* %10, i64 %26
  %29 = getelementptr inbounds i32, i32* %10, i64 %27
  %30 = getelementptr i32, i32* %28, i32 0
  %31 = bitcast i32* %30 to <4 x i32>*
  %wide.load9 = load <4 x i32>, <4 x i32>* %31, align 4
  %32 = getelementptr i32, i32* %28, i32 4
  %33 = bitcast i32* %32 to <4 x i32>*
  %wide.load10 = load <4 x i32>, <4 x i32>* %33, align 4
  %34 = add nsw <4 x i32> %24, %wide.load9
  %35 = add nsw <4 x i32> %25, %wide.load10
  %36 = sext i32 %14 to i64
  %37 = sext i32 %15 to i64
  %38 = getelementptr inbounds i32, i32* %6, i64 %36
  %39 = getelementptr inbounds i32, i32* %6, i64 %37
  %40 = getelementptr i32, i32* %38, i32 0
  %41 = bitcast i32* %40 to <4 x i32>*
  store <4 x i32> %34, <4 x i32>* %41, align 4
  %42 = getelementptr i32, i32* %38, i32 4
  %43 = bitcast i32* %42 to <4 x i32>*
  store <4 x i32> %35, <4 x i32>* %43, align 4
  %index.next = add i32 %index, 8
  %44 = icmp eq i32 %index.next, %n.vec
  br i1 %44, label %middle.block, label %vector.body, !llvm.loop !1

middle.block:                                     ; preds = %vector.body
  %cmp.n = icmp eq i32 %11, %n.vec
  br i1 %cmp.n, label %._crit_edge4, label %scalar.ph

scalar.ph:                                        ; preds = %middle.block, %min.iters.checked, %.lr.ph3
  %bc.resume.val = phi i32 [ %n.vec, %middle.block ], [ 0, %.lr.ph3 ], [ 0, %min.iters.checked ]
  br label %45

; <label>:45:                                     ; preds = %scalar.ph, %45
  %46 = phi i32 [ %bc.resume.val, %scalar.ph ], [ %57, %45 ]
  %47 = sext i32 %46 to i64
  %48 = getelementptr inbounds i32, i32* %8, i64 %47
  %49 = load i32, i32* %48, align 4
  %50 = mul nsw i32 %12, %49
  %51 = sext i32 %46 to i64
  %52 = getelementptr inbounds i32, i32* %10, i64 %51
  %53 = load i32, i32* %52, align 4
  %54 = add nsw i32 %50, %53
  %55 = sext i32 %46 to i64
  %56 = getelementptr inbounds i32, i32* %6, i64 %55
  store i32 %54, i32* %56, align 4
  %57 = add nsw i32 %46, 1
  %58 = icmp slt i32 %57, %11
  br i1 %58, label %45, label %._crit_edge4, !llvm.loop !5

._crit_edge4:                                     ; preds = %middle.block, %45
  br label %59
...
  !1 = distinct !{!1, !2, !3, !4}
  !2 = !{!"llvm.loop.vectorize.enable", i1 true}
  !3 = !{!"llvm.loop.vectorize.width", i32 1}
  !4 = !{!"llvm.loop.interleave.count", i32 1}
  !5 = distinct !{!5, !2, !6, !3, !4}
  !6 = !{!"llvm.loop.unroll.runtime.disable"}
```

Note that clang generated "llvm.loop" metadata both in branch entering the loop body, and branch leaving loop body. I have experimented with adding this metadata to generated code and this also didn't help. I have also removed both of the metadata (also checked the case when only one was left, code was vectorized anyway) from upper code and loop was vectorized by opt anyway (!) - yes, there was no vectorization hint on this loop and clang still managed to vectorize it.

**Comment by coodie** [comment]
Here is code snippet which I've tried to get vectorized:
```
proc mainTestFunctionForall()
{

  var A : [0..100] real;
  var B : [0..100] real;
  var C : [0..100] real;

  for i in 0..100 do
  {
    B[i] = read(real);
    C[i] = read(real);
  }

  var alpha : real = read(real);

  A = B * alpha + C; //implicit forall

  for i in 0..100 do
    writeln(A[i]);
}

mainTestFunctionForall();

```

Idea for this test comes from STREAMS as recommended by @mppf. I compiled this with `--vectorize --llvm --llvm-wide-opt --fast`. Note that I've put the loop into function and changed types from int to real so that finding loop in LLVM code is easier (it's quite a challenge to find it in ~35000 lines of code). 

I took a quick look at code generated by chpl, and there are a lot of loops associated with distributing code for locales. This one looks like the one doing the real calculation job (the one we want to be vectorized the most), and I know that because no other loop has "fmul" operation (I've added "B[i]*alpha" and "+ C[i]" comments).

```
mainTestFunctionForall_chpl_16blk_body_:          ; preds = %mainTestFunctionForall_chpl_16blk_body_, %mainTestFunctionForall_chpl_15blk_body_
  %_ic__F9_i_chpl3.0 = phi i64 [ %370, %mainTestFunctionForall_chpl_15blk_body_ ], [ %401, %mainTestFunctionForall_chpl_16blk_body_ ]
  %_ic__F9_i_chpl2.0 = phi i64 [ %322, %mainTestFunctionForall_chpl_15blk_body_ ], [ %400, %mainTestFunctionForall_chpl_16blk_body_ ]
  %_ic__F9_i_chpl.0 = phi i64 [ %274, %mainTestFunctionForall_chpl_15blk_body_ ], [ %399, %mainTestFunctionForall_chpl_16blk_body_ ]
  %392 = getelementptr inbounds double, double addrspace(100)* %385, i64 %_ic__F9_i_chpl3.0
  %393 = getelementptr inbounds double, double addrspace(100)* %387, i64 %_ic__F9_i_chpl2.0
  %394 = getelementptr inbounds double, double addrspace(100)* %389, i64 %_ic__F9_i_chpl.0
  %395 = load double, double addrspace(100)* %393, !llvm.mem.parallel_loop_access !26
  %396 = fmul double %395, %224 // B[i]*alpha
  %397 = load double, double addrspace(100)* %392, !llvm.mem.parallel_loop_access !26
  %398 = fadd double %396, %397 // + C[i]
  store double %398, double addrspace(100)* %394, !llvm.mem.parallel_loop_access !26
  %399 = add i64 %_ic__F9_i_chpl.0, 1
  %400 = add i64 %_ic__F9_i_chpl2.0, 1
  %401 = add i64 %_ic__F9_i_chpl3.0, 1
  %402 = icmp sle i64 %399, %277
  %403 = zext i1 %402 to i64
  br i1 %402, label %mainTestFunctionForall_chpl_16blk_body_, label %mainTestFunctionForall_chpl_14cond_end_, !llvm.loop !26
```

I've disabled vectorization for other loops manually, so I have clear output for that loop. I've run opt on the loop and here is the warning message it gives:

```
remark: <unknown>:0:0: loop not vectorized: could not determine number of loop iterations
```

That is exactly the same message printed for every other loop. At first sight this loop is greatly different from what clang generates, looks like there are three indices:

_ic__F9_i_chpl.0
_ic__F9_i_chpl2.0
_ic__F9_i_chpl3.0

Let's call them i1, i2, i3 and the loop looks like it is compiled from:

```

//some initialization
while( n > i3)
{
  A[i1] = B[i2]*alpha+C[i3];
  i1++;
  i2++;
  i3++;
}

```

- Maybe this would be better if there was one index inside the loop. 
- I'm not sure if the parallel_loop_access are correctly added here. I'll keep on observing and investigating.

**Comment by mppf** [comment]
hi @coodie - thanks for the note. I think there are two things possible here:
 1) Our "CForLoops" are generating sufficiently obtuse LLVM IR so as to confuse the LLVM optimizations
 2) Our vectorization hints / parallel loop access hints are wrong.

I think you can try manually modifying some LLVM IR to try to figure out which of these two cases is applying.

Also, re. your clang experiments, I think we need to get an LLVM IR dump just *before* vectorization occurs so we can see what we should be generating. Can you get clang to give you an IR dump of the LLVM IR generated by the C frontend, with all the parallel loop access hints in there?

**Comment by coodie** [comment]
@mppf 
Isn't running "clang++ main.cpp -c -emit-llvm" and then running "llvm-dis" on clang generated code enough? It does no optimization on that code (at least not some obvious optimizations). Also, note the EDIT I just did.

**Comment by mppf** [comment]
@coodie - re. clang options - maybe, but I was expecting to see parallel_loop_access metadata and do not.

Anyway, it sounds like the main issue in this example is how the CForLoop bound is written. Thanks for figuring out that changing the icmp gets it to work!

When we added CForLoop, I wanted to make it use the lowest-level LLVM IR possible - phi nodes and SSA loop variables - but ran into challenges doing that b/c of how our bounds checking works (but bounds checking is turned off in --fast).

Anyway, I think it's time to think about another example. Consider this:

``` chapel
config const n = 1000;
var A:[1..n] int;
var B:[1..n] int;

for i in 1..n {
  B[i] = i;
}

for i in vectorizeOnly(1..n) {
  A[i] = 2*B[i];
}

writeln("Sum of A is ", + reduce A);
```

Here, the vectorizeOnly iterator should generate a CForLoop with vectorization enabled (from the Chapel compiler's perspective). At the same time, this is a simpler loop than the one you have been working with. (The promoted one you have been working with uses zippered leader-follower iterators and creates tasks). Does the simpler one vectorize? Can you see what needs to change about it to get it to vectorize? It might be easier to see how the LLVM IR codegen needs to be modified to get this simpler loop to vectorize.

coodie:
TLDR;

Still no vectorization occured.

Hello, I changed your example to simplify it even further (and changed from real to int, I didn't check it on integers, but I don't think that matters):

```
proc testMain()
{
  const n = 1000;
  var A:[1..n] real;
  var B:[1..n] real;

  for i in 1..n
  {
    B[i] = i;
  }

  for i in vectorizeOnly(1..n)
  {
    A[i] = 2*B[i];
  }

  writeln("Sum of A is ", + reduce A);
}
testMain();

```

Here is non-optimized LLVM IR generated by Chapel:

```
%range_int64_t_bounded_F = type { i64, i64, i64, i64, i8 }
...
testMain_chpl_3blk_end_:                          ; preds = %testMain_chpl_3blk_body_, %entry
  call void @chpl_build_bounded_range(i64 1, i64 1000, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5)
  %90 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 0
  %91 = load i64, i64* %90
  %92 = insertvalue %range_int64_t_bounded_F undef, i64 %91, 0
  %93 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 1
  %94 = load i64, i64* %93
  %95 = insertvalue %range_int64_t_bounded_F %92, i64 %94, 1
  %96 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 2
  %97 = load i64, i64* %96
  %98 = insertvalue %range_int64_t_bounded_F %95, i64 %97, 2
  %99 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 3
  %100 = load i64, i64* %99
  %101 = insertvalue %range_int64_t_bounded_F %98, i64 %100, 3
  %102 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 4
  %103 = load i8, i8* %102
  %104 = insertvalue %range_int64_t_bounded_F %101, i8 %103, 4
  %105 = load %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)*, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)** %47
  %106 = getelementptr inbounds %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)* %105, i32 0, i32 8
  %107 = load double addrspace(100)*, double addrspace(100)* addrspace(100)* %106
  %108 = load %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)*, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)** %81
  %109 = getelementptr inbounds %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)* %108, i32 0, i32 8
  %110 = load double addrspace(100)*, double addrspace(100)* addrspace(100)* %109
  %111 = icmp sle i64 %91, %94
  br i1 %111, label %testMain_chpl_4blk_body_, label %testMain_chpl_4blk_end_

testMain_chpl_4blk_body_:                         ; preds = %testMain_chpl_4blk_body_, %testMain_chpl_3blk_end_
  %i_chpl2.0 = phi i64 [ %91, %testMain_chpl_3blk_end_ ], [ %116, %testMain_chpl_4blk_body_ ]
  %112 = getelementptr inbounds double, double addrspace(100)* %107, i64 %i_chpl2.0
  %113 = getelementptr inbounds double, double addrspace(100)* %110, i64 %i_chpl2.0
  %114 = load double, double addrspace(100)* %113, !llvm.mem.parallel_loop_access !25
  %115 = fmul double 2.000000e+00, %114
  store double %115, double addrspace(100)* %112, !llvm.mem.parallel_loop_access !25
  %116 = add i64 %i_chpl2.0, 1
  %117 = icmp sle i64 %116, %94
  br i1 %117, label %testMain_chpl_4blk_body_, label %testMain_chpl_4blk_end_, !llvm.loop !25
```

Let's look at what happens in above code briefly:

1. 
 ```
call void @chpl_build_bounded_range(i64 1, i64 1000, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5)
```
This line build range (1,1000) and stores it in `%_autoCopy_tmp__chpl5`
2.  
``` 
%90 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 0
%91 = load i64, i64* %90
%93 = getelementptr inbounds %range_int64_t_bounded_F, %range_int64_t_bounded_F* %_autoCopy_tmp__chpl5, i32 0, i32 1
%94 = load i64, i64* %93
```
These lines store start and end indices of the loop to %91 and %94
Note that I have removed `insertvalue` instruction, because this builds structure which isn't used later anywhere. 
3.
```
  %105 = load %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)*, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)** %47
  %106 = getelementptr inbounds %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)* %105, i32 0, i32 8
  %107 = load double addrspace(100)*, double addrspace(100)* addrspace(100)* %106
  %108 = load %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)*, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)** %81
  %109 = getelementptr inbounds %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object, %chpl_DefaultRectangularArr__real64_1_int64_t_F_int64_t_object addrspace(100)* %108, i32 0, i32 8
  %110 = load double addrspace(100)*, double addrspace(100)* addrspace(100)* %109
```

These lines store pointer to arrays A and B in %107 and %110.

4.
```
testMain_chpl_4blk_body_:                         ; preds = %testMain_chpl_4blk_body_, %testMain_chpl_3blk_end_
  %i_chpl2.0 = phi i64 [ %91, %testMain_chpl_3blk_end_ ], [ %116, %testMain_chpl_4blk_body_ ]
  %112 = getelementptr inbounds double, double addrspace(100)* %107, i64 %i_chpl2.0
  %113 = getelementptr inbounds double, double addrspace(100)* %110, i64 %i_chpl2.0
  %114 = load double, double addrspace(100)* %113, !llvm.mem.parallel_loop_access !25
  %115 = fmul double 2.000000e+00, %114
  store double %115, double addrspace(100)* %112, !llvm.mem.parallel_loop_access !25
  %116 = add i64 %i_chpl2.0, 1
  %117 = icmp sle i64 %116, %94
  br i1 %117, label %testMain_chpl_4blk_body_, label %testMain_chpl_4blk_end_, !llvm.loop !25
```

That's the body of our loop. Someone who is familiar with LLVM IR will see that the loop is `A[i] = B[i]*2`.

This looks like vectorizer is unable to determine number of loops. Here is how I managed to get the loop to vectorize:

- Change `%117 = icmp sle i64 %116, %94` to  `%117 = icmp sle i64 %116, 1000` (exactly the same thing I did before).
- Change `%94 = load i64, i64* %93` to `%94 = add i64 0, 1000` (I have just assigned 1000 to %94 explicitly).

Here is what did not get the loop to vectorize:
- Change `%94 = load i64, i64* %93` to `%94 = call i64 gimme1000()` (Assign 1000 to %94 from function call). Where gimme1000 is defined as:
```
define internal i64 @gimme1000() {
entry:
  ret i64 1000
}
```
- Change `%94 = load i64, i64* %93` to `%94 = add i64 0, %end`. Where %end is argument of function which contains the loop body.


I did some experiments on C generated code to see how I can kill vectorization by playing with end indice assignment. I have noticed that loop always gets vectorized if the ending index is assigned to variable before loop. Specifically:

```
int N = function(a,b,c);
for(int i = 0; i < N; i++)
{
   A[i] = B[i]*2;
}
```  

The above always gets vectorized.

```
for(int i = 0; i < function(a,b,c); i++)
{
   A[i] = B[i]*2;
}
```
The above never gets vectorized, because end index is evaluated every time inside the loop and LLVM does not analyze whether function(a,b,c) is pure or not. 

I did few more experiments where N didn't change while loop was being executed, but was assigned before in many ways (like load by user, left undefined, was constant etc.), and in all these cases the loop got vectorized.

According to results of above experiments, and how the loop is compiled vectorization should occur in case of Chapel (end index is assigned before the loop), but somehow it does not. Loop only gets vectorized when exact number of iterations is known. This makes me think that we compile loop in such a way that vectorizer only manages to notice the number of iterations in very few ways. 

Next step I think would be to understand when loop might get vectorized (probably by looking at clang generated code) and change compilation.

**Comment by coodie** [comment]
When it comes to clang generated code there is no usage of phi nodes in non-optimized code (while Chapel uses them heavily).

I think I haven't described how I check when clang vectorizes loops:

I write in file `main.cpp` valid C++ program which contains loop with vectorize enable pragma like this:
```
    #pragma clang loop vectorize(enable)    
    for(int i = 0; i < n; i++)
    {
        A[i] = a*B[i]+C[i];
    }
```
One has to be careful when writing the loop, because loop might get treated as dead code while running `opt` as described:

1. I run: clang++ main.cpp -c -emit-llvm.
2. Above command gives me `main.bc` file, I run `llvm-dis main.bc` which gives me `main.ll`.
3. Then I run `opt -licm -adce -simplifycfg -domtree -instcombine -barrier -float2int -domtree -loops -loop-simplify -lcssa -loop-rotate -branch-prob -block-freq -scalar-evolution -loop-accesses -loop-vectorize main.ll -S > main-opt.ll`, and if the loop isn't vectorized the warning is printed. I usualy double check optimized code manually to make sure everything is fine.

Note that I run optimizations on code generated by clang. I provided example of code generated for the loop given above, and there is no `llvm.mem.parallel_loop_access` metadata. Note that without pragma, loop also gets vectorized meaning that `llvm.vectorize.enable` is useless - optimizer can spot vectorizable loops without this metadata.

**Comment by coodie** [comment]
TLDR;

I managed to get loop vectorized incredibly easily.

I was playing with clang options and I've run
`
opt -licm -adce -simplifycfg -domtree -instcombine -barrier -float2int -domtree -loops -loop-simplify -lcssa -loop-rotate -branch-prob -block-freq -scalar-evolution -loop-accesses main.ll -S > main-opt.ll
`
Note that upper command does all transformations before loop is vectorized. I wanted to see how should the loop look like right before it is vectorized (since all of these transformations are supposed to transform the loop to be understood better by vectorizer).

On llvm IR of clang generated code:

```
void func(int n, float *A, float *B)
{
    #pragma clang loop vectorize(enable)
    for(int i = 0; i < n; i++)
    {
        A[i] = B[i]*2;
    }
}
```

Giving really nice and concise LLVM version of the following loop:

```
  %1 = icmp slt i32 0, %n
  br i1 %1, label %.lr.ph, label %12

.lr.ph:                                           ; preds = %0
  br label %2

; <label>:2:                                      ; preds = %.lr.ph, %2
  %3 = phi i32 [ 0, %.lr.ph ], [ %10, %2 ]
  %4 = sext i32 %3 to i64
  %5 = getelementptr inbounds float, float* %B, i64 %4
  %6 = load float, float* %5, align 4
  %7 = fmul float %6, 2.000000e+00
  %8 = sext i32 %3 to i64
  %9 = getelementptr inbounds float, float* %A, i64 %8
  store float %7, float* %9, align 4
  %10 = add nsw i32 %3, 1
  %11 = icmp slt i32 %10, %n
  br i1 %11, label %2, label %._crit_edge, !llvm.loop !1
```

For the sake of completeness here is chapel generated version of similar loop (B[i]*2 with vectorizeOnly iterator mentioned in previous comments):

```
testMain_chpl_4blk_body_:                         ; preds = %testMain_chpl_4blk_body_, %testMain_chpl_3blk_end_
  %i_chpl2.0 = phi i64 [ %91, %testMain_chpl_3blk_end_ ], [ %116, %testMain_chpl_4blk_body_ ]
  %112 = getelementptr inbounds double, double addrspace(100)* %107, i64 %i_chpl2.0
  %113 = getelementptr inbounds double, double addrspace(100)* %110, i64 %i_chpl2.0
  %114 = load double, double addrspace(100)* %113, !llvm.mem.parallel_loop_access !25
  %115 = fmul double 2.000000e+00, %114
  store double %115, double addrspace(100)* %112, !llvm.mem.parallel_loop_access !25
  %116 = add i64 %i_chpl2.0, 1
  %117 = icmp sle i64 %116, %94
  br i1 %117, label %testMain_chpl_4blk_body_, label %testMain_chpl_4blk_end_, !llvm.loop !25
```

I noticed that chapel version is basically the same as clang version, but for some reason it doesn't get vectorized. 

I've had some experience from previous experiments and noticed that the only reason why this loop wasn't vectorized was because vectorizer couldn't calculate number of iterations, the only place which impacts number of iterations is actually `add` instruction. In case of clang generated code it has extra `nsw` argument, and chapel doesn't. So what I did is I've manually added `nsw` argument to the chapel code as such:

```
testMain_chpl_4blk_body_:                         ; preds = %testMain_chpl_4blk_body_, %testMain_chpl_3blk_end_
  %i_chpl2.0 = phi i64 [ %91, %testMain_chpl_3blk_end_ ], [ %116, %testMain_chpl_4blk_body_ ]
  %112 = getelementptr inbounds double, double addrspace(100)* %107, i64 %i_chpl2.0
  %113 = getelementptr inbounds double, double addrspace(100)* %110, i64 %i_chpl2.0
  %114 = load double, double addrspace(100)* %113, !llvm.mem.parallel_loop_access !25
  %115 = fmul double 2.000000e+00, %114
  store double %115, double addrspace(100)* %112, !llvm.mem.parallel_loop_access !25
  %116 = add nsw i64 %i_chpl2.0, 1
  %117 = icmp sle i64 %116, %94
  br i1 %117, label %testMain_chpl_4blk_body_, label %testMain_chpl_4blk_end_, !llvm.loop !25
``` 

Next step would be to investigate a little bit more about add instruction and add proper arguments of it to code generation.

File: compiler/codegen/expr.cpp [hunk 1]
```diff
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, true);
+        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, a_signed or b_signed);
```
mppf: I thought it would be values.isSigned (which indicates whether whatever we promoted to is signed) and I don't think 'or' works in C++ (unless it is some new standard thing I've never seen?)

coodie: `or` has been in C++ standard since 1998 I believe

And changing to values.isSigned won't probably change anything, but is more readable.

mppf: Huh. I learned something. Anyway, I think values.isSigned is better here. Thanks!

mppf:
@coodie - I'm closing this PR since it's not ready to merge & we'd like to limit the number of standing PRs. You should not interpret closing it as any judgement on your work here. Anyway, at what point will the code changes you made here be "ready" to be reviewed and merged? Maybe move them to another PR and track the progress in an issue?

Thanks.

mppf:
After PR #6465, I think you can do
`chpl test.chpl --mllvm -vectorize-scev-check-threshold=0 --mllvm -runtime-memory-check-threshold=0` but I havn't tried it yet. This should disable the run-time check for LLVM's loop vectorizer and so only vectorize if the LLVM opt knows it is safe.

PR #5789, opened 2017-03-22
Fix some memory leaks in ArrayViewRankChange

benharsh:
This commit addresses two problems:

1) old domains in 'dsiSetIndices' were not free'd before being
overwritten
Solution: call dsiDestroyDom() to free the old domains (if they are not
null, as they might be in a freshly-created rank-change domain).

2) too many deep-copies of the downdist _distribution
Solution: Store a pointer to the downward distribution's PID and class
to avoid excessive copying without freeing.

Testing:
- [x] full local
- [x] full gasnet
- [x] valgrind?

### DIFF
--- a/modules/internal/ArrayViewRankChange.chpl
+++ b/modules/internal/ArrayViewRankChange.chpl
@@ -49,7 +49,8 @@ module ArrayViewRankChange {
   class ArrayViewRankChangeDist: BaseDist {
     // a pointer down to the distribution that this class is creating
     // lower-dimensional views of
-    const downdist;
+    var downDistPid:int;
+    var downDistInst;
     
     // These two fields represent whether or not each dimension was
     // collapsed as part of the rank-change; and if so, what the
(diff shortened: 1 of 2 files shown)


benharsh:
@bradcray : here you go!

This passed the memory/sungeun/refCount tests. I reran with --memLeaks for the regressing tests and they are all now as good or better than before (except HPL 2012).

bradcray:
Just to understand your "as good or better" characterization:  Are tests like the following back down to zero leaks?  (which is where I think they were prior to the rank-change merge):

````
     7131800  cholesky.chpl
     7130112  covariance.chpl
     7022080  correlation.chpl
     6914048  2mm.chpl
     6914048  syr2k.chpl
     6806016  lu.chpl
````

benharsh:
Yes, those are all back to zero.

bradcray:
This looks great, thanks Ben!

Based on my memory leak chasings, I would probably suggest running valgrind at least on the memory/sungeun/refCount tests and possibly on some of the tests that you fixed leaks in.  OTOH, a full valgrind test doesn't take overlong (a bit over an hour I think) and might be worthwhile for peace of mind.  I've got a free workspace I could fire off one of these test configurations on if you'd like.

PR #5790, opened 2017-03-22
"no doc" the MPI module deinit()

ben-albrecht:
`chpldoc` has not yet been updated to handle module-level `deinit` so I am going to no-doc the function. I don't think this function should really be considered part of the public interface anyway.

This error was introduced in #5769 

Also, no-doc'd 2 private module-level variables.


ben-albrecht:
Documented chpldoc bug in #5791

PR #5800, opened 2017-03-23
Update specification about "ref-pair" to include 3-way

### LINKED ISSUE #5590: spec update for 3-way ref pair for 1.15
After #5561, we need to update the spec to show 3-way ref pair.

mppf:
Fixes #5590.

Removes uses of the term "ref-pair".
Moves most of the detailed description to the Function Resolution
section.
Includes description of choosing between const ref and value versions.

### DIFF
--- a/spec/Functions.tex
+++ b/spec/Functions.tex
@@ -786,45 +786,26 @@ The \chpl{const ref} return intent is also available. It is a restricted
 form of the \chpl{ref} return intent. Calls to functions marked with
 the \chpl{const ref} return intent are not lvalue expressions.
 
-\subsection{Choosing a Ref Return Intent Function Based on Calling Context}
-\label{Ref_Pair}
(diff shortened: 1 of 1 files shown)


spec/Functions.tex
```diff
@@ -786,45 +786,26 @@ \subsection{The Const Ref Return Intent}
...
-% TODO - this section, or a significant portion of it,
-% should move to the function resolution section.
+\subsection{Return Intent Overloads}
+\label{Return_Intent_Overloads}
+\index{functions!return intent overloads}
+
+In some situations, it is useful to choose the function called based upon
+how the returned value is used.  In particular, suppose that there are
+two functions that have the same formal arguments and differ only in
+their return intent. Normally, such a situation will cause an error since
```
benharsh: It seems like we could say something better than "Normally" here. Maybe something like "In other languages, " ?

PR #5801, opened 2017-03-23
Support where clauses on initializers

lydia-duncan:
Prior to this change, we were failing to accept initializers that defined
where clauses because the class allocator did not recognize and copy the where
clause on the initializer with which it was paired.  This change copies the
where clause and ensures that it refers to the arguments in the _new method
rather than the arguments in the original initializer.

It also adds four new tests of initializers on classes with where clauses, three of which
failed prior to this change.  The first case exercises using a where clause when
the argument signature is fully specified and otherwise identical.  The second
case exercises using a where clause when the argument signature of the two
initializers is generic but would otherwise lead to ambiguity during resolution.
The third case exercises using a where clause when the arguments are fully
specified and so the where clause is redundant.  The fourth case exercises
using a where clause to eliminate other calls that the generic argument would
otherwise allow (in this case, reals).

It also adds four new passing tests of initializers on records with where clauses,
which I believe worked before my change as well, and a future version to track
an unfortunate error message.

Testing over the initializers directory revealed no new failures.  Sanity checked
over full std/


test/classes/initializers/where-clause4.good
```diff
@@ -0,0 +1,2 @@
+where-clause4.chpl:12: error: unresolved call '_new(type Foo, 3.4)'
```
noakesmichael: It's tempting to say that there should also be a "future" version of this test to reflect the
notion that this error message isn't all that great.

lydia-duncan: Yeah.  I'm also tracking the fact that we get a whole lot of extra output when the problem is with the init function itself - this is likely to impact us increasingly as we switch library types to use the new style initializers, so we'll probably want to clamp down on what we output in those cases.

PR #5802, opened 2017-03-23
Array with unsigned domain API test

benharsh:
Adds tests for slices, rank-changes, and reindexes on an array
declared over a domain of unsigned ints.

This test exposed some minor bugs in our internal modules, which
are also fixed by this PR.

Testing:
- [x] full local
- [x] full gasnet

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -995,6 +995,7 @@ module DefaultRectangular {
   //
   proc _remoteAccessData.toRankChange(newDom, cd, idx) {
     compilerAssert(this.rank == idx.size && this.rank != newDom.rank);
+    type idxSignedType = chpl__signedType(idxType);
 
     // Unconditionally sets 'blkChanged'
     //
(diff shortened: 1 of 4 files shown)


benharsh:
@bradcray : Here's the unsigned API test we discussed.

I have one ready for Block, but it currently fails due to the reindexed domain not being distributed.

benharsh:
@ben-albrecht : heads up, there are minor changes to ``domain.shape`` in this PR.

modules/internal/DefaultRectangular.chpl
```diff
@@ -1014,7 +1015,8 @@ module DefaultRectangular {
     for param j in 1..idx.size {
       if !collapsedDims(j) {
         rad.off(curDim) = newDom.dsiDim(curDim).low;
-        rad.origin     += this.blk(j) * (rad.off(curDim) - this.off(j)) / this.str(j);
+        const off       = (rad.off(curDim) - this.off(j)):idxSignedType;
+        rad.origin     += ((this.blk(j):idxSignedType) * off / this.str(j)):idxType;
```
bradcray: I'm not understanding the need to cast 'blk' to a signed type here... is it simply to avoid uint-int math?

benharsh: Unfortunately, yes.

bradcray:
This looks fine to me, thanks!

On the block variation, I'd suggest either having the .good file show the current behavior and adding a second block-reindex-only test as a future to capture the need to add locality.  I'm skeptical that the locality preservation of reindex is going to make it in at this point, though it depends a bit on how quickly I can get through CHANGES...

PR #4745, opened 2016-10-11
First step to transition to qualified references

benharsh:
Co-authored with @mppf 
Reviewed by @vasslitvinov , @noakesmichael , @ronawho 

This is the first step towards a representation of references that does not
rely on embedding ref-ness into a symbol's Chapel type. Instead, symbols have
a QualifiedType that represents a varaible 'kind' like 'const', 'var', or
'ref'.

While there are a number of entries in the 'Qualifier' enum, only a subset are
currently relied upon:
- QUAL_UNKNOWN
- QUAL_REF
- QUAL_CONST_REF
- QUAL_WIDE_REF
- QUAL_VAL

Over time others may be introduced and utilized properly, but for this
initial step it was useful to keep the changes as simple as possible.
I have added a number of 'BHARSH TODO' comments to keep track
of future work required to convert the compiler entirely over to using
qualified references. I plan to continue working on these TODOs for
the upcoming release.

New in the compiler
-------------------

New utility functions:
- isMoveOrAssign
- isDerefMove (moving from a reference into a var)

Every BaseAST now supports the following queries that return a boolean:
- isRef()
- isWideRef()
- isRefOrWideRef()

These methods rely on the QualifiedType of symbol or expression they
represent. For now, they will also consider the Chapel type and look for
FLAG_REF and FLAG_WIDE_REF when appropriate. Eventually they will only
consider the Qualifier.

This change also introduces the primitive PRIM_SET_REFERENCE, which will
ultimately replace PRIM_ADDR_OF in the QualifiedType world.

The pass 'resolveIntents' is currently also responsible for setting the
correct Qualifier for variables.

Codegen
-------

While it would have been nice to use QualifiedTypes for the entirety of
codegen, in the interest of time it was most expedient to change symbols with
a "ref" Qualifier to have a reference type at the start of codegen. This made
it easy to get LLVM operational. That said, much of codegen has been changed
to use 'isRef', or something similar, instead of directly looking at flags on
the type's symbol.

There are many changes to codegen, most of them adding dereferences or
addrOfs when necessary. With QualifiedTypes, the compiler is less likely to
introduce reference or dereference temporaries, and codegen will pick up the
slack. For example, one can now dereference simply by PRIM_MOVE-ing into a
non-reference variable. A number of issues with codegen were exposed because
of arguments with a "ref" intent that did not have a reference type, and
therefore were not correctly identified as references by codegen.

Primitive Calls
---------------

All type functions for primitives now return QualifiedTypes. 'typeInfo'
[...]

PR #5803, opened 2017-03-23
Improve late const checking: check _this

mppf:
The main purpose of this PR is to remove the workaround in late const checking that disables const checking on `this` arguments. However, in order to get everything to work, I needed to take several other steps:

* In blankIntentForThisArg, handle INTENT_CONST like INTENT_THIS, continuing PR #4745. See commit 5361ec4.
* Improve late const checking error messages. Removes duplicate lines for users, prints type of formal (important to know to debug some of the cases). Also prints out information about shadow variables at least in some cases.
* fixes many lvalue issues in modules and tests. The biggest change here is making atomics work by `const` for read/peek/waitFor/writeThis. That is necessary since atomics have a default intent of `ref` and yet we'd like assignment to work, for example, with a `const` RHS.
* disable late const checking for now for tuples-of-tuples, calls to promotion wrappers
* using a pragma, disable late const checking of _getIterator and related calls

Testing:

* full local testing
* hellos with --baseline
* hellos with --verify
* hellos with --baseline --verify
* hellos with GASNet
* hellos with ugni
* hellos with numa


mppf:
@benharsh - re const-checking on promotion, it's disabled (so not really correct but shouldn't generate errors). I kept changes to a few tests (PrintComms.chpl) to limit testing failures if/when the problem comes up again, since what that function is doing doesn't really need it to be const. But I added a new test that tracks it explicitly.

benharsh:
Sounds good to me, thanks!

PR #5805, opened 2017-03-23
Void vars primer

daviditen:
Add a primer for void variables and fields

Add a primer demonstrating void variables and void fields.  Declare some
void variables, create a record with void fields, explain what they are
and how they are useful.


daviditen:
Thanks @benharsh !  The documentation for it builds and looks good.

PR #5808, opened 2017-03-23
Array API tests for Block and fix related bugs

benharsh:
Tests the normal and sliced Block-distributed array API. A future is added
to track bugs in the reindexed ArrayView.

This commit also fixes ``localSubdomain`` for ArrayViewRankChange. First
we query the downward domain's local subdomain. If the collapsed dimensions'
index is not a member of the downward subdomain's dimension, then
this locale does not have local subdomains from the rank-changes point
of view. In that case we return an empty domain. Otherwise we collapse the
appropriate dimensions of the downward subdomain and return.

Also implements ``BlockArr.dsiSerialRead`` using existing helper function.

Testing:
- [x] full local
- [x] full gasnet


benharsh:
@bradcray : how does this look?

benharsh:
@bradcray : I have fixed 'localSubdomain' for rank-changes (at least for Block). I have also added a dedicated test against this functionality.

test/arrays/userAPI/arrayAPItest.chpl
```diff
@@ -57,10 +57,11 @@ proc testArrayAPI2D(lbl, X: [], sliceDom, reindexDom) {
   writeln("X is:\n", X);
   writeln();
 
-  // Test read access via tuples and varargs
-  writeln("low element is: ", X.localAccess[X.domain.alignedLow]);
+  const domForLowHigh = if X.hasSingleLocalSubdomain() then X.localSubdomain() else X.domain;
+  // Test local read access via tuples and varargs
+  writeln("low element is: ", X.localAccess[domForLowHigh.alignedLow]);
```
bradcray: It seems as though it could be that, for a given array+domain, the localSubdomain could be empty or X.domain's low could not be local to the current locale.  Are you thinking that because we control the clients of this helper module, we'll just be sure not to call it in such cases?  (which is fine with me).  So, for example, with Block, all locales will have a piece and with DR we'll just be running with one locale?

test/arrays/userAPI/blockOps2D.comm-none.good
```diff
@@ -0,0 +1,414 @@
...
+idxType is: int(64)
+rank is: 2
+
+size is: 100
+numElements is: 100
+shape is: (10, 10)
+
+X is:
+1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0
+11.0 12.0 13.0 14.0 15.0 16.0 0.0 0.0 0.0 0.0
```
bradcray: Huh... is the implication here that if a file doesn't have enough values to fill an array, we just stop reading and leave the rest uninitialized?  I'd have expected this to result in an error...  It'd also be fine with me to beef up the size of the file such that we can put true interesting values into every location (in which case I think the current cases will just stop reading early, which seems fine.

modules/internal/ArrayViewRankChange.chpl
```diff
@@ -357,7 +357,18 @@ module ArrayViewRankChange {
...
+      const dims = downDom.dsiLocalSubdomain().dims();
+      const empty : domain(rank, idxType, chpl__anyStridable(dims));
+
+      // If the rank-changed dimension's index is not a member of the range
+      // in the same dimension of 'dims', then this locale does not have a
+      // local subdomain.
+      for param d in 1..dims.size {
+        if collapsedDim(d) && !dims(d).member(idx(d)) then
+          return empty;
+      }
```
bradcray: Thinking through this: for a ``{3, ..}`` rank-change slicing dom on a block-distributed array/domain, the down domain would be ``{3..3, ..}`` in which case the processors owning column 3 would see ``{3..3, <whatever>}`` as their local down-subdomain and the others would see ``{1..0, <whatever>}`` as theirs.  So rather than doing a membership check, could we simply see if dims(d) is empty?  (though I guess the two are arguably equivalent, this way just feels slightly more convoluted... I'd guess it's asking "Is 3 a member of this empty range?"

bradcray:
Sorry for the late review -- I didn't have internet access for very long last night.  Unless my comments cause you to want to make any changes, this looks good to me, thanks!

test/arrays/userAPI/blockOps2D.comm-none.good
benharsh: Hmm, I hadn't noticed that. I'd be inclined to hold off on such a change (for another PR) as it would impact every test in this directory.

As far as 'solutions' go, perhaps we could create a file in-memory and generate random numbers with a seed?

modules/internal/ArrayViewRankChange.chpl
benharsh: Yup, you're right. I just pushed a commit that uses 'isEmpty'.

test/arrays/userAPI/arrayAPItest.chpl
benharsh: Yeah, I'm just relying on:
- us being the only ones to call this arrayAPItest
- there being some indices owned by the current locale

benharsh:
How do things look now?

benharsh:
I'm going to go ahead and merge now so that we get this through a night of testing.

test/arrays/userAPI/blockOps2D.comm-none.good
bradcray: I'd be surprised if adding more numbers to the file broke the existing tests...  Only the new block cases read more than 16 elements, don't they?

benharsh: I'm not worried about adding more to the file, I was just thinking about a way to avoid needing a separate file at all.

bradcray: I don't think I understand your "impact every test in this directory" comment then...

benharsh: Huh, fair point. Not sure what I was thinking...

PR #5810, opened 2017-03-23
Don't mess up schedaffinity when touching desired heap for locality.

gbtitus:
The code for touching desired heap pages into existence on the "right"
NUMA domains was careful to first record the thread's (process's) NUMA
locality and then restore it afterward.  Unfortunately, the function
used to get that locality was one designed to return the specific NUMA
domain associated with the thread and, as such, if the thread had
affinity to multiple domains the function reduced this to just the first
before it returned the value.  When we later "restored" that locality,
any NUMA domains the thread had originally had affinity to, other than
the first one, were lost.  The effect was observed in some runs on Cray
XE systems, where the reduced affinity had the effect of changing the
number of physical cores the program thought were available to it.

To work around this, move the page-touching localization code out of the
mem=jemalloc code and into the topology code, in a new function called
chpl_topo_touchMemFromSubloc().  Here we can save the original affinity
and restore it afterward without losing information.

### DIFF
--- a/runtime/src/chpl-topo.c
+++ b/runtime/src/chpl-topo.c
@@ -227,7 +227,7 @@ c_sublocid_t chpl_topo_getThreadLocality(void) {
   }
 
   flags = HWLOC_CPUBIND_THREAD;
-  if (hwloc_get_thread_cpubind(topology, pthread_self(), cpuset, flags)) {
+  if (hwloc_get_cpubind(topology, cpuset, flags)) {
     report_error("hwloc_get_cpubind()", errno);
   }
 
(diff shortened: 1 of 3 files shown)


gbtitus:
@dmk42, would you review this one?  @ronawho is familiar with the problem this fixes, so he might be interested also.

runtime/src/chpl-topo.c
```diff
@@ -310,6 +310,58 @@ void chpl_topo_setMemSubchunkLocality(void* p, size_t size,
...
+
+  if (nPages == 0)
+    return;
+
+  if ((cpuset = hwloc_bitmap_alloc()) == NULL) {
+    report_error("hwloc_bitmap_alloc()", errno);
+  }
+
+  flags = HWLOC_CPUBIND_THREAD;
+  if (hwloc_get_thread_cpubind(topology, pthread_self(), cpuset, flags)) {
```
dmk42: The HWLOC_CPUBIND_THREAD flag is not needed here because the thread-specific function hwloc_get_thread_cpubind() is being used.  However, I would suggest keeping the flag and switching to using hwloc_get_cpubind() instead.  That would keep the abstraction level slightly higher by not having to refer to pthreads here.

But, there is another case like this in chpl_topo_getThreadLocality() which I missed earlier (sorry), so if you'd rather not make the change here and keep the two consistent, that's OK.

gbtitus: Good catch -- I'm going to fix both of those before merging.

PR #3098, opened 2016-01-12
Add a cast from type to string

Kyle-B:
This cast lets us write code like:

    writeln(foo.type:string);

instead of:

    writeln(typeToString(foo.type));

PR #5811, opened 2017-03-24
More tests for the LinearAlgebra module

npadmana:
Additional tests of the LinearAlgebra module. These are just coded versions of matrix identities, with the thought that these might exercise different parts of the linear algebra modules. 

Currently implemented --
 * (AB)^t = B^t A^t
 * (AB)C = A(BC)
 * v^t A w = tr(A w v^t)
 * tr(AB) = tr(BA)

These are relatively simple for now, but hopefully are a good model that can be extended to the LAPACK routines (where they become slightly less trivial).

- [x] finish PR text
- [x] more tests

### DIFF
--- a/test/modules/packages/linearalgebra/correctness/test_identities1.chpl
+++ b/test/modules/packages/linearalgebra/correctness/test_identities1.chpl
@@ -16,6 +16,12 @@ test_assoc(real(64));
 test_assoc(complex(64));
 test_assoc(complex(128));
 
+/* (AB)^t = B^t A^t */
+test_transpose_product(real(32));
+test_transpose_product(real(64));
+test_transpose_product(complex(64));
+test_transpose_product(complex(128));
+
 /* tr(AB) = tr(BA) */
 test_trace_rotate(real(32));
 test_trace_rotate(real(64));
@@ -37,7 +43,7 @@ test_trace_vector_rotate(complex(128));
 /* A(BC) = (AB)C */
 // Do an elementwise comparison
 proc test_assoc(type t) {
-  const testName="trace_assoc [%s]".format(t:string);
+  const testName="assoc [%s]".format(t:string);
   const eps = threshold(t);
   var A = Matrix(L, M, t);
   var B = Matrix(M, N, t);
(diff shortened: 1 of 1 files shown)


npadmana:
@ben-albrecht -- just started this, so that you can look over my shoulder if you like.

I noticed that the Makefile in the directory has ``-lcblas``, which doesn't seem to be necessary on eg. Ubuntu. Do you have an objection to me including a new Makefile in this directory?

npadmana:
One more -- is there a way to print types?

ronawho:
Current way (added in https://github.com/chapel-lang/chapel/pull/3098) to print types is to cast the type to a string:

```chpl
writeln(t:string);
```

test/modules/packages/linearalgebra/correctness/test_identities1.chpl
```diff
@@ -0,0 +1,168 @@
...
+  var diff = abs(actual/expected - 1.0);
+  if diff > thresh {
+    writef("Failure in %s : %er \n",testname, diff);
+  }
+}
+
+// Absolute comparisons
+proc testAbsCompare(expected, actual, thresh : real,testname:string) {
+  var diff = abs(actual - expected);
+  if diff > thresh {
```
monsij: Can you please describe what is the issue briefly!

npadmana: @monsij --- Sorry to be slow here, but I'm not sure what you are referring to. The function above is simply a helper function to compare a result to its expected value and print out an appropriate message if the comparison fails (given some tolerance). Mostly to avoid copying the same logic everywhere.....

Maybe these should be renamed to ``absCompare`` and ``relCompare`` to be clearer.... I'm no longer sure why I chose those names...

ben-albrecht:
> I noticed that the Makefile in the directory has -lcblas, which doesn't seem to be necessary on eg. Ubuntu. Do you have an objection to me including a new Makefile in this directory?

This appears to be necessary for me with OS X + manually-installed lapack. Lazy solution: Could we keep both flag? This seems to work for me...

The tests look good as far as I can tell. Did you run into any issues with the LinearAlgebra module while working on this? I think this just needs a `.good` file and some confirmation testing by myself to merge.

npadmana:
On Ubuntu, there is no libcblas, and so the make breaks. I don't actually care much here; it's easy enough for me to make a copy and fiddle the flags when needed. 

> The tests look good as far as I can tell. Did you run into any issues with the LinearAlgebra module while working on this? I think this just needs a .good file and some confirmation testing by myself to merge.

No, your code was flawless! :-) I'll put in the ``.good`` now.

PR #5812, opened 2017-03-24
First stab at rewriting paratest.chapcs in python

ronawho:
Note that this is a pretty direct transliteration

I think the big TODO is print the output of paratest as it's running

### DIFF
--- a/util/test/paratest.chapcs
+++ b/util/test/paratest.chapcs
@@ -1,81 +1,129 @@
-#!/usr/bin/env bash
+#!/usr/bin/env python
 
-#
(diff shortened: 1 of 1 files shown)


ronawho:
@ben-albrecht just fyi since this was your "nit" of the original version: https://github.com/chapel-lang/chapel/pull/5316#issuecomment-278405148

I could probably use a hand with creating a version of run_command that prints as it goes.

ben-albrecht:
Are you closing this until post-release?

ronawho:
Yeah, I'll wait to merge until after code-freeze, and I still need to add a run_command that prints output as it goes

PR #5814, opened 2017-03-24
Strengthen "illegal cast to non-type" error

### LINKED ISSUE #5691: Values can be used as types in some cast operations
### Summary of Problem
The compiler allows values to be used as types in some cast operations.

### Steps to Reproduce
Cast an integer to another variable, e.g. an int or string.

**Source Code:**
```
var t: string;
const c = 1: t;
```

test/expressions/vass/cast-to-non-type-1.chpl
test/expressions/vass/cast-to-non-type-2.chpl
test/users/jglewis/wrong_error_message_april_26_2010.chpl (after PR #5681 is merged)

vasslitvinov:
Fixes #5691

I aimed at making the compiler take essentially the same code path
with my change as it did before. That is, it now creates a symexpr
over a type whereas before it did over a non-type variable, for the
cast-to argument of the cast call. The next time the compiler looks
at the argument, it looks at that symexpr's type. I added an assertion
that this type is the same before my change and with my change.

I avoided more extensive changes given proximity to code freeze.
After the release branch is cut, we may want to rejigger this code
more deeply, for example:

* Rename 'sym' and 'var' to indicate whether they are the "from"
or the "to".

* Bail out early if "to" is not a SymExpr - because the current code
does nothing in such case. Maybe same w.r.t. "from".

* If "to" is not a type, replace the whole cast with the "to"
expression. Otherwise the "illegal cast to non-type" error currently
may be reported twice.

* Ignore the case where "to" is a type - instead of branching to the
"illegal cast to non-type" error in printResolutionErrorUnresolved() -
because such error has already been reported.

* Instead of the above two bullets - in such case yes, report "illegal
cast to non-type" (of course), and in addition report that a cast
to the corresponding *type* would be illegal.

While there, touched up on tests:

* Clarified types/tuple/marybeth/tuplecast.chpl
following a brief discussion on the internal mailing list.

* With Brad's approval, fixed smallIntBoolTests/realBool2.chpl

* Tightened up, using --verify, nested-class-with-user-defined-constructor-1.chpl

Testing: linux64 --verify


compiler/resolution/functionResolution.cpp
```diff
@@ -8265,7 +8271,17 @@ postFold(Expr* expr) {
...
 
       if (sym->symbol()->type != dtUnknown && sym->symbol()->type != val->type) {
-        CallExpr* cast = createCast(val, sym->symbol());
+        Symbol* toSym = sym->symbol();
+        if (!toSym->hasFlag(FLAG_TYPE_VARIABLE)) {
+          // This assertion is here to ensure compatibility with old code.
+          // Todo: remove after 1.15 release.
+          Symbol* newToSym = toSym->type->symbol;
+          INT_ASSERT(toSym->type       // cast machinery used this previously
+                  == newToSym->type);  // now it will use this
```
mppf: I'm wondering if you mean for this to be an WARNING of some sort, given that the next line wouldn't actually ever run with INT_ASSERT?

test/classes/vass/nested-class-with-user-defined-constructor-1.bad
```diff
@@ -1 +1,9 @@
 nested-class-with-user-defined-constructor-1.chpl:4: error: constructor for class 'inner' requires a generic argument called 'outer'
+nested-class-with-user-defined-constructor-1.chpl:4: In initializer 'inner':
+nested-class-with-user-defined-constructor-1.chpl:4: internal error: EXPnnnn chpl Version mmmm
+
+Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
+and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
+please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
+the filename + line number above may be useful in working around the issue.
```
mppf: It is surprising to me that this .bad file changed to an internal error. Could you investigate more and/or explain why this isn't indicative of a greater problem? I.e. are other tests/user codes now going to run into an internal error?

mppf: Or is the internal error just from running with --verify ?

compiler/resolution/functionResolution.cpp
vasslitvinov: (discussed off-line) I expect INT_ASSERT to pass and the following line to be executed.

test/classes/vass/nested-class-with-user-defined-constructor-1.bad
vasslitvinov: I confirmed that the internal error was there before my changes. It is introduced because I added --verify. In our nightly testing this test was tested without --verify.

PR #4822, opened 2016-11-04
Adjust LLVM Configurations and Buliding

mppf:
This commit adjusts LLVM support to enable new LLVM configurations.

We previously had
 * CHPL_LLVM=none -- no LLVM support
 * CHPL_LLVM=llvm -- full LLVM support with included LLVM

Now we also support
 * CHPL_LLVM=llvm-minimal -- only build and use ADTs from included LLVM
 * CHPL_LLVM=system-minimal -- only use ADTs from a system LLVM
 * CHPL_LLVM=system -- full LLVM support with a system LLVM

TODO:
 * the use of compileline during --llvm compilation is actually really slow because of the repeated invocations of Make. Ideally, compileline would only run Python code, and not even run Make. To do that, some Makefile snippets would have to move to Python.
 * investigate dynamic linking with LLVM libraries
 * upgrade to LLVM 3.8 and then 3.9. 3.8 seems to be the default that is readily available on current distributions. Note that LLVM 3.8 is the last release supporting autotools. and LLVM 3.9 requires at least cmake version 3.4.3.

Testing:

- [x] tested on Ubuntu 16.04 in all configurations with with LLVM 3.7 and 3.8 installed.
- [x] testing on Fedora 25 with system-minimal
- [x] llvm-minimal on darwin, cmake 3.2.1
- [x] llvm-minimal on Ubuntu 16.04, cmake 3.5.1
- [x]  llvm  on darwin, cmake 3.2.1
- [x]  llvm  on Ubuntu, cmake 3.5.1
- [x] llvm-minimal on darwin, no cmake
- [x] llvm-minimal on Ubuntu 16.04, no cmake
- [x]  llvm  on darwin, no cmake
- [x]  llvm  on Ubuntu, no cmake

Note: I used these packages to satisfy dependencies:

Fedora 25:
dnf install llvm3.7 llvm3.7-devel llvm3.7-static zlib-static

Ubuntu 16.04
aptitude install llvm-3.7-dev llvm-3.7 clang-3.7 libclang-3.7-dev libedit-dev

Reviewed by @dmk42 and @vasslitvinov - thanks!

PR #5819, opened 2017-03-24
Update LLVM docs

mppf:
Document new CHPL_LLVM settings added in PR #4822.
Adjust comments in third-party/llvm Makefile and shell script.

### DIFF
--- a/doc/technotes/llvm.rst
+++ b/doc/technotes/llvm.rst
@@ -30,6 +30,8 @@ generation, but no support for ``--llvm-wide-opt``:
 
   source ./util/setchplenv.bash
   export CHPL_LLVM=llvm
+  # or, if you have already installed compatible LLVM libraries
+  # export CHPL_LLVM=system
 
   make # you might want to do e.g. make -j 16 for a parallel build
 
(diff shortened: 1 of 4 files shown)


mppf:
Question for my reviewer: We don't currently make use of CHPL_LLVM settings llvm-minimal or system-minimal. It seems we should document them anyway, if the chplenv.rst is to be the reference for what values are allowed. Should we note that they aren't currently meaningful? Not document them? Document them in an RST comment? Something else?

ben-albrecht:
>Question for my reviewer: We don't currently make use of CHPL_LLVM settings llvm-minimal or system-minimal. It seems we should document them anyway, if the chplenv.rst is to be the reference for what values are allowed. Should we note that they aren't currently meaningful? Not document them? Document them in an RST comment? Something else?

My stance is that we should not document them in the user-facing documentation. I'm OK with keeping them in an rst comment, if we intend to add them in the near future. I prefer to only advertise the currently implemented features in the online documentation.

ben-albrecht:
Other than that, looks good.

mppf:
@ben-albrecht - they are implemented, it's just that we currently have no reason to use llvm-minimal or system-minimal.

ben-albrecht:
Correction -- I misunderstood the behavior of `llvm-minimal` and `system-minimal`. I still think a feature intended for developers shouldn't be in the user-facing documentation. I'd keep it commented out for now, to avoid confusion. We can add it back in if it becomes a user-facing feature.

PR #5822, opened 2017-03-24
Add a technote about void variables and fields

daviditen:
Add a technote document describing void variables and fields. Verified that
"make docs" properly generates the html version of this file.

### DIFF
--- a/doc/technotes/voidVariables.rst
+++ b/doc/technotes/voidVariables.rst
@@ -1,0 +1,66 @@
+.. _readme-voidVariables:
+
+=========================
+Void Variables and Fields
+=========================
+
+This README describes support for ``void`` variables and class/record
+fields.  A variable or field with type ``void`` will be removed from
+the program by the compiler.  It cannot be assigned to a variable of
+any non-void type, and non-void values cannot be assigned to a ``void``
+variable.  A ``void`` variable cannot be used in any context that
+requires a non-void value.
+
+The value with type ``void`` is currently named ``_void``. It is
+expected that this name will be changed.
+
+A ``void`` value can be used to instantiate fully-generic function
+arguments as long as the argument is not used in any non-void context.
+This will cause the function argument to be removed by the compiler.
+
+How is this useful?
+-------------------
+
+Often a class or record will have different implementation strategies
+depending on things like the properties of the hardware or the type of
+a generic field.  When the different implementation strategies each
+require their own class/record fields, the number of fields in the type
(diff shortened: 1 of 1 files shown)


doc/technotes/voidVariables.rst
```diff
@@ -0,0 +1,66 @@
...
+with a conditional type that is ``void`` when it is ``false``. Such
+fields will be removed by the compiler.
+
+.. code-block:: chapel
+
+  record R {
+    param isAccelerator: bool;
+    var accelVar: if isAccelerator then real else void;
+
+    if isAccelerator {
```
mppf: This conditional needs to be in a function, otherwise it's not what we have implemented

doc/technotes/voidVariables.rst
```diff
@@ -0,0 +1,66 @@
...
+``void``. The variables will be removed by the compiler.
+
+.. code-block:: chapel
+
+  var v1: void = noReturn();
+  var v2 = noReturn();
+
+In the future, a function that does not return may be differentiated from
+a function that explicitly returns a ``void`` value. In that case, the
+example above would be an error, but a function that returns a ``void``
```
mppf: I'm having trouble following. Which example would be an error? Why? Both lines in the v1/v2 example in error?

I think you need to spell out `proc noReturn` here and maybe describe the 2nd case in another example (that shows explicitly returning a void vs not returning).

doc/technotes/voidVariables.rst
daviditen: Ah, yes.  I'll add a function.

doc/technotes/voidVariables.rst
daviditen: Both v1 and v2 lines would be errors.  I'll work on making it more clear.

doc/technotes/voidVariables.rst
```diff
@@ -0,0 +1,81 @@
...
+.. code-block:: chapel
+
+  proc noReturn() {
+    writeln("Hello from noReturn");
+  }
+
+  var v1: void = noReturn();
+  var v2 = noReturn();
+
+  proc explicitReturn() {
```
mppf: Could this be written
```
  proc explicitReturn(): void {
```

and have the same behavior? Or does that make it like `noReturn` ? I think it should say.

mppf:
Looks good, except for my latest comment.

mppf:
Looks good!

daviditen:
Thank you!

PR #5824, opened 2017-03-24
Update comm counts now that array aliases are less costly

benharsh:


benharsh:
@ronawho : thanks for letting me know about this

PR #5829, opened 2017-03-24
PR #5798 resolves the bug, so this future should be removed

benharsh:


ronawho:
Should the skipif also be removed?

Issue #5834, opened 2017-03-24
Symmetrize doc/


ben-albrecht:
This issue is intended to track the follow-up effort on the comments from #5183.

PR for this issue: #5838

## Current doc hierarchy for release
```
doc/
    builtins/
    examples/
    html/
    language/
    modules/
    platforms/
    primers/
    technotes/
    tools/
    users-guide/
    usingchapel/
```
## Proposed doc hierarchy for release

```
doc/
    html/
    rst/
    README.rst
```

Changes include:

- Stuffing doc-level rst directories down one level into `rst/`
    - Modifying build paths to account for this
- Update `QUICKSTART.rst` to account for new doc-paths (currently outdated)
- Add a `doc/README.md` that explains `rst` vs. `html`, and describes some key paths, similar to `QUICKSTART.rst`'s [What's Next](http://chapel.cray.com/docs/master/usingchapel/QUICKSTART.html#what-s-next) section (using only raw-paths with no URLs).
- Add a comment to the top of `README.rst` pointing users who are not viewing the web-rendered docs, to view `doc/README.md`
- Update internals pointing to docs (pointed out by @awallace-cray)

ben-albrecht:
Ping @bradcray, who helped design this proposal.

ben-albrecht:
#5838 has been merged

PR #5838, opened 2017-03-24
Symmetrize doc/

ben-albrecht:
PR for #5834 

## Overview

This PR makes changes to the doc hierarchy such that the release doc directory looks like this:

```
doc/
    html/
    rst/
    README.rst
```

and:
- a new [doc/README.rst](https://github.com/chapel-lang/chapel/pull/5838/files#diff-1341b47450991c3744c072d9a816d1d1)
- an updated [doc/rst/usingchapel/QUICKSTART.rst](https://github.com/chapel-lang/chapel/pull/5838/files#diff-5a77b55b0441085449d315729360b363)
- an updated [README.rst](https://github.com/chapel-lang/chapel/pull/5838/files#diff-88b99bb28683bd5b7e3a204826ead112)

Pretty much all of the other changes are updating references to `doc/` -> `doc/rst/` in Makefiles, documentation, scripts, etc.

## Details

- [X] Stuff doc-level rst directories down one level into `rst/`
        - [X] Modify build paths to account for this
- [x] Update `QUICKSTART.rst` to account for new doc-paths (currently outdated)
       - Note that this file isn't intended to be viewed on GitHub, so it looks pretty bad this way.. However, at least the cross-references no longer appear to be dead-links like they are today.
- [x] Add a `doc/README.rst` that explains rst vs. html, and describes some key paths, similar to `QUICKSTART.rst`'s What's Next section (using only raw-paths with no URLs).
- [x] Add a comment to the top of `README.rst` pointing users who are not viewing the web-rendered docs, to view `doc/README.rst`
    - Also, update reference to `QUICKSTART` to point to web-rendered version
- [x] `gen_release` changes
- [x] Update internals pointing to docs
    - [x] PR ready for merge
- [x] Update [chplspell dictionary](https://github.com/ben-albrecht/chapel/pull/1#issuecomment-289188767) -- thanks @cassella 

## Testing

- [x] `make docs`
- [x] `gen_release`
- [x] `make man`
- [x] `make man-chapel`
- [x] `cd doc && make checkdocs`
    - some fairly harmless broken cross-references that show up as *italics* instead of a [hyperlink](https://github.com/chapel-lang/chapel/pull/5838) -- Will address these in a separate PR, given this PR is already pretty big.

## Follow-ups

- Remove extra copy of spec in the release tarball
- Clean up broken cross-references (low-priority)


doc/rst/usingchapel/QUICKSTART.rst
```diff
@@ -160,28 +160,28 @@ What's next?
...
+    platform-specific notes                     `doc/rst/platforms`_
     online Chapel Documentation                 `chapel.cray.com/docs`_
     example Chapel programs                     ``examples/README``
-    setting Chapel environment variables        `doc/chplenv.rst`_
-    building the compiler                       `doc/building.rst`_
-    compiling Chapel programs                   `doc/compiling.rst`_
-    executing Chapel programs                   `doc/executing.rst`_
-    debugging Chapel programs                   `doc/debugging.rst`_
-    reporting bugs                              `doc/bugs.rst`_
+    setting Chapel environment variables        `doc/rst/usingchapel/chplenv.rst`_
```
bradcray: Do you think that these should not be relative to where QUICKSTART.rst now lives?  (i.e., ``chplenv.rst`` since both files are in the same directory rather than ``doc/rst/usingchapel/chplenv.rst`` which doesn't work from this location).  Though that has the downside of requiring something like ``../platforms`` for other cases.  Alternatively, we could make this ``$CHPL_HOME/doc/rst/usingchapel/chplenv.rst`` to remove the ambiguity, though that seems a little messier.

Put another way: Note that all of the paths here were (and still are with your change) written as though QUICKSTART.rst still lives in $CHPL_HOME.  I'm guessing you'd rather not move it back up there, though?

ben-albrecht: Yeah, I noted this in the commit message. I'd be OK with a mix of relative paths for things like `chplenv.rst` and `../platforms`, and absolute paths for top-level items: `$CHPL_HOME/PERFORMANCE.md`.

README.rst
```diff
@@ -24,10 +24,14 @@ Resources
 ---------
 For more information about Chapel, please refer to the following resources:
 
-=====================  ========================================
+.. NOTE
+   If you are viewing this file locally, we recommend referring to
+   doc/README.rst for local references to documentation and resources.
+
+=====================  ==================================================================================
```
bradcray: I think the separator above (and below) would look better when read from the command-line if it fit within 80-columns (even though not all the entries below do).  Or at least if it were no longer than the longest entry below (though I'd probably still make it only 80 columns if that didn't break the .rst rendering in any way).

ben-albrecht: I believe this is necessary for the rst-rendering, but I will double check.

doc/Makefile
```diff
@@ -5,10 +5,10 @@ include Makefile.sphinx
 
 CHPL2RST     = ../util/run-in-venv.bash ./util/chpl2rst.py
 
-CHPL_DOC_PRIMERS_DIR = primers/primers
+CHPL_DOC_PRIMERS_DIR = $(SOURCEDIR)/primers/primers
```
bradcray: I realize this is pre-existing, but it's always surprised me when I cut and paste a URL with primers/primers/ in it.  I think I used to know why we had this, but every time I've run into it since it's seemed surprising...

README.rst
ben-albrecht: Looks like it works! Cutting down to 80-columns for the separators.

doc/README.rst
```diff
@@ -0,0 +1,37 @@
...
+documentation in the web format.
+
+Resources
+---------
+
+Here are some useful resources available in ``{rst,html}/``:
+
+====================================  ========================================
+Building from source                  ``usingchapel/QUICKSTART.{rst,html}``
+Platform-specific notes               ``platforms/``
```
bradcray: I'm wondering whether this should point to index.{rst,html} (but I'm not sure).

bradcray: Eh, maybe not...  Anyway, curious for your thoughts.

ben-albrecht: Not sure either. My rationale was that since this file is intended for raw-text consumption, and so browsing the directory will be more helpful than the index.rst, which is mostly sphinx-syntax:

```rst
Platform-Specific Notes
=======================

Major Platforms
---------------

.. toctree::
   :maxdepth: 1

   macosx
   cray
   knl

...
```

doc/rst/usingchapel/QUICKSTART.rst
```diff
@@ -13,7 +13,7 @@ sections explain how to re-build in the preferred configuration and how to
 enable more features, such as distributed memory execution.
 
 
-0) See `doc/prereqs.rst`_ for more information about system tools and
+0) See :ref:`prereqs.rst <readme-prereqs>` for more information about system tools and
```
bradcray: This doesn't look like it renders properly when viewed online (https://github.com/ben-albrecht/chapel/blob/4ba85c0d15d2756b2eb95c80e68855e3b3d5acd7/doc/rst/usingchapel/QUICKSTART.rst) -- is that to be expected?

doc/Makefile
ben-albrecht: I believe this was originally put into a sub-dir so that we could easily blow the generated primers away on a `make clobber` without protecting the top-level `index.rst`, and it also mirrors `examples/hellos`.

I think we can eliminate the sub-directory now that we copy `index.rst` files from `meta/` into generated directories. Would you like to see that in this PR?

doc/rst/usingchapel/QUICKSTART.rst
ben-albrecht: Yes, `QUICKSTART.rst` is not intended for viewing via GitHub. Previously, all of these links were dead-links on the GitHub-version -- try the link out for yourself [here](https://github.com/chapel-lang/chapel/blob/master/doc/usingchapel/QUICKSTART.rst).

doc/Makefile
bradcray: I think it could be a separate one, but I think it'd be a nice cleanup either way.

ben-albrecht: I'll do a separate follow-up, since this PR is already pretty huge.

ben-albrecht: Actually, I take that back -- since this PR updates so many references to `primers/primers` already, it would be easier to do all the updates at once, and include the change in this PR.

doc/rst/usingchapel/QUICKSTART.rst
```diff
@@ -157,31 +157,20 @@ the Bourne shell (sh)                ``. util/quickstart/setchplenv.sh``
 What's next?
 ------------
 
-=============================================== =========================
+=============================================== ==================================================
```
bradcray: This one could also be reduced to 80-columns.  The table itself can also almost be squeezed to 80 columns if the 5+ spaces between the columns are compressed out.  Some of the entries could be tightened up to help more (e.g., "environment variables" -> "environment vars", "file and directory" -> "file/dir" but the header in the first column is still problematic...  :(

doc/.gitignore
```diff
@@ -1,11 +1,9 @@
...
-/examples
+/rst/builtins
+/rst/modules
+/rst/primers
+/rst/examples
 
-tools/chpldoc/man.rst
-tools/chplvis/examples
-usingchapel/man.rst
+man.rst
```
bradcray: This entry looks wrong.  At least, I don't have a doc/man.rst (and wouldn't expect to).

ben-albrecht: `/man.rst` would refer to `doc/man.rst` -- however, `man.rst` applies to any file named `man.rst` (both the `chpldoc` and `chpl` man pages)

doc/rst/usingchapel/QUICKSTART.rst
ben-albrecht: If we dropped the header and instead used a short introduction before, like we do in the top-level `README.rst`, I think we can squeeze it down to 80-columns. Would you like me to do that?

doc/.gitignore
bradcray: I was not aware of that!

doc/README.rst
bradcray: OK, let's leave it as-is then.

doc/rst/usingchapel/QUICKSTART.rst
bradcray: Eh... no strong opinion.  I wouldn't go much further than low-hanging fruit.

ben-albrecht: Updated

Issue #5724, opened 2017-03-18
Improve Sort module


ben-albrecht:
Next steps for the [Sort module](http://chapel.cray.com/docs/master/modules/packages/Sort.html):

- Utilize first class functions (once FCF support improves) as comparators
- Support sorting arrays with `rank != 1`, as long as comparator specifies which rank to use
- Utilize something like a `raw()` accessor for stride/offset-independent implementations (#5368)
- Parallelism & Performance
    - Improve [shared-memory sort performance](http://chapel.sourceforge.net/perf/chapcs/?suite=standardlibrary)
    - Improve distributed sort performance
    - Add C implementations to performance-test against

xSooDx:
Hi @ben-albrecht ,
I would like to help out with this. 
I was looking at timsort and radix sort. For timsort, the implementation requires binary insertion sort for sorting below a threshold. For radix sort, the implementations require counting sort.

I can find neither binary insertion or counting sort in the Sort.chpl module. I think it would be a good idea to include these as well.

ben-albrecht:
Hi @xSooDx - thanks for your interest in contributing!

As the PR notes (maybe unclearly), @mppf plans to implement radixSort.

Sounds like a good plan for `timSort`. The `insertionSort` might be a good starting point for the `binaryInsertionSort` implementation. Also, be aware that strides and offsets in domains need to be accounted for in the Chapel implementation.

xSooDx:
I have seen the Sort module and will follow the same structure.

I have made an interesting observation:
For `binaryInsertionSort` I want to use the binary search function from the Search Module. However I do not want to use the entire module. 
Using `use Search` and `use Search only binarySearch` globaly, my code works. 
However using `use Search only binarySearch` within my `binaryInsertionSort` procedure, I get 
```
binaryInsertionSort.chpl:4: In function 'binaryInsertionSort':
binaryInsertionSort.chpl:17: error: unresolved call 'binarySearch([domain(1,int(64),false)] int(64), int(64), comparator=DefaultComparator, lo=int(64), hi=int(64))'
```
It does not show a failure on the `use` line.

My test code:
```
use Sort;
use Search only binarySearch;

proc binaryInsertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator) {
  //use Search only binarySearch;  //DOES NOT WORK
  chpl_check_comparator(comparator, eltType);
	
  const low = Dom.low,
        high = Dom.high,
[...]

PR #5840, opened 2017-03-25
timSort and Binary Insertion Sort

xSooDx:
Added `binaryInsertionSort` to the Sort module, for issue #5724


ben-albrecht:
Hi @xSooDx -- thanks for your contribution.

I will look at this PR today, but in the meantime, here are some next steps:

- [ ] Submit a CLA (if you haven't)
- [ ] Add a correctness test (to the existing correctness test suite) and ensure it passes
- [ ] Add a performance test and ensure performance is reasonable w.r.t. other sorts
- [ ] Make sure the docs build correctly

Let me know if you need documentation pointers for any of these tasks.

xSooDx:
I have already submitted the CLA. Please do point me in the right direction for the other tasks.

xSooDx:
- [ ] I still need to implement the galloping phase of the algorithm.

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+    while(n>=64){
+      r|=n&1;
+      n>>=1;
+      
+    }
+    return n+r;
+  }
+  ///PROCEDURES FOR TIMSORT ONLY///
+  //find runs in the data
+  proc getRuns(minrun:int, Data:[?Dom] ?eltType, comparator:?rec=defaultComparator) {
```
ben-albrecht: This could be `private`

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+  proc getMinrun(in n: int) : int{
+    var r: int;
+    while(n>=64){
+      r|=n&1;
+      n>>=1;
+      
+    }
+    return n+r;
+  }
+  ///PROCEDURES FOR TIMSORT ONLY///
```
ben-albrecht: I don't think this is comment is necessary

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+  //find runs in the data
```
ben-albrecht: Instead, state that this function is intended for `timSort`. Also, I wouldn't be opposed to using chpldoc-style comment for this function for 1) consistency and 2) building developer-documentation (when that feature becomes available some day)

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+          len=minrun;
+        }       
+      }
+      runs[runCount+1]=(base,len);
+      runCount+=1;
+      base+=len;      
+    } while(base<high);
+    return (runs,runCount);
+  } 
+  proc _TimSortMerge(Data:[?Dom],run1,run2,comparator:?rec=defaultComparator) {
```
ben-albrecht: Comments from `getRuns` applies here as well.

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+  const low = Dom.low,
+        high = Dom.high,
+        stride = abs(Dom.stride);
+  for i in low..high by stride {
+    var ithVal = Data[i];
+    var inserted=false;
+    var found:bool;
+    var loc:int;
+    var j:int = i-1;
+    (found,loc)=binarySearch(Data,ithVal,comparator=comparator,lo=low,hi=i);
```
ben-albrecht: a little clean-up, this could be:

```chapel
var (found, loc) = binarySearch(Data, ithVal, comparator=comparator, lo=low, hi=i);
```

modules/packages/Sort.chpl
ben-albrecht: Also, I'm OK with these being outside of the scope of `timSort` despite what other sort functions are doing... (I'll probably update those in the future).

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+    binaryInsertionSort(Data,comparator=comparator);
+    return;
+  }
+  const low = Dom.low,
+        high = Dom.high,
+        stride = abs(Dom.stride);
+  //pair for run(start,len)
+  type run = (int,int);
+  //calculate minrun length
+  proc getMinrun(in n: int) : int{
```
ben-albrecht: see `getRuns` comments

modules/packages/Sort.chpl
```diff
@@ -509,7 +509,277 @@ proc insertionSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
...
+        top-=1;
+        runs[top]=X;
+        runs[top-1]=(Z[1],Z[2]+Y[2]);
+      }      
+      
+    }
+  }
+}
+
+proc _TimSortMerge(Data:[?Dom],run1,run2,comparator:?rec=defaultComparator) {
```
ben-albrecht: Maybe I'm missing something... why is this function implemented twice?

xSooDx: This was a mistake. Will correct.

modules/packages/Sort.chpl
xSooDx: I am assuming chpldoc-style comments are 
```
/*
   Sort the 1D array `Data` in-place using a binary insertion sort algorithm.

   :arg Data: The array to be sorted
    :type Data: [] `eltType`
    :arg comparator: :ref:`Comparator <comparators>` record that defines how the
      data is sorted. 
 */
```

xSooDx:
added correctness and performance tests.
Results of performance test:
```
Time taken to sort 65536 bytes (8192 int(64)s)
quickSort (seconds): 0.087629
insertionSort (seconds): 6.62632
mergeSort (seconds): 0.246954
selectionSort (seconds): 9.19721
binaryInsertionSort (seconds): 4.05745
timSort (seconds): 0.611725
```
Suggestion: add to the correctness test an array with size>100. timSort initially seemed to pass because all test arrays were less than 64 elements.

xSooDx:
Docs are being built correctly.

xSooDx:
I am sorry for the delay but I have been spending the last few days on implementing galloping phase (sounds a lot easier than it is).

I have been facing several issues with it, and even after ironing out the bugs, it got a significance performance **decrease**. The code is also turning out to be very ugly.

Apart from that I have also found an issue while dealing with arrays with strides that I need to fix.

xSooDx:
I apologies again for the delay, been busy with college submissions.
I have reworked timSort and its related functions to work with domains, and fixed the issue with stridable arrays.

I have almost completed galloping phase implementation. I just need the answer to [this](http://stackoverflow.com/questions/43405722/how-do-i-perfrom-a-zip-iterator-in-reverse-chapel) stackoverflow question on how to iterate through a zip in reverse.

Currently the performance of timSort (with galloping in 3 out of 4 merge cases implemented):
```
Time taken to sort 131072 bytes (16384 int(64)s)
quickSort (seconds): 0.180702
mergeSort (seconds): 0.722562
binaryInsertionSort (seconds): 15.8835
timSort (seconds): 5.26966
```
Please do have a look at the code, and let me know if any changes are required.

xSooDx:
Finished Galloping phase.
Performance tests;
```
Time taken to sort 131072 bytes (16384 int(64)s)
quickSort (seconds): 0.182228
mergeSort (seconds): 0.571098
binaryInsertionSort (seconds): 16.082
timSort (seconds): 1.89295
```
With this last commit, I think everything is done for timSort.

ben-albrecht:
Hi @xSooDx - Thanks for all the hard work you've put into this implementation. I think we're getting there, but we still need comparable performance and/or a firm understanding of the performance gap before merging.

Beyond investigating the performance bottleneck of the galloping phase within `_TimSortMerge`, here are some low hanging fruit I would suggest starting with:

- Confirm we have reproduced `timSort` correctly by comparing the number of comparisons performed in Chapel's `timSort` vs. Python's `timSort`. This can be done by instrumenting the sorts with a counter comparator, e.g. in Chapel:

```chapel
class Cmp {
  var count = 0;
  proc compare(a, b) {
    this.count += 1;
    if a < b { return -1; }
    else if b < a { return 1; }
    else return 0;
  }
}

var cmp = new Cmp();
// read data from a file, so that Python and Chapel are using same data
A = getData();
timSort(A, comparator=cmp);
writeln(cmp.count);
delete cmp;
```

then do the equivalent for Python, and see if we're getting the same compares for both versions on the same data. Tim demonstrates this in his [introduction of tim sort](https://svn.python.org/projects/python/trunk/Objects/listsort.txt). We might be getting correct sorts but are still missing some key optimizations from the reference implementation. 

It would also be useful to gather Python's timings as a point of reference. I'd encourage checking this test into the branch. We can finagle it into the nightly testing down the road.

- We should re-use the `tmp` array in `_TimSortMerge` rather than re-allocating it every in every call.

- We should drop strided array special-casing and assume `stride = 1`. This should simplify some parts of the code, and improve performance for the non-strided case. We can still support strided arrays with the following wrapper:

```chapel
pragma "no doc"
proc timSort(Data: [?Dom] ?eltType, comparator:?rec=defaultComparator)
  where Dom.stridable {
  const D = {0..#Data.size};
  ref newData = Data.reindex(D);
  timSort(newData, comparator);
}
```

We'll likely pursue this idiom for all of the sort routines. Historically, we didn't because of performance impacts, but some local performance testing suggests this is no longer an issue.

And here are some less pressing comments:

- Try to match the stylistic choices made elsewhere in the `Sort.chpl` module
   - specifically, spacing
- There are some calls in which `comparator` does not get propagated, but should
- Use `private proc` for the private functions
- I'd pull out `countRun` - it might look like this:

```chapel
/* countRun explanation */
private proc countRun(Data:[?Dom], start=Dom.low, comparator:?rec=defaultComparator) {
  if(start>=Dom.last) then
    return start..start;

  const high = Dom.high;
  var last=start;
...
}

```

xSooDx:
Hi @ben-albrecht, I'll be sure to make those changes but progress will be slow as I will be having college exams for the next couple of weeks.

I wanted to know if there would be any performance difference between 
`A[x..y by s]`, `A[x..y]` and `A[y..x by -s]`
if s was =1 (and A was explicitly non stridable)?

ben-albrecht:
@xSooDx -- not a problem. Thanks for letting me know.

>I wanted to know if there would be any performance difference between
`A[x..y by s]`, `A[x..y]` and `A[y..x by -s]`
if `s =1` (and `A` was explicitly non stridable)?

Depending on the operation, arrays declared over stridable domains can have a small performance overhead, even when `stride=1`. For example, indexing into a 1-D strided array would be slightly more expensive, while iterating over a 1-D strided array incurs no additional expense.

In the sort implementations, strides are most problematic when we need to do index-math and need to convert our strided index to a non-strided index within an inner loop. I believe the reindex-wrapper approach proposed in my earlier comment would be the best way to minimize these overheads in the sort routines.

xSooDx:
I was asking because i have used `x..y by stride`, `x..y by -stride` and `i+=stride` in a number of places and wanted to know if is should remove those.

ben-albrecht:
@xSooDx - gotcha - I misinterpreted your original question (edited response).

I would still remove them to make the code cleaner, personally.

xSooDx:
@ben-albrecht I was checking the number of comparisons for timSort and found that even for sorting a 2 element array, 11 comparisons are made. This is mainly due to `chpl_check_comparator()` adding 2 comparisons, and that it is being called redundantly at the start of each procedure `timsort()->binaryInsertionSort()->binarySearch()`.

I figure I can reduce it by doing a custom implementation for binary search to avoid the condition if the element is found, but as binaryInsetionSort is used often, it will add to the comparison count.

ben-albrecht:
@xSooDx - Nice find -- did you try removing the checks to see if it has an impact on performance? I noticed that these checks are evaluated at compile-time for many cases, e.g. sorting an array of ints with a default comparator.

A few ways we could approach this:

- Create an overload or variant of `binaryInsertionSort`and `binarySearch` that does not call `chpl_check_comparator()`.

- Skip the `chpl_check_comparator()` check if one of the flags within `--fast` is enabled (not sure which flag yet).

I'd want to discuss this design decision with the broader team before pursuing one of these approaches. Until then, I would remove the checks in your local branch, and continue to look for other performance bottlenecks.

xSooDx:
@ben-albrecht  As of now, after using a custom binary search instead of the one in the search module, the number of comparisons for a 10000 element array is `337924`, while in python it is `119964`.

I think one major contributor is the galloping algorithm which I will diagnose and check.

In case of the function calls, I can create a public wrapper for `binaryInsertionSort` which performs the check and then calls a private function to do the actual sorting.

ben-albrecht:
Hey @xSooDx - I'm going to close this for now. We can re-open it whenever you make some progress.

Issue #5756, opened 2017-03-21
more array/domain/distribution user API testing


bradcray:
- [ ] add array user API testing:
  - [ ] default rectangular version
    - [ ] slice, reindex, rank-change on default rectangular
  - [ ] add block array version
    - [ ] add slice, reindex, rank-change of block
  - [ ] add version using non-int idxType(s)

- [ ] add domain user API testing
  - [ ] default rectangular version
    - [ ] slice, reindex, rank-change on default rectangular
  - [ ] add block array version
    - [ ] add slice, reindex, rank-change of block
  - [ ] add version using non-int idxType(s)

- [ ] add domain map API testing
  - [ ] default rectangular version
    - [ ] slice, reindex, rank-change on default rectangular
  - [ ] add block array version
    - [ ] add slice, reindex, rank-change of block
  - [ ] add version using non-int idxType(s)

PR #5847, opened 2017-03-26
Domain api test and improvements

bradcray:
This adds a test of the dsi routines for domains and fills in a few missing cases in the rank-change domain.  While I don't think the changes are risky, I also don't think they're necessary to merge prior to the release (as I think they're in less-used routines and won't be missed; I also think that things like dsiClear() should probably just be handled in ChapelArray.chpl for default rectangular arrays (e.g., block doesn't have a dsiClear either).


bradcray:
@benharsh: Want to take a look and see what you think?  (merge or wait)

bradcray:
Thinking about this this morning, I feel inclined to save it until after the release.  Writing the test served its purpose of reassuring me that we weren't missing any major functionality for the rank-change domain (the two routines I provided seeming non-essential for now), and the test itself could stand to be cleaned up further before being merged (taking care of the error cases, determining why I couldn't get the commented out code to work even for DR, etc.).  And I think the direction I'd ultimately really like to take this is to have things like ChapelArray.chpl implement dsiClear() for all rectangular arrays.  So I feel with the peace of mind that the code changes aren't necessary now and could be better for the test itself.  But let me know if you disagree.

benharsh:
Waiting until post-release to merge sounds fine to me.

bradcray:
Closing this for the time being until I have time to get back to it.  Noted in #5756.

PR #5848, opened 2017-03-26
errorHandling technote

psahabu:
Add a technote document describing error handling as currently implemented. `make docs` works as expected.


doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
+.. _readme-errorHandling:
+
+==============
+Error Handling
+==============
+
+This README describes the error handling feature as currently implemented.
```
mppf: Let's make some sort of reference to CHIP  or something to CHIP 8 (you could just say "CHIP 8 describes the complete design and rationale").

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+
+Errors
+------
+A base class ``Error`` is defined in the Chapel standard modules.
+
+::
```
mppf: Let's use the chapel code environment:
```
.. code-block:: chapel
```

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+  class Error {
+    var msg: string;
+
+    proc Error(_msg: string) {
+      msg = _msg;
```
mppf: I'm not so sure if it's worth including some of the details here:
 * that there is a member variable called `msg`
 * that the constructor sets msg = _msg

I think it might be best to just say something like:
```
A base class ``Error`` is defined in the Chapel standard modules. It currently includes an initializer accepting a string argument. That argument is the error message to display if the error is not caught.
```

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+::
+
+  class MyError: Error {}
+
+  class MyIntError: MyError {
+    var i: int;
+  }
+
+Chapel does not provide a standard set of ``Error`` classes.
```
mppf: Let's say "While defining a standard set of ``Error`` subclasses is planned, Chapel currently only provides ``Error``."

doc/technotes/errorHandling.rst
mppf: Also, let's say something like "An early version of error handling is included in Chapel 1.15".

doc/technotes/errorHandling.rst
mppf: (basically every time you use `::`).

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+
+Throwing Errors
+---------------
+``throw`` statements throw errors from a function to its callee.
+
+::
+
+  throw new Error("with a message");
```
mppf: While `"with a message"` is nice in that the above reads like a sentence, it doesn't match how you'd actually use it, since "with a message" is a pretty poor error message. Let's just put something like "error message here" in the string.

mppf: I think the two sections in this code block need to have some exposition splitting them up. That could be in comments or it could be 2 code blocks. It should say something like

You can throw a newly created error:


Or, you can throw an error that is already stored in a variable:

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+
+Handling Errors
+---------------
+
+Default Policy
+++++++++++++++
+
+When a throwing function is called, the default error handling policy of the
+calling function is as follows:
```
mppf: Is there some better way to describe this than "calling function" and "called function"? These terms are visually very similar. I think what would help the most would be to illustrate what you mean with a code sample like this:

``` chapel
  proc inFunction() {
     .... call() ...
  }
```

What does `call` do with respect to error handling? 


But... maybe there are other terms that would work nicely... I don't disagree with anything you're writing, just wonder if it will be hard to follow right here.

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+  proc mayThrowErrors() throws { ... }
+  proc mayThrowErrorsAlso(): A throws where ... { ... }
+  proc mayNotThrowErrors() { ... }
+
+
+Handling Errors
+---------------
+
+Default Policy
```
mppf: I don't really think the term "Policy" makes sense as section headers here - I think of "Policy" as Default vs Strict, but that's not the meaning of "Default Policy" here.

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+  type matches the type filter, the block associated with that ``catch`` clause
+  (and only that block) will be executed. 
+
+  * These filters will be evaluated in order of declaration, so even if a more
+    exact filter is present in the list, it will not be selected if a matching
+    general filter precedes it.
+
+  * If no filter is included, or only an error variable name is provided, then
+    all errors will match. Combined with the rule above, a clause with such a
+    filter must be declared at the end of the list.
```
mppf: Combined with the rule above, a clause with such a filter must be declared at the end of the list. 

->

A clause with such a filter must be declared at the end of the list., since otherwise clauses that follow it could never be executed.

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+    writeln("caught MyError");
+  }
+
+
+Strict Error Mode
++++++++++++++++++
+When the ``--strict-errors`` flag is passed to the Chapel compiler, it will
+disallow the *Default Policy* described above. As such, every throwing call
+must be handled by a *Specified Policy*. This forces throwing calls to be 
+explicitly marked and handled by the user.
```
mppf: ... and the justification for that is that it makes the control flow more apparent. Could you include that in here somehow?

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+
+
+Current Limitations
+-------------------
+* In general it is not possible to use error handling together with parallel or
```
mppf: In general -> It is not currently possible

doc/technotes/errorHandling.rst
```diff
@@ -0,0 +1,190 @@
...
+  multilocale constructs.
+  
+  * ``forall``, ``begin``, ``on``, etc.
+
+  * One exception: throwing calls handled entirely within the scope of a
+    given construct, without propagation.
+
+* Errors may not be generic classes. 
+
+* Virtual methods cannot throw.
```
mppf: Not sure if a Chapel programmer knows what a virtual method is - I mean they might from C++ but I don't think we describe it in that way in the spec. Not too big of a deal though...

doc/technotes/errorHandling.rst
mppf: By the time I got to the end, I figured out what you meant by "Default Policy". I think you need a paragraph above these sections to introduce the outline of the following. e.g.

Error handling behavior can be broken up into several cases. Because it is possible to choose between two modes, default mode and strict mode, we describe the behaviors separately:

mppf: But... It might make more sense as a document to fully describe Strict Mode first and then include a section expressing how Default Mode allows a little bit more.

mppf:
I left of lots of detailed comments but overall I think the most pressing issue is thinking about the structure of the document (in particular the sections defining error handling behavior). I think the introduction of "Default Policy" / "Specified Policy" is unnecessary and confusing and we could handle it just in terms of "Here's Strict mode. You can do XYZ... Here's default mode. In addition to what you can do in strict mode, you can also do ABC".

doc/technotes/errorHandling.rst
psahabu: Would caller/callee work? It at least drops the function bit.

doc/technotes/errorHandling.rst
psahabu: Implicit/Explicit Handling?

doc/technotes/errorHandling.rst
mppf: Something seems strange to me about "caller" in this context. I'd maybe phrase it in terms of "When a function that throws is called, what happens at the call site? Well, it depends if the call site is in a function marked with `throws` or not".

doc/technotes/errorHandling.rst
mppf: I think it'd be better to just describe Strict mode, with subsections, and then describe Default Mode as a relaxation of strict  rules.

psahabu:
Alright, I've addressed the detailed bits and am moving to restructure the whole Handling Errors section.

doc/rst/technotes/errorHandling.rst
```diff
@@ -0,0 +1,193 @@
+.. _readme-errorHandling:
+
+==============
+Error Handling
+==============
+
+This README describes the early version of error handling implemented in Chapel
+1.15. For a complete explanation of the final design and rationale, refer to
+CHIP 8.
```
bradcray: 'CHIP 8' seems like an obvious place for a hyperlink.

PR #5855, opened 2017-03-27
fixed simple TODO in LinearAlgebra.chpl and matrix IO

PrabhanjanPM:
I think using files is one of the easier ways to load a large matrix into a program. 
readMatrix() is a preliminary function. I would like to modify the function based on expected functionality. It would be rather helpful if you could suggest some improvements for the function.


ben-albrecht:
Hi @PrabhanjanPM - I think the `readMatrix` function requires additional discussion before adding, and the minor corrections have already been made on master, but the `expBySquaring` improvement would still be useful. Could you reduce the PR to just that?

ben-albrecht:
Out of curiosity, did you do any timing comparisons between the implementations? I'd be curious what the improvement is. I'll be doing this myself before merging.

PrabhanjanPM:
I reduced the PR just to improve the `expBySquaring`  function. Here are some timing comparisons( I used timer(time in seconds) from Time)-

matrix size          power           while                      recursion
10                          10               0.000143               0.000224
10                          20               0.000237               0.000144        
10                          50               0.000255               0.000198
10                          100             0.000217               0.000253
100                        10               0.003915               0.00133             
100                        20               0.001952               0.001693
100                        50               0.006036               0.002189    
100                        100             0.007631               0.005564
1000                      10               0.886028               0.704105
1000                       20              1.05983                 0.879997
1000                       50              1.41596                 1.22134
1000                       100            1.5844                   1.39682
2000                       10               6.87752                5.45924
2000                       20               8.21648                6.81767

Interestingly the running time is worse after 'improving' the function. @ben-albrecht can you point out where I might be going wrong?
I think the `while` approach might be more efficient in terms of memory, although I don't know how to measure that.

PrabhanjanPM:
I reduced the PR just to improve the `expBySquaring`  function. Here are some timing comparisons( I used timer(time in seconds) from Time)-

matrix size|power | while |recursion
--------------|---------|--------|------------
10              |            10       |        0.000143       |        0.000224
10              |            20       |        0.000237       |        0.000144        
10              |            50       |        0.000255       |        0.000198
10              |            100      |       0.000217         |      0.000253
100            |            10        |       0.003915         |      0.00133             
100            |            20        |       0.001952         |      0.001693
100            |            50        |       0.006036         |     0.002189    
100            |            100      |       0.007631         |      0.005564
1000          |            10        |       0.886028         |      0.704105
1000          |             20       |       1.05983           |      0.879997
1000          |             50       |       1.41596           |      1.22134
1000          |             100     |       1.5844              |     1.39682
2000          |             10       |        6.87752          |      5.45924
2000          |             20       |        8.21648         |       6.81767

Interestingly the running time is worse after 'improving' the function. @ben-albrecht can you point out where I might be going wrong?
I think the `while` approach might be more efficient in terms of memory, although I don't know how to measure that.

modules/packages/LinearAlgebra.chpl
```diff
@@ -137,6 +137,7 @@ module LinearAlgebra {
 
  use Norm; // TODO -- merge Norm into LinearAlgebra
  use BLAS;
+ use IO;
```
ben-albrecht: No longer necessary

ben-albrecht:
> I think the while approach might be more efficient in terms of memory, although I don't know how to measure that.

Try `./test --memStats`

Also, in case you weren't, make sure you're compiling with `--fast`.

PrabhanjanPM:
The running times are still similar to the previous experiment even after compiling with --fast
The maximum simultaneous allocated memory improves with the while loop. When using recursion the memory allocated depends on the number of ones in the binary representation of the exponent.

modules/packages/LinearAlgebra.chpl
```diff
@@ -560,16 +559,29 @@ proc matPow(A: [], b) where isNumeric(b) {
...
+private proc _expBySquaring(x: ?t, n): t{
+  // TODO -- _expBySquaring(pinv(x), -n);
+  if(n<0) then halt("Negative powers not yet supported");
+  else{
+    var product = eye(x.domain);
+    var basis = x;
+    var no = n;
+    while(no>0){
+      if(no%2==1)then{
+        product = dot(product,basis);
```
ben-albrecht: These matrix-matrix multiplications would benefit from re-using the same arrays rather than creating new arrays, which the `dot` interface doesn't currently support. You could directly call the `BLAS.gemm()` routine, or wait until we provide better support for re-using temp variables and in-place operations.

npadmana:
@PrabhanjanPM, @ben-albrecht --- it might be useful to special case out the n=2 case to just directly call dot, instead of doing any more work...... I can't decide whether it's worth doing for n=3 as well.....

modules/packages/LinearAlgebra.chpl
npadmana: I agree, no reason to pay an allocation/initialization cost every time around.

A micro-optimization -- the first multiply for product is an identity multiply, and can probably be optimized away as well. This will matter less and less for higher powers, but will for lower powers (since the code always does that at least once).

PrabhanjanPM: @ben-albrecht  I think I'll go through the Chapel language specification document before working more on this issue
Also, if I resolve some other issue, should I commit it to this pull request, or raise another pull request ?

ben-albrecht: Raise in another pull request please.

ben-albrecht:
Hey @PrabhanjanPM - I'm closing this for now. We can re-open when you make progress again.

PR #5857, opened 2017-03-27
Add forwarding technote, test of examples

### LINKED ISSUE #5682: technote for forwarding
PR  #5058 adds a forwarding feature, but for the release, we also need a technote describing it.

mppf:
Adds doc/technotes/forwarding.rst.

Closes #5682. Documents feature added in PR  #5058.

Reveiwed by @benharsh - thanks!


doc/technotes/forwarding.rst
```diff
@@ -0,0 +1,155 @@
...
+    var impl: MyCircleImpl;
+
+    proc getImplOrFail() {
+      if impl == nil then
+        halt("impl is nil");
+      else
+        return impl;
+    }
+
+    forwarding getImplOrFail();
```
benharsh: I'm not quite following how this works. Does it fail when the record is created, when or when they query ``foo.impl``, or something else?

mppf: e.g. the following would fail with the error:
``` chapel
var x:MyCircle; // impl starts out as nil
x.area(); // "impl is nil" error
```

benharsh: Ah, OK. Could that be added to this code snippet?

mppf: OK, extended the example in a similar way.

Issue #5454, opened 2017-02-28
Try to remove calls to sleep() from test suite


ronawho:
There are a number of tests that use sleep in our test suite. 

It'd be ideal to get rid of as many sleeps as possible for a few reasons

* A lot of the time sleep is being abused as a way to wait for tasks to finish. This is race-y and not safe, and most of the time a sync statement should be used instead. These tests often show up as sporadic regressions.
* Some tests use sleep as a timeout mechanism. i.e. "if this tests doesn't complete in 3 seconds, quit" These tests should be using a .timeout file instead. This makes the errors more obvious in regression mails, and is safer, because there's no gaurentee that your sleeping task will wake up after 3 seconds 
* Sleep increases testing time without adding much if any value, and makes it difficult to run multiple trials of tests 
* Seriously, this adds a lot of pointless time to testing. Most of these sleeps, sleep for at least a second, which takes hundreds of times longer than the actual test.
* even more so than not doing a chapel sleep, we should **never** extern out to system sleep. This is all but guaranteed to not work under tasking layers like qthreads, where doing a pthread sleep is incredibly wrong.
* The only tests that should be using sleep, are ones that are testing sleep

I wonder if we should have some smoke test that looks for sleeps in test/. We'd need some way to mark a test as being allowed to have a sleep in it though

Here's (way too long of) a list of tests that sleep (or more accurately have text "sleep" in them somewhere), please check them off as they are fixed or as you discover that I lied and they don't actually sleep:

- [ ] exercises/boundedBuffer.chpl 
- [ ] exercises/boundedBuffer2.chpl 
- [ ] exercises/RandomNumber6.chpl 
- [ ] extern/bradc/externClock.chpl 
- [ ] extern/sungeun/externProcInClassShouldBeError.chpl 
- [ ] functions/iterators/angeles/distAdaptativeWS.chpl 
- [ ] functions/iterators/angeles/distAdaptativeWSv1.chpl 
- [ ] functions/iterators/angeles/distAdaptativeWSv2.chpl 
- [ ] functions/iterators/angeles/dynamic.chpl 
- [ ] functions/iterators/angeles/guided.chpl 
- [ ] functions/vass/sync-to-this-of-underlying-type.chpl 
- [ ] memory/deitz/test_tricky_heap_case.chpl 
- [X] memory/diten/useArrInBegin.chpl
- [X] memory/diten/useDomInBegin.chpl
- [X] modules/diten/mutualuse.chpl
[...]

PR #5859, opened 2017-03-27
Make MPI SPMD hello.chpl consistent

ben-albrecht:
MPI can result in stdout being merged in unspecified ways even with a flush, so we throw in a sleep for good measure, to ensure output matches the `.good` file.

### DIFF
--- a/test/modules/packages/mpi/spmd/hello-chapel/hello.chpl
+++ b/test/modules/packages/mpi/spmd/hello-chapel/hello.chpl
@@ -1,10 +1,16 @@
 use MPI;
+use IO;
+
+//extern proc fflush(stream);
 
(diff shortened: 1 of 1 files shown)


ben-albrecht:
As @npadmana points out in [comments](https://github.com/chapel-lang/chapel/pull/5924#issuecomment-291010736) from #5924, this test is intended to test the setup/teardown behavior rather than the MPI_Barrier correctness, so I'm going to do a prediff instead.

ronawho:
I think you're already planning on changing this, but please don't add more uses of sleep() to our testing infra: https://github.com/chapel-lang/chapel/issues/5454

ben-albrecht:
Adding `PREDIFF`s here: #6213

PR #5869, opened 2017-03-27
Add a section to the beginning of the initializers CHIP for the status

lydia-duncan:
Intended to cover what parts of the feature are and are not supported today.


bradcray:
Note that taking this approach (as compared to putting the status in a technical note as in #5848) has the downside of making it difficult to edit the CHIP's status section prior to the 1.16 release shipping because 1.15 users may still be relying it (since we don't roll CHIPs into the release documentation yet).

Issue #5368, opened 2017-02-16
array.raw() accessor


ben-albrecht:
# The Problem

In module code, we often need to index into the Nth element, regardless of low, high, stride, and alignment of domain. Computing this translation of indices, performing a `reindex` or calling `orderToIndex()` can impact performance, when it doesn't necessarily need to.

## Example Use Cases

- Routines that need to support with non-strided 0-based indices, but should preserve existing domain offset and stride
- Algorithms that require computing indices within an inner-loop, e.g. `Sort.chpl`
- Improved C interoperability story
  - Libraries that wrap C code can use the 0-based indexing when talking to C, while providing a more Chapel-idiomatic 1-based indexing (or something else, if it makes more sense) through it's interface.

## Potential Solution

Add a new `raw` accessor to arrays, which enables a minimal-overhead 0-based 1-D indexing into the array:

```chapel
var Arr: [1..3] int;

Arr[2] = 42;

// Accesses via 0-based indexing, i.e. the (n+1)th element
writeln(Arr.raw[1]); // prints 42, the 2nd element in the array.

Arr.raw[2] = 10; // writes to the 3rd element in the array

writeln(Arr[3]); // print 10,
```

### Some Questions

- How to handle multi-dimensional arrays?
     - Index into them as a row-major 1-D array.
- How to handle for `associative`?
     - I don't think feature  makes sense for associative arrays.
- How to handle `sparse`?
     - I think raw access of sparse should index into however the underlying sparse format is being stored in a 1-D array, e.g. the `.data` field.
- Is there a better syntax or name than `raw`?
   - Probably. Naming things was never my forte.
- Should this accessor support slicing?
  - I don't see why not, if it's possible to implement.
- What other solutions are available?
  - Optimizing overhead away from `reindex` or `orderToIndex`. Probably some other ideas I have not considered.

ben-albrecht:
Broadening issue to focus on the problem rather than the proposed solution. Will reopen later with more details.

marcoscleison:
Hi Ben, I think that this kind of access would help on writing some matrix related algorithms that use the 1D representation of matrix.

ben-albrecht:
Reopening this as dealing with alignment and stride in `LinearAlgebra` implementations has been a recent pain point for me.

ben-albrecht:
I have not heard much opposition to this feature. It seems the biggest challenge would be identifying the right method name.

PR #5871, opened 2017-03-27
String join() needs domain.first

psahabu:
@cassella caught this bug in ``String.join()``. linux64+flat testing pending.

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -835,7 +835,7 @@ module String {
         return '';
       } else if S.size == 1 {
         // TODO: ensures copy, clean up when no longer needed
-        var ret = S[S.domain.low];
+        var ret = S[S.domain.alignedLow];
         return ret;
       } else {
         var joinedSize: int = this.len * (S.size - 1);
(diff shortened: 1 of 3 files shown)


ben-albrecht:
Out of curiosity - does `domain.first` accomplish the same thing?

psahabu:
I don't believe so? If the array had reverse indexing they might be different...?

ben-albrecht:
Good point. Change looks good to me. 

I'll take this opportunity to document this as another potential use-case for the feature requested in #5368

psahabu:
Actually, in the context of where this code is used, `first` works fine.

bradcray:
While in a half-awake state this morning, I found myself thinking we should consider making low/high into procedures that take an optional 'param aligned=true' argument because I think the number of cases where we get this wrong is huge...

modules/internal/String.chpl
```diff
@@ -835,7 +835,7 @@ module String {
         return '';
       } else if S.size == 1 {
         // TODO: ensures copy, clean up when no longer needed
-        var ret = S[S.domain.low];
```
ben-albrecht: This was a bug before (and still is), but this should not work for the tuple-case... I guess we don't have a test for `string.join(tuple)` anywhere.

PR #5872, opened 2017-03-27
Remove duplicate --cast-checks flag

bradcray:
In PR #5655, a duplicate copy of the flag --[no-]cast-checks was introduced.  I'm removing it here.


bradcray:
@psahabu : Would you review this since it appears you accidentally re-introduced the flag?  It might be worth double-checking the PR to make sure that there isn't any other code that was re-introduced due to merge conflicts.

psahabu:
Looked through it and didn't see any other reintroductions of old code.

PR #5875, opened 2017-03-27
Updated documentation for 'require' feature

bradcray:
This updates the current documentation on the 'require' feature to specify that:

* it now takes general param string expressions
* only those 'require' statements in active code actually add requirements


ben-albrecht:
Looks good to me. I think a brief code snippet would make the usefulness of this feature more obvious, but is not necessary.

e.g.

```chapel
config param requireFoo=true;

if requireFoo then
  require "foo.h";
else
  require "bar.h";

bradcray:
Added an example in the latest commit if you wouldn't mind re-reviewing.

PR #5881, opened 2017-03-28
Clean up malformed cross-references and URLs in docs

ben-albrecht:
This PR fixes most of the broken links and cross-references from `cd doc && make checkdocs`.

**Fixes**

All of the fixed links were fairly harmless (showed up as italics instead of link, or redirected correctly), except for a link in `cray.rst`, which was pointing users to an internal JIRA issue (whoops!). This link has been removed entirely.

Some links referenced `extern proc`s which are not supported in Chapel-sphinx, so these references were changed to ``inline code`` (mostly in the ZMQ module).

**Remaining Broken Links**

The remaining broken links are either false-negatives or are caused by `chpl2rst.py` stripping `//` from URLs. This should be fixed in the future, but redirects correctly still.

**Remaining Broken Cross-References**
It is not clear what is causing remaining broken cross-references in `ChapelArray.chpl`. The warning shows up as:

```
chpl:chplref reference target not found: bool
```

This warning doesn't appear to be serious and can be investigated further in the future.

**Testing**

- [x] `make docs`
- [x] `cd doc && make checkdocs`

### DIFF
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -703,12 +703,12 @@ Successively higher levels of overflow checking are more likely both to
 catch overflow and to catch it earlier, but they also have more overhead
 and thus a greater impact on performance.
 
-Note: in some situations the check as to whether or not the task
-stacks are in hugepage memory gets the wrong answer, leading to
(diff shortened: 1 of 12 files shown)


ronawho:
👍

PR #5883, opened 2017-03-28
Move spec change per Brad's note

mppf:
Brad pointed out it makes more sense in the Records section than in the Functions / default intent section.

I'm not particularly happy with the structure of these sections of the specification. In particular, it seems that we refer to the Classes section for methods, but methods can be declared on any type. 
I've mentioned that before in spec/TODO...

### DIFF
--- a/spec/Records.tex
+++ b/spec/Records.tex
@@ -492,12 +492,12 @@ The receiver formal argument can be referred to within the method
 using the identifier \chpl{this}.
 
 The default intent for the implicit {\em this} formal argument is
-different for record methods than it is for class methods. The default
-intent for the receiver formal argument is \chpl{ref} or \chpl{const ref},
-depending on whether the formal argument is modified inside of the
-function. Programmers wishing to be explicit about whether or not record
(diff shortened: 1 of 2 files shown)


mppf:
@benharsh - can you look at this one too?

spec/Records.tex
```diff
@@ -490,10 +491,13 @@ \subsection{The Method Receiver and the {\em this} Argument}
...
 
-The difference from a class method is that the receiver actual argument,
-which must be a record value, is passed to the record method by reference,
-rather than by copying. Therefore updates to the receiver made in the
-method, if any, are visible outside the method.
+The default intent for the implicit {\em this} formal argument is
+different for record methods than it is for class methods. The default
+intent for the receiver formal argument is \chpl{ref} or \chpl{const ref},
+depending on whether the formal argument is modified inside of the
+function. Programmers wishing to be explicit about whether or not record
```
benharsh: Should this say ``method`` instead of ``function``?

mppf:
@ben-albrecht - I just merged this spec change that Brad hopes will make it into the release.

PR #5888, opened 2017-03-28
Add future for undeclared identifier error message

benharsh:
This test should emit a nice error message for the user instead of an INT_ASSERT. This worked in 1.14, but not in 1.15.


bradcray:
Note that this could be merged even during "code freeze" since it's in test/ and outside of test/release...  (would also be nice to get into testing if the fix goes in soon).

benharsh:
Mike has fixed this issue in #5891

PR #5890, opened 2017-03-28
Update cray.rst info

ben-albrecht:
As I understand it, the issue described in this `.. note` is now resolved from  #4931 and the follow-on work.


gbtitus:
I believe that's correct, yes.

PR #5893, opened 2017-03-28
Update Spec 1.15.0 - again

ben-albrecht:
Another update after #5883 

- [x] Build new spec and check into rst/language
- [x] Sanity checks
    - [x] Index built?
    - [x] figures & tables look OK?
    - [x] Similar file size? (`du -k`)
        - still 1004 KB


mppf:
Looks good, thanks

PR #5894, opened 2017-03-28
Create a Methods chapter

mppf:
* moves Class Methods to new chapter
* move most of Class Method Calls
* including The Method Receiver and This argument

While there:
 * removes Aggregates.tex file, not used
 * removes discussion of multiple class inheritance
   since it is not intended to be implemented


bradcray:
High-level question (*):  Does 'Methods' seem better as a standalone chapter (as compared to a modifier on the functions/procedures chapter?)

* = I haven't actually read anything other than the commit message

mppf:
I don't really have a strong opinion about whether or not Methods should be its own chapter vs. a sub-section in Functions. I think it comes more down to a philosophical / terminology question - are methods a special kind of function? Or does the word "function" only refer to non-methods?

I think I was assuming that "function" only refers to non-methods, but the start of the Functions chapter says:

> A function (either a procedure or an iterator) can be a \emph{method} if it is bound to a type (often a class).

which makes me think that making it a sub-section of Functions would be best for preserving the current terminology / conceptual structure.

bradcray:
I think methods could either be their own chapter or not independently of the terminology relationship with functions (ditto iterators).  Note that while the file is called Functions.tex, I'm pretty sure the chapter is called 'Procedures' (it seems we renamed it at some point but avoided renaming the file to... avoid hiding the history?  simplify merging in a pre-git world?  not sure).

I think the last time we discussed this, we decided that 'functions' was the union of 'procedures' and 'iterators'.  I don't recall whether we decided anything specific about 'methods', but personally I think of 'methods' as being 'functions' that have a special receiver argument and syntax (where 'methods' could be either 'procedures' or 'iterators').

Also, I don't have a preference with respect to my question necessarily, just putting the question out there for consideration.

lydia-duncan:
Personally, I would make it a subsection of Functions

mppf:
@lydia-duncan, @bradcray - re the new chapter or not question, I think we need a new chapter. I came to this conclusion once I renamed Functions.tex to Procedures.tex to match the chapter heading.

Here is our current story (based on the spec):

A function can be:
 * a procedure, or
 * an iterator, or
 * a method

A procedure can be:
 * a method
 * a non-method
 * but not an iterator

A method can be:
 * a procedure
 * an iterator

Therefore, the idea of a method is not a subset of the idea of a procedure, which is what the file that used to be called Functions.tex describes. So, I'll update this PR to rename Functions.tex -> Procedures.tex but I still need somebody to actually review the changes...

spec/Methods.tex
```diff
@@ -0,0 +1,386 @@
...
+
+\section{The {\em this} Method}
+\label{The_this_Method}
+\index{methods!indexing}
+\index{this@\chpl{this}}
+\index{methods!this@\chpl{this}}
+
+A procedure method declared with the name \chpl{this} allows the receiver to be
+``indexed'' similarly to how an array is indexed.  Indexing (as with
+\chpl{A[1]} has the semantics of calling a method
```
lydia-duncan: I believe you never close that open parenthesis on the previous line

spec/Methods.tex
```diff
@@ -0,0 +1,386 @@
...
+\index{methods!iterating}
+\index{these@\chpl{these}}
+\index{methods!these@\chpl{these}}
+
+An iterator method declared with the name \chpl{these} allows the
+receiver to be
+``iterated over'' similarly to how a domain or array supports iteration.
+Using a value in the context of a loop where
+an \sntx{iteratable-expression} is expected has the semantics of calling
+a method on the value named \chpl{these}.
```
lydia-duncan: The line breaks on this paragraph should probably be fixed, though it won't impact the generated pdf

PR #5902, opened 2017-03-29
Proposed modifications to "submitted" shootouts for 1.15

bradcray:
* in meteor-fast, moved a 'begin' to make sure an array didn't leave scope before the 'begin' terminated
* in fasta, changed some 'const' variables to 'var' as they should've been from the start
* in chameneosredux, restored the destructor for the time being to continue to match what's on the shootout website

Note that the submitted meteor-fast is now pretty different from the release version.  I'm hard-pressed to say which is better...


bradcray:
@chapel-lang/perf-team : In case you weren't aware, two of our submitted shootout entries will break if/when Isaac upgrades to 1.15.  meteor-fast due to an array going out of scope before a begin is done with it; fasta due to a const array that we modify but weren't const-checking properly (oops!).  This PR creates versions that I'd submit to fix these.  It also backs out a change to chameneos that, while more in line with the future of the language isn't (yet) necessary to make.

So at some point soon (maybe tomorrow's perf meeting), I'll ask whether we're OK submitting these to the site.  The main question I have is about the differences between the submitted and release versions of meteor-fast.  Looking at the diffs, it wasn't quickly clear to me whether the larger diff for Isaac was worth whatever benefit they had (though the benefit also wasn't clear to me).

PR #5906, opened 2017-03-30
Add future for conditional expressions within init() routines

bradcray:
Using a conditional expression within an initializer generates an error message about returning within initializers.

(My wild guess is that the check for returns within an initializer is looking into a nested function that implements the conditional expression?)


bradcray:
@lydia-duncan , @noakesmichael : Assuming this looks appropriate to you, I'll open an issue when I merge it.

lydia-duncan:
D'oh!  This looks like a reasonable test to me, thanks for catching it!

bradcray:
Capturing in issue #5935

PR #5907, opened 2017-03-30
Update release+submitted binarytrees versions to the new approach

bradcray:
Isaac reset what these benchmarks are supposed to do and how they do it (no longer store item; no longer store positive and negative trees simultaneously, change checksum to count nodes), so I'm updating our versions to match.

Unfortunately, this algorithm change causes the .perfkeys cues that we were looking for previously to consider perf runs successful (-1 on first and last lines) to change and to become problem-size dependent.  This is a good thing, just an inconvenience.  He also bumped the problem size up to 21 (from 20 most recently and 16 prior to that).  So I used this as an excuse to retire both the n=16 and n=20 cases from before and just focus on n=21 for simplicity.

The net effect of this is that there will be a discontinuity in the perf results due to all of these changes.  It'll also look more dramatic than it actually is because I've co-opted the n=16 graphs (due to their use of the default filenames) for the new n=21 problem size, so those graphs will reflect a several-orders-of-magnitude jump for any pre-existing n=16 data that we'll now be splicing in.


bradcray:
@chapel-lang/perf-team : I'm looking for a review for this PR.  Though, possibly, rather than reviewing every version, you may want to:

* make sure that the submitted version matches what's on the CLBG website and understand its delta
* review the changes to the release version
* run a correctness test and performance test on all binary-trees
* see that you get reasonable graphs out

(This is more or less what I've done, but have I done it well?)

bradcray:
@chapel-lang/perf-team : Decided to merge in order to get some runs in on an otherwise quiet night.  I'd still appreciate someone giving it a post-merge review, though.

PR #5912, opened 2017-03-30
Fix rank change yield type

benharsh:
Testing:
- [x] full local
- [x] full gasnet


benharsh:
pinging @bradcray

bradcray:
@benharsh : A small puzzle, though I don't think it's a problem per se... but may be something we run into down the road.  I expanded the test to make sure all variations (serial, standalone, leader, follower) were returning ints for both slice-of-DR and slice-of-Block and put writeln()s into the iterators to make sure the right ones were getting called.  Surprisingly, I could never get the standalone that uses the downDom case to fire -- instead, it would revert to the leader/follower case (which I suspect is another case of our use of tryToken to try standalone and revert to leader/follower if it doesn't resolve).  I tried breaking this out into its own named iterator to try and debug the issue, but failed to get any clarity into the problem.  I suspect this isn't a problem now, but may be something we have to fix once Vass's forall node conversion is merged (and hopefully it'll be easier to debug then as well).  Here's the test, which we may want to file as an extension of the current one after the freeze (I didn't want to further modify the PR tonight):

````chapel
use BlockDist;

proc main() {
  var D = {1..4, 1..4};
  var Space = D dmapped Block(D);

  for i in Space[1, ..] {
    writeln(i.type:string);
  }

  for i in D[1, ..] {
    writeln(i.type:string);
  }

  forall i in Space[1, ..] {
    writeln(i.type:string);
  }
  
  forall i in D[1, ..] {
    writeln(i.type:string);
  }

  forall (i,j) in zip(D[1, ..], 1..4) {
    writeln(i.type:string);
  }

  forall (i,j) in zip(1..4, D[1, ..]) {
    writeln(j.type:string);
  }

  forall (i,j) in zip(Space[1, ..], 1..4) {
    writeln(i.type:string);
  }

  forall (i,j) in zip(1..4, Space[1, ..]) {
    writeln(j.type:string);
  }
}
````

benharsh:
I don't think you were seeing a standalone for rank-changed Block because BlockDom does not implement a standalone iterator.

bradcray:
Ohh... didn't occur to me to look down to the underlying calls.  Thanks!

PR #5913, opened 2017-03-30
Speed up reduction stress test by eliminating repetitive array building

ronawho:
reductions-stress*.chpl claims to be a reduction stress test, but almost all of
its runtime was spent building up an associative domain. This just moves the
building of the associative domain from the inner trial loop to the top level.

This dramatically reduces the runtime on chapcs:

    reductions-stress-mf   : 12.1s => 1.0s
    reductions-stress-fast : 15.5s => 0.1s

And we could probably bump the fast trial size and make it even more of a
reductions stresser now, but I'm not doing that here.

### DIFF
--- a/test/reductions/vass/reductions-stress-mf.chpl
+++ b/test/reductions/vass/reductions-stress-mf.chpl
@@ -6,10 +6,11 @@ config const r = 1000;
 writeln("n=", n, "   r=", r);
 var errors = 0;
 
+var d:domain(int);
+for i in 1..n {  d += i; }
+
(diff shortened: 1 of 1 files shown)


ronawho:
Note that this will "resolve" the reductions-stress-fast timeout under 32 bit, but David is still digging into the root cause of that regression.

ronawho:
@vasslitvinov any objections?

ronawho:
@vasslitvinov I'm going to merge to quiet the regression, but let me know if you have any objections post-commit

PR #5915, opened 2017-03-31
Align atomics

dmk42:
64-bit align 64-bit atomic objects on 32-bit systems.

On some 32-bit systems, the max forced alignment is 32 bits.  This means some 64-bit objects will be misaligned, and therefore some of those will straddle cache line boundaries.  This is not a problem unless those objects are atomic.  Then there is a serious performance penalty.

For the locks implementation of atomics, hardware atomic instructions are not used, so we do not need to worry about alignment.  For the cstdlib implementation of atomics, the compiler vendors are supposed to choose the alignment optimally for each atomic type, relieving us of the need to do so.  For the intrinsics implementation of atomics, we need to force 64-bit atomic objects to be 64-bit aligned, to guarantee that they will never straddle a cache line boundary on 32-bit systems.  This change does that.

### DIFF
--- a/runtime/include/atomics/intrinsics/chpl-atomics.h
+++ b/runtime/include/atomics/intrinsics/chpl-atomics.h
@@ -28,18 +28,33 @@
   #include <intrinsics.h>
 #endif
 
+//
+// SIZE_ALIGN_TYPE:  Declare a version of a type aligned to at least its size.
+//
+// C11 syntax does not support using this on typedefs.  However,
+// all the compilers we support with the intrinsics version of atomics
+// also support the gcc alignment attribute syntax used here.
+//
+// This is needed for 64-bit atomic types on 32-bit machines,
+// to guarantee that the atomic objects will never straddle a
+// cache line boundary, which has a huge performance penalty.
+//
+#define SIZE_ALIGN_TYPE(t) __attribute__ ((aligned (sizeof(t)))) t
+
 typedef int_least8_t atomic_int_least8_t;
 typedef int_least16_t atomic_int_least16_t;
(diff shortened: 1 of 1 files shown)


runtime/include/atomics/intrinsics/chpl-atomics.h
```diff
@@ -28,19 +28,34 @@
   #include <intrinsics.h>
 #endif
 
+//
+// SIZE_ALIGN_TYPE:  Declare a version of a type aligned to at least its size.
+//
+// C11 syntax does not support doing this to typedefs.  However,
+// all the compilers we support with the intrinsics version of atomics
+// also support the gcc alignment attribute syntax used here.
```
ronawho: What do you think of dropping the C11/compiler support info from this comment? I think it's useful information and could go in the PR message, but it doesn't seem especially helpful for context in this file (and hopefully in the future if the compiler supports C11/_Alignas, we'll be using the C11 atomics implementation instead of the intrinsics one.)

dmk42:
Thanks.  I've edited the comment as you suggested.  Brad says we should unfreeze tomorrow if all goes well (fingers crossed).  If so, I'll merge then.  I'll be traveling starting Saturday, so it's OK for anyone to revert this change if needed.  It shouldn't be necessary, though.

ronawho:
Ok. I'm not going to be around over the weekend and I'll be out mon-tues, so if you plan on merging I think it'd be a good idea to find a person to handle any issues that may come up

dmk42:
@noakesmichael has volunteered to watch this post-merge.

PR #5918, opened 2017-03-31
Tweak markup

dmk42:
Change some blocks of Chapel code in docs from code-block sh to code-block chapel.


ben-albrecht:
Can't believe I didn't catch this earlier!

Issue #5922, opened 2017-03-31
Unfriendly behavior of assigning generally real values to real(32) variables


ian-bertolacci:
### Summary of Problem
Assignments from a real-like expression to a real(32) typed variable produce the error:
`error: type mismatch in assignment from real(64) to real(32)`

For example, both of these statements produce this error:
```
var a : real(32) = 0.0;
var b : real(32) = 1 / (2.0 : real(32));
```

While this can be fixed with casting, it seems odd that 
1. The size of real literals is always assumed to be 64-bit
2. Operations with a real(32) and an int (of seemingly any size) result in a real(64)

A user might also be confused since assigning an int literal to a int(8) variable does not require casting:
```
var a : int(8) = 10; // No error
```

While not a 'bug' it is not especially user friendly, and results in lots of seemingly unnecessary casts.

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.14.0
Copyright (c) 2004-2016, Cray Inc.  (See LICENSE file for more details)
```

bradcray:
Hi Ian --

> 1) The size of real literals is always assumed to be 64-bit

Here's my take on this.  I think it's reasonable for real literals to be assumed to be 64-bit because they need to be assumed to be _something_ so that for cases like:

````chapel
var a = 1.0;
````

we have a well-defined type for 'a'.  For your first case:

> ````chapel
> var a : real(32) = 0.0;
> ````

I'd argue what you really want is for the compiler to be willing to coerce param real(64) values into real(32) values, at least in some situations.  This is the feature that permits your int case to work today:

> ````chapel
> var a : int(8) = 10; // No error
> ````

Here, '10' is an int(64), but because it's a param value (one known to the compiler), it enjoys coercions to smaller int types that can hold its value (which happens to be all of them; but if your initializer above was 512, the initialization would not be legal).  

For floating point values, when we were drafting the language, we felt uncertain how to define cases when it would vs. would not be reasonable to coerce a 64-bit float into a 32-bit float (due to a combination of lack of floating-point expertise on our team and fear of doing the wrong thing).

[...]

PR #5923, opened 2017-03-31
Consider permitting ints to coerce to real(32)s

bradcray:
The compiler was explicitly preventing int(64)s from coercing to
real(32)s.  Should it?  Here, I tried backing this out to see what, if
anything, breaks.  I think this was done because (a) we felt that ints
should coerce to _some_ floating point types, but we wanted to be
cognizant of the fact that there are int values that floating point
types can't represent accurately.  So we had them only coerce to
real(64)s in order to minimize the number of cases where loss would
occur without completely preventing them from coercing.

An intermediate point between what's on master today and opening the
gates up this wide would be to permit integers to param coerce to
real(32)s when they fit within the real(32)'s mantissa without loss of
accuracy (?).  But that feels a little...dicey.

I'm also noting that C# permits coercions from ints and longs to floats
and doubles.


bradcray:
This was motivated by issue #5922

bradcray:
This fix is too simplistic.  While it permits ints to be coerced to real(32), it also causes them to prefer real(32) over real(64) overloads, given the choice of both.  I think that it should prefer real(64) on the arguments of (a) reducing potential for inaccuracy and (b) preferring our default type widths.  For example the following call prefers the 32-bit version:

````chapel
proc foo(x: real(32)) {
  writeln("In 32-bit foo()");
}

proc foo(x: real(64)) {
  writeln("In 64-bit foo()");
}

foo(1);
````

The full list of failures was as follows where at least the ones I checked were failing due to this reason:

````
[Error matching program output for expressions/figueroa/sqrt_with_int_arg]
[Error matching man page(s) with compiler environment]
[Error matching program output for modules/standard/gmp/studies/gmp-chudnovsky]
[Error matching program output for modules/standard/math/constants]
[Error matching program output for release/examples/benchmarks/hpcc/ptrans]
[Error matching program output for studies/hpcc/PTRANS/PTRANS]
[Error matching program output for studies/hpcc/PTRANS/jglewis/ptrans_2011]
[Error matching program output for studies/hpcc/PTRANS/jglewis/ptrans_2011-blkcyc]
[Error matching program output for studies/madness/aniruddha/madchap/test_diff]
[Error matching program output for studies/madness/aniruddha/madchap/test_likepy]
[Error matching program output for studies/madness/common/test_diff]
[Error matching program output for studies/madness/common/test_likepy]
[Error matching program output for studies/madness/dinan/mad_chapel/test_diff]
[Error matching program output for studies/madness/dinan/mad_chapel/test_likepy]
````

bradcray:
My latest commit above makes int's prefer real(64)s over real(32)s but introduces ambiguities when, for example, comparing int(32) and int(64) values with '=='.  This is reminiscent of a number of coercion-related problems that I run into with some regularity (e.g., enum->uint coercions...) and is going to require some time to dig into and sort through.

bradcray:
Thinking about this a bit more, it seems to me that as a tie-breaker between multiple overloads that require coercion, we may want/need to keep a "score" of how many arguments required param and/or traditional coercion, and rank the options based on that number.  This would, for example, cause int(32) + int(64) prefer the int(64) + int(64) overload over the real(32) + real(32) and real(64) + real(64) overload because only one argument needs to coerce.

mppf:
For the sake of history, PR #7659 will resolve the param int -> real issue.
Issue #7740 is where we should discuss allowing ints to coerce to real(?w).

PR #4184, opened 2016-07-18
Bugfix : CHPL_COMM_WORLD should be set even if numLocales==1

npadmana:
We were previously only setting CHPL_COMM_WORLD if the number of locales > 1, which would cause the program to crash if the user attempted to use it when the number of locales==1.

This bugfix now sets CHPL_COMM_WORLD to MPI_COMM_WORLD when numLocales==1, which is (happily) the desired behavior even when running single-locale Chapel + MPI.

@mppf -- could you review this? Thanks!

Issue #5722, opened 2017-03-18
Improve MPI Module


ben-albrecht:
Next steps for the MPI module.

- Fix destructor order to resolve `dereference nil` bug for `_initMPI` object (#5573)
- Add `fflush` to SPMD tests (#5582)
- Support `--spmd` with `CHPL_COMM=none` for all launchers
    - Can we make this only available when using MPI module or selected configurations?
- More testing (bcast, send, recv, ....)
- Try out the futures module with MPI module over ugni or gasnet

ben-albrecht:
Credit to @npadmana for several of the above ideas (and the MPI module itself!)

npadmana:
@ben-albrecht -- on the last point, I didn't mean the futures module (which I know nothing of!), but rather that we should test the module over either ugni or gasnet with a more performant backend....

I believe this currently doesn't work.... but it would be good to have an explicit failing test....

ben-albrecht:
Marked [easy] for the additional testing bullet.

npadmana:
A few more todos (since I can't directly update the list) ---

- [ ] Extend support to Intel compilers. I've been currently testing only against gcc; are we testing the Cray versions against the Intel and Cray compilers as well? For non-Cray configurations, we may need to introduce a new target compiler analogous to ``mpi-gnu``. 

- [ ] Extend to MPI-2 and 3 routines (or some well defined subset). We currently support *most* of MPI-1, but a random subset (mostly driven by experiments I was doing) of MPI-2 and 3. We should support a more rational subset of these routines (we may not want all of them).

ben-albrecht:
@npadmana - I've updated the issue description. Feel free to delete your comment or leave it for historical purposes.

npadmana:
- [ ] Support/testing for MPI_THREAD_SERIALIZED.
   * We currently recommend and test with MPI_THREAD_MULTIPLE, which allows multiple threads to make concurrent MPI calls. It would be good to test and see if MPI_THREAD_SERIALIZED works in the case where only one thread is making an MPI call at any given time.

npadmana:
- [ ] Support for ``CHPL_TARGET_ARCH`` in ``mpi-gnu``. 
   * Currently set to ``unknown`` which produces a warning with ``--specialize`` (from ``--fast``).

npadmana:
@ben-albrecht -- I think we can cross off the qthreads stuff here, plus close #6192. Whee.

ben-albrecht:
[...]

PR #5924, opened 2017-04-01
Support CHPL_COMM=ugni, gasnet+aries and the MPI module

npadmana:
This PR fixes the MPI module to allow it to work with ``CHPL_COMM=ugni``.

The original module fails in the ``MPI_Init_thread`` with a ``GNI_CdmAttach (GNI_RC_INVALID_STATE)`` error.  

Following the discussion here -- 
  https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
the MPI module now updates ``PMI_GNI_COOKIE``.
   * In the case where ``PMI_GNI_COOKIE`` is a single number, we just increment by 1. This appears to be the case when ``CHPL_COMM=ugni``.
   * When ``PMI_GNI_COOKIE`` is a colon separated number, we increment the last number by 1, but keep the full list. 

While in the guts here --
 * also fixed an outdated comment about ``CHPL_COMM_WORLD`` not supported for ``numLocales==1`` (including SPMD mode). This was fixed in #4184, and is tested by the current SPMD code.
 * Chased many red herrings on ``gasnet+aries``. Codes here may require setting ``MPICH_GNI_DYNAMIC_CONN=disabled``, although I struggle to reproducibly trigger a failure. As we test more, this may clarify itself. We currently document this....

With lots of contributions/reviews/help from @ben-albrecht, @gbtitus and @ronawho -- thanks!!

Still to do --
- [x] Can we determine what CHPL_COMM is at compile time in the code, to do away with the ``isUGNI``param?
- [x] Update documentation
- [x] Update text in this PR
- [x] Add ugni to our configs for MPI testing?

### DIFF
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -124,10 +124,10 @@ follows:
 
 .. code-block:: sh
 
-  CHPL_TARGET_COMPILER=cray-prgenv-gnu
+  CHPL_TARGET_COMPILER=cray-prgenv-{gnu,intel}
   CHPL_TASKS=fifo
-  CHPL_COMM=gasnet
-  CHPL_COMM_SUBSTRATE=mpi           # or aries
+  CHPL_COMM={ugni, gasnet}
+  CHPL_COMM_SUBSTRATE={aries,mpi}   # if CHPL_COMM=gasnet 
   MPICH_MAX_THREAD_SAFETY=multiple
   AMMPI_MPI_THREAD=multiple         # if CHPL_COMM_SUBSTRATE=mpi
 
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -124,10 +124,10 @@ follows:
 
 .. code-block:: sh
 
-  CHPL_TARGET_COMPILER=cray-prgenv-{gnu,intel}
+  CHPL_TARGET_COMPILER=cray-prgenv-{gnu, intel}
   CHPL_TASKS=fifo
   CHPL_COMM={ugni, gasnet}
-  CHPL_COMM_SUBSTRATE={aries,mpi}   # if CHPL_COMM=gasnet 
+  CHPL_COMM_SUBSTRATE={aries, mpi}   # if CHPL_COMM=gasnet 
   MPICH_MAX_THREAD_SAFETY=multiple
   AMMPI_MPI_THREAD=multiple         # if CHPL_COMM_SUBSTRATE=mpi
 
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -131,6 +131,9 @@ follows:
   MPICH_MAX_THREAD_SAFETY=multiple
   AMMPI_MPI_THREAD=multiple         # if CHPL_COMM_SUBSTRATE=mpi
 
+Running under ``gasnet+aries`` might require setting ``MPICH_GNI_DYNAMIC_CONN=disabled``.
+This is discussed `here <http://chapel.cray.com/docs/latest/platforms/cray.html#known-constraints-and-bugs>`.
+
 These are the configurations in which this module is currently tested. Any
 launcher should work fine for this mode. Support is expected to expand in
 future versions.
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -149,9 +149,8 @@ Communicators
 The GASNet runtime, and therefore Chapel, makes no guarantees that the MPI
 ranks will match the GASNet locales. This module creates a new MPI communicator
 :proc:`CHPL_COMM_WORLD` that ensures that this mapping is true.
-Note that this is only set in mixed Chapel-MPI mode. If numLocales is 1, then
-:proc:`CHPL_COMM_WORLD` is set to :const:`MPI_COMM_NULL`, and will cause an MPI
-error if used.
+Note that if numLocales is 1, :proc:`CHPL_COMM_WORLD` is identical to :const:`MPI_COMM_WORLD`,
+which is the desired behaviour for SPMD mode programs.
 
 .. note::
   #. Pointer arguments are written as ``ref`` arguments, so no casting to a ``c_ptr``
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -263,12 +263,15 @@ module MPI {
     // If we are running using the uGNI layer, then the following hack
     // appears to be necessary in order to run MPI, as well as Chapel
     // See : https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
-    if CHPL_COMM=="ugni" {
+    if (CHPL_COMM=="ugni") || (CHPL_COMM=="gasnet" && CHPL_COMM_SUBSTRATE=="aries") {
       coforall loc in Locales do on loc {
           // This must be done on all locales!
           var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
           if !pmiGniCookie.isEmptyString() {
-            C_Env.setenv("PMI_GNI_COOKIE",("%i".format(pmiGniCookie:int+1)).c_str(),1);
+            // This may be a colon separated string.
+            var cookieJar = pmiGniCookie.split(":");
+            cookieJar[1] = ((cookieJar[1]):int + 1):string;
+            C_Env.setenv("PMI_GNI_COOKIE",("%s".format(":".join(cookieJar))).c_str(),1);
           }
         }
     }
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -262,14 +262,15 @@ module MPI {
     // If we are running using the uGNI layer, then the following hack
     // appears to be necessary in order to run MPI, as well as Chapel
     // See : https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
-    if (CHPL_COMM=="ugni") || (CHPL_COMM=="gasnet" && CHPL_COMM_SUBSTRATE=="aries") {
+    if (CHPL_COMM=="ugni") {
       coforall loc in Locales do on loc {
           // This must be done on all locales!
           var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
           if !pmiGniCookie.isEmptyString() {
             // This may be a colon separated string.
             var cookieJar = pmiGniCookie.split(":");
-            cookieJar[1] = ((cookieJar[1]):int + 1):string;
+            const lastcookie = cookieJar.domain.last;
+            cookieJar[lastcookie] = ((cookieJar[lastcookie]):int + 1):string;
             C_Env.setenv("PMI_GNI_COOKIE",("%s".format(":".join(cookieJar))).c_str(),1);
           }
         }
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -174,12 +174,7 @@ module MPI {
   use SysCTypes;
   require "mpi.h";
 
-
   use UtilReplicatedVar;
-
-  // TODO : This shouldn't need to be set by the user
-  // How do we query CHPL_COMM?
-  config param isUGNI=false;
 
   /*
      Automatically initializes MPI (if not already initialized by the runtime), and
@@ -268,7 +263,7 @@ module MPI {
     // If we are running using the uGNI layer, then the following hack
     // appears to be necessary in order to run MPI, as well as Chapel
     // See : https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
-    if isUGNI {
+    if CHPL_COMM=="ugni" {
       coforall loc in Locales do on loc {
           // This must be done on all locales!
           var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
--- a/modules/packages/MPI.chpl
+++ b/modules/packages/MPI.chpl
@@ -174,7 +174,12 @@ module MPI {
   use SysCTypes;
   require "mpi.h";
 
+
   use UtilReplicatedVar;
+
+  // TODO : This shouldn't need to be set by the user
+  // How do we query CHPL_COMM?
+  config param isUGNI=false;
 
   /*
      Automatically initializes MPI (if not already initialized by the runtime), and
@@ -260,6 +265,20 @@ module MPI {
      processes (world size)
      */
   proc initialize() {
+    // If we are running using the uGNI layer, then the following hack
+    // appears to be necessary in order to run MPI, as well as Chapel
+    // See : https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
+    if isUGNI {
+      coforall loc in Locales do on loc {
+          // This must be done on all locales!
+          var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
+          if !pmiGniCookie.isEmptyString() {
+            C_Env.setenv("PMI_GNI_COOKIE",("%i".format(pmiGniCookie:int+1)).c_str(),1);
+          }
+        }
+    }
+
+    // The actual MPI initialization goes here.
     coforall loc in Locales do on loc {
       // TODO : Need a gasnet barrier here???
       var provided : c_int;
@@ -627,4 +646,11 @@ module MPI {
   extern proc MPI_Graph_map (comm: MPI_Comm, nnodes: c_int, ref iindex: c_int, ref edges: c_int, ref newrank: c_int): c_int;
 
    } // End C_MPI
+
+
+  module C_Env {
+    // Helper routines to access the environment
+    extern proc getenv(name : c_string) : c_string;
+    extern proc setenv(name : c_string, envval : c_string, overwrite : c_int) : c_int;
+  }
 }
--- a/test/modules/packages/mpi/multilocale/hello-chapel-nl-1/SKIPIF
+++ b/test/modules/packages/mpi/multilocale/hello-chapel-nl-1/SKIPIF
@@ -14,7 +14,8 @@ retval = ((compiler !="mpi-gnu")
 # Only support fifo 
 retval |= (os.environ['CHPL_TASKS']!="fifo")
 
-# Only support GASNET+MPI 
+# Check CHPL_COMM support
+# Currently supports GASNET+MPI, GASNET+Aries, uGNI
 comm=os.environ['CHPL_COMM']
 substrate=os.environ['CHPL_COMM_SUBSTRATE']
 
--- a/test/modules/packages/mpi/multilocale/hello-chapel-nl-1/SKIPIF
+++ b/test/modules/packages/mpi/multilocale/hello-chapel-nl-1/SKIPIF
@@ -5,6 +5,7 @@ import os
 retval = True
 
 # Only support the GNU compilers for now
+# except on Crays
 compiler=os.environ['CHPL_TARGET_COMPILER']
 retval = ((compiler !="mpi-gnu")
           & (compiler!="cray-prgenv-gnu")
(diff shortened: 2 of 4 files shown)


npadmana:
Just for the record, I have currently tested this with
  ``test/modules/packages/mpi/multilocale/hello-chapel``
with the following configuration
```
CHPL_TARGET_PLATFORM: cray-xc
CHPL_TARGET_COMPILER: cray-prgenv-intel
CHPL_TARGET_ARCH: haswell
CHPL_LOCALE_MODEL: flat
CHPL_COMM: ugni
CHPL_TASKS: fifo *
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: gmake
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: ugni
CHPL_GMP: gmp
CHPL_HWLOC: none
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```

npadmana:
@ben-albrecht -- care to look at this, since you've been looking at the MPI testing most recently? 

Some questions for the reviewer --
* Is there a way to determine CHPL_COMM, so that the user isn't required to set it?
* The fix involves getting and setting environment variables. I've added in ``extern proc``s for ``getenv`` and ``setenv`` in a submodule, but maybe these should go somewhere else?
* Should we be setting a pre-diff for the tests, that sorts the output? That way we'll avoid the ``fflush`` calls we were doing. I thought there was this functionality in the testing framework.

@bradcray may also find this interesting.

npadmana:
I'll note that this addresses a testing todo in #5722 as well as implicitly testing that the MPI module works with the Intel compilers on a Cray.

modules/packages/MPI.chpl
```diff
@@ -174,8 +174,13 @@ module MPI {
   use SysCTypes;
   require "mpi.h";
 
+
   use UtilReplicatedVar;
 
+  // TODO : This shouldn't need to be set by the user
+  // How do we query CHPL_COMM?
+  config param isUGNI=false;
```
ronawho: You should just be able to say `CHPL_COMM=="ugni"` 

CHPL_COMM is a builtin param string: (http://chapel.cray.com/docs/1.15/builtins/internal/ChapelEnv.html?highlight=chapelenv#ChapelEnv.CHPL_COMM)

modules/packages/MPI.chpl
```diff
@@ -260,6 +265,20 @@ module MPI {
...
+    // If we are running using the uGNI layer, then the following hack
+    // appears to be necessary in order to run MPI, as well as Chapel
+    // See : https://hpcrdm.lbl.gov/pipermail/upc-users/2014-May/002061.html
+    if isUGNI {
+      coforall loc in Locales do on loc {
+          // This must be done on all locales!
+          var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
+          if !pmiGniCookie.isEmptyString() {
+            C_Env.setenv("PMI_GNI_COOKIE",("%i".format(pmiGniCookie:int+1)).c_str(),1);
+          }
```
ronawho: @gbtitus (our resident ugni expert) do you have any thoughts on this?

Curiously is looks like gasnet doesn't actually do this automatically: https://github.com/chapel-lang/chapel/blob/master/third-party/gasnet/gasnet-src/gemini-conduit/gasnet_core.c#L117-L143

modules/packages/MPI.chpl
npadmana: Thanks --- I knew it would be a simple fix! I'll update...

modules/packages/MPI.chpl
npadmana: That's interesting..... it might also explain issues I had trying to get the ``gasnet+aries`` case running when I tried it a few months ago. It would be great if this fixes both cases....

ben-albrecht:
re: prediff idea - would that still be testing the correctness of the barrier? BTW - here's what the `fflush` approach looks like: #5859

npadmana:
@ben-albrecht --- good point. I guess I wasn't really thinking of this test as a test of the Barrier and rather one that just exercised the initialization. I might imagine writing a test that more explicitly required the barrier to work -- i.e. this case could easily accidentally work. Somewhat amusingly, that's an easier test to write for the multilocale case than the SPMD case.

At some point, we might get tired of littering our tests with sleeps, flushes and Barriers, hence my suggestion. Although we may be stuck with the Barrier, since the I/O streams may get jumbled (again, a problem with the SPMD case, not the multi-locale case).

And in lots of cases, the correctness doesn't depend on the order that things get written in....

npadmana:
Going back to the ``CHPL_COMM`` case, I tried this out with gasnet+aries. The current module fails in an identical manner to the ugni case, suggesting that the same fix might work. However, in this case, ``PMI_GNI_COOKIE`` is a colon-separated list of numbers (2 in the cases I saw). I tried a few variants (pick either and add 1, add one to both) and I had some partial success, but the code would die (often with mmap failure unable to allocate memory) some of the time. 

Amusingly, reading the gasnet source code that @ronawho pointed to, it looks like they have just this todo (that this may be a colon separated list, but no information on how to proceed). 

So... I think this probably needs someone who actually knows what is going on here, to suggest the next step (i.e. not me!). Hopefully, we have enough breadcrumbs here to find our way home.

modules/packages/MPI.chpl
gbtitus: I think this is okay, but I'm not positive.  I believe that the "proper" way to do this would be to select an otherwise unused instance ID value (the first argument to `GNI_CdmCreate()`) instead. However, for that to work you'd have to have cooperation among all the callers of `GNI_CdmCreate()` and here you're specifically dealing with a situation where the two users of uGNI (MPI and the Chapel comm layer) don't know about each other.  So, ~~lying about~~ adjusting the cookie value may be the only practical alternative.  (Also, at the limit of its scaling the Chapel ugni comm layer will use all values of the the instance ID itself.  So in that situation there would not be any "free" ones for MPI to use even if you could get it to pick a different one than the 0 (zero) it's likely defaulting to.)

gbtitus:
I'm checking internally to see what I can find out about PMI_GNI_COOKIE.

gbtitus:
@npadmana and/or @ben-albrecht : can you give me a synopsis of the failure mode that seems to be solved by incrementing the PMI_GNI_COOKIE value and also, the (different?) failure mode that occurs when that doesn't work?  Are we sure the 2nd failure mode still has to do with the cookie value?  (The mention of an mmap() failure doesn't really sound like a cookie- or PMI- or GNI-related thing to me.)

npadmana:
@gbtitus --- Sure.

On both ``ugni`` as well as ``gasnet+aries``, when I try to initialize MPI (without fiddling with cookies), I get the following failure -- ``GNI_CdmAttach (GNI_RC_INVALID_STATE)``. 

In the case of ``ugni``, ``PMI_GNI_COOKIE`` is a single value, and incrementing it seems to work just fine. 

In the case of ``gasnet+aries``, it's a colon separated list, so there is the first question of which one of these should one be incrementing. I tried a few cases, but I didn't try them in a particularly systematic manner (sorry!). Let me pick one scheme for updating the cookies in this case and see what errors I get and document these for you. This was a case where it seemed to work sometimes but not others, and I couldn't discern a pattern.

It is absolutely possible that the second set of errors has nothing to do with the cookie values.

gbtitus:
Thanks! PMI_GNI_COOKIE seems not to be documented because it's internal to the place-and-launch support, but based on feedback from some folks inside Cray I think you should be incrementing the 1st one when it's a colon-separated multi-component value.  Once I have the additional info about the failures I'll send that along to them and see what they say.

npadmana:
Thanks --- just to be clear on what to try here -- increment the 1st and then pass the full colon separated list to MPI, or just the first number? Either is straightforward.....

gbtitus:
Retain the colon-separated syntax, but with the first number incremented.  Though this is subject to adjustment if I find out I've misunderstood how things work!  (As I understand it currently, when there are two numbers with a colon separator the 2nd number is the cookie that "belongs" to MPI.)

npadmana:
Progress! So it turns out that combination (which, for some reason, I did not try) seems to work, both on ugni as well as on gasnet.... woohoo!

ben-albrecht:
@npadmana  --

>I guess I wasn't really thinking of this test as a test of the Barrier and rather one that just exercised the initialization

In that case, I'd be OK with the prediff approach. Are there any other tests that might be more reliable for both multilocale and SPMD modes, e.g. a send/recv-based hello-world?

It's a shame we need `extern proc`'s to set env vars in a Chapel program, but I don't see any other way around it. Note that `Sys.chpl` supports: `sys_getenv`, but that's not a very Chapelific interface..

gbtitus:
Early warning: I may have gotten things backward.  It seems that when the PMI_GNI_COOKIE value is a colon-separated list, it's the 2nd component that should be used with uGNI.  The 1st component seems to be reserved for DMAPP, including when MPI uses DMAPP.  Stand by while I await confirmation on this.  (We may need to change the ugni comm layer as well, if this is true.)
Also, this may apply to the PMI_GNI_PTAG env var value as well.

bradcray:
> It's a shame we need extern proc's to set env vars in a Chapel program, but I don't see any other way around it. Note that Sys.chpl supports: sys_getenv, but that's not a very Chapelific interface..

FWIW, I'd love to see a Chapel-tastic way of querying (and maybe setting) env. vars.  One question that would need to be wrestled with in designing it is how it would interact with multi-node behavior, and launcher environment vs. compute environment.  The former could be dealt with by having the environment be accessed via the locale type.  The latter may require a better story for forwarding environment from login node to compute nodes than what we have today (which I think is a bit ad hoc and probably varies from launcher to launcher).

ben-albrecht:
Looking good overall -- before merging I need to:

- [ ] test this myself
- [ ] build and check that the docs look good
- [ ] confirm @gbtitus is OK with all the ugni 🍪 code

gbtitus:
Based on the final few comments from Cray internals folks, I think we should actually be using the _2nd_ component of a multi-component `PMI_GNI_COOKIE`, and the same for the `PMI_GNI_PTAG` value.  The first component turns out to be generally reserved for potentially VMDH-using DMAPP clients including upc, Fortran co-arrays, and certain optionally-enabled Cray MPI features.  The second component is for non-VMDH-using uGNI clients, which would mean MPI and Chapel.  But since MPI and Chapel are both going to want to use that 2nd component, one of us needs to use an incremented value of it.  I think it would be okay for Chapel to increment the value and leave the un-incremented value in the env var for MPI to use.  (Note that we need to be using the the 2nd component of both the `PMI_GNI_COOKIE` and `PMI_GNI_PTAG` env vars, but only the former needs to be incremented.  The ptag should be okay for MPI and Chapel to share, I believe.)

npadmana:
Thanks for digging this out! But a question --- are you suggesting that Chapel should use the incremented value, and then pass along the un-incremented value to MPI? Currently, we're doing the opposite -- Chapel is using the original value and we're passing on the new value to MPI. The advantages with this (I thought) are that we don't do anything non-standard when MPI isn't used. And you don't need to change the Chapel code for this, nor the ugni or gasnet+aries comm layers.

Let me know how you think we should proceed.

gbtitus:
I think Chapel could either use the incremented value, or use the original value and increment the env var.  The downside of the latter is that it only works if all other users of the env var look at it (presumably when they initialize) only after Chapel has incremented it.  The upside of the former is that the initialization order doesn't matter -- Chapel can init either before or after the other programming model(s) and things will still work.  (The current Chapel+MPI work is somewhat Chapel-centric in that it assumes the main program is written in Chapel and Chapel will initialize first.  But that's not the only way to mix things together; one could imagine an MPI-centric program that used Chapel subprograms, once we have separate compilation etc.)

So, I'm inclined toward a long-term solution in which Chapel does the incrementing, and then we don't even have to check for the presence of other programming models.  That said, for now we could certainly leave things as they are, and only implement the more general solution if we need it.  I would be okay either way, now that we understand the whole problem area better.

npadmana:
Good point! Ok, let me suggest the following..... let me update the code to increment the last cookie and verify that this works when I test it. If it does, I'll leave it up to you et al to decide what route to go. It's not a lot of code either way (on the MPI module side), so it isn't costly....

npadmana:
(never a dull moment...) Ok, updating the last cookie failed -- here is the error message...

```
*** Caught a fatal signal: SIGBUS(7) on node 0/2
NOTICE: Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace. 
Wed Apr  5 15:27:12 2017: [PE_1]:inet_recv:inet_recv: recv error (fd=5) Connection reset by peer
Wed Apr  5 15:27:12 2017: [PE_1]:_pmi_network_allgather:_pmi_inet_recv from controller failed
Wed Apr  5 15:27:12 2017: [PE_1]:_pmi_allgather:_pmi_network_allgather failed
Wed Apr  5 15:27:12 2017: [PE_1]:_pmi2_kvs_fence:_pmi_allgather failed
Wed Apr  5 15:27:12 2017: [PE_1]:PMI2_KVS_Fence:_pmi2_kvs_fence failed
Rank 1 [Wed Apr  5 15:27:12 2017] [c0-0c0s3n3] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(525)....: 
MPID_Init(287)...........: channel initialization failed
MPIDI_CH3_Init(102)......: 
MPID_nem_init(467).......: 
MPIDI_PG_SetConnInfo(653):  PMI2 KVS_Fence failed
*** Caught a fatal signal: SIGABRT(6) on node 1/2
NOTICE: Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace. 
```

Running with ``GASNET_BACKTRACE=1`` now.

npadmana:
Here's the longer error dump...

```
*** Caught a fatal signal: SIGBUS(7) on node 0/2
[0] 0: [0x57b902] _gasneti_print_backtrace_ifenabled gasnet_tools.c:? 
[0] 1: [0x60e8d0] gasneti_defaultSignalHandler ??:? 
[0] 2: [0x557a00] __restore_rt sigaction.c:? 
[0] 3: [0x9b2d3d] memset /home/abuild/rpmbuild/BUILD/glibc-2.19/string/../sysdeps/x86_64/memset.S:80 
[0] 4: [0x7ab0d0] MPID_nem_gni_dma_buffers_init ??:? 
[0] 5: [0x7acd19] MPID_nem_gni_init ??:? 
[0] 6: [0x7a4f8f] MPID_nem_init ??:? 
[0] 7: [0x79a33e] MPIDI_CH3_Init ??:? 
[0] 8: [0x791382] MPID_Init ??:? 
[0] 9: [0x760865] MPIR_Init_thread ??:? 
[0] 10: [0x760c78] PMPI_Init_thread ??:? 
[0] 11: [0x72a1a1] MPI_Init_thread ??:? 
[0] 12: [0x4b49da] wrapon_fn_chpl43 _main.c:? 
[0] 13: [0x50bbe1] thread_begin tasks-fifo.c:? 
[0] 14: [0x50cbc7] pthread_func threads-pthreads.c:? 
[0] 15: [0x5540f4] start_thread /home/abuild/rpmbuild/BUILD/glibc-2.19/nptl/pthread_create.c:309 (discriminator 2) 
[0] 16: [0x9f5809] clone /home/abuild/rpmbuild/BUILD/glibc-2.19/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:113 
Wed Apr  5 15:30:42 2017: [PE_1]:inet_recv:inet_recv: recv error (fd=5) Connection reset by peer
Wed Apr  5 15:30:42 2017: [PE_1]:_pmi_network_allgather:_pmi_inet_recv from controller failed
Wed Apr  5 15:30:42 2017: [PE_1]:_pmi_allgather:_pmi_network_allgather failed
Wed Apr  5 15:30:42 2017: [PE_1]:_pmi2_kvs_fence:_pmi_allgather failed
Wed Apr  5 15:30:42 2017: [PE_1]:PMI2_KVS_Fence:_pmi2_kvs_fence failed
Rank 1 [Wed Apr  5 15:30:42 2017] [c0-0c0s3n3] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(525)....: 
MPID_Init(287)...........: channel initialization failed
MPIDI_CH3_Init(102)......: 
MPID_nem_init(467).......: 
MPIDI_PG_SetConnInfo(653):  PMI2 KVS_Fence failed
*** Caught a fatal signal: SIGABRT(6) on node 1/2
srun: error: nid00012: task 0: Bus error
```

I'll note that this is intermittent. It worked once in between the two failures.....

npadmana:
And for completeness, here is the code that sets the cookies...

```
    if (CHPL_COMM=="ugni") || (CHPL_COMM=="gasnet" && CHPL_COMM_SUBSTRATE=="aries") {
          // This must be done on all locales!
          var pmiGniCookie = C_Env.getenv("PMI_GNI_COOKIE") : string;
          if !pmiGniCookie.isEmptyString() {
            // This may be a colon separated string.
            var cookieJar = pmiGniCookie.split(":");
            const lastcookie = cookieJar.domain.last;
            cookieJar[lastcookie] = ((cookieJar[lastcookie]):int + 1):string;
            C_Env.setenv("PMI_GNI_COOKIE",("%s".format(":".join(cookieJar))).c_str(),1);
          }
        }
    }
```

gbtitus:
Based on the traceback, you're using gasnet+mpi comms.  Did you set the environment variable `MPICH_GNI_DYNAMIC_CONN=disabled`?  The need for that is a bit obscure, but it is documented in http://chapel.cray.com/docs/latest/platforms/cray.html#known-constraints-and-bugs.

npadmana:
no, wasn't doing that... I'll try it and see if it fixes it...

npadmana:
@gbtitus -- thanks for your patience while I tracked this down.

Here's what I figured --
 * for ``ugni``, we do need to update the cookie -- that failure is reproducible. I'm following your recommendation and incrementing the last cookie, although for the ``ugni`` case, ``PMI_GNI_COOKIE`` seems to always be just one.
 * As you suspected, the ``gasnet+aries`` failures seem to be unrelated to the cookie issue. I've updated the documentation to discuss ``MPICH_GNI_DYNAMIC_CONN=disabled``, although I do not have a reproducible failure. In this case, I am **not** changing the cookie.

@ben-albrecht -- if @gbtitus signs off on this, I think this is ready for review/testing.

test/modules/packages/mpi/multilocale/hello-chapel-nl-1/SKIPIF
```diff
@@ -5,16 +5,24 @@ import os
...
 compiler=os.environ['CHPL_TARGET_COMPILER']
-retval = (compiler !="mpi-gnu") & (compiler!="cray-prgenv-gnu")
+retval = ((compiler !="mpi-gnu")
+          & (compiler!="cray-prgenv-gnu")
+          & (compiler!='cray-prgenv-intel'))
 
 # Only support fifo 
 retval |= (os.environ['CHPL_TASKS']!="fifo")
 
 # Only support GASNET+MPI
```
gbtitus: Comment needs slight update.

test/modules/packages/mpi/multilocale/hello-chapel/SKIPIF
```diff
@@ -5,16 +5,24 @@ import os
...
 compiler=os.environ['CHPL_TARGET_COMPILER']
-retval = (compiler !="mpi-gnu") & (compiler!="cray-prgenv-gnu")
+retval = ((compiler !="mpi-gnu")
+          & (compiler!="cray-prgenv-gnu")
+          & (compiler!='cray-prgenv-intel'))
 
 # Only support fifo 
 retval |= (os.environ['CHPL_TASKS']!="fifo")
 
 # Only support GASNET+MPI
```
gbtitus: Comment needs slight update.

gbtitus:
Sorry, I had forgotten that this one was awaiting a final okay from me.  I've looked through the changes and spotted just a couple of minor comment updates needed in SKIPIF files.  Everything else looks okay, and I don't need to look at it again after those two little changes are made.
👍

npadmana:
@gbtitus, @ben-albrecht --- SKIPIF comments are updated....

@ben-albrecht -- you may want to file an issue for us to clean up the MPI SKIPIFs, making some of them recursive. Let's do this after we make sure everything is working, though.

Issue #5753, opened 2017-03-21
Improve LinearAlgebra module


ben-albrecht:
There were several outstanding TODOs slated for follow-up from #5717 -- we can track them here as well as others ideas we have for the LinearAlgebra module.

## Features

- LAPACK-backed routines
- sparse array support
- distributed arrays support

## Testing

- more extensive testing: randomly generated matrices
- performance testing

ben-albrecht:
Will transfer more TODOs over from #5717 when time permits... 

Pinging @npadmana in case he misses it.

xSooDx:
Hi @ben-albrecht 

I want to apply for GSoC in your organisation and work on the Linear Algebra Library project.  Since you have already started the module, I would like to help expand the feature set.

PrabhanjanPM:
Hi @ben-albrecht

I raised a PR[1] a couple of days back that might address 'improve _expBySquaring' by fixing TODO- flatten recursion into while-loop.
I also added a preliminary function to read matrices from files which was very helpful in running some tests after making the changes. What would be the expected functionality of such functions ?
I feel file IO is the easier way to load large matrices into program. The compatibility of Chapel IO with octave/numpy IO would help in debugging and performance testing. 
[1] - https://github.com/chapel-lang/chapel/pull/5855

ben-albrecht:
@xSooDx -- Thanks for your interest, I'll mark some of the tasks above as `[easy]` to make it clear where to start.

@PrabhanjanPM - Thanks for your contribution -- I'll review the code in the PR. 

As for the `readMatrix` idea, it's not clear to me whether this should be a matrix-specific function or a more general data-reading function supporting all arrays. For example, we might want to support some subset of the IO utility routines [numpy provides](https://docs.scipy.org/doc/numpy/reference/routines.io.html) in our IO module for reading in generic data. We might also reason that native Chapel IO is elegant enough that it doesn't require such utility functions. Either way, I'd want to bounce these ideas off other core-team members before proceeding.

buddha314:
From the [Intel Page](https://software.intel.com/en-us/mkl-developer-reference-c-lapack-least-squares-and-eigenvalue-problem-driver-routines) we see a list of driver routines.  The [Singular Value Decomposition](https://software.intel.com/en-us/mkl-developer-reference-c-singular-value-decomposition-lapack-driver-routines#D0B702B7-1AED-4C13-A209-B605CEF644DC) is one of the more widespread methods.

[...]

PR #5926, opened 2017-04-01
added diag and offdiag function in package LinearAlgebra.chpl

PrabhanjanPM:
add functions for extracting diagonal/off diagonal elements of a matrix or to create a diagonal matrix with diagonal as elements of the vector argument.
Addresses some of the points in the issue - https://github.com/chapel-lang/chapel/issues/5753


modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
+  else{
+    compilerError("A must have rank 2 or less");
+  }
+
+}
+
+/*
+  Return a Vector containing the off diagonal elements of ``A``
+*/
+proc offdiag(A:[?Adom] ?eltType){
```
npadmana: We weren't particularly clear when we wrote "off-diagonal", but what we meant was ``A[i,i+k]`` where ``k`` would be an input. A good model for what we were thinking of here is the equivalent numpy function --
  https://docs.scipy.org/doc/numpy/reference/generated/numpy.diagonal.html
(where you could ignore the axis pieces).

What you are using here is probably better referred to as the "skew diagonal", and it would be nice to have a function for that (although I can't immediately think of a use for it).

modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
 //
 
+
+/*
+   Return a Vector containing the diagonal elements of ``A`` if the argument ``A`` is of rank 2.
+   Return a diagonal Matrix whose diagonal contains elements of ``A`` if argument ``A`` is of rank 1.
+ */
+proc diag(A: [?Adom] ?eltType){
+  if(Adom.rank == 2) then{
+    if(!isSquare(A)) then halt("diag supports only square matrices");
```
npadmana: I don't think we need to impose a square constraint; it would be fine for this function to return ``A[i,i]`` even if the matrix isn't square. 

In the case where the diagonal is passed it, I think returning a square matrix is fine.

modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
+/*
+  Return a Vector containing the off diagonal elements of ``A``
+*/
+proc offdiag(A:[?Adom] ?eltType){
+  if(Adom.rank != 2) then halt("offdiag expects argument to be of rank 2");
+  if(!isSquare(A)) then halt ("offdiag supports only square matrices");
+  return _diag_vec(A,trans=true);
+}
+
+proc _diag_vec(A:[?Adom] ?eltType, trans=false){
```
npadmana: This may be moot if we use the definition of off-diagonal I gave above, but I suggest using ``trans`` in a linear algebra setting only when a matrix transpose is happening, which is not the case here. It just is a little confusing otherwise...

modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
+  var diagonal = Vector(Adom.dim(1).length);
```
npadmana: I think the ``Vector`` call needs an ``eltType`` argument as well?

modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
+    forall i in Adom.dim(1) do{
+      var n = Adom.dim(1).length;
+      diagonal[i] = A[i, n-1-i];
+    }
+  }
+  return diagonal;
+}
+
+proc _diag_mat(A:[?Adom] ?eltType){
+  var diagonal = Matrix(Adom.dim(1).length);
```
npadmana: Again, I think ``Matrix`` needs an ``eltType`` as well?

modules/packages/LinearAlgebra.chpl
```diff
@@ -605,6 +604,60 @@ private inline proc _raw(D: domain(1), i) {
...
+proc offdiag(A:[?Adom] ?eltType){
+  if(Adom.rank != 2) then halt("offdiag expects argument to be of rank 2");
+  if(!isSquare(A)) then halt ("offdiag supports only square matrices");
+  return _diag_vec(A,trans=true);
+}
+
+proc _diag_vec(A:[?Adom] ?eltType, trans=false){
+  var diagonal = Vector(Adom.dim(1).length);
+  if(!trans) then{
+    forall i in Adom.dim(1) do{
```
npadmana: I think there is a subtlety here that is probably best explained with an example. The input ``A`` can be any 2D array, and in Chapel, these can have indices that start and end at any point. eg. the domain of ``A`` could be ``{4..7, 4..7}``. However, when you build the vector, it will default to ``{0..3}``, following usual C numbering.... 

So in your code below, for this case, ``i`` would run from 4 to 7, while diagonal would expect 0 to 3.....

I think it's an good question of whether, in a case like this, you want the diagonal to have indices 4..7 or 0..3? I don't know.... 

Note that this applies to the cases below as well....

PrabhanjanPM:
@npadmana 
Thank you for reviewing my code. I will make the necessary changes as soon as possible

modules/packages/LinearAlgebra.chpl
ben-albrecht: I think we could add the `k=0` optional argument to `diag` above, and remove this function.

modules/packages/LinearAlgebra.chpl
ben-albrecht: I'd add a 'where'-clause here so that we can always assume Adom is 1D:

```chapel
proc _diag_mat(A:[?Adom] ?eltType) where Adom.rank == 1 { ... }
```

Then creating the diagonal matrix can simplified:

```chapel
var diagonal = Matrix(A.size);
```

as well as the for-loop:

```chapel
forall i in Adom { .. }
```

Also, note that `do` keyword is only necessary if you are omitting the brackets, e.g.

```chapel
 forall i in Adom do diagonal[i, i] = A[i];

**Comment by ben-albrecht** [comment]
As you proceed, I recommend adding some tests to confirm to yourself that the implementation is working and to demonstrate to the reviewers what test cases you have considered. I think this could easily be worked into the `test/modules/packages/linearalgebra/correctness/correctness.chpl` suite.

**Comment by PrabhanjanPM** [comment]
@npadmana @ben-albrecht 
I have made a few changes according to your suggestions and added some simple tests.

**Comment by ben-albrecht** [comment]
Hi @PrabhanjanPM -- Things are looking better, but let's drop the `skewDiag` from this PR since it was not an originally requested feature. I think it's fine for this routine to accept only 0-based indexing as a first pass, but let's make that clear in the documentation. I suggest adding something along the line of:

```
This function only supports 0-based non-strided arrays,
including the vectors and matrices created from this module.
```

**Comment by ben-albrecht** [comment]
I'd say this is ready to merge after dropping `skewDiag` and documenting 0-based indexing.

**Comment by PrabhanjanPM** [comment]
@ben-albrecht  sorry for the late update.
I hope this is fine

**Comment by ben-albrecht** [comment]
Looks good! 

It seems that we don't have a CLA from you yet. Could you follow the instructions from [here](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements/)?

File: modules/packages/LinearAlgebra.chpl [hunk 7]
```diff
@@ -605,6 +604,96 @@ private inline proc _raw(D: domain(1), i) {
...
+  else{
+    var diagonal = Vector(Adom.dim(2).length, eltType);
+    forall i in Adom.dim(2) do{
+      diagonal[i] = A[i,i];
+    }
+    return diagonal;
+  }
+}
+
+proc _diag_vec(A:[?Adom] ?eltType, k){
```
ben-albrecht: I would add `private` to these `_diag_*` functions. Though, we can do this as a short follow-up PR.

awallace-cray:
FYI, the CLA has been checked-in.

ben-albrecht:
Thanks for your contribution @PrabhanjanPM !

Issue #5036, opened 2016-12-14
Add a standard iterator that chunks a range in a round-robin fashion


bradcray:
An idiom that seems reasonably familiar to me (and came up recently for me in test/release/examples/benchmarks/shootout/fasta.chpl) is chunking a range up by a given chunk size in a block-cyclic manner.

For example, fasta contains:

    for i in tid*chunkSize .. n-1 by numTasks*chunkSize {
      const bytes = min(chunkSize, n-i);

where tid ranges from 0 to numTasks.  The 'bytes' variable essentially asks "how large is my chunk".  The loop itself is somewhat ugly looking.  Could/should we come up with a better interface (as part of the RangeChunking module, e.g.) that would permit something more like:

    for chunk in greatIteratorName(0..#n, chunkSize, tid, numTasks) {
      const bytes = chunk.size;

where the yielded 'chunk' is a range?  ('greatIteratorName' is a placeholder because I don't have a great name for this in mind right now).

daviditen:
I've been working with @abhi-rao on this.  I think it's nearly ready, but we need to choose a name that works for everybody.  It is currently named `iter roundRobinBasedChunkOfBlocks(r: range,  blockSize, idx, numChunks)` in PR #5928.

daviditen:
Maybe: `blockCyclicChunks(r: range, blockSize, idx, numInds)`

Some possible names for the last two arguments:
```
(tid, numTasks) // The names from the first comment in this issue
(rank, numRanks)
(idx, numInds)
```

psahabu:
Even though it takes a second for me to get what is meant by `idx`, `(idx, numInds)` fits best with the rest of the library and I support that choice.

rahulghangas:
Here's something I whipped up quickly (would be cleaned properly written later) if this is still not implemented (it hasn;t been in RangeChunk.chpl). Ignore the mix of snake and camel case

```chapel
iter rr(r, blockSize, idx, numInds){
  
  var range_stride = r.stride;
  var block_stride = blockSize * range_stride;
  var low = r.low;
  var high = r.high;
  var first_block_start = (if range_stride > 0 then r.low - 1 else r.high + 1) + block_stride * (idx + 1);
  if first_block_start > r.high || first_block_start < r.low then return;

  var stride_to_next_block = block_stride * numInds;
  
  if range_stride > 0 {
    for block_start in first_block_start..high by stride_to_next_block {
      var block_end = min(high, block_start + block_stride - 1);
[...]

PR #5928, opened 2017-04-03
Addition roundRobinBasedChunkOfBlocks to RangeChunk.chpl

abhi-rao:
Addition of a round robin based iterator, whose chunk contains contiguous ranges of size blocksize.
Adds feature mentioned in #5036.

### DIFF
--- a/modules/packages/RangeChunk.chpl
+++ b/modules/packages/RangeChunk.chpl
@@ -95,7 +95,14 @@ module RangeChunk {
     }
     const first = r.orderToIndex(0);
     const last = r.orderToIndex(nElems - 1);
-    const rangeStride = ((abs(last - first)/nElems) + 1)*sgn(last - first);
+
+    var rangeStride = 0;
+    if(nElems > 1) {
+      rangeStride = (abs(last - first)/(nElems - 1))*sgn(last - first + 1);
+    }
+    if(rangeStride == 0) {
+      rangeStride = abs(last - first + 1);
+    }
     const ordinalStartIndex = idx*blockSize;
     if(ordinalStartIndex >= nElems) {
       return;
--- a/modules/packages/RangeChunk.chpl
+++ b/modules/packages/RangeChunk.chpl
@@ -85,6 +85,23 @@ module RangeChunk {
   }
 
   /*
+     Iterates through blocks of size ``blockSize`` of ``idx`` chunk of range ``r``, in a round robin fashion.
+  */
+  iter roundRobinBasedChunkOfBlocks(r: range(?RT, bounded, ?S),  blockSize: integral,
+             idx:integral, numChunks: integral): range(RT, bounded, S) {
+
+    const nElems = r.length;
+    const start = idx*blockSize;
+    const stride = blockSize*numChunks;
+    const end = nElems - 1;
+    for chunkStartIdx in start..end by stride {
+      const chunkEndIdx = min(chunkStartIdx + blockSize, nElems) - 1;
+      yield chunkStartIdx..chunkEndIdx;
+    }
+    
+  }
+
+  /*
      Iterates through chunks ``0`` to ``numChunks - 1`` of range ``r``, emitting each
      as a 0-based order tuple. The remainders will be distributed according to ``remPol``.
   */
--- a/modules/packages/RangeChunk.chpl
+++ b/modules/packages/RangeChunk.chpl
@@ -89,15 +89,31 @@ module RangeChunk {
   */
   iter roundRobinBasedChunkOfBlocks(r: range(?RT, bounded, ?S),  blockSize: integral,
              idx:integral, numChunks: integral): range(RT, bounded, S) {
-
     const nElems = r.length;
-    const rangeStride = ((r.orderToIndex(nElems - 1) - r.orderToIndex(0))/nElems) + 1;
-    const start = r.orderToIndex(idx*blockSize);
+    if(nElems == 0) {
+      return;
+    }
+    const first = r.orderToIndex(0);
+    const last = r.orderToIndex(nElems - 1);
+    const rangeStride = ((abs(last - first)/nElems) + 1)*sgn(last - first);
+    const ordinalStartIndex = idx*blockSize;
+    if(ordinalStartIndex >= nElems) {
+      return;
+    }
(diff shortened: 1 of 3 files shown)


daviditen:
Hi @abhi-rao,

This looks like a good start.  A few things to watch out for:

* Does this properly block up ranges that don't start at 0? For example: `1..100`, `-7..25`, `25..50`?
* Does it work right with ranges that are already strided? For example: `0..99 by 2`, `1..100 by 3`, `-7..25 by 5`, `3..30 by -2`?

The way I'd typically deal with those two questions is to subtract away the low bound of the input range and divide it by the stride before working with it, then do the opposite to the ranges about to be yielded.

Adding some tests of this new functionality would be a nice way to show that it handles all of the cases right. See: [TestSystem](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst).

I think we may need to work on the iterator name a bit more in order to gain wider approval.

P.S. Have you signed a contributor agreement?  [ContributorAgreements](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements)

Thanks!

abhi-rao:
Hi @daviditen,

I have added tests as well as made some modifications to handle the cases suggested.

Thanks

daviditen:
Hi @abhi-rao,

Thanks for writing the new test, it is looking good.  I think you forgot to commit your new changes to RangeChunk.chpl though.

Thanks

abhi-rao:
Hi @daviditen,
Sorry for the mistake.
Thanks
Abhijith B Rao

daviditen:
Hi @abhi-rao,

Thanks for the update.  I ran your new test and everything passed there.  I also tried my own test with each of the ranges I listed above and most of them worked, but `-7..25 by 5` and `3..30 by -2` each got errors.  I also tried `1..100 by -1`, and it didn't get an error, but it didn't give any chunks at all.

Could you please take another look at cases like these?

Thanks!

abhi-rao:
Hi @daviditen ,
If the range to be chunked is empty or if the range has no chunks at it returns void.
Thanks.

daviditen:
Hi @abhi-rao,

Thanks for continuing on with this PR.

I tried out your test and my test that I mentioned in an earlier comment, and both passed. I extended my test to check a large number of randomly created ranges, and unfortunately I'm still getting unexpected errors and incorrect results.

One of the more common errors I'm seeing is:
`$CHPL_HOME/modules/packages/RangeChunk.chpl:107: error: the step argument of the 'by' operator is zero`

One issue I see that can lead to this error is on line 98 in RangeChunk.chpl.  If the range only has one element then `last == first`, so `sgn(last-first)` is 0, leading to a stride of 0.

I'm also still finding some indices in the ranges yielded by `roundRobinBasedChunkOfBlocks` that weren't in the original range.

Here is my randomized test.  I think it should be able to execute without getting any errors, and it should also never print the word "failed".

[randRangeChunk.chpl](https://gist.github.com/daviditen/dabb44b398581d4502dcd1d9c8b6072b)

It would be good to play with the `block_size` and `num_tasks` config constants to make sure different values for them work as well.

abhi-rao:
Hi @daviditen,
I have tested the module for 10000 random cases, and it passed.
Should it also handle cases of unbounded ranges?

daviditen:
Hi @abhi-rao,

Thanks for the update! This looks like it's getting close to ready to merge.

I think it would be neat to eventually support ranges that are only bounded on one side,  for example chunking up the range `1..` with `blockSize=5` might yield `1..5`, `6..10`, `11..15`, `...`.  I think that this isn't necessary for this PR though, and instead a nice compiler error message would suffice.

You could add that error message by changing the iterator to take a generic range argument `r: range(?)`, and then adding something like the following:

```
if r.boundedType != BoundedRangeType.bounded then
  compilerError("range argument to iterator 'roundRobinBasedChunkOfBlocks' must be bounded");
```

I see that the `blockSize` and `numChunks` arguments don't give very friendly errors if either of them is <= 0.  These errors could have nicer runtime errors added, something like:

```
if (blockSize <= 0) then
  halt("blockSize must be greater than 0 in iterator 'roundRobinBasedChunkOfBlocks'");
```

I'm not yet sure what exactly this iterator should be named, but if you wanted to argue in favor of any particular names, you can make suggestions and comment in issue #5036.

abhi-rao:
I have added appropriate error messages for invalid arguments.
And blockCyclicChunks seemed really an appropriate name.

abhi-rao:
Regarding the contributor agreement, I will send it at the earliest.

daviditen:
Thanks for adding the error checking. I think the error for `boundedType` should be a `compilerError` instead of a `halt`.  This will issue an error during compilation instead of waiting until the program is executing.  It also needs to be a `compilerError` to let you use functions that only apply to bounded ranges in the rest of the iterator, such as `r.length`.

abhi-rao:
Hi @daviditen ,
I have submitted the Chapel Contributor Agreements, and made the necessary changes.

daviditen:
@abhi-rao,

We received your contributor agreement and it looks good, thanks!

daviditen:
@abhi-rao,

My random ranges test caught another case that gets an unexpected error: any range with two elements and a stride of -1, for example `0..1 by -1`.  But before you go looking for a fix, continue reading.

I just discovered that a method that could have simplified your life with this PR is not documented in the Ranges online documentation.  I'm putting together a PR to add documentation for it, and the method is:

```
  /* Returns the stride of the range */
  inline proc range.stride
```

You can call this method on ranges to get the stride instead of needing to compute it yourself, e.g. 

`rangeStride = r.stride;`

I'm sorry that this documentation was missing and I didn't catch it sooner.

daviditen:
I added some documentation for the range methods `range.low`, `range.high`, `range.stride`, `range.alignment`, and `range.aligned`, and they're included in the [master branch documentation](http://chapel.cray.com/docs/master/builtins/internal/ChapelRange.html) online now.

daviditen:
Hi @abhi-rao,

We're trying to reduce the number of open PRs we have, so I'm going to close this for now.  Please ping us here or through one of our mailing lists if you get a chance to work on it again and we can re-open this PR.

PR #5929, opened 2017-04-03
Fix rank changes with mixed stride

bradcray:
@tmacd9 sent a code indicating that we'd lost the ability to mix ranges of different stridable-ness between 1.14 and now as a result of the array view rewrite for how we do rank-change.  This PR fixes that issue and also does a code clarification requested by @mppf in his review of array views that I didn't have the chance to tackle until now.

The root cause of the problem was that the recursive routine that gathered all of the ranges used to represent the upward-facing domain for a rank-change expression created a mix of stridable and non-stridable ranges even though in practice all of a domain's ranges must share the same stridable-ness.  I didn't think that I had changed this behavior since the 1.14 release, but must be mistaken.  As a result, if a rank-change slice contained some stridable and some non-stridable ranges, things would break downstream.

The fix here is to pre-compute a shared type for the upward-facing ranges prior to computing them, by examining all of the rank-change slice arguments and seeing whether any of the range arguments are stridable.  If they are, or if the original domain was stridable, the whole upward-facing domain should be and we can declare its ranges as such.

In implementing this fix, I removed a complicated recursive routine that had been used to gather these ranges and replaced them with a recursive routine that counts the number of ranges, a declaration of the tuple-of-range type, and a more straightforward assignment to the tuple elements as part of a (pre-existing) iterative loop over the rank-change arguments.

I also added TMac's motivating test to prevent a slip like this from occurring again.

TODO:
- [x] linux64 testing
- [x] gasnet testing


modules/internal/ChapelArray.chpl
```diff
@@ -3100,13 +3105,33 @@ module ChapelArray {
...
+  }
+
+  // the following pair of routines counts the number of ranges in its
+  // argument list and is used for rank-change slices
+  proc chpl__countRanges(arg) param {
+    return isRangeValue(arg):int;
+  }
+
+  proc chpl__countRanges(arg, args...) param {
+    return chpl__countRanges(arg):int + chpl__countRanges((...args));
```
mppf: It seems odd to me that this :int cast is necessary here, since the above 1-arg routine clearly returns an int.

modules/internal/ChapelArray.chpl
```diff
@@ -3147,38 +3172,6 @@ module ChapelArray {
...
-        if isRange(args(dim)) {
-          const newRange = downdom.dsiDim(dim)[args(dim)]; // intersect ranges
-          return ((...x), newRange);
-        } else {
-          return x;
-        }
-      }
-    }
-  }
-
```
mppf: 👍

mppf:
I think this is great, strictly from a code clean-up perspective. Since it fixes a bug, all the better!

modules/internal/ChapelArray.chpl
bradcray: Oh, good catch.  This used to be the same isRange call which is why it's there...

bradcray:
@awallace-cray: Sorry for the delay -- this is ready to be promoted to the release branch now.

PR #5931, opened 2017-04-03
Scope update

noakesmichael:
Work on initializers provoked me to look as some "normalization" code for methods that's
currently tucked in to one of the inner loops of scope resolution.   The surrounding code
seems to me to be unreasonably/unnecessarily convoluted.

This PR refactors that portion of scope-resolve a) on principle b) to support possible next
steps for initializers.

This PR does not / should not change the behavior of any Chapel program.  The PR consists
of a relatively large number of relatively small commits in order to simplify the task of confirming
that the business logic is unchanged.

Compiled with/without CHPL_DEVELOPER on clang/darwin and gcc/linux.

Also compiled with CHPL_LLVM=llvm as I made trivial changes to some code inside a LLVM
guard in scopeResolve.cpp

Standard single-locale paratest on linux64 with -futures and then with --llvm

### DIFF
--- a/compiler/passes/scopeResolve.cpp
+++ b/compiler/passes/scopeResolve.cpp
@@ -1742,17 +1742,17 @@ static void insertFieldAccess(FnSymbol*          method,
   const char* name      = usymExpr->unresolved;
   Type*       type      = method->_this->type;
   int         nestDepth = computeNestedDepth(name, type);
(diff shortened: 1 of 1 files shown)


lydia-duncan:
Man, some of that code was pretty crufty.

PR #5933, opened 2017-04-04
Make ReplicatedDist less distinct

bradcray:
This change makes ReplicatedDist behave a lot less "specially" relative to other distributions, primarily by removing operations that implicitly acted on all replicands such as loops and I/O.  In the new world, most operations will only occur on the local locale's replicand.  The cases that still operate on all replicands are:

* creation and destruction of domains and arrays: updates all copies
* setting a domain's indices: updates all copies

So cases that have changed are:

* iterators used to iterate over all indices/elements in all replicands, but now only iterate over the local indices/elements
* numIndices/numElements used to return the number of indices/elements multiplied by the number of replicands but now just returns the number of indices/elements in a single replicand
* printing domains used to print the index set followed by the list of locales that own a replicand; now it just prints the index set
* printing arrays used to print all the replicands, but now just prints the local replicand.

* in modules/dists/ReplicatedDist.chpl:
  - implemented the changes as described above
  - updated the documentation
  - introduced helper routines to get the local replicand or generate an error if there isn't one
* in modules/standard/UtilReplicatedVar.chpl:
  - updated example code to avoid assumption that printing would print all copies
* updated test/release/examples/primers/distributions.chpl to reflect the new semantics
* in test/distributions/vass/testReplicated* and test/parallel/forall/vass/ri-4-arrdom.chpl:
  - updated tests to maintain correctness under new semantics
  - replaced wrapper* versions of tests to use comm-none.good and .numlocales instead
* simplified test/distributions/robustness/arithmetic by:
  - removing *.replicated.good files for cases where it no longer differs
  - moving *.default.good files to *.good files and removing PREDIFFs when possible
  - removing CLEANFILES and .gitignore entries for such cases
* updated comm counts in test/distributions/robustness/performance/multilocale

This is a work-in-progress to try and get ReplicatedDist's behavior less unique and to get testing quiet again.

TODO:
- [x] pass gasnet testing at least once
- [x] get buy-in from team
- [x] code cleanup
- [x] linux64 testing
- [x] gasnet testing
- [ ] add test of "no local replicand" style errors
- [ ] why does testReplWrites get a valgrind error on 1 locale?
- [ ] clean up printReplicands()
- [ ] are we ok with slips in comm counts?
- [ ] update documentation
- [ ] mpi module OK?
- [ ] hdfs module OK?


bradcray:
Closing in favor of #5996

PR #5940, opened 2017-04-04
Fix a linux/darwin difference in DateTime module

daviditen:
The "%Y" (year) format specifier for function strftime zero-pads to 4 digits
on darwin, but doesn't zero-pad on Linux. Explicitly zero-pad it to 4 digits
in the isoformat function.

Removed two .notest files that were added due to this difference.

Changed a call to the non-existent gmtoff() into a call to utcoffset().

Tested on the datetime unit tests on Linux64 and darwin.

### DIFF
--- a/modules/standard/DateTime.chpl
+++ b/modules/standard/DateTime.chpl
@@ -1085,7 +1085,10 @@ module DateTime {
                (if minutes < 10 then "0" + minutes: string else minutes: string);
     }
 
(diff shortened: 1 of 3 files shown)


gbtitus:
I took a quick look and this seems fine to me.  👍

daviditen:
Thanks @gbtitus !

PR #5941, opened 2017-04-04
Add a test of a known chpldoc failure with multiline comments

lydia-duncan:
chpldoc intentionally grabs multiline comments that precede a symbol, no matter
the intervening space present.  However, if there is an intervening single line
or otherwise non-documentation comment, it probably shouldn't assume that the
earlier multiline comment is relevant.

This was reported ages ago, I believe when we were documenting a library module.
Apparently I forgot to file a future when I created the original issue, so I am
fixing that oversight now.


lydia-duncan:
Adds a test of #5943

Issue #5267, opened 2017-01-27
Records declared in unused functions cause internal compiler errors


daviditen:
## Bug Report

Records declared in unused functions cause internal compiler errors

### Summary of Problem

Declaring regular records in unused functions causes an internal compiler error.  Declaring extern records in unused functions causes a different internal compiler error.  If the function is used, the errors go away.

### Steps to Reproduce

`chpl test/types/records/diten/unusedRecordInProc.chpl`
`chpl test/types/records/diten/unusedExternRecordInProc.chpl`

### Configuration Information

- Output of `chpl --version`: chpl Version 1.14.0.495f500

daviditen:
Fixed with PR #5952.

PR #5952, opened 2017-04-05
Fix a problem with removing unused functions at the end of resolution

daviditen:
When an unused function is removed from the AST at the end of resolution,
if it had a record defined in it, the record definition gets removed as
part of the body of the function. The autoCopy, autoDestroy and refType
for the removed type get left behind and continue to refer to the removed
type.

Before removing an unused function at the end of resolution, walk the
function's body looking for types defined within. If any are found, remove
the autoCopy, autoDestroy, and refType if they exist. Also remove the
autoCopy and autoDestroy functions for the refType.

Removed .future files for types/records/diten/unusedRecodInProc and
types/records/diten/unusedExternRecordInProc.


daviditen:
Tested on linux64 with --verify

daviditen:
This PR should fix issue #5267.

PR #5956, opened 2017-04-05
File a future for an internal error in insertWideRefs / task intents

bradcray:
Something about this test makes the compiler go off the rails and
trigger an internal error.  The key idiom is:

````chapel
coforall loc in Locales do
  on loc do
    forall ra in RA do
      ra = here.id;
````

Changing the 'forall' into a whole-array operation works around the
problem.

Here, I've set up the .good files (and arguably the test itself) to
reflect the proposed new ReplicatedDist behavior being introduced in
PR #5933.  However, the internal error seems to occur with either
master or that branch...  (so I'm gambling that PR #5933 is merged
before this is but would be happy to be proven wrong).


bradcray:
This demonstrates issue #5957

PR #5962, opened 2017-04-06
Add '(pre-release)' label to --version output with SHA

bradcray:
For some time now, we've been discussing bumping our compiler
version number just after a release rather than just before it.
In doing so, we wanted to clearly label the version as being
a pre-release version.  This mod notes that this condition is
met whenever we print the SHA of the GitHub repo and simply
extends the logic that prints the SHA with a print of "(pre-release)".
It also updates the script that creates the .good file for our
test of the version number similarly.

### DIFF
--- a/compiler/main/version.cpp
+++ b/compiler/main/version.cpp
@@ -26,6 +26,6 @@ void
 get_version(char *v) {
   v += sprintf(v, "%d.%s.%s", MAJOR_VERSION, MINOR_VERSION, UPDATE_VERSION);
   if (strcmp(BUILD_VERSION, "0") != 0 || developer)
-    sprintf(v, ".%s", BUILD_VERSION);
+    sprintf(v, ".%s (pre-release)", BUILD_VERSION);
 }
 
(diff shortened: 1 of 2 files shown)


benharsh:
I think we should indicate some kind of version number in there like "1.16 pre-release", just to indicate a range of possible SHAs.

bradcray:
I don't understand what you're suggesting.

Just to make sure the change is clear, what I think I've implemented is that for a release version of the compiler ``chpl --version`` gives you:

````
chpl Version 1.15.0
````

whereas with a non-release version, you would get:

````
chpl Version 1.15.0.54e8a05 (pre-release)
````

What change to this are you suggesting?

benharsh:
I sorta misunderstood your proposal, but I think my core concern remains. With your sample output, I still might wonder 'is this before or after 1.15 release?'. But this may be better for a separate topic (the sort-of ongoing 'should we increment right after releasing' discussion).

bradcray:
Do you think a different organization of the output would add clarity?  Say:

````
chpl Version 1.16.0 pre-release (54e8a05)
````

Issue #5963 is the umbrella issue for this PR / deciding how/when we bump the version number.

benharsh:
That format looks good to me.

bradcray:
Updated based on feedback

awallace-cray:
This PR as of dfbccff2 is included in PR #5992.

bradcray:
Subsumed by PR #5992

PR #5964, opened 2017-04-06
Fix typos

cassella:
Fix assorted typos

### DIFF
--- a/compiler/include/flags_list.h
+++ b/compiler/include/flags_list.h
@@ -84,7 +84,7 @@ symbolFlag( FLAG_CONST , npr, "const" , "constant" )
 symbolFlag( FLAG_CONST_DUE_TO_TASK_FORALL_INTENT , npr, "const due to task or forall intent", ncm )
 symbolFlag( FLAG_C_PTR_CLASS , ypr, "c_ptr class" , "marks c_ptr class" )
 symbolFlag( FLAG_CONSTRUCTOR , npr, "constructor" , "constructor (but not type constructor); loosely defined to include constructor wrappers" )
-symbolFlag( FLAG_COPY_MUTATES , ypr, "copy mutates" , "the initCopy funciton / copy initializer takes its argument by ref")
+symbolFlag( FLAG_COPY_MUTATES , ypr, "copy mutates" , "the initCopy function / copy initializer takes its argument by ref")
 symbolFlag( FLAG_DATA_CLASS , ypr, "data class" , ncm )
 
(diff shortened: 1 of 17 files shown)


cassella:
The string containing "funciton" in flags_list.h does not appear elsewhere in the tree.

cassella:
This plus the latest dictionary update PR leaves these two:

```
compiler/include/resolution.h:131: 'eturn' not found in dictionary (from token 'eturn')
compiler/include/resolution.h:185: 'eturn' not found in dictionary (from token 'eturn')
```

which are

```
fortytwo@magrathea:~/src/chapel (typos)$ grep -iIr \\\\return . | grep -v third-party
./compiler/include/resolution.h:   * \return If a valid alignment was found.
./compiler/include/resolution.h:   * \return A constant reference to this disambiguation context.
./compiler/resolution/functionResolution.cpp: * \return True if fn1 is a more specific function than f2, false otherwise.
./compiler/resolution/functionResolution.cpp: * \return The result of the disambiguation process.
./compiler/util/clangUtil.cpp:    // \returns true to continue parsing, or false to abort parsing.
```

Which looks like a documentation directive.  Except that it only
appears (outside of third-party) those five times.  (And I must have
added it to some dictionary to keep it from being reported for the
other three.)  So I wasn't sure if it was intentional.

(scspell parses the "\r" as whitespace, so "\return" as "eturn".)


On the other hand, now that I look for it, "\param" appears a lot more
than that.  And I must have added "aram" to the main dictionary.  So
maybe I should just add "eturn" too.

lydia-duncan:
Do you know what style of documentation directive it comes from?  There's a part of me that thinks letting the spellchecker ignore missing the first letter of these words will just mean we don't catch it when someone actually makes that typo, so I would be tempted to change those comments to not trigger rather than adding to the dictionary, if it isn't too much trouble

lydia-duncan:
I'm running a paratest over this change out of paranoia, but should it come back clean I think this will be ready to merge

cassella:
Yeah, that's why I didn't want to add "eturn" to the dictionary.

I don't know what the style is, or if the ones in the Chapel repo are used for anything, which is why I didn't want to just go ahead and change them.  :)

cassella:
``\param`` and ``\return`` are consistent with doxygen, whose commands may start with either a ``\`` or an ``@``.
https://www.stack.nl/~dimitri/doxygen/manual/commands.html

I've shied away from modifying the source for the convenience of my
interloping spellchecking.

PR #5967, opened 2017-04-06
Fix RVF bug involving a dereference

benharsh:
Previously RVF would transform AST like this (a deref-move):
  ``(move myVar myRefArg)``
into this:
  ``(move myVar (addrof myRefArg))``
Which is clearly wrong. RVF wasn't being made aware of the deref-move
pattern, so this commit adds a case to handle it.

This pattern could later trip up other passes that expect the AST
to adhere to certain rules/conventions.

Testing:
- [x] full local
- [x] full gasnet


benharsh:
This will be merged once the upcoming ReplicatedDist changes are in.

mppf:
Looks good to me. Thanks!

benharsh:
@bradcray : testing came back clean with your recent replicated changes, and the ``test/distributions/replicated/repl-int-error.chpl`` future will be fixed by this PR.

Clear to merge?

PR #5968, opened 2017-04-06
Cache remote testing updates

mppf:
* Updates sub_test to pass expected .good file to prediffs. Since the prediff can create a .good file (and cause a different one to be selected) we now compute the .good file twice; once before running the prediff and once after.
* Updates sub_test to look for .cache-remote.good and .llvm-wide-opt.good files since these can have different comm counts.
* Adds a prediff to tolerate some variance in comm counts under --cache-remote. Naturally, this uses the .good file that is now passed to prediffs by the above change.

* Adjusts tests printing comm counts that differ under --cache-remote to either skip or have a different .good file.

Variance in comm counts for some tests when using --cache-remote could be due to cache-remote's prefetching and sensitivity to where memory is stored. If a different program run has a different memory layout, whether needed data is gotten by reading a cache-line at a time or not can change.

- [x] full local testing
- [x] full GASNet testing
- [ ] full GASNet testing with --cache-remote


mppf:
@ronawho - could you let me know if you think the approach in this PR for enabling clean --cache-remote testing is reasonable?

ronawho:
To me --cache-remote seems like any other optimization, so I don't understand why we would want specific .good files it. i.e. we don't have --no-rvf, or --baseline .good files so why would we want --cache-remote ones?

To me it seems like we should either make --cache-remote the default (and update all the .good files) or leave things as they are.

That said, I understand that in reality --cache-remote is different than a traditional compiler optimization, but I would still prefer to avoid adding extra .good files if we can.

My main concerns here are that we'll never remember to update these .good files if we don't have full nightly testing, and if we do have full nightly testing, there's still a lot more .good files to have to update when comm counts change.

mppf:
@ronawho - I'm making these changes in response to our last discussion about --cache-remote. I asked what it would take to have it be enabled for our reported performance results and you asked for "clean --cache-remote testing". I also understood that "clean --cache-remote testing" would be a prerequisite for turning it on by default.

The reason we don't have --no-rvf or --baseline .good files for comm counts is that we don't ever test those configurations in nightly testing with multi-locale. It seems to me we need to be able to test --cache-remote nightly in multi-locale as part of making it a reported configuration or eventually making it the default.

Personally, I've never like the comms counts correctness testing anyway. I think they should be re-cast as performance graphs rather than correctness tests. They pre-dated the nice performance graphs we have now. I think it's a mistake to create work for ourselves when the comm counts improve, for one thing.

ronawho:
>@ronawho - I'm making these changes in response to our last discussion about --cache-remote. I asked what it would take to have it be enabled for our reported performance results and you asked for "clean --cache-remote testing". I also understood that "clean --cache-remote testing" would be a prerequisite for turning it on by default.

Ok, can you update the PR message to include the motivation/context? Also does 'elliot asked for "clean --cache-remote testing"' mean you don't think clean testing is a requirement?

>The reason we don't have --no-rvf or --baseline .good files for comm counts is that we don't ever test those configurations in nightly testing with multi-locale. It seems to me we need to be able to test --cache-remote nightly in multi-locale as part of making it a reported configuration or eventually making it the default.

What's the plan? Add a full --cache-remote testing config and work towards making it the default? Once it's the default can we basically revert all these changes?

I had imagined this as: 
 - work toward getting all correctness testing passing in a branch
 - do some multi-locale performance testing, and if we're happy with the results make it the default and update all comm counts.

But it's not clear to me what your path forward is so i'm having trouble figuring out if this is a good next step.

mppf:
Hi @ronawho -

For me, "clean testing" means having the comm counts report the correct numbers. That means we have to deal with the comm count changes somehow. It sounds like you expected me to directly update the .good files in a branch, but it's not yet obvious to me (until we do some more work) that --cache-remote is ready to be on by default. It is obvious to me that we should test it more. So I took an approach that allowed us to separate the testing changes from the decision about whether or not it should be on by default.

But even if it were on by default, wouldn't we still want to enable full testing to run correctly with "--no-cache-remote"? For that matter, shouldn't a full run of "--no-rvf" report no testing failures today? (I'm pretty sure we would get failures from comm counts).

I'd prefer to migrate the comm counts tests over to the performance test infrastructure. Do you think that is a good idea, a bad idea, too much work for now, ... ?


Anyway, to answer "What's the plan"? My plan for this PR was:
 1. enable clean testing of both --cache-remote and --no-cache-remote
 2. start regular testing of --cache-remote, including performance testing
 3. investigate any performance bumps compared to --no-cache-remote (and note that this might take months, depending)
 4. decide whether or not --cache-remote can be the default yet; do 3 again if not

ronawho:
I wouldn't expect --no-rvf, --no-loop-invariant-code-motion, or -no-<blah> testing to pass the comm count tests, and that's my point here. I don't really want to start special casing comm count tests for various optimizations. Arguably that means the comm count tests are fragile and we should convert them to perf tests or something (maybe make an issue for that?)

I'd prefer to see this work done on a branch, and then we can just setup some jenkins jobs to run multi-locale correctness against your branch (like we did with the string-as-rec and maybe array-as-rec work.)

mppf:
new issue here: #6136

I'll close this PR for now and keep updating .good files (with its current framework, but that is temporary). And run some local testing. At some point, we'll set up a Jenkins job to run multi-locale with --cache-remote.

mppf:
Tests to look in to from quickstart + gasnet + cache remote run:
memory/ferguson/test1.chpl
modules/packages/ZMQ/multilocale/hello-multilocale.chpl
modules/packages/ZMQ/multilocale/req-rep.chpl
parallel/taskPar/sungeun/barrier/commDiags.chpl
performance/hilde/userSync.chpl
performance/sungeun/multilocale/syncsingle.chpl
regexp/elliot/I-fail-without-re2.chpl
studies/hpcc/CommDiags/stream-ep_commDiags.chpl

PR #5969, opened 2017-04-06
Improve wide-pointer analysis for const-ref args

benharsh:
If an argument has the const-ref intent and points to a wide symbol, we
have historically widened the actual at each callsite. This doesn't
quite make sense because the const-ref intent says that the function
cannot modify the actual. In other words, a local thing passed to a
const-ref formal (of whatever wideness) should remain local.

A change was made to 'insertWideTemp' to handle a new AST pattern.

Testing:
- [x] full no-local
- [x] full gasnet

### DIFF
--- a/compiler/passes/insertWideReferences.cpp
+++ b/compiler/passes/insertWideReferences.cpp
@@ -1884,9 +1884,8 @@ static void insertWideTemp(Type* type, SymExpr* src) {
   if (type->isRefOrWideRef() &&
       !src->isRefOrWideRef()) {
     needsAddrOf = true;
-  }
-  else if (src->isRefOrWideRef() &&
-      !type->isRefOrWideRef()) {
+  } else if (src->isRefOrWideRef() &&
+             !type->isRefOrWideRef()) {
     needsDeref = true;
   } else if (type->isRef() && src->isRef() &&
              type->getValType()->symbol->hasFlag(FLAG_WIDE_CLASS) &&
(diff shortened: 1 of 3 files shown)


benharsh:
For some context, this squashes some overhead in the 'bucketize' stage of ISx. It doesn't help a ton, but is strictly better.

compiler/passes/insertWideReferences.cpp
```diff
@@ -1878,15 +1878,21 @@ static void insertWideTemp(Type* type, SymExpr* src) {
...
+  bool widenVal    = false;
 
   if (type->isRefOrWideRef() &&
-      !src->isRefOrWideRef())
+      !src->isRefOrWideRef()) {
     needsAddrOf = true;
-
-  if (src->isRefOrWideRef() &&
-      !type->isRefOrWideRef())
+  }
```
noakesmichael: Tiny niggle that you can ignore.  To my eye the brace convention on this else if
is inconsistent with very next else-if.  Brace style is somewhat personal but I
think it's good to be consistent.

compiler/passes/insertWideReferences.cpp
```diff
@@ -1895,14 +1901,33 @@ static void insertWideTemp(Type* type, SymExpr* src) {
   Expr* rhs = src->copy();
   if (needsAddrOf)
     rhs = new CallExpr(PRIM_ADDR_OF, rhs);
-  if (needsDeref) {
+  else if (needsDeref) {
```
noakesmichael: Applause.  I like that it's clearer that needsAddrOf and needDeref can't be simultaneously true.
As a footnote this makes it seem as if an enum might be a good choice one day if this logic becomes
more confusing.

PR #5970, opened 2017-04-06
Migrate chpldoc testing directory to rst

### LINKED ISSUE #5951: Convert chpldoc tests to match against generated .rst
Though we lean more strongly on the .rst and .html output generated from chpldoc through Sphinx, our testing of chpldoc still compares against the old .txt output for the most part.  It would be nice to update most of the expected tests to match against .rst (which is nicer to look at than .html), with a couple to verify the .txt output so long as we maintain it.  Unfortunately, this work is pretty thankless, which is why we've put it off so long.

This impacts the tests in the test/chpldoc directory

lydia-duncan:
Resolves #5951 

This updates the chpldoc test directory to match against our generated rst, the basis for the
html, which our current uses of chpldoc rely on most heavily.  It modifies the CHPLDOCOPTS
files in each directory (to use `--no-html --save-sphinx docs` instead of `--text-only` during
testing), updates the .catfiles to look in the new location for the generated files, and uses the
new output as the .good file for the test.

As part of this PR, I also copied the `chpldoc/basics` directory over to
`chpldoc/compflags/textonly`, to ensure the old set up still works until such time as we decide
to retire it.

I spot-checked the new .goods for six of these directories, as part of verifying that my script was reasonable.  I would like to verify the other directories later as the mood strikes me but
agree that doing so is too much effort over all.

TODO: Migrate
- [x] chpldoc/
- [x] chpldoc/basics/
- [x] chpldoc/classes/
- [x] chpldoc/compflags/comment
- [x] chpldoc/compflags/externs
- [x] chpldoc/compflags/folder
- [x] chpldoc/enum
- [x] chpldoc/globals
- [x] chpldoc/linkage-spec
- [x] chpldoc/module
- [x] chpldoc/nodoc
- [x] chpldoc/types
- [x] chpldoc/types/arguments
- [x] chpldoc/types/config
- [x] chpldoc/types/fields
- [x] chpldoc/types/returns

Decided not to update the `chpldoc/compflags/folder/save-sphinx`, and `chpldoc/compflags/creole` folders because they test unusual configuration settings that are distinct from the output we are updating to.

Decided not to update the NOTEST directories `chpldoc/compflags/alphabetical` and `chpldoc/devel` - these directories would require additional effort to update, since we don't currently support the features they are testing - writing a script to convert them over would be either the same effort to implement the missing functionality or as what chpldoc does under the covers.  They should be updated when the feature is implemented, to save effort should we never decide to implement those features.


lydia-duncan:
I used the following script to make the changes, with the exception of copying the PREDIFF and modifying the CHPLDOCOPTS file (the latter of which needed to be updated on a case by case basis)

```
#!/bin/bash

for i in *.catfiles; do
  sed -i -e 's/docs\//docs\/_sources\/modules\//g' $i
done

for i in *.doc.good; do
  cp "$i" "${i/.doc.good/.txtmode.doc.good}"
done 

for i in *.logical.doc.good; do
  rm "$i"
done

start_test

for i in *.comp.out.tmp; do
  mv "$i" "${i/.comp.out.tmp/.good}"
done

```

lydia-duncan:
@bradcray - mind sanity checking my approach before I apply it to all the other directories?

bradcray:
This is generally what I was hoping for, though I'll add a couple of questions inline.

test/chpldoc/linkage-spec/exportTest.doc.catfiles
```diff
@@ -1,2 +1,2 @@
-docs/exportTest.doc.txt
-docs/exportTest.doc/Other.txt
+docs/_sources/modules/exportTest.doc.txt
+docs/_sources/modules/exportTest.doc/Other.txt
```
bradcray: What determines this _sources/modules directory?  Is there a(n easy) change we could make to how chpldoc testing works that would cause it to use the old docs/ directory?

Also: I was surprised that the extension on the file was .txt rather than .rst.  Is that our choice or something else's?

bradcray:
I can't comment on the change directly, but I'd be inclined to throw out the old text .good files rather than preserving them.  Seems like they'd just add clutter and confusion.

bradcray:
Since he has a vested interest in documentation (and probably more cycles than me), I'd like @ben-albrecht to give this a look at some point as well.

test/chpldoc/linkage-spec/exportTest.doc.catfiles
lydia-duncan: We still get the docs directory.  Because I'm not throwing anything explicitly, this is also generating html, but it is saving the source files it used for that html into the subdirectory `_sources/modules`.  I believe it is our choice for that extension to be .txt and it is probably not difficult to change, but I refrained in this PR in an effort to keep it a "pure test" change.

lydia-duncan:
Thanks for commenting on the old text .good files.  I wasn't sure how I felt about them either - on the one hand, they will likely go out of date; on the other hand, they're a touchstone of what we think should and shouldn't be documented and so would be useful if/when we ever get back to verifying the new expected output.

test/chpldoc/linkage-spec/exportTest.doc.catfiles
lydia-duncan: We don't have an explicit flag to only generate rst for chpldoc at this time

bradcray:
> so would be useful if/when we ever get back to verifying the new expected output.

We can always go back into the GitHub repo to get them for that purpose, though.

lydia-duncan:
True, true.  Okay, I will delete them, then.  Thanks!

ben-albrecht:
This approach looks good to me. 

A side note: We could speed these tests up significantly by dropping the html-generation and only generating rst with the flags: `chpldoc --no-html --save-sphinx rst foo.chpl`. 

However, I think there is value in generating html to ensure no errors occur on that step as well. That's assuming `start_test` properly flags a failed sphinx-build as a failure for the test, probably by checking the return code - do either of you happen to know if this is the case?

lydia-duncan:
I don't know if that is the case, but I suspect that my prediff might get rid of Sphinx error output if it occurs . . . the normal Sphinx output includes exact path locations and I didn't think we should keep or maintain that.

I think it would be worth investigating the flags you suggest

bradcray:
> However, I think there is value in generating html to ensure no errors occur on that step as well. 

I could imagine counting on the generation of the entire docs/ hierarchy to serve as sufficient coverage for making sure html generation didn't generate errors if we wanted to avoid generating things we don't need here.  But I don't feel strongly about this either way.

ben-albrecht:
As a rough estimate here's the difference in chpldoc times for a file with 1 function:

|  command | time (ms) |
|---------|----------------|
| `chpldoc foo.chpl` | 655      |
| `chpldoc --text-only foo.chpl` | 67 |
| `chpldoc --no-html --save-sphinx foo-doc foo.chpl`                     | 73 |

If we're OK with the module-build confirming the HTML-generation, then I'd be in favor of throwing the suggested flags to avoid increasing our chpldoc test time by an order of magnitude.

Also, @lydia-duncan's suspicion was correct - the HTML-generation error is not caught by `start_test` anyway.

lydia-duncan:
Thanks for looking into that, @ben-albrecht!  I'll definitely update with your recommendations

lydia-duncan:
Okay, I believe this PR is complete.  Would one of you mind sanity checking the PR message?  I've updated it for the new strategy and to denote which directories are and are not getting updated as part of this effort (and why)

lydia-duncan:
(I'm tempted to remove the creole directory entirely, but I don't think this PR is the place to do it)

bradcray:
PR message looks OK to me.

PR #5971, opened 2017-04-06
CLBG diff tool

ben-albrecht:
This PR does the following:

- Added a diff tool, `clbg-diff.py`, to our submitted shootouts to easily check if our versions have diverged from the CLBG website's versions.
- Renamed some files to reflect the naming scheme used by CLBG
    - The new names do not reflect the CLBG names exactly, but are consistent enough, that the diff-script can map our format to theirs (see `clbgify_name()`)
    - `.dat` files are left alone because I don't want to mess with their mirrors on the data repository...
- Added a `.gitignore` to ignore the checked out cvs repo.

[clbg-diff.py output](https://gist.github.com/ben-albrecht/d01180145eec4f23c66f6b97a65ae173)

- Note that we are missing `fastaredux.chpl` because it has been retired.


ben-albrecht:
Anyone from @chapel-lang/perf-team would be a suitable reviewer I think.

benharsh:
Looks reasonable to me 👍

ben-albrecht:
Thanks for review @benharsh - will merge now. 

Dropping the following comment from the PR message, in case anyone from @chapel-lang/perf-team has an opinion:

> should we update our submitted versions to reflect the changes shown in the diff?

- On the + side, it'd be easier to spot new diffs, if we maintain an empty diff.
- On the - side, it's another thing to maintain..

bradcray:
I'm not understanding the "thing to maintain" comment -- it should be a one-time "bring things into sync" change shouldn't it? (assuming he doesn't continue to mess with our spacing)

ben-albrecht:
@bradcray Good point.

> (assuming he doesn't continue to mess with our spacing)

If we can safely make that assumption, then I'd say it's good to bring them in sync.

bradcray:
I'm guessing that, other than updates due to changes in the ruels, they've only been updated when brought into the system, not thereafter.

I also owe a binarytrees update based on his merge of my new code late last week...

Issue #5909, opened 2017-03-30
Chpldoc error with --comment-style flag and non-matching closing comment


lydia-duncan:
### Summary of Problem
When Chapel code is documented with chpldoc, if the --comment-style flag is used, no verification is performed to ensure that the signifier of the end of the comment matches the start of the comment.  We would like to ensure that all custom comment styles are closed in the reverse order of the starting sequence.  We might also consider adding a flag to specify different opening and closing signifiers, but that would be a bonus feature.

### Steps to Reproduce

**Source Code:**

```chapel
/** This comment is not closed properly */
proc bad() {

}
```

**Compile command:**
`chpldoc --comment-style /** badClose.doc.chpl`

**Associated Future Test(s):**
chpldoc/compflags/comment/badClose.doc.chpl

### Configuration Information

- Output of `chpl --version`:
Chapel 1.10.0 and on (still present as of Chapel 1.15.0)

Shashi456:
how can i go about doing this ?

lydia-duncan:
The handling of chpldoc flags can be found in compiler/main/docsDriver.h.  Looking in there should allow you to determine the global variable used for determining if a block comment is a chpldoc comment (I recommend doing a search on the flag minus the leading "--" and then reading the description above the data structure which that will lead you to, in order to determine which of the components of that record is used).

The detection of whether an ordinary block comment or a chpldoc comment is being used is handled at parse time.  Doing a search in the compiler for the variable you found in the last step will lead you to two potential files - one of them is generated from the other.  The file you want to modify is compiler/parser/chapel.lex.  Once you have made modifications there, you will need to do a `make parser` in the directory containing that source in order to actually see your changes take effect when you rebuild the compiler (a separate step).

saru95:
Is this fixed?

lydia-duncan:
https://github.com/chapel-lang/chapel/pull/5973 is resolving this issue

PR #5973, opened 2017-04-07
fixing non-matching closing comment

krishnakeshav:
I am trying to fix issue #5909 . What current changes do is that it simply returns an error. `error : unterminated comment started here.` on mismatch start and end signifiers. I would like to know what you guys prefer - 

1.  A separate error message and termination.

2. Or no compilation error but warning like - `comment is not closed properly`.


lydia-duncan:
Say we have this code:
```
/** This comment is not closed properly */
proc bad() {

}
```
Under case 2, would this generate a warning but otherwise act as though the `*/` properly closed the comment?

Consider this code: 
```
/*^& This comment is not closed properly &^/
proc bad() {

}
```
Would this code generate the warning?  Or would we just always fall to the end of the file being parsed and run into the error you currently get?

lydia-duncan:
When the future was added, we believed the right course of action was to warn about the comment not being closed properly, and to then not include it as a documentation comment.  There's some cases an improperly closed comment could fall under:

1. It could be closed with the normal '*/'
2. The sequence intended to close it could be similar to the opening sequence but with a small difference (i.e. a character was forgotten)
3. The sequence intended to close it was unrecognizable from the value provided to --comment-style (maybe multiple characters were missing)

I think that cases 1 and 2 should generate a warning, and that case 3 should be treated as though no end sequence was provided and generate the error you see with your change so far

krishnakeshav:
As discussed, I have added warning to comment closing with ***/** and mismatch up to 1 character. I request you to review code as well as if any further possible improvisations.

compiler/parser/chapel.lex
```diff
@@ -751,8 +751,13 @@ static int processBlockComment(yyscan_t scanner) {
     else
      d = 1;
 
-    if (lastc == '*' && c == '/' && d == len + 1) { // close comment
-      if(fDocs && d == len + 1) {
+    if (lastc == '*' && c == '/') { // close comment
+      if(fDocsCommentLabel[0] == '/' && fDocsCommentLabel[1] == '*' && (d < len + 1 && d >= len - 3)) {
+        USR_WARN("chpldoc comment not closed, ignoring comment: %s This comment is not closed properly", fDocsCommentLabel);
```
lydia-duncan: If you look later in this if/else branch, you'll see that we usually output the line number in our error messages.  Would you mind doing that here?

krishnakeshav: Sure! I will.

krishnakeshav: Done!

lydia-duncan:
From my start_test run over this branch:

[Test Summary - 170410.164958]
[Error matching compiler output for chpldoc/multifile.doc]
[Error matching .bad file for chpldoc/compflags/comment/badClose.doc]
[Error matching compiler output for chpldoc/compflags/comment/loseComments.doc]

For the .bad failure, you will either need to update the .good file on the future to match your expected output, or (probably the better course, if possible) update the error message to more closely match the one given in the .good file.  But that's independent of the current failure, which is an internal segfault

lydia-duncan:
You can verify this output by running start_test -futures over the test/chpldoc directory

krishnakeshav:
I made some changes in future tests. Apart from that what I am now concerned with is test in file loseComments.doc.chpl.
`/* This comment should be lost */
proc found () {

}` 
**compile command : chpldoc --comment-style '/**' loseComments.doc.chpl**
I interpreted that the requirement is to disallow  such comments, where label and commenting style in file are different.

lydia-duncan:
The `--comment-style` feature was intended to allow a user to be more selective in the comments that get included for documentation, not to disallow ordinary multi-line comments.  Basically, this means that if a user wishes to have a long comment ignored, they don't have to preface every line with `//`.

compiler/parser/flex-chapel.cpp
```diff
@@ -3508,20 +3508,24 @@ static int processBlockComment(yyscan_t scanner) {
...
-        d = 1;
+      if(fDocs) {
+        if(d == len + 1) {
+          depth--;
+          d = 1;
+        }
+        else if(fDocsCommentLabel[0] == '/' && fDocsCommentLabel[1] == '*') {
+          depth--;
+          d = 1;
+          USR_WARN("chpldoc comment not closed, ignoring comment: %s in Line no: %d\nThis comment is not closed properly", fDocsCommentLabel, startLine);
```
lydia-duncan: I would prefer this to print out the comment that is bad, rather than the doc label being used, as I believe that will be more useful to the user

lydia-duncan: (Sorry, wrong file)

test/chpldoc/compflags/comment/loseComments.doc.good
```diff
@@ -1,6 +1,2 @@
-loseComments.doc
-
-Usage:
-   use loseComments.doc;
-
-   proc found()
+error: improper comment style
+cat: docs/loseComments.doc.txt: No such file or directory
```
lydia-duncan: As I mentioned in the discussion, I would like this test to be unimpacted

compiler/parser/chapel.lex
```diff
@@ -762,7 +762,8 @@ static int processBlockComment(yyscan_t scanner) {
         else if(fDocsCommentLabel[0] == '/' && fDocsCommentLabel[1] == '*') {
           depth--;
           d = 1;
-          USR_WARN("chpldoc comment not closed, ignoring comment: %s in Line no: %d\nThis comment is not closed properly", fDocsCommentLabel, startLine);
+          warning = true;
+          //USR_WARN("Bad comment: %s\nLine no: %d\nThis comment is not closed properly", stringBuffer.c_str(), startLine);
```
lydia-duncan: Hmm, maybe it would be better to match the error style of the rest of the file, rather than the general USR_WARN that we use in the rest of the file.  See line 782 in this file

test/chpldoc/compflags/comment/badClose.doc.bad
```diff
@@ -1,4 +1,5 @@
-warning: chpldoc comment not closed, ignoring comment: /** in Line no: 1
+warning: chpldoc comment not closed, ignoring comment: * This comment is not closed properly
+Line no: 1
```
lydia-duncan: Two things:
1) Do you really want to be concatenating the comment style characters after the opening `/*` into the comment that is output?
2) The result of this PR should be a clean match against the .good file for this test, meaning that the .bad file is no longer necessary and that it and the .future can be removed

krishnakeshav: Okay. into it.

krishnakeshav:
Latest patch implements the check for bad comment. It generates a warning and ignores the bad comment.

compiler/parser/chapel.lex
```diff
@@ -815,16 +809,20 @@ static int processBlockComment(yyscan_t scanner) {
...
     }
+    if(!badComment)
+      yyLval->pch = astr(wholeComment.c_str());
+    else {
+      
+      ParserContext context(scanner);
 
-    yyLval->pch = astr(wholeComment.c_str());
-
+      fprintf(stderr, "Warning:%d:chpldoc comment not closed, ignoring comment:%s\n",
```
lydia-duncan: Sorry, this is nearly perfect!  Could you add a space between the line number and the start of the error message?  So it will be "Warning:%d: chpldoc" instead of "Warning:%d:chpldoc".

Be sure to update the .good one more time, and run start_test over that future to make sure it works!  Then if it says the future passes, you should remove the .bad and .future file.  I will run testing just to be certain nothing broke, but I suspect this will then be good to merge!

krishnakeshav: Done!

lydia-duncan:
You're the best!  Running a full standard paratest now

lydia-duncan:
Full testing passed!

Issue #5523, opened 2017-03-06
Cleanup and simplify Cray platform README once muxed is retired


ronawho:
doc/platforms/cray.rst is pretty long and IMO overly specific. It could use some cleanup, but since a fairly large portion of it is dedicated to the muxed tasking layer, I don't really want to clean it up before muxed is retired. 

This issue is just a placeholder so I don't forget to clean it up after the 1.15 release.

ronawho:
#5975 removes muxed, once that is merged, I think we can/should overhaul cray.rst

ronawho:
@gbtitus I think we should get to this during doc-freeze, especially since there will probably be dynamic registration related things to document

gbtitus:
This might be done now, but someone (you know who you are) should take a run through it to make sure.

ronawho:
Between https://github.com/chapel-lang/chapel/pull/7419 and https://github.com/chapel-lang/chapel/pull/7438 this is done

PR #5975, opened 2017-04-07
Retire muxed tasking

ronawho:
Remove all references to muxed tasking, it was deprecated for 1.15 and has
been officially retired now.


test/studies/shootout/thread-ring/README
```diff
@@ -1,26 +0,0 @@
...
-
-Other changes tried: altering chpl_task_alloc calls in the c code to
-chpl_malloc calls caused little to no improvement on its own, but when combined
-with deletions of extraneous _ic_these_DefaultRectangularDom_1_int64_t_F alloc's
-and free's caused a fair improvement, although not as much as we would like.
-Utilizing writeXF instead of writeEF also seemed to give a slight improvement.
-
-Skipping testing under valgrind because the current settings do not allow enough
-threads to complete even under the (comparatively) low number of threads for
-that sample size
```
cassella: OOC, why remove this file?  Through this point, it doesn't mention muxed.  And there's still a valgrind SKIPIF file.

ronawho: I initially just started updating the file, but decided to just remove it since it's largely out of date and I don't think provides much value. For instance: 

 - the coforall+begin version is no longer the fastest 
 - chpl_task_alloc isn't a thing anymore 
 - there are no instances of _ic_these in the generated code anymore
 - the timings for qthreads/muxed are out of date

The only current info is about skipping under valgrind, and I don't think that's useful enough info to warrant keeping the readme around

compiler/passes/parallel.cpp
```diff
@@ -1037,12 +1037,10 @@ freeHeapAllocatedVars(Vec<Symbol*> heapAllocatedVars) {
 //  CHPL_COMM == "gasnet" && CHPL_GASNET_SEGMENT == "everything";
 // or
 //  CHPL_STACK_CHECKS == 0 && CHPL_TASKS == "fifo"
-// or
-//  CHPL_STACK_CHECKS == 0 && CHPL_TASKS == "muxed"
 //
 // true otherwise.
 //
-// The tasking layer matters because fifo and muxed allocate
+// The tasking layer matters because fifo allocates
```
gbtitus: This sentence doesn't make sense.  It didn't before either, though.  Whether or not it needs fixing is up to you.

doc/rst/platforms/cray.rst
```diff
@@ -535,9 +510,7 @@ When hugepages are used with the ugni comm layer, tasking layers
 cannot use guard pages for stack overflow detection.  Qthreads tasking
 can only use guard pages for stack overflow detection, so if ugni
 communications is combined with qthreads tasking, overflow detection is
-turned off completely.  Muxed tasking can use guard pages for stack
-overflow detection, but it can also drop back to synchronous overflow
-detection, as described below, with ``CHPL_COMM=ugni`` and hugepages.
+turned off completely.
```
gbtitus: Not strictly true any more.  With comm=ugni, if you run without a craype-hugepages module loaded then you're in so-called "minimal registered heap" mode, most of memory including the stacks are on regular pages, and guard pages can be created (and are, other things permitting).  This is mentioned in the big chunk removed later, but it's kind of a passing reference.

gbtitus:
My line comment about stack checks should be addressed, but the rest looks good and I don't need to review again unless you want me to look at that part.

compiler/passes/parallel.cpp
ronawho: I have a PR on a branch that updates this text and code for qthreads -- I'l include this fix there

doc/rst/platforms/cray.rst
ronawho: good point, though I'm not too interested in "improving" documentation in this PR. I'd prefer to save that for #5523

ronawho: wait actually, I don't think this is wrong. It reads:

"""
When hugepages are used with the ugni comm layer, tasking layers cannot use guard pages for stack overflow detection. Qthreads tasking can only use guard pages for stack overflow detection, so if ugni communications is combined with qthreads tasking, overflow detection is turned off completely.
"""

That section starts with "When hugepages are used with the ugni comm layer..." so I think this is only talking about what the behavior is when a hugepage module is loaded. 

I don't disagree that the documentation could be improved, but I don't think it's wrong, is it?

ronawho: Eh whatever, unless you feel like it needs to be fixed now, I'll punt until #5523

gbtitus: Yeah, okay, I didn't spot that part of it.  It's still a little misleading, I think, but certainly any improvement could wait until #5523.

gbtitus:
Ship it!   👍

doc/rst/platforms/cray.rst
ronawho: Yeah, agreed. It's definetly misleading especially when read quickly

PR #5977, opened 2017-04-08
Upgrade to dygraphs 2.0

benharsh:
Some minor changes are needed now that parts of the old API are no longer available.

### DIFF
--- a/util/test/perf/perfgraph.js
+++ b/util/test/perf/perfgraph.js
@@ -783,7 +783,9 @@ function customValueFormatter(val, opts, series_name, dygraph) {
   if (val != 0.0 && (Math.abs(val) >= Math.pow(10, maxWidth) || Math.abs(val) < Math.pow(10, -digits))) {
     return val.toExponential(digits);
   } else {
-    return val.toFixed(digits);
+    // "3" should display as "3" and not "3.00"
+    var shift = Math.pow(10, digits);
+    return Math.round(val * shift) / shift;
   }
 }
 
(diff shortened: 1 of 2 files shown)


benharsh:
@ronawho : this was easier to split off than I thought. I can send you a link to some hosted graphs if you want.

util/test/perf/perfgraph.js
```diff
@@ -732,7 +778,8 @@ function customValueFormatter(val, opts, series_name, dygraph) {
 
   // update digits, but do NOT redraw. Then use the default value formatter
   dygraph.updateOptions({digitsAfterDecimal: digits}, true);
-  return Dygraph.numberValueFormatter(val, opts);
+
+  return val.toFixed(digits);
```
ronawho: This effectively means we lose the `labelsKMB` and `labelsKMB2` options, right? I don't think we use them now so that should be ok.

Does this also mean we'll lose the auto-scientific notation switch?

ronawho:
Do we still have access to `Dygraph.dateString_`, and `Dygraph.DASHED_LINE` (and the other lines)?

ronawho:
A link to a hosted graph would be great.

benharsh:
> Do we still have access to Dygraph.dateString_, and Dygraph.DASHED_LINE (and the other lines)?

Yes, we have access to both.

util/test/perf/perfgraph.js
benharsh: Hmm, I had forgotten about the scientific notation switch. The ``toExponential`` method may be useful. Do you have an example graph in mind?

ronawho: The memory leak graphs use it. And others might if you zoom in close to the axis

benharsh:
Just pushed a commit that restores the scientific notation formatting. It should be reflected in the linked graphs.

ronawho:
hmm, the number of digits after the decimal point doesn't look quite right. Not sure why, the toFixed stuff looks ok to me

PR #5981, opened 2017-04-10
Optimized sparse bulkAdd for when there is no nonzeroes in the domain

e-kayrakli:
This PR optimizes bulk index addition to sparse domains, for the special
case where there is no nonzero indices in the domain (i.e. initial
index addition)

Specifically:
- Circumvents much of the bookkeeping that takes care of existing
  indices in the domain when `nnz == 0`
- Adds one more internal helper and adjusts existing ones.

I did some performance testing (not included in the PR).  Time to create
a scale 20 Graph500 domain, ie square matrix of size 2^20 with 16
nonzeroes per row on average:

```
                      master        this PR
presorted indices     0.55s         0.15s
unsorted indices      3.49s         3.10s
```

The behavior is not significantly different in DefaultSparse and CSR.

It seems this would be especially useful when indices are read from a
presorted file, or created rather regularly as in some benchmarks.
Otherwise, sorting the input indices becomes the bottleneck.

- [x] std linux64 test on `test/users/engin/sparse_bulk*`
- [x] GASNet+linux64 test on `test/users/engin/sparse_bulk*`
- [x] full std linux64 test


e-kayrakli:
I had discussed this with @mppf  quite a while back. He might be interested in reviewing..

mppf:
hi @e-kayrakli, this looks good to me. What testing have you done with it? Could you put TODOs in the PR message with any testing you'd like me to help with? Thanks.

e-kayrakli:
@mppf - I think full standard linux64 test should suffice. I updated the PR message accordingly.

PR #5982, opened 2017-04-11
Add new "fast" version of mandelbrot to release

bradcray:
This cleans up mandelbrot-blc.chpl and adds it to the release as a
'fast' (read: uglier but faster) variation of the benchmark.  It adds
mandelbrot-fast to the README and Makefile and to the two main
mandelbrot*.graph files.

While here, I also tried a style change to mandelbrot-blc.chpl
(zippering over the image domain and array to avoid some silly
manipulation in writing the image) to see how it affects performance.


bradcray:
@chapel-lang/perf-team : For your consideration.  I seem to recall Elliot having some objection to us using -fast.chpl for second copies of benchmarks in the release directory, but can't quite recall what it was, and since we haven't moved away from it yet, continued to lean on that convention here.

ronawho:
I'm fine going with "-fast" for now. I think my objection was that having a -fast version implies the other version is "slow", and that "-tuned" or something would be a "nicer" term for it

ronawho:
I have no objections to merging this tonight

ben-albrecht:
Not that the CLBG site simply uses counters (1,2,3..), so the submitted versions have been changed to reflect that as well. That doesn't mean we can't use `-fast` elsewhere though.

bradcray:
I'm open to "-tuned" or maybe "-opt" as an alternative to "-fast", but will wait until we decide to do that for all the copies in release rather than being the first to do so here.  I don't like the idea of just numbering the versions in the release, though I agree that's the right thing to do in the submitted directory...

bradcray:
@chapel-lang/perf-team: Having merged this, if anyone else would like to give it a close read after it passes testing yet before I submit to the site as a proposed "Chapel #2" entry, please let me know.

PR #5987, opened 2017-04-11
Attempt to make mandelbrot-fast testing more robust

bradcray:
There's a numerical instability in the CLBG benchmark that depends
on whether or not fused multiply-add is used or not.  In the past,
we've tried to call out such cases in skipif files, but this seems
fragile to maintain and also means that we don't test in such
configurations, which seems unfortunate.

This instability has hit us before (#5175) and is discussed a bit here:

  https://alioth.debian.org/forum/forum.php?thread_id=14967&forum_id=2965&group_id=100815

Here, I take a slightly different approach by committing both of the
possible .good files that we see and using a .prediff file to copy the right
one into the expected .good file's location.  The current logic is:

* if we're on the shootout box, use mandelbrot-fast-clbg.good so that
  we don't accidentally submit code that will break our official entry

* otherwise, speculatively compare against mandelbrot-fast-alt.good.
  If that passes, use that, otherwise use mandelbrot-fast-clbg.good.


bradcray:
@ronawho: Following up on our discussion, would you take a look at this?  Feel free to request stylistic changes -- I tend to avoid bash conditional, so didn't create as clean of a nested conditional as I probably could've/should've

So far, I've only tested on linux32 and my Mac.  I also need to see if $HOSTNAME is the short or long-form string on the shootout box...

test/release/examples/benchmarks/shootout/mandelbrot-fast.prediff
```diff
@@ -0,0 +1,35 @@
...
+outputfile=$2
+
+clbg_good=$testname-clbg.good
+alt_good=$testname-alt.good
+real_good=$testname.good
+
+#
+# remove .good file which should not be stored in git
+#
+rm -f $real_good
```
ronawho: Is this needed? Won't the cleanfile take care of this?

ronawho:
the hostname is "shootout", so that should be good

```sh
eronagha@shootout:~/chapel$ echo $HOSTNAME
shootout
```

test/release/examples/benchmarks/shootout/mandelbrot-fast.prediff
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
ronawho: `#!/usr/bin/env bash` instead?

ronawho:
Made some minor nits inline, but looks good to me overall. 

Let me know if you want me to test anywhere

test/release/examples/benchmarks/shootout/mandelbrot-fast.prediff
bradcray: Hmm, that's probably true and I was being overly cautious.

test/release/examples/benchmarks/shootout/mandelbrot-fast.prediff
ronawho: I think our "preferred" scheme is /usr/bin/env. tbh, it probably doesn't really matter, but I tend to go with /usr/bin/env for everything

PR #5989, opened 2017-04-11
Enable original locale of I/O operations to be queried

mppf:
Adds a new field to `record channel` to store the locale on which the I/O originated. That way, it can be queried for things like Replicated arrays to output in a reasonable manner. This original locale information is passed along to I/O calls within writeThis.

Testing:
- [x] passed full GASNet testing

Thanks to @bradcray for input on the design.


mppf:
@bradcray - here is the implementation we talked about

bradcray:
Cool, I'll give it a shot -- thanks!

bradcray:
This succeeded in getting me unstuck.  Now my remaining failures for the cases I've been wrestling with are my own...

bradcray:
I got my "replicated trouble child" tests passing and am now running full testing.  linux64 passed and so far GASNet looks quite good as well.

mppf:
@bradcray - do you expect to include these commits in your PR? Or should we be attempting to merge this PR first? Also, did the GASNet testing go OK?

bradcray:
I'd prefer for these to go in first/separately, and then I'll re-merge against master.  Yep, GASNet testing went smoothly (though it was GASNet testing of my branch -- not that I'd expect this one to break if that one didn't).

mppf:
Do you have an opinion on the Questions I asked in the PR message?

bradcray:
> is this a user-facing feature? (If so, it needs docs updates)

If writeThis() is user-facing then yeah, I think this should be as well.  Seems plausible that other use cases would also want this information.

> is it acceptable that a writeThis called from within a writeThis will use Locale 0 for stdout?

If the author of the writeThis() wants to plumb the original locale through, can they push the writeln() in their routine into an on-clause?  If so, I think this is acceptable if noted in the documentation.  (Alternatively, I suppose they could call writeThis() directly?)

mppf:
I tried modifying the above example to
``` chapel
  proc writeThis(writer, loc) {
    on loc {
      writeln("in B.writeThis loc=", loc.id);
      writer.write(a);
    }
  }
```

but that hangs - I think the on clause causes the recursive locking scheme to lose track of the fact that it *already has* the lock and instead waits to take the lock (that it already has, but on another locale). (These I/O routines use a recursive locking scheme implemented in the runtime - I think this strategy would work if we switched to a on-statement-aware recursive locking scheme). Calling `writeThis` directly does work, though:

``` chapel
  proc writeThis(writer, loc) {
    writeln("in B.writeThis loc=", loc.id);
    a.writeThis(writer, loc);
  }
```

I'll start a PR to update the documentation before we merge this one, but I'd like to merge this one separately to help you make progress.

bradcray:
No big rush for my sake.  I have a valgrind issue in a new test I added to track down still (and need to focus on slides, so this is background anyway).

mppf:
@bradcray - I've been thinking about an additional / alternative change.

I think it's a shame that 'on' statements don't work within writeThis. The reason they don't work is described in one of my earlier comments. I'm thinking that I could make them work while also solving this problem in a different way.

The idea would be to store the original requesting locale (that this PR adds as a locale argument) and whether or not the channel lock is already taken *by this task* in the channel record. When performing I/O, we'd create a new channel record pointing to the same implementation as before, but where the channel record would indicate locking is not necessary & the original locale of the request.

I think the "surprising behavior" that would result from this change is that a pattern like this would still cause problems:
``` chapel
proc writeThis(writer) {
  cobegin {
     writer.writeln("A");
     writer.writeln("B");
  }
}
```
because we'd have disabled additional locking in the writer passed to the writeThis (so the two writelns would not lock the channel even though they are happening in parallel). But, with the current implementation, such examples would just deadlock. So I'm not sure I have a way to (or intend to) make that one work.

Anyway, if the original requesting locale were part of the channel passed to writeThis
 * we wouldn't need to add the optional `loc` arg to writeThis (but we could)
 * the `loc` would be passed through writeln/writef/whatever in writeThis.

Thoughts?

bradcray:
I like the idea of not having to pass the additional argument through writeThis().  W.r.t. the limitation you bring up, I'd be inclined to disable write*s within parallelism within writeThis() functions since (a) it doesn't seem useful and (b) it's likely to just result in trouble.  But detecting and warning about / squashing such cases seems likely to be fiddly compiler code to implement (so we could file an issue against it and just leave it to the user not to do this for now by documenting the assumption/restriction somewhere).

Mostly, I'm at a point where I'd like to get a merge to the replicated distribution fix in soon in order to stop being the recipient of so much passive-aggression in the Monday meetings.

mppf:
I think the new version is cleaner & I'm happy it doesn't require a compiler mod. It now propagates the original locale to call to `writer.write(...)` in a `writeThis`.

But, the problem I originally hoped to solve (I/O hanging with an on statement) turns out to be due to a slightly different issue than I originally thought. It is due to recursive locking, but the recursive locking wasn't a problem with `writer` but it was a problem with `stdout`. I had something that would now be written
``` chapel
  proc writeThis(writer) {
    var loc = writer.readWriteThisFromLocale();
    on loc {
      writeln("in B.writeThis loc=", loc.id);
      writer.write(a);
    }
  }
```

Then, suppose that this writeThis is called from I/O started on Locale 1 on stdout. The process of preparing to run the writeThis will go to Locale 0 and lock stdout (until the I/O completes, naturally). But then the `on loc` here returns to Locale 1. Then it tries to run `stdout.writeln` which uses an `on` statement to go back to Locale 0 and tries to take the lock there. But the `on` statement back on Locale 0 is in practice executed by the active message handler in a different thread, and so the recursive locking scheme fails and we get a deadlock.

It works if you don't include I/O to stdout in the writeThis, e.g.:
``` chapel
  proc writeThis(writer) {
    var loc = writer.readWriteThisFromLocale();
    on loc {
      writer.writeln("in B.writeThis loc=", loc.id);
      writer.write(a);
    }
  }
```

I'm adding a test to keep what works working, but there's a warning to add in documentation somewhere.

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -1511,7 +1563,7 @@ static void buildStringCastFunction(EnumType* et) {
 
 
 void buildDefaultDestructor(AggregateType* ct) {
-  if (function_exists("deinit", 2, dtMethodToken, ct))
+  if (function_exists("deinit", dtMethodToken, ct))
```
mppf: Moved to PR #6168. I will continue to test these together but can merge that one first.

modules/standard/IO.chpl
```diff
@@ -3072,6 +3080,33 @@ proc channel._set_style(style:iostyle) {
...
+
+// Returns the original locale that the I/O started on
+// Uses readWriteThisFromLocale in order to propagate that
+// information across readThis/writeThis/readWriteThis calls.
+// If readWriteThisFromLocale returns nil, that means the channel
+// was not created to call readThis/writeThis/readWriteThis and
+// so the original locale of the I/O is `here`.
+pragma "no doc"
+inline
+proc channel.getIoStartedOnLocale() {
```
mppf: What do you think about getLocaleOfIoRequest?

Note that this method is for internal use only. The user-facing one is `channel.readWriteThisFromLocale()`.

modules/standard/IO.chpl
```diff
@@ -3623,7 +3658,10 @@ private inline proc _write_binary_internal(_channel_internal:qio_channel_ptr_t,
 
 // Channel must be locked, must be running on this.home
 // x is ref (vs out) because it might contain a literal string.
-private inline proc _read_one_internal(_channel_internal:qio_channel_ptr_t, param kind:iokind, ref x:?t):syserr where _isIoPrimitiveTypeOrNewline(t) {
+private inline proc _read_one_internal(_channel_internal:qio_channel_ptr_t,
+                                       param kind:iokind,
+                                       ref x:?t,
+                                       loc:locale):syserr where _isIoPrimitiveTypeOrNewline(t) {
```
mppf: It is intentional. These functions are overloaded - have two implementations, one for built-in types and another for anything else. I needed the loc:locale argument on the "anything else" version so changed both overloads.

modules/standard/IO.chpl
```diff
@@ -3754,12 +3802,13 @@ private inline proc _write_one_internal(_channel_internal:qio_channel_ptr_t,
...
 proc channel.readIt(ref x) {
   if writing then compilerError("read on write-only channel");
+  const origLocale = this.getIoStartedOnLocale();
   on this.home {
     this.lock();
     var error:syserr;
     error = qio_channel_error(_channel_internal);
     if ! error {
-      error = _read_one_internal(_channel_internal, kind, x);
+      error = _read_one_internal(_channel_internal, kind, x, origLocale);
```
mppf: The trivial answer is that in that function, `origLocale` is set by `const origLocale = this.getIoStartedOnLocale();`. But the answer you were probably looking for is that _readWriteThisFromLocale is set in _read_one_internal and _write_one_internal overloads that handle non-primitive types in the construction of a new, unlocked channel. Search for _readWriteThisFromLocale to find it.

modules/standard/IO.chpl
mppf: I've also thought about naming it readWriteThisFromLocaleOrHere()

lydia-duncan: Of the four options, I think getLocaleOfIoRequest is best

modules/standard/IO.chpl
lydia-duncan: Oh, in the constructor.  That's why I missed it.  Carry on!

PR #5993, opened 2017-04-11
Broaden special-casing of mandelbrot.chpl to handle mandelbrot*.chpl

bradcray:
[source of problem found by @ronawho]

This is a silly test which seems to be designed to dodge around the
fact that mandelbrot's output is binary and therefore cannot easily be
grepped.  Previously, we checked for mandelbrot.chpl.  This change
makes us check for mandelbrot*.chpl.  A better test might be to see if
outfile is binary, but unfortunately I can't afford to take more test
failures this week so am leaving a comment instead.

### DIFF
--- a/util/test/prediff-for-aprun-with-moab
+++ b/util/test/prediff-for-aprun-with-moab
@@ -11,7 +11,6 @@
 # variants...
 #
 if [[ "${1}" != mandelbrot* ]] ; then
-    echo "match!"
     outfile=$2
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho: For your special consideration.

ronawho:
👍 

I'll see if we can remove this prediff altogether at some point in the next couple of days

ronawho:
Also, confirmed that this fixes testing on kay

PR #6001, opened 2017-04-13
Continued small updates to visibleCandidates

noakesmichael:
This simple PR begins to update the logic for expandVarArgs().

A. Rename the external uses of this function to be expandIfVarArgs() so that it is more
obvious, at the call sites, that this is a conditional update to the function that is being
processed.

B. expandIfVarArgs() maintains a cache that maps user-defined functions to the expanded
version of the function.  Before this PR *every* function in the code base was inserted in
to the cache.  After this PR only functions with varargs are inserted.  It is unlikely that this
will result in an observable performance difference, especially for the first implementation
of this change, but it makes the intent clearer and supports possible future enhancements.

Passed a single-locale paratest

### DIFF
--- a/compiler/resolution/visibleCandidates.cpp
+++ b/compiler/resolution/visibleCandidates.cpp
@@ -186,7 +186,7 @@ static void filterCandidate(CallInfo&                  info,
 static void filterConcrete(CallInfo&                  info,
                            ResolutionCandidate*       currCandidate,
                            Vec<ResolutionCandidate*>& candidates) {
-  currCandidate->fn = expandIfVarArgs(currCandidate->fn, info.actuals.n);
+  currCandidate->fn = expandIfVarArgs(currCandidate->fn, info);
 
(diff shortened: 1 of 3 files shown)


compiler/resolution/visibleCandidates.cpp
```diff
@@ -28,33 +28,35 @@
...
 #include "symbol.h"
 
-static void doGatherCandidates(CallInfo&                  info,
-                               Vec<FnSymbol*>&            visibleFns,
-                               bool                       compilerGenerated,
-                               Vec<ResolutionCandidate*>& candidates);
+static void       doGatherCandidates(CallInfo&                  info,
+                                     Vec<FnSymbol*>&            visibleFns,
+                                     bool                       generated,
+                                     Vec<ResolutionCandidate*>& candidates);
```
lydia-duncan: One day you must tell me how you do this without inserting all the spaces by hand

PR #6002, opened 2017-04-13
Fix date parsing in perf graphs

benharsh:
Use datejs third-party library to parse dates instead of native Date.parse


benharsh:
Care to review, @ronawho ?

ronawho:
Does this end up overriding the default Date.parse?

benharsh:
Yes, so no perfgraph.js changes needed.

ronawho:
Ok, fine by me. Could you send me links to chapcs/16-node so I can verify this fixes the issues I was seeing on the graphs?

benharsh:
Offline Elliot suggested that we revert the 2.0 PR until things are more stable, and I agree.

PR #6014, opened 2017-04-13
Continue to update expandVarArgs

noakesmichael:
Prior to this PR the body of expandVarArgs() was quite convoluted.

A. There were a couple of points at which there was a comment in the code saying
something to the effect "assume there is one var args set"

B. It was hard to determine whether the function would a) return NULL,
b) return a modified version of the original function c) return a new function.

A little study suggested that the primary source of complexity
wast that it was trying to handle

proc foo(x ... 4) 

and 

proc foo(x ... ?N)

in a single loop even though the current business logic
is quite different for these two cases.

This PR separates the original implementation into three pieces

One function that confirms there is just one var-arg set and which form it is,
and then a pair of functions that are specialized for each case.

This PR also adds an explicit check that generates a user diagnostic if a
function has more than one var-arg set and includes a test to lock in that
this is reported.

This PR consists of multiple commits that help to clarify the transformations that
are applied to do this.

The only change in observable behavior would be if there were a function with
multiple var-args.

Also performed a run with clang and std=C++14.


compiler/resolution/visibleCandidates.cpp
```diff
@@ -404,19 +404,136 @@ static bool hasVariableArgs(FnSymbol* fn) {
...
+
+    if (isQueryVariable == false) {
+      retval = expandVarArgsFixed(fn, info);
+
+    } else {
+      retval = expandVarArgsQuery(fn, info);
+    }
+
+  } else {
+    INT_FATAL(fn, "No support for function with multiple vararg sets");
```
lydia-duncan: Shouldn't this be a USR_FATAL?

noakesmichael:
I think you’re right.

Thanks,

Mike



On Apr 13, 2017, at 2:26 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


In compiler/resolution/visibleCandidates.cpp<https://github.com/chapel-lang/chapel/pull/6014#discussion_r111491011>:

> +  }
+
+  if (numVarArgs == 0) {
+    INT_ASSERT(false);
+
+  } else if (numVarArgs == 1) {
+
+    if (isQueryVariable == false) {
+      retval = expandVarArgsFixed(fn, info);
+
+    } else {
+      retval = expandVarArgsQuery(fn, info);
+    }
+
+  } else {
+    INT_FATAL(fn, "No support for function with multiple vararg sets");


Shouldn't this be a USR_FATAL?

lydia-duncan:
Is there a test to lock this in?

noakesmichael:
Good point.  There has never been a test to lock this in.  I’ll add one as a separate PR.

Mike



On Apr 13, 2017, at 2:33 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


Is there a test to lock this in?

test/functions/varargs/just-one-set.chpl
```diff
@@ -0,0 +1,33 @@
...
+  writeln('t1: ', t1);
+  writeln();
+}
+
+proc test2(t2 : real ... ?N2) {
+  writeln('t2: ', t2);
+  writeln();
+}
+
+proc test2(t1 : int ... ?N1, t2 : real ... ?N2) {
```
lydia-duncan: Shouldn't this be `test3`?

lydia-duncan:
Okay, I think this is right.  There may be something I have missed, though.  Do you have a sense of what other tests there are of varargs?  Should this add more tests of explicit behavior to ensure things haven't changed or is the current testing sufficient?

noakesmichael:
Sigh. Thanks.

Mike


On Apr 13, 2017, at 3:08 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


In test/functions/varargs/just-one-set.chpl<https://github.com/chapel-lang/chapel/pull/6014#discussion_r111497846>:

> +  test2(10.0, 20.0, 30.0);
+
+  test3(1, 2, 3, 10.0, 20.0, 30.0);
+}
+
+proc test1(t1 : int ... ?N1) {
+  writeln('t1: ', t1);
+  writeln();
+}
+
+proc test2(t2 : real ... ?N2) {
+  writeln('t2: ', t2);
+  writeln();
+}
+
+proc test2(t1 : int ... ?N1, t2 : real ... ?N2) {


Shouldn't this be test3?

lydia-duncan:
The dangers of adding futures :)

Issue #5935, opened 2017-04-04
Problems with conditional expressions in initializers


bradcray:
### Summary of Problem

I expected that conditional expressions would be supported in phase I of initializers since they don't introduce control flow that would confuse whether or not the field is being initialized.  However, instead they result in an error message and one that isn't clearly related to the conditional expression:

````
error: initializers cannot return a value
````

My wild guess is that we implement the conditional expression as a nested function and that the compiler is incorrectly assuming that the return within that nested function implies that there's a return in phase I.  If so, this may be a five-minute fix.

### Steps to Reproduce

**Source Code:**

```chapel
class C {
  const x: real;

  proc init(flag: bool) {
    x = if flag then 1.2 else 3.4;
    super.init();
  }
}
```

**Compile command:**
``chpl foo.chpl -o foo``

**Associated Future Test(s):**
``test/classes/initializers/initViaConditional.chpl``

### Configuration Information

- Output of `chpl --version`: 1.15.0.6a45818

bradcray:
Here's another code that runs into similar issues:

````chapel
  proc init(colors: [?colorSpace]) {                                            
    chameneos = for i in colorSpace do new Chameneos(i, colors[i]);             
  }                                                                             
````

(probably for similar reasons)

bradcray:
@lydia-duncan : This seems like a potential candidate for a five-minute fix...  Have you had a chance to look into it at all?  I seem to keep running into it when I write initializers.

lydia-duncan:
My gut says that this won't be too hard to fix, either, but I haven't taken a deep look at what it would take to solve it yet.  Sorry you keep running into it, I'll bump it up my queue if Mike doesn't get to it first!

PR #6018, opened 2017-04-13
Spit-and-polish for Elliot's chameneos

bradcray:
Here, I made proposed changes for a next version of chameneos-redux to my study version, in approximate order of importance:

I wouldn't expect this to result in any behavioral or performance changes relative to Elliot's version -- it's mostly just code cleanup and organization.

* added Elliot's name to the byline due to his work in tuning it
* made spin limit into a config
* combined two print routines into one now that they're called consecutively
* removed 'Population' record which I think is arguably less useful now that main() has to do more juggling of the populations manually.  Moved contents either inline or into standalone procs
* hoisted colors10 to the top as a result and/or because it seems more "global"
* changed from underscores to camelCase
* removed some curly brackets from a simple if-then-else

I also updated Elliot's version (with his permission) to include all of these changes other than the removal of Population so that we could weigh the stylistic impacts on main() vs. having the record.  As part of this, we ended up simplifying the interface to Population to take a color array, removed the population size configs (because they're not really all that flexible anyway), and introducing a 'colors' field to the constructor in order to be able to set up the population properly.  (We tried adding an initializer to avoid having to store the colors field, but this ran afoul of issue #5935)

### DIFF
--- a/test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc.chpl
+++ b/test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc.chpl
@@ -1,26 +1,37 @@
 /* The Computer Language Benchmarks Game
    http://benchmarksgame.alioth.debian.org/
 
-   contributed by Hannah Hemmaplardh, Lydia Duncan, and Brad Chamberlain
+   contributed by Hannah Hemmaplardh, Lydia Duncan, Brad Chamberlain,
+     and Elliot Ronaghan
    derived in part from the GNU C version by Dmitry Vyukov
 */
 
 config const n = 600,              // number of meetings (must be >= 0)
              popSize1 = 3,         // size of population 1 (must be > 1)
-             popSize2 = 10;        // size of population 2 (must be > 1)
+             popSize2 = 10,        // size of population 2 (must be > 1)
+             spinLimit = 15;       // number of times to spin before yielding
 
 enum Color {blue=0, red, yellow};  // the chameneos colors
 use Color;                         // permit unqualified references to them
 
+// special colors to use for a chameneos population of size 10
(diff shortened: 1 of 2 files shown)


bradcray:
@ronawho: See what you think about this version from a stylistic perspective.  You'll probably want to compare against your version rather than the previous version of mine (which is what GitHub will show).

@lydia-duncan will probably want to take a look as well due to her history with this benchmark.

test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc.chpl
```diff
@@ -36,82 +53,35 @@ proc printColorEquations() {
...
-  // place, and then creating per-chameneos tasks to have meetings.
-  //
-  proc holdMeetings(numMeetings) {
-    const place = new MeetingPlace(numMeetings);
-
-    coforall c in chameneos do           // create a task per chameneos
-      c.haveMeetings(place, chameneos);
-
-    delete place;
+proc print(chameneos) {
```
ronawho: arg name `chameneos` confused me. What do you think about `chameneosArr` or `chameneosGroup` or something?

ronawho:
👍 

In person I mentioned some nits and a slight preference for keeping the `Population` wrapper, but I don't feel strongly about any of that. i.e. I'd have no real objections to submitting this version

test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc.chpl
bradcray: I updated this arg name to 'population' in 223fb2a

lydia-duncan:
:+1:

PR #6023, opened 2017-04-14
Fix regex-redux .dat filename for release

bradcray:
Once I got the submitted regex-redux working, I copied its files
over to the release, failing to update the .dat filename specified
in the .perfkeys file.

(I don't think I realized .dat filenames could be specified in both
.perfexecopts and in .perfkeys, and didn't think to look for it here
in any case since each version had a unique name...)


bradcray:
@psahabu : Want to take a quick look?

psahabu:
Looks good to me, sorry I didn't catch that.

Issue #5271, opened 2017-01-27
Upgrade re2 versions, try to contribute patches upstream


ronawho:
Our "version" of re2 is about 3 years old. I don't track re2 development myself, so I'm not sure what all has gone in since then, but it seems likely that it's worth upgrading. If/when we do upgrade, we should also switch over to the new third-party directory structure like we've done for qthreads(#4498), hwloc(#3453), jemalloc(#3424), and gasnet(#3914).

It also think we should try to use pkg-config to determine the link args instead of always throwing `-lpthread`

Like our other third-party libraries (qthreads/hwloc/jemalloc/gasnet) it might be worth finding an owner for upgrading versions.

We also have patches we apply to re2 to "allow regular expressions in the format
string for Chapel's readf() for searching on QIO channels." It'd be nice if we could contribute that upstream and use a clean/non-patched version of re2, especially if we start upgrading regularly. 

 - [X] switch re2 dir structure over to re2-src
 - [X] find an owner to update re2 (@mppf)
 - [X] document how to upgrade version (how to upgrade and apply our patches)
 - [X] try to contribute our re2 patches upstream? (tried unsuccessful, see https://github.com/chapel-lang/chapel/issues/5271#issuecomment-320268636)
 - [X] use pkg-config to determine the link args instead of always throwing `-lpthread`? (punting based on https://github.com/chapel-lang/chapel/issues/5271#issuecomment-321720595)

ronawho:
@mppf I assigned to you since you added the original support for re2, but I don't think you necessarily have to own this. I wouldn't mind adding this to the list of 3rd party libs I upgrade, although I don't know anything about the patches we apply, so I wouldn't feel comfortable trying to contribute those upstream.

ronawho:
#6024 upgraded to the latest re2 version, switched dir structure over to re2-src, and added upgrading documentation 

Thanks @mppf!

ronawho:
@mppf completely naive (probably stupid) question: do we _need_ to patch re2? Could we accomplish the same thing at a higher level? I'm assuming we could, but that it would come at a significant performance penalty. Is that accurate, or does it have to be done in re2?

Anyways, not super important. It'd be nice if we could contribute things upstream. Not a big deal if we can't. It'd be a nice fallback to implementing what the patches achieve at a higher level, but also not a big deal if we can't. re2 isn't making very many dramatic changes anymore, so I don't think it'll be hard to maintain our patches going forward.

mppf:
[...]

PR #6024, opened 2017-04-14
Update RE2

mppf:
* Upgrade RE2 to re2 commit 2b16c27
* Fix a trivial C++ compilation warning

I verified that the current diff with RE2 makes sense.

- [x] test/regexp with clang on Mac OS X
- [x] full local testing
- [x] update commit ID in README, make it look more like other READMEs in third-party
- [x] perf check regex-dna / regex-dna-redux: perf improves by 10% - 25%

```
16 trials
new regex-dna best: 3.19 worst: 3.26
new regex-dna-redux best: 3.80 worst: 3.84
old regex-dna best: 3.50 worst: 3.55
old regex-dna-redux best: 4.82 worst: 4.89
```


mppf:
@ronawho - could you sign off on this or let me know what other steps you think I need to take?

ronawho:
https://github.com/chapel-lang/chapel/issues/5271 probably has some TODOs in it. 

I didn't look at this too carefully yet, but I see some cmake stuff. Has re2 switched from autoconf/configure to cmake? If so, does this mean cmake is now a dependency for us?

mppf:
AFAIK RE2 is currently supporting both CMake and Make, so it's not a dependence. It seems that the code now assumes C++11 (but we're used to needing that for LLVM builds).

ronawho:
llvm is always built with the host compiler though, right ? I'd bet intels C++11 support is good, but what about cray/pgi?

ronawho:
Also, do you have any sense of what improvements this brings us? 

It looks like re2 doesn't have a ChangeLog or releases notes, at least not that I found quickly :(

mppf:
I can't really describe what we're getting with the updates easily either. AFAIK the commit log is all we have. It's multiple years of updates though.

ronawho:
I think it's worth running a regexp heavy benchmark just to check that this won't introduce any major regressions. 

When upgrading jemalloc I test binary-trees. For re2, I'd recommend testing regex-dna/regex-dna redux.

mppf:
@ronawho - see the PR message for the perf test result. Is there anything else you think I need to do for this PR in particular?

third-party/re2/Makefile
```diff
@@ -49,7 +49,7 @@ $(RE2_BUILD_SUBDIR):
 	mkdir -p $@
 
 $(RE2_H_FILE): $(RE2_BUILD_SUBDIR)
-	cd re2 && \
+	cd re2-src && \
```
ronawho: `cd $(RE2_SUBDIR) && \` ?

ronawho:
re2 link in the README should be updated to the github repo. I also think it's worth including the "version" number in the README and maybe making it more uniform with the other third-party READMES (e.g. gasnet, jemalloc). Maybe something like:


```rst
=====================
re2 README for Chapel
=====================

This copy of re2 2017-04-01 (2b16c27) is being released with Chapel
for convenience and was obtained from:

  https://github.com/google/re2

Any Chapel issues that seem to be related to re2 should be directed
to the Chapel team at http://chapel.cray.com/bugs.html.


Chapel modifications to re2
===========================
We modify the RE2 library to allow regular expressions in the format
string for Chapel's readf() for searching on QIO channels.

- moved NFA and OnePass searches to be implemented with templates
  to allow string types other than const char*
  Also modified these to use begin_reading() instead of begin()
  for the pointer that will be dereferenced.
- added file_strings.h to represent a portion of a file accessible
  by calling some read routine (through a pointer)
  and modified string_piece to include it.
- implemented RE2::MatchFile which searches a buffer
  (optionally) and then some other string type.
- RE2 constructor now computes min/max possible match length
  for use in MatchFile.

```

PR #6025, opened 2017-04-14
Add a function asciiToString to convert a uint(8) ascii value into a string

daviditen:
The asciiToString function takes a uint(8) argument and returns a
zero-terminated string containing the character corresponding to that ascii
code.  Also added a simple test that converts a string to uint(8)s and back.


modules/internal/String.chpl
```diff
@@ -1743,6 +1743,16 @@ module String {
...
   }
 
+  /*
+     :returns: A string with the single character with the ASCII value `i`.
+  */
+  inline proc asciiToString(i: uint(8)) {
+    var buffer: [0..1] uint(8);
+    buffer[0] = i;
+    buffer[1] = 0;
+    return new string(c_ptrTo(buffer), 1, 2);
```
mppf: I would probably use chpl_here_alloc / free rather than using the array -> c_ptrTo mechanism, since things like multi-ddata make such uses of arrays problematic. I'd like to limit any problems in such areas to tests of extern functions.

mppf:
Looks good, other than my one note. Can you say in the PR message what caused you to implement this now?

mppf:
OK, consider it reviewed, but double-check testing & check for memory leaks on your new test to be sure.

daviditen:
Thanks @mppf.

I let the string take ownership of the original buffer as you suggested instead of copying and deleting it.  I've run my new test with --memLeaks, and am currently running a full paratest.

I've run into a desire for this functionality a few times in the past, and have heard questions asking how to do it from a few other people too.  This time, I had a toy program that reads a sequence of commands to manipulate an array of bytes, and one of the commands is "output the pointed-to byte as an ascii character".

PR #6030, opened 2017-04-14
Fine-tune Elliot's chameneos

bradcray:
Something about last night's changes (the storage of 'color' as a field
of 'Population'?) made Elliot's chameneos perform much more poorly than
it had been.  Here, I remove the color field, replacing it with a
constructor (I tried to use an initializer but ran afoul of issue #5935).
The constructor-based approach seemed to require the declaration of a
domain field as well -- i.e., I couldn't find a way to simply make
'chameneos' generic and have the constructor resolve.  I also changed
the deinitializer to a destructor for symmtry.

This version seems to be going faster than it has before, and faster
than my variation as well.  Unfortunately, it relies on
constructors/destructors which will be going away soon, so will need
to be updated in the future.

### DIFF
--- a/test/studies/shootout/chameneos-redux/elliot/chameneosredux-ejr.chpl
+++ b/test/studies/shootout/chameneos-redux/elliot/chameneosredux-ejr.chpl
@@ -47,12 +47,14 @@ proc printColorEquations() {
 // a chameneos population
 //
 record Population {
-  var colors;  // the colors of the chameneos in the population
+  var chamSpace: domain(1);
+  var chameneos: [chamSpace] Chameneos;
 
-  //
-  // an array of chameneos objects representing the population
-  //
-  var chameneos = [i in colors.domain] new Chameneos(i, colors[i]);
+  proc Population(colors) {
+    chamSpace = colors.domain;
+    forall i in chamSpace do
+      chameneos[i] = new Chameneos(i, colors[i]);
+  }
 
   //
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho: Check it out, yo.

ronawho:
I'm not opposed to this, but OOC what would the "ideal" init/deinit version look like?

bradcray:
That's a good question and pushes into the limits of my knowledge about the plan for generics.  I'm hoping it could be something like this:

````chapel
record Population {
  var chameneos;

  proc init(colors) {
    chameneos = [i in colors.domain] new Chameneos(i, colors[i]);
  }
...
````

So one approach would be to submit the constructor version today (loosely speaking) and update to initializers when they're ready and constructors are deprecated.  Another would be to submit my version today, and file this version once initializers work.

ronawho:
Ok, cool. That's what I was hoping the initializer might look like. 

I think commit this PR, and we can decide which version we want to submit based on perf results. IMO it seems really unfortunate to "promote" the use of constructor/destructor knowing that we want to retire it, but I don't feel strongly about that

bradcray:
I agree.

FWIW, I now see a best time of 1.18 sec for ejr on the shootout box compared to my best blc timing of 1.24.  I think if I had to decide today based on that, I'd submit the blc version in order to reduce the number of versions we have to update in the future (and because the domain field + constructor approach seems a little awkward).

PR #6032, opened 2017-04-14
File an idealized version of chameneos as a future

bradcray:
This is a variation on chameneos-ejr which attempts to use an
initializer rather than a constructor.  It currently runs afoul of
issue #5935, though I expect that once we get past that, there will be
other challenges related to generics ahead (e.g., am I now writing
cases that I told Mike I didn't expect us to be able to handle a few
weeks ago? :)  And what options would we have for writing this idiom
if not this way?)


bradcray:
@ronawho, FYI: This is a variation on your version that uses init/deinit.

ronawho:
Seems like it'd be better to make a smaller reproducer for this to make debugging easier for the initializers folks.  I also wouldn't be surprised if this version became pretty out of sync due to other chameneous improvements we might make and so it would no longer represent an "ideal" version.)

bradcray:
My goal with this version isn't to file a bug against initializers (I've already got one of those), so much as to track our ability to evolve toward this version of chameneos.  I'm also expecting that chameneos won't change fast enough that maintenance will be difficult...

ronawho:
ok, fair enough. good to merge

bradcray:
(I'm also too lazy at this point to try and boil it down to the core pattern...  Maybe after we get past the first issue and hit a second one I will, though).

PR #6033, opened 2017-04-14
Futures for 'Owned' types

bradcray:
I tried to write a version of Chameneos that used 'Owned' types
and ran into a few failures related to l-values vs. const-ness in
one case and coercions/interoperability between Owned(t) and
t.  See the .future files for details.


bradcray:
Hi @mppf: Will you take a look at these before I merge them to see if I should make any changes to them?  (no rush)  I'd suggest reviewing them in the order:

* owned-init2
* owned-init
* owned-comp

test/classes/delete-free/owned-init2.chpl
```diff
@@ -0,0 +1,20 @@
...
+      writeln("same!");
+    else
+      writeln("different!");
+  }
+}
+
+proc main() {
+  var a: [1..3] Owned(C);
+  for i in 1..3 do
+    a[i] = new Owned(new C(i));
```
mppf: Ah, I see the problem here.
Since = for owned clears out the RHS (since it is an ownership transfer), = takes the RHS by ref.
But other parts of the compiler raise an lvalue error if you pass an expression, such as `new Owned(new C(i))` as ref.

The workaround is this:
``` chapel
  var tmp = new Owned(new C(i));
  a[i] = tmp;
```
which works for me.

We could 
 1) relax the l-value checking in this case (and other cases?) for Owned (maybe with a pragma?)
 2) relax l-value checking more generally (in favor of "is it const?")
 3) require the use of the workaround above

I'm uncertain of the real value that l-value rules offer (as a separate concern from const-ness). I think any consideration of (1) or (2) needs to start by trying to understand the merits of l-value rules. Note that currently we don't really follow l-value rules for arrays (and I think that's for historical reasons rather than really be design at this point).

test/classes/delete-free/owned-init.chpl
```diff
@@ -0,0 +1,18 @@
...
+  proc matches(other) {
+    if this == other then
+      writeln("same!");
+    else
+      writeln("different!");
+  }
+}
+
+proc main() {
+  const a = [i in 1..3] new Owned(new C(i));
```
mppf: Pretty sure this is the same problem as init2; just within the compiler generated promoted routine.

test/classes/delete-free/owned-comp.chpl
```diff
@@ -0,0 +1,20 @@
...
+      writeln("different!");
+  }
+}
+
+proc main() {
+  var a = new Owned(new C(1));
+  var b = new Owned(new C(2));
+
+  a.matches(b);
+  a.matches(a);
```
mppf: This one is interesting and a different problem from the others. It's showing the limits of our current forward-but-don't-coerce strategy for Owned.

Not sure how to "fix" this without user-defined-coercions for Owned, but
the obvious solution is to write

``` chapel
a.matches(b.borrow());
a.matches(a.borrow());
```

``` chapel
```

mppf:
@bradcray - I commented on these; it might make the most sense to include my comments in the .future files; but of course you might decide to make some different tests too.

PR #6035, opened 2017-04-14
Add more tests for compiler error messages

benharsh:
There are three minor changes to the compiler to either enable an error message or switch to a USR_FATAL_CONT.

The rest of the changes are new tests against existing error messages. I did not find any of these error messages in existing .good files.

Testing:
- [x] local paratest


mppf:
Looks good to me. I wonder if you searched for USR_FATAL in the compiler and checked that each error message appears in a .good file? If you did, did you have a script doing that? It seems to be a smoke-test-y thing to do.

benharsh:
Yeah, just USR_FATAL and .good files. I didn't have a script, but I'm sure we could whip one together.

We should also do something similar for USR_WARN and module-level errors/warnings.

lydia-duncan:
You are a saint

PR #6036, opened 2017-04-15
Add make install, adjust compiler to support it

mppf:
This is intended to support Debian packaging.

The normal install locations for certain files will differ (between e.g.  /usr/share/chapel and /usr/lib/chapel).  For this reason, introduce new compiler & environment variables that are computed by the compiler as described below:

```
                   CHPL_HOME set         installed
CHPL_RUNTIME_LIB   lib/                  /usr/lib/chapel/1.15/runtime/lib
CHPL_RUNTIME_INCL  runtime/include       /usr/lib/chapel/1.15/runtime/include
CHPL_THIRD_PARTY   CHPL_HOME/third-party /usr/lib/chapel/1.15/third-party
CHPL_HOME          CHPL_HOME             /usr/share/chapel/1.15/
```

Adds a configure script and 'make install' uses that configuration.

The configure script help shows the idea:
```
Usage: ./configure [--prefix=PREFIX] [--chpl-home=DEST_DIR]

       --prefix=PREFIX selects a split installation
                e.g. PREFIX could be /usr/ to install
                     to /usr/bin/chpl, /usr/share/chapel/
                     e.g.  /usr/share/chapel/<vers>/modules

       --chpl-home=DEST_DIR selects a bundled installation
                e.g. DEST_DIR could be /opt/chapel/ to install
                     to /opt/chapel with subdirectories
                     matching the source code / release
                     e.g.  /opt/chapel/modules
```

Running configure shows some output to show the current configuration and creates a `chplconfig` file in the current directory:

```
$ ./configure --prefix=/tmp/prefix

  Currently selected Chapel configuration:

CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none *
CHPL_TASKS: fifo *
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: cstdlib *
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: none *
CHPL_HWLOC: system *
CHPL_REGEXP: none *
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none

  Selected installation options:

PREFIX: /tmp/prefix

  To change the selected Chapel configuration, set environment variables.
  See doc/rst/usingchapel/chplenv.rst
  To change installation options, see ./configure --help

  If the configuration above is appropriate, build Chapel with:

make
make check

  and possibly install it with

make install
```

The configure script creates these files:
 * chplconfig (to keep track of the selected configuration)
 * configured-chpl-home or configured-prefix (to save where it will be installed)

For the version installing to a prefix (e.g. to /usr/bin/chpl), the compiler needs to know where to find the other resources. Therefore, in that case, the compiler builds in the contents of configured-prefix (or "" if it is not set) which will form the default CHPL_HOME when the program is run. This is accomplished with Makefiles generating compiler/main/CONFIGURED_PREFIX and by modifying setupChplHome in driver.cpp to use the configured prefix.

Then, this PR adds `make install` by adding a script `util/buildRelease/install.sh` that is run for `make install`. That script uses `configured-chpl-home` or `configured-prefix` in order to know where to install files. `make install` supports  `make DESTDIR=/tmp/stage install` which is important for Debian. We do that by passing DESTDIR to to install.sh with `-stage=`. install.sh takes some other cleanup steps - it remove .pyc files and .timestamp that would be installed (these would cause problems on Debian).

Modifies some Makefiles to use the new variables. Fixes some 3rd-party package support to use CHPL_THIRD_PARTY instead of assuming third-party is a subdirectory of CHPL_HOME.

Modified printchplenv to omit CHPL_HOME with --overrides --anonymize

Future Work:
* make distclean remove generated files (lexer and parser)
* make installdocs, installexamples
* make uninstall

Passed full local testing.

Reviewed by @ben-albrecht and @daviditen - thanks!


mppf:
@ben-albrecht - do you think configure should save the settings to $CHPL_HOME/chplconfig ? If it did, it would be "sticky". Vs right now, the configured-chplconfig is only used during 'make install'. Anyway, if we do that, we should also check if we try to build e.g. RE2 even if chplconfig sets it to none. No environment variables should need to be set after the configure step.

ben-albrecht:
Yes, but should users expect environment variables to override the chplconfig for `make install` like they would elsewhere?

mppf:
@ben-albrecht - I think the enviro vars would override chplconfig, but it wouldn't affect the chplconfig that we install - just build an additional configuration. I think it'd be reasonable to error out of `make install` if there is a chplconfig and it doesn't reflect the current configuration.

mppf:
@ben-albrecht I think this is ready for your actual detailed review

compiler/main/driver.cpp
```diff
@@ -270,7 +281,7 @@ static void setupChplHome(const char* argv0) {
   }
 
   if( chpl_home ) {
-    if( strlen(chpl_home) > FILENAME_MAX )
+    if( strlen(chpl_home) >= FILENAME_MAX )
```
ben-albrecht: Was this previously a (minor) bug?

mppf: no, it was right before, since CHPL_HOME has space for FILENAME_MAX + 1. I'll un-do this change.

util/chplenv/chpl_home_utils.py
```diff
@@ -17,6 +17,11 @@ def get_chpl_home():
         chpl_home = dirname(dirname(dirname(os.path.realpath(__file__))))
     return chpl_home
 
+@memoize
+def get_chpl_third_party():
```
ben-albrecht: Follow-up Q: Do we need one of these for `get_chpl_runtime()`, `get_chpl_compiler()`, etc.?

daviditen:
@ben-albrecht asked me to take a high-level look over this.  I'm not super familiar with installation standard practices, so I'm assuming you've done the right thing as far as where this puts things.

I tried:

```
./configure --prefix=$HOME/install
make
make check
make install
```

And I get an error when I try to compile something with `$HOME/install/bin/chpl`

```
$HOME/install//share/chapel/1.15/runtime/etc/Makefile.include:134: Expected runtime configuration: fs-none
$HOME/install//share/chapel/1.15/runtime/etc/Makefile.include:134: *** The runtime has not been built for this configuration. Check $CHPL_HOME/util/printchplenv and try rebuilding with 'make' from $CHPL_HOME.  Stop.
error: compiling generated source
```

Do I need to do something to set up my environment to use the installed chpl?  It isn't obvious to me that I should have to based on the documentation changed here.

I think there should also be a way to automatically uninstall so the user doesn't need to try to figure out what files are part of the Chapel install across multiple directories if they decide to get rid of Chapel.

I think at a high level this looks pretty good, notwithstanding my issues.

util/chplenv/chpl_home_utils.py
mppf: We might one day but at present I don't think any of these Python chplenv scripts would use it and the only reason it's here is to enable that.

mppf:
@daviditen - I've fixed the problem you were seeing; Ben and I had been testing with quickstart. It had to do with problems auto-detecting gmp/re2 was built.

You brought up `make uninstall`. I'd like to leave this as future work.

Additionally, I am reminded that `make check` currently requires CHPL_HOME to be set. Would it be OK to auto-detect it, so the ./configure etc process can work without setting CHPL_HOME? Otherwise our instructions / configure output should not include `make check` I think.

The `configure` process is meant to work even if you haven't done a `setchplenv.bash`.

daviditen:
I tried the `make install` process I did before again and this time it worked.

I'm OK with leaving `make uninstall` as future work, but I do think it's important.

mppf:
the main Future Work is in issues:
 * #6353 for make uninstall
 * #6600 for make distclean

PR #6037, opened 2017-04-16
Promote latest bradc chameneos version to release as 'fast' version

bradcray:
The current 'elliot' and 'bradc' variations of chameneos redux are
performing similarly, but my version doesn't rely on the use of a
slightly crufty record constructor (and initializers can't handle this
case yet).  For this reason, Elliot and I were thinking on Friday
that this is probably the version to put forward for the time being
and to consider switching to the Elliot version once it is supported
by initializers.


bradcray:
@ronawho: this still seems right to me after thinking on it for a few days.  You?

PR #6039, opened 2017-04-16
Fix cross-references and readThis/writeThis docs

mppf:
I noticed that IO.chpl still referred to the language specification for readThis/writeThis, even though this material moved to the "IO Support" document (generated from ChapelIO.chpl).

This PR adds cross references to fix that.

While fixing the references, I noticed some out-of-date and incorrect text about default read/write methods (should be readThis / writeThis methods). Fixed those too.

### DIFF
--- a/modules/internal/ChapelIO.chpl
+++ b/modules/internal/ChapelIO.chpl
@@ -163,15 +163,15 @@ Default ``readThis`` methods are created for all types for which a user-defined
 ``readThis`` method is not provided.  The default ``readThis`` methods are
 defined to read in the output of the default ``writeThis`` method.
 
-Additionally, the Chapel implementation includes ``writeThis`` or methods for
+Additionally, the Chapel implementation includes ``writeThis`` methods for
 built-in types as follows:
 
 * for an array: outputs the elements of the array in row-major order
   where rows are separated by line-feeds and blank lines are used to separate
   other dimensions.
-* for a `domain`: outputs the dimensions of the domain enclosed by
+* for a domain: outputs the dimensions of the domain enclosed by
   ``{`` and ``}``.
-* for a `range`: output the lower bound of the range, output ``..``,
+* for a range: output the lower bound of the range, output ``..``,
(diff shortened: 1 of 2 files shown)


modules/internal/ChapelIO.chpl
```diff
@@ -141,34 +143,45 @@ function resolution error if the class NoRead is read.
...
   comma.  The output is delimited by ``(`` and ``)``.
 
-Default ``read`` methods are created for all types for which a user-defined
-``read`` method is not provided.  The default ``read`` methods are defined to
-read in the output of the default ``write`` method.
+Default ``readThis`` methods are created for all types for which a user-defined
+``readThis`` method is not provided.  The default ``readThis`` methods are
+defined to read in the output of the default ``writeThis`` method.
+
+Additionally, the Chapel implementation includes ``writeThis`` or methods for
```
lydia-duncan: I think there's a missing word here, I doubt you meant to say "includes ``writeThis`` or methods".

modules/internal/ChapelIO.chpl
```diff
@@ -141,34 +143,45 @@ function resolution error if the class NoRead is read.
...
+built-in types as follows:
+
+* for an array: outputs the elements of the array in row-major order
```
lydia-duncan: Did you mean to italicize "domain" and "range" but not "array" and "tuples"?

modules/standard/IO.chpl
```diff
@@ -159,9 +159,8 @@ with :proc:`channel.readf` and :proc:`channel.writef`.  It is possible to write
 data to strings (see "The write and writeln Methods on Strings" in the Chapel
```
lydia-duncan: I don't think this section exists anymore either

modules/internal/ChapelIO.chpl
mppf: I opted to un-italicise domain/range since it seemed to be distracting formatting to me & isn't used later in those bullets.

PR #6042, opened 2017-04-17
c2chapel 0.1.0

benharsh:
Major rewrite of c2chapel - warranting a v0.1.0 designation

The original c2chapel.py script was somewhat limited in its capability.
This PR is a major rewrite of that script to add and solidify the following
functionality:
- Handle new C99 constructs:
  - function pointers
  - structs (with their fields)
  - structs in typedefs
  - typedefs
  - varargs
- standard headers no longer cause as many fatal errors
- Scans for #define'd integer literals and emits extern consts
- Uses c_string and c_void_ptr Chapel types when possible
- Emits 'int(64)' when encountering an 'int64_t'
- Attempts to emit helpful comments
- Allows user to pass flags to the C preprocessor
- Attempts to comment out typedefs from standard headers

Users can now ``cd`` into $CHPLHOME/tools/c2chapel and run
```make && make install``` to build c2chapel. This involves installing
pycparser to a local virtualenv and placing a symbolic link in $CHPL_HOME/bin.

Tests for c2chapel have also been added. There are a few tests in
$CHPL_HOME/test, but users can also run 'make check' from the
c2chapel directory to run a variety of simple test.

Special thanks to @npadmana , the original author of this tool.

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -133,6 +133,10 @@ chplvis: compiler third-party-fltk FORCE
 	cd tools/chplvis && $(MAKE)
 	cd tools/chplvis && $(MAKE) install
 
+c2chapel: FORCE
+	cd tools/c2chapel && $(MAKE)
```
ben-albrecht: Should these top-level make rules be documented?

tools/c2chapel/README.rst
```diff
@@ -0,0 +1,92 @@
+-----------------------------------------------------
```
benharsh: I didn't want to add it to the built docs until we include it in the release.

benharsh:
Just pushed some changes, let me know if you wanted something different.

Makefile
ben-albrecht: I see that you've added docs on `make clean`, etc., but this comment was referring to the `$CHPL_HOME`-level `make` targets. Are you also holding off on documenting these until this is included in the release?

benharsh: Ah, didn't catch that. I'll remove those top-level rules for now.

PR #6046, opened 2017-04-18
add nsw generation in LLVM IR for signed adds

coodie:
This PR adds generation of nsw instruction to Chapel's LLVM IR. This change is supposed to improve loop vectorization as discovered in #5788.

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b);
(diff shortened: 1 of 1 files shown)


mppf:
- [x] Passed full local testing with CHPL_LLVM=llvm
- [ ] Passed full local testing with CHPL_LLVM=llvm and --llvm

compiler/codegen/expr.cpp
```diff
@@ -1498,7 +1498,7 @@ GenRet codegenAdd(GenRet a, GenRet b)
       if(values.a->getType()->isFPOrFPVectorTy()) {
         ret.val = info->builder->CreateFAdd(values.a, values.b);
       } else {
-        ret.val = info->builder->CreateAdd(values.a, values.b);
+        ret.val = info->builder->CreateAdd(values.a, values.b, "", false, values.isSigned);
```
mppf: Could you add a comment near the additional arguments to say that the intent is to use an 'nsw' add for signed integers?

PR #6054, opened 2017-04-19
Update .bad for idealized chameneos

bradcray:
This test was failing due to issue #5935, which was fixed in PR #6053.
I expected it to have additional problems once we got past #5935, and
it seems it does.  I'm currently getting a seg-fault internal error with
the following stack trace:

````
isRecordWrappedType()
resolveFormals()
checkResolveFormalsWhereClauses()
filterInitConcreteCandidate()
````

So I updated the .bad to reflect this.


bradcray:
Attn: @lydia-duncan .  I'll file an issue against this unless you see any reason not to (e.g., if the fix is trivial).

lydia-duncan:
Oooh, fun.  I don't think the fix is going to be trivial, this looks to be similar to an error mode I've seen with one or two other generic records.

lydia-duncan:
(For more details, the problem is with *reusing* an instantiation of a generic record - we convert the 'this' argument to a reference, but then do something stupid when we try to resolve against that version the second time)

bradcray:
Created issue #6058 for this.

PR #6059, opened 2017-04-19
Updates to the issue template to simplify specifying futures

bradcray:
1) Encouraged people to use the "future tests" section rather than the
   inline sample code if a future exists already.  For me, it's
   streamlined filing the issue not to have to include the future code
   explicitly in the issue given that I'm typically going to file a
   future for my issues.

2) Changed the template reference to the future to include a hyperlink

3) Removed '-o foo' from the sample command-line because it's almost
   never necessary and doesn't add value that I can see.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -27,15 +27,15 @@ What behavior did you expect to observe?
 ```
 
 **Compile command:**
-<!-- e.g. ``chpl foo.chpl`` -->
+<!-- e.g. `chpl foo.chpl` -->
 
 **Execution command:**
-<!-- e.g. ``./foo -nl 4``
+<!-- e.g. `./foo -nl 4`
 If an input file is required, include it as well. -->
 
 **Associated Future Test(s):**
 <!-- Are there any tests in Chapel's test system that demonstrate this issue?
-     e.g. ``test/path/to/foo.future`` #1234 -->
+     e.g. `test/path/to/foo.future` #1234 -->
 
 ### Configuration Information
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -33,7 +33,7 @@ If an input file is required, include it as well. -->
 
 **Associated Future Test(s):**
 <!-- Are there any tests in Chapel's test system that demonstrate this issue?
-     e.g. [``test/path/to/foo.chpl``](
+     e.g. [`test/path/to/foo.chpl`](
            https://github.com/chapel-lang/chapel/blob/master/test/path/to/foo.chpl
           ) #1234 -->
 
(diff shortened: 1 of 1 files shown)


bradcray:
@ben-albrecht: I was going to ask you to take a glance at this before asking others for input, but before doing that, I was curious whether there was a way to preview it in the "file an issue" context prior to merging it.

ben-albrecht:
1. Looks good. One concern to keep in mind is that we don't want the template file to get too long / overwhelming, but I think we are still within a reasonable length at the moment.

2. I think single tics still work

3. Yeah, that was not valuable at all...

Viewing the `TEMPLATE.md` as raw text is how it would show up in the issue. If you wanted to get the full experience, you could commit it to your bradcray/chapel repository and open an issue against that. There may be an easier way to do this, e.g. switching your "main" branch for your repository, but I'm not sure if that works or not.

ronawho:
👍  for not inlining the content of futures, but I think it could be nice to link to them. 

e.g. 

``test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc-ideal.chpl``

vs. 

[test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc-ideal.chpl](https://github.com/chapel-lang/chapel/blob/master/test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc-ideal.chpl)

bradcray:
> but I think it could be nice to link to them.

I was wondering how hard this would be.  Thanks for providing an example.  Are you proposing updating the issue template to show this pattern in the comment for the futures section?  (I think it's a good idea, just making sure we're on the same page).

ronawho:
I'm for updating the template to include the linking (though I'm a little bummed out that github doesn't support relative linking.)

I was hoping we'd be able to do:

 ```
  [test/path/to/foo.chpll](test/path/to/foo.chpl) 
```

but instead we have to do:

```
  [test/path/to/foo.chpll](https://github.com/chapel-lang/chapel/blob/master/test/path/to/foo.chpl)
```

File: .github/ISSUE_TEMPLATE.md [hunk 1]
```diff
@@ -21,17 +21,21 @@ What behavior did you expect to observe?
...
 
 **Execution command:**
-<!-- e.g. `./foo -nl 4`. If an input file is required, include it as well. -->
+<!-- e.g. `./foo -nl 4`
+If an input file is required, include it as well. -->
 
 **Associated Future Test(s):**
 <!-- Are there any tests in Chapel's test system that demonstrate this issue?
-     e.g. test/path/to/foo.future #1234 -->
+     e.g. [``test/path/to/foo.chpl``](
```
ben-albrecht: Double back-tics by mistake?

bradcray:
@ronawho , @ben-albrecht : You guys OK with this?  Do you think I should put it in front of the whole team or just run with it and assume everyone will love it?

ben-albrecht:
Looks good to me.

PR #6060, opened 2017-04-19
Offset multi-ddata chunks at creation rather than at access time.

gbtitus:
The index computation for multi-ddata-chunked arrays first computes an
entire-array offset, then splits this into a chunk number and offset
within the chunk.  Here, after allocating the chunks, offset their
addresses backwards in memory so that we can use the original offset
into the entire array to access elements within each chunk, avoiding
having to subtract the chunk's starting offset on each access.  We still
record the chunk offset, because we need it to recreate the allocated
chunk address for certain sub-cases of bulk transfer and when we free
the chunks.

This results in a bit less than a 10% improvement in the serial case for
prk-stencil with locModel=numa, which had one of the worst performance
regressions due to multi-ddata.  However it's still something like 25x
slower with numa than with locModel=flat, so there is plenty of ground
to make up.


gbtitus:
@benharsh, would you please review this one?

PR #6061, opened 2017-04-19
Replace more Vecs with STL

benharsh:
Replaces various Vecs with std::set/vector/queue in the latter half
of the compiler.

Occasionally we iterate over a Vec while appending new elements. I
have handled this by using an std::queue.

An STL variant of the collectSymbolSet utility function was added to
support this change.

Testing:
- [x] full local
- [x] full gasnet


vasslitvinov:
Looks good.

I suggest emphasizing in the commit message that you handle the idiom of iterating over our homegrown Vec while simultaneously adding to it -- using std::queue .

PR #6064, opened 2017-04-20
Fix typos

cassella:
Fix typos in visibleFunctions, visibleCandidates, codegen.cpp, and the shootout README.


cassella:
In addition to letting github run the travis check, I ran a local build to validate the ifdef guard change.

ben-albrecht:
Thanks @cassella !

PR #6065, opened 2017-04-20
Use -std=c++11 when building re2, re2-interface.cc

mppf:
Addresses testing failures from PR #6024. The testing failures are because the 2017-04-01  version of RE2 uses some C++11 features and the previous version we were using did not.

This PR resolves these testing failures by:
 * introducing a Makefile variable, CXX11_STD, set in make/compiler/... to for example `-std=c++11`.
 * using CXX11_STD when compiling RE2 itself and when compiling re2-interface.cc
 * addressing other C++ warnings / errors from some tested configurations
    - defining true and false in chpltypes.h
    - removing #include <algorithm> from re2-interface.cc works around a problem with also including <intrinsics.h>
    - within RE2, fix a format string type mismatch

Verified quickstart+re2 build and test/regexp passes with:
- [x] Apple clang 8 on Mac OS X
- [x] Intel compiler version 15
- [x] Intel compiler version 16
- [x] Cray compiler version 8.4
- [x] Cray compiler version 8.5
- [x] GCC 4.7.3
- [x] GCC 4.8 on Ubuntu Trusty 32-bit
- [x] GCC 4.8 on CentOS 7
- [x] GCC 4.9

Configurations that worked before PR #6024 but still don't work after these fixes:
- PGI compiler version 16, 17: RE2 does not build
- GCC 4.6 on Ubuntu Precise 32-bit: -std=c++11 unrecognized command line option. Note that this distribution is End of Life this month (Apr 2017).
- GCC 4.4 on CentOS 6.8: unrecognized command line option "-std=c++11". Note that this distribution has End of Life in 2020.

Reviewed by @dmk42 - thanks!

### DIFF
--- a/make/compiler/Makefile.clang
+++ b/make/compiler/Makefile.clang
@@ -65,6 +65,7 @@ DEF_C_VER := $(shell echo __STDC_VERSION__ | $(CC) -E -x c - | sed -e '/^\#/d' -
 DEF_CXX_VER := $(shell echo __cplusplus | $(CXX) -E -x c++ - | sed -e '/^\#/d' -e 's/L$$//' -e 's/__cplusplus/0/')
 C_STD := $(shell test $(DEF_C_VER) -lt 199901 && echo -std=gnu99)
 CXX_STD := $(shell test $(DEF_C_VER) -ge 201112 -a $(DEF_CXX_VER) -lt 201103 && echo -std=gnu++11)
+CXX11_STD := -std=c++11
 
 #
 # Flags for compiler, runtime, and generated code
--- a/third-party/re2/re2-src/re2/nfa.cc
+++ b/third-party/re2/re2-src/re2/nfa.cc
@@ -700,8 +700,8 @@ bool NFA<StrPiece>::Search(const StrPiece& text, const StrPiece& const_context,
     for (int i = 0; i < nsubmatch; i++)
       submatch[i].set_ptr_end(match_[2 * i], match_[2 * i + 1]);
     if (ExtraDebug)
-      fprintf(stderr, "match (%td,%td)\n",
-              match_[0] - btext_, match_[1] - btext_);
+      fprintf(stderr, "match (%td,%ld)\n",
+              (long)(match_[0] - btext_), (long)(match_[1] - btext_));
     return true;
   }
   return false;
--- a/third-party/re2/re2-src/re2/nfa.cc
+++ b/third-party/re2/re2-src/re2/nfa.cc
@@ -700,8 +700,8 @@ bool NFA<StrPiece>::Search(const StrPiece& text, const StrPiece& const_context,
     for (int i = 0; i < nsubmatch; i++)
       submatch[i].set_ptr_end(match_[2 * i], match_[2 * i + 1]);
     if (ExtraDebug)
-      fprintf(stderr, "match (%td,%ld)\n",
-              (long)(match_[0] - btext_), (long)(match_[1] - btext_));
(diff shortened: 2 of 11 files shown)


third-party/re2/re2-src/re2/nfa.cc
```diff
@@ -700,8 +700,8 @@ bool NFA<StrPiece>::Search(const StrPiece& text, const StrPiece& const_context,
     for (int i = 0; i < nsubmatch; i++)
       submatch[i].set_ptr_end(match_[2 * i], match_[2 * i + 1]);
     if (ExtraDebug)
-      fprintf(stderr, "match (%td,%td)\n",
-              match_[0] - btext_, match_[1] - btext_);
+      fprintf(stderr, "match (%td,%ld)\n",
+              (long)(match_[0] - btext_), (long)(match_[1] - btext_));
```
dmk42: I don't think this change is right.  These should be type ptrdiff_t, which is what you get by default when you subtract two pointers.  The %td handles that.  Casting to long can chop off bits (in particular on 64-bit Windows).  I think this part of the code should be left as is.

The previous version of re2 was worse, though.  It did a static_cast<int> which will chop off bits in a lot of environments.

mppf: I changed it because I was getting compiler warnings. Our RE2 patches make NFA::Search a tempalet accepting StingPiece or FilePiece. When it's StringPiece, you are right that it's supposed to be ptrdiff_t. When it's FilePiece though, the subtraction is going to yield an int64_t.

I don't really care, since it's debugging code though. But do you have a suggestion for solving the problem in a better way (that gets us past the compiler error I was fixing?)

mppf: (the warnings were fatal errors in some cases)

dmk42: Ah, OK, I see.  In that case, instead of a C cast to long, I would suggest static_cast\<ptrdiff_t\> and keep the %t specifiers in the printf.

dmk42: Oh, wait, if it's a file offset and we're on a 32-bit machine, that will chop bits too.  I guess the best move is static_cast\<unsigned long long\> and use %llu in the printf.

dmk42: Sorry, there's an even better one.  static_cast\<uintmax_t\> and use %ju in the printf.

make/compiler/Makefile.clang
```diff
@@ -65,6 +65,7 @@ DEF_C_VER := $(shell echo __STDC_VERSION__ | $(CC) -E -x c - | sed -e '/^\#/d' -
 DEF_CXX_VER := $(shell echo __cplusplus | $(CXX) -E -x c++ - | sed -e '/^\#/d' -e 's/L$$//' -e 's/__cplusplus/0/')
 C_STD := $(shell test $(DEF_C_VER) -lt 199901 && echo -std=gnu99)
 CXX_STD := $(shell test $(DEF_C_VER) -ge 201112 -a $(DEF_CXX_VER) -lt 201103 && echo -std=gnu++11)
+CXX11_STD := -std=c++11
```
dmk42: How about this instead?

CXX11_STD := $(shell test $(DEF_CXX_VER) -lt 201103 && echo -std=gnu++11)

dmk42: This can wait until tomorrow.  Let's get testing back on track today.  OK to merge.

PR #6070, opened 2017-04-20
Makefile clean-ups for CFLAGS, CPPFLAGS, CXXFLAGS

mppf:
This PR performs the following changes: 
  1) Respect CPPFLAGS, CFLAGS, and CXXFLAGS in compiler and runtime builds.
  2) Rename some CFLAGS variables used to store C++ flags to CXXFLAGS variables
  3) Use `:=` when defining some Makefile variables that are computed by the shell

Updated the make/compiler/Makefiles to respect CPPFLAGS, CFLAGS, and CXXFLAGS in a more principled way. This is important for the Debian packaging effort. Note that CPPFLAGS refers to C pre-processor flags (CXXFLAGS is for C++).

To avoid confusing myself (and others editing these Makefiles), I renamed some variables to make it clearer the compiler flags are C++:
 *  COMP_CFLAGS to COMP_CXXFLAGS
 * COMP_CFLAGS_NONCHPL to COMP_CXXFLAGS_NONCHPL
 * CRAYPE_COMP_CFLAGS to CRAYPE_COMP_CXXFLAGS
The CFLAGS versions of these are no longer used anywhere.

Lastly, while working on these Makefiles, I noticed some shell-computed variables use `=` for assignment - which might mean that the shell command is re-evaluated every time the variable is used. Changed these to use `:=` which evaluates the RHS only once to remove this possibility.

(I tested this PR combined with #6086).

Verified quickstart+re2 build and test/regexp passes with:
- [x] Apple clang 8 on Mac OS X
- [x] Intel compiler version 15 (as host and target compiler)
- [x] Intel compiler version 16
- [x] Intel compiler version 17 (as host and target compiler)
- [x] Cray compiler version 8.4
- [x] Cray compiler version 8.5 (as host and target compiler)
- [x] GCC 4.7.3
- [x] GCC 4.8 on CentOS 7
- [x] GCC 4.8.5 on a Cray XC
- [x] GCC 4.9
- [x] GCC 6.3 on Ubuntu Zesty 17.04
- [x] ugni on a Cray XC
- [x] GASNet-udp on linux64
- [x] CHPL_LLVM=llvm on Mac OS X with and without --llvm
- [x] CHPL_LLVM=llvm on Linux with and without --llvm

And
- [x] full local testing

Reviewed by @dmk42 - thanks!


mppf:
@dmk42 - have a look & when it looks good to you I'll start testing...

dmk42:
Looks good.  After it passes testing, OK to merge.

dmk42:
Hmmm . . . the name CPPFLAGS is rather confusing.  It now looks like there are two sets of C\+\+ flags, CPPFLAGS and CXXFLAGS.  Preprocessor flags normally go in CFLAGS or CXXFLAGS, and CPPFLAGS is never set.  Do we really need that?

mppf:
@dmk42 CFLAGS, CXXFLAGS, and CPPFLAGS are not made up by us or me, but are rather standard names for Makefile variables.
See
 https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html
 http://man7.org/linux/man-pages/man1/dpkg-buildflags.1.html

CFLAGS: flags for C compilation
CXXFLAGS: flags for C++ compilation
CPPFLAGS: C *PreProcessor* flags, for C and C++ compilation

CPPFLAGS is not used anywhere in our code other than in the compiler support files where it is included in the other CFLAGS / CXXFLAGS variables.

dmk42:
Ah, I see.  It's a GNU-specific thing.  We can support it, but we should be careful not to depend on it.

mppf:
I'm going to rebase this on the current master to get a travis check & to clean up the diff.

PR #6080, opened 2017-04-21
Continue to straighten out var-args

noakesmichael:
Another sequence of mostly trivial commits in the effort to make it clearer
how var-arg functions are resolved.  This effort reminded me that I introduced
a "band-aid" during the string-as-rec update to work-around problems with
blank intents and ref types.  I need to come back to this.

Also did a spot check with c++14 enabled.

Passed a single-locale paratest


noakesmichael:
Another sequence of mostly trivial commits in the effort to make it clearer
how var-arg functions are resolved. This effort reminded me that I introduced
a "band-aid" during the string-as-rec update to work-around problems with
blank intents and ref types. I need to come back to this.

Also did a spot check with c++14 enabled.

Passed a single-locale paratest

Issue #6049, opened 2017-04-18
Semantic: Functions that return void values vs. return nothing


daviditen:
Should functions that do not return a value be equivalent to functions that explicitly return a void value?

```
proc f1() {
  writeln("f1");
  // no return
}

proc f2() {
  writeln("f2");
  return; // 0-argument return
}

proc f3() {
  writeln("f3");
  return _void;
}

var v1 = f1();
var v2 = f2();
var v3 = f3();
```
Should all three of the variable assignments be allowed?  Currently all of them are allowed because a function that has no return statement and a function that has a 0-argument return are considered equivalent to a function that returns the value `_void`. I think only the explicit `_void` return should really be allowed in this way.

This is similar to issue #5697, which asks a similar question but about iterators.

bradcray:
I tend to agree with David -- that only a function that returns an expression of void type should be able to be assigned to a variable (or used in a context that needs an expression).  I'm not whether the other cases should generate an error or a warning -- I just don't think they should be silently allowed.  A case that I ran into recently looked something like this:

````chapel
proc foo() {
  ...compute something...
  // meant to return something here but forgot to
}

writeln(foo());
````

compiled and ran but generated no output which puzzled me for awhile.  If non-returning functions used in such cases generated a warning or error, I would've caught my mistake much sooner.

noakesmichael:
I agree with David and Brad.

On Apr 18, 2017, at 2:56 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


I tend to agree with David -- that only a function that returns an expression of void type should be able to be assigned to a variable (or used in a context that needs an expression). I'm not whether the other cases should generate an error or a warning -- I just don't think they should be silently allowed. A case that I ran into recently looked something like this:

proc foo() {
  ...compute something...
  // meant to return something here but forgot to
}

writeln(foo());

compiled and ran but generated no output which puzzled me for awhile. If non-returning functions used in such cases generated a warning or error, I would've caught my mistake much sooner.

bradcray:
The thing I'm wrestling with is whether an error or warning is more appropriate.  Here are some sub-questions:

* Should we think of these different cases as defining functions with different type signatures?  Or do we think of them as simply different flavors of void functions?
* Maybe the same question in a slightly different way: If the user were providing fully typed signatures for these functions, would they all be of return type ': void' or would the two cases be different?
[...]

PR #6091, opened 2017-04-21
Functions that don't return vs. functions that return _void

daviditen:
Update functions that don't return values to no longer be usable as `_void` values.  If a function either doesn't have a return statement, or it has a return statement with no arguments, it cannot assigned to variables or used as actual arguments.  If a function explicitly returns the value `_void`, it can be used as a void variable.

Functions that may have a `void` return type and that should not be treated as a value are marked with the flag FLAG_VOID_NO_RETURN_VALUE.  If a function that returns `void` and has this flag is moved to a variable, an error is issued.

Extern functions with no declared type or a declared type of `void` are treated as functions with no returns, with one exception.  The exception is that extern functions with array arguments have a wrapper function created for handling the array arguments.  The wrapper function returns the result of calling the extern function. To allow this, void extern functions that take array arguments are treated as though they explicitly return `_void`.  The wrapper functions are then treated as though they did not return.

Changed tests that tried to capture the value of non-returning functions back to error tests with the same message they had before PR #5681.

This addresses issue #6049 as it currently stands.

Tested with a full paratest on Linux64 with --verify.


daviditen:
Passed full Linux64 paratest with --verify.

mppf:
Looks good to me. Don't you also need to update `doc/rst/technotes/voidVariables.rst` ?
Also I'm assuming that the discussion on the issue is sufficient for consensus purposes.

daviditen:
Thank you @mppf.  I updated the tech note.

Everybody seemed to be pretty well in agreement in the issue, so I think it's non-controversial at this point.

PR #6100, opened 2017-04-24
ACCU 2017 c-ray cleanup

bradcray:
This takes a pass over the c-ray exercise in preparation for the ACCU 2017 tutorial:

* wordsmithed the comments and, to a minor extent, the README

* added support for a "built-in" scene in order to avoid the need for file input, based on the deadlocks that some users reported at SC16

* fixed an apparent think-o in my channel.readlines() routine in which I was operating on 'infile' rather than 'this'.  I don't think that this could have resulted in the aforementioned deadlock (@mppf -- want to check my thinking on that, as I/O master?) but this seems like the right way to write it anyway.

* passed the input filename into the writeImage() routine and had it print what it was writing to to stdout (if it's not printing to stdout).

* moved the declarations of infile and outfile to local variables


mppf:
@bradcray - the readlines() change seems right, but I don't see any reason it should have deadlocked before. I think that was from an odd C library.

PR #6103, opened 2017-04-24
Address a problem with copy initializers for non-generic records

noakesmichael:
There is a semantic problem if a non-generic record defines an initializer with a single
generic argument.  The user might implement the initializer with the expectation that
they have an understanding of the actual type of the argument based on their uses of
new().  These assumptions may be invalid if the compiler specializes the initializer "as if" it
is a copy initializer.

This is a longer term issue that needs attention.  In the near term it may block the development
of some simple tests.  This PR provides a partial work-around that reduces the impact of
this issue while we continue to develop initializers.

The tactical problem is that function resolution implements the function 

static void resolveAutoCopyEtc(Type* type);

which generates calls to autoCopy/autoDestroy and unalias for every aggregate type.
The code for autoCopy generates a specialization using the generic initializer even if
the application does not require a copy initializer.

The longer term story for initializers should eliminate the use of autoCopy/autoDestroy
for record/class.  This PR disables the autoCopy expansion for non-generic records with
initializers.

The PR consists of a sequence of trivial commits.  The preliminary commits simplify the
implementation of resolveAutoCopies() and propagateNotPOD() without altering behavior.
The sequence is intended to make it easy to verify that the transformations are valid.

The penultimate commit actually disables the insertion of autoCopy for non-generic records.

The final commit retires one passing future and removes the existing .bad file for a future that
has changed behavior.

Compiled with/without CHPL_DEVELOPER on clang/darwin and gcc/linux64.  Also enabled C++14
for local testing on clang.

Passed a single-locale paratest with -futures and then with --no-local

### DIFF
--- a/compiler/AST/AstDumpToNode.cpp
+++ b/compiler/AST/AstDumpToNode.cpp
@@ -182,6 +182,7 @@ void AstDumpToNode::enterNode(BaseAST* node) const
     if (FnSymbol* fn = toFnSymbol(node))
     {
       fprintf(mFP, "%s%-10s", delimitEnter, node->astTagAsString());
+
       writeNodeID(node, true, false);
 
       if (fn->hasFlag(FLAG_GENERIC) == true)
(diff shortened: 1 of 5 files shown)


lydia-duncan:
This looks reasonable to me.  In addition to the testing you have already done, I would like to know the results of a valgrind run over the initializers directory, in case the prevention of autoCopy functions causes problems.  After that, I think this is fine to merge.

lydia-duncan:
It looks like the new output for initCallsDom.chpl is the same as the output I get in a future added by #6105

PR #6111, opened 2017-04-25
Work around broken __sync_bool_compare_and_swap under cce

ronawho:
This effectively reverts #6073. cce 8.4 added __sync_bool_compare_and_swap, but
it doesn't appear to be working correctly. Go back to using
`(__sync_val_compare_and_swap(obj, expected, desired) == expected)`


ronawho:
@dmk42 mind reviewing?

Note that we could use __sync_synchronize instead of __builtin_ia32_mfence still, but I don't think it's worth cluttering up the ifdef.

dmk42:
Just to leave a record, I OK'd this directly with Elliot before it was merged.

PR #6112, opened 2017-04-25
Refactor business logic to parse record/class/union 

noakesmichael:
Consider the following declaration statement

record Foo { ... }

Parsing this statement currently creates an AggregateType with tag AGGREGATE_RECORD.

Before this PR, the function buildClassDefExpr() accepted an AggregateType* and a list of fields.
The AggregateType* is allocated within the parser generated from chapel.ypp and without the
opportunity to inspect the fields.

After this PR, the function buildClassDefExpr() accepts an AggregateTag and a list of fields.
It continues to build the same AST object as before.  In the future it might choose to inspect
the fields and then construct a subtype of AggregateType (e.g. to construct a Record vs. a
GenericRecord.

Passed local testing on test/release and a single-locale paratest.

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -2244,46 +2244,61 @@ BlockStmt* buildVarDecls(BlockStmt* stmts, std::set<Flag> flags, const char* doc
 }
 
 
-DefExpr*
-buildClassDefExpr(const char* name,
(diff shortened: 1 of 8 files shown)


lydia-duncan:
Looks great!  Thanks

PR #6113, opened 2017-04-25
Use at least C++11 when recent enough compiler available

dmk42:
This does for the Chapel compiler what was already being done for the runtime.  If we are using a C\/C\+\+ compiler recent enough to default to C11, then we ensure that we use at least C\+\+11 to match.  However, we do not demote to C\+\+11 if the compiler defaults to something later.

This allows us to use some extra error checking, because fewer implicit type conversions are allowed for C\+\+11 and beyond.  It does not prevent us from testing with older standards; that only requires using an older compiler.

This shifts the amount of time to catch errors around a bit.  Errors caught by C\+\+11 will now be caught immediately, because clang has these checks.  (G\+\+ 7 will also perform the checking.)  Errors caused by using C\+\+11 features in the compiler will now be caught during the next nightly test instead of immediately.

This change is applied everywhere that we have this variable language standard scheme for the runtime:  the clang, gnu, and intel compilers.

Performed build and paratest with CHPL_HOST_COMPILER set to clang, gnu, and intel.
Tested with gcc 4.9, 5.4, and 6.3, which fall into the three separate prongs of the Makefile logic.
Tested with the latest Apple clang (approx. clang 3.9/4.0) and clang 3.3, which exercise the only two cases that clang exhibits to date.
Tested with v17 of the intel compiler, which exercises the only case that intel exhibits to date.
Also tested with and without CHPL_LLVM=llvm.

### DIFF
--- a/make/compiler/Makefile.clang
+++ b/make/compiler/Makefile.clang
@@ -70,7 +70,7 @@ CXX11_STD := $(shell test $(DEF_CXX_VER) -lt 201103 && echo -std=gnu++11)
 #
 # Flags for compiler, runtime, and generated code
 #
-COMP_CXXFLAGS = $(CPPFLAGS) $(CXXFLAGS)
+COMP_CXXFLAGS = $(CXX_STD) $(CPPFLAGS) $(CXXFLAGS)
 COMP_CXXFLAGS_NONCHPL = -Wno-error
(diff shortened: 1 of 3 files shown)


dmk42:
@ronawho and @benharsh , would you mind checking to make sure that this is OK with you?

ronawho:
👍

benharsh:
👍

PR #6116, opened 2017-04-26
Add note about GCC 5 in re2 README, prereqs.rst

mppf:
* Updates prereqs.rst to note GCC 5 for C++11 dependency in third-party
* Clarifies some error wording in Regexp.chpl
* Updates third-party/re2/README to note C++11 dependency and suggest GCC 5 and say why.

Reviewed by @dmk42 and @ben-albrecht - thanks!

### DIFF
--- a/doc/rst/usingchapel/prereqs.rst
+++ b/doc/rst/usingchapel/prereqs.rst
@@ -29,6 +29,10 @@ about your environment for using Chapel:
     relatively recent versions of gcc/g++, clang, and compilers from
     Cray, Intel, and PGI.
 
+    * Note that you will need a C++11 compiler to build LLVM or regular
+      expression support (i.e.  CHPL_LLVM=llvm or CHPL_REGEXP=re2). If
+      GCC is used, we recommend GCC version 5 or newer for this purpose.
+
   * If you wish to use Chapel's test system, python-setuptools and
     python-devel (or equivalent packages for your platform) are required.
 
--- a/modules/standard/Regexp.chpl
+++ b/modules/standard/Regexp.chpl
@@ -42,9 +42,9 @@ in the Chapel distribution.
 
 .. note::
 
-  if re2 support is not enabled (which is the default), all features
-  described below will compile successfully, but will result in an internal
-  error at *run time*, saying "No Regexp Support".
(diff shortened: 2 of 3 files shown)


mppf:
@dmk42 - how does this look?

dmk42:
I agree with this change.  I would like to see an opinion from @ronawho .

ronawho:
I think this needs to be somewhere way more visible like prereqs.rst

doc/rst/usingchapel/prereqs.rst
```diff
@@ -29,6 +29,10 @@ about your environment for using Chapel:
     relatively recent versions of gcc/g++, clang, and compilers from
     Cray, Intel, and PGI.
 
+    * Note that you will need a C++11 compiler to build LLVM or regular
+      expression support (i.e.  CHPL_LLVM=llvm or CHPL_REGEXP=re2). If
+      GCC is used, we recommend GCC version 5 or newer for this purpose.
+
```
dmk42: Also, when gmp builds on chapcs, I see it compile in C++11 mode.  Should something about gmp be mentioned here?  I think on other systems it sometimes uses the system gmp and doesn't need C++11, so the situation might be complicated.

dmk42: Nope, it looks like that was my mistake.  I was misreading my log from the build.

dmk42:
:+1:

modules/standard/Regexp.chpl
```diff
@@ -42,9 +42,9 @@ in the Chapel distribution.
 
 .. note::
 
-  if re2 support is not enabled (which is the default), all features
-  described below will compile successfully, but will result in an internal
-  error at *run time*, saying "No Regexp Support".
+  if re2 support is not enabled (which is the case in quickstart
```
ben-albrecht: Could possibly cross-reference `:ref:quickstart-with-other-shells` to make it clear what 'quickstart' is.

mppf:
@ben-albrecht - it made more sense to me to link to the whole Quickstart document. Sound OK? Otherwise, I think we'd need to add a paragraph somewhere like "What is Quickstart mode?".

ben-albrecht:
Sounds good! LGTM

Issue #5697, opened 2017-03-16
Semantic: Iterators that yield void values vs. yield nothing


daviditen:
### Summary
What should happen with an iterator that does not yield anything?  It is currently a compiler error: "illegal use of a function that does not return a value".  With PR #5681 it turns into the error: "cannot iterate over values of type void".  It could be treated as if it yields a single void value.  @bradcray has suggested that void iterators should work as in this example:

```
iter yieldTenVoids() {
  for i in 1..10 do
    yield _void;
}

iter yieldAVoid() {
  yield _void;
}

iter yieldNothing() {
  writeln("In my iterator!");
}

for v in yieldTenVoids() do
  writeln("Got a void!");    // should print this message ten times

for v in yieldAVoid() do
  writeln("Got a void!");    // should print this message one time

for v in yieldNothing() do
  writeln("Got a void!");    // should not print the loop body's message at all, but should still print the iterator's message
```

bradcray:
When I first thought about this (and commented on the PR), I thought it should collapse the loop into a no-op, but now I think it should invoke the iterator but not the loop body (as in the third iterator / loop example above).  I've edited the text in the issue description slightly to clarify this.

My personal compiler may currently just be a mess, but it seems like if I comment out the yieldNothing()-related code above and compile it today, I get an internal error...?

vasslitvinov:
What is the motivation to allow non-yielding iterators? Would it result in allowing subtle bugs like "I meant to put a yield and forgot" or "I expect my loop body to be executed at least once and it wasn't" ?

daviditen:
IMO the motivation is mostly orthogonality.  It is fine to write and use an iterator that doesn't yield anything dynamically, but we currently require it to have an unused yield statement:

```
iter sometimesYieldNothing(doYield: bool) {
  writeln("in sometimesYieldNothing");
  if doYield {
    yield 1;
  }
}

for i in sometimesYieldNothing(true) {}  // OK
for i in sometimesYieldNothing(false) {} // OK

iter neverYield() {
  writeln("in neverYield");
}

for i in neverYield() {} // not OK
```

I think the `neverYield()` case is analogous to the `sometimesYieldNothing(false)` case, and it seems natural to want them to be equivalent.  It's even more equivalent if the `doYield` argument is a param, so there really isn't a yield when it's false.

[...]

PR #6117, opened 2017-04-26
Allow iterators that explicitly yield '_void' and non-yielding iterators

daviditen:
Change yield statements that yield values with 'void' type to yield the global
`_void` value.  Change for loop index variables that are type `void` to refer
to the global `_void` value.  This gets two of the three cases in issue #5697
working as described in the issue.  Each explicitly yielded `_void` makes the loop
iterate once with a `_void` index.

Remove a few checks that disallow iterators with no yields.  This allows
non-yielding iterators to be used, executing the code in the iterator, but not
executing any loop body iterations.  This is the third case in issue #5697.

Since most void-typed symbols are removed at the end of function resolution,
the few places that remain that still need void values (e.g. yield statements that
yield _void and loop index variables for loops over _void) use the global void
value named gVoid in the compiler.

Added two tests that exercise the three cases from the issue.  Updated a .good
file for a test that expected an error from a yield-nothing iterator.

### DIFF
--- a/compiler/passes/checkParsed.cpp
+++ b/compiler/passes/checkParsed.cpp
@@ -304,8 +304,6 @@ checkFunction(FnSymbol* fn) {
 
   if (numVoidReturns != 0 && numNonVoidReturns != 0)
     USR_FATAL_CONT(fn, "Not all returns in this function return a value");
-  if (isIterator && numYields == 0)
-    USR_FATAL_CONT(fn, "iterator does not yield a value");
   if (!isIterator &&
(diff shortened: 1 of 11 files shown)


daviditen:
Passed full paratest on Linux64 with --verify.

vasslitvinov:
Please add a .future file for yieldNothingIterator.chpl (assuming it's needed).

daviditen:
Oops, I removed that .future, but forgot to remove the .bad file.

PR #6118, opened 2017-04-26
Fix an out-of-bounds array access in the lexer

daviditen:
If the `--comment-style` flag is not used, then `fDocsCommentLabel` is an empty
string with length=0. Make sure it is not zero-length before indexing into it
with `len - d` with `d` positive.

Regenerated flex-chapel.cpp and flex-chapel.h after making the above change.

### DIFF
--- a/compiler/include/flex-chapel.h
+++ b/compiler/include/flex-chapel.h
@@ -51,6 +51,7 @@ typedef int16_t flex_int16_t;
 typedef uint16_t flex_uint16_t;
 typedef int32_t flex_int32_t;
 typedef uint32_t flex_uint32_t;
+typedef uint64_t flex_uint64_t;
 #else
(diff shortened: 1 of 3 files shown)


daviditen:
Passed a full default paratest on Linux64.

lydia-duncan:
This looks reasonable to me.  I'm surprised this wasn't caught earlier, though.  Would you mind adding a test case that failed before but works now?

daviditen:
Thanks Lydia!

I discovered this by building the compiler with address sanitation with clang, then compiling release/examples/hello.chpl. I don't currently have a test that fails because of it without address sanitation enabled, but essentially everything fails with it.

lydia-duncan:
Ah, okay!  Good to go, then!

PR #6126, opened 2017-04-27
Rewrite paratest.chapcs in python

ronawho:
I need to make some changes to paratest.chapcs, but they're going to be
hard/ugly to do in bash, so rewrite the script in python. Note that this is a
pretty direct transliteration for now and should behave identically to the
current version. Improvements and bug fixes will be made in later PRs

Note that the python version is being added as paratest.chapcs.py for now. Once
I have more confidence that it behaves correctly it'll replace paratest.chapcs

### DIFF
--- a/util/chplenv/utils.py
+++ b/util/chplenv/utils.py
@@ -62,3 +62,20 @@ def run_command(command, stdout=True, stderr=False, cmd_input=None):
         else:
             return ''
 
+def run_live_command(command):
+    """Run a command, writing its output live instead of after the process is
+       finished. stderr is merged with stdout"""
+    try:
+        process = subprocess.Popen(command,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.STDOUT)
+    except OSError:
+        error("command not found: {0}".format(command[0]), OSError)
+
+    for stdout_line in iter(process.stdout.readline, str.encode("")):
+        yield stdout_line.decode()
+    process.stdout.close()
+    returncode = process.wait()
+
+    if returncode != 0:
+        error("command failed: {0}".format(command), CommandError)
--- a/util/chplenv/utils.py
+++ b/util/chplenv/utils.py
@@ -63,8 +63,8 @@ def run_command(command, stdout=True, stderr=False, cmd_input=None):
             return ''
 
 def run_live_command(command):
-    """Run a command, writing its output live instead of after the process is
-       finished. stderr is merged with stdout"""
+    """Run a command, yielding the merged output (stdout/stderr) as the process
+       runs rather than returning the output after the process finishes"""
     try:
         process = subprocess.Popen(command,
                                    stdout=subprocess.PIPE,
--- a/util/test/paratest.chapcs.py
+++ b/util/test/paratest.chapcs.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python
+
+"""
+Simple slurm-aware paratest wrapper for testing on the chapcs cluster. This
+script aims to run parallel testing as quickly as possible without interfering
+with nightly testing or other exclusive reservations (e.g.  developers running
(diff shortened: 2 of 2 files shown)


ronawho:
@ben-albrecht and @benharsh mind reviewing?

I'm especially interested to see if you have a better name for `run_live_command`

util/chplenv/utils.py
```diff
@@ -62,3 +62,20 @@ def run_command(command, stdout=True, stderr=False, cmd_input=None):
         else:
             return ''
 
+def run_live_command(command):
+    """Run a command, writing its output live instead of after the process is
```
ben-albrecht: **comment nit-pick:**

 It's really *yielding* the merged output (stdout/stderr) as the process runs, rather than *writing* it live. 

**idea:**

A potential alternative to the iterator approach: Write the merged output to stdout within the function and return the output once the process completes, providing a simple interface similar to `run_command`. 

You could take it a step further and make this function accessible through a `live=False` argument in `run_command` and avoid the hard decision of choosing a name altogether.

util/test/paratest.chapcs.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+    start_time = timeit.default_timer()
+    for line in utils.run_live_command(paratest_cmd):
+        sys.stdout.write(line)
+    elapsed = int(timeit.default_timer() - start_time)
+    minutes, seconds = divmod(elapsed, 60)
+    print('paratest took {0} minutes and {1} seconds'.format(minutes, seconds))
+
+run_paratest()
```
ben-albrecht:
```python

if __name__ == '__main__':
    run_paratest()
```

Not necessary, but this is good practice to enable importing of this module elsewhere, e.g. unit testing.

ben-albrecht: I've generally been trying to include help flags on all scripts in the repository.

It might also be nice to catch the `--help` flag instead of forwarding it to the paratest, e.g.

```python

if '-h' in sys.argv or '--help' in sys.argv:
    print(__doc__)
    return
```

util/chplenv/utils.py
ronawho: Will adjust the docs.


I was originally trying to make `live` an argument to run_command, but I gave up on that because the stdin stuff got weird and because it doesn't really make sense to ignore stdout/stderr for the live version. i.e. in order to support stdin with the live version, you end up having to do `stdin.write(input)`, which isn't recommended. communicate() is recommended to avoid deadlock due to OS pipe buffers filling up, but you can't do live output with communicate so I ended up just creating a second function that doesn't accept intput. There's also no reason to use the live version if you're going to ignore the output so I also removed the stdout and stderr arguments

util/test/paratest.chapcs.py
ronawho: I'll change to `if __name__ == '__main__':`

The --help flag is a good idea, but I'll punt on that for the moment and put that in a future update. Just for testing sanity I want the first approach to be a direct replacement for what we have now.

util/test/paratest.chapcs.py
```diff
@@ -0,0 +1,127 @@
...
+    return [line.strip().strip('"') for line in lines.splitlines()]
+
+
+def get_num_exclusive_nodes():
+    """
+    Get the number of nodes reserved exclusively on the chapel partition
+    """
+    # Grab the "SHARED,NODES" info for all jobs. For each exclusive (SHARED=no)
+    # job add up the number of nodes it's using
+    squeue_cmd = ['squeue',
```
ben-albrecht: Not sure if it is necessarily better, but you could alternatively maintain these as strings, and wrap them with `shlex.split()` when passing to `run_command`. 

This would also remove the necessity of `strip_quotes` helper function.

PR #6130, opened 2017-04-27
Generate file-unique ID for GETs and PUTs

benharsh:
While the line number and filename of a Chapel program is useful output from VisualDebug, it doesn't really help developers map gets/puts to the generated sources. Many gets/puts may share line/file info if they are within an iterator, for example.

This PR adds a new argument to the get/put runtime calls, called ``commID``, which is a unique integer ID for each get/put. This ID is unique within a file. In the generated C, this integer is wrapped in the new COMMID macro for readability.

I incremented the output version from 1.2 to 1.3, and updated ``chplvis`` to handle the new format. Otherwise there is no change to chplvis. The ``commID`` is not displayed anywhere in the GUI.

In the chpl-visual-debug.c file, I add code to compute the length of the get/put.

Otherwise, there are some minor style/formatting cleanups.

Testing:
- [x] test/chplvis and chplvis primers
  - [x] mac
  - [x] linux
  - [x] ugni
- [x] cache-remote
- [x] local paratest
- [x] gasnet paratest

### DIFF
--- a/runtime/include/chpl-comm-compiler-llvm-support.h
+++ b/runtime/include/chpl-comm-compiler-llvm-support.h
@@ -40,7 +40,7 @@ void chpl_gen_comm_get_ctl(void *dst_addr, wide_ptr_t src, int64_t n, int64_t ct
 {
   c_nodeid_t src_node = chpl_wide_ptr_get_node(src);
   void* src_addr = chpl_wide_ptr_get_address(src);
-  chpl_gen_comm_get(dst_addr, src_node, src_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, -1, 0);
+  chpl_gen_comm_get(dst_addr, src_node, src_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, CHPL_UNKNOWN_COMM_ID, -1, 0);
 }
 
 static inline
@@ -48,7 +48,7 @@ void chpl_gen_comm_put_ctl(wide_ptr_t dst, void *src_addr, int64_t n, int64_t ct
 {
   c_nodeid_t dst_node = chpl_wide_ptr_get_node(dst);
   void* dst_addr = chpl_wide_ptr_get_address(dst);
-  chpl_gen_comm_put(src_addr, dst_node, dst_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, -1, 0);
+  chpl_gen_comm_put(src_addr, dst_node, dst_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, CHPL_UNKNOWN_COMM_ID, -1, 0);
 }
 
 // This function implements memcpy/memmove when both the source and destination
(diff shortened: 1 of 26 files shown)


benharsh:
@gbtitus , @tomyhoi : this may be of interest to you.

runtime/include/chpl-comm-compiler-llvm-support.h
```diff
@@ -40,15 +40,15 @@ void chpl_gen_comm_get_ctl(void *dst_addr, wide_ptr_t src, int64_t n, int64_t ct
 {
   c_nodeid_t src_node = chpl_wide_ptr_get_node(src);
   void* src_addr = chpl_wide_ptr_get_address(src);
-  chpl_gen_comm_get(dst_addr, src_node, src_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, -1, 0);
+  chpl_gen_comm_get(dst_addr, src_node, src_addr, sizeof(uint8_t)*n, CHPL_TYPE_uint8_t, CHPL_UNKNOWN_COMM_ID, -1, 0);
```
gbtitus: I'd prefer CHPL_COMM_ID_UNKNOWN or CHPL_COMM_UNKNOWN_ID here, just to group this symbol with the rest of the (case insensitive) chpl_comm_* ones.

runtime/include/chpl-comm-compiler-macros.h
```diff
@@ -53,9 +55,9 @@ void chpl_gen_comm_get(void *addr, c_nodeid_t node, void* raddr,
 #endif
   } else {
 #ifdef CHPL_TASK_COMM_GET
-    chpl_task_comm_get(addr, node, raddr, size, typeIndex, ln, fn);
+    chpl_task_comm_get(addr, node, raddr, size, typeIndex, commID, ln, fn);
```
gbtitus: I think you need to do the chpl_cache_comm_*() interface functions as well (above).

runtime/src/chpl-visual-debug.c
```diff
@@ -66,6 +66,9 @@ int chpl_vdebug = 0;
 
 #define TID_STRING(buff, tid) (chpl_task_idToString(buff, CHPL_TASK_ID_STRING_MAX_LEN, tid))
 
+#define CHPL_VDEBUG_GETPUT_FORMAT_NAMES "tv srcNodeID dstNodeID commTaskID addr raddr elemSize typeIndex length commID lineNumber fileno"
+#define CHPL_VDEBUG_GETPUT_FORMAT_STRING(name) #name ": %lld.%06ld %d %d %s %#lx %#lx %zd %d %zd %d %d %d\n"
```
gbtitus: This is a good change, but you can do without the leading "CHPL_" because it's not a global symbol.  Also, it might be a bit easier to read if you just did something like:

    #define VDEBUG_GETPUT_FORMAT_STRING "%s: %lld.%06ld %d %d %s %#lx %#lx %zd %d %zd %d %d %d\n"

and then in usage, something like:

    printf(VDEBUG_GETPUT_FORMAT_STRING, "nb_put", ...

gbtitus:
Looks good!  I made a few line comments and for sure the chpl_cache_comm_*() functions also need to get the comm ID, but this seems fine otherwise.

gbtitus:
👍

PR #6133, opened 2017-04-28
Enable use of system jemalloc

mppf:
This PR adds support for using a system jemalloc library. This are intended to support the Debian packaging effort, but is useful more generally.

It enables:
 * CHPL_HWLOC=system
 * CHPL_MEM=jemalloc CHPL_JEMALLOC=system

Details:

 * Originally, this PR added CHPL_MEM=jemalloc-system. But that is problematic because it both specifies the memory layer to use (jemalloc) and also specifies where to get it.
  
   Instead of doing that, add a CHPL_JEMALLOC variable that is only useful when CHPL_MEM=jemalloc. That way, each hase one meaning:
     * CHPL_MEM=jemalloc -> use jemalloc
     * CHPL_JEMALLOC -> where to get jemalloc

   Adding a CHPL_JEMALLOC variable involved making these changes:
       - adding util/chplenv/chpl_3p_jemalloc_configs.py
       - calling get_link_args from util/chplenv/chpl_3p_jemalloc_configs.py in printchlenv
       - makefiles:
            - runtime/etc/Makefile.jemalloc-`$CHPL_JEMALLOC`
            - call those from runtime/etc/Makefile.include
            - runtime/make/Makefile.runtime.jemalloc-`$CHPL_JEMALLOC`
            - call those from runtime/make/Makefile.runtime.include

 * To work with a system-installed jemalloc, besides adding Makefiles and scripting to support CHPL_MEM=jemalloc-system, I needed to update some of our C code working with jemalloc in chpl-mem-impl.h and mem-jemalloc.c:

   * update our calls to jemalloc routines to use a configurable prefix (currently chpl_je_ or je_)
   * use #defines to make it compile with jemalloc 3.6
     *  3.6 is the default on Ubuntu 17.04 and Debian sid (so this version is important for the Debian packaging effort)
     * 3.6 doesn't include chunk_hooks_t and doesn't define `bool`
     * we use chunk_hooks_t for multilocale + segment=fast
     * I made it compile but fail at runtime in a multilocale + segment=fast configuration
   * use -I to the directory containing jemalloc/jemalloc.h since the system install has /usr/include/jemalloc/jemalloc.h - that is, we now include "jemalloc/jemalloc.h" instead of just "jemalloc.h"

(note, system packages were tested with Ubuntu 17.04 packages)

`make clobber && make && make check` with:
- [x] quickstart config (system jemalloc not installed)
- [x] quickstart config + CHPL_MEM=jemalloc (system jemalloc not installed)
- [x] quickstart config (system jemalloc installed)
- [x] quickstart config + CHPL_MEM=jemalloc (system jemalloc installed)
- [x] quickstart config + CHPL_MEM=jemalloc CHPL_JEMALLOC=system

Verified that quickstart + CHPL_MEM=jemalloc CHPL_JEMALLOC=system does not build anything in third-party.

Verified an error whet CHPL_JEMALLOC=system but system libjemalloc-dev is not installed
 (`fatal error: jemalloc/jemalloc.h: No such file or directory`)

Passed full local testing.

Reviewed by @ronawho and @gbtitus. Discussed overall strategy with @ben-albrecht and @bradcray.


mppf:
@ronawho - I think this version addresses your feedback. Please let me know if you have other feedback.

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -155,6 +157,11 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
 // *** Chunk hook replacements *** //
 // See http://www.canonware.com/download/jemalloc/jemalloc-latest/doc/jemalloc.html#arena.i.chunk_hooks
 
+// jemalloc < 4.0 didn't support chunk_hooks_t
+// avoid unused function warnings by only defining these if they could be used
+#if JEMALLOC_VERSION_MAJOR >= 4
```
ronawho: Should check for 4.1 instead of 4.0 -- 4.1 changed `opt.narenas` from `size_t` to `unsigned`so our code won't work with < 4.1

https://github.com/jemalloc/jemalloc/releases/4.1.0 -- Incompatible changes sections

https://github.com/chapel-lang/chapel/commit/c20bdc7f5f7cfe75cfe8999a12e23fdcfde92ce1

We _could_ change our shim to deal with the incompatibly, but realistically I can't imagine that anybody will end up using a version between 4.0 and 4.1.

ronawho:
Can you split out just the jemalloc changes so it's easy to discuss/think about without the hwoc/re2 changes mixed in?

mppf:
@ronawho - sure, the hwloc changes are now in PR #6148 and this one is focused on jemalloc.

util/printchplenv
```diff
@@ -111,6 +111,9 @@ def print_mode(mode='list', anonymize=False):
     hwloc = chpl_hwloc.get()
     print_var('CHPL_HWLOC', hwloc, mode, 'hwloc', ('runtime',))
 
+    jemalloc = chpl_jemalloc.get()
+    print_var('CHPL_JEMALLOC', jemalloc, mode, 'jemalloc', ('runtime',))
```
ronawho: Can we print this only if CHPL_MEM=jemalloc?

util/printchplenv
```diff
@@ -156,8 +159,7 @@ def print_mode(mode='list', anonymize=False):
         print_var('  CHPL_JEMALLOC_UNIQ_CFG_PATH',
                   chpl_3p_jemalloc_configs.get_uniq_cfg_path(),
                   mode, filters=('runtime',))
-        if target_mem == 'jemalloc':
-            link_args_3p.extend(chpl_3p_jemalloc_configs.get_link_args())
+        link_args_3p.extend(chpl_3p_jemalloc_configs.get_link_args(jemalloc))
```
ronawho: Why get rid of the `target_mem == 'jemalloc'` check?

util/chplenv/third-party-pkgs
```diff
@@ -65,7 +65,8 @@ for l in chplenv.split('\n'):
         var=var.strip()
         val=val.strip()
         # Ignore platform variables
-        if not var.endswith('_PLATFORM'):
+        # Ignore CHPL_MEM (use CHPL_JEMALLOC)
+        if not (var.endswith('_PLATFORM') or var == 'CHPL_MEM'):
```
ronawho: How does this "ignore" CHPL_MEM?

util/chplenv/chpl_3p_jemalloc_configs.py
```diff
@@ -21,10 +21,15 @@ def get_jemalloc_config_file():
...
+def get_link_args(target_mem):
+    if target_mem == 'jemalloc':
+        jemalloc_config = get_jemalloc_config_file()
+        libs = ['-ljemalloc']
+        if os.access(jemalloc_config, os.X_OK):
+            jemalloc_libs = run_command([jemalloc_config, '--libs'])
+            libs += jemalloc_libs.split()
+        return libs
+    elif target_mem == 'system':
+        return ['-ljemalloc']
```
ronawho: Don't we still need to get the link args from jemalloc somehow? Prior to jemalloc 4.1. the only link arg was probably -lpthread, but if the system version ever gets upgraded, don't we need a way of detecting jemallocs link args?

util/chplenv/third-party-pkgs
mppf: I added `or var == 'CHPL_MEM' ` so if CHPL_MEM=jemalloc, we don't build third-party/jemalloc

util/chplenv/chpl_3p_jemalloc_configs.py
mppf: I don't see any way to get that information out of the jemalloc package on Debian, so I don't think we have to worry about it right now. Anyway it seems to work without it.

Is there some link arg that is particularly relevant with versions >= 4.1 ?

util/chplenv/third-party-pkgs
ronawho: oh duh, was reading this backwards

runtime/make/Makefile.runtime.jemalloc-jemalloc
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+RUNTIME_INCLS += -I$(JEMALLOC_INCLUDE_DIR)
+# Chapel third-party install of jemalloc calls the routines
+# chpl_je_malloc, chpl_je_free, etc
+RUNTIME_DEFS += -DCHPL_JEMALLOC_PREFIX=chpl_je_
```
ronawho: Move this into third-party/jemalloc/Makefile.include and use it to set `--with-jemalloc-prefix=`?

runtime/include/mem/jemalloc/chpl-mem-impl.h
```diff
@@ -23,41 +23,62 @@
...
-  chpl_je_free(ptr);
+  CHPL_JE_(free)(ptr);
 }
 
 static inline size_t chpl_good_alloc_size(size_t minSize) {
   if (minSize == 0) { return 0; }
-  return chpl_je_nallocx(minSize, MALLOCX_NO_FLAGS);
+  return CHPL_JE_(nallocx)(minSize, MALLOCX_NO_FLAGS);
+  //return nallocx(minSize, MALLOCX_NO_FLAGS);
+  //return je_nallocx(minSize, MALLOCX_NO_FLAGS);
```
ronawho: leftover?

util/printchplenv
mppf: Philosophically, I'd prefer to have chpl_3p_jemalloc_configs.get_link_args be totally responsible for any use of jemalloc. However everything would work OK with the target_mem condition. I just viewed adding CHPL_JEMALLOC as separating out these portions. E.g. if we ever had a use of jemalloc that was not dependent on CHPL_MEM.

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -232,6 +250,9 @@ static bool null_split(void *chunk, size_t size, size_t size_a, size_t size_b, b
 static bool null_merge(void *chunk_a, size_t size_a, void *chunk_b, size_t size_b, bool committed, unsigned arena_ind) {
   return true;
 }
+
+#endif // jemalloc >= 4
```
ronawho: out of date comment

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -428,13 +457,14 @@ void chpl_mem_layerInit(void) {
   //   jemalloc 4.5.0 man: "Once, when the first call is made to one of the
   //   memory allocation routines, the allocator initializes its internals"
   if (heap_base != NULL) {
+
```
ronawho: extra newline?

util/printchplenv
mppf: We could, but only for certain configurations (e.g. not in printchplenv --make). I'm half a mind to leave it alone though (like CHPL_HWLOC prints with or without qthreads). The other precedent is CHPL_COMM_SUBSTRATE etc which only print with CHPL_COMM != none.

Is it obvious to you what we should do here?

util/chplenv/chpl_3p_jemalloc_configs.py
ronawho: -lrt

util/chplenv/third-party-pkgs
mppf: instead we will build it by relying on `CHPL_JEMALLOC=jemalloc`

doc/rst/usingchapel/chplenv.rst
```diff
@@ -372,12 +372,14 @@ CHPL_MEM
...
-        ========= =======================================================
-        cstdlib   use the standard C malloc/free commands
-        jemalloc  use Jason Evan's memory allocator
-        ========= =======================================================
+        =============== =======================================================
+        Value           Description
+        =============== =======================================================
+        cstdlib         use the standard C malloc/free commands
+        jemalloc        use included version of Jason Evan's memory allocator
+        jemalloc-system use a system install of jemalloc
```
ronawho: Is this still right? I thought it was now `CHPL_MEM=jemalloc` `CHPL_JEMALLOC=system`

util/chplenv/chpl_3p_jemalloc_configs.py
mppf: I see some Debian packages install .la files to /usr/lib and others install .pc files to /usr/lib/<arch>/pkgconfig. I think it would be better to worry about this issue once there is a jemalloc package needing -lrt on some system that we're trying to work with. That way, we can test it & I don't have to predict what the package maintainers will do.

doc/rst/usingchapel/chplenv.rst
mppf: right, thanks, good catch

runtime/make/Makefile.runtime.jemalloc-jemalloc
mppf: That seems to me to be more trouble than it's worth because we also have to work with --with-private-namespace so it's not a simple trade for 1 variable. Anyway it would add a new Makefile variable and I don't know about you but I'm more likely to be confused by one of those than to be confused by needing to change chpl_je_ in two places.

ronawho: --with-private doesn't impact anything beyond jemalloc internals though. 

I was just thinking of having `CHPL_JEMALLOC_PREFIX=chpl_je` in third-party/jemalloc/Makefile.include. Then here doing `RUNTIME_DEFS += -DCHPL_JEMALLOC_PREFIX=$(CHPL_JEMALLOC_PREFIX)` 

Not sure I'm following what makes that more confusing

mppf: OK, I changed it.

doc/rst/usingchapel/chplenv.rst
```diff
@@ -508,6 +508,28 @@ CHPL_HWLOC
    to ``none`` and rebuild (and please file a bug with the Chapel team.) Note
    that building without hwloc will have a negative impact on performance.
 
+.. _readme-chplenv.CHPL_JEMALLOC:
+
+CHPL_JEMALLOC
+~~~~~~~~~~~~~
+   Optionally, the ``CHPL_JEMALLOC`` environment variable can select
+   between no jemalloc, using the jemalloc distributed with Chapel in
+   third-party, or using a system jemalloc. This setting is intented
```
gbtitus: intended

doc/rst/usingchapel/chplenv.rst
```diff
@@ -508,6 +508,28 @@ CHPL_HWLOC
...
+   to elaborate upon ``CHPL_MEM=jemalloc``.
+
+       ======== ==============================================================
+       Value    Description
+       ======== ==============================================================
+       none     do not build or use jemalloc
```
gbtitus: Is CHPL_JEMALLOC even referenced if CHPL_MEM != jemalloc and, if so, for what purpose?  It seems like CHPL_JEMALLOC=none shouldn't be necessary.

runtime/etc/Makefile.include
```diff
@@ -63,6 +63,7 @@ include $(CHPL_MAKE_HOME)/runtime/etc/Makefile.mem-$(CHPL_MAKE_MEM)
 -include $(CHPL_MAKE_HOME)/runtime/etc/Makefile.unwind-$(CHPL_MAKE_UNWIND)
 include $(CHPL_MAKE_HOME)/runtime/etc/Makefile.gmp-$(CHPL_MAKE_GMP)
 -include $(CHPL_MAKE_HOME)/runtime/etc/Makefile.hwloc-$(CHPL_MAKE_HWLOC)
+-include $(CHPL_MAKE_HOME)/runtime/etc/Makefile.jemalloc-$(CHPL_MAKE_JEMALLOC)
```
gbtitus: Could this include be moved to .../runtime/etc/Makefile.mem-jemalloc?

doc/rst/usingchapel/chplenv.rst
mppf: My intention was to make CHPL_JEMALLOC be a completely independent variable from CHPL_MEM. Of course we could decide to make them more linked (e.g. show CHPL_JEMALLOC only when CHPL_MEM=jemalloc). See discussion with Elliot above, too.

runtime/include/mem/jemalloc/chpl-mem-impl.h
```diff
@@ -23,41 +23,60 @@
...
+// uses the prefix chpl_je_. Use the CHPL_JE_ macro to
+// paste the appropriate prefix (je_ or chpl_je_).
+#define CHPL_JE_TOK_PASTE_INTERNAL(a, b) a ## b
+#define CHPL_JE_TOK_PASTE(a, b) CHPL_JE_TOK_PASTE_INTERNAL(a,b)
+#define CHPL_JE_(NAME) CHPL_JE_TOK_PASTE(CHPL_JEMALLOC_PREFIX, NAME)
+
 
 static inline void* chpl_calloc(size_t n, size_t size) {
-  return chpl_je_calloc(n,size);
+  return CHPL_JE_(calloc)(n,size);
```
gbtitus: Good heavens, that's ugly.  Can we at least do

    #define CHPL_JE_CALLOC CHPL_JE_(calloc)`

and similarly for all the other functions, and then use `CHPL_JE_CALLOC` everywhere we want the symbol?

runtime/make/Makefile.runtime.include
```diff
@@ -103,6 +103,11 @@ UNWIND_INCLUDE=$(RUNTIME_ROOT)/make/Makefile.runtime.unwind-$(CHPL_MAKE_UNWIND)
 HWLOC_INCLUDE=$(RUNTIME_ROOT)/make/Makefile.runtime.hwloc-$(CHPL_MAKE_HWLOC)
 -include $(HWLOC_INCLUDE)
 
+# Add any further includes for JEMALLOC.
```
gbtitus: Should only be done for CHPL_MAKE_MEM=jemalloc?

gbtitus:
This looks okay in principle but I had some issues with implementation, as noted in line comments.

mppf:
@gbtitus - would appreciate your feedback on whether or not CHPL_JEMALLOC should be thought of as a separate variable from CHPL_MEM (precedent: CHPL_TASKS=qthreads and CHPL_HWLOC) or a sub-variable for the case of CHPL_MEM=jemalloc (precedent CHPL_COMM=gasnet and CHPL_COMM_SUBSTRATE).

doc/rst/usingchapel/chplenv.rst
gbtitus: Why?  Is CHPL_JEMALLOC meaningful when CHPL_MEM=cstdlib, for example?

mppf: Perhaps one day it will be. For example, we could start building the compiler with jemalloc. Doing so gives a 10% performance boost, in my measurements.

gbtitus:
I think CHPL_HWLOC with respect to CHPL_QTHREADS is a different case.  Although we started out only using the hwloc package as an adjunct to Qthreads, it quickly became our intention to use hwloc for it's own sake in the runtime, and with the topology support added last release this is now the case.  I definitely think CHPL_JEMALLOC should be treated more like CHPL_COMM_SUBSTRATE.

doc/rst/usingchapel/chplenv.rst
gbtitus: In that case we would need separate settings for compiler use of jemalloc and user program use of jemalloc.  (like host compiler and target compiler)

mppf: Yes, but both could share the "Where do I find jemalloc?" variable.

ben-albrecht:
Just want to comment on 2 design decisions that have been made:

#### `CHPL_JEMALLOC=system` vs. `CHPL_MEM=jemalloc-system`

My initial reaction was that `CHPL_JEMALLOC=system` is going to generate a ton of new env vars to manage/document as we support more system-installed third-parties. However, I think it's reasonable to expect that we'll only support 1 system-installed variant per configuration, e.g. `CHPL_MEM`, and so it shouldn't get too out of hand. There is also a benefit in developer and user code for not having to check the value of `CHPL_MEM` as `jemalloc*`.

#### public vs. internal

Given the difficulties of ensuring that a system-installed jemalloc will work for a given setup, and the limited usefulness of doing so, I would be in favor of making this feature unavailable and/or undocumented to users. It really only benefits the packaging use-case.

bradcray:
@mppf: It looks as though the `CHPL_JEMALLOC` environment variable has been hidden/undocumented more or less since it was added, but I'm wondering whether that decision still feels right to you.  At least, it's not clear to me why they wouldn't be as documented/visible as other `CHPL_` variables.

mppf:
@bradcray - IIRC, it was just from not wanting to change the `printchplenv` output without discussion / without living with the new features & the limited applicability when this was added. See also https://github.com/chapel-lang/chapel/pull/6133#issuecomment-300877045 . It would be fine with me to add `CHPL_HOST_JEMALLOC` and `CHPL_TARGET_JEMALLOC` to the `printchplenv` output, potentially only showing underneath `CHPL_HOST_MEM` / `CHPL_MEM`.

bradcray:
Great, thanks for confirming!  Tagging @jabraham17 on this in case they're not following as we were talking about it w.r.t. the CHANGES.md file (specifically, I went looking for why `CHPL_TARGET_JEMALLOC` was mentioned in the developer-only section rather than the user-facing one).

PR #6138, opened 2017-04-28
Add very prototypical support for instance declarations of generics with initializers

lydia-duncan:
This change allows variable declarations of the form `var foo: Foo(int);` to work
when Foo is a basic generic class that defines initializers.  There are still some kinks
to work out, and it needs to be extended to variable declarations of the form
`var foo: Foo(int) = new Foo(4);` as well as to support generic records, but it is a
good step in the right direction.

What it does handle:
- declaring an instance with an argument to the type specifier per generic argument
  to an initializer
  `proc init(type t)` will allow `: Foo(<type>)` declarations
  `proc init(param p)` will allow `: Foo(<param>)` declarations
  `proc init(val)` will allow `: Foo(<type>)` declarations
  Note that it is impossible to distinguish between case 1 and 3 if both are present,
  so for now these result in a conflict

What will be improved in later efforts:
- Response to multiple initializers generating the same type specifier
  Details: Right now, we generate a type specifier per initializer.  There are
  circumstances where it would be helpful to avoid generating multiple type
  specifiers with the same argument list: for instance, when the initializers
  are only distinguishable by a `where` clause, or if one initializer has no
  concrete arguments, or potentially if one initializer has less generic var
  arguments
- Argument lists with declared instantiations
- Using the old declaration style (argument per field) when no initializer
  supports it (you can declare a class instance of it, but not a record instance)

Details:
Modifies three compiler files in support of this change:

initializerRules - for every initializer we verify, create a type specifier with an
argument per generic argument in the initializer and which returns the
type instantiated by that initializer with the generic arguments specified.

normalize - modify normVarTypeWoutInit to distinguish when the type is
going to be a generic instantiation for a type with initializers.  If so and the
type is a class, create a local type variable to store the instantiation we will
use based on a call to the type specifier, then assign the default of that
type to the variable being declared.

functionResolution - update the error messages for ambiguous calls and
unresolved calls when the call is to my new type specifier functions, so
that they look prettier

Testing:
Resolves six futures.  Also fixed an error in a test case I had previously added,
and removed the .bad files for some tests that now generate failures to
match for a type specifier call.  I'd like to tune those messages further

Passed full std testing and I verified the output was reasonable for my
initializers directory futures.


compiler/passes/initializerRules.cpp
```diff
@@ -725,6 +727,10 @@ static void preNormalizeNonGenericInit(FnSymbol* fn) {
 
     fn->addFlag(FLAG_INLINE);
   }
+
```
noakesmichael: It seems surprising that this is in preNormalizeNonGenericInit().

compiler/passes/initializerRules.cpp
```diff
@@ -2318,3 +2328,61 @@ FnSymbol* buildClassAllocator(FnSymbol* initMethod) {
 
   return fn;
 }
+
+static void buildTypeFunction(FnSymbol* fn) {
```
noakesmichael: Perhaps choose a name that is a little more descriptive than "fn" and that helps
to remind the reader that this is constructing a type function based on an initializer?

compiler/passes/initializerRules.cpp
```diff
@@ -2318,3 +2328,61 @@ FnSymbol* buildClassAllocator(FnSymbol* initMethod) {
...
+  Symbol* _this = fn->_this;
+  AggregateType* at = toAggregateType(_this->type);
+
+  SET_LINENO(at);
```
noakesmichael: It is sometimes tricky to decide what to use for a line number.  I wonder if it would be better
to tie the type-function to the initializer that generates it rather than to the definition of the
type (which could be in a different file etc)?

compiler/passes/normalize.cpp
```diff
@@ -1952,18 +1952,66 @@ static void normVarTypeWoutInit(DefExpr* defExpr) {
     var->type = type;
 
   } else {
-    VarSymbol* typeTemp = newTemp("type_tmp");
-    DefExpr*   typeDefn = new DefExpr(typeTemp);
-    CallExpr*  initCall = new CallExpr(PRIM_INIT, typeExpr);
-    CallExpr*  initMove = new CallExpr(PRIM_MOVE, typeTemp, initCall);
+    bool handled = false;
```
noakesmichael: I think the "bool handled;" path is unwieldy.  As I understand it, handled will be true if
ts->type is a generic class or a generic record with initializers.  Why not make that a predicate
that can be invoked before dropping in to the else?

compiler/passes/initializerRules.cpp
```diff
@@ -2318,3 +2329,66 @@ FnSymbol* buildClassAllocator(FnSymbol* initMethod) {
 
   return fn;
 }
+
+static void buildTypeFunction(FnSymbol* fn) {
```
noakesmichael: I think I'm missing the whole point of buildTypeFunction().  I am surprised that it has so much AST in it
and that is pretending to "call" things.  I thought this type function would be a faux-hook that normalize
would use to break in to the C++ compiler and find/create the appropriate AggregateType for the variable?

Seems I've misunderstood the direction.  I think we'll need to resolve face to face.

compiler/passes/initializerRules.cpp
lydia-duncan: I may not have read that section as carefully as I should have, whoops!

compiler/passes/initializerRules.cpp
lydia-duncan: I ended up doing this later, I think the original line came about from looking at the creation of the _new function

compiler/passes/normalize.cpp
lydia-duncan: This is fair - I was planning on taking the outer if statements and condensing them into a single function call at a later stage, but was in prototyping mode so took the quick path.

compiler/passes/initializerRules.cpp
lydia-duncan: Let's talk when you get in

lydia-duncan:
Offline, Mike raised two things that I hadn't considered:

1. Concrete arguments should impact the type function that is generated (by adding arguments and perhaps where clause restrictions)
2. What if the call to the initializer that I've placed within the generated type function has side effects?

The first is resolvable.  The second might be indicative of a more foundational issue with this approach, and so I'm going to think about if there is a different strategy that will avoid this issue, or will restrict what we support for now so that we can return to this greater support at a later time.

lydia-duncan:
Upon investigation, it looks as though my approach will not result in the side-effects occurring when present (i.e. if the initializer will print something, that is not printed because of the call to the initializer within the type function I have written).  That is reassuring and means that if I can extend the support to include concrete arguments and the cases I listed in the original message, this is a viable strategy

lydia-duncan:
Though it does mean that should type functions start yielding side effects, this implementation will be impacted, and so we will need to document that consideration

lydia-duncan:
Discussion with Brad led to the excellent point that we don't have a way for the user to opt out of this behavior in the case where it causes problems for them, so close for now.  We can revisit this if there is a strong desire for the feature when we have more experience in the initializers world

PR #6152, opened 2017-05-01
Update PRK Transpose

e-kayrakli:
This PR updates the PRK Transpose study

  - Bumps up the version number
  - Changes error printouts from writeln/exit to halt
  - Drops types
  - Adds support for distributed data (without any optimization)
  - Adjusts input initialization and minor changes in the logic
    This minor changes is due to the changes in the main PRK repo. It
    increases the computation slightly, so we can expect some
    performance degradation.
  - Adjusts validation accordingly

Change is limited to the test file. And it has been tested on standard
linux64 with `CHPL_COMM=none` and `CHPL_COMM=gasnet`

General PRK notes
- I will be creating other PRs to contribute other PRKs I have
- After we are done adding new PRKs, I think another cleanup PR is
  necessary for the following reasons:
  - README.md must be updated or removed completely
  - Folder names should be adjusted(capitalization) to align with PRK
    names from the main repo. (Not a must but why not?)
  - Coding style adjustments/alignments. I am trying to keep the diff
    minimal for this PR.
  - IMHO the `validate` flag needs to be changed. (1) The name
    doesn't reflect its function precisely, (2) Its function seems to be
    suppressing output so that start_test can compare against the good
    file, which can (I think should) be handled with prediff.

### DIFF
--- a/test/studies/prk/transpose/transpose.chpl
+++ b/test/studies/prk/transpose/transpose.chpl
@@ -1,5 +1,5 @@
 //
-// Chapel's serial implementation of transpose
+// Chapel's parallel implementation of transpose
 //
 use Time;
 use BlockDist;
--- a/test/studies/prk/transpose/transpose.chpl
+++ b/test/studies/prk/transpose/transpose.chpl
@@ -2,37 +2,51 @@
 // Chapel's serial implementation of transpose
 //
 use Time;
+use BlockDist;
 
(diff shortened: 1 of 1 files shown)


ben-albrecht:
> General PRK notes
> ...

See [response](https://github.com/chapel-lang/chapel/pull/6153#issuecomment-298787440) in #6153

PR #6153, opened 2017-05-01
Update PRK Stencil

e-kayrakli:
This PR updates the PRK Stencil study

  - Bumps up the version number
  - Adds correct initialization for compact stencil
  - param-ifies the compact stencil loop. Before this, star stencil
    loops were getting unrolled whereas compact ones were not. This is
    probably due to avoid creating a very large AST/generated code when
    the radius is large. However, it leads to an unnecessary bias
    against compact stencil code. I think this should be addressed one
    way or another.

    In the OpenMP version this unrolling behaviour is controlled by a
    pound-defined constant, where unrolled loop is created with a script
    (thanks to param loop we don't need to do that). I think we should
    adapt a similar approach. Only problem would be increased code
    redundancy where identical loops would need to be written with and
    without `param`. This can be reduced to some extent by introducing 
    an inlined/nested function to take care of the arithmetic in the loop
    body and fusing some star stencil loops.

Change is limited to the test file. And it has been tested on standard
linux64 with `CHPL_COMM=none` and `CHPL_COMM=gasnet`

General PRK notes
- I will be creating other PRs to contribute other PRKs I have
- After we are done adding new PRKs, I think another cleanup PR is
  necessary for the following reasons:
  - README.md must be updated or removed completely
  - Folder names should be adjusted(capitalization) to align with PRK
    names from the main repo. (Not a must but why not?)
  - Coding style adjustments/alignments. I am trying to keep the diff
    minimal for this PR.
  - IMHO the `validate` flag needs to be changed. (1) The name
    doesn't reflect its function precisely, (2) Its function seems to be
    suppressing output so that start_test can compare against the good
    file, which can (I think should) be handled with prediff.

### DIFF
--- a/test/studies/prk/stencil/stencil.chpl
+++ b/test/studies/prk/stencil/stencil.chpl
@@ -11,7 +11,7 @@ use VisualDebug;
 use StencilDist;
 
 /* Version kept in sync with PRK repository */
-param PRKVERSION = "2.16";
+param PRKVERSION = "2.17";
 
 /* Numerical type for matrix elements */
 config type dtype = real;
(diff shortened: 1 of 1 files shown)


ben-albrecht:
I double checked the performance difference for the compact `param` loops, using the new compact intitialization for both cases:

Configuration: UGNI+qthreads
compopts: `--fast --set order=32000 --set iterations=3 --set compact=true`
execopts: `-nl 16`

| distribution | param-loop (MFlops/s)   |  no-param-loop   (MFlops/s) |
|-------------|:-------------:|:-----------------|
| blockdist:      |  17851.6   |  21319.8  |
| stencildist:    |  23778.9  | 19480.3  |

I observed no performance difference for `COMM=none`.

Given that `stencildist` is closer to what the reference version is doing, I'd argue that the `blockdist` regression is acceptable. However, it is a bit puzzling to me why they'd diverge in performance.

ben-albrecht:
> README.md must be updated or removed completely

Agreed. I like the pointer to the original repository, but the `## Implementations` section can be removed. That documentation can/should live in the comments of the Chapel programs.

> Folder names should be adjusted(capitalization) to align with PRK
names from the main repo. (Not a must but why not?)

👍 

> Coding style adjustments/alignments. I am trying to keep the diff
minimal for this PR.

Sounds good.

> IMHO the validate flag needs to be changed. (1) The name
doesn't reflect its function precisely, (2) Its function seems to be
suppressing output so that start_test can compare against the good
file, which can (I think should) be handled with prediff.

I'd be in favor of renaming it `--correctness` (a pattern employed elsewhere in the test system)

PR #6154, opened 2017-05-02
Flag some compiler temps as consts

vasslitvinov:
... and introduce newTempConst()

`newTempConst` is the same as `newTemp` except it also flags the new temp with `FLAG_CONST`.

This should allow more compiler optimizations such as remote value forwarding, resulting in simpler generated code. I am encouraging wider use of newConstTemp().

I chose the name `newTempConst` instead of `newConstTemp` so that we can grep for "newTemp" and get both const and non-const varieties.

While there, I removed one "addFlag(FLAG_TEMP)" as it is already added on all newTemps.

With this change, this test:

```
distributions/robust/arithmetic/basics/test_distributed_array_of_distributed_array1.chpl
```

when compiled with `-sdistType=DistType.replicated ` no longer triggers a bug in updateLoopBodyClasses().

Testing: linux64
An extended version of this PR also essentially passed valgrind testing.

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -997,7 +997,7 @@ static void buildLeaderIteratorFn(FnSymbol* fn, const char* iteratorName,
   lifn->where = new BlockStmt(new CallExpr("==", lifnTag, tag->copy()));
   fn->insertAtHead(new DefExpr(lifn));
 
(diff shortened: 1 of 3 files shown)


vasslitvinov:
@benharsh or @mppf - would you review?

PR #6155, opened 2017-05-02
Suppress I-fail-without-re2 test if using PGI

awallace-cray:
##  Suppress I-fail-without-re2 test if using PGI
    
    Current version of re2 requires C++11 atomic support, which PGI
    does not have. So, with PGI, re2 does not get built and the
    I-fail-without-re2 test fails.
    
    This change adds I-fail-without-re2.suppressif to quiet this error.
    
    PGI is only used in Cray Whitebox tests, so those are the only tests
    that should be affected by this change.

  ##  Parameterize CHPL_NIGHTLY_LOGDIR for whitebox jobs.
    
    Makes common-whitebox.bash honor the existing env variable, if any,
    instead of always forcing it to /data/sea/chapel/Nightly.
    
    This allows casual whitebox tests to be run using "nightly", without
    disturbing the nightly test logs or sending misleading email
    notifications.

### DIFF
--- a/test/regexp/elliot/I-fail-without-re2.suppressif
+++ b/test/regexp/elliot/I-fail-without-re2.suppressif
@@ -1,0 +1,3 @@
+# re2 requires C++11 atomic support, which pgi doesn't have
+CHPL_TARGET_COMPILER==cray-prgenv-pgi
+CHPL_TARGET_COMPILER==pgi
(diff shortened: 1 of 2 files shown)


awallace-cray:
@ronawho would you mind reviewing? Thanks!

ronawho:
Looks good. 

Can you use more descriptive branch names in the future?

awallace-cray:
@vasslitvinov This should quiet the recurring test failures in xc-wb.*-pgi test configs.

PR #4222, opened 2016-07-21
Denormalize pass

e-kayrakli:
This PR adds a `denormalize` pass right before codegen. Purpose of the pass is
to remove local variables which are def'd and use'd only once.

DETAILS

- A local variable is considered for denormalization if:
  - It has only one def and use.
  - Its def satisfies the following
    - It's a PRIM_MOVE or PRIM_ASSIGN,
    - PRIM_MOVE/PRIM_ASSIGN cannot create a communication, ie
      - Neither LHS nor RHS is wide.
      - RHS is not a MEMBER primitive with wide base.
    - RHS and LHS of PRIM_MOVE are of same type
    - Neither RHS nor LHS is extern.
  - Its use satisfies the following
    - It's not: PRIM_{ADDR_OF, ARRAY_GET, GET_MEMBER, DEREF, GET_MEMBER_VALUE}
    - If it's a PRIM_MOVE, LHS and RHS are of same type.
    - It is not condStmt in a while family loop or condBlock/incrBlock in
      CForLoop

- Given we found candidates, denormalization of the temporary is aborted if
  there is an unsafe CallExpr between def and use

- --[no-]denormalize flag and CHPL_DENORMALIZE controls the behaviour. Default
  is --no-denormalize. You have to pass -denormalize or set CHPL_DENORMALIZE=1
  to enable denormalization.

- Minor change in dead code elimination pass.

  Previously codegen pass was ignoring some of the primitives unless they are
  RHS of PRIM_MOVE. Same assumption no longer holds. From this category of
  primitives, dead code elimination didn't eliminate PRIM_ARRAY_GET for some
  reason, generating bunch of unused statement warnings.

`CallExpr` SAFETY

A CallExpr is considered safe for reordering if:

- It is a non-essential primitive,
- If it's not a primitive:
  - It doesn't have any ref formals,
  - It doesn't touch(neither def nor use) a global
  - It doesn't touch(neither def nor use) an extern
  - Function body doesn't have any essential primitive

CODEGEN CHANGES

This PR also makes important changes in code generation, mostly confined
to `expr.cpp`. Some codegen changes highlights:

- `codegenPrimMoveRhsIsSpecialPrimitive` was used in `codegenPrimMove`
  only to generate code with some primitives, whose generated code might
  change depending on the LHS of the PRIM_MOVE. Now that the AST can be
  denormalized, it is possible to have those "special" primitives
  elsewhere.

  Therefore, said helper function is transformed into

  ```
[...]

PR #6159, opened 2017-05-02
Fix denormalize bug on handling PRIM_UNARY_NOT

e-kayrakli:
Fixes a minor bug in denormalize that was causing different behaviour
w/o --no-denormalize.

Following snippet used to print out 32 with `--denormalize` and 8 with
`--no-denormalize`:

```
var i = 0:uint(8);
writeln(popcount(~i));
```

This issue is caused by integer promotion for "smaller" ints. Here i is
expanded to 32 bits, negated and popcount'ed. Without casting the
variable or assigning it to a uint(8) there is nothing truncating
the leading junk bits.

Similarly
```
uint8_t i = 0;
printf("%d\n", __builtin_popcount(~i));
```
Also prints out 32.

Similar issue also occurs with arithmetic and shift operations, and
denormalize already handles those. This PR adds PRIM_UNARY_NOT to
the list of primitives that needs a typecast when they are moved.

For reference this is the note from the initial denormalize PR #4222:
> We now generate typecasts for all integer operations. This is due to C
> promoting small integers to full ints in arithmetic operations. When
> such integers are in arithmetic operations and moved to somewhere
> other than a variable assignment to a corresponding type, we have to
> make sure they are typecast(and truncated)
> 
> Currently there is no easy way of getting c_int size for target
> architecture in the compiler. Therefore casts are not limited to small
> ints.

Tests -- all in standard linux64:
- [x] test/users/engin/bitwiseDenorm.chpl
- [x] test/release/hello*chpl
- [x] full correctness


benharsh:
Are you running the full correctness suite yourself?

e-kayrakli:
Ah, no I can't do that. Sorry for not being clear.

benharsh:
No problem, I'll fire off a paratest.

benharsh:
Testing came back clean. OK if I merge?

e-kayrakli:
Sure. Thanks!

PR #6163, opened 2017-05-03
Add a script to check for calls to the system allocator in the runtime

ronawho:
Generally speaking, we want our runtime to call the chapel allocator, instead
of the system one. This can offer better performance, and more importantly for
cases where the heap is registered, we usually want allocations in the runtime
to be registered as well.

Historically, we've enforced this with 'chpl-mem-warning-macros.h', which
does "#define malloc  dont_use_malloc_use_chpl_mem_allocMany_instead". This
approach has a couple unfortunate downsides:
 - This prevents the token "malloc" from being used, not just calls the
   function so we have to turn off the warnings in more places than ideal
 - It's caused portability issues (especially on cygwin). We define malloc
   before including system headers in many files, and so if a system header
   uses the token malloc, we fail to build. e.g.

This adds a script that effectively greps for calls to the system allocator in
the runtime. Instead of using grep, it uses cscope which will do a fuzzy parse
(parse the text, but is ok if includes aren't fully satisfied.) My hope is to add
a smoke test that runs this script, and work towards retiring the macro-based
approach. I also plan on checking the qthreads-src in our repo, and pointing
the qthreads team to this script


ronawho:
This script is a little more generalized than it probably needs to be for us, but I think it'll be useful for the qthreads team to ensure their external allocator support doesn't break over time: https://github.com/Qthreads/qthreads/issues/40.

If we end up being happy with this, I'll pass it along to the qthreads team

ronawho:
Here's the current output from the script:

https://gist.github.com/ronawho/f47450c3c20637ffaa24d8cf33a43c6c

ben-albrecht:
Can you please run `chplspell` with this file.

Issue #6162, opened 2017-05-03
Improve PRKs


ben-albrecht:
Here is a meta-issue to track progress on the implementations of Intel's [Parallel Research Kernels](https://github.com/ParRes/Kernels) in Chapel.

## Resources

-  [test/studies/prk](https://github.com/chapel-lang/chapel/tree/master/test/studies/prk)
- [Single locale performance](http://chapel.sourceforge.net/perf/chapcs/?suite=prkbenchmarks)
- [Multilocale performance](http://chapel.sourceforge.net/perf/16-node-xc/?suite=prkperf)

## General

- [ ] Clean up and improve maintainability of prk README
- [ ] Update directory names to reflect names used in PRKs, e.g. casing
- [ ] Add `contributed by` header comments, giving credit to authors and contributors.

### Stencil

- [ ] Consider dynamic unrolling approach used in OpenMP version as described in #6153

### Transpose

- [ ] Rewrite distributed implementation to reflect reference version
    - Current implementation is naive blockDist
- [ ] Enable multilocale performance testing

### Synch_p2p

- Current implementation does not reflect reference version

e-kayrakli:
I have been working on Transpose recently and wanted to capture what is missing in the current implementation:

PRK specifications and the reference MPI1 implementation uses column-major arrays for both matrices and uses column-wise data decomposition. Then, the output array is accessed in column-major order where the input is accessed in row-major order. Current Transpose implementation in Chapel do things rather haphazardly in this context. Given that there is no native column-major layout in Chapel (yet?), I think arrays can be distributed with row-major decomposition and the access orders can be reversed (row-major on output array) to emulate something close to the reference implementation and the specs.

e-kayrakli:
@ben-albrecht, looking at the issue again I think there are few things that can be added:

- Missing PRKs for completeness (some may be more important then others, like AMR):
  - AMR: A variation of Stencil that spawns subgrids to emulate adaptive mesh refinement
  - Branch: Very simple one that tests branch performance
  - Random: Another simple one
  - Reduce: At least a straightforward implementation should be simple. ("Reduce" may be a misnomer  as it seemingly does a element-wise vector addition where vectors are at specific parts of the memory.)
[...]

PR #6165, opened 2017-05-03
Add new PRKs

e-kayrakli:
Adds four new PRKs(the Parellel Research Kernels): DGEMM, Nstream,
PIC and Sparse

Issue #6162 has overall notes for improving PRKs. Some applies to the
PRKs in this PR, as well. Also, #6152 and #6153 updated the existing
Transpose and Stencil implementations recently.

Open questions / random notes:
- All but PIC are naively distributed. (There is no optimization
  whatsover) Should we keep it that way? Or not create distributed
  versions without properly optimizing them? PIC is not distributed at
  all.
- PIC code style needs a revision. I translated it from OpenMP version,
  trying to do things the Chapel way as much as I can. But variable
  names are almost identical to OpenMP variables. I think that's not the
  convention that Chapel community is adopting, so those should be
  changed to camel case where appropriate. (Possibly in a cleanup PR)
- PIC uses random_draw.c from the PRK repo almost as-is. I needed to
  make small adjustments so that variables are always declared in the
  beginning of blocks. It would be cool to have it implemented in
  Chapel, as well.  But I'd say it is very low priority at this point.
- DGEMM: I used unbounded ranges in deeply nested loops. I wonder if it
  has any performance problems compared to bounded ranges?
- DGEMM: In the CHIUW paper this version performed ~60% of OpenMP. But
  using C arrays for tile arrays makes Chapel performance almost
  identical to OpenMP. I wonder if that implementation of DGEMM would be
  frowned upon.
- On a similar note, should we have different flavors of PRKs? Such as
  coforall based Nstream.

Testing:
This PR only adds new tests to studies/prk. Said folder is locally
tested with standard linux64 and standard GASNet configs

Todo:
- [x] Cosmetic changes/more Chapel-ification in PIC
- [x] Investigate a TODO comment regarding unexpected behavior in PIC

### DIFF
--- a/test/studies/prk/DGEMM/dgemm.chpl
+++ b/test/studies/prk/DGEMM/dgemm.chpl
@@ -13,7 +13,8 @@ config const order = 10,
              iterations = 100,
              blockSize = 0,
              debug = false,
-             validate = true;
+             validate = true,
+             correctness = false; // being run in start_test
 
 
 // TODO current logic assumes order is divisible by blockSize. add that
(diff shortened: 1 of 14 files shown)


e-kayrakli:
I think PIC is in a better shape now. I still didn't update the variable names (they are mostly the same with OpenMP version). It might help the review process. I don't mind changing them in this PR, though.

ben-albrecht:
From a correctness point of view, the code for these submissions looks good to me. I've tested to confirm they pass locally as well.

Performance-oriented reviews will be done post-merge by the @chapel-lang/perf-team.

Thanks for the contribution @e-kayrakli!

PR #6107, opened 2017-04-25
Change an internal error message to a user-facing error message

lydia-duncan:
When --library is thrown while compiling a program with a main(), we used to
generate an internal error message.  However, it is really just a user error to
define a main() function when compiling in --library mode, so make this internal
error a user-facing error.

This resolves #6106 

Adds a test to lock in the error message.

Passed full std/ testing

Issue #6106, opened 2017-04-25
Internal compiler error with --library and main


TartanLlama:
### Summary of Problem

The compiler ICEd when I gave it code with `main` in a module and compiled with `--library`

### Steps to Reproduce

**Source Code:**

```chapel
module Hello {
    proc main() {
    }
}
```

**Compile command:**
`chpl --library test.chpl`

### Output

production.chpl:2: internal error: BUI0469 chpl Version 1.15.0

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.

### Configuration Information

- Output of `chpl --version`: chpl Version 1.15.0
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:

```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```

- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

lydia-duncan:
Thanks for reporting!  Turns out the internal compiler error is actually an error we would like to expose to the user, so it's an easy fix.

PR #6166, opened 2017-05-03
Change main() with --library to a warning

bradcray:
It seems to me that there's no reason not to permit main() to be defined in a .chpl program being compiled with the --library flag, it's just that the programmer shouldn't expect main() to be their program's entry point.  For example, doing this could be one way of creating a library that can self-test itself when compiled in non--library mode.

To that end, I changed the user fatal error that Lydia introduced in #6107 into a user warning and changed the wording slightly to reflect that.  In that PR, she mentioned her attempt to remove main() from the IR, but I don't think this is necessary since we rename main() anyway and generate our own calls to it when compiling in non--library mode.  Leaving it in didn't break anything.

I added new tests to (a) see that I could compile a library code containing main(), (b) see that the warning is generated when compiling it as a library, (c) link to the library and call it from a C program without having main() be called.  I also updated Lydia's existing test to ensure that it keeps working but did not remove it because it's strictly simpler than the ones I've added here.  As far as I can tell, this is the first test that locks in calling a Chapel library from C, and only the second that uses the --library flag (scary, scary).

While here, I did a cleanup to our Makefile used by the --library flag to pass '-c' to 'ar'.  Without this, we would get a message to the terminal stating where the .a file being created by the compiler was saved -- in a /tmp directory that the user didn't care about using a temporary name that the user shouldn't care about because right after that we're going to copy it back to their output directory.  This was noisy and not helpful, so squashing it seems the right thing to do.


bradcray:
@lydia-duncan / @daviditen : Would one or both of you guys like to review this?

bradcray:
Oh yeah, I forgot to propagate that to the top-level commit message (just added now).  It squashes the message from 'ar' telling where it deposits the resulting .a file.

bradcray:
Noting belatedly that this PR, like PR #6166, was motivated by issue #6106.

PR #6167, opened 2017-05-04
Do not create .error files in computePerfStats

vasslitvinov:
Historically, computePerfStats has created a .error file
next to the .dat being updated.

In an email discussion on the internal email list today and yesterday
we agreed that .error files are useless. Even if something does go wrong,
they don't currently log any valuable info. Also, whatever they do log
gets wiped out by the final open(errors_file, "w").

This change removes all occurrences of errors_file from computePerfStats.
We should remove the old .error files after this change goes in.

### DIFF
--- a/util/test/computePerfStats
+++ b/util/test/computePerfStats
@@ -19,7 +19,6 @@ import string
 test_output = None
 test_output_raw = None
 data_file = None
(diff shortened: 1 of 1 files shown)


ronawho:
👍

PR #6171, opened 2017-05-04
Extend c2chapel with our own fake defines/typedefs

benharsh:
Users can add to the new utils/custom.h file to resolve parsing issues. Also removes fake limits.h header so that we rely on the system header.


benharsh:
@ben-albrecht : care to review? I'll be opening another PR with changes related to Chapel generation.

PR #6172, opened 2017-05-05
Make replicated distributions less distinct

bradcray:
This change makes ReplicatedDist behave a lot less "specially" relative to other distributions, primarily by removing operations that implicitly acted on all replicands such as loops and I/O.  In doing so, it resolves test failures that have been occurring since array views went in (PR #5338) which seemed best addressed by "fixing" replicated distributions rather than continuing to wrestle with tests in the historical implementation of ReplicatedDist.

In the new world, most operations will only occur on the local locale's replicand.  The operations that still affect all replicands are:

* creation and destruction of domains and arrays: updates all copies
* setting a domain's indices: updates all copies

So cases that have changed are:

* iterators used to iterate over all indices/elements in all replicands, but now only iterate over the local indices/elements
* numIndices/numElements used to return the number of indices/elements multiplied by the number of replicands but now just returns the number of indices/elements in a single replicand
* printing domains used to print the index set followed by the list of locales that own a replicand; now it just prints the index set
* printing arrays used to print all the replicands, but now just prints the local replicand.

Here are the changes, sorted by file:

* in modules/dists/ReplicatedDist.chpl:
  - implemented the changes as described above
  - updated the documentation
  - introduced helper routines to get the local replicand or generate an error if there isn't one
* in modules/standard/UtilReplicatedVar.chpl:
  - updated example code to avoid assumption that printing would print all copies
* updated test/release/examples/primers/distributions.chpl to reflect the new semantics and generally refresh the descriptions (also added section headers for the rendered form of the file)
* in test/distributions/vass/testReplicated* and test/parallel/forall/vass/ri-4-arrdom.chpl:
  - updated tests to maintain correctness under new semantics
  - replaced wrapper* versions of tests to use comm-none.good and .numlocales instead
* significantly simplified test/distributions/robustness/arithmetic by:
  - removing *.replicated.good files for cases where it no longer differs
  - moving *.default.good files to *.good files and removing PREDIFFs when possible
  - removing CLEANFILES and .gitignore entries for such cases
* updated comm counts in test/distributions/robustness/performance/multilocale
  - some of these changes were primarily due to the array views change which changed the representation of the target locales collection into an associative domain of locales and reflect comm count differences that we've been seeing in nightly testing for some time now.
  - others represent cases where we've improved things by only acting on local copies of the domain/array rather than on all locales' copies as we used to.
  - I'm also realizing very belatedly that the comm counts for the alloc*.chpl performance tests are unstable from run to run so these .good updates will not cause them to start passing.

TODO:
- [x] pass gasnet testing at least once
- [x] get buy-in from team
- [x] code cleanup
- [x] linux64 testing
- [x] gasnet testing
- [ ] add test of "no local replicand" style errors
- [x] why does testReplWrites get a valgrind error on 1 locale?
- [x] clean up printReplicands()
- [x] are we ok with slips in comm counts?
- [x] update documentation
- [x] mpi module OK?
- [ ] hdfs module OK?


mppf:
I looked at the changes to the module code, which looks good to me. I didn't look at the test changes.

bradcray:
@vasslitvinov : Do you have time to give this PR a review as the original author of ReplicatedDist.chpl?  Thanks.

modules/dists/ReplicatedDist.chpl
```diff
@@ -62,79 +59,29 @@ That is, changes to one replicand of an array are never propagated to
...
-
-Note that the above behavior may change in the future. In particular,
-we are considering changing it so that replication is never observable.
-For example, only the local replicand would be accessed in all cases.
+When accessing a replicated domain or array from a locale *not* in the
+set of target locales, an error is reported if bounds-checking is on;
+undefined behavior occurs if it is off.
 
+Otherwise, only the replicand on the current locale is accessed by
+default when performing operations on the domain or array.
```
vasslitvinov: Is "by default" meaningful here? If yes, as a user I'd like to know how to change from the default. If no, let's take that out.

modules/dists/ReplicatedDist.chpl
```diff
@@ -410,31 +366,19 @@ proc ReplicatedDom.dsiGetIndices(): rank * range(idxType,
...
+    yield i;
 }
 
 iter ReplicatedDom.these(param tag: iterKind) where tag == iterKind.leader {
-  coforall locDom in localDoms do
-    on locDom do
-      // there, for simplicity, redirect to DefaultRectangular's leader
-      for follow in locDom.domLocalRep._value.these(tag) do
-        yield follow;
+  // for simplicity, redirect to DefaultRectangular's leader
```
vasslitvinov: "for simplicity" implies that we have other choices, which practically speaking we don't.
Let's remove "for simplicity".

test/release/examples/primers/distributions.chpl
```diff
@@ -240,86 +240,88 @@ verifyID(BA);
 // is replicated onto each locale. (Note: consistency among these
 // array replicands is NOT maintained automatically.)
 //
-// This replication is observable in some cases but not others,
-// as shown below. Note: this behavior may change in the future.
+// This replication is observable in some cases but not others, as
+// shown below.  In places this example uses a utility function,
```
vasslitvinov: We should re-jigger everything in this primer that talks about Replicated. Because really I don't see how the user can observe replication, other than by comparing values on distinct locales. A TODO for the doc-freeze week?

vasslitvinov:
Looks good modulo my feedback.

test/release/examples/primers/distributions.chpl
bradcray: I didn't realize that I'd left this in such a mess.  I think I started to update it and then must've been distracted before finishing (?).  I'm going to refresh it now -- no need to put it off.

bradcray:
@vasslitvinov : I just pushed a refresh of the replicated distribution primer.  Would you mind taking a look to see if it's above your bar?

test/release/examples/primers/distributions.chpl
```diff
@@ -233,98 +238,124 @@ verifyID(BCA);
...
-
+// Replicated
+// ----------
+//
+// The ``ReplicatedDist`` distribution is different from the previous
+// cases: each of the original domain's indices is replicated onto
+// each locale, as are the corresponding array elements.  For example,
+// a domain ``{1..3}`` distributed using ``ReplicatedDist`` will store
+// three indices per locale that the distribution is targeting (by
+// default, all of them).  Similarly, an array declared over that
```
vasslitvinov: "all of them" --> "all locales"

test/release/examples/primers/distributions.chpl
```diff
@@ -233,98 +238,124 @@ verifyID(BCA);
...
-forall ra in RA do
-  ra = here.id;
-
-writeln("Replicated Array Index Map, ", RA.numElements, " elements total");
-writeln(RA);
-writeln();
+// Queries about the size of a replicated domain or array will return
+// the size per locale:
+//
+writeln("Replicated Array Index Map, ", RA.numElements, " elements per locale");
```
vasslitvinov: "Index Map," --> "has"

test/release/examples/primers/distributions.chpl
```diff
@@ -233,98 +238,124 @@ verifyID(BCA);
...
+proc writeReplicands(X) {
+  for loc in Locales {
+    on loc {
+      writeln(loc, ":");
+      writeln(X);
+    }
+  }
+}
+
+writeReplicands(RA);
```
vasslitvinov: Before writeReplicands(), let's insert
```
writeln("Replicated Array Index Map");
```

PR #6174, opened 2017-05-05
Make install llvm

mppf:
* Improves debugging output from --print-chpl-settings
* Adds some commented out debugging code to uti/config/compileline to help next time
* Fixes make install + --llvm

The --llvm compiles use the compileline script, so install that.  (It also might be useful in some cases when bulding C code to be integrated with Chapel).

Then, modify compileline to set CHPL_MAKE_RUNTIME_LIB, CHPL_MAKE_RUNTIME_INCL, and CHPL_MAKE_THIRD_PARTY. It was already setting CHPL_MAKE_HOME but these others need to be set too in order for the installed Make to work correctly.

Passed full local testing.

### DIFF
--- a/util/config/compileline
+++ b/util/config/compileline
@@ -28,6 +28,15 @@ if( defined $ENV{"CHPL_HOME"} ) {
 }
 
 $ENV{"CHPL_MAKE_HOME"} = $chpl_home_dir;
+if( defined $ENV{"CHPL_RUNTIME_LIB"} ) {
+  $ENV{"CHPL_MAKE_RUNTIME_LIB"} = $ENV{"CHPL_RUNTIME_LIB"}
+}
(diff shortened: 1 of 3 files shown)


mppf:
@ben-albrecht - could you review? thanks

util/config/compileline
```diff
@@ -37,6 +46,15 @@ $make = "$make --no-print-directory";
 
 sub mysystem {
   my $cmd = shift;
+
+  # Uncomment for debug output
```
ben-albrecht: Would this be used frequently enough to control by a flag rather than uncommenting the code?

mppf: I'm not sure yet... I needed it to do this. How about we turn it in to a flag if we need it again?

PR #6175, opened 2017-05-05
Dynamically Loading Generics CHIP

mppf:
Adds a CHIP that discusses strategies for dynamically loading Chapel code, especially considering generic code.

- [x] move the new chip to the next number


mppf:
@benharsh, I don't think this technically requires review, but since we were discussing this document... do you have any suggestions for adjustments to do to it before it is merged?

PR #6176, opened 2017-05-05
Add param and type versions of the Reflection.getField() functions

daviditen:
If the field to get is a param or a type, the getField() functions can
return a param or type.

Added a test to check that this is working.  Removed standard module
line numbers from .good files for two tests and added .prediff scripts to
strip them from the output.

Passed full local paratest with --verify.


mppf:
Cool, looks good.

daviditen:
Thank you!

PR #6177, opened 2017-05-05
Add AstCount, --print-module-resolution prints counts

mppf:
I thought it would be interesting to investigate how many AST nodes were involved in each module. We could use this information to trim down the amount of AST that the compiler automatically includes.

So this PR does three things:

1) It adds a new AST visitor, AstCount, that counts the various kinds of AST elements
2) It adjusts the handling of --print-module-resolution to count the number of AST elements within each module after that module is resolved and to print out the total. This is not necessarily the best place to gather this information but it was easy to add.
3) It adds a test of --print-module-resolution that is designed to not become a maintenance burden.

- [x] full local testing

### DIFF
--- a/compiler/AST/AstCount.cpp
+++ b/compiler/AST/AstCount.cpp
@@ -1,0 +1,262 @@
+/*
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


compiler/include/AstCount.h
```diff
@@ -0,0 +1,145 @@
...
+  virtual void   exitCForLoop        (CForLoop*          node);
+
+  virtual bool   enterForLoop        (ForLoop*           node);
+  virtual void   exitForLoop         (ForLoop*           node);
+
+  virtual bool   enterParamForLoop   (ParamForLoop*      node);
+  virtual void   exitParamForLoop    (ParamForLoop*      node);
+
+  virtual bool   enterCondStmt       (CondStmt*          node);
+  virtual void   exitCondStmt        (CondStmt*          node);
```
daviditen: Above this line the order of the declarations and the definitions in AstCount.cpp are the same, below this line they are not.  It would be nice to have the order match all the way.

compiler/AST/AstCount.cpp
```diff
@@ -0,0 +1,262 @@
...
+ */
+
+#include "AstCount.h"
+
+
+AstCount::AstCount() {
+#define init_members(type) num##type = 0
+foreach_ast(init_members);
+#undef init_members
+
```
daviditen: Would be nice to put a comment here like in the other places you had to handle things not covered by the macro.

PR #6181, opened 2017-05-05
Miscellaneous c2chapel improvements

benharsh:
- Map 'unsigned' type to 'c_uint'
- Map 'FILE' type to '_file'
- Avoid generating structs, arguments, and functions with identifiers that
  are also Chapel keywords
- Fix bug related to returning a c_fn_ptr
- Generate an overload for empty varargs
- Better handling for typedef'd pointer to struct

### DIFF
--- a/tools/c2chapel/c2chapel.py
+++ b/tools/c2chapel/c2chapel.py
@@ -54,6 +54,8 @@ import re
 noComments = False
 DEBUG      = False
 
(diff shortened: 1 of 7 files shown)


benharsh:
@ben-albrecht : with these changes I can generate and compile the results for sqlite3.h and ncurses.h (on my mac) without having to edit the results.

PR #6187, opened 2017-05-06
Add checkDivByZero checks for %

ben-albrecht:
Previously, modulus-zero was not handled by the Chapel compiler or modules resulting in either a backend C compiler error or a runtime exception. This PR adds in checks to provide Chapel compiler and runtime errors like we do for divide-by-zero.

Also in this PR:

- Added tests that demonstrate this fix. Tests follow pattern used in `checkDivByZero` tests
- Annotated `proc %(param a: uint(64), b: uint(64))` as a temporary solution to handling param coercion for uints
- Added a future that documents a tricky case that requires additional compiler changes (#6188)
- Cleaned trailing whitespace in `ChapelBase.chpl` (it's just habit..)

**Testing**

- [x] `start_test --compiler="" test/compflags/checks/`
- [x] paratest


modules/internal/ChapelBase.chpl
```diff
@@ -307,11 +307,31 @@ module ChapelBase {
   //
   // % on primitive types
   //
-  inline proc %(a: int(?w), b: int(w)) return __primitive("%", a, b);
-  inline proc %(a: uint(?w), b: uint(w)) return __primitive("%", a, b);
+  inline proc %(a: int(?w), b: int(w)) {
+    if (chpl_checkDivByZero) then
+      if b == 0 then
+        halt("Attempt to divide by zero");
```
bradcray: I'm not sure of the answer to this question:  Is "divide" an appropriate term here when it's modulus rather than a divide that we're computing?  If there's precedent in other systems, it's fine with me.

modules/internal/ChapelBase.chpl
```diff
@@ -307,11 +307,31 @@ module ChapelBase {
...
+    return __primitive("%", a, b);
+  }
 
-  inline proc %(param a: int(?w), param b: int(w)) param return __primitive("%", a, b);
-  inline proc %(param a: uint(?w), param b: uint(w)) param return __primitive("%", a, b);
+  inline proc %(param a: int(?w), param b: int(w)) param {
+    if (chpl_checkDivByZero) then
```
bradcray: As we discussed, there's no reason to check the chpl_checkDivByZero flag in the param case since there's no execution-time cost for this operation.  See the param-/ overload for an example.

modules/internal/ChapelBase.chpl
```diff
@@ -1286,23 +1306,29 @@ module ChapelBase {
     __primitive("/=", lhs, rhs);
   }
   inline proc /=(ref lhs:real(?w), rhs:real(w)) {
+    if (chpl_checkDivByZero) then
+      if rhs == 0.0 then
+        halt("Attempt to divide by zero");
```
bradcray: I don't believe we want divide by zero checks for reals because floating point semantics should handle it without a C-level exception.  Note that we don't support them for the plain-old '/' operator, for example.

modules/internal/ChapelBase.chpl
```diff
@@ -1286,23 +1306,29 @@ module ChapelBase {
...
     __primitive("/=", lhs, rhs);
   }
   inline proc /=(ref lhs, rhs) {
     lhs = lhs / rhs;
   }
 
   inline proc %=(ref lhs:int(?w), rhs:int(w)) {
+    if (chpl_checkDivByZero) then
+      if rhs == 0:int(w) then
```
bradcray: Did you find that the casts after these literal '0' expressions were necessary?  If not, I'd just as soon remove them to avoid the code complexity.

modules/internal/ChapelBase.chpl
```diff
@@ -1286,23 +1306,29 @@ module ChapelBase {
...
-  }
-  inline proc %=(ref lhs:real(?w), rhs:real(w)) {
+    if (chpl_checkDivByZero) then
+      if rhs == 0:uint(w) then
+        halt("Attempt to divide by zero");
     __primitive("%=", lhs, rhs);
   }
   inline proc %=(ref lhs, rhs) {
-    lhs = lhs % rhs;
+      lhs = lhs % rhs;
```
bradcray: Unless I'm misreading this diff, it looks to me like this is unusually indented in the new version.

modules/internal/ChapelBase.chpl
```diff
@@ -1485,9 +1511,7 @@ module ChapelBase {
 
   // non-param/param and param/non-param
   inline proc %(a: uint(64), param b: uint(64)) {
-    return __primitive("%", a, b);
-  }
-  inline proc %(param a: uint(64), b: uint(64)) {
+    if b == 0 then compilerError("Attempt to divide by zero");
```
bradcray: It seems to me that we should have a:

````chapel
inline proc %(a: int(64), param b: int(64)) { ... }
````

version as well for symmetry (and since ints can't coerce to uints).

File: test/compflags/checks/checkModByZero.chpl [hunk 7]
```diff
@@ -0,0 +1,62 @@
...
+  when 1 do
+    writeln(10 % 0);
+
+  //proc %(param a: uint(?w), param b: uint(w)) param
+  when 2 do
+    writeln(10 % 0);
+
+  //proc %(a: uint(64), param b: uint(64))
+  when 3 do
+    writeln(unum % (0:uint));
```
bradcray: I'd guess that the cast here isn't necessary (and the test might be slightly more interesting without it since it's what users are more likely to write).  If it is necessary, I'd be curious to know why.

modules/internal/ChapelBase.chpl
ben-albrecht: I originally thought this would be simpler, but now I've changed my mind. I think clarifying the operation would be an improvement. For reference, Python does this:

```
ZeroDivisionError: integer division or modulo by zero
```

and clang does this:

```
warning: remainder by zero is undefined [-Wdivision-by-zero]
```

modules/internal/ChapelBase.chpl
ben-albrecht: Just being overly cautious.

modules/internal/ChapelBase.chpl
ben-albrecht: just being overly cautious

modules/internal/ChapelBase.chpl
bradcray: (meanwhile, I'm doing a quick check to see if I can figure out why the / version has a param/non-param overload.  If not, I'll remove it.  If there is a good reason, we may need to do the same here (but I hope not)).

modules/internal/ChapelBase.chpl
bradcray: Since we can be precise about which operation, I'd be inclined not to follow Python's lead.  It's too bad that the wording I chose for divide-by-zero doesn't have a symmetrical wording for % (that I can see).  "attempt to modulate by zero"?!? :)

modules/internal/ChapelBase.chpl
bradcray: I'd advocate for being brave / breaking superstitions for things that it seems the language should do.  It's a good way to find real problems when it doesn't do what you think it should.

bradcray:
Thinking about this PR a bit more this morning, I think the man page text should be updated to mention modulus/o as well.  And in doing so, would you please change the existing text's use of "routines" to "operations"?

modules/internal/ChapelBase.chpl
bradcray: Unfortunately, it looks like the ``/(param a: uint(64), b: uint(64))`` overload of division is currently required for expressions like "10/myUint" due to a longstanding issue with param coercion for uints that's been in my court for a long time.  In the interest of keeping your PR simple, I wouldn't worry about adding the param/non-param overload for % since we don't have it today...  though you also could do so in order to improve symmetry between the two.

ben-albrecht: I'm going to add `%(param a: uint(64), b: uint(64))` back in for this purpose, with a TODO comment to remove once param coercion for uints is improved.

ben-albrecht:
@bradcray - made changes in response to your feedback. Going to run another paratest and wait for your approval before merging.

modules/internal/ChapelBase.chpl
```diff
@@ -1293,12 +1311,15 @@ module ChapelBase {
...
   inline proc %=(ref lhs:int(?w), rhs:int(w)) {
+    if (chpl_checkDivByZero) then
+      if rhs == 0 then
+        halt("Attempt to compute a modulus by zero");
     __primitive("%=", lhs, rhs);
   }
   inline proc %=(ref lhs:uint(?w), rhs:uint(w)) {
-    __primitive("%=", lhs, rhs);
-  }
-  inline proc %=(ref lhs:real(?w), rhs:real(w)) {
```
bradcray: Unless I'm reading the diff wrong, it appears that you've completely removed the %= overload for reals, which I don't think we want to do.

ben-albrecht: You've read the diff correctly. From the PR msg:

> Dropped `proc %=(ref lhs:real(?w), rhs:real(w))`, since `%` is not supported for real-types

Is this incorrect, or is there a separate reason to support `%=` for reals, when `%` is not supported?

ben-albrecht: For additional info: reals *are* supported in the `Math.mod()` implementation.

bradcray: Oh sorry, I didn't [re?-]read the PR message

And gosh, that's weird... I could've sworn someone asserted that % was supported on reals the other week, and I was surprised by that, but then realized that C supports fmod() and figured we must map % to fmod.

Since your PR isn't related to % on reals, I'd leave that overload in for now and then I think we should add an overload for % on reals as a separate PR (which I'm willing to drive if you're not interested).

ben-albrecht: OK - sounds good to me.

bradcray:
Looks good, thanks!

ben-albrecht:
My paratest run has revealed an enum-related bug that has surfaced with this change, demonstrated by the following snippet:


```chapel
enum E { zero, one, two }
var x = 1;
writeln(x % E.two);
```

This results in the following error:

```
$CHPL_HOME/modules/internal/ChapelBase.chpl:137: error: 'param' functions cannot return non-'param' values
enum-bug.chpl:4: error: Attempt to compute a modulus by zero
```

.. and causes failures for `test/studies/shootout/meteor/kbrady/meteor.chpl` and `test/studies/shootout/meteor/kbrady/meteor-implicit-domain.chpl`, due to the `rotate` function:

```chapel
  /* Returns the direction rotated 60 degrees clockwise */
  proc rotate(dir: direction) : direction {
    return ((dir + 2) % PIVOT): direction;
  }
```

Note that submitted version, derived from `meteor-parallel.chpl` does not fail here, because it computes the modulus by the enum size, `direction.size`, rather than the last enum value (a change made during the pre-submission clean-ups).

@bradcray - I propose doing the following:

- add a future for the snippet above, in this PR
- update these meteor implementations to compute the modulus by the enum size, in this PR
- open an issue for this enum bug

What do you think?

bradcray:
That approach sounds good to me.

bradcray:
(Or better, do the future + meteor changes in a separate PR to be merged first, for cleanliness)

ben-albrecht:
Meteor changes merged. Rebasing and retesting.

PR #6197, opened 2017-05-09
Replace all runtime calls to the system allocator with wrapper functions

ronawho:
Add sys_{malloc, calloc, realloc, memalign, free} wrappers and use them instead
of directly calling the system allocator. This allows the lookForBadAlloc
script (added in #6163) to run cleanly.

Generally speaking, we want our runtime to call the chapel allocator, instead
of the system one. This can offer better performance, and more importantly for
cases where the heap is registered, we usually want allocations in the runtime
to be registered as well. It also allows us to track memory allocations via
'chplmemtrack'. There are a few special cases where we need to call the system
allocator (memory layer not initialized yet, want non-registered memory.) Those
few cases now call the sys_ wrappers instead of directly calling the system
allocator.

Historically, we've enforced our "don't use the system allocator" policy with
'chpl-mem-warning-macros.h/chpl-mem-no-warning-macros.h/', which does something
like "#define malloc dont_use_malloc_use_chpl_mem_allocMany_instead". This
approach has a couple unfortunate downsides:
 - It prevents the token "malloc" from being used, not just calls the function
   so we have to turn off the warnings in more places than ideal
 - It causes portability issues (especially on cygwin). We define malloc
   before including system headers in many files, and so if a system header
   uses the token malloc, we fail to build. e.g.

This patch allows the ./util/devel/lookForBadAlloc script to pass. My intention
is to use that script to check that we're not mistakenly calling the system
allocator and retire chpl-mem-warning-macros.h.

Beyond changing calls to the system allocators to use the sys_ wrappers, there
are some minor cleanups, behavior changes, and bugfixes exposed by this effort:
 - fix mismatched call to system strndup and qio_free (hdfs bug)
 - fix chpl-mem-replace-malloc.c memalign call to __libc_memalign (memtrack bug)
 - remove chpltypes_malloc wrapper, which duplicated chpl_mem_allocMany logic
 - qbytes_iobuf is now system page aligned instead of heap page aligned
   - wastes less space when hugepages are used
   - may want to use system allocator for qbytes_iobuf in the future

### DIFF
--- a/runtime/include/chpl-mem-sys.h
+++ b/runtime/include/chpl-mem-sys.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _chpl_mem_sys_H_
+#define _chpl_mem_sys_H_
+
+// disable mem warnings since we need to call the system allocator
+#include "chpl-mem-no-warning-macros.h"
+
+#include <stdlib.h>
+#ifdef __GLIBC__
+#include <malloc.h>  // get memalign
+#endif
+
+static inline void* sys_calloc(size_t n, size_t size) {
+  return calloc(n, size);
+}
+
+static inline void* sys_malloc(size_t size) {
+  return malloc(size);
+}
+
+static inline void* sys_memalign(size_t boundary, size_t size) {
+#ifdef __GLIBC__
+  return memalign(boundary, size);
+#else
+  void* ret = NULL;
+  int rc;
+  rc = posix_memalign(&ret, boundary, size);
+  if( rc == 0 ) return ret;
+  else return NULL;
+#endif
+}
+
+static inline void* sys_realloc(void* ptr, size_t size) {
+  return realloc(ptr, size);
+}
+
+static inline void sys_free(void* ptr) {
+  free(ptr);
+}
+
+// Now that we've defined our functions, turn the warnings back on.
+#include "chpl-mem-warning-macros.h"
+
+#endif
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -249,14 +249,14 @@ size_t chpl_mem_localizationThreshold(void) {
 
 #else // LAUNCHER
 
-#include <stdlib.h>
+#include "chpl-mem-sys.h"
 #include "arg.h"
 
 #define chpl_mem_allocMany(number, size, description, lineno, filename)        \
-  malloc((number)*(size))
+  sys_malloc((number)*(size))
 
 #define chpl_mem_free(ptr, lineno, filename)        \
-  free(ptr)
+  sys_free(ptr)
 
 #endif // LAUNCHER
 
--- a/runtime/include/qio/qbuffer.h
+++ b/runtime/include/qio/qbuffer.h
@@ -474,6 +474,24 @@ qioerr qbuffer_memset(qbuffer_t* buf, qbuffer_iter_t start, qbuffer_iter_t end, 
 #define qio_free(ptr) chpl_mem_free(ptr, 0, 0)
 #define qio_memcpy(dest, src, num) chpl_memcpy(dest, src, num)
 
+typedef chpl_bool qio_bool;
+
+#else
+
+#include "chpl-mem-sys.h"
+
+#define qio_malloc(size) sys_malloc(size)
+#define qio_calloc(nmemb, size) sys_calloc(nmemb,size)
+#define qio_realloc(ptr, size) sys_realloc(ptr, size)
+#define qio_memalign(boundary, size) sys_memalign(boundary, size)
+#define qio_free(ptr) sys_free(ptr)
+#define qio_memcpy(dest, src, num) memcpy(dest, src, num)
+
+typedef bool qio_bool;
+
+#endif
+
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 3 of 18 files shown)


ronawho:
@gbtitus and @mppf mind reviewing? 

Probably easiest to review commit at a time. Most of the commits are simple, ones that deserve more attention are https://github.com/chapel-lang/chapel/pull/6197/commits/9e132b2daa93b38466cbd1edd92d797525b20b89, https://github.com/chapel-lang/chapel/pull/6197/commits/a2179abe4813f96c6b6d53770870e87f5d706599, https://github.com/chapel-lang/chapel/pull/6197/commits/42a46ca4e376a388c6e879f62c17ca777f4f685d, and https://github.com/chapel-lang/chapel/pull/6197/commits/b18f0d471d2cd3762edd5cbd8e7264fbbe55450e

ronawho:
Oh and tested std, quickstart, gasnet-everything, and gasnet-fast.

runtime/src/qio/qbuffer.c
```diff
@@ -148,14 +148,8 @@ qioerr _qbytes_init_iobuf(qbytes_t* ret)
 {
   void* data = NULL;
   
-  // allocate 4K-aligned (or page size aligned)
-  // multiple of 4K
-  data = qio_valloc(qbytes_iobuf_size);
+  data = qio_memalign(sys_page_size(), qbytes_iobuf_size);
```
mppf: Looks good. For 4K pages it's worth page aligning these (so certain system calls might be faster) but it's probably better to save ~2MB of space than to 2MB-align them. It's also less likely that these are communicated. If later on it turns out that it matters that these are really page aligned, we could conceivably put them on a different heap from the main one if hugepages are in use, so their page size can really be 4K.  (e.g. using the system allocator). But we don't have to do that now.

runtime/include/chpl-mem.h
```diff
@@ -249,14 +249,14 @@ size_t chpl_mem_localizationThreshold(void) {
 
 #else // LAUNCHER
 
-#include <stdlib.h>
+#include "chpl-mem-sys.h"
```
gbtitus: Just curious -- can the check-for-malloc script ignore stuff that is `#ifdef LAUNCHER`?  If so you wouldn't need the changes here.

runtime/src/chpl-mem-replace-malloc.c
```diff
@@ -231,7 +231,7 @@ int posix_memalign(void **memptr, size_t alignment, size_t size)
     *memptr = NULL;
     ret = chpl_posix_memalign_check_valid(alignment);
     if( ret ) return ret;
-    *memptr = memalign(alignment, size);
+    *memptr = __libc_memalign(alignment, size);
```
mppf: Looks like it would have called memalign in this file and that would have checked chpl_mem_inited() and then run __libc_memalign. But the tracking would have accounted for it twice, which is a bug I think.

Looks like a good change to me. Thanks!

runtime/include/qio/qbuffer.h
```diff
@@ -500,13 +485,28 @@ typedef chpl_bool qio_bool;
...
+{
+  char* ret = (char*) qio_malloc(strlen(ptr)+1);
+  if( ret ) strcpy(ret, ptr);
+  return ret;
+}
+
+#ifdef __cplusplus
+} // end extern "C"
+#endif
+
```
mppf: This is fine with me but looks really familiar... I thought I made this change somewhere. Anyway it can be in this PR.

mppf: While we're at it, it'd probably be faster to save the strlen into a len variable and then use memcpy. You don't have to do that now if you don't want to, but I'd appreciate a TODO comment.

runtime/src/qio/auxFilesys/hdfs/qio_plugin_hdfs.c
```diff
@@ -493,10 +493,15 @@ const qio_file_functions_ptr_t hdfs_function_struct_ptr = &hdfs_function_struct;
...
+  char* ret = NULL;
   const char *keys = "."; // We lop off the .us.cray...
-  int i = strcspn (loc, keys);
-
-  return strndup(loc, i);
+  int len = strcspn (loc, keys);
+  ret = (char*) qio_malloc(len+1);
+  if (ret) {
+    ret[len] = '\0';
+    qio_memcpy(ret, loc, len);
```
mppf: Looks good. I'd tend to write the '\0' after the memcpy (easier to optimize), but that's a minor issue & performance is not critical here.

runtime/src/qio/qbuffer.c
mppf: Put another way, I think we should consider making it sys_memalign. I think that was my original intention.

mppf: Note that in some use cases, these channel buffers are just the result of running 'mmap' so they won't necessarily be in the Chapel heap either way.

runtime/include/chpl-mem-sys.h
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _chpl_mem_sys_H_
+#define _chpl_mem_sys_H_
+
+// disable mem warnings since we need to call the system allocator
+#include "chpl-mem-no-warning-macros.h"
```
gbtitus: If this is now the only remaining instance of including this file I'd be in favor of inlining its contents here and getting rid of it.  (And same for `#include "chpl-mem-warning-macros.h"` below.)

gbtitus:
Looks good -- I had a couple of line comments but it's up to you whether to do anything about/with them.

runtime/include/chpl-mem-sys.h
mppf: Since Elliot is planning to get rid of these warning macros files in a subsequent PR I think it makes more sense to leave them as-is for now. That way he can get more glory out of the next PR :)

runtime/include/chpl-mem.h
ronawho: The check-for-malloc script is a glorified grep (does a fuzzy parse of the source code using cscope) so there's no simple way of doing this, and I personally like switching the launcher case over to sys_{malloc, free}

runtime/include/qio/qbuffer.h
ronawho: I'm happy to change it now

runtime/src/qio/auxFilesys/hdfs/qio_plugin_hdfs.c
ronawho: I'm happy to change, and I slightly prefer the ordering you suggest (easy to read/follow IMO). I went with the current approach because I was looking at the glibc strndup and figured they probably have a reason for putting '\0' first: https://code.woboq.org/userspace/glibc/string/strndup.c.html#48

runtime/include/chpl-mem-sys.h
ronawho: You caught me :)

Yeah, saving this for a future PR. There are still other uses of chpl-mem-no-warning-macros in places where we use the token `malloc` (jemalloc shim) and places where we need to disable it because cygwin system includes reference the token.

I _could_ remove them now, but I'd prefer to merge this, and have the jenkins job that runs ./util/devel/lookForBadAlloc start running normally (mail everybody on failure) and then get rid of the macros so there's no "lapse" in checking

runtime/src/qio/qbuffer.c
ronawho: Just double-checking: you're not asking me to convert it to sys_memalign in this PR, right? Just something to think about for the future?

mppf: Up to you. I think we should convert it to sys_memalign but that is more of a behavior change than you probably want to do in this PR.

ronawho: I'll leave that to you and/or another PR.

runtime/include/qio/qbuffer.h
ronawho: done

runtime/src/qio/auxFilesys/hdfs/qio_plugin_hdfs.c
ronawho: moved

ronawho:
Pushed some minor changes brought up in the review. Anything else you want done prior to merging?

gbtitus:
Ship it!

mppf:
All good by me

Issue #6184, opened 2017-05-05
Hoist chpl__getPrivatizedCopy


benharsh:
Consider the following program:

```chpl
use BlockDist;

proc main() {
  var dom = {1..10};
  var space = dom dmapped Block(dom);
  var A : [space] int;

  for i in 1..10 {
    A[i] += 1;
  }

  writeln(A);
}
```

When compiled with --no-local, the ``A[i]`` expression will generate a call to the runtime function chpl__getPrivatizedCopy. We currently do not hoist this call, and it surprisingly turns out to have an impact on the Stencil PRK. In the Stencil PRK, there is more than one call to chpl__getPrivatizedCopy.

On 16-nodes we observe a 15% improvement in performance when hand-modifying the code to manually hoist the chpl__getPrivatizedCopy call.

|               | MFlop/s |
|---------|---------|
| Nightly | 81860.5 |
| HandOpt | 93988.9 |

The performance improvement varies depending on problem size and other potential hand-optimizations.

ronawho:
@benharsh I had a pretty easy time getting chpl__getPrivatizedCopy hoisted for something like:

```chapel
use PrivatizationWrappers;

proc myproc()  {
  var privatizedIdx = 5;
  var newValue = new C(privatizedIdx);
  insertPrivatized(newValue, privatizedIdx);

  for 1..10 {
    var a = getPrivatized(privatizedIdx);
  }
}

myproc();
```

but for the code you gave me it's going to be a lot trickier to do. The privatized class ends up getting passed around to a bunch of different functions and is used in different branches, so I think the defUse analysis and the domination analysis are preventing hoisting. If you have time next week, we should sit down and go through the generated code together.

An easier workaround might be to just move the runtime support for privatization into a header so the backend can inline and optimize it. Here's a first stab at that: https://github.com/chapel-lang/chapel/compare/master...ronawho:inline-privatization

It's not getting quite the performance boost you saw, but it's close at ~90-92 MFlop/s:

```
==> stencil-blockdist.dat <==
# master
05/05/17 	79833.6	0.243646
05/05/17 	79927.7	0.243359
05/05/17 	79953.6	0.24328

# inline privatization
05/05/17 	90730.2	0.214384
05/05/17 	89620.7	0.217038
05/05/17 	90291.2	0.215427

==> stencil-stencildist.dat <==
# master
05/05/17 	81120.2	0.239782
05/05/17 	80971.4	0.240222
[...]

PR #6198, opened 2017-05-09
Move runtime privatization support into chpl-privatization.h

ronawho:
Move runtime privatization support from the .c file to the header.
chpl_getPrivatizedClass() can be called frequently, so we want to allow the
backend compiler to fully optimize/inline calls to it.

Moving the privatization source code into the header has a pretty big
performance impact for the stencil PRK, improving performance by about 15% for
16-node-xc. There's also some minor improvements for fft, and lulesh.

This is motivated by https://github.com/chapel-lang/chapel/issues/6184, though it's not quite enough to close that issue yet.


ronawho:
@benharsh mind reviewing? Will do full gasnet-{everything, fast} testing tomorrow.

PR #6201, opened 2017-05-09
"make check" saves stdout/stderr from chpl compilation in memory, not tmp file

awallace-cray:
"make check" saves stdout/stderr from chpl compilation in memory, not tmp file
    
    A "disk full" on $HOME file system should not prevent us from seeing the "disk full" error messages.


ronawho:
It's been a long time since I've looked at this file. Changes look fine to me, but @ben-albrecht is probably a better reviewer

PR #6202, opened 2017-05-09
Delete the root locale upon program termination

bradcray:
It was pointed out last week that we don't ever clean up the locale
classes, which prevents locale model authors from making clean-up
steps during program teardown.  This seems like an oversight, and
one that looks easy to correct.

TODO:
- [x] valgrind testing
- [x] gasnet testing
- [x] regular testing
- [x] review each locale model to see whether any need deinit() routines on their classes
  - to be done after the fact for numa and knl by @gbtitus and @dmk42

### DIFF
--- a/modules/internal/ChapelLocale.chpl
+++ b/modules/internal/ChapelLocale.chpl
@@ -661,4 +661,11 @@ module ChapelLocale {
   proc chpl_taskRunningCntDec() {
     here.runningTaskCntSub(1);
   }
+
+  //
+  // Free the original root locale when the program is being torn down
+  //
+  proc deinit() {
+    delete origRootLocale;
+  }
 }
--- a/modules/internal/localeModels/flat/LocaleModel.chpl
+++ b/modules/internal/localeModels/flat/LocaleModel.chpl
@@ -218,6 +218,11 @@ module LocaleModel {
       // node.
       return myLocales[chpl_rt_nodeFromLocaleID(id)];
     }
+
+    proc deinit() {
+      for loc in myLocales do
+        delete loc;
+    }
   }
 
   //////////////////////////////////////////
--- a/test/modules/vass/deinit-order-modules.verbose.good
+++ b/test/modules/vass/deinit-order-modules.verbose.good
@@ -34,3 +34,4 @@ uv4ae.deinit UC
   DefaultRectangular
   ChapelArray
   Sort
+  ChapelLocale


bradcray:
@mppf: I took a quick stab at this after last week's deep-dive, and it seems to result in deletion of locale objects upon program termination, at least in the 'flat' locale model.  Still need to test it to make sure there are no surprises.

mppf:
@bradcray - I would think the deinit function would need to traverse the locales tree and delete everything. Is there some special behavior you're using in the destructor that deletes the child locales?

bradcray:
In the flat locale model, I believe it just stores an array of locales as a member and that the array will automatically be freed by the compiler-provided deallocation of the class object.  Adding a deinit() function to the Locale class seemed to confirm this.  My assumption is that any other locale model's implementation of the root locale type would define whatever deinit()s it required to clean up after itself properly...  I haven't looked at our other current locale models to see whether any of them need such user-defined deinit()s, but will take a look.

mppf:
RootLocale stores an array of locales, right, but these are class instances, so while that array will be freed automatically, the class instances themselves won't, right? I mean, doesn't RootLocale need a deinit method to do that?

Maybe you are just saying that since RootLocale is implemented by the locale model, the locale model author could have a deinit method there? If that's the case, I think we should put one in flat so we are not surprised by any testing problems this creates. In particular I expect there is some ordering to the destruction of the individual locale objects that may be important.

bradcray:
Oh, that's a good point... and yet when I added a deinit() routine to the locale model class I did see that it was being called.  I'll look further since I can't explain that...  It seems like I must've messed up (or maybe some other locale class is being freed).

bradcray:
Aha, I see what I did wrong... added the deinit() to the root locale class, not to the locale class itself.

bradcray:
@gbtitus and @dmk42: I'm tagging you on this because I suspect that you'll be able to help me see it through to completion much faster than I could myself.  You may have heard in the deep-dive the other week that Chapel doesn't currently clean up its locale models at program termination which prevents an advanced locale model from doing various teardown activities at that point.  I believe I've introduced the hook that will delete the root locale at the appropriate point in the program (need to do more testing still), but the other thing that needs to happen is that the root locale class for each locale model needs to have a deinitializer (destructor) that cleans up its members.  For example, in the flat locale model, this just requires deleting all of the top-level locales in the locales array (and I suspect that with appropriate destructors for each of the classes that make up other locale models the same would be true for them).

In order to see this effort through, would you have time to introduce such destructors on your locale models, respectively?  My assumption is that you could implement and test them much faster than I could given that you're already working on those locale models.  I'm hoping this is not too hard and will result in a nice bit of clean-up.

gbtitus:
@bradcray, just checking, you're asking us to do our own PRs containing equivalent changes in numa and knl, right?  Not to push commits to your branch for this PR?

bradcray:
I'm open to whatever -- either opening PRs against this branch or doing the work independently and merging after this goes in.  Mostly, I'm trying to get out of implementing and testing the deinits for the locale models that would take me more time to get up-to-speed with and test.

bradcray:
(Also, let me know which approach you want to take, as I think this PR is pretty close to being able to go in).

dmk42:
I'd like to split the KNL locale model off into a separate PR because it's a little complicated, so don't wait for me to merge this.

gbtitus:
I'd like to split the numa locale model off into its own PR as well, so I think you're free to proceed.

PR #6203, opened 2017-05-09
Don't heap promote locals used in on statements under qthreads

ronawho:
As of https://github.com/chapel-lang/chapel/pull/5637, qthread task stacks are
part of the registered heap, so there is no need promote locals used in on
statements anymore.

### DIFF
--- a/compiler/passes/parallel.cpp
+++ b/compiler/passes/parallel.cpp
@@ -1037,14 +1037,14 @@ freeHeapAllocatedVars(Vec<Symbol*> heapAllocatedVars) {
 //  CHPL_COMM == "gasnet" && CHPL_GASNET_SEGMENT == "everything";
 // or
 //  CHPL_STACK_CHECKS == 0 && CHPL_TASKS == "fifo"
+// or
+//  CHPL_TASKS == "qthreads"
 //
 // true otherwise.
 //
-// The tasking layer matters because fifo allocates
-// from task stacks the communication registered heap
-// (unless stack checks is on, in which case it might not
-//  be possible because of huge pages).
-// In the future, we hope that all tasking layers do this.
+// The tasking layer matters because qthreads and fifo allocate task stacks
+// from the communication registered heap (fifo can only do so if stack checks
+// are turned off.)
 static bool
 needHeapVars() {
   if (fLocal) return false;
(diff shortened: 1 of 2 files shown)


ronawho:
@mppf This is a followup on a TODO from #5637 discussed in a mail titled:  "qthreads tasks in Chapel heap". 

However, in putting this together I kept finding myself wondering if we can just get rid of `needHeapVars()` now. As I understand it: `needHeapVars()` was originally needed because some task stacks weren't part of the registered heap and gasnet-{fast, large} didn't support out-of-segment puts/gets. 

However, now that gasnet (and all comm layers) trampoline out-of-segment puts/gets, there's no functional reason for this code, right? There might have been a performance reason to do so, but that's less important now that qthreads task stacks now come from the registered heap, right?

i.e. do you think there's any functional loss if I removed needHeapVars()? If you think that's ok, I'll probably merge this to see if there's any perf impact and then go ahead with removing needHeapVars() (and I think all of makeHeapAllocations())

And just FYI there's also an older internal thread titled "!needHeapVars() when comm=ugni?" that discusses some of this too.

mppf:
@ronawho - thanks for putting this together.

First, about this PR itself, why doesn't needHeapVars need to check fNoStackChecks for qthreads the same way it does for fifo? Since guard pages don't generally make sense on a huge-pages heap... what does qthreads do when stack checks are on and the heap is on huge pages?

Second, I'm absolutely in favor of removing some of this compiler code if we can. As far as I know, we could remove findBlockRefActuals and needHeapVars. It looks like makeHeapAllocations still does something for broadcasting global variables, so it presumably needs to stay.

ronawho:
qthreads force disables guard pages when huge pages are used: https://github.com/ronawho/chapel/blob/648cd8f/runtime/src/tasks/qthreads/tasks-qthreads.c#L621-L623


And yeah, looks like you're right: `findHeapVarsAndRefs()` (terrible name) will still be needed to promote global consts that will be replicated.

mppf:
even so we should check with @benharsh.

benharsh:
I'm not necessarily a parallel.cpp expert, but I think you're both right that ``findHeapVarsAndRefs`` does something useful. I'm not sure about the rest of ``makeHeapAllocations`` though. Testing will likely be required to see what is and isn't useful anymore.

Anyways, there are almost certainly opportunities in parallel.cpp for refactoring and simplification.

ronawho:
@mppf just to confirm: so far as you know, there is no correctness reason that we still have to promote local vars used in on functions to the heap, right? All our comm layers support out-of-segments get/puts?

i.e. you added support for out-of-segment gets/puts in our gasnet shim, and so far as we know that all works correctly?

PR #6207, opened 2017-05-10
Add enum futures

ben-albrecht:
Documenting several enum bugs.


test/types/enum/compare.future
```diff
@@ -0,0 +1,15 @@
...
+
+
+Comparing an enum element to a value results in the following `==` overload
+resolving:
+
+   inline proc ==(param a: int(?w), param b: int(w)) param return __primitive("==", a, b);
+
+Which triggers the following error in postFold.cpp:
+
+   USR_FATAL_CONT(call, "'param' functions cannot return non-'param' values");
```
lydia-duncan: Does this mean that enum constants aren't treated as param values?

test/types/enum/scope.chpl
```diff
@@ -0,0 +1,6 @@
+enumScopeTest(2);
+
+proc enumScopeTest(param x) {
```
lydia-duncan: Looking at the original issue, you should probably add a test for the type argument case, too, in case someone accidentally forgets to fix both (since type and param arguments follow different paths)

ben-albrecht: Good call. Will add.

test/types/enum/compare.future
ben-albrecht: That might be one explanation to the error.

ben-albrecht:
Confirmed that tests work as expected:

```

> start_test test/types/enum/*.chpl

...

Future (bug: Comparing an enum element to a value results in compiler error) [Error matching compiler output for types/enum/compare]
Future (bug: enums declared inside type-arg functions are not initialized correctly) [Error matching compiler output for types/enum/scope-type]
Future (bug: enums declared inside param-arg functions are not initialized correctly) [Error matching compiler output for types/enum/scope]
Future (bug: Enums with explicit types and offsets fail to cast correctly) [Error matching program output for types/enum/typed-offset]
Future (bug: Enums with explicit types fail to cast correctly) [Error matching compiler output for types/enum/typed]
[Summary: #Successes = 1 | #Failures = 0 | #Futures = 5 | #Warnings = 0 ]
[Summary: #Passing Suppressions = 0 | #Passing Futures = 0 ]

```

PR #6212, opened 2017-05-10
Move chpl_getPrivatizedClass() into chpl-privatization.h

ronawho:
This is a second attempt at #6198, but only moves chpl_getPrivatizedClass()
instead of the entire privatization implementation. chpl_getPrivatizedClass()
is just a getter for chpl_privateObjects, so we also need to extern to
chpl_privateObjects.

chpl_getPrivatizedClass() can be called frequently, so we want to allow the
backend compiler to fully optimize/inline calls to it. This has a pretty big
performance impact for the stencil PRK, improving performance by about 15% for
16-node-xc. There's also some minor improvements for fft, and lulesh.

This is motivated by #6184, though it's not quite enough to close that issue
yet.

### DIFF
--- a/runtime/include/chpl-privatization.h
+++ b/runtime/include/chpl-privatization.h
@@ -29,8 +29,8 @@ void chpl_newPrivatizedClass(void*, int64_t);
 
 // Implementation is here for performance: getPrivatizedClass can be called
 // frequently, so putting it in a header allows the backend to fully optimize.
+extern void** chpl_privateObjects;
 static inline void* chpl_getPrivatizedClass(int64_t i) {
-  extern void** chpl_privateObjects;
   return chpl_privateObjects[i];
 }
 
--- a/runtime/include/chpl-privatization.h
+++ b/runtime/include/chpl-privatization.h
@@ -23,8 +23,6 @@
 #include <stdint.h>
 #include "chpltypes.h"
 
-extern void** chpl_privateObjects;
-
 void chpl_privatization_init(void);
 
 void chpl_newPrivatizedClass(void*, int64_t);
@@ -32,6 +30,7 @@ void chpl_newPrivatizedClass(void*, int64_t);
 // Implementation is here for performance: getPrivatizedClass can be called
 // frequently, so putting it in a header allows the backend to fully optimize.
 static inline void* chpl_getPrivatizedClass(int64_t i) {
+  extern void** chpl_privateObjects;
   return chpl_privateObjects[i];
 }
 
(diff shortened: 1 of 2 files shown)


ronawho:
Confirmed that this still gives us the same performance as #6198 for stencil PRK, but resolves all the testing regressions that introduced. Also passed full std and gn-everything testing.

ronawho:
@benharsh, @gbtitus, and @dmk42 mind reviewing (I'm feeling pretty paranoid after the horrible first attempt at this)

runtime/include/chpl-privatization.h
```diff
@@ -23,10 +23,18 @@
 #include <stdint.h>
 #include "chpltypes.h"
 
+extern void** chpl_privateObjects;
+
```
dmk42: I'd recommend putting this extern declaration inside the function chpl_getPrivatizedClass() just to keep the private objects only accessible through the getter.

gbtitus:
👍

dmk42:
So my advice was bad, because we use \-Werror=nested-externs for an unknown reason.

ronawho:
@gbtitus any idea why we throw `-Werror=nested-externs`? Doesn't make much sense to me or David. Unfortunately, the original commit message has no helpful info: https://github.com/chapel-lang/chapel/commit/c8da236e7114a59e20c17510667e151242ab6223

gbtitus:
I have never understood why we throw `-Werror=nested-externs`.  Given the time frame of the change in https://github.com/chapel-lang/chapel/commit/c8da236e7114a59e20c17510667e151242ab6223, I think @bradcray is the only one who could possibly shed light on that.

bradcray:
I can't think of any reason we need to throw -Wnested-externs and don't have a problem with removing it.

dmk42:
Now that we are totally dependent on C99 and later (because of inline functions, for example), would it also be reasonable to remove \-Wdeclaration-after-statement?

bradcray:
I have no opinion on that one.

gbtitus:
I have a mild preference for removing `-Wdeclaration-after-statement`, because I'm prototyping use of a very small third-party package that is written with mixed decls and stmts and I've had to do some Makefile mods to get that to work.  (The package effectively consists entirely of a .h file that is compiled in with the emitted C code for the user program, so I can't just turn off that option when building the third-party package, as I believe we do in some other cases.)

dmk42:
I will submit a PR that removes both `-Wnested-externs` and `-Wdeclaration-after-statement`.

PR #6213, opened 2017-05-10
MPI testing cleanup

ben-albrecht:
In this PR:

- Utilize recursive skipifs instead of directory-wide SKIPIFs
- Utilize directory-wide EXECENVs instead of test-specific execenvs
- Utilize sorting PREDIFFs instead of trying to make MPI barrier test near-deterministic with sleeps, etc.
- Extend `test-mpicc.sh` directory to `test/modules/packages/mpi`, which should only apply to SPMD tests due to skipifs, and will allow adding more SPMD tests.

**Testing**
- [x] `start_test test/modules/packages/mpi` on linux64 with `mpicc` configuration


ben-albrecht:
@awallace-cray - could I get your blessing  on the `test-mpicc.sh` change?

PR #6216, opened 2017-05-11
Factored out most Popen calls in sub_test

vasslitvinov:
sub_test used to have a lot of subprocess.Popen calls
followed by p.communicate and other things. Most of these
calls were similar yet subtly different. How the output
and the return code of the invoked process were handled
was not obvious.

This change standardizes most invocations of Popen into
helper functions `runAnd...` - so that the policies of handling
of the output and return code can be understood and changed easily

This PR does not change any policies, except when calling 'cat'
on some files like LASTCOMPOPTS the presence of stderr output
now results in a line [Error...] in the log.

While there, I replaced one rstrip() with strip(), for consistency,
and changed an error message to print out the actual name of the
program that was invoked.
Also, I factored a bunch of sys.stdout.flush() into runAndShowOutput().

Testing:

* `start_test release/examples -compperformance` with and without my changes
==> contents match, modulo dates and #seconds

* linux64 paratest with --verify:
==> correct number of tests
==> all "Success" lines match those from the nightly linux64 log,
except those printed when compiling .future tests.

* `start_test --performance`:
==> no warnings, errors, failures
==> same set of 'Success matching performance keys' as from a nightly run
==> same 'Success compiling' lines.

### DIFF
--- a/util/test/sub_test
+++ b/util/test/sub_test
@@ -202,6 +202,105 @@ def LauncherTimeoutArgs(seconds):
               useLauncherTimeout)
 
 
+# Execute command.
+# Its stdout+stderr go to sub_test's stdout+stderr.
+# Return its exit code.
+def runAndShowOutput(command):
+    sys.stdout.flush()
+    p = subprocess.Popen(command)
+    rc = p.wait()
+    return rc
+
+# Execute command. Rc = Return code.
+# Gather its combined stdout+stderr.
+# Return (its exit code, the gathered output).
+def runAndReturnRcOutput(command):
+    p = subprocess.Popen(command,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.STDOUT)
+    output = p.communicate()[0]
+    return p.returncode, output
+
+# Execute command. EIEO = Error If Error Output.
+# Gather *separately* its stdout and stderr.
+# If stderr is non-empty or the exit code is non-0, report an error.
+# If errfatal, quit with Fatal(). Otherwise, 
+# return the return code (at least 1 if got stderr) and command's stdout.
+def runAndReturnRcOutputEIEO(command, errfatal=False):
+    p = subprocess.Popen(command,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    output = p.communicate()
+    errmsg = None
+    rc = 0
+
+    if output[1]:
+        errmsg = "unexpected stderr output while executing {0} (rc={1})".format(' '.join(command), p.returncode)
+        rc = 1
+        if errfatal:
+            sys.stdout.write("While executing {0}:\n".format(' '.join(command)))
+            sys.stdout.write("stdout=\n{0}stderr=\n{1}".format(trim_output(output[0]), trim_output(output[1])))
+        else:
+            sys.stdout.write("[Error: {0}]\n".format(errmsg))
+            sys.stdout.write("The stderr output was:\n{0}".format(trim_output(output[1])))
+    elif p.returncode:
+        errmsg = "rc={0} while executing {1}".format(p.returncode, ' '.join(command))
+        if not errfatal:
+            sys.stdout.write("[Error: {0}]\n".format(errmsg))
+        rc = p.returncode
+
+    if errfatal and errmsg:
+        Fatal(errmsg)
+
+    return rc, output[0]
+
+# Same as runAndReturnRcOutputEIEO() except drop return code.
+def runAndReturnOutputEIEO(command, errfatal=False):
+    rc, output = runAndReturnRcOutputEIEO(command, errfatal)
+    return output
+
+# Return the contents of the file. CFE = Checking For Error.
+# Report an error and proceed if something went wrong.
+def fileContentsCFE(filename):
+    p = subprocess.Popen(['cat', filename],
+                         stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


vasslitvinov:
@ben-albrecht or @mppf or @ronawho would you review?

util/test/sub_test
```diff
@@ -202,6 +202,105 @@ def LauncherTimeoutArgs(seconds):
...
+
+# Same as runAndReturnRcOutputEIEO() except drop return code.
+def runAndReturnOutputEIEO(command, errfatal=False):
+    rc, output = runAndReturnRcOutputEIEO(command, errfatal)
+    return output
+
+# Return the contents of the file. CFE = Checking For Error.
+# Report an error and proceed if something went wrong.
+def fileContentsCFE(filename):
+    p = subprocess.Popen(['cat', filename],
```
ronawho: If you're going to go through this effort, we should stop shelling out to call `cat`

We should just read the files contents in pure python

util/test/sub_test
```diff
@@ -202,6 +202,105 @@ def LauncherTimeoutArgs(seconds):
...
+                         stderr=subprocess.STDOUT)
+    output = p.communicate()[0]
+    return p.returncode, output
+
+# Execute command. EIEO = Error If Error Output.
+# Gather *separately* its stdout and stderr.
+# If stderr is non-empty or the exit code is non-0, report an error.
+# If errfatal, quit with Fatal(). Otherwise, 
+# return the return code (at least 1 if got stderr) and command's stdout.
+def runAndReturnRcOutputEIEO(command, errfatal=False):
```
ronawho: `Rc` and `EIEO` are not helpful or easy to remember names in my opinion

util/test/sub_test
```diff
@@ -202,6 +202,105 @@ def LauncherTimeoutArgs(seconds):
...
+            sys.stdout.write("[Error: {0}]\n".format(errmsg))
+            sys.stdout.write("The stderr output was:\n{0}".format(trim_output(output[1])))
+    elif p.returncode:
+        errmsg = "rc={0} while executing {1}".format(p.returncode, ' '.join(command))
+        if not errfatal:
+            sys.stdout.write("[Error: {0}]\n".format(errmsg))
+        rc = p.returncode
+
+    if errfatal and errmsg:
+        Fatal(errmsg)
```
ronawho: Do we really want to use Fatal? Fatal will quite out of sub_test completely, and skip any remaining tests. The code you removed didn't call Fatal, right?

ronawho:
I did not review everything yet. I'd like to see the removal of shelling out to `cat` and then do a full review

util/test/sub_test
vasslitvinov: What do you suggest?

My motivation was to avoid extra-long names.

util/test/sub_test
vasslitvinov: We did, like so:

```python
c_compiler = p.communicate()[0].rstrip()
if p.returncode != 0:
  Fatal('Cannot find c compiler')
```

util/test/sub_test
mppf: I don't think we need all the variations but if you want your PR to not change behavior we do. Perhaps we could choose a handful with good names that we intend to use exclusively in the future?

vasslitvinov: How do you all feel about mostly not changing the behavior upon this first PR, then following up with behavior changes separately?

Meanwhile let us choose good names right away. We could leave non-good names for functions that we expect to retire, if we can't come up with good ones.

vasslitvinov: How about these?
```
runAndReturnStatusAndOutput
runAndReturnStatusAndOutputUnlessError
runAndReturnOutputUnlessError
fileContentsOrError
fileContentsUnlessError
```
What are good variations on the last two for multiple files?

ben-albrecht: I think we could consolidate many of these functions into single functions differentiated with arguments, if we're willing to take advantage of python exception handling and refactor the call-sites a bit:

```python
def run_command(command, verbose=False, errfatal=False):
    """Execute command and return exit code

    Optionally print stdout+stderr if verbose==True
    Optionally raise an error upon non-empty stderr if errfatal==True
    """


# Note: will need to refactor call sites to try/except these calls, instead of
# getting the exit code in the return
def check_output(command, errfatal=False):
    """Run command and return output, raise CalledProcessError if exitcode != 0

    This is a pseudo-backport of subprocess.check_output for python 2.6
    Optionally raise an error upon non-empty stderr if errfatal==True
    """

# Instead of checking getting exit code as an argument, we'd catch the exception thrown, e.g.
cmd = ['ls', '-l']
try:
    output = check_output(cmd)
except subprocess.CalledProcessError:
    # exitcode was not zero - do some stuff



# I've included an example implementation of how these could be combined
# Note that we no longer need a stderr case, since we're not calling out to subprocess
def file_contents(filenames):
    """Return concatenated contents of file(s) as string

    Raise error if file does not exist"""

    if type(filenames) is not list:
        filenames = [filenames]
    contents = []
    for filename in filenames:
        if not os.path.isfile(filename):
            raise IOError(filename + ' not found')
        with open(filename, 'r') as handle:
            contents.extend(handle.readlines())
    return ''.join(contents)
```

What do others think?

vasslitvinov:
Temporarily closing, by Elliot's request.

PR #6217, opened 2017-05-11
Fix two --baseline regressions

daviditen:
In the iterator `getValue()` function, if the return type is `void` then
return the global void value instead of trying to look up the field in the
iterator class.

This problem didn't show up in regular testing because
deadCodeElimination covered it up.

This fixes the `--baseline` regressions for:
```
functions/iterators/diten/yieldNothingIterator
functions/iterators/diten/yieldVoidIterators
```

Tested with:
```
paratest -compopts --baseline (one regression remains)
paratest -compopts --verify (clean)
```


vasslitvinov:
Looks good to me.

daviditen:
Thank you!

PR #6221, opened 2017-05-11
Stabilize replicated alloc comm counts

bradcray:
It turns out that the reason that the number of communications has been unstable for the alloc*.chpl tests in the distribution robustness suite's comm-counts / performance directory since array views went in is that there is remote parallel contention on an atomic lock which can cause a locale to try and get the lock multiple times if someone already has it.  Specifically, the array views work transformed the set of target locales in the replicated distribution into an associative domain, and accesses to associative domains, like the ones used in tearing down the replicated array, currently attempt to take a lock (arguably needlessly as noted in the comments section of this PR #6221).

I found that I could fix this by rewriting the parallel loop to avoid operations that result in the random access by using zippered iteration.


bradcray:
@benharsh : I think you're the obvious reviewer of this, both as someone who knows the associative stuff well, and as the developer who introduced the associative domain of target locales into the replicated distribution.  Bonus points if you can explain why the previous version resulted in more comms and contention on an atomic than I've been able to here.

benharsh:
Nothing really jumps out at me as to the cause for more comms/contention.

Otherwise this PR looks good 👍

bradcray:
Is it clear why we even would try to take an atomic for indexing into an associative array?  I'd think for a read-only operation like that, there'd be no need to take a lock...  Is it the case that we implemented locking for the original read/write dsiAccess() and that now that we've separated reads from writes the reads could avoid needing to take the lock while the writes would keep it?

benharsh:
Don't we need the lock in case the table resizes in another task?

bradcray:
I think that, in general, we assume that it's not safe to read an array (of any type) while some other task is modifying it and that the burden of guarding against this is on the user, not the language/library.  E.g., even for an array:

````chapel
var D = {1..n};
var A: [D] real;
````

if we had to worry some other task resizing D while we were reading A, we'd need to lock every array access.  It's not obvious to me that associative domains/arrays are different in this respect.  The main way in which they are different is that rectangular domains change their index set in an O(1) operation regardless of the number of indices being added to/removed from the domain, so we tend not to worry about concurrent changes to the index set itself.  Whereas with associative (and sparse) domains and arrays, since the nature tends to be O(n) work to add O(n) indices, we go to greater trouble to permit multiple domain updates to occur simultaneously (and this might also include write-style dsiAccess() calls in the event that the array has the 1:1 relationship with its domain that permits writes to turn into resizes).

benharsh:
Good point. With that in mind, I suspect you're correct and that the value and const-ref dsiAccess methods could be rewritten to avoid locking.

PR #6228, opened 2017-05-12
Fix typos

cassella:
Fix typos, in compiler/resolution, man/chpl.rst, and chip 19.

chip 19 note: Google and compiler/codegen suggest it's "Schubert Numbering", not
"Shubert Numbering".


lydia-duncan:
Running a sanity paratest, but since this is impacting comments and documentation, I don't expect any failures

PR #6233, opened 2017-05-12
Make targetLocales more consistent and flexible in ReplicatedDist

bradcray:
This commit makes the use of the target locales accepted by ReplicatedDist
more consistent: previously, we sometimes hashed on locale.id and
sometimes on index(targetLocaleDomain)) which required us to state a
"consistency" requirement on the targetLocales array passed to a ReplicatedDist.
Now we can drop that requirement.  Added three tests that demonstrate that
this is the case.


bradcray:
@benharsh and/or @vasslitvinov: Would one or both of you have time to give this a review?

PR #4517, opened 2016-09-13
Revert #4476 (remove function resolution broken optimization) and comment it

lydia-duncan:
When this optimization was removed (because it is buggy and
unmaintainable), we suffered a 10% compilation slowdown
across all of our tests.  There is not enough time before feature
freeze to reimplement this optimization in a form that is less
fragile, and with the merge of #4503, we are much less likely to
run into this bug (though not assured it is gone even for the moment,
as we know the List module still exhibits this behavior.  However,
List is less likely to be used than the Search and Sort modules),
so we decided to wait to handle the bug further until after the
release.

Add comments to the optimization and its uses, detailing why
it could be a source of errors, in case we lose track of it again
and further code drift causes new errors.

Tested over standard

PR #6235, opened 2017-05-12
Fix a long-standing bug for scoped access to internal modules.

mppf:
See also PR #4476 and PR #4517 which reverted it.  This PR makes a simple change to findVisibleFunctions to correctly handle the case when a scoped access to a module provided automatically does not compile; for example 

``` chapel
  writeln(List.makeList(1,2,3));
```

(with or without a 'use List').

findVisibleFunctions contains an optimization which is important for the performance of the compiler. This optimization is to store all included-by-default symbols into one scope instead of performing extra bookkeeping for these scopes across all of the included-by-default code.  The code handling List.makeList (e.g.) simply needed to be updated to use this one scope when appropriate. 

Passed full local testing.


lydia-duncan:
The second sentence in your first paragraph has a typo but I'm not sure how to fix it.

compiler/resolution/visibleFunctions.cpp
```diff
@@ -31,6 +31,24 @@
...
+
+   In this way, the symbols declared in each block are stored.
+   These tables do not save the transitive symbols available.
+   To do that, code using the table needs to go up blocks/scopes
+   and explicitly consider module 'use's.
+
+   This file also includes an optimization, where the
+   symbols available to all modules (i.e. what is in ChapelStandard)
+   is considered to be in a single block. This optimization
+   provides a significant performance for compiling 'hello'.
```
lydia-duncan: Missing the word "improvement" here?

mppf:
In the PR message? I think I fixed it.

mppf:
I don't expect any impact on compiler performance (but havn't looked at it in detail). I'll fix the missing word in the comment.

PR #6236, opened 2017-05-12
At Mike's suggestion, move definition of defaultTypeConstructor

lydia-duncan:
The field defaultTypeConstructor was present on all Type AST nodes, but
was only set for its subclass AggregateType.  This PR converts the accesses
of this field that weren't already explicitly on AggregateTypes to be on that
subclass, and then moves the definition of the field from the parent class
to the child.

I intend to tackle defaultInitializer in a separate PR.

Passed a full std paratest


compiler/AST/AggregateType.cpp
```diff
@@ -44,8 +44,10 @@ AggregateType* dtOnBundleRecord    = NULL;
 AggregateType* dtTaskBundleRecord  = NULL;
 AggregateType* dtError             = NULL;
 
-AggregateType::AggregateType(AggregateTag initTag)
-  : Type(E_AggregateType, NULL) {
+AggregateType::AggregateType(AggregateTag initTag) :
+  Type(E_AggregateType, NULL),
+
+  defaultTypeConstructor(NULL) {
```
noakesmichael: The other fields are initialized within the body of the constructor.  For consistency
this one should be too.

compiler/resolution/tuples.cpp
```diff
@@ -799,13 +800,14 @@ do_computeTupleWithIntent(bool valueOnly, IntentTag intent, Type* t)
...
 
-  FnSymbol*  typeConstructor    = t->defaultTypeConstructor;
-  BlockStmt* instantiationPoint = typeConstructor->instantiationPoint;
-
   // Construct tuple that would be used for a particular argument intent.
 
   bool allSame = true;
   AggregateType* at = toAggregateType(t);
+
+  FnSymbol*  typeConstructor    = at->defaultTypeConstructor;
```
noakesmichael: I think lines 806 and 808 imply that `t' must also be an AggregateType.  Does that seem right to you?
If so I'd be inclined to update the function signature accordingly.

compiler/resolution/virtualDispatch.cpp
```diff
@@ -75,8 +75,8 @@ static void addToVirtualMaps   (FnSymbol* pfn, AggregateType* ct);
 static void collectMethodsForVirtualMaps(Vec<FnSymbol*>& methods,
                                          Type*           ct,
                                          FnSymbol*       pfn);
-static void collectInstantiatedAggregateTypes(Vec<Type*>& icts,
-                                              Type*       ct);
+static void collectInstantiatedAggregateTypes(Vec<AggregateType*>& icts,
```
noakesmichael: Is this a good opportunity to

a) convert Vec<t> to std::vector<t>

b) Follow the common convention that formals that are effectively output
values should be grouped as the final formals rather than being first.

Not wedded to either of these but I know I would take this opportunity.

compiler/resolution/virtualDispatch.cpp
```diff
@@ -477,16 +477,18 @@ static void collectMethodsForVirtualMaps(Vec<FnSymbol*>& methods,
 //
 // add to vector icts all types instantiated from ct
 //
-static void collectInstantiatedAggregateTypes(Vec<Type*>& icts, Type* ct) {
+static void collectInstantiatedAggregateTypes(Vec<AggregateType*>& icts,
```
noakesmichael: Ditto

compiler/resolution/tuples.cpp
lydia-duncan: I'm little leery of doing that since tuples are so mystical, but I'll give it a shot

compiler/resolution/virtualDispatch.cpp
lydia-duncan: Good point!

PR #6241, opened 2017-05-12
Also at Mike's suggestion, move definition of defaultInitializer

lydia-duncan:
The field defaultInitializer was present on all Type AST nodes, but
was only set for its subclass AggregateType. This PR converts the accesses
of this field that weren't already explicitly on AggregateTypes to be on that
subclass, and then moves the definition of the field from the parent class
to the child.

This PR should be merged after #6236 

Passed a full paratest prior to the first PR.  I intend to run a second with
an update from master before merging


compiler/resolution/lowerIterators.cpp
```diff
@@ -57,9 +57,11 @@ FnSymbol* getTheIteratorFn(CallExpr* call) {
 FnSymbol* getTheIteratorFn(Type* icType)
 {
   // the asserts document the current state
+  AggregateType* icTypeAgg = toAggregateType(icType);
```
noakesmichael: I think lines 60 and 61 imply the icType must be an aggregate type.  So one
wonders why it accepts a static Type?

lydia-duncan: I think I will handle that in a different PR, so I can merge this one today and focus on other things for a little while

PR #6243, opened 2017-05-12
Publicize reindex / localSlice

ben-albrecht:
Add `reindex()` and `localSlice()` methods to the public documentation.

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -618,7 +618,7 @@ module ChapelArray {
   }
 
   //
-  // Returns a domain with a rank equivalent to chpl__getActualArray(arr).rank.
+  // Return a domain with a rank equivalent to chpl__getActualArray(arr).rank.
   // This domain is no larger than the innermost array's domain. It represents
   // the 'active' indices that the top-level ArrayView works with. For example:
   //
@@ -634,7 +634,7 @@ module ChapelArray {
   }
 
   //
-  // Returns the innermost array class (e.g., a DefaultRectangular).
+  // Return the innermost array class (e.g., a DefaultRectangular).
   //
   // 'arr' can be a full-fledged array or a BaseArr-inheriting class
   //
@@ -645,7 +645,7 @@ module ChapelArray {
   }
 
   //
-  // Returns true if 'arr' is a DefaultRectangular array or is an ArrayView
+  // Return true if 'arr' is a DefaultRectangular array or is an ArrayView
   // over a DefaultRectangular array.
   //
   // 'arr' can be a full-fledged array type or a class that inherits from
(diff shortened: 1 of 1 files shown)


ben-albrecht:
@benharsh - Care to review / wordsmith my documentation?

benharsh:
Looks fine to me 👍

modules/internal/ChapelArray.chpl
```diff
@@ -1613,12 +1613,14 @@ module ChapelArray {
       return this((...r));
     }
 
-    pragma "no doc"
+    /* Return an array slice halting if the range contains elements
+       that are not local. */
```
bradcray: This routine doesn't take a range, it takes a vararg number of ranges (essentially, an anonymous rectangular domain).

modules/internal/ChapelArray.chpl
```diff
@@ -2293,7 +2295,17 @@ module ChapelArray {
              "  Actual domain is: ", this.domain);
     }
 
-    pragma "no doc"
+    /* Return an array declared over a new domain, provided that the new
+       domain is of the same rank and size as the original.
```
bradcray: This wording suggests to me that this routine creates a new array (with new storage associated with it), when in fact it returns a view of the original array, simply using new indices.

modules/internal/ChapelArray.chpl
```diff
@@ -1613,12 +1613,14 @@ module ChapelArray {
...
-    pragma "no doc"
+    /* Return an array slice halting if the range contains elements
+       that are not local. */
     proc localSlice(r: range(?)... rank) {
       return _value.dsiLocalSlice(chpl__anyStridable(r), r);
     }
 
-    pragma "no doc"
+    /* Return an array slice halting if the domain contains elements
+       that are not local. */
```
bradcray: In the current wording, I'd put a comma after "slice".

The more important aspect of this routine isn't that it halts if the slice is out-of-bounds -- it's that it returns a local view of the sub-array which is cheaper to index in subsequent operations because the indices are known to be local (so there are no checks to see which locale owns the indices, whether communication is required, etc.)

In this routine and the previous, I worry that the term "slice" may not be sufficiently well-known that these descriptions will be clear to someone who doesn't already know what the routines do.

modules/internal/ChapelArray.chpl
```diff
@@ -2293,7 +2295,17 @@ module ChapelArray {
...
 
-    pragma "no doc"
+    /* Return an array declared over a new domain, provided that the new
+       domain is of the same rank and size as the original.
+
+       For example:
+
+       .. code-block:: chapel
+
+          var A: [{1..10}] int;
```
bradcray: I'd remove these curly brackets.  I don't think we want to propagate a notion that this is required for Chapel (nor the preferred style).

ben-albrecht: Whoops - good catch.

modules/internal/ChapelArray.chpl
ben-albrecht: Do we have documentation on array views that I can cross-reference when using the term *array view* ?

ben-albrecht:
@bradcray  - Thanks for comments. Let me know what you think of recent changes.

PR #6244, opened 2017-05-13
Reduce amount of locking in associative array dsiAccess

### LINKED ISSUE #6223: Associative arrays too conservative in locking
While debugging a nondeterministic number of communications in PR #6221, I was surprised to find that we were taking a lock (via an atomic) in order to do a read-only access to an associative array (see the discussion between @benharsh and myself in the comments section of that PR for details).  This issue is designed to capture an action item that I think we should not be taking these locks for array accesses.

This conservatism w.r.t. locking seems to have originally stemmed from a concern that someone may be resizing the associative array as we're reading from / writing to it. -- however, this risk is true of any of our domains and arrays.  E.g., resizing the domain of a 1D rectangular array while someone is reading from / writing to it could have similarly disastrous consequences.  For this reason, we've traditionally said "don't do that," and the burden is on the user not to do so to avoid having to lock every read/write.  While we could reconsider that position or have a way to opt into something safer, that's not what I'm interested in here.

Initially, I was thinking that we should continue taking the lock for write accesses in order to guard against the event that the write could resize in the array (which is only supported when the associative array and its domain have a 1:1 relationship).  However, upon exploring the implementation and thinking about it a bit more, it seems that even locking in that case is not particularly well-motivated.  I was originally thinking that it would help with cases like this:

```chapel
forall key in collection do
  A[key] += 1;
```

by permitting multiple accesses to associative array `A` to occur simultaneously, potentially growing it.  But @benharsh pointed out that after the dsiAccess() returns the reference to the array element it computes, another task could resize `A` before the += operator had the opportunity to update the array element.  So locking in this case does not seem particularly well-motivated either (see the discussion on outdated code diffs in PR #6244 for further wrestling with this question).

bradcray:
This PR cleans up and simplifies the locking done in associative arrays.  It was originally motivated by an observation that our dsiAccess() routines were locking on every access, presumably due to fear that the domain/array may be resized while the access was occurring.  But this is true of all arrays (e.g., even 1D arrays may be resized during access) and not something we traditionally guard against, so it seemed unnecessarily heavyweight.

Moreover, while such locking may give the illusion that code like the following is safe:

```chapel
forall key in collection do
  A[key] += 1;
```

it actually isn't.  Namely, between the time that the dsiAccess() routine calculates the reference to `A[key]` and the += operator tries to update it, another task may have simultaneously reallocated the array, invalidating the reference returned by dsiAccess().

While here and reviewing the code, Michael noted other opportunities for clean-up and simplification which led to some thematically related changes:

* removed the 'haveLock' arguments from the dsiAccess() routines because they weren't being used anymore.
* similarly, removed the 'haveLock' arguments from clearEntry() and findEmptySlot() routines on the argument that they were never actually used at callsites and/or in the bodies of the routines themselves
* changed existing 'haveLock' arguments to 'needLock' on the argument that it's clearer:  'haveLock' was used to indicate either that we have a lock or that we don't need a lock.   'needLock' says more clearly that we need it and don't have it.  Reversed the sense of actuals sent to such arguments and their default values.

This PR resolves #6223.


bradcray:
@benharsh and/or @mppf seem like the most appropriate reviewers here.

modules/internal/DefaultAssociative.chpl
```diff
@@ -566,14 +566,15 @@ module DefaultAssociative {
 
     // ref version
     proc dsiAccess(idx : idxType, haveLock = false) ref {
-      const shouldLock = dom.parSafe && !haveLock;
+      const arrOwnsDom = dom._arrs.length == 1;
+      const shouldLock = dom.parSafe && !haveLock && arrOwnsDom;
```
mppf: As I understand it, the locking is used to ensure that the domain resize operation happens under a lock. What if this dsiAccess first tried to get the element without locking, and then if it is not found try again with locking? (Noting that on the second try, the element might have appeared in the meantime)? That way, the only time the lock is invoked on element access is when actually adding an index to the domain.

bradcray: Just to make sure I'm interpreting this correctly -- should I read your "What if" as "Do you think it could/should do this?"  (at first I read it as proposing a hypothetical situation that could be problematic).

mppf: yes. it's a suggestion.

bradcray:
@mppf: Is that last commit what you had in mind?  (assuming I got it right -- re-testing now)

bradcray:
Hmmm... hold off, testing suggests that I've done something wrong (the reason I've been putting off replying to your feedback... I wasn't convinced I'd be able to get it right easily).

bradcray:
OK, this is ready for re-review now.

mppf:
I can see that associative domain iteration doesn't take the lock in either case. I assume that's what we want & also that it's consistent with other array types, but I'm not sure it's documented well. It seems like something we'd want to put a warning label on.

modules/internal/DefaultAssociative.chpl
```diff
@@ -566,18 +566,48 @@ module DefaultAssociative {
 
     // ref version
     proc dsiAccess(idx : idxType, haveLock = false) ref {
-      const shouldLock = dom.parSafe && !haveLock;
-      if shouldLock then dom.lockTable();
-      var (found, slotNum) = dom._findFilledSlot(idx, haveLock=true);
+
+      // Start by attempting to do an unlocked lookup of the value
+      var (found, slotNum) = dom._findFilledSlot(idx, haveLock=haveLock);
```
mppf: I would think this should use haveLock=true in order to do unlocked access.

mppf:
I think it would be less confusing if we renamed haveLock from _findFilledSlot to something else.
There are several cases:
 * we don't have the lock and the lock isn't necessary (currently, haveLock=true)
 * we have a lock (currently, haveLock=true)
 * we don't have the lock and the lock is necessary (currently, haveLock=false).

E.g. it could be `needsLock=false` for the first two cases and `needsLock=true` for the last.

I'm also not convinced we need a haveLock argument to dsiAccess at all (since you never have the lock when it's called). The current exception is `clearEntry` which:
 * is only called from one place (dsiRemove)
 * is not user facing
 * is only ever called while holding the lock
 * is only ever called on an element that already exists

Therefore, it could be calling a dsiAccess with haveLock=true. `clearEntry` is a bit confusing but it seems to be a type system hack (it is a virtual method on BaseArr so that it can be called on the elements of _arrs). I'm having trouble seeing what we gain from calling clearEntry instead of dsiAccess in this case.

So, I recommend removing `haveLock` from all the dsiAccess functions. It seems to me that `haveLock` is getting really confusing under the current use, and so tidying up would really help. However it would be OK to do it in a future PR.

Also, I think it's critical that we document the "rules" about locking at the top of the file:
 * concurrent domain update and element access is not supported
 * concurrent domain update and domain iteration is not supported

Therefore,
 * domain updates generally use locking
 * array access doesn't need locking unless it implicitly adds an element
 * domain and array iteration doesn't need locking

bradcray:
> I'm having trouble seeing what we gain from calling clearEntry instead of dsiAccess in this case.

It appears to me that we can't access the `eltType` of the array at the `clearEntry()` callsite because at that point it's a `BaseArr` which does not (currently) have an `eltType` field.

mppf:
@bradcray - thanks for the updates & explanations. I'm happy with this but if you feel nervous about locking & races you might want to ask for a second set of eyes on it. Also, could you make a TODO/issue/comment/something about documenting the rules for concurrent access and modification of domains/arrays (which I brought up in my other comment?).

bradcray:
> could you make a TODO/issue/comment/something about documenting the rules for concurrent access and modification of domains/arrays (which I brought up in my other comment?).

This was not intended to be ready for re-review yet.

modules/internal/DefaultAssociative.chpl
bradcray: @mppf: On reflection, I'm not sure that this was a good change to make.  Imagine the following code:

```chapel
forall key in collection do
  A[key] += 1;
```

If we only lock writes to `A` which resize the table, then any writes to `A` which happen to be to an existing key may happen in an unlocked manner and therefore may run afoul of a simultaneous resizing being done by another write.  This would suggest that all write accesses to `A` should lock, whether or not they result in a resizing.  But we could restrict the locking to writes to `A` which have the 1:1 domain:array property (since writes to other associative cases won't resize but will result in an OOB error).  Do you buy that?

Ideally, we could also avoid the lock in write accesses of associative arrays with const domains (since they also can't be resized), but I don't think we can tell whether an array's domain is const or not from within module code.

mppf: >  I don't think we can tell whether an array's domain is const or not from within module code.

not yet, but it's been coming up a lot lately....

Anyway, regarding the rest... it seems like a slippery slope.
What if I wrote

``` chapel
var total: atomic int;
forall key in collection {
  if A.domain.member(key) {
    total += A[key];
  } else {
    A[key] = 0;
  }
}
```

? Wouldn't that be really confusing too? I'm trying to construct an example where there is a race between unlocked reads and locked writes. (Something I was thinking about before... being able to say an operation is "atomic on" some data structure, would help here. You'd just make the whole loop body "atomic on" the domain).

Maybe the answer is to just say that the 1:1 associative array key creation is not parallel safe.

We've talked before about how it'd probably be better to make such array-operations-modifying-domains just be their own separate array types.

bradcray: To me, your new example seems to be across the line where the user is doing enough disparate operations on a single associative array that they should be expected to handle any locking required to make it safe themselves.  Even the 'if A.domain.member(key)' check seems suspect since there's a potential race (what if it is a member by the time they're done checking it?).

mppf: Right, but what worries me is that if a user didn't know that we supported the example you wrote, if they wrote the longer version of the same, it would be wrong. 
``` chapel
forall key in collection {
  if D.member(key) then
    A[key] += 1;
  else
   D += 1;
```

which makes me think that the 1:1 auto-domain addition is not a shorthand at all.

bradcray: I'm confused:

* It isn't a shorthand for the code you wrote here, but surely it's a shorthand for something...?  
* In any case, I'm not sure what you're concluding...

Since it doesn't seem that we're disagreeing that the "lock only when a write access misses and needs to grow things" isn't the right solution, I've pushed a new commit that locks more aggressively on write dsiAccesses when the 1:1 property exists as a placeholder for the time being. 

I also think we're going to need @benharsh (if not others) to look back in on this PR and weigh in.

bradcray: @benharsh stopped by to point out that even on master or in the current version of this branch, there's still a race even in my simple forall loop:

```chapel
forall key in collection do
  A[key] += 1;
```

Specifically, the `A[key]` expression gets a reference to a particular location, but then nothing prevents another simultaneous access from growing the domain and invalidating the reference before the += operation has a chance to update that reference.  So that also might argue for not doing any locking on any write-base dsiAccesses (which I think is what you were proposing a few messages above, and maybe in your last message as well).

bradcray:
I think the code changes here are ready for re-review by @mppf and/or @benharsh (or even @sungeunchoi).  They've passed linux64 testing as well as 1000-trial stress testing on some key associative tests that were added to avoid races.  I'm holding off on doing any documentation changes until we're in agreement about the code/semantics.

modules/internal/DefaultAssociative.chpl
mppf: Re my example, I meant it to be the long form of the simple forall loop you've been using (repeated above). I just didn't write it right. Here's another attempt:
``` chapel
forall key in collection {
  if D.member(key) then
    A[key] += 1;
  else {
   D += key;
   A[D] = 1;
  }
}
```

My reasoning is this:
 * I think the modifies-domain stuff is a convenience/shorthand and so shouldn't change memory consistency / data races from how you'd write it without using the shortcut
 * I don't see how to write it in a race-free way without using the short form

Therefore I'm questioning:
 a) whether we want the special 1:1 behavior at all
 b) whether it makes sense for the existence of the 1:1 behavior to impact our parallel semantics. That is, if the "long form" of the same loop obviously wouldn't be thread safe, why should we go to effort in the implementation to make the short form be so?

I think I'd rather expect the user to use a more general mechanism (it could be - these statements happen under the domain's lock; or require them to use a separate lock; or ideally the `atomic` statement).

mppf:
I've looked at this enough to be happy with it but I'd appreciate a 2nd opinion from @benharsh.

bradcray:
@mppf and/or @benharsh: In the last commit, I've added documentation regarding parallel safety to try and clarify our current position and the behavior of the 'parSafe' param on associative domains.  Would one of you review please?

PR #6245, opened 2017-05-13
5/11 perf meeting updates

psahabu:
@chapel-lang/perf-team:
- Added annotations for #6212:
  - PRK stencil graphs
  - NPB
- added y-axis to array init tests
- removed empty task spawning from perf tracking
- PRK stencils were already part of 16-node-xc perf tracking

Having trouble running `check_annotations.py`, so I'm holding off until Monday.

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -44,6 +44,8 @@ all:
   03/28/14:
     - Add static flag for performance testing (22992)
   05/08/14:
+  05/11/17:
+    - 
     - switched to gcc-4.7
   07/19/14:
     - no testing occurred, machine update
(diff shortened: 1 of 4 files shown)


ronawho:
What does "Having trouble running check_annotations.py" mean? If it's giving you any output beyond "No fatal annotation errors detected", that probably means you have a badly formatted annotation

test/ANNOTATIONS.yaml
```diff
@@ -44,6 +44,8 @@ all:
   03/28/14:
     - Add static flag for performance testing (22992)
   05/08/14:
+  05/11/17:
```
ben-albrecht: This may be the cause of the `check_annotations.py` failure

psahabu:
@ronawho Yup that was the case, I just had absolutely no recollection of pasting the lines that @ben-albrecht pointed out so I was very confused by the error.

ben-albrecht:
@psahabu -- it doesn't look like you pushed your fix before merging?

psahabu:
@ben-albrecht:
Wow this is a rough Monday, I can't believe I overlooked that. On it.

Issue #6239, opened 2017-05-12
testRelease should fail if making docs fails


bradcray:
It seems that ``util/buildRelease/testRelease`` does not fail if its attempt to make module documentation fails, and this can spiral into other errors downstream.  Specifically, if the temporary modules/standard/SysCTypes.chpl used to generate documentation is not removed (as it is not if this build step fails), the compiler will rightfully complain of an ambiguous module.  It seems to me that a failure to make the docs should kill the testRelease script at that point rather than leaving open the possibility of rippling into other errors or having the error go undetected.  

That said, I'm not sure offhand what it would take to do this.  From what I can tell, 'make module-docs' returns an error code (2), so it would seem to be something upstream from that step.  The following shows a case that went wrong when the deinit() procedure in ChapelLocale.chpl was not marked as "no doc" (and removing the "no doc" pragma should permit this to be reproduced, at least until such time as that internal error is fixed).  If testRelease worked as I believe it should, the following lines should've been the last ones seen in the log rather than lines much further on when trying to do 'make test' on the Chapel compiler itself.  This would make the source of the problem easier to debug and also reduce the time to failure if bad code is checked in.

````
Generating module docs from $CHPL_HOME/modules/ into rst/modules
(cd ../modules && make documentation)
mkdir -p standard
cd standard && /private/var/folders/69/jbk0y4q95lq8xhhwpqgwf_rh000bxs/T/chapel-release.bradc.deleteme.hlmkk/chapel-test/util/config/make_sys_basic_types.py --doc SysCTypes.chpl
Generating module documentation with chpldoc
export CHPLDOC_AUTHOR='Cray Inc' && \
[...]

PR #6256, opened 2017-05-15
Emit correct exit code for `make docs` failures

ben-albrecht:
This PR puts the docs error message in a separate target, and explicitly exits with error code 1. Note that `make` will exit with error code 2 anyway (as noted in the comment).

This eliminates a few symptoms that propagated up to `gen_release` and `testRelease`, which were pointed out in #6239.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -29,8 +29,8 @@ help-source:
 
 
 docs: FORCE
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Thanks @dmk42 for pointing me to this cleaner solution.

PR #6258, opened 2017-05-16
Add deinit to KNL and numa locale models

dmk42:
This change cleans up the numa and KNL locale models at program exit. The prior work by @bradcray for the flat locale model is in #6202 and #6238 .

First, correctness was verified by inserting printfs in the deinits and other strategic places to observe behavior on teardown.  This was done on Mac, Linux (with and without GASNet), and both KNL and non-KNL Cray for the numa locale model, and on both KNL and non-KNL Cray for the KNL locale model.  Testing for the KNL locale model had to be limited to places where the memkind library was available.

Then the printfs were removed and full paratesting was performed, on the same systems as above.

### DIFF
--- a/modules/internal/ChapelLocale.chpl
+++ b/modules/internal/ChapelLocale.chpl
@@ -665,6 +665,9 @@ module ChapelLocale {
   //
   // Free the original root locale when the program is being torn down
   //
+  // Be careful to free only origRootLocale, and never the copy in
+  // rootLocale, or the same locales will be torn down twice.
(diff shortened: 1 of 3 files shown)


dmk42:
@bradcray and @gbtitus , this is ready for review now.

I was only going to do the KNL locale model, but doing the numa one first helped me understand what to do better, so I ended up doing both.

gbtitus:
Looks good!  👍

PR #6261, opened 2017-05-17
Support bundling futures into another future via waitAll()

nspark:
This PR includes
* Implementation of `waitAll()`
* Documentation and tests for `waitAll()`
* Updates to the Future record to use atomic guard variables instead of sync vars (to support more result types)

Thanks @noakesmichael for the compiler support needed for this PR.


ronawho:
@mppf since you reviewed #5135, mind reviewing this one as well?

modules/packages/Futures.chpl
```diff
@@ -99,11 +112,13 @@ module Futures {
 
     type retType;
 
-    var value: single retType;
     var valid: bool = false;
+    var value: retType;
+    var state: atomic bool;
```
mppf: Re change from `single` to `atomic` and new field storing value; you could have switched to using a `single bool` instead of an atomic. Is there a reason to prefer the atomic? (I don't really have a preference; just trying to understand).

nspark: I think I understand the overhead of CPU atomics better than that of sync/single variables. So, it felt more like I was choosing to deal with a known quantity over an unknown one. I can see how a single variable would make sense here. If you think that'd be a better approach, I'm open to the idea of changing it.

mppf: I think atomic is fine and agree that there are fewer moving parts behind it.

mppf:
Full local testing passed. This is ready to merge as far as I'm concerned. Thanks!

PR #6265, opened 2017-05-18
throws for instantiations of generic functions

psahabu:
One of our users discovered a bug where generic functions could be marked `throws`, but their instantiations would not inherit that `throws` declaration. This PR ensures that `_throwsError` is carried over from the generic to any of its instantiations.


mppf:
Looks good!

PR #6267, opened 2017-05-19
Add some tests and futures for error-handling use-cases

nspark:
* Tests (which pass)
  * Throwing from a method
* Futures
  * Throwing from a generic function
  * Throwing from a method of a generic record
  * Functions that return, but may throw
  * Methods that return, but may throw
  * Throwing from a record's deinit()

Tagging @psahabu

### DIFF
--- a/test/errhandling/nspark/fn-returns-or-throws.bad
+++ b/test/errhandling/nspark/fn-returns-or-throws.bad
@@ -1,0 +1,3 @@
+before
+after
+called oops()
--- a/test/errhandling/nspark/fn-returns-or-throws.chpl
+++ b/test/errhandling/nspark/fn-returns-or-throws.chpl
@@ -1,0 +1,12 @@
+proc oops(): int throws {
+  throw new Error("called oops()");
+  return 99;
+}
+
+try {
+  writeln("before");
+  var y = oops();
+  writeln("after");
+} catch err {
+  writeln(err.msg);
+}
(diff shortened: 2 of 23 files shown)


nspark:
@psahabu I see now that #6265 should close out at least one of the futures (maybe two?) in this PR.

psahabu:
@nspark I noticed an interesting, unanticipated pattern in [generic-record-throws.chpl](https://github.com/nspark/chapel/blob/207db6a87cd0de9f627aba6e658c6844ac4d7b20/test/errhandling/nspark/generic-record-throws.chpl).

You throw `err` in `get()`, which is a field of the `Wrapper` record. The compiler's automatic error cleanup inserts a delete for that error after the `try`/`catch` because it isn't thrown on, which is probably unexpected behavior if the user were to access `w` afterward.

Do you think this is an important use case for Chapel to support, throwing errors stored in data structures while leaving them intact?

psahabu:
@mppf, @bradcray Here's the open issue that touches error memory management.

nspark:
> Do you think this is an important use case for Chapel to support, throwing errors stored in data structures while leaving them intact?

@psahabu I am inclined to say "no", however, the use case -- supporting errors in the Futures module -- that motivated this test code wants otherwise. The idea was that a Future would have both `value` and `error` fields; if the asynchronous function that provides the result to `value` were to throw an error, that error would get stored in `error`. A subsequent call to `Future.get()` would throw `error` (if there was one) or return the result. Neither field is intended to be accessed directly.

If, like std::future in C++, `Future.get()` could only be called once, then the compiler-inserted delete after the try-catch block would not cause an issue. However, I chose to make `Future` objects support multiple `get()` calls. If for some reason the first `get()` ignored the error, a subsequent `get()` call would throw an error that points to invalid memory.

Again, I'm inclined to say "no", assuming that distinguishing between the case for automatic cleanup and otherwise might be difficult to distinguish automatically.

psahabu:
see #6428

PR #6268, opened 2017-05-19
Fix typos in UseStmt.cpp

cassella:
Fix typos in UseStmt.cpp


cassella:
These fixes have been overtaken by events.

PR #6276, opened 2017-05-19
Add multilocale performance testing for NPB MG

lydia-duncan:
Adds a 4 node multilocale run for MG size B and adds a graph to track its time
to the general time suite and the NPB time suite.  Note that we get
verification failures for runs with 6, 8 and 16 locales (and I didn't check 5)


test/users/npadmana/npb-mg/mg.ml-compopts
```diff
@@ -0,0 +1 @@
+-suseBulkTransferStride
```
benharsh: I don't think you need to specify this anymore, it should be on by default.

lydia-duncan: Okay.  Should I remove it from the perfcompopts file while I'm at it?

benharsh: Yeah, that sounds good

PR #6278, opened 2017-05-20
Update compiler source for newer llvm

mppf:
This PR uses #ifdefs to enable the Chapel compiler to build with LLVM versions 3.8, 3.9, and 4.0.

I verified that Hello World works for all of these when built with --llvm (but for LLVM 4.0, I had to turn off TBAA metadata because only struct-path TBAA is now allowed).

### Testing & review

- full local testing
- full local --llvm testing

Reviewed by @dmk42 - thanks!

### Notes about future work:

I hope we can upgrade the bundled version to 4.0.

If we upgrade to LLVM 4.0, we will need to start emitting struct-path TBAA (the old-style alias analysis metadata we were generating is no longer supported).

Other changes will be required to Makefiles in third-party/llvm when we change the version bundled with Chapel. If it changes to version 4.0:
  installhdrs -> install-llvm-headers, install-clang-headers
  llvm-config command now needs: passes coroutines lto

Some notes about LLVM versions that I learned in the process:
 * 3.8 is the last release supporting an autoconf-based build system.
 * 3.9.1 requires cmake >= 3.4.3 but includes clang API improvements that might allow us to no longer need an internal clang header.
 * 4.0 includes coroutines (which we could use to implement iterators) and more API improvements to clang that will help us avoid needing the internal header. This version also requires GCC 4.8 to build (if using GCC).

We plan to upgrade to LLVM 4.0 once we update the compiler to emit struct-path TBAA. Then, at some point, we will remove the #ifdef support for older LLVMs.


compiler/include/llvmUtil.h
```diff
@@ -26,7 +26,13 @@
...
+#if   LLVM_VERSION_MAJOR>4 || (LLVM_VERSION_MAJOR==4 && LLVM_VERSION_MINOR>=0 )
+#define HAVE_LLVM_VER 40
+#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=9 )
+#define HAVE_LLVM_VER 39
+#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=8 )
+#define HAVE_LLVM_VER 38
+#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=7 )
 #define HAVE_LLVM_VER 37
 #elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=6 )
 #define HAVE_LLVM_VER 36
```
mppf: I don't think it's likely that LLVM will use a .10 version any time soon, since they just switched to whole-number-increments every release.

At the same time though, we could make the macro better. In the future I hope to combat its growth by only supporting LLVM 4.0 (say).

Nonetheless I'll add a comment with your suggestion.

Issue #6275, opened 2017-05-19
Cannot access module-scope function from shadowed record-scope method


ben-albrecht:
### Summary of Problem

Due to scoping rules in Chapel, the only way to access a function or variable
that shares the same name as a method or field in a record/class is to specify the intended scope with the module prefix. The future linked below demonstrates that this does not work for
methods shadowing functions.

Confusingly, the compiler emits the error:

    use of 'A' before encountering its definition, type unknown

despite the A module being accessible elsewhere within the method.

For what it's worth, the shadowed variable/field case works as expected.

### Steps to Reproduce

**Associated Future Test(s):**
[`test/visibility/shadow.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/visibility/shadow.chpl
          ) #6274

lydia-duncan:
I believe this is because we are recognizing that the function is a method on a class in the closest scope with that function defined, and so think that A could be the name of a local instance of that type.  However, A is a module, not an instance, so it isn't valid to look for it as if it is an instance.

noakesmichael:
This issue was resolved by PR 6279.  The future created in PR #6274 was retired.

PR #6279, opened 2017-05-20
Resolve issue 6275

noakesmichael:
This PR resolves recent issue #6275; an error in which a reference to a module-level function
within a method is incorrectly shadowed by a method with the same name as the function.

Resolution currently relies on convoluted conventions to manage scopes for class/record fields,
method calls, and function calls via explicit module names.  This PR adds an additional work-around
to prevent methods from unintentionally shadowing attempts to access a function defined in a
named module.

This could have been a "one line" change but this PR takes the opportunity to perform a
modest refactoring of the existing business logic and add some comments so that the
implementation is marginally less obscure.

This PR also retires the future that was created for the issue.

Limited local testing on both platforms with both configurations.
Single-locale paratest with -futures


bradcray:
Speedy!  Thanks!

PR #6280, opened 2017-05-20
Noinline attribute when using llvm-print-ir

coodie:
In full optimization stage function sometimes was inlined and removed from llvm module, easiest way to fix this is to add 'noinline' attribute for it.

The preferred way however is to add noinline pragma before function we want to print so that this option doesn't interfere with other optimization options and it is explicitly stated in code.

Other solution would be add _llvm-print-inline_ option which would specify if we want to turn this off or on.

Anyway, now we can see that adding noinline attribute is very easy to do.

### DIFF
--- a/compiler/codegen/symbol.cpp
+++ b/compiler/codegen/symbol.cpp
@@ -45,6 +45,7 @@
 #include "stringutil.h"
 #include "type.h"
 #include "resolution.h"
+#include "llvm/IR/Attributes.h"
 
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


compiler/codegen/symbol.cpp
```diff
@@ -56,6 +54,10 @@
 #include <inttypes.h>
 #include <stdint.h>
 
+#ifdef HAVE_LLVM
+#include "llvm/IR/Attributes.h"
+#endif
```
coodie: I don't like putting these ifdefs here, but I can't see the default place where to put these ifdefs.

mppf: This is OK. I'd probably put the LLVM include before the first system include (`<algorithm>`). compiler/include/llvmUtil.h might work too but it's probably smarter to just put the include you need for this file here. Speaking of that, I'm surprised that the other llvm:: stuff in this file worked addFnAttr didn't. Are you sure you can't just remove the new include entirely?

coodie: Well, apparently it does, I can recall that it didn't compile without this header but apparently my crazy mind made it up.

mppf:
Passed full local testing.

PR #6284, opened 2017-05-22
Fix NPB MG

benharsh:
This program would fail when run with more than 4 locales. The cause
turned out to be some incorrect math in the 'stencilConvolve' helper.
This core operation is written with a temporary array in the reference
version, but we don't have task-private arrays yet so we wrote it
differently for Chapel. It seems like one or more aspects of the
computation were translated incorrectly in the original version of this
code.

After these changes, we can run with 8 and 16 locales without failures.
Performance for 4 nodes seems to be roughly the same as it was before
this change.

I also changed the coefficient type from an array to a tuple, and made
some minor whitespace changes.

Also increased number of locales for performance testing now that the correctness issue is fixed.


benharsh:
@lydia-duncan : here are the changes we discussed offline.

@npadmana : As the original author of this program, do you have any objections to these changes?

lydia-duncan:
This looks reasonable to me, thanks for doing it!

npadmana:
@benharsh -- absolutely no problem --- I remember writing this code, but that's about it! I'm a little sheepish about a math bug though; just out of curiosity, where was the bug?

benharsh:
I didn't nail down a particular line or anything, but I think the issue was in stencilConvolve where we were modifying the previous/next cell. I think there were issues on the edges as well.

npadmana:
ok --- thanks for getting it to work!

PR #6289, opened 2017-05-23
Rename ReplicatedDist (the class) to Replicated

bradcray:
This is nicer for the user's dmapped clauses and more symmetric
to what we do for Block, Cyclic, and other distributions.

TODO:
- [x] linux64 testing
- [x] gasnet testing
- [x] distribution robustness testing for replicated
- [ ] add deprecated version of ReplicatedDist

### DIFF
--- a/modules/dists/DimensionalDist2D.chpl
+++ b/modules/dists/DimensionalDist2D.chpl
@@ -135,7 +135,7 @@ and distributes using block-cyclic distribution in the second dimension.
     for loc in MyLocales do on loc {
 
       // The ReplicatedDim specifier always accesses the local replicand.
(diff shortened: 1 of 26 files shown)


bradcray:
I'm having trouble finding a way of deprecating ReplicatedDist short of cloning all of the 1.15 ReplicatedDist code into this file.  Once we change over to initializers for the BaseDist class hierarchy, I think we should be able to do this more directly.  My current thought is to file a placeholder issue against that and do the renaming now.  Alternatively, I could add the 1.15 ReplicatedDist into the file and replace it with the initializer-based approach when we switch.

bradcray:
Replaced this with #6327

PR #6294, opened 2017-05-23
Add test to demonstrate subtle behavior for shadowed methods

ben-albrecht:
There was a discussion about what the behavior of the following program should be:

```chapel

proc shadow(x) {
  writeln('outer scope');
}

record R {

  proc shadow() {
    writeln('R scope');
    shadow(10);
  }

}


var r = new R();
r.shadow();
```

The conclusion was that the call to `shadow()` without a *module-access-expression* is invalid in Chapel. This follows what other languages that also support implicit *member-access-expression* do, including: C++, D, and Swift

This PR adds a test to ensure that we don't break this subtle behavior in the future.

### DIFF
--- a/test/visibility/shadow-overload.chpl
+++ b/test/visibility/shadow-overload.chpl
@@ -1,0 +1,21 @@
+/* Test to ensure that functions shadowed by methods
+   *cannot* be called without the module-access-expression, even when the
+   function and method have different signatures */
+
+proc shadow(x) {
+  writeln('outer scope');
+}
+
+record R {
+
+  proc shadow() {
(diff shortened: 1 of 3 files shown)


lydia-duncan:
We've ensured that the fix didn't break valid recursive access to the method, correct?

ben-albrecht:
For clarity, I'll provide some additional context:

>We've ensured that the fix didn't break valid recursive access to the method, correct?

I think this comment is referring to #6279 and being addressed to @noakesmichael 

As said offline, #6279 had no effect on the test being added in this PR.

Issue #5787, opened 2017-03-22
~ on bools is confusing


mppf:
``` chapel
  var t:bool(16) = true;
  var f:bool(16) = false;
  proc returnComplement(arg) {
    return ~arg;
  }
  writeln(returnComplement(t));
  writeln(returnComplement(f));
```

currently prints out
```
true
true
```

when one might expect it to print out
```
false
true
```

Related tests:
 *  types/scalar/bradc/bitNegateBool
 * types/scalar/bradc/bools2

Some discussion on this topic revealed these options:

1) Make bitwise operations on bool types (including `|` `&` `~`) result in compilation error.
 * programmers shouldn't be thinking of bools in terms of bit patterns / it's not what the bitwise operators are for
 * can lead to confusion if the element type ever changes to integer
 * could be confusing if coercion is involved

2) Make bitwise operations on bool types operate with 1-bit at a time; so `~` is equivalent to `!` and the and and or operators are equivalent to the logical versions, but without short circuiting.
 * arguably supports generic programming
 * these operations on bool values are unambiguous, so why not support them
 * traditionally, in C-like languages, `|` and `&` are available as non-short-circuiting equivalents to `||` and `&&`. If Chapel users care about this distinction, they might like to have both.

Related question: should it be possible to apply `||` `&&` or `!` to integral types?

mppf:
Note, currently `!` is defined on integers, but `||` and `&&` are not.

mppf:
Note, --llvm behaves differently than the C backend for some of these tests.

vasslitvinov:
I support keeping integers and booleans separate: no bitwise operators on booleans and no logical operators on integers.

gbtitus:
With respect to:

> ... traditionally, in C-like languages, | and & are available as non-short-circuiting equivalents to || and &&. If Chapel users care about this distinction, they might like to have both.

`|` and `&` are only "non-short-circuiting equivalents" for `||` and `&&` when programmers are very sloppy or values are very constrained.  `1 & 2` is 0, which C treats as logical false.  `1 && 2` is logical true.

tmacd9:
Just to follow up on @gbtitus comments - if I write:

   var x : bool = (i < n) & (i > 0);

you are recommending this should be written as:

   var x : bool = (i < n) && (i > 0);

This could force someone to write:

   var x = f();
   var z = g() && x;

if they want both functions to be called instead of:

   var z = f() & g();

gbtitus:
[...]

PR #6300, opened 2017-05-24
Make ~ on bool an error

mppf:
See issue #5787, which discusses the problem with `~` on bools and the strategy for improvement.

If we want to continue to support `~` on bools, the implementation would need to change (possibly to calling `!`). The previous implementation had different behavior for --llvm and when compliing with GCC 7.

Note that this PR adds a `where true` clause to the error overload of `~`. At first I tried to mark it with `pragma "compiler generated"` to make it a last-resort candidate. However, if I do that, the other `~` overload will apply. (I.e. `~` on bool will coerce the argument to int, then run bit complement). Since the disambiguation rules prefer a match without a `where` clause to one with a `where` clause, I used `where true` to achieve the same effect (mostly). I've also added a test that checks that user-defined overload of `~` is possible.

Passed full local testing.

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -441,7 +441,7 @@ module ChapelBase {
   //
   // bitwise operations on primitive types
   //
-  inline proc ~(a: bool) return __primitive("u~", a);
+  inline proc ~(a: bool) { compilerError("~ is not suported on bool"); }
   inline proc ~(a: int(?w)) return __primitive("u~", a);
   inline proc ~(a: uint(?w)) return __primitive("u~", a);
 
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -441,9 +441,10 @@ module ChapelBase {
   //
   // bitwise operations on primitive types
   //
-  inline proc ~(a: bool) { compilerError("~ is not suported on bool"); }
   inline proc ~(a: int(?w)) return __primitive("u~", a);
   inline proc ~(a: uint(?w)) return __primitive("u~", a);
+  // note: where clause used here to enable a user-defined overload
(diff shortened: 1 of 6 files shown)


modules/internal/ChapelBase.chpl
```diff
@@ -441,7 +441,7 @@ module ChapelBase {
   //
   // bitwise operations on primitive types
   //
-  inline proc ~(a: bool) return __primitive("u~", a);
+  inline proc ~(a: bool) { compilerError("~ is not suported on bool"); }
```
bradcray: I think that this should have the "compiler generated" ("last resort") pragma applied to it so that a user or package module could choose to provide its own overload if it didn't like our choice.

In other similar error messages, we tend to use phrases like "...on operands of type bool" or "...on operands of boolean type" which I find more intuitive than simply "...on bool" (which, taken literally, suggests to me that perhaps the operation is not supported on the ``bool`` type itself?).  If keeping the message succinct seems important, even saying "on bools" or "on boolean operands" seems clearer.

mppf: I can't make it "compiler generated" because if I do, the following line's `~` overload will apply. (I.e. `~` on `bool` will coerce the argument to int, then run bit complement). Adding an unnecessary `where` clause covers much the same territory, so I'm doing that. I've also added a test that checks that user-defined overload is possible.

bradcray: On the phone just now we were talking about how the where clause didn't do the right thing as expected.  I forgot to ask the obvious follow-up question:  Is there some other (not-too-contorted) solution available to us that you can see?  In any case, I think the test you wrote should be updated (or cloned) to put the user overload into a sibling "library" module which I suspect would demonstrate the unintended behavior?

PR #6302, opened 2017-05-24
Add hand-optimized version of PRK Stencil to ML performance testing

benharsh:
This hand-optimized version uses localAccess calls and the local block to squash overhead. It also sets the boundingBox of 'output' to be the same as 'input' so that there is zero communication in the core stencil loop.

I removed the BlockDist and tiling options to limit the scope of this study (at least for now).


benharsh:
@ronawho : Here's the hand-optimized version I mentioned a while ago. This version is roughly 2x better than the elegant version, but is still 2x slower than the reference. The remaining performance likely involves:
- misc LICM changes
- NUMA first-touch
- buffered bulk-transfer (only a little)

@ben-albrecht : pinging you as the original author of the stencil PRK.

test/studies/prk/stencil/optimized/SKIPIF
```diff
@@ -0,0 +1,2 @@
+CHPL_TEST_PERF != on
+CHPL_TEST_PERF_LABEL != ml-
```
ronawho: Any reason to not run this as a correctness and/or a single locale performance test as well?

ronawho:
👍

PR #6306, opened 2017-05-24
Improve return intent overloads

### LINKED ISSUE #5656: return intent overload and ambiguity
What happens today if you only define the const ref and value overloads?  An ambiguity?  Or does it just always pick a certain one of them?

After PR #5561, the compiler always chooses the value one. Before that, it would have been an ambiguity error. Note that this problem existed for duplicate ref versions starting with the first ref-pair support in the compiler.

As far as what should happen, I think it should either be:

1) choose between value and const ref version depending on how the expression is used
2) make it a failure to resolve due to ambiguity

There are some other issues with ambiguity & ref pairs that stem from the fact that disambiguateByMatch returns NULL for both 0 matches and also for ambiguity. That means that e.g. declaring 2 ambiguous ref-return functions will disable ref-pair functionality rather than raise an error.

It seems clear to me that duplicate `ref` return versions should create a failure to resolve rather than silently preventing the usual ref-pair behavior. That's a bug, but we have to choose between 1 and 2 above for the value/const ref variant.

mppf:
* allows value and const-ref overloads
* more carefully handles the difference between ambiguity and
  no candidates when choosing return intent overloads.

See issue #5656 for the relevant language design discussion.

Closes #5656.

- [x] full local testing
- [x] spec changes: const ref + value overloading possible (PR #6435)
- [x] spec changes: where ignored for finding first set of most specific (PR #6435)

### DIFF
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -1394,6 +1394,9 @@ bool CallExpr::isRefExternStarTuple(Symbol* formal, Expr* actual) const {
 
 ContextCallExpr::ContextCallExpr() :
   Expr(E_ContextCallExpr),
+  hasValue(false),
+  hasConstRef(false),
+  hasRef(false),
   options()
 {
   options.parent = this;
@@ -1404,6 +1407,9 @@ ContextCallExpr*
 ContextCallExpr::copyInner(SymbolMap* map) {
   ContextCallExpr* _this = 0;
   _this = new ContextCallExpr();
+  _this->hasValue = hasValue;
+  _this->hasConstRef = hasConstRef;
+  _this->hasRef = hasRef;
   for_alist(expr, options)
     _this->options.insertAtTail(COPY_INT(expr));
   return _this;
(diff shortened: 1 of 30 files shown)


mppf:
[Error matching compiler output for modules/standard/Reflection/reflectOnTypesParams]
[Error matching compiler output for release/examples/benchmarks/miniMD/miniMD (compopts: 1)]
[Error matching compiler output for release/examples/benchmarks/miniMD/miniMD (compopts: 2)]
[Error matching compiler output for release/examples/benchmarks/miniMD/explicit/miniMD]

The miniMD one at least seems to have 5 overloads for `this` and only some of them seem to have the right argument types. Maybe a problem in candidate filtering? Or I am missing a coercion from array to tuple and from range to int?

chpl test/release/examples/benchmarks/miniMD/miniMD.chpl  -M test/release/examples/benchmarks/miniMD/helpers/ --gdb

mppf:
I have a fix for that problem, but now I'm running into a problem where the implementation doesn't match the specification. The spec says in 13.13
> From the set of candidate functions, determine the set of most specific functions. In most cases, there is one most specific function, but there can be several if they differ only in return intent. The set of most specific functions is the set of functions that are not more specific than each other but that are more specific than every other candidate function. The more specific relationship is defined in §13.13.3.

A function with a `where` clause is less specific than one without. But, we currently have overloads in the array implementation where the `ref` version does not have a `where` clause but the others do.

compiler/AST/expr.cpp
```diff
@@ -1473,74 +1479,60 @@ Expr* ContextCallExpr::getFirstExpr() {
...
-    // Note: it would be nicer to check retTag to decide between
-    // ref / value versions. However, doing so is challenging because
-    // of the way that iterator functions no longer have the original
-    // retTag.
-    constRefCall = toCallExpr(options.get(1));
-    valueCall    = toCallExpr(options.get(2));
-    refCall      = toCallExpr(options.get(3));
+  // always use order of value, const ref, ref
 
+  int n = 1;
```
vasslitvinov: Makes me wish for a more direct representation of the three calls.
OK to leave as-is.

vasslitvinov:
If there is something non-trivial about `ignoreWhere` esp. why you added it, please add a comment about it to the code.

mppf:
@vasslitvinov - I talked about the reason for ignoreWhere in the comment above "I have a fix...".

vasslitvinov:
@mppf  - Would you copy the relevant pieces of that comment to where disambiguateByMatch* are defined?

compiler/resolution/cullOverReferences.cpp
```diff
@@ -1753,11 +1761,75 @@ bool firstPassLowerContextCall(ContextCallExpr* cc)
...
+void lowerContextCallPreferRefConstRef(ContextCallExpr* cc)
+{
+  CallExpr* refCall = NULL;
+  CallExpr* valueCall = NULL;
+  CallExpr* constRefCall = NULL;
+  choose_type_t which = USE_REF;
+
+  cc->getCalls(refCall, valueCall, constRefCall);
+
+  if (refCall) which = USE_REF;
```
vasslitvinov: `which` is already initialized to `USE_REF`. Why do it again here?

compiler/resolution/cullOverReferences.cpp
```diff
@@ -1753,11 +1761,75 @@ bool firstPassLowerContextCall(ContextCallExpr* cc)
...
+{
+  CallExpr* refCall = NULL;
+  CallExpr* valueCall = NULL;
+  CallExpr* constRefCall = NULL;
+  choose_type_t which = USE_CONST_REF;
+
+  cc->getCalls(refCall, valueCall, constRefCall);
+
+  if(constRefCall) which = USE_CONST_REF;
+  else if(valueCall) which = USE_VALUE;
```
vasslitvinov: Analogously to the above:
`which` is already initialized to `USE_CONST_REF`. Why do it again here?

compiler/AST/expr.cpp
vasslitvinov: OTOH keeping all three in a list makes their removal cheaper.

mppf: You mean having fields for each call, say? I think I was just continuing to use an AList since that was the path of least resistance when I first introduced ContextCallExpr.

vasslitvinov:
Other than my comments, this is good to merge.

compiler/AST/expr.cpp
vasslitvinov: That's fine for now, you encapsulated them nicely.

PR #6307, opened 2017-05-24
move /data/sea/chapel/Nightly* to /data/cf/chapel/Nightly* (except NightlyHistory)

awallace-cray:
This commit is basically a mass grep-and-replace for /data/sea/chapel,
omitting some hits that are nothing to do with the target subdirectories.


awallace-cray:
Must be coordinated with PRs in chapel, chapel-code, and chapel-ci-config, copying or moving the actual data files, etc. CHAPDEV-229.

Issue #6192, opened 2017-05-08
Improve mpi+qthreads interoperability 


ronawho:
Using `CHPL_TASKS=qthreads` with mpi can and does lead to deadlock. As a workaround we're currently forced to use fifo, which has significantly worse performance. We need to figure out a way to get mpi+qthreads interop working. 

Related to https://github.com/chapel-lang/chapel/issues/5722 and https://github.com/chapel-lang/chapel/pull/4193

PR #4193, opened 2016-07-19
Highlight the deadlock that can occur between qthreads and MPI

npadmana:
Motivated by an email exchange with @bradcray, here is a simple standalone case that exposes the deadlock when running MPI and Qthreads. 

Note that this occurs even if Chapel is running single locale, but with the qthreads tasking layer. 

The README file lists the different cases I've considered...

PR #6314, opened 2017-05-25
MPI + Qthreads

npadmana:
A work-around for MPI+qthreads support.

1. Use non-blocking calls whenever possible. We provide wrappers for a small subset of common blocking calls. Most important is an implementation of ``Wait``, which wraps ``MPI_Test`` and calls ``chpl_task_yield`` if a request is not complete.
2. Blocking calls must be preceded by a call to ``Barrier``.
3. Blocking calls should be serialized by the user.

The hello-world test case has been updated to avoid a deadlock using the above. We also include a ring test case that previously used to deadlock (used to be #4193).

With contributions from @ronawho & @ben-albrecht 

Fixes issues in #5722 and #6192, and retires #4193.

Tested by @npadmana using -- 
```
CHPL_TARGET_PLATFORM: cray-xc
CHPL_TARGET_COMPILER: cray-prgenv-intel
CHPL_TARGET_ARCH: haswell
CHPL_LOCALE_MODEL: flat
CHPL_COMM: ugni
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
  CHPL_JEMALLOC: jemalloc
CHPL_MAKE: gmake
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: ugni
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```

@ronawho, @ben-albrecht -- here is a start to getting qthreads+MPI support.  The idea here are drop-in replacements to some blocking calls. Let me know what you think....

For this version, I suggest supporting a subset of all blocking calls. This is a minimal list to get some basic tests in (ring and hello-world), just to get the basic testing going. We can expand this list over time
- [x] Wait
- [x] Barrier
- [x] Send
- [x] Recv

TODO:
- [x] Document each of these routines (simple docs are in)
- [x] Document usage of MPI+qthreads
- [x] Multilocale hello-world should run with gasnet/ugni + qthreads. 
- [x] Add in ring test
- [x] Fill out PR message.


ronawho:
👍  This is almost exactly what I was thinking about, and the implementation looks good to me.

Only difference is that I had been planning to replace the MPI_Send/MPI_Recv/MPI_Wait procs instead of adding wrappers (e.g. implement MPI_Send ourselves, instead of extern'ing to the "real" ones.) If we expect users to be calling the MPI_ routines directly, I think that's probably the right approach, but if those are just there to expose the API and the higher level MPI module is the one you want/expect users to use, this looks good to me.

For reference, here's what I had been working on: https://github.com/ronawho/chapel/commit/e2e7be0. And with that a

npadmana:
Eliot -- oops! Didn't mean to duplicate work.....

Implementing these natively in Chapel is an interesting idea, and probably is worth keeping on a possible todo list. My thought here was simply expediency --- this would let us do a chunk of these quickly and they'd just be drop in replacements. At the end of the day, I'd hope that the user would use  more idiomatic Chapel constructs instead of writing MPI code --- and that this module would just exist to deal with third-party libraries....

Just for completeness and my reference, some other blocking calls I was thinking of wrapping --
Bcast, Reduce, Allreduce, Alltoall

ronawho:
>Eliot -- oops! Didn't mean to duplicate work.....

No problem at all, wasn't a lot of work, and you're obviously much faster than I am :)

>Implementing these natively in Chapel is an interesting idea, and probably is worth keeping on a possible todo list. My thought here was simply expediency --- this would let us do a chunk of these quickly and they'd just be drop in replacements. At the end of the day, I'd hope that the user would use more idiomatic Chapel constructs instead of writing MPI code --- and that this module would just exist to deal with third-party libraries....

Yeah, I agree with all this. I certainly don't want to implement all of them ourselves, at least for now just the ones that easy to convert from blocking to non-blocking routines.

>Just for completeness and my reference, some other blocking calls I was thinking of wrapping --
Bcast, Reduce, Allreduce, Alltoall

Ok good to know. In the commit, I just marked routines that have an `I` version, where it'd be trivial to implement a non-blocking version ourselves. It's not obvious to me how we'd implement Bcast/reduce/etc ourselves with the 1.1 API. We'd need mpi-3 for native non-blocking versions of  Bcast, Reduce, Allreduce, and Alltoall, right?

npadmana:
Yes, some of these will need to pull in newer versions. I've already had to do it for Ibarrier.... 

My feeling here is not to be too finicky about API versions and pull in the newer calls we need. That being said, I didn't want to wrap all of MPI-3 because that was a nightmare in terms of the number of functions.

ronawho:
Ah ok, that makes sense to me

test/modules/packages/mpi/multilocale/hello-chapel-nl-1/hello.chpl
```diff
@@ -7,6 +7,17 @@ coforall loc in Locales do on loc {
...
     if irank == rank then
       writef("Hello, Chapel! This is MPI rank=%i of size=%i, on locale.id=%i\n",rank, size, here.id);
+
+    /* Why the double barrier?
+
+       A single barrier is to ensure that the code outputs in the correct sequence, and
+       doesn't overwrite. This could be achieved with a single Barrier.
+
+       Having a direct call into the MPI_Barrier also tests that the first Barrier prevents
+       deadlocks when running CHPL_TASKS=mpi. This second MPI_Barrier call stands in for any blocking
```
ronawho: `CHPL_TASKS=mpi` => `CHPL_TASKS=qthreads` or `CHPL_COMM_SUBSTRATE=mpi`?

npadmana: qthreads... I'm was being an idiot.

npadmana:
@ronawho, @ben-albrecht -- ok, modulo documentation, this is "done" and ready for review. It should work on ugni/gasnet(aries) + qthreads....

modules/packages/MPI.chpl
```diff
@@ -333,6 +353,63 @@ module MPI {
...
+    Wait(request);
+    return ret;
+  }
+
+  /* Drop in replacement for ``MPI_Recv``, with non-blocking MPI calls.
+
+     This is implemented by a call to ``MPI_Irecv``, followed by a call to ``Wait`` above. The
+     returned value of ``MPI_Status`` is ignored.
+   */
+  proc Recv(ref buf, count: c_int, datatype: MPI_Datatype, dest: c_int, tag: c_int, comm: MPI_Comm, ref status: MPI_Status): c_int {
```
ronawho: `dest` => `source`?

modules/packages/MPI.chpl
```diff
@@ -333,6 +353,63 @@ module MPI {
...
+    ret = C_MPI.MPI_Test(request, flag, status);
+    while (flag==0) {
+      chpl_task_yield();
+      ret = C_MPI.MPI_Test(request, flag, status);
+    }
+    return ret;
+  }
+
+  /* Overloaded version of Wait, which ignores the returned status */
+  inline proc Wait(ref request): c_int {
```
ronawho: `ref request` => `ref request: MPI_Request` for consistency?

test/modules/packages/mpi/multilocale/hello-chapel-nl-1/hello.chpl
```diff
@@ -7,6 +7,17 @@ coforall loc in Locales do on loc {
...
     if irank == rank then
       writef("Hello, Chapel! This is MPI rank=%i of size=%i, on locale.id=%i\n",rank, size, here.id);
+
+    /* Why the double barrier?
+
+       A single barrier is to ensure that the code outputs in the correct sequence, and
+       doesn't overwrite. This could be achieved with a single Barrier.
+
+       Having a direct call into the MPI_Barrier also tests that the first Barrier prevents
+       deadlocks when running with CHPL_TASKS=qthreads. This second MPI_Barrier call stands in for any blocking
```
ronawho: Inconsistent line width in these comments looks a little weird, but no big deal if you don't feel like fixing it.

test/modules/packages/mpi/multilocale/ring/ring.chpl
```diff
@@ -0,0 +1,40 @@
...
+        size = commSize(CHPL_COMM_WORLD) : c_int;
+    var src, val : c_int;
+    src = mod(rank-1,size);
+
+    writef("Rank %i receiving from %i \n",rank, src);
+    var status : MPI_Status;
+    Recv(val, 1, MPI_INT, src, 0, CHPL_COMM_WORLD, status);
+    writef("Rank %i received %i from the left.\n", rank, val);
+  }
+}
```
ronawho: Add a newline to appease github?

test/modules/packages/mpi/multilocale/ring/ring.good
```diff
@@ -0,0 +1,18 @@
...
+Rank 2 done sending...
+Rank 2 received 1 from the left.
+Rank 2 receiving from 1 
+Rank 2 sending to 3 
+Rank 3 done sending...
+Rank 3 received 2 from the left.
+Rank 3 receiving from 2 
+Rank 3 sending to 0 
+The main program ends here...
+This is the main program
```
ronawho: We don't need a CHPL_COMM=none .good file, because there's a multilocale.skipif that ensures this won't run in that case, right?

ronawho:
👍 No major comments, just some nits. I'm happy with the documentation

test/modules/packages/mpi/multilocale/ring/ring.good
npadmana: I believe so, but I'll let @ben-albrecht confirm....

test/modules/packages/mpi/multilocale/ring/ring.chpl
npadmana: hm. This is commit 915efa2, but it doesn't seem to be showing up.... ?

test/modules/packages/mpi/multilocale/ring/ring.good
ben-albrecht: Correct!

ben-albrecht:
I'll test before we merge:

- [ ] Cray
- [ ] linux64

ben-albrecht:
Do we want to remove the `fifo`-only constraint in `mpi/multilocale.skipif` with this PR, and enable MPI testing on the Crays? Or would you prefer this to happen in a separate PR?

npadmana:
We may as well remove that constraint here.... --- seems like the point of this PR. :-)

Which reminds me --- we should also update some of the module level docs to remove the fifo constraint....

ben-albrecht:
> we should also update some of the module level docs to remove the fifo constraint

@npadmana - I'd do that in this PR too, if you're cool with it.

npadmana:
@ben-albrecht -- done. As an added bonus, I made sure the docs compiled and weren't completely messed up. ;-)

PR #6317, opened 2017-05-25
Make where clause for rank change operations more specific

mppf:
I noticed that the rank change overload for `this` was participating in overload resolution for array slices. E.g. for a 1-D array, A[1..10]. I can't explain why it worked before (perhaps a
bug in overload resolution)?

This commit fixes that by checking that there is both a range argument and a non-range argument in _validRankChangeArgs.

Passed full local testing.


daviditen:
This looks good to me.

PR #6326, opened 2017-05-26
Support generic records created via "new" expressions

lydia-duncan:
This change adds support for initializing generic records that define
initializers via "new" expressions (`var foo = new MyRec();`).  It does
not yet add support for initializing such records solely via explicit type
declarations (`var foo: MyRec(args);`) and there is a separate issue
with initializers on generic records having default values for their fields,
but I figured I should get the basic case in first.

The change itself is fairly straight-forward: in a few places, extend the
handling of non-generic records with initializers to also handle generic
records with initializers; add a couple of helper methods to aid in this
action; ensure that the argument passed in for the "this" arg to the
initializer is a named argument, so we can match against it easily; ensured
that we appropriately updated the type of the "this" arg actual once we
knew its concrete form, and fix function resolution's handling of generic
arguments that should be turned into ref types to mark the formal with
QUAL_REF when we haven't created the `_ref(Foo)` wrapper type
(because we can't create that wrapper type when Foo is still generic,
which it will be until we've resolved Phase 1 of the initializer body).  I also
removed the general generic records with initializers error message and
added one specifically for when the record is initialized with an explicit type
declaration.

Shout-out to Mike for making the handling of record initialization much
cleaner, this change would not have been so simple without that.

Testing:
Added a copy of most of my generic class tests where the only difference is
that the type is now a record.  My generic record tests were relying on
explicit type declarations and so didn't give sufficient coverage for me to be
certain my change was correct for the "new" expression case without these
tests.  Adds 14 new passing tests, 4 new futures to track default values for
arguments in generic record initializers and the "this." issue for param fields,
and resolves 2 passing futures in the test/classes/initializers/records/generics
directory.  Updates the .bad file for
test/studies/shootout/chameneos/bradc/chameneos-blc-ideal.chpl, as it
progresses further but now fails because expressions of the form
`[i in 1..10] i` are stored as iterator expressions in the field rather than getting
transformed into arrays (ok'd by Brad).  Updated the source for
test/types/records/ferguson/copy-init/new-init-generic.chpl to avoid the
ambiguity.

Passed full std testing, and no additional passing futures


lydia-duncan:
@noakesmichael - I recommend looking at the final state of the PR rather than going through all the commits, there was some clean up and rearranging that occurred

compiler/passes/normalize.cpp
```diff
@@ -1862,7 +1870,9 @@ static void normVarTypeInference(DefExpr* defExpr) {
     if (initCall->isPrimitive(PRIM_NEW) == true) {
       AggregateType* type = typeForNewExpr(initCall);
 
-      if (isNonGenericRecordWithInitializers(type) == true) {
+      if (type &&
```
noakesmichael: Why not isRecordWithInitializers(type)?

compiler/passes/normalize.cpp
```diff
@@ -1873,14 +1883,15 @@ static void normVarTypeInference(DefExpr* defExpr) {
         argExpr->baseExpr->replace(new UnresolvedSymExpr("init"));
 
         // Add _mt and _this (insert at head in reverse order)
-        argExpr->insertAtHead(var);
+        argExpr->insertAtHead(new NamedExpr("this", new SymExpr(var)));
```
noakesmichael: It troubles me that we are going back to the use of NamedExpr to signal things.
I don't understand this issue very well but it's hard for to understand why it's necessary.

compiler/passes/normalize.cpp
```diff
@@ -1989,7 +2000,8 @@ static void normVarTypeWithInit(DefExpr* defExpr) {
 
     var->type = type;
 
-  } else if (isNonGenericRecordWithInitializers(type) == true) {
+  } else if (isNonGenericRecordWithInitializers(type) == true ||
```
noakesmichael: Why not "isRecordWithInitiaizers(type) == true" ?

compiler/resolution/functionResolution.cpp
```diff
@@ -3285,6 +3289,11 @@ FnSymbol* resolveNormalCall(CallExpr* call, bool checkonly) {
         // If the first actual is an instance of dtMethodToken and the call is
         // to "init" of a generic record that defined initializers
         resolveInitializer(call);
+        NamedExpr* named = toNamedExpr(call->get(2));
```
noakesmichael: This seems unfortunate to me.

compiler/resolution/functionResolution.cpp
```diff
@@ -4078,7 +4087,8 @@ static void resolveInitVar(CallExpr* call) {
     call->primitive = primitives[PRIM_MOVE];
     resolveMove(call);
 
-  } else if (isNonGenericRecordWithInitializers(srcType) == true)  {
+  } else if (isNonGenericRecordWithInitializers(srcType) == true ||
```
noakesmichael: isRecordWithInitializers()

compiler/passes/normalize.cpp
lydia-duncan: Because I forgot it existed, I will fix that.

compiler/passes/normalize.cpp
lydia-duncan: The "this" argument is treated specially, since it starts as generic and its concrete instantiation cannot be known until Phase 1 of the initializer body has been resolved.  Using the NamedExpr on the actual allows us to update the type of the actual once we know its instantiation without relying on the more brittle "it's position 2 in the argument list" signaling.

compiler/resolution/functionResolution.cpp
lydia-duncan: I meant to move the action which updates the type of the "this" actual into resolveInitializer so it would be consistent for classes and records.  Would that solve your concern, or is the use of the NamedExpr at all still troubling?

compiler/passes/normalize.cpp
lydia-duncan: Would you prefer that I only add this for generic records?  I would be happy to perform that limitation, but I do believe it is necessary for them

lydia-duncan:
@noakesmichael - this is ready for a check to make sure things got cleaned up the way you wanted

compiler/passes/normalize.cpp
```diff
@@ -676,9 +676,10 @@ static Symbol* theDefinedSymbol(BaseAST* ast) {
         } else if (isPrimitiveScalar(type) == true) {
           retval = var;
 
-        // non generic records with initializers are defined
+        // records with initializers are defined
         } else if (AggregateType* at = toAggregateType(type)) {
-          if (isNonGenericRecordWithInitializers(at) == true) {
+          if (at->isRecord() == true &&
```
noakesmichael: isRecordWithInitializers()?

lydia-duncan: Crud.  Thanks!

PR #6327, opened 2017-05-26
Rename new 'ReplicatedDist' to 'Replicated' and preserve deprecated 'ReplicatedDist'

bradcray:
This mod renames the new implementation of the 'ReplicatedDist' class / domain map to 'Replicated' to avoid using the same name for both a module and its class, and to improve symmetry with other distributions like 'Block', 'Cyclic', etc.  In doing so, I resurrected the 1.15 implementation of 'ReplicatedDist' into its own class that generates a deprecation warning so that existing code would (a) continue to work as it used to and (b) get a deprecation warning about 'ReplicatedDist' with a pointer to 'Replicated'.

This approach has the downside of increasing the amount of code that is required for programs that 'use ReplicatedDist' (since they will effectively now get two domain maps rather than one), but I've done a quick check that this isn't used by default in our programs, so only those programs that use 'ReplicatedDist' should suffer (and we can remove 'ReplicatedDist' after 1.16 is released).

I put the old 'ReplicatedDist' into a new file 'OldReplicatedDist.chpl' as the simplest way to avoid having it generate documentation.  The only changes I made to the file were to add the deprecation warning by default (with a flag to turn it off in cases like dsiClone() and dsiPrivatize()) and to remove the tracing capability because it was difficult to refer back to the config param in ReplicatedDist.chpl since the class shares the same name.

TODO
- [x] spot-check new code against 1.15-era ReplicatedDist tests to make sure they still work
- [x] full linux64 testing
- [x] full gasnet testing
- [x] replicated distribution robustness


bradcray:
@vasslitvinov : Would you review this one?

modules/dists/DimensionalDist2D.chpl
```diff
@@ -135,7 +135,7 @@ and distributes using block-cyclic distribution in the second dimension.
     for loc in MyLocales do on loc {
 
       // The ReplicatedDim specifier always accesses the local replicand.
-      // (This differs from how the ReplicatedDist distribution works.)
+      // (This differs from how the Replicated distribution works.)
```
vasslitvinov: It doesn't differ now? Maybe that entire line should be deleted?

modules/dists/OldReplicatedDist.chpl
```diff
@@ -0,0 +1,652 @@
...
+
+
+// constructor: replicate over the given locales
+// (by default, over all locales)
+proc ReplicatedDist.ReplicatedDist(targetLocales: [] locale = Locales,
+                                   purposeMessage: string = "used to create a ReplicatedDist",
+                                   param warning = true)
+{
+  if warning then
+    compilerWarning("The `ReplicatedDist` domain map is deprecated.  Please switch over to using `Replicated` instead.");
```
vasslitvinov: Should we also warn the user that Replicated's behavior is different from ReplicatedDist's?

vasslitvinov:
Other than the minor comments, this looks good.

modules/dists/DimensionalDist2D.chpl
bradcray: I'm not sure, but will take your word for it.

modules/dists/OldReplicatedDist.chpl
bradcray: I thought about this, but opted not to for the sake of brevity.  Since we don't suggest that this is merely a renaming (i.e., we are telling them to use a different domain map), I feel like we're covered.  Are you OK with that?

vasslitvinov: I am OK with this. Although sometimes verbosity is a virtue, e.g. in hints sometimes given by git.

modules/dists/DimensionalDist2D.chpl
bradcray: Actually, now I feel more sure.  I hadn't thought much about what this was saying, nor about the fact that we'd already changed what ReplicatedDist does prior to this PR.

PR #6328, opened 2017-05-26
Fix perf graphs date parsing issue

benharsh:
The datejs library seems to cause problems with dygraphs under certain conditions. Use moment.js instead.


benharsh:
@ronawho : I verified that the datejs library was the issue by downloading one of dygraphs' x-axis zooming examples and including the datejs library in the html source. I would then see some different behavior from dygraphs, but not quite as bad as what we encountered.

moment.js seems to be a popular and more-recently-updated option.

PR #6332, opened 2017-05-26
Update PRK PIC inputs

ben-albrecht:
Update `GEOMETRIC` to use `rho=0.99` instead of `1.0`, which resulted in a `NaN` bug due to division by zero.  This matches the reference inputs.

Also updated `LINEAR` to use `alpha=1.0`, `beta=3.0` to match reference inputs.

I've added a comment denoting the `PATCH` reference inputs, but I believe they require a larger problem size, so I am leaving the correctness test size alone (we can use the reference values for performance tests).

Also, minimized a for-loop which wasn't using the iteration variable.


ben-albrecht:
@e-kayrakli -- may be of interest to you.

Issue #6330, opened 2017-05-26
Should iterator expressions stop being turned into arrays on assignment?


lydia-duncan:
Setting the value of a field in an initializer has managed to avoid the special case code which fires for ordinary field assignment that transforms `[i in 1..10] i` from an iterator expression into an array.  This is arguably a more "pure" treatment of the statement, but means that we are being inconsistent in our
handling of that expression, both for initializers and for passing the value of that field as an argument to another function (which is where the issue was noticed, see future test below).  If we do decide to extend this behavior to variable assignment (`var x = [i in 1..10] i;`), how would/should one express its type?

**Source Code:**
`[i in 1..10] i;`

**Compile command:**
`chpl foo.chpl`

**Associated Future Test(s):**
[`test/studies/shootout/chameneos/bradc/chameneos-blc-ideal.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/studies/shootout/chameneos/bradc/chameneos-blc-ideal.chpl)

### Configuration Information

- Output of `chpl --version`:
Chapel 1.15.0

lydia-duncan:
@bradcray - anything to add?

bradcray:
I'd probably use a test like the following to demonstrate the current inconsistency:

````chapel
var x = [i in 1..10] i;
writeln(x.type:string);

proc foo(x) {
  writeln(x.type:string);
}

foo([i in 1..10] i);
````

The question about expressing its type could be thought of as "what type specifier should be added to the variable/argument declarations of 'x' above in order to constrain them to be iteratable expressions?"

**Comment by lydia-duncan** [comment]
Currently that program gives the following output:
```
[domain(1,int(64),false)] int(64)
_ir__iterator_for_loopexpr3
```
with the second line being consistent for both --devel and --no-devel compilation.  This is definitely unfortunate, as it's using a compiler-internal naming scheme.  I don't have specific guidance on what that should look like instead, though.

bradcray:
Though we should improve the compiler's description of the iterator type, I think that's a separate issue (or at least is not what I was suggesting we focus on here -- I just offered the program as a way to demonstrate that the compiler thinks it's an array in one case and an iterator expression in the other (where the first line is the array type)).

ben-albrecht:
Determining a syntax / method in https://github.com/chapel-lang/chapel/issues/14364 could provide a way for specifying whether to create an array or iterator, e.g.

```chpl
var A = [i in 1..10] i;             // iterator
[...]

PR #6333, opened 2017-05-26
Add a future to track the type of iterable expressions in multiple contexts

lydia-duncan:
I rediscovered that the type of an iterable expression is different when it is
saved into a local variable versus when it is passed as an argument to a
function.  This future tracks that difference, and poses the question of
whether the types should match and what it should be if they do.

This issue is tracked in #6330


lydia-duncan:
@bradcray - I changed the code you gave me so that the .bad file wouldn't rely on the "_ir__iterator_for_loopexpr3" output, which seemed like something that could change depending on compiler AST reshuffling.  I verified that `==` would give me the right thing for strings.  Thoughts?

bradcray:
Looks OK to me.

bradcray:
(BTW, to the question raised in the future, I strongly believe we should not expand iterators into arrays upon passing them into functions... I believe that doing so would cause the compiler to insert array temps in places where we wouldn't want them.  So that would suggest either permitting variable declarations to capture iterator expressions or having the cases continue to behave differently).

bradcray:
(Oh, also, I don't think you "discovered" this -- we implemented the language this way very intentionally.  :)  ).

lydia-duncan:
So I am Christopher Columbus?

bradcray:
Hopefully less controversial than him...

PR #6346, opened 2017-05-30
Hide CHPL_JEMALLOC in printchplenv unless explicitly set

ben-albrecht:
`CHPL_JEMALLOC` is not intended for user exposure. This PR hides it from the `printchplenv` output unless it is set to anything except its default value of `jemalloc` when using `CHPL_MEM=jemalloc`.

This follows up on some of the changes made in #6133


ben-albrecht:
@ronawho && @mppf -- thoughts?

ronawho:
Don't went want to update the documentation in chplenv.rst as well?

PR #6350, opened 2017-05-30
Don't ignore running tasks in sublocale-sensitive iterators.

gbtitus:
For locModel!=flat we long ago switched to ignoring the running task
count for purposes of chunking iterations in DefaultRectangular, because
the tasks created for the coforall across sublocales would interfere
with computing the number of intra-sublocale tasks to create.  But since
then we've changed how we compute the number of tasks to use in nested
iteration loops, so we actually shouldn't be ignoring other running
tasks any more.  Here, take into account other running tasks (that is,
outside this parallel construct) when computing how many total tasks to
apply to the nested looping constructs we use when locModel!=flat.


gbtitus:
@ronawho, @benharsh, would you review this one, please?

ronawho:
👍

PR #6355, opened 2017-05-31
Switch ResolutionCandidate's actual and formal tracking to use std::vectors

lydia-duncan:
Prior to this change, the fields formalIdxToActual and actualIdxToFormal were
sparsely populated Vecs.  This change modifies them to be std::vectors.  There was
some reliance on the sparsely populated Vecs having a larger size than they
would necessarily be populated with, so I modified some iterations through these
structures to account for that change.

Passed std/ testing


compiler/resolution/wrappers.cpp
```diff
@@ -477,17 +477,19 @@ buildDefaultWrapper(FnSymbol* fn,
 
 FnSymbol*
 defaultWrap(FnSymbol* fn,
-            Vec<ArgSymbol*>* actualFormals,
+            std::map<int, ArgSymbol*>* actualFormals,
```
noakesmichael: There is a popular convention to define functions such that inputs come first and then outputs
come last.  If there are in/outs and out then I think it is common to place the in/outs before the
outs.  I don't think it is necessary to "fix" this function for this PR but something to keep in mind?

compiler/resolution/wrappers.cpp
```diff
@@ -477,17 +477,19 @@ buildDefaultWrapper(FnSymbol* fn,
...
             CallInfo* info) {
   FnSymbol* wrapper = fn;
-  int num_actuals = actualFormals->n;
+  int num_actuals = actualFormals->size();
   int num_formals = fn->numFormals();
   if (num_formals > num_actuals) {
     Vec<Symbol*> defaults;
```
noakesmichael: Seems a shame not to take the opportunity to upgrade the variable defaults to be a std::vector.
But may that has too much impact on addCache() and buildDefaultWrapper()?

compiler/resolution/functionResolution.cpp
```diff
@@ -1622,23 +1622,20 @@ moreSpecific(FnSymbol* fn, Type* actualType, Type* formalType) {
 
 static bool
 computeActualFormalAlignment(FnSymbol* fn,
```
noakesmichael: Another case in which the procedure formals seem to me to be in a "dis-organized" sequence.

compiler/resolution/functionResolution.cpp
```diff
@@ -1622,23 +1622,20 @@ moreSpecific(FnSymbol* fn, Type* actualType, Type* formalType) {
...
-                             Vec<Symbol*>& formalIdxToActual,
-                             Vec<ArgSymbol*>& actualIdxToFormal,
+                             std::map<int, Symbol*>& formalIdxToActual,
+                             std::map<int, ArgSymbol*>& actualIdxToFormal,
                              CallInfo& info) {
```
noakesmichael: I may be missing something but it seems to me that the actual "info" could be const CallInfo& info.

compiler/resolution/functionResolution.cpp
```diff
@@ -1732,16 +1729,16 @@ getInstantiationType(Type* actualType, Type* formalType) {
...
 computeGenericSubs(SymbolMap &subs,
                    FnSymbol* fn,
-                   Vec<Symbol*>& formalIdxToActual,
+                   std::map<int, Symbol*>& formalIdxToActual,
                    bool inInitRes) {
   int i = 0;
   for_formals(formal, fn) {
     if (formal->intent == INTENT_PARAM) {
-      if (formalIdxToActual.v[i] && formalIdxToActual.v[i]->isParameter()) {
+      if (formalIdxToActual[i] != NULL && formalIdxToActual[i]->isParameter()) {
```
noakesmichael: If I understand correctly formalIdxToActual[i] != NULL is a synonym for "does this map have an
element for key "i".  Per our earlier conversation, we believe this will insert a NULL if the map did
not have an entry for "i" which seems unfortunate to me.  I may be missing something but it looks
to me that formalIdxToActual could be declared as const std::map<int, Symbol*>& formalIdxToActual
if it weren't for this.

PR #6366, opened 2017-06-01
Fix baseline regression involving tuples of void

daviditen:
In the function that removes 'void' variables find and remove tuples of
'void' in the same cases.

In a few places we expected tuples to have at least one field, but after
removing 'void' fields they can have none. Fix them up to allow tuples
with no fields to pass through.

If fn->_this is void and is not a formal argument, set it to NULL

This fixes a --baseline regression for test functions/vass/proc-iter/error-no-yield-in-iter-1.chpl.

* Passed full paratest with -compopts --baseline
* Passed full paratest with -compopts --verify


vasslitvinov:
For the commit message: the test that this fixes is: functions/vass/proc-iter/error-no-yield-in-iter-1

vasslitvinov:
As for the code changes, I would suggest adding a function `isVoidOrVoidTuple(Expr*)` that factors out a common pattern in your changes. For extra coolness, it could be overloaded to take either an Expr* or a Type*.

OK to merge with or without this addition.

daviditen:
Thanks Vass.  I fixed the commit message.  I'll look into your `isVoidOrVoidTuple` suggestion in the morning, but I think that sounds good.

PR #6369, opened 2017-06-01
Increase nightly test coverage for LLVM

awallace-cray:
Changes the existing llvm test config to
(a) run the entire test suite, instead of just examples + extern/ferguson,
(b) use paratest w 8 localhost nodes.

### DIFF
--- a/util/cron/test-llvm.bash
+++ b/util/cron/test-llvm.bash
@@ -1,12 +1,23 @@
 #!/usr/bin/env bash
 #
-# Test default configuration with CHPL_LLVM=llvm and pass --llvm flag to
-# compiler on linux64.
+# Full test suite with CHPL_LLVM=llvm and pass --llvm flag to
+# compiler on linux64. Now uses paratest.server.
+
+# Needs /data/cf/chapel/setup_python27.bash (common-llvm)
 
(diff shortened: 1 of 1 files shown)


awallace-cray:
Tested all these changes, but not all at once. 
There is one final test, that will run for a couple more hours. 
Won't merge unless that test comes out ok.

util/cron/test-llvm.bash
```diff
@@ -1,12 +1,23 @@
...
 CWD=$(cd $(dirname $0) ; pwd)
 source $CWD/common.bash
 source $CWD/common-llvm.bash
+source $CWD/common-localnode-paratest.bash
+
+# common-llvm restricts us to extern/ferguson, but we want all the tests
+unset CHPL_NIGHTLY_TEST_DIRS
+
+# paratest with 8 nodes on localhost
+nightly_args="${nightly_args} $(set +x ; get_nightly_paratest_args 8)"
```
dmk42: Let's keep an eye on this.  It should work, but if we start getting sporadic failures, we may need to dial this down to 4 nodes instead of 8.

awallace-cray:
My 8-way test passed everything, twice.. but will run a 4-way and see how long that takes. (The 8-way run takes ~2:35 total)

PR #6375, opened 2017-06-03
Alphabetize contributors

ben-albrecht:
sleepSorted™ the contributors list.

I've employed the strategy of "sort by the first last name" for people with multiple last names, e.g. "Thomas Van Doren" is sorted by "Van", and "Marcos Cleison Silva Santana" is sorted by "Cleison".

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,7 +9,6 @@ Contributors to the current release
 * Ben Albrecht, [Cray Inc.]
 * Paul Cassella, [Cray Inc.]
 * [Brad Chamberlain], [Cray Inc.]
-* Marcos Cleison Silva Santana, individual contributor
 * Sung-Eun Choi, [Cray Inc.]
 * Lydia Duncan, [Cray Inc.]
 * Michael Ferguson, [Cray Inc.]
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.md
```diff
@@ -52,6 +52,7 @@ Contributors to previous releases
 * Joel Denny, [ETI]
 * Roxana Diaconescu, [CalTech]
 * James Dinan, [Cray Inc.] / [The Ohio State University]
+* Thomas Van Doren, [Cray Inc.]
```
lydia-duncan: I don't think this movement is right.  Thomas' last name is "Van Doren", Van isn't his middle name

ben-albrecht:
It does feel wrong...  I asked the internet and the conclusion seems to be
that there are strict rules for sorting multi-word last names, but they
vary from country to country. It's ultimately more practical to pick a
simple set of rules such as by the very last name, and stick to that. We
could alternatively pick a different set of rules, such as sorting by the
first last name.

Given the minimal importance here, I could be swayed easily if one felt
strongly about our approach. :)

lydia-duncan:
I'd be in favor of sorting by first last name :)

ben-albrecht:
>> Given the minimal importance here, I could be swayed easily if one felt
strongly about our approach. :)

> I'd be in favor of sorting by first last name :)

Consider me swayed. I'm sure @thomasvandoren will appreciate it anyway.

Issue #6356, opened 2017-05-31
Consider using processor atomics for remote coforalls EndCount


ronawho:
https://github.com/chapel-lang/chapel/pull/2451 forced "local" coforalls to use processor atomics for better performance. Processor atomics are faster than network atomics for operations that are only done on a single locale, so it makes sense to have local coforalls use processor atomics. We left "remote" coforalls (coforall+on) using network atomics with the thought that the end count will mostly be manipulated by remote locales, in which case network AMOs will be faster than a remote-fork+processor atomic.

So for someting like:

```chapel
coforall loc in Locales do on loc { }
```

the spawning task will increment the EndCount once for each task, and then do a waitFor, while remote tasks will decrement the EndCount when they're done running. The remote tasks decrementing the EndCount should benefit from using network atomics, but it seems doing the network AMOs possibly including the waitFor hurts things overall. 

Surprisingly, switching to using processor atomics (switching `forceLocalTypes` to `gTrue` in https://github.com/chapel-lang/chapel/blob/master/compiler/AST/build.cpp#L1584) results in significantly better performance for some multi-locale codes including stream, RA, Lulesh , and prk stencil

Here's a few sample timings:

| Benchmark   | u+q default  | u+q procssor atomic EndCount |
| ----------- | ------------ | ---------------------------- |
| lulesh      | 3.35s        | 2.9s                         |
| stencil prk | 98 GFlops/s  | 104 GFlops/s                 |
| stencil-opt | 193 GFlops/s | 235 GFlop/s                  |
| prom stream | 440 GB/s     | 640 GB/s                     |

Here's a full list of perf differences: https://gist.github.com/ronawho/719c6f6a470b10599d73f7b5d36c57d9

A lot of these benchmarks have pretty quick running kernels (noted in https://github.com/chapel-lang/chapel/issues/6323), so we should increase them before making this change or at least evaluate the benefit of this change for different length kernels.

The relatively large impact of changing endcounts also indicates that we should really implement the "bounded-coforall" (https://github.com/chapel-lang/chapel/pull/5018) optimization for remote coforalls as well. 

We should also create a remote coforall microbenchmark to help determine the impact of network vs. processor atomic EndCounts.

TODOs:

 - [x] create a microbenchmark for coforall+on (#6378)
 - [x] implement bounded coforall optimization for remote coforalls (#6357, #6419)
[...]

PR #6378, opened 2017-06-05
Add a multi-locale performance test to track remote task startup

ronawho:
Add a performance test to track the cost of doing:

```chapel
coforall loc in Locales do on loc {}
```

This idiom is a fundamental building block of our distributions and parallel
applications, so we should track its performance. This is similar to the
task-spawning benchmarks added in #4102, except that we're spawning remote
tasks instead of local ones.

We've recently discovered that remote task spawning for coforall+on's is more
expensive when using network atomics (#6356), so this test will serve as a way
to track progress on optimizations that improve that and serve as a baseline
comparisons to tree-based spawning schemes we want to explore in the future.


ronawho:
@benharsh mind reviewing?

PR #6379, opened 2017-06-05
Minor adjustment in PRK-DGEMM

e-kayrakli:
@cassella noticed that unblocked branch in DGEMM has forall that looks unsafe. This PR fixes the issue. Also:

- Diverges from OpenMP loop ordering in favor of the standard ijk order.
- Uses a predefined domain to avoid creating domains at every iteration. (Similar to the blocked branch)

Changed test passes with standard linux64 config.

### DIFF
--- a/test/studies/prk/DGEMM/dgemm.chpl
+++ b/test/studies/prk/DGEMM/dgemm.chpl
@@ -58,9 +58,10 @@ if blockSize == 0 {
   for niter in 0..#iterations {
     if iterations==1 || niter==1 then t.start();
 
-    // loop order is the same as the OpenMP version
-    forall (j,k,i) in {vecRange, vecRange, vecRange} {
-      C[i,j] += A[i,k] * B[k,j];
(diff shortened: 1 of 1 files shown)


ronawho:
It's not obvious to me what the potential race is. Could you elaborate? (Don't hold up the commit over my question, I'm just curious)

e-kayrakli:
@ronawho - Using @cassella's example; without this, there is no guarantee in the language that (j,k1,i) and (j,k2,i) will be mapped to the same task. We just know that in practice default rectangular parallel iterators divide across rows (with sufficiently large data?). But this is by no means enforced, ie default rectangular parallel iterators may change and divide across the second dimension, which would make this code unsafe. Because two separate tasks would end up writing to C[i,j] concurrently. Are we missing something?

ronawho:
Ah ok, thanks for the clarification. Yeah, theoretically every iteration of the loop could be run in separate tasks, so that was definetly a race.

PR #6380, opened 2017-06-05
Increase PRK stencil problem size

benharsh:
Increases problem size to avoid very short subsections.


benharsh:
@ronawho : Look good? These tests will now take at least one second on average per iteration. I tried 256k with the reference but ran into further integer-width issues, which I'd rather not spend time on right now.

Issue #6383, opened 2017-06-06
Iterating over Block with for and forall give different results


npadmana:
### Summary of Problem
The following code iterates over a Block distributed array using ``for`` and ``forall``
and gets different results (when they should be the same).

This occurs even for single locale Chapel.

Note that if I do ``const D = DSpace``, the code does the correct thing, and yields 
``diff=0``.

### Steps to Reproduce

**Source Code:**

```chapel
use BlockDist;
use Random;

config const Ng=128;
const DSpace={0.. #Ng, 0.. #Ng, 0.. #(Ng+2)};
var targets : [0.. #numLocales, 0..0,0..0] locale;
targets[..,0,0] = Locales;
const D : domain(3) dmapped Block(boundingBox=DSpace, targetLocales=targets) = DSpace;
const Dre = D[..,..,0.. #(Ng+2) by 2 align 0];
const Dim = D[..,..,0.. #(Ng+2) by 2 align 1];
var GG,GG1 : [D]real;

fillRandom(GG, seed=1234);
GG1 = GG;
const twopi = 2*pi;

// Do an iteration over this using for
for (x0, x1) in zip(GG[Dre], GG[Dim]) {
  // Box-Muller
  x0 += if (x0 < 1.0e-30) then 1.0e-30 else 0.0;
  const rr = sqrt(-2*log(x0));
  const ang = twopi*x1;
  x0 = rr*cos(ang);
  x1 = rr*sin(ang);
}

// Do an iteration over this using forall
forall (x0, x1) in zip(GG1[Dre], GG1[Dim]) {
  // Box-Muller
  x0 += if (x0 < 1.0e-30) then 1.0e-30 else 0.0;
  const rr = sqrt(-2*log(x0));
  const ang = twopi*x1;
  x0 = rr*cos(ang);
  x1 = rr*sin(ang);
}

// This should be zero!!
var diff = max reduce abs(GG-GG1);
writeln(diff);
```

**Compile command:**
`chpl bug.chpl`

**Execution command:**
`./a.out`

This should return 0, but does not.

### Configuration Information
``chpl Version 1.16.0 pre-release (4924299)``

```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: mpi-gnu *
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none *
CHPL_TASKS: fifo *
CHPL_LAUNCHER: none
[...]

PR #6384, opened 2017-06-06
Fix a bug in parallel iteration over aligned block-distrib. doms/arrs

bradcray:
This is motivated by issue #6383 filed by Nikhil and Paul.

TODO:
- [ ] verify assumption made that alignment is always in 0..#stride
- [ ] add tests to make sure I didn't introduce a bug like the one Nikhil caught below
- [ ] full block dist test run
- [ ] see if other dists have similar issues and test as appropriate

### DIFF
--- a/modules/dists/BlockDist.chpl
+++ b/modules/dists/BlockDist.chpl
@@ -849,9 +849,10 @@ iter BlockDom.these(param tag: iterKind, followThis) where tag == iterKind.follo
   type strType = chpl__signedType(idxType);
   for param i in 1..rank {
     var stride = whole.dim(i).stride: strType;
+    var alignment = whole.dim(i).alignment;
     // not checking here whether the new low and high fit into idxType
-    var low = (stride * followThis(i).low:strType):idxType;
-    var high = (stride * followThis(i).high:strType):idxType;
+    var low = (stride * followThis(i).low:strType):idxType + alignment;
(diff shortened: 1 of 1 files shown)


bradcray:
Note: I likely won't have time to test and merge this let alone work on the TODOs for at least a week if someone wants to push this forward for @npadmana

modules/dists/BlockDist.chpl
```diff
@@ -1125,9 +1126,10 @@ iter BlockArr.these(param tag: iterKind, followThis, param fast: bool = false) r
 
   for param i in 1..rank {
     var stride = dom.whole.dim(i).stride;
+    var alignment = whole.dim(i).alignment;
```
npadmana: Shouldn't this be ``dom.whole.dim(i).alignment``?

bradcray: D'oh!  Yes.  You can tell I wanted to go to bed too badly to test that branch (I was proud to have even looked for it after getting the original test to work.  Did I mention I'd been up since 3am yesterday when I wrote this at 11:30pm?

benharsh:
We should probably make a similar change to StencilDist.

I'll try and take this over once I get past a couple of in-progress tasks.

benharsh:
I have opened PR #6406 that fixes this kind of bug for the Block, Cyclic, and Stencil distributions.

PR #6390, opened 2017-06-06
Mothball multi-ddata.

gbtitus:
Add a config param that by default disables the multi-ddata feature,
thus mothballing it.

The multi-ddata solution for memory affinity on NUMA architectures has
always had poor performance on important idioms, notably iteration over
the domain or range of an array instead of over the array itself.  This
does not seem to be something we can fix -- in most cases we simply
can't remove enough of the multi-ddata array access overhead to bring
the performance at all close to that of locModel=flat.  So, while
multi-ddata does allow an elegant solution to the problem of getting
good NUMA locality in the NIC-registered heap on Cray Aries-based
systems, the performance losses elsewhere more than offset that.  Here,
disable multi-ddata using a config param, thus effectively removing the
feature but retaining the code for now against the likelihood we might
want to use it in the (near?) future for an explicit layout.

There are a couple of not-yet-merged multi-ddata improvements archived
in the form of closed PRs which should be resurrected if we ever bring
multi-ddata back:

- #6385:
  Reorganize the multi-ddata information to use tuples of per-chunk
  info rather than ddatas of chunk info.

- #6387:
  Use multiply-by-reciprocal to implement the divide needed when
  computing the chunk that holds an array element.


gbtitus:
@ronawho and @bradcray, you may be interested in this one as well.

(There will be at least one more commit, to fix the `*.lm-numa.good` conflicts.)

ronawho:
Looks good to me. I'd also be ok with keeping some of the multi-ddata tests around if you wanted to just throw `-sdefRectDisableMultiDData=false` to them, but I don't object to removing them altogether.

ronawho:
Also, can you open an issue to remind us to update chip 18?

gbtitus:
I didn't think about keeping the multi-ddata tests around and turning the feature back on to test them, but it seems like it would put multi-ddata into a sort of zombie state: it would be mothballed but we'd still be maintaining its unit tests and, by extension, the feature itself.  I'm not enthused about that.  Plus we do want the CommDiagnostics tests to revert to the non-multi-ddata values and the tests that print DefaultRectangularArr class instances for lm-numa not to mention the now-nonexistent multi-ddata fields.  It seems better to force anyone resurrecting multi-ddata to bring all the tests back with it, rather than a subset.

ronawho:
If we want to mothball multi-ddata and not test it, I think we should remove the code from the modules as well.

gbtitus:
New issue requesting Chip 18 update is https://github.com/chapel-lang/chapel/issues/6396.

gbtitus:
Against the admittedly slight possibility that in the next couple of months we can make the local-field pragma trickle down from tuple fields to the elements thereof, I'm leaving multi-ddata in for now so as to make it easier to perf-test the combination of multiply-by-reciprocal and tuple-of-ddatas.

PR #6401, opened 2017-06-07
Add futures for bug in chpldoc output of certain arrays

lydia-duncan:
Ben A encountered an issue with type aliases for arrays, where instead of
`.. type:: type t = [1..10] int;` we would get:

```
.. function:: proc _parloopexpr1(iterExpr)

.. type:: type t = AppendExpr.Call08
```

I discovered that the same thing was true for arrays created using the for loop
syntax, `var x = [i in 1..10] i;`

This change adds two futures to track these failures.

Fixing this issue will be a bit tricky, as there are two problems.  First,
_parloopexpr1 is a compiler created function that is defined at the scope of
the assignment and so should not be part of the documentation output.  Second,
the init portion of the global should look the same as the declaration in user
code.


lydia-duncan:
@ben-albrecht - here are the futures for that bug

ben-albrecht:
Thanks!

PR #6402, opened 2017-06-07
Dynamic iters default chunk size

s-geronimoanderson:
Added default value 1 for dynamic iterator's chunkSize formal, and updated shootouts not to pass 1 as the chunkSize actual. Note: My editor seems to have removed trailing spaces automatically, which makes the diff appear more involved than it actually is.


benharsh:
These changes look good to me. Since this PR modifies the release shootouts, @chapel-lang/perf-team might be interested in reviewing as well.

daviditen:
In mandelbrot.chpl and mandelbrot-fast.chpl `chunkSize` was a user-configurable setting, but now it's always forced to 1.  I think it should probably be put back in those two cases.

bradcray:
> In mandelbrot.chpl and mandelbrot-fast.chpl chunkSize was a user-configurable setting, but now it's always forced to 1. I think it should probably be put back in those two cases.

I agree.  Or we could remove the config if we don't anticipate values other than 1 being useful...

test/release/examples/benchmarks/shootout/mandelbrot-fast.chpl
```diff
@@ -10,7 +10,7 @@ use DynamicIters;
 
 config const n = 200,             // image size in pixels (n x n)
              maxIter = 50,        // max # of iterations per pixel
-             limit = 4.0,         // per-pixel convergence limit
+             limit = 4.0;         // per-pixel convergence limit
```
benharsh: I think you need to replace the semicolon with a comma here.

test/release/examples/benchmarks/shootout/mandelbrot.chpl
```diff
@@ -10,7 +10,7 @@ use DynamicIters;
 
 config const n = 200,              // image size in pixels (n x n)
              maxIter = 50,         // max # of iterations per pixel
-             limit = 4.0,          // per-pixel convergence limit
+             limit = 4.0;          // per-pixel convergence limit
```
benharsh: And here

s-geronimoanderson:
I debated what to do in the mandelbrot* tests, but I also agree that it's sensible to leave the chunkSize config const as intended. I've added it back in for those.

Issue #6374, opened 2017-06-03
Inline proc : making return optional with exit()


TomsyPaul:
In an inline proc that returns something, the return statement is compulsory in every branch, even in one which uses exit(). It would be better and more appropriate, to make return optional for such branches. 

https://sourceforge.net/p/chapel/mailman/message/35874026/

lydia-duncan:
I added some futures and regular tests to track this in #6377

lydia-duncan:
On the mailing list, @bradcray pointed out that "halt()" does not trigger the same error message, implying that this might be an easy fix.  I may try to get to that in the next couple of days

TomsyPaul:
Just to mention - An iterator has no issues with exit().

daviditen:
PR #6403 was merged today and should do what you expect.  I'm going to leave this issue open until it has been through nightly testing tonight.

TomsyPaul:
Just to mention - An iterator has no issues with exit().

daviditen:
Nightly testing is looking good, so I think this can be closed now.

PR #6403, opened 2017-06-07
Add a pragma "function never returns" for functions like 'halt' and 'exit'

daviditen:
Add a pragma for functions like 'halt' and 'exit' that don't return but cause
the program to terminate instead. Check for this pragma in returns-on-all-paths
analysis instead of checking for functions named 'halt'.

Added the new pragma to the definitions of 'halt' and 'exit'.

Added a test that defined it's own non-halting 'halt' which caused stack garbage to be assigned to a variable before this change.

Remove .future and .bad files for two newly passing tests.

Passed paratest with -futures and --verify.


daviditen:
This should fix issue #6374

compiler/include/flags_list.h
```diff
@@ -118,6 +118,7 @@ symbolFlag( FLAG_FN_RETARG, npr, "fn returns via _retArg", ncm )
 symbolFlag( FLAG_FORMAL_TEMP,     npr, "formal temp", "a formal temp to back an in, out, or inout argument" )
 symbolFlag( FLAG_FORWARDING_FN , npr, "forwarding function" , ncm )
 symbolFlag( FLAG_FUNCTION_CLASS , npr, "function class" , "first-class function class representation" )
+symbolFlag( FLAG_FUNCTION_NEVER_RETURNS, ypr, "function never returns", "function that causes the program to exit" )
```
lydia-duncan: Maybe FN_TERMINATES_PROGRAM or something?  I could see "never returns" being confused with "has no return value"

daviditen: Thanks!  I've renamed it to "function terminates program".  I'm doing another test run just to be on the safe side.

PR #6404, opened 2017-06-07
Dynamic iters domain guided adaptive

s-geronimoanderson:
Created domain versions for guided and adaptive iterators, using the dynamic domain iterator as an example. (The new code is highly similar.)

Also created domain versions of the checkAdvancedIters tests (dynamic, guided, and adaptive). These are essentially the same as those tests using ranges.


test/functions/iterators/angeles/checkAdvancedIters/checkAdaptiveWSIter.chpl
```diff
@@ -5,43 +5,66 @@ config const nTasks=4;           // number of cores; should be here.maxTaskPar?
...
+var dmn:domain(1)={rng};         // The domains
+var dmns=dmn by 2;
 
-var A:[rng] int=0;              // The test arrays
+var A:[rng] int=0;               // The test arrays
 var B:[rngs] int=0;
 var C:[rngs,rng] int=0;
+var D:[dmn] int=0;
+var E:[dmns] int=0;
+var F:[{rngs,rng}] int=0;
```
benharsh: Can we avoid the domain literal here and just do this?
```chpl
var F:[rngs,rng] int=0;
```

PR #6406, opened 2017-06-07
Fix distribution alignment issues with iteration

benharsh:
Issue #6383 concerned a bug with iteration over an align'd Block-distributed domain. The serial iterator yielded the correct indices, but the parallel iterator seemingly ignored alignment entirely. After adding a new test, it became apparent that other distributions had a similar problem. The general fix is to use ``alignedLow`` instead of ``low``.

I also added a StencilDist option to the distribution robustness suite, but only checked StencilDist with this new test.

Testing:
- [x] full local
- [x] full gasnet
- [x] dist-cyclic
- [x] dist-replicated


vasslitvinov:
Good to go!

PR #6407, opened 2017-06-07
Temporary workaround for gcc - mpich conflict on cray internal test system

awallace-cray:
Changes the build script that runs one automated nightly Chapel test series (test-xc-wb, for whitebox).  No effect on Chapel itself. 

It is a temporary workaround for a specific problem on this internal system, so Chapel testing may proceed.


awallace-cray:
The only thing riding on this change is whether the nightly whitebox test succeeds.

awallace-cray:
Not yet.

PR #6410, opened 2017-06-07
Improve LICM to enable better hoisting from the dgemm PRK

ronawho:
Only check if definitions dominate all loop exists for variables that aren't
known to be dead out of loops.

We have a defDominatesAllExits() check that avoids hoisting for codes like:

```chapel
var b = 2;
for i in 1..maxIter {
  if converged() then
    break;
  b = 4;
}
writeln(b);
```

Even though `b = 4;` is loop invariant, we can't hoist it because the
definition of b does not dominate all loop exists, and if the loop breaks on
the first iteration, we'll print `4` instead of `2`.

Previoulsy we just checked if definitions dominated all exists for all
variables, but we only actually need to check for variables that are live out
of the loop.

e.g. we could hoist b from:

```chapel
for i in 1..maxIter {
  if converged() then
    break;
  var b = 4;
}
```

because b's defPoint is inside the loop so it can't be live out of the loop.

This basically turns our check for `if (defDominatesAllExits()) {...}` into `if
(defPointInLoop() || defDominatesAllExits()) {...}`. We're not actually doing
true liveness analysis, we're just doing a more limited check to note that if a
variable's defPoint is inside the loop, it can't be live outside of it.

### DIFF
--- a/compiler/optimizations/loopInvariantCodeMotion.cpp
+++ b/compiler/optimizations/loopInvariantCodeMotion.cpp
@@ -110,22 +110,29 @@ public:
       delete bitExits;
     }
 
-    // This function exists to place an expr in the
-    // "preheader" of the loop,
-    void insertBefore(Expr* expr) {
+    // Get the BlockStmt AST that corresponds to this BasicBlock representation
+    // of the loop
+    BlockStmt* getLoopAST() {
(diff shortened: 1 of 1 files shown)


ronawho:
Currently failing 7 tests on master, so closing until I have time to get back to this.

PR #6412, opened 2017-06-08
Add a symbolic value for "all sublocales".

gbtitus:
We've long had symbolic values for "no sublocale" and "any sublocale",
in both the module and runtime code.  Here add a symbolic value for "all
sublocales", wanted for some upcoming work in memory localization.

While here, I also modified a tasks=fifo `assert()` that would have let a
caller pass `sublocale==0` to `chpl_task_addToTaskList()`; that should not
occur.

### DIFF
--- a/modules/internal/localeModels/knl/LocaleModel.chpl
+++ b/modules/internal/localeModels/knl/LocaleModel.chpl
@@ -513,11 +513,11 @@ module LocaleModel {
     proc localeIDtoLocale(id : chpl_localeID_t) {
       const node = chpl_nodeFromLocaleID(id);
       const subloc = chpl_sublocFromLocaleID(id);
-      if chpl_isActualSublocID(subloc) then
-        return (myLocales[node:int].getChild(subloc:int)):locale;
-      else if (subloc == numaDomainForAny(
-                        (myLocales[node:int]:LocaleModel).numSublocales)) then
+      if subloc == numaDomainForAny(
+                    (myLocales[node:int]:LocaleModel).numSublocales) then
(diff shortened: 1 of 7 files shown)


gbtitus:
@ronawho, would you please review this one?  And @dmk42, would you also take a look at least at the changes to the KNL locale model code?  I haven't tested that yet -- the appropriate platform won't be available until later today.  (Any feedback you can give on testing modes/techniques would be appreciated, too.)

modules/internal/localeModels/knl/LocaleModel.chpl
```diff
@@ -513,13 +513,13 @@ module LocaleModel {
     proc localeIDtoLocale(id : chpl_localeID_t) {
       const node = chpl_nodeFromLocaleID(id);
       const subloc = chpl_sublocFromLocaleID(id);
-      if (subloc == c_sublocid_none) || (subloc == c_sublocid_any) then
-        return (myLocales[node:int]):locale;
+      if chpl_isActualSublocID(subloc) then
+        return (myLocales[node:int].getChild(subloc:int)):locale;
       else if (subloc == numaDomainForAny(
                         (myLocales[node:int]:LocaleModel).numSublocales)) then
```
dmk42: With the new organization, these two `if` statements need to be inverted.  The `numaDomainForAny()` value passes the `chpl_isActualSublocID()` test, so in the current order, the second `if` statement would never be executed.

gbtitus: Oops, yes, I missed that -- bizarrely, I was thinking that the magic `numaDomainForAny()` value was also not an "actual sublocale ID".  Working on a fix ...

Issue #6413, opened 2017-06-08
DGEMM PRK performance rate computation is incorrect


ronawho:
Only `iterations-1` trials are actually timed, but the computation still does `t.elapsed()/iterations`

Instead of `for niter in 0..#iteration` I think it should be `for niter in 0..iteration` to match the reference and the rest of our PRK implementations. There may be some places that need to be changed from `iterations` to `iterations+1` then, but you should be able to remove the `iterations==1` checks from the trial loops.

I was doing 3 iterations while debugging and 20 iterations while getting ready to report numbers for #6388 and for some reason I was seeing very different performance rates.

ronawho:
@ben-albrecht and/or @e-kayrakli mind taking a look at this?

e-kayrakli:
:( I'm on it. It looks like the same bug is there at least in the SERIAL version in the base PRK repo

ronawho:
Thanks @e-kayrakli. 

I think this inflates our performance compared to the OpenMP version, which is probably why your `Chapel Carr` version was slightly outperforming OpenMP.

e-kayrakli:
Yes, makes sense. I wonder what would the comparison look like with #6414. I'll try to that but not today.

ben-albrecht:
#6414 merged - thanks @e-kayrakli !

PR #6414, opened 2017-06-08
Fix the timing error in PRK-DGEMM

e-kayrakli:
Fix a small bug with potentially big impact in reported performance. Solves #6413 

The test passed with standard linux64 config.

### DIFF
--- a/test/studies/prk/DGEMM/dgemm.chpl
+++ b/test/studies/prk/DGEMM/dgemm.chpl
@@ -54,13 +54,13 @@ if !correctness {
   writeln();
 }
 
-const refChecksum = (iterations) *
+const refChecksum = (iterations+1) *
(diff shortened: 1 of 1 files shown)


e-kayrakli:
@ronawho - It seems this is all needed to be fixed. I quickly compared with base OpenMP implementation, especially in terms of the checksum. Did you notice anything else?

ronawho:
I think you'll want to get rid of the `iterations==1` checks to start the timer. Otherwise for 1 iteration, you'll start the time on iteration 0.

e-kayrakli:
Assuming you meant `niter==1`...

e-kayrakli:
Ah, true.

Small nit: I have used `niter` whereas Stencil and Transpose use `iteration` as the current iteration number. I think this confusion needs to be resolved in favor of the latter.

PR #6415, opened 2017-06-08
Add annotations for this week

lydia-duncan:
Note multi-locale performance changes for the stencil PRK and NPB EP.

Verified the annotation checker passed these changes

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -378,6 +378,10 @@ ep-b:
   11/22/14:
     - revert -E use for nightly on 21st (which failed, unrelatedly)
 
+ep.ml-perf:
+  06/08/17:
+    - Mothball multi-ddata (#6390)
+
 fannkuch-redux:
   07/22/14:
     - Error due to cleaning up file incompletely
@@ -745,6 +749,8 @@ prk-stencil.ml-perf:
     - Move chpl_getPrivatizedClass() into chpl-privatization.h (#6212)
   05/18/17:
     - Use nonblocking transactions for strided transfers (#6262)
+  06/06/17:
+    - Increase the PRK stencil problem size (#6380)
 
 prk-stencil-time:
   05/11/17:


ronawho:
@lydia-duncan why do you suspect the multi-ddata mothballing impacted EP under CHPL_LOCALE_MODEL=flat? multi-ddata should hopefully only impact CHPL_LOCALE_MODEL=numa

lydia-duncan:
I'm aware, but multi-ddata also impacted flat with its introduction.  EP has multiple correlations of its performance changing after ddata changes and then a month's worth of stability, and the performance shift after the introduction of ddata looked like a similar magnitude in the opposite direction.

lydia-duncan:
Still, I can remove the note if you think I should (since you were one of the reviewers, right?), or shift it to a question mark

ronawho:
I don't really see a concrete drop when multi-data first went in. It looks like u+q might have gotten worse, but it was really bimodal/noisy and gn-aries/gn-mpi didn't change, so it seems surprising to me to multi-ddata would have impacted EP. 

So far as I know, the original multi-ddata work only hurt performance for some single-locale benchmarks, because multit-ddata fields increased the array size, which meant we couldn't fit as much in cache. That was resolved by #5723, so I wouldn't expect any changes from mothballing multi-data.

And I don't mean to say it's not multi-ddata, but it seems unlikely to me and worth bisecting.

ronawho:
So bisecting didn't turn up anything. I'm still seeing the better performance even going back a week.

It's probably from the gcc 6.3 -> 7.1 target compiler upgrade then.

ronawho:
Yup, can confirm, it's the gcc 7.1 upgrade 

Swap out the EP annotation for and all annotation with: 
```
  06/08/17:
    - text: upgrade gnu target compiler from 6.3.0 to 7.1.0
      config: Single node XC, 16 node XC
```

?

lydia-duncan:
Ah, good thinking!  Yeah, I'll do that swap, thanks!

PR #6416, opened 2017-06-08
Enable testing LCALS in correctness testing

daviditen:
LCALS has only been getting tested for performance testing because with bounds
checks it takes way to long to run. Add a .compopts file with
'--no-bounds-checks' to enable it to run in regular correctness testing.

Modified/added the test config files to run the large loop size for the RAW and
RAW_OMP variants for correctness.

Modified a config param to allow selecting between three output formats:
* MINIMAL: Don't print timing/checksum dumps
* REFERENCE: Like the reference code
* PERF_TEST: Formatted nicely for perf testing

I tried running with bounds checking, and hit an out of bounds error at around
the 40 minute mark. Looking into this, I found that the reference version also
accesses the array out of bounds with index -1.  I added a comment mentioning
this where it happens.

Tested on linux64 and OSX with correctness and performance testing modes.

### DIFF
--- a/test/release/examples/benchmarks/lcals/LCALSMain.chpl
+++ b/test/release/examples/benchmarks/lcals/LCALSMain.chpl
@@ -131,7 +131,8 @@ module main {
 
     const cache_size = 0;
     const host_name = here.name;
-    writeln("\n Running loop suite on ", host_name);
+    if outputFormat != 0 then
+      writeln("\n Running loop suite on ", host_name);
 
     allocateLoopSuiteRunInfo(host_name,
                              num_suite_passes,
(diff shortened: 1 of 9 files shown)


test/release/examples/benchmarks/lcals/LCALSMain.chpl
```diff
@@ -160,12 +161,24 @@ module main {
...
 
     writeln("\n generate reports....");
-    if !perfTesting {
-      generateTimingReport(run_variants, output_dirname);
-      generateChecksumReport(run_variants, output_dirname);
-      generateFOMReport(run_variants, output_dirname);
-    } else {
-      generatePerfTimingReport(run_variants);
+
+    select outputFormat {
```
ronawho: Switch outputFormat to an enum instead?

ronawho:
I think it'd be nice to use an enum instead of just different integer sentinels, but looks good other than that!

daviditen:
Thanks! I'll change it.

ronawho:
👍

PR #6418, opened 2017-06-08
Optimize updateFluff

benharsh:
The original updateFluff implementation might end up using many GETs or
PUTs for noncontiguous chunks of memory. This commit implements a
variant of updateFluff that packs the elements into a buffer to be used
in a single GET or PUT, and is unpacked on the destination locale. This
optimization will fire if the element types can be bulk-transferred.

To implement this change, the StencilDom.setup() method was modified
such that each locale now knows how to send its own indices or how to
fetch indices. This information is stored arrays of domains which are
used in the optimized updateFluff method.

This optimized variant can be disabled with the new config param:
  stencilDistAllowPackedUpdateFluff

This optimization was motivated by the Stencil PRK when we observed that
our updateFluff stage was performing significantly worse than the
reference's corresponding stage. This commit improves our performance
depending on the comm layer used. In gasnet-mpi, we see a 10x
improvement. In ugni-qthreads, a 3x improvement.

Minor style cleanups were made while working on this change.

Testing:
- [x] full local
- [x] full gasnet
- [x] miniMD performance not impacted (as expected)
- [x] numa testing for distributions/bharshbarg/stencil/

### DIFF
--- a/modules/dists/StencilDist.chpl
+++ b/modules/dists/StencilDist.chpl
@@ -32,12 +32,6 @@
 // mapped to by the distribution.
 //
 
-//
-// TO DO List
-//
-// 1. refactor pid fields from distribution, domain, and array classes
-//
-
 use BlockDist;
 use DSIUtil;
 use ChapelUtil;
@@ -306,6 +300,9 @@ class StencilDom: BaseRectangularDom {
 // stridable: generic domain stridable parameter
 // myBlock: a non-distributed domain that defines the local indices
 //
+// NeighDom will be a rectangular domain where each dimension is the range
+// -1..1
+//
 class LocStencilDom {
   param rank: int;
   type idxType;
@@ -365,6 +362,7 @@ class LocStencilArr {
   var locRADLock: atomicbool; // This will only be accessed locally
                               // force the use of processor atomics
 
+  // TODO: use void type when packed update is disabled
   var recvBufs, sendBufs : [locDom.NeighDom] [locDom.bufDom] eltType;
   var sendRecvFlag : [locDom.NeighDom] atomic bool;
 
@@ -1015,7 +1013,7 @@ proc StencilDom.setup() {
         forall (recvS, recvD, sendS, sendD, N, L) in zip(myLocDom.recvSrc, myLocDom.recvDest,
                                                          myLocDom.sendSrc, myLocDom.sendDest,
                                                          myLocDom.Neighs, ND) {
-          if !zeruTuple(L) && recvS.size != 0 {
+          if !zeroTuple(L) && recvS.size != 0 {
             var other = -1 * L;
 
             proc checker(me, myDom, them, themDom) {
--- a/modules/dists/StencilDist.chpl
+++ b/modules/dists/StencilDist.chpl
@@ -54,6 +54,18 @@ use CommDiagnostics;
 config param debugStencilDist = false;
 config param debugStencilDistBulkTransfer = false;
 
+// TODO: part of constructor?
+config param stencilDistAllowPackedUpdateFluff = true;
+
+// Instructs the _packedUpdate method to only perform the optimized buffer
+// packing if the number of GETs/PUTs would be greater than or equal to the
+// value in this config const.
+//
+// Currently disabled until we can find a way to determine a good value.
+//
+// TODO: find better heuristic through experimentation
+config const stencilDistPackedUpdateMinChunks = 1;
+
 // Re-uses these flags from BlockDist:
 //   - disableAliasedBulkTransfer
 //   - sanityCheckDistribution
@@ -300,7 +312,9 @@ class LocStencilDom {
   param stridable: bool;
   var myBlock, myFluff: domain(rank, idxType, stridable);
   var NeighDom: domain(rank);
-  var Dest, Src: [NeighDom] domain(rank, idxType, stridable);
+  var bufDom : domain(1);
+  var recvDest, recvSrc,
+      sendDest, sendSrc: [NeighDom] domain(rank, idxType, stridable);
   var Neighs: [NeighDom] rank*int;
 }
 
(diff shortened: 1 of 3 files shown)


benharsh:
@ronawho : this should make gasnet-mpi competitive with the reference.

ronawho:
👍  nice!

modules/dists/StencilDist.chpl
```diff
@@ -351,6 +365,9 @@ class LocStencilArr {
   var locRADLock: atomicbool; // This will only be accessed locally
                               // force the use of processor atomics
 
+  var recvBufs, sendBufs : [locDom.NeighDom] [locDom.bufDom] eltType;
```
vasslitvinov: Add a TODO note to give recvBufs, sendBufs void type when !stencilDistAllowPackedUpdateFluff ?

modules/dists/StencilDist.chpl
```diff
@@ -300,7 +312,9 @@ class LocStencilDom {
   param stridable: bool;
   var myBlock, myFluff: domain(rank, idxType, stridable);
   var NeighDom: domain(rank);
```
vasslitvinov: While there, how about to the class comment above an example of what NeighDom looks like. An example for a simple stencil would already be great.

modules/dists/StencilDist.chpl
```diff
@@ -888,69 +905,126 @@ proc StencilDom.setup() {
...
+                                                         myLocDom.sendSrc, myLocDom.sendDest,
+                                                         myLocDom.Neighs, ND)
+          with ( max reduce bufLen) {
+
+          // if L is a tuple of zeroes then we are in the center, which needs
+          // no updating.
+          if !isZeroTuple(L) {
+            const to = chpl__tuplify(L);
+            var dr : rank*whole.dim(1).type;
+            for i in 1..rank {
```
vasslitvinov: Did you mean a `param` for here?

modules/dists/StencilDist.chpl
```diff
@@ -888,69 +905,126 @@ proc StencilDom.setup() {
...
+      on dist.targetLocales(localeIdx) {
+        var nearest : rank*range;
+        for param i in 1..rank do nearest(i) = -1..1;
+        const ND : domain(rank) = nearest;
+        ref myLocDom = locDoms(localeIdx);
+
+        forall (recvS, recvD, sendS, sendD, N, L) in zip(myLocDom.recvSrc, myLocDom.recvDest,
+                                                         myLocDom.sendSrc, myLocDom.sendDest,
+                                                         myLocDom.Neighs, ND) {
+          if !zeruTuple(L) && recvS.size != 0 {
```
vasslitvinov: `zeruTuple` is a typo?

modules/dists/StencilDist.chpl
benharsh: I left ``param`` out because I don't see an advantage to using it in this case.

modules/dists/StencilDist.chpl
benharsh: Indeed, I'll fix that.

modules/dists/StencilDist.chpl
```diff
@@ -1436,25 +1508,144 @@ proc _array.updateFluff() {
...
-// copy over into cache
-proc StencilArr.dsiUpdateFluff() {
-  if zeroTuple(dom.fluff) then return;
+//
+// On each locale, update the local cache by fetching data from neighbors.
+//
+proc StencilArr.naiveUpdateFluff() {
   coforall i in dom.dist.targetLocDom {
     on dom.dist.targetLocales(i) {
       ref myLocDom = locArr[i].locDom;
```
vasslitvinov: I wonder if it is beneficial to cache locArr[i] (assuming it's just a pointer).

modules/dists/StencilDist.chpl
```diff
@@ -1436,25 +1506,144 @@ proc _array.updateFluff() {
...
+// most beneficial when the communicated region would require many GETs in
+// the naive approach. For example, Let's say we have a 10x10 domain with
+// a fluff of (1,1). Our wholeFluff domain would be {0..11, 0..11}. When we go
+// to communicate the region {1..10,1..1}, the naive method requires 10 GETs or
+// PUTs, whereas the packed method requires just one.
+//
+// On each locale:
+// 1) Repeat the following for each region of data that other locales will
+//    want:
+//    to fetch:
```
vasslitvinov: "want to fetch" go together?

modules/dists/StencilDist.chpl
```diff
@@ -1436,25 +1506,144 @@ proc _array.updateFluff() {
...
+//       buffer that other locales will read from.
+//    b) Write 'true' to an atomic flag on the local that intends to read from
+//       the buffer that was just populated.
+// 2) Once we have prepared our local data for reading, repeat the following
+//    for each region of data that *our* cache wants to update:
+//    a) Wait for the corresponding atomic flag to be 'true' so that we know
+//       it is safe to fetch the data from the remote locale. Reset the flag to
+//       false afterwards.
+//    b) Bulk-copy the remote buffer into a local buffer
+//    c) Copy elements from the local buffer into the cache
```
vasslitvinov: Nice comments!

modules/dists/StencilDist.chpl
```diff
@@ -1436,25 +1506,144 @@ proc _array.updateFluff() {
...
+//
+proc StencilArr._packedUpdate() {
+  coforall i in dom.dist.targetLocDom {
+    on dom.dist.targetLocales(i) {
+      var myLocDom = locArr[i].locDom;
```
vasslitvinov: Like above, consider caching locArr[i].

modules/dists/StencilDist.chpl
```diff
@@ -1436,25 +1506,144 @@ proc _array.updateFluff() {
...
+//       the buffer that was just populated.
+// 2) Once we have prepared our local data for reading, repeat the following
+//    for each region of data that *our* cache wants to update:
+//    a) Wait for the corresponding atomic flag to be 'true' so that we know
+//       it is safe to fetch the data from the remote locale. Reset the flag to
+//       false afterwards.
+//    b) Bulk-copy the remote buffer into a local buffer
+//    c) Copy elements from the local buffer into the cache
+//
+proc StencilArr._packedUpdate() {
```
vasslitvinov: Consider replacing the "_" in _packedUpdate with `private` annotation, for a nicer naming convention. If so, add `private` to some other methods as well?

benharsh: Unfortunately, we do not currently allow private secondary methods like this.

Issue #6230, opened 2017-05-12
Ignored-attributes error in Qthreads header included from re2 prevents Chapel build


dmk42:
### Summary of Problem

Gcc 7.1's stricter checking found a case of attributes being ignored in the qthread.hpp header.  It issues a warning, and our use of `-Werror` turns it into a fatal error.  The following output is observed during a build of Chapel.
```
g++ -c   -MMD -MP -g  -Wall -Werror -Wpointer-arith -Wwrite-strings -Wno-strict-aliasing -Wmissing-declarations -DCHPL_TASKS_MODEL_H=\"tasks-qthreads.h\" -DCHPL_THREADS_MODEL_H=\"threads-none.h\" -DCHPL_WIDE_POINTER_STRUCT -DCHPL_DEBUG -MMD -MP -g -I/home/dmk/chapel/third-party/qthread/install/linux64-gnu-native-flat/include -I. -I../../../../include/localeModels/flat -I../../../../include/localeModels -I../../../../include/comm/none -I../../../../include/comm -I../../../../include/tasks/qthreads -I../../../../include/threads/none -I../../../../include -I../../../../include/qio -I../../../../include/atomics/intrinsics -I../../../../include/mem/jemalloc -I/home/dmk/chapel/third-party/utf8-decoder -I../../../../../build/runtime/linux64/gnu/arch-native/loc-flat/comm-none/tasks-qthreads/tmr-generic/unwind-none/mem-jemalloc/atomics-intrinsics/gmp-none/hwloc/re2/wide-struct/fs-none/include  -I/home/dmk/chapel/third-party/jemalloc/install/linux64-gnu-native/include/jemalloc -I/home/dmk/chapel/third-party/hwloc/install/linux64-gnu-native-flat/include -I/home/dmk/chapel/third-party/re2/install/linux64-gnu-native/include -I/home/dmk/chapel/third-party/re2/re2-src -o ../../../../../build/runtime/linux64/gnu/arch-native/loc-flat/comm-none/tasks-qthreads/tmr-generic/unwind-none/mem-jemalloc/atomics-intrinsics/gmp-none/hwloc/re2/wide-struct/fs-none/src/qio/regexp/re2/re2-interface.o re2-interface.cc
In file included from /home/dmk/chapel/third-party/qthread/install/linux64-gnu-native-flat/include/qthread.h:1661:0,
                 from ../../../../include/tasks/qthreads/tasks-qthreads.h:34,
[...]

PR #6420, opened 2017-06-09
Fix ignored attributes error in Qthreads header

dmk42:
This change fixes issue #6230 by fixing an error in a Qthreads header.

This tested change is being pushed immediately to fix nightly testing, but post-commit review is requested from @ronawho and @gbtitus .  This change or something like it will need to be contributed back to the Qthreads folks.

BACKGROUND:
The problem with the `assert()` macro is that it is executable, so it can only appear inside functions and operate at run time.  C11 and C\+\+11 added a `static_assert()` expression that can appear almost anywhere and takes effect at compile time.

Qthreads wanted to use `static_assert()` in their C\+\+ code, but also wanted to be compatible with earlier versions of the language.  Therefore, they invented their own macro, `QTHREAD_STATIC_ASSERT()` which was cleverly implemented to work in a similar way.  Unfortunately, their implementation provokes a warning from the C\+\+14 checking done by gcc 7.1, and our use of \-Werror turns this warning into an error during the nightly tests, as well as when CHPL_DEVELOPER=true.

The warning is only a problem with very recent compilers, so the fix only has to apply to recent compilers.  The fix presented here uses the real `static_assert()` whenever C\+\+11 or later is being used, thus avoiding the warning that we turn into an error.

Also see issue #6230 for additional detail.

### DIFF
--- a/third-party/qthread/qthread-src/include/qthread/qthread.hpp
+++ b/third-party/qthread/qthread-src/include/qthread/qthread.hpp
@@ -14,7 +14,11 @@
 template <bool> class OnlyTrue;
 template <> class OnlyTrue<true>
 {};
+#if __cplusplus >= 201103
+#define QTHREAD_STATIC_ASSERT(X) static_assert(X, #X)
+#else
 #define QTHREAD_STATIC_ASSERT(X) (void)sizeof(OnlyTrue<(bool)(X)>)
+#endif
 #define QTHREAD_CHECKSIZE(X)     QTHREAD_STATIC_ASSERT(sizeof(X) == sizeof(aligned_t))
 
 #define QTHREAD_CHECKINTEGER(X)  QTHREAD_STATIC_ASSERT(std::numeric_limits<X>::is_integer)
--- a/third-party/qthread/qthread-src/include/qthread/qthread.hpp
+++ b/third-party/qthread/qthread-src/include/qthread/qthread.hpp
@@ -15,7 +15,7 @@ template <bool> class OnlyTrue;
 template <> class OnlyTrue<true>
 {};
 #if __cplusplus >= 201103
-#define QTHREAD_STATIC_ASSERT(X) static_assert(X, #X)
+#define QTHREAD_STATIC_ASSERT(X) static_assert((X), #X)
 #else
 #define QTHREAD_STATIC_ASSERT(X) (void)sizeof(OnlyTrue<(bool)(X)>)
 #endif


ronawho:
@dmk42 what changed that this will start causing problems with night?

third-party/qthread/qthread-src/include/qthread/qthread.hpp
```diff
@@ -14,7 +14,11 @@
 template <bool> class OnlyTrue;
 template <> class OnlyTrue<true>
 {};
+#if __cplusplus >= 201103
```
ronawho: What are the odds this actually works under cce?

dmk42:
We're now using gcc 7.1 for nightly testing.

third-party/qthread/qthread-src/include/qthread/qthread.hpp
dmk42: It does work under cce.
The nice thing about this test is that if it fails, it just falls back to the old version.

ronawho:
I still see setup_gcc_nightly using 6.3. I see that we're using gcc 7.1 for our target compiler, but our gen compiler is still much older, so we shouldn't be building qthreads with 7.1, right?

I'm guess I'm confused about the current state of what compilers we're testing. IMO whenever you build a new gcc, it'd be nice to setup a jenkins job to test. e.g. test-linux64-gcc71.bash. That way we can find errors like this earlier, and once we're happy we can switch our target compiler and setup_gcc_nightly compiler at the same time.

third-party/qthread/qthread-src/include/qthread/qthread.hpp
ronawho: Ah, cool

ronawho:
Yeah, please open a PR against qthreads for this before we forget

dmk42:
Qthreads needs to be built with the target compiler so it can link with the generated code.  We actually run into this problem when re2 includes a header which transitively includes qthread.hpp, and re2 is also being built with the target compiler.

ronawho:
Sorry, target compiler is getting a little conflated here. 

For the module, qthreads (and the runtime/third-party) are built with CHPL_TARGET_COMPILER, but we use "gen" compiler version to build them. i.e. we're still using `CHPL_GCC_GEN_VERSION=5.1.0` to build the runtime/third-paty

dmk42:
Ah, right.  I forgot that it goes one level deeper.  We are actually running into this problem when the runtime re2-interface.cc includes a header which eventually leads to qthreads.hpp being included.  The re2-interface.cc is compiled with the target compiler.

ronawho:
But we're still not building the module with gcc 7.1, right? So far as I know the only place gcc 7.1 gets used is when we're testing the module.

Oh.. wait, nevermind. I forgot that we use CHPL_GCC_TARGET_VERSION to build on the whiteboxes

dmk42:
This has now been contributed back to Qthreads in https://github.com/Qthreads/qthreads/pull/55 .

gbtitus:
Long after the fact, but I've taken a look and this seems okay.

PR #6405, opened 2017-06-07
Trivial housekeeping in PRK

e-kayrakli:
Addresses some of the trivial issues in #6162:

- Header comments with contributors
- Dir name updates
- Change `validate` flag to `correctness`

test/studies/prk passes with standard linux64 config

PR #6421, opened 2017-06-09
Fix references to PRK .graph files

ronawho:
PR #6405 uppercased some of the PRK directories, but missed updating the
references to them in GRAPHFILES and ML-GRAPHFILES


ronawho:
@e-kayrakli and @ben-albrecht just a heads up that these needed to be changed.

PR #6423, opened 2017-06-09
Skip LCALS correctness testing in some configurations

daviditen:
Skip valgrind testing because it takes too long and timed out.

Skip linux32, cygwin, CCE because the checksums don't match what is expected
in a few kernels.  I verified that the arrays match what is expected in a
failing kernel on linux32, so I believe it is the checksum calculation itself
that is going wrong.  I think the most likely culprit is that
sizeof(long double) is less than 16 in these configurations, and a long double
is used to calculate the checksums.


ronawho:
No objections from me

daviditen:
OK, thanks.  I'll go ahead and merge then.

PR #6425, opened 2017-06-09
Skip LCALS correctness testing in --baseline testing

daviditen:
Skip LCALS in --baseline testing because it takes too long and times out.


test/release/examples/benchmarks/lcals/LCALSMain.skipif
```diff
@@ -1,8 +1,9 @@
 # This test takes to long for no-local configurations, so skip it for
```
ronawho: While you're editing, want to change "to" to "too"

daviditen: Wow, can't believe I didn't notice that.  Thanks.

PR #6426, opened 2017-06-09
Cordon off the AliasField code

noakesmichael:
Merge a trivial update on my scope-resolve branch to master.

scopeResolve includes a small amount of "remarkable" code to support array aliases in
new expressions.  This code path supports a feature that was deprecated for 1.15 and
is only used in one test in deprecated/

This PR simply extracts two blocks of inline-logic into helper functions, moves these to
the bottom of the file, and adds some comments to clarify what is going on.

Compiled with/without CHPL_DEVELOPER on clang/darwin and gcc/linux64 and ran a
portion of release/ on all 4 configs.  Also tested deprecated/ on all 4 configs.  Then
passed a single-locale paratest


ronawho:
Can we just remove array alias support instead?

PR #6433, opened 2017-06-12
String uses BaseStringType, so add it to ChapelStandard

mppf:
This is a 1-line change to make ChapelStandard more consistent.

### DIFF
--- a/modules/internal/ChapelStandard.chpl
+++ b/modules/internal/ChapelStandard.chpl
@@ -25,6 +25,7 @@ module ChapelStandard {
 
   // Internal modules.
   use CPtr;
+  use BaseStringType;
   use CString;
   use String;
   use ChapelDebugPrint;


mppf:
@noakesmichael pointed out that BaseStringType is private to String, so this makes less sense. And we are not so sure if ChapelStandard ought to include transitively included modules.

mppf:
correction: BaseStringType is declared private, but it is a top-level module; private would only be reasonable if it were an inner module. That should probably be a compilation error on its own. I think the best we can hope for is that `private module` in that context does nothing.

noakesmichael:
On a second look, it’s not actually true that BaseStringType is “private” to String
after all but I think it’s intended to be.


The *file* String.chpl defines 2 top level modules (whatever that really means);
BaseStringType and String.  The former is marked private but it’s not clear what
that should mean.

I am intrigued that everything compiles. String can use BaseStringType and
ChapeStandard was able to use BaseStringType.



Separately I am looking at an issue with using modules that I *think* reflects
a bug / ill-defined behavior.

Mike



On Jun 12, 2017, at 12:28 PM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@noakesmichael<https://github.com/noakesmichael> pointed out that BaseStringType is private to String, so this makes less sense. And we are not so sure if ChapelStandard ought to include transitively included modules.

PR #6435, opened 2017-06-12
Update spec for const ref/value return intent overloading

mppf:
This is the spec change portion of PR #6306.

See issue #6510 for a language design alternative that the review of these changes uncovered.

### DIFF
--- a/spec/Procedures.tex
+++ b/spec/Procedures.tex
@@ -1143,7 +1143,7 @@ defined in ~\rsec{Determining_More_Specific_Functions}.
 \item
 From the set of most specific functions, the compiler determines a best
 function for each return intent as described in
-~\rsec{Determining_Best_Function}. If there is more than one
+~\rsec{Determining_Best_Functions}. If there is more than one
 best function for a given return intent, the compiler will issue
 an error stating that the call is ambiguous. Otherwise, it will choose
 which function to call based on the calling context as described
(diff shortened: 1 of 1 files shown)


spec/Procedures.tex
```diff
@@ -1295,38 +1295,46 @@ \subsection{Determining More Specific Functions}
...
+determined by the following steps:
+
+\begin{itemize}
+\item If $F_1$ is more specific than $F_2$, then $F_1$ is a better match.
+\item If $F_2$ is more specific than $F_1$, then $F_2$ is a better match.
+\item If $F_1$ has a where clause and $F_2$ does not have a where clause,
+then $F_1$ is a better match.
+\item If $F_2$ has a where clause and $F_1$ does not have a where clause,
+then $F_2$ is a better match.
+\end{itemize}
```
vasslitvinov: Let's use "the better function" instead of "a better match" in the four cases above.

spec/Procedures.tex
```diff
@@ -1295,38 +1295,46 @@ \subsection{Determining More Specific Functions}
...
 
+\subsection{Determining Best Function}
+\label{Determining_Best_Function}
+
+Given two functions $F_1$ and $F_2$, the better function is
+determined by the following steps:
+
+\begin{itemize}
+\item If $F_1$ is more specific than $F_2$, then $F_1$ is a better match.
+\item If $F_2$ is more specific than $F_1$, then $F_2$ is a better match.
```
mppf: Is this redundant, since this applies to  only the set of most specific functions?

vasslitvinov: We select the "best" function from the set of "most specific" functions. Currently (before this PR) the spec says "The set of most specific functions is the set of functions that are not more specific than each other [...]." So these two clauses will never be triggered. (Right?)

mppf: I think it is redundant, but only if if selecting the "best" function only occurs in the flow described above. In particular, somebody just scanning the document might look at the section on selecting the "best" function and ignore the overview. Then they would get the right idea if we leave the redundant information in place.

vasslitvinov: This section exists to define "a best function for each return intent" given the set of most specific functions. I propose to simplify and clarify this section as follows:

```
Given the set of most specific functions for a given return intent,
only the following function(s) are selected as best functions:

\begin{itemize}
\item all functions, if none of them contains a \chpl{where} clause;
\item only those functions that have a \chpl{where} clause, otherwise.
\end{itemize}
```

PR #6437, opened 2017-06-12
Fix alphabetization of contributors

bradcray:
Boy are we bad at alphabetizing...  (where I realize that "we" == "I"
for the most part here...)


bradcray:
@lydia-duncan : I believe you reviewed the previous failed attempt at this (PR #6375).  Would you take a look at this renewed attempt to fix what seems clearly broken?

PR #6441, opened 2017-06-13
Spawn pattern for compiling chapel code

Spartee:
Details an example of the spawn module and how is used to create sub-processes within a program. Takes in a file as a command line argument and compiles it using the chpl compiler.


vasslitvinov:
Please be sure to set it up for our testing system (analogously to grep.chpl) before merging.

PR #6442, opened 2017-06-13
Chapel grep

Spartee:
A grep function written in chapel that utilizes chapels parallel constructs to search for matches using the regex module. The simple version sequentially searches a single file provided by the user. The second version takes in no file but rather searches the files in the users current working directory in parallel for a match.


test/patterns/grep.chpl
```diff
@@ -0,0 +1,74 @@
...
+*/
+proc advancedGrep(tofind: string) {
+  
+  // Seaches current working directory ignoring anthing but UTF-8
+  // encoded characters
+  var files = FileSystem.findfiles(Path.curDir);
+  forall file in files {
+    fileGrep(tofind, file);
+  }
+}
```
mppf: seems to be an eof/special character in an odd place

test/patterns/grep.chpl
```diff
@@ -0,0 +1,74 @@
...
+  var toRead = open(errors, file, iomode.r);
+  var r = toRead.reader();
+
+  // Use regex library to compile into searchable regex
+  var regEx = compile(toFind);
+  var line:string;
+  var lineNum = 1;
+
+  // Search while handling errors
+  while(r.readline(line, errors)) {
```
mppf: did you try using these:
http://chapel.cray.com/docs/latest/modules/standard/IO.html#IO.channel.search
?

test/patterns/grep.chpl
```diff
@@ -0,0 +1,74 @@
...
+  r.close();  // close file
+}
+
+
+/*
+ Advanced grep: needs 
+   - string to be searched for only
+ Runs in parallel across any files in the current working directory. 
+*/
+proc advancedGrep(tofind: string) {
```
mppf: I don't think "advanced" is a good term here; maybe recursive or directory or parallel or ... ?

test/patterns/grep.chpl
Spartee: I did, but had trouble parsing the information I wanted to print from the tuple objects it returns without going by line in each file. Suggestions?

test/patterns/grep.chpl
Spartee: fixed this(parallelGrep) and the eof error.

test/patterns/grep.chpl
mppf: I don't think it's necessarily faster, but you could compute the positions in the file of the matches and then separately read the line. Probably most reasonable to leave it the way you had it for now though.

PR #6446, opened 2017-06-13
Move formatted io

mppf:
Makes FormattedIO be a new submodule within IO.chpl.
The reason to do this is to make the documentation clearer since the FormattedIO documentation is quite long. I considered making FormattedIO be a separate standard module, but that effort was hindered by the lack of `private use`.

Passed full local testing.
Trivial, idea discussed with group, but reviewed by @lydia-duncan - thanks!


mppf:
@lydia-duncan - I think this is trivial (and already discussed). The diff is basically useless (since it's just moving a lot of code around) but I'd appreciate if you can have a quick look at it.

bradcray:
> I considered making FormattedIO be a separate standard module, but that effort was hindered by the lack of private use.

Could you provide more detail on this?  I'm curious what the challenges were.

It also seems to me that we should probably document IO.FormattedIO as one of thhe modules that are auto-`use`d.  I believe this involves adding the standard note to the top of its documentation (see IO for an example) and adding it to the list of the modules at the top of the "Standard Modules" page, but @ben-albrecht is the authority here, so I'm tagging him in case I'm forgetting something else.

mppf:
Re private use, the challenge is that there is a raft of extern procs and also internal-to-IO-implementation functions (such as _write_one_internal) that are used both by the rest of IO.chpl and also by FormattedIO.chpl. These would need to go in to a 3rd module that is shared by both IO and FormattedIO and should not be exposed to users - hence 'private use'. Right now they are module private in IO and that only works because they are called from IO or FormattedIO (which is within IO).

PR #6450, opened 2017-06-14
Introduce ForallStmt AST node

vasslitvinov:
Introduce ForallStmt AST node

This introduces node declaration and basic functions for the work in #5295.
This code is not yet intended to be actually used by the compiler.

#6514 outlines our target goal for forall and intent representation and
lowering. This is the first step in that direction.

Pre-rebase individual commits are available from vasslitvinov clone:

960f481 Follow-ups on review feedback
9cf2a00 Brought it closer to the target #6514
30c568a Another edit based on review suggestions
8ea3925 Edits based on review suggestions
67253e6 Added a separate constructor, per reviewer request
6122589 Added comments for clarity
45045dd Introduce ForallStmt AST node
ed38633 Merge pull request #6521 from lydia-duncan/removeDeprecated
...

### DIFF
--- a/compiler/AST/ForallStmt.cpp
+++ b/compiler/AST/ForallStmt.cpp
@@ -1,0 +1,206 @@
+/*
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ForallStmt.h"
+#include "AstVisitor.h"
+#include "passes.h"
+
+/////////////////////////////////////////////////////////////////////////////
+//
+// ForallStmt represents a forall loop statement
+//
+/////////////////////////////////////////////////////////////////////////////
+
+ForallStmt::ForallStmt(bool zippered, BlockStmt* body, ForallIntents* with):
+  Stmt(E_ForallStmt),
+  fZippered(zippered),
+  fLoopBody(body),
+  fWith(with)
+{
+  fIterVars.parent = this;
+  fIterExprs.parent = this;
+  fIntentVars.parent = this;
+  gForallStmts.add(this);
+}
+
+ForallStmt::~ForallStmt() {
+  delete fWith;
+}
+
+ForallStmt* ForallStmt::copyInner(SymbolMap* map) {
+  ForallStmt* _this  = new ForallStmt(fZippered,
+                                      COPY_INT(fLoopBody),
+                                      COPY_INT(fWith));
+  for_alist(expr, fIterVars)
+    _this->fIterVars.insertAtTail(COPY_INT(expr));
+  for_alist(expr, fIterExprs)
+    _this->fIterExprs.insertAtTail(COPY_INT(expr));
+  for_alist(expr, fIntentVars)
+    _this->fIntentVars.insertAtTail(COPY_INT(expr));
+
+  return _this;
+}
+
+void ForallStmt::replaceChild(Expr* oldAst, Expr* newAst) {
+  if (oldAst == fLoopBody) {
+    if (!newAst)
+      fLoopBody = NULL;
+    else if (BlockStmt* newBlock = toBlockStmt(newAst))
+      fLoopBody = newBlock;
+    else
+      // It is caller responsibility to make newAst fit.
+      INT_ASSERT(false);
+
+  } else if (fWith->replaceChildFI(oldAst, newAst)) {
+      // replaceChildFI took care of it
+
+  } else {
+    // We did not find oldAst in our ForallStmt.
+    INT_ASSERT(false);
+  }
+}
+
+// Todo: are these checks done elsewhere?
+static void verifyList(AList& list, Expr* parent) {
+  INT_ASSERT(list.parent == parent);
+  for_alist(expr, list) {
+    INT_ASSERT(expr->list == &list);
+    INT_ASSERT(expr->parentExpr == parent);
+  }
+}
+
+void ForallStmt::verify() {
+  Expr::verify(E_ForallStmt);
+
+  INT_ASSERT(fIterVars.length == fIterExprs.length);
+  if (fZippered) INT_ASSERT(fIterVars.length > 0);
+  else           INT_ASSERT(fIterVars.length == 1);
+
+  verifyList(fIterVars, this);
+  verifyList(fIterExprs, this);
+  verifyList(fIntentVars, this);
+  for_alist(expr, fIterVars)
+    INT_ASSERT(isDefExpr(expr));
+  for_alist(expr, fIntentVars)
+    INT_ASSERT(isDefExpr(expr));
+
+  INT_ASSERT(fLoopBody);
+  verifyParent(fLoopBody);
+  verifyNotOnList(fLoopBody);
+  // should be a normal block
+  INT_ASSERT(fLoopBody->blockTag == BLOCK_NORMAL);
+  INT_ASSERT(!fLoopBody->blockInfoGet());
+  INT_ASSERT(!fLoopBody->isLoopStmt());
+  INT_ASSERT(!fLoopBody->useList);
+  INT_ASSERT(!fLoopBody->userLabel);
+  INT_ASSERT(!fLoopBody->byrefVars);
+  INT_ASSERT(!fLoopBody->forallIntents);
+
+  INT_ASSERT(fWith);
+
+  // Currently ForallStmt are gone during resolve().
+  INT_ASSERT(!resolved);
+}
+
+void ForallStmt::accept(AstVisitor* visitor) {
+/* TODO once AstVisitor is updated:
+  if (visitor->enterForallStmt(this)) {
+    for_alist(next_ast, fIter)
+      next_ast->accept(visitor);
+    // not doing: fIter->accept(visitor);
+    fWith->acceptFI(visitor); // aka visitor->visitForallIntents(fWith);
+    fMBlock->accept(visitor); // includes loopBody
+    visitor->exitForallStmt(this);
+  }
+*/
+}
+
+Expr* ForallStmt::getFirstChild() {
+  return fIterVars.head;
+}
+
+Expr* ForallStmt::getFirstExpr() {
+  if (Expr* iv = fIterVars.head->getFirstExpr())
+    return iv;
+  else if (Expr* ie = fIterExprs.head->getFirstExpr())
+    return ie;
+
+  // we must have found something
+  INT_ASSERT(false);
+  return NULL; //dummy
+}
+
+Expr* ForallStmt::getNextExpr(Expr* expr) {
+  if (expr == fIterVars.tail)
+    return fIterExprs.head;
+
+  if (expr == fIterExprs.tail) {
+    if (Expr* inv = fIntentVars.head)
+      return inv;
+    else
+      return fLoopBody->getFirstExpr();
+  }
+
+  // Should we also descend into fWith?
+
+  if (expr == fIntentVars.tail)
+    return fLoopBody->getFirstExpr();
+
+  return this;
+}
+
+// Is 'expr' an iterable-expression for 'this' ?
+bool ForallStmt::isIteratedExpression(Expr* expr) {
+  return expr->list && expr->list == &fIterExprs;
+}
+
+// Return the enclosing forall statement for 'expr', or NULL if none.
+ForallStmt* enclosingForallStmt(Expr* expr) {
+  for (Expr* curr = expr->parentExpr; curr; curr = curr->parentExpr)
+    if (ForallStmt* fs = toForallStmt(curr))
+      return fs;
+  return NULL;
+}
+
+// If 'var' is listed in this's with-clause with a reduce intent,
+// return its position in the with-clause, otherwise return -1.
+int ForallStmt::reduceIntentIdx(Symbol* var) {
+  ForallIntents* fi = withClause();
+  int nv = fi->numVars();
+  for (int i = 0; i < nv; i++)
+    if (fi->isReduce(i))
+      if (SymExpr* varSE = toSymExpr(fi->fiVars[i]))
+        if (varSE->symbol() == var)
+          return i;
+
+  // Did not see 'var' with a reduce intent.
+  return -1;
+}
+
+GenRet ForallStmt::codegen() {
+  INT_ASSERT(false); // should not be invoked
+  GenRet ret;
+  return ret;
+}
+
+BlockStmt* ForallStmt::build(Expr* indices, Expr* iterator, ForallIntents* fi,
+                             BlockStmt* body, bool zippered) {
+  // TODO
+  return (BlockStmt*)NULL;
+}
--- a/compiler/include/ForallStmt.h
+++ b/compiler/include/ForallStmt.h
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _FORALL_STMT_H_
+#define _FORALL_STMT_H_
+
+#include "stmt.h"
+
+// forall loop statement
+
+class ForallStmt : public Stmt
+{
+public:
+  bool       zippered()       const; // was 'zip' keyword used?
+  AList&     inductionVariables();   // DefExprs, one per iterated expr
+  AList&     iteratedExpressions();  // SymExprs, CallExprs
+  AList&     intentVariables();      // (future) DefExprs of LoopIntentVars
+  BlockStmt* loopBody()       const; // the body of the forall loop
+  // todo: remove this in favor of intentVariables
+  ForallIntents* withClause() const; // forall intents
+
+  ~ForallStmt();
+
+  DECLARE_COPY(ForallStmt);
+
+  virtual void        replaceChild(Expr* oldAst, Expr* newAst);
+  virtual void        verify();
+  virtual void        accept(AstVisitor* visitor);
+  virtual GenRet      codegen();
+
+  virtual Expr*       getFirstChild();
+  virtual Expr*       getFirstExpr();
+  virtual Expr*       getNextExpr(Expr* expr);
+
+  // for the parser
+  static BlockStmt* build(Expr* indices, Expr* iterator, ForallIntents* fi,
+                          BlockStmt* body, bool zippered);
+  // helpers
+  bool isIteratedExpression(Expr* expr);
+  int  reduceIntentIdx(Symbol* var); // todo: replace with LoopIntentVars
+
+private:
+  bool           fZippered;
+  AList          fIterVars;
+  AList          fIterExprs;
+  AList          fIntentVars;  // may be empty
+  BlockStmt*     fLoopBody;    // always present
+  ForallIntents* fWith;   // always present; todo: replace with intentVariables
+
+  ForallStmt(bool zippered, BlockStmt* body, ForallIntents* with);
+};
+
+ForallStmt* enclosingForallStmt(Expr* expr);
+
+inline bool   ForallStmt::zippered() const       { return fZippered; }
+inline AList& ForallStmt::inductionVariables()   { return fIterVars; }
+inline AList& ForallStmt::iteratedExpressions()  { return fIterExprs; }
+inline AList& ForallStmt::intentVariables()      { return fIntentVars; }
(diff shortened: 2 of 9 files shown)


compiler/include/ForallStmt.h
```diff
@@ -0,0 +1,80 @@
...
+  // All pointer fields are always non-null.
+  BlockStmt*     fMBlock;   // Master Block: contains loopBody and defs
+                            // for index variable(s)
+  bool           fZippered;
+  AList          fIter;   // iterable expression(s)
+  ForallIntents* fWith;   // todo: replace with AList of per-variable "clauses"
+};
+
+// serving ForallStmt and forall intents
+bool isDirectlyUnderBlockStmt(const Expr* expr);
```
mppf: I don't see a definition for this one

compiler/include/ForallStmt.h
```diff
@@ -0,0 +1,80 @@
...
+bool isForallStmtIterator(ForallStmt* fs, Expr* expr);
```
mppf: I can't tell from the name of this function what it's purpose is. In what sense can a forall statement be an iterator?

compiler/include/ForallStmt.h
```diff
@@ -0,0 +1,80 @@
...
+ForallStmt* enclosingForallStmt(Expr* expr);
+int reduceIntentIdx(ForallStmt* fs, Symbol* reVar);
```
mppf: Need more explanation of what this function does. It probably needs a better name in addition to an explanatory comment.

compiler/AST/ForallStmt.cpp
```diff
@@ -0,0 +1,257 @@
...
+  INT_ASSERT(!body || body->blockTag == BLOCK_NORMAL);
+
+  if (body) {
+    // We are creating a new node from scratch.
+    fMBlock->insertAtTail(body);
+    convertIterator(fIter, iter, zippered);
+  } else {
+    // Creating a new node when most contents already exist,
+    // ex. when copying an existing node.
+    INT_ASSERT(!iter);
```
mppf: I'd generally prefer to have a different ForallStmt constructor for copying vs creating one during parsing.

compiler/include/baseAST.h
```diff
@@ -546,12 +550,23 @@ static inline const CallExpr* toConstCallExpr(const BaseAST* a)
...
     break;                                                              \
+  case E_ForallStmt: {                                                  \
+    AST_CALL_LIST (_a, ForallStmt,    iterExprs(),  call, __VA_ARGS__); \
+    ForallIntents* fi = ((ForallStmt*)_a)->withClause();                \
+    AST_CALL_STDVEC(fi->fiVars,  Expr,              call, __VA_ARGS__); \
+    AST_CALL_STDVEC(fi->riSpecs, Expr,              call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, iterRec,      call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, leadIdx,      call, __VA_ARGS__); \
+    AST_CALL_CHILD(fi, ForallIntents, leadIdxCopy,  call, __VA_ARGS__); \
+    AST_CALL_CHILD(_a, ForallStmt,    masterBlock(),call, __VA_ARGS__); \
```
mppf: Just a note, we're discussing whether or not these could be simplified to match more closely the fields in ForallStmt.

mppf:
@vasslitvinov - could you describe (maybe in a comment somewhere) how this start at ForallStmt is related to improved representation for for and coforall loops? E.g. ForStmt/CoforallStmt? I think it would be worth taking a minute to think about how these Loop Statement nodes might be related. In particular, one might wish to have a hierarchy with a base class allowing inspection of what is iterated and of the loop body.

I think if you go through the exercise of imagining such a hierarchy, you might be able to determine a more reasonable naming strategy for parIdxVar / parIdxCopyVar since e.g. a for loop won't have shadow variables at all. On the matter of shadow variables, I'm also a little bit confused as to why the ForallStmt AST node represents the shadow variables at all. Shouldn't they be constructed, respecting the forall intents, when the ForallStmt is lowered? Rather than being constructed during parsing? (Isn't such early "lowering" what we are trying to move away from anyway?)

compiler/include/ForallStmt.h
vasslitvinov: Thank you for catching it. Removed.

compiler/include/ForallStmt.h
vasslitvinov: This function tells whether 'expr' an iterable-expression for 'fs'. 52f91f0 adds a corresonding comment.

compiler/include/ForallStmt.h
vasslitvinov: If 'var' is listed in the fs's with-clause with a reduce intent, this function returns its position in the with-clause, otherwise -1. 52f91f0 adds a corresonding comment.

vasslitvinov: I welcome suggestions for a better name.

compiler/include/baseAST.h
vasslitvinov: In other words, how about simply calling `AST_CALL_CHILD` on the `fWith` field (via withClause() accessor, given that the field is private)? Instead of saving it in `fi` and going over each of fi's fields individually?

The answer is that we cannot/should not define AST_CALL_CHILD on the fWith field -- because it contains a ForallIntent object, which is not in the BaseAST hierarchy. I can define a AST_CALL_WITH_CLAUSE macro instead and use it here, if that is preferable.

One of my TODOs is to bring ForallIntent into the BaseAST hierarchy, which will make fWith be an AList of those things. That will resolve this issue.

compiler/AST/ForallStmt.cpp
vasslitvinov: Added in 58955fa

compiler/include/ForallStmt.h
vasslitvinov: ... in 52f91f0

vasslitvinov:
Added a discussion of a potential hierarchy in 52f91f0.

Re parIdxVar / parIdxCopyVar - these are implementation artifacts, not the variables that the user wrote. I left them as-is from the current implementation to reduce the scope of this initial change. That also explains why they are constructed during parsing.

compiler/include/baseAST.h
mppf: Why is it hard to bring ForallIntent into the BaseAST heirarchy? That seems like a trivially simple change to me.

compiler/include/ForallStmt.h
mppf: This sounds to me like a method that would only be useful during lowering of the forall stmt. Let's leave it out of this PR (you could move it to forall stmt lowering in your bigger branch).

Am I wrong about that?

If it's generally useful, I think it should be a method on ForallStmt.

compiler/include/ForallStmt.h
mppf: In that case I think it has a confusing name. What about `ForallStmt::isIterableExpression(Expr* expr)` ?

I.e. change the name and make it a method?

compiler/AST/ForallStmt.cpp
mppf: It didn't really help so much.. I was hoping/expecting the copy constructor to have 1 argument but perhaps that is not easy with the various BaseAST macros.
If a 1-argument copy constructor is unreasonable, it probably makes most sense to:
 * have a constructor with 1-argument per field
 * have a `ForallStmt::build` method similar to how `TryStmt.h` works
 * mark the constructor as private (either really private or with a comment)

mppf:
@vasslitvinov - thanks for your changes so far and discussion. I'm going to ask you to go further but please don't be discouraged. I think much of the hard work is about what is transformed where and here I'm going to ask you to make more changes to the representation. (And a little bit of what is transformed where...)

Re parIdxVar / parIdxCopyVar, I'm having trouble buying the argument that it's worthwhile to "reduce the scope" of this sequence of PRs because it's already pretty big. You're adding new lowering for ForallStmts. I'd rather have the new code work the way we want it to than add it now and have to go in and rewrite it later.

Generally speaking, I think this PR needs to put more effort into getting the AST representation "right" at least from an interface point of view. In this I think parIdxVar and parIdxCopyVar are the worst offenders.

In particular, let's suppose that we moved the introduction of the shadow variables. Wouldn't the ForallStmt AST node match what you described as ideal:

```
// properties and corresponding methods:
//   zippered             -- zippered()
//   index variables      -- indexVars()
//   iterable expressions -- iterExpr()
//   with clause          -- withClause()
//   loop body            -- loopBody()
//
```

?

And anyway why does the parser have any reason to even know shadow variables exist? They are a detail of how we implemented forall intents.

Moving the introduction of the shadow variables from the parser to ForallStmt lowering does not seem to me to be a conceptually difficult change to make. Yes, you will have to update code, but I'd bet it's not even that much.

As a separate, but related matter, right now we have a method `ForallStmt::iterExprs()` that leaks the representation of the iterable expressions (that it is an AList). It would not be possible to write this method if they were not stored in an AList (because AList is intrusive). Can you justify always needing it to be stored as an AList? Or could that method return a std::vector? Or maybe the numIterExprs/getIterExpr(int) is sufficient?

Likewise `ForallStmt::masterBlock()` leaks the internal representation. I don't think having a public method that locks in the internal representation is better than using public fields with a private name (e.g. starting with _). Either way such a method needs to have a name that hints it is only for use in the BaseAST classes. Or, maybe we can start to use C++ `friend` functions or something. These are all fundamentally a way to communicate with other developers, since obviously any of us could go in and make private fields public or create public accessor methods for them.

Personally I would prefer the simple convention of making such fields public and marking them with a leading `_`. This is the approach we took in TryStmt.h. I think it's just making it worse to add public methods that look like part of the stable API when they are in fact about the implementation details.

The last potential difference I see between your current ForallStmt and the above comment (supposing that is our ideal design) has to do with tuples.
Does `indexVars` or `iterExprs` return a tuple? Or an AList of SymExprs? Or something else? Or will there be `getNum...` `get...(int)` like we have now? Why should `indexVars` and `iterExprs` be different in this regard?

Summing up, since
 * this change is mainly to reduce the cruftiness of the compiler;
 * as far as I know, nothing is waiting on it right now; and
 * we'll need to build upon it for `for` and `coforall` -
I think it's important to try to get the AST representation `right`. Let's try a little harder at that.

mppf:
Also, adding to my last remark, in case it's not obvious. I think we need to talk more about how index vars and iterable expressions are represented before you try to make code updates in that area.

mppf:
@noakesmichael suggests out that `ForallStmt::masterBlock()` is better than having public fields; so to satisfy both of us you'd have to mark such methods that depend closely on the implementation as "implementation internal" even though they will be public for BaseAST to use them.

mppf:
Nonetheless we should immediately concern ourselves with what the intended interface for the new AST node is exactly and then figure out how to support that in the code.

noakesmichael:
On Jun 15, 2017, at 8:11 AM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@noakesmichael<https://github.com/noakesmichael> suggests out that ForallStmt::masterBlock() is better than having public fields;

I didn’t intend to appear to say this and I certainly don’t believe this.
To be crystal clear I do not think that ForallStmt::masterBlock()
should exist at all

Mike




so to satisfy both of us you'd have to mark such methods that depend closely on the implementation as "implementation internal" even though they will be public for BaseAST to use them.

compiler/include/baseAST.h
vasslitvinov: "Trivial" is a bit of overstatement. "Straightforward" may be a better adjective. :)
That change is orthogonal to this PR. OK to do it after #5295, to streamline
management of my code changes?

compiler/AST/ForallStmt.cpp
vasslitvinov: Marked it private in bee0328

compiler/include/ForallStmt.h
vasslitvinov: Good suggestion. Implemented in bee0328

compiler/include/ForallStmt.h
vasslitvinov: Without claims of generality, I use it in normalize.
Switched to a method in 2213438

vasslitvinov:
parIdxVar / parIdxCopyVar are removed in bee0328

vasslitvinov:
Re iterExprs():

I keep iterable expressions in an AList because AList is designed
specifically for use with Expr nodes.
One direct benefit is efficiency of checking whether a given expression
is an iterable-expr in a forall statement.

We could consider keeping iterable expressions in some primitive CallExpr,
although that sounds less fitting to me and will cause more trouble
when dealing with normalization.

I do not see a benefit for having the iterExprs() accessor convert
it to a std::vector.

Are numIterExprs/getIterExpr(int) sufficient?

Yes, in Turing-completeness sense. The problem is that this makes
certain operations more expensive, for example traversing iterable
expressions in AST_CALL_CHILD. Analogoulsy, ContextCallExpr
"leaks" its options AList instead of having accessors for
individual options.

vasslitvinov:
Re masterBlock:

I am not sure how to avoid exposing it.  I am using it to store DefExprs for index variables, so it will need to be traversed in AST_CALL_CHILD.  I also use it for scaffolding during ForallStmt resolution and lowering. If we would like to keep ForallStmts alive past resolution, I strongly suspect we will need to keep, after resolution, at least some scaffolding at least initially, so we most likely will need some kind of BlockStmt for them.

One alternative to consider is placing those DefExprs on a special list, analogously to what we do for a FnSymbol's formals.  I think that this is ultimately what we want for all loops with index variables.  However, that will probably necessitate creating a new Symbol class, a sibling to ArgSymbol. And it will still expose representation in one way or another.

mppf:
@vasslitvinov - thanks for making updates. I think there are mainly 3 points to discuss at the moment:

1) About the nested block structure and DefExprs for index variables.
2) About the types returned by indexVars and iterExprs.
3) About whether or not returning an AList is reasonable

mppf:
> Analogoulsy, ContextCallExpr "leaks" its options AList instead of having accessors for individual options.

ContextCallExpr stores the options AList in a public field (so AST_CALL_CHILD can work) but the rest of the interface does not "leak" the representation. For example, if we changed to having individual fields for each of refCall, valueCall, constRefCall, code using ContextCallExpr will still work with set... and get... calls. There is not an API method that returns an AList.

Of course, one issue with ContextCallExpr is that the *order* in which resolution traverses these calls matters and that is connected (perhaps unnecessarily) to the representation in the AST node.

> I do not see a benefit for having the iterExprs() accessor convert it to a std::vector.

It would allow us to change the implementation of ForallStmt without changing all of the code that calls it. What if, for example, we decided to represent iterExprs as a tuple?

However, I'm somewhat convinced by an argument that iterExprs are most reasonably stored in an AList.

What AST_CALL_CHILD does with it doesn't have much weight for me since we can make it work with whatever AST node API we are making (and we probably need to remove AST_CALL_CHILD at some point anyway).

mppf:
About the nested block structure and DefExprs for index variables:

> Re masterBlock:
>
> I am not sure how to avoid exposing it. I am using it to store DefExprs for index variables...
>
> One alternative to consider is placing those DefExprs on a special list, analogously to what we do for a FnSymbol's formals. I think that this is ultimately what we want for all loops with index variables. However, that will probably necessitate creating a new Symbol class, a sibling to ArgSymbol. And it will still expose representation in one way or another.

Exposing representation is not necessarily bad. It's exposing representation that we know is implementation detail (and not fundamental to the type) that is bad.

In this case, I think it would be a great improvement to (say) create an AList that directly stores, in some form, the iterable expressions. And to store that AList directly as a field in ForallStmt. The two approaches that come to mind are an AList of DefExprs (which you suggested) or an AList of SymExprs (which I have seen in some other loop structures).

Let's talk a little bit more about storing an AList of DefExprs for the index variables. What concerns are there about such an approach?

 * It seems to me the nested block structure you had was to enable certain scoping operations. So perhaps it would require some changes to e.g. scopeResolve?
 * Another element of the nested structure you have is to cause a particular traversal order. But that can still be provided by AST_CALL_CHILD / AstVisitor implementation.

I don't understand why you would need a different AST type other than VarSymbol to store these index variables. ArgSymbol is different because arguments are really different from local variables, and that's an arguably fundamental difference.

While we might argue that index variables are fundamentally different from other variables, I don't think they are different enough to justify a different AST node (e.g. IterationVariableSymbol).

vasslitvinov:
Re iterExprs: I remember struggling with the question of whether it should be an AList or a tuple. A tuple would be more traditional for our compiler to pass around (we do not pass AList references around). However, a tuple meant to me a CallExpr, which, to me, did not convey the essence of what is going on.

I suggest going with AList for now. If/when we decide to switch to tuple representation, it will be an easy change - I can do it at that time.

vasslitvinov:
Re masterBlock: Sure, I can move DefExprs from "master block" into a separate AList. That's a good suggestion.

I prefer to contain the DefExprs within ForallStmt, because that's what they are part of. If we put DefExprs outside, then we are creating additional dependencies, we may want an additional required BlockStmt around ForallStmt, etc. That does not sound in line with our goals.

The concept of "master block" still makes sense to me because it will allow us to distinguish the scaffolding that's needed during lowering from the loop body proper.

If we are messing with scopeResolve, I can make the experiment of having a separate BlockStmt for scaffolding. It will contain the BlockStmt for loop body conceptually, just not in the representation.

What do you think?

vasslitvinov:
Re separate AST node for index variables: actually they are different from regular variables in that they have associated forall intents. Right now we keep forall intents as a separate structure from their variables. By contrast, ArgSymbols keep their intents and possibly initial values in their class.

While I am not pushing for creating IterationVariableSymbols, or perhaps IntentVariableSymbols, it **would** make sense to me if we did so.

vasslitvinov:
Re indexVars: I do not represent those explicitly at the moment. When we do... we may need a combination of a set of DefExprs (a BlockStmt with those, an AList with those, or something) plus an Expr that represents a tree of SymExpr - because we need to represent the various ways index variables may be grouped in a tuple:

```chpl
forall singeVar in ...
forall (v1,v2) in ...
forall ((v1,v2), (_,(v3,v4)), v5) in ...
etc.
```

Interestingly, this argument does not apply to iterable expressions because those are organized in a sequence of 1 or more exprs, there is no nesting.

Clarification: the defExprs we discuss above in the context of masterBlock are for implementation variables like today's chpl__leadIdx and chpl__leadIdxCopy.

mppf:
Re indexVars, to represent the various things like

```chapel
forall singeVar in ...
forall (v1,v2) in ...
forall ((v1,v2), (_,(v3,v4)), v5) in ...
etc.
```

what makes sense to me is to have an AList/whatever storing the loop iteration variable for each thing being iterated over, and then to generate code to pack/unpack tuples as needed. In other words, we would represent something like:

``` chapel
forall ((v1,v2), (_,(v3,v4)), v5) in zip(A, B, C) {
}
```

as
``` chapel
forall (a, b, c) in zip(A, B, C) {
  var/ref ((v1,v2), (_,(v3,v4)), v5) = (a,b,c)
}
```

I think that having 1 iteration variable per zippered thing is the
"natural" representation here and that everything else is just a tuple
pack/unpack operation within the loop body.

compiler/include/ForallStmt.h
```diff
@@ -0,0 +1,78 @@
...
+
+// forall loop statement
+
+class ForallStmt : public Stmt
+{
+public:
+  bool       zippered()       const; // was 'zip' keyword used?
+  AList&     inductionVariables();   // DefExprs, one per iterated expr
+  AList&     iteratedExpressions();  // SymExprs, CallExprs
+  AList&     intentVariables();      // DefExprs of LoopIntentVars with intents
```
mppf: Let's update this comment to say future: or (future)

mppf: Also, a nit, but is there a reason not to make inductionVariables, iteratedExpressions, and intentVariables have const this and return a const reference?

compiler/AST/ForallStmt.cpp
```diff
@@ -0,0 +1,204 @@
...
+    if (!newAst)
+      fLoopBody = NULL;
+    else if (BlockStmt* newBlock = toBlockStmt(newAst))
+      fLoopBody = newBlock;
+    else
+      // It is caller responsibility to make newAst fit.
+      INT_ASSERT(false);
+
+  } else if (fWith->replaceChildFI(oldAst, newAst))
+    ; //good
```
mppf: a nit, I think I'd prefer to use curly braces in this situation instead of a bare ;

compiler/AST/ForallStmt.cpp
```diff
@@ -0,0 +1,204 @@
...
+*/
+}
+
+Expr* ForallStmt::getFirstChild() {
+  return fIterVars.head;
+}
+
+Expr* ForallStmt::getFirstExpr() {
+  if (/*Expr* iv =*/ fIterVars.head->getFirstExpr())
+    INT_ASSERT(false); // todo: instead return iv;
```
mppf: I'm confused about this assertion? Why is this todo still here?

compiler/AST/ForallStmt.cpp
```diff
@@ -0,0 +1,204 @@
...
+    return fIterExprs.head;
+
+  if (expr == fIterExprs.tail) {
+    if (Expr* inv = fIntentVars.head)
+      return inv;
+    else
+      return fLoopBody->getFirstExpr();
+  }
+
+  // Should we also descend into fWith?
```
mppf: I don't actually know the answer to that. If it isn't necessary right now, OK. I would expect that once intents are real AST nodes they will need to be visited here (but you already have that with fIntentVars here).

compiler/AST/ForallStmt.cpp
```diff
@@ -0,0 +1,204 @@
...
+        if (varSE->symbol() == var)
+          return i;
+
+  // Did not see 'var' with a reduce intent.
+  return -1;
+}
+
+GenRet ForallStmt::codegen() {
+  INT_ASSERT(false); // should not be invoked
+  return (GenRet)0;
```
mppf: I'm not so sure what (GenRet)0 actually does. I'd normally write
```
GenRet ret;
return ret;
```

but that is a nit since the compiler should quit before getting there anyway.

mppf:
Hi @vasslitvinov - thanks for all your work on improving the new AST node. I already offered some minor feedback on this version but I think this is more or less ready to go in. I don't need to see it again before you merge it.

Two requests:
 * make sure testing passes (even though the new code should not be run)
 * mention issue #6514 as documenting the goal of forall and intent lowering in the PR and merge messages.

Thanks!

compiler/AST/ForallStmt.cpp
vasslitvinov: Right. #5295 works without descending into fWith. This is here as a reminder of sorts.

PR #6452, opened 2017-06-14
Adjusted a call to chpl_glob in the modules.

vasslitvinov:
Replaced pattern:c_string with pattern.localize().c_str()
for consistency with other uses of chpl_glob in FileSystem module.


mppf:
Looks good, thanks.

PR #6453, opened 2017-06-14
Added two verify() variants.

vasslitvinov:
verify(expectedTag) and verifyParent() factor out common operations
in our verify() methods.

To show their utility, I applied them where possible in expr.cpp.
A nice effect is a reduction in code size.

Given that it is an internal error that is raised if there are issues,
I stayed with minimalistic error messages (in agreement with Michael).

### DIFF
--- a/compiler/include/expr.h
+++ b/compiler/include/expr.h
@@ -41,6 +41,9 @@ public:
   virtual QualifiedType qualType();
   virtual void    verify();
 
+  void verify(AstTag expectedTag); // ensure tag is as expected, then verify()
+  void verifyParent(const Expr* child); // verify proper child->parentExpr
+
   // New interface
   virtual Expr*   copy(SymbolMap* map = NULL, bool internal = false)   = 0;
   virtual void    replaceChild(Expr* old_ast, Expr* new_ast)           = 0;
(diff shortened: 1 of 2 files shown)


compiler/include/expr.h
```diff
@@ -40,6 +40,8 @@ class Expr : public BaseAST {
   virtual bool    isStmt()                                           const;
   virtual QualifiedType qualType();
   virtual void    verify();
+          void    verify(AstTag expectedTag);
+          void    verifyParent(const Expr* child, const char* field);
```
mppf: Could you comment these methods here to indicate their purpose?
In particular I can see what the AstTag version is doing from the source code, but I don't understand the purpose of `field` in verifyParent. Given that it will raise an INT_FATAL, why does it need a customizeable error message at all?

vasslitvinov:
Sure I will add comments.

Are you suggesting to skip the `field` argument? I'd be OK with that.

mppf:
Looks good, but please repeat the comments in the .h file. I know that the .h file doesn't have much comments now but that's a flaw to be addressed rather than a pattern to emulate. After that you're good to merge once you're satisfied with testing. Thanks!

PR #6455, opened 2017-06-14
Added testing and changed file location for grep pattern

Spartee:
Added the requested for tests and deleted the grep.notest file. Changed around the file structure as suggested to keep the patterns folder cleaner.


mppf:
Preston's review is sufficient. As an aside, if I were to actually *use* such a parallel grep, I'd want it to recursively descend into directories. E.g. with http://chapel.cray.com/docs/latest/modules/standard/FileSystem.html#FileSystem.findfiles . I would also expect the output to resemble more closely what `grep` does, but that is not a real big deal right now.

PR #6456, opened 2017-06-14
Spawn pattern with added testing

Spartee:
Added testing to chplCompile and changed file location to clean up the patterns directory.


test/patterns/cltools/fileCompile/chplCompile.chpl
```diff
@@ -0,0 +1,54 @@
+
+/*
+A tool written in chapel to compile a chapel program and to demonstrate the use of the 
+spawn module and how it is used to create subprocesses within a chapel program. 
+*/
+module chplCompile {
+
```
psahabu: The usual convention is to indent everything inside of a module. However, since this is a test, you could actually delete this line and the matching curly brace -- the compiler will auto-generate the module name from the filename.

test/patterns/cltools/fileCompile/chplCompile.chpl
```diff
@@ -0,0 +1,54 @@
...
+use Spawn;
+
+//Command to be executed. Starts with chpl to signal the chapel compiler.
+var command: string = "chpl";
```
psahabu: A top level `var` is okay, but it might be better to pass around a command `var` with `"chpl "` as a `const` to prepend to the command.

test/patterns/cltools/fileCompile/chplCompile.chpl
```diff
@@ -0,0 +1,53 @@
...
+  cat_command(args);    // concatenate command 
+  writeln(command);     // print command for user
+  spawn_sub(command);   // spawn command to compile file
+}
+
+
+/*
+Concatenates the given arugments to be sent to a subprocess
+*/
+proc cat_command(args: [] string) {
```
psahabu: Sorry, I should've clarified. I meant returning `command` from this function and storing it in `main()`, rather than at the top-level.

PR #6457, opened 2017-06-14
Use a temporary domain to track indices to remove in associative &= operator

daviditen:
Use a temporary domain to store the indices to remove in the &= operator
so that we don't remove indices while iterating over them.

Added a test that does domain intersections of random domains that failed
before this PR with an index out of bounds error.

Passed linux64 paratest with --verify.


daviditen:
Thanks!

PR #6462, opened 2017-06-14
Add the #chapel140 programs

bradcray:
Here, I'm adding the programs that we've been tweeting as 140-character (#chapel140) programs for safekeeping.


bradcray:
@daviditen : Would you like to review this one?  As contributors, @tmacd9 , @mppf , @cassella , and @ben-albrecht may also want to take a peek.

test/studies/twitter/sleepsort.chpl
```diff
@@ -0,0 +1,11 @@
+use Time;
```
daviditen: I have a feeling this program might turn into a maintenance headache when some system decides that sleeping for 3 seconds is a fine reaction to a sleep(1) call.  Hopefully not though.

daviditen:
Looks good!

test/studies/twitter/sleepsort.chpl
bradcray: Could be.  I'll move to a smaller set of more distantly spaced integers if it does cause problems.  I'm also wondering whether we'll have any portability problems with the file.length logic in the check140*.chpl tests or generally with survey.chpl.

PR #6463, opened 2017-06-14
Reenable gcc value range propagation

dmk42:
Gcc's value range propagation had been turned off because of an optimizer bug in gcc 5.  This change reenables it for later versions of gcc.  This fixes many false positive warnings from gcc 7.

Tested with gcc 6.3 and 7.1.


bradcray:
This looks good to me -- thanks!

PR #6464, opened 2017-06-14
added notest file

Spartee:
added .notest file for the samplefile included in the last commit(#6456)


cassella:
Should there be a CLEANFILES too?

psahabu:
@Spartee:
@cassella has a good point that I hadn't considered. When this test is run, the compiled `samplefile` is left in the directory and it appears when I do a `git status`. If you could create a CLEANFILES file (see test/io/ferguson/CLEANFILE for an example) and put it in this directory, that'd help keep that tidy.

PR #6465, opened 2017-06-15
Improvements to the way chpl uses clang and llvm

mppf:
* Enables --llvm-print-ir to work with more parts of LLVM optimization pipeline. The new options are named after the "extension points" in PassManagerBuilder. For example, optimizer-last or scalar-optimizer-late. This works by adding our own LLVM FunctionPass that dumps IR for the requested function.
* Changes the way the Chapel compiler uses clang. Instead of generating a .bc file and then linking it with a call out to clang, now it generates a .o file. This is helpful (at least in the near term) because all the optimizations now occur in the chpl binary instead of in the separate clang instance. That makes it easier to reason about.
* Changes the way that the Chapel compiler creates command line arguments for the clang instance it uses. It now uses the clang Driver to process the arguments since some arguments are handled differently by the driver. The main purpose of this change is to enable `-march=native`.
* Adds support for --mllvm to set llvm optimizations (e.g. `--mllvm -force-vector-width=16`).

Testing:

- [x] full local testing
- [x] full local CHPL_LLVM=llvm testing
- [x] full local testing with --llvm


mppf:
Closing for now since I know it needs more work. We can still discuss.

mppf:
@dmk42 - I think this PR improves the situation. I'm hoping you'll review it and/or discuss it with me.

dmk42:
Only two issues stick out to me.
* Missing copyright on llvmDumpIR.cpp, preventing travis test from passing
* If an option in llvmFlags ever contains a space, it will be parsed incorrectly
The latter probably doesn't matter right now, but we should keep an eye on it.

OK to commit after copyright corrected.

coodie:
I think it is necessary to test supported targets, because (as I've noticed in my e-mail) the issue with `EmitBackendOutput` function was related to `TargetRegistry`. I suppose we want to test whether TargetRegistry contains the most important targets to us (the ones which chapel is supposed to support). There might be added extra compile option to llvm for printing all supported llvm targets, and then there would be a check in the test whether the list contains the most relevant to us.
Here is easy way to print contents of registry:
```
  for(auto target : llvm::TargetRegistry::targets())
    std::cout << target.getName() << std::endl;
```

The target registry is filled in these lines:
```
  llvm::InitializeAllTargets();
  llvm::InitializeAllTargetMCs();
  llvm::InitializeAllAsmPrinters();
  llvm::InitializeAllAsmParsers();
```

mppf:
@coodie - are you proposing we have a compiler flag to print out supported targets? Or something else? Not sure what you mean by "test supported targets".

coodie:
@mppf 

That is one idea to test targets that are supported by llvm - by adding extra compiler flag to print out these and check what's being printed. We can also add some asserts inside code, or have some simpler unit tests to avoid adding such compiler option (I don't think this is very useful anyway).

mppf:
2 failures (only) with CHPL_LLVM=llvm:

compflags/coodie/llvmBackendCCflagsSplitBySpaces is getting an error
  clang: error: argument unused during compilation: '-mllvm -use-cfl-aa'

runtime/sungeun/chpl-env-gen is getting an error about llvm not matching runtime.

PR #6467, opened 2017-06-15
Squash false positive warnings about string overflows

dmk42:
Gcc 7 has some false positive warnings on the generated code.  Squash them.

Trivial, not reviewed.

### DIFF
--- a/make/compiler/Makefile.gnu
+++ b/make/compiler/Makefile.gnu
@@ -164,6 +164,13 @@ SQUASH_WARN_GEN_CFLAGS += -Wno-tautological-compare
 endif
 
 #
+# Avoid false positive warnings about string overflows
+#
+ifeq ($(shell test $(GNU_GPP_MAJOR_VERSION) -eq 7; echo "$$?"),0)
+SQUASH_WARN_GEN_CFLAGS += -Wno-stringop-overflow
+endif
+
+#
 # 2016/03/28: Help to protect the Chapel compiler from a partially
 # characterized GCC optimizer regression when the compiler is being
 # compiled with gcc 5.X.


bradcray:
Hi @dmk42 -- Can you say more about this one?  E.g., What were the warnings and why were they false positives?

dmk42:
Gcc 7 adds some really impressive new checks, including checks to catch buffer overflows, when possible at compile time.  It also checks the value ranges of integers, and knows that an unsigned value could be very large if it originally came from a signed value that might have been negative.

When building the Chapel compiler and runtime, this is useful to us.  However, there are some unfortunate interactions that cause false positives when users compile their Chapel code.  If in the generated code, the unsigned value for the size of a buffer passed to `memcpy()` originally came from a signed integer, gcc complains to the user trying to compile a Chapel program that the `memcpy()` might overflow its buffer.

One example of this is when the `Locales` array is localized.  There is a call to `chpl_comm_get()` which checks to see if the array is already local so it can do a local copy instead of a network get.  In the case of a local copy, it calls `memcpy()` with a size expression that is derived from `numLocales`.  However, `numLocales` is signed, so gcc thinks that the unsigned size passed to `memcpy()` might be excessively large because `numLocales` has the capability of holding a negative number.  It complains to the poor user who might not even understand C.

This particular case could be fixed by changing `numLocales` to be unsigned.  However, Chapel's use of signed integers in situations where unsigned would be better is pervasive.  The real fix would be to go all the way through Chapel and make it more unsigned-friendly.  In the meantime, we cannot have the backend compiler complaining to Chapel users about the generated code, so this PR disables those warnings for the generated code only.

bradcray:
Thanks!  I'm similarly disinclined to change such values to unsigned to work around such warnings.

dmk42:
Just to clarify, the purpose for changing things to unsigned would not be to work around warnings.  That would just be a bonus.  The real purpose would be to make the code clearer, both to optimizers and to humans.

PR #6470, opened 2017-06-15
Add options to print unused functions

daviditen:
Add options --print-unused-functions and --print-unused-internal-functions.
The first prints the names and source locations of any unused functions in
the user's program.  The second is a developer option, and prints names
and source locations of unused functions across the entire program,
including in the modules.

The printing happens near the end of the function resolution pass, right
before unused functions are pruned away.

Added a small test that contains an unused function and uses the
--print-unused-functions flag.

Update the man page and userhelp test to include --print-unused-functions.

Passed a standard linux64 paratest.


ronawho:
@daviditen can this be closed until it's ready to merge?

daviditen:
I'd prefer to leave it open and get it merged ASAP.

ronawho:
>I'd prefer to leave it open and get it merged ASAP.

Works for me (I didn't realize this was something you were trying to get into the release)

daviditen:
I brought this up to date with master and re-tested with a clean Linux64 paratest two days ago.

compiler/resolution/functionResolution.cpp
```diff
@@ -8249,6 +8253,53 @@ static void cleanupAfterRemoves() {
 }
 
 
+static void printUnusedFunctions() {
+#ifdef PRINT_UNUSED_FNS_TO_FILE
```
lydia-duncan: Was this defined in an earlier commit, or is it just debugging support that got left behind?

test/compflags/diten/unusedFunction.good
```diff
@@ -0,0 +1 @@
+unusedFunction.chpl:3: baz
```
lydia-duncan: I think it would be useful to start the list of unused functions with a label or header, in case other output in generated.  It might also be good to label the end of the list.  Something like
```
These are the unused functions in your program:
unusedFunction.chpl:3: baz
There are no more unused function
```
or
```
These are the unused functions in your program:
unusedFunction.chpl:3: baz
End of unused functions list
```

compiler/resolution/functionResolution.cpp
daviditen: It's useful for running the whole test suite and saving the results to files so we can see what functions are not used by the test suite.  Just define this macro (and maybe also `EXIT_AFTER_PRINTING_UNUSED_FNS`) and run paratest.

lydia-duncan: Would you mind documenting that in the developers documentation or something, then?  (and maybe add a comment pointing to that explanation of its use)  It seems like it would get overlooked otherwise

test/compflags/diten/unusedFunction.good
daviditen: What do you think about indenting the list instead of having an ending message?

```
The following functions are unused:
  unusedFunction.chpl:3: baz
```

lydia-duncan: That works for me!

compiler/resolution/functionResolution.cpp
daviditen: I added some comments about these macros at the top of `printUnusedFunctions`.  Do you think it needs documentation beyond that?  If yes, where would you recommend putting it?

daviditen:
Awesome!  Thank you!

PR #6472, opened 2017-06-15
Fix chplCompile environment issue

benharsh:
Prepend the PATH for this test with the directory containing the chpl binary in case the environment isn't set up for it, as is the case in our nightly testing.


benharsh:
@Spartee : We were on the right track with the .execenv file, it just needed to be a bit fancier to get at the CHPL_HOST_PLATFORM environment variable.

PR #6473, opened 2017-06-15
Add a check for passing "this" to a function in phase1 of an initializer

noakesmichael:
This is a simple update that generates an error if "this" is passed as an actual to a function
in phase1.  Retire the associated future.

Passed start_test on a small portion of release/
Passed single-locale paratest on gcc/linux64


lydia-duncan:
Does this respond appropriately to:

```
proc init(xVal) {
  x = xVal;
  y = this.x;
  super.init();
}
```
?  I think I neglected to make a test with that case

PR #6476, opened 2017-06-15
Update 2 .good files and retire the associated futures

noakesmichael:
Update 2 .good files for initializers and then retire the associated futures

### DIFF
--- a/test/classes/initializers/phase2/parentAccess.good
+++ b/test/classes/initializers/phase2/parentAccess.good
@@ -1,1 +1,1 @@
-{f1 = 1.25, f2 = -1.95, f3 = 1.5625}
+parentAccess.chpl:7: error: Cannot access parent field 'f1' during phase 1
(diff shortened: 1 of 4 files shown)


test/classes/initializers/phase2/parentAccess.good
```diff
@@ -1 +1 @@
-{f1 = 1.25, f2 = -1.95, f3 = 1.5625}
+parentAccess.chpl:7: error: Cannot access parent field 'f1' during phase 1
```
lydia-duncan: It might be more appropriate to update the .chpl file so that it doesn't violate the phase 1 rule with the omitted initialization.  This test was intended to demonstrate that it is okay to rely on a parent field in Phase 2, it was an oversight that it inserted a bad access

noakesmichael: I like this test.  So maybe there's a role for yet another test.

lydia-duncan: That seems reasonable to me!  I always favor more tests :)

PR #6480, opened 2017-06-16
Emit warning suppression flag in the correct order

dmk42:
For a long time, we have been emitting gcc's `-Wno-strict-overflow` flag when compiling the generated code, but we never noticed until gcc 7 that it was placed on the command line in the wrong order.  We had been putting it before `-Wall` which caused it to be overridden.  This change places it after `-Wall`.

This fixes the test `modules/packages/LinearAlgebraJama/TestMatrix` when run with `--fast`.


bradcray:
If it was in the wrong order, getting ignored, and not causing problems, does that suggest that we don't want/need that flag?

dmk42:
No, we need it.  That's why I discovered it was in the wrong order.
Gcc 7.1 optimized a case that earlier versions didn't, which is why I noticed it.

PR #6491, opened 2017-06-16
Initial implementation of default initializers support

lydia-duncan:
This adds a new flag to compilation, --force-initializers, which when
thrown will skip the generation of default constructors for types that
don't define any constructors or initializers.  It will then try to create
a default initializer for that type, and error if the initializer cannot be
created.

We will also now get an error message if a class tries to inherit from
a class which defined an initializer and the child wants to rely on the
default initializer when compiling without --force-initializers.

Limitations:
---------------
At present, I have limited the scope of this flag to *not* impact the
following cases:
- records/unions
- aggregate types found in the internal and library modules
- nested aggregate types
- extern aggregate types
- aggregate types that define an `initialize()` method

This flag will not impact types that have defined a constructor or
an initializer - the former will still get a default constructor (as it
relies on the default constructor being present for functionality),
and the latter will still get neither a default constructor nor a
default initializer.

Implementation:
---------------------
Adds two helper methods to AggregateTypes: needsConstructor() and
wantsDefaultInitializer().  The former is private to AggregateTypes and
only called from the method buildConstructor; it returns true if a default
constructor should be generated for the type, and the method is
recursive in the case of inheritance, resulting in the generation of a
default constructor for classes that inherit from a class which defined a
constructor.  The latter method is public and called by many places when
determining the AST to generate when creating an instance of that
type.  It returns true if a default initializer should be generated for that
type.

Update the checks for if the type relies on initializers to account for when
a default initializer would be generated for that type.

Adds the flag `--force-initializers` and a global boolean to support it.

Adds a function to create a default initializer to the pass
buildDefaultFunctions.  The default initializer will have an argument per
field, which it will use to initialize each field in order.  It will then attempt
to call a parent initializer.  If the parent relies on a default initializer,
it will create and pass in an argument per argument in the default parent
initializer.  If the parent defines a constructor, we will generate an error.
If the parent defines an initializer, we will attempt to call a zero-argument
initializer.  Whether that call is valid or not will be determined at resolution.

At function resolution, if no initializer has been created for the type and we
need a default, generate an error message asking the user to define an
explicit initializer.

Update the man page for this new flag so users can experiment with it.

Testing:
----------
Adds 12 new tests and a future for this feature, and moves the four
dflt-generic-uninit-field error tests into the directory for testing compiler
generated default initializers, so that we verify their behavior when it is
initializers that are generated.  Add a COMPOPTS to that directory to throw
--force-initializers with all of those tests.  Update the expected output of
userhelp.chpl for the new flag.

The new tests:
- verify that an ordinary class with no constructor or initializer will work with
  the --force-initializers flag thrown
- verify that a generic class with no constructor or initializer will work with the
  --force-initializers flag thrown, for all strategies of instance declaration.
- verify that the --force-initializers flag doesn't break ordinary classes that
  define constructors
- verify that a child class which inherits from a parent that defined a constructor
  will throw an error with --force-initializers when trying to create a default initializer
- verify that a child class which inherits from a parent will work when both do not
  define either a constructor or an initializer and --force-initializers is thrown
- verify that a child class which relies on the default initializer and inherits from a
  parent that defines an initializer with no arguments will work when
  --force-initializers is thrown
- verify that a child class which relies on the default initializer and inherits from a
  parent that defines an initializer with arguments will fail when --force-initializers
  is thrown (because we don't know how to call that initializer, so the user should
  specify what that call should look like)
- verify that a child and parent class with no initializers or constructors will fail if
  the grandparent class defines a constructor and --force-initializers is thrown.
- verify that a child, parent, and grandparent chain where no class in the chain
  defines an initializer or constructor works with --force-initializers
- verify that a child and parent class with no initializers or constructors will work
  if the grandparent class defines an initializer with no arguments and
  --force-initializers is thrown.
- adds a future for a child and parent class with no initializers or constructors that
  should fail if the grandparent class defines an initializer with arguments and
  --force-initializers is thrown, but instead matches against the grandparent initializer
  instead of the child default initializer
- verify that a child class which relies on the default initializer and inherits from a
  parent that defines an initializer with no arguments will fail when --force-initializers
  is not thrown

Passed full std testing


lydia-duncan:
@noakesmichael - no rush, especially since testing will be down this weekend.  Also, I think it is best to review all the files at once instead of commit-by-commit

test/compflags/bradc/help/userhelp.good
```diff
@@ -22,6 +22,7 @@ Optimization Control Options:
       --[no-]dead-code-elimination    Enable [disable] dead code elimination
       --fast                          Use fast default settings
       --[no-]fast-followers           Enable [disable] fast followers
+      --[no-]force-initializers       Enable [disable] default initializers
```
mppf: Don't you want to revert this one too, make --force-initializers be a developer only flag, and put tests for it specifically into test/deprecated?

lydia-duncan: I guesssssssss?  I don't want to put the tests into the deprecated directory yet, I think.  I'm not yet sure on the time frame for removing the default constructor code and so might keep it around longer, perhaps as long as the deprecation warning for explicit constructors.  If that's the case, it might be useful for it to be a flag a user could use; if they find a problem with default initializers, we might want to have them see if they still have that problem with default constructors and so might tell them to use that flag.

mppf: I'd keep it developer only and in deprecated/ optimistically, and then consider making it a user-facing flag if we're not where we want to be closer to the release. But that's up to you.

Either way, I'd put *something* in the deprecated directory. I don't think we're obligated to remove everything from test/deprecated each release. If you don't like that, maybe make an issue or something to track removing the flag?

noakesmichael:
If I were handling this I’d be inclined to take the path that Michael proposes.

I hate to try to predict the future but I think we should take a path that
assumes that there won’t be any constructors in MOD_USER by 1.16
and that there will be a simple deprecate message for any new constructors
in at least MOD_USER.

Mike



On Jun 21, 2017, at 2:25 PM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@mppf commented on this pull request.

In test/compflags/bradc/help/userhelp.good<https://github.com/chapel-lang/chapel/pull/6491#discussion_r123370294>:

> @@ -22,6 +22,7 @@ Optimization Control Options:
       --[no-]dead-code-elimination    Enable [disable] dead code elimination
       --fast                          Use fast default settings
       --[no-]fast-followers           Enable [disable] fast followers
+      --[no-]force-initializers       Enable [disable] default initializers


I'd keep it developer only and in deprecated/ optimistically, and then consider making it a user-facing flag if we're not where we want to be closer to the release. But that's up to you.

Either way, I'd put something in the deprecated directory. I don't think we're obligated to remove everything from test/deprecated each release. If you don't like that, maybe make an issue or something to track removing the flag?

test/compflags/bradc/help/userhelp.good
lydia-duncan: Okay, I've got a branch that does this (but it isn't complete yet because I dragged my feet.  I'll have you take a look tomorrow morning when I finish it)

PR #6494, opened 2017-06-17
Delete some user-level SSCA2 classes that had not been

bradcray:
This PR deletes some user-level classes in SSCA2 that had not been before:
* it  flips the sense of DELETE_KERNEL4_DS to true in order to have the
  benchmark clean up after itself when it had not before.  For the perfexecopts
  file, it flips it back to false, just in case this messes up something with the
  timing (even though it's outside of the timed region of the code, I believe).
* it fixes the code in the DELETE_KERNEL4_DS to walk the linked list in
  both directions rather than just one, catching a number of Level_Set
  objects that had previously escaped being freed even if DELETE_KERNEL4_DS
  was set to true.
* it deletes the random number generators that are allocated.

Unfortunately, this doesn't make SSCA2 leak-free, but it does reduce
its leaks significantly.


bradcray:
@sungeunchoi: Could you give me a quick review of this sometime?

bradcray:
Hmm, @sungeunchoi ... I just noticed that test/studies/ssca2/sungeun/ is NOTEST-ed.  Is there any reason not to just blow that directory away, do you think?

sungeunchoi:
Yeah -- just blow it away.

PR #6496, opened 2017-06-19
C2Chapel fix for Enums/Typedef Enums

Spartee:
c2chapel now accepts enums as well as typedef'd enums. The fix is relatively straightforward as c2chapel emits an "extern const" for each enumerated value and an "extern type" for each typedef'd enum.  The test script added to the /test directory details a few different uses of Enums as well as typedef'd enums.

### DIFF
--- a/tools/c2chapel/c2chapel.py
+++ b/tools/c2chapel/c2chapel.py
@@ -346,8 +346,21 @@ def genVar(decl):
 
 def genEnum(decl):
     if type(decl) == c_ast.Enum:
+        if decl.name: 
+            genComment("Enum: " + decl.name)
+        else: 
+            genComment("Enum: anonymous")
         for val in decl.values.enumerators:
             print("extern const " + val.name + " :c_int;")
+        print("\n")
+
+def genTypeEnum(decl):
+    if type(decl.type) == c_ast.Enum:
+        for child in decl.children():
+            for val in child[1].children():
+                for value in val[1].enumerators:
+                    print("extern const " + value.name + " :" + decl.declname + ";")
+        print("\n")
 
 # Simple visitor to all function declarations
 class ChapelVisitor(c_ast.NodeVisitor):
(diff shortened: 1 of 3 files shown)


tools/c2chapel/test/enum.h
```diff
@@ -0,0 +1,32 @@
...
+  FAILED = 2
+};
+
+typedef enum 
+{
+  ENUM_ERROR = 1,
+  ENUM_WVALUES_ERROR = 2,
+  TD_ENUM_ERROR = 3,
+  FUNC_ENUM_ERROR = 4,
+  STRUCT_ENUM_ERROR = 5
```
benharsh: Where are the generated Chapel constants for these?

tools/c2chapel/test/enum.chpl
```diff
@@ -0,0 +1,21 @@
+// Generated with c2chapel version 0.1.0
+
+// Header given to c2chapel:
+require "enum.h";
+
+// Note: Generated with fake std headers
+
+extern const TEST_STATUS_PASSED :c_int;
+extern const TEST_STATUS_FAILED :c_int;
+extern const PASSED :c_int;
```
benharsh: I think we should add a newline after each enum group

tools/c2chapel/test/enum.h
```diff
@@ -0,0 +1,32 @@
...
+  TEST_STATUS_PASSED,
+  TEST_STATUS_FAILED
+};
+
+/* Enum with values */
+enum 
+{
+  PASSED = 1,
+  FAILED = 2
+};
```
benharsh: Let's also add a test case like this:
```
enum direction {
  NORTH,
  SOUTH,
  EAST,
  WEST
};
```

I think it would also be good if the comment before this group of constants mentioned the name (e.g. ``direction``).

tools/c2chapel/c2chapel.py
```diff
@@ -343,6 +343,12 @@ def genVar(decl):
     print("extern var " + name + " : " + ty + ";")
     print()
 
+
+def genEnum(decl):
+    if type(decl) == c_ast.Enum:
+        for val in decl.values.enumerators:
```
benharsh: Let's add a generated comment before the constants. If the enum has a name, we can print it there, otherwise we can just call it "anonymous"

tools/c2chapel/test/enum.h
Spartee: @benharsh the if (type) was getting passed as the typedef'd enums are a TypeDecl class not a Decl class. Added method genTypeEnum() to fix.

PR #6498, opened 2017-06-19
Struct forward declarations in c2chapel

Spartee:
Fixed struct forward declaration corner case in c2chapel. Nothing needs to be emitted by c2chapel for this type of forward declaration so this fix makes it so that the program does not show an error/stop.


benharsh:
Can you add a test for this case somewhere? Maybe in miscTypedef.h ?

benharsh:
Let's also add a test for this case:
```c
struct forwardStruct;
typedef struct forwardStruct fwdStruct;

struct forwardStruct {
  int a;
  int b;
};
```

PR #6419, opened 2017-06-08
Implement bounded-coforall optimization for remote coforalls 

ronawho:
Optimize remote task creation speed. This applies the same optimization done
for local coforalls in #5018, but to remote coforalls

We used to transform code like:

```chapel
coforall loc in Locales do on loc { }
```

into roughly something like:

```chapel
for loc in Locales {
  _upEndCount(_coforallCount);
  executeOnNB(loc, body());
}
_waitEndCount(_coforallCount);
```

and now we transform it into something like:

```chapel
var tmpIter = Locales;
var bounded = isBoundRange(tmpIter) || isDomain(tmpIter) || isArray(tmpIter);
if bounded {
  var numTasks = tmpIter.size;
  // only bump EndCount once, instead of once per task
  _upEndCount(_coforallCount, numTasks);
  for loc in tmpIter {
    executeOnNB(loc, body());
  }
  _waitEndCount(_coforallCount, numTasks);
} else  {
  // old style coforall+on
}
```

Under ugni this turns 2*numTasks network atomic updates into a single atomic
update. We now only update our `EndCount.i` once, and we don't have to touch
`EndCount.taskCnt` at all.

This pretty dramatically improves remote-task creation, especially under ugni.
For the remote task microbenchmark we see a 2.5X speedup:

|               | before | now   |
| ------------- | ------ | ----- |
| u+q AMO       | 36.5s  | 13.5s |
| u+q processor | 15.5s  | 14.5s |
| gn-aries      | 33.5s  | 32.5s |
| gn-mpi        | 43.5s  | 42.5s |

This results in some decent performance improvements for real benchmarks as
well: 25% speedup for lulesh, up to 50% perf boost for stencil PRK, and minor
improvements for some other benchmarks as well:

| Benchmark   | before       | now          |
| ----------- | ------------ | ------------ |
| lulesh      | 3.35s        | 2.5s         |
| stencil prk | 120 GFlops/s | 130 GFlops/s |
| stencil-opt | 208 GFlops/s | 320 GFlop/s  |
| prom stream | 450 GB/s     | 640 GB/s     |

This resolves #6357

PR #6499, opened 2017-06-19
perf graph annotations

psahabu:
- bounded coforall optimization (#6419)
- optimize updateFluff (#6418)
- gcc upgrade to 7.1.0 and reversion to 6.3.0

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -674,7 +674,7 @@ meteor:
   08/10/16:
     - Updated meteor's lock to yield less frequently (#4300)
 
-mg-b:
+mg.ml-time:
   05/11/17:
     - Move chpl_getPrivatizedClass() into chpl-privatization.h (#6212)
   05/23/17:
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -244,6 +244,12 @@ all:
   06/08/17:
     - text: upgrade gnu target compiler from 6.3.0 to 7.1.0
       config: Single node XC, 16 node XC
+  06/14/17:
+    - text: upgrade gnu target compiler from 6.3.0 to 7.1.0
+      config: chapcs, chap03, chap04
+  06/16/17:
+    - text: revert gnu target compiler from 7.1.0 to 6.3.0
+      config: chapcs
 
 
 AllCompTime:
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -871,7 +871,7 @@ stream-promoted.ml-perf:
     - text: Improve const inference for chpl__iterLF (#5736)
       config: 16 node XC
   06/10/17:
-    - bounded-coforall optimization for remote coforalls (#6419)
+    - text: bounded-coforall optimization for remote coforalls (#6419)
       config: 16 node XC
 
 STREAM_study:
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -128,8 +128,6 @@ all:
   05/05/16:
     - text: ubuntu update to 16.04 (gcc upgraded to 5.3)
       config: shootout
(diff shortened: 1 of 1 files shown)


test/ANNOTATIONS.yaml
```diff
@@ -128,6 +128,8 @@ all:
   05/05/16:
     - text: ubuntu update to 16.04 (gcc upgraded to 5.3)
       config: shootout
+  06/10/17:
+    - bounded-coforall optimization for remote coforalls (#6419)
```
ronawho: I would mark this 16 node only

test/ANNOTATIONS.yaml
```diff
@@ -862,6 +876,9 @@ stream-promoted.ml-perf:
   03/21/17:
     - text: Improve const inference for chpl__iterLF (#5736)
       config: 16 node XC
+  06/10/17:
+    - text: bounded-coforall optimization for remote coforalls (#6419)
```
ronawho: Don't need to re-list this if it's already in the "all" list, right?

test/ANNOTATIONS.yaml
```diff
@@ -242,6 +244,12 @@ all:
   06/08/17:
     - text: upgrade gnu target compiler from 6.3.0 to 7.1.0
       config: Single node XC, 16 node XC
+  06/14/17:
+    - text: upgrade gnu target compiler from 6.3.0 to 7.1.0
+      config: chapcs, chap03, chap04
```
ronawho: chap03 testing isn't a thing anymore

test/ANNOTATIONS.yaml
```diff
@@ -242,6 +244,12 @@ all:
...
+  06/16/17:
+    - text: revert gnu target compiler from 7.1.0 to 6.3.0
+      config: chapcs
```
ronawho: Shouldn't the config mirror the above "upgrade" config?

test/ANNOTATIONS.yaml
```diff
@@ -670,6 +678,8 @@ mg-b:
     - Move chpl_getPrivatizedClass() into chpl-privatization.h (#6212)
   05/23/17:
     - Fix NPB MG (#6284)
+  06/13/17:
+    - Optimize updateFluff (#6418)
```
ronawho: This impacted the prk stencil as well. @benharsh any others that you would expect to be impacted?

benharsh: I don't think so, should just be MG and the Stencil PRK.

test/ANNOTATIONS.yaml
psahabu: I thought @dmk42 said it was only reverted on chapcs, but I may have misheard him.

ronawho: setup_gcc_nightly is what gets changed, so all of the chap* and chapcs* machines will get changed at the same time

psahabu:
@ronawho I think I whacked all the goofiness you pointed out

test/ANNOTATIONS.yaml
```diff
@@ -734,6 +745,8 @@ prk-stencil:
     - blockDist for useStencilDist output matrix and more (#3705)
   05/06/16:
     - Problem size increased from --order=1000 to --order=32000 (#3813)
+  06/13/17:
+    - Optimize updateFluff (#6418)
```
ronawho: This is the single-locale graph, and was not impacted by this change

test/ANNOTATIONS.yaml
ronawho: @psahabu not on you, but think this graph is misnamed. `mg-b` is the single-node version, and all of these annotations are 16-node so `mg-b` should probably be renamed to `mg.ml-time`

psahabu: I think I fixed this correctly?

ronawho:
I think this is ok. I didn't verify the dates, but seems reasonable

PR #6502, opened 2017-06-20
streamlined astr(s1)

vasslitvinov:
Most of the time (~90%) astr() with 1 argument is invoked
when the argument is already canonicalized.

In those cases, avoid executing malloc() + free().

testing: linux64 --verify, gasnet --verify


daviditen:
This looks good to me.

PR #6504, opened 2017-06-20
add ./tmp to toplevel gitignore list

awallace-cray:
A handy ./tmp subdirectory for dev and test miscellany, 
which Git will ignore (but git clean -fdx will still wipe).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ tags
 /build
 /lib
 /tar
+/tmp
 
 # Compiler ignores.
 /compiler/BROWSE


benharsh:
Can we give this a more descriptive name? I could imagine a user or developer having their own $CHPL_HOME/tmp directory, and I wouldn't want it to be removed by some script. For example, I believe start_test removes $CHPL_TEST_TMP_DIR when finished.

ronawho:
Agreed with @benharsh. I often use $CHPL_HOME/tmp or $CHPL_HOME/.tmp as a staging area. And for WIP branches, I'll occasionally commit temporary tests there as I'm developing/testing

ronawho:
And from your mail:

>If you hate the idea, please say so and it's gone.

I certainly don't "hate" this idea, but I'm in favor of just punting on this. It's unlikely that most devs will use the same name for their "scratch" space, and if they want it to be git ignored, they could always just add it to their global .gitignore instead

PR #6507, opened 2017-06-20
QualifiedType for copyPropagation and scalarReplace

benharsh:
A minor change to copyPropagation was made to no longer
consider (PRIM_ASSIGN ref ref) to create an alias. This pattern
simply copies the contents of the references, like ``*a = *b``

Testing:
- [x] local
- [x] no-local
- [x] gasnet


benharsh:
I spot checked a dozen 'interesting' performance tests and didn't see any regressions.

PR #6508, opened 2017-06-20
Switch knucleotide's associative domain to use default parSafe

bradcray:
As a result of PR #6244, this isn't necessary to get good performance any longer.


bradcray:
@chapel-lang/perf-team : I think this is a no-brainer, but am tagging you all just to be safe.

PR #6509, opened 2017-06-21
Free strings leaked by Regex.subn() / qio_regexp_replace()

bradcray:
This PR closes our current top memory leak which relates to a c_string returned by our C regular expression library that Chapel never frees.  I took the approach of stealing the c_string into a Chapel string to avoid a copy rather than simply freeing the c_string (which also closes the leak).

Before:
```
16         1414348    io buffer or bytes
```

After:
```
4          544        io buffer or bytes
```

I also diagnosed the source of these remaining leaks (also in regexp), but have not figured out how to close them yet.  I marked the case with a TODO in the source for the time being.

While here, I added __LINE__ information to QIO memory calls to help determine which allocations / reallocations are being leaked.  Supplying a file number is more challenging because we don't have a good way to map runtime files to integers at present, but the line number alone is typically enough to hone in on the offending line by doing, say 'git grep -n qio_malloc' and looking for the line that was reported in the verbose memory trace.

I tested this by:
* spot-checking memory leaks and performance figures for regexp.chpl, both of which seem to have improved
* running valgrind testing on all regexp tests using an older version of re2 that passes valgrind testing
* running complete linux64 testing

### DIFF
--- a/modules/standard/Regexp.chpl
+++ b/modules/standard/Regexp.chpl
@@ -929,7 +929,6 @@ record regexp {
     } else {
       nreplaced = qio_regexp_replace(_regexp, repl.localize().c_str(), repl.length, text.localize().c_str(), text.length, pos, endpos, global, replaced, replaced_len);
     }
-    writeln("replaced.type = ", replaced.type:string);
     const ret = replaced:string;
     chpl_free_c_string_copy(replaced);
     return (ret, nreplaced);
--- a/modules/standard/Regexp.chpl
+++ b/modules/standard/Regexp.chpl
@@ -413,6 +413,7 @@ proc compile(pattern: string, utf8=true, posix=false, literal=false, nocapture=f
   opts.dotnl = dotnl;
   opts.nongreedy = nongreedy;
 
+  // TODO: At present, 'ret' is leaked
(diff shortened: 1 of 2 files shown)


bradcray:
@mppf : This one is solidly in your home court, if you have time to review it.

mppf:
@bradcray - looks good, thanks for looking at this issue & fixing it.

Re. the remaining memory leak, the regexps are supposed to go into a per-pthread cache of regexps. Perhaps the remaining leak in this area is bounded by the cache size? Which is 8 according to the #define REGEXP_CACHE_SIZE. If that's the case, I'd look more at finding a way to clean up the cache than to figure out how to free that particular re_t. That is, assuming I'm reading the situation & your comment correctly. (I didn't dig in to it).

bradcray:
> If that's the case, I'd look more at finding a way to clean up the cache than to figure out how to free that particular re_t. That is, assuming I'm reading the situation & your comment correctly. (I didn't dig in to it).

OK, thanks for that pointer.  Since these leak chases are spare-time activities, I'm going to punt on this additional re2 leak for now and focus on the next biggest offender when I have the time to do more.

PR #6512, opened 2017-06-21
Only perform packed updateFluff when there are multiple locales

benharsh:


benharsh:
@chapel-lang/perf-team : this should resolve the NPB MG single-node regression from last week.

bradcray:
This looks good and like a no-brainer to me.  I'm wondering, though, how hard it would be to make it more sophisticated.  For example, if I were running StencilDist on a 2D array that was only decomposed in one dimension, my instinct is that we'd see similar slowdowns to those we saw for single-locale testing (?).  Is there any way to filter on which dimensions of fluff we're communicating on and whether or not those dimensions are flat in the target locales array?  (this can be left as future work... it just seems related which is why I ask here).

PR #6513, opened 2017-06-21
Add a compiler flag to error when a constructor is defined in user code

lydia-duncan:
This will allow Mike and I to find test cases that need constructors converted
into initializers more quickly.

Will not error for internal and package/standard module constructors

Add four tests for the --user-constructor-error flag, including:
- Primary constructors (expect error message)
- Secondary constructors (expect error message)
- Initializers only (expect no error message)
- No constructor or initializer (expect no error message)

Update the relevant flag tests

Passed full standard paratest


mppf:
My only concern here is that at some future point we need to remember to remove this and some of the other compiler flags being introduced for the initializers effort. Do we track that in an issue? In some kind of test in test/deprecated? Or something else?

lydia-duncan:
This one should be pretty easy to find with a grepcomp on "constructor", but your point is a good one

PR #6517, opened 2017-06-22
Fix deinitialization order in AutoDestroyScope::variablesDestroy

mppf:
Variables need to be deinitialized in the reverse declaration order
(because they could refer to each other - or memory managed by each
other). Here I add a new test that checks deinitialization order in a way
that is meant to trigger different behavior in
AutoDestroyScope::variablesDestroy. Indeed, this new test reveals a
problem with the deinitialization order of variables declared in simple
nested blocks.

Then I fix AutoDestroyScope::variablesDestroy to deinitialize such
examples in the correct order. I did that by adding a PRIM_NOOP at the
end of blocks without a terminating GOTO; then in any case the
autoDestroy calls are added in reverse declaration order. This PRIM_NOOP
is added simply as a placeholder that the following code can insert
before and it is removed after it is no longer useful for this purpose.

Passed full local testing.

### DIFF
--- a/compiler/resolution/AutoDestroyScope.cpp
+++ b/compiler/resolution/AutoDestroyScope.cpp
@@ -115,14 +115,20 @@ void AutoDestroyScope::variablesDestroy(Expr*      refStmt,
                                         VarSymbol* excludeVar) const {
   // Handle the primary locals
   if (mLocalsHandled == false) {
-    bool   insertAfter = false;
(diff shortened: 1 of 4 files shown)


mppf:
@noakesmichael - would you mind having a look at this bug fix? I believe you are the natural reviewer for the change to AutoDestroyScope::variablesDestroy. Thanks!

PR #6531, opened 2017-06-23
Build llvm tools

coodie:
This branch is supposed to add llvm tools to chapel building. By tools I mean commands like `llvm-dis`, `opt`. Couple of other useful tools (which are definitely going to be useful for testing) like `FileCheck` should be built and installed as well. 

I can't see easy and standard way of adding only specified tools. Makefile generated by llvm's CMake contains targets like `install-FileCheck`, and we could use them to build only the necessary tools, but I feel this is a little bit hacky solution and makes build system harder to maintain.

I think one good idea might be adding `llvm` directory to `utils` and move some of tools there, so that the tools are easier to use in tests for example.

I'd be glad if someone offered a little bit help with this because testing these changes (making sure that proper tools get built and installed) takes a lot of time on my computer as I have to rebuild chapel to make sure everything works (still not sure about this).


mppf:
I've verified that the configure-based LLVM build builds & installs these tools already.

mppf:
I've verified that a cmake-based build builds & installs FileCheck and llvm-dis

mppf:
Passed full local testing.

PR #6533, opened 2017-06-23
LLVM vectorization improvements

coodie:
This branch introduces improvements to LLVM vectorization. 

Improvements relate to adding "llvm.mem.parallel_loop_access" metadata to instructions inside parallel loops. Specifically we add the metadata to loads and stores in loops marked as order independent (vectorizeOnly for example). There are two ways of adding metadata to loops:
1) Metadata refers to the innermost loop instruction is in.
2) Metadata refers to the stack of nested loops that instruction is in.

Implementation covers only 1).

### DIFF
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -40,6 +40,10 @@ static llvm::MDNode* generateLoopMetadata(bool vectorize)
   auto tmpNode        = llvm::MDNode::getTemporary(ctx, llvm::None);
   args.push_back(tmpNode.get());
 
+  // llvm.loop.vectorize.enable metadata is only used to either:
+  // 1) Explicitly disable vectorization of particular loop
+  // 2) Print warning when vectorization is enabled (using metadata) and vectorization didn't occur
+  // This means enabling vectorization using this metadata is not necessary for llvm vectorizer to vectorize loop
   if(vectorize)
   {
     llvm::Metadata *loopVectorizeEnable[] = { llvm::MDString::get(ctx, "llvm.loop.vectorize.enable"),
@@ -172,7 +176,7 @@ GenRet CForLoop::codegen()
 
     llvm::MDNode* loopMetadata = nullptr;
     if(fNoVectorize == false && isOrderIndependent()) {
-      loopMetadata = generateLoopMetadata(true);
+      loopMetadata = generateLoopMetadata(false);
       info->loopStack.emplace(loopMetadata, true);
     }
 
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -24,8 +24,40 @@
 #include "build.h"
 #include "codegen.h"
 #include "ForLoop.h"
+#include "driver.h"
 
 #include <algorithm>
+
+#ifdef HAVE_LLVM
+
+
+static llvm::MDNode* generateLoopMetadata(bool vectorize)
+{
+  GenInfo* info = gGenInfo;
+  auto &ctx = info->module->getContext();
+
+  std::vector<llvm::Metadata*> args;
+  auto tmpNode        = llvm::MDNode::getTemporary(ctx, llvm::None);
+  args.push_back(tmpNode.get());
+
+  if(vectorize)
+  {
+    llvm::Metadata *loopVectorizeEnable[] = { llvm::MDString::get(ctx, "llvm.loop.vectorize.enable"),
+                                              llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(llvm::Type::getInt1Ty(ctx), true))};
+    args.push_back(llvm::MDNode::get(ctx, loopVectorizeEnable));
+  }
+
+  llvm::MDNode *loopMetadata = llvm::MDNode::get(ctx, args);
+  loopMetadata->replaceOperandWith(0, loopMetadata);
+  return loopMetadata;
+}
+
+static void addLoopMetadata(llvm::Instruction* instruction, llvm::MDNode* loopMetadata)
+{
+  instruction->setMetadata("llvm.loop", loopMetadata);
+}
+
+#endif
 
 /************************************ | *************************************
 *                                                                           *
@@ -138,7 +170,16 @@ GenRet CForLoop::codegen()
     info->builder->SetInsertPoint(blockStmtBody);
     info->lvt->addLayer();
 
+    llvm::MDNode* loopMetadata = nullptr;
+    if(fNoVectorize == false && isOrderIndependent()) {
+      loopMetadata = generateLoopMetadata(true);
+      info->loopStack.emplace(loopMetadata, true);
+    }
+
     body.codegen("");
+
+    if(loopMetadata)
+      info->loopStack.pop();
 
     info->lvt->removeLayer();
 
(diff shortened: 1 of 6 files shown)


test/llvm/vectorization/SKIPIF
```diff
@@ -0,0 +1 @@
+CHPL_LLVM!=llvm
```
mppf: I think this file should end in a newline

mppf: Or something... never sure what the red characters in GitHub here at the end mean.

test/llvm/vectorization/parallel_loop.chpl
```diff
@@ -0,0 +1,15 @@
+
+proc loop (A, B) {
+  // This loop should get vectorized when passed --vectorize
+  // CHECK: <4 x i32>
+  for i in vectorizeOnly(0..511) {
+      A[A[i]] = B[i];
+      A[i] = B[i+1];
+  }
```
mppf: Does this case vectorize without addLoopVectorizationHint? With or without the run-time check?

mppf:
I'd like to see this in two PRs:

1) PR adding the testing, but as .futures if they don't vectorize right without the hints
2) PR adding the hints and moving those .futures to regular tests

test/llvm/vectorization/parallel_loop.chpl
mppf: Anyway I'd consider passing -vectorize-scev-check-threshold=0  -runtime-memory-check-threshold=0 LLVM options to disable the runtime check.

mppf: (for testing purposes)

coodie: Adding loop vectorization (`llvm.loop.vectorize.enable`) is not necessary to get anything vectorized, it only makes sense when we want to **disable** vectorization of particular loop. What is important is adding `llvm.mem.parallel_loop_access`, because it greatly improves vectorization (we can remove runtime checks entirely and stuff gets vectorized anyway). I couldn't figure out working case by myself so I took this from llvm's test suite, `llvm.mem.parallel_loop_access` is necessary to get this vectorized, because it's very tricky and not obvious that there is no loop carried dependency.

mppf: ah, cool

coodie:
@mppf 
I've split this PR into testing and implementation, and synchronized both of branches with master.

compiler/util/clangUtil.cpp
```diff
@@ -54,6 +54,8 @@
 
 #include "llvmDebug.h"
 
+#include "llvm/Support/TargetRegistry.h"
```
mppf: I must be missing something - why is this #include now necessary? You didn't change anything else in this file.

coodie: Oh well, apparently I forgot to remove this include, it's not necessary for anything to work.

mppf:
- [ ] test this with --vectorize
- [ ] try turning on --vectorize by default for --llvm

compiler/codegen/CForLoop.cpp
```diff
@@ -24,9 +24,35 @@
...
 #include "codegen.h"
 #include "ForLoop.h"
+#include "driver.h"
 
 #include <algorithm>
 
+#ifdef HAVE_LLVM
+namespace
+{
+  llvm::MDNode* addLoopVectorizationHint(llvm::Instruction* instruction)
```
mppf: I'd expect this function to be `static` rather than in an anonymous namespace. Arguably a style issue, but making it `static` will fit in with the rest of the compiler. Vs when defining a type, you can't use `static`.

compiler/codegen/CForLoop.cpp
```diff
@@ -154,7 +180,19 @@ GenRet CForLoop::codegen()
                                                FNAME("condition"));
 
     // Create the conditional branch
-    info->builder->CreateCondBr(condValue1, blockStmtBody, blockStmtEnd);
+    llvm::Instruction* endLoopBranch = info->builder->CreateCondBr(condValue1, blockStmtBody, blockStmtEnd);
+
+    if(fNoVectorize == false && isOrderIndependent())
+    {
+        llvm::MDNode* llvmLoopMetadata = addLoopVectorizationHint(endLoopBranch);
+        for(auto it = blockStmtBody->begin(); it != blockStmtBody->end(); it++)
```
mppf: Prefer `++it` to `it++` for this kind of loop

mppf: Note with LLVM 5.0 (and not before) the recommendation changes to
```
for (Instruction &I : BB)
```

mppf: I think this code might only add parallel loop access metadata to simple loops.
What if we have
``` chapel
var sum = 0;
for i in vectorizeOnly(1..100) with (+ reduce sum) {
  if i > 10 then
     sum += i;
  else
     sum += 1;
}
```

? Won't this loop over blockStmtBody instructions not visit those instructions "inside" the conditional?

compiler/include/codegen.h
```diff
@@ -78,6 +91,8 @@ struct GenInfo {
   llvm::IRBuilder<> *builder;
   LayeredValueTable *lvt;
 
+  std::stack<LoopData> loopStack;
```
mppf: Since this stack only ever contains elements with parallel=true and since you're not extending other loop generation (while loops, other loops, ...), I think this stack should just store llvm::MDNode* for parallel loops and be called something like `parallelLoopStack`.
Do you envision some case in which this stack will need to contain non-parallel loops?

compiler/codegen/expr.cpp
```diff
@@ -389,6 +389,13 @@ llvm::StoreInst* codegenStoreLLVM(llvm::Value* val,
   llvm::MDNode* tbaa = NULL;
   if( USE_TBAA && valType ) tbaa = valType->symbol->llvmTbaaNode;
   if( tbaa ) ret->setMetadata(llvm::LLVMContext::MD_tbaa, tbaa);
+
+  if(!info->loopStack.empty()) {
+    const auto &loopData = info->loopStack.top();
```
mppf: Note in a comment here and in the PR message that you're only adding `parallel_loop_access` hints for the innermost loops.

mppf:
I'd like to see a future PR that adds the parallel loop access to function calls; see https://github.com/chapel-lang/chapel/blob/master/compiler/codegen/expr.cpp#L2222

mppf:
I tested this, with this patch:
```
diff --git a/compiler/codegen/CForLoop.cpp b/compiler/codegen/CForLoop.cpp
index bcf60f8..9f9dc84 100644
--- a/compiler/codegen/CForLoop.cpp
+++ b/compiler/codegen/CForLoop.cpp
@@ -40,7 +40,7 @@ static llvm::MDNode* generateLoopMetadata(bool vectori
ze)
   auto tmpNode        = llvm::MDNode::getTemporary(ctx, llvm::None);
   args.push_back(tmpNode.get());
 
-  if(vectorize)
+  if(vectorize && 0 /* off for testing */)
   {
     llvm::Metadata *loopVectorizeEnable[] = { llvm::MDString::get(ctx, "llvm.lo
op.vectorize.enable"),
                                               llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(llvm::Type::getInt1Ty(ctx), true))};
```

with  --llvm  --vectorize  --fast

and I saw 1 failure:

      compflags/coodie/llvmPrintIr/llvmPrintIrFull.chpl

Please have a look at disabling the vectorize.enable by default (I think it's fine to leave in the code, but make it off by default somehow) and at getting llvmPrintIrFull to work in that configuration:

    start_test -compopts --llvm -compopts --vectorize -compopts --fast compflags/coodie/llvmPrintIr/


One way to do that would be to make llvmPrintIrFull.skipif containing:
```
COMPOPTS <= --fast
```

Once you address these issues I think this is ready to merge. Thanks.

mppf:
Passed full local testing.

Issue #6536, opened 2017-06-23
Dead module with an enum causes internal compiler error


e-kayrakli:
### Summary of Problem
As in the title: dead module with an enum causes internal compiler error

### Steps to Reproduce

**Source Code:**

deadModule.chpl:
```chapel
enum Foo {BAR};
```
mainModule.chpl:
```chapel
proc main() { writeln("Hello"); }
```

**Compile command:**
```
chpl deadModule.chpl mainModule.chpl
```

**Execution command:**
Doesn't compile.

**Associated Future Test(s):**
PR: #6537 

Will add the future `test/users/engin/deadModuleWithEnum/deadModuleWithEnum.chpl`

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.16.0 pre-release (ad166c86f8)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```

- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
```
gcc (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406
```

e-kayrakli:
For background: @LouisJenkinsCS hit this issue with `BigInteger`. Later he figured out that some enum in the module was causing the error.

bradcray:
This is hitting some logic in deadCodeElimination.cpp that seems to be saying "if there was only one thing in this module, I expected it to be an init function!"  

[...]

PR #6537, opened 2017-06-23
Dead module with enum -- future for issue #6536

e-kayrakli:
Adds a future for the issue #6536.

Added future is reported as future with standard linux64 setup.


e-kayrakli:
I don't know if there is an accepted convention for writing `.bad` files for internal errors. Saw something like this after some grepping. I can change the `.bad` file..

bradcray:
This looks good to me, thanks for filing it @e-kayrakli !

Issue #6538, opened 2017-06-23
Missing error message when returning an inner function that uses an outer-local var


e-kayrakli:
### Summary of Problem
Compiler should create an error message when the user's trying to return an inner function that uses local variables from it's parent function(s).

Current behavior is an internal error, that seems to be coming from a slightly irrelevant part of the compiler. Most likely it is due to a missing actual for an additional formal compiler creates in the inner function's signature to capture the outer variable.

### Steps to Reproduce

**Source Code:**

```chapel
proc foo() {
  var dummy = 10;
  return lambda (x:int) { return x*dummy; };
}
writeln(foo()(5));
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
Doesn't compile. (It shouldn't)

**Associated Future Test(s):**
PR: #6539 

Adds `test/users/engin/retInnerFunction.chpl`

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.16.0 pre-release (ad166c86f8)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
```
gcc (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406
```

mppf:
While I agree this is something we should have a better error message for, whether or not you can do this at all is an open design question for Chapel first-class functions. If the inner function is returned, presumably you'd want the `dummy` variable in the example to continue to exist and be accessible? I'm not sure we want to make the compiler do that. (Remember we have no garbage collection...)

e-kayrakli:
[...]

PR #6539, opened 2017-06-23
Add new error message future for issue #6538

e-kayrakli:
Adds a future for issue #6538 

Added test is reported as future with standard linux64. 

TODO before merge (I think):
- [x] Create an appropriate `.good` file (I thought these were optional)

### DIFF
--- a/test/users/engin/retInnerFunction.future
+++ b/test/users/engin/retInnerFunction.future
@@ -1,5 +1,8 @@
-error message: returning an inner function that uses outer-local vars
+bug: returning an inner function that uses outer-local vars
 
 When returning an inner function that uses an outer-local variable, the
 compiler crashes. Such code is user error and compiler should detect it
 before crashing (by luck?)
+
+NOTE: The .good file should be updated with the appropriate error
+message.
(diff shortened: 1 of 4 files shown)


bradcray:
> Create an appropriate .good file (I thought these were optional)

.good files are required so that we know when a future is passing.  But they can be hard to write for cases like this where you're trying to anticipate what a good error message would be.  I usually make sure to at least get the format right (e.g., filename:lineno: error: ...) and see if I can think of a related error message that I think is good to model the proposed one from.  But I also usually note in the future that the implementer can update the .good if they come up with a better error message, unless I'm particularly proud of the one I created.

test/users/engin/retInnerFunction.future
```diff
@@ -0,0 +1,5 @@
+error message: returning an inner function that uses outer-local vars
```
bradcray: Not a huge deal, but I'd call this a bug (I think any internal error is a bug).  I reserve "error message" futures to be cases where we generate a user-facing error message but it's not as good or helpful as it could be and we'd like to improve it (and don't consider internal errors to be in that same class of user-facing error messages I suppose).

e-kayrakli:
The last revision:
- Changes the future to be a bug rather than an error message
- Updates the `.good` file with appropriate format and adds a note in the `.future` file

bradcray:
I think this is ready to merge and can do so -- is that right, @e-kayrakli ?

e-kayrakli:
@bradcray -- yep!

PR #6541, opened 2017-06-24
Have DimensionalDist2D clean up after itself

bradcray:
Implement the dsiDestroy*() routines for DimensionalDist2D in the obvious way
to have them clean up the per-locale local descriptors.

I'm not sure whether or not there's more that I'm missing, but this
makes a significant impact on our number 1 leaking test, schurComplement, and
passes correctness testing for all tests that use DimensionalDist2D.

### DIFF
--- a/modules/dists/DimensionalDist2D.chpl
+++ b/modules/dists/DimensionalDist2D.chpl
@@ -907,7 +907,7 @@ proc DimensionalArr.isAlias
   return this.dom != this.allocDom;
 
 
-//== creation
+//== creation and destruction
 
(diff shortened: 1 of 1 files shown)


bradcray:
This is simple enough that I'm going to merge it today in order to see new graph results on Monday, but tagging @vasslitvinov in case he's online this weekend or can review it post-merge.

vasslitvinov:
This looks good. The only other thing that comes to mind is to define LocDimensionalDom.deinit() to delete this.doml1, doml2 if those are classes.

PR #6545, opened 2017-06-26
Improving --log support

mppf:
### Summary of command line changes (please read if you use --log=):
 * Removes `--log=` in favor of `--log` and `--log-pass=`.
     * `--log=<letters>` is no longer available; use multiple `--log-pass=` arguments if you want to select several passes.
 * Logging now removes the log directory before outputting any AST dumps.
 * `--log` activates logging of all interesting passes to a default `log` directory.
 * `--log-pass=<name>` and `--log-pass=<letter>` activate logging but only of selected passes. Pass several `--log-pass=` arguments to select multiple passes.
 * Only a few passes have 1-character names, these are currently all the 1st letter of the pass name in lowercase now. Only: parse, scopeResolve, normalize, resolve, codegen.
 * Adds `--log-node` is now available, to select AstDumpToNode instead of AstDump.
 * `--log` need not be passed as an argument if any of `--log-module=`, `--log-pass=`, or `--log-dir=` are included as arguments. These all imply `--log`.

### Details of internal-to-compiler changes:

Besides these changes, updated logging code to work with string pass names instead of 1-char pass IDs. This allows the 1-char pass ID to be omitted and for the pass to be still logged when logging all passes.

To facilitate that, log.h now defines LOG_NEVER, which means a pass should not be logged with --log (but it can stil be selected directly with --log-pass=<name>) and LOG_NO_SHORT, which means that that pass does not have a 1-character short name.

While there I removed all of the 1-character short names that do not seem mnemonic to me, since a short name is no longer necessary to get AST dumps.

Before these changes, lowerErrorHandling would not create logging and such logging could not even be requested, because it did not specify a 1-character short name. Since there are 44 passes, maintaining a 1-character short name for each seems to be a maintenance burden to me.

### Testing and review

Passed full local testing.


mppf:
@lydia-duncan - here is the --log changes PR I have created. Thanks!

compiler/main/log.cpp
```diff
@@ -45,37 +49,74 @@ bool             fdump_html_print_block_IDs             =   false;
...
+  if (availableInitialized == false) {
+    initPassesForLogging(); // calls logMakePassAvailable for each pass
+    availableInitialized = true;
+  }
+
+  // Make --log optional by setting it if --log-pass= is set
+  fLog = true;
+
+  // Since we selected a pass, don't log everything
+  logAll = false;
```
lydia-duncan: So this means that `--log --log-pass 'R'` and `--log-pass 'R' --log` will have different behavior (which it did in the old world, iirc), right?  Just making sure I understand, not a problem

compiler/main/log.cpp
```diff
@@ -45,37 +49,74 @@ bool             fdump_html_print_block_IDs             =   false;
...
+  // Check first for a match in the available pass names
+  for( size_t i = 0; i < logAvailableNames.size(); i++ ) {
+    if (logAvailableNames[i] == arg) {
+      logOnlyName.insert(logAvailableNames[i]);
+      return;
+    }
+  }
 
-      arg++;
+  // If none found, check for a match in the short names.
```
lydia-duncan: This confused me for a minute, so it would probably be worth adding a comment about this being in support of the multi-pass form (with an example of what that means)

compiler/main/log.cpp
```diff
@@ -45,37 +49,74 @@ bool             fdump_html_print_block_IDs             =   false;
...
+  if (strlen(arg) == 1 &&
+      arg[0] != LOG_NO_SHORT &&
+      arg[0] != LOG_NEVER) {
+    for( size_t i = 0; i < logAvailableShortNames.size(); i++ ) {
+      if (logAvailableShortNames[i] == arg[0]) {
+        logOnlyName.insert(logAvailableNames[i]);
```
lydia-duncan: Except it returns immediately and now I am confused again.  This is how we handle the multi-pass form, right?  Do we send in a truncated version?

compiler/main/log.cpp
```diff
@@ -45,37 +49,74 @@ bool             fdump_html_print_block_IDs             =   false;
...
+    fLog = true;
+  // Enable logging if --log-pass is used
+  if (logOnlyName.size() > 0)
+    fLog = true;
+  // Enable logging if --log-dir is used
+  if (log_dir[0] != '\0')
+    fLog = true;
+
+  if (fLog || fdump_html || *deletedIdFilename) {
+    // Remove the log directory to make sure there is no stale data
```
lydia-duncan: Might be good to give a warning this is happening so the user isn't confused

compiler/main/runpasses.cpp
```diff
@@ -40,49 +40,49 @@ struct PassInfo {
...
+#define LOG_copyPropagation                    LOG_NO_SHORT
+#define LOG_deadCodeElimination                LOG_NO_SHORT
+#define LOG_removeWrapRecords                  LOG_NO_SHORT
+#define LOG_removeEmptyRecords                 LOG_NO_SHORT
+#define LOG_localizeGlobals                    LOG_NO_SHORT
+#define LOG_loopInvariantCodeMotion            LOG_NO_SHORT
+#define LOG_prune2                             LOG_NO_SHORT
+#define LOG_returnStarTuplesByRefArgs          LOG_NO_SHORT
 #define LOG_insertWideReferences               'W'
 #define LOG_optimizeOnClauses                  'o'
```
lydia-duncan: I think it is confusing to have a mix of capitalized and non-capitalized single-character log letters.  I suspect you have removed all the ones where there is a pair of upper/lower case - would you mind unifying to lowercase?

lydia-duncan:
I think this is reasonable and strongly suspect that we don't have any tests of the --log flag (though that isn't to say that you *shouldn't* do a full paratest).  What cases have you checked by hand to ensure correctness?

compiler/main/log.cpp
mppf: I'm pretty sure they'll have the same behavior. logAll is `true` by default and giving --log-pass will cause it to be set to `false`. In that event only the selected passes are logged.

compiler/main/log.cpp
mppf: It no longer supports e.g. --log-pass=RS. Instead you can pass --log-pass=R --log-pass=S. In that event, this routine is called twice, once with "R" and once with "S". The routine first checks for pass names matching. Since none will be 1-character long, it does this second check.

compiler/main/log.cpp
mppf: I don't think we should give a warning. It would be new & different for this sort of thing. We don't generally output such purely informational notes.

compiler/main/runpasses.cpp
mppf: Sure, I can do that. I think I might remove a few more 1-character shortcuts too (in particular optimizeOnClauses), on the theory that if anybody really cares about a particular one we'll have it back again soon.

mppf: I'm only leaving 1-character names for "the big ones":
 * parse
 * scope resolve
 * normalize
 * resolve
 * codegen

It's likely somebody else would create a different set but this particular set allows each 1-letter abbreviation to be just the 1st letter.

mppf:
I checked --log, --log --log-pass=resolve, --log --log-pass=R, --log-pass=R, --log-module=hello, --log --log-node, --log --log-dir=log2, --log-dir=log2

compiler/main/log.cpp
mppf: I *can* make the old mashup of 1-character names work again (e.g. --log-pass=sr) but I don't personally see the value in supporting that. Either way it wouldn't be hard to adjust this code to do it. So just let me know if you have a preference there.

lydia-duncan: Oh, right, sorry.  I was misreading the iteration space and thought for some reason it was over arg and not the list of passes.  Never mind!

compiler/main/log.cpp
lydia-duncan: We don't blow away the ccode directory, though, and this behavior is a change from previous behavior.  I just think it would be surprising if the user didn't look too closely at the release notes/commit messages, and think it would be helpful for the transition.

PR #6546, opened 2017-06-27
Implement reindex() of an empty domain/array

bradcray:
In some recent work, I was running into an issue in which the .reindex()
operation was not working properly when applied to an empty domain
(or a domain with an empty range).  I've added a small test of this case
here as emptyReindex.chpl.  Then I updated the reindex operation to
make such cases work and changed the approach by having it handle
a dimension at a time.  As part of this effort, I refactored some repeated
indexOrder logic into a helper routine to avoid duplicating the work.


bradcray:
@benharsh: you would be a natural reviewer for this if you have time.

test/arrays/reindex/emptyReindex.good
```diff
@@ -0,0 +1 @@
+
```
benharsh: Is B's domain supposed to be ``{10..9}`` ? If so, can you print it out here as well?

bradcray: Good point.  I was just happy to no longer get an execution-time error.  :)

bradcray:
Hi @benharsh: Would you mind giving this a re-review?  It turns out that I had a stupid error in that I was calling the hardcoded-to-1D arrays version of the convert index utility for a single dimension of a multidimensional array.  This essentially broke multidimensional cases.  I've expanded the test to cover a 2D case, and refactored the code to fix the bug and reduce some of the repeated business logic by moving it into a helper function.

PR #6553, opened 2017-06-27
Study version of LLNL's CoMD

benharsh:
This implementation of CoMD in Chapel will serve as a point of comparison for elegance and performance when we write a more idiomatic version in Chapel.

TODO:
- [x] 500 correctness trials
- [x] verify perf/graph files work


benharsh:
@ben-albrecht : would you be available to review?

test/studies/comd/llnl/LICENSE
```diff
@@ -0,0 +1,70 @@
+/******************************************************************************
+
+CoMD-Chapel v1.0
+
+Copyright (c) 2016, Lawrence Livermore National Security, LLC.
```
ben-albrecht: Does this need to be updated?

benharsh: Perhaps, but I think that's on LLNL. I just copied this license from their repo.

benharsh: @tmacd9  : does that sound right to you?

PR #6554, opened 2017-06-27
Fix for multilocale timing in PRK-DGEMM

e-kayrakli:
PRK-DGEMM was coded to run in a multilocale setting but wasn't tested. I discovered a bug in starting the timer for multilocale setup and fixed it in this PR.

P.S. This PR does *not* add multilocale testing.

The test in question passes with standard linux64 config.

### DIFF
--- a/test/studies/prk/DGEMM/dgemm.chpl
+++ b/test/studies/prk/DGEMM/dgemm.chpl
@@ -87,7 +87,7 @@ else {
             CC: [blockDom] dtype;
 
         for niter in 0..iterations {
-          if tid==0 && niter==1 then t.start();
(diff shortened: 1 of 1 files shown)


e-kayrakli:
pinging @ben-albrecht

e-kayrakli:
(Started a review by mistake, hope it didn't create some noise)

Reading that line again made me think that maybe it makes more sense to check `niter` first. But I don't think it would have any observable performance impact.

ben-albrecht:
> But I don't think it would have any observable performance impact.

It would be good to prove that to ourselves, but I don't think it's necessary for this PR.

Issue #6392, opened 2017-06-06
Conditional local statement


e-kayrakli:
More often than not, I want being able to control `local` statements with conditionals. To be fair, this mostly happens in benchmark code, but at the same time, I think it can have practical use cases, as well. Users may want to do an explicit locality check before entering a performance-critical section of the code, and suppress creating wide pointers if they can determine that everything is local during run time.

Currently, what I end up doing generally is:

```chapel
if condition do
  local do
    someKernel()
else
  someKernel()
```
By itself, this feels clunky. Moreover, distilling `someKernel()` may not be very straightforward/natural.

Instead:
```chapel
local condition do
  someKernel()
```
can be very useful in such cases.

Grammar/syntax can be refined:
```chapel
local if condition do
  someKernel()
```
might be an option. but can be confused with 
```chapel
local do if condition then
  someKernel()
```

Another option may be using `when` instead of `if`:
```chapel
local when condition do
  someKernel()
```

It seems like this should require small modifications in the compiler. If there is enough support, I may take this up as a small parser project.

ronawho:
pinging @benharsh

benharsh:
I generally support this kind of feature. One possible option is to redefine the grammar for ``local`` such that it is similar to ``serial``. ``serial`` currently has the following grammar:
```
TSERIAL expr do_stmt
TSERIAL do_stmt
```

We only have a handful of tests that rely on syntax like ``local foo();``, so it wouldn't be that big of a change. I think we'd prefer the local statement to go away long-term, but this might be a good quality of life improvement in the meantime.

e-kayrakli:
I was wondering if there are any objections to this? Should I go ahead and try to add this to the grammar? (I may be able to work on it as early as this week)

benharsh:
I think @bradcray should have a chance to comment as well.

bradcray:
I don't have any problem adding it if it isn't too much of a distraction and agree that following the pattern set by `serial` is the right way to go (including having the lack of an expression mean `local true`.  If we were wed to the `local` keyword for the long-term, I'd definitely want to take it in this direction.  My only reservation is whether it represents too much wasted work if we manage to retire `local` in the next release cycle or two.  But if it's easy to add and saves some pain and suffering in the meantime, I'd say go for it.

[...]

PR #6555, opened 2017-06-27
Conditional local statement

e-kayrakli:
Modifies the grammar for `local` statement to allow conditional `local` blocks.

With this PR, the `local` statement is identical to the `serial` statement grammar-wise. See issue #6392 for more detailed discussion.

Also fixes some of the tests with old grammar I was able to find with naive grepping. I suspect that there are more that I missed.

TODO:
- [x] full paratest (something I cannot do myself)


e-kayrakli:
@bradcray  and @benharsh -- here is what I have for this.

It seemed to be working with much less effort than I was anticipating, which makes me a bit uncomfortable. I hope I am not missing anything.

benharsh:
Can you add a test case for a program that would fail to run if not for the conditional feature?

e-kayrakli:
@benharsh -- as in a test case that uses the old grammar?

Like `local foo()`(old) instead of `local do foo()`(new)

benharsh:
I meant something that uses the conditional to avoid a local block around remote code. Something along these lines:

```chpl
proc inc(x) {
  local x.locale == here do x += 1;
}
var A = 1;
inc(A);

// should enter non-local branch
on Locales[numLocales-1] do inc(A);
```

Basically a motivating use-case for this feature.

e-kayrakli:
Ah, I see. Will do!

compiler/parser/chapel.ypp
```diff
@@ -480,7 +480,8 @@ stmt:
 | TCONTINUE opt_ident TSEMI            { $$ = buildGotoStmt(GOTO_CONTINUE, $2); }
 | TDELETE expr TSEMI                   { $$ = buildPrimitiveStmt(PRIM_DELETE, $2); }
 | TLABEL ident stmt                    { $$ = buildLabelStmt($2, $3); }
-| TLOCAL stmt                          { $$ = buildLocalStmt($2); }
+| TLOCAL expr do_stmt                  { $$ = buildLocalStmt($2, $3); }
+| TLOCAL do_stmt                       { $$ = buildLocalStmt(new SymExpr(gTrue), $2); }
```
bradcray: Continuing our conversation in the top-level notes, I think I'd expect this to simply be `buildLocalStmt($2)` in order to avoid the early-compiler code bloat that you noted.

e-kayrakli:
I think this is ready for another review/paratest.

P.S. Commit history is a bit messy with some back-and-forths, maybe this should be squash-merged. OTOH there are parser rebuilds, which may not be a good idea to squash with other commits.

e-kayrakli:
@benharsh -- you can send me an email if the list of syntax errors in tests is too long :)

benharsh:
I'll track testing here:

- [ ] full local
- [ ] full no-local
- [ ] release/examples gasnet

benharsh:
Can you also update the documentation in doc/rst/technotes/local.rst ?

benharsh:
I think you missed some local statements in the following files:
- modules/dists/StencilDist.chpl
- test/optimizations/localon/sublocale.chpl
- test/release/examples/benchmarks/ssca2/SSCA2_Modules/SSCA2_RMAT_graph_generator.chpl
- test/studies/hpcc/HPL/vass/schurComplement.chpl
- test/studies/hpcc/HPL/vass/hpl.hpcc2012.chpl

e-kayrakli:
Fixed the doc, modules and tests.

One question, though:

I didn't understand the motivation behind having a separate documentation for "local-on" pattern. There may be some optimizations and whatnot using an on statement tightly nested in a local block. But what is the purpose of documenting it separately? i.e. As a user, I am not sure what to learn different than the previous sections from the "local-on" section. Regardless, I changed "local-on" to "local-do-on" to be more in line with the new grammar.

benharsh:
Thanks for the changes, I'll fire off some testing.

I believe we have documented a 'local-do-on' separately because it will trigger an error if the destination of the on-statement is not on the same node. While this isn't useful for the flat locale model, it is more useful for numa.

e-kayrakli:
> I believe we have documented a 'local-do-on' separately because it will trigger an error if the destination of the on-statement is not on the same node. While this isn't useful for the flat locale model, it is more useful for numa.

Makes more sense now. Thanks

benharsh:
Testing came back clean! Ready for me to merge this?

e-kayrakli:
Good to go from my view.

benharsh:
I think it would make more sense for you to send the email as the main advocate for this change. I'm fine with holding this PR for a couple of days after the email is sent.

e-kayrakli:
@benharsh -- The mail is sent. You can merge this whenever you want.

benharsh:
@bradcray : it's been a few days with no responses on the mailing list, do you feel OK with merging this now?

bradcray:
I do, yes.

PR #6556, opened 2017-06-27
Update tests to use explicit initializers instead of constructors, part 1

lydia-duncan:
Updates some tests to utilize initializers where previously constructors were
used.  This is part of our effort to make initializers the default.

test/arrays/vass/const-checking.fields.chpl:
Placed a super.init() call prior to the array updates, and modified the
spacing so that the test was more legible, resulting in an update of
the expected output.

test/classes/bradc/test-class-init-blc.chpl:
Placed a super.init() call prior to the `for` loop used to update the array's
fields.  This test brought up a consideration: the `for` loop is only invalid
in Phase 1 because we haven't initialized the array it is referencing yet.
That initialization is omitted in Phase 1 and so inserted by the compiler,
but today would be inserted just before the super.init() call (after the
`for` loop).  Brad suggested that we instead insert the omitted initialization
*just after* the last field initialization instead, which would allow the `for`
loop to be valid.

test/classes/bradc/bitOpOnClass.chpl:
Updated the spacing of the test.  Similarly to test-class-init-blc.chpl,
insert a super.init() call prior to the `for` loop used to update the array's
fields.

test/classes/claridge/baseConstructorCall.chpl:
Insert an explicit call to the parent initializer with the appropriate argument,
and insert an explicit super.init() in the parent initializer for the phase 1 divide
(may retract)

test/release/examples/benchmarks/lcals/:
Convert LoopStat and CycleTimer to use initializers.  LoopTimer cannot use
initializers yet, due to a bug with new statements (which Mike is in the process
of fixing, hopefully in the next day or so).  ADomain will be interesting to
transition, as it relies on select statements and has if/else branches that miss
their last else statement.

@vasslitvinov, would you review:
- test/arrays/vass/const-checking.fields.chpl

David reviewed:
- test/release/examples/benchmarks/lcals/*.chpl
- test/classes/claridge/baseConstructorCall.chpl

Brad reviewed:
- test/classes/bradc/bitOpOnClass.chpl
- test/classes/bradc/test-class-init-blc.chpl

Verified that a fresh checkout of all ran cleanly


test/classes/claridge/baseConstructorCall.chpl
```diff
@@ -1,17 +1,18 @@
 class BaseClass {
   var a, b: int;
 
-  proc BaseClass(b: int) {
+  proc init(b: int) {
     this.b = b;
+    super.init();
```
daviditen: This `super.init()` isn't necessary, right?

daviditen:
The changes to LCALS and baseConstructorCall look good to me.

test/classes/claridge/baseConstructorCall.chpl
lydia-duncan: Not necessary, but I think it is preferable, since it means the statement above it executes in Phase 1 instead of having b get an initial value that will just get overwritten

test/classes/bradc/bitOpOnClass.chpl
```diff
@@ -1,29 +1,32 @@
...
+  var dom: domain(1);
+  var arr: [dom] int;
+
+  proc init(length: int) {
+    this.len = length;
     this.dom = {1..length};
-      forall (a,i) in zip(this.arr, 13..) do
-        a = i;
-        }
+    super.init();
```
bradcray: This looks like another case where the refined "where does the compiler insert its default field inits
 rule we discussed (switching to "just after the previous field init" rather than "just before the next field init") would permit all the code here to be on the same side of the super.init().  The more I think about this rule, the more I think we should implement it sooner to see if it leads to any surprises and to be able to make use of it in cases like this.

lydia-duncan: I think I called that out in the commit message, but did move the commit messages into the PR message - I'll do that now, thanks!

test/classes/claridge/baseConstructorCall.chpl
daviditen: Ah, OK.

bradcray:
I'm good with my two changes, but would be interested in seeing a feature request .future go in with a variant of the test-class-init-blc.chpl case which swaps the order of the super.init() and array assignment as a placeholder for the proposed default field initialization insertion rule.

lydia-duncan:
On it, good point!

test/classes/claridge/baseConstructorCall.chpl
bradcray: In rewriting existing tests, I'd be inclined to take the stance of only using super.init() when it's necessary since it's more likely to be what a typical user does and may lend fodder to any debates around switching the default constructor phase going forward.  I'd also keep a list of cases where it seemed regrettable that the default wasn't the other way around (where double-initialization in a case like this is one such case).

lydia-duncan: @bradcray - I've been taking the policy of putting things that are obviously Phase 1 explicitly before a super.init().  There's a couple of more complicated tests (that didn't work for other reasons) but that naturally seemed to have the Phase 1/Phase 2 divide, so I tried inserting one midway through those (for instance, https://github.com/chapel-lang/chapel/blob/master/test/studies/madness/aniruddha/madchap/MRA.chpl#L62).  Should I be taking the policy of "Phase 2 as much as possible" for those as well?

I was going to avoid rearranging to make a clear Phase 1/Phase 2 divide as much as possible, but imo favoring the "easy" approach for Phase 1 only bodies just means less opportunities to iron the errors out of the more complicated portion, which was part of the intention behind switching the constructors over now rather than focusing on default initializers/noinit.

I was just thinking a list of initializers that could trivially be Phase-1-only would be a good idea, and I think such a list would be just as easily populated from "ugh, why did I have to add this super.init()" moments as "oh, I guess that could be Phase 1 only" (and might even be more quickly noticed in the former).

I'm not really arguing that this is a policy anyone that helps me should follow (since that seems annoying), but I think it is good for *me* to do it, at least.  Thoughts?

bradcray: I think if the test requires a super.init() then putting it in the most natural dividing line between "initialization" and "general computation" makes sense (and I realize that the two tests of mine in this PR _could_ be considered of that flavor, though I could also imagine a user naively thinking of the array assignment as being initialization-like even though it doesn't meet our technical definition).

I don't think we should force tests to use phase 1 for the purposes of finding bugs in the implementation if they don't "want" it naturally.  I.e., our main goal should be to convert the tests over so that they keep working in a deprecated initializer world and to be as true to the test as possible.  Personally, if I had a test in the system that looked like this:

```chapel
class C {
  var x: int;
  proc C() {
    x = 42;
  }
}
```
and I were tasked with rewriting it, I'd rewrite it as:

```chapel
proc init() {
  x = 42;
}
```

(but I'm also a person who didn't understand why C++ supported initializer lists for years and avoided using them as a result).  Even understanding them, my preference is not to use super.init when it's unnecessary and hope that the compiler would optimize the duplicate initialization away (the back-end compiler, if not ours).

This might be a reason to focus initially not just on the most complicated tests, but also the ones that don't have an obvious owner.  That way, the owner can make the call as to how to rewrite them (assuming they have the cycles/interest in doing the rewrite).

bradcray: > I don't think we should force tests to use phase 1 for the purposes of finding bugs in the implementation if they don't "want" it naturally.

Put another way, if rewriting the tests exposes bugs, that's great, but the goal shouldn't be to rewrite the tests in a way that maximizes the chances that they expose bugs; it should be to rewrite them to work and be natural.  I'm guessing that even doing so, we'll find plenty of bugs and feature requests to keep us busy—particularly when we turn our attention to the module code.

lydia-duncan:
@bradcray - I added the future in https://github.com/chapel-lang/chapel/pull/6569, and added a discussion on it in https://github.com/chapel-lang/chapel/issues/6568 since I keep waffling (and so would like more approval on it before its implementation)

bradcray:
Thanks!  I agree that it would need to be discussed before being accepted.

vasslitvinov:
Looks good to me!

PR #6564, opened 2017-06-28
Add a boiled down version of an error I am encountering with autoCopy & init()

lydia-duncan:
When attempting to convert test/types/records/noakes/return.chpl to using an
initializer, I encountered an error when we returned a record that defined an
initializer to an instance that already knew its type.  That error is due to
our implementation of copy initializers, I believe.  We don't validate our
autoCopy function for that type (because they might not work and we only want
that to be an issue if an autoCopy is needed), but it seems that we don't
know how to respond if the autoCopy map doesn't have an entry for that type
because we had skipped it.

Add a test to cover that use case in a simpler form.

Verified a fresh checkout of that test


lydia-duncan:
@mppf - I ran into a wall when trying to resolve this issue, do you have any advice as to how I should go about fixing it?

mppf:
I thought we had decided that a record didn't get compiler-provided default initializers if it provided any initializers? Since the `copy initializer` is an initializer, and your example includes a `proc init`, it shouldn't get a compiler-generated copy initializer.  I think the test you are showing correctly fails to compile. It should presumably just be a user-facing error.

lydia-duncan:
Adding a copy initializer did not resolve the issue for me.  I should probably update/add a copy of the test with a copy initializer, but I don't really understand why the `var x: Rec = returnRec();` call should be invalid when `var x = returnRec();` is valid?

mppf:
That's a different question and has more to do with how we're implementing variable initialization. I agree it would be reasonable for `var x: Rec = returnRec()` to not invoke a copy initializer.

The error you are getting is coming from `insertAutoCopyTemps` in callDestructors. We should be totally removing that function since it's the old code (replaced by `ReturnByRef`). It's just that there are some non-obvious connections to other parts of the compiler. Even so, one of us should try to remove that code soon.

Is this issue holding you up, or is it a curiousity?

lydia-duncan:
It's not holding me up, it's just one of the failure modes for initializers currently and seemed relatively easy to tackle so I thought I would give it a whirl for a while.  I'll probably try what you're suggesting and if that doesn't work, then I will log it as a failure mode and move on

mppf:
I think removing `insertAutoCopyTemps` is likely to be a 1-week+ long project but I'd be happy to be surprised.

lydia-duncan:
Heh, fair enough.  Is the future good to go as is, then, or should I make that modification?  And do you think
```
var x: Rec;
x = returnRec();
```
should behave the same as `var x: Rec = returnRec();`?  Or is that more of a @noakesmichael question?

mppf:
I think it should behave the same only if we have some optimizations I've proposed in the past (called "expiring value optimization"). Those are not implemented now and so the version you have default initializes and then assigns. We'll need to discuss things like "expiring value optimization" and how to specify such things.

lydia-duncan:
I tried removing insertAutoCopyTemps by itself and that led to an error during returnRecordsByReferenceArguments due to some malformed AST it was trying to insert.  Since that function seemed like it was also no longer necessary, I tried removing it and its dependencies as well, and that mostly went fine - see the commit comments on [this branch](https://github.com/chapel-lang/chapel/compare/master...lydia-duncan:removeInsertAutoCopyTemps).  There were a few error categories for 36 tests, which seemed to be divided between:
- errors with associate arrays and domains
- failure to copy classes/records with array fields appropriately
- errors with opening files

I'm not sure if that means we are failing to cover those cases appropriately with the new implementation, or if there is code in other parts of the compiler for those cases that was relying on the precise actions of those old functions.  I think I'm going to punt on solving this for now.

Issue #6568, opened 2017-06-28
Discussion: insertion point of omitted initialization in initializers


lydia-duncan:
Brad suggested that it might be good to have the following code be valid:
```
class Params {
  var numProbSizes: int;
  var N: [1..numProbSizes] int;

  proc init() {
    numProbSizes = 3;
    // Option A
    for n in N do n = 10; // This line is problematic when placed in Phase 1
    // Option B
    super.init();
  }
}

```

The compiler is already inserting omitted initialization for the array, but the
current implementation places the omitted initialization right before the
super.init() call ("Option B"), making the `for` loop access a field before it has been
initialized.  If the omitted initialization was placed right after the last
initialized field ("Option A"), then the `for` loop would be valid.

Is this desirable, or letting users write a bug?  Should this be part of the
specification, or left as an implementation detail (and so maybe undefined
behavior)?

lydia-duncan:
@noakesmichael, @mppf, @vasslitvinov - thoughts?  Whenever I talk to Brad, I come away feeling like it is reasonable, but then when I think about it again, I go back to "but maybe that would be violating the Phase 1 rules and letting them access a field without initializing it".

bradcray:
If we make this switch, I think it needs to be specified.  Otherwise, people will likely rely on it and end up with non-portable code.

noakesmichael:
I continue to think it is an intriguing idea.  I think it is a little harder
to describe/specify than the current rule and non-trivially harder to
implement.  I suspect, but can’t prove. that there are some cases
in which the compiler would have to work a bit harder to generate
good error messages with this revised rule.


I think this is exactly the class of good idea that should be captured
and then reviewed in the context of having some experience
converting existing constructors.  I wonder how often this idea
would be valuable esp. in a phase2-is-default world.

I also think that this rule could be introduced later without breaking
any working code.  I think it would simply allow some initializers to
be re-written to be more concise.

MIke




On Jun 28, 2017, at 3:45 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


@noakesmichael<https://github.com/noakesmichael>, @mppf<https://github.com/mppf>, @vasslitvinov<https://github.com/vasslitvinov> - thoughts? Whenever I talk to Brad, I come away feeling like it is reasonable, but then when I think about it again, I go back to "but maybe that would be violating the Phase 1 rules and letting them access a field without initializing it".

bradcray:
[...]

PR #6569, opened 2017-06-28
Add a future with Brad's request for the omitted initialization insertion point

lydia-duncan:
Brad thinks we should have omitted initialization be inserted right after the
last explicit initialization, rather than just before the next or the
`super.init()` call (as it is today).  This would enable loops like the one in
this test to be Phase 1 operations, as the initialization would occur prior to
their operation.


lydia-duncan:
This future tracks #6568

PR #6571, opened 2017-06-29
Update tests to use explicit initializers instead of constructors, part 2

lydia-duncan:
Mike fixed the bug with `new` statements, so these constructors could
safely be converted to initializers.

Passed a clean checkout.


lydia-duncan:
@daviditen, would you mind taking a look at both of these?

daviditen:
Looks good.

lydia-duncan:
Thanks!

Lydia

PR #6572, opened 2017-06-29
Improve well-known function facility

mppf:
I ran into problems with `chpl_delete_error` being marked `export` when extending error handling for `on` statements. That motivated me to find an alternative way to preserve a function through resolution, other than by marking it `export`. Currently `export` indicates that all arguments are local (not wide) and that's not appropriate for `chpl_delete_error`.

This PR takes these steps, in different commits:
1) In preparation, factor well-known function logic out of parser.cpp into its own file
2) Add  a facility to that file by which later passes can query for a list of well known functions. Use this new function in resolution to avoid pruning important functions.
3) As a clean-up, move printModuleInit to the usual well known functions instead of finding it in special code in normalize.cpp.

Passed full local testing.


compiler/include/wellknown.h
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _WELL_KNOWN_H_
+#define _WELL_KNOWN_H_
+
```
noakesmichael: Curious if you have a preference for #include "symbol.h" vs class FnSymbol; here?  I favor the latter where
possible, I believe others prefer the former in general, and believe it's a case-by-case for some people.

compiler/AST/wellknown.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+  std::vector<FnSymbol*> fns;
+
+  int nEntries = sizeof(sWellKnownFns) / sizeof(sWellKnownFns[0]);
+
+  for (int i = 0; i < nEntries; ++i) {
+    WellKnownFn& wkfn = sWellKnownFns[i];
+    fns.push_back(*wkfn.fn);
+  }
+
```
noakesmichael: I am sure your code is right but I can't see why gPrintModuleInitFn wasn't pushed on fns by the general
loop over sWellKnownFns.

compiler/include/wellknown.h
mppf: I'll change it to the forward declaration, which I agree is generally better if it's that simple.

compiler/AST/wellknown.cpp
mppf: Oh, good catch. I had not included gPrintModuleInitFn in the sWellKnownFns array in an earlier version and forgot to remove this code when I added it there.

PR #6573, opened 2017-06-29
PRK Sparse Distributed

ben-albrecht:
This PR adds a distributed Sparse PRK to the shared-memory Sparse PRK, and refactors the code.

- Added distributed implementation (fused with shared kernel)
- Switched to 1-based indexing
- Encapsulated non-kernel code in functions
- Encapsulated domain-definitions in functions (to hide `if distributed` logic)
- Added documentation
- Performance testing added

### Performance-impacting changes

- `temp` for storing partial sums.

### TODO

- [x] Change vector-update to use block-like communication pattern (#6682).
- [ ] Report performance delta of shared-memory test before/after this PR
- [ ] multilocale performance testing
    - 16 locales
- [x] cleanup / void distribution definitions

### Future work
- Further investigate if `vector` is large enough to warrant updates in parallel
   - Update with `vector += vectorDom;`?
- Capture performance delta of iterating over a sparse vs. indexing into it in a separate set of tests

### DIFF
--- a/test/GRAPHFILES
+++ b/test/GRAPHFILES
@@ -169,10 +169,12 @@ studies/isx/isx-no-return.graph
 studies/isx/isx-per-task.graph
 release/examples/benchmarks/isx/isx-release.graph
 # suite: PRK benchmarks
-studies/prk/Synch_p2p/prk-synch_p2p.graph
-studies/prk/Synch_p2p/prk-synch_p2p-time.graph
+studies/prk/Sparse/prk-sparse.graph
+studies/prk/Sparse/prk-sparse-time.graph
 studies/prk/Stencil/prk-stencil.graph
 studies/prk/Stencil/prk-stencil-time.graph
+studies/prk/Synch_p2p/prk-synch_p2p.graph
+studies/prk/Synch_p2p/prk-synch_p2p-time.graph
 studies/prk/Transpose/prk-transpose.graph
 studies/prk/Transpose/prk-transpose-time.graph
 # suite: NAS Parallel benchmarks
--- a/test/studies/prk/Sparse/prk-sparse-time.graph
+++ b/test/studies/prk/Sparse/prk-sparse-time.graph
@@ -1,0 +1,5 @@
+perfkeys: Avg time (s): 
+files: sparse-csr.dat
+graphkeys: CSR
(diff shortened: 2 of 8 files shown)


ben-albrecht:
Made changes and attempted to capture remaining ideas from @chapel-lang/perf-team in PR message.

FYI - @e-kayrakli

e-kayrakli:
The way `vector` is handled seems different than it is in MPI1 version of the reference implementations. *If* MPI version to be followed, then the vector needs to be block-distributed and local copies need to be created/updated every iteration corresponding to the MPI_Allgather.

bradcray:
> If MPI version to be followed, then the vector needs to be block-distributed

(Based on heresay), I thought the MPI version used a 1D distribution of the matrix which required the input vector to be replicated everywhere (all processors need all elements) and the output vector to be block-distributed.

> If MPI version to be followed then... the local copies need to be created/updated every iteration

I agree with this and thought the plan was to check with the PRK team to see what the constraints were.

ben-albrecht:
> I agree with this and thought the plan was to check with the PRK team to see what the constraints were.

That's still the plan.

>> If MPI version to be followed, then the vector needs to be block-distributed

>(Based on heresay), I thought the MPI version used a 1D distribution of the matrix which required the input vector to be replicated everywhere (all processors need all elements) and the output vector to be block-distributed.

After talking to @e-kayrakli offline, I have a better understanding of what the reference version is doing: The `vector` *is* block-distributed in the reference version, each locale gathers a local copy of the full `vector` at the start of the kernel loop, trading update iterations for communication, when compared to our implementation.

I'll wait to hear from PRK team before proceeding.

e-kayrakli:
To elaborate more:

All the ranks allocate enough space to store the whole `vector` (a la ReplicatedDist), but then ranks only initialize (which I strongly believe is buggy in MPI1 implementation) and update their share of it (a la Block). After they update, they do an in-place Allgather on `vector` (a la whole array assignment minus local memcpy).

It gets a bit confusing if you want to find a 1:1 understanding of what it does. Well, it is C+MPI, so it is bound to be confusing...

ben-albrecht:
Even though this is close to being finished, I'm going to close the PR until I can set aside the time to wrap it up and merge.

PR #6574, opened 2017-06-29
Code generation fixes for on+throws

mppf:
These changes were developed in order to get on+throws working.

Before these changes, codegenWideAddr, when passed a class value, would return a class value with .isLVPtr = GEN_WIDE_PTR. That doesn't make sense because the wide class is already a value (our compiler treats class instances as values; e.g. classes are passed by "in" intent of the pointer value). This changes codegenWideAddr to return .isLVPtr = GEN_VAL for class instances and removes some accommodations in related code (mostly calls to codegenAddrOf that are no longer necessary).

Passed release/examples/primers/ and release/examples/users-guide/ with GASNet+fifo.
- [x] full local testing
- [x] full GASNet testing

### DIFF
--- a/compiler/AST/type.cpp
+++ b/compiler/AST/type.cpp
@@ -971,6 +971,10 @@ bool isClass(Type* t) {
   return false;
 }
 
+bool isClassOrNil(Type* t) {
+  if (t == dtNil) return true;
+  return isClass(t);
+}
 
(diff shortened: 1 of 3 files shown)


mppf:
@benharsh - I think this is ready for your review, thanks.

PR #6576, opened 2017-06-29
6/29 triage annotations

psahabu:
- XC reverting to gcc 6.3
- new tmp dir
- less locking on associative arrays
- regex leak improvements
- updateFluff fix

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -254,6 +254,8 @@ all:
   06/20/17:
     - text: revert gnu target compiler from 7.1.0 to 6.3.0
       config: Single node XC, 16 node XC
+  06/21/17:
+    - change tmp directory
 
 
 AllCompTime:
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -481,8 +481,6 @@ hpl_performance:
 hpl.hpcc2012.ml-perf:
   03/03/17:
     - Modified test as part of array views (#5338)
-  06/19/17:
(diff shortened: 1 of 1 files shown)


test/ANNOTATIONS.yaml
```diff
@@ -472,6 +481,8 @@ hpl_performance:
 hpl.hpcc2012.ml-perf:
   03/03/17:
     - Modified test as part of array views (#5338)
+  06/19/17:
+    - Reduce amount of locking in associative array dsiAccess (#6244)
```
benharsh: Did you verify this was the cause?

test/ANNOTATIONS.yaml
```diff
@@ -769,14 +788,10 @@ prk-stencil.ml-perf:
...
   06/06/17:
     - Increase the PRK stencil problem size (#6380)
-  06/13/17:
-    - Optimize updateFluff (#6418)
 
 prk-stencil-time:
   05/11/17:
     - Move chpl_getPrivatizedClass() into chpl-privatization.h (#6212)
-  06/13/17:
-    - Optimize updateFluff (#6418)
```
benharsh: Why remove these?

psahabu: I messed up the first time around, they didn't actually do anything to the stencil as far as I could tell. I can put them back in though.

test/ANNOTATIONS.yaml
psahabu: I didn't verify. Since it was a positive change, should I just remove it?

PR #6579, opened 2017-06-30
added isExternClassType

vasslitvinov:
With this change, these standard-library functions:

```chpl
  isClass()  isClassType()  isClassValue()
```

return `true` when invoked on an `extern` class or on `nil`.

### Is this desirable?

IMO this makes sense. `isClass` functions are user-oriented
and extern classes in most respects behave like Chapel classes.

Re `nil` - while it is somewhat special, it IS of any class type.
So let's have it be `isClass`, too.

What do you think?

### Implementation Outline and Questions

* Added PRIM_IS_EXTERN_CLASS_TYPE to the compiler and isExternClassType()
to the modules, following the lead of isUnionType().

* Needed to add something to `proc ==` for extern classes.
Before this change `==` involving an extern class would not resolve.
Now it needs to be resolvable because in IO.chpl in _write_one_internal()
we have:

```chpl
  if (isClassType(t) || chpl_isDdata(t)) && x == nil ....
```

where `t` is the type of `x`. So when `x` is an extern class,
it is now compared against `nil` whereas before it weren't.

* How should we do that?

The only way I could think of was a `proc ==(a,b)` where
`a` and `b` are completely generic, restricting it with a `where` clause:

```chpl
proc ==(a, b) where isClass(a) && isClass(b)
```

Interestingly (A), before this we did not have any `proc ==`
with completely generic formals. I am unsure whether introducing it
will be an issue w.r.t. the number of specializations/compilation time/...
At least it passes testing.

Interestingly (B), we cannot retire this one:

```chpl
proc ==(a: object, b: object)
```

That's because we need it when one of the actuals is a `sync` or `single`
whose underlying type is a class. In such case, the compiler sees this
as a valid candidate and inserts an implicit coersion that does readFE
or readFF. The completely-generic formal matches, too, however in that
case its type is a sync, so it does not pass the isClass() where-clause.

When the sync/single is over an EXTERN class, this still does not work.
If desired, we should add something like:

```chpl
proc ==(a, b) where isSyncOverClass(a) && isClass(b)
proc ==(a, b) where isClass(a) && isSyncOverClass(b)
```

Then it should be possible (desirable?) to retire the (object,object)
version.

Everything about `==` applies equally to `!=`.

What do you think?

* Another interesting question is: should users be able to use `==`
on one extern class and one non-extern class?
On two different extern classes?

* In ChapelIO.chpl in writeThisFieldsDefaultImpl() and
readThisFieldsDefaultImpl(), do not visit the superclass for
an extern class because, like `object`, it ain't got any.

A minor benefit is that now extern classes are printed in `{}`
like Chapel classes, not in `()` like records.

testing: linux64 --verify on full suite, gasnet on multilocale tests


mppf:
I'm not sure if `isClass()` should return true for extern classes or not. Also for `nil`. These seem like terminology questions. Usually I look to the language specification in such cases to represent whatever terminology choices we already made. Does the language specification / other documentation describe `extern class` as a kind of `class` or something apart? What about `nil`?

But for the fixes to `==` and `!=` here that are presumably your end goal, you could have just as well made `isExternClass` and `isNil` and used these in your where clause.

vasslitvinov:
My goals are unrelated to `isClass()` returning true for extern classes.

`isClass` is defined in online docs (not in the spec) as "Returns true if the argument is a class type or value."

The spec defines extern classes like so:

> An external type which is a pointer to a struct can be referred to from Chapel using an external class declaration.

Sounds like deciding whether an extern class `isClass` is still up to us.

Another interesting question is: should users be able to use `==` on one extern class and one non-extern class? On two different extern classes? (
My goals are unrelated to `isClass()` returning true for extern classes.

`isClass` is defined in online docs (not in the spec) as "Returns true if the argument is a class type or value."

The spec defines extern classes like so in 31.2.1 Shared Types:

> An external type which is a pointer to a struct can be referred to from Chapel using an external class declaration.

Sounds like deciding whether an extern class `isClass` is still up to us. Indeed extern classes are at the same time like and unlike Chapel classes.

Another interesting question is: should users be able to use `==` on one extern class and one non-extern class? On two different extern classes? (I added this to the PR message.)

vasslitvinov:
Closing this because:
* #6646 - we decided that `isClass()` should not return true for extern classes
* #6674 - we implemented a smaller change instead of this PR, simply adding `isExternClass()`

PR #6580, opened 2017-06-30
Clean up after chplCompile test

vasslitvinov:
Remove the generated executable after testing.

At first I had the .prediff also run the generated executable,
to test the compilation output. The prediff would say
"./samplefile -nl 1". However, while discussing with Sam,
I talked myself out of it. The reason being that on some
systems this may not work because of missing setup
like environment variables, launcher redirect/options, etc.

We also considered adding `--no-codegen`
to the compilation command. We decided
that it's prettier without it.


vasslitvinov:
@Spartee would you review?

Spartee:
Looks good to me! Thanks!

PR #6584, opened 2017-06-30
Change to IWR to support error handling

mppf:
error handling was creating a error:Error variable and passing it to a function argument with ref intent (but not ref type).  This fix adjusts IWR to widen the error variable in this case when the argument is also wide (instead of generating a temporary wide class).

- [x] full local testing
- [x] full local --no-local testing
- [x] full GASNet testing


mppf:
@benharsh - would you mind looking at this PR changing insertWideReferences? Thanks.

PR #6585, opened 2017-06-30
Better better match

### LINKED ISSUE #6578: Error compiling code using UtilReplicatedVar
### Summary of Problem
The code below fails to compile with the following error message ---
```
$CHPL_HOME/modules/standard/UtilReplicatedVar.chpl:126: In function 'rcReplicate':
$CHPL_HOME/modules/standard/UtilReplicatedVar.chpl:132: error: invalid ref return pair: mixing proc and iter
$CHPL_HOME/modules/internal/ChapelArray.chpl:2069: error: here
$CHPL_HOME/modules/standard/UtilReplicatedVar.chpl:132: error: here
$CHPL_HOME/modules/internal/ChapelArray.chpl:2078: error: here
```

Commenting out the line ``var D: domain(1);`` prevents the error, as does putting the definition of ``R`` into the main module.

### Steps to Reproduce

**Source Code:**

```chapel
use UtilReplicatedVar;
use Utils;

module Utils {

  record R {
    /* Note that the issue appears to be related to the domain.... if
       we comment this line out, it works again...

       Although defining R in the main module also works
    */
    var D: domain(1);
    var x : int;

    // Old-style constructor
    proc R() {
    }

    // No checks
    inline proc this(x: real): real {
      return 0.0;
    }
  }
}

proc setup() {
  return new R();
}

proc main() {
  var a : [rcDomain] R;
  rcReplicate(a, setup());
}
```

**Compile command:**
`chpl foo.chpl`

### Configuration Information
Fails on version 2659cc6b6aef4a623fbee, but I believe it works on 492429969fae6e26

mppf:
Fixes #6578.

isBetterMatch actually already does most of the work of computing if f1 is a better match than f2 *and* if f2 is a better match than f1.  This commit changes isBetterMatch to compareSpecificity which returns -1 to prefer f1, 0 if they are equally preferred, and 1 to prefer f2. Using this information we can correctly compute the set of most specific functions, which previously could in some cases include the wrong elements (and that is what is going wrong in issue #6578).

While there, I noticed that isBetterMatch/compareSpecificity was not correctly handling the case when both functions contained a most specific argument. In that case, the specificity is ambiguous (according to the language specification) and so the scope should be considered. See

        test/functions/ferguson/most-specific.chpl

which would have been an ambiguity error before this change - and now compiles - for an example.

Passed full local testing.


compiler/resolution/functionResolution.cpp
```diff
@@ -2201,14 +2182,23 @@ static void testArgMapping(FnSymbol* fn1, ArgSymbol* formal1,
...
  *
- * \return True if fn1 is a more specific function than f2, false otherwise.
+ * \return -1 if fn1 is a more specific function than f2
+ * \return 0 if fn1 and fn2 are equally specific
+ * \return 1 if fn2 is a more specific function than f1
  */
-bool isBetterMatch(ResolutionCandidate* candidate1,
+static
+int compareSpecificity(ResolutionCandidate* candidate1,
                           ResolutionCandidate* candidate2,
```
lydia-duncan: Minor nit: this changes the indentation level of the arguments below

PR #6589, opened 2017-06-30
Changes to parallel.cpp to support on+throw

mppf:
Before this change, the argument bundle for an on statement that can throw was storing an Error variable by value instead of by reference.

This PR adjusts for the situation by updating create_arg_bundle_class to consider the intent of the on-statement body formal. However the first attempt at doing so did not function correctly, because apparently parallel.cpp needs to generate QUAL_VAL fields for some on-function formals that are references. Since this seems to depend on const-ness, it's likely that moving that concern to RVF would be worthwhile. It's also likely that index variables need to be by value.

- [ ] full local testing
- [ ] full --no-local testing


mppf:
@benharsh - would you mind looking at this PR changing parallel.cpp? Thanks.

PR #6603, opened 2017-07-03
Initializer bug with inheritance and inlining

benharsh:
Add test for inheriting classes and initializers that fails during inlining

### DIFF
--- a/test/classes/initializers/generics/inheritance/inheritance-inline.chpl
+++ b/test/classes/initializers/generics/inheritance/inheritance-inline.chpl
@@ -1,0 +1,17 @@
+class Base {
+  proc init() {
+  }
+}
+
+class Fancy : Base {
+  proc init() {
+  }
+}
+
+class Simple : Base {
+  proc init() {
+  }
+}
+
+var f = new Fancy();
+var s = new Simple();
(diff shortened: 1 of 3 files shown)


benharsh:
@lydia-duncan , @noakesmichael : I couldn't find an existing test case for this bug. Does it seem familiar to either of you?

test/classes/initializers/generics/inheritance/inheritance-inline.chpl
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+class Simple : Base {
+  proc init() {
+  }
+}
+
+var f = new Fancy();
+var s = new Simple();
```
lydia-duncan: You'll want to delete these instances to avoid memory leaks, btw

lydia-duncan:
That doesn't seem familiar to me at all, that's a really weird failure.

lydia-duncan:
Oh, I think I know why.  Mike, this seems related to making the "init" function inlined when we create the "_new" function.  It probably thinks the implicit "super.init()" call is recursive, but that is still strange.

benharsh:
OK, classes are deleted (good catch). Do you want to add this to the big list of issues, or should it be a separate one?

PR #6604, opened 2017-07-03
Migrate some comm-count tests to performance tests

mppf:
This PR takes the following steps:
 * enable --no-cache-remote compiler option
 * rework test/performance/ferguson as multi-locale performance tests instead of correctness tests. Note that these tests are still correctness tests; in that setting they verify that the number of communications of particular types do not exceed a maximum that is stored in the .chpl file. Otherwise, they are performance tests and will make a graph of communication counts. The idea here is that each of these tests is intended to graph a particular type of communication (e.g. GETs) but it can meanwhile assert that there are not too many communications of the type that is not graphed.

- [x] correctness testing of the new directory with GASNet
- [x] performance testing of the new directory with GASNet
- [x] full local testing


lydia-duncan:
If these are intended as multi-locale performance tests, I think you would have wanted to update ML-GRAPHFILES instead of GRAPHFILES

mppf:
@lydia-duncan - thanks for the pointer, I'll need to do that.

compiler/main/driver.cpp
```diff
@@ -775,7 +775,7 @@ static ArgumentDescription arg_desc[] = {
 
  {"", ' ', NULL, "Optimization Control Options", NULL, NULL, NULL, NULL},
  {"baseline", ' ', NULL, "Disable all Chapel optimizations", "F", &fBaseline, "CHPL_BASELINE", setBaselineFlag},
- {"cache-remote", ' ', NULL, "Enable cache for remote data (must be enabled specifically)", "F", &fCacheRemote, "CHPL_CACHE_REMOTE", setCacheEnable},
+ {"cache-remote", ' ', NULL, "[Don't] enable cache for remote data", "N", &fCacheRemote, "CHPL_CACHE_REMOTE", setCacheEnable},
```
benharsh: I think you will need to update compflags/bradc/help/userhelp.good

modules/standard/CommDiagnostics.chpl
```diff
@@ -332,6 +332,61 @@ module CommDiagnostics
...
+    Get the total number of execute_on, execute_on_fast, or execute_on_nb
+    operations on any locale.
+   */
+  proc totalOns(D:[LocaleSpace] commDiagnostics) : int {
+    var t = totalCounts(D);
+    return t.execute_on + t.execute_on_fast + t.execute_on_nb;
+  }
+
+
+  /*
```
benharsh: These additions seem fine to me, but I think there should be a broader discussion as this changes the API for a standard module.

test/GRAPHFILES
```diff
@@ -308,5 +308,26 @@ statements/lydia/ifVersusSelect.alternates.graph
...
+performance/ferguson/remote-record-write.graph
+performance/ferguson/remote-tuple-read-1.graph
+performance/ferguson/remote-tuple-read.graph
+performance/ferguson/remote-tuple-read-small-1.graph
+performance/ferguson/remote-tuple-read-small.graph
+performance/ferguson/remote-tuple-write-1.graph
+performance/ferguson/remote-tuple-write.graph
+performance/ferguson/remote-tuple-write-small-1.graph
+performance/ferguson/remote-tuple-write-small.graph
+performance/ferguson/remote-two-array-read.graph
```
benharsh: I second Lydia's remark about ML-GRAPHFILES

modules/standard/CommDiagnostics.chpl
benharsh: In the short-term it seems like this could simply move into CommUtil

util/start_test
```diff
@@ -672,6 +672,7 @@ def set_up_environment():
         #   when the memkind package is being used (KNL, e.g.)
         # - any locale model other than 'flat' (because we'll need
         #   libnuma which is only available as a .so)
+        # MPF: this causes problems on my Ubuntu system...
```
benharsh: Should this be a GitHub issue?

test/performance/ferguson/remote-record-write-1.graph
```diff
@@ -0,0 +1,5 @@
+perfkeys: PUTs:, PUTs:
+files: remote-record-write-1-c.dat, remote-record-write-1-c-cache.dat
+graphkeys: c PUTs, c-cache PUTs
+ylabel: Count
+graphtitle: remote-record-write-1
```
benharsh: Is there a way we could better distinguish tests with and without the "-1" suffix? It seems like it could be confusing when discussing/searching graphs.

util/start_test
ronawho: Yeah, please make an issue out of this or let's try to fix whatever is going wrong

PR #6605, opened 2017-07-03
fix throwing errors from catch blocks

psahabu:
## motivation
An overlooked part of the initial error handling design was the semantics of a throwing call is inside a `catch` block. For example:

``` chapel
try {
  writeln("outer try");
  try {
    writeln("inner try");
    throwAnError();
  } catch {
    writeln("inner catch");
    throwAnError();
  }
} catch {
  writeln("outer catch");
}
```
Currently if a `throw` is made inside of a `catch` block, it sends the error to the scope outside of the matching `try` block. This change brings throwing calls in line with that policy, while also ensuring that throwing calls and `throws` can be made from a `catch` of arbitrary depth.

### traversal order
Statements inside of a `catch` are in the error handling scope outside the `try` rather than inside of it. To simplify tracking the correct scope for throwing calls and `throws`, the new traversal order puts `exitTryStmt()` before `enterCatchStmt()`.

### catches stack
Before, the structure of `catch` blocks was lowered in `exitTryStmt()`, after the content of each block was lowered with `enterCatchStmt()`. When the traversal order changed, the content of the `catch` blocks was not properly lowered, so a workaround was needed.

`CatchStmt` lowering was already in a separate function, `lowerCatches()`, but it was dependent on arguments available to `exitTryStmt()`. To get around this, `TryInfo` was expanded to include additional information about the given `TryStmt`, and a separate stack of `TryInfo` tracks which catches still need to be lowered. `exitCatchStmt()` will check if all the catches have been lowered, then pull from the catches stack and pass the `TryInfo` to `lowerCatches()`.

### AList.empty()
`exitTryStmt()` will call `lowerCatches()` if the catch list is empty, since there will be no `exitCatchStmt()`. The catch list is an `AList`, so an `empty()` function has been added to `AList` for convenience.


compiler/passes/errorHandling.cpp
```diff
@@ -131,29 +131,29 @@ class ErrorHandlingVisitor : public AstVisitorTraverse {
...
   LabelSymbol*        epilogue;
-  bool                insideCatch;
 
-  AList     lowerCatches      (TryStmt*   tryStmt,  VarSymbol* errorVar,
-                               TryInfo*   outerTry);
-  AList     setOutGotoEpilogue(VarSymbol* error);
-  AList     errorCond         (VarSymbol* errorVar, BlockStmt* thenBlock,
-                               BlockStmt* elseBlock = NULL);
-  CallExpr* haltExpr          ();
+  void   lowerCatches      (TryInfo    info);
```
mppf: I'd expect you to pass a (const) reference to TryInfo for this function, vs the current copy.

psahabu:
The catches stack helps with tracking which catches still need to be lowered. While the try stack is good for tracking nested `try`, it doesn't naturally track nested `catch`.

PR #6606, opened 2017-07-03
Enable privatization for SparseBlock

benharsh:
This commit enables privatization for the SparseBlock distribution,
resulting in significantly less communication.

A minor change was made to dsiAccess to check index membership in a
dense domain instead of a sparse domain to improve performance.

[ written by @e-kayrakli, committed by @benharsh ]

Testing:
- [x] full local
- [x] subset with gasnet
  - [x] distributions/bradc/spsBlk
  - [x] sparse/engin/
  - [x] users/engin/

### DIFF
--- a/modules/dists/SparseBlockDist.chpl
+++ b/modules/dists/SparseBlockDist.chpl
@@ -60,7 +60,7 @@ class SparseBlockDom: BaseSparseDomImpl {
   type sparseLayoutType;
   param stridable: bool = false;  // TODO: remove default value eventually
   const dist: Block(rank, idxType, sparseLayoutType);
-  const whole: domain(rank=rank, idxType=idxType, stridable=stridable);
+  var whole: domain(rank=rank, idxType=idxType, stridable=stridable);
   var locDoms: [dist.targetLocDom] LocSparseBlockDom(rank, idxType, stridable,
       sparseLayoutType);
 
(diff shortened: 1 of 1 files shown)


benharsh:
@e-kayrakli : does all this look good to you?

@ben-albrecht : have you checked to see if this reduces comm counts for the PRK you were looking at?

e-kayrakli:
@benharsh -- There is just one thing that I don't have a clear answer myself:

Do we need to fully privatize a SparseBlockDom instance? From a more practical standpoint, do internal "index arrays" need to be copied to the all privatized instances?

"No" seems to be the clear answer because of obvious overhead of doing something like that, and hard to answer questions like "do you reprivatize after indices added/removed?". Besides everything seems to be working without doing that. At the same time I cannot be 100% certain that this would not break any assumption regarding privatized domains.

It is just something I cannot be sure myself. If you cannot think of any problem with that, then I am good.

benharsh:
I think those are interesting questions, ones that may be answered as the SparseBlock distribution is improved. For example, DefaultSparseDom and SparseBlockDom don't seem to support dsiNumIndices. In the current implementation of SparseBlockDom, we only appear to update the current locale's ``nnz`` which could be problematic later.

I don't have any clear answers at this time, but I think this PR is a good first step for privatization and should significantly reduce comm counts.

ben-albrecht:
> @ben-albrecht : have you checked to see if this reduces comm counts for the PRK you were looking at?

Some fun chplvis results:


branch | max Comms in `sparse.chpl` with `-nl 4` 
--------|--------------
`chapel-lang:master` | `53574` 
`benharsh:sparseBlockPrivatization` | `6`

PR #6607, opened 2017-07-04
throw errors from a function that returns values

psahabu:
It was not possible to throw errors from functions that returned values due to a bug in how error handling code was inserted. `getStmtExpr()` assures that the error handling code will be inserted in a place where it will be executed.

The NULL check is needed because of changes in another open PR, #6605, which may cause `parentExpr` to be NULL.

This change does _not_ enable single-line assignments, such as:
``` chapel
try! var x = throwOrValue(a);
```


compiler/passes/errorHandling.cpp
```diff
@@ -285,6 +285,9 @@ bool ErrorHandlingVisitor::enterCallExpr(CallExpr* node) {
 
       VarSymbol* errorVar    = NULL;
       BlockStmt* errorPolicy = new BlockStmt();
+      Expr*      insert      = node->getStmtExpr();
+      if (insert == NULL)
+        insert = node;
```
mppf: I don't have a problem with this going in as a bug fix, but I don't really understand why parentExpr can be NULL while node->insertBefore will still work, for a CallExpr.

psahabu: I have no idea why `node->insertAtTail()` still works, but `node->insertBefore()` was switched to `insert->insertBefore()`.

psahabu:
Passes linux64+flat testing.

compiler/passes/errorHandling.cpp
mppf: Right but I'm talking about the case when node->parentExpr == NULL and so insert == node. Then `insert->insertBefore()` might not be reasonable.

I'm not so concerned about `node->insertAtTail()`.

psahabu: Oh, duh. That is concerning.

PR #6608, opened 2017-07-05
Fix CoMD .good file

benharsh:


benharsh:
@ben-albrecht : can you review?

test/studies/comd/llnl/CoMD.prediff
```diff
@@ -6,17 +6,36 @@ test_name = sys.argv[1]
...
+vals['   Initial energy  :'] = -1.166063303477
+vals['   Final energy    :'] = -1.166049767266
+vals['   eFinal/eInitial :'] = 0.999988
+
+def verify(line):
+    info = line.split(' ')
+    pre = ' '.join(info[:-1])
+    if pre in vals:
+        last = float(info[-1])
+        if abs(last - vals[pre]) < EPSILON:
```
ben-albrecht: Could alternatively use [`math.isclose()`](https://docs.python.org/3/library/math.html#math.isclose)

benharsh: Ah, interesting. I don't think we can rely on python 3 for start_test yet though 😢

ben-albrecht: Good point - didn't even realize it's python3-only..

PR #6611, opened 2017-07-05
Skip LLNL CoMD for memory leaks testing

benharsh:
The original reference does not delete classes, so just skip for memory leaks testing


ronawho:
👍

Issue #6530, opened 2017-06-23
Convert tests from using explicit constructors to initializers


lydia-duncan:
Here are all the tests that rely on explicit constructors that don't have an
equivalent in the initializers directory (as of June 21, 2017,  there are
132 of these tests).  These should be updated to use initializers.  That
may not be possible yet for all of them; if so, it is sufficient to note *why*
they couldn't be updated yet and come back to them when that issue is
resolved.

If a constructor is converted that could trivially be Phase 1 (but isn't
because that would require writing super.init() at the end of the body),
please mention it at [this url](https://gist.github.com/lydia-duncan/5842808446e845218646ca3fe5acfada).

If a constructor is converted that could benefit from param and type fields
being "phase independent", please mention it at [this url](https://gist.github.com/lydia-duncan/11f694326f7f9036518c3aa1a2ca4b08).

If a constructor is converted that could benefit from const fields being
reassignable in Phase 2, please mention it at [this url](https://gist.github.com/lydia-duncan/5943b5225dfe57a51e299cece39f13f3)

Note, this list does not include futures.

Known failure categories (i.e., if a test cannot use initializers because of
these reasons, just mark it and move on):
- nested records and classes (mostly just when generics are involved)
  - [ ] test/studies/amr/lib/amr/BergerRigoutsosClustering.chpl
- secondary initializers
- inheritance when both parent and child define a zero-argument initializer
- inheritance when any class in the tree does not explicitly define an initializer
  - but only when --force-initializers is not thrown (though there are some limits
    there as well, so it might be best to punt anyways.  Ask me if you want to try
    anyways)
    - [ ] distributions/bradc/assoc/userAssoc-domain-priv-check
    - [ ] distributions/bradc/assoc/userAssoc-array-domain-zipper
    - [ ] distributions/bradc/assoc/userAssoc-array-noelt
    - [ ] distributions/bradc/assoc/userAssoc-array-priv-check
    - [ ] distributions/bradc/assoc/userAssoc-basic-array
    - [ ] distributions/bradc/assoc/userAssoc-basic-domain
    - [ ] distributions/bradc/assoc/userAssoc-dom-operations
    - [ ] studies/labelprop/labelprop-tweets (depends on the userAssoc helper module)
- not yet supporting certain error handling statements
  - [ ] test/studies/comd/llnl/utils/forceeam.chpl
- using nil as the default value of an argument for a class initializer
  - [ ]  test/users/shetag/objRefAsFormArg.chpl
[...]

PR #6613, opened 2017-07-06
Convert bradc tests to init

bradcray:
This takes a pass over tests in issue #6530 that have `bradc` in the path, attempting to convert them from constructors to initializers.  Most cases were easy, though a few were more challenging / not possible (that I could tell).

Specifically, converting helper module `test/studies/hpcc/common/bradc/BradsBlock1DPar.chpl` led to two issues:

* the first was that I introduced a `this.idxType = idxType;` line into `LocBlock1DDist`s `init()` routine which led to the compiler complaining that `idxType` had not been initialized.  What it's really trying to say is that it wasn't initialized in phase 1 of the initializer, but to a programmer who's not familiar with our phase 1 vs. phase 2 semantics, this would be confusing since it apparently is initialized.  I filed a new "error message" future test against this called `test/classes/initializers/generics/genericInitError.chpl` and can file an issue against it if Mike / Lydia agree with the intent and can confirm there isn't an outstanding issue already.

* the second is that the initialization `boundingBox = bbox` in `Block1DDist`s `init()` routine results in an error saying `cannot assign expression of type domain(1,int(64),false) to field of type _unknown` even though `boundingBox` is declared to be `domain(1, idxType)`.  Interestingly, the line given for this error is wrong (the initializer's `proc` declaration line) if the assignment is in phase 2, but correct if it's in phase 1.

The rest were fairly easy to convert.  For future reference, cases where I had to use a super.init() which seemed regrettable included:

* initializing a `const` field (in `test/parallel/sync/bradc/syncArrInit2.chpl`)
* initializing generic fields (in `test/reductions/bradc/manual/threeclasstypes-construct*.chpl`)

### DIFF
--- a/test/parallel/sync/bradc/syncArrInit2.chpl
+++ b/test/parallel/sync/bradc/syncArrInit2.chpl
@@ -1,8 +1,9 @@
 class C {
   const irng: int;
 
-  proc C(v1: int) {
+  proc init(v1: int) {
     irng = v1;
+    super.init();
   }
 }
 
--- a/test/reductions/bradc/manual/threeclasstypes-construct.chpl
+++ b/test/reductions/bradc/manual/threeclasstypes-construct.chpl
@@ -15,10 +15,11 @@ class mysumreduce {
     return x;
   }
 
-  proc mysumreduce(type in_t) {
+  proc init(type in_t) {
     intype = in_t;
     statetype = in_t;
     outtype = in_t;
+    super.init();
   }
 }
 
--- a/test/reductions/bradc/manual/threeclasstypes-construct.good
+++ b/test/reductions/bradc/manual/threeclasstypes-construct.good
@@ -1,3 +1,1 @@
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'intype'
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'statetype'
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'outtype'
+result is: 55
--- a/test/reductions/bradc/manual/threeclasstypes-construct2.chpl
+++ b/test/reductions/bradc/manual/threeclasstypes-construct2.chpl
@@ -15,7 +15,11 @@ class mysumreduce {
     return x;
   }
 
-  proc mysumreduce(type intype, type statetype=intype, type outtype=intype) {
+  proc init(type intype, type statetype=intype, type outtype=intype) {
+    this.intype = intype;
+    this.statetype = statetype;
+    this.outtype = outtype;
+    super.init();
   }
 }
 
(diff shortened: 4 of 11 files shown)


bradcray:
Attn @lydia-duncan and @noakesmichael.  Also, I'd be happy to continue on and do the other chameneos tests as part of this PR which I suspect are really similar to the one I did here.

lydia-duncan:
Thanks, Brad!  Addressing your PR comments first:
1) the idxType issue: you're right that the error message should be more specific and let you know that it is trying to insert omitted initialization for that field in Phase 1 while the body itself is Phase 2.  I'll add that initializer to the list of "initializers that would benefit from Phase 1 as the default".
2) I believe the issue with the boundingBox is due to its reliance on a previous (type) field.  We also run into this issue for test/classes/initializers/base-class-generic-args.chpl's tuple field "r".  I'm currently working on an effort to get generic initializers and concrete initializers to share more code, which should allow them to make use of Mike's improvements for prior field access and solve this issue.  That's odd about the line number, though.

lydia-duncan:
For 1, there is not an outstanding issue currently, so far as I am aware.

test/parallel/sync/bradc/syncArrInit2.chpl
```diff
@@ -1,8 +1,9 @@
 class C {
   const irng: int;
 
-  proc C(v1: int) {
+  proc init(v1: int) {
     irng = v1;
+    super.init();
```
lydia-duncan: I've marked this as benefiting from Phase 1 as the default.

test/reductions/bradc/manual/threeclasstypes-construct.chpl
```diff
@@ -15,10 +15,11 @@ class mysumreduce {
     return x;
   }
 
-  proc mysumreduce(type in_t) {
+  proc init(type in_t) {
     intype = in_t;
     statetype = in_t;
     outtype = in_t;
+    super.init();
```
lydia-duncan: I've also marked this as benefiting from Phase 1 as the default

test/reductions/bradc/manual/threeclasstypes-construct2.chpl
```diff
@@ -15,7 +15,11 @@ class mysumreduce {
     return x;
   }
 
-  proc mysumreduce(type intype, type statetype=intype, type outtype=intype) {
+  proc init(type intype, type statetype=intype, type outtype=intype) {
+    this.intype = intype;
+    this.statetype = statetype;
+    this.outtype = outtype;
+    super.init();
```
lydia-duncan: And this as well

test/reductions/bradc/manual/threeclasstypes-construct.good
```diff
@@ -1,3 +1 @@
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'intype'
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'statetype'
-threeclasstypes-construct.chpl:18: error: constructor for class 'mysumreduce' requires a generic argument called 'outtype'
+result is: 55
```
lydia-duncan: I'm excited that initializers allow this test to work :)

lydia-duncan:
This looks good to go to me!  If you want to update the other chameneos tests as well, I won't stop you :)

bradcray:
Filed issue #6632 for the future I filed here.

PR #6616, opened 2017-07-06
Supporting throw + on

mppf:
This PR takes a series of steps to improve error handling with a goal of providing initial support for a 'throw' that propagates out of an 'on' statement.

* adds canBlockThrow as a way of computing if a block could throw
* in lowerErrorHandling, detects compiler-generated on statement functions that can throw, and marks these as `throws`.
* Removes export from `chpl_delete_error` since it should not assume argument is a narrow pointer.
* Adds a simple workaround for a c-compilation error - uses PRIM_ASSIGN in one place in error handling lowering.
* Adds a simple test of on + throws
* Moves error handling checking to new CanThrowVisitor; changed fatal errors in ErrorHandlingVisitor to internal errors (since these should all be found by the previous checking).

Passed full local testing

The new test, test/errhandling/mferguson/on-throw.chpl, passes with quickstart+gasnet.

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -445,13 +445,6 @@ private:
   bool fnCanThrow; // only used for error checking
 
   bool   catchesNotExhaustive(TryStmt* tryStmt);
-
-  AList     setOutGotoEpilogue(VarSymbol* error);
-  AList     errorCond         (VarSymbol* errorVar, BlockStmt* thenBlock,
-                               BlockStmt* elseBlock = NULL);
-  CallExpr* haltExpr          ();
-
-  CanThrowVisitor();
 };
 
 CanThrowVisitor::CanThrowVisitor(bool inThrowingFn, bool makeCompileErrors) {
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -524,7 +524,7 @@ bool CanThrowVisitor::enterCallExpr(CallExpr* node) {
     canThrow = true;
 
     if (insideTry) {
-      // OK. non-exaustive try handled in try handling.
+      // OK, error checking for this case done in try handling
     } else if (fnCanThrow == true) {
       // OK, fn can throw
     } else if (errors == true) {
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -126,6 +126,11 @@ try {
 }
 */
 
+
+// Static functions
+static bool canBlockThrow(BlockStmt* blk);
+
+
 namespace {
 
 class ErrorHandlingVisitor : public AstVisitorTraverse {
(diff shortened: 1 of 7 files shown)


compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +415,149 @@ void lowerErrorHandling() {
...
+  bool   canThrow;
+  bool errors;
+  bool fnCanThrow; // only used for error checking
+
+  bool   catchesNotExhaustive(TryStmt* tryStmt);
+
+  AList     setOutGotoEpilogue(VarSymbol* error);
+  AList     errorCond         (VarSymbol* errorVar, BlockStmt* thenBlock,
+                               BlockStmt* elseBlock = NULL);
+  CallExpr* haltExpr          ();
```
psahabu: I'd remove these declarations since they're unused.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +415,149 @@ void lowerErrorHandling() {
...
+          USR_FATAL_CONT(node, "throwing call without try or try! (strict mode)");
+        }
+        // not in a try
+        canThrow = true;
+      }
+    }
+  } else if (node->isPrimitive(PRIM_THROW)) {
+    canThrow = true;
+
+    if (insideTry && !insideCatch) {
```
psahabu: I was concerned that this wouldn't work because of how `ErrorHandlingVisitor` is structured with the catches stack, but I couldn't write a case where this fails.

mppf: I don't think insideCatch belongs here any more. Let me see if it works when it is removed.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +415,149 @@ void lowerErrorHandling() {
...
+  tryDepth--;
+
+  // is it an exhaustive catch?
+
+  bool nonExhaustive = catchesNotExhaustive(node);
+
+  if (node->tryBang()) {
+    canThrow = false;
+  } else {
+    canThrow = nonExhaustive;
```
mppf: I like it better as a separate function. Keeps functions from getting too long, separates analysis from logic, and the function name offers some documentation about what's going on.

psahabu: Oh sorry, I'm bad with technical language. I totally agree with the separate function, originally I meant inline the call instead of assigning the `bool` up top.

However now that we're doing error checking with `catchesNotExhaustive()`, it's fine as it is.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +422,148 @@ void lowerErrorHandling() {
...
+        }
+        // not in a try
+        canThrow = true;
+      }
+    }
+  } else if (node->isPrimitive(PRIM_THROW)) {
+    canThrow = true;
+
+    if (insideTry) {
+      // OK. non-exaustive try handled in try handling.
```
psahabu: *exhaustive

Good call removing the `insideThrows` logic.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +422,148 @@ void lowerErrorHandling() {
...
+
+  return !hasCatchAll;
+}
+
+bool CanThrowVisitor::enterCatchStmt(CatchStmt* node) {
+  return true;
+}
+
+void CanThrowVisitor::exitCatchStmt(CatchStmt* node) {
+}
```
psahabu: You can remove `enter` and `exitCatchStmt`, as these are the same as the `AstVisitorTraverse` implementations.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +422,148 @@ void lowerErrorHandling() {
...
+
+class CanThrowVisitor : public AstVisitorTraverse {
+
+public:
+  CanThrowVisitor       (bool inThrowingFn, bool makeCompileErrors);
+
+  virtual bool enterTryStmt  (TryStmt*   node);
+  virtual void exitTryStmt   (TryStmt*   node);
+  virtual bool enterCatchStmt(CatchStmt*   node);
+  virtual void exitCatchStmt (CatchStmt*   node);
```
psahabu: You can remove the `enter` and `exitCatchStmt` declarations, since you're just using `AstVisitorTraverse` implementations.

compiler/passes/errorHandling.cpp
```diff
@@ -396,3 +422,148 @@ void lowerErrorHandling() {
...
+    canThrow = nonExhaustive;
+    if (errors && tryDepth==0 && nonExhaustive && !fnCanThrow) {
+      USR_FATAL_CONT(node, "try without a catchall in a non-throwing function");
+    }
+  }
+}
+
+bool CanThrowVisitor::catchesNotExhaustive(TryStmt* tryStmt) {
+
+  bool       hasCatchAll = false;
```
psahabu: The spacing is kinda funny on this line, not a big deal though.

PR #6618, opened 2017-07-06
Clarify which field is generating type mismatch error

bradcray:
In running into the error that this change touches, I couldn't figure out
which field was being complained about.  For that reason, I updated
the error message to have it name the field.

Unfortunately, no existing tests exercise this error message, and I was
unable to create a test that did reproduce it in a few tries (and the case
that caused me to run into the error doesn't serve as a good test even
though it exercises the change because it's an error that I don't think
we should be generating).  So this code remains untested.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -4143,8 +4143,9 @@ static void handleSetMemberTypeMismatch(Type*     t,
 
     } else {
       USR_FATAL(userCall(call),
-                "cannot assign expression of type %s to field of type %s",
+                "cannot assign expression of type %s to field '%s' of type %s",
                 toString(t),
+                fs->name,
                 toString(fs->type));
     }
   }
@@ -4252,8 +4253,9 @@ static void resolveInitField(CallExpr* call) {
 
     } else {
       USR_FATAL(userCall(call),
-                "cannot assign expression of type %s to field of type %s",
+                "cannot assign expression of type %s to field '%s' of type %s",
                 toString(t),
+                fs->name,
                 toString(fs->type));
     }
   }


bradcray:
This looks like a job for either @mppf or @noakesmichael based on who's touched this code most recently.  Also, if you have suggestions for how to trigger the error in a test, I'd like to file one.  So far, I've been unable to come up with such a case.

mppf:
@bradcray - this looks "trivial" to me and I certainly don't object to it. It's not too surprising to me that you don't see any tests with the error since the type mismatch would be detected as a problem with a call to a field accessor function.

noakesmichael:
+1


On Jul 6, 2017, at 9:36 AM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


@bradcray<https://github.com/bradcray> - this looks "trivial" to me and I certainly don't object to it. It's not too surprising to me that you don't see any tests with the error since the type mismatch would be detected as a problem with a call to a field accessor function.

lydia-duncan:
Brad - I believe this doesn't have any tests that trigger it because the detection is happening on PRIM_SET_MEMBER, which is only inserted by the compiler (for iterators, for default constructors, for initializers, for some parallel statements, etc.).  The error you are encountering happens when the compiler has done something wrong with the starting type for the field.

bradcray:
Lydia -- Does that suggest it should be an INT_FATAL() rather than a USR_FATAL() since it will necessarily be referring to code that the user is not responsible for?

lydia-duncan:
I believe so, though that may change as the implementation does - for instance, if we stop relying on PRIM_MOVE/PRIM_ASSIGN to detect the type mismatches, then this code might become a line of defense for fields.  It would most likely be better to have a comment to that effect rather than leaving it as a user facing error, though.

PR #6619, opened 2017-07-06
File test guarding against internal error backslide

bradcray:
When I wrote this test a few days ago, it generated an internal error.  @lydia-duncan just fixed that issue in PR #6626 which permits me to file this as a guard against backsliding in this area rather than a future (thanks!)


bradcray:
Attn @lydia-duncan and @noakesmichael.  I'll file an issue against this as well if there isn't one already.

lydia-duncan:
I believe this one is on my radar and has a version for enum constants (test/classes/initializers/phase1/enumFieldNoUseBad.chpl), but it wouldn't hurt to have this variation submitted as well.  Basically, concrete initializers don't expect to encounter unresolved sym expressions, but there still could be some at normalize, so we internal error there instead of generating the nice user-facing error that resolution would give us.

lydia-duncan:
The test is good to merge

PR #6621, opened 2017-07-06
Improve overloads of functions with partially generic arguments

### LINKED ISSUE #6641: Disambiguation of Partially Generic Arguments
Consider the following program:

``` chapel
proc test(arg:[]) where true {
  writeln("generic [] version");
}
proc test(arg:[] int) where true {
  writeln("[] int version");
}

var A:[1..100] int;

test(A);
```

What should this program do? The argument expression `arg:[] int` declares a generic argument that has an integer element type. It could still have any domain type. That might be called a partially generic argument.

The language specification describes how the best function is chosen from a series of overloads in 13.13.3 Determining More Specific Functions. However, it considers only totally generic and totally concrete arguments. It does not consider partially generic arguments.

Nonetheless the compiler supports partially generic arguments. However, it currently does so by adding a `where` clause to the function. Which seems to work on the surface, but it changes the way the function will be selected in an un-principled way (and in particular this is the reason the code listing at the top of this issue results in ambiguity even though the `arg:[] int` argument is *more concrete*).

I think that we should:
 a) Extend the compiler to directly prefer partially generic arguments over fully generic arguments
 b) Update the language specification to consider partially generic arguments

In particular, I think that the program above should result in output:
```
[] int version
```

mppf:
This PR implements a change discussed in issue #6641.

### A User-facing behavior change:
``` chapel
proc t1(A:[]) where true {
  writeln("t1(A:[]");
}
proc t1(A:[] int) {
  writeln("t1(A:[] int)");
}

var A:[1..100] int;

t1(A);
```

The above program would be ambiguous before this change but after this change would output
```
t1(A:[] int)
```

I think that this change is in line with the most-specific-argument mapping rules described in  the language specification 13.13.3 Determining More Specific Functions. However it would probably need to be updated to discuss such partially generic arguments.

### What's going on in the implementation?

Resolution disambiguates functions based upon whether or not they have a `where` clause. But, in certain situations, the normalizer will add a `where` clause to a function even if one did not exist before.

Since the `where` clause was added by the compiler, it's not reasonable to expect Chapel programmers to understand the implications on resolving otherwise ambiguous functions.

This PR
 * adds FLAG_COMPILER_ADDED_WHERE and FLAG_NOT_FULLY_GENERIC and sets these in the relevant parts of normalize
 * checks FLAG_COMPILER_ADDED_WHERE in the resolution code that prefers functions using where clauses, to ignore compiler-added where clauses
 * checks FLAG_NOT_FULLY_GENERIC in resolution code deciding which argument is more specific
 * adds a test that would fail before this change

Closes #6641.

- [x] full local testing
- [x] spec change is in PR #6659


mppf:
See issue #6641.

PR #6626, opened 2017-07-07
Allow unresolved syms through initializer normalization

lydia-duncan:
Prior to this change, if there was an unresolved symbol, we would just abort
with an internal error when encountering the AST node for it during the handling
of initializers.  This meant that code which would have later given a reasonable
error message instead didn't give a helpful one.  I updated the handling of
these AST nodes to allow them through so that function resolution would give us
the appropriate error message.

The motivating future now passes with a slight fix to the expected error message
to reflect that the error was originating inside the initializer.

This partially resolves the error in #6619, though the error message there also
has the initializer prefix since we handle the type of the field in the initializer body
rather than in place at the field declaration.

Passed a full std/ paratest run


lydia-duncan:
@bradcray - heads up, this should improve the output for your new future, but note the caveat above

bradcray:
Cool, this fixed the issue.  I'll still file the test, but as a working case rather than a future (which saves me from having to write up an issue).  Thanks!

lydia-duncan:
Excellent!

PR #6630, opened 2017-07-07
Update insertDynamicDispatchCalls()

noakesmichael:
This PR updates insertDynamicDispatchCalls() to resolve a failure case for super.init() and to
disable dynamic dispatch for uses of this.init() (c.f. convenience initializers in Swift).

In early versions of Chapel every invocation of a class method implemented dynamic dispatch.
First the method is resolved based on the static type of the receiver.  If there were available
overrides for that method, then the compiler would convert the static call in to a sequence of
conditional statements that selected the appropriate method.

In 2009 the implementation was extended to include a more traditional dynamic-dispatch table.
The primary motivation for this was reported to be that there were cases in which the number
of overrides might lead to an excessive amount of generated code.  However it was believed
that retaining the conditional code might prove useful for performance optimizations in the
future in situations with a small number of overrides.  A compiler configuration flag was used
to set a threshold for preferring the dispatch table to the conditional branch code.  This flag
was originally set to 0 i.e. always use the table.  There was an intent to bump this threshold
but there is not a commit message that mentions updating this config and it was set to zero
immediately before this PR.

In July 2016 it was noted that a call to super.foo() from an override of foo() resulted in an
infinite loop; the Chapel compiler failed to include to escape for uses of super that match
the mechanisms found in Java, C#, Swift etc.  The first attempt to address this relied on a
modified version of the conditional branch code.

The recent work on initializers rely on calls to super.init() to invoke parent initializers
and this.init() to invoke peer initializers (c.f. convenience initializers in Swift).  Additionally
initializers are marked "inline".  Details of the conditional-branch code would cause uses
of inlined super.init() to fail in some situations.  When investigating this it was determined that

1. Handling super.foo() was effectively bypassing the dynamic dispatch code altogether.
This is certainly valid if 'super' is being used to invoke the method that is being overridden
by the current method.  There are question marks about what this should mean for other
method calls using 'super' and more research is needed to be certain how other languages
handle this.

2. There was no support for bypassing dynamic dispatch for this.init()

3. There were significant problems with the modern implementation of the conditional branch
code in the context of the current implementation of classes.  We did not attempt to determine
which changes had what effects.  The modern implementation unnecessarily emits far too
much code and 63 tests fail for a single-locale paratest if the threshold is turned up.

The decision was made to

1. Eliminate the portion of the implementation based on conditional branching.
It can be resurrected if future performance analysis supports such an effort.

2. To "hard code" bypassing dynamic dispatch for calls to initializers.

This PR implements these decisions, drops the compiler configuration, updates
the 3 compopts that override the threshold, removes the degenerate future that
was developed to surface the inlining issue,  and adds two passing tests to lock
in the desired behavior for super.init() and this.init().

Compiled with/without CHPL_DEVELOPER on clang/darwin and gcc/linux64.  Passed
start-test on a small portion of release/  Also compiled with CHPL_LLVM=llvm on
gcc/linux64 and passes a small portion of release/.  Passed a single-locale paratest.

Reviewed by @lydia-duncan


compiler/resolution/virtualDispatch.cpp
```diff
@@ -643,116 +660,32 @@ void insertDynamicDispatchCalls() {
...
-        if (isSuperAccess) {
-          // We're in a call on the super type.  We already know the answer to
-          // this if conditional
-          getCid = new SymExpr(gFalse);
-        } else {
-          getCid = new CallExpr(PRIM_TESTCID, call->get(2)->copy(),
-                                type->symbol);
-        }
+    // Exception 2 :- Any call to init()
+    if (call->isNamed("init") == true) continue;
```
lydia-duncan: I think you'll want to be careful.  This might include general init functions (that aren't actually initializers)

lydia-duncan: Waaaaaait, those shouldn't be impacted.  "Shouldn't" being the operable word here.

lydia-duncan:
This looks good!

Issue #6535, opened 2017-06-23
do-while loops and dead block elimination problem


mppf:
The following program should output `in loop` but instead prints nothing:
``` chapel
proc foo() {
  do {
    writeln("in loop");
    break;
  } while true;
}

foo();
```

mppf:
I came across this bug when writing tests for `defer`. Note that the `writeln` is getting removed during resolution, in deadBlockElimination.

PR #6631, opened 2017-07-07
Fix a deadBlockElimination bug for do-while loops

ronawho:
Fix a deadBlockElimination bug that that removed do-while loops with
unconditional breaks/returns.

For something like:

```chapel
proc foo() {
  do {
    writeln("in loop");
    break; // or return;
  } while (true);
}
foo();
```

we mistakenly removed the entire do-while loop. For this case the basic block
construction looks something like:

```
 0:   >  1 3
189546  def const-val tmp:bool

 1:  0 2  >  3
189525  call( fn writeln "in loop" )
189528  goto _breakLabel

 2:   >  3 1
189545  tmp

 3:  2 1  >
189547  def _breakLabel
408343  return( _void )
```

We can see that block 2 (the condition of the do-while loop) is dead because
the break/return prevents it from ever being evaluated. However, the do-while
loops were triggering a while-loop path that removed the entire loop. If the
condition of a while-loop is dead, we can remove the entire while-loop, but
that's not true for do-while loops since they must be executed at least once.
For the above do-while loop we only want to remove block 2, but in reality we
can't do that either since the compiler expects do-while loops to have a
non-null expr for the condition. So now we just don't eliminate anything if we
detect that the block containing a do-while loop condition is dead.

### DIFF
--- a/compiler/optimizations/deadCodeElimination.cpp
+++ b/compiler/optimizations/deadCodeElimination.cpp
@@ -30,6 +30,7 @@
 #include "stlUtil.h"
 #include "stmt.h"
 #include "WhileStmt.h"
+#include "DoWhileStmt.h"
 
 #include <queue>
 #include <set>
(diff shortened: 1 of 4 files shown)


ronawho:
This will allow us to retire `types/records/ferguson/deinit/{break-in-do-while,return-in-do-while}` and resolve https://github.com/chapel-lang/chapel/issues/6535. Running a full paratest now to see if there's any other futures that will be resolved

mppf:
thanks @ronawho !

PR #6633, opened 2017-07-07
Relax filter of which functions will be stack-traced

bradcray:
For some time now, I've been feeling as though there are functions
missing from the CHPL_UNWIND output that shouldn't be -- most notably,
top-level user code like module initialization code and main().  It
turns out that this was due to the filter used in codegen to limit
which functions will be traceable and which not was too strict, based
on the C identifier having the prefix of 'chpl_' which eliminates such
routines.

Here, I apply the same test to the 'name' field such that
anything we munge with 'chpl' to avoid conflicts with C (like 'main')
will still be reported while also opening up reporting to module init
routines so that top-level code will show up in the backtrace.  This
improves the output significantly for me.  It may be that even this
test is a bit silly (e.g., maybe there's a flag that we should be cueing off of
rather than basing it on a string prefix), but this was a simple fix
that will make my use of CHPL_UNWIND much better, so it seemed
worthwhile on its own without getting distracted by trying to find the
optimal solution right now.

Updated existing stacktrace tests to reflect the new output (and the
moved location of halt() which wasn't caught before due to lack of
testing on this directory).

### DIFF
--- a/compiler/codegen/codegen.cpp
+++ b/compiler/codegen/codegen.cpp
@@ -782,7 +782,7 @@ static void genUnwindSymbolTable(){
   if(strcmp(CHPL_UNWIND, "none") != 0){
     // Gets only user symbols
     forv_Vec(FnSymbol, fn, gFnSymbols) {
-      if(strncmp(fn->name, "chpl_", 5)) {
+      if(strncmp(fn->name, "chpl_", 5) || fn->hasFlag(FLAG_MODULE_INIT)) {
         symbols.push_back(fn);
       }
     }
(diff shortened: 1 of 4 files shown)


bradcray:
@mppf: This is a follow-up to my questions today about why top-level stack frames seemed to get dropped in the CHPL_UNWIND view.  It seems to make things better for me, though I haven't lived with it for very long yet.  I've only tested by running it on `test/runtime/panzone` -- let me know if you think additional testing is required.

mppf:
@bradcray - this looks good and I think you've done an appropriate amount of testing for the time being. I think it would be nice if the tests with the location of `halt` had a PREDIFF but presumably we also need some more significant test flexibility should we move towards enabling `CHPL_UNWIND` in nightly testing. For that reason I think it's just fine to merge this as-is.

Thanks!

PR #6634, opened 2017-07-07
Add future for forwarding bug with generic field

benharsh:


benharsh:
@mppf : are you aware of this kind of bug with forwarding?

mppf:
@benharsh no, it is news to me.

PR #6635, opened 2017-07-07
Make chameneos versions all use initializers rather than constructors

bradcray:
This updates the chameneos tests to use initializers rather than
constructors.  The only one that was particularly interesting was
one of Elliot's that needed super.init() for the Population record
due to using a loop to initialize the array, similar to other patterns
that @lydia-duncan and I have discussed in previous initializer
conversion PRs.

I also updated a .notest Hannah version, though it has other problems.


bradcray:
@lydia-duncan: Here are the other chameneos conversions to initializers.

lydia-duncan:
I noted the Elliot version in the issue discussing the insertion point of omitted initialization

PR #6636, opened 2017-07-07
Extend parallel array initialization to POD arrays

ronawho:


ben-albrecht:
For my own reference: This branch is being performance-tested with #6638, which is tracked [here](http://chapel.sourceforge.net/perf/chapcs/pod-parallel-init/?suite=performancetracking).

@ronawho - Do you still intend to pursue this change?

ben-albrecht:
Also, would it be useful to check in the program that ignited this effort, as a performance test?

ronawho:
Yes, but I had completely forgotten that I opened this PR, so I opened a new one (that's ready to be reviewed/merged) over at https://github.com/chapel-lang/chapel/pull/6675

ronawho:
I'm not sure that the small program would be useful as a performance test, but we probably want a correctness test that will print out whether parallel/serial init is used for various types.

PR #6644, opened 2017-07-10
Refactor prim init

noakesmichael:
This is a reasonably simple "close the gap" PR.  No change in behavior.

@benharsh developed a trivial test that exposes a bug when creating arrays of records if the
record has an initializer.  The fix for this bug includes revisions to resolvePrimInit() and resolveMove().

This PR straightens out resolvePrimInit() as a precursor to applying the bug fix.

Submitted as four relatively simple commits to help confirm that this should match
the behavior of the existing code.

Passed a small portion of start_test release/ on all 4 configs.
Passed a single-locale paratest.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -5042,6 +5042,7 @@ Type* resolveTypeAlias(SymExpr* se)
 static bool primInitIsIteratorRecord(Type* type);
 
 static bool primInitIsUnacceptableGeneric(CallExpr* call, Type* type);
+static void primInitHaltForUnacceptableGeneric(CallExpr* call, Type* type);
 
 Expr* resolvePrimInit(CallExpr* call) {
   SymExpr* se     = toSymExpr(call->get(1));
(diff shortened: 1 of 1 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -5069,17 +5071,14 @@ Expr* resolvePrimInit(CallExpr* call) {
...
-      return NULL;
-    }
+  if (primInitIsIteratorRecord(type) == true) {
+    return NULL;
   }
 
   // Generate a more specific USR_FATAL if resolution would fail
-  ensureGenericSafeForDeclarations(call, type);
+  if (primInitIsUnacceptableGeneric(call, type) == true) {
+    INT_FATAL(call, "primInitIsUnacceptbleGeneric should not return true");
```
lydia-duncan: Typo: "Unacceptable" is missing an "a".

lydia-duncan: Ah, I see you fixed that in the next commit

compiler/resolution/functionResolution.cpp
```diff
@@ -5090,45 +5089,64 @@ Expr* resolvePrimInit(CallExpr* call) {
...
+    if (tag != NULL) {
       USR_FATAL(call,
                 "Variables can't be declared using %s generic types like '%s'",
-                (typeCons ? "not-fully-instantiated" : "abstract"),
+                tag,
                 type->symbol->name);
     }
   }
+
+  return false;
```
lydia-duncan: Why make this function always return false or end with a USR_FATAL?  It seems odd to have a return that always gives the same value - one is tempted to ask what is the point of returning at all?

PR #6647, opened 2017-07-10
Fix no-local bug with ref-intent and virtual method calls

benharsh:
A user submitted a test that failed to compile due to a type
mismatch between an actual and a formal. This test had the
ref-intent on a formal in a virtual method. Virtual method
formals are automatically widened in all cases, so the actual
should have also been widened.

Testing:
- [x] full no-local
- [x] full gasnet
- [x] check stencil PRK performance for local field optimization

### DIFF
--- a/compiler/passes/insertWideReferences.cpp
+++ b/compiler/passes/insertWideReferences.cpp
@@ -961,12 +961,20 @@ static void addKnownWides() {
         }
       }
     }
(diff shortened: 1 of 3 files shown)


benharsh:
@LouisJenkinsCS : engin mentioned offline that my initial attempt at this PR didn't solve your problem. I've made a couple of changes since then. Do these latest changes help?

Issue #6628, opened 2017-07-07
Integrate FileCheck into chapel's test suite


coodie:
With recent work on LLVM vectorization and upcoming work with LLVM new tool for testing LLVM related changes will definitely make work easier. 

LLVM uses `FileCheck` tool which is a script that uses regex to assert that some string occured in file.

Here is documentation of FileCheck:
http://llvm.org/docs/CommandGuide/FileCheck.html

It was @mppf idea to use FileCheck for testing code generation and it appears to be very good tool for this task as shown in #6548.

I think this would be very nice to add FileCheck to Chapel's test suite by default, because as of current we need to rely on PREDIFF script and add empty .good file per each test which makes writing tests inconvenient. 

There are however few things to point out and to make sure that we want to use FileCheck anyway and how do we want it to be used.

1. FileCheck is external tool and introduces some dependency into chapel. We need to add FileCheck into LLVM's build, it is easy to add and was already done in #6531 
2. How do we want to use FileCheck in Chapel's test suite. This is a subject of discussion with other developers.

mppf:
Brainstorming some options for opting in to FileCheck based tests:
 * a FILECHECK or test.filecheck file
 * finding some particular magic comment at the top of the .chpl file
 * the lack of a .good file could trigger it (but maybe check for some FileCheck strings?)

coodie:
I think it's good to understand how FileCheck works. The best description is in the document I mentioned in the issue. Basically FileCheck needs two files in order for it to work. One file has commands for FileCheck and the other is the file it checks. Usually file with commands is the one which contains both - the code and the commands (in our case it's .chpl file). Careful reader might notice that .chpl file has both chapel code and FileCheck commands, does this mean that we won't be able to compile .chpl file using chapel compiler in tests? Not really, because FileCheck commands can (and should) occur in commented lines.

Example of test.chpl file with FileCheck commands where we try to check if compiler performed constant folding.

```chpl
for a in A {
  // CHECK: store i32 25
  a = 10+15;
}
```

Possible output (test.ll)

```llvm
mainTest_chpl_3blk_body_:                         ; preds = %mainTest_chpl_3blk_body_, %entry
  %i_chpl.0 = phi i64 [ %4, %entry ], [ %16, %mainTest_chpl_3blk_body_ ]
  %15 = getelementptr inbounds i32, i32* %13, i64 %i_chpl.0
  store i32 25, i32* %15, !tbaa !8
[...]

PR #6648, opened 2017-07-11
Integrating FileCheck with start_test

mppf:
* Update sub_test to support FileCheck and other in-test cfgs
* Updates test/llvm to use these new features

Chapel tests now support a bunch of comments at the top of the file that can adjust the test configuration, for example:

``` chapel
// FileCheck
// COMPOPTS: --no-vectorize
// NOEXEC
```

Currently only FileCheck, COMPOPTS, EXECOPTS, NUMLOCALES, and NOEXEC are supported in this way. It is a (sometimes detected) error to specify both a configuration comment for a given thing and a file-specific one. E.g. don't provide test.compopts and include a `// COMPOPTS:` comment in the header block.

When FileCheck is activated, the test need not have a .good file. Instead, the .chpl file contains comments like `// CHECK:` that indicate portions of the output that should be matched. This kind of verification is particularly useful for LLVM IR dumps but can be used for other sorts of output. However, since FileCheck is an LLVM tool, we currently intend to use FileCheck style tests only in CHPL_LLVM=llvm configurations.

The header block is parsed in a manner to minimize problems with existing tests. It looks for the appropriate keywords in a case-insensitive way. It will generate a warning if one of these is found but there isn't a blank line before the rest of the Chapel program. This is intended to help discover spelling errors.

Resolves issue #6628.

- [ ] full local testing


mppf:
Closing this because we're not sure about specifying-test-properties-in-comments.

PR #6650, opened 2017-07-11
Move refPropagation into its own file

benharsh:
Testing:
- [x] full local


ronawho:
👍 

Do we want to make it it's own "pass" at some point as well?

benharsh:
Isn't it already its own pass? https://github.com/chapel-lang/chapel/blob/master/compiler/main/runpasses.cpp#L139

ronawho:
Oh.. what am I thinking of then. Maybe it's currently controlled by --[no-]copy-propagation instead of having it's own --[no-]ref-propagation?

benharsh:
Yeah, I think it still relies on that same flag.

ronawho:
Ok. Not for this commit, but I'd be in favor of adding a flag for ref propagation

PR #6651, opened 2017-07-11
Test performance of --vectorize for llvm in perf playground

ronawho:


ronawho:
@mppf just FYI. Results should be up tomorrow at http://chapel.sourceforge.net/perf/chapcs/llvm-vectorize

PR #6653, opened 2017-07-11
Remove reachingDefinitionAnalysis and related deadCodeElimination function

benharsh:
Removes reachingDefinitionAnalysis, which seems overly complex for the
amount of cleanup we receive. The logic was also closely tied to some notion
of defs and uses that are ill-defined in today's compiler. After removing
RDA, the generated code was largely the same. Some improvements were
made to deadVariableElimination to recover the majority of the losses. The
main remaining case is that multiple consecutive definitions of variables are
not eliminated. For example:
```chpl
var A : int;
A = 1;
A = 2;
A = 3;
writeln(A);
```

Here, the first two assignments will no longer be removed.

Also updates deadVariableElimination to revisit Symbols with removed uses
deadVariableElimination could be less effective depending on the order
in which Symobls were visted. Consider the following:

A = 1 + 3;
B = A * 5;

If 'A' were visted first, we would consider it a live variable because
of the single use in 'B'. If 'B' is visited first and found to be dead,
the single use of 'A' goes away and 'A' is now dead.

To accomplish this we fill a set with symbols used in statements that
are deleted as a variable is eliminated. If this symbol is not in the
current queue of symbols, we push it onto the queue.

This work exposed some compiler-introduced dead code in parallel and
buildOnStmt that is easily removed.

Single-node performance testing revealed no regressions.


ronawho:
So many 👍

PR #6655, opened 2017-07-11
Fix how we exclude nodes for paratest.chapcs

ronawho:
Previously we would double count exclusive reservations on the same node and
we'd sometimes try to grab a node that was exclusively reserved. This fixes
both of these problems and avoids errors that caused messages along the lines of:

    salloc: error: Unable to allocate resources: Immediate execution impossible
    salloc: error: Unable to allocate resources: Requested nodes are busy

Problem 1 (double counting exclusive reservations):

    SHARED   NODES  NODELIST  REQ_NODES
    no       1                chapcs11
    no       1                chapcs11

we used to just count how many jobs had exclusive reservations, so here we'd
calculate that 2 nodes were reserved exclusively, when in fact only 1 was.

Problem 2 (reserving the wrong nodes):

    SHARED   NODES  NODELIST       REQ_NODES
    no       1                     chapcs11
    yes      1      chapcs10       chapcs10
    yes      12     chapcs[00-11]

we used to see that only one node was being used exclusively, so we'd try to
grab 11 nodes. However, to "load balance" slurm would try to grab chapcs11 since
chapcs10 has 2 jobs on it while others just had (even though chapcs11 had an
exclusive reservation out for it.)

Now we calculate how many nodes are reserved exclusively and maintain a list of
the nodes that have been explicitly requested, and we'll exclude them from our
salloc. e.g. instead of:

    salloc --nodes=11 --immediate

we do:

    salloc --nodes=11 --exclude=chapcs11 --immediate=60

Note that even excluding chapcs11 explicitly it takes slurm a little while to
figure out that it has enough nodes (I haven no idea why) so the immediate=60
is required.

### DIFF
--- a/util/test/paratest.chapcs
+++ b/util/test/paratest.chapcs
@@ -29,24 +29,41 @@ def run_command_wrapper(command):
     return output_lines
 
 
-def get_num_exclusive_nodes():
+def expand_hostnames(hostname):
     """
-    Get the number of nodes reserved exclusively on the chapel partition
+    Expand a possibly shorthand nodelist into all the full hostnames. e.g
+    expands 'chapcs[07-08]' to '[chapcs07, chapcs08]'
     """
-    # Grab the "SHARED,NODES" info for all jobs. For each exclusive (SHARED=no)
-    # job add up the number of nodes it's using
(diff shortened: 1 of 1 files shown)


ronawho:
@ben-albrecht and/or @benharsh mind reviewing?

PR #6656, opened 2017-07-12
Remove '=>' from additional tests

bradcray:
Mike found some more tests that were still using the '=>' operator which
is deprecated and should soon be removed.  This rewrites them to use `ref`
instead.

I completely removed the 'test_illegal_alias_operator' test because it
was designed to make sure that the alias operator couldn't be used as
a standalone statement, but that makes less sense if the operator
doesn't exist anymore (at least, in this context).  Also removed
"4-return-alias-local.chpl" because @mppf pointed out that the
rewrite was redundant with one of the tests in its parent directory.

I also added a .bad file for 'desirable_syntax' which is a future in order
to improve our chances of tracking it as we go.

As far as I can tell, the remaining uses of '=>' are 
* in hpl-hpcc2012.chpl which I'm still working on removing
* in .notest tests that have been kept around for reference and whose value I didn't want to spend the time wrestling with
* associative array literals where `=>` is still valid syntax.


bradcray:
This is primarily for the sake of @noakesmichael's work, but @benharsh and @mppf may want to take a look as their tests are some of the ones I updated.

test/arrays/ferguson/semantic-examples/valgrind-futures/4-return-alias-local.chpl
```diff
@@ -1,7 +1,7 @@
 use samples;
 proc f() {
   var A:[1..4] R;
-  var B => A;
+  ref B = A;
```
mppf: Unless you have some particular reason to keep this around, I'd prefer you simply remove this test. The new version no longer matches the intent of the test, since the array will now return by value. I think it'd be better covered by any future along the lines of issue #5341.

bradcray: I didn't put any significant thought into this edit (other than wondering what "slice" its comments were referring to) and was just looking for tests that needed updating w.r.t. `=>`, doing the obvious rewrite, and then seeing if they still worked.  I hadn't yet gotten around to seeing whether this would still fail under valgrind, but testing today, see that it now passes.

I  don't have a problem with removing it, but also don't see a downside to retaining it (promoting it to its parent directory?) as proof that we don't get memory errors for returning a `ref` alias to a local array.  Your test, though, so your call.

mppf: Let's remove it because the parent directory already has `4-return-local-ref.chpl` which covers this case.

bradcray: That test is a different case than this one, isn't it?  That one returns a local array using `ref` return intent, so its .good file contains an error message.  Whereas this one returns a `ref` to a local array using default intent (which results in a copy out, hopefully with no leaks or valgrind issues when it's working well).

mppf: Sorry, I pasted the wrong test name. `4-return-ref-local.chpl` is the one I was talking about.

PR #6659, opened 2017-07-12
Spec updates for partially generic arguments

mppf:
This commit makes spec changes to go along with issue #6641 and PR #6621.

- [x] full local testing

### DIFF
--- a/spec/Procedures.tex
+++ b/spec/Procedures.tex
@@ -1269,12 +1269,12 @@ is determined by the following steps:
 \item
  If $F1$ is generic over all types and $F2$ is not generic over all
  types, $M_2$ is more specific.
- \item
- If $F1$ is partially concrete generic and $F2$ is generic and not
- partially concrete, $M_1$ is more specific.
 \item
- If $F1$ is generic and not partially concrete  and $F2$ is partially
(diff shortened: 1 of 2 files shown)


spec/Procedures.tex
```diff
@@ -1269,6 +1269,12 @@ \subsection{Determining More Specific Functions}
 \item
  If $F1$ is generic over all types and $F2$ is not generic over all
  types, $M_2$ is more specific.
+ \item
+ If $F1$ is partially concrete generic and $F2$ is generic and not
```
lydia-duncan: I think "partially concrete" and "generic" need to be more separated.  Something about the three-word combination is feeling like not proper English, I suspect because it's two descriptors with an unclear line between them rather than three descriptors or one big one.

I would suggest making it "If $F1$ is partially concrete and generic, and $F2$ is generic and not partially concrete, then $M_1$ is more specific."  And similarly change the next sentence.

ben-albrecht:
Follow-up suggestion: 

For added clarity, we could mention that a `where` clause plays no role in whether or not an argument is partially concrete. This confusion arose for me from the discussion in #6441 of how partially concrete arguments are implemented under the hood using `where` clauses.

PR #6660, opened 2017-07-12
Fix chpl-env-gen.h inclusion guard.

gbtitus:
chpl-env-gen.h is a generated include file that defines C preprocessor
symbols for the configuration settings in effect when the runtime is
built.  It has a standard ifndef-define-endif guard around its body, to
ensure that no matter how many times it is included in a compilation,
the second and subsequent includes have no effect.  Unfortunately that
guard was based on a symbol (_CHPL_ENV_H_) that rightfully belonged to
another file, namely runtime/include/chpl-env.h.  The result was that if
both chpl-env.h and chpl-env-gen.h were included, whichever was first
hid the other one.

To fix this, here we change the inclusion guard symbol in chpl-env-gen.h
to _CHPL_ENV_GEN_H_.


gbtitus:
Hey, @sungeunchoi, wanna do a tiny Chapel runtime review?  :-)

PR #6662, opened 2017-07-12
Convert generic initializers to share code with concrete ones

lydia-duncan:
I realized that when we upgraded the support for concrete classes
and records with initializers, we did not extend that support to
generic classes and records, meaning that many of the verification
checks added in the last four months were only applied to concrete
classes and records.  This change extends the support to include
generic classes and records, and adds tests of generic types with
those error messages.  It also adds some error messages for
assigning to param and type fields in Phase 2.

Details:
----------
InitNormalize.h, .cpp:
Adds three methods to the InitNormalize class, to handle generic
fields with only a type, only an initial value, and both a type and an
initial value.  Changes initializeFieldsBefore and fieldInitFromField
to call those methods when the field is generic (and to give the
expected error message when a field with no init or type expr is
encountered, because we reach that point when no explicit
initialization has been given for them and that is an error).  Changes
fieldInitFromStmt to also call those methods (but allow this to be
applied to fields w/o declared init or type expressions).

initializerRules.cpp:
Modifies preNormalizeFields to skip param and type fields, so that
it can be applied to initializers on generic types without fear.  Changes
preNormalizeNonGenericInit's name to preNormalize and removes the
branch which separated the behavior towards initializers on concrete
and generic types, as well as the old code supporting the generic types.
Add checks that param and type fields are not updated in Phase 2.

normalize.cpp:
Calls preNormalizeFields on generic classes and records rather than
just concrete ones.

Testing:
----------
With Brad's approval, updates .bad file for
test/classes/initializers/generics/genericInitError.chpl now that it gives
both an error for the omitted initialization in Phase 1 and the explicit
initialization in Phase 2.  I'm not going to work on this further in the
immediate future, but will return to it eventually.

Remove the .future and .bad files for the 4 Phase 2 update of param
and type fields cases, and update the .good files for the exact error
message wording.  Resolve the 4 tests of omitted initialization using a
prior param field.

Make test/functions/nspark/generic-varargs1.chpl into a future for a short
period of time - varargs did not work for concrete types with initializers,
but we didn't have a test for it.  I'm going to tackle that in a separate PR.

Adds 43 tests of validation for initializers on generic types.
Covers:
- good and bad uses of "this" during Phase 1
- improper dependence on fields that would be initialized later
- storing enum fields (with explicit naming, use statements, and improper
  accesses to enum constants)
- nested blocks
- param if statements to update param fields (currently encounters an issue
  with generic instantiation naming that I did not expect, both prior to this
  change and after it)
- if statements when missing an else branch to update the param field.
- proper and improper uses of loops in Phase 1
- improper use of a cobegin in Phase 1
- using a local param prior to updating a param field in Phase 1
- improper use of method calls in Phase 1
- duplicate field initialization
- use of nested functions in initializers (some future work here)
- explicitly initializing fields of classes and records with new
- param fields with casts and coercions when the initialization is omitted
  (encounters an issue for the coercion and non-default size int fields,
    both prior to and after this change)
- overriding the default value for a param field's type (generic instantiation
  doesn't label this correctly, both prior to and after this change)
- inserted omitted initialization for a concrete record field when there are
  generic fields present in the type
- in-order initialization of fields
- setting a concrete parent field in the child's initializer
- out of order initialization of fields errors
- properly calling methods in Phase 2
- properly updating the parent field in Phase 2
- properly sending "this" to a function in Phase 2

Passed test/classes/initializers/ with futures, and a full std/ paratest.


lydia-duncan:
Note for reviewer: everything after the first green check is a clean up commit.  But in order should still be useful, albeit with a little out-of-date implementation.

lydia-duncan:
Still waiting on some testing to come back

PR #6666, opened 2017-07-12
Use the Barrier module in SSCA2 instead of defining its own barrier

daviditen:
SSCA2 has its own Barrier class that is very similar to the Barrier record
in the standard modules.  Delete that class and use the one in the modules
instead.

The barrier it was using was stored in a "Task Private Variables" data
structure, but after switching to the Barrier record, it was calling a
NULL virtual method.  Instead, just declare a barrier variable right before
it is used.  Also, pass the barrier to a coforall with 'ref' intent.

Updated the version in test/studies/ssca2/rachaels/ in the same way.

Tested with "start_test test/release/examples/benchmarks/ssca2" and
"start_test test/studies/ssca2/rachaels" with and without gasnet.

Tested the release version for performance on 16 nodes of Cray XC, and
performance appears to be no worse than before, and possibly better.

### DIFF
--- a/test/release/examples/benchmarks/ssca2/SSCA2_Modules/SSCA2_kernels.chpl
+++ b/test/release/examples/benchmarks/ssca2/SSCA2_Modules/SSCA2_kernels.chpl
@@ -15,7 +15,7 @@ module SSCA2_kernels
 //  +==========================================================================+
 
 { 
-  use SSCA2_compilation_config_params, Time;
+  use SSCA2_compilation_config_params, Time, Barrier;
 
   var stopwatch : Timer;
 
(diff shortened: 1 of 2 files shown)


bradcray:
This looks great, thanks David!  It looks like the study version in test/studies/ssca2/rachels/ could benefit from the same transform if you'd be willing to reflect the change there as well.  If not, I can do it in a separate PR.

daviditen:
Thanks @bradcray!

I updated the rachels version now too, and checked that it passes with single locale and with gasnet.

bradcray:
I think this is good to go then, thanks!

PR #6669, opened 2017-07-12
Increase the amount of oversubscription paratest.chapcs does

ronawho:
Bump from nodepara 1 to 2 for gasnet testing. I routinely use nodepara 2 for
gasnet testing and have never encountered any timeouts.

Additionally, if there's no other shared jobs running on the system
(effectively nobody else doing a paratest) bump nodepara by 1.

If chapcs is empty, this should allow a comm=none paratest to finish in ~20
minutes, and a comm=gasnet paratest to finish in ~30 minutes

### DIFF
--- a/util/test/paratest.chapcs
+++ b/util/test/paratest.chapcs
@@ -96,18 +96,16 @@ def get_num_shared_jobs_running():
 
 def get_good_nodepara():
     """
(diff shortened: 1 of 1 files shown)


ronawho:
@benharsh mind reviewing?

Issue #6646, opened 2017-07-10
isClass on extern classes and nil?


vasslitvinov:
Currently our standard function [`isClass()`](http://chapel.cray.com/docs/latest/modules/standard/Types.html#Types.isClass) returns false on an `extern` class or `nil`.

Should it return true?

How about introduce these other functions:

* `isExternClass`
* `isNil`
* `isPointer` - returns true for: ddata, c_ptr, class, extern class, nil

Should it be an error to write `c1 == c2` where c1 and c2 are of **different** extern class types? if c1 is a Chapel class and c2 is an extern class?

Note: these are purely user-facing questions.

An implementation for isClass returning true on extern classes is in #6579. This issue (hopefully) contains the brief discussion there.

----------------

The current proposal (a mix of Michael and Vass suggestions):

* `isClass` - true if Chapel class or nil
* `isExternClass` - true if extern class or nil
* `isPointer` - true if: ddata, c_ptr, class, extern class, nil
* `==` is an error between two different extern classes
* `==` is an error between an extern class and a Chapel class

----------------

#### Pros

* An extern class is a "class", by its name.
* In many respects it behaves like a Chapel class. It can have fields. The default intent is 'const in'. You can `delete` it.
* It is not just a `c_ptr()` to something.

#### Cons

* An extern class comes from outside Chapel.
* It is in fact a pointer.
* It does not inherit from `object`.
* Currently there is no dynamic dispatch on extern classes.

----------------

### Should `isClass` return true on `nil`?

* While nil is somewhat special, it **is** of any class type.

----------------

### Motivating case

We have `proc chpl__delete(arg)` in our internal modules, which implements Chapel's `delete` operator. One thing it does is report an error when `delete` is used illegally. Today it says `if (isRecord(arg)) then compilerError("delete not allowed on records")`. So if `delete` is invoked, say, on a domain (as in BenA's future), `chpl__delete` misses it. The compiler catches it later, at which time caller location is not easily available. So instead I would like `chpl__delete` to report an error whenever `arg` is not a class - because this will catch all cases, not just records. Given that we call `delete` on extern classes as well, we need to check on those too. This led me to the question of isClass on extern classes.

However, this issue is purely about user-facing interface. Error reporting can be fixed simply by adding `isExternClassType`. Which was the first version of #6579. While writing the PR message, I realized I could just fix `isClass` so that I don't need to justify having a separate `isExternClass`. Apparently Michael disagrees that that would be a "fix".

[...]

PR #6674, opened 2017-07-12
Added isExternClassType()

vasslitvinov:
Our standard function [`isClass()`](http://chapel.cray.com/docs/latest/modules/standard/Types.html#Types.isClass) returns false on an `extern` class. However users may want to know if something is an extern class. For example, to determine if it is OK/necessary to apply `delete` to a given expression.

This change:

* introduces `isExternClassType()`, which provides such a test
* upgrades `isClass()` to return true when it is passed `nil`
* adds `=`, `==`, `!=` on extern classes
* assignment/comparison between two different extern classes or between an extern and a non-extern class are not available and will result in "unresolved" error at compile time

See #6646 for discussion and #6579 (not to be merged) for a more extended change where `isClass()` returns true on extern classes. We rejected the latter as discussed in the above issue and PR.

The implementation of `isExternClassType()` follows the lead of `isUnionType()`.

Tested: linux64+verify and gasnet.


vasslitvinov:
@mppf would you review?

mppf:
@vasslitvinov - a belated comment on this PR - I think you need to update the documentation about isClassType - see http://chapel.cray.com/docs/latest/modules/standard/Types.html#Types.isClass - it should mention something like "inherits from object or is nil".

PR #6677, opened 2017-07-13
Tests of initializers using nested classes

bradcray:
These tests show problems when a generic class stores fields of a nested class type and attempts to create new instances of it in its initializer.  These were inspired by test/studies/amr/lib/amr/BergerRigoutsosCLustering.chpl which uses a similar nested class.


bradcray:
Hi @lydia-duncan and @noakesmichael : I ran into the error cases shown by these tests while converting test/studies/amr.  Only then did I notice the mention of nested classes in issue #6530, but thought these tests might still be of interest unless they're redundant with existing ones.

lydia-duncan:
Thanks!  These weren't already covered, I didn't have a version for generic nested classes yet.  These look good to go to me!

bradcray:
Hi @lydia-duncan -- would you let me know if you'd like me to file an issue against these, or whether you have a place you'd like them to be added?

test/classes/initializers/generics/nested2.future
```diff
@@ -1 +1,4 @@
 bug: nested classes cause problems with initializers
+
+This test seems like it should work, but it generates and internal
```
lydia-duncan: "and" -> "an"

lydia-duncan:
I'm content to say listing the issue with nested classes in #6530 sufficient but am willing to open a separate issue if you disagree (I can see the argument that it would be more obvious to the casual user if it had a top-level issue, but also don't want to have to maintain multiple lists).

PR #6678, opened 2017-07-13
Convert studies/amr from constructors to initializers, mostly...

bradcray:
This conversion was more interesting than most and was illuminating to me because it demonstrated to some extent how much more flexible our constructors implementation has been compared to initializers (not to say that means constructors are better...).

Here are some notes:

* test/studies/amr/lib/amr/BergerRigoutsosCLustering.chpl was unable to be converted due to its use of nested classes.  See PR #6677 for some simplified examples that show similar errors.

* AMRHierarchy used to 'delete this' and then set 'this = new AMRHierarchy(...)' -- this converted fairly cleanly to a sibling `this.init()` call.  I still feel a little surprised that this worked so easily without more effort on my part.

* In some cases, fields had to be reordered in order to meet the phase 1 constraints.

* often super.init()s had to be inserted to take care of const initializations in phase 1

* some consts had to be converted to vars due to how they were initialized (e..g, an associative domain that had keys added to it within the body of the initializer).  Lydia points out that a workaround is to use a var temp and then assign it to the const field once it's ready.

* the following pattern came up a few times:

```chapel
const bigD: domain(2);
const smallD: subdomain(bigD);

proc init() {
  smallD = {1..n, 1..N};
  bigD = smallD.expand(1);
}
```

which is interesting in that bigD has to be initialized in phase 1 before smallD, causing a bit of a cycle.  I fixed this by converting the subdomain into a normal domain.  Other options may have been to make them vars and initialize them in phase 2 or to use some temps to set up initial values and then copy into the const fields...?

### DIFF
--- a/test/studies/amr/lib/amr/AMRHierarchy_def.chpl
+++ b/test/studies/amr/lib/amr/AMRHierarchy_def.chpl
@@ -61,7 +61,7 @@ class AMRHierarchy {
 
 
   //|\''''''''''''''''''''|\
-  //| >    constructor    | >
+  //| >    initializer    | >
   //|/....................|/
   proc init (
     x_low:             dimension*real,
@@ -149,11 +149,11 @@ class AMRHierarchy {
 
   }
   // /|''''''''''''''''''''/|
-  //< |    constructor    < |
+  //< |    initializer    < |
   // \|....................\|
 
   //|\''''''''''''''''''''|\
-  //| >     destructor    | >
+  //| >   deinitializer   | >
   //|/....................|/
 
   proc deinit() {
@@ -166,7 +166,7 @@ class AMRHierarchy {
   }
 
   // /|''''''''''''''''''''/|
-  //< |     destructor    < |
+  //< |   deinitializer   < |
   // \|....................\|
 
 
@@ -570,7 +570,7 @@ class PhysicalBoundary
 
 
   //|\''''''''''''''''''''|\
-  //| >    constructor    | >
+  //| >    initializer    | >
   //|/....................|/
   
   proc init ( level: Level ) 
@@ -591,13 +591,13 @@ class PhysicalBoundary
 
   }
   // /|''''''''''''''''''''/|
-  //< |    constructor    < |
+  //< |    initializer    < |
   // \|....................\|
 
 
 
   //|\'''''''''''''''''''|\
-  //| >    destructor    | >
+  //| >  deinitializer   | >
   //|/...................|/
   
   proc deinit () 
@@ -605,7 +605,7 @@ class PhysicalBoundary
     for multidomain in multidomains do delete multidomain;
   }
   // /|'''''''''''''''''''/|
-  //< |    destructor    < |
+  //< |  deinitializer   < |
   // \|...................\|
 
 
@@ -652,11 +652,11 @@ class Flagger {
 
 
 //|\"""""""""""""""""""""""""""""""""""""""""""""|\
-//| >    AMRHierarchy constructor, file-based    | >
+//| >    AMRHierarchy initializer, file-based    | >
 //|/_____________________________________________|/
 //
 //-----------------------------------------------------------------
-// Alternate constructor in which all numerical parameters for the
+// Alternate initializer in which all numerical parameters for the
 // hierarchy are provided using an input file.  This allows those
 // parameters to be changed without recompiling the code.
 //-----------------------------------------------------------------
@@ -708,7 +708,7 @@ proc AMRHierarchy.init (
 
 }
 // /|"""""""""""""""""""""""""""""""""""""""""""""/|
-//< |    AMRHierarchy constructor, file-based    < |
+//< |    AMRHierarchy initializer, file-based    < |
 // \|_____________________________________________\|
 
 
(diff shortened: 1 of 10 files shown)


bradcray:
Attn: @lydia-duncan

test/studies/amr/lib/amr/AMRHierarchy_def.chpl
```diff
@@ -79,10 +79,11 @@ class AMRHierarchy {
     this.x_high            = x_high;
     this.n_coarsest_cells  = n_coarsest_cells;
     this.n_ghost_cells     = n_ghost_cells;
-    this.max_n_levels      = max_n_levels;
     this.ref_ratio         = ref_ratio;
-    this.target_efficiency = target_efficiency;
+    this.max_n_levels      = max_n_levels;
     this.flagger           = flagger;
+    this.target_efficiency = target_efficiency;
+    super.init();
```
lydia-duncan: Nice!

test/studies/amr/lib/amr/AMRHierarchy_def.chpl
```diff
@@ -572,7 +573,7 @@ class PhysicalBoundary
   //| >    constructor    | >
```
lydia-duncan: I'd be tempted to change this and the one after the initializer to say "initializer" instead of "constructor" but that's really nit-picky

test/studies/amr/lib/amr/AMRHierarchy_def.chpl
```diff
@@ -694,16 +695,16 @@ proc AMRHierarchy.AMRHierarchy (
...
+  this.init(
+            x_low,
+            x_high,
+            n_coarsest_cells,
+            n_ghost_cells,
+            max_n_levels,
+            ref_ratio,
+            target_efficiency,
+            flagger,
+            inputIC);
```
lydia-duncan: Nice!

test/studies/amr/lib/amr/AMRHierarchy_def.chpl
lydia-duncan: There's a few other instances in the other files

test/studies/amr/lib/cfboundary/RefiningTransfer.chpl
```diff
@@ -21,8 +21,8 @@ class GridCFGhostRegion {
   //|/...............|/
 
   const grid:             Grid;
-  const coarse_neighbors: domain(Grid);
-  const transfer_regions:  [coarse_neighbors] MultiDomain(dimension,stridable=true);
+  var coarse_neighbors: domain(Grid);
+  var transfer_regions:  [coarse_neighbors] MultiDomain(dimension,stridable=true);
```
lydia-duncan: If you wanted to leave these const, you could make a local variable that gets updated during the computation, and then copy that variable into the const field when you were done building it (though that would obviously have the memory impact creating a copy entails).

test/studies/amr/lib/grid/Grid_def.chpl
```diff
@@ -40,8 +40,8 @@ class Grid {
 
   const dx: dimension*real;
           
+  const cells:          domain(dimension, stridable=true);
   const extended_cells: domain(dimension, stridable=true);
-  const cells:          subdomain(extended_cells);
```
lydia-duncan: Local variables might also allow this to remain in the original order (the assignment to cells in the initializer could be an assignment to a local variable which is used to create extended_cells before it is copied into the cells field).  I think this was the right modification to make, though.

test/studies/amr/lib/level/LevelSolution_def.chpl
```diff
@@ -19,8 +19,9 @@ class LevelSolution {
   //|\''''''''''''''''''''|\
   //| >    constructor    | >
   //|/....................|/
-  proc LevelSolution(level: Level) {
+  proc init(level: Level) {
     this.level  = level;
+    super.init();
```
lydia-duncan: I'm curious about the placement here versus right after the two field updates following it?  This seems like a good candidate for a Phase 1 default body.

test/studies/amr/lib/level/Level_def.chpl
```diff
@@ -90,7 +90,8 @@ class Level {
     // must be multiplied by 2 because a cell is 2 indices wide.
     //---------------------------------------------------------------
 
-    possible_ghost_cells = possible_cells.expand(2*n_ghost_cells);    
+    possible_ghost_cells = possible_cells.expand(2*n_ghost_cells);
+    super.init();
```
lydia-duncan: Also a good candidate for a Phase 1 by default body

lydia-duncan:
You can take or leave my comments before merging.  Thanks for doing these!

test/studies/amr/lib/level/LevelSolution_def.chpl
bradcray: Sounds good.

test/studies/amr/lib/amr/AMRHierarchy_def.chpl
bradcray: Good catch.

test/studies/amr/lib/cfboundary/RefiningTransfer.chpl
bradcray: I think having them be var is not so bad -- that's a good point, though.

PR #6681, opened 2017-07-13
LocaleTree iterators

ben-albrecht:
For logarithmic communication patterns


modules/internal/LocaleTree.chpl
```diff
@@ -51,4 +51,51 @@ module LocaleTree {
...
+
+  iter localeTree(root = here, parent = here) : locale {
+    halt("DONT USE SERIAL ITER");
+    yield parent;
+  }
+
+  /*
+   Logarithmic gather iterator
+   */
+  iter localeTree(param tag:iterKind, root = here, parent = here) : locale where tag == iterKind.standalone {
```
benharsh: I think it would be good to call this something other than ``localeTree``. I would want something called ``localeTree`` to actually do an on-statement onto each locale.

test/multilocale/localeTree/iterators.good
```diff
@@ -0,0 +1,12 @@
...
+Broadcast: 1 -> 3
+Broadcast: 1 -> 4
+Broadcast: 2 -> 5
+Broadcast: 2 -> 6
+Gather: 1 -> 0
+Gather: 2 -> 0
+Gather: 3 -> 1
+Gather: 4 -> 1
+Gather: 5 -> 2
+Gather: 6 -> 2
```
benharsh: I think this test should be written in such a way that we verify that the iteration order is correct.

modules/internal/LocaleTree.chpl
benharsh: I think we could also rewrite this to take an array of Locales and a ``k`` integer representing the max number of children. At that point we might have to use a coforall instead of a cobegin. That way we wouldn't have to rely on chpl_localeTree.

test/multilocale/localeTree/iterators.good
ben-albrecht: How would you recommend doing that? Calling the iterator in a `serial` block?

modules/internal/LocaleTree.chpl
ben-albrecht: >I would want something called localeTree to actually do an on-statement onto each locale.

Fair enough. I wasn't particularly attached to the current naming scheme, and was hoping to get some feedback on that. Have any ideas? Maybe `localeGather` and `localeBroadcast`?

ben-albrecht: > think we could also rewrite this to take an array of Locales and a k integer representing the max number of children. At that point we might have to use a coforall instead of a cobegin. That way we wouldn't have to rely on chpl_localeTree.

Good idea. I'll take a shot at it.

benharsh:
I know I mentioned that we probably wouldn't want a 'gather' method on Replicated, but it seems like a 'broadcast(from : locale)' method could be useful.

ben-albrecht:
So, as an overview, would you agree that we want something like the following?

```chpl

// TODO: Ascii diagrams demonstrating iteration pattern for each iter


/*
Iterates through `locales` via the  `on parent - yield children` pattern, 
where each parent locale has `k` children, ultimately yielding every locale in `locales` 
*/ 
iter localeTree(root: locale, parent = here, locales=Locales, k=2) { }


/*
Iterates through `locales` via the  `on parent - yield children` pattern, 
where each parent locale has `k` children, ultimately yielding every locale in `locales`, 
except for the `root` locale 
*/ 
iter localeTreeTraversal(root = here, parent = here, locales=Locales, k=2) { }


/*
Iterates through `locales` via the  `on child - yield parent` pattern,
where each parent locale has `k` children, ultimately doing an `on-statment` for
every locale in `locales`, except for the `root` locale 
*/ 
iter localeTreeTraversalReversed(parent = here, locales=Locales, k=2) { }


/* Broadcast `from` locale's copy of elements to every other locale */
proc ReplicatedDist.broadcast(from: locale, locales=Locales, k=2)) { }
```

benharsh:
Yeah, something along those lines seems attractive.

ben-albrecht:
@benharsh - I'm going to put this on the back-burner for now - let me know if you're picking it up

I have a WIP branch [here](https://github.com/chapel-lang/chapel/compare/master...ben-albrecht:localeTree), but it is actually working in the wrong direction, since it ends up spawning new tasks from `LOCALE0` (whoops). Therefore, it may not be too helpful as a starting point.

Issue #5900, opened 2017-03-29
Improve Pygments Support


mppf:
This issue is meant to track bugs in pygments / chpl-doc html-ifying in RST containing
```
.. code-block:: chapel
```

mppf:
I noticed that this doesn't get highlighted:
```
.. code-block:: chapel

  proc f() ref {
    return A;
  }
```



and it does if I make the name longer (f -> returnsArrayByRef e.g.)

ben-albrecht:
@mppf  - note that GitHub also uses pygments, so this can be better demonstrated by just code-blocking the snippet:

```chapel
proc f() ref {
  return A;
}
```

ben-albrecht:
Also, here's [Chapel's Pygments lexer](https://bitbucket.org/birkenfeld/pygments-main/src/7941677dc77d4f2bf0bbd6140ade85a9454b8b80/pygments/lexers/chapel.py?at=default&fileviewer=file-view-default) for reference.

mppf:
but... it highlights in github but not in the rst.

ben-albrecht:
I was mistaken, GitHub uses [linguist](https://github.com/github/linguist) for syntax detection which ends up using our [tmBundle](https://github.com/chapel-lang/chapel-tmbundle/).

What about the original code snippet doesn't get highlighted correctly? Trying with the latest release of pygments (2.2.), I get different highlighting on the function name and the ref for `pygmentize -l chapel f.chpl` in:

```chapel
proc f() ref { // ref is not colored, f is colored
  return A;
}
```

vs.

```chapel
proc ff() ref {  // ref is the same color as proc, ff is not colored
  return A;
}
```

mppf:
Needs owned/shared/unmanaged/borrowed/override

mppf:
So, https://bitbucket.org/birkenfeld/pygments-main/pull-requests/735/updating-chapel-highlighter/ is updated to close this issue, but that PR has been open for more than a year.

Asking on the recommended IRC channel, they say "contributing doesn't often go anywhere ... There is a single maintainer and he doesn't seem to be doing much with the project."

mppf:
pygments-main PR 735 has been merged, so I'm closing this issue.

PR #6682, opened 2017-07-13
Replicand method on ReplicatedDist

ben-albrecht:
Adds a `replicand` method to `ReplicatedDist` that allows accessing the elements stored on one locale from another, e.g.

```chpl

use ReplicatedDist;

var D = {1..4} dmapped Replicated();
var A: [D] int;

// Set here.id's local copy to 1
A = 1;

// Broadcast here.id elements to all other locales
for loc in Locales {
  A.replicand(loc.id) = A;
}
```

- [x] Tested Locally


ben-albrecht:
Co-developed with @benharsh 

@bradcray - interested in reviewing this small PR?

benharsh:
I think this addition should be chpldoc'd somewhere.

ben-albrecht:
> I think this addition should be chpldoc'd somewhere.

I assume in the [ReplicatedDist](http://chapel.cray.com/docs/master/modules/dists/ReplicatedDist.html) documentation. Do you know how to add specific methods to this documentation page? It seems like they are all getting stripped away currently... 

(I'll figure it out eventually if you don't have any hints)

modules/dists/ReplicatedDist.chpl
```diff
@@ -424,6 +424,16 @@ class ReplicatedArr : BaseArr {
...
+
+  /* Access another locale's local array representation by locale id */
+  proc replicand(id: here.id.type) ref {
+    return localArrs[id].arrLocalRep;
+  }
+}
+
+pragma "no doc"
+proc _array.replicand(id : here.id.type) ref {
+  return _value.replicand(id);
```
bradcray: Once we can do forwarding as hoped on `_value` fields, we would expect this to go away, is that right?  If so, would you add a comment to that effect indicating that this is a placeholder until such time?  Or, is there any chance that Michael's commit of this morning already gets us to that point?

modules/dists/ReplicatedDist.chpl
```diff
@@ -424,6 +424,16 @@ class ReplicatedArr : BaseArr {
         halt("locale ", here.id, " has no local replicand");
     return localArrs[here.id];
   }
+
+  /* Access another locale's local array representation by locale id */
+  proc replicand(id: here.id.type) ref {
```
bradcray: It seems to me as though this interface would be more natural if it took a locale rather than a locale ID.  Was there a reason it needed to be the latter?

(This isn't something that would normally occur to me, but @mppf was asserting this morning that he believes well-chosen method names should be verb-like rather than nouns in order to make code read like English, which might suggest calling this `getReplicand()` rather than simply `replicand()` if you agree with that philosophy.  That said, I wouldn't have thought of this if we hadn't had the conversation this morning (and I realize that I'm the one who probably suggested we call this `replicand` in the first place)).

Completely unrelated to this PR, I"m noting that `ref` doesn't get highlighted above.  :'(

modules/dists/ReplicatedDist.chpl
benharsh: That commit from this morning does not resolve the ~two-dozen failures on my array-forwarding branch. I would expect this wrapper to go away once forwarding is in.

modules/dists/ReplicatedDist.chpl
ben-albrecht: Good point. I'm happy with `getReplicand(loc: locale) { }`

>Completely unrelated to this PR, I"m noting that ref doesn't get highlighted above. :'(

Might be worth tracking in a new "Improve TmBundle Support" issue (like we have for pygments in #5900)

bradcray:
> I think this addition should be chpldoc'd somewhere.

In addition to the chpldocumentation for ReplicatedDist, I think it would be great to add an indication of this ability to the distributions.chpl primer which has a pretty detailed description of Replicated in it.  As extra credit, I've been thinking that it might be time to break that part of the primer out into its own primer given how different and elaborate it is.

modules/dists/ReplicatedDist.chpl
bradcray: Bummer.  Let's add a comment then so that future readers will know this is intended as a short-term (hopefully) fix and eventually remove it.

bradcray:
Other than those minor comments, this looks like a good change to me!  Glad it was as easy as we'd hoped!

modules/dists/ReplicatedDist.chpl
bradcray: I'll add that I don't know yet that I agree with Michael's philosophy.  For example `student.height(unit=inches)` seems like a reasonable call to support to me since it's querying a property of a student (something we might consider field-like).  In contrast, `student.getHeight(unit=inches)` feels a little unnecessarily verbose.  OTOH, I suppose it could be argued that `student.height(unit=inches)` leaves it unclear whether it's a setter or a getter (though in Chapel, it could be both...).

Anyway, I'm still wrestling with this and just noting the thought, not insisting on the renaming.

mppf: Field names can be nouns. Some methods are sortof accessors or similar to them. In that case I don't have a problem with using a noun. I think that `student.height()` fits into that category.

In the case of `replicand`, I don't have a problem with `A.replicand` because it's arguably accessing a property of A. What is some evidence that it's a "property" rather than an "action"?
 * it's accessing something fundamentally stored in the data structure (A replicated array might be viewed as consisting of replicands...)
 * It returns something that makes sense as the LHS of an assignment statement

In other words, if I was writing complete English sentences in my programming, `A.replicand` would be a *noun phrase* that I would presumably then want to do something else with using a verb. `A.replicand.split()` for an arbitrary example, has reasonable code-to-English properties.

Anyway for these reasons, I prefer it be called `replicand` in this case and not `getReplicand`.

ben-albrecht:
@benharsh && @bradcray - the requested changes have been made. Let me know if you have any additional comments. Ping me if you'd like a snapshot of the updated docs with the new Replicated primer.

modules/dists/ReplicatedDist.chpl
```diff
@@ -70,19 +70,23 @@ referring to the domain or array.
...
 
-    // only the current locale's replicand is accessed
+    // Only the current locale's replicand is accessed
     forall (b,r) in zip(Abase,Arepl) do b = r;
     Abase = Arepl;
 
+    // Access other locales' replicand with the replicand(loc) method
+    Arepl.replicand(Locales[1]) = Arepl;
 
+See the :ref:`primers-distributions` primer for more examples of the Replicated
```
benharsh: Shouldn't this point to the new primer>

test/release/examples/primers/replicated.chpl
```diff
@@ -0,0 +1,169 @@
...
+RA = A;
+writeln("Replicated Array after whole-array assignment:");
+writeReplicands(RA);
+writeln();
+
+//
+// Here, we have each locale update its own copy of `RA` to store its
+// locale ID, which results in a modification to each replicand:
+//
+for loc in Locales do
```
benharsh: I think it would be better to keep this as a coforall-on. If this file is intended to demonstrate the recommended usage of ReplicatedDist, I think a coforall-on is more likely than a serial loop using ``replicand``.

ben-albrecht: I agree w/ switching back to a coforall. However, is on-statement preferred over using `replicand`?

ben-albrecht:
Waiting on final comments from @bradcray

test/distributions/replicated/testReplicand.good
```diff
@@ -0,0 +1,4 @@
+0 0 0 0 0
+1 1 1 1 1
+2 2 2 2 2
+3 3 3 3 3
```
bradcray: Seems like you either need a testReplicand.comm-none.good or a skipif for `CHPL_COMM==none`?  (I tend to prefer the former to make sure that things work on single-locale runs as expected).

bradcray:
Sorry for the delayed review.  Other than the one potential problem I flagged, this looks good to me.  Thanks for taking on the challenge of breaking the replicated distribution documentation out into its own file.

PR #6684, opened 2017-07-13
Better error message upon incorrect 'delete'

vasslitvinov:
When `delete` is invoked on something that's not a class
or an "extern" class, report an error with line number
pointing to the location of the `delete`.

Before this change, this already happened when `delete` was
on a record. Now this happens for all other erroneous cases.

While there:

* Make it an error to invoke `delete` on `nil`.
* Make it a compiler verification check after each pass after resolution that there are no deletes on records.
* Remove the same check in the compiler that used to be done once and reported a user error. Explanation: now the user error is reported by the (updated) module code.
* Remove a compiler check that "free" is not invoked on records. There is nothing special about "free".
* Move the test types/records/compilerErrors/deleteRecordMessage to domains/vass/domain-delete (because that's what it does).

### DIFF
--- a/compiler/main/checks.cpp
+++ b/compiler/main/checks.cpp
@@ -54,6 +54,7 @@ static void checkAggregateTypes(); // Checks that class and record types have
                                    // constructors.
 static void checkResolveRemovedPrims(void); // Checks that certain primitives
                                             // are removed after resolution
+static void checkNoRecordDeletes();  // No 'delete' on records.
 static void checkTaskRemovedPrims(); // Checks that certain primitives are
                                      // removed after task functions are
                                      // created.
@@ -465,6 +466,7 @@ static void check_afterResolution()
   checkReturnTypesHaveRefTypes();
   if (fVerify)
   {
+    checkNoRecordDeletes();
     checkTaskRemovedPrims();
     checkResolveRemovedPrims();
 // Disabled for now because user warnings should not be logged multiple times:
@@ -620,6 +622,16 @@ checkResolveRemovedPrims(void) {
       }
     }
   }
+}
+
+static void checkNoRecordDeletes() {
+  // No need to do for_alive_in_Vec - there shouldn't be any, period.
+  // User errors are to be detected by chpl__delete() in the modules.
+  forv_Vec(CallExpr, call, gCallExprs)
+    if (FnSymbol* fn = call->resolvedFunction())
+      if(fn->hasFlag(FLAG_DESTRUCTOR))
+        if (!isClass(call->get(1)->typeInfo()->getValType()))
+          INT_FATAL(call, "delete not on a class");
 }
 
 static void
--- a/compiler/passes/checkResolved.cpp
+++ b/compiler/passes/checkResolved.cpp
@@ -415,31 +415,6 @@ checkReturnPaths(FnSymbol* fn) {
   }
 }
 
-
-// This test must be run after resolution, since it depends on the knowledge of
-// whether the type of delete's operand is a record type.
-// But it cannot be run after the compiler adds its own record deletes
-// TODO: This violates the "no magic" principle, so the check and associated
-// language in the spec should be considered for removal.
-// In addition, the ability for user code to explicitly call deletes on fields
-// of record type may be necessary for UMM, but this is yet to be demonstrated.
-static void
-checkNoRecordDeletes(CallExpr* call)
-{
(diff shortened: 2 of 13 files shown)


vasslitvinov:

compiler/main/checks.cpp
```diff
@@ -622,6 +624,16 @@ checkResolveRemovedPrims(void) {
   }
 }
 
+static void checkNoRecordDeletes() {
+  // No need to do for_alive_in_Vec - there shouldn't be any, period.
```
ben-albrecht: Being unfamiliar with this code, could you explain this comment a little more? Specifically, what is `for_alive_in_Vec` ? An alternative to `forv_Vec`?

test/domains/vass/domain-delete.chpl
```diff
@@ -0,0 +1,4 @@
+// formerly BenA's types/records/compilerErrors/deleteRecordMessage
```
ben-albrecht: I disagree with moving this test to a user-named path (` test/arrays/vass`). I stand by my philosophy that new tests should be put in descriptive directories, such as `test/types/records/compilerErrors/`. If you feel strongly that it should be described as an `array` or `domain` test rather than a `record` test, then I would at least put it in something like:

- `test/domains/compilerErrors/`
- `test/arrays/compilerErrors/`

compiler/main/checks.cpp
```diff
@@ -622,6 +624,16 @@ checkResolveRemovedPrims(void) {
...
+  // User errors are to be detected by chpl__delete() in the modules.
+  forv_Vec(CallExpr, call, gCallExprs)
+    if (FnSymbol* fn = call->resolvedFunction())
+      if(fn->hasFlag(FLAG_DESTRUCTOR))
+        if (isRecord(call->get(1)->typeInfo()->getValType()))
+          INT_FATAL(call, "delete on records");
```
ben-albrecht: Is there a way we can test this case as well?

compiler/main/checks.cpp
vasslitvinov: for_alive_in_Vec does forv_Vec, except it does not execute the loop body for those nodes that have been "removed" from the AST. Such nodes are typically not pertinent. Here I am saying that there should not be record deletes even on non-pertinent nodes. Makes sense?

compiler/main/checks.cpp
vasslitvinov: Can you clarify what you mean by "this case" ?

test/domains/vass/domain-delete.chpl
vasslitvinov: The test invokes `delete` on a domain, right? I do not see what that has to do with records.

compiler/main/checks.cpp
ben-albrecht: I assumed this is a separate compiler error, and that we could write a test to demonstrate that as well - did I assume incorrectly?

test/domains/vass/domain-delete.chpl
ben-albrecht: arrays and domains are records. Anyway, I don't think it's *too* important. I'm fine with domain- & array- specific paths, but not vass-specific paths 😛

compiler/main/checks.cpp
ben-albrecht: >Makes sense?

Absolutely. Thank you!

compiler/main/checks.cpp
vasslitvinov: Ah good question. Generally, INT_FATAL and INT_ASSERT are triggered by bugs in the compiler. "INT" stands for "internal [to the compiler]". Cf. USR_ stands for "user".

For this one, I am not aware of a bug that would trigger this INT_FATAL. So I do not know how to test it. The same is true for the great majority of INT_FATAL and INT_ASSERT calls in the compiler.

It's like an assert() in C code - it's there to ensure that it is never triggered.

ben-albrecht: Gotcha. I suppose a bug that would trigger this error would be a "bug" in itself. Thanks for clarifying.

test/domains/vass/domain-delete.chpl
vasslitvinov: Arrays and domains are not "records", as far as the user is concerned. This is so according to the current language definition and I do not see that changing any time soon.

I moved the tests in b78ba51

PR #6692, opened 2017-07-13
In LICM, hoist the defPoint of the invariant being hoisted

mlchu:
In LICM, hoist the defPoints of invariants if they are also within the loop.

In our plans to add GPU code generation, we want to automatically create GPU kernels from forall loops. In order to do so, we want to generate code such as: 
```
if (GPU requested) {
   launch GPU kernel;
} else {
   original CPU forall loop code;
}
```
However, to do so, we need to have the def points of invariants also moved up out of the loop, since our GPU loop creation pass is currently after optimizations have completed and we need to reference those variables.

More generally, this fixes a bug with LICM that is currently not surfacing. If you had code like:
```
for i in 1..10 {
  var a: int;
  a = 2;
  b = a + a;
}
```
LICM would transform that into:
```
a = 2;
for i in 1..10 {
  var a: int;
  b = a + a;
}
```
This fix properly changes that to:
```
var a: int;
a = 2;
for i in 1..10 {
  b = a + a;
}
```

### DIFF
--- a/compiler/optimizations/loopInvariantCodeMotion.cpp
+++ b/compiler/optimizations/loopInvariantCodeMotion.cpp
@@ -1,4 +1,5 @@
 /*
+ * Copyright 2017 Advanced Micro Devices, Inc.
  * Copyright 2004-2017 Cray Inc.
  * Other additional copyright holders may be indicated within.
  *
@@ -680,13 +681,13 @@ static bool allOperandsAreLoopInvariant(Expr* expr, std::set<SymExpr*>& loopInva
  * Then a is an alias for b. The current alias analysis is extremely conservative. If there is only one def
  * of a variable check if it is composed of loop invariant operands and operations. 
  */
-static void computeLoopInvariants(std::vector<SymExpr*>& loopInvariants, Loop*
-    loop, symToVecSymExprMap& localDefMap, FnSymbol* fn) {
- 
+static void computeLoopInvariants(std::vector<SymExpr*>& loopInvariants,
+    std::set<Symbol*>& defsInLoop, Loop* loop, symToVecSymExprMap& localDefMap,
+    FnSymbol* fn) {
+
   // collect all of the symExprs, defExprs, and callExprs in the loop
   startTimer(collectSymExprAndDefTimer);
   std::vector<SymExpr*> loopSymExprs;
-  std::set<Symbol*> defsInLoop;
   std::vector<CallExpr*> callsInLoop;
   for_vector(BasicBlock, block, *loop->getBlocks()) {
     for_vector(Expr, expr, block->exprs) {
@@ -1172,7 +1173,8 @@ void loopInvariantCodeMotion(void) {
       //and use the defUseMaps to compute loop invariants 
       startTimer(computeLoopInvariantsTimer);
       std::vector<SymExpr*> loopInvariants;
-      computeLoopInvariants(loopInvariants, curLoop, localDefMap, fn);
+      std::set<Symbol*> defsInLoop;
+      computeLoopInvariants(loopInvariants, defsInLoop, curLoop, localDefMap, fn);
       stopTimer(computeLoopInvariantsTimer);
 
       //For each invariant, only move it if its def, dominates all uses and all exits 
@@ -1180,6 +1182,9 @@ void loopInvariantCodeMotion(void) {
         if(CallExpr* call = toCallExpr(symExpr->parentExpr)) {
           if(defDominatesAllUses(curLoop, symExpr, dominators, localMap, localUseMap)) {
             if(defDominatesAllExits(curLoop, symExpr, dominators, localMap)) {
+              if(defsInLoop.count(symExpr->symbol()) == 1) {
+                curLoop->insertBefore(symExpr->symbol()->defPoint);
+              }
               curLoop->insertBefore(call);
             }
           }   


compiler/optimizations/loopInvariantCodeMotion.cpp
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright 2017 Advanced Micro Devices, Inc.
```
ronawho: @tmacd9 ignorant question on my part -- how do copyrights work with our license/CLA? Do we/cray have sole copyright, or can AMD add theirs?

If they can add theirs is the right place to add it?

ronawho:
Some test failures for:

```
[Error matching program output for arrays/deitz/jacobi-no-inline]
[Error matching program output for memory/figueroa/LeakedMemory6]
[Error matching program output for trivial/deitz/statements/max_range_iteration (compopts: 2)]
[Error matching program output for types/range/elliot/test_min_and_max_range_iteration (compopts: 2, execopts: 1)]
[Error matching program output for types/range/elliot/test_min_and_max_range_iteration (compopts: 2, execopts: 2)]
[Error matching program output for types/range/elliot/test_min_and_max_range_iteration (compopts: 2, execopts: 3)]
[Error matching program output for types/range/elliot/test_min_and_max_range_iteration (compopts: 2, execopts: 4)]
```

Haven't gotten a chance to look into what's going on, but I'll try to get some more info on Monday.

compiler/optimizations/loopInvariantCodeMotion.cpp
mlchu: I'm personally rather apathetic about these types of things, but our legal team originally wanted me to basically put the same entire generic Apache license text you have with our copyright for all the files I modify. I told them that it looked pretty ridiculous to duplicate all that text, so they agreed just adding that single line on top of your text was sufficient. If you guys want something different, I can circle back with AMD legal and see what they think.

ronawho: I also don't care about copyrights/licensing/whatever, so I'll leave that for Tmac :)

ronawho:
Ok, so those regressions look like they come from a pre-existing LICM bug that this PR just happens to expose. I'll continue to dig into them

mlchu:
Interesting. I'm curious as to what the existing bug is, since I thought this change was fairly innocuous.

mlchu:
@ronawho any update on those regressions that this PR exposed?

ronawho:
>Interesting. I'm curious as to what the existing bug is, since I thought this change was fairly innocuous.

The change itself is fairly innocuous, but we only perform LICM on variables whose defPoint in the current loop, and on some globals. For nested loops, this change means more variables are candidates for hoisting out of outer loops. 

>any update on those regressions that this PR exposed?

Yeah, there's at least 2 classes of failures, and I have a fix for one of them. Hoping to dig more into the second case today.

ronawho:
Sorry @mlchu it's been a while since I've done any serious compiler debugging so this is going much slower than I hoped. Might have to get some help from @benharsh

mlchu:
No problem @ronawho. I'm not in a rush. If there's anything I can do to help, let me know.

ronawho:
https://github.com/chapel-lang/chapel/compare/master...ronawho:licm-hoist-defPoint should resolve this issue. I was hoping to get that in today, but unfortunately it's still failing 1 test. 

I'm going to be out for the next 2 weeks, but I'll work on this and some other LICM bugs (https://github.com/chapel-lang/chapel/issues/7003 and https://github.com/chapel-lang/chapel/pull/6739) first thing when i get back.

I may have some time to look at it on the flight too, but I definetly won't be able to merge/test anything until I get back.

mlchu:
@ronawho Any status update on this? I know there's a code freeze happening on Monday. I don't particularly need this to go in before the freeze since I have other stuff to merge before out stuff will work anyway. Just checking if there's anything you need me to do.

ronawho:
@mlchu nope, nothing for you to do. 

I've been working on other licm bug fixes (#7315, #7326) before getting back to this one in hopes that the regressions have the same root cause.

I still plan on getting this into the release (it's a bug-fix anyways, so we have a little more time)

ronawho:
Alright, I finally got to the bottom of this. PR to fix it will be up tomorrow, and then I'll do a full re-test on this PR

---

Some details if you're interested:

I kept trying to debug the licm analysis only to get frustrated because I couldn't find it doing anything wrong for any of the tests that were failing. It turns out that the analysis was fine, we were just hoisting to the wrong loop.

Most of licm operates on basic blocks, which do not directly correlate to the AST (there's no direct way to get from a basic block to the corresponding AST element.) At the end of licm we hoist invariants into the loops "pre-header", and at this point we correlate the basic-block-loop back to the AST-loop with [this code](https://github.com/ROCm-Developer-Tools/chapel/blob/f5b5347149a9769f2a4b9dfcb3c1d11bcdbe920c/compiler/optimizations/loopInvariantCodeMotion.cpp#L116)

We effectively look at the first AST expr in the basic block, and walk up the tree until we find its enclosing loop by doing something like `(header->exprs.at(0)->parentExpr)-insertBefore(expr)`

That `header->exprs.at(0)` is always a defExpr and since they weren't hoisted before `header->exprs.at(0)->parentExpr` pointed to the correct loop. However, now that we are hoisting the defExpr, `header->exprs.at(0)->parentExpr` will now be the loop that the defExpr was hoisted into. So with this PR we were effectively doing something like:


```chpl
for i in 1..10 {
  var x: int;
  x = func();
  for j in 1..10 {
    var y: int; // this is loop 2's `header->exprs.at(0)`
    y = x + 5
  }
}
```

after hoisting the defExpr: 

```chpl
for i in 1..10 {
  var x: int;
  x = func();
  var y: int; // hoisted the defExpr to before `header->exprs.at(0)->parentExpr`, note that this is still loop 2's `header->exprs.at(0)` (header is a basicblock, but parentExpr is the actual AST element)
  for j in 1..10 {
    y = x + 5
  }
}
```

after hoisting the loop invariant: 

```chpl
y = x + 5 // BUG -- still hoisted the invariant to `header->exprs.at(0)->parentExpr`, but parentExpr is now the first for loop
for i in 1..10 {
  var x: int;
  x = func();
  var y: int; // note that this is still loop 2's `header->exprs.at(0)`
  for j in 1..10 {
  }
}
```

ronawho:
Alright with #7396, this passes full standard and gasnet testing.

Sorry for the embarrassingly long amount of time it took to get this merged.

Issue #6685, opened 2017-07-13
GASNET_LD unexpectedly set to X -- with MPI module load


e-kayrakli:
### Summary of Problem
Compilation fails at link time when MPI module is loaded. Sample error message:

```
$CHPL_HOME/runtime/etc/Makefile.comm-gasnet:64: GASNET_LD unexpectedly set to '/usr/mpi/gcc/mvapich2-1.7/bin/mpicc '.  Please file a Chapel bug against this.
$CHPL_HOME/third-party/gasnet/install/linux64-gnu-native/seg-large/nodbg/lib/libgasnet-ibv-par.a(gasnet_bootstrap_mpi-PAR.o): In function `gasneti_bootstrapExchange_mpi':
gasnet_bootstrap_mpi.c:(.text+0x149): undefined reference to `MPI_Allgather'
... bunch of other undefined references on MPI_* functions
```

Some printouts as requested by @bradcray:
```
$ make -f c_Code/Makefile print-GASNET_LD
...
GASNET_LD is /usr/mpi/gcc/mvapich2-1.7/bin/mpicc 
$ make -f c_Code/Makefile print-GASNET_CXX
...
GASNET_CXX is /opt/gcc-4.9.2/bin/g++ 
$ make -f c_Code/Makefile print-GASNET_CC
...
GASNET_CC is /opt/gcc-4.9.2/bin/gcc 
$ make -f c_Code/Makefile print-MPI_CC
...
MPI_CC is 
```

### Steps to Reproduce

**Compile command:**
```
chpl test/release/examples/hello.chpl
```

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.16.0 pre-release (e159b80)
```
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: ibv *
  CHPL_GASNET_SEGMENT: large
CHPL_TASKS: qthreads
CHPL_LAUNCHER: gasnetrun_ibv *
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: gmake
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
[...]

PR #6700, opened 2017-07-14
Fix for issue #6685:  GASNet-with-MPI Makefile bug

e-kayrakli:
Patches a Makefile issue reported in #6685. Fix suggested by the GASNet team.


e-kayrakli:
@bradcray: Do you know of any other Makefile that may have a similar issue? Is this something that can even be tested within the test system?

bradcray:
Hi @e-kayrakli : Thanks for putting this PR together.  This should be the only thing that needs to change.  Unfortunately, our current test configurations wouldn't test this or exercise it (as the longstanding bug proves) -- it's more about testing on a system and in a configuration in which GASNet wants to use mpicc than anything.

I'm ready to merge this unless you have any hesitations.

e-kayrakli:
Thanks for the clarification. You can merge whenever you wish.

Issue #5791, opened 2017-03-22
chpldoc support for module-level deinit


ben-albrecht:
### Summary of Problem
`chpldoc` crashes when a module has a documented module-level `deinit` function, introduced in #5744.

This resulted in the following PR breaking the docs:  #5769 

### Steps to Reproduce

**Source Code:**

```chapel
// deinit.chpl

module Foo {
  /* deinit function */
  proc deinit() {}
}
```
**Compile command:**
`chpldoc deinit.chpl`

**Associated Future Test(s):**
[`test/chpldoc/module/modDeinit.doc.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/chpldoc/module/modDeinit.doc.chpl
          ) #6238 

### Configuration Information

- Output of `chpl --version`: (master as of writing)

lydia-duncan:
Would you or @vasslitvinov add a future to the chpldoc/module test directory to track this?  Assuming the problem isn't specific to having a deinit function in a standard/package module

ben-albrecht:
It looks like the cause of error is that the code-path for `chpldoc` calls `appendClassName` on the function symbol, because it has a destructor flag, resulting in an assertion error in `AST/AstToText.cpp`:

```c

  else if (fn->hasFlag(FLAG_DESTRUCTOR) == true)
  {
    appendClassName(fn);
    mText += ".~";
    appendClassName(fn);
  }
```

bradcray:
@vasslitvinov : Should the module-level deinit() have a destructor?  I'd guess not, but am not sure what the consequences are...

lydia-duncan:
(Note also that new style initializers do not rely on FLAG_CONSTRUCTOR at the moment)

lydia-duncan:
I looked into what would happen if I limited when FLAG_DESTRUCTOR was applied.  Unfortunately it gets added at parse time, when we don't know if we're adding the function as a primary method or a module level function (unless it's a secondary method).  I think what I need to do is look into attaching the flag later in compilation and how that impacts the compiler.

vasslitvinov:
Module deinit() is in no way a destructor or deinitializer, as far as our init/deinit story is concerned.

FLAG_DESTRUCTOR is an implementation helper. If it causes trouble, I recommend removing it from module deinit() functions after parsing, when such functions are distinguished from actual deinitializers.

lydia-duncan:
[...]

PR #6702, opened 2017-07-14
No-doc PrivateDist deinit

ben-albrecht:
Avoid documentation-breaking bug documented in #5791


ben-albrecht:
@vasslitvinov - #5791 is a pretty annoying bug.. Do you know if it's a low-hanging fruit?

@ronawho - Can you refresh my memory on why we don't do docs-builds in Travis testing? Would it be more acceptable now that we don't do OS X builds (which were really slow to begin with)?

ronawho:
@ben-albrecht I'm not sure if there's any reason we don't/cant do a docs-build in travis testing

PR #6706, opened 2017-07-14
Mark const variables using llvm.invariant.start

coodie:
This PR marks const variables using llvm.invariant.start.

I am not 100% sure if I check for constness correctly, I have huge difficulties understanding what chapel compiler is doing, @mppf should be able to see instantly if I add this in correct spots.


mppf:
- [ ] full local --llvm testing
- [ ] full local --llvm --fast

mppf:
Passed full local testing.

PR #6708, opened 2017-07-14
Update array userAPI tests for operations on Block-distributed arrays

bradcray:
As a result of the localization of reindex operations merged yesterday,
blockOpsReindex.chpl started mismatching its .bad file because it was
essentially working.  However, it did not match its .good file because
the .good file wasn't accurate.

In updating the .good, I was reminded that because the early arrays
used for the userAPI tests only had <16 elements, our file full of
values to load into the arrays was too short and these large 10x10
block-distributed arrays would store a number of 0.0s instead.  This
seemed unfortunate (e.g., in some cases, it made it difficult to tell
whether the right value was being fetched or not, since so many
elements shared the same value), so I expanded the file of values
(realValues.txt) to include 100 elements such that each array element
would be unique.

However, expanding the file required updating the .goods of the other
tests that used > 16 elements, so I did that here as well.

Also, enabled these tests for CHPL_COMM=none because they work and
there is some value in testing them there as well ("does Block work as
expected on a single locale?").

Retired the .future that originally started me down this path.


bradcray:
@benharsh: These are tests that I believe you added... are you interested in giving them a review?

PR #6710, opened 2017-07-14
Convert constructor-based future and test to initializers

bradcray:
This updates these tests (one working, one a future) to reflect the new initializers world.


lydia-duncan:
It looks like the description in the .future file still references constructors when perhaps it should reference both constructors and initializers (since the problem was not introduced by initializers)

PR #6722, opened 2017-07-17
Update PR 5296

daviditen:
PR #5296 by @mppf added a new required procedure to the DSI for domains called
`dsiAssignDomain`.  This allows domains to implement assignment in a specialized
way when appropriate.  As of this PR, all rectangular domains implement it the
same way, and all other domains implement it in a single different way.

This PR also moved the definition of common fields in rectangular domains
(`rank`, `idxType`, `stridable`) into a base class.

#5296 never got merged due to proximity to a release and due to expected overlap
with the array views work that was nearing completion at the same time.  This
new PR brings #5296 more up to date with master, resolves the conflicts, and
fixes a handful of multilocale tests that were failing.

Passed a full paratest with CHPL_COMM=gasnet.


modules/internal/ChapelArray.chpl
```diff
@@ -3263,7 +3263,9 @@ module ChapelArray {
       if !a.stridable && b.stridable then
         compilerError("cannot assign from a stridable domain to an unstridable domain without an explicit cast");
 
-    a._instance.dsiAssignDomain(b, lhsPrivate=false);
+    // distributions/robust/arithmetic/resizing/resizeFromOtherLocale.chpl
+    // fails if this uses b._instance instead of b._value
+    a._value.dsiAssignDomain(b, lhsPrivate=false);
```
mppf: @daviditen - I suspect that this reintroduces the bug that @benharsh fixed in PR #5171 but I'm not certain. I'm also not certain how to fix it. I'm worried that lhs._arrs will always be empty in chpl_assignDomainWithGetSetIndices if a was privatized. @benharsh - do you have any suggestions here?

mppf:
@daviditen - sorry for taking so long to review. It only looked at your new commits, since the work before that was by me and already reviewed by others. Thanks!

modules/internal/ChapelArray.chpl
benharsh: David, can you verify that lhs._arrs is indeed always empty if lhs is privatized? It may be that the ``resizeFromOtherLocale`` test isn't as robust as it ought to be.

daviditen: No, it's not always empty.  I added the following to chpl_assignDomainWithGetSetIndices:

```
if _isPrivatized(lhs) {
  writeln(lhs._arrs.size);
}
```

In the resizeFromOtherLocale.chpl test mentioned in this comment, `chpl_assignDomainWithGetSetIndices` is called 3 times with a privatized lhs, and the size of `lhs._arrs` in those 3 calls are 0, 1, 0.

benharsh: Hmm, interesting. I'm going to check out this branch and poke around to see if my mental model of _arrs is still correct.

daviditen:
@benharsh suggested switching both `dsiAssignDomain` calls to use `_instance` instead of `_value`, as well as a `_reprivatize`.  With that, a full GASNet paratest run passed cleanly.

mppf:
This is good to go as far as I'm concerned.

daviditen:
Thanks!

PR #6723, opened 2017-07-17
Switch elegant miniMD to initializers

benharsh:
Also adds a future for a compiler failure when a coforall is within an initializer.

### DIFF
--- a/test/classes/initializers/miscBlocks.chpl
+++ b/test/classes/initializers/miscBlocks.chpl
@@ -1,0 +1,7 @@
+
+class Foo {
+  proc init() {
+    coforall i in 1..here.maxTaskPar do writeln("Hello from task ", i);
+    coforall loc in Locales do on loc do writeln("Hello from ", here);
+  }
+}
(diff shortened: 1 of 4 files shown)


benharsh:
@lydia-duncan , @noakesmichael : I've added a new future in this PR. I suspect it could be resolved quickly, so let me know if you would rather I not add it at all.

lydia-duncan:
That is a known issue, but we don't have a test for it yet, so I think it would be useful to have one, regardless of whether it gets resolved quickly or not.

test/classes/initializers/miscBlocks.chpl
```diff
@@ -0,0 +1,7 @@
+
+class Foo {
+  proc init() {
+    coforall i in 1..here.maxTaskPar do writeln("Hello from task ", i);
+    coforall loc in Locales do on loc do writeln("Hello from ", here);
+  }
+}
```
lydia-duncan: It looks like your .good file is empty, but these writelns should give output, right?

benharsh: I do not invoke the initializer, so these lines won't execute.

lydia-duncan:
The miniMD update looks good to me!

test/classes/initializers/miscBlocks.chpl
lydia-duncan: Ah, right.  Okay, never mind!

PR #6724, opened 2017-07-17
IO.channel.lines (serial-only)

ben-albrecht:
This PR adds `channel.lines()` iterator to the IO module.

We have always intended to support this, but haven't yet due to challenges in supporting a parallel `channel.lines()`. This PR only adds support for a serial `channel.lines()`.

This implementation was derived from `file.lines()`, with all of the arguments inferred from the `channel` object's fields.

This PR also updates `c-ray.chpl` examples to use `channel.lines()`, which were previously using a custom `channel.readlines()` iterator.

Note: `channel.lines()` is implemented as an iterator, whereas `file.lines()` is a procedure that returns an iterable object (`ItemReader`). It would be ideal to update `file.lines()` to also be an iterator in the future.

- [x] Test all flavors of `channels`
  - [x] vanilla `channel`
  - [x] non-vanilla `channel.kind`
  - [x] non-vanilla `channel.locking`

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -3291,8 +3291,7 @@ iter channel.lines() {
     this._set_style(newline_style);
 
     // Iterate over lines
-    var lineReader = new ItemReader(string, this.kind, this.locking, this);
-    for line in lineReader {
+    for line in this.itemReader(string, this.kind) {
       yield line;
     }
 
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -3267,6 +3267,17 @@ inline proc channel.read(ref args ...?k,
   return !error;
 }
 
+/* Iterate over all of the lines in a channel.
+    Only serial iteration is supported.
+
+   :returns: an object which yields strings read from the file
+ */
+proc channel.lines() {
+  var ret = new ItemReader(string, channel.kind, channel.locking, this);
+  return ret;
+}
+
+
 pragma "no doc"
 proc _can_stringify_direct(t) param : bool {
   if (t.type == string ||
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -3267,14 +3267,25 @@ inline proc channel.read(ref args ...?k,
   return !error;
 }
 
+
 /* Iterate over all of the lines in a channel.
     Only serial iteration is supported.
 
    :returns: an object which yields strings read from the file
  */
-proc channel.lines() {
-  var ret = new ItemReader(string, channel.kind, channel.locking, this);
-  return ret;
+iter channel.lines() {
+  var local_style: iostyle;
+
+  // Update local_style to iterate over newlines
+  qio_channel_get_style(this._channel_internal, local_style);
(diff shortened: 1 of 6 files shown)


ben-albrecht:
Still adding tests, but I'll do a preliminary @mppf - ping, in case he has any suggestions on current implementation or other tests that should be added.

ben-albrecht:
In reviewing, @mppf suggested that I implement `channel.lines()` as an iterator instead of returning an `ItemReader()`, which has a `.these()` method. I agree that this is a more Chapel-rific interface, but this would differ from `file.lines()`, which is a method that returns an `ItemReader()`. 

If we are implementing `channel.lines()` as an iterator, should we do the same for `file.lines()`? It may be difficult to give a proper deprecation warning, since the iterator cannot be distinguished from the method. However, the end result should be same for both the method and iterator interface, such that changing the interface would have no impact on user code. A paratest could give us more confidence in this.

vasslitvinov:
While I am OK with this implementation as-is, it modifies the channel's style to set a particular string_end. I don't understand the implications of that. However what if the user knows that they are reading, for example, a DOS-format file and have already adjusted the channel's style appropriately?

modules/standard/IO.chpl
```diff
@@ -3267,6 +3267,29 @@ inline proc channel.read(ref args ...?k,
...
+  // Update local_style to iterate over newlines
+  qio_channel_get_style(this._channel_internal, local_style);
+  local_style.string_format = QIO_STRING_FORMAT_TOEND;
+  local_style.string_end = 0x0a; // '\n'
+  qio_channel_set_style(this._channel_internal, local_style);
+
+  var lineReader = new ItemReader(string, this.kind, this.locking, this);
+  for line in this.itemReader(string, this.kind) {
+    yield line;
+  }
```
mppf: To @vasslitvinov's comment, it should save style and then set it back again. so that the string_end change only applies to within this function. I'd also consider locking the channel for the duration of channel.lines(), to prevent any possible race conditions on the style.

mppf:
Responding to the second half of @vasslitvinov's comment - `channel.lines` should just say it reads lines ending in "\n" and that's all it will ever do.

vasslitvinov:
I am open to be talked into lines() only ever reading \n lines. However, on the surface this does not feel like the right way.

mppf:
@vasslitvinov - we could add a default line separator argument:
``` chapel
iter channel.lines(lineEnding="\n") {
   ...
}
```

Anyway dos line endings are "\r\n" which is more than 1 byte and it'd be a pain to actually support that as the terminator. (But we could, I just don't think it's worth the time). "\r" used to be used by old Mac OS but it doesn't come up as much anymore. With the DOS line endings, you'd just see a "\r" at the end of each line which could be stripped away with `line.strip()`.

I don't want to pretend we actually support DOS line endings. E.g. `writeln` isn't going to emit "\r\n" on DOS. It could, but I don't think it's worth the effort to try to change this.

vasslitvinov:
OK, I have been talked into it. Sounds good.

PR #6725, opened 2017-07-17
Update tests to use initializers instead of constructors

daviditen:
Update tests matching patterns: \*diten\*, \*datetime\*, \*parboil\*, \*types/void\*
to use initializers instead of constructors.

These could all be phase 1:
classes/diten/test_destructor.chpl
classes/diten/test_destructor2.chpl
modules/standard/datetime/testTimezone.chpl
modules/standard/datetime/testDatetimeTZ.chpl

This would benefit from generic fields being phase independent:
studies/parboil/BFS/bfs.chpl

These are not updated yet because they were non-trivial to update:
distributions/diten/testCyclic.chpl
reductions/diten/deleteReductionClass.chpl
modules/standard/datetime/testTimezoneConversions.chpl

### DIFF
--- a/test/classes/diten/test_destructor.chpl
+++ b/test/classes/diten/test_destructor.chpl
@@ -6,7 +6,7 @@ record R {
 // private:
   var c: C;
 // public:
-  proc R(a:int, b:int) {
+  proc init(a:int, b:int) {
     writeln("R");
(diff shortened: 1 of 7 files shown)


daviditen:
Thank you!

PR #6730, opened 2017-07-17
Fix nonscalar types when used for type fields w/ initializers

lydia-duncan:
In the new normalization of initializers on generic types that I added,
some branches used PRIM_INIT_VAR, but that isn't a valid operation on type
variables.  Rearrange the contents of genericFieldInitTypeInference to
accommodate this.

This allows functions/nspark/generic-varargs1.chpl to pass again, so I
removed the .future and .bad file added last week.

Passed test/classes/initializers with futures and a full paratest.


lydia-duncan:
@noakesmichael - this is ready for review when you get the chance

PR #6734, opened 2017-07-18
Convert more tests from constructors to initializers and file some futures

bradcray:
These are tests that presented slightly more challenge in being converted from constructors to initializers.  I filed a pair of futures against some of the challenges that I encountered when they weren't personal probems.

The new futures are:
* genericArgUsesTypeQuery: when this pattern happens: `proc init(x: C(?t), type t2 = t)` we have problems
* initAtomicsViaWrite: initializing atomics using a write causes problems in generic classes

The test changes are as follows:
* boundedBuffer-atomic: in the obvious rewrite, this led to problems captured in genericArgUsesTypeQuery; then I realized that the initializer isn't required since it was just setting the atomics to their default values, so rewrote the fields to share this initializer and explain the change
  - rewrote boundedBuffer-sync to similarly share the field type + initializer
* did the obvious rewrite for precedence.chpl and then realized that it made sense that the .good didn't match any longer since _defaultOf() has no meaning in the initializers world (yay!)
* made a number of updates to learnChapelInYMinutes to update the code and comments from constructors to initializers
  - the biggest here was that the old description leaned on the fact that both compiler-generated and user-created initializers could co-exist, which is no longer the case (by default at least)
  - also rewrote cases of `1..#n` into `1..n` and tried to improve the explanation of the `#` operator.
  - also updated .prediff file to simplify maintenance by inserting linefeeds in the expected serial output.
* in graph.chpl, had to remove some initializations of arrays of lists with nil, which isn't legal and must've squeaked by somehow in the constructors world.  The .good file needed to be updated because we got further in the execution than ever before (though it still doesn't run to completion).


bradcray:
Attn: @lydia-duncan

test/classes/initializers/generics/genericArgUsesTypeQuery.future
```diff
@@ -0,0 +1,5 @@
+bug: using type query field as default argument in generic initializer
+
+Something about this use of a type query field as the default for
+another argument in this generic class's initializer results in
+an internal error.
```
lydia-duncan: Missing a newline

test/functions/hilde/defaultOf/precedence.chpl
```diff
@@ -19,7 +19,7 @@ proc _defaultOf(type t) where t == R
   return result;
 }
 
-proc R.R()
+proc R.init()
```
lydia-duncan: Maybe this test should just be removed?

test/release/examples/primers/learnChapelInYMinutes.chpl
```diff
@@ -850,19 +861,24 @@ class GenericClass {
...
-// Further, we can take advantage of this to allow our copy constructor
-// to copy classes of different types and cast on the fly.
-  proc GenericClass(other : GenericClass(?otherType),
-                     type classType = otherType) {
+// Copy-style initializer.
+// Note: We include a type argument whose default is the type of the first
+// argument.  This lets our initializer copy classes of different
+// types and cast on the fly.
+  proc init(other : GenericClass(?),
+            type classType = other.classType) {
```
lydia-duncan: I'm not entirely convinced that makes this a copy initializer.  @mppf, what do you think?

lydia-duncan:
Thanks for making the futures!

test/release/examples/primers/learnChapelInYMinutes.chpl
mppf: Whatever a.init(a) resolves to should be considered a copy initializer. I think this would qualify.
*however* this is a class, not a record, and for classes the compiler doesn't ever actually add a`copy initailize`.

test/functions/hilde/defaultOf/precedence.chpl
bradcray: I proposed that in my original commit, but decided against it as a means of ensuring that "_defaultOf()" doesn't have any (visible) effect in the initializers world anymore.  I think of this as an example of whitebox testing -- something you'd never write if you just read the (final) language definition (i.e., "What if I had a function called _defaultOf()?" is a silly question in such a world), but something worth testing against knowing that the compiler does have special code for how it treats "defaultOf()" in it today.

I decided this before I ran into a related problem in which I was converting `channel` in the standard modules over to initializers, yet the compiler gave an error relating to _defaultOf() functions even though (as far as I could tell) there were no such functions in the module code.  This emphasized for me that _defaultOf() may still play some role in the initializers world and so we should make sure that those impacts aren't user-visible (I think @noakesmichael mentioned that he ran into a _defaultOf()-related problem in initializers code recently also when trying to fix a BHarsh bug?).

I'd be comfortable removing this test at the point when a `git grep defaultOf` had no hits outside of the `test/` directory.

test/release/examples/primers/learnChapelInYMinutes.chpl
bradcray: That's why I changed the comment to call it a "copy-style initializer".  Are you objecting to that wording?   If so, what do you suggest?

mppf: It doesn't really matter to me; I know what a *copy initializer* is for a record but I don't think we've any agreement on what the term is for the corresponding operation on a class. (and oddly enough that section of CHIP 10 seems to assume it's talking about a record without saying so).
However I think for the purposes of learnChapelInYMinutes I'd advocate just calling it an "initializer for performing a copy", "initializer for cloning", or some other wording that avoids the exact term "copy initializer". I don't really have a problem with "copy-style initializer" but I was responding to Lydia's comment.

test/functions/hilde/defaultOf/precedence.chpl
lydia-duncan: That seems reasonable to me.

We know that _defaultOf is still applied to generic classes and records (I think), and there is at least one more spot that is inserting a call to it for concrete classes and records when we don't want it to do so, but we seem to be moving away from it otherwise

test/release/examples/primers/learnChapelInYMinutes.chpl
lydia-duncan: I'm okay with "copy-style initializer" as the wording.

lydia-duncan: But the options that Michael is proposing make serve to avoid unnecessary confusion

bradcray: I think this whole primer would really benefit from being rewritten by someone (and was thinking that a few people in a deep-dive doing it interactively might be a way to choke that medicine down), so think I'll leave it as-is for now and we can consider whether having this initializer in the primer is even high enough value to be worthwhile (I was trying to limit myself to obvious functionality and style improvements without getting sucked into the trap of reviewing the whole thing in this pass).

PR #6736, opened 2017-07-18
Improve usable functions in CPtr module

mppf:
I found that it was hard to actually use `c_memcpy` and related functions because I was working with `c_void_ptr` and there wasn't a user-visible way to get the size of an object. So I made these changes.

* makes `c_sizeof(type)` user-visible and adds documentation to it
* adds user-visible `isAnyCPtr(type)` to check for `c_ptr` or `c_void_ptr`.
* enables `c_memmove`, `c_memcpy`, and `c_memcmp` to work `c_void_ptr` in addition to any `c_ptr`
 * Adds `c_memset` to fill out the set of C memory functions

Additionally adds tests of these functions.

Passed full local testing.
Discussed with @ronawho and @bradcray .


e-kayrakli:
Hey @mppf, I was wondering if there is any particular reason we don't have a `c_memset`. I had to extern `memset` in the application code for my optimization for PRK-DGEMM. It'd be nice if it was already in this module.

mppf:
@e-kayrakli - I just added a (draft) memset. I did it right away (before adding tests) because I'm asking for API review on this PR & would like feedback on it as well.

e-kayrakli:
Thanks @mppf! This PR will be definitely useful to me.

modules/internal/CPtr.chpl
```diff
@@ -364,8 +376,28 @@ module CPtr {
...
     extern proc memcmp(const s1: c_void_ptr, const s2: c_ptr, n: size_t) : c_int;
     return memcmp(s1, s2, n.safeCast(size_t)).safeCast(int);
   }
+
+  /*
+    Fill bytes of memory with a particular byte value.
+
+    This is a simple wrapper over the C memset() function.
+
+    :arg b: the destination memory area to fill
```
ronawho: Can you use `s` instead of `b` for the arg name

modules/internal/CPtr.chpl
```diff
@@ -364,8 +376,28 @@ module CPtr {
...
+    :arg c: the byte value to use
+    :arg n: the number of bytes of b to fill
+
+    :returns: b
+   */
+  inline proc c_memset(b, c:integral, n: integral)
+  where isAnyCPtr(s1.type) && isAnyCPtr(s2.type) {
```
ronawho: bogus where clause?

modules/internal/CPtr.chpl
```diff
@@ -364,8 +376,28 @@ module CPtr {
...
+    extern proc memset(b: c_void_ptr, c: c_int, n: size_t) : c_void_ptr;
+    memset(b, c:c_int, n.safeCast(size_t));
```
ronawho: safeCast `c` as well?

ronawho:
No objections for the memmove, memcpy, or memcmp changes. Also happy with the memset addition. 

I'm not sure if I like exposing c_sizeof though. Are you intending to be able to call c_sizeof for chapel types, or just on cptr/void_ptr types?

bradcray:
I don't have any objections to most of the proposed changes.  I have a slight reluctance to c_sizeof() that's similar to @ronawho's.  I always worry that such routines have the potential to lead to confusion and/or abuse.  E.g., how do we guide user expectations w.r.t. what `var A: [1..100] real;  c_sizeof(A);` should mean?  Should we limit it to certain types?  (e.g., value types?)

mppf:
@bradcray, @ronawho - let's talk about `c_sizeof`. In particular I'm thinking that we should expose it because many typical C examples of using `memcpy` or the related functions also use `sizeof`. To the point that I'm not sure how to portably write a call to `c_memcpy` without using `c_sizeof` (or writing `extern proc sizeof(type t)` each time, which is hardly better).

I'm open to changing the name of `c_sizeof` if there is a better name you can think of to reduce confusion. But I don't think it's so bad - it is after all prefixed with `c_` to indicate it should only be used for C interoperability.

Regarding @ronawho's concern about `c_sizeof` or `sizeof` giving incorrect sizes for wide pointers, I think that is just a bug to fix. I have also run in to it. I wonder if it would be easier to get correct if it were `c_sizeof(someVariable)`. Anyway, if that's the main issue, we can just put a warning label on `c_sizeof` in the docs for now. (or actually try to fix the bug, which I havn't done yet).

mppf:
> Are you intending to be able to call c_sizeof for chapel types, or just on cptr/void_ptr types?

I think it should work on Chapel types (but buyer beware - it is a gun for shooting one's foot - but useful if you know what you are doing).

bradcray:
I'm open to `c_sizeof()` if it comes with significant warnings in the documentation.  But I also still wonder whether constraining it just to certain types (or variables of certain types) might even permit us to avoid those warnings.  With or without the wide pointer fix for classes, it seems as though it could still lead to questions like "is this returning the size of the pointer/reference, or the object itself?"

mppf:
How is the commit I just pushed for warnings in the documentation for `c_sizeof`? Or would you like to push more for limiting the types it can apply to?

bradcray:
I'm OK with the warning if we go out as-is, but remain curious about the limiting types approach.  E.g., if you didn't have personal need or expectation to call it on classes or arrays, it seems it would make it more bulletproof and less surprising (wouldn't rely on people actually reading documentation).  It's also yet another one of those cases where it'd be easier to expand it when someone had a compelling reason to do so than it might be to limit it later (and potentially deal with confused users/developers between now and then).  But I'm also imagining something as simple as a fairly straightforward where clause -- if it was much more than that, I might not push for it.

mppf:
So, I'd expect that you might reasonably be able to `memcpy` any Chapel type that has a fairly fixed size and layout. That would include all the primitive numeric types but not arbitrary records or classes. It would include all extern records, c pointers, c void pointers, and extern classes too.

To me, that where clause would be complicated enough to explain in the documentation that I'd rather just have the warning.

bradcray:
I'm surprised that Chapel records wouldn't have a sufficiently fixed size/layout for c_sizeof() -- why would that be?  (I understand that we can't accurately predict their size, but since we're asking C to answer the question for us...?)

For your list of types, I might document it as "all numeric Chapel types as well as all extern and C types".  If records were included, I might say "all value types as well as all extern and C types."  Alternatively, we could write the where clause as an exclusion of certain types (syncs, singles, atomics?, classes, arrays) rather than trying to make it list all types that were supported and document similarly ("This routine is not supported for ...")

mppf:
I think it's reasonable for records to be supported for now, but I think the Chapel language shouldn't promise that record fields are contiguous in memory (to enable array-of-structure vs structure-of-array). In fact tuples are not guaranteed to be contiguous now.

I'd be happy to exclude classes, arrays, syncs, singles and atomics; but I think we still might need a warning for records/tuples.

bradcray:
I'd be fine excluding records as well -- had just anticipated them being on the other side of the line.  Hadn't thought about tuples, but no objections to either excluding or warning about them either.

mppf:
@bradcray, @ronawho - I've added a 'where' clause and possible documentation for `c_sizeof` but I still prefer the simplicity of allowing it on everything but having a warning label. The list of things it's not supported on seems, looking back at it, as a random subset of Chapel types. I think having a `where` clause there adds complexity and I'm not sure there's a real benefit to it. We can just say "don't do that" to users who rely on `c_sizeof` in a way that isn't forward compatible. And anyway, it's trivial to make your own `extern proc sizeof(type t)` so we're not really preventing anybody from doing anything here. By adding the where clause we are just ensuring that more people will have to write `extern proc sizeof(type t)` themselves.

While I committed a version with the `where` clause, so you could look at it, I'd like to go back to the version with just a warning and no `where` clause and I'm pretty convinced that is the best choice here.

ronawho:
Yeah, looking at the where clause, I'm in favor of dropping it and just keeping the warning in the comments. Plus Cptr is an internal module after all so it shouldn't be used by users anyway.

ronawho:
My preference would just be to remove the where clause, merge this, and open an issue about supporting a chapel-aware sizeof or something that handles chapel types (including wide types). While I started the sizeof thread, I personally don't think it's important enough to spend too much more time wrestling with it for now

mppf:
I think CPtr can/should be used by users. Features from it (such as `c_ptr`) are documented here: http://chapel.cray.com/docs/latest/technotes/extern.html

For some reason (maybe oversight?) it's not included in the chpldoc'd internal modules, even though most of the methods have chpldoc comments.

ronawho:
I don't view technotes as "user facing" either, but maybe I'm wrong

bradcray:
* I view technotes as user facing.
* If CPtr is meant to be user-facing, I think it should be moved to modules/standard or modules/packages.  The internal modules are meant to be the kinds of things that either aren't user facing or are baked into the language (e.g., arrays, syncs).  I don't think CPtr is the latter.
* I'm OK with the proposed "document rather than where clause" approach.  Thanks for exploring the alternative.

PR #6737, opened 2017-07-18
 Improve well-known functions to work after resolution 

mppf:
In implementing prototype error handling support within tasks, I needed to create some Chapel functions that can be called during code generation. We do this right now by marking such functions as `export` but I couldn't do that because I needed my function to take a wide pointer argument.

So, this PR makes minor adjustments to the compiler in order to keep the list of well-known-functions respected and accurate after function resolution.

In particular it:
 * adds well-known functions to the set of visible functions in visitVisibleFunctions (so that the prune pass does not remove them)
 * adjusts `removeUnusedFunctions()` in function resolution to remove any generic well-known functions from the list, since these cannot be used after resolution (at least in the current compiler).

Passed full local testing.

### DIFF
--- a/compiler/AST/astutil.cpp
+++ b/compiler/AST/astutil.cpp
@@ -34,6 +34,7 @@
 #include "TryStmt.h"
 #include "type.h"
 #include "virtualDispatch.h"
+#include "wellknown.h"
 #include "WhileStmt.h"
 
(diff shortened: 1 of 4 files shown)


mppf:
@noakesmichael - this further improves the well-known function facility to work after function resolution. I'm wondering if you'd be interested in reviewing? thanks.

PR #6739, opened 2017-07-18
Add LICM bug

benharsh:


ronawho:
@benharsh could you add an issue for this and assign it to me?

PR #6740, opened 2017-07-18
Update test to use initializers instead of constructors

daviditen:
Update distributions/diten/Cyclic.chpl to use initializers.  Reordered the
body of the initializer to meet ordering constraints.


daviditen:
Thank you!

Issue #6726, opened 2017-07-17
Ordering of deinitializers: modules vs. records


ben-albrecht:
What should be called first: *module deinitializers* or *record deinitializers*? Today (as of 1.15), module deinitializers are called first. However, this can be problematic in some use-cases.

Consider the following code:

```chpl
var c = new C(1);
var r = new R(c);

class C { var x: int; }

record R {
  var c: C;
  proc init(c) { this.c = c; }
  proc deinit() {
    writeln('Record deinitializer called');
    writeln('I need to read field c, in order to deinitialize');
    writeln(c.x);
  };
}


proc deinit() {
  writeln('Module deinitializer called');
  writeln('Deleting c, even though r still holds a reference to it');
  delete c;
}
```

Today on master, the sequence of events for the above program goes as follows:

- `c` instantiated
- `r` instantiated, with reference to `c`
- module deinitialized, deleting `c`
- `r` deinitialized, accessing field of `c`, which has already been deleted.

The program reports an error with valgrind, since the reference to `c` is being accessed after it's been freed. So this leads to the open question(s): 

- **Should the record deinitializer get called before the module deinitializer?**
  - **If not, how should programs like this ensure `c` is being freed in the correct order?**

bradcray:
Can you give a practical example in which `R` would need to know the value of `C` in order to invoke its deinitializer?

vasslitvinov:
The sequence you are describing is new as of 1.15. The rationale is that global variables should be deinitialized after module deinit(), because the latter may refer to the former.

ben-albrecht:
> Can you give a practical example in which R would need to know the value of C in order to invoke its deinitializer?

[Here](https://github.com/chapel-lang/chapel/pull/6695/files#diff-5a736cac1292deb92c19b3fff952c242R205) is the example that surfaced this issue.

`privateDist` is accessed by `PrivateSpace ` during the record (domain) deinitialization. I haven't dug further into the domain code to see *why* it's accessing the distribution for deinitialization, since it seems like a reasonable thing to do.

vasslitvinov:
@ben-albrecht In this example, seems like privateDist will be deleted automatically at the end of the scope, probably because you used it in dmap(privateDist). So not deleting it in module deinit() would be a solution.

PrivateSpace keeps a pointer to privateDist, so the latter should not be deleted while the former is still around.

ben-albrecht:
>So not deleting it in module deinit() would be a solution.

[...]

PR #6695, opened 2017-07-13
Eliminate memory leak in PrivateDist

ben-albrecht:
There is a minor memory leak when using `PrivateDist`, due to a module-level instance of `Private()` never gets freed. This PR introduces a module-level deinit to ensure the module-level instance of `Private()` is deallocated at the end of programs.

When checking the tests that use `PrivateDist`, I discovered some obvious memory leaks that could be fixed in the tests (user-code), so I went ahead and did this. However, I did not address *all* memory leaks in these tests.

- [x] Confirm all tests that use `PrivateDist` pass locally (`gasnet` and `none`)
    - Double checked that no modules utilize `PrivateDist` either

```
> start_test
    extern/sungeun/externProcInClassShouldBeError.chpl
    modules/vass/SSCA2-with-Private-1.chpl
    modules/vass/SSCA2-with-Private-2.chpl
    multilocale/deitz/needMultiLocales/dist/test_private_space.chpl
    multilocale/deitz/needMultiLocales/test_frag_main.chpl
    optimizations/cache-remote/ferguson/correctness/atomic.chpl
    optimizations/cache-remote/ferguson/correctness/atomic1.chpl
    parallel/taskPar/sungeun/private.chpl

...

[Summary: #Successes = 7 | #Failures = 0 | #Futures = 1 | #Warnings = 0 ]
```

PR #6741, opened 2017-07-18
Partially revert #6695

ben-albrecht:
The quick-fix from #6695 did not account for the domain accessing the domain map during domain deinitialization. 

Solutions to this problem are being discussed in #6726 

This change reverts the changes to `PrivateDist.chpl`, but does not revert the leak-fixes to user-code tests.


mppf:
Looks good, thanks

PR #6742, opened 2017-07-18
Fix string.isAlpha() to return false for non-alphabetical characters

daviditen:
string.isAlpha() was returning true for the characters between uppercase
and lowercase letters in the ascii table.  Changed it to be true only for
characters that are lowercase or uppercase.

Fixed a couple typos in chpldoc comments.

Added, then retired a .future for string.isAlpha().


modules/internal/String.chpl
```diff
@@ -1094,9 +1094,9 @@ module String {
     /*
      Checks if all the characters in the string are printable. Characters are
      defined as being printable if they are not within the range of `0x00-0x1f`
-     or are `0x7f`.
+     and are not `0x7f`.
```
mppf: I'd probably prefer a positive form of the printable range; that is "defined as being printable if they are within the range 0x20 - 0x7e, including the bounds"

test/types/string/diten/stringIsAlpha.chpl
```diff
@@ -0,0 +1,3 @@
+var s = "[\\]^_`";
+
+writeln(s.isAlpha());
```
mppf: Let's make this test also check each 1-character substring of that.

daviditen:
Thanks @mppf.  I'll make your suggested changes.

PR #6756, opened 2017-07-19
Upgrade re2 from 2017-04-01(2b16c27) to 2017-07-01(a810d71)

ronawho:
Upgrade to the latest re2 release. This mainly brings in a patch that restores
limited valgrind support: https://github.com/google/re2/commit/b6e2ddf

Valgrind won't work out of the box, but you can build with -DRE2_ON_VALGRIND.
I'll add a chapel env var to control that next.


ronawho:
@mppf mind reviewing? I've done a full paratest, and confirmed that the tests that fail under valgrind will pass when built with `DRE2_ON_VALGRIND`

I still don't understand the tar-based upgrade instructions, so I just created a diff against a vanilla 2017-04-01 release, and reverse-applied it to the 2017-07-01 release. There was only a minor merge conflict in a comment, so I'm pretty confident I didn't break anything.

third-party/re2/README
```diff
@@ -2,7 +2,7 @@
 re2 README for Chapel
 =====================
 
-This copy of re2 2017-04-01 (2b16c27) is being released with Chapel
+This copy of re2 2017-07-01 (a810d71) is being released with Chapel
```
mppf: I know you're just upgrading RE2 but I'm not sure I'd say anymore that it's "released with Chapel for convenience" because we patch it to add functionality. In other words, you wouldn't get it to work if you replaced third-party/re2/re2-src with a checkout of RE2.
I think it made sense to say this in the past because it was third-party boilerplate and because we thought te RE2 project would eventually accept our patches. But I no longer think they will.

mppf:
@ronawho - thanks, looks good to me. You're welcome to adjust the upgrade instructions to something that you understand, too.
I verified that test/regexp/ferguson passes after a clobber in third-party/re2 on this branch, with Mac OS X.

mppf:
> Valgrind won't work out of the box, but you can build with -DRE2_ON_VALGRIND.
> I'll add a chapel env var to control that next.

Can't you just set CCFLAGS or something like that? Also, is this just when building RE2 or also when running `chpl` ? Just wondering if it's worth having a Chapel variable.

third-party/re2/README
ronawho: Modified to say "based on X version"

ronawho:
>>Valgrind won't work out of the box, but you can build with -DRE2_ON_VALGRIND.
I'll add a chapel env var to control that next.
>Can't you just set CCFLAGS or something like that? Also, is this just when building RE2 or also when running chpl ? Just wondering if it's worth having a (edit: new) Chapel variable.

It just has to be set at re2 build time. I think a chapel var like `CHPL_RE2_VALGRIND_SUPPORT` is appropriate because nobody is going to want to (or remember to) change their ccflags anytime they want re2 support with valgrind. i.e. I imagine @bradcray just wants something he can set in his .bashrc and not have to think about it ever again

third-party/re2/README
ronawho: Ok. I don't think anybody but us will ever read these READMEs anyways, but I changed it again

mppf:
looks good thanks

PR #6759, opened 2017-07-19
Fix support for coforall loops in initializers

lydia-duncan:
This was very straight-forward.  Just add a new block label to InitNormalize's
enum, with a method to return whether we are in a coforall and update the
"describe" method.  Then use the "inCoforall" check when validating this.init(),
super.init(), and field initializations.

Add four new tests of coforalls in initializers.  These are all for initializers on
concrete types, but I have no reason to suspect initializers on generic types
are not appropriately supported.  These tests all failed before the change I
will in this branch.  Remove the .bad and .future files from Ben H's future
(which covers the presence of coforalls in Phase 2).

Passed test/classes/initializers and a full paratest


lydia-duncan:
@benharsh - this should fix the error you were encountering with release/examples/benchmarks/miniMD/explicit/miniMD.chpl

PR #6761, opened 2017-07-19
Further conversions of constructors to initializers, after bug fixes

lydia-duncan:
Updates five tests to use initializers instead of constructors.  Notes and
details below.  Note that 4 of these updates could not be performed
prior to a couple of recent bug fixes.

test/classes/figueroa/RecordConstructor1.chpl:
This test would benefit from Phase 1 as the default.

test/functions/vass/ref-intent-bug-2big.chpl:
Used to fail due to the presence of coforalls in the body.  This test
would benefit from being able to assign const fields in Phase 2.

test/parallel/forall/vass/alist-error.chpl:
Used to fail due to issues with initializing fields via iterator expressions.
Would benefit from Phase 1 as the default.

test/studies/madness/aniruddha/madchap/FTree.chpl:
Used to fail due to the presence of coforalls in the body.  This initializer
would benefit from being able to assign const fields in Phase 2.  One
other helper module for the madness tests remains to update, due to
an issue with copyPropagation that I am hoping will be resolved by
removing the pass ;)

test/studies/matrixMult/stonea/distributed/simpleBlockDist.chpl:
The non-trivial initializer would benefit from Phase 1 as the default.
Used to fail due to an issue with non-generic records and initializing
local variables with a type specified by type arguments.

Passed full standard paratest.

### DIFF
--- a/test/classes/figueroa/RecordConstructor1.chpl
+++ b/test/classes/figueroa/RecordConstructor1.chpl
@@ -1,6 +1,6 @@
 record R {
   var c: int;
-  proc R () {c = 2;}
+  proc init () {c = 2;}
 }
 
 var a: R;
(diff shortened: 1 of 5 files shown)


lydia-duncan:
@vasslitvinov - since a couple of these have your name on them, would you mind reviewing the bunch?

(The ones with your name on them were ones I had grabbed prior to Brad suggesting you look at the "*vass*" ones, and knew they would fail until these issues were resolved)

vasslitvinov:
All sound good to me.

PR #6762, opened 2017-07-19
Use initializers instead of constructors ini explicit miniMD

benharsh:


benharsh:
@lydia-duncan : look OK?

PR #6765, opened 2017-07-19
moving EndCount to task-local storage

mppf:
Adds a task-local storage region, managed by Chapel code. After this PR, this storage region contains the serial state and also end count for fire-and-forget tasks.

 * this task-local storage is stored in the argument bundle structure.
 * Chapel locale model code copies this task-local storage region to the argument bundle for each new task.
 * the runtime task/comm layers no longer track serial_state. The comm layers now need to communicate the Chapel state (which includes serial state).
* adds get/set dynamic end count primitives, which use the  task-local storage
* switch sync, begin, begin on to use task-local-storage to manage end count
* coforall/cobegin end counts work as before

Allows new test end-count-virtual.chpl to pass where it did not before.

(There are other ways we could have tried to get this test to pass; however putting end-count in task local storage is both relatively straightforward and likely to be correct going forward. For example if we tried to detect uses of `begin` in virtual methods and make overridden methods pass an EndCount too, we'd have future problems with separate compilation. Adding serial state to the task-local storage is a clean-up activity while modifying these portions of the runtime.).

Tested Hellos, new tdata-test.chpl, new crazy-serial tests, new end-count-virtual.chpl test with:
- [x] ugni+qthreads on XC
- [x] ugni+fifo on XC
- [x] gasnet+qthreads on XC
- [x] gasnet+fifo on XC
- [x] gasnet UDP+qthreads
- [x] gasnet UDP+fifo
- [x] NUMA locale model, qthreads, gasnet
- [x] fifo comm=none
- [x] qthreads comm=none
- [x] massivethreads comm=none

More testing:
- [x] primers, new tests with GASNet+fifo
- [x] full local testing
- [x] full gasnet testing

Reviewed by @benharsh, @gbtitus, @ronawho - thanks!


mppf:
@ronawho - I think this task-local-end-count PR is ready for your review, thanks. I'd also appreciate any help you want to offer in checking performance impact (whether before or after we merge).

ronawho:
Perf results will be up tomorrow at http://chapel.sourceforge.net/perf/chapcs/dyn-end-count

I'll take a look at the perf results in the morning and start reviewing shortly after that. 

I added Greg and Bharsh as reviewers too. I think Bharsh is more familiar with the module/compiler EndCount code, and Greg is more familiar with the runtime code, and I should be a good middle ground reviewer with decent knowledge of both

ronawho:
Minor regressions for the no-op begin tests

```chpl
for 1..500000 {
  sync {
    for 1..here.maxTaskPar do 
      begin { /* no-op */ } 
  }
}
```

http://chapel.sourceforge.net/perf/chapcs/dyn-end-count/?startdate=2017/07/19&enddate=2017/07/22&suite=taskspawning

But it's a pretty small regression for an already slow pattern, so I'm not worried. Additionally I didn't see any regressions for any real benchmarks.

mppf:
@ronawho - Thanks for checking outh the performance. I can't tell if the apparent speedups (say in forall) are real or not - are these some form of noise or testing artifacts? The reason I'm uncertain is that omp parallel-for also shows an improvement - even more than the others, which seems surprising.

ronawho:
Lcals perf didn't change, so the forall is probably an artifact. I really wish the chapcs machines had more consistent performance between them.. It really makes perf comparisons frustrating

runtime/include/chpl-tasks-prvdata.h
```diff
@@ -23,7 +23,14 @@
 // This header file provides chpl_comm_taskPrvData_t
 #include "chpl-comm-task-decls.h"
 
-// The type for task private data
+// Type for Chapel-managed task private data
+// to be copied to new tasks.
+// Includes 'serial_state'
+typedef struct {
+  unsigned char data[32];
```
gbtitus: If this is too small, the failures may be difficult to diagnose (segfaults, silent wrong answers, etc.).  Is there a guarantee or check anywhere that it is big enough?  The dependence between this and far-away Chapel module code makes me nervous.  Ideally we'd export this runtime type into the module code and then check at startup that the module definition of the same space was no larger.  Either way, there should be a comment in the module code telling developers to make sure this runtime type is big enough, if they change the module definition of the same space.

mppf: @gbtitus - I already added the check here:
https://github.com/chapel-lang/chapel/pull/6765/files#diff-d5ae6549828c154a2a8864af5cf85f21R118

I have a feeling you havn't looked at that file (ChapelTaskData.chpl) yet. Please add a comment on that part if you still feel we need such a comment. I personally think the assert there is enough.

runtime/src/chpl-tasks.c
```diff
@@ -207,3 +207,5 @@ size_t chpl_task_getDefaultCallStackSize(void)
 
   return deflt;
 }
+
```
gbtitus: please remove these new trailing blank lines

runtime/include/chpl-tasks-prvdata.h
gbtitus: Ah!  I'd looked at that file but rather rapidly just for context and missed the assert().

gbtitus:
Runtime looks okay to me.  👍

modules/internal/ChapelBase.chpl
```diff
@@ -956,18 +968,21 @@ module ChapelBase {
...
-    var e = __primitive("get end count");
+  proc _upDynamicEndCount(param countRunningTasks=true) {
+    var e = __primitive("get dynamic end count");
     _upEndCount(e, countRunningTasks);
   }
 
-  proc _downEndCount() {
-    var e = __primitive("get end count");
+  // MPF - should this have
+  //pragma "dont disable remote value forwarding"
```
ronawho: @benharsh any opinion here?

modules/internal/ChapelTaskData.chpl
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+// ChapelTaskData.chpl
+//
+module ChapelTaskData {
+
+  use ChapelStandard;
+
+  // use ChapelError
+  // use CPtr
```
ronawho: leftover comments?

modules/internal/ChapelTaskData.chpl
```diff
@@ -0,0 +1,146 @@
...
+
+  private extern proc sizeof(type t):size_t;
+  private extern proc memcpy(dst:c_ptr, src:c_ptr, n:size_t);
```
ronawho: Use the routines in cptr once that gets merged?

modules/internal/ChapelTaskData.chpl
```diff
@@ -0,0 +1,146 @@
...
+
+  // Chapel task-local data format:
+  // up to 16 bytes of wide pointer for _remoteEndCountType
+  // 1 byte for serial_state
+  private const chpl_offset_endCount = 0:size_t;
+  private const  chpl_offset_serial = sizeof_endcount_ptr();
+  private const     chpl_offset_end = chpl_offset_serial+1;
+
+  // workaround to make sure some things are wide
+  private var neverExecuted = false;
```
ronawho: This feels awful. Can we please find a better way to force wideness. Even a pragma feels better

modules/internal/LocaleModelHelpNUMA.chpl
```diff
@@ -151,16 +155,22 @@ module LocaleModelHelpNUMA {
     //
     const dnode =  chpl_nodeFromLocaleID(loc);
     const dsubloc =  chpl_sublocFromLocaleID(loc);
+    var tls = chpl_task_getChapelData();
+    var isSerial = chpl_task_data_getSerial(tls);
     if dnode == chpl_nodeID {
-      if __primitive("task_get_serial") then
```
ronawho: can/should  we remove the set/get serial primitives now?

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -5421,6 +5426,8 @@ void chpl_comm_execute_on(int locale, c_sublocid_t subloc,
            (int) locale, (int) subloc, (int) fid, arg, arg_size);
 
   if (locale == chpl_nodeID) {
+    assert(0); // locale model code should prevent this...
```
ronawho: halt with error msg instead of assert?

modules/internal/ChapelTaskData.chpl
mppf: I originally did but I'd rather not tie these two together in that way if we don't have to.

modules/internal/ChapelTaskData.chpl
mppf: @benharsh - do you think a pragma/flag would be better than the `on` trick to force wideness? Or, are you OK with a pragma/flag?

runtime/src/comm/ugni/comm-ugni.c
mppf: I was duplicating some code in GASNet which uses asserts for this reason too. We could make them all error messages but there are several of them and I think assert is good enough. Did you see some particular advantage to doing halt-with error?

modules/internal/LocaleModelHelpNUMA.chpl
mppf: It's still used, but anybody using it could just call `chpl_task_getSerial` directly since it's a Chapel function. I'd guess there's some benefit to leaving it a primitive so the compiler can analyze it, but I don't know if we actually do that. Anyway I intend to leave it alone for this PR at least.

modules/internal/ChapelTaskData.chpl
ronawho: Not tie what together. The PRs or the modules?

runtime/src/comm/ugni/comm-ugni.c
ronawho: Just seems better to get an error message for ease of reporting, but I don't feel strongly

modules/internal/ChapelTaskData.chpl
mppf: @ronawho - I was also thinking about adding a pragma to construct a wide pointer with a particular type from a c_void_ptr and a localeID_t. That might be easier to get IWR to work with. But I was also hoping to put it off for this PR.

modules/internal/ChapelTaskData.chpl
mppf: Both

ronawho: I'm not following why you wouldn't want to use the ones from cptr here once they're availble

modules/internal/ChapelTaskData.chpl
benharsh: The easiest solution would likely be to modify IWR's ``addKnownWides`` function to always widen EndCounts. Adding a general-purpose pragma seems like overkill for this PR.

mppf: But there are different types of _EndCount since it's generic. In this case, I only care that one of the overloads is wide.

modules/internal/ChapelBase.chpl
benharsh: Not a well-informed or strong one. I'd be inclined to add the pragma and do some testing as our existing downEndCount has the pragma.

I'm unfamiliar with the historical motivating use-cases for this pragma or what impact is has, so I'm not too sure one way or another.

modules/internal/ChapelTaskData.chpl
mppf: I'm planning to build on PR #6796 to do this. Besides using the new primitive, I'll memcpy just the local pointer and localeID portions. That should clean up some of the part Elliot doesn't like.

benharsh:
Though there are still some wide-pointer issues to be ironed out, the compiler changes in general look good to me.

modules/internal/ChapelTaskData.chpl
mppf: I switched to using the c_memcpy etc now that the other PR is merged.

modules/internal/ChapelBase.chpl
```diff
@@ -943,6 +954,7 @@ module ChapelBase {
     }
   }
 
+  // called for cobegin/coforall
```
ronawho: This overload is only called for bounded coforalls

modules/internal/ChapelBase.chpl
```diff
@@ -914,12 +922,15 @@ module ChapelBase {
...
   pragma "dont disable remote value forwarding"
+  pragma "down end count fn"
   proc _downEndCount(e: _EndCount) {
+    // inform anybody waiting that we're done
     e.i.sub(1, memory_order_release);
   }
 
   // This function is called once by the initiating task.  As above, no
   // on statement needed.
+  // called for implicit sync at end of main
```
ronawho: Pretty sure this one is called for implicit sync, explicit syncs, unbounded coforalls, and cobegins.

modules/internal/ChapelTaskDataHelp.chpl
```diff
@@ -0,0 +1,40 @@
...
+
+  extern type chpl_task_ChapelData_t;
+  extern proc chpl_task_getChapelData(): c_ptr(chpl_task_ChapelData_t);
+  extern proc chpl_task_getBundleChapelData(args: chpl_task_bundle_p):c_ptr(chpl_task_ChapelData_t);
+
+  // This function is called to set up the Chapel-managed portion
+  // of task-local storage in the argument bundle when creating
+  // a new task.
+  proc chpl_task_data_setup(args: chpl_task_bundle_p, tls:c_ptr(chpl_task_ChapelData_t)) {
+    extern proc sizeof(type t):size_t;
```
ronawho: c_sizeof/c_memcpy?

modules/internal/ChapelTaskData.chpl
```diff
@@ -0,0 +1,128 @@
...
+// ChapelTaskData.chpl
+//
+module ChapelTaskData {
+
+  use ChapelStandard;
+
+  // Chapel task-local data format:
+  // up to 16 bytes of wide pointer for _remoteEndCountType
+  // 1 byte for serial_state
+  private const chpl_offset_endCount = 0:size_t;
```
ronawho: Spacing here looks a little weird. I think either align them or (my personal preference) just drop to 1 space after `const`

modules/internal/ChapelTaskData.chpl
```diff
@@ -0,0 +1,128 @@
...
+  private const  chpl_offset_serial = sizeof_endcount_ptr();
+  private const     chpl_offset_end = chpl_offset_serial+1;
+
+  // What is the size of a wide _EndCount pointer?
+  private
+  proc sizeof_endcount_ptr() {
+    return c_sizeof(chpl_localeID_t) + c_sizeof(c_void_ptr);
```
ronawho: Is this still right for --no-local?

ronawho:
👍 looks good to me

mppf:
I saw these failures in full gasnet testing:

[Error matching program output for parallel/taskPar/sungeun/taskCount (compopts: 1)]
[Error matching program output for parallel/taskPar/sungeun/taskCount (compopts: 2)]
[Error matching program output for performance/sungeun/multilocale/syncsingle]

These are all comm count differences. Should we try to fix these before merging?

modules/internal/ChapelTaskData.chpl
mppf: yes we just always store the locale ID even on --local

ronawho:
Assuming the comm count regressions are for sync/begin only I'm ok living with them

mppf:
Yes all of these have to do with `begin`

mppf:
The comm count regressions turned out to be an existing bug in the compiler revealed by the new code. See PR #6822.

mppf:
I've rebased now that  PR #6822 is in & confirmed that it fixes the comm counts failures. I think this is ready to merge.

PR #6772, opened 2017-07-20
Two configure/make install fixes

### LINKED ISSUE #6352: make check doesn't work with ./configure
Running the following sequence, without any Chapel environment variables set, results in a failed `make check`:

``` bash
./configure
make
make check
```

The `./configure` step prints out this:
```
  If the configuration above is appropriate, build Chapel with:

make
make check

  and possibly install it with

make install
```

But `make check` fails in that order like so:
```
$ make check
[Info] Running minimal test script: $CHPL_HOME/util/test/checkChplInstall
[Fail] chpl not found. Make sure it available in the current PATH.
Makefile:179: recipe for target 'check' failed
make: *** [check] Error 1
```

We need to do either or both of:

a) Update `./configure` to suggest running `make check` after `make install` and point out that chpl must be in your path at that point
b) Update `make check` to auto-detect the location of the built chpl to run

### LINKED ISSUE #6749: gasnet configure + make install results in bad path
When building a GASNet-enabled release of Chapel via `configure --prefix=$INSTALL_PREFIX`, `make`, `make install` I find that I get a bad path when trying to invoke the resulting compiler, as follows:

```
$INSTALL_PREFIX/share/chapel/1.16/make/Makefile.base:167: $INSTALL_PREFIX/share/chapel/1.16/third-party/amudprun/Makefile.include
```

It appears that the path where `Makefile.include` is installed is actually under `$INSTALL_PREFIX/lib/` rather than `share/`

``chpl --version``: ``chpl Version 1.16.0 pre-release (913c97e)``

mppf:
* Fix configure output per issue #6352
* Fix gasnet + make install issue #6749
  - updated some launcher build paths to use Makefile variables added in #6036
  - forwarded these same new Makefile variables to the secondary `make` invocation for the launcher

Closes #6352.
Closes #6749.

Tested make check with GASNet+fifo and GASNet+qthreads and ugni+fifo.


mppf:
@ben-albrecht or @bradcray - could you review this PR fixing some make install issues? thanks

ben-albrecht:
It appears that this PR broke the following workflow:

```bash
>chpl program.chpl --savec=gen
>make -f gen/Makefile
```

bradcray:
PR #6960 attempts to fix the newly broken behavior @ben-albrecht mentioned above.

PR #6773, opened 2017-07-20
Fix record type alias bug with initializers

lydia-duncan:
Prior to this change, the conversion of constructors to initializers for
code that utilized type aliases on records was failing with the output:

`error: illegal use of function that does not return a value: 'init'`

This was because the AST for calls of the form `var a = new alias(args);`
was being maladjusted - we would end up with:

`call(PRIM_MOVE, a, call('init', tmp: Foo, args))`

instead of:

```
call('init', tmp: Foo, args);
call(PRIM_MOVE, a, tmp);
```

This changes modifies the handling of PRIM_NEW during resolution
to insert the correct AST.

It also commits the conversion of these two tests to use initializers,
which was what brought our attention to this issue.  All initializers
involved would benefit from Phase 1 as the default.  I also modified
one of them further to remove record inheritance, which was
still present.

Passed test/classes/initializers with futures.  Passed a full std/
paratest run.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -4985,7 +4985,13 @@ static void resolveNew(CallExpr* call) {
             if (isBlockStmt(call->parentExpr) == true) {
               call->insertBefore(def);
             } else {
-              call->parentExpr->insertBefore(def);
+              Expr* parent = call->parentExpr;
+              parent->insertBefore(def);
+
+              if (isClass(at) == false) {
+                call->replace(new SymExpr(newTmp));
+                parent->insertBefore(call);
+              }
             }
 
             if (isClass(at) == true) {
--- a/test/types/records/hilde/callConstructorThruTypeFn.chpl
+++ b/test/types/records/hilde/callConstructorThruTypeFn.chpl
@@ -8,9 +8,10 @@ record R {
   proc init(i : int) { _i = i; }
 }
 
-record S : R {
+record S {
+  var _i : int;
   var _r : real;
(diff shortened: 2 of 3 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -4985,7 +4985,13 @@ static void resolveNew(CallExpr* call) {
             if (isBlockStmt(call->parentExpr) == true) {
               call->insertBefore(def);
             } else {
-              call->parentExpr->insertBefore(def);
+              Expr* parent = call->parentExpr;
```
noakesmichael: This looks a bit odd to me but I am sure that there's something I don't understand.  The concern
is that the new code in the else branch has the same "flavor" as the code in the conditional expression
that immediately follows.  Of course the expression you modified is conditional on whether the
parent is a blockStmt or not while the next test is only based on whether at is a class or not.

Still it feels as if there might be some interesting business logic here that is being obscured.
I am going to spend a few minutes contemplating resolveNew() a bit more.  Meanwhile I bet
you understand this better than I do.

lydia-duncan: I believe both "in a block" and "on a record" need to be true for the transformation I'm inserting to be necessary - if the inserted init call started as a top-level statement, then we weren't expecting it to return anything in the first place.

It might be worthwhile to split the handling of non-generic classes and non-generic records out from each other, to allow the control flow in this area to be more straight-forward.  I'm happy to do that in this PR if you agree

lydia-duncan: (off list, Mike said to merge for now, but we might want to take a look at this function again soon)

PR #6776, opened 2017-07-20
Fix Several ArrayViewReindex Leaks

benharsh:
1) For whatever reason, using _getDistribution in the constructor
resulted in a new instance of 'thisDomClass.dist'. Instead we can just
pass the class straight in.

2) Free 'updom' in dsiSetIndices if it is not nil. Before we were
writing over 'updom' and never freeing it.

3) Delete 'downdomInst' if we point it to a class that ArrayViewReindex
allocated. This state is tracked in the new 'ownsDownDomInst' field.

Testing:
- [x] full local
- [x] full gasnet
- [x] dist-replicated, dist-cyclic
- [x] sporadic valgrind testing

### DIFF
--- a/modules/internal/ArrayViewReindex.chpl
+++ b/modules/internal/ArrayViewReindex.chpl
@@ -82,13 +82,16 @@ module ArrayViewReindex {
 
     var dist;
 
+    var ownsDownDomInst = false;
+
     //
     // TODO: If we put this expression into the variable declaration
     // above, we get a memory leak.  File a future against this?
     //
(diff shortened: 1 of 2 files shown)


benharsh:
@bradcray : I believe this PR will squash the leaks introduced in #5757 .

bradcray:
Thanks for jumping on this!  I've given this at least a cursory look and don't have objections, but may have missed something and won't likely have time to look more carefully this week.  I might suggest doing a valgrind run on a few key tests that use reindexing to make sure cleaning up better didn't introduce any use-after-frees.  Thanks again.

PR #6779, opened 2017-07-20
Remove handling of references from copyPropagation

benharsh:
As a part of our goal to simplify and/or rework copyPropagation, this
commit removes the parts of the copyPropagation pass that attempt to
propagate references. Our alias analysis capabilities are poor and the
benefit of this code is tiny compared to the effort required to maintain
it.

Handling of addr-of was moved into the newly-renamed 'needsKilling'
helper so that available pairs are correctly culled.

I have also added some very naive/crude code to determine if a
symbol has a live reference to it. Once a reference is taken to a
symbol, no more copy pairs will be created for that symbol.

After this change, miniMD only suffers a 2% increase in AST size.

While here I made some minor style cleanups:
- used a printf macro instead of checking against DEBUG_CP repeatedly
- cache the result of 'call->get(1) == se'

Testing:
- [x] full local
- [x] full no-local
- [x] full gasnet
- [x] perf testing


benharsh:
Want to review, @ronawho ?

ronawho:
FYI there are 3 current futures (that I know of) that fail because of copy prop. This seems to resolve `ifi.chpl`, but not the other 2:

```
expressions/ferguson/if-expr/ifi.chpl
functions/deitz/nested/test_nested_var_iterator3.chpl 
optimizations/copyPropagation/cp-problem1.chpl
```

compiler/optimizations/copyPropagation.cpp
```diff
@@ -323,43 +216,27 @@ static bool isDef(SymExpr* se)
   }
   else
   {
+    const bool isFirst = call->get(1) == se;
```
ronawho: I don't find `isFirst` to be a particularly helpful name

ronawho:
Changes look sane to me, though I wouldn't mind a different name for `isFirst`. Don't have a great suggestion off the top of my though.

I'd also be curious what it would take to resolve those other two futures since they both look at least somewhat related to references in copy-prop, but up to you if you want to do that in this PR or a different one.

ronawho:
👍

benharsh:
@mppf : this PR may be of interest to you as well.

benharsh:
I'll be re-testing this PR since there have been some large changes on master since the last round of testing (EndCount, ForallStmt).

PR #6780, opened 2017-07-20
Bad new

noakesmichael:
I reviewed a small change to resolveNew() this morning and noted it was getting a little
wooly.

When I took a second look it seemed to me that some simple user-level errors e.g.
var x = new int;

were simply being ignored. I confirmed that this statement was largely ignored by
resolve() and ultimately generated an INT_FATAL much deeper in the pipeline.

In this PR I 

1. catch non-aggregate types and generate more pleasing error messages,

2. add a few tests to help lock in the error messages for inappropriate uses of new,

3. take the opportunity to break the monolithic implementation of resolveNew() 
in to a few helpers.  These changes do not affect behavior.

Ran start_test on a small portion of release/ for all 4 configs.
Passed a single-locale paratest with futures

### DIFF
--- a/test/expressions/new-expr/new-on-domain.good
+++ b/test/expressions/new-expr/new-on-domain.good
@@ -1,0 +1,2 @@
+new-on-domain.chpl:1: In function 'main':
+new-on-domain.chpl:2: error: invalid use of 'new' on chpl__buildDomainRuntimeType
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -4952,80 +4952,25 @@ static SymExpr* resolveNewTypeExpr(CallExpr* call);
 
 static bool     resolveNewHasInitializer(AggregateType* at);
 
+static void     resolveNewHandleInitializer(CallExpr*      call,
+                                            AggregateType* at,
+                                            SymExpr*       typeExpr);
+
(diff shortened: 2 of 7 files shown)


lydia-duncan:
You may want to English the PR message :) (that's unusual for you!)

test/expressions/new-expr/new-on-domain.good
```diff
@@ -0,0 +1,2 @@
+new-on-domain.chpl:1: In function 'main':
+new-on-domain.chpl:2: error: invalid use of 'new' on chpl__buildDomainRuntimeType
```
lydia-duncan: Is there a prettify function you could use on this error message so that we don't bleed internal compiler implementation details?

noakesmichael: In fairness to me, this is an existing message from a path I didn't change.  I added the
test to the 2 I had a hand in to get this path on the table.  I am open to ideas on how
to make it more elegant but I'd be tempted to do that in a follow up PR

lydia-duncan: Ah okay.  I'd be tempted to open an issue on it with the label "easy" and see if anyone running through the compiler feels so moved.

lydia-duncan:
This looks good, thanks!

PR #6781, opened 2017-07-20
Native Chapel transpose in LinearAlgebra

ben-albrecht:
Using BLAS for dense matrix transpose was pretty bad for performance, due to the temp-array allocations required. This PR adds a native transpose, and a performance benchmark to compare it to the BLAS transpose.

Note that the LinearAlgebra tests only run on Crays, so this data will not be generated on the linux testing machines.

### Local performance testing

| m | 10 | 1000 | 10000 |
|---|----|------|----|
native | `2.591e-06` | `0.0051657` | `1.22584`|
BLAS | `3.276e-06` |  `0.0330011`| `37.4587` |
---------

Note: for sufficiently large matrices, the following form was shown to be faster:

```chpl
 [i in Dom.dim(1)] C[.., i] = A[i, ..];
```

I might add this to the elegant vs. performant suite (or as another variation to the transpose perf tests)

Also updated development Makefiles to assume MKL instead of netlib BLAS

**Correctness Testing**
  - [x] Cray


modules/packages/LinearAlgebra.chpl
```diff
@@ -312,17 +312,16 @@ proc _array.T where this.domain.rank == 1 { return transpose(this); }
 
 */
 proc transpose(A: [?Dom] ?eltType) where Dom.rank == 2 {
```
benharsh: If we never intend for users to call this with a distributed array, it would be good to add a compilerError if ``Dom`` is distributed. This could just be a TODO for now.

PR #6783, opened 2017-07-21
Add FLAG_LAST_RESORT, use it

mppf:
FLAG_COMPILER_GENERATED previously included multiple overloaded meanings.  The most confusing of these overloaded meanings was that a function marked with FLAG_COMPILER_GENERATED is only selected during resolution if no candidate functions without that flag are found. This was sometimes used to create a last-resort function for errors (such as with deprecation errors) even though the function involved is not compiler generated (or even internal to the implementation, as happened in Random.chpl).

This PR creates a new flag, FLAG_LAST_RESORT, to request specifically this behavior in function resolution.

To implement the change, I looked at every use of FLAG_COMPILER_GENERATED in the compiler and modules.

When FLAG_COMPILER_GENERATED was added to a function, I generally also added FLAG_LAST_RESORT. In some cases I was confident that the intent was *only* to make the function last-resort in resolution (as with cases where the function exists to produce an error message). In those cases I removed the FLAG_COMPILER_GENERATED from the function.

When FLAG_COMPILER_GENERATED was checked, I verified that its use made sense - e.g. that it is controlling error output or had to do with identifying POD types. I updated the comment in flags_list.h for FLAG_COMPILER_GENERATED to reflect my understanding of what impact this flag has.

Of course, I updated the resolution portion that preferred functions without FLAG_COMPILER_GENERATED to instead check for FLAG_LAST_RESORT.

Passed full local testing. 44358cb722d58bd0a7fd3b7656bdfd588efaef03
- [x] final full local testing

### DIFF
--- a/compiler/include/flags_list.h
+++ b/compiler/include/flags_list.h
@@ -69,14 +69,14 @@ symbolFlag( FLAG_COERCE_TEMP , npr, "coerce temp" , "a temporary that was stores
 symbolFlag( FLAG_CODEGENNED , npr, "codegenned" , "code has been generated for this type" )
 symbolFlag( FLAG_COFORALL_INDEX_VAR , npr, "coforall index var" , ncm )
 symbolFlag( FLAG_COMMAND_LINE_SETTING , ypr, "command line setting" , ncm )
-// The compiler-generated flag is already overloaded in three ways.  We may
-// want to split it if this becomes cumbersome:
-// 1. In resolution, functions marked as compiler-generated are considered only if
-// no functions without that flag (i.e. user-supplied) functions are found.
-// 2. In printing filename/lineno information in error messages (when developer
-// == false), the callstack is searched ignoring compiler-generated functions.
-// 3. Assignment operations flagged as 'compiler generated' shall contain only
-// field assignments and assignment primitives.
+// The compiler-generated flag has these meanings:
+// 1. In various parts of the compiler, when printing printing filename/lineno
+//    information in error messages, callstack locations marked
+//    compiler-generated may be ignored when developer == false.
+// 2. In relation to determination of POD types when "compiler generated"
+//    applies to assignment, copy, initialization routines
+// 3. When additional checking for user-written code can be relaxed
+//    for functions added by the compiler (e.g. with error handling)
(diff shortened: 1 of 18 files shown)


bradcray:
I have not reviewed this PR, but am happy for this conceptual change and agree that these two concerns have been entangled for too long.  It seems likely to me that there will be other routines that will want the "last resort" flag over time -- e.g., I could imagine it to all/most "throwOpError" routines in ChapelBase.chpl in order to permit user overloads of operators that we don't support by default and want to generate nice error messages for, but that can be separate ongoing work.

compiler/include/flags_list.h
```diff
@@ -67,14 +67,14 @@ symbolFlag( FLAG_COERCE_TEMP , npr, "coerce temp" , "a temporary that was stores
...
-// The compiler-generated flag is already overloaded in three ways.  We may
-// want to split it if this becomes cumbersome:
-// 1. In resolution, functions marked as compiler-generated are considered only if
-// no functions without that flag (i.e. user-supplied) functions are found.
-// 2. In printing filename/lineno information in error messages (when developer
-// == false), the callstack is searched ignoring compiler-generated functions.
-// 3. Assignment operations flagged as 'compiler generated' shall contain only
-// field assignments and assignment primitives.
+// The compiler-generated flag has these meanings:
+// 1. In various parts of the compiler, when printing printing filename/lineno
```
lydia-duncan: duplicate "printing"

compiler/resolution/initializerResolution.cpp
```diff
@@ -193,11 +193,11 @@ static void gatherInitCandidates(CallInfo&                  info,
 
 static void doGatherInitCandidates(CallInfo&                  info,
                                    Vec<FnSymbol*>&            visibleFns,
-                                   bool                       generated,
+                                   bool                       lastResort,
                                    Vec<ResolutionCandidate*>& candidates) {
   forv_Vec(FnSymbol, visibleFn, visibleFns) {
     // Only consider user functions or compiler-generated functions
```
lydia-duncan: Maybe update this comment and the one in visibleFunctions.cpp?

PR #6792, opened 2017-07-21
Performance warning on channel.lines()

ben-albrecht:
Warn users about locality-unawareness of the `channel.lines()` iterator in the documentation.

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -3274,6 +3274,12 @@ inline proc channel.read(ref args ...?k,
 
   Only serial iteration is supported.
 
+  .. warning::
+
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Once there is a leader/follower 2.0 issue up, this example should be linked in it.

mppf:
I don't think we should make a leader/follower 2.0 issue; instead just make issues for the specific things (like can't yield from on).

ben-albrecht:
"can't yield from on" is a design choice, rather than a bug though. This design choice may be revisited in the leader/follower 2.0 discussion. This is what I gather from talking to @benharsh about it, at least.

PR #6795, opened 2017-07-21
Fix gasnet installed

### LINKED ISSUE #6793: ./configure + make install for gasnet tries to use launcher from build directory
If configuring and installing a GASNet-enabled Chapel build, as in issue #6749, the launcher tries to refer to binaries in the original build directory such that, if it has been cleaned up, the launch will fail (at least for the amudp launcher, I have not tried others).  E.g., when running a multi-locale program from a gasnet installation of Chapel, I get:

```
./a.out -nl 5 --verbose
$CHPL_HOME/third-party/amudprun/install/darwin-clang/nodbg/bin/amudprun  -np 5 ./a.out_real -nl 5 --verbose
```

where `$CHPL_HOME` refers to the original Chapel directory where I did the `make`, not the installation location.  If that directory has not been removed / clobbered, the installation works fine.  If it has been, the result is:

```
internal error: execvp() failed: No such file or directory
```

This suggests that `make install` is still not working correctly for GASNet.

mppf:
Helps with #6793.

Fix gasnet installed

[developed by @mppf, reviewed by @bradcray]

GASNet install was not working because the path that the launcher uses to access amudprun and the like was hard-coded in the runtime to point to the location from which Chapel was built, not where it was installed.  This change captures these key paths at compile-time, relative to the compiler's installation, and has the runtime query those when invoking the launcher instead.

So far, this change is specific to the amudprun launcher.  The equivalent changes would be required for other GASNet launchers in sibling directories.  Noting that in the original issue #6793 as a placeholder.

- [x] testing

### DIFF
--- a/compiler/main/driver.cpp
+++ b/compiler/main/driver.cpp
@@ -261,10 +261,13 @@ static void setChplHomeDerivedVars() {
 
 static void saveChplHomeDerivedInEnv() {
   int rc;
+  envMap["CHPL_RUNTIME_LIB"] = strdup(CHPL_RUNTIME_LIB);
   rc = setenv("CHPL_RUNTIME_LIB", CHPL_RUNTIME_LIB, 1);
   if( rc ) USR_FATAL("Could not setenv CHPL_RUNTIME_LIB");
+  envMap["CHPL_RUNTIME_INCL"] = strdup(CHPL_RUNTIME_INCL);
   rc = setenv("CHPL_RUNTIME_INCL", CHPL_RUNTIME_INCL, 1);
   if( rc ) USR_FATAL("Could not setenv CHPL_RUNTIME_INCL");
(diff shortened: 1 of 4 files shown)


mppf:
@ben-albrecht: review me?

bradcray:
Assuming this works, I believe there are a bunch of other sibling Makefile.share files that need a similar mod.  Essentially, anything else that sets LAUNCH_PATH, I believe.

bradcray:
@ben-albrecht : I haven't done any in-depth review of this, but gave it a quick try and believe that it fixes the issue.

bradcray:
I've reviewed this and think it looks good.  Rather than worrying about the other conduits for now, I'm going to merge as-is and then add a TODO to deal with the other conduits to the open issue, primarily due to lack of cycles today and a desire to try this out on TIO before vacation.

PR #6802, opened 2017-07-23
llvm.invariant.start tests

coodie:
This PR introduces tests for changes in #6706.


test/llvm/llvm-invariant/function_local_const.chpl
```diff
@@ -0,0 +1,21 @@
+// This test checks if llvm.invariant.start is added to function local constants
+// Normally primitive type constants could be used, but chapel compiler
+// performs some constant propagation, which doesn't happen in case of records.
+
+record A
+{
+  var a : int;
+}
```
mppf: We have some ongoing work with initializers. Can you add a local and global test using this pattern?

``` chapel
record A
{
  var a : int;
  proc init(arg:int) {
    a = arg;
    super.init();
  }
}
```

That will cause the compiler to follow some code paths that aren't exercised in your tests so far but which we expect to become widely usedin the future.

coodie: @mppf 
I'm not sure what should I test as initializers make creating new records differently:

Here is how constructors work:
```
%tmp = alloca %A_chpl;
%localConst = alloca %A_chpl;
call construct_A(%A_chpl* %tmp, i64 %arg);
store %A_chpl %tmp, %A_chpl* %localConst
```

and here is how initializers work:

```
%localConst = alloca %A_chpl;
call init_A(%A_chpl* %tmp, i64 %arg);
```

Thus there is no final store which would mark that this variable is constant from this point giving opportunity to mark it using llvm.invariant.start, my algorithm covers a lot of cases because chapel compiler uses `store from temporary to final` a lot while in this case it doesn't happen so it doesn't do that.

mppf: @coodie - right, so in other words:
 1. we'll need to use another approach to put llvm.invariant.start after these `init` calls
 2. the current approach just leaves this case alone (and in particular doesn't generate incorrect IR for it)

mppf:
@coodie - could you please double-check that these tests pass the way you have written them? I'm seeing testing failures with the following ones when run with start_test in an LLVM configuration with this branch merged with master and with the llvm-invariant branch.

[Error matching program output for llvm/llvm-invariant/function_local_const]
[Error matching program output for llvm/llvm-invariant/global_const]
[Error matching program output for llvm/llvm-invariant/program_constructs]

coodie:
I fixed that. These changes were related to tbaa, which generation was changed.

PR #6807, opened 2017-07-24
Upgrade to gasnet 1.28.2

ronawho:
This upgrades to GASNet 1.28.2, while keeping our traditional patch to support
running gdb/lldb on each Chapel program instance with the amudprun launcher
and our 64-bit ARM cross-configuration file.

1.28.2 changed the Cray cross-configure script names, and removed some amudprun
makefiles we used, so they had to be pulled in from GASNet's source tree.


third-party/gasnet/Makefile
```diff
@@ -51,15 +51,16 @@ CHPL_GASNET_CFG_OPTIONS += --disable-gemini --disable-aries
 endif
 endif
 ifeq ($(CHPL_MAKE_TARGET_PLATFORM),cray-xc)
-CHPL_GASNET_CFG_SCRIPT=cross-configure-crayxc-linux
+CHPL_GASNET_CFG_SCRIPT=cross-configure-cray-aries-alps
```
ronawho: There are *-slurm variants as well, but the only diference between them is the MPIRUN_CMD, which I don't think we care about (and I really don't want to have to rebuild gasnet based on the launcher, since that'll add ~20 minutes to our module build time):

`diff cross-configure-cray-aries-alps cross-configure-cray-aries-slurm`
```diff 
< MPIRUN_CMD="${APRUN:-aprun} -n %N %C" ; export MPIRUN_CMD  # launch command for MPI jobs
---
> MPIRUN_CMD="${SRUN:-srun} -K0 -W60 %V -mblock -n %N %C" ; export MPIRUN_CMD  # launch command for MPI jobs
```

third-party/gasnet/Makefile
```diff
@@ -51,15 +51,16 @@ CHPL_GASNET_CFG_OPTIONS += --disable-gemini --disable-aries
...
 else
-CHPL_GASNET_CFG_SCRIPT=cross-configure-crayxe-linux
+CHPL_GASNET_CFG_SCRIPT=cross-configure-cray-gemini-alps
 endif
 XTRA_CONFIG_COMMAND=cp --update $(GASNET_SUBDIR)/other/contrib/$(CHPL_GASNET_CFG_SCRIPT) $(GASNET_SUBDIR)
 XTRA_POST_INSTALL_COMMAND=rm -f $(GASNET_SUBDIR)/$(CHPL_GASNET_CFG_SCRIPT)
 else
 
 ifeq ($(CHPL_MAKE_TARGET_PLATFORM),cray-xt)
+# TODO cross-configure script has been removed, drop xt support?
```
ronawho: Not sure what to do about this yet. I think we can/should just retire XT support, but I'll check with the rest of the team on Monday.

ronawho:
Did a full gasnet-everything run and a `make check` for gasnet-aries and gasnet-gemini. Did a full performance run for gasnet-aries as well and didn't see any noticeable performance differences between 1.28.0 and 1.282

ronawho:
@dmk42 could you verify aarch support?

third-party/gasnet/Makefile
gbtitus: I also think dropping XT support is probably okay.  I'd be surprised if any were still running.

gbtitus:
Looks good!

dmk42:
Looks good.
Tested and verified working on aarch64.
👍

PR #6811, opened 2017-07-24
Default executable name based on main module name not a.out

### LINKED ISSUE #6283: chpl foo.chpl -> what executable?
Today, if you compile a .chpl file without providing a `-o` argument, you get `a.out`.

``` chapel
// foo.chpl
writeln("Hi");
```

```
$ ls
foo.chpl
$ chpl foo.chpl

$ ls
a.out    foo.chpl
```

I think this is an archaic choice. (the name 'a.out' refers to the predecessor binary format to ELF). It seems better to name the executable `foo` in this case. A little survey:
 * C, C++ compilers default to a.out (at least on some systems)
 * D defaults to the file prefix, e.g. 'foo' in this case
 * Swift defaults to the file prefix, e.g. 'foo' in this case
 * Rust defaults to the file prefix, e.g. 'foo' in this case

Should we change the compiler to emit `foo` in this case instead of `a.out` ?

mppf:
Closes #6283.

Following the proposal in that issue, the default executable name is now based upon the main module (which is typically the name of the .chpl file supplied on the command line).

Adds a custom sub_test-based test to check that this works correctly. The normal sub_test script always supplies a `-o` argument and that is why a custom script is used to test this feature.

Verified new test passes with quickstart+GASNet.
Passed full local testing.

### DIFF
--- a/test/compflags/ferguson/default-binary-name/sub_test
+++ b/test/compflags/ferguson/default-binary-name/sub_test
@@ -1,0 +1,62 @@
+#! /usr/bin/env python
+
+import subprocess
+import sys
+import os
+
+if len(sys.argv)!=2:
+  print 'usage: sub_test COMPILER'
+  sys.exit(0)
+
+# Find the base installation
+compiler=sys.argv[1]
+if not os.access(compiler,os.R_OK|os.X_OK):
+  Fatal('Cannot execute compiler \''+compiler+'\'')
+
+# find current directory
+
+def run(source, exe_name, compopts, expect): 
+  p = subprocess.Popen([compiler, source] + compopts,
+                       stdout=subprocess.PIPE)
+  myoutput = p.communicate()[0]
+  sys.stdout.write(myoutput)
+  if p.returncode != 0:
+    sys.stdout.write("[Error matching compiler output for %s (%s)]\n"%(source, exe_name))
+    sys.exit(1)
+  else:
+    sys.stdout.write("[Success compiling %s (%s)]\n"%(source, exe_name))
+
+  returncode = -1
+  myoutput = ""
+
+  if os.access(exe_name,os.R_OK|os.X_OK):
+
+    numlocales = "0"
+    if "NUMLOCALES" in os.environ:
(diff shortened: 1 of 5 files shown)


test/compflags/ferguson/default-binary-name/sub_test
```diff
@@ -0,0 +1,62 @@
...
+    sys.stdout.write("[Error matching program output for %s (%s)]\n"%(source, exe_name))
+
+  if os.access(exe_name, os.F_OK):
+    os.remove(exe_name)
+  if os.access(exe_name + "_real", os.F_OK):
+    os.remove(exe_name + "_real")
+
+
+run("test-default-binary-name.chpl", "test-default-binary-name", [], "Hello")
+run("test-default-binary-name.chpl", "test_exe", ["-o", "test_exe"], "Hello")
```
daviditen: Could you add another test that sets CHPL_EXE_NAME to "test_exe" to show that setting it through the environment variable still works as expected?

daviditen:
I asked to add another test, but looks good overall.

test/compflags/ferguson/default-binary-name/sub_test
mppf: Done, and that directory still passes testing. I plan to merge soon.

bradcray:
I'm noting that there are documentation changes that need to go along with this change.  For example, at the very least, the man page and executing.rst are out-of-date w.r.t. the current behavior of the compiler, but a git grep suggests that there are several other potential cases as well.

mppf:
@bradcray - thanks for noticing. I've taken a crack at fixing these in PR  #6978.

bradcray:
Now that I'm a bit more caught up on email, I think this change warrant[ed/s] a mail to chapel-developers as well.  It took me a long time while on the road to determine why running ./a.out wasn't running the program I had just compiled anymore.  I could imagine anyone not accustomed to using `-o` and not tracking every commit running into the same confusion.  It might also be worth telling them about the workaround to set CHPL_EXE_NAME to `a.out` to restore the old behavior.

PR #6813, opened 2017-07-24
Add a warning for potentially confusing module name

lydia-duncan:
Via chat, a developer ran into an issue when their code was of the form:

Test.chpl
```chapel
require "blah.h";
module Test { ... }
```
They ended up needing to `use Test;` twice in order to access the contents
of the explicitly defined module, because it wasn't clear that having top
level statements beyond the module definition caused the insertion of an
implicit wrapper module named after the file.

This change adds a warning when a module has the same name as its
parent module if the parent module is an implicit file module.

It does this by creating a new flag and attaching it to the ModuleSymbol for
compiler inserted implicit file modules, then checking against that flag during
checkParsed.cpp.

I created two tests to verify this output, one that does not generate the
warning and one which does.

Passed a full std paratest


compiler/passes/checkParsed.cpp
```diff
@@ -314,6 +317,21 @@ checkFunction(FnSymbol* fn) {
...
 }
 
+static void nestedName(ModuleSymbol* mod) {
+  if (mod->defPoint == NULL) {
+    return;
+  }
+
+  ModuleSymbol* parent = mod->defPoint->getModule();
+  if (parent != mod &&
+      strcmp(mod->name, parent->name) == 0 &&
```
mppf: I'm pretty sure you should be able to pointer-compare the names here because of 'astr'

compiler/passes/checkParsed.cpp
```diff
@@ -314,6 +317,21 @@ checkFunction(FnSymbol* fn) {
   }
 }
 
+static void nestedName(ModuleSymbol* mod) {
+  if (mod->defPoint == NULL) {
+    return;
+  }
+
+  ModuleSymbol* parent = mod->defPoint->getModule();
+  if (parent != mod &&
```
mppf: parent == mod seems pretty weird. Does that have something to do with internal/standard module ?

mppf:
Looks pretty good. I'm curious if the strcmps could be pointer compares (relying on astr). Also, I think it'd be worthwhile in this case to add a USR_PRINT note after the error describing how to fix it (e.g. "Move all statements within the module").

compiler/passes/checkParsed.cpp
lydia-duncan: I thought I remembered the rootModule being its own parent for some strange reason, but I didn't verify that came up.  I'll do that now

lydia-duncan:
Thanks Michael!  Is the USR_PRINT sufficient or potentially confusing?

mppf:
> warning: module 'modSameNameAsFileBad' has the same name as the implicit file module
> note: move all statements within the module to avoid this

Not bad, but 'avoid this' seems awkward, and I'm hoping for something more suggestion-y and less certain of itself. Maybe this:

> warning: module 'modSameNameAsFileBad' has the same name as the implicit file module
> note: did you mean to include all statements in the module declaration?

?

lydia-duncan:
Hmm, yeah, that sounds better to me, too!

mppf:
@lydia-duncan I don't object to any of this but it just occurred to me to wonder if it would be philosophically better design to put the warning in parseFile instead of adding a flag and checking the flag.

lydia-duncan:
That's an interesting thought, but I think we should stick with the current implementation - as it stands today, the check gets run on the module that would cause the warning (the explicit module) rather than starting from the parent module and looking at its children to see if any of them are invalid.  That seems more straight-forward to me, but I'm happy to discuss.

Is your objection to adding the flag and checking it simply because it has that one purpose that is served fairly quickly?

mppf:
I think it's good to look for ways to reduce inter-pass dependencies. I'm not worried about this one though (it's pretty simple). It was a philosophical thought.

PR #6816, opened 2017-07-24
Test perf of new forall-stmt node in perf playground

ronawho:


ronawho:
@vasslitvinov merge this whenever you're ready to start perf testing. 

Results will be up at chapel.sourceforge.net/perf/chapcs/forall-stmt the following day

vasslitvinov:
@ronawho how to disable this testing, assuming that it is not needed now that #5295 is on master?

PR #6820, opened 2017-07-25
Remove XT support

ronawho:


ronawho:
@gbtitus this is in response to https://github.com/chapel-lang/chapel/pull/6807#discussion_r128943114

I effectively "grepped" for `cray-xt` and `-w xt`. Any other references you can think of and/or find?

PR #6824, opened 2017-07-25
Don't let complex.h define I for C code used in Chapel

mppf:
It's not necessary for Chapel complex numbers and it causes interoperability problems. In particular, openssl's rsa.h includes a function argument named I. The included new test `test/users/ferguson/using-I-in-c/` demonstrates how that might be a problem. Without this fix, it fails to compile with gcc 6.3 with errors like this: 

```
./test.h: In function ‘g’:
./test.h:4:7: error: expected identifier or ‘(’ before ‘__extension__’
   int I = 1;
       ^
./test.h:5:4: error: lvalue required as increment operand
   I++;
    ^~
./test.h: At top level:
./test.h:9:32: error: expected ‘)’ before ‘__extension__’
 static inline int f(const int* I) {
                                ^
```

Passed test/types/complex/.
- [x] full local testing


mppf:
@dmk42 - could you look at this one? thanks

dmk42:
I don't agree with this approach.  If we are having problems with openssl, we should push back against them rather than undo a standard definition.

In the case of this particular issue, there are some implementation changes likely for complex numbers, and there is no guarantee that the things we need from complex.h will not change.

mppf:
@dmk42 - I would normally agree with you, but in this case the standard is a little bit mushy. The standard explicitly says that `I` is a macro and that people can undefine it if they don't want it. There is another name for it: `_Imaginary_I`. It's certain that we could use `_Imaginary_I` anywhere we might conceivable need to generate `I` with the Chapel compiler.

See http://pubs.opengroup.org/onlinepubs/9699919799/ which contains:

> An application may undefine and then, perhaps, redefine the complex, imaginary, and I macros.

...

> The choice of I instead of i for the imaginary unit concedes to the widespread use of the identifier i for other purposes. The application can use a different identifier, say j, for the imaginary unit by following the inclusion of the <complex.h> header with:
>
> #undef I
> #define j _Imaginary_I

For these reason's, I'm not convinced:
 a) that the OpenSSL developers will regard it as a bug, and
 b) that we'd be "undoing a standard definition" if we use this `#undef` approach

dmk42:
`_Imaginary_I` is not a valid replacement in general, because imaginary numbers are optional.  To define something else in place of `I` would require something more complicated, like this.
```c
#ifdef _Imaginary_I
#define J _Imaginary_I
#else
#define J _Complex_I
#endif
```
As for what the OpenSSL people would be willing to change, that has nothing to do with whether or not they consider it a bug, and we will not know their willingness unless we ask them.

We really need to push back against them before we consider something like this.

mppf:
@saru95 @panzone - looks like you'll have to work with OpenSSL to fix that bug.

saru95:
@mppf @panzone @dmk42 OpenSSL doesn't agree to this. More here: https://github.com/openssl/openssl/issues/4022

dmk42:
Well, I'm pretty miffed at the OpenSSL folks for messing this up.  They have left us no way to get it right.  The biggest issue is that the problem works both ways.  If we leave things as they are, OpenSSL doesn't work.  If we `#undef I`, then some reasonable header file that does want `complex.h` will fail.  Consider this.
```c
// in the C part
#undef I
```
```chapel
// in the Chapel part
require "goodguy.h";
```
If `goodguy.h` does a `#include <complex.h>`, it will not really get the header file, because the inclusion guard will have already been set the first time around.  Therefore, the definition of `I` will be missing and `goodguy.h` is hosed.

Since Chapel is, among other things, intended to be used for scientific applications, there is a good chance that someone will want to include a header that does need complex numbers.

Could we somehow mark offending header files and then wrap them like this?
```c
#undef I
#include "badguy.h"
#undef I
#ifdef _Imaginary_I
#define I _Imaginary_I
#else
#define I _Complex_I
#endif
```

dmk42:
@saru95 @panzone - does OpenSSL work when it's wrapped that way?  I guess that would be the first thing to find out before we decide on what mechanism to use.

panzone:
@mppf @dmk42 I've tried the wrapper.h and it seems to work. @saru95, can you confirm it works with your examples?

PR #6825, opened 2017-07-25
Add dsiAssignDomain method and remove inherited fields in ArrayViewReindexDom

daviditen:
The method dsiAssignDomain is now required for domains, so add it to
ArrayViewReindexDom. The fields rank, idxType, stridable are now inherited
from BaseRectangularDom, so remove them from ArrayViewReindexDom.

Passed a full local paratest.


mppf:
👍

PR #6828, opened 2017-07-26
Cleaning up some futures

vasslitvinov:
De-futurizing:

#### due probably to #6722
`distributions/kushal/assignDomain`

#### due probably to #6765
`parallel/begin/vass/begin-in-module-init`
`users/ferguson/bug-zhao`

#### while there

Removed a .prediff as it should not be needed any more.

Converted the following test so that it always compiles with --verify.
Compiler's crashing under --verify  is a more immediate
and more reliable bug than compiling the program
all the way (which can be done without --verify).

`types/records/engin/parentheseslessConstErrorField`


vasslitvinov:
@mppf and @daviditen please review post-merge.

mppf:
looks good, thanks

daviditen:
Thanks @vasslitvinov.  Looks good.

PR #6830, opened 2017-07-26
Add a warning for the install script if installing to CHPL_HOME

mppf:
- [x] verified that make install works still with prefix or target home directory
- [x] verified new error appears with either of the following sequences:
```
./configure --chpl_home=`pwd` && make && make install
./configure --prefix=`pwd` && make && make install
```

Passed full local testing.

### DIFF
--- a/util/buildRelease/install.sh
+++ b/util/buildRelease/install.sh
@@ -121,12 +121,26 @@ then
   DEST_THIRD_PARTY="$PREFIX/lib/chapel/$VERS/third-party"
   DEST_CHPL_HOME="$PREFIX/share/chapel/$VERS"
   echo "Installing Chapel split to bin, lib, share to $PREFIX"
+  if [ "$CHPL_HOME" = "$PREFIX" ]
+  then
+    echo
+    echo "Error: destination prefix is Chapel source directory"
+    echo "Please run configure again to select a different installation path"
+    exit -1
+  fi
 else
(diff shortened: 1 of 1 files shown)


mppf:
@ben-albrecht - how is this for the error for misuse of configure?

util/buildRelease/install.sh
```diff
@@ -121,12 +121,26 @@ then
   DEST_THIRD_PARTY="$PREFIX/lib/chapel/$VERS/third-party"
   DEST_CHPL_HOME="$PREFIX/share/chapel/$VERS"
   echo "Installing Chapel split to bin, lib, share to $PREFIX"
+  if [ "$CHPL_HOME" = "$PREFIX" ]
```
ben-albrecht: I expected `==`

mppf: Look at http://tldp.org/LDP/abs/html/comparison-ops.html and search for "string comparison"

ben-albrecht: < This is a synonym for =.

TIL - cool. You are a bash wizard.

mppf: I think `sh` / `dash` etc don't support `==` but do support `=` for string comparison.

ben-albrecht: Great, another reason for me to dislike `dash`

mppf: you probably mean "another reason to dislike anything but `bash`"

PR #6832, opened 2017-07-26
Fix const record init error

mppf:
PR #6818 added a future for a bug @benharsh noticed. This PR makes a simple change to fix the bug. While there, it adds `astrInit` for use by the compiler, next to `astrDeinit`.

Passed full local testing.


mppf:
@benharsh - how does this look?

PR #6833, opened 2017-07-26
Fix forVersusWhilePerf.chpl

coodie:
Looking at current performance graphs one can notice that in case of LLVM chapel's version of this test there is slight performance degradation, after some investigation I realized that this test is not written well due to few factors:

- Chapel's default size for integer is 64-bits while in case of C it is usually 32-bits, so the tests had different performance due to usage of different types of integers, in this case there was better performance for C version.
- Chapel's version cannot perform constant propagation due to marking variables as "config const". C's version uses "static const" what enables to do constant propagation which reduces register pressure, that in this case made some difference.

After changing all "int" to "int64_t" and "static const int64_t" to "int64_t" the Chapel's version performance closely matches the C version performance.

I guess @lydia-duncan is the best person to review this PR since she wrote those tests in first place.

### DIFF
--- a/test/statements/lydia/forVersusWhilePerf.chpl
+++ b/test/statements/lydia/forVersusWhilePerf.chpl
@@ -32,7 +32,7 @@ proc while_loop()
       else
         res2 = res2 / 2;
       j += 1;
-    } 
+    }
     i += 1;
   }
   return res2;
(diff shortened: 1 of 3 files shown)


lydia-duncan:
Sorry, didn't get a notification from the edit.  Looking at this now!

test/statements/lydia/forVersusWhilePerf.chpl
```diff
@@ -29,9 +32,19 @@ proc main() {
       else
         res2 = res2 / 2;
       j += 1;
-    }
+    }
```
lydia-duncan: This line adds some trailing whitespace

test/statements/lydia/forVersusWhilePerf.chpl
```diff
@@ -43,7 +56,7 @@ proc main() {
     writeln("For loop underwent ", numIterations, " iterations ", numTrials,
             " times in ", t1.elapsed(TimeUnits.milliseconds)/1000, " seconds");
     writeln("While loop underwent ", numIterations, " iterations ", numTrials,
-            " times in ", t2.elapsed(TimeUnits.milliseconds)/1000, " seconds");
+          " times in ", t2.elapsed(TimeUnits.milliseconds)/1000, " seconds");
```
lydia-duncan: This changes the whitespace for the worse

lydia-duncan:
Awesome, thanks!

PR #6835, opened 2017-07-26
Extend 'use' warning when it causes an implicit file module to also cover 'require'

lydia-duncan:
When discussing the warning message added in #6813, it was noted that
if the warning for 'use' statements as exercised by [topleveluse.chpl](https://github.com/chapel-lang/chapel/blob/master/test/users/chris/topleveluse.chpl) had been
extended to 'require' statements, there would have been less confusion.

This change partially extends the warning.  Unfortunately, to fully support
this change for 'require' statements will involve a change to the
implementation so that we actually create an interesting AST node in all
cases, instead of handling `require someFile.chpl;` at parse time and then
inserting a no-op.

While I think it would be better to have the warning in all cases, I would
like to discuss those changes with Brad first, and I had an implementation
that handles the other cases anyways, so we might as well benefit from it.

Adds two tests - one for 'require' statements on a .h file which demonstrates
the new warning, and a future for 'require' statements on a .chpl file (which
currently does not demonstrate the new warning).

Full std paratest passed


lydia-duncan:
@mppf - thoughts?

compiler/parser/parser.cpp
```diff
@@ -674,19 +679,33 @@ static bool containsOnlyModules(BlockStmt* block, const char* path) {
     }
   }
 
-  if (hasUses == true && hasOther == false && moduleDefs == 1) {
+  if ((hasUses == true || hasRequires == true) &&
+      hasOther == false &&
+      moduleDefs == 1) {
```
mppf: I think this would be easier to read as nested conditionals, of pulling some parts of the conditional out as variables. But that's a style issue & it's OK this way.

test/modules/lydia/modSameNameAsFileBadRequire.chpl
```diff
@@ -0,0 +1,7 @@
+require "modSameNameAsFileGood.chpl";
```
mppf: Can you make a test that triggers the warning whilst having both a top-level `require` and `use` ?

PR #6836, opened 2017-07-26
Add a new .compopts file to suppress a new warning message from cce/8.6.*

awallace-cray:
Adds -hnomessage=1046 for Cray cce compiler, to suppress this warning message in 
examples/users-guide/base/floatValues.chpl  : 

CC-1046 craycc: WARNING File = floatValues.c, Line = 39

  The floating-point value cannot be represented exactly.
  myReal_chpl = 0x1p-1;

### DIFF
--- a/test/release/examples/users-guide/base/floatValues.compopts
+++ b/test/release/examples/users-guide/base/floatValues.compopts
@@ -1,0 +1,20 @@
+#!/usr/bin/env python
+
+# With cce/8.6.0.*, floatValues.chpl emits new warning messages like,
+#
+# CC-1046 craycc: WARNING File = floatValues.c, Line = 39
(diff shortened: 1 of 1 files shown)


awallace-cray:
For use with Cray target compiler version cce/8.6.1, which is still in test.

ronawho:
Do we think this is a useful warning at all? 

Should we just disable the warning in [Makefile.cray-prgenv-cray](https://github.com/chapel-lang/chapel/blob/master/make/compiler/Makefile.cray-prgenv-cray#L18-L26) like we do for some others

awallace-cray:
Makes sense to me.  Let's do that.

awallace-cray:
Replaced by PR #6906

Issue #6514, opened 2017-06-21
target organization of ForallStmt


vasslitvinov:
While reviewing #5295, we are setting the goals for the structure and handling of ForallStmt node:

* We want to retain ForallStmt nodes through cullOverReferences. We clearly need to have them lowered by the time today's lowerIterators comes about.

* During resolution, the iterator calls get resolved. Specifically, standalone, leader, follower(s) calls. If we need to decide at run time between fast-followers and non-fast followers, we retain both iterator calls, storing them in dedicated areas of the ForallStmt node. The loop body is resolved and we still retain only a single copy of it.

* During resolution also, the types and qualifiers of user loop variables are established.

* TBD how to handle the iterator machinery that's currently present during resolution, such as _getIterator calls, IC/IR variables, ForLoop::mIndex and ForLoop:mIterator. Idealy we want to defer those until lowerIterators, too.

* We do not want to obscure the contents of ForallStmt nodes during resolution beyond the above considerations.

* Before lowerIterators, we lower ForallStmt nodes and remove them from the tree.

ct-clmsn:
@noakesmichael @vasslitvinov are changes to the AST and compiler making it into the compiler documentation along with updates to CompilerIRTricks.txt?

vasslitvinov:
@ct-clmsn - no, not making. CompilerIRTricks.txt does not go into that level of detail so there is nothing there that could reflect these changes.
As for compilerOverview.pdf, we do not have plans to update it at the moment. Hopefully these changes will result in simpler handling of forall loops, which has not been reflected in this document anyway.

Issue #6797, opened 2017-07-21
ForallStmt future work


vasslitvinov:
This issue tracks left-over items for #5295 .

### Primary items

- [ ] Convert remaining old-style Forall loops,
i.e. where they are created using the old buildForallLoopStmt().
Including: (a) promoted expressions, (b) reduce expressions.
Convert them either to ForallStmt or to other dedicated AST nodes.

- [ ] Resolve #5255 about iterator forwarding.
In particular, give users the ability to forward iterators
without resorting to __primitive("can resolve") and similar.

- [ ] Remove the tryToken completely.
The main issue I think is its use in chpl__transferArray().
Some ideas to get rid of it there:
(a) require at least leader/follower for any domain map, or
(b) allow a forall loop to use the serial iterator if there are no
parallel iterators.

- [ ] Push lowering of ForallStmt until later in compiler pipeline.
This will preserve valuable information for later passes.

- [ ] Finalize the organization of ForallStmt #6514

- [ ] Get rid of _getIterator and _toFollower etc calls.

- [ ] Implement forall intents in a less complex way.
For example, in propagateExtraLeaderArgs(), use a dedicated data structure
to indicate intents on individual variables. So far this has relied
on argument intents and so does not convey information properly.

### Secondary items

- [ ] Consider converting `ForallExpr`s to using ForallStmt (see also
`buildForallLoopExprFromForallExpr`). One consideration is whether
there are cases where a forall expression does not necessarily
lead to executing a corresponding forall statement.

- [ ] Switch to yielding (from the specialized parallel iterators) all
shadow variables (except for reduce intents) by reference.
This is needed for example to implement `in` intents properly.
Currently they `in`-intented shadow variables are yielded by value,
making them iteration-private rather than task-private how they are
supposed to be. My only concern for this change is whether performance
will be affected.

- [ ] Un-futurize `test/parallel/forall/in-intents/both-arr-dom-const-const`

PR #6838, opened 2017-07-26
Converted withClause() to forallIntents() in ForallStmt

vasslitvinov:
We prefer forall intents to be AST nodes.
Cf. prior to this change they were in a non-AST class ForallIntents,
which was pointed to by ForallStmt::withClause().

Here we introduce ForallIntent and ForallStmt::forallIntents().
'forallIntents()' is an AList with one ForallIntent per one
Chapel-level forall intent.

Note the two class names are similar. Plural ForallIntents
is the non-AST class in BlockStmt::forallIntents representing
the entire with clause.
Singular ForallIntent is the new AST node representing
a single intent/variable, stored in ForallStmt::forallIntents()
AList.

In retrospect, I probably should have gone straight to
LoopIntentVar as per #6514. However, this change
is a simpler and more immediate improvement over #5295 
and one that Michael and I have wanted for a while.

The next steps, independent of each other, are:

* Retire BlockStmt::forallIntents and the ForallIntents class.
This means switching ALL forall loops to new representation,
probably based on ForallStmt at least initially.

* Move from ForallIntent class to LoopIntentVar class.

Testing: linux64 - full paratest; gasnet - spot check.

The commits before the rebase were bb84e1b..69baa92

Visitor changes are in a separate commit.

### DIFF
--- a/compiler/AST/ForallStmt.cpp
+++ b/compiler/AST/ForallStmt.cpp
@@ -26,36 +26,108 @@
 
 /////////////////////////////////////////////////////////////////////////////
 //
+// ForallIntent represents one forall intent in a with clause
+//
+// This is a transitional step on the way to LoopIntentVar.
+// We also plan to retire BlockStmt::forallIntents and ForallIntents class
+// by switching ALL forall statements (and the likes) to new representation.
+//
+/////////////////////////////////////////////////////////////////////////////
+
+ForallIntent::ForallIntent(TFITag intent, Expr* var, Expr* reduceExpr):
+  Expr(E_ForallIntent),
+  fiIntent(intent),
+  fiVar(var),
+  riSpec(reduceExpr)
+{
+  gForallIntents.add(this);
+}
+
+ForallIntent* ForallIntent::copyInner(SymbolMap* map) {
+  ForallIntent* _this = new ForallIntent(fiIntent,
+                                         COPY_INT(fiVar),
+                                         COPY_INT(riSpec));
+  return _this;
+}
+
+void ForallIntent::replaceChild(Expr* oldAst, Expr* newAst) {
+  if (oldAst == fiVar)
+    fiVar = newAst;
+  else if (oldAst == riSpec)
+    riSpec = newAst;
+  else
+    INT_FATAL(this, "Unexpected case in ForallIntent::replaceChild");
+}
+
+Expr* ForallIntent::getFirstChild() {
+  return fiVar;
+}
+
+Expr* ForallIntent::getFirstExpr() {
+  return fiVar->getFirstExpr();
+}
+
+Expr* ForallIntent::getNextExpr(Expr* expr) {
+  if (expr == fiVar && riSpec)
+    return riSpec;
+  else
+    return this;
+}
+
+void ForallIntent::verify() {
+  Expr::verify(E_ForallIntent);
+  verifyParent(fiVar);
+  verifyParent(riSpec);
+  bool isRI = isReduce();
+  INT_ASSERT(isRI == (fiIntent == TFI_REDUCE));
+  INT_ASSERT(isRI == (riSpec != NULL));
+}
+
+GenRet ForallIntent::codegen() {
+  INT_ASSERT(false); // this method should not be invoked
+  GenRet ret;
+  return ret;
+}
+
+void ForallIntent::accept(AstVisitor* visitor) {
+  if (visitor->enterForallIntent(this)) {
+    variable()->accept(visitor);
+    if (Expr* ri = reduceExpr())
+      ri->accept(visitor);
+    visitor->exitForallIntent(this);
+  }
+}
+
+
+/////////////////////////////////////////////////////////////////////////////
+//
 // ForallStmt represents a forall loop statement
 //
 /////////////////////////////////////////////////////////////////////////////
 
-ForallStmt::ForallStmt(bool zippered, BlockStmt* body, ForallIntents* with):
+ForallStmt::ForallStmt(bool zippered, BlockStmt* body):
   Stmt(E_ForallStmt),
   fZippered(zippered),
-  fLoopBody(body),
-  fWith(with)
+  fLoopBody(body)
 {
   fIterVars.parent = this;
   fIterExprs.parent = this;
   fIntentVars.parent = this;
+  fFIntents.parent = this;
   gForallStmts.add(this);
-}
-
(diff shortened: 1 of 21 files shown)


compiler/include/ForallStmt.h
```diff
@@ -55,30 +93,39 @@ class ForallStmt : public Stmt
...
 inline AList& ForallStmt::intentVariables()      { return fIntentVars; }
-inline BlockStmt*     ForallStmt::loopBody()   const { return fLoopBody; }
-inline ForallIntents* ForallStmt::withClause() const { return fWith; }
+inline AList& ForallStmt::forallIntents()        { return fFIntents;    }
+inline BlockStmt* ForallStmt::loopBody()   const { return fLoopBody; }
 
 // conveniences
 inline Expr* ForallStmt::firstIteratedExpr() const { return fIterExprs.head; }
 inline int   ForallStmt::numIteratedExprs() const { return fIterExprs.length; }
+inline int   ForallStmt::numFIntents()     const { return fFIntents.length; }
```
mppf: Rename to numForallIntents please

compiler/include/ForallStmt.h
```diff
@@ -55,30 +93,39 @@ class ForallStmt : public Stmt
...
+inline ForallIntent* ForallStmt::getFIntent(int index) const
```
mppf: Rename to getForallIntent please

compiler/include/ForallStmt.h
```diff
@@ -55,30 +93,39 @@ class ForallStmt : public Stmt
...
+  int           numFIntents() const;
+  ForallIntent* getFIntent(int index) const; //todo replace with LoopIntentVars
 
 private:
   bool           fZippered;
   AList          fIterVars;
   AList          fIterExprs;
-  AList          fIntentVars;  // may be empty
+  AList          fIntentVars;  // may be empty; elems are ForallIntent
+  AList          fFIntents;    // todo: replace with intentVariables
```
mppf: Can you explain what the difference is between fIntentVars  and fFIntents? The comment says fIntentVars contains ForallIntent elements. Are the comments just reversed?

compiler/include/ForallStmt.h
```diff
@@ -55,30 +93,39 @@ class ForallStmt : public Stmt
   Expr* firstIteratedExpr() const;
   int   numIteratedExprs() const;
   bool isIteratedExpression(Expr* expr);
-  int  reduceIntentIdx(Symbol* var); // todo: replace with LoopIntentVars
+  int  reduceIntentIdx(Symbol* var); // todo: replace with ForallIntents
+  int           numFIntents() const;
```
mppf: rename to numForallIntents please

compiler/include/ForallStmt.h
```diff
@@ -55,30 +93,39 @@ class ForallStmt : public Stmt
...
+  ForallIntent* getFIntent(int index) const; //todo replace with LoopIntentVars
```
mppf: rename to getForallIntent please

compiler/include/baseAST.h
```diff
@@ -563,19 +567,17 @@ static inline const CallExpr* toConstCallExpr(const BaseAST* a)
...
-    ForallIntents* fi = ((ForallStmt*)_a)->withClause();                \
-    AST_CALL_STDVEC(fi->fiVars,  Expr,              call, __VA_ARGS__); \
-    AST_CALL_STDVEC(fi->riSpecs, Expr,              call, __VA_ARGS__); \
-    AST_CALL_CHILD(fi, ForallIntents, iterRec,      call, __VA_ARGS__); \
-    AST_CALL_CHILD(fi, ForallIntents, leadIdx,      call, __VA_ARGS__); \
-    AST_CALL_CHILD(fi, ForallIntents, leadIdxCopy,  call, __VA_ARGS__); \
-    AST_CALL_CHILD(_a, ForallStmt,    loopBody(),   call, __VA_ARGS__); \
-    break;                                                              \
-  }                                                                     \
+    AST_CALL_LIST (_a, ForallStmt, forallIntents(),       call, __VA_ARGS__); \
```
mppf: 👍

compiler/include/ForallStmt.h
```diff
@@ -22,7 +22,47 @@
 
 #include "stmt.h"
 
-// forall loop statement
+
+//////////////////////////
+    // a forall intent //
+   //////////////////////////
```
mppf: Let's put a TODO somewhere (in an issue? in a comment?) that we need to think about how to merge ForallIntent and TaskIntent so that the parser only needs to emit one of these, since they are syntactically the same (aren't they)?

compiler/AST/ForallStmt.cpp
```diff
@@ -389,16 +455,24 @@ BlockStmt* ForallStmt::build(Expr* indices, Expr* iterator,
...
+
+  if (forall_intents) {
+    // transfer from 'forall_intents' into 'fs'
+    // todo: how to make the parser produce AList directly?
+    AList& fis = fs->forallIntents();
+    for (int i = 0; i < forall_intents->numVars(); i++)
+      fis.insertAtTail(new ForallIntent(forall_intents->fIntents[i],
+                                        forall_intents->fiVars[i],
+                                        forall_intents->riSpecs[i]));
+    delete forall_intents;
```
mppf: Something (PR message, comments, TODOs, issues) needs to clearly say that we don't intend to have both ForallIntents and ForallIntent going forward. What will it take to get the parser to produce ForallIntent only? Could you retire ForallIntents in this PR? If you can, you should. I think it's confusing to have both, if we can avoid it.

compiler/AST/ForallStmt.cpp
```diff
@@ -26,36 +26,106 @@
 
 /////////////////////////////////////////////////////////////////////////////
 //
+// ForallIntent represents one forall intent in a with clause
+//
+// This may be a transitional measure on the way to LoopIntentVar.
```
mppf: I don't understand why we made ForallIntent instead of just immediately introducing LoopIntentVar or TaskIntent or something along those lines that is fairly general in name. However I'm OK with it "transitionally".

compiler/AST/ForallStmt.cpp
vasslitvinov: We can't get rid of ForallIntents class right now. It is still used for representing forall loops via BlockStmt. Eliminating ForallIntents is part of the "convert the remaining forall loops" item in #6797.

PR #6839, opened 2017-07-26
More on statement support in initializers

lydia-duncan:
Begin+on and cobegin+on statements are allowed in initializers, so long as
they do not contain field initialization, super.init() calls, or this.init calls.

Prior to this change, we would get an internal error with --no-local, COMM !=
none and LOCALE_MODEL != flat if begin+on statements were present.
However, cobegin+on and on+cobegin statements were treated normally.

Adds testing for these cases over field inits, this.init() calls, super.init() calls
and in Phase 1 but not over field initializations.

Testing:
- [x] classes/initializers
- [x] classes/initializers with --no-local
- [x] classes/initializers with NUMA
- [x] classes/initializers with COMM=gasnet
- [x] paratest
- [x] paratest with --no-local


lydia-duncan:
I ran paratest and paratest --no-local prior to the cobegin tests.  I haven't run the initializers directory with gasnet or NUMA yet

PR #6842, opened 2017-07-26
Add datetime.ctime() method

benharsh:
This method behaves like the python equivalent:
https://docs.python.org/2/library/datetime.html#datetime.datetime.ctime


benharsh:
@daviditen : how does this look? I'll likely merge tomorrow to reduce potential noise for ForallStmt.

PR #6843, opened 2017-07-27
Refactor some logic in AstToText::appendName()

### LINKED ISSUE #5791: chpldoc support for module-level deinit

vasslitvinov:
Resolves #5791
De-futurizes the test: `chpldoc/module/modDeinit.doc.chpl`

While examining `AstToText::appendName(fn)` prior to this change,
some things did not feel quite right so I fixed them:

* I moved checking for FLAG_DESTRUCTOR into the case for methods.
Otherwise this check fires also for module deinit() functions,
causing appendClassName() to be invoked. The latter observes no formals
and crashes with an assert. Which is correct because a function
with no formals got no class.

* Destructor methods were displayed as MyClass.~MyClass().
I changed that to be MyClass.deinit().

* For methods, "MyClass." is prepended uniformly when `fn` is not
a primary method. I am not aware of the motivation for doing it differently.

* "this" methods are now displayed as "this".
Prior to this change, they were displayed as "MyClass".
I am not aware of the rationale for that.

* Note that the checks for `"_type_construct_"`, `"_construct_"`, FLAG_DESTRUCTOR
are useful (and unaffected by this change) for uses of AstToText other than
for chpldoc. In those cases these checks avoid introducing the underscore
ugliness.

* Todo: display constructors as init() rather than MyClass().
The challenge is that when fn->hasFlag(FLAG_CONSTRUCTOR), we may not know
whether it should be displayed as a primary or secondary method, i.e.
"MyClass.init" or just "init". Lydia adds that "initializers are not currently tagged with FLAG_CONSTRUCTOR and instead print out their name just like any other method."

Testing: linux64 --verify


vasslitvinov:
@lydia-duncan would you review?

lydia-duncan:
Reading your PR message, I should point out that initializers are not currently tagged with FLAG_CONSTRUCTOR and instead print out their name just like any other method.  Did you write a test that proved otherwise?

vasslitvinov:
@lydia-duncan I ran a full linux64 paratest. I added the tests as  you suggested. Thank you for the review. OK to merge?

PR #6849, opened 2017-07-27
Forwarding bug fixes

mppf:
* Fix my earlier misunderstanding of the role of FLAG_INVISIBLE_FN and generics.
  In particular, finding visible functions only works with visible
  functions (not generic instantiations which are marked with
  FLAG_INVISIBLE_FN). Since forwarding is really a process of finding
  additional visible functions, it should only work with generic
  functions (not instantiations thereof). So it should ignore methods
  with FLAG_INVISIBLE_FN.
* For a forwarding statement in a generic type, now the type of 'this' in
  the forwarding function is also generic; and the function is marked
  with FLAG_GENERIC in that case in order to work with the rest of
  resolution.
* No longer hard-codes the best candidate for the method call in the
  generic forwarder function, so that it can work naturally when
  instantiated multiple times.
* Fixes forwarding of methods using default expressions or type expressions based upon other arguments
* Fixes forwarding of methods using default expressions or type expressions based upon `this` expressions
* Additionally added some more tests for some more corner cases

Passed full local testing.


mppf:
@vasslitvinov - you reviewed the PR adding forwarding (#5058). Would you be willing to look at this additional bug fix for it? Thanks.

PR #6856, opened 2017-07-27
Move global well-known types/fns to wellknown.h/.cpp

mppf:
The (relatively new) files wellknown.h and wellknown.cpp are responsible for managing certain types and functions that are identified during parse time and are known to the compiler.

This PR simply moves the global variables managed by wellknown.h/.cpp to be declared in those files. I think that represents an improvement since it keeps the managed globals close to where they are managed.

This is a trivial clean-up.
Discussed with @noakesmichael, who agreed this is an improvement.
Passed full local testing.

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -35,17 +35,6 @@
 AggregateType* dtObject            = NULL;
 
 AggregateType* dtString            = NULL;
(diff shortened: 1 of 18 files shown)


mppf:
@noakesmichael - if you think this code-movement PR represents an improvement, let's do it. If not, I'll drop it. What do you think? thanks.

Issue #6851, opened 2017-07-27
Define record "inheritance"


lydia-duncan:
We used to define record inheritance as "field only".  We've since decided that we hadn't sufficiently defined what that would mean, what should happen with methods, etc, and so removed it from the language specification and will (shortly) make it an error for a record to inherit.

This is a placeholder issue for the discussion on what record inheritance should mean, so that the error message can point users to it and we can get feedback on what would be wanted.  We don't anticipate working on this issue immediately, but would likely be disappointed if Chapel had not figured this out 5-10 years down the road

psahabu:
Record inheritance would be useful for error memory management if we don't get an Owned/Shared story first. I don't know if that should be a driving motivation though.

vasslitvinov:
@lydia-duncan @bradcray I suggest to close this issue. We did away with record inheritance officially in #5892 because we had - and have - no current plans to consider it. We can reopen this issue - or create a new one - when+if appropriate.

vasslitvinov:
Bonus: how about remove the tests that, since #6860, say "This test will probably be useful when we rethink the feature" and reference this issue:
```
test/classes/deitz/inherit-class-record/record_inherit1.chpl
test/classes/deitz/inherit-class-record/record_with1.chpl
test/classes/deitz/inherit/test_value2.chpl
test/classes/ferguson/generic-inherit-record-no-parent-methods.chpl
test/classes/figueroa/RecordDestructor3.chpl
test/classes/marybeth/test_dispatch_record.chpl
test/functions/jplevyak/scoping/visibility-4.chpl
test/users/vass/type-tests.isSubtype-records.chpl
```

lydia-duncan:
I'm ambivalent to closing the issue but against removing the tests.

mppf:
I'm closing this issue to reflect that our current plan is not to support record inheritance. Of course, in the future, that can be revisited with a new issue proposing a particular meaning and showing a compelling use case.

bradcray:
At one point, we discussed having a "mix-in" concept instead of supporting record inheritance.  However, since we  implemented `forwarding` fields, that has felt to me to be sufficiently mix-in-like and to provide inheritance-like behavior to not make me wish for more than that, personally.

> I'm ... against removing the tests.

[...]

PR #6860, opened 2017-07-27
Add an error when a record tries to inherit

### LINKED ISSUE #4954: record inheritance should be an error
Currently, record inheritance is syntactically allowed but can only be used to inherit fields.

We would rather have a 'mix-in' type strategy and disable record inheritance entirely.

lydia-duncan:
Resolves #4954.

We removed record inheritance from the language specification due to not having
ironed out enough of the details.  However, field inheritance for records was
already implemented, and we did not remove it.

This change adds an error message after parsing if we detect that a record has
been defined which inherits.  It also points users encountering this to an
issue I created for discussion on record inheritance (#6851).

It does not remove field inheritance for records.

Testing:
Adds a test to lock in the error message.  Converts 7 tests into futures, and
updates the expected output for 5 futures.  Also, splits one test into a
working portion and a now failing portion.

Passed a full paratest with futures.


lydia-duncan:
Retested after the merge from master, all clean.

PR #6752, opened 2017-07-19
TOML library

Spartee:
Add the Toml library to the tests directory along with ten toml files in the form of .good tests along with a unit test for the Node class. This is the toml library for use in the package manager. The interface is under the issue #6654 right now and I am still looking for more feedback on it. 

- [x] Use `channel.lines()` in `TomlParse(input: channel)` ~~when #6724 is merged~~
- [x] Rename Node to something more relevant to Toml

PR #6864, opened 2017-07-28
Renamed Toml.chpl to TOML.chpl

vasslitvinov:
Toml.chpl was introduced in #6752
Judging by the error message in this file:

    test/modules/packages/Toml/test/quoteKeyPeriod.bad

it was tested as TOML.chpl. I am renaming it so that the tests pass.

I am also adding TOML.notest because Toml.chpl lacks a .good file
and is probably not intended to be tested by itself anyway.

@Spartee please check if this is what you intended.


Spartee:
@vasslitvinov Thank you for catching this; yes, this is what I intended.

PR #6865, opened 2017-07-28
Retire XT support

ronawho:
So far as we know nobody is running Chapel on an XT (or even running an XT for
that matter) so retire the remaining XT support.

GASNet 1.28.2 (#6807) removed the XT cross-configure script, which motivated
removing any remaining support we had.


runtime/etc/Makefile.comm-gasnet
```diff
@@ -20,12 +20,12 @@ ifneq ($(MAKE_LAUNCHER), 1)
 include $(GASNET_SETUP_MAKEFILE)
 
 #
-# On the XT/XE, we often build GASNet using a different compiler than the
+# On the XE/XC, we often build GASNet using a different compiler than the
```
gbtitus: Could start this sentence with "On Cray X*, ..." to future-proof it.

gbtitus:
This looks fine as-is, but I think we could even consider changing the Makefile filter() checks to just be against "cray-x" instead of enumerating the individual models.  It's hard to predict marketing, but I suspect that for the foreseeable future we'll continue with the "Cray X*" nomenclature for the large systems.  (Of course if I'm wrong and the next one is "Cray Z2" you'll curse me. :-)

ronawho:
>This looks fine as-is, but I think we could even consider changing the Makefile filter() checks to just be against "cray-x" instead of enumerating the individual models. It's hard to predict marketing, but I suspect that for the foreseeable future we'll continue with the "Cray X*" nomenclature for the large systems. (Of course if I'm wrong and the next one is "Cray Z2" you'll curse me. :-)

I'm not opposed to that, but I'd prefer to do that as a separate PR

PR #6873, opened 2017-07-28
Enhanced viewFlags()

vasslitvinov:
For additional debugging convenience:

* Also display the type qualifier.

* When given a Type, DefExpr, of SymExpr, redirect to the corresponding Symbol instead of giving up.

While there, in a separate commit moved the "Qualifier to string" conversion into `qualifierToStr(Qualifier q)` and to type.cpp, out of type.h. Also brought together Qualifier-related code within type.cpp.

### DIFF
--- a/compiler/AST/flags.cpp
+++ b/compiler/AST/flags.cpp
@@ -19,9 +19,8 @@
 
 #include "flags.h"
 
-#include "baseAST.h"
+#include "expr.h"
 #include "stringutil.h"
-#include "symbol.h"
 
 
 typedef Map<const char*, int> FlagMap;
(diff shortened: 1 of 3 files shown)


compiler/AST/flags.cpp
```diff
@@ -71,6 +70,35 @@ void writeFlags(FILE* fp, Symbol* sym) {
   }
 }
 
+static const char* qualDebugStr(Qualifier q) {
```
ronawho: Can this share code with `type::qualStr()`?

benharsh:
The QualifiedType changes look fine to me.

PR #6876, opened 2017-07-28
Significantly reduce nondeterminism in the compiler

benharsh:
This PR implements templated functions that hash BaseAST-inheriting classes based on their AST IDs. For a long time we have hashed on the pointer value which can change from run to run, which made debugging the compiler difficult due to differing AST IDs.

To test the effectiveness of this PR I compiled HPCC stream and saved the generated code with --gen-ids as a reference point. I then compiled with the same flags roughly ten times. On master there were on average 80,000 lines different between the reference and a given run's generated code. With this PR, there are no differences run to run.

For our home-grown Map and Vec classes, I have implemented a templated function called ``_vec_hasher``. I have left the fully generic version of this function undefined, forcing us to consider what the hash of an object should be lest we mistakenly use pointer values in the future.

For c++ standard containers I have implemented the ``less`` comparator, which compares AST IDs.

I have also modified ``Map:: get_values`` such that it will no longer return a Vec of unique values. This method seemed to only be used for virtual methods and children, where the values were pointers to a Vec<FnSymbol*>. Instead of trying to figure out what a good hash of a Vec would be, I observed that we do not re-use values across different keys. This means that ``Map::get_values`` need not attempt to create a Vec of unique values.

Testing:
- [x] full local
- [x] full no-local
- [x] full gasnet

### DIFF
--- a/compiler/include/baseAST.h
+++ b/compiler/include/baseAST.h
@@ -415,6 +415,86 @@ def_to_ast(ParamForLoop);
 
 #undef def_to_ast
 
+#define def_hash_ast(SomeType) \
+  namespace std { \
+    template<> struct less<SomeType*> { \
+      bool operator()(const SomeType* lhs, const SomeType* rhs) const { \
+        if (lhs == NULL && rhs != NULL) return true; \
+        if (lhs != NULL && rhs == NULL) return false; \
+        if (lhs == NULL && rhs == NULL) return false; \
+        return ((BaseAST*)lhs)->id < ((BaseAST*)rhs)->id; \
+      } \
+    }; \
+  }
+
+def_hash_ast(SymExpr)
+def_hash_ast(UnresolvedSymExpr)
+def_hash_ast(DefExpr)
+def_hash_ast(ContextCallExpr)
+def_hash_ast(ForallExpr)
+def_hash_ast(NamedExpr)
+def_hash_ast(UseStmt)
+def_hash_ast(BlockStmt)
+def_hash_ast(CondStmt)
+def_hash_ast(GotoStmt)
+def_hash_ast(DeferStmt)
+def_hash_ast(ForallStmt)
+def_hash_ast(TryStmt)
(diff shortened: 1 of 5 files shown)


mppf:
I took a quick look at this and don't have any objections. I can see how this solves the problem for Vec/Map. My main question is if there's any corresponding strategy change to std::map or std::set that we ought to be using? Or does this PR cover that case too somehow? (Or had we already handled it?)

benharsh:
std::map and std::set should use the ``less`` comparator by default. This PR implements specialized instantiations of this comparator for BaseAST-inheriting classes.

compiler/include/baseAST.h
```diff
@@ -415,6 +427,58 @@ def_to_ast(ParamForLoop);
...
+      bool operator()(const SomeType* lhs, const SomeType* rhs) const { \
+        if (lhs == NULL && rhs != NULL) return true; \
+        if (lhs != NULL && rhs == NULL) return false; \
+        if (lhs == NULL && rhs == NULL) return false; \
+        return ((BaseAST*)lhs)->id < ((BaseAST*)rhs)->id; \
+      } \
+    }; \
+  }
+
+def_hash_ast(SymExpr)
```
lydia-duncan: Out of curiosity, is there a reason defining this for BaseAST would not be sufficient for all subtypes?

lydia-duncan: Oh, right, to avoid having a version we default to.  Seems painful to add one for every node if this is the behavior we always want, though

lydia-duncan:
I'm fine with this going in as is, but I suspect we will never want the pointer comparison again and so defining it on BaseAST (or some way to apply it to every node type by default) would be less of a maintenance burden

benharsh:
I agree, but I don't know how to do that (at least without C++11 or later features).

vasslitvinov:
Thanks to Ben for working it out.
def_hash_ast on every node is a reasonable price to pay IMO.

Fine to merge as-is.

PR #6877, opened 2017-07-28
Add Symbol::getInitialization(), use it in parallel.cpp

mppf:
In this first commit, parallel.cpp computes the initialization point two
ways - the old way and also the new way. The intent is to verify that the
new code behaves the same.

In a future commit, we'll need to remove the old implementation.

Passed quickstart+gasnet testing.
Passed qthreads+gasnet testing.
Passed full local testing.

### DIFF
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -418,13 +418,11 @@ Expr* Symbol::getInitialization() const {
       }
     }
 
-    if (isDef || isUse) {
+    if (isDef) {
+      // first use/def of the variable is a def (normal case)
+      return stmt->getStmtExpr();
 
-      if (isDef) {
-        // first use/def of the variable is a def (normal case)
-        return stmt->getStmtExpr();
-      }
-
+    } else if (isUse) {
       bool handled = false;
 
       // handle PRIM_MOVE refTmp, PRIM_ADDR_OF curSym or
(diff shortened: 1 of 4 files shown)


mppf:
@benharsh - here is my attempt at solving the const-record.chpl issue we were talking about. I plan to do this in multiple PRs so we can get full testing that the new solution is as good as the old one. This PR does it both ways and then checks it got the same result.

compiler/AST/symbol.cpp
```diff
@@ -379,6 +379,84 @@ SymExpr* Symbol::getSingleDef() const {
...
+      if (sym == curSym || sym == refSym) {
+        int result = isDefAndOrUse(se);
+        isDef |= (result & 1);
+        isUse |= (result & 2);
+      }
+    }
+
+    if (isDef || isUse) {
+
+      if (isDef) {
```
benharsh: Could you hoist this out of the parent if-statement, and instead structure the code like this?
```cpp
if (isDef) {
  // return ...
} else if (isUse) {
  // ...
}
```

compiler/passes/parallel.cpp
```diff
@@ -911,11 +911,17 @@ replicateGlobalRecordWrappedVars(DefExpr *def) {
...
-  if (found)
+
+  Expr* initialization = def->sym->getInitialization();
+
+  if (found) {
+    INT_ASSERT(stmt == initialization);
     stmt->insertAfter(new CallExpr(PRIM_PRIVATE_BROADCAST, def->sym));
-  else
+  } else {
+    INT_ASSERT(0);
```
benharsh: Could this be an ``INT_FATAL`` with a helpful error message for developers?

PR #4911, opened 2016-11-29
Fix third-party gasnet package dependence flaw.

gbtitus:
The 'build' target needs to depend on the 'config' target, and the
'install' target on 'build'.  Otherwise, we can get ahead of ourselves
in parallel makes.  However, with those dependences added we can
re-enable parallelism.

PR #6878, opened 2017-07-28
Re-enable parallel make support for GASNet

ronawho:
GASNet 1.28.2 fixed support for parallel make, so stop forcing serial make.
This effectively reverts #4911


ronawho:
@gbtitus https://upc-bugs.lbl.gov/bugzilla/show_bug.cgi?id=3389 has been fixed, and I've been running a parallel make with 1.28.2 with this patch for a couple of hours without issue (1.28.0 failed 1/3 times with this patch applied):

Here's what I've been running to test on chapelmac:
```bash
#!/usr/bin/env bash

for i in $(seq 1000); do
  git clean -fdx third-party/gasnet
  make -C third-party/gasnet -j || exit 1
done

```

PR #6884, opened 2017-07-29
Quiet forVersusWhilePerf regressions

ronawho:
PR #6833 changed forVersusWhilePerf.c to be more comparable to the Chapel
version. It changed some int's to int64_t's, but used the wrong printf
formatter and didn't include stdint.h.

Here I include stdint.h and cast the results in printf to long ints. I could
have used the PRId64 formatter, but I never remember how portable that is, and
we don't use it in any other tests, so I opted to just cast in order to quiet
the regression for now.

### DIFF
--- a/test/statements/lydia/forVersusWhilePerf.c
+++ b/test/statements/lydia/forVersusWhilePerf.c
@@ -2,6 +2,7 @@
 #include <sys/time.h>
 #include <time.h>
 #include <stdlib.h>
+#include <stdint.h>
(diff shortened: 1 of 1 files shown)


ronawho:
just fyi @coodie and @lydia-duncan

lydia-duncan:
Elliot, I'm not seeing the regressions you are referring to, either in cron mails or in our job UI.  Can you point me to them?

lydia-duncan:
Oh, was it the cygwin failure?

PR #6900, opened 2017-07-31
Copy constructor for Toml class

Spartee:
This PR adds a copy constructor for the Toml library. This is very useful in the creation of .lock files and therefore is needed in the TOML library for Mason. Also added is a corner case for arrays that should have been included in the first pr. 

@ben-albrecht

### DIFF
--- a/test/modules/packages/Toml/TOML.chpl
+++ b/test/modules/packages/Toml/TOML.chpl
@@ -434,6 +434,20 @@ Parser module with the Toml class for the Chapel TOML library.
        this.tag = fieldArr;
      }
 
+     // Clone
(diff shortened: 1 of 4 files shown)


test/modules/packages/Toml/TOML.chpl
```diff
@@ -434,6 +434,20 @@ Parser module with the Toml class for the Chapel TOML library.
        this.tag = fieldArr;
      }
 
+     // Clone
+     proc init(root: Toml) {
```
ben-albrecht: I think this deserves a test

Spartee: On it

PR #6903, opened 2017-08-01
Adjust call destuctors to correctly deinit in try blocks

mppf:
See the new err-return test for an example where the new behavior is relevant.

callDestructors needs to be aware of error hadling because it has both control flow impact and also because it has an impact on where deinitialization should occur. In particular variables local to a try block should be deinitialized before the catch blocks execute.  Additionally, calls that throw in a try block should deinitialize locals in that try block if an error occurred - with the exception of a local that was not fully initialized (as in the case when a function returning a record throws an error).

This PR takes the approach of making the AST added for error-handling simpler and more amenable to special treatment in callDestructors. To that end, it adds a flag and a new primitive.  Then, it adjusts callDestructors to recognize and take particular action at the error-handling label in a try block and in the error-handling condititional that occurs after a call to a throwing function. The case that proved to be the most complex to implement is that a throwing function that returns should not have the return value be deinitialized if the function threw an error. To make that happen, such variables are identified in addAutoDestroyCalls and then passed along as variables to ignore to the subsequent traversal that will add auto-destroy calls before a goto-error-handler.

- [x] Passed test/errhandling in quickstart, quickstart+gasnet.
- [x] Passed full local testing.

Reviewed by @psahabu  and discussed with @noakesmichael - thanks!


mppf:
@psahabu - could you have a look at this PR? thanks.

PR #6906, opened 2017-08-01
Suppress unneeded warning from CCE 8.6 - floating point values

awallace-cray:
Replaces PR #6836


make/compiler/Makefile.cray-prgenv-cray
```diff
@@ -25,6 +25,9 @@ CRAYPE_GEN_CFLAGS += -hnomessage=186
 # certain cases of __asm__
 CRAYPE_GEN_CFLAGS += -hnomessage=3140
 
+# Suppress warnings about floating-point value cannot be represented exactly.
```
ronawho: Read's a little weird, and is the only one with a period. 

I'd suggest `Suppress warnings about floating-points that cannot be represented exactly`

PR #6907, opened 2017-08-01
LayoutCS

ben-albrecht:
This PR adds `LayoutCS`, a general compressed sparse layout that supports both `CSR` and `CSC`.


ben-albrecht:
This PR depends on a type-alias modification in the compiler (thanks @noakesmichael && @benharsh) - Will reopen when that PR is opened & merged.

PR #6909, opened 2017-08-01
Future showing issue with non-generic copy-init

mppf:
I ran in to this issue while working on some error-handling test cases.

I thought that we had concrete copy-init working but it seems not to be the case.

I think this is related to PR #6103.


lydia-duncan:
Likely related to #6564, right?

mppf:
@lydia-duncan - yes but I think the PR I referenced at the outset effectively disabled populating the auto-copy map for non-generic records.

lydia-duncan:
Ah, true.

PR #6910, opened 2017-08-01
Stop disabling GASNet's pshm support for OSX

ronawho:
pshm support for mac was fixed with GASNet 1.28.2, so stop disabling it.


ronawho:
@gbtitus any objections? 

Added gasnet-fast testing for OSX with https://github.com/chapel-lang/chapel/pull/6881 to ensure we'd have a config to test this.

PR #6913, opened 2017-08-01
Set break and continue labels of LoopStmts to NULL if removed

benharsh:
Without this change, LoopStmt would store dangling pointers and cause valgrind failures when hashing the dead LabelSymbols (by dereferencing "id").

Testing:
- [x] full local


vasslitvinov:
Looks good to merge, if testing passes.

benharsh:
I've decided against doing full valgrind testing due to the amount of time required. I have checked a handful of tests that were failing, and they no longer fail. I'll let nightly testing catch further issues.

PR #6916, opened 2017-08-02
Remove some copy propagation futures

benharsh:
These were all fixed by #6779.


cassella:
Some of these ex-futures still have ``.skipif``s to skip conditions
where the past future passed:

```
fortytwo@magrathea:~/src/chapel (master)$ cat test/functions/deitz/nested/test_nested_var_iterator3.skipif 
# This bug does not show up if inlining is turned off.
COMPOPTS <= --baseline
COMPOPTS <= --no-inline
```

and

```
fortytwo@magrathea:~/src/chapel (master)$ cat test/expressions/ferguson/if-expr/ifi.skipif 
# skip this version in baseline testing so it fails
COMPOPTS <= --baseline
```

The ifi.future file used to say

```
If the bug is fixed, besides removing this future, remove
the ifi-baseline test and ifi.skipif.
```

benharsh:
Good catch, I'll remove those as well.

PR #6918, opened 2017-08-02
Initial AoS elegant implementation of CoMD

benharsh:
This version differs from the LLNL implementation by using a domain map,
AccumStencilDist, to represent its distributed arrays. This domain map
behaves like the standard StencilDist, except that its updateFluff will
accumulate cached cells onto their source cells. A += overload is
required for array elements for this distribution to work.

This implementation is largely based off of the reference MPI+openmp
implementation.

TODOs:
- get rid of "localAccess" calls
- this would be a lot cleaner if we could update the boundingBox after
  initialization
- performance

Testing:
- [x] test ML perf files


ben-albrecht:
Do we have any perf comparisons yet?

benharsh:
Not yet. After discussions with @bradcray we felt it would be better to stop at correctness and shift to other priorities (e.g. sparse matrices).

ben-albrecht:
Would still like to do a post-merge review when time permits..

PR #6926, opened 2017-08-03
Generate fast flags for floating operations in llvm build

coodie:
This change is not necessary to make fast flag optimizations, however none fast-flags optimizations are visible in `chpl__module.bc` nor using `llvm-print-ir` because they are performed at clang invocation time. This change is mostly for developer convenience, but it is also related in http://releases.llvm.org/3.7.0/docs/Frontend/PerformanceTips.html#other-things-to-consider.


compiler/util/clangUtil.cpp
```diff
@@ -1191,6 +1191,17 @@ void prepareCodegenLLVM()
...
+
+  if(ffloatOpt == 1)
+  {
+    llvm::FastMathFlags FM;
+    FM.setNoNaNs();
+    FM.setNoInfs();
+    FM.setNoSignedZeros();
+    FM.setAllowReciprocal();
+    FM.setUnsafeAlgebra();
+    info->builder->SetFastMathFlags(FM);
```
mppf: I had to change it to setFastMathFlags for LLVM 4. Could you fix it in the PR too?

mppf:
I've tested this in full local testing, --llvm testing, and --llvm --fast testing.

PR #6934, opened 2017-08-03
Improve annotation checker

ronawho:
Exit with a non-zero exit-code if the annotation checker finds any warnings

The annotation checker is now being run as part the travis smoke test so
warnings won't be caught unless there's a non-zero exit code. The warnings
aren't fatal (won't screw up nightly testing), but they should still be caught
before merging.

Also does some misc cleanup to the script:

  - use docstrings instead of # comments
  - remove outdated comments
  - use get_chpl_home() instead of just looking for $CHPL_HOME
  - trim down some long lines
  - cleanup other misc things suggested by the linter

### DIFF
--- a/util/test/check_annotations.py
+++ b/util/test/check_annotations.py
@@ -21,6 +21,7 @@ import re
 import subprocess
 import sys
 import time
+import warnings
 
 import annotate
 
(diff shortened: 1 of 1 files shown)


util/test/check_annotations.py
```diff
@@ -1,101 +1,109 @@
 #!/usr/bin/env python
+""" Do basic validation and checks on the ANNOTATIONS.yaml file
 
-import annotate
+This is a kinda-ugly script that does basic validation of ANNOTATIONS.yaml
```
ben-albrecht: `s/kinda-ugly/pretty/`

PR #6936, opened 2017-08-04
LayoutCS

ben-albrecht:
This PR generalizes `LayoutCSR` to `LayoutCS` that supports both `CSR` (`CS(compressRows=true)`) and `CSC` (`CS(compressRows=false)`) sparse arrays.

This PR also introduces a test demonstrating `CSR-CSC` matrix multiplication.

## Finishing Touches

- [x] Test domain map standard interface (DSI) for `CS`
    - `test/sparse/CS/correctness.chpl`
- [x] Polish naming scheme
-  **backwards-compatibility**: Confirm that all existing `CSR` tests work with `CS(row=true)`
    - [x] `local`
    - [x] `no-local`
    - [x] Deprecation message for `LayoutCSR`
    - [x] Deprecation test for `LayoutCSR`
-  Performance testing
    - [x] Add `CSR . CSC` performance test
    - [x] Add `_grow()` performance test
    - (future PR) compare performance to other libraries such as `MKL Sparse BLAS` and`scipy.sparse`

### LayoutCSR

Since `LayoutCS`'s `type CSR` is fully backwards-compatible with `LayoutCSR` (aside from the `config const debugCSR`), `LayoutCSR` will be deprecated and temporarily replaced with the following contents:

```chpl
// LayoutCSR.chpl

use LayoutCS;
```

### Printing

Currently, `CSC` arrays print how they are stored, by columns rather the rows. This is more performant and simpler to implement, but looks odd when trying to visually compare the output of a `CSR` array to a `CSC` array.

**Example:**

Sample array:
```
11 12 13 14  0  0
 0 22 23  0  0  0
 0  0 33 34 35 36
 0  0  0 44 45  0
 0  0  0  0  0 56
 0  0  0  0  0 66
```

Printed as `CSR`:
```
{
 (1, 1) (1, 2) (1, 3) (1, 4)
 (2, 2) (2, 3)
 (3, 3) (3, 4) (3, 5) (3, 6)
 (4, 4) (4, 5)
 (5, 6)
 (6, 6)
}
```
Printed as `CSC`:
```
{
 (1, 1)
 (1, 2) (2, 2)
 (1, 3) (2, 3) (3, 3)
 (1, 4) (3, 4) (4, 4)
 (3, 5) (4, 5)
 (3, 6) (5, 6) (6, 6)
}
```

### row & column major semantics

The following overload assumes that the indices are sorted because domains are sorted by default. However, it's possible that a column-major domain could be sorted by the column index rather than the row index, making this assumption wrong. 

```chpl
proc +=(ref sd: domain, d: domain) where isSparseDom(sd) && ...
```

In this PR, we lose an optimization for bulk-addition of a domain to a sparse domain, since we can no longer assume the indices are sorted. It would be nice to have a way to query whether a domain is sorted by column or row indices for cases like this. We can re-enable this optimization once this is in place.

*TODO:* Open an issue for this feature request.


ben-albrecht:
**Reviewers:** Many trailing whitespaces were harmed in the making of this pull request, therefore this diff is best viewed with `?w=1`:

https://github.com/chapel-lang/chapel/pull/6936/files?w=1

modules/internal/ChapelDistribution.chpl
```diff
@@ -453,13 +472,13 @@ module ChapelDistribution {
...
-    proc bulkAdd_prepareInds(inds, dataSorted, isUnique) {
+    proc bulkAdd_prepareInds(inds, dataSorted, isUnique, cmp) {
       use Sort;
-      if !dataSorted then sort(inds);
+      if !dataSorted then sort(inds, comparator=cmp);
 
       //verify sorted and no duplicates if not --fast
       if boundsChecking {
-        if dataSorted && !isSorted(inds) then
+        if dataSorted && !isSorted(inds, comparator=defaultComparator) then
```
benharsh: Why not use ``cmp`` here?

modules/internal/ChapelDistribution.chpl
```diff
@@ -607,8 +629,13 @@ module ChapelDistribution {
...
 
   // Currently this is not optimized for addition of a sparse
+  // This assumes that a domain is sorted by row, which may not always be true.
+  // There could be domains sorted by column instead.
+  // It would be nice if there was a common interface in BaseDom to query
+  // whether a domain was row-major or column-major to distinguish these.
+  // The current implementation is a work-around for this shortcoming.
   proc +=(ref sd: domain, d: domain)
-  where isSparseDom(sd) && d.rank==sd.rank && sd.idxType==d.idxType {
+  where sd._value: DefaultSparseDom && d.rank==sd.rank && sd.idxType==d.idxType {
```
benharsh: If we're going to leave ``+=`` unimplemented for CSDom, could we at least add an overload with a helpful ``compilerError`` ?

modules/internal/ChapelDistribution.chpl
```diff
@@ -621,8 +648,41 @@ module ChapelDistribution {
 
     sd._value.dsiBulkAdd(arr, true, true, false);
   }
+
+  // TODO: Implement bulkRemove
+  proc -=(ref sd: domain, inds: [] sd.idxType) where isSparseDom(sd) &&
+    sd.rank==1 {
+
+    if inds.size == 0 then return;
```
benharsh: I don't think these returns are necessary. If the arrays or domains are empty, the loops won't do anything.

modules/internal/ChapelDistribution.chpl
```diff
@@ -621,8 +648,41 @@ module ChapelDistribution {
 
     sd._value.dsiBulkAdd(arr, true, true, false);
   }
+
+  // TODO: Implement bulkRemove
+  proc -=(ref sd: domain, inds: [] sd.idxType) where isSparseDom(sd) &&
```
benharsh: These overloads seem better suited to ``ChapelArray.chpl``

test/sparse/CS/correctness.chpl
```diff
@@ -0,0 +1,164 @@
...
+
+  writeln('DSI');
+  writeln('===');
+
+  // dsiBulkAdd with domain
+  csrDom += D;
+  cscDom += D;
+
+  assert(csrDom.size == D.size);
+  assert(cscDom.size == D.size);
```
benharsh: Maybe also assert that ``csrDom == cscDom`` ?

test/sparse/CS/multiplication/correctness.chpl
```diff
@@ -0,0 +1,161 @@
...
+  var csrArr: [csrDom] real,
+      cscArr: [cscDom] real;
+
+  var A: [ADom] real,
+      B: [BDom] real;
+
+  const rA1 = ADom.dim(1) by 2,
+        rA2 = ADom.dim(2) by 2;
+
+  csrDom += {rA1, rA2};
```
benharsh: Can't this just be ``ADom by 2`` ?

test/sparse/CS/multiplication/correctness.chpl
```diff
@@ -0,0 +1,161 @@
...
+    writeln('csrArr x cscArr:');
+    writeDense(csrcscArr);
+  }
+
+  // TODO: Validate sparse csrcscArr against AB
+}
+
+/* Dense matrix-matrix multiplication */
+proc multiply(A: [?ADom] ?eltType, B: [?BDom] eltType) {
+  var C: [{ADom.dim(1), BDom.dim(2)}] eltType;
```
benharsh: I don't think you need the curly braces here.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,657 @@
...
+on the locale where the array variable is declared.
+*/
+class CS: BaseDist {
+  const row: bool = true;
+
+  proc dsiNewSparseDom(param rank: int, type idxType, dom: domain) {
+    return new CSDom(rank=rank, row=row, idxType=idxType, dist=this, parentDom=dom);
+  }
+
+  proc dsiClone() return new CS();
```
benharsh: Shouldn't ``dsiClone`` return ``new CS(this.row)`` ?

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,657 @@
...
+    rowRange = parentDom.dim(1);
+    colRange = parentDom.dim(2);
+    dom = if row then {rowRange.low..rowRange.high+1} else {colRange.low..colRange.high+1};
+    nnzDom = {1..nnz};
+    dsiClear();
+  }
+
+  proc dsiMyDist() return dist;
+
+  proc dsiGetIndices() return 0;
```
benharsh: If we're not supporting get/setIndices, could we at least add some kind of error/warning message?

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,657 @@
...
+    else
+      coforall chunk in chunks(1..numElems, numChunks) do
+        yield (this, chunk.first, chunk.last);
+    // TODO: to handle large numElems and numChunks faster, it would be great
+    // to run the binary search in _private_findStart smarter, e.g.
+    // pass to the tasks created in 'coforall' smaller ranges to search over.
+  }
+
+  // TODO: Test
+  iter these(param tag: iterKind, followThis: (?,?,?)) where tag == iterKind.follower {
```
benharsh: Have you tested zipping a CSR with a CSC?

modules/internal/ChapelDistribution.chpl
ben-albrecht: Perhaps my comment was not clear enough. We *do* support `+=` on `CSDom`, but it is an awkward work-around (implemented at the bottom of `LayoutCS.chpl`)

modules/internal/ChapelDistribution.chpl
ben-albrecht: I pondered this as well, in a comment above:

```
614   // TODO: Should these be in ChapelArray with other overloads?
```

test/sparse/CS/correctness.chpl
ben-albrecht: I think this behavior is a bit of an open question... Would be interested on @bradcray's thoughts.

test/sparse/CS/multiplication/correctness.chpl
ben-albrecht: Yep! `rA1` and `rA2` originally had different strides, but I settled on this for now.

test/sparse/CS/correctness.chpl
bradcray: Is the question whether we could/should support domain equality across sparse domains of different implementations?

benharsh: Yes. My impression of ``==`` between domains is to check if they both have the same indices, regardless of internal representation.

bradcray: I agree.  It would surprise me if the default implementation of `==` on sparse worked (but I'm saying that without taking a look to check).  It won't be particularly efficient, but I think that's acceptable.

modules/layouts/LayoutCS.chpl
ben-albrecht: Good call - that should be tested as well.

modules/layouts/LayoutCS.chpl
ben-albrecht: This originated from `CSR` implementation, but I agree with you. I'll emit a compiler error.

test/sparse/CS/correctness.chpl
ben-albrecht: > My impression of == between domains is to check if they both have the same indices, regardless of internal representation.

I like this definition.

> It would surprise me if the default implementation of == on sparse worked

`csrDom == cscDom` evaluates to `true` with the current implementation.

modules/layouts/LayoutCS.chpl
ben-albrecht: Not yet.. coming in next commit.

test/sparse/CS/correctness.chpl
bradcray: Sheesh, I'm getting old.  That was supposed to read "it wouldn't surprise me if it worked" (or maybe "it would surprise me if it didn't work").  Glad I was right even though the words don't get from my brain to the keyboard correctly.

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+    sd._value.dsiBulkAdd(inds);
+  }
+
+  proc +=(ref sd: domain, inds: [] sd.rank*sd.idxType) where isSparseDom(sd) &&
+    sd.rank>1 {
+
+    if inds.size == 0 then return;
+
+    sd._value.dsiBulkAdd(inds);
+  }
```
bradcray: Can these two overloads be combined into one by replacing the element type on the `inds` arguments with `index(sd)`  ("the index type of `sd`")?

I may be missing something, but it seems like these new `+=` operators were added out of thin air in this PR without explanation (the PR message suggests to me that they are pre-existing, but I can't see any indication of that — has there been support for them in the past that I'm overlooking?).  Was there any design discussion on adding support for them?  Should they be documented?

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+
+  // Currently this is not optimized for addition of a sparse
```
bradcray: This seems incomplete?

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+  // This assumes that a domain is sorted by row, which may not always be true.
+  // There could be domains sorted by column instead.
+  // It would be nice if there was a common interface in BaseDom to query
+  // whether a domain was row-major or column-major to distinguish these.
+  // The current implementation is a work-around for this shortcoming.
+  proc +=(ref sd: domain, d: domain)
+  where sd._value: DefaultSparseDom && d.rank==sd.rank && sd.idxType==d.idxType {
```
bradcray: Is it correct that the effect of this where clause is that we don't support a += overload for CS* layouts on the LHS that take a domain on the RHS?   If so, that seems unfortunate...

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+
+    if d.size == 0 then return;
+
+    type _idxType = if sd.rank==1 then int else sd.rank*int;
+    const indCount = d.numIndices;
+    const arr: [{0..#indCount}] _idxType;
```
bradcray: This is another case where I think we could use `index(sd)` for simplicity in place of computing `_idxType`.  I feel surprised that `_idxType` is defined in terms of `int` rather than the `idxType` of `d` or `sd`.

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+
+    //this could be a parallel loop. but ranks don't match -- doesn't compile
+    for (a,i) in zip(arr,d) do a=i;
```
bradcray: Even if ranks matched, this would be difficult to execute in parallel, since it would require us to be able to know a priori what the unique position of any index 'i' in domain 'd' had (i.e., what index of 'arr' each index in 'd' corresponds with) which is a hard problem in general.  I.e., I'm not sure "could be a parallel loop" is accurate without some sort of pre-computation of how many indices are in any given range of rows.

I felt surprised that we permit rank mismatches in serial loops like this, and am thinking that this might be one of the issues leader-follower 2.0 is supposed to fix.  Since it's already serial, it might be better to zip `d` with `0..` in order to make it future-proof.  (i.e., I'm a little worried that what you've written might break in the future).  That is:

```chapel
for (i,j) in zip(d, 0..) do arr[j] = i;
```

modules/internal/ChapelArray.chpl
```diff
@@ -3211,6 +3211,68 @@ module ChapelArray {
...
+
+    sd._value.dsiBulkAdd(arr, true, true, false);
```
bradcray: If I understand correctly, dsiBulkAdd() says whether or not the `arr` value being passed in is sorted.  But here, if the RHS domain is a CSC, they won't be if I'm understanding your PR's message properly.  So I would've expected this to pass in `false` for `sorted` (in which case I'd also wonder if the where clause restricting the LHS to be default sparse could also be removed safely).

modules/internal/ChapelDistribution.chpl
```diff
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2017 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
```
bradcray: In the future, when doing a big trailing whitespace cleanup, I'd recommend doing it in a PR of its own, merging it as trivial, and then pulling it into your working branch rather than having it clutter up an interesting PR...

modules/internal/ChapelArray.chpl
ben-albrecht: Looks like it...

FYI - this section originated from [ChapelDistribution](https://github.com/chapel-lang/chapel/pull/6936/files/96829f529a582ae347fbd0513a3236a71bfb4d0f?diff=split#diff-7aa1af090d3c4b7704aa8aadc35ccd68L592)

modules/internal/ChapelDistribution.chpl
```diff
@@ -404,25 +404,44 @@ module ChapelDistribution {
...
       halt("Helper function called on the BaseSparseDomImpl");
 
       return -1;
     }
 
-    inline proc _grow(size: int){
+    // TODO: Would ChapelArray.resizeAllocRange() be too expensive?
+    //       - would have to put the conditional outside of the call
+    /* Grow domain if necesary */
+    inline proc _grow(size: int, factor=arrayAsVecGrowthFactor) {
```
bradcray: What's the motivation for adding this factor argument, changing it from `2`?  (esp since it's not overridden anywhere).  Does it result in better performance?

modules/internal/ChapelArray.chpl
bradcray: Oops, missed that these had been moved from ChapelDistribution.chpl, sorry.  This also means that many of my comments are on code that you inherited (but I think should still be cleaned up while we're touching it).

modules/internal/ChapelDistribution.chpl
ben-albrecht: In hindsight, I may have gone overboard and these `_grow()` & `_shrink()` changes may be better fit for a follow-up PR with more thorough performance analysis. I assumed defaulting to what we do for array-resizing would be a more reasonable behavior than a hard-coded factor of `2`.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* Debug flag */
+config param debugCS = false;
+use RangeChunk only ;
+
+/* For backwards-compatibility with LayoutCSR */
+type CSR = CS;
```
bradcray: Is this sufficient to keep existing `LayoutCSR` programs working?  If so, I think we should rewrite all `use LayoutCSR` tests to `use LayoutCS`, rip out the guts of LayoutCSR as you propose in your PR message, add a deprecation warning to that module, and add a test/deprecated/ test to help us remember to remove it after 1.16 goes out.  This could either be done in this PR or as a quick follow-on.

If it's not sufficient, I'd remove this line.

modules/internal/DefaultSparse.chpl
```diff
@@ -239,7 +239,7 @@ module DefaultSparse {
     proc bulkAdd_help(inds: [?indsDom] index(rank, idxType), dataSorted=false,
         isUnique=false){
 
-      bulkAdd_prepareInds(inds, dataSorted, isUnique);
+      bulkAdd_prepareInds(inds, dataSorted, isUnique, Sort.defaultComparator);
```
bradcray: We've clearly had it for a very long time, but today I'm finding myself surprised that we defined '>' and '<' on tuples given that one might have different definitions of what less than means in a row-major vs. column-major order (as one example).  Just curious if you found that choice of a default surprising as well...  (as compared to not providing a default at all).

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+pragma "no doc"
+/* Comparator used for sorting by columns */
+record _ColumnComparator {
+  proc init() { }
+  proc key(idx: _tuple) { return (idx(2), idx(1));}
+}
+
+const _columnComparator: _ColumnComparator;
```
bradcray: Should/could this be private?  No doc'd?

bradcray: Related: If I 'make docs' on this branch, I'm not seeing docs for this file.  I can't recall where this stands:  Do we need to explicitly add it to an index file?

modules/layouts/LayoutCS.chpl
ben-albrecht: > Is this sufficient to keep existing LayoutCSR programs working? 

That's what I've observed based on testing.

> I think we should rewrite all use LayoutCSR tests to use LayoutCS, rip out the guts of LayoutCSR as you propose in your PR message.

I will try this in this PR and do a full paratest once changes to `LayoutCS` converge.

modules/internal/DefaultSparse.chpl
ben-albrecht: > Just curious if you found that choice of a default surprising as well... (as compared to not providing a default at all).

I find the default behavior reasonable and useful.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+/*
+This CS layout provides a Compressed Sparse Row and Compressed Sparse Column
+implementation for Chapel's sparse domains and arrays.
+
+To declare a CS domain, invoke the ``CS`` constructor without arguments
+in a `dmapped` clause. For example:
```
bradcray: This documentation seems specific to CSR and like it should be updated to illustrate the CSR, CSC, and default-CS cases?

modules/internal/ChapelDistribution.chpl
ben-albrecht: Maybe I can accomplish this within this PR actually... @ronawho - know of any good benchmarks I can try to mimic with sparse to test the perf-delta here?

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+    // assume the above declarations
+    var CS_Array: [CS_Domain] real;
+
+This domain map is a layout, i.e. it maps all indices to the current locale.
+All elements of a CS-distributed array are stored
+on the locale where the array variable is declared.
+*/
+class CS: BaseDist {
+  const row: bool = true;
```
bradcray: I'd be tempted to rename this because a variable named 'row' to me suggests that it stores a row (and some declarations later in this file do exactly that).  Possible replacements: 'RMO', 'CSR', 'compressRows', ...?

In our (offline) discussions about optimizing the += routine for CSR vs. CSC, I was imagining that we'd know at compile-time whether a CS layout was CSR or CSC, which would require making this a `param`.  Were there competing reasons for making it a `param`?

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+  proc dsiNewSparseDom(param rank: int, type idxType, dom: domain) {
+    return new CSDom(rank=rank, row=row, idxType=idxType, dist=this, parentDom=dom);
+  }
+
+  proc dsiClone() return new CS(row=this.row);
+
+  proc dsiEqualDMaps(that: CS) param {
+    return true;
```
bradcray: Thus should probably return `true` only if the two domain maps have the same `row` setting?  (that said, I'm not sure how this is used in practice...).

bradcray: Ah... though doing so would require `rows` being a `param` as inquired about above...

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+class CSDom: BaseSparseDomImpl {
+  const row: bool;                  // would like param
+
+  var dist: CS;
+
+  var rowRange: range(idxType);
+  var colRange: range(idxType);
+
+  /* (row|col) dom */
+  const dom: domain(1, idxType);
```
bradcray: This is another variable I'd suggest giving a more descriptive/unique name than simply `dom` (which says to me "this domain is a domain").  What would we want to call this dimension in the general case?  It's the dimension that supports random access and is not compressed...  "raDimDom"?  "idxDimDom"?

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+  /* (row|col) start */
+  var startIdx: [dom] idxType;      // would like index(nnzDom)
+  /* (row|col) idx */
+  var idx: [nnzDom] idxType;        // would like index(parentDom.dim(1))
```
bradcray: This one isn't quite as bad as 'dom', but if we came up with some way of referring to the compressed vs. uncompressed dimensions consistently / conveniently, I'd use that term here too (e.g., "compressedDimIdx" comes to mind, though that's a mouthful).

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+
+  /* Initializer */
+  proc CSDom(param rank, type idxType, const row=true,
```
bradcray: I don't think there's a code path that would ever use the default value of `row` here, so I'd probably remove it for simplicity.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+             dist: CS, parentDom: domain(rank, idxType)) {
+    if (rank != 2) then
+      compilerError("Only 2D sparse domains are supported by the CS distribution");
+    this.row = row;
+    this.dist = dist;
+    this.parentDom = parentDom;
+    rowRange = parentDom.dim(1);
+    colRange = parentDom.dim(2);
+    dom = if row then {rowRange.low..rowRange.high+1} else {colRange.low..colRange.high+1};
```
bradcray: This is something you've inherited, but I'm realizing that if the parent domain was strided, we'd end up storing a dense array here rather than a strided one.  Might be worth adding a TODO comment to acknowledge.  Or we could throw a warning or assertion error in the event that rowRange was strided saying we don't optimize/handle that case yet).

modules/internal/ChapelDistribution.chpl
ronawho: I don't know of any good benchmarks, but our growth factor is based off of facebooks optimized folly /FBVector, which is a much faster and optimized version of std:: vector. 

https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md#the-jemalloc-connection

modules/layouts/LayoutCS.chpl
ben-albrecht: > Or we could throw a warning or assertion error in the event that rowRange was strided

I think this is the right course of action. Sparse domains over strided arrays is currently unsupported / undefined. In fact, sparse domains don't even have a `param stridable` field like all other domains.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+    nnzDom = {1..nnz};
+    dsiClear();
+  }
+
+  proc dsiMyDist() return dist;
+
+  // TODO: Open question -- How should slicing on CS domains and arrays work?
```
bradcray: This comment threw me off because slicing of CS arrays shouldn't rely on these routines and should "just work" using array views (though it very well may not).  Checking the sources, I see you're probably referring to domain-domain slicing, so I'd rewrite this to say "slicing between CS domains" or "domain intersection (slicing)" work?"

Related, I might simply make the errors just below say "dsiGetIndices() is not yet supported..." to be more precise.  E.g., it looks to me as though dsiSetIndices() probably isn't used for sparse domains anymore.

So here's a third approach: Simply pull this comment and these two routines and wait for something to prove it needs them rather than supplying error overloads "just in case".  I'm noticing that DefaultSparse.chpl doesn't support them (anymore?) for example...

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+  proc dsiBuildArray(type eltType)
+    return new CSArr(eltType=eltType, rank=rank, idxType=idxType, dom=this);
+
+  iter dsiIndsIterSafeForRemoving() {
+    var cursor = if this.row then rowRange.high else colRange.high;
+    for i in 1..nnz by -1 {
+      while (startIdx(cursor) > i) {
+        cursor -= 1;
+      }
+      if this.row {
```
bradcray: This line is a great example of why we'd want a param -- without that, all our iterators become "two yield" iterators which we're much worse at optimizing than those with a single yield.  If 'row' is a param, we'll fold this conditional and have a single yield.

If we can't make row a param for some reason, we should at least store the index to be yielded into a variable in a conditional and then yield that variable to get back to a single-yield world.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+        return (i-1, idx[idx.domain.low]);
+
+    halt("Something went wrong in dsiFirst");
+    return (0, 0);
+  }
+
+  proc dsiLast {
+    if nnz == 0 then return (parentDom.low) - (1,1);
+    const _low = nnzDom.low;
+    var _lastRow = parentDom.low[1] - 1;
```
bradcray: This looks like it's still suspiciously row-oriented.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+    // increment number of nonzeroes
+    nnz += 1;
+
+    // double nnzDom if we've outgrown it; grab current size otherwise
+    var oldNNZDomSize = nnzDom.size;
+    _grow(nnz);
+
+    const (row,col) = ind;
+
+    // shift column indices up
```
bradcray: Minor, but this comment is still very CSR-oriented.

modules/internal/ChapelArray.chpl
ben-albrecht: (discussed offline) Disabling this optimization for this PR - we will reintroduce with an `CMO vs. RMO` domain interface addition

modules/internal/ChapelArray.chpl
ben-albrecht: see [other comment](https://github.com/chapel-lang/chapel/pull/6936#discussion_r132338710)

modules/internal/ChapelDistribution.chpl
ben-albrecht: Noted! Maybe `?w=1` can suffice for now.

modules/layouts/LayoutCS.chpl
ben-albrecht: That is correct. Coming in next commit.

modules/layouts/LayoutCS.chpl
ben-albrecht: >  Were there competing reasons for making it a param?

Only technical difficulties which have been overcome in the upcoming commit.

I like `compressRows` out of the options available.

ben-albrecht: Well, users will have to interact with this argument in the form:

```chpl

var subD: sparse subdomain D dmapped CS(compressRows=true);

```

That's pretty lengthy...

```chpl
CS(CSR=true)
```

is also a little awkward..

ben-albrecht: Going with `compressRows` for now...

modules/layouts/LayoutCS.chpl
ben-albrecht: Option 3 sounds nice.

modules/layouts/LayoutCS.chpl
ben-albrecht: Good catch. Same with `dsiFirst`. Unfortunately this method does not produce the correct result on master today, so I need to do some debugging. Working on a fix..

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+          newIndIdx -= 1;
+          if newIndIdx == indsDom.low-1 then break; // There were duplicates -- now done
+          newLoc = actualInsertPts[newIndIdx];
+        }
+      }
+      else halt("Something went wrong");
+    }
+
+    // Aggregated row || col shift
+    var prevCursor = if this.row then parentDom.dim(1).low else parentDom.dim(2).low;
```
bradcray: It seems as though, if we set up constants like 'compressedDim' and 'indexedDim' (to pick some potentially lame names) that store 1 and 2 (or 2 and 1) depending on whether we were doing CSR or CSC, a bunch of code like this could be simplified from using conditionals to simpler expressions like 'var prevCursor = parentDom.dim(compressedDim)'.

modules/layouts/LayoutCS.chpl
```diff
@@ -0,0 +1,656 @@
...
+    return 1;
+  }
+
+  proc dsiClear() {
+    nnz = 0;
+    startIdx = 1;
+  }
+
+  iter dimIter(param d, ind) {
+    if (d != 2) {
```
bradcray: This should probably be '2' for CSR, '1' for CSC.

modules/layouts/LayoutCS.chpl
bradcray: (This would also have the benefit of giving symbolic names that are otherwise magic values of a sort in both CSR and CS).

test/sparse/CS/correctness.chpl
```diff
@@ -0,0 +1,217 @@
+use LayoutCS;
+use Random;
+
```
bradcray: This is a nice test!

modules/layouts/LayoutCS.chpl
ben-albrecht: Going with `startIdxDom` for now...

ben-albrecht:
Currently looking into the last remaining test failure: `test/users/engin/partial_reduction_support/unit_tests/sparsePartialThese.chpl`, which relies on some custom DSI methods in `test/users/engin/partial_reduction_support/modules/dsiMethods.chpl`

PR #6938, opened 2017-08-04
Remove now-redundant ifi-baseline test

cassella:
ifi.chpl used to be a future that passed with --baseline.  So ifi.chpl
had a skipif baseline.  ifi-baseline.chpl was a copy of ifi.chpl that
ran with baseline to lock in it passing.

Now that ifi.chpl passes with and without --baseline and is no longer
skipped with --baseline, ifi-baseline.chpl is redundant with it.


cassella:
@benharsh ifi.future also said to remove this test upon defuturification.

benharsh:
D'oh, thanks!

PR #6939, opened 2017-08-04
Fix typo in GASNET_LFLAGS

ronawho:
Add a missing "D" to GASNET_LFLAGS:

    GASNET_LFLAGS => GASNET_LDFLAGS
    CHPL_GASNET_LFLAGS => CHPL_GASNET_LDFLAGS

Because of the typo we weren't actually adding gasnet's LD flags. Typo was 
noticed by the GASNet team.


ronawho:
@gbtitus and maybe @bradcray any objections?

Tested with gasnet-everything and gasnet-fast on linux64 and OSX as well as gasnet-aries

bradcray:
D'oh!

PHHargrove:
Just a note to help anybody who finds this issue due to GASNET_LDFLAGS problems.

I think the fix in this PR is correct, but it turns out insufficient to fix the originally observed failures that lead us to report the typos.  If I build under conditions where GASNET_LDFLAGS contains necessary flags (such as `-Wl,-rpath=/usr/local/pkg/gcc/7.1.0/lib64`) I find these are *still* not getting used.

As I have just reported in #7010  the `ifeq ($(CC),$(GASNET_CC))` that guards the change made in third-party/gasnet/Makefile.setup is (unintentionally) false under normal build conditions.  So, GASNET_LDFLAGS are still not getting used to link.

PR #6940, opened 2017-08-04
Make this future a normal test since it now passes

mppf:
statements/defer/defer-manual-container-throws now passes.
Trivial testing adjustment.

Verified test passes in quickstart, gasnet configurations.


mppf:
I believe this passed due to #6903.

PR #6941, opened 2017-08-04
Fix valgrind failures by using autoCopyMap::find to test for membership

benharsh:
Directly indexing with operator[] will create an entry with Type* as a
key. If that Type* is later deleted, future accesses of autoCopyMap will
compare against that key with std::less. We define std::less for various
BaseASTs with the def_less_ast macro, which will query the 'id' field
of the given arguments. If an argument was deleted, this will result in
valgrind failures.

Testing:
- [x] full local
- [x] fixes last night's failed tests


benharsh:
I feel pretty confident about this change, so I will merge without a review.

Issue #4982, opened 2016-12-08
printchplenv should infer CHPL_HOME similar to chpl binary


ben-albrecht:
### Summary of Problem

The `chpl` binary can infer the path of `CHPL_HOME` if it is not set based on a variety of checks defined in `compiler/driver.cpp`. `printchplenv` does not have the equivalent checks, and therefore can lead to disagreement between the two, when `CHPL_HOME` is being inferred. 

### Proposed Solution

`printchplenv` should use the same rules that the compiler uses to infer `CHPL_HOME` when it is not set.

### Example

```bash
unset CHPL_HOME

# CHPL_HOME is inferred using the rules defined in compiler/driver.cpp:
>chpl --print-chpl-home
/path/to/CHPL_HOME /path/to/chpl


# CHPL_HOME is empty, and confusingly labeled with a '*'
# which is intended to mean the default value has been overridden (which it hasn't in this case):
>printchplenv --anonymize
machine info: ...
CHPL_HOME:  *
script location: /path/to/CHPL_HOME/util/
CHPL_TARGET_PLATFORM: darwin
...
```

bradcray:
In the interest of coming up with common understandings of labels: this seems more like a feature request (or enhancement) than a bug to me...?

PR #6946, opened 2017-08-07
Have printchplenv infer CHPL_HOME

ronawho:
Infer CHPL_HOME in printchplenv and only print "*" if CHPL_HOME is explicitly
set.

Resolves https://github.com/chapel-lang/chapel/issues/4982

### DIFF
--- a/util/chplenv/chpl_home_utils.py
+++ b/util/chplenv/chpl_home_utils.py
@@ -18,6 +18,10 @@ def get_chpl_home():
     return chpl_home
 
 @memoize
+def get_chpl_module_home():
+    return os.environ.get('CHPL_MODULE_HOME', '')
+
+@memoize
 def get_chpl_third_party():
     default = os.path.join(get_chpl_home(), 'third-party')
     chpl_third_party = overrides.get('CHPL_THIRD_PARTY', default)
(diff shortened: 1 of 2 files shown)


util/chplenv/chpl_home_utils.py
```diff
@@ -18,15 +18,20 @@ def get_chpl_home():
...
+@memoize
 def get_chpl_third_party():
     default = os.path.join(get_chpl_home(), 'third-party')
     chpl_third_party = overrides.get('CHPL_THIRD_PARTY', default)
     return chpl_third_party
 
 @memoize
 def using_chapel_module():
-    chpl_home = overrides.get('CHPL_HOME', '')
+    chpl_home = os.path.normpath(overrides.get('CHPL_HOME', ''))
```
ben-albrecht: Useful! Should `overrides.get()` just normalize all paths before returning?

ben-albrecht:
Post-approval afterthought: Are there any tests against `chpl`'s `CHPL_HOME` inferred values? If so, it'd be nice to confirm `printchplenv` matches what `driver.cpp` does.

util/chplenv/chpl_home_utils.py
ronawho: Nah, CHPL_HOME is the only path that overrides.get() supports

We could do the normpath in get_chpl_home() and get_chpl_module_home(), but I don't really think that's needed. It's not really needed here either since the module is what sets CHPL_HOME and CHPL_MODULE_HOME, but I happened to notice this while testing the change

ronawho:
>Post-approval afterthought: Are there any tests against `chpl`'s `CHPL_HOME` inferred values? If so, it'd be nice to confirm `printchplenv` matches what `driver.cpp` does.

Nightly (and all our testing) always sets CHPL_HOME, so I can't think of a sane way to test this

ben-albrecht:
> Nightly (and all our testing) always sets CHPL_HOME, so I can't think of a sane way to test this

Yeah.. that's a tough one. I'm fine with punting on it for now. It would be good to hand-test a few scenarios that `driver.cpp` covers if possible.

ronawho:
Yup, hand tested a few already

PR #6947, opened 2017-08-07
With comm=ugni, register large arrays dynamically.

gbtitus:
Under uGNI we used to allocate and register our entire heap with the NIC
at program startup and we'd satisfy dynamic allocation using this heap.
But registering memory touches it, so this would set the affinity of the
entire heap to NUMA domain 0.  This meant that we were getting terrible
performance for benchmarks that are NUMA sensitive like stream and the
stencil PRK.

This is the start of dynamic memory registration work.  For now, we're
just dynamically registering arrays.  Essentially we allocate large
arrays directly using hugepages, let the Chapel initializer run, and
then register the array memory. This way Chapel code is responsible for
first-touch and we get the NUMA locality we want.

This has a dramatic performance impact on NUMA-sensitive applications:

~2.5X improvement for stream (now on par with gn-mpi)
~2.5X improvement for optimized stencil (close to gn-mpi, remaining gap
      being investigated by @benharsh and @ronawho)
~2.5X improvement for single 1D reduction
~15% improvement for ISx

There are recommendations for follow-up work in https://github.com/chapel-lang/chapel/issues/6949.

The technical details are as follows.

This is a preliminary implementation of non-heap allocation and dynamic
NIC registration for arrays.  It is used in the `comm=ugni` configuration
when there is a `craype-hugepage` module loaded.  Currently it only
applies to arrays that are at least 2 hugepages in size.  The major
benefits of this feature is that it allows first-touch to achieve good
NUMA locality for such arrays.  It also means that the registered heap
can be much smaller by default, since large arrays are not allocated
from it.  The major downside is increased overhead for both array
allocations and dynamic registration broadcasting.

This feature improves stream and stream-ep performance with `comm=ugni`
and a registered heap by somewhat more than 2x, bringing it roughly up
to par with `comm=gasnet` and the MPI substrate, our best-performing
configuration for these tests.

A more detailed breakdown of the work is as follows.

In the runtime comm layers:
Add new `chpl_comm_regMem*()` functions to work with comm layer registered
memory: threshold size, allocate, post-allocate (register), and free.
Arrange for these to be an optional part of the interface, and implement
them only for `comm=ugni`.  Note that the implementation is a somewhat
preliminary one.  In particular, the broadcasts of updated memory maps
when arrays are registered and deregistered on one node are done by
means of serial `executeOn` operations to each of the other nodes to
update their map entries for that nodes' memory regions.

While here, in the `comm=ugni` alignment macros, debugging, and internal
statistics reporting, I did some other feature work, clean-ups, and
simplifications that arose out of developing and debugging this feature.

In the runtime comm layer pre-launch code:
For this feature to be effective we have to set `HUGETLB_NO_RESERVE=yes`
in the environment, to work around a kernel bug that cause hugepage
locality to be set at reservation time, before first-touch occurs.  We
expect this kernel bug to be addressed soon, but in the meantime we
perform this workaround in the Chapel launcher.  (It would be good to
eventually remove the workaround, because it has the side effect of
causing programs to get a `SIGBUS` on first-touch if they over-reserve the
hugepages, rather than simply getting a synchronous allocation failure
that we can report with the usual "no-more-memory" message.)

In the runtime memory layers:
Update `chpl_mem_array_alloc()` and its wide partner to try to allocate
through the comm layer if the array is big enough.  If successful, tell
the caller to call back again after initializing, passing back the
allocated pointer.  At that point, call the comm layer's post-allocate
routine.  (This back and forth dance is temporary and will be fixed by
future changes to how the `array_alloc` primitive works.)  Then, in
`chpl_mem_array_free()`, try to free through the comm layer and fall back
to freeing to the regular memory layer code iff that fails.

In the module code:
Adapt ChapelBase.chpl:`_ddata_allocate()` to the changed interface to
`chpl_mem_array_alloc()`, via the "array_alloc" primitive.  And then in
DefaultRectangular.chpl, adapt the `_ddata_allocate()` calls to its new
interface.

In the compiler:
Adapt the codegen logic for `primitive("array_alloc")` to the new
`chpl_mem_array_alloc()` interface.

In tests:
Unset the heap size setting for a number of tests that allocate really
big amounts of space.  These settings are now not needed and indeed may
cause us to fail with the` HUGETLB_NO_RESERVE`-related `SIGBUS` on aprun
systems, because they limit how many hugepages are available for
non-heap allocations.  (On srun systems we already limit the heap to
16g, so this is less of a concern there.)

There is a fair amount of future work that could or should follow this.
In no particular order:

- It may be desirable to allow users to enable/disable this feature by
  means of a config const, presumably in ChapelBase.
- The array size threshold should be tunable.
- The comm=ugni registration update broadcast logic needs to be
  scalable.
- We should probably expand this support to dynamically register
  everything, by reworking the jemalloc chunk allocation (and the
  cstdlib memory layer "morecore" hook?) to get memory from the comm
  layer by this interface instead of the old pre-allocated fixed heap
  one.  Then we could have NUMA-sensitivity for things like stacks and
  smaller heap-allocated data.
- It might be nice to add a way to set "interleaved" memory binding for
  the hugepages, to improve performance for applications that are not
  sensitive to sublocale (NUMA) locality but can benefit from the added
  bandwidth when all the NUMA memory pathways are used.


runtime/include/comm/ugni/chpl-comm-impl.h
```diff
@@ -21,6 +21,28 @@
 #define _chpl_comm_impl_h_
 
 //
+// This is the comm layer sub-interface for dynamic allocation and
+// registration of (typically) array memory.
```
ronawho: I'd drop the `(typically) array` since this will (hopefully) change in the future

gbtitus: Will do.

compiler/codegen/expr.cpp
```diff
@@ -2407,9 +2409,22 @@ GenRet codegenCallExpr(const char* fnName, GenRet a1, GenRet a2, GenRet a3)
   args.push_back(a3);
   return codegenCallExpr(fnName, args);
 }
+//static
+//GenRet codegenCallExpr(const char* fnName, GenRet a1, GenRet a2, GenRet a3,
```
ronawho: Remove vs comment out?

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -350,20 +370,35 @@ static int    registered_heap_info_set;
...
-#define NUM_MEM_REGIONS 10
+static int    using_hugepages;
+static size_t hugepage_size;
+
+//
+// Memory regions.  mem_regions contains the address/length pairs and
+// uGNI memory domain handles for each memory region of interest to us.
+// mem_region_map contains a copy of every node's mem_regions.
+//
+#define MAX_MEM_REGIONS 100
```
ronawho: Can/should this be env var settable? Obviously this would require doing dynamic allocation instead, but I have no idea how invasive of a change that would be.

I'm ok punting on this for now (and putting it in a post-commit issue) since we've talked about doing full dynamic registration, which might require a different data structure anyways.

ronawho:
Code looks good. 

Here's what I think needs to be done prior to merging: 

The PR message is really detailed. I understand it since I'm familiar with the work, but I think we need a laymans intro, motivation,  and summary. Something along the lines of: 

"""
Under ugni we used to allocate and register our entire heap at program startup and we'd satisfy dynamic allocation using this heap. Registering touches memory, and set the affinity of the entire heap to numa-domain 0, which meant that we were getting terrible performance for benchmarks that are numa sensitive like stream and the stencil PRK. 

This is the start of dynamic memory registration work. To start, we're just dynamically registering arrays. Essentially we allocate large arrays directly using hugepages, let the Chapel initializer run, and then register the arrays. This way chapel code is responsible for first-touch. 

This has a dramatic performance impact on numa-sensitive applications:
- ~2.5X improvement for stream (now on par with gn-mpi)
- ~2.5X improvement for optimized stencil (close to gn-mpi, remaining gap being investigated by BHarsh and Elliot)
- ~2.5X improvement for single 1D reduction
- ~15% improvement for ISx

Some details: 
 ... rest of the PR message here.
"""


I also think we should open the issue for future TODOs and link to that in the PR message. Here's the TODO's that I know of, though I'm sure we talked about others:
 - Cleanup the array_alloc interface 
 - Determine if we can stop setting HUGETLB_NO_RESERVE for up05
 - try to trap and report a clean error message for OOMs (since we're using HUGETLB_NO_RESERVE, an allocation might "succeed" only to fail later when the kernel actual tries to cobble pages together)
 - Use dynamic registration for all allocations (and eliminate need to ever set CHPL_RT_MAX_HEAP_SIZE for ugni)

compiler/codegen/expr.cpp
gbtitus: Oops!

runtime/src/comm/ugni/comm-ugni.c
gbtitus: Yeah, this could definitely be set at execution time.  I'd prefer to do that as part of post-commit work, though.  I'll add it to the list for that, and I'll do the post-commit issue right after merging this.

Issue #6945, opened 2017-08-06
Exporting function with generic array argument should throw user error


buddha314:
As noted in https://github.com/chapel-lang/pychapel/issues/80 trying to do an `export` on a function with an array argument as in

```
  export
  proc flip_bit(v: [?D] int) {
    v[0] = -v[0];
    return v;
  }
```

causes an internal error
```
(pychapel) buddha314@notes:~/pychapel/tmp$ python call.py
/tmp/temp-lnjLSb.chpl:10: In function 'flip_bit':
/tmp/temp-lnjLSb.chpl:11: internal error: CAL0126 chpl Version 1.16.0 pre-release (c3663a54f0)

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug --
please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.

g++: error: /tmp/tmpEB8cif.a: No such file or directory
Traceback (most recent call last):
  File "call.py", line 12, in <module>
    shout_out()
  File "/home/buddha314/.virtualenvs/pychapel/local/lib/python2.7/site-packages/pych/extern.py", line 212, in wrapped_f
    raise MaterializationError(self)
pych.exceptions.MaterializationError: Failed materializing ({'anames': [],
 'atypes': [],
 'bfile': None,
 'dec_fp': '/home/buddha314/pychapel/tmp/response.chpl',
 'dec_hs': '7ecfac2d168f3423f7104eeb38057ac3',
 'dec_ts': 1501804367,
 'doc': None,
 'efunc': None,
 'ename': 'shout_out',
 'lib': 'sfile-chapel-7ecfac2d168f3423f7104eeb38057ac3-1501804367.so',
 'module_dirs': [],
 'pfunc': <function shout_out at 0x7fe326f428c0>,
 'pname': 'shout_out',
 'rtype': None,
 'sfile': '/home/buddha314/pychapel/tmp/response.chpl',
 'slang': 'chapel',
 'source': None}).
```

So... whuz up wid dat?

ben-albrecht:
> So... whuz up wid dat?

I think the [docs](http://chapel.cray.com/docs/master/technotes/libraries.html) say it all:

>As mentioned above, this feature is not very sturdy. Please refer to Reporting Chapel Issues if any problems are encountered.

Thanks for reporting, so that we can fix this!

[...]

PR #6950, opened 2017-08-07
Add user-facing error for exported function with generic arguments

bradcray:
As reported in issue #6945, using a generic argument in an exported
function was previously causing an internal error in some cases (and
being ignored in others).  This mod generates a user-facing error message
in such cases at the time that we are marking the function
as being generic and adds a test to lock in the behavior.

### DIFF
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -1591,7 +1591,7 @@ bool FnSymbol::tagIfGeneric() {
   bool retval = false;
 
   if (hasFlag(FLAG_GENERIC) == false) {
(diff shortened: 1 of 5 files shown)


bradcray:
Hi @noakesmichael : This is the code change that I was discussing with you, if you'd have the chance to review this PR.  I think it's reasonably straightforward and definitely results in better
errors than what's on master today.

bradcray:
Hi @noakesmichael : Would you take another look?  Note that in taking the refactoring you suggested, I also specialized the error message a bit more such that each formal gets its own line in order to make things a bit more coherent in the event that multiple arguments make a function generic.  See what you think.

PR #6952, opened 2017-08-08
Define SIZE_MAX if needed for runtime

vasslitvinov:
Before this change the runtime was not building under simple configurations.
This is a stopgap measure to hopefully reduce noise in nightly testing.
Copied from qbuffer.h.


dmk42:
Hi, Vass.  The correct place to get `SIZE_MAX` is `<stdint.h>`.  Since it is a standard macro, it is not a good idea for us to define it ourselves because that may have unknown consequences.  (The C compiler might know `SIZE_MAX` internally and our definition might mess it up.)

vasslitvinov:
This change is a part of the workaround in runtime/include/qio/qbuffer.h , which does #include <stdint.h> right before. However, the context there is/may be different from chpl-comm.h. So it may or not work here.

PR #6953, opened 2017-08-08
Convert this future to valgrind-only test

vasslitvinov:
This clarifies that the failure we observe in nightly testing
is due to memory corruption.

While there, add .bad and .prediff (for clean matches against .bad).

Add .skipif because the test has been timing out without valgrind.

Adjust .good because I think the last empty line should not be there.


lydia-duncan:
It seems mysterious to me that a test would run faster with valgrind than without.  Considering there are other differences between our linux64 build and our valgrind testing, is it possible that this is due to one of those (say, using qthreads instead of fifo)?

vasslitvinov:
When there is memory corruption, anything can happen, right?

I wish valgrind had the option "quit after the first violation".
In a sense this is what the .prediff is mimicking.
If this test times out under valgrind, which is possible,
we can revisit.
Of course the best answer is to fix the bug. :)

PR #6955, opened 2017-08-08
Fix use of stdint.h by C++ code

dmk42:
In #457 , there was a workaround to get `SIZE_MAX` when it should have been available to C\+\+ runtime code but was not.

Another such case was found, and #6952 was an attempt at another workaround.  This was reverted in #6954 .

Both problems were due to `__STDC_LIMIT_MACROS` not being set on the first inclusion of `<stdint.h>`.  Setting this macro before subsequent inclusions was too late because the inclusion guard had already been set.  This change sets `__STDC_LIMIT_MACROS` before the first inclusion so that no workarounds are needed.

This issue only affects old C\+\+ compilers such as gcc 4.7.2.  Newer compilers do not need `__STDC_LIMIT_MACROS` set to retrieve the macros from `<stdint.h>`.

Tested with gcc 4.7.2, gcc 6.3.0, and clang (latest Mac release).

Since this has affected many people over the years, I will not merge until the following people have had a chance to look at it.
@gbtitus @mppf @ronawho @vasslitvinov

### DIFF
--- a/runtime/include/qio/qbuffer.h
+++ b/runtime/include/qio/qbuffer.h
@@ -24,10 +24,6 @@
 #ifndef __STDC_FORMAT_MACROS
 #define __STDC_FORMAT_MACROS 1
 #endif
-// This macro set to obtain SIZE_MAX
-#ifndef __STDC_LIMIT_MACROS
-#define __STDC_LIMIT_MACROS 1
-#endif
 
 #include "sys_basic.h"
 #include "qio_error.h"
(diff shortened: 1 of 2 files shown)


gbtitus:
Looks good, and thank you for the informative message!

gbtitus:
BTW, I would say don't wait for @mppf since he's on vacation this week.

dmk42:
@gbtitus thanks, I somehow mixed it up and thought @mppf  was returning Tuesday instead of Thursday.

vasslitvinov:
What Greg said.

dmk42:
@mppf I will go ahead and merge, but feel free to comment later.

mppf:
@dmk42 - thanks for clearing this up. I personally don't view `__STDC_LIMIT_MACROS` as less of a workaround than the earlier strategy of defining `SIZE_MAX`, but I'm happy to defer to your judgement on this subject.

PR #6957, opened 2017-08-08
Add a test to make sure building from --savec generated code works

daviditen:
This test currently fails for CHPL_COMM != none, so is filed as a .future
that is skipped if CHPL_COMM == none.  Once the problem is fixed, both the
.future and the .skipif should be removed.

@ben-albrecht said that PR #6772 is the cause of this failure.

### DIFF
--- a/test/compflags/diten/savec.chpl
+++ b/test/compflags/diten/savec.chpl
@@ -1,0 +1,18 @@
+const outdir = "savec_output";
+const filename = "savec.chpl";
+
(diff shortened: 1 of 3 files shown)


daviditen:
@benharsh @ben-albrecht Here's a test to make sure the --savec problem you discovered/investigated doesn't happen again.  Does this look good to you?

daviditen:
Thanks!

PR #6960, opened 2017-08-08
Fix --savec + make -f behavior

bradcray:
This is a simple fix for the recently broken 'make -f output/Makefile' behavior due to a variable referring to CHPL_RUNTIME_LIB where it seems it should have referred to CHPL_MAKE_RUNTIME_LIB.  While here, I renamed some CHAPEL_* Makefile variables to CHPL_* to conform to other variables.

This fixes and retires the future introduced in PR #6957.


bradcray:
@daviditen and/or @ben-albrecht : Would you review?  I've tested manually, in a 'make install' environment, and using David's test (but, as with any Makefile change, that doesn't necessarily make me confident).

bradcray:
As the author of #6772, @mppf may want to double-check this upon returning.

PR #6962, opened 2017-08-09
Port Chapel to LLVM 4.0.1

dmk42:
This modifies the Chapel compiler to use LLVM 4.0.1 as a back end.

Note that this causes test/llvm/vectorization/simple_loop_novec.chpl to vectorize because LLVM 4.0.1 is smart enough to vectorize it without our help.  We will need to decide what to do with that test.

Other than that, this was verified to pass a full test on linux64 and release/examples on a Mac, with and without WARNINGS=1.


dmk42:
@mppf please review at your convenience.
@coodie this might be of use to you.

mppf:
@dmk42 - thanks for putting this together. Here is a checklist of sorts for this PR:
 - [ ] check with @ronawho about the strategy for including third-party code. I think that on our next LLVM upgrade (ie. this PR) we're expected to change to including LLVM in an llvm-src subdirectory in third-party. However we might make an exception for LLVM/clang since they are big.
 - [ ] check with Tony about testing dependency issues. In particular we'll need cmake >= 3.4.1 as far as I know. We'll need nightly testing systems to have that version of cmake available and enabled.
 - [ ] consider removing the support in the build system for the auto-tools builds. If we keep it, it will be so that the tree works with LLVM <= 3.8 without cmake for a while. See the next point.
 - [ ] think about when we will remove the #ifdef conditionals for older versions of LLVM. How far back should the Chapel compiler support? I think it'd be reasonable to drop at least to 3.7 and possibly require 4.0. One hazard of having these #ifdefs in place is that they aren't tested with any version of LLVM other than the one we ship. I think they're becoming crufty enough that we should clean them up. In my view, they make sense transitionally but not long-term.

dmk42:
@ronawho I'm happy to make this use llvm-src if that's the decision.  What should factor into the decision is that LLVM is large enough that it takes a long time to extract.  Do we want to pay that cost during a git clone or during a build?

dmk42:
@mppf LLVM 5.0.0 comes out in two weeks.  I would suggest leaving support for earlier versions in until we can add preliminary support for 5.0, just to make it easier to see where differences are likely to occur.  Once we have moved to 4.0.1 and have at least initial changes toward 5.0, that would be a good time to remove everything prior to 4.0.

ronawho:
LLVM is sufficiently different that I don't feel strongly about what we do. 

Note that the main reason for switching to `*-src` format when we switched to git was so that the diff for version upgrades is smaller, but that was for packages that we stored untarred. e.g. for gasnet upgrades we used to: 
 - commit the raw tarball so we could find it later if we needed
 - unpack the new tarball into a gasnet-<VER> dir
 - update our makefile to point to the new gasnet-<VER> dir
 - eventually remove the old gasnet-<OLD-VER> dir

and this resulted in checking in a tarball, and creating a new directory every time, which really bloats the git history (and since git is DVCS increases the size of every clone.) Now we just commit the diffs between versions. 

For llvm, we're not currently storing the unpacked source, so that initial commit will quite big, but upgrades might result in smaller diffs (assuming llvm doesn't dramatically alter their src layout.)

dmk42:
Thanks @ronawho for clarifying.  Given this information, I would suggest the following way forward.

Because the number of people using LLVM is in the minority prior to Chapel 1.16, keep distributing it as a tarball for now, so only those who use it pay the price.

Because Brad wants to move toward LLVM being the default after Chapel 1.16, start using an llvm-src directory right after the 1.16 release.

dmk42:
Rethinking this, delaying the switch to llvm-src results in a permanently larger git history, so I might as well do it now.  Anybody object?

mppf:
@dmk42 - I'm fine either way. It's not obvious to me if the tarball approach creates smaller or larger git history vs checking in the files. I suppose you could answer that question with some experiments on different branches.

mppf:
I think at the moment I'm leaning towards leaving it a tarball for now (and possibly re-considering when we upgrade to LLVM 5 - in more of a context of llvm-is-the-default).

dmk42:
@mppf OK, experiments show I was wrong, and it's a smaller hit to the git history size to delay the move to llvm-src anyway.  Sticking with the way we do it now keeps it simple pre-Chapel-1.16, so I now agree that the upgrade to LLVM 5 is the time to move to llvm-src.

mppf:
did you actually do the experiment? How big was the size difference, I'm curious?

dmk42:
The diff between llvm-src's is about 51M, and storing an extra tarball costs about 27M.

ronawho:
Just to be clear, the benefit is about upgrades over time. The source directory may be large, but hopefully the different between versions is relatively small as opposed to committing a 30 mb tarball every time, so it may take a few releases to see the benefit.

dmk42:
I gave Tony a new version of cmake (which I've been testing all along).  He's making it accessible to all the LLVM test jobs.

For the autotools support, I would suggest removing that right after the Chapel 1.16 release just in case.  Any objections?

dmk42:
Tony's PR to use the new cmake in nightly tests is #6983 .

mppf:
I'm fine with removing the autotools support after 1.16 but it would also be fine with me to do that once we are removing support for some older LLVM versions.

Here's how I see the Future Work we've been discussing here:
 - Just after 1.16 is released, switch to LLVM-is-the-default
 - After LLVM 5.0 is working with Chapel (but not necessarily included), remove support for LLVM < 4
 - Remove autotools-build support for LLVM at one of the times above
 - When we upgrade to LLVM 5.0, switch to an llvm-src strategy instead of tarballs

Thanks!

dmk42:
The results we decided on are documented in #6993 .

This PR is now ready to merge, but I am headed to London today.  I will plan to merge it when I return on Monday, August 21st, unless @mppf would like to merge it earlier.

Issue #6566, opened 2017-06-28
Adjusting error handling policy to avoid implicit halt


mppf:
### CHIP 8 describes implicit `try!`

Right now, CHIP 8 describes a design that will implicitly add `try!` to calls to functions marked `throws` if those calls are not enclosed in any of a throwing function, a `try`, or a `try!`. This behavior is described by 1.ii. of the following piece of CHIP 8:

> To address the desire for both fast prototyping and enforced error handling, the compiler will run checks in one of two modes, based on a flag.
>
> If a call to a throwing procedure is not enclosed by a try or try!:
>
>    1. Default.:
>        i. if the call is contained in a procedure labelled with throws, any error encountered will be propagated.
>        ii. if the call is contained in a procedure not labelled with throws, the program will halt() if an error is encountered.
>    2. Strict.: the compiler will raise an error.

For example, the test() function below compiles without error and results in a halt at run-time, because `mayThrow` generated an error that was not handled.

``` chapel
proc mayThrow() throws {
  throw new Error("Test");
}

proc test() {
  mayThrow();
}

test();
```

In contrast, a similar example that throws an error directly is a compilation error (cannot throw from a non-throwing function):

``` chapel
proc test() {
  throw new Error("Test");
}

test();
```

### Problems with implicit `try!`

This implicit `try!` behavior is problematic because:
 * it introduces implicit halts, where the user might not notice them. Recall that one of the goals of the error handling effort was to cut down on implicit halts.
 * it produces inconsistent behavior across similarly structure code. In particular, the above examples only differ in the introduction of a function and so it is surprising that they behave so differently.
 * the implicit `try!` was added as a part of design improvements to add an implicit `try`. In retrospect, implicit `try!` is not necessary to achieve the design goals.

### Proposed Solution

Instead of *default mode* and *strict mode*, let's talk about 3 ways in which the current *strict mode* can be relaxed.

1. *relaxed try*: a `try` is not necessary around a call to a function that can throw if that call is contained in a function that `throws`.
   For example
   ``` chapel
     proc callMayThrow() throws {
       mayThrow(); // compiler automatically adds try
     }
   ```

2. *relaxed throws*: includes *relaxed try* and also: the compiler will infer when a function should be marked `throws`. In particular, functions that are not marked `throws` but that contain un-handled calls to throwing functions or `throws` itself will be marked `throws`. Note that this analysis might need to be recursive.
   For example
   ``` chapel
[...]

PR #6963, opened 2017-08-09
use error handling in standard IO module

psahabu:
## purpose
Use the IO module as a major test case for the error handling system, in turn making the module more resilient.

### leave syserr intact
The IO module uses `syserr` as a representation for error codes returned from the runtime, passing them around internally and `out` externally. Replacing each function that takes an `out syserr` with `throws` and strict error handling would require extensive changes by library users, outside the intended scope of this change.

### `ioerror()`
Instead of replacing each instance of `syserr`, this change focuses on the function `ioerror()`, which was used in the library to halt the program on certain system errors. This functionality has been changed to `throw` a `SystemError`, a subclass of `Error` that contains a message about the `syserr`. Some intentionally failing tests had their output changed because the `halt()` is done through the error handling system instead of by `ioerror()` directly.

### `throws`, `try`, `try!` throughout
`throws` was added to any function that called `ioerror()`, along with the `try` annotation so that the IO module itself would compile in strict mode. This naturally extended to any functions that called those throwing functions. However the throwing functions can be safely ignored by users in default mode.

## results and outlook
linux64+flat testing is clean in default mode and fails 1529/5962 tests in strict mode, which was expected given that many tests do not wrap with `try`. The goal is to eventually replace `syserr` with the error handling system, but it would require deprecation warnings so that users can have a smooth transition.

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1381,10 +1381,10 @@ proc file.close(out error:syserr) {
 
 // documented in error= version
 pragma "no doc"
-proc file.close() {
+proc file.close() throws {
   var err:syserr = ENOERR;
   this.close(err);
-  if err then ioerror(err, "in file.close", this.tryGetPath());
+  if err then try ioerror(err, "in file.close", this.tryGetPath());
 }
 
 /*
@@ -1411,10 +1411,10 @@ proc file.fsync(out error:syserr) {
 
 // documented in the error= version
 pragma "no doc"
-proc file.fsync() {
+proc file.fsync() throws {
   var err:syserr = ENOERR;
   this.fsync(err);
-  if err then ioerror(err, "in file.fsync", this.tryGetPath());
+  if err then try ioerror(err, "in file.fsync", this.tryGetPath());
 }
 
 
@@ -1469,11 +1469,11 @@ proc file.tryGetPath() : string {
 Get the path to an open file. Halt if there is an error getting the path.
 
 */
-proc file.path : string {
+proc file.path : string throws {
   var err:syserr = ENOERR;
   var ret:string;
   ret = this.getPath(err);
-  if err then ioerror(err, "in file.path");
+  if err then try ioerror(err, "in file.path");
   return ret;
 }
 
@@ -1485,13 +1485,13 @@ change if other channels, tasks or programs are writing to the file.
 :returns: the current file length
 
 */
-proc file.length():int(64) {
+proc file.length():int(64) throws {
   var err:syserr = ENOERR;
   var len:int(64) = 0;
   on this.home {
     err = qio_file_length(this._file_internal, len);
   }
-  if err then ioerror(err, "in file.length()");
+  if err then try ioerror(err, "in file.length()");
   return len;
 }
 
@@ -1591,6 +1591,7 @@ proc open(out error:syserr, path:string="", mode:iomode, hints:iohints=IOHINT_NO
       var (host, port, file_path) = parse_hdfs_path(url);
       var fs:c_void_ptr;
       error = hdfs_connect(fs, host.c_str(), port);
+      // psahabu: strict mode should be complaining about this
       if error then ioerror(error, "Unable to connect to HDFS", host);
       /* TODO: This code is an alternative to the above line, which breaks the
          function's original invariant of not generating errors within itself.
@@ -1664,10 +1665,10 @@ proc open(out error:syserr, path:string="", mode:iomode, hints:iohints=IOHINT_NO
 // documented in open(error=) version
 pragma "no doc"
 proc open(path:string="", mode:iomode, hints:iohints=IOHINT_NONE, style:iostyle =
-    defaultIOStyle(), url:string=""):file {
+    defaultIOStyle(), url:string=""):file throws {
   var err:syserr = ENOERR;
   var ret = open(err, path, mode, hints, style, url);
-  if err then ioerror(err, "in open", path);
+  if err then try ioerror(err, "in open", path);
   return ret;
 }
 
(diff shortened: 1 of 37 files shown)


modules/standard/SysError.chpl
```diff
@@ -41,6 +41,19 @@ module SysError {
...
+class SystemError : Error {
+  var err: syserr;
+
+  proc SystemError(err: syserr, msg: string) {
+    this.err = err;
+    this.msg = msg;
+  }
+
+  proc writeThis(f) {
+    try! f.write(msg);
```
mppf: Should writeThis methods be `throws` ? If so we'll need to update buildDefaultFunctions & documentation about readThis/writeThis/readWriteThis. Let's make sure to write down a TODO somewhere.

mppf:
I'm wondering how we might be able to opt-in to strict errors at a module level. It seems to me that we ought to enable strict error checking for library code (say the IO.chpl module) but leave it off for most tests (since they aren't meant to be library code). Seeing > 1000 failures for strict mode indicates to me more the limits of the compiler flag approach (rather than suggesting that we need to change all of those tests).

As a straw-man, what if you did "use StrictErrors;" to opt-in?

modules/standard/IO.chpl
```diff
@@ -3981,8 +3980,10 @@ proc channel.writeln(out error:syserr):bool {
 
 // documented in style= error= version
 pragma "no doc"
-proc channel.writeln():bool {
-  return this.write(new ioNewline());
+proc channel.writeln():bool throws {
+  try {
+    return this.write(new ioNewline());
+  }
```
mppf: In Swift you can write such a method like this:
``` swift

func f() throws {
    return try g();
}
```

Is there a reason we can't support the same? (can you make a note in our error handling TODOs issue?)

modules/standard/IO.chpl
```diff
@@ -2169,7 +2169,7 @@ inline proc channel.unlock() {
 proc channel.offset():int(64) {
   var ret:int(64);
   on this.home {
-    this.lock();
+    try! this.lock();
```
mppf: A future TODO: replace try! this.lock with this.lock; defer this.unlock();
I think it's reasonable to leave this as try! for now, but we need to record a TODO somewhere. In functions that throw, it seems odd to use try!.

test/types/file/unableToOpenFile.good
```diff
@@ -1 +1 @@
-unableToOpenFile.chpl:1: error: No such file or directory in open with path "_test_cannotOpenMe.txt"
+unableToOpenFile.chpl:1: error: uncaught error - No such file or directory in open with path "_test_cannotOpenMe.txt"
```
mppf: The style like
```
unableToOpenFile.chpl:1: error: uncaught error - No such file or directory in open with path "_test_cannotOpenMe.txt"
```

seems redundant - maybe it could be:
```
unableToOpenFile.chpl:1: uncaught error: No such file or directory in open with path "_test_cannotOpenMe.txt"
```

On a related note, in an ideal world, creating an Error object will capture the file/line information of where the error was created (and possibly more of a stack trace). In that world, we'd probably have multiple lines output per error, since we'd want to report two (or more) locations. What would the ideal output look like in that case? 

In most of these examples, the compiler is intentionally hiding the distinction (by reporting the "user code location" ignoring internal modules). But we could easily construct a case where a user module creates a function that throws and then calls it. Should the linenumber reported be to where the error is created? Where it is thrown? Or where it is not caught? (I think it should show where it is created & where it is not caught).

psahabu: While I like the second style you offer, the `error: ` comes from the `chpl_error` primitive called to halt on an uncaught error. This would require a new runtime function, which might be worthwhile.

After talking with @lydia-duncan about this, we preferred that the type of the error (e.g. `SysError`) be included in the message at some point. Here's one concept:

```
unableToOpenFile.chpl:1: uncaught SystemError - No such file or directory in open with path "_test_cannotOpenMe.txt"
```

In the ideal world, we might want to look at how Java does [stack traces on exceptions](https://en.wikibooks.org/wiki/Java_Programming/Stack_trace) as a starting point. If I understand what you said correctly, I agree that we should show where the error was initially created and where it was not caught. Something like:

```
uncaught SystemError in thread 0 on locale 0:
    at unableToOpenFile.main() [unableToOpenFile:1]
    at [where the error was created]
```

A stack trace would fill in the `throw` locations between those.

mppf: Let's call what happens in this PR "good enough for now" and open an issue to talk about the exact details of the format, wherein we can discuss further the Java analogue, original line numbers, etc. I do have more to say, so @ mention me when you make the issue.

modules/standard/SysError.chpl
psahabu: I've put it under the design questions section in #4977.

modules/standard/IO.chpl
psahabu: In the initial design phase I thought of `try` as being used on statements rather than expressions, but this might be a distinction without difference and I would prefer this form to `try return g();`. I've put it under the everything else section in #4977.

psahabu:
`use StrictErrors;` seems like the most reasonable first step, utilizing our current namespacing rules for strict mode. However, how would this approach interact with the proposals in #6566?

test/types/file/unableToOpenFile.good
psahabu: I've filed #7013.

mppf:
@psahabu - I think we can just create other variants, e.g. `use RelaxedErrors;` and `use RelaxedTry;` (modulo discussion about naming). I think each relaxation proposal makes sense when applied to a module at a time (even though the result might be confusing in some cases).

PR #6964, opened 2017-08-09
Fix using_chapel_module()

ronawho:
In #6946 I changed using_chapel_module() to do a normpath on CHPL_HOME and
CHPL_MODULE_HOME before comparing them. If both were unset, this would
incorrectly result in the function thinking the chapel module was being used,
which resulted in building the runtime for the lowest common denominator on the
whiteboxes.

This updates using_chapel_module() to only do the normpath after checking if
CHPL_HOME and CHPL_MODULE_HOME were actually set.

### DIFF
--- a/util/chplenv/chpl_home_utils.py
+++ b/util/chplenv/chpl_home_utils.py
@@ -18,10 +18,6 @@ def get_chpl_home():
     return chpl_home
 
 @memoize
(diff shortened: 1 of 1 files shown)


ronawho:
@ben-albrecht I'm probably going to merge now, but would appreciate a post-commit review

PR #6965, opened 2017-08-09
Add nonnull attribute to reference function arguments

coodie:
This PR adds nonnull attribute to LLVM IR for function parameters  that are references.


compiler/codegen/symbol.cpp
```diff
@@ -1181,12 +1181,17 @@ void FnSymbol::codegenPrototype() {
...
+    std::vector<int> refArgs;
     for_formals(arg, this) {
       if (arg->hasFlag(FLAG_NO_CODEGEN))
         continue; // do not print locale argument, end count, dummy class
       argumentTypes.push_back(arg->codegenType().type);
       argumentNames.push_back(arg->cname);
+
+      arg->codegenType();
       numArgs++;
+      if(arg->intent & INTENT_REF)
```
mppf: I think you can use arg->isRef() here

compiler/codegen/symbol.cpp
```diff
@@ -1181,12 +1181,17 @@ void FnSymbol::codegenPrototype() {
...
+        refArgs.push_back(numArgs);
```
mppf: Here, numArgs is 1-based, since this happens after we incremented it. I would expect LLVM argument indices to be 0-based... but it's not immediately obvious to me from the LLVM documentation. If in fact the are 1-based, could you add a comment to that effect here?

mppf:
Passed full local testing, --llvm testing, and --llvm --fast testing.

PR #6969, opened 2017-08-09
Add future for llvm bug with extern arrays

benharsh:


benharsh:
@mppf : I ran into this while working on serialization.

PR #6973, opened 2017-08-09
Fix this.init() calls in initializers on generic types

lydia-duncan:
Prior to this change, we would get an internal failure for initializers on
generic types with `this.init()` calls in them, because we wouldn't have
determined the instantiation at the time of the call - we'd only know it after
the call completes.

We also get that failure for `super.init()` calls when the parent type is generic,
but that will be addressed in a separate PR.

Details:
---------
Adds a flag, FLAG_DELAY_GENERIC_EXPANSION, and applies it to "this"
arguments for initializers on generic types, as well as the actuals used for
calls to them.  Then, once we have finished resolving a call with an actual
with that flag, use the type of the matching formal (after verifying it is no
longer generic) to update the actual.  This solution could be applied to the
normal strategy used for updating the actuals to initializer calls, generalizing
the code somewhat.

Testing:
----------
Added a new test for initializers that use `this.init()` when the type is generic.
Modified the expected output for one of the futures in test/classes/initializers.

Passed a paratest over the initializers directory with futures and a full paratest
over std/


lydia-duncan:
TODO: maybe remove some more code from initializersResolution?

lydia-duncan:
Okay, I don't think there's anything else I can remove right now.  At least, not without stepping on Mike's toes.  @noakesmichael, would you mind giving this a review when you get the chance?

PR #6975, opened 2017-08-10
Add path to 'chpl' into system command in case it isn't in user's path

bradcray:
@vasslitvinov pointed out that this test doesn't work if 'chpl' isn't in the user's path.  Here, I'm updating the test to form an absolute path to 'chpl' to deal with that potential problem.

### DIFF
--- a/test/compflags/diten/savec.chpl
+++ b/test/compflags/diten/savec.chpl
@@ -5,7 +5,7 @@ var ret: int;
 
 extern proc system(s: c_string): int(32);
 
-ret = system(("chpl --savec " + outdir + " " + filename).c_str());
+ret = system((CHPL_HOME+"/bin/"+CHPL_HOST_PLATFORM+"/"+"chpl --savec " + outdir + " " + filename).c_str());
 if ret != 0 then
   halt("Error compiling Chapel code");
 


ronawho:
Looks fine to me. 

Unrelated, and not your test, but it seems like we could/should use the spawn module instead of externing to system

bradcray:
I wondered whether @daviditen had a reason to use system() rather than spawn() as well...

ronawho:
Just noticed this while looking for other tests that use Spawn, but patterns/cltools/fileCompile/chplCompile has a slightly different solution for putting chpl in the path. I don't really have a preference between the two approaches though.

bradcray:
I had a vague memory of a test like that, but couldn't find it grepping for tests containing spawn/system and 'chpl' on the same line.  And no wonder given how much this test chops things up.

It looks like the trick there is just to add the path to .execenv before running the test?  Seems nicer/more straightforward to me to have the test embed the path.  (For that matter, that test makes using system() seem much more straightforward than spawn() which seems too bad...  I assume spawn() has virtues in its array-of-strings approach that I'm not detecting, but the "one string for the whole command" definitely seems more straightforward and intuitive at a glance.

ronawho:
array-of-strings is based on pythons subprocess and is done for security measures (can give more details on that if you want), but I'm pretty sure that test just uses a single string and `spawnshell()` instead of the array-of-strings and `spawn()`

You should be able to do something like `spawnshell(CHPL_HOME+"/bin/"+CHPL_HOST_PLATFORM+"/"+"chpl --savec " + outdir + " " + filename).wait();
`

daviditen:
I used the extern `system()` because I didn't remember that the Spawn module existed.

Rewriting them to use `spawn()` or `spawnshell()` feels a little clunky because the test needs to look at the return codes, so it would actually be more like:

```
var sub = spawnshell(CHPL_HOME + "/bin/"+CHPL_HOST_PLATFORM + "/" +
                     "chpl --savec " + outdir + " " + filename);
sub.wait();
ret = sub.exit_status;
```

vs.

```
ret = system((CHPL_HOME + "/bin/" + CHPL_HOST_PLATFORM + "/" +
              "chpl --savec " + outdir + " " + filename).c_str());
```

bradcray:
@daviditen's comment makes me wonder if we should have the `Spawn` module support a `system()` call for convenience that essentially wraps the spawnshell() + wait + status idiom that he shows above.  I'm not familiar enough with the design of the `Spawn` module to know if there'd be a reason not to do this, though.  @mppf?

mppf:
Sure, I think it'd be reasonable to add a combined spawn+wait command. I guess I just don't see the pattern with sub.wait() and ret = sub.exit_status as so clunky.
We based the functions in Spawn on the similar support in Python. In 3.5 it looks like they added a combined method (run). But even in Perl or Bash you have the problem that when calling out to a command you might be interested in the return code and/or the output. So I think returning a record and then querying the exit status makes sense (i.e. I wouldn't want to get rid of the ret = sub.exit_status line). I'm not sure I'd buy an argument that querying the return code is more common than querying the output (since backtics in bash say get the output but not the return code).

So, all that said, we could combine the sub.wait() call in, but it wouldn't work for redirecting output to a pipe, and so the interface might look different anyway. I'd personally lean towards leaving it alone but I wouldn't object if somebody wanted to look at adding another function to Spawn. If they did, I'd recommend trying to base it off of `subprocess.run` in Python 3.5 (even if we choose a different name).

PR #6978, opened 2017-08-10
Update a.out in documentation & test comments

mppf:
PR #6811 changed the default executable name from a.out. This PR updates user-facing documentation and test comments to match.

- [x] full local testing


ben-albrecht:
Should we update the `Language_Overview.tex` section of the spec as well?

There are also a handful of non-user facing cases that would be easy to fix if you're interesting:

- `util/devel/runManyTimes` 
- `modules/standard/SysBasic.chpl`

mppf:
The SysBasic one is supposed to stay a.out since it's a linux error description. I'll update the files.cpp comment but the others don't seem worth fixing to me. (I don't see a problem with cleaning up a.out; runManyTimes can default to a.out for now).

I've updated the language overview spec section.

mppf:
@ben-albrecht - once you give this PR a nod I think it's ready to merge. Thanks.

PR #6984, opened 2017-08-10
fix PRIM_ARRAY_SET_FIRST on a ref with --no-local

vasslitvinov:
The following used not to work with --no-local:

    proc init_elts(ref x: _ddata(int)) {
      var i:int, elm:int;
      __primitive("array_set_first", x, i, elm);
    }

This simple change fixes that.

Testing: linux64 --verify; gasnet on multilocale tests.


vasslitvinov:
@benharsh - OK with you? Testing passes.

PR #6989, opened 2017-08-11
Use pkg-config to determine re2 link args

ronawho:
Use pkg-config instead of hard-coding `-lpthread` for re2's link args.

This also removes our custom install-static target (which didn't install the
.pc file) and just removes the shared-object so we continue to link against the
static lib.


ronawho:
This fails for clang with `clang: warning: argument unused during compilation: '-pthread' [-Wunused-command-line-argument]` :(

Issue #6972, opened 2017-08-09
Design questions about `chplconfig` files, `./configure` and `printchplenv`


bradcray:
* It seems that when doing a `./configure` and then `make install` the `chplconfig` file that's created is left in the original copy of the source tree, affecting future uses of it until it's manually removed or another `./configure` step is run.  Is this intentional?  Desirable?

* When using a `chplconfig` file and `printchplenv`, settings that are established by the `chplconfig` file are marked with a `*` similar to those set via environment variables.  This confused me for awhile because I checked my environment and saw that I had not set them myself explicitly (which is what I think of the `*` as indicating).  Should they be marked with some second, distcint symbol?  (or not marked at all?)

bradcray:
Tagging @mppf, @ben-albrecht, and maybe @ronawho as maybe being interested in this.

mppf:
For the first point, it's typical for a `./configure` based project to have that behavior. The way to avoid it is to use out-of-tree builds - where you configure in say `$CHPL_HOME/../mybuild` and then run `make` in that directory. We don't support out-of-tree builds right now. One goal of our `./configure` support is to make Chapel fit in a bit more with the build (and install) behavior people are used to. So I'm not sure we should try to change it directly, but we could seek to support out-of-tree builds (say).

For the second, I think it'd be great to have an alternative to `*` to mean "it came from config file".

bradcray:
On the first point, I'm OK leaving it as-is (and am not particularly excited about having any of us put in the effort to support out-of-tree builds at this point).  It would also seem natural to me to have `make install` `mv` the `chplconfig` file rather than copying it, based on an assumption that once people do `make install` they wouldn't necessarily care whether it was still there or not, but maybe there's a use case I'm not anticipating.

For the second, I took a tab at this in PR #6991.

mppf:
I like the idea of separating but I think `+` would be a better marker than `*` since the environment variables take precedence over the file values (i.e. they bind more tightly). But it's mostly arbitrary.

vasslitvinov:
Sounds good to me. I don't have a preference for the character.

ronawho:
I also have a slight preference for `+`, but I don't feel strongly about it either

bradcray:
PR #6991 addressed item 2 above.  Though I think item 1 still has merit, it didn't gain much support this time around, so I'm closing this issue until someone else gets bitten by it.

ben-albrecht:
> Though I think item 1 still has merit

[...]

PR #6991, opened 2017-08-11
Printchplenv distinguish config env

bradcray:
I got into some confusion the other day due to a lingering chplconfig file that I wasn't aware of, made worse because printchplenv doesn't distinguish between variables set in the environment vs. a chplconfig file (which led to my questions in #6972).  This is my Python-ignorant attempt to have the `[print]chplenv` scripts be able to distinguish between those two cases.

### DIFF
--- a/util/chplenv/overrides.py
+++ b/util/chplenv/overrides.py
@@ -197,11 +197,12 @@ def get(var, default=None):
 
 @memoize
 def get_environ(var, default=None):
-    """ Check if variable has a default defined somewhere and return value """
+    """ Check if variable has a default defined in the environment and return value """
     return os.environ.get(var) or default
 
+@memoize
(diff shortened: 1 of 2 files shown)


bradcray:
@ben-albrecht : GitHub thinks that you would be the best to review this, and who am I to disagree?

ben-albrecht:
I think `printchplenv --help` output needs to be updated to distinguish `*` from `@`

bradcray:
Good catch!

bradcray:
@ben-albrecht : Help message has now been updated.

ben-albrecht:
Could you confirm that the `test/chplenv/` tests pass? Other than that, looks good to me.

bradcray:
Good catch -- there would've been one failure (I'd forgotten we had chplenv tests).

PR #6992, opened 2017-08-11
Re-enable -Wunused for GASNet-related files

bradcray:
As of GASNet 1.28.2, we can compile GASNet's headers with -Wunused
(see https://upc-bugs.lbl.gov/bugzilla/show_bug.cgi?id=2538),
so no longer need to throw -Wno-unused when compiling our runtime
files that #include them.  This turned up a few unused variables
which I removed.


bradcray:
@ronawho: Would you review this and let me know what degree of testing you think it should get?

ronawho:
Looks good! I think just a `make check` with gcc and clang is enough

PR #7001, opened 2017-08-11
Add a number of memory leak annotations

bradcray:
I've had an IOU to add these for far too long...  Track memory-related
ups and downs from the past month or so.


bradcray:
@ronawho : Want to be my sanity check?

ronawho:
👍

PR #7002, opened 2017-08-11
Test perf of forall-intent changes in perf playground

vasslitvinov:
(following the suite of  #6816)


vasslitvinov:
@benharsh I am taking it over, OKed by @ronawho .

Issue #5577, opened 2017-03-09
Automatically rebuild third-party dependencies when they change


ronawho:
We have good dependency checking for the compiler and runtime, but not for third-party libraries. 

i.e. when we upgrade qthreads/hwloc/jemalloc/gasnet versions we don't automatically rebuild. Additionally, we don't rebuild when breaking makefile changes are made (like with #5566.)

This is really a developer-only problem, but even those of us on the core team are often caught off-guard by breakages that result from these sorts of things.

ronawho:
Relatively simple idea might be to compute the checksum of the the third-party source and our Makefiles, and use that checksum as a dependency

upcrob:
I just got burned by #5566 when I tried to build today as well.  I imagine more thought should go into this than I've put into it, but the following worked for me:

```
diff --git a/Makefile b/Makefile
index 8aed32c..664e809 100644
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ all: comprt
 
 comprt: FORCE
 	@$(MAKE) compiler
+	@$(MAKE) jemalloc
 	@$(MAKE) third-party-try-opt
 	@$(MAKE) always-build-test-venv
 	@$(MAKE) always-build-chpldoc
@@ -83,6 +84,11 @@ runtime: FORCE
 third-party: FORCE
 	cd third-party && $(MAKE)
 
+jemalloc: FORCE
+	-@if [ "$(git show | grep \"third-party/jemalloc\" | wc -c)" != "0" ]; then \
+	cd third-party/jemalloc && $(MAKE); \
+	fi
+
 third-party-try-opt: third-party-try-re2 third-party-try-gmp
 
 third-party-try-re2: FORCE
```

It's a little (a lot?) naive and hacky, especially since it will always result in a third-party rebuild from that commit, but for simple scenarios it seems to work.

ronawho:
@upcrob interesting idea. That'll only work if the last commit involved a change to jemalloc though, right? At least for me `git show` just shows the last commit. If you pulled in 20 commits, and the middle one was the jemalloc upgrade I think it would be missed. 

Still we could do something along those lines -- maybe store the sha at the time a 3rd party was built, and do a diff against the current one checking for changes to third-parties. I worry about the time that would add to a no-op build since git operations can be slow, especially on non-local filesystems.

I had originally been thinking we'd want a solution that didn't rely on git for people using the tarball, but we probably only have to be concerned about this for developers that are using the git repo.

[...]

PR #7004, opened 2017-08-13
Automatically rebuild third-party libraries

ronawho:
We're pretty good at rebuilding the compiler/runtime when code changes, but we
never rebuild the third-party libraries. Even when we upgrade third-party
versions or start throwing different configuration options, we don't do
anything to rebuild. This usually means that developers aren't testing the new
version. However, some changes cause ABI incompatibles that break the build in
obscure ways and force devs to manually clobber their workspace, which often
confuses even internal devs.

This adds code to automatically rebuild third-party libraries if any of their
C/C++ source files or our makefiles change.

Ideally we would also rebuild if any of the third-party files change, but
unfortunately for several packages that use autoconf such as gasnet, hwloc, and
qthreads we touch some configure/autoconf files in a certain order to prevent
autoconf from running, and this changes the timestamps, which causes make to
think the files were modified and we need to rebuild.

Closes https://github.com/chapel-lang/chapel/issues/5577

### DIFF
--- a/third-party/Makefile
+++ b/third-party/Makefile
@@ -67,16 +67,25 @@ depend:
 # automatically rebuild on 3rd party version upgrades or if we specify
 # different option in the Makefile.
 
-SRC_FILES = -iname '*.cc'  -o -iname '*.cp'  -o -iname '*.cxx' -o -iname '*.cpp' -o  -iname '*.CPP' -o -iname '*.c++' -o  -iname '*.hh'  -o -iname '*.hp' -o -iname '*.hxx' -o -iname '*.hpp' -o -iname '*.HPP' -o -iname '*.h++' -o -iname '*.tcc' -o -iname '*.c'   -o -iname '*.h'
-AMUDPRUN_DEPEND := $(shell find $(AMUDPRUN_SRC_DIR) $(SRC_FILES) 2>/dev/null) $(AMUDPRUN_DIR)/Makefile*
-GASNET_DEPEND := $(shell find $(GASNET_SUBDIR) $(SRC_FILES) 2>/dev/null) $(GASNET_DIR)/Makefile*
-GMP_DEPEND := $(shell find $(GMP_SUBDIR) $(SRC_FILES) 2>/dev/null) $(GMP_DIR)/Makefile*
-HWLOC_DEPEND := $(shell find $(HWLOC_SUBDIR) $(SRC_FILES) 2>/dev/null) $(HWLOC_DIR)/Makefile*
-JEMALLOC_DEPEND := $(shell find $(JEMALLOC_SUBDIR) $(SRC_FILES) 2>/dev/null) $(JEMALLOC_DIR)/Makefile*
-LIBUNWIND_DEPEND := $(shell find $(LIBUNWIND_SUBDIR) $(SRC_FILES) 2>/dev/null) $(LIBUNWIND_DIR)/Makefile*
-MASSIVETHREADS_DEPEND := $(shell find $(MASSIVETHREADS_SUBDIR) $(SRC_FILES) 2>/dev/null) $(MASSIVETHREADS_DIR)/Makefile*
-QTHREAD_DEPEND := $(shell find $(QTHREAD_SUBDIR) $(SRC_FILES) 2>/dev/null) $(QTHREAD_DIR)/Makefile*
-RE2_DEPEND := $(shell find $(RE2_SUBDIR) $(SRC_FILES) 2>/dev/null) $(RE2_DIR)/Makefile*
+# Common C/C++ file extensions
+SRC_FILES_FILTER += -iname '*.c'   -o -iname '*.h'   -o -iname '*.cpp' -o
+SRC_FILES_FILTER += -iname '*.cc'  -o -iname '*.cxx' -o -iname '*.hpp' -o
+# Less common C++ extensions
+SRC_FILES_FILTER += -iname '*.c++' -o -iname '*.hp'  -o -iname '*.hxx' -o
+SRC_FILES_FILTER += -iname '*.h++' -o -iname '*.tcc' -o -iname '*.cp'
+
+# Find 3rd party dependencies -- arg 1 is src dir, arg 2 is top-level dir with Makefiles/README
+find_3p_depend = $(shell find $(1) $(SRC_FILES_FILTER) 2>/dev/null) $(2)/Makefile* $(2)/README
+
+AMUDPRUN_DEPEND       := $(call find_3p_depend, $(AMUDPRUN_SRC_DIR),      $(AMUDPRUN_DIR))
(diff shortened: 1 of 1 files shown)


third-party/Makefile
```diff
@@ -88,31 +107,42 @@ else ifeq ($(wildcard $(GMP_H_FILE)),)
...
+QTHREAD_DEPEND_INSTALL_DIRS += $(HWLOC_INSTALL_DIR)
 endif
-$(QTHREAD_INSTALL_DIR):
+ifeq ($(CHPL_MAKE_JEMALLOC), jemalloc)
+QTHREAD_DEPEND_INSTALL_DIRS += $(JEMALLOC_INSTALL_DIR)
+endif
+QTHREAD_DEPEND_INSTALL_DIRS += $(QTHREAD_INSTALL_DIR)
+qthread: $(QTHREAD_DEPEND_INSTALL_DIRS)
+
+$(QTHREAD_INSTALL_DIR): $(QTHREAD_DEPEND) $(HWLOC_DEPEND) $(JEMALLOC_DEPEND)
```
gbtitus: When hwloc and/or jemalloc are not in use we could save a little time by not including their `$(*_DEPEND)` files in the `$(QTHREAD_INSTALL_DIR)` target dependencies.

third-party/Makefile
```diff
@@ -62,6 +62,23 @@ clobber: FORCE
 
 depend:
 
+# Third-party dependencies. If any of these files (3rd party source dirs or our
+# Makefiles) change, then rebuild the 3rd party lib. This allows us to
+# automatically rebuild on 3rd party version upgrades or if we specify
+# different option in the Makefile.
+
+SRC_FILES = -iname '*.cc'  -o -iname '*.cp'  -o -iname '*.cxx' -o -iname '*.cpp' -o  -iname '*.CPP' -o -iname '*.c++' -o  -iname '*.hh'  -o -iname '*.hp' -o -iname '*.hxx' -o -iname '*.hpp' -o -iname '*.HPP' -o -iname '*.h++' -o -iname '*.tcc' -o -iname '*.c'   -o -iname '*.h'
```
gbtitus: `SRC_FILES` is a pretty general name and I can imagine some poor naive soul using it for another purpose in such a way that it conflicts here.  Maybe something more descriptive, like `REBUILD_DEP_FILENAMES`?

ronawho: Renamed to SRC_FILES_FILTER

third-party/Makefile
ronawho: Maybe, but that seems like a lot of complexity to add. Any thoughts on a clean way to achieve that?

ronawho:
>One concern is: how much time does it take to do these find commands? If it's just seconds that's no problem, but if it's a minute that will be kind of a pain. (And will it be worse if and when we build LLVM from source and that has to be included here?)

Yeah, that was a big concern for me too. In the worst case (that i've tested) it adds 6 seconds. That was on an 8 year old dual core dev box. On chapcs nfs it takes 4 seconds, and on my mac it takes under 2 seconds. This extra time is most noticable when you're doing a top-level "no-op" build (a `make` where nothing has changed), but that already takes 6 seconds on my mac, so the overhead didn't seem like that much. 

>The other concern is that the list of source file extensions seems like a source (heh) of obscure problems. If a package adds a new source file extension and that list doesn't get updated, and then a file with that extension changes, we won't rebuild when we ought to. I don't know how likely that is, but if it ever does happen it could be a difficult to figure out what's gone wrong.

That's a list of all (I think) legal C/C++ file extensions. It's based off of [look_for_alloc.py:: find_source_files()](https://github.com/chapel-lang/chapel/blob/master/util/devel/look_for_alloc.py#L69-L77), which I put together based gccs list of valid filename extensions. Our third-party code only uses the "common" subset

ronawho:
Oh and the `:=` ensures that we only do the find command once per make invocation, so we won't be executing the find commands multiple times

third-party/Makefile
gbtitus: Something like this is what I had in mind, just building on the logic that's already there:

    QTHREAD_ALL_DEPS = $(QTHREAD_DEPEND)
    ifeq ($(CHPL_MAKE_JEMALLOC), jemalloc)
      QTHREAD_ALL_DEPS += $(JEMALLOC_DEPEND)
      QTHREAD_DEPEND_INSTALL_DIRS += $(JEMALLOC_INSTALL_DIR)
    endif
    ifeq ($(CHPL_MAKE_HWLOC), hwloc)
      QTHREAD_ALL_DEPS += $(HWLOC_DEPEND)
      QTHREAD_DEPEND_INSTALL_DIRS += $(HWLOC_INSTALL_DIR)
    endif
    ...
    $(QTHREAD_INSTALL_DIR): $(QTHREAD_ALL_DEPS)
    ...

gbtitus:
I like 99ccc53a!   👍

third-party/Makefile
ronawho: I'll probably punt on this --

The expensive part is the actual `find` command, which we'll do regardless. Make seems to be pretty fast at computing dependencies. 

And if somebody is using qthreads, 99% of the time (my complete fabrication/guess) they'll be using hwloc and likely jemalloc. I imagine most users are either using the quickstart configuration or the default configuration, not some mix of qthreads and no hwloc.

What I thought you were suggesting was conditionalizing the find commands.. which make we could do now that the find_3p_deps is function-ified?

ronawho: i.e. we could do something like:


```make
find_3p_depend = $(shell if [ '$(3)' = '$(4)' ]; then find $(1) $(SRC_FILES_FILTER) 2>/dev/null; fi) $(2)/Makefile* $(2)/README
AMUDPRUN_DEPEND := $(call find_3p_depend, $(AMUDPRUN_SRC_DIR), $(AMUDPRUN_DIR), $(CHPL_MAKE_LAUNCHER), amudprun)
```

or 

```make
ifeq ($(CHPL_MAKE_LAUNCHER), amudprun)
AMUDPRUN_DEPEND := $(call find_3p_depend, $(AMUDPRUN_SRC_DIR),$(AMUDPRUN_DIR))
endif
```

gbtitus: I was thinking that the find commands were already conditionalized, but they're not.  We do them in every third-party subdir, even the ones for packages we're not going to build.  I'd be inclined to make those conditional.

gbtitus: Yeah, I'd do the latter, checking the condition before firing up a shell rather than in the shell itself.

ronawho: Ok, do you need that to be part of this PR, or can that be a "if people notice a slowdown, fix it" sort of thing? My (relatively strong) preference would be to wait and see because of the impact making them conditional will have on the readability/maintainability of this code

gbtitus: Yeah, I think it'd be okay to wait.

PR #7006, opened 2017-08-14
Fix perf graph generation when --perflabel "label" is used

ronawho:
start_test always used $CHPL_HOME/test/GRAPHFILES as the graphfile list even if
a perflabel was being used. We never noticed this before because we manually
call gengraphs for the `--perflabel ml-` graphs.

@mppf stumbled on this while trying to add a "cc-" label in #6604

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -519,7 +519,11 @@ def compiler_performance():
 
 
 def generate_graph_files_graphs():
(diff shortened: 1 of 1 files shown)


mppf:
Solves my problem, thanks!

PR #7007, opened 2017-08-14
Add test for bug with initializing an extern C array

benharsh:


benharsh:
@mppf, @lydia-duncan , and @noakesmichael : any thoughts on what should happen here? Do we need a different type for an extern C array?

lydia-duncan:
That's fun.  I'm afraid I don't have a good suggestion, though :(

mppf:
@benharsh - this is with the default constructor = old style initializer. I can think of several strategies here:
 1. Mark wrapper's myBuffer field as "no init" with say myBuffer = noinit; in the declaration.
 2. Use a tuple to represent the fixed-sized array instead of the c_ptr (this is how it is done in GMP.chpl)
 3. Use an opaque type for myBuffer instead of an extern record, and then use void ptr/memcpy to work with it.

It's a bit of a lie to declare the extern record myBuffer with a c_ptr field, so I don't think it's surprising that it's an error.

benharsh:
OK, so should I remove this future?

And should we have a helper proc like this for C arrays?
```chpl
proc cArray(type eltType) type {
  return 1*eltType;
}
```

mppf:
Did you mean your helper proc to have an int argument? C arrays don't necessarily have a size in them and they're pretty different if they do...

PR #7009, opened 2017-08-14
Make this test a .future to pass nightly testing

mppf:
Continuing PR #6962.


mppf:
@coodie - could you have a look at this test? Perhaps in LLVM 4.0 we need to pass an additional option to the loop vectorizer to disable vectorization? Alternatively, we might decide this test isn't that valuable. Please advise.

Issue #6854, opened 2017-07-27
error handling: groups of errors


mppf:
[CHIP 8](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/8.rst) includes the idea that tasks can raise errors, and if they do, the errors will be propagated to the parent task at a *task join point* (such as a the end of a `sync`, `cobegin`, or `coforall` statement).

Let's consider a `coforall` for a minute. A `coforall` could create an error for each task - and so the *task join point* might have to handle many errors. How do we represent these errors?

CHIP 8 describes an Error subtype called `AsyncErrors`. 

My implementation work has started with an error subtype called `ErrorGroup`. Either way, we need to decide upon the API for this collection of errors and also how they behave with `catch` blocks.

So, here are the questions I'd like to answer in this issue:

 1. What is the name of the Error subtype that collects multiple errors? (currently `ErrorGroup`)
 2. If there happened to be just one error thrown from a sub-task, should we raise that error directly instead of raising an `ErrorGroup` containing a single error? (currently: no)
 3. If there happened to be just one class of error (e.g. `IoError`), should a catch block for that type of error catch the entire list? And then each `Error` subtype would have a way to enumerate multiple errors encountered? (currently: no)
 4. When nesting is used, as with a `coforall` within a `coforall`, does that result in an `ErrorGroup` containing an `ErrorGroup`? Or a single `ErrorGroup` containing all of the errors? (currently: no)
 5. Should we provide methods on `ErrorGroup` that
    a) indicate if a particular type of error is present in the list? (currently: yes)
    b) allow retrieving the errors by type? (currently: yes)
 6. Should printing out an `ErrorGroup` with `writeln` e.g.
     a) Print out every contained error? (currently: no)
     b) Print out some kind of summary of the contained errors? (currently: yes)
     c) Print out a fixed number of errors (using numbers to indicate how much wasn't printed)? (currently: yes)
     d) Print out one error of each error type encountered? (currenly: no)
 7. Based upon the above, what should be the user-facing API for `ErrorGroup` ?
      Currently, we have
      ``` chapel
      class ErrorGroup : Error {
        /* yield the errors contained in this group */
        iter these();
        /* Write a summary of the errors contained in this group.
           The summary attempts to use limited space. In particular,
           it prints only a constant number of errors, using
           numbers to indicate how many other errors there are.
         */
        proc writeThis(f);
        /* Iterate only errors that are a subtype of the passed type */
        iter filter(type t);
[...]

Issue #6853, opened 2017-07-27
error handling and sync blocks


mppf:
If you have a 'sync' block, it's supposed to track errors created by any task that might have been spawned within the 'sync' block and 'throw' them at the conclusion of the 'sync' block.

That's fine, but how does the compiler know if error handling was required?

E.g.
``` chapel
sync {
  someFunction();
}
```

Since tasks created in `begin` statements within `someFunction` might throw, `sync` might also throw. Should the user be required to handle the potential for errors from a `sync` in strict error handling mode?

I can see these options:
 1. The compiler assumes that `sync` can `throw` and user's just live with that
 2. The compiler tries to do analysis to find the `sync` blocks that should `throw`. Note that this would complicate separate compilation because the compiler has to know both
 
     a) which functions can (transitively) create use `begin`
     b) which of those `begin` blocks can create tasks that might throw
 3. `sync` can `throw` but the user isn't required to handle it. (a.k.a. "unchecked throws").

mppf:
Based upon discussions with @psahabu and @bradcray, I'm initially trying to implement option 1 above.

mppf:
We discussed this in a meeting but some team members had reservations about the approach.

lydia-duncan:
My understanding of the current approach is that sync blocks must always be either in a throwing function or surrounded by a `try` block in strict/pedantic mode.

This makes me uneasy, but that might stem from a misunderstanding of strict/pedantic mode as they stand today.

As an example, consider the following program (apologies if the syntax is incorrect):
```chapel
proc beginThatThrows () {
  begin functionThatThrows();
}

proc syncFn () {
  sync {
    beginThatThrows();
  }
  sync {
     functionThatDoesNotThrow();
  }
}
```
My understanding of the current proposal is that there is a mode where both `sync` blocks would be marked as needing a `try` or `try!`.  My instinct is that we would only want to require that of the first `sync` block, and only if the functions it calls are marked `throws`.  But I also think that if a function has a `begin` block that might throw and is not surrounded by a `sync` and error handling, it would have to be marked as `throws` since the throwing call might outlive the scope of the function.  I think it would be inappropriate and confusing for a `sync` block to be treated differently than the functions it called in this manner.

Is that consistent with the current strategy?  Is there an issue with my understanding of `begin`/`sync`?

mppf:
> My instinct is that we would only want to require that of the first sync block, and only if the functions it calls are marked throws.

[...]

PR #7019, opened 2017-08-15
Support throwing errors across tasks

mppf:
This PR adds support for throwing an error in a begin, cobegin, or coforall and then passing that error to the place where that task joins with the parent task (end of sync block for begin, end of cobegin/coforall statement for the other two).

These bullets describe the overall strategy:
 * store errors created by tasks in a task-safe list of errors. This list is stored in the EndCount (since that EndCount is available to the code that waits for the tasks so it's convenient to propagate the errors; and also since the EndCount is known to the compiler and so can be adjusted as necessary in parallel.cpp).
 * when catching errors from a coforall/cobegin/sync, the errors from child tasks are raised in an ErrorGroup which can contain multiple errors. Printing out one of these attempts to summarize the errors. There are convenience methods for filtering particular types of errors out. And, of course, the included errors can be enumerated.
 * A `cobegin` or `coforall` will count as `throws` if the code inside could throw. In contrast, since we can't easily determine what fire-and-forget tasks were launched (and which of those could throw), all `sync` blocks are marked `throws`.

Here are a few more notes on the implementation strategy:
 * errorHandling.cpp split CanThrowVisitor into CanThrowVisitor and ErrorCheckingVisitor in preparation for improving CanThrowVisitor
 * Added FLAG_UNCHECKED_THROWS to mean a function can throw but we shouldn't insist the error is caught.
 * Added FLAG_ERROR_VARIABLE to aid in adjustments that parallel.cpp makes to propagate errors.
 * downEndCount now takes in an error argument (to add to the list of errors from spawned tasks that is stored in the EndCount). parallel.cpp now adjusts the downEndCount calls to pass any uncaught error encountered in the task.

Related issues:
 * sync blocks throw: issue #6853.
 * Is it ErrorGroup or something else? What methods should be available? Discussed in issue  #6854.

Testing:

 - test/errhandling passes
 - Passes full local testing except for some tests that don't correctly print out the correct file name / line number on a halt. This is fixed by improvements to lowerIterators in the follow-on PR - in which these changes pass full local testing.

Reviewed by @psahabu, @ronawho, and @vasslitvinov (in PR #7036) - thanks!


mppf:
@psahabu and @ronawho - could you help review this PR? In particular, I'm expecting Preston to look at the error-handling approach, semantics, ChapelError.chpl changes, and errorHandling.cpp changes. Elliot, I'd like you to help review the ChapelBase.chpl, build.cpp, and parallel.cpp changes. Thanks!

compiler/passes/errorHandling.cpp
```diff
@@ -546,30 +668,48 @@ bool CanThrowVisitor::enterCallExpr(CallExpr* node) {
...
+    (*reasons)[fn] = visit.reason();
+    fn->throwsErrorInit();
+
+    if (visit.unchecked())
+      fn->addFlag(FLAG_UNCHECKED_THROWS);
+  }
 }
 
-static void checkErrorHandling(FnSymbol* fn)
+static void checkErrorHandling(FnSymbol* fn, implicitThrowsReasons_t* reasons)
```
psahabu: Same as `checkImplicitThrows`, maybe `checkErrorHandling` makes more sense as a static method of `ErrorCheckingVisitor`.

compiler/passes/errorHandling.cpp
```diff
@@ -603,3 +743,59 @@ bool isCheckErrorStmt(Expr* e)
...
+      outError = new ArgSymbol(INTENT_REF, "error_out", dtError);
+      outError->addFlag(FLAG_ERROR_VARIABLE);
+      fn->insertFormalAtTail(outError);
+
+      epilogue = fn->getOrCreateEpilogueLabel();
+      INT_ASSERT(epilogue); // throws requires an epilogue
+    }
+
+    ErrorHandlingVisitor visitor = ErrorHandlingVisitor(outError, epilogue);
+    fn->accept(&visitor);
```
psahabu: I suppose by my `checkImplicitThrows` logic, lines 784 to 799 should be made into a static method of `ErrorHandlingVisitor`. I'm not sure that'd be best, but it'd make `lowerErrorHandling` aesthetically pleasing with three single-statement loops.

mppf:
The test with the halt filename issue are:
```
arrays/bradc/badParZip.chpl
arrays/deitz/parallelism/stream/test_stream_is_parallel.chpl
arrays/deitz/parallelism/stream/test_stream_no_fast_follow.chpl
arrays/deitz/parallelism/stream/test_whole_array_stream_is_parallel.chpl
arrays/deitz/parallelism/stream/test_whole_array_stream_no_fast_followers.chpl
arrays/deitz/parallelism/test_whole_array_assignment_is_parallel.chpl
arrays/deitz/parallelism/test_whole_array_assignment_of_array_promotion_is_parallel.chpl
```

compiler/passes/errorHandling.cpp
```diff
@@ -546,30 +668,48 @@ bool CanThrowVisitor::enterCallExpr(CallExpr* node) {
...
 
-
 // Returns `true` if a block can exit with an error
 //  (e.g. by calling 'throw' or a throwing function,
 //   when these are not handled by try! or catch).
 // This function is useful to infer 'throws' for
 // certain compiler-introduced functions.
 
-static bool canBlockThrow(BlockStmt* block)
+static void markImplicitThrows(FnSymbol* fn, std::set<FnSymbol*>* visited, implicitThrowsReasons_t* reasons)
```
mppf: I'd prefer to keep it and checkErrorHandling as functions since the fact that they are implemented with the various visitors seems to be to be an implementation point; where these functions are an internal API of sorts.

psahabu: Makes sense to me.

psahabu:
The new changes look good, thanks for the comments and refactoring.

PR #7021, opened 2017-08-15
Fix memory annotations that used # within description

bradcray:
Something about the use of a # in my description of some leak changes
seems to have truncated the description in the online graphs, so
I've reworded to avoid the hash in hopes that that helps.

### DIFF
--- a/test/ANNOTATIONS.yaml
+++ b/test/ANNOTATIONS.yaml
@@ -692,7 +692,7 @@ memleaks:
   07/14/17:
     - Have reindex domains/distributions preserve locality (#5757)
   07/25/17:
-    - fix leaks introduced by reindex views in #5757 (#6776)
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho : Would you double-check me on this?  (that you see the truncation as well, that this seems like a reasonable workaround)

ronawho:
Whoops. Yeah, I see the truncation too. `<space>#` is a .yaml comment, so that's being stripped out (whereas something like` (#5210)` isn't because there's no space before the #)

This seems like a good workaround, or you could quote the line to be:

```
    - "fix leaks introduced by reindex views in #5757 (#6776)"
```

bradcray:
Thanks for the other workaround.  Since the # notation isn't clickable in this context anyway, I'm going to stick with what I've got for simplicity / sufficient clarity.

PR #7022, opened 2017-08-15
Fix compflags/diten/savec.chpl test for CHPL_GASNET_SEGMENT=fast

daviditen:
Any use of the extern function 'system' causes seg faults for gasnet-fast.
Change 'system' calls into calls to a function that uses the 'Spawn' module to
do something equivalent. Ran the modified test 100 times with gasnet-fast.

Add a simple .future that only runs for gasnet-fast that calls 'system' and
crashes.


ben-albrecht:
> Any use of the extern function 'system' causes seg faults for gasnet-fast.

That's interesting - any idea why?

test/compflags/diten/savec.chpl
```diff
@@ -1,26 +1,30 @@
 const outdir = "savec_output";
 const filename = "savec.chpl";
 
-var ret: int;
+proc mysystem(cmd: string): int {
+  use Spawn;
+  var sub = spawnshell(cmd);
```
ben-albrecht: I generally prefer `spawn()`, but since no user input is involved here, it doesn't really matter.

daviditen:
No, I don't know why.  I spent some time looking at it in the debugger and looking at the generated code, but there isn't much to see without a debug version of the libc.  Here's the backtrace from where it crashes:

```
#0  0x00007fcdaa98b280 in ?? ()
#1  0x00007fcdaa9d91c0 in ?? ()
#2  0x00007fcdaba00d6c in ?? ()
#3  0x00007fcdaba00c00 in ?? ()
#4  0x00007ffff6be8ffd in do_system () from /lib64/libc.so.6
#5  0x00007ffff6be907c in system () from /lib64/libc.so.6
#6  0x000000000044c925 in chpl__init_externSystem (_ln_chpl=1, _fn_chpl=53)
    at output/externSystem.c:48
#7  0x000000000044cb9a in chpl_gen_main (_arg=0x82b4d0 <chpl_gen_main_arg>)
    at output/externSystem.c:121
#8  0x000000000047cff5 in chpl_executable_init ()
#9  0x00000000004815c2 in main_wrapper ()
#10 0x000000000053c2b8 in qthread_wrapper ()
#11 0x0000000000000000 in ?? ()
```

daviditen:
Thanks!

PR #7024, opened 2017-08-15
Fix bug with initializers on types that inherit from generic parents

lydia-duncan:
Prior to this change, we would get an internal error complaining that the super
field was generic when we tried to resolve `super.init()` calls on classes
that inherited from a generic parent class.  This changes allows those
initializers to compile and run correctly.

Details:
--------

AggregateType.h,
AggregateType.cpp:
Extend setNextGenericField to also recursively set the index of the first
generic field of the parent type (as this cannot happen upon creation of the
parent type).  Add a new method getInstantiationParent, which takes in the
instantiated parent type and provides an instantiation for the child which takes
it into account.

symbol.h,
symbol.cpp:
Add a new method renameInstantiatedFromSuper to TypeSymbols, which renames the
current type based on the renaming that has occurred to the parent type.

InitNormalize.h,
InitNormalize.cpp:
Add a new method transformSuperInit, which replaces the old AST for a
`super.init()` call with a PRIM_GET_MEMBER_VALUE, since the old AST resulted in
a call to the accessor for super and that call would never determine the
appropriate instantiation (also, methods aren't allowed in Phase 1).  Call that
method when we complete Phase 1.  Move the declarations of a couple of helper
functions higher up, so that we can use them in this method.  Also, when we
"skip" the first field in the initialization of the InitNormalize object, mark
it with the FLAG_DELAY_GENERIC_EXPANSION flag if the parent type is generic.

functionResolution.cpp:
If we are moving a GET_MEMBER_VALUE call into a temporary, propagate the
FLAG_DELAY_GENERIC_EXPANSION flag when present.  Make temporaryInitializerFixup
operate on the valType of the argument to the initializer.  When we back-update
the temporary for an initializer call, if the temporary is a temporary for the
super field, propagate the new type to the super for this method.

initializerResolution.cpp:
Only internal error after the call to setNextGenericField if the type wasn't
updated and the parent type wasn't generic.

Testing:
----------
Had to fix the .good files for the tests involving generic vars (because I forgot
the generic var fields, whoops!).  Otherwise, removed seven .future files for now
passing tests.

Passed full standard paratest with futures (modulo current failures on master)


compiler/resolution/functionResolution.cpp
```diff
@@ -6082,6 +6082,25 @@ resolveExpr(Expr* expr) {
           // delayed instance to have the right type.
           actualSym->type = formal->type;
           actualSym->removeFlag(FLAG_DELAY_GENERIC_EXPANSION);
+
```
noakesmichael: I am just beginning to get my head around this but I think I noticed something I overlooked when
this stuff started to go in.  I wonder if the loop that this update is embedded in terminate with
i <= call->numActuals() rather than i < call->numActuals()?

noakesmichael: Hmmm.  Further consideration of call->numActuals() leads me to be more confused than I was
before.  I am going to keep working on this but I suspect we'll want to thrash this out face to face
rather than by email

lydia-duncan: Hmm, good point!  I forgot about that, since it likes 1-based indexing

PR #7026, opened 2017-08-15
Reduce the comm=ugni default heap size and make other improvements.

gbtitus:
With the advent of dynamic array registration in comm=ugni we no longer
need as large a default for the heap size.  Here, reduce it to 16 gb or
as much as the NIC TLB (Gemini) or NIC TLB cache (Aries) can cover,
whichever is smaller.  Also, if the user specifies a heap size larger
than the NIC TLB can cover, improve our messaging and print that only
from one node (since the situation will be identical on all), and only
reduce that heap size when we're using a Gemini NIC.  When we're using
an Aries NIC, allow such over-large heaps and just warn that NIC TLB
refills might reduce performance.


gbtitus:
@ronawho, care to take a look?

PR #7027, opened 2017-08-15
Atomic warning fix

mppf:
I noticed the warning you get when using an atomic statement refers to the "atomic keyword" but now that atomic types are supported it should refer specifically to the "atomic statement". This PR makes this simple wording change in the error message.

Passed full local testing.


bradcray:
It seems to me this feature has been deprecated so long that we could just pull the code and tests in question rather than wordsmithing it?

bradcray:
Ironically, I believe the thing that blocked it from working last time it was pursued was a portable implementation of atomics, which we now have.  :(  I've often wondered how hard it would be to resurrect that branch (not that I think it's anything like a priority at present).

All that said, I think it's sufficiently speculative and far-off in our Chapel plans that removing it from the parser makes the most sense (not that I'm suggesting you personally need to do this).  E.g., partial reductions are much more on our planned support path, yet even they are not supported by the parser yet.  Keeping them in the parser seems likely only to add confusion and potentially lead to false assumptions.

PR #7028, opened 2017-08-15
Enlarge registered memory tables and report if/when they overflow.

gbtitus:
In comm=ugni we record the registered memory regions in a fixed-size
table.  Here, greatly increase that table's size (from 100 entries to
1000) and add a warning emitted the first time we discover that we want
more entries but don't have any.  The ISx test, when run on compute
nodes with high core counts, was silently running out of such table
entries, dropping back to allocating arrays out of the regular heap, and
then running out of memory.  With this change it should no longer do
this, but if it or any other test overflows the table at least we'll
know that has happened.

Note that expanding the tables arguably makes it even more imperative
that we improve the scalability of the broadcast and search algorithms
used on them.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -378,7 +378,7 @@ static size_t hugepage_size;
 // uGNI memory domain handles for each memory region of interest to us.
 // mem_region_map contains a copy of every node's mem_regions.
 //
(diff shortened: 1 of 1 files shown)


gbtitus:
@ronawho, here's the other one we talked about.

gbtitus:
@benharsh, you might be interested in this one.

PR #7029, opened 2017-08-16
Improved forall intents

### LINKED ISSUE #6351: Passing variable via in intent to a forall loop seems to create an iteration-private variable, not a task-private one
### Summary of Problem

The following program reports "0 => 1" for all iterations.

The spec section 25.3 (p. 216) says

> The in intent is a way to obtain task-private variables in a forall loop.

A comment in issue #6325 suggests the same thing.  The behavior I see
isn't what I expect "task-private" to mean.

```
fortytwo@magrathea:~/src/chpl$ chp boop.chpl
here.maxTaskPar: 2
51: 0 => 1
52: 0 => 1
53: 0 => 1
54: 0 => 1
55: 0 => 1
1: 0 => 1
56: 0 => 1
...
45: 0 => 1
100: 0 => 1
46: 0 => 1
47: 0 => 1
48: 0 => 1
49: 0 => 1
50: 0 => 1
fortytwo@magrathea:~/src/chpl$ 
```

### Steps to Reproduce

**Source Code:**

```chapel
var x = 0;

writeln("here.maxTaskPar: ", here.maxTaskPar);

forall i in 1..100 with (in x) {
  var xtmp = x;
  x += 1;
  writeln(i, ": ", xtmp, " => ", x);
}
```

**Compile command:**
chpl boop.chpl -o boop

**Execution command:**
./boop -nl 1

None yet

### Configuration Information

```
fortytwo@magrathea:~/src/chpl$ chpl --version
chpl Version 1.16.0 pre-release (3c4a4d9)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)

fortytwo@magrathea:~/src/chpl$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
  CHPL_JEMALLOC: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none

fortytwo@magrathea:~/src/chpl$ gcc --version
```

vasslitvinov:
Improved handling of forall intents

### Representation and code improvements

* add FIinfo class to store intent information

* propagate the exact intent information to those parts of
  implementForallIntents.cpp that need it

* streamline the code, removing some unwanted special cases

### Semantics and generated code

* all VarSymbols in the loop body that represent shadow variables
  are now aliases, i.e. "refs", to the shadow variables created
  within the parallel iterator (which is extended to do so
  by extendLeaderNew()); this resolves #6351

* the outer variable is passed to the extendLeaderNew()-extended parallel
  iterator by the intent that corresponds to the forall intent
  i.e. by value for '[const] in' and by reference for '[const] ref'

* remove the warning "Arrays, domains, and distributions currently can be
  passed into the forall loop by 'const', 'ref' or 'const ref' intent only"

* fix several forall-intent futures
  (to be merged separately)

### Future work - code clarity

* We may want to make heavier use of FIinfo instead of passing
outerVars and shadowVars vectors around.

* Right now FIinfo::svar is very lightly used. Also we have both
FIinfo::reduceGVar and FIinfo::reduceMark for a reduce intent.
Some of these could be either removed or utilized more extensively.

  Note that outer vars and shadow vars are going to change as we
descend into task functions in extendLeaderNew / propagateExtraLeaderArgsNew.
Up to that point, however, we may want to fold outerVars and shadowVars
vectors into 'fivec'.

### Future work - semantics and AST

Do we want to continue on the path of handling ALL outer variables
uniformly w.r.t. passing them to the parallel iterator and yielding
them into the shadow variables?  For example, we do not need to
pass through the variables passed by 'ref' intent or 'const'
variables passed by 'const ref' intent.

* isOuterVarNew() has the check for FLAG_CONST. It excludes a
const variable from being handled as an outer variable.
This is incorrect when the variable is passed by 'in' intent.
Ex. test/parallel/forall/in-intents/both-arr-dom-const-var

  If I remove this check, a few tests break, ex.
arrays/bradc/workarounds/arrayOfArray
arrays/deitz/jacobi5*
arrays/deitz/matrix
Interestingly, they break in a seemingly-unrelated, much later
spot in the compiler.

  Maybe the thing to do is to remove this check, however
later in createShadowVarsNew() prune those 'const' variables
that are passed by 'const ref' intent. While there, may as well
complain if one is passed by 'ref' intent.

* createShadowVarsNew() also prunes variables for INTENT_REF
and INTENT_REF_MAYBE_CONST. I.e. we do not create shadow variables
for these two intents, so the body of the loop refers directly
to the outer variable. Is this what we want? as much as possible?

  We have to do at least something for 'const ref' intent.
In the case when the outer variable is not const, we want
constness checking in the body of the loop to see the shadow
variable as a const.

  Also, is relying on lexical scoping any different than
passing these through, in terms of the generated code?
They will end up being passed through the task functions anyway
when those task functions are flattened, right?

### Status

- [x] testing: linux64 --verify
- [x] testing: gasnet (an earlier version passed)
- [x] LCALS inner_prod slowdown - fix if needed (consult with David I)

### DIFF
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -2074,19 +2074,25 @@ static void pruneFIVec(FIIvec& fivec, int numShadowVars) {
 
 // Not to be invoked upon a reduce intent.
 static void setShadowVarFlagsNew(Symbol* ovar, VarSymbol* svar, IntentTag intent) {
+  // These do not make sense for task/forall intents.
+  INT_ASSERT(!(intent & (INTENT_FLAG_PARAM | INTENT_FLAG_TYPE)));
+  // If this assert fails, we need to handle this case.
+  INT_ASSERT(!(intent & INTENT_FLAG_OUT));
+
+  // All our shadow variables are expected to be refs.
+  INT_ASSERT(svar->qual == QUAL_REF);  //see below: svar->qual = QUAL_CONST_REF
+  svar->addFlag(FLAG_REF_VAR);
+
   if (intent & INTENT_FLAG_CONST) {
+    svar->qual = QUAL_CONST_REF;
     svar->addFlag(FLAG_CONST);
+    if (intent == INTENT_CONST_IN)
+      // Enables canForwardValue(), ex.
+      //   release/examples/benchmarks/hpcc/fft.chpl
+      svar->addFlag(FLAG_REF_TO_CONST);
     if (!ovar->isConstant())
       svar->addFlag(FLAG_CONST_DUE_TO_TASK_FORALL_INTENT);
   }
-
-  svar->addFlag(FLAG_REF_VAR);
-  INT_ASSERT(svar->qual == QUAL_REF); // because it has a ref type
-
-  // If this assert fails, we need to handle this case.
-  INT_ASSERT(!(intent & INTENT_FLAG_OUT));
-  // These do not make sense for task/forall intents.
-  INT_ASSERT(!(intent & (INTENT_FLAG_PARAM | INTENT_FLAG_TYPE)));
 }
 
 //
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -28,6 +28,7 @@
 #include "stlUtil.h"
 #include "stringutil.h"
 #include "visibleFunctions.h"
+#include <utility>
 
 //
 //-----------------------------------------------------------------------------
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
 // for the (new) ForallStmt-based representation.
 //
 
+// Forall Intent information about each outer variable.
+class FIinfo {
+public:
+  Symbol* ovar;       // outer variable
+  Symbol* svar;       // shadow variable
+  Symbol* reduceGVar; // reduction class, non-NULL only for reduce intents
+  Symbol* reduceMark; // transitional measure; TODO: remove
+  TFITag  intent;     // the forall intent for this variable
+
+  bool isReduce() { return intent == TFI_REDUCE; }
+
+  FIinfo(Symbol* outerVar):
+    ovar(outerVar), svar(NULL),
+    reduceGVar(NULL), reduceMark(NULL), intent(TFI_DEFAULT)
+  { }
+};
+
+/*
+We need to:
+  * Associate each outer variable with a unique FIinfo.
+  * Adjust FIinfo with user-specified intent for each variable in with-clause.
+  * Iterate over the variables in a deterministic order.
+
+We can use just std::map<Symbol*,FIinfo>.
+
+Or we can use std::vector<FIinfo>.
+Also, to ensure uniqueness and for processing the with-clause,
+use a temporary std::map.
+
+Given my fuzzy-warm feeling about traversing a vector, i.e. faster than
+traversing a map, I chose to use the vector + temporary map.
+*/
+
+typedef std::vector<FIinfo> FIIvec;
+typedef std::map<Symbol*,int> FIImap;  // map to the index into FIIvec
+
+// Add 'sym' to 'fivec' and 'fimap', if it is not there already.
+static void addFII(Symbol* sym, FIIvec& fivec, FIImap& fimap) {
+  FIinfo info(sym);
+
+  if (fimap.count(sym))
+    // already there
+    return;
+
+  fivec.push_back(info);
+  int newIndex = (int)fivec.size()-1;
+  fimap.insert(std::make_pair(sym,newIndex));
+}
+
+/////////////////////////////////////////////////////////////////////////////
+// TFITag <-> IntentTag
+
+static IntentTag tfi2it(TFITag tfi) {
+  switch (tfi) {
+  case TFI_DEFAULT:   return INTENT_BLANK;
+  case TFI_CONST:     return INTENT_CONST;
+  case TFI_IN:        return INTENT_IN;
+  case TFI_CONST_IN:  return INTENT_CONST_IN;
+  case TFI_REF:       return INTENT_REF;
+  case TFI_CONST_REF: return INTENT_CONST_REF;
+  case TFI_REDUCE:
+    INT_ASSERT(false);   // don't know what to return
+    return INTENT_BLANK; //dummy
+  }
+  INT_ASSERT(false);   // unexpected TFITag; 'tfi' contains garbage?
+  return INTENT_BLANK; // dummy
+}
+
+static TFITag it2tfi(IntentTag intent) {
+  switch (intent) {
+  case INTENT_IN:        return TFI_IN;
+  case INTENT_CONST:     return TFI_CONST;
+  case INTENT_CONST_IN:  return TFI_CONST_IN;
+  case INTENT_REF:       return TFI_REF;
+  case INTENT_CONST_REF: return TFI_CONST_REF;
+  case INTENT_BLANK:     return TFI_DEFAULT;
+  default:
+    INT_ASSERT(false);  // don't know what to return
+    return TFI_DEFAULT; // dummy
+  }
+  INT_ASSERT(false); // unexpected IntentTag; 'intent' contains garbage?
+  return TFI_DEFAULT; // dummy
+}
+
+static TFITag concreteTFI(TFITag origTFI, Type* type) {
+  if (origTFI == TFI_DEFAULT)
+    return it2tfi(concreteIntent(INTENT_BLANK, type->getValType()));
+
+  else if (origTFI == TFI_CONST)
+    return it2tfi(concreteIntent(INTENT_CONST, type->getValType()));
+
+  else
+    return origTFI;
+}
+
+/////////////////////////////////////////////////////////////////////////////
+// findOuterVarsNew() and helpers
+//
+
 //
 // A forall-intents variation on isOuterVar() in createTaskFunctions.cpp:
 // Is 'sym' defined outside 'block'?
@@ -1778,7 +1878,9 @@ static bool isFsIndexVar(ForallStmt* fs, Symbol* sym)
 // A forall-intents variation on findOuterVars() in createTaskFunctions.cpp:
 // Find all symbols used in 'block' and defined outside of it.
 //
-static void findOuterVarsNew(ForallStmt* fs, BlockStmt* block, SymbolMap& uses) {
+static void findOuterVarsNew(ForallStmt* fs, BlockStmt* block,
+                             FIIvec& fivec, FIImap& fimap)
+{
   std::vector<SymExpr*> symExprs;
 
   collectSymExprs(block, symExprs);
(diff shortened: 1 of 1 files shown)


mppf:
> Also, is relying on lexical scoping any different than
> passing these through, in terms of the generated code?
> They will end up being passed through the task functions anyway
> when those task functions are flattened, right?

That is my understanding.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+  Symbol* reduceMark; // transitional measure; TODO: remove
+  TFITag  intent;     // the forall intent for this variable
+
+  bool isReduce() { return intent == TFI_REDUCE; }
+
+  FIinfo(Symbol* outerVar):
+    ovar(outerVar), svar(NULL),
+    reduceGVar(NULL), reduceMark(NULL), intent(TFI_DEFAULT)
+  { }
+};
```
mppf: Why is FIinfo local to this file, vs being a full-on AST node with its own header file etc etc?
Also, can we please name it ForallIntentInfo ? Or even just ForallIntent ?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
 //
 
+// Forall Intent information about each outer variable.
+class FIinfo {
+public:
+  Symbol* ovar;       // outer variable
+  Symbol* svar;       // shadow variable
+  Symbol* reduceGVar; // reduction class, non-NULL only for reduce intents
+  Symbol* reduceMark; // transitional measure; TODO: remove
+  TFITag  intent;     // the forall intent for this variable
```
mppf: I wish it was not TFITag because I have no idea what TFI means. This comes up every time I see the type. Can we rename it?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+
+Given my fuzzy-warm feeling about traversing a vector, i.e. faster than
+traversing a map, I chose to use the vector + temporary map.
+*/
+
+typedef std::vector<FIinfo> FIIvec;
+typedef std::map<Symbol*,int> FIImap;  // map to the index into FIIvec
+
+// Add 'sym' to 'fivec' and 'fimap', if it is not there already.
+static void addFII(Symbol* sym, FIIvec& fivec, FIImap& fimap) {
```
mppf: addForallIntent

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+Or we can use std::vector<FIinfo>.
+Also, to ensure uniqueness and for processing the with-clause,
+use a temporary std::map.
+
+Given my fuzzy-warm feeling about traversing a vector, i.e. faster than
+traversing a map, I chose to use the vector + temporary map.
+*/
+
+typedef std::vector<FIinfo> FIIvec;
+typedef std::map<Symbol*,int> FIImap;  // map to the index into FIIvec
```
mppf: ForallIntentVec and ForallIntentMap

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+
+  fivec.push_back(info);
+  int newIndex = (int)fivec.size()-1;
+  fimap.insert(std::make_pair(sym,newIndex));
+}
+
+/////////////////////////////////////////////////////////////////////////////
+// TFITag <-> IntentTag
+
+static IntentTag tfi2it(TFITag tfi) {
```
mppf: Can you name this function in English? like getArgIntentForTaskIntent or something?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+  case TFI_CONST_REF: return INTENT_CONST_REF;
+  case TFI_REDUCE:
+    INT_ASSERT(false);   // don't know what to return
+    return INTENT_BLANK; //dummy
+  }
+  INT_ASSERT(false);   // unexpected TFITag; 'tfi' contains garbage?
+  return INTENT_BLANK; // dummy
+}
+
+static TFITag it2tfi(IntentTag intent) {
```
mppf: This name is also unreasonably abbreviated IMO.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1723,6 +1724,105 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+  case INTENT_BLANK:     return TFI_DEFAULT;
+  default:
+    INT_ASSERT(false);  // don't know what to return
+    return TFI_DEFAULT; // dummy
+  }
+  INT_ASSERT(false); // unexpected IntentTag; 'intent' contains garbage?
+  return TFI_DEFAULT; // dummy
+}
+
+static TFITag concreteTFI(TFITag origTFI, Type* type) {
```
mppf: concreteTaskIntent ?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1778,38 +1878,221 @@ static bool isFsIndexVar(ForallStmt* fs, Symbol* sym)
...
   collectSymExprs(block, symExprs);
 
   for_vector(SymExpr, symExpr, symExprs) {
     Symbol* sym = symExpr->symbol();
 
-    if (isLcnSymbol(sym))
-      if (!isFsIndexVar(fs, sym) && isOuterVarNew(sym, block))
-        uses.put(sym, markUnspecified);
+    if (isLcnSymbol(sym) &&
+        // exclude a few things
```
mppf: this is a useless comment. I can see that things are being excluded! It would be more useful to comment each line with the intent.  In particular I'd appreciate a comment next to isLcnSymbol saying "ArgSymbol or VarSymbol". Also, it seems a little strange to me that `this` is excluded. What's the justification for that? Can't `this` be an outer variable? Say, with a method containing an inner function? I can certainly see that gMethodToken can be ignored for this purpose.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1778,38 +1878,221 @@ static bool isFsIndexVar(ForallStmt* fs, Symbol* sym)
...
+      VarSymbol* typeofTemp = newTemp("chpl__typeofET");
+      typeofTemp->addFlag(FLAG_TYPE_VARIABLE);
+      hld->insertAtTail(new DefExpr(typeofTemp));
+      hld->insertAtTail("'move'(%S,'typeof'(%S))", typeofTemp, fiVarSym);
+      eltTypeArg = new SymExpr(typeofTemp);
+    } else {
+      // Forego typeof(), use the type directly.
+      eltTypeArg = new SymExpr(eltType->symbol);
+    }
+  }
```
mppf: I can see that the above two were moved, but it's hard for me to tell if you also changed them. Such rearrangements are best done in separate commits so they don't distract from the other changes. For now, just let me know if you changed them.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: Have you looked at the definition of this type? That may clarify it for you.

I can rename it to TFIntentTag if it helps. I'd be reluctant to make it any longer.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: At the moment this is a local helper data structure. Hopefully it will be obviated by IntentVariables. The way it is now, it will not help to make it into an AST node.

I will rename it to ForallIntentInfo.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: Thinking about it some more... I'd be OK calling it "TaskIntentTag" if you prefer. The downside is that at the moment it has nothing to do with task intents, although in the future it may.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: Will do. "get" does not help in this case, so simply argIntentForTaskIntent.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: Sorry about that. No, just moved, no change, for insertGenerateAfter, setupRiGlobalOp, handleRISpec.

compiler/resolution/implementForallIntents.cpp
mppf: If it's currently only used for forall intents, it should be called ForallIntentTag.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1778,38 +1878,221 @@ static bool isFsIndexVar(ForallStmt* fs, Symbol* sym)
...
+    // What else can this be?
+    INT_FATAL(fs, "not implemented");
+  }
+
+  return result;
+}
+
+// Mark the variables listed in 'with' clauses, if any, with tiMark markers.
+// Same as markOuterVarsWithIntents() in createTaskFunctions.cpp,
+// except uses ForallIntents.
```
mppf: When do you plan to remove the old implementation? It this PR getting us any closer to doing that?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1778,38 +1878,221 @@ static bool isFsIndexVar(ForallStmt* fs, Symbol* sym)
...
+  // do the same as in 'if (needsCapture(fn))' in createTaskFunctions()
+  findOuterVarsNew(fs, body, fivec, fimap);
+  markOuterVarsWithIntentsNew(fs, fivec, fimap);
+
+}
+
+/////////////////////////////////////////////////////////////////////////////
+
+// We could do it as-we-go in createShadowVarsNew(), it may be more complex.
+static void pruneFIVec(FIIvec& fivec, int numShadowVars) {
```
mppf: pruneForallIntentVec ?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1819,88 +2102,46 @@ static void setShadowVarFlagsNew(Symbol* ovar, VarSymbol* svar, IntentTag intent
 // so an ordered traversal of the two lists gives matching pairs.
 // Count the number of these variables into 'numOuterVars'.
 //
-static void createShadowVarsNew(Expr* refNode, SymbolMap& uses,
-                             int& numShadowVars, int& totOuterVars,
+static void createShadowVarsNew(Expr* refNode, FIIvec& fivec,
```
mppf: I got to here in my most recent review attempt. I'll have to return to it again.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: We need to convert other forall loops to the new representation
to retire the old implementation.
I do not have a specific date set for this work.
This PR does not address it.

compiler/resolution/implementForallIntents.cpp
vasslitvinov: +1

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1819,88 +2102,46 @@ static void setShadowVarFlagsNew(Symbol* ovar, VarSymbol* svar, IntentTag intent
...
 
       if (tiIntent == INTENT_REF || tiIntent == INTENT_REF_MAYBE_CONST) {
-        // for efficiency
-        pruneit = true;
-
-      } else if (isMethodToken) {
-        // If MT is present, _toLeader(..., _mt...) does not get resolved.
-        // See e.g. parallel/taskPar/figueroa/taskParallel.chpl
-        pruneit = true;
+        // do we want this? does this lead to more efficient generated code?
```
mppf: This is where the question of forall intents vs. handling in flatten functions comes in. It's not obvious to me if it's better to leave this case to flatten functions but I lean a little bit towards having the forall intent code handle all captured variables. But, it's certainly the case that flatten functions is going to run anyway... Perhaps we want to outline these functions early? Perhaps forall intent processing is an early sort of flatten functions?

Anyway it might be less of a change to leave it this way.

Regarding the comment, I think it'd be worth directly saying that by pruning the outer variable reference, we leave it for flatten functions to handle. Unless that's wrong and I'm misunderstanding.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1910,27 +2151,18 @@ static void createShadowVarsNew(Expr* refNode, SymbolMap& uses,
...
-                 tiMarker ? intentDescrString(tiMarker->intent) : "default intent");
-
-      } else if (isRecordWrappedType(ovar->type) &&
-                 !(tiIntent & INTENT_FLAG_REF)) {
-        // Threading through the leader for non-ref intents
-        // may not work correctly for arrays/domains, so we avoid it.
-        tiIntent = (tiIntent & INTENT_FLAG_CONST) ?
-          INTENT_CONST_REF : INTENT_REF;
-        USR_WARN(refNode, "Arrays, domains, and distributions currently can be passed into the forall loop by 'const', 'ref' or 'const ref' intent only. '%s' will be passed by %s.", ovar->name, intentDescrString(tiIntent));
+        USR_WARN(refNode, "sync, single, or atomic var '%s' currently can be passed into the forall loop by 'ref' intent only - %s is ignored", ovar->name, intentDescrString(tiIntent));
```
mppf: I'm confused - it looks like this code is in a if isAtomicType but the error message is about sync/single/atomic. Am I missing something? Or could/should we update the error message to be only about atomic.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1940,88 +2172,77 @@ static void createShadowVarsNew(Expr* refNode, SymbolMap& uses,
...
-    reduceGVars.push_back(globalOp);
-
-    e->value = svar;
+    info.reduceGVar = globalOp;
+    info.svar = svar;
 
     numShadowVars++;
   }
+
+  if ((int)fivec.size() != numShadowVars)
```
mppf: I got this far, will continue.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -28,6 +28,7 @@
 #include "stlUtil.h"
 #include "stringutil.h"
 #include "visibleFunctions.h"
+#include <utility>
```
mppf: The comment below:
> * does some AST transformations

Really? As if that explains anything...

mppf: It came from PR #614.

vasslitvinov:
I have follow-up work that supersedes this PR, so I prefer not to merge this.

PR #7032, opened 2017-08-16
Prevent GMP from modifying gmp.info* files

ronawho:
Because we store gmp untarred, differing timestamps can cause gmp to mistakenly
recreate some gmp.info doc files. To avoid this, just touch the gmp.info files
to prevent autoconf from trying to recreate them. This is similar to what we do
to prevent hwloc from rebuilding some doxygen files (added in 19a6dbb6af)


gbtitus:
Superb!   👍

PR #7033, opened 2017-08-16
Upgrade from hwloc 1.11.6 to 1.11.7

ronawho:
This release mostly just includes bugfixes/improvements to command lines tools
that we don't use. I don't think there's anything we need, but I wanted to keep
up to date with versions to catch any breaking changes sooner.

Full a full changelog see:
  https://mail-archive.com/hwloc-announce@lists.open-mpi.org/msg00098.html


ronawho:
@gbtitus probably not much to review here, so just more of an fyi

Issue #6979, opened 2017-08-10
Rename Barrier module


e-kayrakli:
### Summary of Problem
The `Barrier` module has a record named `Barrier` and its constructor `proc Barrier`. This creates major issues as can be seen in the source code sample. Although this brings up the possibility of user writing such a module and trying to use it in this fashion, I think the Barrier module should be renamed until the fundamental issue is fixed in a more principled way.

### Steps to Reproduce

**Source Code:**

```chapel
module Foo {
  use Barrier;
}
use Foo;
use Barrier;
```

**Compile command:**
`chpl foo.chpl`

Results in:

```
foo.chpl:2: error: 'use' of non-module/enum symbol Barrier
$CHPL_HOME/modules/standard/Barrier.chpl:68: error: Definition of symbol Barrier
```

### Configuration Information

- Output of `chpl --version`:
`chpl Version 1.16.0 pre-release (a9493a6b17)`

e-kayrakli:
@lydia-duncan helped me figure out the issue offline.

It appears that the issue happens when there is even only a class/record with the same name as the module but it can be worked around as such:

```chapel
module Foo {
  use Barrier only Barrier as MyBarrier;
}
use Foo;
use Barrier;
```

Note that this fix doesn't help as `use Barrier` also brings in `proc Barrier`.

lydia-duncan:
I mentioned to Engin that this behavior is currently intended, as Chapel is a single namespace language, so we would have conflicts for a Barrier module and class with the same reachability at the point where the use statement was made.  It is possible that we should rethink whether use statements even consider symbols that are not enums or modules, but I agree that for now the best path forward is to rename the enclosing module

bradcray:
How about renaming the record type to `barrier` (lowercase)?  Tagging @daviditen who I think wrote this code (?).

e-kayrakli:
> How about renaming the record type to barrier (lowercase)? 

I like this idea as it would be similar to `module List` / `record list`. However, wouldn't it create a conflict btw the actual `barrier` method and the constructor?

bradcray:
That's a good point.  Though if we rewrote the barrier type to use initializers rather than constructors, we'd probably be back on firm ground again (and I suspect that initializers are far enough along that this could be done today).

daviditen:
Switching the record to be named `barrier` and changing the constructor to an initializer, we still get into trouble with the `barrier()` method because the compiler thinks it's a constructor:

[...]

PR #7035, opened 2017-08-16
Change the name of the 'Barrier' module to the plural 'Barriers'

daviditen:
Previously, the `Barrier` module and the `Barrier` record shared a name, which
could lead to problems, as discussed in issue #6979. Update the name of the
module to `Barriers`.

One other idea was to change the name of the record to all lowercase `barrier`,
and change the constructor to an `init` function, but then the `barrier`
method looked like a constructor, so would also need to change names.
Updating the module name appears to be a lower impact change.

The `Barrier` module can still be used by its old name, but now generates
a deprecated warning.

Update tests to use the new module name.  Added a test for the deprecated
warning.

Passed a regular Linux64 paratest.

### DIFF
--- a/modules/standard/Barrier.chpl
+++ b/modules/standard/Barrier.chpl
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* Deprecated. Use the module :mod:`Barriers` instead. */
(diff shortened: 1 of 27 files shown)


modules/standard/Barrier.chpl
```diff
@@ -17,349 +17,9 @@
...
-   In the future, we expect to add more language-level support for a
-   "task-team" concept.  A task-team will more directly support collective
-   operations such as barriers between the tasks within a team.
-
-   The current implementation is designed for correctness, but is not expected
-   to perform well at scale.  We expect performance at scale to improve as
-   this barrier implementation is optimized and as the task-team concept is
-   implemented and optimized.
-*/
+/* Deprecated. Use the module :mod:`Barriers` instead. */
```
ben-albrecht: Should we remove this from the documentation or do we tend to leave the deprecated module documented for a release?

daviditen: IMO if the code is there it should be documented, even if the documentation is as simple as saying "Don't use this - use this other thing instead."

daviditen:
Thanks @ben-albrecht and @vasslitvinov.

modules/standard/Barrier.chpl
mppf: I think it's reasonable to have a comment here but I don't think we should be putting deprecated modules into the chpldoc'd set & on to the web site. I think that'll just confuse new users.

ben-albrecht: Note that we removed the documentation of `LayoutCSR.chpl` from the public docs, so we should resolve this inconsistency in how we document deprecated modules. If this isn't a well-defined protocol (which it doesn't appear to be), then we should get input from the broader team and try to define a protocol.

PR #7036, opened 2017-08-16
Support throw in forall statements

mppf:
This PR adds support and tests for throw used in the context of forall statements.
This is a complex task because of the way in which lowerIterators works directly with task functions - including possibly inlining them.

The approach of this PR is to adjust lowerIterators to fix up the error handling constructs in a way that the follow-on parallel pass can understand. A reasonable alternative design would be to perform lowerIterators before the errorHandling pass - but that is impossible right now because lowerIterators needs to be after callDestructors and errorHandling needs to be before callDestructors.

Anyway, this PR modifies lowerIterators as follows:
 * Adjusts expandBodyForIteratorInline to more obviously track whether or not it is processing a task function and to include an argument indicating if the calling context should update the CallExpr to pass in a dummy error argument.
 * Add fixupErrorHandlingExits, which is called from expandBodyForIteratorInline. fixupErrorHandlingExits handles error-handling structures that are no longer legal because they refer outside of the function. These might be referring to goto targets outside of the current function or to formals outside of the current function. Either way, this function translates it to use the task error argument and exit the task function on an error. (The parallel pass will take care of adding any error saved in the error argument to a list of errors encountered in subtasks).
 * Add replaceErrorArgumentWithEnclosingError, which is called from expandIteratorInline. This function adjusts certain error-handling patterns to use the out error argument instead of an enclosing variable. That is needed because a task function in a forall iterator needs to send the error to the list of sub-task errors rather than trying to exit control flow otherwise.
 * Added other support routines to help these two new functions.

- [x] full local testing
- [x] passes Hellos and test/errhandling in a GASNet configuration 
- [x] passes Hellos and test/errhandling in a NUMA configuration

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -702,9 +702,8 @@ bool ErrorCheckingVisitor::enterCallExpr(CallExpr* node) {
 
 static void markImplicitThrows(FnSymbol* fn, std::set<FnSymbol*>* visited, implicitThrowsReasons_t* reasons)
 {
-  // Currently, only task functions can be implicitly throws.
-  // Future: include iterators and any task functions.
-  if (!isTaskFun(fn)) // TODO: for foralls && !fn->isIterator())
+  // Currently, only task functions or iterators can be implicitly throws.
+  if (!isTaskFun(fn) && !fn->isIterator())
     return;
 
   // If we already visited this function, don't visit it again.
--- a/compiler/passes/parallel.cpp
+++ b/compiler/passes/parallel.cpp
@@ -27,6 +27,7 @@
 #include "driver.h"
 #include "errorHandling.h"
 #include "expr.h"
+#include "errorHandling.h"
 #include "files.h"
 #include "optimizations.h"
 #include "resolution.h"
--- a/compiler/passes/parallel.cpp
+++ b/compiler/passes/parallel.cpp
@@ -27,7 +27,6 @@
 #include "driver.h"
 #include "errorHandling.h"
 #include "expr.h"
-#include "errorHandling.h"
 #include "files.h"
 #include "optimizations.h"
 #include "resolution.h"
--- a/compiler/resolution/lowerIterators.cpp
+++ b/compiler/resolution/lowerIterators.cpp
@@ -512,11 +512,22 @@ replaceIteratorFormalsWithIteratorFields(FnSymbol* iterator, Symbol* ic,
   }
 }
 
+static void replaceErrorFormalWithEnclosingError(SymExpr* se);
+
 static void
-replaceIteratorFormalsWithIteratorFields(FnSymbol* iterator, Symbol* ic,
-                                         std::vector<SymExpr*> & symExprs) {
+replaceIteratorFormals(FnSymbol* iterator, Symbol* ic,
+                       std::vector<SymExpr*> & symExprs) {
+  bool throws = iterator->throwsError();
+
   for_vector(SymExpr, se, symExprs) {
-    replaceIteratorFormalsWithIteratorFields(iterator, ic, se);
+    // For an inlined iterator, we don't want to track the error
+    // argument through the iterator class. Instead of setting
+    // a formal error argument, set an enclosing error variable.
+    if (throws)
+      replaceErrorFormalWithEnclosingError(se);
+    // if se was not replaced by the above call...
+    if (se->parentSymbol != NULL)
+      replaceIteratorFormalsWithIteratorFields(iterator, ic, se);
   }
 }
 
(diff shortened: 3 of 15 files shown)


mppf:
@vasslitvinov - I'm still in the process of testing this PR but I hope you will look at it soon & offer feedback. It mostly consists of changes to lowerIterators. Thanks!

compiler/passes/errorHandling.cpp
```diff
@@ -128,13 +128,30 @@ try {
...
 
 
-// Static functions
-static bool canBlockThrow(BlockStmt* blk);
-static void checkErrorHandling(FnSymbol* fn);
+// This is the type for a map that is used when printing out the reason
+// that a function is marked throws. That is relevant for task
+// functions for a coforall, say, where the fact that the body of
+// the coforall throws isn't as relevant to the user as what in
+// particular in the body of the coforall throws.
```
vasslitvinov: The above sentence is quite a puzzle. Let's capitalize the last "what" to help the reader.

compiler/passes/errorHandling.cpp
```diff
@@ -128,13 +128,30 @@ try {
...
+//
+// Here the key is a FnSymbol and the value is he reason it was
```
vasslitvinov: he -> the

compiler/passes/parallel.cpp
```diff
@@ -25,7 +25,9 @@
 
 #include "astutil.h"
 #include "driver.h"
+#include "errorHandling.h"
 #include "expr.h"
+#include "errorHandling.h"
```
vasslitvinov: duplicate - please remove

compiler/resolution/lowerIterators.cpp
```diff
@@ -1290,8 +1302,16 @@ expandIteratorInline(ForLoop* forLoop) {
...
+    if (iterator->throwsError()) {
+      // For an inlined iterator, we don't want to track the error
+      // argument through the iterator class. Instead of setting
+      // a formal error argument, set an enclosing error variable.
+      collect_asts(ibody, asts);
+      replaceErrorArgumentWithEnclosingError(iterator, asts);
+    }
+
+    collect_asts(ibody, asts);
     replaceIteratorFormalsWithIteratorFields(iterator, ic, asts);
```
vasslitvinov: First, replaceErrorArgumentWithEnclosingError and replaceIteratorFormalsWithIteratorFields need only SymExprs. So I'd like 'asts' to be a vector of SymExprs, not all BaseASTs.

Second, replaceErrorArgumentWithEnclosingError does not change the set of SymExprs. So collect_asts or collectSymExprs should be done just once.

If either of the above is not the case, please ignore.

compiler/resolution/lowerIterators.cpp
```diff
@@ -1305,21 +1325,268 @@ expandBodyForIteratorInline(ForLoop*       forLoop,
                             BlockStmt*     ibody,
                             Symbol*        index) {
   TaskFnCopyMap taskFnCopies;
+  bool callWithErrorArg = false;
+  expandBodyForIteratorInline(forLoop, ibody, index, false,
+                              taskFnCopies, callWithErrorArg);
+}
+
+/*
+ A Goto ErrorHandling label in the body of an loop will not
```
vasslitvinov: an -> a

modules/internal/ChapelError.chpl
```diff
@@ -20,14 +20,176 @@
...
+      _next = nil;
+    }
+
+    proc writeThis(f) {
+      //f <~> "Error: " <~> msg;
+      f <~> "{msg = " <~> msg <~> "}";
+    }
+
+    pragma "no doc"
+    var _next: Error = nil; // managed by lock in record ErrorGroupRecord
```
vasslitvinov: This is an odd placement of a field. I'd prefer for all fields to be declared together.

vasslitvinov: ... esp. since both fields are referenced in a constructor above.

modules/internal/ChapelError.chpl
```diff
@@ -20,14 +20,176 @@
...
+  }
+
+  // Used by the runtime to accumulate errors. Needs
+  // only support adding errors concurrently. Errors
```
vasslitvinov: I'd prefer "Needs support only ..." if that's what you mean.

vasslitvinov:
Other than my minor comments, this is good to go.

compiler/passes/errorHandling.cpp
mppf: I fixed this and the below comment but the fixes will appear in the other PR.

compiler/resolution/lowerIterators.cpp
mppf: replaceErrorArgumentWithEnclosingError does change the set of SymExprs, since it creates a CallExpr and adds dummy error arguments. But perhaps it only adds SymExprs that replaceIteratorFormalsWithIteratorFields does not need to concern itself with. But in that event I think it's better to err on clarity/less potential bugs - in case that changes - than to optimize it.

Anyway I'll change them to use a vector of SymExprs. This change involves changes to other functions here, but it seems like an improvement... It'll be in a separate commit though so you can have a look.

modules/internal/ChapelError.chpl
mppf: I'll fix but it will appear in the other PR

PR #7038, opened 2017-08-16
Add future for broadcasting an empty record

benharsh:
Add future for compiler assertion when attempting to broadcast an empty record with an empty initializer


benharsh:
@noakesmichael : here's the bug we talked about.

ben-albrecht:
Thanks @benharsh !

PR #7040, opened 2017-08-17
Fix a dsiPartialThese bug exposed by changing growth factor

e-kayrakli:
partialThese methods were added as self-contained tests for partial reduction support. One of the helpers had a bug that was exposed by changing sparse growth factor in #6936 

This PR fixes the bug. `test/users/engin/partial_reduction_support` passed a no-local test.


e-kayrakli:
ping @ben-albrecht

ben-albrecht:
Thanks @e-kayrakli !!

PR #7043, opened 2017-08-17
Skip RVF and makeHeapAllocations if there are no on-statements

benharsh:
By skipping these sections of the compiler that are useless without
on-statements, we can slightly improve compilation speed.

Testing:
- [x] full local
- [x] full gasnet
- [x] numa examples/

### DIFF
--- a/compiler/optimizations/remoteValueForwarding.cpp
+++ b/compiler/optimizations/remoteValueForwarding.cpp
@@ -67,7 +67,7 @@ static void replaceRecordWrappedRefs();
 ************************************** | *************************************/
 
 void remoteValueForwarding() {
-  if (fNoRemoteValueForwarding == false) {
+  if (fNoRemoteValueForwarding == false && requireOutlinedOn()) {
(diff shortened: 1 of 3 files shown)


benharsh:
@mppf : here's a change we talked about a while back. It helps miniMD performance on my laptop by about 5%. Not a big win, but better than nothing.

mppf:
@benharsh - does it help miniMD execution performance or compilation performance?

benharsh:
Compilation performance.

benharsh:
@ronawho : I recall at one point there was some discussion on RVF and --local execution performance. Do you recall whether disabling RVF hurt performance?

ronawho:
Disabling rvf for task functions (https://github.com/chapel-lang/chapel/pull/4240) hurt performance for the task-spawning microbenchmark by ~3-4%. But we have a more principled way of getting that back. We're only doing rvf for on-statements now, so disabling it for --local shouldn't have any impact on execution performance.

Issue #7037, opened 2017-08-16
Dependency management for interoperability tests


nspark:
### Summary of Problem

In addressing #7008, I'm adding some Chapel-Python interoperability tests for the `ZMQ` module. These require Python to have the `pyzmq` module installed for the Python interpreter invoked via `start_test`. Currently, `pyzmq` is not installed in the `chpl-venv` used by `start_test`.

In addition, I'd ultimately like to test Chapel-* interoperability for `ZMQ`, where the other main idea for "*" might be Node.js. Of course, there may be others in time.

### Potential Solutions

1. Add `pyzmq` to the `test-requirements.txt` file for `chpl-venv`
1. Create a special virtual environment for these tests (Can this be invoked without conflicts with `chpl-venv` and `start_test`?)

I'm leaning toward (2) with a wrapper shell script to invoke `source bin/activate` and `deactivate` around the Python side of the test. That is:
* `server.chpl` spawns `wrapper.sh`
* `wrapper.sh` runs `source bin/activate ; python client.py ; deactivate`

I think this approach could also be portable to tests with Node.js and NPM.

ben-albrecht:
> Create a special virtual environment for these tests

I think this is the best we can do today. I see you have a commit that attempts this - did that work?

> (Can this be invoked without conflicts with chpl-venv and start_test?)

I don't anticipate any conflicts as long as `start_test` is still running in the `chpl-venv` and the test is running in the new virtualenv.

nspark:
> I think this is the best we can do today. I see you have a commit that attempts this - did that work?

As far as I've tested it, yes, it does. It uses a `PREEXEC` file to create the virtual environment if it does not already exist and the Python script is invoked via `wrapper.sh` (which activates/deactivates this new venv by the Chapel program via `spawn()`.

Issue #7008, opened 2017-08-14
ZMQ string/record serialization incompatible with REQ/REP pattern


nspark:
### Summary of Problem
As identified by [this Stack Overflow question and discussion](https://stackoverflow.com/questions/45651724/zeromq-w-chapel-and-python-cannot-answer-in-current-state), the ZMQ module does not correctly serialize Chapel strings (or other records) in a way that is compatible with the `REQ`/`REP` socket pattern or, likely, compatibility with other language bindings (e.g., Python).

### Steps to Reproduce

**Source Code:**

Server code (`server.chpl`):
```chapel
use ZMQ;
var context: Context;
var socket = context.socket(ZMQ.REP);
socket.bind("tcp://*:5555");

while ( 1 < 2) {
  var msg = socket.recv(string);
  writeln("got something");
  socket.send("back from chapel");
}
```

Client code (`client.py`):
```python
import zmq

context = zmq.Context()
socket = context.socket(zmq.REQ)
socket.connect("tcp://localhost:5555")

for request in range(10):
    print("Sending request %s ..." % request)
    socket.send(str("Yo"))
    message = socket.recv()
    print("OMG!! He said %s" % message)
```

**Compile command:**
`chpl -o server server.chpl`

**Execution command:**
```bash
# in parallel terminals
./server
python client.py
```

**Associated Future Test(s):**

Add tests for:
* `REQ`/`REP` pattern
* Chapel-Python interoperability

Please assign to @nspark

buddha314:
If it makes any kind of difference, the first use case I had in mind was passing JSON back and forth.

nspark:
@buddha314 I have a rough fix for sending and receiving strings such that Python and Chapel can talk back and forth, as in your example. It's "rough" in the sense that the fix -- that is, to use `zmq_msg_t` objects in a Python-compatible way -- only works for the string type (and not other Chapel records or primitive) and is only tested with the Python example in this issue.

As a minor issue, it also leaks memory internal to ZeroMQ because I still need to figure out how to pass a C function pointer into an extern C call within a Chapel program. I'll poke @bradcray about this one.

ben-albrecht:
> I still need to figure out how to pass a C function pointer into an extern C call within a Chapel program.

@nspark - just in case you aren't aware, here are some helpful resources on this:

- [`c_fn_ptr()`](http://chapel.cray.com/docs/master/technotes/extern.html#c-fn-ptr)
[...]

PR #7049, opened 2017-08-18
Chapel-Python interoperability for ZMQ's send/recv(string)

nspark:
This PR:
1. Reimplements `Socket.send/recv(string)` in the `ZMQ` module to be compatible/interoperable with Python's `PyZMQ` module
2. Adds a basic Chapel-Python interoperability test for ZeroMQ (see also #7037)
3. Updates module documentation on string serialization and interoperability

This PR should close the specific test that triggered #7008. Thanks for identifying the issue, @buddha314!


nspark:
@awallace-cray Can you clarify what exactly is failing in the Travis build? It seems to be coming from the `check_annotations.py` test, but I don't really understand how that relates to this PR.

awallace-cray:
Hi @nspark - I see it, but it doesn't mean anything to me either.  Investigating.

benharsh:
I don't think the annotation travis failure has anything to do with this PR. #7050 also fails travis with the same message.

awallace-cray:
Travis is generating "warning" messages from somewhere- not sure- and our annotation_check sees "warning" and reacts by throwing error.  
https://github.com/chapel-lang/chapel/commit/1816f3296bfc969bc10568dbddc0ff889dfcc5f5 
I think you could override this Travis error.

benharsh:
@nspark : Travis is back to normal now.

test/modules/packages/ZMQ/interop-py/wrapper.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pyscript=$1
+
+if [[ -f pyzmq-venv/bin/activate ]] ; then
+    source pyzmq-venv/bin/activate
+    python $pyscript
+    deactivate
```
ben-albrecht: What happens if this `deactivate` doesn't get called due to something going wrong? Will the parent environment be impacted?

util/start_test
```diff
@@ -90,7 +90,7 @@ def run_tests(tests):
 
     # print out flags
     logger.write('[tests: "{0}"]'.format(" ".join(files)))
-    if args.recurse:
+    if args.recurse and not os.path.isfile("NORECURSE"):
```
ben-albrecht: How is this feature different from the `NOTEST` file? Could `NOTEST` suffice for your case?

test/modules/packages/ZMQ/interop-py/wrapper.sh
nspark: I think the parent venv isn't affected because the `pyzmq-venv` is used only within a new Bash process when `wrapper.sh` is executed via `spawn()`.

util/start_test
nspark: I created a `NOTEST` file in `pyzmq-venv/` and this only keeps `start_test` from testing anything within that folder, but it still recurses into the directory tree of the Python venv. It's not a big deal because there aren't `*.chpl` files in that directory tree, but it produces output "cruft" that I'd _prefer_ to avoid if it's not too much trouble.

ben-albrecht: Ah, could you try creating a `pyzmq-venv.notest`? I think this should accomplish what you're trying to do.

For `start_test` rule-files (where `notest` is a *rule*), `filename.rule` will act recursively, whereas `RULE` will act on the current directory only. 

note-to-self: This is should be documented in [TestSystem.rst](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst)...

ben-albrecht:
@nspark - ping me when you'd like me to review again (and potentially merge).

nspark:
@ben-albrecht I think this one is ready

modules/packages/ZMQ.chpl
```diff
@@ -165,13 +165,52 @@ versions may expose more features native to ZeroMQ.
...
-(in bytes).
+In Chapel, sending or receiving messages is supported for a variety of types.
+Primitive numeric types and strings are supported as the foundation.
+In addition, user-defined :type:`record` types may be serialized automatically
+as `multipart messages <http://zguide.zeromq.org/page:all#Multipart-Messages>`_
+by internal use of the :chpl:mod:`Reflection` module.
+Currently, the ZMQ module can serialize records of primitive numeric types,
+strings, and other serializable records.
+
+.. note::
```
ben-albrecht: (nitpick) Do you think this note would be as effective without referencing "Chapel 1.16"? I don't see any references to Chapel versions elsewhere in module documentation, presumably because the documentation is versioned itself and user-facing changes are usually handled with deprecation messages.

nspark: I think this is a fair nitpick. Does Chapel have module-level deprecation warnings? This also isn't strictly a deprecation, but a protocol change. With this change Chapel 1.14/15 programs using `ZMQ` can't talk to Chapel 1.16 programs, though they can talk intra-version just fine. I'm not sure who really uses old Chapel binaries with new Chapel binaries _and_ the `ZMQ` module. That set of potential users is almost certainly the empty set.

I think this is a case where the module versioning really has nothing to do with the Chapel release, but there isn't another release mechanism until `mason` comes online. I was trying to be explicit and clear about the changes in `ZMQ` from 1.15 to 1.16, since there isn't a change log or release notes for this (or any?) specific module.

If you still want it gone, I can take it out and tweak the wording as needed.

ben-albrecht: > Chapel 1.14/15 programs using ZMQ can't talk to Chapel 1.16 programs, though they can talk intra-version just fine.

This point made it clearer to me. I'm fine with referencing the version numbers for a caveat like this. I wonder if this text should be included in the first paragraph of the `.. note:`, e.g.

```rst

   The serialization protocol for strings changed in Chapel 1.16 in order to
   support inter-language messaging through ZeroMQ. (See
   :ref:`notes on interoperability <interop>` below.) As a result, 
   programs using ZMQ compiled with Chapel 1.15 or earlier can't 
   communicate with programs compiled with Chapel 1.16 or later, 
   though they can communicate intra-version just fine.

```

nspark: That makes sense. I can make and push those changes tonight.

nspark: @ben-albrecht Good to go!

ben-albrecht:
Follow-up: I just realized we support a *version-changed* syntax for symbols in chpldoc:

```chpl
/* 
    .. versionchanged:: 1.16

        Stuff that changed in 1.16...
*/

This might be a more effective way to express a non API-breaking change like this one.

**Comment by nspark** [comment]
Nice! I'll work on a doc-edit soon-ish.

**Comment by nspark** [comment]
I guess it's not actually used yet:
```bash
$ grep 'versionchanged' $(find $CHPL_HOME/modules -name '*.chpl')
$ echo $?
1
```

It also stands out less than using `.. note::`, which makes me think the documentation might be better as-is.

ben-albrecht:
> It also stands out less than using .. note::, which makes me think the documentation might be better as-is.

Ah, OK - I did not realize this. I think the `.. note::` works fine then.

PR #7052, opened 2017-08-18
Comm-counts perf testing

awallace-cray:
First cut at Jenkins job script for Comm-counts perf testing, req by @mppf . 
Also see https://github.com/chapel-lang/chapel/pull/6604

Perf data goes into /data/sea/chapel/NightlyPerformance/chapcs/cc
(http://chapel.sourceforge.net/perf/chapcs/cc/)

### DIFF
--- a/util/cron/test-perf.chapcs.cc.bash
+++ b/util/cron/test-perf.chapcs.cc.bash
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# Run -perflabel cc- performance tests on a chapcs machine with gasnet
+
+CWD=$(cd $(dirname $0) ; pwd)
+
+export CHPL_TEST_PERF_CONFIG_NAME='chapcs'
+
+source $CWD/common-perf.bash
+source $CWD/common-gasnet.bash
+
+export CHPL_NIGHTLY_TEST_CONFIG_NAME="perf.chapcs.cc"
+
+SHORT_NAME=cc
+START_DATE=08/17/17
+
+perf_args="-performance-description $SHORT_NAME -sync-dir-suffix $SHORT_NAME -perflabel cc-"
+perf_args="${perf_args} -performance -numtrials 5 -startdate $START_DATE"
+
+# "make check" will fail on chapcs because Gasnet wants to use ibv.
+#  The warning message causes the good-output v. test-output comparison to fail.
+
+#  -no-buildcheck skips the "make check"
+nightly_args=-no-buildcheck
+
+$CWD/nightly -cron ${nightly_args} ${perf_args}
--- a/util/cron/nightly
+++ b/util/cron/nightly
@@ -36,6 +36,7 @@ $examples = 0;
 $performance = 0;
 $performancedescription = "";
 $performanceconfigs = "";
+$perflabel = "";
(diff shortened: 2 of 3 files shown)


util/cron/test-perf.chapcs.cc.bash
```diff
@@ -0,0 +1,24 @@
...
+
+export CHPL_NIGHTLY_TEST_CONFIG_NAME="perf.chapcs.cc"
+
+SHORT_NAME=cc
+START_DATE=08/17/17
+
+perf_args="-performance-description $SHORT_NAME -sync-dir-suffix $SHORT_NAME -perflabel cc-"
+perf_args="${perf_args} -performance -numtrials 5 -startdate $START_DATE"
+
+# make test will fail on chapcs because Gasnet wants to use ibv, and that warning
```
awallace-cray: That should read "make check" will fail..
Will correct that for final.

util/cron/test-perf.chapcs.cc.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# Run -perflabel cc- performance tests on a chapcs machine with gasnet
+
+CWD=$(cd $(dirname $0) ; pwd)
+
+export CHPL_TEST_PERF_CONFIG_NAME='chapcs'
```
mppf: I don't really know the impact of these variables but I can't tell if this should be chapcs-comm-counts, say

util/cron/test-perf.chapcs.cc.bash
```diff
@@ -0,0 +1,26 @@
...
+
+source $CWD/common-perf.bash
+source $CWD/common-gasnet.bash
+
+export CHPL_NIGHTLY_TEST_CONFIG_NAME="perf.chapcs.cc"
```
mppf: Can we call it perf.chapcs.comm-counts or something like that?

util/cron/test-perf.chapcs.cc.bash
```diff
@@ -0,0 +1,26 @@
...
+
+SHORT_NAME=cc
+START_DATE=08/17/17
+
+perf_args="-performance-description $SHORT_NAME -sync-dir-suffix $SHORT_NAME -perflabel cc-"
```
mppf: What does performance-description do? 'cc' might be too short if it's for something human-readable.

mppf:
This seems good to me, but I'm not sure about the various names yet.

awallace-cray:
I'm not sure about the various names, either!  I regard this as an initial guess, which should at least do no harm- that was why I added that new "cc" subdir. 

I know about some patterns and constraints on the job name.  For example, instead of job name "chapcs-perf-test-perf.chapcs.cc" (with script file test-perf.chapcs.cc.bash), it could easily have job name "performance-test-perf.chapcs.comm-count" (test-perf.chapcs.comm-count.bash).  

I know almost nothing about the perf data related names. Worse, this looks like a new type of perf test, not just another instance of some existing pattern.  

IMO, we could at least get these builds going and put some data out there, then consult with Elliot when he gets back.

mppf:
@awallace-cray - sure that is fine with me. thanks.

PR #7053, opened 2017-08-18
Fix annotation checker regex bug

benharsh:
Use regex that won't find other PR numbers in the merge message. For example, the old regex would find '123' in a merge message like "Merge pull request #500 that deals with #123"


benharsh:
@ronawho : Can you do a post-merge review?

awallace-cray:
+1

ronawho:
Ah, good catch. 👍

PR #7056, opened 2017-08-19
Simplify ErrorGroup.writeThis

mppf:
The previous ErrorGroup.writeThis used an associative array but its only
purpose was to print out the number of unique error messages encountered.
But it had the nasty side effect of significantly impacting compilation
speed.

Since it's not necessarily important to print out the number of unique
error messages, this PR changes that writeThis to no longer use an
associative array.  It's arguably better not to be allocating a lot of
memory in code run in a catch block, anyway, in case (in the future) we
are processing an error indicating that we are low on memory.

- [x] full local testing

### DIFF
--- a/modules/internal/ChapelError.chpl
+++ b/modules/internal/ChapelError.chpl
@@ -113,56 +113,43 @@ module ChapelError {
     }
 
     proc writeThis(f) {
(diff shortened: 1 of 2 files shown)


mppf:
@psahabu - could you review? thanks.

PR #7058, opened 2017-08-19
Drop implausible .bad file

noakesmichael:
A new future was developed that demonstrates that an initializer for a derived class/record
is allowed to access an uninitialized field for base class that is more than one step away.

Unfortunately the future included a .bad file that implicitly reflected the "uninitialized value"
and hence has undefined behavior for nightly testing.  Drop the .bad file.


lydia-duncan:
Thanks!  Should have thought about that more deeply before adding the .bad file

PR #7060, opened 2017-08-20
Future for distributed bulk transfer on rank change

bradcray:
This test demonstrates that something about the distributed bulk
transfer interface breaks down when rank change is involved.  It
may also have required me to use a subset of the dimensions that
are preserved, though I don't feel confident about that as I
type this.  At a quick glance, it appears to me that there's no
conversion going on in the domain between low- and high-D when
being passed through the doiBulkTransfer() interface.


bradcray:
@benharsh: This one looks right up your alley.  Nothing's blocked on it, but I stumbled across it while trying to get reindexed array view distributions properly privatized.  I'm not going to merge since I won't be around to mop up if anything goes wrong, but if you agree that it's a bug and ready to go in, please either feel free to, or I can do it when I get back (and file an issue too).

benharsh:
I'll merge this now, and I expect this bug will be resolved with the upcoming bulk transfer API rework.

Issue #6812, opened 2017-07-24
[Review] Data Structure API


LouisJenkinsCS:
# Data Structure API

Looking for feedback and review on the API for Data Structures, chpldocs hosted [here](https://louisjenkinscs.github.io/Distributed_Queue/index.html). Below are questions that help determine the overall direction of further development of the queue.

**Edit:** The chpldocs link is now updated. Old docs are [here](https://louisjenkinscs.github.io/Distributed_Queue/modules/queues/Queue.html). 

**Edit2:** I have removed the `QueueFactory` entirely for now.

**Edit3:** Generalizing for any kind of data structures.

~~## The QueueFactory~~

~~Emulates [Go's 'make'](https://golang.org/pkg/builtin/#make).~~

~~### Open Questions~~

~~* Should we maintain an interface at all?~~
  ~~* If not, then should we make each data structure a `record` instead?~~
   ~~* Resilience to Change vs Flexibility~~

### 'Freezing' Data Structures

While parallel-safe data structures are desired, not all things can be done in a
parallel-safe manner. For these situations, I believe the data structures should be armed with two
states: Mutable (unfrozen) and Immutable (frozen). The benefit is two-fold, as it
enables optimizations such as adopting a more optimistic manner of accessing data
by not contesting for locks meaning more performance, and the user has the benefit
of enforcing immutability across nodes without need for external synchronization. 
I believe that this is the only 'safe' way to allow things like reduction, zipper, 
and mapping iterations without modifying the queue while providing a significant performance boost.

#### Open Questions

* Are the semantic changes of having dual-mode operations too confusing?
* What happens to concurrent ongoing operations while freezing/unfreezing?
   * Should we halt? Block until unfrozen? Make it 'implementation-defined' behavior?

### Iteration

Iteration allows for the data structures to become more than just a simple containers, making things
such as reductions, zipper-iterables, and mappings possible.

### Open Questions

* Should we allow read-only iteration or 'draining' iteration, perhaps both?
   * Proposal: Based on if the data structure is 'frozen' or 'unfrozen'.

### Transactional Additions

Adding in bulk 'all-or-nothing' transactional approach.

#### Open Questions

* Should we allow adding non-scalar types? From other collections?

### Operator Overloading

While some operators can be helpful, others can be rendered irrelevant. Syntactic
sugar operators such as `+=` can help produce more readable code and are easy
to overload. Perhaps the `=` operator can be overloaded to perform a 'clear-and-add'
[...]

PR #7062, opened 2017-08-21
[GSoC] Parallel-Safe Collections Module and Distributed Data Structures

LouisJenkinsCS:
# Collections

I present a parallel-safe distributed data structures framework, similar to Java's 'Collection' framework. Original discussion can be seen [here](https://github.com/chapel-lang/chapel/issues/6812). Original repository containing benchmarks can be found [here](https://github.com/LouisJenkinsCS/Distributed-Data-Structures). Chpldocs documentation can be seen [here](https://louisjenkinscs.github.io/Distributed-Data-Structures/).

**Note:** This is the official PR for the GSoC program, but I do realize it may seem a bit like a rough-draft. Any feedback or constructive criticism would be appreciated it.

## Deque

Provides a strict ordering without sacrificing too much performance. Supports insertion
and removal from both ends, allowing FIFO, LIFO, and a Total ordering, which is
preserved across all nodes in a cluster, and employs a wait-free round-robin approach
to load distribution that ensures fairness in memory, bandwidth, and computation.

**Disclaimer:** The deque provided, while scalable, rely heavily on network atomics.
The benchmark results are produced using said network atomic operations.

## Bag

With performance that scales both in the number of nodes in a cluster and the
number of cores per node, we offer a multiset implementation, called a 'Bag',
which is a medium that allows storing and retrieving data in any arbitrary order.
This type of data structure is ideal for work queues as it employs it's own load
balancing, and offers unparalleled performance.

**Disclaimer:** A node can request a 'privatized' copy, which retrieves a clone
that is allocated on the node requesting it, reducing any excess communication.
Usage of `getPrivatizedInstance()` is highly advised for performance-critical
sections.

### Performance

We compare our data structures to a naive synchronized list implementation
as that is all that is available in the Chapel standard library.
In all cases, the data structures scale and outperform the naive implementation.

#### Insert

Implementation | Performance over Naive (at 64 nodes)
-------------- | :-----------:
SynchronizedList | 1x
DistributedDeque | 63x
DistributedBag | 403x

#### Remove

Implementation | Performance over Naive (at 64 nodes)
-------------- | :-----------:
SynchronizedList | 1x
DistributedDeque | 123x
DistributedBag | 651x

## Revisions

- [x] Remove 'freeze' logic
- [x] Engin's suggested changes
- [ ] Michael's suggested changes


mppf:
Two things for the PR message:
1. One place refers to MultiSet while the other refers to Bag
2. These data structures are designed for parallel use cases. I don't think that's true of the Java analogues, but I might be wrong about that. Either way, I think it's important to put that information up-front near the top of the PR message.

mppf:
@LouisJenkinsCS - yes I think that renaming is a good idea. Thanks!

e-kayrakli:
Thanks Louis! I discussed with @mppf offline and will give this an initial review a little later today.

Meanwhile, I don't know whether you have signed a CLA yet. If not, maybe you can take a look at this and get it out of the way while waiting for reviews:

https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements

mppf:
We already have a contributor agreement for Louis.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  }
+
+  inline proc takeElements(n) {
+    var iterations = n : int;
+    var arr : [{0..#n : int}] eltType;
+    var arrIdx = 0;
+
+    for 1 .. n : int{
+      if isEmpty {
+        halt("Attempted to take ", n, " elements when insufficient");
```
e-kayrakli: Similarly this and other halt messages should at least be preceded with something like "Internal DistributedBag Error:"

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  proc Bag(type eltType, parentHandle) {
+    this.parentHandle = parentHandle;
+  }
+
+  proc add(elt : eltType) : bool {
+    var startIdx = nextStartIdxEnq : int;
+    var phase = ADD_BEST_CASE;
+
+    while true {
+      select phase {
```
e-kayrakli: `select` seemed a bit of an overkill here, but maybe it's just me

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+                              // Sanity check: ensure segment did not fall under minimum since last check
+                              if  targetSegment.nElems.read() < WORK_STEALING_MINIMUM {
+                                targetSegment.releaseStatus();
+                                break;
+                              }
+
+                              extern proc sizeof(type x): size_t;
+                              // We steal at most 1MB worth of data. If the user has less than that, we steal a %, at least 1.
+                              const mb = WORK_STEALING_MAX_MEMORY_MB * 1024 * 1024;
+                              var toSteal = max(WORK_STEALING_MINIMUM, min(mb / sizeof(eltType), targetSegment.nElems.read() * WORK_STEALING_RATIO)) : int;
```
e-kayrakli: There is already `c_sizeof` no need to extern `sizeof` yourself.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+                  // at most a % of work from other nodes to give to ourselves.
+                  // As load balancer, we also are the only one who knows whether
+                  // or not all bags are empty.
+                  var isEmpty : atomic bool;
+                  isEmpty.write(true);
+                  segment.releaseStatus();
+                  coforall segmentIdx in 0..#here.maxTaskPar {
+                    var stolenWork : [{0..#numLocales}] (int, c_ptr(eltType));
+                    coforall loc in parentHandle.targetLocales {
+                      if loc != here then on loc {
```
e-kayrakli: I think performance-wise `on loc` should do the same as this (i.e. compiler will add the same check in the C code)

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+        otherwise do halt("Invalid phase #", phase);
+      }
+
+      // Reset variables...
+      idx = startIdx;
+      iterations = 0;
+      backoff = 0;
+    }
+
+    halt("DEADCODE");
```
e-kayrakli: Remove or change the message

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+  we use an unrolled linked list which further has its own (non-atomic) counter.
+*/
+
+/*
+  Frozen states... If we are FREEZE_UNFROZEN, we are mutable. If we are FREEZE_FROZEN,
+  we are immutable. If we are FREEZE_MARKED, we are in the middle of a state change.
+*/
+private const FREEZE_UNFROZEN = 0;
+private const FREEZE_MARKED = 1;
+private const FREEZE_FROZEN = 2;
```
e-kayrakli: similarly param these

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+/*
+    Turn off checks for freezing the data structure.
+*/
+config param DEQUE_NO_FREEZE = false;
+
+// For each node we manage an unroll block. This block needs to also support Deque
+// operations, and as such we maintain our own mini headIdx and tailIdx. Since we can
+// add and remove from either direction, we must also maintain the size ourselves...
+pragma "no doc"
+class LocalDequeNode {
```
e-kayrakli: I wonder if there is enough similarity between unroll blocks of the bag and the deque and can they be merged?

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+  var tail : LocalDequeNode(eltType);
+
+  // We cache the last deleted node to handle cases where we have rapid mixed push/pop
+  var cached : LocalDequeNode(eltType);
+
+  // The size of a segment. This is used as both a means of knowing when an element
+  // gets added, as well as a barrier to prevent the head and tail from being cached.
+  var size : atomic int;
+
+  inline proc recycleNode() {
```
e-kayrakli: I'd find `createNode` more appropriate but it is more personal taste

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+
+  pragma "no doc"
+  proc dsiGetPrivatizeData() {
+    return pid;
+  }
+
+  pragma "no doc"
+  inline proc getPrivatizedThis {
+    return chpl_getPrivatizedCopy(this.type, pid);
+  }
```
e-kayrakli: Similarly privatization methods should be marked with TODO or FIXME or something

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+  proc push(elt : eltType) : bool {
+    return pushBack(elt);
+  }
+
+  /*
+    Syntactic sugar for `popBack`.
+  */
+  proc pop() : (bool, eltType) {
+    return popBack();
+  }
```
e-kayrakli: I think push/pop should use front instead of back, but I don't know if that makes any practical difference

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,1000 @@
...
+          nodes[idx] = (node.size, node.tailIdx, node);
+        } else {
+          nodes[idx] = (-1, -1, node);
+        }
+      } else {
+        // Else update state...
+        nodes[idx] = (size, tailIdx, node);
+      }
+    }
+  }
```
e-kayrakli: Maybe names FIFO/LIFO can be clearer.. `iterFIFO`/`iterLIFO` or something.. or even `iter iterator(param order: orderKind` where `orderKind` is an enum or something

test/collections/DequeParity.compopts
```diff
@@ -0,0 +1,2 @@
+-sisBounded --main-module=DequeParity
+--main-module=DequeParity
```
e-kayrakli: Is there a need for --main-module flag?

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  be disabled for a raw performance boost. If the user attempts to freeze the
+  bag with this set to true, it merely returns false; all other checks are elided.
+*/
+config param BAG_NO_FREEZE = false;
+/*
+  The minimum number of elements a horizontal segment must have to become eligible
+  to be stolen from. This may be useful if some segments produce less elements than
+  others and should not be stolen from.
+*/
+config param WORK_STEALING_MINIMUM = 1;
```
mppf: yes, camelCase would be more typical. I agree about adding some sort of further detail to it - since it will arguably turn into a compilation option.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+    }
+  }
+
+  /*
+    Iterate over each bag in each node. If the bag is frozen, we elide the need
+    to acquire any locks and iterate directly over the bag. If the bag is not
+    frozen, we must acquire not only the needed locks, but also take a snapshot
+    approach, increasing memory consumption.
+  */
+  iter these() : eltType {
```
mppf: IIRC on statements are not allowed in serial iterators.

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,153 @@
...
+    Check if this data structure is empty.
+  */
+  inline proc isEmpty() : bool {
+    return size() == 0;
+  }
+
+  /*
+    Obtain the number of elements contained in this collection.
+  */
+  inline proc size() : int {
```
LouisJenkinsCS: Parenthesis methods cannot be overloaded, but having a synonym for 'length' seems spot on since 'length' can just call 'size()'.

mppf: Right but you could create a `proc Collection.size { return this.getSize(); }` e.g.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: I can agree to the prefixing and camel case. I'll hold off on this in case anyone else want to give their input.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  pragma "no doc"
+  proc dsiGetPrivatizeData() {
+    return pid;
+  }
+
+  pragma "no doc"
+  inline proc getPrivatizedThis {
+    if this.locale == here {
+      return this;
+    }
```
LouisJenkinsCS: @mppf 

Is [this](https://github.com/chapel-lang/chapel/blob/86164f1a7433a12855cd5a002522a4639194c1e4/modules/internal/ChapelArray.chpl#L213-L216) comment still relevant? If so, then we need to avoid excessive calls to get a privatized instance if its managed by Locale 0. If not, I can easily remove it.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+      }
+    }
+
+    localThis.concurrentTasks.sub(1);
+  }
+
+  /*
+    Triggers a more static approach to load balancing; note that this operation
+    is parallel-safe but is performed in parallel across all `targetLocales` and
+    should only be called at a point of synchronization.
```
LouisJenkinsCS: I agree on the 'from a sequential context' note. If you're using it in parallel with operations, it is safe to do so. The issue comes down to performance. To do this kind of operation, we establish a global locking order and proceed to grab *all* locks in said order, and is dead-lock free and even safe to use while another task is using dynamic load balancing (they'll just be blocking until static load balancing is finished). Even other calls to 'balance' is safe while a balance is in progress because it acquires it in that locking order.

The issue is *performance*. You should not call this excessively. It should only be called when you know, from a sequential context, that you have a severe imbalance.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  */
+  proc balance() {
```
LouisJenkinsCS: Doing a coforall over segments is necessary in that we steal horizontally, we steal from all segments with the same segment index as ours. Since everything is going to be evenly locally distributed, then  this distribution can be kept and balanced across all horizontal segments.

The latter is definitely due to ease of implementation.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+        if nElems > average {
+          var take = nElems - average;
+          var tmpBuffer = __primitive("+", buffer, bufferOffset);
+          segment.transferElements(tmpBuffer, take, buffer.locale.id);
+          bufferOffset += take;
+        }
+      }
+
+      // With the excess elements, redistribute it...
+      // BUG: Cannot seem to actually move data this way...
```
LouisJenkinsCS: The comment is outdated, and I'm working on a more valid replacement now.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+      // Then fill it.
+      var buffer = c_malloc(eltType, excess);
+      var bufferOffset = 0;
+      for loc in localThis.targetLocales do on loc {
+        var average = avg;
+        ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+        var nElems = segment.nElems.read() : int;
+        if nElems > average {
+          var take = nElems - average;
+          var tmpBuffer = __primitive("+", buffer, bufferOffset);
```
LouisJenkinsCS: I don't believe so. I'm not sure how it is implemented, but since the buffer is allocated on another node, dereferencing 'buffer[bufferOffset]' at all would cause it to dereference invalid memory. Would it work because `buffer` is taken by reference and would call a `C_Ptr` method to handle this in a delicate enough manner?

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Technically, a lot of things that use remote execution, such as remote atomic operations without network atomics would violate an agreement of 'not using `on` statements'. I'm all for it, especially because you cannot remotely use a C_Ptr and must an `on` statement to make use of it.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: The reason why we skip `if loc != here` is because we do not want to steal from ourselves.

modules/standard/DistributedBag.chpl
e-kayrakli: Maybe your comment here can be incorporated in the documentation, as well. Mainly the fact that this is an expensive operation and for better performance should be called from a sequential context etc.

modules/standard/DistributedBag.chpl
e-kayrakli: Apparently you can do:

https://github.com/chapel-lang/chapel/blob/master/modules/internal/CPtr.chpl#L224

modules/standard/DistributedBag.chpl
e-kayrakli: Ughh, I got this wrong. Now makes more sense

modules/standard/DistributedDeque.chpl
LouisJenkinsCS: I'm not sure about merging them. I use constant sized tuples, but truth be told I did not have time to benchmark whether or not increasing the size of the tuples benefits performance (it likely does), and converting to C pointers so that I can double the size of each successor node likely would also yield another significant boost for when we have a large workload. 

Time constraints being as they are, I left them as tuples.

modules/standard/DistributedDeque.chpl
LouisJenkinsCS: Since `Enqueue` pushes to the back, I'd expect that if I did `enqueue(); pop();` to get the same thing I just added, hence I have both use the same end.

test/collections/DequeParity.compopts
LouisJenkinsCS: Good question, that was from before, but I do need it to run both with and withou t`-sisBounded` set, and I wasn't sure if blank lines get counted.

modules/standard/DistributedDeque.chpl
LouisJenkinsCS: Yeah, I was thinking of having `these(param order : orderKind = orderKind.NONE)`, but I wasn't sure. That way I'd have `orderKind.NONE, orderKind.FIFO, orderKind.LIFO`, etc. As well, since enums no longer causes a compiler internal error with deadCodeElimination (fixed in #6570), I have no issue with enums anymore.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: @mppf 

It simply states that you cannot `yield` in and `on` statement, nothing against using them.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Whoaaaa, I had no idea this existed! Definitely will use that. Its not in the documentation for C_Ptr module though, also strangely it has `pragma "no doc"`

test/collections/DequeParity.compopts
e-kayrakli: You can do:

```
-sisBounded=true
-sisBounded=false
```

modules/standard/DistributedBag.chpl
mppf: That comment is still relevant, but it's not about *getting a privatized instance* but rather *creating a privatized ID*. The ID creation step is the one that has a possibly unbalanced reliance on locale 0. Going from a privatized ID to a class instance pointer is the same cost on all locales.

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
...
+    in a variable not declared outside of loop...
+
+    ::
+
+      for i in 1 .. N {
+        var retval = c.remove();
+      }
+
+  */
+  inline proc remove() : (bool, eltType) {
```
mppf: I'm used to seeing remove calls refer to a specific element - where the pattern here is more what I'd probably call `pop`. Is there a particular reason you chose add/remove vs. push/pop? Any precedent?

Of course it's the case that we don't want to imply that all of these Collections meet some particular definition of Stack ordering.

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
...
+class Collection {
+  /*
+    The type of element that this Collection holds.
+  */
+  type eltType;
+
+  /*
+    Adds an element to this data structure.
+  */
+  inline proc add(elt : eltType) : bool {
```
mppf: I wouldn't expect virtual methods that just halt should be marked inline.

modules/standard/Collection.chpl
mppf: No-inline here too?

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
...
+
+  */
+  inline proc remove() : (bool, eltType) {
+    halt("'proc remove() : (bool, eltType)' is not supported...");
+  }
+
+  /*
+    Determine whether an element exists in this collection.
+  */
+  inline proc contains(elt : eltType) : bool {
```
mppf: no-inline

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
...
+    Syntactic sugar for `getSize`.
+  */
+  inline proc size : int {
+    return getSize();
+  }
+
+  /*
+    Obtain the number of elements contained in this collection.
+  */
+  inline proc getSize() : int {
```
mppf: no inline here

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
...
+
+/*
+  Syntactic sugar for 'add'.
+
+  **BUG:** Compiler produces a warning that `c` should be a `ref`, when it is not needed
+  because it is a class and can be used by value. Need a pragma to disable this?
+  As such, this cannot be used where `ref` intents cannot be used, such as ``forall``
+  and ``coforall`` loops.
+*/
+proc +=(ref c : Collection(?eltType), elt : eltType) {
```
mppf: This one seems more reasonable to inline, if you want.

modules/standard/Collection.chpl
LouisJenkinsCS: The reasoning behind declaring it `inline` is due so that the user will know the precisely location they have invoked the unsupported method. Normally, if it is not inlined, it shows a line number in Collections.chpl, which isn't entirely helpful.

modules/standard/Collection.chpl
```diff
@@ -0,0 +1,116 @@
+/*
+  A 'Collection' is a data structure, a container for elements that provide support
+  for insert, lookup, remove, and iteration operations.
```
mppf: Based on the number of open issues / compiler bugs here, I'd like to recommend that this module go into modules/packages. It is used in the same way as it would be if it went in to standard/. My expectation would be that once some of these user-facing issues that limit its usefulness are addressed, it can move to standard/.

mppf: (and if it wasn't clear, we'd want to put all of your new modules in either standard/ or packages/ together)

modules/standard/Collection.chpl
LouisJenkinsCS: Right, I didn't want to imply any particular ordering, and `add` and `remove` seemed descriptive enough, as `remove` in this case means "remove *any* item".

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+  A highly parallel segmented multiset. Each node gets its own bag, and in each bag
+  it is segmented into 'here.maxTaskPar' segments. Segments allow for actual parallelism
+  while operating in that it enables us to manage 'best-case', 'average-case',
+  and 'worst-case' scenarios by making multiple passes over each segment. In the
+  case where there is no oversubscription, the best-case will always be achieved
+  (considering any other conditions are also met), while in the case of oversubscription
+  or, for example a near empty bag, we fall into the 'average-case', etc. Examples
+  of 'best-case' scenarios for a removal would be when a segment is unlocked
+  and contains some elements we can drain, and the 'average-case' would be to find
+  any segment, unlocked or not, that contains elements we can drain, and so on.
```
mppf: Have you tried to use chpldoc to render the documentation for your new modules? (While that's a good idea, it's not necessarily required). In this case in particular, there's usually a comment at the top of the module that offers a high level introduction. If that's what this comment is, it's too detailed. I'm not objecting to this comment's existence - I'm just saying that we'd need an even more basic comment (What is this and why would I use it?) before any comment that goes into as much detail about performance as this one.

modules/standard/Collection.chpl
LouisJenkinsCS: I have no objections to that. I placed it in standard/ because the `List` module was also in standard/.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: The documentation (using chpldoc) can be seen [here](https://louisjenkinscs.github.io/Distributed-Data-Structures/). I normally left the high-level description for the main structure itself.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+  dramatically causes a performance drop. Furthermore, a severe imbalance across
+  nodes, such as an unfair distribution of work to a single or small subset of nodes,
+  may also causes an equally severe performance degradation. To circumvent this,
+  we allow the user to determine manually when to 'balance', which will evenly distribute all
+  elements across nodes. This data structure scales in terms of nodes, processors per node,
+  and even work load. The larger the work load, the more data that gets stolen when
+  stealing work, and better locality of elements among segments. As well, to
+  achieve true parallelism, usage of a privatized instance is a requirement, as
+  it avoids the overhead of remotely accessing class fields, bounding performance
+  on communication.
```
mppf: With the array implementation (say), the priviatized instances are handled transparently by the module code. I think that doing so in your case would amount to:
 * creating a record wrapping these types
 * storing the privatized ID a well as the class pointer in the record
 * creating forwarding methods from the record to the class instance / privatized class

I can see that you've been making use of the class system in Chapel, so that Bag and Deque are subclasses of Collection. I can think of several reasons why you might have wanted to do that, but my personal expectation would be that Collection would be an "interface" (which does not exist yet in Chapel) and that Bag and Deque are implementations of that interface. I bring this up because if you do the record-wrapping as I describe above, one would no longer be able to store a mixed-type array of Collections that are either Bag or Deque. But I don't think doing so is actually useful, in this particular case.

So, if you were to follow the array strategy:
 * Collection, DistributedBag, DistributedDeque would become CollectionImpl, DistributedBagImpl, DistributedDequeImpl
 * The new records wrapping these would be called DistributedBag and DistributedDeque (or something, you could consider using lower case to name them for example)
 * Users would no longer have to manually `delete` or manage privatized instances

Lastly, I'm noting that this module creates a class with the same name as the module. We need to find a way not to do that - Chapel is a single namespace language - so one namespace needs to be able to identify both the module and the class - otherwise we get some weird compilation problems.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+private param REMOVE_AVERAGE_CASE = 3;
+private param REMOVE_WORST_CASE = 4;
+
+/*
+  The initial amount of elements in an unroll block. Each successive unroll block
+  is double the size of it's predecessor, allowing for better locality for when
+  there are larger numbers of elements. The better the locality, the better raw
+  performance and easier it is to redistribute work.
+*/
+config param distributedBagInitialBlockSize = 1024;
```
mppf: I'm wondering - is there a particular reason these need to be *compile time* constants? Certainly I can think of reasons we might prefer to hide them from the -help output, but I'm having trouble imagining that such configuration settings couldn't be set at launch time? (I.e. config const).

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+*/
+config param distributedBagWorkStealingMinElems = 1;
+
+/*
+  A parallel-safe distributed multiset implementation that scales in terms of
+  nodes, processors per node (PPN), and workload; The more PPN, the more segments
+  we allocate to increase raw parallelism, and the larger the workload the better
+  locality (see `distributedBagInitialBlockSize`). This data structure is unordered and employs
+  its own work-stealing algorithm, and provides a means to obtain a privatized instance of
+  the data structure for maximized performance.
```
mppf: This paragraph is a pretty good introduction. For full-on user documentation you'd also need to include an example of using it.

modules/standard/DistributedBag.chpl
mppf: Meaning I wouldn't expect that removing the conditional above would affect performance.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  /*
+    Obtains a privatized version of this instance. The privatized version is a
+    cloned instance that is allocated on this node, which is useful for eliding
+    communications generated from accessing instance fields. Using another node's
+    instance will significantly penalize performance by bounding overall throughput
+    on the communication between the two nodes.
+  */
+  proc getPrivatizedInstance() {
+    return getPrivatizedThis;
+  }
```
mppf: Right, that's just because it's a class not a record. The privatized instance ID is currently stored within the class itself - meaning that if you had an object of that type, you'd have to do a GET to retrieve the privatized ID in order to get your local copy, which sortof defeats the purpose of privatization. The pattern of using a record containing {pointer, private id} I described above is a reasonable way to solve this problem. I'm not so sure that it makes sense to expect users to manually call getPrivatizedInstance.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+
+  /*
+    Obtain the number of elements held in all bags across all nodes. This method
+    is best-effort and can be non-deterministic for concurrent updates across nodes,
+    and may miss elements or even count duplicates resulting from any concurrent
+    insertion or removal operations.
+  */
+  proc getSize() : int {
+    var sz : atomic int;
+    forall loc in targetLocales do on loc {
```
mppf: This should be a coforall

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+
+  /*
+    Performs a lookup to determine if the requested element exists in this bag.
+    This method is best-effort and can be non-deterministic for concurrent
+    updates across nodes, and may miss elements resulting from any concurrent
+    insertion or removal operations.
+  */
+  proc contains(elt : eltType) : bool {
+    var foundElt : atomic bool;
+    forall elem in getPrivatizedThis {
```
mppf: I havn't dug in to the these call below, but I would've expected this one to be written with somehting like
``` chapel
coforall loc in targetLocales do on loc {
  forall elem in getPrivatizedThis.localElements() {
        if elem == elt {
          foundElt.write(true);
       }
  }
}
```

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+    Triggers a more static approach to load balancing, fairly redistributing all
+    elements fairly for bags across nodes. The result will result in all segments
+    having roughly the same amount of elements.
+
+    **Note:** This method is very heavy-weight in that it should not be called too
+    often. Dynamic work stealing handles cases where there is a relatively fair
+    distribution across majority of nodes, but this should be called when you have
+    a severe imbalance, or when you have a smaller number of elements to balance.
+    Furthermore, while this operation is parallel-safe, it should be called in a
+    sequential context.
```
mppf: I think - but am not sure - that what you mean by "it should be called in a sequential context" is that it should only be called from one task at a time - since it redistributes the entire data structure and uses on statements etc to handle data elsewhere. I don't think "should be called in a sequential context" means that for me. I'd probably just say "Furthermore, while this operation is parallel-safe, it's only really useful to call it from one task at a time, since that call will update the entire distributed data structure".

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+      var excess : int;
+      for nElems in nSegmentElems {
+        if nElems > avg {
+          excess += nElems - avg;
+        }
+      }
+
+      // Allocate buffer, which holds the 'excess' elements for redistribution.
+      // Then fill it.
+      var buffer = c_malloc(eltType, excess);
```
mppf: Why use `c_malloc` instead of just a normal Chapel array?

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+      bufferOffset = 0;
+      for loc in localThis.targetLocales do on loc {
+        var average = avg;
+        ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+        var nElems = segment.nElems.read() : int;
+        if average > nElems {
+          var give = average - nElems;
+          var arr : [1..give] eltType;
+          on bufferOffset {
+            var tmpBuffer = buffer;
```
mppf: If buffer were also a Chapel array, you'd write this transfer in a way that the Chapel compiler can optimize into a small number of PUTs by using array slices. E.g.

  arr[1..give] = buffer[bufferOffset..bufferOffset+give-1];

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+      // beforehand, causing more undefined behavior. For now, I use Chapel Arrays.
+      bufferOffset = 0;
+      for loc in localThis.targetLocales do on loc {
+        var average = avg;
+        ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+        var nElems = segment.nElems.read() : int;
+        if average > nElems {
+          var give = average - nElems;
+          var arr : [1..give] eltType;
+          on bufferOffset {
```
mppf: I'm pretty sure you meant on Locales[bufferOffset] or something like that - as it is, on bufferOffset will probably be a no-op since bufferOffset is running on the current locale.

LouisJenkinsCS: We call `on bufferOffset` because bufferOffset is located on the node that holds the `buffer`, which is the node that called `balance`. On line 341, you see that we do a `for loc in Locales do on loc`. We have to jump to the locale owning the `buffer`  because its a C_Ptr.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Currently I have a helper method that takes both a c_ptr, a size, and a locale and writes into it, and one that does the opposite (reads from a c_ptr). I chose c_ptr over chapel arrays as they are more lightweight, but I didn't know about the bulk transfer support you mentioned before.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: While I can agree that record wrapping is a very good idea, it wouldn't allow me to have any kind of inheritance (issue #6851). Not only would you no longer be able to store mixed-type arrays, you wouldn't be able to store *any* kind of `Collection` as you can no longer inherit, and even if you could you wouldn't be able to use the `Collection` in an opaque way (since before removal, they would truncate all methods and fields to the base type).

LouisJenkinsCS: Furthermore, there is an issue with using records in a `forall` or `coforall` loop. Example below

```chpl
record bag {
	var someData : atomic int;
	proc add(elt) { someData.add(elt); } 
}

var b : bag;
forall i in 1 .. 100 {
	b.add(i);
}

writeln(b);
```

[TIO (Online Compiler)](https://tio.run/##PY0xDsIwEATr3Cu2TBoraXHLR872AZacODosKKK83dhIUO5qZ9Y/eJdUq4rPGuD4joOGFyueeZUrF8YFXPIaPeJWLA27Zg8OYZRUJhz/nfl1FifoJOoS1@jmtHTLyikhNgkWGINlnvuT@2Jxsp14ayySttG1SLV@AA)

modules/standard/DistributedBag.chpl
mppf: Thanks for the link - I had forgotten that you'd already generated that.

Right, but it's good for such documentation to start with an introductory paragraph. For this reason, we typically explicitly declare the module (e.g. ``` module DistributedBag { } ```) and attach comments to that.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: That does sound like a good idea, having each `Bag` implement some parallel and serial iterator to reduce the amount of redundant code.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: I remember originally it had to do with using tuples (which is no longer the case), I suppose that slipped my mind. I'll make that change.

modules/standard/DistributedBag.chpl
mppf: You can use a Collection in an opaque way with generic programming. I'm not sure that supporting mixed-type arrays of Collection is worth the trade-off of requiring users to call privatization functions manually.

I'm having trouble interpreting this:
> you wouldn't be able to store *any* kind of `Collection` as you can no longer inherit
> you wouldn't be able to use the `Collection` in an opaque way (since before removal, they would truncate all methods and fields to the base type).

What does inheriting have to do with storing a Collection? What do you mean by removal?

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+        var give = excess - bufferOffset;
+        var tmpBuffer = buffer + bufferOffset;
+        segment.addElementsPtr(tmpBuffer, give, buffer.locale.id);
+      }
+
+      c_free(buffer);
+    }
+
+    // Phase 3: Release all locks from first node and segment to last node and segment.
+    for loc in localThis.targetLocales do on loc {
```
mppf: Doesn't this need to happen in the reverse of the acquire order to prevent some deadlock case? Or am I just being paranoid?

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+    from concurrent operations.
+
+    **Warning:** Breaking from an iterator will leak the snapshot due to issue #6912.
+    It will however leave all Bags in a safe state.
+  */
+  iter these() : eltType {
+    for loc in targetLocales {
+      for segmentIdx in 0..#here.maxTaskPar {
+        // The size of the snapshot is only known once we have the lock.
+        // BUG: c_ptr does not seem to work here and causes a segmentation fault at runtime.
```
mppf: well yeah... c_ptr is always local and doesn't support use from another locale. _ddata would do that, but it's almost certainly better to just use a Chapel array in the first place.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: No, we hold the locks on all segments, so we don't need to bother checking to see if they are locked or not, we just unlock them. We start by locking the first node segment, then continue on to the last, so we unlock in the same way. Also should be noted that we use atomic spinlocks over `sync` variables.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+    return size == 0;
+  }
+
+  inline proc isFull {
+    return size == cap;
+  }
+
+  inline proc push(elt : eltType) {
+    if elems == nil {
+      halt("DistributedBag Internal Error: 'elems' is nil");
```
mppf: We have `assert` too, which might be more succinct for errors like this.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+      yield buffer[i];
+    }
+  }
+}
+
+/*
+  A segment block is an unrolled linked list node that holds a contiguous buffer
+  of memory. Each segment block size *should* be a power of two, as we increase the
+  size of each subsequent unroll block by twice the size. This is so that stealing
+  work is faster in that majority of elements are confined to one area.
```
mppf: It's probably also worth pointing out here that any locking / parallel safety is taken care of outside of BagSegmentBlock.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: I mean, you wouldn't be able to do this...

```chpl
proc addSomeData(c : Collection) {
   c.add(1);
  c.add(2);
  c.add(3);
}

var bag = new DistributedBag(int);
addSomeData(bag);
```

You would not be able to use it that way at all. You can't say "Give me any collection so I can add elements to it", which I believed was one of the main key-points to using it. Unless, you're saying that we have it like this...

```chpl
proc addSomeData(c : Collection(?implType));
```

Where `Collection` is generic on the implementation? Can you provide an example of how you envision it?

LouisJenkinsCS: I take back what I said above on being generic on the implementation, as implementations can have different methods from the base class.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: That explains why its a huge deal for the classes to not share the same name as the module itself. I'll make that change.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Also by 'removal' I mean removal of record inheritance from the language.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+
+      // With the excess elements, redistribute it...
+      // BUG: Cannot seem to actually move data this way...
+      bufferOffset = 0;
+      for loc in localThis.targetLocales do on loc {
+        var average = avg;
+        ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+        var nElems = segment.nElems.read() : int;
+        if average > nElems {
+          var give = average - nElems;
```
LouisJenkinsCS: Forgot about this (and on `take`) I'll see about renaming

modules/standard/DistributedBag.chpl
LouisJenkinsCS: The issue I've had with assert is that for some reason I can't seem to find the source of an assertion, normally it shows a line number that points into the Assert module. Furthermore, it doesn't seem to allow me to provide an error message for failing an assertion.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: In this situation, I'll give it to you, that does look like it'd save me a ton of logic and triple checking, I'll see about using that, and in particular, using slices like that.

modules/standard/DistributedBag.chpl
e-kayrakli: My two cents: In a separate discussion I have mentioned to Louis that I don't find being able to use data structures polymorphically super-useful and such functionality can be sacrificed. I think that discussion was about not being able to override parallel iterators.

Also, I was never a big fan of the whole getPrivatizedInstance business. But I regarded it as something we may have to do for performance until there is a better story for privatization.

I haven't considered record-wrapping for more seamless privatization. If all that would happen is that we lose the ability to use the structures polymorphically yet get rid of a user-facing privatization interface for performance, I totally support making that change.

mppf: @LouisJenkinsCS - in fact the record-wrapping strategy I described still allows for inheritance on the Impl classes. I'm not sure why else you would want record inheritance.

LouisJenkinsCS: @mppf 

Here are some potential issues I would need to know how to handle with record wrapping...

1) How can the user use these parallel-safe data structures in a parallel context, such as `forall` and `coforall`? Engin informed me of `pragma "default intent is ref if modified"` which does seem to work, but just to be sure.

2) How do we manage copies of the original record wrapped collection if the user were to make a copy by-value? For example, if you were to allocate the structure in a small scope but have it be used outside of it, and the original goes out of scope, then how do we ensure we do not clean this up? Engin suggested [`SharedObject`](http://chapel.cray.com/docs/latest/modules/packages/SharedObject.html?module-SharedObject), and I could overload assignment operator to automatically update the reference to it. Just to make sure you'd want this change.

3) The record-wrapped types are not generic on the implementation, only on the element itself. The type cannot be used polymorphically/dynamically at all. `CollectionImpl` becomes more of a 'pseudo-interface', a simple 'promise' to implement at least those methods (although at this point, do we really need the base `Collection` anymore? It'd be less restrictive to just remove it all-together and not use any inheritance.) This is a rather drastic change, I'm not sure how long this would take me.


Finally, would this need to be done by the initial PR and not in a separate one? If we're going to be making this change, then `Collection` kind of goes away entirely, or if it does exist it wouldn't really be entirely useful.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: - [x] ...

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+  }
+
+  /*
+    Obtains a privatized version of this instance. The privatized version is a
+    cloned instance that is allocated on this node, which is useful for eliding
+    communications generated from accessing instance fields. Using another node's
+    instance will significantly penalize performance by bounding overall throughput
+    on the communication between the two nodes.
+  */
+  proc getPrivatizedInstance() {
```
LouisJenkinsCS: This method may not be needed soon.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1245 @@
...
+      // Allocate buffer, which holds the 'excess' elements for redistribution.
+      // Then fill it.
+      var buffer = c_malloc(eltType, excess);
+      var bufferOffset = 0;
+      for loc in localThis.targetLocales do on loc {
+        var average = avg;
+        ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+        var nElems = segment.nElems.read() : int;
+        if nElems > average {
+          var take = nElems - average;
```
LouisJenkinsCS: - [x] ...

modules/standard/DistributedBag.chpl
LouisJenkinsCS: - [x] ...

modules/standard/Collection.chpl
LouisJenkinsCS: - [x] ...

modules/standard/Collection.chpl
LouisJenkinsCS: - [x] ...

modules/standard/Collection.chpl
LouisJenkinsCS: - [x] ...

modules/standard/Collection.chpl
LouisJenkinsCS: - [x] ...

modules/standard/Collection.chpl
LouisJenkinsCS: - [x] ...

modules/standard/DistributedBag.chpl
LouisJenkinsCS: - [x] ...

modules/standard/DistributedBag.chpl
LouisJenkinsCS: - [x] ...

modules/standard/DistributedBag.chpl
LouisJenkinsCS: After thinking it over as well, I can see the novelty. However, at this point, I'm 200% for discarding the inheritance entirely for their own particular data structures. It'd make my job 10x easier than having to have them all inherit from a base class. I may even be able to do it before Friday. Also @mppf I already told @e-kayrakli but for me, school starts Monday, and I need Saturday and Sunday for getting school supplies and moving in, so in terms of significant changes, I was hoping I would have everything finished by Friday night latest. I'm all for coming back and making the necessary changes when I get the time. I personally think as well that the focus should shift from privatization-related issues to reviewing the rest of the code.

test/collections/CollectionNQueens.chpl
```diff
@@ -0,0 +1,192 @@
+use Barrier;
```
ben-albrecht: You'll want to update this to `Barriers`, assuming #7035 is getting merged before this.

modules/standard/DistributedBag.chpl
mppf: OK. As you work towards removing the inheritance, just make Collection exist for documentation purposes (i.e. it shows a human what the interface is that is supported by the others). In the future, the compiler will support actual interface declarations for this kind of thing.

mppf: Also, it makes more sense to me to use OwnedObject either directly or as inspiration. Generally speaking, we've been moving away from reference counting things in preference to requiring that the user not delete them before they go out of scope. I think if that strategy is good enough for language-supported arrays, it's good enough for Bag etc.

mppf: Lastly, even if you (say) replaced `class DistributedBag` with `record DistributedBag`, that would probably cause problems with the privatization code. In order to privatize you probably do need `record DistributedBag` and `class DistributedBagImpl`.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+        var buffer : [dom] eltType;
+
+        on loc {
+          ref segment = getPrivatizedThis.bag.segments[segmentIdx];
+
+          if segment.acquireIfNonEmpty(STATUS_LOOKUP) {
+            var block = segment.headBlock;
+            while block != nil {
+              for i in 0 .. #block.size {
+                  buffer.push_back(block.elems[i]);
```
mppf: I'm not certain, but this loop might perform better if you allocate space for some maximum / large segment in buffer and if most of the time you never reallocate it. Additionally in that event you can use the bulk array transfer idiom I described above (with array slices). It's not obvious to me if segments have a maximum size. If they do not, this loop could pick a maximum size to work with at a time, and loop over adding chunks of that size efficiently.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+              halt("DistributedBag Internal Error: Snapshot attempt with bufferSz(", bufferSz, ") with offset bufferOffset(", bufferOffset + block.size, ")");
+            }
+            __primitive("chpl_comm_array_put", block.elems[0], here.id, buffer[bufferOffset], block.size);
+            bufferOffset += block.size;
+            block = block.next;
+          }
+
+          // Yield this chunk to be process...
+          segment.releaseStatus();
+          yield (bufferSz, buffer);
```
mppf: This is fine but as a documentation matter you'll have to say that this data structure does not `zip` with rectangular data structures (e.g. forall i,j in zip( myDistributedBag, 1..) )

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+          var block = segment.headBlock;
+          var bufferSz = segment.nElems.read() : int;
+          var buffer = c_malloc(eltType, bufferSz);
+          var bufferOffset = 0;
+
+          while block != nil {
+            if bufferOffset + block.size > bufferSz {
+              halt("DistributedBag Internal Error: Snapshot attempt with bufferSz(", bufferSz, ") with offset bufferOffset(", bufferOffset + block.size, ")");
+            }
+            __primitive("chpl_comm_array_put", block.elems[0], here.id, buffer[bufferOffset], block.size);
```
mppf: It's almost certainly a better idea to use the array slice idiom to do this transfer, instead of c_malloc.

modules/standard/DistributedBag.chpl
mppf: Fair enough, it's not worth too much trouble at this point. However we could certainly add error messages for asserts. And for the line numbers, try CHPL_UNWIND=libunwind and see where that gets you. Or compile without CHPL_DEVELOPER=1.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+    return elt;
+  }
+
+  proc BagSegmentBlock(type eltType, capacity) {
+    if capacity == 0 {
+      halt("DistributedBag Internal Error: Capacity is 0...");
+    }
+
+    cap = capacity;
+    elems = c_malloc(eltType, capacity);
```
mppf: Not so sure why c_malloc makes sense in this case. I'd expect BagSegmentBlock to just be using Chapel arrays.
```
  // The capacity of this block.
  var cap : int;
  var dom = 0..#cap;
  // Contiguous memory containing all elements
  var elems :  [dom] eltType;
  var next : BagSegmentBlock(eltType);
  // The number of occupied elements in this block.
  var size : int;
```

I havn't gotten to the part where you initialize a BagSegmentBlock with a ptr. If you need to do that to, c_ptr makes a little bit more sense, but there are ways of doing that too with arrays.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+  }
+
+  inline proc acquireWithStatus(newStatus) {
+    return status.compareExchangeStrong(STATUS_UNLOCKED, newStatus);
+  }
+
+  // Set status with a test-and-test-and-set loop...
+  inline proc acquire(newStatus) {
+    while true {
+      if currentStatus == STATUS_UNLOCKED && acquireWithStatus(newStatus) {
```
mppf: I don't think it's required, but this loop is the reason compareExchangeWeak exists...

mppf: I mean this kind of loop.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+    while destOffset < n {
+      if headBlock == nil || isEmpty {
+        halt(here, ": DistributedBag Internal Error: Attempted transfer ", n, " elements to ", locId, " but failed... destOffset=", destOffset);
+      }
+
+      var len = headBlock.size;
+      var need = n - destOffset;
+      // If the amount in this block is greater than what is left to transfer, we
+      // cannot begin transferring at the beginning, so we set our offset from the end.
+      if len > need {
```
mppf: Pretty sure this would be simpler using the array slice transfer, but this is OK if that doesn't work out for some reason.

modules/standard/DistributedBag.chpl
mppf: I think it's fine, if you want to write it that way. The compiler literally translates it into some `if` statements.

modules/standard/DistributedBag.chpl
mppf: Engin does have a point though about optimization. the `coforall loc in locales on loc do` pattern is optimized by the compiler to not create a new local task (only a remote one). In this case you could achieve the same effect if you created a serial iterator over targetLocales that yielded all but the locale == here, and then you'd be able to write the compiler optimized pattern once more.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+  proc BagSegmentBlock(type eltType, capacity) {
+    if capacity == 0 {
+      halt("DistributedBag Internal Error: Capacity is 0...");
+    }
+
+    cap = capacity;
+    elems = c_malloc(eltType, capacity);
+  }
+
+  proc BagSegmentBlock(type eltType, ptr, capacity) {
```
mppf: I don't see where this constructor is used but I'm probably missing something.

modules/standard/DistributedBag.chpl
mppf: By the way, it's possible one of these array transformations will add unwanted GETs. So it's important to sanity check after you do one of them to make sure performance is not adversely impacted.
One thing to know is that local arrays aren't privatized. That means that there are some GETs for data in the array header when doing the bulk transfer. The thing is, that might not have significant impact on performance, depending.
So, if you try using the array slice idiom, and performance is still good, I'd call it a win. If you try it, and performance is 2x worse, just keep to the c_ptrs and explicit transfers.

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+  benefitted from local distribution from the round-robin algorithm discussed above.
+  We also steal a ratio of elements, say 25%, because it leaves all other segments
+  with 75% of their work; the more elements they have, the more we take, but the
+  less they have, the less we steal; this also has the added benefit of reducing
+  unnecessary work stealing between segments when the bag is nearly emptied. Furthermore,
+  stealing horizontally ensures that all segments remain filled and that we still
+  achieve parallelism across segments for removal operations. Lastly, we attempt
+  to steal a maximum of `N / sizeof(eltType)`, where N is some size in megabytes
+  (representing how much data can be sent in one network request), which keeps
+  down excessive communication.
```
mppf: I think it'd be nice in this comment somewhere to talk about the overall algorithm. You have some of that above but it doesn't say you're using spin locks or why there are different lock statuses (acquire for writing vs acquire for reading).

modules/standard/DistributedBag.chpl
```diff
@@ -0,0 +1,1076 @@
...
+
+      // Reset variables...
+      idx = startIdx;
+      iterations = 0;
+      backoff = 0;
+    }
+
+    halt("DistributedBag Internal Error: DEADCODE");
+  }
+}
```
mppf: I've finished looking over this file.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Interesting point. I wonder, is it possible to do something like...

```chpl
coforall loc in (targetLocales - here) do on loc {
   /* Code here */
}
```

Basically, create a temporary array with the element removed? I suppose I could define a inner iterator like such (if this is legal chapel code)

```chpl
iter otherTargetLocales() {
   for loc in targetLocales {
      if loc != here {
         yield loc;
      }
   }
}

coforall loc in otherTargetLocales do on loc {
   /* Code here */
}
```

modules/standard/DistributedBag.chpl
LouisJenkinsCS: Just to be sure, this is an affirmative on committing to record wrapping. I'll begin working on these changes in a new branch immediately (stop me if I assumed incorrectly). 

Also, looking at the `OwnedObject` module, doesn't it seem more than a bit... restrictive? Looking at the implementation, all it does is disallow the user from using the object in certain contexts using pragmas. `SharedObject`, reference counting or not, allows the user to make use of the all data structures as if they are classes, which I strongly believe is the way to go about this. *If* privatization support improves for classes, then this will not be an issue in the future and since this becomes a 'module/package' it doesn't have to strictly follow the 'standard'.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: A few things here. Originally, I used Chapel Arrays and found that after making the switch to using tuples, I found a significant performance improvement (I believe Chapel perfs show the comparisong [here](http://chapel.sourceforge.net/perf/chapcs/?startdate=2015/09/10&enddate=2017/08/23&graphs=arrayvstupleserialaccesses)) after the switch (at least 4 - 5x improvement). After deciding to double the size of each successor unroll block, I saw an improvement of at least 1.5x by going from Tuples to C_Ptr. Those improvements weren't even factoring in bulk transfer, that was just from adding elements one at a time.

We also use bulk transfer across nodes for load balancing (dynamic load balancing). Currently, the explicit transfers definitely could be improved a bit for more performance, like implementing some direct zero-copy from one segment to another, but overall C_Ptrs have dramatically improved performance. Even if array bulk transfers do not cause a significant performance degradation, I can see it causing issues in other areas. Especially while adding a large number of elements, and even during load balancing where we likely would create a significant number of unroll blocks.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: I can convert `acquireWithStatus` to use `compareExchangeWeak`, sure, or is this about using the 'Test-And-Test-And-Set' over the 'Test-And-Set' pattern?

modules/standard/DistributedBag.chpl
LouisJenkinsCS: You're correct, I used to make use of this function during dynamic work stealing, but no longer. I'll remove it.

modules/standard/DistributedBag.chpl
LouisJenkinsCS: @mppf 

I'm thinking that with record wrapping, I might be able to do this without requiring explicit privatization (perhaps this might be best?). What if I just create an array of `DistributedBagImpl` in the record and have each Locale pick which to use based on that? The only issue is that this type should almost never be used by `ref` or else it incurs the cost we're trying to avoid (hence the need for `SharedObject`). Speaking of `SharedObject`, I think I can make it work by just making a `SharedObject` hold on to some helper class that has the sole purpose of destroying the `DistributedBag` fields when it gets destroyed. 

Furthermore, with all this indirection, I can probably proceed this way, and it probably is best in general. I still have concerns based on how to ensure that as a record it never is used by wide reference  (is there a pragma for that?).

> As you work towards removing the inheritance, just make Collection exist for documentation purposes

Would it be possible to make it into a CHIP instead? However, if we were to keep the inheritance with the implementations to make `Collection` relevant, the issue would next come down to forwarding. If I want to forward only on the privatized instance, is this currently allowed? Etc.

modules/standard/DistributedDeque.chpl
```diff
@@ -0,0 +1,864 @@
...
+  the necessary checks to ensure two invariants: A `pop` operation never proceeds
+  if the deque is empty, and that a `push` operation never proceeds if the deque
+  is full. In the case where we have a `pop` and a non-empty deque, or a `push`
+  with a non-full deque, we proceed within a single atomic fetchSub or fetchAdd
+  respectively, hence is wait-free. In the case that we have a `pop` with a
+  empty deque, or a `push` with a full deque, the queueSize becomes either
+  negative or over capacity. When this happens, the faulty operation will, via
+  remote execution (on-statement), perform a CAS-retry loop. If the opposite operation
+  (`push` is the opposite of `pop` and vice-verse) sees the size in a incorrect state,
+  it will continue its fetchSub/fetchAdd cycle, also bringing the size closer to
```
mppf: vice-verse -> vice-versa

modules/standard/DistributedBag.chpl
mppf: OK, works for me then. You might want to add a comment around the uses of c_ptr / c_malloc to say something like "This code uses low-level C-style arrays in order to control as much of the communication overhead as possible".

modules/standard/DistributedBag.chpl
mppf: Right defining the iterator as you did above is exactly what I'm suggesting

modules/standard/DistributedBag.chpl
mppf: I'd lean towards mimicking the pattern of record wrapping done by arrays. That would not necessarily mean using OwnedObject. In particular, the main point there is that when the record is destroyed, the Impl instance is deleted, and there's no provision for keeping the instance alive longer if there are record copies of it. In this way it is similar to using a bare class.

It's probably not currently a good idea to try to use OwnedObject for this purpose, as OwnedObject has some issues (to do with lvalue checking and compiler errors at least) that you could easily avoid just by writing your own constructor / destructor.

If you created an array of DistributedBagImpl, you'd still need privatization / replication in that array to avoid communication when getting the local instance ID. I think it's reasonable to use the privatization mechanism, but having a Replicated array field (say) is a pretty reasonable alternative.

About not wanting to use the record as a wide ref - Ben H. recently committed a mechanism that can be used to serialize/deserialize instances. Obviously it might be more than you're able to do in your last week, but you could try to use that to control what happens when your data structure is used on the other side of an `on` statement.

I'm not sure what you mean by forwarding... "If I want to forward only on the privatized instance, is this currently allowed?" I think that's what the array record does, depending on exactly what you mean. Might be best to find me on irc to ask about these.

mppf:
Are there any papers etc. that you can refer to for some description of how/why these work? If so, can you put citations in the comments somewhere?

test/collections/DequeParity.good
```diff
@@ -0,0 +1 @@
+SUCCESS
```
mppf: these tests should be in test/modules/packages/collections e.g. or at a minimum somewhere in test/modules/packages.

mppf:
Can you write comments that point to the related publications and that summarize what is different about these from the published versions?

mppf:
I ran some testing in the standard (local) configuration and get a failure for CollectionNQueens: 
> CollectionNQueens.chpl:168: error: halt reached - Spun: 1000000

mppf:
Passed full local testing except for CollectionNQueens.

mppf:
Passes new test directory with:
 * quickstart
 * quickstart + gasnet
 * qthreads
 * qthreads + gasnet

modules/packages/DistributedBag.chpl
```diff
@@ -0,0 +1,1084 @@
...
+              return true;
+            }
+          }
+
+          phase = ADD_AVERAGE_CASE;
+        }
+        /*
+          Pass 2: Average Case
+
+          Find any segment (locked or unlocked) and make an attempt to acquire it.
```
mppf: The comment says "Find any segment" but the average case code only looks in 1 segment. Is that intentional?

LouisJenkinsCS: Yep, that's outdated documentation. Originally we checked the status of the segment before trying to add to it (back when I had work stealing keep the lock on the segment, it would just look for an adjacent segment which wasn't). I can make a OR correcting it if needed

LouisJenkinsCS: Yep, that's outdated documentation. Originally we checked the status of the segment before trying to add to it (back when I had work stealing keep the lock on the segment, it would just look for an adjacent segment which wasn't). I can make a OR correcting it if needed

LouisJenkinsCS: Btw, has there been a reported issue so far with the collections?

mppf: I havn't seen one, I'd let you know. I think we're still waiting for somebody to battle-test them.

You can fix the comment if you like, no big deal.

Issue #6997, opened 2017-08-11
ambiguous param method overloads cause a segfault in resolution


bradcray:
### Summary of Problem

I accidentally created a duplicate overloaded param method on a type, and this resulted in a segfault in the compiler rather than an "ambiguous call" style error message. 

### Steps to Reproduce

**Associated Future Test(s):**
[`test/functions/resolution/ambig-param-overloads.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/resolution/ambig-param-overloads.chpl) #6996 

**Source Code:**

```chapel
record R {
  param flag;
  var x: int = 42;
}

proc R.getX where flag return x;
proc R.getX param where !flag return 0;
proc R.getX param where !flag return 0;

var myR1 = new R(false);
var myR2 = new R(true);

writeln(myR1.getX);
writeln(myR2.getX);
```

**Compile command:**
`chpl ambig-param-overloads.chpl`

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.16.0 pre-release (2624f85)`

daviditen:
I just merged PR #7064 which fixes this issue.

bradcray:
Thanks David!

PR #7064, opened 2017-08-21
Fix a bug for ambiguous param methods

daviditen:
The disambiguateByMatch function was incorrectly returning 1 match when there
was no best match in this case. When there is no best match, return 0 instead.

Remove a .future and .bad for a test revealing this bug.  Fixes issue #6997.


daviditen:
Thanks @mppf!

PR #7065, opened 2017-08-21
Fix codegen for comparisons to be able to handle refs

e-kayrakli:
Without this patch codegen seems to omit dereferences for `a` in the following snippet:
```chapel
proc foo() {
  var a = 10;
  on Locales[0] {
    writeln(a>10);
    writeln(a>=10);
    writeln(a<10);
    writeln(a<=10);
  }
}

foo();
```

Such code compiled fine with --no-denormalize, since the dereferencing was handled by PRIM_MOVE. However, since denormalize removes that PRIM_MOVE this code was compiled without derefrenceing causing pointer vs integer comparison.

It seems that ref intent for `a` is inserted post denormalize, so there doesn't seem to be a direct way of handling this at denormalize time. Moreover, such issues are handled at codegen time for arithmetic operations. I followed the same strategy for comparisons.

This PR also adds a test that generates faulty code on master.

(I can't easily reproduce this issue without on statements. I tried to do it with ref variables but it seems to work fine for those)

Tests:
- [x] release/examples/hello*chpl passed with GASNet
- [x] users/engin/refComparison.chpl passed with GASNet
- [ ] Full GASNet
- [ ] Full LLVM (?)


e-kayrakli:
This was encountered by @LouisJenkinsCS.

Also, I can't run paratest myself.

benharsh:
I'll start some paratesting:
- full gasnet
- gasnet llvm examples/

test/users/engin/refComparison.skipif
```diff
@@ -0,0 +1 @@
+CHPL_COMM!=none
```
benharsh: Does this test really need a skipif?

e-kayrakli: Hmm.. not that it fails with `CHPL_COMM!=none`. Just that the on statement is optimized away (I think) with that environment, thus it doesn't test anything new. In other words, with `CHPL_COMM=none` this test passes without this PR, too.

I can remove it if you want.

benharsh: Should it really be "CHPL_COMM==none" , which would only run the test in multilocale configurations?

e-kayrakli: Oh boy :( You are right, sorry about the confusion. Do you want me to fix it or remove it altogether?

benharsh: Let's just remove it

PR #7068, opened 2017-08-22
Fix error with error handling exits of forall loops

mppf:
Previous work with throwing across foralls adjusted GOTO_ERROR_HANDLING in lowerIterators in order to account for iterator inlining. Unfortunately, some patterns of GOTO out of a forall loop show incorrect AST before lowerIterators. In particular, callDestructors adds DeferStmts for _freeIterators before the error handling block and in any GOTO_ERROR_HANDLING to that error handling block. The result was that a GOTO_ERROR_HANDLING within the body of a forall loop would deallocate the parallel iterator inside one of the task functions. That resulted in a use-after-free when other code (including other tasks) tried to use the parallel iterator.

This PR addresses this issue by changing the error handling exit of a forall loop to come in two parts:
 1. Break out of the loop.
 2. Handle error exits after the forall loop.

That way, the break out of the loop does not deallocate the parallel iterator while in the loop.

To do this:
 * Added GOTO_BREAK_ERROR_HANDLING as a type of goto, so that errorHandling can add this new break from a forall loop and follow-on passes can treat it appropriately
 * Added ForLoop::isLoweredForallLoop, so that errorHandling can identify loops that are implementing a forall loop.
 * Create new helper methods in ErrorHandlingVisitor: setOuterErrorAndGotoHandler and gotoHandler.
 * Add throwingForall as a field in TryInfo
 * Add enterForLoop and exitForLoop implementation to add the new break-on-error construct

Additionally, in the process of preparing this PR I realized that a change previously added - to make iterators throw implicitly if their body does - was incorrect. I found another way to solve the same issue - by adding chpl_propagate_error and handling it in lowerIterators. The issue was that a leader/standalone iterator might include a coforall. The coforall counts as "throwing" if its body throws. So, the error handling for that coforall needs to work in the context that the iterator is inlined into.

- [x] passed test/errhandling with -valgrindexe
- [x] passed test/errhandling with --baseline
- [x] full local testing
- [x] passed hellos and test/errhandling with gasnet, numa
- [x] passed hellos and test/errhandling with numa

Reviewed by @psahabu and @vasslitvinov - thanks!


mppf:
@psahabu - I'd like you to have a look at this in addition to @vasslitvinov, thanks.

compiler/passes/errorHandling.cpp
```diff
@@ -797,6 +903,21 @@ bool isCheckErrorStmt(Expr* e)
   return false;
 }
 
+Symbol* getErrorSymbolFromCheckErrorStmt(Expr* e)
```
psahabu: I can't seem to find where this function is called, or if it used to be in another part of the compiler. Could you point it out to me?

mppf: It's not called right now but I'd need it if I try again to make progress on non-inlined iterators that throw. I could remove it from this PR if you like.

psahabu: You can leave it in, but it might be good to add a comment.

vasslitvinov:
Makes me wonder whether we should have an "error throwing" intent for foralls.

PR #7070, opened 2017-08-22
[GSoC] Add clang builtins header

coodie:
This PR adds builtins wrapper for some of clang headers along with script that generates them based on "Builtins.def" file from clang sources. This solves problem with performance issue as noted in mandelbrot-complex test.

TODOs:
- [x] Add custom implementation of 'cabs' function.

I noticed that 'proper' cabs implementation function is a little complex and might be quite difficult to maintain. We could simply copy-paste it from some open library (julia lang has one), but I think it's best if someone else made the decision. I think wrapper for very simple yet fast version (used only when passed --no-ieee-float flag) of cabs should be enough for now.

@mppf and @dmk42 are proposed reviewers


mppf:
Travis says:
```
runtime/include/llvm/clang_builtins_wrapper.h

Add the copyright with: $CHPL_HOME/util/buildRelease/add_license_to_sources.py <files>
```

I think you need to run this script from the new script.

mppf:
Passed full local testing

mppf:
@coodie - I'm seeing some testing failures with --llvm
[Error matching compiler output for parallel/taskCompare/elliot/empty-qthreads-taskspawn (compopts: 1)]
[Error matching compiler output for parallel/taskCompare/elliot/empty-qthreads-taskspawn (compopts: 2)]
[Error matching compiler output for parallel/taskCompare/elliot/empty-qthreads-taskspawn (compopts: 3)]
[Error matching compiler output for parallel/taskCompare/elliot/empty-qthreads-taskspawn (compopts: 4)]
[Error matching compiler output for runtime/gbt/tasks/callbacks/tcb]
[Error matching compiler output for runtime/gbt/tasks/callbacks/tcb-unin]

In particular it looks like we should be adding the header files only for the LLVM compilation but it looks like we're also adding them when compiling C code. I'm not immediately sure how to achieve that, but we probably need to move the include.

mppf:
I'm thinking that the best way to handle this would be to name the header wrapper functions differently. Then, we adjust our LLVM code generator to know about the current set of builtins that get special treatment - and then the code generator uses the wrapper.

E.g. name the function, instead of cabs, chpl_llvm_wrapper_cabs; and then adjust getFunctionLLVM to find chpl_llvm_wrapper_cabs when it is looking for cabs. I would expect that would entail generating another header file - this one for the compiler - to specify this mapping.

coodie:
In all tests the error is related to having the counterparts of our functions have different declaration than the ones header 'complex.h', specifically:

```
static declaration of 'conj' follows non-static declaration
```

I'm not sure, but I'd expect simpler solution to simply remove static (i removed it and it works nicely). I'd expect that some test would include 'complex.h' and our header still be able to override that with our custom implementations.

mppf:
Right, but the problem is that it relies on the particulars of how math.h works; so it would be better to use different function names so as not to be dependent on that.

compiler/codegen/codegen.cpp
```diff
@@ -1962,6 +1973,17 @@ void codegen(void) {
     }
   }
 
+  // Wrap calls to chosen functions from c library
+  if( llvmCodegen ) {
+#ifdef HAVE_LLVM
+    forv_Vec(FnSymbol, fn, gFnSymbols) {
+      if (fn->hasFlag(FLAG_EXTERN)) {
+        fn->cname = getClangBuiltinWrappedName(fn->cname);
```
mppf: I think it'd be nicer to keep the conditional in here, since most of the time fn->cname does not change. Having the conditional here will make that more obvious to people reading the code.

compiler/codegen/codegen.cpp
```diff
@@ -1855,6 +1856,16 @@ adjustArgSymbolTypesForIntent(void)
 extern bool printCppLineno;
 debug_data *debug_info=NULL;
 
+
+static const char* getClangBuiltinWrappedName(const char* name)
+{
+  auto it = chplClangBuiltinWrappedFunctions.find(name);
```
mppf: Since this is not in a HAVE_LLVM ifdef, you can't use C++11 features. Of course, you could put it in such an ifdef...

runtime/include/llvm/clang_builtins_wrapper.h
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* Generated by clang_builtin_gen.py based on clang's Builtins.def */
+#define ADD_WRAPPER_PREFIX(S) chpl_clang_builtin_wrapper_##S
+static inline double ADD_WRAPPER_PREFIX(cimag)(_Complex double arg0)
```
mppf: Since we're generating this file anyway, it seems an odd choice to use the macro to add the prefix. I think it'd be better to leave the generated file more straightforward and change that in the .py script if necessary.

mppf:
This is looking good to me, and I think it's ready to go once my feedback is addressed.  I'll need @dmk42 to help with the testing since I'll be away next week. Thanks!

runtime/include/llvm/clang_builtins_wrapper.h
coodie: Macro is used here for consistency with with `chapel_libc_wrapper.h` file and how it's supposed to be used in future when we want to wrap more builtins.

dmk42: OK, I understand this point about consistency.  I will go ahead and allow this.

dmk42:
Passed `release/examples` with and without `--fast` on a Mac.
Passed full local testing with and without `--fast` on linux64.

PR #7074, opened 2017-08-22
Quiet PGI compiler error by emitting a NULL element

benharsh:
After #7050 we would see failures when using PGI as the target compiler. PGI would complain that the initializer for the chpl_global_serialize_table was empty. This PR simply adds a NULL element if there are no globals to add to the table.

Testing:
- [x] full local
- [x] gasnet
  - [x] release/examples
  - [x] optimizations/remoteValueForwarding/serialization


mppf:
👍 . Sometimes I prefer the pattern of always adding a NULL element at the end of such global tables.

PR #7075, opened 2017-08-22
Update last of the madchap helper constructors to be an initializer

lydia-duncan:
Prior to the update I made to the _new function, this change was encountering
a segfault in copyPropagation.  It turns out that the _new function was not
properly copying arguments, especially in the case of argument dependencies - in
this case it was trying to reference the 'k' argument in the init function
instead of the corresponding 'k' argument it added to itself.  Now that I have
resolved that error, I can commit this change.

This test would benefit from the ability to update const fields during Phase 2.

Passed a sanity std/ paratest

### DIFF
--- a/test/studies/madness/aniruddha/madchap/MRA.chpl
+++ b/test/studies/madness/aniruddha/madchap/MRA.chpl
@@ -59,9 +59,9 @@ class Function {
     const r0   : [dcDom] real;
     const rp   : [dcDom] real;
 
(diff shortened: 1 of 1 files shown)


lydia-duncan:
@vasslitvinov - would you like to review this test, or are you satisfied with the look you took at it when we talked about it a week or so ago?

vasslitvinov:
Looks perfect. Thanks for doing the work!

The diffs are pretty pathetic. Git should do better than that for such a trivial change. :)

PR #7076, opened 2017-08-22
Removed OwnedObject; Reduced memory leaks

Spartee:
This PR fixes the debugger for the toml parser as well as a few minor improvements in functionality and code cleanliness. The use of the OwnedObject module was the reason for the breakage of the debugger. OwnedObject also prevented the toml parser from deleting unnecessary tokens. This has also been fixed.


ben-albrecht:
Curious - what was the overall memory leak reduction (for an arbitrary test)?

PR #7083, opened 2017-08-23
DistributedIters performance tests

s-geronimoanderson:
Performance tests for DistributedIters module. Currently used for comparing distributed guided iterator with block-distributed default iterator. Includes NOTEST for now, but also perfexecopts and perfkeys for future reference. (I'm willing to remove these *perf* files if they'll not be useful.)


test/modules/packages/distributedIters/perf/benchmarkDistributedIters.perfexecopts
```diff
@@ -0,0 +1,2 @@
+--test=uniform --mode=guided --n=10000 # benchmarkDistributeIters-guided.perfkeys
```
benharsh: I think we can do without the distinct 'perfkeys' files here and instead just do something like "# distributedGuided". See the following perfexecopts for examples:

[`linear algebra`](https://github.com/chapel-lang/chapel/blob/master/test/modules/packages/linearalgebra/performance/transpose-perf.perfexecopts)
[`sort`](https://github.com/chapel-lang/chapel/blob/master/test/modules/packages/Sort/performance/performance.perfexecopts)

test/modules/packages/distributedIters/perf/benchmarkDistributedIters.chpl
```diff
@@ -0,0 +1,473 @@
+/*
+  Benchmarks for testing the DistributedIters module performance.
+
+  ..
+    Part of a 2017 Cray summer intern project by Sean I. Geronimo Anderson
+    (sgeronimo@cray.com) as mentored by Ben Harshbarger (bharshbarg@cray.com).
```
benharsh: github usernames again 😄

s-geronimoanderson: Right! Ah!

test/modules/packages/distributedIters/perf/benchmarkDistributedIters.perfexecopts
s-geronimoanderson: Okay! Does this mean "use the default perfkeys file and store the result in distributedGuided.dat"? I could update the testing docs.

benharsh: Yup, that's what it means.

Let's leave the testing docs alone for this PR.

Issue #6160, opened 2017-05-03
How should Chapel handle integer overflow?


mppf:
For example, if you have
``` chapel
var x:int(8) = 127;
x += 1;
writeln(x);
```

what should happen? Let's introduce some terminology for the basic options:

 * *undefined behavior*: this program is wrong, who knows what it will do
  * *wrap*: x is -128
  * *trap*: process dies with a signal (as with, say, dereferencing NULL)
  * *throw*: an overflow is thrown and can be caught 

Note that the main difference in functionality is that *trap* is expected to be unrecoverable but *throw* is explicitly something the user's code can handle.

### What do other languages do?

* C: signed integer overflow is *undefined behavior*; unsigned integer overflow will *wrap*
* [Swift:](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AdvancedOperators.html) signed and unsigned overflow *trap*
* [Rust](http://huonw.github.io/blog/2016/04/myths-and-legends-about-integer-overflow-in-rust/):
   * in debug mode, "panic" (AFAIK panic maps to *trap* in our terminology since it's not meant to be recovered from; note though a Rust panic does involve stack unwind+cleanup).
   * in release mode, *wrap*
   * further controllable by compiler flags and explicit add functions

### Areas of concern

* performance of generated code
* safety

### Loops and Bounded Ranges

When optimizing loops, it's important to know that the loop induction variables cannot wrap. That is important for optimizations such as loop unrolling or vectorization.

For loops that iterate over bounded ranges, optimizations can generally assume that the loop induction variable does not wrap. The main wrinkle here is that the form of the loop might be written in a way that requires a *sentinel value*. This comes up if the loop follows typical C style, where the loop condition is checked after an increment.

The Chapel spec currently requires that *sentinel values* are representable:

> An attempt to iterate over a range causes an error if adding stride to the range's last index overflows its index type, i.e. if the sum is greater than the index type's maximum value, or smaller than its minimum value.

While it would be possible to write the loop implementing bounded range iteration without assuming that the *sentinel value* is representable, doing so might add complexity to almost all loops and lead to lower performance in general.

The following loop is an example of bounded range iteration that currently causes a runtime error (without --fast) because the *sentinel value* is not representable:

``` chapel
var a:int(8) = 125;
var b:int(8) = 127;
for i in a..b {
  writeln(i);
}
```

This will currently turn into C like this:
``` c
int8_t a = 125;
[...]

PR #7092, opened 2017-08-23
[GSoC] Generate nsw for signed integers operations

coodie:
Couple of months ago I have added 'nsw' (meaning undefined behaviour when overflowing on signed operations instead of wrapping) keyword to add instruction in LLVM IR since it was related to vectorization not occuring.  According to #6160 discussion there was a little inconsistency until this PR, because multiplications and subtractions would wrap while adds wouldn't, this tiny PR fixes that.

Proposed reviewers: @mppf and @dmk42


dmk42:
This looks good.  I think the same thing should also be done with the CreateShl and CreateNeg calls.

test/llvm/generate-nsw-for-signed/SKIPIF
```diff
@@ -0,0 +1 @@
+CHPL_
```
mppf: Pretty sure this SKIPIF isn't finished... probably meant to be CHPL_LLVM!=llvm

mppf:
Passed full local testing, with the SKIPIF fix

dmk42:
I retested this after the additional nsws were added.  I got the following results.

* Passes full local testing without `--fast` on linux64
* Passes release/examples with and without `--fast` on a Mac
* Passes almost all of full local testing with `--fast` on linux64

The only test that fails with `--fast` is spec/marybeth/casts.  I believe that test is improper because it tries to use undefined behavior to detect undefined behavior, which cannot be relied upon.  I have asked the Chapel team if we can remove the undefined behavior part of that test.  Once that problem is resolved, this PR will be ready to merge.

PR #7098, opened 2017-08-24
Adjust sync statements to better handle throws

mppf:
Consider an example like the following:

``` chapel
  try {
    sync {
      begin someFunction();
      someThrowingFunction();
    }
  } catch {
    ...
  }
```

Before this PR, that translates into something more or less like this:

``` chapel
  {
    begin someFunction();
    someThrowingFunction(err);
    if err != nil then
      goto handler;
    waitEndCount; // wait for dynamically enclosed begins
  }
  handler:
  if err != nil then
    ...
```

The problem with this is that if the body of the sync block threw
an error after tasks were created, the sync would not wait for these
tasks. In order to address that issue, this PR wraps the body of each
sync statement in a try/catch block. Additionally, since the sync
statement might 'throw' errors from nested begins, any error 'thrown'
in the body of the sync statement itself needs to be included among
the task errors (otherwise we'd have possibly 2 errors and only one
place to put them - we have to propagate all errors). So, now build.cpp
translates the above example into something like this:

``` chapel
  try {
    try {
      begin someFunction();
      someThrowingFunction(err);
    } catch e:Error {
      chpl_save_task_error(e);
      e = nil;
    }
    waitEndCount; // wait for dynamically enclosed begins
  } catch {
    ...
  }
```

This change caused problems for an initializer check to allow sync statements in phase 1. Based on discussion with @lydia-duncan, I've .future-ized that test.

Besides this change in build.cpp, this PR does the following:
 * adjusts chpl_delete_error to check the error is not nil first, which is apparently necessary
 * adjusts error handling lowering to enable the pattern of setting the error to nil so that it is not deleted. This is a workaround to enable this pattern (one day, I hope these errors are wrapped in Owned).

- [x] full local testing


test/classes/initializers/phase1/syncStatements.future
```diff
@@ -0,0 +1,9 @@
+feature request: sync statements in phase 1
+
+Philisophically, it should not be a big deal for phase 1 initializers to
```
lydia-duncan: "Philisophically" -> "Philosophically"

lydia-duncan: This future looks good otherwise, thanks!

mppf:
@psahabu - that's right, a sync still "throws". The existing test begin-throws.chpl exercises this, but I don't think we have any tests showing an error in strict mode (yet).

PR #7101, opened 2017-08-24
[GSoC] Add tests for check againt const

coodie:
These tests make sure that emitting checking against constants emits proper check, this is mentioned in 8. in the document: http://releases.llvm.org/4.0.0/docs/Frontend/PerformanceTips.html#other-things-to-consider.

Proposed reviewers: @mppf, @dmk42


mppf:
Looks good to me, but it will need a testing check.

dmk42:
Looks good.
Test passes on Mac and linux64.

PR #7108, opened 2017-08-25
move IO calls in tests to error handling

psahabu:
Some tests were handling `syserr` explicitly from `IO.chpl` when it would be better to handle `SystemError` through the error handling system. This hardens the error handling implementation while showing that transitioning is relatively easy for users.

- [x] pending linux64+flat testing


test/io/ferguson/json.chpl
```diff
@@ -64,12 +64,13 @@ var f = opentmp();
...
 
-  var err:syserr;
-
-  var got = reader.readf("%~jt", r, error=err);
-
-  writeln("got is ", got, " error is ", errorToString(err));
+  try! {
+    got = reader.readf("%~jt", r);
+  } catch e: SystemError {
+    writeln("got is ", got, " error is ", errorToString(e.err));
```
mppf: I think this test should print out "got is " whether or not there was an error. You'll need to adjust the .good file.

test/modules/packages/Toml/test/fileTest.chpl
```diff
@@ -1,13 +1,14 @@
-
 use TOML;
 
 config const f: string;
 
 proc main() {
-  var error: syserr;
-  var TomlFile = open(error, f, iomode.r);
```
mppf: huh, I'm not sure why this code used the error argument at all, since it didn't check it.
It seems to me simply removing the error argument is more reasonable in this case, at least based on what I see here.

test/patterns/cltools/grep/grep.chpl
```diff
@@ -33,23 +33,29 @@ proc main(args: [] string) {
...
   // Search while handling errors
-  while(r.readline(line, errors)) {
-    if errors != ENOERR {
-      break; }
+  while (true) {
+    try {
+      r.readline(line);
+    } catch {
+      break;
+    }
```
mppf: Why not write this as
``` chapel
try {
  while r.readline(line) {
    if line.search(regEx) {
      ...;
    }
  }
}
```

Or, maybe even better, simply make fileGrep throw?

test/patterns/cltools/grep/grep.chpl
```diff
@@ -59,12 +65,12 @@ proc fileGrep(toFind: string, file: string) {
 
 
 /*
- Parallel grep: needs 
+ Parallel grep: needs
    - string to be searched for only
- Runs in parallel across any files in the current working directory. 
+ Runs in parallel across any files in the current working directory.
 */
 proc parallelGrep(tofind: string) {
```
mppf: Let's make it throw.

test/regexp/ferguson/readfre.chpl
```diff
@@ -19,8 +18,14 @@ writeln("#3 captures az, should match");
...
 writeln(r.offset(), " ", ok, ":", str);
 writeln("#4 should not match");
-ok = r.readf("%/(0-9)/", str, error=error);
-writeln(r.offset(), " ", ok);
+
+try {
+  ok = r.readf("%/(0-9)/", str);
+} catch {
+  ok = false;
+  writeln(r.offset(), " ", ok);
```
mppf: It's supposed to print out the offset either way; move it out of the catch block.

test/release/examples/primers/fileIO.chpl
```diff
@@ -274,19 +274,22 @@ Error handling
...
+    unlink(testfile);
 
-  // Who knows, maybe 1st unlink succeeds.
-  unlink(testfile, error=err);
-
-  // File does not exist by now, for sure.
-  unlink(testfile, error=err);
-  assert(err == ENOENT);
+    // File does not exist by now, for sure.
+    unlink(testfile);
```
mppf: Let's put an `assert(false); // never reached` here

test/release/examples/primers/fileIO.chpl
```diff
@@ -274,19 +274,22 @@ Error handling
 if example == 0 || example == 5 {
   writeln("Running Example 5");
 
-  // Error handling.
-  var err:syserr;
+  try! {
```
mppf: It's not so clear to me that `try! { } catch e: SystemError { }` is the right pattern to show in these primers. Certainly it is closest to the previous behavior, but does it represent a "best practice"? I suppose it's unclear what we would do with another sort of error anyway... and that's when `try!` is useful.

test/release/examples/primers/fileIO.chpl
```diff
@@ -274,19 +274,22 @@ Error handling
...
+  } catch e: SystemError {
+    assert(e.err == ENOENT);
+  }
 
-  // What happens if we try to open a non-existent file?
-  var f = open(testfile, iomode.r, error=err);
-  assert(err == ENOENT);
+  try! {
+    // What happens if we try to open a non-existent file?
+    var f = open(testfile, iomode.r);
```
mppf: Let's put an `assert(false); // never reached` here too

test/studies/labelprop/labelprop-tweets.chpl
```diff
@@ -187,10 +186,34 @@ proc process_json(logfile:channel, fname:string, ref Pairs) {
 
   if progress then
     writeln(fname, " : processing");
- 
+
   while true {
-    got = logfile.readf("%~jt", tweet, error=err);
-    if got && !err {
+    try! {
+      got = logfile.readf("%~jt", tweet);
```
mppf: if `readf` throws, then `got` won't be initialized here. I think you need to set it to `false` before the call here or as the first thing in the catch block.

test/studies/labelprop/labelprop-tweets.chpl
```diff
@@ -187,10 +186,34 @@ proc process_json(logfile:channel, fname:string, ref Pairs) {
...
+      if e.err == EEOF {
+        break;
+      } else if e.err == EFORMAT {
+        if verbose then
+            stdout.writeln("error reading tweets ", fname, " offset ",
+              logfile.offset(), " : ", errorToString(e.err));
+
+        // read over something else
+        try! {
+          got = logfile.readf("%~jt", empty);
```
mppf: this one definately shouldn't be in a `try!` - instead, failure here should output the "severe error" message.
Maybe it should use a two-level try/catch, where the inner level handles EOF/read over something, and the outer level prints out the sever error message?

test/release/examples/primers/fileIO.chpl
psahabu: This was my thought, that `try!` says "I don't expect any other errors here".

test/studies/labelprop/labelprop-tweets.chpl
```diff
@@ -208,8 +208,10 @@ proc process_json(logfile:channel, fname:string, ref Pairs) {
     } catch e: SystemError {
       if e.err == EEOF then break;
 
-      stderr.writeln("severe error reading tweets ", fname, " offset ",
+      try! {
+        stderr.writeln("severe error reading tweets ", fname, " offset ",
           logfile.offset(), " : ", errorToString(e.err));
+      }
```
psahabu: This `try!` was needed due to the bug described in #7130, and should be removed when that is closed.

PR #7109, opened 2017-08-25
Include inherited methods when collecting methods for forwarding

mppf:
This was an omission from collectVisibleMethodsNamed.
The new test shows a case where it was not working before.

Thanks to @LouisJenkinsCS for pointing out the issue.

- [x] full local testing


mppf:
@vasslitvinov - could you have a look at this short one too? thanks!

Issue #7013, opened 2017-08-15
unhandled error message format 


psahabu:
## today
Here is an example of unhandled error messages in Chapel:
```
unableToOpenFile.chpl:1: error: uncaught error - No such file or directory in open with path "_test_cannotOpenMe.txt"
```

This has a few deficiencies:
- redundant: `error: uncaught error`
- uninformative: not clear what type of error was unhandled
- minimal insight: no information beyond where the error was unhandled

A new design is needed to address these issues, but the best way forward is unclear and has many stakeholders.

### simple reformatting
There are a few different ways to go here, but here are some ideas:

- remove the leading error:
```
unableToOpenFile.chpl:1: uncaught error: No such file or directory in open with path "_test_cannotOpenMe.txt"
```

- include the error type:
```
unableToOpenFile.chpl:1: uncaught SystemError - No such file or directory in open with path "_test_cannotOpenMe.txt"
```

### stack trace
This opens up a lot of useful possibilities, but that means there's also a lot of possibilities to choose from. Starting from [Java's stack trace](https://en.wikibooks.org/wiki/Java_Programming/Stack_trace) seems reasonable, here's one possible example:
```
uncaught SystemError in thread 0 on locale 0:
    at unableToOpenFile.main() [unableToOpenFile:1]
    at [where the error was created]
```

mppf:
I'd generally lean towards the *stack trace* version, even if we only have 2 entries rather than a full stack trace (location error created & location it was not caught). I'm not certain of the value of listing thread and locale numbers there though.

lydia-duncan:
I also like noting where the error originally came from.  Would that end up being a hidden field/set of fields on the error type, or just something the compiler knows to do?

mppf:
@lydia-duncan - we'd just store the file name (or int id) and line number in the Error type. I don't think these necessarily need to be hidden fields. I'd expect it to be almost entirely handled in module code.

bradcray:
Is the stack trace capability redundant with CHPL_UNWIND?  (or: does it make CHPL_UNWIND redundant?)

mppf:
[...]

PR #7119, opened 2017-08-25
Improve uncaught error message

mppf:
See issue  #7013.

After this PR, uncaught error messages look like this:

```
uncaught MyError: test error
  e.chpl:5: thrown here
  e.chpl:9: uncaught here
```

It will take more effort to track functions. In particular,
there is a bit of a design question there because the function
where the error was created/thrown might (often) be deep within
the standard library. How can we show a function name while
avoiding reporting implementation functions? And how can
that work without actually doing a stack trace?

Additionally, I'm inclined to continue to make the error
store where it is thrown vs where it is created because
marking where it is created is connected to initializers &
keeping these efforts separate for now is better.

- [x] full local testing
- [x] test/errhandling with --llvm ; with --verify ; with GASNet


compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -1767,6 +1767,19 @@ static void buildRecordQueryVarField(FnSymbol*  fn,
...
 ************************************** | *************************************/
 
+static bool inheritsFromError(Type* t) {
+  if (t == dtError)
+    return true;
+
+  bool ret = false;
+
+  forv_Vec(Type, parent, t->dispatchParents) {
+    ret = ret || inheritsFromError(parent);
```
psahabu: Is it possible for there to be parents beyond `dtError` in `dispatchParents`? If so, maybe it'd be better to check if each parent is `dtError` and return true if it is, rather than going through the whole list.

modules/internal/ChapelError.chpl
```diff
@@ -177,12 +187,54 @@ module ChapelError {
...
+  pragma "insert line file info"
   proc chpl_uncaught_error(err: Error) {
-    var tmpstring = "uncaught error - " + stringify(err);
-    __primitive("chpl_error", tmpstring.c_str());
+    extern proc chpl_error_preformatted(c_string);
+
+    const myFileC:c_string = __primitive("chpl_lookupFilename",
+                                         __primitive("_get_user_file"));
+    const myFileS = myFileC:string;
+    const myLine = __primitive("_get_user_line");
```
psahabu: I might call these `uncaught` instead of `my` for clarity, and then swap them with the `thrown` section since that's the order they're printed in. It's fine as it is though.

mppf:
re parents in dispatchParents - normally there is 1 or 0 entries. It is a list because originally the compiler might've supported multiple inheritance. So it's actually the recursive call that's really useful; going through the dispatchParents list is probably going to only find 1 entry.

compiler/passes/buildDefaultFunctions.cpp
psahabu: Or is it possible for a parent's `dispatchParents` to contain classes that are not in the child's `dispatchParents`? In that case this loop makes sense.

mppf: yes, if you had FileNotFoundError inherit from SystemError inherit from Error, for example.

Issue #7106, opened 2017-08-24
Improve Mason


Spartee:
This issue is to gather all of the ideas to improve chapel's package manager Mason. If there is something you would like to add please comment it below. 

## Commands to add
- [ ] add (add dependency)
- [ ] rm (remove dependency)
- [ ] init (create project in existing directory)
- [ ] search (search mason-registry)
- [ ] doc (chpldoc project)
- [ ] publish (upload toml to mason-registry)
- [ ] bench (run benchmarks)
- [ ] test (run tests)
- [ ] examples (run examples)

## Features (needed)
- [ ] Chapel Versioning
- [ ] Use of other git platforms(bitbucket, gitlab, etc..)
- [ ] C dependencies
- [ ] Developer dependencies
- [ ] Resolution strategies for major versions
- [ ] Lazy versioning (e.g. 2.3.x)
- [ ] Build on last modified

## Mason-Registry
- [ ] More test directories
- [ ] Add tutorial of git commands for uploading a package
- [ ] Move registry off of github (centralized registry with cached packages)
- [ ] Blessed vs. open packages
- [ ] Checksumming
- [ ] Binary packages(applications)
 

## Tests 
- [ ] More multi-dependency testing
- [ ] Major version conflict testing (IVRS)
- [ ] A method of testing that does not involve cloning entire registry

## Interaction with TOML library (See  #7104)
- [ ] Use of array of tables for dependencies once that functionality is put in
- [ ] Make [root] print at the top of Mason.lock

## Features (just cool)
- [ ] Curses loading bar for parallel downloads
- [ ] Color for standard output

Spartee:
In reference to one of the features listed in this issue, there has been discussion of how mason will handle multiple versions of chapel. So far, the idea that is the most favorable amoung team members who have discussed this is to add a field to that project `mason.toml` that indicates chapel version. 

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.16.0"

```

We have also proposed the idea of using chapel range syntax to indicate multiple compatible versions as follows

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.13.0..1.16.0"

``` 

We are welcoming all ideas and thoughts on this topic.

ben-albrecht:
> Extend mason doc to document dependencies

I'm wondering if this is really that useful of a feature to support..

benharsh:
[...]

PR #7124, opened 2017-08-25
Mason doc, runtime/compile args, TOML update

Spartee:
This Pr includes three features: an update to the documentation on runtime and compile time args; the mason doc command, and an update to the TOML library.  Two of these features are listed and will by crossed off the mason issue #7106. 

First, an update to the documentation about how mason now supports forwarding all compile time and runtime arguments with examples that are hopefully clear. This was one of the first bits of feedback we got, so it was important that it got documented well.

Second, a first pass at a ``mason doc`` procedure has been introduced. It does not currently document dependencies, but does document the main file of the project. Documenting dependencies will be a "next step".

Third, the toml library has been adjusted so that if any Mason.lock file is generated, we now ensure that, for readability, the main project signified by "[root]" will always be printed at the top of the lock file. 

All local testing had passed for these changes.

### DIFF
--- a/doc/rst/tools/mason/mason.rst
+++ b/doc/rst/tools/mason/mason.rst
@@ -79,7 +79,7 @@ When invoked, ``mason build [ options ]`` will do the following:
 
     - Invoke build.
     - Run the resulting executable out of ``target/``, if it exists.
-    - All options not recongized by ``mason`` will be forwarded to the executable.
+    - All options not recognized by ``mason`` will be forwarded to the executable.
         
 For example, after ``mason run [ options ]``, the project directory appears as so::
 
@@ -137,7 +137,7 @@ To try out different values at runtime, pass the values for ``number`` to ``maso
 
    For the case when a flag intended for the ``chpl`` compiler or executable is recognized by 
    ``mason build`` or ``mason run``, respectively, the flag can be thrown after ``--`` 
-   to override this conflict. For example, ``mason run -- -nl 4``. Intsead of mason recognizing
+   to override this conflict. For example, ``mason run -- -nl 4``. Instead of mason recognizing
    this argument, this command will run the executable over 4 locales.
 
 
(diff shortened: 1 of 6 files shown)


tools/mason/mason.chpl
```diff
@@ -116,10 +117,28 @@ proc masonClean() {
   runCommand('rm -rf target');
 }
 
+ 
+proc masonDoc(args) {
+  var toDoc = basename(getEnv('PWD'));
+  var project = toDoc + '.chpl';
+  if isDir('src/') {
+    if isFile('src/' + project) {
+      var command = 'chpldoc src/' + project;
```
ben-albrecht: Note: `mason` will now depend on `chpldoc`, so this dependency should be marked in the top-level `Makefile`

doc/rst/tools/mason/mason.rst
```diff
@@ -93,8 +93,38 @@ For example, after ``mason run [ options ]``, the project directory appears as s
 	  MyPackage
 	
    
+Projects can also be multiple files. Mason handles this by forwarding all arguments
```
ben-albrecht: Note that projects can be multiple files without getting compiler flags involved. `-M` is actually not necessary for the example below, but would be example for a case more like:

```
MyPackage/
  Mason.toml
  Mason.lock
  src/
  MyPackage.chpl
  utils/
    MyPackageUtils.chpl
```

ben-albrecht: I think it'd be good to document the handling of compiler flags / execution flags as a standalone topic. The usage of `-M` to use modules in subdirectories seems more FAQ-ish than appropriate for documentation. An FAQ at the bottom of the docs wouldn't be a bad idea.

Spartee: I think this is too important of a topic to push to the bottom of the documentation before we have compile flags in ``Mason.toml``.

PR #7125, opened 2017-08-25
Update collectVisibleMethodsNamed to handle generic inheritance

mppf:
When a type inherits from a generic type, collectVisibleMethodsNamed
should not add generic methods from the parent type in two ways
(inheritance/instantiation as well as instantiation/inheritance).
Additionally, overridden methods should not be added in
collectVisibleMethodsNamed.

Adds a new test that shows the previous problem and now passes.

Thanks to @LouisJenkinsCS for pointing out this issue.

- [x] passes test/classes/ferguson/forwarding
- [x] full local testing


mppf:
@vasslitvinov - here is another forwarding fix for you. thanks!

PR #7126, opened 2017-08-25
Fix compilation infinite loops for default initializers

lydia-duncan:
Turns out the timeouts/compiler segfaults for paratests with --force-initializers
were due to an infinite loop after the handling of initializers on generic types
changed, because the "this" argument wasn't getting labeled appropriately.

Resolve this error by attaching the FLAG_DELAY_GENERIC_EXPANSION to the "this"
arg and "super" field when each is generic.

Allows an additional 14 tests to pass, and many more to change their error mode.

Also fix an additional issue with the error message when skipping an actual for a
param field defined without an initial value (but with an initial type).  Turns out I
shouldn't insert a defaultExpr for those, and that the typeExpr should be a
BLOCK_TYPE instead of a BLOCK_SCOPELESS.  This is a little annoyingly mystical
but I don't want to deal with it right now.

Adds eight tests to lock in default initializer behavior with generics.  Prior to the
changes fixing the handling of default initializers on generic types, 4 of these tests
failed and 4 of them passed.  Now, all of them pass.

Covers:
- Initializing a generic type with a default initializer via `var foo = new Foo(args);` when
  - the type is generic due to a generic var field
  - the type is generic due to a generic type field with no initial value
  - the type is generic due to a generic type field with an initial value
  - the type is generic due to a generic param field with no initial value or type
  - the type is generic due to a generic param field with an initial value but no type
  - the type is generic due to a generic param field with no initial value but a type
  - the type is generic due to a generic param field with an initial value and a type
  - the type is generic due to a generic param field with no initial value but a type and
    the new call contains no arguments

Testing: passed classes/initializers with futures, a full standard paratest, running
a full standard paratest with --force-initializers


lydia-duncan:
These passed testing prior to a rebase, but I'm rerunning afterwards just to be certain

lydia-duncan:
Sanity recheck complete!

Issue #7046, opened 2017-08-17
error handling: errors in foralls may or may not be in ErrorGroup


mppf:
Consider a program like forall-range-throws.chpl:
``` chapel
config const n = 1000;

proc test() {
  try {
    writeln("before forall block");
    forall i in 1..n {
      if i == 1 then
        throw new Error("test error");
    }
    writeln("after forall block");
  } catch errors: ErrorGroup {
    for e in errors {
      writeln("Caught group error e ", e);
    }
  } catch e {
    writeln("Caught other error ", e);
  }
}

test();
```

Are the errors thrown in the body of the forall in an ErrorGroup? Or not? Or nested in two ErrorGroups?

Currently, it depends on the implementation of the iterator that is being invoked by the forall statement. If it does not create any tasks / include coforall/cobegin, it won't use an ErrorGroup.

The test forall-range-throws.chpl is skipped in the NUMA configuration due to this issue.

What should we do?

mppf:
@psahabu what do you think?

bradcray:
My intuition is that the forall should always use an ErrorGroup for consistency from the user's perspective.  What would the downsides of that approach be?

psahabu:
I'm with Brad, for consistency I think it's best if an ErrorGroup was always used, but there may be implementation or design issues making that difficult.

mppf:
Right, consider this example:
``` chapel
 forall i in 1..n {
    throw new Error("test error");
  }
```

as compared to this example 
``` chapel
forall i in 1..n {
   coforall j in 1..m {
        throw new Error("test error");
   }
}
```

Should the 1st case throw an ErrorGroup of Errors, but the second case throw an ErrorGroup of ErrorGroups? If so, we'll have to distinguish between whether a coforall in a forall is from an iterator or from the body of the loop.

mppf:
My current favorite strategy for this issue is to do the following two things:
 1. make ErrorGroups not support nesting. Adding an ErrorGroup A to an ErrorGroup B will add the contents of B to A.
 2. Adjust the compiler to add a call to ensure that any Error returned by a forall is in an ErrorGroup

I discussed this with @psahabu and we're on the same page. We don't think that supporting nesting in ErrorGroups adds much value, and this strategy greatly reduces the complexity of the adjustment the compiler needs to make for forall statements to be more uniform in what they can throw.

bradcray:
This plan matches my intuition as well.

mppf:
[...]

PR #7127, opened 2017-08-26
rename ErrorGroup to TaskErrors

psahabu:
`coforall` and `forall` will eventually emit a flat list of errors from tasks (#7046), so this PR changes the name of that list to `TaskErrors` instead of `ErrorGroup`. Note that this PR does not flatten the errors from `coforall` and `forall`.

In the future, an `ErrorGroup` may be made available to users as a way to collect errors separate of the compiler-inserted `TaskErrors`, either as the parent class of `TaskErrors` or as a standalone class.


mppf:
Looks good, once you fix the conflicts. Might be nice to say in the PR message that this does only the renaming and doesn't flatten them yet.

PR #7129, opened 2017-08-26
Remove undefined behavior from a test

dmk42:
There was undefined behavior in `test/marybeth/casts` that was being used to detect the undefined behavior of a signed overflow.  This change replaces the signed integers with unsigned integers, eliminating the undefined behavior but still achieving the intended test.

### DIFF
--- a/test/spec/marybeth/casts.chpl
+++ b/test/spec/marybeth/casts.chpl
@@ -9,18 +9,6 @@ writeln(z);
 
 var m = 2: int(64);
 var n = 2: int(32);
-var i = 1;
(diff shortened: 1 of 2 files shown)


dmk42:
This PR will allow #7092 to be merged.

mppf:
@dmk42 - this is fine with me, but I wonder if it'd be better to adjust the test to use unsigned integers, for which overflow is not undefined behavior?

dmk42:
@mppf - agreed.  I've changed it to use the same method as the original but with unsigned integers.

Issue #6998, opened 2017-08-11
Dynamic dispatch of overloaded leader-follower iterators causes internal error


LouisJenkinsCS:
### Summary of Problem

If you wanted to overload a base class to provide parallel iteration, but cast it to it's parent type, an internal error is produced.

### Steps to Reproduce

**Source Code:**

[TIO (Online Compiler)](https://tio.run/##tVDRSsQwEHw2X7H0qQWvoL5d6Rcc@HL3AzHZNoE1KUm8Ikd/3bhnOEVBi6Bvy87szM4oIyeknBXJGGGSAV2Ck7iyCQMkgxFrXspHSHKELZzXO@t0c56ZORsM@Ib1/TvYEkrN5ycwklJd7Y1/Ig3OJ3hAUJIIddu2VdPBs0VGbrpl1fEaBk/k54Ox8Wf3wvuN/@0X/4v@usAiFiFKd8pY5mz/vMN/rmitgYLflaRHGWD6yNiDw7kEr5tODD6A5TOmaA9zYD1ytS0Il/YNeJmrezzyAyOm@KljkfOLGkiOMW82mjn0Cg)

```chapel
class parent {
	iter these(param tag : iterKind) : int where tag == iterKind.leader { halt("Should not be called..."); yield 1;}
	iter these(param tag : iterKind, followThis) : int where tag == iterKind.follower { halt("Should not be called..."); yield 2;}
	iter these() : int { halt("Should not be called...");}
}

class child : parent {
	iter these(param tag : iterKind) : int where tag == iterKind.leader { yield 1;}
	iter these(param tag : iterKind, followThis) : int where tag == iterKind.follower { yield 2;}
	iter these() : int { yield 3;}
}

var p : parent = new child();
for i in p do writeln(i);
forall i in p do writeln(i);
writeln("Never gets called...");
```

**Output (w/ --devel flag):** 
```
internal error: no line number available [baseAST.cpp:296]
```

**Output (w/o --devel flag):**
```
internal error: BAS0296 chpl Version 1.16.0 pre-release (-999)

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.
```

ronawho:
@vasslitvinov and/or @daviditen could you look at this one?

[...]

PR #7136, opened 2017-08-27
GSoC Collections Module Improvements

LouisJenkinsCS:
* Added a lot of documentation
* Added new utility methods `addBulk` and `removeBulk` along with the affiliated unit test
* Collections are now 'record-wrapped' and implicitly use privatized versions of data, meaning we now get all of the benefits of privatization without requiring the user to care. We use `SharedObject` for reference counting so that it can be passed around like a class can.

**Disclaimer:** These revisions brought out a few more compiler bugs, I will gradually work to create some official MWEs or make changes to side-step these in the near future.


mppf:
Travis is complaining about this:
> Found tabs :-(
>  modules/packages/DistributedBag.chpl:242:

modules/packages/Collection.chpl
```diff
@@ -18,118 +18,212 @@
...
+    }
+
+    /*
+      Determine whether an element exists in this collection.
+    */
+    proc contains(elt : eltType) : bool {
+      var found = false;
+
+      // Iterate to find it... note that we cannot break early due to
+      // issues with leaking during serial iteration.
```
mppf: I know this isn't new in this PR, but why shouldn't `contains` use parallel iteration?

test/modules/packages/collections/CollectionBulk.skipif
```diff
@@ -0,0 +1,2 @@
+# 'fifo' tasking layer results in a double free in the runtime of a pthread_condvar. Works on Qthread
+CHPL_TASKS == fifo
```
mppf: I think this file needs a newline

test/modules/packages/collections/NUMLOCALES
```diff
@@ -1 +1 @@
-4
+4
```
mppf: This one also needs a newline

modules/packages/Collection.chpl
LouisJenkinsCS: Chapel currently has issues allowing overloading of parallel iterators, #6998

modules/packages/DistributedBag.chpl
```diff
@@ -64,631 +62,788 @@ use BlockDist;
...
+
+    // Privatized id...
+    pragma "no doc"
+    var _pid : int = -1;
+
+    // Reference Counting...
+    pragma "no doc"
+    var _rc : Shared(DistributedBagRC(eltType));
+		
+		pragma "no doc"
```
mppf: Spacing here seems weird

LouisJenkinsCS: Oh wow... I didn't even see that in my editor.

modules/packages/DistributedBag.chpl
```diff
@@ -64,631 +62,788 @@ use BlockDist;
...
+    proc DistBag(type eltType, targetLocales = Locales) {
+      _pid = (new DistributedBagImpl(eltType, targetLocales = targetLocales)).pid;
```
mppf: I would have expected this constructor to initialize _impl as well

modules/packages/DistributedBag.chpl
LouisJenkinsCS: My local copy doesn't show this. Git doesn't pick up on it either

mppf: I'd delete the line, add it back in, and see if that solves the issue

modules/packages/DistributedBag.chpl
LouisJenkinsCS: `_impl` is ignored for now, I merely have it for documentation sake. We obtain the privatized copy each time in `_value` and use that. I needed it because `forwarding` shows weird unhelpful stuff in `chpldoc`

modules/packages/DistributedDeque.chpl
```diff
@@ -59,638 +57,709 @@ use Collection;
...
+
+    // Privatization id
+    pragma "no doc"
+    var _pid : int;
+    // Reference counting
+    pragma "no doc"
+    var _rc : Shared(DistributedDequeRC(eltType));
+
+    proc DistDeque(type eltType, cap = -1, targetLocales = Locales) {
+      _pid = (new DistributedDequeImpl(eltType, cap, targetLocales)).pid;
```
mppf: also here, I would have thought that _impl would be initialized.

modules/packages/DistributedDeque.chpl
```diff
@@ -59,638 +57,709 @@ use Collection;
...
-            // If it is empty, fix the head...
-            if tail != nil {
-              tail.next = nil;
-            } else {
-              head = nil;
-            }
+    pragma "no doc"
+    inline proc these(param order : Ordering = Ordering.NONE, param tag) where
+        (tag == iterKind.leader || tag == iterKind.standalone) &&
+        __primitive("method call resolves", _value, "these", tag=tag)
```
mppf: You shouldn't need the "method call resolves" business here. And I don't remember seeing these forwarding  'these' in DistributedBag but maybe I missed them. Anyway, you know if you implemented leader/follower or standalone so you can just directly create the versions that you implemented, instead of using "method call resolves". That is a workaround appropriate for the internal array code.

modules/packages/DistributedBag.chpl
mppf: I think in that case it'd be more reasonable to chpldoc _value but I can believe that the _impl version might look nicer. Anyway if that's the case you can add a `//` comment to say that this field exists only for documentation purposes.

mppf:
Now travis says
> Found tabs :-(
> modules/packages/DistributedBag.chpl:237: pragma "no doc"

modules/packages/DistributedBag.chpl
LouisJenkinsCS: That worked! Thanks, wouldn't have been able to spot it on my own.

mppf:
PR looks pretty good to me! I looked at the forwarding changes & commented on anything I noticed that seemed odd. I didn't review the documentation changes though. (For some reason, the diff is really quite hard to read...).

mppf:
I won't be able to merge it this week, perhaps @ben-albrecht can help? Also ideally @e-kayrakli can review the documentation changes I didn't get to. 

Before merging, needs testing of the relevant test/packages/collections directory at least.

modules/packages/DistributedDeque.chpl
LouisJenkinsCS: After removing and testing, it turns out they are needed, I get `unresolved call 'DistributedDequeImpl(int(64)).these(Ordering, tag=iterKind)'` otherwise.

e-kayrakli:
I will give this a look (especially the documentation parts) later today.

modules/packages/Collection.chpl
```diff
@@ -18,118 +18,212 @@
...
-  A 'Collection' is a data structure, a container for elements that provide support
-  for insert, lookup, remove, and iteration operations.
+  
+  Summary
+  _______
+
+  A 'Collection' is an interface for a data structure that is a 'contract' between
+  the user and implementer that it fulfills the following:
+
+  1. Is safe parallel-safe, hence is safe to use across multiple tasks across multiple locales.
```
e-kayrakli: I am not sure if Collection provides parallel-safety guarantee as it is now. Nor I think it should. i.e. I think Collection can be inherited by a data structure that is not parallel-safe.

modules/packages/Collection.chpl
```diff
@@ -18,118 +18,212 @@
...
+      rejects an element, we cease to offer more. We return the number of elements
+      successfully added to this data structure.
+
+      **Warning:** While this method will add as many items as possible, it will `break`
+      when it is unable to consume more elements. Due to the fact that using `break` in
+      an iterator will cause a memory leak and potentially leave this data structure in
+      a undefined state, this should not be used to append another data structure to a
+      bounded structure unless it is known it will succeed.
+    */
+    proc addBulk(elts) : int {
```
e-kayrakli: I think I prefer `bulkAdd` (similarly `bulkRemove`). That's what sparse domains use. Sparse domains and collections are not closely related, but nonetheless..

modules/packages/Collection.chpl
```diff
@@ -18,118 +18,212 @@
...
+        arr[idx] = elt;
+        idx += 1;
+      }
 
-  /*
-    Obtain the number of elements contained in this collection.
-  */
-  inline proc getSize() : int {
-    halt("'proc size() : int' is not supported...");
+      return arr;
```
e-kayrakli: I wonder if you can return an array slice safely here. It might deliver better performance as you probably won't need to shrink the array, which is obviously costly.

modules/packages/DistributedBag.chpl
```diff
@@ -64,631 +62,791 @@ use BlockDist;
...
+      Triggers a more static approach to load balancing, fairly redistributing all
+      elements fairly for bags across nodes. The result will result in all segments
+      having roughly the same amount of elements.
+      
+      .. note::
+
+        This method is very heavy-weight in that it should not be called too
+        often. Dynamic work stealing handles cases where there is a relatively fair
+        distribution across majority of nodes, but this should be called when you have
+        a severe imbalance, or when you have a smaller number of elements to balance.
```
e-kayrakli: "or when you have a smaller number of elements to balance" sounds a bit contradictory to what you are saying before.

modules/packages/DistributedBag.chpl
```diff
@@ -64,631 +62,791 @@ use BlockDist;
...
+    /*
+      Iterate over each bag in each node. To avoid holding onto locks, we take
+      a snapshot approach, increasing memory consumption but also increasing parallelism.
+      This allows other concurrent, even mutating, operations while iterating,
+      but opens the possibility to iterating over duplicates or missing elements
+      from concurrent operations.
+
+      .. note:: 
+
+        `zip` iteration is not yet supported with rectangular data structures.
```
e-kayrakli: Let's rewrite this. Something like "This data structure cannot be used in a zippered iteration." Just to make it more general.

modules/packages/DistributedDeque.chpl
```diff
@@ -59,638 +57,712 @@ use Collection;
...
+      or by making the lock reentrant and forcing all iterators to acquire all locks in
+      some global locking order.
+
+      **FIXME:** Ordered serial iterators currently are not working well when the head or tail are negative
+      (because the head and tail may need to move backwards, its easily possible for it to be negative. We
+      handle this right now with the core methods, but iteration wasn't the focus during development...), 
+      but there is not enough time to fix it. This is an official warning against using anything but 
+      iteration when `order` == `Ordering.NONE` (A.K.A the default unordered iteration). 
+
+      **TODO:** Need to create a parallel iterator (not really parallel, just allowed to be used in a
```
e-kayrakli: I think parallel unordered iteration *can* be allowed. If a user uses a deque as an iterand in a `forall` I think we can assume that they want parallel iteration and by nature, it is unordered. Similarly reduction can be unordered, as well.

test/modules/packages/collections/CollectionBulk.compopts
```diff
@@ -0,0 +1,3 @@
+-sisDeque --baseline
+-sisBag --no-loop-invariant-code-motion
+-sisBoundedDeque --baseline
```
e-kayrakli: Why do we need `--baseline` here? If it is b/c of a bug, I think you should a reference to it, otherwise, let's remove it.

modules/packages/DistributedDeque.chpl
LouisJenkinsCS: By default, parallel iteration *is* unordered. My complaint is that there is no way to reduce it in a way that is ordered.

"Why would it need to be ordered?" - Imagine the following...

```chpl
var deque = new DistDeque(string);
deque.pushBack("World");
deque.pushFront("Hello");
writeln(+ reduce deque);
```

This will not be in any ordering at all. Reduction isn't only useful for summation, sometimes ordering can matter, I would like to the user to be able to do `+ reduce deque.these(ordering.FIFO)` or if thats not possible syntactically:

`deque.reduceFIFO(lambda(s1 : string, s2 : string) : string { return s1 + s2; });`

That would be pretty cool overall, and very useful.

modules/packages/Collection.chpl
LouisJenkinsCS: Eh... I guess relaxing this property means more Collections, sure. I can't wait until we get proper interfaces and inheritance of said interfaces though.

modules/packages/DistributedBag.chpl
LouisJenkinsCS: Can I potentially defer this until next commit? I have a ton of work to do now that school started, I'll promise to come back to it later (as comment said its incremental revision)

modules/packages/DistributedDeque.chpl
e-kayrakli: I think my main concern in the comment was "not really parallel" part. And my question is "why not?"

In terms of ordering in reduction:

First, you cannot use an iterator of your choosing, nor pass an argument to `these` while doing reduction. It simply is not in the grammar.

Second, the use case you are presenting is not a very realistic use case for reductions. And I can't think of one, myself. I think reduction operations should always be commutative, so ordering is not necessary. Even if this wasn't the case, I think this is a bigger question than your implementation of the parallel iterator, and parallel iterator should still be unordered and *parallel*.

What you are showing here needs to be achieved by a scan operation and not by reduction. And scan operations currently use serial iterators anyways.

modules/packages/DistributedBag.chpl
e-kayrakli: This one is probably the simplest one :) But anyways, I think it would be nice if it is addressed before the release at least (sometime in the Fall, I guess)

modules/packages/DistributedDeque.chpl
LouisJenkinsCS: I suppose I was a bit unclear, but what I meant in my message is that we need *another* parallel iterator, that wasn't parallel and just yielded sequentially in order. However, I realize now that this isn't needed, as reductions are supported if you declare the iterator to use, like [this](https://tio.run/##dVHBaoRADD3rVzy8qFT2A1Y8Fhb2UtrCHkoPwYl1IOvIaGtL6bfbGbWuPXgJk7z3kpdJWVPLMo5hKdR10D3b@0@6tsL4DgOfoq@54yT1eVAZi7KGbhCdWMTgYqyoCMrgS7MoB@Zh8BMu0kpXxivxT/rXcCvCRjXjLVm6oqc3HCdbZ92oFEPNlqdqUazlgzApJ/QOP8iCrEWBl/gUZ4jZB1mD8QE@XNbUrqiKX/PZK4ksdn23nf32nWaojPuf4bnW3Z7rmbH4ntvfRNMYN8jvo3u3TsPD9jxJmofhYF1FmiR6Kqk5Ispwh849HdHBN5StJsEjq/eSF5qdEkc8zEfa8h/Ir897Ckcdx18).

As well, in terms of reduction, all languages I have seen so far have reductions be ordered on an ordered data structure. For example, see Haskell's [fold](https://wiki.haskell.org/Fold). I was merely suggesting providing a means to reduce in order. `scan` is nice and all, but it has an O(N) space overhead, especially when you only need the last value produced.

modules/packages/DistributedBag.chpl
LouisJenkinsCS: Speaking of, when is the next release? You're right in that I need to have my data structures improved by then, but I'd like to know when.

modules/packages/Collection.chpl
LouisJenkinsCS: Will do.

modules/packages/DistributedDeque.chpl
LouisJenkinsCS: Following this, its unfortunately I can't just do it this way, but if I were to provide this (big 'if' at this point), I would likely do so in parallel, and assume that all values are associative. Like, for example, imagine if I emulated a block distribution instead of cyclic, then I would have `maxTaskPar` adjacent `LocalDeque` that maintain a relative ordering...

```
LD1      LD2    LD3       LD4
-H1-    -H2-   -H3-     -H4-
-H5-    -H6-   -H7-     -H8-
-H9-    -H10- -H11-   -H12-
```

Where 'LD#' is a `LocalDeque` index, and `H#` is the index of each element from the Head. (This is how ordering is preserved). Reduction can take advantage of this by reducing each 'row' from each 'column', producing sub-results which get combined with sub-results from other nodes, etc. It remains ordered, and it is still parallel.

modules/packages/Collection.chpl
mppf: Github isn't showing me the related code in a way that I understand, but generally speaking array slices return by value right now.

modules/packages/DistributedBag.chpl
mppf: Feature freeze is Sept 18th, doc/code freeze is Sept 25th.

mppf:
@LouisJenkinsCS - what's the current status of this PR, from your point of view?

modules/packages/DistributedBag.chpl
LouisJenkinsCS: @mppf 

This month?! So soon after GSoC too... If you were asking whether or not this should be in official release, I'd say 'no'. I can't commit enough time to have it completely bug-free, not to mention there are other gating issues as well which would require my code to change once they are resolved. Everything about it hits places which are bug-ridden, both in terms of the implementation and due to Chapel itself, and I can imagine users running into many more that I haven't in the limited coverage of my tests.

With that said, the data structures make amazing additions to the language, but if anything I'd label them 'experimental' right now, and should remain in modules/packages.

test/modules/packages/collections/CollectionBulk.compopts
LouisJenkinsCS: I believe I used `--baseline` as I was experiencing some issues with non-deterministic behavior even with just '--no-loop-invariant-code-motion', but now that I've re-tested it, it seems to work pretty well. I believe the issue had to do with `fifo` tasking layer and I forgot to remove it, fixing it right now. (Gonna make these fixes in small incremental patches).

mppf:
- [ ] full local testing

mppf:
@LouisJenkinsCS I'm seeing timeouts with quickstart+gasnet configuration that didn't appear to occur before this PR. Could you look in to it?

mppf:
Well, the fifo deadlock potential comes about when there are many more tasks than threads. The fifo tasking layer multiplexes the tasks onto threads, but it's not smart enough to always swap out idle/waiting tasks.
Earlier, we had some problems when the code was creating numlocales**2 threads. I'm guessing that there's no reason your benchmark codes need to create > # locales * num cores tasks. Am I wrong about that? FIFO should work OK if the number of tasks is approximately the number of cores.

modules/packages/DistributedDeque.chpl
```diff
@@ -705,168 +762,300 @@ class DistributedDeque : Collection {
...
-      slot.lock$;
+      followThis.lock$;
     }
-  }
 
-  iter these(param order : Ordering = Ordering.NONE) : eltType where order == Ordering.FIFO {
-    var localThis = getPrivatizedThis;
+    pragma "no doc"
+    proc Destroy() {
+      for slot in slots do delete slot;
```
LouisJenkinsCS: @mppf 

Hey Michael, can you give me some feedback on this. Would this potentially cause an issue since about `maxTaskPar * (numLocales - 1)` will be remote, hence resulting in `on` statements? I comment this out and it works fine for 'fifo', although there is nothing inherently wrong here (as we do not acquire any locks) and this is done sequentially, so at most we only have 2 tasks running...

mppf: I don't see anything wrong with it either...

mppf:
@LouisJenkinsCS - this patch seems to fix it for me:
```
diff --git a/modules/packages/DistributedDeque.chpl b/modules/packages/DistributedDeque.chpl
index 4e14183..655b6e9 100644
--- a/modules/packages/DistributedDeque.chpl
+++ b/modules/packages/DistributedDeque.chpl
@@ -767,7 +767,7 @@ module DistributedDeque {

     pragma "no doc"
     proc Destroy() {
-      for slot in slots do delete slot;
+      for slot in slots do on slot do delete slot;
       delete globalHead;
       delete globalTail;
       delete queueSize;
```

The only thing is, I'm at a loss as to explaining why that fixes it.
You have an `on` statement already in the LocalDeque destructor. Without the on statement I added above, there are 2 on statements to the remote locale:
 * run the destructor
 * free the memory

So, it's less communication to do that in 1 on statement. But... I suspect there is a bug here somewhere. I'm also really surprised not to see the 'free' call running with a fast executeOn.

mppf:
@LouisJenkinsCS - let's add my workaround to this PR and then I can merge it. I'll make an issue about the deadlock, and @ronawho is planning to look at it. Thanks!

mppf:
- [ ] quickstart + gasnet
- [ ] qthreads + gasnet
- [ ] qthreads local

gbtitus:
Just a comment on this from @mppf earlier:

> Well, the fifo deadlock potential comes about when there are many more tasks than threads. The fifo tasking layer multiplexes the tasks onto threads, but it's not smart enough to always swap out idle/waiting tasks.

Fifo doesn't multiplex tasks on threads.  A given thread can run more than one task but only one after another, serially.  As a special case the thread that runs the parent task of a parallel construct such as a coforall can also one one or more of the child tasks of that construct, but it has to run such a child task to completion before returning to running the parent.  It can't switch between two tasks in any other way.  Basically fifo assigns each task to a thread for the duration of the task's life, and switches tasks by having the kernel switch the pthreads.  This makes fifo nice and simple, but it also limits its applicability and in some cases its performance.

PR #7145, opened 2017-08-28
Fix memory leak by avoiding insertion of PRIM_DEREF in flattenFunctions

benharsh:
The modified portion of flattenFunctions was inserting PRIM_DEREFs for
record-wrapped formals for some reason that no longer applies in today's
compiler. Improvements to `updateAssignmentsFromRefTypeToValue` in the
callDestructors pass caused the compiler to insert an autoCopy for this
PRIM_DEREF without a matching autoDestroy.

The following test was most negatively impacted by the memory leaks:
optimizations/bulkcomm/bharshbarg/stridedAssignDR.chpl

Adds a new test case to make it easier to catch this issue in the future.

Testing:
- [x] full local
- [x] full gasnet
- [x] valgrind release/examples


benharsh:
@vasslitvinov : The "crash2VarSymbolTypeNull" test failed on this branch due to a bug that already exists on master. The bug doesn't yet have a future test, but I will create one soon (hopefully with a fix).

From what I can tell, this test was created in 2011 to test a fatal compiler error. It doesn't seem like the error has been present for a long time, and it seems plausible that the compiler has changed significantly since the test was first created such that this test is no longer useful.

In short, are you OK with the removal of this test?

vasslitvinov:
@benharsh Sounds good, assuming you create a new .future.

benharsh:
Thanks, I fixed the issue and added a new test in #7150

PR #7147, opened 2017-08-28
Shadow var symbol for forall intents

vasslitvinov:
This PR replaces ForallIntent class and corresponding code.
It continues on the path defined in #6514 for representing forall statements.

ShadowVarSymbol a subclass of VarSymbol so that the rest of the compiler
downstream from resolution works as before.

A shadow variable within the forall loop body is represented by
a ShadowVarSymbol:

* upon scopeResolve(), if it is given a forall intent explicitly
in a with clause,

* upon implementForallIntents1New(), otherwise.

The expressions ShadowVarSymbol::outerVarRep and specBlock
do not have a parentExpr. This is unexpected by the compiler.
To deal with this, specBlock is a BlockStmt containing just
the specification expression. Instead of having specBlock
be the expression itself.
Also, I wipe out those two fields at the start of lowerIterators
because otherwise when the latter performs flattenFunctions,
it tries to propagate the variables that these two fields reference.
That's does not necessarily work and in any case is useless.

Another new-ish thing is that their DefExprs are not part of any BlockStmt.
Instead those DefExprs are on the list ForallStmt::intentVariables().
In that sense they are similar to ForallStmt::inductionVariables().
So I adjust scopeResolve in a similar manner.

While there, did some renamings as suggested by Michael.
Removed no-longer-needed ForallStmt::forallIntents() and fFIntents.

Key future work items specific to this change are:

* createShadowVarsNew() marks certain shadow vars to be "pruned",
then pruneIntentVars() does "pruning". Instead, do not prune at all.

Right now we "prune" those variables whose forall intent is ref or
INTENT_REF_MAYBE_CONST. Instead these variables are handled while
flattening lexical scopes. We probably want to handle them the same
way as other intents i.e. by threading through the parallel iterator.

Unless we handle 'ref' forall intents uniformly, pruning will be needed.
For example, if user gives an explicit 'ref' intent to the variable 'x',
it will be represented by ShadowVarSymbol inside the loop body. So if
we decide not to propagate 'x' through the leader, the shadow 'x'
needs to be replaced by the outer variable in the loop body.

* isOuterVarNew() essentially prunes out those outer variables
that are "const" (FLAG_CONST). This prevents consts to be passed
by non-const 'in' intent. Should be fixed.

* Extend visitors to handle ShadowVarSymbol explicitly.

* In resolveVarIntent(), also check sym->isConstant() and set one of CONST qualifiers.

* Refactor the code for clarity.

Testing passes: linux64 --verify; gasnet


compiler/include/symbol.h
```diff
@@ -91,6 +91,20 @@ typedef std::bitset<NUM_FLAGS> FlagSet;
...
+  TFI_DEFAULT, // aka TFI_BLANK
+  TFI_CONST,
+  TFI_IN,
+  TFI_CONST_IN,
+  TFI_REF,
+  TFI_CONST_REF,
+  TFI_REDUCE,
+};
+
+const char* tfiTagDescrString(ForallIntentTag tfiTag);
```
mppf: getForallIntentTagDescription ?

compiler/include/symbol.h
```diff
@@ -361,6 +379,45 @@ class ArgSymbol : public LcnSymbol {
...
+  // The corresponding outer var or NULL if not applicable.
+  SymExpr* outerVarSE() const { return (SymExpr*)outerVarRep; }
+  bool     isReduce()   const { return intent == TFI_REDUCE;  }
+  Expr*    spec()       const;
+  void     removeSupportingReferences();
+
+  ForallIntentTag intent;
+
+  // This would be a SymExpr*, if not for checkIdInsideWithClause().
+  // See also: sv->outerVarSE() and outerVarSym(sv).
```
mppf: Huh? outerVarSE casts it to a SymExpr. Is that invalid in some situation? Should it use toSymExpr? It seems to me that it should be a SymExpr.

compiler/include/symbol.h
```diff
@@ -361,6 +379,45 @@ class ArgSymbol : public LcnSymbol {
...
+  Expr* outerVarRep;
+
+  // For a reduce intent, the reduce expression.
+  // For a task-private variable, the initialization expression.
+  // Either way, wrapped in a block.  Otherwise it is NULL.
+  BlockStmt* specBlock;
+
+  Symbol* reduceGVar;
```
mppf: I don't know what reduceGVar is. Is that the "global" variable? Maybe name it reduceTotal or something like that?

compiler/include/symbol.h
```diff
@@ -676,9 +731,16 @@ void initAstrConsts();
 // pass-by-reference intents are used.
 bool argMustUseCPtr(Type* t);
 
+// Is 'expr' a SymExpr for the outerVar of some ShadowVarSymbol?
+bool isOuterVarOfShadowVar(Expr* expr);
+
+// Verification support.
+void verifyOnFSIntentVarList(DefExpr* def);
```
mppf: No idea what FSIntentVarList is supposed to mean to me. I'd expect ForallStmt / ShadowVarSymbol to have methods to do any appropriate verification step.

compiler/include/symbol.h
```diff
@@ -361,6 +379,45 @@ class ArgSymbol : public LcnSymbol {
...
+  // See also: sv->outerVarSE() and outerVarSym(sv).
+  Expr* outerVarRep;
+
+  // For a reduce intent, the reduce expression.
+  // For a task-private variable, the initialization expression.
+  // Either way, wrapped in a block.  Otherwise it is NULL.
+  BlockStmt* specBlock;
+
+  Symbol* reduceGVar;
+  bool pruneit;
```
mppf: I think it's debatable whether or not we should have any public fields in a class like this. But pruneit seems "obviously" private to me. Is that impossible to do in an incremental way in this implementation? Would it be easier to remove pruneit entirely, say?

compiler/include/symbol.h
```diff
@@ -361,6 +379,45 @@ class ArgSymbol : public LcnSymbol {
...
+
+  virtual void    replaceChild(BaseAST* oldAst, BaseAST* newAst);
+  virtual bool    isConstant()                              const;
+  virtual bool    isConstValWillNotChange()                 const;
+
+  const char* intentDescrString() const;
+  // The corresponding outer var or NULL if not applicable.
+  SymExpr* outerVarSE() const { return (SymExpr*)outerVarRep; }
+  bool     isReduce()   const { return intent == TFI_REDUCE;  }
+  Expr*    spec()       const;
```
mppf: Need a comment saying what this spec function does. It's not immediately obvious to me.

compiler/include/symbol.h
```diff
@@ -361,6 +379,45 @@ class ArgSymbol : public LcnSymbol {
...
+  void     removeSupportingReferences();
```
mppf: Is this method basically implementing some part of lowering? Or ... ? At least we need a comment about it.

compiler/include/ForallStmt.h
```diff
@@ -30,12 +30,12 @@
 class ForallIntent : public Expr
```
mppf: Why is there still a class ForallIntent? I thought you were replacing it with ShadowVarSymbol?

compiler/include/expr.h
```diff
@@ -466,6 +466,10 @@ static inline bool needsCapture(FnSymbol* taskFn) {
          taskFn->hasFlag(FLAG_NON_BLOCKING);
 }
 
+static inline Symbol* outerVarSym(ShadowVarSymbol* sv) {
+  return sv->outerVarSE()->symbol();
+}
```
mppf: It's not obvious to me why this should be a function and not a method on ShadowVarSymbol. It's obvious it should be viewed as a "convenience method" though.

compiler/include/ForallStmt.h
vasslitvinov: This is correct. This class should not be used. I would like to remove it in a separate "trivial" PR. OK?

compiler/include/expr.h
vasslitvinov: Good point. I will switch it to a convenience method.

compiler/include/symbol.h
vasslitvinov: I don't like "get". Also, we have intentDescrString() and retTagDescrString(). So how about forallIntentDescrString() ? I also like forallIntentTagDescription(). Which would you prefer?

compiler/include/symbol.h
vasslitvinov: I will add this:
// The reduce intent specification.

compiler/include/symbol.h
vasslitvinov: You are right. I will add this:
  // Remove no-longer-needed references to outside symbols when lowering.

compiler/include/symbol.h
vasslitvinov: Yes, it definitely should be a SymExpr. Except in one spot in scopeResolve (that's for checkIdInsideWithClause()) I need it to be something an UnresolvedSymExpr. Meaning that without it I'd have to replace checkIdInsideWithClause() with something ugly. :(

compiler/include/symbol.h
vasslitvinov: Given that "globalOp" is a name that has been used throughout our implementation of reduce intents, historically and at present, how about reduceGVar --> reduceGlobalOp ?

compiler/include/symbol.h
vasslitvinov: While it is indeed obviously private, it is both read and written in implementForallIntents.cpp in non-method functions. So marking it as "private" will prevent that code from compiling.

My hopes are to eliminate "pruning" altogether so that this field can be removed. OK to leave it as-is for now?

compiler/include/symbol.h
vasslitvinov: Never mind. I will get rid of this function.

compiler/include/ForallStmt.h
mppf: If it's trivial enough, why not remove it now? I don't see any advantage to putting it off, other than confusing me :) That is to say, I'd like you to remove it now if it's trivial. If it's not trivial, then you'll learn something, right?

If you want to do this in multiple PRs, the thing to split out would be adding the ShadowVarSymbol - just add it as a class without using it anywhere...

compiler/include/symbol.h
mppf: I'm still confused about this. What is a reduce intent specification? Why is it an Expr? Can you give an example?

compiler/include/symbol.h
mppf: Yes but let's at least put a comment about how we plan to get rid of it.

compiler/include/symbol.h
mppf: I don't have a problem having a field with the name reduceGlobalOp, but it would seem to me that such a field would store something different from a field named reduceGlobalVar. (The reduceGlobalOp would store a reduction class instance or "operator"; but a reduceGlobalVar might store the result of the reduction, say).

compiler/include/symbol.h
mppf: Right, so in that event it seems to me the comment should say:
```
// Either a SymExpr* or an UnresolvedSymExpr*
// After scope resolve completes, there should be no more UnresolvedSymExprs
```

A reasonable alternative strategy would be to store the UnresolvedSymExpr in a separate field from the SymExpr.

compiler/include/symbol.h
mppf: forallIntentDescrString() is reasonable (since other code looks like that) but I'd slightly prefer forallIntentTagDescription().

compiler/include/symbol.h
vasslitvinov: Since it is indeed a reduction class instance, not the reduction state or final result, I will go with reduceGlobalOp.

compiler/include/symbol.h
vasslitvinov: Given this code:
```chpl
var x: MyReduceResultType;
forall ... with (MyReduceOp(int) reduce x) { ... }
```
spec() is the CallExpr for `MyReduceOp(int)`.

This btw means that I want to reduce using MyReduceOp class and the type of the input data is `int`.

mppf: So maybe we should call the method something like `reduceOpExpr` ?

mppf: `spec` just doesn't communicate it to me...

vasslitvinov: reduceOpExpr() it is.

mppf:
Note that this PR continues on the path defined in issue #6514 for how we want to represent foralls.

compiler/passes/resolveIntents.cpp
```diff
@@ -263,6 +263,17 @@ void resolveArgIntent(ArgSymbol* arg) {
...
 
+static void resolveVarIntent(VarSymbol* sym) {
+  QualifiedType q = sym->qualType();
+  if (q.getQual() == QUAL_UNKNOWN) {
+    if (sym->isRef()) {
+      sym->qual = QUAL_REF;
+    } else {
+      sym->qual = QUAL_VAL;
+    }
+  }
```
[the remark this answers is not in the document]
benharsh: I think it would be worth attempting, but maybe in a future PR.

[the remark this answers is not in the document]
mppf: OK, in that case, Vass, could you add a TODO comment about it? Thanks.

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1719,10 +1720,70 @@ void implementForallIntents2wrapper(CallExpr* call, CallExpr* eflopiHelper)
...
+    case TFI_CONST_REF: return INTENT_CONST_REF;
+    case TFI_REDUCE:
+      INT_ASSERT(false);    // don't know what to return
+      return INTENT_BLANK;  // dummy
+  }
+  INT_ASSERT(false);    // unexpected ForallIntentTag; 'tfi' contains garbage?
+  return INTENT_BLANK;  // dummy
+}
+
+static ForallIntentTag forallIntentForArgIntent(IntentTag intent) {
```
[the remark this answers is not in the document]
vasslitvinov: They are different. They handle a different set of intents and have different behavior outside of that set.

compiler/passes/normalize.cpp
```diff
@@ -1165,21 +1182,17 @@ static void callConstructor(CallExpr* call) {
 // That would be incorrect because this is a special syntax for reduce intent.
 //
 static SymExpr* callUsedInRiSpec(Expr* call, CallExpr* parent) {
-  SymExpr* retval = NULL;
-
-  if (parent != NULL && parent->isPrimitive(PRIM_MOVE) == true) {
+  if (parent && parent->isPrimitive(PRIM_MOVE)) {
```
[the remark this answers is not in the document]
vasslitvinov: I am simply restoring the code to what I originally wrote.

compiler/passes/resolveIntents.cpp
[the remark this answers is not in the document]
vasslitvinov: Done. Good point.

compiler/passes/normalize.cpp
```diff
@@ -1270,12 +1283,12 @@ static void  evaluateAutoDestroy(CallExpr* call, VarSymbol* tmp);
 static bool  moveMakesTypeAlias(CallExpr* call);
 static Type* typeForNewNonGenericRecord(CallExpr* call);
 
-static void insertCallTemps(CallExpr* call) {
-  if (shouldInsertCallTemps(call) == true) {
+static void insertCallTemps(CallExpr* call, Expr* stmt) {
+  if (shouldInsertCallTemps(call) || stmt != NULL) {
```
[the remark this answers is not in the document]
vasslitvinov: Done. insertCallTempsWithStmt(call, stmt).

compiler/AST/foralls.cpp
```diff
@@ -186,7 +186,7 @@ void ForallIntents::acceptFI(AstVisitor* visitor) {
 // These functions report a user error for an unexpected intent.
 //
 
-static TFITag it2tfi(Expr* ref, IntentTag intent) {
+static ForallIntentTag it2tfi(Expr* ref, IntentTag intent) {
```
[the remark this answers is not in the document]
vasslitvinov: Done.

compiler/AST/AstDumpToNode.cpp
```diff
@@ -396,34 +396,6 @@ bool AstDumpToNode::enterBlockStmt(BlockStmt* node)
...
-    fputs("reduceExpr: ", mFP);
-    node->reduceExpr()->accept(this);
-  }
-
-  mOffset = mOffset - 2;
-  if (!compact) newline();
-  exitNode(node);
-
-  return false;
-}
```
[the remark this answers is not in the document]
vasslitvinov: Done, too. Good point.

Issue #5569, opened 2017-03-08
KNL locale model post-commit TODOs


dmk42:
This issue tracks any items that need to be addressed after the KNL locale model PR #5555  is committed.

More sophisticated method of choosing memory kinds:
It would be nice to be able to specify more than one property.  What if we had a function
  locale.selectCharacteristics(args...)
where each argument is a set of available sublocales?  The sets could be created by functions like
  memoryPerSublocale(s: int)
  highBandwidth()
  lowLatency()
and they return a set of the memory kinds that match that characteristic.  The selectCharacteristics function intersects the sets in order, excluding any that cause the resulting set to become empty.

Non-KNL issues uncovered by the PR, to be addressed later:
* In modules/internal/defaultRectangular.chpl, dataParIgnoreRunningTasks is set to false for the flat locale model, which could cause heavy oversubscription.  More thought should be given to this.  See [720c769](https://github.com/chapel-lang/chapel/commit/720c769658358e406745ed8dc11160dd9e61e023) for more info.
* Overall, as more locale models are added, we will need to take a closer look at the prediff and skipif files to make sure we apply tests to the right locale models.  Some tests have been converted from ==numa to !=flat, and that may need to be revisited if locale models with completely different characteristics are added.

dmk42:
I'm deleting the following from the list, which has now been done:

---

The following files need additional documentation on the KNL locale model.

* doc/technotes/localeModels.rst
* doc/platforms/knl.rst

ronawho:
KNL locale model was removed in https://github.com/chapel-lang/chapel/pull/15832

Issue #5323, opened 2017-02-08
Proposed language changes for GPU programming features


daniellowell:
**AMD Research is proposing changes to the Chapel language to enable support for the GPU programming model.**

This Issues topic has been created to open up discussion with Chapel Language developers and users on what these proposed changes should look like and what features might be need, or are superfluous.

The changes are described in the [Chapel Improvement Proposal (CHIP) 17](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/17.rst)

 This will give GPU programmers the ability to:
    1. allocate and access GPU local scratch pad memory
    2. allow access to GPU primitives such as get_local_id()
    3. enforce proper execution by use of workgroup scope synchronization
    4. specify the size of workgroups.
    5. specify the number of workitems in a kernel launch
    6. specify the dimensions of the global workitems and workgroups

The goal is to provide Chapel programmers the tools to create diverse and more efficient programs on a GPU. This CHIP however does not cover data movement between a GPU locale and other locales and assumes all required data and logic is available to the GPU at runtime.

Please feel free comment.

Current collaborators are:

Michael Ferguson, Cray
Daniel Lowell, AMD
Mike Chu, AMD
Ashwin Aji, AMD
Michael Ferguson, Cray

#5319

mppf:
I'm particularly interested in the question of how a `forall` loop might possibly be combined with the sense of a `workgroup`. In particular, I have a hypothesis that the concept of a `workgroup` is useful even for parallel programs on CPUs. Certainly it is critical for GPU programming, but if it's also useful for CPU programming, we might want to consider ways to modify the Chapel language that make something like `workgroup` an optional component of all `forall` loops.

One of the features I've wished for with Chapel's `forall` loops is a way to provide a manually vectorized kernel (written in assembly, possibly) that should execute inside a `forall` loop. Including a sense of a `workgroup` that maps down to the vector width on a CPU is one way to do that.

Let's consider a simple STREAM-like example to understand the issue and the idea.

``` chapel
config const n: int;
var D = {1..n}; // possibly dmapped
var A:[D] int;
var B:[D] int;
var C:[D] int;

forall (a,b,c) in zip(A,B,C) {
  a = b + c;
}
```

Now suppose I have manually created an 8-wide vector addition routine, `add8`, in C or assembly that is available to Chapel as an `extern proc` taking in addresses for the first of 8 elements. Then I might write the loop as:

``` chapel
// for now, ignoring left-over elements
[...]

PR #7152, opened 2017-08-29
APU locale model

mlchu:
This defines a locale model for an AMD Accelerated Processing Unit (APU).

APUs consist of a CPU and an integrated GPU. This PR is the first part of a larger set of changes to allow targeting of GPUs from Chapel code. The hierarchical sublocale functionality is used so users can write code such as:

```
on Locale[0].GPU {
  // GPU code
}
```

This change will eventually require the HSA runtime to also be installed on the system being run. For now, the calls to the HSA runtime have been commented out, until the required runtime changes are checked in later. Additionally, we are currently hardcoding a single CPU and GPU sublocale. This could also change after more runtime components are added. The compiler changes to generate GPU executable code will be another PR added in the future.

### DIFF
--- a/modules/internal/LocaleModelHelpAPU.chpl
+++ b/modules/internal/LocaleModelHelpAPU.chpl
@@ -1,5 +1,7 @@
 /*
  * Copyright 2017 Advanced Micro Devices, Inc.
+ * Copyright 2004-2017 Cray Inc.
+ * Other additional copyright holders may be indicated within.
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/internal/LocaleModelHelpAPU.chpl
+++ b/modules/internal/LocaleModelHelpAPU.chpl
@@ -23,6 +23,9 @@ module LocaleModelHelpAPU {
   use LocaleModelHelpSetup;
   use LocaleModelHelpRuntime;
 
+  pragma "no doc"
+  config param debugAPULocale = false;
+
   //////////////////////////////////////////
   //
   // utilities
@@ -84,14 +87,19 @@ module LocaleModelHelpAPU {
                      ) {
     const dnode =  chpl_nodeFromLocaleID(loc);
     const dsubloc =  chpl_sublocFromLocaleID(loc);
-    if (dsubloc == 0) {
-        warning("executing on CPU");
-    }
-    else if (dsubloc == 1) {
-        warning("executing on GPU");
+    
+    if (debugAPULocale) {
+      if (dsubloc == 0) {
+          chpl_debug_writeln("** executing on CPU");
+      }
+      else if (dsubloc == 1) {
+          chpl_debug_writeln("** executing on GPU");
+      }
     }
 
    if dnode != chpl_nodeID {
+      var tls = chpl_task_getChapelData();
+      chpl_task_data_setup(chpl_comm_on_bundle_task_bundle(args), tls);
       chpl_comm_execute_on(dnode, dsubloc, fn, args, args_size);
     } else {
       // run directly on this node
@@ -121,6 +129,8 @@ module LocaleModelHelpAPU {
     const dnode =  chpl_nodeFromLocaleID(loc);
     const dsubloc =  chpl_sublocFromLocaleID(loc);
     if dnode != chpl_nodeID {
+      var tls = chpl_task_getChapelData();
+      chpl_task_data_setup(chpl_comm_on_bundle_task_bundle(args), tls);
       chpl_comm_execute_on_fast(dnode, dsubloc, fn, args, args_size);
     } else {
       var origSubloc = chpl_task_getRequestedSubloc();
@@ -151,16 +161,22 @@ module LocaleModelHelpAPU {
     //
     const dnode =  chpl_nodeFromLocaleID(loc);
     const dsubloc =  chpl_sublocFromLocaleID(loc);
+    var tls = chpl_task_getChapelData();
+    var isSerial = chpl_task_data_getSerial(tls);
     if dnode == chpl_nodeID {
-      if __primitive("task_get_serial") then
+      if isSerial {
         chpl_ftable_call(fn, args);
-      else
+      } else {
+        chpl_task_data_setup(chpl_comm_on_bundle_task_bundle(args), tls);
         chpl_comm_taskCallFTable(fn, args, args_size, dsubloc);
+      }
     } else {
-      if __primitive("task_get_serial") then
+      chpl_task_data_setup(chpl_comm_on_bundle_task_bundle(args), tls);
+      if isSerial {
         chpl_comm_execute_on(dnode, dsubloc, fn, args, args_size);
-      else
+      } else {
         chpl_comm_execute_on_nb(dnode, dsubloc, fn, args, args_size);
+      }
     }
   }
 }
--- a/modules/internal/LocaleModelHelpAPU.chpl
+++ b/modules/internal/LocaleModelHelpAPU.chpl
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2017 Advanced Micro Devices, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


mlchu:
@dmk42 @mppf could you code review?

One item I'm not sure about: I noticed the functions chpl_localeModel_sublocToExecutionSubloc and chpl_localeModel_sublocMerge were added in the last few months to the other locale models. I'm unsure of the intended purpose for these functions so I used the same implementation as the NUMA locale.

mlchu:
Looks like it failed a check because of copyrights. Are Cray copyrights supposed to be added to new files? I left the copyrights on the modified files.

tmacd9:
If it goes into the Chapel repository then it needs to have a Cray, Inc. copyright.  There can also be an AMD copyright.

mlchu:
@tmacd9 Ok. I assume that rule excludes things in the third-party directory? I was just following the directions I got from our legal department. I'm fine with adding a Cray copyright to these files since we largely based them off Cray code anyway -- I'll make that revision soon. However, I do have some runtime files I'd like merged in the future which are entirely new AMD/HSA code. It definitely belongs in the runtime directory rather than third-party. I'll have to circle back with our legal team for their guidance.

dmk42:
@mlchu - Thanks very much for this PR!  While I look at it, I have an answer for your earlier question.

I added `chpl_localeModel_sublocToExecutionSubloc` and `chpl_localeModel_sublocMerge` because of a property of the KNL locale model, but I implemented them generically because there may come a time when another locale model has similar properties that need to be taken into account.  Here are the details.

Right now, the Qthreads tasking layer is the only one that supports sublocales.  When Chapel runs on Qthreads, a Chapel sublocale maps to a Qthreads shepherd thread.  There is one shepherd thread for each NUMA node that has cores in it, and each shepherd thread farms out work to the worker threads in its NUMA node.  There is one worker thread per core.

Therefore, for the purposes of executing code, a Chapel sublocale is a NUMA node that has one or more cores in it.  However, KNL has two kinds of NUMA nodes, the ones that contain cores and the ones that contain only a special kind of memory but no cores.  I found it useful to plant information in the Chapel sublocale ID that tells which kind of memory is being used.  `chpl_localeModel_sublocToExecutionSubloc` strips off the memory information, leaving only the information that tells which Qthreads shepherd is being used.  `chpl_localeModel_sublocMerge` recombines the memory information with the Qthreads shepherd ID to construct a complete Chapel sublocale ID.

So, those two functions make it possible for a Chapel sublocale to model both the NUMA node on which execution takes place, and information about whether or not some other resource is being used (special memory in the case of KNL).

dmk42:
@mlchu - Thanks for your patience.  While reviewing, I came across this commit that you will probably need to merge into `localeModelHelpAPU.chpl`.

https://github.com/chapel-lang/chapel/commit/aeb31dbf413bf5a43311618630d7c0be14fabdbe

It comes from PR #6765 .

Continuing with my review . . .

modules/internal/LocaleModelHelpAPU.chpl
```diff
@@ -0,0 +1,166 @@
...
+  export
+  proc chpl_executeOn(loc: chpl_localeID_t, // target locale
+                      fn: int,              // on-body function idx
+                      args: chpl_comm_on_bundle_p,     // function args
+                      args_size: size_t     // args size
+                     ) {
+    const dnode =  chpl_nodeFromLocaleID(loc);
+    const dsubloc =  chpl_sublocFromLocaleID(loc);
+    if (dsubloc == 0) {
+        warning("executing on CPU");
```
mppf: I'd expect such printouts to be controlled by a config param (maybe they are and I'm missing something?)

mppf: Also, I don't see so much different here from LocaleModelHelpNuma (but I might be missing something). Makes me wonder if the APU model could use LocaleModelHelpNuma. That would cause it to use more regularly tested code. I'm not too worried about it either way, though - philosophically the APU locale model needs to specify these functions and could do so by implementing them from scratch. I'd just like to avoid copy/paste code we have to maintain if we can.

modules/internal/LocaleModelHelpSetup.chpl
```diff
@@ -89,6 +90,20 @@ module LocaleModelHelpSetup {
...
+
+    forall locIdx in dst.chpl_initOnLocales() with (ref root_accum) {
+      chpl_task_setSubloc(c_sublocid_any);
+      const node = new LocaleModel(dst);
+      dst.myLocales[locIdx] = node;
+      root_accum.accum(node);
+    }
+
+    root_accum.setRootLocaleValues(dst);
+    here.runningTaskCntSet(0);  // locale init parallelism mis-sets this
```
mppf: This is identical to helpSetupRootLocaleNUMA. Is there some philosophical reason for this copy/paste, or can we update callers to just call helpSetupRootLocaleNUMA ?

modules/internal/localeModels/apu/LocaleModel.chpl
```diff
@@ -0,0 +1,295 @@
...
+      // but here it is defined thus for backward compatibility.
+      f <~> new ioLiteral("LOCALE") <~> _node_id;
+    }
+
+    proc getChildSpace() {
+      halt("error!");
+      return {0..#numSublocales};
+    }
+
+    proc getChildCount() return 0;
```
mppf: It's clear that these getChild functions are inconsistent. @dmk42 - we had some plan to allow for querying sublocales by type. Do you have written down somewhere what that interface will look like? We should at least start a github issue about it.

Anyway, right now code like the range/default rectangular array parallel iterators will use getChildCount() to compute the number of sublocales (https://github.com/chapel-lang/chapel/blob/master/modules/internal/ChapelRange.chpl#L1713). We don't want it to by default share work between the CPU and the GPU (and we want it to only run on the GPU when requested), so I'd recommend treating the GPU similarly to the memory types in the KNL locale model. That is, I don't think the GPU sublocale should be a "child" of this locale... (at least not until we fix up the interface).

Clearly we have some interface work to do. I don't think this issue has to block merging the APU locale model, but I'd like too see it addressed. At the very least, it needs to be covered in an issue.

@dmk42 - what do you think about this?

mppf:
I'm happy with this whevenever @mlchu and @dmk42 think it's ready to merge. Like I said, I think we at least need an issue about the sub-locale querying interface, and I think it's on @dmk42 to create that.

Thanks!

dmk42:
The plan for a more sophisticated sublocale interface is discussed in issue #5569 .

I agree that APU support could use a similar mechanism to the KNL locale model, and that would be a good idea because I don't think we would want `c_sublocid_any` to mean it's random whether the CPU or GPU is being used.  I'll suggest the following way forward.

If I understand correctly, the APU locale model currently works in preliminary testing, right?  How about if we first make the minor cleanups suggested here and merge what works, and then in a separate PR, migrate toward a sublocale scheme more similar to what the KNL locale model does?

If we do that, then `LocaleModelHelpAPU.chpl` will probably need to remain as a separate file, because it will eventually have more differences from NUMA.

modules/internal/localeModels/apu/LocaleModel.chpl
```diff
@@ -0,0 +1,295 @@
...
+    //------------------------------------------------------------------------{
+    //- Implementation (private)
+    //-
+    proc setup() {
+      _node_id = chpl_nodeID: int;
+
+      helpSetupLocaleAPU(this, local_name, numSublocales);
+    }
+    //------------------------------------------------------------------------}
+ }
```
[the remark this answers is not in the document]
mlchu: Oh, great, I'm glad this got added. I think this was actually a change that we requested because we needed a place to clean up the HSA runtime. I'll probably add a commented out cleanup routine in the deinit()

modules/internal/LocaleModelHelpAPU.chpl
mlchu: I actually meant to remove those printouts. I'll consider putting them under some config param though. Your suggestion of making it just use LocaleModelHelpNuma is an interesting one (for this and the function in LocaleModelHelpSetup). I'll have to think about this a little bit more.

dmk42:
@mlchu - Hi.  I know you have your own schedules to deal with.  Just wanted to let you know that we have a Chapel feature freeze for the 1.16 release coming up next Monday, September 18.  Is there any chance we could get this PR ready to merge by then?  Is there anything I can do to help?

mlchu:
@dmk42 Yeah, Michael told me about the code freeze last week. I think I have the changes done, I'll get them pushed as soon as I get a break in my meetings later this afternoon. Thanks!

dmk42:
Looks great, thanks!  I'll go ahead and merge this.

buddha314:
Is there a ticket for general GPU support?

dmk42:
Not as such, but it's definitely on our radar.  If there's something specific you are interested in, I would encourage you to create an issue for it.

One example of a specific item for GPU support is https://github.com/chapel-lang/chapel/issues/5323 .

PR #7154, opened 2017-08-30
Dynamically determine location of FileCheck

dmk42:
The location of the FileCheck program used in LLVM tests was hardcoded in `test/llvm/PREDIFF` for `linux64-gnu` systems.  This change dynamically determines the platform and compiler so the tests can be run on other systems.

Tested on `linux64-gnu` and `darwin-clang`.

### DIFF
--- a/test/llvm/PREDIFF
+++ b/test/llvm/PREDIFF
@@ -4,7 +4,9 @@ OUTFILE=$2
 
 TMPFILE="$outfile.prediff.tmp"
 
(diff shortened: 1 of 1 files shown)


awallace-cray:
+1, so long as it's understood that there's no FileCheck if LLVM wasn't built.

dmk42:
Thanks, Tony.  Right, all the subdirectories that use this PREDIFF have a SKIPIF that skips the tests if we don't have LLVM.

Issue #4998, opened 2016-12-09
Check annotations file as part of the smoke test


ronawho:
Check the [ANNOTATIONS.yaml](https://github.com/chapel-lang/chapel/blob/master/test/ANNOTATIONS.yaml) file as part of the smoke test.

We often get the syntax wrong as seen in https://github.com/chapel-lang/chapel/pull/4828, https://github.com/chapel-lang/chapel/pull/3911, https://github.com/chapel-lang/chapel/pull/4997, and many others. 

Should be relatively simple to do: I think we just need to add a list of valid configs to [annotate.py](https://github.com/chapel-lang/chapel/blob/master/util/test/annotate.py) and then have a self check mode that reads through all the annotations and validates that they're formatted corrected. 

Bonus points for going through the PR numbers and making sure that we annotated the dates correctly.

Note that this would require that the smoke test has PyYAML available, so to start we could just make this a developer only tool that perf team members would run.

lydia-duncan:
Would a reasonable alternative be having Travis run "start_test -generate-graphs"?

ronawho:
I don't think so. The problem with that (and why we often don't catch this issue before merging) is because gengraphs (and thus start_test) will only pull annotations for days that it has data and will leave the other annotations unparsed. 

We want something that force parses all annotations and it'd be nice to check that configs listed are valid by storing a list of all our valid configs (chapcs, Single node XC, 16 node XC, etc)

ronawho:
We should also warn when there's no .graph file for a given annotation to avoid "losing" annotations from .graph splits or similar changes. For example we "lost" lcals annotations like https://github.com/chapel-lang/chapel/commit/8f730c2c when https://github.com/chapel-lang/chapel/commit/16614a7a2 went in

ronawho:
We also lost annotations when we switched xc-* to ml-* (#5191). Fixed in #5525 along with a typo that would have caught with a "warn when there's no.graph file" check

ronawho:
Added a basic checker with #5538. It should catch syntax errors, invalid 'config' names ('1 node XC' instead of 'Single node XC'), and using the commit date as the annotation date. It also warns when there's no .graph for a given annotation. 

The script isn't super pretty, but I'm pretty happy with it overall. I think we should try it out manually for the next few weeks, and if all goes well add it to the smoke test.

ronawho:
Running as part of travis testing with #6917

ben-albrecht:
[...]

PR #7158, opened 2017-08-30
Bump git depth for annotation checker

ronawho:
The annotation checker uses the current git repo to check that PR numbers match
up with commit dates. By default, travis limits the git depth to 50, so we were
only checking "recent" commits. This bumps the default to 100,000 (which should
effectively be a full clone of the currrent branch.)

This only adds ~45 seconds for the annotation checker config, so it should
still finish well before the `make check` configs.


ben-albrecht:
Does this have a significant impact on Travis testing time?

ronawho:
>Does this have a significant impact on Travis testing time?

No, not really -- https://github.com/chapel-lang/chapel/issues/4998#issuecomment-326094313

PR #7163, opened 2017-08-30
Remove _arrAlias field

benharsh:
The _arrAlias field was created when we still implemented slices in
closed-form. With the advent of ArrayViews, this is no longer necessary
and _arrAlias was basically only used to determine if we needed to
decrement the reference counts for array elements that are arrays.

This commit replaces the _arrAlias field with a bool field named
'_decEltRefCounts'. This field is set when the utility function
chpl_incRefCountsForDomainsInArrayEltTypes is called. The _delete_arr
function in ChapelDistribution will now check this field to determine if
we need to call chpl_decRefCountsForDomainsInArrayEltTypes.

Consequently, the 'isslice' bool formal of the 'dsiDestroyArr' method is
no longer necessary.

Testing:
- [x] full local
- [x] full gasnet

### DIFF
--- a/modules/dists/BlockDist.chpl
+++ b/modules/dists/BlockDist.chpl
@@ -917,7 +917,7 @@ proc BlockArr.setup() {
   if doRADOpt && disableBlockLazyRAD then setupRADOpt();
 }
 
-proc BlockArr.dsiDestroyArr(isslice:bool) {
+proc BlockArr.dsiDestroyArr() {
   coforall localeIdx in dom.dist.targetLocDom {
     on locArr(localeIdx) {
       delete locArr(localeIdx);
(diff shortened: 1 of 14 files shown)


benharsh:
@bradcray : how does this look?

modules/dists/BlockDist.chpl
```diff
@@ -1288,10 +1288,6 @@ proc BlockArr.doiCanBulkTransfer(viewDom) {
     for param i in 1..rank do
       if viewDom.dim(i).stride != 1 then return false;
 
-  // See above note regarding aliased arrays
```
mppf: Is there a note above to remove as well?

benharsh: There is a config const that could be removed called `disableAliasedBulkTransfer`. It's only used in a couple of compopts and in miniMD's release Makefile. Removing it with this PR would not change the behavior of those programs.

mppf: Seems like we should remove it soon though, right?

benharsh: I believe we should, yes. Any objections?

mppf: I'm happy with removing it, if it is no longer used after this PR. But you could leave that to another PR if you like.

benharsh: I'll leave it for another PR for now, and give @bradcray a chance to chime in.

PR #7165, opened 2017-08-30
 Rename the first argument to the _new function generated for initializers

lydia-duncan:
Having it be named "t" was causing conflicts when we had a field named "t" and
tried to set the corresponding argument "t" via explicit naming (a NamedExpr
in the AST).  So I renamed the argument to "chpl_t".

Adds two tests which failed prior to this change.  One exercises the issue with
default initializers, the other exercises the issue if the user happens to write an
initializer with "t" as the name of the type argument it takes.

Passed classes/initializers with futures, a full standard paratest, and allowed 32
more tests to pass in a full paratest with --force-initializers


lydia-duncan:
Mike says I should rename it to "chpl_t".  Going to do that and test

noakesmichael:
I suspect that we’re going to need to do something better than this at some
point but I sense that even in the shorter term “chpl_t” is less likely to bite us
than “_t”.  But I am open to something like _buzzle_humperdkink if “chpl_t” is
too bland even in the short term.


One possible step in the longer term is to be able to flag ArgSymbols in
a manner that is comparable to “compiler generated” and then disregard
these when doing the “named-expr” matching stuff.


Mike



On Aug 30, 2017, at 4:44 PM, Lydia Duncan <notifications@github.com<mailto:notifications@github.com>> wrote:


Mike says I should rename it to "chpl_t". Going to do that and test

lydia-duncan:
That's an interesting idea!  Maybe FLAG_LAST_RESORT would be a good one to use?  I think there is some reliance on matching certain inserted arguments (I know we do it a little with the "this" arg and I suspect constructors do for "meme", though both of these also have a corresponding flag to double check when looking for that particular argument).

lydia-duncan:
Should I open an issue on this?

PR #7166, opened 2017-08-30
update Spawn module to use error handling

psahabu:
Several procedures in Spawn would halt if they encountered an error, but this PR enables users to handle a `SystemError` if they prefer. No tests were handling `syserr` from Spawn.

- [x] rebase onto master after #7218 is merged
- [ ] passes linux64+flat testing


modules/standard/Spawn.chpl
```diff
@@ -276,10 +276,10 @@ module Spawn {
        Causes a fatal error if the subprocess does not have a
        stdin pipe open.
      */
-    proc stdin {
-      _halt_on_launch_error();
+    proc stdin throws {
+      try _throw_on_launch_error();
       if stdin_pipe == false {
-        halt("subprocess was not configured with a stdin pipe");
+        throw new Error("subprocess was not configured with a stdin pipe");
```
mppf: Seems like we shouldn't be throwing a bare Error in a standard module. I know you're working on the Error subclasses in parallel. This one could be a SystemError with EINVAL. This comment applies to the other "not configure with" errors.

psahabu: I'll do this one after I merge the `SystemError` hierarchy, so I can use `SystemError.fromSyserr(EINVAL)`.

mppf: I hope you make a SystemError.fromSyserr(EINVAL, "some particular error message") work

psahabu:
[reviewed by @mppf]

Several procedures in Spawn would halt if they encountered an error, but this PR enables users to handle a `SystemError` if they prefer. No tests were handling `syserr` from Spawn.

PR #7169, opened 2017-08-31
Broadcast registration updates directly instead of using executeOn

gbtitus:
As each node dynamically registers and deregisters memory with the NIC,
it needs to broadcast those changes to all other nodes.  We've been
doing that by looping serially over the node indices and performing an
executeOn to each target node to update the source node's entry in the
targets' registered memory maps.  Here, replace this with direct PUTs of
new values into the remote memory maps.  And to improve performance, do
the PUTs in chunks of up to 64 at a time, using the chained transaction
feature introduced recently for GNI PostFMA.  This should improve the
scalability of registration broadcast, though the effect will probably
only be noticeable at fairly large scale.


gbtitus:
@ronawho, care to take a look at this one?

PR #7174, opened 2017-08-31
update FileSystem module to use error handling

psahabu:
Several procedures in `FileSystem` would halt if they encountered an error, but this PR enables users to handle a SystemError if they prefer. No tests were handling syserr from `FileSystem`.

### DIFF
--- a/modules/standard/FileSystem.chpl
+++ b/modules/standard/FileSystem.chpl
@@ -172,10 +172,10 @@ proc locale.chdir(out error: syserr, name: string) {
    :arg name: The intended current working directory
    :type name: `string`
 */
-proc locale.chdir(name: string) {
+proc locale.chdir(name: string) throws {
   var err: syserr = ENOERR;
   chdir(err, name);
-  if err != ENOERR then ioerror(err, "in chdir", name);
+  if err != ENOERR then try ioerror(err, "in chdir", name);
 }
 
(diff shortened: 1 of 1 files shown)


modules/standard/FileSystem.chpl
```diff
@@ -172,10 +172,10 @@ proc locale.chdir(out error: syserr, name: string) {
    :arg name: The intended current working directory
    :type name: `string`
 */
-proc locale.chdir(name: string) {
+proc locale.chdir(name: string) throws {
   var err: syserr = ENOERR;
   chdir(err, name);
-  if err != ENOERR then ioerror(err, "in chdir", name);
+  if err != ENOERR then try ioerror(err, "in chdir", name);
```
lydia-duncan: I'm a little surprised this is a `try` instead of a direct `throw`, though I'm assuming ioerror is a function that throws?

lydia-duncan: I say that because the intention of those statements is that if we enter the `then` block, we are in an error situation and so it would be problematic not to `throw` if that's the behavior we should perform

psahabu: Right, `ioerror()` is a function that `throws` if an `err` is passed to it. It was easier to change the code in `ioerror` once rather than replace all calls with a `throw`, at least for now.

PR #7176, opened 2017-08-31
update HDFS module to use error handling

psahabu:
Several procedures in `HDFS` would halt if they encountered an error, but this PR enables users to handle a `SystemError` if they prefer. No tests were handling a syserr from `HDFS`.


modules/packages/HDFS.chpl
```diff
@@ -180,26 +180,30 @@ proc hdfsChapelConnect(out error: syserr, path: c_string, port: int): c_void_ptr
...
-proc hdfsChapelConnect(path: string, port: int): hdfsChapelFileSystem {
+proc hdfsChapelConnect(path: string, port: int): hdfsChapelFileSystem throws {
   var ret: hdfsChapelFileSystem;
-  forall loc in Locales with (ref ret) {
-    on loc {
-      var err: syserr;
-      const tmpstr = path.localize();
-      rcLocal(ret._internal_file) = hdfsChapelConnect(err, tmpstr.c_str(), port);
-      if err then ioerror(err, "Unable to connect to HDFS", tmpstr);
+  try {
```
mppf: Why did we need this outer try block?

psahabu: My understanding is that because the code inside the `forall` might throw, the `forall` needs to be wrapped in a `try` to pass strict mode.

mppf: I believe this is a bug with foralls. Could you create a simple reproducer / future test showing the problem in strict mode?

psahabu: Yeah, here it is: https://github.com/chapel-lang/chapel/compare/master...psahabu:forall-bug

Should I wait until we solve this future to merge this PR, or should I merge this now and include a note to fix it in the future test?

mppf: Hmm... I think it's debatable whether or not that is actually a bug. In strict mode, we're supposed to be marking with 'try' anywhere that control flow can escape, and indeed, at the end of the forall, control flow could leave the functions. I think you should merge the future and this PR.

psahabu: Will do, I'll add a note to the future saying it might not actually be a bug.

PR #7177, opened 2017-09-01
update Buffers and Path modules to use error handling

psahabu:
Several procedures in `Buffers` and `Path` would halt if they encountered an error, but this PR enables users to handle a `SystemError` if they prefer. No tests were handling a syserr from `Buffers` or `Path`.

### DIFF
--- a/modules/standard/Buffers.chpl
+++ b/modules/standard/Buffers.chpl
@@ -146,10 +146,10 @@ module Buffers {
     this.home = here;
   }
   pragma "no doc"
-  proc bytes.bytes(len:int(64)) {
+  proc bytes.bytes(len:int(64)) throws {
     var error:syserr = ENOERR;
     error = qbytes_create_calloc(this._bytes_internal, len);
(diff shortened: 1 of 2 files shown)


modules/standard/Buffers.chpl
```diff
@@ -361,19 +361,19 @@ module Buffers {
...
-    if error then ioerror(error, "in buffer.flatten");
+    if error then try ioerror(error, "in buffer.flatten");
     return ret;
   }
 
   // TODO -- shouldn't have to write this this way!
   pragma "init copy fn"
   pragma "no doc"
-  proc chpl__initCopy(x: buffer) {
+  proc chpl__initCopy(x: buffer) throws {
```
mppf: Take care, this is the first throwing initCopy that I know of. It would be OK with me to make this one try!.

psahabu: `try!` seems prudent here, thanks.

PR #7180, opened 2017-09-01
Support arrays as args to array.{push_front,push_back,insert}

ben-albrecht:
## New array method overloads

This PR adds 3 new separately documented overloads for array methods:

```chpl
proc array.push_back(arr) where isArray(arr) { .. }
proc array.push_front(arr) where isArray(arr) { .. }
proc array.insert(pos, arr) where isArray(arr) { .. }
```

with the behavior of:

```chpl
[0].push_back([1,2,3])  // [0, 1, 2, 3]
[0].push_front([1,2,3]) // [1, 2, 3, 0]
[0].insert(1, [1,2,3]); // [1, 2, 3, 0]
```

Today, the above calls get turned into promoted operations that gets executed in parallel which is unhelpful and a potential "gotcha" in the array interface.

This PR turns these confusing cases into ways to *prepend*, *append*, or *bulk insert* to arrays.

See ` test/arrays/methods/vectorMethods.chpl` for the edge cases considered when developing these changes.

## Other Changes

In writing this change, I refactored the implementations of these methods in an effort to reduce the code duplication in resizing an array under the hood (see `reallocateArray()`)

In writing tests for this change, I encountered a bug in `array.equals()` where arrays-of-arrays are unsupported, so I fixed that too.

This PR depends on #7240 (merged)

## Testing

- [x] linux64 testing


daviditen:
I generally like the idea.  It would be nice to do the same for `insert`.

Instead of making it serial, you could extend the array by `vals.size` and then insert the elements in parallel.

modules/standard/Types.chpl
```diff
@@ -414,6 +414,10 @@ proc isProperSubtype(type sub, type super) param
   return false;
 
 
+/* Returns `true` if the object has an iterable `these()` method */
+proc isIterable(x) param where canResolveMethod(x, 'these') return true;
```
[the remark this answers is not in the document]
ben-albrecht: I like!

ben-albrecht:
> Instead of making it serial, you could extend the array by vals.size and then insert the elements in parallel.

This thought occurred to me since the initial PR as well... I am favor of this, though it will require a separate overload that checks for a `size` field.

ben-albrecht:
> It would be nice to do the same for `insert`.

Can do.

modules/internal/ChapelArray.chpl
```diff
@@ -2712,6 +2721,13 @@ module ChapelArray {
       }
     }
 
+    pragma "no doc"
+    /* Ensure promoted-push_front is serial */
+    proc push_front(vals) where isIterable(vals) {
+      for val in vals do
```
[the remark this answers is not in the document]
ben-albrecht: Yeah, that's tricky... If users think about this like a "promoted" `push_front`, then they might expect the current implementation (`for val in vals`). If they think about this like a way to prepend to an array, then they might expect the order to be preserved.

The latter seems more useful, while the former seems less surprising IMO.

modules/standard/Types.chpl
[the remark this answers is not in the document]
ben-albrecht: It was inheriting the `use Reflection` statement from `ChapelArray.chpl` (now moved to `Types.chpl`).

modules/internal/ChapelArray.chpl
```diff
@@ -2636,6 +2638,13 @@ module ChapelArray {
       }
     }
 
+    pragma "no doc"
```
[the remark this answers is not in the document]
ben-albrecht: Separately, or as part of the main `push_front` / `push_back` documentation?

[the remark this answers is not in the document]
daviditen: I lean toward separately since I see them as having different signatures vs. just being promoted versions of the single element ones.  Either way could work for me though.

modules/internal/ChapelArray.chpl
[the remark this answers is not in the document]
benharsh: My gut reaction is that we should preserve the order of elements (i.e. the first element yielded is the first element of the array after push_front). While this is simple for arrays, the implementation will be more difficult for iterators yielding an unknown number of elements.

Do we want to split array push_front/back to a separate method with unique behavior, and have the isIterable method as a way to avoid the need for locking in promoted expressions?

ben-albrecht:
I'd like to adjust the scope of this PR to the following:

3 new separately documented overloads for array methods:

```chpl
proc array.push_back(arr) where isArray(arr) { .. }
proc array.push_front(arr) where isArray(arr) { .. }
proc array.insert(pos, arr) where isArray(arr) { .. }
// Note: where-clause is necessary so that the typed function takes precedence 
//       for cases like arrays-of-arrays
```

with the behavior of:

```chpl
[0].push_front([1,2,3]) // [0, 1, 2, 3]
[0].push_back([1,2,3])  // [1, 2, 3, 0]
[0].insert(1, [1,2,3]); // [1, 2, 3, 0]
```

This allows users to *extend* and *prepend* to arrays in-place, and prevents users from tripping over the currently unhelpful promoted `push_*` behavior.

`isIterable()` seems generally useful, but I think it belongs in a separate PR.

@benharsh, @daviditen, @ronawho -- thoughts?

benharsh:
I like the proposed overloads specifically for arrays. I think the promotion behavior (even serially) would be a common 'gotcha' and annoying for users to have to write themselves.

daviditen:
The adjustment in scope works for me.

I don't understand what exactly this comment is saying:

```chpl
// Note: where-clause is necessary so that the typed function takes precedence 
//       for cases like arrays-of-arrays
```

Is it saying that the where clause is necessary and this won't work?

```chpl
proc array.push_back(arr: [] this.eltType) { .. }
```

ben-albrecht:
@daviditen - Good question. I didn't explain it well in that comment, because I didn't fully understand the subtle difference it until I did some further experimentation. For clarity, I am proposing the following overloads:

```chpl
// typed element overload - takes precedence over generic overload
proc array.push_back(val: this.eltType) { .. } // already exists

// generic array overload
proc array.push_back(arr) where isArray(arr) { .. }  // new
```

With the typed array overload, we'd have these overloads:

```chpl
// typed element overload
proc array.push_back(val: this.eltType) { .. } // already exists

// typed array overload
proc array.push_back(arr: [] this.eltType) { .. } // new
```

This works in almost all cases, but becomes problematic when type coercion gets involved... With the above overloads, the following still resolves to the promoted typed element overload, which we are trying to avoid:

```chpl
// array of ints pushed onto array of reals.
[0.0].push_back([1,2,3,4]);
```

ben-albrecht:
Anyone subscribed to this PR - I'd appreciate a second pass over the latest changes. linux64 testing in progress.

modules/internal/ChapelArray.chpl
```diff
@@ -2642,43 +2644,76 @@ module ChapelArray {
...
+          const nextAllocRange = resizeAllocRange(newRange, direction=direction);
           if debugArrayAsVec then
-            writeln("push_back reallocate: ",
-                    oldRng, " => ", nextAllocRange,
+            writeln("reallocateArray: ",
+                    oldRange, " => ", nextAllocRange,
                     " (", newRange, ")");
-          this._value.dsiReallocate((nextAllocRange,));
           // note: dsiReallocate sets _value.dataAllocRange = nextAllocRange
+          //this._value.dsiReallocate((nextAllocRange,));
```
benharsh: Mistakenly commented out dsiReallocate?

modules/standard/Types.chpl
```diff
@@ -414,6 +416,8 @@ proc isProperSubtype(type sub, type super) param
   return false;
 
 
+/* Returns `true` if the object has an iterable `these()` method */
+proc isIterable(x) param return canResolveMethod(x, 'these');
```
benharsh: I think ``isIterable`` should be a separate PR.

test/arrays/diten/arrayAsVecShrinkBug.good
```diff
@@ -1,21 +1,21 @@
...
-push_back reallocate: 1..28 => 1..42 (1..29)
-push_back reallocate: 1..42 => 1..63 (1..43)
-push_back reallocate: 1..63 => 1..94 (1..64)
-push_back reallocate: 1..94 => 1..141 (1..95)
-push_back reallocate: 1..141 => 1..211 (1..142)
-push_back reallocate: 1..211 => 1..316 (1..212)
-push_back reallocate: 1..316 => 1..474 (1..317)
-push_back reallocate: 1..474 => 1..711 (1..475)
-push_back reallocate: 1..711 => 1..1066 (1..712)
+reallocateArray: 1..0 => 1..1 (1..1)
```
benharsh: I think it would be good to preserve which method called reallocateArray.

test/arrays/methods/vectorMethods.chpl
```diff
@@ -0,0 +1,99 @@
...
+
+// insert
+{
+  var A: [1..0] [1..2] int;
+  const B: [1..3] [1..2] int = [[1,1],[2,2],[3,3]];
+
+  A.insert(1, B);
+
+  assert(A.equals(B));
+}
```
benharsh: What if A does a push_back/push_front/insert with itself? Or a slice of itself?

ben-albrecht: It crashes and burns. *The plot thickens.*

modules/internal/ChapelArray.chpl
```diff
@@ -2642,43 +2644,75 @@ module ChapelArray {
...
+        compilerError("push_back() is only supported on dense 1D arrays");
+
+      chpl__assertSingleArrayDomain("push_back");
+
+      var idx = this.domain.high+1;
+
+      const newRange = this.domain.low..(this.domain.high + vals.size);
+      reallocateArray(newRange);
+
+      for val in vals {
```
daviditen: This could be: `this[idx..] = vals;`

modules/internal/ChapelArray.chpl
```diff
@@ -2731,26 +2765,33 @@ module ChapelArray {
...
+      if (!chpl__isDense1DArray()) then
+        compilerError("push_front() is only supported on dense 1D arrays");
+
+      chpl__assertSingleArrayDomain("push_front");
+
+      const newRange = (this.domain.low - vals.size)..this.domain.high;
+      reallocateArray(newRange, direction=-1);
+
+      var idx = this.domain.low;
+      for val in vals {
```
daviditen: This could be `this[idx..#vals.size] = vals;`.

test/arrays/methods/vectorMethods.chpl
daviditen: I think that should be undefined, so crash and burn is fine.

ben-albrecht: @daviditen  - How strongly do you feel about this being undefined? I have this working now in my upcoming commit..

daviditen: If it's working, parallel, and doesn't use a temp array then I'm good with that.

test/arrays/diten/arrayAsVecShrinkBug.good
ben-albrecht: Actually, doing that would require undoing the refactor, since the check for whether or not an allocation occurs within `reallocateArray`.

modules/internal/ChapelArray.chpl
```diff
@@ -2861,25 +2858,35 @@ module ChapelArray {
 
        The array must be a rectangular 1-D array; its domain must be
        non-stridable and not shared with other arrays.
+
+       Aliasing arguments is not supported for this method. For example, the
```
benharsh: 'is not' -> 'are not' ?

test/arrays/methods/vectorMethods.chpl
ben-albrecht: Got `push_*` to work, but we'll leave `insert()` for a rainy day.

PR #7189, opened 2017-09-04
Improved const properties

vasslitvinov:
This PR flags more things as constants, with other accomodating changes.

#### What is different now

* In ArgSymbol::isConstant(), return true always for INTENT_CONST.
Previously it would handle this case as if it were the default intent.
I think this change by itself enabled const checking in a couple of cases
where it was not performed before. (See changes to tests.)

* In flattenFunctions, use a more comprehensive check sym->isConstant()
instead of sym->hasFlag(FLAG_CONST). This enables it to pick ArgSymbols
and (upcoming) ShadowVarSymbols, in addition to VarSymbols.

* Also in flattenFunctions / addVarsToFormals(), when adding a formal
to replace a variable in a lexically-enclosing scope, also give the formal
the flag(s) FLAG_REF_TO_CONST and/or FLAG_CONST_DUE_TO_TASK_FORALL_INTENT
when the replaced variable has those. Formerly this was done only for
FLAG_ARG_THIS.

* During remoteValueForwarding, isSufficientlyConst() returns true
also when the reference has FLAG_REF_TO_CONST.
Note that we cannot simply check that flag in canForwardValue().
That's because we should not forward the variable if it is used in
a dot-locale expression, even though it does not change.

* Minor improvement of comm counts in the test:
```
studies/ssca2/performance/SSCA2_commDiags
```

#### Supporting changes

* functionResolution: the top of the callStack can be *two* parentExprs
up from the call being considered, if the latter is in a ContextCallExpr.

* The change in implementForallIntents.cpp is (a) presentation and (b)
instead of the default intent, use the corresponding concrete intent.

#### While there

* Clarification from BenH (not reflected in the code):
FLAG_REF_TO_CONST means that the pointed-to thing does not change.

* Removed the historic comment in addVarsToFormals():
```
// BLC: TODO: This routine is part of the reason that we aren't
// consistent in representing 'ref' argument intents in the AST.
......
```

It is at least somewhat outdated, and in any case not helpful any longer.

#### More RVF is better?

There was also an unexpected challenge with RVF (remote value forwarding).
Consider our Barriers module, whose simple version is this:

    record Barrier {
      var bar = new SomeBarrierImplementation();
    }
    proc Barrier.notify() {
      on this {
        this.bar.notify();
      }
    }

A simple version of its use is:

    var b = new Barrier();
    coforall loc in Locales do
      on loc do
        b.notify();

The two on-statements make the compiler try RVF of 'b'.

(A) The compiler cannot RVF across the on-statement in the coforall.
That's because 'b' is passed into Barrier.notify(). That queries its locale
to see where its on-statement should go. If 'b' were RVF-ed, the correct
locale would be lost.

(B) For the on-statement in Barrier.notify(), however, the compiler did not RVF
before this change, it does now. This is an improvement, you say?

Turns out, this introduced additional communication.

That's because when the coforall invokes notify() on a non-0 locale,
notify() now RVFs its 'this' formal. (See (B).) However, to do so, it needs
the contents of 'this'. As the actual for 'this' is not RVF-ed (see (A)),
notify() has to read the contents from Locale 0. Before this change, this
read was not needed so this communication was not done. Now it is.

My solution is to replace, in notify(), "on this" with "on this.bar".
This way this's locale is not accessed and it can now be RVF-ed for (A).
The additional communication is no longer inferred.

Besides changing the module, I had to change the test itself
for this same reason. There, I instead removed the "on" clause
altogether, by BenH's suggestion. While there, I simplified
the test and wrote some more explanations in comments.

Perhaps a more principled solution, though, is to forward
the record's original locale in addition to the record's contents,
if the locale is accessed remotely. I did not take it on in this PR.

#### Testing

- [X] linux64 --verify
- [X] gasnet --verify


vasslitvinov:
@benharsh would you review?

test/parallel/taskPar/sungeun/barrier/commDiags.chpl
```diff
@@ -34,11 +34,11 @@ proc remoteTestSplitPhase(b: Barrier, numRemoteTasks) {
   coforall t in barSpace do on A.domain.dist.idxToLocale(t) {
     B[t] = A[t];
     b.notify();
-    if t!=hi {            // Use an 'on' to get same results for atomics
-      on b { b.wait(); }  //  and sync/single.  Without the on, there will
-    } else {              //  be additional comm for the sync/single case
-      on b { b.wait(); }  //  due to returning the value to the original
-    }                     //  locale.
+    if t!=hi {                // Use an 'on' to get same results for atomics
+      on b.bar { b.wait(); }  //  and sync/single.  Without the on, there will
```
benharsh: Could this just be ``b.wait();`` ?

PR #7193, opened 2017-09-05
Improve the accuracy of our running task counter

ronawho:
We use `here.runningTasks()` in most of our foralls to determine how much
parallelism to create. An inaccurate counter will result in either under or
oversubscribing the node, which in both cases can hurt performance. This work
fixes our task counting, which improves performance for multi-locale stream-ep
and ISx. This also adds a pretty comprehensive set of task counting tests.

The bulk of this work moves our task counting for on-stmts from the runtime to
the compiler. Previously we would try to increment the runningTaskCounter in
the runtime for "moved" tasks, but we did a poor job of actually tracking what
constituted a moved task. We would increment for locally spawned tasks, but not
for remotely spawned tasks, so code like `coforall loc in locales do on loc`
would increment for locale 0, but no other locales, which is the opposite of
what we want.

Here we remove all the logic from the runtime and move it into the compiler. We
increment the remote runningTaskCount for non-fast ons, and we decrement the
local runningTaskCount for non-fast, blocking on stmts. We don't want to
increment the remote taskCount for fast ons because they'll run in the
comm-handler and won't actually create any tasks. We don't want to decrement
the local taskCount for non-blocking ons since we're spawning new remote tasks,
not moving the current one and we don't want to decrement the local taskCount
for fast ons since they'll finish quickly anyways.

Some other improvements included in this PR:

Use the "bounded" upEndCount/waitEndCount for cobegins. We always know how many
tasks a cobegin stmt will create, so use the bounded version of upEndCount and
waitEndCount. This should have a performance improvement for cobegins (though
we don't use them much) and it allows us to modify the runningTaskCount before
spawning any tasks, so we can be sure that the runningTaskCounter is accurate
by the time any tasks have spawned. This was mostly done to make task counting
tests for cobegins deterministic, but it's a nice improvement regardless.

Fix task counting for `coforall loc in Locales do on loc` ("bounded" remote
coforalls). Previously we weren't subtracting the spawning task that just does
the waitFor() from the runningTaskCount when countRunningTasks=false. We were
already doing it for the countRunningTasks=true case though.

Fix task counting for unbounded coforalls. We need to decrement the task count
before doing chpl_taskListExecute since the main task will do that work, not a
newly spawned task.

Misc notes:
-----------

Before locales are initialized we have a dummyLocale that lives on locale 0.
We don't want to access `here` from remote locales during initialization
because that introduces communication, so this adds a replicated var
rootLocaleInitialized to track whether we should modify `here`.

Replacing the dummyLocale also results in a bogus runningTaskCounter, since
some increment/decrement pairs are split across the dummy and real locales.
This only used to be an issue on locale 0, but the on-stmt task-counting now
changes the count on all locales. We now do the reset in the runtime on all
locales prior to executing user code. We could probably move this into the
module code, but the runtime was the easiest place to do it for now since we
won't have to spawn a coforall+on to run on all nodes.

We only modify the locale/remote task count for "actual" on stmts, not direct
ons (ons to the current locale) since those are directly executed without
spawning any tasks.

Individual commits have more details and some followups/todos are captured in
https://github.com/chapel-lang/chapel/issues/7194


ronawho:
@benharsh mind reviewing? It's probably easiest to review commit at a time.

PR #7203, opened 2017-09-06
Improve isConstValWillNotChange()

vasslitvinov:
In Symbol::isConstValWillNotChange() :

* Do not return 'true' for VarSymbols corresponding to 'const ref' aliases.

* Return 'true' when the symbol has the flag FLAG_REF_TO_CONST.

Do the same in the standalone function isConstValWillNotChange(), too.

### DIFF
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -567,7 +567,8 @@ bool VarSymbol::isConstant() const {
 
 
 bool VarSymbol::isConstValWillNotChange() const {
-  return hasFlag(FLAG_CONST);
+  return (hasFlag(FLAG_CONST) && !hasFlag(FLAG_REF_VAR)) ||
+         hasFlag(FLAG_REF_TO_CONST);
 }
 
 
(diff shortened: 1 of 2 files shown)


compiler/AST/symbol.cpp
```diff
@@ -567,7 +567,8 @@ bool VarSymbol::isConstant() const {
 
 
 bool VarSymbol::isConstValWillNotChange() const {
-  return hasFlag(FLAG_CONST);
+  return (hasFlag(FLAG_CONST) && !hasFlag(FLAG_REF_VAR)) ||
+         hasFlag(FLAG_REF_TO_CONST);
```
mppf: I'd expect this to return `true` for Qualifier QUAL_CONST.

compiler/AST/symbol.cpp
```diff
@@ -778,12 +779,14 @@ bool ArgSymbol::isConstant() const {
   return retval;
 }
 
+// keep in sync with isConstValWillNotChange() in visibleFunctions.cpp
```
mppf: Why doesn't that function call this method for ArgSymbols in addition to VarSymbols?

vasslitvinov:
Superseded by #7302 .

PR #7204, opened 2017-09-06
Include modules/packages in C compiler search path

mppf:
This is to enable the new Crypto.chpl module to include a C header file implementing a workaround. I expect in the future, we'll want mason packages to be able to #include C files in their directory ( that is, pass a -I path to the C compiler).

Passed full local testing.

### DIFF
--- a/runtime/make/Makefile.runtime.include
+++ b/runtime/make/Makefile.runtime.include
@@ -63,6 +63,7 @@ endif
 # Some tasking layers put something in RUNTIME_INCLS first.
 RUNTIME_INCLS += \
         -I. \
+        -I$(CHPL_MAKE_HOME)/modules/packages \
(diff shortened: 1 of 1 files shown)


ben-albrecht:
>  I expect in the future, we'll want mason packages to be able to #include C files in their directory ( that is, pass a -I path to the C compiler).

Agreed. For reference, this is tracked in #7106

gbtitus:
(I came to this long after the fact, while creating the monthly runtime report.)

I'm curious why it is that runtime/make/Makefile.runtime.include:RUNTIME_INCLS
is the best place to put this new `-I`.  Based on the description it sounds like this
is to support a #include done when building a target program.  If that's correct, it
seems like runtime/etc/Makefile.include:HEADER_CFLAGS would have been more
appropriate. Or am I misunderstanding?

mppf:
@gbtitus - I believe you are right & we should move it to HEADER_CFLAGS (and verify that the Crypto module tests still pass).

PR #7205, opened 2017-09-06
Optimize alignment-related fields out of non-stridable ranges

bradcray:
This PR optimizes the storage related to alignment out of the range type for non-stridable ranges, similar to what was done for the `stride` field earlier.  Initial results suggest that performance is not significantly impacted positively or negatively, but the more compact execution time representation seems obviously desirable.  In order to maintain as much of the past behavior as possible, this changes causes the application of the `align` operator to a non-stridable range or domain to result in a stridable range (one with stride `1`).

More specifically, prior to this change, non-stridable ranges stored an alignment when the `align` operator was applied, which was then seemingly overwritten upon applying a 'by' operator to them making it nearly pointless (except perhaps in the case of fully- and lower-unbounded ranges).  Here, I'm having non-stridable ranges with a lower bound return their low bound as their alignment and those without a lower bound return '0' as their alignment, which is consistent with what we did before.  Here are the resulting user-visible semantic changes that take place as a result of this PR:

range | alignment before | alignment now
------|------------------|--------------
1..n    | 0                             | 1
1..      | 0                             | 1

The above are attractive because, previously, `1..n` returned alignment `0` while `1..n by 1` returned alignment `1` which seems surprising/confusing/inconsistent/weird.

In addition, I changed the interpretation of "natural alignment" such that non-stridable unbounded ranges and upper bounded ranges are never naturally aligned.  My rationale is that these ranges start from an unspecified lower bound and if one applied a 'by 2' to them, there wouldn't be any obvious indication whether counting should start from odd or even numbers.  Due to this change, the range expression `.. align 0` used to print out simply as `..` where now it prints out as `.. align 0` on the argument that since `..` has no natural alignment, any specified alignment should print out as such.  Upper bounded ranges with a negative stride are defined as being naturally aligned if their aligned high bound is the same as their high bound.

A mostly unrelated change in this PR is that strided block-sparse domains/arrays are now supported; this is a follow-on to #7573 which simply wasn't exercised by previous tests; it is now because we have a test that applies 'align' operators to the dense base domain (resulting in a `stridable` domain now).

The work involved here related to:
* guarding writes to the internal `_alignment` field to only do them in `stridable` cases to avoid attempts to assign to a `void` field
* making the `_alignment` and `_aligned` fields `void` in the case that `stridable` is false
* created a second range() constructor for non-stridable ranges that can handle void alignment and aligned fields
* specialized the `.alignment` query to handle stridable, non-stridable, and non-bounded cases appropriately.
* make code that used to refer to `_aligned` and `_alignment` now refer to `aligned` and `alignment` in order to get the logical values rather than a potential `void` field
* made the `align` operator always generate a `stridable=true` result

TODO:
- [x] finish writing this merge message by adding detail
- [x] get team buy-in / approval
- [x] test perf impact in playground
- [x] see whether documentation needs updating


vasslitvinov:
This is great.

PR #7207, opened 2017-09-06
Update two more initializer tests after Mike's improvements from yesterday

lydia-duncan:
With Mike's improvements to copy initializers, and my improvement to
PRIM_SET_MEMBER while supporting nested concrete classes, these tests
now compile and run as expected with a constructor to initializer conversion.

Details:
test/types/records/noakes/return.chpl:
Removed the autoCopy and initCopy functions as they are no longer necessary.
These initializers would benefit from Phase 1 as the default.

test/types/records/sungeun/ctorWithOnFn.chpl:
Converts R0's constructor into an initializer and adds a zero-argument
initializer (since we don't generate that when an explicit initializer is
present).
If we decide to allow on clauses around field initialization, this test would
benefit from Phase 1 as the default.

test/parallel/sync/bradc/syncArrInit.chpl:
This test would benefit from Phase 1 as the default, or the ability to set const
fields in Phase 2.

Passed a sanity paratest

### DIFF
--- a/test/parallel/sync/bradc/syncArrInit.chpl
+++ b/test/parallel/sync/bradc/syncArrInit.chpl
@@ -1,8 +1,9 @@
 class C {
   const irng: range;
 
-  proc C(v1: int) {
+  proc init(v1: int) {
     irng = v1..v1;
+    super.init();
   }
 }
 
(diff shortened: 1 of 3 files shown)


psahabu:
Looks good to me. 👍

PR #7215, opened 2017-09-06
Fix errors + inheritance

### LINKED ISSUE #6315: errors with subclasses that throw errors
### Summary of Problem

The following program produces:

    subclass-throws.chpl:13: internal error:
    number of actuals does not match number of formals in foo() [callDestructors.cpp:1302]

I expected

```
C.foo()
D.foo()
D.foo()
Fini
```

### Steps to Reproduce

**Source Code:**

```chapel
class C {
  proc foo() throws { writeln("C.foo()");}
}

class D : C {
  proc foo() throws { writeln("D.foo()"); }
}

var c = new C();
var d = new D();
var dc:C = new D();

try { c.foo(); } catch { writeln("c.foo threw"); }
try { d.foo(); } catch { writeln("d.foo threw"); }
try {dc.foo(); } catch { writeln("dc.foo threw"); }

delete c;
delete d;
delete dc;

writeln("Fini");
```

Curiously, though the error is reported on the line calling c.foo(),
if I remove from the program everything having to do with class D, var
d, and var dc, it compiles and executes the c.foo() line.

The `class D` code is there because this is slightly simplified from a
test I expected to fail to compile, but which does compile but
segfaults when run.  If the `throws` is removed from `C.foo()`, and
`D.foo()` is changed to `throw new Error();` after its `writeln()`, the
program compiles and runs, but segfaults:

```
fortytwo@magrathea:~/src/chpl$ chp subclass-throws.chpl
C.foo()
D.foo()
D.foo()
*** Caught a fatal signal: SIGSEGV(11) on node 0/1
NOTICE: Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace. 
```

Curiously, `d.foo threw` wasn't printed, and the segfault was between
the second `D.foo()` and the `Fini`.

(The gasnet backtrace doesn't look interesting to me, but I can provide it.)

I expected this to be a compilation failure because `D.foo()`
shouldn't be allowed to throw when `C.foo()` doesn't.  And that's
because since `C.foo()` doesn't throw, the call to `dc.foo()`
shouldn't be able to.  (IMO)

**Compile command:**
chpl subclass-throws.chpl -o subclass-throws

**Execution command:**
./subclass-throws -nl 1

None yet

### Configuration Information

```
fortytwo@magrathea:~/src/chpl$ chpl --version
chpl Version 1.16.0 pre-release (23d7df3)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
fortytwo@magrathea:~/src/chpl$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
  CHPL_JEMALLOC: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
fortytwo@magrathea:~/src/chpl$ gcc --version
```

mppf:
Fixes #6315.

In addition to making throwing functions work with inheritance at all, this PR also adds a compilation error in the event that an overridden function has a different throws/not throwing option.

Passed full local testing.

### DIFF
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -390,7 +390,25 @@ static void addToVirtualMaps(FnSymbol* pfn, AggregateType* ct) {
               USR_FATAL_CONT(pfn, "conflicting return type specified for '%s: %s'", toString(pfn), pfn->retType->symbol->name);
               USR_FATAL_CONT(fn, "  overridden by '%s: %s'", toString(fn), fn->retType->symbol->name);
               USR_STOP();
+            } else if (fn->throwsError() != pfn->throwsError()) {
+              USR_FATAL_CONT(fn, "conflicting throws for '%s'", toString(fn));
+              const char* pfnThrowing = NULL;
+              const char* fnThrowing = NULL;
+
+              if (pfn->throwsError()) {
+                pfnThrowing = "throwing";
+                fnThrowing = "non-throwing";
+              } else {
+                pfnThrowing = "non-throwing";
+                fnThrowing = "throwing";
+              }
+
+              USR_FATAL_CONT(pfn, "%s function '%s'",pfnThrowing,toString(pfn));
+              USR_FATAL_CONT(fn, "overridden by %s function '%s'",
+                             fnThrowing, toString(fn));
+              USR_STOP();
             }
+
             {
               Vec<FnSymbol*>* fns = virtualChildrenMap.get(pfn);
               if (!fns) fns = new Vec<FnSymbol*>();
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -323,7 +323,18 @@ void ErrorHandlingVisitor::lowerCatches(const TryInfo& info) {
 bool ErrorHandlingVisitor::enterCallExpr(CallExpr* node) {
   bool insideTry = !tryStack.empty();
 
-  if (FnSymbol* calledFn = node->resolvedFunction()) {
+  // The common case of a user-level call to a resolved function
+  FnSymbol* calledFn = node->resolvedFunction();
+
+  // Also handle the PRIMOP for a virtual method call
+  if (calledFn == NULL) {
+    if (node->isPrimitive(PRIM_VIRTUAL_METHOD_CALL)) {
+        SymExpr* arg1 = toSymExpr(node->get(1));
+        calledFn = toFnSymbol(arg1->symbol());
+    }
+  }
+
+  if (calledFn != NULL) {
     if (calledFn->throwsError()) {
 
       SET_LINENO(node);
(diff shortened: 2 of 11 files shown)


compiler/resolution/virtualDispatch.cpp
```diff
@@ -390,7 +390,25 @@ static void addToVirtualMaps(FnSymbol* pfn, AggregateType* ct) {
               USR_FATAL_CONT(pfn, "conflicting return type specified for '%s: %s'", toString(pfn), pfn->retType->symbol->name);
               USR_FATAL_CONT(fn, "  overridden by '%s: %s'", toString(fn), fn->retType->symbol->name);
               USR_STOP();
+            } else if (fn->throwsError() != pfn->throwsError()) {
```
psahabu: If the parent class function throws, then I think the child class function could be throws or not throws. In a way, a non-throwing function is a subtype of a throwing function. What do you think?

mppf: I agree that a non-throwing function can be viewed as a subtype of a throwing one. However, I'm uncomfortable with the implementation changes we'd have to do at this time.

In particular, in the implementation of virtual dispatch, what we have is a table from runtime class ID to the method that should be called. If we just allowed it, we would pass an extra argument (out error) to a dynamically called function. That seems to be at least surprising and potentially an error. (Do you know if when calling a function through a function pointer, it is always OK to pass extra arguments not expected by the function?)

psahabu: This might be a question for @dmk42? Theoretically the arguments get pushed on the stack in reverse order so the later arguments just get ignored, but that might just be happy coincidence rather than something spec enforced.

dmk42: When targeting C, if the number of arguments does not equal the expected number, it is undefined behavior.  In practice, most ABIs end up just ignoring an extra argument as long as it's tacked onto the end, though we can't count on that.  When targeting LLVM, we have more control and could set things up to ensure extra arguments just get ignored, if we want to do it that way.

mppf: @dmk42 - thanks for the clarification. I think it's sufficient for it to be an error for now & we can consider relaxing it in the future.

PR #7216, opened 2017-09-06
Remove task counting vestiges from the runtime

ronawho:
Remove chpl_task_getNumRunningTasks() from the runtime interface.
chpl_task_getNumRunningTasks was "deprecated" with bdc315e5172, and #7193
moved the remaining task-counting code from the runtime into the
compiler/modules so there's no reason to keep this interface around.


ronawho:
@gbtitus any objections?

gbtitus:
No objections -- 👍 !

PR #7218, opened 2017-09-07
create a SystemError hierarchy

psahabu:
While changing from `syserr` to `SystemError` was straightforward enough, it didn't allow users to handle different kinds of `SystemError`s. Rather, they had to handle `SystemError` and inspect the internal `err: syserr` manually. This PR addresses that awkwardness by creating a `SystemError` hierarchy [modeled after that of OSError in Python](https://docs.python.org/3/library/exceptions.html#os-exceptions), as well as a factory type method that takes in a `syserr` and emits the appropriate subtype of `SystemError`.

Tests have also been updated to reflect the change.


modules/standard/SysError.chpl
```diff
@@ -52,6 +52,164 @@ class SystemError : Error {
   proc writeThis(f) {
     try! f.write(msg);
   }
+
+  proc type fromSyserr(err: syserr, msg: string = "") {
+    var strerror_err: err_t = ENOERR;
+    var errstr : c_string = sys_strerror_syserr_str(err, strerror_err);
+    var err_msg: string   = errstr:string;
+    if !msg.isEmptyString() then
+      err_msg = errstr:string + " (" + msg + ")";
```
mppf: I see you are combining the msg argument with the result of strerror here. A reasonable alternative would be to create an overload of Error's proc message() in SystemError that does this combining. It's not strictly necessary to do it when creating the SystemError.

modules/standard/SysError.chpl
```diff
@@ -52,6 +52,164 @@ class SystemError : Error {
   proc writeThis(f) {
     try! f.write(msg);
   }
+
+  proc type fromSyserr(err: syserr, msg: string = "") {
```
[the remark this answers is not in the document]
mppf: I don't think the 'msg' argument here should be called 'msg' since it's an elaboration on the error message from strerror. Maybe "details" or something? Obviously it will work no matter what we name this argument. It's just a documentation thing.

modules/standard/SysError.chpl
```diff
@@ -52,6 +52,164 @@ class SystemError : Error {
...
+      return new IsADirectoryError(err_msg);
+    } else if err == ENOTDIR {
+      return new NotADirectoryError(err_msg);
+    } else if err == EACCES || err == EPERM {
+      return new PermissionError(err, err_msg);
+    } else if err == ESRCH {
+      return new ProcessLookupError(err_msg);
+    } else if err == ETIMEDOUT {
+      return new TimeoutError(err_msg);
+    }
```
mppf: Chapel's I/O system adds 3 "system" error codes. Of course these aren't thrown by system calls, but they can be raised doing I/O. They are:
 * EEOF - An error code indicating the end of file has been reached
 * ESHORT - An error code indicating that the end of file or the end of the input was reached before the requested amount of data could be read.
 * EFORMAT - An error code indicating a format error; for example when reading a quoted string literal, this would be returned if we never encountered the opening quote.

It seems to me that each of these should get its own subclass of SystemError.
Of course we will have to make up names for them, but these will almost certainly each be multi-word names to fit in with the others. Maybe
 * EEOF -> EndOfFileError
 * ESHORT -> UnexpectedEndOfFileError
 * EFORMAT -> BadFormatError

test/io/sungeun/ioerror.EACCES.good
```diff
@@ -1,3 +1,3 @@
-uncaught SystemError: Permission denied EACCES with path "this/is/my/path" offset -1
+uncaught SystemError: Input/output error (Permission denied EACCES with path "this/is/my/path" offset -1)
```
mppf: This one looks weird, but it's in fact just because the test does something weird in this case (it doesn't actually pass the EACCESS code).

test/modules/standard/Path/lydia/realpath/doesNotExist.good
```diff
@@ -1,3 +1,3 @@
-uncaught SystemError: No such file or directory in realPath of with path "foo.txt"
+uncaught FileNotFoundError: No such file or directory (in realPath of with path "foo.txt")
```
lydia-duncan: Pre-existing problem, but there's a grammar error in this message

test/types/file/unableToOpenFile.good
```diff
@@ -1,3 +1,3 @@
-uncaught SystemError: No such file or directory in open with path "_test_cannotOpenMe.txt"
+uncaught FileNotFoundError: No such file or directory (in open with path "_test_cannotOpenMe.txt")
```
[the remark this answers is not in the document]
lydia-duncan: Seems reasonable to me.  I don't have strong feelings one way or the other

modules/standard/SysError.chpl
```diff
@@ -52,6 +52,164 @@ class SystemError : Error {
...
   }
+
+  proc type fromSyserr(err: syserr, msg: string = "") {
+    var strerror_err: err_t = ENOERR;
+    var errstr : c_string = sys_strerror_syserr_str(err, strerror_err);
+    var err_msg: string   = errstr:string;
+    if !msg.isEmptyString() then
+      err_msg = errstr:string + " (" + msg + ")";
+
+    if err == EAGAIN || err == EALREADY || err == EWOULDBLOCK || err == EINPROGRESS {
```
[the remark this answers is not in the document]
psahabu: Yes, this came from Python's hierarchy.

psahabu:
@mppf do you mind taking another look? I decided to give each subclass a single constructor with optional `details` and `err` arguments.

modules/standard/SysError.chpl
```diff
@@ -41,17 +41,342 @@ module SysError {
...
+
+*/
 class SystemError : Error {
-  var err: syserr;
+  var err:     syserr;
+  var details: string;
+
+  proc SystemError(err: syserr, details: string = "") {
+    this.err     = err;
+    this.details = details;
```
mppf: FYI. I think this kind of spacing-lining-up can quickly become more trouble than it's worth. I usually try to reserve that effort for limited cases where the spacing makes comparing different lines easier. That said, I don't mind if you do it.

psahabu: Yeah this was probably overkill.

psahabu:
Since it's quick to do, I'm going to handle the comments in your review summary now:
- repeat the syserr documentation
- move EOF/etc. into a separate subclass `FileError`
- remove `InvalidArgumentError` and leave `EINVAL` as a plain `SystemError`

ben-albrecht:
FYI @psahabu - `git bisect` points me to this PR for the regressions in [AST size and compiler performance](http://chapel.sourceforge.net/perf/chapcs/?startdate=2017/09/01&enddate=2017/09/14&graphs=compilationtime,jacobiastsize,jacobiemittedcodesize).

ben-albrecht:
@mppf - is this generated code increase surprising? It's not clear to me if there's anything that can be done about the regressions.

mppf:
It's just more included-by-default code. Until we build separate compilation, the compiler will get slowly slower as we work on the modules. I think building separate compilation is what we can do about it.

PR #7233, opened 2017-09-07
Remove chpl-mem-warning-macros.h

ronawho:
chpl-mem-warning-macros.h was used to prevent accidental calls to the system
allocator in the runtime. It did so by redefining malloc and friends to bogus
values. This often caused build failures on cygwin because system headers were
updated and included the token "malloc", which prevented us from building:

We also had to turn off the warnings anytime we wanted to reference the token
malloc (such as in the jemalloc shim), which was kind of a pain.

We now have travis/jenkins jobs that run ./util/devel/lookForBalloc, so rely on
them instead of chpl-mem-warning-macros.h and chpl-mem-no-warning-macros.h.

This is a follow on to https://github.com/chapel-lang/chapel/pull/6163 and https://github.com/chapel-lang/chapel/pull/6197


ronawho:
@gbtitus and/or @mppf any objections?

mppf:
OK by me, but you should check LLVM testing.

gbtitus:
Good!   👍

ronawho:
Passes extern/ferguson with llvm, that good enough or did you want a full llvm run?

PR #7234, opened 2017-09-07
Attempt to rewrite array-vs-ddata perf test to unify two cases

bradcray:
The goal of this rewrite is to remove some surprising performance
deltas we see in which ddata is slower than arrays even though
arrays wrap ddatas.  The approach I've taken here is to allocate
each data structure just before the loop that times it rather
than allocating each at program start and then running the loops.
To do this, I also changed how the verification takes place since
both arrays no longer live simultaneously.

Experiments by Elliot and myself suggest that this fixes the
strange inversion we were seeing before.

### DIFF
--- a/test/arrays/diten/time_array_vs_ddata.chpl
+++ b/test/arrays/diten/time_array_vs_ddata.chpl
@@ -6,16 +6,21 @@ config const niters = 10000000;
 config param printTimes = false;
 
 proc main {
-  var A: [0..#n] int;
-  var AA = _ddata_allocate(int, n);
   var t = new Timer();
-  t.start();
-  for j in 0..#niters {
-    for i in 0..#n {
-      A[i] += i;
+
+  {
+    var A: [0..#n] int;
+    t.start();
+    for j in 0..#niters {
+      for i in 0..#n {
+        A[i] += i;
+      }
     }
+    t.stop();
+    for i in 0..#n do
+      if A[i] != niters*i then
+        writeln("error: A[",i,"] = ", A[i]);
   }
-  t.stop();
 
   if printTimes {
     writeln("array ", t.elapsed());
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho: Take a look at this rewrite as an attempt to remove some of the mystery of arrays seeming to outperform ddatas.  In a quick check on my laptop this seemed to make ddatas slightly faster than arrays as you'd expect (where they were slightly slower on master).  In any case, it seems like it should be no more poorly behaved than the current version.

@daviditen may also want to take a look as the original author of the test.

ronawho:
👍 I think this is worth running on chapcs07 too to verify your perf results (I can do that if you want)

test/arrays/diten/time_array_vs_ddata.chpl
```diff
@@ -6,34 +6,45 @@ config const niters = 10000000;
...
+  {
+    var A = _ddata_allocate(int, n);
+    t.start();
+    for j in 0..#niters {
+      for i in 0..#n {
+        A[i] += i;
+      }
     }
+    t.stop();
+    _ddata_free(A, n);
```
daviditen: This `free`s the `_ddata` right before  walking it to check for correctness.

daviditen:
Other than my comment about the `_ddata_free` this looks fine to me.

bradcray:
Good catch, @daviditen, thanks.  @ronawho: I don't feel the need to time it per se (again, seems like it could be no worse and is only more principled; and any results may be subject to the same noise we've seen), but if you want to, I'm not going to protest.

ronawho:
Results on chapcs07 mimicked what you saw on your machine (ddata is now slightly faster than array)

```
# Date  array   ddata
09/07/17        0.161287        0.227237
09/07/17        0.165906        0.227126
09/07/17        0.161361        0.227893

09/07/17        0.160183        0.153577
09/07/17        0.162146        0.155957
09/07/17        0.160446        0.153761
```

PR #3388, opened 2016-03-02
Clean up jemalloc shared heap support

ronawho:
Move all our static variables that are used to support the shared heap into a
struct. This is a minor code cleanup, it shouldn't impact performance or
anything.

See JIRA 186 for more info: https://chapel.atlassian.net/browse/CHAPEL-186

PR #7237, opened 2017-09-07
Remove numa-localization support from jemalloc shim

ronawho:
This effectively reverts #5689. This cripples the already mothballed (but still
somewhat available) multi-ddata support, so we should pobably remove it
completely.

This simplifies the jemalloc shim in preparation for the jemalloc 5.0 upgrade
which will require non-trivial updates to our allocation hooks.

### DIFF
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -164,16 +164,11 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
     if (p == NULL) {
       p = chpl_mem_allocMany(nmemb, eltSize, CHPL_RT_MD_ARRAY_ELEMENTS,
                              lineno, filename);
-      // TODO is this right?
       do_localize = (subloc == c_sublocid_all) ? true : false;
     }
 
     if (do_localize) {
-      if (subloc == c_sublocid_all) {
-        // TODO can this be removed? I think it's multi-ddata only?
-        chpl_topo_setMemSubchunkLocality(p, size, true, NULL);
-      }
-      else if (isActualSublocID(subloc)) {
+      if (isActualSublocID(subloc)) {
         chpl_topo_setMemLocality(p, size, true, subloc);
       }
     }
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -29,7 +29,6 @@
 #include "arg.h"
 #include "chpl-comm.h"
 #include "chpl-mem-hook.h"
-#include "chplsys.h"
 #include "chpl-topo.h"
 #include "chpltypes.h"
 #include "chpl-tasks.h"
@@ -65,10 +64,6 @@ void chpl_mem_exit(void);
 
 int chpl_mem_inited(void);
 
-// predeclared here because we need them below; actual definitions
-// are near the end
-static chpl_bool chpl_mem_localizes(void);
-static size_t chpl_mem_localizationThreshold(void);
 
 static inline
 void* chpl_mem_allocMany(size_t number, size_t size,
@@ -169,11 +164,13 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
     if (p == NULL) {
       p = chpl_mem_allocMany(nmemb, eltSize, CHPL_RT_MD_ARRAY_ELEMENTS,
                              lineno, filename);
-      do_localize = !chpl_mem_localizes();
+      // TODO is this right?
+      do_localize = (subloc == c_sublocid_all) ? true : false;
     }
 
     if (do_localize) {
       if (subloc == c_sublocid_all) {
+        // TODO can this be removed? I think it's multi-ddata only?
         chpl_topo_setMemSubchunkLocality(p, size, true, NULL);
       }
       else if (isActualSublocID(subloc)) {
@@ -264,29 +261,6 @@ void chpl_mem_layerExit(void);
 void* chpl_mem_layerAlloc(size_t, int32_t lineno, int32_t filename);
 void* chpl_mem_layerRealloc(void*, size_t, int32_t lineno, int32_t filename);
 void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
-
-//
-// Does the implementation provide allocated memory that is already
-// localized to the calling sublocale?  That is, when the locale model
-// has sublocales and an allocation is done while running on one, will
-// the allocated memory be localized to that sublocale?  Note that the
-// answer doesn't have to be completely truthful, but inaccuracy can
-// result in poor performance for applications sensitive to sublocale
-// (NUMA, for example) affinity.
-//
-#ifndef CHPL_MEM_IMPL_LOCALIZES
-  #define CHPL_MEM_IMPL_LOCALIZES() false
-#endif
-
-static inline
-chpl_bool chpl_mem_localizes(void) {
-  return CHPL_MEM_IMPL_LOCALIZES();
-}
-
-static inline
-size_t chpl_mem_localizationThreshold(void) {
-  return chpl_topo_getNumNumaDomains() * 2 * chpl_getHeapPageSize();
-}
 
 #else // LAUNCHER
 
--- a/runtime/src/mem/jemalloc/mem-jemalloc.c
+++ b/runtime/src/mem/jemalloc/mem-jemalloc.c
@@ -28,8 +28,6 @@
 #include "chpl-comm.h"
 #include "chpl-mem.h"
 #include "chplmemtrack.h"
-#include "chplsys.h"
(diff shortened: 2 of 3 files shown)


ronawho:
@gbtitus thoughts?

I basically manually reverted #5689 (and did some comparison against modified diffs to make sure I didn't miss anything) though I didn't touch anything from chpl-topo.

Note that the only functions from chpl-topo that are currently used after the PR are `chpl_topo_setMemSubchunkLocality()` and `chpl_topo_setMemLocality()` both from `chpl-mem.h::chpl_mem_array_alloc()`, though I'm not even sure if `chpl_topo_setMemSubchunkLocality()` is needed. However, I did not take on removing anything from chpl-topo myself, as I'm not sure what might be useful in the future.

runtime/include/chpl-mem.h
```diff
@@ -169,11 +164,13 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
     if (p == NULL) {
       p = chpl_mem_allocMany(nmemb, eltSize, CHPL_RT_MD_ARRAY_ELEMENTS,
                              lineno, filename);
-      do_localize = !chpl_mem_localizes();
+      // TODO is this right?
```
gbtitus: Would you mind leaving chpl_mem_localizes() and the associated logic?  While it's true that with this change it collapses to false in all cases including jemalloc and therefore could be removed, as soon as jemalloc can allocate already-localized memory it will be true again.  (Which is worthwhile because then we can avoid the chpl_topo_setMemLocality() below.)  And the check against the subloc value is handled later if we're going to localize at all.

runtime/include/chpl-mem.h
```diff
@@ -169,11 +164,13 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
...
+      do_localize = (subloc == c_sublocid_all) ? true : false;
     }
 
     if (do_localize) {
       if (subloc == c_sublocid_all) {
+        // TODO can this be removed? I think it's multi-ddata only?
```
gbtitus: Yes, subloc==c_sublocid_all can only be true when we have multi-ddata, so this branch can go.

runtime/include/chpl-mem.h
```diff
@@ -265,29 +262,6 @@ void* chpl_mem_layerAlloc(size_t, int32_t lineno, int32_t filename);
 void* chpl_mem_layerRealloc(void*, size_t, int32_t lineno, int32_t filename);
 void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
 
-//
```
gbtitus: As noted above, I think that although chpl_mem_localizes() isn't needed once this change goes in, it will be useful in the near future and we should consider keeping it.  (If we do keep it now it reduces to compile-time constant, so it doesn't hurt execution time.)  chpl_mem_localizationThreshold() seems to be unused (oops!) so it can definitely go.

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -44,111 +42,12 @@
...
-    // The chpl_topo_setMemSubchunkLocality() call above isn't
-    // having the effect we desire, for reasons that are mostly but
-    // not completely understood.  So for now, to get the locality
-    // we want, touch the pages into existence while executing on
-    // the desired NUMA domains.
-    //
-    chpl_topo_touchMemFromSubloc(heaps[hpi].base, heaps[hpi].size, true, hpi);
-  }
-}
+} heap; // static, will be "zero" initialized automatically
```
gbtitus: Do we really need to tell programmers that file=level C variables are zero-initialized?

ronawho: Note that this is a reversion back to what was there. 

That said it could probably be removed. I originally added it to note why there was no explicit initializer: https://github.com/chapel-lang/chapel/pull/3388#r54660489

runtime/include/chpl-mem.h
ronawho: I'm not sure I'm following. What are we going to need that for? Not for dynamic registration or chunks, right?

ronawho: (personally I'd like to remove it still, it's trivial to re-add in the future if we do need it)

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: It does seem odd to have this comment on this one variable and not have a similar one on all the rest.  I won't die if it stays, though.

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -371,19 +265,12 @@ static void get_small_and_large_class_sizes(size_t* classes) {
...
-    uintptr_t u_base = (uintptr_t)heaps[hpi].base;
-    uintptr_t u_top =  u_base + heaps[hpi].size;
-    if (u_ptr >= u_base && u_ptr < u_top) {
-      return false;
-    }
-  }
-  return true;
+  uintptr_t u_base = (uintptr_t)heap.base;
+  uintptr_t u_top = u_base + heap.size;
+  return (u_ptr < u_base) || (u_ptr > u_top);
```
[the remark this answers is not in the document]
ronawho: (in a followup PR though, I'd really like to keep this as close as possible to a direct reversion to make testing easier)

runtime/include/chpl-mem.h
gbtitus: It seems like it would be useful for chunks.  With the chunk work we can localize (via the topo layer) each chunk that we add to an arena so that it's in the same NUMA domain as the thread which is extending that arena (which is presumably also running in the NUMA domain).  Subsequent allocations from the same or other tasks running on that thread will then also have good affinity.  Given this, we want to say that the memory layer localizes allocations, so we can avoid calling chpl_topo_whatever() to post-localize the allocation later in chpl_mem_array_alloc().

runtime/src/mem/jemalloc/mem-jemalloc.c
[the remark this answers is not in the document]
gbtitus: Sure, that's okay.

runtime/include/chpl-mem.h
gbtitus: We can add it back later, yes.

ronawho: Ah ok. Good point. I forgot that nic registering will pin to numa 0 unless it's already pinned. For some reason I was thinking it would still use sched_affinity to pin to whatever domain the current thread is running on. 

So I definitely agree that we'll want to localize a chunk, but I'm not sure there will be any clients that need to know that information. Currently nobody calls chpl_mem_array_alloc() with anything other than `c_sublocid_all` or `c_sublocid_none`, so the chpl_topo_setMemLocality() is never actually called, right? Do we think that will change in the near future?

runtime/include/chpl-mem.h
ronawho: Cool, thanks

If it's ok with you, I plan to leave all the current functions in chpl-topo for now, and I'll let you remove any you don't think we'll need

runtime/include/chpl-mem.h
ronawho: (and I don't mean to make a big deal out of this, if you think it's important I'm happy to leave it)

gbtitus: Ah.  Maybe not.  Maybe we can just yank out the whole do_localize-based if-stmt that controls the post-alloc topo localization calls further down, and rely on first-touch in all cases where it can have an effect in the first place.

A bit of background: chpl_mem_localizes() used to be called chpl_mem_impl_localizes(), which is arguably a better name because this is a function that the general code here maybe, or maybe not) needs to call but nobody else does.  So even if we keep it it doesn't need to be defined as part of the publicized interface but it does need to be provided by every implementation (even if vacuously, by not overriding the default), so the it's available to be used by the interface code.

runtime/include/chpl-mem.h
gbtitus: Yes, I'd prefer we leave chpl-topo alone for now.

gbtitus:
Using github as an IM system: we're headed for lunch and I have to leave this meeting room.  I think this is okay to go in -- if I have to or want to resurrect any of the code you're removing later it will certainly be easy to do so.  So it's okay to take it all out; I think the discussion has still been useful.
👍

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: Will remove it

ronawho:
Removed static comment, fixed heap bug, and removed the TODOs. 

I'll merge later today, unless you catch anything else

gbtitus:
I'm good with this.  👍

PR #7240, opened 2017-09-08
Support grow by more than 1 in resizeAllocRange

ben-albrecht:
`resizeAllocRange` only expected a difference of `1` between range sizes `r` and `r2`.  Consequently, it assumes `r.size+1` or `r.size*1.5` will be sufficiently large to accommodate `r2`.

This worked previously, because all the calls to `resizeAllocRange` only had a size delta of `1`. However, In #7180 we need to make bulk array resizes, which breaks down this assumption.

This PR modifies `resizeAllocRange` to account for `r2.size` when resizing, to accommodate any range delta `> 1`.

I also added a test for this functionality.

## Testing

- [x] linux64

### DIFF
--- a/test/arrays/methods/internal/resizeAllocRange.chpl
+++ b/test/arrays/methods/internal/resizeAllocRange.chpl
@@ -1,0 +1,33 @@
+/* Test to exercise the Grow code paths in _array.resizeAllocRange */
+
+// Dummy array to call methods on
+var Array = [1];
+
+
+//
+// direction = +1
+//
+
+// Grow by 1, small (r1+1)
+assert(Array.resizeAllocRange(1..3, 1..4) == 1..4);
+
+// Grow by 1, large (1.5x)
+assert(Array.resizeAllocRange(1..100, 1..1) == 1..150);
+
+// Grow by more than 1 (r2)
+assert(Array.resizeAllocRange(1..3, 1..7) == 1..7);
+
+
+//
+// direction = -1
+//
+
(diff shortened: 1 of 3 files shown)


ben-albrecht:
@daviditen - requesting review.

Something that occurred to me when making this change and adding tests:

Should `resizeAllocRange` inherit the value of `r` from `this._value.dataAllocRange`? If not, should it be a private helper function, since it doesn't access/modify any array fields?

test/arrays/methods/internal/resizeAllocRange.chpl
```diff
@@ -0,0 +1,33 @@
+/* Test to exercise the Grow code paths in _array.resizeAllocRange */
```
daviditen: As a hedge against arrayAsVecGrowthFactor changing in the future and breaking this test you could explicitly set it to 1.5 in a .compopts file.

ben-albrecht: good call

daviditen: This looks good to me assuming your testing is clean.

ben-albrecht: Cool - paratesting now!

ben-albrecht:
> Should resizeAllocRange inherit the value of r from this._value.dataAllocRange? 

Offline, @daviditen responded: 👍 

I will make this change in #7180 along with some other refactoring.

PR #7241, opened 2017-09-08
Increase default gasnet exit timeout for gemini/aries

ronawho:
GASNet 1.30.0 included a gemini/aries update to "Improve robustness of exit
code". It appears that the improved robustness in combination with re-enabling
GASNet's multi-domain feature in #6893 has resulted in significantly higher
exit times. I'll ping the GASNet folks, but for now just increase the exit
timeout to avoid testing regressions.

### DIFF
--- a/runtime/src/comm/gasnet/comm-gasnet.c
+++ b/runtime/src/comm/gasnet/comm-gasnet.c
@@ -778,6 +778,7 @@ static void set_max_segsize() {
 }
 
 static void set_num_comm_domains() {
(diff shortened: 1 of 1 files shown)


ronawho:
@dmk42 this should resolve the sigalarms on knl for now, but I'll ping the gasnet folks to see if there's a better solution

ronawho:
init and exit should be called from the same, unless there's an abnormal exit, but the regressions I'm seeing are for clean exits with just the basic hello.chpl

Also, how did you possibly see this issue already?? Do you guys have something that scrapes for references to GASNet?

PR #7243, opened 2017-09-08
gen_release should confirm the Git source version it finally uses

awallace-cray:
Adds "git status; git log -1" right after Git clone and (optional)
git checkout are finish. Makes it easier to confirm the exact
version of Chapel being built.

Flags all messages with "[gen_release]" to make them easier to
recognize.


awallace-cray:
Tested. Doesn't change anything about the release tarball or how it is generated. Only some messages to the console.

PR #7245, opened 2017-09-08
Enable strict error handling by default for modules

mppf:
Continuing PR #7244, this PR turns on strict error handling for more code and adjusts the modules and tests appropriately to match.

This PR adjusts several `writeThis` methods to use `<~>` instead of `channel.write` since the operator saves an error code in the channel instead of throwing. This is a workaround for the problem discussed in issue #7261.

This PR follows on earlier work in which the top-level `proc writeln` and associated routines (`write`, `writef`) do not throw. The idea is that they appear frequently enough & a program would be messed up enough if they didn't work that it should be a fatal error. In contrast, the methods on channels, such as `proc channel.writeln()` *do* throw since it's possible to encounter errors when doing file or network I/O.

Along the same lines, we considered possibly making `string.format()` not throw, but there is really only one name for this routine, and it's definitely less common than `writeln`.

- [x] full local testing


modules/packages/RecordParser.chpl
```diff
@@ -121,11 +121,11 @@ class RecordReader {
 
   /* Create a RecordReader to match an auto-generated regular expression
      for a record created by the :proc:`createRegexp` routine.
-     
+
      :arg t: the record type to read
      :arg myReader: the channel to read from
    */
-  proc RecordReader(type t, myReader) {
+  proc RecordReader(type t, myReader) throws {
```
psahabu: Have we done a `throws` on a constructor?

mppf: No, but it seems to pass testing. I plan to turn it in to try! if anything goes wrong.

test/io/ferguson/writef.chpl
```diff
@@ -1,3 +1,4 @@
+pragma "error mode fatal"
```
bradcray: I was just reviewing this to see what tests needed to be updated with this switch.  Is it the use of writef()s in this test that requires a switch or something else?

test/release/examples/primers/opaque.chpl
```diff
@@ -7,6 +7,7 @@
    support unstructured arrays like graphs. They are a special case of
    :ref:`associative domains and arrays <primers-associative>`.
 */
+pragma "error mode fatal"
```
bradcray: It seems unfortunate to require the pragma in release/examples tests, and particularly in a primer.  In this specific case, I wonder if the explicit `module` declaration could/should just be removed (most primers don't seem to use them unless they're teaching modules?).  In other cases (benchmarks), I'm wondering whether try! could/should be used instead?

bradcray: Looking at the rendered version of this online makes me think even more that we should just remove the module declaration in this case.  It doesn't seem to have a purpose.

test/io/ferguson/writef.chpl
mppf: Actually this one doesn't really need the switch (it did in an earlier version of the PR, when I had the top-level writef throw - it does not now).

test/release/examples/primers/opaque.chpl
mppf: I think it's a matter of file naming. We don't want the implicit module to be named `opaque` because that might interfere with things like `domain(opaque);`. If we're willing to rename it to opaque-domains or something like that, I think we can drop the module declaration.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
```diff
@@ -101,7 +101,7 @@ module main {
     run_loop[LoopKernelID.FIND_FIRST_MIN] = runC_findFirstMin;
 
     if output_dirname.length > 0 then
-      mkdir(output_dirname, parents=true);
+      try! mkdir(output_dirname, parents=true);
```
bradcray: Having just run release-over-release performance gathering, it occurs to me that a downside of this edit (which I liked on first blush due to its precision) is that it will prevent getting performance on this test for past releases of Chapel.  :(  

This wouldn't be much better, but is something I've been curious about for other reasons: Do we still have a global compiler flag that can be used to dial down strictness?  How does it play w.r.t. the pragma?

test/release/examples/primers/opaque.chpl
bradcray: In updating it, I'd pull the module declaration anyway I think.  It just clutters the documentation.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
mppf: We don't currently have a global compiler flag to dial down the strictness. Recall that before, we had only --strict-errors and --no-strict-errors, where --no-strict-errors was the global default. But I removed that flag with my earlier PR. We could certainly add a flag for --fatal-errors e.g.. Do you think we should?

test/release/examples/primers/opaque.chpl
mppf: I have PR #7277 to fix this about the opaque primer.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
mppf: Also, if we want this test to work release-over-release, we could just have it call the out error= version of mkdir and include its own halt call. That should work the same before & after all the error handling changes.

test/release/examples/primers/opaque.chpl
bradcray: Sorry, I got my wires crossed:  I'd seen your "this doesn't need the pragma" comment on email and thought it was on this comment stream rather than the previous one (which is why my last response probably didn't parse well).  Bummer about the need to dodge the implicit module name, but I'm OK with that PR.

The comment that kicked this stream off was also intended to wonder about what could be done to avoid using the pragma in other release/examples tests as well—I just attached it to this one since it was one of the most user-facing.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
bradcray: Thanks for the info (I'd lost track of the flag).  W.r.t. adding a flag: I don't feel certain about it, but while musing over the "fine-grain control" conversation yesterday and possible usage modes, it seemed likely that one might be "I'm going to use explicit modules, but I really don't want to be strict about my error-handling (yet?) and don't want to have to use a pragma on each module," which might suggest supporting such a compiler flag.

I'm not sure whether --fatal-errors is the right name for it or not.  Intuitively, it seems to me that the flag would be saying something like "explicit modules are relaxed w.r.t. error-handling requirements."  But this may be another conversation that suggests "Brad needs to refresh his understanding of the three modes."

bradcray: > Also, if we want this test to work release-over-release, we could just have it call the out error= version of mkdir and include its own halt call.

That might be nice, but is it a temporary solution?  (i.e., are we planning on deprecating those interfaces in a future release?)

test/release/examples/primers/opaque.chpl
mppf: I can get rid of the test/release/examples/benchmarks/ssca2 ones by removing some module declarations (so we instead use the implicit module). Arguably the module statements make it clearer what is going on, but I think it'd be acceptable to remove them and rely on the implicit default. Sound OK?

test/release/examples/benchmarks/lcals/LCALSMain.chpl
mppf: I'm uncertain about how temporary it would be. I think we intend to deprecate those interfaces at some point, yes, but using an `error=` argument seems like a reasonable "for now" change here.

For the compiler flag, what do you think about `--halt-on-errors` ?

bradcray: OK, let's do the "for now" change since this is an important performance test.  In non-performance cases, though, I like leaning on the new features more (so wish we could here as well... but not enough I guess).

In English, is the behavior of this flag "halt on errors" or "halt on uncaught errors" or "..."?

mppf: "Instead of having a compiler error on unhandled errors, halt at runtime if one comes up"

test/release/examples/primers/opaque.chpl
bradcray: I think this sounds OK.  If the modified versions in the studies directory is also a performance test, I might do the same there in order to keep it compiling with older compilers.  If not, leaving it as-is is fine.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
bradcray: --halt-on-unhandled-errors?
--unhandled-errors-halt?

The main hesitation I have with all three of the above suggestions (including yours) is that it almost sounds as though we're requesting a different runtime behavior for a given program when in fact we're really requesting a different compile-time behavior, aren't we?  Something like "permit errors to be unhandled and halt on them"?

(You can tell me to go back and read about the three modes again at any time that I'm talking nonsense).

mppf: PR #7283 is about the lcals changes. I'll ask David to look at it.

test/release/examples/primers/opaque.chpl
mppf: See PR #7287.

test/release/examples/benchmarks/lcals/LCALSMain.chpl
mppf: It's not exactly only compile-time or run-time, because it's a choice between compile-time failure for the code that doesn't handle errors vs. run-time error checking. I think it *is* different runtime behavior - checking for errors and halting if any are found. That is different run-time behavior than you'd get with marking a function 'throws' say (which is probably the most common way to consume the errors).

mppf: Other potential flag names: `--lazy-errors` or `--prototype-errors`

bradcray: Of the ones we've discussed, ``--lazy-errors`` is most appealing to me in its conciseness and the broad net it casts.  Though it's not as concise, I also find ``--enable-lazy-errors`` or ``--permit-lazy-errors`` to be slightly clearer.  Might be time to get some opinions other than mine on this question.

Apart from name, is the feature easy to implement?  (I'm guessing so).  Is it correct to characterize that it would not change the behavior of anything that is making an effort to handle errors properly and successfully does so?

mppf: Yes - easy to implement and also wouldn't change the behavior of anything making an effort to handle errors properly. And I think I'd implement it to just change the default for modules from "error mode relaxed" to "error mode fatal".

bradcray: That sounds right to me.  You're likely already on this, but I'd implement it, and then send out a mail or issue explaining it and containing a short poll with favorite names to see if there's a clear winner.

Issue #7246, opened 2017-09-08
LLVM 5.0.0 issues


dmk42:
This issue tracks the current items to be addressed with LLVM 5 as a Chapel target.

- [x]  LLVM 5 currently only works with `CHPL_LLVM_DEVELOPER` set due to the following problem.
Chapel with LLVM 5 gets an undefined symbol `llvm::Value::dump()` in the link step because the new version of the LLVM file `lib/IR/AsmWriter.cpp` wraps the implementation of that method with the following.
```c
#if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
```
I have not yet found a way to get cmake to pass `LLVM_ENABLE_DUMP` through to the LLVM build.  Currently a debug version of LLVM must be built so that `NDEBUG` will not be defined and `llvm::Value::dump()` will be compiled in.

- [ ]  When `--llvm --fast` is used, Chapel with LLVM 5 fails some fft tests with anomalous numeric results.  This may be due to aggressive optimization.  The tests pass without `--fast`.  The tests that fail with optimization turned on are as follows.
* `release/examples/benchmarks/hpcc/fft`
* `chplvis/benchmarks-hpcc/fft-vdb`
* `performance/compiler/bradc/fft-timecomp` (program times out)
* `studies/hpcc/FFT/fft-testPow4` (program times out)
* `studies/hpcc/FFT/diten/fft`

mppf:
Re. the aggressive optimization, I wonder if these settings are too much:
https://github.com/chapel-lang/chapel/blob/master/compiler/util/clangUtil.cpp#L1204
But for the first fft test, it passes `--ieee-float`. Perhaps we are not properly respecting this flag?

mppf:
- [ ] with commit updating master to use LLVM 5, be sure to update third-party/llvm/README

dmk42:
We have avoided this issue by skipping LLVM 5 and going straight to LLVM 6.  I'm closing this now.

Issue #6993, opened 2017-08-11
LLVM progression plans


dmk42:
As worked out in the discussion of #6962 , the following are planned for LLVM.

- [ ] Between releases 1.16 and 1.17, switch to LLVM-is-the-default.
- [x] After LLVM 5.0 is working with Chapel (but not necessarily included), remove support for LLVM < 4.  The purpose is to keep the source as clean as possible consistent with leaving enough history there to make future edits easier.
- [x] Remove autotools-build support for LLVM at one of the times above, as LLVM >= 4 will only build with cmake.
- [ ] When we upgrade to LLVM 5.0, switch to an llvm-src strategy instead of tarballs.

mppf:
@dmk42 - I think we should close this issue (we have other ways to track LLVM is default).

Issue #6231, opened 2017-05-12
Codegen to LLVM gets internal compiler error on casted extern constant


dmk42:
### Summary of Problem

When a C header file defines a macro as a casted integer constant, and the macro is used by a Chapel program as an `extern const`, the Chapel program compiles and runs when using the C back end, but the Chapel compiler's codegen gets an internal error when using the LLVM back end.

This problem was originally encountered when trying to compile lcals/Timer.chpl on Ubuntu, where the `CLOCKS_PER_SEC` macro is defined as a casted integer constant.  The source code below has been simplified so Ubuntu is no longer necessary to reproduce the problem.

The following error is observed when attempting to use the LLVM back end.  This message is generated with `CHPL_DEVELOPER=true` for extra detail.
```
cps.chpl:5: internal error: Could not code generate MY_CLOCKS_PER_SEC - perhaps it is a complex macro? [symbol.cpp:517]
Note: This source location is a guess.
```

### Steps to Reproduce

This problem has been reproduced under the following conditions.

* Either clang or gcc as the host/target compiler
* WIth or without `CHPL_DEVELOPER=true`
* `CHPL_LLVM=llvm`
* On either Linux or a Mac

**Source Code:**

cps.h:
```c
#define MY_CLOCKS_PER_SEC ((long) 1000000)
```
cps.chpl:
```chapel
require "cps.h";

extern const MY_CLOCKS_PER_SEC: int(64);

writeln(MY_CLOCKS_PER_SEC);
```

**Compile command:**

Working:
```
chpl cps.chpl
```
Nonworking:
```
chpl --llvm cps.chpl
```

**Execution command:**

Working:
```
a.out
```
Nonworking:
None available because the compiler aborts.

**Associated Future Test(s):**
N/A

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.16.0 pre-release (0aab9c3)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
[...]

PR #7248, opened 2017-09-08
Preliminary LLVM 5 port (not as the default LLVM)

dmk42:
This change ports Chapel to LLVM 5.  Two issues remain to be addressed, as documented in #7246 .
* LLVM 5 requires `CHPL_LLVM_DEVELOPER` to be set.
* Using `--llvm --fast` causes `release/examples/benchmerks/hpcc/fft` to fail with anomalous numeric results, along with some other fft tests.

Submitting this change now allows setting aside the remaining issues until after the Chapel 1.16 release, and also allows others to try LLVM 5 without having to duplicate my work.

I expect the items in #6993 to be addressed after the Chapel 1.16 release.

The primary changes to the LLVM API are:
* LLVM has changed more of its API to use named variadic parameters, which allows a function with a variable number of arguments to handle the exact number of arguments safely without requiring a null-terminated argument list.  Therefore, `NULL` as an argument must now be deleted from several method calls.
* The BBVectorize pass has been fully deleted in favor of SLP vectorization.
* Some methods now take an extra argument to specify an address space into which allocation should occur.
* `SynchronizationScope` has been renamed `SyncScope::ID`.
* Call site attributes are now stored in an `AttributeList` instead of an `AttributeSet`.

Testing (all done with `CHPL_LLVM_DEVELOPER` set due to the issue noted above):
- [x] Passes `release/examples` and `llvm` on a Mac without `--fast`
- [x] Passes `release/examples` and `llvm` on a Mac with `--fast` except for `fft` as noted above.
- [x] Passes the same tests on Ubuntu except for the casting issue noted in #6231
- [x] Full local linux64 testing without `--fast`
- [x] Full local linux64 testing with `--fast` (failing `fft` as noted above)
- [x] LLVM 4 verified to still work after this change (full local linux64 testing with and without `--fast`, and Mac testing on `release/examples` and `llvm`)


dmk42:
@mppf - Would you mind taking a look at this?

compiler/codegen/type.cpp
```diff
@@ -350,8 +350,14 @@ void AggregateType::codegenDef() {
...
-        if( largestType ) st = llvm::StructType::get(largestType, NULL);
-        else st = llvm::StructType::get(info->module->getContext());
+        if( largestType ) {
+          st = llvm::StructType::get(largestType
+#if HAVE_LLVM_VER < 50
+                                     , NULL
+#endif
+                                     );
+        } else
+          st = llvm::StructType::get(info->module->getContext());
```
mppf: This is a trivial/style thing, but I always use curly braces on an else block if I used them on an if block...

compiler/include/llvmUtil.h
```diff
@@ -26,30 +26,7 @@
...
-#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=4 )
-#define HAVE_LLVM_VER 34
-#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=3 )
-#define HAVE_LLVM_VER 33
-#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=2 )
-#define HAVE_LLVM_VER 32
-#elif LLVM_VERSION_MAJOR>3 || (LLVM_VERSION_MAJOR==3 && LLVM_VERSION_MINOR>=1 )
-#define HAVE_LLVM_VER 31
-#endif
+#define HAVE_LLVM_VER (LLVM_VERSION_MAJOR*10 + LLVM_VERSION_MINOR)
```
mppf: 👍

compiler/util/llvmDebug.cpp
```diff
@@ -34,10 +34,15 @@
...
-#include "llvm/Support/Dwarf.h"
 #include "llvm/Support/raw_os_ostream.h"
 #include <iostream>
 #include "llvm/ADT/DenseMap.h"
+
+#if HAVE_LLVM_VER >= 50
+#include "llvm/BinaryFormat/Dwarf.h"
+#else
+#include "llvm/Support/Dwarf.h"
+#endif
```
mppf: Shouldn't this be covered by the include of `llvmDebug.h` above?

compiler/codegen/type.cpp
dmk42: Sounds reasonable.  I will push a fix to make the curly braces the same for the else block.

compiler/util/llvmDebug.cpp
dmk42: Good point.  llvmDebug.cpp includes several files that are already included by llvmDebug.h.  I will push a change to delete the duplicates from llvmDebug.cpp so we don't have to port the files to new versions of LLVM twice.

dmk42:
Style fixes tested with the hellos and `test/llvm`.

Issue #7231, opened 2017-09-07
Directly inlined iterators modify local tmp instead of arg


ronawho:
### Summary of Problem
Updates to arguments passed to an iterator aren't preserved by direct iterator
inlining (--inline-iterators). e.g.

```chpl
iter myiter(in arg) {
  for i in 1..10 {
    arg += 1;
    yield arg;
  }
}

for i in myiter(1) do writeln(i);
```

will always yield `1` instead of `1, 2, 3, ...`

Note that this is a direct iterator inlining only bug. The test works if
inlining is done using the zippered iterator inlining code
(--no-inline-iterators --optimize-loop-iterators)

I _think_ this is likely a bug in replaceIteratorFormalsWithIteratorFields(),
but I'm not positive.

**Associated Future Test(s):**
[`test/functions/iterators/elliot/yieldArgIter/yieldArgIter.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/iterators/elliot/yieldArgIter/yieldArgIter.chpl) #7232

ronawho:
I have a branch that attempts to inline all non-zippered iterators, not just those with only 1 yield, and ~12 tests break because of this bug.

Related to https://github.com/chapel-lang/chapel/issues/6757

ronawho:
@daviditen Do you have any time to look at this?

daviditen:
I looked at older versions to see if this has ever worked and 1.14.0 works.  I did a binary search to see when it broke and PR #5624 on March 16 is when it stopped working.  With that hint I'll continue looking.

daviditen:
This change to [functionResolution.cpp:5585](https://github.com/chapel-lang/chapel/pull/5624/files#diff-e19fcde76769a1d3f15597da1da08352L5585) is where this test goes wrong.  Changing that line back makes the test pass.

The commit message for that [commit](https://github.com/chapel-lang/chapel/pull/5624/commits/eb60654d2e302cc8bdd4ba8560b430cf39ad70f6) states that the removed condition was apparently the reverse of what it should be.  I think this is correct, and the reverse should be added back in, along with another case for `fn->hasFlag(FLAG_ITERATOR_FN)` since we typically try to inline iterators.

I have this [change](https://github.com/chapel-lang/chapel/compare/master...daviditen:inlineIteratorBug?expand=1) in testing right now.

daviditen:
[Test Summary - 170908.165338]
[...]

PR #7253, opened 2017-09-08
Fix a bug for inlined iterators with formals with in intent

daviditen:
PR #5624 changed `formalRequiresTemp()` to no longer depend on if the function
is inlined, stating that it seems the condition was the reverse of what it
should have been anyway. I believe that this is correct and the reverse
should be added back.

In addition, an `in` formal to an iterator should get a temp because we want
to inline iterators even though they don't have `FLAG_INLINE` on them.

This fixes issue #7231.


ronawho:
Thanks for taking this on @daviditen!

I don't feel particularly qualified to review this, but @benharsh and/or @mppf would probably be good reviewers

daviditen:
Thanks @mppf, I'll note that in the commit message.

PR #3371, opened 2016-02-25
Allow c_ptr's to be implicitly coerced to c_void_ptr

ronawho:
This is to support using the chpl_here_* routines instead of externing to the
chpl_mem_* functions in the string module. Currently the string module defines
the chpl_mem_* routines as taking/return a c_ptr instead of a c_void_ptr,
because we didn't coerce c_ptr's to c_void_ptr's.

This should also allow us to get rid of a lot of code in CPtr module, but
that's cleanup for another day

PR #7254, opened 2017-09-09
Make c_ptr and c_void_ptr objects printable

### LINKED ISSUE #7097: I/O formatting for c_ptr and c_void_ptr
### Feature Request

When dealing with C interoperability, I often find it helpful to print out pointers as hexadecimal values. In C, I can use the `%p` format specifier, but Chapel does not have that feature. This issue requests the addition of an explicit `%p` format specifier (and equivalent `%t`) for formatted I/O on `c_ptr` and `c_void_ptr` types.

nspark:
Per the title, this PR makes `c_ptr` and `c_void_ptr` objects printable.

This PR would close #7097 to my satisfaction.

### DIFF
--- a/modules/internal/CPtr.chpl
+++ b/modules/internal/CPtr.chpl
@@ -353,7 +353,7 @@ module CPtr {
 
     :arg data: the c_ptr to memory that was allocated
     */
-  inline proc c_free(data: c_ptr) {
+  inline proc c_free(data: c_void_ptr) {
     chpl_here_free(data);
   }
 
--- a/modules/internal/CPtr.chpl
+++ b/modules/internal/CPtr.chpl
@@ -79,6 +79,18 @@ module CPtr {
     inline proc deref() ref {
       return __primitive("array_get", this, 0);
     }
+    /* Print this pointer */
+    inline proc writeThis(ch) {
+      (this:c_void_ptr).writeThis(ch);
+    }
+  }
+
+  pragma "no doc"
+  inline proc c_void_ptr.writeThis(ch) {
+    if this == c_nil then
+      ch.writef("(nil)");
+    else
+      ch.writef("0x%xu", this:uint(64));
   }
 
   pragma "no doc"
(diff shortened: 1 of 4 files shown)


modules/internal/CPtr.chpl
```diff
@@ -353,7 +372,7 @@ module CPtr {
 
     :arg data: the c_ptr to memory that was allocated
     */
-  inline proc c_free(data: c_ptr) {
+  inline proc c_free(data: c_void_ptr) {
```
mppf: I think we still need the old c_free and that you should copy the function to make a version that applies to c_void_ptr. (or make it generic with `where isAnyCPtr(t)`). As far as I know we don't have any implicit conversions from c_ptr to c_void_ptr.

nspark: Sure. Still, [this line of my test](https://github.com/chapel-lang/chapel/pull/7254/files/6dbcf375317047aa5165187ee1b81a3824add12a#diff-50dd7781e51e1b51a1c34b0fef8ddb8fR21) works and made it seem like `c_ptr` gets implicitly converted to `c_void_ptr`. Plus, [`c_free` is a pass-through call to `chpl_here_free`](https://github.com/chapel-lang/chapel/blob/master/modules/internal/CPtr.chpl#L356) (which accepts `c_void_ptr`), so the conversion seems to happen implicitly.

I like the `isAnyCPtr(t)` approach.

mppf: I stand corrected - Elliot added implicit conversions from c_ptr to c_void_ptr in PR #3371. So, I don't think you need to add the 2nd method. Probably best to leave it alone (unless you really want the isAnyCPtr approach).

mppf: I suppose that from a documentation point of view, the where clause is clearer than just accepting a c_void_ptr. If we leave it as-is, we probably should add a doc-comment that indicates the method will accept c_ptr in addition to c_void_ptr.

mppf:
Passed full local testing, except for an error-handling issue. I'm going to merge and then fix the issue in a separate PR.

nspark:
Thanks @mppf!

PR #7257, opened 2017-09-11
Improve try compiler error

mppf:
Improve the compiler error messages that were previously `throwing call in non-throwing function without try or try!` to include more information and to remove unnecessary spaces.

- [x] full local testing

Reviewed by @lydia-duncan and @psahabu - thanks!

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -537,7 +537,7 @@ static void printReason(BaseAST* node, implicitThrowsReasons_t* reasons)
       if (calledFn->throwsError()) {
         if (reasons->count(calledFn)) {
           BaseAST* reason = (*reasons)[calledFn];
-          USR_PRINT(reason, " is reason function throws");
+          USR_PRINT(reason, "is reason function throws");
           printReason(reason, reasons);
         }
       }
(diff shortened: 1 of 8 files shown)


compiler/passes/errorHandling.cpp
```diff
@@ -537,7 +537,7 @@ static void printReason(BaseAST* node, implicitThrowsReasons_t* reasons)
       if (calledFn->throwsError()) {
         if (reasons->count(calledFn)) {
           BaseAST* reason = (*reasons)[calledFn];
-          USR_PRINT(reason, " is reason function throws");
+          USR_PRINT(reason, "is reason function throws");
```
lydia-duncan: I might extend this somewhat so that the message reads: "note: this line is the reason the function throws" or something along those lines.  It was a bit too curt for my taste.

mppf: how about `reason function throws here` ? Or `called throwing function here` ?

lydia-duncan: Of those two options, I would prefer `called throwing function here`.  My problem with the other one and the old version is that the message seems a little passive and vague.  `called throwing function here` seems clearer to me

mppf: I'm currently proceeding with `call to throwing function here`. thanks!

compiler/passes/errorHandling.cpp
```diff
@@ -799,10 +799,18 @@ bool ErrorCheckingVisitor::enterCallExpr(CallExpr* node) {
             inThrowingFunction = parentFn->throwsError();
           }
           if (mode == ERROR_MODE_STRICT) {
-            USR_FATAL_CONT(node, "throwing call without try or try! (strict mode)");
+            USR_FATAL_CONT(node, "call to throwing function %s "
+                                 "without try or try! (strict mode)",
+                                 calledFn->name);
+            USR_PRINT(calledFn, "throwing function %s defined here",
```
psahabu: Some ideas for this second line, though I'm not particularly invested:
- "throwing function" is redundant because it's already called a throwing function in the `FATAL_CONT`
- indent to associate it with the above line

```
implicit-relaxed-try-nothrows-inner.chpl:6: error: call to throwing function throwAnError without throws, try, or try! (relaxed mode)
  ./ThrowError.chpl:1: note: throwAnError defined here
```

psahabu: @lydia-duncan: Feel free to also reject these suggestions if you think they're off-base.

mppf: I think it'd be new & different to indent the 2nd line (AFAIK no other compiler error/warning does this). I think it's reasonable to repeat 'throwing function' in this case but I'd be OK removing it from the 'defined here' line if you & Lydia both view it as an improvement.

psahabu: To the 2nd line: ah okay, wasn't sure of what was typical.

To repeating throwing function: that's reasonable too, ignore me.

lydia-duncan: I think this error message is fine as is.  I'm okay with repeating that the function is throwing, to ensure the connection between the two lines is obvious, but I don't feel strongly enough to object if it was removed.

lydia-duncan: I agree the indentation looks a little odd, but I think it would be unnecessarily complicated to change

PR #7263, opened 2017-09-11
Use global domain in ISx

benharsh:
Use a global domain instead of a local per-array domain now that serialization eliminates the overhead at scale.


bradcray:
Looks good, thanks!

PR #7264, opened 2017-09-11
Do not document serialization-related variables or procs

benharsh:


modules/internal/String.chpl
```diff
@@ -83,11 +83,13 @@ module String {
 
   type bufferType = c_ptr(uint(8));
```
ben-albrecht: Since we're leaving this publicly documented, would it be worth adding a small doc-string for it?

benharsh: @noakesmichael : your name is on the commit that brought this type out of an undocumented private module. Do you think it should be documented now?

noakesmichael: Kyle placed it in a private module to work around a scope/resolution bug that he
couldn't diagnose at the time.  I am not sure what the bug was and hence how it
got fixed but I noticed that it was fixed and so I lifted it out.

I don't have a strong opinion on whether it should be documented but my leaning
would be to regard it as an internal implementation detail, likely based on the
encoding, and so not document it.

PR #7265, opened 2017-09-11
Added .mason to .gitignore

vasslitvinov:
... given that it's in CLEANFILES


vasslitvinov:

PR #7267, opened 2017-09-11
Add bug for use-after-free for heap var and recursive iterators

benharsh:


benharsh:
@vasslitvinov : here's a small reproducer for the same bug that was worked-around in #7252 .

I suspect the actual fix should be implemented in parallel's freeHeapAllocatedVars.

Issue #6757, opened 2017-07-19
yield stmt in else part affects time.


TomsyPaul:
### Summary of Problem
The yield statement in else part of an if-else is also adding to the execution time.

### Steps to Reproduce

**Source Code:**

```chapel
use Time;
const n=10000;
var A:[1..n,1..n] int, t = new Timer();
//A.thankGod;
iter myiter()
{
  var x:bool=true;
  if x then
    for (i,j) in A.domain do yield (i,j);
  else
    for (i,j) in A.domain do yield (i,j);
}

iter youriter()
{
  if true then
    for (i,j) in A.domain do yield (i,j);
  else
    for (i,j) in A.domain do yield (i,j);
}

t.start();
for (x,y) in myiter() do A[x,y]=1;
t.stop();
writeln("myiter: ", t.elapsed(), " seconds");

t.clear();
t.start();
for (x,y) in youriter() do A[x,y]=1;
t.stop();
writeln("youriter: ", t.elapsed(), " seconds");

```

**Compile command:**
chpl iftrueiter.chpl

**Execution command:**
./a.out

**Output**
myiter: 1.00722 seconds
youriter: 0.193151 seconds

### Configuration Information

- Output of `chpl --version`: 
chpl Version 1.14.0.953f818
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: 
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
gcc (GCC) 4.8.0 20130311 (GNU UPC 4.8.0-3)

ronawho:
[...]

PR #7270, opened 2017-09-12
Allow direct inlining of iterators that have up to 10 yields

ronawho:
Previously we only inlined non-zippered iterators that had exactly 1 yield
statement. This limit was to prevent massive code growth since the body will be
inlined at every yield site. However, given the huge performance penalty paid
for not inlining an iterator, it's worthwhile to pay for some code growth.

Use a default limit of 10, that can be set with --inline-iterators-yield-limit.

Additionally, this adds --report-inlined-iterators, which is similar to
--report-optimized-loop-iterators but for direct iterators only. Use this flag
to lock in the yield limit and show when we will and will not inline iterators.

Closes https://github.com/chapel-lang/chapel/issues/6757


ronawho:
Enabled by https://github.com/chapel-lang/chapel/pull/7253

@daviditen mind reviewing? Probably easiest to review commit at a time.

compiler/resolution/lowerIterators.cpp
```diff
@@ -1246,6 +1246,17 @@ expandIteratorInline(ForLoop* forLoop) {
   Symbol*   ic       = forLoop->iteratorGet()->symbol();
   FnSymbol* iterator = getTheIteratorFn(ic);
 
+  if (fReportInlinedIterators) {
+    ModuleSymbol *mod = toModuleSymbol(iterator->getModule());
+    INT_ASSERT(mod);
+
+    if (developer || mod->modTag == MOD_USER) {
+      printf("Inlined iterator (%s) in module %s (%s:%d)\n", iterator->cname,
```
daviditen: I think the location where the iterator is getting inlined e.g. `forLoop` is as important as the location where the iterator is defined.

daviditen:
Overall looks good.

Have you done any checking on the compiler performance impact of this?  We may need to tune the limit if 10 proves to cause too much generated code bloat.

test/functions/iterators/elliot/report/inline-iters-report.default.good
```diff
@@ -0,0 +1,58 @@
+Inlined iterator (tenYieldIter) in module inline-iters-report (inline-iters-report.chpl:10)
```
daviditen: Is the order of these `--report-inlined-iterators` messages guaranteed, or do you need something like a `.prediff` to sort them?

compiler/AST/iterator.cpp
```diff
@@ -1501,7 +1501,7 @@ void lowerIterator(FnSymbol* fn) {
     // Isn't that bad?
     if (singleLoop) {
       if (fReportOptimizedLoopIterators) {
-        ModuleSymbol *mod = toModuleSymbol(fn->defPoint->parentSymbol);
+        ModuleSymbol *mod = toModuleSymbol(fn->getModule());
```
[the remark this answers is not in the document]
ronawho: Good catch. Will remove this along with the assert(mod)

compiler/resolution/lowerIterators.cpp
```diff
@@ -1758,13 +1769,14 @@ countYieldsInFn(FnSymbol* fn)
...
+// Iterators can be inlined if they contains between 1 and
+// inline_iter_yield_limit yield statements
 static bool
 canInlineIterator(FnSymbol* iterator) {
-  unsigned count = countYieldsInFn(iterator);
+  int count = countYieldsInFn(iterator);
 
   // count==0 e.g. in users/biesack/test_recursive_iterator.chpl
-  return (count == 1) ? true : false;
+  return (count >= 1 && count <= inline_iter_yield_limit) ? true : false;
```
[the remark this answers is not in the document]
ronawho: will do

compiler/resolution/lowerIterators.cpp
ronawho: I don't really disagree, but for now I want to keep the output consistent with `--report-optimized-loop-iterators`.

test/functions/iterators/elliot/report/inline-iters-report.default.good
ronawho: I was wondering the same thing. `functions/iterators/sungeun/report/` (which tests `--report-optimized-loop-iterators`) does not sort and it has been stable for the past 4 years, so I think the ordering is relatively safe. Note that the `--inline-iterators` and `--optimize-loop-iterators` functions are both called from `expandForLoop()`, so this is no less stable than that test.

ronawho:
>Have you done any checking on the compiler performance impact of this? We may need to tune the limit if 10 proves to cause too much generated code bloat. 

I did not see any compiler performance impact, but we don't have many tests for iters with multiple yields. And our core range/domain/array iterators have all been optimized to only have a single yield, so we wouldn't see an impact from our core modules.

I agree that we may have to tune in the future, but at least for now a max of 10x growth for an iterator doesn't seem too horrendous given how large our generated code is anyways.

daviditen:
OK, that works for me.

PR #7272, opened 2017-09-12
Add compilation errors for when initializers involve error handling

lydia-duncan:
While we would like to be able to allow throw, try, and try! inside an
initializer during Phase 2 only, we didn't have time to implement that this
release.  Instead, add a temporary error about using those structures (with
the except of try! without a catch block, as it will cause an execution halt
and thus will not return the instance in a bad state), and declaring an
initializer as throws.

Adds 9 tests to verify the new error messages, covering:
- declaring the initializer as throws
- having a throw in an initializer body (currently just checks Phase 1)
- having a `try stmt` in an initializer body (ditto above)
- having a `try { stmts }` in an initializer body (ditto above)
- having a `try! stmt` in an initializer body (ditto above)
- having a `try! { stmts }` in an initializer body (ditto above)
- having a `try { stmts } catch { stmts }` in an initializer body (ditto above)
- having a `try! { stmts } catch { stmts }` in an initializer body (ditto above)
- having one of the above in an initializer body when it is not a top level
  statement (ditto above)

Testing: running a full paratest w/ futures

### DIFF
--- a/test/classes/initializers/errors/errHandling/initWithTryBang.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryBang.chpl
@@ -1,0 +1,22 @@
+// Verifies we currently don't allow initializers to have try! statements in
+// their body.
+// We hope to eventually allow it.
+class Foo {
+  var x: int;
+
+  proc init() {
+    x = 10;
+    try! {
+      outerFunc();
+    }
+    super.init();
+  }
+}
+
+proc outerFunc() throws {
+  throw new Error();
+}
+
+var foo = new Foo();
+writeln(foo);
+delete foo;
--- a/test/classes/initializers/errors/errHandling/initWithTryBang.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryBang.chpl
@@ -1,6 +1,5 @@
-// Verifies we currently don't allow initializers to have try! statements in
-// their body.
-// We hope to eventually allow it.
+// Verifies we currently allow initializers to have try! statements in their
+// body, as long as it doesn't have a catch block
 class Foo {
   var x: int;
 
--- a/test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
@@ -1,6 +1,5 @@
-// Verifies we currently don't allow initializers to have try! statements in
-// their body.
-// We hope to eventually allow it.
+// Verifies we currently allow initializers to have try! statements in their
+// body, as long as it doesn't have a catch block.
 class Foo {
   var x: int;
 
--- a/test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
@@ -1,0 +1,20 @@
+// Verifies we currently don't allow initializers to have try! statements in
+// their body.
+// We hope to eventually allow it.
+class Foo {
+  var x: int;
+
+  proc init() {
+    x = 10;
+    try! outerFunc();
+    super.init();
+  }
+}
+
+proc outerFunc() throws {
+  throw new Error();
+}
+
+var foo = new Foo();
+writeln(foo);
+delete foo;
--- a/test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
@@ -1,0 +1,24 @@
+// Verifies we currently don't allow initializers to have try statements in
+// their body.
+// We hope to eventually allow it.
+class Foo {
+  var x: int;
+
+  proc init() {
+    x = 10;
+    try {
+      outerFunc();
+    } catch {
+      writeln("Look ma, I caught an error!");
+    }
+    super.init();
+  }
+}
+
+proc outerFunc() throws {
+  throw new Error();
+}
+
+var foo = new Foo();
+writeln(foo);
+delete foo;
--- a/test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
+++ b/test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
@@ -5,9 +5,9 @@ class Foo {
   var x: int;
 
   proc init() {
-    x = 10;
     try {
       outerFunc();
+      x = 10;
     } catch {
       writeln("Look ma, I caught an error!");
     }
--- a/compiler/passes/initializerRules.cpp
+++ b/compiler/passes/initializerRules.cpp
@@ -24,6 +24,7 @@
 #include "InitNormalize.h"
 #include "passes.h"
 #include "stmt.h"
+#include "TryStmt.h"
 #include "type.h"
 #include "typeSpecifier.h"
 
@@ -38,7 +39,7 @@ static bool      isInitStmt (Expr* stmt);
 static bool      isSuperInit(Expr* stmt);
 static bool      isThisInit (Expr* stmt);
 
-static bool      isErrHandlingStmt(Expr* stmt);
+static bool      isUnacceptableTry(Expr* stmt);
 
 static void      preNormalize(FnSymbol* fn);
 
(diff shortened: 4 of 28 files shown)


lydia-duncan:
@noakesmichael - would you review?  @psahabu or @mppf - would one of you double check me on the testing coverage?

psahabu:
The problem with banning `try!` in initializers is that it can cause programs written in fatal error mode to be unwritable in relaxed or strict error mode.
``` chapel
proc init() {
  thisFunctionThrows();
}
```
In this example, fatal mode would wrap a halt around `thisFunctionThrows();`, but in relaxed mode would require a `try!` that is disallowed by this PR.

There are four possible resolutions to this, in order of effort:
1. Leave it as is.
2. Permit `try!` and only `try!` inside initializers.
3. Permit `try!` and `try` with a `catch`-all inside initializers.
4. Check all functions calls in `init()` to see if they throw.

While (1) seems unappealing at first, relaxed mode already bans some reasonable programs from being written, so this would just be a harsher enforcement than intended.

(2) is the simplest to tack on to this PR, but would leave strange inconsistencies like `try! { } catch { }` would be allowed while `try { } catch { }` would not, even though they're functionally the same.

(3) is more philosophically consistent with error handling generally, but would a bit more effort determining if there is a catch-all on a plain `try`. It might be possible to leverage code in the errorHandling pass to do this, though.

(4) is needlessly limiting and infeasible in this timeframe.

@mppf do you have any thoughts on this?

mppf:
I think that `try!` (without catch) should always be allowed, even in Phase 1. If the initializer can't be `throws`, there's no way to propagate the error out when `try` (without catch) is used, so there's no sense in checking it.

I can certainly see that we don't want control flow in Phase 1, but that seems similar for conditionals, `try/catch`, and even `defer`. Is there a way to make the check for isErrHandlingStmt more similar to how conditionals are detected? Or, what would the testing look like if we removed the `isErrHandlingStmt` entirely? Or, perhaps, only disallow `try/catch` and `try!/catch` in Phase 1 for now?

lydia-duncan:
Again, this is a temporary commit because I don't think we have time to fully implement the "error handling in initializers" rules we wanted for this release.  The question is whether banning `try!` for this release is reasonable temporary behavior.  Removing isErrHandlingStmt would mean going out with whatever support happened to work as is, which definitely allows us to throw from Phase 1 currently.

While I don't think it would be necessarily that much harder to implement the more fine-grained approach for initializers (though some parts would require interaction between the error handling and initializer-verifying passes that will be more difficult), it seems late in the release cycle to do so and be certain we weren't allowing something that we would ultimately want to clamp down on later.

mppf:
@lydia-duncan - For the immediate term, I'd be more comfortable with "whatever happens to work" than with "banning try!". I think most uses of error handling + initializers will want the initializer to `throw` or will amount to simple uses of `try!` - and the former won't compile anyway (with your other check) and the latter will probably just work.

lydia-duncan:
Ah, I see.  So only banning throw on an initializer would allow try! and try with a catchall, but would give compilation errors for everything else.

test/classes/initializers/errors/errHandling/initWithTryBang.chpl
```diff
@@ -0,0 +1,22 @@
...
+    }
+    super.init();
+  }
+}
+
+proc outerFunc() throws {
+  throw new Error();
+}
+
+var foo = new Foo();
```
lydia-duncan: @mppf - without the changes on this branch, this test generates the output:
```
uncaught Error:
   initWithTryBang.chpl:17: thrown here
   initWithTryBang.chpl:20: uncaught here
```

test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
```diff
@@ -0,0 +1,20 @@
...
+    try! outerFunc();
+    super.init();
+  }
+}
+
+proc outerFunc() throws {
+  throw new Error();
+}
+
+var foo = new Foo();
```
lydia-duncan: Similarly, this test generates this output:
```
uncaught Error:
   initWithTryBang2.chpl:15: thrown here
   initWithTryBang2.chpl:18: uncaught here
```

test/classes/initializers/errors/errHandling/initWithTryBang.chpl
mppf: That seems just fine.

test/classes/initializers/errors/errHandling/initWithTryBang2.chpl
mppf: Also looks OK to me.

test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
```diff
@@ -0,0 +1,24 @@
...
+class Foo {
+  var x: int;
+
+  proc init() {
+    x = 10;
+    try {
+      outerFunc();
+    } catch {
+      writeln("Look ma, I caught an error!");
+    }
```
lydia-duncan: I think this case ends up being the most worrisome without banning all error handling structures, as its behavior seems to indicate it would permit field assignment and super.init() calls within the try catch, potentially violating single-initialization of fields and single-super.init() calls (for instance, an early thrown error could prevent the assignment to the field that our pass detected within the later body of the try, and would not insert the appropriate omitted initialization before returning)

test/classes/initializers/errors/errHandling/initWithTryBang.chpl
lydia-duncan: Shouldn't the line number be within the 9-11 range, since it is the `try!` that didn't catch the error rather than the `var foo = new Foo();`?

mppf: Yes, the second line number ought to be 9-11 and not 20. I'm not sure what's going wrong there, but would guess it's simply some mis-attributed line numbers on AST elements in the compiler.

test/classes/initializers/errors/errHandling/initWithTryCatch.chpl
mppf: I don't think you've said yet what this actually does without the "no error handling structures" rule?

lydia-duncan: Without the changes on this branch, it gives the output:
```
Look ma, I caught an error!
{x = 10}
```

I suspect its behavior when I only ban the initializer from being declared as throws should be equivalent, since the catchall does not rethrow anything, so the remainder of the initializer would complete as normal

lydia-duncan: Going back to your earlier point, we could get around this by not allowing try/try! with catch in Phase 1, but allow try! only.  To me, that would imply that we should make the error message more fine-grained as the current one doesn't indicate it would allow some but not others (which is definitely acceptable to me)

mppf: Nothing about this test makes me think it must be an error... but anyway I feel my viewpoint has been heard & I hope you do something we've recently discussed.

lydia-duncan: Would changing the test to:
```
try {
  outerFunc();
  x = 10;
} catch { ... }
```
clarify my concerns?  I can do that as part of the change

mppf: Yes that is a much better test showing what should be an error.

test/classes/initializers/errors/errHandling/initWithTryBang.chpl
lydia-duncan: Should I accept the currently generated numbers for this commit, even though it is less than ideal?  Or should I fix that now (or in a separate PR)?

mppf: I certainly don't think you need to do it in this PR. I think it's worth looking in to at some point but it doesn't have to happen before Monday.

lydia-duncan:
This is ready for a second glance if you guys want to sanity check the new error messages

lydia-duncan:
(Except I forgot to update the sync .bad files, just ignore those)

test/classes/initializers/errors/errHandling/initWithTryBang.chpl
psahabu: If the line numbers are wrong for this test, shouldn't there be a `.bad` file for it? Or did I just miss that?

psahabu: If the line numbers are wrong for this test, shouldn't there be a `.future` and `.bad` file for it? Or did I just miss that?

lydia-duncan: I decided to make only one of the three a future, with a note to update the other two when it gets fixed

PR #7278, opened 2017-09-12
update spawn error good files

psahabu:
Forgot to update the good files on these tests. Thanks to @ronawho for catching this.


psahabu:
Turns out the good files for these tests depend on the platform they are tested on, with different systems generating different `SystemError`s. I'll be opening a prediff PR shortly.

psahabu:
After actually looking at the test directory, the test only runs on darwin, which is why linux64+flat didn't turn up anything. This fix should be sufficient to fix the failures on darwin.

Issue #7130, opened 2017-08-26
unhandled stderr.writeln() in default mode catch block


psahabu:
### Summary of Problem
Calling `stderr.writeln()` from a catch block in default mode causes incomprehensible errors to be emitted, rather than a properly inserted `halt()`. Explicitly wrapping the call in a `try!` preempts the problem.

### Steps to Reproduce

**Source Code:**

```chapel
proc unrelated() {
  try {
    writeln("inside try");
  } catch {
    stderr.writeln("stderr");
  }
}

unrelated();
```

**Compile command:**
`chpl unrelated.chpl`

**Execution command:**
`./unrelated`

### Configuration Information

- Output of `chpl --version`: 1.16 prerelease
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
  - CHPL_TARGET_PLATFORM: darwin
  - CHPL_TARGET_COMPILER: clang
  - CHPL_TARGET_ARCH: native
  - CHPL_LOCALE_MODEL: flat
  - CHPL_COMM: none
  - CHPL_TASKS: qthreads
  - CHPL_LAUNCHER: none
  - CHPL_TIMERS: generic
  - CHPL_UNWIND: none
  - CHPL_MEM: jemalloc
  - CHPL_MAKE: make
  - CHPL_ATOMICS: intrinsics
  - CHPL_GMP: gmp
  - CHPL_HWLOC: hwloc
  - CHPL_REGEXP: re2
  - CHPL_WIDE_POINTERS: struct
  - CHPL_AUX_FILESYS: none
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
  - Apple LLVM version 8.0.0 (clang-800.0.42.1)
  - Target: x86_64-apple-darwin15.6.0
  - Thread model: posix
  - InstalledDir: /Library/Developer/CommandLineTools/usr/bin

psahabu:
Ah, this is the band-aid fix in #6607 coming back to bite me, `errorHandling.cpp:337`:

``` cpp
Expr* insert = node->getStmtExpr();
if (insert == NULL)
  insert = node;
```

Here's the AST that is given to the `lowerErrorHandling` pass from `resolve`:

```
val _return_tmp_[1201912]:bool[7] "temp"
(1201914 'move' _return_tmp_[1201912](193214 call writeln[1074430] stderr[235915] "stderr"))
```

[...]

PR #7281, opened 2017-09-12
update labelprop with error handling

psahabu:
#7130 was preventing labelprop from being upgraded to error handling, but #7269 was a nice fix for that. This is the follow-up to that issue and PR, bringing labelprop into the error handling world.

- [ ] test passes on chapcs


psahabu:
Mysterious timeouts on chapcs, punting to post-release.

PR #7283, opened 2017-09-12
Use halt-on-error mode for LCALS

mppf:
To allow LCALS to run in release-over-release testing,
don't use try! and instead just remove the `module main` declaration
so we get halt-on-error mode.

Discussed with @daviditen.
Tested release/examples/benchmarks/lcals with and without -performance.
Verified LCALS still builds with make.

### DIFF
--- a/test/release/examples/benchmarks/lcals/LCALSMain.chpl
+++ b/test/release/examples/benchmarks/lcals/LCALSMain.chpl
@@ -1,1160 +1,1154 @@
-module main {
-  use FileSystem;
-  use LCALSParams;
-  use LCALSEnums;
-  use LCALSDataTypes;
-  use LCALSLoops;
-  use LCALSStatic;
-  use LCALSConfiguration;
-  use LCALSChecksums;
+use FileSystem;
+use LCALSParams;
+use LCALSEnums;
+use LCALSDataTypes;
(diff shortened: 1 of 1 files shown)


mppf:
@daviditen - could you look at this? Should we get LCALS to compile with generateTimingReport? If so, would it be acceptable to remove the outer `module main` ?

daviditen:
The old-style out error change looks fine.

Yes, I think `generateTimingReport` should be kept working - it's the output version that is intended to be identical to the reference's output.

If removing the `module main` helps with that it's fine with me to remove it.

mppf:
How do I activate generateTimingReport, for testing?

daviditen:
It's controlled by the `config param outputFormat`.  Setting it to OutputStyle.REFERENCE, or just leaving it as the default will use `generateTimingReport`.  Compiling with the Makefile leaves it alone, so that works too.

bradcray:
Belatedly, I'm realizing that LCALS still won't give us release-over-release numbers back very far due to the Barrier -> Barriers renaming.  Sorry to have introduced hassle, and if you guys like the try! approach better, feel free to back this out.  That said, I sortof like having the "entry point" module not be named, stylistically (so would stick with this refactoring anyway, personally).

PR #7285, opened 2017-09-12
Fix block rankchange bug

benharsh:
Testing:
- [x] full local
- [x] full gasnet

### DIFF
--- a/modules/dists/BlockDist.chpl
+++ b/modules/dists/BlockDist.chpl
@@ -1531,11 +1531,11 @@ proc BlockArr.doiBulkTransferFrom(Barg, viewDom)
   if debugBlockDistBulkTransfer then
     writeln("In BlockArr.doiBulkTransferFrom()");
 
-  if this.rank == Barg.rank {
-    const Dest = this;
-    const Src = chpl__getActualArray(Barg);
-    const srcView = chpl__getViewDom(Barg);
-    type el = Dest.idxType;
+  const Dest    = this;
+  const Src     = chpl__getActualArray(Barg);
+  const srcView = chpl__getViewDom(Barg);
+  type el       = Dest.idxType;
+  if Dest.rank == Src.rank {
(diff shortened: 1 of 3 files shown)


benharsh:
@bradcray : here's the quick fix for this particular bug. A more general solution is provided by the API changes described in #5310 .

modules/dists/BlockDist.chpl
```diff
@@ -1531,11 +1531,11 @@ proc BlockArr.doiBulkTransferFrom(Barg, viewDom)
...
-  if this.rank == Barg.rank {
-    const Dest = this;
-    const Src = chpl__getActualArray(Barg);
-    const srcView = chpl__getViewDom(Barg);
-    type el = Dest.idxType;
+  const Dest    = this;
+  const Src     = chpl__getActualArray(Barg);
+  const srcView = chpl__getViewDom(Barg);
+  type el       = Dest.idxType;
+  if Dest.rank == Src.rank {
```
bradcray: Does this conditional have the effect of simply saying "never try to do a bulk transfer on a rank-change slice?"

benharsh: Yes, and that's what we do already. The problem was that the rank check was done at runtime, leading the compiler to attempt to resolve the problematic code in the if-statement.

bradcray: OK, I'm good to merge then.  

To whet my appetite: Does the "more general solution" of the future attempt to do bulk transfers on rank change slices?

benharsh: The more general solution requires domain map authors to opt-in to bulk-transfers involving rank-changes by implementing a param method named ``doiCanBulkTransferRankChange`` that returns true. Though I generally want to reduce the number of bulk-transfer methods, rank-changes are sufficiently tricky that it seems worth the extra cruft.

Issue #7261, opened 2017-09-11
error handling: returning errors from writeThis


mppf:
The `writeThis`, `readWriteThis`, and `readThis` methods can be provided on a type to implement how that type is written or read.

As part of the error handling effort for Chapel 1.16, many I/O functions now are marked `throws`. Some of these might be called by the above methods like `writeThis`.

Currently, if there is an error performing I/O in support of `writeThis`/..., an error code is saved in the relevant channel. Then, subsequent attempts to do I/O on that channel as part of `writeThis` will instead do nothing and preserve the error.

This method of saving an error code in the channel doesn't fit with the `throws` style errors that we'd like to be able to use, however. In particular, one might imagine wanting to have a `writeThis`/... that can throw an arbitrary user error.

At the time of this writing, `channel.writef` or `channel.write` are marked `throws`, but the I/O operator `<~>` is not. Instead, the I/O operator always saves the "error" state in to the channel. For this reason, PR #7245 changes several calls to `write` to calls to `<~>` as a workaround for this issue. Longer term though, I think we want `<~>` to be equivalent to `channel.write`.

We could:
1. make writeThis/... throws, and save the error state in the channel record. This strategy has the advantage of allowing arbitrary exceptions to be thrown by writeThis/..
2. make writeThis/... optionally return an error code
3. expect writeThis/... authors to manage explicitly setting an error code on the channel if an error is encountered

BryantLam:
Just ran into this again (because I keep forgetting). Tried to do the equivalent of `try! f <~> data`.

BryantLam:
>     3. expect writeThis/... authors to manage explicitly setting an error code on the channel if an error is encountered

This is the status quo today and I haven't internalized the implications until now. Using the `<~>` operator seems somewhat problematic in that if the channel ever hits an error, all subsequent writes to it silently fail. (RIght?..) While I've not seen stdout/stderr fail before, I won't rule it out in an e.g., long-running job that suddenly stops writing to stdout/stderr/some log file vs. throwing an exception which has to be handled (but maybe the application is safe even if these writes silently fail?..).

mppf:
> Using the <~> operator seems somewhat problematic in that if the channel ever hits an error, all subsequent writes to it silently fail

It's possible to clear the error. Also I'm not sure I'd call it "silently failing". It's just within a `write` call that the error is hidden, and I/O calls ignored, until the `write` call retrieves the saved error and then `throws`.

[...]

PR #7293, opened 2017-09-13
Fix writeThis in network atomics module

ronawho:
Continuation/bug-fix of #7245. That PR adjusted several `writeThis` methods to
use `<~>` instead of `channel.write` since the operator saves an error code in
the channel instead of throwing. This was a workaround for the problem
discussed in issue #7261.

This makes the same change for network atomics, which was missed in that PR.

### DIFF
--- a/modules/internal/NetworkAtomics.chpl
+++ b/modules/internal/NetworkAtomics.chpl
@@ -183,7 +183,7 @@ module NetworkAtomics {
     }
 
     proc const writeThis(x) {
-      x.write(read());
+      x <~> read();
     }
   }
 
@@ -374,7 +374,7 @@ module NetworkAtomics {
     }
 
     proc const writeThis(x) {
-      x.write(read());
+      x <~> read();
     }
   }
 
@@ -565,7 +565,7 @@ module NetworkAtomics {
     }
 
     proc const writeThis(x) {
-      x.write(read());
+      x <~> read();
     }
   }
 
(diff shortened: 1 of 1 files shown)


ronawho:
Passes release/examples for ugni as well as `--performance --perflabel ml-` for ugni

ronawho:
@mppf I plan to merge this tonight to avoid testing failures, but I would appreciate a post-commit review.

Might also be worth taking a look at all the other `writeThis` methods in the module code to ensure there's no other infrequently tested cases being missed.

modules/internal/NetworkAtomics.chpl
```diff
@@ -755,8 +755,8 @@ module NetworkAtomics {
       _v = value;
     }
 
-    proc writeThis(const x) {
```
ronawho: Note that I changed the signature of this `writeThis` to match the others. I think it was just a typo in 41ae1eed844

mppf:
@ronawho - thanks very much for your help here. PR #7294, besides fixing nightly testing failures that have come up so far, also fixes a bunch of writeThis methods I found with grepping. Note that any writeThis that isn't in a module declaration is OK the old way (at least for now).

I'm finding it surprising how little we seem to test writeThis and dsiSerialWrite. I would have expected failures for the main distributions (block, cyclic, etc) with these but that didn't seem to happen in a standard paratest.

ronawho:
I don't think we're expected to manually test the different distributions (unless you're making huge changes to the distribution code)

IMO it's perfectly fine to let nightly testing do the heavy lifting and test all the various configs, because it's such a pain for us to manually configure/setup and test all the different configurations

For the PR that caused this, I probably wouldn't have thought to test ugni, I just happened to be doing some ugni performance runs and ran into this.

PR #7298, opened 2017-09-13
Support try/try! as an expression

mppf:
* Adds inTry and inTryBang fields on CallExpr
* adds PRIM_TRY and PRIM_TRY_BANG
* Adjusts parser to add PRIM_TRY / PRIM_TRY_BANG around try/try! expressions
* Adjust normalize pass to lower PRIM_TRY / PRIM_TRY_BANG by
   * marking the contained CallExprs appropriately
   * removing the primitive
   * (the actual transformation to do this is in errorHandling.cpp but called from normalize)
* Adds tests.

Why isn't it sufficient to use TryStmt? Because TryStmt is a *statement* and in particular it includes a block. We'd like to support try *expressions* in order to do things like support variable declarations that are initialized with a `try someCall()` or `try! someCall()`.

When modifying the parser, I learned from the Swift parser structure:
https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Expressions.html#//apple_ref/doc/uid/TP40014097-CH32-ID383

Passed full local testing.

Reviewed by @psahabu and @lydia-duncan - thanks!


mppf:
@lydia-duncan - this is the try-expression change I was telling you about. Please let me know if you have any feedback on it.

@psahabu - could you review? Thanks.

mppf:
(note, I havn't committed a rebuilt parser yet - you have to do that if you want to try it out)

compiler/AST/expr.cpp
```diff
@@ -902,7 +910,10 @@ CallExpr::CallExpr(const char* name,
   argList(),
   partialTag(false),
   methodTag(false),
-  square(false) {
+  square(false),
+  inTry(false),
+  inTryBang(false)
```
lydia-duncan: Would it be useful to have this be set via an enum selector of some kind, since a particular expression is not able to be both a try and a try! expression?

compiler/passes/errorHandling.cpp
```diff
@@ -828,6 +841,84 @@ bool ErrorCheckingVisitor::enterCallExpr(CallExpr* node) {
   return true;
 }
 
+class NormalizeTryExprsVisitor : public AstVisitorTraverse {
```
lydia-duncan: I'd define this in a separate file, since it is intended to be used at normalize(), right?

mppf: yes. New file? Or in normalize?

compiler/AST/expr.cpp
mppf: I didn't do the enum because it makes it more complex but I could do it & then we'd just have 1 new field. Do you think that would be obviously better?

compiler/passes/errorHandling.cpp
lydia-duncan: I think new file

compiler/AST/expr.cpp
lydia-duncan: I'm not sure.  This is obviously a complicated issue and I know you've thought about it (likely more than I have).

At first, it seemed like if the outermost `try` was a `try!`, then it wouldn't particularly matter because the line number for the uncaught error would be the same to the user, but then I realized that you would care about any side effects caused by the operations in the other part of the expression (though do we guarantee a particular execution order for them semantically, or is it an implementation detail?).  So if it were true that a `try!` would bleed to the inner parts of the expression, then I think it would simplify things to just have that be the representation once one has been encountered.

And maybe it is true since the implementation here fills in both fields without necessarily denoting which one was most immediately relevant?  I'm still trying to wrap my head around the idea, apologies.

What was your thinking on how an outer `try!` should impact the inner statements?

lydia-duncan: Ugh, that's a little stream of consciousness.

To simplify that, the implementation seems to carry over the context into inner statements, which makes me wonder about the precedence of `try` and `try!` on the inner statements, and whether it would simplify the analysis should one be a stronger statement than the other.

compiler/passes/errorHandling.cpp
```diff
@@ -790,14 +790,27 @@ bool ErrorCheckingVisitor::enterCallExpr(CallExpr* node) {
...
+      bool inThrowingFunction = false;
+      if (FnSymbol* parentFn = toFnSymbol(node->parentSymbol)) {
+        inThrowingFunction = parentFn->throwsError();
+      }
+
+      if (insideTry || node->inTryBang) {
+
         // OK
+      } else if(node->inTry) {
+        if (!inThrowingFunction) {
```
psahabu: I'd add an ``// OK`` here to be consistent with the rest of these checks.

PR #7301, opened 2017-09-13
Nested types: error for generics and initializers, and some code motion

lydia-duncan:
Add an error in a nested type context if the outer type and/or the inner type
has defined an initializer and the outer type and/or the inner type is generic,
when we first create the "outer" field for the inner type.

Additionally, it seemed odd that we were setting the "outer" field as part of
the creation of the type constructor, so I decided to make it occur independently
and prior to the creation of the type constructor (since it is needed for the
creation of the type constructor, default constructor, and default initializer).

Add a test where only the inner type is generic and the outer type is concrete
(when the outer uses an initializer), which fails today.  Also, update the .bad
files for the other generic nested types tests to reflect the new error message.

Passed a full std/ paratest with futures.  Running a paratest with
--force-initializers out of curiosity.


lydia-duncan:
Travis failure is due to "Couldn't bind to cpuset 0x00000001 because Invalid argument" error Elliot brought up in chat.  Merging anyways

PR #7302, opened 2017-09-13
split FLAG_REF_TO_IMMUTABLE from FLAG_REF_TO_CONST

vasslitvinov:
#### Now, two "const" flags

* FLAG_REF_TO_CONST means "illegal to modify through this reference"

* FLAG_REF_TO_IMMUTABLE means "the referenced thing does not change
during its lifetime"

Replaced the uses of the former with the latter when appropriate.

Besides code changes, this also takes a stand on the terminology,
* "const" means "illegal to modify",
* "immutable" means "will not change".

#### Enhanced isConstValWillNotChange()

It can now be invoked on ArgSymbols with abstract intents, too,
as long as can compute concreteIntent().

Looking at the implementation of isConstValWillNotChange(),
I did not see why it shouldn't work for an abstract intent.

#### Future work

The FLAG_REF_TO_CONST-illegal-to-modify really should be retired
in favor of QUAL_CONST_REF and INTENT_CONST_REF.

### DIFF
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -222,10 +222,6 @@ bool Symbol::isConstValWillNotChange() const {
   return false;
 }
 
-bool Symbol::isConstValWillNotChange_AnyIntent() const {
-  return false;
-}
-
 bool Symbol::isParameter() const {
   return false;
 }
@@ -572,19 +568,11 @@ bool VarSymbol::isConstant() const {
 }
 
 
-static bool isConstVWNChgHelp(const VarSymbol* var) {
-  // todo: how about QUAL_CONST ?
-  return var->qual == QUAL_CONST_VAL         ||
-         var->hasFlag(FLAG_REF_TO_IMMUTABLE) ||
-         (var->hasFlag(FLAG_CONST) && !var->hasFlag(FLAG_REF_VAR));
-}
-
 bool VarSymbol::isConstValWillNotChange() const {
-  return isConstVWNChgHelp(this);
-}
-
-bool VarSymbol::isConstValWillNotChange_AnyIntent() const {
-  return isConstVWNChgHelp(this);
+  // todo: how about QUAL_CONST ?
+  return qual == QUAL_CONST_VAL         ||
+         hasFlag(FLAG_REF_TO_IMMUTABLE) ||
+         (hasFlag(FLAG_CONST) && !hasFlag(FLAG_REF_VAR));
 }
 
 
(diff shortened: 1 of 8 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -3947,7 +3945,7 @@ static void handleTaskIntentArgs(CallInfo& info, FnSymbol* taskFn) {
       // records and strings are (incorrectly) captured at the point
       // when the task function/arg bundle is created.
       if (taskFn->hasFlag(FLAG_COBEGIN_OR_COFORALL) == true &&
-          isConstValWillNotChange(varActual)        == false &&
+          !varActual->isConstValWillNotChange_AnyIntent()   &&
```
benharsh: We should try and keep the style here uniform ( '!' vs. '== false').

benharsh: I don't see much benefit at this time for the "_AnyIntent" method. I'd prefer rolling back to the 'isConstValWillNotChange' standalone function.

vasslitvinov: @benharsh - I would like to use the "_AnyIntent" method in my upcoming forall-intents PR. Seems you'd be OK to keep it? We can surely tune the names.

benharsh: In that case let's keep it for now, and think of better names later.

compiler/AST/symbol.cpp
```diff
@@ -568,7 +569,10 @@ bool VarSymbol::isConstant() const {
 
 
 bool VarSymbol::isConstValWillNotChange() const {
-  return hasFlag(FLAG_CONST);
+  // todo: how about QUAL_CONST ?
+  return qual == QUAL_CONST_VAL         ||
+         hasFlag(FLAG_REF_TO_IMMUTABLE) ||
+         (hasFlag(FLAG_CONST) && !hasFlag(FLAG_REF_VAR));
```
mppf: I'm not so certain of this last clause. Wouldn't it be better to add FLAG_REG_TO_IMMUTABLE in the right places instead? In particular, I'm worried that a compiler-introduced reference temporary might end up with FLAG_CONST and be a reference. But maybe FLAG_CONST does not work that way. Can you confirm that this is OK?

vasslitvinov: @mppf my only confirmation is that this passes testing.

I think this is a really gray area because I don't think any of us knows exactly how FLAG_CONST works.

While I am not following exactly on what you are saying, one thing we can do to strengthen this is to switch from hasFlag(FLAG_REF_VAR) to isRef(). This should reduce the change of isConstValWillNotChange() returning true incorrectly. @benharsh what do you think about it?

mppf: I'm suspicious of this one:
https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/functionResolution.cpp#L5034
(because I can't explain it)

and this one:

https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/functionResolution.cpp#L5034

because it might set FLAG_CONST on a reference (a Qualifier type reference)

and this one:

https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/functionResolution.cpp#L5792

because the ArgSymbol might be a reference

After that I got tired of looking... I do think that switching to isRef() there is probably a good idea.

benharsh: Switching to isRef seems worth trying.

constness representation in the compiler is still a bit of a mess, so if our existing tests (much improved over the past year) still pass, then I'm content for now.

PR #7310, opened 2017-09-14
Use full filenames for project dependencies in mason

daviditen:
When a project named Project has a dependency named Foo-1.0.0, mason uses
the compile line:
`chpl src/Project.chpl -o target/debug/Project -M $MASON_HOME/src/Foo-1.0.0/src`

But we'd rather have it specify the dependencies source filename and specify
a main module:
`chpl src/Project -o target/debug/Project --main-module Project $MASON_HOME/src/Foo-1.0.0/src/Foo.chpl`

Modify `genSourceList()` to return a 3-tuple containing
(source URL, name, version) where before it combined name and version into
a single string.  Modify the two functions that call it to expect this.

Update the generated compile command to match what we'd like shown above.

Update the mason Makefile so it rebuilds when the mason source code changes.

Passed a linux64 paratest.

### DIFF
--- a/tools/mason/MasonBuild.chpl
+++ b/tools/mason/MasonBuild.chpl
@@ -114,9 +114,10 @@ proc compileSrc(lockFile: Toml, binLoc: string, show: bool,
   if isFile(pathToProj) {
     var command: string = 'chpl ' + pathToProj + moveTo + ' ' + ' '.join(compopts);
     if release then command += " --fast";
+    if sourceList.numElements > 0 then command += " --main-module " + project;
 
-    for dep in sourceList {
-      var depSrc = ' -M '+ depPath + dep(2) + '/src';
+    for (_, name, version) in sourceList {
+      var depSrc = ' ' + depPath + name + "-" + version + '/src/' + name + ".chpl";
       command += depSrc;
     }
 
@@ -145,14 +146,14 @@ proc compileSrc(lockFile: Toml, binLoc: string, show: bool,
 /* Generates a list of tuples that holds the git repo
    url and the name for local mason dependency pool */
 proc genSourceList(lockFile: Toml) {
-  var sourceList: [1..0] (string, string);
+  var sourceList: [1..0] (string, string, string);
   for (name, package) in zip(lockFile.D, lockFile.A) {
     if package.tag == fieldToml {
       if name == "root" then continue;
       else {
         var version = lockFile[name]["version"].s;
         var source = lockFile[name]["source"].s;
-        sourceList.push_back((source, name+'-'+version));
+        sourceList.push_back((source, name, version));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


tools/mason/MasonBuild.chpl
```diff
@@ -114,9 +114,10 @@ proc compileSrc(lockFile: Toml, binLoc: string, show: bool,
   if isFile(pathToProj) {
     var command: string = 'chpl ' + pathToProj + moveTo + ' ' + ' '.join(compopts);
     if release then command += " --fast";
+    if sourceList.numElements > 0 then command += " --main-module " + project;
```
ben-albrecht: Is there any downside to always throwing the flag `--main-module`?

ben-albrecht:
For context, cases where this could be problematic are:

1. The actual module name is different than the file name, e.g.

```chpl
// Foo.chpl
module Bar { .. }
```
I don't think this case will work still, since we currently assume that the `project name == module name`. I'm fine with this being against mason protocol for version `0.1.0`

2. There are multiple modules sharing the same name, e.g. if a mason brick conflicts with a package module or standard library module name. This is important for migrating package modules out to mason bricks.

ben-albrecht:
@daviditen - Do you think we should add a test for case 2 above?

tools/mason/MasonBuild.chpl
daviditen: Assuming that the module name and source file name match, no downside except being unnecessary.

ben-albrecht: I'm in favor of always throwing it to reduce additional code-complexity of this `if` statement, but it's trivial enough that I'll leave it up to you.

daviditen:
1. Yes, that would cause problems in the current scheme.  A way to solve it would be to add a main-module field to the .toml file.

2. Multiple modules sharing a name seems like it would always be a potential problem. I think the way the compiler currently handles it is to use the module that came from the command line specified file.

This currently works:
```
% mason new List
% cd List
% mason build
```

tools/mason/MasonBuild.chpl
daviditen: I think it's preferable to not use --main-module when it's not needed from a users perspective.  It just makes the command unnecessarily more complicated.

The added complexity is less than one line of code, so I'd prefer to leave it.

ben-albrecht: > but it's trivial enough that I'll leave it up to you.

Sounds good.

ben-albrecht:
> A way to solve it would be to add a main-module field to the .toml file.

True. We can punt that down the road. I think requiring the filename to share the module name is a reasonable restriction for `mason` packages, for now.

> Multiple modules sharing a name seems like it would always be a potential problem. I think the way the compiler currently handles it is to use the module that came from the command line specified file.

That's what I'd expect too. It might be worth adding a test case for this in `test/mason/` but I wouldn't hold up this PR for adding it.

PR #7313, opened 2017-09-14
Mason: Compile options in toml file

Spartee:
This PR is a feature for mason that allows users to put their compopts right in the ``Mason.toml`` file as listed in issue #7106. When this feature is used, the format for the ``Mason.toml`` will be a follows:

``` toml
[brick]
name = "MyProject"
version = "0.1.0"
compopts = "-M /test/somePackage"

[dependencies]
```

This is a step towards being able to support C dependencies. I do not believe this will conflict with the changes that @daviditen is proposing in #7310 but I am mentioning him in this just in case. 

This change has passed testing on my linux64 machine.

### DIFF
--- a/tools/mason/MasonUtils.chpl
+++ b/tools/mason/MasonUtils.chpl
@@ -22,7 +22,7 @@
 /* A helper file of utilities for Mason */
 use Spawn;
 use FileSystem;
-use TOML;
+
 
 /* Gets environment variables for spawn commands */
 extern proc getenv(name : c_string) : c_string;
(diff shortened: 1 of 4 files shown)


tools/mason/MasonUtils.chpl
```diff
@@ -31,6 +33,16 @@ proc getEnv(name: string): string {
 }
 
 
+/* Retrieves the compilation flags from lock file */
+proc getCompopts(lock: Toml) {
+  if lock.pathExists('root.compopts') {
+    var compopts = lock['root']['compopts'].s;
```
benharsh: I don't think you need the temporary 'compopts' here.

benharsh:
Does the 'compopts' field automatically get copied to 'root'?

tools/mason/MasonUtils.chpl
```diff
@@ -31,6 +33,16 @@ proc getEnv(name: string): string {
...
+    return compopts;
+  }
+  else return ' ';
```
daviditen: Why not return an empty string instead of a string containing a space?

daviditen:
I don't think this and #7310 conflict, so you should be good on that front.

Spartee:
@benharsh requesting a second review whenever you have time

PR #7314, opened 2017-09-14
Forwarding bug fixes

mppf:
This PR fixes several bugs with forwarding discovered in the effort to try to use forwarding from _array records to their _value.

Bugs fixed:
 * test calls were not kept in the AST, which caused problems with generic instantiation, because these can use the instantiation point of a generic (which was NULL).
 * methodTag/partialTag were not handled correctly in the test call
 * forwarding of iterators was not working correctly 
 
While there, added some compiler breakpoints:
 * before a few resolution warnings
 * for breakOnResolveID, if the type of that ID is established in resolveParallelIteratorAndIdxVar

- [x] full local testing

Reviewed by @vasslitvinov -thanks!

### DIFF
--- a/compiler/AST/foralls.cpp
+++ b/compiler/AST/foralls.cpp
@@ -545,6 +545,9 @@ static void resolveParallelIteratorAndIdxVar(ForallStmt* pfs,
   // Set QualifiedType of the index variable.
   QualifiedType iType = fsIterYieldType(parIter, iterCall);
   VarSymbol* idxVar = parIdxVar(pfs);
+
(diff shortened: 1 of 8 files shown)


mppf:
@vasslitvinov - would mind reviewing these further bug fixes to forwarding? thanks.

mppf:
Re. the file name, it's called "bug" because it was a bug report at one point. That information might be useful to future maintainers of the test.

vasslitvinov:
Sure. Let's put that as a comment in the .chpl file, then.

Issue #7003, opened 2017-08-12
LICM does not detect defs from virtual method calls


LouisJenkinsCS:
### Summary of Problem

Loop Invariant Code Motion (LICM) seems to cause some sort of 'undefined behavior' in that it causes some things to be compiled away entirely. Like how the tuple variable seems to be given undefined/uninitialized values. Furthermore, how the entire loop gets optimized away if we do not add that extra 'writeln' (of which is not needed with the `--no-loop-invariant-code-motion` flag).

As well, this is another opportunity to bring another issue that @e-kayrakli had made in my place, #6542 , which can also be reproduced here by calling `fn()` without capturing it in the the tuples `(a,b)` will produce a segfault (with and without `--no-loop-invariant-code-motion` flag).

The amount of time I've spent double, triple, and quadruple checking my data structure thinking that I corrupted my data structure or something was exhausting.

### Steps to Reproduce

**Source Code:**

[TIO (Online Compiler)](https://tio.run/##jZGxboMwEIZ3nuIvE0iEJEOXVFGHdMmSDs0LGDiCpcNG2IDaKq9eejhVsnY523f@P/93LhvVEc9zyco5vOE7ArrelqhNkmKHpLCWM2jj01ADGsU@iT8aO3AFYz0KQqmYqcrzPE5f5NI1ukbRjXgQxn@oPfmhN0hqxY4ybB@cUfUoA2UPQxMOidRCNlEZilTSie8HEW2kUNseW@Q5ngN4vca50Q6usZODb5SXQBgVLwJdiysxQdWyitJbDE5dpOSIWrecSzU4gjLQzg0kSBEVeNpjs5AMpl57YpPEaoc4g1iSWIR9EZq4uyzzpfsX6UlcnaynHY61QMQea0PygEyhtSNVWTDJ1nbQ3hHXmDRzGHY/yJBs53Wrv8S2mtTnaxrdTZyIKkk7K8KHaPkh23aa/xSwBufj@@2/5vmnrFld3LxaVTQS/wI)

**With `--no-loop-invariant-code-motion`**

[...]

PR #7315, opened 2017-09-14
Fix licm bug with dynamic dispatch

ronawho:
`astutil.cpp:isDefAndOrUse()` doesn't currently handle virtual methods, so we
weren't detecting a def of a variable through a virtual method call. Workaround
this limitation by checking for virtual methods defs in licm.

Note that the correct solution here is to update isDefAndOrUse, but that led to
regressions that I don't have time to look into yet.

Closes https://github.com/chapel-lang/chapel/issues/7003


ronawho:
@benharsh mind reviewing?

PR #7317, opened 2017-09-14
check if thrown errors are nil

psahabu:
Before, throwing a nil error would result in a no-op. The team determined that it would be beneficial for users to see that they had thrown nil by generating a `NilThrownError` and throwing that instead. This PR combines that functionality with the previously existing `gSaveLineInErrorFn`, now renamed to `gChplFixThrownError`.

While there, reorganized `gChplDeleteError` and renamed `gSaveTaskErrorFn` to `gChplSaveTaskError`.

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -379,21 +379,20 @@ bool ErrorHandlingVisitor::enterCallExpr(CallExpr* node) {
     BlockStmt* throwBlock = new BlockStmt();
     node->replace(throwBlock);
 
-    SymExpr*   thrownExpr    = toSymExpr(node->get(1)->remove());
-    VarSymbol* thrownError   = toVarSymbol(thrownExpr->symbol());
+    SymExpr*   thrownExpr  = toSymExpr(node->get(1)->remove());
+    VarSymbol* thrownError = toVarSymbol(thrownExpr->symbol());
 
-    VarSymbol* notNilError   = newTemp("not_nil_error", dtError);
-    CallExpr*  checkNilError = new CallExpr(gChplCheckNilError,
-                                            castToError(thrownError));
+    VarSymbol* fixedError  = newTemp("fixed_error", dtError);
+    CallExpr*  fixError    = new CallExpr(gChplFixThrownError,
+                                          castToError(thrownError));
 
-    throwBlock->insertAtTail(new DefExpr(notNilError));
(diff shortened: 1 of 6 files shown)


compiler/passes/errorHandling.cpp
```diff
@@ -379,16 +379,21 @@ bool ErrorHandlingVisitor::enterCallExpr(CallExpr* node) {
...
 
-    throwBlock->insertAtTail(new CallExpr(gSaveLineInErrorFn,
-                                          castToError(thrownError)));
+    VarSymbol* notNilError   = newTemp("not_nil_error", dtError);
+    CallExpr*  checkNilError = new CallExpr(gChplCheckNilError,
+                                            castToError(thrownError));
+
+    throwBlock->insertAtTail(new DefExpr(notNilError));
+    throwBlock->insertAtTail(new CallExpr(PRIM_MOVE, notNilError, checkNilError));
+    throwBlock->insertAtTail(new CallExpr(gSaveLineInErrorFn, notNilError));
```
mppf: Seeing it here, I wonder if we should merge gChplCheckNilError with gSaveLineInErrorFn, so that we only need to call one of these form the compiler? Not urgent, though. If you choose to merge them, chpl_fix_thrown_error might be a reasonable method name for it.

modules/internal/ChapelError.chpl
```diff
@@ -53,6 +53,14 @@ module ChapelError {
     }
   }
 
+  class NilThrownError : Error {
+    const nil_thrown_msg = "thrown error was nil";
```
mppf: I don't really see why not just return this string in message(), but it's not a big deal either way.

modules/internal/ChapelError.chpl
```diff
@@ -254,14 +252,19 @@ module ChapelError {
 
     return ret;
   }
-
   pragma "no doc"
-  pragma "insert line file info"
-  proc chpl_save_line_in_error(err: Error) {
+  pragma "replace nil, insert line file info"
```
mppf: what is pragma "replace nil" doing here?

modules/internal/ChapelError.chpl
```diff
@@ -254,14 +252,19 @@ module ChapelError {
...
+  proc chpl_fix_thrown_error(err: Error): Error {
+    var err_fix: Error = err;
```
mppf: idiomatically Chapel code uses camelCase

Issue #6532, opened 2017-06-23
Should Chapel raise an error on an infinite loop?


mppf:
``` chapel
while true { }
```

Should such a program compile? Or should there be an error or warning from the compiler?

mppf:
Compiler contained an implementation of this kind of check but it only applied to `while` and didn't handle `break` or `return` in the loop. For that reason I'm looking at removing the check.

bradcray:
I'd be inclined to remove it, on the basis that gcc doesn't seem to warn against such things even with -Wall on.

Checking the commit that added it (590e2c0), my sense is that it was motivated by the introduction of task intents, and so maybe concern about the case that a user was looping over something they thought was a var based on its declaration but was actually a 'const' shadow variable?  Maybe @vasslitvinov can provide more implementation.

Note that the original commit claimed it took breaks/returns into account.  Not sure if that capability was lost during a refactoring since then, or was mis-advertised.

mppf:
@bradcray - actually Vass' commit added a different error message (that I am not trying to remove). The error message I want to remove was added by Tom H in a "refactor" in commit 88287437c9854cdc2666382d8d5daffe93ea863f.

bradcray:
> The error message I want to remove was added by Tom H in a "refactor" in commit 8828743.

Oh, that wasn't clear to me in the issue description.  I assumed you were hoping to pull checkConstLoops() in general.  Could you clarify what the proposed change is?

mppf:
I'd like to:
1) Know if we want the compiler to be in this business at all, and
2) immediately remove Tom's USR_WARN (as shown in b9f96b37d649c6ae512c78c1fd2134aca46e90a1)

bradcray:
> Know if we want the compiler to be in this business at all

That's what led me to inquire about the reasons that Vass added the original code.

> immediately remove Tom's USR_WARN (as shown in b9f96b3)

I have no qualms about removing that diff.  I couldn't pick it out of the previous SHA you cited.

vasslitvinov:
As 590e2c0 says, "The warning will not be generated if the loop contains a break, return or yield statement [or] when the condition is a param." So, in a sense, break statements WERE handled.

Based on that message, I think the motivation was that some loops that used to be alright before task intents - will silently become infinite when the task intent feature is introduced. Today, this may be due to the user being unaware of task intents.

Here is an example:

```chpl
var done = false;
proc setDoneToTrue() { done = true; }

coforall ... {
[...]

PR #7322, opened 2017-09-14
GSoC Collections - LICM Changes

LouisJenkinsCS:
PR to remove LICM concerns/bug notes, and revised units tests to no longer avoid triggering LICM bug.

Currently there is another bug exposed with nested declarations being hoisted?

### DIFF
--- a/test/modules/packages/collections/CollectionBulk.skipif
+++ b/test/modules/packages/collections/CollectionBulk.skipif
@@ -1,2 +1,0 @@
-# 'fifo' tasking layer results in a double free in the runtime of a pthread_condvar. Works on Qthread
-CHPL_TASKS == fifo
--- a/test/modules/packages/collections/CollectionCounter.chpl
+++ b/test/modules/packages/collections/CollectionCounter.chpl
@@ -51,10 +51,12 @@ var barrier = new Barrier(here.maxTaskPar * numLocales);
 coforall loc in Locales do on loc {
   var perLocaleActual : atomic int;
   const _c = c;
-  var (hasElem, elt) : (bool, int);
+
   coforall tid in 0..#here.maxTaskPar {
     barrier.barrier();
-    (hasElem, elt) = (true, 0);
+    // BUG: Moving this declaration above the loop results in an incorrect
+    // compiler warning: 'A while loop with a constant condition'
+    var (hasElem, elt) : (bool, int) = (true, 0);
     var perTaskActual : int;
     while hasElem {
       perTaskActual = perTaskActual + elt;
--- a/test/modules/packages/collections/CollectionCounter.chpl
+++ b/test/modules/packages/collections/CollectionCounter.chpl
@@ -51,9 +51,10 @@ var barrier = new Barrier(here.maxTaskPar * numLocales);
 coforall loc in Locales do on loc {
   var perLocaleActual : atomic int;
   const _c = c;
+  var (hasElem, elt) : (bool, int);
   coforall tid in 0..#here.maxTaskPar {
     barrier.barrier();
-    var (hasElem, elt) : (bool, int) = (true, 0);
+    (hasElem, elt) = (true, 0);
     var perTaskActual : int;
     while hasElem {
       perTaskActual = perTaskActual + elt;
--- a/test/modules/packages/collections/CollectionNQueens.chpl
+++ b/test/modules/packages/collections/CollectionNQueens.chpl
@@ -156,9 +156,9 @@ coforall loc in Locales do on loc {
   coforall tid in 0 .. #here.maxTaskPar {
     barrier.barrier();
     var nSpins : int;
+    // BUG: If we do not specify the type here, we get a compiler internal error...
     var (exists, myBoard) : (bool, boardType);
     while found.read() < totalSolutions {
-      // BUG: If we do not specify the type here, we get a compiler internal error...
       (exists, myBoard) = c.remove();
       // Spin: We haven't found solution yet...
       if !exists {
--- a/test/modules/packages/collections/CollectionNQueens.chpl
+++ b/test/modules/packages/collections/CollectionNQueens.chpl
@@ -156,9 +156,10 @@ coforall loc in Locales do on loc {
   coforall tid in 0 .. #here.maxTaskPar {
     barrier.barrier();
     var nSpins : int;
+    var (exists, myBoard) : (bool, boardType);
     while found.read() < totalSolutions {
       // BUG: If we do not specify the type here, we get a compiler internal error...
-      var (exists, myBoard) : (bool, boardType) = c.remove();
+      (exists, myBoard) = c.remove();
       // Spin: We haven't found solution yet...
       if !exists {
         nSpins = nSpins + 1;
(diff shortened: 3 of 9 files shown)


test/modules/packages/collections/CollectionCounter.chpl
```diff
@@ -51,9 +51,10 @@ var barrier = new Barrier(here.maxTaskPar * numLocales);
 coforall loc in Locales do on loc {
   var perLocaleActual : atomic int;
   const _c = c;
+  var (hasElem, elt) : (bool, int);
   coforall tid in 0..#here.maxTaskPar {
     barrier.barrier();
-    var (hasElem, elt) : (bool, int) = (true, 0);
+    (hasElem, elt) = (true, 0);
     var perTaskActual : int;
     while hasElem {
```
LouisJenkinsCS: @ronawho 

```
A while loop with a constant condition
```

ronawho: Hmm, this check occurs long before LICM is actually run, so I think it's unrelated to that. Does this cause any failures beyond getting the warning itself?

https://github.com/chapel-lang/chapel/issues/6532 has some related info

test/modules/packages/collections/CollectionNQueens.chpl
```diff
@@ -156,9 +156,10 @@ coforall loc in Locales do on loc {
   coforall tid in 0 .. #here.maxTaskPar {
     barrier.barrier();
     var nSpins : int;
+    var (exists, myBoard) : (bool, boardType);
     while found.read() < totalSolutions {
       // BUG: If we do not specify the type here, we get a compiler internal error...
```
ronawho: move this comment up with the declaration?

LouisJenkinsCS: Done.

test/modules/packages/collections/CollectionCounter.chpl
LouisJenkinsCS: @ronawho 

No errors, get a good old `SUCCESS`, thanks for the fix!

@mppf 

Seeing as you merged a patch for infinite loop checking, was wondering if you could take a look on this when you have the time.

test/modules/packages/collections/CollectionBulk.skipif
```diff
@@ -1,2 +0,0 @@
-# 'fifo' tasking layer results in a double free in the runtime of a pthread_condvar. Works on Qthread
```
ronawho: Is this skipif removal intentional?

test/modules/packages/collections/CollectionCounter.chpl
LouisJenkinsCS: In fact, if I got some information on where to go to approach this myself, I can potentially attempt it myself (such as, where these checks occur, potentially 'why' it occurs, a decent tour around the Chapel compiler, etc.)

ronawho: For this PR I think we should move the declaration back down to avoid the `A while loop with a constant condition` warning. That warning code either needs to be improved (or IMO removed), but I think we should that in a separate PR.

test/modules/packages/collections/CollectionBulk.skipif
LouisJenkinsCS: Yeah, I forgot to remove it last time. #7136 fixed the issue for the 'fifo' tasking layer, and I just remembered now. I suppose this kind of widens the topic of this PR.

ronawho: I'm fine with it being removed here, just wanted to make sure it was intentional since I didn't see it noted in the PR message

test/modules/packages/collections/CollectionCounter.chpl
ronawho: And FYI: the check occurs in `compiler/AST/WhileStmt.cpp::checkConstWhileLoop()` as part of `WhileStmt::checkConstLoops()`, which is ultimately called during `functionResolution()`

The warning is (clearly) inaccurate, and at least I don't think it's very valuable anyways

ronawho:
I'm actually not sure what the testing policy for external contributors is. For this small PR I think local testing on your machine for a single configuration is plenty. I'll do some sanity checks on my end too.

test/modules/packages/collections/CollectionNQueens.chpl
```diff
@@ -156,9 +156,10 @@ coforall loc in Locales do on loc {
   coforall tid in 0 .. #here.maxTaskPar {
     barrier.barrier();
     var nSpins : int;
+    // BUG: If we do not specify the type here, we get a compiler internal error...
```
ronawho: Is there already an issue for this bug? If not, I'll add one

Also.. sorry that you've hit so many bugs :(

LouisJenkinsCS: It is an issue, but I didn't have time to figure out what the cause was. It had something to do with 'Visibility Blocks', and I couldn't figure out why, all I know is that it cropped up around the time I converted `DistributedDeque` to record wrapping. 

Also no problem, occupational hazards and all that.

test/modules/packages/collections/CollectionCounter.chpl
LouisJenkinsCS: So, tracing it, it looks like the issue is [this](https://github.com/chapel-lang/chapel/blob/master/compiler/AST/WhileStmt.cpp#L313-L315), that it incorrectly sees it as a constant expression. It seems the issue is [here](https://github.com/chapel-lang/chapel/blob/master/compiler/AST/WhileStmt.cpp#L320)? It only goes one parent up, when it seems to need to check all parents.

LouisJenkinsCS: In particular, [this](https://github.com/chapel-lang/chapel/blob/master/compiler/AST/WhileStmt.cpp#L320) line may stop it too early, if `parent == this`

LouisJenkinsCS: Actually, nevermind, ignore that, issue would be that it sees `condSym` as `true` to begin with. I'll just investigate silently until I actually arrive at the conclusion.

LouisJenkinsCS: @ronawho 

Sorry to bother you again, but is it possible for `GotoStmt::gotoTag` to be this value: `42262848`? If the answer is no, its possible that it got mangled or something, because if you look at the type, its an `enum GotoTag`, defined [here](https://github.com/chapel-lang/chapel/blob/e67f28cfd0a81094e05830c9cc6fb55da97a9f38/compiler/include/stmt.h#L191-L201). As can be seen, an `enum` shouldn't have an undefined value unless it wasn't initialized correctly, leading to undefined behavior.  I got the value by using gdb. If this is due to it simply being uninitialized its an easy fix but can lead to some major complications down the road if anything else wanted to use any other other uninitialized values. In any event, this undefined value causes it to fail [this](https://github.com/chapel-lang/chapel/blob/master/compiler/AST/WhileStmt.cpp#L356) check.

Actually, after typing this, it seems that it may be a pointer, and thats the issue. Casting it in gdb to another `GotoStmt *` shows that it directs to something with `GOTO_NORMAL` tag. Perhaps the issue is that its too specific with what it expects, or is there a reason that it only checks for `return` and `break`?

mppf: Regarding the compiler changes - might be time to move the discussion to an issue / another PR. I personally wouldn't cry if we entirely removed those checks for infinite loops though.

Regarding GotoTag having a weird value - usually I try do debug that sort of thing by running the compiler in valgrind.

PR #7324, opened 2017-09-15
Support for 'chplVersion' in mason.

### LINKED ISSUE #7279: Mason: Chapel Versioning
## Chapel Versioning in Mason

Add support for a `chplVersion` field in `Mason.toml`:

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.16.0"

```

### field name options

- **`chplVersion`**
- ~`chpl`~
- ~`chapelVersion`~

## Range Syntax

Ranges are specified with Chapel syntax:

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.13.0..1.16.0"

``` 

minor-minor releases can be implied:

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.13..1.16"

``` 

minor releases *cannot* be implied:

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1" # invalid

``` 
"... or later" ranges are expressed by a single version and include later major versions:

``` toml

[brick]
name = "MyPackage"
version = "0.1.0"
chplVersion = "1.16.0" # 1.16, 1.17, 2.0, 10.10.10

``` 

## Default Value

When creating a package with `mason new`, the default `chplVersion` is the version of the available `chpl` compiler that `mason` is currently using. This means that by default the package is assumed to work with later versions of the compiler and standard modules.

## Dependency Resolution

When resolving  package versions, only packages matching the current Chapel compiler version should be considered. There is no dependency resolution strategy for Chapel versions themselves.

benharsh:
This field can control the allowed version of Chapel, as determined by
the output of 'chpl --version'. A string like "1.16" allows for any
version (including major versions) later than 1.16. A string like
"1.16.0..1.20" allows for any version between 1.16 and 1.20 (inclusive).
Versions can optionally leave off the 'bug' part of the version (i.e.
'9' in 1.0.9).

By default, new projects will use the current Chapel version as an
unbounded range (e.g. "1.16").

The 'update' subcommand now supports an argument '--no-update-registry'
to allow for local fake registries to be used in testing. Local
registry directories can be specified with the environment variable
MASON_LOCAL_REGISTRY_DIR, though this should only be used for testing.

If multiple dependencies fail the chplVersion check, they will all be
listed in the error message.

If all dependencies in IVRS fail the chplVersion check, mason will
currently halt and only print an error message for that package. This
could be improved upon in the future.

This commit also fixes a bug where IVRS would list the wrong version in
the "dependencies" array in the lock file. This could happen if a later
minor version was encountered before a lower minor version, and the
later version was disqualified by a chplVersion check.

Lock files restrict the chplVersion to the current version of Chapel
with a bounded range.

Resolves #7279


tools/mason/MasonUpdate.chpl
```diff
@@ -36,16 +36,46 @@ The current resolution strategy for Mason 0.1.0 is the IVRS as described below:
        (ex. 1.13.0, 2.1.0 --> incompatible)
 */
 
+private var failedChapelVersion : [1..0] string;
 
+proc getRegistryDir() {
```
ben-albrecht: Any downsides to making this `proc registryHome { ... }` ?

benharsh:
Passes on both OS X and linux64.

Issue #7042, opened 2017-08-17
LICM incorrectly hoists expression dependent on record in global array


benharsh:
### Summary of Problem
I ran into an LICM bug while developing CoMD where an expression
was incorrectly hoisted from the conditional of a while loop. An operand
of the comparison was updated on each loop iteration, and should not
be hoisted.

**Associated Future Test(s):**
[test/optimizations/licm/whileExpr.chpl](https://github.com/chapel-lang/chapel/blob/master/test/optimizations/licm/whileExpr.chpl)

PR #7326, opened 2017-09-15
Fix licm for defs to variables through function calls

ronawho:
LICM previously assumed that any function calls may modify global variables,
but it didn't do this for local variables that were defined outside the loop's
scope. A nested function could change a local variable, without LICM having
visibility, so update licm to assume a function call can change any variable
not declared in the loop's scope.

This resolves https://github.com/chapel-lang/chapel/issues/7042


ronawho:
@benharsh another one for you

ronawho:
Yeah, it could have some performance regressions. Want me to playground it for a night?

benharsh:
Not sure, I could go either way. This would be an easy one to revert.

ronawho:
Turns out this led to large comm count regressions, and didn't actually solve the real issue, it just happened to work in the test cases since they used a nested function. Closing in favor of https://github.com/chapel-lang/chapel/pull/7371 (which includes a test case that doesn't use a nested function that fails with this PR)

PR #7328, opened 2017-09-15
Chameneos initializer updates

bradcray:
While reviewing release-over-release performance numbers and looking at chameneos versions, I was reminded that now that initializers are much improved, we can write nicer initializers for our chameneos versions.  This PR:

* moves a phase 2 array assignment in Elliot's version's initializer into a phase 1 assignment and uses promotion of a constructor (a forall expression could be used instead if that felt like the right clarity / code size tradeoff)
* copies my "ideal" version to the "blc" version now that it's working to get timings on it (the previous blc version was virtually identical to the submitted version 2.  The main difference between this version and Elliot's is the use of a generic field rather than explicit domain/array fields.  Since my "ideal" version was never timed for performance, this will give us a first opportunity to do so
* creates a new (futurized) ideal version that results in an internal error in the compiler.  Interestingly, it uses concepts from Elliot's (the promoted constructor) and mine (the generic field) either of which works separately, but which blow up when used together.

### DIFF
--- a/test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc-ideal.bad
+++ b/test/studies/shootout/chameneos-redux/bradc/chameneosredux-blc-ideal.bad
@@ -1,0 +1,8 @@
+chameneosredux-blc-ideal.chpl:56: internal error: INI0465 chpl Version 1.16.0 pre-release (5007c1a)
+Note: This source location is a guess.
+
+Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
+and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
+please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
+the filename + line number above may be useful in working around the issue.
+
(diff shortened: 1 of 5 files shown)


bradcray:
@ronawho: Will you review this?
@lydia-duncan: Will you check to see whether the new future in the -ideal version is known or whether I should file an issue against it?

ronawho:
Looks good to me

bradcray:
@ronawho: FYI, I updated to include the promotion using an anonymous range that I mentioned at the end of the day today.  I think it adds a lot of clarity to the promotion.

lydia-duncan:
This is a new error.  Go ahead and file an issue, I might take a look at how difficult it would be to solve this (from what I can tell, the problem seems to be independent of the promotion and specific to a generic field getting initialized with `new`)

lydia-duncan:
Further information - the problem is also present when the field has an initial value but no type (var y = new Rec();).  The error source is different (due to it following different but similar code paths), but the solution for both should be similar.  I believe this was an oversight from fixing the same error when the type of the field is explicitly declared

lydia-duncan:
See #7336 for more cases that run into this error

bradcray:
Hi Lydia -- I think you've got a better grasp on the error than I do.  Would you take care of filing the issue?  (unless you think the fix is imminent?)

lydia-duncan:
Gladly!

PR #7329, opened 2017-09-15
Add futures for arrays of records of const fields that can't be init'd

bradcray:
Here are some tests that show how a record with a const field is troublesome in initialization contexts at present.  This is believed to be at least partially due to the fact that the compiler creates constructors for the records rather than initializers, though there are likely additional challenges hiding behind that one:

* initRecFieldWithConstField.chpl: This is the simplest and a good starting point.  It shows that because we use assignment for initialization of a record field, the compiler complains when that field has a const field (since it can't be reassigned; yet we aren't assigning it, we're init-ing it, so this should be OK in Chapel, one way or another)

* initArrOfRecConstField*.chpl: This is a series of tests in which the record has an array of records with a const field and runs into similar troubles (though hopefully these are mostly just variations on the above).

* arrOfRecConstField.chpl: This shows that creating an array literal of records with a const field has a similar issue.  This is arguably another form of initialization though I think it goes through a different (and probably reasonably specialized) code path.  This variant is the first stumbling block for the initArrOfRecConstField.chpl and initArrOfRecConstField2.chpl variants above.


bradcray:
@lydia-duncan and @noakesmichael, could you take a look to see whether I'm doing something dumb and whether this is a known issue?  Thanks!

noakesmichael:
I think it’s easy to agree that this is exposing “a problem” but I am
not 100% clear what the problem is.



I took a very  quick glance at the simpler version of the issue
i.e. initArrofRecConstField.chpl




If I am reading things correctly then this is not an initializer issue.



1) I believe the error is occurring while the compiler is trying to build
the literal expression [myR, myR, myR].  This appears to be the
expression that’s generating the call to chpl__buildArrayExpr.

Line 485 of that function is indeed an assignment statement in a
param loop.



2) The following variation of that program seems to me to show
that the same problem exists for a “constructor” based program.
That trivial constructor should ensure we’re not invoked new
initializer paths and a glance at the AST suggests that’s right.


record MyRec {
  proc MyRec() {
    c = 10;
    v = 30;
  }

  const c: int;
  var   v: int;
}

var myR = new MyRec(c=1, v=2);

var A: [1..3] MyRec = [myR, myR, myR];

writeln(A);



The piece that confuses me at this point is how one could write
an assignment operator for MyRec given that it has a const
field


Presumably there’s a way to get that part to work but I don’t
know the trick off the top of my head.  Or perhaps this has
exposed an issue with chpl__buildArrayExpr that we haven’t
thought about before?


Mike



On Sep 14, 2017, at 9:40 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


@lydia-duncan<https://github.com/lydia-duncan> and @noakesmichael<https://github.com/noakesmichael>, could you take a look to see whether I'm doing something dumb and whether this is a known issue? Thanks!

bradcray:
@noakesmichael : I think I moved too quickly in condensing the original case I found that seemed to be hitting an initializer-oriented issue down into these futures.  Long story short, I think the `initRecFieldWithConstField.chpl` future that I just added to the PR captures what I was originally going for.

You're right of course that the first order problem in the tests I filed last night was due to the creation of the array literal rather than its assignment to an object field, and I just added a new future to this PR, `arrOfRecConstField.chpl` to capture that issue distinctly from initializers.  But I think there's an initializer-oriented issue as well, and have created new versions of last night's test `initArrOfRecConstField[1|2][a|b].chpl` to try and demonstrate it.  These use RHS initializers for the array field that aren't array literals (one uses a forall loop, the other relies on a promoted assignment from `R` to `array of R`) and result in similar issues due to the use of assignment within the initializer.

But even these tests are overkill now that I think about it a bit more, and `initRecFieldWithConstField.chpl` demonstrates what I think is at the heart of the issue, which I think you touched on last night in your comment as well:  I think it's correct that assignments for records with const fields don't make sense (at least, assignments that try to assign that const field for the LHS record), but that in the initialization context, we either shouldn't be using assignments to initialize such fields or should be relaxing the const field assignment rules within them in order to permit records with const fields (and arrays of same) to be members.  That probably wasn't clear.  Hopefully the `initRecFieldWithConstField.chpl` test is.

This seems _somewhat_ related to me to the whole copy constructor / move/steal constructor question so I'm tagging @mppf on this as well due to the things he's done there.

bradcray:
I've updated the PR message above to try and describe the issue and tests I'm proposing adding.

mppf:
I just looked at `initRecFieldWithConstField.chpl` and it seems clear enough to me that it should be copy-initializing from the `const ref` `initR` argument in to the field `r`. There should not be a problem copy-initializing a `const` field. I can't think of any reason a call to `=` is necessary in this case.

What I don't know is why there is currently an error about assignment or how hard it would be to adjust the implementation. I wonder if this is related to some of the issues/workarounds with copy initializers (especially issues with generic records), but I havn't tracked the implementation very closely.

bradcray:
@noakesmichael / @lydia-duncan / @mppf : This PR is independent of the release (in the sense that it doesn't affect it to be merged or not)—should I merge it and file an issue against it or something else?

bradcray:
@noakesmichael points out what should've probably been obvious: that the issue with my simplest test is that `R` doesn't define an initializer and so the compiler creates a constructor for it rather than an initializer which isn't invoked by phase 1 of the initializer that I do define.  This probably accounts for the errors in some of the other array cases as well, though Mike and I hypothesize that there will still be an issue with the array literal case (and discussed possibly using a constructor on the _array type to implement the creation of such literals in a way that would work in the future).

I'm left unsure what aspects of this PR are new issues vs. ones that are already tested.  @lydia-duncan, you seem to be keeping good tabs on things -- could you help with that?

(Mike also suggested running this with the experimental "compiler, make initializers for me" flag to see which passed or failed in that case)

test/types/records/const-checking/initArrOfRecConstField.future
```diff
@@ -0,0 +1,7 @@
+bug: should be able to init arrays of records of const fields
+
+This test demonstrates that arrays of records that have const fields
+can't be initialized, though it seems that they should be able to.  As
+Mike points out, the zeroth order concern here is that we can't create
+array literals of record values with const fields as
+arrOfRecConstField.chpl shows.
```
lydia-duncan: No newline at the end of this file, fyi

test/types/records/const-checking/initArrOfRecConstField1a.future
```diff
@@ -0,0 +1,4 @@
+bug: should be able to init arrays of records of const fields
+
+This test demonstrates that arrays of records that have const fields
+can't be initialized, though it seems that they should be able to.
```
lydia-duncan: And here

test/types/records/const-checking/initArrOfRecConstField1b.future
```diff
@@ -0,0 +1,4 @@
+bug: should be able to init arrays of records of const fields
+
+This test demonstrates that arrays of records that have const fields
+can't be initialized, though it seems that they should be able to.
```
lydia-duncan: And here

test/types/records/const-checking/initArrOfRecConstField2.future
```diff
@@ -0,0 +1,7 @@
+bug: should be able to init arrays of records of const fields
+
+This test demonstrates that arrays of records that have const fields
+can't be initialized, though it seems that they should be able to.  As
+Mike points out, the zeroth order concern here is that we can't create
+array literals of record values with const fields as
+arrOfRecConstField.chpl shows.
```
lydia-duncan: And this one

test/types/records/const-checking/initArrOfRecConstField2a.future
```diff
@@ -0,0 +1,4 @@
+bug: should be able to init arrays of records of const fields
+
+This test demonstrates that arrays of records that have const fields
+can't be initialized, though it seems that they should be able to.
```
lydia-duncan: I suspect all the .future files are impacted at this point

lydia-duncan:
Since all your tests share the use of a default constructor for the record being put in the array, I suspect these are all old issues.  I doubt default initializers have fixed this problem but it is probably worth checking.

I do not know of any tests that already cover this issue.  I have no problem with you merging this today and filing an issue for it.

bradcray:
Following up on the discussion thread... I tried adding explicit initializers to several of the tests to see if they made the errors go away, but no dice.  So the first thing to focus on is getting default initializers online, but then there is likely to be additional work after that point.

Issue #6542, opened 2017-06-26
Returning by-ref values from overridden methods and not capturing them at call site


e-kayrakli:
### Summary of Problem
Returning some values from overridden methods seems to cause compiler crashes, if the value returned is not captured at the call site.

(I am not entirely sure if the title of the issue captures the problem in a good way)

A little bit of GDB'ing shows that `calledFn` in https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/callDestructors.cpp#L515 seems to be null if the `call` is a `PRIM_VIRTUAL_METHOD_CALL` causing a null dereference in the next line.

### Steps to Reproduce

**Source Code:**

```chapel
class A {
  proc foo() { return (3, 3); }
}

class B:A {
  proc foo() { return (1,1); }
}

var b:A = new B();
b.foo();
```

I have tried and confirmed that this crashes if `foo` returns: homogeneous/heterogeneous tuples and strings. Returning a class instance field of A seems to compile fine, any primitives are also fine. Moreover returning a primitive class field as such also seems fine:

```chapel
proc foo() ref { return someIntField; }
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
Doesn't compile.

**Associated Future Test(s):**
[`test/classes/elliot/non-captured-virtual-return.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/elliot/non-captured-virtual-return.chpl) #7318 

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.16.0 pre-release (405f2a2a7a)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```

- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```

e-kayrakli:
[...]

PR #7332, opened 2017-09-15
Add resolvedOrVirtualFunction, fix issue #6542

mppf:
This PR takes two steps to fix issue #6542:
  1. It introduces a function, `CallExpr::resolvedOrVirtualFunction`, factoring out a code pattern used already in callDestructors and errorHandling. The pattern is to find the resolved function, or, for a PRIM_VIRTUAL_METHOD_CALL, find one of the possible virtual methods that could be called at runtime. This function is appropriate to call for code that identifies calls to transform based upon the signature of the called function.
 2. It calls this function in two places in order to resolve issue #6542.

What was going wrong with issue #6542? The compiler includes a transformation that happens during resolution, `insertReturnTemps`, which finds all calls to functions that return a value that don't use the result of that return. In that case it adds a temporary variable to capture the returned value.

Code in callDestructors / ReturnByRef relied on calls to transformable functions having a variable to capture the returned value.

In the case of virtual method calls, `insertReturnTemps` was not transforming the call, and so the code pattern in issue #6542 led to failed assertions in callDestructors.

The important part of this fix is to adjust `insertReturnTemps` to handle virtual calls (PRIM_VIRTUAL_METHOD_CALL).

Passed full local testing.

### DIFF
--- a/compiler/AST/expr.cpp
+++ b/compiler/AST/expr.cpp
@@ -1198,6 +1198,30 @@ void CallExpr::setResolvedFunction(FnSymbol* fn) {
   }
 }
 
+// This function returns the resolved function, if it's totally resolved,
+// or some resolved virtual function (probably a virtual parent that
+// won't actually called at runtime) if it's a virtual method call.
+// This function is useful for transformations on calls that work
+// depending on the called function's signature (since the virtual
+// parent and children will have the same signature).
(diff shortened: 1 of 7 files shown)


mppf:
@noakesmichael - would you mind looking at this bug fix? Thanks.

PR #7336, opened 2017-09-15
Add a future for the variations of a field initialized with 'new' failing

lydia-duncan:
For initializers that explicitly initialize fields of class or record type with
a `new`, we previously fixed the errors for fields declared in the form of:

```
var x: Rec = new Rec();
var x: Rec;
```

But apparently missed the errors for fields declared in the form of:

```
var x = new Rec();
var x;
```

This adds futures to cover that oversight.

See also #7328 for where this issue surfaced.

My kingdom for a unit test generator . . .


lydia-duncan:
Heads up, @noakesmichael

PR #7338, opened 2017-09-15
Improve c2chapel visibility

benharsh:
Adds c2chapel to the top-level Makefile, online docs, and release.


doc/rst/tools/c2chapel/c2chapel.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Building
+--------
+
+``c2chapel`` is built on top of `pycparser <https://github.com/eliben/pycparser>`_.
+Building ``c2chapel`` will install pycparser into a local virtualenv, leaving
+the user's python environment untouched. A symbolic link will be created in the
+appropriate directory under ``$CHPL_HOME/bin``.
+
+To build, run ``make c2chapel`` from ``$CHPL_HOME``.
```
daviditen: This didn't work for me.
```
virtualenv /Users/diten/chapel/tools/c2chapel/install/venv
/bin/bash: virtualenv: command not found
```
Is there a prerequisite that a virtualenv be set up somehow?

benharsh: Yes, though I (apparently incorrectly) thought having python implied having virtualenv. I'll add that to the prerequisites.

daviditen: I'm able to build chpldoc and our test environment with `make chpldoc` or `make test-venv`, and they both use virtualenv.  Can the Makefile be set up like for them so it just works?

benharsh: Possibly, but I'm not sure what it would take to get that working. Could that be part of a later version of c2chapel?

daviditen: I think it would be really nice to use the same trick the other tools use so virtualenv isn't a prerequisite.  If it won't be straightforward then it can wait until later.

benharsh:
@daviditen : this PR will also add c2chapel files to the release tarball.

I'm leaning towards going without automatic virtualenv installation for this release. There's a lot about chpl-venv's Makefiles that I'm not familiar with and would rather not try and duplicate in the next few days.

doc/rst/tools/index.rst
```diff
@@ -14,4 +14,5 @@ Contents:
 
    chpldoc/chpldoc
    chplvis/chplvis
-   mason/mason
\ No newline at end of file
+   c2chapel/c2chapel/
```
daviditen: The trailing slash can be removed here, and c2chapel is at the top alphabetically.

tools/c2chapel/README.rst
```diff
@@ -16,86 +16,5 @@ function declaration in a header file:
...
-Known issues:
-- enums may cause fatal errors
-- fake standard headers are incomplete
-
-============
-Contributors
-============
-| Ben Harshbarger [Cray Inc.]
-| Nikhil Padmanabhan [Yale University]
+See the documentation in ``$CHPL_HOME/doc/rst/tools/c2chapel/`` for more
```
daviditen: It might be nicer to link to the online version here.

daviditen:
I think you should sort the tools by name in doc/rst/tools/index.rst
The comment about linking the online vs plaintext versions is a matter of personal preference, take it or leave it.

Otherwise I think this looks good to go.

daviditen:
Thanks for getting it working for users without virtualenv preinstalled.  That is a big win IMO.

PR #7340, opened 2017-09-15
Set the line number before copying leader iterators in implementForallIntents

daviditen:
Issue #6998 hit an internal error because the line number was not set when
adding an AST. Set the line number with SET_LINENO to fix the internal error.

The test from #6998 still fails because dynamic dispatch doesn't appear to
be working on leader iterators. Oddly, after removing the halts from that
test, the parent leader is called, but the child follower is called.

Update a test's .bad and .future files to match the new failure mode.


daviditen:
Thanks!

bradcray:
Nice, David!

Issue #7296, opened 2017-09-13
Compress zeroes out of comm diagnostic counts


bradcray:
When we print out communication counts for a multilocale program, the output typically looks like this:

```
(get = 0, get_nb = 0, put = 0, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 8, execute_on_fast = 0, execute_on_nb = 6) (get = 101, get_nb = 0, put = 1, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 4, execute_on_fast = 2, execute_on_nb = 0) (get = 101, get_nb = 0, put = 1, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 0, execute_on_fast = 2, execute_on_nb = 0) (get = 101, get_nb = 0, put = 1, put_nb = 0, test_nb = 0, wait_nb = 0, try_nb = 0, execute_on = 0, execute_on_fast = 2, execute_on_nb = 0)
```

which makes it hard to compare across two outputs or to find key values of interest.  In this issue, I'm proposing that we drop all 0 values for a more compact and readable output, like:

```
(execute_on = 8, execute_on_nb = 6) (get = 101, put = 1, execute_on = 4, execute_on_fast = 2) (get = 101, put = 1, execute_on_fast = 2) (get = 101, put = 1, execute_on_fast = 2)
```

I think we should also print out something in the event that all values are 0, like:

```
(<no communication>)
```

ronawho:
👍  for this. 

FWIW `git diff --word-diff` can make the diffs a little easier to read for things like this (if you have a shell that supports colors)

gbtitus:
I spazzed my addition of a thumbs-up and it appears in doing so erased the one from @ronawho .

mppf:
My work on making comm-counts tests into perf tests is certainly related to this. See PR #6604.

bradcray:
@mppf: Just wanted to check and make sure you didn't object to the direction here for tests that haven't been converted away from using printing of comm diagnostics?

mppf:
Nope, it's just fine.

PR #7343, opened 2017-09-15
Printing comm diagnostics squashes zeroes

bradcray:
This PR adds a `writeThis()` overload to the record used to collect communication
diagnostics such that it won't print out 0 entries.  This permits the human reader to
focus on the interesting values given that, for many use cases, most of the values are
zero.

Most of the work here is in updating existing .good files.  As part of this effort, I
retired the ssca2_commDiags.chpl test which counted communications for SSCA2
using some fuzzy matching logic.  Now that we're (a) tracking SSCA2 performance
on multiple locales nightly and (b) not as focused on SSCA2 specifically as we were
in Chapel's early days, this test doesn't seem worth the maintenance burden any
longer.


bradcray:
This was proposed and received support in issue #7296.

modules/standard/CommDiagnostics.chpl
```diff
@@ -215,6 +215,53 @@ module CommDiagnostics
       non-blocking remote executions
      */
     var execute_on_nb: uint(64);
+
+    proc writeThis(c) {
```
benharsh: This is fine for now, but could probably be more elegantly written using Reflection.

bradcray: I'm not understanding this comment...

benharsh: I think you could use numFields, getFieldName, and getField within a param loop instead of many if-statements.

bradcray:
Just pushed a nice reflection-based cleanup.  What a great language!  (if only I weren't too lazy to go read documentation without prompting...)

PR #7346, opened 2017-09-16
Don't inline virtual leader/standalone iterators

ronawho:
Note that this requires building up the zip* functions for leader/standalone
iterators. 94f83afd47 originally stopped building the zip* for parallel iters
under the assumption that they would always be directly inlined, but now we
will need them for virtual leader/standalone iters. We could probably do
something smarter to avoid building up the zip* functions for any iterator that
will always be directly inlined, but I did not take that step here.

Also note that FLAG_INLINE_ITERATOR is really just a bad name for marking
leader/standalone iters.

This resolves https://github.com/chapel-lang/chapel/issues/6998


ronawho:
Follow on to #7340

@daviditen mind reviewing?

ronawho:
Probably easiest to review ignoring whitespace: https://github.com/chapel-lang/chapel/pull/7346/files?w=1

cassella:
I get an error on this branch trying to use a virtual standalone
iterator that generates parallelism:

```chapel
class Parent {
  iter these(): int {  yield 1; }
  iter these(param tag: iterKind): int where tag == iterKind.standalone {
    coforall i in 2..4 {
      yield i;
    }
  }
}


class Child : Parent {
  iter these(): int {  yield 1; }
  iter these(param tag: iterKind): int where tag == iterKind.standalone {
    coforall i in 42..44 {
      yield i;
    }
  }
}


var o: Parent = new Child();

forall i in o {
  writeln(i);
}
```


On master, this test prints e.g.

```
3
2
4
```

but on this branch, I get

```
fortytwo@magrathea:~/src/chpl$ ./pariter -nl 1
internal error: compiler generated error
```

ronawho:
Interesting.. I guess I should have added a more complicated example than the one in virtual-par-iters.chpl

ronawho:
I added some more tests/futures in #7384. Unfortunately, they show that anything beyond the most basic iterators don't work even with this PR. 

The virtual/dynamic-dispatch iterator code was originally written for serial iterators only and it looks like it doesn't support parallel or distributed constructs (coforalls or on-stmts).

I'm going to close this PR and merge #7385, which at least issues a clear compiler error instead of an internal error or confusing runtime behavior

PR #7351, opened 2017-09-18
Use initial value of outer variable upon reduce intent

### LINKED ISSUE #6738: Reductions should perform a final combination with the initial value of the variable...
(Prompted by the following exchange on stackoverflow:
   https://stackoverflow.com/questions/45111312
)

Currently Chapel reductions ignore the external value of the variable being reduced  when setting the initial value of the reduction and they overwrite the variable on completion (instead of combining with this value). While the former behavior make sense, the latter can result in surprising results. For instance, the following code
```chapel
var x = 5;
for i in 1..3 {
  forall j in 1..10 with (+ reduce x) {
    x += 1;
  }
}
writeln(x);
```
returns 10, instead of 35. By contrast, a similar loop in OpenMP (thanks @bradcray) does return 35.
```c
#include <stdio.h>
#include <omp.h>

int main(int argc, char* argv[]) {
  int tot = 5;
  for (int i=0; i<3; i++) {
  #pragma omp parallel for reduction(+:tot)
    for (int j=0; j<10; j++) {
      tot += 1;
    }
  }
  printf("tot is: %d\n", tot);
}
```

This issue suggests that Chapel reductions should accumulate against the final value as a final step.

vasslitvinov:
Up to now, we had discarded the initial value of the outer variable used in a forall loop with a reduce intent. For example:

    var x = 7;
    forall 1..10 with (+ reduce x) {
      x += 1;
    }
    writeln(x);

had produced 10 instead of 17.

Responding to multiple user requests, and as discussed on the internal mailing list, this PR changes the behavior to take the initial value into consideration.

For user-defined reductions, this will be done by calling a method on the reduction class instance that is used in the forall loop. This call will be made before invoking the parallel iterator. The call will be:

    globalOp.initialAccumulate(outerVariable);

if it is available, otherwise it will be:

    globalOp.accumulate(outerVariable);

It is a compilation error if neither is available or applicable.

In the above example, "outerVariable" is "x".

I also adjusted the tests that relied on the old behavior.

Resolves #6738.

### DIFF
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -1866,8 +1866,53 @@ static void findOuterVarsNew(ForallStmt* fs, SymbolMap& outer2shadow,
 // markOuterVarsWithIntentsNew() and helpers
 //
 
+//
+// Incorporate the initial value of the outer variable:
+//   globalOp.initialAccumulate(outerVar)
+// If this does not resolve, instead call:
+//   globalOp.accumulate(outerVar)
+//
+static void insertAndResolveInitialAccumulate(ForallStmt* fs, BlockStmt* hld,
+                                            Symbol* globalOp, Symbol* outerVar)
+{
+  CallExpr* initAccum = new CallExpr("initialAccumulate", gMethodToken,
+                                     globalOp, outerVar);
+  if (hld)
+    hld->insertAtTail(initAccum);
+  else
+    fs->insertBefore(initAccum);
+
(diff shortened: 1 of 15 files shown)


vasslitvinov:
@e-kayrakli how does this look to you?

vasslitvinov:
P.S. Looking at test changes, it really makes me want the kind of reduce intent that gives me just the result of the reduction -- and infers its type so I don't have to declare it or figure out what it is.

PR #7358, opened 2017-09-18
Add a note on rebasing to ContributorInfo

### LINKED ISSUE #6748: Add section on rebasing to ContributorInfo.rst
When reviewing the update to ContributorInfo.rst, @mppf pointed out that the list of common git commands could use a section on `git rebase`.  I don't want to handle that right away, but I don't want to forget about it either, so log it here.

lydia-duncan:
Adds it under "How To Modify Git History", the section for things that
are useful, but may cause problems if the changes have already been
pulled.

Also, fix a couple of things I noticed while here - One of the sections
was missing a colon, and one of the sentences had "commands are may
cause", instead of just "commands may cause"

Resolves #6748

### DIFF
--- a/doc/rst/developer/bestPractices/ContributorInfo.rst
+++ b/doc/rst/developer/bestPractices/ContributorInfo.rst
@@ -440,7 +440,7 @@ If you want to understand the changes that occurred upstream, see
 How to modify git history
 +++++++++++++++++++++++++
 
-The following commands are **may cause problems** if the changes they overwrite
+The following commands **may cause problems** if the changes they overwrite
 have been pulled by other repositories.
 
 Fixing a commit message:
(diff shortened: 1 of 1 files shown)


doc/rst/developer/bestPractices/ContributorInfo.rst
```diff
@@ -449,12 +449,18 @@ Fixing a commit message:
...
 
 .. code-block:: bash
 
     git reset --soft HEAD~1
 
+Reapplying changes from the current branch onto an updated version of master:
+
+.. code-block:: bash
+
+    git rebase master
```
mppf: Let's make it be
```
git fetch upstream
git rebase upstream/master
```

lydia-duncan: (we continued conversation after the initial "looks good" and decided this would be useful too)

PR #7360, opened 2017-09-18
Adjust string pass/return test

mppf:
It looks like this test relies on some extreme amounts of optimization
from GCC that only reliably apply when the tested loops are in functions.
PR #7019 seemed to slow down the returning strings path, but it seems
that the reason was that a GCC optimization wasn't as effective with the
error-handling changes, due to the code pattern in the (seemingly
unrelated to the loops) waitEndCount. Moving the loops to their own
function lets GCC avoid confusion.

Additionally, this PR adds a 3rd line to the graph to measure the time
for constructing and returning a string created on the heap, since
strings referring to global constant strings behave slightly differently.

### DIFF
--- a/test/types/string/psahabu/perf/arguments.chpl
+++ b/test/types/string/psahabu/perf/arguments.chpl
@@ -3,34 +3,64 @@ use Time;
 config const n = 100000000;
 config const timing = true;
 
-// passing
 var pass = "passing a string";
 var acc = 0;
-var tPassing: Timer;
-if timing then tPassing.start();
-for i in 1..n {
-  acc += receive(pass);
+
+test();
+
+// passing
+proc testPassing():Timer {
+  var tPassing: Timer;
+  if timing then tPassing.start();
+  for i in 1..n {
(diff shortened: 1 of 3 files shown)


test/types/string/psahabu/perf/arguments.chpl
```diff
@@ -3,34 +3,64 @@ use Time;
...
-if timing then tPassing.start();
-for i in 1..n {
-  acc += receive(pass);
+
+test();
+
+// passing
+proc testPassing():Timer {
+  var tPassing: Timer;
+  if timing then tPassing.start();
```
psahabu: I wonder if we need the `if timing` bits in the function calls, because we already do the `if timing` check in `test()`.

psahabu:
Oh, it also might be worth adding the perf annotation now, although I don't know if it's possible to do future annotations like that.

bradcray:
I'm ready for this to go in and to re-run the release-over-release numbers for this test once it does.  Is anything holding it up?

PR #7364, opened 2017-09-18
Close memory leak for vertex coloring problem

bradcray:
Graph nodes needed to be freed at end of run.


bradcray:
Heads-up to @kushalsingh007 (who developed this) and @benharsh (who reviewed it).

PR #7365, opened 2017-09-18
Privatize domains and distributions for reindex array views

bradcray:
This privatizes the domain and distribution created for a reindex array view in the event that the thing being reindexed is privatized.

The work to implement the privatization was fairly simple and followed the pattern that was established for rank change array views.  The bigger part of this PR was changing references to domains to refer to the privatized domain when available.

Communication counts for the use of a reindex array view are reduced by this change (as shown by the assignReindex.chpl test results), though the comm counts for creating the reindex view increase slightly.  Comparing the communication counts for a reindex view to that of a rank-change view, it seems that there is still far more communication than there should be for the use of the reindex view, yet I did not put in a lot of effort to optimize it further at this point given how rarely reindex views are used in practice.


bradcray:
@benharsh: It's been awhile since I've tested this, so I'm running that while I type, but expect it to come back clean.  Would you have time to review this today?

PR #7366, opened 2017-09-18
Reduce number of memleaks in Jama testing

bradcray:
In a background activity to close memory leaks in our tests, I tried
to tackle those being generated by TestMatrix.chpl, but unfortunately
was only able to make a small dent in them because the
LinearAlgebraJama module itself allocates and returns a lot of classes
which the caller doesn't free in this test (and it also often
overwrites existing matrix pointers, effectively orphaning the
objects).  I didn't have the time to determine which matrices needed
freeing beyond the obvious (I tried to add a 'delete' for each 'new'
within the test itself).


bradcray:
Attn: @ct-clmsn and @ben-albrecht

bradcray:
Closing this PR in favor of #7539 and https://github.com/ct-clmsn/LinearAlgebraJama/issues/2.

PR #7369, opened 2017-09-18
Improve mason build process

benharsh:
This commit updates mason's Makefile to only build mason with:
CHPL_COMM=none
CHPL_LAUNCHER=none
CHPL_REGEXP=re2

It also unloads hugepages if the 'module' command is available. This is an attempt to detect that we're on a Cray, in which case we don't want or need hugepages.

Instead of using 'git -C', use ``chdir`` instead in order to be compatible with older versions of git.

$MASON_HOME/.mason will now be created on the first call to 'mason update' instead of at installation time.

### DIFF
--- a/tools/mason/MasonUtils.chpl
+++ b/tools/mason/MasonUtils.chpl
@@ -174,3 +174,12 @@ proc getChapelVersionStr() {
   return chplVersion;
 }
 
+proc gitC(newDir, command) {
+  const oldDir = here.cwd();
+  here.chdir(newDir);
+
+  runCommand(command);
+
+  here.chdir(oldDir);
+}
+
--- a/tools/mason/MasonUtils.chpl
+++ b/tools/mason/MasonUtils.chpl
@@ -75,18 +75,18 @@ proc depExists(dependency: string) {
 
 proc MASON_HOME: string {
   // possible locations
-  var masonHome = getEnv("MASON_HOME");
-  var home = getEnv('HOME');
-  if masonHome == '' {
-    if isDir(home + '/.mason') then
-      return home;
-    else {
-      writeln("Mason could not find MASON_HOME");
-      writeln("Consider setting MASON_HOME in your .bashrc");
-      halt();
-    }
+  const envHome   = getEnv("MASON_HOME");
+  const home      = getEnv('HOME');
+  const masonHome = if envHome != "" then envHome else home;
+
+  const dotMason = masonHome + "/.mason";
+
+  if isDir(dotMason) == false {
+    writeln(dotMason + " not found, creating...");
+    mkdir(dotMason);
   }
-  else return masonHome;
+
+  return masonHome;
 }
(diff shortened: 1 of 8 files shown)


benharsh:
Latest commit passes test/mason on darwin and linux64.

tools/mason/MasonUtils.chpl
```diff
@@ -75,18 +75,18 @@ proc depExists(dependency: string) {
...
-    }
+  const envHome   = getEnv("MASON_HOME");
+  const home      = getEnv('HOME');
+  const masonHome = if envHome != "" then envHome else home;
+
+  const dotMason = masonHome + "/.mason";
+
+  if isDir(dotMason) == false {
+    writeln(dotMason + " not found, creating...");
+    mkdir(dotMason);
```
ben-albrecht: Would it be more robust to do a `mkdir(dotMason, parents=true)` here?

tools/mason/README.rst
```diff
@@ -20,25 +20,18 @@ In the mason directory found in ``$CHPL_HOME/tools/mason`` run the following:
...
-dependencies. This installs mason in the same place as the chapel compiler (``chpl``)
-so that mason can be used anywhere in the user's file system.
+It builds the mason binary so that the command line interface can be used. This
+installs mason in the same place as the chapel compiler (``chpl``) so that
+mason can be used anywhere in the user's file system.
 
 
+The ``mason update`` command will download the registry to ``$MASON_HOME/registry``
+if it there is not already a registry at that location.
+
```
ben-albrecht: Would this be a good place to state the minimum `git` version supported? I guess determining that version is a task in itself..

benharsh: Probably, but I fixed the "git -C" issue so I think we can punt on it for now.

tools/mason/buildMason
```diff
@@ -0,0 +1,18 @@
...
+
+export CHPL_COMM=none
+export CHPL_LAUNCHER=none
+
+if command -v module > /dev/null 2>&1 ; then
+  echo "Unloading hugepages..."
+  module unload $(module list 2>&1 | \grep "craype-hugepage" | sed -e"s:[0-9]*)::g")
+fi
+
+if ! $chplBinDir/chpl mason.chpl ; then
```
ben-albrecht: Do we check that `chpl` is in the `$PATH` elsewhere in the module build, and if not, would it be valuable to check it and give a separate error message for that case?

benharsh: Hmm, I'm not sure. Currently we're planning on people running 'make mason' from $CHPL_HOME, so I feel like it should be OK to use CHPL_BIN_DIR in this way.

ben-albrecht: Ah, right. Catching this kind of error might be useful elsewhere (like a `make check` for `mason` or something), but definitely not necessary here.

PR #7371, opened 2017-09-18
Fix licm bug involving references

ronawho:
LICM isn't currently able to accurately track definitions through references,
so don't hoist references or things that alias references.

As a simple example for something like:

```chpl
var B = [10];
ref box = B[1];
box = 2;
```

licm doesn't detect that `box = 2` is a definition of B's ddata. In order to
solve this in a satisfying manner we'd have to teach licm to understand arrays
and significantly ramp up our analysis. That's a large undertaking and we're
hoping that eventually we'll be able to rely on llvm's LICM instead of our
version so for now just prevent hoisting of references or aliases to
references.

This results in a few more gets for allocating a block-dist array, but I don't
expect to see noticeable single or multi-locale performance regressions.

This resolves https://github.com/chapel-lang/chapel/issues/7042


ronawho:
@benharsh mind reviewing?

I didn't see any perf regressions in the single-locale playground or in a manual run of 16-node

PR #7372, opened 2017-09-18
Update whitespace in Collections modules, beef up warning

mppf:
* prominent note now says something like

 This package module is new in 1.16 and may contain bugs. The interface
 may change. The documentation is being incrementally revised and
 improved.

 * converted to unix line endings

 * removed space at the end of lines

Reviewed by @e-kayrakli - thanks!


mppf:
@ronawho - FYI - these modules are already in the chpldoc Makefile

@LouisJenkinsCS - FYI

mppf:
I used a vi command that is effectively equivalent to dos2unix. We don't have a policy about line-endings, but the rest are unix, so it helps it to fit in / be less unusual. I'd recommend using unix line endings in the future if you remember to.

mppf:
@e-kayrakli or @LouisJenkinsCS - could whomever gets to it first review this PR? The only meaningful change is the docs change in 93964ea3ecd54c7ebb9a3c614e18b75893f58bd8 - everything else is whitespace. Thanks.

e-kayrakli:
Looks good to me

PR #7373, opened 2017-09-18
Add skipif files to skip Crypto tests on non-linux64 and if OpenSSL is old

daviditen:
Some of the Crypto tests require OpenSSL 1.0.1 or newer, so skip them if
it isn't available.

OpenSSL isn't installed by default on most non-Linux platforms, so skip the
tests on non-linux64.

### DIFF
--- a/test/modules/packages/Crypto/saru.skipif
+++ b/test/modules/packages/Crypto/saru.skipif
@@ -1,0 +1,2 @@
+# skip non-linux64 platforms for now
+CHPL_TARGET_PLATFORM != linux64
--- a/test/modules/packages/Crypto/saru/kdf/pbkdf2-2.skipif
+++ b/test/modules/packages/Crypto/saru/kdf/pbkdf2-2.skipif
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+reqMajor=1
+reqMinor=0
+reqPatch=1
+
+version=`openssl version | sed -e 's/^OpenSSL \([^ ]*\) .*$/\1/'`
+major=`echo $version | sed -e 's/\..*$//'`
+minor=`echo $version | sed -e 's/.*\.\([^\.]*\)\..*$/\1/'`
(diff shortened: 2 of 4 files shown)


daviditen:
Thanks!

test/modules/packages/Crypto/saru.skipif
```diff
@@ -0,0 +1,2 @@
+# skip non-linux64 platforms for now
+CHPL_TARGET_PLATFORM != linux64
```
ronawho: Executable skipif that checks if openssl is installed?

test/modules/packages/Crypto/saru/kdf/pbkdf2-2.skipif
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+reqMajor=1
+reqMinor=0
```
ronawho: Can this and one of the other ones symlink to a real one?

test/modules/packages/Crypto/saru.skipif
daviditen: On OSX there is an OpenSSL installed, but not compatible with the homebrew version.  So it needs to be installed and be the homebrew one on Macs.  So I just said I'll skip non-Linux for now.

test/modules/packages/Crypto/saru/kdf/pbkdf2-2.skipif
daviditen: Sounds good.

test/modules/packages/Crypto/saru.skipif
ronawho: Ah ok.

In addition to skipping on non-linux then should we also skip if openssl inn't installed, or do we expect every linux system to have it?

daviditen: For now I feel fairly safe saying that Linux systems that we test on have a reasonably standard (though sometimes very old) OpenSSL installed.

PR #7375, opened 2017-09-18
Require deinit functions to have parentheses.

vasslitvinov:
Our group consensus has moved towards requiring deinit functions
to have parentheses, as a style preference. This PR implements that.
This is both for class/record deinitializers and for module deinit functions.

Luckily, normalize() already iterates over all the former and all the latter.
This PR adds a simple check to each.

We used to remove FLAG_NO_PARENS from all deinit functions in the parser.
This is so that deinit resolves the same whether with parens or not.
This PR removes that removal so that normalization can detect
paren-less deinits.

- [x] test under linux64 --verify


vasslitvinov:
@lydia-duncan would you review?

PR #7376, opened 2017-09-18
Remove Bug Comment from Collections Module for Virtual Function segfault

LouisJenkinsCS:
Removed the bug comment, also removed an extra white-space in `DequeParity.chpl`


mppf:
@LouisJenkinsCS - can you also add a .future test that shows the code that causes the internal compiler error with remove, so that one of us can more easily take a look? Thanks.

mppf:
I don't see any unit tests yet? Also if the bug is no longer there, do we still need "There are issues with :proc:`remove`, of which the cause is currently unknown, ..." ?

modules/packages/Collection.chpl
```diff
@@ -50,16 +50,6 @@
       fact with the fact that ``break`` ing out of a serial iterator will result in resource leakage, where destructors
       are not called, hence Collections using some RAII-based resource cleanup will end up leaking and potentially leaving
       the Collection in an undefined state. This has been documented under issue `#6912 <https://github.com/chapel-lang/chapel/issues/6912>`_ .
-  2.  There are issues with :proc:`remove` where the compiler will crash (segmentation fault) if the
-      return value is not captured at the callsite, documented as issue `#6542 <https://github.com/chapel-lang/chapel/issues/6542>`_ .
-      Lastly, there is yet another bug, of which the cause is currently unknown, that triggers a compiler internal error, which requires
-      the user to declare the express type at callsite. A 'safe' way to use :proc:`remove` in a loop is documented below.
```
mppf: Do you have any tests that use remove in the "unsafe" way ("unsafe" without the recent bug fix)?

LouisJenkinsCS: Hm... now that I think about it, no. I'm not sure of a valid reason to use `remove` without capturing and checking the results. Then again, I can see about adding it to a Deque unit test.

LouisJenkinsCS: I manually tested `remove` without capturing at callsite, it doesn't produce a segmentation fault in compiler.

PR #7377, opened 2017-09-18
Make two new .skipif files be symlinks to a third

daviditen:
PR #7373 added 3 identical .skipif files.  So that only one needs to be
updated if they change, remove two of them and make them instead symlinks.
As suggested by @ronawho.


ronawho:
👍  thanks

PR #7379, opened 2017-09-18
Update and extend the initializers technote

lydia-duncan:
While for last release, the emphasis was on "some things work, others don't",
the message for this release is "most things work except for these known
sticking points".  Hopefully, the final cut of this technote for the release
will be promotable to the spec next release, modulo some minor tweaks if we
change some of the rules with more experience.

With that in mind, change a little of the introductory paragraph, then add a
more robust description of the initializer method and its various rules, trying
to mimic the style of the spec's constructor section as much as possible.  I
remove the "Status" section in favor of a "Remaining Work" section (title can
change).  The "Remaining Work" section contains some more detailed callouts for
compiler generated initializers and the error handling support, minor notes on
noinit and the known remaining bugs, and then a list of other TODO items that
didn't fall under the previous categories.

Verified the output when looking at it via github.  I intend to look at the generated
docs output once the content has been finalized.

### DIFF
--- a/doc/rst/technotes/initializers.rst
+++ b/doc/rst/technotes/initializers.rst
@@ -287,6 +287,7 @@ Bugs
   initializer in its original module
 - nested types when the outer type and/or the inner type defines an initializer
   and the outer type and/or the inner type is generic.
+- others
 
 Other TODOs
 +++++++++++
@@ -295,3 +296,5 @@ Other TODOs
 - Improve some slightly cryptic error messages
 - Ensure we *always* error when a method is called in Phase 1 (we only sometimes
   do today)
+- Extend on statement support to allow field initialization within its bounds
+  after getting larger team buy in.
--- a/doc/rst/technotes/initializers.rst
+++ b/doc/rst/technotes/initializers.rst
@@ -9,8 +9,10 @@ initialization of variables with record type or instances of class
 type.  This approach relies on methods known as initializers rather
 than the original methods known as constructors.
 
-A discussion of the current design and rationale is provided in
-`CHIP 10 <https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/10.rst>`_.
+A discussion of the current design and rationale is provided in `CHIP 10`_.
+
+.. _CHIP 10:
+   https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/10.rst
 
 Release Chapel 1.16.0 provides a strong preview implementation of this
 new feature.  Though some known bugs remain, the feature is rapidly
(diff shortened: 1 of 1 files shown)


doc/rst/technotes/initializers.rst
```diff
@@ -12,123 +12,289 @@ than the original methods known as constructors.
...
 
-The functionality of an initializer method has four major aspects
+An initializer is a method on a class or record named "init".  It is invoked
+by the ``new`` operator, where the type name and initializer arguments are
+preceded with the ``new`` keyword.
 
-  - user-defined code that runs in phase 2
+If the program declares a type initializer method, it is a user-defined
+initializer.  If the program declares no initializers or constructors for a
+class or record, a compiler generated constructor for that type is created
```
noakesmichael: "a compiler generated constructor for that type" --> "a compiler generated initializer for that type"

doc/rst/technotes/initializers.rst
```diff
@@ -12,123 +12,289 @@ than the original methods known as constructors.
...
+-------------------------
 
-  - compiler generated code that runs in phase 1
+A user-defined initializer is an initializer method explicitly declared in the
+program.  An initializer declaration has the same syntax as a method
+declaration, except that the name of the function is "init", and there is no
+return type specifier.
 
-  - the generation of effective user-facing diagnostic messages
+When an initializer is called, the usual function resolution mechanism is
```
noakesmichael: I am unsure if "When an initializer is called," in this context might be confusing or not.  Some readers
may have skimmed the previous text that speaks of new expressions and uses the word "invoked"
and may fail to be clear that typical user code should not call init() directly.

doc/rst/technotes/initializers.rst
```diff
@@ -12,123 +12,289 @@ than the original methods known as constructors.
...
+applied to determine which user-defined initializer to invoke.
 
-in approximate order of robustness and maturity.
+In contrast to user-defined constructors for generic classes and records,
```
noakesmichael: I wonder if the whole issue of generic class/record is sufficiently complex/confusing that it deserves
a separate section?

doc/rst/technotes/initializers.rst
```diff
@@ -12,123 +12,289 @@ than the original methods known as constructors.
...
 
-We believe that initializer methods are ready for use in application code
-but we also expect that the implementation will mature significantly
-during the 1.15 release.  We encourage their use but recognize that they
-may prove to be less robust than we would like at the start of the release.
-If you do choose to experiment with this feature then please do not
-hesitate to seek help, report bugs, or offer suggestions.
+The code written in an initializer can be divided into two categories,
+referred to as "Phase 1" and "Phase 2" for the remainder of this document.
+When the phase division indicator is not present, the body of an initializer
```
noakesmichael: I wonder if there should be a summary of what phase1/phase2 are, and perhaps even a summary of the motivation, before talking about the "division indicator" etc?

doc/rst/technotes/initializers.rst
lydia-duncan: I'm describing the current state here, rather than the intended direction, so it should stay "constructor" for now.  Should I call out there that we intend to move to compiler generated initializers, rather than just having a reference to the section on them?

doc/rst/technotes/initializers.rst
lydia-duncan: That's fair.  I stole those lines directly from the constructor section of the spec, should I update it there as well?

doc/rst/technotes/initializers.rst
lydia-duncan: I'm definitely open to that.  I wasn't sure how much additional detail was needed on them, so I kept them close to the rest in an effort to imply their general similarity (e.g. both generic and concrete fields must be initialized in order, etc.).  Should I go in more depth on their differences?

lydia-duncan:
@noakesmichael - I am debating about whether to add more to the generics section about type constructors and the like.  I'm going to spend some time hammering out my changes to that section of the CHIP and see what that inspires.

PR #7381, opened 2017-09-19
Improve some "multiply-defined" errors

vasslitvinov:
Before this change the "multiply defined" error did not indicate
the conflicting definition when it was a function.
This change ensures that when the conflict is between a variable
and a function, the compiler points to the function as well.

I was affected by the lack of the function reference
when I compiled this test: users/npadmana/mpi/ring.chpl
whose extract I am also adding as
  test/modules/vass/multiply-defined-error-with-fn.chpl

To implement this, I created a separate function
that iterates over the symbols the second time, for clarity
and simplicity.

While there, I added "is" and switched to the more conventional
"note: also defined here".

- [x] test under linux64 --verify

### DIFF
--- a/compiler/passes/scopeResolve.cpp
+++ b/compiler/passes/scopeResolve.cpp
@@ -1164,6 +1164,23 @@ static void lookup(const char*           name,
 
                    std::vector<Symbol*>& symbols);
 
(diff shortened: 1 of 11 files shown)


dmk42:
Looks good to me.

Issue #7225, opened 2017-09-07
Simplify the construction of simple random numbers


buddha314:
In chatting with bHarsh, I learned that Chapel is designed to preserve a random sequence in forall loops.  (Don't ask me to go to deep.)  However, a simple coin flip is pretty important.  In Python, one can do

```
>>> import random
>>> random.random()
0.5547507267722159
```

Perhaps we can have two easier functions, one that is consistent across `forall` loops and one that is seeded locally.  Maybe

```
var x = prandom()  // Get the same sequence in each forall
var y = lrandom()  // A globally seeded random sequence, will not be consistent across tasks.
```

The different names could encourage careful usage.

ben-albrecht:
Implementation detail: I think this would involve instantiating a default `RandomStream` in the `Random` module, and providing helper functions simply called `random` or `lrandom` to get their next values.

Design detail: `RandomStreams` are instantiated with a type. What types would be supported by default? `real` (`0.0-1.0`) seems like a safe choice. Maybe bounded `ints` too?

mppf:
I'll note that RandomStreams don't *have* to use a type. E.g. with PCGRandomStream, there is not a problem getting random values for different types with each call.

I believe we shied away from having a global RandomStream in the past because it might not be as clear that there is a shared stream when many tasks use it. OTOH, a top-level global RandomStream created by the user is subject to the same issues.

Lastly, the use-case that brought this up is a program that generates a single random value. I'm suspicious of using time to seed an RNG and extract one value - seems like we're just using munged time. Likewise, in a distributed run, if the nodes all have their clocks set exactly, might we get the same values from different random streams on different nodes? For these reasons I think we should change the default seed to use /dev/urandom on Linux when it is available.

nspark:
This was why I wanted a module-level `deinit()` function that could free the module-level "convenience feature" `RandomStream` object.

bradcray:
Hi Nick --  Just to make sure we're on the same page, you remember that we have module-level deinit() now, right?  (I couldn't tell from the verb tense above).

nspark:
@bradcray Yes, I just never got around to writing such a convenience random function. It's feature-freeze today, right? 😉

mppf:
Note that there is a fair amount of discussion for the design of this in #7382.

vasslitvinov:
See also #12540

PR #7382, opened 2017-09-19
Add convenience functions for simple RNG use

nspark:
- [x] Add `getRandom` and `seedRandom` convenience routines to the `Random` module
- [ ] Documentation
- [ ] Tests

This PR attempts to close out #7225.

It's still Sept. 18... did I make "Feature Freeze"? :wink:

### DIFF
--- a/modules/standard/Random.chpl
+++ b/modules/standard/Random.chpl
@@ -87,7 +87,56 @@ module Random {
   }
 
   /* The default RNG. The current default is PCG - see :mod:`PCGRandom`. */
-  param defaultRNG = RNG.PCG;
+  config param defaultRNG = RNG.PCG;
+
+  private
+  var globalRandomStream:
+    if defaultRNG == RNG.PCG then PCGRandom.RandomStream(eltType=real, parSafe=true)
+    else NPBRandom.NPBRandomStream(eltType=real, parSafe=true);
+
+  {
+    var seedVal: uint(64) = 0;
+    try {
+      var f = open("/dev/urandom", iomode.r);
+      var c = f.reader();
+      c.readbits(seedVal, 62);
+      c.close();
+      f.close();
+    } catch e: Error {
+      halt("Error initializing ");
+    }
+    seedVal = (seedVal << 1) | 1:uint(64);
+    globalRandomStream =
+      makeRandomStream(eltType=real, seed=seedVal.safeCast(int(64));
+  }
+
+  /*
+    Docs go here
+   */
+  proc getRandom(type eltType = globalRandomStream.eltType) {
+    if !isNumericType(eltType) then
+      compilerError("Random.getRandom() only supports numeric element types");
+    if defaultRNG == RNG.PCG then
+      return globalRandomStream.getNext(eltType);
+    else {
+      if eltType != globalRandomStream.eltType then
+        compilerError("Random.getRandom() only supports eltType == real");
+      return globalRandomStream.getNext();
+    }
+  }
+
+  /*
+    Docs go here
+   */
+  proc seedRandom(seed: int(64)) {
+    delete globalRandomStream;
+    globalRandomStream = makeRandomStream(eltType=real, seed=seed);
+  }
+
+  pragma "no doc"
+  proc deinit() {
+    delete globalRandomStream;
+  }
 
   // CHPLDOC FEEDBACK: If easy, I'd suggest either deprecating the
   // :arg <type> <name>: form or else switching the order to


modules/standard/Random.chpl
```diff
@@ -87,7 +87,56 @@ module Random {
...
+      var c = f.reader();
+      c.readbits(seedVal, 62);
+      c.close();
+      f.close();
+    } catch e: Error {
+      halt("Error initializing ");
+    }
+    seedVal = (seedVal << 1) | 1:uint(64);
+    globalRandomStream =
+      makeRandomStream(eltType=real, seed=seedVal.safeCast(int(64));
```
ben-albrecht: `chpldoc` reports:

```
standard/Random.chpl:110: syntax error: near ';'
```

ben-albrecht:
> did I make "Feature Freeze"?

Sorry that we missed this one @nspark - it appears no one was eagerly looking for incoming PRs to review last night..

It looks like there is a failure in Travis testing that you'll need to resolve before moving forward.

Thanks for implementing this feature. I've wanted this for a while now!

modules/standard/Random.chpl
ben-albrecht: ah, missing a closing parens: `makeRandomStream(eltType=real, seed=seedVal.safeCast(int(64)));`

bradcray:
Design question: What are the implications for this PR on multi-locale programs?  Do they just grab the remote global random stream (and is that what we want?)

nspark:
@ben-albrecht That's odd about the missing close parenthesis. I must have added the `safeCast` change after I ran my tests.

> Sorry that we missed this one @nspark - it appears no one was eagerly looking for incoming PRs to review last night..

You mean submitting a PR late on the "due date" doesn't work like slipping one's paper under the professor's door? 😉 It's too bad about missing the feature freeze, but it's not a big deal to me.

@bradcray Good question! I haven't thought deeply about this feature and multi-locale programs. I'd guess that we'd want such a feature to be fast even in a multi-locale setting, however, there are probably better methods (e.g., `fillRandom`) when one needs lots of random numbers quickly. I can look at having a distributed array of `RandomStream` objects, where `getRandom` operates on the `here.locale`, but that feels a little heavier weight than I'd like.

@buddha314 Do you have any strong feelings about multilocale performance in any particular use-cases, given that you raised #7225?

Another design question is whether `fillRandom`, `shuffle`, and `permutation` should be updated to use the `globalRandomStream` object (introduced here) or continue to create their own streams via `makeRandomStream`.

buddha314:
@nspark Not at this time, no.  I haven't worked enough in a multilocale setting.  I am working up a few MCMC methods now and that might inform me.  At the moment, a simple random() would help me out.  Thanks for asking!

modules/standard/Random.chpl
mppf: I think it'd be worth explicitly passing `parSafe=true` here for emphasis.

modules/standard/Random.chpl
```diff
@@ -87,7 +87,56 @@ module Random {
...
+
+  private
+  var globalRandomStream:
+    if defaultRNG == RNG.PCG then PCGRandom.RandomStream(eltType=real, parSafe=true)
+    else NPBRandom.NPBRandomStream(eltType=real, parSafe=true);
+
+  {
+    var seedVal: uint(64) = 0;
+    try {
+      var f = open("/dev/urandom", iomode.r);
```
mppf: Not all platforms will have "/dev/urandom". So I'm not sure if it's reasonable to make any program using Random not run in such an environment. I'd consider printing out a warning and falling back on seeding with time if there is an error initializing.

Generally speaking, I think we should include a "/dev/urandom" and a "/dev/random" strategy in RandomSupport.SeedGenerator (which isn't necessarily a step you need to take for this PR). I think we should move away from the current default of using time in various Random procs, at least when "/dev/urandom" is available.

mppf:
> Design question: What are the implications for this PR on multi-locale programs? Do they just grab the remote global random stream (and is that what we want?)

I'd just like to elaborate on this question a little bit. We could:
 a. Make the global RandomStream have state per locale
 b. Make all locales share the global RandomStream

(And, if you want more to think about, as a thought experiment consider also the possibility of making each task or thread have its own RandomStream).

Each one of these has implications on what kind of programs would be correct/deterministic.
Either way though, I think we need to expose the globalRandomStream as a documented variable so that a user could call `globalRandomStream.fillRandom` on it (for example, as a way to efficiently fill in a distributed array while deterministically updating the state).

With a, if there is state per locale, we have the benefit that independent uses of the RandomStream don't interfere. (I.e. calls to getRandom() would be more efficient). But we have the drawback that something like `globalRandomStream.fillRandom` would only be deterministic if all the global RNG calls are on the same locale. So with that design, doing things like `globalRandomStream.fillRandom` is probably frowned upon, but might be the right move in certain circumstances. Additionally, it might get confusing if deterministic program behavior is required - for example, varying the number of locales might change program output.

With b, there might be more contention / interference when using the shared RandomStream. (I.e. calls to getRandom() on different locales would be slower). But the advantage is that `globalRandomStream.fillRandom` would have clear, deterministic, and reasonably efficient behavior.

(To make that efficient multi-locale - I'd expect we'd want to make a temporary copy of the RNG on the locales running fillRandom and use `skipToNth` on the global copy... but I think the implementation actually already does this.).

Anyway, my personal opinion on the matter is that I'm nervous about the global RandomStream because it would be easy to misuse it (say in a forall loop, setting array elements - both nondeterministic and inefficient). Using a strategy that makes `globalRandomStream.fillRandom` work well makes me feel better about the potential for misuse, since there is a pattern that we can document & support to do the same thing efficiently.

buddha314:
bump.

In writing some probability functions, I did this workaround:

```{chapel}
var u: [1..1] real;
fillRandom(u);
p = p * u[1];
```

ben-albrecht:
@nspark - Thanks for the initial effort. FWIW, I believe @bradcray considered himself the current blocker on this (annotated by his assignment).

> We could:
> a. Make the global RandomStream have state per locale
> b. Make all locales share the global RandomStream

> Anyway, my personal opinion on the matter is that I'm nervous about the global RandomStream because it would be easy to misuse it 

What if we provide both options to users, with explicit variable names, such as `localRandomStream` and `globalRandomStream` to help inform users of what they're opting into?

I would really like to see this feature added, as I continually find myself implementing something like this in projects that need random numbers generated across multiple modules, but do not need a separate random stream per module.

PR #7383, opened 2017-09-19
Mason installation should not use absolute paths

benharsh:
Use a relative symbolic link for mason instead of an absolute path in case $CHPL_HOME is moved around.


benharsh:
@ben-albrecht : I don't know if this is the best solution, but I'm going with it for now so we can test mason with the nightly build (which moves CHPL_HOME).

Issue #5760, opened 2017-03-21
Some data parallel statements are serialized with warning "X has been serialized"


daviditen:
**Note:** If you found your way to this issue via a compiler warning about scans being serialized, please see issue #12482 instead.

### Summary of Problem
Some data parallel statements that are meant to be parallelized in Chapel are serialized with a warning message "X has been serialized".  For example:

1. Scans are always serialized.
2. Assignments from ranges to multidimensional arrays are always serialized.
3. Parallel iteration over opaque domains and arrays are always serialized.

**Associated Future Test(s):**
arrays/deitz/parallelism/test_scan_is_parallel.future

LouisJenkinsCS:
Just to check in, is there an active effort to parallelize scans?

daviditen:
I don't think there is currently an active effort on parallelizing scans.  If this is something you'd like us to prioritize (or you'd like to work on it yourself!) a good first step would be to open an issue about it with a link back to this issue.

bradcray:
Hi @daviditen: I've recently opened more specific issues on parallelizing scans (e.g., issue #12482) and am not seeing other cases where the compiler or modules emit serialization warnings any longer, so am inclined to close this issue in favor of more specific ones (after changing the compiler's warnings to refer to the new issue for scans).  Let me know if you disagree with this plan.

daviditen:
@bradcray, That sounds fine to me.

bradcray:
I'm closing this issue because:

* issue #12482 now describes the situation for parallel scans more directly
* We're no longer generating a warning for assigning ranges to multidimensional arrays, and I'm not convinced the language should support this case anyway (see issue #12865).
* opaque domain/array iteration is parallel

PR #7385, opened 2017-09-19
Add a check for virtual parallel iterators

ronawho:
Issue a compiler error if virtual parallel iterators are used. They are
currently not working, so issue a clear compiler error instead of either
runtime failures or uninterpretable compiler errors.

### DIFF
--- a/compiler/resolution/lowerIterators.cpp
+++ b/compiler/resolution/lowerIterators.cpp
@@ -139,6 +139,24 @@ static void nonLeaderParCheck()
   USR_STOP();
 }
 
+static bool isVirtualIterator(Symbol* iterator) {
+  Vec<Type*> children = iterator->type->dispatchChildren;
+  return !((children.n == 0) || (children.n == 1 && children.v[0] == dtObject));
+}
+
+static void parallelIterVirtualCheck() {
+  forv_Vec(BlockStmt, block, gBlockStmts) {
+    if (isAlive(block) && block->isForLoop()) {
+      ForLoop* forLoop = toForLoop(block);
+      Symbol* iterator = toSymExpr(forLoop->iteratorGet())->symbol();
+      FnSymbol* ifn = getTheIteratorFn(iterator);
+      if (ifn->hasFlag(FLAG_INLINE_ITERATOR) && isVirtualIterator(iterator)) {
+        USR_FATAL_CONT(forLoop, "virtual parallel iterators are not yet supported (see issue #6998)");
+      }
+    }
+  }
+  USR_STOP();
+}
 
 static void nonLeaderParCheckInt(FnSymbol* fn, bool allowYields)
 {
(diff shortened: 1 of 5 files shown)


ronawho:
@daviditen mind reviewing? (Note that I still need to do testing)

@bradcray is it ok/preferred to put issue numbers in error messages? We do this for serialized scans but nothing else: `scan has been serialized (see issue #5760)`

compiler/resolution/lowerIterators.cpp
```diff
@@ -139,6 +139,24 @@ static void nonLeaderParCheck()
   USR_STOP();
 }
 
+static bool isVirtualIterator(Symbol* iterator) {
+  Vec<Type*> children = iterator->type->dispatchChildren;
+  return !((children.n == 0) || (children.n == 1 && children.v[0] == dtObject));
```
daviditen: How would dtObject ever be a child of of an iterator type?

daviditen: I see that similar checks have been around for a long time, but it seems very strange.

compiler/resolution/lowerIterators.cpp
```diff
@@ -139,6 +139,24 @@ static void nonLeaderParCheck()
...
+}
+
+static void parallelIterVirtualCheck() {
+  forv_Vec(BlockStmt, block, gBlockStmts) {
+    if (isAlive(block) && block->isForLoop()) {
+      ForLoop* forLoop = toForLoop(block);
+      Symbol* iterator = toSymExpr(forLoop->iteratorGet())->symbol();
+      FnSymbol* ifn = getTheIteratorFn(iterator);
+      if (ifn->hasFlag(FLAG_INLINE_ITERATOR) && isVirtualIterator(iterator)) {
```
daviditen: A comment saying that FLAG_INLINE_ITERATOR implies that the iterator is parallel or something like that would be helpful here.

ronawho: I have a branch that I plan on merging post-release that changes `FLAG_INLINE_ITERATOR` to `FLAG_PARALLEL_ITERATOR`.

compiler/resolution/lowerIterators.cpp
ronawho: I was similarly confused by this, but I wanted to limit the scope of this change since we're passed feature freeze. I have an untested branch that just converts this to `return children.n > 0`, which is also what we do in other places in the compiler including later in `expandForLoop()` and in `handlePolymorphicIterators()`

Any objections to sticking with what was there for now?

daviditen: No objection.

daviditen:
I think that having the issue number in the error message is fine.

daviditen:
Assuming this has passed a paratest I approve.

bradcray:
> is it ok/preferred to put issue numbers in error messages? We do this for serialized scans but nothing else

Just my opinion: I think it's OK and possibly preferred in cases where the error message is intended as a short-term placeholder particularly when there's more that we'd like to say beyond what the error message can say itself in a reasonable amount of space.  I'm not sure if the serialized scans item meets that latter definition presently (the issue doesn't add much, though it does still log the intention to improve it in the future and for users to comment on the issue?).

daviditen:
The serialized scan one came from retiring `STATUS`.  It used to say:
```
scan has been serialized (see note in $CHPL_HOME/STATUS)
```

And `STATUS` had the text that is in issue #5760.

Issue #7331, opened 2017-09-15
error handling: Error leaks with preemptive throws


psahabu:
### Description

Every `Error` in the Chapel error handling system is a class, but to hide this implementation detail from the user, the compiler inserts deletes for the handled error at the end of every catch block.

``` chapel
try {
  throw new SomeError();
} catch e {
  writeln("handling error " + e.msg);
  // compiler inserts delete e
}
```

This works even in cases when the handled error is thrown from the catch block, as the delete is dead code eliminated in a later pass. However, the following pattern leaks the handled `Error`:

``` chapel
try {
  throw new SomeError();
} catch e {
  writeln("handling error " + e.msg);
  throw new DifferentError();
  // delete e is dead code eliminated
  // e is never deleted
}
```

### Solutions

The current compiler provides no easy solutions because there is no way for `e` to be intelligently deleted, short of inspecting every statement in the `catch` block for a `throw` or a throwing call. This is a motivating case for a stronger error memory management story, be it errors as records instead of classes, or Shared/Owned.

### Affected tests
- `errhandling/psahabu/throw-from-catch`
- `errhandling/psahabu/throw-inner-catch`
- `errhandling/psahabu/unmarked-call-catch`

mppf:
@psahabu - I'm not convinced that there's no easy solution. What if we simply put `delete e` in a DeferStmt at the start of every catch block body?

psahabu:
Closed by #7387.

PR #7387, opened 2017-09-19
compiler inserts deferred delete error into catch statements

psahabu:
## Motivation

When a new error is thrown from a catch block, the old error is leaked. This bug is detailed in #7331.

``` chapel
try {
  throw new SomeError();
} catch e {
  writeln("handling error " + e.msg);
  throw new DifferentError();
  // delete e is dead code eliminated
  // e is never deleted
}
```

### deferred delete

Instead of inserting `delete e` at the end of the catch block, it is inserted at the beginning of the catch block as a `DeferStmt`:

``` chapel
} catch e {
  // compiler inserted
  defer {
    if e then delete e;
  }
  throw new DifferentError();
}
```

### nil-out error variable if thrown

The deferred delete is effective, but `e` should not always be deleted:

``` chapel
} catch e {
  defer {
    if e then delete e;
  }
  // psuedo-code lowering of throw
  var thrownError = cleanError(e);
  throw thrownError;
  // DeferStmt implementation causes double free of error
}
```

To pre-empt this possibility, the compiler will insert checks before throws in catch blocks and nil-out `e`:

``` chapel
} catch e {
  defer {
    if e then delete e;
  }
  var thrownError = cleanError(e);
  if thrownError == e then e = nil;
  throw thrownError;
  // DeferStmt does not free e
}
```

### note

The solution could be simplified if `delete nil;` were a no-op rather than a fatal error. In that case, the defer block would not need a conditional check.

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -21,6 +21,7 @@
 
 #include "AstVisitorTraverse.h"
 #include "CatchStmt.h"
+#include "DeferStmt.h"
 #include "ForLoop.h"
 #include "driver.h"
 #include "stmt.h"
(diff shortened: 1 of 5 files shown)


compiler/passes/errorHandling.cpp
```diff
@@ -381,6 +390,27 @@ bool ErrorHandlingVisitor::enterCallExpr(CallExpr* node) {
...
 
+    if (!catchesStack.empty()) {
+      Expr*      parent      = throwBlock;
+      CatchStmt* parentCatch = NULL;
+      while (parentCatch == NULL) {
+        parent      = parent->parentExpr;
+        parentCatch = toCatchStmt(parent);
+      }
+
+      if (DefExpr* catchFilter = parentCatch->expr()) {
```
mppf: This can be tricky, because we will cast the error variable (and create temporaries in the process). Setting one of these to nil won't affect the other copies of it. Can you make sure to add some tests that exercise this code? Are those tests already there? Or are they not yet in this PR for some reason?

psahabu: I've added `nil-check-catchall` and `nil-check-filter`, which exercise the code inside the conditional for named catchalls and filtered catches respectively. `throw-from-catch`, `throw-inner-catch`, and `unmarked-call-catch` ensure that the conditional is not entered for unnamed catchalls.

mppf: Cool. I think this PR is ready now, thanks.

PR #7389, opened 2017-09-19
Fix whitespace in mason's default .chpl file

benharsh:
Mason's default .chpl file should use two spaces, not three.

TODO: should we add an option to let users specify tabs/4-spaces/2-spaces?

Testing:
- [x] test/mason
  - [x] OS X
  - [x] linux


ben-albrecht:
> TODO: should we add an option to let users specify tabs/4-spaces/2-spaces?

Seems good to push people towards (what I interpret as) the standard convention of 2-spaces..

PR #7390, opened 2017-09-19
Fix --llvm test failures for Crypto module

mppf:
* EVP_MD_CTX_init is documented to return void, so change its return type
* Adjusted the LLVM backend so this type of mismatch won't cause an error if the returned value is never used. (See below for more on this).
* Adds a test to lock in this behavior. This test would fail under `--llvm` without the above adjustment.

About the type mismatch we are tolerating:

With the C backend, if we are wrong about the return type of a function declared with `extern proc`, nothing bad comes of it unless a returned value is actually used.

This commit adjusts --llvm code generation to behave similarly and adds a test to verify the behavior. This is an alternative way to fix the issue with EVP_MD_CTX_init.

- [x] verified functionality on Mac OS X with `start_test test/modules/packages/Crypto/saru/* --compopts '-I  /usr/local/opt/openssl/include -L /usr/local/opt/openssl/lib'`
- [x] full local testing

Reviewed by @dmk42 - thanks!

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -3418,7 +3418,13 @@ GenRet CallExpr::codegen() {
 #ifdef HAVE_LLVM
       // We might have to convert the return from the function
       // if clang did some structure-expanding.
-      if (this->typeInfo() != dtVoid) {
+
+      bool returnedValueUsed = false;
+      if (CallExpr* parentCall = toCallExpr(this->parentExpr))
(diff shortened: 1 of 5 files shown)


saru95:
I agree with this change. As `EVP_MD_CTX_init` is a macro for `EVP_MD_CTX_reset` which seems to be undocumented and `EVP_MD_CTX_init` is defined with a return type of `void`.

mppf:
@dmk42 - could you have a quick look at this? Thanks.

PR #7391, opened 2017-09-19
Update the module search path technote

lydia-duncan:
While looking through technote examples, I noticed that this document was out of
date; specificially, it was missing the package module directory.  This change
fixes that oversight.

Details:
- Brad asked that I change the word "README" to something else in the opening
  paragraph
- change the entry for standard modules in the list of module search paths to
  include the packages directory.
- For maintainability, at Brad's recommendation, replace the individual
  directories in the "internal" module list with a general mention.

### DIFF
--- a/doc/rst/technotes/module_search.rst
+++ b/doc/rst/technotes/module_search.rst
@@ -3,8 +3,8 @@
 Module Search Paths
 ===================
 
-This README describes how modules are located by the current Cray implementation
-of Chapel.
+This document describes how modules are located by the current Cray
+implementation of Chapel.
 
 Internal modules are always loaded by a chapel program *as if* the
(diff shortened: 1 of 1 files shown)


daviditen:
> Change the note about user modules overriding a standard module to mention
> both standard and package modules, and for the sake of conciseness replace
> later mentions with the more general "Chapel provided" term.

I'd be inclined to leave it as just "standard".  The bullet list above includes `packages` in the "standard" category.  If you really think it needs to be separately called out, it should probably be "the standard, package, layout, or dist version".

daviditen:
Looks good otherwise.

lydia-duncan:
Good point, thanks David!

Issue #5003, opened 2016-12-10
comm=ugni internal error (transfer too large?)


gbtitus:
### Summary of Problem

An internal user (Chapel team member) reports an internal error in a system library called from the ugni comm layer, in a program that attempts to transfer 10\*\*30 ints between top-level locales in a single operation.  The root cause has not yet been diagnosed, but it wouldn't be a bit surprising if this larger transfer needed to be broken up into smaller chunks, for the network to be happy.

### Steps to Reproduce

**Source Code:**

Use the existing Chapel ISx test with large enough parameters, on a Cray XC with CHPL_COMM=ugni and a craype-hugepages module loaded.  The heap has to be on hugepages because if it is not, there is already code to do the transfer using bounce buffers, breaking it up as needed, and the bounce buffers are much smaller (megabytes) and already known to work.

mppf:
@gbtitus - was this fixed by PR #5009?

gbtitus:
No, one can still build a program that has the problem -- you just need the array big enough that it uses eough of the heap to span the boundary at which we did an mprotect().  I'll fix this remnant problem later.

mppf:
That's technically true but I'm not sure it's true in practice. In particular, a Chapel user doesn't know how much of the Chapel heap is used by the runtime. So the program might fail if you get within 8MB of the total heap to be allocated, either way.
That's assuming that the last 8MB in the heap is allocated when the rest is, which isn't necessarily true...

gbtitus:
If the registered heap spans more than one kernel memory segment, and thus more than one kGNI memory segment, and thus more than one uGNI memory region, then it CAN happen that a comm-ugni transfer spans a region boundary.  Since we know this is possible and an internal error will result, it's appropriate to have an issue for it.

ronawho:
We think the heap spanning multiple kernel memory segments was resolved with dynamic registration, right?

There's still an issue where we're not breaking up large puts/gets into FMA allowable bite chunks

Here's a simple reproducer (which I'll add as a future soon):

```chapel
use Memory;

type elemType = uint(8);
config const memFraction = 3;
config const numMB = here.physicalMemory(unit = MemUnits.MB) / memFraction;

const n = (numMB * 1024 * 1024) / numBytes(elemType);


var A: [1..n] elemType;
on Locales[1] {
  var B: [1..n] elemType;
  B = A;
}
```

```
chapel (master) $ ./f -nl2 --numMB=128
[...]

PR #7394, opened 2017-09-19
Split large transfers in comm=ugni.

gbtitus:
The Gemini and Aries NICs cannot do GETs or PUTs of more than 1gb at a
time, and the uGNI interface enforces this.  Here, if we're asked to do
transfers larger than that, split them into smaller chunks that are
legal.  Add a test (based on one from @ronawho ) to check that this
works.

While here, also add handling for GETs larger than 4kb when we don't
have a hugepage module loaded.  There was already support for this in
the PUT code, but the GET side had never been done.

This resolves https://github.com/chapel-lang/chapel/issues/5003.


test/runtime/gbt/comm/bigBulkTran.skipif
```diff
@@ -0,0 +1 @@
+CHPL_COMM != ugni
```
ronawho: I think is a valuable "stress" test regardless of config (I think we should run it with gasnet and even comm=none)

test/runtime/gbt/comm/bigBulkTran.compopts
```diff
@@ -0,0 +1 @@
+--fast
```
ronawho: Why `--fast`? We're trying to avoid throwing --fast in compopts if possible. If there's some value in it I'm not opposed. If it's because bounds-checking slows this test down you could either just throw `--no-bounds-checks` or you could rewrite your initializer loops to be:

```chpl
forall (a,  i) in zip(A, A.domain) do a = i;
```

test/runtime/gbt/comm/bigBulkTran.chpl
```diff
@@ -0,0 +1,41 @@
+use Time;
+
+type elemType = uint;
```
ronawho: super minor and it shouldn't matter, but I would prefer to use `int` since I imagine way more user arrays will be ints vs uints.

test/runtime/gbt/comm/bigBulkTran.chpl
```diff
@@ -0,0 +1,41 @@
...
+
+  const startTime = getCurrentTime();
+  if doGET then
+    B = A;
+  else
+    A = B;
+  const elapsedTime = getCurrentTime() - startTime;
+
+  if verify then
+    writeln(if && reduce [i in 1..n by verifyStride] B(i) == A(i) then
```
ronawho: I find this really hard to read (not super important since it's a test)

test/runtime/gbt/comm/bigBulkTran.execopts
```diff
@@ -0,0 +1,2 @@
+--doGET=true
+--doGET=false
```
ronawho: I think it'd be worth running this test at a couple different sizes. Maybe 512, 1024, 2048, and possibly 1/3 of available mem (although that will probably increase the running time significantly)

test/runtime/gbt/comm/bigBulkTran.chpl
gbtitus: I mainly just wanted the default elemType to be big enough to hold the full initializer values in order to increase the odds of catching mis-copies.  So `int` is fine with me.

test/runtime/gbt/comm/bigBulkTran.compopts
gbtitus: Yes, it's just for execution time reduction.  I'll switch to `--no-bounds-checks`; it seems to work nearly as well.

test/runtime/gbt/comm/bigBulkTran.chpl
gbtitus: I could adds parens around the `&& reduce ...` expr.  Just kidding!  Just kidding!  Will simplify.

test/runtime/gbt/comm/bigBulkTran.execopts
gbtitus: Whether that's worthwhile would seem to depend on the purpose of the test.  If it's just a test that we're dealing with the Gem/Ari limitations properly then I think 2gb and running only on comm=ugni is sufficient.  (This was my intent.)  If we want it to be more of a general test then running it with other comm layers is appropriate, and we probably would at least want different sizes for each comm layer, depending on how their logic flowed.

ronawho: Even if we only run it for ugni, I think it's worth running at "boundary sizes" to make sure things work as expected (that said I think it's worth running for other configs too)

gbtitus:
@ronawho , would you mind taking a quick look at the followup changes?

PR #7395, opened 2017-09-19
Delete some user-allocated classes in stencil.chpl

bradcray:
This change closes some user-allocated memory leaks in v2/stencil.chpl
and in ref-intent-bug-2big.chpl which appears to be a code clone of
the same.

At this point, these tests are only leaking what appear to be domain
map-related data structures.


bradcray:
Attn: @vasslitvinov and @benharsh.

vasslitvinov:
Looks good to me.

PR #7398, opened 2017-09-19
Fix bulk-transfer rank change compilation failures

benharsh:
This PR fixes various compile-time errors due to rank mismatches. A new test has been added to test the cross product of rank changes between default, Block, and Cyclic distributions.

Testing:
- [x] full gasnet
- [x] dist-cyclic
- [x] dist-replicated

### DIFF
--- a/modules/dists/CyclicDist.chpl
+++ b/modules/dists/CyclicDist.chpl
@@ -1039,11 +1039,11 @@ proc CyclicArr.doiBulkTransferFrom(Barg, viewDom)
   if debugCyclicDistBulkTransfer then
     writeln("In CyclicArr.doiBulkTransferFrom()");
 
-  if this.rank == Barg.rank {
-    const Dest = this;
-    const Src = chpl__getActualArray(Barg);
-    const srcView = chpl__getViewDom(Barg);
-    type el = Dest.idxType;
+  const Dest = this;
+  const Src = chpl__getActualArray(Barg);
+  const srcView = chpl__getViewDom(Barg);
+  type el = Dest.idxType;
+  if this.rank == Src.rank {
     coforall i in Dest.dom.dist.targetLocDom do // for all locales
       on Dest.dom.dist.targetLocs(i)
       {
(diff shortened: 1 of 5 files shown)


benharsh:
@bradcray : this should resolve the easily-visible rank-change compile-time failures for default, Block, and Cyclic dists. As I mentioned before, a more general solution will be included in the bulk-transfer API redesign.

modules/dists/CyclicDist.chpl
```diff
@@ -1069,7 +1069,7 @@ proc CyclicArr.doiBulkTransferFrom(Barg, viewDom)
           if debugCyclicDistBulkTransfer then
             writeln("B[",(...r1),"] ToDR A[",regionDest, "] ");
 
-          Barg._value.doiBulkTransferToDR(Dest.locArr[i].myElems[regionDest], Barg.domain[(...r1)]);
+          Barg._value.doiBulkTransferToDR(Dest.locArr[i].myElems[regionDest], {(...r1)});
```
bradcray: Can you tell me about this change?  (What was wrong before, what it fixes)

benharsh: Barg.rank is not necessarily equal to r1.size, and would cause a compilation failure under certain conditions. We already pass ``{(...r1)}`` in BlockDist, so doing the same here felt safe.

test/distributions/robust/arithmetic/collapsing/rankChangeAssign.chpl
```diff
@@ -0,0 +1,114 @@
...
+config const debug = false;
+
+proc DEBUG() { if debug then writeln(); }
+proc DEBUG(const args ...?n) {
+  if debug then writeln((...args));
+}
+
+test(DistType.default);
+test(DistType.block);
+test(DistType.cyclic);
```
bradcray: Just to make sure I"m understanding the high-level approach here:  This test explicitly tests default, block, and cyclic and then relies on this test being run against those same distributions (plus others... replicated?) as part of the robustness testing suite to get the cross product, is that right?

bradcray:
The bulk of this PR looks like the same pattern that was applied in #7285 to additional contexts that weren't covered in that PR, is that right?

test/distributions/robust/arithmetic/collapsing/rankChangeAssign.chpl
benharsh: Yes, it relies on different ``distType`` config param values being run in different testing configurations.

benharsh:
Correct, it's mostly the same pattern.

bradcray:
OK, looks good to me, thanks!

PR #7400, opened 2017-09-19
Squash most mason leaks

benharsh:
In the TOML package, do not needlessly create Toml instances when we can just change the tag.

In mason-update:
- delete dependencies as they are processed
- do not clear a dependency's children (therefore leaking them) by assigning an empty associative array to an existing Toml

Testing:
- [x] correctness + valgrind
  - [x] modules/packages/Toml
  - [x] mason/

### DIFF
--- a/modules/packages/TOML.chpl
+++ b/modules/packages/TOML.chpl
@@ -350,37 +350,74 @@ pragma "no doc"
 
  pragma "no doc"
  proc =(ref t: Toml, s: string) {
-   t = new Toml(s);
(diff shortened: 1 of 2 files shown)


benharsh:
@bradcray : I'll get someone else to review this, but this should significantly drop the leaks associated with mason testing from 10s of thousands to 'just' hundreds of bytes. The remaining leaks are BaseRectangularDom and io buffer stuff.

bradcray:
Sweet!

PR #7401, opened 2017-09-19
Remove dont_use_exit_use_chpl_exit_instead

ronawho:
The macro `define exit dont_use_exit_use_chpl_exit_instead` was used to prevent
accidental calls to exit instead of chpl_exit. Unfortunately this sort of macro
definition is error-prone because it replaces every `exit` token, not just
function calls. This is causing compilation failures with new versions of clang
on OSX with an error along the lines of:

```
 no member named 'dont_use_exit_use_chpl_exit_instead' in the global namespace
 using ::exit;
       ~~^
 ../../../../include/chplrt.h:41:17: note: expanded from macro 'exit'
 #define exit    dont_use_exit_use_chpl_exit_instead
```

This just removes the macro, because we don't believe it's very useful anymore.
We could add a lookForBadExit script (similar to ./util/devel/lookForBadAlloc
that replaced the similar macros for the system allocator) but that seems like
overkill. Calling the wrong allocator can result in undefined behavior and
result in all sorts of nasty bugs, while calling exit has no such problems.
There's also no cases where we've accidentally added exit() or cases where we
needed to call the real exit, so it doesn't seem worthwhile to check for calls.

### DIFF
--- a/runtime/include/chplrt.h
+++ b/runtime/include/chplrt.h
@@ -38,8 +38,6 @@
 
 #define _noop(x)
 
-#define exit    dont_use_exit_use_chpl_exit_instead
-
 #endif // LAUNCHER
 
 #endif
--- a/runtime/src/chplexit.c
+++ b/runtime/src/chplexit.c
@@ -30,8 +30,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-#undef exit
-
 static void chpl_exit_common(int status, int all) {
   fflush(stdout);
   fflush(stderr);


ronawho:
@bradcray any objections to just removing the macro (you originally added it in fdf3f01c9b2)

bradcray:
If it's not difficult to extend our existing checks, I'd prefer to guard against direct exit() calls getting introduced.  For multi-locale runs in particular, I think it would be unfortunate if exit()s crept into the runtime without tearing down the comm layer, bypassing gdbShouldBreakHere, etc.  I also like forcing the developer to decide between "all locales are going to be calling exit in this case" vs. "we can't guarantee that."

Rather than adding a new script, can we add it to the existing one, potentially (eventually) renaming it to `lookForBadRTCalls` or something like that?

ronawho:
Replacing with https://github.com/chapel-lang/chapel/pull/7431

PR #7405, opened 2017-09-20
Keep deprecated LayoutCSR documented

ben-albrecht:
LayoutCSR was deprecated for LayoutCS (#6936), and its documentation was removed. It was later decided that we should keep the deprecated module documentation up for 1 release cycle.

I did not add `LayoutCSR` back to the list of modules for the docs post-processing script because I want the `type CSR` documentation shown.

Also, added a note about deprecation policies for future chapel devs.


daviditen:
Looks good.

PR #7410, opened 2017-09-20
Fix write cptr baseline

mppf:
Follow-on to PR #7254.

This PR adds some documentation and moves the nil-check to _write_one_internal.

Checking for nil in the writeThis method on c_ptr causes
runtime failures in --baseline (from checking that the `this`
for the called method is not `nil`, but it is).

Passed full local testing and --baseline testing for the test that failed.


mppf:
Verified the failing test works with --baseline, and full local testing is still running (but I'm not expecting any issues).

PR #7419, opened 2017-09-20
Update Cray platform documentation.

gbtitus:
Update the Cray platform documentation to reflect changes for configs
with CHPL_COMM=ugni and a hugepage module loaded.  Mainly the changes
have to do with array allocation and its side effect, the SIGBUS that
can result when we run out of memory allocating an array.  I also did
some minor rephrasing, changed `2^10` (bc(1) syntax?) to `2**10` (Chapel
syntax) for KiB etc., and (I hope) improved the discussion about why and
how to select a craype-hugepages module.  While here I also added a
mention of the Aries 64-bit real add disability, for completeness.

### DIFF
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -369,33 +369,40 @@ Controlling the Heap Size
 
 The "heap" is an area of memory used for dynamic allocation of
 everything from user data to internal management data structures.
-When running on Cray XC/XE/XK systems using either of the following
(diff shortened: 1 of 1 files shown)


gbtitus:
Changes to be made based on IM with @ronawho:

- Remove 1.16 and 1.17 version numbers and just refer to current and upcoming releases.
- Move the SIGBUS-on-OOM discussion down to the Known Constraints and Bugs section.

PR #7420, opened 2017-09-20
Add basic 'mason search' functionality

benharsh:


ben-albrecht:
Since the `_MasonTest*` packages are intended for testing purposes only, I think it'd be valuable to only show them in `mason search` if some sort of debugging flag is turned on. What do you think?

tools/mason/MasonSearch.chpl
```diff
@@ -0,0 +1,78 @@
...
+    masonSearchHelp();
+    exit(0);
+  }
+
+  updateRegistry("", args);
+
+  consumeArgs(args);
+
+  // If no query is provided, list all packages in registry
+  const query = if args.size > 0 then args.tail().toLower()
```
ben-albrecht: What are your thoughts on making the default search behavior follow *smart-casing*?

if `query.isUpper()` then `ignorecase=false`, otherwise follow the same logic you have.

benharsh: I'd prefer to keep it simple for this first pass.

benharsh:
How about 'mason search' hides packages starting with an underscore by default?

tools/mason/MasonSearch.chpl
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+proc findLatest(packageDir) {
+  var ret = new VersionInfo(0, 0, 0);
+  const suffix = ".toml";
+
+  for fi in listdir(packageDir, files=true, dirs=false) {
+    assert(fi.endsWith(suffix));
+    const end = fi.length - suffix.length;
```
ben-albrecht: Just a thought - [Path](http://chapel.cray.com/docs/master/modules/standard/Path.html) should really have procedures for extracting / dropping the file extension...

tools/mason/MasonSearch.chpl
```diff
@@ -0,0 +1,78 @@
...
+
+  // If no query is provided, list all packages in registry
+  const query = if args.size > 0 then args.tail().toLower()
+                else ".*";
+
+  const searchDir = MasonUpdate.getRegistryDir() + "/Bricks/";
+  const pattern   = compile(query, ignorecase=true);
+
+  var results : [1..0] string;
+  for dir in listdir(searchDir, files=false, dirs=true) {
```
ben-albrecht: Is this searching all packages or only the subset of packages that supports the current compiler version? It's not clear to me which it should search, but it seems like something that should be toggle-able via flag. Not necessary for version `0.1.0`, but we *should* make it clear in the `--help` output.

tools/mason/MasonSearch.chpl
```diff
@@ -0,0 +1,78 @@
...
+// TODO:
+// - order results by some kind of 'best match' metric
+// - allow regex searches
+// - allow for exclusion of a pattern
+//
+
+proc masonSearch(origArgs : [] string) {
+  var args : [1..origArgs.size] string = origArgs;
+
+  const printHelp = || reduce [a in args] (a == "-h" || a == "--help");
```
ben-albrecht: Should `--no-update-registry` be a flag displayed in the help output?

tools/mason/MasonSearch.chpl
benharsh: All packages for now.

tools/mason/MasonSearch.chpl
benharsh: I don't think so, for now it's just for testing.

ben-albrecht:
> How about 'mason search' hides packages starting with an underscore by default?

Works for me.

tools/mason/MasonSearch.chpl
ben-albrecht: Mind making that more explicit in the help documentation?

tools/mason/MasonSearch.chpl
ben-albrecht: Works for me - we should make it clear that `mason search` is case insensitive in the help output, if possible.

test/mason/search/help.good
```diff
@@ -6,4 +6,8 @@ Usage:
 Options:
     -h, --help                  Display this message
 
-When no query is provided, all packages in the registry will be listed
+When no query is provided, all packages in the registry will be listed. The
+query will be used in a case-insensitive search of all packages in the
+registry.
+
+Packages will be listed regardless of their chplVersion compatibility.
```
ben-albrecht: Fortunately, this won't be an issue for the first version :)

Issue #6347, opened 2017-05-30
Unsigned integer overflow with #0:uint operation


ben-albrecht:
### Summary of Problem

The code shown below results in an (unsigned) integer overflow, due to the way the count operator is translated from `start..#end` to `start..(end-1)`. 

The runtime error is:

```
$CHPL_HOME/modules/internal/ChapelRange.chpl:1388: error: halt reached - Iteration over a bounded range may be incorrect due to overflow
```

### Steps to Reproduce

**Source Code:**

```chapel
for 0..#(0:uint) { }
```

PR #7422, opened 2017-09-21
Fix the anonymous range optimization for low-bounded counted ranges

ronawho:
We used to turn `for low.. #count { }` into `for low..(low + (count - 1))`, but
this was wrong for a couple of reasons. `low.. #0:uint` was mistakenly turned
into `low..max(uint)` and `low:uint.. #count:int` didn't type-cast correctly.

Here we fix that by introducing an explicit `chpl_direct_counted_range_iter()`
instead of re-using `chpl_direct_range_iter()`. It does proper casting and
checks if `count == 0`.

This also adds some tests to ensure that error messages for anonymous and
non-anonymous ranges are the same and fixes some cases that were wrong before.

This resolves https://github.com/chapel-lang/chapel/issues/6347


ronawho:
@bradcray you've reviewed the previous anonymous range optimization PRs, mind taking a look at this one too? (Probably easiest to review commit at a time)

modules/internal/ChapelRange.chpl
```diff
@@ -1454,21 +1454,67 @@ proc _cast(type t, r: range(?)) where isRangeType(t) {
...
+  iter chpl_direct_counted_range_iter(low, count) {
+    compilerError("Bound of 'low..' must be an integer");
+    yield nil;
+  }
+
+  // The "actual" counted range iter. Turn the bounds of a low bounded counted
+  // range into the bounds of a fully bounded non-strided range. `low..#count`
+  // becomes `low..(low + (count - 1))`. Needs to check for negative counts,
+  // and for zero counts iterates over a degenerate `1..0`.
+  iter chpl_direct_real_counted_range_iter(low, count) {
```
bradcray: The use of `real` here suggests "real floating point value" to me rather than "I do the real work."  Maybe "helper"?  (in which case I'd probably move it to the end of the routine's name: `chpl_direct_counted_range_iter_helper`

bradcray:
This looks OK to me.  As the author of #6347, did @ben-albrecht get a chance to verify that it does what he wants/expects?  This feels like a biggish change this late in code freeze week (maybe particularly since it hasn't been the quietest of weeks).  What's the benefit of getting this in vs. sitting on it until post-release?  (e.g., do we think it's likely that real user codes will benefit?).  Has this run through the performance playground?

bradcray:
BTW, completely at random, I just happened to re-read #6117 which seems to permit yield-less iterators if you want to try removing all those `yield nil` calls in your error overloads.

ronawho:
I have not run it in the playground. I don't mind sitting on it until after the release if you think that's the right call, but I didn't think of this is a very big or impactful change. It's a decent amount of code, but that's just because of the overloads needed to get coercion right.

Practically speaking, I doubt many (any?) users would run into this bug, and Ben didn't actually run into this bug, but we thought of it while debugging another issue.

Anyways, I don't feel strongly either way, so I'll defer to you whether this should be post-release or not.

bradcray:
I agree it's probably low-risk (the main risk I worry about is inadvertently disabling some other range optimization), but if it's also low-reward...seems like there's no major harm in post-releasing it.

ben-albrecht:
>  did @ben-albrecht get a chance to verify that it does what he wants/expects?

```chpl
// a.chpl
for 0..#(0:uint) { }
```
```bash
master:> chpl a.chpl && ./a
halt reached - Iteration over a bounded range may be incorrect due to overflow.

fix-counted-anon-ranges> chpl a.chpl && ./a

```

Success!

ronawho:
>>did @ben-albrecht get a chance to verify that it does what he wants/expects?

>fix-counted-anon-ranges:> chpl a.chpl && ./a
>Success!

Just FYI types/range/elliot/anonymousCountedRange.chpl (added in this PR) tests many variations of the test in the original issue.

ronawho:
>I agree it's probably low-risk (the main risk I worry about is inadvertently disabling some other range optimization), but if it's also low-reward...seems like there's no major harm in post-releasing it.

Ok, post-release it is.

All your other feedback should be addressed now.

Issue #6016, opened 2017-04-13
Make delete more productive?


bradcray:
* Should we permit delete to take a comma-separated list of things to delete?

````chapel
delete obj1, obj2, obj3;
````

* If we have an array of objects, should we permit delete to promote?

````chapel
delete ArrOfObj;
````

rather than:

````chapel
forall a in ArrOfObj do delete a;
````

vasslitvinov:
I say "yes" to both.

For the array case, did you mean `forall` instead of `for` ?

bradcray:
> For the array case, did you mean forall instead of for ?

I was mostly trying to capture the need to do an explicit loop, but since I'm proposing we use a forall in the implementation, I've switched it now.

psahabu:
@bradcray should we close this issue given that we've merged the branch?

bradcray:
Oops, yes.  Addressed in #7424

PR #7424, opened 2017-09-21
Let 'delete' take multiple expressions, arrays / remove PRIM_DELETE

bradcray:
While working on cleaning up memory leaks, I once again came across the desire expressed in issue #6016 to (a) delete multiple things with a single delete statement or (b) delete arrays as a means of deleting all their elements.  This seemed like it should be a five-minute fix and probably would've been if I'd spent more time in the compiler recently and had the insight that I eventually did as to the best way to approach it.

Specifically, I added a varargs version of `chpl__delete()` (which delete statements get translated into in postFold()) and had it call `chpl__delete()` on each of its arguments, as well as an overload of `chpl_delete()` that accepts arrays (relying on promotion didn't work because of a preference to call the current overload that is completely generic).

But in wrestling through the handling of PRIM_DELETE, I found that there didn't seem to be any need for PRIM_DELETE to exist, so removed it to simplify the compiler and handling of deletes.

Updated the spec to describe this behavior, and to refresh the content in this area at the same time.

Filed a future against a bug I found in which calling delete on `nil` is legal but calls the `deinit()` routine.

TODO:
- [x] get group buy-in
- [x] test
- [x] get code review
- [x] commit parser/lexer generated files once this passes review
- [x] spot-check that some memleaks reports are similar
- [x] update spec


mppf:
I'm confused by the PR message note about promotions. Is chpl__delete() being a varargs function what prevents promotion from working with it? In that case, couldn't it be 2 overloads, one that is not varargs, and a varargs one for 2+ arguments?

How would a 1-argument chpl__delete() be any different from a user `f` function, say? It's certainly the case that a function with a totally generic argument doesn't promote (`proc f(x)`), say, since it's intended to be able to receive an array argument directly. But in this case there's no reason why `chpl_delete` needs to have totally generic arguments. Shouldn't it's arguments be declared with type`:object`?

bradcray:
Here's what I was trying to say:

* assume that users have the ability to write data structures that result in promotion (ideally, maybe as proposed in issue #5114)
* assume that rather than simply permitting `delete` on arrays of objects as proposed in issue #6016, we support `delete` on any collection of objects that supports promotion
* now assume that the user has three such collections of objects named `A`, `B`, and `C`

My concern is that a statement `delete A, B, C;` would dispatch to the varargs version of `chpl__delete()` which would then call `chpl__delete()` on each of the arguments individually.  But since there isn't an overload of `chpl__delete()` that takes the user's collection type and since promotion resolutions are handled by the compiler, those individual calls won't resolve.

If we don't worry about the user promotion case, adding a `chpl__delete()` overload that accepts arrays ought to work for cases we'd care about today (and I have a branch of this branch that starts on that, but then got distracted by the release).

I feel like this might be a more general problem that we've run into in another case or two (use of varargs hinders promotion), but I can't think of what those cases are offhand anymore.  Something with respect to I/O?

mppf:
I'm still missing something. Is the fact that varargs + promotion isn't working a *language* issue or a *compiler bug*?

In particular, we already promote binary operators and functions with arbitrary numbers of arguments. Why is promoting a two-argument call to a varargs function philisophically different?

Is there a future test case somewhere about the varargs + promotion issue? I'd expect you can easily create one that doesn't involve `chpl_delete`.

bradcray:
I was thinking it's a language issue, but now I think I may have been mistaken.  Consider the following:

```chapel
proc foo(x: object) {
  writeln(x);
}

proc foo(xs: ...) {
  for param i in xs.size do
    foo(xs(i));
}

class C { ... }

var X, Y, Z: [1..100] C; 
foo(new C(), new C(), X, new C(), Y, Z, new C());
```

I was thinking that since the second foo() is varargs, that it wouldn't be able to make promoted calls to foo().  However, I forgot until typing this up that the varargs foo() would necessarily have to use a param loop to make the calls to the original foo() in which case the compiler will know the type of each of the calls and will be able to resolve the promotion.  I had been mis-thinking that because the varargs routine mushed all the arrays and singletons together, it wouldn't be able to resolve the promoted calls.

That said, I think we might still have trouble if the user wrote a 2+level promotable data structure because I don't think the language applies multiple promotions automatically (i.e., for a formal of type t, we only promote via array-of-t not array-of-array-of-t).  For arrays, we could add the overload:

```chapel
proc foo(X: []) {
  forall x in X do
    foo(x);
}
```

but for the user's data structure, we wouldn't have a similar overload (for delete).  Unless I'm missing something again.

But maybe the above is also OK — i.e., arguably it's not our job to do multi-level promotions within user data structures that support promotion...

When I get back to this branch, I'll give the promotion+varargs case a try and make sure it works as I argued above, try the array-of-array case without an explicit array overload to see if it fails as anticipated, and add the explicit array overload to see if that makes it work again.

bradcray:
Turns out that without the explicit array overload, we get an ambiguous function call error, so I just added the array overload which seems to work as expected.  Thanks for talking me through this.

mppf:
Can you say more about the ambiguous function call error? In particular I think we should determine if it's a language issue or if it's an implementation issue, and then make a GitHub issue that describes the problem. It's not immediately obvious to me why it shouldn't work.

mppf:
Sounds right, thanks for explaining it.

bradcray:
I think with effort, we could tighten up the catch-all error overload to only catch the types that aren't handled elsewhere.  We'd just want to make sure our where clauses didn't leave any gaps...

Or...  If a "last resort" pragma were applied to the error overload, would calling with an array prefer the promotion case or the last resort case?  (and what would we want to have happen in such cases?)

mppf:
Yeah, I think we could use "last resort" on the error overload & then the promoted array/iterator case ought to work without having a specific overload for it. Pretty sure we'd prefer the promotion case in that event, and I'm pretty sure that's what we'd want it to do.

PR #7427, opened 2017-09-21
Make --llvm-print-ir a developer flag

mppf:
Suggested by @bradcray.

PR #6135 added --llvm-print-ir and documented it as a user feature... But this feature is really only useful to developers or expert users who might used developer flags anyway.

Passed full local testing.


bradcray:
Thanks!  I think you'll need to update userHelp.good as well.  And not that I expect problems, but you might run that test that makes sure all user flags are matched on the man page (in `test/man/`, I think?)

mppf:
I've learned to just run full testing on changes of compiler flags... so I'll sort them out.

PR #7431, opened 2017-09-21
Remove dont_use_exit_use_chpl_exit_instead

ronawho:
The macro `define exit dont_use_exit_use_chpl_exit_instead` was used to prevent
accidental calls to exit instead of chpl_exit. Unfortunately this sort of macro
definition is error-prone because it replaces every `exit` token, not just
function calls. This is causing compilation failures with new versions of clang
on OSX with an error along the lines of:

```
 no member named 'dont_use_exit_use_chpl_exit_instead' in the global namespace
 using ::exit;
       ~~^
 ../../../../include/chplrt.h:41:17: note: expanded from macro 'exit'
 #define exit    dont_use_exit_use_chpl_exit_instead
```

This removes the macro and instead uses look_for_calls to check the runtime for
calls to exit(). It basically renames lookForBadAlloc to lookForBadRTCalls and
instead of just looking for calls to the system allocator it looks for calls to
exit() as well.


ronawho:
@ben-albrecht mind reviewing the python changes? Probably easiest to review commit at a time?

@bradcray this is the replacement for #7401 that checks the runtime for calls to exit(). Ok with you?

.travis.yml
```diff
@@ -11,7 +11,7 @@ matrix:
     - env: TEST_COMMAND="make test-venv && CHPL_HOME=$PWD ./util/run-in-venv.bash ./util/test/check_annotations.py"
       git:
         depth: 100000
-    - env: TEST_COMMAND="./util/devel/lookForBadAlloc"
+    - env: TEST_COMMAND="./util/devel/lookForBadRTCalls"
```
gbtitus: I might have called it `lookForBadCallsInRT` instead of `lookForBadRTCalls` since the latter sounds like we're looking for bad calls *to* the runtime instead of bad calls *in* it, but I'm okay if you leave it as is.

gbtitus:
👍  with one line-specific comment you are free to take or ignore, your choice.

bradcray:
Thanks for taking this on in spite of your reluctance!

util/devel/look_for_calls.py
```diff
@@ -77,29 +71,42 @@ def find_source_files(search_dir):
...
 
 
-def check_for_alloc_calls(search_dir, exclude_paths=None, rel_paths=True):
-    """Check source files in search_dir for calls to the system allocator.
-       Don't check files that contain any of the exclude_paths.  Report files
-       relative to search_dir/../ if rel_paths is True, otherwise use absolute
-       paths. Returns True if alloc calls were found, False otherwise"""
+@contextlib.contextmanager
+def get_tmp_dir():
+    """Context lib for a temp fir. Would use tempfile.TemporaryDirectory(), but
```
ben-albrecht: temp fir?

image

util/devel/look_for_calls.py
```diff
@@ -77,29 +71,42 @@ def find_source_files(search_dir):
...
+       that's python >= 3.2"""
+    try:
+        tmp_dir = tempfile.mkdtemp()
+        yield tmp_dir
+    finally:
```
ben-albrecht: no `except`?

util/devel/look_for_calls.py
ronawho: lol

util/devel/look_for_calls.py
ronawho: no? Why would there be one?

PR #7432, opened 2017-09-21
Manually hoist loop-invariant array access in miniMD

benharsh:
This change recovers performance lost due to #7371 , and improves performance by 30% on 16-node-xc.


ronawho:
👍

Issue #7251, opened 2017-09-08
Long shutdown times for gasnet-aries with multi-domain enabled


ronawho:
GASNet 1.30.0 included a gemini/aries update to "Improve robustness of exit
code". It appears that the improved robustness in combination with re-enabling
GASNet's multi-domain feature in #6893 has resulted in significantly higher
exit times.

Currently working around this for now with https://github.com/chapel-lang/chapel/pull/7241

**edit**

based on discussion with the GASNet team below, there wasn't actually a regression in shutdown time between 1.28.2 and 1.30.0, it's just that the final exit now has an `alarm()` watchdog. With that in mind, I'm happy with our current work-around in https://github.com/chapel-lang/chapel/pull/7241.

ronawho:
Copying comments from https://github.com/chapel-lang/chapel/commit/552763d:

>>>>>>This increase in exit times in the presence of multi-domain is not something we would have expected. However, looking over the multi-domain code I have identified a possible cause.
>>>>>>
>>>>>>I hope to investigate this more on Friday.
>>>>>>
>>>>>>It would help if you can confirm that Chapel might call gasnet_exit() from a thread other than the first one. That is a pre-condition for the possible cause I've identified (and something I know we don't test often, if ever). However, if you were to tell me that can never happen then I'll dismiss that theory and look elsewhere.
>>>>>
>>>>>init and exit should be called from the same thread, unless there's an abnormal exit, but the regressions I'm seeing are for clean exits with just a basic hello.chpl
>>>>>
>>>>>Although, we do a gasnet_exit() from each locale after doing a barrier(https://github.com/chapel-lang/chapel/blob/master/runtime/src/comm/gasnet/comm-gasnet.c#L1056-L1058).
>>>>
>>>>>I'm not confident on how gasnet_exit works, but it's my understanding that if you exit on one node, it will try to shutdown the others too, so maybe a non-0 locale is doing the exit first, propagating the exit to other nodes, and the AM handler on locale 0 is doing an exit instead of the main thread? (it's getting late, and I'm not confident in anything I just said, so don't quote me on any of that)
>>>>
>>>>I will see what I can reproduce w/ hello.chpl w/o too many assumptions about the behavior of Chapel. Can you give me an idea of the scale of the run necessary to reproduce? Any special conditions?
>>>>
>>>
[...]

PR #7435, opened 2017-09-21
Increase the gasnet exit timeout under gasnet-{aries,gemini} again

ronawho:
Bump the default timeout again (first done in #7241). GASNet 1.30.0 added an
additional shutdown watchdog, which has caused us to see some sigalarms. We use
the multi-domain feature, which increases the shutdowm times so for now just
raise the default timeout. Note that there was no actual regression in the
shutdown times in 1.30.0, it's just that there is an extra watchdog whose limit
is too low when the multi-domain feature is being used.

For more info see https://github.com/chapel-lang/chapel/issues/7251 and
https://upc-bugs.lbl.gov/bugzilla/show_bug.cgi?id=3621

### DIFF
--- a/runtime/src/comm/gasnet/comm-gasnet.c
+++ b/runtime/src/comm/gasnet/comm-gasnet.c
@@ -793,11 +793,13 @@ static void set_num_comm_domains() {
     chpl_error("Cannot setenv(\"GASNET_AM_DOMAIN_POLL_MASK\")", 0, 0);
   }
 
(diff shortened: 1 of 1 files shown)


ronawho:
@dmk42 and @gbtitus this should resolve the infrequent failures we're seeing on KNL.

gbtitus:
Surely 10 seconds will be enough!  👍

dmk42:
Looks good, thanks!

PR #7438, opened 2017-09-22
Minor cleanups to "running on a cray" doc 

ronawho:
Remove references to pgi. We don't build the module with support for pgi as a
target compiler anymore, and we don't trust pgi that much anyways, so don't
list it as a backend compiler.

Remove the "change your host compiler section". We really recommend using gnu,
so don't list alternatives.

Remove note about recommending cce 8.4.0 or newer. 8.4 is pretty old at this
point. Any site using Chapel will certainly have at least 8.4.

Remove note about setting CHPL_TASKS. We really want people to just default to
qthreads. This might have been more relevant when muxed was around, but now we
really don't want fifo to be used.

Remove note about setting the comm layer. ugni provides the best performance,
so we want users to use it if they're on a xe/xc. If they're on a cs, gasnet is
there only option so there's no reason for them to change it.

Fix a reference from CHPL_COMM to CHPL_NETWORK_ATOMICS in the AMO section

### DIFF
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -177,7 +177,7 @@ Using Chapel on a Cray System
         source util/setchplenv.bash
 
 
-4) Compile your Chapel program.  For example:
+2) Compile your Chapel program.  For example:
 
    .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


doc/rst/platforms/cray.rst
```diff
@@ -204,27 +177,7 @@ Using Chapel on a Cray System
...
-   installation being used, whether that is a source distribution or the
-   pre-built module.  If it is not, the Chapel compiler will produce an
-   error message saying so when you try to compile anything.  If you get
-   this error, you will need to build the desired configuration (if you
-   are working from source) or modify your configuration so that it is
-   one of those supplied (if you are working with the pre-built module).
-
-
-3) Compile your Chapel program.  For example:
+4) Compile your Chapel program.  For example:
```
gbtitus: Numbering is wonky starting here and continuing below -- this should be 2) instead of 4).

ronawho: haha, whoops. I promise I can count.. 1, 4, 5,...

Issue #7411, opened 2017-09-20
rare SEGV in determineSlurmVersion()


PHHargrove:
In our CI testing of GASNet with Chapel's master branch on NERSC's Cori and Edison systems, we have been seeing somewhere below 1% of the runs failing with a SEGV within the first second of execution.  I have finally been able to determine with certainty that this is in the launcher, and have a core file with debug info:
```
Core was generated by `./hello4-datapar-dist -nl 2'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00002aaaab031f7d in __isoc99_fscanf () from /lib64/libc.so.6
Missing separate debuginfos, use: zypper install glibc-debuginfo-2.22-61.3.x86_64
(gdb) where
#0  0x00002aaaab031f7d in __isoc99_fscanf () from /lib64/libc.so.6
#1  0x0000000000407b86 in determineSlurmVersion () at launch-slurm-srun.c:87
#2  0x000000000040888f in chpl_launch (argc=3, argv=0x7fffffff6b58, numLocales=2) at launch-slurm-srun.c:456
#3  0x00000000004029db in main (argc=3, argv=0x7fffffff6b58) at main_launcher.c:458
```

Examining the code it appears that the return from `fopen()` is not checked.  
Indeed it appears to have failed:
```
(gdb) up
#1  0x0000000000407b86 in determineSlurmVersion () at launch-slurm-srun.c:87
87          fscanf(sysFile, "%c", &tmp);
(gdb) print sysFile
$1 = (FILE *) 0x0
```

I strongly suspect that the reason I am seeing some small fraction on my runs fail in this manner is because our CI scripts are putting all the chapel executables in *one* directory.  If I read the code correctly this means that multiple concurrent runs (split over distinct batch runs for each value of NumLocales) are going to be using the **same** temporary file for determining the slurm version!  

I am guessing that `fopen()` fails when another simultaneous launch removes the file.  
I would suggest replacing `system()` redirected to temporary file with use of `popen()`.  

I do not have a contributor agreement, but consider the following to be _de minimis_
```diff
--- ./runtime/src/launch/slurm-srun/launch-slurm-srun.c~        2017-09-20 03:30:21.000000000 -0700
+++ ./runtime/src/launch/slurm-srun/launch-slurm-srun.c 2017-09-20 03:34:33.000000000 -0700
@@ -30,7 +30,6 @@
 #include "error.h"

[...]

PR #7441, opened 2017-09-22
Fix a race in our slurm-srun launcher while determining the slurm version

ronawho:
Fix a race in slurm-srun:determineSlurmVersion(). Previously we wrote the
contents of `sbatch --version` out to a tmp file and then read the file to
determine versions. Unfortunately, we used the same tmp file so this would
result in segfaults and other sporadic failures if 2 programs were running from
the same directory. While we run concurrent testing on crays, we use unique
subdirs so we haven't run into this in nightly testing. This was discovered by
the gasnet team's CI testing.

There's no reason to be using a tmp file anyways (this code was originally
copied from the slurm-gasnetrun_ibv launcher) so just use chpl_run_utility1K
instead. I copied the pattern we use in determineQsubVersion.

Note that slurm-gasnetrun_ibv and pbs-gasnetrun_ibv also need to be updated,
but I'll do that in a later PR. slurm-srun is one of our most commonly used
launchers, so I want to get this in a soon as possible.

This resolves https://github.com/chapel-lang/chapel/issues/7411

### DIFF
--- a/runtime/src/launch/slurm-srun/launch-slurm-srun.c
+++ b/runtime/src/launch/slurm-srun/launch-slurm-srun.c
@@ -30,7 +30,6 @@
 #include "error.h"
 
 #define baseSBATCHFilename ".chpl-slurm-sbatch-"
-#define baseSysFilename ".chpl-sys-"
 
 #define CHPL_WALLTIME_FLAG "--walltime"
 #define CHPL_GENERATE_SBATCH_SCRIPT "--generate-sbatch-script"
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus mind reviewing?

Passes with `NUM_TRIALS=100 NUM_PROGRAMS=10` for the reproducer in https://github.com/chapel-lang/chapel/issues/7411#issuecomment-331492537 (previously it failed 5/10 times for `NUM_TRIALS=10 NUM_PROGRAMS=2`

Running a paratest for release/examples now

ronawho:
paratest was clean

gbtitus:
Yeah, this looks good.   👍

PR #7444, opened 2017-09-22
Remove msg field from Error

mppf:
Historically we've had the Error class contain
a 'msg' field of type string. This made sense when
error handling was a prototype feature.

Note that the Error class is basically required to
exist by the language as a super type of other Errors.

This PR removes the msg field. Why?

* It's easier to add this sort of thing than remove it
   (i.e. not having that field is the conservative design)
* Error should be as close to a pure virtual base
   class as possible since it's really expressing an
   interface
* I can imagine user-defined Errors for which it does
   not make sense to have a string .msg field.
   (e.g. they might store another error, which
    itself contains a message if there is one).
* Users ought to make their own subclasses of Error
   for different error types
* Now we have a 'proc message()' that subclasses are
   supposed to override to provide their error message.
   Using .msg on the parent class might be confusingly
   ineffective in that setting.

- [x] full local testing
- [x] add error message if they are used, tests in test/deprecated


mppf:
@psahabu - can you review?

Issue #6422, opened 2017-06-09
Check .graph/GRAPHFILES as part of a smoke test


ronawho:
We have some code in start_test that validates that all .graph files in the test directory are listed in a *GRAPHFILES file and that for each file listed in the *GRAPHFILES files, there is a corresponding .graph file in the test directory. i.e. we ensure there is a 1:1 mapping of .graph files that exist and that are listed in one of the GRAPHFILES. 

We should pull this code out into a helper script and run it as some sort of smoke test or linter or something since it's easy to get this wrong and have a bunch of nightly jobs mail warnings (#6421, and several other cases.)

It's the code inside of [start_test::check_for_duplicates()](https://github.com/chapel-lang/chapel/blob/main/util/test/start_test.py#L1187)

Edit: we'd probably add a call to this helper script in a new entry under jobs in [this file](https://github.com/chapel-lang/chapel/blob/main/.github/workflows/CI.yml).  To test that works, it should trigger a 7th "check" when opening a PR on the repo (remembering that sometimes it takes a second for the non-DCO checks to turn up)

Shivansh-khare:
if I can help in this issue...

pree-T:
Is this issue still open?

lydia-duncan:
I believe so, yes

PR #7446, opened 2017-09-22
Futurize race-y HPL HPCC 2012

bradcray:
In rewriting this version of HPL to avoid the use of the (deprecated)
'=>' operator and use reindexing instead, a race seems to have been
introduced in which a few different failing behaviors occur, such as
failing verification, having bogus indices being computed on in the
reindex array view, etc.  It's unclear whether the cause of the race
is in the benchmark itself, the Dimensional distribution, or the
reindex array view.  A bug in the latter two would be problematic, but
neither has had much weight put on them yet, so it's not clear where
the source is.  We haven't had a chance to dig into this to determine
the source and don't currently consider this as high-priority as other
issues, so are futurizing the test for now.

The specific changes I made were:

* Made hpl.hpcc2012.chpl the version we would like it to be given the current language definition
* Added a .skipif and a .future file to only test it in gasnet where the race exhibits itself and to skip it for performance testing.
* Removed the *.ml* files to disable multilocale testing for now (these can be restored if/when the race is fixed)
* Preserved comm=none testing for it by adding .comm-none.chpl, .comm-none.compopts, .comm-none.skipif, and .comm-none.execopts files and updating the .comm-none.good file.  When the future is fixed, these files can be removed.


ronawho:
@benharsh any idea if we skip performance tests if they have a .future?

benharsh:
I just tried locally and got a weird start_test error. Probably easiest to just remove the perf files.

bradcray:
I already have a commit (not yet pushed) to remove the .ml* files.  I thought the .skipif would take care of traditional performance testing.

bradcray:
@ronawho and @benharsh: I was going to send out a general request for a reviewer for this (and in writing it realized that I wanted to add the last few commits), but since you've already started looking at it, would one of you grab it?  I've tested the whole directory with 'start_test' for comm=none, comm=gasnet, -performance, and -memleaks and all seem to work as expected.

test/studies/hpcc/HPL/vass/hpl.hpcc2012.ml-perf.graph
```diff
@@ -1,5 +0,0 @@
-perfkeys: Performance (Gflop/s) =
```
ronawho: You'll need to remove the .graph files from $CHPL_HOME/test/ML-GRAPHFILES too

test/studies/hpcc/HPL/vass/hpl.hpcc2012.future
```diff
@@ -0,0 +1,13 @@
+???: hpcc2012 version of HPL is race-y when rewritten to avoid '=>'
```
ronawho: Replace `???` ?

test/studies/hpcc/HPL/vass/hpl.hpcc2012.ml-perf.graph
bradcray: I was wondering about this -- what happens if I don't?  Do we just stop getting data after that point, or do we get an error / failure of some sort?  (I was hoping the former...)

ronawho: start_test checks that all .graph files are listed in a *GRAPHFILES file and vice-versa. It'll issue a warning if that's not true. (We wanted to split out this check and run it as part of the smoke test https://github.com/chapel-lang/chapel/issues/6422 ) If you want to keep the graph around, just leave the .graph files in.

ronawho: FYI I think `start_test --performance test/studies/hpcc/HPL/vass/` should have issued the warning

bradcray: Thanks for the clarification / reminder.  (I got dozens of warnings on that run, so must've missed it in the noise.  But... the normal `-performance` run would check the ML-GRAPHFILES as well?)

ronawho: Yeah, it should. We just look for *GRAPHFILES and *.graph so it should catch all perf configs

bradcray: OK, didn't anticipate that, thanks.

bradcray: Do you not get dozens of warnings about studies/prk/* .graph files being missing from GRAPHFILES when you run start_test -performance on a specific subdirectory/test?  I can't think what I'd have done personally to trigger that.

bradcray:
OK, I think I've pushed the necessary changes to keep *GRAPHFILES working.

ronawho:
Looks good to me. Might be worth having a second look from @benharsh still

test/studies/hpcc/HPL/vass/hpl.hpcc2012.chpl
```diff
@@ -398,30 +398,7 @@ proc DimensionalArr.dsiLocalSlice1((sliceDim1, sliceDim2)) {
...
 
-  // This works but uses the '=>' operator...
-  pragma "no auto destroy"
-  var slice => locAdesc.myStorageArr[r1, r2];
-  pragma "no auto destroy"
-  var result: [(...reindexExpr)] => slice;
-  return result;
-
-  /* This rewrite avoiding the '=>' operator doesn't:
   return locAdesc.myStorageArr[r1, r2].reindex({(...reindexExpr)});
```
benharsh: Might be worth a quick -valgrindexe test run, I believe the 'no auto destroy' pragmas were required in order to prevent use-after-free errors.

test/studies/hpcc/HPL/vass/hpl.hpcc2012.ml-perf.graph
ronawho: Hmm, are you on a case-insensitive system (mac)?

#6405 changed the case of some prk tests. I'm wondering if on a case-insensitive system if the git pull of #6405 didn't change the names locally. I would try doing `rm -r $CHPL_HOME/test/studies/prk && git co $CHPL_HOME/test/sudies/prk`

ronawho:
FYI the travis failure is because the annotations for hpcc.hpl no longer have a .graph file. So probably just remove those annotations? (That check was added to catch cases when we rename .graph files or if you got the name of the .graph file wrong)

test/studies/hpcc/HPL/vass/hpl.hpcc2012.chpl
bradcray: Looks like it works fine.

bradcray:
Oops, I forgot to say `[reviwed by @ronawho and @benharsh]`

Issue #6774, opened 2017-07-20
Plain on statements and initializers


lydia-duncan:
Within an on statement that is not within a parallel or loop context, should we allow:
- field initialization in Phase 1?
- super.init()/this.init() calls?

Relevant context:
- Parallel statements are not allowed to contain these operations already
- Neither are loops
So if those operations are in parallel or loop statements within an on statement, or an on statement is in a parallel or loop statement and contains these operations, then that is already an error

I'm going to disallow these cases for now, but I think this is worth discussing as it
may impact the implementation of:
- types with distributed array fields
- arrays, domains, and distributions

lydia-duncan:
I don't think it makes sense for an on statement to wrap the field initialization when the type is a primitive, but more complicated types seem reasonable to me.  Thoughts?

mppf:
The main case that I know of for caring about this is when creating a class instance, e.g.
``` chapel
proc MyType.init(targetLocale:locale) {
  on targetLocale {
    myField = new SomeClass();
  }
  super.init();
}
```

However that case is easy enough to work around:
``` chapel
proc MyType.init(targetLocale:locale) {
  var tmpMyField:SomeClass;
  on targetLocale {
    tmpMyField = new SomeClass();
  }
  myField = tmpMyField
  super.init();
}
```

(assuming I am not missing something that would disallow the workaround).

Separately, I don't favor allowing any complicated constructs around super.init()/this.init(); these are really a marker and should pretty much be at the top level of the init function in my view (with a possible exception for param conditionals).

vasslitvinov:
super/this.init() should indeed not be inside complicated constructs.

Field initializations either, except here the "complicated" bar is lower. For a field initialization, I think the requirement is that the compiler must be able to reason whether it does or does not happen.

In particular, initalizing within an on statement should be fine.

lydia-duncan:
So that's two :+1:  for plain on around field initialization and two :-1: for plain on around super.init()/this.init()?

mppf:
@lydia-duncan - more or less on my part - I'm OK either way for on around field init especially for now.

bradcray:
I'm just catching up on this issue and think I'm OK with the trending agreement.

PR #7447, opened 2017-09-22
Update initializer chip

lydia-duncan:
Contains our decisions on `on` statements, and generic instantiations using
type declarations, since neither of those had been documented.  The former
is mostly just a mention in the list of what is and is not allowed to surround a
`this.init()` or `super.init()` call, while the latter is a new section with more
details and examples.


lydia-duncan:
@noakesmichael - I think this is ready for an initial look (at least, I'm a lot happier about it than I was at first).

lydia-duncan:
@mppf and @vasslitvinov - just a heads up, this incorporates the results of our discussion on issue #6774 and the conversation we had in May (which I can forward to you if you need a reminder).  It shouldn't be controversial and Mike has reviewed it so I am going to merge, but feel free to give it a look yourselves at your leisure.

PR #7450, opened 2017-09-22
Fix race in slurm-gasnetrun_ibv launcher while determining the slurm version

ronawho:
Same as #7441, but for slurm-gasnetrun_ibv instead of slurm-srun. You can
really tell how much our launchers are clones of each other when the version
number function was still called `determineQsubVersion()`

I just copied the determineSlurmVersion() code from slurm-srun, and removed
some checks for nccs/torque, which were leftover from when this was copied from
the qsub version.


ronawho:
@gbtitus mind reviewing. Tested the hellos with gasnet-ibv on chapcs

gbtitus:
Looks good!  👍

PR #7451, opened 2017-09-22
Init record this intent

noakesmichael:
Convert blank intent for 'this' of record-initializer to ref.

This PR resolves the initializer future for classes/initializers/records/emptyGlobalConst.chpl
That test failed in the parallel pass because that pass could not replicate a const record.
The problem was that the 'this' concrete formal to the initializer had been set to const-ref
rather than 'ref'.

I tried a few paths to fixing this but ultimately decided to update the intent during
pre-normalization for record initializers.  This addressed the future but caused
approximately 10 tests to start to value during an lvalue check in function-resolve.

This problem was addressed by a minor update to the logic in a late sub-phase of
normalize.

This PR has 4 commits

1. A minor change in initializerRules.cpp to convert a blank intent to a ref intent for
record initializers.  There are also a handful of alignment fixes etc.

2. A handfull of non-functional changes to normalize.cpp that were related to the
call path for the function I intended to update.

3. A minor functional change to stop the compiling from adding FLAG_EXPR_TEMP
to the LHS for uses of a new expr for non-generic records.  The original code includes
a comment that hints that this is a "hack" to trigger auto-copy/auto-destroy behavior.

4. Convert two passing futures to be two passing tests.  The second future documented
an existing situation that exposed the issue that commit 3 is solving.

Compiled on clang/darwin and gcc/linux with/without CHPL_DEVELOPER.
Passed with paratest with -futures.  Also passed with future and --no-local.

Also evaluated memory leaks before/after this change and confirmed no change
in metrics.


test/classes/delete-free/owned-init2.good
```diff
@@ -1 +1,2 @@
-{x = 1} {x = 2} {x = 3}
+owned-init2.chpl:14: In function 'main':
+owned-init2.chpl:17: error: illegal lvalue in assignment
```
lydia-duncan: It looks like this changes the .bad output to be the expected output for the test.  Have we verified that this was an oversight?

lydia-duncan:
Oh, and it might be good to do a quick valgrind check over the passing futures

PR #7452, opened 2017-09-22
Fix a race in pbs-gasnetrun_ibv launcher while determining the pbs version

ronawho:
Similar to #7441/#7450, but for pbs-gasnetrun_ibv. I partially copied the code
from pbs-aprun, but note that this launcher checks for torque instead of moab.
torque/moab are one in the same, so I probably could have updated this launcher
to use the pbs-aprun check, but we don't have a machine to fully test this
launcher on, so I didn't want to change too much.

Since we don't have a machine to test this on, I just made sure the launcher
compiles and that determineQsubVersion works correctly on PBSPro and
moab/torque machines.


ronawho:
@gbtitus one last one for you. 

As mentioned in the PR we can't fully test this, but I did the best I could to ensure this is correct

gbtitus:
Looks good!   👍

Issue #7182, opened 2017-09-02
Gmp sometimes fails to build without bison or yacc


dmk42:
### Summary of Problem
Sometimes in the gmp distribution, in the `third-party/gmp/gmp-src/demos/calc` directory, `calc.y` ends up being newer than `calc.c`.  When that happens, gmp will try to find either bison or yacc to rebuild `calc.c`.  If it fails, the build of gmp fails.

We could either touch the `calc.c` file to make sure it is always newer than `calc.y`, or we could add to the Chapel prerequisites that either bison or yacc is required to build gmp from sources.

PR #7454, opened 2017-09-22
Don't depend on third-party source files to auto-rebuild a third-party lib

ronawho:
PR #7004 added code to automatically rebuild a third-party library if our
README/Makefiles changes or if any source file in the third-party dir changed.
Unfortunately depending on third-party sources had some unforeseen consequences
because of built-in make rules, so now only rebuild if our README/Makefiles
changes. We'll still auto-rebuild on version upgrades or if we bring in an
upstream patch or something so long as we're careful to update the README to
note version changes and mods.

The specific problem we ran into with depending on third-party sources was that
a built-in rule was causing us to auto-rebuild some gmp files that have a .y
file. Make will automatically rebuild a .c file if it has a .y file with a
newer timestamp than it. We could probably work around this, but it makes me
nervous about other built-in rules and I don't want to take any chances. For
gmp, this was effectively requiring a user to have yacc/bison installed in
order for `make check` to pass, which we do as part of the speculative build.

Also touch the specific .c files that were being rebuilt. They won't be rebuilt
by an implicit make rule anymore, but the gmp build might rebuild them, so make
sure that the generated .c has a newer timestamp than the source .y.

This resolves an issue where developers were seeing modifications to
`third-party/gmp/gmp-src/demos/calc/{calc.c, calclex.c}`. This also 
resolves https://github.com/chapel-lang/chapel/issues/7182

### DIFF
--- a/third-party/Makefile
+++ b/third-party/Makefile
@@ -59,29 +59,25 @@ clobber: FORCE
 
 depend:
 
-# Third-party dependencies. If any of these files (3rd party source dirs or our
+# Third-party dependencies. If any of these files (our 3rd party README or
 # Makefiles) change, then rebuild the 3rd party lib. This allows us to
 # automatically rebuild on 3rd party version upgrades or if we specify
-# different option in the Makefile.
+# different options in the Makefile. Note that this requires us to update the
+# version number and list patches in our README file. We used to use
+# third-party source files as dependencies too, but that introduced problems
+# with some files getting rebuilt because of built-in make rules.
 
-# Common C/C++ file extensions
-SRC_FILES_FILTER += -iname '*.c'   -o -iname '*.h'   -o -iname '*.cpp' -o
-SRC_FILES_FILTER += -iname '*.cc'  -o -iname '*.cxx' -o -iname '*.hpp' -o
-# Less common C++ extensions
-SRC_FILES_FILTER += -iname '*.c++' -o -iname '*.hp'  -o -iname '*.hxx' -o
-SRC_FILES_FILTER += -iname '*.h++' -o -iname '*.tcc' -o -iname '*.cp'
+# Find 3rd party dependencies -- arg 1 is the top-level dir with Makefiles/README
+find_3p_depend = $(1)/Makefile* $(1)/README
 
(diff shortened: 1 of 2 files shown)


ronawho:
@dmk42 do you mind reviewing and seeing if this resolves https://github.com/chapel-lang/chapel/issues/7182?

@gbtitus mind giving this a post-commit look on monday?

dmk42:
Yes, this resolves #7182 .  Looks good, thanks!

third-party/Makefile
```diff
@@ -59,29 +59,25 @@ clobber: FORCE
...
-MASSIVETHREADS_DEPEND := $(call find_3p_depend, $(MASSIVETHREADS_SUBDIR), $(MASSIVETHREADS_DIR))
-QTHREAD_DEPEND        := $(call find_3p_depend, $(QTHREAD_SUBDIR),        $(QTHREAD_DIR))
-RE2_DEPEND            := $(call find_3p_depend, $(RE2_SUBDIR),            $(RE2_DIR))
+# different options in the Makefile. Note that this requires us to update the
+# version number and list patches in our README file. We used to use
+# third-party source files as dependencies too, but that introduced problems
+# with some files getting rebuilt because of built-in make rules.
+
+# Find 3rd party dependencies -- arg 1 is the top-level dir with Makefiles/README
+find_3p_depend = $(1)/Makefile* $(1)/README
```
gbtitus: I don't see how this works now.  How does the `shell find ...` that used to be part of the expansion of find_3p_depend happen in, with this change?  (It seems like we wouldn't actually need the `find` any more, but it also seems like we would still need the `shell` to expand the wildcarded `Makefile*`.)

ronawho: Not sure I'm following. We're not doing a `find` anymore. And make will expand wildcards in dependencies, so we don't need to do that. For something like jemalloc this effectively does something like:

`jemalloc: $(CHPL_MAKE_HOME)/third-party/jemalloc/README $(CHPL_MAKE_HOME)/third-party/jemalloc/Makefile*`

and then make will expand `Makefile*`

gbtitus: Oh, that's funny!  I continue to learn new things that Gnu make can do.  Wildcards in targets and prerequisites is an example.

PR #7458, opened 2017-09-22
Array alias remove

noakesmichael:
Remove vestiges of the '=>' operator

This operator was deprecated for this release.  Multiple
people have worked to replace deprecated uses of '=>'
to use conventional ref.

1. Updates the Chapel parser to remove the '=>' operator.

2. Removes all uses of FLAG_ARRAY_ALIAS

3. Remove the deprecated/ test for array alias


bradcray:
Yay!

compiler/passes/normalize.cpp
```diff
@@ -1541,35 +1537,6 @@ static void normalizeTypeAlias(DefExpr* defExpr) {
...
-static void normalizeArrayAlias(DefExpr* defExpr) {
-  SET_LINENO(defExpr);
-
-  Symbol* var  = defExpr->sym;
-  Expr*   init = defExpr->init->remove();
-
-  if (defExpr->exprType == NULL) {
-    CallExpr* newAlias = new CallExpr("newAlias", gMethodToken, init);
-
-    defExpr->insertAfter(new CallExpr(PRIM_MOVE, var, newAlias));
```
bradcray: @noakesmichael: I just stumbed across this by chance:  It appears to me that the newAlias() functions should be able to be removed from ChapelArray.chpl now that '=>' has been retired.  (there's also a "new alias fn"/"newAlias function" pragma in the compiler and some logic relating to destructors, but that is unrelated I believe.

PR #7466, opened 2017-09-25
Fix reduce intents for parallel iterators with forall.

### LINKED ISSUE #5223: + reduce not working when a normal forall loop is used inside the leader

vasslitvinov:
This fixes the generated code for forall loops with
in intent and/or reduce intent, ex.:

    forall ... with (in somevar, + reduce anothervar) ...;

where the corresponding parallel iterator itself has a forall loop, ex.:

    iter myParallelIter(param tag: iterKind) {
      ...
      forall ... {
        yield ...;
      }
      ...
    }

Before this change, the same shadow variable for the above "somevar"
and "anothervar" would be accessed for all iterations of the forall
loop inside myParallelIter(). This can cause unintended data races for
reduce and 'in' forall intents. Cf. no data races for const forall intents
(cannot modify the shadow variables) and already-intended data race
for the ref forall intent.

This fix adds shadow variables for this forall loop as well.

This fix does not handle the case where the parallel iterator has
**nested** forall loops. Handling of that case should be straightforward.
I did not do it because I timed out and because it seems we run into
other/unrelated issues with nested forall loops in the parallel iterator.
I am leaving it for future work.

So if there are nested forall loops, the compiler now generates an error
when an 'in' or reduce intent is present. This is to avoid unintended
races in the generated code, given the state of implementation.

Also I am handling only reduce intents for the "old" implementation.
This is because in the old implementation the specific intent is not
easily discernible. The reduce intent is discernible only by the virtue
of the shadow variable being a subclass of the ReduceScanOp class.

Retiring BenH's original future and adding a more extensive test case.
Resolves #5223.

Testing: linux64 --verify and gasnet.


vasslitvinov:
@benharsh would you review?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -1062,6 +1062,152 @@ static const char* astrArg(int ix, const char* add1) {
...
+
+      case TFI_IN:
+      case TFI_REDUCE:
+        // May result in data races or incorrect behavior
+        // if we don't handle multiple enclosing foralls.
+        USR_FATAL_CONT(yieldCall, "a parallel iterator with a yield nested in two or more enclosing forall statements is not currently implemented in the presence of an 'in' or 'reduce' intent.");
+        USR_PRINT(efs, "the immediately enclosing forall statement is here");
+        USR_PRINT(efs2, "the next enclosing forall statement is here");
+        // If we continue, we may get a const violation on x*_shadowVarReduc.
+        USR_STOP();
```
benharsh: Shouldn't this error be triggered in the new test you added? Or does it only fire for multiple nested foralls like this?
```
iter ( ... )
  forall ...
    forall ...
      yield ...
```

benharsh:
Does this help miniMD at all?

compiler/resolution/implementForallIntents.cpp
vasslitvinov: This should not be triggered in my test. As you say, the error requires at least 2 nested foralls around the yield statement.

vasslitvinov:
Testing update: the test I added and the test I defuturized run clean under valgrind and with --memLeaks, on a single locale.

PR #7467, opened 2017-09-25
Prevent certain data races upon 'in' forall intents

vasslitvinov:
Before this change, the compiler would generate racy code
for a forall loop with an 'in'-intented shadow variable, ex.:

    forall ... with (in somevar) ...;

when the corresponding parallel iterator has a for loop over
another parallel iterator, ex.:

    iter myParallelIter(param tag: iterKind) {
      ...
      for anotherIter(iterKind.standalone) {
        yield ...;
      }
      ...
    }

The reason is that all iterations of the for-loop over anotherIter
would use the same shadow variable for 'somevar', allowing for
unintended data races.

The compiler changes to handle this situation properly would be too invasive
so I did not want to put them in so close to code freeze.
Leaving this as future work.

Instead, this change detects this situation and changes the 'in' intent
to 'const in' intent. So if 'somevar' is modified within the loop,
const-checking flags that as an error. If it is not modified
within the loop, there will be no data races. The user is warned
about the change from 'in' to 'const in'.

A couple of miniMD and stencil variants pass a tuple by 'in' intent.
Changing those to 'const in' intent.

This originally exhibited itself as a performance issue with miniMD.
Thanks to BenH for narrowing down this issue to a small reproducer
and finding a workaround.

Testing: linux64 --verify, gasnet.

### DIFF
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -2399,12 +2399,14 @@ static void propagateThroughYieldNew(ForallStmt* fs,
   // Check for an "eflopi" (Enclosing For-Loop Over a Parallel Iterator).
   // This is specific to a given yield.
   // Todo: handle multiple yields in a single eflopi.
-  bool      eflopiChecked = false;
   ForLoop*  eflopiLoop    = NULL;
   CallExpr* eflopiCall    = NULL;
   int       eflopiIdx     = 1;
   CallExpr* eflopiHelper  = NULL;
   int       ix            = -1;
+
+  // sets eflopiLoop and eflopiCall if appropriate:
+  eflopiFind(rcall, eflopiLoop, eflopiCall);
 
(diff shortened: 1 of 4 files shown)


vasslitvinov:
@benharsh and @mppf - could you look at this and/or review?

benharsh:
The code changes look reasonable to me.

@bradcray : are you OK with this approach? Or should the PR that caused this change be reverted despite the bugs it fixed?

mppf:
I'm OK with it.

bradcray:
This seems like a discussion topic for today's meeting.  @vasslitvinov, could you be prepared to describe what bugs were fixed by the original PR, what the instability with 'in' intents is, and this proposed PR so that everyone can weigh in on what they'd like to see happen?

bradcray:
(let me know if there are any links or visual aids that I can put up to help the discussion)

PR #7468, opened 2017-09-25
Make --print-callgraph print calls to functions in standard modules

daviditen:
The --print-callgraph option only printed calls to user functions, but we'd
like to also see calls to functions in the standard modules.  Make it print
calls to standard modules, but not go any further down.

Update a test for printCallGraph to expect this.


daviditen:

benharsh:
Will this catch ``private`` functions?

daviditen:
It does print `private` functions.

```
private proc isPrivate() { }

proc main {
  isPrivate();
}
```

Outputs:
```
main
  isPrivate
```

benharsh:
OK, looks fine to me!

daviditen:
Thanks!

PR #7470, opened 2017-09-25
Add 'mason env' command

benharsh:
The 'mason env' command will print environment variables recognized by
mason, similar to 'printchplenv'.

The default for ``MASON_HOME`` is changed from ``$HOME`` to
``$HOME/.mason``. This was likely an oversight during initial
development.

A new variable, ``MASON_REGISTRY``, has been added. This variable will
point to the registry from which to pull package information. The
default is ``https://github.com/chapel-lang/mason-registry``. It must be
a valid git URL (i.e. it could be a local file path).

If ``MASON_REGISTRY`` is changed between updates, the local copies of the registry and sources will be removed.

Testing passes on linux64 and darwin for test/mason/


test/mason/env/mason-registry.chpl
```diff
@@ -0,0 +1,87 @@
...
+
+  runCommand("git add .", true);
+  runCommand("git commit -m \"nothing\"", true);
+
+  here.chdir(oldCWD);
+}
+
+proc main() {
+  // Clear environment for testing
+  unsetEnv("MASON_HOME");
```
ben-albrecht: A helper functions like `clearMasonEnvs()` would be useful for these tests and future tests. Not necessary at this moment though.

tools/mason/MasonUtils.chpl
```diff
@@ -58,35 +60,19 @@ proc runWithStatus(command): int {
...
-  const envHome   = getEnv("MASON_HOME");
-  const home      = getEnv('HOME');
-  const masonHome = if envHome != "" then envHome else home;
-
-  const dotMason = masonHome + "/.mason";
-
-  if isDir(dotMason) == false {
-    writeln(dotMason + " not found, creating...");
-    mkdir(dotMason, parents=true);
+  for a in args {
```
ben-albrecht: Why not use `_array.find()` ?

tools/mason/MasonUtils.chpl
```diff
@@ -35,13 +35,15 @@ proc getEnv(name: string): string {
 
 
 /* Uses the Spawn module to create a subprocess */
-proc runCommand(cmd) {
+proc runCommand(cmd, quiet = false) {
```
ben-albrecht: *nitpick:* spacing inconsistency -- typically args are squished: `quiet=false`.

tools/mason/MasonUpdate.chpl
```diff
@@ -75,6 +75,31 @@ proc genLock(lock: Toml, lf) {
...
+  }
+
+  const oldRegistry = gitC(MASON_CACHED_REGISTRY, "git config --get remote.origin.url", quiet=true).strip();
+
+  if oldRegistry != MASON_REGISTRY {
+    writeln("MASON_REGISTRY changed since last update:");
+    writeln("  old: ", oldRegistry);
+    writeln("  new: ", MASON_REGISTRY);
+    writeln();
+    writeln("Removing cached registry and sources to avoid conflicts");
```
ben-albrecht: Open Q: Should this be an interactive `[Y/n]` request?

ben-albrecht:
Are we able to check if the new `MASON_REGISTRY` is valid before removing the old one?

benharsh:
I'm not sure how we would do that within the current 'runCommand' framework. Could definitely be future work.

PR #7474, opened 2017-09-25
Update voidVariables.rst tech note to include void iterators

daviditen:
Update the tech note to include iterators that `yield _void` and iterators
that don't yield anything.  Update the example to include these cases.

Removed a TODO comment from the Iterators spec section that said iterators
currently require at least one yield.


daviditen:
Thanks again @benharsh!

PR #7477, opened 2017-09-26
Emphasize parens for deinit

vasslitvinov:
With #7375, we require deinit functions to be declared with parens,
both as class/record deinitializers and module deinit functions.

Emphasize that in the spec by writing parens next to "deinit"
where they are discussed.
Emphasize that in CHIP 10 explicitly.

While there, promote reduce intents from "initial implementation"
and indicate that they are available for coforalls too.

### DIFF
--- a/doc/rst/developer/chips/10.rst
+++ b/doc/rst/developer/chips/10.rst
@@ -679,7 +679,8 @@ Deinitialization
 ++++++++++++++++
 
 To specify actions that occur upon deinitialization, specify a ``proc
-deinit()`` method. This method must have zero arguments and cannot be called
+deinit()`` method. This method must have zero arguments and be declared
+with parentheses. It cannot be called
 directly. If a type does not specify a ``proc deinit()`` method, the compiler
 will generate one.
 It will be equivalent to a user-written ``deinit`` with an empty body.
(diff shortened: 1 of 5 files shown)


spec/Data_Parallelism.tex
```diff
@@ -264,8 +264,8 @@ \section{Forall Intents}
 \end{rationale}
 
 \begin{craychapel}
-An initial implementation of "reduce" intents is also available,
-which permits users to reduce values across iterations of a forall loop.
+To reduce values across iterations of a forall or coforall loop,
+"reduce" intents can be used.
```
lydia-duncan: This looks like it was intended to be a separate commit - was it accidentally committed?  (It seems reasonable to me, I just wanted to make sure you were aware it was included)

doc/rst/developer/chips/10.rst
```diff
@@ -679,7 +679,8 @@ Deinitialization
 ++++++++++++++++
 
 To specify actions that occur upon deinitialization, specify a ``proc
-deinit()`` method. This method must have zero arguments and cannot be called
+deinit()`` method. This method must have zero arguments and be declared
```
lydia-duncan: I would insert another "must" here, as in "This method must have zero arguments and must be declared with parentheses", though you certainly don't have to do so.

PR #7487, opened 2017-09-26
Enable --llvm-wide-opt with struct wide pointers

### LINKED ISSUE #6149: CHPL_WIDE_POINTERS: node16 not currently working
### Summary of Problem
There is LLVM assertion in compiler when  compiling with LLVM for multilocales (setting CHPL_COMM makes this work fine).

Here is message of assert:
```
chpl: /home/goovie/Dokumenty/Repozytoria/coodie_chapel/third-party/llvm/install/linux64-gnu/include/llvm/IR/Instructions.h:823: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
```

I did git bisect with hope of fixing this bug by myself and following commit introduces it:
[https://github.com/chapel-lang/chapel/commit/7af9b09aadb4c992051f0027a5a956d3c5648e90](url)

### Steps to Reproduce

**Source Code:**

```chapel
writeln(1000);
```

**Compile command:**
chpl test.chpl --llvm -o test

### Configuration Information

- Output of `chpl --version`:
chpl Version 1.14.0.7af9b09
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
HPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: none *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: cstdlib *
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: node16 *
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:

mppf:
Previous to this PR, --llvm-wide-opt has not been working, because `CHPL_WIDE_POINTERS=node16` was not working. See issue #6149.

But the packed wide pointer representation was introduced along with `--llvm-wide-opt` as a workaround for problems with LLVM optimizations not understanding that pointers in different address spaces can have different sizes. That was a serious issue in the LLVM 3.3 era but it appears to have been fixed in later LLVMs.

The packed wide pointer representation has never had much support, other than enabling the experimental `--llvm-wide-opt`. For one thing, the representation limits maximum scaling. For another, it appears to inhibit some C/LLVM optimization because the packed pointers lose type information.

Therefore this PR takes two steps:
 1. Completele remove CHPL_WIDE_POINTERS and support for packed wide pointers.
 2. Update `--llvm-wide-opt` support to work with a structure wide pointer representation
 3. Make use of new interfaces provided by LLVM to remove some workarounds
 4. Fix bugs and get `--llvm-wide-opt` in many cases.

I don't think `--llvm-wide-opt` has ever passed full local testing, and it hasn't worked at all for a release or so. So even if this PR doesn't get all tests passing with `--llvm-wide-opt`, it is an improvement to the current situation.

### What is --llvm-wide-opt?

See also doc/rst/technotes/llvm.rst, but --llvm-wide-opt activates a mode where Chapel generates wide pointers as address-space-100 pointers, and GET is load, PUT is store (or these could possibly be the LLVM memcpy instrinsic in more general cases). Then normal LLVM optimizations run and can do things like hoist a GET/load out of a loop. After LLVM optimizations run, we run llvmAggregateGlobalOps, which un-does some deoptimization that some other LLVM passes might do, and combines loads/stores to provably adjacent memory locations. Finally, we run llvmGlobalToWide, which "lowers" all of this address-space-100 business back to wide pointers and calls to runtime functions to GET and PUT.

### Performance Impact

* `performance/ferguson/remote-record-read-licm.chpl` shows LLVM optimizations hoisting a GET out of a loop
* observed no change in performance for stream-ep, lulesh with or without --llvm-wide-opt
* miniMD shows about 25% performance improvement with --llvm-wide-opt over C backend
* PTRANS shows about a 4x performance speedup with --llvm-wide-opt vs the C backend.

Elliot ran some performance testing with revision 3218725 and there were performance regressions that are fixed by later commits.

### Notes on particular changes:

dc82aa321a3354dce32f892220487d35c9d8f5c3 makes a few methods on AST classes const and adds Type::isWidePtrType() which enables a workaround in a later commit.
5c4c51474672dcca13b0490ea0a3053c27b69e1e uses Type::isWidePtrType in code generation of types to work around a hack in parallel.cpp and then 96ec460d5582bde616f61670de6c279a285605c7 adjusts codegenWideThingField to further work around this issue.
ad9319b3820798962c76b1c3c2817f3a672c54a5 cleans up how we use clang and separates out details of the clang invocation from GenInfo into a new ClangInfo structure. Additionally, it uses CreateLLVMCodeGen (new in 3.9) now instead of relying on internal Clang types. Unfortunately, I couldn't find a way to generate an llvm::Type* from a clang::Type* using the public API and so we still use some internal clang headers, but we are much closer to being API-clean, so to speak. We might be able to get this Type information and also GEP offset generation (the other missing piece) in to an additional interface for clang's CodeGenModule.
8783dbabb9a43cedc460a113b3b8b92d025d3668 switches over to running our own LLVM optimization pipeline, instead of relying on clang's pipeline. Without this change, I was unable to get the data layout string in the module to reflect that address space 100 pointers are 128 bits (and that causes all sorts of problems with the lowering in llvmGlobalToWide). Within EmitBackendOutput, the clang code *really* wants to replace the data layout string with whatever it thinks is appropriate. But we need to adjust it before/after the llvmGlobalToWide pass.
70b0a10918ea68691392b13627788ec36635e879 uses an alternative strategy to EmitBackendOutput so we don't run optimizations twice and 0520cedf5739a342ed0adb7c8541cce7ca4f3407 fixes some problems with detecting target Features (such as sse) and -mllvm parsing.
1313e3eafcbe5facb5aba9199e0d1f470f3bee4a was originally in PR #7496 and is included here so that this PR can be tested with LLVM 5 and to minimize conflicts. That PR has a good description of the change.

Resolves #6149.

Reviewed by @dmk42 - thanks!

Further TODOs (including getting more tests working with --llvm-wide-opt) are in #7523

- [x] full local testing with CHPL_LLVM=llvm


dmk42:
👍 
This PR includes a considerable amount of cleanup that should make things easier to read and maintain.  Nice.

I understand there may be more commits, but up to this point, I have read it all and it looks good to me.

mppf:
@dmk42 @ronawho - I'm planning to merge this PR as it stands now even though not all of the --llvm-wide-opt tasks are complete. At least then master will work with --llvm-wide-opt at all... I put the future TODOs in an issue and plan to continue to investigate.

Issue #7484, opened 2017-09-26
chpl_memalign() should check for out-of-memory.


gbtitus:
### Summary of Problem

`chpl_memalign()` and probably a few other Chapel runtime memory allocation functions don't do out-of-memory checking.

### Steps to Reproduce

Run the `users/shetag/fock/fock-3.chpl` test on linux32 in the default configuration with 8mb call stacks.  After about 150 runs it will segfault.  If you rebuild Qthreads with `--enable-debugging --enable-asserts` you can collect a core file and observe that a stack memory pool allocation has failed.  Qthreads doesn't notice this normally (without `--enable-asserts`) and tries to dereference the pointer, causing the segfault. Despite that Qthreads calls the Chapel runtime memory layer for pool memory, the expected error halt in the Chapel allocator for out-of-memory doesn't happen, because the runtime function Qthreads calls to get more memory is `chpl_memalign()` and that was not coded to check for allocation failure.  It should be.

This is the cause of the segfault (which should instead be an out-of-memory halt) described in [CHAPEL-316](https://chapel.atlassian.net/browse/CHAPEL-316).

gbtitus:
@mppf (originator of `chpl_memalign()`) and @ronawho (Qthreads, jemalloc, runtime in general), you might be interested in this.  I've assigned it to myself and am intending to do the work, though, so only looking for commentary and not actual help.

gbtitus:
Yes, it is still available and that would be great!  You can look at the other functions in the chpl_mem layer of the runtime to see what they do for out-of-memory.  (Basically at present it is just reported as an internal error.)

bradcray:
We had a user run afoul of this today, so I'm noting it here to give this a bump and also tagging it as a user issue.

PR #7490, opened 2017-09-26
Limit fock-3 call stack to 1mb to avoid OOM on low-memory systems.

gbtitus:
`users/shetag/fock/fock-3` has been failing intermittently (1 time in
about 150) with a segfault in linux32 testing on a system with 4gb of
memory and 4 cores.  It turns out that this is due to running out of
memory for Qthreads stack pools.  Here, as an interim measure, reduce
the call stack size for this test to 1 mb.  This passed 1000 runs on
the linux32 testing system without a failure.

This resolves [CHAPEL-316](https://chapel.atlassian.net/browse/CHAPEL-316).

#7484 is related.


gbtitus:
@dmk42 and/or @ronawho might be interested in this one.  Note that although it's probably quite innocuous, it's nevertheless marked "post-release".

dmk42:
👍

PR #7491, opened 2017-09-26
Update the tech note on reduce intents

vasslitvinov:
Bringing it up-to-date with the current master and the current thinking.

For example, all Open Issues that were listed there are resolved
right now, so I removed them.

While there, touched CHIP 10 to follow up on Lydia's feedback on #7477
by adding another "must" and re-filled that paragraph.

### DIFF
--- a/doc/rst/technotes/reduceIntents.rst
+++ b/doc/rst/technotes/reduceIntents.rst
@@ -18,22 +18,25 @@ As with any forall intent, a reduce intent can be specified on any
 *outer variable* - that is, a variable used within the body of a
 forall loop and declared outside that loop.  References to such a
 variable within the loop implicitly refer to the corresponding formal
-argument of the task function or the leader iterator.
+argument of the task function created by the parallel iterator.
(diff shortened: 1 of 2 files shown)


doc/rst/technotes/reduceIntents.rst
```diff
@@ -73,7 +81,7 @@ Increment ``x`` in the loop -- counts the number of iterations:
 
  .. code-block:: chapel
 
-  var x = 5;
+  var x = 0;
   forall myIterator() with (+ reduce x) {
     x += 1;
   }
```
bradcray: The writeln() that follows this loop no longer seems accurate due to the non-zero initial value of 'x'.

vasslitvinov: I changed the initial value for 'x' from 5 to 0, so this should be correct now. ?

bradcray: Oh wait, I read the diff backwards.  Nevermind.

PR #7496, opened 2017-09-27
Change LLVM dump() to print()

dmk42:
Beginning with LLVM 5, the dump() methods are no longer exposed for non-debug builds.  This change switches to using print() instead.

A lot of the LLVM community didn't know this (myself included) but dump() was only intended as something that could be run from inside gdb.  That is why it is now stripped from non-debug builds.

The print() methods take varying numbers of arguments depending on what is being printed.  The last argument is `isForDebug` which needs to be set to `true` to duplicate what we used to get from the dump() methods.

This change is sufficient to get LLVM 5 to work without setting CHPL_LLVM_DEVELOPER.  However, it will conflict with some of the files that @mppf is changing with the llvm-wide-opt PR.  Therefore, I will submit this PR as a record of what needs to be done, and then close it to get out of the way.

Passes full local testing with and without `--fast`, except for known errors in fft optimization that will need to be investigated (documented in #7246 ).

### DIFF
--- a/compiler/util/llvmAggregateGlobalOps.cpp
+++ b/compiler/util/llvmAggregateGlobalOps.cpp
@@ -49,6 +49,7 @@
 
 #include "llvm/Pass.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/Debug.h"
 #include "llvm/ADT/SmallSet.h"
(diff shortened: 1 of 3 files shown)


dmk42:
Closing to get out of @mppf 's way.

PR #7497, opened 2017-09-28
Fix bug where string assignment did not always add terminating zero byte

### LINKED ISSUE #7352: Issue recasting array value to integer over 5 digits
### Summary of Problem
Passing array to subroutine, which are string values (but numbers)

In subroutine, if I attempt to recast the Array[x] values directly to int(64) variables, and the numeric value is over 9999, the operation fails with:
unexpected character "l" when converting from string to int(64)  **(lowercase L)

If I put the string value in an intermediate string variable, the operation works as expected.

### Steps to Reproduce
Array containing string values "1001","3004","10004" (or pick your own). If any value of Array is over 9999, error is observed.

Fails with error:
testSub(A) {
   var a1 int = A[1] int;
   var a2 int = A[2] int;
   var a3 int = A[3] int;   (ERROR HERE)
}

Works:
testSub(A) {
   var as1 string = A[1];
   var as2 string = A[2];
   var as3 string = A[3];
   var a1 int = as1 int;
   var a2 int = as2 int;
   var a3 int = as3 int;
}

**Compile command:**
chpl -o test test.chpl   (same with --fast)

**Execution command:**
./test

### Configuration Information
- Output of `chpl --version`: chpl Version 1.15.0
- Output of `$CHPL_HOME/util/printchplenv --anonymize`: bash: CHPL_TARGET_PLATFORM:: command not found
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: gcc (GCC) 7.2.0
- (For Cray systems only) Output of `module list`:
Platform: Linux64 on Arch Linux

mppf:
The string assignment function (really in reinitString) wasn't setting the null
terminator at the end of a string in certain cases.

The included new test demonstrates the previous bug and now passes.

Closes #7352.

- [x] full local testing
- [x] valgrind testing of test/types/string
- [x] tested test/types/string with GASNet


daviditen:
Looks good to me.

PR #7501, opened 2017-09-28
Use paratest to speed up long running nightly chapcs jobs

ronawho:
Use paratest for gasnet everything/fast testing and increase the amount of
oversubscription we do valgrind. chapcs is our main dev/testing platform.
Speeding up nightly testing will let us fully utilize the system for our own
development paratests earlier in the day.


ronawho:
@benharsh quick sanity check?

PR #7514, opened 2017-09-30
Close leaks due to LocRadCache objects

bradcray:
It appears that we have not been cleaning up LocRadCache objects
in our Block, Cyclic, Stencil distributions.  Doing so seems to
be as easy as creating a deinit() routine that frees them if
they're non-`nil` on the local array descriptors.  This takes
MultiBlockDist2DLocales.chpl down to zero leaks.

TODO:
- [x] testing

### DIFF
--- a/modules/dists/BlockDist.chpl
+++ b/modules/dists/BlockDist.chpl
@@ -412,6 +412,11 @@ class LocBlockArr {
   inline proc unlockLocRAD() {
     locRADLock.clear();
   }
+
+  proc deinit() {
+    if locRAD != nil then
+      delete locRAD;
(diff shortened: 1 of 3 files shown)


bradcray:
No, I only did linux64 and valgrind.  How extensive do you think the gasnet testing should be?

benharsh:
I'd normally recommend a full paratest. A smaller subset could be:
- release/examples
- distributions/

vasslitvinov:
For gasnet, I often run paratest on the set of tests+dirs produced by `util/devel/test/findMultilocaleTests`

PR #7515, opened 2017-09-30
Fix memory leak caused by "".join(("",""))

cassella:
This currently results in returning a string with len 0 but buf[]
allocated and _size > 0.  The len being 0 results in isEmptyString()
returning true, so deinit() doesn't free the buf[].

Avoid this by handing a resulting joinedSize 0 as a special case, and
return a real empty string.

This fixes the leak in test/types/string/psahabu/join-tuples.chpl

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -890,6 +890,9 @@ module String {
         var joinedSize: int = this.len * (S.size - 1);
         for s in S do joinedSize += s.length;
 
+        if joinedSize == 0 then
+          return '';
+
         var joined: string;
         joined.len = joinedSize;
         const allocSize = chpl_here_good_alloc_size(joined.len + 1);


cassella:
I don't know if this is the best way to fix it, v.s. some other adjustment to the resulting string, and/or maybe changing ``isEmptyString`` to only consider ``buf`` or ``_size``, not ``len``.

I verified the leak is fixed in ``test/types/string/psahabu/join-tuples.chpl``. 
``start_test test/types/string`` reported no errors.

cassella:
(whoops, force push to fix the name of the test in the commit message.)

benharsh:
Testing on my end looks good. This actually fixes three leaks:
- types/string/kbrady/join.chpl
- types/string/psahabu/join-tuples.chpl
- types/string/psahabu/join-varargs.chpl

Thanks!

PR #7516, opened 2017-10-02
Adjust cpp-lines test to tolerate environment variable being set

mppf:
In some testing configurations, the perftools-base module is setting
CHPL_CG_CPP_LINES=1. That was causing failures for noCppLines.

This commit makes this test more robust by passing --no-cpp-lines
which has the effect of overriding the environment variable.

While there, made these tests .noexec since the only useful thing being
tested here is whether or not #line numbers were emitted in the generated C,
and that doesn't require that the program be run.

### DIFF
--- a/test/compflags/shannon/cpplines/cppLines.good
+++ b/test/compflags/shannon/cpplines/cppLines.good
@@ -1,2 +1,1 @@
-m = 10
 NOTE: #line directives were generated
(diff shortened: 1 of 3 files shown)


awallace-cray:
Fine by me, from a release-management point of view.

mppf:
@ronawho has another strategy he likes better, so closing this PR.

PR #7518, opened 2017-10-02
Fix dynamic dispatch of serial iterators with nesting deeper than 2

daviditen:
When the class hierarchy is at least three levels deep with an iterator
overloaded at each level, the compiler was tripping an assert. This was
because the iterator classes for both the parent and grandparent iterators
were getting added to the child iterator class's dispatchParents list.

Check that the classes the iterators are defined in have a direct subclass
relationship before linking the iterator classes.


daviditen:
@noakesmichael would you mind reviewing this?

PR #7520, opened 2017-10-02
Fix some sparse matrix-matrix mult bugs

benharsh:
1) SPA used the range '0..#cols', when it should use the first dimension
of A's domain. Otherwise there would be OOB errors if the array wasn't
zero-based.

2) The resulting sparse array did not sort its indices internally. This
would result in various failures. For example, domain membership checks
would fail despite those indices being yielded from the domain's
iterator. The fix is to sort SPA.ls before inserting into 'C'.

Testing:
- [x] LA primer
- [x] modules/packages/linearAgebra/correctness

### DIFF
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -1225,7 +1225,7 @@ module Sparse {
   pragma "no doc"
   /* Sparse-accumulator */
   record _SPA {
-    var cols: range;
+    var cols; // range(?)
     type eltType = int;
     var b: [cols] bool,      // occupation
         w: [cols] eltType,   // values
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -1189,7 +1189,7 @@ module Sparse {
     const nnzAB = Adom._value.nnz + Bdom._value.nnz;
     Cdom._value.nnzDom = {1..nnzAB};
 
-    var spa = new _SPA(cols=D.dim(1).size, eltType=A.eltType);
+    var spa = new _SPA(cols=D.dim(1), eltType=A.eltType);
 
     /*
      IR (row)     - nnz-rows  - A.domain._value.startIdx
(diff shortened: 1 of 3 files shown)


modules/packages/LinearAlgebra.chpl
```diff
@@ -1225,34 +1225,37 @@ module Sparse {
   pragma "no doc"
   /* Sparse-accumulator */
   record _SPA {
-    var cols: int;
+    var cols: range;
```
bradcray: IIRC, I think that this field will necessarily be a bounded, non-stridable range.  It might make sense to have it be a generic field in order to support domains that are stridable (unless we know for some reason that they will not be).  Really, we'd like to declare this field to be of type `range(?)` but I think at present, that only works in formal argument lists, not in field declarations (?).  I usually handle cases like this where I control the callsites by using a commented out type: `cols /* : range(?); */;`

bradcray:
That's the only comment I have -- thanks for taking care of this.

bradcray:
Thinking about the performance hit that this took yesterday, it occurs to me belatedly that we might want to have CS domains not store each row in sorted order (or have a param that permits that option) in order to avoid the overhead of Sort.  I did it in order to get O(lg(nnz/n)) random access time, but given that nnz/n is typically a small, this may be a case of premature / unnecessary optimization.

benharsh:
Maybe we could add a 'sortIndices' method for users that want faster random access once they're done adding indices?

bradcray:
That's a possibility, though it'd require dynamic checks (conditionals) against whether or not the indices were sorted along key code paths.  But maybe such code paths are already heavyweight enough that the conditional wouldn't matter (though it would add code bloat).  That's what made me suggest the param.

My conjecture, though, is that maybe the binary search within a row is not worth the complexity given that rows will typically not be very long and the cache might be happier if we just ripped through them consecutively...  It's pretty unusual for CS* formats to store row contents in sorted order which is partially why I suggested that I might have been optimizing prematurely.

benharsh:
What about serial iteration over a CSR domain? Usually rectangular domains yield their indices in-order for serial iteration, but perhaps sparse is sufficiently different such that we can yield in an undefined order?

bradcray:
Hmm, that's an interesting question and point.  I could be open to having them yield in an unusual order.  I'm guessing that CSC domains might also yield in an unusual CMO serial order currently (unless they sort prior to yielding?).  Still thinking about this...

ct-clmsn:
@benharsh to be a little off topic - some good reference implementations from the C++ community for linear algebra libraries are [blaze](https://bitbucket.org/blaze-lib/blaze) and [Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page).

PR #7522, opened 2017-10-02
Increase the number of updates for RA to stabilize performance timings

ronawho:
We used to just do 10 million updates for all ra variations, but ra-rmo is very
fast under ugni/gasnet-aries and ra-atomics is very fast under ugni. Bump the
number of updates to 1 billion for those configs to stabilize the results
(they're very noisy right now.)

Note that this also turns off verification. For ra-on and ra-rmo the
verification is ~50 times slower than the actual kernel and takes too long to
run at problem sizes that give us stable kernel timings.


ronawho:
@benharsh mind reviewing? 

Note that I plan on changing the nightly .dat files to match the new expected names.

This is a little ugly, but I think it's the cleanest way to get stable timings. Note that I'm using a .ml-compopts because `useOn` is a param. Using the .ml-compopts also requires casting `n` and `N_U` to the `indexType` since they aren't `int`s in the source code.

PR #7527, opened 2017-10-03
When running gasnet testing with valgrind, use its suppression file

bradcray:
Unfortunately, this does not result in clean execution for simple
hello world programs, but it's better than nothing.  As a background
task, I'm trying to determine what we can do to tighten this up (with
some help from the GASNet team as well).

### DIFF
--- a/util/test/sub_test
+++ b/util/test/sub_test
@@ -798,6 +798,8 @@ if (os.getenv('CHPL_TEST_VGRND_EXE')=='on' and vgrbin):
     valgrindbinopts = chpl_valgrind_opts.split()+['-q']
     if (chplcomm!='none'):
         valgrindbinopts+=['--trace-children=yes']
+        if (chplcomm=='gasnet'):
+            valgrindbinopts+=['--suppressions=%s/third-party/gasnet/gasnet-src/other/contrib/gasnet.supp'%(chpl_home)]
 else:
     valgrindbin = None
     valgrindbinopts = None


bradcray:
@ronawho : Totally forgot about this one.  Will you give it a quick look?  It's a bit of a GIGO PR because we don't get clean runs even with this suppression file (and it'll take more work to get them cleaner), but it seems more obviously correct to utilize the suppression file than to not do so.

util/test/sub_test
```diff
@@ -798,6 +798,8 @@ if (os.getenv('CHPL_TEST_VGRND_EXE')=='on' and vgrbin):
     valgrindbinopts = chpl_valgrind_opts.split()+['-q']
     if (chplcomm!='none'):
         valgrindbinopts+=['--trace-children=yes']
+        if (chplcomm=='gasnet'):
+            valgrindbinopts+=['--suppressions=%s/third-party/gasnet/gasnet-src/other/contrib/gasnet.supp'%(chpl_home)]
```
ronawho: Nit. sub_test has a hodgepodge of formatters so it doesn't really matter, but the new/preferred way to format strings in python is with .format():
```python
valgrindbinopts+=['--suppressions={0}/third-party/gasnet/gasnet-src/other/contrib/gasnet.supp'.format(chpl_home)]

```

bradcray: OK.  Not knowing a thing about Python, I just pattern matched off of line 789.  Do you want me to change it/them?

bradcray:
> I'm surprised we weren't already doing this considering it's documented in doc/rst/developer/bestPractices/GASNetOnDesktops.txt

Yeah, I'm pretty sure that we used to and that it got lost somewhere along the way (maybe in the Python rewrite?)

util/test/sub_test
ronawho: Up to you. sub_test is enough of a mismatch that I don't mind either way, was more of an fyi then something that needs to be fixed.

ronawho:
>>I'm surprised we weren't already doing this considering it's documented in doc/rst/developer/bestPractices/GASNetOnDesktops.txt

>Yeah, I'm pretty sure that we used to and that it got lost somewhere along the way (maybe in the Python rewrite?)

Could be, though Sung rewrote sub_test years and years ago. The more recent rewrite of start_test shouldn't have affected this. Holy crap.. that "recent" rewrite was over 2 years ago..

`git rev-list --all | xargs git grep gasnet.supp -- util` isn't turning up any references in the util dir either. Looks like we just never thought of it.

PR #7529, opened 2017-10-03
Fix nightly gasnet-{everything,fast} testing

ronawho:
We started using paratest for nightly gasnet testing with #7501, but nightly
hasn't used paratest for gasnet testing in a long time. It was still trying to
call paratest with `-comm gasnet`, which was removed years ago with c79342b81b.

Stop throwing `-comm` in nightly to fix this.


ronawho:
@benharsh quick sanity check?

PR #7548, opened 2017-10-04
Add dsiDestroy* calls to BlockSparse Dom and Arr fields

bradcray:
It seems we haven't had cleanup steps for distributed sparse
domains and arrays, so any codes using them have been leaking
them.


bradcray:
@e-kayrakli : Could you give this a look and let me know if anything seems amiss or if there's an obvious reason not to clean up in this way?  (I essentially copied the code straight from BlockDist.chpl).  Thanks!

PR #7552, opened 2017-10-05
Replace initialize methods on 9 tests and their helpers

lydia-duncan:
The calls to create instances of these classes and records involved a
single call style that did not provide a value for every field.  Often, the
initialize method was responsible for setting the contents of one of the
fields that did not get passed a corresponding argument, regardless of
its previous contents - meaning that there was no point having an
argument to the initializer for that field.  This change turns those
initialize() methods into a single initializer per type, with the contents
of the old initialize() method used as its Phase 2.

Details:
arrays/dinan/init_arraymember/test3.chpl and
arrays/dinan/init_arraymember/test6.chpl - 2 of 6 variations on setting
the value of an array field that is dependent on a previous domain field.
None of the other variations contain constructors, but they do rely on the
default constructor and the value of the field would be just overwritten by
the initialize() method - it seems likely that the initialize method is
interchangeable with a constructor for the purpose of these tests.  Make
the initialize() method in test3 into an initializer where the operation occurs
in Phase 1 (arbitrarily) - if Phase 1 were the default then the super.init()
call would not be needed.  Make the initialize() method in test6 into a Phase
2 only initializer (also arbitrarily).

classes/bradc/overloadMethods/v1NoReturnValue.chpl:
The default constructor for this test was called with just the param argument
and the initialize method overwrote the value of the other field without
inspecting its prior contents.  Convert the initialize method into an initializer
with just a param argument, and have its old body compose Phase 2.  This
initializer would benefit from being able to set generic fields during Phase 2.

Note: I was not able to convert the other six v*.chpl tests in that directory, as
they exposed a bug with generic initializers and inheritance impacting the
virtual dispatch of method calls.

distributions/diten/testCyclic.chpl's helper file, Cyclic.chpl:
The only `new` calls on the Cyclic1DDom and Cyclic1DArr relied on three
of the four fields, while the initialize() method on each overwrote any
potential previous value for the last field without checking its contents.
Change these initialize methods into initializers with only arguments for
the three fields that were explicitly set.  The initializer on Cyclic1DDom
would benefit from being able to set generic fields in Phase 2 and from
being able to set const fields in Phase 2, while the initializer on
Cyclic1DArr would benefit from being able to set generic fields in Phase
2.

multilocale/deitz/linked/test_linked1.chpl,
multilocale/deitz/linked/test_linked2.chpl,
multilocale/deitz/linked/test_linked3.chpl,
multilocale/deitz/linked/test_linked4.chpl:
These four tests all called the default constructor with no arguments and
the initialize() method overwrote the field without checking its contents.
Change these methods into zero argument Phase 2 only initializers.

users/shetag/assocArrayOfArithArrayInClass.chpl:
The operations of it are entirely Phase 2 like, and the default constructor was
called with zero arguments, so make it an initializer with no arguments and a
Phase 2 only body

Tested over a standard paratest (due to the modification of helper
files which might be used elsewhere).

### DIFF
--- a/test/studies/hpcc/common/bradc/BradsBlock1D.chpl
+++ b/test/studies/hpcc/common/bradc/BradsBlock1D.chpl
@@ -38,10 +38,10 @@ class Block1DDist {
 
   //
   // an associative array of local distribution class descriptors --
-  // set up in initializer below
+  // set up in initialize() below
   //
   // TODO: would like this to be const and initialize in-place,
-  // removing the initializer
+  // removing the initialize method
   //
   // TODO: Remove second generic parameter -- seems confusing and wrong.  Replace
   // with explicit typing of locid field in LocBlock1DDist class.  Particularly
@@ -49,13 +49,7 @@ class Block1DDist {
   //
   var locDist: [targetLocDom] LocBlock1DDist(glbIdxType, index(targetLocs.domain));
 
-  proc init(type idxType = int(64), bbox, targetLocs,
-            targetLocDom = targetLocs) {
-    glbIdxType = idxType;
-    this.bbox = bbox;
-    this.targetLocs = targetLocs;
-    this.targetLocDom = targetLocDom;
-    super.init();
+  proc initialize() {
     for (loc, locid) in zip(targetLocs, 0..) do
       on loc do
         locDist(loc) = new LocBlock1DDist(glbIdxType, locid, this);
@@ -177,19 +171,14 @@ class Block1DDom {
 
   //
   // an associative array of local domain class descriptors -- set up
-  // in initializer below
+  // in initialize() below
   //
   // TODO: would like this to be const and initialize in-place,
-  // removing the initializer
+  // removing the initialize method
   //
   var locDom: [dist.targetLocDom] LocBlock1DDom(glbIdxType, lclIdxType);
 
-  proc init(type idxType, type locIdxType, myDist, myWhole) {
-    glbIdxType = idxType;
-    lclIdxType = locIdxType;
-    dist = myDist;
-    whole = myWhole;
-    super.init();
+  proc initialize() {
     for loc in dist.targetLocs do
       on loc do
         locDom(loc) = new LocBlock1DDom(glbIdxType, lclIdxType, this, dist.getChunk(whole));
@@ -326,16 +315,11 @@ class Block1DArr {
   // an associative array of local array classes, indexed by locale
   //
   // TODO: would like this to be const and initialize in-place,
-  // removing the initializer
+  // removing the initialize method
   //
   var locArr: [dom.dist.targetLocDom] LocBlock1DArr(glbIdxType, lclIdxType, elemType);
 
-  proc init(type idxType, type localIdxType, type eltType, myDom) {
-    glbIdxType = idxType;
-    lclIdxType = localIdxType;
-    elemType = eltType;
-    dom = myDom;
-    super.init();
+  proc initialize() {
     for loc in dom.dist.targetLocs do
       on loc do
         locArr(loc) = new LocBlock1DArr(glbIdxType, lclIdxType, elemType, dom.locDom(loc));
--- a/test/studies/hpcc/common/bradc/BradsBlock1D.chpl
+++ b/test/studies/hpcc/common/bradc/BradsBlock1D.chpl
@@ -38,10 +38,10 @@ class Block1DDist {
 
   //
   // an associative array of local distribution class descriptors --
-  // set up in initialize() below
+  // set up in initializer below
   //
   // TODO: would like this to be const and initialize in-place,
-  // removing the initialize method
+  // removing the initializer
   //
   // TODO: Remove second generic parameter -- seems confusing and wrong.  Replace
   // with explicit typing of locid field in LocBlock1DDist class.  Particularly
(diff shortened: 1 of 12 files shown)


test/studies/hpcc/common/bradc/BradsBlock1D.chpl
```diff
@@ -38,18 +38,24 @@ class Block1DDist {
...
   var locDist: [targetLocDom] LocBlock1DDist(glbIdxType, index(targetLocs.domain));
 
-  proc initialize() {
+  proc init(type idxType = int(64), bbox, targetLocs,
+            targetLocDom = targetLocs) {
+    glbIdxType = idxType;
+    this.bbox = bbox;
+    this.targetLocs = targetLocs;
+    this.targetLocDom = targetLocDom;
+    super.init();
```
bradcray: This seems like the kind of case where being able to supply a postInit()-style hook would be beneficial—haven't you essentially written the compiler-provided initializer explicitly here?

Patterns like this are the kind that I would propose putting off rewriting pending further decisions regarding enabling users to leverage the compiler-provided initializer or create a postInt() hook.

lydia-duncan: It's missing the last field.

This one only got called with two arguments, but it seemed like we would probably want the other two in some situations (otherwise the targetLocDom would always match targetLocs, and the index type would always be int).

I'm not bothered by rewriting four fields, personally.  5+ is where I draw the line.  But I can revert this change if you prefer.

bradcray: I don't think the point should be how the initializer is called in practice in the test system (particularly for code that is library-like for which the test may not show all calls that a client of the library might make); nor how many fields you/the author would have to explicitly initialize.  Instead, I think we should focus on identifying cases where the compiler-provided initializer was being leveraged by initialize() in a way that permitted the user to avoid rewriting what the compiler provided (or even 80% of it) themselves.

I'm OK with this change going in, I'm just not comfortable with writing it off as though it was not an interesting use case for initialize().

lydia-duncan: Sorry, I'm using these metrics as short hand for the larger point.  Let me go into a little more detail on what I mean.

> I don't think the point should be how the initializer is called in practice in the test system

In normal cases, I think having a single call style for the creation of the type indicates either a short-sightedness on part of the test author (didn't realize that their type could be initialized without their preprogrammed field relationships), or that we need more unit tests on the type to ensure proper behavior.

Having the initialize method as a fall back enables a lazy response to being confronted with this case, which is why I called it out in my commit message rather than sweeping it entirely under the rug.

> (particularly for code that is library-like for which the test may not show all calls that a client of the library might make);

That is very reasonable.

> nor how many fields you/the author would have to explicitly initialize.

I think we're saying the same thing here: if the amount of complexity needed is sufficient to warrant more than just the default initializer, there is a particular problem space for "I want a post-hook" versus "I will gladly write my own initializer".  The number I gave should be thought of more as the number of fields that aren't impacted by the code you want to add - if it's small, I wouldn't even consider an initialize method myself.

lydia-duncan: > The number I gave should be thought of more as the number of fields that aren't impacted by the code you want to add - if it's small, I wouldn't even consider an initialize method myself.

Err, it's kinda an interplay between percentage and number of fields?  The default initializer being sufficient for 50% of fields is one thing when there are 14 fields versus 2.

vasslitvinov:
Re test3.chpl and test6.chpl - from a quick glance at the revision history, these tests were added in 0602b0f , >10 years ago, with this commit message:

    Several tests that attempt to initialize an array member from a function.
    Most of these are futures except for test6.chpl which uses initialize() as a
    workaround.

I did not see significant modifications since then.

If these tests merely test initialization of array fields, using `initialize` as a workaround, then converting them to proper initializers and removing initialize() is adequate.

vasslitvinov:
Re Cyclic.chpl: the changes in this PR are fine.

Both initializers are great examples of "do whatever the default initializer does, then execute <this code>". At least for rapid prototyping, even if a "polished"/production initializer may be more meaningful than that.

@lydia-duncan , I think you are keeping track of tests that could benefit from that? If so, please add this Cyclic.chpl to that list.

Setting generic/const fields in Phase 2 is one way to write this initializer. Other ways that would be convenient are initializing array elements in Phase 1 (which would allow `locDoms` and `locArr` to be declared `const` btw) and/or automatic inference of Phase 1/2 boundary.

vasslitvinov:
The changes in BradsBlock1D* are OK if we are certain that we want to get rid of initialize(). Otherwise, like Brad suggests, let's hold off.

Akin to Cyclic.chpl, the Phase 2 of the initializers here are merely initializing array fields. We should find a good way to do so in Phase 1.

vasslitvinov:
OK to merge modulo BradsBlock1D*

lydia-duncan:
> Both initializers are great examples of "do whatever the default initializer does, then execute ". At least for rapid prototyping, even if a "polished"/production initializer may be more meaningful than that.
>
> @lydia-duncan , I think you are keeping track of tests that could benefit from that? If so, please add this Cyclic.chpl to that list.

If by keeping track you mean as part of my list of initialize() methods, then yes, I have been.  I haven't noted cases that were not initialize() methods to begin with, though.

Is there a conclusion you want from that list other than "maybe we need an initialize method equivalent"?  It kinda sounds like you are arguing for our inserted omitted initialization to mimic the default initializer behavior (inserting arguments for the omitted ones and setting the omitted fields to their values), but I'm not sure where I got that impression ...

The array fields operation is a common pattern, it will be interesting to see our conclusions once we have considered all example initializers in our test and module code.

lydia-duncan:
Thanks for the review!

vasslitvinov:
@lydia-duncan I am arguing that we may want to support the use case(s) that currently rely on the initialize() method. Maybe one way to say it is "do what the default initializer does, then perform these actions." The examples on your list - thank you for keeping it - should help us understand what language support/features we want to provide.

PR #7554, opened 2017-10-05
Remove Multi-ddata from DefaultRectangular

benharsh:
This PR removes multi-ddata from our modules, which has been disabled since #6390.

Testing:
- [x] full local
- [x] full numa
- [x] full gasnet
- [x] valgrind


bradcray:
I missed the context on this one—did we decide that we're unlikely to want to resurrect multi-ddata in the future?  Did @gbtitus agree?

gbtitus:
I did agree that it was okay (and am the reviewer).  @benharsh and I had discussed that in its current disabled state, the code is both in the way and not being tested, a tough combination. 
@ronawho has also pressed for removal, noting that after all, the code will be in the history if we want to revive it some day.

PR #7559, opened 2017-10-06
Fix memory leaks in Crypto tests

daviditen:
The tests for the Crypto module create, but don't delete classes, so they
leak memory. Add delete statements so they don't leak.

### DIFF
--- a/test/modules/packages/Crypto/saru/aes/aesTest1.chpl
+++ b/test/modules/packages/Crypto/saru/aes/aesTest1.chpl
@@ -27,4 +27,15 @@ proc main(){
   /* Decrypt the message using the key and IV */
   var orig = a.decrypt(ct, key, iv);
   writeln("Obtained Plaintext: ", orig.toHex());
+
+  /* Clean up allocated classes */
+  delete orig;
+  delete ct;
+  delete msg;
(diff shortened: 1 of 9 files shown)


daviditen:
@saru95 Does this look ok to you?  You can run the tests before and after this PR with commands like:

```
cd test/modules/packages/Crypto/saru
chpl aes/aesTest1.chpl
aesTest1 --memLeaks
```

To see the leaks / see them fixed.

daviditen:
I think we should think about switching all of these data types to be records instead of classes so they will manage their own memory, but this fixes the leaks in the near term.

saru95:
Oh yeah. I totally overlooked this. Adding `delete` at the end of each test avoids this. However, do we expect the user to do that every time he/she uses the `Crypto` module?

daviditen:
If the various types stay implemented as classes, then yes, the user needs to `delete` them or they will leak.

If we switch them to records with carefully defined copies and deinitializers they should be able to clean up after themselves without `delete`.

saru95:
You could try merging this for now. Meanwhile, I'll try on working on a version using `record`.

daviditen:
OK, thanks!

PR #7560, opened 2017-10-08
Added symmetric cipher Blowfish

saru95:


ronawho:
@daviditen Could you review this?

test/modules/packages/Crypto/saru/blowfish/bfTest2.skipif
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
ronawho: Could you make this a symlink to aesTest1.skipif like we did in https://github.com/chapel-lang/chapel/pull/7377?

modules/packages/Crypto.chpl
```diff
@@ -638,6 +639,169 @@ module Crypto {
...
+
+  class Blowfish {
+    pragma "no doc"
+    var cipher: CONST_EVP_CIPHER_PTR;
+    pragma "no doc"
+    var byteLen: int;
+
+    /* The `Blowfish` class constructor that initializes the Blowfish encryption
+       algorithm with the right key length and chaining mode.
+
```
daviditen: It would be nice to list the valid modes in this comment along with what they mean. Even just spelling out what the acronyms stand for would be good.

test/modules/packages/Crypto/saru/blowfish/bfTest1.chpl
```diff
@@ -0,0 +1,19 @@
...
+  var iv = new CryptoBuffer("iv");
+  var key = new CryptoBuffer("key");
+
+  writeln("MSG: ", msg.toHex());
+
+  var ct = bf.encrypt(msg, key, iv);
+  writeln("CT: ", ct.toHex());
+
+  var pt = bf.decrypt(ct, key, iv);
+  writeln("PT: ", pt.toHex());
```
daviditen: Please avoid leaking your classes by calling delete on all allocated classes like in #7559.  You can check for leaks by compiling the tests and running them with the `--memLeaks` flag.

```
chpl bfTest1.chpl
bfTest1 --memLeaks
```

modules/packages/Crypto.chpl
```diff
@@ -638,6 +639,179 @@ module Crypto {
...
+                      ciphertextLen: c_int);
+    EVP_DecryptFinal_ex(ctx,
+                        c_ptrTo(plaintext[plaintextLen..]): c_ptr(c_uchar),
+                        c_ptrTo(updatedPlainLen): c_ptr(c_int));
+
+    plainDomain = {0..#(plaintextLen + updatedPlainLen)};
+    return plaintext;
+  }
+
+  class Blowfish {
```
daviditen: Looking at this again, I think a slightly edited version of the comment for the `init()` function would be good as a comment for the `Blowfish` class.  Then the comment for `init()` could be simplified to something like "Initialize an instance of the Blowfish class. Arguments: ... Returns: ... Return type: ..."

test/modules/packages/Crypto/saru/blowfish/bfTest2.skipif
```diff
@@ -0,0 +1 @@
+../aes/aesTest1.skipif
```
daviditen: This should be a symlink to `../aes/aesTest1.skipif` instead of a file containing that text. e.g.

`ln -s ../aes/aesTest1.skipif bfTest2.skipif`

daviditen:
@saru95 I think this looks good.  I'd still like to see a comment on the `Blowfish` class - the one currently on `Blowfish.init()` would work without too much of a change.

Then with the symlink change I just commented on this should be good to go.

test/modules/packages/Crypto/saru/blowfish/bfTest2.skipif
saru95: This is what I had done before putting the text. xD

saru95:
Pushed the requested changes. Extremely sorry that I overlooked your reviews. Was caught with something important at work. 💯

daviditen:
Thanks for the updates @saru95.

As a sanity check I ran a full Linux64 test run and got:

`[Summary: #Successes = 7918 | #Failures = 0 | #Futures = 0]`

I also ran the two new blowfish tests on OSX and both passed.  I think this looks good to merge.

daviditen:
@saru95 the bfTest1 test appears to have a memory corruption problem.  It failed our nightly LLVM testing and valgrind testing last night. Would you please look into the failures?  I suspect fixing the valgrind errors will also fix the LLVM errors, so I recommend starting there.

The valgrind testing runs with:
```
export CHPL_TASKS=fifo
export CHPL_MEM=cstdlib
```

(rebuild Chapel after setting these)

With `CHPL_MEM=cstdlib` the output has the `CT` and `PT` lines appearing to be random, changing each time it is executed.

Here is the [valgrind output](https://gist.github.com/daviditen/63d7a4dd5fbd5b56082321b701f3b958) from:
```
chpl bfTest1.chpl
valgrind bfTest1
```

saru95:
@daviditen Is it only `bfTest1.chpl`?

daviditen:
It looks like `bfTest2.chpl` fails in a similar way.  We just didn't run it in our valgrind testing because openssl is too old on that machine, so the .skipif blocked it.

vasslitvinov:
@saru95 - the function `EVP_CIPHER_CTX_init` is declared `void` in the system header file and the man page. Please adjust the `extern` declaration in Crypto.chpl to match.

saru95:
@vasslitvinov It is already declared as a `void`. This PR https://github.com/chapel-lang/chapel/pull/7390 did that.

daviditen:
@saru95 Sorry for the delayed response - for some reason I didn't get a notification about your most recent reply.  I've spent some time digging into this, and I believe the problem is that the `IV` needs to be 8 bytes long.

The [BF_encrypt man page](https://www.openssl.org/docs/man1.1.0/crypto/BF_encrypt.html) isn't talking about the `EVP_EncryptInit_ex` function directly, but it mentions 8 byte requirements for the `IV` for other functions:

> BF_cbc_encrypt() is the Cipher Block Chaining function for Blowfish. It encrypts or decrypts the 64 bits chunks of in using the key schedule, putting the result in out. enc decides if encryption (BF_ENCRYPT) or decryption (BF_DECRYPT) shall be performed. ivec must point at an 8 byte long initialization vector.

If you can confirm this 8 byte requirement it would be nice to have `Blowfish.encrypt()` check the length of `IV` and issue an error if it is wrong.

PR #7563, opened 2017-10-09
Close a series of reasonably obvious user-level memory leaks

bradcray:
Close a number of memory leaks in user-level code that represent the top of our second tier of leakers.


bradcray:
@daviditen and @ben-albrecht may want to take a quick look at these since they're their tests (though I don't think the changes are likely to be controversial).

PR #7567, opened 2017-10-10
chplspell dictionary updates

cassella:
chplspell dictionary updates

### DIFF
--- a/util/devel/chplspell-dictionary
+++ b/util/devel/chplspell-dictionary
@@ -1115,6 +1115,7 @@ subminor
 subobject
 subparts
 substring
+subtarget
 subtask
 suppresswarnings
 suseconds
@@ -2735,8 +2736,10 @@ ures
 urse
 
 FILEID: 4f065520-1d85-11e6-a3a6-10ddb1d4c3d5
+intented
 orse
 shadowvar
+ssvar
 ytemp
 
 FILEID: 803ee256-1d85-11e6-a3a6-10ddb1d4c3d5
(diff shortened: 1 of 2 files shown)


util/devel/chplspell-dictionary
```diff
@@ -2735,8 +2736,10 @@ ures
 urse
 
 FILEID: 4f065520-1d85-11e6-a3a6-10ddb1d4c3d5
+intented
```
lydia-duncan: I'm inclined to say that "intented" as a word we made up does not follow English grammar rules, since an "intent" is a property and thus a noun, rather than a verb that can have "ed" applied to it.  I think the relevant section should read "then we need to create a shadow variable with a reduce intent for this forall" instead, and would bring it up to the author.

util/devel/chplspell-dictionary
```diff
@@ -5645,6 +5670,7 @@ malayalam
 malloc
 maltzahn
 mandelbrot
+mannari
```
lydia-duncan: Where does this one come from?

cassella: A new entry in the contributors file.

util/devel/chplspell-dictionary
cassella: The use was intentional, and I don't think my writ extends to editorializing.  So in cases like this, I've been adding err, chapelisms, like this to the dictionary.   e.g. densified, detupled, and anded.

I can remove this one from the PR.

util/devel/chplspell-dictionary
lydia-duncan: Cool, that makes sense (and was kinda what I was guessing)

util/devel/chplspell-dictionary
lydia-duncan: With the precedent, that's fair.  I think leaving it as is would be fine, then.  I'll go ahead and merge this PR, thanks!

Issue #7570, opened 2017-10-11
Curl setopt - error: call to throwing function ioerror without throws, try, or try!


ljdursi:
Hi, all:

### Summary of Problem

In trying to use the setopt function with curl, at compilation time a compile error is thrown about setopt calling ioerror without a throws/try/try!:

```
$ chpl -o testcurl testcurl.chpl
$CHPL_HOME/modules/packages/Curl.chpl:302: In function 'setopt':
$CHPL_HOME/modules/packages/Curl.chpl:313: error: call to throwing function ioerror without throws, try, or try! (relaxed mode)
$CHPL_HOME/modules/standard/SysError.chpl:407: note: throwing function ioerror defined here
```

### Steps to Reproduce

**Source Code:**

```chapel
use Curl;

proc main() {
  const uniprotid = "P07204_TRBM_HUMAN";
  const uniproturl = "http://www.uniprot.org/uniprot/" + uniprotid + ".fasta";
  var response = "";

  try! {
    var c = open(url=uniproturl, mode=iomode.r);
    c.setopt(curlopt_followlocation, 1);
    var reader = c.reader();
    var line: string;

    while (reader.readline(line)) do
      response += line;
  }

  writeln(uniprotid);
  writeln(response);
}
```

**Compile command:**

`chpl -o testcurl testcurl.chpl`

### Fix?

A crude change to get setopt to compile seems to work:

```
302c302
< proc file.setopt(opt:c_int, arg):bool {
---
> proc file.setopt(opt:c_int, arg):bool throws {
313c313,315
<   if err then ioerror(err, "in file.setopt(opt:c_int, arg)");
---
>   try! {
>     if err then ioerror(err, "in file.setopt(opt:c_int, arg)");
>   }
746c748
< } /* end of module */
\ No newline at end of file
---
> } /* end of module */
```

### Configuration Information

- Output of `chpl --version`:

```
chpl Version 1.16.0
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```

- Output of `$CHPL_HOME/util/printchplenv --anonymize`:

```
CHPL_TARGET_PLATFORM: darwin
CHPL_TARGET_COMPILER: clang
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
[...]

PR #7572, opened 2017-10-11
add error handling to Curl module

psahabu:
Fixes the issues raised in #7570 by marking `ioerror()` calls with `try` and making the calling methods throw.


bradcray:
I'm leaving this one to @mppf since he's more familiar with both error-handling and the curl module.

mppf:
This patch looks fine to me. @psahabu - the worrying thing is that this issue did not cause test failures. That means that these functions are not tested nightly. Why not? Do we have tests for them at all? Do they pass now? (If we have them, we need to run them nighty. If we don't have them, we need to add them).

bradcray:
I agree with Michael's point in the following respect: If a PR is fixing a previously unknown bug, it should also include a new test that would've failed before and passes now to lock in the fix.

W.r.t. the broader question of what kind of curl testing we already have, I don't recall.  But this wouldn't be the first instance of a library/feature that has insufficient testing... (not to excuse it, just to say that many of us are guilty of this).

psahabu:
@mppf it does not appear that we have Curl tests. I've added the repro in #7570 as a test.

Is it possible that there aren't Curl tests because they might be inconsistent from machine to machine?

mppf:
@psahabu - I suspect it was simply an oversight. PR #89 added Curl support and that PR message refers to an apparently uncommitted set of unit tests:   https://github.com/tzakian/chapel-unit-tests/tree/curl

Could you look at migrating those tests in to the Chapel repo? You'll need a SKIPIF to skip them if curl support is not enabled. (And we'll need to enable curl in some nightly testing config).

psahabu:
I'm taking a look at the tests now.

mppf:
@psahabu - perhaps we should merge this PR, but continue to look at testing Curl nightly as a separate effort?

test/modules/packages/Curl/setopt.chpl
```diff
@@ -0,0 +1,20 @@
+use Curl;
+
+proc main() {
+  const uniprotid = "P07204_TRBM_HUMAN";
+  const uniproturl = "http://www.uniprot.org/uniprot/" + uniprotid + ".fasta";
```
mppf: Seems to me that we should probably create testing traffic against our own website

test/modules/packages/Curl/setopt.good
```diff
@@ -0,0 +1 @@
+internal error: No Curl Support
```
mppf: Is this a compilation failure or a runtime failure? I'd expect the .good file to have the normal output and then the .skipif file to skip the test if curl is not available.

psahabu:
I'll split out the setopt tests and merge this PR, then open another PR when the rehabilitated Curl tests are ready.

psahabu:
Supplanted by #7713.

Issue #7575, opened 2017-10-11
Bounds checking is incorrect in array.insert method


daviditen:
### Summary of Problem
The `array.insert(pos: idxType, a: [])` method fails to catch out of bounds insertions in its bounds check.  A method it uses does catch it though, resulting in an "array slice out of bounds" message.

**Associated Future Test(s):**
[test/arrays/diten/arrayInsertPastEnd.chpl](https://github.com/chapel-lang/chapel/blob/master/test/arrays/diten/arrayInsertPastEnd.chpl)

### Configuration Information
- Output of `chpl --version`: chpl Version 1.17.0 pre-release (15496ff)

daviditen:
Fixed by #7577

PR #7577, opened 2017-10-11
Fix bounds checking in array.insert

daviditen:
Bounds checking in array.insert was allowing the insert position to be
anywhere in the newly resized array, but the position should be somewhere in
the non-resized array (or 1 past the end).

Fixes issue #7575.


daviditen:
Passed a full local linux64 paratest.

daviditen:
Thanks!

PR #7578, opened 2017-10-11
Switch from these(tag) to foralls for some standalone iters

benharsh:
Back when we were working on the rank-change array view, we ran into some bugs with standalone iterators and reductions:

https://github.com/chapel-lang/chapel/pull/5694#issuecomment-287206496

We went with the workaround of invoking "foo.these(tag)" directly, and the bug was tracked in #5223. This bug was recently fixed, so we should now be able to use the more elegant ``forall`` approach.

The primitives that verify a resolvable standalone iterator exists have not been changed. Until we know more about how we want to express this pattern, I'm hesitant to remove those primitives.

Testing:
- [x] full local
- [x] full gasnet
- [x] full numa

### DIFF
--- a/modules/dists/SparseBlockDist.chpl
+++ b/modules/dists/SparseBlockDist.chpl
@@ -245,7 +245,7 @@ class SparseBlockDom: BaseSparseDomImpl {
   {
     coforall locDom in locDoms {
       on locDom {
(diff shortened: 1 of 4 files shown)


benharsh:
@bradcray : no action necessary on your part, but this may be of interest if it's on your "ArrayViews cleanup" TODO list.

Issue #7569, opened 2017-10-10
multi-locale support for TIO (try it online)


bradcray:
As a background task for awhile now, I've been working on enabling multi-locale support for Chapel on the TIO website, hitting and addressing various stumbling blocks.  The current one is that TIO's sandbox doesn't support sockets.  I thought we could work around this by using `--enable-pshm` in GASNet, but it turns out that the udp conduit still uses sockets to bootstrap in that case.  Paul Hargrove tells me that the smp conduit may represent a path forward, but that doesn't work with Chapel out of the box (seems likely to require a new launcher and/or some new runtime code):

```
mbp-bradc (master)$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: darwin
CHPL_TARGET_COMPILER: clang
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: smp *
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: system *
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none
mbp-bradc (master)$ chpl test/release/examples/hello6-taskpar-dist.chpl 
mbp-bradc (master)$ ./hello6-taskpar-dist -nl 4
Hello, world! (from locale 0 of 4 named brads-workstation)
Hello, world! (from locale 0 of 4 named brads-workstation)
Hello, world! (from locale 0 of 4 named brads-workstation)
Hello, world! (from locale 0 of 4 named brads-workstation)
```

ronawho:
I'm pretty confident that smp doesn't require a launcher. From the [docs](https://github.com/chapel-lang/chapel/blob/master/third-party/gasnet/gasnet-src/smp-conduit/README#L17-L18): 
>smp-conduit jobs can be launched directly in the same manner as normal POSIX executables - ie on the command line, no spawner script required.


[...]

PR #7579, opened 2017-10-11
Stop throwing --enable-allow-gcc4 for the GASNet build

ronawho:
`--enable-allow-gcc4` allows "the use of a broken gcc/g++ 4.0-4.2 compiler". It
doesn't actually do anything other than quiet a warning/error that gasnet will
throw if it detects you're using a broken compiler.

gcc 4.2 is almost 10 years old at this point, so I think we can stop throwing
it and just prefer that people upgrade gcc versions or quiet the warning
themselves.

This was originally added in e7f7577d40 in 2008, when gcc 4 was new.

### DIFF
--- a/third-party/gasnet/Makefile
+++ b/third-party/gasnet/Makefile
@@ -6,7 +6,7 @@ export CHPL_COMM=gasnet
 CHPL_MAKE_HOST_TARGET = --target
 include $(CHPL_MAKE_HOME)/make/Makefile.base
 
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus and/or @bradcray any objections?

For context, I'm going to surround the `--enable-segment-$(CHPL_MAKE_COMM_SEGMENT)` code with a conditional for #7569. I didn't want to impact whether enable-allow-gcc4 was thrown, and looking at it I decided to stop throwing it instead of always adding it.

bradcray:
OK with me.

gbtitus:
I looked post-merge, and this seems like a wise change.

PR #7585, opened 2017-10-12
Add bug for forwarding a virtual method

benharsh:
I ran into this bug while working on array forwarding. The 'doiBulkTransfer'
method in BaseArr was causing bizarre resolution errors in the following test:

test/optimizations/sungeun/bulk_transfer/assign.chpl

Somehow the compiler was attempting to call doiBulkTransfer between a 1D
DefaultRectangular array of locales and a 3D DR array of reals.


benharsh:
@mppf : Do you agree that this is a bug? If so, are you aware of any duplicates?

benharsh:
Also, this isn't blocking the array forwarding branch. I can work around this issue by simply removing 'doiBulkTransfer' from BaseArr.

mppf:
Looks like a bug to me but I might learn something interesting by digging in to it...

mppf:
Also no, I don't know of any duplicates.

mppf:
@benharsh - this seems to have something to do with addAllToVirtualMaps. Could you make a GitHub issue for it?

benharsh:
See #7786.

PR #7590, opened 2017-10-12
"make check" should printchplenv w debug output

awallace-cray:
Export CHPL_CHECK_DEBUG=1 to see printchplenv immediately before test compile and execution.
This should help "make check" failures a little easier to diagnose.


awallace-cray:
No hurry. Thanks-

PR #7591, opened 2017-10-12
Check for gasnet segment when setting gasnet's --enable-segment-<segment>

ronawho:
I mistakenly checked the substrate in #7580.


gbtitus:
Crud.  Sorry about missing this in the original.  👍

PR #7597, opened 2017-10-13
These tests rely on bounds checking, so skip under --fast

noakesmichael:
These tests rely on bounds checking.  Add .skipif for --fast


bradcray:
Thanks Mike!  My subconscious must've been churning on this still because I woke up at 6:30am on Saturday morning and thought... I wonder if any of those tests won't come back clean.  Wait... my --fast tests will have the same problem that the other Thursday failure did.  If only my conscious could keep up (and clock a bit faster...).

Issue #7574, opened 2017-10-11
make install should install chpl man page


mppf:
make install doesn't currently install a man page for chpl

bradcray:
@mppf: @ronawho and I were noticing that 'make install' effectively does a 'make comprt' before doing the install, which seemed surprising (w.r.t. how I think make installs normally work).  Was this intentional (e.g., required for Debian) or an oversight?

Also, in looking at the git blame for the line and the associated commit message, I read "Don't install chpldoc yet." which suggests it may have been intentional for Debian's packaging requirements (maybe?)

awallace-cray:
2018-09-07 : pre-release testing : 
- configure (--prefix and --chpl-home) + make install, source = release tarball
  - installs chpl and mason but not chpldoc in bin
  - man not available
- configure (--prefix and --chpl-home) + make install, source = full Git repo
  - same- no chpldoc
  - man was available but not installed

awallace-cray:
Noticed that "make mason" depends on "make chpldoc" (ie, Makefile targets are connected). 
I wonder if mason (when installed by "make install" needs chpldoc to be make-installed as well. 
If not, then it might make sense to decouple the Makefile build targets.

ben-albrecht:
I am redacting the `[good first issue]` label here because this is infrastructure-related.

PR #7598, opened 2017-10-16
Updated "make install" with mason support

panzone:
This PR solves one of the points of issue #7574, adding support to make install for mason.

I've only made mason to be sure that I wasn't missing anything in how the script/make install works on chapel. After this is considered OK I will expand it (or makes new PRs) to cover the entire issue.

### DIFF
--- a/util/buildRelease/install.sh
+++ b/util/buildRelease/install.sh
@@ -317,6 +317,19 @@ done
 # copy utf8-decoder header
 myinstallfile third-party/utf8-decoder/utf8-decoder.h "$DEST_THIRD_PARTY"/utf8-decoder/
 
+# copy mason
+if [ -f tools/mason/mason ]
+then
+  if [ ! -z "$PREFIX" ]
+  then
+    myinstallfile tools/mason/mason "$PREFIX/share/chapel/$VERS/tools/mason"
+    ln -s "$PREFIX/share/chapel/$VERS/tools/mason/mason" "$PREFIX/bin/mason"
+  else
+    myinstallfile tools/mason/mason "$DEST_CHPL_HOME/tools/mason"
+    ln -s "$DEST_CHPL_HOME/tools/mason/mason" "$DEST_DIR/bin/$CHPL_HOST_PLATFORM"/mason
+  fi
+fi
+
 # copy chplconfig
 if [ -f chplconfig ]
 then
--- a/util/buildRelease/install.sh
+++ b/util/buildRelease/install.sh
@@ -322,8 +322,7 @@ if [ -f tools/mason/mason ]
 then
   if [ ! -z "$PREFIX" ]
   then
-    myinstallfile tools/mason/mason "$PREFIX/share/chapel/$VERS/tools/mason"
-    ln -s "$PREFIX/share/chapel/$VERS/tools/mason/mason" "$PREFIX/bin/mason"
+    myinstallfile tools/mason/mason "$PREFIX/bin"
   else
     myinstallfile tools/mason/mason "$DEST_CHPL_HOME/tools/mason"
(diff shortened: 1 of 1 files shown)


bradcray:
@panzone: Thanks for jumping onto this issue, which we greatly appreciate.  I was hoping someone (maybe @ben-albrecht or @mppf?) who was more familiar with the Debian packaging effort than I was might be able to flesh out what was intended by the comment I unearthed on the issue suggesting that it might be a reason not to have 'make install' work on these binaries.  That was the only thing that gave me pause in proceeding with the issue.

(Note that I haven't done anything to review your actual PR here, and will leave that to those who know 'make install' better...  same name as above, actually :) ).

Thanks again!

ben-albrecht:
My thinking is that this is fine, given the installation only takes place if `mason` is built, and `mason` wouldn't be built for the Debian package (for now, at least). 

I would appreciate @mppf double checking my thinking before merging though.

util/buildRelease/install.sh
```diff
@@ -317,6 +317,19 @@ done
 # copy utf8-decoder header
 myinstallfile third-party/utf8-decoder/utf8-decoder.h "$DEST_THIRD_PARTY"/utf8-decoder/
 
+# copy mason
+if [ -f tools/mason/mason ]
+then
+  if [ ! -z "$PREFIX" ]
+  then
+    myinstallfile tools/mason/mason "$PREFIX/share/chapel/$VERS/tools/mason"
+    ln -s "$PREFIX/share/chapel/$VERS/tools/mason/mason" "$PREFIX/bin/mason"
```
mppf: Is there some reason we can't actually but the main version in $PREFIX/bin/mason ? Is there something depending on the version in the tools directory? (If so, maybe it needs to be updated?).

I'm pretty sure that Debian Filesystem Hierarchy Standard rules would have us putting such binaries in to /usr/bin or maybe to /usr/lib. /usr/share is for processor-independent files.

util/buildRelease/install.sh
```diff
@@ -317,6 +317,19 @@ done
...
+  else
+    myinstallfile tools/mason/mason "$DEST_CHPL_HOME/tools/mason"
+    ln -s "$DEST_CHPL_HOME/tools/mason/mason" "$DEST_DIR/bin/$CHPL_HOST_PLATFORM"/mason
```
mppf: This version is fine as-is and in fact should mirror whatever we are used to doing in the source tree.

util/buildRelease/install.sh
panzone: I've done the symbolic link because the Mason makefile does it too. I'll explore if it actually needs to stay in the tool section.

bradcray: @panzone: As we've been writing up our 1.16 release notes, I only just learned that the $CHPL_HOME/bin/ copy of mason is a symbolic link, as you say here, and wondered why, but don't think I've heard the answer yet.  Can one of @ben-albrecht / @benharsh / @spartee explain the rationale?  Thanks.

benharsh: I think we used a symbolic link because that's what c2chapel did/does. I don't think there's a deep reason why it couldn't be changed.

bradcray: Intuitively, I'd think in either a $CHPL_HOME-based build or a make install-style setting, we'd want to copy those binaries such that the original source tree could be clobbered / removed without breaking the tool...?

mppf: The file is copied to an installation directory in the original version of this PR already. I was just bringing up that Debian won't like it if we put a binary in /usr/share.

panzone:
I've tried with mason in $PREFIX/bin and it seems to work ok on my machine.

mppf:
@panzone - Thanks!

@ben-albrecht - this looks good to me, but I'd appreciate it if you can check it works for you & manage merging it.

PR #7602, opened 2017-10-16
Silence GCC bug exposed by --baseline -O

benharsh:
This test began to fail after an unrelated PR (#7578) was merged. We
observed that GCC was generating an unaligned vector instruction in
deeply-nested inlined functions. Since chapel inlines many functions by
default under normal conditions (disabled by baseline), we think it's
unlikely that we would encounter this behavior in normal testing.

Throwing -O defaults to O3 under gcc, and this test will pass under baseline
with O2 enabled.

### DIFF
--- a/test/studies/sudoku/dinan/sudoku-simpler.compopts
+++ b/test/studies/sudoku/dinan/sudoku-simpler.compopts
@@ -1,0 +1,1 @@
+--ccflags -O2


bradcray:
I'm probably stating the obvious here, but note that this is going to throw -O2 for all back-end compilers, gcc or otherwise.  I'm not aware of a compiler that doesn't accept -O2 as a flag, but just wanted to point out that we're applying a big hammer here...

Also, there was something I'm not quite understanding about the diagnosis of the issue:  Is the thought that with --baseline, we're generating a lot of smaller, non-inlined functions that gcc is automatically inlining, resulting in the bug in question?  (I ask because I don't think anybody's ever accused Chapel of generating small, trivially inline-able functions in our generated code before).

benharsh:
I don't think we're generating "a lot" of smaller functions, but it's my best guess. But this test is also subject to tiny changes in the generated source, so it's hard to say what the cause might truly be. Running with -fno-inline-functions (or something similar) did squash the error, but I don't know how many other optimizations might have been affected by that flag.

benharsh:
For some more information: Two functions involved in this issue are '_build_tuple_always_allow_ref' and '_construct__tuple36', both of which are inlined by the Chapel compiler by default. Of course, many other transformations involving tuples are enabled unless --baseline is thrown.

PR #7603, opened 2017-10-16
Default to segment fast for gasnet's smp conduit

ronawho:
I had been mistakenly thinking you couldn't set a segment for gasnet, but it
turns out that you just can't use segment-everything with pshm. Gasnet defaults
to segment-fast, so that's what we were getting by not enabling a specific
segment.

Default to segment-fast to reflect this understanding. See Dan's explanation in
https://github.com/chapel-lang/chapel/commit/57726dc84c#commitcomment-24973648


ronawho:
@gbtitus quick sanity check?

gbtitus:
Looks good!   👍

PR #7604, opened 2017-10-16
Add smp to the list of compatible launchers for `make check`

ronawho:
Not all launchers are compatible with `make check` so we have to explicitly
list the supported ones. CHPL_LAUNCHER=smp works, so add it to the approved
list.


ronawho:
@gbtitus another quick sanity check?

gbtitus:
Yep, good to go.

PR #7605, opened 2017-10-16
Forward array method calls

benharsh:
This PR allows domain map authors to more easily implement custom methods for their array types. Prior to this PR, they would have to implement a secondary method on the ``_array`` type in order to forward the call to the actual method.

This PR applies ``forwarding`` to the ``_value`` field of the ``_array`` record, and to the downward-pointing fields in array views. As a result, we could eliminate many dsi* routines from the array view implementations and more elegantly implement StencilDist's ``noFluffView`` and ``updateFluff`` methods.

Testing:
- [x] linux64
- [x] gasnet
- [x] numa
- [x] valgrind
- [x] baseline

### DIFF
--- a/modules/internal/ArrayViewReindex.chpl
+++ b/modules/internal/ArrayViewReindex.chpl
@@ -99,7 +99,7 @@ module ArrayViewReindex {
   //
  class ArrayViewReindexDom: BaseRectangularDom {
     // the new reindexed index set that we represent upwards
-    var updom: DefaultRectangularDom(rank, idxType, stridable);
+    forwarding var updom: DefaultRectangularDom(rank, idxType, stridable);
 
     // the old original index set that we're equivalent to
     var downdomPid;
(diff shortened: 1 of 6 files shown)


benharsh:
@bradcray , @mppf : Finally got this working! Thanks Michael for all the help with forwarding issues/questions.

modules/internal/ArrayViewReindex.chpl
```diff
@@ -620,11 +548,18 @@ module ArrayViewReindex {
     //
     // bulk-transfer
     //
-
+    //
```
noakesmichael: nonexistant -> nonexistent

modules/internal/ArrayViewSlice.chpl
```diff
@@ -251,13 +228,16 @@ module ArrayViewSlice {
     //
     // bulk-transfer
     //
-
+    // If these methods were nonexistant, calls to these methods would use
```
noakesmichael: nonexistant -> nonexistent

PR #7607, opened 2017-10-18
Convert a passing range alignment future to a passing test

noakesmichael:
I happened to notice that Brad's recent PR resolved a future.

Convert the future to a passing test


bradcray:
Thanks Mike!  After merging, I thought "Oh, I should check futures," but then got distracted before acting.

PR #7609, opened 2017-10-18
Delete stmt

noakesmichael:
Do not invoke deinit() if a class variable is nil

Prior to this PR the user-defined method deinit() would be invoked for a class
variable with value 'nil'.  This PR fixes that error and also avoids calling chpl_here_free()
for these variables.

Two passing tests were updated to reflect the fix in behavior and a passing future was
converted to a passing test.

Two futures were impacted by this change.  This impact revealed that the futures had
been mischaracterized.  One future is for constructors and the other is for the converted
version with initializers and so will be a priority for this release.

All updates are in Chapel code.

paratested for single-locale and gasnet.

### DIFF
--- a/modules/internal/ChapelBase.chpl
+++ b/modules/internal/ChapelBase.chpl
@@ -1263,16 +1263,20 @@ module ChapelBase {
 
   // implements 'delete' statement
   inline proc chpl__delete(arg)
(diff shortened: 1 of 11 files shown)


bradcray:
Thanks for grabbing this, Mike!

PR #7627, opened 2017-10-20
Reduce wide pointer overhead for sparse domains/arrays

benharsh:
This PR expands upon the optimization implemented in #5487, which reduced wide pointer overhead for the _array record. Now this optimization will also apply to domains and distributions.

Now we can apply 'local field' pragmas to internal arrays used in LayoutCS and DefaultSparse.

Testing:
- [x] full local
- [x] full no-local
- [x] full gasnet


benharsh:
@ronawho : this fixes the no-local regression for cg-sparse.

PR #7632, opened 2017-10-23
The LCALS INNER_PROD kernel was using regular assignment instead of +=

daviditen:
This kernel should have been using += instead of just =.  This bug led to the
backend C compiler optimizing away the loop after PR #7422.

Verified that the reference version uses +=.

### DIFF
--- a/test/release/examples/benchmarks/lcals/RunCRawLoops.chpl
+++ b/test/release/examples/benchmarks/lcals/RunCRawLoops.chpl
@@ -63,7 +63,7 @@ module RunCRawLoops {
             for isamp in 0..#num_samples {
               q = 0.0;
               for k in 0..#len {
-                q = z[k]*x[k];
+                q += z[k] * x[k];
               }
               val = q*isamp;
             }


test/release/examples/benchmarks/lcals/RunCRawLoops.chpl
```diff
@@ -63,7 +63,7 @@ module RunCRawLoops {
             for isamp in 0..#num_samples {
               q = 0.0;
               for k in 0..#len {
-                q = z[k]*x[k];
+                q += z[k] * x[k];
```
ben-albrecht: I am not super familiar with the LCALS code. Is there some kind of check that validates the result that should have caught this? If so, is it just coincidental that this didn't impact the result?

daviditen:
It does a fake "checksum" to verify results, but this slipped by while still passing that.

test/release/examples/benchmarks/lcals/RunCRawLoops.chpl
ben-albrecht: > It does a fake "checksum" to verify results, but this slipped by while still passing that.

Gotcha. Looks good to me then.

daviditen:
Thanks!

PR #7638, opened 2017-10-24
Allow associative domain indices that contain ranges as fields/elements

### LINKED ISSUE #6110: Internal error for associative domain of tuple of ranges
### Summary of Problem
An associative domain with idxType tuple-of-range causes an internal compiler error.  It currently fails an assert in the scalarReplace pass.

### Steps to Reproduce

**Compile command:**
`chpl assocOfTupleOfTupleOfRange.chpl`

[test/domains/diten/assocOfTupleOfTupleOfRange.chpl](https://github.com/chapel-lang/chapel/blob/master/test/domains/diten/assocOfTupleOfRange.chpl)

### Configuration Information
`chpl Version 1.16.0 pre-release (9adb343)`

benharsh:
Implements ``chpl__defaultHash`` for ``range`` so that other valid indices (records, tuples, etc.) that contain ranges can still be hashed. 

Resolves #6110.

Testing:
- [x] full local


benharsh:
@daviditen : Could you review this? From what I can tell, the compiler would try to build a defaultHash for ranges, but would get tripped up by range promotion at some point, leading to invalid AST.

An alternative to resolving #6110 would be to disallow any record/tuple/etc. that contained a range.

daviditen:
Does `build_record_hash_function` in `buildDefaultFunctions.cpp` not cover this case for some reason?  It looks to me like it should.

benharsh:
``build_record_hash_function`` does not cover this case. I believe it gets hung up on range's promotion capability and generates incorrect code.

daviditen:
OK, thanks.  Looks good.

PR #7643, opened 2017-10-24
Clean up code by removing support for LLVM older than 4

dmk42:
Supporting old versions of LLVM caused considerable clutter (over 500 lines worth).  This change removes the old `#ifdef`s.  Where reasonable, it also further consolidates the code once the conditionals are removed.

This implements part of #6993 .

While testing this, I ran up against `ClangInfo` being declared as a struct in one place and a class in another.  To get it to compile with the latest Mac Clang, I had to make it consistently a struct.

The following tests were run with both LLVM 4 and LLVM 5:
Passed full local testing with `--llvm` on linux64.
Passed the hellos with `--llvm`, with and without `--fast`, plus test/llvm on a Mac building LLVM with clang.

### DIFF
--- a/compiler/include/llvmDebug.h
+++ b/compiler/include/llvmDebug.h
@@ -33,20 +33,14 @@
 #include "llvm/Support/Dwarf.h"
 #endif
 
-#if HAVE_LLVM_VER >= 35
 #include "llvm/IR/DebugInfo.h"
 #include "llvm/IR/DIBuilder.h"
-#else
-#include "llvm/DebugInfo.h"
-#include "llvm/DIBuilder.h"
-#endif
 
 #include <vector>
 
 //#define DW_LANG_chapel (llvm::dwarf::DW_LANG_lo_user+37)
 #define DW_LANG_chapel llvm::dwarf::DW_LANG_lo_user
 
-#if HAVE_LLVM_VER >= 37
 // This should be renamed to DI versions
 #define LLVM_DITYPE llvm::DIType*
 #define LLVM_DIFILE llvm::DIFile*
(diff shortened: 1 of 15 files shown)


compiler/include/llvmDebug.h
```diff
@@ -33,20 +33,14 @@
...
-#endif
 
 #include <vector>
 
 //#define DW_LANG_chapel (llvm::dwarf::DW_LANG_lo_user+37)
 #define DW_LANG_chapel llvm::dwarf::DW_LANG_lo_user
 
-#if HAVE_LLVM_VER >= 37
 // This should be renamed to DI versions
 #define LLVM_DITYPE llvm::DIType*
```
mppf: I recommend we remove these #defines and static inline functions and replace their uses manually. They are a source of clutter / confusion. The only reason they were there was to support different LLVM versions.

compiler/include/llvmDebug.h
```diff
@@ -85,62 +79,9 @@ static inline LLVM_DIVARIABLE toDIVARIABLE(llvm::MDNode* N)
...
-
-
-#endif
-
-
-
 // LLVM 4.0 added DIGlobalVariableExpression but previously
 // it was basically stored in DIGlobalVariable
-#if HAVE_LLVM_VER >= 40
 #define LLVM_DIGLOBALVARIABLEEXPRESSION llvm::DIGlobalVariableExpression*
```
mppf: Let's remove the macro and replace its uses with the RHS.

compiler/include/llvmUtil.h
```diff
@@ -46,39 +46,16 @@ static inline bool llvm_fn_param_has_attr(llvm::Function* f, unsigned idx, llvm:
...
 #error LLVM version is too old for this version of Chapel
 
 #endif
 
-// PassManager also changed names..
-#if HAVE_LLVM_VER >= 37
 #include "llvm/IR/LegacyPassManager.h"
 #define LEGACY_FUNCTION_PASS_MANAGER llvm::legacy::FunctionPassManager
 #define LEGACY_PASS_MANAGER llvm::legacy::PassManagerBase
 #define LEGACY_MODULE_PASS_MANAGER llvm::legacy::PassManager
```
mppf: Let's remove the macros

compiler/include/llvmUtil.h
```diff
@@ -103,7 +80,6 @@ uint64_t getTypeFieldNext(const llvm::DataLayout& layout, llvm::Type* ty, uint64
 
 
 // And create a type for a metadata operand 
-#if HAVE_LLVM_VER >= 36
 #define LLVM_METADATA_OPERAND_TYPE llvm::Metadata
```
mppf: this one too

compiler/include/llvmUtil.h
```diff
@@ -103,7 +80,6 @@ uint64_t getTypeFieldNext(const llvm::DataLayout& layout, llvm::Type* ty, uint64
...
 static inline llvm::ConstantAsMetadata* llvm_constant_as_metadata(llvm::Constant* C)
```
mppf: and this and the following function

compiler/include/llvmUtil.h
```diff
@@ -114,24 +90,9 @@ static inline
...
-
-#endif
 
 // LLVMs after 3.5 require C++11 support
-#if HAVE_LLVM_VER >= 35
 #define HAVE_LLVM_CXX11 1
-#endif
 
 #ifdef HAVE_LLVM_CXX11
 #define LLVM_CXX_OVERRIDE override
```
mppf: We can remove this macro as well, along with HAVE_LLVM_CXX11

compiler/util/llvmDebug.cpp
```diff
@@ -61,11 +61,7 @@ for more information on LLVM debug information.
 */
 
 // smooth changes in zero DIFlags
-#if HAVE_LLVM_VER >= 40
 #define FLAG_ZERO llvm::DINode::FlagZero
```
mppf: Let's remove this macro

dmk42:
Thanks @mppf !  I've made the suggested changes and retested.  Since I already have approval, I'll go ahead and merge so I can check the nightly tests tomorrow.

PR #7644, opened 2017-10-24
Fix the post-parse check for yields and returns from module scope

### LINKED ISSUE #7617: Compiler fails to emit error about yield outside of iterator at module scope
### Summary of Problem
The compiler will fail with an internal error for yields outside of an iterator at module scope. While this is certainly an invalid program, we should emit a better error message.

### Steps to Reproduce

**Source Code:**

```chapel
for i in 1..10 do yield i;
```
Output of `chpl foo.chpl`
```
internal error: CHE0564 chpl Version 1.17.0 pre-release (...)
...
```

**Compile command:**
`chpl foo.chpl`

  [`functions/iterators/bharshbarg/moduleYield.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/functions/iterators/bharshbarg/moduleYield.chpl
          )

### Configuration Information

All configurations, likely present since Mesozoic Era.

daviditen:
The checkModule function was walking the statements in the module's body
looking for yields and returns to issue errors about.  However, it didn't
dive into any statements, so skipped over whole BlockStmts without looking
inside of them.

Update it to look at all of the CallExprs in the module body that have the
module itself as a parentSymbol.

Remove a .future that yielded from the module level.  This will close #7617.

### DIFF
--- a/compiler/passes/checkParsed.cpp
+++ b/compiler/passes/checkParsed.cpp
@@ -352,11 +352,12 @@ static void nestedName(ModuleSymbol* mod) {
 
 static void
 checkModule(ModuleSymbol* mod) {
-  for_alist(stmt, mod->block->body) {
-    if (CallExpr* call = toCallExpr(stmt)) {
+  std::vector<CallExpr*> calls;
+  collectCallExprs(mod->block, calls);
+  for_vector(CallExpr, call, calls) {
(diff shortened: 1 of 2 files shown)


daviditen:
Passed a full linux64 paratest

daviditen:
Thank you!

PR #7647, opened 2017-10-25
Added chpl_internal_error to chpl_memalign()

panzone:
This PR adds checks to chpl_memalign for avoiding the bug presented in issue #7484

### DIFF
--- a/runtime/include/mem/cstdlib/chpl-mem-impl.h
+++ b/runtime/include/mem/cstdlib/chpl-mem-impl.h
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2017 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
  * The entirety of this work is licensed under the Apache License,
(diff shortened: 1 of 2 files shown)


runtime/include/mem/cstdlib/chpl-mem-impl.h
```diff
@@ -37,7 +37,10 @@ static inline void* chpl_malloc(size_t size) {
 }
 
 static inline void* chpl_memalign(size_t boundary, size_t size) {
-  return sys_memalign(boundary, size);
+  void *rp = sys_memalign(boundary, size);
+  if(rp==NULL)
```
ronawho: NULL is a valid valid if `size == 0`, so I think we need to check `if (rp == NULL && size != 0)`

Might also want to use `ret` instead of `rp` to more closely match the style in other runtime memory functions

ronawho:
@panzone I'm gonna close this to minimize the number of open PRs we have, but feel free to ping us or re-open if there's any updates

PR #7649, opened 2017-10-25
Avoid baseline error for chpl__defaultHash(range)

benharsh:
This PR changes the order of conditionals to work around baseline error for getField and enums. The original code could incorrectly attempt to acquire a const-reference to a param/type field, which could result in odd errors under baseline. Overall this change results in a more principled implementation.

Testing:
 - [x] full local


benharsh:
I was about to add a future for this issue, but realized that it should be illegal to take a const-reference of a param anyways.

PR #7652, opened 2017-10-25
Fix issues with using enums in generic functions

### LINKED ISSUE #5414: Scoping for enums
### Summary of Problem

Declaring and using enums inside of functions that take compile-time arguments such as params or types results in compilation error due to the enum type being unknown.

### Steps to Reproduce

**Source Code:**

`param` case:

```chapel
enumScopeTest(2);

proc enumScopeTest(param x) {
  enum Numbers {one, two, three};
  writeln(Numbers.one); // error: use of 'one' before encountering its definition, type unknown
}
```

`type` case:

```chapel
enumScopeTest(int);

proc enumScopeTest(type t) {
  enum Numbers {one, two, three};
  writeln(Numbers.one); // error: use of 'one' before encountering its definition, type unknown
}
```

working case:

```chapel
enumScopeTest(2);

proc enumScopeTest(x) {
  enum Numbers {one, two, three};
  writeln(Numbers.one); // prints '1'
}
```

[`test/types/enum/scope.chpl`](https://github.com/chapel-lang/chapel/tree/master/test/types/enum/scope.chpl) #6207

[`test/types/enum/scope-type.chpl`](https://github.com/chapel-lang/chapel/tree/master/test/types/enum/scope-type.chpl)

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.14.0.8dc84bd` (master)

lydia-duncan:
Resolves #5414.

Previously, we experienced an issue with enums declared within the scope of a
generic function - when the function was instantiated, we would fail to find the
enum constants.

As it turns out, our copy function on EnumType AST nodes was broken and we
hadn't noticed because we never defined EnumTypes within generic functions.  It
was broken in that it never set the enum constants to have the type of their
defining enum, and it never created a version of all the compiler-generated enum
default functions for the new copy of the EnumType.

This change makes the generation of the enum constant copies use the new type.
It also makes the generation of the enum default functions into something that
can be called outside of buildDefaultFunctions, and calls it when a copy of the
type is created (via copying the DefExpr defining it at the BlockStmt level, as these
functions rely on the EnumType being completed updated and inserted in the tree).
Finally, it tags the generated functions as generic when it creates them, since
creating them after the start of functionResolution means that they avoid the
tagIfGeneric iteration through all functions - this allows us to appropriately match
against them during resolution.

Allows test/types/enum/scope.chpl and test/types/enum/scope-type.chpl to pass,
as well as test/types/enum/scope-also-generic.chpl (which was written as part of
this work).  Also allowed the futures test/types/enum/diten/enumDeclInFn.chpl and
test/types/enum/sungeun/cast_enum_weird_4.chpl to pass.

Testing changes:
The .good for scope.chpl and scope-type.chpl was expecting the numerical value,
but it should have been expecting the string version of the constant.  Fix that.

Adds two new variations: scope-also-generic.chpl (for generic functions due
to having no type on an argument), and scope-works.chpl (which is the concrete
function case and worked prior to this change).  All four now pass.

Additionally, adds a version each of the test that performs assignment between
enum instances, iterates over the enum type, casts from an int to an enum, gets
the first enum for the type, and calls enum.size.  Also adds a test that multiple
instantiations of the function do not break

Passed a full std/ paratest


lydia-duncan:
TODO: 
- [ ] add tests of enum assignment, first, size, and enumerate in a generic function which declared the type, as those generated functions were impacted by this change.

PR #7657, opened 2017-10-26
Fix bug when writing a Stencil-distributed array

benharsh:
Prior to this change, a Stencil-distributed array would write its cached values when called with ``writeln`` or some other IO function. This is confusing and surprising behavior, so instead never read from the cache when printing a Stencil-distributed array. This allows such arrays to be printed before calling ``updateFluff``.


bradcray:
Cool!  Thanks for the quick fix!

bradcray:
Does the test need a .numlocales file?  Or maybe there's a directory-wide one?  In any case, I think this is good to merge when you're happy with it (no hurry on our part... I think we're winding down).

benharsh:
There's a directory-wide ``NUMLOCALES`` already.

PR #7659, opened 2017-10-26
Improve resolution of coercions & enums

### LINKED ISSUE #5922: Unfriendly behavior of assigning generally real values to real(32) variables

mppf:
This PR:
 * Enables param coercions from int/real/complex to real and complex when the value is perfectly representable
 * Enables param handling of real/imag/complex
    + (but in order for that to be user-facing, we'd need param operator overloads in ChapelBase - the subject of PR #7668).
 * Significantly improves enum behavior, so that enums can coerce to integral types in which the enum constants will fit.

What's the headline behavior improvement?
 * the literal 0.0 can coerce to real(32)
 * the literal 1 can coerce to real(32)

Resolves #5922.

### Summary of Behavior Changes

Here are some top-level features that continue to work as they did before:
* param i = 1:uint(32) + 1;  results in y.type being uint(32)
* integer literals such as 1 prefer to resolve to 'int', if other arguments did not indicate a stronger preference
* int, uint, enum, bool param/literals resolve to "default size" when crossing type category (e.g. bool to int(?w)) but otherwise prefer whichever overload is smallest (e.g. int(8) will prefer int(16) over int(64)).

Here are some top-level features that work after this PR (and not before):
* 0.0, 1, or other numeric values that are exactly representable can pass to a real(32) argument
     + for param reals, that means verifying both the mantissa and exponent will fit
* int(8), int(16) non-param values can pass to a real(32) argument (since they are exactly representable)
* param r = 1:real(32) + 0.0;  results in r.type being real(32) (after we add real param ops in PR #7688)
* enum X { a, b, c = a + b } works at all
* enum Y { a=1 }    var  x:myenum;   f(x); can now dispatch to f(any integral type), since the enum value fits into any integral type. This works even without resulting in ambiguity if overloads for all integer types are present!
* real param/literals resolve to "default size" when crossing type but otherwise prefer whichever overload is smallest
* 15 futures retired!

### Enum Examples that Behave Differently

``` chapel
      enum myenum { v }

      proc foo(x:uint(8)) {
        writeln("uint(8)");
      }

      proc foo(x:uint) {
        writeln("uint");
      }

      var x:myenum;

      foo(x);
```

After this PR, this resolves to the 'uint' foo; on master it fails to resolve, with ambiguity. I think it makes sense to prefer the "default sized type" in this case (even before this PR, we would prefer an 'int' version), but more importantly adding this rule helps to prevent surprising ambiguity errors that arise from complex relationships between the various most-specific-argument rules.

``` chapel
      enum myenum { X=1, Y=1000 }

      proc foo(x:int(8)) {
        writeln("int(8) version");
      }

      proc foo(x:int(32)) {
        writeln("int(32) version");
      }

      foo(myenum.X);
```

After this PR, this resolves to "int(32) version" rather than "int(8) version". In fact it could resolve to the int(8) version if it were the only option, but the int(32) version is preferred because the enum type, myenum, which is the type of myenum.X, can always dispatch to int(32), but myenum.X can only dispatch to the int(8) version because it happens to be a param with the value 1.

### Testing and Review

Passes full local futures testing.

### Future Work

 * consider also enabling coercions from int(32) to real(32) (discuss in issue #7740)
 * actually adding 'real' param functions (PR #7688 is a start)
 * spec changes to describe actual behavior of "param prefers" in the resolution section (note though that some of the features, such as passing an int(8) variable to a real(32) argument, are already described as something that should work in the spec. Other areas, such as how function disambiguation works with param arguments, are incorrectly described in the spec.).


lydia-duncan:
Heads up, it looks like this gets very close to, but doesn't seem to touch the same code that I fixed in #7676.  Are you still working on this change, or should I try to see the impact of merging it with mine?

lydia-duncan:
Note that the .bad file mismatch is the same test that I fixed in my PR

mppf:
@lydia-duncan - thanks, I was just working on that. I'll leave it to your PR.

test/types/enum/ferguson/enum-overflow.chpl
```diff
@@ -0,0 +1,17 @@
+enum myenum {
```
vasslitvinov: Could you please add a top-level comment to this file explaining whether this is a legal program i.e. it has a well-defined behavior, and why i.e. what language rules talk about "overflow" going from a to b to c.

test/types/range/hilde/alignTypeError1.chpl
```diff
@@ -1,30 +1,50 @@
...
+var i32 = 3:int(32);
+var i64 = 3:int(64);
+var i   = 3:int;
+var u8  = 3:uint(8);
+var u16 = 3:uint(16);
+var u32 = 3:uint(32);
+var u64 = 3:uint(64);
+var u   = 3:uint;
+ 
+
```
vasslitvinov: Looks like the line numbers changed in .good file. Why have these empty lines here?

I am ok keeping them if you have a good reason. It would be nice to explain
in a comment, to save other people from questioning.

test/types/range/hilde/alignTypeError2.chpl
```diff
@@ -1,13 +1,33 @@
...
+var i32 = 3:int(32);
+var i64 = 3:int(64);
+var i   = 3:int;
+var u8  = 3:uint(8);
+var u16 = 3:uint(16);
+var u32 = 3:uint(32);
+var u64 = 3:uint(64);
+var u   = 3:uint;
+ 
+
```
vasslitvinov: ditto re empty lines
and the same for the other two tests below

compiler/resolution/functionResolution.cpp
```diff
@@ -71,18 +72,20 @@ class DisambiguationState {
...
-  // 1 == fn1, 2 == fn2, -1 == conflicting signals
-  int   paramPrefers;
+  bool fn1WeakPreferred;
+  bool fn2WeakPreferred;
+
+  bool fn1WeakerPreferred;
+  bool fn2WeakerPreferred;
+
+  bool fn1WeakestPreferred;
+  bool fn2WeakestPreferred;
```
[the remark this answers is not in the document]
mppf: I was expecting to explain that in the specification. But at a high level, it's because something like 'param x:int(16) = 5' should be able to coerce to 'int' or 'uint' or 'int(8)' or .... Meanwhile 'param y = 5' should be able to coerce to 'int(16)' or 'uint' or 'int(8)' or ....  Now imagine we are resolving 'x+y'. We want it to resolve to the 'int(16)' version because 'x' has a type specified, but 'y' is a default type. It used to be the case that this version was chosen simply because non-default-sized ints didn't allow param conversion.

PR #7660, opened 2017-10-26
Add bug for printing slice of Stencil-distributed array

benharsh:


benharsh:
@bradcray : it occurred to me that printing a slice of a Stencil-distributed array could be problematic, and indeed it is. The output is the same as before #7657 . In my opinion, this is more of a general array-view issue.

Issue #5343, opened 2017-02-10
promoted array assignment with downcasting of complex(128)


ben-albrecht:
### Summary of Problem

The following program results in a compile-time error during function resolution as a result of the promoted assignment of A to a `complex(128)` literal. The error message (with `CHPL_DEVELOPER=true`) is:

```
$CHPL_HOME/modules/internal/ChapelIteratorSupport.chpl:95: error: cannot iterate over values of type complex(128)
$CHPL_HOME/modules/internal/ChapelIteratorSupport.chpl:95: note: unresolved call had id 1089401
```

with `--no-devel`, the error is:

```
complex_downcast.chpl:3: error: cannot iterate over values of type complex(128)
```

Ideally, I'd expect this program to succeed, with the literal assignment `A = i +0i` being downcasted from `complex(128)` to `complex(64)`. A potential short-term solution might be to give a more meaningful error message, such as a type error.

### Steps to Reproduce

**Source Code:**

```chapel
var A: [0..3] complex(64);
// Note: If rewritten (more correctly) as `A = (1 + 0i): t;`, error does not occur
A = 1 + 0i;
```

**Associated Future Test(s):**
`test/arrays/promotion/complex_downcast.chpl`  #5344

### Configuration Information

- Output of `chpl --version`: chpl Version 1.14.0.f3cacb7 (master branch at the time of posting)
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: darwin
CHPL_TARGET_COMPILER: clang
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: none *
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2 *
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: Apple LLVM version 8.0.0 (clang-800.0.42.1)

bradcray:
At present, there are a few challenges to making this work as you'd like:

[...]

PR #7668, opened 2017-10-27
Enable param real operations

mppf:
This PR adds support for some param operations on real/imag numbers.

- [ ] are we OK with having param real operations?
- [x] allow param complex addition to resolve issue #5343
- [ ] full local testing


mppf:
This PR is replaced by PR #12386.

PR #7672, opened 2017-10-27
Enable accessing type/param fields from a type

### LINKED ISSUE #5937: Querying compile-time properties of types
Should the following be valid Chapel?

```chpl

class X {
  type eltee;
  param strided;
  param rank;
}

proc printer(type T) {
  writeln(T.eltee:string);
  writeln(T.strided);
  writeln(T.rank);
}

proc main() {
  var A : X(int, false, 1);
  var B : X(int, true, 3);

  printer(A.type);
  printer(B.type);
}
```

There are places in our internal modules where we create temporaries from a type argument to get information that we could have known at compile-time. For example, we have an array helper routine that increments domain ref counts for inner arrays. It looks a bit like this:
```chpl
proc incRefHelper(parentArr, type eltType) {
  if isArrayType(eltType) {
    var temp : eltType; // potentially expensive!
    temp.domain._ref_count_helper();
    incRefHelper(parentArr, temp.eltType);
  }
}
```

It would be nice if we could just say ```eltType.eltType``` at the end, instead of relying on a temporary. I'd also argue that in the case of arrays we should be able to query the domain from the type, but that's probably best for a separate discussion.

**Edit**: updated code to cast type to string

### LINKED ISSUE #7614: size method does not resolve for type tuples
### Summary of Problem
The Chapel compiler cannot resolve the call to `type <tuple of types>.size`.

### Steps to Reproduce

**Source Code:**

```chapel
proc test(type eltTypes) {
  writeln(eltTypes.size);
}
test((int, string, real));
```

**Compile command:**
`chpl test.chpl`

**Execution command:**
`./test`

**Actual output:**
```
test.chpl:1: In function 'test':
test.chpl:2: error: unresolved call 'type (int(64),string,real(64)).size'
[...]
```

**Expected output:**
`3`

### Configuration Information

- Output of `chpl --version`: 1.16.0 

### Miscellaneous

I know that an alternative form of this example could work:
```chapel
proc test(type eltTypes...) {
  writeln(eltTypes.size);
}
test(int, string, real);
```

However, this pattern does not fit my actual use-case.

mppf:
This PR adjusts the compiler-generated accessor generation in buildDefaultFunctions to generate a type method for type/param fields. That allows the field to be accessed from a type, rather than just from an object.

It seems odd to me that we generate param field accesses with PRIM_MEMBER (and not PRIM_MEMBER_VALUE) but I wasn't able to resolve that quickly.

Besides adjusting buildDefaultFunctions, this PR makes a minor change to postFoldPrimop to correctly handle retrieving a param from a type variable.

Closes #5937.
Fixes #7614.

- [x] full local testing
- [x] test/release/examples/primers with
- [x] --baseline
- [x] --verify
- [x] --baseline --verify


mppf:
Testing:
[Error matching program output for classes/deitz/types/type_in_class3]
[Error matching .bad file for classes/ferguson/generic-type-method-simple1]
[Error matching .bad file for classes/ferguson/generic-type-method-simple2]
[Error matching .bad file for classes/ferguson/generic-type-method-simple3]
[Error matching .bad file for classes/ferguson/generic-type-method-tuple]
Future (feature request: Syntactic sugar for PRIM_QUERY_TYPE and PRIM_QUERY_PARAM) [Success matching program output for classes/lydia/prettierQueryField]
Future (bug: printing slice of StencilDist prints cached values) [Success matching program output for distributions/bharshbarg/stencil/writeSlice]

mppf:
@lydia-duncan - do you know if PRIM_QUERY_TYPE_FIELD and PRIM_QUERY_PARAM_FIELD will still be needed after this PR?

lydia-duncan:
It looks like PRIM_QUERY_TYPE_FIELD and PRIM_QUERY_PARAM_FIELD are still used in postFold and buildDefaultFunctions.  Based on the future text, I suspect you have implemented a similar functionality and so it could be unified, but I would have to check to be certain

lydia-duncan:
Do you want me to review this PR?

mppf:
@lydia-duncan - yes, I'd very much appreciate a review of this PR. Thanks!

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -311,15 +308,18 @@ static void build_accessor(AggregateType* ct, Symbol* field, bool setter) {
 
   fn->addFlag(FLAG_FIELD_ACCESSOR);
 
-  if (fieldIsConst)
+  if (fieldIsConst && !typeMethod)
     fn->addFlag(FLAG_REF_TO_CONST);
-  else if (recordLike)
+  else if (recordLike && !typeMethod)
```
lydia-duncan: I would have the typeMethod check be unified outside rather than added to both, i.e.:

```
if (!typeMethod) {
  if (fieldIsConst)
    fn->addFlag(FLAG_REF_TO_CONST);
  else if (recordLike)
  // etc.
```

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -418,8 +426,21 @@ static void build_accessors(AggregateType* ct, Symbol *field) {
     return;
 
   // Otherwise, build compiler-default getter and setter.
-  build_accessor(ct, field, true);
-  build_accessor(ct, field, false);
+
+  if (isUnion(ct)) {
+    // Unions need a special getter and setter.
+    build_accessor(ct, field, /* setter */ false, false);
+    build_accessor(ct, field, /* setter */ true,  false);
```
lydia-duncan: Shouldn't one of these comments say "getter" instead?  If it is intended to label the boolean argument, why is there not also one for the second boolean, and an "="?

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -418,8 +426,21 @@ static void build_accessors(AggregateType* ct, Symbol *field) {
...
+  } else {
+    // Otherwise, only build one version for records and classes.
+    // This is normally the 'ref' version.
+    build_accessor(ct, field, /* setter */ !fieldIsConst, false);
+  }
+
+  // If the field is type/param, add a type-method accessor.
+  if (fieldTypeOrParam) {
+    build_accessor(ct, field, /* setter */ false, true);
```
lydia-duncan: The above applies here as well

bradcray:
This seems likely to want some sort of specification update...?

mppf:
@bradcray - I think it's debatable whether or not it's "obvious" (i.e. that people reading the spec assume you can do this) but I created PR #7735 to update the spec to affirm that we can do this.

Issue #5690, opened 2017-03-16
Include where-clauses on unresolved call candidate signatures


ben-albrecht:
The following code:
```chapel
proc foo(x) where isIntegral(x) { }
proc foo(x) where isReal(x) { }
foo("bar");
```

Results in the following unresolved call error upon compilation:

```
foo.chpl:4: error: unresolved call 'foo("bar")'
foo.chpl:1: note: candidates are: foo(x)
foo.chpl:2: note:                 foo(x)
```

The candidate signatures do not contain sufficient information for the user to deduce what the problem is. It would be helpful to include the where clauses in these candidate signatures for this error.

PR #7675, opened 2017-10-27
Print where clauses

daviditen:
Update toString(FnSymbol*) to include the where clause if it exists

This will cause the where clause to be included in messages such as the
candidate list in an unresolved call error.

Updated a handful of test .good and .bad files to include where clauses.

This is requested in issue #5690.

### DIFF
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -2213,7 +2213,7 @@ const char* toString(FnSymbol* fn) {
     }
   }
 
-  if (fn->where) {
+  if (fn->where && fn->where->body.length == 1) {
     AstToText info;
     info.appendExpr(fn->where->body.only(), false);
     retval = astr(retval, " where ", info.text().c_str());
--- a/compiler/AST/symbol.cpp
+++ b/compiler/AST/symbol.cpp
@@ -2085,12 +2085,7 @@ const char* toString(FnSymbol* fn) {
   const char* retval = NULL;
 
   if (fn->userString != NULL) {
-    if (developer == true) {
-      retval = astr(fn->userString, " [", istr(fn->id), "]");
-    } else {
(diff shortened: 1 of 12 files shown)


daviditen:

compiler/AST/symbol.cpp
```diff
@@ -2216,10 +2211,16 @@ const char* toString(FnSymbol* fn) {
     if (skipParens == false) {
       retval = astr(retval, ")");
     }
+  }
 
-    if (developer  == true) {
-      retval = astr(retval, " [", istr(fn->id), "]");
-    }
+  if (fn->where && fn->where->body.length == 1) {
```
ben-albrecht: Could you clarify the purpose of the `fn->where->body.length == 1` condition to me?

ben-albrecht:
As a sanity check, have you tested any `where` clauses with more than 1 concatenated conditional? I didn't see any tests like that.

compiler/AST/symbol.cpp
daviditen: In the second commit message I mentioned it, but misspelled the word "shouldn't" as "should".

The compiler sometimes inserts multi-statement where clauses and we shouldn't print them.  One such case is for functions that have a queried argument.

daviditen:
Ah, shoot.  You're right about the multiple conditions case.

```
proc foo(x) where isIntegral(x) && !isReal(x) { }
proc foo(x) where isReal(x) { }
foo("bar");
```

Doesn't print the where clause for the first version.

daviditen:
I think I have a way make it work by caching the where clause earlier in the compiler before it gets normalized.  I'm going to close this PR for the moment though.  Thanks for looking at it.

ben-albrecht:
I knew it was too good to be true!

Issue #7011, opened 2017-08-15
CHPL_COMM_DEBUG=1 w/ GASNet requires either CHPL_DEVELOPER=1 or OPTIMIZE=0


PHHargrove:
I am trying to deploy some CI scripts to test Chapel's "master" with GASNet's "develop" on a daily basis. While I am having reasonably good luck with default builds, I am not having much luck with getting CHPL_COMM_DEBUG to work as I am expecting.

This is the second (less pressing for me) issue I've encountered.

I am building Chapel with the 1-line work-around I describe in #7010, and passing `CHPL_COMM=gasnet CHPL_COMM_SUBSTRATE=udp CHPL_DEBUG=1` to gmake.
By default I am seeing the following while compiling comm-chapel.c:
```
/tmp/BLD/dbg/chapel/third-party/gasnet/install/linux64-gnu-unknown/seg-everything/debug/include/udp-conduit/amudp_const.h:64:6: error: #error Tried to compile AMUDP client code with optimization enabled but also AMUDP_DEBUG (which seriously hurts performance). Disable C and C++ compiler optimization or reconfigure/rebuild without --enable-debug
     #error Tried to compile AMUDP client code with optimization enabled but also AMUDP_DEBUG (which seriously hurts performance). Disable C and C++ compiler optimization or reconfigure/rebuild without --enable-debug
      ^
```

I find that I can avoid this problem by adding `CHPL_DEVELOPER=1` to the gmake command line.  
Personally, I am OK with that one added setting for use within our CI scripting.  
However, my reading of the Chapel documentation didn't sound as if `CHPL_DEVELOPER=1` is an *intended* prerequisite for `CHPL_COMM_DEBUG=1` (but I could have missed it).

Also, since I am uncertain what else might be happening as a side-effect of setting `CHPL_DEVELOPER=1`, I am wondering: is there some other approach I could/should be using to avoid this error?  I see `OPTIMIZE` in the Makefiles, but that doesn't appear to be "user-facing" since it lacks a `CHPL_` prefix.

bradcray:
On your command line, `CHPL_DEBUG` is the setting that stands out as being unfamiliar to me.  Does it work if you don't set that?  And what caused you to set it?

You may have better luck setting `DEBUG=1` which is what we do more often (and it looks like it results in `CHPL_DEBUG` being set as well, so may result in better compatibility.

[...]

PR #1099, opened 2015-01-08
Set NDEBUG for optimized runtime builds

Kyle-B:
Setting NDEBUG will turn any asserts in our runtime code into no-ops.

This was being done in an ad-hoc basis in a few files, those checks have
been removed to avoid duplicate definitions.

PR #7678, opened 2017-10-27
Fix CHPL_COMM_DEBUG=1 OPTIMIZE=1 builds

ronawho:
CHPL_COMM_DEBUG builds gasnet with --enable-debug and GASNet has some checks to
ensure that --enable-debug isn't combined with -O3 or -DNDEBUG. We have force
thrown -O0 for CHPL_COMM_DEBUG for a while, but we've been mistakenly throwing
-DNDEBUG since #1099.

This just throws -UNDEBUG if CHPL_COMM_DEBUG is set. Note that if somebody is
building with CHPL_COMM_DEBUG, they might want to build with OPTIMIZE=0 so the
rest of the runtime is built with debugging enabled, but there's no reason you
shouldn't be able to just enable debugging for gasnet.

This resolves https://github.com/chapel-lang/chapel/issues/7011


ronawho:
Builds with `CHPL_COMM=gasnet OPTIMIZE=1 CHPL_COMM_DEBUG=1` and passes `make check` for gnu, clang, intel, and cce

Issue #7061, opened 2017-08-21
Docs unclear with respect to support status of ofi and psm conduits


PHHargrove:
In the [docs for CHAPEL_COMM_SUBSTRATE](http://chapel.cray.com/docs/1.15/usingchapel/multilocale.html#set-comm-conduit) I find ofi and psm conduits listed (among others).  
However, they do not appear in the [table of default CHPL_LAUNCHER values](http://chapel.cray.com/docs/1.15/usingchapel/launcher.html#currently-supported-launchers).  
So, I am left uncertain if Chapel intends to  
  1. support these two conduits, but w/o a default launcher.
  1. not support these two conduits, but has listed them in the substrate list for completeness.

**If (and only)** if you wish to consider psm-conduit as supported, then I noticed a `runtime/src/launch/gasnetrun_psm` which appears unused in master.  I think one just needs a couple more lines in util/chplenv/chpl_launcher.py` to make it the default for psm-conduit.

**If (and only)** if you wish to consider ofi-conduit as supported, then you need pretty much the same logic as for psm-conduit, which in turn is more or less identical to ibv-conduit.  I *have* actually tested this, FWIW.

Let me know if you need help testing.

Related:   
I am at least 90% sure that `gasnetrun_mpi` is a very good default for pami-conduit, and that it will work for many (most?) configurations using portals4 or shmem.  However, I have not tested those 3 conduits (yet?).

ben-albrecht:
Hey @PHHargrove  -- thanks for reporting this. I think @gbtitus, @ronawho, (or maybe @bradcray) might be the right people to respond to this. They are currently unavailable, but we'll get you a response when they get back.

ronawho:
`runtime/src/launch/gasnetrun_psm` was added by an external user with: https://github.com/chapel-lang/chapel/pull/5843. The docs you linked to are from 1.15, gasnetrun_psm is actually documented in the master (future 1.16 docs): http://chapel.cray.com/docs/master/usingchapel/launcher.html#currently-supported-launchers

That said, you're right that we don't currently default to gasnetrun_psm in `util/chplenv/chpl_launcher.py`.  This is because we don't have a machine to test on, so we weren't confident enough to make it the default launcher.

For ofi -- are you saying we could create a gasnetrun_ofi launcher, that's similar to the psm/ibv launchers?

PHHargrove:
[...]

PR #7679, opened 2017-10-28
Add an OFI launcher

ronawho:
Add a gasnetrun_ofi launcher that's a code clone of gasnetrun_{psm, mpi, ibv}
launchers.

We default to CHPL_LAUNCHER=gasnetrun_ofi for CHPL_COMM_SUBSTRATE=ofi. While I
was here I also made CHPL_LAUNCHER=gasnetrun_psm for CHPL_COMM_SUBSTRATE=psm

Tested the ofi launcher locally on my mac with the ofi sockets profider. Tested
psm on an omnipath system. We need regular testing for both of these configs,
but they work well enough to make them defaults.

This resolves https://github.com/chapel-lang/chapel/issues/7061


ronawho:
@gbtitus mind giving this a sanity check post-commit?

PR #7682, opened 2017-10-30
Improve tuple-out-of-bounds error to include bounds

mppf:
The current tuple-out-of-bounds error looks like this:
```
t.chpl:3: error: tuple index out-of-bounds error (0)
```

I think this error message misses an opportunity to educate Chapel users about tuples. This PR changes it to indicate that tuples start at 1 (if element 0 is accessed) or otherwise print the range of indices that the tuple contains. Please see the .good files for new tests to see exactly what the error message changes to.


mppf:
@lydia-duncan - would you mind reviewing this PR improving tuple out of bounds error messages? Thanks.

Issue #7673, opened 2017-10-27
Remove uses of _value from tests and standard modules


benharsh:
The array, domain, and distribution records all have a field named ``_value``. This is a potentially-privatized class representing the array/domain/dist implementation. In the past users have been forced to use this field to access the underlying implementation in a non-standard way that could change out from under them in the next release.

This issue will be used to track remaining uses of ``_value`` in our tests and standard modules. Further issues may be spawned if we are unable to replace the use of ``_value`` with something better. I would recommend that we eventually make ``_value`` a private method, but then I'm not sure how we would write tests to verify/inspect internal state. It is also unclear to me how we should expose things like bulk-transfer and RADopt for use in distributions.

Below I have categorized tests and standard modules based on how they use ``_value``. Some tests may appear in multiple categories.

# Iterators
## Leader/Follower forwarding
- [x] [`modules/standard/DynamicIters.chpl`](https://github.com/chapel-lang/chapel/blob/master/modules/standard/DynamicIters.chpl)
- [x] [`modules/standard/Random.chpl`](https://github.com/chapel-lang/chapel/blob/master/modules/standard/Random.chpl)
- [x] [`test/distributions/bradc/assoc/UserMapAssoc.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/distributions/bradc/assoc/UserMapAssoc.chpl)
- [x] [`test/functions/iterators/sungeun/forwarding/forwardDefaultArr.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/iterators/sungeun/forwarding/forwardDefaultArr.chpl)
- [x] [`test/functions/iterators/sungeun/forwarding/forwardDefaultDom.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/iterators/sungeun/forwarding/forwardDefaultDom.chpl)
- [x] [`test/functions/iterators/sungeun/localDataPar.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/iterators/sungeun/localDataPar.chpl)
- [x] [`test/reductions/bharshbarg/reduceWithoutTag.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/reductions/bharshbarg/reduceWithoutTag.chpl)
[...]

PR #7683, opened 2017-10-30
Remove uses of _value in various tests

benharsh:
This PR addresses some of the tests listed in #7673. The changes in this PR fall into the following categories:
- use targetLocales/localSubdomain
- use top-level ``type``, ``eltType``, or ``idxType`` queries
- use ``domain.dim`` instead of ``_value.dsiDim``
- remove some unused code

Testing:
- [x] full local
- [x] full gasnet


benharsh:
@vasslitvinov : This PR removes two futures associated with your name:
- test/classes/vass/generic-method-with-param-arg-2.chpl
- test/classes/vass/generic-method-with-param-arg-3.chpl

Any objections?

test/arrays/vass/initialization-mapped-1.compopts
```diff
@@ -0,0 +1 @@
+--devel
```
vasslitvinov: What is the motivation for adding --devel ?

benharsh: So that we print ``Block`` or ``DefaultDist`` or whatever instead of the cleaner ``domain(1, int ,false)``.

PR #7686, opened 2017-10-30
Retire SPMD variation of miniMD

benharsh:
This version of miniMD is inelegant, doesn't receive much attention, and performs worse compared to the elegant version. There is little value in keeping it around, so this PR removes it entirely. Relevant docs and Makefiles have been updated to reflect this change.

Testing:
- [x] release/examples
- [x] 'make' in miniMD directory


benharsh:
@ronawho : do we need to do anything for performance .dat files if a column goes away?

PR #7695, opened 2017-10-31
Retire outdated distribution futures

benharsh:
Four futures have been removed or retired in this PR:

- test/distributions/bradc/block1Dlocale.chpl
  At some point started failing to compile, but we didn't notice due to lack of .bad file. The correct behavior has likely been in place for over a year.

- test/distributions/bradc/clonedForall.chpl
  This test was created out of concern for compiler efficiency when someone noticed we duplicate loop bodies for some internal iterators. Manually preventing the duplication did not improve compilation time for LCALS, which uses loops extensively.

- test/distributions/robust/arithmetic/reindexing/test_strided_reindex3.chpl
  This test was incorrectly updated during the removal of the ``=>`` operator, which resulted in a runtime failure because the reindex range did not match the slice's length. Furthermore, the future was originally created when it was unknown whether reindexes should preserve the distribution. We have since decided that reindexes *should* preserve the distribution, so this test is updated to reflect that behavior and the .future file removed.

- test/distributions/sungeun/misc/useDummyDist.chpl
  This test was created to track a resolution error that has since gone away, and in the meantime became out of sync with the expected distribution interface and misleadingly still failed to compile. This test does not hold value anymore and is removed.


benharsh:
@bradcray : does this look OK to you? It seems as though this test was left behind at some point and failed to compile. It's entirely possible that this was before .bad files.

benharsh:
I've found some other futures under distributions/ that also need some updating, so I'll close this while I work.

Issue #7665, opened 2017-10-26
Figure out a better way to prevent autoconf reruns for third-party packages


ronawho:
Because we store our third-parties untarred, we have to touch some files in a
certain order to prevent a rebuild of the configure script. Tarball releases of
third-party codes don't have this issue, because the tarball maintains
timestamps, but those timestamps are lost when we untar and commit to git. To
combat this we try to touch some configure/autoconf files in a specific order.
Unfortunately, we can never seem to get this quite right in a portable manner.
Our original `touch` code comes from GASNet's Bootstrap file, which seems to
work well, so maybe we should just follow that more closely.

Here's just a few instances where we've had to adjust:

Hwloc:
 - https://github.com/chapel-lang/chapel/commit/277db01a8e
 - https://github.com/chapel-lang/chapel/commit/19a6dbb6af
 - https://github.com/chapel-lang/chapel/commit/10fd4c9700

Qthreads:
 - https://github.com/chapel-lang/chapel/commit/c5b7aea6a7
 - https://github.com/chapel-lang/chapel/commit/07812e2e37
 - https://github.com/chapel-lang/chapel/commit/2071a3b2da

GMP:
 - https://github.com/chapel-lang/chapel/commit/bcdc8b5201

We're not touching any configure/autoconf files for gmp or jemalloc, should we
be?

Are we going to run into issues when we store llvm untarred? I think llvm will
use cmake instead of configure/autoconf, but there could be similar issues

As far as trying to address this:

We could consider storing the tarballs again, but this has several drawbacks:
 - For packages that we apply patches/fixes to, we'd have to untar and retar
 - We transitioned to not store the tarballs so we wouldn't blow up our git
   history as much
 - Searching third-party sources would be harder

For our tarball releases, using third-party tarballs would be fine, but it
seems like a bad idea for our git repo.  Maybe we could do something with
submodules or some other git feature, but this really isn't an area I'm
familiar with.

ronawho:
Pinging @gbtitus and @dmk42 

I don't think we need to do anything with this just yet, but I think it's useful to track how many times this bites us

ronawho:
It's been a while since we've had any issues with this, closing.

PR #7014, opened 2017-08-15
Change GMP source directory structure

ronawho:
Untar gmp-6.0.0a into gmp-src so we can avoid adding and removing new/old
source versions when we upgrade.

This is similar to #3424 (jemalloc), #3453 (hwloc), #3914 (gasnet), and #4498
(qthreads)

PR #7696, opened 2017-10-31
Add note about GMP's requirement of m4 to prereq.rst

lydia-duncan:
This recently tripped up a user, so call it out explicitly.  The
GMP installation information notes the package dependency and some
situations where the version in use on a particular system may be
incompatible.

### DIFF
--- a/doc/rst/usingchapel/prereqs.rst
+++ b/doc/rst/usingchapel/prereqs.rst
@@ -35,6 +35,9 @@ about your environment for using Chapel:
 
   * Building LLVM requires cmake version 3.4.3 or later.
 
+  * Building GMP requires the m4 macro processing package, which may
+    not be installed by default in all systems.
+
   * If you wish to use Chapel's test system, python-setuptools and
     python-devel (or equivalent packages for your platform) are required.
 


daviditen:
Looks good to me.

ronawho:
Can you link to the gmp documentation that says m4 is required? I would imagine that m4 is only required for gmp users building from source, not users using the gmp tarball. i.e. autoconf/m4/aclocal should only be required if configure needs to be rebuilt from configure.ac, which we don't want users to do. It's possible that we need to touch some files like we do for [qthreads](https://github.com/chapel-lang/chapel/blob/master/third-party/qthread/Makefile#L114-L120)/[hwloc](https://github.com/chapel-lang/chapel/blob/master/third-party/hwloc/Makefile#L63-L69) to prevent this.

lydia-duncan:
https://gmplib.org/manual/Notes-for-Particular-Systems.html#Notes-for-Particular-Systems, which I reached through https://gmplib.org/manual/Installing-GMP.html#Installing-GMP with the Notes for Particular Systems link

lydia-duncan:
My understanding of our configuration set up for GMP is very limited, so it may very well be that something we're doing is messing it up somehow

ronawho:
I don't see anything talking about m4 being required for official gmp releases. And we didn't see issues last release (before we started storing gmp untarred.)

I would guess we need to do something like what we do for qthreads/hwloc (https://github.com/chapel-lang/chapel/issues/7665 has some more info)

For now I think I would close this and open an issue along the lines of "gmp seems to require m4" and assign it to me (and sorry, I wasn't following this thread at all before, so I didn't piece together that it might have been a bug introduced from https://github.com/chapel-lang/chapel/pull/7014)

PR #7700, opened 2017-10-31
With mem=jemalloc, comm=ugni, and hugepages, support dynamically extendable heap.

gbtitus:
In the past we have only provided a fixed-size registered heap when
CHPL_COMM=ugni.  Here, build on the dynamic registration capability
recently added for array allocations, to support dynamic extension of
the main heap also, even when it is composed of registered memory.

### DIFF
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -131,7 +131,7 @@ static inline
 void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
                            chpl_bool* callAgain, void* repeat_p,
                            int32_t lineno, int32_t filename) {
-  size_t size = nmemb * eltSize;
+  const size_t size = nmemb * eltSize;
   void* p;
 
   //
@@ -153,7 +153,8 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
 
     p = NULL;
     *callAgain = false;
-    if (size >= chpl_comm_regMemAllocThreshold()) {
+    if (size >= chpl_comm_regMemAllocThreshold()
+        && size >= 2 * chpl_comm_regMemHeapPageSize()) {
       p = chpl_comm_regMemAlloc(size);
       if (p != NULL) {
         *callAgain = true;
@@ -198,8 +199,15 @@ static inline
 void chpl_mem_array_free(void* p,
                          size_t nmemb, size_t eltSize,
                          int32_t lineno, int32_t filename) {
-  if (!chpl_comm_regMemFree(p, nmemb * eltSize))
-    chpl_mem_free(p, lineno, filename);
+  const size_t size = nmemb * eltSize;
+
+  if (size >= chpl_comm_regMemAllocThreshold()
+      && size >= 2 * chpl_comm_regMemHeapPageSize()) {
+    if (chpl_comm_regMemFree(p, size))
+      return;
+  }
+
+  chpl_mem_free(p, lineno, filename);
 }
 
 static inline
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -128,6 +128,17 @@ void chpl_mem_free(void* memAlloc, int32_t lineno, int32_t filename) {
 }
 
 static inline
+chpl_bool chpl_mem_size_justifies_comm_alloc(size_t size) {
+  //
+  // Don't try to use comm layer allocation unless the size is beyond
+  // the comm layer threshold and enough pages to make localization
+  // worthwhile.
+  //
+  return (size >= chpl_comm_regMemAllocThreshold()
+          && size >= 2 * chpl_comm_regMemHeapPageSize());
+}
+
+static inline
 void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
                            chpl_bool* callAgain, void* repeat_p,
                            int32_t lineno, int32_t filename) {
@@ -153,8 +164,7 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
 
     p = NULL;
     *callAgain = false;
-    if (size >= chpl_comm_regMemAllocThreshold()
-        && size >= 2 * chpl_comm_regMemHeapPageSize()) {
+    if (chpl_mem_size_justifies_comm_alloc(size)) {
       p = chpl_comm_regMemAlloc(size);
       if (p != NULL) {
         *callAgain = true;
@@ -201,10 +211,15 @@ void chpl_mem_array_free(void* p,
                          int32_t lineno, int32_t filename) {
   const size_t size = nmemb * eltSize;
 
-  if (size >= chpl_comm_regMemAllocThreshold()
-      && size >= 2 * chpl_comm_regMemHeapPageSize()) {
-    if (chpl_comm_regMemFree(p, size))
-      return;
+  //
+  // If the size indicates we might have gotten this memory from the
+  // comm layer then try to free it there.  If not, or if so but the
+  // comm layer says it didn't come from there, free it in the memory
+  // layer.
+  //
+  if (chpl_mem_size_justifies_comm_alloc(size)
+      && chpl_comm_regMemFree(p, size)) {
+    return;
   }
 
   chpl_mem_free(p, lineno, filename);
--- a/runtime/src/comm/ugni/comm-ugni-heap-pages.c
+++ b/runtime/src/comm/ugni/comm-ugni-heap-pages.c
@@ -21,11 +21,13 @@
 // System and heap page sizes, for the uGNI communication interface.
 //
 
+#include <pthread.h>
 #include <stdint.h>
 #include <stdlib.h>
 
 #include "chplrt.h"
 #include "chplcgfns.h"
+#include "chpl-env.h"
 #include "comm-ugni-heap-pages.h"
 #include "error.h"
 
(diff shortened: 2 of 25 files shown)


gbtitus:
@ronawho, would you do the honors?  It'll almost certainly be easiest to review this commit by commit.

runtime/include/chpl-mem.h
```diff
@@ -153,7 +153,8 @@ void* chpl_mem_array_alloc(size_t nmemb, size_t eltSize, c_sublocid_t subloc,
 
     p = NULL;
     *callAgain = false;
-    if (size >= chpl_comm_regMemAllocThreshold()) {
+    if (size >= chpl_comm_regMemAllocThreshold()
+        && size >= 2 * chpl_comm_regMemHeapPageSize()) {
```
ronawho: I'd prefer to put the `size >= chpl_comm_regMemAllocThreshold()  && size >= 2 * chpl_comm_regMemHeapPageSize()` code in some helper function instead of duplicating the logic in alloc and free

runtime/include/chpl-mem.h
```diff
@@ -198,8 +199,15 @@ static inline
 void chpl_mem_array_free(void* p,
                          size_t nmemb, size_t eltSize,
                          int32_t lineno, int32_t filename) {
-  if (!chpl_comm_regMemFree(p, nmemb * eltSize))
-    chpl_mem_free(p, lineno, filename);
+  const size_t size = nmemb * eltSize;
+
+  if (size >= chpl_comm_regMemAllocThreshold()
+      && size >= 2 * chpl_comm_regMemHeapPageSize()) {
+    if (chpl_comm_regMemFree(p, size))
```
ronawho: This seems shady to me. Why does `chpl_comm_regMemFree` have a return value? It seems like it's return value is whether memory can actually be registered with the NIC, but that seems like a bit of an abuse to me.

runtime/src/comm/ugni/comm-ugni-heap-pages.c
```diff
@@ -0,0 +1,104 @@
...
+#include "comm-ugni-heap-pages.h"
+#include "error.h"
+
+
+//
+// This is really just a utility function, but exposing it makes things
+// easier for chpl_comm_ugni_getHeapPageSize() callers in the launcher,
+// where the regular chplsys.c support isn't available.
+//
+size_t chpl_comm_ugni_getSysPageSize(void) {
```
ronawho: This feels gross to me. I really don't like having this and `chpl_getSysPageSize()`. I think we should find some way to unify their usage.

runtime/src/comm/ugni/comm-ugni-heap-pages.c
```diff
@@ -0,0 +1,104 @@
...
+    chpl_internal_error("can't get system page size");
+
+  return sps;
+}
+
+
+//
+// Figure out what the heap page size will be.
+//
+size_t chpl_comm_ugni_getHeapPageSize(void) {
```
ronawho: Similarly, it seems like we should find some way of using `chpl_getHeapPageSize()` or something

ronawho: Oh wait, maybe it's not that simple since this check if jemalloc and other things are enabled

runtime/src/comm/ugni/comm-ugni-heap-pages.c
```diff
@@ -0,0 +1,104 @@
...
+    return (hps = chpl_comm_ugni_getSysPageSize());
+  }
+
+  //
+  // Figure out the heap page size.
+  //
+  int num_scanned;
+  char units;
+
+  if ((num_scanned = sscanf(ev, "%zi%c", &hps, &units)) != 1) {
```
ronawho: Do we still not have any common env processing code? I think we have some for parsing bool env vars, but it seems like we duplicate this logic enough that we should have a wrapper function to parse B/KB/GB env vars

runtime/src/comm/ugni/comm-ugni-launch.c
```diff
@@ -23,16 +23,72 @@
...
+  char buf[1000];
+
+  // sanity check: power of 2 and >= sys page size
+  assert((hps & (hps - 1)) == 0
+         && hps >= chpl_comm_ugni_getSysPageSize());
+
+  // lg_chunk is specified as the base-2 log of the expansion chunk size
+  hps_log2 = lrint(log2((double) hps));
+
+  if ((ev = getenv("CHPL_JE_MALLOC_CONF")) == NULL) {
```
ronawho: We shouldn't be hardcoding `CHPL_JE_MALLOC_CONF`. We should be using an uppercased `CHPL_JEMALLOC_PREFIX`, which is defined in third-party/jemalloc/Makefile.include

If that isn't defined then we may be using a system jemalloc (not officially supported today, but may be in the future) so we should default to `JE_`. 

The top of `runtime/include/mem/jemalloc/chpl-mem-impl.h` shows what we do with getting the prefix for the alloc routines

ronawho: Also, I'm trying to figure out the best way to maintain this code if for example a jemalloc upgrade changes the name of the env var. I'm not going to remember that this code needs to be updated, so I was wondering if we could put a function in the jemalloc shim or header for this code to call. 

I'm not sure how that would work since this will be called from the launcher though.. 

At the very least I think we need a note in the jemalloc README about checking this file during version upgrades, but i'm open to other suggestions

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -70,46 +72,79 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
...
+    }
+
+    //
+    // Localize the new memory via first-touch, by storing to each page.
+    // This will give the memory affinity to the NUMA domain (if any)
+    // we were running on when we allocated it.  It also commits the
+    // memory, in the jemalloc sense.
+    //
+    for (int i = 0; i < size; i += heap.page_size) {
+      ((char*) cur_chunk_base)[i] = 0;
```
ronawho: Seems like this should use memset, since memset should be faster (vectorized and maybe other optimizations)

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -70,46 +72,79 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
...
+    // Support useUpMemNotInHeap by hiding the fixed/dynamic distinction
+    heap.base = cur_chunk_base;
+    heap.size = size;
+    heap.cur_offset = 0;
+  }
 
   // jemalloc 4.5.0 man: "Zeroing is mandatory if *zero is true upon entry."
   if (*zero) {
-     memset(cur_chunk_base, 0, size);
+    memset(cur_chunk_base, 0, size);
```
ronawho: You can skip this in the case where you've already zero'd above, right?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -43,9 +43,11 @@
 #endif
 
 static struct shared_heap {
+  chpl_bool isFixed;
   void* base;
   size_t size;
   size_t cur_offset;
+  size_t page_size;  // valid iff !isFixed
```
ronawho: I don't see any reason to store this in the shared_heap struct. Seems like it could just be a static var in the `!isFixed` branch of the alloc routine

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -43,9 +43,11 @@
 #endif
 
 static struct shared_heap {
+  chpl_bool isFixed;
```
ronawho: More descriptive name? Maybe `isFixedSize` or `isExtendable`? Oh, and maybe underscore instead of camelCase to match the others?

runtime/src/mem/jemalloc/mem-jemalloc.c
```diff
@@ -70,46 +72,79 @@ static inline void* alignHelper(void* base_ptr, size_t offset, size_t alignment)
...
+    // Update the current pointer, now that we've past any early returns.
+    heap.cur_offset = cur_heap_size + size;
 
-  // If there's not enough space on the heap for this allocation, return NULL
-  if (size > heap.size - cur_heap_size) {
+    // now that cur_heap_offset is updated, we can unlock
     pthread_mutex_unlock(&heap.alloc_lock);
-    return NULL;
-  }
+  } else {
```
ronawho: How confident are you that we don't need to use the alloc_lock here? Is it safe for `chpl_comm_regMemAlloc()` and  `chpl_comm_regMemPostAlloc()` be called concurrently?

ronawho:
Overall 👍 

I had a bunch of feedback/questions in line. 

The only must change is that we shouldn't be hardcoding `CHPL_JEMALLOC_PREFIX`, others can either be ignored or cleaned up post-commit.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2507,164 +2541,198 @@ void chpl_comm_rollcall(void)
...
-    size_t size;
-    size_t decrement;
-    void*  start;
+  //
+  // The heap is supposed to be of fixed size and on hugepages.  Set
+  // it up.  (If it's to be dynamically extensible, that's handled
+  // separately through chpl_comm_regMem*().)
+  //
+  const size_t nic_max_pages = (size_t) 1 << 14; // not publicly defined
+  const size_t nic_max_mem = nic_max_pages * hugepage_size;
```
ronawho: I have a really hard time following when hugepage_size is set or not. It might be easier to follow if it was a static inline function instead of a global, but if you're confident that it's set when it needs to be, that's good enough for me

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: Notice the gigantic stride.  We're not actually zeroing memory here.  We're just touching each page to force it into existence and thus localize it.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: Not necessarily.  If the memory delivered by the comm layer a few lines previously is on old pages instead of new ones, then it may have non-0 contents and need zeroing.  However, we could look into whether there's a way to tell this and avoid the zeroing cost if the pages are new, because new pages are guaranteed to be zero I believe (though I'd have to go search for confirmation of that).  Agreed it would be nice to avoid the zeroing code here if we can, because these are big chunks.

runtime/include/chpl-mem.h
gbtitus: Done (5a1b0ec7).

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: Oooh, duh. Makes sense, nevermind me then. 

I'm fine with this code -- just curious if there's an equivalent hwloc function, and if so if there's any advantages to that approach.

runtime/src/mem/jemalloc/mem-jemalloc.c
ronawho: Nah, I missed that the above code wasn't zeroing the entire chunk, so ignore this comment

runtime/include/chpl-mem.h
gbtitus: 5a1b0ec7 added a comment here to help explain, but also see the existing [block comment](https://github.com/chapel-lang/chapel/blob/master/runtime/include/chpl-comm.h#L246) which explains why `chpl_comm_regMemFree()` returns a "mine vs. not mine" indicator.

runtime/src/comm/ugni/comm-ugni-heap-pages.c
gbtitus: This has now been addressed, in 79d37249.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: There is a hwloc function that will set the locality of memory (that is, pin it).  And there are reasons for and against using it.  The main differences are that touching the memory as we do here sets the locality implicitly (to match the NUMA domain the touching thread executed on) and only sets locality if the page doesn't have any yet, that is, it only has an affect the first time a page is referenced.  But if the page already exists, then it nothing changes.  The hwloc function, on the other hand, accepts an explicit NUMA domain id.  So it can localize the page to a different NUMA domain than we're running on, which might be useful if our expected reference pattern doesn't match that of our parallel initialization code.  Also, it can optionally force the desired locality.  That is, if the page exists and isn't on the desired NUMA domain, it can physically move it to the desired one.  That always gives the desired locality, but does require actually copying the page contents.  So we might want to use the hwloc function if we have memory the parallel init code doesn't work well on, or if we find ourselves reusing memory, but so far neither of those seem to happen much.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: Libhugetlbfs can back and/or replace malloc, so it should be thread-safe.  And the kernel code that handles hugepages (touch, etc.) certainly should be.  The comm layer would need to protect concurrent access to its data structures for managing registrations and so on, but the only implementation (comm=ugni, so far) does do so.  So yes, this should be safe.

runtime/src/comm/ugni/comm-ugni-launch.c
gbtitus: Fixed the jemalloc README in 88c499a6.

runtime/src/mem/jemalloc/mem-jemalloc.c
gbtitus: Fixed page_size and isFixed in 51a7c566.

runtime/src/comm/ugni/comm-ugni.c
gbtitus: Done, in e3507715.

PR #7707, opened 2017-11-01
isIterable

ben-albrecht:
Add a way to check if an object is iterable.

### DIFF
--- a/modules/standard/Types.chpl
+++ b/modules/standard/Types.chpl
@@ -413,8 +413,8 @@ pragma "no doc"
 proc isProperSubtype(type sub, type super) param
   return false;
 
-
-
+/* Returns `true` if `x` contains a `these()` method. */
+proc isIterable(x) param return canResolveMethod(x, 'these');
 
 // Returns true if it is legal to coerce t1 to t2, false otherwise.
 pragma "no doc"
--- a/test/types/iterable/isiterable.chpl
+++ b/test/types/iterable/isiterable.chpl
@@ -1,0 +1,48 @@
+// Some types
+youAreNotIterable(bool);
+youAreNotIterable(int);
+youAreNotIterable(uint);
+youAreNotIterable(real);
+youAreNotIterable(imag);
+youAreNotIterable(complex);
+youAreIterable(string);
+youAreIterable(domain(1));
+
+// Some values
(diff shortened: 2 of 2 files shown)


modules/standard/Types.chpl
```diff
@@ -413,8 +413,8 @@ pragma "no doc"
 proc isProperSubtype(type sub, type super) param
   return false;
 
-
-
+/* Returns `true` if `x` contains a `these()` method. */
+proc isIterable(x) param return canResolveMethod(x, 'these');
```
cassella: Should this check that ``these`` is an ``iter``, not a ``proc``?

ben-albrecht: I think that's a good idea, but it will require some modifications to the Reflection module and compiler.

daviditen:
Should `isIterable` return `true` for iterable expressions?

Both of these are `false` in this PR:

```
isIterable(for i in 1..4 do i*2);
isIterable(myIterator());
```

ben-albrecht:
@daviditen - That's an interesting question. Both of the cases you pointed out return `true` in this PR if you capture the expressions in a variable, which I'd argue is the more valuable use-case:

```chpl
use Reflection;

iter f() {
  yield 1;
}

var g = f();
writeln(canResolveMethod(g, 'these')); // true

var h = for i in 1..4 do i**2;
writeln(canResolveMethod(h, 'these')); // true (because h is now an array)
```

I think supporting expressions like this will require some changes in resolution, and will likely open a bigger can of worms. I'd argue this PR shouldn't depend on supporting these edge cases, but they should be documented as separate issues & futures. What do you think?

daviditen:
As you note in the comment in your example, once you capture them into a variable they're just regular arrays.  I'm hesitant to have a function called `isIterable` that returns false for iterable expressions.

ben-albrecht:
> I'm hesitant to have a function called isIterable that returns false for iterable expressions.

@mppf expressed a similar concern offline. That makes 2 to my 1. I'll look into what resolution changes are necessary to make this work as expected.

ben-albrecht:
Since the scope of this PR has grown significantly beyond my 1-line solution, I'm not sure when I'll get a chance to implement it. Going to close until I have time to revisit this.

PR #7709, opened 2017-11-01
Support subtype queries on distributions

benharsh:
This PR allows users to write subtype queries between distributions and classes inheriting from BaseDist. In the old world users would have to use _value to check the type:
```chpl
where domain.dist._value : Block
```
Now, users can simply query ``dist``:
```chpl
where domain.dist : Block
```

Testing:
- [x] full local
- [x] full gasnet


bradcray:
Belatedly: Do you think we need any documentation updates related to this PR?

benharsh:
Good question. I'd be inclined to make an update where we describe "domain.dist", but I don't think we specify that anywhere today.

PR #7719, opened 2017-11-01
Handle CHPL_COMM_USE_GDB in Chapel runtime (vs. GASNet)

### LINKED ISSUE #5269: Try to avoid patching GASNet
We have a couple of patches that we apply on top of GASNet releases in order to support launching an xterm window to run a debugger and for initial 64-bit arm support. 

I imagine the 64-bit arm support will come in later GASNet releases (it's just a cross-configure file), but it'd be nice to generalize the xterm support and contribute it upstream so we don't have to patch every time we upgrade gasnet versions.

gbtitus:
With `comm=gasnet` and the `amudprun` launcher, by setting `CHPL_COMM_USE_GDB`
or `CHPL_COMM_USE_LLDB` a user can cause the target program instances to
be run under the `gdb` or `lldb` debugger, respectively, with each instance
in its own `xterm` window.  In the past we've supported this through a mod
to the GASNet `amudprun` launcher, which we've had to re-apply each time
we upgraded GASNet.  Here we remove that patch and replace it with code
in our own Chapel `amudprun` launcher to achieve the same effect.

This resolves #5269.

### DIFF
--- a/runtime/include/chpl-mem.h
+++ b/runtime/include/chpl-mem.h
@@ -265,6 +265,9 @@ void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
 #define chpl_mem_allocMany(number, size, description, lineno, filename)        \
   sys_malloc((number)*(size))
 
+#define chpl_mem_alloc(size, description, lineno, filename)        \
+  sys_malloc(size)
+
 #define chpl_mem_free(ptr, lineno, filename)        \
   sys_free(ptr)
 
--- a/runtime/src/launch/amudprun/launch-amudprun.c
+++ b/runtime/src/launch/amudprun/launch-amudprun.c
@@ -102,8 +102,15 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
       if (chpl_run_cmdstr("which xterm", xterm_path, sizeof(xterm_path)) > 0) {
         largv[largc++] = (char *) strdup(xterm_path);
         largv[largc++] = (char *) "-e";
-        largv[largc++] = (char *) "gdb";
-        largv[largc++] = (char *) "--args";
+        if (ev_use_gdb != NULL) {
+          largv[largc++] = (char *) "gdb";
+          largv[largc++] = (char *) "--args";
+        } else {
+          largv[largc++] = (char *) "lldb";
+          largv[largc++] = (char *) "--";
+        }
+      } else {
+        chpl_warning("CHPL_COMM_USE_(G|LL)DB ignored because no xterm", 0, 0);
       }
     }
   }
--- a/runtime/src/launch/amudprun/launch-amudprun.c
+++ b/runtime/src/launch/amudprun/launch-amudprun.c
@@ -97,10 +97,13 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
     const char* ev_use_lldb = getenv("CHPL_COMM_USE_LLDB");
 
     if (ev_use_gdb != NULL || ev_use_lldb != NULL) {
-      char xterm_path[PATH_MAX];  // understood questionable; hopefully enough
+      // hopefully big enough; PATH_MAX is problematic, but what's better?  
+      const size_t xterm_path_size = PATH_MAX;
+      char *xterm_path = chpl_mem_alloc(xterm_path_size,
+                                        CHPL_RT_MD_COMMAND_BUFFER, -1, 0);
 
-      if (chpl_run_cmdstr("which xterm", xterm_path, sizeof(xterm_path)) > 0) {
-        largv[largc++] = (char *) strdup(xterm_path);
+      if (chpl_run_cmdstr("which xterm", xterm_path, xterm_path_size) > 0) {
+        largv[largc++] = xterm_path;
         largv[largc++] = (char *) "-e";
         if (ev_use_gdb != NULL) {
           largv[largc++] = (char *) "gdb";
--- a/runtime/src/launch/amudprun/launch-amudprun.c
+++ b/runtime/src/launch/amudprun/launch-amudprun.c
@@ -78,13 +78,35 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
                                       int argc, char* argv[],
                                       int32_t numLocales) {
   static char nlbuf[16];
-  const int largc = 3;
-  char *largv[largc];
+  int largc;
+  const int largv_size = 7;
+  char *largv[largv_size];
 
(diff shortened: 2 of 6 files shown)


gbtitus:
I haven't tested this on the full suite but that seems not worthwhile anyway, because `git grep` says the string `CHPL_COMM_USE` doesn't occur in the `test/` hierarchy.  I have tested it on both Linux (with gdb) and MacOS (with lldb), and on MacOS I also temporarily made my `/opt/X11/bin/xterm` unrunnable to make sure the warning message for "can't find xterm" worked properly.

gbtitus:
@ronawho, care to take a look?

ronawho:
Need to update the GASNet README deleting the note about this patch.

runtime/src/main_launcher.c
```diff
@@ -186,6 +186,40 @@ chpl_run_utility1K(const char *command, char *const argv[], char *outbuf, int ou
...
+// This is an even simpler run-a-command utility.  The command is just
+// a string to be run, by something like `/bin/sh -c "commandStr"`.  It
+// must not expect any input.  Its output to stdout is placed in outbuf,
+// truncated to fit if necessary.  Its output to stderr is discarded.
+// On success, the return value is the number of bytes in outbuf.  On
+// failure, it is -1; output may have been placed in outbuf in this
+// case, but if so it should be ignored.
+//
+int
+chpl_run_cmdstr(const char *commandStr, char *outbuf, int outbuflen) {
```
ronawho: Why do you need to add this instead of just using `chpl_run_utility1K()`?

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -78,13 +78,42 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
...
+  // debugger window, arrange for that.
+  //
+  {
+    const char* ev_use_gdb = getenv("CHPL_COMM_USE_GDB");
+    const char* ev_use_lldb = getenv("CHPL_COMM_USE_LLDB");
+
+    if (ev_use_gdb != NULL || ev_use_lldb != NULL) {
+      char xterm_path[PATH_MAX];  // understood questionable; hopefully enough
+
+      if (chpl_run_cmdstr("which xterm", xterm_path, sizeof(xterm_path)) > 0) {
```
ronawho: What's the advantage of shelling out to `which` instead of using `access()` (full disclosure, I have not looked up what `access()` does

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -78,13 +78,42 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
...
+        largv[largc++] = (char *) strdup(xterm_path);
```
ronawho: Use of `strdup` is going to trip up the "check for sys alloc calls in the runtime" check

I'm pretty confident we don't have a chpl_strdup (there's a qio_strdup, but that's not available to the launchers)

I think you could either duplicate the stdup logic (copying what qio_strdup does), or we could tell the check to ignore launcher code.

runtime/src/launch/amudprun/launch-amudprun.c
gbtitus: They're quite different.  `which` looks for the given executable in all paths in PATH and returns the first one it finds; basically it does what the shell would do if you ran that executable without a path.  `access()` gives the access bits for a given path.

ronawho:
👍 Looks good to me. Had some nits inline, the only big thing is the smoke check won't pass with `strdup` being used.

I have never used `CHPL_COMM_USE_{GDB,LLDB}` personally though. Maybe @benharsh could give it a try?

runtime/src/main_launcher.c
gbtitus: The interface to chpl_run_utility1K() seemed unnecessarily complicated because of the argv[] vector.  In many cases all that's needed is just to give a command string to a shell, which is what this does.

runtime/src/launch/amudprun/launch-amudprun.c
gbtitus: Oopsie.  Fixed this -- thanks.

runtime/src/main_launcher.c
ronawho: I have a mild preference for using the other function since it's more tested. That's not a great reason though, so I'm fine with keeping this

runtime/include/chpl-mem.h
```diff
@@ -265,6 +265,9 @@ void chpl_mem_layerFree(void*, int32_t lineno, int32_t filename);
 #define chpl_mem_allocMany(number, size, description, lineno, filename)        \
   sys_malloc((number)*(size))
 
+#define chpl_mem_alloc(size, description, lineno, filename)        \
```
ronawho: While I sympathize with wanting this, I don't think we should have this here, and not in the non-launcher case above. Historically we just call chpl_mem_allocMany with number=1, and I'd prefer to keep doing instead of introducing the number=1 special case here.

gbtitus: I don't think historical usage is a very good guide here.  As far as I've ever been able to tell the only reason the launchers were in the runtime hierarchy was to share a little bit of the runtime code.  Or really, *appear* to share, since in most cases the `.h` files define the same thing quite differently on two sides of a big `#ifndef LAUNCHER ... #else ... #endif` construct.  That's the case for `chpl_mem_allocMany()`, for example.  But then the desire to minimize the resulting special-casing in the defining .h files has led us to provide the bare minimum of support in, again for example, things like the memory interface.  And that in turn has led to stilted code: unless I missed something in my grep just now, with but one exception, every `chpl_mem_allocMany()` call in the `runtime/src/launch/` hierarchy passes `sizeof(char)` as its second argument.  That's silly.  I appreciate the desire not to do too much work in the existing launcher code because we'd like to do a wholesale replacement at some point, but until that's the most important next thing for someone to work on I don't believe it's unreasonable to make improvements, and this seems like an improvement.  (In fact, if it weren't for the fact that I would have had to test every single launcher, I would have changed all those `chpl_mem_allocMany(..., sizeof(char), ...)` calls to `chpl_mem_alloc(...)` calls also, in that last commit.)

ronawho: Man, I'm really off my reviewing game..

I had thought that you were adding a brand new`chpl_mem_alloc()` routine, but now I see that it was already defined for regular runtime (non-launcher) code, and you're just adding it to the launcher code now. Ignore me

Issue #7718, opened 2017-11-01
Sporadic shutdown failures with gasnet udp and segment fast/large


ronawho:
Originally from https://chapel.atlassian.net/browse/CHAPEL-274 (internal)

For our udp-fast testing, we often seen nightly failures along the lines of:

```
*** FATAL ERROR: Master got an xSocket: error in sendAll() - connection closed (32:Broken pipe)
```

This is for `-nl1` tests that create some amount of tasks and halt (abnormal termination.) Here's the simplest reproducer we have:

```chpl
coforall i in 1..here.maxTaskPar do
  if i == 1 then
    halt("Halting!");
```

Note that only one task is calling halt, and the others are no-ops. It doesn't seem to fail with just 1 task, and it fails much less often with just 2 tasks. Unfortunately, we've been unable to come up with a simple C+GASNet reproducer.

Here's the little bash script I'm using to repro to compile and run the above example a bunch of times (it fails 1/50 times for me on a 24 core linux box):

```bash
#!/usr/bin/env bash

rm -f gn
chpl gn.chpl

NUM_TRIALS=${NUM_TRIALS:-1000}

for i in `seq ${NUM_TRIALS}`; do
  output=$(./gn -nl1 2>&1)
  if [ "${output}" != "gn.chpl:3: error: halt reached - Halting!" ]; then
    echo "Trial ${i} failed:"
    echo "${output}"
    exit
  fi
done

echo "No failures after ${NUM_TRIALS} trials!"
```

Note that it sporadically fails for udp with segment large/fast, but doesn't seem to fail with segment everything. I also can't reproduce with smp, mpi, ibv, or aries substrates.

Here's an (almost certainly wrong) patch to amudp that seems to quiet the regression.

```diff
diff --git a/third-party/gasnet/gasnet-src/other/amudp/amudp_spmd.cpp b/third-party/gasnet/gasnet-src/other/amudp/amudp_spmd.cpp
index d8a7296..bcdb43d 100644
--- a/third-party/gasnet/gasnet-src/other/amudp/amudp_spmd.cpp
+++ b/third-party/gasnet/gasnet-src/other/amudp/amudp_spmd.cpp
@@ -927,9 +927,11 @@ pollentry:
                 // TODO: perhaps use an active message for this? for now, just rely on coord socket dying
                 exitCode_nb = hton32(0);
                 for (int i=0; i < (int)coordList.getCount(); i++) {
-                  sendAll(coordList[i], "E");
[...]

PR #7722, opened 2017-11-01
Shutdown the memory layer after the comm layer

ronawho:
Previously we were shutting down the memory layer before the comm layer, but
the comm layer should be allowed to allocate/deallocate during exit, so this
was the wrong order. Move mem exit after comm exit, and only run mem exit if
we're doing an orderly "all" exit.

We don't shutdown the tasking layer for an abnormal "any" exit because anything
could still be going on and we just want to shutdown without cleaning up much.
Similarly we don't want to try to shutdown the memory layer, because other
things (like the tasking or comm layer) could still be trying to allocate and
deallocate memory and if the memory layer is shutdown this could just lead to
other confusing errors instead of just exiting.

This resolves https://github.com/chapel-lang/chapel/issues/7718 and https://chapel.atlassian.net/browse/CHAPEL-274

### DIFF
--- a/runtime/src/chplexit.c
+++ b/runtime/src/chplexit.c
@@ -41,9 +41,11 @@ static void chpl_exit_common(int status, int all) {
     chpl_task_exit();
     chpl_reportMemInfo();
   }
-  chpl_mem_exit();
   chpl_comm_exit(all, status);
-  chpl_topo_exit();
+  if (all) {
+    chpl_mem_exit();
+    chpl_topo_exit();
+  }
   exit(status);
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus and @benharsh mind reviewing?

Passes 10,000 trials for the reproducer in #7718 and for arrays/johnk/associative/tooFewKeys

gbtitus:
Looks great -- good point about not shutting down mem or topo during abnormal exit due to other layers possibly using them.   👍

PR #7735, opened 2017-11-02
Proposed spec change for param/type fields accessible from type

mppf:
This is a proposed specification change to go along with PR #7672.


daviditen:
For each of the updated sections, I'd add the word _instantiated_ to make it clear that you can't access the fields through a generic class or record.  For example:

> In addition to being available for access with a class
> instance receiver, parameter and type fields can be accessed from the
> _instantiated_ class type itself.

Issue #7511, opened 2017-09-29
Compiler failure in main module naming conflict


lydia-duncan:
### Summary of Problem
A user discovered that when two inner modules are provided with the same name
and both define a main, if the ambiguous name is specified with the
--main-module flag then the compiler will arbitrarily choose the later module
as the desired module and, in deadCodeElimination, have an issue with cleaning
up the module whose main was not used. Though the bug gets tickled at
deadCodeElimination, the theory is that this is related to resolution of the
main function and potentially a bug in scope resolve's resolution of module
names - ideally, we would like to see an error because M is ambiguous instead
of arbitrarily choosing one or the other.

This relates to #7510 

### Steps to Reproduce

**Source Code:**

```chapel
module X {
  module M {
    proc main() { writeln("X.M"); }
  }
}

module Y {
  module M {
    proc main() { writeln("Y.M"); }
  }
}
```

**Compile command:**
`chpl --main-module M foo.chpl`

**Associated Future Test(s):**
[`test/modules/cassella/ambigmodmain.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/modules/cassella/ambigmodmain.chpl)
[`test/modules/cassella/ambigmodmain.unnested.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/modules/cassella/ambigmodmain.unnested.chpl)
[`test/modules/cassella/ambigmodmain.unnested2.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/modules/cassella/ambigmodmain.unnested2.chpl)

### Configuration Information

- Output of `chpl --version`: 1.16.0 (pre-release)

noakesmichael:
This is just a mess with two different bugs.

1) An early part of the compiler will make an "arbitrary"
choice of the 'M' to use without regard to nesting.

2) Depending on details of the ordering of top level
modules on files and nesting, the compiler may then
get an internal error in dead code elimination.


Problem (1) can be solved by requiring that the "main module"
be a top level module or allowing the --main-module command
to accept a path.  These are equally easy to implement and so
Brad proposed interpreting --main-module as a path.

We've seen variations of (2) before.  I will solve this once we can
pick modules reliably.

proposed

lydia-duncan:
Okay, sounds good to me, just as long as we don't forget about the other part (which your comment should prevent)

noakesmichael:
I’ve already fixed bug 1 and am now working on bug 2



[...]

PR #7736, opened 2017-11-02
Main module

noakesmichael:
Resolve issue #7511 

This issue contained several sub-issues.

1) If there are multiple modules with a definition of main() that
have the same "local name" (e.g. 'M') but that are nested in
differing ways in other modules then

a) the --main-module option did not provide a way to define
the "path" to the desired 'M'.

b) if the user said simply 'M' then the compiler would pick the
"last one" it found.

2) Depending on other issues the compiler might then
get an internal fatal in dead code elimination.

The first commit extends ModuleSymbol to a) introduce a
notion of a "path name" for a module rather than merely a
"name".  b) extends the handling of --main-module to accept
a "path" and match against the module math.

The second commit addresses the trivial hole in dead code
elimination.

The remaining commits clean up the associated futures and
extend .compopts for each test to get the full coverage.

Compiled with/without CHPL_DEVELOPER on clang/darwin
and gcc/linux64.  Focussed testing of modules/cassella for
all configs.  Also some limited testing of release/


compiler/AST/ModuleSymbol.cpp
```diff
@@ -424,8 +422,33 @@ void ModuleSymbol::printTableOfContents(std::ostream* file) {
...
+  return name;
+}
+
+/*
+ * Generate a name that represents the path to the module
+ * For a top-level module this is simply the name.
+ * For nested modules that name corresponds to the "use name"
+ */
+
+std::string ModuleSymbol::path() const {
```
lydia-duncan: I suspect that chpldoc might do something similar when generating the name for a module.  Could it make use of this new function?  (not a priority, just a musing)

bradcray:
Belatedly: Should this PR have accompanying documentation changes?  I'm imagining that the --main-module description in the man page might need an update, for instance.

PR #7738, opened 2017-11-03
Restructure extendLeaderNew() and its helpers

vasslitvinov:
* Restructure most of the "new" "Part 2" part of the implementation
of forall intents, specifically extendLeaderNew() and its helpers.
It still handles a variety of cases as before, now hopefully in
a more comprehensible manner.

* The following case is now handled nicely and uniformly
for all forall intents. The case is when the parallel iterator
(the one that is invoked by the forall loop)
itself contains a parallel loop (i.e. a forall or a for-loop
that invokes a leader or standalone iterator explicitly).

* The case from the previous bullet where the parallel loop is
a for-loop is handled by converting it to a forall loop first.
Handling a forall loop in this context is much simpler than
handling a for-loop, partly due to a better AST representation.
As a result, `eflopiMap` is no longer needed for the "new" implementation.

* The case when a yield in the parallel iterator is nested
inside 2 (or more) levels of parallel iterators has been causing INT_ASSERT.
Now it generates a "currently not implemented" error instead.

* ... except for the standalone iterator `walkdirs` in `FileSystem` module,
as tested by `modules/standard/FileSystem/filerator/bradc/findfiles-par`.

* To handle walkdirs and findfiles-par, recursive iterators are now supported
when the recursive invocation is made in a forall loop. (Previously it worked
only when such invocation was in a for-loop.) This works as long as there
are no reduce intents involved.

This PR's commits before squashing are 7278920..15b6a7e.

Testing:
- [x] linux64 --verify
- [x] baseline

### DIFF
--- a/compiler/include/ForallStmt.h
+++ b/compiler/include/ForallStmt.h
@@ -32,9 +32,13 @@ class ForallStmt : public Stmt
 public:
   bool       zippered()       const; // was 'zip' keyword used?
   AList&     inductionVariables();   // DefExprs, one per iterated expr
-  AList&     iteratedExpressions();  // SymExprs, CallExprs
+  AList&     iteratedExpressions();  // SymExprs, one per iterated expr
   AList&     intentVariables();      // DefExprs of LoopIntentVars
   BlockStmt* loopBody()       const; // the body of the forall loop
+
+  // when originating from a ForLoop
+  bool       iterCallAlreadyTagged() const;  // already has 'tag' actual
+  bool       noFindOuterVars()       const;  // no add'l shadow variables
 
   DECLARE_COPY(ForallStmt);
 
@@ -50,6 +54,9 @@ public:
   // for the parser
   static BlockStmt* build(Expr* indices, Expr* iterator, CallExpr* intents,
                           BlockStmt* body, bool zippered = false);
+
+  static ForallStmt* fromForLoop(ForLoop* forLoop);
+
   // helpers
   Expr* firstIteratedExpr()        const;
   int   numIteratedExprs()         const;
@@ -64,6 +71,7 @@ private:
   AList          fIterExprs;
   AList          fIntentVars;  // may be empty
   BlockStmt*     fLoopBody;    // always present
+  bool           fFromForLoop;
 
   ForallStmt(bool zippered, BlockStmt* body);
 };
@@ -74,6 +82,8 @@ inline AList& ForallStmt::inductionVariables()   { return fIterVars;   }
 inline AList& ForallStmt::iteratedExpressions()  { return fIterExprs;  }
 inline AList& ForallStmt::intentVariables()      { return fIntentVars; }
 inline BlockStmt* ForallStmt::loopBody()   const { return fLoopBody;   }
+inline bool ForallStmt::iterCallAlreadyTagged() const { return fFromForLoop; }
+inline bool ForallStmt::noFindOuterVars()       const { return fFromForLoop; }
 
 // conveniences
 inline Expr* ForallStmt::firstIteratedExpr() const { return fIterExprs.head;  }
--- a/compiler/include/ForallStmt.h
+++ b/compiler/include/ForallStmt.h
@@ -37,6 +37,7 @@ public:
   BlockStmt* loopBody()       const; // the body of the forall loop
 
   // when originating from a ForLoop
+  bool       createdFromForLoop()    const;  // is converted from a for-loop
   bool       iterCallAlreadyTagged() const;  // already has 'tag' actual
   bool       needToHandleOuterVars() const;  // yes, convert to shadow vars
 
@@ -76,14 +77,14 @@ private:
   ForallStmt(bool zippered, BlockStmt* body);
 };
 
-/* fFromForLoop / iterCallAlreadyTagged / needToHandleOuterVars
+/* fFromForLoop and its accessors
 
 These support handling of some ForLoops by converting them to ForallStmts.
 They cause skipping certain actions for these "conversion" ForallStmt nodes.
 
 Why not just have a single accessor to fFromForLoop? This is to emphasize
-that the two accessors check different properties. These two properties could
-potentially be independent of each other and of fFromForLoop.
+that the three accessors check different properties. These properties could
+potentially be independent of each other.
 
 As fFromForLoop is currently local to implementForallIntents, we may be able
 to replace fFromForLoop with a HashSet. If so, we need to ensure that the
@@ -98,6 +99,7 @@ inline AList& ForallStmt::intentVariables()      { return fIntentVars; }
 inline BlockStmt* ForallStmt::loopBody()   const { return fLoopBody;   }
 inline bool ForallStmt::iterCallAlreadyTagged() const { return fFromForLoop; }
 inline bool ForallStmt::needToHandleOuterVars() const { return !fFromForLoop; }
+inline bool ForallStmt::createdFromForLoop()    const { return fFromForLoop; }
 
 // conveniences
 inline Expr* ForallStmt::firstIteratedExpr() const { return fIterExprs.head;  }
--- a/compiler/include/ForallStmt.h
+++ b/compiler/include/ForallStmt.h
@@ -38,7 +38,7 @@ public:
 
   // when originating from a ForLoop
   bool       iterCallAlreadyTagged() const;  // already has 'tag' actual
-  bool       noFindOuterVars()       const;  // no add'l shadow variables
+  bool       needToHandleOuterVars() const;  // yes, convert to shadow vars
 
   DECLARE_COPY(ForallStmt);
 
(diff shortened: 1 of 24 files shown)


test/parallel/forall/vass/par-loops/2-enclosing-par-loops-error-4.chpl
```diff
@@ -0,0 +1,41 @@
...
+  coforall ooo in 1..3 {
+    yield 666;
+  }
+}
+
+iter pariter() {
+  yield 333;
+}
+
+iter pariter(param tag) {
```
mppf: Why doesn't this check tag== iterKind.standalone?

compiler/include/ForallStmt.h
```diff
@@ -32,10 +32,14 @@ class ForallStmt : public Stmt
...
   bool       zippered()       const; // was 'zip' keyword used?
   AList&     inductionVariables();   // DefExprs, one per iterated expr
-  AList&     iteratedExpressions();  // SymExprs, CallExprs
+  AList&     iteratedExpressions();  // SymExprs, one per iterated expr
   AList&     intentVariables();      // DefExprs of LoopIntentVars
   BlockStmt* loopBody()       const; // the body of the forall loop
 
+  // when originating from a ForLoop
+  bool       iterCallAlreadyTagged() const;  // already has 'tag' actual
+  bool       noFindOuterVars()       const;  // no add'l shadow variables
```
mppf: It seems odd to me that these two functions both just return fFromForLoop. Anyway, 'noFindOuterVars' seems like a strange method name; maybe it should be "outerVarsHandled" or something... Anyway both of these seem to have more to do with the flow of forall intents / iterator lowering than with ForallStmts fundamentally. For that reason it seems it might be better to just have an accessor for FromForLoop. Or, even better, avoid the need to even store 'fromForLoop' in this AST node at all, if it can become local information for some transformation pass.

compiler/include/expr.h
```diff
@@ -474,7 +474,10 @@ static inline bool needsCapture(FnSymbol* taskFn) {
 }
 
 inline Symbol* ShadowVarSymbol::outerVarSym() const {
-  return this->outerVarSE()->symbol();
+  if (SymExpr* ovse = this->outerVarSE())
```
mppf: It's not clear to me why you needed this change, but at least there's no reason for the assignment within the conditional. e.g.
```
SymExpr* ovse = this->outerVarSE();
if (ovse != NULL) return ovse->symbol();
else return NULL;
```

But it's obviously a style thing.

compiler/include/implementForallIntents.h
```diff
@@ -0,0 +1,56 @@
...
+                                                   Symbol* extraFormals[]);
+VarSymbol* localizeYieldForExtendLeader(Expr* origRetExpr, Expr* ref);
+void       checkAndRemoveOrigRetSym(Symbol* origRet, FnSymbol* parentFn);
+FnSymbol*  copyLeaderFn(FnSymbol* origFn, bool ignore_isResolved);
+
+void extendLeaderNew(ForallStmt* fs, FnSymbol* origIterFn, CallExpr* iterCall);
+
+static const char* astrArg(int ix, const char* base) {
+  return astr("x", istr(ix+1), "_", base);
+}
```
mppf: maybe name it something more to do with forall intents?

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+This file contains Part 2 of the new implementation of forall intents.
+"Part 2" is mostly about extending a copy of the parallel iterator in question
+with arguments and yield values corresponding to forall intents.
+
+For a reduce intent, the global reduce op instance needs to be
+properly allocated, deallocated and connected to the corresponding
+user outer variable. This is the job of "Part 1".
+
+For now, implementForallIntents2New() and immediately-adjacent functions
+are in implementForallIntents.cpp, for historical reasons.
```
mppf: Eh... why should you be conservative about moving code between the two files? If they fit better here, move them. My one "it would have been nice" request about this is that this PR seems to have a single commit combining code changes and code movement. That makes it much harder to follow. Try to move code in a separate commit.

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+  // Calling a function nested in the iterator?
+  bool         nested;
+
+  // The iterator's return symbol.
+  VarSymbol*   retSym;
+
+  // The FnSymbol for task startup/teardown code, usually 'curFn'.
+  // It is an enclosing function when descending into a task function
+  // that does not create a user-level task. Such a task function
+  // does not affect the intents.
```
mppf: I'm having trouble understanding this comment. wouldn't you want to add task startup/teardown code within the body of the task?

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+
+//
+// Is 'forLoop' a loop over a parallel iterator?
+// If so, fill in 'eInfo' with details.
+//
+static bool findCallToParallelIterator(ForLoop* forLoop, EflopiInfo& eInfo)
+{
+  Symbol* iterSym = forLoop->iteratorGet()->symbol();
+
+  // Find an assignment to 'iterSym'.
```
mppf: Is this usually located just a few statements before forLoop? If so, add a comment to that effect. If not, you might use for_SymbolSymExprs.

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+  Symbol* calltemp = toSymExpr(rhs1->get(1))->symbol();
+  CallExpr* asgnToCallTemp = NULL;
+  for (Expr* curr = asgnToIter->prev; curr; curr = curr->prev)
+    if (CallExpr* call = toCallExpr(curr))
+      if (call->isPrimitive(PRIM_MOVE))
+        if (SymExpr* lhs2 = toSymExpr(call->get(1)))
+          if (lhs2->symbol() == calltemp) {
+            asgnToCallTemp = call;
+            break;
+          }
```
mppf: To be clear, I don't like that we have to write this kind of pattern-matching-something-lowered code, but I don't have an alternative in mind. Would it help to add something-or-other to ForLoop? Or to promote the 'getIterator' idea to its own AST node or something?

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+  USR_PRINT(ctx.forall, "the forall loop that invokes the iterator is here");
+
+  // If we continue, we may get a const violation on x*_reduceShadowVar.
+  USR_STOP();
+}
+
+
+/////////// extendYieldNew, simple case ///////////
+
+static Symbol* tupcomForSerialYield(FIcontext& ctx, int ix)
```
mppf: let's expand out whatever you meant by 'tupcom'.

compiler/resolution/implementForallIntents2.cpp
```diff
@@ -0,0 +1,1097 @@
...
+static ArgSymbol* newExtraFormal(FIcontext& ctx, FIinfo& fii, int ix,
+                                 Symbol* eActual)
+{
+    bool     isReduce   = fii.svSymbol->isReduce();
+    const char* eName   =
+      isReduce ? astrArg(ix, "reduceParent") :
+        ctx.nested ? eActual->name :
+          !strcmp(eActual->name, "_tuple_expand_tmp_") ?
+            astrArg(ix, "tet") :
+              astrArg(ix, eActual->name); // uniquify user arg name
```
mppf: Would this be clearer as a series of conditionals? Not sure we're getting much by using the ternary operator here.

test/parallel/forall/vass/par-loops/2-enclosing-par-loops-error-4.chpl
vasslitvinov: To explain my thinking:  the check tag== iterKind.standalone is not needed in this simple test. In the sense that it should not affect the test outcome or the thing being tested. I omitted this check to shorted the code.

compiler/include/ForallStmt.h
vasslitvinov: I added a comment rationalizing the current design.

Replacing fFromForLoop with a HashSet is a good idea. I would like to defer that until later.

vasslitvinov:
> I'd recommend putting everything back in to the original file for now, so that I can review this in a sane manner.

As Michael and I discussed off-PR, these changes are mostly "a complete rewrite". Therefore I did not follow this suggestion.

compiler/include/expr.h
vasslitvinov: This change is necessary because this->outerVarSE() can be NULL.

My reason for having a declaration within a conditional is that this style is quite common in our compiler. Now that I am used to it, I find it quite elegant.

compiler/include/ForallStmt.h
vasslitvinov: I renamed the second accessor to needToHandleOuterVars.

test/parallel/forall/vass/par-loops/2-enclosing-par-loops-error-4.chpl
mppf: It seems to me that we should prefer easy to understand test cases over shorter test cases. They are not the same thing. Standalone iterators normally include the where clause.

bradcray:
`implementForallIntents-2.cpp`?!?  We really couldn't come up with a more descriptive name for a file than that?  Is the intention that it will eventually replace `implementForallIntents.cpp`?

mppf:
My personal viewpoint is that we need to actually remove the old implementation. I do not expect this to be so much work that we need to put it off another year. Then we can remove the `-2` suffix.

vasslitvinov:
There used to be two phases with forall intents. 2 is for the second phase.
Like Michael says, the old implementation needs to go.

PR #7741, opened 2017-11-03
Add some quality-of-life features to perf graphs

benharsh:
0) "Invert Selection" button

1) Add "Last Two Weeks" button as shortcut to common-case date slicing.

2) Add button to clear filter box.

3) Redirect ctrl+f and meta+f to filter box.

4) Add button to each graph that resets the Y-axis zoom.

5) Preserve Y-axis zoom when annotations are enabled

6) Instead of using a tooltip to see annotation text, a box will appear
below the annotation when hovered over. This box will contain the annotation
text with strings like "(#NNN)" linked to GitHub PRs.

### DIFF
--- a/util/test/perf/perfgraph.js
+++ b/util/test/perf/perfgraph.js
@@ -232,7 +232,7 @@ function getNextDivs(afterDiv) {
   var annToggle        = addButtonHelper('annotations');
   var screenshotToggle = addButtonHelper('screenshot');
   var closeGraphToggle = addButtonHelper('close');
-  var resetY           = addButtonHelper('reset Y zoom');
+  var resetY           = addButtonHelper('reset y zoom');
 
   return {
     div: div,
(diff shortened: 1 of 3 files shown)


util/test/perf/perfgraph.js
```diff
@@ -442,18 +427,95 @@ function setupLogToggle(g, graphInfo, logToggle) {
   }
 }
 
+function computeInfoLeft(info, box, parentDiv) {
```
ronawho: Can you add a comment to this function

util/test/perf/perfgraph.js
```diff
@@ -250,6 +232,7 @@ function getNextDivs(afterDiv) {
   var annToggle        = addButtonHelper('annotations');
   var screenshotToggle = addButtonHelper('screenshot');
   var closeGraphToggle = addButtonHelper('close');
+  var resetY           = addButtonHelper('reset Y zoom');
```
ronawho: On the graphs the text looks a little off. Maybe lowercase 'Y' or do "reset y-axis zoom", but that starts to get long

util/test/perf/perfgraph.js
ronawho: (and the other new ones, just a short 1-liner explanation would be fine)

PR #7746, opened 2017-11-03
Add bfTest1.notest

vasslitvinov:
bfTest1.chpl test has a couple of issues, so do not test it
in nightly testing for now.


vasslitvinov:
@saru95 - please remove this file when you fix the known issues with bfTest1,
so we can resume testing.

PR #7747, opened 2017-11-04
Fix gasnet.darwin test errors on chapelmac since 2017-10-11

awallace-cray:
Cray I.T. updated chapelmac build slave on 2017-10-10, and since
then, gasnet.darwin fails "make test" unless the following are added:
    export GASNET_MASTERIP=127.0.0.1
    export GASNET_WORKERIP=127.0.0.0

The same setting is needed when running w gasnet from a Macbook
connected over a VPN.

See also https://chapel-lang.org/docs/1.15/platforms/udp.html#troubleshooting-the-udp-conduit


mppf:
This looks fine to me & reasonable to commit for now, but we care about what @ronawho thinks about it, when he has time to look.

Issue #7748, opened 2017-11-04
Spinwaiting in I/O lock can significantly degrade performance under qthreads


ronawho:
By default our I/O is parallel-safe and to implement the parallel-safety we use a lock. qio_lock ends up using chpl_sync_lock, which under qthreads spinwaits, which can lead to significant performance degradation for concurrent I/O on a busy machine.

The following code runs in ~.1 second on chapcs if the machine is empty, but if you run 2 concurrent instances of this test (or even just other non-chapel applications) it can take ~40 seconds or more:

```chpl
use Barriers;

config const numTasks = here.maxTaskPar;
var barrier = new Barrier(numTasks);

config const writesPerTask = 100;

coforall t in 1..numTasks {
  barrier.barrier();
  for i in 1..writesPerTask do writeln(t, ": " , i);
}
```

What ends up happening is that all tasks are spinning on an atomic, but only about half of the pthreads hosting the tasks are running for each program. So we're spinnwaiting but most of the tasks aren't making any progress, and the 1 that can will have to wait for its pthread to get rescheduled. On chapcs we essentially end up with 96 pthreads (48 per process) spinwaiting and contending with each other preventing the 1 thread that can actually grab the lock from making progress.

In order to solve this we should really do some sort of adaptive-locking, where we spin for a little while and then go to sleep.

Here's the bash script i'm using to reproduce this:

```bash
#!/usr/bin/env bash

NUM_TRIALS=${NUM_TRIALS:-20}
START_INSTANCE_TWO_TRIAL=${START_INSTANCE_TWO_TRIAL:-5}
WRITES_PER_TASK=${WRITES_PER_TASK:-100}

PRG_NAME=${PRG_NAME:-io-slowdown}
EXEC_CMD=${EXEC_CMD:-"./${PRG_NAME} --writesPerTask ${WRITES_PER_TASK}"}
EXEC_CMD2=${EXEC_CMD2:-${EXEC_CMD}}

# ensure spawned instances get killed if we ctrl-c
trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM

# Compile
echo "Compiling program"
rm -f ${PRG_NAME}
chpl ${PRG_NAME}.chpl -o ${PRG_NAME} || exit
echo "Compiled program"

# For qthreads use all cores and disable internal spinwaiting/pinning
if [ `$CHPL_HOME/util/chplenv/chpl_tasks.py` == "qthreads" ]; then
[...]

PR #7749, opened 2017-11-04
Quiet timeouts for privatization test

ronawho:
addAndGetPrivatizedInParallel has been sporadically timing out for cce whitebox
testing ever since that configuration was moved onto a busier machine. It turns
out that concurrent parallel-safe I/O uses a spinlock under qthreads, and this
test was doing a writeln on all cores, which led to timeouts when the machine
was busy. In order to quiet this test in the meantime, this just stores the
results in an array and writes out the array at the end to avoid concurrent IO

For more info see https://github.com/chapel-lang/chapel/issues/7748


ronawho:
Passes 1000 trials for the reproducer program in #7748: 

```
cp test/distributions/privatization/runtime/{PrivatizationWrappers.chpl,addAndGetPrivatizedInParallel.chpl} . && PRG_NAME=addAndGetPrivatizedInParallel EXEC_CMD=./addAndGetPrivatizedInParallel NUM_TRIALS=1000 ./concurrent-repro.sh
```

PR #7754, opened 2017-11-06
Offset module-level CHPL_RT_MD constants

benharsh:
The alloc functions in LocaleModelHelpMem were adding
`chpl_memhook_md_num` to the CHPL_RT_MD constants resulting in incorrect
memory descriptor strings in memory leaks output. Instead, pass
`CHPL_RT_MD* - chpl_memhook_md_num` to the alloc functions.

Testing:
- [x] full local
- [x] full gasnet
- [x] memleaks

### DIFF
--- a/modules/internal/CPtr.chpl
+++ b/modules/internal/CPtr.chpl
@@ -306,7 +306,14 @@ module CPtr {
   }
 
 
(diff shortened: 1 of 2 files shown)


benharsh:
@ronawho : I think you were the last person to touch these alloc functions. We have been incorrectly labeling "string copy" leaks as other things for quite a while now. This should get us back on track.

PR #7756, opened 2017-11-06
Tidy use of LLVM headers

mppf:
* Uses forward declaration in preference to header inclusion when possible.
* Makes each .cpp file include the headers it needs and removes "all the LLVM things" and "all the Clang things" type headers.
* Adds error checking to a few LLVM open file calls
* Enables the use of LLVM functions instead of 'mkdir -p' and 'rm -Rf' when they are available

Reviewed by @dmk42 - thanks!


mppf:
@dmk42 - could you review? I'd consider the "Tidy up uses of LLVM headers" commit trivial but want you to double-check the other two. Thanks.

PR #7757, opened 2017-11-06
Don't run compileline for --llvm builds

mppf:
Why? Running compileline takes ~8 seconds on some systems... and nothing else in a chpl invocation is running Perl.

How can we stop using compileline? The compileline script was supplying several different kinds of information:
 * paths to runtime libraries (such as main.o, libchpl.a)
 * C flags necessary to compile with the Chapel runtime
 * LD flags necessary to link a program using the Chapel runtime

This PR makes each of these pieces of information available to the Chapel compiler without requiring that a subprocess be spawned.

To get paths to runtime libraries, we extend `printchplenv` to print these out when it's called by the Chapel compiler. (Note that this is a little bit weird because with `--llvm` compiles there are actually *two* relevant target compilers: clang-included (i.e. llvm) and whatever we would have used with the C backend).

To get C and LD flags, this PR modifies the runtime Makefiles to simply save two files: list-includes-and-defines and list-libraries. These are saved in the lib directory next to libchpl.a. Since arguments we get from a PrgEnv depend on loaded modules, we move getting that information to the time we run a `chpl` compile.

One other change that I realized was necessary: now that the compiler correctly detects 'clang-included' as the target compiler with '--llvm' builds, we need to build a SysCTypes.chpl with the 'clang-included' compiler.

The individual commit messages include more detail on these changes.

Note that since the calls to chpl_compiler.get('target') are memoized, I needed to make that function depend on an environment variable for its default in the --llvm compilation so that the appropriate value ('clang-included') is always returned. But it also needed to be able to return the original target compiler, since we need to print that out too, so I added a 2nd argument that allows one to request the original target compiler (the one we would use with the C backend).

Note that clangLDArgs was removed from ClangInfo since this is only used in makeBinaryLLVM.

Future work:
 * CHPL_LLVM=system was broken before this PR. This PR will not correctly handle that case. PR #7759 will fix it.

- [x] full local testing
- [x] test Hello --llvm on a Cray  with CHPL_COMM=ugni
- [x] test Hello --llvm on a Cray  with CHPL_COMM=gasnet
- [x] test Hello --llvm on Mac OS X


mppf:
@dmk42 or @ronawho - could you review? This is removing the call to `compileline` which adds about 8s to --llvm compile time on some systems.

dmk42:
I'd like to make sure @ronawho agrees, but from my point of view, this looks awesome, and saving 8 seconds per compile is great.

ronawho:
I really like the idea of not running `compileline`, but there's a lot of code here and I probably won't get a chance to do a full review until Monday. chplenv changes are core enough that I would prefer you wait to merge until I get a chance to review instead of just doing a post-commit review. 

I also think we'll want @ben-albrecht to take a look at the chplenv changes as well. 

Could some of these changes like shuffling llvm headers and using llvm-tools instead of shelling out be split off into a separate PR?

mppf:
@ronawho - thanks for your feedback. Given the level of review you suggested, I'm working on tidying up the history for this commit. I'll move the header changes and new uses of LLVM tools to PR #7756.

compiler/include/driver.h
```diff
@@ -106,6 +106,8 @@ extern const char* CHPL_WIDE_POINTERS;
 extern const char* CHPL_LLVM;
 extern const char* CHPL_AUX_FILESYS;
 extern const char* CHPL_UNWIND;
+extern const char* CHPL_RUNTIME_SUBDIR;
```
ronawho: Since these are globals, I don't think we should only set them for --llvm compiles. I think we should always set them. 

I think this would also allow us to remove the `--llvm` special casing in printchplenv and some of the extra code added in `runPrintChplEnv()`

compiler/include/driver.h
```diff
@@ -106,6 +106,8 @@ extern const char* CHPL_WIDE_POINTERS;
...
+extern const char* CHPL_LAUNCHER_SUBDIR;
```
ronawho: Maybe I'm just missing something obvious, but where is `CHPL_LAUNCHER_SUBDIR` actually used?

compiler/main/driver.cpp
```diff
@@ -1157,6 +1173,11 @@ static void setupChplGlobals(const char* argv0) {
     envMap["CHPL_HOME"] = CHPL_HOME;
   }
 
+  // tell printchplenv that we're doing an LLVM build
+  if (llvmCodegen) {
+    envMap["CHPL_LLVM_CODEGEN"] = "llvm";
```
ronawho: Couldn't think just set `envMap["CHPL_TARGET_COMPILER"] = "clang-included"` and get rid of most of the chplenv changes?

ronawho:
I don't understand the code in `compiler/util/clangUtil.cpp` well enough to give that a good review, but overall this looks ok to me. If we can minimize the changes needed to chplenv, that'd be great. The interaction between various chplenv components is already pretty subtle and if we can avoid making things more complicated that'd be great.

compiler/util/clangUtil.cpp
```diff
@@ -1215,52 +1205,88 @@ bool setAlreadyConvertedExtern(ModuleSymbol* module, const char* name)
...
-    // We're parsing extern blocks AND any parts of the runtime
-    // in order to prepare for an --llvm compilation.
-    // Use compiler-specific flags for clang-included.
-    readargsfrom = compileline + " --llvm"
-                                 " --clang"
-                                 " --clang-sysroot-arguments"
-                                 " --cflags"
-                                 " --includes-and-defines";
-  }
+  const char* clang_warn[] = {"-Wall", "-Werror", "-Wpointer-arith",
```
ronawho: Seems a little weird that we're duplicating flags here instead of relying on the existing `make/compiler/Makefile.*` infrastructure, but I don't really understand this code to know if there's a good reason for this.

mppf: The whole point is to avoid shelling out to other things. The main job of the previous Perl compileline script was actually to run some Make targets so that we could use the make/compiler/Makefile.* makefiles (among other things). However that design doesn't make so much sense because there's only one compiler we're working with here - clang - and even mostly only one version - the version we bundle.

compiler/include/driver.h
mppf: No, it's not, but one day I'd like the included clang to build the launcher (vs running Make etc). I was "planning for the future". I could remove it if that's important to you.

compiler/main/driver.cpp
mppf: I got rid of many of the chplenv changes but it's important for e.g. detecting target architecture that the "original compiler" (i.e. what we would use with the C backend) is available as well as clang-included. In order to do that, we used to effectively run printchplenv twice, but by making the small change I made to it we can run it just once now.

ronawho: If we're doing an `--llvm` compile, why is the "original" target arch needed?

mppf: Because we detect Cray stuff with PrgEnv compilers, e.g. https://github.com/chapel-lang/chapel/blob/master/util/chplenv/chpl_arch.py#L286

ronawho: I'm not following. If llvm is being used, why should a user be required to have a cray-prgenv loaded? Shouldn't we update chpl_arch to check for CRAY_CPU_TARGET if a prgenv is being used or if we're on a cray and clang-included is being used?

Also can we assume that clang/llvm accepts the cray arch names?

mppf:
@ronawho, @ben-albrecht - we've documented that printchplenv --simple can create a chplconfig file. But I need to extend what the compiler gets from printchplenv... including some settings that probably don't make for chplconfigs. Should we make printchplenv --chpl or something?

PR #7758, opened 2017-11-06
Include the substrate setting in the gasnet unique config path.

### LINKED ISSUE #7733: Include gasnet substrate in install directory
Currently we only include the segment and whether debug was enabled or not. So for instance if you just change `CHPL_COMM_SUBSTRATE=udp` to `CHPL_COMM_SUBSTRATE=mpi`, we don't trigger a rebuild and you'll be left without mpi support.

gbtitus:
The substrate is part of the build-time configuration of GASNet.
Include that setting in our unique config path, so that users can
maintain side-by-side GASNet builds with different substrates and
everything else the same.

This resolves #7733.


gbtitus:
@ronawho, care to do the honors?

ronawho:
Huh.. I really thought this was going to be more work for some reason. Looks good to me!

PR #7760, opened 2017-11-07
Forwarding wrappers should propagate errors

benharsh:
Wrapper methods for forwarded fields should propagate errors, otherwise try-catch blocks will not work and the program will halt due to uncaught errors.

Testing:
- [x] full local


benharsh:
@mppf : Could you review? I'm not sure this has been discussed anywhere, but it seems like a no-brainer to me.

PR #7766, opened 2017-11-08
Switch perf playground to measure bulk-transfer-interface

benharsh:


benharsh:
@ronawho : are you done with the existing playground config?

ronawho:
All yours

PR #7770, opened 2017-11-10
correct distributedDynamic debug statements

cassella:
Two assert messages named the serial iterator instead of the iterator kind they're in.


cassella:
One .good file includes this message.   Its test is actually testing the serial iterator, so the message is correct.  That test still passes.  (``test/modules/packages/distributedIters/checkBadInput-dynamic.chpl``)

PR #7776, opened 2017-11-10
Fix arguments like type t:object

mppf:
After working with function resolution, I've realized that it's
significantly more work for the compiler to resolve a 'where' clause than
a type constraint expressed with a : on the argument. Additionally, I
know that many _cast overrides are created and resolved in the process of
normal program compilation, and these typically use where clauses. For
this reason, I was wondering if declaring arguments like 'type t:object'
or 'type t:integral' would work.

So, I tried a few cases and to my surprise most of them were working
already! I only needed to make a minor modification to allow it in the
remaining cases. (In particular, only the type t:someclasstype were not
working).

For example, this PR enables functions like this:
``` chapel
proc foo(type t:object) {
  writeln("object");
}
```
The fix amounts to adjusting needToAddCoercion to ignore coercing between type variables.

This pattern does not seem to be used much (if at all) in our code base
currently, so I added a test to exercise it.

Passed full local testing.

Future work: Replace some where clauses with the `type t:sometype`
argument declarations to reduce the work the compiler must do.

### DIFF
--- a/compiler/resolution/wrappers.cpp
+++ b/compiler/resolution/wrappers.cpp
@@ -780,6 +780,11 @@ static bool needToAddCoercion(Type*      actualType,
              (getIntent(formal) & INTENT_FLAG_REF) != 0) {
     retval = false;
 
+  // If actual and formal are type symbols, no coercion is necessary
+  } else if (actualSym->hasFlag(FLAG_TYPE_VARIABLE) &&
+             formal->hasFlag(FLAG_TYPE_VARIABLE)) {
+    retval = false;
+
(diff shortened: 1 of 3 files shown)


mppf:
@vasslitvinov - would you mind reviewing this PR?

vasslitvinov:
Please discuss in the commit message whether we should go ahead and replace some where clauses with the new `type t: appropriate type` form.

bradcray:
Does this need a spec update?

mppf:
@bradcray - arguably. I drafted one in PR #7778 and would be interested in feedback on it.

PR #7778, opened 2017-11-11
Proposed specification changes for type t:someType arguments

mppf:
PR  #7776 adjusts the compiler to allow arguments like `type t:object`. Many other cases were already working.

However the specification does not explicitly say whether or not this is legal. This PR modifies the specification to assert that `type t:someType` arguments are legal and to show an example of them. Additionally it adjusts one of the `param` examples to show that the type for a `param` argument can be omitted.

### DIFF
--- a/spec/Generics.tex
+++ b/spec/Generics.tex
@@ -74,10 +74,7 @@ accepting type arguments that only apply to a specific group of types.
 Suppose that we'd like to define a function that accepts a type argument
 and returns 1 represented in that type.
 \begin{chapel}
-proc getOne(type t:real) {
-  return 1.0;
-}
-proc getOne(type t:integral) {
+proc getOne(type t:numeric) {
   return 1:t;
 }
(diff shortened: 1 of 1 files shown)


spec/Generics.tex
```diff
@@ -65,6 +66,38 @@ \subsection{Formal Type Arguments}
...
 
+A formal type argument can include a formal type (a colon followed by a
+type). This pattern is sometimes useful to create generic functions
+accepting type arguments that only apply to a specific group of types.
+
+\begin{chapelexample}{typeColonArgument.chpl}
+Suppose that we'd like to define a function that accepts a type argument
+and returns 1 represented in that type.
+\begin{chapel}
+proc getOne(type t:real) {
```
daviditen: A single function might be a better example here:

```
proc getOne(type t: numeric) {
  return 1: t;
}
```

PR #7780, opened 2017-11-11
Don't set argument type to dtAny for type variables with type exprs

mppf:
e.g. proc foo(type t:int(8)) should not have type dtAny but rather type dtUnknown that is resolved later. This allows the determination about whether or not that type is generic to be made later in compilation and enables more 'type t:someType' argument patterns to work.

- [x] full local testing
- [x] test/functions/ferguson with --baseline, --verify, --baseline --verify


mppf:
@vasslitvinov - could you have a look at this one as well? thanks.

mppf:
@vasslitvinov - could you have a look at this?

PR #7785, opened 2017-11-13
Improve parse-time error messages

mppf:
This PR improves the error messages the compiler emits during parse time. In particular, it allows the parser to emit multiple syntax errors and it provides a more useful error for the common beginner programming think-o of using `=` in a conditional.

Multiple errors: This PR adjusts chapel.ypp to use the Bison error recovery mechanism, which amounts to a built-in nonterminal called `error`. To do so, it adds a few `error` rules at some high-level places and also removes a few old error messages (like "bad type specification") that do not appear to be tested.

Checking for `=` in a conditional: adds new nonterminals for if statement parsing to match this case. Adds the helper function in build.cpp convertAssignmentAndWarn to handle this case in a resume-able way.

Other changes: Now that yyerror can return, I needed to slightly adjust control flow in eatStringLiteral in chapel.lex.

- [x] full local testing


compiler/AST/build.cpp
```diff
@@ -3130,3 +3133,15 @@ Expr* tryBangExpr(Expr* e)
 {
   return new CallExpr(PRIM_TRYBANG_EXPR, e);
 }
+
+Expr* convertAssignmentAndWarn(Expr* a, const char* op, Expr* b)
+{
+  if (0 == strcmp("=", op)) {
+    USR_FATAL_CONT(a, "Use == to check for equality in a conditional");
```
vasslitvinov: I think the error message needs to be stronger, by saying "assignment is illegal in a conditional".
It, then, can go on to say what it says now.

compiler/AST/build.cpp
```diff
@@ -3130,3 +3133,15 @@ Expr* tryBangExpr(Expr* e)
...
+  } else {
+    USR_FATAL_CONT(a, "Invalid conditional %s", op);
```
vasslitvinov: Makes me wonder what Chapel code can lead to this case. Could you discuss this in a comment here (ok if briefly) or add a test?

compiler/parser/chapel.ypp
```diff
@@ -235,10 +236,14 @@
...
                const char*    str) {
+
+    // like USR_FATAL_CONT
+    setupError(__FILE__, __LINE__, 3);
+
     // TODO -- should this begin with error:
     if (!chplParseString) {
       const char* yyText = yyget_text(context->scanner);
-
+
```
vasslitvinov: Would be nice to remove the space and eliminate this diff line.

compiler/parser/chapel.ypp
```diff
@@ -266,6 +269,9 @@
 
 %debug
 %verbose
+// Thes would print out expected token names
```
vasslitvinov: "Thes" -> "These" ?
Otherwise Paul will come after you with the spell checker. :)

compiler/AST/build.cpp
mppf: if a += b then
I'll add a test

PR #7789, opened 2017-11-13
Convert to more tests to use initializers

lydia-duncan:
These tests were accidentally overlooked in my initial conversion to
use initializers instead of constructors.  Some remain that still have
issues, but these were all capable of being converted.

In more detail:
test/studies/comd/elegant/arrayOfStructs/CoMD.chpl's helper file,
test/studies/comd/elegant/arrayOfStructs/util/AccumStencilDist.chpl,
saw the constructor for LocAccumStencil updated to an initializer.  The
other constructor in that file, for AccumStencil, could not yet be updated
because the type inherits from a library type.  This initializer would benefit
from Phase 1 being the default, and from the ability to define generic and
const fields in Phase 2.

test/studies/comd/llnl/CoMD.chpl's helper types InterpolationObject,
ForceLJ, and Ticker types have been updated to use initializers.  Ticker
would benefit from Phase 1 being the default.  I was unable to update
ForceEAM due to the presence of not yet supported error handling
statements in its body.

test/users/npadmana/bugs/replicated_invalid_ref_return/replicated_bug.chpl
was trivially updated to an initializer.

test/users/shetag/fock/fock-orig.chpl's blockIndices class was updated to
use an initializer.  This initializer would benefit from Phase 1 as the default
and the ability to set const fields during Phase 2.

test/users/shetag/fock/blockindices-blc.chpl, the helper file for
test/users/shetag/fock/fock-3-blc.chpl, test/users/shetag/fock/fock-stc-1.chpl,
test/users/shetag/fock/fock-stc.chpl,
test/users/shetag/fock/fock-dyn-prog-pool.chpl, and
test/users/shetag/fock/fock-dyn-prog-cntr.chpl, was updated.  It would also
benefit from Phase 1 as the default and the ability to set const fields during
Phase 2.

test/users/shetag/fock/fock-1-begin-blc.chpl would also benefit from Phase
1 as the default and the ability to set const fields during Phase 2.

test/users/thom/arrInClass-tuple.chpl would benefit from Phase 1 as the
default.

test/users/thom/arrInClass-dommember.chpl similarly would benefit from
Phase 1 as the default.

The helper file of test/users/thom/returnGenericLineno/strassen_main.chpl,
test/users/thom/returnGenericLineno/rosslyn.chpl, would also benefit from
Phase 1 as the default.

Passed a full std/ paratest

### DIFF
--- a/test/studies/comd/elegant/arrayOfStructs/util/AccumStencilDist.chpl
+++ b/test/studies/comd/elegant/arrayOfStructs/util/AccumStencilDist.chpl
@@ -453,11 +453,13 @@ proc AccumStencil.dsiCreateReindexDist(newSpace, oldSpace) {
 }
 
 
-proc LocAccumStencil.LocAccumStencil(param rank: int,
-                      type idxType,
-                      locid, // the locale index from the target domain
-                      boundingBox: rank*range(idxType),
-                      targetLocBox: rank*range) {
+proc LocAccumStencil.init(param rank: int,
+                          type idxType,
+                          locid, // the locale index from the target domain
+                          boundingBox: rank*range(idxType),
+                          targetLocBox: rank*range) {
+  this.rank = rank;
+  this.idxType = idxType;
   if rank == 1 {
     const lo = boundingBox(1).low;
     const hi = boundingBox(1).high;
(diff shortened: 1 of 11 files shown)


lydia-duncan:
@benharsh - is it alright for me to be updating the CoMD helpers like this?
@noakesmichael - would you mind giving this a review?

PR #7794, opened 2017-11-14
Fix build issues for systems that have cuda installed

ronawho:
Hwloc probes for cuda and if cuda is found, hwloc must be linked with
`-lcudart`. However, when qthreads is checking if hwloc works it only throws
`-lhwloc` or `-lhwloc -lnuma`, so the link step fails. Note that we build hwloc
and qthreads statically, for .so builds this wouldn't be an issue.

For now just throw `--disable-cuda` to hwloc to avoid the issue. The "right"
solution is probably for qthreads to use libtool or pkg-config to figure out
hwloc's link args: https://github.com/Qthreads/qthreads/issues/56

### DIFF
--- a/third-party/hwloc/Makefile
+++ b/third-party/hwloc/Makefile
@@ -27,6 +27,7 @@ CHPL_HWLOC_CFG_OPTIONS += --enable-static \
                           --disable-cairo \
                           --disable-libxml2 \
                           --disable-libudev \
+                          --disable-cuda \
                           --disable-opencl \
                           --disable-pci
 


ronawho:
@mppf ran into this on an external system, but I was able to repro on chapcs08 (one of our systems with cuda available) if I built with the following set:

```sh
export LIBRARY_PATH=/usr/local/cuda-7.0/lib64:$LIBRARY_PATH
export LD_LIBRARY_PATH=/usr/local/cuda-7.0/lib64:$LD_LIBRARY_PATH
export PATH=/usr/local/cuda-7.0/bin/:$PATH
export CPATH=/usr/local/cuda-7.0/include/:$CPATH
```

ronawho:
Here's an alternative (but ugly) solution: 

```diff
diff --git a/third-party/qthread/Makefile b/third-party/qthread/Makefile
index e7cfbf7..6f8feff 100644
--- a/third-party/qthread/Makefile
+++ b/third-party/qthread/Makefile
@@ -26,6 +26,8 @@ endif

 ifeq ($(CHPL_MAKE_HWLOC),hwloc)
 CHPL_QTHREAD_CFG_OPTIONS += --with-topology=hwloc --with-hwloc=$(HWLOC_INSTALL_DIR)
+RUNTIME_LFLAGS += -L$(HWLOC_LIB_DIR)
+RUNTIME_LFLAGS += $(shell python -c "import os, sys; sys.path.insert(0, os.path.join('$(CHPL_MAKE_HOME)', 'util', 'chplenv')); from chpl_3p_hwloc_configs import get_link_args; sys.stdout.write(' '.join(get_link_args('hwloc')))")
 endif
 ifeq ($(CHPL_MAKE_HWLOC),system)
 CHPL_QTHREAD_CFG_OPTIONS += --with-topology=hwloc
```

It adds hwloc link args to the qthreads build via `chpl_3p_hwloc_configs.get_link_args('hwloc')`

gbtitus:
This looks fine at least as a short-term solution while we pursue a more principled one.  👍

Issue #7795, opened 2017-11-15
Our execution limiter is racey


ronawho:
chpl executables like to monopolize a machine (especially when qthreads is being used.) Because of this, concurrent chapel executables often don't play well together, which can cause timeouts. https://github.com/chapel-lang/chapel/pull/3928 added an "execution_limiter" to try to prevent multiple executables from running at the same time. It used a file as a lock in racey manner, but we didn't think the race would be that big of a deal. However, recently we've seen timeouts for some whitebox and llvm configurations that are using paratest in heavily a oversubscribed manner and in all cases it seems to be because multiple chpl executables are running, meaning our race is hurting us. 

Here's (internal) links to the whitebox/llvm timeouts: 
 - https://chapel.atlassian.net/browse/CHAPEL-302 
 - https://chapel.atlassian.net/browse/CHAPEL-318

Here's a script I'm using demonstrate the race:

```bash
#!/usr/bin/env bash

file="no-op"
sleep_amt=.05

# use ps to look concurrent instances of the test
check_for_duplicates() {
  while true; do
    # find instances of the executable (ignoring commands that are concurrent
    # compiles, this grep, the timedexec wrapper, and the test diff)
    case "$(uname)" in
     CYGWIN*) myps=procps ;;
        *)    myps=ps     ;;
    esac
    dupes=$(${myps} x | grep ${file} | grep -v chpl | grep -v grep | grep -v timedexec | grep -v diff)
    dupes_count=$(echo "${dupes}" | wc -l)
    if [ ${dupes_count} -gt 1 ]; then
      echo "${dupes}"
      echo ""
      sleep ${sleep_amt}
    fi
  done
}

# Create a bunch of dirs with a single test that sleeps for a short amount of
# time. Allows us to run a lot of concurrent paratests and catch if executables
# are running at the same time
cd ${CHPL_HOME}/test
nproc=$(python -c "import multiprocessing; print(multiprocessing.cpu_count())")
for i in `seq ${nproc}`; do
  dir="tmp/dir${i}"

  rm -rf ${dir}
  mkdir -p ${dir}

  echo "use Time; sleep(${sleep_amt});" > "${dir}/${file}.chpl"
  touch "${dir}/${file}.good"
done

check_for_duplicates &
# ensure we kill the check_for_dupes subprocess
trap "pkill -P $$" SIGINT SIGTERM EXIT


[...]

PR #3928, opened 2016-05-28
Add ability to limit how many tests will execute concurrently

ronawho:
Paratest has the ability to run oversubscribed with the `-nodepara` option.
This essentially runs multiple start_tests per machine, which means multiple
compiles/executions can happen simultaneously. However, some chapel programs
(particularly when using qthreads) can use up a significant portion of the
machine, which can lead to timeouts when multiple heavy-weight tests are
running concurrently.

This adds the ability to limit our testing infrastructure so that only one
executable can run at a time. Compilation is single threaded, so this doesn't
limit how many compiles occur and it still lets compilation occur while other
tests are running.

In order to enable this just set `CHPL_TEST_LIMIT_RUNNING_EXECUTABLES=yes`

In order to limit how many executables run at once, we basically use a file as
a mutex. This is race-y, which means occasionally two or more executables might
be running at the same time. It's pretty unlikely this will happen though,
since there's small window for that to happen. And note that the code is
careful to consider cases where paratest is killed and doesn't remove the lock,
or where multiple executables end up running concurrently.

The simplified version of what this code does is:

```python
  while os.path.exists('/tmp/<username>-test-running') and modified < timeout:
      time.sleep(.02)

  # race-y since other process could have created file since while loop
  open('/tmp/<username>-test-running', 'w')

  run_executable()

  os.remove('/tmp/<username>-test-running')
```

Note that arguably this is a hacky/ugly solution, and that really I should
have updated sub_test to use python's multiprocessing to get concurrent
compiles instead of just abusing paratest. However, that would have probably
taken me a few days, and this only took a few hours.

PR #7799, opened 2017-11-15
Make our execution limiter race-free

ronawho:
Use filelock (https://pypi.python.org/pypi/filelock) to make our execution
limiter race free in order to avoid timeouts under llvm/whitebox testing.

PR #3928 added an "execution_limiter" to try to prevent multiple executables
from running at the same time. It used a file as a lock in racey manner, but we
didn't think the race would be that big of a deal. However, recently we've seen
timeouts for some whitebox and llvm configurations that are using paratest in
heavily a oversubscribed manner and in all cases it seems to be because
multiple chpl executables are running, meaning the race is hurting us.

Our old implementation used a soft file-based lock, whereas filelock will use a
hard lock when available. filelock is a platform independent lock that uses 1
of 3 platform dependent implementations. It will use `fcntl.flock` on unix
systems (including mac and cygwin) to create a hard lock, `msvcrt.locking` on
windows systems, and a soft file-based lock otherwise. fcntl and msvcrt don't
have issues with 'stale' locks because the lock is held on the file descriptor
and gets dropped by the OS even if the program crashes. The soft file-based
lock can cause stale locks, so if we ever run into a platform where
fcntl/msvcrt aren't supported we should re-add our own implementation that can
deal with stale lock files.

Also note that fcntl/msvcrt will leave the actual file used for locking around.
Removing the file would "unlock" the lock, so there'd be a race between locking
and unlocking/removing (we'd effectively have a double unlock). This is a
common problem for file-based locks. Since we're only using a lockfile per
user, this shouldn't be a big deal, but if we start using the execution limiter
for more things we might need to find a different solution.

This should resolve https://github.com/chapel-lang/chapel/issues/7795

### DIFF
--- a/third-party/chpl-venv/test-requirements.txt
+++ b/third-party/chpl-venv/test-requirements.txt
@@ -1,3 +1,4 @@
 subprocess32==3.2.6
 argparse==1.3.0
 PyYAML==3.11
+filelock==2.0.13
--- a/util/test/execution_limiter.py
+++ b/util/test/execution_limiter.py
@@ -1,107 +1,49 @@
-from __future__ import with_statement
+""" Execution limiter implementations to help limit the number of concurrently
+running chpl executables """
 
 import getpass
 import os
-import sys
 import tempfile
-import time
+import filelock
 
-def log_info(msg):
-    if os.getenv('CHPL_TEST_DEBUG_RUNNING_EXECUTABLE_LOCK') is not None:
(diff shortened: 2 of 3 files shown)


ronawho:
Passed 5,000 trials of the reproducer in #7795 on mac, darwin, and linux64

ronawho:
@benharsh mind reviewing?

@ben-albrecht might be interesting in reviewing as well since he reviewed the original #3928, but I'm think he'll be busy until next week

PR #7801, opened 2017-11-15
Store paths saved for --llvm relative to CHPL_HOME etc

mppf:
PR #7757 started saving CC and LD flags in files to avoid a lot of
computation at chpl-time. But these flags contained some paths that
may not exist after installation. To fix, normalize these paths
to use e.g. $CHPL_HOME, and then expand them again in the compiler.

- [x] verified 'make install'd chpl works after source directory moved

### DIFF
--- a/compiler/util/files.cpp
+++ b/compiler/util/files.cpp
@@ -823,7 +823,7 @@ void expandInstallationPaths(std::vector<std::string>& args) {
 
       size_t off = 0;
       while (true) {
-        off = s.find(tofix[j], off);
+        off = s.find(key, off);
         if (off == std::string::npos)
           break; // no more occurrences to replace
         s.replace(off, key_len, val);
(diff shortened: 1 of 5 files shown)


compiler/util/files.cpp
```diff
@@ -802,6 +802,39 @@ void readArgsFromFile(std::string path, std::vector<std::string>& args) {
...
+    for (int j = 0; tofix[j] != NULL; j += 2) {
+
+      const char* key = tofix[j];
+      const char* val = tofix[j+1];
+      size_t key_len = strlen(key);
+      size_t val_len = strlen(val);
+
+      size_t off = 0;
+      while (true) {
+        off = s.find(tofix[j], off);
```
dmk42: `tofix[j]` should be `key` here in case the implementation up above is ever changed.

ronawho:
Unfortunate that we need this much code for this, but looks ok to me. 

FYI we do someting similiar for getting the path to third-party stuff in util/chplenv/third_party_utils.py:handle_la()

PR #7807, opened 2017-11-16
Prevent signed/unsigned comparison errors

dmk42:
Some compilers will generate a warning for comparing signed and unsigned values, which we turn into an error with \-Werror.  In particular, we have seen this happen with gcc 4.7.  This change adds explicit casts to quiet the warnings introduced by #7659 .

Verified not to introduce any new errors into a full local test.


dmk42:
Thanks, @vasslitvinov !

mppf:
Looks good, thanks!

PR #7809, opened 2017-11-16
Fix typos

cassella:
Fix typos in clangUtil.cpp, implementForallIntents2.cpp, functionResolution.cpp, normalize.cpp, and foralls.cpp.


cassella:
The "Unspported" message does not appear under ``tests/``.

I'm not 100% sure about changing the second "don't coerce bools to reals" to "don't coerce bools to complexes".  Nor did I check that the spec can be cited for the modified comment.

daviditen:
In section 9.1.1 of the spec there is a rationale section that states:

> We disallow implicit conversion of a boolean to a real, imaginary, or complex type because of the following. We expect that the cases where such a conversion is needed will more likely be unintended by the programmer. Marking those cases as errors will draw the programmer’s attention. If such a conversion is actually desired, a cast §9.2 can be inserted.

These changes all look good to me.

PR #7811, opened 2017-11-16
Adjusted .bad files for #7785

vasslitvinov:
With #7785, the compiler is now capable of reporting multiple syntax errors in a single run.

This PR adjusts the .bad files that are affected by this change because they contain multiple syntax errors according to the currently-implemented syntax.


mppf:
Great, thanks!

PR #7815, opened 2017-11-16
printchplenv --debug no longer print stuff needed for --llvm

mppf:
This PR adjusts printchplenv to avoid adding the new variables needed for --llvm compiles - CHPL_ORIG_TARGET_COMPILER, CHPL_MAKE_RUNTIME_SUBDIR, and CHPL_MAKE_LAUNCHER_SUBDIR - to `printchplenv --debug`. These were added in PR #7757.

 Additionally, we need to resolve the mismatch introduced in PR . The [documentation](https://chapel-lang.org/docs/master/usingchapel/chplenv.html#generating-configuration-files) says to use `printchplenv --simple` to save the current configuration, but now that prints out CHPL_ORIG_TARGET_COMPILER, CHPL_MAKE_RUNTIME_SUBDIR, and CHPL_MAKE_LAUNCHER_SUBDIR which we probably don't want to end up in chplconfig files.

- [ ] full local --llvm testing
- [ ] full local testing

### DIFF
--- a/util/printchplenv
+++ b/util/printchplenv
@@ -76,7 +76,7 @@ def print_mode(mode='list', anonymize=False):
 
         print_var('CHPL_TARGET_PLATFORM', target_platform, mode)
         print_var('CHPL_TARGET_COMPILER', target_compiler, mode)
-        if mode == 'simple' or mode == 'make' or mode == 'debug':
+        if mode == 'simple' or mode == 'make':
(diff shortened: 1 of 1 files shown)


mppf:
@ben-albrecht - could you discuss this with me? thanks.

mppf:
Replaced by #7976.

PR #7816, opened 2017-11-16
Fix filelock import on cle6 machines

ronawho:
sub_test/execution_limiter are run as a subprocess of start_test. Depending on
the level of support subprocess32 was built with, subprocesses may or may not
inherit the venv activated in start_test, so to be safe re-activate it in the
execution_limiter before importing filelock.

Our module gets built on a cle5 machine and builds a python 2.6 venv with the
system python and a python 2.7 venv with a custom installed python. Something
about this causes the 2.7 subprocesses32 to not have its full functionality and
as a result spawned processes don't inherit the venv (at least that's my best
guess as to what's going on.)

### DIFF
--- a/util/test/execution_limiter.py
+++ b/util/test/execution_limiter.py
@@ -5,8 +5,11 @@ import getpass
 import os
 import tempfile
 
-import activate_chpl_test_venv
-import filelock
+try:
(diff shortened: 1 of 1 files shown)


ronawho:
@benharsh mind reviewing?

PR #7824, opened 2017-11-17
Exit if build fails for `mason run --build`

ben-albrecht:
Exit if build fails for `mason run --build`.


ben-albrecht:
Initially pointed out by @nspark in #7106

PR #7826, opened 2017-11-17
Introduce SET_YYLINENO

vasslitvinov:
I need to do SET_LINENO during parsing to get a correct line number
for error reporting. SET_LINENO per se does not work during parsing
because BaseAST::BaseAST gives precedence to (yystartlineno, yyfilename)
over currentAstLoc.

To enable that, I added SET_YYLINENO macro.
It behaves just like SET_LINENO, except it works with
(yystartlineno, yyfilename) - so it is usable during parsing.

This allows me to remove some less-than-ideal code in foralls.cpp.

Replaces #7820.

### DIFF
--- a/compiler/AST/baseAST.cpp
+++ b/compiler/AST/baseAST.cpp
@@ -747,7 +747,7 @@ bool isCForLoop(const BaseAST* a)
 
 /************************************* | **************************************
 *                                                                             *
-* Definitions for astlocMarker                                                *
+* Definitions for astlocMarker and astlocYyMarker                             *
 *                                                                             *
 ************************************** | *************************************/
 
@@ -770,4 +770,28 @@ astlocMarker::astlocMarker(int lineno, const char* filename)
 // destructor, invoked upon leaving SET_LINENO's scope
 astlocMarker::~astlocMarker() {
   currentAstLoc = previousAstLoc;
+}
+
+// constructor, invoked upon SET_YYLINENO
+astlocYyMarker::astlocYyMarker(BaseAST* ast) :
+  prevYystartlineno (yystartlineno),
+  prevYyfilename    (yyfilename)
+{
+  yystartlineno = ast->linenum();
+  yyfilename    = ast->fname();
+}
+
+// constructor, for special occasions
+astlocYyMarker::astlocYyMarker(int lineno, const char* filename) :
+  prevYystartlineno (yystartlineno),
+  prevYyfilename    (yyfilename)
+{
+  yystartlineno = lineno;
+  yyfilename    = filename;
+}
+
+// destructor, invoked upon leaving SET_YYLINENO's scope
+astlocYyMarker::~astlocYyMarker() {
+  yystartlineno = prevYystartlineno;
+  yyfilename    = prevYyfilename;
 }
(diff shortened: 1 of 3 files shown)


mppf:
I don't like the idea of changing yystartlineno and yyfilename. These seem to me to be parser variables that we shouldn't mutate. Is there another way to adjust these macros? Could we change BaseAST::BaseAST to no longer "give precedence" to the parser position in some situation? E.g. we could make SET_LINENO_DURING_PARSING that sets currentParsingAstLoc, which is normally unset, and which would have precedence over yyfilename etc as well as currentAstLoc?

Or, what would happen if BaseAST just preferred currentAstLoc over yyfilename etc? Have you run testing in that mode, to try to understand why it is written this way? If that passed testing I'd expect it to be a better strategy.

noakesmichael:
I think it is easy to suggest that it would seem to be a
"mistake" to modify internal state of flex/bison, even in
a relatively controlled way, while flex/bison is scanning
Chapel text.


However it's hard to know what to suggest as an alternative
without a better understanding of what the essential problem is.

Is this really a new use-case for parsing that we haven't seen
before or is this something that would be better handled in a
different way?

Is the C-code in question trying to augment a bison production
as a work-around for challenges with the bison grammar, or is
it trying to implement transformations that could be done equally
well at some later point?

vasslitvinov:
I tried swapping the order of yy* and currentAstLoc as suggested above by @mppf.

I am including that change in this PR as commit 7cfcdfa .

Alas it causes these failures - seems only 2-3 failure modes, all of which are in compiler error messages:

errhandling/parallel/strict/coforall-throws-strict-error.chpl
functions/ferguson/ref-pair/const-error-iterated2.chpl
functions/ferguson/ref-pair/const-error-iterated.chpl
functions/iterators/diten/zipCoforallIter.chpl
parallel/coforall/bradc/nested/nestedForall2.chpl
parallel/taskPar/taskIntents/01-error-global.chpl
parallel/taskPar/taskIntents/02-error-local.chpl
parallel/taskPar/taskIntents/03-error-in-fun.chpl
parallel/taskPar/taskIntents/04-error-in-var-iter.chpl
parallel/taskPar/vass/const-in-intent-arrays-domains.chpl
types/enum/diten/badConfigParamEnum.chpl

These could be easy to fix, in which case we will not need SET_YYLINENO and it is a nicer arrangement in any case. I am not planning to do so at the moment, however.

vasslitvinov:
I was tickled by @noakesmichael comment above, suggesting that there is a design glitch in our parser production(s). I found and fixed it in #7835, which obviates the need for SET_YYLINENO that motivated this PR.

I am keeping this PR open for now as we might still want SET_YYLINENO or a similar facility for other uses in the future.

mppf:
> I am keeping this PR open for now as we might still want SET_YYLINENO or a similar facility for other uses in the future.

@vasslitvinov - if that's the indefinite future, you should closes this PR. It will still be there. I think it'd be OK to leave it open for a week (say) but after that it shouldn't be on the list.

noakesmichael:
I don’t see any value in leaving it open for a week
and then closing it.  Better to simply close it now.


On Nov 21, 2017, at 8:09 AM, Michael Ferguson <notifications@github.com<mailto:notifications@github.com>> wrote:


I am keeping this PR open for now as we might still want SET_YYLINENO or a similar facility for other uses in the future.

@vasslitvinov<https://github.com/vasslitvinov> - if that's the indefinite future, you should closes this PR. It will still be there. I think it'd be OK to leave it open for a week (say) but after that it shouldn't be on the list.

vasslitvinov:
We do not expect to act on this PR at this point. Closing it.

Issue #7728, opened 2017-11-02
Calling clear() on associative domain does not reset associative array values


benharsh:
### Summary of Problem
Calling ``clear()`` on an associative domain does not reset the corresponding values in related associative arrays to their default value. This is surprising as ``remove(idx)`` will reset the value.

**Associated Future Test(s):**
[`test/domains/sungeun/assoc/clear.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/domains/sungeun/assoc/clear.chpl
          )

bradcray:
This seems related to my recently merged PR #7555 and seems as though it may be low-hanging fruit?

daviditen:
Fixed by clearing array elements as indices are added to the domain in PR #7828.

PR #7828, opened 2017-11-17
Clear associative array elements when they are created by dsiAdd

daviditen:
In the dsiAdd method on associative domains walk the arrays associated with
the domain and default initialize all of the newly added array elements.

This fixes test/domains/sungeun/assoc/clear.future, issue #7728.


daviditen:
Thanks @benharsh!

PR #7834, opened 2017-11-18
Added FnSymbol::singleInvocation()

vasslitvinov:
singleInvocation() returns the CallExpr that is the single call
to 'this' FnSymbol. If there is no such call, return NULL.

There are a couple of places in the compiler where I would like
to assert that the call I am looking at is a single invocation.
singleInvocation() allows me to do that cheaply,
thanks to firstSymExpr() and lastSymExpr().

This PR also uses singleInvocation() to add an assertion in
implementForallIntents2.cpp and to replace an expensive check
in iterator.cpp.

While there, removed unused fields of eflopiInfo.

Testing: linux64 --verify -futures

### DIFF
--- a/compiler/AST/iterator.cpp
+++ b/compiler/AST/iterator.cpp
@@ -1160,25 +1160,6 @@ addAllLocalVariables(Vec<Symbol*>& syms, Vec<BaseAST*>& asts) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


vasslitvinov:
@mppf or @noakesmichael or @benharsh - would you review?

PR #7835, opened 2017-11-18
Cleanup for addForallIntent(), addTaskIntent()

vasslitvinov:
Inspired by Mike's comment in #7826, I switched the parser to produce
a ShadowVarSymbol instead of the struct IntentExpr,
when parsing each 'intent_expr" of a where-clause.

This necessitated some cleanup in related code.
As a side effect, the "intent XXX is not supported in a where-clause"
error is now reported in the parser.

The primary shortcoming of the old scheme is that we still needed to
generate a ShadowVarSymbol() for an IntentExpr, if the where-clause
is for a forall loop. By the time we run "new ShadowVarSymbol()",
however, the parser's yystartlineno points to the start of the where-
clause, rather than to the current 'intent_expr'. So I had to patch
the resulting nodes' astlocs manually, see the addForallIntent() change
in #7823.

- [x] paratest linux64 --verify -futures
- [ ] commit updates to parser files

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -1514,20 +1514,17 @@ buildForallLoopStmt(Expr*      indices,
 }
 
 // Todo: replace with ForallIntents or similar.
-void addTaskIntent(CallExpr* ti, Expr* var, IntentTag intent, Expr* ri) {
-  if (ri) {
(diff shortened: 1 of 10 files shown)


vasslitvinov:
@mppf or @noakesmichael - would you review?

Issue #6758, opened 2017-07-19
Improve Travis CI testing


ben-albrecht:
This issue tracks improvements we'd like to make to pre-merge testing (Travis CI), which executes the test script: `util/buildRelease/smokeTest`

- [ ] `make docs` support
- [ ] `.futures` tracking (#4964)
- [ ] Awareness of tests modified in PR (#4951)
- [ ] `annotate.py` support (#4998)
- [ ] `chplspell` support (#6731)

ben-albrecht:
A small correction to my earlier estimate -- because we already build `chpldoc` in the Travis CI testing, building the docs would add an additional ~30 seconds to the testing time.

ronawho:
I'd move `Awareness of tests modified in PR (#4951)`  into a new issue. I think that's a little bigger, and if a lot of tests change, we wouldn't want travis to chew on that forever

I'd also be in support of adding a python linter as well as running `util/devel/lookForBadAlloc` (#6163)

ben-albrecht:
When discussing the addition of adding `make docs` to the testing, some suggested adding it as a separate Travis CI job that runs in parallel. Because `chpldoc` requires a built compiler, a separate job for `chpldoc` and `make docs` would not save a significant amount of time. 

I think this optimization may be useful for us in the future, especially as we beef up Travis CI testing. However, I don't think it's necessary for adding `make docs`, or any of the other changes listed in this issue (assuming they run relatively quickly).

For reference, `make docs` would add ~30 seconds of build time to the current build time, and current build times fluctuate between [2-4 minutes](https://travis-ci.org/chapel-lang/chapel/builds).

ronawho:
Running `make docs` as a concurrent travis build with https://github.com/chapel-lang/chapel/pull/6921

We weren't building chpldoc as part of the travis testing before, so doing `make docs` would have added 90-120 seconds to the travis build.

ronawho:
Added a link to an issue for running the .graph/GRAPHFILES checks that we do as part of start_test -- https://github.com/chapel-lang/chapel/issues/6422

lydia-duncan:
Low priority: added a line about make the configurations more visually distinct, because looking at this isn't helpful when I just want to know what the build is that is still running - 
![screenshot from 2017-08-09 15 08 07](https://user-images.githubusercontent.com/2454710/29146272-3c45286c-7d15-11e7-861a-584114f2ab78.png)

ronawho:
[...]

PR #7837, opened 2017-11-20
Clean up tabs & whitespace in Mason source code

ben-albrecht:
I don't want these changes to pollute more meaningful commits in upcoming changes.


ben-albrecht:
Just another motivating data point for adding CI support in Travis to catch and/or fix these cases (#6758) in the future.

PR #7839, opened 2017-11-20
Simplify getting yielded type for recursive standalone

mppf:
A variety of functions implementing forall intents end up calling buildIterYieldType in order to handle recursive parallel iterators. The previous implementation used a map to save the original yielded type but that map isn't really necessary. This PR uses a simpler solution that amounts to passing the resolved function to several pieces of the implementation.

- [x] passed test/modules/standard/FileSystem/filerator/bradc/ with --baseline, --verify, --baseline --verify


mppf:
@vasslitvinov - here is that change, I'm interested to know if it's obvious to you how to address test/functions/iterators/diten/iteratorNoYield.chpl which fails due to an assert in fsIterYieldType.

vasslitvinov:
@mppf - I have dealt with that test before, I will look.

PR #7848, opened 2017-11-21
Add checks for trailing spaces to smokeTest

daviditen:
Add checks for trailing spaces in compiler/runtime/modules to the smokeTest
script.  These checks are currently commented out because a lot of files
have lines with trailing spaces. Once those are cleaned up we can uncomment
the new code in smokeTest.

I verified that the new test is working by cleaning up the trailing spaces
across compiler/runtime/modules and running smokeTest with the new code
uncommented.

The trailing spaces can be cleaned up with the command:

```
for f in `util/devel/lookForTrailingSpaces | sed -e 's/:.*$//' | sort | uniq` ; do
  sed -i -e 's/  *$//' $f
  rm ${f}-e
  git add $f
done
```

### DIFF
--- a/util/buildRelease/smokeTest
+++ b/util/buildRelease/smokeTest
@@ -45,6 +45,15 @@ if (( $num_tabs > 0 )) ; then
     exit 1
 fi
 
+# Need to clean up all of the trailing spaces before adding this
+# Check for trailing spaces
+#num_trailing_space_lines=$($CHPL_HOME/util/devel/lookForTrailingSpaces | wc -l)
+#if (( $num_trailing_space_lines > 0 )) ; then
+#    echo "Found $num_trailing_space_lines lines with trailing spaces :-("
+#    echo $($CHPL_HOME/util/devel/lookForTrailingSpaces)
+#    exit 1
+#fi
+
 # Check for standard C headers.
 standard_c_headers=$($CHPL_HOME/util/devel/grepstdchdrs || :)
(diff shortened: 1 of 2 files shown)


ben-albrecht:
Can we do tabs while we're at it?

ben-albrecht:
Following the edits on my last comment, perhaps `tools/` should be added to the `grep` list.

util/buildRelease/smokeTest
```diff
@@ -45,6 +45,15 @@ if (( $num_tabs > 0 )) ; then
     exit 1
 fi
 
+# Need to clean up all of the trailing spaces before adding this
+# Check for trailing spaces
+#num_trailing_space_lines=$($CHPL_HOME/util/devel/lookForTrailingSpaces | wc -l)
+#if (( $num_trailing_space_lines > 0 )) ; then
+#    echo "Found $num_trailing_space_lines lines with trailing spaces :-("
+#    echo $($CHPL_HOME/util/devel/lookForTrailingSpaces)
+#    exit 1
```
ben-albrecht: Can we uncomment this block except for the `exit 1` for now?

daviditen:
I don't think we should uncomment it until trailing spaces have been cleaned up.  The current output of smokeTest is ~2400 lines, and uncommenting that block would generate another ~4400 lines.  It would add around 200 KB to every smokeTest log file.

PR #7849, opened 2017-11-21
Remove unnecessary autoCopy for reductions state

vasslitvinov:
Initialization of the shadow variable for a reduce intent
included an autoCopy of the value returned by reduceOp.identity.

Valgrind paratest run shows that this is unnecessary,
i.e. it passes with the autoCopy removed.
So, making this change on master.

Testing: paratest on valgrind+quickstart configuration

### DIFF
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -1440,13 +1440,9 @@ static void propagateThroughYield(CallExpr* rcall,
           // within a param conditional on a not-taken branch.
           svar = new VarSymbol(intentArgName(ix, "shadowVarReduc"));
           svar->addFlag(FLAG_INSERT_AUTO_DESTROY);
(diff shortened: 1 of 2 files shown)


vasslitvinov:
@mppf or @noakesmichael - would you review?

mppf:
@vasslitvinov - this seems OK, but I'd be cautious in merging it - just be ready for testing failures the next day. The identity method should return an "owned" thing, so you shouldn't need to auto-copy it. Also, can you try to write a test that shows a memory leak before this change, but that this change fixes?

mppf:
Oh, and one more test to think about. Does it work well if the identity method uses the ref return intent?

vasslitvinov:
This breaks when the identity method returns by ref. See the test I just added in #7863 :

    test/parallel/forall/reduce-intents/ri-a1-ident-ref-ret.chpl

Closing this for now.

PR #7854, opened 2017-11-22
Move PRIM_SIZEOF fixup from resolveMoveForRhsCallExpr to preFoldPrimOp

mppf:
Since this fixup doesn't actually do anything with the PRIM_MOVE in which it is contained (it only modifies the RHS), it makes more sense in preFold than as a special case for resolveMove.

- [x] full local testing
- [x] hellos with CHPL_LOCALE_MODEL=numa


mppf:
@noakesmichael - I think this is trivial and an improvement but I'd appreciate if you can have a quick look. Thanks!

PR #7858, opened 2017-11-22
Rewrite default argument handling

### LINKED ISSUE #7855: Default arguments for iterators are destroyed before iterator completes
For example, the following program has unpredictable (aka stack trash) output:

``` chapel
proc makeTuple() {
  var s = "hello"*2;
  return (0,s);
}

iter myiter(offset=makeTuple()) {
  for i in 1..10 {
    yield offset;
  }
}

proc test() {

  writeln(makeTuple());

  for x in myiter() {
    writeln(x);
  }
}

test();
```

### LINKED ISSUE #7868: function visibility and default arguments
In PR #7858, I'm exploring how adjustments to the implementation of default arguments might resolve some current issues (and that effort is part of an effort to clean up iterator memory management).

I've encountered one issue, which I simply worked around in the PR. The issue is that before that PR, default arguments are resolved in the scope of the function in which they are declared, but after the PR, they are resolved in the scope of the call. That matters when the default argument is a function call that is marked private, as in the following example:.

``` chapel
module Library {
  private proc myZero(param rank : int, type idxType) {
    var ret : rank*idxType;
    return ret;
  }

  proc foo(param rank=1,
           type idxType=int,
           tup:rank*idxType = myZero(rank, idxType)) {
    writeln("In foo ", tup);
  }

}

module Test {
  use Library;

  proc main() {
    foo(1, real);
  }
}
```

I think we could support such code patterns with the work of the PR, but they would require some special treatment. One strategy would be for the compiler to create a function for each default argument at the scope of the function that they apply to. However I'm uncertain how valuable this pattern is from a language design point of view.

Such private default arguments seem questionable to me since I'd like users to be able to implement themselves whatever the compiler is doing with the default arguments. In other words, since `myZero` above is private, somebody wanting to explicitly pass it in to the `tup` argument of `foo` cannot do so.

Do we need to support this pattern?

### LINKED ISSUE #7889: How does promotion interact with default arguments?
When default arguments are combined with promotions, how many times should the default argument expression be run? Just once, or once per promoted element?

For example:

``` chapel
proc makeit() {
  writeln("IN makeit");
  return 1;
}

proc foo(x:int, y=makeit()) {
  return x+y;
}

var A = [1,2,3,4,5,6,7,8,9];
writeln(A);
var B = foo(A);
writeln(B);
```

could print out "IN makeit" only once or it could print it out 10 times.

A program like the following could work either way, since the default argument depends on a previous argument, which itself is a promoted array - meaning that the default argument is also promoted.

``` chapel
proc bar(x:int) {
  return x + 1;
}

proc foo(x:int, y:int = bar(x)) {
  assert(x+1==y);
  return x+y;
}
var A = [1,2,3,4,5,6,7,8,9];
writeln(A);
var B = foo(A);
writeln(B);
```

A worrying pattern would be something like this:
``` chapel
class C { var x:int; }
proc foo(ref x:C, y:C = new C(1)) {
  if x == nil then x = y;
}
var A:[1..3] C;
foo(A);
A[1].x = 100;
writeln(A); // are these all pointing to the same C instance?
```

While that example is contrived... it might be more plausible than programs that do output in the promoted expression. In any case it seems to me that either behavior would be OK for that example, provided the specification is clear.

Anybody have insight as to whether or not just running the default expression once in these cases is sufficient - or if it needs to be run many times. Master runs it many times right now, but I don't yet know if that was intentional or merely an artifact of the implementation strategy.

### LINKED ISSUE #7940: should Chapel support promotion caused by a default argument?
Consider the test here: test/functions/deitz/test_default_promote_instantiate.chpl, a simplified version of which is reproduced below:

 ``` chapel
proc foo(x: int = 1..3) {
  writeln(x);
}

foo();
```

This test relies on the default-argument-wrapper behavior of master (before #7858, at least) and the .good file expects the default argument to cause the zero-argument call to `foo` to be a promoted call like `foo(1..3)`.

My mental model for how promotion and default arguments interact involves two conceptual transformations, from the discussion over on #7889. My understanding from that issue is that we imagine doing the promotion transformation (i.e. turn it in to a forall loop) before doing the default arguments transformation (i.e. supply arguments with the default expressions). But that model doesn't allow the functionality that this test checks for:

 * original program
    ```chpl
    foo();
    ```
 * after promotion transformation (there is no promotion if there are no arguments, right?)
   ```chpl
   foo();
   ```
 * after default argument transformation
    ```chpl
   foo(1..3); // error: can't call foo(integral) with 1..3
    ```

Hence I am advocating that we make this test result in compilation error.

### PR DESCRIPTION
**Pull request by mppf**
This PR reworks default argument handling.

I created this PR in order to resolve the issue #7855 in which a default argument is destroyed before it can be captured by an iterator (since it's being captured in the iterator by refererence - see the issue for details), which in turn will allow me to make further progress on fixing iterator memory management problems. This change solves that issue by storing the default argument at the call site (say the call to the iterator). A possible alternative way to solve that problem might have been to change how such default arguments are stored into the iterator record, but that seemed to require more interaction between the passes than I would like. Additionally, this same pattern could come up without default arguments with a function that returns an iterator or with a value default argument that is passed by reference to an iterator.

The old wrapper function implementation is entangled with compiler-generated default constructors and so this PR leaves the old implementation and calls it for default constructors only. Once initializers are further along we can expect to remove that branch.

The new default wrapper implementation works as follows:
 * For each default argument expression that is relied upon, generate a function that returns that value. This new function is scoped next to the original function.
 * At the call site, when an argument is not provided and it has a default argument expression, call the generated function to get it

In order to handle the combination of default arguments and promotion as discussed and decided upon in issue #7889, this PR additionally makes some changes to the creation of promotion wrappers. After this PR, `promotionWrap` is responsible for handling default arguments and it does so by calling `addDefaultsAndReorder`. To make it easier to reason about the combination of promotion and default arguments, this PR introduces a type PromotionInfo that tracks information about which formals have defaults and which are promoted.

Resolves #7855.
Closes #7868, closes #7889, closes #7940 .

- [x] full local baseline testing
- [x] full --no-local testing

### CODE REVIEW

**Comment by mppf** [comment]
This PR has the side-effect of making a default argument be executed only once in a promoted context; for example:
``` chapel
proc makeit() {
  writeln("IN makeit");
  return 1;
}

proc foo(x:int, y=makeit()) {
  return x+y;
}

var A = [1,2,3,4,5,6,7,8,9];
writeln(A);
var B = foo(A);
writeln(B);
```

prints "IN makeit" only once after this PR, but printed 10 such lines before.

test/functions/default-arguments/default-ref.chpl
```diff
@@ -0,0 +1,7 @@
+proc foo(ref a:int, ref b:int = a) {
```
vasslitvinov: This is pretty cool.

compiler/resolution/wrappers.cpp
```diff
@@ -71,16 +84,26 @@ static void       reorderActuals(FnSymbol*                fn,
...
 static ArgSymbol* copyFormalForWrapper(ArgSymbol* formal);
 
+typedef struct DefaultExprFnEntry_s {
+  FnSymbol* defaultExprFn;
+  std::vector<std::pair<ArgSymbol*,ArgSymbol*> > usedFormals;
+} DefaultExprFnEntry;
+
+typedef std::map<ArgSymbol*, DefaultExprFnEntry> formalToDefaultExprEntryMap;
+formalToDefaultExprEntryMap formalToDefaultExprEntry;
+
```
vasslitvinov: While there, could you clarify the comment below about 'actualIdxToFormals' ?
The comment says it contains the actuals, however that cannot be true
because an actual is not necessarily an ArgSymbol.

Understanding actualIdxToFormals would help understanding wrapAndCleanUpActuals()
as well as addDefaultsAndReorder().

compiler/resolution/wrappers.cpp
```diff
@@ -156,20 +221,507 @@ static void      formalIsDefaulted(FnSymbol*  fn,
...
+  // Remove the actuals from the call
+  // (we'll add them back again in a moment)
+  for_actuals(actual, call) {
+    actual->remove();
+  }
+
+  // Create a copyMap to handle cases like
+  //   proc f(a, b=a)
+  // in which event, we can't refer to formal 'a' in the default value.
```
vasslitvinov: Can you clarify the above comment? I see in your tests that the default value CAN refer to 'a'.
Is the comment a typo? Did you mean something else?

Issue #7104, opened 2017-08-24
Improve TOML library 


Spartee:
To compare against TOML spec -> [TOML Spec](https://github.com/toml-lang/toml)

## TOML Spec to add support for
- [ ] Array of tables (hardest to implement, most needed)
- [ ] Exponent reals
- [ ] Underscore notation for integers and reals
- [ ] Date
- [ ] Time
- [ ] Datetime with offset

## Known bugs
- [ ] Quoted keys with periods such as "this.key" = value
- [ ] Comment on the last line of the file (easy)

## User errors currently not caught
- [ ] Table name repeated (overwrites table)
- [ ] key name repeated in same table (overwrites value)
- [ ] Arrays of mixed data types e.g. [2, "string", 4.5]

## Testing
- [ ] More futures with user errors like key or table name repeated
- [ ] Tests for handled errors (once error handling is put in)

## Other
- [ ] plug some of the remaining memory leaks
- [ ] error handling instead of halt()

sanket143:
Added date support in #12802 :arrow_up:

sanket143:
Known Bug: #12843 (Inline comment treated as if it's in written in a different line)

ankingcodes:
@ben-albrecht  
> Quoted keys with periods such as "this.key" = value

Is this related to #13332 ?

ben-albrecht:
> > Quoted keys with periods such as "this.key" = value
> 
> Is this related to #13332 ?

No, it is supporting quoted keys, such as described in the [TOML spec](https://github.com/toml-lang/toml#keys).



> > Comment on the last line of the file (easy)
> 
> Could you elaborate this issue ?

I'm actually not sure about this one. Does having a comment on the last line of the file currently produce an error? @Spartee may have a better idea.

Spartee:
So there was a issue(which could be resolved now I haven't tested it recently) where having a comment on the last line of the file would produce a rather unhelpful TomlError.

ankingcodes:
@ben-albrecht I have started to work on `Datetime with offset`. Should the regex validate for leap years too ? 
Also there's a case mentioned for replacing `T` in `1979-05-27T00:32:00-07:00` with a space. Should I consider that case?
If yes, our regex would fail there ! Additionally, the TOML module would also consider `1979-05-27` and `00:32:00-07:00` as different tokens if a space exists.

This is the regex I plan to use for this feature: 
```
[...]

PR #7859, opened 2017-11-22
TOML test suites

ben-albrecht:
This PR adds 2 test suites to the `test/modules/packages/Toml` directory:

- [specification](https://github.com/toml-lang/toml/tree/master/tests)
- [BurntSushi](https://github.com/BurntSushi/toml-test/tree/master/tests/valid)
  - `valid` tests only

Both test suites are futures for now, as the majority of tests fail.

Ideally, we'd have a future per execution option, but this is not yet supported in `start_test`

## Details

In supporting the BurntSushi test suite, I added a JSON writer to the TOML module.

The BurntSushi "good" files are provided as validation JSON files, and diffing is done comparing the JSON data structure produced by the TOML module. For now, this is handled with Python scripts (see `compare.py` and the caller, `PREDIFF`)

The specification tests do not provide validation JSON files, so I stuck to the provided TOML files as good files. We may want to generate the validation JSON files and use the same testing methodology for that suite eventually.

## Other changes

- Added `tomlType` method for getting the `tag` in a string format (compatible with the JSON writer)
- Added empty domain checks to fix case where parsing an empty toml file would produce an error
- Did some minor refactoring in `TOML.chpl`
   - Added `writeTOML()` which `writeThis()` calls
   - Made newlines more explicit in `writeTOML()`
- Hoisted the `re2` `SKIPIF` out to `Toml.skipif` for all `Toml/` tests

### Testing

- [x] `test/modules/packages/Toml`
    - `specification/` : `0/5`
    - `BurntSushi/` : `7/37`
- [x] `test/mason`


ben-albrecht:
Side note: there were no big surprises in running these tests - most of the cases that these tests caught are already documented as TODO features in #7104

ben-albrecht:
note about licensing: both projects use the `WTFPL` license

modules/packages/TOML.chpl
```diff
@@ -76,6 +74,8 @@ module TomlParser {
 
   /* Prints a line by line output of parsing process */
   config const debugTomlParser = false;
+  /* Number of spaces in an indentation for JSON output */
+  config const tabSpace = 4;
```
benharsh: Maybe just make this a 'const' for now? 'tabSpace' seems like a config that could conflict with a lot of other packages.

ben-albrecht: Sure thing

PR #7869, opened 2017-11-27
Simplify check in promotion wrapper for void return

mppf:
The checks for a void return type in buildPromotionWrapper were unnecessarily
complex (see commit bf90301a9844cc7e933018e1f9cc9dda9ebf522d for some history
in this area). In particular, they just need to check if the promoted scalar
function returns 'void' or not, but were doing it two ways depending on
FLAG_EXTERN. That oddly enough led to compilation failures for the following
program:

``` chapel
  var A:[1..10] atomic int;

  A.add(1, memory_order_seq_cst);

  writeln(A[1].read());
```

even though the version without the optional memory order argument compiles
just fine.

I resolved this issue by considering the promoted function to return 'void' if
it returns gVoid or has dtVoid return type.


daviditen:
This change looks good to me.

PR #4518, opened 2016-09-13
Fix a compile problem with fltk on El Capitan.

tomyhoi:
Compile errors on El Capitan due to more strict enum checking.
Add the fixed file and change make file to copy fixed file after untar of fltk.

PR #7871, opened 2017-11-27
Upgrade fltk from 1.3.3. to 1.3.4-2

ronawho:
This fixes build issues on newer versions of OSX.

With fltk 1.3.3 on Sierra with clang 9.0.0 I was seeing errors like:

    Fl_Tree_Item.cxx:552:38: error: ordered comparison between pointer and zero ('Fl_Tree_Item *' and 'int')
       if ( from_parent->deparent(from) < 0 )      // deparent self from current parent

This just upgrades to 1.3.4-2, which includes a fix for that build error as
well as a fix for the issue that #4518 worked around, so I reverted that PR.


ronawho:
@tomyhoi I'm a chplvis noob, so if you have any time to look at this and verify I didn't break anything I'd appreciate it

ronawho:
I'm going to merge this, but I would still appreciate a post-commit review whenever you get a chance

PR #7873, opened 2017-11-28
Update LLVM alias analysis metadata part 1

dmk42:
Update LLVM TBAA metadata to handle structs better in the new TBAA format.  All of the following metadata types are created where appropriate.
*  Scalar type node
*  Struct type node
*  Scalar access tag (with and without const)
*  Struct access tag (with and without const)
*  Memcpy access tag (tbaa.struct)

Update to use an MDBuilder helper wherever reasonable to improve readability.

IRBuilder objects are renamed from `builder` to `irBuilder` to make room for `mdBuilder`.

Tested with LLVM 4.0 and 5.0.


compiler/include/llvmUtil.h
```diff
@@ -27,6 +27,7 @@
 #include "llvmVer.h"
 
 #include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/MDBuilder.h"
```
mppf: I don't think llvmUtil should include MDBuilder. It includes IRBuilder.h only because it uses llvm::IRBuilder<> in the prototypes below.

compiler/include/codegen.h
```diff
@@ -92,6 +93,7 @@ struct GenInfo {
   // Once we get to code generation....
   llvm::Module *module;
   llvm::IRBuilder<> *builder;
+  llvm::MDBuilder *mdHelper;
```
mppf: I wonder if we should name it mdBuilder? And consider renaming 'builder' to irBuilder?

compiler/util/llvmDebug.cpp
```diff
@@ -18,6 +18,7 @@
  */
 
 #include "llvmDebug.h"
+#include "llvmVer.h"
```
mppf: We've been sorting includes for these LLVM files into 4 parts:
 1. Include the header for the file we're working on (e.g. llvmDebug.h here)
 2. Include, in alphabetical order, other Chapel compiler headers
 3. Include system headers
 4. Include LLVM headers, in HAVE_LLVM

Here, I think the llvmVer header should be sorted within the other Chapel headers or possibly included within the HAVE_LLVM.

compiler/codegen/expr.cpp
```diff
@@ -453,7 +453,8 @@ llvm::StoreInst* codegenStoreLLVM(llvm::Value* val,
   GenInfo *info = gGenInfo;
   llvm::StoreInst* ret = info->builder->CreateStore(val, ptr);
   llvm::MDNode* tbaa = NULL;
-  if( USE_TBAA && valType ) tbaa = valType->symbol->llvmTbaaAccessTag;
+  if( USE_TBAA && valType && !valType->symbol->llvmTbaaStructCopyNode )
+    tbaa = valType->symbol->llvmTbaaAccessTag;
```
mppf: It seems surprising to me that we won't mark a store or a load with TBAA metadata if a StructCopyNode is set for that type.

compiler/codegen/expr.cpp
```diff
@@ -511,7 +512,7 @@ llvm::LoadInst* codegenLoadLLVM(llvm::Value* ptr,
   GenInfo* info = gGenInfo;
   llvm::LoadInst* ret = info->builder->CreateLoad(ptr);
   llvm::MDNode* tbaa = NULL;
-  if( USE_TBAA && valType ) {
+  if( USE_TBAA && valType && !valType->symbol->llvmTbaaStructCopyNode ) {
```
mppf: Ditto here

compiler/codegen/expr.cpp
dmk42: This is to prevent the wrong kind of TBAA node from being used.  More is coming.

dmk42:
I agree that tests will be useful.  I have been disassembling the LLVM IR to look at the TBAA nodes.

PR #7875, opened 2017-11-28
Update arrayatomics.future and test

cassella:
test/types/atomic/albrecht/arrayatomics.chpl fails, but not how the
.future file describes it.  So tweak the .future.

Link the future to a github issue. (#7874)

Make the test's variable n a config to make it easier to experiment
with.


cassella:
The incorrect output is consistent on my machine, but it has the feel of something that might be different on different configurations.  Running with --n 10, or 20, or 100, produces output that has some sort of pattern to it.  At all values 10 and higher that I've tried, the incorrect output's largest number is one less than the correct largest number.

mppf:
The output varies from machine to machine. In particular, it seems to produce #cores 0s. I have a theory about what is going on. I modified the program to use the following, as a substitute for the `outputs[D] = a.fetchAdd(inputs);` line.
``` chapel
proc apply(x:int) {
  return a.fetchAdd(x);
}
outputs[D] = apply(inputs);
```

and this worked correctly. I think what is happening is that the variable `a` is not being captured by reference - but is instead being copied - in the promoted forall. This is happening because the iterator record captures it by value instead of by reference. My ongoing work (in branch removeInsertAutoCopyTemps, no PR yet) resolves this problem.

mppf:
@cassella - I'd like to close this PR, because the .bad will vary from machine to machine, and because I'm hoping to fix this .future in the relative near term. (Alternatively, we could take out the .bad and leave in the .chpl/.future file updates).

cassella:
Ok.  I'm fine either way.  Feel free to merge or close the PR.

ronawho:
I believe this future was closed with https://github.com/chapel-lang/chapel/pull/8073 (4be18607a35067290008e91237c65f0cd4330553), so I'm going to close this

Issue #7866, opened 2017-11-27
captured promotion of function returning void is not an error


mppf:
The following program compiles and executes but I would have expected it to result in compilation error, since the commented out section at the end results in compilation error.

``` chapel
proc ref int.foo() { // returns void
  this += 1;
}

var A = [1,2,3,4,5,6,7,8,9];
writeln(A);

var B = A.foo();  // shouldn't this be an error?
writeln(B.type:string);
writeln(B);


//var C = A[1].foo();  // this is an error
//writeln(C.type:string);
//writeln(C);
```

mppf:
@daviditen - I'm curious of your opinion of this issue

daviditen:
I agree that the assignment to `B` should cause a compiler error.  I suspect that the promotion is throwing off the void checking.

daviditen:
@mppf could you please take a look at #7878?

PR #7878, opened 2017-11-28
Fix a bug where a function that doesn't return is assignable to a variable

daviditen:
Wrappers for functions with FLAG_VOID_NO_RETURN_VALUE should have the same flag

When creating a wrapper for a function with FLAG_VOID_NO_RETURN_VALUE, give
the wrapper the same flag.  This fixes an issue where a function that doesn't
return is assignable to a variable if it is promoted.  It should be an error.

Added a test taken from issue #7866 that was failing by allowing a non-returning
function to be assigned to a variable.


daviditen:

daviditen:
Thank you!

PR #7879, opened 2017-11-28
Retry performance graph sync a few times

ronawho:
We've had a few cases where our perf graphs have failed to sync to dreamhost.
The failures seem to be sporadic and resolve quickly, so retry 5 times with a
30 second delay to hopefully minimize the number of sync failures we see.

If this isn't enough time, we can always switch to an exponential backoff
scheme, but I'm keeping it simple to start.


ronawho:
@awallace-cray thoughts?

awallace-cray:
Looks good!

PR #7887, opened 2017-11-29
Fix for memory leak on Blowfish tests

saru95:
Apparently, Blowfish has certain restrictions on its key and IV sizes. Since the key size in blowfish is variable (recommended is 16 bytes), it is advised to use a key-derivation function with size 16 (as done in `bfTest2.chpl`). I'm also adding a note to the docs for the same. We can also strictly enforce a 16-byte key, just in case.

P.S: Sorry, for the huge delay. Was (and still am) stuck with a major release at work.


saru95:
cc @daviditen

daviditen:
Thanks for the update.  I tested it on OSX and Linux, and with valgrind.  I think it looks ready to merge.

PR #7901, opened 2017-11-30
Fix CHPL_LLVM=system compilation flags

dmk42:
The command-line flags for compiling LLVM come from the tool `llvm-config`.  When using the system LLVM, `llvm-config` often reports flags that cause us problems, such as a `-Wall` flag in the wrong sequence for our purposes.

Makefile.share-system already filters out several undesired flags.  This change filters out all added warnings to prevent compilation errors when either CHPL_DEVELOPER or WARNINGS is set.

Also, cause a `make clean` to remove CLANG_SETTINGS so it gets reinitialized properly.

Support is added for Mac Homebrew compilations with CHPL_LLVM=system.

### DIFF
--- a/third-party/llvm/Makefile.share-system
+++ b/third-party/llvm/Makefile.share-system
@@ -18,11 +18,11 @@ endif
 
 # LLVM preprocessor flags (ie -Dbla and -Ibla) 
 ifndef LLVM_CONFIG_CXXFLAGS
-  export LLVM_CONFIG_CXXFLAGS=$(shell test -x $(LLVM_CONFIG) && $(LLVM_CONFIG) --cxxflags | sed 's/ -O[0-4s] / /' | sed 's/ -DNDEBUG / /' | sed 's/ -pedantic / /' | sed 's/ -W / /')
+  export LLVM_CONFIG_CXXFLAGS=$(shell test -x $(LLVM_CONFIG) && $(LLVM_CONFIG) --cxxflags | sed 's/ -O[0-4s] / /' | sed 's/ -DNDEBUG / /' | sed 's/ -pedantic / /' | sed 's/ -W / /' | sed 's/ -Wall / /')
 endif
 
 ifndef LLVM_CONFIG_CFLAGS
-  export LLVM_CONFIG_CFLAGS=$(shell test -x $(LLVM_CONFIG) && $(LLVM_CONFIG) --cflags | sed 's/ -O[0-4s] / /' | sed 's/ -DNDEBUG / /' | sed 's/ -pedantic/ /' | sed 's/ -W / /' )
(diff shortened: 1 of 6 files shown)


third-party/llvm/Makefile.share-system
```diff
@@ -18,11 +18,11 @@ endif
 
 # LLVM preprocessor flags (ie -Dbla and -Ibla) 
 ifndef LLVM_CONFIG_CXXFLAGS
-  export LLVM_CONFIG_CXXFLAGS=$(shell test -x $(LLVM_CONFIG) && $(LLVM_CONFIG) --cxxflags | sed 's/ -O[0-4s] / /' | sed 's/ -DNDEBUG / /' | sed 's/ -pedantic / /' | sed 's/ -W / /' | sed 's/ -Wall / /')
+  export LLVM_CONFIG_CXXFLAGS=$(shell test -x $(LLVM_CONFIG) && $(LLVM_CONFIG) --cxxflags | xargs -n 1 | awk -f $(THIRD_PARTY_DIR)/llvm/filter-llvm-config.awk | xargs)
```
mppf: If we're writing a script anyway, why not write a .sh / .bash script that returns the right stuff when passed the llvm-config arguments to run?
Separately, I'm not sure if xargs would be a new dependency if we use it here.

third-party/llvm/find-llvm-config.sh
```diff
@@ -14,6 +15,15 @@ then
 elif command_exists llvm-config
 then
   command -v llvm-config
+elif command_exists /usr/local/opt/llvm@$PREFERRED_VERSION/bin/llvm-config
```
mppf: are these MacPorts or Homebrew paths or something? Might be worth a comment saying what they're intending to support.

third-party/llvm/Makefile.share-system
dmk42: It seems to me that an sh or bash script would be much more complicated than the awk script.  I don't think it would be worth it.

A standard POSIX environment is required to include xargs, though we could mention it the same place we mention awk.

dmk42:
@mppf , what do you think of using the same mechanism for third-party/llvm/Makefile.share-included ?  Then the llvm-config filtering would be consistent across different compilation modes.

mppf:
@dmk42 - I honestly can't remember why Makefile.share-included runs llvm-config at all. I would have thought that Makefile.share-included would only be needed for using the clang-included C compiler. Anyway, as long as we are running llvm-config there, it should use the same filtering. Thanks.

third-party/llvm/Makefile.share-system
mppf: That's fair - I didn't already know it was part of the standard POSIX environment.

PR #7907, opened 2017-11-30
Fix the name of an internal QT profiling counter.

gbtitus:
When we changed `profile_task_taskCall` to `profile_task_taskCallFTable` we
missed changing the only reference to it.  And apparently we've never
turned on internal Qthreads shim profiling snce then, so we didn't
realize that it wouldn't build any more.  Fix this.


gbtitus:
@ronawho, quick review?

PR #7908, opened 2017-12-01
file.check throws

psahabu:
@mppf pointed out that `file.check()` was halting when it should throw. It has been updated to throw a `FileNotFoundError`.

In order to build a `FileNotFoundError` with just a string, it was necessary to add a cast to its constructor's default `syserr` argument. This was repeated across all `SystemError` subtypes. (There may be a better way to achieve this coercion.)

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1307,7 +1307,7 @@ proc file.check() throws {
     throw SystemError.fromSyserr(EBADF, "Operation attempted on an invalid file");
 }
 
-/* Emit a syserr if a file is invalid */
+/* Return a syserr through out error if a file is invalid */
 proc file.check(out error:syserr) {
   try! {
     check();
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1385,8 +1385,10 @@ proc file._style:iostyle throws {
 proc file.close(out error:syserr) {
   check(error);
 
-  on this.home {
-    error = qio_file_close(_file_internal);
+  if !error {
+    on this.home {
+      error = qio_file_close(_file_internal);
+    }
   }
 }
 
@@ -1416,8 +1418,10 @@ This function will typically call the ``fsync`` system call.
 proc file.fsync(out error:syserr) {
   check(error);
 
-  on this.home {
-    error = qio_file_sync(_file_internal);
+  if !error {
+    on this.home {
+      error = qio_file_sync(_file_internal);
+    }
   }
 }
 
@@ -1448,17 +1452,20 @@ to get the path to a file.
 proc file.getPath(out error:syserr) : string {
   check(error);
 
-  var ret: string;
-  on this.home {
-    var tmp:c_string_copy;
-    var tmp2:c_string_copy;
-    error = qio_file_path(_file_internal, tmp);
-    if !error {
-      error = qio_shortest_path(_file_internal, tmp2, tmp);
+  var ret: string = "unknown";
+  if !error {
+    on this.home {
+      var tmp:c_string_copy;
+      var tmp2:c_string_copy;
+      error = qio_file_path(_file_internal, tmp);
+      if !error {
+        error = qio_shortest_path(_file_internal, tmp2, tmp);
+      }
+      chpl_free_c_string_copy(tmp);
+      if !error {
+        ret = new string(tmp2, needToCopy=false);
+      }
     }
-    chpl_free_c_string_copy(tmp);
-    ret = if error then "unknown"
-                   else new string(tmp2, needToCopy=false);
   }
   return ret;
 }
@@ -1471,12 +1478,7 @@ a problem getting the path to the open file.
 */
 proc file.tryGetPath() : string {
   var err:syserr = ENOERR;
-  var ret:string;
-  try! {
-    ret = this.getPath(err);
-  } catch e: SystemError {
-    err = EBADF;
-  }
+  var ret = this.getPath(err);
 
   if err then return "unknown";
   else return ret;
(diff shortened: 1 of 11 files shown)


modules/standard/IO.chpl
```diff
@@ -1337,7 +1336,7 @@ proc file.unlock() {
 // channel style is protected by channel lock, can be modified.
 pragma "no doc"
 proc file._style:iostyle {
-  check();
+  try! check();
```
mppf: Confused. Why is using `try!` here better than just halting in `file.check` ? Isn't the point here to remove halts from the standard library?

modules/standard/IO.chpl
```diff
@@ -1302,10 +1302,9 @@ proc =(ref ret:file, x:file) {
 }
 
 /* Halt if a file is invalid */
-proc file.check() {
-  if(is_c_nil(_file_internal)) {
-    halt("Operation attempted on an invalid file");
-  }
+proc file.check() throws {
+  if is_c_nil(_file_internal) then
+    throw new FileNotFoundError("Operation attempted on an invalid file");
```
mppf: The error ends up looking a bit odd this way, e.g.
```
$ ./util/start_test  test/library/standard/FileSystem/lydia/sameFile/givenNull.chpl
...
[Executing diff givenNull.good givenNull.exec.out.tmp]
1c1,3
< givenNull.chpl:7: error: halt reached - Operation attempted on an invalid file
---
> uncaught FileNotFoundError: No such file or directory (Operation attempted on an invalid file)
>   givenNull.chpl:7: thrown here
>   givenNull.chpl:7: uncaught here
```

psahabu: Ah that is a bit weird, I'll switch to a `SystemError`.

modules/standard/IO.chpl
psahabu: I have no idea what I was thinking. Corrected.

modules/standard/IO.chpl
```diff
@@ -1301,12 +1301,21 @@ proc =(ref ret:file, x:file) {
...
 }
 
-/* Halt if a file is invalid */
+/* Throw an error if a file is invalid */
 proc file.check() throws {
   if is_c_nil(_file_internal) then
     throw SystemError.fromSyserr(EBADF, "Operation attempted on an invalid file");
 }
 
+/* Emit a syserr if a file is invalid */
```
mppf: I think you mean "Return an error in the error argument if..."

psahabu:
@mppf:

I'm running into a lot of tests that aren't prepared to handle an `out error` from `check()` (or a function that calls `check()`). Since I'll have to modify the tests anyway, I'm thinking I'll transition them to proper error handling and deprecate the `out error` functions in IO.

I think the best way to achieve that is by closing this PR and using this branch as the basis for a larger PR. What do you think?

mppf:
@psahabu - I don't think we should deprecate `out error` in this PR. If tests are relying on code calling `check` to not throw, for example because that code used the `out error` mechanism, and the test also uses the `out error` mechanism, the error from `check` should just be stored in the `out error`.

Can you show an example?

psahabu:
Alright, I discovered that `check(out error)` was making things go haywire in codegen. This is the compiler creates a formal inside the function that gets copied out at the end, and that formal was being initialized with garbage memory. (This might be because `syserr` has no default value, or because the `out` variable was passed through.)

To patch this, I had `check(out error)` use an internal `err` set to `ENOERR`, setting `error` at the end of the function. It's not the prettiest workaround, but it's the fastest and cleanest.

One remaining issue is that `test/memory/figueroa/LeakedMemory6.chpl` is still failing for a very explosive, hard to parse reason. When run with start_test, this error is emitted:

```
> In file included from /var/folders/x3/62y0d0t16qnglkph2yr6gcmw000xk6/T//chpl-psahabu-3722.deleteme/_main.c:55:
> /var/folders/x3/62y0d0t16qnglkph2yr6gcmw000xk6/T//chpl-psahabu-3722.deleteme/IO.c:378:19: error: passing 'chpl____wide_Error' (aka 'struct chpl_chpl____wide_Error_s') to parameter of incompatible type 'Error' (aka 'struct chpl_Error_s *')
> chpl_delete_error(chpl_macro_tmp_1917, _ln_chpl, _fn_chpl);
>                   ^~~~~~~~~~~~~~~~~~~
> /var/folders/x3/62y0d0t16qnglkph2yr6gcmw000xk6/T//chpl-psahabu-3722.deleteme/ChapelError.c:1583:37: note: passing argument to parameter 'err' here
> static void chpl_delete_error(Error err,
>                                     ^
> 1 error generated.
> make: *** [/var/folders/x3/62y0d0t16qnglkph2yr6gcmw000xk6/T//chpl-psahabu-3722.deleteme/LeakedMemory6.tmp] Error 1
> error: compiling generated source
```

I'm having trouble reproducing it when compiling and running manually. Any ideas on what might be going on here?

mppf:
Yes, syserr does not have a default value. That itself might be easy to fix. But could you make/add to an issue saying that extern types don't have default values?

Re test/memory/figueroa/LeakedMemory6.chpl, this test throws --no-local and it's almost certainly the case that the error you are seeing is from chpl_delete_error being called with a wide pointer instead of a narrow one. Why this PR caused that to come up in the generated C code is unclear to me. I'd try compiling hello.chpl with --no-local (and I'd wager it probably doesn't compile...). I'm expecting the compiler to generate a chpl_delete_error that accepts a wide pointer in that context. Sometimes a workaround for such issues is to add something to the standard library that calls chpl_delete_error with a guaranteed-to-be-wide pointer.

psahabu:
Note: `errorHandling.cpp:377` also contains a normalization violation, but replacing it with the normalized `errorCond()` causes a deinit to be inserted on invalid memory for `test/errhandling/ferguson/err-return.chpl`.

PR #7909, opened 2017-12-01
Avoid premature primitive type errors in normalize

bradcray:
This change adds an argument to typeForTypeSpecifier() that indicates whether or not it is legal for the routine to throw fatal errors.  Early in the compiler (e.g., normalize) when code is being processed that may be folded out or never executed, generating such errors may be incorrect since the code may never be reached.  For this reason, made all early calls to the routine pass in `false` for this argument.

Added some tests indicating such cases which did not work prior to this change.


bradcray:
This branch currently has two errors due to early normalization-time error messages being replaced with later resolution-time errors:

```
[Error matching compiler output for types/scalar/bharshbarg/illegalSizes]
[Error matching compiler output for types/scalar/diten/floatingPointBadSize]
```

mppf:
Closing, as this has been replaced by PR #7911 which is now merged.

PR #7911, opened 2017-12-01
Type specifier

noakesmichael:
Suppress premature type-specifier errors

Historically all evaluation of type-specifiers was done during the
"function resolution" sub-phase of resolution.  Early work to
support initializers expanded a subset of these cases to enable
types to determined during sub-parts of normalization.

When doing this work I failed to appreciate the role of conditional
param expressions.  It is possible to generate invalid type specifiers
within param conditionals that should evaluate to false.  In the
previous flow the conditional would be folded away before the
type specifier was evaluated.  In the new flow the type specifier
may be evaluated, with a corresponding error message, before
param expressions are evaluated.

This PR enables the error message associated with this action to be
suppressed, and then suppresses them until param conditionals are
processed.

@bradcray developed the approach and the initial implementation.
I have simply completed the work and handled two test regressions.

One .good file was updated in a trivial way to respond to a minor
change to the error output.

The second test explicitly generated multiple error messages.  This
update causes that test to generate just one error in function-resolution.
This PR expands that failing test to both a passing test and a failing future.

Compiled with/without CHPL_DEVELOPER on clang/darwin and
gcc/linux64.  Tested each configuration on a portion of release/.

Passed a single-locale paratest with futures


bradcray:
Very belatedly, thanks for the assist on this, and sorry to have mistakenly thought that this was on the critical path to getting some of Michael's bigger efforts merged (though I'm still very glad to see it fixed).

PR #7919, opened 2017-12-02
Add llvm.system nightly correctness-test job

awallace-cray:
New Jenkins job runs a new kind of Chapel test config, w CHPL_LLVM=system.

Just like the existing "llvm" test config, except:
* CHPL_LLVM=system instead of llvm
* Runs on the recently-built Ubuntu 16.04 64-bit VM with LLVM-4 pre-installed
* Does not use paratest.
 (this job will take > 10 hours)

### DIFF
--- a/util/cron/common-llvm.bash
+++ b/util/cron/common-llvm.bash
@@ -23,14 +23,7 @@ esac
 
 # setup environment based on CHPL_LLVM = "llvm" or "system"
 
-case "$CHPL_LLVM" in
-( system )
-    export CHPL_LLVM=system
-    log_info setting CHPL_LLVM=$CHPL_LLVM
-    ;;
-( llvm )
-    export CHPL_LLVM=llvm
-    log_info setting CHPL_LLVM=$CHPL_LLVM
+if test "$CHPL_LLVM" = llvm; then
 
     # Ensure that python 2.7 is at front of PATH. This is only done for
     # llvm configuration because the test systems are _very_ finicky about
@@ -56,8 +49,7 @@ case "$CHPL_LLVM" in
     else
         echo "[Warning: llvm may not build correctly with cmake: $(which cmake)]"
     fi
-    ;;
-esac
+fi
 
 # Run examples and test/extern/ferguson/.
 export CHPL_NIGHTLY_TEST_DIRS="extern/ferguson"
--- a/util/cron/common-llvm.bash
+++ b/util/cron/common-llvm.bash
@@ -1,35 +1,63 @@
 #!/usr/bin/env bash
 #
 # Setup environment for LLVM testing.
+#
+# We examine the first commandline parameter,
+#   $ source $CWD/common-llvm.bash system # $1 is "system", for example
+# if $1 is "llvm" or "system", we set CHPL_LLVM = $1
+# all other cases, we set CHPL_LLVM = "llvm" for backward-compatibility
 
 CWD=$(cd $(dirname ${BASH_SOURCE[0]}) ; pwd)
 
-export CHPL_LLVM=llvm
+# set CHPL_LLVM = "llvm" or "system", based on $1 if any
+# (default CHPL_LLVM = "llvm", and CHPL_LLVM = "none" is not allowed)
 
-# Ensure that python 2.7 is at front of PATH. This is only done for
-# llvm configuration because the test systems are _very_ finicky about
-# python 2.6 due to some environmental configurations.
-#
-# Ideally, this would only apply to the `make compile` step in
-# nightly, but there is not a good way to do that right now. One can
-# imagine adding a "pre-make-compile hook" to nightly. Maybe an env
-# var with a comment to executes, but it seems ok for the chpldoc test
-# to fail on llvm for now.
-# (thomasvandoren, 2015-03-11)
-py27_setup=/data/cf/chapel/setup_python27.bash
-if [ -f "${py27_setup}" ] ; then
-    source ${py27_setup}
-else
-    echo "[Warning: llvm may not build correctly with python: $(which python)]"
-fi
+case "$1" in
+( llvm | system )
+    export CHPL_LLVM=$1
+    ;;
+( * )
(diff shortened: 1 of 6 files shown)


awallace-cray:
Tested in sandbox.

awallace-cray:
@ronawho would you mind taking a glance at this too? Thanks

util/cron/common-llvm.bash
```diff
@@ -1,35 +1,63 @@
...
+( * )
+    export CHPL_LLVM=llvm
+    ;;
+esac
+
+# setup environment based on CHPL_LLVM = "llvm" or "system"
+
+case "$CHPL_LLVM" in
+( system )
+    export CHPL_LLVM=system
```
ronawho: Why do we need to re-export CHPL_LLVM here? Seems like you could get rid of this new case stmt and reduce the diff a lot

util/cron/common-llvm.bash
```diff
@@ -1,35 +1,63 @@
...
-fi
-
-# 2017-08-10 : Developer-installed cmake version required by LLVM 4
-cmake_setup=/data/cf/chapel/setup_cmake39.bash
-if [ -f "${cmake_setup}" ] ; then
-    source ${cmake_setup}
-else
-    echo "[Warning: llvm may not build correctly with cmake: $(which cmake)]"
-fi
+case "$1" in
```
ronawho: I find this a little hard to parse, can this case stmt just be replaced with:

```sh
export CHPL_LLVM=${1:-llvm}
```

ronawho: Or maybe `export CHPL_LLVM=${CHPL_LLVM:-llvm}` and have the test-system script set `CHPL_LLVM=system` before sourcing instead of passing in the config to the script

ronawho:
I'm hoping the diff could be something more like:

```diff
diff --git a/util/cron/common-llvm.bash b/util/cron/common-llvm.bash
index 99fb7838f1..84f2333306 100644
--- a/util/cron/common-llvm.bash
+++ b/util/cron/common-llvm.bash
@@ -4,7 +4,7 @@

 CWD=$(cd $(dirname ${BASH_SOURCE[0]}) ; pwd)

-export CHPL_LLVM=llvm
+export CHPL_LLVM=${CHPL_LLVM:-llvm}

 # Ensure that python 2.7 is at front of PATH. This is only done for
 # llvm configuration because the test systems are _very_ finicky about
diff --git a/util/cron/test-llvm.system.bash b/util/cron/test-llvm.system.bash
new file mode 100755
index 0000000000..81134a59a6
--- /dev/null
+++ b/util/cron/test-llvm.system.bash
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+#
+# Full test suite with CHPL_LLVM=system and pass --llvm flag to
+# compiler on linux64.
+
+# Needs /data/cf/chapel/setup_python27.bash (common-llvm)
+
+CWD=$(cd $(dirname $0) ; pwd)
+source $CWD/common.bash
+
+export CHPL_LLVM=system
+source $CWD/common-llvm.bash
+
+# common-llvm restricts us to extern/ferguson, but we want all the tests
+unset CHPL_NIGHTLY_TEST_DIRS
+
+export CHPL_NIGHTLY_TEST_CONFIG_NAME="llvm.system"
+
+log_info START nightly -cron ${nightly_args}
+$CWD/nightly -cron ${nightly_args}
+log_info nightly EXIT status $?
```

But maybe I'm missing something (i.e. not understanding why some of the changes to common-llvm are needed)

dmk42:
I think the issue is that python 2.7 and cmake are not needed when CHPL_LLVM=system.  Ubuntu 16.04 includes both anyway, though, so the only reason to have them in the script right now is so the same script can be used for SLES when CHPL_LLVM=llvm.

I don't think there's any harm in including them when they are not needed, so either way is fine with me: conditionally loading python 2.7 and cmake, or always loading them.

This looks fine to me.  OK to commit after @ronawho is satisfied.

util/cron/common-llvm.bash
dmk42: I agree the re-export is not needed.

ronawho:
Oh, I had missed that we weren't sourcing python 2.7/cmake for the system case.. 

I'd be ok with sourcing them anyways even though they aren't needed, or i'd be ok with just wrapping that block with an `if CHPL_LLVM != system` instead of the case

awallace-cray:
Originally planned to have the LLVM==system branch test if LLVM 4.* was installed, then decided against doing that. I think that was why that case stmt didn't look right.    
Now I want to put that LLVM 4.* test in, after all. 
W.I.P.

awallace-cray:
I tested the CHPL_LLVM=system build & test on Ubuntu 64, after removing+installing LLVM and Clang 3.8 and 4.0. 

I think the existing error messages from Chapel's Makefiles are explanatory enough, and will not add a redundant check+warning to this test script.

awallace-cray:
Ready to merge this. All OK?
Thanks

dmk42:
LGTM

util/cron/test-gasnet.llvm.bash
```diff
@@ -5,7 +5,7 @@
 
 CWD=$(cd $(dirname $0) ; pwd)
 source $CWD/common-gasnet.bash
-source $CWD/common-llvm.bash
+source $CWD/common-llvm.bash llvm
```
ronawho: I have a slight preference for not modifying scripts that are testing `CHPL_LLVM=llvm`, but I won't push for that if you think it's worthwhile for some reason. 

If you do want to modify them, please update `util/cron/test-linux64-vm.bash` too

awallace-cray: The reason I changed the other llvm scripts was just to help remind us (next time we copy-paste one of these test-llvm.bash things) that common-llvm can now set up for LLVM=llvm or LLVM=system via the new commandline parameter. This was not a strong reason, so we'll just forget it.

PR #7920, opened 2017-12-02
Filing a future against copy initializers with default args

bradcray:
This pair of tests demonstrates that if an initializer meets
the copy initializer requirements, but also has additional
arguments, it is not considered.  The second test shows an
unfortunate workaround that can be used in which a second
copy initializer is added to work around the issue.


bradcray:
@lydia-duncan and @mppf: This looks like it's a bug to me, but I wanted to get one or both of you to verify before merging it and opening an issue against it.  Thanks!

bradcray:
(This was motivated by trying to convert the `string` constructor to an initializer)

test/types/records/init/copy/copyWithDefaultArgs.chpl
```diff
@@ -0,0 +1,31 @@
+record R {
+  var x: int;
+
+  proc init(x: int, y: bool = true) {
+    if y then
+      this.x = x;
+  }
+
+  proc init(r: R, y: bool = true) {
```
mppf: It seems odd to me that there are two duplicate copies of `init` here - wouldn't that lead to ambiguity if everything was "working" ?

mppf:
I've also had trouble when I tried to use copy initializers in practice but I'm not sure what's going wrong.

test/types/records/init/copy/copyWithDefaultArgs.chpl
lydia-duncan: Not for these two, you can distinguish them based on the argument without a default value (one only takes an int for the first argument, the other only takes an instance of R for the first argument)

mppf: Ah, I missed that for some reason. Thanks.

lydia-duncan:
I agree that this is a bug on copy initializers, they should allow an initializer where all additional arguments have a default value.

PR #7921, opened 2017-12-02
Futures for conditional expressions in fields when using initializers

bradcray:
These futures show that when using conditional expressions to declare
the type or initial value of a field, problems occur when the type
uses an initializer.  (I ran into these cases while trying to convert
the 'range' type over from constructors to initializers)


bradcray:
@lydia-duncan and @noakesmichael : This seems likely to be a known issue, but is something I ran into when working on converting ranges over from constructors to initializers.  I didn't experiment to see how specific it is to generic types.  Let me know if there's an existing issue for this or whether I should file one.

lydia-duncan:
Oh, interesting.  I've seen the error before (along with a similar one for parallel loop expressions), but I thought it was limited to default initializers.  Good to know that it is impacting both compiler-generated and user-defined initializers.

There isn't an issue open for this currently (it's just in my list of known failure modes for default initializers), so you can file one if you want to.

lydia-duncan:
And these tests are good to merge!

PR #7923, opened 2017-12-02
Replaced an autoCopy with PRIM_INIT_VAR

vasslitvinov:
Looks like the right way to initialize a variable is via PRIM_INIT_VAR, rather than with an autoCopy.
Making this change for the new implementForallIntents2.

There are two autocopies still in the old implementForallIntents. That code needs to go away altogether, so I am leaving those autocopies alone.

Testing: valgrind in quickstart configuration, with --verify.


vasslitvinov:
@lydia-duncan or @noakesmichael - am I right about PRIM_INIT_VAR ?

(The code I modified runs during resolveFunction() right before it gets resolved.)

@mppf - FYI. I think this PR is an improvement over #7849 and handles identity() returning by ref.

lydia-duncan:
I know nothing about PRIM_INIT_VAR beyond that it exists, sorry.

PR #7926, opened 2017-12-02
Always append to nightly_args in nightly testing scripts

ronawho:
Always append to nightly_args instead of overwriting for nightly test scripts.
None of the scripts I'm changing were sourcing scripts that set nightly_args so
this shouldn't have any impact on testing. This is just intended to prevent
issues like we saw in #7925 from happening in the future


ronawho:
@awallace-cray mind taking a look at this?

PR #7928, opened 2017-12-04
Skip tryCResolve in scopeResolve if there are no extern C blocks

mppf:
This addresses a slowdown in scopeResolve for CHPL_LLVM=llvm compiles
relative to CHPL_LLVM=none compiles, when extern blocks are not used.
It would be better to further improve scopeResolve for the case in which
an extern block is used in one module, or to generally improve the performance
of this feature. However this is a quick and easy performance fix.

- [x] full local CHPL_LLVM=llvm testing

### DIFF
--- a/compiler/passes/scopeResolve.cpp
+++ b/compiler/passes/scopeResolve.cpp
@@ -659,7 +659,8 @@ static void resolveUnresolvedSymExpr(UnresolvedSymExpr* usymExpr) {
     updateMethod(usymExpr);
 
 #ifdef HAVE_LLVM
-    if (tryCResolve(usymExpr->getModule(), name) == true) {
+    if (gExternBlockStmts.size() > 0 &&
+        tryCResolve(usymExpr->getModule(), name) == true) {
       // Try resolution again since the symbol should exist now
       resolveUnresolvedSymExpr(usymExpr);
     }


mppf:
@ronawho - I think this is "trivial" but would appreciate it if you can have a quick look. Thanks.

ronawho:
This seems like an ok workaround to me.

I was hoping that the real solution would be something like "stop calling tryCResolve so many times", but this is easy. If you have a single extern block, then scopeResole time will double again, right?

mppf:
@ronawho - yes, a single extern block will slow it down again. As you said, this is the easy fix. I agree a deeper solution would be better. I think we could track which modules had any extern blocks on their path to the root module. It might get a little weird / tricky to track with nested modules, though. Anyway, any module that doesn't have extern blocks and isn't contained in (or using?) a module using extern blocks doesn't need to call tryCResolve.

Also, it wouldn't surprise me if a small amount of optimization effort could make scopeResolve or tryCResolve significantly faster. But my personal reaction is that I still don't understand why we have scopeResolve at all...

ronawho:
>Also, it wouldn't surprise me if a small amount of optimization effort could make scopeResolve or tryCResolve significantly faster. But my personal reaction is that I still don't understand why we have scopeResolve at all...

👍  agreed

Issue #5348, opened 2017-02-11
Add support for tab completion for compiler and testing scripts


ronawho:
Not having tab completion is such a drag

I have no idea what's involved to support bash_completion files, but http://askubuntu.com/questions/68175/how-to-create-script-with-auto-complete might be a good lead.

For our testing scripts that use argparse, we might be able to take advantage of argcomplete: https://argcomplete.readthedocs.io/en/latest/

ronawho:
https://github.com/chapel-lang/chapel/pull/7929 added support for compiler autocompletion

ben-albrecht:
I got the args for `mason` with these commands, but creating the mason completion script will require some further thought due to handling subcommands (not accounted for here):

```bash
optional_abbr=`mason --help | grep -v '^[A-Za-z]' | grep '^.*--' |        \
        sed -e 's;^\(.*--[^ ]*\).*$;\1;' | grep ',' | sed -e 's;,.*$;;' | \
        xargs echo`

optional=`mason --help | grep -v '^[A-Za-z]' | grep '^.*--' |      \
          sed -e 's;^\(.*--[^ ]*\).*$;\1;' | sed -e 's;-[A-Za-z],;   ;' | \
          xargs echo`

arg=`mason --help | grep -v '^[A-Za-z]' | grep -v '^.*--'  | \
     grep -v '<command>' | grep -v '\[options\]'  | sed 's/^[ \t]*//' | \
     cut -d" " -f1 | xargs echo`
```

PR #7929, opened 2017-12-04
Add a bash autocompletion script for chpl options

daviditen:
Sourcing this script enables tab-autocompletion for Chapel compiler options in
bash. It contains a static list of compiler options. The commands used to
generate the lists are included as comments. It should be possible to generate
these lists on the fly, but I didn't want to require a working chpl to be in
the path.


ronawho:
👍 Awesome!

I'm assuming you've already seen this, but if not there's an issue related to this: https://github.com/chapel-lang/chapel/issues/5348

BTW I think it'd be nice to add test to make sure this stays up to date

daviditen:
I've thought a fair bit about testing it to keep it up to date, but so far haven't come up with anything that would work well.  I was hoping there was some command I could run that would answer a question like "What would the possible matches be if I typed `chpl -<tab><tab>`?", but I haven't found anything like that so far.

ronawho:
I was thinking of something that would take the sed commands listed in the file and use them to generate a "current" version of the file, then diff it against util/chpl-completion.bash. If the files are the same, great. If not, then probably some compilation option has changed and ` util/chpl-completion.bash` needs to be regenerated. 

i.e. maybe add "gen-chpl-bash-completion" and "chpl-completion-template" files that will produce chpl-completion.bash, and add a test that generates a new one and compares to the checked in copy.

daviditen:
@ronawho I added generation scripts and a test as you suggested.  Could you please review?

test/chplenv/autocomplete/autocomplete.chpl
```diff
@@ -0,0 +1,27 @@
...
+/* This test runs a script to generate a bash autocomplete script from
+   the output of `chpl --help --devel`.  It diffs the generated script
+   against the version in $CHPL_HOME/util/chpl-completion.bash.  If this
+   test produces any output it probably means that an option has been
+   added or removed from `chpl --help --devel`, so the autocomplete
+   script should be regenerated.
+ */
+
+var homeC: c_string;
+sys_getenv("CHPL_HOME", homeC);
```
ronawho: Any reason to not just use the builtin `CHPL_HOME`? (https://chapel-lang.org/docs/latest/builtins/internal/ChapelEnv.html#ChapelEnv.CHPL_HOME)

daviditen: No. Good call - I'll update that.

util/chpl-completion.bash
```diff
@@ -0,0 +1,72 @@
...
+# https://superuser.com/questions/564716/bash-completion-for-filename-patterns-or-directories
+
+_chpl ()
+{
+  local cur
+  COMPREPLY=()
+  cur=${COMP_WORDS[COMP_CWORD]}
+  case "$cur" in
+    -*)
+      # The following lists were generated with the commands preceding them.
```
ronawho: Maybe remove these comments and just direct people to util/devel/gen-chpl-bash-completion? (I doubt we'll remember to update these comments whenver that script gets updated)

ronawho:
Looks great to me!

Some followups that I think would be nice, but don't need to be in this PR:

I think we should consider adding this to gen_release, documentation, and maybe have the setchplenv* scripts source it.

If we release it, I think that also raises the question of whether we want users to see the "developer" options. Maybe we could make a -devel version (or maybe the script can key off of whether `CHPL_DEVELOPER` is set?)

util/chpl-completion.bash
daviditen: Sounds good.  I'll just remove the comments since there's already a comment at the top of the script pointing at `gen-chpl-bash-completion`.

daviditen:
Thanks!

I think all of those followups sound good.

If we don't want to show `--devel` options in `--no-devel` mode we can make it choose the right mode by looking at both `--devel`/`--no-devel` being on the command line and the `CHPL_DEVELOPER` environment variable.

PR #7930, opened 2017-12-04
Convert locale models over to use init() rather than constructors

bradcray:
This change converts locale models over to use init() rather than
constructors.  For the most part this was straightforward, though for
the current compiler the 'parent' field seems to only be able to be
initialized in phase 2 (in spite of seemingly being 'const'?) and this
required using the `_parent` formal argument in a few cases that
wanted to use the field in phase 1 prior to its being initialized.

### DIFF
--- a/modules/internal/localeModels/flat/LocaleModel.chpl
+++ b/modules/internal/localeModels/flat/LocaleModel.chpl
@@ -63,14 +63,14 @@ module LocaleModel {
     // that it is intended to represent.  This trick is used
     // to establish the equivalence the "locale" field of the locale object
     // and the node ID portion of any wide pointer referring to it.
-    proc LocaleModel() {
+    proc init() {
       if doneCreatingLocales {
         halt("Cannot create additional LocaleModel instances");
       }
       setup();
     }
 
-    proc LocaleModel(parent_loc : locale) {
+    proc init(parent_loc : locale) {
       if doneCreatingLocales {
         halt("Cannot create additional LocaleModel instances");
       }
@@ -161,7 +161,7 @@ module LocaleModel {
     const myLocaleSpace: domain(1) = {0..numLocales-1};
     var myLocales: [myLocaleSpace] locale;
 
-    proc RootLocale() {
+    proc init() {
       parent = nil;
       nPUsPhysAcc = 0;
       nPUsPhysAll = 0;
--- a/modules/internal/localeModels/apu/LocaleModel.chpl
+++ b/modules/internal/localeModels/apu/LocaleModel.chpl
@@ -59,13 +59,14 @@ module LocaleModel {
     }
     proc chpl_name() return name;
 
-    proc CPULocale() {
+    proc init() {
     }
 
-    proc CPULocale(_sid, _parent) {
+    proc init(_sid, _parent) {
       sid = _sid;
+      name = _parent.chpl_name() + "-CPU" + sid;
+      super.init();
       parent = _parent;
-      name = parent.chpl_name() + "-CPU" + sid;
     }
 
     proc writeThis(f) {
@@ -95,9 +96,9 @@ module LocaleModel {
     }
     proc chpl_name() return name;
 
-    proc GPULocale() {
+    proc init() {
     }
-    proc ~GPULocale() {
+    proc deinit() {
      extern proc hsa_shutdown(): void ;
 
      // Comment out this until runtime support for HSA is added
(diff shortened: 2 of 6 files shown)


bradcray:
@lydia-duncan : The changes in this branch cause new candidates to be listed for your test/classes/initializers/where-clause4.chpl test for reasons that aren't immediately obvious to me.  My first instinct was to just update the .good file, but the problem is that the candidates vary depending on which CHPL_LOCALE_MODEL is being used.  Do you understand this behavior and have any suggestions for working around it?

@gbtitus : How do you suggest testing the 'numa' locale model?  I ran on a few tests and looked for ones whose skipifs were specifically designed for 'numa', but maybe I should just do a full run?

@dmk42 : What's the best practice for testing changes to the KNL locale model?

@mlchu : What's the best practice for testing changes to the APU locale model?

lydia-duncan:
The new candidates for where-clause4.chpl is expected.  One of our longstanding TODOs for initializers is to limit the candidates to only functions with their "this" argument in the same hierarchy instead of all "init" methods.  That test is tracking that desire/effort.  Most of the tests I've written that run into this problem use a prediff I've written to remove the library/internal module candidates - I need one test that ensures we're still getting too many candidates.  I think your best bet is to add a skipif for alternate locale models, update the .bad for the locale model you've chosen, and note in the README that it should be removed once those candidates are no longer considered viable (as we will stop getting library output then).

dmk42:
@bradcray : KNL testing has become more complicated now that it has moved to a non-slurm machine.  I will send you details by e-mail.

mlchu:
@bradcray you raise a good question. This change itself seems fine to me. I have an APU machine I sitting in Austin that we are planning to eventually send your guys' way so that you can test it regularly. I've been waiting until I can get more of my code checked in. Right now the APU model doesn't really do all that much other than create a CPU and GPU sublocale without the rest of my changes.

gbtitus:
@bradcray I think you can just run examples, or maybe even just the hellos plus examples/benchmarks/hpcc, say.  Possibly also do perf runs of stream-ep and ssca2 and compare to nightly numa perf.  This seems like a change that ought to either work, or fail pretty obviously.

dmk42:
Just to document our offline discussion:  We agreed to let nightly testing handle the KNL case, mainly due to the newly-increased work to do KNL testing.

bradcray:
Based on the feedback, I think I've done enough testing on this now.  If you haven't already, could one or more of you locale model authors (@dmk42, @gbtitus) / init champions (@lydia-duncan) give this a proper review?  (or let me know if you think you already have).  Thanks!

gbtitus:
This looks good to me, but I reviewed it fairly mechanically since I'm by no means an initialization expert.  It might be a good idea for @lydia-duncan or @dmk42 to also take a look.

dmk42:
After some offline discussion, I'm now convinced that this will work in the short term.  There are some issues with converting `locale` to use an initializer instead of a constructor, so I'll approve this from the KNL locale model point of view and leave it up to @lydia-duncan whether to tackle the `locale` problem now or wait until later.

modules/internal/localeModels/flat/LocaleModel.chpl
```diff
@@ -63,14 +63,14 @@ module LocaleModel {
     // that it is intended to represent.  This trick is used
     // to establish the equivalence the "locale" field of the locale object
     // and the node ID portion of any wide pointer referring to it.
-    proc LocaleModel() {
```
lydia-duncan: Am I reading this correctly that this is another instance of a class type sharing the same name as the module in which it is defined?

dmk42: Correct.  Filename: LocaleModel.chpl, Module name: LocaleModel, class name: LocaleModel.

lydia-duncan: We've run into problems with this in the past, and have updated other types/modules to have a different name.  Do we think that's something we should change here or do we want to wait until that actually becomes a problem?

dmk42: I have no objection to being proactive and taking care of it now.

bradcray: I think taking care of it now is smart, but that we should do it in a distinct PR since it isn't directly related to this one.

bradcray:
Offline, @lydia-duncan and I have been exploring what it would take to convert the parent classes over to initializers as well, which seems like the more principled thing to do than the current reliance on default constructors in the parents and an abuse of assigning a const field (`parent`) in phase 2 (a case that the compiler seems not to be catching).

lydia-duncan:
#7943 is now open to follow up on this change

PR #7931, opened 2017-12-04
Free c_strings allocated by IO runtime

benharsh:
The first leak is resolved by updating SysErr.chpl to correctly take ownership of c_strings by calling string initializers instead of casting to `string`.

The second leak is resolved by manually freeing a couple of c_strings allocated by QIO within the runtime.

Testing:
- [x] full local
- [x] full gasnet
- [x] memleaks (~8 tests no longer leak)


benharsh:
@mppf : can you review?

PR #7943, opened 2017-12-05
Convert locale models to use initializers

lydia-duncan:
Convert the LocaleModel hierarchy to use initializers instead of
constructors or default constructors.  Modifications are fairly
straight-forward, and required an update to one test for the new
expected output.

This is a follow-on/replacement to #7930 that also converts the parent
and grandparent classes (AbstractRootLocale and locale), as well as
their other inherited types, to use initializers as well.

Passed a full std paratest, passed hello.chpl and test/localeModels/numa
for numa.  Will punt on KNL testing barring fearful exclamations from
@dmk42, and APU testing barring fearful exclamations from @mlchu.


lydia-duncan:
Look good, @dmk42, @bradcray and @mlchu?

bradcray:
I gave it only the quickest of looks, but this is definitely along the lines of what I was hoping to write / see!

modules/internal/localeModels/knl/LocaleModel.chpl
```diff
@@ -287,6 +285,9 @@ module LocaleModel {
       const origSubloc = chpl_task_getRequestedSubloc();
       chpl_task_setSubloc(hbm_id);
       hbm = new MemoryLocale(hbm_id, this);
+
+      super.init(_parent);
+
       chpl_task_setSubloc(origSubloc);
```
dmk42: I believe the super.init() should be moved down one line so we will be in the original sublocale when it is called.  Note that we only change the sublocale for one brief moment when we create the MemoryLocale for the high-bandwidth memory, and then we change it back.

I missed this in the original PR because I was focused on understanding what the super.init() did.  Sorry about that.

mlchu:
It looks fine to me.

PR #7947, opened 2017-12-06
Add a function to updateDatFiles.py that can remove "bogus" values

ronawho:
Add a replace_values_over_threshold method, which can be used to fix timings
that are over a certain threshold by replacing the bogus data with the last
valid data point. This is what I'm using to replace the bogus "90 year" timings
in our compiler performance .dat files. Note that the source of the bogus
timings was fixed by https://github.com/chapel-lang/chapel/pull/7942


ronawho:
@ben-albrecht quick sanity check?

Note that this script is from my early days of python, so it could use some cleanup but I don't feel like tackling that here

PR #7958, opened 2017-12-06
Make the autocomplete test work even if chpl is not in PATH

daviditen:
Some of our testing appears to run with chpl not in PATH. Make this test set
PATH to include "CHPL_HOME/bin/CHPL_HOST_PLATFORM" before it tries to use chpl.

Tested with PATH including chpl on Cygwin64, Linux64, OSX, Valgrind.
Tested with PATH not including chpl on Cygwin64, Linux64, OSX, Valgrind.

### DIFF
--- a/test/chplenv/autocomplete/autocomplete.chpl
+++ b/test/chplenv/autocomplete/autocomplete.chpl
@@ -10,6 +10,12 @@ use Spawn;
 
 var home = CHPL_HOME;
 
+extern proc getenv(name: c_string): c_string;
+extern proc setenv(name: c_string, val: c_string, overwrite: c_int): c_int;
+
+var path = getenv("PATH"):string;
+setenv("PATH", (path + ":" + home + "/bin/" + CHPL_HOST_PLATFORM).c_str(), 1);
+
 var genScript = home + "/util/devel/gen-chpl-bash-completion";
 var completeScript = home + "/util/chpl-completion.bash";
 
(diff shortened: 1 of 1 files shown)


daviditen:
@ronawho could you please review this autocomplete test update?  It should fix the failures that we saw in nightly testing by making sure `chpl` is in the `PATH`.

ronawho:
Shoot, I never remember that `chpl` isn't in our PATH for nightly testing. I wonder if we should just update start_test to put it in the PATH

I think the fix in https://github.com/chapel-lang/chapel/pull/6975 is a little cleaner (just call abs path to chpl instead of putting it in the path) since you don't need to extern to getenv/setenv

daviditen:
The test only calls `chpl` through the `gen-chpl-bash-completion` script, so I'd need to modify that script to take an argument or something similar to match #6975, but that would make the script less nice to call by hand.

daviditen:
...And I think updating the testing scripts to put `chpl` into PATH would be nice.

ronawho:
Oh, duh. Yeah, you're right. In that case we could go with something like https://github.com/chapel-lang/chapel/blob/master/test/patterns/cltools/fileCompile/chplCompile.execenv, which just does the path setting in an execenv. It's not really any different other than it saves us from externing in the chpl code.

I have a slight preference for the execenv approach, though I'm not really sure why.

ronawho:
I'll take a look at what it would take to add `chpl` into the `PATH` in start_test

daviditen:
My preference is a little in the other direction.  I prefer making it not depend on the testing system so it can be run by hand:

```
$CHPL_HOME/bin/$CHPL_HOST_PLATFORM/chpl autocomplete.chpl
./autocomplete
```

Would you be OK with leaving it as-is?

ronawho:
I would imagine chpl will be in your path if you're going to compile manually. 

That said, I don't feel very strongly about any of this. I have no problem with this going in as-is.

daviditen:
Thank you!

PR #7965, opened 2017-12-06
Add an error message for constructor when any parent type defines an init()

lydia-duncan:
The default constructor relies on calling the parent type's default constructor.
If any type we are inheriting from defines an initializer, we will run into
problems.  It isn't worth extending constructors to support inheriting from such
types, so just generate a friendly compiler error.

Also adds two tests to cover the new error message for inheritance, covering
both the default constructor world and the default initializer world.

Passed classes/initializers with futures, and a full std/ paratest


bradcray:
This made working through issues converting constructors to initializers much nicer, thank you!

PR #7971, opened 2017-12-07
Eliminate memory leaks in reductionex_record_class.chpl

vasslitvinov:
Each instance of `myhisto` allocates an instance of `histo` and so
should delete it. Add a deinitializer to do so.

The one exception is when generate() is invoked. Which happens
for the top-level reduction op object that produces the value
of the reduction for consumption by the user.
In that scenario, pass ownership to the user and do not
delete it in the deinitializer.

Passes valgrind. -memLeaks reports no leaks.


vasslitvinov:
@benharsh would you review?

Issue #6547, opened 2017-06-27
chpl-env-gen.h contains non-runtime defines


mppf:
chpl-env-gen.h is stored in the runtime build directory, but it contains some information that is not encoded in the runtime build directory.

E.g. in one configuration, I have this runtime build directory:

`../build/runtime/linux64/gnu/arch-native/loc-flat/comm-none/tasks-qthreads/tmr-generic/unwind-none/mem-jemalloc/atomics-intrinsics/gmp/hwloc/re2/wide-struct/fs-none/`

and this in chpl-env-gen.h:

```
#define CHPL_HOST_PLATFORM_LINUX64
#define CHPL_HOST_COMPILER_GNU
#define CHPL_TARGET_PLATFORM_LINUX64
#define CHPL_TARGET_COMPILER_GNU
#define CHPL_TARGET_ARCH_NATIVE
#define CHPL_LOCALE_MODEL_FLAT
#define CHPL_COMM_NONE
#define CHPL_COMM_SUBSTRATE_NONE
#define CHPL_GASNET_SEGMENT_NONE
#define CHPL_TASKS_QTHREADS
#define CHPL_LAUNCHER_NONE
#define CHPL_TIMERS_GENERIC
#define CHPL_UNWIND_NONE
#define CHPL_MEM_JEMALLOC
#define CHPL_JEMALLOC_JEMALLOC
#define CHPL_MAKE_GMAKE
#define CHPL_ATOMICS_INTRINSICS
#define CHPL_NETWORK_ATOMICS_NONE
#define CHPL_GMP_GMP
#define CHPL_HWLOC_HWLOC
#define CHPL_REGEXP_RE2
#define CHPL_WIDE_POINTERS_STRUCT
#define CHPL_LLVM_LLVM
#define CHPL_AUX_FILESYS_NONE
```

Where in particular the following are NOT represented in the runtime build path:

`../build/runtime/linux64/gnu/arch-native/loc-flat/comm-none/tasks-qthreads/tmr-generic/unwind-none/mem-jemalloc/atomics-intrinsics/gmp/hwloc/re2/wide-struct/fs-none/`

and this in chpl-env-gen.h:

```
CHPL_HOST_PLATFORM_LINUX64
CHPL_HOST_COMPILER_GNU
CHPL_LAUNCHER_NONE
CHPL_MAKE_GMAKE
CHPL_LLVM_LLVM
```

[...]

PR #2929, opened 2015-11-19
Fix tests for ChapelEnv Module

ben-albrecht:
This patch sets out to fix the tests that broke in nightly testing:

* No longer passing `CHPL_TARGET_ARCH` to `printchplenv` when using Cray programming environment
* Initialize first element of `stringBuffer` in `stringBufferInit()` to please the valgrind god (Odin?)
* Calling `genGlobalString` for CHPL_vars in `codegen.cpp` so that CHPL_vars are initialized in `chpl_compilation_config.c`
* Bash functions are now POSIX-compliant to please our linux32 test box.

**[ √ paratested]**

PR #7976, opened 2017-12-07
New printchplenv interface

### LINKED ISSUE #7827: Proposal to change printchplenv interface
PR #7757 changed the behavior of printchplenv to enable --llvm compiles to go faster. But, it left some open questions about printchplenv modes and more generally how to adjust that script so it's less confusing to maintain.

The below proposed interface for printchplenv came out of a discussion I had with @ben-albrecht that is related to PRs #7815 and #7757. In particular, the present problem is that `printchplenv --simple` is [documented as a way to produce chplconfig files](https://chapel-lang.org/docs/master/usingchapel/chplenv.html#generating-configuration-files)  but now outputs additional variables (for use by the Chapel compiler) that aren't appropriate for chplconfig files. A less severe, but still relevant problem, is that `printchplenv --debug` now outputs these same extra variables.

It seems that we need to add a mode to `printchplenv` in order to resolve these issues, but the script is already unreasonably tangled and confusing, and so it seemed time to think through how it might more reasonably be structured and what might be a better set of command line options for the script to support.

The proposal below preserves the following behaviors:
 * printchplenv with no arguments works exactly as it does today
 * printchplenv --simple and printchplenv --overrides can still produce chplconfig files

The proposal below makes the following changes in order to better structure the options:
 * enables separate selection of what variables to print / how to filter / output format
 * removes --sh and --csh modes (these are no longer needed now that we have chplconfig)
 * removes --debug in favor of --all
 * printchplenv --simple and printchplenv --overrides no longer outputs CHPL_LLVM by default (to keep the option sets & defaults reasonable. I expect this to be low impact because CHPL_LLVM is automatically enabled if it is built, but now users could use e.g. --simple --all)

Note that `chpl` and Makefiles can still get the output they are used to now
 * Makefiles would use `printchplenv --all --internal --no-tidy --make`
 * chpl compiler would use `printchplenv --all --internal --no-tidy --simple`

### Proposed printchplenv -h output

```
Usage: printchplenv [options]

Print the current Chapel configuration. Arguments allow selectiion of
what gets printed [content], how it gets printed [format], and what gets
filtered [filter].

The default contents are: --runtime --launcher

Options:
  -h, --help   show this help message and exit

  [shortcut]
  --all        shortcut for --compiler --runtime --launcher

  [content]
  --compiler    Select variables describing the configuration of the compiler
  --runtime     Select variables describing the configuration of the runtime
  --launcher    Select variables describing the configuration of the launcher
  --internal    Select additional variables used during builds

  [filter]
  --[no-]tidy  (default) [don't] omit sub-variables irrelevant to the current
                configuration
  --anonymize  avoid printing machine specific details such as host name and
               file paths
  --overrides  only show variables that have been user supplied
               via environment or chplconfig

  [format]
  --pretty     (default) print variables in format: CHPL_KEY: VALUE
               indicating which options are set by environment variables (*)
               and which are set by configuration files (+)
  --simple     print variables in format: CHPL_KEY=VALUE
               output is compatible with chplconfig format
  --make       print variables in format: CHPL_MAKE_KEY=VALUE
  --path       print variables in format: VALUE1/VALUE2/...
```

####  --compiler    Select variables describing the configuration of the compiler
    CHPL_HOST_PLATFORM
    CHPL_HOST_COMPILER
    CHPL_LLVM
    CHPL_MAKE

####  --runtime     Select variables describing the configuration of the runtime
    CHPL_TARGET_PLATFORM
    CHPL_TARGET_COMPILER
    CHPL_TARGET_ARCH
    CHPL_LOCALE_MODEL
    CHPL_COMM
    CHPL_COMM_SUBSTRATE
    CHPL_GASNET_SEGMENT
    CHPL_TASKS
    CHPL_TIMERS
    CHPL_UNWIND
    CHPL_MEM
    CHPL_JEMALLOC
    CHPL_MAKE
    CHPL_ATOMICS
    CHPL_NETWORK_ATOMICS
    CHPL_GMP
    CHPL_HWLOC
    CHPL_REGEXP
    CHPL_AUX_FILESYS

####  --launcher    Select variables describing the configuration of the launcher

```
CHPL_HOST_PLATFORM
CHPL_HOST_COMPILER
CHPL_LOCALE_MODEL
CHPL_COMM
CHPL_TASKS
CHPL_LAUNCHER
CHPL_TIMERS
CHPL_UNWIND
CHPL_MEM
CHPL_ATOMICS
```

####  --internal    Select additional variables used during builds
    CHPL_JEMALLOC
    CHPL_HOST_MEM
    CHPL_ORIG_TARGET_COMPILER
    CHPL_RUNTIME_SUBDIR
    CHPL_LAUNCHER_SUBDIR
    CHPL_COMPILER_SUBDIR
    CHPL_GMP_UNIQ_CFG_PATH
    CHPL_HWLOC_UNIQ_CFG_PATH
    CHPL_JEMALLOC_UNIQ_CFG_PATH
    CHPL_LIBUNWIND_UNIQ_CFG_PATH
    CHPL_MASSIVETHREADS_UNIQ_CFG_PATH
    CHPL_QTHREAD_UNIQ_CFG_PATH
    CHPL_RE2_UNIQ_CFG_PATH
    CHPL_THIRD_PARTY_LINK_ARGS

ben-albrecht:
Clean up `printchplenv` implementation and interface to separate logic of what gets printed vs. how it gets printed.

Closes #7827

The help output describes the new interface:

```
Usage: printchplenv [options]

Print the current Chapel configuration. Arguments allow selectiion of
what gets printed [content], how it gets printed [format], and what gets
filtered [filter].

The default [content] arguments are: --runtime --launcher

Options:
  -h, --help    Show this help message and exit

  [shortcut]
  --all         Shortcut for --compiler --runtime --launcher

  [content]
  --compiler    Select variables describing the configuration of the compiler
  --runtime     Select variables describing the configuration of the runtime
  --launcher    Select variables describing the configuration of the launcher
  --internal    Select additional variables used during builds
                 this flag is incompatible with [format]: --path

  [filter]
  --[no-]tidy   (default) [don't] Omit sub-variables irrelevant to the current
                 configuration
  --anonymize   Omit machine specific detailss, script location, and CHPL_HOME
  --overrides   Omit variables that have not been user supplied via environment
                 or chplconfig

  [format]
  --pretty      (default) Print variables in format: CHPL_KEY: VALUE
                 indicating which options are set by environment variables (*)
                 and which are set by configuration files (+)
  --simple      Print variables in format: CHPL_KEY=VALUE
                 output is compatible with chplconfig format
  --make        Print variables in format: CHPL_MAKE_KEY=VALUE
  --path        Print variables in format: VALUE1/VALUE2/...
                 this flag always excludes CHPL_HOME and CHPL_MAKE
```

### Old Interface vs. New Interface

| printchplenv today | printchplenv with this PR          |
|------------------|----------------------------|
| (no args)        | (no args)                  | 
| `-h, --help`       | `-h --help`                  |
| `--simple`         | `--simple`  | 
|` --overrides`      | `--simple --overrides` |
| `--anonymize`      | `--anonymize`                |
| `--make`           | `--all --internal --make`    |
| `--compiler`       | `--compiler --path`          |
| `--runtime`        | `--runtime --path`           |
|` --launcher`       | `--launcher --path`          |
| `--sh`             | (no longer available)          |
| `--csh`            | (no longer available)          |
| `--debug`          | (no longer available)          |

### Changes to `--simple` and `--simple --overrides`

The new `--simple` and `--simple --overrides` flags will omit some variables from previous implementation, since it is printing all variables for `--runtime --launcher` by default. `--simple` will also include `CHPL_HOME` by default now. Omitted variables are:

```
CHPL_HOST_PLATFORM
CHPL_HOST_COMPILER
CHPL_ORIG_TARGET_COMPILER
CHPL_COMM_SUBSTRATE
CHPL_GASNET_SEGMENT
CHPL_JEMALLOC
CHPL_NETWORK_ATOMICS
CHPL_LLVM
CHPL_RUNTIME_SUBDIR
CHPL_LAUNCHER_SUBDIR
```

This is not a big deal from a user perspective. Makefiles and other internal scripts that rely on `--simple` now call `--simple --internal --all --no-tidy --anonymize` to get all of the variables without `CHPL_HOME` included.

### New Chapel Environment Variables
Previously, the value of `CHPL_TARGET_ARCH` depended on:

- the target architecture
- if chapel is being used on a cray
- if `--make` flag is passed

In order to separate this logic, two new environment variables were added :

- `CHPL_RUNTIME_ARCH`
- `CHPL_TARGET_ARCH_ARG`

`CHPL_RUNTIME_ARCH` is the architecture the runtime is built for. When using a prebuilt Chapel  module on a Cray, It is set to the lowest common denominator (LCD) architecture available for usage by the runtime.

`CHPL_TARGET_BACKEND_ARCH` is set to the *cpu-type* to be used for the `-march` flag for passing to the compiler, e.g.  `ivybridge` maps to `core-avx-i` for gcc-4.7

## Testing

- [x] `linux64 paratest`

## Next Steps

Potential follow-up PRs & issues:

- Remove `useDefaultEnv` from `driver.cpp`
- Update `--launcher` output:
   -  Merge with `--runtime`
- Add a `[shortcut]` flag for `--internal --all --no-tidy` (commonly called internally)


mppf:
New variables: CHPL_RUNTIME_ARCH, CHPL_TARGET_ARCH_ARG

compiler/main/driver.cpp
```diff
@@ -1074,6 +1076,11 @@ bool useDefaultEnv(std::string key) {
     }
   }
 
+  // Always use default env for internal variables that could include spaces
+  if (key == "CHPL_THIRD_PARTY_LINK_ARGS") {
+    return true;
+  }
+
```
mppf: I don't think we need this anymore.... it was a workaround for CHPL_MAKE_ALL_VARS but now we don't even generate that anymore in the generated Makefile.

configure
```diff
@@ -144,13 +144,13 @@ fi
 # and then move it.
 # Note that the resulting chplconfig is expected to cover the same
 # material as any existing chplconfig.
-./util/printchplenv --overrides --anonymize > configured-chplconfig
+./util/printchplenv --simple --no-tidy --overrides --anonymize > configured-chplconfig
```
mppf: It seems to me that we want `--tidy` for configure - if a variable isn't relevant, wouldn't we prefer to leave the default value? (Additionally it seems hard to believe there would be an overridden variable that is not relevant.. but maybe I'm just having a failure of imagination.).

configure
```diff
@@ -144,13 +144,13 @@ fi
...
 mv configured-chplconfig chplconfig
 
 echo
 echo "  Currently selected Chapel configuration:"
 echo
-./util/printchplenv --anonymize
+./util/printchplenv --simple --no-tidy --anonymize
```
mppf: I don't think this should add `--simple` - we want it to be "pretty". Additionally, I'm not convinced --no-tidy is helpful here.

doc/rst/usingchapel/chplenv.rst
```diff
@@ -713,6 +713,11 @@ Below is an example of a Chapel configuration file with comments:
     CHPL_GMP=system
 
 
+To confirm the configuration file is written correctly, you can run
+``printchplenv --overrides``, which will show a list of variables that are
```
mppf: --all --overrides here?

doc/rst/usingchapel/chplenv.rst
```diff
@@ -721,26 +726,25 @@ Generating Configuration Files
...
+    printchplenv --all --simple --overrides > ~/.chplconfig
 
-The ``printchplenv --simple`` flag can be used to print all the variables
+The ``printchplenv --all --simple`` flag can be used to print all the variables
 of the current configuration. For example:
 
 .. code-block:: sh
 
-    printchplenv --simple > ~/.chplconfig
+    printchplenv --all --simple > ~/.chplconfig
```
mppf: It would be nice if the documentation here referred to printchplenv -h somewhere. I'm happy that printchplenv -h already says its output is compatible with chplconfig format.

runtime/Makefile
```diff
@@ -57,7 +57,7 @@ $(CHPL_ENV_HEADER):
 	@echo "/* THIS FILE IS AUTOMATICALLY GENERATED */" >> $(CHPL_ENV_HEADER)
 	@echo "#ifndef _CHPL_ENV_GEN_H_" >> $(CHPL_ENV_HEADER)
 	@echo "#define _CHPL_ENV_GEN_H_" >> $(CHPL_ENV_HEADER)
-	@$(CHPL_MAKE_HOME)/util/printchplenv --simple | \
+	@$(CHPL_MAKE_HOME)/util/printchplenv --all --no-tidy --anonymize --simple | \
```
mppf: I know that this used to work with the equivalent of --all, but actually it really really should not print out compiler variables. Can we change it to --runtime --no-tidy --anonymize --simple? That will solve issue #6547.

runtime/Makefile.help
```diff
@@ -138,7 +140,7 @@ $(RUNTIME_LIB): $(RUNTIME_OBJS) $(RUNTIME_DIR_TIMESTAMP) $(RUNTIME_LIST_INCLUDES
 	$(AR) -rc $@ $(RUNTIME_OBJS)
 	$(RANLIB) $@
 	$(TAGS_COMMAND)
-	../util/printchplenv --simple > $(RUNTIME_CHPLCONFIG)
+	../util/printchplenv --all --no-tidy --anonymize --simple > $(RUNTIME_CHPLCONFIG)
```
mppf: Like the previous runtime one, since this is generating a chplconfig stored in the *runtime* directory, it shouldn't include the launcher or compiler settings. So I think this should use --runtime instead of --all.

test/compflags/albrecht/chplenv/chplenv.chpl
```diff
@@ -12,13 +10,11 @@ writeln("CHPL_LAUNCHER=",CHPL_LAUNCHER);
 writeln("CHPL_TIMERS=",CHPL_TIMERS);
 writeln("CHPL_UNWIND=",CHPL_UNWIND);
 writeln("CHPL_MEM=",CHPL_MEM);
-writeln("CHPL_JEMALLOC=",CHPL_JEMALLOC);
```
mppf: Removed since CHPL_JEMALLOC is now "internal" (ie not printed for --all e.g.) right?

test/compflags/albrecht/chplenv/chplenv.chpl
```diff
@@ -12,13 +10,11 @@ writeln("CHPL_LAUNCHER=",CHPL_LAUNCHER);
...
 writeln("CHPL_MAKE=",CHPL_MAKE);
 writeln("CHPL_ATOMICS=",CHPL_ATOMICS);
 writeln("CHPL_NETWORK_ATOMICS=",CHPL_NETWORK_ATOMICS);
 writeln("CHPL_GMP=",CHPL_GMP);
 writeln("CHPL_HWLOC=",CHPL_HWLOC);
 writeln("CHPL_REGEXP=",CHPL_REGEXP);
-writeln("CHPL_LLVM=",CHPL_LLVM);
```
mppf: Removed since now by default we don't print any compile vars, even for --simple, right?

test/localeModels/gbt/maxTaskPar.prediff
```diff
@@ -40,7 +40,7 @@ case $target in
   linux*|cray-xc|cray-xe|cygwin*)
     case $locModel-$tasks in
       numa-qthreads)
-        ucp=$( $CHPL_HOME/util/printchplenv --make |
+        ucp=$( $CHPL_HOME/util/printchplenv --make --path |
```
mppf: I think we mean for this one to use $CHPL_HOME/util/printchplenv --all --internal--make 
For one thing --path overrides --make... and for another the grep is checking for an internal variable so we definitely need --internal.

test/runtime/gbt/topo/array-loc.prediff
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
-if $CHPL_HOME/util/printchplenv --make \
+if $CHPL_HOME/util/printchplenv --make --path \
```
mppf: Ditto other test: $CHPL_HOME/util/printchplenv --all --internal--make

test/runtime/gbt/topo/array-loc.suppressif
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if $CHPL_HOME/util/printchplenv --make \
+if $CHPL_HOME/util/printchplenv --make --path \
```
mppf: here too: $CHPL_HOME/util/printchplenv --all --internal--make

util/test/paratest.server
```diff
@@ -748,7 +748,7 @@ sub main {
     }
 
     # Package up the environment
-    my $tmpchplenv = `$ENV{CHPL_HOME}/util/printchplenv --sh`;
+    my $tmpchplenv = `$ENV{CHPL_HOME}/util/printchplenv --simple --all`;
```
mppf: At first I thought we needed to change tmpchplenv to store them like e.g. export CHPL_HOME='/some/path/' but now I see that stuff is removed a few lines down. Probably worth at least putting a comment indicating that the below removal is no longer relevant - or deleting it entirely.

e.g. replace the for loop with:
```
     for my $line (@lines) {
         my ($key,$value) = split(/=/, $line, 2);
         $chplenv .= "$key=$value ";
}
```

util/test/sub_test
```diff
@@ -257,7 +257,7 @@ def ReadFileWithComments(f, ignoreLeadingSpace=True):
     if os.access(f, os.X_OK):
         try:
             # grab the chplenv so it can be stuffed into the subprocess env
-            env_cmd = [os.path.join(utildir, 'printchplenv'), '--simple']
+            env_cmd = [os.path.join(utildir, 'printchplenv'), '--all', '--simple', '--no-tidy']
```
mppf: I think this should include --internal, in case some test wants a prediff to check CHPL_JEMALLOC (say) or some other internal variable.

util/test/testEnv
```diff
@@ -5,7 +5,7 @@ use Cwd 'abs_path';
 #
 # Set standard CHPL_* environment variables, if they are not already set.
 #
-my $env = `$ENV{CHPL_HOME}/util/printchplenv --sh`;
+my $env = `$ENV{CHPL_HOME}/util/printchplenv --all --simple --no-tidy`;
```
mppf: I think it should use --internal and also replace the following for loop.
```
 for my $line (@lines) {
   my ($key,$value) = split(/=/, $line, 2);
   next if $key eq "CHPL_HOME";
   $ENV{$key} = $value;
}
```

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+        seen = set()
+        link_args_3p_dedup = [arg for arg in reversed(link_args_3p)
+                              if not (arg in seen or seen.add(arg))]
+        ENV_VALS['  CHPL_THIRD_PARTY_LINK_ARGS'] = ' '.join(reversed(link_args_3p_dedup))
+
+
+"""Return non-empty string if var is set via environment or chplconfig"""
+def user_set(env):
+    env_set = overrides.get_environ(env, '')
+    config_set = overrides.get_chplconfig(env.strip(), '')
```
mppf: It seems odd that you needed env.strip only for the chplconfig version? I'd probably put env.strip at the start of this routine and use it for both.

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+    if env_set:
+        return ' *'
+    elif config_set:
+        return ' +'
+    return ''
+
+
+"""Filter variables that are not user set"""
+def filter_overrides(chpl_env):
+    return user_set(chpl_env.name)
```
mppf: I'd probably put 
```
return user_set(chpl_env.name) != ''
```

but that's partly my discomfort with Python magic.

compiler/main/driver.cpp
ben-albrecht: Just to be clear, are you referring to the if block or the entire `useDefaultEnv` function?

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+
+
+"""Return string to be printed for a given variable and print_format
+Requires a print_format argument
+"""
+def _print_var(env_var, value, print_format=None, shortname=None):
+    env = env_var.strip()
+    if print_format == 'pretty':
+        user_set_symbol = user_set(env)
+        return "{1}: {2}{0}\n".format(user_set_symbol, env_var, value)
```
mppf: I'd probably renumber this so the key/value are {0} and {1}.

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+"""
+def _filter_content(chpl_env, contents=None):
+    return chpl_env.content.intersection(contents)
+
+
+"""Return string to be printed for a given variable and print_format
+Requires a print_format argument
+"""
+def _print_var(env_var, value, print_format=None, shortname=None):
+    env = env_var.strip()
```
mppf: I think the choice of variable names 'env_var' and 'env' is confusing.
I'd probably name them "key" and "key_nospace" or "key_stripped".

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+    if print_format == 'pretty':
+        user_set_symbol = user_set(env)
+        return "{1}: {2}{0}\n".format(user_set_symbol, env_var, value)
+    elif print_format == 'simple':
+        return "{0}={1}\n".format(env, value)
+    elif print_format == 'make':
+        make_env_var = env_var.replace("CHPL_", "CHPL_MAKE_", 1).lstrip()
```
mppf: why not
```
make_env_var = env.replace("CHPL_", "CHPL_MAKE_", 1)
```

i.e. you don't need lstrip if we start out with the stripped variable name?

configure
ben-albrecht: I wasn't sure here, but assumed `--no-tidy` would be safer somehow. I am also failing to imagine a scenario where it would matter. I'll remove the `--no-tidy`.

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
-                stdout.write(value)
-            else:
-                stdout.write("{0}-{1}".format(short_name, value))
-    elif mode == 'shell':
-        stdout.write("export {0}='{1}'\n".format(env_var.strip(), value))
-    elif mode == 'cshell':
-        stdout.write("setenv {0} '{1}'\n".format(env_var.strip(), value))
-    else:
-        raise ValueError("Invalid mode '{0}'".format(mode))
+    if not recursive:
```
mppf: I'm not really in to this 'recursive' argument. I think I'd do this instead:
 * call populate_values before calling printchplenv in main
 * remove the call to populate_values here
 * remove the 'recursive' argument
 * change populate_values argument from contents to compute_internal that is True or False
 * set compute_internal in main to 'internal' in contents

This reduces the complexity of the control flow and more clearly shows the current assumption that a recursive call within printchplenv cannot add internal to contents when it wasn't there before.

test/compflags/albrecht/chplenv/chplenv.chpl
ben-albrecht: That is correct.

test/compflags/albrecht/chplenv/chplenv.chpl
ben-albrecht: Right.

util/printchplenv
```diff
@@ -39,260 +109,278 @@ def print_mode(mode='list', anonymize=False):
...
+
+
+def main():
+    (options, args) = parse_args()
+
+    # Handle --all flag
+    if options.all:
+        options.content.extend(['runtime', 'launcher', 'compiler'])
+
+    # Default [filter] = --tidy
```
mppf: Not really about the default - just say "Handle --tidy / --no-tidy "

compiler/main/driver.cpp
mppf: The block. I suspect we don't need useDefaultEnv either but am not ready to go that far yet. You added it in PR #2929, maybe you can figure out if we need it anymore.

I suspect that the factor out of CHPL_RUNTIME_ARCH will solve the issue.

util/printchplenv
```diff
@@ -1,36 +1,106 @@
 #!/usr/bin/env python
-import optparse, os, subprocess, sys
-from sys import stdout
 
-chplenv_dir = os.path.dirname(__file__)
-sys.path.insert(0, os.path.abspath(chplenv_dir))
+"""Usage: printchplenv [options]
+
+Print the current Chapel configuration. Arguments allow selectiion of
```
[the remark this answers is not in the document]
ben-albrecht: Should have used `chplspell`.... don't tell @cassella

util/printchplenv
ben-albrecht: How about:

```python
return bool(user_set(chpl_env.name))
```

mppf: Fine with me, if you think it makes sense...

util/printchplenv
ben-albrecht: Good call - I've opted for splitting `populate_values` into `compute_all_values()` and `compute_internal_values()`

compiler/include/driver.h
```diff
@@ -89,6 +89,8 @@ extern const char* CHPL_TARGET_PLATFORM;
 extern const char* CHPL_TARGET_COMPILER;
 extern const char* CHPL_ORIG_TARGET_COMPILER;
 extern const char* CHPL_TARGET_ARCH;
+extern const char* CHPL_RUNTIME_ARCH;
+extern const char* CHPL_TARGET_ARCH_ARG;
```
ronawho: This isn't a particularly meaningful name to me. I don't have any great suggestions though. Maybe `CHPL_TARGET_MAPPED_ARCH`?

Also maybe swap the ordering so it's next to `CHPL_TARGET_ARCH`?

compiler/util/files.cpp
```diff
@@ -635,12 +635,16 @@ void codegen_makefile(fileinfo* mainfile, const char** tmpbinname, bool skip_com
...
   {
-    if(!useDefaultEnv(env->first)) {
-      chplmakeallvars += env->first + "=" + std::string(env->second) + " ";
-    }
+    const std::string& key = env->first;
+    const char* oldPrefix = "CHPL_";
+    const char* newPrefix = "CHPL_MAKE_";
+    INT_ASSERT(key.substr(0, strlen(oldPrefix)) == oldPrefix);
+    std::string keySuffix = key.substr(strlen(oldPrefix), std::string::npos);
+    std::string chpl_make_key = newPrefix + keySuffix;
```
ronawho: Seems a little unfortunate to do the manual replacement instead of just using `printchplenv --make` (if we change the prefix or something we have 2 points of change), but since `printchplenv` takes a non-trivial amount of time to run, I can see the desire for this.

Also, is this code change just to prevent running `printchplenv` at generated code `make` time? If it is, can we split this into a different PR?

compiler/include/driver.h
mppf: I'm fine with CHPL_TARGET_MAPPED_ARCH if that's more satisfying to you. It doesn't say what it's mapped to, though...

compiler/util/files.cpp
mppf: @ronawho - it could be a separate PR, but then we'd have to do more workarounds in this PR, since we have to adjust how CHPL_MAKE_ALL_VARS is computed anyway (since it might now include variables with spaces). So it's not quite purely an optimization. Nonetheless we could do the adjustment to CHPL_MAKE_ALL_VARS here and then make this change in another PR if that's important to you. I just think it's extra work, in this situation, since I'm pretty convinced it's "better" to run printchplenv once if we can.

ronawho: Agreed that we want to only run printchplenv once -- I'm guessing this PR is already going to blow up nightly testing, so I'd prefer to minimize whatever changes we can, but if it's hard to do in this case, I'm fine with leaving it in.

ronawho:
Looks ok to me. I think this should be merged on an otherwise quiet day

compiler/include/driver.h
ronawho: I don't know if I like it any better, but it was the only suggestion I could think of

ben-albrecht: It is mapped to a `cpu-type` according to `gcc` man pages. Would `CHPL_TARGET_CPUTYPE` or something along those lines be any less ambiguous?

mppf: I like combining the proposals for this weird variable, so we get `CHPL_TARGET_ARCH_BACKEND_ARG`

compiler/main/driver.cpp
ben-albrecht: I'm pretty sure it can be removed, but let's save it for follow-up

ben-albrecht:
@mppf && @ronawho - I intend to merge when the current paratest finishes (cleanly).

PR #7977, opened 2017-12-07
Add a pragma to create a default initializer for a type

lydia-duncan:
This will allow us to convert some types to use default initializers instead of
constructors on an individual basis (especially in the modules), instead of
waiting until we can make the switch wholesale.

Some limitations:
- Not yet applicable to records
  - We haven't tested default initializers with those types yet, so it is
    unlikely to work without effort.
- Intentionally not applicable to other symbols
- Not applicable if the class defines an explicit constructor
- Currently not applicable if the class defines an explicit initializer
  - We have played with the idea of allowing users to specify the need for a
    default initializer even if one or more explicit initializers have been
    provided.  This flag could be a stepping stone to providing that support,
    but we don't yet have a proposed permanent syntax and so should wait until
    that is more finalized.

Adds tests to lock in the expected error messages and to show that basic
functionality is provided.

Passed classes/initializers with futures.  Running a full std/ paratest


bradcray:
This was very helpful and seems to be working like a charm—thanks!

PR #7987, opened 2017-12-09
Convert the locRADCache from a constructor to an initializer

bradcray:
This is a simple change to change the locRADCache type from using a constructor to an initializer.


bradcray:
@lydia-duncan: An easy one for your consideration.

PR #7988, opened 2017-12-09
Make Error hierarchy use initializers

bradcray:
This converts the Error hierarchy over from constructors to initializers.  In some cases, I had to use the new `pragma "use default init"` in cases where the Error class did not provide an explicit initializer.  The main code change here was to have all of the subclasses of `SysError` pass their arguments up to the parent class rather than initialize them explicitly.  Once we're in a full default init world, I believe we should just be able to remove these initializers (we may even be able to do it now if we use the pragma... I did not try that).  Most other changes involved changing module and/or test code to use initializers and the pragma in cases where constructors were used on new Error subtypes.


bradcray:
@psahabu , @mppf, @lydia-duncan : I've mailed you about this work, but now it's a PR.

mppf:
> While here, I also removed a few deprecated features from the Error hierarchy since they were deprecated in 1.16.

msg was in 1.16 I think you should double-check this - I think they need to stay in for 1.17. Or we decide that in this case they don't need to be deprecated in a release.

bradcray:
> I think they need to stay in for 1.17.

Oops, you're absolutely right.  Usually my bad sense of time works in the opposite direction.  I also just pushed a few other minor changes revealed in testing this morning.

PR #7993, opened 2017-12-11
Update LLVM alias analysis metadata part 2

dmk42:
Update LLVM TBAA metadata to take struct members into account.  Member access is now reported as part of the struct as well as the member type itself, adding to the information available for the alias analysis passes.

Another PR will recursively handle members of members as well as performing some cleanup and adding some tests.

Passes full local LLVM testing.

### DIFF
--- a/compiler/include/genret.h
+++ b/compiler/include/genret.h
@@ -96,10 +96,14 @@ public:
   // one of the following is set when generating LLVM
   llvm::Value *val; // use val->getType() to obtain LLVM type
   llvm::Type *type; // set when generating a type only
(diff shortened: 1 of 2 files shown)


compiler/include/genret.h
```diff
@@ -96,10 +96,14 @@ class GenRet {
   // one of the following is set when generating LLVM
   llvm::Value *val; // use val->getType() to obtain LLVM type
   llvm::Type *type; // set when generating a type only
+  Type *baseType; // surrounding structure, if this is a field
```
mppf: We might be able to choose a better field name than `baseType` but it is OK and I appreciate the comment.

PR #7997, opened 2017-12-11
Only consider "online" nodes in paratest.chapcs when making a reservation

ronawho:
Previously we just computed the total number of "responding" nodes, but this
didn't work correctly if a node was actually online but marked as "drain" (or
a handful of other states.)

Fix this to only check for nodes that are "online" (allocated, completing,
idle, or mixed -- see https://slurm.schedmd.com/sinfo.html "Node State Codes"
for more info)

### DIFF
--- a/util/test/paratest.chapcs
+++ b/util/test/paratest.chapcs
@@ -78,13 +78,15 @@ def get_exclusive_nodes():
 def get_num_non_exclusive_nodes():
     """
     Get the number of nodes available for testing on the chapel partition
-    (total - exclusive)
+    (online - exclusive)
     """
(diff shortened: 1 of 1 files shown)


util/test/paratest.chapcs
```diff
@@ -78,13 +78,15 @@ def get_exclusive_nodes():
 def get_num_non_exclusive_nodes():
     """
     Get the number of nodes available for testing on the chapel partition
-    (total - exclusive)
+    (online - exclusive)
     """
+    online_states = 'ALLOC,ALLOCATED,COMP,COMPLETING,IDLE,MIX,MIXED'
```
ben-albrecht: I think the variable name `online_node_states` would better describe this string

Issue #6334, opened 2017-05-26
task startup/teardown code


vasslitvinov:
We would like to specify code to be executed at the beginning and end of each task of a forall loop. Here is an outline example:

```chapel
  forall ... {
    // Do this at start of each task.
    var ran: RanT = initializeRandomState();

    // Do this at each iteration of forall,
    // using the 'ran' variable set up at start of the corresponding task.
    generateRandom(ran);

    // Do this at end of each task.
    deleteRandomState(ran);
  }
```

One way to do so is using task-init/deinit functions. The above example could be written:

```chapel
  forall ... {
    proc begin.init() {
      ran = initializeRandomState();
    }
    proc begin.deinit() {
      deleteRandomState(ran);
    }

    generateRandom(ran);
  }
```

The `ran` variable would be a task-private variable, as in #6325 .

We could tweak task-private variable semantics so their type is determined automatically from the loop body, including the init/deinit functions, which would be helpful in this example.

bradcray:
Assuming we add a 'defer' concept to be executed upon function termination, could we perhaps rely on that rather than adding task startup/teardown hooks?  (for the sake of feature minimization).  I.e., imagine writing the above as:

```chapel
  forall ... with (const ran = initializeRandomState()) {
    defer deleteRandomState(ran);
    generateRandom(ran);
  }
```

Hmm, except... is it problematic if a ``defer`` statement is encountered multiple times dynamically during a function execution?  If so, that would require something more like:

```chapel
  forall ... with (const ran = initializeRandomState(),
                         defer deleteRandomState(ran)) {
    generateRandom(ran);
  }
```

which is not quite as nice, but maybe still nicer than having to introduce nested procedures and associating them with the keyword ``begin``?

vasslitvinov:
Here is some syntax:

```chpl
forall ... with (in x, ref y,
  begin init     { execute at START of each task; }
  begin deinit   { execute at END of each task; }
  locale init     { at START of each on-block; }
  locale deinit   { at END of each on-block; }
) {
  loop body;
}
```

Task-private variables can be declared within these begin/local init blocks using normal syntax.

The "locale" annotation above feels natural to me. If we can't get it to parse, we could use `local` instead.

(For convenience in a common case, we could also provide the strawman syntax in #6325.)

vasslitvinov:
Tagging @bradcray @benharsh @mppf for feedback.

mppf:
Here are some more ideas:

[...]

Issue #6325, opened 2017-05-25
Task-private variables


benharsh:
See also: #9529 : per-locale task-private variables; #6334 : task startup/teardown code.

### Motivation

We'd like to develop a local slice/view feature soon that would benefit from task-private variables. We've tossed around the following as potential syntax:
```chpl
forall (i, j) in Space with (ref A = A[local]) {
  B[i,j] = (A[i,j-1] + A[i,j] + A[i,j+1]) / 3;
}
```

Our motivating benchmark desires this functionality for stencil computations. However, I don't think we've nailed down the syntax for task-private variables, so the snippet above may not be what we want.

In the k-nucleotide benchmark we also might want task-private variables:
```chpl
forall i in 1..n with (var D : domain(int, parSafe=false), var A : [D] int) {
  // add to array/domain without need to lock
  // when the task dies, acquire a lock and add our array to a global
  // array
}
```

The NPB MG benchmark will also probably want task-private arrays for some scratch space:
```chpl
forall (i,j) in outer with (var r1, r2 : [0..k] real) {
  for k in inner {
    // stencil computation using r1/r2 as scratch space
  }
}
```

### Strawman proposal

This proposal reflects the syntax in the above examples.
It supports only task-private variables, not locale-private variables.

The proposal is that a task-private variable can be declared
in the `with` clause just like a forall intent, using the syntax
of a variable declaration:
```
variable-kind identifier [: type-specifier] [= expression]
```

`variable-kind` is one of: `var` `const` `ref` `const-ref`

A `ref` or `const-ref` TPV requires `= expression` and do not allow `: type-specifier`.
A `var` or `const` TPV reqiuies at least one of the two clauses.
I.e. its type and/or initialization expression must be specified.

benharsh:
@bradcray , @vasslitvinov , @mppf : Here's a starting point for discussing task-private variables.

bradcray:
[...]

PR #8000, opened 2017-12-12
Introduce task-private variables (TPV)

vasslitvinov:
This is the first step in implementing all we want/need
about task-private variables (TPV). See also:

#8009 : Next steps/remaining TODOs
#6325 : design of TPV
#6334 : design of task startup / tear-down blocks

A task-private variable is similar to a forall intent:

    forall idx in myIter() with (var myTPV = 5) ...

Currently only "var VARNAME = INIT EXPR" form is supported.

COMPILER REPRESENTATION

* a ShadowVarSymbol
  - handled similarly to other forall intents

* a VarSymbol
  - is defined and initialized in the task init block
  - serves as the outerVarSym() for the ShadowVarSymbol

* taskInit block
  - for now, keeps only the defs and inits for the above VarSymbols
  - could also contain user-specified task startup code

* taskDeinit block
  - always empty for now
  - could contain user-specified task tear-down code

* within taskInit and taskDeinit lexical scopes:
  - a reference to a TPV refers to the above VarSymbol
  - a reference to a non-TPV outer variable refers
    to the corresponding ShadowVarSymbol
  - cannot reference a loop induction variable

* the loop body can access all variables declared in taskInit
  (currently, only TPVs), plus shadow variables and inductionvariables

* the lexical scope of taskDeinit is nested within taskInit

Conceptually, a ShadowVarSymbol has been strictly a creature
of the body of the forall loop. This change also allows it
to be used in taskInit/taskDeinit blocks. ShadowVarSymbols
for TPVs and non-TPVs are handled very similarly, see next.

Recall that a key part of the existing implementation of forall intents
is `extendLeaderNew()`. It extends the parallel iterator to propagate
outer variables through its task-parallel and data-parallel constructs
and yield them as part of a newly-created tuple. (We have been considering
getting away from this approach.)

This change keeps this implementation and extends it for TPVs.
Namely:

* Each task construct (lexically present in the parallel iterator)
is extended to execute a clone of taskInit at the beginning
and a clone of taskDeinit at the end.

* Recall that for a given forall intent the yielded tuple
contains the corresponding formal of the enclosing task construct.
(There is a bit more for a reduce intent.)

* For a TPV, there is no corresponding formal.
This is because the TPV starts its lifetime inside the task construct.
Instead, the role of the formal is played by the clone of
the corresponding VarSymbol that is a part of the clone of taskInit
that is executed at the beginning of the task construct (see above).

When the parallel iterator contains a forall loop:

* For each ShadowVarSymbol in the original forall loop,
a corresponding new ShadowVarSymbol is added to shadowVariables()
of the forall in the iterator.

* New clones of taskInit and taskDeinit of the original forall loop
are added to taskInit and taskDeinit blocks of the forall in the iterator.

(This handling of foralls in parallel iterators is to be implemented.)

WHILE THERE

Improved list_view() of ForallStmts.

Fixed ShadowVarSymbol::isConstValWillNotChange().

Minor refactoring. Also, moved bodies of two functions out
from QualifiedType class declaration.

Testing: linux64 --verify; gasnet.


vasslitvinov:
@mppf - would you review?

mppf:
I have some language design questions about this that I asked on issue #6325.

vasslitvinov:
Closing this until we decide whether to change the implementation of forall intents in the compiler.

PR #8003, opened 2017-12-12
Revert "Adjust addVarsToActuals to handle Qualified-Ref fields"

mppf:
This reverts commit d1ceac07d0be8897a01851f2be3ffff205250ed5.

That commit caused failures with --no-local or GASNet for test/studies/comd/llnl/CoMD (the symptom is that the energy is wrong at the start of the simulation). Additionally it caused a performance regression for the force part of the calculation. I'm reverting the change at least until there is a solution for the CoMD issue.

Passed full local testing.

### DIFF
--- a/compiler/passes/flattenFunctions.cpp
+++ b/compiler/passes/flattenFunctions.cpp
@@ -321,14 +321,10 @@ addVarsToActuals(CallExpr* call, SymbolMap* vars, bool outerCall) {
         // This is only a performance issue.
         INT_ASSERT(!sym->hasFlag(FLAG_SHOULD_NOT_PASS_BY_REF));
         /* NOTE: See note above in addVarsToFormals() */
-        if (sym->isRef())
-          call->insertAtTail(sym);
-        else {
-          VarSymbol* tmp = newTemp(sym->type->getValType()->refType);
-          call->getStmtExpr()->insertBefore(new DefExpr(tmp));
-          call->getStmtExpr()->insertBefore(new CallExpr(PRIM_MOVE, tmp, new CallExpr(PRIM_ADDR_OF, sym)));
-          call->insertAtTail(tmp);
-        }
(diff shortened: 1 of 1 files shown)


mppf:
Note, the reverted commit also caused

 Future (bug: use-after-free error for heap var in recursive iterator with coforall-on) [Success matching program output for functions/iterators/bharshbarg/recurseHeapBug]

mppf:
Smaller reproducer found by @benharsh that fails with --no-local but passes with --local:

``` chapel

class D {
  const ld = {1..2, 1..2, 1..2};
  var pe : [ld] [1..64] real;
}

proc main() {
  var d = new D();

  // If it's a coforall, result is '0' but was non-zero before
  // If it's a 'for', result is non-zero as it was before
  coforall (x, y) in zip(d.ld, d.pe) {
    y = x(3) + x(2) * 2 + x(1) * 4;
  }

  writeln(d.pe);
}
```

benharsh:
I was able to write a simpler reproducer:
```chpl
proc main() {
  var A : [1..2, 1..2] [1..10] int;
  coforall a in A {
    a = 1;
  }
  for i in A.domain do writeln(i, ": ", A[i]);
}
```

```
> ./a.out
(1, 1): 0 0 0 0 0 0 0 0 0 0
(1, 2): 0 0 0 0 0 0 0 0 0 0
(2, 1): 0 0 0 0 0 0 0 0 0 0
(2, 2): 0 0 0 0 0 0 0 0 0 0
```

Initial debugging suggests this is a problem in the ``parallel`` pass.

benharsh:
I believe this shows up in ``--no-local`` because of ``remoteValueForwarding``. RVF turns a formal from ``INTENT_REF`` to ``INTENT_CONST_REF``, and later we only check for ``INTENT_REF``. This is likely an oversight, and we should check for any ref-like intent.

PR #8006, opened 2017-12-12
Provide a nice error when BSD Make is used

mppf:
Renames Makefile to Makefile.toplevel.
Puts in Makefile checks for GNU Make and then forwards to Makefile.toplevel.

- [x] Gives error for `make` on FreeBSD 11.1, works with `gmake` there
- [x] error works as expected on omnios-r151020
- [x] tested with `bmake` on Ubuntu
- [x] parallel top-level make still works correctly
- [ ] full local testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,188 +15,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#
-# top-level Chapel Makefile
-#
+# This toplevel Makefile exists simply to make sure that we're not
+# using BSD Make. It could also check a GNU Make version.
+# Real toplevel rules are stored in Makefile.toplevel
 
-#
-# This is the one Makefile that does not/should not include
-# $CHPL_HOME/make/Makefile.base.  The reasons are:
-#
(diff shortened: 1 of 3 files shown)


ben-albrecht:
Does `gen_release` need to be updated to include the new Makefile?

bradcray:
Just glancing, it seems pretty regrettable to have to introduce a whole new Makefile plus indirection to get this check.  I guess I'm not convinced it's worth the additional confusion given that this isn't a problem all that often (and I've got to think that the BSD user community is used to it?)

mppf:
@bradcray - either way, I don't think it's reasonable for us to leave CHPL_MAKE in the default printchplenv output. I think the error message here is superior to including CHPL_MAKE in printchplenv. Also, I don't actually think this adds significant complexity to our Makefiles (since they're already pretty complicated).

bradcray:
I haven't fought back against removing CHPL_MAKE from the default printchplenv output, I've just tried to state why I think it's occasionally been valuable in the past.  I'd much prefer to not have it in the default printchplenv output than adopt this approach.

PR #8007, opened 2017-12-12
Modify expected output for new future, and add a new version

lydia-duncan:
Michael pointed out that I'd merged the two cases related to this failure
in this future - we want an error message asking for a copy initializer
when none is provided, and we want the test to work when it is.

Currently, neither behave as intended.


lydia-duncan:
@mppf - does this look better?

PR #8008, opened 2017-12-12
Fix some failing futures caused by my introduction of new initializers

bradcray:
The approach I took here was to apply the same prediff as I used to
quiet noise on the where-clause4.chpl test to these tests so that we
wouldn't have to keep updating the .bad files as new initializers were
added to the module code.


bradcray:
@lydia-duncan : Would you have time to give this a quick review?  (one of the tests is yours, and you've at least touched the others as well).

PR #8012, opened 2017-12-13
Remove CHPL_LOCALE_MODEL!=flat suppressif

lydia-duncan:
On my advice, Brad added a prediff to this test so we wouldn't have to keep
updating it after every new module code initializer.  But the prediff made it
so that the output across locale models was reunified, so the suppressif is no
longer necessary.  Remove it.


lydia-duncan:
(forgot Brad was out today, I'll just commit the changes)

PR #8013, opened 2017-12-13
Only add FLAG_REF_TO_IMMUTABLE to ref variables

mppf:
This change addresses the --verify failures in studies/prk/Stencil/stencil.

- [x] full local testing


mppf:
@vasslitvinov - could you review this? In particular, it's not so clear to me why newExtraFormal adds a value formal argument for a const-ref actual in some cases.

It was running into the assert here:
https://github.com/chapel-lang/chapel/blob/master/compiler/AST/symbol.cpp#L746

vasslitvinov:
Looks good, thanks!

PR #8014, opened 2017-12-13
Remove c_string_copy

benharsh:
The ``c_string_copy`` type was used to distinguish between owned (``c_string``) and unowned ``char*`` data from extern functions. Now that we have properly managed strings, ``c_string_copy`` seems unnecessary and potentially confusing. Clients of extern functions will need to look at the documentation for the function to determine what to do with the returned ``c_string``.

Testing:
- [x] full local
- [x] full no-local, gasnet
- [x] full memleaks, valgrind


test/types/string/thomasvandoren/int-to-c_string.bad
```diff
@@ -1,4 +1,3 @@
 5 : int(64)
-5.0 : c_string_copy
-5 : c_string_copy
+5.0 : c_string
```
ronawho: Update the .good file for this test too?

benharsh: D'oh, thanks!

modules/standard/IO.chpl
```diff
@@ -1439,13 +1436,13 @@ proc file.getPath(out error:syserr) : string {
...
-    var tmp:c_string_copy;
-    var tmp2:c_string_copy;
+    var tmp:c_string;
+    var tmp2:c_string;
     error = qio_file_path(_file_internal, tmp);
     if !error {
       error = qio_shortest_path(_file_internal, tmp2, tmp);
     }
-    chpl_free_c_string_copy(tmp);
+    qio_free_string(tmp);
```
ronawho: This is just a macro in the runtime that says "This is temporary for now ..."

Can we use chpl_free_c_string instead? (and remove qio_free_string, which seems odd, especially since there's no corresponding qio_alloc_string)

benharsh: Probably... could that be part of another PR? It seems like there are a lot of little things we could do to clean up c_strings in general.

test/runtime/gbt/tasks/idToString.chpl
```diff
@@ -13,7 +13,7 @@ proc main() {
     var buf: [1..bufLen] buf_t;
     var idStr = chpl_task_idToString(c_ptrTo(buf), buf.size:size_t, id);
     writeln('task ID of ', what, ' is: ',
-            if idStr==_nullString then '<OVF>' else idStr:string);
+            if idStr==nil then '<OVF>' else idStr:string);
```
ronawho: Should this (and all other places that were _nullString ) use c_nil instead?

modules/standard/FileSystem.chpl
```diff
@@ -544,19 +544,18 @@ proc copyTree(src: string, dest: string, copySymbolically: bool=false) throws {
...
   on this {
-    var tmp:c_string_copy;
-    // c_strings and c_string_copy's can't cross on statements.
+    var tmp:c_string;
+    // c_strings can't cross on statements.
     error = chpl_fs_cwd(tmp);
     if (error != ENOERR) {
       ret = "";
     } else {
       var tmp_len = tmp.length;
```
ronawho: Get rid of `tmp_len` too?

modules/standard/IO.chpl
ronawho: Yeah, that sounds fine. I mentioned it only because this adds a second use of it (and the only other use of it is in this file too)

PR #8017, opened 2017-12-13
File a future against an important generic inheritance init() pattern

bradcray:
This test shows a common pattern that we use in our array/domain hierarchy
which doesn't seem to be expressible today, and which I fear demonstrates
the need for there to be the concept of a "phase 0" during initialization in
which compile-time generics are resolved prior to phase 1 execution-time
field initialization (as has been anticipated by some users).


bradcray:
Initializers team (@lydia-duncan, @noakesmichael, @mppf, @vasslitvinov): I ran into this last night on my way out the door while starting work on converting the domain map hierarchy over to using initializers and am looking for insights.  Thanks.

bradcray:
(PS -- I'll open an issue against this as a better place for discussion once someone confirms that I'm not missing something obvious)

lydia-duncan:
That's a known issue, unfortunately.  I called it out in the slides for last release as a case we need a solution for, but I haven't initiated a discussion about it yet.  I have a test for a variation of it [here](https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/generics/inheritance/concrete-sub-inherits-type.chpl) (where the old test is related to an inherited type field but really the issue is any inherited information used for determining the type of a sub class field.

bradcray:
Is there an issue open for it that you're aware of?

lydia-duncan:
A quick look didn't reveal one to me, so I don't think so.

PR #8021, opened 2017-12-13
Adjust addVarsToActuals for Qualified-ref actuals again

mppf:
Fixes the problem with CoMD from PR #7990 (reverted in #8003).

- [ ] full local testing
- [x] CoMD performance does not regress
- [x] CoMD passes with --no-local


mppf:
TODO: 1 failure functions/deitz/iterators/test_coforall_iterator_yields_domains

benharsh:
With this PR we now have a pattern like:
```
chpl__buildDomainExpr(..., &_ic_thing);
tmp = chpl__autoCopy(&_ic_thing);
i = tmp;
argbundle->i = i;
<launch coforall>
chpl__autoDestroy(i);
chpl__autoDestroy(_ic_thing);
```
This means that ``i`` may be destroyed while the coforall body is executing. On master we have an extra autoCopy before placing ``i`` into the arg bundle. The autoCopy is missing because of the modification to parallel.cpp.

Is this the sort of thing your other branch could fix? It seems to me like the autoDestroy of ``i`` is in the wrong place, and that it should be in the coforall wrapper/body.

mppf:
Yeah, the other branch concerns itself with such things. I'm just not yet sure how to approach the --baseline failures in the meantime.

benharsh:
Though unfortunate, perhaps it would be best to revert the PR that originally modified flattenFunctions and include those changes in the other branch?

mppf:
That pattern now works correctly on my other branch. I'll see about reverting some/all of the other PR in the meantime.

mppf:
This will be addressed in PR #8073.

PR #8022, opened 2017-12-13
Perf test to measure array view creation

benharsh:


benharsh:
@ronawho : I was motivated to create this test while working on a line-by-line variant of reverse-complement. The revcomp variation created a slice of an IO buffer when scanning for a line. The 'fasta.big' input file has many lines, so the number of slices has a large impact.

Care to review?

test/performance/bharshbarg/create-views.skipif
```diff
@@ -0,0 +1,2 @@
+# We only care about this test for performance
+CHPL_TEST_PERF != on
```
ronawho: I don't see any reason to not run this as a correctness test too

PR #8028, opened 2017-12-14
Remove hpl.hpcc2012 perf graphs

ronawho:
hpl.hpcc2012 was disabled (most recently) with #7744, but we didn't stop
graphing the old results. The graphs aren't useful right now and are slightly
annoying when doing performance triage, so just remove the graphs. They can
always be re-added if we start performance testing again.


ronawho:
@benharsh quick sanity check?

PR #8030, opened 2017-12-14
Fix `nil` dereferences from `Shared(nil)` objects

daviditen:
In some cases involving `Shared(nil)` objects, `Shared` was calling `retain()`
on `nil`.  Fix it by testing for `nil` before calling `retain()`.

Add 3 tests that made `nil` dereferences before this change but pass now.

### DIFF
--- a/test/library/packages/SharedObject/shared-nil-in-arg.chpl
+++ b/test/library/packages/SharedObject/shared-nil-in-arg.chpl
@@ -1,0 +1,9 @@
+use SharedObject;
+
+class C { }
+
+proc foo(in sc: Shared(C)) { }
+
+var sc = new Shared(nil:C);
+
+foo(sc);
--- a/modules/packages/SharedObject.chpl
+++ b/modules/packages/SharedObject.chpl
@@ -150,7 +150,8 @@ module SharedObject {
       this.p = src.p;
       this.pn = src.pn;
 
-      this.pn.retain();
+      if this.pn != nil then
+        this.pn.retain();
     }
 
     /*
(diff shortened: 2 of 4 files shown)


daviditen:
Standard Linux64 paratest:
[Summary: #Successes = 8103 | #Failures = 0 | #Futures = 0]

test/library/packages/SharedObject/shared-nil-in-arg.chpl
```diff
@@ -0,0 +1,9 @@
+use SharedObject;
+
+class C { }
+
+proc foo(in sc: Shared(C)) { }
+
+var sc = new Shared(nil:C);
```
vasslitvinov: @mppf, @daviditen - I wonder if we should provide an initializer for Shared that just accepts a type, for convenience. That way the user would write `new Shared(C)` instead of `new Shared(nil:C)`.

Ditto for Unique.

mppf: @vasslitvinov - you could always write 
``` chapel
var sc:Shared(C);
```

today, couldn't you?

Additionally I'm uncertain what the future will be for Owned and Shared storing `nil` - this is one of the things we might seek to make compile-time checking for. If it's going to be a compile-time error in some situations I'm not sure we're ready to have a shortcut for it.

daviditen: The way I'm using it in the `DateTime` module it needs to be able to store `nil`.  The `TZInfo` class representing the time zone is `nil` for instances that don't want to bother with time zones.  I think this will be the common case, but I want to make it `Shared` for when it is actually used.

daviditen: > you could always write
> 
> `var sc:Shared(C);`
>
> today, couldn't you?

Yes - in fact one of the tests added in this PR did that.

daviditen:
Thanks for reviewing @vasslitvinov!

PR #8033, opened 2017-12-14
Fix the insertion of omitted initialization for fields using cond/loop exprs

lydia-duncan:
Made InitNormalize::updateFieldsMember check if what it was provided was
a call that turned out to be a method (and has not already been transformed
into a method).  If so, it alters the AST so that we can recognize it as a
method call later.  Then, if we recognize it is a method call and the method
is an inserted conditional or loop expression function, don't generate our
method warning.

The AST set up I used for this was a bit different from what normally is
present prior to normalize's applyGetterTransform().  It seemed silly to
insert a SymExpr to a VarSymbol with an Immediate storing the name that was
just going to get transformed back into an UnresolvedSymExpr when we already
had an UnresolvedSymExpr.  So I added a case to applyGetterTransform to handle
this new set up.

Note: this implementation does not currently help the default initializers case,
due to the inserted initialization being placed as the default value for the
corresponding argument (and thus not where we expect to find it in the
validation of initializer rules).  I'm making that work a separate PR.

Add a test each for cond/loop expr initial values with omitted initialization,
and a future for when inserted omitted initialization involves a method call
(which we fail to detect).

Passed test/classes/initializers with futures, and a full paratest run.


lydia-duncan:
Retested after the changes I made: passed classes/initializers with futures, running another full std/ paratest

PR #8037, opened 2017-12-14
Add barrier subtimings to isx-hand-optimized

ronawho:
Previously the exchange subtimer timed the exchange as well as the barrier.
Split the barrier out into a separate subtimer so we can better tell where time
is spent and maybe narrow down the noisy part of the computation.


ronawho:
@benharsh just checking if you're ok with this

Note that I still need to test this and will need to update the .dat files when it goes in

ronawho:
Belatedly, I'm realizing this information might not be all that useful. Even with the slow barrier that we have, this will mostly likely just tell us how much time is spent waiting on the barrier (i.e. that the exchange is noisy, so some tasks finish earlier and wait longer), not how much overhead the barrier is adding.

Do you think this is worth adding, or skip it?

benharsh:
Probably skip for now. A microbenchmark for Barrier will likely be more useful for the foreseeable future.

Issue #7932, opened 2017-12-05
Copy initializers with extra optional arguments aren't considered


bradcray:
### Summary of Problem

A copy initializer that has additional optional arguments (with default values) is not considered as it seems it should be.

### Steps to Reproduce

**Associated Future Test(s):**
[`test/types/records/init/copy/copyWithDefaultArgs.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/types/records/init/copy/copyWithDefaultArgs.chpl
          ) #7920

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.17.0 pre-release (6c08a44)`

bradcray:
We use this pattern in the string module currently, so to convert it from constructors to initializers in the most obvious way, we'd need to resolve this.

bradcray:
I hadn't realized until just now when I was taking stock of things, but... it looks like this has been resolved and can be closed?  (@lydia-duncan?)

lydia-duncan:
Yup!  I'll take care of it (I meant to earlier, but wanted to make sure testing didn't blow up)

PR #8038, opened 2017-12-14
Add test of copy initializer taking `const ref` argument

bradcray:
Add a test of an explicit copy initializer whose argument is taken in by `const ref`.
Prior to PR #8087, this test did not work and this PR included a modest change to
the compiler to enable it.  PR #8087 has obviated the need for that compiler change
but adding the test still seems valuable.

Here's the original PR message for the historical record (and so that GitHub comments
from Michael make more sense to future readers):

This is a modest change to the compiler's current logic to find copy initializers which
permits the 'other' argument to be taken in by 'ref'.  In converting the BigInteger and
GMP modules over to use initializers rather than constructors, this was required in order
to have the current bigint copy constructor/initializer be recognized as such since it
takes its 'bigint' argument by 'const ref' intent.  My guess is that this will probably not
break any existing tests but am still running testing on this mod.

This mod also makes it obvious to me why copy initializers with additional arguments are
not recognized as such as noted in issue #7932.  Our current logic for identifying
copy initializers is to simply look for any expected signature rather than trying to
make a call that would dispatch to such a signature.  I did not undertake anything more
significant that would address that issue here (and don't currently intend to).


bradcray:
@mppf / @noakesmichael : I'm curious for your thoughts on this.  Is there a reason that a copy initializer should not / would not take its argument as ref that I'm not thinking of?  Thanks!

bradcray:
linux64 testing came back clean

bradcray:
@mppf / @lydia-duncan : It appears to me that this PR was obviated by PR #8087, so I'm planning on closing it unless that surprises you.  A question would be whether the test in this PR is sufficiently different from those that exist to be worth merging on its own without the compiler change.

lydia-duncan:
I don't have any tests with explicit intents for initializers yet, so I think merging this test would be worthwhile.

bradcray:
@mppf: I was mentioning to Lydia en route to lunch that I was expecting that if/when I merged master into this branch, it would conflict due to whatever changes were made in PR #8087 to get this case working.  It turns out that that wasn't the case—the routine in question was unchanged.  This made me a bit surprised since previously my change had been required in order to "find" candidate copy constructors.  While out at lunch, I conjectured that maybe the routine was dead now, but that seems not to be the case (which is to say, inspecting the source, I can find a call to it).  That left me confused about how PR #8087 did cause this test to start working...  On one hand, I don't need to know, but on the other, I'm curious if either of you happen to understand.

lydia-duncan:
#8087 caused some autoCopy resolution to occur that had not previously happened.  Perhaps the reason the refType check was necessary was because it had not followed the appropriate resolution path to use QUAL_REF instead of refType in the old world?

It also might be (but I hope not) that the const ref version is being ignored in favor of the compiler generated one - you could check that by inserting a writeln into the defined copy initializer, to ensure it was the one being called (as otherwise its execution is pretty indistinguishable)

bradcray:
Putting the writeln() in, it is calling the one in the code (whew!)

bradcray:
Also, your guess about QUAL_REF sounds reasonable to me.

mppf:
I don't know the answer to why the earlier compiler change in this PR is no longer necessary.

PR #8041, opened 2017-12-15
Filter out LLVM compile options that inhibit debugging

dmk42:
The `-gsplit-dwarf` option prevents gdb from being able to see the source files used to build chpl.  This change filters it out from the `llvm-config` output.

While there, also filter out `-fPIC` which can increase code size and/or execution time, and is not needed except in shared libraries.  Empty options are also filtered out, to remove unnecessary blanks.

Tested on linux64 and Mac, with CHPL_LLVM=llvm and CHPL_LLVM=system.

### DIFF
--- a/third-party/llvm/filter-llvm-config.awk
+++ b/third-party/llvm/filter-llvm-config.awk
@@ -1,5 +1,8 @@
 BEGIN { RS=" " }
+/^$/ { next }
 /^-DNDEBUG$/ { next }
+/^-fPIC$/ { next }
+/^-gsplit-dwarf$/ { next }
 /^-O[0-4s]?$/ { next }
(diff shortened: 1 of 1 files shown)


mppf:
I thought that using -fPIC was necessary when building with a library using -fPIC but maybe I was mistaken?

dmk42:
It turns out that \-fPIC just causes all global symbols to be accessed indirectly through a Global Offset Table (GOT).  A shared library needs to be compiled that way, but other code can link with it with or without \-fPIC.  That's because the library's symbol names all point to the GOT, so even code compiled without \-fPIC knows where to find them.

So, the answer is no, it isn't needed outside of the shared library itself.  (Even a non-shared library doesn't need it.)

PR #8046, opened 2017-12-15
Update disambiguation rules in spec

mppf:
This PR updates the spec to match the currently implemented disambiguation rules. This update is a follow-on to PR #7659.

### DIFF
--- a/spec/Procedures.tex
+++ b/spec/Procedures.tex
@@ -1229,9 +1229,13 @@ mapping to $F_2$ and none of the legal argument mappings to $F_2$ is a
 more specific argument mapping than the corresponding legal argument
 mapping to $F_1$, then $F_1$ is more specific.
 
-\item Similarly, if at least one argument mapping to $F_2$ is {\em more
-specific} and none of the argument mappings to $F_1$ are, then $F_2$ is
-more specific.
+\item
+If at least one of the legal argument mappings to $F_2$ is a {\em more
+specific argument mapping} than the corresponding legal argument
+mapping to $F_1$ and none of the legal argument mappings to $F_1$ is a
+more specific argument mapping than the corresponding legal argument
+mapping to $F_2$, then $F_2$ is more specific.
+
 
 \item If $F_1$ shadows $F_2$, then $F_1$ is more specific.
 \item If $F_2$ shadows $F_1$, then $F_2$ is more specific.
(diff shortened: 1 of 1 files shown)


spec/Procedures.tex
```diff
@@ -1224,24 +1225,47 @@ \subsection{Determining More Specific Functions}
...
 more specific argument mapping than the corresponding legal argument
 mapping to $F_1$, then $F_1$ is more specific.
-\item
-If at least one of the legal argument mappings to $F_2$ is a {\em more
-specific argument mapping} than the corresponding legal argument
-mapping to $F_1$ and none of the legal argument mappings to $F_1$ is a
-more specific argument mapping than the corresponding legal argument
-mapping to $F_2$, then $F_2$ is more specific.
+
+\item Similarly, if at least one argument mapping to $F_2$ is {\em more
```
daviditen: I think this should go back to matching the pattern of the rest of the list items - worded the same as the previous point but with F_1 and F_2 swapped.

spec/Procedures.tex
```diff
@@ -1224,24 +1225,47 @@ \subsection{Determining More Specific Functions}
...
+specific} and none of the argument mappings to $F_1$ are, then $F_2$ is
+more specific.
+
 \item If $F_1$ shadows $F_2$, then $F_1$ is more specific.
 \item If $F_2$ shadows $F_1$, then $F_2$ is more specific.
-\item If all \chpl{param} arguments prefer $F_1$ over $F_2$, then $F_1$ is more specific.  In order of preference, a \chpl{param} argument prefers to be passed to (a) a \chpl{param} formal of matching type; (b) a \chpl{param} formal large enough to store the \chpl{param} value; (c) a non-\chpl{param} formal of matching type.
-\item If all \chpl{param} arguments prefer $F_2$ over $F_1$, then $F_2$ is more specific.
+
+\item If at least one of the legal argument mappings to $F_1$ is {\em
```
daviditen: Would it be possible to go back to a single level of specific/preferred by defining more ordered rules in the "determining more specific" section below?  Having four different levels is a pretty big increase in how confusing this is.

mppf: It's true that it's more confusing. The additional levels are helping with cases where there are say 2 functions f1 and f2 under considerations, and each has two arguments, let's say f1a f1b and f2a f2b. The situation (that comes up a lot) is that f1a is a better match than f2a but f2b is a better match than f2a. The multiple levels help this case. Rearranging the order within the determining more specific functions / arguments can't help that case.

PR #3550, opened 2016-03-22
Fix lustre support on non-esl crays

ronawho:
Our lustre support requires lustre/lustreapi.h, but this header wasn't
implicitly available on login nodes. It is on esl nodes (where we happen to be
building our modules) but if you try to build lustre support on a non-esl node
you'll get errors about the missing header.

In order to resolve this we need bring in the devel api when we're on login
nodes. The devel api isn't available on all machines (it's not on esls) so we
first check if a pkg-config call to it will be successful before trying to
include it. There might be a more principled way to check esl vs non-esl, but
this also handles cases where the devel api might not have been installed.

PR #8051, opened 2017-12-16
Rely on externally-set CHPL_AUXIO* Make variables to provide lustre api.

awallace-cray:
This change simply reverts PR #3550. PR #3550 was not entirely successful
because it only fixes the "missing file lustre/lustreapi.h" problem for
Cray-PrgEnv compilers- it does not help the clang-included compiler, which
does not interact with the Cray PrgEnv at all.

As we have recently decided to use the CHPL_AUXIO* Make variables to
provide lustre api paths and flags for clang-included, we may as well let
the the Cray-PrgEnv compilers get the same info the same way.

### DIFF
--- a/make/compiler/Makefile.cray-prgenv
+++ b/make/compiler/Makefile.cray-prgenv
@@ -76,12 +76,3 @@ ifneq (,$(filter $(CHPL_MAKE_COMM_SUBSTRATE),gemini aries))
 export PE_CHAPEL_PKGCONFIG_LIBS := cray-pmi:cray-udreg:cray-ugni:$(PE_CHAPEL_PKGCONFIG_LIBS)
 endif
 endif
-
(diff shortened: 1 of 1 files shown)


awallace-cray:
This is entirely optional, and I do not care very much whether it is merged or discarded. It simply removes some now-redundant Makefile clutter, which seems a good thing.
This was tested on Cray-XC, both with and without AUX_FILESYS=lustre. 
Not tested anywhere else.

ronawho:
I don't think we want to revert this. Your fix was for the internal module build, but this patch works for users building from source too. I think the right longer term solution is to apply the internal patch to the public repo. But I think there's going to be a fair amount of changes to clang-included and getting --llvm to work without a PrgEnv-* wrapper loaded, so I think the internal patch is a good short-term solution for now

PR #8058, opened 2017-12-18
Suppress lcals performance testing under cray-prgenv-gnu

ronawho:
This test is failing with an undefined reference to "_ZGVcN4v___exp_finite",
because the driver isn't throwing -lmvec. Suppress it until the driver is fixed


ronawho:
@daviditen any objections to this?

daviditen:
@ronawho I'm fine with that.

PR #8070, opened 2017-12-19
Use default-args wrapper fn for default init for now

mppf:
Without this, tests such as

  classes/initializers/compilerGenerated/syncFieldTypeOnly.chpl

that use sync variables and throw --force-initializers will fail in fifo
configurations because the allocated sync variable will be destroyed
twice.

I'd consider the plan to improve 'in' intent and use it for
default initializers to be the proper way to fix this problem.
Until then, this change just puts default initializers back to
using the old default argument wrappers (which have special
cases for this kind of thing).

- [x] full local fifo testing
- [x] full local qthreads testing


mppf:
@lydia-duncan - can you review this workaround? thanks.

PR #8073, opened 2017-12-19
Remove FLAG_INSERT_AUTO_COPY & rework iterator memory management

### LINKED ISSUE #7994: What should happen when a promoted expression is returned?
When returning a promoted expression, a reasonable question to ask is when the promoted expression should be evaluated. Consider for example the following program, which promotes `add`. Should "In add" be printed after "In f" or after "In g"? (In other words, should the promoted expression be evaluated in the process of returning from f() ? Or should it be evaluated in the process of being assigned to C?).

``` chapel
var A:[1..4] int;
for i in 1..4 do
  A[i] = i;
var B:[1..4] int;
for i in 1..4 do
  B[i] = 10*i;

proc add(a: int, b: int) {
  writeln("In add");
  return a + b;
}

proc f() {
  writeln("In f");
  return add(A, B);
}

proc g(x) {
  writeln("In g");
  A[2] = 1;

  var C = x;
  writeln(C[2]); // Does this print 21 or 22?
}

g(f());
```

On master today, `f` returns an iterator record that contains the promoted expression. That causes the evaluation of `add` to be put off until the initialization of `C` in `g`.

This strategy has the downside that it is unclear how to correctly compile a program such as the following, in which the promoted iterator captures a local variable:

``` chapel
var A:[1..4] int;
for i in 1..4 do
  A[i] = i;

proc add(a: int, b: int) {
  writeln("In add");
  return a + b;
}

proc f() {
  writeln("In f");
  var B:[1..4] int;
  for i in 1..4 do
    B[i] = 10*i;
  return add(A, B);
}

proc g(x) {
  writeln("In g");
  A[2] = 1;

  var C = x;
  writeln(C[2]); // does it print out 22 or 21?
}

g(f());
```

This program prints neither 21 or 22 on master today (and has memory access errors instead).

test/arrays/ferguson/semantic-examples/1-return-promoted-expr.chpl
test/arrays/ferguson/semantic-examples/valgrind-futures/4-return-promoted-expr-local.chpl
test/arrays/ferguson/semantic-examples/valgrind-futures/4-return-promoted-expr-records.chpl

mppf:
### Fixing a User-Discovered Memory Leak

I started on this branch in order to fix a memory leak that was causing problems for a Chapel user. This PR adds reproducers for these and also fixes the memory leak. The reproducers are:
 * test/functions/iterators/ferguson/no-leak-string-bug.chpl	
 * test/functions/iterators/ferguson/no-leak-record-bug.chpl

### FLAG_INSERT_AUTO_COPY

FLAG_INSERT_AUTO_COPY is removed by this PR. It and the accompanying code in callDestructors that added copies based on that flag are no longer needed.

### Returning iterator records
Returning an iterator record now returns an array ( see issue #7994 for discussion ). That includes the case of returning a promoted expression and also the case of returning a call to an iterator in a non-iterator function. Since this might change the behavior of existing programs, for this release there is a warning when that happens. To support various odd cases in the compiler and to reduce the number of tests that needed to change immediately, FLAG_FN_RETURNS_ITERATOR allows a function to return an iterator record (instead of always converting it into an array).

### Managing iterator memory
Iterators now treat yielding more similarly to returning - in particular, a yield by value represents a transfer of ownership (i.e. the responsibility to destroy, in the case of a record) from the iterator to the loop using the iterator.

Iterator records are now managed similarly to other records - isUserDefinedRecord returns true for them.

callDestructors is now aware of the value yielded by an iterator and treats it specially (just as the value returned by a function needs special treatment).

### Iterator yield value normalization
Moves the normalization of iterator yields - where there is one "yield value variable" per yield - to normalize instead of insertYieldTemps. 

Commit 8d3b2c4065d6466dbaadab0ae0c8444e6645dae6 introduced a separate pass to create a different "yield temp" per yield, instead of having all of the yields share one. In the commit message, it points out that it would be better to adjust the normalizer.
    
Since an iterator can continue to run after the yield (but not a return) it makes sense to represent yield and return with different AST. In particular, since each yield represents a different occasion for transfering a value to a calling loop, it makes sense to have each yield work with its own "yield value variable". 

Along these lines, resolveReturnType now computes the yielded type using a similar algorithm to returns (can one of the yielded types coerce into the other?).

### Adds mechanism to retrieve static type & static promotion type
Adds PRIM_STATIC_TYPEOF and PRIM_SCALAR_PROMOTION_TYPE in order to build the following conversion.

### Adjusts conversion of iterator to array

The `proc chpl__initCopy(ir: _iteratorRecord)` in ChapelArray implements the conversion from an iterator to an array. As before, this conversion is serial now. I reworked it in order to be sure that the yielded value is "moved" into the array and because I was having problems with the way the iterator record was used previously to get the return type (so the returned array could be allocated). In particular the iterator record was previously run twice, but now it is run once. This change has the side effect of supporting some skyline array patterns.

In particular, the following code creates a skyline array after this PR:
``` chapel
proc makeArray(i:int) {
  var A:[1..i] int;
  A[1] = i;
  return A;
}

// AA is an array of arrays
var AA = for i in r do makeArray(i);
```

Supporting this pattern is a side effect of the improvements to memory management in `proc chpl__initCopy(ir: _iteratorRecord)`. There are certainly many ways in which skyline arrays need more design and implementation attention. (including issue #8152 which asks what AA.eltType should be in the above example).

### Other notes

This PR tidies up test/modules/figueroa/DuplicateConfigVars/ a little bit to make it more focused on the issue of shadowing config vars. That test was a feature request for a compilation warning but was failing to compile for various reasons.

### PR Testing, Review, and Related Issues

Resolves #7994.

Tested single-locale performance in the performance playground & looks OK with probably real improvements to:
* k-nucleotide
* Array initialization (slower idioms)
* HPCC HPL
* Allocating string operations (char iteration)

- [x] PR message
- [x] full local testing
- [x] baseline testing
- [x] --no-local testing
- [x] GASNet testing
- [x] performance evaluation


mppf:
@vasslitvinov - I believe this is ready for your review. I'll run more testing on it but I believe I've got the --baseline situation under control. I realize this is large, but my earlier efforts to split parts of it off have been stymied by testing failures in one configuration or another.

mppf:
@vasslitvinov - things to consider when reviewing:
 * can you think of tests of iterator memory management that we should be doing? Or tricky cases I might not have gotten right?
 * I wanted to remove localizeReturnSymbols by replacing it with code in the normalizer, but wasn't able to go all the way on that. Search for 8d3b2c4065d6466dbaadab0ae0c8444e6645dae6 in the source code. Any ideas here?

mppf:
Here is a reproducer for the user memory leak`stringLeak.chpl`:

``` chapel
iter foo() {
  var A : [1..10] string = "hello";

  for a in A do yield a;
}

iter foo(param tag:iterKind) where tag == iterKind.standalone {
  var A : [1..10] string = "hello";

  for a in A do yield a;
}

proc main() {
  var x = {50..100};
  forall f in foo() with (const ref x) {
    writeln(x.member(f.length));
  }
}
```

modules/internal/ChapelArray.chpl
```diff
@@ -3988,49 +4016,119 @@ module ChapelArray {
...
+      // The intent here is to transfer ownership to the array.
+      __primitive("=", data[i], eltCopy);
+
+      i += 1;
     }
 
-    D = { 1 .. i - 1 };
+    // let the comm layer adjust array allocation
+    if callAgain then
+      __primitive("array_alloc", data, size, subloc, c_nil, data);
```
vasslitvinov: Please consider moving the above conditional under "if data != nil" below.

modules/internal/ChapelArray.chpl
```diff
@@ -3988,49 +4016,119 @@ module ChapelArray {
...
+
+      // Future: we should generally remove this copy.
+      // Note though that in some cases it invokes this function
+      // recursively - in that case it shouldn't be removed!
+      pragma "no auto destroy"
+      pragma "no copy"
+      var eltCopy = chpl__initCopy(elt);
+
+      pragma "insert auto destroy"
+      type eltType = eltCopy.type;
```
vasslitvinov: I do not see you use `eltType`. If this is intentional, please add a brief comment explaining why. Otherwise please remove `eltType`.

modules/internal/ChapelArray.chpl
```diff
@@ -3988,49 +4016,119 @@ module ChapelArray {
...
+    // let the comm layer adjust array allocation
+    if callAgain then
+      __primitive("array_alloc", data, size, subloc, c_nil, data);
 
-    return A;
+
+    if data != nil {
+      // Now construct a DefaultRectangular array using the data
+      pragma "insert auto destroy"
+      var D = { 1 .. #i };
```
vasslitvinov: I am curious: why `pragma "insert auto destroy"` on `D` ? The user can write this code without such a pragma, right?

    proc foo() {
      var D = {1..2};
      var A: [D] int;
      return A;
    }

test/arrays/ferguson/from-iterator/skyline.chpl
```diff
@@ -0,0 +1,21 @@
+config const n = 2;
```
vasslitvinov: I suggest adding the phrases "skyline array" and "jagged array" into the body of this test. That way this test will will be picked up by grep.

compiler/AST/iterator.cpp
```diff
@@ -87,16 +87,23 @@ removeRetSymbolAndUses(FnSymbol* fn) {
...
   INT_ASSERT(ret && ret->isPrimitive(PRIM_RETURN));
-  SymExpr* rse = toSymExpr(ret->get(1));
-  INT_ASSERT(rse);
-  Symbol*  rsym = rse->symbol();
+
+  if (CallExpr* assign = toCallExpr(ret->prev))
+    if (assign->isPrimitive(PRIM_MOVE) || assign->isPrimitive(PRIM_ASSIGN))
+      if (SymExpr* se = toSymExpr(assign->get(1)))
+        if (se->symbol()->hasFlag(FLAG_RETARG))
+          assign->remove();
```
vasslitvinov: Why should we delete an assignment to _retArg? Maybe add a brief comment about it?

compiler/AST/primitive.cpp
```diff
@@ -272,12 +284,21 @@ returnInfoGetMember(CallExpr* call) {
   VarSymbol* var = toVarSymbol(sym->symbol());
   if (!var)
     INT_FATAL(call, "bad member primitive");
-  if (var->immediate) {
-    const char* name = var->immediate->v_string;
-    for_fields(field, ct) {
-      if (!strcmp(field->name, name))
-        return field->qualType();
+  if (Immediate* imm = var->immediate) {
```
vasslitvinov: Could you also remove the "if (!ct)" test above, since you are editing anyway?

compiler/AST/primitive.cpp
```diff
@@ -532,10 +553,24 @@ initPrimitive() {
...
   prim_def(PRIM_GET_MEMBER, ".", returnInfoGetMemberRef);
   prim_def(PRIM_GET_MEMBER_VALUE, ".v", returnInfoGetMember, false, true);
-  // base, field, value
+
+  // PRIM_SET_MEMBER: base, field, value
+  // if the field is a ref, and the value is a ref, sets the ptr.
+  // if the field is a ref, and the value is a not ref, invalid AST
+  // if the field is not ref, and the value is a ref, derefs value first
+  // if neither are references, sets the field
   prim_def(PRIM_SET_MEMBER, ".=", returnInfoVoid, true, true);
```
vasslitvinov: Thanks for adding the comments on PRIM_SET*/GET* !

compiler/AST/primitive.cpp
```diff
@@ -576,7 +611,19 @@ initPrimitive() {
   // PRIM_CAST arguments are (type to cast to, value to cast)
   prim_def(PRIM_CAST, "cast", returnInfoCast, false, true);
   prim_def(PRIM_DYNAMIC_CAST, "dynamic_cast", returnInfoCast, false, true);
+
+  // PRIM_TYPEOF an array returns a runtime type (containing its domain)
+  // For values without a runtime type component, it works the same as
+  // PRIM_STATIC_TYPEOF
   prim_def(PRIM_TYPEOF, "typeof", returnInfoFirstDeref);
+
+  // Return the compile-time component of a type (ignoring runtime types)
```
vasslitvinov: Could you also say what it returns if the type is an array or domain type?

compiler/AST/primitive.cpp
```diff
@@ -576,7 +611,19 @@ initPrimitive() {
   // PRIM_CAST arguments are (type to cast to, value to cast)
   prim_def(PRIM_CAST, "cast", returnInfoCast, false, true);
   prim_def(PRIM_DYNAMIC_CAST, "dynamic_cast", returnInfoCast, false, true);
+
+  // PRIM_TYPEOF an array returns a runtime type (containing its domain)
```
vasslitvinov: Did you mean "**of** an array" ?

compiler/passes/flattenFunctions.cpp
```diff
@@ -27,6 +27,7 @@
 #include "stlUtil.h"
 
 static void flattenNestedFunctions(Vec<FnSymbol*>& nestedFunctions);
+static void markTaskFunctionsInIterators(Vec<FnSymbol*>& nestedFunctions);
```
vasslitvinov: Geez, so much stuff will go away when we reverse the introduction of task functions early in the compiler!

compiler/AST/primitive.cpp
mppf: I moved it above the check for a flag on ct->symbol

modules/internal/ChapelArray.chpl
mppf: Yes. But we've historically had special treatment for 'initCopy' functions.
I don't know that this pragma is required, but such a pragma was in the previous version, and I'm getting myself into enough trouble as it is, so I'd like to leave it alone for now.

compiler/resolution/callDestructors.cpp
```diff
@@ -1278,29 +889,49 @@ static void insertYieldTemps()
...
-      Symbol* tmp = newTemp("_yield_expr_tmp_", type);
-      Expr* stmt = call->getStmtExpr();
-      stmt->insertBefore(new DefExpr(tmp));
-      stmt->insertBefore(new CallExpr(PRIM_MOVE, tmp, new CallExpr(getAutoCopyForType(type), yieldExpr->remove())));
-      call->insertAtHead(new SymExpr(tmp)); // New first argument.
+
+    // If the yielded symbol is subject to auto-copy/destroy discipline
+    // and it's returned by value (not by reference)
+    // and the yielded value is not an expression temporary
+    //  (e.g. for yield someCall(), the result of someCall() doesn't need copy)
```
vasslitvinov: Seems like the opening "If" of this sentence is lacking "then". Would you double-check?

compiler/resolution/lowerIterators.cpp
```diff
@@ -971,6 +981,10 @@ static void localizeReturnSymbols(FnSymbol* iteratorFn, std::vector<SymExpr*> sy
 // will be included in 'asts' and handled when 'fn' is the enclosing
 // iterator.
 //
+// see commit 8d3b2c4065d6466dbaadab0ae0c8444e6645dae6
+// Now we're using a return temp per yield, but this change is still
+// apparently necessary. I think that lowerIterators isn't copying some
+// variables when it should be.
```
vasslitvinov: Maybe add "todo: remove this" ?

compiler/AST/type.cpp
```diff
@@ -1392,12 +1405,14 @@ bool isString(Type* type) {
 // Noakes 2017/03/02
 // This function now includes range and atomics
 //
+// MPF    2017/08/03
+// This function now includes iterator records
```
vasslitvinov: Maybe add *"so it really should be called `isUserLikeRecord`"* ?

test/arrays/ferguson/from-iterator/for-yields-no-arrays.future
```diff
@@ -0,0 +1,2 @@
+design: runtime type missing for yield-nothing
+
```
vasslitvinov: Need some more explanation here. (I had to figure it out myself).
There should probably be a github issue for this as well,
or just cross-reference with #8152.

test/modules/figueroa/DuplicateConfigVars/RandomNumber.chpl
```diff
@@ -1,24 +0,0 @@
-class Random {
```
vasslitvinov: Please mention in the PR message why you are removing this.

compiler/AST/iterator.cpp
vasslitvinov: Please follow up on this.

vasslitvinov: Ah nevermind, I see it now.

compiler/passes/normalize.cpp
```diff
@@ -989,52 +989,67 @@ static void normalizeReturns(FnSymbol* fn) {
...
+    }
+
+    if (fn->retTag == RET_TYPE) {
+      retval->addFlag(FLAG_TYPE_VARIABLE);
+    }
+
+    if (fn->hasFlag(FLAG_MAYBE_TYPE)) {
+      retval->addFlag(FLAG_MAYBE_TYPE);
+    }
+
```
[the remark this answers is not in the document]
mppf: I'm not so sure FLAG_MAYBE_TYPE or FLAG_MAYBE_PARAM make sense, since AFAIK we can't yield a param or a type from an iterator.

>  error: iterators may not yield or return parameters
>  error: iterators may not yield or return types

So I'll remove those two.

At this point in the compiler, we're supposed to know whether we're returning a ref or not by the function's return intent. This code calls insertRetMove which adds PRIM_ADDR_OF or PRIM_DEREF depending on whether or not the function returns with ref intent.

compiler/passes/normalize.cpp
```diff
@@ -989,52 +989,67 @@ static void normalizeReturns(FnSymbol* fn) {
...
+    yield->insertBefore(new DefExpr(retval));
+    insertRetMove(fn, retval, yield);
+    yield->insertBefore(new CallExpr(PRIM_YIELD, retval));
+    yield->remove();
```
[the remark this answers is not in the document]
mppf: I have the opposite experience. When we replace only part of the expression, it's harder to see where in the compiler it changed. I'd like to leave it as-is and consider it a "style" choice.

test/arrays/ferguson/from-iterator/for-yields-no-arrays.future
mppf: I made issue  #8156 for it, I just forgot to put it in the .future. I've fixed that now, thanks for pointing it out!

test/modules/figueroa/DuplicateConfigVars/RandomNumber.chpl
mppf: Done (but I also reshuffled a bit...)

bradcray:
> Returning an iterator record now returns an array ( see issue #7994 for discussion ). That includes the case of returning a promoted expression and also the case of returning a call to an iterator in a non-iterator function. Since this might change the behavior of existing programs, for this release there is a warning when that happens.

I'm wondering whether this warning is sufficiently useful / non-confusing to warrant inclusion in the release.  I understand that the behavior of some programs will have potentially changed, but looking through the test modifications in this PR, it seems pretty innocuous in most cases, and I'm not sure that the warning sends the message that it's trying to inform you of a potential semantic change since version 1.16 of Chapel.  Put another way, when I get a warning, it makes me assume I've done something wrong, but this warning pops up for programs that seem legal / reasonable to me, and therefore makes me worry that I did something wrong when I very well may not have.  I also feel that there are likely many things that change between releases that do not generate warnings.  Thus, I'm wondering how exposed we would feel if 1.17 did not have this warning?

mppf:
I'd be fine with not putting the warning this release. But, would simply rewording it help? Or, if we added something like `--warn-unstable` which I've wanted for things like changes to what `new` does? So people could request warnings when the compiler detects uses of language elements that have changed recently / will change soon?

bradcray:
Do we have a most realistic case of when a user's code would be sensitive to this change in a way that would surprise them?  Most of the diffs I recall from the testing changes seemed more related to things like how many copies were done or when they were done than things that I would expect to impact the semantics or behavior of real user code.  Without an example that makes me think "Yeah, lots of users might fall into this trap", having the warning on by default (with this wording or another one) makes it feel as though the user's done something they shouldn't when that doesn't feel like the case to me for this change.  Changing it from a warning to a note would help somewhat, but would still add noise to compilation for what's presumably an otherwise correct program.

I wouldn't have a problem with a way to opt into warnings (or notes) that pointed things like this out as you propose with `--warn-unstable`, but I also expect that not many people would learn about it / use it if it weren't on by default (I'm thinking about how hard it is to get people to realize that the `--fast` flag exists, for example).  And if it were on by default then it feels back to the former paragraph again.  But I suppose that, like `--fast`, once people said "my code seems to have broken for some reason I can't guess", having them flip the flag to see if it pointed things out to them could help them/us (though moreso if it tracked lots of things rather than just this one).

mppf:
I'll look at adding a `--warn-unstable` developer (for now) flag and limiting this check to that.

Issue #8077, opened 2017-12-20
Function hijacking


mppf:
CHIP 20 describes several examples where the Chapel language allows "function hijacking" - that is, a situation when a library might unintentionally change behavior based upon changes in other libraries or in code using that library.

This issue tracks discussion related to that CHIP.

mppf:
CHIP 20 suggests, for the case of hijacking by (accidentally) implementing a more specific function, that it might be possible to simply put 'more visible' before 'more specific' in the list of arguments.

That strategy is challenging when working some generally overloaded function, such as _getIterator, _cast, _new, init, this, these, etc. that should be overridden specifically for each type. In particular, in the event that there is a generic function providing a default implementation, the default implementation might be selected instead of one specifically tailored to the type in question. (That pattern can be handled with FLAG_LAST_RESORT).

Additionally, this approach presents problems with promotion. If promotion is less important than visibility, some obvious programs don't work (Say a module defines CustomRecord.rank and then they have Array:[1..n] CustomRecord - then Array.rank will resolve to the promoted rank over the array rather than returning the Array.rank). If promotion is more important than visibility, programs relying on a promoted `foo(x:int)` call might accidentally start calling a `foo(x:[] int)` provided by a library. (but overloads sets would solve that last problem).

vasslitvinov:
The adjustment to the POI rule in #15948, currently being implemented in #16158, prefers "more visible" over more specific.

For example, this test:
```
test/functions/ferguson/hijacking/Application4.chpl
```
which was a future prior to #16158, with that PR exhibits the behavior expected by the .good .

PR #8079, opened 2017-12-20
Stop using point-of-instantiation in isMoreVisible

mppf:
Function disambiguation rules use a function, isMoreVisible, to tell when one
function is more visible than another. This function has historically treated
the point of instantiation of a generic as the "parent block" for that
function. However that leads to confusing program behavior in some cases,
as described in CHIP 20 and its connected issue #8077.

This behavior originated in commit 1da623b5f50906858891056584d5187d1bb478a0.


mppf:
Closing for now.

mppf:
I think this is ready but I'm going to wait until tomorrow to merge it to give a little more time for feedback.

PR #8082, opened 2017-12-21
wrappers.cpp: remove reference to nonexistent test

cassella:
There's no such test as ``test/functions/bradc/defaultVals/defaultArrCausesCopy-reffn.chpl``, so remove this reference to it.


cassella:
@mppf I can tell that this test doesn't exist, and doesn't seem to have ever existed.  I can't tell if you meant another test or had one in mind.  Feel free to reject the PR if removing the reference isn't the right thing to do.

mppf:
Removing it is fine with me.

PR #8092, opened 2017-12-23
Fix plus reduce in csrmatvecMult

dmk42:
#8089 fixed a race condition but missed some accumulations of the reduction variable on certain architectures, in the numa and knl locale models.  This restores the accumulations but retains the fix to the prior race condition.


dmk42:
@ben-albrecht :  I'm going to merge this to quiet testing noise, but would you please do a post-commit review?

PR #8099, opened 2018-01-02
joinPath() function issue #6012

victor-ludorum:
Adding **joinPath()** function to **path.chpl** file . This function is helpful in joining path names . 
@lydia-duncan Kindly review this , references #6012 and suggest necessary changes

PR #8100, opened 2018-01-02
Testfile for joinPath() function

victor-ludorum:
Adding **testfile and expected output file** for the **joinPath()** function added in Path.chpl .
@lydia-duncan Kindly review this with respect to PR #8099 .


lydia-duncan:
Usually we try to keep test files for a change in the same PR as the change, would you mind committing them on that PR?

victor-ludorum:
Okay @lydia-duncan I will do the same .

PR #8045, opened 2017-12-15
Buffered reverse-complement studies

benharsh:
These versions are based off of Rust entries 2 and 3, and mimic Rust's ``BufReader`` struct. One version reads entire sections at a time, and the other reads line-by-line. Both versions use a buffer to reduce the number of IO operations, and copy bytes into a given 1D array up until a particular character (> or \n) is found.

The rust versions parallelize differently, but these entries are mainly to test an alternate approach to IO.

The line-by-line version is currently performing significantly worse, likely due to slicing overhead. PR #8022 added a test dedicated to measuring creation time for array views.

The 'entire sections at a time' version is competitive with the top C entry.

Issue #7954, opened 2017-12-06
I/O Design Meta-Issue


mppf:
This issue is just meant to list and track I/O design issues (or at least issues that potentially will change the I/O API).

- [ ] writeThis and inheritance #7118
- [ ] error handling: returning errors from writeThis #7261
- [ ] Supporting I/O of class instance cycles? #7951
- [ ] I/O: When should output be machine readable? #7952 
- [ ] when should an array be resized when it is read? #5053 (also #7524)
- [ ] Improve Channel Formats #7846 
- [ ] Remote file local channel #7953
- [ ] parallel file.lines() #4959

bradcray:
Related: @benharsh has been looking at I/O overheads in the context of the revcomp benchmark and expects to put together a list of I/O issues / wishes in response to that exercise, as I understand it.

benharsh:
Question: does anything in the release/examples/ directory rely on the functionality we intend to deprecate?

benharsh:
My take on a rough prioritization of these tasks (grouped by coarse prioritization):
1. Improve/add channel formats
    Rationale: probably deprecates "%j", breaking existing code
2. machine readable output
    Rationale: behavior-changing, possibly significantly
3. IO plugins written in Chapel
    Rationale: deprecates HDFS, CURL from stdlib
-------------
4. Return errors from writeThis
    Rationale: Future programs may break if they do not handle errors.
5. Arrays resized when read (or, "how to read an unknown number of things")
    Rationale: Useful idiom. **Note:** May depend on whether we keep mutate-through-array functionality.
---------------
6. Buffer/bytes/strings with channels
    Rationale: Useful idiom.
7. Methods beginning with underscore
    Rationale: Solidify API
8. DefaultRectangular and readBytes
    Rationale: may break existing code
9. allocate-as-read classes
    Rationale: Useful idiom.
10. class instance cycles
    Rationale: Annoying for users
11. writeThis inheritance
    Rationale: Annoying default behavior, but can be overrided
---------------
12. Remote file, local channel
    Rationale: correctness, productivity > performance. Though probably a common idiom.
13. parallel file.lines()
    Rationale: correctness, productivity > performance. Though this may have design repercussions.
14. Spinwaiting under qthreads
  Rationale: correctness, productivity > performance

e-kayrakli:
It might be good to keep a record of code style issues with the IO module here. See https://github.com/chapel-lang/chapel/issues/14646

[...]

PR #8103, opened 2018-01-02
Add channel.advancePastByte, use it to improve revcomp

### LINKED ISSUE #8105: adding channel.advancePastByte ?
I've been working with @benharsh on some revcomp benchmarks game I/O speed issues.
I'm proposing to add a method to IO.chpl called channel.advancePastByte that reads
until a particular byte is found and leaves the channel cursor just after that byte.
(And raises an EOF error if the byte is not found).

This function enables succinct expression of the fastest known I/O pattern for revcomp.
It could be used in other contexts as well, for example, we could use it in the implementation
of readln to skip characters until the newline.

See also PR #8103.

``` chapel
/* Reads until ``byte`` is found and then leave the channel offset
    just after it. If that byte is never found, raises an UnexpectedEOFError. */
proc channel.advancePastByte(byte:c_int) throws
```

mppf:
PR #8045 added some messy rev-comp studies that improve performance by improving the I/O pattern. What the performance comes down to is two things:
 1. Copying large chunks of the input from the channel buffer in to the array to be used
 2. Using memchr to identify the relevant chunks of the input

I experimented with a version that used regexp format strings to replace the memchr call but that had unsatisfying performance.

This PR adds channel.advancePastByte in order to enable the expression of the fast I/O pattern in revcomp easily. Now the revcomp version does the following:
 * "mark" (indicate to the I/O system not to drop the buffer as we might return)
 * identify the offset of the newline (end of the sequence description)
 * identify the offset of the > (start of the next sequence)
 * "revert" (go back to where we "marked")
 * read the data again in one go with readBytes

I'm seeing a 10% speedup for this version beyond revcomp-buf.chpl, and it is much simpler.

While there, I noticed that qio_channel_advance might not set up the buffer in some cases, so added code to do that.

- [x] full local testing
- [x] docs for advancePastByte
- [x] update catch statement to specify type
- [x] check for 32-bit issues ala #8116

Closes #8105.


test/studies/shootout/reverse-complement/bharshbarg/revcomp-mark-skip-read-begin.chpl
```diff
@@ -0,0 +1,87 @@
...
+
+      // Scan forward until we get to the \n (end of description)
+      input.advancePastByte(ascii("\n"));
+      seqOffset = input._offset();
+
+      try {
+        // Scan forward until we get to the > (end of sequence)
+        input.advancePastByte(ascii(">"));
+        nextDescOffset = input._offset();
+      } catch UnexpectedEOFError {
```
psahabu: Does the compiler currently support this pattern? My thought is that it would need to be `catch e: UnexpectedEOFError`.

mppf: seems to work...

psahabu: Huh, cool.

bradcray:
Stab in the dark to head off potential testing noise: does the need for PR #8116 have any implications for this one?

test/studies/shootout/reverse-complement/bharshbarg/revcomp-mark-skip-read-begin.chpl
mppf: I bet that it's not matching on the type but rather I'm naming an error variable `UnexpectedEOFError` and it shadows the type. I.e. it's like I wrote `catch e`. I'll update it.

mppf:
@benharsh - could you review this PR? we certainly talked about it. It passed full local testing in an earlier version - after review I plan to run that again. Thanks.

mppf:
@benharsh - not sure where to put an example but at least it's a documented function...

As far as adjusting readBytes, I made a note of that issue #7954.

Issue #5122, opened 2017-01-06
Add support for real(32) params


bradcray:
This issue relates to test/param/errors/param32BitFloat.future which was added in PR#5072.

It seems that we don't / can't support real(32) params today due to the fact that the _defaultOf() routines can't be written in a way that returns a param 32-bit 0.0 value.  In user code, we'd typically write "0.0:real(32)" but we can't do this cast at param-evaluation time because we have traditionally not done floating point operations at compile-time (for fear that the floating point semantics might differ at compile-time from what the user would get at execution time -- perhaps this is being overly paranoid, though, since the user did request us to do the computation at compile-time.... but that's a separate issue).

This seems like it's probably just a bootstrapping problem and could potentially be handled by special-casing this value using a primitive or somesuch.

nimitbhardwaj:
Hey, I am encountering a bug here, its related to the behaviour of real(32).

What I did was simply

```chapel
var a:real(32) = -1.5;
writeln(a);
```
Its giving error `a.chpl:1: error: type mismatch in assignment from real(64) to real(32)`
while for the code
```chapel
var a:real(32) = 1.5;
writeln(a);
```
there is no error, smooth compilation, there is an error related to parsing of negative values which are real(32), in my sense real(32) is a float in C or C++, so there are 11 bits for decimal part and one bit for sign, it should work, -1.5 is also a real(32) or float, why this is, am I wrong ??

One more thing, it works correctly for

```chapel
var a:real(32) = -100;
writeln(a);
```
This above program is compiled correctly, only negatives with a decimal provides error

bradcray:
The issue in your code is that on master today:
* we do permit coercions from `int(64)`s (like `-100`) to `real(32)`s when there is no loss of information (@mppf: did I get that right?), but
* we don't permit coercions from `real(64)`s (like `-1.0`) to `real(32)`s (ever).
The workaround for the time being is to do an explicit downcast, like: `(-1.0):real(32)`

We have discussed adding support for automatic downcasting from `real(64)` values to `real(32)` values in the event that there would be no loss of information, but that has not yet been implemented.


> there is an error related to parsing of negative values

The parsing of the values is not the problem here, it's the automatic conversion from a more precise floating point types to less-precise ones.

[...]

PR #8111, opened 2018-01-03
Add Bessel Functions in standard `Math` module by wrapping std C functions

### LINKED ISSUE #7646: Bessel functions
It would be great to have the bessel functions available in the math module.

nimitbhardwaj:
Add the Bessel functions in the standard `Math` module, by wrapping the
standard C library functions. Add `j0`, `j1`, `jn`, `y0`, `y1` and `yn` procedures

Closes #7646


modules/internal/ChapelArray.chpl
```diff
@@ -1650,6 +1650,11 @@ module ChapelArray {
        return translate(offTup);
      }
 
+     /* Return true if the domain has no elements */
```
ben-albrecht: I think this change was accidentally included in this PR

bradcray:
I've only taken the briefest of looks (mostly to see how 32- vs. 64-bit reals were handled), but this looks like a nice first/second PR to me.  Thanks!

nimitbhardwaj:
This is done, error was I branched out from other branch accidentally, error handled

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,66 @@ module Math {
     return truncf(x);
   }
 
+
+  /* Returns the Bessel function of the first kind of order `0` of `x`. */
+  extern proc j0(x: real(64)): real(64);
+
+  inline proc j0(x : real(32)): real(32) {
+  /* Returns the Bessel function of the first kind of order `0` of `x`. */
```
ben-albrecht: Note that these doc-strings need to be pulled out of the procedure body in order to get rendered in the documentation. Right now it looks like this:

screen shot 2018-01-10 at 11 00 20 am

nimitbhardwaj:
@ben-albrecht , OK, I do it, I think I didn't notice it, just I need to do is pragma "no doc"

nimitbhardwaj:
no sorry, just put the comment on the function, my bad

ben-albrecht:
@nimitbhardwaj  - No problem! In case you weren't aware, you can build and view the docs locally with `make docs` from `$CHPL_HOME`.

nimitbhardwaj:
Now it will may be correct, but as I told before, the Bessel function of 2nd kind is complex, because C didn't support complex numbers originally, but Chapel supports them, so the function y0, y1, yn return NaN for negative numbers, so I think, we need to get some implementation for y0, y1, yn for negative values, so other way for them is needed, may be we have to put this as issue further, if this PR is accepted

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,66 @@ module Math {
...
+
+  /* Returns the Bessel function of the first kind of order `1` of `x`. */
+  inline proc j1(x : real(32)): real(32) {
+    extern proc j1f(x: real(32)): real(32);
+    return j1f(x);
+  }
+
+
+  /* Returns the Bessel function of the first kind of order `n` of `x`. */
+  extern proc jn(n: int, x: real(64)): real(64);
```
mppf: Note, Chapel int != C int; Chapel int corresponds to int64_t.
It would be possible to use c_int in the argument but a better practice is probably to use safeCast.

nimitbhardwaj: OK, thanks I make changes, I didn't thought of that, that time, thanks

nimitbhardwaj: Its done

mppf: I think you've correctly updated the 'extern proc' but I think the Chapel versions should use `int` as the argument type and then use `safeCast` to convert from int to c_int.

nimitbhardwaj:
now is fine may be

mppf:
I think the integer arguments should be `int` for Chapel and then use `safeCast` to convert to `c_int`. See e.g. https://github.com/chapel-lang/chapel/blob/master/modules/standard/BigInteger.chpl#L152

ben-albrecht:
> we need to get some implementation for y0, y1, yn for negative values, so other way for them is needed, may be we have to put this as issue further, if this PR is accepted

I am OK to follow libc and document that negative values are not supported for now. Implementing these procedures would be an effort on a different scale than what is done in this PR. Many languages opt for calling out to [GSL](https://www.gnu.org/software/gsl/manual/html_node/Bessel-Functions.html) or [netlib/amos](http://www.netlib.org/amos/), and [netlib/cephes](http://www.netlib.org/cephes/) in their math libraries rather than re-implementing, so that's something to consider as well.

I think this could be pursued as a follow-up effort.

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,66 @@ module Math {
...
+  extern proc y0(x: real(64)): real(64);
+
+  /* Returns the Bessel function of the second kind of order `0` of `x`. */
+  inline proc y0(x : real(32)): real(32) {
+    extern proc y0f(x: real(32)): real(32);
+    return y0f(x);
+  }
+
+
+  /* Returns the Bessel function of the second kind of order `1` of `x`. */
```
ben-albrecht: Let's document that negative values of `x` are not supported.

nimitbhardwaj: yes, its good, then I think a test for only positive values will be cool for bessel function of second kind, what should be done for negetive values, an exception is thrown or what?

ben-albrecht: We might want this to throw an error eventually (@psahabu can confirm if that's our intention in `Math.chpl` eventually), but for now, let's stay consistent with our procedures in the module by halting if `x < 0`.

daviditen:
All of the new tests are testing only the 64-bit functions.  It would be good to also test the 32-bit functions.  I tried extending the j0.chpl test to also test 32-bit and on my OSX computer, it fails during the C compilation step with errors like:

```
j0.c:75:19: error:
      implicit declaration of function 'j0f' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
call_tmp_chpl15 = j0f(call_tmp_chpl10);

j0.c:75:19: note:
      did you mean 'j0'?
/usr/include/math.h:687:15: note: 'j0' declared here
extern double j0(double) __OSX_AVAILABLE_STARTING(__MAC_10_0,__IPHONE_3_2);
```

My understanding is that only the `double` versions of these functions are standard.

modules/standard/Math.chpl
nimitbhardwaj: yes, I did

nimitbhardwaj:
yes, there is error arising, when i convert, int to c_int, it starts, it says, there are 2 candidates for jn function

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,66 @@ module Math {
     return truncf(x);
   }
 
+
+  /* Returns the Bessel function of the first kind of order `0` of `x`. */
+  extern proc j0(x: real(64)): real(64);
+
+  /* Returns the Bessel function of the first kind of order `0` of `x`. */
+  inline proc j0(x : real(32)): real(32) {
```
daviditen: I'm OK with the way these 32-bit wrapper functions are written, but am just noting that they could also be written as a [renaming extern declaration](https://chapel-lang.org/docs/master/technotes/extern.html#declaring-external-c-functions) (example near the end of this section):

`extern "j0f" proc j0(x: real(32)): real(32);`

nimitbhardwaj: On my PC using Linux, it was not giving error for that, I also tried, `var a: real(32) = 10;`, it worked fine, but if in mac its giving error, I have to change it

nimitbhardwaj:
Hey, there is a serious error, I don't know whether only I am encountering, it is regarding keyword compilerError, for code
`var a = 10;`
`if a == 10 then compilerError("Error");`
and 

`var a = 10;`
`if a != 10 then compilerError("Error");`

in both the cases, I am getting compiler error, I don't know why

ben-albrecht:
> in both the cases, I am getting compiler error, I don't know why

It is because the value of the variable `a` is only known at runtime, but you have a `compilerError` dependent on its value. For this case, you'll want a `halt()`, which is a runtime error.

nimitbhardwaj:
OK, thanks, I was revolving around this mistake, thanks

mppf:
(For your future knowledge) that if example with compilerError would have to declare the variable as `param a = 10` to work as you intended.

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,72 @@ module Math {
...
+  inline proc jn(n: int, x : real(64)): real(64) {
+    extern proc jnf(n: c_int, x: real(64)): real(64);
+    return jnf(n: c_int, x);
+  }
+
+
+  /* Returns the Bessel function of the second kind of order `0` of `x`. */
+  extern "y0f" proc y0(x: real(32)): real(32);
+
+  /* Returns the Bessel function of the second kind of order `0` of `x`. */
```
ben-albrecht: We should mention that `y*` Bessel functions do not support `x < 0` values in the documentation string

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,72 @@ module Math {
...
+  inline proc y0(x : real(64)): real(64) {
+    if x < 0 then
```
ben-albrecht: @mppf suggested this check should be omitted when `boundsChecking` is disabled, so I would update my suggestion to:

```chpl
if boundsChecking {
  if x < 0 then
    halt("Input value for Bessel function of second kind must be greater than 0");
}
```

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,72 @@ module Math {
...
+
+  /* Returns the Bessel function of the first kind of order `1` of `x`. */
+  inline proc j1(x : real(64)): real(64) {
+    extern proc j1(x: real(64)): real(64);
+    return j1(x);
+  }
+
+
+  /* Returns the Bessel function of the first kind of order `n` of `x`. */
+  extern "jnf" proc jn(n: int, x: real(32)): real(32);
```
ben-albrecht: Note that `int != c_int`. Chapel `int` is implicitly `int(64)`, whereas `c_int` is a type alias for `int(32)`, the default `int` size for C.

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,72 @@ module Math {
...
+  inline proc y1(x : real(64)): real(64) {
+    if x < 0 then
+      halt("Input value for Bessel function of second kind must be greater than 0");
+    extern proc y1(x: real(64)): real(64);
+    return y1(x);
+  }
+
+
+  /* Returns the Bessel function of the second kind of order `n` of `x`. */
+  extern proc yn(n: int, x: real(32)): real(32);
```
ben-albrecht: Another `int` that should be `c_int`.

modules/standard/Math.chpl
mppf: The default int size for C isn't even necessarily int(32) although that is the currently common configuration.

nimitbhardwaj:
Its done

modules/standard/Math.chpl
ben-albrecht: @nimitbhardwaj  - this still needs to be updated to `n: c_int`

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,96 @@ module Math {
...
+    extern proc y1(x: real(64)): real(64);
+    return y1(x);
+  }
+
+
+  /* Returns the Bessel function of the second kind of order `n` of `x`,
+     if and only if the value of `x` is greater than 0, if 
+     `boundChecking` is `true` then for `x < 0` there will be an error,
+     otherwise an `NaN` value will be returned*/
+  extern "y1f" proc yn(n: int, x: real(32), boundChecking = true): real(32);
```
ben-albrecht: This still needs to be updated to `c_int` as well.

mppf: This should look like the yn below. You can't use the `extern "y1f" proc yn` syntax here because we need the Chapel function to convert from int to c_int.

modules/standard/Math.chpl
mppf: This should look more like jn below, since we have to cast. The extern proc must have c_int not int in the type.

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,96 @@ module Math {
...
+     if and only if the value of `x` is greater than 0, if 
+     `boundChecking` is `true` then for `x < 0` there will be an error,
+     otherwise an `NaN` value will be returned*/
+  extern "y1f" proc y1(x: real(32), boundChecking = true): real(32);
+
+  /* Returns the Bessel function of the second kind of order `1` of `x`,
+     if and only if the value of `x` is greater than 0, if 
+     `boundChecking` is `true` then for `x < 0` there will be an error,
+     otherwise an `NaN` value will be returned*/
+  inline proc y1(x: real(64), boundChecking = true): real(64) {
```
ben-albrecht: `boundsChecking` is a global configuration param that is turned off when a Chapel program is compiled with `-sboundsChecking=true`, which is part of the `--fast` optimization flag. You can drop this argument in the procedures and rely on the global `boundsChecking` configuration param instead. We don't generally mention `boundsChecking` behavior in documentation, so you can drop mentions of `boundsChecking` in the docs as well.

Basically, when `boundsChecking` is on (which it is by default), users will get cleaner and more helpful error messages for out of bounds errors, such as `"Input value for Bessel function of second kind must be greater than 0"` in this case. When it is disabled, users lose those cleaner error messages in a trade off for performance. In this case, users will get whatever error message the C implementation spits out, when `boundsChecking` is off.

You can grep the `modules/` directory for `boundsChecking` to see other example its use.

modules/standard/Math.chpl
nimitbhardwaj: Oh sorry, just copied above lines, bad  in such names.

nimitbhardwaj: I want to ask a thing, proc yn is the wrapper function to yn <which belongs to C stdlib> for the user, the user will call yn procedure, giving n as input, this n will be integer wrt to chapel, then this wrapper will call C function, for that C function the arguments needed will be a C type integer which is a 32 bit int (in contrast to 64 bit integer of chapel), so the user is giving integer it should be int not c_int, and it is casted to C type integer at the time of calling, so I left it there in my sense, because at that time I didn't get the need to take c_int, because caller is chapel, chapel will provide the simple integers.

nimitbhardwaj: @mppf @ben-albrecht

nimitbhardwaj: oh, in this context, I understand it now, thanks

nimitbhardwaj:
Its done this issue

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,84 @@ module Math {
...
+
+  /* Returns the Bessel function of the first kind of order `1` of `x`. */
+  inline proc j1(x: real(64)): real(64) {
+    extern proc j1(x: real(64)): real(64);
+    return j1(x);
+  }
+
+
+  /* Returns the Bessel function of the first kind of order `n` of `x`. */
+  extern "jnf" proc jn(n: c_int, x: real(32)): real(32);
```
mppf: Still not right.
``` chapel
  /* Returns the Bessel function of the first kind of order `n` of `x`. */
  inline proc jn(n: int, x: real(32)): real(32) {
    extern proc jnf(n: c_int, x: real(32)): real(32);
    return jn(n.safeCast(c_int), x);
  }
```

modules/standard/Math.chpl
```diff
@@ -1057,6 +1057,84 @@ module Math {
...
+        halt("Input value for Bessel function of second kind must be greater than 0");
+    }
+    extern proc y1(x: real(64)): real(64);
+    return y1(x);
+  }
+
+
+  /* Returns the Bessel function of the second kind of order `n` of `x`,
+     if and only if the value of `x` is greater than 0 */
+  extern "ynf" proc yn(n: c_int, x: real(32)): real(32);
```
mppf: here too

modules/standard/Math.chpl
nimitbhardwaj: so, it means explicily I had to write the function, oh sorry, I didn't understood it, i thought working correct it was correct, no problem ;-)

nimitbhardwaj:
This PR should be merged, I solved the requested changes

daviditen:
As I mentioned 16 days ago, I'd still like to see the 32-bit versions of these functions tested in the new tests.

nimitbhardwaj:
Oh, I am extremely sorry for that I missed that comment, sorry for that, I make the adjustments now.

nimitbhardwaj:
Hey, I am encountering a bug here, its not related to my work, its related to the behaviour of real(32).

What I did was simply

```chapel
var a:real(32) = -1.5;
writeln(a);
```
Its giving error `a.chpl:1: error: type mismatch in assignment from real(64) to real(32)`
while for the code
```chapel
var a:real(32) = 1.5;
writeln(a);
```
there is no error, smooth compilation, there is an error related to parsing of negative values which are real(32), in my sense real(32) is a float in C or C++, so there are 11 bits for decimal part and one bit for sign, it should work, -1.5 is also a real(32) or float, why this is, am I wrong ??

nimitbhardwaj:
Please everyone check it, or if its unknown issue then perhaps create an issue.

nimitbhardwaj:
Or only I am encountering it

mppf:
I think it's issue #5122. It might help for you to comment with your example on that issue.

nimitbhardwaj:
OK, thanks I put it there

bradcray:
@nimitbhardwaj : As a workaround, you should be able to use `var a = (-1.0): real(32);`

nimitbhardwaj:
Yes, its an alternative here, it can be done, but hope original will also work soon, may be a compiler error, can't say anything of it, by the way thanks.

nimitbhardwaj:
I made it

bradcray:
@nimitbhardwaj : It's not a compiler error per se, it's the issue Michael pointed to above:  `1.0` is a `real(64)`, so negating it yields a `real(64)` and we don't currently support automatic downcasts from `real(64)` to `real(32)` (but are discussing changing that).

nimitbhardwaj:
Okk, perhaps it should be changed, as like in many other languages, like C or C++, it is done, assigning this way works, may be only get warning sometimes.

nimitbhardwaj:
They are unsupported, but why, I add this file, may be because there is some difference between Clang and gcc or something ??

nimitbhardwaj:
I have added for darwin os

daviditen:
> They are unsupported, but why, I add this file, may be because there is some difference between Clang and gcc or something ??

According to a FreeBSD [man page](https://www.freebsd.org/cgi/man.cgi?apropos=0&sektion=3&query=y0f&manpath=FreeBSD+7.0-current&format=html):

> The j0(), j1(), jn(), y0(), y1(), and yn() functions conform to IEEE Std
> 1003.1-2001 (``POSIX.1'').  The float versions are extensions.

So I think the float versions are extensions that aren't available on OSX.

nimitbhardwaj:
Okk

nimitbhardwaj:
Is it fine?

daviditen:
My concerns have been addressed, thanks!

nimitbhardwaj:
Problem is that, I don't have mac, I don't know what is error, they don't have function or that.
And I can't afford mac now, but hope of it some day.

mppf:
My concerns were also addressed, so at this point I'm happy if @ben-albrecht is happy.

Issue #7939, opened 2017-12-05
autoCopy internal error


nspark:
### Summary of Problem

An internal error (`FUN0334`) is encountered when returning an object from a function. (**Edit**: removed the generic aspect, as this produces the error without it.)

### Steps to Reproduce

**Source Code:**

```chapel
// error-2.chpl
record Foo {
  proc init() {
    super.init();
  }
}
proc quux(x: Foo) return x;
var foo = new Foo();
var bar = quux(foo);
```

**Compile command:**
```
$ chpl error-2.chpl
error-2.chpl:1: internal error: FUN0334 chpl Version 1.17.0 pre-release (febd6ff96f)

$ CHPL_DEVELOPER=1 chpl error-2.chpl
error-2.chpl:1: internal error: Trying to obtain autoCopy for type 'Foo()', which defines none [functionResolution.cpp:334]
```

### Configuration Information

- Output of `chpl --version`: 1.17.0 pre-release (febd6ff96f)

lydia-duncan:
Thanks for reporting!  This looks like a failure on our part to warn you about not defining a copy initializer

lydia-duncan:
Notes for future attempts to solve this problem:
This is occurring during the callDestructors oass, when we try to insert an autoCopy for the return value from quux because it is returning the argument.  However, we don't create autoCopies for types that define initializers (because we try to resolve autoCopies before we know if they are needed, and this causes errors in the case where the copy initializer is not defined or a generic, one argument function is thought to be a copy initializer when really it isn't one), so getAutoCopyForType() fails, leading to the internal error listed here.
I attempted to insert a call to an initializer and resolve it when we know the type has initializers, but that didn't work because resolution is not able to be called this late in compilation (which is why we resolved the autoCopies early and stored them in the autoCopy map in the first place).

In short, solving this problem will require a significant rework of the callDestructors pass and possibly resolution as well.  We want to do this work, but it likely won't happen for a while, unfortunately.

bradcray:
I believe that I ran into this same error today while trying to convert more internal / standard modules over from constructors to initializers.  So we may not be able to put it off for _too_ long.

lydia-duncan:
Michael and I are going to try to pair program on this while he is here.

lydia-duncan:
#8064, #8065, and #8066 are related to the effort to fix this issue.  Michael and I would welcome input

PR #8115, opened 2018-01-03
BigInteger initializers conversion

bradcray:
This converts the BigInteger module from using constructors to initializers.  My original effort was thwarted by issue #7939, now resolved.

As part of this effort, the old-style autocopy and initcopy functions were able to be removed, which is attractive.


bradcray:
@lydia-duncan and/or @noakesmichael : Any interest in reviewing?

Issue #8023, opened 2017-12-13
Support slicing of sparse domains


buddha314:
I'm not sure how much overlap there with [this issue](https://github.com/chapel-lang/chapel/issues/7374), but it would be very nice to be able to traverse rows or columns (and sets thereof) in sparse matrices.

```chapel
var D: domain = [1..5];
var SD: sparse subdomain(D);
var A: [SD] real;
SD += (..);

// maybe something like
for c in A[3,..] {
  writeln("col ", c, " is ", A[3,c]);
}
```

Good things to implement on top of this are
* Number of non-zero entries in that row
* reduce intents like make, min, sum on that row
* Assignment of values to that row

buddha314:
In the mean time, I wrote this terrible hack to get me the actual neighbors. Is there an intermediate solution that won't be so slow?

```chapel
  proc Graph.neighbors(vid: int) {
    var result: [1..0] int;
    for (i,j) in this.A.domain {
      if i == vid {
        result.push_back(j);
      }
    }
    return result;
  }
```

bradcray:
You're interested in the neighbors of a given row, right?  I'd do it as follows:

```chapel
use LayoutCS;

const D = {1..4, 1..4};
var SD: sparse subdomain(D) dmapped CS();
SD += (1,2);
SD += (2,1);
SD += (3,3);
SD += (3,4);
SD += (4,1);
SD += (4,4);

iter neighbors(SparseDom, row) {
  for col in SparseDom.dimIter(2, row) do
    yield col;
}

config const row = 3;

for col	in neighbors(SD, row) do
  writeln(col, " is a neighbor of ", row);
```

buddha314:
Is there a cool method like that for the diagonal?  I don't see how you would use `.dimIter`.  I want to set the diagonal of a sparse matrix to 0.

buddha314:
And how do I put an `if` in there, like
```chapel
iter neighbors(SparseDom, row) {
  for col in SparseDom.dimIter(2, row) do
    if col != row yield col;
}```

E.g., I want to eliminate the diagonal.  You may be sensing a theme.

bradcray:
Putting the `if` in there just like that ought to work.  Does it not?

buddha314:
Yes, like on [Stack Overflow](https://stackoverflow.com/questions/47823225/chapel-iterators-with-conditionals/47823373#47823373)  When you're right, you're right.  And you, my friend, are right.

buddha314:
[...]

PR #8117, opened 2018-01-03
Support stridable queries on sparse domains

bradcray:
While investigating support for slicing of sparse domains/arrays in support of issue #8023, the first thing I ran into was the lack of support for querying the stridability of sparse domains.  This change adds the ability for sparse domains to support a `stridable` query that is implemented by forwarding the query on to their parent domains.  It also adds tests to make sure this works with our three main sparse domain maps today, and that it works when creating sparse domains of other sparse domains.

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -1084,13 +1084,8 @@ module ChapelArray {
     }
 
     /* Return true if this is a stridable domain */
-    proc stridable param where isRectangularDom(this) {
+    proc stridable param where isRectangularDom(this) || isSparseDom(this) {
       return _value.stridable;
-    }
-
-    pragma "no doc"
-    proc stridable param where isSparseDom(this) {
-      compilerError("sparse domains do not currently support .stridable");
     }
 
     pragma "no doc"
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -1084,8 +1084,13 @@ module ChapelArray {
     }
 
     /* Return true if this is a stridable domain */
-    proc stridable param where isRectangularDom(this) || isSparseDom(this) {
+    proc stridable param where isRectangularDom(this) {
       return _value.stridable;
+    }
+
+    /* Return true if this sparse domain's parent is stridable */
+    proc stridable param where isSparseDom(this) {
+      return _value.parentDom.stridable;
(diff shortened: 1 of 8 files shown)


ben-albrecht:
Do we expect arrays to now support stridable method too, or is this PR strictly for sparse domains?

bradcray:
> Do we expect arrays to now support stridable method too, or is this PR strictly for sparse domains?

This PR is only focused on the sparse domain case because that was the case I ran into while working on sparse slicing.  It looks to me like dense arrays don't support stridable queries currently, though I could be misinterpreting.  If/when dense arrays do, sparse probably should as well.

ben-albrecht:
> It looks to me like dense arrays don't support stridable queries currently, though I could be misinterpreting. 

Sorry I was not clear - I was curious about what error we would get when trying to slice a sparse array with this branch:

```chpl
const D = {1..10};
var sD: sparse subdomain(D);
var A: [sD] int;
sD += [1,2,3,4];
writeln(A[1..4]);
```

This still fails with:

```
sps.chpl:5: error: unresolved call 'DefaultSparseDom(1,int(64),domain(1,int(64),false)).stridable'
```

I know that we don't yet expect sparse array slicing to magically work, but is this error expected?

bradcray:
> is this error expected?

I didn't expect it, and think that my first commit would've fixed it, but that my final one would not.  My guess is that the issues is that the slicing code is trying to call `stridable` not on the record implementing the sparse domain, but on the underlying class.  And for some reason, DefaultSparseDom is the one sparse domain class we have that doesn't have a `stridable` field.

I think that this PR is still valuable in its exposure of `stridable` to users (seems like a no-brainer, and I can't think of any reason we haven't in the past), but your observation points out that more work is required to get past this first bump toward sparse array slicing.

buddha314:
For those of us playing the home game, would it be worth it to provide an example?  Does this commit provide something I may be able to use directly?

bradcray:
@buddha314 : This PR just permits you to say:

```chapel
mySparseDomain.stridable
```

for sparse domains as you already could for dense domains where it'll return "true" if the parent domain is stridable, "false" otherwise.  I.e., given:

```chapel
var mySparseDomain: sparse subdomain(D);
```

it essentially will tell you whether or not `D` is stridable.  So yes, it's usable, but it may not be terribly useful.

Issue #7960, opened 2017-12-06
Internal error in initializer resolution


nspark:
### Summary of Problem

When compiling a generic record that calls `this.init(...)`, I see an internal error (`INI0061`) in initializer resolution.

### Steps to Reproduce

**Source Code:**

```chapel
// bug.chpl
record Foo {
  type idxType;
  var D: domain(1, idxType);
  proc init(type idxType, space: range) {
    this.idxType = idxType;
    this.D = space;
    super.init();
  }
  proc init(type idxType, size: idxType) {
    this.init(idxType, 0..#size);
  }
}
var foo = new Foo(int, 10);
```

**Compile command:**
`chpl bug.chpl`

**Futures**
[test/classes/initializers/records/generics/siblingCall-generic.chpl](https://github.com/chapel-lang/chapel/tree/master/test/classes/initializers/records/generics/siblingCall-generic.chpl) #7969 
[test/classes/initializers/records/generics/siblingCall-moreComplex.chpl](https://github.com/chapel-lang/chapel/tree/master/test/classes/initializers/records/generics/siblingCall-moreComplex.chpl) #7969

### Configuration Information

- Output of `chpl --version`: 1.17.0 pre-release (febd6ff)

lydia-duncan:
At a guess, this is a bug with our handling of range declarations, or range declarations of that form.  I'll look into this shortly, thanks for reporting!

lydia-duncan:
Oh, weird.  This seems to be limited to records only.  The same code works just fine for classes.

PR #8121, opened 2018-01-04
Convert channel type from constructors to initializers

bradcray:
Converting the channel type to use initializers was a bit tricky in that they required:

* a translation of the old constructor to an initializer
* a types/params-only overload due to the way we declare channels in type signatures
* a copy initializer, where the implicit one created by the compiler caused segfaults and so could not be used
* a manual transcription of the default initializer that the compiler would normally make since the other initializers overrode it (a good motivator for the "opt-in to exposing the compiler initializer" feature that's been discussed)
  
In the code review, @mppf pointed out that some of the initializers are not intended to be end-user facing such that, if/when we had private methods (in the sense of "private to this module"), they could be tagged as such.  He also pointed out that even if we had a way of opting into making the compiler-provided initializer available, he may not want to make use of it stylistically, in order to make it abundantly clear that that initializer does nothing w.r.t. ownership.  I mentioned this in issue #8136.

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1945,16 +1945,6 @@ record channel {
   var _readWriteThisFromLocale:locale;
 }
 
-// TODO -- shouldn't have to write this this way!
-pragma "no doc"
-pragma "init copy fn"
-proc chpl__initCopy(x: channel) {
-  on x.home {
-    qio_channel_retain(x._channel_internal);
-  }
-  return x;
-}
-
 pragma "no doc"
 proc =(ref ret:channel, x:channel) {
   // retain -- release
@@ -1971,7 +1961,19 @@ proc =(ref ret:channel, x:channel) {
 }
 
 pragma "no doc"
-proc channel.channel(param writing:bool, param kind:iokind, param locking:bool, f:file, out error:syserr, hints:c_int, start:int(64), end:int(64), in local_style:iostyle) {
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool) {
+  this.writing = writing;
+  this.kind = kind;
+  this.locking = locking;
+  super.init();
+}
+
+pragma "no doc"
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool, f:file, out error:syserr, hints:c_int, start:int(64), end:int(64), in local_style:iostyle) {
+  this.writing = writing;
+  this.kind = kind;
+  this.locking = locking;
+  super.init();
   on f.home {
     this.home = f.home;
     if kind != iokind.dynamic {
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1968,8 +1968,29 @@ proc channel.init(param writing:bool, param kind:iokind, param locking:bool) {
   super.init();
 }
 
+//
+// Note that this is effectively the initializer that the compiler
+// would typically provide and that, by providing the next initializer
+// below, we have to write it out manually...  A good case for having
+// a means to "opt-in" to including the compiler-provided initializer?
+//
+pragma "no doc"
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool,
+                  home: locale, _channel_internal:qio_channel_ptr_t,
+                  _readWriteThisFromLocale: locale) {
+  this.writing = writing;
+  this.kind = kind;
+  this.locking = locking;
+  this.home = home;
+  this._channel_internal = _channel_internal;
+  this._readWriteThisFromLocale = _readWriteThisFromLocale;
+  super.init();
+}
+
 pragma "no doc"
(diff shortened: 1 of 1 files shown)


modules/standard/IO.chpl
```diff
@@ -1971,7 +1961,19 @@ proc =(ref ret:channel, x:channel) {
...
+  this.locking = locking;
+  super.init();
+}
+
+pragma "no doc"
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool, f:file, out error:syserr, hints:c_int, start:int(64), end:int(64), in local_style:iostyle) {
+  this.writing = writing;
+  this.kind = kind;
+  this.locking = locking;
+  super.init();
```
lydia-duncan: I would be tempted to have one of these initializers call the other one instead of having them both utilize the same Phase 1 code.  (Likely having the bottom one call the top one, though I would check with @mppf to see if the top initializer is okay or if it should really be calling into the bottom one in some form).

lydia-duncan: (Update for those following along at home: the above strategy currently encounters a bug for generic records, see #7960)

bradcray:
@mppf, would you review this when you get back to the office since you're the father of channels?  @lydia-duncan and/or @noakesmichael may want to take a look just to see a case that required a bunch of init() overloads as well.

modules/standard/IO.chpl
```diff
@@ -1971,7 +1961,54 @@ proc =(ref ret:channel, x:channel) {
...
+  on x.home {
+    qio_channel_retain(x._channel_internal);
+  }
+}
+
+//
+// Note that this is effectively the initializer that the compiler
+// would typically provide and that, by providing the next initializer
+// below, we have to write it out manually...  A good case for having
+// a means to "opt-in" to including the compiler-provided initializer?
```
mppf: I think it makes sense for this to be an explicitly written initializer because it should be 'private'. Can you check to see if making it 'private' works here? AFAIK it should only be called from _read_one_internal / _write_one_internal.

modules/standard/IO.chpl
```diff
@@ -1971,7 +1961,54 @@ proc =(ref ret:channel, x:channel) {
...
+  this.kind = kind;
+  this.locking = locking;
+  this.home = home;
+  this._channel_internal = _channel_internal;
+  this._readWriteThisFromLocale = _readWriteThisFromLocale;
+  super.init();
+}
+
+pragma "no doc"
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool, f:file, out error:syserr, hints:c_int, start:int(64), end:int(64), in local_style:iostyle) {
```
mppf: I think this one could be private too. Does it work if it's private?

Users should be constructing channels with 'openreader' 'file.writer' etc.

modules/standard/IO.chpl
```diff
@@ -1971,7 +1961,54 @@ proc =(ref ret:channel, x:channel) {
 }
 
 pragma "no doc"
-proc channel.channel(param writing:bool, param kind:iokind, param locking:bool, f:file, out error:syserr, hints:c_int, start:int(64), end:int(64), in local_style:iostyle) {
+proc channel.init(param writing:bool, param kind:iokind, param locking:bool) {
```
mppf: I feel a little surprised that this is the way you write the default initializer for a generic record. Is that something we've talked about in a GitHub issue somewhere? I probably just forgot everything...

lydia-duncan: I thought private couldn't be applied to methods yet, am I misremembering?

modules/standard/IO.chpl
lydia-duncan: I thought private couldn't be applied to methods yet, am I misremembering?

modules/standard/IO.chpl
lydia-duncan: [See here](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/10.rst#type-declarations-of-generic-instantiations), which was a result of discussion from back in May

mppf: ah, thanks. I'm surprised that document doesn't describe default initialization at all - I was looking for a place discussion of default initialization could refer to the section you pointed to.

modules/standard/IO.chpl
mppf: No you're right, it gives a compilation error today. But I think it's only not implemented because we've got the wrong idea about what 'private' should mean. I think it should always mean module-private and we shouldn't bother with C++ or Java style field private. But of course that's another issue.

lydia-duncan: Nah, it's only not implemented because we haven't made an official decision yet :)

modules/standard/IO.chpl
lydia-duncan: Hmm, yeah, there's probably room for a crosslink in there...

PR #8123, opened 2018-01-04
Add a no-op multi-locale performance test

ronawho:
Extend the no-op test so we can track multi-locale startup times too.

We expect https://github.com/chapel-lang/chapel/pull/7700 to improve ugni startup times, so I want to track the before and after timings


ronawho:
@benharsh mind reviewing?

PR #8127, opened 2018-01-05
Convert Random to use initializers

bradcray:
This converts the Random module to use initializers.  In doing this, I reordered the fields
to match the initializer argument order because I initially ran into problems by trying to
initialize fields in argument order rather than field order which the compiler doesn't like
(and I think the field order isn't terribly important and reflects the old / deprecated
constructor argument ordering).

Other than that, this was a fairly trivial translation.

### DIFF
--- a/modules/standard/Random.chpl
+++ b/modules/standard/Random.chpl
@@ -556,7 +556,7 @@ module Random {
 
 
       /*
-        Constructs a new stream of random numbers using the specified seed
+        Creates a new stream of random numbers using the specified seed
         and parallel safety.
 
         :arg eltType: The element type to be generated.
@@ -571,10 +571,14 @@ module Random {
         :type parSafe: `bool`
 
       */
-      proc RandomStream(type eltType,
-                        seed: int(64) = SeedGenerator.currentTime,
-                        param parSafe: bool = true) {
+      proc init(type eltType,
+                seed: int(64) = SeedGenerator.currentTime,
+                param parSafe: bool = true) {
+        this.eltType = eltType;
         this.seed = seed;
+        // TODO: This can't be right:
+        //        this.parSafe = parSafe;
+        super.init();
         for param i in 1..numGenerators(eltType) {
           param inc = pcg_getvalid_inc(i);
           PCGRandomStreamPrivate_rngs[i].srandom(seed:uint(64), inc);
@@ -2032,7 +2036,7 @@ module Random {
 
 
       /*
-        Constructs a new stream of random numbers using the specified seed
+        Creates a new stream of random numbers using the specified seed
         and parallel safety.
 
         .. note::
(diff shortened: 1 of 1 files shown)


bradcray:
@lydia-duncan, @noakesmichael: If either of you have the chance, this case was odd in that if I uncommented the lines in my initializers which set `this.parSafe = parSafe` the compiler was complaining that it was being initialized multiple times even though it doesn't seem to me that it is.  I tried some various manipulations to stop it (e.g., no field initializer, no default value in argument list), but couldn't find something that worked other than commenting the line out completely (which obviously isn't what we want).  I didn't think to try renaming the `parSafe` field to something else to see if that helped avoid some conflict with something else.  I think I'll need some help sorting this out.  I believe that any test that uses the Random module should generate the error (once the offending lines have been commented out).

noakesmichael:
The error message isn’t very elegant but it’s “easy”
to spot the error with un-biased eyes.

The fields of RandomStream are

type eltType;
param parSafe : bool = true;
const seed: int(64);

but your tentative code tried to “init” seed and then parSafe.

I switched the order and the test I was working with passed.

Mike



On Jan 4, 2018, at 5:34 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


@lydia-duncan<https://github.com/lydia-duncan>, @noakesmichael<https://github.com/noakesmichael>: If either of you have the chance, this case was odd in that if I uncommented the lines in my initializers which set this.parSafe = parSafe the compiler was complaining that it was being initialized multiple times even though it doesn't seem to me that it is. I tried some various manipulations to stop it (e.g., no field initializer, no default value in argument list), but couldn't find something that worked other than commenting the line out completely (which obviously isn't what we want). I didn't think to try renaming the parSafe field to something else to see if that helped avoid some conflict with something else. I think I'll need some help sorting this out. I believe that any test that uses the Random module should generate the error (once the offending lines have been commented out).

lydia-duncan:
That was going to be my first guess as well.  We have an action item to improve that particular error message, but it hasn't been a high priority so far

bradcray:
Aha, OK, thanks!  Probably a question for Lydia: How hard would it be to specialize this error message to refer somehow to field ordering rather than simply re-initialization?  Since the first initialization is implicit, it seems misleading or at least less helpful than ideal.

lydia-duncan:
I don't think it would be hard, but it would involve tracking a little more information.  Probably an hour's worth of work at most.

bradcray:
Oops, I hadn't seen Lydia's first response before posting mine for some reason (failed to refresh?).  Should I post an issue/future against the error message request or does one exist?

lydia-duncan:
We have a future for it (test/classes/initializers/phase1/unorderedFields-preferred, I think), but thanks for checking!

bradcray:
Perfect, thanks!

bradcray:
@mppf: As the last significant contributor to the Random module, would you review this?  As keepers of the initializers, @lydia-duncan and @noakesmichael may be interested in a look as well (though they already have).

PR #8128, opened 2018-01-05
Filter out gmake "clock skew detected" warnings from chpl tests and "make check"

awallace-cray:
This is to quiet Chapel test errors caused by gmake warnings like this
in the Chapel test output (the chpl compile output, actually):
    
      gmake: Warning: File `foo/bar.o' has modification time 0.01 s in the future
      gmake: warning:  Clock skew detected.  Your build may be incomplete.
    
These warning msgs typically happen when CHPL_HOME or TMPDIR is located on
a remote-mounted NFS file system. They are nothing to worry about- at least,
not from the chpl compiler. However, they will cause the test output to
differ from the "good"- unless filtered out

### DIFF
--- a/util/test/checkChplInstall
+++ b/util/test/checkChplInstall
@@ -179,7 +179,7 @@ fi
 
     # no errors on stdout/stderr?
     # first, apply "prediff"-like filter to remove gmake "clock skew detected" warnings, if any
-TEST_COMP_OUT=$( grep < ${TMP_TEST_DIR}/chpl.out -Ev '^g?make(|\[[0-9]+\]): *[Ww]arning: *(File .* has modification time .* in the future|Clock skew detected\. *Your build may be incomplete\.) *$' 2>&1 )
+TEST_COMP_OUT=$( grep < ${TMP_TEST_DIR}/chpl.out -Ev '^g?make(\[[0-9]+\])?: *[Ww]arning: *(File .* has modification time .* in the future|Clock skew detected\. *Your build may be incomplete\.) *$' 2>&1 )
 
 if [ -n "${TEST_COMP_OUT}" ]; then
     log_error "Test job compiled with output - Chapel is not installed correctly"
--- a/util/test/checkChplInstall
+++ b/util/test/checkChplInstall
@@ -162,19 +162,26 @@ COMP_FLAGS="--cc-warnings"
 
 # Compile test job into temporary directory
 log_info "Compiling \$CHPL_HOME/${REL_TEST_DIR}/${TEST_JOB}.chpl"
-TEST_COMP_OUT=$( ${CHPL_BIN} ${TEST_DIR}/${TEST_JOB}.chpl ${COMP_FLAGS} -o ${TMP_TEST_DIR}/${TEST_JOB} 2>&1 )
+${CHPL_BIN} ${TEST_DIR}/${TEST_JOB}.chpl ${COMP_FLAGS} -o ${TMP_TEST_DIR}/${TEST_JOB} > ${TMP_TEST_DIR}/chpl.out 2>&1
 COMPILE_STATUS=$?
 
-# Check that compile was successful
+# Check that chpl compile was successful
 log_debug "Compilation exit status was ${COMPILE_STATUS}"
 
+    # exit status 0?
 if [ ${COMPILE_STATUS} -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


awallace-cray:
The test harness won't use the new prediff file unless CHPL_SYSTEM_PREDIFF points to it. 
However, I plan to set that env variable in the Jenkins job configs, not in the individual util/cron/test-*.sh scripts in chapel.git. I think it will work better that way.

awallace-cray:
This has been tested in a Jenkins sandbox, with a Cray module test config and a chapcs multi-locale test config.

ben-albrecht:
Looks good in general. A few clarifying questions:

> The test harness won't use the new prediff file unless CHPL_SYSTEM_PREDIFF points to it.

Will we only be setting this for Cray testing?  Also, is it possible for users running `start_test` on a Cray to encounter this error, since they are unlikely to have set `CHPL_SYSTEM_PREDIFF`?

> However, I plan to set that env variable in the Jenkins job configs, not in the individual util/cron/test-*.sh scripts in chapel.git. I think it will work better that way.

Could you explain why this is preferred over setting the default in the scripts? I don't have a preference, but am curious what you had in mind.

awallace-cray:
> Will we only be setting this for Cray testing? 

For the new prediff file, I tried to select only the Jenkins jobs that run start_test, with either WORKSPACE or TMPDIR located on the remote NFS-mounted file systems that typically generate those gmake clock-skew warnings.  This includes nightly Cray module correctness tests, and the multi-locale (Slurm/Gasnet/Ibv) nightly tests run on the chapcs cluster.  

> is it possible for users running start_test on a Cray to encounter this error, since they are unlikely to have set CHPL_SYSTEM_PREDIFF?

Yes, it is possible for other users running start_test to have these errors.  If that happens, they can set CHPL_SYSTEM_PREDIFF.  There are a number of other systemwide prediffs in the same location which are not currently used in our nightly tests, but could be used if the need arose.  Many of them do contain more "README"-type documentation than this new one does. I will fix that. 

The easiest way to make sure no user anywhere would ever have to find out about using this new prediff file, would be to run the new prediff every time- needed or not. The trouble is, I am trying NOT to run this prediff if it isn't needed.  Two reasons: 

(a) Start_test/sub_test can only handle one systemwide prediff at a time. We already use another systemwide prediff in a few jobs ("incr" test configs, for example), and I expect other new ones will be wanted in the future.  I considered changing start_test to handle multiple systemwide prediffs if needed- it would not be difficult. But that leads to the 2nd reason:

(b) The Chapel dev team's Jenkins currently runs over 200,000 individual Chapel test cases every night. I think we should try to avoid adding any more little bits of processing to each test if it isn't needed. 

In the Jenkins config, it is fairly easy to set the env variable (or not) for each job, config based on context knowledge of our site-specific Jenkins environment, including build slave configs. 

If we set the env variable in the little job script files in util/cron, we would probably need to add some rather site-specific "if" tests- or, just invoke the prediff every time whether needed or not.  If we need to make adjustments in the future, it would mean more changes in chapel.git- adding to the commit noise.  Plus, the little job script files for the Cray module tests are not even stored in chapel.git.

util/test/checkChplInstall
```diff
@@ -162,19 +162,26 @@ COMP_FLAGS="--cc-warnings"
...
-    echo "${TEST_COMP_OUT}"
+    cat ${TMP_TEST_DIR}/chpl.out
     log_debug "Exit \"make check\" script with status code 1"
     exit 1
-elif [ -n "${TEST_COMP_OUT}" ]; then
+fi
+
+    # no errors on stdout/stderr?
+    # first, apply "prediff"-like filter to remove gmake "clock skew detected" warnings, if any
+TEST_COMP_OUT=$( grep < ${TMP_TEST_DIR}/chpl.out -Ev '^g?make(\[[0-9]+\])?: *[Ww]arning: *(File .* has modification time .* in the future|Clock skew detected\. *Your build may be incomplete\.) *$' 2>&1 )
```
ronawho: `-E` is a gnu extension. This fails on omnios and likely other machines. Can we use something more portable?

And could we also shove this inside of the `if [ -n "${TEST_COMP_OUT}" ]; then` check so that we only try to apply the prediff if there was compiler output?

awallace-cray: Okay

PR #8130, opened 2018-01-05
Implements Kronecker Product in LinearAlgebra module as `kron()`.

### LINKED ISSUE #8072: Provide Kronecker product in Linear Algebra package
The [Kronecker Product](https://en.wikipedia.org/wiki/Kronecker_product) is useful in many matrix operations and hence many machine learning routines.  It is relatively simple to implement recursively.

nimitbhardwaj:
Add the procedure `kron` in `LinearAlgebra` module and provide corresponding tests.

Implement `kron` using parallelism of chapel, input two 2-d matrices and output the
resultant matrix which equals the Kronecker Product of the matrices.

Closes #8072


nimitbhardwaj:
I think implementation is correct, working fine for some sample test cases

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,25 @@ proc trace(A: [?D] ?eltType) {
   return trace;
 }
 
+proc kroneckerProduct(A: [?ADom] ?eltType, B: [?BDom] eltType) {
```
ben-albrecht: I'd prefer a less verbose procedure name like `kron()`. It seems that several other linear algebra libraries made a similar decision in their interface.

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,25 @@ proc trace(A: [?D] ?eltType) {
...
+  if ADom.rank != 2 || BDom.rank != 2 then compilerError("Rank size not 2");
+  const (rowA, colA) = A.shape;
+  const (rowB, colB) = B.shape;
+  var C = Matrix(rowA*rowB, colA * colB);
+  for i in 0..rowA-1 {
```
ben-albrecht: Try the count operator for these, e.g.

```chpl

for i in 0..#rowA {
  ...
}
```

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,25 @@ proc trace(A: [?D] ?eltType) {
   return trace;
 }
 
+proc kroneckerProduct(A: [?ADom] ?eltType, B: [?BDom] eltType) {
+  if ADom.rank != 2 || BDom.rank != 2 then compilerError("Rank size not 2");
+  const (rowA, colA) = A.shape;
+  const (rowB, colB) = B.shape;
+  var C = Matrix(rowA*rowB, colA * colB);
```
ben-albrecht: nitpick - try use consistent spacing

ben-albrecht: This should also take an argument `eltType=eltType` to inherit the element type of the matrix arguments.

ben-albrecht:
>  its basic Implementation now, there are a few things like commenting, documentation and tests left now

Ah, I see now that you omitted these intentionally :)

modules/packages/LinearAlgebra.chpl
ben-albrecht: Also, a `forall` seems appropriate here, since each iteration can be computed independently.

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: Ok

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: yes, I was thinking of it too, forall may give better performance

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: ok

nimitbhardwaj: OK, I didn't know of it, thanks

nimitbhardwaj:
@ben-albrecht , thanks for review, i make the changes quickly and also was making test and the doc for it

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,28 @@ proc trace(A: [?D] ?eltType) {
...
+proc kron(A: [?ADom] ?eltType, B: [?BDom] eltType) {
+  if ADom.rank != 2 || BDom.rank != 2 then compilerError("Rank size not 2");
+  const (rowA, colA) = A.shape;
+  const (rowB, colB) = B.shape;
+  var C = Matrix(rowA*rowB, colA*colB, eltType=eltType);
+  forall i in 0..#rowA {
+    forall j in 0..#colA {
+      var stR = i*rowB;
+      var stC = j*colB;
+      forall k in 0..rowB-1 {
```
ben-albrecht: I would change these 2 inner `forall` loops back into `for` loops. The `forall` loops will only spawn tasks based on the available resources, which will be consumed by the outer `forall` loop tasks.

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,28 @@ proc trace(A: [?D] ?eltType) {
...
+/* Return the Kronecker Product of matrix ``A`` and matrix ``B``.
+   If the size of A is ``x * y`` and of B is ``a * b`` then size of the resulting
+   matrix will be ``(x * a) * (y * b)`` */
+proc kron(A: [?ADom] ?eltType, B: [?BDom] eltType) {
+  if ADom.rank != 2 || BDom.rank != 2 then compilerError("Rank size not 2");
+  const (rowA, colA) = A.shape;
+  const (rowB, colB) = B.shape;
+  var C = Matrix(rowA*rowB, colA*colB, eltType=eltType);
+  forall i in 0..#rowA {
+    forall j in 0..#colA {
```
ben-albrecht: I am on the fence about leaving this 2nd `forall` vs. changing it back to `for` loop. It could benefit performance in cases where `rowA < numPUs`. 

Does anyone from @chapel-lang/perf-team have an opinion?

nimitbhardwaj:
I also fix it, but first there is some error in my pc regarding python, so `make docs` was not working, I am first recovering it

ben-albrecht:
> but first there is some error in my pc regarding python

Feel free to post more info on the issue you're having if you'd like help. It might be worth posting to stack overflow, in case others will or already have run into the same issue.

nimitbhardwaj:
yes, its nearly fixed, but only one error now, sphinx error, sphinx-build is not there so im unable to make the documentation

nimitbhardwaj:
it will be done

nimitbhardwaj:
Or I will do this PR tomorrow, as in India now its night, and tomorrow is my class, I best try to fix it now

ben-albrecht:
> sphinx-build is not there so im unable to make the documentation

`make docs` from `$CHPL_HOME` should build the virtualenv with `sphinx` as a prerequisite.

nimitbhardwaj:
yes, i did it, export CHPL_HOME=path, but sphinx error is it is using old version, but i will later figure out it, first I fix the issue

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: OK, I understand it

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: but I want to understand it, the 2nd forall loop and 1st forall loop will do the thing, parallely multiplying an integer with the matrix

nimitbhardwaj: like kron(A, B) is needed, then A[i][j] * B is calculated parallely, if we use 2 forall loops, it will be fast, if we have enough number of threads, which should be equal to the size of matrix A, but now considering the complexity of algo, A would have size atmost 100 * 100, because then it will cause 10^8 ops, this may take 1s of time, so I think it will fast the things,
Where am I incorrect?

nimitbhardwaj: the loop 3 and  4 is understandable as getting huge number of threads cann be a pain in optimization, but can avg 10^4 threads we get, if threads use one one model, then 10^4 LWP

nimitbhardwaj:
I removed the forall loops, sorry, maybe wrong commit message, 3 loops are removed, now, but I would like to understand the reason for the 2nd loop to be removed, I didn't get it much

modules/packages/LinearAlgebra.chpl
```diff
@@ -927,6 +927,28 @@ proc trace(A: [?D] ?eltType) {
   return trace;
 }
 
+/* Return the Kronecker Product of matrix ``A`` and matrix ``B``.
+   If the size of A is ``x * y`` and of B is ``a * b`` then size of the resulting
+   matrix will be ``(x * a) * (y * b)`` */
+proc kron(A: [?ADom] ?eltType, B: [?BDom] eltType) {
+  if ADom.rank != 2 || BDom.rank != 2 then compilerError("Rank size not 2");
+  const (rowA, colA) = A.shape;
```
ben-albrecht: Thinking on this more, this implementation makes the assumption that `A` and `B` have 0-based domains, i.e. `{0..#rowA, 0..colA}`. The `LinearAlgebra` library is intended to be generic w.r.t. domain indices. For this reason I suggest we do a reindexing for `A` and `B` so that we can access them as if they were 0-based within the function. Also, I now notice that we can combine the nested for loops into 2 loops over the 2-dimensional domains, using the `forall` in the outer loop. Here's what I suggest for these changes:

```chpl

  const AzDom = {0..#rowA, 0..#colA},
        BzDom = {0..#rowB, 0..#colB};

  ref Az = A.reindex(AzDom),
      Bz = B.reindex(BzDom);

  var C = Matrix(rowA*rowB, colA*colB, eltType=eltType);

  forall (i,j) in AzDom {
    const stR = i*rowB,
          stC = j*colB;
    for (k,l) in BzDom {
      C[stR+k, stC+l] = Az[i, j]*Bz[k, l];
    }
  }
```

modules/packages/LinearAlgebra.chpl
ben-albrecht: This is a good question. First, I suggest taking a look at the [`forall` section of the users guide](https://chapel-lang.org/docs/latest/users-guide/datapar/forall.html) if you haven't yet.

>  if we use 2 forall loops, it will be fast, if we have enough number of threads,

This is why I was on the fence about this. If there are more elements in `0..#rowA` than available processors, then all the available processors would be saturated with tasks from the outer-most loop, making the inner `forall` effectively a `for` loop, but with a little extra overhead from querying the available resources. The best choice depends on how small that extra overhead is and how often `numProcs > rowA`, which is hard to predict.

Fortunately, we can simplify this decision by making the outer `forall` over a 2-dimensional domain `{0..#rowA, 0..#colA}` instead of 2 nested 1-dimensional loops (see my other comment).

modules/packages/LinearAlgebra.chpl
nimitbhardwaj: ok, this sounds interesting, I change it, thanks

nimitbhardwaj:
Its done, and thanks for coordinating, it was really nice.

ben-albrecht:
@nimitbhardwaj - please update the PR description to follow the merge message format, and we will be good to merge:

[final merge message](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/ContributorInfo.rst#final-merge-message)

nimitbhardwaj:
I didn't find any words more giving more brief explaination

ben-albrecht:
I'm specifically asking you to reword this:

> With ref to #8072, I implemented the Kronecker Product, its basic Implementation now, there are a few things like commenting, documentation and tests left now, but I left them now intentionally, I want to get feedback for it first, like if there any way to optimize it further like by using forall loop or other tricks of parallelism, and now its complexity is O(abxy) where A[a, b] and B[x, y].
Please have a check
>
> Closes #8072

to fit these guidelines:

> - start with a single topic line with at most 75 characters
> - then have a blank line
> - then have a more detailed explanation including motivation for the change and how it changes the previous behavior
> - use present tense (e.g. "Fix file iterator bug")

I would start with:

> Implements KroneckerProduct in LinearAlgebra module as `kron()`.
>
> (some details)
>
> Closes #8072

nimitbhardwaj:
Thanks for guiding

nimitbhardwaj:
Thanks for merge  :smile:  :+1:

ben-albrecht:
Thanks for contributing!

Issue #6010, opened 2017-04-13
Path library - add support for isAbsPath


lydia-duncan:
Back in late 2014, we started undertaking the implementation of the FileSystem and Path libraries, with input from the user community ([start of thread](https://sourceforge.net/p/chapel/mailman/message/32800541/), [mail with updated proposal after feedback](https://sourceforge.net/p/chapel/mailman/message/32902710/)) Much progress has been made in getting the support onto master, but 10 functions still remain unimplemented due to the youth of our string implementation at the time.  With our string implementation in a good state, these functions should be returned to and implemented

This issue is for the function `isAbsPath`.

Here is the expected documentation and declaration:
```chapel
/* Determines whether the path specified is an absolute path and returns
   that information

   :arg name: the path to be checked.
   :type name: `string`

   :return: `true` if `name` is an absolute path, `false` otherwise
   :rtype: `bool`
*/
proc isAbsPath(name: string): bool;
```

This function will probably be used in the implementation of `absPath` (#6005)

lydia-duncan:
Tests will need to be written for this feature

victor-ludorum:
Hello @lydia-duncan  I have worked on this issue and made an **isAbsPath function** .
``
proc isAbsPath(name: string): bool {
  
     if name.isEmptyString() {
	    return false;
	 }
  
     const len: int = name.length ;
  
    /*
	First, there will be a check that
	whether first character of name 
	are '/' or '\'. If it is then it 
	is Absolute .
	*/
	
    var str: string = name[1];
    if (str == '/' || str == '\\') {
      return true;
    }

	/*
	Then , there will be a check that
	whether the name will be like "C:\\"
	or "D:/" or "c:/" . So, we first check for
	alphabets (capital & small) letters . Then, 
	check ':' character and lastly checking
	has been done for  '/' or '\' .
	*/

	else if str.isAlpha() {
      // Possible device root

      if len > 2 && name[2] == ':' {
        var strp: string = name[3];
        if (strp == '/' || strp == '\\') {
          return true;
		}  
      }
    }
    return false;
}
``
Kindly check it and suggest the necessary changes that should be done .

lydia-duncan:
I'm going to focus on other work for the rest of today, but will be sure to review this tomorrow!

victor-ludorum:
Okay Mam No problem 😊 !!

lydia-duncan:
Support was added for this function for linux and OSX, but not for cygwin

[...]

PR #8133, opened 2018-01-05
isAbsPath function issue #6010

victor-ludorum:
Implementing **isAbsPath** function to check absolute path resolves issue #6010 .
@lydia-duncan  Kindly review it and suggest the necessary changes required .


lydia-duncan:
(earlier Travis failures were an unrelated connection issue)

lydia-duncan:
Passed std/ testing.  Getting set up to double check windows

lydia-duncan:
New test passed cygwin.  No failures for the test/library/standard/Path/ test directory

cassella:
Should the Windows-style absolute paths be accepted on e.g. Linux?  For example, on my Linux box, I can 

```
fortytwo@magrathea:~/names$ touch 'C:\AUTOEXEC.BAT'
fortytwo@magrathea:~/names$ touch '\\server'
fortytwo@magrathea:~/names$ ls
C:\AUTOEXEC.BAT  \\server
```

I would expect ``isAbsPath()`` to return ``false`` for those on Linux.

lydia-duncan:
That seems reasonable to me (sorry for not realizing that myself, @victor-ludorum)

victor-ludorum:
@cassella Thanks for your evaluation!! . @lydia-duncan  Yeah !! I think also that it is reasonable . This (**isAbsPath()** function ) and also **normcase()** function have the same issue that it is different for different OS. So, I think that there should be separate files for the different OS as present in _Python_ . It would be helpful if anyone will give any solution to avoid this problem instead of separating files that I have stated above . It would be very much helpful 😊 .

modules/standard/Path.chpl
```diff
@@ -260,5 +260,32 @@ proc file.realPath(): string throws {
...
+*/
+
+  proc isAbsPath(name: string): bool {
+     if (CHPL_TARGET_PLATFORM == 'linux64' || CHPL_TARGET_PLATFORM == 'linux32') {
+        if name.isEmptyString() {
+           return false;
+        }
+        const len: int = name.length;
+        var str: string = name[1];
+        if (str == '/' || str == '\\') {
```
lydia-duncan: If I recall correctly, backslashes are only useful in a Windows environment.

modules/standard/Path.chpl
```diff
@@ -260,5 +260,32 @@ proc file.realPath(): string throws {
...
+
+   :arg name: the path to be checked.
+   :type name: `string`
+
+   :return: `true` if `name` is an absolute path, `false` otherwise
+   :rtype: `bool`
+*/
+
+  proc isAbsPath(name: string): bool {
+     if (CHPL_TARGET_PLATFORM == 'linux64' || CHPL_TARGET_PLATFORM == 'linux32') {
```
lydia-duncan: Mac OS also has a similar style of absolute path, but it is not covered here.

victor-ludorum:
@lydia-duncan Thanks for pointing . I have updated it .

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
```diff
@@ -0,0 +1,15 @@
+   use Path;
```
lydia-duncan: This test will currently generate the error message on cygwin.  You'll need to either make a separate test for that environment (and skip this test when we are on cygwin), or add an alternative .good file for that setting.  Since we would like to support cygwin eventually I recommend making a separate test as a future, with the test cases you removed, their expected output in the .good file, and the error message as your .bad file for that test.  The future should mention that we want to support isAbsPath on Windows

lydia-duncan: This future would also need to limit itself to only run on Windows

victor-ludorum: @lydia-duncan  Can I make one separate file for cygwin testcases and their output so that when it will get implemented then it may get tested on it .

lydia-duncan: Yup!

lydia-duncan: There should be examples of tests that only use cygwin in the test directory, or that exclude cygwin.

victor-ludorum: okay I will make one separate directory for cygwin testcase and testoutput file

victor-ludorum: @lydia-duncan  I have added separate file for the test cases . Kindly suggest other necessary changes that should be done

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
```diff
@@ -0,0 +1,15 @@
...
+
+   writeln(isAbsPath("./baz"));
+	 
+   writeln(isAbsPath("/baz/.."));
+	 
+   writeln(isAbsPath("baz/"));
+	 
+   writeln(isAbsPath("//server"));
+	 
+   writeln(isAbsPath("\\server"));
```
lydia-duncan: Shouldn't this case fail for non-windows environments?

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
lydia-duncan: I'm not seeing a skipif to limit these tests to only certain configurations, as I asked (like [this file](https://github.com/chapel-lang/chapel/blob/master/test/studies/shootout/k-nucleotide/bradc/SKIPIF), except the file name should be testcases_isAbsPath.skipif).

I also asked for the cygwin test to have a .future and a .bad file (see [here](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst#futures-a-mechanism-for-tracking-bugs-feature-requests-etc) for details on what that means).  The new test will cause noise in our nightly runs without these files.

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
victor-ludorum: @lydia-duncan Yeah mam!! This test case fail for non-windows environment . I have thought that we have to add testcases for windows environment in a separate file . Therefore I have added this testcase file as a separate file .

lydia-duncan: Ooooh, I misread the output.  Sorry!

victor-ludorum: No problem mam 😊 !!

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
victor-ludorum: @lydia-duncan I have added **skipif , .future and .bad file** . Kindly review it and suggest any further changes that should be done .

test/library/standard/Path/victor-ludorum/cygwin/testcases_isAbsPath.future
```diff
@@ -1,3 +1,11 @@
+   // Test file for isAbsPath() function for cygwin
```
lydia-duncan: Sorry, the .future file is needed *in addition* to the .chpl file.  Was this just misreading the instructions, or is there a particular part of the description of futures that led you to this conclusion?

victor-ludorum: In the doc which you have suggested , only 3 files are mentioned .future, .good and .bad . Therefore I have added this only.

lydia-duncan: "To mark a test as a future, you add a .future file sharing the same base name as the test. For example, adding a hi.future file would make the simple correctness test at the start of this document into a future test." (second paragraph of the futures section)

This snippet indicates to me that the .future file should be additional, not replacing the .chpl file.

victor-ludorum: Okay!! I think I have not understood it correctly . Thanks !!

victor-ludorum: I have added testcases_isAbsPath.chpl file .

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.skipif
```diff
@@ -0,0 +1 @@
+CHPL_HOST_PLATFORM == linux32 , CHPL_HOST_PLATFORM == linux64 and CHPL_HOST_PLATFORM == darwin
```
cassella: I don't know the details of the skipif file syntax, but I can't make sense of this one.

victor-ludorum: @cassella  Actually I also don't know about the syntax . I have implemented it by considering linux32, linux64 and darwin environment . This is what I want to specify . Suggest me How can I write this ?? 
Thanks 😊 !!

lydia-duncan: You'll be able to find that information by searching through the .skipifs used in the test directory.  I suggest looking up how to do a search to find all files with a specific suffix, and then looking through those for examples of chaining conditions

victor-ludorum: Thanks @lydia-duncan  for your advice !! I have updated the **skipif** file . Kindly check it and suggest if any further modification is needed .

modules/standard/Path.chpl
```diff
@@ -260,5 +260,32 @@ proc file.realPath(): string throws {
...
+
+   :arg name: the path to be checked.
+   :type name: `string`
+
+   :return: `true` if `name` is an absolute path, `false` otherwise
+   :rtype: `bool`
+*/
+
+  proc isAbsPath(name: string): bool {
+     if (CHPL_TARGET_PLATFORM == 'linux64' || CHPL_TARGET_PLATFORM == 'linux32' || CHPL_TARGET_PLATFORM == 'darwin') {
```
lydia-duncan: This line is likely longer than 80 characters, I would recommend putting each part of the conditional on a separate line:
```chapel
if (CHPL_TARGET_PLATFORM == 'linux64' ||
    CHPL_TARGET_PLATFORM == 'linux32' ||
    CHPL_TARGET_PLATFORM == 'darwin') {
```

modules/standard/Path.chpl
```diff
@@ -260,5 +260,32 @@ proc file.realPath(): string throws {
...
+        const len: int = name.length;
+        var str: string = name[1];
+        if (str == '/') {
+          return true;
+        }
+        else 
+          return false;
+     }
+     else
+     writeln("Error : Target platform should have Unix like environment");
```
lydia-duncan: The writeln should be indented under the else, especially since there are no curly braces to make the dependence obvious.

lydia-duncan: Actually, the more I think about it, the more I wonder if throwing an Error subclass would be useful here (see [this document](https://chapel-lang.org/docs/latest/technotes/errorHandling.html) for general error handling in Chapel)

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.skipif
```diff
@@ -0,0 +1,5 @@
+# It is supportable in 'Unix' and
```
lydia-duncan: This will do the opposite of what you intend, sadly.  It will skip this test on darwin, linux32 and linux64, but those are the operating systems where it will work.

You want to use this file's current contents for the cygwin test's .skipif.
After that, you'll want to change the skipif for this test to check against cygwin32 and cygwin64.

Have you tried running `start_test $CHPL_HOME/test/library/standard/Path/victor-ludorum/` on your machine?

modules/standard/Path.chpl
lydia-duncan: @psahabu - is there a handy list somewhere of all the Error subtypes, that @victor-ludorum could use?

psahabu: We've created a SystemError subtype hierarchy, which can be found here: https://chapel-lang.org/docs/1.16/modules/standard/SysError.html?highlight=syserror#SysError.SystemError

We're open to creating other subtype hierarchies, though it's not an area of focus right now.

victor-ludorum:
Thanks @lydia-duncan for the review .
I have updated the three things suggested by you .
**1. .skpif file
2. Error message 
3. maintained 80 characters or less in a single line**
Kindly suggest if any further modification is required .

test/library/standard/Path/victor-ludorum/cygwin/testcases_isAbsPath.chpl
```diff
@@ -0,0 +1,19 @@
+   use Path;
```
lydia-duncan: I'm not seeing the skipif so that this test only runs on cygwin.  There should be two skipif files: 1) the one for the Linux and Mac environment (which you have), and one for this test that lives in the same directory

modules/standard/Path.chpl
lydia-duncan: Would making another subtype of SystemError be reasonable for this error?  The error is basically "we don't support this in your environment yet", and it doesn't look like anything is already present that would work for that.

lydia-duncan:
Awesome, thanks!  I'm just waiting on @psahabu's guidance for the Error class.  After that, I'll make sure testing runs fine and get back to you if it doesn't.

psahabu:
A new subtype of `SystemError` to describe a bad environment seems reasonable, although I'd like @mppf to weigh in. Perhaps `BadEnvironmentError`?
```
class BadEnvironmentError : SystemError { ... }
```

mppf:
@psahabu - I missed the relevant discussion for why we'd want to have `BadEnvironmentError` or what it would mean?

victor-ludorum:
@mppf It actually signifies that this function is mainly implemented to work on some environment (Unix and darwin) and on cygwin **it will give error message**

modules/standard/Path.chpl
psahabu: @mppf here's the discussion

mppf:
I don't think we should add `BadEnvironmentError` or a corresponding subclass with a different name. I think we should just use SystemError with EINVAL and a reasonable string error message.

mppf:
I think the intent of the class hierarchy is that there are subtypes for common error types. I don't think this will be a common error type ( or at least not commonly one that needs to be handled).

victor-ludorum:
@mppf  Thanks sir !! I have one question as you have suggested that " **a reasonable error message** " is good to handle this issue . Then , Is my implementation i.e. using compilerError() and printing the message is good or I should change it . As by using this type I can easily print a reasonable message .

mppf:
compilerError is better if it works for the use case

lydia-duncan:
@mppf - we might want this bad environment error for other Path functions, like normCase if we don't start by supporting all systems.

The use case I was thinking of that would motivate an Error subclass over a compiler error is this: say the user wrote a program that they intended to be executable on multiple environments, and wanted to use isAbsPath for all environments it supported and their own homegrown implementation for environments that weren't.  They could use isAbsPath throwing an Error as a way of distinguishing between the cases, rather than having to write their own logic on what should happen in each environment and having to update that logic when we expand our support within this function.

I can see the above example not being our intended use of errors, though.  Thoughts?

mppf:
I'd probably expect such a use case of isAbsPath to just call a home-grown function that sometimes calls the home-grown equivalent and sometimes calls isAbsPath. That can work even with the compiler error.

For the other path functions, I think they can continue to be compilerError if that big hammer makes sense, but at any point they can throw new SystemError(ENOSYS, "This function not implemented on Windows"); e.g. I don't think there's justification for a new subclass of SystemError for these.

lydia-duncan:
Okay, sounds good!  @victor-ludorum - I'll test against the branch as it stands now

lydia-duncan:
The new tests worked fine on my linux testing box.  Running full testing there now.

lydia-duncan:
On the cygwin testing machine, I get this output instead of the expected output (because the compiler error is only executing once, and because it tells you the location of the error):
```
testcases_isAbsPath.chpl:3: error: Target platform should have Unix like environment
```
You will want to update the test/library/standard/Path/victor-ludorum/cygwin/testcases_isAbsPath.bad for this output.

lydia-duncan:
Linux full std/ testing passed, and cygwin test/library/standard/Path & test/library/standard/FileSystem passed aside from the .bad file mismatched I mentioned earlier

victor-ludorum:
Thanks @lydia-duncan  I have updated **.bad** file . kindly review it.

victor-ludorum:
Thanks @lydia-duncan  I have updated the file . please check it.

lydia-duncan:
Perfect, thanks!

PR #8135, opened 2018-01-05
Make the the `tzinfo` time zone field in DateTime.time use Shared

daviditen:
Make the `tzinfo` field in DateTime.time Shared so that it will be
automatically deleted when there are no more references to it.  Updated
functions/methods that have a `TZInfo` argument to use `Shared(TZInfo)`
instead.  Added calls to `borrow()` as needed, most commonly when comparing
it with `nil`.

Updated DateTime tests to use `Shared(TZInfo)` instead of `TZInfo` directly.
One pain point is that real uses of time zones use a subclass of `TZInfo`
and need to cast it to the `TZInfo` type before making it shared.

  var sharedTZ = new Shared(new DerivedTZ(...): TZInfo);


daviditen:
This has passed a full linux64 paratest.

It is not currently backward compatible for codes that use timezones in the `DateTime` module.  Does it need to be deprecated for a release?

modules/standard/DateTime.chpl
```diff
@@ -577,7 +584,7 @@ module DateTime {
       ret = ret + "." + makeNDigits(6, microsecond);
     }
     var offset = utcoffset();
-    if tzinfo != nil {
+    if tzinfo.borrow() != nil {
```
mppf: Do you think it'd be worthwhile having a `proc isEmpty` or something on Shared?

modules/standard/DateTime.chpl
```diff
@@ -850,7 +857,10 @@ module DateTime {
    */
   proc datetime.datetime(year, month, day,
                 hour=0, minute=0, second=0, microsecond=0,
-                tzinfo: TZInfo=nil) {
+                tzinfo: Shared(TZInfo)=new Shared(nil: TZInfo)) {
+    // For some reason, the compiler fails if we use nilTZ for the
+    // tzinfo argument above.  Almost everywhere else it works fine.
+    // Testcase: test/library/standard/DateTime/testTimezone.chpl
```
mppf: Is there an issue about this?

modules/standard/DateTime.chpl
mppf: Actually it might be even slicker if we made support for `==` and `!=` on Owned/Shared combined with a nil type.

modules/standard/DateTime.chpl
```diff
@@ -474,7 +478,7 @@ module DateTime {
     pragma "no doc"
     var chpl_hour, chpl_minute, chpl_second, chpl_microsecond: int;
     pragma "no doc"
-    var chpl_tzinfo: TZInfo;
+    var chpl_tzinfo: Shared(TZInfo);
```
mppf: Can you point me to wherever this is set to something other than Shared(nil)?

daviditen: It's used in some of the DateTime tests: [testDatetimeTZ](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testDatetimeTZ.chpl), [testTimezone](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testTimezone.chpl), [testTimezoneConversions](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testTimezoneConversions.chpl)

modules/standard/DateTime.chpl
daviditen: Yes, `==` and `!=` for Owned/Shared with `nil` would be nice.

modules/standard/DateTime.chpl
daviditen: Not yet.  I'll create one.

daviditen:
@mppf Thank you for reviewing!

mppf:
@daviditen - could you make issue / issues about the Owned/Shared improvements that experience (such as this effort) has identified would be worthwhile?

daviditen:
Yes, I'll make issues for them, thanks.

PR #8141, opened 2018-01-06
Convert DateTime modules from constructors to initializers

bradcray:
Converting the DateTime module from constructors to initializers resulted in a few
surprises, noted here:

* Because the code contains some variable declarations of the form `var d: date;`
  or `var dt: datetime`, I had to add zero-argument initializers to those records to
  keep the code working.  This surprised me a little and may be another case of
  initializers in type contexts not yet working as we would hope, though I'd have to
  review notes or call on @lydia-duncan / @noakesmichael / @mppf to help me be
  sure.  @daviditen should also check that the bodies of these are reasonable (I
  left them empty)—tests didn't fail, but I worry a little that they may result in
  internally inconsistent / illegal values.

* One of the cases that @daviditen called out in #8166 could be written the way
  he had hoped now that we're using initializers, but the other still resulted in the
  same problem.  I converted the one that could be, but the inconsistency made
  me wonder whether we're just getting lucky in some way (e.g., our tests don't
  exercise the problematic code path for that type).  Or maybe the issue is related
  to nested or inherited initializer invocations?

* The `testTimezoneConversions.chpl` test was intriguing in that it required a
  few nontrivial changes.  One of these was the need to create an explicit
  initializer in the child class where we hadn't needed to in the constructor world
  (and, intuitively, it seems like we shouldn't need to from a productivity
  perspective).  I filed a feature request future against this in defaultForInheritedInit.chpl.
  This is low priority for me, but feels like it'd be a nice convenience (and would
  permit an uninteresting initializer added to testTimezoneConversions.chpl to
  be removed.

* In that same area, I needed to replace an explicit `new timedelta(hours=-5)`
  call in a constructor invocation with simply `[hours=]-5` to get it to resolve
  with initializers.  The initializer code seems correct, but I'm at a slight loss as
  to why the constructor version worked.  Maybe we were calling into the
  compiler-provided constructor which is not visible in the initializer world since
  we overrode it?

* Ben Harshbarger's elegant CoMD code stopped working in the initializer world
  because his local `use`s of "DateTime" caused the copy initializer not to be
  visible in some context that wanted it (compiler-generated code, as far as I can
  tell).  As a workaround, I moved this to the module scope, which permitted it to
  work again and filed a new future "localUseNoCopy.chpl" against the failing case
  for further investigation by @lydia-duncan and/or @mppf.

* I also filed a future against a bug that I inadvertently stumbled into in which I
  accidentally called a superclass initializer and got surprising results: 
  inheritedInitWrongName.chpl.  This bug isn't holding me up at all, but feels
  like it might be low-hanging fruit and may head off confusion for future users
  who hit it.

TODO:
- [ ] open issues for the futures added here.


bradcray:
@mppf / @lydia-duncan : This PR has an interesting failure case that I haven't seen before: it causes a compilation error in `studies/comd/elegant/arrayOfStructs/CoMD.chpl` of the form "error: copy-initialization invoked for a type that does not have a copy initializer".  The cause is that the uses of the `DateTime` module in `test/studies/comd/elegant/arrayOfStructs/util/Util.chpl` are local to the routine in question.  If I move the `use`s into the module scope, things work again.

I'm having trouble creating a small reproducer of this for some reason, though...  maybe I just haven't made my callchain sufficiently deep?  Anyway, before trying harder, I thought I'd see whether you'd understand the root cause well enough to make creating a reproducer easier.

lydia-duncan:
I think it's because you are returning the type, so a copy initializer is invoked to save the value.  That would imply that the copy is inserted after the function has exited, though, which seems wrong . . .

mppf:
I don't have any insight about it & think a small reproducer would help.
Which function is trying to get the copy initializer? Is it part of the old FLAG_INSERT_AUTO_COPY path that I'm in the process of removing?

bradcray:
If you're willing to grab this PR, a small reproducer is:

```chapel
proc dateString() {
  use DateTime;
  return datetime.now().strftime("%F, %T");
}

writeln(dateString());
```

I was trying to create one that didn't rely on the DateTime module, but that's what I was unsuccessful at (so far).

bradcray:
@daviditen: I think this is ready to review and that you're the most obvious person to do it as the DateTime author.

@lydia-duncan / @mppf: You'll see that I worked around my copy initializer scoping challenge by filing a future against it and applying a workaround to the original test.  I'll also file an issue against this.

@lydia-duncan / @noakesmichael : May also be interested in the other two futures added here, one of which is a bug relating to inherited initializers.  I'll file issues against these as well.

test/classes/initializers/inheritance/inheritedInitWrongName.good
```diff
@@ -0,0 +1,4 @@
+inheritedInitWrongName.chpl:20: error: unresolved call '_new(type D, x=3, "hi", 4.6)'
+inheritedInitWrongName.chpl:1: note: candidates are: _new(type chpl_t, x, y, z)
+inheritedInitWrongName.chpl:14: note:                 _new(type chpl_t, a, b, c)
+deleted module lines
```
mppf: I don't think there should be an ambiguity error here. The source code clearly says it's creating a `D` : `new D ( ... )`. So it should be calling a `D` initializer *no matter what the arguments are*.

test/classes/initializers/inheritance/inheritedInitWrongName.future
```diff
@@ -0,0 +1,6 @@
+error message: accidental call to superclass initializer
+
+It seems to me that calls to 'new subclass(...)' (like 'new D' in this
```
mppf: I created issue #8176 with the intention that it would track this case - you can refer to it here if you like.

test/classes/initializers/inheritance/inheritedInitWrongName.good
daviditen: The `new D(...)` call uses named argument passing for `x`, but the `D` initializer only has formals `a, b, c`.  I think `unresolved call` makes sense in that case.

mppf: Quite right, quite right. I stand corrected.

test/classes/initializers/inheritance/inheritedInitWrongName.prediff
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+output=$2
+cat $output | sed '/^\$CHPL_HOME\/modules\// d' > $output.tmp
```
daviditen: I'd personally write this line as: `grep -v '\$CHPL_HOME/modules/' $output > $output.tmp`.

test/classes/initializers/inheritance/defaultForInheritedInit.future
```diff
@@ -0,0 +1,7 @@
+feature request: default initializers for single-init inheritance cases
+
+WHen converting some code from constructors to initializers, it seemed
```
lydia-duncan: Capitalization error here ("WHen")

test/classes/initializers/inheritance/defaultForInheritedInit.future
```diff
@@ -0,0 +1,7 @@
...
+natural to me in a case like this that the subclass should get a
+default initializer that was essentially a pass-through to the parent
+class initializer, yet that didn't happen.  From a productivity
+perspective, I think it should.
```
lydia-duncan: I support this future, thanks!

test/classes/initializers/inheritance/inheritedInitWrongName.good
lydia-duncan: I think that means that the test file should fail no matter what, yes?

lydia-duncan:
No issues for me, aside from the question about the future David and Michael were discussing

test/classes/initializers/inheritance/inheritedInitWrongName.good
mppf: Right, it shouldn't compile.

bradcray: As our compiler gets better at printing out the actual candidates, I think this .good file output will change (improve), but this was my best guess as to what it should be in the immediate-ish term (if this issue were fixed in the next week or two, say).

test/classes/initializers/inheritance/inheritedInitWrongName.prediff
bradcray: I copied this file from a number of other identical existing prediffs on initializer testse and while I'm a big fan of never using sed, would prefer to keep them consistent rather than have one differ (or rewrite all of them in this PR).

bradcray: (I'm interpreting this comment as feedback rather than something you thought I needed to necessarily do, so am going to proceed, but we can tackle it post-merge if I'm wrong).

PR #8142, opened 2018-01-06
Ensure fnReturnsGeneric.chpl is compiled with --verify

vasslitvinov:
#8129 changed .bad, which I attributed to #8125.
Looking at it again, the different in compiler output is instead
due to my compiling with --verify. Cf. nightly testing runs without
--verify, so the test output was unchanged from yesterday.

Having the compiler fail due to --verify is preferable because
it shows a problem earlier in during compilation.
So instead of reverting #8129, I am simply forcing the test
to be compiled with --verify.


lydia-duncan:
I would prefer you update the .future file to note the behavior change when --verify is thrown.  This is a bug that is encountered in user code and while it is useful to know that it fails earlier, I think it is more useful to track the behavior the user will encounter by default.

PR #8145, opened 2018-01-08
Improve extern shadow test

mppf:
This future was not failing in the manner it used to.  This commit adds
to the test to make it clearer what is going on and make it no longer a
future.


mppf:
It's working by design that a Chapel function is called in preference to an available function in an extern proc. The original future was seeing compilation failures from the backend compiler in this scenario.

PR #8138, opened 2018-01-05
Update Barriers module to use initializers

bradcray:
This converts the Barriers module over from constructors to initializers.
This was fairly straightforward.  Most interesting part was probably
changing the chpl__initCopy() into a copy initializer.

PR #8149, opened 2018-01-08
Improve handling of initializers in 'inferConstRefs'

benharsh:
This commit updates the 'inferConstRefs' mini-pass to identify
initializers as a single 'def' of a variable, similar to the
pre-existing handling of retArgs. This allows more things to be inferred
to be const, which helps with RVF.

Testing:
- [x] local
- [x] no-local
- [x] gasnet


benharsh:
@bradcray : this resolves the comm count regression resulting from #8138

bradcray:
Thank you!!

PR #8148, opened 2018-01-08
Fix prediff-for-gmake-clock-skew-warning causes false errors in mandelbrot* tests

awallace-cray:
The grep filter in prediff-for-gmake-clock-skew-warning does not work with
"binary" test output files, like mandelbrot*, causing false test errors to
be reported.
    
This change checks if the test output is "binary" before filtering, and if
so, the prediff just skips the filtering and returns normally without
doing anything.

This change symlinks the prediff-for-gmake-clock-skew-warning file to appear
as new files mandelbrot.preexec and mandelbrot-fast.preexec. That way, the
clock-skew-warning filter will run BEFORE the mandelbrot tests add their
"binary" output to the test output file.

Note: other chpl tests create binary output, but these mandelbrot tests are 
the only ones under release/examples.  We could create symlinks to "preexec" for
those tests as well, but they wouldn't make any difference until someone ran the
full test suite AND set CHPL_SYSTEM_PREDIFF=prediff-for-gmake-clock-skew-warning. 

The worst that could happen from NOT setting all the other possible preexec symlinks, is that these tests might throw some false errors someday. 
```
./test/exercises/c-ray/c-ray.4samples.chpl
./test/exercises/c-ray/c-ray.chpl
./test/exercises/c-ray/c-ray.sphfract.chpl
./test/exercises/c-ray/forStudents/c-ray.chpl
./test/exercises/Mandelbrot/mandelbrot-bmptest.chpl
./test/io/ferguson/writefbinary.chpl
./test/release/examples/benchmarks/shootout/mandelbrot-fast-alt.chpl
./test/release/examples/benchmarks/shootout/mandelbrot-fast-clbg.chpl
./test/release/examples/benchmarks/shootout/mandelbrot.chpl
./test/studies/shootout/mandelbrot/bharshbarg/mandelbrot-unrolled.chpl
./test/studies/shootout/mandelbrot/bradc/mandelbrot-blc.chpl
./test/studies/shootout/mandelbrot/ferguson/mandelbrot-opt2.chpl
./test/studies/shootout/mandelbrot/ferguson/mandelbrot-stdout-putchar.chpl
./test/studies/shootout/mandelbrot/ferguson/mandelbrot-stdout.chpl
./test/studies/shootout/mandelbrot/ferguson/mandelbrot-tricks.chpl
./test/studies/shootout/mandelbrot/ferguson/mandelbrot2-opt.chpl
./test/studies/shootout/mandelbrot/jacobnelson/mandelbrot-complex.chpl
[...]

PR #8151, opened 2018-01-09
gmake clock skew warning filter is a global preexec, not a global prediff

awallace-cray:
Replaces PR #8148. 
This simply uses the gmake-clock-skew-warning filter as a preexec every time, without trying to make exceptions for binary test data. In PR #8128 and #8148, it was trying to be used as a prediff, where it caused trouble with binary test output data (mandelbrot) and any other test outputs that lack a trailing newline. 
Best of all, the CHPL_SYSTEM_PREEXEC handler was already implemented in start_test/sub_test.

### DIFF
--- a/util/test/preexec-for-gmake-clock-skew-warning
+++ b/util/test/preexec-for-gmake-clock-skew-warning
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# This script is for the use with 'start_test' as the system-wide prediff, when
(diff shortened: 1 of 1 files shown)


awallace-cray:
Corresponding PR in Jenkins config is ready. 
Tests (including both PRs) passed. 
@ronawho, is that an approval to merge? Thanks!

PR #8153, opened 2018-01-09
Convert MatrixMarket package from constructors to initializers

bradcray:
This was a completely straightforward case of making constructors into initializers.


bradcray:
@lydia-duncan : Here's an easy one for you.

ben-albrecht:
Should this change be propagated [upstream](https://github.com/ct-clmsn/MatrixMarket)?

bradcray:
@ben-albrecht : I think it should, but was holding off on suggesting it until resolving the current open initializer questions (for fear that we'll need to change it a second time).

Issue #8114, opened 2018-01-03
Provide version selection menu on Chapel Documentation page


buddha314:
When doing a Google search for [Chapel Docs](https://chapel-lang.org/docs/1.12/index.html), you typically get older versions.  V. 12 seems popular.  Postgres has provided a version selection menu at the top of their [documentation](https://www.postgresql.org/docs/9.5/static/index.html) that allows you to get to the current version independent of what your search engine thinks you should be reading.

bradcray:
We've long discussed making the version number in the upper-left of the Chapel docs pages be a pull-down menu allowing one to select the version, but nobody's ever taken that on (I'm not sure what the level of effort would be.

I wondered whether v1.12's popularity with Google might be because we stopped embedding Google-specific metadata into the html files with that release (out of laziness / forgetfulness), but it looks like we included them for 1.13 as well.  So maybe it's just when Google and users first started discovering them heavily?  This makes me wonder whether there's metadata that can be provided to Google to say "prefer this version" of these files that are very similar from release to release...

buddha314:
Here is a [Stack Overflow](https://stackoverflow.com/questions/28849284/how-do-i-have-multiple-versions-of-my-doc-with-sphinx) about multiple versions on Sphinx.

FWIW, DuckDuckGo returns a hodgepodge of 1.12, 1.13, 1.15 and 1.16.

bradcray:
@ben-albrecht : I feel like you know our use of the sphinx stuff best.  Any chance you'd be able to look into the Sphinx extension mentioned on the SO page @buddha314 references to see what level of effort would be required for us to make use of it?

ben-albrecht:
I believe the challenge with the `sphinxcontrib-versioning` extension is that it requires rebuilding all of the existing documentation with the extension, which would be difficult, since there were many post-release documentation edits throughout the releases. 

That said, I think it'd be a good idea to start using this extension going forward. We can probably hack together a solution for previous versions if we think it'd be useful enough.

bradcray:
> which would be difficult, since there were many post-release documentation edits throughout the releases

It seems in any reasonable solution, the "menu" providing the list of versions would be stored in a file independent from the source html files themselves so that old versions of docs wouldn't need to be updated to be able to refer to the newer versions (?), is that right?  If so, it may be a simple matter of finding the delta in html between our current docs and the ones that "`#include`" the menu code and running a script over the static html to update it in that respect (?).

[...]

PR #8155, opened 2018-01-09
Version selection for online docs

benharsh:
This PR adds a template override for the sphinx RTD theme's "layout.html" template. A php-include is inserted between the documentation title ("Chapel Documentation") and the search box. The included file represents a drop-down linking to the same page in different versions of Chapel docs. This included file copied into the root of the output HTML directory during 'make docs'. We will likely end up using a symlink to a master copy when pushing to chapel-lang.org.

Testing:
- [x] OS X
  - [x] chrome
  - [x] firefox
  - [x] safari
- [x] Windows
  - [x] firefox


benharsh:
Pinging @ben-albrecht and @bradcray .

ben-albrecht:
Closes #8114 ?

ben-albrecht:
What is the behavior for this feature when building the docs locally?

doc/util/versionButton.php
```diff
@@ -0,0 +1,142 @@
...
+    dropDiv.style.display = "block";
+  } else {
+    dropDiv.style.display = "none";
+  }
+}
+
+var chplTitle = "<?php echo "$chplTitle";?>";
+
+// Note: assumes second element is most-recent release
+var chplVersions = [
```
ben-albrecht: This is something we'll need to update every release. Do you think it's possible to generate it automatically instead?

ben-albrecht:
A few minor design comments:

- It might fit in better if you use colors from the read-the-docs theme

- It's unfortunate that the drop-down menu is so wide, since `pre-release` is included in the next version. One idea is to drop `pre-release` from the menu (having it as static text outside the menu). Another is to adopt a less verbose form of `pre-release`, e.g. `1.17rc`, `1.17 (dev)`, or `1.17 (2532b28)`

benharsh:
When building the docs locally, we will still generate the PHP include. But unless the user has set up a server (that also parses HTML as PHP), they won't see the button.

I looked at using RTD theme colors, but I ended up feeling that they wouldn't be appropriate. The orange-ish color used for warnings is, in my opinion, too desaturated and not 'bold' enough to effectively catch the reader's eye. I didn't want to use the same blue-ish color as the header for the 'current' version because I felt that the button should be distinguishable from the background. I did not see a yellow-ish color to consider for the pre-release.

Are you talking about the width of the button or the width of the thing the button makes visible? I don't think we have much room on the button. I toyed around with it just now, and while "1.17rc" might work, the others would still be on a new line. Also, I don't think "1.17rc" would be accurate until the end of the release cycle.

doc/util/versionButton.php
benharsh: I think point releases make this difficult to automate, but I'm open to hearing ideas. I just did this manually so that we can make progress.

bradcray: I agree that desire for auto-updating shouldn't hold up getting a first draft in.  I also don't think doing this update is that much worse than other documentation and version number-related updates that we already require and could be added to the list.

bradcray:
I don't have a problem with the pre-release version of the button dropping to a second line, and might even consider it appropriate since it visually emphasizes that you're not on the main version.  If anything, I might be tempted to make the older versions also include some longer label like "(outdated)" or "(old)" in order to make them also stand out more (more orange color).  While I think it's a little weird to have the button jump around in that header, I also think that it's appropriate that the one configuration where it would "look right / look good" is the current released documentation (this also relates to why I would support making the label longer for out-of-date versions).

ben-albrecht:
Side note: @benharsh - What do you use to test this locally? I used `php -S 0.0.0.0:8000` and renamed `index.html` to `index.php`. Is there an easier way that does not require renaming `index.html`?

bradcray:
Interesting idea...

benharsh:
I believe this is possible, and I'll start looking into it.

PR #8158, opened 2018-01-09
Skip this test on 32-bit platforms

mppf:
The image output is slightly different on a 32-bit platform. I believe this is due to the fewer number of iterations used.


bradcray:
Note that the CLBG still measures 32-bit results as well:  http://benchmarksgame.alioth.debian.org/u32/mandelbrot.html

I'd be curious whether the programs you modeled your changes after get bad output as well (they show up at the end of the list on the page above, if they haven't been swept under the rug completely).

bradcray:
@mppf: Before I forget, I'm still curious about this.  Poking around myself, I'm guessing for your description that your version was based on gcc #6 which does seem to get correct results in the CLBG 32-bit results.  But I don't feel confident that I'm looking at the right version, nor that there weren't other changes that weren't captured in that version.

PR #8159, opened 2018-01-10
Convert LinearAlgebraJama to initializers

bradcray:
This converts the LinearAlgebraJama package to use initializers.
It required slightly more work than average because
the old constructor-based scheme initialized fields
in a fairly arbitrary order, so the main effort here was
reordering things and teasing phase 1 away from phase 2.

Most of the constructors assigned m and n (the matrix sizes)
first, so I moved those fields higher up in the object
structures rather than trying to create workarounds to
initialize them later (this also matched a few classes that
did declare them first).  As a result of this reordering,
the default class print output changed, requiring an update
to the .good file, but this still seemed better than trying to
delay initialization of m and n.

Where here, and converting things to field initialization, I
also could not resist converting some rather laborious
explicit loop-based serial array initializations to a more
concise, parallel whole-array assignment approach.  This
made it easier / more succinct to initialize the arrays in
phase 1.

### DIFF
--- a/modules/packages/LinearAlgebraJama.chpl
+++ b/modules/packages/LinearAlgebraJama.chpl
@@ -165,14 +165,14 @@ proc hypot(a, b:real) {
 
 class CholeskyDecomposition {
 
+   /* Row and column dimension (square matrix). */
+   var n:int;
+
    pragma "no doc"
    var lDom = {0..1,0..1};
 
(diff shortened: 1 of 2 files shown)


bradcray:
@lydia-duncan: Do you want to take this one?  Pinging @ct-clmsn to let him know that this effort is underway as well (though it may need to evolve further before we're done).

PR #8162, opened 2018-01-10
Avoid task yielding during dynamic memory allocation/registration under ugni

ronawho:
This fixes the oversubscribed array allocation test, and will be important for
some upcoming fully dynamic registration work. We could have fixed the
oversubscribed array allocation test by using a chpl_sync_var as a lock or by
using a try-lock pthread_mutex, but this is the "better" long term solution.
For full dynamic memory registration we need to ensure that an allocation
finishes before a task yields because some allocations in qthreads (and the
chunk hooks in jemalloc) can't handle concurrent/interrupted allocations on the
same pthread.

This hurts ISx performance. On 16 node this takes us from ~8s to ~11s. Note
that that perf loss was on 28-core BW nodes. On 36-core BW nodes, we (somewhat
confusingly) did not see any perf loss even at higher node counts.


ronawho:
Perf testings shows only the isx regressions

Correctness testing with 50 trials also passed: `$CHPL_HOME/util/test/paratest.server -env CHPL_TEST_NUM_TRIALS=50 -nodepara 8 -dirs release/examples/ -dirs runtime/gbt/comm/`

PR #8163, opened 2018-01-10
Add 6 tests to track the bug Russel encountered earlier with bigint

lydia-duncan:
These cover modifications that do and do not reveal the bug found.

reductionOverCreatedArray-zero.chpl is a simplified version of the bug, showing
that the problem occurs when the index scope of the array created is 1..0

reductionOverCreatedArray-arrayTemp.chpl is the simpliest modification to allow
the program to work correctly.  It saves the created array into a temporary
before using it in the reduction.

reductionOverCreatedArray-generalType.chpl is my attempt to mimic bigint so that
I could get a more general case.  It does not encounter the bug, indicating that
one of the additional qualities of bigint is triggering this issue.

reductionOverCreatedArray-reductionTemp.chpl, and
reductionOverCreatedArray-reductionTemp2.chpl are variations involving saving
the result of the reduction operation with the array creation prior to returning
it.  reductionOverCreatedArray-reductionTemp2 demonstrates the bug for a
single statement, while reductionOverCreatedArray-reductionTemp shows the
loop case (I found a situation where this made a difference in output, but the
valgrind bug remained).

Save these cases.  The issue seems a lot more bigint-specific, so I am going to
wait to resolve it for now


lydia-duncan:
Hmm.  While the futures fail on my mac, they don't seem to fail in my testing directory with qthreads.  With fifo, reductionOverCreatedArray-reductionTemp is the only one that fails (and it outputs everything and then segfaults).

Things to check that might impact the results:
- [ ] CHPL_DEVELOPER=false/true
- [ ] linux versus mac.
- [ ] valgrind

lydia-duncan:
Valgrind results: "conditional jump or move depends on uninitialized value" for the futures, and the developer mode check.  The loop example gives this error for all iterations, looks like.

lydia-duncan:
Setting CHPL_DEVELOPER=true on the linux testing environment caused the individual reductionTemp cases to fail but the loop version was fine, and the version that returned the reduction immediately also failed.  I'm going to just make these valgrind tests.

lydia-duncan:
And remove the "--devel" version

PR #8165, opened 2018-01-10
Fix high precision timer launcher issue

mppf:
### util/test/timers/highPrecisionTimer

Output time information to stdout not stderr

Otherwise we see sporadic failures with
  studies/shootout/fasta/kbrady/fasta-printf
in a configuration with a launcher.

Adding a sys.stdout.flush() beforehand didn't solve the problem either.
I suspect that there's a launcher or other multi-node software in the way
of the flush strategy working.

I think it's reasonable to always output the time on stdout because if a
perf test is outputting anything to stderr it's probably failing.

### util/test/computePerfStats

Make it clearer in the failure message that output is truncated 

### Review & Testing:

Reviewed by @daviditen and @ronawho - thanks!

- [x] 10 trials with highPrecisionTimer in the test/configuration that failed last night
- [x] 10 trials with defaultTimer in the test/configuration that failed last night
- [x] 100 trials with highPrecisionTimer in the test/configuration that failed last night
- [x] checked performance testing on test/studies/shootout

### DIFF
--- a/util/test/computePerfStats
+++ b/util/test/computePerfStats
@@ -267,11 +267,13 @@ def cleanup():
     print("output was: ")
     # if test output was too long, only print first and last 1000 characters
     if len(test_output_raw) > 2000: 
+        print("(first 1000 followed by last 1000)")
         head = test_output_raw[:1000].strip()
         tail = test_output_raw[-1000:].strip()
         head = ''.join(x if x in string.printable else "~" for x in head)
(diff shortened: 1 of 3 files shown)


gbtitus:
I suspect the odd ordering is caused by how the combining tree of processes and pipes puts all the output together for a multi-node job on a Cray X* system, and getting all the output into the same stream as you've done here is probably the best solution.

PR #8167, opened 2018-01-10
Move some "ugni" tests to a new directory

ronawho:
Move some tests that were originally created to track ugni bugs into a new
"test/runtime/configMatters/" dir. I'm going to add this directory to cray
nightly testing, but I don't want to waste time testing all of test/runtime,
since most of the tests aren't dependent on the runtime layer being used.


gbtitus:
I like it!

PR #8170, opened 2018-01-11
Convert SharedObject from constructors to initializers

bradcray:
This case was very straightforward once all the bugs on master were fixed.  :)
(See PR for details)


modules/packages/SharedObject.chpl
```diff
@@ -117,7 +114,8 @@ module SharedObject {
 
        :arg p: the class instance to manage. Must be of class type.
      */
-    proc Shared(p, type t=p.type) {
+    proc init(p, type t=p.type) {
```
bradcray: @mppf: Belatedly, I'm thinking that the `type t` argument was a requirement in the constructors
world that should be moved from the argument list to the initializer body in the current world.
I.e., a user is not expected to be able to pass a type 't' into the initializer, is that right?  (and
ditto for other ones below).

bradcray: (i.e., I don't think I was thinking hard enough about the interface when mechanically converting these).

mppf: Yes that's right. It was there simply to satisfy the compiler.

bradcray:
Testing reveals that this PR is currently held up by a problem with deinit() for (generic?) records with initializers that @mppf is working to resolve in PR #8185.

bradcray:
Testing is now failing for a single test with a complaint that "No copy constructor for initializer" can be found for the second line below:

```chapel
  var x = new Shared(new Impl(1));
  var y = x;
```

in `test/classes/ferguson/delete-free/shared2.chpl` even though there is a copy initializer.  @mppf, do you have any tips on debugging this issue or thoughts as to what could be causing it?

bradcray:
Thanks to Michael's #8255, this is now passing linux64 testing.  @mppf, would you be up for reviewing, as the author of SharedObject?  Also pinging @lydia-duncan and @noakesmichael as keepers of initializers.

lydia-duncan:
Looks reasonable to me!

bradcray:
I think that the correctness of that directory of tests is dependent on being leak-free, but I reran all delete-free tests with --memleaks and also with --valgrindexe and saw no surprises.

Issue #7697, opened 2017-10-31
GMP seems to require m4


lydia-duncan:
Elliot suspects this is a result of #7014

### Configuration Information

- Output of `chpl --version`: chpl Version 1.17.0 pre-release

lydia-duncan:
Requested from #7696

ronawho:
I have a box I can reproduce this on now, so I'll take a look and see what I find

ronawho:
I was wrong / lydia was right -- m4 is a requirement. 

I think there's 2 next steps here:
 - adjust our documentation to list m4 as a prereq (revive https://github.com/chapel-lang/chapel/pull/7696, and add m4 to the apt/yum/zypper commands)
 - Think of a better way to let users know a speculative build failed. The gmp build actually fails with a pretty clear `configure: error: No usable m4 in $PATH`, but since it's a speculative build we sweep that under the rug and users only find out at chpl compilation time with incredibly unhelpful   unknown __mpz_struct (and other gmp type) errors. Maybe we could do something like set CHPL_GMP=failed if the speculative build failed, and in the module code emit a clear error? Or we could have a "production" Chapel install that did a non-speculative build of gmp? Something to think about -- I'll probably open a new issue for this.

bradcray:
> Think of a better way to let users know a speculative build failed.

Couldn't the solution be even simpler?  If the speculative build fails, I'm pretty sure CHPL_GMP defaults to `none`.  So couldn't we add code to the GMP.chpl module along the lines of:

```chapel
if (CHPL_GMP == "none") then
  compilerError("Your Chapel configuration does not currently have GMP enabled.");
```

and maybe a pointer to some documentation for more information about how to enable it?  (set the env. var., rebuild if necessary, recompile)

ronawho:
Yeah, I was just wondering why we didn't already have that check (I thought we did, but turns out we only have it for CHPL_REGEXP) -- I think that's a good first step and will take that on

bradcray:
I'd guess it was just an oversight...

bradcray:
@ronawho: Just to see if I can close this tab: Are you going to follow up with the GMP check in its module, or do you want me / someone else to?

ronawho:
I own that (and I'm working on it ATM)

My current error message is `GMP support not compiled in`, but this is just to match the re2 error of `Regular expression support not compiled in`. I'm open to other error messages if you have a preference

ronawho:
[...]

PR #8171, opened 2018-01-11
Add a compilerError when trying to use the GMP module with CHPL_GMP=none

ronawho:
Currently, if you try to use the GMP module and CHPL_GMP=none, you'll get
non-obvious errors from the backend compiler about implicit declarations of GMP
types and GMP functions. This is especially confusing with speculative builds
of GMP, where the build failure is hidden and CHPL_GMP is silently set to none.
As an improvement, give an explicit compilerError that GMP support is not
available if CHPL_GMP=none.

This also updates a pre-existing error message in the regexp module to be
consistent with the wording used in the GMP one.

Note that the GMP error currently points to the GMP module instead of the user
line. This is a bug: https://github.com/chapel-lang/chapel/issues/8191

Related to https://github.com/chapel-lang/chapel/issues/7697


test/library/standard/GMP/elliot/check-gmp-none-compiler-error.good
```diff
@@ -0,0 +1 @@
+$CHPL_HOME/modules/standard/GMP.chpl:1226: error: GMP support not compiled in
```
ronawho: The bad/wrong part of this PR is that we're reporting the internal GMP line number, not the user line number. I think this is because the compilerError is just top-level in the module, and not inside a user-callable function. I'm not sure I understand the compilers handling of error reporting to solve this myself though.

modules/standard/GMP.chpl
```diff
@@ -1222,9 +1222,9 @@ module GMP {
     }
   }
 
-  /* FUTURE -- GMP numbers with record semantics,
-      expression and operator overloads.
-  */
+  if CHPL_GMP == "none" {
+    compilerError("GMP support not compiled in");
```
ronawho: I'm open to different error messages -- this is just based on a similiar error message for regexp support: https://github.com/chapel-lang/chapel/blob/master/modules/standard/Regexp.chpl#L414

test/library/standard/GMP/elliot/check-gmp-none-compiler-error.good
lydia-duncan: That would probably require recognizing `use`s as contributing to the module init call chain, which is something our compiler doesn't do strongly right now.  If there's a way to remove the line number completely from the output, I would consider that a suitable compromise (though I'm not sure that's possible)

bradcray: Does the Regexp module have the same issue?  compilerError() takes an errorDepth argument which should permit you to pop up a few levels of callchain potentially, but I can never quite remember how it works nor how well it works.  See test/param/diten/errorDepth.chpl, e.g.

modules/standard/GMP.chpl
bradcray: This error is better than no error, but I'd probably want something that helped a newbie user help themselves more (in both GMP and Regexp cases).  Referring directly to `CHPL_GMP` being set to `none` might help them find what they need with a docs search.  Or referring to the need to set CHPL_GMP to a value other than none and rebuilding might also be useful.

test/library/standard/GMP/elliot/check-gmp-none-compiler-error.good
ronawho: Regexp has the compilerError inside the `compile()` proc, since you have to compile a regex in order to use it, so it correctly reports the user error line.  For gmp we _could_ make wrappers for all the gmp init routines and put the compilerError in there, but there's a lot of init routines (and currently they're all just naked externs)

I tried an errorDepth of 0, but that didn't change anything since I think that's the default

modules/standard/GMP.chpl
ronawho: Do you like any of these?

 - "GMP support is not available with CHPL_GMP=none"
 - "Cannot use GMP with CHPL_GMP=none"
 - "Cannot use GMP with GMP_GMP=none, see https://chapel-lang.org/docs/master/usingchapel/chplenv.html#chpl-regexp for more information"


I don't love hardcoding a link, but I also don't love saying something like "set CHPL_GMP and rebuild" because it doesn't feel obvious that CHPL_GMP is an env variable, and it also doesn't seem like it's obvious that we mean rebuild the chpl runtime/third-party vs "recompile" your app

lydia-duncan: Heads up, you're linking against chpl-regexp, not chpl-gmp

bradcray: Maybe the second one?  I could also imagine changing it to say "the GMP module" rather than just "GMP".

It occurs to me belatedly that we should probably add a similar error in the BigInteger module.

test/library/standard/GMP/elliot/check-gmp-none-compiler-error.good
bradcray: I'm OK with putting this in as-is and striving to improve it in the future.

modules/standard/GMP.chpl
lydia-duncan: Wouldn't the BigInteger module benefit from the GMP error, since it uses the GMP module?

lydia-duncan: Or do you mean that the error should be more explicit, since it isn't necessarily clear that BigInteger relies on GMP?

bradcray: Exactly.  And I was thinking / hoping that putting a more explicit error into BigInteger.chpl would cause it to print first (though now I'm worrying that that's incorrect...).

ronawho: Since BigInteger uses GMP you get the compilerError from the gmp module, and I couldn't emit a different error message for BigInteger, even if I put the compilerError before the [`use GMP;`](https://github.com/chapel-lang/chapel/blob/master/modules/standard/BigInteger.chpl#L112).

That's part of the reason I was trying to leave the error a little more vague and not say "the gmp module"

bradcray: I see.  OK, let's go with this then and wait for users to express confusion before trying to do something more involved.

ronawho:
Alright, pushed a slightly updated error message (and made the regexp one match).

I also converted the test to a future (so changes to GMP.chpl line numbers won't cause testing failures) and added a smaller reproducer for the issue.

Any other comments or good to merge?

bradcray:
I haven't given it a line-by-line review or anything like that (esp the tests) but what I've seen is above the bar and I'm guessing the rest of it is reasonably straightforward.

ronawho:
I'm not sure that error would make much sense if a user is just doing a `use BigInteger;`, so I'll probably leave this alone for now

PR #3242, opened 2016-02-03
Bug fix: make PRIM_IS_SUBTYPE handle transitive instantiation

mppf:
Retires test/users/ferguson/generic-default-record.future

The problem was that PRIM_IS_SUBTYPE only checked one level of
instantiation. Arguably it should have a different name, but
PRIM_IS_SUBTYPE is what a : expression maps to in a where clause. Even
before this PR, it would return true if lt is an instantiation of rt.

In this PR, I adjust it to transitively check instantiation and
sub-typing. For some reason, in the particuar generic-default-record
example, the record R had two levels of instantiation.

Passed full quickstart testing.

PR #8173, opened 2018-01-11
Allow :integral to work in where clauses

### LINKED ISSUE #8172: where clause not working with :integral
I would expect the following program to print out "integral" but it does not:

``` chapel
proc foo(x) where x:integral {
  writeln("integral");
}
proc foo(x) {
  writeln("any type");
}

writeln("foo");
var i8:int(8) = 1;
bar(i8);
```

The problem is with isSubTypeOrInstantiation, which does not follow all of the possible "instatiations" that the compiler knows about (through other means than instantiatedFrom) such as :integral).

mppf:
I would expect the following program to print out "integral" but it does not before this PR:

``` chapel
proc foo(x) where x:integral {
  writeln("integral");
}
proc foo(x) {
  writeln("any type");
}

writeln("foo");
var i8:int(8) = 1;
bar(i8);
```

The compiler has special treatment for several numeric types in canInstantiate that needs to be used in the handling for a:b in a where clause, which is computed by the function
isSubTypeOrInstantiation. (For history, note that PR #3242 introduced isSubTypeOrInstantiation)

This PR simply adjusts isSubTypeOrInstantiation to call canInstantiate to handle these special cases.

Fixes #8172.

- [x] full local testing

### DIFF
--- a/compiler/resolution/postFold.cpp
+++ b/compiler/resolution/postFold.cpp
@@ -464,6 +464,11 @@ static bool isSubTypeOrInstantiation(Type* sub, Type* super) {
   if (sub == super) {
     retval = true;
 
+  } else if (canInstantiate(sub, super)) {
+    // handles special cases like dtIntegral, which aren't covered
+    // by at->instantiatedFrom
+    retval = true;
+
   } else if (AggregateType* at = toAggregateType(sub)) {
     for (int i = 0; i < at->dispatchParents.n && retval == false; i++) {
       retval = isSubTypeOrInstantiation(at->dispatchParents.v[i], super);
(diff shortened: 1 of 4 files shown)


mppf:
@lydia-duncan - would you mind reviewing this one? thanks.

compiler/resolution/postFold.cpp
```diff
@@ -464,6 +464,11 @@ static bool isSubTypeOrInstantiation(Type* sub, Type* super) {
   if (sub == super) {
     retval = true;
 
+  } else if (canInstantiate(sub, super)) {
+    // handles special cases like dtIntegral, which aren't covered
+    // by at->instantiatedFrom
+    retval = true;
```
lydia-duncan: Thinking aloud, it might be useful to at some point merge canInstantiate and isSubTypeOrInstantiation into a single function

Issue #8154, opened 2018-01-09
Read configuration files at run time


buddha314:
For instance, if I want to use the `dev` database during this run and the `prod` database during that run, it would be nice to have

dev.cfg
```bash
DB_HOST=localhost
DB_USER=sally
DB_PW=secret
```

prod.cfg
```bash
DB_HOST=10.11.12.hike
DB_USER=ckaepernick
DB_PW=afro
```

Then be able to run my chapel program and point to the file

```bash
./dotslashables --config=prod.cfg
```

There was some discussion of this a while ago, but I don't see a ticket for it.

daviditen:
Are `DB_HOST`, `DB_USER`, and `DB_PW` all declared as `config const` or `config var` in your program?  If yes, they can be set with config files as you request with the option `-f`.

`./dotslashables -f prod.cfg`

buddha314:
Oh, that sounds nice.  Do you have an example lying around I could follow?

daviditen:
Sure thing:

configs.chpl:
```
config const a = 1,
             b = 2.0,
             c = "hello";

writeln((a, b, c));
```

configs.cfg:
```
a=2
b=3.14
c="world"
```

% chpl configs.chpl
% ./configs
(1, 2.0, hello)
% ./configs -f configs.cfg
(2, 3.14, world)

buddha314:
hells^{yes}.  Let me try this out. Will be back shortly.

ben-albrecht:
More about the `-f` flag at the bottom of the [setting configuration variables](https://chapel-lang.org/docs/master/usingchapel/executing.html#setting-configuration-variables) section in the docs.

buddha314:
so, RTFM, Ben? et tu?

buddha314:
I'm not sure what I should do about this: 

```bash
 error: Configuration constants are allowed only at module scope.
```

and I'm not seeing it in the freakin manual, BEN!!!! I'M LOOKING AT YOU ALBRECHT!

buddha314:
Nota bene: I edited the file to be have the variables be `config` variables, but they were inside `main()` so that's what caused the error.  Also, in experimenting I learned you can't have spaces in your config file like

```
FAIREST_IN_THE_LAND = "Betty White"
```

It will barf on the spaces (but not, presumably, Betty White)

buddha314:
[...]

PR #8174, opened 2018-01-11
Update -f configuration file documentation

ben-albrecht:
- Add more examples for the `-f` configuration file documentation
- Explicitly warn users about the whitespace issue
- Add subsections to make it easier to find documentation within configuration variable section.
- Add cross-reference to users guide for configuration variables

This follows discussion from #8154

### DIFF
--- a/doc/rst/usingchapel/executing.rst
+++ b/doc/rst/usingchapel/executing.rst
@@ -100,9 +100,10 @@ Configuration File
 ~~~~~~~~~~~~~~~~~~
 
 Configuration values can also be passed to a Chapel program through a
-configuration file, specified by the execution time ``-f`` option. The format
-is described in the example below.
-
+configuration file, specified by the execution time ``-f`` option.
+Configuration files can contain a whitespace- or newline-delimited list of
+keys and values separated by an assignment operator ``=``. Comments begin
+with the ``#`` character. The examples below demonstrate this format.
 
 Consider the following program:
 
@@ -179,7 +180,7 @@ Setting the Number of Locales
 
 For multi-locale Chapel executions, the number of locales on which to
 execute a program is specified on the executable's command-line.  This
-can be set either using the -nl flag, or by assigning to the built-in
+can be set either using the ``-nl`` flag, or by assigning to the built-in
 numLocales configuration constant using the normal mechanisms.  So, to
 execute on four locales, one could use:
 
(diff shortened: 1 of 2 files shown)


doc/rst/usingchapel/executing.rst
```diff
@@ -91,21 +95,77 @@ or:
...
 
     .. code-block:: sh
 
-        ./hello2-module -fhello2-module.input
+        # config variables are populated by program.input values
+        ./program -fprogram.input
+
+
+.. warning::
+    Assignments cannot contain whitespaces outside of quotes, so the following
```
daviditen: `chplspell` doesn't like the word "whitespaces" - maybe drop the 's' from the end?

doc/rst/usingchapel/executing.rst
```diff
@@ -91,21 +95,77 @@ or:
...
+
+    .. code-block:: python
+
+        # program.input
+
+        msg="hello world"
+        val1=1.61803
+        val2=3.14159
+
+Configuration files can contain a whitespace- or newline-separated list of
```
daviditen: I'd prefer to leave this explanation up at the top paragraph so it's a fairly complete explanation there followed by a nice example.  Then you can just say that multiple configs can be set on one line and order doesn't matter here.

doc/rst/usingchapel/executing.rst
```diff
@@ -91,21 +95,77 @@ or:
...
+        msg="hello world" # This is a comment
+
+
+The ``program.input`` is passed during execution with the ``-f`` flag:
 
     .. code-block:: sh
 
-        ./hello2-module -fhello2-module.input
+        # config variables are populated by program.input values
+        ./program -fprogram.input
```
daviditen: One more.  `chplspell` doesn't like "fprogram" and it could be fixed with a space, but I think it's fair to leave as is.

ben-albrecht: I could just update the dictionary..

doc/rst/usingchapel/executing.rst
cassella: FWIW, I've taken the tack of adding to the chplspell dictionary any word that's natural to use in the project.  As long as it's spelled right. 

Including "words" that are formed by command line option, as "fhello" for the "-fhello" removed below.  :(

ben-albrecht:
@daviditen - let me know if this looks good to you.

daviditen:
Looks good, thanks!

PR #8177, opened 2018-01-11
Reduce time spent in the registered memory table critical sections.

gbtitus:
Some operations on the registered memory region information have to be
done under mutual exclusion in order to avoid corruption.  Specifically,
these are acquiring an empty table entry for a new memory region, and
updating the highest used index.  The old code protected much more than
this in critical sections; here, we reduce the protection to just what
is needed.

When allocating memory, we used to protect the search for an empty table
slot and by extension the check if one was available, the allocation of
hugepages, filling in the entry, and updating the highest used index if
that was necessary.  Now the hugepage allocation and the check for an
available table entry are outside the critical section, and only finding
and filling in the table entry and updating the count are protected.

In the post-allocation function, which registers memory regions and
broadcasts registrations, we used to protect both operations.  Now the
critical section covers only the check to see if we need to broadcast
both the registration count and any new table entries and, if we do, the
broadcast as well.  If only the new entry needs to be broadcast that can
be done safely without mutual exclusion, because no other entity can be
communicating that.

Then when freeing registered memory, we used to protect everything:
deregistering the memory, clearing then entry, and broadcasting it.  By
reordering the operations now we don't need any critical section at all.
First we deregister the memory, then we broadcast a zero entry to all
other nodes, then we zero our own entry, and finally we free the memory.
We no longer reduce the table length, so there is no need to update the
highest-entry index on remote nodes.


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2807,67 +2812,60 @@ void chpl_comm_impl_regMemPostAlloc(void* p, size_t size)
...
-      } else {
-        src_v[ci] = (char*) &mem_regions + off;
-        node_v[ci] = ni;
-        tgt_v[ci] = (char*) &mem_regions_map_addr_map[ni][chpl_nodeID] + off;
-        size_v[ci] = size;
-        ci++;
-      }
-    }
+  if (mr_i < mem_regions_map[chpl_nodeID].mreg_cnt) {
+    mem_regions_unlock();
```
ronawho: Found myself wondering post-review -- even if we don't need the mutex lock, don't we still need to prevent yields here?

gbtitus: We don't need it as long as we're just dealing with array memory, but yes, in the dynamic stack extension case we'll need it.  Good catch!

ronawho: Depending on the perf results, we probably want to split out the lock and yield, so that we can release the mutex, but continue to not yield

PR #8181, opened 2018-01-11
Ask if issue is blocking users on GitHub issue template

ben-albrecht:
Ask issue filers: "Is this a blocking issue with no known work-arounds?"


ben-albrecht:
Separately, I feel that our issue template is a little verbose. I'm open for suggestions on making it more concise without losing any valuable information.

bradcray:
This looks fine to me.  I could also imagine promoting it to a top-level thing with a clear yes/no choice to make sure it doesn't get lost, but maybe we should start here and promote it if people don't answer it.

PR #8184, opened 2018-01-12
retire future: record-deinit-throws

psahabu:
@nspark provided a future where a record deinitializer was marked throws. Given the decision to ban deinit from throwing as of #8161, this future has been retired and modified.


bradcray:
@nspark -- I hadn't realized you'd filed a future against this previously.  Does the presence of the future suggest that you'd hoped to be able to throw from a deinit() routine?  (i.e., that the decision in #8161 isn't what you'd have picked?)  Thanks.

nspark:
@bradcray I think the idea I had was that the callable that drove a `Future` object might throw. The future could capture the error and hold it (in lieu of the value) until either `Future.get()` was called or `Future.deinit()` ran. In both cases, that would be when the future would throw the error from the callable, although the idea that it would throw on `deinit()` was pretty loose.

PR #8187, opened 2018-01-12
Fix default arguments for init

mppf:
After PR #8141, we were seeing failures in getInitialization
(which, as called in parallel.cpp, looks for the initialization
point for global const variables). The AST looked something like this:

```
 DefExpr unixEpoch
 DefExpr def_tmp
 init_default_tzinfo( &unixEpoch, /*ret*/ &def_tmp)
 init( &unixEpoch, /*ret*/ &def_tmp )
```

(Note that PR #7858 rewrote default argument handling to use
 functions returning the default value - instead of a wrapper
 function calling the entire original function).

In other words, the function producing the default value for
an argument to an init function was consuming the 'this' argument.
But that doesn't make sense because 'this' won't be initialized yet
(it's initialized in the initializer). Therefore, I extended logic
in buildDefaultedActualFn and createDefaultedActual that avoids
passing 'this' to the default-argument function to apply to init
functions in addition to constructors.

- [x] full local testing

### DIFF
--- a/compiler/resolution/wrappers.cpp
+++ b/compiler/resolution/wrappers.cpp
@@ -476,10 +476,13 @@ static DefaultExprFnEntry buildDefaultedActualFn(FnSymbol*  fn,
 
   SymbolMap copyMap;
 
+  bool isConstructorOrInit = fn->hasFlag(FLAG_CONSTRUCTOR) ||
+                             0 == strcmp(fn->name, "init");
+
   // Set up the arguments
   if (fn->hasFlag(FLAG_METHOD) &&
       fn->_this != NULL &&
-      !fn->hasFlag(FLAG_CONSTRUCTOR)) {
+      !isConstructorOrInit) {
     // Set up mt and this arguments
     Symbol* thisArg = fn->_this;
     ArgSymbol* mt = new ArgSymbol(INTENT_BLANK, "_mt", dtMethodToken);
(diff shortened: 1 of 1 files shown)


mppf:
@vasslitvinov - could you have a look at this?

bradcray:
This looks like a no-brainer to me.  The only question I had (maybe for @lydia-duncan or @noakesmichael) is whether there is a better way to identify initializers than via a strcmp?

bradcray:
(I should also add the caveat that I don't really know this code at all, so my no-brainer detector may be way out of  kilter...)

lydia-duncan:
We've avoided having a FLAG_INITIALIZER in the past, but I'm not entirely convinced that we should

bradcray:
> We've avoided having a FLAG_INITIALIZER in the past

On what rationale?

noakesmichael:
I can’t recall any specific discussions about this
particular case but I certainly don’t like the convention
of defining flags for purposes like this.  Ultimately they
are set by doing a strcmp somewhere in the compiler
and the code base is littered with ambiguous flags.

However I do feel that it is an oversight that we
have yet to define a C++ method on FnSymbol
that returns true for initializers and false otherwise.







On Jan 12, 2018, at 1:49 PM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


We've avoided having a FLAG_INITIALIZER in the past

On what rationale?

Issue #7935, opened 2017-12-05
Support calling 'new' on type function returning generic class


bradcray:
### Summary of Problem

It appears that, at present, for a generic class supporting an `init()` method, we are unable to write a function that returns that type and call `new` on the result to create an object of that type.  This seems to be true whether the function returns the generic class name or a concrete instantiation of the class.

### Steps to Reproduce

**Associated Future Test(s):**
[`test/classes/initializers/generics/typeFunctions/fnReturnsConcrete.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/generics/typeFunctions/fnReturnsConcrete.chpl
          ) #7934
[`test/classes/initializers/generics/typeFunctions/fnReturnsGeneric.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/generics/typeFunctions/fnReturnsGeneric.chpl
          ) #7934

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.17.0 pre-release (ae93ddd)`

bradcray:
We use this pattern in the sync/single module currently, so to convert it from constructors to initializers in the most obvious way, we'd need to resolve this.

lydia-duncan:
I've started looking into this bug

lydia-duncan:
I need to take a look into how the sync/single implementation works.  The generics test at least doesn't work regardless of whether the type defines an initializer, a constructor, or relies on the default constructor, implying that there is some special compiler support occurring.

lydia-duncan:
#8188 handles the concrete case, which seemed to be what syncs/singles relied upon.

It looks like #5143 doesn't ask for the generic case and syncs/singles don't rely on it (because it doesn't work for constructors) - should I explore what it would take to handle the generic test, or just add an explicit user facing error since both initializers and constructors encounter the same internal compiler error?

bradcray:
No need to handle the generic test in the short-term for my sake.  I wrote it while exploring different ways of expressing the pattern I was shooting for.

lydia-duncan:
#8205 adds a user-facing error message for the function that returned a generic.  We can still implement the support for it later, so I think let's leave this issue open for now but not prioritize it?

bradcray:
Works for me, thanks!

aconsroe-hpe:
Closing because issue has been resolved

Issue #5143, opened 2017-01-10
Expand support for type aliases of concrete-ized generic types


nspark:
I've been trying out some ideas with generic records and classes, and ran into a few issues. I saw #3406 and some of the [associated examples](https://github.com/chapel-lang/chapel/tree/master/test/classes/ferguson), and wanted to see whether the compiler could handle what I was trying to do. I thought filing an issue here might help me document what I'm trying to do.

## Part 1: A problem with type aliases of generic types

After looking at [this example](https://github.com/chapel-lang/chapel/blob/master/test/classes/ferguson/record-alias-generic-concrete.chpl) and experimenting with a few variants, I noticed that the generic type specified by the alias is not actually respected.

The following updated test shows that the constructor argument of "1" is what determines the type of `A.x`, not the type alias:
```chapel
record A {
  var x;
}

type C = A(real);

var c = new C(1);
if (c.x.type != real) then compilerError("Inner type not respected by alias!");
writeln(c);
writeln(c.x);
```

## Part 2: Generic records with type fields

Assuming that Part 1 can be resolved, I'm hoping to work with generic types that have a type field. Again comparing to [this example](https://github.com/chapel-lang/chapel/blob/master/test/classes/ferguson/record-alias-generic-concrete.chpl), I'm hoping to have something like:
```chapel
record A {
  type T;
  var a: T;
}
 
type C = A(int);
 
var c = new C(1);
writeln(c);
writeln(c.a);
```
which currently produces the following error message:
```
test.chpl:8: error: unresolved call 'A.init(1)'
test.chpl:1: note: candidates are: A.init(type T, a: T)
```

## Part 3: Forwarding constructor arguments

Ultimately, I want a type alias that concrete-izes a generic record, where the "inner" type is a class, and the constructor arguments to the type alias get forwarded to the class's constructor. I think this would look something like:

```chapel
class B {
  var b1: real;
  var b2: int;
  proc B(v1: real, v2: int) {
    b1 = v1;
    b2 = v2;
  }
}

record A {
  type T;
  var a: T;
  proc A(args...) {
    a = new T((...args));
  }
}

type C = A(B);

var c = new C(99.0, 42);
writeln(c);
writeln(c.a);
writeln(c.T:string); // prints B
writeln(c.a.b1);     // prints 99.0
writeln(c.a.b2);     // prints 42
```

bradcray:
@mppf: Tagging you to make sure you see this issue.

mppf:
[...]

PR #8188, opened 2018-01-12
Fix issue with initializers on complete instantiations returned from type functions

lydia-duncan:
Prior to this change, we would fail to resolve initializer calls when the type
provided was a generic instantiation of a class instead of the generic base. 
This was because we were recognizing it as non generic, but for generic
classes we only create the _new function after we have resolved an
initializer for that type.

This change adds a new path to resolveNew, which distinguishes non-generic types
from generic instantiations.  It then attempts to resolve the initializer as
though the fully generic type was provided, then compares the determined
instantiation to the one that was initially provided.  If they don't match, then
the best initializer match for the call could not generate the instantiation
desired, so the call was incorrect.  Note that this strategy does not handle the
case where a field could be coerced into the proper type.

Adds a new test, fnReturnsConcrete2, which separated the instance
declaration from the init call.  Fixed fnReturnsConcrete and fnReturnsGeneric
so that they delete the instances they create, and fixed the `new` call
so that it would actually resolve.

Resolves the futures
test/classes/initializers/generics/typeFunctions/fnReturnsConcrete and
test/classes/initializers/generics/typeAlias.chpl and allows the new test
fnReturnsConcrete2 to pass.  It also updates the expected error message
for when records attempted the same operation (which passed because
you could find the initializer from the instantiated type).

Passed classes/initializers with futures, and a full std paratest


lydia-duncan:
@noakesmichael - this is ready for review, but I don't intend to merge it until Monday, so no rush!

lydia-duncan:
Resolves case 2b of #5143, and the concrete case of #7935

compiler/resolution/functionResolution.cpp
```diff
@@ -5798,6 +5807,101 @@ static void resolveNewHandleNonGenericInitializer(CallExpr*      call,
   }
 }
 
+// Similar to resolveNewHandleGenericInitializer, except we have been
```
noakesmichael: I am pretty sure there's something wrong with the tail of this comment i.e.
"and then double check that the type we were given".  Not sure if there's
text missing or if the word "that" is extraneous.

lydia-duncan: I think I forgot to finish that sentence, thanks!

PR #8190, opened 2018-01-12
Initial support for promotion on types with initializers

benharsh:
Instead of dealing with the '_promotionType' field, types with
initializers can implement the 'chpl__promotionType()' method in order
to enable promotion. For example, if your class or record contained a
type field called 'eltType', you can write:

```chpl
proc chpl__promotionType() type {
  return eltType;
}
```

The 'chpl__promotionType' method is currently resolved **after**
initializers, meaning that promotion on the initialized type will not
work within an initializer.

Testing:
- [ ] full local
- [ ] full gasnet


benharsh:
It has been pointed out to me that ``resolveInitializer`` is currently only called for generic initializers. I'll close this until I implement a approach for both generic and non-generic initializers.

PR #8194, opened 2018-01-15
LLVM struct-path alias analysis for records and class objects

dmk42:
This change adds records and class objects to Chapel's LLVM struct-path alias analysis.  Several special case Chapel types were implemented because they are used as record or class object fields, such as the void type, the object type, and complex types.

FileCheck's TBAA testing capabilities are sufficiently complicated that they are best placed in a separate PR.

- [x] Full local testing on linux64 with `CHPL_DEVELOPER` and `CHPL_LLVM_DEVELOPER` set, and `--llvm --fast`
- [x] Mac release/examples with `CHPL_LLVM=system`, LLVM 5, and `--llvm --fast` (expected fft failure)
- [x] Retest after syncing with master
- [x] Test non-LLVM release/examples
- [x] Retest after addressing review feedback


dmk42:

compiler/codegen/symbol.cpp
```diff
@@ -948,64 +962,213 @@ void TypeSymbol::codegenMetadata() {
...
+    // The only other way to handle unions correctly, given the limitations
+    // of LLVM's TBAA metadata design, is to make all unions ancestors
+    // of everything, as we do here.  Clang does this as well.
+    //
+    // This means we don't get any help with alias disambiguation on
+    // unions.  We still need to supply a TBAA type descriptor for them,
+    // though, in case they appear as a member of a class or record.
+    llvmTbaaTypeDescriptor = info->tbaaUnionsNode;
+  } else if (is_complex_type(type)) {
+    // At present, complex types are not records, so we have to
```
mppf: Let's put this long 'complex' case into a different function.

compiler/codegen/symbol.cpp
```diff
@@ -948,64 +962,213 @@ void TypeSymbol::codegenMetadata() {
...
     llvmTbaaTypeDescriptor =
       info->mdBuilder->createTBAAScalarTypeNode(cname, parent);
-  } else if (ct && !isUnion(type) && !hasFlag(FLAG_STAR_TUPLE)) {
-    // Create the TBAA struct type descriptors and tbaa.struct metadata nodes.
-    llvm::SmallVector<llvm::Metadata*, 16> TypeOps;
-    llvm::SmallVector<llvm::Metadata*, 16> CopyOps;
-    llvm::SmallVector<llvm::Metadata*, 16> ConstCopyOps;
+  } else if (isRecord(type)) {
+    if (ct == dtLocaleID) {
+      // Report this as if it were a scalar type because it is a runtime type.
```
mppf: "because it is a runtime type" -> "because it is a type defined by the C runtime"

compiler/codegen/symbol.cpp
```diff
@@ -948,64 +962,213 @@ void TypeSymbol::codegenMetadata() {
...
+      // Change this if the locale ID can contain other than 32-bit integers.
```
mppf: It seems to me that the Chapel compiler should be able to know this? I believe you should be able to request the LLVM type information for "chpl_localeID_t" (using some of the clang integration) and then verify it only has 32-bit integers at this point. Otherwise I'm worried that a seemingly innocent change there will mess up optimization without anybody noticing.

mppf: E.g. can you do `getTypeLLVM("chpl_localeID_t")` ? Or `codegenCType` with the result of `lvt->getType("chpl_localeID_t")` ?

PR #8195, opened 2018-01-15
Convert ZMQ to initializers

bradcray:
When converting the Futures module to use initializers, I also
converted the ExplicitRefCount module, but failed to notice that
ZMQ also uses that module.  And worse, most of our testing
environments don't have ZMQ available, so don't test it by default.
As a result, some ZMQ failures slipped into nightly testing due to
having a mix of initializers and constructors within a given class
hierarchy.

This PR converts ZMQ to initializers.  These cases seemed fairly
straightforward.

### DIFF
--- a/modules/packages/ZMQ.chpl
+++ b/modules/packages/ZMQ.chpl
@@ -471,9 +471,10 @@ module ZMQ {
     var ctx: c_void_ptr;
     var home: locale;
 
-    proc ContextClass() {
-      this.home = here;
+    proc init() {
       this.ctx = zmq_ctx_new();
+      this.home = here;
+      super.init();
       if this.ctx == nil {
         var errmsg = zmq_strerror(errno):string;
         halt("Error in ContextClass(): %s\n", errmsg);
@@ -503,7 +504,7 @@ module ZMQ {
     /*
       Create a ZMQ context.
      */
-    proc Context() {
+    proc init() {
       acquire(new ContextClass());
     }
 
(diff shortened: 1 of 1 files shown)


bradcray:
@lydia-duncan / @mppf: This will resolve the ZMQ failures on 'numa' from last week (apparently the only configuration that tests ZMQ?) if one of you have time to review.  Tagging @nspark for his reference.

bradcray:
I'll drag my feet a bit just in case Lydia wants to take a look from an initializers perspective.

lydia-duncan:
Complete sidebar - this file highlights really weirdly on github.  It looks like it is due to line 147, which "closes" a "nested comment".  @victor-ludorum - would you like to take a look at this?  I think you have the most expertise on updating the tmbundle syntax highlighter.  I think the problem is with its handling of nested comments, which seems to differ from our compiler.

victor-ludorum:
Okay @lydia-duncan I will try my best to solve this issue 😊 !!

bradcray:
> I think the problem is with its handling of nested comments

Though I'm not familiar with tmbundle, I was worried that it might not be powerful enough to handle nested comments.  This is a classic case used in intro CS theory classes to demonstrate how regular expressions / finite automata can't do everything due to the need to store state (in this case, nesting depth).  That said, many modern "regular expression" systems do more than basic regular expressions, so maybe there's hope...

Googling 'tmbundle nested comments' gives a hit for the Swift tmbundle package, and commits like [this one](https://github.com/textmate/swift.tmbundle/commit/662fd22bf8e6d2ed1dfcb5881c23838bad620404) suggest that they have implemented it.

PR #8197, opened 2018-01-15
Allow generic record initializers to call this.init()

### LINKED ISSUE #7960: Internal error in initializer resolution

lydia-duncan:
Prior to this change we would get an internal assertion error in
initializerResolution.cpp when trying to resolve the this.init() call in an
initializer on a generic record.  This was because we were not accounting for
partial calls, which do not resolve until they have been flattened when preFold
is called on their parent call.  The this.init() call was partial when any
arguments were present.

To resolve this change, I modified resolveInitializer to respond appropriately
to when the call fails to resolve the first time (by just returning instead of
acting as though the call had resolved).  I then adjusted the handling
for generic records when the call resolved so that it would handle the
this.init() calls, as their format differs slightly from other record
initializer calls.

With this change, and some slight modifications to the tests themselves, the
futures test/classes/initializers/records/generics/siblingCall-*.chpl now pass.

Resolves #7960 

Test changes:
I had accidentally used the wrong output format for these tests (the output
would match if they were class tests, but they are record tests).  Also,
siblingCall-generic needed to eliminate its single-argument initializer from
consideration as a copy initializer.  Remove their .bad and .future files.

Tested over classes/initializers with futures and a full std/ paratest.


lydia-duncan:
@noakesmichael - do you think it would be appropriate to split the handling of generic records and generic classes into slightly different functions, instead of keeping them both in resolveInitializer?

PR #8198, opened 2018-01-15
Fix chplsys mem functions

mppf:
This PR resolves multiple problems with chpl_bytesPerLocale and
chpl_bytesAvailOnThisLocale.

First, it renames them to better reflect that they come from the runtime
and to accurately state what they do.

  * chpl_bytesPerLocale renamed to to chpl_sys_physicalMemoryBytes
  * chpl_bytesAvailOnThisLocale renamed to chpl_sys_availMemoryBytes

Second, updated FreeBSD/NetBSD in chpl_sys_physicalMemoryBytes to get the
hw.physmem/hw.physmem64 system property. As far as I know, these could
actually use _SC_PHYS_PAGES the way linux does, but we don't test on them
regularly and apparently _SC_PHYS_PAGES was not sufficient for NetBSD at
some point (see commit cab70c9).

Third, update chpl_sys_availMemoryBytes() implementation for Mac OS X,
FreeBSD, and NetBSD. Before this commit, on my Mac OS X system, this call
returned a value that I could not explain.  Note, I think it'd be more
reasonable to use MemAvailable (from /proc/meminfo) on linux for this -
because there might be quite a bit of freeable memory in the page cache,
but I didn't try to do that in this commit. Instead, this commit just
makes the other platforms also return the amount of free memory (which is
at least a reasonable guess at the amount of available memory).

chpl_sys_physicalMemoryBytes supports the locale.physicalMemory function.
It's debateable whether or not we should have a function returning the
amount of user memory a program could allocate with dedicated scheduling -
i.e. total physical memory - kernel memory.  If we decided to add such
  a thing, it would need a different Chapel-level interface.

- [x] full local testing
- [x] memory functions return sane values on various platforms:
    - [x] netbsd
    - [x] freebsd
    - [x] mac os x
    - [x] linux

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -358,22 +358,32 @@ size_t chpl_getHeapPageSize(void) {
 }
 
 
-uint64_t chpl_bytesPerLocale(void) {
+// This function returns the amount of physical memory the
+// system reports there is. This amount of memory is not generally
+// allocatable, since there has to be space for the kernel. This routine
+// makes no attempt to account for kernel space.
+uint64_t chpl_sys_physicalMemoryBytes(void) {
 #ifdef NO_BYTES_PER_LOCALE
   chpl_internal_error("sorry- bytesPerLocale not supported on this platform");
   return 0;
 #elif defined __APPLE__
   uint64_t membytes;
   size_t len = sizeof(membytes);
-  if (sysctlbyname("hw.memsize", &membytes, &len, NULL, 0)) 
+  if (sysctlbyname("hw.memsize", &membytes, &len, NULL, 0))
     chpl_internal_error("query of physical memory failed");
   return membytes;
 #elif defined _AIX
   return _system_configuration.physmem;
+#elif defined __FreeBSD__
+  uint64_t membytes;
+  size_t len = sizeof(membytes);
+  if (sysctlbyname("hw.physmem", &membytes, &len, NULL, 0))
+    chpl_internal_error("query of physical memory failed");
+  return membytes;
 #elif defined __NetBSD__
   uint64_t membytes;
   size_t len = sizeof(membytes);
-  if (sysctlbyname("hw.usermem", &membytes, &len, NULL, 0))
+  if (sysctlbyname("hw.physmem64", &membytes, &len, NULL, 0))
     chpl_internal_error("query of physical memory failed");
   return membytes;
 #elif defined __CYGWIN__
@@ -389,6 +399,8 @@ uint64_t chpl_bytesPerLocale(void) {
   //
 #else
   long int numPages, pageSize;
+  // This code runs for linux, but probably works for FreeBSD/NetBSD.
+  //
   numPages = sysconf(_SC_PHYS_PAGES);
   if (numPages < 0)
     chpl_internal_error("query of physical memory failed");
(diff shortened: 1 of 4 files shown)


mppf:
I was using this program to test the memory functions:

``` chapel
extern proc chpl_sys_physicalMemoryBytes():uint(64);
extern proc chpl_sys_availMemoryBytes():uint(64);

writeln("bytes per locale");
writeln(chpl_sys_physicalMemoryBytes());

writeln("bytes avail on this locale");
writeln(chpl_sys_availMemoryBytes());

config const n = 0.5;
config const nElems = ((n*1024*1024*1024)/8):int;
var A:[1..nElems] int;
A = 1;

writeln("bytes per locale after allocating ", n, " GB");
writeln(chpl_sys_physicalMemoryBytes());

writeln("bytes avail on this locale after allocating ", n, " GB");
writeln(chpl_sys_availMemoryBytes());

A[nElems] = 2;
```

I didn't commit it as a test since the output varies from system-to-system and run-to-run.

mppf:
@ronawho, @gbtitus - you might be interested in looking over this PR fixing up the memory functions including chpl_bytesAvailOnThisLocale.

@gbtitus - while I appreciate that it might be reasonable to have a function returning (physical memory - kernel memory), I checked with Brad and he wants here.physicalMemory to return the amount of system memory (not subtracting anything; i.e. not the allocateable memory). If we decide to add a function for "allocateable memory" we'll need to come up with a new Chapel-level name for it.

@ronawho is planning to ask the GASNet team about when we should be setting GASNET_MAX_SEGSIZE. (If we don't figure it out ourselves first...)

ronawho:
FWIW I think the segsize parts of this PR should be split off into a new PR

mppf:
I can do that.

runtime/src/chplsys.c
```diff
@@ -399,23 +411,82 @@ uint64_t chpl_bytesPerLocale(void) {
...
-#elif defined __NetBSD__
-  int64_t membytes;
-  size_t len = sizeof(membytes);
-  if (sysctlbyname("hw.usermem64", &membytes, &len, NULL, 0))
-    chpl_internal_error("query of hw.usermem64 failed");
-  return (size_t) membytes;
 #elif defined(__linux__)
   struct sysinfo s;
 
+  // MPF: I think this should return MemAvailable not MemFree
```
gbtitus: If the change described in this comment would result in the value returned for Linux being more similar to what we return on other platforms, then you should go ahead and do it now, unless there's a reason to put it off that I'm not aware of.

runtime/src/comm/gasnet/comm-gasnet.c
```diff
@@ -759,17 +759,20 @@ static void set_max_segsize() {
     return;
   }
 
-  // If GASNET_NEEDS_MAX_SEGSIZE is defined then we have to have
-  // GASNET_MAX_SEGSIZE set.  Otherwise, we don't.
-#ifdef GASNET_NEEDS_MAX_SEGSIZE
+  // If segment is large or fast, we need to set GASNET_MAX_SEGSIZE.
+  // Note: this used to be ifdef GASNET_NEEDS_MAX_SEGSIZE
```
gbtitus: It's unusual to leave a comment behind to describe how the code used to work -- is there a particular reason for doing it this time given git's (or whatever's) ability to show history?

gbtitus:
Modulo a couple of line comments this looks good to me.

runtime/src/comm/gasnet/comm-gasnet.c
mppf: I've moved those changes to another branch (per @ronawho's request) and I'll leave him in charge of carrying them forward in whatever form makes sense.

runtime/src/chplsys.c
mppf: This commit makes the other platforms be in line with linux - namely they all currently return the amount of free memory. I don't think that's the right answer long-term (because of the page cache) but I think we need to first resolve issue #8199 and understand a little bit the performance impact of changing this number (since we're not using this value on Cray systems currently anyway to set the GASNet segment size - which is all we do with this function). E.g. if it turns out not to matter for benchmarks, we might remove this function entirely, or tolerate its continued inaccuracies.

PR #8205, opened 2018-01-16
Add a user-facing error message when the return type of a function is generic

lydia-duncan:
Prior to this change, we would encounter some weird internal errors when the
user defined a function that returned a generic type: in normal compilation we
would get the message "every return type must also have a ref type", but generic
types don't get refTypes; when compiling with --verify we would get the message
that the type was no longer in the tree, because types that aren't used get
cleaned up and being a return type does not count as a use (only instances
created count, by causing the resolution of their type constructors).

This change adds a user-facing error message when resolving the return type of a
type function.  If the type is still generic after all operations on it have
finished, then generate this error message.

It also updates test/classes/initializers/generics/typeFunctions/fnReturnsGeneric.
This test was tracking the behavior when a type function returns a generic type.
Since we might eventually want to support this case, leave the test as a future,
but update its .bad file to track the new error message and update the .future
file with a description of the current state.  Since the user-facing error
message prevents both the normal output and output when compiling with --verify,
remove the .compopts file.

Tested over std with futures.


lydia-duncan:
One failure with test/functions/firstClassFns/captureGenericFn

test/classes/initializers/generics/typeFunctions/fnReturnsGeneric.future
```diff
@@ -1,5 +1,9 @@
-bug: Can't write a type function that returns a generic class
+Feature Request: Allow type functions that return a generic class
```
mppf: I'm not really convinced of the value of this particular future. Nonetheless I think you've updated it in a reasonable manner.

mppf:
I can't make up my mind about whether it's better to leave the error tested by that future or if it'd be better to add something with a .good file, so that problems with the error will be caught in normal testing.

lydia-duncan:
Heh, that's fair.  I think it's possible down the road that we might want this to be a feature that gets added (but definitely don't think it's a priority now), and I'm leery of adding both a regular test and a future that match against the same behavior, so decided letting the future check the error message was the best idea.  But I'm happy to change it if you end up deciding that's a bad idea

lydia-duncan:
Maybe linking to the test from Brad's issue is sufficient to track the feature request?

lydia-duncan:
Follow up from chat: still good to merge

Issue #8189, opened 2018-01-12
Resolution failure for function with if-expr as default value for an argument


benharsh:
### Summary of Problem
The compiler fails to resolve a function that uses an if-expr for a default-value of an argument. The if-expr in this case depends on a param bool argument.

The example below is of an initializer that encounters this problem, but normal methods and functions are also impacted.

### Steps to Reproduce

**Source Code:**

```chapel
record Foo {
  param stridable:bool;
  var _stride;

  proc init(param stridable:bool = false,
            _stride = if stridable then 1 else _void) {
    this.stridable = stridable;
    this._stride = _stride;
    super.init();
  }
}

proc main() {
  var f = new Foo();
}
```

**Compile command:**
```
> chpl foo.chpl
foo.chpl:5: In initializer:
foo.chpl:6: error: unresolved call '_if_fn126()'
```

**Associated Future Test(s):**
[`test/classes/lydia/defaultArgumentIsConditional.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/lydia/defaultArgumentIsConditional.chpl) #8206
[`test/classes/lydia/defaultArgumentIsConditional-field.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/lydia/defaultArgumentIsConditional-global.chpl) #8222
[`test/classes/lydia/defaultArgumentIsConditional-global.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/lydia/defaultArgumentIsConditional-field.chpl) #8222
[`test/functions/lydia/defaultArgumentIsConditional.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/lydia/defaultArgumentIsConditional.chpl) #8206
[`test/functions/lydia/defaultArgumentIsConditional-global.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/lydia/defaultArgumentIsConditional-field.chpl) #8222

### Configuration Information
chpl Version 1.17.0 pre-release (fe20d15)

Programs failing under ``--force-initializers``:
- classes/deitz/class/conditional_init_expr
- classes/marybeth/test-init
- parallel/taskPar/sungeun/private
- release/examples/primers/voidVariables
- studies/ssca2/rachels/SSCA2_test (x5)

lydia-duncan:
Never mind, not specific to classes and records.  Just impacts every function.

lydia-duncan:
[...]

PR #8206, opened 2018-01-16
Add two test variants of the bug Ben found with default arguments and cond exprs

lydia-duncan:
The version Ben found involved initializers (see #8189), but it turns out that
regular functions and regular methods are also impacted.  Save both of
these more general versions as futures.


lydia-duncan:
Verified a fresh checkout behaved as expected

PR #8208, opened 2018-01-16
Allow functions implementing forwarding to return iterators

mppf:
See test added by this commit which inadverdently changed
behavior from PR #8073.

This commit allows the compiler-generated forwarding functions
to return iterators, so that forwarding an iterator is possible.

I'm not certain that is the long-term strategy for forwarding iterators,
but it used to work and was broken by PR #8073 and this is a simple
repair.

- [x] full local testing


mppf:
@vasslitvinov - could you review this fix for a bug that came about with forwarding iterators after PR #8073 ?

PR #8220, opened 2018-01-17
"similar" module

aditmehta9:
It can convert string to int and int to string. 👍

### DIFF
--- a/similar.chpl
+++ b/similar.chpl
@@ -1,0 +1,60 @@
+/*
+
+use similar;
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Hi, these conversions can be accomplished with casts (`foo: int` or `foo: string`). Additionally, new modules are generally proposed and discussed as GitHub issues before being developed.

I'm not sure what this module is trying to accomplish, but I think it would be a better fit for a Mason package rather than a part of the Chapel repository.

PR #8222, opened 2018-01-17
Add an error message for conditional expressions in an argument list

lydia-duncan:
This is a continuation of investigating #8189 

Since the solution for this problem and the solution for default initializers
will differ due to the context in which the conditional expression is defined,
punt on the explicit argument version for now by adding a user-facing error
message.  This error will only apply in the inferred argument case, as giving
the argument an explicit type allows it to work.

Adds three additional tests.  Ensures that the implementation would handle
using a global in the if expression, both for the method case and the function
case, and that the implementation would handle using a field in the if expression
in the method case.  The other two tests only handled referencing other
arguments - fix a typo in them where I was accidentally printing an extra line.

Tested over std/ with futures


lydia-duncan:
One failure from test/functions/deitz/test_default_conditional.chpl.  Apparently arguments with default expressions of conditionals worked if the argument they were relying on did not also have a default value of its own . . .

lydia-duncan:
test/functions/deitz/test_default_conditional.chpl worked because the argument with a default value had an explicit type.  The issue is with resolving the conditional expression function call for the type of the argument.

lydia-duncan:
Passed std/ testing

lydia-duncan:
@benharsh  - mind reviewing?

Issue #6145, opened 2017-04-29
try in phase1


cassella:
### Summary of Problem

Fields can't be initialized in a try block in phase 1.

### Steps to Reproduce

**Source Code:**
```chapel
record Try {
  var x = 88;
  proc init() {
    super.init();
    try { writeln("try"); x = 42; }
    catch { x = 77; }
  }
}

record TryAgain {
  var x = 88;
  proc init() {
    try { writeln("try again"); x = 42; }
    catch { x = 77; }
    super.init();
  }
}

var t = new Try();
writeln("Try       ", t);

var ta = new TryAgain();
writeln("Try again ", ta);
```

**Compile command:**
`chpl try-try-again.chpl`

**Execution command:**
```
./try-try-again -nl 1
try
Try       (x = 42)
try again
Try again (x = 88)
```

**Associated Future Test(s):**
None yet

### Configuration Information

```
fortytwo@magrathea:~/src/chapel/test/classes/initializers (try-try-again.chpl)$ chpl --version
chpl Version 1.16.0 pre-release (f5cebb7)

CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none

gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
```

cassella:
I had been thinking that the ``TryAgain`` case should work (printing
``42``).  But then I considered a new case:

```chapel
record TryTryAgain {
  var x = 88;
  proc init() {
    try { writeln("try try again"); x = 42; throw new Error(); }
    catch { writeln("caught");  x = 77; }
    super.init();
  }
}
```

which also prints ``88`` today.  But if it did work it would
[...]

PR #8224, opened 2018-01-17
Convert Buffers from constructors to initializers

bradcray:
This converts Buffers from constructors to initializers, but my
confidence in the rewrite is low because over the course of
working on this PR, the implementation of copy initializers
changed dramatically, yet no failures showed up in testing
(including memory leaks or valgrind testing).  From what I
can tell, only two tests use this module.

For that reason, I've proposed moving this module from standard
to packages as a follow-on PR to indicate reduced confidence
in its implementation and not having much weight exercising it
at present.


bradcray:
@mppf: Would you give this one a careful review when you have time?  As noted in the PR message, I don't have much confidence that I got the copy initializers right, and there doesn't seem to be much testing for `Buffers.chpl` at present (I found two cases).  I did test those two cases with valgrind on and memleaks on and things seemed OK, but I don't feel confident they exercise all the right code paths...

Tagging @lydia-duncan as well from an initializers point of view, though you may want to wait until Michael's had a look first.

modules/standard/Buffers.chpl
```diff
@@ -772,5 +775,5 @@ module Buffers {
     if err then try ioerror(err, "in buffer.copyin");
     return ret;
   }
-}
+  }
```
mppf: Why add a space at the end of the file?

modules/standard/Buffers.chpl
```diff
@@ -140,18 +141,35 @@ module Buffers {
...
+
+  pragma "no doc"
+  proc bytes.init(x: bytes) {
+    if x.home == here {
+      this.home = here;
+      this._bytes_internal = x._bytes_internal;
+      super.init();
+      qbytes_release(x._bytes_internal);
+    } else {
+      this.home = here;
```
lydia-duncan: If we always set `this.home` to `here`, why not do that outside the if statement?

modules/standard/Buffers.chpl
```diff
@@ -140,18 +141,35 @@ module Buffers {
...
+  }
+
+
+  pragma "no doc"
+  proc bytes.init(x: bytes) {
+    if x.home == here {
+      this.home = here;
+      this._bytes_internal = x._bytes_internal;
+      super.init();
+      qbytes_release(x._bytes_internal);
```
lydia-duncan: This looks different from the initCopy function - this performs a qbytes_release while the initCopy performs a qbytes_retain.  Was that intentional?

modules/standard/Buffers.chpl
```diff
@@ -306,16 +308,65 @@ module Buffers {
...
   }
   pragma "no doc"
-  proc buffer.buffer() throws {
+  proc buffer.init() /*throws*/ {
     var error:syserr = ENOERR;
     this.home = here;
+    super.init();
     error = qbuffer_create(this._buf_internal);
-    if error then try ioerror(error, "in buffer constructor");
+    if error then try! ioerror(error, "in buffer constructor");
```
lydia-duncan: I'd leave a note on this function about wanting to eventually allow it to throw when you can do that from initializers.

modules/standard/Buffers.chpl
```diff
@@ -306,16 +308,65 @@ module Buffers {
...
+  }
+  pragma "no doc"
+  proc buffer.init(x: buffer) {
+    if x.home == here {
+      this.home = here;
+      this._buf_internal = x._buf_internal;
+      super.init();
+      qbuffer_release(x._buf_internal);
```
lydia-duncan: Same question about retain versus release here

modules/standard/Buffers.chpl
```diff
@@ -306,16 +308,65 @@ module Buffers {
...
+    } else {
+      var error: syserr = ENOERR;
+      this.init(error);
+      if error then halt("Got error on buffer_create");
```
lydia-duncan: I might say "when initializing the type" instead, since "buffer_create" to me implies a call to a function named buffer_create.

modules/standard/Buffers.chpl
mppf: Now I'm embarassed I didn't catch that...

lydia-duncan:
I suspect that you are getting slightly different behavior here.  It looks like the initCopy functions are defined to reuse the same variable, while a copy initializer should actually create a new instance.  But I'm not an expert on writing initCopy functions . . .  @mppf, does that match your interpretation of those functions?

mppf:
the initCopy functions could return the argument & do some fixup on it (which works when reference counting). With a copy initializer that's not really a form you can write it in.

modules/standard/Buffers.chpl
bradcray: Oops, sorry that slipped in... I had a mis-matched curly bracket at some point and when trying to track it down, it seems I forgot to put things back as they were...  My confidence was sufficiently low that this would pass review that I didn't scrutinize the diff as much as I should've.

modules/standard/Buffers.chpl
bradcray: I guess because I was trying to retain as much of the original code as possible (not that it helped minimize the diffs) and didn't really expend any effort to try and understand what it did.

modules/standard/Buffers.chpl
bradcray: This relates back to my lack of confidence in this change.  For some reason, I was thinking that since the copyInit() returned `x`, the behavior needed to be different in a copy initializer that was creating a new value and tried this pattern based on how the Owned/Shared modules implement assignment.  When it passed testing I wasn't sure whether to feel like I got it right or whether testing just didn't exercise copies.  Based on Michael's feedback I'll change it to qbytes_retain(), assume that's right and re-test.

modules/standard/Buffers.chpl
bradcray: I was going to open an issue against that and rely on the commented out `throws` as a placeholder for the intention, but I'll add a comment to this line as well.

bradcray:
> while a copy initializer should actually create a new instance

My understanding is that these copy initializers are creating new instances, but that their guts are pointing to an existing class (and that that's the intended behavior).  But again, I don't really understand this module well and was surprised I could pass testing without new leaks or valgrind errors.

modules/standard/Buffers.chpl
lydia-duncan: #6145 has the most recently documented discussion of that, but I believe the thinking has advanced beyond that.  You'll want to ask @psahabu

modules/standard/Buffers.chpl
```diff
@@ -319,6 +318,8 @@ module Buffers {
     this.home = here;
     super.init();
     error = qbuffer_create(this._buf_internal);
+    // TODO: really want the following to by `try` once we can throw from
```
lydia-duncan: Typo: "by" -> "be"?

bradcray:
Hi @mppf and @lydia-duncan : I think I've implemented the changes you guys wanted correctly.  I'm still passing full linux64 testing plus valgrind and memleaks testing on the tests that exercise `Buffers`.  This makes me even less confident that the code is correct (either version gave the same behavior) and more confident that this module should not yet be considered a standard module.  I think we should move it to packages in a separate PR but am not sure how to proceed with this PR with confidence.  Any suggestions?

mppf:
@bradcray - we could certainly add testing to increase confidence. But I think you should merge this PR and then move it to packages.

bradcray:
OK.  And have you re-reviewed the latest changes to have confidence that they're more correct than the original draft?

modules/standard/Buffers.chpl
```diff
@@ -140,18 +141,34 @@ module Buffers {
...
+  }
+
+
+  pragma "no doc"
+  proc bytes.init(x: bytes) {
     this.home = here;
+    if x.home == here {
+      this._bytes_internal = x._bytes_internal;
+      super.init();
+      qbytes_retain(x._bytes_internal);
```
mppf: Note that this could also say `qbytes_retain(this._bytes_internal)`. I think that'd be clearer, but it shouldn't matter.

modules/standard/Buffers.chpl
```diff
@@ -306,16 +307,67 @@ module Buffers {
...
+    // initializers
+    if error then try! ioerror(error, "in buffer constructor");
+  }
+  pragma "no doc"
+  proc buffer.init(x: buffer) {
+    if x.home == here {
+      this.home = here;
+      this._buf_internal = x._buf_internal;
+      super.init();
+      qbuffer_retain(x._buf_internal);
```
mppf: ditto here.

modules/standard/Buffers.chpl
mppf: Actually on second thought the ideal story for reference counting is to increase the reference count immediately once we are in the function. Can we do `qbytes_retain(x._bytes_internal);` before setting `this._bytes_internal` ?
But all of this feedback is probably in the noise - you could go ahead as-is.

PR #8225, opened 2018-01-17
Adjust RVF serialize/deserialize for new auto destroy placement

mppf:
PR #8073 migrated an auto-destroy for coforall index variables
to the task function. If the task function is also an on, there
might be RVF and serialization occurring. For the following two
tests, that was causing a yielded domain to be destroyed twice.

    studies/hpcc/STREAMS/bradc/stream-block1dpar-arr
    studies/hpcc/STREAMS/bradc/stream-block1dpar

To fix the issue, this commit adjusts insertSerialization to
explicitly handle the case in which an argument is destroyed
in the task function. If the argument is still passed in some
form, it will still be destroyed there. If the argument is no
longer passed (which is the case when there isn't a runtime type),
instead destroy the index variable immediately after the
serialize call.

The adjustment adds asserts to check that the auto-destroy
call in question is only present for coforall index variables,
since this code should be double-checked if that ever changes.

Passed full gasnet testing.
Passed full local testing.

### DIFF
--- a/compiler/optimizations/remoteValueForwarding.cpp
+++ b/compiler/optimizations/remoteValueForwarding.cpp
@@ -397,6 +397,7 @@ static bool isSufficientlyConst(ArgSymbol* arg) {
 // BHARSH TODO: Split this function into more easily-digestible pieces
 // BHARSH TODO: capture the assumptions made here in documentation
 //
+static CallExpr* findDestroyCallForArg(ArgSymbol* arg);
 static void insertSerialization(FnSymbol*  fn,
                                 ArgSymbol* arg) {
   Type* oldArgType    = arg->type;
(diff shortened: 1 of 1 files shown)


compiler/optimizations/remoteValueForwarding.cpp
```diff
@@ -511,6 +544,27 @@ static void insertSerialization(FnSymbol*  fn,
   }
 }
 
+static CallExpr* findDestroyCallForArg(ArgSymbol* arg) {
+  // Scroll backwards looking for an auto-destroy call for arg
+  // in the function defining arg.
+
+  FnSymbol* fn = arg->getFunction();
+  for (Expr* stmt = fn->body->body.tail;
```
benharsh: Would it be better to use the list of SymExprs stored in ``arg``? I don't really care about the performance here. It just seems like a more straightforward way than scanning the entire function body.

mppf: I thought it important not to find autoDestroy calls inside conditionals etc, but I debated about this. Certainly the AST pattern we have seen for this always has it at the end of the function.

benharsh: Fair point, works for me!

PR #8226, opened 2018-01-17
Cleanup how we detect the shared heap type in jemalloc shim

ronawho:
Instead of a boolean for `have_fixed_comm_layer_heap` add an enum to track `FIXED`,
`DYNAMIC`, and` NONE`.

The only functional difference is that we won't do an "initial" setup
allocation after initializing the shared heap for the DYNAMIC case.


ronawho:
@gbtitus I find this a little easier to follow, any objections?

gbtitus:
No object and looks fine.   👍

Issue #8209, opened 2018-01-16
Errors while building


buddha314:
I don't know if this is useful, please close if not.  When building from a fresh pull, I get these messages:

```
clang++ -c -std=gnu++11  -MMD -MP -O3 -I. -I../include/darwin -I../include    -Wno-error -o ../../build/compiler/darwin/clang/llvm-none/parser/bison-chapel.o bison-chapel.cpp
clang++ -c -std=gnu++11  -MMD -MP -O3 -I. -I../include/darwin -I../include    -Wno-error -o ../../build/compiler/darwin/clang/llvm-none/parser/flex-chapel.o flex-chapel.cpp
flex-chapel.cpp:1015:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register yy_state_type yy_current_state;
        ^~~~~~~~~
flex-chapel.cpp:1016:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register char *yy_cp, *yy_bp;
        ^~~~~~~~~
flex-chapel.cpp:1016:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register char *yy_cp, *yy_bp;
        ^~~~~~~~~
flex-chapel.cpp:1017:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register int yy_act;
        ^~~~~~~~~
flex-chapel.cpp:1071:4: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
                        register YY_CHAR yy_c = yy_ec[YY_SC_TO_UI(*yy_cp)];
                        ^~~~~~~~~
flex-chapel.cpp:2005:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register char *dest = YY_CURRENT_BUFFER_LVALUE->yy_ch_buf;
        ^~~~~~~~~
flex-chapel.cpp:2006:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register char *source = yyg->yytext_ptr;
        ^~~~~~~~~
flex-chapel.cpp:2007:2: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
        register int number_to_move, i;
        ^~~~~~~~~
[...]

PR #8227, opened 2018-01-17
Update the version of flex used to generate the lexer

daviditen:
The version we have used up to now was flex 2.5.35.  It generated variables
with 'register' storage class specifiers, which modern C++ compilers warn
about being deprecated.

The version I updated to is flex 2.6.4.  This version does not use the
'register' storage class specifier.  It does use the `noreturn` attribute on
a function that calls `exit()` on all paths, which is fine except that we've
replaced `exit` with `clean_exit()` and `clean_exit()` wasn't marked with the
`noreturn` attribute, so generated a new warning.  I added the attribute to
`clean_exit()` and the parser directory now builds without warnings.

To make sure we don't fall back to an old version of flex, I added a check to
the "parser" rule in compiler/parser/Makefile to check the flex version and
require at least 2.6.0. This shouldn't cause problems for end-users because we
build the generated files ourselves and keep them checked in to the repository.

### DIFF
--- a/compiler/include/bison-chapel.h
+++ b/compiler/include/bison-chapel.h
@@ -1,19 +1,19 @@
-/* A Bison parser, made by GNU Bison 2.5.  */
+/* A Bison parser, made by GNU Bison 3.0.4.  */
 
 /* Bison interface for Yacc-like parsers in C
-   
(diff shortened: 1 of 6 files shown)


daviditen:
This passed a full linux64 paratest.

This will fix issue #8209 .

daviditen:
Thanks @ronawho!

PR #8229, opened 2018-01-17
Update the checkCopyrights script

awallace-cray:
This change updates the script used to check source files for Cray copyright notices (including the current year). 
    
The list of files in previously-scanned toplevel subdirs is adjusted to reflect the kinds of "source" files they now contain, or seem most likely to be added from now on.
    
Recently-added top-level subdir "tools" will now be scanned- before this change, "tools" was not scanned at all.

The regexp used to check for valid copyright year is slightly modified. Now, the LAST 4-digit year in the "Copyright ... Cray Inc" line must be the current year. For ex, "2018-2004" would no longer be accepted.

This change also adds the Cray copyright notice to some source files which had been missed for some reason.

### DIFF
--- a/runtime/src/qio/regexp/re2/re2-interface.cc
+++ b/runtime/src/qio/regexp/re2/re2-interface.cc
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2004-2018 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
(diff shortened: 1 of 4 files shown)


awallace-cray:
tested on:
- Chapel release tarball 
- Cray module 
- chapel.git workspace, before and after local "make": Linux and Mac OS

PR #8233, opened 2018-01-18
Limit darwin gnu testing to just release/examples

ronawho:
gnu on darwin isn't a configuration we care deeply about and with #8214 our
testing times have increased by several hours, so limit this config to only
release/examples.


ronawho:
@awallace-cray just fyi

PR #8236, opened 2018-01-18
update ZMQ and Crypto to throw errors

psahabu:
To put more weight on the error handling system, this PR moves the ZMQ and Crypto modules from halts to throwing errors.


psahabu:
Splitting this PR in two, see #8244 and #8245.

PR #8237, opened 2018-01-18
Allow FileCheck to be used when CHPL_LLVM=system

dmk42:
Previously, the LLVM tests requiring FileCheck were only run when CHPL_LLVM=llvm.  This change adds a search for the location of FileCheck, which is needed when CHPL_LLVM=system, so that the tests can run in that environment.

Tested on Ubuntu and Mac with CHPL_LLVM=system.
Tested on linux64 with CHPL_LLVM=llvm to make sure that still works.
Tested on Mac with CHPL_LLVM unset to make sure the tests are still properly skipped in that case.

### DIFF
--- a/test/llvm/PREDIFF
+++ b/test/llvm/PREDIFF
@@ -8,6 +8,12 @@ PLATFORM=`$CHPL_HOME/util/chplenv/chpl_platform.py --host`
 COMPILER=`$CHPL_HOME/util/chplenv/chpl_compiler.py --host`
 FILECHECK=${CHPL_HOME}/third-party/llvm/install/${PLATFORM}-${COMPILER}/bin/FileCheck
 
+if [ ! -x $FILECHECK ]; then
+    PREFERRED_LLVM_VERS=`cat ${CHPL_HOME}/third-party/llvm/LLVM_VERSION`
+    LLVM_CONFIG=`${CHPL_HOME}/third-party/llvm/find-llvm-config.sh $PREFERRED_LLVM_VERS`
+    FILECHECK=${LLVM_CONFIG//llvm-config/FileCheck}
+fi
+
 mv $OUTFILE $TMPFILE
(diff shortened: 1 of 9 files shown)


dmk42:

test/llvm/PREDIFF
```diff
@@ -8,6 +8,12 @@ PLATFORM=`$CHPL_HOME/util/chplenv/chpl_platform.py --host`
 COMPILER=`$CHPL_HOME/util/chplenv/chpl_compiler.py --host`
 FILECHECK=${CHPL_HOME}/third-party/llvm/install/${PLATFORM}-${COMPILER}/bin/FileCheck
 
+if [ ! -x $FILECHECK ]; then
+    PREFERRED_LLVM_VERS=`cat ${CHPL_HOME}/third-party/llvm/LLVM_VERSION`
+    LLVM_CONFIG=`${CHPL_HOME}/third-party/llvm/find-llvm-config.sh $PREFERRED_LLVM_VERS`
+    FILECHECK=${LLVM_CONFIG//llvm-config/FileCheck}
```
mppf: I assume this is intended to replace llvm-config with FileCheck?
I can't find this syntax in the Make manual. I'd assume it'd be documented here 
  https://www.gnu.org/software/make/manual/html_node/Text-Functions.html
Anyway would it be clearer to use ${subst llvm-config,FileCheck,${LLVM_CONFIG}} (but I havn't tested that)?

dmk42: It's funny that we both made the same mistaken assumption.  I first wrote this in Makefile syntax, but that didn't work because it is a bash script instead.

mppf: haha OK right you are now it's in the bash manual. Thanks. Not sure why I thought we were writing Makefiles (I guess it just looks Make-y).

PR #8238, opened 2018-01-18
Use just a single transaction for each memory region table broadcast.

gbtitus:
In the case where we needed to broadcast both one or more memory region
table entries and a new entry count we've been doing two transactions,
one for the entries and one for the count.  Here, combine the two into a
single broadcast.  This will make some upcoming planned changes simpler,
and might marginally improve performance.  Limited 16-node testing on a
Cray XC system seems to show a slight performance benefit, but that
could be illusory.  In any event, it's hard to imagine it hurting.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1305,7 +1305,7 @@ static void      ensure_registered_heap_info_set(void);
 static void      make_registered_heap(void);
 static size_t    get_hugepage_size(void);
 static void      set_hugepage_info(void);
-static void      regMemBroadcast(void*, size_t);
+static void      regMemBroadcast(int, int, chpl_bool);
 static void      exit_all(int);
 static void      exit_any(int);
 static void      rf_handler(gni_cq_entry_t*, void*);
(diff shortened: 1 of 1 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2983,7 +2983,7 @@ void chpl_comm_impl_regMemPostAlloc(void* p, size_t size)
             "chpl_comm_regMemPostAlloc(): entry %d, bcast",
             mr_i);
     PERFSTATS_INC(regMem_bCast_cnt);
-    regMemBroadcast(mr, sizeof(*mr));
+    regMemBroadcast(mr_i, 1, false /*send_mreg_cnt*/);
```
ronawho: I tend to prefer `/*send_mreg_cnt=*/false` over `false /*send_mreg_cnt*/`

gbtitus: I'm tempted, but in this file there are already other instances of this sort of thing, using the same construction I did here.  So I think in the name of consistency I'll leave it for now.

Issue #8146, opened 2018-01-08
Different path functions for different OS


victor-ludorum:
I am working to add two path function in Chapel language i.e **normCase()** function #6013 and **isAbsPath()** function #6010 . Each having the same issue that the function definition are different for different OS. But, there is only one separate file in Chapel module for all path function instead of two or three for each OS like Python. Kindly suggest any ways to implement these functions .

mppf:
> function definition are different for different OS

Could it be as simple as using some `if` conditionals based upon `CHPL_TARGET_PLATFORM` ?

What is the behavior that differs for different platforms? So far, Chapel assumes a unix-like environment.

lydia-duncan:
normCase's behavior varies depending on if the file system is case sensitive, and converts forward-slashes to back-slashes on Windows systems.  isAbsPath should only accept paths like "C://path/to/foo" if it is on a windows system

mppf:
For normCase, it's not platform dependent, it's filesystem dependent. E.g. on Mac OS X you can make HFS+ filesystems that are case sensitive even though they are usually case insensitive. So it seems to me the code has to query whether or not the filesystem is case sensitive.

For isAbsPath, I'd expect that we'd ignore functionality for such windows-only paths for now, since that's what we've done everywhere else AFAIK.

victor-ludorum:
@mppf Thanks!! for your advice . But Can you elaborate how can I use _CHPL_TARGET_PLATFORM_  as in the [doc](https://chapel-lang.org/docs/1.15/usingchapel/chplenv.html#readme-chplenv-chpl-target-platform) too there is no example to suggest how can I use this . Also Can you suggest any method to check **case sensitivity of file system** as in other languages (like in _python_ and _node.js_ ) they have separated these functions in separate modules so I have not got any help  . 

> For isAbsPath, I'd expect that we'd ignore functionality for such windows-only paths for now, since that's what we've done everywhere else AFAIK.

As the problem of **isAbsPath()** function , the same problem is with other functions too **expandVars()** and **expandUser()** . So , do I only implement for Unix-like environment ignoring Windows environment .

bradcray:
Taking one piece of this:

> Can you elaborate how can I use CHPL_TARGET_PLATFORM

[...]

PR #8239, opened 2018-01-18
Fix issue with isAbsPath on crays

lydia-duncan:
cray wasn't included in the list of Unix-like environments for isAbsPath.
This just changes the check so that it only errors for cygwin.  We think this
is okay because most systems are Unix-like and so even if we don't know about
them or test against them now, they will probably be accurate.

Testing:
- [x] full paratest over std with futures
- [x] test/library/standard/Path, test/library/standard/FileSystem with futures on cygwin64
- [x] double check test on cray

### DIFF
--- a/modules/standard/Path.chpl
+++ b/modules/standard/Path.chpl
@@ -273,7 +273,8 @@ proc file.realPath(): string throws {
 */
 
   proc isAbsPath(name: string): bool {
-    if (isUnixEnv()) {
+    if (CHPL_TARGET_PLATFORM != "cygwin64" &&
+        CHPL_TARGET_PLATFORM != "cygwin32") {
         if name.isEmptyString() {
            return false;
         }
@@ -286,25 +287,6 @@ proc file.realPath(): string throws {
           return false;
     } else {
       compilerError("Target platform should have Unix like environment");
-    }
-  }
-
-  /* Helper to determine if we are in a Unix environment.  Not sure yet how
-     useful this will be to other functions, but I didn't want to write that
-     big conditional in the function itself */
-  private proc isUnixEnv() param {
-    if (CHPL_TARGET_PLATFORM == "linux64" ||
-        CHPL_TARGET_PLATFORM == "linux32" ||
-        CHPL_TARGET_PLATFORM == "darwin" ||
-        CHPL_TARGET_PLATFORM == "netbsd64" ||
-        CHPL_TARGET_PLATFORM == "netbsd32" ||
-        CHPL_TARGET_PLATFORM == "cray-xk" ||
-        CHPL_TARGET_PLATFORM == "cray-xe" ||
-        CHPL_TARGET_PLATFORM == "cray-xc" ||
-        CHPL_TARGET_PLATFORM == "cray-cs") {
-      return true;
-    } else {
-      return false;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


modules/standard/Path.chpl
```diff
@@ -286,8 +284,27 @@ proc file.realPath(): string throws {
...
+        CHPL_TARGET_PLATFORM == "netbsd32" ||
+        CHPL_TARGET_PLATFORM == "cray-xk" ||
+        CHPL_TARGET_PLATFORM == "cray-xe" ||
+        CHPL_TARGET_PLATFORM == "cray-xc" ||
+        CHPL_TARGET_PLATFORM == "cray-cs") {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
ronawho: This feels like a real maintenance burden to me (and one that we will always be reactionary about -- i.e. we will only add new platforms when users file a bug, we will never remember to update this when new platforms are added.)

Surely there has to be a programatic way to check if we're using a posix filesystem

ronawho: (I'd be ok with this going in to fix nightly testing, but at the very least I think we need to open an issue about doing this programmatically)

mppf: I think it'd be reasonable & acceptable to list the *non-unix* platforms we know about. Is that just cygwin right now?

lydia-duncan: I agree that it is a maintenance burden.  We also have an issue about how to perform different behavior on different file systems (#8146, which I pointed people to earlier), but the answer from the only person who commented on the issue was "punt for now" - performing the work and performing the check in the way you are describing will require figuring this out.

I considered listing all the non-unix platforms, but was concerned about accidentally claiming to support cases that we don't actually support.  If we were able to add support for native Windows as a separate config from cygwin, we would then have isAbsPath approve of a path that is not valid on Windows, silently.  It seems better to give an error when we aren't sure than to fail to give an error when we should.

For logging sake: @dmk42 asked me this question offline and the answer I gave then was that what he was proposed on canonicalizing the filename was another Path function we have proposed but have not yet implemented (but that the developer who contributed this function is working his way towards).

mppf: >  It seems better to give an error when we aren't sure than to fail to give an error when we should.

Isn't this the role of testing? As long as we have testing that will fail in such a configuration, I'm comfortable with the result. I.e. if we list Cygwin* for now, and somebody adds windows, can we arrange for one of these tests to fail if they ran full local testing?

lydia-duncan: > can we arrange for one of these tests to fail if they ran full local testing?

I don't think we can reasonably predict how to detect an operating system that we haven't supported before.  The windows case was just an example and it happened to be something that exists today.

mppf: We can reasonably detect when isAbsPath doesn't work on an OS, can't we?
Or is the problem that there might be new & different types of paths that isAbsPath doesn't support? Isn't it sufficient to just say "We assume everything is unix" like we do everywhere else?

ronawho: There's really only posix and windows filesystems (I think this is all boost and python support) -- https://docs.python.org/3/library/os.path.html (python also has an old-style mac path, but that's not really old at this point)

I'm very much in favor of just doing an error for cygwin.

lydia-duncan: Okay, I updated the code and am re-running testing now.  Thanks for the feedback!  Anything else of concern?

lydia-duncan: Upon discussing further with Elliot, he's got a modification to the test file itself so that it will actually detect incorrect usage.

PR #8243, opened 2018-01-18
Pull gatherLoopDetails into its own file

mppf:
This PR moves gatherLoopDetails and supporting functions into their own file, loopDetails.cpp. These functions are declared in the header file compiler/resolution/loopDetails.h since the expectation is that they only should be called from other resolution passes.

I found gatherLoopDetails necessary for my prototype work on lifetime checking. Since it would be used beyond cullOverReferences, it seemed time to pull it out of cullOverReferences.cpp and create its own .h file too.

Passed full local testing.


mppf:
@psahabu - I should probably remove that comment in this PR :)

It's my current belief is that with better loop representation we might not need `gatherLoopDetails()`. Adding it to ForLoop might just confuse any future work on improving the loop representation.

Additionally,`gatherLoopDetails()` in its current form only makes sense in late-resolution passes. Adding it to ForLoop might suggest to compiler developers that it can be called throughout compilation.

PR #8244, opened 2018-01-18
update Crypto to throw errors

psahabu:
To put more weight on the error handling system, this PR moves the Crypto module from halts to throwing errors.

### DIFF
--- a/modules/packages/Crypto.chpl
+++ b/modules/packages/Crypto.chpl
@@ -190,10 +190,10 @@ module Crypto {
        :rtype: `[] string`
 
     */
-    proc toHex() {
+    proc toHex() throws {
       var buffHex: [this.buffDomain] string;
       for i in this.buffDomain do {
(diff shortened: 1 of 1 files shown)


bradcray:
Same comment as on #8244:  If no tests were changed or added as part of this PR, are we really putting any more weight on error-handling?  What would it take to add such tests or update existing ones to put that weight on them?

saru95:
As a first step, we should try to replace all the `halts` with error handling. Maybe then we can replace the tests that test for `halt` with these ones.

bradcray:
I didn't realize that hadn't already been done.  I agree that that would be a good evolution of this code.

psahabu:
#8329 has been created to address `halt`s in the Crypto module.

Lots of the standard modules have halts as well. I'm going to explore removing them after I finish deprecating the 'out error' pattern.

PR #8245, opened 2018-01-18
update ZMQ to throw errors

psahabu:
To put more weight on the error handling system, this PR moves the ZMQ module from halts to throwing errors.

### DIFF
--- a/modules/packages/ZMQ.chpl
+++ b/modules/packages/ZMQ.chpl
@@ -772,7 +772,7 @@ module ZMQ {
         if (0 != zmq_msg_init_data(msg, copy.c_str():c_void_ptr,
                                    copy.length:size_t, c_ptrTo(free_helper),
                                    c_nil)) {
-          try throw_socket_error("Error in Socket.send(%s): %s\n");
+          try throw_socket_error(errno, "send");
         }
 
         // Send the message
@@ -781,7 +781,7 @@ module ZMQ {
           if errno == EAGAIN then
             chpl_task_yield();
           else {
-            try throw_socket_error("Error in Socket.send(%s): %s\n");
+            try throw_socket_error(errno, "send");
           }
         }
       }
@@ -798,7 +798,7 @@ module ZMQ {
           if errno == EAGAIN then
             chpl_task_yield();
           else {
-            try throw_socket_error("Error in Socket.send(%s): %s\n");
+            try throw_socket_error(errno, "send");
           }
         }
       }
@@ -817,12 +817,8 @@ module ZMQ {
       on classRef.home {
         var copy = data;
         param N = numFields(T);
-        try {
-          for param i in 1..(N-1) do
-            send(getField(copy,i), ZMQ_SNDMORE | flags);
-        } catch e: TaskErrors {
-          throw e;
-        }
+        for param i in 1..(N-1) do
+          try send(getField(copy,i), ZMQ_SNDMORE | flags);
 
         try send(getField(copy,N), flags);
       }
@@ -853,7 +849,7 @@ module ZMQ {
         // Initialize an empty ZeroMQ message
         var msg: zmq_msg_t;
         if (0 != zmq_msg_init(msg)) {
-          try throw_socket_error("Error in Socket.recv(%s): %s\n");
+          try throw_socket_error(errno, "recv");
         }
 
         // Receive the message
@@ -862,7 +858,7 @@ module ZMQ {
           if errno == EAGAIN then
             chpl_task_yield();
           else {
-            try throw_socket_error("Error in Socket.recv(%s): %s\n");
+            try throw_socket_error(errno, "recv");
           }
         }
 
@@ -873,7 +869,7 @@ module ZMQ {
                              length=len, size=len+1,
                              owned=true, needToCopy=true);
         if (0 != zmq_msg_close(msg)) {
-          try throw_socket_error("Error in Socket.recv(%s): %s\n");
+          try throw_socket_error(errno, "recv");
         }
 
         // Return the string to the calling locale
@@ -894,7 +890,7 @@ module ZMQ {
           if errno == EAGAIN then
             chpl_task_yield();
           else {
-            try throw_socket_error("Error in Socket.recv(%s): %s\n");
+            try throw_socket_error(errno, "recv");
           }
         }
         ret = data;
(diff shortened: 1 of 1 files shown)


modules/packages/ZMQ.chpl
```diff
@@ -802,29 +798,33 @@ module ZMQ {
...
       on classRef.home {
         var copy = data;
         param N = numFields(T);
-        for param i in 1..(N-1) do
-          send(getField(copy,i), ZMQ_SNDMORE | flags);
-        send(getField(copy,N), flags);
+        try {
+          for param i in 1..(N-1) do
+            send(getField(copy,i), ZMQ_SNDMORE | flags);
+        } catch e: TaskErrors {
```
mppf: Why are we catching TaskErrors here? Where's the tasks?

modules/packages/ZMQ.chpl
```diff
@@ -847,14 +847,13 @@ module ZMQ {
...
-    proc recv(type T, flags: int = 0) where isString(T) {
+    proc recv(type T, flags: int = 0) throws where isString(T) {
       var ret: T;
       on classRef.home {
         // Initialize an empty ZeroMQ message
         var msg: zmq_msg_t;
         if (0 != zmq_msg_init(msg)) {
-          var errmsg = zmq_strerror(errno):string;
-          try! halt("Error in Socket.recv(%s): %s\n".format(string:string, errmsg));
+          try throw_socket_error("Error in Socket.recv(%s): %s\n");
```
mppf: I don't like the way that throw_socket_error assumes that the string has 2 format specifiers in it.
Can we just make throw_socket_error take in "send" or "recv" and put together the rest of the existing error format?

modules/packages/ZMQ.chpl
```diff
@@ -908,23 +904,35 @@ module ZMQ {
...
+          throw e;
+        }
         ret = data;
       }
       return ret;
     }
 
+    pragma "no doc"
+    proc throw_socket_error(fmtstr: string) throws {
+      var _errno: c_int = errno;
```
mppf: I tend to prefer it when uses of errno are explicit and at the call site of the place producing the error. For that reason, I'd add errno as an argument to this function and have the callers pass it.

modules/packages/ZMQ.chpl
```diff
@@ -908,23 +904,35 @@ module ZMQ {
...
+                                                   (!isString(T))) {
       var ret: T;
       on classRef.home {
         var data: T;
-        for param i in 1..numFields(T) do
-          getFieldRef(data,i) = recv(getField(data,i).type);
+        try {
+          for param i in 1..numFields(T) do
+            getFieldRef(data,i) = recv(getField(data,i).type);
+        } catch e: TaskErrors {
```
mppf: Same question here.

modules/packages/ZMQ.chpl
psahabu: I must've had a stroke or something, I thought that was a `forall`. My bad.

modules/packages/ZMQ.chpl
```diff
@@ -772,7 +772,7 @@ module ZMQ {
         if (0 != zmq_msg_init_data(msg, copy.c_str():c_void_ptr,
                                    copy.length:size_t, c_ptrTo(free_helper),
                                    c_nil)) {
-          try throw_socket_error("Error in Socket.send(%s): %s\n");
+          try throw_socket_error(errno, "send");
```
mppf: 👍

bradcray:
If no tests were changed or added as part of this PR, are we really putting any more weight on error-handling?  What would it take to add such tests (or update existing ones to put that weight on them).

psahabu:
Answering both for here and #8244: Since all the relevant tests were written using prototype modules, the calls to newly throwing functions are now implicitly halting on errors. I agree this isn't the most weight we could be putting on error handling, since we could be using real handling instead of implicit halts.

I could wrap the test calls with `try` blocks and handle the errors with `catch` statements. But actually testing those code pathways might be tricky for these modules since that would require SystemErrors to be generated from error codes.

One other problem I've found while doing this is tracking what types of errors are thrown from each function. While adding compiler checks isn't the path we want to go down, a `throws` section in chpldoc would be helpful.

bradcray:
Oh, so the message printed by failing to try/catch in a prototype module for these errors is identical to that which was generated by the halt()s previously?  I would've guessed that there'd be minor changes in terms of wording, formatting, or line numbers which is why I was expecting changes to `.good` files.  If not, that's pretty encouraging.

I don't understand your second paragraph about challenges w.r.t. triggering error codes to create `SystemErrors`.

I completely agree that having a `throws` section in `chpldoc` would be great.  I'm not sure whether you're implying that `chpldoc` should automatically determine which error classes are thrown by a function (which seems much more aggressive than anything `chpldoc` does today and would require interprocedural analysis I believe, for the case when a function throws the result of another function upward?), or whether you just mean giving the chpldoc author for a given routine the ability to list the errors they expect it to throw manually.

psahabu:
The "error" tests we see in ZMQ are from compilerErrors, not `halt()`, which is why I didn't have to change them. ZMQ and Crypto tests never checked their `halt()` cases because they would only occur due to failures in the runtime, and those failures are difficult to set off purposefully from these libraries. That's what I was getting at in the second paragraph: even if non-halt handling were inserted, it's not clear how those code paths would be tested.

Sorry I was unclear on `chpldoc`, I just meant that the user should be able to document the errors thrown by their function, not that we should send `chpldoc` on a chase to find them.

bradcray:
OK, that makes it sound to me that there isn't really any significant additional weight put onto error-handling (in terms of nightly testing) as a result of these changes because nothing is actually exercising those error paths currently.  I guess it's arguably exercising "What happens if you invoke a throwing procedure from top-level code and no errors are thrown," but that doesn't feel like a very interesting case to me — or at least, one that I'm guessing we've covered well elsewhere.  In trying to put more weight on error handling, my hope was that we'd get more experience throwing and catching errors and seeing whether that reveals any new problems.  I'm concerned if we're not doing either of those, we're not gaining any greater confidence.

It does appear that there are a handful of tests that check for halts in the Crypto module (though only a few of them), so those might be another candidate to convert.

psahabu:
Unfortunately those tests are checking for halts in initializers. However, while implementing #8329, I found that lots of the standard modules have halts as well. Removing them and updating tests should give us good error handling coverage.

I'm going to explore that after I finish deprecating the 'out error' pattern.

mppf:
@psahabu - I think @bradcray is also hoping for more "real" application type tests that use error handling in a non-trivial way - i.e. that handle errors, don't just halt on them.

>  In trying to put more weight on error handling, my hope was that we'd get more experience throwing and catching errors and seeing whether that reveals any new problems

In other words while migrating halts over is useful, there's more to do here.

psahabu:
Right, my intent was to break some tests by removing `halt`s, then fix them with proper handling. Failing that, I'll make tests that use the proper handling.

PR #8246, opened 2018-01-18
Upper lower triangular matrix

aditmehta9:


ben-albrecht:
These procedures are already [implemented](http://0.0.0.0:8000/docs/latest/modules/packages/LinearAlgebra.html#LinearAlgebra.isTril).

PR #8247, opened 2018-01-18
Update isAbsPath test so that it will fail on platforms we don't support

ronawho:
Previously the isAbsPath test just checked fixed string literals. Update the
test to check if Python's abspath for the CWD is detected as an absolute path.
This will let us know if our abspath implementation is wrong on other platforms
that we may test on in the future. e.g. this test would (correctly) fail on
windows since we don't support windows style paths yet.

Related to https://github.com/chapel-lang/chapel/pull/8239

### DIFF
--- a/test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
+++ b/test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
@@ -1,4 +1,5 @@
    use Path;
+   use Spawn;
 
    writeln(isAbsPath("/foo/bar"));
 
@@ -13,3 +14,19 @@
    writeln(isAbsPath("\\server"));
 	 	 	 
    writeln(isAbsPath("."));
+
+   // Check to see if we correctly detect Pythons's abspath to the $CWD as an
+   // absolute path (This is to help identify platforms we don't support.)
+   writeln(isAbsPath(getPythonsAbspath()));
+
+   proc getPythonsAbspath() {
+     var command = "python -c 'import os; print(os.path.abspath(\"\"))'";
+     var sub = spawnshell(command, stdout=PIPE);
+
+     var absPath:string;
+     sub.stdout.readline(absPath);
+     absPath = absPath.strip();
+     sub.wait();
+     return absPath;
+     sub.wait();
+   }
--- a/test/library/standard/Path/victor-ludorum/testcases_isAbsPath.good
+++ b/test/library/standard/Path/victor-ludorum/testcases_isAbsPath.good
@@ -5,3 +5,4 @@ false
 true
 false
 false
+true


test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
```diff
@@ -13,3 +14,19 @@
...
+   proc getPythonsAbspath() {
+     var command = "python -c 'import os; print(os.path.abspath(\"\"))'";
+     var sub = spawnshell(command, stdout=PIPE);
+
+     var absPath:string;
+     sub.stdout.readline(absPath);
+     absPath = absPath.strip();
+     sub.wait();
+     return absPath;
+     sub.wait();
```
lydia-duncan: You've got a sub.wait() after the return statement, won't that never execute?

ronawho: whoopsie

test/library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
```diff
@@ -13,3 +14,19 @@
    writeln(isAbsPath("\\server"));
 	 	 	 
    writeln(isAbsPath("."));
+
+   // Check to see if we correctly detect Pythons's abspath to the $CWD as an
+   // absolute path (This is to help identify platforms we don't support.)
+   writeln(isAbsPath(getPythonsAbspath()));
+
+   proc getPythonsAbspath() {
+     var command = "python -c 'import os; print(os.path.abspath(\"\"))'";
```
lydia-duncan: Couldn't you just use [getcwd](https://docs.python.org/2/library/os.html#os.getcwd)?

ronawho:
If I remove the cygwin compiler error, this tests passes for me (my repo is in a posix-style dir: `/cygdrive/c/Users/eronagha/chapel`)

What was the case where cywin wasn't working?

I'm pretty sure our nightly cygwin testing uses a posix path too, so it's no longer clear to me what we don't support about cygwin -- can you use windows paths in cygwin (I honestly have no idea)

lydia-duncan:
Looking it up online: cygwin uses POSIX style paths for its operation, but because it lives in a Windows system, it could call a native Windows program and that would require a Windows-style path ([source](https://cygwin.com/cygwin-ug-net/using-effectively.html)).  Cygwin provides a program to perform this conversion, [cygpath](https://cygwin.com/cygwin-ug-net/cygpath.html).

I had incorrectly assumed that cygwin would use the Windows path style, and I should have verified that.  I'll check on whether python also labels full native Windows paths of the form "C:\\path\to\blah" as absolute on cygwin, or whether it only accepts POSIX style paths.  That information should allow us to make a better decision on whether we should view cygwin as supported by isAbsPath as it is currently implemented.

lydia-duncan:
Looks like python on cygwin does not view Windows-style absolute paths as absolute paths.  It seems reasonable for us to follow that behavior.

PR #8250, opened 2018-01-19
Fix LLVM performance regression for class pointers

dmk42:
Loads and stores cannot correctly handle `tbaa` access tags whose access type is a struct.  In #7873 , as additional metadata were added, I added a check to prevent struct `tbaa` access tags from being attached to loads and stores.

In #8194 , I began generating `tbaa.struct` nodes for class objects, which triggered this check and then prevented the metadata for class pointers from being attached to loads and stores.  This slowed down the stencil PRK by a factor of 3.

This change qualifies the check so that class pointers can get through, while still protecting against aggregate objects having their metadata attached to loads and stores.

The lesson here is that we were deriving tremendous benefit from scalar TBAA metadata on class pointers.

Passed full local testing with `CHPL_DEVELOPER` and `CHPL_LLVM_DEVELOPER` set, and `--llvm --fast`

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -454,13 +454,14 @@ llvm::StoreInst* codegenStoreLLVM(llvm::Value* val,
   GenInfo *info = gGenInfo;
   llvm::StoreInst* ret = info->irBuilder->CreateStore(val, ptr);
   llvm::MDNode* tbaa = NULL;
-  if( USE_TBAA && valType ) {
+  if (USE_TBAA && valType &&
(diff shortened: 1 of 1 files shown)


dmk42:
I am merging this now to clean up the performance noise.

@mppf - Would you please do a post-commit review?

@ronawho - This will undo the performance regression.

dmk42:
This is the long-term fix, not the temporary patch I was contemplating if I hadn't figured it out.

I have some ideas for improving TBAA further, but they are not related to this and are low priority.

PR #8253, opened 2018-01-19
Remove compiler error in isAbsPath for cygwin

lydia-duncan:
I hadn't realized that cygwin's path style was POSIX instead of Windows, which
means that the current implementation of isAbsPath would work for it and that
we don't have a CHPL_TARGET_PLATFORM setting where the function would *not*
work.  Remove the compiler error branch, since we don't have a way to detect
we're on a filesystem that this wouldn't work for.  Also updates the
documentation.

Remove the cygwin future, and the normal test's skipif.  Since cygwin actually
works, we don't need a separate future test.  We just need to know that we
don't support native Windows yet.

Passed testing over a full std/ paratest, and test/library/standard/Path and
test/library/standard/FileSystem on cygwin.


lydia-duncan:
@ronawho - mind taking a look?

Issue #8240, opened 2018-01-18
Internal compilation error RES0472 encountered for attempted param range class member declaration


ian-bertolacci:
### Summary of Problem
Attempting to create a param range in a class results in internal error RES0472.
This is in contrast to the more helpful `error: 'symbol' is not of a supported param type` when trying to do the same **outside** a class declaration.
(I am now aware that param ranges are not yet supported.)

### Steps to Reproduce

**Source Code:**
```chapel
class Foo {
  param _range_ = 1..1;
}

var f = new Foo();
writeln( f._range_ );
```

(see [this TIO instance](https://tio.run/##S85ILEjN@f8/OSexuFjBLT9foZpLQaEgsSgxVyG@KDEvPTVewVbBUE/P0JqrlourLLFIIQ0okJdaDlKsoWnNVV6UWZKak6ehkKYH06Bp/f8/AA))

### Configuration Information

- Output of `chpl --version`: 
```
> chpl --version
chpl Version 1.16.0.e43acc7
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)

```
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
> $CHPL_HOME/util/printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none *
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: none
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
```
>  gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

ian-bertolacci:
[...]

PR #8254, opened 2018-01-19
Convert an internal error about param formal default value into two user errors

daviditen:
A param formal with a non-param default value or a type-mismatched default
value resulted in an internal error, as reported in issue #8240. Change the
internal error into two user errors:

`error: default value for param formal is not a param`
or
`error: type mismatch between declared formal type and default value type`

Added tests to trigger each of these errors.

### DIFF
--- a/test/functions/diten/paramFormalGenericDefaultMismatchError.chpl
+++ b/test/functions/diten/paramFormalGenericDefaultMismatchError.chpl
@@ -1,0 +1,3 @@
+proc foo(param r: integral = 2.718) { }
+
+foo();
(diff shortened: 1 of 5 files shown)


test/functions/diten/paramFormalGenericDefaultMismatchError.chpl
```diff
@@ -0,0 +1,3 @@
+proc foo(param r: integral = 2.718) { }
```
lydia-duncan: Out of curiosity, why `integral` instead of `int`?

PR #8255, opened 2018-01-19
Improve hasCopyConstructor

mppf:
PR #5642 added hasCopyConstructor to detect the presence of a
copy-initializer. However the apprach used in that PR (namely to loop
over all of the FnSymbols) doesn't work in the event that the type in
question is a generic type and the specific copy initializer hasn't been
resolved yet.

To resolve that issue (and at the same time to remove the loop over all
FnSymbols), change hasCopyConstructor to try resolving an appropriate
copy init call.

While there, renamed hasCopyConstructor to hasCopyInit since it's
detecting a copy initializer (vs an old-style copy constructor).

This PR is intended to resolve an issue that came up when
converting SharedObject to use initializers (see PR #8170).

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -4963,7 +4963,7 @@ static void resolveInitField(CallExpr* call) {
 *                                                                             *
 ************************************** | *************************************/
 
-static bool hasCopyConstructor(AggregateType* ct);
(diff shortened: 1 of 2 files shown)


mppf:
@noakesmichael - this PR improves upon something you added in PR #5642 in order to resolve an issue @bradcray was seeing in PR #8170. Would you mind having a look at reviewing it?

PR #8256, opened 2018-01-19
Fix copy, deinit resolution for types with initializers

### LINKED ISSUE #8231: Local module `use` breaks visibility of copy initializers?
### Summary of Problem

It appears that a local `use` of a module that defines a copy initializer can break compilation
due to the copy initializer not being visible, even within that same scope.  I ran into this when
converting the `DateTime` module over to initializers (in PR #8141) which broke @benharsh's
elegant CoMD test due to its local `use` of DateTime.  Moving the `use` to module scope
fixed the issue and is a workaround we applied to make forward progress (but one that should
be reverted once this issue is fixed).

I reproduced this in a simpler standalone test in:
[`test/types/records/init/copy/localUseNoCopy.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/types/records/init/copy/localUseNoCopy.chpl
          ) #8141

With a minimal effort, I was not able to reproduce it in a standalone test that didn't rely
on the `DateTime` issue (but timed out before trying very hard).

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.17.0 pre-release (14176f2)`

mppf:
This PR makes multiple improvements aimed at improving resolution of
helper functions for types that have initializers. Helper functions
include `deinit` and `initCopy`.

Fixes #8231.

- [x] full local testing
- [x] primers with --verify, --baseline --verify
- [x] undo PR #8185 change to test/studies/parboil/BFS/bfs.chpl
- [x] undo PR #8141 change to test/studies/comd/elegant/arrayOfStructs/util/Util.chpl

### Warm up
 * Updated resolveUninsertedCall to enable it to return NULL instead
    of raising a fatal error. That's useful in fixupDefaultInitCopy, for example.

### Main change
 * Use resolveUninsertedCall to resolve deinit and copy-init for
    types with initializers
 * Unify calls to find the copy-init in fixupDefaultInitCopy and
    resolveInitVar. Now both of these call a shared function findCopyInit
    which is an evolution of findCopyInit from PR #8255.

### Important Enabling Change

Enable getInsertPointForTypeFunction to work with initializers

This function looks for the instantiation point to find where the
function such as auto-copy/auto-destroy should be added.

This logic originated in 3788ee3 and I
believe it would be better to store such functions within the
instantiated type, somehow.

Anyway, it was finding the instantiation point by looking at
at->defaultInitializer->instantiationPoint. But, types with initializers
don't have a defaultInitializer set. So, this commit extends it to look
for the instantiation point from the defaultTypeConstructor.

Additionally, any AggregateType that doesn't have a defaultInitializer or
a defaultTypeConstructor simply uses the defPoint of the type as the
place to add the new functions. That makes sense since the only
complicated thing here is handling generic instantiations.


mppf:
@benharsh - would you be up for reviewing this PR? thanks.

mppf:
Testing showed a memory leak in 4 Shared tests. I'll have a look.

mppf:
Passed full local testing.

PR #8259, opened 2018-01-19
Make parent init functions inapplicable for child init calls

benharsh:
This PR updates candidate filtering to reject an initializer function if the 'this' argument and corresponding actual do not have the same type. This prevents the compiler from choosing a parent class' initializer, which will cause odd execution-time behavior.

It turns out that a few internal/standard classes were relying on this behavior, and tests started to fail with compilation errors. We didn't see bugs on master because these classes did not contain fields that might be clobbered by the parent initializer. To solve the new compiler errors, I implemented the expected initializers by simply passing the arguments to the parent initializer.

The 'parent_inherits_initializer_bad' future can be retired as a result of this work.

Testing:
- [x] full local + futures


compiler/resolution/ResolutionCandidate.cpp
```diff
@@ -671,6 +671,18 @@ bool ResolutionCandidate::checkResolveFormalsWhereClauses() {
...
+      } else if (strcmp(fn->name,"init") == 0 &&
+                 formal->hasFlag(FLAG_ARG_THIS)) {
+        AggregateType* ft = toAggregateType(formal->getValType());
+        AggregateType* at = toAggregateType(actual->getValType());
+
+        // Do not allow dispatch for 'this' formal with initializers
+        //
+        // Types may only mismatch if the formal's type is an instantiation of
+        // the actual's type.
+        if (ft != at && ft->isInstantiatedFrom(at) == false) {
```
lydia-duncan: Note that for initializers on generic types, the "this" argument might itself be generic until we have resolved Phase 1 of that initializer - so you might need another check for when the actual passed is an instantiation of the formal.  E.g.
```chapel
var foo: Foo(int, real); // Here we obtain an instantiation for Foo

// But we have no idea if the initializers for Foo have been used yet,
// so they might still be generic
foo = new Foo(...);
```

lydia-duncan:
This seems nicely straight-forward.  I'd like to see a test using the case I mentioned, but other than that I think this is good to go!

compiler/resolution/ResolutionCandidate.cpp
benharsh: I think the function signature of an initializer is guaranteed to be resolved by this point (at least today it is).

benharsh:
Wouldn't the example above translate into something like:
```
init(&foo, int real);
init(&tmp, ...);
=(foo, tmp);
```

Would this result in a concrete 'this' formal?
```chpl
var foo : Foo(int, real) = new Foo(int, real, 100, 1.0);
```

compiler/resolution/ResolutionCandidate.cpp
lydia-duncan: Nah.  We mark the "this" argument with FLAG_DELAY_GENERIC_EXPANSION and instantiate everything else.  Then, when we go through resolution of the body, if we encounter the primitive PRIM_INIT_FIELDS, we get to [this point](https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/functionResolution.cpp#L4894) of [this function](https://github.com/chapel-lang/chapel/blob/master/compiler/resolution/functionResolution.cpp#L4843), where we instantiate the type

lydia-duncan:
Now I'm doubting myself, hmmm

benharsh:
I haven't had any luck trying to write a program that has a concrete type for the 'this' actual in the desired spot in ResolutionCandidate.

I poked around and saw ``resolveNewHandleInstantiatedGenericInit`` in ``functionResolution``, which appears to handle the case we're interested in adding a test for. If I'm reading things correctly, we currently and intentionally create the 'new_temp' this-actual with the fully generic type.

lydia-duncan:
Okay, then this is probably good to go!  Thanks, Ben!

PR #8263, opened 2018-01-21
Fix re2 builds on cygwin

ronawho:
New versions of cygwin require -std=gnu++11 to build re2, but it looks like re2
only throws that for CMake builds, not regular make builds. Explicitly throw
-std=gnu++11 ourselves under cygwin.

See https://github.com/google/re2/issues/150 for more information

### DIFF
--- a/third-party/re2/Makefile
+++ b/third-party/re2/Makefile
@@ -42,7 +42,12 @@ endif
 # CXX11_STD will be blank for compilers that default to C++11,
 # or it will use e.g. gnu++11 or c++11 if not.
 # Note that RE2 itself sometimes throws -std=c++11, but this is
-# not done for some compilers that Chapel supports.
+# not done for some compilers that Chapel supports. Cygwin
+# requires -std=gnu++11, but re2 will only throw that for CMake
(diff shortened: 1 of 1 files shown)


dmk42:
Hi.  CXX11_STD is already set in some compiler Makefiles.  Shouldn't we check if that is set before setting it here?  That would allow us to set it to something other than the GNU-compatible string for compilers that do something different.

ronawho:
>Hi. CXX11_STD is already set in some compiler Makefiles. Shouldn't we check if that is set before setting it here? That would allow us to set it to something other than the GNU-compatible string for compilers that do something different.

There is probably a cleaner solution, but note that this is just for cygwin, where I'm pretty sure gcc is the only choice. I think gcc 6.4 is the default on our current cygwin box, so `CXX11_STD` is empty, re2 is then throwing -std=c++11 in its Makefile. Even if gcc defaults to c++11 or newer, we need it to be gnu++11 (or newer) under cygwin, so I think we want to force this

dmk42:
Ah, got it.  Thanks.

PR #8267, opened 2018-01-22
Updating Datetime.chpl file comments section

victor-ludorum:
In the DateTime.chpl file there is **record time** and several methods associated with it . But in the comments section **"/* The hour represented by this date value */"** which should be **"/* The hour represented by this time value */"** . So, it should be updated .

@daviditen Would you mind to have a review .
You may close it if it is not useful .


daviditen:
@victor-ludorum Looks good, thanks!

PR #8273, opened 2018-01-22
The gmake clock-skew-warning filters should use a more portable regex

awallace-cray:
PR #8151 added filters to silently remove "clock skew detected" warning messages from gmake (if any), because those messages (when they occur) caused false test failures- both in Chapel's test harness, and in the shell script used to implement "make check". When the modified "make check" was run on omnios, the grep command failed because it did not accept "-E".

This change introduces a grep/regex that solves the omnios compatibility problem.

In the "make check" script (checkChplInstall) only, it also reverts some unnecessary complexity in PR #8151.  To understand this PR, it may be helpful to
>    git diff 7926fcce -- util/test/checkChplInstall

(ie, diff the version BEFORE PR #8151)

### DIFF
--- a/util/test/checkChplInstall
+++ b/util/test/checkChplInstall
@@ -162,24 +162,28 @@ COMP_FLAGS="--cc-warnings"
 
 # Compile test job into temporary directory
 log_info "Compiling \$CHPL_HOME/${REL_TEST_DIR}/${TEST_JOB}.chpl"
-${CHPL_BIN} ${TEST_DIR}/${TEST_JOB}.chpl ${COMP_FLAGS} -o ${TMP_TEST_DIR}/${TEST_JOB} > ${TMP_TEST_DIR}/chpl.out 2>&1
+TEST_COMP_OUT=$( ${CHPL_BIN} ${TEST_DIR}/${TEST_JOB}.chpl ${COMP_FLAGS} -o ${TMP_TEST_DIR}/${TEST_JOB} 2>&1 )
 COMPILE_STATUS=$?
 
-# Check that chpl compile was successful
+# Check that compile was successful
 log_debug "Compilation exit status was ${COMPILE_STATUS}"
 
     # exit status 0?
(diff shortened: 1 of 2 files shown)


awallace-cray:
Testing now, but I do not have access to a system where PR #8151 did not work.

util/test/checkChplInstall
```diff
@@ -177,9 +177,15 @@ if [ ${COMPILE_STATUS} -ne 0 ]; then
     exit 1
 fi
 
+
     # no errors on stdout/stderr?
     # first, apply "prediff"-like filter to remove gmake "clock skew detected" warnings, if any
-TEST_COMP_OUT=$( grep < ${TMP_TEST_DIR}/chpl.out -Ev '^g?make(\[[0-9]+\])?: *[Ww]arning: *(File .* has modification time .* in the future|Clock skew detected\. *Your build may be incomplete\.) *$' 2>&1 )
+TEST_COMP_OUT=$(<${TMP_TEST_DIR}/chpl.out)
+if [ -n "${TEST_COMP_OUT}" ]; then
+    TEST_COMP_OUT=$( grep < "${TMP_TEST_DIR}/chpl.out" -v \
```
ronawho: Should this be grepping `TEST_COMP_OUT` instead of `"${TMP_TEST_DIR}/chpl.out"`?

awallace-cray: No. Line 183 means it doesn't matter.

ronawho: Doesn't matter as in either would work? Seems like it'd be a little better/cleaner to use `TEST_COMP_OUT`  since that's the "is not empty" condition we're checking

ronawho:
I can test on omnios. 

I would update the PR message to say something like "The previous filter was using `-E` (a gnu grep extension), which isn't portable. Swap to a more portable version and only run the filter if there was any compiler output"

awallace-cray:
Rebased and modified the grep/regex. I'll test vagrant boxes later (except omnios) later.

PR #8274, opened 2018-01-22
Add tests to check LLVM alias analysis metadata

dmk42:
This change uses FileCheck to validate the TBAA type descriptors and access tags for scalars, records, class objects, and their fields.

- [x] Test on Mac with `CHPL_LLVM=system` and LLVM 5
- [x] Test on linux64 with `CHPL_LLVM=llvm`, and with `CHPL_DEVELOPER` and `CHPL_LLVM_DEVELOPER` set, and LLVM 4


dmk42:

dmk42:
Good point, thanks.  I've added scalar tests as well, and verified that they would catch the most recently fixed bug.

mppf:
great, thanks!

PR #8282, opened 2018-01-23
Add triple-quoted multi-line uninterpreted strings

daviditen:
Add a new style of strings that uses either ''' or """ as the delimiter.
These are uninterpreted and allow literal newlines.  Escape sequences such
as "\n" are treated as two separate characters, not a newline.

Add a test that demonstrates the behavior of these strings.

Update the string literals section of the spec to include uninterpreted string
literals.


compiler/parser/chapel.lex
```diff
@@ -468,6 +487,63 @@ static const char* eatStringLiteral(yyscan_t scanner, const char* startChar) {
...
+static const char* eatMultilineStringLiteral(yyscan_t scanner,
+                                             const char* startChar) {
+  YYLTYPE*   yyLloc  = yyget_lloc(scanner);
+  const char startCh = *startChar;
+  int startChCount = 0;
+  int        c       = 0;
+
+  newString();
+
+  while (((c = getNextYYChar(scanner)) != startCh || startChCount < 2) && c != 0) {
```
mppf: I think this would be clearer to write as say 'while true' with a break in the event we've just read 3 of the start character. I'm finding the body of this while conditional to be a bit more complex than I'd like.

compiler/parser/chapel.lex
```diff
@@ -468,6 +487,63 @@ static const char* eatStringLiteral(yyscan_t scanner, const char* startChar) {
...
+    if (c == startCh) {
+      startChCount++;
+    } else {
+      startChCount = 0;
+    }
+
+    if (c == '\"') {
```
mppf: I think 'addCharEscape' is unfortunately named. I'd prefer to rename it to 'addCharEscapeNonprint'. It's debateable whether or not much of the special handling below should go into a method named something like addCharEscape. But given that all of the strings represented inside the compiler are treated as C `"bla bla"` type strings, including appropriate backslashes, I think it'd be worthwhile to add a function for properly adding a character with appropriate escaping to such a string.

mppf: But anyway if we keep the current structure for these cases, we know the character we're adding is printable, so shouldn't we just call addChar with it?

compiler/parser/chapel.lex
```diff
@@ -468,6 +487,63 @@ static const char* eatStringLiteral(yyscan_t scanner, const char* startChar) {
...
+      addCharEscape('\\');
+      addCharEscape('n');
+    } else if (c == '\t') {
+      // translate tab into two characters "\t"
+      addCharEscape('\\');
+      addCharEscape('t');
+    } else if (c == '\\') {
+      // translate backslash into an escaped double-backslash
+      addCharEscape('\\');
+      addCharEscape('\\');
```
mppf: If this is a list of escape sequences in C, it's incomplete. See https://en.wikipedia.org/wiki/Escape_sequences_in_C

it should also handle `\a` `\b` `\f` `\r` `\v`, shouldn't it?

Alternatively, maybe you can explain why we don't need to handle those? I could believe an argument that only `\` `"` and `\n` need to be handled - but then I don't see why you are handling tab.

mppf: I'd prefer whatever logic is necessary here be moved to a new function named something like addCharEscape.

daviditen:
@mppf Thanks for reviewing!  I just pushed some commits to address your comments.  How does it look now?

compiler/parser/chapel.lex
```diff
@@ -884,6 +948,55 @@ static void addCharEscape(char c) {
   }
 }
 
+// Convert C escape characters into two characters '\\' and the other character
```
mppf: I'd appreciate a colon after 'two characters' above

compiler/parser/chapel.lex
```diff
@@ -425,36 +445,36 @@ static const char* eatStringLiteral(yyscan_t scanner, const char* startChar) {
...
+        addCharEscapeNonprint('t'); // add a valid escape to continue parsing
       } else if ('0' <= c && c <= '7' ) {
         ParserContext context(scanner);
         yyerror(yyLloc, &context, "octal escape not supported in string literal");
-        addCharEscape('t'); // add a valid escape to continue parsing
+        addCharEscapeNonprint('t'); // add a valid escape to continue parsing
       } else if (c == 0) {
         // we've reached EOF
-        addCharEscape('t'); // add a valid escape to continue parsing
+        addCharEscapeNonprint('t'); // add a valid escape to continue parsing
```
mppf: I think any of these with a known character should just use addChar, but that's not strictly required for your change.

compiler/parser/chapel.lex
```diff
@@ -468,6 +488,50 @@ static const char* eatStringLiteral(yyscan_t scanner, const char* startChar) {
   return astr(stringBuffer);
 }
 
+static const char* eatMultilineStringLiteral(yyscan_t scanner,
+                                             const char* startChar) {
+  YYLTYPE*   yyLloc  = yyget_lloc(scanner);
+  const char startCh = *startChar;
+  int startChCount = 0;
+  int        c       = 0;
```
mppf: spacing seems really bizarre here

daviditen: Yes it does.  I copied it from eatStringLiteral, but then added startChCount and didn't follow the pattern.  I'll fix up the spacing.

compiler/parser/chapel.lex
daviditen: Colon added.

With the name `addUnescapedChar` I was trying to convey that it takes a possibly escaped character and turns it into something without escape characters.  "Unescaped" was meant to imply the breaking up of escape characters.  I can use your suggested name though.

daviditen:
Thanks @mppf!

PR #2537, opened 2015-09-15
Quiet parallel/forall/vass/memleaks* regression

ronawho:
I had to run ~1000 times to see the regression so it appears to be very
sporadic and we were unlucky enough for it to occur on the first night of
testing.

This adds a serial to the forall statements to quiet the regressions. Serial
was attached to the forall in memleaks2-BlockDist, but I thought I could remove
it in 880d1520e57c8b9979d4ea9fce47e48dc520f22e. I'm re-adding that and adding
it to memleaks1-minimal as well.

See JIRA issue 112: https://chapel.atlassian.net/browse/CHAPEL-112

PR #8287, opened 2018-01-23
Stop serializing foralls in parallel/forall/vass/memleaks*

ronawho:
This reverts #2537. That PR quieted some memleaks by wrapping a forall with a
serial stmt. The root cause was never understood, but the sporadic leaks seem
to be gone now, so remove the serial stmts and references to the JIRA issue.

Passes 100,000 trials for test/parallel/forall/vass/memleaks*.chpl with the
quickstart configuration for --local and --no-local.


ronawho:
@vasslitvinov any objections?

PR #8292, opened 2018-01-24
 Catch more use-before-init phase 1 errors

benharsh:
Before this PR users could call methods on fields during phase 1, and the
compiler would fail to complain. For example:

```chpl
// 'x' is a field
proc init() {
  x[1] = 1;
  x.foo = 1;
  x.m();
  super.init();
}
```

The three statements preceeding the 'super.init' should be flagged as
'use before init' errors at compile-time. The second statement was in
fact identified as an error, but the message was confusing (can't pass
"this" as an actual to a function...).

Three changes are made to InitNormalize::fieldUsedBeforeInitialized:

1) "x[1] = 1": If the LHS of an assignment looks like a field access and the
   call is flagged with the 'square' tag, then we are dealing with a field use.

2) "x.foo = 1": If the LHS of an assignment is not a simple field access, then
   we need to walk through it to determine if it contains field uses.

3) "x.m()": When examining an arbitrary CallExpr that is not an assignment
   or field access, we need to examine its baseExpr as well as its actuals.
   In this case, the baseExpr is a field access like "this.x".

These cases now result in an error message like:
  Field used before it is initialized

Testing:
- [x] full local + futures

### DIFF
--- a/compiler/passes/InitNormalize.cpp
+++ b/compiler/passes/InitNormalize.cpp
@@ -1419,16 +1419,32 @@ bool InitNormalize::fieldUsedBeforeInitialized(CallExpr* callExpr) const {
   bool retval = false;
 
   if (isAssignment(callExpr) == true) {
-    retval = fieldUsedBeforeInitialized(callExpr->get(2));
+    if (CallExpr* LHS = toCallExpr(callExpr->get(1))) {
+      if (isFieldAccess(LHS)) {
+        // Want to watch out for array-like accesses that appear as field
+        // accesses: x[1] = 1;
+        retval = LHS->square;
+      } else {
+        // Look for expressions like: x.foo = 1;
+        retval = fieldUsedBeforeInitialized(callExpr->get(1));
+      }
+    }
+    retval = retval || fieldUsedBeforeInitialized(callExpr->get(2));
 
   } else if (DefExpr* field = type()->toLocalField(callExpr)) {
     retval = isFieldInitialized(field) == true ? false : true;
 
   } else {
-    for_actuals(actual, callExpr) {
-      if (fieldUsedBeforeInitialized(actual) == true) {
-        retval = true;
-        break;
+    // Need to check the baseExpr in cases like:
+    //   myField.set(1)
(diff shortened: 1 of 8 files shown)


bradcray:
Cool!  This fixes a surprise that I ran into last week and had been intending to file a future + issue against:

```chapel
record R {
  var x: atomic int;
  
  proc init() {
    this.x.write(1);
    super.init();
    writeln("x init'd to ", x.read());
  }
}

var myR = new R();
var myR2: R;

writeln(myR);
writeln(myR2);
```

Previously, this had the surprising effect of making the atomic field '0' (presumably because the compiler would default init it after the .write() to it).  Do you think adding this as a test to help lock in the behavior would be beneficial?

Also, in the error message:

```
testit2.chpl:5: error: Field used before it is initialized
```

would it be easy to name the field that's causing the problem?  (seems like it should be...).  For a complex statement/expression that potentially refers to many fields, I could imagine this being helpful.  I.e.,

```
testit2.chpl:5: error: Field 'x'used before it is initialized
```

benharsh:
There's a related future that was modified by this PR related to atomics:

https://github.com/benharsh/chapel/blob/d1b2c5003410c1413adfe4f30f396ec9b1f9f241/test/classes/initializers/generics/initAtomicsViaWrite.chpl

Is that sufficient?

I'm not sure what it would take to name the offending field. I'll open an issue if I can't find one.

bradcray:
Oh OK, yes, thanks for pointing it out to me (whoever filed it is clearly a person of taste and distinction! :) ).

It appears to me that this future could be retired now.  I.e., it suggests in the text itself that it didn't expect it to work per se, just didn't expect an internal error.  Even though the .good file suggests that maybe it should work as written, I don't think I agree with that anymore.  I think the path forward here is to support initialization via `=` for atomics (in initializers and in normal variable initialization), if not assignment, and then to initialize them in that manner.  Let me know if you disagree (I'd be happy to retire it if you don't).

benharsh:
Retiring it sounds reasonable to me, go for it!

I have also opened #8295 to track the error message feature request.

bradcray:
Thanks!

PR #8293, opened 2018-01-24
Fix LLVM backend field access type and supertype TBAA handling

dmk42:
In the translation to LLVM, we were sometimes passing an incorrect Chapel type to the low-level version of the `codegenLoadLLVM()` and `codegenStoreLLVM()` functions.  Nothing was wrong with the LLVM type, only the Chapel type, which is why it only became apparent while I was working on TBAA.  This change fixes that problem.  As a result, it is no longer necessary to carry around a `fieldTbaaTypeDescriptor` in a `GenRet`.

While there, I noticed that I was generating TBAA information the same for superclasses as for other classes, which meant that they were erroneously marked as pointers in the TBAA metadata.  This change fixes that.

Expected test results have been updated accordingly.

Test plan:
- [x] Mac on release/examples with `CHPL_LLVM=system`, `--llvm --fast`, and LLVM 5
- [x] Full local testing on linux64 with `CHPL_LLVM=llvm`, `--llvm --fast`, and LLVM 4, with `CHPL_DEVELOPER` and `CHPL_LLVM_DEVELOPER` set

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -440,7 +440,7 @@ void codegenInvariantStart(llvm::Value *val, llvm::Value *addr)
 }
 
 // Create an LLVM store instruction possibly adding
-// appropriate metadata based upon the Chapel type of val.
+// appropriate metadata based upon the Chapel type destType.
 //
 static
 llvm::StoreInst* codegenStoreLLVM(llvm::Value* val,
@@ -483,6 +483,8 @@ llvm::StoreInst* codegenStoreLLVM(llvm::Value* val,
 
 
 
+// The valType argument is a hint, in case we don't already know the
+// Chapel type of the memory into which we are storing.
 static
 llvm::StoreInst* codegenStoreLLVM(GenRet val,
                                   GenRet ptr,
@@ -550,6 +552,8 @@ llvm::LoadInst* codegenLoadLLVM(llvm::Value* ptr,
   return ret;
 }
 
+// The valType argument is a hint, in case we don't already know the
+// Chapel type of the memory from which we are loading.
 static
 llvm::LoadInst* codegenLoadLLVM(GenRet ptr,
                                 Type* valType = NULL,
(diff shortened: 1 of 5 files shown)


dmk42:

compiler/codegen/expr.cpp
```diff
@@ -555,13 +555,16 @@ llvm::LoadInst* codegenLoadLLVM(GenRet ptr,
...
 {
-  if( ptr.chplType && !valType ) {
-    if( ptr.isLVPtr ) valType = ptr.chplType;
-    else valType = ptr.chplType->getValType();
+  Type *srcType;
+  if (ptr.chplType) {
+    if (ptr.isLVPtr) srcType = ptr.chplType;
+    else srcType = ptr.chplType->getValType();
+  } else {
+    srcType = valType;
```
mppf: So doesn't this have the effect of ignoring valType that was passed in the event that ptr.chplType is set?
I would think calls to codegenLoad/StoreLLVM that explicitly pass valType should respect that. Would it be hard to adjust the compiler to have an INT_ASSERT in the event that both are non-null and they differ, and adjust the code calling here to either not pass valType at all or pass a consistent one?

dmk42: It turns out that valType is not what we want, unless ptr.chplType is unset.  The Chapel type being dereferenced is what matters.  We only need help from valType when we don't know what that is.

I have tried exactly that INT_ASSERT and it triggers on hello.chpl while processing the Error hierarchy, among many other places.  Also, if we give valType precedence, the LLVM verifier fails because the access type is not one of the fields in the surrounding struct.

mppf: Can you comment the appropriate places passing that type to indicate that it's only used as a last-resort, then?

dmk42: An example is the call to `codegenLoadLLVM()` from  `codegenCallExpr()` where, when the type of the return object is already known, valType gets it out of sync with what is actually stored there.  If we don't know what is already stored there, it is safe to use the valType passed because that is the return type.

dmk42: OK, I'll add a comment in the code.  There are only two spots.

dmk42: I ended up adding comments in four places.  You can search for the word "hint".

dmk42:
Thanks to you too!  I agree those comments helped.

dmk42:
@ronawho - This just-merged PR causes fewer pointers to be aliased than before.  Based on recent experience, I don't expect a big impact, but there might be a slight performance improvement in some cases.

PR #8297, opened 2018-01-24
Do nonblocking responses in the polling task

gbtitus:
Every time the polling task handles a request it has to respond to the
initiator.  It may need to release the remote-fork request buffer the
initiator used and/or it may need to send a "done" indication to the
initiator.  In the past the polling task has used synchronous PUTs for
these, that is, it did a PUT and then waited for the network response
before continuing.  That has the advantage that if an error occurs
either in the network or at the remote node, the polling thread will
kill the job before starting any other work.  However, it also reduces
the polling thread's throughput because at least one network round-
trip latency must pass between when the polling task finishes handling
one request and starts handling the next.

Here, switch to handling request reponse completions asynchronously.
When it sends a request response the polling task no longer waits for
the corresponding CQ event.  A periodic check for completion events is
done in the main polling loop alongside looking for requests, and such
events are consumed and checked then.  An error in the network or at the
remote node will still kill the job, but the polling thread may have
processed a few additional requests in the meantime.  This doesn't seem
like a problem since such network or remote errors are always fatal; the
former must be due to hardware problems and the latter due to internal
errors in the comm layer itself.

This results in some significant performance improvements. For 16-node-xc,
on-stmts are about 1.5x faster and "fast" on-stmts are about 2.5x faster.
Some representative test results are:
- 1.5x improvement for ra-on
- 15% improvement for SSCA2
- 10% improvement for lulesh-dense

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1317,6 +1317,7 @@ static void      fork_amo_wrapper(fork_amo_info_t*);
 static void      release_req_buf(uint32_t, int, int);
 static void      indicate_done(fork_base_info_t* b);
 static void      indicate_done2(int, rf_done_t *);
+static void      send_polling_response(void*, c_nodeid_t, void*, size_t);
 static nb_desc_idx_t nb_desc_idx_encode(int, int);
 static void      nb_desc_idx_decode(int*, int*, nb_desc_idx_t);
 static nb_desc_t* nb_desc_idx_2_ptr(nb_desc_idx_t);
@@ -1366,7 +1367,8 @@ static void      acquire_comm_dom_and_req_buf(c_nodeid_t, int*);
 static void      release_comm_dom(void);
 static chpl_bool reacquire_comm_dom(int);
 static int       post_fma(c_nodeid_t, gni_post_descriptor_t*);
-static void      post_fma_and_wait(c_nodeid_t, gni_post_descriptor_t*, chpl_bool);
+static void      post_fma_and_wait(c_nodeid_t, gni_post_descriptor_t*,
+                                   chpl_bool);
 static int       post_fma_ct(c_nodeid_t*, gni_post_descriptor_t*);
 static void      post_fma_ct_and_wait(c_nodeid_t*, gni_post_descriptor_t*);
 static void      local_yield(void);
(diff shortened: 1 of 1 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -3859,16 +3875,94 @@ inline
...
+  }
+
+  //
+  // Fill in the POST descriptor.
+  //
+  post_desc->type            = GNI_POST_FMA_PUT;
+  post_desc->cq_mode         = GNI_CQMODE_GLOBAL_EVENT;
+  post_desc->dlvr_mode       = GNI_DLVMODE_PERFORMANCE;
+  post_desc->rdma_mode       = 0;
+  post_desc->src_cq_hndl     = 0;
```
cassella: I'm thinking/guessing/assuming that these first five fields aren't modified by ``post_fma()`` or anything else?  Would there be an advantage to getting them all set once up front, and not setting them again each time through?

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -3859,16 +3875,94 @@ inline
...
+    // for request responses (buffer releases and "done" indicators).
+    // To improve throughput, we don't wait for the remote completions
+    // when we send these.  Instead, we collect the CQ completion events
+    // later, as they come in.  We can have as many request responses in
+    // flight simultaneously as we have Post descriptors in the pool.
+    // Small-scale runs on Aries-based systems show the polling task can
+    // do about 3 minimum-cost requests ("fast" remote forks, etc.) in
+    // the minimum network latency of ~1 usec.  The pool size we use
+    // here is enough larger that it will cover the max network latency
+    // of ~1.5 usec.
```
cassella: OOC, is 1.5 usec the max theoretical latency, or the max observed in the small-scale testing?

gbtitus: See https://www.cray.com/sites/default/files/resources/CrayXCNetwork.pdf (public info), page 22.  ~1.5 usec is the measured maximum Aries network latency for 8-byte PUTs.

runtime/src/comm/ugni/comm-ugni.c
gbtitus: The documentation doesn't say whether uGNI might modify these or not, so I don't think we can rely on it not doing so.  I doubt any benefit would be noticeable in practice anyway, since the total number of stores needed to initiate a network transaction is large and some of them are expensive (the ones to the memory-mapped FMA descriptor and window).  And the code is a bit more clear with these explicit at the point of the post.

PR #8304, opened 2018-01-25
Move Buffers from modules/standard/ to packages/

bradcray:
In PR #8224, I became concerned about the lack of testing currently
done on the `Buffers` module (in spite of having some fairly complex
autocopy/init / copy initializer logic, that logic seems to be largely
untested), so proposed that it be moved from modules/standard to
modules/packages in that PR.  This effort tries to make good on that
proposal.


bradcray:
@mppf this is the PR for the move of `Buffers.chpl` that we were discussing on PR #8224.  One thing that gives me pause is that grepping for references to `Buffers`, I"m seemingly finding documentation on `Buffers` and `Bytes` in `IO.chpl` (though IO.chpl doesn't seem to `use Buffers`).  Does that belong there?

@ben-albrecht: I think you have the most experience with what's required to move modules between standard/ and packages/.  Would you let me know if I've missed anything obvious?

mppf:
@bradcray - yes let's move the "Bytes Type" and "Buffers" sections from the IO.chpl module chpldoc to this module.

bradcray:
@mppf: Like this?  (Note that there was already documentation on those types in Buffers.chpl itself, so I took a best stab at merging them.  I also removed a reference that channels contain buffers because it seemed out of place here and also because I didn't feel confident that they did in a quick search.

mppf:
@bradcray - yes I am happy with the merged documentation, thanks. I hope one day we can revisit & enable the integration I designed for between channels and buffer, but this PR leaves it in a better intermediate state in the meantime.

bradcray:
@mppf: Would you file an issue against that vision, just as a placeholder?  Doesn't need to be super-detailed in my opinion, just enough to capture the intent.

mppf:
Done, issue #8314

PR #8308, opened 2018-01-25
Update the comments of replace function in string file

victor-ludorum:
The comments of **replace function** of the string should be updated:  _here a copy of the string where `needle` replaces `replacement` up_ this should be actually replaced with _a copy of the string where `replacement` replaces `needle` up ._ .


victor-ludorum:
Thanks @ben-albrecht !!

PR #8311, opened 2018-01-25
Use up to 120 comm domains on Aries-based systems

### LINKED ISSUE #8215: Don't limit to 30 comm domains on Aries-based systems.
Legacy code in the comm=ugni implementation limits us to 30 comm domains. While that's appropriate for Gemini-based systems which only have 32 NIC tiles, it's not for Aries-based systems which have 128 NIC tiles.  (The actual maximum number one can use is actually a few less than this in both cases.)

We should change the logic to allow using more comm domains on Aries-based systems.  The default number the comm layer uses, subject to limiting by the maximum, is 1 more than the number of threads the tasking layer says it will use, so on BW36 compute nodes we're actually overloading 36 threads on 30 comm domains, which definitely won't be optimal.  In a prototype implementation @ronawho found that upping the limit for Aries improved ra-atomic by 25% when running on BW36 compute nodes, but hurt ISx performance.  (ISx seems to be our most temperamental benchmark at present.)

gbtitus:
We have long limited the number of uGNI comm domains we use to 30.  This
number was chosen for the original comm-ugni implementation, and was
appropriate at that time since the Gemini chip only has 32 NIC tiles.
We didn't update it when Aries arrived, although that chip has 128 NIC
tiles.  For a long time this was not a problem, since we didn't have
compute nodes with enough processors to use more than 30 comm domains.
(Generally speaking one only needs as many comm domains as one has CPUs
to initiate transactions on them.)  But with the advent of 36-core
compute nodes in Aries-based systems that's no longer true.

Here, up the limit on uGNI comm domains to 120 for Aries-based systems.
In order to make this work with the remote fork request indexes sent via
the remote CQ mechanisms, I also switched to using the 32-bit remote
event data for those encoded indexes instead of the 24-bit local event
data.  (We actually should have been doing this all along.)  And while
here, I recoded the event data (AKA instance ID) encoding and decoding
macros to make things a bit more clear.

This resolves #8215.

This proves ra-atomic performance on BW36 XC nodes by 20-25%.  It
also slows ISx even further, and we will address that separately.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -371,30 +371,39 @@ static gni_nic_device_t nic_type;
 // The uGNI instance IDs we use are built from locale, communication
 // domain, and request buffer indices, plus remote operation codes.
 //
-#define _IID_LI_BITS  18
-#define _IID_LI_MASK  ((1 << _IID_LI_BITS) - 1)
-#define _IID_LI_POS   0
+#define _IID_LI_BITS        17
+#define _IID_LI_MASK        ((1 << _IID_LI_BITS) - 1)
+#define _IID_LI_POS         0
+#define _IID_ENCODE_LI(li)  (((li) & _IID_LI_MASK) << _IID_LI_POS)
+#define _IID_DECODE_LI(iid) (((iid) >> _IID_LI_POS) & _IID_LI_MASK)
 
-#define _IID_CDI_BITS 5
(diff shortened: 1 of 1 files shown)


ronawho:
Code looks good, we may want to update the PR message a little as the perf results come in.

gbtitus:
XC functional testing passed.  XE functional testing passed (hello, and then ra, ra-atomics, stream, ISx, SSCA2 all at performance scale).

PR #8312, opened 2018-01-26
Revert portion of #8293 relating to field access type

dmk42:
Testing of #8293 was incomplete because I forgot that `start_test` unsets `CHPL_DEVELOPER`.  When a Chapel program is compiled with `CHPL_DEVELOPER` and `CHPL_LLVM_DEVELOPER` set and `--llvm` (outside of the test harness), the LLVM verifier fails due to invalid metadata.  Testing only succeeds because `start_test` turns off the verifier.

This change is an exact reversion of expr.cpp and genret.h from #8293 , which is the portion related to field access types.  The other files, related to the handling of superclasses, are not reverted.  In hindsight, I would have done better to split the two purposes into two separate PRs.

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -440,28 +440,29 @@ void codegenInvariantStart(llvm::Value *val, llvm::Value *addr)
 }
 
 // Create an LLVM store instruction possibly adding
-// appropriate metadata based upon the Chapel type destType.
+// appropriate metadata based upon the Chapel type of val.
 //
 static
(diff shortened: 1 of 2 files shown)


dmk42:
@mppf - Sorry for the confusion. Due to testing problems as noted in this PR, I was not testing #8293 as well as I had thought.  Would you mind reviewing this reversion?

I'm not reverting the part related to superclasses because that works.  I've tested this PR with a hacked version of `start_test` that doesn't turn off `CHPL_DEVELOPER`.

Issue #8324, opened 2018-01-29
Implementing gcd function in Math.chpl


victor-ludorum:
As many languages have **gcd function** to calculate greatest common divisor , so I think this function should be implemented in Chapel too.
Please suggest !!

mppf:
I think it'd be reasonable to add `proc gcd(arg1: int, arg2: int) : int ` to Math.chpl.

nimitbhardwaj:
Hey, isn't it already in BigInteger, if its still required, then why not add the other number theory functions in Math.chpl.
But they are already there in BigInteger, just wrappers over GMP, as I saw code there, I also thought there would be need of em, but then I got that they are implemented.

One thing not there is nCr, nPr etc which are necessary functions for Combinatorics. Catalan's number etc, I thought they would be there, but I feel they are sometimes necessary.

mppf:
It'd be reasonable to create a new package if we want to add a lot of combinatiorics functions.

nimitbhardwaj:
Yeah, a new package is better, because Math is something different, number theory is different, even number theory is part of math, may be not explaining correctly, in computer science we study the algorithms regarding the number theory, like implementing these functions, sieve, prime numbers etc other things

victor-ludorum:
@nimitbhardwaj I have also checked those modules . But those are related with BigInteger and gcd is common function which should be present in standard math module . Therefore, I have thought to implement in Chapel module. 

> One thing not there is nCr, nPr etc which are necessary functions for Combinatorics. Catalan's number etc, I thought they would be there, but I feel they are sometimes necessary.

Actually problem with nCr and nPr function is that they may give very large values to some small input due to factorial which is difficult to be stored in any datatype . I think this may be reason that C++ or python has no inbuilt function for nCr or nPr .

nimitbhardwaj:
Cool bro

mppf:
https://chapel-lang.org/docs/master/modules/standard/BigInteger.html#BigInteger.bigint.bin

> One thing not there is nCr, nPr etc which are necessary functions for Combinatorics. Catalan's number etc, I thought they would be there, but I feel they are sometimes necessary.

The BigInteger module does include the way to compute the binomial coefficient (n choose k) https://chapel-lang.org/docs/master/modules/standard/BigInteger.html#BigInteger.bigint.bin

I'm not familiar with all of the functions you're talking about but it seems to me that most of these do not operate simply on integers (and would require bigint, say).

[...]

PR #8325, opened 2018-01-29
Creating GCD function in Math.chpl file 

victor-ludorum:
As **gcd** is common function which is  used frequently so, this pull request deals with the implementation of gcd function in Chapel math module referencing issue #8324  .
@mppf  Kindly review it and suggest what necessary improvements should be done . 

- [x] full local testing


victor-ludorum:
@nimitbhardwaj Thanks 😊 !! for the review . 
> gcd procedure can be made fast by writing it in iterative form, rather than recursive. 

If you are thinking that I should impose [Binary GCD](https://en.wikipedia.org/wiki/Binary_GCD_algorithm) Then , I can't implement it here as it computes **gcd** **of two non-negative integers** but I have to implement it for negative number too . Or if you are suggesting to implement this recursive version into iterative then, I have researched on this and I found that both work in the same time and mostly all uses this recursive version of Euclid theorem and it is faster too .

> Then if gcd is calculated, lcm function should also be there, then there comes gcd of array of numbers, lcm of array of numbers, modular exponentiation, modular inverse etc other terms of number theory.

Actually , its all depend upon need what is necessary for the chapel module , if those are necessary I will implement that too .

nimitbhardwaj:
GCD for negatives is not a difficult, remember my friend GCD(a, -b) = GCD(-a, b) = GCD(-a, -b) = GCD(abs(a), abs(b)), not too difficult

victor-ludorum:
gcd(4,-8) = 4, gcd(-4,-8) = -4 and gcd(-4,8) = -4 so it is different .

nimitbhardwaj:
Here is iterative version, its easy
```chapel
proc gcd(in a: int, in b: int): int {
  a = abs(a);
  b = abs(b);
  while(b%a != 0) {
    var c = b%a;
    b = a;
    a = c;
  }
  return a;
}
```

nimitbhardwaj:
Bro, talking of negative number GCD check [here](https://math.stackexchange.com/questions/927050/can-we-find-the-gcd-of-a-positive-and-negative-number)

nimitbhardwaj:
I can also prove it with some easy trivial informal way, what is GCD of 2 numbers, its the greatest number that divides both the two numbers, so if we have -12 and 18, here 6 divides both of em, so it is enough to say 6 is their GCD.

nimitbhardwaj:
the link at proof wiki also says it all

victor-ludorum:
Yeah.. I know this iterative version and it is easy too.. But as you have suggested for time complexity . I didn't find any article on internet suggesting iterative version is faster than recursive version . Therefore, I have implemented in recursive way. But I will check again if it is really faster than I will implement in iterative way.

nimitbhardwaj:
Nah, no difference in complexity, I was just thinking of complexity of function overhead, so, its not a big deal of complexity, and also the range of int is 2^63 - 1, at max, so, 10^19, not a big deal for O(logn)

victor-ludorum:
Actually I have thought you are suggesting about Binary GCD algo for computing therefore , I have told you that it is only for non negative number . And the iterative approach can work on negative numbers which you have suggested .

mppf:
> I was just thinking of complexity of function overhead

Right, iterative versions are generally preferred (if they're not much harder to write) because they use less stack space. If optimizations like tail recursion don't apply, a recursive version that is O(n) might use O(n) stack frames. Whether or not the iterative version is really "better" or not probably depends more upon if it's easy to understand.

victor-ludorum:
Okay @mppf  sir !! I will update the code if it has good space complexity .

victor-ludorum:
@nimitbhardwaj Thankyou very much for review , I have learned many new things with the discussion 😊

nimitbhardwaj:
But I didn't understand, why worry for negatives, GCD(-a, b) = GCD(abs(a), abs(b)), this is true in mathematics, no matter what algorithm you use. Mathematics can never be failed, so Binary GCD will work with abs(a) and abs(b) correctly, the GCD is always a +ve quantity, as I read in my math class.

nimitbhardwaj:
@victor-ludorum , thanks you too, I also learned Bin GCD, I didn't know of it, its 60% better.

victor-ludorum:
Yeah !! I too have thought to implement Binary GCD but as per the research done by me it is not much popular and little bit complex too for implementation . Therefore, I don't have thought to implement as I want to create a simple function .

victor-ludorum:
@mppf I have updated the code and made it in a **iterative manner** handling all cases . Please review the code .

victor-ludorum:
@mppf Sir, I have updated the code , actually previously I get some result which shows that gcd may be negative . But I have confirmed that it will always be positive . Sorry, for confusion .

test/library/standard/Math/victor-ludorum/testcases_gcd.chpl
```diff
@@ -0,0 +1,17 @@
+writeln(gcd(4,8));
```
mppf: Can you put this test in  test/library/standard/Math ? Adding user-names in the test directory is a pattern we are trying to move away from.

modules/standard/Math.chpl
```diff
@@ -1056,6 +1056,19 @@ module Math {
     extern proc truncf(x: real(32)): real(32);
     return truncf(x);
   }
+  
+  /* Returns the greatest common divisor of the integer argument `arg1` and
+     `arg2`. */
+  proc gcd(in a: int,in b: int): int {
+     var r: int;
+     while(b != 0) {
+       r = a % b;
```
mppf: I have some discomfort with `%` applied to negative numbers. Would you mind taking abs of a and b before the loop, instead of taking the abs when returning?

test/library/standard/Math/victor-ludorum/testcases_gcd.chpl
victor-ludorum: Okay sir !! I will do that . Actually previously this idea is preferred . Therefore, I have followed this

modules/standard/Math.chpl
victor-ludorum: Okay sir !! I will update this though it do not have much affect . But I think your suggestion is good . I will follow it. Thanks !! 😊

victor-ludorum:
@mppf Sir, I have made the changes suggested by you . Please review it !!

victor-ludorum:
I have updated the code so that if there will be any negative number then, it will first convert it into positive number . Then all the operation will be done on them so that there will be no problem with any operators too.

modules/standard/Math.chpl
```diff
@@ -1056,6 +1056,25 @@ module Math {
...
+  
+  /* Returns the greatest common divisor of the integer argument `arg1` and
+     `arg2`. */
+  proc gcd(in a: int,in b: int): int {
+     if(a < 0) {
+     a = -a;
+     }
+     if(b < 0) {
+     b = -b;
+     }
```
mppf: I'd expect
```
a = abs(a);
b = abs(b);
```

Anyway, if you keep the conditionals here, it's important to indent their bodies. Besides the parens are not necessary around a conditional in Chapel. I'd probably write such a conditional like this:
``` chapel
if a < 0 then
  a = -a;
```

modules/standard/Math.chpl
```diff
@@ -1056,6 +1056,25 @@ module Math {
...
+     var r: int;
+     while(b != 0) {
+       r = a % b;
+       a = b;
+       b = r;
+     }
+    return abs(a);
```
mppf: Since everything up to this point is positive, we don't need this final abs, right?

victor-ludorum: Oh.. sorry sir !! I will update this thanks !!

modules/standard/Math.chpl
victor-ludorum: Thanks sir !! I will update this

victor-ludorum:
@mppf I have updated the code as you have suggested . Kindly review it .

modules/standard/Math.chpl
```diff
@@ -1056,6 +1056,21 @@ module Math {
     extern proc truncf(x: real(32)): real(32);
     return truncf(x);
   }
+  
+  /* Returns the greatest common divisor of the integer argument `arg1` and
+     `arg2`. */
```
mppf: The documentation refers to `arg1` and `arg2` but these arguments are called `a` and `b`. This is important because you can pass arguments by name in Chapel (e.g. `gcd(b=24, a=8)`.

mppf:
I think this is very close. I just need to run testing on it before we can merge it.

modules/standard/Math.chpl
victor-ludorum: Yeah.. minor mistake sorry sir for bothering again . I will update it.

victor-ludorum:
@mppf Updated the documentation part .

mppf:
@victor-ludorum - this isn't a big deal right now, but I wanted to point out something to help your future development. In general, it's important to make each individual commit have a message describing what that individual commit changes. E.g. in your last commit, it was titled "Adding gcd function in Math.chpl file" but it should have been something like "Updating gcd documentation".

Alternatively, you can develop without regard to writing good commit messages, and then rebase and squash all the commits into one. The problem there is that - say you need to split your work into 2 PRs - if you didn't write good commit messages you'll be stuck staring at a diff. Anyway, you can rewrite your history with 'git rebase -i' and send it to GitHub with 'git push -f'. I recommend you read up on these. 

But, it's not critical for this PR. Assuming I'm the one to merge this PR, I can use a GitHub feature to rebase it all into one commit. You're welcome to try the 'git rebase -i' but if you don't, I'll squash it this way.

A lot of that is described here, (for others, and in case you havn't looked at it yet) https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/ContributorInfo.rst but I know it's pretty long.

victor-ludorum:
Thanks @mppf sir !! I will always follow it in future and apologize me for this time

mppf:
@victor-ludorum - No need to apologize, I understand you're learning.

PR #8329, opened 2018-01-30
remove non-init halts from Crypto module

psahabu:
Discussion after the merge of #8244 prompted a closer examination of the Crypto module. While it was not possible to remove halts from `init()` due to the present inability to throw from initializers, this PR locates all the other halts in the module and has them throw a new standard error type, `IllegalArgumentError`.

### DIFF
--- a/modules/internal/ChapelError.chpl
+++ b/modules/internal/ChapelError.chpl
@@ -106,8 +106,7 @@ module ChapelError {
     }
 
     proc message() {
-      var msg_fmt = "illegal argument '%s': %s";
-      return msg_fmt.format(arg_name, arg_info);
+      return "illegal argument '" + arg_name + ': ' + arg_info;
     }
   }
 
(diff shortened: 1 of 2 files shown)


saru95:
The tests checking for `halts` should fail. The `.good` file for those will also have to be updated to reflect the error message. Something like this file - https://github.com/chapel-lang/chapel/blob/master/test/library/packages/Crypto/saru/utils/buffer3.good

psahabu:
@saru95 Right, once we can throw errors in initializers then these tests will need to be updated. Until then this is the most that can be done.

@mppf Will do.

lydia-duncan:
My two cents: I might be tempted to have the IllegalArgumentError class store the name of the argument and the reason for it being illegal and then give a standardized message of "Argument <arg> was illegal because <reason>" or something along those lines.  It doesn't seem very distinguished from other error subclasses otherwise.

psahabu:
@mppf what do you think of the changes to `IllegalArgumentError`?

mppf:
@psahabu - look good to me!

PR #8330, opened 2018-01-30
Minor cleanup in parallel.cpp

vasslitvinov:
* Pass the formal to needsAutoCopyAutoDestroyForArg()
  because it is already available.

* Remove a couple of checks that we expected to be unnecessary
  and that testing confirms that they are.

* Add a comment.

Passes paratest under linux64 --verify and gasnet.


vasslitvinov:
@benharsh - would you review?

PR #8331, opened 2018-01-30
Improve view.cpp debugging facilities

vasslitvinov:
* Improved list_view() for ForallStmt and DeferStmt.

* Added a couple of variants on aid(), for convenience while debugging.

* Added debugParentExpr to return an Expr*.
  Changed debugParentSym to return a Symbol*.

* Added vec_view on std::vector<Symbol*>,
  following its counterpart on 'Vec'.

* While there, added and used ForallStmt::isForallIterExpr(),isForallLoopBody(),
which replaced ForallStmt::isIteratedExpression().


lydia-duncan:
These seem like pure improvements, thanks! (I didn't review this closely, just checked to make sure it didn't worry me)

PR #8332, opened 2018-01-30
Minor cleanup of uses of breakOnResolveID:

vasslitvinov:
* Do not print_view the FnSymbol for which the breakpoint hits.
  These printouts can be too lengthy.
  Instead, print a couple of pieces of information about the FnSymbol.

* Print CallExpr id before print_view on it upon breakOnResolveID.

* Uniformize code formatting in foralls.cpp w.r.t. breakOnResolveID.

### DIFF
--- a/compiler/AST/foralls.cpp
+++ b/compiler/AST/foralls.cpp
@@ -670,8 +670,7 @@ static void resolveParallelIteratorAndIdxVar(ForallStmt* pfs,
                                         origIterator, alreadyResolved);
   VarSymbol* idxVar = parIdxVar(pfs);
 
-  if (idxVar->id == breakOnResolveID)
-    gdbShouldBreakHere();
+  if (idxVar->id == breakOnResolveID) gdbShouldBreakHere();
   idxVar->type = iType.type();
   idxVar->qual = iType.getQual();
 }
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -2040,6 +2040,8 @@ FnSymbol* resolveNormalCall(CallExpr* call, bool checkOnly) {
   FnSymbol* retval = NULL;
 
   if (call->id == breakOnResolveID) {
+    printf("breaking on resolve call %d:\n", call->id);
+    print_view(call);
     gdbShouldBreakHere();
   }
 
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -2040,8 +2040,6 @@ FnSymbol* resolveNormalCall(CallExpr* call, bool checkOnly) {
   FnSymbol* retval = NULL;
 
   if (call->id == breakOnResolveID) {
-    printf("breaking on resolve call:\n");
-    print_view(call);
     gdbShouldBreakHere();
   }
 
--- a/compiler/resolution/initializerResolution.cpp
+++ b/compiler/resolution/initializerResolution.cpp
@@ -111,6 +111,8 @@ static void resolveInitCall(CallExpr* call) {
   CallInfo info;
 
   if (call->id == breakOnResolveID) {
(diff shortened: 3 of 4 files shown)


lydia-duncan:
Can you provide some examples of how things were before?  I'm having trouble visualizing it.

vasslitvinov:
**Group 1** : breaking upon resolving a call or a var symbol:

Before:

* In most cases:
```
    Breakpoint 1, gdbShouldBreakHere () at misc.cpp:61
    (gdb) 
    // i.e. just the gdb prompt
```
* In a couple of cases:
```
    breaking on resolve call:
    (CallExpr move
      (SymExpr 'val AS_result2:int(64)')
      (SymExpr 'val initTemp:int(64)'))
    
    Breakpoint 1, gdbShouldBreakHere () at misc.cpp:61
    (gdb) 
```
Now:

* Uniformly in all cases:
```
    Breakpoint 1, gdbShouldBreakHere () at misc.cpp:61
    (gdb) 
```
**Group 2** : breaking upon resolving a function:

Before:

    breaking on resolve fn:
    'fn coforall_fn:void'
      ( DefExpr 'const-val arg _coforallCount:_EndCount(atomic_int64,int(64))' intent-'const in' )
      (DefExpr 'const-val arg _x2_loc:locale' intent-'const in' )
      (DefExpr 'val arg _x3_result1:int(64)' intent-'in' )
      (DefExpr 'const-val arg _x4_RP_result2:SumReduceScanOp(int(64))' intent-'const in' )
      (BlockStmt
        (DefExpr 'val RP_RP_result2:SumReduceScanOp(int(64))' )
        (CallExpr move
          SymExpr 'val RP_RP_result2:SumReduceScanOp(int(64))')
          (CallExpr
            (SymExpr 'fn clone:SumReduceScanOp(int(64))')
    
      .... THE ENTIRE FUNCTION IS PRINTED OUT ....
      )
    
    Breakpoint 1, gdbShouldBreakHere () at misc.cpp:61
    (gdb) 

Now:

    breaking on resolve fn coforall_fn[131775] (3 args)
    Breakpoint 1, gdbShouldBreakHere () at misc.cpp:61
    (gdb)

lydia-duncan:
Thanks for the explanation.  I've found it helpful in the past when the call expression has been printed out.  This is the workflow I've used:
1) A program I was running got an internal error.
2) I see that the problem is something to do with a call expression that is being resolved
3) I exit lldb and run again with --break-on-resolve-id with the id of the call expression
4) Sometimes the problem is with the resolution of one of its inner pieces which comes first, so I'll reach gdbShouldBreakHere and not get the call expression output
  - Seeing the call expression output is a good indicator to me that I was able to break prior to when the compiler hard fails and so could analyze the behavior prior to the failure.  I can then adjust as needed instead of having to swap to a higher frame to determine what was going on. (because pausing on gdbShouldBreakHere doesn't give me any information about what has happened to cause the break, just that one of the call sites of it was trigger)

I can see the output from breaking on the function as being too much (especially in the case of some functions).  And I'm not entirely reliant on the output - I'm mostly putting this out there because it's a mode I've found convenient to operate in and it seems different than the one you use.

lydia-duncan:
At the end of the day, I can always print the call expr out when I get to the frame where it is visible (frame select 1; lview call;).  My point is that having it appear when the program interrupts is faster than having me make the two operations necessary to do so, and printing out the call expression does not take very much space.

At the end of the day it doesn't matter too much overall, but generally we want to avoid making arbitrary preference changes and having one person like it the current way and another person like it a different way doesn't seem like enough groundswell to justify making the change.  I am curious if other people use the feature and could break the tie, and would prefer to wait until we know no one else is interested before having a final decision.

I am okay with the other changes in this PR.  I haven't reviewed them super much but the general concept seems fine (and I can actually review if you want).  Maybe they should be separated from the bit we disagree on?

PR #8333, opened 2018-01-30
Remove getFirstChild()

vasslitvinov:
This member function is currently unused
and not well understood.

Removing it completely.


mppf:
Related to this, is getFirstExpr actually used, other than to implement other getFirstExpr?

PR #8182, opened 2018-01-11
Introducing Quiescent State-Based Reclamation to Chapel

LouisJenkinsCS:
I introduce Quiescent State-Based Reclamation, a memory reclamation algorithm that can be from the confines of the runtime and potentially from Chapel user code (with some GOTCHAS). The memory reclamation algorithm comes with very little performance regression and can ensure the eventual cleanup of memory so long as checkpoints are periodically called from all threads (although their placement is up to debate). 

QSBR can come in handy for any performance-critical data structure, and currently is used in Chapel's privatization table. Future uses can be for the thread-safety of Chapel's callback system and perhaps in more dynamic task local storage. In the future, a specific QSBR table is planned for users that will not be able to interfere with the runtime.

**Potential Uses of QSBR:**
1. Non-Blocking Data Structures: It is safe to defer the deletion of anything logically removed from the data structure in question. For example, a lock-free queue that removes a node from the head can safely defer deletion as it will be the only thread to do so (since the CAS operation will return true for only one successful operation)
2. Multiword Compare-And-Swap: This also applies to anything using descriptors for helping mechanisms, but a K-Word Compare-And-Swap where we make use of an object that holds enough state to represent some in-progress operation to claim ownership of the location. The descriptor can be accessed by concurrent threads and so it is safe to defer for deletion after the owning thread/task completes its operation.
3. Software Transactional Memory: While things like reading and writing are trivial to implement, a huge problem that arises is the deletion of data during a transaction and even after committing it. Deleting the object immediately can result in a segmentation fault for tasks that have not yet aborted their transaction, so deferring deletion would be a solution so long as no checkpoint is invoked within a transaction.

There are many other potential uses, the sky is the limit.

Reviewed by @mppf, @gbtitus, @ronawho.

Testing:

 - [x] 10000 trials of test/distributions/privatization/runtime/ in the standard configuration
 - [x] full local testing
 - [x] ugni testing for test/release/example/primers
 - [x] full quickstart testing
 - [x] full gasnet testing
 - [x] full gasnet+fifo testing

PR #4902, opened 2016-11-23
Dramatically improve task creation speed

ronawho:
Improve task creation times by using a hybrid spin/condwait scheme for qthreads
nemesis scheduler. While digging into LCALS performance we discovered a large
portion of our overhead compared to OpenMP is in task creation. This improves
the performance of a task startup microbenchmark from 21s to 7s compared to
3.2s for an OpenMP variant. While we're still off from OpenMP, this is a vast
improvement, and I have a good lead on the remaining sources of overhead. With
additional work, we should be able to get close if not on par with OpenMP task
creation times.

Details of this commit:

qthreads --enable-condwait-queue controls whether idle workers spinwait or do a
condwait while looking for new tasks. Pure spinwaiting kills single threaded
performance because idle threads pound the cpu nonstop, but pure condwait (our
previous default) increases task startup.

Since neither pure spinwaiting nor pure condwaiting work, this implements a
simple hybrid spin/condwait approach where we spin QT_SPINCOUNT (default
300,000) iterations before doing a condwait.

Initially I thought this approach was too naive, but it's actually very similar
to what gnus omp implementation does (and is also where the 300,000 spins comes
from.)

This has a dramatic performance improvement for several benchmarks:

  - 3X speedup on task startup microbenchmark
  - 2.5 speedup for LCALS and Lulesh
  - 50% speedup for HPL, NAS CG, and NAS FT
  - 25% speedup for MiniMD and SSCA#2
  - minor improvements for other benchmarks

I only saw one minor regression (10% perf loss) for the serial synch_p2p PRK,
but considering the other improvements, this seems acceptable.

This hybrid scheme still needs to be applied to other schedulers, and needs to
be contributed upstream, but given it's impact it's worth adding it now.

PR #8335, opened 2018-01-30
Switch perf playground to measure the impact of using RCU for privatization

ronawho:


ronawho:
Yeah, it'll run your branch nightly and compare performance to master. It's single-locale only, so it won't test privatization overhead, but it will tell us if your code adds any significant overhead to task-spawning or other operations.

I can do multi-locale testing as well, but that's more manual. 

I was waiting for https://github.com/chapel-lang/chapel/pull/8182 to converge/settle-down before merging, but I can start testing tonight if you're ready.

mppf:
@ronawho @LouisJenkinsCS - I think it'd be great to get it testing in the perf playground tonight. I havn't run a full local testing of it recently (though it did pass earlier and I think it's more likely an error would be discovered by the 10000 trials of test/distributions/privatization/runtime/ that I did... but you never know).

mppf:
@LouisJenkinsCS - now that this PR is merged, we'll get performance graphs from each night. If you update your branch, the testing after that will reflect the change.

ronawho:
Yeah, this just grabs whatever code is in your branch at ~11:45 PST and will test that against master.

ronawho:
Results are up at https://chapel-lang.org/perf/chapcs/RCU-Privatization. There is only a single data point for your branch, but after tonight your branch will be represented with a dashed line, while master is solid. 

Unfortunately there are some non-trivial performance regressions, likely due to increased task-creation time:  https://chapel-lang.org/perf/chapcs/RCU-Privatization/?startdate=2018/01/30&enddate=2018/02/02&graphs=luleshrelease,minimdljsize10time,lcalsrawompshort,nasparallelbenchmarkscgtimingssizea,emptytaskspawntimings500000xmaxtaskpar,timetobuildtrees

All of these tests create a lot tasks (usually in a trial loop, not all at once.) I would start with "Empty Task Spawn Timings" which really just times something like:

```chpl
for 1..numTrials {
  coforall 1..here.maxTaskPar() {
    // do nothing, no-op
  } 
}
```

LCALS (especially the short problem size) is dominated by task spawning times. https://chapel-lang.org/releaseNotes/1.15/09-perf-opts.pdf details recent optimizations we did to task spawning and has more info on LCALS and the empty task-spawn test since they were what motivated the improvements.

mppf:
start_test -performance does, and you can give it the path to the particular test you're interested in.

ronawho:
This sort of "create tasks in a trial loop" is what LCALS does too (and improving LCALS was the motivation for adding the empty task spawn test)

https://github.com/chapel-lang/chapel/pull/4902 changed things so we would only put worker pthreads that are looking for new tasks to sleep after ~300,000 iterations through a loop. That's all inside qthreads-core though, so I don't think we'll want to make any mods or add any hooks there. 

Nothing immediately comes to mind for how you would cleanly get that sort of information, but I'll stew on it for a little

ronawho:
>I already have a qthread_registerOnPark and qthread_registerOnUnpark hooks that are called, so I should be able to move the callbacks to being after spinning num_spins_before_condwait and if it works just add another environment variable to the Qthread library to allow fine-tuning.

The spinwaiting is inside qthreads-core though, not inside our qthreads shim, so you don't really have access to it.

>However, I'm gonna say that its messed up that a revolutionary language like Chapel has to bend over backwards for legacy applications written for another paradigm and memory model; it'd be like if C had to prove it had equivalent performance to assembly. I believe it'd be more efficient if our port of applications was less of an emulation of OMP and was more PGAS/Chapel appropriate.

I don't necessarily disagree, but fast task spawning speed is still important even for more idiomatic Chapel codes. Threre's also nothing about LCALS that's terribly un-idiomatic chapel. It uses direct array indexing, instead of iterating over arrays directly, but that doesn't change the task-spawning implications.

For spaces like ML/AI/whatever we don't have to be competitive with MPI/OpenMP, but the HPC space tends to care about performance above all else.

PR #8340, opened 2018-01-31
Report an error for a return inside a forall.

### LINKED ISSUE #8251: Returning from a forall loop should generate a compiler error
```chpl
//Functions for uppper triangular matrix and lower triangular matrix
var row:int =3;
var col:int =3;
var ar=[[1,2,3],[1,1,1],[10,10,10]];
proc upper(arr,r, c): (bool) {
	forall i in 1..#r{
		forall j in 1..#c{

			if(i<j&&arr[i][j]!=0)
			{
				return false;
			}
		}
	}
	return true;
}
writeln(upper(ar,row,col));
proc lower(arr,r,c): (bool) {
	forall i in 1..r{
		forall j in 1..c{
			if(i>j&&arr[i][j]!=0)
			{
				return false;
			}
		}
	}
	return true;
}

```

It enters if condition but not returning false in upper triangular matrix. I tried to solve issue  #5753  end up creating new issue.

vasslitvinov:
Resolves #8251.

When I switch to representing forall loops with ForallStmts in #5295, I omitted a call to checkControlFlow(). I am adding it here.

There are a couple of interesting pieces in addition to that.

* We DO allow continue inside a forall. Unless that continue jumps out of the forall by naming a loop outside of the forall. I adjusted checkControlFlow() for that.

* I added some break/continue checking inside forall in scopeResolve.cpp - see handleForallGoto() (invoked from resolveGotoLabels()) - without being aware of the missing call to checkControlFlow(). That checking is insensitive to jumping outside of the forall and is now superceded by checkControlFlow(). So I removed the checking component of handleForallGoto().

I beefed up the existing test of checkControlFlow(), test_control_flow.chpl.

I also removed offending pieces from the test forall-continue-good.chpl and removed forall-continue-errors.chpl altogether in favor of the beefed-up test_control_flow.chpl


vasslitvinov:
@benharsh would you review?

PR #8343, opened 2018-02-01
Fix issue with default initializers called on type aliases to instantiations

lydia-duncan:
Prior to this change, if a `new` call was made using a type alias to a generic
instantiation of a class with default initializers, we would fail to find the
default initializer.  This was because "resolveNewHasInitializer()" was not
recognizing that types with default initializers did not define constructors
when we were executing "resolveNew()".

With this change, three more tests pass with --force-initializers, and one has
changed its failure mode.

Adds a test to the test/classes/initializers/compilerGenerated/generics
directory to lock in the improved behavior.

Testing:
- [x] full paratest
- [x] full paratest with --force-initializers


lydia-duncan:
@noakesmichael or @benharsh, would one of you mind reviewing?  The change is fairly straight-forward

PR #8344, opened 2018-02-01
Fix opaque domain assignment

lydia-duncan:
The assignment function for domains did not work for opaque domains due to
accessing their idxType, which is forbidden.  Add a check prior to accessing
the idxType to avoid making the check against opaque domains so that they can
still use assignment.

Also adds a test of opaque domain assignment, to lock in the behavior.  This
test failed prior to this change.

I encountered this issue via compiler generated default initializers.  With this
fix, two of the four tests that failed with the message "opaque domains do not
currently support .idxType" now pass and the other two have changed their
failure mode.

Passed a full standard paratest


lydia-duncan:
@bradcray - would you like to review this?  It should be short.  Otherwise, I can grab someone else

bradcray:
This looks good to me, thanks!

PR #8346, opened 2018-02-01
Add Cholesky factorization and eigenvector solvers to LinearAlgebra

daviditen:
Add some linear algebra functions that call into functions in the LAPACK
module to implement Cholesky factorization and to find eigenvalues and
eigenvectors.

Added a new test for each function.


test/library/packages/LinearAlgebra/correctness/testCholesky.chpl
```diff
@@ -0,0 +1,52 @@
+use LinearAlgebra;
+/*
+proc cholesky(A: [] real, uplo: string = 'L') where A.rank == 2 {
```
ben-albrecht: Did you mean to remove this comment block?

modules/packages/LinearAlgebra.chpl
```diff
@@ -937,6 +937,132 @@ proc trace(A: [?D] ?eltType) {
   return trace;
 }
 
+/* Type to choose which triangle of a matrix is desired */
+enum lowerOrUpper {
```
ben-albrecht: It'd be nice if we could get away from the `enum` arguments for the `LinearAlgebra` interface. What do you think of a `lower=false` argument instead, similar to [scipy](https://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.linalg.cholesky.html)

modules/packages/LinearAlgebra.chpl
```diff
@@ -937,6 +937,132 @@ proc trace(A: [?D] ?eltType) {
...
+  potrf(lapack_memory_order.row_major, uploStr, copy);
+
+  // tril and triu make/return an extra copy.  Should we zero the unused
+  // triangle of the array manually instead to avoid the copy?
+  return if uplo == lowerOrUpper.lower then tril(copy) else triu(copy);
+}
+
+
+/* Type to choose which eigenvectors to compute in the ``eigvals`` function. */
+enum EigenVecType {
```
ben-albrecht: I feel similarly about this `enum` argument, but am not sure how it could map down to something simpler. Do you know how other libraries (such as scipy, matlab, julia, etc.) handle this argument?

test/library/packages/LinearAlgebra/correctness/testEigen.chpl
```diff
@@ -0,0 +1,120 @@
+use LinearAlgebra, Random;
+
+config const epsilon = 0.0000000001;
+
+proc isClose(a: complex, b: complex) {
```
ben-albrecht: We really need to add this to `Math.chpl` or something...

test/library/packages/LinearAlgebra/correctness/testCholesky.chpl
daviditen: Yes, I did mean to remove it.  Gone now.

modules/packages/LinearAlgebra.chpl
daviditen: I removed this one in favor of a `lower=true` argument.

modules/packages/LinearAlgebra.chpl
daviditen: Julia and SciPy both have separate functions for computing only eigenvalues vs. both eigenvalues and eigenvectors.  Neither appears to have a direct way to get the left eigenvectors.

Matlab bases it on what the return value is assigned to:
```matlab
e = eig(A)       % just eigenvalues
[V,D] = eig(A)   % eigenvalues and right eigenvectors
[V,D,W] = eig(A) % eigenvalues, right eigenvectors, and left eigenvectors
```

ben-albrecht: Thanks for looking into.

Matlab is pretty slick. I don't know of a way to do that in Chapel..

Splitting into separate functions like SciPy and Julia seems appealing to me. Do you think we could do something similar with our interface?

daviditen: I agree we can't do what Matlab does without a pretty significant language change, so that's out.

If you squint just right this is kind of already multiple functions since `rl` is a param.

If we were to break them up it would need to be four different functions with four different names.  I initially did have them split and overloaded, and which one got picked was based on the arguments, but as I simplified away arguments wasn't able to do that anymore, so I combined them to keep the overload feel.

modules/packages/LinearAlgebra.chpl
```diff
@@ -938,6 +938,125 @@ proc trace(A: [?D] ?eltType) {
...
+
+   * If ``left`` and ``right`` are both ``true`` then both eigenvectors are
+     computed.  The return value is a tuple with three elements:
+     ``(eigenvalues, leftEigenvectors, rightEigenvectors)``
+
+   * If ``left`` and ``right`` are both ``false`` only the eigenvalues are
+     computed, and returned as a single array.
+
+ */
+proc eigvals(A: [] ?t, param left = false, param right = false)
```
ben-albrecht: Just occurred to me: Would it be more Chapelistic to camelCase this as `eigVals()`?

daviditen: It's short for the single word "eigenvalues", so IMO, no.

modules/packages/LinearAlgebra.chpl
```diff
@@ -938,6 +938,125 @@ proc trace(A: [?D] ?eltType) {
...
+        skipNext = true;
+        cplx[.., j  ] = vec[.., j] + 1.0i*vec[.., j+1];
+        cplx[.., j+1] = vec[.., j] - 1.0i*vec[.., j+1];
+      }
+    }
+    return cplx;
+  }
+
+  const n = A.domain.dim(1).length;
+  assert(A.domain.dim(1) == A.domain.dim(2));
```
ben-albrecht: Error message

test/library/packages/LinearAlgebra/correctness/testEigen.chpl
```diff
@@ -0,0 +1,120 @@
+use LinearAlgebra, Random;
```
ben-albrecht: Were these tests derived from another test suite?

modules/packages/LinearAlgebra.chpl
```diff
@@ -938,6 +938,125 @@ proc trace(A: [?D] ?eltType) {
 }
 
 
+/* Perform a Cholesky factorization on matrix ``A``.  ``A`` must be square.
+   Argument ``lower`` indicates whether to return the lower or upper
+   triangular factor.  Matrix ``A`` is not modified.
```
ben-albrecht: Mention return type?

test/library/packages/LinearAlgebra/correctness/testEigen.chpl
daviditen: No, I wrote them from scratch.  The numbers used in the matrix in testCholesky were taken from https://rosettacode.org/wiki/Cholesky_decomposition to make sure I ended up with something reasonable.

daviditen:
Thanks for reviewing!

PR #8348, opened 2018-02-01
bug future: modifying parent const fields in phase 2 of init()

bradcray:
"Phase 2" will soon be redefined, but it seems likely/possible that this
problem may persist even in the new initializers world, so I'm filing this
before I forget.


bradcray:
@lydia-duncan : Any idea whether we already have an issue / bug filed against this already?

lydia-duncan:
That sounds vaguely familiar but I'm not seeing a test that would match it, so I suspect this is a future we don't have already.  Thanks!

PR #8352, opened 2018-02-01
Remove removeWrapRecords from compiler perf since it is no longer a pass

ronawho:
It was removed in https://github.com/chapel-lang/chapel/pull/8349


ronawho:
@mppf quick sanity check?

PR #8353, opened 2018-02-01
Use cobegin to create llTree in parallel with others

bradcray:
A recent discussion on the CLBG site suggested that it's legal to build the long-lived
tree in parallel with the others.  This rewrite does that using a cobegin, to see whether
it'll result in any savings or not, by getting more processors involved sooner.  In doing
this, I've repurposed my personal study version of the benchmark and brought it more
in line with the release version to minimize the diff.


bradcray:
@ronawho : This is what we've been (very slowly) discussing every month or two for the past several months.  Can you give it a careful review and see what you think?

Issue #6559, opened 2017-06-28
one-character developer-only compiler flags


mppf:
I noticed that some developer-only flags have one-character equivalents. For example `-d` is equivalent to `--log` and `-t` is equivalent to `--html`.

I think it might be confusing to users for the compiler's argument parsing to depend on features that they can't see. For example, the error for `-devel` only makes sense if one understands that `-d` is an option.

Furthermore these developer-only 1-character equivalents prevent us from adding user-facing flags with obvious names.

Lastly, 1-character equivalents for little used flags (such as `--parser-debug`) do not seem useful especially when the 1-character short name is not immediately obvious.

What 1-character developer-only options are there?
 * `-d` for `--log`
 * `-t` for `--html`
 * `-D` for `--parser-debug`

I think we should remove these 1-character equivalents and as a matter of policy avoid adding 1-character short names for any developer-only command line arguments.

daviditen:
I use `-t` very frequently.  I think improving our error messages is a better choice than removing the short options.  Developer options should also be completely removed from a non-developer build so that `-devel` would give the same error as `-unknown-option`.

I don't think we should have any spelled-out flags that use a single dash, so I'm not sure how these prevent any obvious user-facing flags.

bradcray:
I'd be OK with removing one-character developer-only flags.  Alternatively, it seems that we could error on developer flags unless --devel is thrown first (or CHPL_DEVELOPER is set).  I wouldn't make this a compile-time choice as David suggests because I think it's often useful to flit between user and developer modes without having to rebuild, or to ask a user to throw a developer flag to see what happens (and asking them to rebuild, try it, and then rebuild again once they want to go back to user mode seems likely to stop the conversation).  I also don't see any real downside to making it an execution-time conditional rather than a compile-time one.

mppf:
@daviditen - we might want to use `-t` for something else in the future as a user-facing 1-character flag. Additionally, I'd personally prefer it of the compiler didn't care about the difference between `-` and `--` for most options.

mppf:
After discussion with @daviditen, I've created a PR to remove these 1-arg flags. David says he's OK as he's the only one who was using them.

mppf:
Closing since I meant to close it when merging PR #8357.

PR #8357, opened 2018-02-02
Remove 1-character developer-only compiler flags

mppf:
Removes -d -t and -D which are 1-character developer only
flags. The reason for doing this is to reduce confusion for
users and to make these 1-character flags possibly available
for use as user flags.

Resolves issue #6559.

Passed full local testing.

Discussed with many in issue #6559. PR reviewed by @ben-albrecht - thanks!


daviditen:
I'm not sure if you've hit this yet, but `test/chplenv/autocomplete/autocomplete.chpl` is going to fail since options were removed.  You can run `util/devel/gen-chpl-bash-completion` to regenerate `/Users/diten/chapel7/util/chpl-completion.bash` and it will be back to passing.

mppf:
I need to adjust the following two tests:
 * chplenv/autocomplete/autocomplete
 * compflags/bradc/oneDash/dashDevel

Everything else passed cleanly.

PR #8365, opened 2018-02-02
Demote UtilReplicatedVar to package module

benharsh:
Long-term we expect to have a more elegant way to represent replicated variables, at which point ``UtilReplicatedVar`` will no longer be useful or appropriate as a standard module. If we think we're going to deprecate something in the standard modules, we should do so sooner rather than later. This PR demotes ``UtilReplicatedVar`` to a package and renames it as ``ReplicatedVar``.

For now ``UtilReplicatedVar`` will still be usable, but will result in a compiler warning when used. All uses have been replaced with ``ReplicatedVar``.

Testing:
- [x] full local
- [x] full gasnet
- [x] library/packages/MPI

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -110,6 +110,7 @@ PACKAGES_TO_DOCUMENT = \
 	packages/OwnedObject.chpl \
 	packages/RangeChunk.chpl \
 	packages/RecordParser.chpl \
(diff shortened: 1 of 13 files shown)


bradcray:
Works for me.  (This isn't a detailed review, I just looked at the general approach).

PR #8368, opened 2018-02-03
Change default intent for range to const in

### LINKED ISSUE #5256: default intent for ranges?
What should be the default intent for ranges?
At the very least this should be documented in the spec.

It should probably be `const`.
Concretely, should it be `const in` or `const ref` ?

An important consideration is in handling of range literals and begin statements. E.g.:

    proc processEm(myRange) {
      // do something with the range...
      writeln(myRange);
    }
    proc startIt(myRange) {
      begin { processEm(myRange); }
    }
    startIt(1..10);

Cf. the same for records:

    startIt(new MyRecord());

See also the thread "default argument intent for ranges?" on the internal list in October 2016.

mppf:
This PR adjusts the compiler to make range have similar intents to int. The main difference is `const in` vs `const ref` by default.

Resolves #5256.

Passed full local testing and full gasnet testing.

### DIFF
--- a/test/studies/amr/lib/amr/BergerRigoutsosClustering.chpl
+++ b/test/studies/amr/lib/amr/BergerRigoutsosClustering.chpl
@@ -237,7 +237,8 @@ proc CandidateDomain.trim()
       //---- Approximately center the enlarged range ----
       
       var n_overflow_low = (min_width(d) - R.length) / 2;
-      R = ((trim_low - n_overflow_low*stride .. by stride) #min_width(d)).alignHigh();
+      var tmp = ((trim_low - n_overflow_low*stride .. by stride) #min_width(d));
+      R = tmp.alignHigh();
 
 
       //---- Enforce bounds of D ----
--- a/test/types/range/ferguson/range-intents.chpl
+++ b/test/types/range/ferguson/range-intents.chpl
@@ -84,6 +84,9 @@ proc const int.test_const_this() {
 global = start;
 global.test_const_this();
 
+// We don't support 'in' or 'const in' as 'this' intents.
+// If we did, I'd want these functions uncommented to test it.
+
 /*proc in range.test_in_this() {
   global = change;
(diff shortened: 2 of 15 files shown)


mppf:
@benharsh - would you like to review this change-range-default-intent PR? I'm also interested to know if you think it needs perf analysis before merging. Thanks!

test/studies/amr/lib/amr/BergerRigoutsosClustering.chpl
```diff
@@ -237,7 +237,8 @@ proc CandidateDomain.trim()
       //---- Approximately center the enlarged range ----
       
       var n_overflow_low = (min_width(d) - R.length) / 2;
-      R = ((trim_low - n_overflow_low*stride .. by stride) #min_width(d)).alignHigh();
+      var tmp = ((trim_low - n_overflow_low*stride .. by stride) #min_width(d));
+      R = tmp.alignHigh();
```
benharsh: Can we start an issue for this?

test/types/range/ferguson/range-intents.chpl
```diff
@@ -0,0 +1,184 @@
...
+  writeln("const this ", upper(this));
+}
+proc const int.test_const_this() {
+  global = change;
+  writeln("const this ", upper(this));
+}
+global = start;
+global.test_const_this();
+
+/*proc in range.test_in_this() {
```
benharsh: Should this commented-out block still be here? If so, could you add a comment explaining why?

mppf: We don't support 'in' or 'const in' as 'this' intents. (Why not? I dunno.) If we did, I'd want these functions uncommented to test it.

test/studies/amr/lib/amr/BergerRigoutsosClustering.chpl
mppf: issue #8386 is started.

PR #8369, opened 2018-02-03
Remove try! and halt from BLAS module

psahabu:
Continues the effort of removing try! and halt from our modules where possible.

- [x] passes BLAS testing on cray-xc

### DIFF
--- a/modules/packages/BLAS.chpl
+++ b/modules/packages/BLAS.chpl
@@ -1763,10 +1763,10 @@ module BLAS {
 
 
   */
-  proc rotmg(ref d1: ?eltType, ref d2: eltType, ref b1: eltType, b2: eltType, P: []eltType) {
+  proc rotmg(ref d1: ?eltType, ref d2: eltType, ref b1: eltType, b2: eltType, P: []eltType) throws {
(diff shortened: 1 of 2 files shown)


modules/packages/BLAS.chpl
```diff
@@ -641,7 +640,7 @@ module BLAS {
         n = Bdom.dim(2).size : c_int;
 
     if m != n then
-      try! halt("Non-square array of dimensions %ix%i passed to trmm".format(m, n));
+      throw new IllegalArgumentError("B", "Non-square array of dimensions %ix%i".format(m, n));
```
mppf: There are many of these that include the method name (trmm in this case). Should we include that in the new error? I'd presume it'd help, until we have stack traces on errors...

ben-albrecht:
We should probably confirm tests depending on BLAS are tested, namely `LinearAlgebra`

modules/packages/BLAS.chpl
psahabu: I forgot that traces aren't on by default, so I've restored the original error messages.

PR #8370, opened 2018-02-03
Allow Chapel to compile on an ARM-based Cray system

dmk42:
Right now, third-party Cray builds are hard coded to x86_64.  This change lets them compile on whichever architecture is being used.

Note that `$HOSTTYPE` only provides the machine type in bash, and `/bin/arch` reports some x86_64 systems as i386.  `uname -m` works for Cray Intel, AMD, and ARM systems.

This change only affects systems that report as `cray-x`-something.

Test plan - hellos and studies/shootout/pidigits/pidigits-gmp.chpl on:
- [x] Intel-based Cray systems
- [x] ARM-based Cray systems
- [x] AMD-based Cray systems

Note that MassiveThreads gets far enough to complain about missing architectural features, but has not yet been ported to 64-bit ARM.

### DIFF
--- a/third-party/gmp/Makefile
+++ b/third-party/gmp/Makefile
@@ -10,7 +10,7 @@ include $(CHPL_MAKE_HOME)/make/Makefile.base
 #
 GMP_CROSS_COMPILED=no
 ifneq (, $(filter cray-x%,$(CHPL_MAKE_TARGET_PLATFORM)))
-CHPL_GMP_CFG_OPTIONS += --host=$(HOSTTYPE)-cle-linux-gnu
+CHPL_GMP_CFG_OPTIONS += --host=$(shell uname -m)-cle-linux-gnu
 GMP_CROSS_COMPILED=yes
 else ifneq ($(CHPL_MAKE_HOST_PLATFORM),$(CHPL_MAKE_TARGET_PLATFORM))
 CHPL_GMP_CFG_OPTIONS += --host=$(CHPL_MAKE_TARGET_PLATFORM)-unknown-linux-gnu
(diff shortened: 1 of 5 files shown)


dmk42:
@gbtitus , @ronawho - Would you please review?

ronawho:
This looks good to me, but I think @gbtitus is the more qualified reviewer here.

third-party/gmp/Makefile
```diff
@@ -10,7 +10,7 @@ include $(CHPL_MAKE_HOME)/make/Makefile.base
 #
 GMP_CROSS_COMPILED=no
 ifneq (, $(filter cray-x%,$(CHPL_MAKE_TARGET_PLATFORM)))
-CHPL_GMP_CFG_OPTIONS += --host=x86_64-cle-linux-gnu
+CHPL_GMP_CFG_OPTIONS += --host=$(shell uname -m)-cle-linux-gnu
```
gbtitus: It's a little unclear to me whether `uname -m` or `uname -p` is what we want here.  `git grep` in the gmp-src subdir seems to imply that `uname -p` might be what's needed, but I'm far from certain about that.

dmk42: You aren't the only one to wonder about it.  It varies depending on the type of system.  Solaris man pages say to use `uname -p` and not `uname -m`.  Linux man pages say `uname -p` is not portable.  On a Mac, `uname -p` yields i386 and `uname -m` yields x86_64.

It turns out that on the current Cray systems, `uname -p` and `uname -m` return identical results (x86_64 or aarch64).  Since Cray systems run Linux, I followed the Linux `uname` man page and avoided `uname -p` because they say it is nonportable.  However, no matter what I choose, there is no guarantee we won't have to revisit it someday.

dmk42: By the way, in all the third-party packages, they use both `uname -p` and `uname -m` in their configure scripts.

gbtitus: Based on their saying that `uname -p` is nonportable I'd say stick with your original `uname -m`, then.  Good to go!

PR #8379, opened 2018-02-03
Mark regressions for JIRA issue 329

awallace-cray:
Detect regressions with the following error message in the output,
    "Fatal MPP reservation error on confirm",
and mark them with a recognizable error msg,
    "Jira 329 -- Fatal MPP reservation error",
for easier triage.


awallace-cray:
This is optional- your call.

vasslitvinov:
I say let's go. However I suggest to let @ronawho weigh in.

PR #8389, opened 2018-02-06
User programs can build with CHPL_COMM=ofi

gbtitus:
Make some relatively minor changes to allow the runtime and user programs to build with CHPL_COMM=ofi:
- Remove includes of the now-nonexistent memory warning macros include files, and recode where those were needed.
- For now, use the default of no registered shared heap.
- Fix a misnamed function definition.
- With Cray PrgEnv target compilers, default to dynamic linking to avoid a couple of link-time warnings.

### DIFF
--- a/make/compiler/Makefile.cray-prgenv
+++ b/make/compiler/Makefile.cray-prgenv
@@ -47,8 +47,8 @@ LIB_STATIC_FLAG =
 LIB_DYNAMIC_FLAG = -shared
 
 #
-# Circumstances may push us to choose dynamic linking instead of the
-# PrgEnv default static linking.
+# Normally we let the PrgEnv compilers choose the linking mode, but in
+# some circumstances we push dynamic linking as the default.
 #
 ifneq ($(CHPL_MAKE_LOCALE_MODEL), flat)
 # With locale models other than 'flat' we will need libnuma.  That is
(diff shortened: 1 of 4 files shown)


make/compiler/Makefile.cray-prgenv
```diff
@@ -46,10 +46,19 @@ GEN_DYNAMIC_FLAG = -dynamic
 LIB_STATIC_FLAG =
 LIB_DYNAMIC_FLAG = -shared
 
-# With locale models other than 'flat' we will need libnuma.  That is
-# so far available only in .so form, but the PrgEnv compilers default
-# to forcing static linking.  So, override them in this case.
+#
+# Circumstances may push us to choose dynamic linking instead of the
+# PrgEnv default static linking.
```
ronawho: The default link method is site-specific, so maybe change the wording a little bit?

make/compiler/Makefile.cray-prgenv
```diff
@@ -46,10 +46,19 @@ GEN_DYNAMIC_FLAG = -dynamic
...
 ifneq ($(CHPL_MAKE_LOCALE_MODEL), flat)
+# With locale models other than 'flat' we will need libnuma.  That is
+# so far available only in .so form, so override them in this case.
+RUNTIME_LFLAGS += -dynamic
+else ifeq ($(CHPL_MAKE_COMM), ofi)
+# libfabric calls dlopen() and getaddrinfo().  With static linking, the
+# references to these cause link-time warnings that the corresponding
+# dynamic libraries will be required at execution time.  With dynamic
+# linking we avoid these warnings.
 RUNTIME_LFLAGS += -dynamic
```
ronawho: Not required for this PR, but we should see what other groups are doing about this. We know dynamic linking on Crays can lead to non-trivial performance regressions, and it seems hard to believe other groups will rely on full dynamic linking for ofi

make/compiler/Makefile.cray-prgenv
gbtitus: Ah, I'd forgotten that detail and the original comment didn't mention it to remind me!  I'll change the new comment.

PR #8390, opened 2018-02-06
Stop normalizing ContextCallExprs

lydia-duncan:
ContextCallExprs are generated at resolution time, but default initializers were
running into a situation where we copied a ContextCallExpr into a new function
that we then normalized and tried to resolve.  Prior to this change,
normalization would replace the contents with call_tmps, but that confused
ContextCallExprs when we tried to retrieve the expected type from it.  This
change prevents the creation of those call_tmps, so the ContextCallExpr is
handled as normal.

This change resolves an issue with default initializers where a field which
was supposed to be initialized from a global array's contents was generating
a "unknown type" error.

Adds two tests to lock in the fixed behavior for default initializers.
The first is a copy of the test which discovered this error, the second is a
variation so that the access is not just copied wholesale (as we want the
computation to be handled well in all contexts).  Both failed prior to this
change.

Passed classes/initializers with futures, a full std/ paratest, and no new
errors on a paratest with --force-initializers


lydia-duncan:
I suspect it wouldn't be accepted, but I can't say with certainty.  We would for sure encounter the same error I got if a ContextCallExpr whose last option was a SymExpr was used as the rhs of a move, but other situations may allow it

PR #8391, opened 2018-02-06
Add 'top' information for compiler timeouts

bradcray:
This simple change permits 'timedexec's output (which includes the top
processes running) to be printed in the event of a compilation timeout.
Previously we didn't print anything in this case beyond the fact that
the timeout had occurred.


bradcray:
@ronawho: Would you give this a quick review?  Also let me know what you'd suggest doing in terms of testing.  I made sure it printed out the 'top' output in the event of a compilation timeout and didn't in the event that it didn't time out, but haven't run it over the full test suite or anything like that (since it seems pretty clearly orthogonal).

PR #8397, opened 2018-02-07
BUILD_VERSION_FILE runs updateBuildVersion if it exists

mppf:
This can be viewed as a follow-on to PR #7498.

The old command (looking for .git directory) does not work for me since I
use git worktree. In a git worktree setup, $CHPL_HOME will be a git
checkout and respond to git commands like 'git log', but it contains
a .git file rather than a .git directory. That causes failures with
test/compflags/bradc/printstuff/version.chpl (among others) because the
BUILD_VERSION file contains "0" instead of a git hash.

So, this commit changes the relevant Makefile snippet to check for
the $CHPL_HOME/.git in file or directory form. That solves
the problem in my git worktree configuration.

- [x] full local testing

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -134,7 +134,7 @@ UPDATE_BUILD_VERSION = $(CHPL_MAKE_HOME)/util/devel/updateBuildVersion
 CLEAN_TARGS += $(CLANG_SETTINGS_FILE)
 
 $(BUILD_VERSION_FILE): FORCE
-	@({ test -x $(CHPL_MAKE_HOME)/.git ; } && \
+	@({ test -x $(UPDATE_BUILD_VERSION) ; } && \
 	test -x $(UPDATE_BUILD_VERSION) && $(UPDATE_BUILD_VERSION) $@) || \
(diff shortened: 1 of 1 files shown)


mppf:
@bradcray - do you have a moment to look at this PR making a 1-line-change in a Makefile? Thanks.

bradcray:
This looks acceptable to me, though it seems like we could do better.  Specifically, I don't think we'd want to have `updateBuildVersion` run if it were accidentally packaged into the release in some future change because having BUILD_VERSION be anything other than "0" causes the `--version` output to mark it as a pre-release.  Moreover, if `updateBuildVersion` is run outside of a git context, it won't work as intended.  So if I'm understanding correctly your situation correctly, rather than taking this approach, I'd suggest checking for the existence of `.git` (`test -e`) rather than checking for execute permissions (which looks like it was a poor-man's way of testing for a directory that dates back to the days when we were using svn).

mppf:
@bradcray - OK, I've done that. You can check the current version of the PR if you like. I'll still run testing on it before merging.

bradcray:
Looks good to me if it works for you.

PR #8361, opened 2018-02-02
Enable type inference from 'new' within initializers

benharsh:
This PR leverages existing code to allow initializers to handle type inference from a 'new' expression. Previously we would see a PRIM_NEW and emit an internal error, presumably in order to make progress in other initializer tasks.

A new compiler error is added for the case where a type field is initialized from a 'new' expression.

Related futures have been updated to avoid copy initializer confusion by specifying the type of the generic argument to 'init'.

Closes #7341

Testing:
- [x] full local

PR #8398, opened 2018-02-07
Fix 'new' expression bugs in phase 1

### LINKED ISSUE #8388: type-from-new param variations
### Summary of Problem

I saw the comment in #8361 (type inference from ``new`` within
initializers) about a special case for ``type`` fields, and was
curious what happens with a ``param`` field.  A few things happen.

### Steps to Reproduce

**Source Code:**

I started with the test added there, [test/classes/initializers/generics/phase1/type-from-new.chpl](https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/generics/phase1/type-from-new.chpl )

```chapel

record R {
  var x : int;
}

class C {
  type t;
  var v = new R();

  proc init() {
    t = new R();
    super.init();
    v = new R();
  }
}

proc main() {
  var c = new C();
  delete c;
}
```

Compiling this produces the desired error: ``Cannot initialize type
field 't' with 'new' expression``.

1. But if I change ``type t`` to ``param t``,

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl param-from-new.chpl 
param-from-new.chpl:12: internal error: seg fault [misc.cpp:558]
Note: This source location is a guess.
```

That shouldn't compile, but shouldn't be a segfault.

2. The same error occurs with ``var t``, which AFAIK should compile.

I removed the field ``v`` and its initialization in ``C.init()`` phase2.  Both the ``param`` and ``var`` versions compile that way.

3. However, I think the ``param`` version shouldn't compile.

4. Curious about what ``c`` had ended up looking like in the param, "no v" case, I added ``writeln(c)`` to ``main()``, and got

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl param-from-new-no-v-writeln.chpl 
In file included from /tmp/chpl-fortytwo-22186.deleteme/_main.c:28:0:
/tmp/chpl-fortytwo-22186.deleteme/ChapelIO.c: In function ‘writeThisFieldsDefaultImpl’:
/tmp/chpl-fortytwo-22186.deleteme/ChapelIO.c:310:13: error: ‘tmp_chpl’ undeclared (first use in this function)
 call_tmp7 = tmp_chpl;
             ^
/tmp/chpl-fortytwo-22186.deleteme/ChapelIO.c:310:13: note: each undeclared identifier is reported only once for each function it appears in
/home/fortytwo/src/chapel/runtime/etc/Makefile.exe:27: recipe for target '/tmp/chpl-fortytwo-22186.deleteme/param-from-new-no-v-writeln.tmp' failed
make: *** [/tmp/chpl-fortytwo-22186.deleteme/param-from-new-no-v-writeln.tmp] Error 1
error: compiling generated source
```

None yet.

### Configuration Information

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl --version
chpl version 1.17.0 pre-release (48805b1)
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)
```
```
fortytwo@magrathea:~/src/chpl/tfn$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none
```
```
fortytwo@magrathea:~/src/chpl/tfn$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.6) 5.4.0 20160609
```

benharsh:
This PR fixes two bugs:

1) Fixes compilation segfaults for type-inferred fields.
2) Adds error message for initializing a param field from a 'new' expression

The compiler could segfault if it attempted to insert initialization
implicitly for a type-inferred field. A small example:

```chpl
class C {
  var t = new R();
  proc init() {}
}
```

The compiler will try to insert something like 'this.t = new R()' within
the initializer body.

A couple of branches in ``InitNormalize::initializeFieldsBefore`` could
call functions if the exprType was NULL, but the field's type was not
dtUnknown. However, those functions relied on exprType and so we ran
into segfaults. The solution is to remove the check against dtUnknown
and to let other existing branches handle type inference.

The new error message expands on the work in #8361 , which introduced
a similar error message for type fields.

Closes #8388.

Testing:
- [x] full local


test/classes/initializers/generics/phase1/param-from-new.chpl
```diff
@@ -0,0 +1,20 @@
+
+record R {
+  var x : int;
+}
+
+class C {
+  param t;
```
lydia-duncan: What happens if we do the same thing but via omitted initialization for the param field? E.g.
```chapel
class C {
  param t = new R();
  var v = new R();

  proc init() {
    v = new R();
    super.init();
  }
}
```

benharsh: That case is also caught by this PR.

lydia-duncan: Excellent!  I figured it would be.  Your call whether to add a test for it or not

PR #8404, opened 2018-02-07
Solve "tuple size must be static" issue for default initializers

lydia-duncan:
Prior to this change, the _new function for classes with default initializers
was failing to treat param fields in the appropriate manner to specify that
a type field should be a homogeneous tuple.  That is, it was failing to pick
up the substitution information.  This change allows it to preserve that
information.

Add a copy of the test which triggered this failure for default initializers.

Passed classes/initializers with futures, a std/ paratest and no new
failures for a full paratest with --force-initializers


lydia-duncan:
@benharsh, @noakesmichael - would one of you mind reviewing this?

benharsh:
Isn't ``classes/bradc/genericTypes`` a related future that could be retired?

You mentioned elsewhere that this doesn't help user-written initializers. Could you add a test against that failure mode?

Otherwise, this looks good to me.

benharsh:
Oops, I think ``classes/bradc/genericTypes`` was just a failure we say when throwing ``--force-initializers``, so nevermind.

lydia-duncan:
I did originally write a test for the user initializer, but ended up thinking it was a style no one would ever write:
```chapel
proc init(param p: int, type t1; type t2: p*t1 = p*t1) { ... }
```
I would expect a user to either only use the colon specification (in which case they won't run into this particular case because they must provide a value, but it's kinda silly to have an argument that can't accept anything else), or to only use the default value specification (which works), not both.

But I'm fine with adding the test anyways if you still think that is worthwhile

PR #8405, opened 2018-02-08
Add .bad file for classes/initializers/assign-param-segfault future

cassella:
This is the current output of the test: writeln() of an instantiation of a generic class doesn't include the param member.


cassella:
I ran across this future while looking for existing reports of the problem I was having in type-from-new.chpl.

I don't understand what the ``.future`` is about, but I don't think either of the problems it describes are happening.  The output doesn't match the ``.good`` file though.

However, it does match how master behaves today for a default-constructed class:

```chapel
class C { param x = 7; var y = 42; }

var c = new C(17);

writeln(c);

delete c;
```
```
fortytwo@magrathea:~/src/chpl/tfn$ ./wrparmem -nl 1
{y = 42}
```

So maybe the ``.good`` is bad, the ``.skipif`` should be skipped, and the ``.future`` is now?

I don't know, so I at least updated the ``.bad`` in order to call attention to it in case the output changes.

lydia-duncan:
This test is a copy of a constructors future.  We have not been adding .bads religiously to initializers tests because it is an undue burden to continuously update them as the implementation has evolved.  Once the implementation is more finalized, we will probably want to go through the futures and add .bads then, but not until that time.

PR #8407, opened 2018-02-08
Add simple string.size

### LINKED ISSUE #8395: introduce string.size
In a discussion on an internal mailing list, we think that we want to introduce the method `string.size` by analogy to `array.size`, `range.size` `tuple.size` and so on.

There is some ambiguity in what 'size' could mean for different string encodings, though.

nimitbhardwaj:
Just add simple size function, so to add analogy with other data containers

Closes #8395


nimitbhardwaj:
I will add the tests for it, but please guide, where to put them

nimitbhardwaj:
Okk, no problem, I didn't know it would be better to talk of it first. :smiley:

nimitbhardwaj:
Sir, please for this time, can you deal with this issue, or give to someone else, because this time I am trying to apply for gsoc, because I have to do internship this time for college grade, so, otherwise I surly will do the work but please if needed immediately, then you should give it to someone else.

I will come back after 12 march for sure.
Apologies.

nimitbhardwaj:
Or I do it, not a big issue this is now, so I do it

nimitbhardwaj:
Where is the test located for .length

nimitbhardwaj:
Done, may be it was a simple test, if anything to edit here, tell me I will do that.

test/types/string/diten/stringSize.chpl
```diff
@@ -0,0 +1,5 @@
+var str1 = "hello";
+var str2 = "world";
+var str3 = "ell";
```
ben-albrecht: This works.. Can you add a string like `йцы` to track the UTF-8 behavior too?

nimitbhardwaj: OK sure.

nimitbhardwaj: what should be size, its giving 6

nimitbhardwaj: I don't know this language, but it seems like the size should be 3 or 4, I don't know of it, what should be ans.

nimitbhardwaj: yes, 3 may be, python3 is saying 3

ben-albrecht: > what should be size, its giving 6

That's the current behavior, since it's assuming ascii string. This test will lock that behavior in for now.

nimitbhardwaj:
if the answer is 3 then, we have to interpret the unicode strings, this is given by a library https://www.gnu.org/software/libunistring/manual/libunistring.html
May be it can help

test/types/string/diten/stringSize.chpl
ben-albrecht: One more - I'd suggest adding a `str1.size == str1.length` check to ensure these fields remain the same in the future.

nimitbhardwaj: OK, I do it

ben-albrecht:
> if the answer is 3 then, we have to interpret the unicode strings, this is given by a library 

This is discussed in #8395. This behavior may change in the future, but currently the strings are assumed to be ascii.

nimitbhardwaj:
Its done

nimitbhardwaj:
But if there is need of modification in the implementations, I would like to help too, but these days I am buzy to get a seat in GSoC, because I have to get one for my college grades, I would help more, I liked working here.

Apologies.
Please don't mind

ben-albrecht:
Going to merge after running full test suite.

PR #8409, opened 2018-02-08
Add annotation for recent machine upgrade

lydia-duncan:


test/ANNOTATIONS.yaml
```diff
@@ -289,6 +289,8 @@ all:
   01/25/18:
     - text: Do nonblocking responses in the polling task (#8297)
       config: 16 node XC
+  02/07/18:
+    - machine operating system update
```
ronawho: Mind making this a chapcs only annotation? (that'll apply to all perf configs that run on chapcs machines)

Super unimportant/trivial nit -- maybe change to "Machine upgraded to SLES12" to match what we've done CLE upgrades?

PR #8415, opened 2018-02-08
Keep track of the line number when handling multi-line strings

daviditen:
Multi-line strings weren't bumping the line number when a literal newline was
encountered.  This made the line numbers reported in error messages incorrect
when using multi-line strings.  Call `processNewline()` to handle newlines
inside of multi-line strings.

Add a test that got an incorrect line number in an error prior to this change.


daviditen:

daviditen:
Thanks @lydia-duncan!

PR #8417, opened 2018-02-08
Make in intent similar to variable initialization

### LINKED ISSUE #8416: is in task intent racey?
I'm working on a PR (#8417) that adjusts the `in` argument to perform the copy at the call site rather than in the called function. This effort led me to the realization that the current task intent implementation, for `in` intent, piggybacks on the compiler's existing argument intent implementation, which adds the copy to the body of the called function. That has the result of creating potentially surprising behavior, since a user might imagine that if the `in` task intent is used, it's not possible to race on that variable. But it is possible to have a race there, since the copy happens within the task.

Below is a complete program that reproduces this problem. On master, this consistenly prints out something like:
```
serial has 0 mismatches
begins has 6007 mismatches
```

but on my in-progress in-intent branch (PR #8417), which moves the copy to the call site (i.e. out of the task function and instead performs the copy in the process of creating the task argument bundle), it prints out:
```
serial has 0 mismatches
begins has 0 mismatches
```

The question is, do we consider the below program a wrong program? (If so, what should the user have written instead to synchronize correctly?).

``` chapel
class Impl {
  var x:int;
}

record R {
  var ptr:Impl;
}

proc R.init() {
  this.ptr = new Impl(0);
}

proc R.deinit() {
  delete this.ptr;
}

proc R.init(from:R) {
  this.ptr = new Impl(from.ptr.x);
}

proc =(ref lhs:R, rhs:R) {
  lhs.ptr = new Impl(rhs.ptr.x);
}

proc R.increment() {
  this.ptr.x += 1;
}


config const n = 10000;
config const debug = false;

proc report(phase, hits) {

  var mismatches = 0;
  for i in 0..#n {
    var count = hits[i].read();
    if count != 1 {
      mismatches += 1;
      if debug then
        writeln(phase, " hits count ", i, " is ", count, " not 1!");
    }
  }

  writeln(phase, " has ", mismatches, " mismatches");
}

proc test() {

  {
    var hits:[0..n] atomic int;

    var r:R;

    // First, run the pattern in a serial loop
    for i in 0..#n {
      hits[min(n,r.ptr.x)].add(1);
      r.increment();
    }

    // Report any hits != 1
    report("serial", hits);
  }

  {
    var hits:[0..n] atomic int;

    var r:R;

    // Now, run it with a begin per iteration
    sync {
      for i in 0..#n {
        begin with (in r) {
          hits[min(n,r.ptr.x)].add(1);
        }
        r.increment();
      }
    }

    // Report any hits != 1
    report("begins", hits);
  }
}

test();
```

mppf:
This PR makes `in` intent similar to variable initialization, in terms of when copy initialization occurs. In particular copy initialization is not needed for the result of a function call, e.g.

``` chapel

  record R { var x:int; }
  proc makeIt() { return new R(1); }
  proc returnIt(in x) { return x; }

  proc test() {
    var x = returnIt(makeIt()); // no copy is necessary here
  }
```

But note that a copy would be required were another variable, say, passed in to the `in` intent argument:
``` chapel
  proc test2() {
    var a = makeIt();
    var x = returnIt(a); // a copy is necessary here
  }
```

This PR achieves this change by handling the copy-initialize sometimes required for `in` intent at the call site. It leaves the old implementation in place for initializers and constructors though, since these have historically had special cases. The function `shouldAddFormalTempAtCallSite` can be adjusted to choose between the old and new `in` intent implementations (it returns `true` for the new implementation).

This PR does not adjust `out` or `inout`. While a similar adjustment might make sense for them, it is left as future work.

Closes #8416.

- [x] full local testing 
- [x] gasnet testing
- [x] baseline testing


mppf:
Gasnet Testing result:

Everything passed except for functions/vass/setter-passed-to-another-fun, which fails reliably.

mppf:
Valgrind testing results:
[Error matching program output for library/packages/ZMQ/interop-py/server]
[Error matching program output for memory/figueroa/LeakedMemory6]
[Error: Timed out executing program parallel/begin/deitz/test_global_for_begin]
[Error: Timed out executing program parallel/bundles/bundles]
[Error matching compiler output for regexp/elliot/I-fail-without-re2]
[Error matching program output for studies/comd/llnl/CoMD]

mppf:
These valgrind failures are from my misconfiguration in testing (nightly sets `export CHPL_RT_NUM_THREADS_PER_LOCALE=100`) or are issues currently present on master.

Issue #8406, opened 2018-02-08
Initializers allow initializing a param field with a non-param value


cassella:
### Summary of Problem

```chapel
class C {
  param p: int;
  proc init() {
    var x = 7;
    p = x;
    super.init();
  }
}

var c = new C();
writeln(c);
delete c;
```

[edit 2/9 cassella -- simplify test case to no longer use a record. ]

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl param-from-record.chpl 
In file included from /tmp/chpl-fortytwo-8306.deleteme/_main.c:28:0:
/tmp/chpl-fortytwo-8306.deleteme/ChapelIO.c: In function ‘writeThisFieldsDefaultImpl’:
/tmp/chpl-fortytwo-8306.deleteme/ChapelIO.c:237:13: error: ‘call_tmp_chpl’ undeclared (first use in this function)
 call_tmp7 = call_tmp_chpl;
             ^
/tmp/chpl-fortytwo-8306.deleteme/ChapelIO.c:237:13: note: each undeclared identifier is reported only once for each function it appears in
/home/fortytwo/src/chapel/runtime/etc/Makefile.exe:27: recipe for target '/tmp/chpl-fortytwo-8306.deleteme/param-from-record.tmp' failed
make: *** [/tmp/chpl-fortytwo-8306.deleteme/param-from-record.tmp] Error 1
error: compiling generated source
```

I'd have expected an error on the ``p = x`` line since ``x`` isn't a param.

[ edit again 2/9 cassella -- adjust the expected behavior too. ]

**Associated Future Test(s):**
None yet.

### Configuration Information

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl --version
chpl version 1.17.0 pre-release (ee5378d)
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)

fortytwo@magrathea:~/src/chpl/tfn$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
[...]

PR #8421, opened 2018-02-09
new future: init() initializing a param field with a non-param value

cassella:
This should fail to compile.

This is the future for issue #8406

### DIFF
--- a/test/classes/initializers/generics/init-param-from-non-param.chpl
+++ b/test/classes/initializers/generics/init-param-from-non-param.chpl
@@ -1,8 +1,7 @@
-config var notaparam = 77;
-
 class C {
   param p:int;
   proc init() {
+    var notaparam = 77;
     p = notaparam;
     super.init();
   }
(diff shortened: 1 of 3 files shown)


test/classes/initializers/generics/init-param-from-non-param.chpl
```diff
@@ -0,0 +1,13 @@
+config var notaparam = 77;
```
lydia-duncan: Not a bad thing, but I'm curious why you went with this program as opposed to using the local var you mentioned in the original issue

cassella: In the original issue, I was hung up on the ``record`` type, and the local var was just there to separate the param initialization from the ``new R()``.  When I found the issue applied to ``int`` as well, I used a ``config var`` so I didn't have to contemplate whether the compiler believes the value below is "known at compile time" as the error message says:

```chapel
var notaparam = 7;
param p = notaparam;
```

With a config var, there's no sense in which it's known at compile time.

lydia-duncan: I can see that logic, but a local variable is a more basic case (assuming it still causes the same failure) and really should be treated reasonably similarly by the compiler.

cassella: Ok.  I'll try making that change tonight.

PR #8427, opened 2018-02-09
Restore gcd() future about lack of return from a function that always does

bradcray:
This future started passing inadvertantly last night because it
uses a very generic function, gcd(), which was considered a
worse match than the new gcd() library function.  The future
is still relevant, though, so I just renamed the function to
keep it working as it was.


bradcray:
Hi @lydia-duncan : You futurized this (appropriately) some time ago -- would you give this a quick review?

PR #8428, opened 2018-02-09
cleanup inTree()

vasslitvinov:
It used to be the case that the presence (non-NULL-ness)
of parentSymbol was not enough to decide Expr::inTree().

#### This PR

* establishes Expr::inTree() === (Expr::parentSymbol != NULL)

* adjusts Expr::inTree() correspondingly,

* replaces a few explicit checks for "parentSymbol != NULL"
  with calls to inTree(), for clarity.

The only case where this discipline did not hold was for ModuleSymbol::block.
I adjusted that, almost trivially.

While there, I made "this == rootModule" the second check in Symbol::inTree().
Because we expect the now-first check to succeed in most cases and
the "this == rootModule" check to fail almost always.

While there, I made {Expr,Type,Symbol}::inTree() non-virtual,
as we do not override them in subclasses.

#### Other potential improvements

* There may be other cases where we could replace "parentSymbol != NULL"
with inTree().

* Add INT_ASSERT(symbol != NULL) in Type::verify().

* Replace isAlive() with inTree().

The only reason not to do that that I see is that isAlive() executes
slightly fewer conditionals than inTree(). This could be mitigated:

  - To avoid the check for rootModule, we can add another version of
    Symbol::inTree(). We could apply that version whenever we do not need the
    precise answer for rootModule, ex. when the Symbol is not a ModuleSymbol.
    Maybe Symbol::inTree() could be virtual, overriden for ModuleSymbol.

  - To avoid the check for Type::symbol!=NULL, we could ensure that
    it is always the case. Add INT_ASSERT(symbol!=NULL) to Type::verify().

Testing: linux64 --verify, gasnet, baseline.

### DIFF
--- a/compiler/AST/ModuleSymbol.cpp
+++ b/compiler/AST/ModuleSymbol.cpp
@@ -35,7 +35,7 @@ ModuleSymbol*                      standardModule        = NULL;
 ModuleSymbol*                      printModuleInitModule = NULL;
 
 Vec<ModuleSymbol*>                 userModules; // Contains user + main modules
-Vec<ModuleSymbol*>                 allModules;  // Contains all modules
+Vec<ModuleSymbol*>                 allModules;  // Contains all modules except rootModule
 
 static ModuleSymbol*               sMainModule           = NULL;
 static std::string                 sMainModuleName;
@@ -213,8 +213,6 @@ ModuleSymbol::ModuleSymbol(const char* iName,
   doc                 = NULL;
   extern_info         = NULL;
   llvmDINameSpace     = NULL;
-
-  block->parentSymbol = this;
 
   registerModule(this);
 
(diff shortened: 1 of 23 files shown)


vasslitvinov:
@benharsh would you review?

benharsh:
> It used to be the case that the presence (non-NULL-ness)
of parentSymbol was not enough to decide Expr::inTree().

What changed that lowers the bar? What if I called ``inTree()`` on an expr in a function that was not in the tree?

vasslitvinov:
There was only one case where parentSymbol is non-NULL for something not inTree(). That was for BlockStmts stored in ModuleSymbol::block fields. I fixed that, so now no such cases remain. I ran a paratest with an assertion in Expr::inTree() to confirm that.

When some node is removed from the tree, whether it is `call->remove()` or `fn->defPoint->remove()`, the `parentSymbol` field on all sub-nodes is reset to NULL. I just double-checked that.

benharsh:
OK, looks good to merge.

PR #8429, opened 2018-02-09
Do not lowerIterator() on dead parallel iterators

vasslitvinov:
After we inline iterators, most iterators become unused.
So there is no reason to run lowerIterator(fn) on these iterators.

Instead, now we delete them from the tree.

The "unused" check is that the FnSymbol has no SymExprs referring to it.

This "unused" check does not work on serial iterators.
This is probably because of the _toLeader/_toFollower/_toStandalone
business, where a parallel iterator - that may still be used -
needs to go back to its serial counterpart at some point.
So limit the "unused" check to parallel iterators.

The task functions invoked by an unused iterator could also
be removed. However, this change is simpler and does not do that.
If we revert the task functions back to BlockStmts or dedicated
non-FnSymbol AST nodes, then such removal will happen automatically.

I needed to bump the depth argument to `maybeCalled` because
the test `binary-tree-spawn.no-recurse.chpl` now exposes dead-yet-inTree
task functions that are 5-deep call stack from the iterator invoking them.

Testing: linux64 --verify, gasnet, baseline.

### DIFF
--- a/compiler/resolution/lowerIterators.cpp
+++ b/compiler/resolution/lowerIterators.cpp
@@ -2365,8 +2365,13 @@ static void cleanupTemporaryVectors() {
 }
 
 
+//
 // 'depth' is a heuristic to avoid the risk of unbounded recursion.
 // Ex. what if 'parentSym' is a recursive function?
+// When 'parentSym' is a task function, 'depth' should correspond at least
+// to how many task constructs deep the yield call is in the iterator.
+// Ex. test/parallel/forall/vass/binary-tree-spawn.no-recurse.chpl
+//
 static bool maybeCalled(Symbol* parentSym, int depth) {
   if (!parentSym || !parentSym->inTree())
     return false;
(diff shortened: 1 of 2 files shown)


vasslitvinov:
@mppf or @benharsh would you review?

compiler/resolution/lowerIterators.cpp
```diff
@@ -2509,10 +2514,16 @@ void lowerIterators() {
 
   forv_Vec(FnSymbol, fn, gFnSymbols) {
     if (fn->isIterator()) {
+     if (fn->firstSymExpr() == NULL &&
```
mppf: I think this section needs a comment describing the general goal of this snippet, like "Remove parallel iterators that are not used"

mppf:
In your PR message, it's talking about removing dead parallel iterators, but can't a live serial iterator cause a parallel iterator to appear, with toFollower / toLeader / toStandalone? The PR message seems to indicate it can only go the other way - from parallel to serial - but these details are rather unclear to me. Can you investigate further / better explain the situation?

Besides that, the code actually removes dead iterators that are marked with FLAG_INLINE_ITERATOR. It's not obvious to me what the relationship between FLAG_INLINE_ITERATOR and being parallel is.

Without a better explanation for why you'd need to leave in serial iterators, I'm left thinking that it just happens to be the case that this change works with the AST the compiler is currently producing, but that to be "right" we'd need to decide that iterators returned by toLeader etc. also count as a "use".

But it's fairly likely that I'm not seeing something that you are.

mppf:
I should also say, I've pretty much always wished that we could replace toFollower/toLeader/toStandalone with some more compiler smarts. I'm continually confused by them and usually respond my making them a special case in compiler code that should generally apply.

vasslitvinov:
Closing this until I have a chance to investigate further, as requested by the reviewer.

Issue #6019, opened 2017-04-13
Permit promotion of constructors?


bradcray:
Given a class and (optionally) constructor like this:

````chapel
class C {
  int id;
  ...
  proc init(initID) {
    id = initID;
  }
}
````

Should we be permitted to create an array of such objects by promoting the constructor as follows?

````chapel
var arrOfC = new C(1..10);
````
(It does not work as of 1.15.0.2781929, generating the message:

````
error: type mismatch in assignment from _ir__construct_C to C
````

**Comment by benharsh** [comment]
I think we should also consider how promotion might interact with copy initializers. Let's consider the following sketch of a program:

```chpl
class C {
  type eltType;
  var x : eltType;

  proc chpl__promotionType() type return eltType;

  iter these() { for i in 1..10 do yield i*x; }

  proc init(x : int) { /* whatever */ }
  proc init(x : real) { /* whatever */ }

  proc init(other:C(int)) { /* whatever */ }
}

var cr = new C(1.0); // C(real)

var q = new C(cr); // compile-time error, or promote to init(x:real) ?
```

I think that there are at least two interpretations of the initialization of ``q``:
1) Allow promotion over C.init(x:real).
2) Compile-time error that a copy initializer was not found.

I think allowing promotion in this case would confuse most users, who probably expect to be calling a copy initializer. A compile-time error would require users who intentionally wanted promotion to write something more explicit (``forall`` or ``[]`` expr). I think this is a tradeoff worth making to prevent users from accidentally encountering promotion.

cassella:
It works now, if the initializer's arg isn't generic:

```chapel
class C {
  var id: int;

  proc init(initID: int) {
    id = initID;
  }
}

var arrOfC = new C(1..10);

writeln(arrOfC.type:string);
writeln(arrOfC);
```

```
fortytwo@magrathea:~/src/chapel (master)$ ./promoinit -nl 1
[domain(1,int(64),false)] C
{id = 1} {id = 2} {id = 3} {id = 4} {id = 5} {id = 6} {id = 7} {id = 8} {id = 9} {id = 10}
```

cassella:
Granted, the specific question was "should this be allowed?", and in retrospect I don't know if the generic arg to the initializer was specifically part of the question.

FWIW, @benharsh's example (modified to have each `init()` set `this.eltType` before doing whatever) does fail, but I think due to not applying the `.these()` (?):

```
[...]

PR #8431, opened 2018-02-10
Support promotion on types with initializers

benharsh:
The '_promotionType' field was used to indicate and implement promotion over a
class or record. This was a hacky approach likely implemented long ago to make
progress. More recently, this field complicates the implementation and
transition for the ongoing initializers work.

Ideally we would infer promotion support by resolving the yield type of the
'these' iterator. This is not easy in the current compiler architecture. A
simpler approach is to attempt to resolve a method that returns a type. If
resolvable, promotion is enabled.

Constructors and initializers now use a common function,
'resolvePromotionType', to infer support for promotion. This function assumes
that the given type is non-generic and that it does not already have an
associated promotion type. In order to support promotion over 'this' within
initializers, the compiler will attempt to infer the promotion type before the
phase 2 portion of an initializer has been resolved.

Types that use generic initializers are fully instantiated within
'resolveInitField', at which point we can safely infer the promotion type.
Types with concrete initializers will have their promotion type inferred at the
start of 'resolveFunction'. Constructors will continue to infer promotion type
at the start of 'resolveTypeConstructor'.

Promotion is disallowed for calls that **look** like copy initialization. My rationale for this decision is that a user writing ``var x = new C(otherC);`` probably wants a copy initializer. This rule holds even if there is not a copy initializer for ``otherC`` (e.g. if ``otherC`` is an unsupported generic instantiation). Though this rule will likely prompt further discussion, I think this is the safest initial behavior and can be relaxed later.

Testing:
- [x] full local
- [ ] full gasnet
- [ ] baseline


benharsh:
If interested, I advocate that we discuss the copy-init promotion restriction over in #6019.

compiler/resolution/ResolutionCandidate.cpp
```diff
@@ -649,18 +670,36 @@ bool ResolutionCandidate::checkResolveFormalsWhereClauses() {
...
+    AggregateType* base  = getActualType(this, 1);
+    AggregateType* other = getActualType(this, 2);
+    if (base != NULL && other != NULL) {
+      if (base == other) {
+        looksLikeCopyInit = true;
+      } else if (getRootInstantiation(base) == getRootInstantiation(other)) {
+        looksLikeCopyInit = true;
+      }
+    }
+  }
```
mppf: It seems like we need a comment here indicating why copy inits get different treatment.
Also it seems likely that computing looksLikeCopyInit belongs in another function / method, but maybe it's very specific to this code path?

benharsh: Adding a comment sounds reasonable.

I think this check is very specific to this code path, but it wouldn't hurt to move it into a local utility function.

compiler/AST/AggregateType.cpp
```diff
@@ -1053,16 +1053,7 @@ void AggregateType::typeConstrSetFields(FnSymbol* fn,
...
-        Symbol*   _this     = fn->_this;
-        Symbol*   name      = new_CStringSymbol(field->name);
-        Expr*     exprType  = field->defPoint->exprType->remove();
-        CallExpr* call      = new CallExpr(PRIM_TYPE_INIT,  exprType);
-        CallExpr* setMember = new CallExpr(PRIM_SET_MEMBER, _this, name, call);
-
-        fn->insertAtTail(new BlockStmt(setMember, BLOCK_TYPE));
-
-      } else {
+      }  else {
```
vasslitvinov: Care to remove the extra space? (Unless you put it there for style.)

PR #8432, opened 2018-02-10
Link in libmvec.a when necessary

dmk42:
Gcc now needs a vectorized math library, libmvec.  This is made available automatically in most cases, but when linking statically on Linux, it needs to be specified explicitly on the link command line.

When using gcc, this change looks to see if libmvec is available, and if so, assumes it is needed whenever we are linking statically.

The \-lmvec flag for the vectorized math library must appear before the \-lm flag for the regular math library, according to https://sourceware.org/glibc/wiki/libmvec .

Tested on Ubuntu, SLES, Mac, and Intel/AMD/ARM-based Cray systems.


dmk42:
@ronawho - Mind taking a look at this?  I know you approved an earlier version, but I've now made the new link flag conditional on static linking, because it isn't needed otherwise.  I tack it onto the end of the LIBS variable in the makefile that gets built for each Chapel program as it is being compiled.

runtime/etc/Makefile.exe
```diff
@@ -28,7 +28,7 @@ $(TMPBINNAME): $(CHPL_CL_OBJS) checkRtLibDir FORCE
 ifneq ($(SKIP_COMPILE_LINK),skip)
 	$(CC) $(CHPL_MAKE_BASE_CFLAGS) $(GEN_CFLAGS) $(COMP_GEN_CFLAGS) -c -o $(TMPBINNAME).o $(CHPL_RT_INC_DIR) $(CHPLSRC)
 	$(foreach srcFile, $(CHPLUSEROBJ),$(CC) $(CHPL_MAKE_BASE_CFLAGS) $(GEN_CFLAGS) $(COMP_GEN_CFLAGS) -c -o $(srcFile) $(CHPL_RT_INC_DIR) $(srcFile).c ;)
-	$(LD) $(GEN_LFLAGS) $(COMP_GEN_LFLAGS) -o $(TMPBINNAME) -L$(CHPL_RT_LIB_DIR) $(TMPBINNAME).o $(CHPLUSEROBJ) $(CHPL_RT_LIB_DIR)/main.o $(CHPL_CL_OBJS) -lchpl -lm $(LIBS) $(CHPL_MAKE_THIRD_PARTY_LINK_ARGS) $(CHPL_MAKE_BASE_LFLAGS)
+	$(LD) $(GEN_LFLAGS) $(COMP_GEN_LFLAGS) -o $(TMPBINNAME) -L$(CHPL_RT_LIB_DIR) $(TMPBINNAME).o $(CHPLUSEROBJ) $(CHPL_RT_LIB_DIR)/main.o $(CHPL_CL_OBJS) -lchpl $(LIBS) -lm $(CHPL_MAKE_THIRD_PARTY_LINK_ARGS) $(CHPL_MAKE_BASE_LFLAGS)
```
gbtitus: We could probably even move the '-lm' clear to the end of our command line.  `libm` shouldn't reference any globals that aren't defined by other system libraries which the `$(LD)` driver (whatever it is) will add following everything mentioned explicitly on the command line.  More specifically, it shouldn't reference anything that will be defined by libraries appearing in `$(CHPL_MAKE_THIRD_PARTY_LINK_ARGS)` or `$(CHPL_MAKE_BASE_LFLAGS)`.  That said, it also shouldn't hurt to leave it where it is.

gbtitus:
Looks good!  I made a line comment but it only needs addressing if you want to and I don't need to look at it again if you do change it.

dmk42:
@gbtitus - Thanks.  I agree, though it would require another round of thorough testing to make sure I wasn't breaking an assumption somewhere else.  I'll keep this in mind, but for now, in the interest of resolving a triage issue, I'll merge it this way.

PR #8433, opened 2018-02-10
Add a scalable AllLocalesBarrier

ronawho:
Add a scalable AllLocalesBarrier and use it in ISX. 

The main scalability issue for ISx is that we use a barrier that lives on locale 0, and every other locale must do an on-stmt back to that locale to barrier locally. This is a huge bottleneck and significantly overloads locale 0 (at 256 nodes, we're creating ~10,000 tasks on locale 0, which will trigger an OOM with default call stack sizes.)

With this barrier, there are no execute-on's in ISx. There's only numLocales-1 execute-on-nb's for the initial `coforall loc in Locales do on loc ` , and everything else is just gets/puts.

The basic way it works is we do a local barrier (reusing our centralized busy-wait atomic barrier) on each locale, and the last task to enter the barrier will do a chpl_comm_barrier before releasing the barrier. This is much more scalable than the current approach, since chpl_comm_barrier is optimized/scalable (tree-based for ugni, and probably a tree or something for gasnet)

The scalability for this is good, and the overall performance is pretty high. The centralized busy-wait atomic barrier doesn't perform very well with high core-counts, but that's a fixed cost and doesn't impact locale-count scalability. 

Note, this PR is for discussion only. The individual commits will be broken out and merged separately (assuming the overall approach is ok'd)

### DIFF
--- a/modules/packages/AllLocalesBarriers.chpl
+++ b/modules/packages/AllLocalesBarriers.chpl
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2004-2018 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


ronawho:
@benharsh I'm curious to see what your impressions are. Ignore the runtime changes, I'll sort those out with Greg later, but I'm interested to see if you're ok with the AllLocalesBarrier API and the changes to ISx.

benharsh:
Changes to ISx look fine to me.

I'd slightly prefer a type named ``LocalesBarrier`` that took an optional array of locales. But this is 'just' a package, so I'm not too concerned.

ronawho:
Closing for now -- was prototype only to gain feedback

Issue #8422, opened 2018-02-09
classes allow param record fields


cassella:
### Summary of Problem

```chapel
record R { }

class C {
  param r:R;
  proc init() { }
}

var c = new C();
writeln(c);
delete c;
```

Without the ``writeln()``, this compiles.  With it,

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl wrparmem.chpl
In file included from /tmp/chpl-fortytwo-23814.deleteme/_main.c:28:0:
/tmp/chpl-fortytwo-23814.deleteme/ChapelIO.c: In function ‘writeThisFieldsDefaultImpl2’:
/tmp/chpl-fortytwo-23814.deleteme/ChapelIO.c:236:26: error: ‘tmp_chpl’ undeclared (first use in this function)
 writeIt_chpl(_ref_tmp_6, tmp_chpl);
                          ^
/tmp/chpl-fortytwo-23814.deleteme/ChapelIO.c:236:26: note: each undeclared identifier is reported only once for each function it appears in
/home/fortytwo/src/chapel/runtime/etc/Makefile.exe:27: recipe for target '/tmp/chpl-fortytwo-23814.deleteme/wrparmem.tmp' failed
make: *** [/tmp/chpl-fortytwo-23814.deleteme/wrparmem.tmp] Error 1
error: compiling generated source
```

I expected an error along the lines of the one you get declaring such a ``param r:R`` outside of a class:

```
param-record-field.chpl:4: error: 'r' is not of a supported param type
```
**Associated Future Test(s):**
[`test/classes/initializers/generics/param-record-field.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/generics/param-record-field.chpl)

[edit 2/12 cassella to add path to future]

### Configuration Information

```
fortytwo@magrathea:~/src/chpl/tfn$ chpl --version
chpl version 1.17.0 pre-release (44fa845)
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)

fortytwo@magrathea:~/src/chpl/tfn$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
[...]

PR #8440, opened 2018-02-11
future: classes allow param record fields

cassella:
Add the future test/classes/initializers/generics/param-record-field.chpl


cassella:
This is the future for issue #8422.

PR #8445, opened 2018-02-12
chip23 :- Update the chip to match recent work

noakesmichael:
Update the Chip for the initializer proposal to better match the latest view of this work.

### DIFF
--- a/doc/rst/developer/chips/23.rst
+++ b/doc/rst/developer/chips/23.rst
@@ -37,7 +37,7 @@ conclusions are
 The first concern is clearly the most pressing and it is the focus of
 this note.  We propose a key change to deliver the missing
 functionality, and then a few smaller changes to compensate for this
-change.  Beyond the fact that this proosal provides key functionality
+change.  Beyond the fact that this proposal provides key functionality
 that the current design does not, it is believed that it will be
 easier for some newer users to adopt.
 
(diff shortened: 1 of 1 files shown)


doc/rst/developer/chips/23.rst
```diff
@@ -375,243 +370,407 @@ How does this impact our current view of two-phase initialization?
...
+   var x = allocate(MyClass1);
+
+   x.init(...args...);
+   x.postInit();
+
+i.e. an instance is allocated and then the appropriate overload of the
+MyClass1.init() method is invoked.  When this method returns 'x' is a
+fully initialized instance of MyClass1.  Then the method postInit() is
+invoked.  If this call results in method calls with virtual overloads
+then the dispatch will be for objects of type MyClass1.
```
cassella: Hi @noakesmichael, I'm curious about some of the details here.

Particularly, how ``postInit()``s work with a class hierarchy.  For example the ``A -> B -> C -> D``, as in the "Two-phase initialization" section above.

If only ``A`` defines a ``postInit()``, then ``var d = new D()`` will as above do ``d.init(); d.postInit()``, and ``d.postInit()`` will resolve to the ``A.postInit()``.

But if multiple levels define ``postInit()``s, how is control handed off between them?  Would ``B.postInit()`` have a call to ``super.postInit()``?  If ``B.postInit()`` was defined without such, would the compiler insert one, declare it an error, or would control just not reach ``A.postInit()``?

If ``A`` and ``C`` define ``postInit()``, would ``B`` be required to define one to pass control along?  Or would the compiler insert one?  Or would ``C.postInit()``'s call to ``super.postInit()`` just work out to calling ``A.postInit()``?

Or would no ``postInit()`` invoke super.postInit()?  I could image the compiler, upon seeing ``var d = new D()``, could, as it knows the static type D and its static type hierarchy, insert the appropriate independent ``postInit()`` calls in some defined order, instead of just one ``d.postInit()``?

PR #8446, opened 2018-02-12
Prevent paratest.chapcs from running during nightly testing

ronawho:
Prevent paratest.chapcs from running between 11:30 PM and 2:30 AM PST when
nightly has exclusive access to machines. Nightly testing may run longer, but
this is the rough time frame from when multi-locale jobs will be running.


ronawho:
@benharsh mind reviewing?

PR #8448, opened 2018-02-12
Improve error messages for openfd failures

mppf:
In nightly testing, we've seen several errors of this form:
```
uncaught SystemError: 'file type not openable (in openfd with path "/")
```

This is probably coming from some sort of system / launcher issue, but the error message leaves a few open questions. This PR simply improves the error message to add more detail so we can better understand what is going wrong if it happens again. Additionally, it adds a test to make sure we get a reasonable error in this case.

- [x] full local testing

### DIFF
--- a/runtime/src/qio/qio.c
+++ b/runtime/src/qio/qio.c
@@ -769,12 +769,22 @@ qioerr qio_file_init(qio_file_t** file_out, FILE* fp, fd_t fd, qio_hint_t iohint
              ftype == S_IFBLK ) {
     // regular file or block device can seek
     seekable = 1;
+  } else if( ftype == S_IFDIR ) {
+    if( fd == 0 ) {
+      QIO_RETURN_CONSTANT_ERROR(EINVAL, "fd 0 (stdin) is a directory");
+    } else if( fd == 1 ) {
+      QIO_RETURN_CONSTANT_ERROR(EINVAL, "fd 1 (stdout) is a directory");
+    } else if( fd == 2 ) {
+      QIO_RETURN_CONSTANT_ERROR(EINVAL, "fd 2 (stderr) is a directory");
+    } else {
+      QIO_RETURN_CONSTANT_ERROR(EINVAL, "cannot openfd a directory");
+    }
+  } else if( ftype == S_IFLNK ) {
+    QIO_RETURN_CONSTANT_ERROR(EINVAL, "cannot openfd a symbolic link");
   } else {
-    // ftype == S_IFDIR
-    // ftype == S_IFLNK
     // ftype == S_IFWHT on Mac OS X
-    // can't open symlink/dir/whiteout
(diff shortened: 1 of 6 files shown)


bradcray:
This isn't a review, but after a quick glance at the PR:  If I'm understanding the mod correctly, we'll still see an "uncaught error" prefix in the output, which seems a little unfortunate in that the user didn't fail to catch anything (we didn't throw anything to them) which is what that phrase suggests to me.  Meanwhile,  we could've caught something in `modules/standard/IO.chpl` if we were to put the effort into it to spit out the error message without the "uncaught error" prefix.  Is that right?

mppf:
> Meanwhile, we could've caught something in modules/standard/IO.chpl if we were to put the effort into it to spit out the error message without the "uncaught error" prefix. Is that right?

I think you mean - if we put in the effort to write a try/catch in stdinInit - we could emit an error without the "uncaught error" prefix? (If so, I agree that would be possible).

I don't think the "uncaught error" prefix by itself is problematic in this case. The error was not caught & the default handler dealt with it (by halting). "uncaught error" doesn't necessarily mean that the user could or should catch the error. AFAIK Python (say) can print out a stack trace on a fatal error even if it happens before the main interpreted code begins. Besides all of that, I don't think this particular error is likely to come up much for users, so I'm hesitant to expend energy on word-smithing it to make it as friendly looking as possible.

bradcray:
> I think you mean - if we put in the effort to write a try/catch in stdinInit - we could emit an error without the "uncaught error" prefix? (If so, I agree that would be possible).

Right, and part of my motivation in asking, as with the email that led to this PR, was to put more weight on error-handling than we do at present (where I feel pretty confident `try!` works but less so about other things).  That said, this was also based on an assumption that "uncaught error" implies to the user that they shouldn't done something but didn't, which you disagree with:

> "uncaught error" doesn't necessarily mean that the user could or should catch the error.

I haven't worked much at all in languages with exceptions / error-handling, so will defer on this issue if that's the consensus.

bradcray:
(I also generally think that the presence of `try!` in module code is likely to indicate a quick fix for which something more should be done (including `try...catch...halt()`, but could be wrong about that as well...).

mppf:
> I also generally think that the presence of try! in module code is likely to indicate a quick fix for which something more should be done 

I generally agree with that statement, though. It's just that if some form of "halt" is the right answer, `try!` seems OK to me.

bradcray:
>  It's just that if some form of "halt" is the right answer, `try!` seems OK to me.

To me (again, as an instance of the unwashed non-exception-using masses), if we're curating an error message for the user in code we own, it seems odd to label it as an "uncaught error".  Intuitively, as an English speaker, that phrase suggests to me "nobody anticipated this" whereas putting a `try!` in suggests "we did anticipate it.  But again, if you guys think this is standard practice, I defer.

PR #8451, opened 2018-02-13
coforall/try! bug

psahabu:
@benharsh discovered a bug in error handling where a `try!` at the end of a `coforall` block allows unhandled throwing calls to escape within the `coforall`. For example:

``` chapel
// throwme() is unhandled, but this passes checking
proc test() {
  coforall i in 1..10 {
    writeln(i);
    throwme();
    try! { }
  }
}
```

The error ended up being in `ImplicitThrowsVisitor`, where `exitTryStmt()` would mark an entire function block as non-throwing if `try!` was the last statement. This was triggered by a `coforall` because they are lowered into a `coforall_fn()` during compilation.

This change required additional checking on `throw` statements in order to maintain correctness.

- [x] linux64+flat testing

### DIFF
--- a/compiler/passes/errorHandling.cpp
+++ b/compiler/passes/errorHandling.cpp
@@ -696,15 +696,14 @@ void ImplicitThrowsVisitor::exitTryStmt(TryStmt* node) {
 
   bool nonExhaustive = catchesNotExhaustive(node);
 
-  if (node->tryBang()) {
-    canThrow = false;
-  } else {
+  if (!node->tryBang()) {
     canThrow = nonExhaustive;
     if (nonExhaustive)
       onlyUnchecked = false;
     if (reasonThrows == NULL)
       reasonThrows = node;
   }
+  // try! does not change the throwing status of a block
 }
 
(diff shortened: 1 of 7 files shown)


mppf:
@psahabu - I'm not sure the logic is right here. In particular, looking at the code, it seems like this case:

``` chapel
// throwme() is unhandled, but this passes checking
proc test() {
  coforall i in 1..10 {
    try! {
      writeln(i);
      throwme();
    }
  }
}
```

wouldn't be handled right but it would have worked before. Could you add such a test & if it works, explain why?

psahabu:
@mppf - The case you provided works because `canThrow` is initialized to false when `ImplicitThrowsVisitor` is created. When the `try!` is exited, `canThrow` is not set due to the conditional checking if it's a `try!`.

However, this case does not work:

``` chapel
proc test() {
  coforall i in 1..10 {
    try! {
      try {
        writeln(i);
        throwme();
      }
    }
  }
}
```

I'll keep looking at this to straighten out the logic.

Issue #8385, opened 2018-02-05
Bad references to moved `multilocale.rst`


BryantLam:
### Summary of Problem
Erroneous multilocale executions print bad path to `multilocale.rst` that was moved to `$CHPL_HOME/doc/rst/**usingchapel**/multilocale.rst`. When fixed, some tests are also affected with hard-coded output expecting the old path.

### Steps to Reproduce
**Execution command:**
`./foo -nl 4`

**Output**
```
error: Only 1 locale may be used for CHPL_COMM layer 'none'
To use multiple locales, see $CHPL_HOME/doc/rst/multilocale.rst
```

### Configuration Information
```
chpl version 1.17.0 pre-release (81470ac9ba)
```

ben-albrecht:
Thanks for reporting!

mppf:
@cassella - I know this might sound strange, but would it be reasonable for chplspell to find such path issues?

cassella:
@mppf Nothing's occurring to me immediately.  Do you have an approach in mind?

mppf:
Well, paths could be "words" and a "dictionary" could be populated with paths that exist. Of course, that's all said from a place of having no idea how chplspell works... so it might not have a reasonable implementation there.

cassella:
It's not a natural fit for chplspell -- the scpell script tokenizes everything up front on spaces and punctuation including ``/``, so it's already "CHPL_HOME" "doc" "rst" "multilocale" "rst" before any spell checking can happen.

Though  for a first cut, only considering paths that are anchored by "CHPL_HOME/",

```
grep -n -I -o  -r CHPL_HOME[-_./a-zA-Z0-9]\* . | sed s,CHPL_HOME/,, | sed s,CHPL_HOME,, | sed s/:/\ /g | while read f l p; do [ -z "$p" -o -e "$p" ] || echo $f:$l names $p; done
```

There are a fair number of false positives, but also what look like a bunch of out-of-date docs.  (And CHANGES.md lines refering to files that no longer exist, I think.)  But it didn't pick out any more instances in runtime/ (or compiler/ besides an llvm RUN_CONFIG script.)

I'm not sure this can be made into something reliable enough to be smoke-testable.  Or if it'd be reasonable to look for e.g ``doc/`` ``compiler/`` ``test/`` ... as anchors.

PR #8456, opened 2018-02-13
Fix assorted dead links

cassella:
Fix assorted dead links

### DIFF
--- a/doc/rst/developer/bestPractices/TestSystem.rst
+++ b/doc/rst/developer/bestPractices/TestSystem.rst
@@ -433,9 +433,7 @@ associated GitHub issue number in the `#issue-number` format, e.g. `#1`.
 
 The rest of the file is optional and free-form. It can be used over the
 future's lifetime to describe in what way the test isn't working or should be
-working, implementation notes, philosophical arguments, etc.  The one-line
-summaries of all outstanding futures can be viewed by running
-``$CHPL_HOME/util/devel/test/list_futures``.
+working, implementation notes, philosophical arguments, etc.
 
 The current categories of futures reflect GitHub labels:
 
--- a/doc/rst/tools/chpldoc/chpldoc.rst
+++ b/doc/rst/tools/chpldoc/chpldoc.rst
@@ -77,11 +77,11 @@ As an example, there is a primer on the chpldoc capability in
 ``$CHPL_HOME/examples/primers/`` that can be turned into documentation using::
 
    cd $CHPL_HOME/examples/primers
-   chpldoc chpldoc.chpl
+   chpldoc chpldoc.doc.chpl
 
 The output documentation will be located in::
 
-   $CHPL_HOME/examples/primers/docs/chpldoc.html
+   $CHPL_HOME/examples/primers/docs/index.html
 
 
 Comment style
--- a/doc/rst/usingchapel/tasks.rst
+++ b/doc/rst/usingchapel/tasks.rst
@@ -135,7 +135,7 @@ be found in the ENVIRONMENT section of the qthread_init man page.  (Note
 that although this man page documents variables named ``QTHREAD_*``, each
 variable is actually present in both ``QT_*`` and ``QTHREAD_*`` forms,
 with the former superseding the latter.)  The qthreads man pages are
-available by means of the man(1) -m option, for example:
+available by means of the man -M option, for example:
 
 .. code-block:: sh
 
--- a/doc/rst/usingchapel/tasks.rst
+++ b/doc/rst/usingchapel/tasks.rst
@@ -139,7 +139,7 @@ available by means of the man(1) -m option, for example:
 
 .. code-block:: sh
 
-  man -M $CHPL_HOME/third-party/qthread/qthread-*/man qthread_init
+  man -M $CHPL_HOME/third-party/qthread/qthread-src/man qthread_init
 
 Note that in some cases there are Chapel environment variables that
 override Qthreads counterparts. ``CHPL_RT_NUM_THREADS_PER_LOCALE`` overrides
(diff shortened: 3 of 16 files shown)


cassella:
As suggested in #8385.  I looked for ``CHPL_HOME/foo/bar`` where ``foo/bar`` doesn't exist:

```
grep -n -I -o  -r CHPL_HOME[-_./a-zA-Z0-9]\* . | sed s,CHPL_HOME/,, | sed s,CHPL_HOME,, | sed s/:/\ /g | while read f l p; do [ -z "$p" -o -e "$p" ] || echo $f:$l names $p; done
```

The first step in reducing false positives would be for the script to ignore a trailing literal ``.``, as in ``See CHPL_HOME/README.rst.``.

cassella:
I didn't update the error messages in chip 2 or in assorted .bad files.




modules/internal/String.chpl says

```
// We'd like this to be by ref, but doing so leads to an internal
// compiler error.  See
// $CHPL_HOME/test/types/records/sungeun/recordWithRefCopyFns.future
```

But that future was retired by 3ebff5a64e7bf787403a08f9d03b741246a6622a in 2014.

A few other tests cite the same future:
```
./test/types/records/ferguson/must-init-record/myrecord.chpl:58 names test/types/records/sungeun/recordWithRefCopyFns.future
./test/types/records/ferguson/assert-locale-record/myrecord.chpl:44 names test/types/records/sungeun/recordWithRefCopyFns.future
./test/types/records/ferguson/tracking-record/myrecord.chpl:93 names test/types/records/sungeun/recordWithRefCopyFns.future
./test/types/records/ferguson/expiring-value-alias.chpl:335 names test/types/records/sungeun/recordWithRefCopyFns.future
./test/types/records/ferguson/self-pointer-record/myrecord.chpl:59 names test/types/records/sungeun/recordWithRefCopyFns.future
```

I left all those references alone, with no ability to tell whether the described change can be made.

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -433,9 +433,7 @@ associated GitHub issue number in the `#issue-number` format, e.g. `#1`.
 
 The rest of the file is optional and free-form. It can be used over the
 future's lifetime to describe in what way the test isn't working or should be
-working, implementation notes, philosophical arguments, etc.  The one-line
-summaries of all outstanding futures can be viewed by running
-``$CHPL_HOME/util/devel/test/list_futures``.
+working, implementation notes, philosophical arguments, etc.
```
cassella: ``list_futures`` was deleted by c0b57deb585509ed9265708e603ad722aa0134e6 in 2011, which introduced ``futures_stats``.  I can't tell if that script can be used for this purpose.

doc/rst/tools/chpldoc/chpldoc.rst
```diff
@@ -77,11 +77,11 @@ As an example, there is a primer on the chpldoc capability in
 ``$CHPL_HOME/examples/primers/`` that can be turned into documentation using::
 
    cd $CHPL_HOME/examples/primers
-   chpldoc chpldoc.chpl
+   chpldoc chpldoc.doc.chpl
 
 The output documentation will be located in::
 
-   $CHPL_HOME/examples/primers/docs/chpldoc.html
+   $CHPL_HOME/examples/primers/docs/index.html
```
cassella: I'm not sure this change is correct.  When I ran the command above, there was no ``chpldoc.html``.  There is an ``index.html`` and a ``modules/chpldoc.doc.html``.  I picked the first.

test/reductions/thomasvandoren/test/TestScanCountsInherit.good
```diff
@@ -1,2 +1,2 @@
-TestScanCounts.chpl:8: warning: scan has been serialized (see note in $CHPL_HOME/STATUS)
+TestScanCountsInherit.chpl:8: warning: scan has been serialized (see issue #5760)
```
cassella: This future now passes, at least for me.  :)

cassella:
``util/README`` starts to talk about ``util/quickstart``:
```
The following are designed for beginning users (see $CHPL_HOME/README
for details):
```

I can't tell what file or files it means.  None of the top-level ``README*`` mention ``quickstart``.

``doc/rst/usingchapel/QUICKSTART.rst`` maybe?

doc/rst/usingchapel/tasks.rst
```diff
@@ -139,7 +139,7 @@ available by means of the man(1) -m option, for example:
 
 .. code-block:: sh
```
daviditen: Since you're already updating this - on the line just above this example, could you please change "man(1) -m" to "man -M"?

doc/rst/tools/chpldoc/chpldoc.rst
daviditen: I think pointing to `index.html` is a good choice for this.

doc/rst/developer/bestPractices/TestSystem.rst
daviditen: I think just removing that sentence is fine.  `future_stats` isn't a direct replacement for `list_futures`

test/reductions/thomasvandoren/test/TestScanCountsInherit.good
daviditen: Cool! Could you please do `git rm test/reductions/thomasvandoren/test/TestScanCountsInherit.future` to turn it into a regular passing test?

cassella: I'll try to do that tonight.  I wasn't sure if you'd want to run the test on more platforms before removing the future.

daviditen:
> util/README starts to talk about util/quickstart:
> 
> The following are designed for beginning users (see $CHPL_HOME/README
> for details):
> I can't tell what file or files it means. None of the top-level README* mention quickstart.
> 
> doc/rst/usingchapel/QUICKSTART.rst maybe?

Sorry, I missed this comment.  Yes, I think you're right it should be `doc/rst/usingchapel/QUICKSTART.rst`.

cassella:
Ok @daviditen I think that's everything.

Issue #5482, opened 2017-03-02
Be more principled about initializing and destroying atomics used in the runtime 


ronawho:
Our atomics implementation has init and destroy functions. We're good about calling these from module code, but we're not consistent about calling the init or destroy routines from runtime code (particularly for ugni.)

For CHPL_ATOMICS=intrinsics, init and destroy are no-ops so we don't usually get bitten by this, but for CHPL_ATOMICS=locks, it's really important that we actually initialize the atomic.

ronawho:
One idea to track this down would be to stuff some extra bits into the locks implementation to track whether a lock has been initialized, and just error out in the atomic ops if it hasn't. This isn't perfect, because our canary value could happen to be set, but in practice that's unlikely

ronawho:
We should also add an init and destroy interface to network atomics.

gbtitus:
It's worth noting that, as it is currently coded, comm=ugni wouldn't work at all with CHPL_ATOMICS=locks  because it does at least one `atomic_init` operation in `chpl_comm_init()`, which is before the tasking layer is initialized and thus not a point at which one can be certain that sync variables will even work.  To some extent this highlights a problem that has always been present with the runtime atomics interface, which is that while it was introduced to support the semantics of Chapel atomics, it was quickly co-opted to also support mutual exclusion in the runtime.  But at least the "locks" implementation of runtime atomics is not actually suited to the latter, since a sync variable based approach can't necessarily be used before the tasking layer is initialized or within the tasking layer sync var implementation itself, and its performance for things like managing runtime counters would probably be very poor.  Probably the runtime should define a separate interface for its own internal atomics.

ronawho:
Ensuring that we init atomic variables will also be important as we start considering making `CHPL_ATOMICS=cstdlib` the default on more platforms: https://github.com/chapel-lang/chapel/issues/11029

gbtitus:
> I think this is outdated now that we use pthread_mutexes instead of sync variables to protect access fo the `CHPL_ATOMICS=locks` implementation (#10305)

[...]

PR #8460, opened 2018-02-13
Comm=ofi single-node execution.

gbtitus:
It turned out we only needed two things to reach single-node execution.
First, since fi_freeinfo() uses free() on the provider name member of
the fabric attributes (this is documented, though obscurely), we needed
to either allocate that from the system (malloc() e.g.) or free it and
NULL the pointer before calling fi_freeinfo().  Here, choose the former
and acquire the space via sys_malloc().

Also, fix a race which allowed us to start tearing everything down (in
chpl_comm_pre_task_exit()) before we had even finished setting it up.
For this we use a pthreads condition variable and associated mutex, in
both the setup and the teardown code.

While here, clean up the non-Cray-XC out-of-band code somewhat, such
that for now we only support (and only minimally for now) a solution
based on sockets and only under slurm.

With these changes in place a single-node hello-world execution works.
A 2-node one still fails, though, with gripes about unimplemented GET
and PUT on Cray XC and unimplemented out-of-band barrier elsewhere.

### DIFF
--- a/runtime/src/comm/ofi/comm-ofi.c
+++ b/runtime/src/comm/ofi/comm-ofi.c
@@ -24,6 +24,7 @@
 #include "chpl-comm-callbacks.h"
 #include "chpl-comm-callbacks-internal.h"
 #include "chpl-mem.h"
+#include "chpl-mem-sys.h"
 // #include "chpl-cache.h"
 #include "chpl-tasks.h"
 #include "chpl-gen-includes.h"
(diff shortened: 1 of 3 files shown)


runtime/src/comm/ofi/comm-ofi.c
```diff
@@ -77,6 +87,7 @@ int32_t chpl_comm_getMaxThreads(void) {
 }
 
 void chpl_comm_init(int *argc_p, char ***argv_p) {
+  atomic_init_bool(&progress_threads_please_exit, false);
```
ronawho: Need to call `atomic_destroy_*` on all these atomics at some point (only really matters for the locks implementation, but still) -- related to https://github.com/chapel-lang/chapel/issues/5482

gbtitus: Done, and thanks!

Issue #8464, opened 2018-02-13
problem with if expressions containing range field accessors


mppf:
### Summary of Problem

Generated code for an if-expr involving internal range fields can return a pointer to a local argument.

### Steps to Reproduce

**Source Code:**

```chapel
proc outputIt(y:int) {
  if y == 1 then
    writeln("OK");
  else
    writeln("BAD");
}

proc const range.testIfExpr(other:range(?)) {

  var lo1 = if hasLowBound() then this._low else other._low;
  outputIt(lo1);
}

proc main() {
  var a=1..10;
  var b=1..10;

  a.testIfExpr(b);
}
```

**Compile command:**
`chpl range-field-ifexpr.chpl --no-inline`

**Execution command:**
`valgrind ./range-field-ifexpr`

**Associated Future Test(s):**
test/expressions/ferguson/if-expr/range-field-ifexpr.chpl

### Configuration Information

quickstart

benharsh:
Resolved by #9649 .

PR #8465, opened 2018-02-13
Fix valgrind issue after range intent PR

mppf:
Follow-up to PR #8368.

Observed failures with --no-local and tests that use
a config bool that is passed on the command line
when run under valgrind.

See issue #8464 and the new .future test showing the if-expr issue.

This PR adds the new .future test and also adjusts range.this(range)
to work around the issue. I believe this if-expr issue would be straightforward
to resolve if if-exprs had their own AST node.

- [x] verified test/memory/figueroa/LeakedMemory6.chpl  now passes with -valgrindexe
- [x] full local testing
- [x] full --no-local testing
- [x] full valgrindexe testing


lydia-duncan:
This seems fine, but since failures occurred on --no-local and valgrind, I'd like to see test runs there listed in the checklist

PR #8473, opened 2018-02-14
Enable default initializers on records

lydia-duncan:
Allows us to start working through potential issues.  I solved a couple for the
basic implementation, but there are likely more I have missed

- Inserts the super.init() call for records, to keep us in line with the old
proposal for default initializers (since Mike doesn't think I should move them
to the new one just yet)
- Remove the check in needsConstructor() and wantsDefaultInitializer() that
prevented records from being included.
- Update the 'use default init' pragma error to not error for records and to
specify that it can only be applied to classes or records.
- Update build_record_copy_function to also return early if the type is using
a default initializer.

Add a couple of minor tests to validate records and default initializers,
and remove the future and bad file for the pragma test now that it
works.

Passed classes/initializers with futures, a full paratest, currently running a
paratest with --force-initializers.


lydia-duncan:
@noakesmichael or @benharsh, care to review?  I'm not going to let the additional --force-initializers failures block this going in, since I expect at least some, and everything else is clean

lydia-duncan:
Worked like a dream, but might as well save it!

lydia-duncan:
Still walking through the --force-initializers errors.  Nothing terrifying yet, but want to be sure.

lydia-duncan:
65% of the way through the log, so I'm not going to merge this until Monday, when I will hopefully be done.

lydia-duncan:
I have now finished categorizing the new record failures with default initializers.  While some of them are new failure categories, none of them make me worried that this change has harmed compilation without --force-initializers.

PR #8476, opened 2018-02-14
Make it easier to run valgrind testing

### LINKED ISSUE #5492: Have start_test -valgrind[exe] flag exit for bad configs
valgrind doesn't work particularly well when using qthreads or jemalloc, which are part of our default settings.  Unfortunately, I never remember this very quickly, so inevitably waste a fair amount of time when doing valgrind testing before I get everything set up correctly.

This issue proposes that we have the start_test script halt after parsing arguments and setting up if the -valgrind or -valgrindexe flags are thrown and CHPL_TASKS=qthreads or CHPL_MEM=jemalloc with an message indicating that you should use CHPL_TASKS=fifo and/or CHPL_MEM=cstdlib instead.

I was initially tempted to say that this should just print a warning, but my suspicion is that I wouldn't see them, hence I'm suggesting a halt instead.  My assumption is that developers who really want to push through this issue could comment out the halt for a short-term workaround.  If this was common behavior, we could have a way of disabling the halt via a flag, but that feels like overkill in the first draft solution to me.

mppf:
This PR makes paratest.server and start_test check for a common misconfiguration of valgrind (running it not in quickstart mode) and error out.

Additionally, it adjust start_test to set CHPL_RT_NUM_THREADS_PER_LOCALE=100 as nightly valgrind testing does if that environment variable is not already set.

Resolves #5492.

- [x] verified that studies/comd/llnl/CoMD can now pass with -valgrind
- [x] verified error on paratest in misconfiguration
- [x] verified error on start_test in misconfiguration
- [x] full local paratest

### DIFF
--- a/util/test/paratest.server
+++ b/util/test/paratest.server
@@ -662,8 +662,8 @@ sub main {
 
     # Check for a compatible valgrind configuration
     if ($valgrind > 0) {
-      $tasks = `../util/chplenv/chpl_tasks.py`; chomp $tasks;
-      $mem = `../util/chplenv/chpl_mem.py`; chomp $mem;
+      $tasks = `$ENV{CHPL_HOME}/util/chplenv/chpl_tasks.py`; chomp $tasks;
+      $mem = `$ENV{CHPL_HOME}/util/chplenv/chpl_mem.py`; chomp $mem;
 
       if ($tasks ne "fifo") {
         die "Error: valgrind requires tasks=fifo - try quickstart";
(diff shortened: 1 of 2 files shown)


mppf:
Relates to issue ##5777 but doesn't come address any other configurations besides valgrind.

util/test/paratest.server
```diff
@@ -660,6 +660,19 @@ sub main {
 
     die "Error: CHPL_HOME must be set\n" unless defined $ENV{CHPL_HOME};
 
+    # Check for a compatible valgrind configuration
+    if ($valgrind > 0) {
+      $tasks = `../util/chplenv/chpl_tasks.py`; chomp $tasks;
```
ronawho: Think the path to util/chplenv should either be prefixed with `"$pwd"` or just use `$CHPL_HOME` like the call to printchplenv does

PR #8477, opened 2018-02-14
Init parent 1

noakesmichael:
Extend business logic to support new initializer proposal

The production version of initializers generates a user fatal if a
class initializer attempts to access a parent field during phase 1.
This is valid in the new proposal.  Extend in the checks in a simple
way to allow this when an initializer includes an initDone().

Compiled with the common configurations.  Limited testing for
each configuration.  Passed a single-locale paratest.


compiler/passes/InitNormalize.cpp
```diff
@@ -905,8 +905,9 @@ void InitNormalize::fieldInitTypeInference(Expr*    insertBefore,
 ************************************** | *************************************/
 
 bool InitNormalize::isFieldAccessible(Expr* expr) const {
-  AggregateType* at     = type();
-  bool           retval = true;
+  AggregateType* at      = type();
+  bool           initNew = hasInitDone(mFn->body);
```
benharsh: Not specific to this PR, but could we store this 'initNew' property in the ``InitNormalize`` class?

PR #8480, opened 2018-02-14
Fix baseline issues after range intent PR

mppf:
Follow-up to PR #8368 and similar to PR #8465.

Adjusts several functions in ChapelRange.chpl to
avoid using if-exprs where the value portions are
both refs to range fields. This is more workaround
to issue #8464. I'm hoping that this PR applies the
workaround to all if-exprs where the parts other
than the conditional are range field accessors
(like `_low`, e.g.).

- [x] full baseline testing
- [x] full local testing


lydia-duncan:
A --no-local and valgrindexe run would also probably not go amiss here.  Other than that, this seems reasonable (if unfortunate)

PR #8486, opened 2018-02-15
Init parent 3

noakesmichael:
Continue to extend initializers to handle accessing parent fields

This PR supports the ability to reference a type alias in a parent class from a
field declaration in a child class.  This required

1. A minor change to InitNormalize::updateFieldsMember() to use ".v" rather than "." when
a field is a type-alias rather than a variable.

2. A extension to resolvePrimInit() to support the case when the RHS is a ".v" primop rather
than a SymExpr.

Compiled using the standard configurations.  Limited testing of release/ for each config.


lydia-duncan:
I was debating about doing something similar for the execution segfaults with default initializers.  Hopefully this helps there, too!

Issue #8493, opened 2018-02-15
Provide names of missing module or Enum


buddha314:
I keep getting errors like this:

```
src/Ayatollah.chpl:6: error: Cannot find module or enum
```

But It won't tell me which module or enum.  Grrr..  Is that possible?  To get around it, I have to `use` multiple statements, and abuse the `use/mention` distinction.  HOW META!

bradcray:
Just for a bit more context, what does line 6 of that file say?  Thanks.

buddha314:
So I just tracked this down and was thinking about opening a ticket.  I have these files

```
src/Modesl.chpl
    /Api.chpl
    /Setup.chpl
```

The calling file asked for

```
use Models,
      API,
      Setup;
```

For the file and the module, `API` has a case mis-match.  Within the file, the module is identified as 

```chapel
module API {
  ...
}
```

Building this on my mac, this hasn't been a problem. I had to change the `use`, filename and `module` call all to the same case.  Here `API`;

bradcray:
I think I see: you were getting the error when you moved to Linux?  Yeah, I believe we can generate a better error for that, I just needed to know how to reproduce it.  I've got it now, thanks.

bradcray:
@buddha314: Who loves ya?  #8495 should resolve this.

buddha314:
Do you really?  Could have said something yesterday before I spent $200 on dinner with my wife...

Thanks brad!

bradcray:
This error message was improved as requested by PR #8495.

PR #8495, opened 2018-02-16
Name module when cannot find it

bradcray:
This improves the error message we generate when a `use`d module or enum
name can't be found by naming the module to help the user track down the
error.  It also adds some tests to lock in the behavior and test other related
existing error messages.

### DIFF
--- a/compiler/AST/UseStmt.cpp
+++ b/compiler/AST/UseStmt.cpp
@@ -215,7 +215,11 @@ void UseStmt::scopeResolve(ResolveScope* scope) {
       }
 
     } else {
-      USR_FATAL(this, "Cannot find module or enum");
+      if (UnresolvedSymExpr* use = toUnresolvedSymExpr(src)) {
+        USR_FATAL(this, "Cannot find module or enum '%s'", use->unresolved);
+      } else {
+        USR_FATAL(this, "Cannot find module or enum");
+      }
     }
 
   } else {
--- a/test/modules/errors/useMissingModule.chpl
+++ b/test/modules/errors/useMissingModule.chpl
@@ -1,0 +1,1 @@
+use GooberPeas;
--- a/test/modules/errors/useMissingModule.good
+++ b/test/modules/errors/useMissingModule.good
@@ -1,0 +1,1 @@
+useMissingModule.chpl:1: error: Cannot find module or enum 'GooberPeas'
--- a/test/modules/errors/useMissingModule2.chpl
+++ b/test/modules/errors/useMissingModule2.chpl
@@ -1,0 +1,2 @@
+use Goober.Peas;
+
--- a/test/modules/errors/useMissingModule2.good
+++ b/test/modules/errors/useMissingModule2.good
@@ -1,0 +1,1 @@
+useMissingModule2.chpl:1: error: Cannot find object 'Goober'
--- a/test/modules/errors/useMissingModule3.chpl
+++ b/test/modules/errors/useMissingModule3.chpl
@@ -1,0 +1,4 @@
+use Goober.Peas;
+
+module Goober {
+}
--- a/test/modules/errors/useMissingModule3.compopts
+++ b/test/modules/errors/useMissingModule3.compopts
@@ -1,0 +1,1 @@
+--no-warnings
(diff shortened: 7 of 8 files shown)


bradcray:
@noakesmichael / @lydia-duncan : Would one of you give me a review on this change, motivated by issue #8493?

bradcray:
Thanks Mike.

> I believe src could be a CallExpr e.g. "use M1.M2;" One of these
> days it might be helpful to handle this case too.

Aha, I didn't anticipate that (but given how generic `src` was, you can tell I did fear
asserting that it was an UnresolvedSymExpr).  I've lost track of our non-developer pretty
printers.  Would one of them be rich enough to call on `src` to handle both cases?  Can it
print to strings in order to be incorporated into an error message like this?

> Separately one might argue that it's a shame that we allow this
> to work on mac.

I don't feel as strongly about this...  I think the real shame is that Macs are case
insensitive, but I'm not sure we should expend too much effort to work around the
nonportable decisions an operating system has made (really, I'd prefer to move
Chapel away from the assumption that a file named 'foo.chpl' is necessarily the
right place to find a module named 'foo'...  We should really be scanning all .chpl
files to see what modules they define independent of filename).

lydia-duncan:
> really, I'd prefer to move
Chapel away from the assumption that a file named 'foo.chpl' is necessarily the
right place to find a module named 'foo'... We should really be scanning all .chpl
files to see what modules they define independent of filename

That would likely break some test directories.  I think there are at least a few directories where there are five or six files that have defined the same module name, but we've gotten away with it because they all had different file names and the only way to `use` them is to explicitly name the file you want when compiling

bradcray:
> That would likely break some test directories. I think there are at least a few directories where there are five or six files that have defined the same module name, but we've gotten away with it because they all had different file names and the only way to use them is to explicitly name the file you want when compiling

The presence of challenges in the current tests doesn't strike me as a reason not to improve the compiler in general, and for me, definitely not in this respect.  But I'm not sure it would break tests:  Note that if a file is explicitly listed on the command line, it would still be used and preferred over one that is not, like today.  It's only when the compiler goes out searching for modules that weren't provided to it by the command-line files that it would use a better heuristic than "only look in files that have a name that happens to match".

bradcray:
Back on this PR's focus, though:  

@noakesmichael, I tried a pair of tests like the one you suggested and find that it doesn't seem to fall into the non-UnresolvedSymExpr case as you anticipated, but generates semi-useful error messages (in that they name reasonable things, but use terminology that sounds more like field access of an object than module access:

Case 1:

```chapel
use Goober.Peas;
```

results in:

```
useMissingModule2.chpl:1: error: Cannot find object 'Goober'
```

Case 2:

```chapel
use Goober.Peas;

module Goober {
}
```

results in:

```
useMissingModule3.chpl:1: error: Cannot find field 'Peas'
```

This behavior is consistent with what happens on master today.

Let me know if there are other cases you can think of for me to try and tickle the non-UnresolvedSymExpr case or other changes you think should happen before merging this.

noakesmichael:
I can see that there is support for src being a CallExpr but am
obviously wrong about how it gets down to that point.

Rather than guess/poke I remain more than happy to wait until
we have a user-facing problem case that’s worth prioritizing.


On Feb 16, 2018, at 10:34 AM, Brad Chamberlain <notifications@github.com<mailto:notifications@github.com>> wrote:


Back on this PR's focus, though:

@noakesmichael<https://github.com/noakesmichael>, I tried a pair of tests like the one you suggested and find that it doesn't seem to fall into the non-UnresolvedSymExpr case as you anticipated, but generates semi-useful error messages (in that they name reasonable things, but use terminology that sounds more like field access of an object than module access:

Case 1:

use Goober.Peas;

results in:

useMissingModule2.chpl:1: error: Cannot find object 'Goober'


Case 2:

use Goober.Peas;

module Goober {
}

results in:

useMissingModule3.chpl:1: error: Cannot find field 'Peas'


This behavior is consistent with what happens on master today.

Let me know if there are other cases you can think of for me to try and tickle the non-UnresolvedSymExpr case or other changes you think should happen before merging this.

Issue #8271, opened 2018-01-22
min/max procedures for a single string


nimitbhardwaj:
I was wandering around the issues, and got this #6312, as I googled the minimum string, I got lexicographical minimal string rotation algorithm, I think it can be handy sometimes and is a good way to provide functioning to min and max procedures for a single string.

Here I give the example of using it.

```chapel
writeln(max("hello"));
```
its output will be `lohel`

similarly for
```chapel
writeln(min("BCABDADAB"));
```
its output will be `ABBCABDAD`

which results the lexicographically maximum and minimum strings by rotating the string.

Moreover acts as a nice chapel algorithm design exercise.

bradcray:
As a gut reaction, this seems like something I might considering supporting, but using a name other than `max()` and `min()` (unless there was a compelling precedent for doing so).  Specifically, Chapel already supports a few min()/max() overloads—one flavor that returns the extreme value for a given type; another that returns the extreme value of its arguments.  I worry that adding a third flavor that permutes its argument to get an extreme value would be a third / different flavor and add confusion (e.g., it begs the question for me whether "max(10)" should return the maximum integer that can be created from two 1's and a 0, but I don't want max on integers to do that).

Supporting this but calling it something else seems reasonable to me (though I don't have any uses for it in mind...  so again, studying precedents set by other popular libraries might suggest doing it or not, depending on what is found).

nimitbhardwaj:
Yes, i got it, adding third kind of overloading will result in adding the function for other types, but now I think we can create a new module for it, it will be something called StringAlgorithms,chpl or some other appropriate name, which will include useful string algorithms like it, along with string search like KMP, Z buffer, string search by dfa and others, whoes names at this point of time i dont know, but by reading them and understanding them, can be implemented.
Hows this idea

nimitbhardwaj:
String algorithms are sometimes useful, the information about them can be gathered from Geeksforgeeks and other sites regarding competitive coding.

bradcray:
I'm not enough of a string expert myself to be able to anticipate what will or won't be useful, but I almost never object to people writing libraries for Chapel that they think will be useful.  This might be a good package to add mason support for, e.g.?

nimitbhardwaj:
[...]

PR #8500, opened 2018-02-16
String max-min rotation added

Deeksha96:
This pull request is in reference to the issue #8271 
Please have a look and suggest any changes required.

### DIFF
--- a/test/users/Deeksha96/StringAlgorithms.chpl
+++ b/test/users/Deeksha96/StringAlgorithms.chpl
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2004-2018 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module StringAlgorithms{
+proc minLexographicRotation(s: string): string{
+  var len = s.length;
+  var minstr = s;
+  var S = s + s;
+  for i in 1..len
+  {
+    var str: string;
+    str = S.this(i..i+len-1);
+    if(str < minstr)
+    {
+      minstr = str;
+    }
+  }
+  return minstr;
+}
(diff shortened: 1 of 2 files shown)


Deeksha96:
@mppf Please review this PR. 
We are considering applying for Rgsoc these summers and are quite new to contributing to chapel. We were facing a few problems about how to test the new module that we added to this PR locally. It would be great if you could guide us in that direction.

mppf:
hi @Deeksha96 - I can provide some feedback but I'm not so sure that it makes sense to add these to Chapel releases. I.e. I think they'd make more sense in your own GitHub repo, at least until somebody convinces me otherwise.

Even so, I'd be happy to have these added to the `tests/` directory.

Regarding testing, basically you'll want to create a .chpl file that does `use StringAlgorithms;` and calls the new functions. I like to write tests that use `assert` to check that they got the right answer and this is a good way to start. Once you have such a program, say `testStringAlgorithms.chpl`, you can add an empty .good file next to it, say `testStringAlgorithms.good`, and then you should be able to `start_test testStringAlgorithms.chpl` and see it reported as passing.

Test system docs here: https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst and https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst#creating-a-simple-test might be a good place to start.

mppf:
hi @Deeksha96 - thanks for adding a test! I think it will require discussion in order to know if StringAlgorithms can go into modules/packages (this directory is bundled with releases) so I'd propose you take the following two next steps:
 1. Adjust this PR to add the new module somewhere inside of test/. E.g. just move modules/packages/StringAlgorithms.chpl and your new test into test/users/Deeksha96/StringAlgorithms.chpl for now. This will allow this PR to merge without solving the other issue.
 2. Create a new GitHub issue - or use the existing issue #8271 - to ask whether or not it makes sense for these functions to be included in Chapel releases - or if they should just go in the test system & possibly be a Mason package.

Also in order for the PR to merge, we'll need a contributor agreement - see https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements/

mppf:
Also - regarding the should it be in Chapel installations question - on such questions it often helps to know if other languages have similar support included in the language distribution or as a separate package? It would help if you are able to answer that question for some other languages (hopefully including Python and one or two others).

Deeksha96:
Hi @mppf, I have made the required changes. I have also mailed the contributor agreement.

mppf:
We've got the contributor agreement, thanks.

verma-varsha:
@mppf I'm not getting you. Don't we already have a testStringAlgorithms.good file?

mppf:
Right, you can 
`touch StringAlgorithms.notest`
and the test command should start passing.

The start_test script thinks anything ending in .chpl is a "test". The issue is with StringAlgorithms.chpl in this case.

PR #8501, opened 2018-02-16
Avoid l-value errors for common Owned use cases

mppf:
* Allow an Owned call-temp to be passed by ref into another call (e.g. to initCopy).
* Adjust copyInit and copy-initialize functions to take in the RHS argument by ref for types containing Owned
* Add test for record containing Owned and Shared fields
* Add workarounds for several patterns that come about with constructors

Passed full local, --verify, -futures, --baseline, and --no-local testing.

### DIFF
--- a/test/classes/ferguson/delete-free/owned-shared-fields.chpl
+++ b/test/classes/ferguson/delete-free/owned-shared-fields.chpl
@@ -1,3 +1,8 @@
+// This test checks several patterns of
+// records and classes with Owned and Shared fields.
+// This test is intended to make sure that fields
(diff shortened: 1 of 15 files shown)


test/classes/ferguson/delete-free/owned-shared-fields.chpl
```diff
@@ -0,0 +1,109 @@
+use OwnedObject;
```
vasslitvinov: Would be nice to add a comment explaining what this test is testing / why does it exist.

PR #8502, opened 2018-02-16
Correct ISx reported times

ronawho:
Previously, if numTrials was greater than 1 our timing values weren't correct.
Instead of computing the average across all trials and tasks, we took the min
for each task from any trial. e.g. if we had timings like:

|         | task 1 | task 2 |
| ------- | ------ | ------ |
| trial 1 |   5s   |     1s |
| trial 2 |   1s   |     5s |

We'd report an average time of 1 second per task, which looks great but isn't
accurate. We will now correctly report the average for each task as 3 seconds,
which matches the reference version. Note that this isn't an issue for nightly
testing where we only do 1 trial, but this was giving misleading results for
scalability numbers where we tend to run 5 trials.

### DIFF
--- a/test/release/examples/benchmarks/isx/isx.chpl
+++ b/test/release/examples/benchmarks/isx/isx.chpl
@@ -429,26 +429,23 @@ proc printTimeTable(timeTable, units, timerName) {
 }
 
 //
-// print timing statistics (avg/min/max across tasks of min across
-// trials)
+// print timing statistics (avg/min/max across tasks and trials)
 //
 proc printTimingStats(timeTable, timerName) {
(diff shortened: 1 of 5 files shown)


ronawho:
@bradcray and @benharsh mind reviewing? If you're ok with this, I'll make the same change to the other ISx variations

ronawho:
Yes, that is what I think I'm doing (and I think what we want)

The [reference](https://github.com/ParRes/ISx/blob/b3a9a5de811d9944c5d171d5ad39907f00a1206b/SHMEM/isx.c#L635-L642) just computes the average across all trials/tasks (and we print the min/max for our own consumption):

PR #8504, opened 2018-02-17
Add a barrier performance test

ronawho:
This adds a barrier micro-benchmark to track the performance of barriers for
both single and multi-locale. It's somewhat lame, in the sense that it's
testing a barrier by repeatedly doing barrier.barrier() in a trial loop, so all
tasks will be perfectly aligned/synchronize, which doesn't happen in real
applications. That said, it's better than nothing and will help track upcoming
improvements to barriers used across multiple locales. This only tests the
atomic barrier, since the sync one isn't reusable.

The performance for our barrier is terrible for multi-locale since all locales
have to go back to the locale that owns the barrier (we knew/expected this.) This
poor scalability significantly hurts ISx. We have plans to add a more scalable
barrier (which is part of the motivation for adding this test.)

As a side node, our atomic barrier doesn't scale terribly well for a single
node either. We use a pretty simple linear barrier (centralized atomic counter
and wait variables.) This creates a lot of contention and cache-thrash. We
probably want some sort of dissemination barrier even within a single node.
There's a ton of papers/research out there on the best barrier for a given
situation (it's highly dependent on core/node count) so I'm sure we can lean on
those whenever we end up needing a better barrier.

The basic plan for the more scalable multi-locale barrier is to do a local
barrier on each locale, and last task to enter will do a chpl_comm_barrier().
For ISx, the poor scalability of the local barrier won't really hurt us, since
ISx only does 1 or 2 barriers in the timed section.


ronawho:
@benharsh mind reviewing?

test/performance/comm/barrier/empty-chpl-barrier.chpl
```diff
@@ -0,0 +1,39 @@
...
+  t.stop();
+
+  if printTimings {
+    writeln("Elapsed time: ", t.elapsed());
+  }
+}
+
+proc LocalBarrierBarrier() {
+  var barrier = new Barrier(numTasks, BarrierType.Atomic);
+  coforall loc in Locales do on loc do
```
benharsh: Is the single task spawned here worth worrying about for the single-node case?

test/performance/comm/barrier/empty-chpl-barrier.chpl
```diff
@@ -0,0 +1,39 @@
...
+  LocalAtomic
+};
+use BarrierMode;
+
+config param barrierMode = LocalAtomic;
+
+proc main() {
+  var t: Timer;
+
+  var LocalAtomicBarrier = new Barrier(numTasks, BarrierType.Sync);
```
benharsh: Leftover from development?

test/performance/comm/barrier/empty-chpl-barrier.chpl
ronawho: I don't think so. The task creation time will be very, very short compared to the time spent in the trial loop

test/performance/comm/barrier/empty-chpl-barrier.chpl
```diff
@@ -0,0 +1,39 @@
...
+
+  t.start();
+  select barrierMode {
+    when LocalAtomic do LocalBarrierBarrier();
```
benharsh: This can be a TODO, but I'd slightly prefer that we use ``barrierMode`` to create the barrier, and then pass it to some common function as an argument.

test/performance/comm/barrier/empty-chpl-barrier.chpl
```diff
@@ -0,0 +1,37 @@
...
+use BarrierMode;
+
+config param barrierMode = LocalAtomic;
+
+proc main() {
+  var t: Timer;
+
+  t.start();
+  select barrierMode {
+    when LocalAtomic do LocalBarrierBarrier();
```
benharsh: This can be a TODO, but I'd slightly prefer that we use ``barrierMode`` to create the barrier, and then pass it to some common function as an argument.

ronawho: That gets tricky in cases I want to add where we have 2 barriers -- One per-locale and 1 between locales, so we only have to do numLocales on-stmts instead of numLocales * numTasksPerLocales.

PR #8511, opened 2018-02-17
Dead links

cassella:
Fix more dead links.

Reviewed by @ben-albrecht.

### DIFF
--- a/compiler/parser/parser.cpp
+++ b/compiler/parser/parser.cpp
@@ -426,9 +426,7 @@ static void ensureRequiredStandardModulesAreParsed() {
             if (mod == NULL) {
               INT_FATAL("Trying to rename a standard module that's part of\n"
                         "a file defining multiple\nmodules doesn't work yet;\n"
-                        "see "
-                        "test/modules/bradc/modNamedNewStringBreaks.future "
-                        "for details");
+                        "see b2397b0052 and a94694b2aa for details");
             }
 
             mod->name      = astr("chpl_", modName);
--- a/compiler/parser/parser.cpp
+++ b/compiler/parser/parser.cpp
@@ -425,8 +425,7 @@ static void ensureRequiredStandardModulesAreParsed() {
 
             if (mod == NULL) {
               INT_FATAL("Trying to rename a standard module that's part of\n"
-                        "a file defining multiple\nmodules doesn't work yet;\n"
-                        "see b2397b0052 and a94694b2aa for details");
+                        "a file defining multiple\nmodules doesn't work yet");
             }
 
             mod->name      = astr("chpl_", modName);
--- a/compiler/resolution/resolveFunction.cpp
+++ b/compiler/resolution/resolveFunction.cpp
@@ -1384,7 +1384,7 @@ static void addLocalCopiesAndWritebacks(FnSymbol*  fn,
            //  destructor calls delete on any fields.  I think we
            //  probably need a similar change in the INOUT/IN case
            //  above.  See test/types/records/sungeun/destructor3.chpl
-           //  and test/users/recordbug3.chpl.
+           //  and test/users/ferguson/recordbug3.chpl.
            //
            // For records, this problem should go away if/when we
            //  implement 'const ref' intents and make them the default
--- a/compiler/resolution/resolveFunction.cpp
+++ b/compiler/resolution/resolveFunction.cpp
@@ -245,7 +245,7 @@ static bool needRefFormal(FnSymbol* fn, ArgSymbol* formal,
 //   test/types/atomic/sungeun/no_atomic_assign.chpl
 //   test/functions/bradc/intents/test_construct_atomic_intent.chpl
 //   test/users/vass/barrierWF.test-1.chpl
-//   test/studies/shootout/spectral-norm/spectralnorm.chpl
+//   test/studies/shootout/spectral-norm/sidelnik/spectralnorm.chpl
 //   test/release/examples/benchmarks/ssca2/SSCA2_main.chpl
 //   test/parallel/taskPar/sungeun/barrier/*.chpl
 //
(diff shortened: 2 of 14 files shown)


compiler/resolution/resolveFunction.cpp
```diff
@@ -1335,7 +1335,7 @@ static void addLocalCopiesAndWritebacks(FnSymbol*  fn,
            //  destructor calls delete on any fields.  I think we
            //  probably need a similar change in the INOUT/IN case
            //  above.  See test/types/records/sungeun/destructor3.chpl
-           //  and test/users/recordbug3.chpl.
+           //  and test/users/ferguson/recordbug3.chpl.
```
cassella: I think this comment is referring to these tests as futures.  But both have been retired.

recordbug3 by 70d5ea40, and destructor3 by 5f19b38.

modules/internal/DefaultRectangular.chpl
```diff
@@ -1685,7 +1685,7 @@ module DefaultRectangular {
     // compilation does not work right now.  The calls to chpl_comm_get
     // and chpl_comm_put should be changed once that is fixed.
     if Adata.locale.id==here.id {
-      if debugDefaultDistBulkTransfer then //See bug in test/optimizations/bulkcomm/alberto/rafatest2.chpl
+      if debugDefaultDistBulkTransfer then //See bug in test/optimizations/bulkcomm/alberto/Block/rafatest2.chpl
```
cassella: I can't tell what bug this comment means.  AFAICT, this test was never a future.

third-party/README
```diff
@@ -117,7 +117,7 @@ qthread/
 
 re2/
   Summary: The RE2 regular expression library used to support regular
-           expression operations described in doc/technotes/regexp.rst
+           expression operations described in doc/rst/technotes/regexp.rst
```
cassella: I can't find a regexp.rst or anything that looks like a regexp technote.

At least now it refers to a directory that exists...

compiler/parser/parser.cpp
```diff
@@ -426,9 +426,7 @@ static void ensureRequiredStandardModulesAreParsed() {
             if (mod == NULL) {
               INT_FATAL("Trying to rename a standard module that's part of\n"
                         "a file defining multiple\nmodules doesn't work yet;\n"
-                        "see "
-                        "test/modules/bradc/modNamedNewStringBreaks.future "
-                        "for details");
+                        "see b2397b0052 and a94694b2aa for details");
```
cassella: I'm not sure what to do here.

b2397b0052 added the cited future.
a94694b2aa removed it, not because it was fixed, but because (IIUC) the bug was specific to handling of standard modules, and the last standard module that exposed it was changed to not expose it any more.

I don't like the change I came up with at all.  But I think it's at least no less useful to a user than what's there now.

cassella: I guess maybe open a new github issue, and point the user at that?  I don't feel particularly qualified to open the issue, or even verify it's still present.

cassella:
>  - [ ]   Once the parser.cpp change is final, rebuild the parser.

Or is parser/parser.cpp not one of the files that requires rebuilding the parser?

lydia-duncan:
We're planning to review these changes the week of doc/code freeze

third-party/README
mppf: it should refer to the Regexp module docs (these used to be in a technote).

compiler/parser/parser.cpp
mppf: I think it should simply remove any reference to the test or to the commits. I think it's somewhat likely that the bug still exists and this internal error might be useful one day, but as far as I know we don't have any standard-inculded modules that meet this pattern now.

modules/internal/DefaultRectangular.chpl
ben-albrecht: (If not, then I think adjusting the path is sufficient for this PR)

compiler/resolution/resolveFunction.cpp
ben-albrecht: Something is certainly out of date here, but I'm not sure to what extent it requires updating...  Maybe @noakesmichael has a better idea?

compiler/parser/parser.cpp
ben-albrecht: Leaving this to @mppf

cassella: If someone who understands the bug could describe it in a github issue, I can point it at that.  Otherwise I'll just remove the "see ... for details" part.

mppf: I don't think it makes sense to be a github issue at present time, please just remove the details part.

modules/standard/IO.chpl
```diff
@@ -388,7 +388,7 @@ module IO {
       will also apply to class tag numbers once we start encoding types.
     - The Chapel compiler does not currently allow RAII/reference counting
       to work correctly in all cases (bug report 'records containing pointers'
-      and test files in test/users/ferguson/{byvalue.chpl,refcnt.chpl};
+      and test files in test/users/ferguson/byvalue.chpl;
```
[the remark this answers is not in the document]
mppf: That's true but I think it's important to focus this PR on removing dead links.

modules/internal/DefaultRectangular.chpl
benharsh: I this this comment should just be removed because:
1) Things have changed so much since 2012 (when this comment was added) that it is probably not accurate
2) I am unaware of a bug with that test

cassella:
Those last four commits should address all the requested changes.  I believe the only open items are the change to ``resolveFunction.cpp`` (the first review comment at the top of this page), and removing from the PR description the bit about rebuilding the parser, as on reflection I'm pretty sure changes to parser.cpp don't require that.

compiler/resolution/resolveFunction.cpp
ben-albrecht: Let's follow @mppf's suggestion of keeping this PR critical-thinking-free™, and just leave the rest of the comment as is. I'll still point this comment out to Mike to see how it should be updated separately.

ben-albrecht:
I've removed the parser rebuild TODO, because it is not necessary (confirmed with @mppf as well)

PR #8516, opened 2018-02-20
Include MasonEnv.chpl in the list of sources in the Mason Makefile

daviditen:
In the Makefile for building Mason, MasonEnv.chpl wasn't included in the list
of source files, so changes to MasonEnv didn't cause Mason to be rebuilt.  Add
it and alphabetize the Mason*.chpl files.


daviditen:
Thank you!

Issue #8497, opened 2018-02-16
ISx TODOs


ronawho:
List of next steps/todos for ISx so I don't have to rely on my (poor) memory: 

Correctness: 
 - [ ] Fix timers when using multiple trials (we take the best time from any trial instead of the average)
 - [ ] Fix pcg.bounded_random cast (we cast result to `int(32)` instead of `keyType`)

Scalability:
 - [ ] Add a scalable barrier:
   *  [ ] Permit `chpl_comm_barrier()` use in module code (currently tied up in `chpl_comm_pre_task_exit()`
   * [ ] Provide a user-facing barrier between all locales 
   * [ ] Add a barrier microbenchmark

Noise reduction (limit impact from background/runtime threads):
 - [ ] Limit interference from the progress thread (park progress thread)
 - [ ] Limit interference on non-0 locales from end-barrier (park main process, AM to wake up?)

Cleanup:
 - [ ] Minimize the difference between the release and hand-optimized variants
   *  Started this in #8290, but that [hurt performance](https://chapel-lang.org/perf/16-node-xc/?startdate=2018/01/20&enddate=2018/01/26&graphs=isxhandoptimized)

Per-locale perf:
 - [ ] Don't waste time default initializing local arrays before the app inits them? (probably future item)

PR #8519, opened 2018-02-20
Stop hard-coding the keyType in ISx makeInput

ronawho:
Cast the results of pcg.bounded_random() back to `keyType` instead of
explicitly casting to `int(32)`, which happens to be the default keyType.

Related to https://github.com/chapel-lang/chapel/issues/8497

### DIFF
--- a/test/release/examples/benchmarks/isx/isx.chpl
+++ b/test/release/examples/benchmarks/isx/isx.chpl
@@ -355,7 +355,7 @@ proc makeInput(taskID, myKeys) {
   // Fill local array
   //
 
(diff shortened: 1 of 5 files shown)


ronawho:
@benharsh and/or @bradcray this seems like a no-brainer, but maybe I'm missing something. Mind sanity checking this?

benharsh:
Good catch. Should we do something about the ``uint(32)`` as well?

ronawho:
>Good catch. Should we do something about the uint(32) as well? 

I think the `uint(32)` is ok, since that's the signature of `bounded_random()`:

https://github.com/chapel-lang/chapel/blob/c3ac79cfeeb8771f55e8af64a8914f875d6c3e2e/modules/standard/Random.chpl#L1365

Issue #8305, opened 2018-01-25
Create a set of core communication performance tests


ronawho:
Create a set of core/fundamental communication performance tests. These should allow us to compare the performance of various comm layers and track improvements/regressions to core idioms over time. 

Here's the current list of benchmarks that we’d like to add:
 - [ ] gets/puts
    - [x] small (64-bit, say) (done in #8520)
    - [x] large (1 mb?)
    - [ ] strided
 - [x] on-stmts
    - [x] fast (done directly by AM handler, no runtime task needed) (done in #8520)
    - [x] small (single transaction to initiate) (done in #8520)
    - [x] large (multi-transaction to initiate, i.e. for arg bytes)
 - [x] atomic operations (network and processor) (done in #8520)
 - [x] coforall-on idiom (done in #6378)
 - [x] start-up cost (done in #8123)
 
For all of those we should probably track varying levels of concurrency (serial, 1/4, maxed out) and think about if there's some way we can test non-blocking gets/puts. We should also track different variants including:
 - [x] many-to-one (emphasizes target-side limitations) (done in #8520)
 - [ ] one-to-many (emphasizes source-side limitations)
 - [x] balanced (emphasizes network limitations and/or interference between source-oriented and target-oriented mechanisms)

 Not as fundamental, but might be worth tracking:
 - [x] barrier performance (done in #8504)
 - [ ] concurrent array creation performance (because of dynamic array registration)

Note that some larger benchmarks stress some of these components already (and allow us to compare to references in other languages): 
 - ISx tests concurrent array creation, large all-to-all puts, barriers, and probably more
 - ra-on tests performance of on-stmts 
 - ra-atomic tests performance of atomic operations
 - ra-rmo tests performance of many small gets/puts
 - stream (particularly at scale) tests the coforall-on idiom

ronawho:
Closing, we've been pretty happy with the set of benchmarks we have currently, and expect we'd add more on-demand as we hit performance issues instead of proactively adding more.

PR #8520, opened 2018-02-20
Add low-level communication performance testing

gbtitus:
This adds perf testing for low-level communication: GETs, PUTs,
on-stmts, and "fast" on-stmts.  So far the only test is for many-to-one
communication, where N-1 nodes all communicate with the same target
node.  We envision adding one-to-one and one-to-many testing also.  We
may add many-to-many testing, although the results of that may tell us
more about the network than about Chapel communication, so it's
secondary.  So far the ml-compopts are only set up for comm=ugni.

Related to #8305 .


ronawho:
Can you put this test in test/performance/comm/ (it's a new dir added in #8504, so you may have to rebase)

test/performance/commCompare/gbt/many-to-one.chpl
```diff
@@ -0,0 +1,98 @@
...
+            numTasksPerNode, ' worker tasks per node, ',
+            numOps, ' ', op, (if numOps == 1 then '' else 's'));
+  }
+
+  if printCommDiags {
+    resetCommDiagnostics();
+    startCommDiagnostics();
+  }
+
+  const startTime = getCurrentTime();
```
ronawho: Nit, can you use a `Timer` instead of `getCurrentTime()`?

test/performance/commCompare/gbt/many-to-one.chpl
```diff
@@ -0,0 +1,98 @@
...
+
+  coforall locIdx in 1..numWorkerNodes with ( ref x0, ref x0Atomic ) {
```
ronawho: I think the default intent for atomics is already ref, can you drop the `ref x0Atomic`?

test/performance/commCompare/gbt/many-to-one.chpl
```diff
@@ -0,0 +1,98 @@
...
+    on Locales(locIdx) {
+      coforall 1..numTasksPerNode with ( ref x0, ref x0Atomic ) {
+        for i in 1..numOpsPerTask {
+          select op {
+            when opNone {
+            }
+
+            when opGet {
+              if i == 1 then x0 = 0; // prevent compiler loop-hoisting GET
```
ronawho: Do you actually see this happening? If so that's a LICM bug I need to fix

test/performance/commCompare/gbt/many-to-one.chpl
```diff
@@ -0,0 +1,98 @@
...
+              infiniteSink(x0);
+            }
+
+            when opPut {
+              x0 = infiniteSource();
```
ronawho: This violates the MCM, right? Are we ok with that (and will  weassume that the compiler will not optimize the puts away or anything?)

test/performance/commCompare/gbt/many-to-one.compopts
```diff
@@ -0,0 +1,4 @@
+-sop=opGet -snumOps=1000 many-to-one-helper.h
+-sop=opPut -snumOps=1000 many-to-one-helper.h
+-sop=opFastOn -snumOps=1000 many-to-one-helper.h
+-sop=opOn -snumOps=1000 many-to-one-helper.h
```
ronawho: We can do this later, (separate PR) but I think we'll probably want the correctness test to ensure that the correct  number of gets/puts/AMs occur. We could also add additional checks to ensure the fast-on is actually triggered for opFastOn, and not triggered for opOn.

test/performance/commCompare/gbt/many-to-one.ml-compopts
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
+
+# Aim for about 1 second of run time.  The number of operations needed
```
ronawho: Personally, I'd be in favor of run times between 5-10 seconds

test/performance/commCompare/gbt/many-to-one.ml-compopts
```diff
@@ -0,0 +1,35 @@
...
+import os
+
+comm = os.getenv('CHPL_COMM')
+comm_sub = os.getenv('CHPL_COMM_SUBSTRATE')
+ugni = comm == 'ugni'
+gn_aries = comm == 'gasnet' and comm_sub  == 'aries'
+
+if ugni or gn_aries:
+  numGets    =  20000000
+  numPuts    = 100000000
```
ronawho: Interesting -- any idea why puts seem to be much faster than gets?

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: I did see it happening, but is it really a LICM bug?  I figured LICM didn't see any stores within the loop, so it assumed that in the absence of ordering constraints, it could hoist the load out of the loop.

ronawho: Ah, I guess you're right. Since the extern proc is declared without an intent, it should default to in-intent, in which case we can assume `x0` isn't changing. I think if you declared it with a ref or an inout intent we wouldn't hoist (I actually don't know if it's legal to do mark extern procs with inout)

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: Yes, it does violate the MCM.  We don't care about the result value as such, though; we're just concerned that the communication happen.  I don't know how to force that to happen without impacting the timing, though, possibly severely.

test/performance/commCompare/gbt/many-to-one.ml-compopts
gbtitus: GETs require a PCIe round trip on the remote side to fetch the gotten value, and the network response doesn't happen until after that occurs.  That reduces the rate per initiator, and then having so many requests inbound (numWorkerNodes*numWorkerCores) loads the PCIe as well. 
PUTs don't require that PCIe round trip.  (This is rough; see [here](https://www.cray.com/sites/default/files/resources/CrayXCNetwork.pdf), top of page 22, for more detail.  Also note that we're seeing much more difference than one would expect based on reading that document.  I'm not sure why.)

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: Will do.

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: Will do.

test/performance/commCompare/gbt/many-to-one.compopts
gbtitus: Yep, sounds good and should be straightforward.

test/performance/commCompare/gbt/many-to-one.ml-compopts
gbtitus: I'll adjust to aim for about 5 secs.

test/performance/commCompare/gbt/many-to-one.ml-compopts
gbtitus: I've just now checked the emitted code for all 5 ops and they're doing what we expect them to do.  So, the much larger observed difference between GETs and PUTs, compared to what's expected, remains unexplained.

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: We could and I don't think that would invalidate the test since we're actually trying to get at the comm performance on a per-node basis rather than a per-memory-location basis.  But just a `numWorkerNodes`-sized receiver array wouldn't make this obey the MCM.  We'd actually need a `(numWorkerNodes*numTasksPerNode)`-sized array.

test/performance/commCompare/gbt/many-to-one.chpl
gbtitus: If it was `ref` we wouldn't even do the GET, because we'd pass the wide pointer rather than the value as the actual argument.  I don't think `inout` would work either, because then we'd probably do a GET and then pass the pointer and then do a PUT afterward.

Issue #8513, opened 2018-02-20
Casting array of domains to string varies depending on domain type


ben-albrecht:
### Summary of Problem

Casting an array of domains to a string results in an array of arrays of strings if the domains are 1D, and just a string if the domains are 2D.

Not 100% sure if this is a bug or not, but wanted to document it and see if I could find an explanation, i.e. maybe this is a confusing result of promotion mechanics.

### Steps to Reproduce

**Source Code:**

```chapel
var b = [{1..3}, {1..5}]: string;
writeln(b.type: string); // [domain(1,int(64),false)] [domain(1,int(64),false)] string
```

```chpl
var c = [{1..3, 1..4}, {1..3, 1..4}]: string;
writeln(c.type: string); // string
```

### Configuration Information

- Output of `chpl --version`: `chpl version 1.17.0 pre-release (7a30cf6)`

bradcray:
I agree that there are bugs here.  Mainly, I'm surprised that either of these are accepted by the parser / front-end.  I'd think that one you put curly brackets into an anonymous domain expression, you'd be limited to just one of them.  I.e., either of the following would be legal:

```chapel
var A: [1..3, 1..5] real;     // rely on sugar to introduce an anonymous domain {1..3, 1..5}
var B: [{1..3, 1..5}] real;  // call out the anonymous domain explicitly because "I like to type"
var C: [1..3] [1..5] real;   // rely on sugar to create an array of arrays over anonymous domains
var D: [{1..3}] [{1..5}] real;  // I really do like to type
```

But wouldn't think that declaring an array over multiple domains would be legal in any event.  I'm not sure what it would mean even (did you expect it to mean something specific?)

There may have been a point in the early days of the language when we thought we might want to accept:

```chapel
const D = {1..3};
var E: [D, D] real;
```

and so maybe but having lived with the language for a lot longer now, this no longer seems right to me.

The only other thing I can imagine is that we might consider these some sort of anonymous associative domain, but that doesn't seem very attractive to me either...

bradcray:
I agree that the inconsistency here is worrying.  I think you may be right that a surprise/inconsistent promotion may be at fault.  Intuitively, the behavior for `c` seems more correct to me.

ben-albrecht:
> did you expect it to mean something specific?

For context, a user encountered this when mixing up the syntax. They meant to do this:

```chpl
var b: [{1..3, 1..5}] string;
```

bradcray:
[...]

PR #8527, opened 2018-02-20
Add support for casting arrays to strings

bradcray:
It seems that, previously, casting arrays to strings was not supported
directly and so would sometimes fall into a promotion case and
sometimes fall into the catch-all 'stringify'-based implementation that
occurs at the bottom of `ChapelIO.chpl`.  This resulted in seemingly
inconsistent results from case to case depending on the array's type,
as shown in issue #8513.  This PR attempts to resolve the instability
by introducing a specific cast overload for casting arrays to strings
and adds in the tests from the issue to lock the behavior in.


bradcray:
@mppf: I put this solution together based on tracing through what was happening in the test cases and stumbling upon the catch-all cast you added at the end of `ChapelIO.chpl`.  Would you be willing / available to review as a result?

@ben-albrecht: Tagging you since you filed the issue.

PR #8528, opened 2018-02-20
Ignore default initializers for iterator records and classes

lydia-duncan:
Iterator records and classes do not create a default constructor today, so there
isn't any reason to try and use default initializers.  Once the implementation
for constructors is removed, this check should no longer be necessary.

Testing:
- [x] full paratest
- [x] full paratest with --force-initializers: no new failures and removes 54


lydia-duncan:
@benharsh or @noakesmichael, would one of you review?

PR #8529, opened 2018-02-21
Add support for -futures-only to paratest.server

bradcray:
start_test has long supported a -futures-only flag that tested the
future tests but no others.  paratest.server (and by extension,
paratest.chapcs) has _seemed_ to support such a flag, in that it
accepted it; but in practice, it ran all tests and futures because
the flag parsing for paratest.server is weak and checks for prefix
strings rather than complete strings.  So it fell into the -futures
case silently and ran everything.

Being able to run paratest in this mode has long seemed attractive
to me because I commonly run paratest.chapcs, get clean output,
and then think "Oh shoot, I should've checked futures as well"
so re-run the whole thing, wasting time and cycles.

Here, I added support for a -futures-only flag, building on
pre-existing support in subsequent scripts for this being supported
via futures mode "2".  Thanks to Elliot for doing the detective
work to determine that it would be this easy to implement.

Note that my futures-only flag is as lame as the other paratest
flags in that you could also say -futures-only-stupid-script
and it would accept the flag and run the futures only.

### DIFF
--- a/util/test/paratest.server
+++ b/util/test/paratest.server
@@ -491,7 +491,7 @@ sub find_files {
 
 
 sub print_help {
-    print "Usage: paratest.server [-compopts s] [-dirfile d] [-dirs d] [-env s] [-execopts s] [-filedist] [-futures] [-logfile l] [-memleaks f] [-nodefile n] [-nodepara m] [-valgrind[exe]] [-help|-h] [-timeout t]\n";
+    print "Usage: paratest.server [-compopts s] [-dirfile d] [-dirs d] [-env s] [-execopts s] [-filedist] [-futures] [-futures-only] [-logfile l] [-memleaks f] [-nodefile n] [-nodepara m] [-valgrind[exe]] [-help|-h] [-timeout t]\n";
     print "    -compopts s: s is a string that is passed with -compopts to start_test.\n";
     print "    -dirfile  d: d is a file listing directories to test. Default is the current diretory.\n";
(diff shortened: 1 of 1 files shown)


bradcray:
@ronawho: Seemed to work as expected.  Want to review this?

ronawho:
Nice, glad it was so easy. If you want this to be an "undocumented" option, this is fine as-is. Otherwise, probably add it to the --help output (See https://github.com/chapel-lang/chapel/pull/5304 for an example of where you need to edit the help output)

bradcray:
Good catch, I've updated the `-help` output now.

lydia-duncan:
I'm not seeing this working for me, does it maybe need to be pipe'd through paratest.chapcs as well?

bradcray:
I've only run it from within paratest.chapcs.  In what way is it not working?

lydia-duncan:
I'm still getting successes from non-future tests, even after a `make test-venv`:
`[Summary: #Successes = 8388 | #Failures = 0 | #Futures = 812]`

lydia-duncan:
Oh, maybe I'm using the wrong flag - I thought it was `-futuresonly`?

bradcray:
Right, there's a hyphen.  And as noted in the PR, unfortunately our current flag parsing is lame and only looks for prefixes, which allows all sorts of other stuff to slip through incorrectly.  :P  I'm not enough of a perl-head to have been motivated to fix that while I was adding this.

PR #8530, opened 2018-02-21
Fix support for arrays of barriers

ronawho:
Add a default 0-arg initializer for Barriers

Without this, an array of barriers fails to resolve because the default array
initialization is trying to call a default initializer.  An even simpler case
that failed is something like:

```chpl
var bar: Barrier; // resolution failure, no default init
bar = new Barrier(1);
```

### DIFF
--- a/modules/standard/Barriers.chpl
+++ b/modules/standard/Barriers.chpl
@@ -103,6 +103,11 @@ module Barriers {
       owned = true;
     }
 
+    pragma "no doc"
+    proc init() {
+      this.init(0);
+    }
+
     /* copy initializer */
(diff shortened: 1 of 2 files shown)


ronawho:
@bradcray picking on you to review since you recently updated the barrier module to use initializers. I didn't check if the test-case would fail before that PR (though I'm assuming it would). 

FWIW this is motivated by the upcoming AllLocalesBarrier PR that wants to create a block-distributed array of barriers.

ronawho:
(and it's not actually clear to me that this is the "right" solution, but it seemed liked the most obvious fix to me)

PR #8533, opened 2018-02-21
Put the main process to sleep while waiting to shutdown

ronawho:
This PR frees up `chpl_comm_barrier()` for use in module code, and prevents the
main process from interfering with tasks running on the same CPU.

Previously, the main process on non-0 locales would spin-wait on a barrier in
`chpl_comm_pre_task_exit()`, waiting for locale-0 to tell them to do an orderly
shutdown. This barrier prevented chpl_comm_barrier from being used in user
applications, and the spin-waiting hurts performance for tasks running on the
same CPU. This alleviates both issues by putting the main process to sleep with
a condition variable, and having locale-0 wake everybody up with a AM that
signals the cond var when it's time to shutdown.

This is motivated by an upcoming scalable barrier PR that wants to use
`chpl_comm_barrier()` in module code. The scalable barrier will significantly
improve ISx performance and scalability. Putting the main process to sleep
reduces interference, but we probably won't notice this in any tests yet
because the progress-thread currently causes more interference. i.e. tasks
sharing a core with the progress-thread are already the slowest tasks, so the
benefit of parking the main process will only be visible once that issue is
resolved.

### DIFF
--- a/runtime/src/comm/gasnet/comm-gasnet.c
+++ b/runtime/src/comm/gasnet/comm-gasnet.c
@@ -248,6 +248,7 @@ typedef enum {
   PRIV_BCAST_LARGE,     // put data at addr (used for private broadcast)
   FREE,                 // free data at addr
   EXIT_ANY,             // <unused> to be used for exit_any() cleanup
+  SHUTDOWN,             // tell nodes to get ready for shutdown
   BCAST_SEGINFO,        // broadcast for segment info table
   DO_REPLY_PUT,         // do a PUT here from another locale
   DO_COPY_PAYLOAD       // copy AM payload to another address
@@ -535,6 +536,17 @@ static void AM_exit_any(gasnet_token_t token, void* buf, size_t nbytes) {
   // ensure only one thread calls chpl_exit_all on this locale.
 }
 
+static chpl_bool can_shutdown = false;
+static pthread_mutex_t shutdown_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t shutdown_cond = PTHREAD_COND_INITIALIZER;
+
+static void AM_shutdown(gasnet_token_t token) {
+  pthread_mutex_lock(&shutdown_mutex);
+  can_shutdown = true;
+  pthread_cond_signal(&shutdown_cond);
+  pthread_mutex_unlock(&shutdown_mutex);
+}
+
 //
 // This global and routine are used to broadcast the seginfo_table at the outset
 // of the program's execution.  It is designed to only be used once.  This code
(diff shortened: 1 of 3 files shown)


ronawho:
@gbtitus mind reviewing?

This is similar to the branch I showed you before, but I took your early feedback into account. https://github.com/chapel-lang/chapel/compare/master...ronawho:allow-barrier-in-user-code has the not cleaned up history. 2d854b7 is the commit that cleaned up ugni, but the rest probably aren't worth reviewing individually.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -3283,6 +3298,22 @@ void chpl_comm_pre_task_exit(int all)
     return;
 
   if (all) {
+
+    if (chpl_nodeID == 0) {
+      int i;
+      for (i = 0; i < chpl_numNodes; i++) {
+	if (i != chpl_nodeID) {
```
ronawho: Just noticed that I could just start at 1 instead of 0 and skip this conditional -- thoughts?

runtime/src/comm/gasnet/comm-gasnet.c
```diff
@@ -1035,6 +1048,22 @@ void chpl_comm_barrier(const char *msg) {
 
 void chpl_comm_pre_task_exit(int all) {
   if (all) {
+
+    if (chpl_nodeID == 0) {
+     int node;
+     for (node = 0; node < chpl_numNodes; node++) {
+       if (node != chpl_nodeID) {
```
ronawho: Just noticed that I could just start at 1 instead of 0 and skip this conditional -- thoughts?

gbtitus: I saw this in the original branch and thought about mentioning it, but then I considered that when we expand on the capability you're adding here so that any node (rather than just node 0) can initiate an orderly shutdown we'll probably end up with code just like this.  That being so, I'm fine with leaving it how it is.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -890,6 +891,10 @@ typedef struct {
   void*            res;                 // if non-NULL, where to send result
 } fork_amo_info_t;
 
+typedef struct {
```
gbtitus: You could actually get rid of this and just use a `fork_base_info_t` for shutdown AMs but if you want to keep it that's fine too, and indeed there may be a need for it when orderly shutdowns can be initiated by any node.

gbtitus:
Looks good!  👍 

I had a couple of line-specific comments but you're free to merge this as is.

runtime/src/comm/ugni/comm-ugni.c
ronawho: Ah interesting and good to know. I'll probably leave it, it doesn't add much extra code and like you said we may want it in the future

ronawho:
FYI I did full gasnet segment fast and everything testing and `release/examples runtime/configMatters` for ugni. 

Also I didn't see any performance differences with ugni from this PR. This isn't surprising, though we know this will help ISx, stream, and prk-stencil (the task that shares a core with the main process was the second slowest after the task that shared with the progress thread)

PR #8534, opened 2018-02-21
Use processor atomics for Barrier variables

ronawho:
The atomics are always manipulated inside an on-stmt back to the locale where
the barrier lives, so using processor atomics will be faster.

This shaves about a second off the empty barrier test under ugni (10s to 9s),
and slightly helps ISx, but ISx is currently so noisy it's hard to see the
difference.


ronawho:
@daviditen mind reviewing?

PR #8535, opened 2018-02-21
Convert future to valgrind-only test

bradcray:
It turns out that this future had a valgrind failure which caused its
output to be less predictable than I'd first realized.  Changing it
over to be tested under valgrind only for the time being.


bradcray:
@daviditen : I put this together quickly this morning in response to the failures that have come in (so far).  Would you be able to give it a quick review?

PR #8537, opened 2018-02-21
Initializers for sync/single

benharsh:
This PR updates sync and single types to use initializers. Note that these types rely on ``chpl__initCopy`` returning a different type, so this isn't a total conversion to pure initializers.

Two .bad files are removed in order to reduce overhead of updating them as the initializer work progresses.

A breakpoint has been added for ``resolveInitVar`` to aid debugging.

Testing:
- [x] full local

### DIFF
--- a/modules/internal/ChapelSyncvar.chpl
+++ b/modules/internal/ChapelSyncvar.chpl
@@ -115,7 +115,6 @@ module ChapelSyncvar {
       ensureFEType(valType);
       this.valType = valType;
       this.wrapped = new (getSyncClassType(valType))(valType);
-      this.isOwned = true;
       super.init();
     }
 
@@ -633,7 +632,6 @@ module ChapelSyncvar {
       ensureFEType(valType);
       this.valType = valType;
       wrapped = new _singlecls(valType);
-      isOwned = true;
       super.init();
     }
 
--- a/modules/internal/ChapelSyncvar.chpl
+++ b/modules/internal/ChapelSyncvar.chpl
@@ -111,18 +111,30 @@ module ChapelSyncvar {
     var  wrapped : getSyncClassType(valType) = nil;
     var  isOwned : bool                      = true;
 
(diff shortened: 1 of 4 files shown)


benharsh:
@lydia-duncan : would you mind reviewing?

modules/internal/ChapelSyncvar.chpl
```diff
@@ -111,18 +111,30 @@ module ChapelSyncvar {
...
-    }
-
-    proc _syncvar(type valType, const other : _syncvar(valType)) {
-      ensureFEType(valType);
-
-      wrapped = other.wrapped;
-      isOwned = false;
+      this.valType = valType;
+      this.wrapped = new (getSyncClassType(valType))(valType);
+      this.isOwned = true;
```
lydia-duncan: You could omit the initialization of isOwned, I think

modules/internal/ChapelSyncvar.chpl
```diff
@@ -613,18 +629,30 @@ module ChapelSyncvar {
     var  wrapped : _singlecls(valType) = nil;
     var  isOwned : bool                = true;
 
-    proc _singlevar(type valType) {
+    proc init(type valType) {
       ensureFEType(valType);
-
+      this.valType = valType;
       wrapped = new _singlecls(valType);
       isOwned = true;
```
lydia-duncan: Could also omit this one

modules/internal/ChapelSyncvar.chpl
benharsh: Omit because of default initialization?

lydia-duncan: Yeah.  This is the same as the declared default value for the field, so omitted initialization would just insert the same line

benharsh:
Made the change and ran a 100 or so tests to make sure nothing blew up. It all looks good, thanks for the review!

PR #8538, opened 2018-02-21
Fix codegen failure for assignment between voids

benharsh:
The compiler was leaving a stray statement-level SymExpr to ``_void`` in the tree, which caused codegen failures for ``--llvm`` testing.

This PR removes any ``_void`` SymExprs that are at the statement level. Also adds a test to more easily catch this case in the future.

Testing:
- [x] full local
- [x] full llvm


benharsh:
@daviditen : would you mind reviewing?

benharsh:
Just ran across a couple of errors which lead me to believe I'll need to handle PRIM_YIELD as well. I'll update the PR soon.

benharsh:
Ran into issues with loops as well, so I went with a simpler and more focused condition. Would you mind taking a quick look again?

PR #8540, opened 2018-02-21
make copyFnBodyForInlining() externally accessible

vasslitvinov:
I need to use copyBody() for my ForallStmt work,
so moving it from 'static' to passes.h.

While there, make minor formatting tweaks.

Testing (just in case): linux64 --verify.


vasslitvinov:
@benharsh would you review?

benharsh:
I think this function should be renamed if we're going to expose it like this. Otherwise this seems OK to me.

Issue #8452, opened 2018-02-13
documented use of try!, halt() in modules


psahabu:
To make better use of error handling, we are trying to convert as many uses of `try!` and `halt()` over to `throws` in the standard library. However this is not always possible due to limitations in error handling or truly fatal conditions. This issue documents instances of `try!` and `halt()` with the hope that their blocking reasons will be resolved in the future.

### pending work on string casts
- internal/ChapelBase.chpl

### can't throw from initializers
- packages/Buffers.chpl

### can't throw from non-inlined iterators
- packages/RecordParser.chpl

### `writeln` throws
- internal/tasktable/on/ChapelTaskTable.chpl
- standard/Time.chpl

### IO
- standard/IO.chpl

psahabu:
@mppf, @bradcray: Here's the working list of `try!` in modules, I'll add the instances of `halt()` when I start removing them from modules.

mppf:
@psahabu - I've updated the issue description to go more in-depth on the IO cases, since just saying "IO" isn't very useful.

mppf:
@psahabu - while we've triaged most uses of try!, not all of them are good uses:

> can't throw from non-inlined iterators
> should be throws / some recoverable error 

Also there were some I couldn't quickly diagnose, listed under ?

> ? 

... these need additional diagnosis.

Presumably, if the goal here was to drive improvements to error handling, the next step would be to actually fix any of these that can / should be fixed?

psahabu:
@mppf - I agree that there are non-ideal uses in this list, I'll look into getting those resolved. I'll also look into triaging the unknown category.

PR #8547, opened 2018-02-22
add error handling to casts from string

psahabu:
### motivation
Failed casts of strings to primitive data types (`bool`, `int`, `real`, `imag`, `complex`) and `enum` types used to halt.

``` chapel
var s = "brad";
var i = s: int;  // this will halt!
```

To help users write resilient programs and put more weight on error handling, this PR allows such casts to throw errors.

``` chapel
var s = "brad";
try! {
  var i = s: int;
} catch e: IllegalArgumentError {
  writeln("caught cast error");
}
```

### approach
- Change the runtime to return an error flag if a cast is not possible.
- Have the matching `extern` declarations and calls pass an error flag variable by ref.
- Check and translate the error flag into an `IllegalArgumentError` in Chapel `_cast(...)` functions.
- Do the same for the generated `enum` casts in `buildDefaultFunctions`.

### misc
`integral_to_c_string()` is a special case where a failure would indicate a truly fatal error in the runtime. As such, its matching Chapel module function `_cast(type t, x: integral) where t == string` has been set to halt on such an error. This has been noted in #8452.

- [x] update tests with improperly handled casts

### DIFF
--- a/modules/internal/CString.chpl
+++ b/modules/internal/CString.chpl
@@ -117,89 +117,29 @@ module CString {
   // casts from c_string to bool types
   //
   inline proc _cast(type t, x:c_string) throws where isBoolType(t)
-  {
-    pragma "insert line file info"
-    extern proc c_string_to_chpl_bool(x:c_string, ref err: bool) : bool;
-
-    var isErr: bool;
-    var retVal = c_string_to_chpl_bool((x:string).strip().c_str(), isErr): t;
-
-    if isErr then
-      throw new IllegalArgumentError(x:string, "Unexpected value when converting from string to bool");
-
-    return retVal;
-  }
+    return try ((x:string).strip()): t;
 
   //
   // casts from c_string to integer types
   //
-  pragma "insert line file info"
-  extern proc c_string_to_int8_t  (x:c_string) : int(8);
-  pragma "insert line file info"
-  extern proc c_string_to_int16_t (x:c_string) : int(16);
-  pragma "insert line file info"
-  extern proc c_string_to_int32_t (x:c_string) : int(32);
-  pragma "insert line file info"
-  extern proc c_string_to_int64_t (x:c_string) : int(64);
-  pragma "insert line file info"
-  extern proc c_string_to_uint8_t (x:c_string) : uint(8);
-  pragma "insert line file info"
-  extern proc c_string_to_uint16_t(x:c_string) : uint(16);
-  pragma "insert line file info"
-  extern proc c_string_to_uint32_t(x:c_string) : uint(32);
-  pragma "insert line file info"
-  extern proc c_string_to_uint64_t(x:c_string) : uint(64);
-
-  inline proc _cast(type t, x:c_string) where t == int(8)
-    return c_string_to_int8_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == int(16)
-    return c_string_to_int16_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == int(32)
-    return c_string_to_int32_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == int(64)
-    return c_string_to_int64_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == uint(8)
-    return c_string_to_uint8_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == uint(16)
-    return c_string_to_uint16_t((x:string).strip().c_str());
-  inline proc _cast(type t, x:c_string) where t == uint(32)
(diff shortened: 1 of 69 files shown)


modules/internal/CString.chpl
```diff
@@ -116,11 +116,18 @@ module CString {
...
-    return c_string_to_chpl_bool((x:string).strip().c_str()) : t;
+    extern proc c_string_to_chpl_bool(x:c_string, ref err: bool) : bool;
+
+    var isErr: bool;
+    var retVal = c_string_to_chpl_bool((x:string).strip().c_str(), isErr): t;
+
+    if isErr then
+      throw new IllegalArgumentError(x:string, "Unexpected value when converting from string to bool");
+
+    return retVal;
```
mppf: Why isn't this method written the same as the string version?
Why aren't the methods below updated (e.g. `inline proc _cast(type t, x:c_string) where t == int(8)
return c_string_to_int8_t((x:string).strip().c_str());`)?
Since these c_string casts call strip anyway, can they just convert the c_string to a string and then use the string : int (or whatever) cast?

modules/internal/StringCasts.chpl
```diff
@@ -75,43 +83,60 @@ module StringCasts {
...
 
-  inline proc _cast(type t, x: string) where isIntegralType(t) {
+  inline proc _cast(type t, x: string) throws where isIntegralType(t) {
     //TODO: switch to using qio's readf somehow
     pragma "insert line file info"
-    extern proc c_string_to_int8_t  (x:c_string) : int(8);
+    extern proc c_string_to_int8_t  (x:c_string, ref err: bool) : int(8);
     pragma "insert line file info"
-    extern proc c_string_to_int16_t (x:c_string) : int(16);
+    extern proc c_string_to_int16_t (x:c_string, ref err: bool) : int(16);
```
mppf: CString also has these: ```extern proc c_string_to_int16_t (x:c_string) : int(16);```
which is presumably now out of date.
Can we remove the C string ones? Either this PR won't pass testing, or they are always unused. Can you check?

psahabu:
@mppf: I've followed your suggestions for `CString` after getting snagged on a few tests. I'm running testing on it now, I believe it should come back clean but we'll see.

modules/standard/IO.chpl
```diff
@@ -1579,11 +1579,19 @@ proc open(out error:syserr, path:string="", mode:iomode, hints:iohints=IOHINT_NO
...
-    if port_start > 0 then port = path[port_start..port_end];
+    var port_str = "";
+    if port_start > 0 then port_str = path[port_start..port_end];
+
+    var port: int;
+    try! {
+      port = port_str: int;
+    } catch e: IllegalArgumentError {
+      error = ENOENT;
+    }
```
psahabu: @mppf: I'm not confident that this is the right error code to raise if the port_str fails to cast into an int. Any ideas?

modules/standard/IO.chpl
```diff
@@ -5551,7 +5602,11 @@ proc _setIfPrimitive(ref lhs:?t, rhs:?t2, argi:int):syserr where t==bool&&_isIoP
 private inline
 proc _setIfPrimitive(ref lhs:?t, rhs:?t2, argi:int):syserr where t!=bool&&_isIoPrimitiveType(t)
 {
-  lhs = rhs:t;
+  try {
+    lhs = rhs:t;
+  } catch {
+    return EFORMAT;
```
psahabu: @mppf: I'm also not confident with this error code, any suggestions?

modules/standard/IO.chpl
mppf: I think EINVAL just like the others is reasonable.

modules/standard/IO.chpl
mppf: I think EFORMAT is reasonable here, but so is ERANGE (as the comment above points out)

psahabu:
@mppf: I've got a clean testing run, do you mind giving this a quick re-review? I can find someone else if you're swamped.

psahabu:
While #8749 accused this PR of being the cause of performance regressions, further investigation has shown it to be indirect at best. What may be happening is that string cast error handling has significantly impacted the node ids used in the compiler, which can in turn impact the argument order of generated `coforall` functions. Different versions of `gcc` respond differently to these reorganized functions, with `6.3` being negatively impacted while `4.2` and `7.3` see no impact.

mppf:
> What may be happening is that string cast error handling has significantly impacted the node ids used in the compiler, which can in turn impact the argument order of generated coforall functions

node ids? are you saying that performance changed because e.g. someSymbol->id changed? In other words that performance depends on the values of AST ids? That seems unfortunate to be sure.

psahabu:
> In other words that performance depends on the values of AST ids? That seems unfortunate to be sure.

It's the plausible theory we have at the moment, I'll need to spend time fiddling with generated code to make sure that's the case.

PR #8548, opened 2018-02-22
FLAG_ALWAYS_PROPAGATE_LINE_FILE_INFO

psahabu:
While working on #8547, I was running into a problem where errors were being reported as uncaught deep in module code, rather than user code. @mppf and I discovered two reasons for this occurrence:

- `chpl_propagate_error()` was not decorated with `pragma "insert line file info"`
- `CHPL_DEVELOPER` was set, triggering actively unhelpful code in `insertLineNumbers.cpp`

`halt()` was being exempted from this actively unhelpful code so that `CHPL_DEVELOPER` users would always receive a clear error message. To clean up that part of the codebase and generalize the `halt()` protection, we added `FLAG_ALWAYS_PROPAGATE_LINE_FILE_INFO` to the compiler and decorated it on these module functions:

- `chpl_propagate_error()`
- `halt()`
- `warning()`
- `chpl_here_alloc()`
- `chpl_here_calloc()`
- `chpl_here_relloc()`
- `chpl_here_good_alloc_size()`
- `chpl_here_free()`

- [X] passes linux64+flat testing


compiler/passes/insertLineNumbers.cpp
```diff
@@ -153,15 +152,11 @@ insertLineNumber(CallExpr* call) {
   } else if (fn->hasFlag(FLAG_EXTERN)                           ||
              (fn->hasFlag(FLAG_EXPORT) &&
               !fn->hasFlag(FLAG_INSERT_LINE_FILE_INFO))         ||
-             strcmp(fn->name, "chpl__heapAllocateGlobals") == 0 ||
-             strcmp(fn->name, "chpl__initStringLiterals")  == 0 ||
-             strcmp(fn->name, "chpl__initModuleGuards")    == 0 ||
-             strcmp(fn->name, "chpl_gen_main")             == 0 ||
```
mppf: We changed these removed functions into an assert, which has since been removed. I'd recommend leaving the assert in for this PR, with a TODO note that it should be removed. In any case it needs to pass full local testing with the assert on. Also AFAIK heapAllocateGlobals might only be relevant in GASNet configurations. Hence the assert to catch errors in our logic with nightly testing.

compiler/passes/insertLineNumbers.cpp
```diff
@@ -153,15 +152,11 @@ insertLineNumber(CallExpr* call) {
...
              ftableMap.count(fn)                                ||
              (mod->modTag == MOD_USER               &&
               !fn->hasFlag(FLAG_COMPILER_GENERATED) &&
               !fn->hasFlag(FLAG_INLINE)) ||
-             (developer && strcmp(fn->name, "halt"))) {
+             (developer && !fn->hasFlag(FLAG_ALWAYS_PROPAGATE_LINE_FILE_INFO))) {
```
mppf: I think it would be nice if the PR message were clearer that we're adding a flag to allow certain functions to get the same treatment `halt` was already getting.

test/memory/psahabu/lineno-memLeak.execopts
```diff
@@ -0,0 +1 @@
+--memLeaks
```
mppf: You have to add --devel for this to actually test the improvement

test/memory/psahabu/lineno-memLeak.good
```diff
@@ -0,0 +1,7 @@
+
+=================================================================================================================
+Allocated Memory (Bytes)         Number   Size     Total    Description                      Address             
+=================================================================================================================
+lineno-memLeak.chpl:2            1        16       16       MyClass                          0x00007fa2edc03f70
```
mppf: You'll have to put a prediff of some sort, since the address will change run-to-run. Also, we almost certainly only care about the .chpl:2 part. Probably grepping for it in the prediff is the way to go.

PR #8550, opened 2018-02-22
Add a scalable allLocalesBarrier

ronawho:
This adds a barrier that must be called on all locales, optionally with a
specific number of tasks per locale. This uses the current processor atomic
barrier to barrier between tasks on a locale, and the last task to enter the
local barrier will do a `chpl_comm_barrier()` and wait for all locales to sync
before releasing the tasks.

This scales extremely well, especially compared to the current barrier. As a
quick comparison, the table below shows elapsed time to do 1,000 barriers on 36
core nodes. At 256 nodes the old barrier ends up doing ~10,000 on-stmts back to
locale 0, which exhausts memory unless the call stack size is lowered. The new
barrier does not use any on-stmts (it's a tree based put barrier under ugni.)

| Barrier Type (1K trials  ) | 1 node | 2 node | 16 node | 64 node | 256 node |
| -------------------------- | -------| ------ | ------- | ------- | -------- |
| ugni localAtomicBarrier    |  0.07s |  0.4s  |  3.4s   | 14.1s   | 64.0s    |
| ugni allLocalesBarrier     |  0.08s |  0.2s  |  0.2s   |  0.2s   |  0.2s    |

Most of the time recorded for the allLocalesBarrier is actually the time to
spawn the remote tasks. Since it's scalable we can bump 100,000 trials:

| Barrier Type (100K trials) | 1 node | 2 node | 16 node | 64 node | 256 node |
| -------------------------- | -------| ------ | ------- | ------- | -------- |
| ugni allLocalesBarrier     |  3.2s  |  4.5s  |  7.5s   | 20.2s   | 64.2s    |
| gn-aries allLocalesBarrier |  3.2s  |  4.2s  |  7.0s   | 18.4s   | 62.2s    |

Here we see that the tree barrier that ugni provides, and the dissemination
barrier gasnet-aries provides both result in some sort of logarithmic curve.

Note that the implementation is pretty ugly, because it hacks into the existing
barrier code (to avoid code duplication) and calls `chpl_comm_barrier()`.

Using this barrier is similar to using something like `shmem_barrier_all()` or
`MPI_Barrier(MPI_COMM_WORLD)` in the sense that it's a singleton global barrier
between all locales in a job.

Eventually we want to be able to barrier between subsets of locales, but that
implementation will require more work, whereas this allows us to tie into the
already optimized `chpl_comm_barrier()` code.

This will allow us to significantly improve the scalability and performance of
ISx, which is the primary motivation for this PR.

Related to https://github.com/chapel-lang/chapel/issues/8497


ronawho:
Rendered docs available at: http://web/~eronagha/add-all-locales-barrier-2/modules/packages/AllLocalesBarriers.html

ronawho:
This "ugly" part of this PR is that I'm hacking into the standard barrier module in order to avoid cloning the atomic barrier code. We could avoid this by creating a new barrier subclass, but it involve a lot of code duplication

ronawho:
@benharsh and/or @daviditen mind reviewing?

PR #8551, opened 2018-02-22
Fix failure with default initializers with record fields that utilize initializers

lydia-duncan:
Prior to this change, we would get the following error message in the
default initializer of a type that had a record field when the record
utilized initializers (either explicitly or via default initializers) `non-lvalue
actual is passed to ‘ref’ formal ‘this’ of init()`.  This was because the
default value for the argument created was marked as a specific type
of temporary and our lvalue checks prevent those temporaries from
being used as `ref` arguments to functions, but the initializer for
that temporary required the instance to be `ref`.  It turned out that
there was a flag that compiler-generated constructors were using
which prevented them from encountering this error, so utilize that
flag for compiler-generated initializers as well (and propagate it to the
_new function for classes).

Adds a test of a class using a compiler generated initializer with a
record field that uses explicit initializers, since that was not a case
that was handled before.

With this change, 22 failures with --force-initiailizers now pass, and a
couple have changed their failure mode.

Passed classes/initializers with futures, a full paratest, and no new
failures on a paratest with --force-initializers


lydia-duncan:
Summary of --force-initializers run:
- No new failures as compared to master
- 22 failures now pass
- 3 failures now have a different failure mode
  - 1 returns to an old failure mode from prior to enabling over records
  - 1 shifts to a known records and noinit failure mode
  - 1 shifts to a variation on a known failure mode, but not one that was previously seen so far as I can tell
- 3 failures continue to fail in the same manner (due to testing the incorrect use of the record type for a param)

lydia-duncan:
@noakesmichael or @benharsh, care to review?

PR #8558, opened 2018-02-23
Fix a bug in ugni when all puts in a vector put are proxied

ronawho:
If all puts in a vector/chained put are proxied (because the put locations
aren't registered) don't try to do a post_fma_ct_and_wait(). The vector put is
currently only used for the dynamic memory registration broadcast, which can't
proxy, but an upcoming use of put_V for chpl_comm_barrier() will proxy if a
hugepage module isn't loaded. In this case all puts were proxied (and done with
the non vector put) so there was nothing in the chained transaction arrays to
post, which resulted in an internal error.


ronawho:
@gbtitus not sure if this is "best" fix, but it was pretty simple. Without this https://github.com/chapel-lang/chapel/pull/8557 without a hugepage module loaded leads to internal errors along the lines of:

```
internal error: 0:comm-ugni.c:7188: internal error: CTPostFMA() failed
srun: error: nid00076: task 0: Exited with exit code 2
```

gbtitus:
That fix looks fine to me -- okay to proceed.  👍

PR #8572, opened 2018-02-23
start_test to process files in a deterministic order

vasslitvinov:
To simplify comparison of results of different runs of `start_test`,
force it to process the files in a deterministic order.

For that, have `sub_test` and `sub_clean` sort their file lists.

The sorting order may differ between locales, OSes, etc.
Ex. capital vs. lower-case characters may be ordered differently.
I let it be, in favor of simplicity of the change in this PR.

UPDATE: also add sort() in `start_test` proper,
for deterministic order of recursion into sub-directories.

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -207,6 +207,8 @@ def test_directory(test, test_type):
         # stop recursing if flag is set
         if not args.recurse:
             del dirs[:]
+        else:
+            dirs.sort()
 
         # ignore skipifs for --clean-only run
         if not args.clean_only:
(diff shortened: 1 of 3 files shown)


dmk42:
I like this.  I'd give others a chance to object, but if you get no other comments in a reasonable time, OK to merge.

ben-albrecht:
Do the filenames in the sort call include the file paths, i.e. would this keep files in the same directory grouped together?

vasslitvinov:
> Do the filenames in the sort call include the file paths, i.e. would this keep files in the same directory grouped together?

The path is not included in the filenames in sub_test, where the filenames are produced by `os.listdir()` . The path IS included in the filenames in sub_clean - produced by `glob.glob()`.

Regardless of that, in both these cases (where I added `sort`), all files come from a single directory. So grouping is not an issue afaik.

Aside: the path is typically "." . The only exception that I see happens in sub_clean when running start_test on a specific directory. In that case the path is that directory.

vasslitvinov:
UPDATE: I also just added sort() in `start_test` proper,
for deterministic order of recursion into sub-directories.

Issue #6493, opened 2017-06-17
Convert random number generators to record-wrapped classes


bradcray:
It seems unfortunate that our random number generators are raw classes and require explicit user deletes.  Is it time to make them record-wrapped classes, as a case of using Owned/Shared?

bradcray:
@mppf: As the most recent author of RNG stuff and the author of Owned/Shared, what do you think about this?

mppf:
Mainly, I suspect we might run into compiler bugs if/when we try to make appropriate type aliases.
But I don't see any problem with attempting it. I suspect we'd need to make e.g. RandomStream be a record that contains an Owned(RandomStreamImpl) or a Shared(RandomStreamImpl). The odd thing here is that I'm not sure I can justify choosing Owned or Shared. Perhaps we need RandomStream and SharedRandomStream? Or maybe it is really better to have users wrap with Owned/Shared? (e.g. new Owned(new RandomStream(...)) ).

cassella:
Does the delete-free support obviate the need to do anything here?

mppf:
I think it's reasonable for users to write `owned RandomStream` etc. However others might disagree.

bradcray:
I agree and will close this, thanks!

PR #8575, opened 2018-02-23
makeRandomStream returns an Owned random stream 

mppf:
Since makeRandomStream allocates a new object,
it can communicate to the caller that it should be
freed by the caller by returning an owned object. 

This PR does not attempt to deprecate makeRandomStream
for two reasons:
 1. Can't add a warning overload for return type changes like this
 2. Owned coerces to borrowed so code using the older
      makeRandomStream has a chance of continuing to work.

Relates to issue #6493 and #9087.

- [x] full local testing

### DIFF
--- a/modules/standard/Random.chpl
+++ b/modules/standard/Random.chpl
@@ -136,7 +136,6 @@ module Random {
     where isSupportedNumericType(arr.eltType) {
     var randNums = makeRandomStream(seed, eltType=arr.eltType, parSafe=false, algorithm=algorithm);
     randNums.fillRandom(arr);
-    delete randNums;
   }
 
   pragma "no doc"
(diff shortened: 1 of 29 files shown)


mppf:
[Error matching compiler output for distributions/robust/arithmetic/modules/test_module_Random]
[Error matching program output for library/standard/Random/bradc/writeRandomStream (compopts: 1)]
[Error matching program output for library/standard/Random/bradc/writeRandomStream (compopts: 2)]
[Error matching compiler output for library/standard/Random/ferguson/check-time-seed]

mppf:
The error with distributions/robust/arithmetic/modules/test_module_Random looks like an issue with forwarding that goes away if we update to use coercions.

bradcray:
This is very exciting (a poster child for "owned" classes) and looks good to me!

PR #8576, opened 2018-02-23
Let push_back/push_front use in intent

### LINKED ISSUE #6841: push_back() on an Owned(C) type
### Summary of Problem

The `array.push_back(val)` method takes `val` as a *blank intent*, which defaults to *const*. The `Owned` record needs to set the RHS to `nil` when transferring ownership from `val` to the array element, thus we get the error shown below.

I suspect this impacts other array methods, but haven't done much testing.

One potential solution is to make `array.push_back(val)` and other related methods *ref intent*, but this may have other unwanted effects that I am currently not aware of.

**Work-around**

Note that one can work around this today by writing a custom `push_back` routine with a ref intent (not great, but it works).

### Steps to Reproduce

**Source Code:**

```chapel
use OwnedObject;

class MyClass {
  var x: int;
}

var D = {1..3};
var A: [D] Owned(MyClass);

var c = new Owned(new MyClass(1));
A.push_back(c);

writeln(A[4].x);

```

**Compiler error message:**

```
$CHPL_HOME/modules/internal/ChapelArray.chpl:2631: In function 'push_back':
$CHPL_HOME/modules/internal/ChapelArray.chpl:2661: error: illegal lvalue in assignment
```

*TODO*

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.16.0 pre-release (96295b9)`

mppf:
To help with arrays of Owned. A reasonable alternative would be for push_back to have a different overload for Owned.

Closes  #6841.

- [x] add test from  #6841
- [x] full local testing


modules/internal/ChapelArray.chpl
```diff
@@ -2744,6 +2744,9 @@ module ChapelArray {
 
       reallocateArray(newRange, debugMsg="push_back reallocate");
 
+      // This could "move" from val to the array element, but
```
daviditen: I don't understand this comment. I think that `this[this.domain.high]` should effectively be uninitialized memory.  It was allocated by `reallocateArray()` above (likely as part of a previous allocation for several not-yet-used elements) and is not used until this assignment.

daviditen: Thanks for explaining.  I'm OK with just doing 1. for now.

PR #8577, opened 2018-02-23
Ignore erroneous auto-copy calls in auto-copy fns

### LINKED ISSUE #8603: Compilation error when `Owned` is tuple member
### Summary of Problem
```
$CHPL_HOME/modules/internal/ChapelBase.chpl:1167: error: implicit copy-initialization invoked for a type that does not allow it
```

### Steps to Reproduce

**Source Code:**

```chapel
use OwnedObject;
class Foo {
  var x: int;
};

type Bar = (Owned(foo), int); // <-- Error here?
```

**Compile command:**
`chpl --force-initializers <source>`

### Configuration Information
`chpl version 1.17.0 pre-release (2f0af5c5db)`

mppf:
This change helps with the effort to migrate 'new' to return Owned.
In particular, on that branch, I was seeing tuple autocopy functions for a tuple containing Owned calling the Owned autocopy function. But the tuple-of-owned autocopy was never called, so that's not a problem. This pattern can come up with tuples, for which the compiler directly generates init/auto copy.

This PR adjusts the handling of FLAG_ERRONEOUS_INITCOPY and FLAG_ERRONEOUS_AUTOCOPY to mark init/auto copy calls that call erroneous init/auto copies. Then, it ignores errors within init/auto copy functions themselves, since if they are never called, no error needs to be generated.

This PR can be viewed as a follow-on to PR #5683.
Resolves #8603.

- [x] full local testing


compiler/resolution/callDestructors.cpp
```diff
@@ -1065,22 +1065,57 @@ void insertReferenceTemps(CallExpr* call) {
 // flag is ever called and raises an error if so.
 static void checkForErroneousInitCopies() {
 
+  // Propagate uses into autoCopy/initCopy fns
```
daviditen: This should clarify what 'uses' are being propagated.  e.g. "Propagate uses of FLAG_ERRONEOUS_INITCOPY/FLAG_ERRONEUS_AUTOCOPY into init/autocopy functions that call erroneous init/autocopy functions."

compiler/resolution/callDestructors.cpp
```diff
@@ -1065,22 +1065,57 @@ void insertReferenceTemps(CallExpr* call) {
...
+
   forv_Vec(FnSymbol, fn, gFnSymbols) {
     if (fn->hasFlag(FLAG_ERRONEOUS_INITCOPY)) {
       // Error on each call site
       for_SymbolSymExprs(se, fn) {
-        USR_FATAL_CONT(se,
-                       "copy-initialization invoked for a type "
-                       "that does not have a copy initializer");
+        if (FnSymbol* callInFn = se->getFunction())
+          if (!callInFn->hasFlag(FLAG_INIT_COPY_FN))
```
daviditen: Would an assert that `callInFn->hasFlag(FLAG_ERRONEOUS_INITCOPY)` in the else side of this branch be good?

compiler/resolution/callDestructors.cpp
```diff
@@ -1065,22 +1065,57 @@ void insertReferenceTemps(CallExpr* call) {
...
     }
 
     if (fn->hasFlag(FLAG_ERRONEOUS_AUTOCOPY)) {
       // Error on each call site
       for_SymbolSymExprs(se, fn) {
-        USR_FATAL_CONT(se,
-                       "implicit copy-initialization invoked for a type "
-                       "that does not allow it");
+        if (FnSymbol* callInFn = se->getFunction())
+          if (!callInFn->hasFlag(FLAG_AUTO_COPY_FN))
```
daviditen: And a similar assert here?

PR #8579, opened 2018-02-23
Fix think-o in testArgMapping

mppf:
testArgMapping includes a mechanism to print out the reason one function in chosen over another in overload disambiguation. This included misleading wording: "generic any vs generic" is actually not what the spec says; the spec says "generic any vs not generic any". This PR adjusts the wording to resolve the thinko.

- [x] full local testing


mppf:
[Error matching compiler output for domains/marybeth/test_compare_range]
[Error matching compiler output for studies/beer/bradc/beer-promoted-infer]

PR #8580, opened 2018-02-23
Don't print out Owned representation

mppf:
Don't print out 'p' when printing an Owned, just print class pointer.
This helps as tests are updated to use Owned in more places, since the
output does not need to change as much.

- [ ] full local testing


mppf:
[Error matching program output for classes/delete-free/owned-init2]

PR #8338, opened 2018-01-31
Add lifetime checking pass

mppf:
PR message forthcoming

- [x] full local testing ccda7fd1adfcab4d18950cc37117e98485aa2a18 c909238a69b79e627392b6686ad96422a6319556

PR #8581, opened 2018-02-23
Add comparison operators for astloc

mppf:
These comparison operators for AST locations are useful for simplifying error messages and I used them in my upcoming work on lifetime checking (PR #8338).

- [x] full local testing

### DIFF
--- a/compiler/include/baseAST.h
+++ b/compiler/include/baseAST.h
@@ -148,6 +148,15 @@ public:
   int         lineno;    // line number of location
 };
 
+static inline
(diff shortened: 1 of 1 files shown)


compiler/include/baseAST.h
```diff
@@ -148,6 +148,15 @@ class astlocT {
   int         lineno;    // line number of location
 };
 
+static inline
```
daviditen: I think it would be better to make these member functions in the astlocT class.

PR #8582, opened 2018-02-23
Improve gatherLoopDetails to work after ReturnByRef

mppf:
Upcoming work in PR #8338 adds a lifetime checking pass that runs after callDestructors and uses gatherLoopDetails. This PR updates gatherLoopDetails in order to function at this slightly later point in compilation.

- [x] full local testing

### DIFF
--- a/compiler/resolution/loopDetails.cpp
+++ b/compiler/resolution/loopDetails.cpp
@@ -431,18 +431,21 @@ void gatherLoopDetails(ForLoop*  forLoop,
           INT_ASSERT(actualSe); // otherwise not normalized
           // Find the single definition of actualSe->var to find
           // the call to _getIterator.
-          Expr* iterable = NULL;
+          Expr* iterable = actualSe;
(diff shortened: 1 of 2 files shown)


compiler/resolution/loopDetails.cpp
```diff
@@ -146,6 +146,49 @@ Symbol* collapseIndexVarReferences(Symbol* index)
   return index;
 }
 
+/* Return whatever is producing the Symbol iterator.
```
vasslitvinov: How is this function specific to iterators?

vasslitvinov: If it is not specific to iterators, it would be nice to have it as a globally-visible function, perhaps in symbol.h.

mppf: Most of the work is actually in Symbol::getInitialization. But the way it ignores FLAG_TEMP variables is specific to what we actually need here and I'm not so confident it's broadly useful. We can certainly move it to symbol if we find more cases needing the same pattern, though.

Issue #8375, opened 2018-02-03
lifetime checking: implicit coercions between kinds class pointers


mppf:
It seems likely that we'll want to allow implicit coercion from an 'owned' to a 'borrow'.

Allowing an implicit coercion from 'borrow' to 'owned' would break the analysis.

Do we want this implicit coercion? Or any others related the effort?

psahabu:
There might be cases where the user wants to pass ownership to another scope, which this coercion would preempt. If we go forward with this coercion, then there needs to be a way to express the other pattern.

mppf:
> There might be cases where the user wants to pass ownership to another scope, which this coercion would preempt. If we go forward with this coercion, then there needs to be a way to express the other pattern.

There already is a way to express that. Say you want to pass ownership from one function to another function that it calls. The called function just needs to take in the value to "get" ownership from by `in` intent (or possibly `ref` if it wants to control when ownership is taken). Another way to put it is, if `Owned` is the way we represent an owned class instance, `Owned` already includes methods to "move" ownership, including the `=` operator.

mppf:
If we have

```
class Parent { }
class Child : Parent { }
```

We also need `Owned(Child)` to be a subtype of `Owned(Parent)`

psahabu:
> The called function just needs to take in the value to "get" ownership from by in intent (or possibly ref if it wants to control when ownership is taken).

I'm good with `in` intent being the way to pass ownership, with `default` being coerced into a borrow.

> We also need Owned(Child) to be a subtype of Owned(Parent).

I'm pretty sure my programming languages professor literally jumped up and down to emphasize that `U` being a subtype of `T` does not imply that `W(U)` is a subtype of `W(T)`. That said, given that `Owned` doesn't have functionality beyond tracking ownership, this coercion does make sense.

vasslitvinov:
It makes a good sense to make a <lifetimemodifier>(Child) to subtype the same modifier on Parent.

Yes for owned->borrow conversion implicitly.

I am not excited about 'in' intent to mean "pass ownership" implicitly. 'const in' is the default intent and it would be messy to explain that the default intent means one thing and 'const in' another.

> The called function just needs to take in the value to "get" ownership from by in intent (or possibly ref if it wants to control when ownership is taken).

How are we going to check the lifetime when a class pointer is passed by ref then the callee takes over the ownership? We'd need to look inside the function, at the very least, which violates "interface invariance".

[...]

PR #8583, opened 2018-02-23
Enable Owned(C)->C coercions

### LINKED ISSUE #6045: Surprising type mismatch with Owned, matches method
### Summary of Problem

It is surprising that, when calling a method on a class wrapped in an Owned, that `this` gets the type of the class but a generic argument does not.

### Steps to Reproduce

**Source Code:**

```chapel
use OwnedObject;

class C {
  var x: int;

  proc matches(other) {
    if this == other then
      writeln("same!");
    else
      writeln("different!");
  }
}

proc main() {
  var a = new Owned(new C(1));
  var b = new Owned(new C(2));

  a.matches(b);
  a.matches(a);
}
```

test/classes/delete-free/owned-comp.chpl 

### Discussion

The example above shows the limits of our current forward-but-don't-coerce strategy for Owned.

Not sure how to "fix" this without user-defined-coercions for Owned, but the obvious solution is to write

``` chapel
a.matches(b.borrow());
a.matches(a.borrow());
```

``` chapel
```

mppf:
This PR enables implicit conversions from Owned(C) to C.

In order to inform compiler about Owned, move it internal, and create a flag marking Owned and Shared types.

Additionally, don't add default record cast functions for Owned in order to reduce confusion in error messages.

Note, this PR adds dtOwned, which is not used yet, but will be in work adjusting 'new' to return 'Owned' (a future PR).

Relates to issue #8375.
Closes #6045.


mppf:
[Summary: #Successes = 7422 | #Failures = 794 | #Futures = 0]
there seems to be a problem with the adjustment to autoDestroy overloads

mppf:
This PR is replaced by #8791.

Issue #8474, opened 2018-02-14
Name executable after file containing main module, not main module itself?


bradcray:
In issue #6283 / PR #6811, we changed the default executable name created by the compiler
from `a.out` to the name of the main module.  Every Chapel program defines a unique main
module so this results in an unambiguous name, and often the main module takes its name
from the file containing it, so in those cases, this results in `chpl testit.chpl` creating an 
executable named `testit`.

However, when the main module and the filename don't match, this can cause significant
confusion, which has led to the following counter-proposal:  Perhaps we should be naming
the executable after the file containing the main module rather than the main module itself.
For a case like the following:

MyProgram.chpl:
-----
```chapel
module MyModule {
  proc main() { ... }
}
```

```
chpl MyProgram.chpl
```

this would result in an executable named `MyProgram` rather than `MyModule`.  This should avoid some of the cases of confusion that have been reported or experienced in the current
approach.

I'm curious whether people see downsides / arguments against taking this approach.  And
whether there's precedent in other languages for either our current approach or this
counterproposal that we should consider following.

bradcray:
My only hesitation about this proposal is what the behavior should be if/when we add a
Chapel-level equivalent of `include` (that is, if the main module is broken across multiple files).  
My instinct would be to use the name of the file that the `module M` declaration itself appeared
in.

bradcray:
@ben-albrecht may be supportive of this based on his comments back on issue #6283.

ben-albrecht:
> @ben-albrecht may be supportive of this based on his comments back on issue #6283.

Yes, very supportive.

ben-albrecht:
> My only hesitation about this proposal is what the behavior should be if/when we add a
Chapel-level equivalent of include (that is, if the main module is broken across multiple files).
My instinct would be to use the name of the file that the module M declaration itself appeared
in.

I agree with this. There should always be a top-level file that contains the module M declaration or is the implicit main module.

bradcray:
I ran into this yet again last night, so am giving myself a thumbs-up on the proposal.

lydia-duncan:
I remain vaguely uneasy that we had a discussion on this aspect so recently, but I don't have anything specific against this particular choice.  So maybe take this comment as "we really need to be more careful about our design decisions overall".

bradcray:
[...]

PR #8587, opened 2018-02-24
Base generated executable's name on main module's filename

bradcray:
This PR implements the proposal in issue #8474.

In issue #6283 / PR #6811, we changed the default executable name
created by the compiler from a.out to the name of the main
module. Every Chapel program defines a unique main module so this
results in an unambiguous name, and often the main module takes its
name from the file containing it, so in those cases, this results in
chpl testit.chpl creating an executable named testit.

However, when the main module and the filename don't match, this can
cause significant confusion, which has led to the following
counter-proposal: Perhaps we should be naming the executable after the
file containing the main module rather than the main module itself.


test/compflags/ferguson/default-binary-name/test-default-binary-name2.chpl
```diff
@@ -0,0 +1,7 @@
+module M1 {
+  use M2;
+  
+  proc main() {
+    writeln("Hello");
+  }
+}
```
mppf: Needs newline at end of file

test/compflags/ferguson/default-binary-name/test-default-binary-name3.chpl
```diff
@@ -0,0 +1,3 @@
+module M1 {
+  use M3;
+}
```
mppf: Needs newline at end of file

bradcray:
I'm realizing belatedly that this PR needs some doc updates as well.  Will take that on before merging to keep it all together.

test/compflags/ferguson/default-binary-name/test-default-binary-name3.chpl
bradcray: I don't know why you guys are so obsessed with spending precious disk space on final newlines... :)

doc/rst/usingchapel/compiling.rst
```diff
@@ -50,9 +50,9 @@ We note some of the most useful flags here:
 =====================   ======================================================
 Flags                   Description
 ---------------------   ------------------------------------------------------
-``-o <filename>``       specify the filename of the generated
-                        executable, otherwise the main module name is
-                        used by default
+``-o <filename>``       specify the filename of the generated executable,
+                        otherwise the filename of the main module (minus
```
mppf: 'filename of the main module' seems strange to me, but what about 'name of the file containing the main module' ?

test/compflags/ferguson/default-binary-name/sub_test
```diff
@@ -80,8 +80,11 @@ def run(source, exe_name, compopts, expect):
 
 run("test-default-binary-name.chpl", "test-default-binary-name", [], "Hello")
 run("test-default-binary-name.chpl", "test_exe", ["-o", "test_exe"], "Hello")
-run("test-default-binary-name-main-module.chpl", "M1", ["--main-module", "M1"], "M1 main")
-run("test-default-binary-name-main-module.chpl", "M2", ["--main-module", "M2"], "M2 main")
+run("test-default-binary-name-main-module.chpl", "test-default-binary-name-main-module", ["--main-module", "M1"], "M1 main")
+run("test-default-binary-name-main-module.chpl", "test-default-binary-name-main-module", ["--main-module", "M2"], "M2 main")
+run("test-default-binary-name2.chpl", "test-default-binary-name2", [], "Hi\nHello")
+run("test-default-binary-name3.chpl", "test-default-binary-name3", [], "Hi")
```
mppf: The difference between
```
chpl test-default-binary-name3.chpl
```

and
```
chpl test-default-binary-name3.chpl M3.chpl
```

is surprising to me. It seems that if there is only one .chpl file present on the command line, Chapel considers that file the main module, even if `main` is defined in another module. I was imagining that the default "main module" would be "the module defining main". I think I've always thought it worked this way, but it does not.

As far as how to proceed I think:

 1. I'd recommend merging this PR as it's a good improvement that's been discussed. This potential misunderstanding is not new with this PR (the same test would have similar behavior on master).
 2. I think we need to either change this property of main module or document it better in the language specification. We could possibly consider renaming "main module" to "primary module" e.g. to reduce confusion.
 3. Other people involved in the discussion on the issue #8474 might appreciate a "heads up" about this.

bradcray: You're right that the question of identifying / defining the main module is independent of this PR and hasn't changed for some time.  I.e., we could change how the main module is defined or identified and this PR would still work and be documented correctly.  To describe how we got here:

Given that Chapel programs can just contain top-level code and, in that case, are given an implicit empty main, as in:

```chapel
writeln("Hello, world!");
```

we felt that it would be too confusing to say that if your simple Chapel program happened to `use` a module that defined a routine named `main()` that module would hijack the execution from you if it were not named on the command-line.  I.e., imagine I evolved the above program to:

```chapel
use SomeLib;
writeln("Hello, world!");
someLibCall();
```

and compiled it with:

```
chpl hello.chpl
```

If `SomeLib` ended up having a routine called `main()`, it'd be pretty surprising as a user to have the behavior of my program do something beyond printing hello world and making the obvious library call.

Now, arguably, a good library shouldn't define a `main()` and this is something we could check for, but that's how we got here.  We also rationalized it slightly by sometimes relying on a `main()` in a library module being used as a way to run standalone unit tests on that library, lacking another unit testing strategy and being reluctant at that time to add alternative ways of specifying the program's entry point (where some ideas included a `--main-function=unitTest()` flag or a well-defined entry point `proc unitTest()` that would be used instead of `proc main()` if compiling with a `--unitTest` flag).

Anyway, I don't disagree that there's the potential for confusion here, though I also don't think the current strategy is all that bad either (a potential piece of evidence being how long you've lived with Chapel without ever being bitten or surprised by it :) ).

Anyway, since we agree, I'll continue with this PR after making the documentation update you requested.

mppf: The rationale for the current "main module" behavior sounds reasonable - I just think it is not clear from the language specification how nuanced it is. i.e. we need documentation improvement.

cassella:
BTW, was it intentional that this changed the behavior of ``chpl foo/bar/baz.chpl`` from producing the executable ``./baz`` to producing ``./foo/bar/baz``?

bradcray:
Ooh, good question.  I think you're right that we should not be doing that.  Thanks for pointing it out.

PR #8593, opened 2018-02-25
Directly create aBarrier classes in the AllLocalesBarrier

ronawho:
Instead of using the record wrapper, directly create the atomic barrier class.
This should avoid any potential dynamic dispatch, which was causing an issue
for --baseline testing.

Avoiding dynamic dispatch also slightly improves the performance of this
barrier, though not significantly since a majority of the time is spent
communicating. For the 16-node micro-benchmark that does 100,000 barriers this
this shaves a 1/3 of a second off (6.5s -> 6.2s)


ronawho:
@benharsh mind reviewing? 

This was originally motivated by working around a `--baseline` failure that I'm still trying to create a smaller reproducer for. However, since it helps performance and gets rid of one of the hacky initializers it seems like the right thing to do anyways.

Issue #6007, opened 2017-04-13
Path library - add support for commonPrefix


lydia-duncan:
Back in late 2014, we started undertaking the implementation of the FileSystem and Path libraries, with input from the user community ([start of thread](https://sourceforge.net/p/chapel/mailman/message/32800541/), [mail with updated proposal after feedback](https://sourceforge.net/p/chapel/mailman/message/32902710/)) Much progress has been made in getting the support onto master, but 10 functions still remain unimplemented due to the youth of our string implementation at the time.  With our string implementation in a good state, these functions should be returned to and implemented

This issue is for the function `commonPrefix`.

Here is the expected documentation and declaration:
```chapel
/* Determines and returns the longest common path prefix (taken character-by-character) of
   all the string pathnames provided. If no arguments are specified, returns
   the empty string.

   :arg paths: Any number of paths
   :type paths: `string`

   :return: The longest common path prefix
   :rtype: `string`
*/
proc commonPrefix(paths: string ...?n): string;
```

lydia-duncan:
Open question: would a version of this function which takes an array argument be useful or desired?

lydia-duncan:
Tests will need to be written for this feature

Unnati-P:
Hi @lydia-duncan! 
My partner and I want to apply for RGSoC in your organization and I would like to work on this feature.
I feel implementing the function by taking an array of strings as argument would be useful.

lydia-duncan:
Hi @Unnati-P!  I'm still digging my way out of email from the holiday weekend, I will try to get back to you soon!

lydia-duncan:
Sorry about the delay.  It would be great to have you and your partner work on this feature!  I'm happy to provide guidance (though I may be a bit slow to respond at times, sorry :().

Unnati-P:
Hi @lydia-duncan !
We have worked out a possible solution for implementing this feature and created a process commonPrefix. We have not done a character-by-character comparison, as the feature involves finding common prefix of paths rather than strings.
```javascript
proc commonPrefix(paths: []): string {

var prefix: string = "";    // result string
var itr: string;
var inputLength = paths.size;   // size of input array
var shortestPath = paths[1];
var i: int = 1;
var ind: int = 1;
var delimiter: string;
var shortestPathIndex: int = 1;
var prefixArray:[1..1] string;    // array of resultant prefix string

// if input is empty, return empty string.
// if input is just one string, return that string as longest common prefix path.

if inputLength==0 then {
[...]

PR #8595, opened 2018-02-25
add commonPath to Path.chpl

Unnati-P:
In reference to issue #6007 
TODO:

- [x] Adding more test cases

- [x] Varargs implementation

Hi @lydia-duncan ! Sorry for the delay in getting back. 
We have made some changes as you had suggested. It would be great if you could clear a few doubts:

- Can you elaborate what you meant by passing a `range`.

- Is it possible to use zippered iterations for unequal ranges?
Please look into the code and suggest the changes we can make.
We're working on passing varying number of arguments to function and will commit accordingly.


pri2810:
Hi!!
The above function works fine for the following test Cases:
```javascript
	Input:"Users\\templates\\a\\b","Users\\templates\\a\\b\\c","Users\\templates\\a\\"
	Output: Users\templates\a\
		
	Input:"Users\\templates\\a\\b","Users\\templates\\a\\b\\c","/Users/templates/a/"
	Output: null string
	
	Input:"/Users/templates/a/b","/Users/templates/a/b/c","/Users/templates/a"
	Output: /Users/templates/a/
		
	Input:"/Users/templates/a/b","Users/templates/a"
	Output: null string
		
	Input: "/Users/templates/a/b","/Users/examples"
	Output: /Users/
	
	Input: "/Users/templates/a/B","/Users/templates/a/B/C","/Users/templates/a/B/D/C"
	Output: /Users/templates/a/B/
```
In the above cases, "\\" is used for '\'

lydia-duncan:
I've seen this PR, will try to get to reviewing it in the next couple of hours.  Thanks! :)

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
       return ret;
     }
 
+    /* This function takes an array of stings as input.
```
lydia-duncan: I think it would be more appropriate to put this function in modules/standard/Path.chpl, since its behavior is dependent on paths rather than strings themselves (i.e. you aren't performing a character-by-character comparison, you are comparing based on file path and directory chunks)

lydia-duncan:
What should the behavior be when a file is provided?  For example:
Input: "/Users/foo.chpl", "/Users/foo.good"
Output: "/Users/foo" or "/Users"?
(We try to guide our choices for these functions based on what Python does in this situations, so I would verify the behavior against that of Python)

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+
+    returns:  The longest common prefix path of all the input paths
+    */
```
lydia-duncan: I'd prefer you to use the comment specified in the issue as the documentation for this function (though you will need to update for the array argument)

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+    proc commonPrefix(paths: []): string {
+
+    var prefix: string = "";    // result string
+    var inputLength = paths.size;   // size of input array
+    var firstPath = paths[1];
```
lydia-duncan: This assumes that the array provided is 1's based.  However, that may not be the case, as Chapel arrays can start at any index.  You could also be provided a slice of an array.
The array's domain has a method `first` that you can use like this:
```chapel
var firstPath = paths[paths.domain.first];
```

Since this will impact other parts of your algorithm, I recommend looking at the overall correctness in other places first, then coming back and extending the function in this way

lydia-duncan: (Note also that the varargs version will always have the same index set, I believe)

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+    pos = prefixArray.size;
+    minPathLength = prefixArray.size;
+
+    for i in 2..inputLength do {
+
+      cnt = paths[i].count(delimiter);
+      var tempArray:[1..cnt+1] string;    // temporary array storing the current path under consideration
+      tempArray = paths[i].split(delimiter, -1, false);
+
+      var minimum = min(prefixArray.size, cnt+1);
```
lydia-duncan: In answer to your question about zippering, it looks like we only support zippering of different lengths when one is unbounded (see [here](https://chapel-lang.org/docs/latest/users-guide/base/zip.html?highlight=zippering#zippered-iterand-size-matching) for more information on what that means).  The documentation mentions being able to send in a size argument to the iterator, but I'm not seeing an example of how to do that quickly.  This seems like a reasonable question to ask either in the gitter/irc chatroom or on Stack Overflow, as someone else probably knows the answer to that.

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+
+    if firstPath.find("\\", 1..firstPath.length) == 0 then {
+      delimiter = "/";
+    } else {
+      delimiter = "\\";
+    }
+    cnt = firstPath.count(delimiter);
+
+    var prefixArray:[1..cnt+1] string;    // array of resultant prefix string
+    prefixArray = firstPath.split(delimiter, -1, false);
```
lydia-duncan: You could just declare this all in one go (and wouldn't need to specify the exact array being created), but since the line would get a bit long it's up to you.
```chapel
var prefixArray = firstPath.split(delimiter, -1, false);
```

lydia-duncan: This would remove the need for calling `count` above, as then you could just get `cnt` from the size of the prefix array created

lydia-duncan: (Or even make it a local variable to the `for` loop below)

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+      tempArray = paths[i].split(delimiter, -1, false);
+
+      var minimum = min(prefixArray.size, cnt+1);
+      if minimum < minPathLength then {
+        minPathLength = minimum;
+      }
+
+      for (itr, ind) in zip(1..minimum, 1..minimum) do {
+        if tempArray[itr]!=prefixArray[ind] then {
+          pos = ind;
```
lydia-duncan: Shouldn't this be a check of the minimum between the old value for `pos` and the proposed new value?  I suspect you would run into problems with this input set: "/a/b/c", "/a/d", "/a/b/e"

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+          break;
+        }
+      }
+    }
+
+    if pos < prefixArray.size then {
+      prefixArray.remove(pos..prefixArray.size);
+      prefixArray.push_back(" ");
+    }   else {
+      prefixArray.remove(minPathLength+1..prefixArray.size);
```
lydia-duncan: I'm not sure I understand the purpose of this `else` branch.  Why would you remove something if the first path is entirely a prefix of all the other ones?

modules/internal/String.chpl
```diff
@@ -427,6 +427,75 @@ module String {
...
+        if tempArray[itr]!=prefixArray[ind] then {
+          pos = ind;
+          break;
+        }
+      }
+    }
+
+    if pos < prefixArray.size then {
+      prefixArray.remove(pos..prefixArray.size);
+      prefixArray.push_back(" ");
```
lydia-duncan: There's probably something I'm missing, but why are you adding a " " here?

lydia-duncan:
Let's also have you transform the input and output combinations you've written into a test file (see the guidance [here](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst) for how to do that.  Let me know if you find it confusing, there's definitely things we've been meaning to do to improve that document)

lydia-duncan:
Thanks again for all your hard work!  I look forward to seeing the next round of changes :)

modules/internal/String.chpl
pri2810: As the function aims at finding the longest common path prefix, thus we had performed word-by-byword comparison. It has been added to modules/standard/Path.chpl as mentioned by you.
Also, it is not like Python's feature as we have returned the common path in all the paths provided by the user and have not done it character by character. But the input is an array of strings here (Python has lists). Python feature returns invalid paths but this feature doe not.

modules/internal/String.chpl
pri2810: Thank you for mentioning this feature. We have changed the code accordingly for arrays

modules/internal/String.chpl
pri2810: Thanks for mentioning this feature.

modules/internal/String.chpl
pri2810: Thank you for mentioning the bug, we have looked into it and updated the condition accordingly.

modules/internal/String.chpl
pri2810: It's for making a valid path. By adding the " " at the end of the array, we can add the delimiter at that position while using the join function.

modules/internal/String.chpl
pri2810: The else part is present because if the pos was never updated then it must be the case that the one of the path (which is not the first path in the array) could have been subset of the all the paths provided, thus we remove the array elements from minPathLength+1.

lydia-duncan:
Also, don't forget to both submit a contributor [agreement](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements)

modules/internal/String.chpl
lydia-duncan: > Also, it is not like Python's feature as we have returned the common path in all the paths provided by the user and have not done it character by character. But the input is an array of strings here (Python has lists). Python feature returns invalid paths but this feature does not.

I think that's a reasonable change to the Python implementation.  However, it is different from the specification that was discussed when we first proposed this function - that means it would be best to bring this up for discussion to ensure that is an acceptable change to the community at large (instead of just me :) ).

I would recommend commenting on the trade-offs of this decision on the original issue and sending an email to the developers and users mailing lists asking for comment on the issue with this proposal in mind.  One such trade-off would be the handling of the file prefix test case I mentioned - it's still supportable if you think that's reasonable, but would require more effort to do so

modules/internal/String.chpl
lydia-duncan: Oh, I see.  Because you've had to limit the iteration of the `for` loop to only traverse the minimum of the two paths being compared, it is possible not to have updated `pos`.  Got it.  I think that would be worth a comment explaining, since I found it confusing to follow when just looking at the code

modules/standard/Path.chpl
```diff
@@ -286,4 +286,80 @@ proc file.realPath(): string throws {
...
+  var prefixArray = firstPath.split(delimiter, -1, false);    // array of resultant prefix string
+
+  var pos = prefixArray.size;   // rightmost index of common prefix
+  var minPathLength = prefixArray.size;
+
+  for i in (start+1)..end do {
+
+    var tempArray = paths[i].split(delimiter, -1, false);   // temporary array storing the current path under consideration
+
+    minimum = min(prefixArray.size, tempArray.size);
```
lydia-duncan: `minimum` could be local to the `for` loop, right?

modules/standard/Path.chpl
```diff
@@ -286,4 +286,80 @@ proc file.realPath(): string throws {
...
+
+    if minimum < minPathLength then {
+      minPathLength = minimum;
+    }
+
+    for (itr, ind) in zip(1..minimum, 1..minimum) do {
+      if (tempArray[itr]!=prefixArray[ind] && ind<=pos) {
+        pos = ind;
+        flag=1;   // indicating that pos was changed
```
lydia-duncan: If you're not updating the `prefixArray` you could just compare `pos` against its size instead of using the `flag` variable, right?

modules/standard/Path.chpl
```diff
@@ -286,4 +286,80 @@ proc file.realPath(): string throws {
...
+        break;
+      }
+    }
+  }
+
+  if (flag == 1) {
+    prefixArray.remove(pos..prefixArray.size);
+    prefixArray.push_back(" ");
```
lydia-duncan: Your updated comment on the old code seems to indicate this should be removed but it looks like it wasn't.

modules/standard/Path.chpl
```diff
@@ -286,4 +286,80 @@ proc file.realPath(): string throws {
...
+
+    var tempArray = paths[i].split(delimiter, -1, false);   // temporary array storing the current path under consideration
+
+    minimum = min(prefixArray.size, tempArray.size);
+
+    if minimum < minPathLength then {
+      minPathLength = minimum;
+    }
+
+    for (itr, ind) in zip(1..minimum, 1..minimum) do {
```
lydia-duncan: Sorry I didn't catch this earlier, but `itr` and `ind` should always be the same value, so you don't need both of them to access into tempArray and prefixArray, just one.  I think it was a hold-over from when you were iterating over the contents of those arrays, which would be different.

lydia-duncan:
I think this is really close.  Unfortunately the discussion I mentioned will need to happen before the code can get merged, but assuming that goes well the implementation itself should be ready to go after the minor nits.  But I do want a submitted test file :)

modules/standard/Path.chpl
Unnati-P: For input like ['/Users/a','/Users/b'], `pos>=prefixArray.size` but removal needs to be done from `pos..prefixArray.size`. Thus, we realise it's a case where `pos` is being updated but due to `pos==prefixArray.size`, it goes to the wrong condition.
So it is now simply checking that if pos was changed, remove starting from `pos` else the strings are subsets so remove starting from `minPathLength`.

lydia-duncan: I think going from pos to prefixArray.size assumes that prefixArray.size == prefixArray.last, but that won't always be the case (you'll need to create an array with a different index set to verify this)

lydia-duncan: You will want to check against prefixArray.domain.last, but perhaps the initial value for pos should be set higher than the highest index to avoid confusion

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
```diff
@@ -0,0 +1,27 @@
+/*
+Test cases to validate the commonPrefix([]) in Path.chpl
+*/
+
+use Path;
+
+//paths=null
+writeln(commonPrefix([""]));
```
lydia-duncan: What happens if you send in an empty array?

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
```diff
@@ -0,0 +1,27 @@
...
+writeln(commonPrefix(["/home/user/Documents/foo/bar","/home/user/Documents/foo","/home/user/Documents/foo/bar/baz"]));
+
+//testing when one path is absolute and the other is relative
+writeln(commonPrefix(["/home/user/Documents/foo/bar","home/user/Documents/foo/bar"]));
+
+//Only a substring is common
+writeln(commonPrefix(["/home/user/Documents/foo/bar","/home/user/Desktop/bar","/home/user/Documents/bar"]));
+
+//only relative paths are provided
+writeln(commonPrefix(["../Documents/foo/bar","../Documents/foo/bar/baz","../Documents/foo"]));
```
lydia-duncan: Good thinking!  What about `["/bar/baz/../foo", "/bar/foo"]`?  Do we want to behave specially for this case?  Technically their path is the same since `/baz` is removed by the `/..`

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
```diff
@@ -0,0 +1,27 @@
...
+
+
```
lydia-duncan: We try not to leave so much extra space at the end of files, but it is not a big deal

lydia-duncan:
(the "needs testing" label I just added was for me - I will run your branch through `start_test` prior to merging it, but I expect things to work)

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
pri2810: The compiler throws an error: "syntax error: near ']'"

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
pri2810: The code will return /bar as the common path. Even the os.path.commonpath version of Python behaves similarly. 

The point raised by you is correct but for such behaviour, we may have to get the realPath of all the paths given.

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
lydia-duncan: I think you'll need to create an empty array and then send that array to the function.  Empty arrays have domain `1..0` (so a declaration of one would look like this: `var a1: [1..0] int; writeln(commonPrefix(a1));`).

I'm also realizing we didn't add the test case to ensure the array was not modified that I mentioned a while back, which you would need a local variable array.

modules/standard/Path.chpl
```diff
@@ -288,17 +288,79 @@ proc file.realPath(): string throws {
...
+  } else if inputLength == 1 then{
+    return firstPath;
+  }
+
+  var prefixArray = firstPath.split(pathSep, -1, false);    // array of resultant prefix string
+
+  var pos = prefixArray.size;   // rightmost index of common prefix
+  var minPathLength = prefixArray.size;
+
+  for i in (start+1)..end do {
```
lydia-duncan: Since `end == n`, and in the varargs version `start+1` will always be `2`, it would be safe to use those values directly for this loop instead of defining `start` and `end` (with a comment to ensure it is clear)

test/library/standard/Path/prithvi-unnati/commonPath/testCommonPathStringArguments.chpl
```diff
@@ -0,0 +1,26 @@
+/*
+Test cases to validate the commonPath(paths: string ...?n) in Path.chpl
+*/
+
+use Path;
+
+//paths=null
+writeln(commonPath(""));
```
lydia-duncan: Similarly to the empty array, it would be nice to test `commonPath()` (note: no arguments provided)

pri2810: But as the function is defined to have ```paths: string``` as an argument, the compiler throws an error for ```commonPath()```

lydia-duncan: Huh, I must not have understood varargs correctly, I thought there was a way for zero args to be specified.  Not a big deal if not, but you ought to be able to remove a condition in your code if so

test/library/standard/Path/prithvi-unnati/commonPrefix/testCommonPrefix.chpl
pri2810: I have added the testcase accordingly.

lydia-duncan:
Thanks!  I'll run testing over this branch and let you know how it goes!

lydia-duncan:
Do you have a specific commit message you would like me to use should I merge this soon, per [this guidance](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/ContributorInfo.rst#final-merge-message)?

Unnati-P:
Add commonPath to Path.chpl

Implementation of commonPrefix function, later renamed commonPath, as it takes a list of string pathnames as input and returns longest common path prefix which is a valid path.

lydia-duncan:
Passed testing, thanks!  Going to merge shortly

PR #8598, opened 2018-02-26
Fix handling of default value of argument with record type (with initializers)

lydia-duncan:
Prior to this change, we would get an error when using the default value
of an argument if it was a `new` on a record with initializers, due to
thinking the type of the argument was void instead of the record type. I.e.

```chapel
proc foo (x = new Rec()) { ... }
// Where Rec is a record that defines initializers
```

This was because when the type being created is a record with default initializers,
the return value is void but that's not what we want to use to determine the type
of the argument.  So, when handling resolveNew for a concrete record, if we
know we're the last statement of a block for an ArgSymbol, add a SymExpr to the
temporary we send in to the initializer, so that we can get the right type back.

This is a hack and I am not proud of it, but I'm not sure there's a better way,
aside from redoing argument handling.

Adds five tests to verify the behavior of this change.  Two are for explicit
functions with default arguments (one for an initializer, one for an ordinary
function), the third is for default initializers when the field is declared with
an explicit `new` call, and the last two are for explicit functions with
default arguments when the record is generic (one for an initializer, one for
an ordinary function).  All three failed prior to this change and now pass.

Passed:
- classes/initializers with futures
- full std paratest
- paratest with --force-initializers showed reasonable changes and a few
resolved tests


compiler/resolution/functionResolution.cpp
```diff
@@ -5864,6 +5864,11 @@ static void resolveNewHandleNonGenericInitializer(CallExpr*      call,
     if (isBlockStmt(call->parentExpr) == true) {
       call->insertBefore(def);
 
+      if (isArgSymbol(call->parentSymbol) &&
```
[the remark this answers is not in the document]
lydia-duncan: This one is already in an if statement that eliminates classes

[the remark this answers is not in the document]
lydia-duncan: (but I will add the comment)

compiler/resolution/functionResolution.cpp
```diff
@@ -5864,6 +5864,16 @@ static void resolveNewHandleNonGenericInitializer(CallExpr*      call,
...
 
+      // Provide the return type for new expressions in default arguments to
+      // other parts of resolution. (Because the `new` expression is top level
+      // for argument symbols, we would fail to insert the temporary needed for
+      // argument analysis to appropriately determine the type without inserting
+      // this SymExpr directly)
+      if (isArgSymbol(call->parentSymbol) &&
+          toBlockStmt(call->parentExpr)->body.tail == call) {
+        call->insertAfter(new SymExpr(newTmp));
+      }
```
mppf: Besides the question about at->isClass here, now with the comment, it's turning out to be enough lines it might be worth adding a function, where you can put the comment once.

lydia-duncan: Classes don't run into this problem, because their "init" call is wrapped in a call to "_new", which does return the type.  But I definitely should make a function.

lydia-duncan: So, the helper function I made currently doesn't have the isClass() check or a check if the parent of the call is a BlockStmt.  Both of these are necessary for the insertion to be valid, but the BlockStmt guard is already there for all the additions, and the isClass guard is there for all (albeit more obviously for 2 of the 3).  Should I move both of them into the helper and place the call at a more top level location in these functions?  Just the class one?

lydia-duncan: Imma do both

Issue #6009, opened 2017-04-13
Path library - add support for expandVars


lydia-duncan:
Back in late 2014, we started undertaking the implementation of the FileSystem and Path libraries, with input from the user community ([start of thread](https://sourceforge.net/p/chapel/mailman/message/32800541/), [mail with updated proposal after feedback](https://sourceforge.net/p/chapel/mailman/message/32902710/)) Much progress has been made in getting the support onto master, but 10 functions still remain unimplemented due to the youth of our string implementation at the time.  With our string implementation in a good state, these functions should be returned to and implemented

This issue is for the function `expandVars`.

Here is the expected documentation and declaration:
```chapel
/* Expands any environment variables in the path of the form `$<name>` or
   `${<name>}` into their values.  If <name> does not exist, they are left
   in place. Returns the path which includes these expansions.

   :arg path: a string representation of a path, which may or may not include
                   `$<name>` or `${<name>}`.
   :type path: `string`

   :return: `path`, having replaced all references to environment variables with
                their values
   :rtype: `string`
*/
proc expandVars(path: string): string;
```

lydia-duncan:
Tests will need to be written for this feature

lydia-duncan:
Resolved by #8617

Issue #5939, opened 2017-04-04
Env module


ben-albrecht:
## Motivation
To manage environment variables from Chapel today, a user must use `extern proc setenv()` / `extern proc getenv()`, as used in #5924. I propose we add native support for these features through an `Env` module. 

Note that many other languages support env-var features as part of their `os` libraries. However, Chapel supports the majority of these `os` features in the `FileSystem` and `IO` modules, therefore we only need an `Env` module, rather than a more general `OS` module.

## Features

- set env var
- get env var (with optional default value, if not set)
- check if env var is set (could be done through checking result of `getEnv()`)
- iterate over env vars
- save the state of environment into a variable (e.g. associate array)
- handle locality in a smart way
    - Distinguish environments on different locales
    - Distinguish launcher environment vs. compute environment

mppf:
I'd call the module Environment or EnvironmentVariables or EnvVariables or something longer like that and reserve Env for a type the module might (one day) define.

gbtitus:
Note sure about "Distinguish launcher environment vs. compute environment".  We don't run Chapel code in the launchers, so any support in the modules only has to work on the compute nodes.  You could expand this Issue to be about environment support in general (to include for example what @bradcray said about cleaning up environment forward from launch side to compute side), but then it'll be about more than just a module.

bradcray:
> Distinguish launcher environment vs. compute environment

I assumed this was meant to be a placeholder for the issue I tried to flag on email about whether a user might/should expect that there'd be any relationship between the environment on a given locale and the environment on the node from which they launched a multi-locale job (but I agree it's not clear in the current wording whether that's what this item is about).

ben-albrecht:
@bradcray - I tried to capture the point you had brought up earlier, but may have misunderstood it - dropping the bullet for now, and updated the issue with some more details about locality-oriented design decisions.

cassella:
I tried to take a stab at this, but couldn't get a reference to the C ``char **environ`` in a gcc- and ``modules/``-friendly manner due to #7983.

bradcray:
@cassella: Could using void pointers (`c_void_ptr`) and casting be used as a workaround?

cassella:
[...]

PR #8617, opened 2018-02-27
expandVars function added to Path library

verma-varsha:
This PR addresses the issue #6009 
However, there was one problem I faced while testing this function. The compiler kept throwing  ``error: 'CHPL_WIDE_POINTERS' undeclared (first use this function)`` while this environment variable is given in the list here:  https://chapel-lang.org/docs/latest/builtins/internal/ChapelEnv.html
That is why, you might notice that I have commented out the CHPL_WIDE_POINTERS environment variable in Path.chpl
@lydia-duncan Please review this PR and guide me about any changes to be made.


lydia-duncan:
Thanks for your work!  I'll try to get to reviewing this quickly (but may get held up, feel free to poke me if you don't hear from me in the next 24 hours).  In the meanwhile, have you submitted a contributor [agreement](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements)?

verma-varsha:
Yes @lydia-duncan , I've submitted the CLA on the said email id.

cassella:
I'm not on the Chapel team, but I'd recommend testing expanding the string 
```
bin/$CHPL_COMM/$CHPL_COMM_SUBSTRATE/
```

verma-varsha:
There's no harm in adding a new test. I'm on it!

test/library/standard/Path/verma-varsha/test_expandVars.chpl
```diff
@@ -0,0 +1,7 @@
+use Path;
+
+var x1:string = "$CHPL_HOME/bin/${CHPL_HOST_PLATFORM}";
+assert(expandVars(x1) == CHPL_HOME+"/bin/"+CHPL_HOST_PLATFORM);
+
+var x2:string = "bin/$CHPL_COMM/$CHPL_COMM_SUBSTRATE/";
+assert(expandVars(x2) == "bin/"+CHPL_COMM+"/"+CHPL_COMM_SUBSTRATE+"/");
```
cassella: This assert fails when I run it.

verma-varsha: I don't know what's going wrong. It's passing all the tests on the Travis CI as well as on my machine. :(

cassella: Ah, I see.  I hadn't noticed that your most recent commit not only added the test, but also addressed the bug in the function.  I was running it with the old version of the function.

It works for me as well with the new version of the function.

modules/standard/Path.chpl
```diff
@@ -286,4 +286,55 @@ proc file.realPath(): string throws {
...
+                   ("CHPL_MAKE", CHPL_MAKE),
+                   ("CHPL_ATOMICS", CHPL_ATOMICS),
+                   ("CHPL_NETWORK_ATOMICS", CHPL_NETWORK_ATOMICS),
+                   ("CHPL_GMP", CHPL_GMP),
+                   ("CHPL_HWLOC", CHPL_HWLOC),
+                   ("CHPL_REGEXP", CHPL_REGEXP),
+                   //("CHPL_WIDE_POINTERS", CHPL_WIDE_POINTERS),
+                   ("CHPL_LLVM", CHPL_LLVM)
+                  ];
+    for i in 1..23{
```
lydia-duncan: You've hard-coded the size of `env_arr` here.  I recommend instead iterating over the contents of `env_arr`, like this:
```chapel
for env_var in env_arr { // env_var is a tuple, e.g. ("CHPL_LLVM", CHPL_LLVM)
  if (path_expanded.find("$"+env_var(1))> 0) {
...
```
or
```chapel
for (name, value) in env_arr {
  // (name, value) breaks down the contents of the array so you can access them cleanly
  // e.g. ("CHPL_LLVM", CHPL_LLVM)
  if (path_expanded.find("$"+name)> 0) {
    path_expanded = path_expanded.replace("$"+name, value);
...
```

modules/standard/Path.chpl
```diff
@@ -286,4 +286,55 @@ proc file.realPath(): string throws {
...
+                   ("CHPL_TIMERS", CHPL_TIMERS),
+                   ("CHPL_UNWIND", CHPL_UNWIND),
+                   ("CHPL_MEM", CHPL_MEM),
+                   ("CHPL_MAKE", CHPL_MAKE),
+                   ("CHPL_ATOMICS", CHPL_ATOMICS),
+                   ("CHPL_NETWORK_ATOMICS", CHPL_NETWORK_ATOMICS),
+                   ("CHPL_GMP", CHPL_GMP),
+                   ("CHPL_HWLOC", CHPL_HWLOC),
+                   ("CHPL_REGEXP", CHPL_REGEXP),
+                   //("CHPL_WIDE_POINTERS", CHPL_WIDE_POINTERS),
```
lydia-duncan: I think the intention of this function is to replace *any* environment variable, not just the ones Chapel knows about.  That will be slightly trickier.  I recommend using the Spawn module to get a subprocess; in that subprocess use known bash commands to get the list of environment variables and their values to create your array, and then follow the implementation you have outlined below

Unfortunately writing tests for that will be slightly more complicated - you'll need to use a .execenv file for your test to set dummy environment variables to values you know and compare their output explicitly.  An example of a .execenv file can be found at test/distributions/sungeun/dataPar2.execenv

lydia-duncan:
Thanks for taking this task on!  I think this is a promising start.  What other edge cases can you think of that would potentially lead to problems?

modules/standard/Path.chpl
verma-varsha: The commands to be used to get the list of the environment variables will be different on different platforms. For instance, the bash commands that I use might not work for a windows system. How do I tackle that?

lydia-duncan: Chapel itself doesn't work on pure Windows, only linux-esque setups (i.e. cygwin, bash for Windows), so I'm not as worried about that - we've handled those cases in the past by saying they aren't supported yet in the comment on the function itself.

However, multiple bash versions is a good point.  Is there a command you know of that would at least be common for most linux-based systems?

verma-varsha: I think printenv should work. I'll need to look up how to construct the array from the rendered output.

cassella: I think this approach has a few drawbacks:

1. The program has to create a new process for this, which is pretty high overhead for a string manipulation operation.
2. You have to parse the output of the command.  Note it's legal for an environment variable to contain newlines:

```
fortytwo@magrathea:~$ export FOO="asdf
fdsa=notavar
> 
> oops
> "
fortytwo@magrathea:~$ echo "$FOO"
asdf
fdsa=notavar

oops

fortytwo@magrathea:~$ env
...
TERM=xterm
FVWM_USERDIR=/home/fortytwo/.fvwm
FOO=asdf
fdsa=notavar

oops

HOSTDISPLAY=magrathea:0
```
I don't know how you would parse that reliably.  (It's not entirely an artificial concern -- I've seen a ``$TERMCAP`` environment variable with newlines in it, and lots of ``=``s.)

Another approach would be to use the ``Environment`` module to get an associative array of all the environment variables. 

Unfortunately, the ``Environment`` module is not yet implemented:  #5939 

An entirely different approach would be to turn the problem around.  Instead of getting a list of all environment variables and searching the string for each of them, go through the string and look for things that appear to be environment variables.  Then use ``getenv()`` to get the replacement text.

This has the drawback that you have to carefully process the input string.  And you have to decide what to do about input strings like ``"foo/${BAR${BA${DOS}}}"`` or ``"${OPEN_AND ooh squirrel"``.  (It's my understanding that the python ``os.path.expandvars()`` would be an appropriate model of behavior for such cases.  https://docs.python.org/2/library/os.path.html)

I think you can get at the C-level ``getenv()`` as is done in ``test/execflags/gbt/getenv.chpl``.  I don't know if that's appropriate for module code vs. for a test though.

lydia-duncan:
Heads up, I'll be out of the office until Monday starting shortly.  I'm sure others would be happy to answer questions that come up between now and then, either on the mailing lists or in the chatroom, otherwise I'll check back in on the status of this when I get back.

modules/standard/Path.chpl
mppf: I agree with @cassella and furthermore think calling getenv is appropriate in a Chapel module. For example, the MPI module already does this. It would be nice to have an actual Env module focused on having a reasonable API for such queries.

The Chapel runtime includes a sys_getenv call that is intended to be more portable than the bare getenv, and to limit possible problems with the global errno. This function is available from Chapel in the Sys module. It is already called in LocaleModelHelpSetup.chpl which is probably included by default (it could depend on the configuration).

lydia-duncan:
Checking in - how are you feeling about the direction and changes proposed?

verma-varsha:
Hi, Apologies for the late response. I am caught up a bit with my college examinations.
The approach that I am thinking of following is via using the `sys_getenv()` procedure as suggested by @mppf. First all, I will parse the string to search for any $ signs and the moment I encounter one, I will look for where the name of the env variable terminates so as to get the complete name to pass to `sys_getenv()` . The searching part will be quite easy for variables of the type `${ABCD_FGH}` and for ones of the type `$ABCD`, I will search for a `/` to demarcate the variable name. 
The part of the task that I am not sure about is how to handle the strings of the type `"foo/${BAR${BA${DOS}}}"` or `"${OPEN_AND ooh squirrel"` as pointed out by @cassella .

lydia-duncan:
> Hi, Apologies for the late response. I am caught up a bit with my college examinations.

No worries!  Just making sure you weren't giving up because we did something wrong.

verma-varsha:
@lydia-duncan I have modified the function. Please review the code.

modules/standard/Path.chpl
```diff
@@ -287,6 +288,85 @@ proc file.realPath(): string throws {
...
+                   `$<name>` or `${<name>}`.
+   :type path: `string`
+
+   :return: `path`, having replaced all references to environment variables with
+                their values
+   :rtype: `string`
+*/
+  proc expandVars(path: string): string{
+    var path_p:string = path;
+    var varChars:string = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-_";
```
cassella: I was all ready to comment that ``-`` isn't a legal character in an environment variable name.  I'm glad I checked the ``setenv`` man page before speaking up.  It turns out that any character is legal in an environment variable name, apart from ``=`` and the NULL character.  However, what to do about it is more nuanced.

bash at least doesn't accept anything but alphanumeric and ``_``.  But the ``setenv()`` libc call itself (and presumably ``getenv()``) do allow variables with other characters, as reported by ``/usr/bin/env``, and more immediately relevant, by python's ``os.path.expandvars()`` in some cases.

I'll spare you the experimentation by sharing the code I used to check.

First, a program to call ``setenv()``, and then exec another program to see what actually got set in the environment:

setenv.c:
```c
#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>

int main(int argc, char **argv) {
  char *e = argv[1];
  char *v = argv[2];
  char *b;
  int rc;

  rc = setenv(e, v, 1);
  if (rc != 0) {
    fprintf(stderr, "setenv('%s', '%s') failed: %s\n",
	    e, v, strerror(errno));
    exit(1);
  }

  //rc = execl("/bin/bash", "bash", (char*)NULL);
  //rc = execl("/usr/bin/env", "env", (char*)NULL);
  b = malloc(strlen(e) + 4);
  if (b == NULL) {
    fprintf(stderr, "malloc failed:\n");
    exit(1);
  }
  sprintf(b, "${%s}", e);
   
  rc = execl("./expandvars.py", "expandvars.py", b, (char*)NULL);

  fprintf(stderr, "execl failed: %d: %s\n", rc, strerror(errno));
  free(b);
  return rc;
}

```

And the expandvars.py script to see what ``os.path.expandvars()`` does:

```python
#!/usr/bin/env python

import os.path
import argparse


def main():
    parser = argparse.ArgumentParser(description='call expandvars')
    parser.add_argument('vn', type=str, help="var name")
    args = parser.parse_args()

    startstr = "/foo/" + args.vn + "/bar"
    newstr = os.path.expandvars(startstr)
    print("got {newstr}".format(newstr=newstr))
    return 0


if __name__ == "__main__":
    main()
```

Observations:
* With expandvars.py execing /usr/bin/env, ``./setenv "SNARF SNARF" "GLEEP"`` invokes env which does include the line ``SNARF SNARF=GLEEP``.
* With expandvars.py execing bash, and running ``env`` within bash
** ``./setenv SNARF GLEEP`` does result in ``SNARF=GLEEP``, showing that bash startup isn't obliterating the new variables.
** But ``./setenv "SNARF SNARF" GLEEP`` doesn't get set, revealing (I think) that bash suppresses environment variables it doesn't like.
* With expandvars.py execing expandpath.py,
** ``./setenv "SNARF SNARF" GLEEP`` shows that ``os.path.expandvars()`` _does_ expand variables with spaces in them.  (Similarly, hyphens and dollar signs.)
*** But only in the ``${SNARF SNARF}`` or ``${SNARF-SNARF}`` syntax.  It doesn't expand ``$SNARF SNARF`` or ``$SNARF-SNARF``.  (``setenv.c`` code to invoke this not shown, but trivial.)


Which I think is pretty much what you've implemented, except for the ``-`` in ``varChars``.  So this is just a very long way of saying that I think there's one extra character in the commit.  :smile:

lydia-duncan: I would try using a [regular expression](https://chapel-lang.org/docs/modules/standard/Regexp.html) here instead of listing all the characters out (note that using this support will require a `.skipif` file for your test when regex support is not built it, like [this file](https://github.com/chapel-lang/chapel/blob/master/test/release/examples/benchmarks/shootout/pidigits.skipif), but there aren't many situations where that support won't build)

modules/standard/Path.chpl
```diff
@@ -300,41 +301,205 @@ proc file.realPath(): string throws {
...
+        var c:string = path_p(ind);
+        if(c == "$"){
+            if(path_p(ind+1) == "$"){
+                res = res + c;
+                ind += 1;
+            }
+            else if(path_p(ind+1) == "{"){
+                path_p = path_p((ind+2)..);
+                pathlen = path_p.length;
+                ind = path_p.find("}");
```
lydia-duncan: This seems like it assumes the next `}` you find will close the current open one.  Am I wrong in thinking we could have a nested situation with `{ }`, for example `${FOO${BAR}}`?

lydia-duncan:
I'd like to see a test case for the nested version (regardless of the intended behavior for that case) and for an environment variable that wasn't defined in the path (I know your code attempts to handle the latter but I don't see a corresponding case in the test file).

modules/standard/Path.chpl
verma-varsha: This kind of a nesting in environment variables is not supported in bash. That is why, I excluded handling that case in the code. For reference, https://stackoverflow.com/questions/917260/can-var-parameter-expansion-expressions-be-nested-in-bash .

cassella:
Another interesting test would be a string that ends with a ``$``, like

``var basic_string = "A$";``

(And similarly ``"A${"``.)

It might also be worth having a test that handles a few other special cases, like ``"$ $$ $$$ $$$$ ${}"``.  I don't see a problem in your code for those, but a significant purpose of the test suite is to "lock in" correct behavior.  If someone tweaks or optimizes the code months or years down the road, those tests will ensure they didn't break any of these special cases.

modules/standard/Path.chpl
lydia-duncan: True, but we're not running the function itself in bash, just checking against the contents of the string provided to us.  In the post, it sounds like zsh does support that kind of nesting, so it wouldn't be unheard of for someone to try sending that kind of path to us.  What does Python do when run from bash over that kind of string?

verma-varsha: I checked what python returns for a nested env variable using the following code:
```
import os
os.environ["ridge"]="bar"
os.environ["foobar"]="wall"
path1 = "${foo${ridge}}"
path2 = os.path.expandvars(path1)
print path2
```
And the output that was returned was ```${foo${ridge}}```, which means that python leaves such variables unchanged.

lydia-duncan:
Once the new test modes have been verified, this is probably ready to go (after I run a full suite against it).  Unfortunately, feature freeze is today (sorry, I thought I warned everyone, looks like I missed this PR) so we won't be able to get this merged for a little while.  Not to worry, it'll just go into next release.

lydia-duncan:
Sorry about the merge conflicts - had some documentation related updates that needed to go in for the release.  I recommend waiting to resolve the conflicts until closer to when we'll merge the PR

verma-varsha:
@lydia-duncan I've added the new tests. Please have a look.

lydia-duncan:
The new tests look nice!  I'll validate it locally tomorrow and see how things go

lydia-duncan:
I checked out your branch and ran testing with a clean version of the most recent commit.  I got an assertion error on the first line and printing the output looks like the environment variables didn't get replaced

verma-varsha:
Hey @lydia-duncan  I just resolved the merge conflicts to check the testing status and looks like it's passing all the tests on the Travis CI.

lydia-duncan:
Travis CI doesn't do much beyond checking that very basic Chapel programs work.  We would like to add functionality that allows it to check tests adding in the PR itself, but haven't had the time for that yet.

I'll double check the updated branch

lydia-duncan:
Yeah, still getting the assertion failure.  The result of the expandVars call is "$CHPL_HOME/bin/${CHPL_HOST_PLATFORM}" while what it is being compared to looks appropriate (and the variables are definitely in my environment when I am running the test).  Can you run it successfully yourself?

lydia-duncan:
```
$ printchplenv --anonymize --all
CHPL_HOST_PLATFORM: linux64 *
CHPL_HOST_COMPILER: gnu
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_LLVM: none *
CHPL_AUX_FILESYS: none
```

verma-varsha:
For some reason on my machine, sys_getenv is returning an error code of `1` along with a successful return of the environment variables value. So to account for that, when I check for `h!=1` instead of `h!=0`, it's working fine. So I thought there might be some problem with my environment settings and believed in whatever results the Travis CI gave me.

lydia-duncan:
Ah, so it's assuming it gets an error code back and won't update the environment variable otherwise.  You should be able to check the exact error you are getting from the call, possibly by accessing some C level stuff - some of the FileSystem functions rely on internal runtime wrappers that understand the error codes and return a translated version back, you could base this checking off of their implementation

cassella:
The Chapel ``sys_getenv()`` in ``runtime/src/qio/sys.c`` does return 1 on success.

modules/standard/Path.chpl
```diff
@@ -457,4 +458,182 @@ proc file.realPath(): string throws {
...
+ /*
+ Returns the parent directory of the :type:`~IO.file` record.  For instance,
+ a file with path `/foo/bar/baz` would return `/foo/bar`
+
+ Will halt with an error message if one is detected.
+
+  :return: The parent directory of the file
+  :rtype: `string`
+ */
+ proc file.getParentName(): string throws {
```
cassella: Something's gone awry with your merge.  There are now two of each of the ``proc file.getParentName()`` definitions, and two ``joinPath()``s and two ``isAbsPath()``s.  :(

verma-varsha: Thanks for pointing that out. I resolved it!

verma-varsha:
After merging my branch upstream, the compiler is giving me this error.
`$CHPL_HOME/modules/internal/ChapelBase.chpl:1054: error: unknown pragma: "unsafe"`
This is a file that I haven't even touched. Any idea on why is that coming up?

cassella:
A recent commit added support for that pragma to the compiler, and added uses of the pragma to some of the modules.  Rebuilding the chapel compiler should take care of it -- ``make`` at the top-level.

verma-varsha:
Wooaah! Thank you so much. Had been struggling with that error for quite some time!

verma-varsha:
@lydia-duncan Now I think it should work fine.

cassella:
Hmm.  When I run it now, I get an error from the ``x2`` test.  I think it's because ``CHPL_COMM_SUBSTRATE`` isn't an environment variable, only a chpl variable?  It's printed by ``printchplenv``, but not by the regular shell ``env``.  So it may depend on each particular user's configuration.

verma-varsha:
Yeah. That happened with me too. I had to set `CHPL_COMM_SUBSTRATE` explicitly to pass that assert.

mppf:
> Yeah. That happened with me too. I had to set CHPL_COMM_SUBSTRATE explicitly to pass that assert.

It seems to me this test should use only environment variables set in the .execenv for the test. Otherwise it might give different output in different configurations. I don't see why it needs to check things like CHPL_COMM_SUBSTRATE.

lydia-duncan:
Your new test now passes for me as well!  Running a full test suite now

lydia-duncan:
Standard testing passed as well, I think this is ready to go.   What would you like to be used for the merge message, following the guidance from [here](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/ContributorInfo.rst#final-merge-message)?

PR #8650, opened 2018-02-28
Add special case to codegen for casting wide pointers to c_void_ptr

### LINKED ISSUE #8512: Need compiler error or implicit conversion for casting wide pointers to c_void_ptr
When a wide pointer is cast to a `c_void_ptr` it produces an error in CodeGen that resembles this...

```
error: cannot convert to a pointer type
 fn_chpl(((c_void_ptr)(tmp_chpl6)));
 ^~~~~~~
```

This needs to either raise a compiler error or have it take the address portion and cast that to `c_void_ptr`. Note that this isn't really expected behavior and is very confusing for the user when the code that triggers such an error does not even make use of any locality at all. For this reason, I would think that if the user does cast it to a `c_void_ptr` and if the data is actually remote, it is the users fault (hence it should not result in a compiler error).

**Minimum Working Example:**

```chpl
proc fn(f : c_void_ptr) {
	writeln(f);
}

class Obj { var x : int; var y : int; var z : int;}
var obj : Obj = new Obj(1,2,3);
fn(obj : c_void_ptr);
```

[TIO](https://tio.run/##TY1BDsIgFETXcoq/bJOyUHclnsEjNPgLivkBAqTVNr26SKuL7uZl8mbwIb2inH1wCNpWGlrAbnCm73wKNczsMAaTFJWqFmxhDEnGCNfbE2YYZIBXMYxNYoP3HqY/LGwlV4x28y5g1bim6ticmnOZLce/enctcv6gJnmPmXPrODmU9AU)

**Proposed Change:**

This is a decent test I believe which should double as a good example for why I think it should handle the conversion implicitly. This will show two codegen errors, one for each cast of `a` and `b`. 

```chpl
 extern proc memmove(dest: c_void_ptr, const src: c_void_ptr, n: size_t);
var a = new object();
var b = new object();
memmove(a : c_void_ptr, b : c_void_ptr, c_sizeof(object));
```
[TIO](https://tio.run/##ZY1BCsIwEEX3nmKWDZgLtHiWMJlOtZJkQjJE6eGNiLqoLv@D/x5dMHPone/KJUEuQhA5Rmk8zFx1BHJN1tllLUcgSVWhFtrjNEJdN3ZqpkPDAggnSHwD8VcmHT7U/9FvCGHv8z@b3Esvy/C@GjP1/qAl4Ll2a5PYIIThCQ)

LouisJenkinsCS:
Fixes issue discussed in #8512 

Closes #8512 .


compiler/codegen/expr.cpp
```diff
@@ -4770,10 +4773,15 @@ GenRet CallExpr::codegenPrimitive() {
       } else if (isRecord(typeInfo()) || isUnion(typeInfo())) {
         INT_FATAL("TODO - don't like type-punning record/union");
 
+      } else if (src->symbol->hasEitherFlag(FLAG_WIDE_CLASS, FLAG_WIDE_REF) && typeInfo() == dtCVoidPtr) {
```
mppf: I suspect this should only check for FLAG_WIDE_CLASS

compiler/codegen/expr.cpp
```diff
@@ -4770,10 +4773,15 @@ GenRet CallExpr::codegenPrimitive() {
...
+        // Special case: If we are casting a wide-ptr to a c_void_ptr we need to ensure
+        // that we perform the cast on the actual addression portion of the wide-ptr.
```
mppf: > addression
is that a typeo?

mppf:
I don't think it should include a local check, because c_void_ptr "obviously" doesn't include a locale, and I see no reason why you wouldn't want to be able to create a remote C pointer from a wide Chapel pointer (i.e. maybe you'll turn it into a GET of some sort, or store the locale id in a different struct element, or ...)

test/expressions/AtomicObject/voidPtrTest.chpl
```diff
@@ -0,0 +1,9 @@
+class Obj { var x : int; var y : int; var z : int;}
+
+proc fn(f : c_void_ptr) {
+	writeln(f : Obj);
+}
+
+
+var obj : Obj = new Obj(1,2,3);
+fn(obj : c_void_ptr);
```
mppf: missing newline at end of file

compiler/codegen/expr.cpp
```diff
@@ -4743,6 +4745,7 @@ GenRet CallExpr::codegenPrimitive() {
       // should deref the 'src'.
       if (src->symbol->hasFlag(FLAG_REF) &&
           dst->symbol->hasFlag(FLAG_REF) == false) {
+        // (*src)
```
mppf: I'm unsure of the value of this and the above comment changes. Usually when I add such comments they reflect my increasing understanding, but I'm not sure how helpful they'd be to others. I recommend you either remove them or turn them into more words.

test/expressions/AtomicObject/voidPtrTest.chpl
LouisJenkinsCS: Done

mppf:
bcdf00c8ca5503c4f65dfb4e50d1880e4dd56718 passed full local testing. I'm gonna run --no-local testing too.

mppf:
I did, and do.

mppf:
Passed full --no-local testing.

mppf:
TODO before merging:
- [ ] LLVM backend --no-local testing with the new test

mppf:
@LouisJenkinsCS - it would help if you can do the --llvm --no-local testing on your new test, thanks. Once that is done, this can merge.

PR #8659, opened 2018-03-01
Add comm=ugni perfstats to measure dynamic alloc related activities.

### LINKED ISSUE #8625: Investigate overheads of dynamic array registration
https://github.com/chapel-lang/chapel/pull/6947 added dynamic array registration support for ugni in order to improve numa affinity for arrays. We've had some outstanding questions about the overheads and non-numa-affinity performance implications of this work, particularly with regard to ISx.

We thought the broadcast critical section might be slow and that it could have a significant serializing impact on applications. We took a look reducing the critical section in https://github.com/chapel-lang/chapel/pull/8177, but we didn't see any large performance wins with that. Prior to https://github.com/chapel-lang/chapel/issues/8497 we also didn't have a good platform to investigate because ISx was so noisy. 

This issue is meant to serve as a placeholder to investigate as well as a place to store the results for future reference.

gbtitus:
This adds some more internal performance statistics, this time having to
do with dynamic allocation.  I also did some cleaning up in perfstats,
reducing the number of times we repeat the counter type and things like
that.  All of this is normally turned off; we only use it for low-level
performance investigations because it does slow down normal operations
noticeably.

Measurements produced by these changes led to the conclusions which
allowed us to close #8625.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -267,6 +267,8 @@ static uint64_t debug_stats_flag = 0;
         MACRO(regMem_lock_nsecs)                                        \
         MACRO(regMem_critsec_nsecs)                                     \
         MACRO(regMem_alloc_nsecs)                                       \
+        MACRO(regMem_kpage_nsecs)                                       \
+        MACRO(regMem_defaultInit_nsecs)                                 \
         MACRO(regMem_reg_nsecs)                                         \
         MACRO(regMem_dereg_nsecs)                                       \
         MACRO(regMem_free_nsecs)                                        \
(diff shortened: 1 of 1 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -315,28 +331,52 @@ chpl_comm_pstats_t chpl_comm_pstats;
...
+#define PERFSTATS_TSTAMP(ts)    _PSV_C_TYPE ts = perfstats_timer_get()
+#define PERFSTATS_TDIFF(te, ts) ((te) - (ts) - timer_overhead)
+#define PERFSTATS_TELAPSED(ts)  PERFSTATS_TDIFF(perfstats_timer_get(), ts)
 
 static void perfstats_init(void)
 {
   (void) clock_gettime(CLOCK_MONOTONIC, &perfstats_timeBase);
+
+  //
+  // Measure the timer overhead.  We only use timer_nsecs so that we
```
ronawho: If you're worried about timer overhead you could consider using `CLOCK_MONOTONIC_COARSE` instead (requires 2.6.32, but that shouldn't be an issue for ugni)

ronawho: Although it's less accurate, so maybe that's not what you want anyways

gbtitus: Yeah, these are fairly fine-grained measurements in some cases.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2975,6 +3048,17 @@ void chpl_comm_impl_regMemPostAlloc(void* p, size_t size)
   mem_region_t* mr;
   int mr_i;
 
+#if 0
```
ronawho: I don't love the `if 0` uses, I would prefer we just use a new define to trigger this. i.e. `#if PERFSTATS_COMM_UGNI_ZERO_ARRAY_ALLOC` or something

PR #8660, opened 2018-03-01
Allow fields to be used in phase 1 after they are initialized

benharsh:
This PR allows initializer authors to use fields in phase 1 as long as they have been initialized (implicitly or explicitly). This is significantly more flexible than our current approach.

Since this functionality is largely about whether errors are emitted, this PR focuses on the implementation of error checking for initializers.

### Use after initialization

Previously we had a few helper functions that told us whether a method was used, or a field was used before it was initialized, etc. These helper functions all need to look for expressions like ``'.' this <something>``, so this PR creates a new helper function to just look for this-dot expressions. This new helper allows us to collect potentially-problematic expressions and emit multiple errors instead of exiting upon the first error.

The "collect then check" refactor stops the unhelpful "this used in actual" errors when passing fields to functions. By updating some key places in ``initializerRules.cpp`` to use this new error checking, fields can now be used in phase1 as long as they have already been initialized.

### Out of order initialization

A potentially confusing side effect of this work was that users could write code that looked like fields being initialized out of order, which should be an error. For example:
```chpl
this.second = 1;
this.first  = 2;
```
Would really be transformed by normalization into:
```chpl
this.first  = 0;
this.second = 1;
this.first  = 2; // actually assignment!!
```
 This PR adds functionality to check either a field was implicitly initialized. This enables us to detect expressions that look like out-of-order initialization and emit errors.

### Testing
- [x] full local w/ futures


compiler/passes/initializerRules.cpp
```diff
@@ -373,6 +373,51 @@ static void preNormalizeInitClass(FnSymbol* fn) {
   }
 }
 
+static void checkLocalPhaseOneErrors(InitNormalize state,
```
noakesmichael: I think this *could* be "const InitNormalize& state" but I am not worried about it.

compiler/include/InitNormalize.h
```diff
@@ -165,6 +158,8 @@ class InitNormalize {
   InitPhase       mPhase;
   BlockType       mBlockType;
   BlockType       mPrevBlockType;
+
+  std::set<DefExpr*> implicitFields;
```
noakesmichael: Not the end of the world, and we don't have a formal convention, but I note that the
other member fields of this particular class use the "m" prefix (for "member") to help
distinguish them from procedure locals within methods.

test/classes/initializers/generics/phase1/hasLoopBad3.chpl
```diff
@@ -1,12 +1,16 @@
-// Alternate version of hasLoopBad2.chpl, where the incorrect loop actions
-// occur after we've already seen the initialization of all fields.
+// Alternate version of hasLoopBad2.chpl, where the loop occurs after all
```
noakesmichael: I am not concerned at all but this is maybe the kind of hybrid case we might want to
be sure we're all OK about.

Arguably this initializer isn't valid under the old definition or the new definition.
1) I recognize that we'll have some uncomfortable hybrids during the transition
2) One could claim that the new rules apply retroactively to phase 1 of "old style" initializers.

In either case this just helps us keep the pressure on to deprecate old-style initializers

test/classes/initializers/errors/multipleErrors.good
```diff
@@ -0,0 +1,5 @@
+multipleErrors.chpl:10: In initializer:
+multipleErrors.chpl:11: error: Field "x" used before it is initialized
+multipleErrors.chpl:11: error: Field "y" used before it is initialized
+multipleErrors.chpl:11: error: cannot call a method during phase 1 of initialization
+multipleErrors.chpl:11: error: cannot pass "this" to a function in phase 1 of initialization
```
cassella: This is surprising to me.  I thought the proposal (at least in the chip) allows the object in phase 1 to be an instance of the parent class, and allows parent class method calls.  Shouldn't it also be passable-to-functions as an instance of the parent class? (``object`` in this case?)

test/classes/initializers/phase1/useAfterInit/passThisFn.good
```diff
@@ -0,0 +1,2 @@
+passThisFn.chpl:10: In initializer:
+passThisFn.chpl:15: error: cannot call a method during phase 1 of initialization
```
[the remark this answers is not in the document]
benharsh: Hmmm, fair point. I'll add a future for this case because I suspect we can't catch this reliably yet. For example, it's possible to define a method only on a particular instantiation of a type.

test/classes/initializers/errors/multipleErrors.good
benharsh: This error currently will apply to the old proposal. Eventually it will be updated to mention 'phase 0' or something about the region before super.init/this.init.

cassella: I still feel confused.  This initializer includes an ``initDone()`` call, so it can't be the old proposal?  Unless there's a new new proposal already.

And chip 23, under "Simple classes" says if a ``super.init()`` is omitted, the compiler will insert one at the beginning of ``init()``.  So this ``printer()`` call will be in phase 1 in the new proposal.

benharsh: It's an artifact of the hybrid world on master as we transition towards the new initializer proposal. As you point out it's definitely a new initializer, but we don't currently have the 'behave like your parent' functionality implemented so I left the error message in (for now).

test/classes/initializers/phase1/useAfterInit/passThisFn.good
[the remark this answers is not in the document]
cassella: You mean like

```chapel
record R {
  type t;
  proc foo() where t == int { writeln("Hello from int foo"); }
}
```

and this error is generated before the type is instantiated enough to know if the method resolves?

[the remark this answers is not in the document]
benharsh: Yes, something like the example you wrote is problematic for the current compiler architecture. Internally we have briefly discussed the likely need to move some error checking until after types are resolved, but that probably won't happen until after the 1.17 release.

PR #8661, opened 2018-03-01
Move code to determine the virtualenv dir into chpl_home_utils

ronawho:
Instead of hard-coding it in activate_chpl_test_venv.py, move it into a common
location so that we can use it in other scripts. This is motivated by wanting
to activate the venv in a script that will do start_test tab-completion.


ronawho:
Tested with default venv, `CHPL_TEST_VENV_DIR=none` and `CHPL_TEST_VENV_DIR=<some manually created venv dir>`

ronawho:
Probably easiest to review if you ignore whitespace and look at it commit at a time

PR #8662, opened 2018-03-01
Fix c2chapel: change source to . for ubuntu systems

npadmana:
On Ubuntu systems, /bin/sh points to dash, which does not have `source` (see https://wiki.ubuntu.com/DashAsBinSh). This pull request simply switches `source` to `.`.


npadmana:
@benharsh, @ben-albrecht  -- minor bugfix to c2chapel.

ben-albrecht:
Thanks @npadmana !

PR #8665, opened 2018-03-01
Stop hard-coding path to virtualenv dir in ./run-in-venv

ronawho:
Use the new common helper routines added in #8661

### DIFF
--- a/util/chplenv/chpl_home_utils.py
+++ b/util/chplenv/chpl_home_utils.py
@@ -1,4 +1,6 @@
 """ Access to home variables """
+
+import optparse
 import os
 import sys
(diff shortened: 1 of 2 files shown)


ronawho:
Verified that `make docs` and `chplspell` still work. Building the chips fails, but it fails for the same reasons on master (warnings treated as errors for a nonlocal image and duplicate labels in chip10/23)

ben-albrecht:
> but it fails for the same reasons on master (warnings treated as errors for a nonlocal image and duplicate labels in chip10/23)

Note to self: Add this to Travis check (#6758 )

PR #8666, opened 2018-03-01
Add start_test tab completion support

ronawho:
Use argcomplete to provide tab completion for start_test. argcomplete hooks
into python scripts and dynamically executes them to determine a list a
completion hooks. This adds a little overhead to completions because the python
interpreter has to be spun up, but it avoids having to manually create/maintain
completions files like we do for the compiler.

To enable tab completion you should just be able to source:
$CHPL_HOME/util/devel/chpl-scripts-completion.bash

Related to https://github.com/chapel-lang/chapel/issues/5348

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# PYTHON_ARGCOMPLETE_OK
 #
 # CHAPEL TESTING
 
@@ -27,6 +28,10 @@ from chplenv import *
 
 from test import activate_chpl_test_venv
 import argparse
+try:
+    import argcomplete
+except ImportError:
+    argcomplete = None
 import subprocess32 as subprocess
 
 # PROGRAM ENTRY POINT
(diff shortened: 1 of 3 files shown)


util/start_test
```diff
@@ -27,6 +28,10 @@ from chplenv import *
 
 from test import activate_chpl_test_venv
 import argparse
+try:
+    import argcomplete
```
ben-albrecht: Pretty slick, but why do we need this? For users with outdate virtual envs?

ronawho: Out of date virtualenv, or if they don't source chpl-scripts-completion. Hmm. actually so long as argcomplete is installed it shouldn't matter if chpl-scripts-completion was sourced (from a start_test working point of view that is, obviously tab completion wouldn't work.)

So yeah, just outdated virtualenv (and it's a core script, so I don't mind the extra checks for something that's not strictly needed for usability)

PR #8671, opened 2018-03-01
Stop hard-coding path to virtualenv dir in the man Makefile

ronawho:
Use the new common helper routines.


ronawho:
Verified that the man pages still build on darwin and linux64

PR #8672, opened 2018-03-01
Stop hard-coding path to virtualenv dir in the chpldoc code

ronawho:
Use the new common helper routines


ronawho:
I've done some basic chpldoc testing, but I still need to do a full paratest

Issue #8377, opened 2018-02-03
lifetime checking: specifying function return lifetime


mppf:
I know of the following interesting patterns:
* lifetime is based upon 'this'
* lifetime is infinite
* lifetime is based upon a minimum of the lifetimes of (certain?) actual arguments

What syntax can be used to specify these, in case the default behavior is not sufficient?

Here are a few examples where we'd like to be able to specify the lifetime of the return:

#### Function returning an element of a global hashtable
``` chapel
proc getGlobalHashtableElement( const ref key ) const ref {
  return globalHashTable[key];
}
```

Here, lifetime inference is likely to assume that the function could return the `key` argument. But, once `key` is used to find the hashtable element, the `key` is no longer used.

#### Method returning an array element
``` chapel
proc MyType.getElement( const ref key ) const ref {
  return this.array[key];
}
```

#### Function returning an element of an outer local hashtable
``` chapel
proc outerFunction() {

  var D:domain(string);
  var hashtable:[D] MyClass;

  proc innerFunction( key )  {
    return hashtable[key];
  }

  var element:MyClass = nil;
  {
    var key = "test";
     element = innerFunction( key );
  }
}
```
The annotation for `innerFunction` might need to indicate lifetime matches an outer variable rather than an argument.

psahabu:
Taking another peek at Rust, [they seem to have some pretty decent lifetime elision rules.](https://doc.rust-lang.org/book/second-edition/ch10-03-lifetime-syntax.html#lifetime-elision)

We might differ by adding a fourth rule, having the default be the minimum of the provided lifetimes.

mppf:
I think "infinite" is somewhat the same as saying "it's a raw pointer" so perhaps the way we'd say that is by using a "raw" pointer type (whatever that is).

psahabu:
We could call it `Unsafe(T)`.

vasslitvinov:
The case "copyObjectTree" returns an Owned thing. My understanding is that we need to concern ourselves with the lifetime only of borrowed things. (Correct?)
```chpl
proc copyObjectTree(...) : owned MyTree {...}
```

To indicate the lifetime, I propose the syntax `borrowed(variable or field reference(s))`. Which indicates the shortest of those variable(s)/field(s)' lifetimes. Gives us an excuse to introduce the `borrowed` keyword. :) For example:

```chpl
proc getGlobalHashtableElement(key) borrowed(globalHashTable) const ref {...}

[...]

Issue #8373, opened 2018-02-03
Lifetime checking meta-issue


mppf:
Lifetime checking allows for compile-time detection of many use-after-free errors. I have a prototype adding lifetime checking to the compiler in PR #8338.

This issue links to other issues that represent language design questions related to the effort.

mppf:
I'm closing this issue because it has served its purpose and almost everything is done. The remaining open questions are covered in these issues:

 * #11926
 * #9274
 * #10879
 * #8382

PR #8673, opened 2018-03-01
Add lifetime checking

mppf:
This PR adds a lifetime checking component to the compiler which is off by default. The checker can be activated with the --lifetime-checking compiler flag or by using `pragma "safe"` on a function or module in which checking is desired.

There are several language design questions that need to be answered before this checker can be enabled by default. These design questions are in many issues but are linked to from #8373. The primary open question relevant to this pass specifically is about how to override the default lifetime inference for functions and methods - issue #8377. This PR uses pragmas for that purpose. Most of the other open issues are about the evolution of `Owned` more generally.

Lifetime checking allows for compile-time detection of many use-after-free errors. I have a prototype adding lifetime checking to the compiler in PR #8338. The lifetime checker is concerned with finding potential use-after-free errors for references and borrows.

The lifetime checker infers the lifetime of each symbol (how long what it refers to will be allocated). Globals have infinite lifetime, but a local simple value variable (e.g. an integer) has lifetime of the function block it is declared in. The lifetime of a symbol storing the result of a call is the minimum of the actual lifetimes relevant to that call.

The checker emits errors for several cases:

* reachability of a symbol is greater than its lifetime
* setting a symbol to another value with shorter lifetime
* returning a symbol with non-infinite and non-return lifetime (e.g. reference arguments are marked with return lifetime)

The checker looks for misuses of `ref`/`const ref` in addition to borrows from Owned etc. The checker combines lifetimes for refs and lifetimes for borrows into a single ScopeLifetime, but these can be conceptually be viewed as totally unrelated - there could be 2 separate checking passes, one for refs, and one for borrows.

- [x] full local testing


mppf:
There are currently two failures with --lifetime-checking:
[Error matching compiler output for optimizations/bulkcomm/asenjo/stencilDR/v2/stencil]
[Error matching compiler output for users/ferguson/histogram/reductionex_record_class]


These should not need to stop this PR from merging.

compiler/resolution/lifetime.cpp
```diff
@@ -0,0 +1,1417 @@
...
+   lifetimes of actual arguments with return scope.
+
+   Records are subject to lifetime checking if they contain
+   fields that are subject to lifetime checking.
+
+   Assumption: = operator overloads for records or classes
+   handle lifetimes as if by-value. But note, ownership
+   operations with Shared or Owned records count as "by value"
+   here.
+
```
vasslitvinov: I suggest adding brief explanations of how this analysis handles aggregates: records, classes, arrays, an array's domain and distribution. Sounds like it happens implicitly by appropriate annotations on certain successor functions?

test/classes/delete-free/lifetimes/bad-return-borrow-owned.chpl
```diff
@@ -0,0 +1,77 @@
...
+module l10 {
+
+class MyClass {
+  var x:int;
+}
+record Rinit {
+  pragma "owned"
+  var c:MyClass;
+  proc init(c:MyClass) {
+    this.c = c;
```
vasslitvinov: Hm how is this supposed to work? My understanding is that 'c' is not supposed to escape because it is not "return scope"; however it does because it is stashed in 'this.c' and 'this' escapes. ???

mppf: yes it magically turned into an "owned" thing; once we have owned/borrowed/raw types this example can be more reasonable & the checker can be more complete.

compiler/resolution/lifetime.cpp
```diff
@@ -0,0 +1,1417 @@
...
+  // Mark all arguments with FLAG_SCOPE or FLAG_RETURN_SCOPE.
+  // Default for methods is to mark 'this' argument with return scope.
+  // Default for everything else is to mark all arguments with return scope.
+  forv_Vec(FnSymbol, fn, gFnSymbols) {
+
+    bool anyReturnScope = false;
+    // Figure out if any arguments have FLAG_RETURN_SCOPE on them.
+    // If not, we default to putting it on all of them.
+    // If so, we put FLAG_SCOPE on the rest.
```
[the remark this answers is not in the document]
mppf: it is technically unnecessary but I wanted to know all the formals were marked instead of relying on FLAG_RETURN_SCOPE missing = FLAG_SCOPE.

PR #8684, opened 2018-03-02
Add support for multiple registries to Mason

daviditen:
Update Mason to support multiple registries.  The environment variable
`MASON_REGISTRY` is used to list the available registries. It is a
comma-separated list of "name|location" pairs where "name" is what
the registry should be named locally and location is either a local
path or a URL pointing to a registry.  If unset, it defaults to
"mason-registry|https://github.com/chapel-lang/mason-registry", which
matches the behavior before this change.

The "name|" part of the of the "name|location" pairs is optional and
if not specified will be inferred to be the text following the final slash in
the "location" part.

If `MASON_REGISTRY` includes multiple registries, they are searched in
left-to-right order for any dependencies.

Removed support for the environment variable
`MASON_CACHED_REGISTRY` which is unnecessary given the ability to
name the local registry in `MASON_REGISTRY`.

Updated mason tests to quit using `MASON_CACHED_REGISTRY` and
specify appropriate values for `MASON_REGISTRY` instead.


tools/mason/MasonEnv.chpl
```diff
@@ -71,3 +142,13 @@ proc masonEnv(args) {
...
 }
+
+private proc getRegNameFromLoc(location: string): string {
+  var strippedLoc  = location.strip("/", leading=false);
+  var lastSlashPos = strippedLoc.rfind("/");
+  if lastSlashPos == 0 {
+    stderr.writeln("location should be an absolute path or URL");
+    exit(1);
+  }
+  return strippedLoc[lastSlashPos+1..];
```
ben-albrecht: Could this include `.git` for URLs?

daviditen:
> Are we choosing registry as the default name instead of the last string mason-registry in order to match the current behavior, i.e. for backwards-compatibility? If so, I wonder how painful it would be to switch that directory name to mason-registry to stay consistent with the rule of "choose the string after the last / by default."

I started with it using `registry` for backward compatibility, but switched to `mason-registry` when I added the "choose the string after the last / by default" rule.  I missed updating the message.  The message is fixed now.

daviditen:
Thanks @ben-albrecht!  I just pushed a commit to address your other comments.

PR #8690, opened 2018-03-02
Fix locale model init phase violations

dmk42:
The locale models were updated to use initializers while the rules were new and in flux.  #8660 started enforcing some of the rules (in particular, not passing "this" to something during phase 1), so the locale models need to be updated to correct that problem.  While doing this, I noticed other violations of the rules that are not yet being enforced.  Currently, const members should not be initialized in phase 2, but they are.

This change gets testing back on track by fixing the enforced violations, and fixes violations anticipated to be enforced in the future.

In some places, a const field had to be changed to a var field to avoid a full refactoring of locale models.  In the long run, all locale models will need to be refactored to take new init rules into account.

- [x] Passed hellos on the knl locale model
- [x] Passed release/examples on the numa locale model
- [x] Passed full local testing on the flat locale model

### DIFF
--- a/modules/internal/localeModels/apu/LocaleModel.chpl
+++ b/modules/internal/localeModels/apu/LocaleModel.chpl
@@ -136,9 +136,9 @@ module LocaleModel {
   //
   class LocaleModel : AbstractLocaleModel {
     const _node_id : int;
-    const local_name : string;
+    var local_name : string; // should never be modified after first assignment
 
-    const numSublocales: int;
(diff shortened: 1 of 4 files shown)


dmk42:
@benharsh - Would you please review from the point of view of #8660 ?

@mlchu - Would you mind doing a post-commit review of the APU locale model change?

dmk42:
@gbtitus - Just FYI, wanted to make sure you saw this.  No need to do anything.

PR #8696, opened 2018-03-03
Port Chapel to LLVM 6

dmk42:
This change ports Chapel to LLVM 6.  Some of the LLVM API has changed.
While there, ARM support is improved as well.
LLVM 6 is not made the default in this change.  This only allows people to use LLVM 6 themselves.  A forthcoming PR will make version 6 the default version of LLVM.

Passes full local testing.
Passes more of gasnet-fast than previous versions of LLVM.
Passes as much on Cray XC as previous versions of LLVM.

This is the first version of LLVM that fully supports the processors on the Cray XC50-AC ARM systems.  Work remains to be done there.


dmk42:

compiler/include/misc.h
```diff
@@ -23,7 +23,12 @@
 #include <cstdio>
 #include <cstdlib>
 
+#ifdef HAVE_LLVM
+#define exit(x) clean_exit(x)
+#else
+// This interferes with uses of exit() in LLVM header files.
 #define exit(x) dont_use_exit_use_clean_exit_instead
+#endif
```
mppf: I think this is OK for now but note that there is a better solution that amounts to extending the work done in PR #7233 - that is, extend util/devel/lookForBalloc to also check for bad exits.

compiler/util/clangUtil.cpp
```diff
@@ -1571,7 +1577,15 @@ void runClang(const char* just_parse_filename) {
...
-    march += CHPL_TARGET_BACKEND_ARCH;
+    const char *backend_arch = CHPL_TARGET_BACKEND_ARCH;
+    if (strncmp(backend_arch, "arm-", 4) == 0) {
+      backend_arch += 4;
+      march = "-mcpu=";
+    }
+    march += backend_arch;
+    if (strcmp(backend_arch, "thunderx2") == 0) {
+      march += "t99";
+    }
```
mppf: Shouldn't this happen in util/chplenv/chpl_arch.py ? If you change it much from what you've tested here, let's put it in a different PR.

compiler/include/misc.h
dmk42: Oh, nice catch.  I didn't realize we were doing that.  It looks like the current place is util/devel/look_for_calls.py, which is already looking for exit!

What I'll do is commit this as is, but post-commit:

@ronawho - Can we remove this entirely now that we sanitize the code of calls to exit?
```c
#define exit(x) dont_use_exit_use_clean_exit_instead
```

compiler/util/clangUtil.cpp
dmk42: OK, I'll leave this here for now, but we can think about whether this should migrate to chpl_arch.py in the future.  If we migrate it, it gets more complicated.

The answer we would want from chpl_arch.py depends on whether we are compiling with the `--llvm` flag.  If so, we need to translate from Cray module syntax to Clang syntax, but if not, we need to keep the Cray module syntax so the PrgEnv compiler can find the architecture properly.  If we were to put it in chpl_arch.py, we would have to default to the PrgEnv-style answer, and have the Clang driver run the Python program with a special command line argument to say that we want the Clang-style answer, and then wait for the response, or something like that.

The complexity of migrating it to chpl_arch.py might decrease if at some time in the future, compiling with LLVM means we always use LLVM and never have to worry about whether or not we have used the `--llvm` flag.

mppf: chpl_arch.py / chpl_compiler.py already works that way. Search for CHPL_LLVM_CODEGEN in the util/chplenv scripts to see more. That environment variable is set when `--llvm` is used.

dmk42: OK, thanks.  Sorry I missed that before.  I looked for several things, including chpl_arch.py, but I didn't think to look for printchplenv being called from the driver.

Now that I see this, I agree that in the long run, this should be migrated to chpl_arch.py.  That will require a rework of chpl_arch.py to handle the difference between x86-64, where `-march=` is needed, and ARM, where `mcpu=` is needed instead.  Therefore, migrating it to chpl_arch.py does belong in a separate PR.

compiler/include/misc.h
ronawho: We only check the runtime for calls to exit. See [util/devel/lookForBadRTCalls](https://github.com/chapel-lang/chapel/blob/master/util/devel/lookForBadRTCalls) or the PR Michael pointed out. You're welcome to beef up the checks to include the compiler too, but you'll either want to rename the script or create a new `checkForBadCompCalls` or something

PR #8705, opened 2018-03-05
Phase one by default for record initializers

benharsh:
This PR enforces phase one by default for records, meaning that we are now in the new initializer proposal outlined in CHIP 23. Note that classes still abide by the old proposal (unless an initDone is present).

Many tests were updated to replace super.init() in a record initializer with this.initDone(). Some tests that assumed phase two was the default required an initDone to be inserted at the top of the initializer.

In the compiler there are now a number of additional branches in which we do different transformations/checks if we're in the new proposal. These branches will collapse soon on master once classes are transitioned to the new proposal.

Some notable compiler changes:
- Added ``InitNormalize::completePhase0`` for new super.init rules
- extracted some error checking into ``checkInvalidInit``

Testing:
- [x] full local + futures
- [x] full gasnet


benharsh:
A future PR will enforce the rule that record initializers may not contain ``super.init``. That change was left out of this PR to reduce the diff size.

Issue #8555, opened 2018-02-22
Internal error for fully qualified `new` expressions with type specifiers


BryantLam:
### Summary of Problem
Internal compiler error.
- Works if `record` is a `class`.
- Works if module prefix is omitted.

### Steps to Reproduce

**Source Code:**
```chpl
module A {
  record R {
    var a: int;

    proc init(a: int) {
      this.a = a + 1;
    }
  }
}

use A;
var r: A.R = new A.R(1); // Doesn't work.
/* var r: R = new R(1); // Works. */
writeln(r); // {a = 2}
```

### Configuration Information

`chpl version 1.17.0 pre-release (2f0af5c5db)`

bradcray:
Thanks for filling this.  @lydia-duncan, @noakesmichael, @benharsh, does this failure mode look familiar to you?

I'm noting that the following variant works as well suggesting that it's specific to the `new A.R...` expression:

```chapel
var r: A.R = new R(1);
```

cassella:
I found a few more variants that work, and one that results in a different error.

For conciseness, combining the ``use``, ``new``, and ``writeln``
statements onto one line, and then variably removing parts of it,

These fail as the original does (``internal error: assertion error [normalize.cpp:826]``):

```chapel
use A;  var r : A.R = new A.R(1);
use A;  var r : A.R = new A.R(1); writeln(r);
use A;  var r :   R = new A.R(1);
use A;  var r :   R = new A.R(1); writeln(r);
        var r : A.R = new A.R(1);
        var r : A.R = new A.R(1); writeln(r);
```

These work:

```chapel
use A;  var r       = new A.R(1);
use A;  var r       = new A.R(1); writeln(r);   // working variant with new A.R(1).
use A;  var r       = new   R(1);
use A;  var r       = new   R(1); writeln(r);
use A;  var r :   R = new   R(1); 
use A;  var r :   R = new   R(1); writeln(r);   // initial working example above
use A;  var r : A.R = new   R(1);
use A;  var r : A.R = new   R(1); writeln(r);   // Brad's working example above

        var r       = new A.R(1);
```


This fails to resolve ``R.writeThis``:

```chapel
        var r       = new A.R(1); writeln(r);
```

lydia-duncan:
This is a variation on #6483, which notably did not include an explicit type for the declared variable and was resolved by #6810 and friends in July

bradcray:
[...]

PR #8708, opened 2018-03-05
Add tests showing problems with fully qualified 'new' expressions

bradcray:
Here, I'm adding four tests that all have the characteristic of using
fully-qualified new expressions: 'new moduleName.typeName()`.

The first two of these capture working and non-working cases that were
identified on issue #8555 but never captured as tests:

* qualifiedTypeAndNew.chpl   (doesn't work)
* qualifiedNew.chpl          (works)

The next two capture a different issue reported by Brian Rogoff in
which a module and type share the same name (filed as new issue
#8709):

* modTypeShareName.chpl (doesn't work)
* modTypeDiffName.chpl  (works)


bradcray:
@lydia-duncan, @cassella, @bryantlam: In this PR, I'm filing a future against the problem Bryant identified in #8555 as well as a related one that showed up on email (and working variants of each).

lydia-duncan:
Did you double check the working test cases against those used in the issues I referenced on #8555?

bradcray:
No.  Do you mean from the perspective of "do we already have a test that covers this case?"

lydia-duncan:
> Do you mean from the perspective of "do we already have a test that covers this case?"

Yeah.  Since a variant was fixed a while back, it seems reasonable that some tests for the specific cases that work already exist

bradcray:
True.  I generally don't think it's worth a lot of developer time to try and determine whether tests for user-submitted bugs are pre-existing or not.  And as we all know, it doesn't take that much of a delta between two tests to make one work when another doesn't.  My filing of the working tests was only intended to provide a point of comparison for the .futures (and is also how I generated their .good files).

lydia-duncan:
Agreed, and I'm not against having duplicate tests.  Was just pointing it out as a quick way to have maybe saved you some work since I had the link handy, but I think it is too late for that now so I don't think it's worth dwelling on further.

PR #8734, opened 2018-03-06
Phase one by default for classes

benharsh:
This PR enforces phase one by default for classes.

Many tests were updated to replace super.init() in a record initializer with this.initDone(). Some tests that assumed phase two was the default required an initDone to be inserted at the top of the initializer.

Many branches testing 'initNew' have been collapsed. The ``hasInit`` helper function has been updated to return a value representing the kinds of init statements found within a block. This function will go away once we emit errors for field initialization before super.init.

Testing:
- [x] full local + futures
- [x] full gasnet


test/classes/initializers/cond-only-else-1.good
```diff
@@ -1,2 +1,2 @@
 cond-only-else-1.chpl:16: In initializer:
-cond-only-else-1.chpl:19: error: Both arms of a conditional must use this.init() or super.init() in phase 1
+cond-only-else-1.chpl:19: error: Both arms of a conditional must use this.init() or this.initDone() in phase 1
```
cassella: This seems odd.

benharsh: The rationale behind this error is that the phase of the initializer should be the same at the end of all branches of a conditional. this.init() and initDone() both transition to phase two.

Does that make this seem less odd, or was there something else?

cassella: I meant, this.init() and super.init() seem like peers in many senses, but the error mentions only this.init() as an option.

benharsh: In the new proposal when super.init() is called the initializer transitions to phase one, which doesn't match the phase after this.init() (phase two).

cassella: Ah, ok, I hadn't considered that.

Will that same error message be used if there is an explicit super.init(), followed a conditional with one arm that uses an initDone()?

benharsh: You mean like this?
```chpl
if foo then
  super.init();
else
  this.initDone();
```

If so, then yes: that same error message will be used.

cassella: I meant

```chapel
super.init();
if foo then
   ...;
else
  this.initDone();
```

in which case the error message would be "suggesting" this.init() when I think it's not allowed?

(Or a field initialization before the conditional.)

benharsh: Ah, I see. Good catch! I'll put it on the growing list of error message improvements 😄 

I plan on improving error messages over the next couple of weeks, so hopefully it will be resolved/improved soon. Feel free to create an issue if you like, but don't feel the need to submit a test case just yet. I can do that as I dig into error messages.

cassella: Another approach might be to make the error message concerned with phase transitions: "If one branch transitions from phase 1 to phase 2, then all need to."  (for my case), or "from phase 0 to phase 2" if one branch had a legal this.init() and the other branch had nothing.

In the world where this.initDone() can be inferred (I think I read that somewhere), it would have to be more nuanced about one branch requires a transition that another branch isn't compatible with, or something.

PR #8760, opened 2018-03-09
Improve the comm performance many-to-one test.

### LINKED ISSUE #8681: many-to-one comm perf test needs improvement
### Task

The performance/comm/low-level/many-to-one performance test needs improvements:

- The homespun barrier should be replaced with a `Standard` all-locales barrier now that we have the latter in a high-performance version.

- We should check if changing the target node of the many-to-one loading from node 0 to node `numLocales-1` would improve the results stability, especially for the `GET` and `fetchAMO` operations where we're seeing rather variable results.

gbtitus:
Replace the homegrown barrier used before in this test with a standard
`allLocalesBarrier`.  The barrier improvements in #8550 and follow-ons
hadn't quite arrived when this test was written, so it used a homegrown
barrier.  Now that the standard barrier is available, use it instead.
This code for this change is largely the work of @ronawho.

In testing this seems to have improved execution time stability and the
rates we measure when doing GETs and fetching AMOs.  Those used to be
quite variable, so the homegrown barrier may have had unappreciated
problems.

Restructure to remove load conflicts between node 0 and the target.  In
the past node 0 has served as the target of the many-to-one loading for
this test.  However, it is simultaneously serving as the host, if you
will, of the coforall-stmt across nodes.  Switch to using node 1 as the
many-to-one target, so that node 0 doesn't have 2 jobs.  This doesn't
have much impact on the performance results, but it should only affect
the non-fast executeOn performance anyway, and it still seems like the
right thing to do.

Improve the var names associated with performance measurement and
reporting, and also the phrases used to report the performance.
Change the .ml-keys and .ml-perf.graph files to match the new output
phrasing.

Note that if we adopt this latter change any performance .dat files
containing results for this test will have to be changed to match the
new key phrases also, or graph generation will fail.

While here, also remove some unused code left over from earlier
development.

This closes #8681.

### DIFF
--- a/test/performance/comm/low-level/many-to-one.ml-keys
+++ b/test/performance/comm/low-level/many-to-one.ml-keys
@@ -1,2 +1,2 @@
-Execution time per source =
-Performance (mOps/sec to target) =
+Execution time =
+Performance (mOps/sec) =
--- a/test/performance/comm/low-level/many-to-one.ml-keys
+++ b/test/performance/comm/low-level/many-to-one.ml-keys
@@ -1,2 +1,2 @@
-Execution time =
-Performance (mOps/sec) =
+Execution time per source =
+Performance (mOps/sec to target) =
--- a/test/performance/comm/low-level/many-to-one.chpl
+++ b/test/performance/comm/low-level/many-to-one.chpl
@@ -45,14 +45,14 @@ proc main() {
   var numOpsPerNode: [1..numWorkerNodes] int;
   var timePerNode: [1..numWorkerNodes] real;
 
-  coforall locIdx in 1..numWorkerNodes with ( ref x0 ) {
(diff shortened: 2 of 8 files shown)


test/performance/comm/low-level/many-to-one.ml-keys
```diff
@@ -1,2 +1,2 @@
-Execution time =
-Performance (mOps/sec) =
+Execution time per source =
```
ronawho: Note that this key renaming will requiring editing the .dat files, which is pretty annoying/cumbersome. It's possible if the renaming is important, but if it's not I would try to avoid it. 

See internal chapel-docs/bestPractices/updatingDatFiles.rst and [util/devel/updateDatFiles.py](https://github.com/chapel-lang/chapel/blob/master/util/devel/updateDatFiles.py)

PR #8762, opened 2018-03-09
Add six tests of records and initializers with arrays

lydia-duncan:
Default initializers revealed an issue when a record stored as a field an array
of other records that used default initializers.  These tests demonstrate that
the problem only occurs when both types utilize default initializers.

However, there is a separate issue with inferring the type for the explicit
initializer array argument.  It seems to only impact initializers, thankfully
(as another new test demonstrates)


lydia-duncan:
TODO: add a future for the implicit type version

test/classes/initializers/records/arrayOfOtherRecs-implicitType.chpl
```diff
@@ -0,0 +1,20 @@
...
+  proc init(xVal: int = 0) {
+    x = xVal;
+    this.initDone();
+  }
+}
+
+record StoresR {
+  var x: [1..3] R;
+
+  proc init(arr = [new R(2), new R(3), new R(4)]) {
```
lydia-duncan: Fails with an internal compiler segfault as written

PR #8783, opened 2018-03-11
Add overload for eye for CSDomain

### LINKED ISSUE #8192: Implement matPow in LinearAlgebra.Sparse
Or, if it is implemented, please help me figure out why it's failing.  Here is my message.  I seem to have a special talent for finding these things Friday afternoons after everyone has gone home.

I've tried this using `LinearAlgebra, LinearAlgebra.Sparse` and it's power set.  Same error every time.

This guy works
```chapel
  var S = fcnBigSparseMat();
  var P = S.dot(S);
```

This version fails
```chapel
  var S = fcnBigSparseMat();
  var P = matPow(S,2);
```

And the crying starts...

```bash
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:309: In function 'eye':
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:310: error: can't apply '#' to a range with idxType int(64) using a count of type CSDom(2,int(64),domain(2,int(64),false),true,false)
make: *** [setup] Error 1
```

npadmana:
A possible solution to #8192, by adding in a simple overload for CSDomain as suggested by @cassella .This simple fix adds that in, and includes a simple test. Ideally, this should work for all sparse domains, but I can't seem to figure out how to get a more generic sparse domain.

As an added bonus, fix an issue where matPow would fail for integer arrays (we were not passing the `eltType` to `eye`).

(From 35,000 feet above sea level, during a long flight!)

Resolves #8192


npadmana:
@ben-albrecht  -- care to review?

test/library/packages/LinearAlgebra/correctness/testPowSparse.chpl
```diff
@@ -0,0 +1,20 @@
+use LinearAlgebra;
```
ben-albrecht: I have generally been trying to keep most of the correctness coverage tests in `correctness.chpl` to make it easier to identify if a feature has a test and to reduce overall compilation time. I'm fine with this test being a standalone test, but it may get merged into another file at some point.

modules/packages/LinearAlgebra.chpl
```diff
@@ -631,7 +631,7 @@ pragma "no doc"
 private proc _expBySquaring(x: ?t, n): t {
     // TODO -- _expBySquaring(pinv(x), -n);
     if n < 0  then halt("Negative powers not yet supported");
-    else if n == 0  then return eye(x.domain);
+    else if n == 0  then return eye(x.domain, x.eltType);
```
ben-albrecht: Nice catch!

modules/packages/LinearAlgebra.chpl
```diff
@@ -1572,6 +1572,17 @@ module Sparse {
     compilerError("Matrix division not yet supported for sparse matrices */");
   }
 
+  /* Return an identity matrix over sparse domain ``Dom`` */
+  proc eye(Dom, type eltType=real) where isCSDom(Dom) {
+    const (m,n) = Dom.shape;
+    var D = CSRDomain(m,n);
```
ben-albrecht: I know you're following the dense `eye()` implementation, but I just noticed an issue - this does not preserve the input `Dom` indices. Perhaps this line should do `var D = CSRDomain(Dom.parentDom)` instead.

npadmana: Quick question -- what does `Dom.parentDom` return?

test/library/packages/LinearAlgebra/correctness/testPowSparse.chpl
npadmana: Do you want me to merge this into `correctness.chpl`?

modules/packages/LinearAlgebra.chpl
npadmana: I think I answered my last question, but here's another -- are we requiring that matrix indices start at 1?

ben-albrecht: > I think I answered my last question, but here's another -- are we requiring that matrix indices start at 1?

The `LinearAlgebra` library's stance is: We'll preserve indices where we can, otherwise we'll give you 1-based indices. Implementations should not be index-dependent.

test/library/packages/LinearAlgebra/correctness/testPowSparse.chpl
ben-albrecht: If it's not a hassle. It may make sense to break `correctness.chpl` up into subsections or at least the submodules some day, but I don't expect it to ever reach the granularity of individual functions.

modules/packages/LinearAlgebra.chpl
ben-albrecht: and if that stance is not fulfilled anywhere, I would consider it a bug.

npadmana: Ok, well -- I think `_eyeDiagonal` isn't strictly correct then for 
```
proc eye(Dom: domain(2), type eltType=real) {
```
overload of `eye`, right?

npadmana: Ignore the last comment -- I was being an idiot; the code is fine.

ben-albrecht: I almost agreed with you :) That's a tricky one.

npadmana:
@ben-albrecht -- fixed the domain issue, and merged in the correctness tests.

PR #8787, opened 2018-03-12
Add read/write methods for date, time, and datetime types

daviditen:
Add `readWriteThis` methods for the date, time, datetime types. Add a test to
ensure that the methods can be called and that they output and input matching
values when appropriate.


buddha314:
Go David Go!!

marcoscleison:
This issue is important. Thanks!

modules/standard/DateTime.chpl
```diff
@@ -434,6 +434,13 @@ module DateTime {
     return str;
   }
 
+  /* Read or write a date value from channel `f` */
+  proc date.readWriteThis(f) {
+    var dash = new ioLiteral("-");
+    f <~> new ioLiteral("{") <~> chpl_year <~> dash <~> chpl_month <~> dash
+      <~> chpl_day <~> new ioLiteral("}");
```
mppf: Is there a specific reason for the { } around the date? I'd think 2018-03-12 is clear enough as a date already. Are you following some Python precedent, e.g.?

modules/standard/DateTime.chpl
```diff
@@ -658,6 +664,15 @@ module DateTime {
     return str;
   }
 
+  /* Read or write a time value from channel `f` */
+  proc time.readWriteThis(f) {
+    var colon = new ioLiteral(":");
+    f <~> new ioLiteral("{") <~> chpl_hour <~> colon <~> chpl_minute <~> colon
+      <~> chpl_second <~> new ioLiteral(".") <~> chpl_microsecond
+      <~> new ioLiteral("}");
```
mppf: Same question here about {.

test/library/standard/DateTime/readWriteDatesTimes.good
```diff
@@ -0,0 +1,3 @@
+datetime: true
+date: true
+time: true
```
mppf: Do we have any tests that actually show the date-time formats in the .good file? Might be worth adding one so we can be sure it doesn't accidentally change.

mppf:
Should this be following [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)?

buddha314:
@mppf you probably have a point.  I'm not sure the proper way to set it.

mppf:
The readWriteThis methods can check for json like so:
``` chapel
      var binary = f.binary();
      var arrayStyle = f.styleElement(QIO_STYLE_ELEMENT_ARRAY);
      var isjson = arrayStyle == QIO_ARRAY_FORMAT_JSON && !binary;
```

modules/standard/DateTime.chpl
daviditen: I was following the default behavior of Chapel classes, which doesn't make a whole lot of sense given that these are records.  I'll remove the curly braces.

test/library/standard/DateTime/readWriteDatesTimes.good
daviditen: I'll add one.

daviditen:
> Should this be following ISO 8601?

This means zero-padding the year to 4 digits and everything else to 2 digits, plus putting a 'T' between the date and time, right?  The padding sounds fine, but I'm not crazy about the 'T'.

mppf:
> > Should this be following ISO 8601?
> 
> This means zero-padding the year to 4 digits and everything else to 2 digits, plus putting a 'T' between the date and time, right?  The padding sounds fine, but I'm not crazy about the 'T'.

I think so... and generally agree. But at least ISO 8601 is a standard.

mppf:
Looking good, thanks.

daviditen:
Thank you @mppf!

PR #8790, opened 2018-03-12
Enable phase one dynamic dispatch

benharsh:
This PR updates the compiler to transform ``this`` in phase one to have the type and class ID of the class it inherits. The class ID is reset upon a call to initDone or at the end of the initializer.

Users can now call methods in phase one, provided that the parent implements the invoked method. Users can also now pass ``this`` to functions in phase one. Passing ``this`` before super.init or this.init is an error.

Testing:
- [x] full local + futures
- [x] full gasnet
- [x] partial numa
- [x] partial LLVM

### DIFF
--- a/test/classes/initializers/generics/phase1/accessThisBad.good
+++ b/test/classes/initializers/generics/phase1/accessThisBad.good
@@ -1,2 +1,2 @@
 accessThisBad.chpl:4: In initializer:
-accessThisBad.chpl:5: error: cannot pass "this" to a function in phase 1 of initialization
+accessThisBad.chpl:5: error: cannot pass "this" to a function before calling super.init() or this.init()
--- a/test/classes/initializers/phase1/dynamicThis/castThis.chpl
+++ b/test/classes/initializers/phase1/dynamicThis/castThis.chpl
@@ -1,0 +1,26 @@
+
+class Parent {
+  var x : int;
+
+  proc init() {
+    this.x = 5;
+  }
+
+  proc inheritedMethod() {
+    writeln("Parent.inheritedMethod()");
+  }
+}
+
+class Child : Parent {
+  var y : real;
+
+  proc init() {
+    this.y = x ** 2;
+    inheritedMethod();
+    (this:Child).inheritedMethod();
+  }
+
+  proc inheritedMethod() {
+    writeln("Child.inheritedMethod()");
+  }
+}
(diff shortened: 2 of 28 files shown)


test/classes/initializers/generics/phase1/accessThisBad.good
```diff
@@ -1,2 +1,2 @@
 accessThisBad.chpl:4: In initializer:
-accessThisBad.chpl:5: error: cannot pass "this" to a function in phase 1 of initialization
+accessThisBad.chpl:5: error: cannot pass "this" to a function before calling super.init() or this.init()
```
bradcray: I really like that this message (and others in this PR) go in the direction of getting away from the phrase "phase 1" in the error messages.

test/classes/initializers/phase1/dynamicThis/genericParent.chpl
```diff
@@ -0,0 +1,49 @@
...
+  proc init(type t) {
+    super.init(t);
+    inheritedMethod();
+    foobar(this);
+    initDone();
+    inheritedMethod();
+  }
+
+  proc inheritedMethod() {
+    writeln(this.type:string, ".inheritedMethod()");
```
bradcray: One quick comment on this test:  I may be missing something, but I think that the definitions of inheritedMethod() on both the parent and the child class are identical which makes me think that we can't tell which was called if something was going wrong (?).

test/classes/initializers/phase1/dynamicThis/castThis.chpl
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+
+class Child : Parent {
+  var y : real;
+
+  proc init() {
+    this.y = x ** 2;
+    inheritedMethod();
+    (this:Child).inheritedMethod();
```
bradcray: I agree that this cast shouldn't be legal in phase 1.  The presence of the test (and the fact that it fired even when we don't create a `new Child()` suggesting it's a syntactic pattern rather than a semantic one) made me wonder whether the user ought to be able to cast to an ancestor class like `Grandparent` in phase 1 (and/or whether doing so would have any actual utility... I have to admit that I can't quite remember the behavior of a dynamic cast on a class variable in Chapel when the result is non-`nil`).

test/classes/initializers/phase1/dynamicThis/genericParent.chpl
benharsh: Those methods should print the type of 'this' when called, so I think we're fine. Unless you're seeing something that I'm not?

bradcray: Right, but if, during phase 2, we called the `parent.inheritedMethod()` version, wouldn't it print the child's type even if the child version was supposed to have been called?  My thought was that if we put a `(parent)` or `(child)` label into each of the two routines, we could validate not only that the dynamic type was as expected, but also that the correct method was being called.  But again, I may be missing something else.

PR #8791, opened 2018-03-12
Add support for Owned,Shared -> class coercions

### LINKED ISSUE #6045: Surprising type mismatch with Owned, matches method

mppf:
This PR enables implicit conversions from Owned(C) to C.

In order to inform compiler about Owned create a flag marking Owned and Shared types.

Additionally, don't add default record cast functions for Owned in order to reduce confusion in error messages.

Relates to issue #8375.
Closes #6045.

* Adjust compiler to be aware of Owned,Shared types via flag
* Adjust compiler to allow coercions from Owned,Shared->class type
  by calling borrow()
* Apply workaround to resolve owned/shared autoDestroy
* Add new tests

More about the workaround:

I added chpl__autoDestroy overloads for Owned and Shared. This
is a workaround but I tried several other strategies:
 - changing the "any type" less specific than concrete resolution rule
 - inline proc chpl__autoDestroy(x: ?t) where t:object { }
 - if !isClassType(t) in generic proc chpl__autoDestroy(x: ?t)

All of these led to many failures. Consider

  arrays/stonea/outOfBoundsSlice.chpl
  arrays/diten/sliceStridableByNon.chpl

which for the latter 2 strategies above give compilation errors
such as

 ChapelBase.chpl:1168: error: type mismatch in assignment from [domain(1,int(64),false)] int(64) to [domain(1,int(64),false)] int(64)

where the type mismatch has to do with an array view vs an array.

- [x] full local testing
- [x] full --no-local testing


compiler/resolution/wrappers.cpp
```diff
@@ -1467,14 +1493,37 @@ static void addArgCoercion(FnSymbol*  fn,
...
+
   } else if (isSingleType(ats->getValType()) == true) {
     checkAgain = true;
 
     castCall   = new CallExpr("readFF", gMethodToken, prevActual);
 
+    if (argumentCanModifyActual(formal))
+      issueErrorValueTemporaryToRef(call, formal);
+
+  } else if (isManagedPtrType(ats->getValType()) == true &&
```
daviditen: I know that this code is already full of these, but comparing `boolean == true` ='(

Issue #8301, opened 2018-01-24
Convert standard/internal/package modules to initializers


bradcray:
This issue is intended to capture the work remaining to convert the modules/ directory's
contents over from constructors to initializers.  This is an effort I've been working on during
spare-time cycles over the past month or so and have gotten pretty close on.  Here are
cases that still need work:

### Internal
- [x] **OwnedObject.chpl:** In its purest form, this is failing because the compiler-provided chpl__autoCopy/initCopy() routines don't support passing RHS arguments by `ref` to copy initializers.  We could work around this if necessary, but I've been dragging my feet in hopes of not requiring this to be a special-case.  See issue #8234.
- [x] **first-class functions** #9535
- [x] **Array, Domain, and Distribution record wrappers:** Tricky due to existing initCopies returning a different type. (#9313)
- [x] **Reductions** (#9315)
- [x] **Atomics**
- [x] **Tuples**
- [x] **SharedObject.chpl:** This is ready to go in PR #8170
- [x] *all modules supporting domain maps, domains, arrays:* These ran afoul of limitations in our current initializer story, being discussed in issue #8283.
- [x] **Range.chpl:** Originally ran into a promotionType issue that BHarsh resolved, and is ready for another look.  In its current form, likely to run into limitations relating to conditional expressions and compiler-provided initializers, but the conditional expression has also been noted as being arguably illegal (due to calling a method), so needs a look.
- [x] **ChapelSyncVar.chpl:** Originally ran into an issue with calling `new` on the result of a type function (issue #7935), now believed to be fixed for this case and ready for another look.
- [x] **String.chpl:** These run into a number of tricky cases due to the compiler's built-in support for strings (e.g., string literals like "hello" get special treatment in the compiler). Resolved by #9217.

### Standard
- [x] IO.file and other miscellaneous records in IO module
- [x] List
- [x] Regexp.regexp
- [x] Random
- [x] Loc* classes in distributions (#9314)
- [x] **Buffers.chpl:** Simply needs post-review edits and merge; however, current testing is arguably insufficient to have confidence in copy initializers.

### Package
- [x] **DistributedBag.chpl / DistributedQueue.chpl:** Same as above.

bradcray:
Filtering out the cases known not to be converted above (domain maps, OwnedObject, string), I got failures for the following tests, some of which are designed to ensure constructors are still working.  The only case I think I detected which was left unconverted in modules was `RecordParser` which I'll take a quick look at:

[...]

PR #8807, opened 2018-03-13
Convert more Distributed* constructors to initializers

bradcray:
In checking to see whether we could check off the Distributed* modules from issue #8301, I
noticed that some constructors still existed in these modules, so converted them to
initializers.  These were quite straightforward.


bradcray:
@benharsh or @noakesmichael: Would one of you give this a review?

benharsh:
Looks reasonable to me.

PR #8810, opened 2018-03-13
Convert RecordParser over from constructors to initializers

bradcray:
This case was only complicated by the fact that the constructor
had thrown an error upward and the initializer does not.  For
now, I put a try! around the throwing call and commented out
the throws.


bradcray:
This case is fairly trivial, but since it involved squashing a throw from an init (since that's not supported yet), I didn't want to merge it without @psahabu , @mppf, @lydia-duncan, and/or whoever else was interested taking a look at it.

PR #8811, opened 2018-03-13
Remove remaining destructors from modules code

bradcray:
A quick sanity check to see whether we still had destructors in
the modules turned up only two instances as well as some error
messages that were still written from the constructor/destructor
perspective.  Removing these primarily as housecleaning.

### DIFF
--- a/modules/internal/ChapelDistribution.chpl
+++ b/modules/internal/ChapelDistribution.chpl
@@ -780,7 +780,7 @@ module ChapelDistribution {
     proc dsiPostReallocate() {
     }
 
-    proc ~BaseArrOverRectangularDom() {
+    proc deinit() {
       // this is a bug workaround
     }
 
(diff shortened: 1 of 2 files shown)


modules/internal/ChapelDistribution.chpl
```diff
@@ -780,7 +780,7 @@ module ChapelDistribution {
     proc dsiPostReallocate() {
     }
 
-    proc ~BaseArrOverRectangularDom() {
+    proc deinit() {
       // this is a bug workaround
```
cassella: OOC, is the bug being worked around still present in the initializers world?

@mppf added it in a15c19cc128087, and @daviditen merged it in #6722, last January and July, respectively.

bradcray: I wondered that as well but did not have time to look into it.

Issue #8489, opened 2018-02-15
subclassing and ref arguments


mppf:
``` chapel
class MyClass {
  var x:int;
}
class SubClass : MyClass {
  var y:int;
}

proc acceptsConstRefMyClass(const ref arg:MyClass) {
  writeln(arg);
}

proc acceptsRefMyClass(ref arg:MyClass) {
  arg = new MyClass(1);
}


proc test() {
  var mySubClass = new SubClass(1,2);
  acceptsConstRefMyClass(mySubClass);
  acceptsRefMyClass(mySubClass);
  writeln(x);
}

test();
```

I'm expecting that `acceptsConstRefMyClass(mySubClass)` work without error but that `acceptsRefMyClass(mySubClass)` cause some sort of compilation error.

Instead, both of these cases get a warning from GCC during C compilation.

mppf:
I'm thinking about this issue since it relates to the topic of allowing implicit conversion from `Owned(SubClass)` to `Owned(MyClass)` and also similarly from `const ref Owned(SubClass)` to `const ref Owned(MyClass)`

dmk42:
The idea looks understandable in Chapel.

If we implement this, we'll need to change LLVM TBAA metadata to take it into account.

The reason for the warnings in C (which become errors in testing) is that a C cast to the base class would be needed to make the types compatible.

vasslitvinov:
The `ref` case is definitely a NO. Michael's `acceptsRefMyClass(mySubClass)` makes `mySubClass` point to an instance of `MyClass`, which is a violation of type safety.

I do not see how we can break things with `const ref`. So let us allow `acceptsConstRefMyClass(mySubClass)`. (Unless we find an argument to the contrary :).)

For `Owned(SubClass)` vs. `Owned(MyClass)`, I still think these can be in a subtype relationship. Passing a subtype to a `ref` argument expecting a supertype should be disallowed, other cases are OK.

```chpl
proc acceptsConstRefOwnedMyClass(const ref arg: Owned(MyClass)) { ... }

proc acceptsRefOwnedMyClass(ref arg: Owned(MyClass)) { ... }

var oChild: Owned(SubClass) = ...;
var oParent: Owned(MyClass) = oChild; // OK: cannot "break" oChild

acceptsConstRefOwnedMyClass(oChild);  // OK: by const ref
acceptsRefOwnedMyClass(oChild);       // ERROR: passing subtype by ref
```

mppf:
[...]

PR #8812, opened 2018-03-13
Support coercion from Owned(Child) to Owned(Parent)

mppf:
This PR enables value coercions from Owned(Child) to Owned(Parent). Coercions in a reference context, such as with a `const ref` argument are not currently supported. Additionally coercing a class (not owned) to a parent class type for a `const ref` argument results in backend C compiler warnings from pointer type mismatches. This PR creates a new error message in resolution for that case which also causes errors for passing an Owned that needs coercion to a const ref argument.

- [x] full local testing
- [x] add Owned example from #8489
- [x] futures testing
- [x] baseline testing


compiler/resolution/wrappers.cpp
```diff
@@ -1592,6 +1592,14 @@ static void addArgCoercion(FnSymbol*  fn,
   }
 
   resolveCall(castMove);
+
+  IntentTag concreteIntent = concreteIntentForArg(formal);
+  bool formalIsRef = formal->isRef() || (concreteIntent & INTENT_REF);
+
+  if (!castTemp->isRef() && formalIsRef) {
+    USR_FATAL_CONT(call, "Cannot coerce argument for const ref intent");
+  }
```
mppf: I think it'd be better to fold this into a checking function that includes/replaces calls to argumentCanModifyActual/issueErrorValueTemporaryToRef

mppf:
@daviditen - do you think the error in classes/diten/passSubClassByRef is good enough to un-future it? I mention this question in the last comment in issue #8489. AFAIK we don't currently choose candidates based upon the mutability of the actual and I think it'd be a big change to do so. I suspect the unresolved call error was there more as an example of some sort of error, though.

compiler/resolution/wrappers.cpp
```diff
@@ -1435,8 +1431,33 @@ static bool argumentCanModifyActual(ArgSymbol* formal) {
...
+    // OK with a 'const ref' but right now there are errors at C
+    // compilation time if this error is left out.
+    // Additionally, if a new value is created for this kind of
+    // coercion, it disrupts the desired semantics (a value passed
+    // by const ref could be modified during the call & the change
+    // visible).
+    bool formalIsRef = formal->isRef() || (intent & INTENT_REF);
+
+    if (formalIsRef) {
+      USR_FATAL_CONT(call, "Cannot coerce argument for const ref intent");
```
daviditen: It would be nice to print the formal name here too.

modules/packages/OwnedObject.chpl
```diff
@@ -204,7 +204,7 @@ module OwnedObject {
 
   pragma "no doc"
   inline proc _cast(type t, in x) where t:Owned && x:Owned && x.t:t.t {
-    var ret = new Owned(x.release());
+    var ret = new Owned(x.release():t.t);
```
daviditen: A comment here explaining why the cast is needed would be helpful.  It took a minute staring at it to figure it out.

test/classes/diten/passSubClassByRef.bad
```diff
@@ -1,7 +1 @@
-passSubClassByRef.chpl:6: internal error: CHEnnnn chpl version mmmm
-
-Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
-and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
-please see https://chapel-lang.org/bugs.html for instructions.  In the meantime,
-the filename + line number above may be useful in working around the issue.
-
+passSubClassByRef.chpl:6: error: value from coercion passed to ref formal 'obj'
```
daviditen: This works for me to de-futurize.

PR #8816, opened 2018-03-14
Add skipif for time-write performance test with launcher

lydia-duncan:
This test was on the verge of timing out before, but with some machine updates
starting doing so.  We could adjust the problem size for this new behavior, but
I'm not too worried about the difference between launcher writelns and local
ones so at Elliot's suggestion skip the performance test version when the
launcher is on

Validated a fresh checkout that this doesn't prevent a correctness run or
an ordinary perf run for the test, and does prevent a perf run with a launcher
(but not a correctness launcher run)

### DIFF
--- a/test/io/vass/time-write.skipif
+++ b/test/io/vass/time-write.skipif
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+
+import os
+
+print(os.getenv('CHPL_TEST_PERF') == 'on' and
+      os.getenv('CHPL_LAUNCHER') != 'none')
--- a/test/io/vass/time-write.skipif
+++ b/test/io/vass/time-write.skipif
@@ -2,5 +2,7 @@
 
 import os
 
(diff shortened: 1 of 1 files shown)


lydia-duncan:
@ben-albrecht - as triager, mind giving this a sanity check?

test/io/vass/time-write.skipif
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
+
+import os
+
```
ben-albrecht: It may be worth documenting why this skipif exists as a doc-string at the top of this file.

test/io/vass/time-write.skipif
```diff
@@ -2,7 +2,10 @@
 
 import os
 
-# test is long running, but no need to validate it outside of normal
-# perf testing.
+"""
+This test measures performance writes, where we only care about on-node
```
ben-albrecht: 10/10 comment, would read again.

PR #8818, opened 2018-03-14
Allow Phase 1 of initialization to contain if statements with field init and no else branch

lydia-duncan:
Prior to this change, we would generate an error if an initializer contained an
`if` statement with just a single branch that handled field initialization.  E.g.
```chapel
proc init(xVal) {
  if xVal > 10 {
    this.x = xVal;
  }
}
```
We intended to eventually support this change, but originally viewed it as a
bit of a distraction.  However, adding the support was fairly straight-forward:
if no `else` branch is provided, create one and fill it with field initialization up to
the last field covered in the `then` branch.

Note that this change does not support conditionals where the branches initialize
different fields, e.g.:
```chapel
proc init(val) {
  if val > 10 {
    this.x = val;
  } else {
    this.x = val / 2;
    this.y = val;
  }
}
```
I intend to tackle that case next.

Modifies a few tests that were validating the old error message for the
new expected behavior, including fixing one that was determining a param
field value based on a var field (which was only allowed before because we
were checking the error message about if statements with no else branches)
and making it a future because it encounters the same issue as another if
statement test.

Note that this also slightly changed the error reporting for a few tests.  The
change reports one correct error instead of a different correct error (that
was closer in scope), to avoid failing to report an error at all in some cases.
Ideally we would like to report both but that is beyond the scope of this PR.

Testing:
- [x] full std/ paratest
- [ ] full std/ futures


lydia-duncan:
@noakesmichael - care to review?

PR #8821, opened 2018-03-14
Define a test macro directly instead of indirectly

dmk42:
When the result of expanding a macro contains the `defined` preprocessor operator, the behavior is undefined. (C 6.10.1p4)  Clang caught this use of `defined` for us.  This change defines the macro directly instead of defining it in terms of `defined` operators.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -39,9 +39,12 @@
 #include <gni_pub.h>    // <stddef.h> and <stdint.h> must come first
 #include <hugetlbfs.h>  // <sys/types.h> must come first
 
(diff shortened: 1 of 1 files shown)


dmk42:
@gbtitus -
The earlier thing that Clang made me do was to add the double braces to the following.  That was just being too picky.
```c
static gni_post_descriptor_t polling_post_descs[PPDESCS_CNT] = {{ 0 }};
```
However, this time Clang caught actual undefined behavior.

If the solution in this PR is OK with you, would you please review it?  If not, that's fine.  I can close this PR and you can do it a different way.

gbtitus:
Ouch!  I didn't realize that generating a `defined` token as a result of expansion in the conditional of an `#if` was undefined!  Thanks for fixing this.
👍

PR #8822, opened 2018-03-14
Remove domain preservation LinearAlgebra.Sparse.dot test

ben-albrecht:
Domain preservation and support for offset domains is not yet implemented for `LinearAlgebra.Sparse.dot`, so I am removing this test to quiet nightly testing until we have an implementation. 

I have a WIP implementation that uses reindexing to address this, but will need more time to ensure performance does not degrade as a result of reindexing.

Domain preservation for matrix multiplication is a bit of an open question, which I will create a separate issue for. 

Also, removed trailing whitespace and added wordsmithed comments to match style of other test subsections.

### DIFF
--- a/test/library/packages/LinearAlgebra/correctness/correctness.chpl
+++ b/test/library/packages/LinearAlgebra/correctness/correctness.chpl
@@ -856,42 +856,47 @@ use TestUtils;
   }
 
 
-  // Tests for matPow with sparse matrices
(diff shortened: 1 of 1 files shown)


ben-albrecht:
FYI @npadmana

ben-albrecht:
Recording this as a use-case data point for #5368

PR #8824, opened 2018-03-14
Improved support for postInit

benharsh:
This PR improves support for postInit and enables the following:
- Compiler insertion of super.postInit
- handling absence of postInit between child and ancestor with postInit
- postInit implies default init
- postInit with where clause
- error for postInit with arguments

At the start of normalization we will look at each type and determine if a postInit needs to be inserted. If a type inherits from a type that implements a postInit, then the inheriting type must also have a postInit. If a postInit does not exist, one will be created.

A special case is if all postInits on a type have a where clause. In this case we need to generate a where-less postInit so that ancestor postInits will still be invoked.

Open questions:
- Should we do anything special for ``super.postInit`` calls in conditionals?
- Should users be allowed to call postInit manually?

Testing:
- [x] full local + futures
- [x] full gasnet
- [x] partial baseline

### DIFF
--- a/test/classes/initializers/postInit/args.good
+++ b/test/classes/initializers/postInit/args.good
@@ -1,0 +1,1 @@
+args.chpl:3: error: postInit must have zero arguments
--- a/test/classes/initializers/postInit/conditional.chpl
+++ b/test/classes/initializers/postInit/conditional.chpl
@@ -1,6 +1,4 @@
 
-config const flag = true;
-
 class A {
   var a : int;
   proc init() {
@@ -28,5 +26,10 @@ class B : A {
   }
 }
 
-var b = new B(flag);
-delete b;
+var t = new B(true);
+delete t;
+
+writeln();
+
+var f = new B(false);
+delete f;
--- a/test/classes/initializers/postInit/conditional.chpl
+++ b/test/classes/initializers/postInit/conditional.chpl
@@ -1,0 +1,32 @@
+
+config const flag = true;
+
+class A {
+  var a : int;
+  proc init() {
+    writeln("A.init");
+  }
+  proc postInit() {
+    writeln("A.postInit");
+  }
+}
+
+class B : A {
+  var b : bool;
+  proc init(b:bool) {
+    writeln("B.init");
+    this.b = b;
+  }
+  proc postInit() {
+    if b {
+      writeln("B.postInit");
+      super.postInit();
+    } else {
+      super.postInit();
+      writeln("B.postInit");
+    }
+  }
+}
+
+var b = new B(flag);
+delete b;
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -2021,6 +2021,10 @@ bool AggregateType::needsConstructor() {
   if (symbol->hasFlag(FLAG_USE_DEFAULT_INIT))
     return false;
 
+  if (hasPostInitializer() == true) {
+    return false;
+  }
+
   ModuleSymbol* mod = getModule();
 
   // For now, always generate a default constructor for types in the internal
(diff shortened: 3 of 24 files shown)


benharsh:
@bradcray : Would you be interested in reviewing the new tests?

test/classes/initializers/postInit/args.good
```diff
@@ -0,0 +1 @@
+args.chpl:3: error: postInit must have zero arguments
```
bradcray: I'm not sure we should disallow `postInit()` methods with arguments since the user could choose to use that name for some other purpose (though stylistically, we'd discourage that).  We could create a style-/novice-oriented warning to point out the possibility of a mistake, like "note that the compiler won't insert automatic calls to this postInit() routine because of its required arguments."

We should also distinguish between those that have zero arguments and those that can be called with zero arguments (i.e., if `x` had a default value above, I think it'd be a valid `postInit()` for the compiler to call).

test/classes/initializers/postInit/conditional.chpl
```diff
@@ -0,0 +1,32 @@
...
+      writeln("B.postInit");
+      super.postInit();
+    } else {
+      super.postInit();
+      writeln("B.postInit");
+    }
+  }
+}
+
+var b = new B(flag);
```
bradcray: I'd either add a .execopts that would test both values of this flag or, probably simpler, just create two objects of the type, one with B(true) one with B(false).

bradcray:
The tests that are here look good.  I've got one more test I'd like to see added which I suspect will work.  Let me write it up and send it to you because I think it's easier than trying to describe it.

test/classes/initializers/postInit/args.good
benharsh: I feel like it would be easier for users and ourselves if methods named 'postInit' were strictly limited to being a compiler hook. If a user wants to do something else, they should name it differently. From that point of view, I don't think there's much value in having a postInit where all formals have default values. It seems like an edge case not worth supporting (for now).

Would you be open to releasing with this error and relaxing later if demand is sufficient?

bradcray: (if it's simplest, I'd also be very OK (certainly for now) to just not have any kind of processing of postInit() routines whose signature is "surprising" to us.  This kindof relates to the conversation Vass and I were having the other day about the degree to which the compiler should prevent various signatures of init() and deinit() on modules.  One school of thought would be "anything that doesn't match what we're planning to call should be illegal because the user will only confuse themselves.  Another, and the one I'm leaning towards in both these conversations is, since those aren't reserved names, the compiler shouldn't prevent their use in other cases (but warnings might be useful to help prevent obvious mistakes)).

bradcray: Oops, I hadn't seen your comment before posting my follow-up

I suppose that what you've done is consistent with what we have for `deinit()`, so I wouldn't stand in its way.  In both cases, I do think that an error might be heavy-handed (though I also agree that the chances that a user cares very much is slim).

benharsh: OK. I'm happy to discuss this further if you want.

bradcray:
Here's the test I was thinking of.  It's designed to test dynamic dispatch within postInit() routines given that that's one of the two main reasons I think we have it.  It looks to me like it's working as I'd expect, but you should think through it and see if you agree:

```chapel
class C {
  proc foo() {
    writeln("In C.foo()");
  }

  proc bar() {
    writeln("In C.bar()");
  }

  proc postInit() {
    foo();
    bar();
    //    baz();  // not possible since C doesn't define it
  }
}

class D : C {
  proc foo() {
    writeln("In D.foo()");
  }

  proc baz() {
    writeln("In D.baz()");
  }
}

class E : C {
  proc foo() {
    writeln("In E.foo()");
  }

  proc baz() {
    writeln("In E.baz()");
  }

  proc postInit() {
    foo();
    bar();
    baz();
  }
}

writeln("C:");
var myC = new C();
writeln("D:");
var myD = new D();
writeln("E:");
var myE = new E();

delete myE;
delete myD;
delete myC;
```

benharsh:
Thanks, I'll take a look and add a test.

PR #8825, opened 2018-03-14
Fix least common denominator CPU type for ARM-based Cray systems

dmk42:
For a Cray XC system, we were always setting the least common denominator CPU architecture to "sandybridge".  That is no longer always right, because of ARM systems.  This change sets the LCD to arm-thunderx2, which is currently the only ARM processor on Cray products.

This allows the ARM module to build without references to x86-64 architectures.


dmk42:
@awallace-cray - Would you please review?  The module build script will need to be in sync with this.

PR #8826, opened 2018-03-14
ResolveNew :- Merge the two functions that handle new exprs for "generic" functions

noakesmichael:
Continue to adjust the resolution of PRIM_NEW

Before this PR there were two helper functions that handled two variations on
generic functions.  These functions are reasonably large/complex and are almost
identical.

This intermediate PR merges these two functions together.

Passed limited testing for each configuration.
Passed single-locale paratest with -futures

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -5699,25 +5699,25 @@ bool isDispatchParent(Type* t, Type* pt) {
 *                                                                             *
 ************************************** | *************************************/
 
-static void     resolveNewAT(CallExpr* call);
+static void           resolveNewAT(CallExpr* call);
 
-static bool     resolveNewHasInitializer(AggregateType* at);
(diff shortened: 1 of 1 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -6006,79 +5997,27 @@ static void resolveNewHandleInstantiatedInit(CallExpr* call) {
...
-
-  if (isBlockStmt(call->parentExpr) == true) {
-    call->insertBefore(def);
-
-  } else {
-    Expr* parent = call->parentExpr;
+  AggregateType* retval   = at;
 
-    parent->insertBefore(def);
+  if (at->instantiatedFrom != NULL) {
```
benharsh: There's similar code in ResolutionCandidate.cpp (and perhaps elsewhere), so we should probably make this a method on AggregateType eventually. No need to do so for this PR.

PR #8828, opened 2018-03-14
Add support for if/else branches ending with different field inits 

lydia-duncan:
Prior to this change, the following code would generate an error message.
With this change, it will now compile and run, inserting omitted initialization.
```chapel
proc init(xVal) {
  if (xVal > 5) {
    x = xVal;
    y = true;
  } else {
    x = xVal;
  }
}
```

This assumes that the intention is for omitted initialization to handle the
remaining fields in the branch that did not initialize as many.

Determines which branch is further ahead, and inserts omitted initialization for
all the remaining fields of the branch that is further behind.

Adds five tests.  One covers when both branches end on the same explicit
initialization, the others cover when the branches end on different fields,
whether fields remain or not.

Updates a few tests for new expected behavior.

Testing over:
- [x] full std/ with futures

### DIFF
--- a/compiler/passes/initializerRules.cpp
+++ b/compiler/passes/initializerRules.cpp
@@ -36,6 +36,11 @@ static bool     isUnacceptableTry(Expr* stmt);
 
 static void     preNormalizeInit(FnSymbol* fn);
 
+static void     unifyConditionalBranchLastField(AggregateType* at,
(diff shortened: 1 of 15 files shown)


lydia-duncan:
@noakesmichael, care to review?

PR #8829, opened 2018-03-15
Convert readBytes() to a public, Chapeltastic interface

bradcray:
This converts the throwing (public) version of readBytes() into a
routine intended for end users by documenting it and reducing the
reliance on C types/casts.  It also updates the revcomp study
versions to make use of this new interface.

The throwing writeBytes is similarly updated.

There's one internal call to readBytes() for strings that will need to
be adjusted still (I timed out today and wanted to snapshot this
effort).  It uses the throwing version but passes in C-like types.


bradcray:
Closing in favor of #8855.

PR #8835, opened 2018-03-15
Make testEnv call CHPL_TEST_UTIL_DIR's version of printchplenv

bradcray:
For some time now, `testEnv` has called the version of `printchplenv`
in `CHPL_HOME/util`, making it one of the only remaining scripts used
by the test system that uses this location rather than `CHPL_TEST_UTIL_DIR`. 
This caused a problem in generating release-over-release test
results this week because the `printchplenv` interface has changed since
the last release such that any tests with `.skipif` files were failing for past
versions of the release since their `printchplenv` version didn't support those
flags.

This patch makes it use the version in `CHPL_TEST_UTIL_DIR`, when that
variable is set, on the argument that it should be calling the version that
matches the scripts being used to run the test system itself not the one
corresponding to the version of `CHPL_HOME` that's being tested.  When
`CHPL_TEST_UTIL_DIR` is not set, it falls back to `CHPL_HOME/util/` as it
traditionally has (note that, in most cases, these two locations are identical,
so there's no real difference for most uses of the script).


mppf:
It's not obvious to me that using the printchplenv from the test dir rather than from CHPL_HOME will fix the issue for release-over-release testing. I suppose though it makes failure less likely, since the failure mode I would expect would be a change in a CHPL_BLA variable between releases that was used in a test setup.

Anyway, does this allow you to run release-over-release tests?

mppf:
The alternative strategy would be to detect old printchplenv and call it appropriately.

mppf:
I'm happy with the change itself; would just appreciate clarification on how far it got you.

bradcray:
I think you're correct that if a CHPL_* variable were to change over time, this doesn't provide any guarantee that a given test will work.  But I think that's similar to the general problem with release-over-release testing:  That if the current tests rely on some capability that wasn't present in older releases, things are unlikely to work (which is why our release-over-release graphs are missing so many historical data points).

The problem that I was hitting, which this PR does help with, is that when any test has a `.skipif`, `.suppressif`, etc. we run `testEnv` to populate the `CHPL_*` variables before evaluating the file.  But the interface to `printchplenv` changed since 1.16 such that it doesn't know about the [flag(s) that we now pass to it](https://github.com/chapel-lang/chapel/blob/master/util/test/testEnv#L8).  So any test that has a `.skipif`, `.suppressif`, etc. would fail for all older versions of the compiler even if the source code would compile because `testEnv` would try to run an older version of `printchplenv` that would complain about the flag being passed to it:

```
printchplenv: error: no such option: --all
```

and nothing further would happen for that test (even if the CHPL_* variable being queried was in the appropriate state, the source code would work with the older version of the compiler, etc.)

So this change runs the version of `printchplenv` that matches all of the other scripts being used to run the test system to ensure that they're compatible w.r.t. the basic interface.  Grepping the scripts used when running testing, this appeared to be the only remaining case that referred to a script in `$CHPL_HOME/util` rather than `$CHPL_TEST_UTIL_DIR` when that makes sense.  So I think it was an outlier case that got away when we implemented this capability rather than an indication of a major slippery slope.

PR #8836, opened 2018-03-15
Comm=ugni: print usual out-of-mem message for hugepage fault-in SIGBUS.

### LINKED ISSUE #8801: Say "out of memory" when SIGBUS occurs during hugepage fault-in
### Task

With `CHPL_COMM=ugni` we set the environment variable `HUGETLB_NO_RESERVE`  so that NIC-registered hugepages get NUMA locality corresponding to the membind settings of the threads that first touch them, that is, the threads that fault them in.  An unfortunate side effect of this is that if we cannot allocate a hugepage at fault-in time we get a `SIGBUS` signal, because that env var also setting prevents checking availability at allocation time.  A lack of available pages at allocation time would lead to a simple NULL return from `get_huge_pages()` and then we would issue the usual "Out of memory" message and a halt.  The `SIGBUS` kills the program without a useful message and may result in a (potentially very large) core dump.

The task here is to catch `SIGBUS` signals resulting from hugepage allocation failures during fault-in and issue the usual "Out of memory" message in that case.  But only signals due to fault-in should get this response; to the extent possible, `SIGBUS` signals due to other causes should continue killing the program as they do today.

#8675 was a misguided attempt to solve the same problem in a different way.  #6949 is an umbrella issue covering this and other topics having to do with the dynamic registration work.

gbtitus:
With comm=ugni we set the environment variable `HUGETLB_NO_RESERVE=yes` in
order to get hugepage NUMA locality matching the membind settings of the
threads that fault in those pages on first touch, rather than that of
the thread that allocates them.  This applies both to memory for large
arrays and to mem=jemalloc heap extension chunks with a dynamically
extended heap.  A side effect of this setting is that when hugepages are
allocated we do not acquire or even reserve backing physical pages.
Instead we get those those later, lazily, as the corresponding virtual
hugepages are touched.  If suitable memory cannot be found at that time
the process gets a `SIGBUS` signal.  Up until now such signals have killed
the program in an uncontrolled way.

Here we arrange instead to catch these `SIGBUS` signals and print "Out of
memory allocating \<description\>" and exit, just as would have happened
if the allocation itself had failed.  A couple of new tests check that
this works.

While here, I also fixed a off-by-1 error in counting free memory region
table entries.

This fixes #8801.


gbtitus:
Stand by on reviewing this one -- I need to rebase in order to get clean testing.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2929,6 +2980,26 @@ void set_hugepage_info(void)
...
+  {
+    char const *ev;
+
+    if ((ev = getenv("HUGETLB_NO_RESERVE")) != NULL
+        && strcasecmp(ev, "yes") == 0) {
+      struct sigaction act;
+
+      act.sa_flags = SA_RESTART | SA_SIGINFO;
+      act.sa_sigaction = SIGBUS_handler;
+      if (sigaction(SIGBUS, &act, &previous_SIGBUS_sigact) != 0) {
```
dmk42: Since `act` is an automatic variable, the signal mask might contain random bits.  You'll need something like this at the same time that you're initializing the other fields.
```c
sigemptyset(&act.sa_mask);
```

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2938,6 +3009,61 @@ void set_hugepage_info(void)
...
+    }
+  }
+
+  //
+  // This was not due to allocation failure on a hugepage we were trying
+  // to fault in.  Restore the previous handler and re-raise the SIGBUS.
+  // It should get dealt with as it would have without our handler.
+  //
+  (void) sigaction(SIGBUS, &previous_SIGBUS_sigact, NULL);
+  raise(SIGBUS);
```
cassella: Can any ``previous_SIGBUS_sigact`` result in handling the SIGBUS and resuming without the signal having been fatal?  If so, this would leave the thread running without this handler.

test/runtime/gbt/comm/ugni/SIGBUS-array-alloc.good
```diff
@@ -0,0 +1 @@
+SIGBUS-array-alloc.chpl:15: error: Out of memory allocating "array elements"
```
cassella: Do you want the message to call attention to it being a hugepage allocation failure specifically?  (If so, is the particular hugepage size of the region available in the table the signal handler is consulting?)

runtime/src/comm/ugni/comm-ugni.c
dmk42: Good point.  Currently nothing installs a SIGBUS handler that can keep going, but this does impose a requirement on future code not to do that.

One way around this is to do what `runtime/src/chplsys.c:check_page_size_segv_handler()` does and just run the old handler directly, without letting go of ownership of the signal.

runtime/src/comm/ugni/comm-ugni.c
gbtitus: Ugh, how did I miss that?  Will fix ...

test/runtime/gbt/comm/ugni/SIGBUS-array-alloc.good
gbtitus: No, it should look just like any other allocation failure.  Whether or not a particular allocation was satisfied out of hugepage or non-hugepage memory is not something we want Chapel programmers or programs to care about.

runtime/src/comm/ugni/comm-ugni.c
gbtitus: As I read the man pages the only safe way to invoke the old handler (if any) is to reinstall it and reraise the signal.  The relevant paragraph in the `sigaction(3p)` man page is this:

    If the previous action for sig had been established by signal(), the
    values of the fields returned in the structure pointed to by oact are
    unspecified, and in particular oact->sa_handler is not necessarily the
    same value passed to signal().  However, if a pointer to the same
    structure or a copy thereof is passed to a subsequent call to sigac-
    tion() via the act argument, handling of the signal shall be as if the
    original call to signal() were repeated.

So without knowing for sure that the previous handler was installed using `sigaction()`, we can't call the handler it installed directly.  Does this match your understanding?

dmk42: Ah, I didn't realize that.  That quote from the man page was lifted right out of the POSIX standard, so I should have known.

Fortunately, `raise()` does not return until the other handler returns, so you can use `sigaction()` to reestablish control over the signal afterward, if the other handler doesn't halt the program.

gbtitus: Ah, and I missed that detail.  I'll reinstall the new handler if/after `raise()` returns.  Thanks!

gbtitus: P.S. We don't expect you to have the entire POSIX standard memorized.   :-)

gbtitus:
@dmk42, @ronawho, @cassella: Full comm=gasnet testing passed on linux64 cluster (to check for any problem with the chpl-mem.h and associated changes).  XC examples testing had passed before but I'm running it again to make sure the most recent commits haven't destabilized something.  I'll also test on XE.  OK to merge assuming XC and XE pass?

dmk42:
:+1:

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2941,6 +3007,82 @@ void set_hugepage_info(void)
 }
 
 
+static
+void install_SIGBUS_handler(void)
+{
+  struct sigaction act;
+
+  act.sa_flags = SA_RESTART | SA_SIGINFO;
+  if (sigemptyset(&act.sa_mask) != 0) {
```
cassella: Hmm.  My copy of ``sigaction(2)`` says after describing ``sa_mask``: "In addition, the signal which triggered the handler will be blocked, unless teh ``SA_NODEFER`` flag is used."  So I think with this setup, ``SIGBUS`` will be blocked in the handler, and I don't think the ``raise()`` will have an effect until return from the (intended-to-be-outer) handler.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2941,6 +3007,82 @@ void set_hugepage_info(void)
...
+  // to fault in.  Restore the previous handler and re-raise the SIGBUS.
+  // It should get dealt with as it would have without our handler.  If
+  // the old handler returns and execution continues, install our new
+  // handler again.
+  //
+  if (sigaction(SIGBUS, &previous_SIGBUS_sigact, NULL) != 0) {
+    CHPL_INTERNAL_ERROR("sigaction(SIGBUS) to reinstall old handler failed");
+  }
+  raise(SIGBUS);
+  install_SIGBUS_handler();
```
cassella: Does it matter what happens if different threads are taking ``SIGBUS`` at the same time, some from hugepage fault-ins, and some not?  ``sigaction()`` sets the behavior for the whole process, but more than one thread can be in the handler at the same time.  (Or in different handlers if the signal arrives with a different ``sigaction`` state.)  Maybe it doesn't matter as long as at least one of the handlers will do something fatal.

gbtitus: Hmm indeed.  The solution would be to do `pthread_sigmask(SIG_UNBLOCK, <mask_with_just_SIGBUS>, NULL)` before the `raise()` call, right?  (And maybe restore it afterward, just for completeness.)  Without `SA_NODEFER`, while we're handling a `SIGBUS` due to fault-in that signal will be blocked in the process, if we can believe the `sigaction(3p)` man page.  So we won't get another `SIGBUS` for a hugepage fault-in in this thread while we have the previous handler re-installed in place of our new one.  Then we're safe just changing the sigmask for our thread and trying the old handler, then putting the sigmask and new handler back if the `raise()` returns.

gbtitus:
@cassella, how does the unblock(SIGBUS) code look to you?

PR #8842, opened 2018-03-16
Quiescent-State Based Reclamation - Overhaul [ W.I.P ]

LouisJenkinsCS:
Reverts and makes changes to: #8841 by @mppf
Original PR: #8182

I introduce Quiescent State-Based Reclamation, a memory reclamation algorithm that can be from the confines of the runtime and potentially from Chapel user code (with some GOTCHAS). The memory reclamation algorithm comes with very little performance regression and can ensure the eventual cleanup of memory so long as checkpoints are periodically called from all threads (although their placement is up to debate). 

QSBR can come in handy for any performance-critical data structure, and currently is used in Chapel's privatization table. Future uses can be for the thread-safety of Chapel's callback system and perhaps in more dynamic task local storage. In the future, a specific QSBR table is planned for users that will not be able to interfere with the runtime.

**Potential Uses of QSBR:**
1. Non-Blocking Data Structures: It is safe to defer the deletion of anything logically removed from the data structure in question. For example, a lock-free queue that removes a node from the head can safely defer deletion as it will be the only thread to do so (since the CAS operation will return true for only one successful operation)
2. Multiword Compare-And-Swap: This also applies to anything using descriptors for helping mechanisms, but a K-Word Compare-And-Swap where we make use of an object that holds enough state to represent some in-progress operation to claim ownership of the location. The descriptor can be accessed by concurrent threads and so it is safe to defer for deletion after the owning thread/task completes its operation.
3. Software Transactional Memory: While things like reading and writing are trivial to implement, a huge problem that arises is the deletion of data during a transaction and even after committing it. Deleting the object immediately can result in a segmentation fault for tasks that have not yet aborted their transaction, so deferring deletion would be a solution so long as no checkpoint is invoked within a transaction.

There are many other potential uses, the sky is the limit.


mppf:
### What needs to be addressed before merging QSBR again

 - [ ] chpl-qsbr.c needs to compile with CHPL_ATOMICS=cstdlib and CHPL_ATOMICS=locks. With CHPL_ATOMICS=locks, an atomic uint_ptr_t is larger than a uint_ptr_t (since it includes a lock) and so there shouldn't be pointer casting from the atomic type to the value type. Additionally all reads of atomic values should be through atomic functions e.g. atomic_load_....
 - [ ] valgrind testing needs to come back clean (e.g. `source util/setchplenv/quickstart.bash` `start_test -valgrindexe <tests>`)
 - [ ] resolve core dump for `make check` on Mac OS X with GASNet
 - [ ] Give David K and Brad a chance to comment (I didn't @-mention them here yet intentionally)

### Testing to do before attempting to merge it again
 - [ ] make check works with PGI compiler
 - [ ] make check works in Mac OS X with GASNet segment=everything
 - [ ] make check works in Mac OS X with GASNet segment=fast
 - [ ] make check works in Mac OS X with CHPL_ATOMICS=cstdlib
 - [ ] make WARNINGS=1 does not have errors with CHPL_COMM=ugni
 - [ ] hellos compile and run on a Cray XC with intel compiler
 - [ ] memory/qsbr/ tests pass on Cygwin 32-bit
 - [ ] domains/sungeun/assoc/stress.numthr  passes on Cygwin 64-bit
 - [ ] full valgrind testing

PR #8849, opened 2018-03-18
Replace confusing legacy LLVM build error

dmk42:
Now that LLVM no longer supports building without cmake, a missing cmake results in a large and confusing dump of information.  This change replaces it with a simple error message.

While there, I ripped out the remaining support for configure, which is already gone from LLVM.

Tested on Mac and Linux, with `CHPL_LLVM=llvm` and `CHPL_LLVM=system`.

### DIFF
--- a/third-party/llvm/Makefile
+++ b/third-party/llvm/Makefile
@@ -7,16 +7,13 @@ include $(CHPL_MAKE_HOME)/make/Makefile.base
 
 # decide whether to do a debug or no-debug build
 ifdef CHPL_LLVM_DEVELOPER
-CHPL_LLVM_CONFIGURE_DEBUG = --enable-debug-runtime --enable-debug-symbols --enable-assertions
 CHPL_LLVM_DEBUG = -DCMAKE_BUILD_TYPE=Debug -DLLVM_OPTIMIZED_TABLEGEN=1
 else
-CHPL_LLVM_CONFIGURE_DEBUG = --disable-assertions --enable-optimized
 CHPL_LLVM_DEBUG = -DCMAKE_BUILD_TYPE=Release
 endif
 
 LLVM_SRC_FILE := llvm/CMakeLists.txt
 LLVM_CMAKE_FILE := $(LLVM_BUILD_DIR)/CMakeCache.txt
-LLVM_CONFIGURE_FILE := $(LLVM_BUILD_DIR)/config.status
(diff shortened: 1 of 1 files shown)


bradcray:
I haven't tried this, but it looks like a nice clean-up to me.

dmk42:
@mppf - I've ripped out support for building LLVM with configure, which is something you've been wanting.  This works for me, but would you please check to make sure I didn't rip out too much?

third-party/llvm/Makefile
```diff
@@ -73,55 +67,25 @@ $(LLVM_CONFIGURED_HEADER_FILE): $(LLVM_SRC_FILE)
...
+	cd $(LLVM_BUILD_DIR) && cmake --build . --target install-llvm-headers
 
 # note: the conditional below allows parallel make to continue
 # if cmake us also using Make.
 $(LLVM_SUPPORT_FILE):
-	if [ -f $(LLVM_CONFIGURE_FILE) ]; then \
-	  cd $(LLVM_BUILD_DIR) && $(MAKE) install-cmake-exports ; \
-	  cd $(LLVM_BUILD_DIR)/lib/Support && $(MAKE) ; \
-	  cd $(LLVM_BUILD_DIR)/lib/Support && $(MAKE) install ; \
-	elif [ -f $(LLVM_BUILD_DIR)/Makefile ]; then \
```
mppf: AFAIK we want this 2nd element of the conditinoal to allow parallel make (make -j 8 say) to work for building LLVMSupport but you might have a different solution to that problem.

third-party/llvm/Makefile
```diff
@@ -41,9 +38,6 @@ clobber: FORCE
 
 # -enable-debug-runtime --enable-debug-symbols creates huge binaries
 #  so we by default do not include LLVM debugging information.
-# Note that the configured-clang-sysroot-arguments could possibly be added
-# to LLVM as configure options (--with-clang-resource-dir,
-# --with-c-include-dirs, --with-default-sysroot)
```
mppf: I think this comment is worth keeping but maybe just change it to "as cmake options."

third-party/llvm/Makefile
dmk42: Thanks.  That was an oversight.

third-party/llvm/Makefile
dmk42: Sounds good.

PR #8853, opened 2018-03-19
Add versions of channel.mark, commit, revert for locking=false channels

bradcray:
For channels for which locking==false, this PR adds versions of the existing
_mark(), _commit(), and _revert() methods, yet without underscores for
clarity.  In the long-run, we should probably add no-underscore versions for
the case with manual locking as well, but that felt like too big of a change to
take on for this release.

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -2251,22 +2251,6 @@ proc channel.advancePastByte(byte:uint(8)) throws {
 }
 
 
-// These begin with an _ to indicated that
-// you should have a lock before you use these... there is probably
-// a better name for them...
-
-/*
-   For a channel locked with :proc:`channel.lock`, return the offset
-   of that channel.
- */
-inline proc channel._offset():int(64) {
-  var ret:int(64);
-  on this.home {
-    ret = qio_channel_offset_unlocked(_channel_internal);
-  }
-  return ret;
-}
-
 /*
    *mark* a channel - that is, save the current offset of the channel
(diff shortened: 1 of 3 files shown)


bradcray:
@mppf: Would you review this and make sure it matches what you proposed / were thinking?  The diff is pretty ugly on GitHub, but essentially the main changes here are:

* clone the routines to create non-underscore versions with where clauses
* modify the documentation such that the no-underscore versions are more primary and the underscore versions defer to the no-underscore versions.
* put the no-underscore versions just before the section with underscore routines starts.

bradcray:
Oops, I also meant to tag @benharsh due to his role in revcomp versions.

bradcray:
Thanks @mppf!  

I was curious about something philosophical while I was thinking about this PR on the way to work this morning:  What's the advantage of using a mark/commit/revert discipline for this type of pattern rather than using some sort of getOffset() / setOffset() routines where the callee would be responsible for storing the offsets rather than relying on a stack of them within the channel itself?  It seemed to me that it could potentially simplify the locking discipline by putting the management of the state in the user's hands rather than the channels, but maybe I'm missing something.

mppf:
@bradcray -`mark` actually causes changes in what the channel stores - data that is between a `mark` and the boundary of the buffer won't be discarded from the buffer since it might still be modified (with `revert).

That's why it'd have to be something like "getTransaction" or "getMarkedChannel" etc.

mppf:
I agree it should be throwing. I'm fine either way on returning the current offset.

PR #8854, opened 2018-03-19
Fix LLVM tests for gasnet-fast

dmk42:
This change fixes some LLVM-specific tests to either work or skip when CHPL_COMM!=none, as appropriate.


dmk42:
@mppf - Would you mind giving this a post-commit look?

PR #8856, opened 2018-03-19
Make write/readBytes Chapeltastic and symmetrical

bradcray:
This converts the throwing (public) version of readBytes() into a
routine intended for end users by documenting it and reducing the
reliance on C types/casts. It also updates the revcomp study
versions to make use of this new interface.

As part of this PR, I also added the deprecating warnings for
the versions of readBytes() hat take an `out` error code.

The writeBytes routines are similarly updated.


bradcray:
@benharsh , @mppf : This is a cleanup of readBytes/writeBytes motivated by the revcomp example we've been discussing if one of you would have time to review.

@psahabu : While here, I deprecated the `out` argument versions of the routines based on the pattern you showed me.  Let me know if this looks OK.  Also, note the `halt()` that I didn't rewrite
and let me know if you have any thoughts on it (I added a TODO comment to it).

modules/standard/IO.chpl
```diff
@@ -3300,27 +3292,37 @@ inline proc channel.readwrite(ref x) where !this.writing {
...
+    var retval = true;
     on this.home {
       try! this.lock();
-      error = qio_channel_write_amt(false, _channel_internal, x, len);
+      err = qio_channel_write_amt(false, _channel_internal, c_ptrTo(val),
+                                  numBytes.safeCast(ssize_t));
       this.unlock();
+      if err {
+        try this._ch_ioerror(err, "in channel.writeBytes");
+        retval = false;
```
psahabu: `retval = false` will automatically be inserted by the compiler, so I don't think this line is needed.

psahabu:
The conversion looks good to me, thanks for tackling it while you were here.

modules/standard/IO.chpl
bradcray: I don't see what would make the compiler insert `retval = false`... can you explain?

psahabu: Ah, I misspoke. When an error is thrown, the `lowerErrorHandling` pass sets the return value to a default value in the function epilogue. In this case it'll set the compiler-inserted return value to `false` and ignore `retval` entirely.

As such, it isn't necessary to set `retval = false`.

mppf: it's basically dead code, like this
``` chapel
proc f() throws {
  var x = 12;
  throw new MyError();
  x = 22; // this is dead code
  return x; // this is also dead code
```

The call site of a call to `f()` doesn't get a chance to examine the return value, since the function didn't return normally (it threw instead).

bradcray: OK, that suggests to me that this function shouldn't really have a return value anymore (and shouldn't ever have) -- does that look right?

psahabu: Oh, yeah I don't think this needs to return a `bool`. The `read` functions do because `false` indicates `EOF`.

bradcray:
Parking this until after the release (at least) in order to address Michael's feedback.

bradcray:
I haven't had the motivation to look at this PR since the need for it has passed.  I think off-line from this PR, @mppf expressed reservations that this was an interface we'd want to promote, which led to my lack of motivation.  Closing it until someone champions it again.

PR #8859, opened 2018-03-19
Add --warn-unstable developer flag, require it for return iterator error

mppf:
As a  follow-on to PR #8073, make the warning for a returned iterator being coerced to an array only appear with `--warn-unstable` which is a developer flag for now.

The rationale here is that a warning for that case is not called for because the change in language semantics is unlikely to cause problems. On the other hand, the compiler has the capability to warn about it, and we'd like to be able to activate that if a user has a question about it.

I expect `--warn-unstable` will be useful for upcoming language changes to do with `new`.

Passed full local testing.


util/chpl-completion.bash
```diff
@@ -26,10 +26,10 @@ _chpl ()
...
...infer-local-fields --no-vectorize --no-bounds-checks --no-cast-checks --no-div-by-zero-checks --no-formal-domain-checks --no-local-checks --no-nil-checks --no-stack-checks --no-codegen --no-cpp-lines --no-munge-user-idents --no-debug --no-optimize --no-specialize --no-llvm --no-llvm-wide-opt --no-print-commands --no-print-passes --no-devel --no-explain-verbose --no-print-callgraph --no-print-callstack-on-error --no-print-unused-functions --no-permit-unhandled-module-errors --no-task-tracking --no-warn-const-loops --no-warn-special --no-warn-domain-literal --no-warn-tuple-iteration --no-warnings --no-cc-warnings --no-gen-ids --no-html-wrap-lines --no-html-print-block-ids --no-log-ids --no-verify --no-parse-only --no-debug-short-loc --no-warn-unstable --no-denormalize --no-force-initializers --no-lifetime-checking --no-ignore-errors --no-ignore-errors-for-pass --no-localize-global-consts --no-local-temp-names --no-memory-frees --no-preserve-inlined-line-numbers --no-print-id-on-error --no-print-unused-internal-functions --no-remove-empty-records --no-remove-unreachable-blocks --no-replace-array-accesses-with-ref-temps --no-incremental --no-minimal-modules --no-user-constructor-error"
```
bradcray: Oh wow, I'd never looked at this file or the PRs that introduced it.  This seems like a bit of a maintenance nightmare...  :(  I wonder what it would take to have these lists be auto-generated by processing the compiler sources or `chpl --help` output...  I could also imagine wanting to tease this into options that would be available when CHPL_DEVELOPER is set in the environment vs. not if that were possible (to avoid polluting the user's space of options and/or reduce maintenance burden).  This is more a comment on the maintainability and architecture of this file than on this specific PR, so I'm tagging @daviditen on it as food for thought.

bradcray:
I think that in addition to the changes here, we could/should also remove the following three compopts files that were introduced in PR #8073 (I tried this on a copy of your branch and it seemed to work, but you should double-check me):

* test/functions/deitz/iterators/temps/test_iterator_temp4.compopts
* test/functions/deitz/iterators/test_misc.compopts
* test/functions/vass/passing-iterator-as-argument.compopts

Other than that, this looks good to me.  Thanks for taking it on and sorry not to have noticed it / thought of it earlier in the release cycle.

daviditen:
For `chapel-completion.bash` it's not quite as much of a nightmare as you're probably imagining.  If it gets out of date, the test `test/chplenv/autocomplete/autocomplete.chpl` will fail with the message:

```
diff failed. You may need to run $CHPL_HOME/util/devel/gen-chpl-bash-completion  to regenerate $CHPL_HOME/util/chpl-completion.bash. Try:
/util/devel/gen-chpl-bash-completion > $CHPL_HOME/util/chpl-completion.bash
```

Running the script with the output redirected as suggested will re-generate it for the current options.  That said, I agree it would be good to have the completion script generate the flags itself from `chpl --help` and to differentiate between developer and non-developer modes.

bradcray:
Thanks David, that's reassuring.  Given that sanity check, I'd prioritize differentiating between developer and non-developer flags sooner than auto-generating the flags for the script (in that it seems of greater value to the common user not to suggest flags they're not supposed to be using as autocomplete options).

PR #8861, opened 2018-03-19
Fix segfault with inheritance and initializers/constructors

lydia-duncan:
Prior to this change, if a parent class relied on the default initializer and a child
of that parent defined an explicit constructor, we would encounter a segfault
during compilation instead of the nice error message about mixing your object
initialization styles.  This change resolves that error.

It turns out the check for if the parent utilizes an initializer (either that it
defines or the compiler generated one) was only looking for if an explicit
initializer was in the hierarchy.  Instead, use the check to determine whether a
default initializer should be generated for the type and remove the
no-longer-used parentDefinesInitializer

This resolves two tests that were failing prior to this change, which are now
added to the code base.  One has the parent type get a default initializer
via `--force-initializers`, the other via the pragma.  Both failed on master
and pass with this change.

Testing:
- [x] full std/ paratest


lydia-duncan:
Works, but doesn't seem to help a larger hierarchy.

lydia-duncan:
Replacing with #8862

PR #8862, opened 2018-03-19
Fix segfault with inheritance and initializers/constructors, attempt 2

lydia-duncan:
Prior to this change, if a parent class relied on the default initializer and a child
of that parent defined an explicit constructor, we would encounter a segfault
during compilation instead of the nice error message about mixing your object
initialization styles. This change resolves that error.

It turns out the check for if the parent utilizes an initializer (either that it
defines or the compiler generated one) was only looking for if an explicit
initializer was in the hierarchy.  Expand the check to also error when a default
initializer is present above the class with a constructor.

Add three tests validating this behavior.  One has the parent type get a default
initializer via `--force-initializers`, one has the parent type obtain it via the
pragma, and the third examines a three class hierarchy where the child defines
an explicit constructor and the grandparent uses the pragma (and the parent
relies on the default constructor).

Testing:
- [x] full std/ paratest

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -1415,6 +1415,7 @@ void AggregateType::buildConstructor() {
           // defines an initializer - we should not be creating a default
           // constructor in that case, it won't know what to do with it.
           if (at->initializerStyle == DEFINES_INITIALIZER ||
+              at->wantsDefaultInitializer() ||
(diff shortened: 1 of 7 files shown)


lydia-duncan:
Replaces #8861

lydia-duncan:
@benharsh - do you have time to review this today?

PR #8869, opened 2018-03-19
Make LLVM 6 the default version of LLVM

dmk42:
This change upgrades to LLVM 6.

Heads up:
* If you are working with LLVM, you will probably need to do a `make clobber` after this PR goes in.
* If you are using `CHPL_LLVM=system` on a Mac, you will probably need to do a `brew upgrade` to pull in the latest version of LLVM.

A few remaining commented-out build artifacts supporting LLVM 3.x are removed here, since support for those versions was already removed from our code some time ago.

Tested on both x86-64 and ARM Cray systems, as well as Mac and Linux.
Tested with `CHPL_LLVM=llvm` and `CHPL_LLVM=system`.


dmk42:
@mppf - Would you please take a post-commit look?

PR #8870, opened 2018-03-19
Add a prediff to handle uninitialized mason installation

ben-albrecht:
If mason was not initialized, message would print `initializing` instead of `updating` resulting in a failure. 

- Modified `PREDIFF` to filter that out


ben-albrecht:
Also, other tests don't happen to hit this, because of their `EXECENV`. However, they could be updated to use this `PREDIFF` string replacement as well.

thomasvandoren:
Thanks @ben-albrecht

PR #8872, opened 2018-03-20
Switch revcomp-blc from readBytes() to read() of an array slice

bradcray:
In code reviewing my readBytes() clean-up with Michael, he pointed out
that binary I/O on an array slice felt like a much more natural way to
write this to him.  I think we must've shied away from it in the past
because our line-by-line reading options tanked performance wise due
to all the slicing.  But for this case, we're reading big enough
chunks that the slice performance seems to be in the noise.  This also
feels more Chapeltastic than readBytes() for this situation.

### DIFF
--- a/test/studies/shootout/reverse-complement/bradc/revcomp-blc.chpl
+++ b/test/studies/shootout/reverse-complement/bradc/revcomp-blc.chpl
@@ -43,8 +43,7 @@ proc main(args: [] string) {
         input.revert();
 
         // Read until nextDescOffset into the data array.
-        input.readBytes(c_ptrTo(data[descOffset]),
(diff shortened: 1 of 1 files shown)


bradcray:
@benharsh and @mppf: Turns out reading a slice works correctly (Michael, I went back to see what you said that made me think it might not and can't tell what "DR" was supposed to stand for) and doesn't have a big performance impact: the average timing on chapcs07 was only 0.2% greater than the previous version, and while the max was a bit bigger, the min and median timings were smaller.  So it seems to mostly be in the noise for this version given the small number of slices and large amount of data being processed.

benharsh:
DR = DefaultRectangular?

PR #8875, opened 2018-03-20
integration with clang >= 6 uses only public API

mppf:
This PR adjusts our integration with clang so that it
only uses public clang API functions for clang >=6.

For earlier clang versions, we needed to use internal
clang headers in order to perform two operations.

- [x] full local testing

Reviewed by @dmk42 - thanks!


mppf:
@dmk42 - could you review and also share your opinion of if it should go in for the release?

PR #8878, opened 2018-03-20
Fix --llvm-wide-opt after PR #8194

mppf:
PR #8194 changed codegenFieldPtr in a way that
inadverdently broke --llvm-wide-opt.

In --llvm-wide-opt mode, wide pointers are represented
as address-space-100 pointers in the LLVM IR, so they
are still subject to GEP / load / etc.

codegenFieldPtr was using Type* castType for two purposes:
 1. to compute the GEP offset to use
 2. to cast the type to.

For --llvm-wide-opt mode, the first use needed to
use a "wide pointer" so it would code generate to
an address-space-100 pointer. But the second use
needed the non-wide pointer, since the wide pointer
doesn't actually contain the field.

This commit addresses this problem by moving some
fLLVMWideOpt conditionals later in the code. The
change to when ret.isLVPtr is set seems more correct
since base might have be modified after this condition
in the previous version.

- [x] test/performance/ferguson/ with CHPL_LLVM=llvm and Gasnet
        (this configuration failed before)
- [x] test/release/examples with Gasnet and --llvm-wide-opt --llvm --fast
- [x] full local testing

Reviewed by @dmk42 - thanks!

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -1007,14 +1007,7 @@ GenRet codegenFieldPtr(
     ret.chplType = fieldSymbol->type;
   }
 
-  if( fLLVMWideOpt && castType && isWide(base) ) {
-    // for fLLVMWideOpt
-    castType = getOrMakeWideTypeDuringCodegen(castType);
-  }
-
   ret.isLVPtr = GEN_PTR;
-  // with LLVMWideOpt, we might return a wide ptr.
-  if( fLLVMWideOpt && isWide(base) ) ret.isLVPtr = GEN_WIDE_PTR;
 
   if (isClass(ct) ) {
     base = codegenValue(base);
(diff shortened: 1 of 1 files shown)


mppf:
@dmk42 - could you have a look at this fix for --llvm-wide-opt ?

dmk42:
This is a fix of a bug that has been causing problems, and it belongs in the release.

bradcray:
Is there some testing we ought to have that would catch backslides like this closer to when they went wrong?  (e.g., a compiler diagnostic that could print whether or not the optimization fired for some key or simple cases?)

dmk42:
Yes, the testing is gasnet-fast.  That would have caught it right away.  I didn't realize I was affecting wide pointers, so I didn't try that at the time, and we don't have a gasnet-fast nightly test that throws everything at LLVM.

mppf:
To be clear, I think I'd call the testing "--llvm --fast" with GASNet. I don't think the segment (fast vs everything) matters.

Because test/performance/ferguson/ throws --llvm --llvm-wide-opt --fast, it'd be sufficient to test "--llvm" with any gasnet configuration to get basic coverage (as far as I know, anyway).

PR #8882, opened 2018-03-20
Remove the deprecated Barrier module

daviditen:
Remove the deprecated Barrier module and the test that checks for the deprecated
warning for using it.


daviditen:
Thank you!

PR #8883, opened 2018-03-20
Repair the runtime unit-testing side of error.h after #8836.

gbtitus:
There is limited support for a "standalone" build of the runtime to
support certain kinds of unit testing.  When I did the work in #8836 I
threw together a standalone version of `chpl_error_vs()`, intending to
come back and correct it later before merging, but I never returned to
finish up.  I also never tested it, which would have demonstrated that I
hadn't done so.  Complete and test it now.

### DIFF
--- a/runtime/include/error.h
+++ b/runtime/include/error.h
@@ -67,19 +67,23 @@ void chpl_internal_error(const char* message);
   } while (0)
 
 static inline
+void chpl_error_vs(char *restrict, size_t, int32_t, int32_t,
+                   const char *restrict, ...)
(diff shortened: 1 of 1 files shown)


gbtitus:
For now at least I'm leaving the instance of `__attribute__` outside of `CHPL_RT_UNIT_TEST` protection as-is, not under a target-compiler-based `#ifdef`, because there are other places in the runtime where this occurs, specifically one each in `runtime/include/atomics/intrinsics/chpl-atomics.h` and `runtime/include/chpl-string-support.h`.

PR #8887, opened 2018-03-20
Categories for CHANGES.md for 1.17

bradcray:


CHANGES.md
```diff
@@ -1,6 +1,135 @@
...
 
+version 1.17.0
+==============
+
+Twentieth public release of Chapel, April 5, 2018
+
+Highlights (see subsequent sections for further details)
+--------------------------------------------------------
+
+New Dependences / Configuration Changes
```
cassella: Do you mean "Dependences", or "Dependencies"?

bradcray: I've never understood the difference between these two worlds, and googling, I'm still not sure I do understand it, but I take it you think dependencies is more appropriate here?

cassella: I found this, which agrees with me:
https://ell.stackexchange.com/questions/41528/differences-between-dependence-and-dependency

and a few others that don't agree with me so I'm not going to tell you about them.



Well, ok, though AFAICT they don't generally come to a conclusion, here's one:

https://forum.wordreference.com/threads/dependence-vs-dependency-computers.1072451/

bradcray:
Closing in favor of #8957

Issue #4949, opened 2016-12-06
Argument parsing errors with start_test


sungeunchoi:
### Summary of Problem
I found a few issues with argument parsing:

- The `--compopts` and `--execopts` flags do not work when using `=` (a space must be used).  It appears as though the first character is truncated.
- Using shortened versions of the above flags, e.g., `--compo` or `--execo`, with a space and an argument, results in a start_test usage: error:
```
start_test: error: argument -compopts/--compopts: expected one argument
```
- Certain styles of invalid flags results in the flags being interpreted as test names (e.g., `--compt="--fast --static"` does so, but not `--compt="--fast"` (correctly returns usage message), `--numx="2 x"` but not `--numx=2`)

### Steps to Reproduce

**Source Code:**

N/A

**Compile command:**

N/A

**Execution command:**

N/A

### Configuration Information

- Output of `chpl --version`:
chpl Version 1.14.0.a10da81
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
N/A
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
N/A
- (For Cray systems only) Output of `module list`:
N/A

ben-albrecht:
> The --compopts and --execopts flags do not work when using = (a space must be used). It appears as though the first character is truncated.

This is caused by lines [633](https://github.com/chapel-lang/chapel/blob/ad6e21921762629950c97c4a914fdd22730577d8/util/start_test#L633) and [640](https://github.com/chapel-lang/chapel/blob/ad6e21921762629950c97c4a914fdd22730577d8/util/start_test#L640) of `start_test`.

Not sure what purpose this truncation serves. Will have to look into it more.

ben-albrecht:
> Using shortened versions of the above flags, e.g., --compo or --execo, with a space and an argument, results in a start_test usage ...

I reproduced this in a [toy program](https://gist.github.com/ben-albrecht/f73bf354bdea4f021e8d3bad926140a3).

Out of all the combinations I tried, I found only`--compo=--fast` works as expected.

As far as I can tell, this is an issue with Python's argparse. Our best bet might be to  recommend against using abbreviated flags (offering single character abbreviations of our own, instead maybe?) Sadly, disabling abbreviations isn't supported until Python 3.5.

ben-albrecht:
In summary:

> The --compopts and --execopts flags do not work when using = (a space must be used). It appears as though the first character is truncated.

[...]

PR #8906, opened 2018-03-21
Fix handling of compopts/execopts in start_test

ronawho:
Fix how we preprocess and fixup --compopts= and --execopts= arguments to
start_test. We convert things like `--compopts --fast` to `--compopts '--fast'`
so that argparse doesn't interpret `--fast` as an option to start_test and we
later strip the leading and trailing characters to remove the quotes. However,
we were removing the leading and trailing characters even when we didn't add
quotes for things like `--compopts=fast`. Correct that by only stripping the
leading and trailing characters if they were quotes.

Also prevent abbreviations for `--compopts` and `--execopts`. argparse allows
abbreviations for non-ambiguous options, but abbreviations for `--compopts` and
`--execopts` breaks the preprocessing that we do. Instead of trying to handle
abbreviations, just prevent them for compopts/execopts

This resolves https://github.com/chapel-lang/chapel/issues/4949

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -632,6 +632,13 @@ def check_environment_with_args():
             .format(getpass.getuser(), tgt_platform))
 
 
+# Strip quotes that might have been added by preprocess_options()
+def strip_preprocessing(arg):
+    if arg.startswith("'") and arg.endswith("'"):
+        return arg[1:-1]
+    else:
+        return arg
+
 def set_up_environment():
     # pre-
     if args.preexec:
@@ -640,16 +647,16 @@ def set_up_environment():
     if args.prediff:
         os.environ["CHPL_SYSTEM_PREDIFF"] = args.prediff
 
-    # compopts
+    # compopts (note that we have to strip out our preprocessing)
     if args.compopts:
-        args.compopts = [x[1:-1] for x in args.compopts]
+        args.compopts = [strip_preprocessing(x) for x in args.compopts]
         args.compopts = "--cc-warnings " + " ".join(args.compopts)
     else:
         args.compopts = "--cc-warnings"
 
-    # execopts
+    # execopts (note that we have to strip out our preprocessing)
     if args.execopts:
-        args.execopts = [x[1:-1] for x in args.execopts]
+        args.execopts = [strip_preprocessing(x) for x in args.execopts]
         args.execopts = " ".join(args.execopts)
     else:
         args.execopts = ""
@@ -1416,6 +1423,10 @@ def parser_setup():
     return parser
 
 
+# argparse interprets anything that starts with a `--` as an option, so
+# something like `--compopts --fast` would get interpreted as two options. Here
+# we preprocess to turn that into `--compopts '--fast'` (note that we'll have
+# to remove the quotes later)
 def preprocess_options():
     index = 0
     for arg in sys.argv:
@@ -1423,7 +1434,6 @@ def preprocess_options():
         if arg in ["-compopts", "--compopts", "-execopts", "--execopts"]:
             escape_next_argument(sys.argv, index)
         index += 1
-
 
 def escape_next_argument(args, index):
     next = args[index + 1].strip()
--- a/util/start_test
+++ b/util/start_test
@@ -1276,6 +1276,10 @@ def parser_setup():
     # executing options
     parser.add_argument("-execopts", "--execopts", action="append", 
             dest="execopts", help="set options for executing tests")
+    # lame hack to prevent abbreviations from not getting preprocessed in
+    # preprocess_options() (makes things like --execop ambiguous)
+    parser.add_argument("-execopts ", "--execopts ", action="append",
+            dest="execopts", help=argparse.SUPPRESS)
     # compiler
     parser.add_argument("-compiler", "--compiler", action="store",
             dest="compiler", help=help_all("set alternate compiler"))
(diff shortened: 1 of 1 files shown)


util/start_test
```diff
@@ -640,16 +647,16 @@ def set_up_environment():
...
+    # compopts (note that we have to strip out or preprocessing)
     if args.compopts:
-        args.compopts = [x[1:-1] for x in args.compopts]
+        args.compopts = [strip_preprocessing(x) for x in args.compopts]
         args.compopts = "--cc-warnings " + " ".join(args.compopts)
     else:
         args.compopts = "--cc-warnings"
 
-    # execopts
+    # execopts (note that we have to strip out or preprocessing)
```
cassella: *"strip out our"?

(and also for compopts above)

util/start_test
```diff
@@ -1424,7 +1443,6 @@ def preprocess_options():
             escape_next_argument(sys.argv, index)
```
cassella: Another option, instead of escaping/unescaping, would be to do your own arg parsing here -- pull ``--compopts --foo`` and ``--compopts --bar`` out of sys.argv, and keep ``compopts = [ "--foo", "--bar" ]`` off to the side to combine with the results of any ``--compopts=--baz`` the regular argparse handles.


Escaping/unescaping is a bit confusing, and means if someone did do the correct escaping to get the string ``--compopts='--foo'`` passed into start_test, you'll end up un-escaping the the string the user "meant" to provide.  (Though this is silly.)


An aside: what's the ``arg.replace()`` doing here?  It's worried about a string that's e.g. ``--compopts`` but also contains quotes?

util/start_test
```diff
@@ -1424,7 +1443,6 @@ def preprocess_options():
...
         index += 1
 
-
 def escape_next_argument(args, index):
     next = args[index + 1].strip()
```
cassella: ``start_test --compopts`` raises an unhandled python ``list index out of range`` error here.

ronawho: That's not a new issue, but good catch -- I'll lump the fix in for that into this PR too

util/start_test
ronawho: We could, though I'm not sure that will be any more robust (or less prone to issues) than our current mechanism (and this has been an issue for years without anyone complaining too loudly) 

The downside to doing our own parsing is that we lose all the functionality that argparse already provides (--help output, tab-completion support, error handling, and other things). And note that we can't just pull out `--compopts`/`--execopts`, we also have to pull out the next arg and handle the `--compopts=` case. 

We will currently pull out quotes that we didn't add for something like `--compopts='--fast'`, but at least for the cases that come to mind, that's not an issue. i.e. `--compopts='--fast --fast'` will still work since argparse has already parsed it and we're interpreting it as a single arg. If there are cases where that can hurt us, maybe we could build up a dict of which args we changed instead of always removing ones that start with quotes or something.

`arg.replace()` comes from https://github.com/chapel-lang/chapel/pull/2374. I don't remember the exact details off the top of my head, but without that change `$CHPL_HOME/util/paratest.server -compopts <option>` had `<option>` getting truncated (and note we're not replacing the value in sys.argv, just a local temp copy)

cassella: Hmm.  I can't figure out how removing quotes from the ``arg`` there (that should correspond to the word ``--compopts``, not the comp opt being passed) could have helped with that.  Though paratest.server is a bit complicated, and adds lots of quotes to things.

>  i.e. something like '-compopts --no-checks' would be seen by start_test as "--no-check"

I wonder if the "something like" was actually "-no-check", because it was hitting the unescaping-something-that's-not-escaped bug you're fixing here.

cassella: I did mean just those four things and their next args, leaving the rest of the command line for the usual argparse handling.

I don't think you'd have to handle the ``--compopts=--baz`` case, since argparse will handle that without any escaping needed.  You would need to combine the opts you collected with the ones argparse collected though.

It wouldn't handle ``--compopts= --foo`` (with a space after the ``=``) (If that's supposed to be allowed), but neither will the current code or the code after this PR.


Testing that was a little confusing, because one of the first tests it tries to run is ``Samples/Performance/foo``.  But that's unrelated to the ``--foo``.  :laughing:

cassella: Though the separate set of compopts "off to the side" would mean losing the order between ``--compopts --foo --compopts=--bar --compopts --baz``

cassella: Oh well.  I think the case I'm talking about -- where you'd still end up unescaping something you didn't escape to start with, because it already has ``'``s -- is really silly and not worth worrying about anyway.

ronawho: > Hmm. I can't figure out how removing quotes from the arg there (that should correspond to the word --compopts, not the comp opt being passed) could have helped with that. Though paratest.server is a bit complicated, and adds lots of quotes to things.

>>i.e. something like '-compopts --no-checks' would be seen by start_test as "--no-check"

> I wonder if the "something like" was actually "-no-check", because it was hitting the unescaping-something-that's-not-escaped bug you're fixing here.

Yea.. I can't quite figure it out either, but I just used the "not sure off the top of my head" excuse ;) Probably worth digging into a little more, but I'll punt until post-release when I'm thinking about merging this to dig any further.

ronawho: And if you can think of a compelling case where our escaping/unescaping really breaks things I'd be interested, but beyond that I really, really don't want to do custom argument processing. argparse offers a ton of functionality, and I don't want to try to replicate any of that even if it's just for 2 options.

util/start_test
ronawho: fixed

util/start_test
cassella: Nah, I just saw a loose thread and was pulling on it.  I agree that nothing it's led to is worth pursuing.

PR #8908, opened 2018-03-21
Document Bulk-Transfer Interface

benharsh:
This PR updates the DSI technote to describe the bulk-transfer interface defined in #5310.

### DIFF
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -1067,6 +1067,7 @@ occur:
 * The element types of the arrays must be identical
 
 * The element type must be one of the following:
+
     * ``integral``
     * ``real``
     * ``complex``
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -962,7 +962,7 @@ assignment, in practice the values are assigned individually:
 
  .. code-block:: chapel
 
-  for (d, s) in zip(dest, src) {
+  for[all] (d, s) in zip(dest, src) {
     d = s;
   }
 
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -1004,7 +1004,8 @@ Each method shares these arguments:
 
 * ``otherDom:domain`` - the indices to be transferred to/from ``otherClass``
 
-These arguments have the following guarantees:
+Callers of these methods must satisfy the following guarantees for these
+arguments:
 
 * ``this.rank == myDom.rank``
 
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -1062,7 +1062,7 @@ When Does a Bulk-Transfer Occur?
 There are a number of conditions that need to be true for a bulk-transfer to
 occur:
 
-* The config-param ``useBulkTransfer`` must be true
+* The config param ``useBulkTransfer`` must be true
 
 * The element types of the arrays must be identical
 
@@ -1078,6 +1078,13 @@ The element type is restricted in order to preserve copy semantics of each
 element. For example, ``integral`` types can be bit-copied and so have no
 side-effects. But a record with a copy-initializer may have side effects that
 should not be eliminated by the bulk-transfer optimization.
+
+The ``useBulkTransfer`` config param is ``true`` by default but can be disabled
+at compilation time. For example:
+
+.. code-block:: bash
+
+  chpl myProgram.chpl -suseBulkTransfer=false
 
 Transfers for Aliasing Domain Maps
 ----------------------------------
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -997,12 +997,12 @@ methods to opt-in to support for bulk-transfers:
 
 Each method shares these arguments:
 
-* ``myDom:domain`` - the indices used when transferring to/from ``this``
+* ``myDom:domain`` - the indices to be transferred to/from ``this``
 
 * ``otherClass`` - a class inheriting from ``BaseArr`` that represents the
   other array in the assignment
 
-* ``otherDom:domain`` - the indices used when transferring to/from ``otherClass``
+* ``otherDom:domain`` - the indices to be transferred to/from ``otherClass``
 
 These arguments have the following guarantees:
 
--- a/doc/rst/technotes/dsi.rst
+++ b/doc/rst/technotes/dsi.rst
@@ -954,5 +954,172 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
 Phase 4: Bulk-Transfer Interface
 =================================
 
-The bulk-transfer interface design is still in flux. Once finalized, it will
-be documented here.
+What is the bulk-transfer interface?
+------------------------------------
+
+While the Chapel language does not require a specific implementation for array
+assignment, in practice the values are assigned individually:
+
+ .. code-block:: chapel
+
+  for (d, s) in zip(dest, src) {
+    d = s;
+  }
+
+When assignment occurs between arrays on different locales, this can result in
+significant communication overhead. The bulk-transfer interface aims at
+reducing communication between locales by enabling domain maps to handle
+certain kinds of assignments themselves.
+
+How does the bulk-transfer interface work with DSI?
+---------------------------------------------------
+
+The bulk-transfer interface is an optional interface supported only on array
+descriptors. A descriptor can implement one or more of the following methods to
+indicate support for bulk-transfers:
+
+.. method:: proc GlobalArray.doiBulkTransferToKnown(myDom:domain, otherClass, otherDom:domain) : bool
+
+  Perform an array assignment from ``this`` to ``otherClass``.
+
+.. method:: proc GlobalArray.doiBulkTransferFromKnown(myDom:domain, otherClass, otherDom:domain) : bool
+
+  Perform an array assignment from ``otherClass`` to ``this``
+
+.. method:: proc GlobalArray.doiBulkTransferToAny(myDom:domain, otherClass, otherDom:domain) : bool
+
+  Perform an array assignment from ``this`` to ``otherClass``.
+
+.. method:: proc GlobalArray.doiBulkTransferFromAny(myDom:domain, otherClass, otherDom:domain) : bool
+
+  Perform an array assignment from ``otherClass`` to ``this``
+
+Each method shares these arguments:
+
+* ``myDom:domain`` - the indices used when transferring to/from ``this``
+
+* ``otherClass`` - a class inheriting from ``BaseArr`` that represents the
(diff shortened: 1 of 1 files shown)


doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+
+How does the bulk-transfer interface work with DSI?
+---------------------------------------------------
+
+The bulk-transfer interface is an optional interface supported only on array
+descriptors. The following methods are prefixed with ``doi`` for "Domain map
+Optional Interface." An array descriptor can implement one or more of these
+methods to opt-in to support for bulk-transfers:
+
+.. method:: proc GlobalArray.doiBulkTransferToKnown(myDom:domain, otherClass, otherDom:domain) : bool
```
lydia-duncan: This style does not seem to render properly.  I'm not entirely certain why not, though it might have something to do with the extra space here: `) :`.  That's my hypothesis, though.

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+
+.. method:: proc GlobalArray.doiBulkTransferFromAny(myDom:domain, otherClass, otherDom:domain) : bool
+
+  Perform an array assignment from ``otherClass`` to ``this``
+
+Each method shares these arguments:
+
+* ``myDom:domain`` - the indices used when transferring to/from ``this``
+
+* ``otherClass`` - a class inheriting from ``BaseArr`` that represents the
```
lydia-duncan: Is BaseArr documented explicitly, such that we could link to it?  I think someone reading this document would try to search for it, so it would be better to give them a link or maybe be more explicit about where they would find it.

benharsh: BaseArr is not currently documented. It is briefly introduced in an earlier section of this technote, but I don't think we can say much more until we have time to nail down more of DSI.

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+--------------------------------
+
+There are a number of conditions that need to be true for a bulk-transfer to
+occur:
+
+* The config-param ``useBulkTransfer`` must be true
+
+* The element types of the arrays must be identical
+
+* The element type must be one of the following:
```
lydia-duncan: This bolds really weirdly for me (but the previous two don't.  Maybe separate this and the indented lists with a new line?

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+where-clause that constrains the type of the ``otherClass`` argument. This
+prevents the ``Known`` method from being called with unknown types.
+
+When Does a Bulk-Transfer Occur?
+--------------------------------
+
+There are a number of conditions that need to be true for a bulk-transfer to
+occur:
+
+* The config-param ``useBulkTransfer`` must be true
```
lydia-duncan: Can you link to the documentation for this param?

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+
+* The element types of the arrays must be identical
+
+* The element type must be one of the following:
+    * ``integral``
+    * ``real``
+    * ``complex``
+    * record for which ``isPODType`` returns true
```
lydia-duncan: If possible, it'd be great to link to this as well

doc/rst/technotes/dsi.rst
benharsh: Hmm, good point about the bolding. I'll see what I can do.

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,173 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+should not be eliminated by the bulk-transfer optimization.
+
+Transfers for Aliasing Domain Maps
+----------------------------------
+
+It is valid to write a domain map that aliases another domain map's data. For
+example, a user could write their own variant of a Chapel array slice in terms
+of the Domain Map Standard Interface.
+
+When aliasing other domain maps it is recommended to only implement ``Known``
```
lydia-duncan: It'd be friendly for references to "Known" and "Any" outside of their section to link back to the section in which they were defined.

doc/rst/technotes/dsi.rst
lydia-duncan: Should we make it more vague for now, then?

lydia-duncan:
Suggestions can be followed or ignored, I think this looks great!

doc/rst/technotes/dsi.rst
benharsh: I was thinking this document could serve as the documentation for this param. Would a sentence like "the useBulkTransfer config param is true by default" suffice? It's located in ``ChapelArray``, but that doesn't seem like the right place to document it.

doc/rst/technotes/dsi.rst
benharsh: Perhaps unfortunately, inheriting from ``BaseArr`` is a requirement for domain maps at this time so I think we should mention it. This document in general is still in a fair amount of flux, so I'm comfortable leaving this line as-is.

lydia-duncan: Seems fair to me!

doc/rst/technotes/dsi.rst
benharsh: ``isPODType`` is currently undocumented. I'll see if I can rewrite this in a way that doesn't rely on a specific function.

doc/rst/technotes/dsi.rst
lydia-duncan: That seems reasonable, since this is why it was defined.  If that's the intention, maybe add an example of how to set it and/or give it its own section?  Something that makes it feel like you've gotten enough information and don't need to go looking for more.

benharsh:
I plan to merge this tomorrow in order to avoid a late-night scramble to fix an inadvertent documentation bug. Feel free to add more comments if something occurs to you.

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,185 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+arrays as a common interface.
+
+This interface recognizes these kinds of transfers because otherwise it is
+difficult to determine which distribution implements the best transfer.
+
+For example, the author of ``PopularDist`` could implement ``Any`` methods for
+better performance with unknown third-party domain maps. The author of
+``NewDist`` could implement ``Known`` methods to ``PopularDist`` due to its
+popularity. By doing so the author of ``NewDist`` has informed Chapel that it
+knows better than ``PopularDist`` and a more optimal transfer occurs.
```
cassella: Comments from the peanut gallery:

Given ``A = B``, is it defined whether the call is to ``A.transferFrom(B)`` or ``B.transferTo(A)``?
Or, how are the pair of directions meant to be used?  Having both suggests that either there are cases the compiler will generate either; or it's to allow a ``this.transferFrom(that)`` to turn around and call ``that.transferTo(this)``.  But the latter seems ripe with opportunity for unintentional recursion without guidance to implementers.  (e.g., "``From``s can call ``To``s, but not vice versa.")  Or is it only that the compiler will always try ``A.from(B)`` and if that fails, ``B.to(A)``?

doc/rst/technotes/dsi.rst
```diff
@@ -954,5 +954,185 @@ implicitly for each ``GlobalDomain`` object and so will be local in any case.
...
+
+The second kind of transfer is an ``Any`` transfer. These are transfers to/from
+domain maps whose type and internals are **not** known. These transfers rely on
+a common interface, which may involve unnecessary overhead compared to a
+``Known`` transfer. For example, it is likely that many domain maps will
+implement ``Known`` transfers to/from Chapel's default rectangular arrays. It
+is also likely that many domain maps will be implemented using default
+rectangular arrays. Knowing this, implementers of ``Any`` transfers can
+implement in terms of ``Known`` transfers to/from their own default rectangular
+arrays as a common interface.
```
cassella: I'm having a little trouble grokking this without an example.  I'm thinking the idea is that I'm writing ``FancyDist``, and array ``A`` distributed using it will have its data stored in ``this.myArray[]``, a ``DefaultRectangular``.  Then ``FancyDistArrayDescriptor.transferFromAny(myDom, otherClass, otherDom)`` can figure it's likely ``otherClass`` implements ``Known`` transfers relative to ``DefaultRectangular``, so it can call ``otherClass.transferToKnown(otherDom, this.myArray, someTransform(myDom))`` (assuming it resolves) and if it returns true, be done?

Can ``FancyDist``'s storage of the elements in ``myArray[]`` require multiple calls to ``otherClass.transferToKnown()`` on different subsets of the domains?  Can it assume that if the first call returns true, that the subsequent calls to the same ``otherClass.transferToKnown()`` (passing the same array and having domains of comparable kind as the first call) will return true as well?  That is, is it bad / a non-issue if ``otherClass.transferToKnown()`` returns false for some cases that are type-wise the same, e.g. because of alignment or the provided domain has more elements than it's prepared to deal with, or so few that it doesn't think it's worthwhile to do "in bulk", or the transfer has some allocation or trylock step that could fail?

benharsh: Yes, I think you have the right idea when it comes to how one could implement a transfer in terms of DefaultRectangular. Personally I would like to add some helper functions to aid domain map authors in writing such transfers, but I haven't discussed that with anyone and don't really know what it would look like.

The rule for calling the transferTo/From methods is that the caller must complete the transfer if the returned value is false. So if a call for a given locale's data returns false, then the domain map author should write something like:
```chpl
for (i, j) in zip(myDom, otherDom) do
  localData[i] = otherClass.dsiAccess(j);
```

I don't want to add a full tutorial on writing this kind of transfer, but I recognize that documentation I've written isn't very helpful in this area. I'll see if I can add a brief section on ``Any`` transfers.

doc/rst/technotes/dsi.rst
benharsh: The order is currently:
1) Try either ``Dest.transferFromKnown(Src)`` or ``Src.transferToKnown(Dest)``
2) Try either ``Dest.transferFromAny(Src)`` or ``Src.transferToAny(Dest)``

We currently do not define whether ``To`` or ``From`` is attempted first. The compiler will not generate any of these methods. Our internal implementation of arrays will simply attempt to resolve and then call these methods.

You make a good point about possible recursion, I'll add a warning calling out this problematic pattern.

I'll also add a note about undefined To/From preference.

Issue #6008, opened 2017-04-13
Path library - add support for expandUser


lydia-duncan:
Back in late 2014, we started undertaking the implementation of the FileSystem and Path libraries, with input from the user community ([start of thread](https://sourceforge.net/p/chapel/mailman/message/32800541/), [mail with updated proposal after feedback](https://sourceforge.net/p/chapel/mailman/message/32902710/)) Much progress has been made in getting the support onto master, but 10 functions still remain unimplemented due to the youth of our string implementation at the time.  With our string implementation in a good state, these functions should be returned to and implemented

This issue is for the function `expandUser`.

Here is the expected documentation and declaration:
```chapel
/* Returns the argument.  If the argument begins with `~` or `~<user>`, replace
   that substring with that user's home directory, otherwise returns the
   argument unchanged.

   :arg path: a file or directory, which may or may not include `~` or `~<user>`
   :type path: `string`

   :return: the argument, replacing any `~` or `~<user>` with the user's home
                directory
   :rtype: `string`
*/
proc expandUser(path: string): string;
```

lydia-duncan:
Tests will need to be written for this feature

pachchi:
Hello @lydia-duncan , I and my colleague worked on this issue and made this expandUser function.

```chapel 
proc expandUser(path: string) : string
  {
    var i: int = 1;
    var j: int = 1;

    //if string starts with '~' then returns the result string
    if path[1] == '~' 
    {
      var lenpath = path.length;

      //i goes in the path directory till where home directory is to be replaced 
      while(i <= lenpath && path(i) != '/')
      {
        i = i+1;
      }

       //take the environment variable of HOME
       var x: c_string = "HOME";
       var y: c_string;
       sys_getenv(x,y);
	
       if(sys_getenv(x,y) != 0) then
         return path.replace(path[1..i-1], y:string, -1);

       //if environment variable HOME is not set it checks the present directory by PWD
       else
       {
	 var lenz: int;
	 var count: int = 0;
	 x = "PWD";
    	 sys_getenv(x,y);
     	 var z = y:string;
    	 lenz = z.length;
      	 while(j <= leny && count < 3)
      	 {
	   if z(j) == '/' then
      	   count = count+1;
      	   j = j+1;
         }
       	 return path.replace(path[1..i-1], z[1..j-2], -1);
    	}
     }

     //when '~' is not present at the starting of string it returns the given path
     else
      return path;
  } 
}
```
Kindly go through this code

[...]

PR #8909, opened 2018-03-21
Adding expandUser Function

pachchi:
This PR addresses the issue #6008 
This PR adds expandUser Function
@lydia-duncan  Please review this PR and guide me about any changes to be made.

### DIFF
--- a/modules/standard/Path.chpl
+++ b/modules/standard/Path.chpl
@@ -457,4 +457,78 @@ proc file.realPath(): string throws {
      return ("", name);
    }
  }
+
+/* Returns the argument.  If the argument begins with `~` or `~<user>`, replace
+   that substring with that user's home directory, otherwise returns the
+   argument unchanged.
+
+   :arg path: a file or directory, which may or may not include `~` or `~<user>`
+   :type path: `string`
+
+   :return: the argument, replacing any `~` or `~<user>` with the user's home
+                directory
+   :rtype: `string`
+*/
+proc expandUser(path: string) : string
+  {
+    var i: int = 1;
+    var j: int = 1;
+    var k: int = 1;
+    var lenz: int;
+    var lenp: int = path.length;;
+    var count: int = 0;
+    var x: c_string ;
+    var y: c_string;
+       
+    //If the argument begins with `~` or `~<user>`, it replaces that substring with that user's home directory
+    if path[1] == '~' 
+    {
+      //checks the present directory by PWD
+      x = "PWD";
+      sys_getenv(x,y);
+      var  z = y:string;
+      lenz = z.length;
+
+       while(i <= lenz && count < 2)
+      	 {
+	   if z(i) == '/' then
+      	   count = count + 1;
+      	   i = i + 1;
+         }
+	 
+      	count = 0;
+      	 while(j <= lenz && count < 3)
+      	 {
+	   if z(j) == '/' then
+      	   count = count + 1;
+      	   j = j + 1;
+         }
+
+        //checks if argument begins with '~<user>'
+        if (lenp > 1 && path[2] != '/')
+        {
+         count = 0;
+         while(k <= lenp && count < 2)
+      	 {
+	   if path(k) == '/' then
+      	   count = count + 1;
+      	   k = k + 1;
+         }
+
+         if(z[i..j-2] == path[2..k-2]) then
+       	 return path.replace(path[1..1], z[1..i-1], -1);
+	 else
+	 return path;
+        }
+      
+       // checks if argument begins with '~'
+       else
+        return path.replace(path[1..1], z[1..j-2], -1);
+  
+     }
+
+     //when '~' is not present at the starting of string it returns the given path
+     else
+      return path;
+  } 
 }
--- a/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
+++ b/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
@@ -1,0 +1,10 @@
+/* Test cases to validate expandUser function in Path.chpl */
+
+use Path;
+
+writeln(expandUser('~user/'));
+writeln(expandUser('~USER/'));
+writeln(expandUser('~/'));
+writeln(expandUser('user/'));
+writeln(expandUser('~'));
+
--- a/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.good
+++ b/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.good
@@ -1,0 +1,5 @@
+/home/user/
+~USER/
+/home/user/
+user/
+/home/user
--- a/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.prediff
+++ b/test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.prediff
@@ -1,0 +1,11 @@
+#!/usr/bin/env python
+
+import os
+
+f= open('testcases_expandUser.good','w')
+f.write(os.path.expanduser('~user/') + '\n')
+f.write(os.path.expanduser('~USER/') + '\n')
+f.write(os.path.expanduser('~/') + '\n')
+f.write(os.path.expanduser('user/') + '\n')
+f.write(os.path.expanduser('~') + '\n')
+f.close()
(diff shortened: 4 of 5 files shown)


lydia-duncan:
It looks like the build failed due to the presence of tabs in your implementation.  Our standard practice is to replace any tab with two spaces.  I use a modification in emacs to automatically replace them for me when I try to use a tab, there should be something similar available to your text editor of choice

lydia-duncan:
We are past feature freeze, so this code will unfortunately not get merged until after 1.17 is branched.  But that doesn't mean we have to stop thinking about it, so I will continue to review it :)

modules/standard/Path.chpl
```diff
@@ -457,4 +457,78 @@ proc file.realPath(): string throws {
...
+   argument unchanged.
+
+   :arg path: a file or directory, which may or may not include `~` or `~<user>`
+   :type path: `string`
+
+   :return: the argument, replacing any `~` or `~<user>` with the user's home
+                directory
+   :rtype: `string`
+*/
+proc expandUser(path: string) : string
```
lydia-duncan: As a heads up, I spent a little effort yesterday to get this file organized alphabetically.  You are under no obligation to do so yourself, but note that I will move the function myself later if you don't (I would place between the implementation for dirname and file.getParentName)

modules/standard/Path.chpl
```diff
@@ -457,4 +457,78 @@ proc file.realPath(): string throws {
...
+  {
+    var i: int = 1;
+    var j: int = 1;
+    var k: int = 1;
```
lydia-duncan: Note that Chapel allows you to declare all of these on the same line, and would allow you to infer the type from the value provided (since the value `1` defaults to having a type of `int`).  To take advantage of either of those features, you would write these variable declarations like this:
```chapel
var i, j, k = 1;
```
or
```chapel
var i, j, k: int = 1;
```

modules/standard/Path.chpl
```diff
@@ -457,4 +457,78 @@ proc file.realPath(): string throws {
...
+    var lenz: int;
+    var lenp: int = path.length;;
+    var count: int = 0;
+    var x: c_string ;
+    var y: c_string;
```
lydia-duncan: Similarly, `x` and `y` could be declared on the same line:
```chapel
var x, y: c_string;
```

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.good
```diff
@@ -0,0 +1,5 @@
+/home/user/
```
lydia-duncan: The prediff will generate this file and the cleanfiles will remove it, so you don't want to commit it to the repository

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.prediff
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+
+import os
+
+f= open('testcases_expandUser.good','w')
+f.write(os.path.expanduser('~user/') + '\n')
```
lydia-duncan: Hmm.  It'd be beneficial to use a username that the program knows exists on the system (like the current user).  I'm not sure how to go about doing that myself, what resources can you find that would help you with this?

lydia-duncan: If you find out how to get that information when you're in bash, I can tell you how to send it to Chapel :)

modules/standard/Path.chpl
lydia-duncan: However, I would instead recommend declaring them closer to where they are used (inside the `if` statement), and giving `x` the value of `"PWD"` when you declare it.

modules/standard/Path.chpl
```diff
@@ -457,4 +457,78 @@ proc file.realPath(): string throws {
...
+       
+    //If the argument begins with `~` or `~<user>`, it replaces that substring with that user's home directory
+    if path[1] == '~' 
+    {
+      //checks the present directory by PWD
+      x = "PWD";
+      sys_getenv(x,y);
+      var  z = y:string;
+      lenz = z.length;
```
lydia-duncan: Okay, I figured out why I was confused by the PWD portion.  This code assumes that the directory where the program was run will be somewhere in the user's home directory hierarchy.  However, this isn't always the case - sometimes the user will be working in a mount that is separate from their home directory and they'll want to get the same answer regardless.

For example, say we have a file system defined like this:
```
- /
  - homes
     - Lydia
     - someoneElse
  - placeWhereIWork
     - myTestDir
  - placeWhereSomeoneElseWorks
     -theirTestDir
```
I'll want the result of this call to return the same thing regardless of if I'm executing as me in `/homes/Lydia` or `/placeWhereIWork/myTestDir`, but the code as it is implemented today would give different answers depending on where it was run.

This will take a bit more effort to implement.  I recommend taking a look at how python implements this function - you may need to add some explicit runtime support, similar to what [the chmod implementation](https://github.com/chapel-lang/chapel/pull/618) has done, for instance, or perhaps something more involved.  If you do end up using aspects of the python implementation, be sure to cite your work, similar to [this](https://github.com/chapel-lang/chapel/blob/master/modules/standard/FileSystem.chpl#L314).

modules/standard/Path.chpl
lydia-duncan: It would be good to give these variables more descriptive names than `x`, `y`, or `z`.  This will make it easier to maintain or understand the code.  Iteration variables like `i` are fine because they are generally understood to be for that purpose, though since there are so many of them it might be good to give them more precise names as well.

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
```diff
@@ -0,0 +1,10 @@
+/* Test cases to validate expandUser function in Path.chpl */
+
+use Path;
+
+writeln(expandUser('~user/'));
+writeln(expandUser('~USER/'));
```
lydia-duncan: This test appears to check against a user name you know exists on the system and one you know doesn't.  It would be good to make those cases more explicit (i.e. comment this so we know what you're checking!), and more robust (if this is run on a system that just happens to have both `user` and `USER` as user names, then you aren't testing the same thing you would be testing on a system that only defines one, or defines neither).

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
```diff
@@ -0,0 +1,10 @@
...
+writeln(expandUser('~/'));
+writeln(expandUser('user/'));
+writeln(expandUser('~'));
```
lydia-duncan: I don't have a problem with these last three test cases and am pleased you thought of them :)

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
lydia-duncan: You always want to test:
- current user
- a user that exists on the system and is not the current user
- a user that explicitly does not exist

test/library/standard/Path/Prachi-Muskaan/testcases_expandUser.chpl
lydia-duncan: It also might be good to double check the behavior when `~` is later in the path, e.g. `/path/gets/to/here/~/whoops`

lydia-duncan:
Thanks for all your hard work on this function!  I've made some suggestions in line :)

lydia-duncan:
Side note: I noticed in your test directory you have Prachi-Muskaan.  Does this mean you and @MuskaanJain are working together on this and the normPath implementation?  If so, please let me know so I can attribute you both correctly when I merge the changes :)

Also, you two appear to have different capitalization in the directory name.  Since the test files have different names that should be okay, but some systems are sensitive to capitalization and some are not so you'll want to be careful!

pachchi:
@lydia-duncan  Yes, I and @MuskaanJain are working on this and normPath function together.

lydia-duncan:
Master is basically unfrozen, so it is again possible to merge this change once it has been completed.  Thank you for your patience!

lydia-duncan:
Hi @MuskaanJain and @pachchi, what's the status on this PR?  Let me know if there's anything I can do to help :)

lydia-duncan:
Not hearing anything back from you guys, I'm going to close this PR.  Feel free to ping or reopen if you guys want to continue with it later!

PR #8910, opened 2018-03-21
Update language spec to document domain.dist

benharsh:
This PR documents ``domain.dist`` in the language spec and includes an example demonstrating valid usage in a where-clause to constrain the domain's distribution's type.

### DIFF
--- a/spec/Domains.tex
+++ b/spec/Domains.tex
@@ -958,6 +958,35 @@ D has 0 indices.
 \end{chapelprintoutput}
 \end{chapelexample}
 
+\index{domains!dist@\chpl{dist}}
+\index{predefined functions!dist@\chpl{dist}}
+\begin{protohead}
+proc $Domain$.dist() : dmap
+\end{protohead}
+\begin{protobody}
+Returns the domain map that implements this domain
+\end{protobody}
+
+\begin{chapelexample}{getDomainMap}
+In the code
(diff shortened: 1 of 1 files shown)


benharsh:
@bradcray : do you have the time and interest to review this?

Whoever does review this: note that the return type isn't actually ``dmap`` today. But I do think that's what we want long-term.

Also, are there any meta-tasks (e.g. regenerate PDF) that I need to do because I'm updating the spec?

bradcray:
I probably don't have time to review today.  You should build the PDF to make sure it works and looks as you'd expect, but you shouldn't commit a new PDF to the repository.

spec/Domains.tex
```diff
@@ -958,6 +958,35 @@ \subsection{Methods on All Domain Types}
 \end{chapelprintoutput}
 \end{chapelexample}
 
+\index{domains!dist@\chpl{dist}}
+\index{predefined functions!dist@\chpl{dist}}
+\begin{protohead}
+proc $Domain$.dist() : dmap
```
cassella: Should it be listed without parentheses here?

benharsh: Yup! I'll fix momentarily.

vasslitvinov:
Please run start_test on the new test, if you haven't already.

PR #8911, opened 2018-03-21
Spec and documentation updates

mppf:
See individual commits, but this PR combines documentation improvements.

* describe #define improvements for extern blocks
* describe coercion for Owned/Shared
* improve spec description of coercions for argument passing
* significantly rework and improve spec description of default, const intents

- [x] full local testing

Reviewed by @vasslitvinov and @daviditen - thanks!

### DIFF
--- a/doc/rst/technotes/extern.rst
+++ b/doc/rst/technotes/extern.rst
@@ -683,15 +683,18 @@ it in inline assembly.
 --------
 
 The extern block functionality allows simple #defines to be used from Chapel
-code. Simple defines take no arguments and define an integer value or use
-another #define. For example:
+code. Simple defines take no arguments and define an integer, string, or
+real value or use another #define that does so. Simple casts are also
+supported, as are macros that are simply an alternative name for another
+C variable. For example:
 
 .. code-block:: chapel
 
   extern {
    #define NEG_ONE (-1)
-   #define MY_NUMBER (NEG_ONE)
+   #define MY_NUMBER ((unsigned int)NEG_ONE)
   }
+  writeln(NEG_ONE);
   writeln(MY_NUMBER);
 
 However, it is easy to create #defines that are not supported because the
--- a/doc/rst/technotes/extern.rst
+++ b/doc/rst/technotes/extern.rst
@@ -682,11 +682,11 @@ it in inline assembly.
 #defines
 --------
 
-The extern block functionality allows simple #defines to be used from Chapel
-code. Simple defines take no arguments and define an integer, string, or
-real value or use another #define that does so. Simple casts are also
-supported, as are macros that are simply an alternative name for another
-C variable. For example:
+The extern block functionality allows simple #defines to be used from
+Chapel code. Simple defines take no arguments and define an integer,
+string, or real value or use another #define that does so. Casts to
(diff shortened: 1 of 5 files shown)


mppf:
@daviditen  or @vasslitvinov - would one of you review these doc changes? Thanks.

daviditen:
I can review this.

doc/rst/technotes/extern.rst
```diff
@@ -683,15 +683,18 @@ it in inline assembly.
 --------
 
 The extern block functionality allows simple #defines to be used from Chapel
-code. Simple defines take no arguments and define an integer value or use
-another #define. For example:
+code. Simple defines take no arguments and define an integer, string, or
+real value or use another #define that does so. Simple casts are also
```
daviditen: Do you think this should also define what a "simple cast" is?

spec/Conversions.tex
```diff
@@ -35,11 +35,18 @@ \section{Implicit Conversions}
...
+      \chpl{in} or \chpl{const in}. This conversion does not occur
+      for \chpl{type} arguments~(\rsec{Formal_Type_Arguments} even
+      if they specify a concrete type. It does not occur for generic
+      arguments~(\rsec{Generic_Functions}). It does
+      occur for param arguments~(\rsec{Formal_Parameter_Arguments})
+      with a specified type.
+
+% MPF: I don't see how this rule can be type correct
+%      and it also doesn't seem to be implemented right now
+%\item The formal argument of a function call is converted
```
daviditen: @vasslitvinov, Can you check if you think this should be removed?

spec/Procedures.tex
```diff
@@ -537,7 +537,8 @@ \subsubsection{The Const Intent}
 
 \begin{craychapel}
 The current implementation uses the following mapping:
-
+% TODO: does this really need to be about the "current implementation"?
+% Couldn't it just be about "the current language"?
```
daviditen: I agree that we should make it a language choice, not an implementation choice.  Assuming there's consensus on that, you can also remove the "Open issue" section just above this section.

spec/Conversions.tex
vasslitvinov: Right. This does not make sense to me. Let us remove this altogether.

What our implementation does is a separate issue, though, which we can discuss another time.

vasslitvinov: Let me take this back. It IS meaningful to convert the formal to the actual. The catch is that this conversion happens upon return from the function. So I propose to leave it in, clarifying "upon return".

We are not planning to implement this rule, probably ever. However we can still put it in the spec.

vasslitvinov: Why put it in the spec if we do not implement it?
The spec is not a transcription of what we have or will implement, it is more of an ideal language design.
Still, why bother? I suggest to allow this conversion on `out` because it makes the language more general i.e. permits more programs.

mppf: Thanks for the discussion, for now I'll leave a comment in the spec noting it is not implemented, and I'll leave the rule in there for now. If we ever get around to revising `out` intents the way I revised `in` intents in this past release cycle, we could get this behavior pretty easily I think.

PR #8912, opened 2018-03-21
Add documentation for chpl tab autocompletion script

daviditen:
Document the chpl-completion.bash script in usingchapel/compiling.


daviditen:
Thanks @ronawho!  One of the nice things about autocompletion is that it helps remember the obscure flags that you rarely use.  :)

But I get what you're saying about using a more important flag.

PR #8915, opened 2018-03-21
tidy URLs in spec

vasslitvinov:
* remove `latest` from paths
* point to specific pages instead of to the doc root
* uniform formatting
* click on each link in the document to verify that it works


vasslitvinov:
@benharsh would you give a quick look?

PR #8918, opened 2018-03-22
Add skipif to squelch openssl error for mason-run

ben-albrecht:
Use @daviditen's previous `.skipif` for avoiding sles 11 machines, which have outdated openssl.


ben-albrecht:
I have reproduced the failure and confirmed that this resolves the error on a sles 11 machine.

PR #8919, opened 2018-03-22
paratest.client: update ARGV test, usage message, and commentary

cassella:
paratest.client unconditionally references `$ARGV[6]`, so update its "insufficient arguments" check to ensure there are that many args.

Fix the usage message to show that passing `execopts` is possible only if `compopts` are passed.

Update the comments at the top of the script to describe the `memleaks` and `execopts` options.

### DIFF
--- a/util/test/paratest.client
+++ b/util/test/paratest.client
@@ -3,7 +3,7 @@
 # Client-side of the parallel testing script, paratest.server.
 # Used remotely by paratest.server to run start_test locally.
 #
-# Usage: paratest.client id chapeltestdir testdir chplenv futures-mode valgrind [compopts]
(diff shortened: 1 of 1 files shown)


cassella:
I have no way of testing a paratest, so I didn't.

ronawho:
This looks fine and harmless to me, but I'll wait till post-release since I don't think there's any pressing need to get it in.

ronawho:
and fyi you should be able to do a local paratest and limit to subset of dirs so it doesn't take forever. Something like:

```sh
cd $CHPL_HOME/test/
$CHPL_HOME/util/test/paratest.server -dirs release/examples
```

`util/test/paratest.local` might make it even easier to run locally, but I'm not that familiar with that script. I think it basically just does a local paratest.server run and tries to figure out a good `-nodepara` based on the number of cores and avail mem.

PR #8920, opened 2018-03-22
Cray platform doc updates

gbtitus:
Update the Cray platform doc file for comm=ugni extendable heap and number of comm domains on Aries and no more 'Bus error' on out-of-memory, plus reorganize now that default heap behavior is so different between ugni and gasnet.


ronawho:
Looks good to me, but I'm pretty numb to this document anymore, so it wouldn't hurt to get some fresh eyes on it

daviditen:
I looked this over and it looks generally good to me.

Issue #8131, opened 2018-01-05
type resolve error under unusual circumstances


ben-albrecht:
### Summary of Problem

A class with a domain of strings incorrectly raises a type resolve compilation error under a specific circumstance (see code below). I was not able to reduce the reproducing example any further, but find it hard to imagine it can't be distilled further.

### Steps to Reproduce

**Source Code:**

The following program fails to compile with the `unable to resolve type` error:

```chapel
// Compiling gives " error: unable to resolve type (annotated below)

var watDom: domain(string);
var wat: [watDom] WAT;

class WAT { }

class R {
  var dom: domain(string); // <-- unable to resolve this type
}

class C {
  proc this(idx: int):R { return new R(); }
}
```

Changing almost anything in the above example results in the program compiling correctly, e.g. removing the `: int` type in `proc this`, or changing `watDom` to a `domain(int)`.

For reference, the following program compiles successfully:

```chapel
// For reference, this works
var r = new R();

var c = new C();
var r2 = c[2];


class R {
  var dom: domain(string);
}

class C {
  proc this(idx: int):R { return new R(); }
}
```

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.17.0 pre-release (81b7d17)`

ben-albrecht:
This bug was reported by @marcoscleison in https://github.com/marcoscleison/cdo/issues/7

ben-albrecht:
Tried compiling the working case with valgrind to see if were just getting lucky when it worked, but it did not change the result.

mppf:
Removing the return type declared with the `proc this` should also fix it.

I think there's something going wrong with the default type constructor for R when it's trying to be created from the return type.

daviditen:
The index type of the two domains doesn't appear to matter as long as they match.  The following version fails in the same way with `t` set to `int`, `real`, `complex`, etc.  But if you set them to any two different types it compiles.

```
config type t = string;

var watDom: domain(t);

class R {
  var dom: domain(t); // <-- unable to resolve this type
}

class C {
  proc this(idx: int):R { return new R(); }
}
```

mppf:
Here's another interesting tidbit: the error occurs when trying to resolve _type_construct_R from within a call to resolve dsiNewAssociativeDom for the `watDom:domain(string)` declaration. That seems pretty odd.

cassella:
If I change ``proc this(`` to ``proc that(``, it compiles.

bradcray:
[...]

PR #8921, opened 2018-03-22
Specify a type in DefaultAssociative as a workaround

mppf:
Issue #8131 describes a problem with function resolution.
This change is a simple workaround for that particular issue
and takes the form of specifying DefaultAssociative's
field tableSize as :int rather than letting the compiler infer it.

- [x] full local testing
- [x] baseline testing
- [x] --no-local testing
- [x] --verify testing
- [x] GASNet testing

### DIFF
--- a/modules/internal/DefaultAssociative.chpl
+++ b/modules/internal/DefaultAssociative.chpl
@@ -64,7 +64,7 @@ module DefaultAssociative {
     var numEntries: atomic_int64;
     var tableLock: atomicbool; // do not access directly, use function below
     var tableSizeNum = 1;
-    var tableSize = chpl__primes(tableSizeNum);
+    var tableSize:int = chpl__primes(tableSizeNum);
     var tableDom = {0..tableSize-1};
     var table: [tableDom] chpl_TableEntry(idxType);
   


bradcray:
This looks attractively simple and would be nice to close that issue.  That said, do we have any hope of creating a standalone test that runs into the same function resolution issue so that we're not (completely) sweeping it under the rug?  Or do we think it's specific to our default associative stuff in some way?

I'm also noticing what a large tuple chpl_primes is, and tend to think of tuples in general, but particularly large ones, as being a bog on compile-time at present.  Would rewriting this in some other way (e.g., a const global array, or ...) save us some compile-time without hurting execution time performance?  (though I suspect a change of that magnitude should be saved until after the release).

mppf:
@bradcray @benharsh - so far I havn't been able to create a standalone test that shows the same issue. I do however have a branch https://github.com/chapel-lang/chapel/compare/master...mppf:move-resolve-specified that resolves the issue in a different way that we might expect to be more general. I'm just not so sure that branch makes sense for the release at this point.

The issue seems to have to do with ending up in a recursive resolution situation, where we're partly through resolving one function (buildDomainRuntimeType) and then we're resolving the return type for a candidate (R) and we need the other function to be resolved in order to do so. The compiler currently seems to assume that if it started resolving a function that it has completed doing so - or else the function has a declared return type (e.g. for recursive functions).

mppf:
@bradcray @benharsh - I found a non-domain version of the failing example that my WIP branch resolves. It's both now in a comment at the end of issue #8131 and in my WIP PR for that change.

PR #8923, opened 2018-03-22
Remove this.complete() calls from LCALS

bradcray:
While reviewing how our examples/ directory had changed between
releases, in general and w.r.t. gaterhing release-over-release
timings, it seemed like some cases of this.complete() would not be
necessary with minor adjustements to the initializers (with the goal
of making the initializers simpler, avoiding re-initialization, and
making the code backwards compatible with 1.16.  LCALS seems to
have been a pretty easy change.


bradcray:
@benharsh and @daviditen : Would you guys let me know if you're OK with these changes?  I did correctness testing on master and ensured that 1.16.0 could compile the code.  Have not done any performance testing.

PR #8925, opened 2018-03-22
Minor touch-ups to primer examples w.r.t. initializers

bradcray:
* added explicit initializer to errorHandling.chpl primer to avoid
  use of pragma "use default init" (when adding it, I'd hoped that
  it would no longer be necessary by now)

* Rewrote copy initializer in learnChapelInYMinutes to make it
  phase 1 only (arguably unnecessary, but seemed "simpler"?)

### DIFF
--- a/test/release/examples/primers/errorHandling.chpl
+++ b/test/release/examples/primers/errorHandling.chpl
@@ -60,8 +60,10 @@ try {
   types, developers can also create their own hierarchies.
  */
 
-pragma "use default init"
-class EmptyFilenameError : Error { }
+class EmptyFilenameError : Error {
+  proc init() {
+  }
+}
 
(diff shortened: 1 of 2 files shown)


bradcray:
@psahabu : Would you check the changes to errorHanding.chpl here?
@benharsh: Any objections to this change to learnChapelInYMinutes?  (also, have we put any effort, that you know of, into keeping the version on the learnXInYMinutes site up-to-date?)

benharsh:
There were a few updates to the version on the learnXInY site last year. A quick comparison between the two files shows two major differences:
1) We now use reStructuredText but the site uses Markdown
2) Initializers are missing from the site's version

The code change looks fine to me.

PR #8926, opened 2018-03-22
call resolveSpecifiedReturnType only for candidates

### LINKED ISSUE #7836: Where clause cannot filter functions with uint(?w) arguments
The `where` clause in the below program fails to limit the error cases produced by the compiler (namely generating real(8) for example). The program fails to compile with this error 

```
d.chpl:5: error: illegal size 16 for real
d.chpl:5: error: illegal size 8 for real
```

``` chapel
proc fp754putZ(v : uint(?w)) : real(w)
   where w == 32 || w == 64
{
  var x:real(w) = v:real(w);
  writeln(v.type:string);
  writeln(x.type:string);
  return x;
}

proc test() {
  type R = real(32);
  type U = uint(numBits(R));
  const zero = 0:U;

  writeln(fp754putZ(zero));
}

test();
```

The error is being produced in normalize but whether the `where` clause applies can only be known in resolution.

### LINKED ISSUE #8131: type resolve error under unusual circumstances

mppf:
Before this PR, resolveSpecifiedReturnType was called in resolveSignature, 
which was called from checkResolveFormalsWhereClauses during candidate
filtering (and so was applied to all visible functions). This commit changes
resolveSpecifiedReturnType to only be called on candidates that are potential
matches - that is, those that have passed argument type based filtering as well
as where clause filtering.
    
Initially I tried to move the resolution of specified return types closer to 
the inference of return types when the return type is not specified; however a
straightforward attempt in that direction breaks resolution of recursive
functions.

Besides moving resolveSpecifiedReturnType, in order to get testing to pass, I 
needed to change the way that specified return type blocks are removed from the
AST. Now they are all removed at the end of resolution, in pruneResolvedTree,
instead of being removed in resolveSpecifiedReturnType. While there, I added
code to move functions in the return type block out, to resolve issue #8131.

Closes #8131.
Closes #7836.

- [x] full --baseline testing
- [x] full --no-local testing
- [x] full --verify testing
- [x] full GASNet testing


compiler/resolution/ResolutionCandidate.cpp
```diff
@@ -55,6 +55,11 @@ bool ResolutionCandidate::isApplicable(CallInfo& info) {
     retval = isApplicableGeneric (info);
   }
 
+  if (retval && fn->retExprType != NULL && fn->retType == dtUnknown) {
+    // Should this only happen for FLAG_GENERIC not set?
```
mppf: I think FLAG_GENERIC does not need to be checked here because by the time it's a candidate it won't be generic anymore (but rather instantiated).

PR #8927, opened 2018-03-22
Add util/chpl-completion.bash to gen_release

daviditen:
Added util/chpl-completion.bash to the list of files gen_release should
include.  Tested by running gen_release and unpacking the result to verify that
it included util/chpl-completion.bash.


daviditen:
Thank you!

PR #8928, opened 2018-03-22
Optional command line parameters for easy_install and pip install, in Makefiles

awallace-cray:
Optional env variables CHPL_EASY_INSTALL_PARAMS and CHPL_PIP_INSTALL_PARAMS
are added to easy_install and pip install command lines in the Make
rules for "make test-venv" and similar.
    
This facilitates using a local PyPI mirror or cache, as workaround
for PyPI's rejecting https connections from clients built with older
versions of OpenSSL. On our SLES-11 machines, for example.
    
From PyPI website:
    In preparation for our CDN provider deprecating TLSv1.0 and TLSv1.1
    protocols, we have begun rolling brownouts for these protocols for
    the first ten (10) minutes of each hour.

### DIFF
--- a/third-party/chpl-venv/Makefile
+++ b/third-party/chpl-venv/Makefile
@@ -46,7 +46,7 @@ $(CHPL_VENV_INSTALL_DIR):
 # Install virtualenv program.
 $(CHPL_VENV_VIRTUALENV): $(CHPL_VENV_INSTALL_DIR) check-exes
 	export PYTHONPATH=$(CHPL_VENV_INSTALL_DIR):$$PYTHONPATH && \
-	$(EASY_INSTALL) --install-dir=$(CHPL_VENV_INSTALL_DIR) $(shell cat virtualenv.txt)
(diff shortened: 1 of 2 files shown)


awallace-cray:
This change has been tested by using it in the ARM module builds (and in normal module builds from the same src).

dmk42:
If you want to add the other uses of pip install to this PR, that would be fine.

awallace-cray:
Notes: 
There are ways to set pip options more globally. ~/.pip/pip.conf for example. 
I don't think any of them help easy_install. 

If we use a PyPI mirror w http but not https URL, then pip will complain unless we put in another command line parameter, --trusted-host <the bare hostname used in the mirror URL>.
I don't see any reason we could not stuff both the URL and the --trusted-host <hostname> into this env variable. 

I did not have that problem when I used this change before. That time, I used it to point to a file URL that I generated with pip2pi

awallace-cray:
Latest commit was tested in sandbox

PR #8936, opened 2018-03-23
Update documentation for LLVM version and for ARM-based Cray systems

dmk42:
This updates the documentation to include changes to supported versions of LLVM, and support for the Cray XC50 system with ARM processors.

### DIFF
--- a/doc/rst/usingchapel/chplenv.rst
+++ b/doc/rst/usingchapel/chplenv.rst
@@ -624,8 +624,9 @@ CHPL_LLVM
    If unset, ``CHPL_LLVM`` defaults to ``llvm`` if you've already installed
    llvm in third-party and ``none`` otherwise.
 
-   Chapel currently supports LLVM 4.0 and 6.0. LLVM 5.0 has a known
-   optimization bug that affects Chapel.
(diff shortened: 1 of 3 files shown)


doc/rst/usingchapel/chplenv.rst
```diff
@@ -624,16 +624,21 @@ CHPL_LLVM
    If unset, ``CHPL_LLVM`` defaults to ``llvm`` if you've already installed
    llvm in third-party and ``none`` otherwise.
 
-   Chapel currently supports LLVM 3.7 through 4.0.
+   Chapel currently supports LLVM 4.0 and 6.0. LLVM 5.0 has a known
+   optimization bug that affects Chapel.
```
mppf: This makes it sound like we can install llvm 4 and use CHPL_LLVM=system but that's not the case. maybe above it needs to say "note: the LLVM must be the version supported by Chapel" in the table above or something?

mppf: either that or explicitly say that versions of LLVM before 6 needed internal headers.

dmk42:
@mppf - Thanks!  Does this look OK?

Issue #7961, opened 2017-12-06
[Mason] Illegal characters used as identifier result in unbuildable default project.


BryantLam:
`mason new chpl-project` will create a new module named "chpl-project". The dash is an illegal character and the project will not build. Other illegal characters include the period (`.`).

Potential solutions
- Convert all illegal characters to legal characters, such as an underscore. May affect name-based path traversals in Mason.
- Truncate the name (Cargo takes this approach when it finds a hyphen), but this gets into corner-case situations. For example: `mason new chpl-myproject` --> `module myproject`, but what should `my-project-name` do?
- Forbid creating the project with an illegal character used as an identifier. NOTE: There will be a difference between modules (`module <name>`) and applications (`proc main()` without direct use for the project name.

ben-albrecht:
My preference is to avoid modifying the module name implicitly. Therefore, I find this potential solution most appealing:

> - Forbid creating the project with an illegal character used as an identifier. NOTE: There will be a difference between modules (module <name>) and applications (proc main() without direct use for the project name.

However, I think we could still support package names with illegal characters by only forbidding projects with an illegal character if a separate module name is not explicitly provided, e.g.

```
> mason new chpl-project
Error: Illegal character in module name. Try: --moduleName=<legal module name>

> mason new chpl-project --moduleName=project
Creating package: chpl-project
```

I can see the appeal for package names differing from module names for applications / command-line tools, but I am not sure the same is true for library packages (which is all that `mason` supports today). Is there a motivating use-case for the latter?

Also, I wonder if it would be worth special-casing something like `chpl-*` to `*` to handle the potential package naming convention. It may be too soon to tell.

BryantLam:
We can take it in steps since these all build on each other from most restrictive to least restrictive:
1. Forbid illegal characters in module name. I'm okay with this.

2. Add an option like `--moduleName`. I'm neutral on a new option since it doesn't have a straightforward value proposition (e.g., applications can already write all of their code in separate modules that are not tied to the Mason project name). Is the project name in Mason.toml used for anything (and does it have any illegal-character issues, esp. w.r.t. a Mason registry)?

3. Transform the project name to make it a legal module name. I'm against this since you'll have to document a transformation and ensure it works (it's also not straightforward behavior). But if this is important for people, it can be added because it relaxes the steps above.

ben-albrecht:
[...]

PR #8941, opened 2018-03-23
Tidy up mason new package name checking

mppf:
See also issue #7961.

This PR adjusts `mason new` to better check the name of the package. While there, it removes the default package name of `MyPackage` and tightens up the check for an existing directory.

Package names that aren't Chapel identifiers aren't allowed and neither are package names that contain `$` (which is legal in Chapel identifiers). The rationale for disallowing `$` in package names is that it won't go well when used from a shell.

- [x] full local testing

### DIFF
--- a/tools/mason/MasonNew.chpl
+++ b/tools/mason/MasonNew.chpl
@@ -29,17 +29,13 @@ use MasonEnv;
 
 proc masonNew(args) {
   if args.size < 3 {
-      writeln('error: Invalid arguments.');
-      masonNewHelp();
-      exit();
-    }
-  else if isDir(args[2]) {
-    writeln('A directory with that name already exists');
-  }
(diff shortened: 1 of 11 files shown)


tools/mason/MasonNew.chpl
```diff
@@ -55,12 +51,48 @@ proc masonNew(args) {
...
+      writeln("A directory named '", name, "' already exists");
+    } else {
+      InitProject(name, vcs, show);
+    }
   }
 }
 
-
-
+proc isIdentifier(name:string) {
```
ben-albrecht: I'd be in favor of a simple doc-string clarifying this function, e.g.

```
/* Return 'true' for valid identifiers according to Chapel spec, otherwise 'false' */
```

PR #8944, opened 2018-03-23
Move chplgmp.c out of runtime

mppf:
Instead of chplgmp.c being in the runtime, this PR moves it
to modules/standard/GMPHelper and it is compiled on-demand
when GMP is used (via a require statement).

Why do this? So that a system GMP with a different version from
the one used to build Chapel may be more easily used.

Removes GMP from the runtime build path, since the runtime no longer
depends on or uses GMP.

Adjusts printchplenv to print CHPL_GMP by default but otherwise to treat
it as an internal option. That removes it from the runtime lib path.

Removes the GMPRandom initializer that copied state from another
GMPRandom since it is too hard to support across locales. (See
issue #9390).

To enable this change, added modules/standard to the C compiler include
path. modules/packages was already there. As a follow-on to PR #7204,
which added modules/packages to the C compiler include path, move
these paths from RUNTIME_INCLS to HEADER_CFLAGS since they are used
only when building the target program (and not when building the
runtime). Since RUNTIME_INCLS ends up in list-includes-and-defines but
HEADER_CFLAGS does not, these `-I` paths also needed to be added
to the clang invocation in clangUtil.cpp.

- [x] full local testing
- [x] test/library/standard/GMP with GASNet
- [x] test/library/standard/GMP with --llvm
- [x] test/library/packages/Crypto


modules/standard/GMPHelper/chplgmp.h
```diff
@@ -17,15 +17,10 @@
  * limitations under the License.
  */
 
-#ifdef CHPL_HAS_GMP
```
ronawho: Seems a little unusual to have C header files be a part of the chapel standard library. 

Looking at this file, I don't see any reason this needs to be in C, could these just be chapel functions that call extern procs?

mppf: First, I agree. Second, this PR was intended to take the easiest approach, at least initially.

It's certainly the case that many if not all of these could be re-implemented in Chapel, but I didn't view taking the time to do so as required to make this change. If you think it *is* required, then I can find time to do that. I'm not yet sure if all of them can be implemented in Chapel (some of them are workarounds for LLVM issues with GMP headers).

ronawho: Personally, I'd be in favor of trying to convert it to a Chapel module, and if that takes more than an hour or two making it into an issue. 

I worry that if we put it off it'll basically never get prioritized because things work "well enough"

mppf: @ronawho - I had another look & updated this PR to move as much as I can with reasonable effort to module code. It took more effort than expected to do so because in testing the GMPRandom copy-across-locales code it became clear that it wasn't working properly.

Small functions continue to exist in the C header for one reason or another:
 * Chapel module doesn't know how to look inside __mpz_struct / __gmp_randstate_struct
 * chpl_gmp_mpz_get_str: extern interoperability with strings is awkward
 * chpl_mpz_cmp_si ... -- the LLVM backend can't call these GMP functions because they are written as macros

modules/standard/GMP.chpl
```diff
@@ -1124,7 +1204,82 @@ module GMP {
       if a.locale == here {
         gmp_randinit_set(this.state, a.state);
       } else {
-        chpl_gmp_get_randstate(this.state, a.locale.id, a.state[1]);
+        // Figure out which known algo RNG a.state is using
+        var algo = 0;
```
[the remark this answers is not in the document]
mppf: We were notionally supporting copying a GMPRandom across locales. But that didn't actually work before. In order to get it to work, we need to initialize a new GMP RNG and then copy the state from the old one. This works for the Mersenne Twister but not other GMP RNG (the LCG uses a nested pointer-based state that isn't so easy to copy). Of course, we could decide to stop trying to support that.

modules/standard/GMPHelper/chplgmp.h
```diff
@@ -0,0 +1,122 @@
+/*
```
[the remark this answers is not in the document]
mppf: I just include it as "GMPHelper/chplgmp.h" so I don't believe the subpath needs to be in the include paths.

ronawho:
I'm not following some of these changes (most notably the randstate ones). 

It's also not obvious to me if this PR is a net win/improvement. Might be worth having a second person look at it.

modules/standard/GMP.chpl
```diff
@@ -109,11 +109,51 @@ module GMP {
   use SysError;
   use BigInteger;
 
+  require "GMPHelper/chplgmp.h";
+
+  export proc chpl_gmp_alloc(size:size_t) : c_void_ptr {
+    pragma "insert line file info"
+    extern proc chpl_mem_alloc(size:size_t, md:chpl_mem_descInt_t) : c_void_ptr;
+    extern const CHPL_RT_MD_GMP:chpl_mem_descInt_t;
+    return chpl_mem_alloc(size, CHPL_RT_MD_GMP);
```
ronawho: I wonder if we can just use `chpl_here_alloc()` and friends now instead of duplicating the externs. I won't hold up the PR over that, but it might be a nice cleanup

ronawho:
There are still some references to gmp in the runtime. Some of them are still needed, but maybe some of the makefiles are no longer needed?

mppf:
> There are still some references to gmp in the runtime. Some of them are still needed, but maybe some of the makefiles are no longer needed?

I didn't try to sort this out further in this PR but I agree it's worth looking into at some point.

PR #8955, opened 2018-03-25
Resolve new 9

noakesmichael:
Bypass _new() wrappers for generic class allocations

Early efforts to support initializers for classes attempted to wrap the alloc/init pair in a generated
_new() function.  Recent efforts to use initializers for generic classes exposed cases that where
the approach used was failing to interact with function resolution in an acceptable way.

This PR bypasses the _new() function by open-coding the calls instead.  In the longer term
we would prefer to find a path to rely on the _new() functions/methods again.

This PR also refactors portions of resolveNew*() to clarify the overall business logic.
This PR includes a test that locks in the required behavior.  This test failed to resolve
before this PR.

Limited testing for each configuration.
Passed a full single-locale paratest with -futures


benharsh:
Looks good to me!

PR #8970, opened 2018-03-26
Fix typo in sub_test comment on chapel testing valgrind environment variables

lydia-duncan:
CHPL_TEST_VGRND_EXE was accidentally CHPL_TEST_VBRND_EXE (note: B instead of G)

Testing
- [x] paratest
- [x] valgrind exec run on one test
- [x] valgrind run on one test


lydia-duncan:
I want to do a double check on this before merging, but it was a comment change so I'm not actually worried

ronawho:
yeah, that's way more testing that I would do, but obviously it doesn't hurt

PR #8977, opened 2018-03-27
Fix cut and paste duplication that BenA found

bradcray:


bradcray:
@ben-albrecht : Just double-checking I got it right this time.

Issue #8472, opened 2018-02-14
Section in Docs on Configuring Tests


Tshimanga:
Maybe I just haven't found it, but it may be helpful for newcomers to have a section in the docs on configuring chapel tests. An overview on the default settings, how it works, and  things like the process/syntax for changing the runtime upper bound, pointing to db credentials, modifying compopts, etc.

daviditen:
This isn't particularly easy to find and could probably stand to be linked more directly from the [developer resources](https://chapel-lang.org/developers.html) page:

https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst

cassella:
FWIW, that doc seems far from comprehensive.  For example, it doesn't mention ``.skipif`` or ``.suppressif``, or the difference between them, or the syntax of the files.  Or what the developer mindset should be when creating a future of a failure that doesn't fail on all platforms or configs.

It doesn't describe how a ``.compopts`` or ``.execopts`` can cause a single ``.chpl`` test to have multiple ``.good`` files.  Though it does have a TODO noting the absence.

Also, it suggests to the reader:

> For more information on these capabilities, review the comments at the
> top of ``start_test`` which is currently the most complete documentation
> on the testing system.

But there are no comments at the top of the python ``start_test``.

lydia-duncan:
I have an improved version of the document over in #8978.  @Tshimanga and/or @cassella, feel free to give feedback on the approach.

jabraham17:
Closing, the [Test System docs](https://chapel-lang.org/docs/main/developer/bestPractices/TestSystem.html) seem much improved since this issue was opened

PR #8978, opened 2018-03-27
Revamp the TestSystem helper document

lydia-duncan:
We received feedback that the document wasn't as useful to new developers.
This change attempts to restructure it so that information is organized based
on the problem you are trying to solve, rather than code words that you need
to already know exist.  It also extends the list of relevant files to cover more
cases (though it is intentionally still not a complete list), includes information
about some of these additional file types, and adds an outline.

This is best reviewed by looking at document as a whole, using the "view"
button, instead of looking at the diff.


lydia-duncan:
@ronawho - I think this is ready to take a look.  I'm also intending to run it past some of our more recent contributors.

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -466,6 +638,182 @@ include a future for the issue tracked on the `GitHub issues page`_.
...
+                    *category*, *title*, *issue #*
+foo.bad             output generated on a failing test, to track if a known
+                    failing future begins failing a different way.  See
+                    `Tracking Current Failure Mode`_ for more information
+..
+=================   ===========================================================
+
 .. _extensions:
 
 Planned Changes of Testing System
```
ronawho: You haven't changed this, but I would just remove this section, or at least make it a non-rendered section

We're not going to get to YAML anytime soon (and if we ever do this document will change drastically), tracking of reference benchmarks isn't that relevant to external devs, and I think this PR improves documentation enough that the last bullet can just be removed.

lydia-duncan:
> Some "recipes" for common but non-trivial cases. For example, a future that passes in some configurations. I've seen a few examples, so I think the approach is to write it as a normal future, but add a .skipif to skip the configurations it passes on, and to add another copy with the inverse .skipif to lock in the passing behavior. (And a note in the .future that when it's retired, to merge the tests.)

I think that's a fair point, but since futures are intended as action items, I think it would be confusing to merge an example into the test system.  I can point to a PR with a dummy future like that which will never get merged, but that's not future-proof if we eventually move away from GitHub to the next big thing (that said, there are already references to GitHub specific aspects, so maybe that's not a big detractor).  Thoughts?  (also curious what @ronawho thinks)

> Another example I don't know how to do is a future recording a memory leak. I imagine it would involve an .execopts specifying --memLeaks, and a .prediff to account for line numbers and addresses changing, but I don't have a good example .prediff for that. Are there other common but non-obvious cases? Maybe give pointers to a few tests that exemplify them.

I'm not entirely certain how to do this either, but I wouldn't consider it a common case since memleaks tracking has its own nightly configuration independent of futures.  When we eventually reach a state where memory leaks are almost non-existent, the answer will be different.

> Speaking of .prediff, I think it could be a bit more exposition. Looking at a few examples, I think the .prediff script is invoked with the name of the test in $1 and the file containing the test output in $2, and should leave its modified output in a file of the same name as $2. Is that right? does the script have any other interfaces?

Those are the most important parts that I'm aware of.  I debated about adding more information for this case but it felt very nitty-gritty so I backed away from it, especially because the document was getting rather large.

> The new doc mentions a few different execution modes (performance, comm-none, no-local, lm-numa, chpldoc), and how to identify a performance run in a .skipif file (CHPL_TEST_PERF). Are those all the modes that a test author needs to be cognizant of? Are all combinations of them valid?

Those are more execution modes than the test author needs to be cognizant of, and sometimes less.  It really depends on what the test author is validating.  In addition to these, we cover behavior on different machines, with different compilers, memory leaks, certain specific compilation modes, etc.  But for the most part, if a developer is getting used to the testing system, they really only need to know about a couple of modes.

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -466,28 +637,178 @@ include a future for the issue tracked on the `GitHub issues page`_.
...
+                    information.
+foo.notest          empty file. Indicates the file should not be run explicitly
+                    See `Controlling How It Runs`_ for more information.
+NOTEST              empty file. Indicates the directory should not be run
+foo.skipif          line separated list of conditions under which the test
+                    should not be run, or a script to compute the same.  See
+                    `Limiting Where the Test Runs`_ for more information
+SKIPIF              same as above, but applied to the entire directory
+foo.suppressif      line separated list of conditions under which the test is
+                    expected to fail, or a script to compute the same
```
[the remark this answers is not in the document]
ronawho: It's pretty subjective, and we're not that consistent ourselves. Generally a supressif is for something we expect/want to fail forever in one config (pretty rare), or something that was passing and is now failing for some reason (usually outside our control), but we expect/hope it to work in the future. 

For some examples see https://github.com/chapel-lang/chapel/pull/8258 or https://github.com/chapel-lang/chapel/pulls?q=is%3Apr+suppressif+is%3Aclosed+author%3Aronawho

ronawho:
>>Some "recipes" for common but non-trivial cases. For example, a future that passes in some configurations. I've seen a few examples, so I think the approach is to write it as a normal future, but add a .skipif to skip the configurations it passes on, and to add another copy with the inverse .skipif to lock in the passing behavior. (And a note in the .future that when it's retired, to merge the tests.)

>I think that's a fair point, but since futures are intended as action items, I think it would be confusing to merge an example into the test system. I can point to a PR with a dummy future like that which will never get merged, but that's not future-proof if we eventually move away from GitHub to the next big thing (that said, there are already references to GitHub specific aspects, so maybe that's not a big detractor). Thoughts? (also curious what @ronawho thinks)

I think we should keep this document as simple as possible and cover common cases that typical external devs will need to care about. @cassella is clearly not a "typical" external developer :) For more complicated (and less frequent) cases we could create a secondary document, or TBH what I do is usually find an example in the test system and look through the git log

doc/rst/developer/bestPractices/TestSystem.rst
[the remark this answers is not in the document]
lydia-duncan: > something that was passing and is now failing for some reason (usually outside our control), but we expect/hope it to work in the future

One example of this is a failure due to a bug in a target compiler (gcc, icc, pgi), or a bug in a package

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -466,28 +637,178 @@ include a future for the issue tracked on the `GitHub issues page`_.
...
-=================================
+Tracking Current Failure Mode
++++++++++++++++++++++++++++++
+
+Sometimes a future will change its behavior, but not be resolved.  The future
+should be updated to continue to track the issue as much as possible - to alert
+developers when this happens, it is necessary to track not only the expected
+good output but also the output indicating the current failure.  This is done
+via a ``.bad`` file.  The contents of a ``.bad`` file are similar to a ``.good``
+file and should match the currently generated output of the test.
```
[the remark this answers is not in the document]
lydia-duncan: Providing `.compopts` or `.execopts` results in the output file being named `<test_name>.1.exec.out.tmp` or something along those lines (`.1-2.` when both are used).  I don't know how useful going into those details are, but it might be worth calling that strategy out since the presence of `CHPL_DEVELOPER=true` in the environment will alter the output generated in some cases (so compiling the program on outside of the testing system might lead to frustration)

doc/rst/developer/bestPractices/TestSystem.rst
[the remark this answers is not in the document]
lydia-duncan: Because the use case of `suppressif` is so variable (and not necessarily well defined), Elliot and I thought it best not to give `suppressif` its own section.  @cassella - would you prefer we defined it more thoroughly here?

lydia-duncan:
I think this is ready for a second look

ben-albrecht:
Should this close #8472 ?

lydia-duncan:
> Should this close #8472 ?

I was under the impression that issue was also asking for a more obvious link to the file, either from our developer resources page or in the documentation section

doc/rst/developer/bestPractices/TestSystem.rst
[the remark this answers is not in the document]
cassella: Maybe just enough that someone reading the doc to write a test would know which to use.  Or at least, would get the impression to use a ``.skipif`` unless a reviewer says otherwise.

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -285,15 +295,22 @@ single-locale setting:
      CHPL_COMM != none
 
 This is useful when the conditions required to skip a test can be easily
-determined from the environment.
+determined from the environment.  A condition of ``>=`` indicates that the test
+should be skipped when the environment variable on the left does not contain the
+contents on the right, while ``<=`` indicates the opposite - this is useful for
+imprecise matches, e.g. ``CHPL_HOST_PLATFORM >= cygwin`` would cause the test to
+run on both ``cygwin64`` and ``cygwin32``.
```
cassella: I think this would be a bit easier to follow if ``<=`` were given the explicit definition, since it has fewer "not"s.

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -327,9 +508,24 @@ would use ``bar-100.perfkeys`` for its performance keys and write its
...
+performance testing.
+
+C versions do not have to be performance tests, but this is their most common
+use case.
+
 
 Creating a graph comparing multiple variations
-----------------------------------------------
+++++++++++++++++++++++++++++++++++++++++++++++
+
```
cassella: An existing line later on refers the reader to the performance graphs on sourceforge.  Is the chapel-lang site the official one now?

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -103,86 +103,290 @@ example, given a directory containing two files:
...
+     --x=false # foo.false.good
+
+will compare test output to ``foo.true.good`` for the first execution and
+``foo.false.good`` for the second.
+
+Any line that is unlabeled will use the default ``.good`` for that test.
+Undefined behavior will occur when both the ``.compopts`` and ``.execopts``
+files specify a ``.good`` file in this way.
+
+A Performance Test
```
cassella: I think this section would benefit from a "tell 'em what you're gonna tell 'em" introduction, covering (or at least mentioning) the steps from performance test to graph: compilation/execution, extraction of key values to ``.dat`` files, and graph generation.

Also a note along the lines of "Though your new checkout doesn't have any ``.dat`` files, the graphs on sourceforge/chapel-lang are generated from existing files.  So if you need to modify any ``.perfkeys`` or ``.graph`` files, please keep in mind the following so as to maintain compatibility: ..."

doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -232,24 +437,9 @@ verify that the last line of output contains the string "SUCCESS":
...
-  ``start_test --performance foo.chpl``
-
-When crawling a directory hierarchy, only tests with ``.perfkeys`` files
-will be considered when testing in performance mode.
-
-All performance tests are compiled with ``--fast`` by default and ``--static``
-when it's not problematic for the target configuration.
-
-
 Accumulating Performance Data in .dat files
```
cassella: I think the description of the ``.dat`` file could be improved by being more precise.  It starts off with "the values collected ... are stored as a tab-delimited ``.dat`` file", coming back a few sentences later to say that each run is its own line.   I think it would be better if it said that in the first place.  I get the impression from the first sentence that the file is overwritten by each run, even though I know that's not how it works.  Maybe move that last sentence of the paragraph to the beginning, and merge it with the first one?

Also, it'd help to understand the graphing to say that the first entry on each line is the date the test was run.

cassella: Mentioning that the date's on each line will help make sense of the example ``graph`` file later,

> ylabel: Time (seconds)

> ylabel: a label for the graph's y-axis (the x-axis will be time by default)

If you can refer to the date the test was run, it'd be clearer that it's not the same "time" on both axes.

cassella:
> I think that's a fair point, but since futures are intended as action items, I think it would be confusing to merge an example into the test system. I can point to a PR with a dummy future like that which will never get merged,

Would it be too detailed to describe in the doc how to create a future for an "It fails on my Mac, but not on Linux" sort of bug?  I don't think it would actually need to be checked in or even on a PR.  But maybe that's too much for this doc.  And you can just work with the submitter to revise their PR.

cassella:
Looks good to me, thanks!

PR #8264, opened 2018-01-21
Upgrade from hwloc 1.11.8 to 1.11.9

ronawho:
This release mostly includes bugfixes/improvements to the x86 backend. I don't
think there's anything we need, but I wanted to keep up to date with versions
to catch any breaking changes sooner.

For a full changelog see:
https://mail-archive.com/hwloc-announce@lists.open-mpi.org/msg00105.html

PR #8981, opened 2018-03-27
Upgrade from hwloc 1.11.9 to 1.11.10

ronawho:
This release includes a number of bugfixes and improvements.

A notable fix that's not mentioned in the release notes is that the check for
sysctl is skipped for linux. sysctl wasn't used for the linux backends, but the
check for it led to it being linked in, which caused linker warnings on some
aarch64 systems: https://github.com/open-mpi/hwloc/commit/58b867ac

For a full changelog see:
https://mail-archive.com/hwloc-announce@lists.open-mpi.org/msg00111.html


ronawho:
@gbtitus similar to #8264, there's probably nothing to review here, so just more of an fyi

dmk42:
Thanks, @ronawho !  This works on aarch64 and makes the advertised improvements.

dmk42:
Because you pointed out to me that this upgrade affects memory binding, I also tried it on KNL which is sensitive to that.  It works there too.

PR #8982, opened 2018-03-27
Compiler represents raw and borrow class types differently

### LINKED ISSUE #8964: delete-free subtask: implement raw/borrowed/owned class types
As a Chapel user, I want to be able to distinguish between class instances that have different memory management discipline, so that I can take advantage of the additional safety checking the compiler can perform.

Acceptance criteria: The locales primer, modified to use `raw Node`, compiles and runs.

In this subtask, I'll extend the compiler to be able to distinguish between different kinds of class types (borrow, owned, raw) that have the same internal representation. This is following from the general direction in #8654.

This task includes:
 * parsing things like `new owned MyClass` and `var x:owned MyClass`
 * getting programs involving such constructs to compile and run

mppf:
Enable the compiler to parse `new unmanaged SomeClass` as well as `new owned SomeClass`.
Inside the compiler's AST representation, create a `unmanaged` type for every class type that is associated with that class. The default class represents a `borrow`. The `borrow` forms the canonical class representation, which is used for most purposes within the compiler. The `unmanaged` class can coerce to the `borrow` class.

This effort has been hindered by the lack of good encapsulation in the compiler's AST classes. In particular, a probably better way to solve this problem would be to introduce a subclass of AggregateType that points to another AggregateType and generates information (such as the list of fields or methods) from it. This is not easily attainable without doing significant work to improve encapsulation of AggregateType.

Instead, this PR adds a new type, UnmanagedClassType, that inherits from Type (and not AggregateType). It refers to an AggregateType. Then, relevant code that was working with AggregateType is updated to work with UnmanagedClassType. The UnmanagedClassTypes are changed back into AggregateType late in callDestructors.

It would also be preferable to transform all AggregateTypes that are borrows into something like BorrowedClassType around the time of resolution. However doing so interferes with a lot of code that is used to working with AggregateType and so this PR does not go that far.

After this PR, the result of 'new SomeClass' continues to be represented as an AggregateType in the compiler. This PR simply considers AggregateType class types to be "borrows" during resolution and lifetime checking through callDestructors. After callDestructors, AggregateType classes should be considered "raw" pointers.

A further wrinkle to explain. The parser changes type construct `owned T` and `shared T` by calling `_owned` or `_shared` but for `unmanaged` they call `_to_unmanaged`. The reason for this is that the type `_unmanaged` is a built-in type and doesn't currently have a type constructor. The simplifying workaround is to implement the type constructor `_to_unmanaged` in module code and then arrange for the parser to call that instead of the typical type constructor name.

Closes #8964.

- [x] full local testing


compiler/AST/AggregateType.cpp
```diff
@@ -2528,3 +2532,87 @@ Symbol* AggregateType::getSubstitution(const char* name) {
...
+
+    at->classKind = CLASS_BORROW;
+    raw->classKind = CLASS_RAW;
+
+    at->defaultClass = at;
+    raw->defaultClass = at;
+    at->nextAssociatedClass = raw;
+
+
+    TypeSymbol* tsRaw = new TypeSymbol(astr("raw ", at->symbol->name), raw);
```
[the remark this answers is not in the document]
mppf: Is changing the keyword to unmanaged sufficient for this? Note that the name of this type symbol is visible to users (e.g. in compiler errors, type to string output). For that reason I think the space is more natural since it matches the syntax we are trying to support.

compiler/codegen/symbol.cpp
```diff
@@ -1213,7 +1213,13 @@ void TypeSymbol::codegenAggMetadata() {
 GenRet TypeSymbol::codegen() {
   GenInfo *info = gGenInfo;
   GenRet ret;
-  ret.chplType = typeInfo();
+  ret.chplType = type;
+
+  // Should not be code generating raw pointers
+  if (AggregateType* at = toAggregateType(type))
```
vasslitvinov: Should probably add "by convention".
To me, all class types in codegen are "unmanaged", not "borrowed".

vasslitvinov: Should probably add "by convention".
To me, all class types in codegen are "unmanaged", not "borrowed".

compiler/passes/normalize.cpp
```diff
@@ -887,6 +889,54 @@ static void processSyntacticDistributions(CallExpr* call) {
   }
 }
 
+/* Find patterns like
+     (new (call manager (call ClassType init-args)))
+     ... where manager might be _owned _to_raw _shared
```
vasslitvinov: It's a bit odd that you use "_to" prefix for unmanaged only. Worth a comment somewhere as to this choice.

compiler/codegen/symbol.cpp
vasslitvinov: [Moving an earlier edited comment to the review]
Should probably add "by convention" to the comment.
Semantically, all class types during codegen are "unmanaged", not "borrowed".

compiler/resolution/callDestructors.cpp
```diff
@@ -1205,6 +1205,57 @@ static void destroyFormalInTaskFn(ArgSymbol* formal, FnSymbol* taskFn) {
   downEndCount->insertBefore(autoDestroyCall);
 }
 
+static void convertRawPointerTypes() {
```
vasslitvinov: How about call it convertClassTypesToCanonical() ?

Also I welcome factoring out the common pattern, so the loops in this function look like:

```c++
forv_Vec(VarSymbol, var, gVarSymbols)
  convertHelp(var->type); // whatever you call the helper
```

compiler/resolution/wrappers.cpp
```diff
@@ -1462,6 +1462,20 @@ static void errorIfValueCoercionToRef(CallExpr* call, ArgSymbol* formal) {
   }
 }
 
+static bool isUnmanagedClass(Type* t) {
+  if (AggregateType* at = toAggregateType(t)) {
+    return at->isUnmanagedClass();
+  }
+  return NULL;
```
vasslitvinov: how about 'return false' instead?

compiler/resolution/wrappers.cpp
```diff
@@ -1462,6 +1462,20 @@ static void errorIfValueCoercionToRef(CallExpr* call, ArgSymbol* formal) {
...
+}
+static bool isBorrowClass(Type* t) {
+  if (AggregateType* at = toAggregateType(t)) {
+    return at->isBorrowedClass();
+  }
+  return NULL;
```
vasslitvinov: also 'return false'

modules/internal/ChapelBase.chpl
```diff
@@ -1882,11 +1885,29 @@ module ChapelBase {
   pragma "no doc"
   inline proc _defaultOf(type t) where t: _sync_aux_t return _nullSyncVarAuxFields;
 
-  pragma "no doc"
+  /*pragma "no doc"
   inline proc _defaultOf(type t) where t: _ddata
-    return __primitive("cast", t, nil);
+    return __primitive("cast", t, nil);*/
```
vasslitvinov: Let's clean this up before merging.
Or leave it commented out for now if you prefer.
If so, I'd prefer having /* and */ on their own lines, for cleaner revision history.

modules/internal/ChapelBase.chpl
```diff
@@ -1882,11 +1885,29 @@ module ChapelBase {
...
+    return rt;
+  }
+  // cast from nil to raw
+  inline proc _cast(type t, x) where t:_unmanaged && x:_nilType {
+    return __primitive("cast", t, x);
+  }
+  // cast from raw to borrow
+  inline proc _cast(type t, x) where t:object && x:_unmanaged {
+    return __primitive("cast", t, x);
+  }
```
vasslitvinov: Let's replace raw -> unmanaged in the above 20-or-so lines.

compiler/AST/UnmanagedClassType.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  if (UnmanagedClassType* mt = toUnmanagedClassType(a))
+    return mt->getCanonicalClass();
+
+  return a;
+}
+
+
+static Type* convertClassTypeToCanonical(Type* t) {
+  // If it's a class, get the canonical class type,
```
vasslitvinov: If you really need this function, it should be just a call to canonicalClasstype(t).
Otherwise the reader will be trying to find differences.

compiler/include/UnmanagedClassType.h
```diff
@@ -0,0 +1,63 @@
...
+class UnmanagedClassType : public Type {
+
+public:
+                          UnmanagedClassType(AggregateType* cls);
+                          ~UnmanagedClassType();
+
+  void                    accept(AstVisitor* visitor);
+  void                    replaceChild(BaseAST* oldAst, BaseAST* newAst);
+  Expr*                   getFirstExpr();
+  Expr*                   getNextExpr(Expr* expr);
```
vasslitvinov: There should not be getFirstExpr/getNextExpr.

compiler/resolution/functionResolution.cpp
```diff
@@ -8564,50 +8782,72 @@ static void removeUnusedTypes() {
...
-  } else if (ct->defaultInitializer                       != NULL &&
-             ct->defaultInitializer->isResolved()         == true) {
+  } else if (at && at->defaultInitializer &&
+             at->defaultInitializer->isResolved()) {
     retval = false;
 
   // FALSE if the type constructor is used.
-  } else if (ct->typeConstructor                          != NULL &&
-             ct->typeConstructor->isResolved()            == true) {
+  } else if (at && at->typeConstructor && at->typeConstructor->isResolved()) {
```
vasslitvinov: I suggest moving at->typeConstructor->isResolved() to a separate line, to match the layout of the other else-if clauses.

bradcray:
Just noting that, at some point, the spec and highlighting modes will need to be updated to reflect the new keywords.  I think it makes sense to drag our feet until we're further along with this feature set, but we shouldn't forget to do it either.  Issue #9088 could be considered to cover the spec part of this but opening a separate issue for the highlighting modes might be wise.

PR #8984, opened 2018-03-27
Fix incorrect documentation of postinit for the initializers technote

lydia-duncan:
The description said that writing a postinit would force the user to write
an explicit init as well for the type.  This was incorrect - the compiler
will generate an init method for the type is none is provided.


bradcray:
Is it OK if I take a stab at this?  I'm worried that it'll be more work for both of us for me to add comments regarding the surrounding paragraphs than it will be to just draft proposed text.

bradcray:
Oops, I missed that it's already been merged (why is it not visually clearer to me on GitHub than it is).  Hmmm...

PR #8988, opened 2018-03-27
Update a slew of tests with initialize() methods to use postinit() instead

lydia-duncan:
Basically a pretty brainless conversion of tests that used initialize() methods
to rely upon postinit() instead.  Not all tests could be converted at this time,
but the vast majority could.

Checked multilocale specific tests with comm=gasnet, rest through a std/
paratest.


lydia-duncan:
Still passes testing.

PR #8992, opened 2018-03-28
Allow specifying a node list to aprun

dmk42:
On a Cray system with no workload manager, `util/test/chpl_launchcmd.py` cannot be used, and therefore there is no way to select nodes listed in `CHPL_LAUNCHCMD_HOSTLIST`.  This change adds the option to specify a list of nodes in `CHPL_LAUNCHER_NODELIST` for aprun.

### DIFF
--- a/runtime/src/launch/aprun/aprun-utils.c
+++ b/runtime/src/launch/aprun/aprun-utils.c
@@ -52,7 +52,8 @@ static char const *aprun_arg_strings[aprun_none] = { "-cc",
                                                      "-n",
                                                      "-d",
                                                      "-N",
-                                                     "-j"};
+                                                     "-j",
+                                                     "-L"};
 
 //
 // Return the appropriate integer value for given argument type
@@ -201,6 +202,22 @@ int getCPUsPerCU() {
   return numCPUsPerCU;
 }
 
+const char *getHostListStr() {
+  return getAprunArgStr(aprun_L);
+}
+char *getHostListOpt() {
+  const char *hostList = getenv("CHPL_HOSTLIST");
+  char *hostListOpt = NULL;
+
+  if (hostList) {
+    hostListOpt = malloc(strlen(getHostListStr()) + strlen(hostList) + 1);
+    strcpy(hostListOpt, getHostListStr());
+    strcat(hostListOpt, hostList);
+  }
+
+  return hostListOpt;
+}
+
 //
 // This function allocates and returns a NULL terminated argument list
 // with the aprun command to be run
@@ -215,6 +232,7 @@ char** chpl_create_aprun_cmd(int argc, char* argv[],
   int largc = 0;
   const char *ccArg = _ccArg ? _ccArg : "none";
   int CPUsPerCU;
+  char *hostListOpt;
 
   initAprunAttributes();
 
@@ -233,6 +251,9 @@ char** chpl_create_aprun_cmd(int argc, char* argv[],
   if ((CPUsPerCU = getCPUsPerCU()) >= 0) {
     sprintf(_jbuf, "%s%d", getCPUsPerCUStr(), getCPUsPerCU());
     largv[largc++] = _jbuf;
+  }
+  if ((hostListOpt = getHostListOpt()) != NULL) {
+    largv[largc++] = hostListOpt;
   }
 
   return chpl_bundle_exec_args(argc, argv, largc, largv);
--- a/runtime/src/launch/aprun/aprun-utils.c
+++ b/runtime/src/launch/aprun/aprun-utils.c
@@ -211,8 +211,8 @@ char *getNodeListOpt() {
 
   if (nodeList) {
     nodeListOpt = chpl_mem_allocMany(strlen(getNodeListStr()) +
-                                     strlen(nodeList) + 1,
-                                     sizeof(char), 0, 0, 0);
+                                     strlen(nodeList) + 1, sizeof(char),
+                                     CHPL_RT_MD_COMMAND_BUFFER, -1, 0);
     strcpy(nodeListOpt, getNodeListStr());
     strcat(nodeListOpt, nodeList);
   }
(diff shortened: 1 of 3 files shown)


dmk42:
@gbtitus - Would you please review?  I'm not sure about having a separate environment variable, but it seemed safer.  I'm open to reusing the existing one if it is safe.

runtime/src/launch/aprun/aprun-utils.c
```diff
@@ -201,6 +202,22 @@ int getCPUsPerCU() {
   return numCPUsPerCU;
 }
 
+const char *getHostListStr() {
+  return getAprunArgStr(aprun_L);
+}
+char *getHostListOpt() {
+  const char *hostList = getenv("CHPL_HOSTLIST");
```
ronawho: Launcher envs should use the `CHPL_LAUNCHER` prefix (see other getenvs in runtime/src/launch)

slurm-srun uses `CHPL_LAUNCHER_NODELIST`, though I'm not opposed to using `CHPL_LAUNCHER_HOSTLIST` if you prefer. 

Also note that you'll want to update pbs-aprun/aprun-utils.c too (at some point we tried making that file a symlink, but we ran into some limitations that I'm not remembering off the top of my head)

dmk42: Thanks.  I'll change it to look like slurm.

runtime/src/launch/aprun/aprun-utils.c
```diff
@@ -201,6 +202,22 @@ int getCPUsPerCU() {
...
 
+const char *getNodeListStr() {
+  return getAprunArgStr(aprun_L);
+}
+char *getNodeListOpt() {
+  const char *nodeList = getenv("CHPL_LAUNCHER_NODELIST");
+  char *nodeListOpt = NULL;
+
+  if (nodeList) {
+    nodeListOpt = sys_malloc(strlen(getNodeListStr()) + strlen(nodeList) + 1);
```
ronawho: You'll actually want to use `chpl_mem_allocMany` instead of `sys_malloc`. chpl_mem_allocMany will map to sys_malloc for launchers, but it's the "user" facing API. Alternatively, I think similar cases in other launchers just use a fixed buffer and truncate or error if the string is too long

runtime/src/launch/aprun/aprun-utils.c
ronawho: Cool, thanks -- it looks like the aprun man page uses the term `node-list`/`node_list` anyways

ronawho: FYI I was curious why we stoped using a symlink -- looks like a few years ago we had some old git clients that couldn't support symlinks correctly or something: https://github.com/chapel-lang/chapel/commit/51ddc2c03cbb6de01bc1bd25f8f90a7055364661 https://github.com/chapel-lang/chapel/commit/9f48237ef1814a9b3dcf9e66b8373067c300b0be

ronawho:
Code looks good to me

runtime/src/launch/aprun/aprun-utils.c
```diff
@@ -201,6 +202,24 @@ int getCPUsPerCU() {
...
 
+const char *getNodeListStr() {
+  return getAprunArgStr(aprun_L);
+}
+char *getNodeListOpt() {
+  const char *nodeList = getenv("CHPL_LAUNCHER_NODELIST");
+  char *nodeListOpt = NULL;
+
+  if (nodeList) {
+    nodeListOpt = chpl_mem_allocMany(strlen(getNodeListStr()) +
```
gbtitus: There's also `chpl_mem_alloc()` as a user-facing memory interface function, for when `sizeof(baseType)==1`.  Personally I like that better for cases where either `number` or `size` is 1, because it short-circuits the deep philosophical question "Is this really 1 item of length <len>, or is it <len> items of length 1?"  (The same occurs in the pbs-aprun/ one.)  You don't have to change this on my account, though.

runtime/src/launch/aprun/aprun-utils.c
```diff
@@ -211,10 +230,11 @@ static char _Nbuf[16];
 static char _jbuf[16];
 char** chpl_create_aprun_cmd(int argc, char* argv[],
```
gbtitus: Not a comment on this PR but just in general: how I wish we hadn't had so much pushback against making incremental improvements to the launchers while maintaining them.  Every time I gaze upon this function I'm tempted to weep.

runtime/src/launch/aprun/aprun-utils.c
```diff
@@ -211,10 +230,11 @@ static char _Nbuf[16];
...
                              int32_t numLocales, const char* _ccArg) {
-  char *largv[8];
+  char *largv[16];
```
gbtitus: Why not 9?  That's all we need.  Making it larger (against future needs?) seems like it will increase the likelihood someone will overrun it because of not realizing they need to expand it.  (Ideally we'd add limit checking, but see "weep" above ...)

runtime/src/launch/aprun/aprun-utils.c
ronawho: Oh, sorry David -- I always forget about `chpl_mem_alloc()` -- feel free to use that if you want

runtime/src/launch/aprun/aprun-utils.c
dmk42: I'll size this to fit.  Sadly, the function would need to be rearchitected for a limit check to even work, since the way it is now, a limit check would have to be at the end, after the undefined behavior has already occurred.

gbtitus:
👍

awallace-cray:
@dmk42 @gbtitus @bradcray, How would you feel about cherry-picking this to the release/1.17 branch? Would help me improve https://github.com/chapel-lang/chapel/issues/9321. Even if we did not use it in an official release.

dmk42:
Having tested this heavily on machines where it is needed, the risk is low.  I'm in favor of cherry-picking this to improve our ability to test the dot release for ARM.

PR #8995, opened 2018-03-28
Use on-clause to compute defaultHash on locale owning string

### LINKED ISSUE #8990: Remote string references can cause problems
### Summary of Problem

We've known for some time that referring to remote strings may not be working as well as it ought to, though it's been awhile since we've run into a problem with it.  This test declares string keys on one locale and then uses them as indices for an associative domain of strings on another locale and has problems, seemingly for this reason.  Namely, the code seems on the surface like it ought to work, yet when run on 2 or more locales, it either gets an OOB error or a segfault.  I believe the issue is related to strings because:

* if the keys are changed to ints, the problem goes away
* if the keys are moved onto the same (remote) locale as the associative domain/array it goes away
* if the intents of the DefaultAssociative.dsiAccess() functions are changed to `const in` it goes away (though I don't think this is what we want, nor that we should require it)

### Steps to Reproduce

[`test/multilocale/strings/assocOfString.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/multilocale/strings/assocOfString.chpl
          ) #8989

### Configuration Information

- Output of `chpl --version`: `chpl version 1.17.0 pre-release (405f366)`

bradcray:
This simple change seems to address the problem in issue #8990
by using an on-clause to ensure that defaultHash is computed on
the same locale where a string is allocated.

This leads me to wonder what we could do to guard against such
accidental "remote meddling in string buffer data" mistakes
in the future (e.g., use a getter/setter for the internal data
that contains locality assertions when --checks are on?)

Fixes #8990.


modules/internal/String.chpl
```diff
@@ -1870,10 +1870,15 @@ module String {
...
-      hash = ((hash << 5) + hash) ^ x.buff[c];
+    var hash: int(64);
+    on __primitive("chpl_on_locale_num",
+                   chpl_buildLocaleID(x.locale_id, c_sublocid_any)) {
+      // Use djb2 (Dan Bernstein in comp.lang.c), XOR version
+      var locHash: int(64) = 5381;
+      for c in 0..#(x.length) {
+        locHash = ((locHash << 5) + locHash) ^ x.buff[c];
+      }
+      hash = locHash;;
```
ronawho: Super nit -- double semicolon

mppf:
I wondered, when looking at this patch, "Why didn't x.buff[c] cause a GET"? It's because the string type uses c_ptr for its internal buffer, and c_ptr is always narrow (because C doesn't have wide pointers). It makes me wonder if string should be using _ddata or some other kind of "wide c pointer".

bradcray:
I'd been hoping to retire _ddata rather than bolster its use...  :(  Though I think putting in some defensive hooks in the short-term (post-release) would probably be wise, i'd be disinclined to mess too much with the internal representation before we take up UTF-8 support (and then wrestle with the best representation at that point).

ronawho:
No 16-node-xc perf differences.

Also with this the reproducer in https://github.com/chapel-lang/chapel/pull/8989 passes 10,000 trials for gasnet-udp (segment fast and everything)

bradcray:
> passes 10,000 trials

Aw man, I only did 1,000 trials...  I'm such a slacker!

bradcray:
@ronawho / @mppf: Would it be fair to consider that one or both of you have reviewed this PR, or do you feel it needs further review?

PR #9004, opened 2018-03-28
Prevent candidate filtering from resolving signatures of incompatible initializers

### LINKED ISSUE #8761: Default initializers and records that store arrays of other records
### Summary of Problem
When a record with a default initializer stores as a field an array of other records with default initializers, we get an internal compiler error when trying to determine the argument type for the argument corresponding to the array field.

This is because when resolving the argument type the first time, we go down through array methods and when we try to initialize the contents, we consider the outer type's initializer as a candidate for the initializer call.  In considering that initializer, we attempt to determine its type from the expressions used to initialize it, but because we haven't finished resolving the array functions yet, their return type is dtUnknown.  We set the type of the array argument to dtUnknown and then try to determine the intent of that argument which triggers the internal compiler error.

If we limit the candidates searched for when resolving initializers to methods on the type we intend to get, we should be able to avoid this problem.  It is unclear whether this error will hide the appropriate error for records storing arrays of themselves, though.

### Steps to Reproduce

**Source Code:**

```chapel
record R {
  var x: int;
}

record StoresR {
  var x: [1..3] R;
}

var storing = new StoresR();
writeln(storing);
```

**Compile command:**
`chpl --force-initializers foo.chpl`

[`test/classes/initializers/compilerGenerated/arrayOfOtherRecs-allDefault.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/compilerGenerated/arrayOfOtherRecs-allDefault.chpl) #8762

### Configuration Information

- Output of `chpl --version`: chapel 1.17.0 pre-release

benharsh:
Our existing resolution infrastructure finds many visible methods named
'init' that we the attempt to filter for 'good' candidates. The majority
of these initializers are not for the type that we're trying to
initialize.

We can get into trouble if we attempt to resolve the signature of a
candidate initializer because that candidate may be in the middle of
being initialized. For example, consider the default initializer for a
record storing an array of another record type:

```chpl
record R {
  var x : [1..3] OtherRec;
}
proc R.init(x = {var temp : [1..3] OtherRec; temp } ) { } // pseudocode
```

Resolution may look like this:
- resolve R.init()
- resolve R.init's formals
- resolve x's default value
- resolve [1..3] OtherRec
- resolve OtherRec.init
  - find visible 'init' methods
  - is 'R.init' a candidate?
    - resolve signature of R.init <---- Recursion!

Instead of attempting to resolve the signature of every candidate
initializer, this commit adds a check to return early if the
initializer's 'this' formal and the corresponding actual are not
instantiated from the same type.

Resolves #8761

Testing:
- [x] full local + futures
- [x] full gasnet

### DIFF
--- a/compiler/include/ResolutionCandidate.h
+++ b/compiler/include/ResolutionCandidate.h
@@ -63,7 +63,7 @@ private:
 
   void                      resolveTypedefedArgTypes();
 
-  bool                      checkResolveFormalsWhereClauses();
+  bool                      checkResolveFormalsWhereClauses(CallInfo& info);
 
   bool                      checkGenericFormals();
 
--- a/compiler/resolution/ResolutionCandidate.cpp
+++ b/compiler/resolution/ResolutionCandidate.cpp
@@ -521,7 +521,9 @@ static bool isCandidateInit(ResolutionCandidate* res, CallInfo& info) {
   AggregateType* ft = toAggregateType(res->fn->_this->getValType());
   AggregateType* at = toAggregateType(info.call->get(2)->getValType());
 
-  if (ft == at || ft->isInstantiatedFrom(at) == true) {
+  if (ft == at) {
+    retval = true;
+  } else if (ft->getRootInstantiation() == at->getRootInstantiation()) {
     retval = true;
   }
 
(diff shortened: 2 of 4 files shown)


benharsh:
@lydia-duncan : would you mind reviewing?

benharsh:
Down from 267 to 262 (possibly due to a variety of PRs since 3/18).

lydia-duncan:
I last ran it on the branch Wednesday, so it's probably your change :)

Issue #8820, opened 2018-03-14
Update chpl_arch.py to include ARM processors


dmk42:
As a Chapel implementor, I want to have all the architecture-selection code in one place, chpl_arch.py, because it is easier to make it correct that way, and easier to maintain it.

Acceptance criteria:
ARM-specific code in compiler/util/clangUtil.cpp introduced in #8696 is no longer needed.

Note:  We also need to translate mic-knl to knl.

dmk42:
This was addressed in #9013 .

Issue #5686, opened 2017-03-16
Maintain chpl_arch.py better


ronawho:
[util/chplenv/chpl_arch.py](https://github.com/chapel-lang/chapel/blob/master/util/chplenv/chpl_arch.py) maintains a mapping of architectures to cpu ISA targeting. e.g. setting `CHPL_TARGET_ARCH=broadwell` will result in targetting `core-avx2` under intel and either `core2`, `core-avx2`, or `broadwell` under gnu (depending on the gcc version)

Unfortunately, we haven't done a good job of maintaining these mappings. For example, we're missing mappings for newer gcc's and we're missing mappings for KNL, the latest AMD chips, and probably some other things. 

For more info see the original commit that added architecture specialization: 1c3bf118717110672d4965ee0eb5ec9fc28a7c90

ronawho:
@ben-albrecht can we add an `area: chplenv` label?

ben-albrecht:
Seems oddly specific, but might be valuable. Are there other issues that would fit it? 

Will have to discuss with the **@chapel-lang/labels-team**

ronawho:
Wow.. I didn't realize we had a labels team.

ronawho:
I'd also be ok with something more generic like `util` to cover things like chplenv scripts, util/config stuff, buildRelase stuff, etc. I don't think we have a good label to characterize those things now. 

But TBH I'm not interested in starting another labels discussion, so if there's any opposition, I'll just drop it.

ronawho:
Actually, I don't know what I was thinking by asking for a new label -- gonna delete all my comments in hopes that nobody else sees them :)

ronawho:
@dmk42 I think you follow gcc releases and upcoming architectures pretty closely. Is this something you'd be interesting in maintaining?

dmk42:
@ronawho You're right, I pretty much have to keep up with that information anyway, so I would be a logical person to maintain that file.  I'm willing to sign up for that.

dmk42:
This was addressed in #9013 .

PR #9013, opened 2018-03-29
Update chpl_arch.py to handle 64-bit ARM, KNL, and Skylake

dmk42:
ARM presents a challenge for `chpl_arch.py` because Intel architectures are accessed with the `-march=` compiler flag, and ARM architectures are accessed with the `-mcpu=` flag.  This change includes an indication of which flag to use in the output of `chpl_arch.py` and updates it to handle the relevant ARM architectures.

This allowed the removal of some special cases from `clangUtil.cpp`.

Also, KNL and Skylake have been added.

Makefiles had to be changed to expect `CHPL_MAKE_TARGET_BACKEND_ARCH` to include part of the flag.  Some Python scripts had to be changed to stop after the first `=` when parsing environment variables.

This PR addresses issues #5686 and #8820 .


dmk42:

dmk42:
I'm labeling this a work in progress until we figure out how we want to handle splitting the (arch|cpu)=architecture strings.

dmk42:
@ronawho - This is now reimplemented to return a tuple and verified working on a wide variety of architectures, so it's ready for re-review.

util/chplenv/chpl_comm_substrate.py
```diff
@@ -15,7 +15,6 @@ def get():
     if not substrate_val:
         comm_val = chpl_comm.get()
         platform_val = chpl_platform.get('target')
-        arch_val = chpl_arch.get('target', get_lcd=True)
```
ronawho: Why is this here?

ronawho: Oh, nevermind. I see it's dead code  (that I missed removing in https://github.com/chapel-lang/chapel/commit/29ec9bce3d2b30cd15d84c2d0d07e8c3b806f5d0#diff-45c544a18723ce3d378babf320393a0fL28)

ronawho:
Overall, this looks good to me. 

I wonder if we could use a shorter name than `CHPL_TARGET_BACKEND_ARCH_FLAG`. Maybe just `CHPL_TARGET_ARCH_FLAG`

I wonder if it would be cleaner to use a named tuple. And that way you could do things like: `chpl_arch.get('target').arch` instead of manually detupling

@ben-albrecht may be interested in reviewing some of the chplenv infrastructure changes (though probably not the chpl_arch.py changes)

util/chplenv/third_party_utils.py
```diff
@@ -15,10 +15,11 @@
 #
 @memoize
 def default_uniq_cfg_path():
+    (flag, arch) = chpl_arch.get('target', map_to_compiler=True,
```
ronawho: I would go with `(_, arch)` anywhere the `flag` isn't going to be used. Either that or maybe have chpl_arch return a named tuple so that you could just do `chpl_arch.get(...).arch`

ben-albrecht:
> @ben-albrecht may be interested in reviewing some of the chplenv infrastructure changes (though probably not the chpl_arch.py changes)

LGTM

PR #9018, opened 2018-03-29
Support dynamic dispatch on vararg methods

daviditen:
When a subclass overrides a vararg method defined by a parent class, dynamic
dispatch wasn't happening, so a call to the method through a variable that is
dynamically the subclass type but statically the parent class type would call
the parent version of the method.

When building up the list of methods for a parent class method dispatch set,
expand the varargs of the subclass methods to match the arguments the parent
method was vararg-expanded to.

This needed to look at calls to the parent method in order to expand the
subclass methods, so I moved most of the body of `compute_call_sites()` into
a new function `compute_fn_call_sites(FnSymbol*)`, to allow computing the call
sites for a single function.

Stop removing FLAG_EXPANDED_VARARGS in FnSymbol::finalizeCopy().
FLAG_EXPANDED_VARARGS is useful information even after a generic function is
instantiated.  It shouldn't be removed from a function that was stamped out
from a varargs function.

Added three tests taken from user-issues requesting this functionality.


daviditen:
This has passed a full standard paratest.

mppf:
It's probably worth running a --verify test and/or a --baseline test, at least with release/examples/primers.

daviditen:
Thanks @mppf! I just started a --verify run, and will merge once it's completed assuming no surprises.

daviditen:
The full --verify run was clean.  Running `start_test -compopts --baseline` from test/release/examples was clean.

PR #9030, opened 2018-03-30
Add CHPL_INCLUDE_PATH and colon paths to CHPL_LIB_PATH, -L, and -I flags

### LINKED ISSUE #8889: Add multiple-path support to CHPL_LIB_PATH
`CHPL_LIB_PATH` accepts one library-directory linker path. I would like formal support for multiple paths:

```bash
# Workaround for today's Chapel
export CHPL_LIB_PATH="/usr/lib -L/usr/lib64 -L/opt/lib"

# Example of what I want with path delimiters
export CHPL_LIB_PATH="/usr/lib:/usr/lib64:/opt/lib"

# --> chpl <source> -L/usr/lib -L/usr/lib64 -L/opt/lib
```

### LINKED ISSUE #8890: Add multi-path support for new environment variable CHPL_INCLUDE_PATH
1. Add new environment variable `CHPL_INCLUDE_PATH` that mimics the functionality of `CHPL_LIB_PATH` but for include/header directories.

2. `CHPL_INCLUDE_PATH` should support multiple paths with path delimiter. Related #8889 

```bash
# Example
export CHPL_INCLUDE_PATH="/usr/include:/opt/include"

# ---> chpl <source> -I/usr/include -I/opt/include
```

bradcray:
This PR addresses issues #8889 and #8890 by adding support for multiple paths to be specified via CHPL_LIB_PATH and CHPL_INCLUDE_PATH using `:` as a separator.  Previously, there was no official way to specify multiple paths like this, though `lib/path1 -L lib/path2 -L lib/path3` could be used as a lame workaround.  Because the implementation of our flag parsing is shared between both environment variables and command-line flags, this also has the effect of permitting `-L` and `-I` flags provided to the compiler to specify multiple `:`-separated paths for better or worse (I don't think it's ideal, but don't think it's worth re-implementing the command-line parsing over this, at least at present).

Updated the man page to reflect these changes.

I've also added some tests that link in C libraries in various ways via flags and the environment variables to lock in this behavior.  The tests use `gcc` and `ar` to make the libraries, so are currently `skipif`ed if `CHPL_TARGET_COMPILER` != gnu.

Fixes #8889.
Fixed #8890.


compiler/util/files.cpp
```diff
@@ -69,17 +69,36 @@ static const char* intDirName        = NULL;
...
 
-void addLibInfo(const char* libName) {
+//
+// Convert a libString of the form "-Lfoo:bar:baz" to entries in libFlag[]
+//
+void addLibInfo(const char* libStringOrig) {
   static int libSpace = 0;
+  const char* flag = "";    // do we need a '-L' flag?  (initially, no)
+  char* colon;              // used to refer to ':'s in libString
+  int libStringLen = strlen(libStringOrig)+1;
```
gbtitus: Is there a strong reason for not just using `strdup()` here?  It doesn't look like libStringLen is needed other than to drive the `malloc()` and `strncpy()`.

gbtitus:
A post-review `grep` driven by curiosity about a different question led to the discovery that unlike `CHPL_MODULE_PATH`, `CHPL_LIB_PATH` isn't documented at all.  Should it be?

compiler/util/files.cpp
bradcray: Just my own ignorance, thanks for the suggestion!

bradcray:
I'll look into the documentation question.  Generally, we rely on the environment section of the man page to help people find the environment variables backing flags, but you're right that the module path isn't handled consistently.  Let me take a look at what we have and try to improve it.  The point about include path was pointed out in [a separate issue](https://github.com/chapel-lang/chapel/issues/8890) and that was going to be my next spare-time fix.

bradcray:
Greg: On reflection, I decided to tackle issue #8890 in this PR as well, and made some progress there, but am not done yet.  Just noting this so you don't spend time on a re-review too early.

PR #9033, opened 2018-03-30
Add a future for the enum array default initializer bug, and a test that works

lydia-duncan:
When converting initialize methods to use postinit, I discovered an error with a
record containing an enum array.  If the record utilized a default initializer
and default copy initializer, copies of the instance would not appropriately
copy the enum array field, leading to double deletes and accessing freed memory.
Explicit initializers do not seem to trigger this problem, oddly enough.

This commit adds a future for the error case and locks in the explicit
initializer case.  (Though my local machine was not as reliably reproducing the
error so we'll see if the "passing" version actually works.)

Testing:
- [x] valgrind over classes/initializers with futures


lydia-duncan:
Note: the passing version did work with valgrind on the machine where the original had failed, so I'm not worried about it (I just forgot to update the merge message, whoops!)

PR #9043, opened 2018-03-30
Update revcomp problem size; filenames for fasta-based inputs

bradcray:
The CLBG site changed the problem size of revcomp for the production run.  This catches us up to that size while also doing some refactoring.  Specifically:

* revcomp problem size has been increased to 100000000 and the revcomp perfkeys files have been updated to reflect the new expected output
* I changed from all the per-test .preexec files within a shootout directory to a directory-wide PREEXEC file
* I renamed the `fasta.` files based on the name of the benchmark using them for improved clarity and updated the `.perfexecopts` to reflect the new filenames


bradcray:
@benharsh: I'd already switched the `fasta.*` input filenames to this scheme before you mentioned having them embed the problem size in their names, which I also like.  Thought I'd have you take a look at this and let me know which you prefer before going further with it).

Once we agree on an approach, I'll push similar mods to the test/studies directories.

benharsh:
This generally looks good to me. I'd advocate for a comment in the PREEXEC stating the problem size in english or scientific notation. E.g., 100000000 is "100 million" or "1e8".

bradcray:
And take away our opportunities to mock @ronawho?

bradcray:
Like this?

benharsh:
Looks great, thanks!

PR #9044, opened 2018-03-31
Call postinit on declared generic record

### LINKED ISSUE #9001: postinit not applied for type declaration when record is generic
### Summary of Problem
When a record instance is declared with just a type (no explicit initial value), `postinit` is not called.

### Steps to Reproduce

**Source Code:**

```chapel
record MyRec {
  type t;
  var x : t;

  proc init() {
    this.t = int;
    this.x = 10;

    writeln('MyRec.init default');
  }

  proc init(type t) {
    this.t = t;

    writeln('MyRec.init type');
  }

  proc postinit() { // This should get run, but accidentally is skipped today
    writeln('MyRec.postinit');
    writeln();
  }
}

proc main() {
  var r1 : MyRec(int);

  writeln(r1);
}
```

generates this output:
```
MyRec.init type
(x = 0)
```

**Compile command:**
`chpl foo.chpl`

**Execution command:**
`./foo`

[`test/classes/initializers/postInit/record-generic-fails.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/initializers/postInit/record-generic-fails.chpl) #9002

### Configuration Information

- Output of `chpl --version`: Chapel 1.18.0 pre-release

benharsh:
Issue #9001 exposed a bug where postinit would not be called on a declared generic record, like:
```chpl
var x : R(int);
```

This commit:
- adds the postinit call where appropriate
- updates a .good file with correct output
- converts a test using initialize() to use postinit, a task blocked by this bug

Resolves #9001 

Testing:
- [x] full local

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -9346,6 +9346,13 @@ static Expr* resolvePrimInit(CallExpr* call, Type* type) {
     resolveCallAndCallee(defOfCall);
 
     retval = foldTryCond(postFold(defOfCall));
+
+    if (at && at->isRecord() && at->hasPostInitializer()) {
+      CallExpr* move = toCallExpr(defOfCall->parentExpr);
+      INT_ASSERT(move->isPrimitive(PRIM_MOVE));
+      SymExpr*  var  = toSymExpr(move->get(1));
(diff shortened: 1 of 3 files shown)


benharsh:
I know about the pre-existing double-init in ``record-generic.chpl``, and will tackle it eventually.

PR #9054, opened 2018-04-02
Default Initializers: Parent formals before child formals

### LINKED ISSUE #9056: Default Initializers: Use parent param in child type expression
### Summary of Problem
A compiler-generated initializer fails to handle a case where a derived class field uses a parent param field in its type expression.

### Steps to Reproduce

**Source Code:**

```chapel

class Dummy {
  param stridable : bool;
}

class Parent {
  param rank : int;
  type idxType;
  param stridable : bool;
}

class Child : Parent {
  var x : Dummy(stridable);
}

var c = new Child(1, int, false);

```
Compiler output:
```
foo.chpl:13: error: use of 'stridable' before encountering its definition, type unknown
```

benharsh:
This PR updates default initializers such that the formals for
parent fields precede the formals for child fields. This matches the
existing behavior for constructors and matches the flow of
initialization now that parents are initialized before children.

To accomplish this, it was first necessary to remove some incorrect
error messages claiming that parent fields could not be accessed in
phase one. These were likely left over from a time before the new
initializer proposal was implemented.

Next it was necessary to update 'resolveMaybeSyncSingleField' to update
symbols in the field's type expression to refer to the instantiated
parent field instead of the fully generic field. This is accomplished by
adding a new field to AggregateType, 'genericFieldMap', which maps fully
generic fields from the root instantiation to concrete fields in another
instantiation. This also maps param fields to their computed value.
This map is built up as instantiations are created.

A handful of tests have been updated to reflect this new argument ordering.

Resolves #9056 

Testing:
- [x] full local


benharsh:
Testing with --force-initializers exposed ~245 errors, down from ~265 on master.

compiler/resolution/functionResolution.cpp
```diff
@@ -4675,7 +4675,15 @@ static void resolveMaybeSyncSingleField(CallExpr* call) {
     DefExpr*   tmpDefn   = new DefExpr(tmp);
 
     // Applies a type to TMP
-    CallExpr*  tmpExpr   = new CallExpr(PRIM_INIT, fieldDef->exprType->copy());
+    // BHARSH TODO: Why can't we just use the exprType of the formal? Shouldn't
+    // it be the same?
```
lydia-duncan: We can't have the same AST inserted in two places in the code.  It shouldn't particularly matter whether we copy from the field definition or the argument we just made, though.

compiler/resolution/functionResolution.cpp
```diff
@@ -4675,7 +4675,15 @@ static void resolveMaybeSyncSingleField(CallExpr* call) {
...
+    Expr*      typeExpr  = fieldDef->exprType->copy();
+    if (ct->isClass()) {
+      // Update symbols
+      update_symbols(typeExpr, &ct->getGenericFieldMap());
```
lydia-duncan: So this is only important in the case of inheritance?  The code you inserted in AggregateType doesn't distinguish between classes and records, unless I missed something

benharsh: I didn't distinguish between classes and records in AgreggateType just to keep things simple and general-purpose, in case we find a use for this map later. I suspect we may need it in other cases, but can't point to anything in particular.

compiler/resolution/functionResolution.cpp
benharsh: I quickly noticed that the corresponding formal is not readily available, so I'll leave this as a future TODO.

lydia-duncan: Okay, cool.  I think this is good to go, then!

Issue #8969, opened 2018-03-26
Investigate performance regressions from Meltdown/Spectre patches


ronawho:
We saw some non-trivial performance regressions for 16-node-xc around the time patches for the Meltdown and Spectre were applied: https://chapel-lang.org/perf/16-node-xc/?startdate=2018/02/21&enddate=2018/03/12&graphs=npbepperfmopssized,hpccraatomicsperfgupsn233,hpccrarmoperfgupsn233,hpccglobalstreamperfgbsn5723827200,emptyremotetaskspawntime

The two most concerning cases that were impacted were RA atomic/rmo and stream-global. These are particularly concerning because the reference version wasn't impacted. Our numbers look fine in release-over-release timings, but we are now performing worse than the reference C+MPI+OpenMP versions when we were on par before.

When investigating we should probably start with stream-global since it's the simplest benchmark. I don't think we're doing many (any?) kernel calls for stream-global's timed section, so I was surprised to see the performance regression (though I don't fully understand that implications of the meltdown/spectre patches yet.)

Since stream-ep wasn't really impacted my guess is that our coforall+on idiom was hurt, but the [coforall+on microbenchmark](https://chapel-lang.org/perf/16-node-xc/?startdate=2018/02/21&enddate=2018/03/12&graphs=emptyremotetaskspawntime) wasn't hurt under ugni. The microbenchmark probably ends up using small AMs, so it's possible that large AMs (which will be used for stream-global) were hurt. It's probably worth beefing up that benchmark to include a case that will trigger large AMs. 

When investigating this we should also try running on newer hardware (skylake or newer) to see if the performance impacted is smaller on newer hardware.

ronawho:
It looks like interference/contention from the progress thread is much more significant with the patches. Maybe context switching is slower now, and so the pthread/worker hosting the last task (that shares a core with the progress thread) is slowed down even further?

Fortunately we had already been looking into progress thread interference: https://github.com/chapel-lang/chapel/pull/8562

Running with `CHPL_RT_COMM_UGNI_BLOCKING_CQ=y,` I see even more significant improvements than before, and more importantly the [regressions](https://chapel-lang.org/perf/16-node-xc/?startdate=2018/02/21&enddate=2018/03/12&graphs=npbepperfmopssized,hpccraatomicsperfgupsn233,hpccrarmoperfgupsn233,hpccglobalstreamperfgbsn5723827200,emptyremotetaskspawntime) noted above are resolved.

[...]

PR #9068, opened 2018-04-02
Reduce contention from ugni's polling thread

### LINKED ISSUE #9067: Limit interference from the progress thread
We currently run a single progress thread as a separate pthread that's pinned to the "last" CPU. For ~~ugni and~~ gasnet we run it as an "active" poller in the sense that we do something like:

```c
while (run_poller) {
  check_for_new_am(...);
  chpl_task_yield();
}
```

Actively polling can create non-trivial interference/contention with tasks running on the same CPU. This is true even if there's no incoming AMs, since the poller keeps spinning and getting rescheduled even if there's no work to do. The slowdown created by the polling thread was increased with spectre/meltdown patches (we believe because context switches are now more expensive): https://github.com/chapel-lang/chapel/issues/8969

For ugni we have the option of running the progress thread in a blocking mode (#8562). This will slightly increase the latency of AMs, but it significantly reduces the contention from the polling thread. **Edit** default changed to blocking in #9068

For gasnet we'll need to come up with a separate strategy since I don't think there's any way to do a "blocking" `gasnet_AMPoll()`. Possible solutions might be to run 2 progress "threads" -- one as a low-priority pthread that only runs every ~100 milliseconds and a second that's run in a qthread. The qthread one can run at full speed when there's no other tasks running, and the pthread one can ensure AMs are still delivered even all the qthreads are busy doing user work. We could also just experiment with continuing to run a single pthread, but with lower priority and no pinning. It's probably also worth pinging the GASNet team to see if they have any thoughts on this (maybe EX adds a way to do a blocking poller?)

 - [x] limit interference for ugni (resolved in #9068)
 - [ ] limit interference for gasnet

ronawho:
This updates ugni to use a blocking CQ by default for the progress thread.
Instead of actively polling for new events in a loop we do a CqWaitEvent to
block and wait for a completion event to be generated. Really, this just
enables the feature added in #8562 by default.

Spectre/meltdown patches have increased the cost of context switches (#8969),
which exacerbates the contention between the progress thread and a worker/task
that's running on the same core.

This will resolve the performance regressions we saw after the spectre/meltdown
patches were applied, and in some cases performance will be better than before.
We saw performance improvements when we originally added this as an opt-in
feature but there are some regression for coforall+on microbenchmarks, so we
weren't quite ready to enable it before. However, given how much more
interference/contention there is now, it seems worth enabling.

In short, this will reduce contention from the polling thread, but it does
slightly increase the latency for active messages. This means that tests that
create a lot of of AMs without doing much work (coforall+on mirco-benchmark)
will slow down slightly, but most benchmarks will improve. Here's a brief-ish
summary:

| Benchmark              | Perf gain       |
| ---------------------- | --------------- |
| CoMD elegant           | 10% improvement |
| CoMD llnl              | 30% improvement |
| NPB EP                 | 90% improvement |
| RA-rmo                 | 20% improvement |
| PRK stencil block      | 70% improvement |
| PRK stencil stencil    | 70% improvement |
| stream-ep              |  2% improvement |
| stream-global/prom     | 10% improvement |
| single large reduction | 20% improvement |

| Benchmark              | Perf loss       |
| ---------------------- | --------------- |
| coforall+on microbench | 90% regression  |
| coforall+on net AMO    | 15% regression  |
| many small reductions  | 25% regression  |
| lulesh                 | 10% regression  |
| hpl                    |  5% regression  |

This partially resolves #9067 (resolves for ugni, but not gasnet)

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1383,7 +1383,7 @@ static volatile chpl_bool polling_task_running     = false;
 static volatile chpl_bool polling_task_please_exit = false;
 static volatile chpl_bool polling_task_done        = false;
 
-static chpl_bool polling_task_blocking_cq = false;
+static chpl_bool polling_task_blocking_cq = true;
 
 
 //
@@ -2731,7 +2731,7 @@ void set_up_for_polling(void)
     uint32_t cq_mode = GNI_CQ_NOBLOCK;
 
     polling_task_blocking_cq = chpl_env_rt_get_bool("COMM_UGNI_BLOCKING_CQ",
-                                                    false);
+                                                    polling_task_blocking_cq);
 
     if (polling_task_blocking_cq)
       cq_mode = GNI_CQ_BLOCKING;
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1383,7 +1383,7 @@ static volatile chpl_bool polling_task_running     = false;
 static volatile chpl_bool polling_task_please_exit = false;
 static volatile chpl_bool polling_task_done        = false;
 
-static chpl_bool polling_task_blocking_cq = true;
+static chpl_bool polling_task_blocking_cq;
 
 
 //
@@ -2731,7 +2731,7 @@ void set_up_for_polling(void)
     uint32_t cq_mode = GNI_CQ_NOBLOCK;
 
     polling_task_blocking_cq = chpl_env_rt_get_bool("COMM_UGNI_BLOCKING_CQ",
-                                                    polling_task_blocking_cq);
+                                                    true);
 
     if (polling_task_blocking_cq)
       cq_mode = GNI_CQ_BLOCKING;


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2731,7 +2731,7 @@ void set_up_for_polling(void)
     uint32_t cq_mode = GNI_CQ_NOBLOCK;
 
     polling_task_blocking_cq = chpl_env_rt_get_bool("COMM_UGNI_BLOCKING_CQ",
-                                                    false);
+                                                    polling_task_blocking_cq);
```
gbtitus: I would actually prefer `true` here and no initializer on the variable decl.  The latter is unneeded unless there's a path through execution that doesn't pass through this `getenv` call, and the former would make the default easier for a reader to see.  Otherwise you have to go backwards 1400 lines to find it.

ronawho: Yeah, good call. I was originally thinking -- "Oh, it's nice to see the default at the declaration site so you don't have to go through the code to find it", but in hindsight I think you're right.

PR #9072, opened 2018-04-03
Use Chapel-installed virtualenv for ZMQ interop testing

ben-albrecht:
Previously, ZMQ interop testing required a system-installed `virtualenv` to work, which is not necessarily a requirement for Chapel. This PR modifies the `PREEXEC` to use the Chapel-installed `virtualenv` if it can be found, otherwise it tries to find a system-installed version. If that is not found, the test will fail.

Note the Chapel-installed `virtualenv` may not exist if `CHPL_TEST_VENV_DIR` is set to `none`, for example.

Other changes:

- Clean up virtualenv
- Print an error message if the virtualenv is not built, which should trigger a clear failure instead of a timeout or success
- Use the `CHPL_PIP_INSTALL_PARAMS` to stay consistent with how `test-venv` packages are installed.

### DIFF
--- a/test/library/packages/ZMQ/interop-py/CLEANFILES
+++ b/test/library/packages/ZMQ/interop-py/CLEANFILES
@@ -1,0 +1,1 @@
+pyzmq-venv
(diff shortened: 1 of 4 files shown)


ben-albrecht:
- Confirmed testing passes locally, with Chapel-installed `virtualenv` and system-installed `virtualenv` (`CHPL_TEST_VENV_DIR=none`)
- This test still fails during the PyPI brownout period (first 15 minutes of the hour). That won't be resolved until the virtualenv version is updated.

PR #9076, opened 2018-04-03
Convert docs/latest URLs to docs for simplicity

bradcray:
This is mostly cosmetic, reflecting that we made chapel-lang.org/docs/ equivalent to .../docs/latest but may allow us to retire the docs/latest/ link eventually.  In one case, I also fixed an absolute URL to use a Chapel module cross-reference instead.

### DIFF
--- a/doc/rst/language/archivedSpecs.rst
+++ b/doc/rst/language/archivedSpecs.rst
@@ -44,7 +44,7 @@ Spec *<version>* (Chapel *<release in which specification was made available>*)
 
 
 
-.. _Language Specification:     https://chapel-lang.org/docs/latest/language/spec.html
+.. _Language Specification:     https://chapel-lang.org/docs/language/spec.html
 .. _Spec 0.981 (Chapel 1.13.0): https://chapel-lang.org/docs/1.13/language/spec.html
 .. _Spec 0.98  (Chapel 1.12.0): https://chapel-lang.org/spec/spec-0.98.pdf
 .. _Spec 0.97  (Chapel 1.11.0): https://chapel-lang.org/spec/spec-0.97.pdf
--- a/doc/rst/technotes/initializers.rst
+++ b/doc/rst/technotes/initializers.rst
@@ -740,7 +740,7 @@ types use initializers when possible.  Please report any bugs
 encountered using the guidance described at the `bugs`_ page.
 
 .. _bugs:
-   https://chapel-lang.org/docs/latest/usingchapel/bugs.html
+   https://chapel-lang.org/docs/usingchapel/bugs.html
 
 
 
--- a/doc/rst/technotes/local.rst
+++ b/doc/rst/technotes/local.rst
@@ -36,10 +36,8 @@ the performance of a program, as communication usually slows down
 execution.
 
 The ``local`` construct does not necessarily indicate the cause of
-communication when present. See the `Module\: CommDiagnostics`_ for ways to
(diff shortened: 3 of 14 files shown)


doc/rst/technotes/local.rst
```diff
@@ -39,7 +39,7 @@ The ``local`` construct does not necessarily indicate the cause of
 communication when present. See the `Module\: CommDiagnostics`_ for ways to
 diagnose communication.
 
-.. _Module\: CommDiagnostics:    https://chapel-lang.org/docs/latest/modules/standard/CommDiagnostics.html
+.. _Module\: CommDiagnostics:    https://chapel-lang.org/docs/modules/standard/CommDiagnostics.html
```
bradcray: @ben-albrecht : For each release, should this be pointing to that releases current version of the CommDiagnostics documentation rather than the "latest" version of that documentation?

ben-albrecht: Yes, but we shouldn't be cross-referencing documentation with full URLs like this in the first place. This should be accomplished with a sphinx cross-reference, in this case a [Chapel domain module reference](https://sphinxcontrib-chapeldomain.readthedocs.io/en/latest/reference.html#role-chpl:mod):

```rst
See the :chpl:mod:`Module: CommDiagnostics <CommDiagnostics>` for ways to diagnose communication.
```

doc/rst/technotes/initializers.rst
```diff
@@ -740,7 +740,7 @@ types use initializers when possible.  Please report any bugs
 encountered using the guidance described at the `bugs`_ page.
 
 .. _bugs:
-   https://chapel-lang.org/docs/latest/usingchapel/bugs.html
+   https://chapel-lang.org/docs/usingchapel/bugs.html
```
ben-albrecht: In this case, I think it's OK to stick to the latest version, since we'll always want users to get the latest bugs page

doc/rst/technotes/local.rst
bradcray: If only there were some way to know who injected this in the first place...  :)

bradcray:
@ben-albrecht : Since you've already reviewed the most thought-provoking parts of this PR, would you be open to reviewing the rest of it?

ben-albrecht:
Sure thing.

doc/rst/language/archivedSpecs.rst
```diff
@@ -44,7 +44,7 @@ Spec *<version>* (Chapel *<release in which specification was made available>*)
 
 
 
-.. _Language Specification:     https://chapel-lang.org/docs/latest/language/spec.html
+.. _Language Specification:     https://chapel-lang.org/docs/language/spec.html
```
ben-albrecht: This could arguably be a cross-reference to the current version's spec rather than latest using a `:ref:`

bradcray: I'm not necessarily opposed to that, except that the text just above it says "Latest Spec Version" which seems like it could be unnecessarily confusing if it were pointing to the latest version as of some past release date.  (It's also excerbated by the fact that the link to this page scrolls down just enough, in my Chrome browser at least, that it hides the little color-coded version menu).  Try this link to see what I mean, both in terms of the scrolling and what the wording implies if you just found yourself on this page:  https://chapel-lang.org/docs/1.14/language/archivedSpecs.html#chapel-archived-specs  

Thoughts?

PR #9078, opened 2018-04-04
Prevent path splitting when fixing path on fish (#9049)

### LINKED ISSUE #9049: fixpath.py doesn't handle PATH that contains space on fish shell correctly
https://github.com/chapel-lang/chapel/blob/master/util/config/fixpath.py#L43

If one entry of `PATH` contains space, `fixpath.py` will treat it as multiple entries, because of space being used as the delimiter. Therefore, after `source ./util/setchplenv.fish`, `PATH` might become invalid.

caizixian:
Say, my `PATH` looks like `"/foo bar/42" "/usr/bin"` on fish. Fish will export `/foo bar/42:/usr/bin` to subcommands. The result of splitting in`fixpath.py` will be `["/foo bar/42", "/usr/bin"]`, and the result of join by delim is `/foo bar/42 /usr/bin`, which will be consumed by, for example `setchplenv.fish`. Fish will then treat this as `"/foo" "bar/42" "/usr/bin"`

The solution I proposed is to double quote each entry of `PATH`, so that `setchplenv.fish` can directly pass them to fish shell

Closes #9049

### DIFF
--- a/util/config/fixpath.py
+++ b/util/config/fixpath.py
@@ -6,6 +6,11 @@ from os import getenv
 from sys import argv
 import re
 
+def escape_path(p):
+    if len(argv) >=3 and argv[2] == 'fish':
+        # Suppress splitting of fish so that the path will be treated as a whole
+        return '"{}"'.format(p)
+    return p
 
 def main(env='PATH', delim=':'):
     """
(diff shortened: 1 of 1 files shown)


util/config/fixpath.py
```diff
@@ -25,7 +30,7 @@ def main(env='PATH', delim=':'):
     pattern = r'(?<!\\)\:'
```
ben-albrecht: This change may be a simpler way to accomplish the same thing:

```python
 pattern = r'(?<!\\)\{0}'.format(delim)
```

ben-albrecht: Ah, I am sorry I misunderstood the original issue (not a fish user, as you can probably tell)

ben-albrecht:
@caizixian - Thanks for contributing this fix. 

I don't see a CLA on file for you, which we'll need in order to merge. Could you submit one?

[Here are the instructions](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements/)

caizixian:
@ben-albrecht I've sent the CLA.

PR #9080, opened 2018-04-04
tidy up in remoteValueForwarding

vasslitvinov:
Minor removal of seemingly-obsolete code.
Convert 'ref' intent to 'in' intent of the formal that underwent RVF.

Testing:
- [x] linux64 --verify
- [x] gasnet


vasslitvinov:
@benharsh what do you think?
It passes testing, so should be good?
P.S. += on string now takes both arguments by ref.

PR #9096, opened 2018-04-05
Enable promotion on inheriting classes with initializers

benharsh:
Some helper methods on AggregateType that create new instantiations were
incorrectly removing FLAG_GENERIC, which fooled some of the existing
promotion resolution machinery into not running. This commit updates
those helper methods to preserve FLAG_GENERIC if there are still generic
fields in the instantiation.

This commit also adds a call to resolvePromotionType for the case that
the parent is generic but the child is not.

A new test is added to verify this functionality.

Testing:
- [x] full local + futures


benharsh:
This is a partial blocker on the BaseDom-initializers effort. The FLAG_GENERIC changes are necessary to perform some operations on that branch, and it was convenient to add the call to resolvePromotionType at the same time.

benharsh:
@noakesmichael : Would you be willing to review?

PR #9098, opened 2018-04-05
Housekeeping for Cray-specific tests

dmk42:
This cleans up some items found in testing for Cray XC50 with ARM compute nodes.

The environment variable `CRAY_LIBSCI_PREFIX_DIR` was being used explicitly in compopts files to find the cray-libsci headers and libraries, but these are found automatically by the PrgEnv modules when cray-libsci is loaded.  The environment variable is not always kept up to date by cray-libsci when other modules are loaded, so sometimes we were pointing to the wrong place when the compiler would have otherwise found the right place.  I removed our uses of this environment variable to let the PrgEnv and cray-libsci modules cooperate to do the work for us.  (We can depend on the environment variable's existence, but not its contents.)

Some tests were being skipped on Cray XC systems running CLE 6 because the file that reports the operating system version has moved.  Also, the tests for this checked the directory above first.  They only needed to check whether the file itself existed.  I've simplified the tests and added a search for the CLE 6 version of the file so these tests will now begin running again on Cray XC systems.

Tested on both Cray XC systems, where these tests should run, and whitebox systems, where they should not run.

### DIFF
--- a/test/execflags/bradc/gdbddash/gdbSetConfig.skipif
+++ b/test/execflags/bradc/gdbddash/gdbSetConfig.skipif
@@ -9,7 +9,7 @@ import os, os.path
 # installation and thus nothing to worry about.  To squash the noise, we will
 # avoid running it there.
 if os.getenv('CHPL_TARGET_PLATFORM') == 'cray-xc':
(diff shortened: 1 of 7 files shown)


dmk42:
@ben-albrecht - Would you please review this?

PR #9111, opened 2018-04-06
Fix some initializer leaks

benharsh:
This PR resolves two kinds of initializer leaks:
1) Parent deinits were not always called. This leak is resolved by setting the destructor on the appropriate AggregateType once fully instantiated.
2) Some default initializer values were not destroyed. This is resolved by adding FLAG_INSERT_AUTO_DESTROY to certain default values.

Two tests are added to lock in this behavior.

Testing:
- [x] full local + futures
- [x] full valgrind
- [x] full memleaks

This PR also recovers the leaks introduced by #9075

### DIFF
--- a/compiler/include/resolution.h
+++ b/compiler/include/resolution.h
@@ -261,4 +261,6 @@ SymExpr* findSourceOfYield(CallExpr* yield);
 
 void resolvePromotionType(AggregateType* at);
 
+void resolveDestructor(AggregateType* at);
+
 #endif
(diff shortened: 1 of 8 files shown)


lydia-duncan:
Does this resolve the default initializer failures in the --force-initializers run?

benharsh:
It resolved some of the failures under 'memory leaks'. The remaining failures are due to different reasons and do not appear to be memory related.

PR #9116, opened 2018-04-06
Issue error on any type assignment

bradcray:
While working on an [SO answer](https://stackoverflow.com/questions/49682634/domain-resizing-on-an-array-of-records-hangs/49683440#49683440) this morning, I ran into an error in which the compiler issues a confusing error on type assignments.  Specifically, given a statement like:

```chapel
type t = int;
t = real;
```

it would say:

```
type mismatch in assignment from real(64) to int(64)
```

as though the symbols were variables rather than types.  With this PR, it now says:

```
error: illegal assignment to type
```

Previously, we had a check to make sure that values couldn't be assigned to types, but really nothing should be able to be assigned to types.  Types can be initialized-assigned in their declarations and "phase 1" of a constructor of course, but these aren't really assignments (so don't trigger this error).

Added some tests to lock in this behavior and updated one existing test's output.


test/types/errors/assignTypes.chpl
```diff
@@ -0,0 +1,22 @@
...
+    this.t = int;
+  }
+}
+
+var myR: R(int);
+var myR2: R(int);
+var myR3 = new R();
+
+proc =(ref rhs: R(?t), lhs: R(t)) {
+  rhs.t = lhs.t;
```
cassella: The which-hand-side?

bradcray: It causes me no end of confusion that on QWERTY keyboards we type "L" with our right hand and "R" with our left.  :D

PR #9125, opened 2018-04-06
Add example of "slow sum" user-defined reduction

bradcray:
In reviewing tweets from new Chapel followers this week, I came
across one that said:

  Let's talk about the array, [1.0, 10e100, 1.0, -10e100], and whether
  programming languages use a slow, but sufficiently accurate sum by
  default to figure out that the sum is 2.0 rather than 0.0.

I went to bed thinking "yep, we'd get the wrong answer here too", but
then woke up thinking "...unless we wrote a user-defined reduction!"
See the [resulting twitter conversation here](https://twitter.com/ChapelLanguage/status/981538695739142144).

This program writes such a user-defined reduction, creating an
associative domain and array indexed by the exponent of the values and
combining values by bucket before combining across buckets, resulting
in the correct sum.  I thought it was a sufficiently cute program, as
well as an interesting example of a reduction whose state type doesn't
match its input/output types that I thought it'd be worth capturing.

Note that I intentionally left out some of the more esoteric routines
in the user-defined reduction interface for brevity (and also because
I expect that this interface will evolve as partial reductions come
online).


bradcray:
@vasslitvinov : FYI, I decided to capture this case as a test since it seemed intriguing.

PR #9135, opened 2018-04-07
chplspell dictionary updates

cassella:
chplspell dictionary updates


cassella:
One of these commits modifies the chplspell script itself to add ``HDF5_HL.chpl`` to the list of files to skip, as it's a generated file, and it has dozens of unique words.

Issue #9117, opened 2018-04-06
Support ranges of enums


bradcray:
As a user, I'd like to create ranges of enumerated values where the interpretation would relate to the ordering of the enums in their declaration rather than their underlying integer values because it seems like a natural 1D space to iterate over.

Acceptance criteria:
ranges of enums are implemented, support all normal/reasonable range operations, and are documented

bradcray:
This was completed in PR #9820.

PR #9138, opened 2018-04-08
Disable enum ranges for now

bradcray:
This change disables the creation of enum ranges for now.  We've never supported enum ranges, but because we've supported enum->int coercion, a range of enums has resulted in a range of ints.  This has led to some confusion because (a) the index variable when iterating over that range is an int rather than an enum and (b) if the enum's integer values aren't consecutive, values will be yielded that aren't in the enumeration.

Ultimately, we think we want to support ranges of enums (issue #9117), but before embarking on that, I wanted to turn the current ability off in order to stop relying on the coercion-oriented behavior.

This change puts in an explicit overload for creating a range of enums that currently generates an error message, but ultimately will work.  It also puts in enum overloads of the direct range iterators for similar reasons (otherwise, for loops over enum ranges work due to the coercion).

While here, I refactored some of the code in the range module to better support non-integer types.  Specifically, I improved the helper functions that determine the stride type and introduced helper functions to determine the low and high bounds and the alignment type in the explicit initializer.

Updated tests that relied on enum ranges to have them stop doing so (either using casts or by using the enum's enumerator) and added some new tests to lock in the current behavior (these will be updated as enum ranges get better support).


test/types/range/nonIntTypes/colorRange.good
```diff
@@ -0,0 +1 @@
+colorRange.chpl:3: error: ranges of enum type are not currently supported
```
cassella: Should the tests that result in this message be futures?

Would an enum range that turns out to be empty be an interesting test?  (e.g. ``color.blue..color.yellow``)

cassella:
Do you also need to change the spec

> An enumerated type defines a set of named constants that can be referred to via a member access on the enumerated type. These constants are treated as parameters of integral type.

so the last sentence has an exception for use as a range low or high value?

test/types/range/nonIntTypes/colorRange.good
bradcray: They could be, but it seems equivalent to me and this approach requires fewer files.

Once enum ranges are supported it would be, but we'll want/need dozens more tests by that point.

bradcray:
Right, I'm not planning to update the spec until enums are in their new steady state.  It'll probably just take me a number of smaller PRs to do that in a sane manner.

ronawho:
> @ronawho: Fixed the infinite recursion you noted and added a test for it.

Cool. Just curious, what error message do you get if you have a non-integral stride like: `red..blue by yellow`

ronawho:
The anonymous range optimization also triggers for `low..#high`, so its' probably worth adding a test case for that too? (It calls `chpl_build_low_bounded_range()`)

bradcray:
@ronawho: Good catch.  I put in overloads for that case and a test.  That said, at this point, I'm more inclined to continue focusing on disabling enum->int coercions (which will stop causing wrong cases to get called accidentally) rather than trying to plug all other holes where an enum might coerce to an int and generate a surprising result (e.g., I didn't provide an overload that takes an enum as the stride or count).

bradcray:
> Just curious, what error message do you get if you have a non-integral stride like: `red..blue by yellow`

Today:

```
error: ranges of enum type are not currently supported
```

Ultimately this won't be supported though without casting the stride to an int.

PR #9157, opened 2018-04-09
Fix library-related flags for LLVM back-end

bradcray:
In working on PR #9030, I didn't test the LLVM back-end and failed
to notice that my changes would break it.  This replaces the old
array of library flags generation for LLVM with the new STL Vector-based
approach.


bradcray:
@mppf: Can you give this a quick look / review?  I think it's trivial and it passes the new test, but since I don't know the LLVM code base...  Could also be a post-merge review.

PR #9158, opened 2018-04-09
Strawman partial reductions for DefaultRectangular

vasslitvinov:


bradcray:
My excitement by seeing this PR dissipated a bit when I saw that these are .notest-ed.  What issues prevent them from working?

bradcray:
Oh nevermind.  I missed that only the helper files are notested.  Deleting both these comments.

PR #9160, opened 2018-04-09
Convert BaseArr and its derived classes to use initializers

### LINKED ISSUE #9149: Convert arrays to use initializers
As a Chapel implementer, I want arrays to use initializers because that would demonstrate stability and improve confidence in the implementation of initializers.

Acceptance criteria:
All classes inheriting from BaseArr either:
- have an explicit initializer
- have a "use default init" pragma

benharsh:
This PR converts BaseArr and its derived classes to use initializers. This is generally accomplished by either adding a 'use default init' pragma or converting an existing constructor to an initializer.

A number of bugs are resolved by this PR:
1) Point of instantiation issue for types with initializers is resolved by using the same 'fix' as constructors: set the point of instantiation for virtually dispatched methods to be the same as the initializer that instantiated the type.

2) Allow field default expressions to use previous fields that are generic. Solved by only setting ``dtAny`` on default formals that are type variables.

3) Fixes an issue with array runtime types as fields by using PRIM_GET_MEMBER_VALUE instead of PRIM_GET_MEMBER.

4) A minor segfault is avoided by checking if the type has a type constructor before attempting to use it.

5) Reproducing the recursive-resolution avoidance from #9004 for "_new" wrappers.

These bugs primarily impact complex hierarchies like those found in the array classes.

Resolves #9149 

Testing:
- [x] full local + futures
- [x] full gasnet
- [x] numa on arrays/, release/examples

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -1805,8 +1805,10 @@ void AggregateType::fieldToArg(FnSymbol*              fn,
 
           typeExpr->insertAtTail(new CallExpr(PRIM_TYPEOF, tmp));
 
-          arg->typeExpr    = typeExpr;
-          arg->type        = dtAny;
+          arg->typeExpr = typeExpr;
+          if (arg->hasFlag(FLAG_TYPE_VARIABLE)) {
+            arg->type = dtAny;
+          }
 
           // set up the ArgSymbol appropriately for the type
           // and initialization from the field declaration.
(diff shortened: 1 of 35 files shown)


benharsh:
This PR is waiting on #9113 to be merged.

benharsh:

benharsh:
Following up, performance on single-node and 16-node-xc looks good.

Issue #7924, opened 2017-12-02
Internal error for default value of ref argument


cassella:
### Summary of Problem

```chapel
proc foo(ref x = 7) { }
foo();
```

```
fortytwo@magrathea:~/src/chpl$ chpl --devel def-value.chpl 
def-value.chpl:2: internal error: misuse of codegenAddrOf [expr.cpp:1382]
```

There's no error if the call is ``foo(y)``.

If instead it's ``out x = 7``, the error is

```
def-value.chpl:1: error: invalid type specification
```

but again ony if ``foo()`` is called without argument.

If it's ``inout x = 7``, it compiles and runs.

### Steps to Reproduce

**Associated Future Test(s):**
 [`test/functions/default-arguments/default-argument-inout.chpl`](
     https://github.com/chapel-lang/chapel/blob/master/test/functions/default-arguments/default-argument-inout.chpl) #9165
 [`test/functions/default-arguments/default-argument-ref.chpl`](
     https://github.com/chapel-lang/chapel/blob/master/test/functions/default-arguments/default-argument-ref.chpl) #9165  (retired by #13725)
 [`test/functions/default-arguments/default-argument-out.chpl`](
     https://github.com/chapel-lang/chapel/blob/master/test/functions/default-arguments/default-argument-out.chpl) #13725

[edit 4/24 cassella to add links to futures in a PR ]

### Configuration Information

```
fortytwo@magrathea:~/src/chpl$ chpl --version
chpl Version 1.17.0 pre-release (49cca06)
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)

fortytwo@magrathea:~/src/chpl$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
[...]

PR #9165, opened 2018-04-10
Add futures for ref and inout arguments with default values (issue #7924)

cassella:
Add futures for issue #7924


cassella:
Is this the right way to write a ``.bad`` file for an internal error?

I didn't add futures for the versions where the call is made with a var for an actual, in which case the compiler doesn't complain about the argument having a default value.  I think it should complain in that case too, as the presence of the default value suggests the function can be called with that argument omitted.

I'm not attached to the phrasing of the error I used in the ``.good`` files.

lydia-duncan:
> I didn't add futures for the versions where the call is made with a var for an actual, in which case the compiler doesn't complain about the argument having a default value. I think it should complain in that case too, as the presence of the default value suggests the function can be called with that argument omitted.

This falls under the general category of us only generating errors for structures that are used at compilation, which is something we are aware of but haven't worked on

test/functions/default-arguments/default-argument-inout.chpl
```diff
@@ -0,0 +1,3 @@
+proc foo(inout x = 7) { x = 17; writeln(x); }
+foo();
+
```
lydia-duncan: I don't like that these tests have an extra line at the end of them, but I'm willing to live with it.  That said, I'm going to wait until Monday to merge so if you get motivated... :p

test/functions/default-arguments/default-argument-ref.bad
```diff
@@ -0,0 +1,8 @@
+default-argument-ref.chpl:2: internal error: EXP1418 chpl version 1.18.0 pre-release (a1d1fb9)
+Note: This source location is a guess.
+
+Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
+and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
+please see https://chapel-lang.org/bugs.html for instructions.  In the meantime,
+the filename + line number above may be useful in working around the issue.
+
```
lydia-duncan: Yeah, this is right.

test/functions/default-arguments/default-argument-inout.chpl
cassella: Motivation found.  Here y'go.

PR #9172, opened 2018-04-11
Be sure to compute max comm domains before needing it!

### LINKED ISSUE #9154: comm=ugni minimal-registered mode sometimes doesn't register enough memory
While investigating another issue I unloaded the `craype-hugepages` module I had been using, so I was then in minimal-registered mode.  I then encountered this comm=ugni internal error trying to run ra-rmo:

    internal error: out of NIC-registered space

The way this mode works is by computing the total amount of registered space that will be needed, allocating and registering that, and then handing out pieces of it as the comm layer requests them.  Each request is supposed to have a matching reservation done earlier, so that by the time we need the space it's all ready and there's enough.  The message above is printed when an allocation call finds that there isn't sufficient space.

This isn't a mode that we test regularly, but nevertheless it should work.  Probably this is just a bit-rot problem.

gbtitus:
Recent (last several months) code movement led to a situation in which,
for minimal-registered-memory mode, we used the job-wide maximum number
of comm domains per node before we had computed that.  Here, move the
computation and jobwide distribution of that value prior to the point at
which we need it.

Subsequent testing revealed an additional bug: we have been using the
number of logical processors when estimating how many comm domains we
will need.  Using the tasking layer's max level of parallelism is much
better.

That in turn pointed out yet one more bug, less serious but still noisy:
the check to make sure the pool of remote fork completion flags was big
enough was comparing the expected maximum demand for flags to the number
of flags per pool, not the total number of flags in all pools.  So here,
fix that as well.

This fixes #9154.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1841,11 +1841,16 @@ void chpl_comm_post_task_init(void)
     return;
 
   //
-  // Now that the memory layer has been set up, we can allocate the various
-  // data we need and fill it in.  Some of this need not be shared and so we
-  // could have allocated it earlier, but the logical flow is clearer if we
-  // do all this in one place.
+  // Now that the memory layer and tasking have been set up, we can
+  // allocate the various data we need and fill it in.  Some of this
+  // need not be shared and so we could have allocated it earlier, but
+  // the logical flow is clearer if we do all this in one place.
   //
+
+  //
+  // Figure out how many comm domains we need.
+  //
+  compute_comm_dom_cnt();
 
   //
   // Get our NIC address and share it around the job.
@@ -1856,7 +1861,8 @@ void chpl_comm_post_task_init(void)
   // do this by gathering (locale, nic_addr) pairs and then scattering
   // the nic_addr values into the map, which is indexed by locale.
   //
-  // While we're at it, also share the barrier info struct addresses
+  // While we're at it, also compute the maximum number of communication
+  // domains on any node and share the barrier info struct addresses
   // around the job.
   //
   {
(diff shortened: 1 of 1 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -2057,7 +2066,7 @@ static void compute_comm_dom_cnt(void)
     if (comm_dom_cnt == 0) {
       uint32_t num_PUs;
 
-      if ((num_PUs = chpl_getNumLogicalCpus(true)) > 0)
+      if ((num_PUs = chpl_task_getMaxPar()) > 0)
```
ronawho: Also, in the past I have tried to call `chpl_task_getMaxPar()` before the tasking layer was initialized (in an early version of https://github.com/chapel-lang/chapel/pull/6893 -- but that was from `chpl_comm_init()`, not `chpl_comm_post_task_init()`. 

Because of my previous screw-up I wanted to check to make sure this call was after the tasking layer was initialized (and indeed itis). I wonder if we should add a comment above the `compute_comm_dom_cnt()` declaration that says "Note, must be called after the tasking layer is initialized" just to ensure we don't move it too early in the future? (I'm a little torn on that because that comment will likely become out of date)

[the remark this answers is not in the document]
gbtitus: Yeah, good idea; I'll do that.

gbtitus: I feel like it's okay to leave this without a comment, since by its name we know `chpl_comm_post_task_init()` is called after tasking is running, and there's no reason to need the comm domain count before where it now is computed.  (Could be "famous last words", I guess.)

gbtitus: Oh, hmm, you were talking about adding a comment before the `compute_comm_dom_cnt()` decl, not the call.  Yeah, probably a good idea.  I'll do that.

PR #9175, opened 2018-04-12
Convert Vec<> data structures to std::vector in util/files.cpp

bradcray:
While debugging the issue in PR #9174, I realized that the `Vec<>` data structures used for these lists of `-I`, `-L`, and `-l` flags would be an easy case to convert to `stl::vector`, so took that on as a cleanup exercise while debugging.


compiler/util/files.cpp
```diff
@@ -740,10 +741,11 @@ void codegen_makefile(fileinfo* mainfile, const char** tmpbinname, bool skip_com
   genCFiles(makefile.fptr);
   genObjFiles(makefile.fptr);
   fprintf(makefile.fptr, "\nLIBS =");
-  forv_Vec(const char*, dirName, libDirs)
+  for_vector(const char, dirName, libDirs)
     fprintf(makefile.fptr, " -L%s", dirName);
-  forv_Vec(const char*, libName, libFiles)
+  for_vector(const char, libName, libFiles) {
     fprintf(makefile.fptr, " -l%s", libName);
+  }
```
mppf: Just curious, why the curlies here but not in the previous one?

bradcray: Uh..... I think I must've added some debug printing at some point and then removed it without noticing I hadn't "cleaned it back up" again.

bradcray:
> I think it's weird that forv_Vec used const char* but the for_vector version doesn't 

Yeah, I got tripped up by that for awhile as well and don't understand the reason behind it (if there was one... checking the comments and commits that introduced it, I'm not seeing a clear explanation, and I haven't tried reading through the code itself).  It also seems problematic if, say, the vector was storing `int`s.  That said, there's a TODO in the header essentially saying "We should get off of our macros and use better features to do this" but I wasn't feeling that brave / luxurious in terms of free time.

PR #8785, opened 2018-03-12
move lowering of ForallStmts to lowerIterators

vasslitvinov:
Pipeline issue: #8701 

### TESTS	

* Un-futurized the test:
        multilocale/vass/recursive-iterator-twice-multiloc

* Removed the following test as it is identical to the above under comm=none
  (all it does is remove on-clauses):
        functions/iterators/vass/recursive-iterator-twice-local

* Resolves #7159 i.e. compiler no longer crashes on this test:
        functions/iterators/vass/forall-in-standalone-iterator

### Merge TODOs

General
- [ ] PR message
- [ ] need more on test changes in the PR message?
- [x] top-level comment in lowerForalls.cpp on the approach, code structure, etc.
- [x] rename certain functions, per Michael's request
- [x] add an explanation for the added ref intent in modules/standard/FileSystem.chpl
- [x] remove/address VASS comments
- [x] update the "or 1 idx var" comment
- [x] performance - meteor-parallel, meteor release version ?

Fix known correctness-testing failures
- [x] baseline - parallel/forall/vass/reciter/forall-reciter-works
- [x] futures - virtual parallel iterators

Re-run paratests that will be expected to pass:
- [ ] linux64 + futures
- [ ] gasnet
- [x] valgrind
- [x] memleaks
- [ ] baseline
- [x] LLVM

PR #9176, opened 2018-04-12
Fix an incorrect usage of sys_getenv()

### LINKED ISSUE #9029: helpSetupLocaleFlat() checks for wrong return from sys_getenv()
### Summary of Problem

```chapel
  proc helpSetupLocaleFlat(dst:LocaleModel, out local_name:string) {
    const _node_id = chpl_nodeID: int;
...
    // sys_getenv returns zero on success.
    if sys_getenv(c"CHPL_COMM", comm) == 0 && comm == c"gasnet" &&
      sys_getenv(c"GASNET_SPAWNFN", spawnfn) == 0 && spawnfn == c"L"
    then local_name = chpl_nodeName():string + "-" + _node_id:string;
    else local_name = chpl_nodeName():string;
```

But ``runtime/src/qio/sys.c`` ``sys_getenv()`` returns 1 on success:

```c
// Returns true if an environment variable called name was found,
// in which case *string_out will point to the string found in
// the process's environment block. Don't try to free it.
// Returns false if the name was not found in the environment
//  (ie getenv returned NULL).
int sys_getenv(const char* name, const char** string_out)
{
  char *buf;

  buf = getenv(name);
  if (buf==NULL) {
    return 0;
  } else {
    *string_out = buf;
    return 1;
  }
}
```

I'm not prepared to evaluate what happens if I change it, so I'm just reporting it.

### Steps to Reproduce

**Source Code:**

```chapel
// Please provide source code that will reproduce the problem.
// You can insert your code inline if it's not too long.
// Otherwise, you can attach it as a file or provide a URL to it.
// To the extent possible, providing simplified programs demonstrating the
// problem will be appreciated.
// You can also replace this section with "Associated Future Test(s)" below.
```

**Compile command:**

**Execution command:**

### Configuration Information

- Output of `chpl --version`:
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
- (For Cray systems only) Output of `module list`:

dmk42:
Some of the locale setup code assumes that `sys_getenv()` returns zero on success, but it returns one on success.  This change fixes the locale model assumption, and supplies a test for the result.

This fixes #9029 .


ronawho:
Just curious if you think it'd be better to change `sys_getenv()` to return 0 on success instead?

dmk42:
For system routines, I like returning 0 on success because then you can return an `errno`, but this is the only location that assumed that.  There are a bunch of others that assume 1 on success.  It doesn't really matter, so I changed the minimum amount of code.

dmk42:
@ronawho - Since you're looking, would you mind reviewing?

dmk42:
Although I'm confident of this PR, I cannot complete testing properly until the issues caused by #8785 are resolved, so I will wait.

dmk42:
#9205 made it possible to finish testing this PR.

PR #9177, opened 2018-04-12
Fix this one weird assignment bug

cassella:
This test's inititializer still had super.init() after the field
initialization.  Removing the superfluous super.init() restores the
failure described in the future.


cassella:
weirdAssignmentBug2 was weirdly already without this bug.

PR #9188, opened 2018-04-13
Process /proc/cpuinfo for CPU counts in a better way.

### LINKED ISSUE #9170: issue installing chapel (internal error: varying number of cpu cores)
I use standart gnu/linux x64 (void linux) and I am trying to install chapel:

```
$ source util/setchplenv.bash #also tried source util/quickstart/setchplenv.bash
$ make
$ chpl -o hello examples/hello.chpl
$ ./hello
> internal error: varying number of cpu cores
(make check produces the same error)
```
I am not sure what am I doing wrong. The compilations seems to finish successfully.

gbtitus:
We've been computing node core counts by dividing the number of
'processor' entries in `/proc/cpuinfo` by the number of hardware threads
per core.  We've been getting the latter two values by looking at the
'siblings' and 'cpu cores' entries, and to avoid being confused we've
insisted that all the 'siblings' entries have the same value and all the
'cpu cores' entries have the same value.  But we've begun to encounter
situations where that doesn't hold, which leads to an internal error and
we cannot continue.

Here, switch to just counting the number of distinct pairs of 'physical
id' and 'core id' values to give us a core count.  For at least the one
internal error situation for which we have a copy of the `/proc/cpuinfo`
file, this gives the correct result.

I believe this will resolve #9170 and the problem reported [here](https://github.com/buddha314/Charcoal/issues/6).

### DIFF
--- a/runtime/src/chplsys.c
+++ b/runtime/src/chplsys.c
@@ -587,66 +587,67 @@ static void getCpuInfo_once(void) {
 
   if (cpuTabLen > 0) {
     numCores = cpuTabLen;
-  } else {
-    // We have a limited-format /proc/cpuinfo.
-    // See if the /sys filesystem has any more information for us.
-    int threads_per_core = 0;
-    if ((f = fopen("/sys/devices/system/cpu/cpu0/topology/thread_siblings",
-                   "r")) != NULL) {
-      int c;
-      while ((c = getc(f)) != EOF) {
-        // The number of threads per core is the total number of bits
-        // set in the hex digits of the thread_siblings map.
-        //
-        // The most authoritative source, kernel.org, does not dictate
-        // the formatting of the mask.  However, the following document
-        // indicates that it is either a hex or binary bit mask.  The code
-        // below works in either case.
-        //
-        // https://www.ibm.com/support/knowledgecenter/en/linuxonibm/liaat/liaattunproctop.htm
-        //
-        // Also note that hwloc itself parses the file as a hex bitmap only.
-        if (isxdigit(c)) {
-          switch (tolower(c)) {
-          case '0':
-            break;
-          case '1':
-          case '2':
-          case '4':
-          case '8':
-            threads_per_core += 1;
-            break;
-          case '3':
-          case '5':
-          case '6':
-          case '9':
-          case 'a':
-          case 'c':
-            threads_per_core += 2;
-            break;
-          case '7':
-          case 'b':
-          case 'd':
-          case 'e':
-            threads_per_core += 3;
-            break;
-          case 'f':
-            threads_per_core += 4;
-            break;
-          }
-        } else if (c != ',' && c != '\n' && tolower(c) != 'x') {
-          // unknown file format -- don't use
-          threads_per_core = 1;
+    return;
+  }
+
+  // We have a limited-format /proc/cpuinfo.
+  // See if the /sys filesystem has any more information for us.
+  int threads_per_core = 0;
+  if ((f = fopen("/sys/devices/system/cpu/cpu0/topology/thread_siblings", "r"))
+      != NULL) {
+    int c;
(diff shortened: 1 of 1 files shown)


gbtitus:
@dmk42 would you take the lead on reviewing this one?  It's right near some work that you just did in #8560.  In fact it slightly modifies that structurally, though not functionally.

gbtitus:
I need to depart and will be away tomorrow/Friday.  It would be nice if we could get one or both of the users who reported the problem this addresses to try out the fix.  I don't have time to do this right now but I will tackle it Monday.

runtime/src/chplsys.c
```diff
@@ -502,19 +502,39 @@ uint64_t chpl_sys_availMemoryBytes(void) {
...
-  //
-  // Currently this is pretty limited -- it only returns the number of
-  // physical and logical (hyperthread, e.g.) CPUs, mostly relies on
-  // /proc/cpuinfo, and only supports homogeneous compute nodes with
-  // the same number of cores and siblings on every physical CPU.  It
-  // will probably need to become more complicated in the future.
-  //
+static struct {
+  int physId, coreId;
+} cpuTab[1024];
```
ronawho: Does this mean we'll fail on machines that have more than 1024 cores? That seems like an unfortunate limitation. I think the largest the kernel currently supports is 8192: https://github.com/torvalds/linux/blob/1bad9ce155a7c010a9a5f3261ad12a6a8eccfb2c/arch/x86/Kconfig#L991-L1002

I'd prefer to go with 8192 here, though that'll only become obsolete over time, so I wonder if there's a more programatic way to check.

ronawho:
Belatedly I'm thinking we should have an env var that allows us to read from a file other than `/proc/cpuinfo` for the purpose of adding some nightly testing. i.e. so that we can grab some `/proc/cpuinfo` files from a bunch of configs and make sure we can parse them correctly

It might also be nice to have have an env var to force set the number of cpus just as a workaround if in the future there are more `/proc/cpuinfo` files we can't read. 

And longer term, we probably just want to rely on hwloc, at least when qthreads is being used -- but that's a discussion for another time, not something to worry about for this PR.

runtime/src/chplsys.c
```diff
@@ -526,103 +546,126 @@ static void getCpuInfo(int* p_numPhysCpus, int* p_numLogCpus) {
...
-            break;
-          case 'f':
-            threads_per_core += 4;
-            break;
-          }
-        } else if (c != ',' && c != '\n' && tolower(c) != 'x') {
-          // unknown file format -- don't use
-          threads_per_core = 1;
+  if ((numPUs = procs) <= 0)
+    numPUs = 1;
```
dmk42: We ought to see the `processor : <integer>` lines in every `/proc/cpuinfo` but I'm finding some pretty strange `/proc/cpuinfo` instances out there, so it's probably better to assume they might be missing even when we see the `core id` and `physical id` lines.  In that case, instead of defaulting `numPUs` to 1, it would be better to default it to whatever we end up setting `numCores` to.  That way, `numPUs` will always be at least as big as `numCores`, which I think is something we should expect.

runtime/src/chplsys.c
gbtitus: I could count the newlines in `/proc/cpuinfo` first and dyamically allocate the table to that size, then rewind the file and process it for real.  Surely the system cannot have more CPUs, either physical or logical, than the number of lines in `/proc/cpuinfo`.

More seriously, I can certainly make the table dynamically sized.

gbtitus:
@dmk42 and @ronawho: okay to merge now?

gbtitus:
Looking back on the comments, I believe I have an okay to merge after doing the last two minor fixes already, so I'm going to proceed with that.  The failure in pre-merge testing is due to some pypi problem.

PR #9191, opened 2018-04-13
create Index and Series ADTs for Dataframes

psahabu:
### Motivation
The motivation for the Index and Series ADTs can be found in the connected issue.

### Limitations
While the implementation was mostly straightforward, the thorniest issue was actually the Index because it needs a two-way map with a notion of order. The draft idea is to have it be two arrays, but there are ideas for future implementations in the comments.

Due to the lack of constrained generics, the Dataframes module had to resort to distinct Series/TypedSeries and Index/TypedIndex classes. In the future, the hope is that Index and Series would not need the Typed* implementations beneath them.

### Notes
Included in this PR are pandas equivalents to our tests. Note that the only significant differences are:
- 0 (pandas) vs 1 (chapel) based indexing
- indexes do not have `items()` iterators in pandas, whereas they do in chapel


buddha314:
Did this commit provide any functionality I can tinker with?

psahabu:
@buddha314 #9342 has the latest and greatest on Indexes and Series, working my way up to Dataframes.

PR #9192, opened 2018-04-13
Stop permitting enums to coerce to ints

### LINKED ISSUE #5282: What should enum comparisons do?
While looking through the param folding code for enums, @ronawho  and I found that enum equality was implemented by comparing the enums' integer values.  That is, given enums:

    enum color { red, green, blue };
    enum day { sunday, monday, tuesday };

the param comparison:

    if (color.green == day.monday) then ...

folds to true because green and monday both have the same integer value.  This struck me as surprising because many of our param enum cases are designed to see if the enum is the same symbol.  For example, when checking to see whether an iterator has the 'standalone' tag, it isn't our intention to see if the argument is some other enum or integer whose value happens to equal 'standalone'.  On the other hand, if you view enums less as "new values" and more as "convenient aliases for integer values" maybe the current behavior is less surprising.

I was thinking this was specific to params, but in digging a bit deeper, it turned out that this is what happens for execution-time enum comparisons as well, apparently based on the fact that both enums coerce to integers and then match the integer comparison operators.

To that end, the questions I'm interested in addressing in this issue are:

1) Do we want to support inter-enum comparisons, or should we only support such comparisons within:
    a) a given enum type
    b) enum/integral pairs
    c) never (a user would have to cast the enums to ints in order to do any comparisons on them.

2) Is the above decision only for == and != or for <, <=, >, >= as well?

### LINKED ISSUE #8658: adding enums and uint(8)s result in a surprising ambiguity
### Summary of Problem

In talking about enums on issue #8552 last week, Michael noted that when we add an enum and
a `uint(8)` we get an ambiguous function error that we don't get with a `param int(64)` and a
`uint(8)` in spite of the fact that enums generally behave like `param int(64)`s.  This issue captures that inconsistency.

It's worth noting that, on email, we're discussing potentially having enums stop coercing to
ints, in which case this issue could be closed (as things work correctly when a cast it
used, as shown in the future below).

[`test/types/enum/addEnumUint8.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/types/enum/addEnumUint8.chpl
          ) #8657

### Configuration Information

- Output of `chpl --version`: `chpl version 1.17.0 pre-release (fb061f5)`

### LINKED ISSUE #8794: Should enums coerce to `int`s as they do now?
At present, enums can coerce to integer types, as in C, and via rules that are fairly similar to how we handle `param` integers.  This sometimes results in surprising behavior such as when a range over enums is created (`color.red..color.green` results in an integer range using the enums' underlying values rather than a range of `color`s) and when applying operators to enums of different types (e.g., see issue #5282).

This issue asks the question whether we should disable enum->int coercion, requiring users to cast enums to integers when they wish to use them as their underlying integer values.  In such a world, the compiler would create `==` and `!=` operators for comparing two symbols within a given enum type but comparisons across enum types would no longer be allowed (unless the user chose to cast them to integers and compare the integer values).

bradcray:
[thanks to @mppf for some pointers that got me off to a good start]

This PR removes coercions from enums to ints because they are more likely to lead to confusion than benefit (see issue #8794, for instance).  Enums can still be cast to ints for users who want/need to use their underlying integer values.

The main effort here was to remove code from `functionResolution.cpp` that supported these coercions.  This permitted a few now-dead methods to be removed from `EnumType`.  The other main effort here was to introduce `==` and `!=` functions between two values of a given enum type so that unordered comparisons could still be performed between them.  I did this via function overloads in ChapelBase.chpl.  While here, I took the opportunity to move the compiler-generated assignment function between enums out to ChapelBase.chpl as well.  It should be noted that these overloads are not as clean as I'd like them to be because of issue #9310.  Namely, where I wanted to write:

```chapel
proc ==(x: enumerated, y: x.type) { ... }
```

I instead had to write:

```chapel
proc ==(x:enumerated, y:enumerated) where (x.type == y.type) { ... }
```

Once #9310 is fixed, these can/should be cleaned up.

One impact of this PR is that inequality comparisons (`<`, `<=`, `>=`, `>`) on enums are no longer supported, as that's what was proposed in issue #8794.  Issue #9306 asks whether this is the right choice or not, but will be tackled in a future PR if agreed upon.

One other change that this PR resulted in was to introduce the `enum`'s enumerate function at the global scope rather than at the enum's scope as we did before.  Prior to this PR, it and the assignment were the only enum functions inserted at the enum's scope, so this makes all the cases more consistent (though Michael points out in his comments on this PR that this is something we'd like to fix/improve for other types as well, so it's a shame that we're relying on this approach to work around a more general compiler issue).

In addition, I had to add casts from enums to imaginaries and complexes to keep the current I/O module working (previously, enums had coerced to the int->imag/complex case).  And a default hash function for DefaultAssociative so that we could store associative domains of enums (again, previously, we had coerced to the int hash).

Interestingly, the new stricter enums caught a long-standing point of confusion in the `MatrixMarket` module where it passed an `iokind` enum to a `c_int` hint argument where it was interpreted based on its integer value (giving it an entirely different interpretation than the `iokind` had implied.  The hint turned out to be a non-buffered hint, so fixed this by removing the bad argument and inserting manual flush() calls instead (to get the right behavior with better performance).

A few other modules had to be updated in order to stop relying on enum->int coercions in simple cases involving math and passing enums off to external routines expecting integers (DateTime.chpl and IO.chpl)

This test resolved two futures:
* test/types/coerce/diten/test_override_equals.future (because enums no longer coerce to int overloads in ChapelBase)
* test/types/enum/addEnumUint8.future (because the candidates in question are no longer viable without coercions)

(Note that these changes futurized three submitted shootout benchmarks: fasta, fasta2, and meteor because they require updates on the CLBG site to work without enum->int coercions)

The bulk of the work required test updates and additions:

Added some new tests to lock in certain behavior
* test/types/enum/otherModEnum.chpl -- ensures that compiler-generated enum support is accessible from a remote module
* test/types/enum/primCasts.chpl -- ensures that casts from enums to primitive types work as intended

Filed a few new tests which are variations on old ones to show different behavior
---------------------------------------------------------------------------------------
* test/arrays/bradc/enumarray3.chpl: a future regarding range-of-enum support
* test/types/enum/bradc/dispatch/dispatchenums-ints.chpl -- split the integer cases off from dispatchenums.chpl
* test/types/enum/bradc/dispatch/enumarray2.chpl -- a variant on enumarray.chpl that works as the original test did
* test/types/enum/bradc/dispatch/enumdispatch2.chpl -- a variant on enumdispatch.chpl that works as the original test did
* test/types/enum/enumsLikeParams2.chpl -- a variant on enumsLikeParams.chpl that shows what happens to the old code (now updated to preserve original intent)
* test/types/enum/diten/paramenums2.chpl -- a variant on paramenums.chpl that splits out the cases that relied on inequality comparisons on enums, which are no longer supported.

Inserted casts to keep an enum range as an int range
-----------------------------------------------------
* test/arrays/bradc/enumarray.chpl
* test/studies/amr/lib/grid/Grid_def.chpl
* test/studies/comd/llnl/CoMD.chpl

Changed from arithmetic domain to associative domain on enum to avoid enum ranges
-------------------------------------------------------------------------------------
* test/npb/cg/bradc/cg-arithDomArr.chpl
* test/studies/comd/llnl/utils/helpers.chpl
* test/users/npadmana/npb-mg/mg.chpl

Cast enums to ints in order to index into arrays with them
----------------------------------------------------------
* test/distributions/robust/arithmetic/modules/test_module_Norm.chpl
* test/npb/is/mcahir/intsort.mtml.chpl
* test/studies/comd/llnl/CoMD.chpl

Added enum comparisons in order to call the sorted iterator on a domain of enums
-----------------------------------------------------------------------------------
* test/arrays/diten/enumArrayMethods.chpl
* test/associative/diten/enumArrays.chpl

Added enum comparisons in order to preserve test's old behavior
------------------------------------------------------------------
* test/types/enum/sungeun/test_foldEnumOp.chpl

Inserted casts in order to call functions on ints
* test/functions/diten/param_enum_to_int.chpl

Cast to ints to do math / comparisons
--------------------------------------
* test/puzzles/diten/roman.chpl
* test/release/examples/benchmarks/shootout/meteor.chpl
* test/studies/amr/lib/grid/GridBC_def.chpl
* test/studies/amr/lib/grid/Grid_def.chpl
* test/studies/shootout/chameneos-redux/hannah/chameneos-redux-v1.chpl
* test/studies/shootout/chameneos-redux/hannah/chameneos-redux.chpl
* test/studies/shootout/meteor/kbrady/meteor-implicit-domain.chpl
* test/studies/shootout/meteor/kbrady/meteor-parallel.chpl
* test/studies/shootout/meteor/kbrady/meteor.chpl
* test/types/enum/compare.chpl
* test/types/enum/ferguson/dependent-uint.chpl
* test/types/enum/ferguson/enum-overflow.chpl
* test/types/enum/nspark/dependentElements.chpl
* test/types/enum/typed-offset2.chpl
* test/users/kreider/bug_enum.chpl

Inserted casts in order to assign between enums and ints
-------------------------------------------------------
* test/studies/shootout/fasta/kbrady/fasta-enum.chpl
* test/trivial/bradc/protoast/codegen/enum.chpl
* test/types/enum/deitz/test_enum1.chpl
* test/types/enum/deitz/test_enum2.chpl
* test/types/enum/deitz/test_enum3.chpl
* test/types/enum/deitz/test_enum4.chpl
* test/types/enum/deitz/test_enum5.chpl
* test/types/enum/deitz/test_enum6.chpl
* test/types/enum/deitz/test_enum_overload.chpl
* test/types/enum/ldelaney/enum-iterator.chpl
* test/types/enum/lydia/enumIterateInGenericFunc.chpl
* test/types/enum/vass/enum-to-uint-param-1.chpl

Switched from enums to params for simplicity
* test/release/examples/benchmarks/lulesh/luleshInit.chpl
* test/studies/lulesh/bradc/xyztuple/luleshInit3.chpl

Changed from array of uint(8)s to array of enums + casts for simplicity
----------------------------------------------------------------------
* test/release/examples/benchmarks/shootout/fasta.chpl
* test/studies/shootout/fasta/bradc/fasta-blc-numa.chpl

Updated tests testing dispatch choices for enums vs. integers
-----------------------------------------------------------------
* test/functions/bradc/resolution/badEnumDispatch.chpl
* test/functions/bradc/resolution/intChoices.chpl
* test/functions/bradc/resolution/intChoicesDupTypes.chpl
* test/functions/bradc/resolution/smallIntBoolTests/boolInt.chpl
* test/functions/bradc/resolution/smallIntBoolTests/boolToUint.chpl
* test/functions/bradc/resolution/smallIntBoolTests/paramNon-withenum.chpl

Updated test to distinguish between 'sync enum' and 'sync int'
* test/types/enum/bradc/dispatch/syncenum.chpl

Updated test to make sure it continued showing that big values work
---------------------------------------------------------------------
* test/types/enum/ferguson/dependent-uint.chpl
* test/types/enum/ferguson/enum-overflow.chpl

Removed test as it seemed to no longer have value
* test/types/enum/ferguson/enum-uint.chpl

Changed the test as requested by Vass (to reflect original intent?)
------------------------------------------------------------------
* test/types/enum/vass/enum-to-uint-param-1.chpl

Fixes #8658 
Fixes #5282 
Fixes #8794


test/arrays/bradc/enumarray3.chpl
```diff
@@ -0,0 +1,19 @@
+enum classVals { S, W, A, B, C, D, O };
+
+var DClass = {classVals.S..classVals.O};
```
cassella: Will the resulting domain be a ``domain(1)``?

What if the enum had specified values with gaps, or out of order values?

Will it be possible to write this as ``var DClass = {classVals};``?

bradcray: A `domain(rank=1, idxType=classVals)`, yes.  The current notion is that ranges over enums would hit all of the enums in the order of the enumeration independent of underlying numerical value.  Of course, the compiler would have optimization opportunities when the underlying numerical values happened to be consecutive as in this case.

bradcray: > Will it be possible to write this as ``var DClass = {classVals};``?

I'd guess "no" though this might be considered a reasonable shorthand for `domain(classVals)` (an associative domain of enums).

test/types/enum/ferguson/dependent-uint.compopts
```diff
@@ -0,0 +1 @@
+--ignore-errors-for-pass
```
mppf: I don't think I'm familiar with `--ignore-errors-for-pass`. For this test though I think it'd be better if we just remove the is_uint calls since which type they coerce to is now moot. The enum itself is still worthy of testing since it contains large numbers in an unusual manner.

test/types/enum/ferguson/enum-overflow.compopts
```diff
@@ -0,0 +1 @@
+--ignore-errors-for-pass
```
mppf: Similar here - I'd rather take out the writeint calls and just replace them with writelns.

benharsh:
The CoMD changes look good to me.

test/associative/diten/enumArrays.chpl
```diff
@@ -47,3 +47,7 @@ for (a, d) in zip(A2, D2) {
   if a != A2(d) then
     writeln("Error A2(", d, ") = ", A2(d), " a = ", a);
 }
+
+proc <(x: F, y: F) {
+  return x:int < y:int;
+}
```
daviditen: Missing a newline at the end of the file.

test/functions/diten/param_enum_to_int.good
```diff
@@ -1,2 +1,6 @@
-param 2
-non-param 2
+param_enum_to_int.chpl:13: error: unresolved call 'foo(color)'
```
daviditen: I think I'd rather add explicit casts in this test to make sure that a param enum->int cast still exists.  (It does.)

daviditen:
I had little comments on these two:
```
test/associative/diten/enumArrays.chpl
test/functions/diten/param_enum_to_int.chpl
```

The rest look good.

I didn't see any changes under `test/types/enum/diten`.  I reviewed the changes under `test/types/enum/deitz` instead.

test/types/enum/vass/enum-to-uint-1.chpl
```diff
@@ -5,17 +5,17 @@ config var e1var: e1;
 config var e2var: e2;
```
vasslitvinov: With this change, this test checks explicit casting from simple enums to uints. If we have such checks elsewhere, this test can be removed. Otherwise it looks good.

test/types/enum/vass/enum-to-uint-param-1.chpl
```diff
@@ -1,4 +1,4 @@
 
 enum e { a=-1, b=1 };
-var i:uint(8) = e.b;
+var i = e.b: uint(8);
```
vasslitvinov: Let us also make 'i' a `param`, if that works. I think that was my original intention. Otherwise this test offers little in addition to enum-to-uint-1.chpl .

vasslitvinov:
Other than the two comments just above, test/types/enum/vass looks good.

bradcray:
@mppf: If you could re-review to make sure I did something approximating what you wanted, that'd be great (my first draft in which I literally removed all the `is_uint()` calls was not so interesting.

nspark:
"My" test looks good! 👍

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -951,27 +947,48 @@ static void build_enum_cast_function(EnumType* et) {
...
+  bool isAssign = (prim == PRIM_ASSIGN);
 
-  FnSymbol* fn = new FnSymbol("=");
-  fn->addFlag(FLAG_ASSIGNOP);
+  FnSymbol* fn = new FnSymbol(op);
+  if (prim == PRIM_ASSIGN) {
+    fn->addFlag(FLAG_ASSIGNOP);
+  }
   fn->addFlag(FLAG_COMPILER_GENERATED);
-  fn->addFlag(FLAG_LAST_RESORT);
```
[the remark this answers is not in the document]
bradcray: I went back and forth on this question while writing the PR.  In writing it, I worried that it's more likely to result in confusion than benefit which is why I disabled it here (I also figured this would be the sort of thing that would be easier to relax later than it would be to tighten up on).  But I think you're right that it'd be consistent with what we do for other types to let users write their own comparison overloads and potentially shoot themselves in the foot.  Would you like me to put this back in?

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -951,27 +947,48 @@ static void build_enum_cast_function(EnumType* et) {
 }
 
 
-static void build_enum_assignment_function(EnumType* et) {
-  if (function_exists("=", et, et))
+// Helper function used to create "=", "==", and "!=" for enums
```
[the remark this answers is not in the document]
bradcray: The original issue that proposed this proposed only supporting `==` and `!=` on enums, which is why I didn't implement inequalities as well.  But I agree that there's an argument for doing so.  I've captured that in (new) issue #9306.

I hadn't noticed the overloads for enums in ChapelBase, thanks for pointing them out.  From what I'm seeing, it's only the param-oriented overloads that are defined in ChapelBase.chpl (and, unfortunately, until they're tightened up, they still support inter-enum comparisons on params... :( ).  Like you, seeing these raises a question for me about whether we'd prefer to support `==` and `!=` (and `=`?) on enums as module code vs. as the compiler-generated functions that I added in this PR.  I'm currently inclined towards "module code" if it doesn't cause problems (and it seems it shouldn't).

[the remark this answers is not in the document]
bradcray: It seems that a downside to defining `==(x: enumerated, y...)` in module code is that when it's defined in the compiler for each enum type, we get automatic comparisons between sync enums and enums.  Once it's in the more generic module-based form, the sync enums don't match this signature any longer, so for example test/types/enum/bradc/dispatch/syncenum.chpl:22 doesn't resolve any more.  Potential fixes:  (1) add [sync] enum vs. [sync] enum overloads to ChapelBase.chpl (3 more copies of each function), (2) improve ability for sync x to dispatch to generic 'x' arguments, (3) keep defining these functions in the compiler.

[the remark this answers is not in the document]
mppf: I bet (2) is not that hard. See PR #9322.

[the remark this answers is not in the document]
bradcray: Cool!  Though not directly related, this makes me wonder whether you happened to see my issue #9310 yesterday and whether it's similarly easy.  :)

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -829,7 +825,7 @@ static void build_enum_enumerate_function(EnumType* et) {
   fn->insertFormalAtTail(arg);
   fn->where = new BlockStmt(new CallExpr("==", arg, et->symbol));
 
-  et->symbol->defPoint->insertAfter(new DefExpr(fn));
+  baseModule->block->insertAtTail(new DefExpr(fn));
```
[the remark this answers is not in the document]
bradcray: Since you made this comment, I moved `==`, `!=` and `=` on enums out into module code.  This left the enumeration routine as the only one of the compiler-generated routines that was being inserted at the enum's scope rather than the global scope, so this improves consistency.  That said, I hear what you are saying about making methods, and perhaps standard operators, on types available when that type is available, but that's not something I'm going to be able to fix in this PR.

modules/internal/ChapelBase.chpl
```diff
@@ -1102,6 +1105,9 @@ module ChapelBase {
   inline proc _cast(type t, x: complex(?w)) where isImagType(t)
     return let xim = x.im in __primitive("cast", t, xim);
 
+  inline proc _cast(type t, x: enumerated) where isImagType(t)
+    return x:int:imag;
```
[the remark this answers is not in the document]
bradcray: Good catch.  Arguably we had this problem before when the enum->imag cast was (presumably) relying on enum->int coercions, but nice to improve it while here.

daviditen:
> @daviditen: @mppf's review revealed that I had accidentally failed to support comparisons on mixed enum types if they were param values. When I fixed that, your paramenum.chpl started failing (correctly), so I split it into two tests. Would you review commit b0879f3 to make sure you're OK with what I've done?

The test change in that commit looks fine to me.

bradcray:
I ran --baseline and --verify testing on test/types/enum which came back clean.  I could do more, but it seems like the net changes aren't really doing much with the IR other than simplifying it an moving the enumeration iterator, so I'd be surprised if there were additional issues lurking.

mppf:
Seems reasonable to me. I just try not to break every test in those configurations.

PR #9199, opened 2018-04-13
Be conservative about IPA (cp PR #9115)

awallace-cray:
We don't always control which version of CCE is being used. This change conservatively backs off interprocedural analysis (IPA) optimizations slightly in case someone is using a version that has had difficulty with high levels of IPA.

Verified to work and to avoid under- or over-optimizing on x86-64 and ARM Cray systems.


awallace-cray:
git cherry-pick 4a20fb8

PR #9200, opened 2018-04-13
Fix debug build problem and slightly restructure (cp PR #9162)

awallace-cray:
The mem_regions_lock() and mem_regions_unlock() functions were added to
encapsulate entering and exiting the critical section for memory region
updates.  Unfortunately we placed them at a point at which they wouldn't
compile if DEBUG were turned on, because of an implicit reference to a
variable which hasn't been defined yet.  Here, rename them to match the
other regMem*() functions and move to be with those.

While here, also move can_task_yield() (which was created at the same
time) near where it's used.


awallace-cray:
git cherry-pick 8f48c96a1742ef088c0528d385ea3881da7df43a

PR #9201, opened 2018-04-13
In comm=ugni, skip unregisterable preexisting memory segments that don't matter (cp PR #9168)

awallace-cray:
In the past there have been unexpected changes with respect to which
preexisting memory segments we could register, and recent pre-release
experience with an upcoming CLE version indicates this will continue.
So here, rewrite the code that registers preexisting segments so that it
is less fragile.  Specifically, attempt to register each segment as we
process it, instead of doing them all at the end, and only report an
internal error and halt if registration fails on a segment we really
expect we should be able to register and which has functional or
performance implications.  Basically this comes down to those with
"huge" anywhere in their pathname.


awallace-cray:
git cherry-pick 1859860e984fe5237734baabd9d65f0b4abbcce1 0f01ce4d04e95e71b98faa72895ef7ca44e632b1

PR #9202, opened 2018-04-13
Be sure to compute max comm domains before needing it (cp PR #9172)

awallace-cray:
Recent (last several months) code movement led to a situation in which,
for minimal-registered-memory mode, we used the job-wide maximum number
of comm domains per node before we had computed that.  Here, move the
computation and jobwide distribution of that value prior to the point at
which we need it.

Subsequent testing revealed an additional bug: we have been using the
number of logical processors when estimating how many comm domains we
will need.  Using the tasking layer's max level of parallelism is much
better.

That in turn pointed out yet one more bug, less serious but still noisy:
the check to make sure the pool of remote fork completion flags was big
enough was comparing the expected maximum demand for flags to the number
of flags per pool, not the total number of flags in all pools.  So here,
fix that as well.

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -2088,10 +2088,10 @@ static void compute_comm_dom_cnt(void)
     }
 
     if (comm_dom_cnt == 0) {
-      uint32_t num_PUs;
+      uint32_t maxPar;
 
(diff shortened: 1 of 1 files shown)


awallace-cray:
git cherry-pick 88e11f1e2e91b39fa3439c59afcd0e9d13d423a0 3d20a072f40bdb5e36122e579736ab19fe8e6f05 3fe0efccd1b4bfa8dac2523a80ef85f4ae250f8e

Issue #9025, opened 2018-03-29
Add flag that will end compilation at or before a specified pass


LouisJenkinsCS:
I believe there should be a flag that can be used to end compilation early. Currently there are flags to disable specific optimizations, to disable all optimization, and even to stop before code generation. A flag that stops compilation at a particular point helps compiler developers test up to a specific pass without having to wait for everything to finish or have to filter their results. It can also benefit developers that interact with the compiler as it provides more customization.

A proposed flag could be something like `--stop-after-pass <passname>`.

Edit:

Related issue: #9003

mppf:
Sounds fine to me, and relatively simple to implement. I suspect there would be no objections to adding such a flag as a developer-only flag.

LouisJenkinsCS:
I believe I have found where to implement `--stop-at`, which is [runpasses.cpp:168-188](https://github.com/chapel-lang/chapel/blob/79a44c6edd6590f4f753b011af3ca99aad5f68c9/compiler/main/runpasses.cpp#L168-L188). I can see about adding some code that obtains the index of the desired pass if it exists and loops up to that instead of `passListSize`, should be easy enough to do.

LouisJenkinsCS:
Added the flag `--stop-after-pass <passname>` and I handle it appropriately. Will be in PR once finished all fixes and changes.

bradcray:
@LouisJenkinsCS: Can this issue be closed now that PR #9206 is merged?

LouisJenkinsCS:
Yep, forgot about it

Issue #9024, opened 2018-03-29
Add a variant of '--ignore-errors' that only ignores user errors


LouisJenkinsCS:
I believe that there should be a variant that ignores user errors but will still halt on internal errors. This way if there is an internal error that occurs from skipping a user error, there may be some undefined and non-deterministic behavior, but it will at least allow the user to see multiple errors in the best-case. I believe this change can be performed easily enough without much change to the Chapel compiler and that it can only be a positive thing to provide this option.

bradcray:
Note that even skipping past a user error can result in undefined and nondeterministic behavior since the compiler will have to proceed with potentially incorrect or incomplete information.  Maybe put another way: while I wouldn't be opposed to adding this proposed flag (or maybe just changing ignore-errors take a 1 vs. 2 level indicator), I don't think that this should remain a developer flag because I think it's more likely to lead to pain and confusion than benefit.  If there are cases where it's known to be safe to proceed after an error is triggered, the error itself should be changed to not be a "halt compilation" error.

mppf:
> I don't think that this should remain a developer flag because I think it's more likely to lead to pain and confusion than benefit.

I suspect you meant: *I think this it should remain a developer flag*.

LouisJenkinsCS:
This can be done by having it check whether `ignore_user_error && err_user` is set [here](https://github.com/chapel-lang/chapel/blob/d9288b53e24eea3782cf8b87706ada257b09b1f2/compiler/util/misc.cpp#L417-L419). Although if `tag == 2` it seems that [coverity will panic](https://github.com/chapel-lang/chapel/blob/694fea052e17da0634eb6797e0a85d38fc03bffa/util/devel/coverity/coverity_model.cpp#L23-L25), although this may not be important at all.

mppf:
We used to run coverity testing regularly but stopped doing so.

LouisJenkinsCS:
While I am able to see multiple errors, I'm unable to see all of them... this is still a significant improvement (See screenshot in Issue #9003) so I'll consider it 'good enough' for now.

LouisJenkinsCS:
Yep, sorry forgot about this one.

PR #9206, opened 2018-04-14
Add '--ignore-user-errors' and '--stop-after-pass' compiler flags

LouisJenkinsCS:
Introduced new flag, `--ignore-user-errors` which functions similar to `--ignore-errors` where it will continue compilation after discovering a user error, but unlike `--ignore-errors` it will stop at internal errors. The rationale for such a change is discussed in issue #9024.

Also introduced is a new flag, `--stop-after-pass` which will cease compilation after the requested pass, which must match in name. If a pass is declared, it will check on each pass whether or not the names match and if so it will 'short-circuit' compilation to end early. The rationale for such a change is discussed in issue #9025 

@mppf @bradcray


mppf:
You need to run
```
util/devel/gen-chpl-bash-completion > util/chpl-completion.bash
```
and commit the result.

Verify it worked with
```
./util/start_test test/chplenv/autocomplete/autocomplete.chpl
```

mppf:
@LouisJenkinsCS - that last commit added conflicts. Recommend undoing it, rebasing onto current master, and then doing it again.

PR #9212, opened 2018-04-15
Test strings with initializers in perf playground

benharsh:


benharsh:
Observed some regressions for release versions of regex-dna, but not knucleotide, fasta, or revcomp. Wondering if there are any others.

Issue #9171, opened 2018-04-11
Should string's init-copy create a deep copy by default?


benharsh:
In the old initCopy world, string's [initCopy](https://github.com/chapel-lang/chapel/blob/d67e929695e8bc66d11740517733717111da7867/modules/internal/String.chpl#L1381) would only create a deep copy if the original string was owned. In the initializer world, that would look something like:
```chpl
proc init(s:string, owned : bool = s.owned);
```

Yet there was a string constructor that by default always created a copy:
```chpl
proc string(s:string, owned : bool = true);
```
This string constructor should become the copy-initializer, but that would change memory usage (and perhaps program behavior).

Should string's copy-initializers create a deep copy by default?

bradcray:
I'd think "by default, yes."  As in other cases, I'd hope that we'd eventually have a way to steal the string from a dead RHS.  I'm not convinced the `owned` argument is the right way to do this, however.

mppf:
I would think no, because the common case for an "unowned" string is in fact a string literal. These have no relevant "ownership" and are just available for the whole program run. Shouldn't copying a string literal result in something that shares the (const) string literal?

``` chapel
var x = "hi";
var y = x; // Why can't this also point to the constant literal "hi" ?
```

mppf:
I think the problem in this case is the string constructor's awkward use of default arguments.

benharsh:
Whatever the case, I believe this has been settled for 2.0. At a glance, it looks like we always copy now. My guess is we handle string literals through some of the move-steal analysis.

Issue #9007, opened 2018-03-28
Resolve double-free when initializing from casted record


benharsh:
### Summary of Problem
If a variable is initialized from an expression that casts a record of type ``T`` to ``T``, the compiler fails to insert a call to the copy initializer.

### Steps to Reproduce

**Source Code:**

```chapel
record R {
  var x : int;

  proc init() {
    writeln("R.init()");
  }
  proc init(r:R) {
    this.x = r.x;
    writeln("R.init(R)");
  }
  proc deinit() {
    writeln("R.deinit");
  }
}

proc bad(n:int) {
  var err_args : R;
  for i in 1..n do err_args.x += 1;

  const ret = err_args:R;
  return ret;
}

proc main() {
  var b = bad(10);
  writeln("b = ", b);
}
```
Output:
```
R.init()
R.deinit
b = (x = 10)
R.deinit
```
In this example we expect to see a call to "R.init(R)"

### Configuration Information

- Output of `chpl --version`: 1.17.0

benharsh:
Thinking about this some more, I'm not sure what the semantics really are if a variable is cast to the same type as itself.

```chpl
const foo = original:R;
```

In this case it seems reasonable to me that ``foo`` is some kind of copy of ``original``. Whether it's from the ``_cast`` function or ``R.init(R)`` I'm not exactly sure.

What about this case?

```chpl
(original:R).something(); // equivalent to ``original.something()`` ??
```

Is ``something`` invoked on a copy of ``original``?

bradcray:
> I'm not sure what the semantics really are if a variable is cast to the same type as itself.

My general intuition is that casts create new values, so by default, I'd tend to expect that to happen when casting something to its own type as well.  I could imagine the compiler optimizing the copy away if it found it was legal / unobservable for it to do so.

PR #9217, opened 2018-04-16
Update strings to use initializers

### LINKED ISSUE #9006: Convert strings to use initializers
As a Chapel implementer, I want strings to use initializers because strings are frequently used and would help to demonstrate initializer stability.

Acceptance criteria: Replacement of 'proc string' constructors with initializers.

### LINKED ISSUE #9173: Double-free for variables with declared type initialized from call
### Summary of Problem
A variable with a declared type fails to copy the initialization expression if the expression is a field access:
```chpl
const temp : R = myClass.getMyR;
```

### Steps to Reproduce

**Source Code:**

```chapel

record R {
  var x : int;
  proc init() {
    writeln("R.init()");
    this.x = 0;
  }
  proc init(other:R) {
    this.x = other.x + 1;
    writeln("R.init(R): ", this.x);
  }
  proc deinit() {
    writeln("R.deinit(): ", this.x);
  }
}

class C {
  var x : R;

  proc init() {
    writeln("start: C.init()");
    this.x = new R();
    writeln("end: C.init()");
  }
}

var c = new C();
const x : R = c.x;
writeln("temp = ", x);
delete c;
```

Output:
```
start: C.init()
R.init()
end: C.init()
temp = (x = 0)
R.deinit(): 0
R.deinit(): 0
```

We should see a matching number of "R.init" and "R.deinit" lines. In the statement ``const x : R = c.x;``, the compiler simply moves ``c.x`` into ``x`` without copying. Note that leaving the type off, ``const x = c.x``, results in correct behavior.

benharsh:
This PR converts strings to use initializers. The autoCopy/initCopy functions for strings can now be retired.

The old initCopy would only create a deep copy if the original string was owned. The new copy-initializer for strings will always make a deep copy by default. This results in increased memory usage. See Issue #9171 for further discussion.

This work exposed bugs #9007 and #9173.

Some tests under memory/figueroa were incorrectly measuring memory usage instead of memory leaks. These tests were either updated to reduce the memory usage or removed entirely.

Some compiler quirks that were worked around:
- param strings don't use initializers
- worked around issue with string mismatching to c_string's initCopy
- string literals needed to be worked around in a hacky manner
- Fix invalid AST for a deref

Resolves #9173 by replacing a PRIM_MOVE with a PRIM_INIT_VAR.
Resolves #9006.

Testing:
- [x] local + futures
- [x] gasnet
- [x] valgrind
- [x] memleaks
- [x] verify dead-string-literal elimination still works
- [x] string shootout performance


benharsh:
@lydia-duncan : would you mind reviewing? I still need to look at perf playground results (tomorrow)

compiler/AST/symbol.cpp
```diff
@@ -1531,7 +1524,23 @@ VarSymbol *new_StringSymbol(const char *str) {
...
+    newFirst = new UnresolvedSymExpr("string");
+  }
+
+  CallExpr *initCall = new CallExpr(PRIM_NEW,
+      newFirst,
+      castTemp,
+      new_IntSymbol(strLength),   // length
+      new_IntSymbol(strLength ? strLength+1 : 0)); // size, empty string needs 0
+  initCall->insertAtTail(gFalse); // owned = false
+  initCall->insertAtTail(gFalse); // needToCopy = false
```
lydia-duncan: I'm not sure why these aren't included in the CallExpr when you create it, since it already takes a line per argument.

lydia-duncan: Does the CallExpr constructor only accept five arguments?

compiler/resolution/ResolutionCandidate.cpp
```diff
@@ -603,7 +604,11 @@ bool ResolutionCandidate::checkResolveFormalsWhereClauses(CallInfo& info) {
 
       bool promotes          = false;
 
-      if (actualIsTypeAlias != formalIsTypeAlias) {
+      if (isInitCopy && isString(actual) && formal->getValType() == dtStringC) {
+        // Do not allow an initCopy of a string to find the c_string initCopy,
+        // which is considered first because it is not compiler generated.
```
lydia-duncan: Would it be harmful to mark the c_string initCopy as "last resort"?  Or would it just not resolve the problem?

benharsh: I initially tried that, but when it didn't work I just did this. I'm not very familiar with the last-resort stuff, so I may have messed something up.

lydia-duncan: It's probably because the compiler generated initializer for strings is also marked last resort so it thinks there is an ambiguity due to the coercion.  Oh well

benharsh:
Due to internal testing infrastructure issues, I decided to just check the string shootouts (fasta, revcomp, regexdna, knucleotide). Performance looked good.

PR #9229, opened 2018-04-16
Add error on --warn-unstable for undecorated new

### LINKED ISSUE #9101: delete free subtask: compile error on un-decorated 'new'
As a Chapel implementer, I need the compiler to be able to locate 'new' calls that will change behavior as the delete-free features are implemented so that I can update these appropriately.

As a Chapel user, I need to be able to see which parts of my program need to change in order to work with the changes to 'new' and class types.

mppf:
Compiler raises an error for a `new` that is not `new owned`/`new shared`/`new unmanaged` when the flag `--warn-unstable` is used.

Closes #9101.

- [x] full local testing

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -6017,7 +6017,7 @@ static void handleUnstableNewError(CallExpr* newExpr) {
         if (se == checkSe) {
           // OK
         } else if (CallExpr* parentCall = toCallExpr(se->parentExpr)) {
-          if (parentCall->isNamed("chpl__toraw") || // TODO -- remove case
+          if (parentCall->isNamed("chpl__tounmanaged") || // TODO -- remove case
               parentCall->isNamed("chpl__delete") || // TODO -- remove case
               parentCall->isNamed("chpl__buildDistValue") ||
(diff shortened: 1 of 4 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -5989,6 +5994,53 @@ static void resolveNewManaged(CallExpr* newExpr, Type* manager, CallExpr* manage
...
+          } else if (parentCall->isPrimitive(PRIM_NEW) &&
+                     parentCall->get(1)->typeInfo()->symbol->hasFlag(FLAG_MANAGED_POINTER)) {
+            // OK e.g. new Owned(new MyClass())
+          } else if (parentCall->isPrimitive(PRIM_TO_UNMANAGED_CLASS)) {
+            // OK e.g. new raw MyClass() / new owned MyClass()
+          } else {
+            // TODO -- enable warning for internal/standard modules
+            // along with updating them.
+            if (newExpr->getModule()->modTag == MOD_USER) {
+              USR_WARN(newExpr, "new in is unstable - "
```
vasslitvinov: "new in is unstable" does not flow smoothly IMO. How about "new \<class name> is unstable" ?

Correspondingly, the next line would read "use 'new raw \<class name>' or .....".

Issue #9210, opened 2018-04-15
GASNet ibv-conduit performance fix available upstream


PHHargrove:
We recently discovered and fixed an issue in GASNet's InfiniBand support on modern HCAs which was leading to Gets under-performing Puts by nearly a factor of 2.

We encourage you to cherry-pick [this commit](https://bitbucket.org/berkeleylab/gasnet/commits/5cb86a7e), which provides more details.  
Alternatively, `GASNET_MAX_MTU=-1` in the environment should have the same effect.

ronawho:
Thanks for the heads up @PHHargrove. I'll bring that commit in this week.

ronawho:
For my future self, since I can never remember how I bring in third-party patches while maintaining committer info: 

```bash
commit=<upstream sha>
wget https://bitbucket.org/berkeleylab/gasnet/commits/${commit}/raw -O ${commit}.diff
git am --directory third-party/gasnet/gasnet-src/ ${commit}.diff
```

PR #9255, opened 2018-04-18
Pull in an upstream GASNet patch for an InfiniBand performance improvement

ronawho:
Pull in an upstream GASNet patch for an InfiniBand performance improvement.
This pulls in:
 - https://bitbucket.org/berkeleylab/gasnet/commits/79315e5fe
 - https://bitbucket.org/berkeleylab/gasnet/commits/5cb86a7e0

Note that the first commit is just a comment change, but without it the real
patch doesn't apply cleanly.

Closes https://github.com/chapel-lang/chapel/issues/9210


ronawho:
@gbtitus just FYI

PR #9257, opened 2018-04-18
Fix chpl tab-completion for home directories with '~'

daviditen:
The tab-completion script was not expanding paths leading with tilde, so it
didn't think e.g. "~/chapel/test/release" was a directory.  Expand this to
an absolute path before trying to decide if it is a directory or not.


daviditen:
Thanks!

PR #9262, opened 2018-04-18
Update miscellaneous standard/package types to use initializers

benharsh:
This PR updates remaining classes and records in standard and package modules to use initializers, either explicitly or via the pragma.

Testing:
- [x] local + futures
- [x] gasnet


benharsh:
@mppf : Could you review the changes in IO and Regexp? The rest seem trivial to me.

modules/standard/IO.chpl
```diff
@@ -1255,16 +1255,22 @@ record file {
...
-pragma "init copy fn"
-proc chpl__initCopy(x: file) {
-  on x.home {
-    qio_file_retain(x._file_internal);
+proc file.init(x: file) {
+  this.home = x.home;
+  this._file_internal = x._file_internal;
+  this.complete();
+  on home {
+    qio_file_retain(_file_internal);
```
mppf: Just curious - why is the 'this.complete' needed, if it is?

benharsh: Hmm, not sure... I thought it was needed, but trying out a few tests it doesn't appear to be necessary.

PR #9271, opened 2018-04-19
Correct how we set the locale name with gasnet local spawning

ronawho:
Correct how we set the locale name. While testing #9148 I noticed that we were
using `<node_name>-<node_id>` instead of just `<node_name>` for configs like
gasnet_ibv.

Previously we only checked for comm=gasnet and spawnfn=L, but this got a few
cases wrong. Specifically, spawnfn only means anything for the udp substrate,
so we don't want to add a node_id for something like ibv. We also weren't
adding the node_id to the gasnet-smp case.

The node/locale name is now `<node_name>-<node_id>` for gasnet-smp or
gasnet-udp w/ GASNET_SPAWNFN=L. Otherwise it is just `<node_name>`

This is a logical followup on #9176. Before that PR we never detected
local-spawning because of an incorrect sys_getenv. Fixing that bug exposed this
bug in the implementation of locale name detection.

### DIFF
--- a/modules/internal/LocaleModelHelpSetup.chpl
+++ b/modules/internal/LocaleModelHelpSetup.chpl
@@ -111,10 +111,10 @@ module LocaleModelHelpSetup {
     // current node.  For this reason (as well), the constructor (or
     // at least this setup method) must be run on the node it is
     // intended to describe.
-    var comm, spawnfn : c_string;
+    var spawnfn: c_string;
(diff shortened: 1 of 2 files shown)


ronawho:
Ran `test/localeModels/dmk/locale_name` with smp, udp (local and real spawning) as well as with ibv. 

Ran a full paratest for udp as well.

PR #9272, opened 2018-04-19
Split against all instances of = in overrides.py

dmk42:
#9013 tried to protect against environment variables that contained equals signs, but may have gone too far.  This reverts a portion of that change by having overrides.py split on all occurrences of the equals sign, not just the first one.

### DIFF
--- a/util/chplenv/overrides.py
+++ b/util/chplenv/overrides.py
@@ -133,7 +133,7 @@ class ChapelConfig(object):
                 if self.skip_line(line, linenum):
                     continue
 
-                var, val = [f.strip() for f in line.split('=', 1)]
+                var, val = [f.strip() for f in line.split('=')]
                 self.chplconfig[var] = val
 
     def skip_line(self, line, linenum):
@@ -147,7 +147,7 @@ class ChapelConfig(object):
 
         # Check for syntax errors
         try:
-            var, val = [f.strip() for f in line.split('=', 1)]
+            var, val = [f.strip() for f in line.split('=')]
         except ValueError:
             self.warnings.append(
             (


dmk42:
@ronawho and @ben-albrecht - This is one possible solution to the failing test.  Is this the right thing, though?

ronawho:
Another solution that makes a little more sense to me would be to split on `' = '`.

dmk42:
Splitting on `'='` and allowing multiple `=` signs passes a full paratest.  Splitting on `' = '` fails `chplenv/chplconfig/correctness/correctness`.  Not that I understand it.

ronawho:
Huh ok, shows what I know. Looks like `val=var` is the documented format, but the scripts allow for an arbitrary amount of whitespace around the `=`.

Maybe just merge this to quiet testing, and open up an issue for how to address it "correctly"? (or wait for @ben-albrecht to comment, he knows this code best)

dmk42:
Thanks.  I will wait for @ben-albrecht so I can get more eyes on this.

ben-albrecht:
> but may have gone too far.

Could you elaborate on what went wrong with the previous split?

dmk42:
If any environment variables contain an `=` sign then we don't see them correctly.  That's what I was trying to correct.

I originally had an `=` sign in `CHPL_TARGET_BACKEND_ARCH`, which is how I found out about the problem.  Later I switched to a different strategy, but left the protection against environment variables containing an `=` sign in, because nobody knows what environment variables that we don't control might contain.

ben-albrecht:
Oh I think I understand now.

The test failed, because the new behavior accepted `CHPL_TASKS == fifo` in a `chplconfig` file instead of printing a warning that the incorrect syntax was used.

Would it be reasonable to still split on the first `=`, but then print a warning if a variable began with the character `=`?

dmk42:
I agree with both of you.

- It would be reasonable to still split on the first `=`, but then print a warning if a variable began with the character `=`.

- Whatever we do, the behavior should be similar between an env var and chplconfig.

ben-albrecht:
> I think my preference would be to leave things as they are now (`CHPL_TASKS==fifo` in env or config file results in `CHPL_TASKS: =fifo`, and just update the test.

I'm fine with this actually. I think it'd be pretty easy to catch this error without a special case error message, as soon as you tried to compile something or called `printchplenv`

ben-albrecht:
> I could merge this now, to stop the errors, and @ben-albrecht could implement the long-term desired behavior at his leisure without a deadline, or

Let's do this. LGTM as is.

dmk42:
Great, thanks to both of you!

PR #9275, opened 2018-04-19
Update various internal types to use initializers

benharsh:
These various internal types are fairly simple and not 'special' like tuples or arrays, and so are easy to convert to initializers.

Testing:
- [x] local + futures
- [x] gasnet
- [x] multilocale task-spawn perf
- [x] single-locale task-spawn perf

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -1008,6 +1008,7 @@ module ChapelArray {
   }
 
 
+  pragma "use default init"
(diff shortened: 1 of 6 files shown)


benharsh:
@ronawho : given your past encounters with EndCount, do you think it's worth doing some perf testing on this branch?

ronawho:
I don't think it's worth doing full perf testing, but I would spot check the empty-task-spawn tests for single-locale and multi-locale.

PR #9103, opened 2018-04-05
Fix problem with --llvm and GASNet with MPI

mppf:
As described in issue #9097, when using --llvm,
the MPI library is not linked with the resulting
executable, resulting in link errors for some GASNet
configurations.

These errors don't come up without --llvm because the
GASNet Makefiles set LD to e.g. mpicxx. This commit
adjusts the --llvm support to take similar action by
saving the LD to invoke in a lib/.../override-ld file
(similarly to how list-libraries is handled) so that
make need not be invoked during a chpl --llvm compile.

This can be viewed as a follow-on to PR #7757.

Fixes #9097.

- [x] test on a Cray XC
- [x] hello compiles and runs with GASNet ibv and --llvm
- [x] full local testing

Reviewed by @dmk42 - thanks!

PR #9277, opened 2018-04-19
Fix replace-paths to avoid normalizing file name

mppf:
This avoids a problem on Mac OS X where clang++ turns
into clang-6.0 for --llvm configurations.
(That happens because clang++ is a symbolic link pointing
 to clang-6.0).

This can be viewed as a follow-on to #7801 and #9103.

Reviewed by @dmk42 - thanks!

- [x] full local testing
- [x] verified prefix installation works with --llvm on Mac OS X including with a regexp test

### DIFF
--- a/util/config/replace-paths.py
+++ b/util/config/replace-paths.py
@@ -48,6 +48,14 @@ if __name__ == '__main__':
 
         origpath = m
         path = os.path.realpath(origpath)
+
+        if os.path.isfile(path):
+          # Leave out the filename, since we're only trying to replace
+          # directories. This prevents clang++ from being replaced
+          # with e.g. clang-6.0 when clang++ is a symbolic link.
+          origpath = os.path.dirname(origpath)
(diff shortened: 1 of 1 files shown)


util/config/replace-paths.py
```diff
@@ -48,6 +48,14 @@
 
         origpath = m
         path = os.path.realpath(origpath)
+
+        if os.path.isfile(path):
+          # Leave out the filename, since we're only trying to replace
+          # directories. This prevents clang++ from being replaced
+          # with e.g. clang-6.0 when clang++ is a symbolic link.
+          origpath = os.path.dirname(path)
+          path = os.path.dirname(path)
```
dmk42: Shouldn't `path = os.path.realpath(origpath)` to get the canonicalized path name?  (That's a question, not a request.  I'm not sure I fully understand the original code.)

mppf: yes it should thanks for spotting that

PR #9286, opened 2018-04-20
Add the correct fltk tarball (cp PR #9264)

awallace-cray:
Somehow I added an uncompressed (and misnamed) tarball in #7871. Add the right
one this time. Without this the `tar xzf` fails on linux, though it happens to
work on mac, which is where I tested the original PR.

See https://sourceforge.net/p/chapel/mailman/message/36297481/ for more info


awallace-cray:
git cherry-pick 607d8ce038147145678166283db26863905b191a

PR #9298, opened 2018-04-20
Add annotations for last week

lydia-duncan:
Covers a bunch of memory leak changes, and various initializer performance
impacts, plus notes that a couple of tests got impacted by lowering ForallStmts
at a different point in the compiler.


lydia-duncan:
Passed annotation checker

Issue #9299, opened 2018-04-20
uniquifyName in presence of extern symbols


vasslitvinov:
When an extern symbol has the same name as another symbol that has already been seen, we want to make sure that either both symbols refer to the same extern symbol (i.e. both are functions, or both are types, or both are variables of the same type) or that the earlier symbol is a non-extern/non-export and is uniquified properly. For example:

```chpl
{
  var my_extern_var: int;  // not an extern, should be uniquified
  USEME(my_extern_var);
}
{
  extern var my_extern_var: int;
  USEMEb(my_extern_var);
}
```

Currently the compiler, at best, detects this situation. (It does not, for example, when the symbols in question are globals.) If detected, it aborts instead of handling it. If not detected, incorrect code may be generated where different variables are conflated due to the same C name.

Tests:

```
test/users/vass/uniquify/u1-local-user-extern.chpl
test/users/vass/uniquify/u2-local-extern-extern.chpl
test/users/vass/uniquify/u3-forall-shadow-vars.chpl
```

Note that u3-forall-shadow-vars.chpl may not test this issue, depending on how in-intents are handled by the compiler.

PR #9300, opened 2018-04-20
improve uniquifyName w.r.t. externs etc.

### LINKED ISSUE #9223: add name munging to lowering of ForallStmts
As #9205 shows, lowerForallStmtsInline() may result in incorrect generated C code because of not munging the name of a local variable.

This task is to add munging in appropriate cases, so that the workaround in #9205 can be reverted. This will also prevent similar bugs in the future.

vasslitvinov:
As worked around in #9205, the compiler failed to uniquify a Symbol's name
when there was another Symbol with the same name that was an 'extern'
or 'export'

This PR fixes that issue.

It also adds detection of the situation where the extern/export symbol is
encountered **after** the like-named non-extern/export symbol was encountered.

* When detected, the compiler crashes with INT_FATAL. Proper handling
  of these cases is left as future work.

* This detection fires only for local variables. This is because
  this situation currently happens a lot for global types, functions
  and variables, ex. c_pointer_return, qbuffer_ptr_t, QBUFFER_PTR_NULL.
  We do not want the compiler to crash because of that.

See also #9299.

#### DETAILS

uniquifyName() records the names it has seen.
The names of non-renameable (i.e. extern/export) symbols
were ignored so conflicts with those went undetected/unresolved.
With this PR, these names are recorded, too.

While there:

* Change the interface of uniquifyName() to accept the Symbol
and update its cname directly.

* Move the isRenameable() check into uniquifyName().
That way it is applied to all symbols uniformly. Most importantly,
this check is not applied to local variables in functions,
which could be marked as 'extern'.

Testing:
- [x] linux64
- [x] gasnet - multilocale tests (a previous version)
- [ ] numa
- [ ] numa gasnet - multilocale tests
- [ ] llvm

Resolves #9223.


compiler/codegen/codegen.cpp
```diff
@@ -953,10 +953,48 @@ static const int maxUniquifyAddedChars = 25;
...
+    // to avoid generating erroneous C code silently.
+    // Do this only for things local to a function, as we use
+    // multiple Symbols for the same extern at the global scope,
+    // ex c_pointer_return, qbuffer_ptr_t, QBUFFER_PTR_NULL.
+    //
+    if (set2 && (set1->find(name) != set1->end()) )
+      INT_FATAL(sym, "name conflict with a non-renameable symbol");
+
+    set1->insert(name);
+    return;
```
bradcray: To make the control flow of this function clearer, I'd recommend taking the approach of:

```c
  if (sym->isRenameable()) {
    ... do all the normal stuff ..
  } else {
    ... do the int fatal if the symbol's already in the set ...
  }
  set1->insert(name);
}
```

such that the actions that are different in each case are separated from the one that's common (inserting the name into the set).

PR #9305, opened 2018-04-23
Add asserts about A's domain in arrayVectorOps primer

daviditen:
Add asserts about what A.domain is after each change.  This way it will show
up in the code blocks of the online version of the primer.

Verified that the test still passes and that the generated document looks as
expected.


daviditen:
Thanks @vasslitvinov!

PR #9317, opened 2018-04-23
Update tuples to use special function instead of constructors

benharsh:
We aim to eventually deprecate constructors, so this commit updates
tuples to use a special function instead of a constructor. Tuples are
sufficiently special that I don't think it makes sense to use an
initializer just yet.

This special function is named 'chpl__init_tuple', and is generated by
the compiler in the same way the old constructor was generated. Various
conditionals in the compiler have been updated to look for FLAG_TUPLE
instead of a function name.

Testing:
- [x] local + futures
- [x] gasnet
- [x] --verify on release/examples/ and types/tuples/
- [x] --baseline on release/examples/ and types/tuples/
- [x] 16-node-xc miniMD performance (uses tuples)

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -1354,6 +1354,11 @@ void AggregateType::buildConstructor() {
     }
   }
 
+  // Don't use initializers or constructors for tuples.
+  if (symbol->hasFlag(FLAG_TUPLE)) {
+    return;
+  }
+
   // Create the default constructor function symbol,
   FnSymbol* fn = new FnSymbol(astr("_construct_", symbol->name));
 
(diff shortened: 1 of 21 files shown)


benharsh:
@mppf : would you be willing to review? Do you (or anyone else) think I should do more testing?

mppf:
Oh I see, build_tuple is probably calling init tuple...

benharsh:
Yes, build_tuple does call _construct__tuple/chpl__init_tuple.

It seemed odd to me as well to have these two functions, but that's sort of a separate issue from initializers/constructors and so I just worked through it.

I'll work on adding ``FLAG_INIT_TUPLE``. It doesn't catch the type constructor because we don't explicitly add FLAG_TUPLE when creating the type constructor. The only things with FLAG_TUPLE (at the time of this comment on this branch) are the '_tuple' type symbol and chpl__init_tuple.

benharsh:
@mppf : are you satisfied with the last couple of commits and the amount of testing?

mppf:
@benharsh - yes but I suppose I'm still surprised that tuple type doesn't need ->defaultInitializer, but if it works, it works.

PR #9319, opened 2018-04-24
Fix a typo/think-o in primers.

vasslitvinov:
Replaced an unintended reference to a global variable with 'this'.

// This function with this glitch was introduced back in 0dcb0c3504f.


vasslitvinov:
@daviditen would you double-check?
Passes start_test in linux64 configuration.

PR #9102, opened 2018-04-05
Replace usage of easy_install with pip

ben-albrecht:
Use curl to install `get-pip.py` to install pip to install virtualenv instead of using easy_install to install virtualenv. It's more sane than it sounds.

This fixes `make test-venv` and `make chpldoc` for OS X users using system python.

Closes #9095 

## Testing

```
cd third-party/chpl-venv
make test-venv
make chpldoc-venv
```
### Crays
- [x] Cray (CLE 5, python 2.6)
  - Requires override with `CHPL_PIP` (outdated TLS)
- [x] Cray (CLE 6, python 2.7)
- [x] Cray module build process: Build on CLE 5, Run on CLE 5/6

### Other OSes
- [x] OS X (python 2.7)
- [x] linux64 (SLES 12, python 2.7)
- [x] linux64 (SLES 11, python 2.6)
  - Requires override with `CHPL_PIP` (outdated TLS)
- [x] Cygwin (python 2.7)

## Outside of this PR

- [x] Make common `pip` available in for testing infrastructure
  - Testing must set `CHPL_PIP` to this `pip` installation to override the use of `curl`
- [x] Update Jenkins to define `CHPL_PIP` and `PYTHONPATH` on specific systems

## Followup work

- Update `prereqs.rst` to mention `curl` as a dependency
- Add better error message for when `curl` fails (suggest `CHPL_PIP`)
- Refactor makefile variable names to be a little more cohesive.

PR #9330, opened 2018-04-24
Use CHPL_PYTHONPATH for python -S calls

ben-albrecht:
Relying on changing `PYTHONPATH` ended up being problematic in some testing configurations, so using `CHPL_PYTHONPATH` to override the `pip` is a more reasonable approach.

This follows up on #9102


awallace-cray:
2018-04-25 
git checkout release/1.17
git cherry-pick 1bd04cb38ba1d5b8622cccd3235557888dca7599

Issue #9037, opened 2018-03-30
support 'None' entries in Series


psahabu:
## user story
As a Chapel Programmer, I want to include `None` entries in my Series because it represents my data set better than a default value.

### acceptance criteria
- Iteration through the elements of a Series will skip `None` entires.
- Selecting a `None` element in a Series will `throw` an error.
- Otherwise, iteration and selection works as before.
- The `None` element is indicated when the Series is printed.

## design and implementation tasks
- [ ] determine if a bitset is the best way to represent `None`
- [ ] use an array of `bool` as a proof of concept bitset
- [ ] explore using an array of `int(64)` as a bitset
  - likely works best as a separate library
- [ ] address `FilterNoIndexSeries` future

PR #9342, opened 2018-04-25
arithmetic operations on Series ADT

psahabu:
### Motivation
The motivation for the Series ADT arithmetic can be found in the connected issue.

### Implementation
The `uni()` method combines the indexes of both Series while taking a `SeriesUnifier` object, which contains three overridable functions: `f(lhs, rhs)`, `f_lhs(lhs)`, and `f_rhs(rhs)`. These three functions are used to fold the values of the `lhs` and `rhs` Series, respectively.

`add()`, `subtr()`, and `mult()` utilize `uni()`, and operators (`+`, `-`, and `*`) are built on top of those functions in turn. (The nested implementation of `uni()` and the other functions is necessary for dynamic dispatch to realize the types of the Series and the Index.)

### Notes
Included in this PR are pandas equivalents to our tests. Note that the only significant difference is that, if an index is not shared between two combined Series, Python emits `NaN`. This aspect will be tackled as part of #9037.

### DIFF
--- a/test/library/standard/Dataframes/psahabu/AccessSeries.chpl
+++ b/test/library/standard/Dataframes/psahabu/AccessSeries.chpl
@@ -1,9 +1,9 @@
 use Dataframes;
 
-var I = ["A", "B", "C", "D", "E"];
 var A = ["a", "b", "c", "d", "e"];
+var I = ["A", "B", "C", "D", "E"];
 
-var letters: TypedSeries(string) = new TypedSeries(I, A);
+var letters: TypedSeries(string) = new TypedSeries(A, I);
 
 writeln("A: " + letters["A"]);
 writeln("C: " + letters["C"]);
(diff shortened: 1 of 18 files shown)


test/library/standard/Dataframes/psahabu/AccessSeries.chpl
```diff
@@ -1,9 +1,9 @@
 use Dataframes;
 
-var I = ["A", "B", "C", "D", "E"];
 var A = ["a", "b", "c", "d", "e"];
+var I = ["A", "B", "C", "D", "E"];
 
-var letters: TypedSeries(string) = new TypedSeries(I, A);
+var letters: TypedSeries(string) = new TypedSeries(A, new TypedIndex(I));
```
benharsh: Is the 'new TypedIndex' part temporary?

psahabu: As I was writing the pandas examples, I realized that it was kind of weird to pass bare arrays into the TypedSeries constructor for the index, so now I have the initializer take an Index instead. Hence `new TypedIndex()` is the Index counterpart to `new TypedSeries()` for Series.

PR #9343, opened 2018-04-25
Fix for swap of owned

### LINKED ISSUE #9260: Can't swap Owned
### Summary of Problem
`error: illegal lvalue in assignment`

### Steps to Reproduce

**Source Code:**

```chapel
use OwnedObject;

class Foo {
  var x: int;

  proc init(x: int) {
    this.x = x;
  }
}

var a = new Owned(new Foo(1));
var b = new Owned(new Foo(2));

a <=> b;
```

**Compile command:**
`chpl --force-initializers app.chpl`

### Configuration Information

`chpl version 1.18.0 pre-release (593b3d1e25)`

bradcray:
I needed a study break last night, so checked to see whether this was the five-minute fix that I expected.  It was!

Fixes #9260.


mppf:
Looks good. Yes, swap on shared would swap the instance pointer and the reference count pointer.

bradcray:
@mppf: Looks like the default `<=>` implementation on shared "just works".  Could you verify that this is behaving as expected?  If so, I'll just add a test to lock things in:

```chapel
var c = new Shared(new Foo(3));
var d = new Shared(new Foo(4));
var copyOfD = d;
var anotherCopOfD = d;

writeln(c.pn);
writeln(d.pn);

c <=> d;

writeln(c);
writeln(d);
writeln(c.pn);
writeln(d.pn);
```

output:
```
{count = 1}
{count = 3}
{x = 4}
{x = 3}
{count = 3}
{count = 1}
```

mppf:
@bradcray - yes I'd expect the default implementation to "just work" however it does more reference count twiddling than necessary so I would recommend an overload anyway. Test looks OK.

bradcray:
Makes sense, thanks.

PR #9351, opened 2018-04-25
Implement <=> for 'Shared'

bradcray:
This implements the swap operator for 'Shared' just as PR #9343 did
for Owned.  This isn't strictly necessary, as the default swap
implementation worked; but this results in a more efficient and native
implementation.

Extended my previous swap-owned.chpl file to include shared objects
and renamed it swap-managed.chpl.

While here, I realized that I'd dropped the swap operator into a
fairly arbitrary place in OwnedObject, and moved it closer to the
assignment operator overloads since they're thematically related.


bradcray:
@mppf: Needed another (mid-meeting) study break.  Care to review?

modules/internal/OwnedObject.chpl
```diff
@@ -208,6 +208,10 @@ module OwnedObject {
   proc =(ref lhs:_owned, rhs:_nilType) {
     lhs.clear();
   }
+  pragma "no doc"
```
mppf: can we chpldoc it?

bradcray: We could.  I was following the lead of the assignment overloads (in both files).

mppf: Oh, I see, owned = is not chpldoc'd. This is fine for now then.

PR #9358, opened 2018-04-26
Update prereq.rst for #9102 changes

ben-albrecht:
`chpldoc` and `test-venv` no longer require `easy_install`, so python-setuptools has been removed as a dependency. These features now require `curl` to install a local pip, so `curl` has been added as a dependency.

Note: I opted to not describe the caveat of requiring an openssl that support TLS 1.2+ on non-OS X systems, but could be convinced to do so if the reviewer(s) had a strong opinion on it.


ben-albrecht:
@bradcray && @ronawho - would you mind checking this, and expressing an opinion on my note in the PR description?

PR #9359, opened 2018-04-26
Fix a typo in the ZMQ documentation

lydia-duncan:
The interoperability example said the socket pair was PUSH-PULL, but the example
itself was REQ-REP.  Update the description to match the example.


lydia-duncan:
@nspark - this was the original intention, right?  I can also update it to go the other way (make the example match the old description) if you prefer

nspark:
👍  (You're right that the documentation should be updated to match the code; i.e., use of `REQ`-`REP` is correct.)

nspark:
Nice catch! Thanks!

PR #9360, opened 2018-04-26
Make some late-breaking CHANGES.md edits

bradcray:


bradcray:
@ben-albrecht: Would you review this?  Note that I've told @awallace-cray that we shouldn't re-spin the release over this, but we'll grab it if something else forces us to re-spin before releasing.

PR #9361, opened 2018-04-26
Update ZMQ interop-py PREEXEC to use new virtualenv.py path

ben-albrecht:
#9102 and friends fixed the heart of the issue that was causing this test to fail nightly on OS X. However, the virtualenv path in the `PREEXEC` became stale, so the test continued to fail.  This PR updates the path.

Also added a comment about my desire to have this functionality built into the test framework, rather than have test authors write `PREEXEC`s like this.

- [x] Test locally (OS X)
- [x] Test on failing machine (OS X)

### DIFF
--- a/test/library/packages/ZMQ/interop-py/PREEXEC
+++ b/test/library/packages/ZMQ/interop-py/PREEXEC
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+#Note: This is a temporary solution to supporting additional python dependencies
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Not as ugly as I thought it'd have to be. I don't think the pip-installed site packages suffers from the same issue as the `get-pip.py` pip installation, where the path can be `python` or `python<version>`, and `lib64` or `lib` depending on your OS.

Still, we should have a better story for handling python dependencies in the testing infrastructure.

Issue #9364, opened 2018-04-26
Slow Sparse Matrix Addition


Tshimanga:
I tried using `.plus()` to add a large (square 10^6 with 10^6 nonzeros) matrix to it's transpose using

`var ppt = P.plus(PT);`

and the computation hadn't finished after 45 minutes so I killed it.

I wrote my own matrix addition method to work around this as follows

````
proc matPlus(A:[],B:[]) {
  var dom = {A.domain.dim(1),B.domain.dim(2)};
  var sps = CSRDomain(dom);
  sps += A.domain;
  sps += B.domain;
  var S: [sps] real;
  for (i,j) in sps {
    S(i,j) = A(i,j) + B(i,j);
  }
  return S;
}
````

and with that instead, the same line ran in 26 seconds.

**Comment by ben-albrecht** [comment]
I believe this is related to #9344.

**Comment by ben-albrecht** [comment]
Here's a standalone reproducer of this perf difference:

```chpl
// Sparse case

use Time;
use LinearAlgebra.Sparse;

var t: Timer;

var ADom = CSRDomain(2553515);
var BDom = CSRDomain(2553515);

var Aindices = for i in 1..1000000 do (i,i);
var Bindices = for i in 2..2000000 by 2 do (i,i);

ADom += Aindices;
BDom += Bindices;

var A: [ADom] real = 1.0;
var B: [BDom] real = 2.0;

writeln('Inputs created');

t.start();
{
  var CDom = CSRDomain(ADom.parentDom);
  CDom += ADom;
  CDom += BDom;
  var C: [CDom] A.eltType  = [ij in CDom] A[ij] + B[ij];
}
t.stop();
writeln('forall ij in CDom');
writeln(t.elapsed());

t.clear();
t.start();
{
  var C = A.plus(B);
}
t.stop();
writeln('A.plus(B)');
writeln(t.elapsed());
t.clear();
```

ben-albrecht:
This was caused by slow sparse domain assignment, which was sped up by ~900x for CSR domains in https://github.com/chapel-lang/chapel/pull/14112.

@Tshimanga - Thanks again for making us aware of this, and thanks for addressing it @e-kayrakli!

Issue #9344, opened 2018-04-25
Slow Assignment of Sparse Arrays


Tshimanga:
I need to build matrix polynomials out of sparse square matrices that have 10^6 rows/cols and 10^6 or 10^7 nonzeros. I need to be able to take loaded `NamedMatrix` objects and build the polynomials out of sums of the powers of it's underlying sparse array and its transpose. I can load the matrix, but the step of assigning it's underlying matrix to a new variable

```
var t: Timer;
t.start();
var P = namedMatrix.X;
t.stop();
writeln("Time to Handle Underlying Sparse Matrix: ",t.elapsed());
```

takes a prohibitively long time. I don't have enough time to let it run to completion.

bradcray:
@Tshimanga : Would you be willing to put together a standalone test case that illustrates the issue for study / improvement?

Tshimanga:
Okay, I'll try to get to that soon. Actually I've been wondering if it would be useful to have a `chapel-lang/reproducers` repo to store all of these open problems @bradcray

Tshimanga:
If I write the standalone test and submit a PR do I get granted admittance to the list of Chapel Contributors? :P

bradcray:
If you put the test on this issue, I'm happy to do the test harness and PR work for you (though obviously, I'd be happy if you were to do that yourself as well).  We typically only add people to the contributors list when they do PRs that show up in the release tarball and for the most part the test/ directory does not go into the release (where test/release/ is the exception).

ben-albrecht:
If the issue is with sparse array assignment, I'd expect something like this to be a work-around:

```chpl
var PDom: namedMatrix.X.domain.type;
PDom += namedMatrix.X.domain;
var P: namedMatrix.X.type;
P = A;
```

ben-albrecht:
> We typically only add people to the contributors list when they do PRs that show up in the release tarball and for the most part the test/ directory does not go into the release

However, he *would* get added to the [Chapel Contributors Team](https://github.com/orgs/chapel-lang/teams/chapel-contributors) when he turned in his CLA, which would be required to submit the test.

bradcray:
[...]

PR #9366, opened 2018-04-26
Improved performance of LinearAlgebra.Sparse _array.plus()

Tshimanga:
The matrix methods  `_array.plus()` and `_array.minus()` don't seem to scale well on large matrices. This implementation does. This PR is a work around for #9364 but not a solution, as I believe that issue points to an important underlying language issue related to recent ticket #9344

### DIFF
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -396,11 +396,17 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
   return transpose(this);
 }
 
-/* Add matrices, maintaining dimensions, deprecated for ``_array.plus`` */
+/*Add matrices, maintaining dimensions. Preferred method at scale */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(B) {
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
-  return A.plus(B);
+  var dom = {A.domain.dim(1),B.domain.dim(2)};
+  var sps = CSRDomain(dom);
+  sps += A.domain;
+  sps += B.domain;
+  var S: [sps] real;
+  for (i,j) in sps {
+    S(i,j) = A(i,j) + B(i,j);
+  }
+  return S;
 }
 
 /* Add matrices, maintaining dimensions */
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -398,23 +398,22 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
 
 /*Add matrices, maintaining dimensions. Preferred method at scale */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(B) {
-  var dom = {A.domain.dim(1),B.domain.dim(2)};
-  var sps = CSRDomain(dom);
-  sps += A.domain;
-  sps += B.domain;
-  var S: [sps] real;
-  for (i,j) in sps {
-    S(i,j) = A(i,j) + B(i,j);
-  }
-  return S;
+  compilerWarning('matMinus has been deprecated for _array.plus, ' +
+                  'try: A.minus(B)');
+  return A.plus(B);
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
-  if Adom.rank != this.domain.rank then compilerError("Unmatched ranks");
-  if Adom.shape != this.domain.shape then halt("Unmatched shapes");
-  var C: [Adom] eltType = this + A;
-  return C;
+proc _array.plus(A: [?Adom]) where isCSArr(A) && isCSArr(this) {
+  if this.domain.parentDom != Adom.parentDom then halt("Unmatched Shapes");
+  var sps = CSRDomain(Adom.parentDom);
+  sps += this.domain;
+  sps += Adom;
+  var S: [sps] real;
+  forall (i,j) in sps {
+    S(i,j) = this(i,j) + A(i,j);
+  }
+  return S;
 }
 
 /* Subtract matrices, maintaining dimensions, deprecated for ``_array.minus``*/
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -396,7 +396,7 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
   return transpose(this);
 }
 
-/*Add matrices, maintaining dimensions. Preferred method at scale */
+/* Add matrices, maintaining dimensions, deprecated for ``_array.plus`` */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(B) {
   compilerWarning('matMinus has been deprecated for _array.plus, ' +
                   'try: A.minus(B)');
@@ -404,14 +404,15 @@ proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangular
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isCSArr(A) && isCSArr(this) {
-  if this.domain.parentDom != Adom.parentDom then halt("Unmatched Shapes");
+proc _array.plus(A: [?Adom] eltType) where isCSArr(A) && isCSArr(this) {
+  if Adom.rank != this.domain.rank then compilerError("Unmatched ranks");
+  if this.domain.shape != Adom.shape then halt("Unmatched shapes");
   var sps = CSRDomain(Adom.parentDom);
   sps += this.domain;
   sps += Adom;
-  var S: [sps] real;
+  var S: [sps] eltType;
   forall (i,j) in sps {
-    S(i,j) = this(i,j) + A(i,j);
+    S[i,j] = this[i,j] + A[i,j];
   }
   return S;
 }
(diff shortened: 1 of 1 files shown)


ben-albrecht:
Let's leave `matPlus` alone (intended for deprecation), and make this change to `_array.plus() .. where isCSArr(A) && isCSArr(B)`

Tshimanga:
I changed `matPlus()` rather than `_array.plus()` because `matPlus` was just calling `_array.plus()` and didn't actualy exist on its own; and, so the problematic `_array.plus()` would still be around for anyone wanting to study it.

ben-albrecht:
Note we'll also need a [CLA](https://github.com/chapel-lang/chapel/tree/master/doc/rst/developer/contributorAgreements) to merge.

ben-albrecht:
> the problematic _array.plus() would still be around for anyone wanting to study it.

I'd prefer users get the non-problematic case by default. We can capture the problematic case in a performance test and study that separately :)

Tshimanga:
sent the CLA to chapel_admin@cray.com but I got an auto response say "Recipient Inbox Full"

Tshimanga:
@ben-albrecht the CLA has been sent in 👍

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,11 +396,17 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
 }
 
-/* Add matrices, maintaining dimensions, deprecated for ``_array.plus`` */
+/*Add matrices, maintaining dimensions. Preferred method at scale */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(B) {
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
-  return A.plus(B);
+  var dom = {A.domain.dim(1),B.domain.dim(2)};
+  var sps = CSRDomain(dom);
```
ben-albrecht: We should check that `Adom.parentDom == Bdom.parentDom`, emitting an error message if false, and then just pass one of the parent domains here, e.g.

```chpl
var sps = CSRDomain(Adom.parentDom);
```

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,11 +396,17 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
+  sps += A.domain;
```
ben-albrecht: We can use `Adom` and `Bdom` directly

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,11 +396,17 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
+  sps += B.domain;
+  var S: [sps] real;
+  for (i,j) in sps {
```
ben-albrecht: Let's utilize parallelism and do a `forall`

Tshimanga:
Okay, @ben-albrecht made the changes and retested. everything checked out on my side of things

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,19 +396,24 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
-  if Adom.rank != this.domain.rank then compilerError("Unmatched ranks");
-  if Adom.shape != this.domain.shape then halt("Unmatched shapes");
-  var C: [Adom] eltType = this + A;
-  return C;
+proc _array.plus(A: [?Adom]) where isCSArr(A) && isCSArr(this) {
+  if this.domain.parentDom != Adom.parentDom then halt("Unmatched Shapes");
```
ben-albrecht: We can actually just check the shape (I may have misled you earlier), which would be more consistent with other functions in the library:

```chpl
  if this.domain.shape != Adom.shape then halt("Unmatched shapes");
```

Also note the lower case `shapes` for error message consistency

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,19 +396,24 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
+  var sps = CSRDomain(Adom.parentDom);
+  sps += this.domain;
+  sps += Adom;
+  var S: [sps] real;
```
ben-albrecht: We'd like this to be generic w.r.t. element type, so we should refer to the array `eltType` here. You can do this cleanly by adding an element type query to the signature, and use that type variable here, e.g.

```chpl
proc _array.plus(A: [?Adom] ?eltType) where isCSArr(A) && isCSArr(this) {

...

var S: [sps] eltType;
```

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,19 +396,24 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
   return transpose(this);
 }
 
-/* Add matrices, maintaining dimensions, deprecated for ``_array.plus`` */
+/*Add matrices, maintaining dimensions. Preferred method at scale */
```
ben-albrecht: I think you unintentionally left some changes to `matPlus`

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,19 +396,24 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
+  compilerWarning('matMinus has been deprecated for _array.plus, ' +
+                  'try: A.minus(B)');
   return A.plus(B);
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
-  if Adom.rank != this.domain.rank then compilerError("Unmatched ranks");
```
ben-albrecht: We should continue to check the rank - since it will emit a compile-time error, whereas we can only check shape at runtime.

modules/packages/LinearAlgebra.chpl
```diff
@@ -396,19 +396,24 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
-  var C: [Adom] eltType = this + A;
-  return C;
+proc _array.plus(A: [?Adom]) where isCSArr(A) && isCSArr(this) {
+  if this.domain.parentDom != Adom.parentDom then halt("Unmatched Shapes");
+  var sps = CSRDomain(Adom.parentDom);
+  sps += this.domain;
+  sps += Adom;
+  var S: [sps] real;
+  forall (i,j) in sps {
+    S(i,j) = this(i,j) + A(i,j);
```
ben-albrecht: For style-consistency, let's use brackets for array indexing, e.g. `S[i,j] = ...`

modules/packages/LinearAlgebra.chpl
Tshimanga: ah good point

Tshimanga:
@ben-albrecht i made the changes. also, I made the analogous modifications to `_array.minus()` as well

modules/packages/LinearAlgebra.chpl
```diff
@@ -398,17 +398,23 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
 
 /* Add matrices, maintaining dimensions, deprecated for ``_array.plus`` */
 proc matPlus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(B) {
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
+  compilerWarning('matMinus has been deprecated for _array.plus, ' +
```
ben-albrecht: Still an unintentional change here.

modules/packages/LinearAlgebra.chpl
```diff
@@ -398,17 +398,23 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
+                  'try: A.minus(B)');
   return A.plus(B);
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
```
ben-albrecht: You modified the dense case - you should modify the sparse case defined in `LinearAlgebra.Sparse` instead

modules/packages/LinearAlgebra.chpl
```diff
@@ -419,11 +425,17 @@ proc matMinus(A: [?Adom] ?eltType, B: [?Bdom] eltType) where isDefaultRectangula
 }
 
 /* Subtract matrices, maintaining dimensions */
-proc _array.minus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
```
ben-albrecht: You modified the dense case - you should modify the sparse case defined in `LinearAlgebra.Sparse` instead

modules/packages/LinearAlgebra.chpl
```diff
@@ -398,17 +398,23 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
+  compilerWarning('matMinus has been deprecated for _array.plus, ' +
+                  'try: A.minus(B)');
   return A.plus(B);
 }
 
 /* Add matrices, maintaining dimensions */
-proc _array.plus(A: [?Adom]) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
+proc _array.plus(A: [?Adom] eltType) where isCSArr(A) && isCSArr(this) {
```
ben-albrecht: `eltType` needs to be a type *query*, i.e. `A: [?Adom] ?eltType`

modules/packages/LinearAlgebra.chpl
Tshimanga: this is still flagged because it was previously missing a type query `?eltType` on `A` i think

Tshimanga:
Sounds good! Hopefully the underlying issue that this works around ends up being fixable soon.

Issue #8843, opened 2018-03-16
Should Owned allow compiler-added copies to transfer ownership?


mppf:
With the current record-based Owned implementation, the compiler will create an error at compile time rather than transfering ownership due to an auto-copy. Consider for example:

``` chapel
class MyClass { var x:int; }
proc f( ref arg:Owned(MyClass) ) {
  return arg;
  // compiler adds copy initialization to this return
  // since it's returning a reference by value
  // is it surprising that arg stores nil after the return?
}

proc main() {
  var x = new Owned(new MyClass(1));
  f(x);
  writeln(x); // is it surprising that x now stores nil?
}
```

Should such compiler-added ownership transfers be enabled?

mppf:
I think so, once we have checking for empty Owned things, but I'm not totally certain.

vasslitvinov:
From the user's view, the function returns an owned thing. That ownership has to come from somewhere right? In this case it happens to come from the formal. So the ownership being taken from 'x' is what f() is doing. Sure it may be surprising for the caller of f(), yet this is what f() is written to do.

So I think the program should compile and writeln a nil.

mppf:
I'm inclined to agree that such compiler-added copies should be allowed, especially since the current alternative makes a common patterns into errors (e.g. #9259)

bradcray:
I think I tend to agree with @mppf and @vasslitvinov here.

psahabu:
I agree with the decision that was made here.

PR #9367, opened 2018-04-26
Enable implicit copies for Owned

### LINKED ISSUE #9259: Can't create tuple of Owned
### Summary of Problem
`error: implicit copy-initialization invoked for a type that does not allow it`

### Steps to Reproduce

**Source Code:**

```chapel
use OwnedObject;

class Foo {
  var x: int;

  proc init(x: int) {
    this.x = x;
  }
}

var a = new Owned(new Foo(1));
var b = new Owned(new Foo(2));

var c = (a, b);
```

**Compile command:**
`chpl --force-initializers app.chpl`

### Configuration Information

`chpl version 1.18.0 pre-release (593b3d1e25)`

### LINKED ISSUE #9280: delete-free subtask: enable compiler-added copies for owned
As a Chapel user, I need to be able to use`owned` class values in more code and I am surprised by errors that the compiler emits when an `owned` cannot be copied. I need to be able to use `owned` class values the way I can use a record.

See issue #8843 for the design discussion.

Acceptance criteria:
The example from issue #8843 compiles.
To the extent possible, the compiler performs lvalue checking on owned.

mppf:
* splits lateConstChecking out of cullOverReferences
* invokes lateConstChecking in callDestructors, to check autoCopies added there
* adds error checking for build_tuple calls with Owned

See issue #8843 for the language design discussion.

Fixes #9259.
Closes #9280.

- [x] pass full local testing


modules/internal/OwnedObject.chpl
```diff
@@ -231,7 +231,6 @@ module OwnedObject {
   pragma "no doc"
   pragma "donor fn"
   pragma "auto copy fn"
-  pragma "erroneous autocopy"
   pragma "unsafe"
```
daviditen: It looks like this was the only thing using "erroneous autocopy", can the flag be removed from the compiler?

test/classes/ferguson/delete-free/owned-shared-fields2.bad
```diff
@@ -1,4 +1,8 @@
-owned-shared-fields2.chpl:9: error: implicit copy-initialization invoked for a type that does not allow it
-owned-shared-fields2.chpl:18: error: implicit copy-initialization invoked for a type that does not allow it
-owned-shared-fields2.chpl:13: error: implicit copy-initialization invoked for a type that does not allow it
-owned-shared-fields2.chpl:22: error: implicit copy-initialization invoked for a type that does not allow it
+owned-shared-fields2.chpl:11: internal error: CAL0078 chpl version 1.18.0 pre-release (573fcb6282)
```
daviditen: It's unfortunate that this went from a user error to an internal error.  Since it's just a .bad file I think we can let it slide.

test/classes/ferguson/delete-free/owned-shared-fields2.chpl
```diff
@@ -1,23 +1,30 @@
 use OwnedObject;
 use SharedObject;
 
+pragma "use default init"
```
daviditen: I haven't gone looking to figure out what this pragma is for, but can you explain why these classes and records need this pragma?

compiler/include/symbol.h
```diff
@@ -22,6 +22,7 @@
 
 #include "baseAST.h"
 
+#include "astutil.h"
```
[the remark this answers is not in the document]
mppf: It's not unnecessary - symbol.h has inline functions that use macros defined in astutil.h, so the "where to include" rules I think we've been following mean it should be here.

test/classes/ferguson/delete-free/owned-shared-fields2.chpl
mppf: The pragma enables the default initializer (vs the old default constructor). The old default constructor is not working for owned fields, but we're trying to move away from it presently, so that didn't seem worth digging in to. So I just marked it as "use default init" to require the new default initializer code is used.

modules/internal/OwnedObject.chpl
mppf: The compiler uses it to generate errors for the case of no copy initializer defined for a type.

daviditen: Hmm.  I could have sworn that I searched the compiler for this flag and saw no other uses, but trying again this morning I see them.  Thanks!

PR #9397, opened 2018-04-30
Add a test of a 'type' argument with a constraint and a query expression

bradcray:
While answering a user question last week, I was excited to find that
the following case worked as I'd hoped: putting a generic type
constraint on a `type` argument supported the ability to do queries of
the generic elements of the type, as in value argument contexts:

```chapel
proc foo(type t: C(?t2)) ...
```

Doing some grepping I wasn't easily finding instances of other tests
that tried to exercise this pattern so thought I'd add this to lock
the behavior in or at least put epsilon more weight on it.


bradcray:
Credit to @mppf who seems most likely to have enabled this capability in his recent-ish work.

PR #9400, opened 2018-04-30
Use different code for makeRawNew in resolve

mppf:
The makeRawNew function was called both from build and
from implementForallIntents during resolution. The problem
with that is that the AST for new is slightly different before
and after normalize. So, solve the problem by moving the function
to build.cpp and getting it to generate the pre-normalize pattern
and then using the body the function in implementForallIntents with
the post-normalize pattern.

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -912,6 +912,8 @@ static void processSyntacticDistributions(CallExpr* call) {
  */
 static void processManagedNew(CallExpr* newCall) {
   SET_LINENO(newCall);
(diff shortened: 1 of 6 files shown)


compiler/passes/normalize.cpp
```diff
@@ -912,6 +912,8 @@ static void processSyntacticDistributions(CallExpr* call) {
  */
 static void processManagedNew(CallExpr* newCall) {
   SET_LINENO(newCall);
+  /*if (newCall->id == 190325 || newCall->id == 190284)
```
benharsh: Looks like a leftover from debugging.

Issue #9411, opened 2018-05-01
c_fn_ptr is one type but it should be many


mppf:
When creating function pointers from Chapel code for interoperability, the Chapel compiler converts all of these function pointers to void*. This can cause problems with certain compilers, but one can verify this situation by simply inspecting the generated code.

``` chapel
// ex.chpl
extern proc acceptsFnPtr(arg: c_fn_ptr);

proc f(arg: c_int): c_int {
  return arg;
}

proc test() {
  acceptsFnPtr(c_ptrTo(f));
}

test();
```

chpl ex.chpl  --savec=tmp
(and then ignore the failure to C compile).

See tmp/ex.c:
``` c
static void test_chpl(void) {
  acceptsFnPtr(((c_fn_ptr)(f_chpl)));
  return;
}
```

Some C compilers will warn about this (c_fn_ptr) cast. If denormalize is off, or if the C function pointer is saved in a variable, the c_fn_ptr will be a variable type rather than just a cast.

lydia-duncan:
From #9768:

> I was trying to send a function pointer to an extern C function from Chapel, but because of naivety in our c_fn_ptr and FCF implementation, I ended up needing to make an extern type to represent the type of the function pointer and defining explicit instances of that type in C that I could pass around in Chapel. This seemed unnecessarily painful
> From Michael: "c_ptrTo(c_free) or c_fn_ptr are jamming all function ptrs into 1 type which can cause problems in C compilation ...
"it would be pretty nice to fix the problems with c_fn_ptr ... mainly c_fn_ptr has to take arguments and the codegen needs to know how to generate it"

lydia-duncan:
Having extern records with function pointer fields causes problems in baseline, due to inserting refs to the field when the expected type of the ref is void** but the type will be a pointer to a function pointer.  This problem was present both when the field was listed as c_void_ptr or c_fn_ptr in Chapel

PR #9412, opened 2018-05-01
Address problems with GMP after moving out of runtime

mppf:
This PR resolves testing failures after PR #8944.

### Addressing problems with void* vs C function pointer types

Some C compilers in nightly testing caused failures because they cared
that void* wasn't the matching argument type for mp_set_memory_functions.
To overcome that, I've wrapped mp_set_memory_functions with a C function
that simply casts the c_fn_ptr (void*) to the appropriate function
pointer type.

Issue #9411 describes why this workaround is necessary and what the
real solution would be.

### Addressing problems with pointer type mismatch for mp_limb_t*

Because GMP can be configured to create a different mp_limb_t
on different systems (and in fact it does so on Cygwin), the
GMP module should not assume anything about the size of mp_limb_t.
(Previous to this PR it assumed it matched `uint(64)`).
This PR makes that type opaque. Some related code needed
to be adjusted based on that change:
    
* chpl_gmp_get_mpz now uses chpl_comm_get and sizeof
   instead of chpl_comm_array_get
* added chpl_gmp_mpz_nlimbs and chpl_gmp_mpz_getlimbn. These are
   callable from user code & used in the BigInteger module to
   get the number of limbs / the i'th limb. These both return uint(64)
   always, so that the issue with mp_size_t varying is less problematic.
 * BigInteger functions numLimbs and get_limbn now use the uint(64)
   versions above.
 * mpz_getlimbn is now private to GMP.chpl since it can't be
   used right now without internal compiler knowledge, since the
   Chapel compiler won't understand its return type.
    
While adjusting bigint's numLimbs and get_limbn I noticed that these do not
appear to be tested, so I added a (multilocale even!) test that
they work correctly.

### Testing & Review

- [x] verified that it resolves the problem with pidigits and PrgEnv-Cray on a Cray XC
- [x] cygwin GMP failures resolved
- [x] test/library/standard/GMP passes with GASNet
- [x] full local testing

### DIFF
--- a/modules/standard/BigInteger.chpl
+++ b/modules/standard/BigInteger.chpl
@@ -246,30 +246,27 @@ module BigInteger {
     }
 
     proc numLimbs : uint {
-      var mpz_struct = this.mpz[1];
-
-      return chpl_gmp_mpz_nlimbs(mpz_struct).safeCast(uint);
+      return chpl_gmp_mpz_nlimbs(this.mpz);
     }
 
-    proc get_limbn(n: uint) : uint {
-      const n_ = n.safeCast(mp_size_t);
-      var   ret: mp_limb_t;
+    proc get_limbn(n: integral) : uint {
+      var   ret: uint;
 
       if _local {
-        ret = mpz_getlimbn(this.mpz, n_);
+        ret = chpl_gmp_mpz_getlimbn(this.mpz, n);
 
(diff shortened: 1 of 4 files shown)


mppf:
@ronawho - could you review this or delegate that task to somebody else? Thanks.

PR #9355, opened 2018-04-26
Enable variables declared with generic type

mppf:
This PR enables variables declared with generic type, for example
``` chapel
var x: integral = 1;
var y: unmanaged = new unmanaged MyClass();
```

To do so, it updates the normalization of typed variable initialization. It needed to do so because normalize was using a default-init then assign pattern for such variable initialization, but that doesn't work right because you can't default-init with a generic type. So, now the normalizer represents typed variable initialization with a new third argument to PRIM_VAR_INIT that contains the type. During resolution of PRIM_VAR_INIT, if the third argument is present and represents a possibly different type, it is converted to a PRIM_MOVE + PRIM_COERCE. PRIM_COERCE already existed in resolution so that the normalizer could handle coercions for returns in functions with declared type. This change got PRIM_COERCE to be used in more places, so this PR also makes several important improvements to it.

Helps with #9325 (but a parser change will be required to go the rest of the way).
Fixes one of the cases in #8467 (but not the other because of other iterator -> array conversion issues).
Fixes #8392.

- [x] full local testing d2bf68d6bdfd83b94ffd10308a43a8ba8766d1f9

PR #9422, opened 2018-05-01
Fix typed sync variable initialization

mppf:
This is a follow-on to PR #9355.

It fixes a failure in fifo configurations for the test
```
 parallel/sync/diten/returnSync
```

Since sync/single have an initCopy that returns a different type,
they need to generate PRIM_COERCE rather than initCopy for the
case of typed variable initialization like

``` chapel
 var x: sync int = returnsSync();
```

- [x] full local testing
- [x] full local quickstart testing

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -5151,9 +5151,12 @@ static void resolveInitVar(CallExpr* call) {
       targetType = useType;
     }
 
-    // Ignore the target type if it's the same & not a runtime type.
+    // Ignore the target type if it's the same & not a runtime type
+    // and not sync/single (since sync/single initCopy returns a different type)
     if (targetType == srcType &&
-        !targetType->symbol->hasFlag(FLAG_HAS_RUNTIME_TYPE)) {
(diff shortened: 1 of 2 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -5151,9 +5151,12 @@ static void resolveInitVar(CallExpr* call) {
...
     }
 
-    // Ignore the target type if it's the same & not a runtime type.
+    // Ignore the target type if it's the same & not a runtime type
+    // and not sync/single (since sync/single initCopy returns a different type)
     if (targetType == srcType &&
-        !targetType->symbol->hasFlag(FLAG_HAS_RUNTIME_TYPE)) {
+        !targetType->symbol->hasFlag(FLAG_HAS_RUNTIME_TYPE) &&
+        !isSyncType(srcType) &&
+        !isSingleType(srcType)) {
```
lydia-duncan: Man, I coulda sworn we'd already done that.  Did it get accidentally removed in the other PR?

mppf: We did it in the code handling PRIM_COERCE and also lower down here but not in this specific spot.

lydia-duncan: Ah, okay

PR #9428, opened 2018-05-01
Fix compilation error in knl LocaleModel

mppf:
This is a follow-on to PR #9355.

Fixes compilation error for hello.chpl with CHPL_LOCALE_MODEL=knl.


dmk42:
This looks OK to my eye and it passes release/examples on KNL, but it fails localeModels/knl/on_mem.chpl.  I'm investigating.

PR #9433, opened 2018-05-01
Dataframes -> DataFrames

psahabu:
Brings DataFrames in line with Spark and Pandas casing.


psahabu:
Thanks for the reminder, all the tests in the `DataFrames` directory should be good now.

Issue #8867, opened 2018-03-19
chpldoc 'throws'


psahabu:
## user story
As a Chapel programmer, I want a library's documentation to provide a list of `Error`s that can be thrown from each function because I want to be able to handle all the potential `Error`s.

### acceptance criteria
- `:throws` is a valid keyword on methods in `chpldoc`
- All methods marked `throws` in the standard and package modules are documented with `:throws`.

lydia-duncan:
@bradcray - does the sorting on this issue seem right to you?  I'm thinking about my next sprint and realized that this one might have been placed in Backlog before we had strong guidelines about whether that was appropriate or not.

bradcray:
I think the backlog is appropriate for this one.

lydia-duncan:
(In case it wasn't clear, we do output throws on functions in chpldoc today, but we don't have a category in the documentation handling for lists of what errors are thrown.  I've been interpreting this issue as referring to the later concept, since the other support was added half a year before this issue was opened)

bradcray:
My assumption is that if a user reads a library routine's documentation and can't determine what errors s/he might be responsible for catching at the callsite, just knowing that it throws isn't terribly helpful (?).

lydia-duncan:
Yeah, that's what we were thinking as well.  Just making sure that didn't change the answer.  I think I'll add this to my next sprint cycle

bradcray:
I think there's a design question here as to whether the author of the documentation must manually list all the errors it can throw (as they must for the current "returns" section of the documentation) or whether that list would be auto-generated by chpldoc.  In either case, the task seems complicated in cases where the routine is passing along an error from a throwing routine that it called.

lydia-duncan:
My expectation is that it would be manual, but we can (and should) bring it up to the larger group for discussion.  

My rationale for this expectation is that no other part of chpldoc requires resolution to run - while it would be handy, it opens a can of worms implementation-wise and it seems like something the author should know ahead of time instead of relying on chpldoc to tell them.

Relatedly, we have expected that the user could write documentation prior to the implementation of the function itself - return statements don't have to be present in the body for the function to be documented with a return type, but such functions would encounter an issue should they get resolved.  As an example, the following file is perfectly documentable prior to the implementation of foo's body:
```chapel
/* Some documentation */
proc foo (): int { }
```
I think we should treat `throws` consistently with the return type in this regard.

[...]

PR #9450, opened 2018-05-03
deprecate 'out error' pattern from modules

psahabu:
## Motivation
The `out error: syserr` pattern was initially created for the `IO` module, and became the de facto error handling approach within Chapel modules. Now that a fleshed out approach is in place, the standard modules ought to use it. Additionally, the standard modules provide a solid proving ground for the error handling feature.

## Approach
- Move functionality from the `out error: syserr` function to the `throws` function.
- Rewrite the `out error: syserr` function as a wrapper for the `throws` function.
- Add a deprecation warning to the `out error: syserr` function.

### Next Steps
- Have deprecation warnings in place for 1.18.
- Remove `out error: syserr` functions altogether in 1.19.

## Suggested review order
- `modules/`
  - [x] `standard/IO.chpl`
  - [x] `standard/FileSystem.chpl`
  - [x] `standard/Path.chpl`
  - [x] `standard/Spawn.chpl`, `standard/Regexp.chpl`
  - [x] `packages/Buffer.chpl`, `packages/HDFS.chpl`
  - [x] `internal/CPtr.chpl`, `internal/ChapelIO.chpl`, `internal/DefaultRectangular.chpl`
- [x] test/*


modules/standard/IO.chpl
```diff
@@ -4161,66 +4225,73 @@ proc channel.writeln(const args ...?k,
...
-   :arg error: an optional string argument which will be printed
-               out if the assert fails. The default prints "Not at EOF".
+/* Returns true if a channel has reached end-of-file, false if not.
+   Throws an error if this is a writing channel, or if there was
+   an error doing the read.
  */
-proc channel.assertEOF(error:string = "- Not at EOF") throws {
+proc channel.atEOF(): bool throws {
   if writing {
     try this._ch_ioerror(EINVAL, "assertEOF on writing channel");
```
mppf: atEOF on writing channel?

modules/standard/IO.chpl
```diff
@@ -4161,66 +4225,73 @@ proc channel.writeln(const args ...?k,
...
-   at EOF.
-
-   :arg error: an optional string argument which will be printed
-               out if the assert fails. The default prints "Not at EOF".
+/* Returns true if a channel has reached end-of-file, false if not.
+   Throws an error if this is a writing channel, or if there was
+   an error doing the read.
  */
-proc channel.assertEOF(error:string = "- Not at EOF") throws {
+proc channel.atEOF(): bool throws {
```
mppf: It was intentional that we didn't have an `atEOF` function because such a function is inherently racy for channels that can get more data (e.g. a socket). Can this be a no-doc'd function that we just use to implement assertEOF?

modules/standard/IO.chpl
```diff
@@ -6761,119 +6876,120 @@ proc readf(fmt:string):bool throws {
...
+  } catch {
+    error = EINVAL;
+  }
+}
+
+
+// This function is no longer available.
+pragma "no doc"
+proc format(fmt:string, args ...?k):string {
+  compilerError("use string.format(args ...) not format(fmt, args ...)");
```
mppf: I think this can be removed - it was deprecated some time ago.

modules/standard/IO.chpl
```diff
@@ -6761,119 +6876,120 @@ proc readf(fmt:string):bool throws {
...
   defer {
-    var closeError:syserr;
-    f.close(error=closeError);
-    // Propagate an error on closing only if there wasn't an original error.
-    if !error then
-      error = closeError;
+    try {
+      f.close();
+    } catch {
+      // ignore close errors
```
mppf: I'd prefer it if this code still reported an error on close if one occurred.
It could do that by calling close explicitly outside of the defer block, and then redundantly calling it in the defer block. Or replacing the defer block with a try / catch where the error handling case closes the files and re-throws.

modules/standard/IO.chpl
```diff
@@ -1369,27 +1370,29 @@ proc file._style:iostyle throws {
    Files are automatically closed when the file variable
    goes out of scope and all channels using that file are closed. Programs
    may also explicitly close a file using this method.
-
-   :arg error: optional argument to capture an error code. If this argument
-               is not provided and an error is encountered, this function
-               will halt with an error message.
```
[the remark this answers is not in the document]
psahabu: It is always a `SystemError`, but we don't yet have a `throws` equivalent for `arg` yet. That's part of #8867.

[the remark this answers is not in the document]
mppf: Can you just make the docs say it in the description?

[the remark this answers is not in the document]
mppf: i.e. the information is what matters, less so the formatting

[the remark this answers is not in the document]
psahabu: Ah, gotcha. Will do.

modules/standard/IO.chpl
psahabu: Should this function be allowed to return a valid output when any of the `close()`s throw an error? The function seems to have allowed that before, so if we need to maintain that, the patterns you describe won't work.

mppf: I think it shouldn't return output if any of the closes throw an error.

psahabu: Alright, I've updated this function to throw on an close error, supposing another error did not occur.

modules/standard/FileSystem.chpl
```diff
@@ -302,79 +283,136 @@ proc copy(out error: syserr, src: string, dest: string, metadata: bool = false)
...
-  if error {
-    var ignore:syserr;
-    srcFile.close(error=ignore);
-    return;
+  var srcFile = try open(src, iomode.r);
+  defer {
+    try {
+      srcFile.close();
+    } catch {
+      // ignore errors
```
mppf: As with IO functions, we shouldn't be ignoring an error on a close. We can ignore an error on close if it happens in the process of returning an error. But if everything else ran OK, an error on close should cause the function to throw. This pattern occurs several times in this function.

modules/standard/FileSystem.chpl
```diff
@@ -574,19 +571,34 @@ proc locale.cwd(out error: syserr): string {
...
+  var ret:string;
   var err: syserr = ENOERR;
-  var ret = cwd(err);
+  on this {
+    var tmp:c_string;
+    // c_strings can't cross on statements.
+    err = chpl_fs_cwd(tmp);
+    ret = if err
+          then ""
+          else new string(tmp, needToCopy=false);
```
mppf: Can this just be `ret = tmp:string`? I don't like the call to the string constructor here.

modules/standard/IO.chpl
```diff
@@ -2287,9 +2309,6 @@ proc channel.advancePastByte(byte:uint(8)) throws {
     using this feature, for example by *marking* at offset=0 and then
     advancing to the end of the file. It is important to be aware of these
     memory space requirements.
-
-  :returns: an error code, if an error was encountered.
-
```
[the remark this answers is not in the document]
mppf: Actually let's leave the comment and adjust mark / _mark to throws in a separate PR.

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,81 @@ module Spawn {
...
       // Close stdin.
       if this.stdin_pipe {
         // send data to stdin
         _stop_stdin_buffering();
-        this.stdin_channel.close(error=error);
+        try {
+          this.stdin_channel.close();
+        } catch e: SystemError {
+          err = e.err;
+        } catch { }
```
mppf: err = INVAL?

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,81 @@ module Spawn {
...
-      // If the channels are to stay open, use buffer=true or communicate.
+      // If these channels are to stay open, use buffer=true or communicate.
+      // Close stdout channel.
       if this.stdout_pipe {
-        this.stdout_channel.close(error=error);
+        try {
+          this.stdout_channel.close();
+        } catch e: SystemError {
+          err = e.err;
+        } catch { }
```
mppf: ditto

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,81 @@ module Spawn {
...
       }
+
+      // Close stderr channel.
       if this.stderr_pipe {
-        this.stderr_channel.close(error=error);
+        try {
+          this.stderr_channel.close();
+        } catch e: SystemError {
+          err = e.err;
+        } catch { }
```
mppf: ditto

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,81 @@ module Spawn {
...
       }
 
       // wait for child process to terminate
       var done:c_int = 0;
       var exitcode:c_int = 0;
-      error = qio_waitpid(pid, 1, done, exitcode);
+      err = qio_waitpid(pid, 1, done, exitcode);
```
mppf: This function had this problem before you got to it, but we should be using different error variables for the different operations, initializing them each with 0, and then combining them at the end, so that if any one operation fails we get an error out.

modules/standard/Spawn.chpl
```diff
@@ -779,82 +810,99 @@ module Spawn {
...
+      } catch { }
     }
     // Close stdout.
     if this.stdout_pipe {
-      this.stdout_channel.close(error=error);
+      try {
+        this.stdout_channel.close();
+      } catch e: SystemError {
+        err = e.err;
+      } catch { }
```
mppf: Same issues as the other one, don't want empty catchall, should be using different error variables for the different calls and combining them at the end.

mppf:
Done looking at  standard/Path.chpl standard/Spawn.chpl, standard/Regexp.chpl

modules/packages/Buffers.chpl
```diff
@@ -392,39 +395,45 @@ module Buffers {
...
      :arg range: the region of the buffer to copy, for example buffer.all()
-     :arg error: (optional) capture an error that was encountered instead of
-                 halting on error
      :returns: a newly initialized bytes object on the current locale
    */
-  proc buffer.flatten(range:buffer_range, out error:syserr) {
-  //proc buffer.flatten(range:buffer_range, out error:syserr):bytes {
-    var ret:bytes = new bytes();
+  proc buffer.flatten(range:buffer_range) throws {
+  //proc buffer.flatten(range:buffer_range):bytes throws {
```
mppf: Just remove commented out line?

modules/packages/Buffers.chpl
```diff
@@ -556,19 +576,25 @@ module Buffers {
...
-           halting on error
   */
-  proc buffer.prepend(b:bytes, skip_bytes:int(64) = 0, len_bytes:int(64) = b.len, out error:syserr) {
+  proc buffer.prepend(b:bytes, skip_bytes:int(64) = 0, len_bytes:int(64) = b.len) throws {
+    var err:syserr = ENOERR;
     on this.home {
-      qbuffer_prepend(this._buf_internal, b._bytes_internal, skip_bytes, len_bytes);
+      err = qbuffer_prepend(this._buf_internal, b._bytes_internal, skip_bytes, len_bytes);
     }
+    if err then try ioerror(err, "in buffer.append");
```
mppf: append -> prepend

modules/standard/Spawn.chpl
psahabu: I tried having each error case check if `err` was already set, with the `qio_waitpid` error taking precedence. Let me know if this is what you had in mind.

psahabu: Alright I gave it another shot. If it looks good to you I'll fire off testing.

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,99 @@ module Spawn {
...
+    if !running {
+      if this.spawn_error then
+        try ioerror(this.spawn_error, "in subprocess.wait");
+    }
 
-    on home {
+    // TODO: rethink this approach, perhaps with an Error list, or var Error
+    // - complete all these throwing calls, even if they throw
+    //   - extract a syserr if a SystemError is caught
+    // - at the end, throw an error if err is set
```
mppf: I don't see this approach as so bad - it seems fine to me. So I don't really see why the above comment needs to be here.

modules/standard/Spawn.chpl
```diff
@@ -711,60 +721,99 @@ module Spawn {
...
+    }
 
+    if wait_err {
+      try ioerror(wait_err, "in subprocess.wait");
+    } else if stdin_err {
+      try ioerror(stdin_err, "in subprocess.wait");
+    } else if stdout_err {
+      try ioerror(stdout_err, "in subprocess.wait");
+    } else if stderr_err {
+      try ioerror(stderr_err, "in subprocess.wait");
```
mppf: Looks good! I like this much better. Thanks for adjusting it!

PR #9454, opened 2018-05-03
Fix Chapel buffer overflows that trigger gcc 8.1 errors

dmk42:
Some of our users are early adopters of new gcc versions.  Because we use `-Werror`, Chapel will not compile in nightly testing mode (`WARNINGS` set) or with `CHPL_DEVELOPER` set, when gcc 8.1 is the back end C compiler.  This change fixes the buffer overflows that prevent compilation.

There is also a gcc 8.1 complaint about class member access that does not apply.  This change updates the gnu Makefile to turn that off.

Passes full local testing and full GASNet testing.

### DIFF
--- a/compiler/AST/AstDump.cpp
+++ b/compiler/AST/AstDump.cpp
@@ -249,7 +249,7 @@ void AstDump::visitSymExpr(SymExpr* node) {
   if (var != 0 && var->immediate != 0) {
     const size_t bufSize = 128;
     char         imm[bufSize];
-    char         buff[bufSize];
+    char         buff[bufSize + 1];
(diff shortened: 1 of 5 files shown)


dmk42:
@ronawho - Would you mind taking a look at this?  I think the Makefile change is reasonable but you might have an opinion on that.

The new warning that needs to be disabled is C\+\+ only, and the option causes an error when it's on a C command line.  If I add it directly to `WARN_CXXFLAGS`, it gets sucked into `WARN_CFLAGS` as well, so I created a new `WARN_COMMONFLAGS`.

Issue #9414, opened 2018-05-01
Investigate non-kernel scalability issues for ISx


ronawho:
The timed kernel for ISx is scaling well, but we seem to be hitting some bottleneck in a non-timed section. Below is the kernel time for 1 trail (we did 4 trials -- 3 real, 1 burn) and the total srun time. This is with `--mode=weakISO --numTrials 3 --n=$((2**25))`:

| nodes | kernel time | srun time |
| ----- | ----------- | --------- |
| 64    | 2.13        |  23       |
| 128   | 2.20        |  38       |
| 256   | 2.42        |  90       |
| 512   | 2.61        | 200       |

ronawho:
Switching out the `[1..numTrials]` inner array for a tuple like `(1*(real));`gets significant better scalability and I think shows us a "peak" speed that we want to achieve. Using a tuple would force numTrials to be a param though. For some reason the deinit time is actually highest for 32 node runs -- not sure why that is, but it's repeatable.

| nodes | init time | deinit time | aprun time |
| ----- | --------- | ----------- | ---------- |
|  32   |  0.01     |  0.14       |  7.5       |
|  64   |  0.02     |  0.01       |  8.1       |
| 128   |  0.06     |  0.01       |  8.5       |
| 256   |  0.14     |  0.02       |  9.9       |
| 512   |  0.31     |  0.02       | 11.9       |

On the bright side, even if our array destruction time is embarrassing, at least our runtime startup/shutdown code is scalable

ronawho:
Switching out the `[1..numTrials]` inner array with a record wrapper like `record TimerArr { var A: [1..numTrials] real; }` also works. I think that's the version I'm going to go with as a workaround for isx-hand-optimized. Timers are virtually identical to the tuple version:

| nodes | init time | deinit time | aprun time |
| ----- | --------- | ----------- | ---------- |
|  32   |  0.01     |  0.14       |  7.4       |
|  64   |  0.03     |  0.01       |  8.2       |
| 128   |  0.06     |  0.01       |  8.7       |
| 256   |  0.14     |  0.02       |  9.9       |
| 512   |  0.32     |  0.04       | 12.0       |

ronawho:
Here are some ISx timings on master and with the record wrapper. Note that this is on a different machine than the original values in the issue description: 

Master (`[1..numTrials]` inner array):

 nodes | 1 kernel time | all kernel time | aprun time |
| ----- | ------------- | --------------- | ---------- |
[...]

PR #9455, opened 2018-05-03
Improve non-kernel scalability of isx-hand-optimized

ronawho:
The timed kernel of ISx was scaling well, but the total running time was
scaling horrendously. The root cause of this was slow destruction of some
arrays. We have a bunch of block distributed arrays with inner [1..numTrials]
arrays for timer values. During deinit, all remote locales were going to locale
0 to remove themselves from the list of arrays declared over `1..numTrials`.
This led to a ton of on-stmts, each trying to acquire a shared atomic-lock
which killed scalability.

As a hacky workaround this adds a simple record wrapper so that we end up
creating unique `1..numTrials` domains instead of there being a single one.
This significantly improves scalability of ISx's total running time.

| nodes | kernel time | total before | total now |
| ----- | ----------- | ------------ | --------- |
|  64   |  9.6        |  24.6        |  19.6     |
| 128   | 10.4        |  40.1        |  22.1     |
| 256   | 11.6        |  95.5        |  25.5     |
| 512   | 12.8        | 218.1        |  28.1     |

Related to https://github.com/chapel-lang/chapel/issues/9414

### DIFF
--- a/test/studies/isx/isx-hand-optimized.chpl
+++ b/test/studies/isx/isx-hand-optimized.chpl
@@ -132,6 +132,12 @@ config const recvBuffFactor = 2.0,
 config const numBurnInRuns = 1,
              numTrials = 1;
 
+// Horrible hack to help https://github.com/chapel-lang/chapel/issues/9414
+record TimerArr {
+  var A: [1..numTrials] real;
+  proc this(i) ref { return A[i]; }
+  iter these() ref { for a in A do yield a; }
+}
 
 if printConfig then
   printConfiguration();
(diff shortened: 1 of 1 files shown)


test/studies/isx/isx-hand-optimized.chpl
```diff
@@ -132,6 +132,12 @@ config const recvBuffFactor = 2.0,
 config const numBurnInRuns = 1,
              numTrials = 1;
 
+// Horrible hack to help https://github.com/chapel-lang/chapel/issues/9414
+record TimerArr {
+  var A: [1..numTrials] real;
+  proc this(i) ref { return A[i]; }
+  iter these() ref { for a in A do yield a; }
```
ronawho: @mppf I tried to use `forwarding A` here instead of manually forwading this/these, but I got an "already in AST" internal error. I'm having an annoyingly hard time coming up with a smaller reproducer. If you get a second would you mind taking a look and if you don't see anything obvious, just punt it back to me and I'll work more on a smaller reproducer.

ronawho:
@bradcray any objections to this workaround? (note this is isx-hand-optimized, and we'll obviously still want a more principled fix before promoting to the release version)

bradcray:
OK with me.

test/studies/isx/isx-hand-optimized.chpl
mppf: hey @ronawho, I'm trying to fix it in PR #9456.

PR #9456, opened 2018-05-04
Resolve problem with forwarding in isx-hand-optimized

mppf:
Adjusts preFoldPrimOp handling of PRIM_TO_LEADER and PRIM_TO_STANDALONE to 
avoid adding a duplicate tag argument. It's quite likely that this should be
handled somewhere else (something to do with forall intents, perhaps?) but this
change is simple and unlikely to cause new problems.

Adds tests of the isx case that was failing to compile and also a few other 
iterator forwarding tests.

- [x] full local testing


mppf:
@vasslitvinov - I suspect this bug has to do with forall intents. Would you mind having a look when you get a moment? You can reproduce the original issue by changing the conditional for a tag already found for PRIM_TO_STANDALONE into an assert and then compile the new isx-bug.chpl test.

mppf:
@vasslitvinov - not sure what you mean by a nod to #7323 - is that just a wrong PR number or do you see a relationship I don't here?

vasslitvinov:
@mppf - they are related because they both deal with the extra tag argument. Upon a second look, no big deal, prim-to-leader/follower/standalone need to go away in any case.

PR #9458, opened 2018-05-04
Fix issue #9007 - useless cast double-free

### LINKED ISSUE #9007: Resolve double-free when initializing from casted record

mppf:
Fixes #9007.

- [x] add code from 9007 as test
- [x] full local testing *
- [ ] merge on same day as follow-on-branch PR #9489

### DIFF
--- a/compiler/resolution/preFold.cpp
+++ b/compiler/resolution/preFold.cpp
@@ -1747,25 +1747,30 @@ static Expr* dropUnnecessaryCast(CallExpr* call) {
   if (!call->isCast())
     INT_FATAL("dropUnnecessaryCasts called on non cast call");
 
-  if (SymExpr* sym = toSymExpr(call->castFrom())) {
-    if (LcnSymbol* var = toLcnSymbol(sym->symbol())) {
+  if (SymExpr* fromSe = toSymExpr(call->castFrom())) {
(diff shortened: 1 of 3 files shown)


mppf:
Passed testing except [Error matching program output for memory/sungeun/refCount/domainMaps]

mppf:
PR #9489 includes this change and resolves the testing issue.

PR #9460, opened 2018-05-04
Fix sliced reindex bulk transfers

### LINKED ISSUE #9457: array.reindex produces wrong results in Chapel 1.17.1
### Summary of Problem
Chapel 1.16.0 produced correct results for the following code, but as of 1.17.1 it now produces wrong results. Not sure what's going on.

#### Chapel 1.16.0 output
```
0 1 2 3 4 5 6 7
4 5 6 7 4 5 6 7 // Correct
```

#### Chapel 1.17.1 output up to 1.18.0 pre-release (a6005281c5)
```
0 1 2 3 4 5 6 7
0 1 2 3 4 5 6 7 // Wrong
```

### Steps to Reproduce

**Source Code:**

```chapel
module ReIndex {
  config var data_size = 8;

  proc reindex_test(A: [?D] ?t) {
    const half = data_size / 2;
    var tmp: [0..#half] t;

    // Generate view into A via array.reindex
    ref reA = A.reindex(D);
    // ref reA = A; // This works.

    tmp = reA[(half + D.low)..#half];
    reA[D.low..#half] = tmp;
  }

  proc main() {
    var data = [i in 0..#data_size] i;
    writeln(data);
    reindex_test(data);
    writeln(data);
  }
}
```

benharsh:
An array reindex was incorrectly translating indices from the upward
domain to the downward domain. This was caused by passing the wrong
upward domain class to chpl_reindexConvertDom, and is fixed by passing
'privDom.updom'.

After this bug was fixed, it became apparent that we were also
incorrectly translating strided slices of reindexes. This is solved by
cloning chpl_reindexConvertDom and adjusting it to preserve the slice's
stride.

Some tests have been added or improved to catch related failures.

Resolves #9457 

Testing:
- [x] local + futures
- [x] gasnet


test/optimizations/bulkcomm/bharshbarg/stridedAssignDR.chpl
```diff
@@ -110,6 +110,35 @@ proc test(param rank : int, Dom) {
   if printCount then writeln("Performed ", count, " assignments");
 }
 
+proc testReindex(param rank : int, Dom) {
+
+  var A : [Dom] int;
+  var B : [Dom] int;
+
+  var slices = buildSlices(rank, Dom);
```
mppf: Is this testing a rank-change slice? If not, can you add a test of the similar pattern when including rank changing?

benharsh: There's a test called 'rankChange.chpl' in that same directory, is that adequate?

mppf: yes

PR #9462, opened 2018-05-05
Avoid false positive errors from gcc 8.1

dmk42:
In the normal course of Chapel development while gcc 8.1 was the target compiler, I tripped over a false positive warning not prevented by the Makefile change in #9454 .  We turned the warning into an error with `-Werror`, so this change prevents it.

Passes full local testing.


dmk42:
@ronawho - Would you mind taking a post-commit look at this?

PR #9463, opened 2018-05-05
fix errant resolution in Series

psahabu:
## Motivation
A bug was preventing Series from being used productively by preventing two operations from being used in the same program. More information can be found in the connected issue.

### Probable Cause
Investigation by @cassella showed that a `SeriesAdd(bool)` was being instantiated, which was confusing because such an operation doesn't appear in the reproducer. However, filtering operations have `TypedSeries(bool)` as their return type (i.e. `mySeries < 3`). Creating a separate `TypedSeries(bool)` while commenting out the filter operation reproduced the original error, suggesting that `SeriesAdd(bool)` is being instantiated because `TypedSeries(bool)` exists.

## Fix
@cassella discovered this workaround: Gating `f()`, `f_lhs()`, and `f_rhs()` of `SeriesUnifier` subclasses with `where isNumeric(eltType)`.

Ultimately, a concept like `Summable` (for example) would be preferable so that all types with a `+` operator could be put into a Series, but this works for now.

### DIFF
--- a/test/library/standard/DataFrames/DataFrames.chpl
+++ b/test/library/standard/DataFrames/DataFrames.chpl
@@ -532,7 +532,6 @@ module DataFrames {
    * FIRST CLASS FUNCTION SERENITY NOW.
    */
 
-  // TODO: return tuple versions where first element is "None"
   class SeriesUnifier {
     type eltType;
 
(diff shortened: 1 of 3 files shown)


test/library/standard/DataFrames/DataFrames.chpl
```diff
@@ -550,32 +549,35 @@ module DataFrames {
     }
   }
 
+  // TODO: isNumericType prevents instantiation with bools
+  // would prefer "is summable" type here
```
ben-albrecht: I'm somewhat out of the loop, but would it be worth defining `isSummableType(type t)` in `DataFrames` for now?

psahabu: I don't think so, since at present that would just be `isNumeric` and `isString`. What I meant is that classes should be able to implement a concept such that they are `Summable`.

PR #9464, opened 2018-05-05
basic reductions on Series

psahabu:
## Motivation
See the connected issue.

## Approach
Add methods on `TypedSeries` for `sum()`, `min()`, `max()`, `and() where eltType == bool` and `or() where eltType == bool`. Run a reduction on the `None`-excluding iterator.

### Notes
Included with this PR is a Pandas equivalent to the SumMinMaxSeries test. The most notable difference is that `None` causes the type of the Series to be `float` instead of `int`. A Pandas equivalent for AndOrSeries is not included because the approach is not immediately obvious.

### DIFF
--- a/test/library/standard/DataFrames/DataFrames.chpl
+++ b/test/library/standard/DataFrames/DataFrames.chpl
@@ -265,6 +265,36 @@ module DataFrames {
       halt("generic Series cannot be compared");
       return this;
     }
+
+    pragma "no doc"
+    proc sum() {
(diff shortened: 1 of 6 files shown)


benharsh:
Generally this looks good.

I would sort of prefer that these methods be implemented only on TypedSeries because they need to return the eltType, which Series cannot know. What do you think?

psahabu:
@benharsh That makes sense to me, I'll remove the methods from `Series`.

PR #9471, opened 2018-05-07
Preserve runtime type for formals with 'in' intent

benharsh:
This commit updates wrapper generation to preserve the runtime type of
formals with the 'in' intent. For example:

```chpl
proc foo(in D : domain(1), in A : [D] int);
var X : [1..4] int;
foo({1..4}, X);
```

The formal 'A' is declared to have the runtime type '[D] int'. Anything
else would be surprising/confusing. Prior to this commit we would simply
initCopy X and pass the result to 'foo', which would have a different
runtime type than '[D] int'. To correct this, we can instead create a
temporary for 'A' at the callsite and assign 'X' into that temporary:

```chpl
var default_arg_A : [actual_D] int;
default_arg_A = X;
foo(actual_D, default_arg_A);
```

This commit leverages the pre-existing default_arg creation
functionality to create the temporary into which X is assigned.
Generally, with this commit formal temps are created as if they were
variable declarations.

In order to support this new functionality, FLAG_DEFAULT_ACTUAL is
applied to a certain temporary and later used to avoid inserting an
initCopy. FLAG_NO_AUTO_DESTROY is also applied to default actuals passed
by in-intent.

Testing:
- [x] local + futures
- [x] gasnet
- [x] memleaks


benharsh:
@mppf : Would you be willing to review? This is the first of two PRs that will resolve the array-of-syncs deadlock with initializers.

test/arrays/intents/in/no-domain-linkage.chpl
```diff
@@ -0,0 +1,11 @@
+
```
mppf: I'm having trouble seeing what this test is trying to test? It seems to work on master...

compiler/resolution/wrappers.cpp
```diff
@@ -1678,7 +1824,8 @@ static void handleInIntents(FnSymbol* fn,
     Symbol* actualSym  = info.actuals.v[j];
 
     if (formalRequiresTemp(formal, fn) &&
-        shouldAddFormalTempAtCallSite(formal, fn)) {
+        shouldAddFormalTempAtCallSite(formal, fn) &&
+        actualSym->hasFlag(FLAG_DEFAULT_ACTUAL) == false) {
```
mppf: It'd be nice to have a comment here saying something like "The result of a default argument for an 'in' intent is already owned and doesn't need to be copied"

test/arrays/intents/in/default-value-linkage.chpl
```diff
@@ -0,0 +1,69 @@
+
+var numErrors = 0;
```
mppf: This test is good!

compiler/resolution/wrappers.cpp
```diff
@@ -1649,6 +1692,109 @@ static void addArgCoercion(FnSymbol*  fn,
...
+}
+
+// We have to use the array default if:
+// 1) There is a valid type expr
+//   OR
+// 2) There is a defaultExpr that is not just gTypeDefaultToken
+//
+// We do not want to generate defaults for fully or paritally generic cases:
+//   in A : [] real;
+//   in A = [1,2,3,4];
```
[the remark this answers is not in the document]
benharsh: Yes, you're correct, I'll remove that line.

test/arrays/intents/in/no-domain-linkage.chpl
benharsh: It does work on master, but I didn't find an example of this behavior.

It is verifying that 'in A : []' works like variable initialization in the sense that 'A' shares the same domain as the actual.

benharsh:
@mppf : would you like to discuss any of these comments further?

mppf:
@benharsh - nope, I think your clarifications suffice, thanks

PR #9472, opened 2018-05-07
Update reduction implementation to use initializers

### LINKED ISSUE #9315: Update reductions to use initializers
As a Chapel developer, I want the reductions implementation to use initializers so that we can have more confidence in the initializer implementation and be closer to deprecating constructors.

This issue is currently blocked by the problem described in #9268, which involves the current (and in-flux) partial-reductions implementation.

benharsh:
This PR updates the classes implementing reductions to use initializers. Default constructors have been replaced by applying the 'use default init' pragma, and explicit constructors have been replaced with explicit initializers.

A small compiler change was required to work around a point-of-instantiation issue.

A handful of tests have been notest'd or futurized for one of two reasons:
- incorrect usage of generic instantiation initialization
- usage of nested classes, which are ill-supported for initializers

Resolves #9315 

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/passes/createTaskFunctions.cpp
+++ b/compiler/passes/createTaskFunctions.cpp
@@ -338,7 +338,8 @@ static void addReduceIntentSupport(FnSymbol* fn, CallExpr* call,
   AggregateType* reduceAt = toAggregateType(reduceType->type);
   INT_ASSERT(reduceAt);
 
-  CallExpr* newOp = new CallExpr(reduceAt->defaultInitializer->name,
+  CallExpr* newOp = new CallExpr(PRIM_NEW,
+                                 reduceAt->symbol,
                                  new NamedExpr("eltType", new SymExpr(eltType)));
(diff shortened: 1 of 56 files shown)


benharsh:
@mppf : Did you want those histogram reduction tests to be futurized instead of notest'd?

mppf:
@benharsh - yeah I'd prefer they be .future'd, but a reasonable alternative would be to mention un-no-testing them in some other .future file / linked issue if it's clear enough to you what other future would be fixed at the same tim.

benharsh:
Still wrestling with some point-of-instantiation issues, going to close this issue while I investigate.

benharsh:
@vasslitvinov : can you review the compiler changes, given your familiarity with reductions and foralls?

benharsh:
Just pushed some changes based on vass' feedback. I'll re-test some subset of the suite.

PR #9479, opened 2018-05-08
Fix more buffer overflows and truncations found by gcc 8.1

dmk42:
Now that gcc 8.1 is available on Cray systems, it is finding additional buffer overflows and truncations that were specific to slurm-srun.  This change fixes them.


dmk42:
@ronawho - Would you mind reviewing this?

Issue #6481, opened 2017-06-16
Chapel gcc 7 status


dmk42:
### Summary
Because we have received inquiries from the community regarding using gcc 7.1 with Chapel, this page documents the situation.  We do not normally need to comment on a particular back-end compiler version, but this one has been sought out by our users, since even before it was officially released.

We tested compiling the Chapel compiler, runtime, and generated code with gcc 7.1.  We found it to provide excellent new static analysis features which helped us improve our code.

Due to two optimizer bugs in gcc 7.1, we are not recommending that version of gcc for use with Chapel.  We are switching our nightly gcc testing back to 6.3, with only a small, unoptimized test still running with 7.1.  We are eagerly anticipating future versions of gcc for their static analysis.

### Errors
Both of the following errors occur only when the optimizer is turned on (level `-O2` or higher).  When Chapel is used with the `--fast` flag, it invokes its back-end compiler with `-O3`.

When converting an integer to a floating-point number, gcc 7.1 sometimes copies the bit pattern of the integer instead of performing a conversion.  In C++ terms, it changes a `static_cast` into a `reinterpret_cast`.  This is problematic for Chapel users because of HPC's emphasis on floating-point calculations, and the fact that integers and floating-point numbers are often intermixed.

The other error is more esoteric but can affect the generated code.  Certain combinations of static inline functions and type punning, including valid methods allowed by the C standard, cause gcc 7.1 to abort with an internal compiler error.  We are not sure of the full range of situations that can cause an abort.

As time permits, we intend to narrow these errors to small test cases and submit them to the gcc developers.

ronawho:
Can we disable whatever optimizations are broken if we detect 7.1 is being used?

dmk42:
I just took a look at this.  With a little trickery, we should be able to substitute \-O1 for \-O3 when we are using gcc 7.1.  It would take a lot of hacking to reach down into the third-party configure scripts and make a substitution for that code, though.

ronawho:
Do we really need to drop down to 01? I would have imagined that there's just one or 2 optimizations that are broken that we could disable.

>It would take a lot of hacking to reach down into the third-party configure scripts and make a substitution for that code, though.

Not sure I'm following this

dmk42:
Yes, we really need to drop down to \-O1.  There is an enormous number of optimizations to turn on and off to discover which ones are broken, and they all interact with each other so it isn't guaranteed that removing one from the list would produce something useful.

[...]

PR #9481, opened 2018-05-09
Disable ipa-cp-clone for gcc 7

ronawho:
Disable ipa-cp-clone (function cloning for interprocedural constant
propagation) for gcc 7. This optimization seems to be the cause of the
multi-locale lulesh regression. This seems to have been fixed in gcc 8, so only
disable for gcc 7.1

Without this multi-locale lulesh perf test fails 1/5 runs. It passed 500 trials
with this patch.

This resolves https://chapel.atlassian.net/browse/CHAPEL-306
Related to https://github.com/chapel-lang/chapel/issues/6481


ronawho:
@dmk42 I'm not quite ready to merge this (want to do some more testing and stuff), but wanted to see what you think of this in general

dmk42:
Looks good.  I'm happy to look again if you want to add more.  Otherwise, this is good to merge as is.

PR #9487, opened 2018-05-09
Implement task-private variables:

vasslitvinov:
Task-private variables are discussed in #6325.

The syntax implemented here looks like this:

```chpl
    forall ... with (

      var V1 : v1Type,
      var V2 = v2Init,
      var V3 : v3Type = v3Init,

      const C <the type/init options as for 'var'>,

      ref R = refInit,
      const ref CF = crInit

    ) do .......
```

Overview of changes:

* Parser and related changes in how shadow variables are created during parsing.

* scopeResolve change 1 - now there are two ResolveScope objects
created for a ForallStmt:

    - One ResolveScope is associated with ForallStmt itself. This is new
with this PR. This ResolveScope contains loop index variables, shadow
variables, task-private variables.  It is used for scope-resolving the
with-clause, in particular the initializers and types of task-private
variables.

    - Note that the initializers of task-private variables should not be
referencing loop index variables. Having loop index variables visible
while scope-resolving will allow us to detect such occurrences and
issue errors. This check is currently not implemented.

    - The other ResolveScope is associated with the body of the ForallStmt.
This is the same as before this PR.

    - Note that the iterable expressions such as iterator calls are scope-
resolved outside of both ResolveScopes. So are the references to outer
variables for (non-TPV) shadow variables, which is already on master
since #9354.

* scopeResolve change 2 - getScope() : to map a node to an appropriate
ResolveScope as per the above.

* Arrangement to normalize() so that task-private variables' DefExprs
are normalized by the existing code.

* getOuterVarsNew() et al. need to cater to task-private variables'
init/deinit blocks in addition to the forall loop body.

* Added support for task-private variables to lowerForalls.cpp.
This was one of the easiest/smallest changes in the entire PR,
due to the architecture of this piece of the compiler.

* lowerForalls reorders task-private variables after proper shadow
variables. This is to make lowering work smoothly - see the comment
above reorderShadowVsTaskPrivateVars() for details. We may want to
revisit this, in the long run.

* Added tests in parallel/forall/task-private-vars.

* test/parallel/taskPar/vass/taskintents-badintents
is now a .future - because the errors it checks are now
parse-time errors, so are reported 1 error at a time.
Reporting them all in a single compilation is left as
a (low-priority) future work.
For now I added a non-future test of a single such error.

* While there, minor tweak to an unrelated u1-local-user-extern.bad

Detailed development history for cca56c909c is available
in a96747c961..5f66f5cd8e.


mppf:
Since this adds a language feature, is the language design change discussed somewhere? If so, can you link to the issue? Has the language design had sufficient review?

compiler/AST/foralls.cpp
```diff
@@ -210,31 +210,131 @@ static ForallIntentTag argIntentToForallIntent(Expr* ref, IntentTag intent) {
...
+            const char* nameString, Expr* nameExpr, Expr* type, Expr* init)
+{
+
+  // TPV - task-private variable, as we have a type and/or an initializer.
+  ShadowVarSymbol* result = new ShadowVarSymbol(TFI_TASK_PRIVATE,
+                                                nameString, NULL);
+
+  switch (prefix)
+  {
+  case SVP_VAR:        result->qual = QUAL_VAL;       break;
```
mppf: I generally expect { to increase indent level, is there a particular reason the contents of the switch is not indented?

compiler/AST/foralls.cpp
```diff
@@ -357,7 +457,11 @@ static QualifiedType buildIterYieldType(ForallStmt* fs, FnSymbol* iterFn, FnSymb
         break;
 
       case TFI_TASK_PRIVATE:
-        INT_ASSERT(false); // TODO
+        // task-private variables are TODO too.
+        USR_FATAL_CONT(svar, "Tast-private variables are currently not implemented"
```
mppf: Tast-private sounds tasty but I think you mean Task-private

compiler/AST/symbol.cpp
```diff
@@ -1028,11 +1028,28 @@ bool ShadowVarSymbol::isConstant() const {
...
-  return intent == TFI_CONST_IN;
+  switch (intent) {
+    case TFI_DEFAULT:
+    case TFI_CONST:
+      // Caller responsibility - no abstract intents please.
+      INT_ASSERT(false);
+      return false;
+
+    case TFI_CONST_IN:
+    case TFI_IN_OUTERVAR: // should these two be here?
```
mppf: If "Should these two be here" is a question for me, I don't know, but I suppose it's probably OK to consider them const...

compiler/resolution/implementForallIntents.cpp
```diff
@@ -2176,7 +2184,8 @@ static void appendNewShadowVars(ForallStmt* fs, SymbolMap& outer2shadow) {
 }
 
 // Not to be invoked upon a reduce intent.
-// All our shadow variables are expected to be refs.
+// TODO aren't these flags already set?
```
mppf: Just noting this TODO in case you mean to update it

compiler/resolution/implementForallIntents.cpp
```diff
@@ -2209,7 +2218,8 @@ static void setShadowVarFlagsNew(Symbol* ovar, ShadowVarSymbol* svar, IntentTag
     svar->addFlag(FLAG_CONST_DUE_TO_TASK_FORALL_INTENT);
 }
 
-static void processShadowVarsNew(ForallStmt* fs, BlockStmt* body, int& numShadowVars)
+// Todo: is now time to get rid of pruning?
```
mppf: Just noting this TODO - is this future work or something you are asking for this PR?

compiler/resolution/implementForallIntents.cpp
```diff
@@ -2226,6 +2236,10 @@ static void processShadowVarsNew(ForallStmt* fs, BlockStmt* body, int& numShadow
       svar->qual = QUAL_REF;
       svar->type = dtUnknown;
     }
+    else if (svar->isTaskPrivate())
+    {
+      // Anything to do here?
```
mppf: Probably want to remove / adjust this comment

compiler/resolution/lowerForalls.cpp
```diff
@@ -873,6 +893,89 @@ static void expandTopLevel(ExpandVisitor* outerVis,
...
+  case TFI_TASK_PRIVATE:
+    return false;
+  }
+
+  return false; // dummy
+}
+
+//
+// We want all proper shadow variables to be ahead of task-private variables.
+// Explanation: (a) The latter may reference the former - in TPVs initBlocks.
```
mppf: It might help to have a code example showing how one can refer to the other and another example showing how the other direction is not possible.

compiler/resolution/lowerForalls.cpp
```diff
@@ -873,6 +893,89 @@ static void expandTopLevel(ExpandVisitor* outerVis,
...
+// - because initialization actions may have visible side-effects.
+//
+static void reorderShadowVsTaskPrivateVars(ForallStmt* fs) {
+  AList& svars = fs->shadowVariables();
+  if (svars.empty()) return; // nothing to reorder
+
+  Expr* currDef = svars.head;
+  Expr* lastReorderedDef = NULL;
+
+  do {
```
mppf: I think this loop would probably be easier to understand if you wrote it with 2 lists, where you put the "should go first" ones in one list, and the others in another list, and then output the list elements in the right order. I'm not sure what's gonig on with "not reorderable".

PR #9488, opened 2018-05-10
Good riddance to .good files with no test (and a future too)

cassella:
Remove a few stray .good files and a .future that had no corresponding test.


cassella:
No failures from

```
start_test test/classes/bradc/ test/classes/delete-free test/domains/shannon test/errhandling/psahabu
```

ben-albrecht:
Thanks @cassella - did you use a script to find these, or just happened to notice them?

ben-albrecht:
I'll confirm it passes a paratest before merging.

cassella:
I did use a script, but I don't think it's worth sharing.  Even after hard-coding exceptions for `.lm_flat` and `.comm_none` and `.no-local` etc., there are still hundreds of false positives due to the .good filenames set by  `.compopts` or `.execopts`, and files in `NOTEST` directories, etc.  I think to be done reliably, it would have to be done by something in ``start_test`` that knows all the ways to form a .good filename.

cassella:
Well, fine, here's the "script".  I believe all 590 things it reports
once this PR is merged are false positives.  All the ones I've checked
have `.compopts` or `.execopts` naming them.

```shell
for g in $(find test -name \*.good ); do
   base=${g%.good}
   base=${base%.no-local}
   base=${base%.lm-numa}
   base=${base%.comm-none}
   chpl=${base}.chpl
   c=${base}.test.c
   [ -f $chpl -o -f $c ] || echo $g
done
```


I guess in retrospect it shouldn't be too hard to grep for the exact
`.good` filename in the `*.execopts` and `*.compopts`.

cassella:
This cuts the false positives down a lot, which made it easier to spot at least two true positives, which are incoming.

```bash
#!/bin/bash

for g in $(find test -name \*.good ); do
   base=${g%.good}
   base=${base%.no-local}
   base=${base%.lm-flat}
   base=${base%.lm-numa}
   base=${base%.comm-gasnet}
   base=${base%.comm-none}
   base=${base/.logical.doc/.doc} # ?  many in test/chpldoc
   c=${base}.test.c

   # This dir uses a PREDIFF file to pick a specific .good file
   d=$(dirname $g)
   if [ $d == "test/distributions/robust/arithmetic/performance/multilocale" ]; then
      base=${base%.cyclic}
      base=${base%.replicated}
      base=${base%.block}
      base=${base%.default}
   fi
   chpl=${base}.chpl
   toml=${base}.toml

   if [ -f $chpl -o -f $c -o -f $toml ]; then
      continue
   fi

   # See if this file is named in a compopts or execopts file
   co=$(compgen -G $d/'*.compopts')
   eo=$(compgen -G $d/'*.execopts')

   # Some opts files have "# goodname.good", others just "# goodname"
   if grep $(basename $base) $co $eo /dev/null > /dev/null 2>&1; then
      continue
   fi

   # compopts and execopts with more than one line but no explicit
   # filenames cause foo.chpl to have good files foo.x-y.chpl for x
   # and y in 0 or 1..#lines in file.
   if [ -n "$co" -o -n "$eo" -o -f $d/COMPOPTS -o -f $d/EXECOPTS ]; then
      base=${base%.[0-9]*-[0-9]*}
      chpl=$base.chpl
      c=$base.test.c
      if [ -f $chpl -o -f $c ]; then
      	 continue
      fi
    fi

   echo $g
done
```

I couldn't spot quickly where the `.logical.doc.good` files in `chpldoc` are coming from, so I didn't come up with a more principled way to skip them.

lydia-duncan:
The .logical.doc.good files are for a particular sorting order of chpldoc, which was deprecated but may be re-enabled by a flag in the future

cassella:
Thanks!  I think what the script's on the right track then to consider those .good files the same as `s/.logical.doc/.doc/`, for the purpose of ensuring there's a test for them.

There are a few more improvements I'd like to make to the script before submitting it.  In the meantime, I found another stale .good file.  Incoming...

cassella:
There's also `test/release/examples/benchmarks/lulesh/lulesh-sedov15cube.good`.  AFAICT, this is unreferenced by the test system.  But the commit adding it alongside other updates specifically calls out that it's adding it, but doesn't say why.  (e44778d90)

PR #9489, opened 2018-05-10
Enable fields with generic declared type

### LINKED ISSUE #9007: Resolve double-free when initializing from casted record

### LINKED ISSUE #9234: delete-free subtask: unmanaged as a type
As a Chapel programmer, I would like to be able to specify that arguments or fields are unmanaged pointers, without specifying the exact type.

Acceptance criteria:
Tests exist demonstrating this functionality. To the extent possible, the array/domain/distribution code makes use of the feature.

mppf:
This PR enables records and classes to contain fields which have a generic
declared type, for example:

``` chapel

record R {
  var x: integral;
  var y: MyGenericClass;
}

class MyGenericClass { var x; }
```

Before this PR, such declarations would be an error. In order to do this, it
changes the pattern the normalizer produces to use PRIM_VAR_INIT instead of
default-init/assign.

This PR can be viewed as a continuation of PR #9355.
Closes #9234.
Fixes #9007.

Note that this PR changes the behavior of initializers handling sync fields. It
still does not work but now fails in a different way than it did before. See
classes/initializers/phase1/syncFieldTypeAndInitSyncArg.chpl  and
classes/initializers/phase1/syncFieldTypeOnlySyncArg.chpl. The new test
types/sync/ferguson/sync-copy-init-other-sync.chpl demonstrates the current
failure mode (which also existed before this PR but now also applies to the
field cases - see issue #9516).

- [x] full local testing


compiler/AST/AggregateType.cpp
```diff
@@ -1193,7 +1274,7 @@ CallExpr* AggregateType::typeConstrSuperCall(FnSymbol* fn) const {
   }
 
   if (superTypeCtor->numFormals() > 0) {
-    retval = new CallExpr(parent->symbol->name);
+    retval = new CallExpr(parent->symbol); // parent->symbol->name);
```
vasslitvinov: Clean this up if appropriate.

compiler/passes/scopeResolve.cpp
```diff
@@ -162,12 +162,17 @@ void scopeResolve() {
     }
   }
 
+  //nprint_view(((Symbol*)aid(88514))->defPoint->init);
```
vasslitvinov: Please clean this up.

compiler/resolution/ResolutionCandidate.cpp
```diff
@@ -281,10 +281,10 @@ int ResolutionCandidate::computeSubstitutions() {
 
     if (formal->intent                              == INTENT_PARAM ||
         formal->type->symbol->hasFlag(FLAG_GENERIC) == true) {
-      if (verifyGenericFormal(formal) == false) {
+      /*if (verifyGenericFormal(formal) == false) {
```
vasslitvinov: Remove the commented-out code?

compiler/resolution/functionResolution.cpp
```diff
@@ -4705,7 +4705,14 @@ static bool isFieldAccess(CallExpr* callExpr);
...
 * If this is not the case, then we will expand the call into the normal field *
 * initialization.                                                             *
-*                                                                             *
+*
+* concern for default inits:
+*
+*   make default sync (in default argument handling)
+*   =
+*   deadlock b/c RHS is empty
+*
```
vasslitvinov: The above comment is a bit cryptic. If important, please clarify it. Otherwise remove altogether?

compiler/resolution/functionResolution.cpp
```diff
@@ -5077,38 +5144,42 @@ static void resolveInitField(CallExpr* call) {
...
   call->primitive = primitives[PRIM_SET_MEMBER];
+
+  resolveSetMember(call); // Can we remove some of the above with this?
 }
 
 /************************************* | **************************************
 *                                                                             *
 * This handles expressions of the form                                        *
-*      CallExpr(PRIM_INIT_VAR, dst, src)                                      *
+*      CallExpr(PRIM_INIT_VAR, dst, src, [optional type to coerce to])        *
```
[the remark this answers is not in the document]
mppf: it has 2 or 3 args. I think you are thinking of PRIM_INIT_FIELD, which had 3 before this PR but has 3 or 4 after. But it turns out that I can use PRIM_INIT_VAR and PRIM_INIT_FIELD with the result of that, which is what this PR is actually doing. So the PRIM_INIT_FIELD changes in resolveInitField aren't necessary and I'm expecting to remove them before the final testing runs.

mppf:
@vasslitvinov - see the last 3 commits - it appears some of the work I did in this PR was unnecessary and in fact caused failures with Dimensional after rebasing (since LocDimensionalArr was set to use initializers). I'm running more testing on it now but the dimensional failure is resolved and my new tests are passing. I also added a new test showing the case that I had not taken in to account during the other modifications. Please let me know if you have any concerns.

PR #9494, opened 2018-05-10
Update helper classes in distributions to use initializers

### LINKED ISSUE #9314: Update supporting classes in distributions to use initializers
As a Chapel developer, I want supporting "Loc*" classes in distributions to use initializers so that we can have more confidence in the implementation and be closer to deprecating constructors.

benharsh:
This PR updates the various ``Loc`` helper classes in distributions to use initializers. Many can be trivially converted by applying the 'use default init' pragma, but some need to be translated from constructors to initializers.

Resolves #9314 

Testing:
- [x] local + futures
- [x] gasnet + futures

### DIFF
--- a/modules/dists/BlockCycDist.chpl
+++ b/modules/dists/BlockCycDist.chpl
@@ -449,10 +449,15 @@ class LocBlockCyclic {
   // Constructor computes what chunk of index(1) is owned by the
   // current locale
   //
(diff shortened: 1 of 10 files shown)


benharsh:
This feels borderline trivial to me. If not, would you be willing to review, @lydia-duncan ?

Issue #9243, opened 2018-04-16
Move temp. Chapel changes made for the hand-built Cray XC module w ARM to master branch


awallace-cray:
https://github.com/chapel-lang/chapel/issues/9065 resulted in some platform-specific changes in Chapel.git that I saved on a special dev branch.
* https://github.com/awallace-cray/chapel/commit/df2375006ec53c9f6eedc9a4882daf512c5e6084
* ~https://github.com/awallace-cray/chapel/commit/65806eac232c8654cb8f77d142b26ab20d628b56~
  * 2018-05-10 : no longer needed since hwloc was updated

This task is to get these changes onto master in some form that works for all build/test platforms.
Does not include temp. changes in chapel-code or Jenkins.

awallace-cray:
2018-05-10 : working on https://github.com/awallace-cray/chapel/commit/df2375006ec53c9f6eedc9a4882daf512c5e6084 
aka
https://github.com/awallace-cray/chapel/commit/9360dd6b593b560417eba3b1fe8fa96330b1d223

awallace-cray:
2018-05-15 09:00 : the original task has been completed.

awallace-cray:
2018-05-15 11:00 : cherry-picks to release/1.17 branch
* https://github.com/chapel-lang/chapel/pull/9449 - Allow specifying a node list to aprun
(dmk42 fix enables testing on a 2nd Cray internal system)
* https://github.com/chapel-lang/chapel/pull/9536 - Optionally limit the number of parallel makes to CHPL_MAKE_MAX_CPU_COUNT

PR #9496, opened 2018-05-10
Optionally limit the number of parallel makes to CHPL_MAKE_MAX_CPU_COUNT

awallace-cray:
Chapel build and test scripts have often used parallel makes, like
    "make -j $cpu_count"
where $cpu_count is obtained from the multiprocessing.cpu_count() Python
function.

Unfortunately, Python's cpu_count() can return a ridiculously high number,
when run on ARM-based processors.

This change introduces an optional env variable CHPL_MAKE_MAX_CPU_COUNT,
which sets an upper bound on the number of parallel makes ($cpu_count).
CHPL_MAKE_MAX_CPU_COUNT is highly recommended when running on ARM-based processors.
Also may be used on x86 machines to voluntarily limit the load.

If CHPL_MAKE_MAX_CPU_COUNT is not found or not a positive number, then Python's
cpu_count is used, same as before this change - even on ARM-based processors.

Implements https://github.com/chapel-lang/chapel/issues/9243

Note: requires a corresponding change in chapel-code, to pass CHPL_MAKE_MAX_CPU_COUNT to the ssh processes that execute the builds for each configuration.

### DIFF
--- a/util/buildRelease/chpl-make-cpu_count
+++ b/util/buildRelease/chpl-make-cpu_count
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+""" prints Python cpu_count(), optionally capped by env var CHPL_MAKE_MAX_CPU_COUNT
+"""
+import os
+import multiprocessing
+cpus = multiprocessing.cpu_count()
+try:
+  max = os.getenv('CHPL_MAKE_MAX_CPU_COUNT', '0')
+  if int(max) > 0:
+    cpus = min(int(max), cpus)
+except:
+  pass
+print(cpus)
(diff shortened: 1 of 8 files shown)


ronawho:
Please use something like `CHPL_MAKE_MAX_CPU_COUNT` or something doesn't conflict with the regular CHPL_ env var namespace (test system uses CHPL_TEST_, launchers use CHPL_LAUNCHER_, etc)

awallace-cray:
Note: PR https://github.com/chapel-lang/chapel/pull/9502 is an alternative to this change. #9502 is a much smaller change set that trades flexibility for simplicity.

PR #9504, opened 2018-05-11
Suppress pip 10 warnings for venv builds

ben-albrecht:
`pip 10` includes new warnings and flags to suppress them:

- a warning about installing a scripts into a location that is not included in `PATH`
- a warning about missing dependencies for any pip packages globally installed

We can guarantee that we're using `pip 10` or greater when installing pip from `get-pip.py`, so this PR adds the suppression flags under that condition.

This PR suppresses the `get-pip.py` message unless we are actually installing `pip`.


ben-albrecht:
I plan to merge Monday, since I won't be around for the weekend to monitor testing. I don't take chances when it comes to test infrastructure changes, no matter how trivial.

Issue #9425, opened 2018-05-01
ZMQ interoperability test times out with gasnet


ben-albrecht:
### Summary of Problem

The ZMQ `interop-py` test (added in #7049) currently times out with gasnet configuration. This test was failing silently on nightly gasnet testing until #9072, which surfaced the test timing out. The configuration will be skipped until this issue is resolved.

Something about the gasnet communication layer is interfering with Chapel's ability to connect to the process it spawns. When running `client.py` separately, outside of the Chapel program, the connection succeeds and no timeout occurs.

It's unclear to me whether this is an issue with the test or the module so far.

Note that I have only tested gasnet locally (`GASNET_SPAWNFN=L`) on OS X so far.

### Steps to Reproduce

**Associated Future Test(s):**

 [`test/library/packages/ZMQ/interop-py`](https://github.com/chapel-lang/chapel/blob/master/test/library/packages/ZMQ/interop-py)
### Configuration Information

- Output of `chpl --version`: `chpl version 1.18.0 pre-release (ed162f8)`
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:

```
CHPL_TARGET_PLATFORM: darwin
CHPL_TARGET_COMPILER: clang
CHPL_TARGET_ARCH: unknown
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: none *
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2 *
CHPL_AUX_FILESYS: none
```

nspark:
Any updates on whether this is specific to OS X? I originally added the interoperability tests when running on a Linux laptop, but I don't think I ran the `interop-py` tests with GASNet (only `multilocale`). I can probably try `interop-py` with GASNet this soon-ish.

Are there any known issues between `Spawn.spawn()` and GASNet?

cassella:
I get the hang on Linux too.

If I change the ``Spawn.spawn()`` call to launch ``sleep 60``, and
after starting the server, run ``client.py`` (via wrapper.sh)
[...]

PR #9507, opened 2018-05-11
Convert blocking calls in Spawn module to yielding non-blocking loops

ben-albrecht:
Blocking calls in Spawn module can result in deadlocks when the
subprocess depends on another task that in queue. This converts those
blocking system calls to while loop non-blocking calls with continuous
chpl_task_yields.

Resolves ZMQ gasnet testing  (#9425)

- [x] Remove `CHPL_COMM == none` constraint on ZMQ skipif
- [x] Full `linux64` testing
- [x] `CHPL_COMM=gasnet start_test test/library/packages/ZMQ`

### DIFF
--- a/runtime/src/qio/qio_popen.c
+++ b/runtime/src/qio/qio_popen.c
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2018 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
  * The entirety of this work is licensed under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License.
- * 
+ *
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


ben-albrecht:
@ronawho  - Could you point me to an example of a yielding back-off strategy that I could start from (ideally in the runtime code)?

ronawho:
I think this is fine as-is. It's not obvious to me if there's a "right" back-off strategy here. i.e. when would you increase the length of the blocking `waitpid()`? We don't have a way to determine how busy the subprocess is, and we don't have an easy way to determine how many chapel tasks are queued behind the current one. If there's no tasks queued, chpl_task_yield should be very fast and shouldn't add much interference to any subprocess running.

There are definetly some interesting things to consider here in terms of how to spawn cpu intensive subprocesses and not have chpl tasks (or the qthread workers hosting them) interfere, but I think we should punt on that here, and just address the correctness issue.

For Paul's example. If you have a 4 core system and you spawn a cpu intensive task. If there's no other chpl tasks running, the qthread workers (things that host qthreads/tasks) will eventually wait on a condvar effectively putting themselves to sleep so they shouldn't actively interfere.

PR #9518, opened 2018-05-14
Fix CHPL_PIP_INSTALL_PARAMS append from #9504

ben-albrecht:
Use `+=` to append to `CHPL_PIP_INSTALL_PARAMS`

Also, added `--no-warn-conflicts` to the `get-pip.py` call, since it can also emit conflict warnings, and will always be using pip versions `10` or greater.


ben-albrecht:
Tested locally (OS X) and on the smoke test machine that caused the failure (also OS X)

PR #9536, opened 2018-05-15
Optionally limit the number of parallel makes to CHPL_MAKE_MAX_CPU_COUNT (cp PR #9496)

awallace-cray:
Chapel build and test scripts have often used parallel makes, like
    "make -j $cpu_count"
where $cpu_count is obtained from the multiprocessing.cpu_count() Python
function.

Unfortunately, Python's cpu_count() can return a ridiculously high number,
when run on ARM-based processors.

This change introduces an optional env variable CHPL_MAKE_MAX_CPU_COUNT,
which sets an upper bound on the number of parallel makes ($cpu_count).
CHPL_MAKE_MAX_CPU_COUNT is highly recommended when running on ARM-based
processors. Also may be used on x86 machines to voluntarily limit the load.

If CHPL_MAKE_MAX_CPU_COUNT is not found or not a positive number, then
Python's cpu_count is used, same as before this change - even on ARM-based
processors.

### DIFF
--- a/util/buildRelease/chpl-make-cpu_count
+++ b/util/buildRelease/chpl-make-cpu_count
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+""" prints Python cpu_count(), optionally capped by env var CHPL_MAKE_MAX_CPU_COUNT
+"""
+import os
+import multiprocessing
+cpus = multiprocessing.cpu_count()
+try:
+  max = os.getenv('CHPL_MAKE_MAX_CPU_COUNT', '0')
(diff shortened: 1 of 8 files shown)


awallace-cray:
git cherry-pick 1e759c32421648456a665893a2895107a67142a5

awallace-cray:
2018-05-15 11:00
I meant to get this on master yesterday, but it only merged this morming (PR #9496). 
Now we should wait a day for builds on master branch to succeed before merging this.

PR #9544, opened 2018-05-15
Separate suppression flags out of CHPL_PIP_INSTALL_PARAMS

ben-albrecht:
`make c2chapel` failed on a testing machine using an older system `virtualenv`, which consequently installed an older `pip` version.  This was caused by the `c2chapel/Makefile` including the `chpl-venv/Makefile.include`, which was assuming `pip` 10 or greater whenever `CHPL_PIP` was not set. That assumption broke down here, since the version of `pip` in `c2chapel/Makefile` relies on the version of `virtualenv` instead.

To work around this, I have split off the version-dependent flags into a separate variable which is set and used in the correct conditions.

Separately, this PR fixes a bug I discovered where using the Chapel `virtualenv` from `c2chapel/Makefile` resulted in an import error (needed to update `PYTHONPATH`).

**Testing**
- [x] locally (OS X)
- [x] linux64


ben-albrecht:
@benharsh - would you mind checking the `Makefile/c2chapel` changes?

benharsh:
The c2chapel changes look reasonable to me.

PR #9546, opened 2018-05-15
Remove warning about returned iterator promoted to array

mppf:
This warning was added in #8859. It's OK to remove it since
it was in the 1.17 release. The motivation for removing it right
now is that #it's more convenient for --warn-unstable to
focus on `new` errors.

- [x] full local testing


bradcray:
Looks fine to me.

Issue #9551, opened 2018-05-15
Spike: What scripts could we provide to users to build their own matrix of configurations?


bradcray:
As a system administrator, I'd like to have some support for specifying a matrix of Chapel configurations (e.g., CHPL_COMM=none, ugni, gasnet/aries vs. CHPL_TASKS=qthreads|fifo vs. CHPL_TARGET_COMPILER=...) and having all those configurations be built in one fell swoop rather than doing that scripting myself.

bradcray:
I believe that @thomasvandoren did some groundwork in this area, but don't know that it ever got completed, so I'd imagine this spike being to learn about what we have and what would need to be done to make it user-ready.

ben-albrecht:
> I believe that @thomasvandoren did some groundwork in this area, but don't know that it ever got completed

It is checked in as [`build_configs.py`](https://github.com/chapel-lang/chapel/blob/master/util/build_configs.py).

I wrote something simpler for building the AWS image configurations, but the above looks much further along. I'd start from that.

awallace-cray:
2018-05-16 10:00 : 
* util/build_configs.py is a very general-purpose script to run "makes" for any chosen matrix of CHPL env settings.  Running those "makes" is all it does- there's no help at all for downloading sources, packaging the resulting binaries, or miscellaeous bits like man pages or Python venv. The user can easily select which CHPL env vars to vary (ie, assign values to) and what those values are, choosing from a fixed inventory built-into the script - and the script will grind out the resulting matrix of "makes".  That built-in inventory of available CHPL env settings is a few years old now, but the Python coding makes it easy to update this.  Another thing missing from build_configs.py, is any provision for some CHPL env settings depending on other ones. For example, the Cray module build starts out as a general config matrix like this, but then different things happen to other env settings depending whether COMM is ugni or gasnet.  
[...]

PR #9559, opened 2018-05-16
build_config.py: add --dry-run, --llvm=system, fix CHPL_HOME default

awallace-cray:
Minor updates and improvements to util/build_config.py
Noticed while working https://github.com/chapel-lang/chapel/issues/9551

### DIFF
--- a/util/build_configs.py
+++ b/util/build_configs.py
@@ -102,7 +102,7 @@ Dimensions = [
         'comm', 'CHPL_COMM',
         values=['none', 'gasnet'],
         default=chpl_comm.get(),
-        help_text='Chapel communcation ({var_name}) value to build.',
+        help_text='Chapel communication ({var_name}) value to build.',
     ),
     Dimension(
         'task', 'CHPL_TASKS',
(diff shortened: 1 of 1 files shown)


awallace-cray:
Not urgent.

util/build_configs.py
```diff
@@ -470,6 +475,11 @@ def _format_text(self, text):
         action='store_true',
         help='Verbose output.'
     )
+    parser.add_option(
+        '-n', '--dryrun',
```
ben-albrecht: Maybe nitpicky: `-n` seems like an odd 1-character flag for `--dryrun` to me. Is this a common flag for dry run?

awallace-cray:
```
$ man make
...
       -n, --just-print, --dry-run, --recon
            Print the commands that would be executed, but do not execute them.
```
----
```
$ git clean --help
...
       -n, --dry-run
           Don't actually remove anything, just show what would be done.
```

How about I change the long name to --dry-run? That's what I meant anyway! 
Thanks.

ben-albrecht: Sounds good to me!

Issue #5599, opened 2017-03-09
Organize test/


ben-albrecht:
The `test/` directory is pretty messy. As of writing, here is what the directory hierarchy looks like on master: 

[tree -d $CHPL_HOME/test](https://gist.github.com/ben-albrecht/cc7376bcad9f31fc0a88e3a65ae35709)

This is a **low priority** task, but this issue will allow us to track what is problematic about the current test hierarchy and what can be done to improve it.

For example, an easier first target might be [test/modules](https://gist.github.com/ben-albrecht/cc7376bcad9f31fc0a88e3a65ae35709#file-tree-dat-L551). I think this directory could be significantly cleaned up by consolidating module tests into `standard`, `packages`, and `meta`\* :

```
├── modules
│   ├── standard
│   │   ├── assert
│   │   ├── barrier
│   │   ├── ...
│   ├── packages
│   │   ├── blas
│   │   ├── curl
│   │   ├── ...
│   ├── meta
│   │   ├── ambiguous
│   │   ├── scope
│   │   ├── warnings
│   │   ├── ... 
```

\* `built-ins`, `distributions` and `layouts` could also be included, but they are currently tested elsewhere, so keeping it simple for now.

bradcray:
I don't think 'meta' is right...  the challenge here is that 'test/modules' is currently serving dual purposes:  both testing the language concept of a module (what I think you're calling 'meta', but that sounds way too esoteric for something that's a core component of the language) and the modules that make up the standard library.  I think a cleaner organization would be to make a test/library/ directory that contains the standard/ and packages/ directories you've got above and then a test/modules/ directory that tests the language concept.

bradcray:
(That said, I also want to capture some of the reasons that I generally haven't been supportive of spending the time and energy reorganizing the test/ directory:

* it's not zero effort, which means it distracts from other things that we've identified as higher priorities
* it's not clear to me that the current suboptimal organization is hindering us from making progress or moving the language forward so much as that it isn't as pretty/well-organized as it could be
* git moving files makes it at least slightly more painful to get their history
* moving files when people are working on them can cause challenges

).  To that end I'd be much more supportive of this effort after we've achieved a greater rate of adoption than now.

lydia-duncan:
[...]

PR #9560, opened 2018-05-16
Enable a notest'd MPI test

lydia-duncan:
In looking for extern array examples, I ran across this test, which looked mostly
ready to go but was in a NOTEST directory (due to being submitted before we
had a test setting for it).  Since we didn't otherwise test the functionality in it,
I decided to see what it would take to get it up and running - the
answer was not much!  I copied the execenv and skipif from the multilocale
mpi tests we already run, added a numlocales file since the test doesn't want to
run in a single locale setting, and fixed the .prediff and .good.  With these changes,
the test runs well, so I added it to the gasnet mpi test suite.

The other tests in this directory don't have expected output, so I will save them
for another day.

Verified that the test runs in the appropriate settings, and that the skipif fires
correctly.


ben-albrecht:
Also, FYI @npadmana

ben-albrecht:
These changes seem straightforward to me.

Continuing my campaign of #5599, I'd advocate for moving these tests into `library/packages/MPI` rather than adding this directory to testing, unless there is a good reason not to.

lydia-duncan:
I wasn't sure with a quick glance whether the others were better suited for the multilocale directory or the general one, and figured we would come back to the NOTEST ones faster if they were in a directory with other tests that actually ran (so I didn't want to move test_mpi away from them)

lydia-duncan:
But @npadmana likely has better insight into that

lydia-duncan:
Going to merge now, but Nikhil should feel free to take a look post-merge :)

npadmana:
@lydia-duncan --- thanks for doing this; this looks great!

npadmana:
@lydia-duncan , @ben-albrecht -- I finally got a chance to do a little archaeology to remind myself where this test came from. This was an initial test suite that I had put together for the MPI module. I had started to move some of these tests into ``test/library/packages/MPI/multilocale``, but had never fully finished that move. 

An issue we might run into with the test suite is that it uses blocking MPI calls which is not recommended with ``qthreads``. We've built some wrappers for some of these blocking calls, but not for all (which is probably why I never fully finished migrating all the tests).

I think these should be fine, but we may run into a deadlock (but again, we might not).  

If you'd like, you might want to file a bug report on migrating all of these tests into the official MPI testing repository and assign me to it. We can then fully deprecate this directory....

npadmana:
Final comment -- I think we need to update the skipifs to enforce fifo tasks when using gasnet+mpi, which can deadlock. Just as an aside, I can cause a deadlock on my laptop with gasnet+mpi+qthreads.

lydia-duncan:
#9608 handles this for the test I re-enabled, but not yet for the test/library ones.

Issue #8544, opened 2018-02-21
Improve support for valgrind+gasnet testing


ronawho:
We do nightly single-locale valgrind testing, and that's been incredibly valuable for finding and fixing invalid memory accesses (use-after-free and those sorts of things.)

Unfortunately, multi-locale valgrind testing isn't clean, so it makes diagnosing failures difficult. We should try to clean up multi-locale valgrind testing and run at least a subset of the test suite nightly. For motivation valgrind+gasnet testing would have caught the issue that https://github.com/chapel-lang/chapel/pull/8524 fixed immediately instead of the months it took us to stumble upon it.

I don't think we can expect gasnet+valgrind to work for every configuration, but I'm hoping we can at least get things working for the udp w/ local spawning and smp substrates.

We may also have better luck using an address sanitizer (#6158)

For future reference I found running valgrind manually with `--track-origins=yes` to be helpful for trying to find the origin of the bad memory access.

I tried getting smp and udp conduits working, but I ran into some issues. I timed out trying to get things working, so I'm just noting what I found for future reference:

TL;DR: 

Here's how to try with smp:
```sh
source util/cron/common-valgrind.bash
source util/cron/common-gasnet.bash
export CHPL_COMM_SUBSTRATE=smp
export CHPL_MEM=jemalloc
export CHPL_RT_MAX_HEAP_SIZE=2G
export CHPL_RT_CALL_STACK_SIZE=1M
export GASNET_SUPPRESS=$CHPL_HOME/third-party/gasnet/gasnet-src/other/contrib/gasnet.supp

chpl test/release/examples/hello.chpl -g
valgrind --tool=memcheck -q --trace-children=yes --suppressions=$GASNET_SUPPRESS --track-origins=yes ./hello -nl2
# should be clean (but take ~60s), hello4 will report issues though
```

Here's how to try with udp:
```sh
source util/cron/common-valgrind.bash
source util/cron/common-gasnet.bash
export GASNET_SUPPRESS=$CHPL_HOME/third-party/gasnet/gasnet-src/other/contrib/gasnet.supp
# maybe apply diff below

chpl test/release/examples/hello.chpl -g
valgrind --tool=memcheck -q --trace-children=yes --suppressions=$GASNET_SUPPRESS --track-origins=yes ./hello -nl2

```

SMP:
----
[...]

PR #9561, opened 2018-05-16
Valgrind documentation, with re2 support

psahabu:
The `CHPL_RE2_VALGRIND_SUPPORT` compiler flag was not well documented. This PR creates `Valgrind.rst` to document both the compiler flag and other environment variables needed for `chpl` generated binaries to work with `valgrind`.

This PR also edits `start_test` to require that `CHPL_RE2_VALGRIND_SUPPORT` is set if the `valgrind` and `CHPL_REGEXP` flags are set.


doc/rst/developer/bestPractices/Valgrind.rst
```diff
@@ -0,0 +1,26 @@
...
+------
+
+To run valgrind on an executable generated by the chpl compiler:
+
+.. code-block:: bash
+
+     export CHPL_RT_NUM_THREADS_PER_LOCALE=100
+     export CHPL_TASKS=fifo
+     export CHPL_MEM=cstdlib
+     export CHPL_RE2_VALGRIND_SUPPORT=true
```
ronawho: I would somehow try to note that this may requiring clobbering/rebuilding

doc/rst/developer/bestPractices/Valgrind.rst
```diff
@@ -0,0 +1,26 @@
...
+
+     cd $CHPL_HOME
+     make
+     chpl <program.chpl>
+     valgrind ./<program>
+
+Explanation
+-----------
+
+Some third party libraries in the Chapel runtime will garble the output of
```
ronawho: I would try to describe this a little better than "garble". Something along the lines of:

"""
The above options are needed because not all third-party libraries support valgrind. In particular:
 - jemalloc doesn't doesn't support valgrind, which prevents valgrind from accurately tracking allocations/frees
 -  qthreads performs task-switching in user-space, so valgrind isn't able to keep track of the stack frame correctly
 - re2 intentionally leaves some memory uninitialized for performance reasons (unless `CHPL_RE2_VALGRIND_SUPPORT=true` is set at build time)
 - GASNet support for valgrind is experimental at this time -- see https://github.com/chapel-lang/chapel/issues/8544 for the current status

"""

util/start_test
```diff
@@ -784,6 +784,12 @@ def set_up_general():
         except:
             logger.write("[Error: Could not find valgrind.]")
             finish()
+
+        if "CHPL_REGEXP" in os.environ:
+            if "CHPL_RE2_VALGRIND_SUPPORT" not in os.environ:
+                logger.write("[Error: Rebuild re2 with CHPL_RE2_VALGRIND_SUPPORT set.]")
```
ronawho: I don't know how I feel about this check. I guess it's better than nothing, but since it's a build time option this check can be wrong: 
 - if you had `CHPL_RE2_VALGRIND_SUPPORT` set at build time, but not at start_test time, it will incorrectly error
 - if you have `CHPL_RE2_VALGRIND_SUPPORT` set at start_test time, but not build time, this gives you a false sense of security

PR #9565, opened 2018-05-17
update README, TestSystem.rst for Valgrind.rst

psahabu:
At Vass's suggestion, I have updated `bestPractices/README` and `bestPractices/TestSystem.rst` to reflect the new Valgrind instructions.

### DIFF
--- a/doc/rst/developer/bestPractices/TestSystem.rst
+++ b/doc/rst/developer/bestPractices/TestSystem.rst
@@ -746,6 +746,10 @@ stored in the ``typeTests/`` and ``OOPTests/`` subdirectories.
 If invoked without any arguments, ``start_test`` will start in the current
 directory and recursively look for tests in subdirectories.
 
+If invoked with the ``--valgrind`` flag, ``start_test`` will compile the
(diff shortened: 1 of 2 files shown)


doc/rst/developer/bestPractices/TestSystem.rst
```diff
@@ -746,6 +746,10 @@ stored in the ``typeTests/`` and ``OOPTests/`` subdirectories.
 If invoked without any arguments, ``start_test`` will start in the current
 directory and recursively look for tests in subdirectories.
 
+If invoked with the ``--valgrind`` flag, ``start_test`` will compile the
+program and execute it with ``valgrind``. To learn about best practices with
+``valgrind``, see ``Valgrind.rst``.
```
vasslitvinov: Let us also mention `--valgrindexe`.  For example:

* After "If invoked", use `--valgrindexe` instead of `--valgrind`.
* After the first sentence, add "The flag --valgrind does the same, plus it also runs the compiler under valgrind (which increases testing time compared to --valgrindexe)."

PR #9566, opened 2018-05-17
Fix for createRandomBuffer duplicate names

saru95:
This is a fix for the bug where both the primitive procedure and API procedure had identical names. (and another bug in RSA that I discovered out of the blue)

cc/ @lydia-duncan

### DIFF
--- a/modules/packages/Crypto.chpl
+++ b/modules/packages/Crypto.chpl
@@ -888,7 +888,7 @@ proc bfEncrypt(plaintext: CryptoBuffer, key: CryptoBuffer, IV: CryptoBuffer, cip
 
        .. code-block:: chapel
 
-          var a = (new CryptoRandom()).createRandomBuffer(5)
+          var a = (new CryptoRandom()).getRandomBuffer(5)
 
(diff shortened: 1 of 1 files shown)


lydia-duncan:
These seem reasonable.  Mind checking in a test that exercises them?

lydia-duncan:
(I'm having difficulty testing this branch due to unrelated internal build issues.  Once those are resolved, I'll try a full paratest again)

saru95:
> Mind checking in a test that exercises them as well?

I manually checked them by using them in small snippets but as far as tests are concerned, I still haven't figured out a way to test it since it returns a random value on every run. For RSA, I'm going to add a bunch of tests separately in a new PR, since it requires some changes.

lydia-duncan:
Yeah, but we want to know it at least compiles and runs.  Maybe a good test would call it but not output the random characters directly?  Or maybe ensure they aren't all the same value or something?

lydia-duncan:
Still would like to see the test for getRandomBuffer, but since testing random numbers is tricky I was fine with merging anyways

PR #9569, opened 2018-05-17
Tighten up resolveCoerce

mppf:
resolveCoerce adjusts PRIM_COERCE in a special manner
to handle the return type for tuples that are returned.
But this adjustment does not make sense for variable initialization,
and now PRIM_COERCE is used in variable initialization. So limit
the adjustment to the case in which the result of PRIM_COERCE is returned.

- [x] full local testing


mppf:
PR #9572 removed the need for the logic that was adjusted here.

PR #9570, opened 2018-05-17
Remove BlockCycDist's copy initializer

lydia-duncan:
The copy initializer wasn't a true copy initializer, since it had an extra
argument for the type and param fields.  Those aren't necessary in the
initializer world, and with the arguments removed, the initializer looks
basically identical to what the compiler generated copy initializer would look
like, so remove it entirely.

Still needs testing over:
- [x] local paratest (out of paranoia)
- [ ] dist cyclic nightly testing
- [ ] gasnet with release/examples and multilocale/


lydia-duncan:
Blocked by #9580

PR #9572, opened 2018-05-17
Add tests from delete-free design, fix a few bugs

mppf:
* Add tests from the delete-free design release notes
* Fixes a few bugs discovered in the process of running these tests
   * default intent for owned needs to be `in` vs `const in`
   * resolveSetMember needed updating to handle generic fields with declared type
   * insertCasts adjusted so that `var x:MyClass = new owned MyClass()` destroys the temp at some point
 * Fix problems with tuples
   * initCopy wasn't removing reference types
    * blank intent tuples of tuples did not store the inner tuple by ref (which it should, because that's blank intent for tuples)
    * returning a tuple didn't interact correctly with array type return, see tuple-slices-returned.chpl for a new test that now passes
    * adjusting return type for function returning a tuple was unnecessarily complex in comparison to the similar operation for arrays
* Fixes problems with classes/delete-free/owned-implicit-copy-tuple passes under valgrind

- [x] full local testing


test/classes/ferguson/generic-field/generic-field-record-not-inited3.future
```diff
@@ -0,0 +1,2 @@
+bug: record with generic field default-of problem
+#9568
```
cassella: As far as I can tell, the `...inited3.chpl` test is identical to the `...inited2.chpl` test, but has a `.future` with a different bug description, but citing the same issue number.  The `.good` files are the same as well.  Was that what was intended?

test/classes/delete-free/tests-from-design-overview/borrowed-arguments.chpl
```diff
@@ -0,0 +1,22 @@
...
+    writeln("in MyClass.init");
+  }
+  proc deinit() {
+    writeln("in MyClass.deinit");
+  }
+}
+
+var global: borrowed MyClass; // 'borrowed' optional here
+proc saveit(arg: borrowed MyClass) { // and here
+  global = arg; // Error! trying to store borrow from local 'x' into 'global'
```
cassella: Should the comment say "local 'arg'"?

test/classes/ferguson/generic-field/generic-field-record-not-inited3.future
mppf: The .future and .good contents were intentional but I meant to commit the version of the .chpl that included the commented out code. Now fixed, thanks.

mppf: I intentionally didn't put one because it would be a pain to maintain (it includes all _defaultOf overloads in a compile-time error message)

test/classes/delete-free/tests-from-design-overview/borrowed-arguments.chpl
mppf: In this case, it's referring to 'x' in the other routine (test); it's intentional, but at the moment 'arg' has the value of 'x'. So the other interpretation is not wrong.

compiler/include/resolution.h
```diff
@@ -255,6 +255,10 @@ AggregateType* computeNonRefTuple(AggregateType* t);
 
 AggregateType* computeTupleWithIntent(IntentTag intent, AggregateType* t);
 
+void addTupleCoercion(AggregateType* fromT, AggregateType* toT, Symbol* fromSym, Symbol* toSym, Expr* insertBefore);
+
```
vasslitvinov: Too many blank lines. :)

compiler/resolution/functionResolution.cpp
```diff
@@ -6565,6 +6549,7 @@ static bool isRefWrapperForNonGenericRecord(AggregateType* at) {
 static void resolveCoerce(CallExpr* call) {
   resolveGenericActuals(call);
 
+  /*
```
vasslitvinov: Please clean up this commented-out code as appropriate.

compiler/resolution/resolveFunction.cpp
```diff
@@ -1026,7 +1080,7 @@ void resolveReturnTypeAndYieldedType(FnSymbol* fn, Type** yieldedType) {
 
   // For tuples, generally do not allow a tuple to contain a reference
   // when it is returned
-  if (retType->symbol->hasFlag(FLAG_TUPLE)   ==  true &&
+  /*if (retType->symbol->hasFlag(FLAG_TUPLE)   ==  true &&
```
vasslitvinov: Please clean up this commented-out code as appropriate.

compiler/resolution/tuples.cpp
```diff
@@ -584,6 +587,129 @@ static void instantiate_tuple_init(FnSymbol* fn) {
 *                                                                             *
 ************************************** | *************************************/
 
+void addTupleCoercion(AggregateType* fromT,
```
vasslitvinov: This function is too long. How about split it into shorter chunks?

For example, the code handling the five cases can be moved out to a separate function - or two five separate functions.

Also, the USR_FATAL_CONT call at the beginning can fit in two lines instead of four.
The formals can fit in three lines. These will make the function easier to read.

compiler/resolution/tuples.cpp
```diff
@@ -758,7 +896,9 @@ instantiate_tuple_unref(FnSymbol* fn)
   AggregateType* ct;
   getTupleArgAndType(fn, arg, origCt);
 
-  ct = computeNonRefTuple(origCt);
+  //const char* useCopy = "chpl__autoCopy"; // TODO -- shouldn't it be initCopy?
+  const char* useCopy = "chpl__initCopy";
```
vasslitvinov: Please reduce the two lines above to one.

compiler/resolution/wrappers.cpp
```diff
@@ -1613,9 +1618,57 @@ static void addArgCoercion(FnSymbol*  fn,
...
+    if (prevActualSym->hasFlag(FLAG_EXPR_TEMP))
+      castTemp->addFlag(FLAG_EXPR_TEMP); // for lvalue checking
+
+    if (prevActualSym->hasFlag(FLAG_REF_TO_CONST) ||
+        prevActualSym->isConstant() ||
+        prevActualSym->isParameter()) {
+      castTemp->addFlag(FLAG_REF_TO_CONST);
+    }
+
+    /*
```
vasslitvinov: Please clean up.

compiler/resolution/wrappers.cpp
```diff
@@ -1613,9 +1618,57 @@ static void addArgCoercion(FnSymbol*  fn,
...
+                     formal->name,
+                     fn->name);
+      USR_STOP();
+    }*/
+
   } else if (ats->hasFlag(FLAG_REF) &&
-             !(ats->getValType()->symbol->hasFlag(FLAG_TUPLE) &&
-               formal->getValType()->symbol->hasFlag(FLAG_TUPLE)) ) {
+            !(ats->getValType()->symbol->hasFlag(FLAG_TUPLE) &&
+              formal->getValType()->symbol->hasFlag(FLAG_TUPLE))) {
```
vasslitvinov: I prefer to keep the original formatting in the two lines above.

PR #9567, opened 2018-05-17
Improve Chapel LLVM support

mppf:
Update Chapel to work with master LLVM and improve script for working with LLVM master

* fixes minor API changes from LLVM 7
* enables update-llvm script to take an argument specifying branch to use
* update-llvm also checks out Polly to support current development efforts
* prototype RV integration included but off by default. It is not ready for prime-time but I didn't want to lose track of the small number of changes.

For example, here is the way to use update-llvm.sh to checkout master version of llvm/clang/...:

``` bash
# to enable LLVM debug information (not required but might be helpful)
export CHPL_LLVM_DEVELOPER=1
# to enable building LLVM with Chapel compiler
export CHPL_LLVM

# Clear out previous llvm directory / build
make clobber

# check out the branch of LLVM you want to use
# e.g. release_601
# (if no branch is specified, the script will use a branch name
#  computed from LLVM_VERSION
#  which is the version bundled with Chapel)
# Note, running update-llvm again without clobbering
# will 'git pull --rebase'
cd third-party/llvm
./update-llvm.sh master
cd ../..

# build LLVM and Chapel compiler using it
make

# add built LLVM tools to your path
export PATH=`pwd`/third-party/llvm/install/linux64-gnu/bin/:$PATH
```

- [x] full local --llvm testing

Reviewed by @dmk42 - thanks!

PR #9577, opened 2018-05-17
Fix reversed sense of vectorize enable flag

dmk42:
In #9567 , explicit vectorization metadata was enabled when not using the Region Vectorizer, causing messages about failed vectorization to print out.  This causes some tests to fail.

This happened because the metadata was enabled when not using the RV, and disabled when using the RV.  I believe it was meant to be the other way around.


dmk42:
@mppf - Would you please do a post-commit review?

mppf:
@dmk42 - thanks for the fix - I believe you are right about this

PR #9585, opened 2018-05-18
Improve tolerance of LLVM 64-bit ARM ABI

dmk42:
The 64-bit ARM calling sequence ABI is a little different from x86_64.  A struct with a size less than 128 bits is rounded up to a multiple of 64 bits.  This change tolerates that difference so that ARM code can work.

This change brings us from zero working ARM tests to all of release/examples working except for the ra benchmarks.  Because testing shows that x86_64 has not been broken in the process, and because this is an improvement, I would like to preserve this change by committing it even though not all of the ARM tests work yet.

The ra benchmarks on ARM get an internal LLVM assertion failure from "extracting off end of register," which means I must be missing a special case.  The assertion failure does not occur until the makeBinaryLLVM() stage, so it will take a while to debug.


dmk42:
@mppf - Would you please review this?

dmk42:
@mppf - Thanks.  Yes, the Clang interface would be helpful if we could use it.  I looked into this in detail.  It requires that we generate Clang AST nodes, so that would be a major overhaul.

There is an interesting ABI library that should work better in the long run.  We ought to keep an eye on this.  For now, it only works on x86, though.

https://github.com/scrossuk/llvm-abi

PR #9586, opened 2018-05-18
Preserve non-ref type for ref-intent formals

vasslitvinov:
This is a step towards eliminating ref types from (most of) the compiler,
by disabling the conversion to a ref type during resolution
for formals with explicit ref intents.

This matches the already-existing behavior of formals with the default
intent, which have non-ref types even if they are records.

To make this work, this change adjusts codegen to add address-of in
certain cases, and removeDeadStringLiteral() to match the more concise
string-literal-initialization pattern that this change results in.

Testing: linux64 --verify and gasnet on multilocale tests.


vasslitvinov:
@benharsh would you review?

PR #9588, opened 2018-05-18
fix RandomNumber6 on baseline

psahabu:
`test/exercises/RandomNumber6.chpl` was segfaulting in the baseline testing configuration as a result of #9450. The likely culprit is that the deprecation of out error in favor of error handling increased the size of each stack frame, thus requiring a larger stack size to allocate all the necessary memory.  This change expands the stack size under start_test.


cassella:
Is that really the issue number you mean?

psahabu:
Whoops, #9450.

PR #9590, opened 2018-05-19
Update two futures' .goods with real real output

cassella:
Passes `start_test --futures test/classes/delete-free/unmanaged`


lydia-duncan:
Please read the tests more carefully.  The type field does not actually impact the var field's type, so these are always going to be ints

cassella:
Doh, sorry!

PR #9608, opened 2018-05-23
Update skipif to enforce fifo tasking, at Nikhil's recommendation

lydia-duncan:
As recommended [here](https://github.com/chapel-lang/chapel/pull/9560#issuecomment-391231126)

TODO: apply this also to the tests in library/packages/MPI?


ben-albrecht:
>> We've built some wrappers for some of these blocking calls, but not for all (which is probably why I never fully finished migrating all the tests).

>  I think the tests in library/packages/MPI should be safe, since they use the wrappers he built?

That sounds right, actually.

Issue #9580, opened 2018-05-18
Can't take unmanaged generic class as function argument


lydia-duncan:
### Summary of Problem
We encounter an internal compiler error when a function declares that its argument is unmanaged but also the generic version of the type.  This seems like an unfortunate backslide in functionality when compared to the old strategy for classes.

### Steps to Reproduce

**Source Code:**

```chapel
class Foo {
  type t;
  var v: int;
}

proc takesUnmanagedFoo(x: unmanaged Foo) {
  writeln(x);
}

var foo1 = new unmanaged Foo(int, 3);
var foo2 = new unmanaged Foo(real, 4);
takesUnmanagedFoo(foo1);
takesUnmanagedFoo(foo2);
delete foo1, foo2;
```

**Compile command:**
`chpl unmanaged-generic2.chpl`

**Associated Future Test(s):**
[`test/classes/delete-free/unmanaged/unmanaged-generic2.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/classes/delete-free/unmanaged/unmanaged-generic2.chpl) #9581

### Configuration Information

- Output of `chpl --version`: chapel 1.18.0 (pre-release)

lydia-duncan:
@mppf - heads up, I think this is causing one of the default initializer failures (and I encountered it in another setting)

lydia-duncan:
I suspect owned and shared are similarly impacted

mppf:
Yes this is a known issue in the backlog -- #9521

lydia-duncan:
This one doesn't involve a query, but the examples I see in that issue all do, so I think this is a separate (but related) issue

lydia-duncan:
I'll close the duplicate of that one I just opened, sorry

mppf:
Yeah, anyway thanks for the .futures. I intend to fix both at once when I work on #9521.

lydia-duncan:
Sounds good!  Would you mind checking to make sure they are otherwise written correctly, since I'm not used to writing for owned/shared/unmanaged?

Issue #9521, opened 2018-05-14
delete-free subtask: spike on generic arguments with owned/unmanaged/etc


mppf:
As a Chapel user, I would like to be able to use the right type query mechanism available to me in Chapel for field types within a generic class. I would like to use these type query mechanisms with `unmanaged` and `owned` because it would not occur to me that `unmanaged` or `owned` would impact them, since the types are independent from memory management strategy.

This is not expected to be a breaking change. It fills out the delete-free story.

Generic arguments with queried type parameters worked with "raw" classes but don't currently work with `unmanaged` or `owned`. The purpose of this spike is to understand how these queried type parameters are implemented and identify a strategy to get them working for `owned` or `unmanaged`.

For example:
``` chapel
class MyClass {
  var x;
}
proc f(arg: MyClass(?t)) { }
f(new MyClass(1));
```

works today but this does not:
``` chapel
class MyClass {
  var x;
}
proc f(arg: unmanaged MyClass(?t)) { }
f(new unmanaged MyClass(1));
```

See also #9580.

mppf:
I think this could work in a straightforward way with unmanaged, but for owned, it would be best to use a where clause about the inner type, for now at least.

mppf:
Or, since we can have a type with a generic field, maybe owned partially instantiates to that type?

mppf:
The compiler translates arg:C(?t) into arg:C and then a use of PRIM_QUERY. This mechanism in normalize (isQueryForGenericTypeSpecifier, expandQueryForGenericTypeSpecifier) doesn't currently handle nested cases like arg:A(B(?t)), which is basically what's happening with unmanaged.

I think that it's possible to change the meaning of PRIM_QUERY when it's supplied more than one index, but it seems to already do something there now so I need to check if that's used anywhere.

PR #9612, opened 2018-05-23
warn-unstable flag warns for undecorated class type

### LINKED ISSUE #9288: delete-free subtask: warning on un-decorated class type
As a Chapel user migrating code to the new class ownership system, it would help significantly in such migrations if the compiler can issue a warning every time I use a class type in a way that will change meaning in the new system. This will allow me to more easily identify which portions of my code need to be updated.

See design discussion here: https://github.com/chapel-lang/chapel/issues/9278#issuecomment-382852278

This should help migrating code under the existing breaking changes for the delete-free design.

Acceptance criteria:
Warning implemented for `--warn-unstable`. Test demonstrates the new functionality.

mppf:
This PR adds a warning when `--warn-unstable` is used.  With the flag, the
compiler now warns for undecorated class type usage.

For example:

``` chapel
class MyClass { var x:int; }
var a: MyClass; // this results in warning
proc f(arg: MyClass) { } // so does this

var b: unmanaged MyClass; // this does not
var c: borrowed MyClass; // this does not
var d: owned MyClass; // this does not
var e: shared MyClass; // this does not
```

This warning is useful since the undecorated class type is changing
(from something like `unmanaged MyClass` to `borrowed MyClass`).

Additionally this PR includes initial attempts to get Hello World to compile
without warning when the warning applies to the standard and internal modules.
Since I was not able to get all of the way there (due to issues #9521 / #9580),
the warning only applies to user code at the moment.

* error handling compiler code now tolerates throwing unmanaged
* new warning is added to scopeResolve
* function resolution updated to avoid error for unmanaged = nil
* many internal and standard modules now use unmanaged and borrowed

Note, this PR does not warn on un-decorated uses of generic classes.
That case will be handled in a follow-on PR (#9677).

Resolves #9288.

### DIFF
--- a/modules/internal/ChapelIO.chpl
+++ b/modules/internal/ChapelIO.chpl
@@ -203,8 +203,8 @@ module ChapelIO {
   // TODO -- this should probably be private
   pragma "no doc"
   proc _isNilObject(val) {
-    proc helper(o: object) return o == nil;
-    proc helper(o)         return false;
+    proc helper(o: borrowed object) return o == nil;
+    proc helper(o)                  return false;
     return helper(val);
   }
 
@@ -257,7 +257,7 @@ module ChapelIO {
       var isBinary = writer.binary();
 
       if (isClassType(t)) {
-        if t != object {
+        if t != borrowed object {
           // only write parent fields for subclasses of object
           // since object has no .super field.
           writeThisFieldsDefaultImpl(writer, x.super, first);
@@ -401,7 +401,7 @@ module ChapelIO {
       var superclass_error : syserr = ENOERR;
 
       if (isClassType(t)) {
-        if t != object {
+        if t != borrowed object {
           // only write parent fields for subclasses of object
           // since object has no .super field.
           type superType = x.super.type;
--- a/modules/internal/ChapelIO.chpl
+++ b/modules/internal/ChapelIO.chpl
@@ -257,7 +257,7 @@ module ChapelIO {
       var isBinary = writer.binary();
 
       if (isClassType(t)) {
-        if t != borrowed object {
+        if _to_borrowed(t) != borrowed object {
           // only write parent fields for subclasses of object
           // since object has no .super field.
           writeThisFieldsDefaultImpl(writer, x.super, first);
@@ -401,7 +401,7 @@ module ChapelIO {
       var superclass_error : syserr = ENOERR;
 
       if (isClassType(t)) {
-        if t != borrowed object {
+        if _to_borrowed(t) != borrowed object {
           // only write parent fields for subclasses of object
           // since object has no .super field.
           type superType = x.super.type;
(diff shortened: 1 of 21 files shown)


test/compflags/ferguson/unstable-class.chpl
```diff
@@ -0,0 +1,46 @@
...
+  var dg:unmanaged MyGenericClass(int);
+  var eg:owned MyGenericClass(int);
+  var fg:shared MyGenericClass(int);
+  var gg:Owned(MyGenericClass(int));
+  var hg:Shared(MyGenericClass(int));
+}
+
+
+errors();
+ok();
```
vasslitvinov: Do we also need to test the warning on undecorated types of formals?

modules/internal/ChapelIO.chpl
```diff
@@ -257,7 +257,7 @@ module ChapelIO {
       var isBinary = writer.binary();
 
       if (isClassType(t)) {
-        if t != object {
+        if t != borrowed object {
```
vasslitvinov: Please double-check that all class types arriving here are "borrowed".

Seems like they are, because this is invoked from writeThis() that's generated by compiler's buildDefaultReadWriteFunctions(AggregateType* ct), and an AggregateType means it's a "borrow" ?

What if the user calls writeThisFieldsDefaultImpl()?

Maybe worth adding compilerAssert(t is a borrowed type).
Or just convert `t` to `borrowed` before comparing against `borrowed object`.

modules/internal/ChapelIO.chpl
```diff
@@ -401,7 +401,7 @@ module ChapelIO {
       var superclass_error : syserr = ENOERR;
 
       if (isClassType(t)) {
-        if t != object {
+        if t != borrowed object {
```
vasslitvinov: Same as above.

modules/internal/StringCasts.chpl
```diff
@@ -46,13 +46,13 @@ module StringCasts {
   proc _cast(type t, x: string) throws where isBoolType(t) {
     var str = x.strip();
     if str.isEmptyString() {
-      throw new IllegalArgumentError("bad cast from empty string to bool");
+      throw new unmanaged IllegalArgumentError("bad cast from empty string to bool");
```
vasslitvinov: Geez, it is really unfortunate that the users will need to say what kind of error they are allocating for a "throw". I wish we could allow simply `throw new MyErrorClass(...)` and have the compiler figure out whether it should be unmanaged or what.

test/compflags/ferguson/unstable-class.chpl
vasslitvinov: Also, please add a test of types in an extern function, to illustrate that those do not need decorations because they are assumed to be unmanaged.

modules/internal/StringCasts.chpl
mppf: I moved this to issue #9698.

PR #9619, opened 2018-05-23
Create, but do not use, an IfExpr node

benharsh:
This is the first step towards changing our implementation of if-expressions, which currently lower to a nested function. The nested-function implementation has caused a variety of problems over the years when implementing other features.

This new 'IfExpr' node will take the place of the nested function from parse() through normalize(), at which point it will be lowered to a CondStmt. This PR simply introduces the type and wires it into the existing BaseAST-related macros/templates/functions. This PR does not cause the compiler to create any new AST nodes.

Testing:
- [x] local + futures


benharsh:
@mppf : I intend to implement AST-viewing capability in the follow-on PR.

compiler/AST/expr.cpp
```diff
@@ -1091,6 +1091,99 @@ void NamedExpr::accept(AstVisitor* visitor) {
   }
 }
 
+//
+// Implementation of 'IfExpr'
```
mppf: Can we put the implementation in compiler/AST/IfExpr.cpp ?

compiler/codegen/expr.cpp
```diff
@@ -5666,6 +5666,12 @@ GenRet NamedExpr::codegen() {
   return ret;
 }
 
+GenRet IfExpr::codegen() {
+  GenRet ret;
+  INT_FATAL(this, "IfExpr::codegen not implemented");
+  return ret;
+}
```
mppf: For this kind of thing, I wouldn't add compiler/codegen/IfExpr.cpp. It could stay here or go in compiler/AST/IfExpr.cpp.

compiler/include/expr.h
```diff
@@ -294,6 +294,33 @@ class NamedExpr : public Expr {
   virtual Expr*   getFirstExpr();
 };
 
+class IfExpr : public Expr {
```
mppf: Can it go in to compiler/include/IfExpr.h ?

compiler/AST/expr.cpp
```diff
@@ -1091,6 +1091,99 @@ void NamedExpr::accept(AstVisitor* visitor) {
...
+    condition->accept(visitor);
+    thenStmt->accept(visitor);
+    elseStmt->accept(visitor);
+
+    visitor->exitIfExpr(this);
+  }
+}
+
+QualifiedType IfExpr::qualType() {
+  return thenStmt->body.tail->qualType();
```
mppf: This is a lie, but it's probably harmless. I can't really tell just from the AST nodes how you'll handle type resolution for these. But, I'd expect that if-exprs won't actually exist by resolution. Does that mean that this function could be an assert? If that works OK I think I'd prefer it, since it's not this simple, and I worry the lie here will cause bugs later.

benharsh: Good point. I didn't write this because I observed a need for an IfExpr's type, I just assumed something during normalize() would call it. I'll add an assertion on the follow-up branch and see if anything fails.

PR #9636, opened 2018-05-26
Clean up sub_clean's cleanCleanfiles

cassella:
Correct and improve the comment (c is the file, not f),
and rename f to d since it's a directory.

### DIFF
--- a/util/test/sub_clean
+++ b/util/test/sub_clean
@@ -53,10 +53,10 @@ def cleanChapelTest(f):
             pass
         cleanCleanfiles(None, execname+'.cleanfiles', False)
 
-# Clean files listed in the file f
-def cleanCleanfiles(f, c, rmCore):
-    if f:
-        cleanfile = f+'/'+c
+# Clean files listed in the file c, in the directory d (if given)
+def cleanCleanfiles(d, c, rmCore):
+    if d:
+        cleanfile = d+'/'+c
     else:
         cleanfile = c
     if os.access(cleanfile,os.R_OK):
@@ -64,8 +64,8 @@ def cleanCleanfiles(f, c, rmCore):
         if rmCore:
             cleanfiles += ['core']
         for file in cleanfiles:
-            if f:
-                globfiles = glob.glob(f+'/'+file)
+            if d:
+                globfiles = glob.glob(d+'/'+file)
             else:
                 globfiles = glob.glob(file)
             for g in globfiles:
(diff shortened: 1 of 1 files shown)


cassella:
Lightly tested by running start_test on a test with a `.cleanfiles`.

util/test/sub_clean
```diff
@@ -53,19 +53,19 @@ def cleanChapelTest(f):
...
         cleanfile = c
     if os.access(cleanfile,os.R_OK):
         cleanfiles = ReadCleanfiles(cleanfile)
         if rmCore:
             cleanfiles += ['core']
         for file in cleanfiles:
-            if f:
-                globfiles = glob.glob(f+'/'+file)
+            if d:
+                globfiles = glob.glob(d+'/'+file)
```
ronawho: Any interest in switching things like `d+'/'+file` over to `os.path.join(d, file)` while you're here?

util/test/sub_clean
```diff
@@ -53,19 +53,19 @@ def cleanChapelTest(f):
...
+def cleanCleanfiles(d, c, rmCore):
+    if d:
+        cleanfile = d+'/'+c
     else:
         cleanfile = c
     if os.access(cleanfile,os.R_OK):
         cleanfiles = ReadCleanfiles(cleanfile)
         if rmCore:
             cleanfiles += ['core']
         for file in cleanfiles:
```
ronawho: Maybe `file` => `f` since `file` gets highlighed and is kinda reserved?

ronawho:
Nice, looks good. I left a few inline comments for other potential cleanups if you're interested in making those while you're here. If not, I'm happy to merge as-is

cassella:
Ok, those changes are incoming.

ronawho:
Ah, I see what you were talking about with some of the calls to glob not making sense with os.path.join. You could replace the remaining `/`'s with `os.path.sep`, but I don't feel strongly about that.

This looks great to me, and testing passed (I just did a full paratest and ran `start_test --clean-only` to make sure the same files were being cleaned up w/ and w/o this PR.)

I'll merge tomorrow.

cassella:
I eventually decided the join call wouldn't be worse than the `d+"/"+file`, whatever `d` and `file` are.

Note, at least as far as the globbing is concerned, the globs won't match anything if there aren't files left over from running the tests once first.

cassella:
Oh, I think I was reading too quickly.  The `--clean-only` was after the paratest, not how you ran it.  Never mind that last part.

PR #9637, opened 2018-05-26
Fix copyFile() to be newline-agnostic

cassella:
Use readstring() instead of readline() to avoid complications from
files that don't end with newlines.

Add tests.


cassella:
I looked at `readBytes()`, but that interface seems to require the
caller not to try to read more bytes than are left in the file, even
after #8856.  I couldn't find another interface that reads into an
array of bytes, vs a string.

I didn't test for a performance impact from this change.

mppf:
Passed full local testing.

cassella:
> Note also that I think it's a bug that readline has problems with
  null bytes in strings. I'd appreciate an issue / future about that
  if you're up for making one.

I don't think it has a problem with null bytes.  AFAICT it's only the
lack of a trailing newline that confuses it.  In a test copying some
random data that has 0x0 bytes before the last 0x0a newline byte, the
0x0 bytes are copied correctly, before everything after the last
newline is dropped.



That did expose some curious `readline()` behavior at EOF though.  The
sum of the `numRead`s from all the calls (including the last one where
the return value was false) was one more than the size of the file.
That's due to this code:

```chapel
    while i <= maxIdx {
      got = qio_channel_read_byte(false, this._channel_internal);
      arg[i] = got:uint(8);
      i += 1;
      if got < 0 || got == newLineChar then break;
    }
    numRead = i - start;
    if got < 0 then err = (-got):syserr;
```

At EOF, I'm thinking `qio_channel_read_byte()` returns a negative int
as an error.  Then `readline()` stores that error byte as `uint(8)`
into the buffer, and counts it in the `numRead` being returned.  Is
that right?

mppf:
> At EOF, I'm thinking qio_channel_read_byte() returns a negative int as an error. Then readline() stores that error byte as uint(8) into the buffer, and counts it in the numRead being returned. Is that right?


I believe you understand the situation correctly. It's true that read_byte returns negative on an error (including EOF) - the rest looks like a good explanation from your experiments and the code here.

What if you change it to
``` chapel
   while i <= maxIdx {
      got = qio_channel_read_byte(false, this._channel_internal);
      arg[i] = got:uint(8);
      if got < 0 then break;
      i += 1;
      if got == newLineChar then break;
    }
    numRead = i - start;
    if got < 0 then err = (-got):syserr;
```

?

cassella:
How about moving the `arg[i] =` line after the `got < 0` break too?

I'm hesitating because any direct test of this change will lock in the current behavior of returning false for a non-newline-terminated line.  Would it make sense to change it to return true in that case, and leave the EOF or other error to be re-encountered on the next readline()?  E.g., also change the last line above to

```chapel
if i == start && got < 0 then err = (-got):syserr;
```

Then the readline that hits EOF mid-line will leave err as `ENOERR`, and return true with the correct numBytes.  The next readline() will still hit EOF, but will have i == start so return false with `numBytes == 0`.

The off-by-one suggests that no caller is actually using it the way it's written -- even for files that end with newlines, anyone who sees the false return and checks numBytes today would see `numBytes==1` and consume the EOF character.  So maybe it should be changed to match the way `copyFiles()` was using it, and all other the users must be using it now?  (And matching the C `readline()` AFAICT). 

At least, EOF would repeat on re-readline().  I haven't thought about whether this would be true for all other errors.

(I haven't actually checked that no callers of readline in test/ or modules/ are consuming `outBytes` bytes when the return value is false.  I'm just asserting it must be so.  :) )

I don't think I'll have time in the near future to write an issue more coherent than this comment, let alone a future or PR.

cassella:
Now that you mention it, I think there *is* a null byte problem after all.  :)

```chapel
    if got < 0 then err = (-got):syserr;
    got_copy = got;
    this.unlock();
  }

  if !err && got_copy != 0 {
    return true;
  } else if err == EEOF || got_copy == 0 {
    return false;
```

If `got` happened to be a null byte on the iteration that incremented `i` to `maxIdx`, `got_copy` will also be 0 and it'll spuriously return false.

PR #9649, opened 2018-05-29
Represent if-expressions with IfExpr node instead of if-functions

### LINKED ISSUE #8189: Resolution failure for function with if-expr as default value for an argument

### LINKED ISSUE #8479: Should we change the conditional expression implementation to use a separate AST node?
Today, we represent if expressions (`var x = if a then b else c;`) in the compiler as a nested function.  However, this causes problems:
- In the case of if expressions used as the default value of an argument when no type is specified (see #8189), the function call is copied to the type of the argument, but the definition is not.
- In the case of field defaults when default initializers are involved, the function is defined to utilize fields, but we want those fields to be references to the necessary arguments to the default initializer (since the fields are initialized when the default initializer is called).  However, the function definition has been entirely separated from the call to it, so it can't be updated when this is happening.
- #8464 demonstrates that the generated code may cause the return of argument pointers when used with the internal fields of the range type.

We believe that representing if expressions as a separate AST type would help alleviate these problems.  It would allow the representation to remain within the scope in which it was defined, enabling the necessary analysis and code transformations that would allow these cases to succeed (or at least allow the bug fixes that the current implementation thwarts).

benharsh:
This PR updates the compiler to use the IfExpr node from parse() through normalize(), at which point it is lowered to a CondStmt. A variety of changes were required to make this possible:
- do not lower && or || operators immediately after parsing
- handle IfExprs during scopeResolve
- Lower && and || operators to IfExprs at the start of normalization, pre-order
- special handling of IfExprs for initializers
- CondStmts lowered from IfExprs are special-cased in type resolution, and are handled like functions that return different types from different branches.

As a result, FLAG_IF_EXPR_FN is removed. A handful of futures have been retired as well.

Resolves #8189 
Resolves #8479 

Testing:
- [x] local + futures
- [x] gasnet + futures
- [x] memleaks
- [x] release/examples/, expressions/if-expr, arrays-ferguson/if-expr, expressions/ferguson/if-expr
  - [x] baseline
  - [x] valgrind


compiler/resolution/functionResolution.cpp
```diff
@@ -5356,6 +5363,12 @@ static bool moveSupportsUnresolvedFunctionReturn(CallExpr* call) {
   return retval;
 }
 
+static bool isIfExprResult(CallExpr* call) {
```
mppf: I think the LHS is "the if expr result" but call is something producing the if-expr.
So, I'm asking you to rename this function, to perhaps "moveToIfExprResult" or "setsIfExprResult" or something along those lines. Additionally, I think it should check that call is PRIM_MOVE or PRIM_ASSIGN.

Alternatively, you could keep the current name and just have it take in LHS of the move, since it's only called in resolveMove.

compiler/resolution/resolveFunction.cpp
```diff
@@ -984,6 +984,109 @@ static void computeReturnTypeParamVectors(BaseAST*      ast,
...
+
+    Symbol* thenSym = retSymbols.v[0];
+    Symbol* elseSym = retSymbols.v[1];
+
+    bool thenTypeVar = thenSym->hasFlag(FLAG_TYPE_VARIABLE);
+    bool elseTypeVar = elseSym->hasFlag(FLAG_TYPE_VARIABLE);
+
+    // If only one branch returns a reference, that branch needs to copy the
+    // result so that we can always free the result.
+    if (isReferenceType(thenType) != isReferenceType(elseType)) {
```
mppf: Is there a reason this only works with reference types and not also QUAL_REF? I thought we had another function (like isReference maybe?) that handled the Qualifier case as well?

compiler/resolution/resolveFunction.cpp
```diff
@@ -984,6 +984,109 @@ static void computeReturnTypeParamVectors(BaseAST*      ast,
...
+        retType = thenType;
+      } else if (canDispatch(thenType, thenSym, elseType, fn, &promote) &&
+                 promote == false) {
+        retType = elseType;
+      }
+    }
+
+    // For tuples, generally do not allow a tuple to contain a reference
+    // when it is returned
+    if (retType != NULL && retType->symbol->hasFlag(FLAG_TUPLE) == true) {
```
mppf: I don't think we need this tuple case here - it should be handled by normal variable initialization?

Try removing it and check what the type of x is in:
``` chapel
var A:[1..10] int;
var B:[1..10] int;
var x = if something then (A,B) else (B,A);
```

x should have a tuple type storing the arrays by value, but the if-expr can return a tuple type storing the arrays by reference.

compiler/passes/normalize.cpp
```diff
@@ -857,6 +864,111 @@ static void insertUseForExplicitModuleCalls() {
...
+{
+  public:
+    LowerIfExprVisitor() { }
+    virtual ~LowerIfExprVisitor() { }
+
+    virtual bool enterIfExpr(IfExpr* node);
+};
+
+bool LowerIfExprVisitor::enterIfExpr(IfExpr* ife) {
+  if (isAlive(ife) == false) return false;
```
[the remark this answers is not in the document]
benharsh: Sorta, but I don't feel that strongly about it and would be willing to change it. Sometimes I find that with ``!something()`` I can easily overlook the ``!``.

compiler/passes/normalize.cpp
```diff
@@ -857,6 +864,111 @@ static void insertUseForExplicitModuleCalls() {
...
+  result->addFlag(FLAG_EXPR_TEMP);
+  result->addFlag(FLAG_IF_EXPR_RESULT);
+
+  // Don't auto-destroy local result if returning from a branch of a parent
+  // if-expression.
+  bool isNestedIfExpr = false;
+  if (CallExpr* call = toCallExpr(ife->getStmtExpr()->next)) {
+    if (call->isPrimitive(PRIM_MOVE)) {
+      if (SymExpr* se = toSymExpr(call->get(1))) {
+        isNestedIfExpr = se->symbol()->hasFlag(FLAG_IF_EXPR_RESULT);
```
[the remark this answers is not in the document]
benharsh: Good point, I'll see if I can rewrite to use ``exitIfExpr``.

compiler/passes/normalize.cpp
```diff
@@ -1633,7 +1745,9 @@ static void normalizeTypeAlias(DefExpr* defExpr) {
   INT_ASSERT(type == NULL);
   INT_ASSERT(init != NULL);
 
-  defExpr->insertAfter(new CallExpr(PRIM_MOVE, var, init->copy()));
+  Expr* initCopy = init->copy();
+
+  defExpr->insertAfter(new CallExpr(PRIM_MOVE, var, initCopy));
```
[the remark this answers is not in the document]
benharsh: I think it was left in place after some other changes, I'll undo these.

compiler/passes/scopeResolve.cpp
```diff
@@ -467,6 +483,15 @@ static void scopeResolve(const AList& alist, ResolveScope* scope) {
         }
       }
 
+      // Look for IfExprs
+      if (def->init != NULL) {
```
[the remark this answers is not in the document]
benharsh: Hmm, the compiler doesn't emit an error for that pattern. I'll investigate.

compiler/resolution/resolveFunction.cpp
benharsh: I just copied what ``resolveReturnTypeAndYieldedType`` did, and it uses ``isReferenceType``. For some reason I have it in my head that there are issues with ``isRef`` during resolution.

compiler/passes/normalize.cpp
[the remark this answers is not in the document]
mppf: Don't sweat it, but maybe we can leave out something == true. The one thing I really don't like is `!!x` which I've seen before in LLVM codebase; apparently it's considered a reasonable way to check a pointer is non-null. (Just not by me).

compiler/resolution/resolveFunction.cpp
mppf: There are, but it's mostly that resolution doesn't use QUAL_REF. Anyway you can just add a comment about how it should eventually be updated to handle the qualifiers.

compiler/passes/scopeResolve.cpp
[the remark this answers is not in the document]
benharsh: Perhaps unsurprisingly, your example doesn't work correctly because normalize is making decisions based on x's type (known to be a primitive, int). If we leave off the type expression then we correctly catch the use-before-def.

compiler/resolution/resolveFunction.cpp
benharsh: I originally omitted this case, but ran into a testing failure and so copied it from the resolve-return-type function.

mppf: That's odd, I wonder if something else is wrong / the test needs updating. It's at least worth trying the example I showed above and looking at the --log output to see if we need to make an issue about it.

mppf: The case in which a program would behave differently is like this:
``` chapel
f( if something then (A,B) else (B,A) );
```

benharsh: I have opened #9684 for further discussion of this issue.

Issue #9306, opened 2018-04-23
Should enums support `<`, `<=`, `>`, `>=` by default?


bradcray:
Issue #8794 proposed that the compiler should only create `==` and `!=` comparisons for enums by default.  It could be argued that it should also create `>`, `>=`, `<`, and `<=` since an enumerated type has a well-defined order.

One challenge to implementing such comparisons is that if an enum's underlying integer values were not statically known to be monotonically increasing, these couldn't be implemented simply by casting to integers and calling the integer comparisons.  Instead, we'd need to have some sort of "enum to order" function that would convert an enum to an ordinal value that we could do the comparison on.  This is functionality that would likely be required to support enum ranges as well (see #9117).  Presumably, with additional effort the compiler could distinguish between cases that need this extra conversion step and those that don't in order to implement the monotonic case more efficiently.

[edited to capture further discussion below]:

In discussions, some argued that they would want these operators to compare the underlying integer values rather than the declaration order of the enums.  However, we ended up agreeing/acquiescing that respecting the declaration order was the right thing for enums based on the following argument:

1) In most cases, the two orders are the same, either because the user has either not given their enums integer values, or because they've given them monotonically increasing integer values.

2) When the orders are not the same, we'll argue that the enumeration order is preferable because:
  a) comparisons against the enum values can always be obtained by casting to integers or overloading these operators themselves if that's what's truly wanted
  b) if users want a specific order, they can reorder the enum to reflect any order; but if the integer values encode certain properties (say by using a bitmask that doesn't result in monotonically increasing values), they may not be able to give them different values.

bradcray:
One argument against supporting such comparisons is that, intuitively for many enums, there's no reason to believe that "red" should be less than "blue" (say) just because they happened to be listed in a given order.

mppf:
I find the point about ranges reasonably compelling but am unconvinced by the argument that the declaration order is somehow more natural / better than the integer value order.

mppf:
But, it's also unclear to me why we would want to support enum ranges.

bradcray:
Capturing some cases where codes wanted to use comparisons on enums, for future reference:

* Our `BLAS` module used comparisons in its `gemm()` routine to see what opA was passed in
* Some programs that use the `sorted()` iterator (the "sorterator") on associative domains required it in order to apply the sort to the enumerated indices

bradcray:
[...]

PR #9670, opened 2018-05-29
Add support for enum comparison functions (<, <=, >, >=)

bradcray:
This change adds support for `<`, `<=`, `>`, `>=` comparisons between values of a given enum as proposed in issue #9306, where these operators are defined to respect the declaration order of the enum's symbols (i.e., not their integer values, though the two orders are typically the same in most cases).  The operators are defined both for param combinations of enums and non-.  

The implementation is based on the compiler defining two internal routines that map between enums and integers:

* `chpl__enumToOrder(e: enumerated): int` converts an enumerated value to its order (0, 1, 2, 3, ...) and is also supported in a param->param version.
* `chpl__orderToEnum(i: integral, type et: enumerated): et` which does the reverse, converting a 0-based integer order into the `i`th value of enumerated type `et`.

While enabling the `param` versions of these comparison operators, I realized that the previous param versions of `==` and `!=` were still incorrectly using the underlying integer values of the enums rather than testing for comparisons on the enums themselves.  In this effort, I've implemented a correct primitive `==` comparison for enums in the compiler and have implemented all other comparison functions as module code that performs comparisons on the enums' `chpl__enumToOrder()` values.

New tests include:

* test/types/enum/operations/comparisons/enumComparisons*.chpl: A new pair of tests that ensure all comparison operators for enums work for both param and non-param versions independent of the enums' numerical values.

* test/types/enum/operations/comparisons/enumComparisonsMixed.chpl: Makes sure that we don't permit enum comparisons between enums of different types.

* test/types/enum/operations/order/enumOrder*.chpl: Checks that the enum<->order functions work for param and non-param values as expected.  The three versions of this are intended to exercise different implementations that may exist as we optimize the implementations of these functions for enums with consecutive values (see the first follow-on work item below).

Chapel code-related changes are as follows:

* modules/packages/BLAS.chpl: Updated to avoid using the cast, as it did prior to #9192 now that comparisons are supported again.

* test/arrays/diten/enumArrayMethods.chpl: Updated to remove the explicit `<` overload, as it was prior to #9192.  Updated the .good file to reflect the new sorting order based on enumeration order, not integer value.

* test/associative/diten/enumArrays.chpl: See previous entry

* test/types/enum/sungeun/test_foldEnumOp.chpl: See previous entries

* test/types/enum/diten/paramenums*.chpl: Updated to move cases that are now working again from paramenums2.chpl to paramenums.chpl.  Also updated the error message which is now more specific.

Follow-on work:
* the current implementation uses a switch table to do the conversions in either direction; in the event that the enum symbols are not given integer values (or just the first one is), we could implement these functions using some math and casting and should do so for the best performance.
* ultimately, we may want to make these two routines public in which case they should probably be converted into methods on the enumerated type (for order to enum) and its values (for enum to order)


modules/internal/ChapelBase.chpl
```diff
@@ -153,12 +153,17 @@ module ChapelBase {
   inline proc ==(param a: bool, param b: bool) param return __primitive("==", a, b);
   inline proc ==(param a: int(?w), param b: int(w)) param return __primitive("==", a, b);
   inline proc ==(param a: uint(?w), param b: uint(w)) param return __primitive("==", a, b);
+  //
   inline proc ==(param a: enumerated, param b: enumerated) param where (a.type == b.type) return __primitive("==", a, b);
+  //
+  // NOTE: Only '==' is implemented in the compiler on param enums
+  // (and only only on param enums of the same type), so all other
```
vasslitvinov: "only" repeats twice.
Also, I did not see where the compiler checks whether the two enums are of the same type.
Adjust comment? Adjust the compiler?

modules/internal/ChapelBase.chpl
```diff
@@ -153,12 +153,17 @@ module ChapelBase {
...
+  // param enum routines need to be defined as module code.
 
   inline proc !=(param a: bool, param b: bool) param return __primitive("!=", a, b);
   inline proc !=(param a: int(?w), param b: int(w)) param return __primitive("!=", a, b);
   inline proc !=(param a: uint(?w), param b: uint(w)) param return __primitive("!=", a, b);
-  inline proc !=(param a: enumerated, param b: enumerated) param where (a.type == b.type) return __primitive("!=", a, b);
+  inline proc !=(param a: enumerated, param b: enumerated) param where (a.type == b.type) return __primitive("!=", chpl__enumToOrder(a), chpl__enumToOrder(b));
```
vasslitvinov: I would prefer to avoid using chpl__enumToOrder() here,
because the == case does not use those.
`!__primitive("==", a, b)` seems appropriate.

bradcray: The `==` case doesn't use them because it's implemented in the compiler as noted in the comment that follows it.  This case is symmetrical with all of the other comparison operators (which can't be implemented in terms of "=="), which seems preferable to me to introducing asymmetry just for this case.

vasslitvinov: OK, that's fine, then.

compiler/resolution/postFold.cpp
```diff
@@ -64,22 +64,30 @@ static Expr* postFoldSymExpr(SymExpr* symExpr);
...
+        }                                                               \
+        if (!isEnumSymbol(rhsSym)) {                                    \
+          INT_FATAL("Trying to do a mixed enum non-enum primitive");    \
+        }                                                               \
+        Immediate enumcomp;                                             \
+        enumcomp = false;                                               \
+        if (lhsSym == rhsSym) {                                         \
+          enumcomp = true;                                              \
+        }                                                               \
+        retval = new SymExpr(new_ImmediateSymbol(&enumcomp));           \
```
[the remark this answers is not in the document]
bradcray: Seemingly not, but:

```chapel
	Immediate enumcomp;                                             \
        enumcomp = (lhsSym == rhsSym);                                  \
```

works (the longer version above dated from a point when the true vs. false cases were further apart).

modules/internal/ChapelBase.chpl
bradcray: The compiler doesn't check, but assumes they're the same.  I've updated the comment to clarify that the where clause here is what's keeping things safe.

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -942,6 +944,131 @@ static void build_enum_cast_function(EnumType* et) {
 }
 
 
+static void build_enum_order_functions(EnumType* et) {
```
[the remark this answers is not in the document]
bradcray: Good idea about combining the two versions into one function.  See new commits.

bradcray:
@vasslitvinov : Let me know when you've had a chance to re-review the changes you requested.

vasslitvinov:
@bradcray - the changes look good!

PR #9673, opened 2018-05-30
Add "babel<2.6.0" constraint to chpldoc-venv Pip requirements

awallace-cray:
Cray-internal Cray module builds broke today 5/29.

chpldoc no longer builds, because its Python virtualenv no longer builds,
because Babel (a PyPI dependency) updated itself to version 2.6.0 today
on PyPI, and Babel 2.6.0 is no longer compatible with Python < 2.7.
(Babel version yesterday was 2.5.3)

As long as Chapel must build RPMs for older Cray-XE and Cray-XC CLE-5
systems, which are SLES-11 based, Chapel must be able to use Python 2.6.9.


awallace-cray:
Post-merge review invited. 

This change (constrain babel version < 2.6.0, the latest version on PyPI this morning) is only needed for (old) Cray module builds. AFAIK, everything else we do in Jenkins already expects Python 2.7 for some other reason. Nevertheless, this Pip constraint will be applied every time on every platform, not just Cray module builds for CLE-5 Cray machines. 

IMO, putting in this change was the right thing to do right now, because we need the Cray module builds fixed. However, I invite others to consider if this change should just stay in as-is until such time as we drop all support for Python<2.7, or if we should make some more sophisticated kind of change that (perhaps) only applies when Python<2.7 but has no effect on "normal" build environments.  Maybe, add this "babel<2.6.0" constraint dynamically at build time, for example. 

OTOH, we probably should just drop support for CLE-5 Crays, or adopt Python-3, or both.

awallace-cray:
Invited reviewers: this is a post-merge review. It's optional. Please see above comment. Thanks.

PR #9677, opened 2018-05-30
Add flag to specify classes are unmanaged by default 

### LINKED ISSUE #9289: delete-free subtask: compiler flag to toggle default class type meaning
As a Chapel user migrating code to the new class ownership system, it would help significantly in such migrations if I can toggle between the old and new system within the same version of the compiler. This will allow me to more easily convert code, since I can verify it works with that version of the compiler before starting migration.

See design discussion here: https://github.com/chapel-lang/chapel/issues/9278#issuecomment-382852278

This flag will help with migrating user code because the delete-free language changes can be toggled on and off separately from the other language changes.

Acceptance criteria:
New compiler flag allows toggling between class types / new default to unmanaged and default to borrow.

mppf:
Adds `--default-unmanaged` as a (currently) internal/developer flag.
It should get more testing somehow if it's to be a user-facing flag.

Resolves #9289.

Passed full local testing.


mppf:
@vasslitvinov - could you have a look at this one after PR #9612 ? thanks

mppf:
@vasslitvinov - this PR is ready for your review, thanks!

compiler/passes/scopeResolve.cpp
```diff
@@ -162,6 +165,36 @@ void scopeResolve() {
...
+          Expr* result = se;
+          if (fWarnUnstable || fDefaultUnmanaged)
+            result = handleUnstableClassType(se);
+          // Amend _this->type
+          fn->_this->type = result->typeInfo();
+        } else {
+          INT_ASSERT(0);
+        }
+      }
+    }
```
vasslitvinov: Please move the above into a separate function over 'fn', ex. adjustMethodsForDefaultUnmanaged(FnSymbol* fn).

vasslitvinov: While there, how about move 'fDefaultUnmanaged' to be the first expression in the initial if-condition ?
That IMO makes it easier to read and faster to execute in the common case.

compiler/passes/scopeResolve.cpp
```diff
@@ -725,43 +758,80 @@ void resolveUnresolvedSymExprs(BaseAST* inAst) {
...
+          call->isPrimitive(PRIM_TO_BORROWED_CLASS) ||
+          call->isNamed("_to_borrowed") ||
+          call->isNamed("_to_unmanaged") ||
+          call->isNamed("_owned") ||
+          call->isNamed("_shared") ||
+          call->isNamed("Owned") ||
+          call->isNamed("Shared"));
+}
+
+// Returns the expr resulting, either se or a call containing it
```
vasslitvinov: What if this function returns the resulting Type* directly?

Either way, I'd change the above comment to clarify the intention. For example:

"Handle the case where se->symbol() is "unstable" i.e. an undecorated class type. With --warn-unstable, issue a warning. With --default-unmanaged, wrap it in a _toUnmanaged call. Return either se or a call _toUnmanaged(se)".

modules/dists/SparseBlockDist.chpl
```diff
@@ -125,21 +137,15 @@ class SparseBlockDom: BaseSparseDomImpl {
...
   proc bulkAdd_help(inds: [] index(rank,idxType),
       dataSorted=false, isUnique=false) {
     use Sort;
     use Search;
 
     // without _new_, record functions throw null deref
-    var comp = new TargetLocaleComparator();
+    var comp = new TargetLocaleComparator(rank=rank, idxType=idxType,
+                                          sparseLayoutType=sparseLayoutType,
+                                          dist=dist);
```
vasslitvinov: If the changes to this file in this PR are intentional, please explain them briefly in the PR message.

mppf:
@vasslitvinov - I addressed your feedback and moved the SparseBlock change to #9739. I'm planning to merge once another testing run completes successfully.

vasslitvinov:
@mppf - sounds good.

Issue #9615, opened 2018-05-23
Use BTE RDMA for large puts/gets under ugni


ronawho:
Currently we always use Fast Memory Access (FMA) for puts/gets under ugni. The block transfer engine (BTE) or simply called RDMA in some ugni documentation is supposed to be better for "large" transfers. 

From the [aries docs](https://www.cray.com/sites/default/files/resources/CrayXCNetwork.pdf):
"""
In general, FMA is used for small transfers and BTE for large. FMA transfers are lower latency but require the involvement of the CPU. BTE transfers take longer to start and impose a higher load on the host interface but once initiated can transfer large amounts of data (up to 4 GB) without CPU involvement. 
"""

We probably want to do some tuning to figure out what the right threshold is for us, but here are some points of comparison: 
 -  cray-mpich 7.6.2 cuts over to BTE at 1024 bytes. (from MPICH_GNI_RDMA_THRESHOLD in `man mpi` on edison)
 - GASNet 1.30.0 cuts over to BTE at 4096 bytes (from GASNET_GNI_GET_FMA_RDMA_CUTOVER in aries readme)

I expect this to help with benchmark that do large transfers (such as ISx). 

The ISx section in https://chapel-lang.org/releaseNotes/1.17/05-benchmarks-opts.pdf may have a little more info on this.

ronawho:
This significantly improved sustained bandwidth for [large gets/puts](https://chapel-lang.org/perf/16-node-xc/?startdate=2018/06/02&enddate=2018/06/05&graphs=largearraygetperformance,largearrayputperformance):

![image](https://user-images.githubusercontent.com/1588337/41005645-3fbcbafa-68d3-11e8-9e11-a935303eb0f5.png)

![image](https://user-images.githubusercontent.com/1588337/41005672-5677acdc-68d3-11e8-8832-fbb7e0fac4a2.png)

Which speeds up ISx, particularly at scale:

![isx-time](https://user-images.githubusercontent.com/1588337/40895518-db88b2d8-6764-11e8-82a7-bae0aefe9297.png)

PR #9678, opened 2018-05-30
Add a low-level array transfer performance test

ronawho:
This adds perf testing for low-ish level communication: large gets/puts that
are initiated as part of an array bulk transfer. This does a point-to-point
transfer for small (512B, 4096B) and large (1MB, 1GB, 1/4 mem) array transfers.

This test will be useful to determine the threshold for BTE transfers and also
shows that ugni sustained bandwidth is lagging because it does not use BTE for
large transfers yet. For 1MB and larger transfers ugni only achieves ~3 GB/s,
while gn-aries achieves ~8 GB/s. The sustained bandwidth for aries is somewhere
between 8-10 GB/s for unidirectional point-to-point puts/gets, so 8 GB/s is
probably pretty close to optimal. (Our bulk transfer includes some operations
and comm beyond the large get/put, so we shouldn't expect to quite see the
hardware limit in this test.)

Below is a table of results for the put version comparing ugni to gn-aries at
different transfer sizes. ugni is slightly ahead for small transfers, but is
over 2x behind for larger transfers that test sustained bandwidth. ugni maxes
out at around 3.7 GB/s for puts (2.7 for gets) and gn-aries maxes out at around
8.5 GB/s for both puts/gets.

| size    | ugni-put   | gn-put    |
| ------- | ---------- | --------- |
|  512B   |   35 MB/s  |   28 MB/s |
| 4096B   |  290 MB/s  |  228 MB/s |
|    1MB  | 3650 MB/s  | 7956 MB/s |
|    1GB  | 3707 MB/s  | 8552 MB/s |
| 1/4 mem | 3799 MB/s  | 8696 MB/s |

Related to https://github.com/chapel-lang/chapel/issues/8305 and https://github.com/chapel-lang/chapel/issues/9615


ronawho:
This test is an amalgamation of performance/comm/low-level/many-to-one and runtime/configMatters/comm/bigTransfer

test/performance/comm/low-level/arrayTransfer.ml-execopts
```diff
@@ -0,0 +1,11 @@
+--xferMem=512   --runSecs=1 --op=opPut # array-put-512B
+--xferMem=4096  --runSecs=1 --op=opPut # array-put-4096B
+--xferMB=1      --runSecs=1 --op=opPut # array-put-1M
+--xferMB=1024   --runSecs=1 --op=opPut # array-put-1G
+--memFraction=4 --runSecs=1 --op=opPut # array-put-FR4
```
gbtitus: Transferring 1/4 of memory in 1 second seems unachievable in practice.  If that's so then the runs across different transfer sizes will vary in length and things won't be as comparable as we'd like.  That may not matter much, though, since only the 1/4 mem ones will be out of range.  The rest should be able to get done in 1 second.

ronawho: Yeah, you're right. It doesn't actually finish in a second, this really just limits it to 1 trial at least on large-memory cray compute nodes. On a 128GB compute node it takes ~4 seconds. That said I have no problem with the elapsed time being different between versions since we're graphing the rate.

Issue #9680, opened 2018-05-30
Sporadic hwloc warnings under travis


ronawho:
We've had a bunch of sporadic hwloc failures under travis-ci testing recently. Here's a few: 
 - https://travis-ci.org/chapel-lang/chapel/jobs/380407274
 - https://travis-ci.org/chapel-lang/chapel/jobs/380929447
 - https://travis-ci.org/chapel-lang/chapel/jobs/382332502
 - https://travis-ci.org/chapel-lang/chapel/jobs/385737426
 - https://travis-ci.org/chapel-lang/chapel/jobs/365382846
 - https://travis-ci.org/chapel-lang/chapel/jobs/368834830
 - https://travis-ci.org/chapel-lang/chapel/jobs/371306015
 - https://travis-ci.org/chapel-lang/chapel/jobs/373095113
 - https://travis-ci.org/chapel-lang/chapel/jobs/379426258
 - https://travis-ci.org/chapel-lang/chapel/jobs/379460378
 - https://travis-ci.org/chapel-lang/chapel/jobs/385824542
 - ... many others ...

We get a warning from hwloc that looks something like:
```
****************************************************************************
* hwloc 1.11.10 has encountered what looks like an error from the operating system.
*
* L3 (cpuset 0x00001000,0x02100002) intersects with NUMANode (P#0 cpuset 0x00001111,0x11111111 nodeset 0x00000001) without inclusion!
* Error occurred in topology.c line 1088
*
* The following FAQ entry in the hwloc documentation may help:
*   What should I do when hwloc reports "operating system" warnings?
* Otherwise please report this error message to the hwloc user's mailing list,
* along with the files generated by the hwloc-gather-topology script.
****************************************************************************
```

All the failed build logs seem to be using an older kernel (4.4.0), and successful builds seem to be using a newer kernel (4.14.12). I looked back through a couple hundred builds over the last 4 months and it looks like all the ones that didn't fail were using a 4.14 kernel. 

Maybe related -- we upgraded hwloc version to 1.11.10 on April 2 (https://github.com/chapel-lang/chapel/pull/8981). I haven't been able to find a build before that merge that failed, but I also haven't been able to find a build before that that used 4.4.0 (and I'm getting bored looking through logs)

https://www.open-mpi.org/projects/hwloc/doc/v1.11.10/a00293.php#faq_os_error has more info about the error/warning from hwloc. It also suggests: 

"""
[...]

PR #9681, opened 2018-05-30
Try to quiet hwloc warnings under travis testing

ronawho:
Set `HWLOC_HIDE_ERRORS=1` during travis testing to hopefully avoid the sporadic
hwloc warnings we've been seeing recently. When travis testing happens to run
on a 4.4 kernel we've been getting errors about bad L3 cache detection. We
don't care about that for travis testing (and we already disable affinity under
travis because of weirdness with trying to pin under a VM/container.)

This isn't an ideal solution (HWLOC_HIDE_ERRORS=1 sounds extreme/bad), but I
don't think this is worth digging into too much.

This closes https://github.com/chapel-lang/chapel/issues/9680

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ compiler:
 
 matrix:
   include:
-    - env: NIGHTLY_TEST_SETTINGS=true QTHREAD_AFFINITY=no CHPL_SMOKE_SKIP_DOC=true TEST_COMMAND="./util/buildRelease/smokeTest"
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus any severe objections to this? I don't love it, but I don't think it's worth digging into any more.

ronawho:
We can't get much useful info to report to the hwloc team because the failure is inside a travis instance that we can't directly access. Note that we're only setting this env var for travis testing. Our regular smoke-testing (and certainly our performance testing) does not set this. Given that the failure is only turning up on travis instances running an older kernel, I'm not too worried about this biting us later.

ronawho:
(Ideally I think we would try to replicate the travis environment and see if we can repro, but I'm not sure that's worth the time)

PR #9683, opened 2018-05-30
Use BTE RDMA for large transfers under ugni

ronawho:
Bring ISx performance on par with the reference version by using uGNI's block
transfer engine (BTE) to significantly increase sustained bandwidth.

This enables uGNI's block transfer engine for puts/gets larger than 4K. This
significantly increases the performance of the arrayTransfer microbenchmark,
allowing point-to-point puts/gets over 1M to achieve over 8 GB/s, which is very
close to maximum hardware bandwidth. Using BTE significantly improves the
all-to-all bucket exchange step in ISx, which closes the main performance gap
between Chapel and the reference SHMEM version. At 256 locales, ISx is now ~30%
faster, which brings it within ~5% of the reference version. This also seems to
stabilize ISx timings, particularly on busy systems.

Below is a table comparing the PUT version of the arrayTransfer benchmark at
different sizes. It compares ugni with FMA vs. ugni with BTE for large
transfers vs. gn-aries. gn-aries appears to be slightly ahead of ugni for large
transfers still, but this is just because the entire array under gn-aries is on
numa-domain 0, which is closer to the NIC. ugni permits first-touch, so half
the array is on numa-domain 1. Putting the entire array on numa-domain 0 by
wrapping the benchmark with a `serial` block lets ugni achieve 8700 MB/s as
well.

| size    | FMA-ugni   | BTE-ugni   | gn-aries  |
| ------- | ---------- | ---------- | --------- |
|  512B   |   35 MB/s  |   35 MB/s  |   30 MB/s |
| 4096B   |  290 MB/s  |  295 MB/s  |  230 MB/s |
|    1MB  | 3650 MB/s  | 7500 MB/s  | 7900 MB/s |
|    1GB  | 3707 MB/s  | 8200 MB/s  | 8500 MB/s |
| 1/4 mem | 3799 MB/s  | 8400 MB/s  | 8700 MB/s |

And the new ISx results:

isx-time

This closes https://github.com/chapel-lang/chapel/issues/9615

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -5833,6 +5833,7 @@ chpl_comm_nb_handle_t chpl_comm_get_nb(void* addr, c_nodeid_t locale,
   nb_desc_idx_t          nbdi;
   nb_desc_t*             nbdp;
   gni_post_descriptor_t* post_desc;
+  chpl_bool              do_rdma = false;
 
   DBG_P_LP(DBGF_IFACE|DBGF_GETPUT, "IFACE chpl_comm_get_nb(%p, %d, %p, %zd)",
            addr, (int) locale, raddr, size);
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus I don't plan to merge this until Sunday or Monday (want to give the new arrayTransfer test a couple days to run), but I think this is ready for review.

I still need to do XE testing, and do some ISx scalability testing with this latest variation, though I don't expect any surprised for ISx scaling.

PR #9694, opened 2018-06-01
Create an array/domain/distribution class for external arrays

lydia-duncan:
Adds a distribution/domain/array implementation in support of arrays
obtained from outside of Chapel, where their contents are not owned
by Chapel itself.  Adds a function to create an instance of this array
from a c_ptr and the number of elements.

This is part one of supporting the exporting of functions that take
an array.  The next step is allowing the compiler to generate
code similar to what I have written in my test files when it sees an
exported function that takes an array.

Note that this doesn't account for privatization or locality yet.  The former
is because I'm not quite sure how to test that, the latter due to wanting to
get the basic functionality working and reviewed.

Adds three tests of the functionality.  One example using ints, one example
using reals (demonstrating that this enables some type generality from the
get-go, which is very exciting), and one demonstrating the behavior when
a copy of the external array wrapper is made.

Testing:
- [x] fresh checkout of new tests
- [x] new tests with valgrindexe
- [x] full local paratest (in progress)


lydia-duncan:
@mppf - what are your thoughts on the two strategies?  Should I provide both or just one?  If just one, which one do you favor?

I'm also curious if you have thoughts about syntactic sugar for how to create a domain of this external type (which would make buildArrayWith potentially less weird?)

modules/internal/ExternalArray.chpl
```diff
@@ -0,0 +1,284 @@
...
+    // Prolly want the privatization stuff eventually, but I don't need it right
+    // now.
+  }
+
+  class ArrayViewExternArr: BaseArr {
+    type eltType;
+
+    const dom;
+    
+    const _ArrInstance;
```
mppf: Is this intended to be a C ptr? or something else?
Should it be a "local field" ddata the way that DefaultRectangularArr uses?
If it's a C ptr, are we going to have problems multi-locale?

modules/internal/ExternalArray.chpl
```diff
@@ -0,0 +1,284 @@
...
+    // Do I want the locality-oriented queries? (see
+    // ArrayViewSlice.chpl:225-230)
+    // What about the privatization?
+    // doiCanBulkTransferRankChange? doiBulkTransferFromKnown?
+    // doiBulkTransferToKnown?
+
+    proc dsiDestroyArr() {
+      if (_owned) {
+        c_free(_ArrInstance);
+      }
```
mppf: I thought the idea was that this array never owned the data? I don't have a problem with this though if it's useful to you.

modules/internal/ExternalArray.chpl
```diff
@@ -0,0 +1,284 @@
...
+        return chpl_getPrivatizedCopy(dom.type, _DomPid);
+        } else {*/
+        return dom;
+        //}
+    }
+
+    // proc dsiReallocate(d: domain) is not supported, so don't override.
+    // proc _resize(length: int, old_map) is not supported, don't override.
+  }
+}
```
mppf: Let's add a newline at the end to keep from this red zero thingy.

modules/internal/ExternalArray.chpl
```diff
@@ -0,0 +1,284 @@
...
+      chpl_assignDomainWithGetSetIndices(this, rhs);
+    }
+
+    // What about _getActualDom?  dsiDestroyDom?
+
+    // Prolly want the privatization stuff eventually, but I don't need it right
+    // now.
+  }
+
+  class ArrayViewExternArr: BaseArr {
```
mppf: If it has the option of "owning" the data, I wouldn't call it ArrayViewExternArr. We could call it just ExternArr. Or we could make it not allowed to "own" the data.

modules/internal/ChapelArray.chpl
```diff
@@ -1378,6 +1392,24 @@ module ChapelArray {
...
+      pragma "dont disable remote value forwarding"
+      proc help() {
+        _value.add_arr(x);
+      }
+      help();
+
+      chpl_incRefCountsForDomainsInArrayEltTypes(x, x.eltType);
+
+      return _newArray(x);
+    }
```
mppf: I think you're trying to ask if buildArrayWith or _getArray make a better user-facing interface? I wouldn't call it _getArray but it makes sense to me that you wouldn't need to create the domain before creating the array, in cases like this, since the domain is always 0..#size anyway.

modules/internal/ExternalArray.chpl
lydia-duncan: It's necessary for the initCopy cases I was talking to you about earlier.  If we have to create new memory for the copy, then the copy should be responsible for cleaning that memory up.  But I could be misunderstanding something.

mppf:
> @mppf - what are your thoughts on the two strategies (_getArray versus buildArrayWith)? Should I provide both or just one? If just one, which one do you favor?
>
> I'm also curious if you have thoughts about syntactic sugar for how to create a domain of this external type (which would make buildArrayWith potentially less weird?)

I wouldn't keep either name, as a user-facing function. Instead it'd be a function like 'makeArrayFromPointer' or something like that. I don't see a fundamental reason why forcing users to create the domain first is useful in this context.

modules/internal/ExternalArray.chpl
lydia-duncan: That's an open question in my mind.  I've been working with C ptrs for now because that seems more like a type I would get from an outside function, though _ddata is very similar.  Would _ddata fit into the tests I defined, do you think?  I'm not sure I fully understand the consequences of one choice over the other.

modules/internal/ChapelArray.chpl
```diff
@@ -166,6 +166,7 @@ module ChapelArray {
   use ArrayViewSlice;
   use ArrayViewRankChange;
   use ArrayViewReindex;
+  use ExternalArray;
```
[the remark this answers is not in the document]
lydia-duncan: I was mostly keeping it close to the other _getArray function since they had a similar origin, but I can make it local to the ExternalArray module (and insert uses when compiling with --library mode)

modules/internal/ExternalArray.chpl
mppf: Right, but initCopy can return a different type, so you could consider having it return a DefaultRectangular array.

lydia-duncan: The current implementation of initCopy for arrays has one branch for a shallow copy and one branch for a deep copy.  It seems like the shallow copy should give us the same type, and when it was accidentally giving us a DefaultRectangular for the deep copy branch I ran into compiler resolution errors because it didn't like the types were different - I suppose I could explicitly declare the return type to be a common parent type, though.  Thoughts?

modules/internal/ChapelArray.chpl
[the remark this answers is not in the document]
lydia-duncan: I'm mostly intending this type to be under the covers, so a user could say:
```chapel
export proc foo(x: [0..3] int) { ... }
```
and we would insert the transformation from what was actually provided to us into a form that the original foo would understand.  If ExternalArray is not included by default, then we would also need to insert a `use` in the module for this transformation to work, right?

modules/internal/ExternalArray.chpl
mppf: I've run into such errors before - I don't think the compiler is very good right now at changing the type on a copy. But you could do it with the same mechanisms as the array slices. I think it's fine to create the same array type, though. It just makes me want to call it something other than "view".

modules/internal/ChapelArray.chpl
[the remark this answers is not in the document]
mppf: That's cool, but let's leave it as a non-default-included module until such time as the compiler is using it.

modules/internal/ExternalArray.chpl
```diff
@@ -0,0 +1,284 @@
...
+
+    proc dsiClone() {
+      return new unmanaged ArrayViewExternDist();
+    }
+  }
+
+  class ArrayViewExternDom: BaseRectangularDom {
+    const size: uint; // We don't need a lower bound, it will always be zero
+
+    const dist;
```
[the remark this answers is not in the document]
lydia-duncan: We rely on at least having a .dist and .dsiMyDist method on the domain, so I'm just going to leave it for now

modules/internal/ChapelArray.chpl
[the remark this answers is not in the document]
lydia-duncan: If I leave it in modules/internal, a use statement has difficulty finding it (I think because we assume everything in modules/internal is `use`d by ChapelStandard or something else), so I'm moving it to modules/packages for now.

modules/internal/ExternalArray.chpl
lydia-duncan: Aww, illegal cast.  Should we be allowed to cast from a c_ptr into a __ddata?  I don't know enough about the compiler handling of __dddata to know if the error is reasonable or just us being cautious

mppf: We should allow it. You'll have to add a _cast method in CPtr.chpl (e.g.  `inline proc _cast(type t, x) where t:_ddata && x.type:c_ptr && t.eltType == x.eltType` )

lydia-duncan:
This isn't quite sufficient for all use cases - in the case where you are creating an array and "returning it", I think you want to be able to create an instance of this class without needing to supply the explicit memory, and have the pointer not get cleaned up when you return it.  Something that under the covers looks like:

```chapel
export proc giveIntArr(size: uint): c_ptr(int) {
  var a = // make an array that can be returned to C/Python
  return a.getExportableContents(); // Domain, etc. gets cleaned up but underlying data gets returned
}
```

I think I should handle that use case in another PR.

mppf:
> I think I should handle that use case in another PR.

I agree - it seems to be a whole other can of worms. It might be worthwhile to talk about how it should be expressed in an issue.

Issue #9660, opened 2018-05-29
spike: how should DataFrame rows be represented?


psahabu:
## user story
As a Chapel Programmer, I want to inspect and iterate over rows of a DataFrame because that is a natural way to survey a dataset.

### known problems
- Variable length and variable type.
  - Arrays are variable length, but only have a single type.
    - Could put each datatype in a wrapper, creating a new ADT. (Bag?)
  - Tuples are variable type, but fixed length.
    - Would need a variable length tuple type, which seems adventurous.

## questions
- [ ] does this require a whole new ADT?
  - [ ] if so, what ADT is it and what is its interface?
  - [ ] if not, how will the user access a row?

psahabu:
This may be of interest to you, @ben-albrecht.

Issue #9658, opened 2018-05-29
spike: should Index be represented by a sorted domain?


psahabu:
## user story
As a Chapel Implementer, I might want the Index ADT to be represented by a sorted domain because that would simplify the implementation of DataFrames.

## questions
- [ ] does a sorted domain fit all the use cases of an Index?
- [ ] what would the performance characteristics be for a sorted domain?
- [ ] how much effort would be required to implement a sorted domain?

PR #9696, opened 2018-06-01
DataFrame ADT

psahabu:
### Motivation
The motivation for the DataFrame ADT can be found in the connected issue.

### Limitations
Currently it is not possible to emit a row from a DataFrame because a row requires a variable length, variable type ADT. See #9660 for exploration of this issue.

### Notes
Included in this PR are pandas equivalents to our tests. Note that the only significant differences are:
- slightly different spacing when printing a DataFrame

### DIFF
--- a/test/library/standard/DataFrames/DataFrames.chpl
+++ b/test/library/standard/DataFrames/DataFrames.chpl
@@ -700,11 +700,8 @@ module DataFrames {
       this.idx = idx;
       this.complete();
 
-      for (lab, s) in zip(labels, columns) {
-        var sCopy = s.copy();
-        sCopy.reindex(idx);
-        this.columns[lab] = sCopy;
-      }
+      for (lab, s) in zip(labels, columns) do
+        this.insert(lab, s);
     }
 
     iter these() {
@@ -716,8 +713,10 @@ module DataFrames {
       return columns[lab];
     }
 
-    proc this(lab: string) ref {
-      return columns[lab];
+    proc insert(lab: string, s: Series) {
+      var sCopy = s.copy();
+      sCopy.reindex(idx);
+      columns[lab] = sCopy;
     }
 
     proc reindex(idx: Index) {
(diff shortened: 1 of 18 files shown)


test/library/standard/DataFrames/DataFrames.chpl
```diff
@@ -539,15 +616,116 @@ module DataFrames {
...
-          f <~> i + "\t" + d + "\n";
+          f <~> i + "    " + d + "\n";
       }
       f <~> "dtype: " + eltType:string;
     }
+
+    pragma "no doc"
+    proc writeElem(f, i, len: int) {
+      var output = if this.valid(i)
+                   then new string(this[i]: string)
```
benharsh: Can you add a TODO comment to clean this up once the bug is fixed?

test/library/standard/DataFrames/DataFrames.chpl
```diff
@@ -539,15 +616,116 @@ module DataFrames {
...
+        var n = s.nrows();
+        if n > nMax then nMax = n;
+      }
+      return nMax;
+    }
+
+    proc writeThis(f) {
+      if idx {
+        idx.writeThis(f, this);
+      } else {
```
benharsh: I feel like we should create an actual Index if the user doesn't provide one, then we can remove this branch.

test/library/standard/DataFrames/DataFrames.chpl
```diff
@@ -539,15 +616,116 @@ module DataFrames {
...
+    var labels: domain(string);
+    var columns: [labels] Series;
+    var idx: Index;
+
+    // TODO: init with labels arg
+
+    proc init(columns: [?D] Series) {
+      this.labels = D;
+      this.columns = columns;
+      this.idx = nil;
```
benharsh: Isn't a dataframe supposed to take a copy of the given series?

test/library/standard/DataFrames/DataFrames.chpl
psahabu: Right now the practice in the ADTs is that a `nil` idx implies a dense range. #9658 should address this, at which point I agree that we should remove these weird `if idx` branches.

psahabu:
@benharsh mind taking another look? Pretty sure I tackled it all.

PR #9701, opened 2018-06-01
Fix internal error with nested calls in new expressions

### LINKED ISSUE #8503: Default initializers: compilation failure for dmapped domains
### Summary of Problem
Calling a default initializer for a class with a dmapped-domain field results in an internal compiler error.

It looks like domain literals in the AST aren't normalized properly (or in the expected way) for code like ``dmapped Block({1..10})``.

### Steps to Reproduce

**Source Code:**

```chpl
use BlockDist;

class C {
  var dom = {1..10} dmapped Block({1..10});
}

proc main() {
  var c = new C();
}
```

**Compile command:**
``chpl foo.chpl --force-initializers``

Output:

```txt
foo.chpl:4: internal error: CAL0078 chpl version 1.17.0 pre-release (e00d06a)
Note: This source location is a guess.

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see https://chapel-lang.org/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.
```

**Associated Test(s):**
[`test/users/npadmana/npb-mg/mg.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/users/npadmana/npb-mg/mg.chpl
          )

benharsh:
Nested calls within a ``new`` expression could sometimes result in an internal compiler error if that expression was used as the default value for a field. The internal error was encountered because normalization left the ``new`` expression in an invalid state.

At the end of normalization, we expect a ``new`` expression to both:
- no longer have a single call as the expression's actual
- not have any nested calls

If the expression was stored within a DefExpr, normalization would not introduce call temporaries, which would leave nested calls in the ``new`` expression. Unaware of this fact, we would still "unpack" the ``new`` expression's single actual into some form expected later. This unpacking would prevent call temps from being inserted, and result in unexpected AST.

This problem is fixed by choosing not to unpack the ``new`` expression if the expression does not belong to a statement. This indicates that the expression is within a DefExpr or something else that will be lowered later.

Resolves #8503

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -1282,6 +1282,13 @@ static void insertRetMove(FnSymbol* fn, VarSymbol* retval, CallExpr* ret) {
 * Transform   new (call (call (partial) C _mt this) args...)) args2...        *
 *      into   new (call       (partial) C _mt this) args...   args2...        *
 *                                                                             *
+* Do not transform calls that are nested within a DefExpr (if stmt-expr is    *
+* NULL). Calls within DefExprs do not have call-temps inserted, which if      *
+* transformed would lead to incorrect AST like:                               *
+*   (new C (call foo))                                                        *
+* The expectation is that the expressions will be normalized later once the   *
+* DefExpr's init/type expressions are copied into a BlockStmt.                *
+*                                                                             *
 ************************************** | *************************************/
 
 static void fixPrimNew(CallExpr* primNewToFix);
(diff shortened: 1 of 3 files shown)


benharsh:
@mppf : A delete-free future changed from an internal error to a program output mismatch as a result of this PR. I just updated the .bad file for now.

PR #9706, opened 2018-06-04
Remove .bad files with no corresponding .future files

daviditen:
.bad files should only exist when there is a corresponding .future.  Remove
them otherwise.


daviditen:
Thanks!

PR #9707, opened 2018-06-04
Hdf5 parallel reads

daviditen:
Add a procedure for reading arrays of basic types from multiple HDF5 files in parallel

The currently supported types are integers, reals and c_strings.  The parallelism
is one task per locale since multiple tasks accessing HDF5 at the same time from
the same process is not supported.  The arrays to read can be any rank.

Add a test that reads arrays of int(64), real(64) and c_strings.


modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+      const hdf5Type = getHDF5Type(eltType);
+      H5LTread_dataset(file_id, dsetName.c_str(), hdf5Type, c_ptrTo(data));
+    }
+
+    record ArrRec {
+      type eltType;
+      param rank: int;
+      var D: domain(rank);
+      var A: [D] eltType;
+    }
```
vasslitvinov: What would it take to make it an "official" type for use by Chapel users? I wanted it myself the other day.

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+
+      var filenames: [1..0] string;
+      for f in findfiles(dirname) {
+        if f.startsWith(dirname + '/' + filenameStart:string) &&
+           f.endsWith(".h5") {
+          filenames.push_back(f);
+        }
+      }
+      const Space = {1..filenames.size};
+      const BlockSpace = Space dmapped Block(Space, locs, 1);
```
vasslitvinov: Would be nice to put `dataParTasksPerLocale=` in front of the `1` , to help future readers and maintainers of this code.

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+      proc deinit() {
+        H5Fclose(file_id);
+      }
+    }
+    */
+
+    /* Read the dataset named `dsetName` out of all HDF5 files in the
+       directory `dirname` with filenames that begin with `filenameStart`.
+       This will read the files in parallel with one task per locale in the
+       `locs` array.
```
vasslitvinov: Seems like it `locs` contains duplictate locales then you can read using multiple tasks per locale. Because each task will have its own process. If so, how about reflect it in this chpldoc comment somehow?

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+        }
+      }
+      const Space = {1..filenames.size};
+      const BlockSpace = Space dmapped Block(Space, locs, 1);
+      var files: [BlockSpace] ArrRec(eltType, rank);
+      forall (f, name) in zip(files, filenames) {
+        var locName = name; // copy this string to be local
+        var file_id = H5Fopen(locName.c_str(), H5F_ACC_RDONLY, H5P_DEFAULT);
+        var dims: [0..#rank] hsize_t;
+        H5LTget_dataset_info_WAR(file_id, dsetName.c_str(), c_ptrTo(dims), nil, nil);
```
vasslitvinov: Hm is there protection against out-of-bounds access to `dims` ? Should we add it?

modules/packages/HDF5_HL.chpl
vasslitvinov: How about add something like "It will return a distributed array of ArrRec instances. Each instance will reside on the locale where the corresponding data was read."

What do you think whether this is the right behavior or the return thing should have all the arrays on the same locale?

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+        var data: [0..# (* reduce dims)] eltType;
+        readHDF5Dataset(file_id, dsetName, data);
+
+        var rngTup: rank*range;
+
+        for i in 0..#rank {
+          rngTup[i+1] = 1..dims[i]:int;
+        }
```
vasslitvinov: How about `for param` in the above?
I would prefer for the for-loop to be all in one line right after the "var rngTup" line.

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+    proc readHDF5Dataset(file_id, dsetName: string, data) {
+      type eltType = data.eltType;
+
+      proc getHDF5Type(type eltType) {
+        var hdf5Type: hid_t;
+        select eltType {
+          when int(8) {
+            hdf5Type = H5T_STD_I8LE;
+          }
+          when int(16) {
```
vasslitvinov: I would prefer for each "when" case to take just 1 line, it is easier to read the entire function. C_string is OK as-is.

modules/packages/HDF5_HL.chpl
```diff
@@ -3516,4 +3517,128 @@ module HDF5_HL {
...
+      }
+      return files;
+    }
+
+    /* Read the dataset named `dsetName` out of the open file that `file_id`
+       refers to.  Store the input into the array `data`.
+       Can read data of type int/uint (size 8, 16, 32, 64), real (size 32, 64),
+       and c_string.
+     */
+    proc readHDF5Dataset(file_id, dsetName: string, data) {
```
vasslitvinov: I suggest to start this function with something like:

```chpl
if !isArray(data) then compilerError("'data' must be an array");
```

modules/packages/HDF5_HL.chpl
vasslitvinov: Also, if we reference it in the above chpldoc comment, we could either copy it into the comment or add chpldoc on it. If desired, rename it to `ArrayWrapper`.

daviditen: > What would it take to make it an "official" type for use by Chapel users? I wanted it myself the other day.

We'd have to come up with a name for it that people like, decide where to put it (maybe ChapelArray?), and document it.  I'd be fine with the name `ArrayWrapper`.

I'll change the name and add/modify some chpldoc comments.

modules/packages/HDF5_HL.chpl
daviditen: 👍

modules/packages/HDF5_HL.chpl
daviditen: If `locs` contains duplicate locales they will refer to the same process, and it will cause undefined behavior within HDF5.  I'll add this to the comment.

I'll add a comment about the return type.

I think the returned thing being all on one locale or being distributed could both be the right behavior depending on how many arrays there are, how big they are, and how they will be accessed.

modules/packages/HDF5_HL.chpl
daviditen: I can add a check that the specified `rank` matches the number of dimensions of the data in the file.

modules/packages/HDF5_HL.chpl
daviditen: 👍

modules/packages/HDF5_HL.chpl
daviditen: 👍

modules/packages/HDF5_HL.chpl
daviditen: 👍

PR #9712, opened 2018-06-04
Improve fixTypeNames

mppf:
* adjusts type.cpp toString to contain the same logic, in
   case order of resolution changes
 * fixTypeNames now calls toString

I considered having a late stage of resolution call toString
on all of the types. However the current issue with that is
that array fields in records would cause rebuilding the record
or class type name, e.g.:

    functions/vass/compile-time-messages.chpl

which complicates the implementation enough that it made
me question whether or not it would be an improvement
(over computing it at generic-field-resolution time).

We can't just call toString later in compilation (and never
bother to adjust the ->name fields) because the eltType
field won't generally exist after resolution.

Passed full local testing.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -199,6 +199,7 @@ static void insertReturnTemps();
 static void initializeClass(Expr* stmt, Symbol* sym);
 static void ensureAndResolveInitStringLiterals();
 static void handleRuntimeTypes();
+//static void fixTypeNames();
 static void pruneResolvedTree();
 static void removeUnusedFunctions();
 static void removeUnusedTypes();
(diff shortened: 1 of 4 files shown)


mppf:
@benharsh - could you have a look at this fixTypeNames change? it doesn't go as far as I think we eventually want but it does solve the immediate issue I was facing.

compiler/resolution/functionResolution.cpp
```diff
@@ -8441,6 +8447,18 @@ static void handleRuntimeTypes()
...
+  if (developer == false) {
+    forv_Vec(TypeSymbol, ts, gTypeSymbols) {
+      const char* typeName = toString(ts->type);
+      if (ts->name != typeName)
+        ts->name = typeName;
+    }
+  }
+}
+*/
+
```
benharsh: Were these left in accidentally?

mppf: I wanted to show you what it could look like, if we went further in that direction.

PR #9713, opened 2018-06-04
Catch use-before-def errors in default initializers

benharsh:
Prior to this PR the compiler would not catch use-before-def errors in default initializers because ``checkUseBeforeDefs`` was only called during the ``normalize`` pass. Default initializers are created after normalization, and so were not checked. This PR updates the compiler to:
A) Pull use-before-def logic into a function that operates on FnSymbols
B) Invoke this new function after a default initializer is normalized.

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/include/passes.h
+++ b/compiler/include/passes.h
@@ -124,6 +124,7 @@ bool isVirtualIterator(Symbol* iterator);
 // normalize.cpp
 void normalize(FnSymbol* fn);
 void normalize(Expr* expr);
+void checkUseBeforeDefs(FnSymbol* fn);
 
 // parallel.cpp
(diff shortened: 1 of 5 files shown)


compiler/include/passes.h
```diff
@@ -124,6 +124,7 @@ bool isVirtualIterator(Symbol* iterator);
 // normalize.cpp
 void normalize(FnSymbol* fn);
 void normalize(Expr* expr);
+void checkUseBeforeDefs(FnSymbol* fn);
```
lydia-duncan: Is it worth giving normalize its own .h file?

Issue #9715, opened 2018-06-04
redundant .type for array type fails


mppf:
### Summary of Problem

I expect .type applied to something that is already a type to do nothing, but when the type is an array type, it causes compilation failures:

``` chapel
var A:[1..10] int;

proc tf(type t) type {
  return t.type;
}

tf(A.type);
```

**Associated Future Test(s):**
  test/arrays/ferguson/runtime-type-type.chpl

mppf:
I think preFoldPrimOp / PRIM_TYPEOF handling is related but fixing that to do nothing for type variables doesn't solve the problem.

lydia-duncan:
I was actually just looking at this.  In the handling of PRIM_TYPEOF, we insert a call to convert the runtime value to a type when a runtime type is involved, but we don't check to ensure what we had was a value beforehand.

lydia-duncan:
I don't have a fix myself, though, I was running into it because I was doing something else wrong with my new array types

lydia-duncan:
That's probably about the same conclusion you came to yourself

bradcray:
Am I understanding that you think that `t.type` should return `t` for a type t?

I'm not convinced that `.type` should be able to be applied to types (i.e., that instead, the compiler should warn/error when this happens).  It strikes me as being more likely to be a misunderstanding on the user's part than a useful feature.  If we had things that were generic over both types and values, perhaps it would make sense, but I don't think we do?

mppf:
> Am I understanding that you think that `t.type` should return `t` for a type t?

Yes -- I think it should be idempotent. But at the very least the program in this issue should be a reasonable error.

ben-albrecht:
>> Am I understanding that you think that `t.type` should return `t` for a type t?

> Yes -- I think it should be idempotent.

It would make more sense to me for the `type.type` to be a bare "type" or compiler error than the original value type.

Here's what some other modern languages do with this design decision:

| Language | `print(type(type(1))`        |
|----------|------------------------------|
| Go       | Compiler error               |
| Rust     | Printing types not supported |
| Swift    | `int`                        |
| Python   | `type`                       |

bradcray:
[...]

PR #9716, opened 2018-06-04
Cast unmanaged to void star

### LINKED ISSUE #9524: delete-free subtask: spike on casting unmanaged to c_void_ptr
As a Chapel user, I expect that since unmanaged pointers are low level pointers that I can easily cast them to `c_void_ptr` for interoperability.

But it does not work now, for example the following fails to compile:
``` chapel
class ParentClass {
  var a:int;
}

class MyClass : ParentClass {
  var x:int;
}

proc test() {
  var x = new unmanaged MyClass(1);
  var y = x:ParentClass;
  var z = x:c_void_ptr;

  writeln(x);
  writeln(y);
  writeln(z);

  var x2 = x:unmanaged ParentClass;
  var y2 = x2:MyClass;
  var z2 = x2:c_void_ptr;

  writeln(x2);
  writeln(y2);
  writeln(z2);
}

test();
```

Getting this working should allow converting c_void_ptr expressions/AtomicObject/voidPtrTest

mppf:
Allow casts to c_void_ptr from unmanaged objects.

Resolves #9524.

Passed full local testing.


mppf:
>    Seems we are not settled on the desired behavior of t.type in #9715 .

It doesn't actually matter for this PR. This PR removes a .type that was apparently not necessary, so that _to_borrowed(type t) and _to_unmanaged(type t) do not rely on someType.type. This PR does nothing to address #9715 other than filing a future about it, and a future does not need to represent consensus.

>    The test mentioned in #9524 - expressions/AtomicObject/voidPtrTest - casts borrowed types, not unmanaged types.

I've added an unmanaged version of that test which does pass.

> While I think it should be OK to pass borrowed types out to the C world -- yes that begs for an explicit OK from the team if there hasn't been one yet -- it should be discussed/mentioned in the PR. Assuming that the test passes.

Casting is the user's way to override the defaults of the type system. If it's *coercion* from borrowed to c_void_ptr or c_ptr, then I think it needs discussion. But this PR doesn't add that and it doesn't exist now AFAIK. For *casting*, the user has explicitly indicated that they wish to convert the pointer types, and then I think we need to do it, rather than getting in their way by standing on some principle.

PR #9718, opened 2018-06-04
Update LLNL CoMD to use initializers

benharsh:
This PR updates the Force classes in LLNL's CoMD to use initializers. Prior to this PR this test would fail under --force-initializers.


benharsh:
Oops, accidentally pasted the wrong title/message into this PR (first-class fn init). Title/message are now updated.

PR #9723, opened 2018-06-05
Fix unmanaged for generic arguments and variables

### LINKED ISSUE #9521: delete-free subtask: spike on generic arguments with owned/unmanaged/etc

### LINKED ISSUE #9523: delete-free subtask: spike on variables with unmanaged generic declared type
As a Chapel user, I expect generic capabilities to be independent of owned/unmanaged memory management strategy. As a result, I expect to be able to declare variables of generic type that is owned or unmanaged.

For example, the following test works now on master:

``` chapel
class C {
  type t;
  var x: t;
}
var myC: C = new C(int, 1);
```

Updating it to the following causes it to no longer resolve:
``` chapel
class C {
  type t;
  var x: t;
}
var myC: unmanaged C = new unmanaged C(int, 1);
```

This spike will investigate why it is failing and try to find a strategy to fix it.

### LINKED ISSUE #9580: Can't take unmanaged generic class as function argument

### LINKED ISSUE #9692: varargs and type query syntax
In looking at resolving issue #9521, I've noticed several problems with the implementation of type query arguments in functions with variable arguments.

First, my position is that the type supplied for a varargs argument indicates that the function accepts only 1 argument type (i.e. that the argument is a homogenous tuple) and that the type is the argument type (rather than a tuple type). The language specification, section 13.6, says:

> Either or both the number of variable arguments and their types can
be specified. For example, a basic procedure to sum the values of three integers can be written as
> ``` chapel
>   proc sum(x: int ...3) return x(1) + x(2) + x(3);
> ```
> Specifying the type is useful if it is important that each argument have the same type.  Specifying the number is useful in, for example, defining a method on a class that is instantiated over a rank parameter.

Now, what if instead of `int` that function queried the type?
``` chapel
proc sum(x: ?t...3) return x(1) + x(2) + x(3);
```

Should it be callable as `sum("hi", 1, 2)` ? I think the answer must be "no". The reasoning is that the `?t` syntax indicates that we don't know what type we would write there - but that there is a type that could be written there. However that is not what master (as of 17d0976eb2b4a89b699041f6a653babd8b843421) does:

``` chapel
proc sum(x: ?t...3) return x(1) + x(2) + x(3);
writeln(sum("hi",1,2)); 
// outputs hi12
```

I would expect this program to fail to compile since the `sum` call is not resolveable.

This issue is even more pronounced for queries on arguments to generic records in a varargs context, for example:

``` chapel
record R {
  var x;
}
proc foo(arg:R(?t) ... ?k) {
  writeln(arg);
}
foo(new R(1), new R("hi"));
// outputs ((x = 1), (x = hi))
```

Here again I expect that the ?t stands in for *some type* and that therefore it must be the same type for each argument.

My proposal here is that the ?t used on the type of a vararg refers to the element type and indicates that all of the vararg arguments must be of the same type. The result would be that both of the above examples would no longer compile.

mppf:
This PR makes two groups of changes to improve the combination of `unmanaged`
with generic arguments.

1. Allow 'unmanaged' to apply to a generic argument, as in
   `proc f(arg:unmanaged MyGenericClass)`.

   * Better track the generic-ness of the unmanaged type

   * Parser change to use a primitive instead of a call to get the unmanaged
     version of a type (since resolution can't handle generic arguments to type
     functions).

    * Fix a bug in removeUnusedTypes where unused `ref(unmanaged ClassType)`
      was not being removed

2. Rework type query syntax so that arguments like
   `proc f(arg:unmanaged MyClass(?t))` or `proc g(arg:owned MyClass(?t))`
    can work. These ultimately transform into nested type queries which were
    not supported before this PR.

   * removes PRIM_TUPLE_AND_EXPAND

   * adjusts the normalizer to handle nested type queries

   * treat type queries on var-args as insisting on all vararg components
     having same type (see #9692 for related discussion)

   * adjust varargs functions in module code that were relying on the previous
     behavior (where `proc f(args:?t...)` or `proc f(args:range(?t)...)`
     would allow arguments of different types).

Resolves #9521
Resolves #9523
Resolves #9580
Closes #9692.

This PR leaves as future work the same patterns with owned, shared, or borrowed
(issue #9725). This PR adds these patterns as futures.

Passed full local testing.


modules/internal/ChapelArray.chpl
```diff
@@ -593,7 +593,17 @@ module ChapelArray {
...
+  proc chpl__isTupleOfRanges(tup) param {
+    for param i in 1..tup.size {
+      if !isRangeType(tup(i).type) then
+        return false;
+    }
+    return true;
+  }
+
+  proc chpl__buildDomainExpr(ranges...)
+  where chpl__isTupleOfRanges(ranges) {
```
benharsh: This feels unfortunate to me. I see how ``range(?t) ...`` could require all elements to have the same type, but could we allow the old behavior for ``range(?)...``?

test/functions/ferguson/query/subquery-crazy.chpl
```diff
@@ -0,0 +1,22 @@
+record C {
+  var x;
+}
+record D {
+  var z;
+}
+
+proc f(arg: (C(?t), D( (?u, C(?v))), (?w, ?x)) ...) {
```
benharsh: Nice!

modules/internal/ChapelArray.chpl
mppf: (Off-list you @benharsh made it clear this wasn't meant to be a request for this PR). Even so, I think I'd rather have the language be more consistent in this area than have a convenience for `range(?)`.

Issue #9722, opened 2018-06-05
override checking for methods not called


mppf:
### Summary of Problem
Override checking only works on methods that are actually called. That's because it's currently happening in function resolution. If we moved it to scope resolve, it could do better checking of this sort, but it would no longer allow overriding a generic function with a concrete one or vice versa.

### Steps to Reproduce
``` chapel
class Parent {
  proc f() { }
}

class Child : Parent {
  // overrides parent f()
  proc f() { }
}
// f() never called; Parent, Child never created
```

**Associated Future Test(s):**
test/classes/ferguson/override-missing3.chpl

PR #9724, opened 2018-06-05
Add override keyword and checking for it

### LINKED ISSUE #8567: add override keyword and require its use
Over at CHIP 20, https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/20.rst, there is some discussion of suprising behavior that can happen with libraries and inheritance. To avoid these suprising behaviors, other languages include an `override` keyword
 * D https://dlang.org/spec/attribute.html#override
 * Swift https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Inheritance.html (see section titled Overriding)
 * C++11 http://en.cppreference.com/w/cpp/language/override

These languages differ in whether or not the override keyword is required (to my knowledge, it is required in D and Swift, but not in C++11).

This issue is to propose specifically that we:
 * add an `override` keyword that decorates overriden methods
 * allow this keyword on the left of a `proc` for a method declaration
 * require this keyword for overridden methods

See also #8565, and #8566 which are related improvement suggestions.

### LINKED ISSUE #9094: implement override keyword
As a Chapel user, I want to be able to specify that a method I am implementing is intended to replace a method in a parent class so that I can understand my programs and receive more understandable errors.

Acceptance Criteria: Override keyword is implemented, compiler checks for it on overrides as demonstrated by tests, and all test/module code is converted to use it.

mppf:
This PR adds `override` as a keyword and checking to require its use for 
methods overriding a parent class (aka superclass) method. See CHIP 20 and
#8567 for discussion.

The checking is off by default (as a transitional aid to enable updating code 
gradually) but can be enabled with the new compiler flag --override-checking.

The checking only applies to functions that are resolved (see #9722 for 
discussion of this design tradeoff).

Resolves #9094.
Closes #8567.

Issues #9732 and #9733 record future work to do with the `override` keyword.

Passed full local testing.


compiler/resolution/virtualDispatch.cpp
```diff
@@ -751,6 +781,155 @@ static void filterVirtualChildren() {
...
+}
+
+// This function checks that the override keyword is used appropriately
+// checkOverrides would also be a reasonable name for it.
+static void checkMethodsOverride() {
+
+  TypeToNameToFns map;
+
+  // populate the map
+  forv_Vec(FnSymbol, fn, gFnSymbols) {
```
benharsh: Couldn't you just iterate over gTypeSymbols, then each class' methods?

benharsh:
Does this PR warn/error for overriden methods with differently-named formals? The current compiler considers such methods not to be overriden:
```
proc Parent.foo(temp);
proc Child.foo(tmp); // will not override!?
```

compiler/resolution/virtualDispatch.cpp
mppf: I did it this way because it matches the pattern of other code in this file and so is less likely to find compiler inconsistencies. I expect that all of the definitions of "method" I might want to use aren't always honored by our AST. E.g. fn->_this could be set, whether or not it's in AggregateType->methods.

mppf:
> Does this PR warn/error for overriden methods with differently-named formals? The current compiler considers such methods not to be overriden:

Yes, see the `f` case in  test/classes/ferguson/override/override-mix.chpl

ben-albrecht:
Reminder: Update syntax highlighters

PR #9729, opened 2018-06-05
Force first-class functions to use initializers

### LINKED ISSUE #9535: Update first-class functions to use initializers
The first-class functions implementation still uses constructors when they should use initializers. This issue tracks the progress of converting FCFs to use initializers.

benharsh:
This PR builds upon prior work to allow initializers for first-class-functions, and simply requires that we always generate initializers instead of only generating them under --force-initializers.

Resolves #9535 

Testing:
- [x] local + futures


benharsh:
@lydia-duncan : would you be willing to review? It builds on your previous work and just flips the switch permanently to initializers.

PR #9731, opened 2018-06-05
Switch perf playground to measure enum range implementation

ronawho:


ronawho:
@vasslitvinov are you done with the playground?

PR #9734, opened 2018-06-05
Make our sync-based barrier reusable

ronawho:
Add a reusable sync barrier that's based off the qthreads feb barrier in
[third-party/qthread/qthread-src/src/barrier/feb.c](https://github.com/chapel-lang/chapel/blob/master/third-party/qthread/qthread-src/src/barrier/feb.c).

The barrier actually uses a combination of atomics and syncs, but blocked tasks
still wait on a sync variable, which means they still get de-scheduled.


ronawho:
@daviditen mind reviewing? 

In addition to regular testing, I've also done 10,000 trials for some of the barrier subdirs including `performance/comm/barrier/ studies/hpcc/STREAMS/elliot/ parallel/taskPar/sungeun/barrier/ parallel/taskPar/diten/`

I also did a test run where I forced the sync barrier to be the default just to get some extra sanity checking for tests that just use the default implementation.

test/studies/hpcc/STREAMS/elliot/stream-spmd-barrier.chpl
```diff
@@ -35,9 +35,8 @@ proc main() {
 
   var execTime: [1..numTrials] real;
 
+  var barrier = new Barrier(numTasks, barrierType);
```
daviditen: Doesn't this need to be `reset` between trials?

ronawho: I don't think so, since it's a reusable barrier.

test/performance/comm/barrier/empty-chpl-barrier.chpl
```diff
@@ -22,6 +23,7 @@ proc main() {
   t.start();
   select barrierMode {
     when LocalAtomic do LocalBarrierBarrier();
+    when LocalSync do LocalBarrierBarrier();
```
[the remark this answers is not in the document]
ronawho: whoops, good catch

modules/standard/Barriers.chpl
```diff
@@ -319,63 +319,96 @@ module Barriers {
...
      as a simple barrier or as a split-phase barrier.
    */
-  pragma "no doc" class sBarrier: BarrierBaseType {
+ pragma "no doc" class sBarrier: BarrierBaseType {
+    /* If true the barrier can be used multiple times.  When using this as a
+       split-phase barrier this causes :proc:`wait` to block until all tasks
+       have reached the wait */
+    param reusable = true;
+
+    pragma "no doc"
```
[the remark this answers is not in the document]
ronawho: Unless you feel strongly, I'll probably just leave this as-is since most of these fields already had `no doc` and the ones in the aBarrier class do as well.

test/studies/hpcc/STREAMS/elliot/stream-spmd-barrier.chpl
daviditen: Can the `halt` at `Barriers.chpl:358` ever be reached?  The fact that it doesn't need to be `reset` seems to imply that any number of tasks can call into the barrier.

modules/standard/Barriers.chpl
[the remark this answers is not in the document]
daviditen: 👍

test/studies/hpcc/STREAMS/elliot/stream-spmd-barrier.chpl
ronawho: It can, but it's racey (as is the aBarrier version)


```chpl
var b = new Barrier(here.maxTaskPar);
coforall 1..here.maxTaskPar+1 do
  b.barrier();
```

will halt if all tasks hit the barrier before any of them get released from it.

I was going to add a test for this, but that's when I realized it's racey because it only failed every other time or so.

ronawho:
Oh shoot, I merged without realizing we might have more to discuss with halt for too many tasks barrier-ing. Let me know if you want to discuss that more.

daviditen:
No, you're set.  Merging was fine with me.

PR #9736, opened 2018-06-05
Produce "out of mem" message for Qthreads use of Chapel runtime-managed memory

### LINKED ISSUE #7484: chpl_memalign() should check for out-of-memory.

gbtitus:
We have arranged for Qthreads to get memory from Chapel memory layer
rather than the regular `malloc()` interface.  But we've been using the
internal memory-layer-specific functions (`chpl_malloc()` for example)
rather than the public functions (`chpl_mem_alloc()` for example) that
report out-of-memory conditions and support tracking.  Here, switch
Qthreads' usage to the latter.  This will allow us to get the usual
Chapel "Out of memory allocating <something>" message when the Qthreads
tasking layer asks for memory and the runtime can't supply it.

This fixes #7484.


runtime/src/chplmemtrack.c
```diff
@@ -98,7 +99,19 @@ static size_t totalAllocated = 0; /* total memory allocated */
 static size_t totalFreed = 0;     /* total memory freed */
 static size_t totalEntries = 0;     /* number of entries in hash table */
 
-static chpl_sync_aux_t memTrack_sync;
+
```
ronawho: Add a comment about why we use a pthread_mutex (and why it's safe here)?

third-party/qthread/qthread-src/src/alloc/chapel.c
```diff
@@ -42,10 +46,36 @@ void qt_internal_alignment_init(void) {
...
-    return chpl_memalign(alignment, alloc_size);
+  void *ret;
+
+  assert(alloc_size > 0);
+  switch (alignment) {
+    case 0:
+      ret = chpl_mem_alloc(alloc_size, CHPL_RT_MD_TASK_LAYER_UNSPEC,
+                           0, CHPL_FILE_IDX_INTERNAL);
+      break;
+#if defined(HAVE_16ALIGNED_MALLOC)
```
ronawho: The `configure` check explicitly checks the alignment of `malloc()` and not the wrappers , so I think this is correct. I also don't think this special casing is worthwhile, I'ld just have this function be:

```c
return chpl_mem_memalign(alignment, alloc_size,
                         CHPL_RT_MD_TASK_LAYER_UNSPEC,
                         0, CHPL_FILE_IDX_INTERNAL);
```

ronawho: And jemalloc is really optimized. I can't imagine the branch for the size here will be any faster than just calling the jemalloc aligned interface (and qthreads is pooling anyways, so this routine won't get called much). I'd be in favor of just reverting your second commit that added this fast-path

gbtitus: I could paraphrase from the 3rd commit's message into the PR message -- I think it may be worthwhile to keep this because one could imagine that `malloc()` for small allocs might be faster than `memalign()`.

gbtitus: (and also that that code is duplicated from the regular Qthreads mem shim `src/alloc/base.c`)

runtime/src/chplmemtrack.c
gbtitus: WIll do.

third-party/qthread/Makefile
```diff
@@ -35,8 +35,9 @@ endif
 # Have qthreads use Chapel's allocator, unless directed not to
 ifeq (, $(call isTrue, $(CHPL_QTHREAD_NO_CHPL_ALLOC)))
 CHPL_QTHREAD_CFG_OPTIONS += --with-alloc=chapel
-CFLAGS += -I$(CHPL_MAKE_HOME)/runtime/include/mem/$(CHPL_MAKE_MEM)/
-CFLAGS += -I$(CHPL_MAKE_HOME)/runtime/include/
+CFLAGS += -I$(CHPL_MAKE_HOME)/runtime/include/localeModels/$(CHPL_MAKE_LOCALE_MODEL)
```
ronawho: Is this needed because `#include "chpltypes.h"` includes `chpl-locale-model.h`? Now that we're including a relatively high level chapel header (that could easily change what headers are included) instead of the low-level one, I wonder if we should be more principled about what includes we add here. i.e. I wonder if we should use `./util/config/compileline --includes-and-defines` or something?

third-party/qthread/qthread-src/src/alloc/chapel.c
gbtitus: Hmm, good points (jemalloc highly optimized, this not called much).  Okay, I'll yank it back out.

third-party/qthread/Makefile
gbtitus: Yeah, it could be time for that.  I'll give it a try.

ronawho:
I think we'll want to update the qthreads README (like we do for gasnet/re2) with something like:


```
Chapel modifications to Qthreads
================================

The modifications that we have made to the official Qthreads release are
as follows:

* Something about memory tracking
```

and when we contribute this change upstream we can change it to something like 

```
* Pulled in an upstream update to memory allocation
   - <github link>
```

third-party/qthread/qthread-src/src/alloc/chapel.c
ronawho: While I don't think there's much of a perf win here, I was more concerned about the correctness issue (HAVE_16ALIGNED_MALLOC only being valid for `malloc()` and not for our memory allocator).

PR #9737, opened 2018-06-05
Prevent serialization from qthread sync vars

### LINKED ISSUE #9651: Investigate sync var serialization issue under qthreads
We seem to have an issue where using sync vars serializes execution such that all tasks end up running on one qthread worker.

I believe the issue is around: https://github.com/chapel-lang/chapel/blob/1bab5a891757648fac132de749c83803e436d0e3/third-party/qthread/qthread-src/src/feb.c#L178-L187

Basically, the code that wakes up a task that is now unblocked from a sync var seems to assume a work-stealing scheduler is being used. The nemesis scheduler (the one we default to) doesn't support work-stealing  (and tasks aren't marked with `QTHREAD_UNSTEALABLE`) so all tasks are getting put into the current shepherd, and then never getting redistributed.

For this I want to create spmd versions of stream that will barrier on a sync var. This should allow us to see poor performance on master (should basically be single core perf) and once this issue is fixed achieving performance with the reference stream will give us confidence that execution isn't serialized and that we're maintaining the correct task/numa affinity between sync var sleeps/wake-ups.

To actually fix the issue we just need to add some conditionals to the check that spawns to `waiter->rdata->shepherd_ptr->ready`. We want to spawn to that shepherd if the task can't be stolen or if work stealing isn't supported (because the scheduler doesn't support it, or because work-stealing was disabled at runtime.)

ronawho:
Previously, qthreads/tasks that were blocked on a sync var got scheduled onto
whatever shepherd happened to wake them up. The implementation then relied on
work stealing to redistribute the tasks. However, we either use schedulers that
don't support stealing (nemesis) or we run with stealing disabled (distrib w/
STEAL_RATIO=0) so this resulted in all tasks getting scheduled onto the same
shepherd, which effectively serializes execution for those tasks until they
complete.

This just forces qthreads/tasks to get rescheduled on their original shepherd.
Ideally we would check if the scheduler supports stealing and if stealing is
enabled, but there's currently no reliable way to detect those things for all
schedulers, and adding them is more work than I want to take on right now.

This prevents serialization from sync vars, which allows the SPMD version of
stream to achieve full performance when using a sync barrier:

| config           | performance |
| ---------------- | ----------- |
| no barrier       | 84 GB/s     |
| atomic barrier   | 84 GB/s     |
| old sync barrier | 10 GB/s     |
| sync barrier     | 84 GB/s     |

Note that this will hurt threadring's performance. Threadring just passes a
sync token between ~500 tasks one at a time, so running the tasks serially was
actually beneficial. I expect threadring to have a 2x or so slowdown, but it's
not an important benchmark and the serializing behavior was wrong.

This resolves https://github.com/chapel-lang/chapel/issues/9651

### DIFF
--- a/third-party/qthread/README
+++ b/third-party/qthread/README
@@ -7,5 +7,17 @@ convenience and was obtained from:
 
   https://github.com/qthreads/qthreads
 
-Any Chapel issues that seem to be related to Qthreads should be
-directed to the Chapel team at https://chapel-lang.org/bugs.html.
+Any Chapel issues that seem to be related to Qthreads should be directed
+to the Chapel team at https://chapel-lang.org/bugs.html.
+
+
+Chapel modifications to Qthreads
+================================
+
+The modifications that we have made to the official Qthreads release are
+as follows:
+
+* We force tasks being woken from sync vars to run on their original
+  shepherd. This is to work around a task serialization bug that stems
+  from us using schedulers that don't support work stealing (nemesis)
+  or running with work stealing disabled (distrib w/ QT_STEAL_RATIO=0)
--- a/third-party/qthread/qthread-src/src/feb.c
+++ b/third-party/qthread/qthread-src/src/feb.c
@@ -175,7 +175,11 @@ static inline void qt_feb_schedule(qthread_t          *waiter,
 {
     qthread_debug(FEB_DETAILS, "waiter(%p:%i), shep(%p:%i): setting waiter to 'RUNNING'\n", waiter, (int)waiter->thread_id, shep, (int)shep->shepherd_id);
     waiter->thread_state = QTHREAD_STATE_RUNNING;
-    if ((waiter->flags & QTHREAD_UNSTEALABLE) && (waiter->rdata->shepherd_ptr != shep)) {
+    // Chapel hack -- we either use schedulers that don't support work stealing
+    // (nemesis), or we run with stealing disabled (distrib w/ STEAL_RATIO=0).
+    // Spawning to the current shepherd below serializes execution and relies
+    // on work stealing to redistribute, so just always spawn to the orig shep
+    if (1) { //(waiter->flags & QTHREAD_UNSTEALABLE) && (waiter->rdata->shepherd_ptr != shep)) {
         qthread_debug(FEB_DETAILS, "waiter(%p:%i), shep(%p:%i): enqueueing waiter in target_shep's ready queue (%p:%i)\n", waiter, (int)waiter->thread_id, shep, (int)shep->shepherd_id, waiter->rdata->shepherd_ptr, waiter->rdata->shepherd_ptr->shepherd_id);
         qt_threadqueue_enqueue(waiter->rdata->shepherd_ptr->ready, waiter);
     } else


third-party/qthread/qthread-src/src/feb.c
```diff
@@ -175,7 +175,11 @@ static inline void qt_feb_schedule(qthread_t          *waiter,
 {
     qthread_debug(FEB_DETAILS, "waiter(%p:%i), shep(%p:%i): setting waiter to 'RUNNING'\n", waiter, (int)waiter->thread_id, shep, (int)shep->shepherd_id);
     waiter->thread_state = QTHREAD_STATE_RUNNING;
-    if ((waiter->flags & QTHREAD_UNSTEALABLE) && (waiter->rdata->shepherd_ptr != shep)) {
+    // Chapel hack -- we either use schedulers that don't support work stealing
+    // (nemesis), or we run with stealing disabled (distrib w/ STEAL_RATIO=0).
+    // Spawning to the current shepherd below serializes execution and relies
+    // on work stealing to redistribute, so just always spawn to the orig shep
+    if (1) { //(waiter->flags & QTHREAD_UNSTEALABLE) && (waiter->rdata->shepherd_ptr != shep)) {
```
gbtitus: Rather a nit, but this seems more clear to me:

    if (1 || ((waiter->flags & QTHREAD_UNSTEALABLE) && (waiter->rdata->shepherd_ptr != shep))) {

The target compiler should be able to constant-fold this at compile time, so no perf implications. It's okay if you don't want to change it, though.

gbtitus (approved):
This looks good, whether or not you choose to do anything about my line-specific comment.

I poked around and didn't see any way to set `QTHREAD_UNSTEALABLE` on all qthreads from the Chapel tasking shim, plus we might want that set only when rescheduling a waiter, not everywhere.  Another solution might be to set it on a qthread at the time it is de-scheduled because the sync var is not in the desired state, but looking at the code that would require changes in many more places, it seems.  (You probably looked into this also; I'm just trying to cover all the bases.)

third-party/qthread/README
```diff
@@ -7,5 +7,17 @@ convenience and was obtained from:
...
+Chapel modifications to Qthreads
+================================
+
+The modifications that we have made to the official Qthreads release are
+as follows:
+
+* We force tasks being woken from sync vars to run on their original
+  shepherd. This is to work around a task serialization bug that stems
+  from us using schedulers that don't support work stealing (nemesis)
+  or running with work stealing disabled (distrib w/ QT_STREAL_RATIO=0)
```
cassella: *`QT_STEAL_RATIO`

ronawho:
>I poked around and didn't see any way to set QTHREAD_UNSTEALABLE on all qthreads from the Chapel tasking shim, plus we might want that set only when rescheduling a waiter, not everywhere. Another solution might be to set it on a qthread at the time it is de-scheduled because the sync var is not in the desired state, but looking at the code that would require changes in many more places, it seems. (You probably looked into this also; I'm just trying to cover all the bases.)

I spent a fair amount of time digging through qthreads looking for a cleaner solution than this. For now I'd like to just go with the simple hack (since we always want to take this branch with our current schedulers anyway). I'll open up a followup issue against us and probably against upstream qthreads to come up with a cleaner solution

PR #9739, opened 2018-06-06
Move TargetLocaleComparator out of class SparseBlockDom

mppf:
Some of my changes during development of PR #9677 caused problems with
this code (when changing scope resolution for the this type for primary
methods to make them more like secondary methods). Since nested classes
are not working well with initializers, it seems smart to avoid using
nested classes in module code at this point. For this reason, this PR
moves TargetLocaleComparator to a standalone record outside of
SparseBlockDom.

Passed full local testing and gasnet testing.


benharsh:
Looks good to me. Have you done any gasnet testing?

mppf:
I have done some, but I'll give it a full gasnet run before merging it.

PR #9744, opened 2018-06-06
Clean error message when forgetting new in a throw

ben-albrecht:
This PR changes the error message for a throwing an `Error` with a missing `new` from this:

```
throwNew.chpl:2: In function 'f':
throwNew.chpl:3: error: unresolved call 'chpl_fix_thrown_error(type Error)'
$CHPL_HOME/modules/internal/ChapelError.chpl:338: note: candidates are: chpl_fix_thrown_error(err: Error)
$CHPL_HOME/modules/internal/ChapelError.chpl:352: note:                 chpl_fix_thrown_error(ref err: Owned(Error))
```

to this:

```
throwNew.chpl:2: In function 'f':
throwNew.chpl:3: error: Cannot throw a type: 'Error'. Did you forget the keyword 'new'?
```

### DIFF
--- a/modules/internal/ChapelError.chpl
+++ b/modules/internal/ChapelError.chpl
@@ -342,8 +342,8 @@ module ChapelError {
 
   pragma "no doc"
   pragma "insert line file info"
-  proc chpl_fix_thrown_error(type Error) {
-    compilerError("Cannot throw a type: '", Error:string, "'. Did you forget the keyword 'new'?");
+  proc chpl_fix_thrown_error(type errType) {
+    compilerError("Cannot throw a type: '", errType:string, "'. Did you forget the keyword 'new'?");
   }
 
   pragma "no doc"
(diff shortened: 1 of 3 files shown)


modules/internal/ChapelError.chpl
```diff
@@ -340,6 +340,12 @@ module ChapelError {
     return chpl_fix_thrown_error(_to_unmanaged(err));
   }
 
+  pragma "no doc"
+  pragma "insert line file info"
+  proc chpl_fix_thrown_error(type Error) {
```
mppf: Can we rename this argument to errorType, to distinguish from the global type named Error?

mppf: the other args are named `err` and have type `Error`

mppf: (you could rename this argument to `err` and that'd be reasonable)

PR #9748, opened 2018-06-06
Fix managed new in loop expressions

### LINKED ISSUE #9525: delete-free subtask: promotion and new
As a Chapel user, I expect to be able to use promotion to initialize managed class instances since I previously could use promotion to initialize unmanaged class instances.

```
$ cat test/functions/deitz/test_promote_generic.chpl 
```

``` chapel
class C {
  type t;
  var x: t;
}

proc foo(c: C) {
  return c.x;
}

var A: [1..4] C(int) = [i in 1..4] new C(int, i);

writeln(A);

writeln(foo(A));

for a in A do delete a;
```

Converting this test to `new unmanaged` or `new owned` leads to failures.

mppf:
This PR fixes several problems with the combination of `new owned` or `new 
unmanaged` with a loop expression. For example

``` chapel
class C { }
var arr = for 1..3 do new owned C();
```

and
``` chapel
class C {
  type t;
  var x: t;
}

// note: arg c here is generic
proc foo(c: C) {
  return c.x;
}

var A = [i in 1..4] new owned C(int, i);
foo(A);
```

* Fixes problems with l-value checking an owned in this context
* Fixes promotion for array of owned to a generic argument
* Adds tests for loop expressions with new shared/owned/unamanged.

Resolves #9525

Passed full local testing.


compiler/resolution/functionResolution.cpp
```diff
@@ -5459,7 +5459,8 @@ static void resolveMoveForRhsSymExpr(CallExpr* call) {
       // todo: not all const if it is zippered and one of iterators is var
       if (isReferenceType(rhsType)                == false &&
           isTupleContainingAnyReferences(rhsType) == false &&
-          rhsType->symbol->hasFlag(FLAG_ARRAY)    == false) {
+          rhsType->symbol->hasFlag(FLAG_ARRAY)    == false &&
+          rhsType->symbol->hasFlag(FLAG_COPY_MUTATES) == false) {
```
vasslitvinov: I would prefer to get rid of "== false" in the above, while we are here.

mppf: I'm going to leave it for now but we should at some point have a broader discussion about code style. While I think if (something == true) is unnecessarily verbose, if (something == false) seems about as reasonable to me as writing if (!something).

PR #9756, opened 2018-06-07
Type tuple error messages

daviditen:
Add user errors for accessing a tuple of types out of bounds.  This was previously an
internal error.

Add a user error for passing too many arguments to a type-tuple indexing expression.
This was previously allowed to compile and run with the extra arguments ignored.

Add new tests for each of these errors.


daviditen:

compiler/resolution/preFold.cpp
```diff
@@ -1123,14 +1123,27 @@ static Expr* preFoldNamed(CallExpr* call) {
...
+
+      if (!at) {
         USR_FATAL(call, "illegal type index expression");
       }
 
+      if (index <= 0 || index > at->fields.length-1) {
+        char msg[60];
+        snprintf(msg, 60,
+                 "type index expression '%" PRId64 "' out of bounds", index);
+        USR_FATAL(call, msg);
```
mppf: Shouldn't this just be USR_FATAL(call, "type index expression '%i out of bounds", (int) index) ?

daviditen: Yes, that sounds better to me.  I'll update it.

daviditen:
Thanks @mppf!  I just pushed your suggested improvement.

PR #9758, opened 2018-06-07
Adjust definition of new external array types to use runtime defined type

lydia-duncan:
This is a step along the way towards allowing exported functions to return arrays.

We realized that if we returned a pointer to the memory directly then we wouldn't
know the size of the array, so we needed to return a wrapper with the memory,
the size, and a pointer to the appropriate free function.  This implements the
wrapper we need in the runtime for design review, the next step will be to block
out what that return would look like.

Testing:
- [x] fresh checkout
- [x] fifo
- [x] valgrindexe
- [ ] full paratest

### DIFF
--- a/modules/packages/ExternalArray.chpl
+++ b/modules/packages/ExternalArray.chpl
@@ -27,6 +27,7 @@ module ExternalArray {
 
   extern type free_func;
   private extern const FREE_FUNC_NIL: free_func;
+  private extern const FREE_FUNC_CHAPEL_WRAP: free_func;
 
   extern record external_array {
     var elts: c_void_ptr;
@@ -82,7 +83,7 @@ module ExternalArray {
     proc dsiBuildArray(type eltType) {
       var data = new external_array(c_malloc(eltType, this.size),
                                     this.size,
-                                    __primitive("cast", free_func, c_free));
+                                    FREE_FUNC_CHAPEL_WRAP);
       var arr = new unmanaged ExternArr(eltType,
                                         _to_unmanaged(this),
                                         data,
--- a/modules/packages/ExternalArray.chpl
+++ b/modules/packages/ExternalArray.chpl
@@ -25,7 +25,6 @@
 module ExternalArray {
   use ChapelStandard;
 
-  extern proc callFree(x: external_array);
   extern type free_func;
   private extern const FREE_FUNC_NIL: free_func;
 
@@ -35,6 +34,7 @@ module ExternalArray {
 
     var free: free_func;
   }
+  extern proc call_free(x: external_array);
 
   pragma "use default init"
   class ExternDist: BaseDist {
@@ -272,7 +272,7 @@ module ExternalArray {
 
     proc dsiDestroyArr() {
       if (_owned) {
-        callFree(_ArrInstance);
+        call_free(_ArrInstance);
       }
     }
 
--- a/modules/packages/ExternalArray.chpl
+++ b/modules/packages/ExternalArray.chpl
@@ -25,6 +25,17 @@
 module ExternalArray {
   use ChapelStandard;
 
+  extern proc callFree(x: external_array);
+  extern type free_func;
+  private extern const FREE_FUNC_NIL: free_func;
+
+  extern record external_array {
+    var elts: c_void_ptr;
+    var size: uint;
+
+    var free: free_func;
+  }
+
   pragma "use default init"
   class ExternDist: BaseDist {
 
@@ -69,9 +80,12 @@ module ExternalArray {
     }
 
     proc dsiBuildArray(type eltType) {
+      var data = new external_array(c_malloc(eltType, this.size),
+                                    this.size,
+                                    __primitive("cast", free_func, c_free));
       var arr = new unmanaged ExternArr(eltType,
                                         _to_unmanaged(this),
-                                        c_malloc(eltType, this.size),
+                                        data,
                                         true);
       return arr;
     }
@@ -148,7 +162,7 @@ module ExternalArray {
 
     const dom;
     
-    const _ArrInstance: _ddata;
+    const _ArrInstance: external_array;
 
     const _owned: bool;
 
@@ -156,7 +170,7 @@ module ExternalArray {
       super.init(_decEltRefCounts = false);
       this.eltType = eltType;
       this.dom = dom;
-      this._ArrInstance = _ArrInstance: _ddata(eltType);
+      this._ArrInstance = _ArrInstance;
       this._owned = _owned;
     }
 
@@ -219,19 +233,19 @@ module ExternalArray {
 
     inline proc dsiAccess(i) ref {
       checkBounds(i);
-      return _ArrInstance(i(1));
+      return (_ArrInstance.elts: c_ptr(eltType))(i(1));
     }
 
     inline proc dsiAccess(i)
       where shouldReturnRvalueByValue(eltType) {
       checkBounds(i);
-      return _ArrInstance(i(1));
+      return (_ArrInstance.elts: c_ptr(eltType))(i(1));
     }
 
     inline proc dsiAccess(i) const ref
       where shouldReturnRvalueByConstRef(eltType)  {
       checkBounds(i);
-      return _ArrInstance(i(1));
+      return (_ArrInstance.elts: c_ptr(eltType))(i(1));
     }
 
     inline proc checkBounds(i) {
@@ -258,7 +272,7 @@ module ExternalArray {
 
     proc dsiDestroyArr() {
       if (_owned) {
-        _ddata_free(_ArrInstance, dom.size);
+        callFree(_ArrInstance);
       }
     }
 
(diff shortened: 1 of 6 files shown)


lydia-duncan:
lol, the Travis failure is because it thinks a field named free is the same as the normal free function.  I'm going to rename the field

modules/packages/ExternalArray.chpl
```diff
@@ -25,6 +25,18 @@
...
 
+  extern type free_func;
+  private extern const FREE_FUNC_NIL: free_func;
+  private extern const FREE_FUNC_CHAPEL_WRAP: free_func;
+
+  extern record external_array {
+    var elts: c_void_ptr;
+    var size: uint;
+
+    var freer: free_func;
```
mppf: Since you don't use it, I wouldn't even put freer in this record, but you can if you want.

runtime/include/chpl-external-array.h
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _chplexternarray_H_
+#define _chplexternarray_H_
+
+#include <stdint.h>
+
+typedef void (*free_func)(void*);
```
mppf: If these are ever going to be included by default, and possibly in any case, they should use a chpl_ prefix. Prefixing like that is important in C since it doesn't really have modules/namespacing. I wouldn't want link failures if somebody else's library made a function called call_free, say.

lydia-duncan:
@gbtitus or @ronawho - would one of you mind taking a look at the runtime changes?

gbtitus:
I'm concerned about the asymmetry in the allocate and free operations.  Allocating with `c_malloc()` and freeing with `chpl_mem_free()` isn't safe.  We really need to allocate and free from the same interface.  Although in current coding `c_malloc()` calls `chpl_here_alloc()` which calls `chpl_mem_alloc()`, there's no guarantee that will always be the case.  A change to `chpl_here_malloc()` could cause us to call some other allocator and then we'd be allocating from one management package and freeing to another, with results ranging from segfaults to silent wrong answers.

Somewhat related to this, is there a strong reason that the allocate is coded in the modules while the free is coded in the C runtime?  Splitting a small interface like this across a major division in the software stack seems awkward.  Can `chpl_call_free()` not be coded in the modules, basically by putting its actions in `dsiDestroyArr()`?

lydia-duncan:
Our c_fn_ptr and first class function support does not allow us to grab the function to store directly in the module code, hence saving it as a function pointer variable in the runtime code.  I think that means that you'd prefer the allocation call be done through a function defined in the runtime as well?

lydia-duncan:
I worry about adding yet another strategy of calling chpl_mem_alloc.  My hope was for something that wouldn't involve a lot of specialized functions, which was why I originally went with c_malloc/c_free.  But I ended up unable to use c_free, which led to the current dichotomy

gbtitus:
Do we really need the function pointer in the record, if it's only there to hold the address of the "freer" function and that address is always (I think) either NULL or `chpl_mem_free`?  Could we instead just put a boolean flag called (for lack of a better name) `weNeedToFreeThis` in the record, and then in `dsiDestroyArr()` do the following?  Then we wouldn't need any runtime content at all, would we?

```chpl
if weNeedToFreeThis then c_free(_ArrInstance.elts);
```

mppf:
@gbtitus - there's some context missing. The purpose of this record is for C or Python interoperability, where C or Python might allocate the memory with something other than the Chapel allocator, and if we "own" it, we need to be able to free it appropriately. Or C/Python might need to free something allocated on the Chapel heap but do so with a simple interface (just call the function pointer).

lydia-duncan:
That's basically it.  The other allocators won't need to be visible in Chapel code, but we might get memory that was created using them that we would need to be able to clean up.

gbtitus:
> ... C or Python might allocate the memory with something other than the Chapel allocator, and if we "own" it, we need to be able to free it appropriately.

Okay, I thought I had to be missing some context; this helps a lot, thanks.  So, the `freer` member is just to make sure that the memory is freed to the same interface from which it was allocated.  I.e., it's incumbent upon whoever allocates the memory to make sure `freer` contains the matching `free()` function address.  That being the case, since we can't manage and call arbitrary function pointers in the modules and thus can't do the `free()` there, let's go for second-best and put both the allocate and the free in the runtime.  This will avoid the interface splitting I mentioned before.  Whether to use `chpl_mem_alloc|free` or `chpl_mem_array_alloc|free` is an open question, but for now it won't hurt anything to choose the former and optimize to the latter at some future point.

I think this also means we don't need `CHPL_FREE_FUNC_*`, right, since the value in the `freer` function pointer can be set by the new allocation function in the runtime code in both places where we create these in the module code?

lydia-duncan:
(retesting)

lydia-duncan:
Testing passed, this is ready for another look @gbtitus

modules/packages/ExternalArray.chpl
```diff
@@ -25,6 +25,19 @@
...
   use ChapelStandard;
 
+  extern type chpl_free_func;
+  private extern const CHPL_FREE_FUNC_NIL: chpl_free_func;
+
+  extern record chpl_external_array {
+    var elts: c_void_ptr;
+    var size: uint;
+
+    var freer: chpl_free_func;
```
gbtitus: `freer` isn't used in the Chapel code.  If we abstracted it to be just a `c_void_ptr` then we could get rid of both the `chpl_free_func` and `CHPL_FREE_FUNC_NIL` externs, couldn't we?  (And either pass `nil` or just elide the third argument to `new chpl_external_array()` in `makeArrayFromPtr()`.)

modules/packages/ExternalArray.chpl
```diff
@@ -25,6 +25,19 @@
...
+  }
+  extern proc chpl_make_external_array(elt_size: uint, num_elts: uint):
+    chpl_external_array;
+  extern proc chpl_call_free(x: chpl_external_array);
```
gbtitus: `chpl_call_free` is pretty nonspecific.  Can we name this `chpl_free_external_array()`, by analogy with the `make` function?  (It's okay that the actual free operation is indirect, that's true for a long way down the call stack.)

runtime/src/chpl-external-array.c
```diff
@@ -0,0 +1,47 @@
...
+  return ret;
+}
+
+void chpl_call_free(chpl_external_array x) {
+  if (x.freer != CHPL_FREE_FUNC_NIL) {
+    x.freer(x.elts);
+  }
+}
+
+void chpl_wrap_chapel_free_call(void* mem) {
```
gbtitus: The only reference to this is within this same file, right?  If so, it should be declared static and removed from the .h file.

modules/packages/ExternalArray.chpl
gbtitus: Replying to my own comment: We would then have to cast it when calling via the pointer, but that seems a small price to pay for the added interface information-hiding.

runtime/include/chpl-external-array.h
```diff
@@ -0,0 +1,40 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _chplexternarray_H_
+#define _chplexternarray_H_
```
mppf: this guard does not actually match the filename

modules/packages/ExternalArray.chpl
```diff
@@ -25,6 +25,16 @@
...
   use ChapelStandard;
 
+  extern record chpl_external_array {
+    var elts: c_void_ptr;
+    var size: uint;
+
+    var freer: c_void_ptr;
+  }
+  extern proc chpl_make_external_array(elt_size: uint, num_elts: uint):
+    chpl_external_array;
```
mppf: This looks odd - I'd probably give up on not overflowing the line or wrap it differently.

runtime/include/chpl-external-array.h
```diff
@@ -0,0 +1,40 @@
...
+typedef void (*chpl_free_func)(void*);
+
+typedef struct {
+  void* elts;
+  uint64_t size;
+
+  chpl_free_func freer;
+} chpl_external_array;
+
+const chpl_free_func CHPL_FREE_FUNC_NIL;
```
mppf: Do we still need CHPL_FREE_FUNC_NIL ?

runtime/include/chpl-external-array.h
lydia-duncan: Based on my experience with hyphens in C names, I got scared so I avoided using them...

mppf: it's not the hyphens - I would have expected it to be _chpl_external_array_H_

There is 'external' vs 'extern'

modules/packages/ExternalArray.chpl
lydia-duncan: Like this?
```chapel
  extern proc
  chpl_make_external_array(elt_size: uint, num_elts: uint): chpl_external_array;

  extern proc chpl_free_external_array(x: chpl_external_array);
```

mppf: I like that better but such things are definitely subjective.

runtime/include/chpl-external-array.h
lydia-duncan: Oooooohhhhhhh

Issue #9772, opened 2018-06-08
problem with promoted new for generic class


mppf:
Promoting `new` for a generic class isn't working right now. For example, this program doesn't work correctly:

``` chapel
class C {
  type t;
  proc init(arg:int) {
    t = int;
  }
}

var A:[1..10] int;
var B = new C(A);

writeln(B);

```

It should print out 10 elements but instead prints out 1.

PR #9775, opened 2018-06-08
Get more tests passing with decorated class types

mppf:
* Gets the first 50 failing tests with --warn-unstable passing

- [ ] full local testing


compiler/resolution/functionResolution.cpp
```diff
@@ -6246,6 +6329,11 @@ static void resolveNewHandleGenericClass(CallExpr* newExpr) {
 
   resolveNewGenericInit(initCall, at, initTemp);
 
+  if (FnSymbol* fn = initCall->resolvedFunction())
+    if (fn->hasFlag(FLAG_PROMOTION_WRAPPER))
+      USR_FATAL(initCall,
+                "Promoted initialization of generic classes not implemented");
```
benharsh: Can we add a .future or something for this error message?

mppf: Oh sorry I meant to remove that statement - the check doesn't work. This PR adds futures 
    test/classes/initializers/promotion/generic-promote-new.chpl
    test/classes/initializers/promotion/concrete-promote-new.chpl
 but it's mainly left as future work in issue #9772.

Issue #8793, opened 2018-03-12
Error handling within initializers


bradcray:
As a Chapel Programmer, I would like the ability to throw errors from initializers and catch them when new objects are created because sometimes exceptional conditions can occur during object initialization and need to be handled within the program text without tearing the program down.

Acceptance criteria:
Initializers that throw errors can be written and their errors can be caught

Issue #7134, opened 2017-08-27
error handling: non-inlined iterators cannot throw


mppf:
See for example
```
test/errhandling/ferguson/it-throws-noinline.chpl
```

At the present time, only iterators that are inlined can throw. Otherwise the compiler generates an error because this code path is not implemented completely yet. This issue is a place to refer to this problem.

Perhaps the iterator lowering process should make 'throwing' those functions built from the iterator (advance, zip1, zip2, etc) that contain code that was setting the `error` argument in the original iterator.

ronawho:
Considering that iterator inlining is an optimization that can be disabled with `--baseline`, or `--no-optimize-loop-iterators --no-inline-iterators` I don't think we should add error-handling to any iterator until this is implemented. 

And note that some iterators can be direct inlined (`--inline-iterators`), but cannot be zippered inlined (`--optimize-loop-iterators`) since the limitations on the iterator for the 2 inlining methods are different.

mppf:
@ronawho - right, we havn't been adding error handling to iterators in the modules for this reason.

e-kayrakli:
Today I hit this issue while writing some Arkouda code. It is not a blocker for me, but I would have preferred to be able to do this.

PR #9776, opened 2018-06-08
Add halt wrappers for use in initializers and iterators

ronawho:
Error handling isn't supported for initializers or iterators today. These halt
wrappers are intended to mark such cases so that we can more easily
find/identify them in the future.

See https://github.com/chapel-lang/chapel/issues/7134 and https://github.com/chapel-lang/chapel/issues/8793 for more info

### DIFF
--- a/modules/internal/ChapelIO.chpl
+++ b/modules/internal/ChapelIO.chpl
@@ -689,6 +689,34 @@ module ChapelIO {
     __primitive("chpl_error", tmpstring.c_str());
   }
 
+
+  //
+  // Halt wrappers for cases where we want error-handling, but error-handling
+  // isn't supported yet
+  //
+
+  /*
+     Halt wrapper for cases where we want error-handling in initializers. For
+     more info see: https://github.com/chapel-lang/chapel/issues/8793
+   */
+  pragma "no doc"
+  pragma "function terminates program"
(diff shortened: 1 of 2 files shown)


mppf:
Are you intending to add calls to these in the compiled code in this PR?

ronawho:
Not in this PR, but in PRs that I'll be opening today and monday

mppf:
This PR seems OK to me then. I'd consider prefixing the new routines with `chpl_` though.

ronawho:
Hmm, that's a good point. I'd like to keep the names relatively short/clean, since I plan on adding other wrappers for misc things like:
 
 - pureVirtualMethodHalt
 - unreachableCodeHalt

and wrappers for execution time halts: 
 - nilCheckHalt
 - boundsCheckHalt
 - formalDomainCheckHalt
 - localCheckHalt
 - safeCastCheckHalt
 - divByZeroCheckHalt

Thoughts on moving these wrappers into a new not-included-by-default module named ChapelHaltWrappers or something instead of prefixing with chpl_?

mppf:
> Thoughts on moving these wrappers into a new not-included-by-default module named ChapelHaltWrappers or something instead of prefixing with chpl_?

My thought is simply "+1".

PR #9781, opened 2018-06-08
Support for enum ranges as well as domains and arrays over same

bradcray:
This adds support for enum ranges, such as:

```chapel
enum color {red, orange, yellow, green, blue, indigo, violet};
color.orange..color.blue;
```

It also adds support for domains and arrays over enum ranges.

This PR resolves issue #9648.

The key to the approach taken here is to have the range store and compute on enum indices based on their 0-based "order" rather than their underlying integer values.  This is necessary since the integer values may not be consecutive nor monotonically increasing for a given enum, yet we didn't necessarily want to restrict ranges to those enums which did rely on consecutive sequences of integers.  The challenge to this approach is that it requires ranges to deal with two types where previously one was sufficient:  `idxType`, which is the logical index type of the range in the user's eyes.  And `intIdxType` which is the integer type used in its internal representation of the range (and set to `int` when `idxType` is an enum, `idxType` otherwise).  This affected a lot of code at the boundaries between a range's internals and the public interfaces, since values had to potentially be converted back and forth between enums and ints.

To support these conversions back and forth, I introduced utility functions, `chpl__idxToInt()` and `chpl__intToIdx()` which convert between values of `idxType` and `intIdxType` as required by any given scenario.  Note that for traditional integer index types, these conversions are essentially no-ops.  In addition, I created helper methods `chpl_intToIdx()` on the range and DefaultRectangularDom types to avoid having to pass in an explicit idxType argument (necessary for the standalone versions).

Changes to ChapelRange and DefaultRectangular
Most of the work involved here required generalizing the ChapelRange and DefaultRectangular modules to support distinct types for idxType versus their internal representations:  Lots of conversions at the boundaries, while trying to stay working with ints as much as possible within the code itself.  In support of this, some notable changes were:

* added support for `intIdxType` routines that gives the integral type corresponding to `idxType` (it's the same for integral types and `int` for enums)

* added support for `chpl__idxTypeToIntIdxType()` which converts from `idxType` to `intIdxType`

* removed some needlessly explicitly specified `idxType` field type specifications and relied on type inference, just to keep the code simpler

* removed compiler errors which had blocked ranges of enums from compiling previously

* made variations of `first`, `last`, `alignedLow`, and `alignedHigh` with an `AsInt` suffix so that internal code wanting to work with these values could do so directly without having to convert needlessly back and forth from ints to idxTypes back to ints again.

* similarly, switched some internal code that used to work on `low`, `high`, `alignment` to use the internal `_low`, `_high`, `_alignment` instead to avoid conversions

* changed the public interfaces of `.size` and `.length` to return `intIdxType` rather than `idxType`

* renamed some public-facing arguments for two reasons:  (1) to minimize code changes within the routine itself; (2) because the original argument names were not very descriptive (e.g., `i`)

* made a few `param` routines non-`param` because we don't have param order->enum routines at present, and it didn't look like cases where `param`-ness would matter

* removed the chpl__rangeAlignType() as no longer necessary

* removed the chpl__defaultLowBound()/HighBound() routines as being needlessly complex

I also took a pretty significant pass over the ChapelRange module's documentation.  Originally, this was motivated by the desire to get it correct w.r.t. enum ranges, but while here, I found other ways to improve it, and did so, such as saying more about values and types, moving iteration higher in the documentation given its importance, and tweaking things here and there.  I also took a quick pass over the language specification, but a more careful one is probably warranted (and in doing so, I'd be tempted to move most of the range "library" calls into the online documentation).

Other miscellaneous changes made while here:
* while here, removed BHarsh's `typeHelper()` workaround, no longer necessary now that initializers are better able to handle conditional type expressions
* simplified the definition of .length() which had distinct cases for signed and unsigned `idxType`s that seemed essentially identical
* simplified / generalized the logic for `ident()`

Changes in other modules
---------------------------
* ChapelArray.chpl: changed argument types for expand(), interior(), etc. to take `intIdxType` rather than `idxType`
* ArrayViewReindex.chpl: made the index conversion routine stop constraining arguments to integral types now that idxType can be non-integral

New tests
----------
Many of the following tests test against three distinct enum declarations—one that uses default numbering; one that uses a non-default starting value; and one that uses decreasing numbering—to make sure that we're not relying on the enum's underlying values or their increasing monotonicity accidentally.

* initFromArr.chpl: captures a case that broke until very late in the effort where the bulk transfer interface assumed that the source and destination index types were the same
* zipEnumRangeDomArr.chpl: tries a bunch of loop idioms over enum ranges, domains, arrays including zippered iteration
* zipStridedEnumRangeDomArr.chpl: similar but for strided enum ranges
* arrays/userAPI/enumArray.chpl: makes sure that the existing array API test works for arrays with enum indices
* domains/userAPI/*: added a new set of tests that check that the domain API works for 2D domains and created versions of it that test arithmetic domains (test2Ddomain.chpl) and enum domains (testEnumDomain.chpl)
* baseEnumRangeTest.chpl: replaces a bunch of distinct simple tests that used to ensure enum ranges didn't work
* enumRange.chpl: tries a bunch of simple operations on enum ranges to exercise a bunch of the common interfaces.
* enumRangeStrided.chpl: similar but for strided ranges of enums
* negStridEnum.chpl: make sure that negative strides work

Testing changes
-----------------
* moved existing enum range tests from test/types/range/nonIntTypes/ to test/types/range/enum/
* restored uses of enum ranges that I'd previously removed when enums stopped coercing to integers: enumarray.chpl, test_module_Norm.chpl, cg-arithDomArr.chpl, test/studies/amr/*, comd/llnl/*, mg.chpl


cassella:
Are you concerned about enums with repeated values?  E.g., a `color4.chpl` like

```chapel
module Color {
  enum color {red=1, orange=2, yellow=1, green=2, blue=1, indigo=2, violet=1};
}
```

bradcray:
> Are you concerned about enums with repeated values?

I think that's a general question w.r.t. our enums that's independent of this change since ranges respect the declaration order, not the underlying integer values.  I.e., I think that adding that color4.chpl file wouldn't break anything (and would be happy to add it if that's what you're proposing).

bradcray:
Thinking about this PR this morning and what remains for enums in general, I realized that if we were to support enum op int combinations as in issue #9784, this PR may have been much less impactful on the existing range code...  Hmm...

cassella:
> I think that adding that color4.chpl file wouldn't break anything (and would be happy to add it if that's what you're proposing).

I should have mentioned that there were more errors than I wanted to include in the comment.  Here's a taste


```
[Executing diff enumRange.good enumRange.4-0.exec.out.tmp]
1c1
< red..violet
---
> red..red
5,10d4
< orange
< yellow
< green
< blue
< indigo
< violet
12,13c6,7
< length = 7
< size = 7
---
> length = 1
> size = 1
15c9
< last = violet
---
> last = red
...
---
> orderToIndex(1) = orange
> r.translate(1) = orange..red
> interior(1) = orange..orange
> interior(-1) = red..red
> r.exterior(1) = red..red
> enumRange.chpl:44: error: halt reached - enumerated type out of bounds in chpl__orderToEnum()
[Error matching program output for types/range/enum/enumRange (compopts: 4)]
```

bradcray:
Hmmm... back to the drawing board?

bradcray:
Here's a simpler test that gets to the heart of what breaks Paul's case:

```chapel
enum color {red=1, orange=2, yellow=1, green=2, blue=1, indigo=2, violet=1};

writeln(chpl__enumToOrder(color.violet));
writeln(chpl__orderToEnum(6, color));
```

bradcray:
Taking another stab at this support in [a new branch](https://github.com/bradcray/chapel/tree/enumRange2) that uses math on enums to try and minimize changes.

bradcray:
I've dug into @cassella's  mystery a bit more, and it's independent of this PR (but still worth wrestling with).  See issue #9786.

vasslitvinov:
> simplified the definition of .length() which had distinct cases for signed and unsigned idxTypes that seemed essentially identical (@vasslitvinov: check me on this, as the original author?)

Looks good!

modules/internal/ChapelRange.chpl
```diff
@@ -2350,38 +2454,54 @@ proc _cast(type t, r: range(?)) where isRangeType(t) {
...
-    if isIntegralType(idxType) {
-      return 0:idxType;
-    } else if isEnumType(idxType) {
-      // TODO: This is only correct for default-numbered enums; need
-      // to generalize
-      return 1:idxType;
-    } else {
-      chpl__rangeTypeError(idxType);
-    }
+  inline proc chpl__intToIdx(type idxType, i: integral, j: integral) {
```
benharsh: We briefly discussed this offline, but for the record: I think it could help make this change less verbose if we had a method variant of this function such that we did not need to pass 'idxType' every time.

bradcray:
Thanks for the review, @benharsh.

> What configurations have you tested?

linux64, gasnet, and valgrind

benharsh:
You may want to consider doing some NUMA testing over a subset of the suite, just to head off nightly errors. Otherwise I can't think of any additional configurations to test (that are worth testing manually).

modules/internal/ChapelRange.chpl
```diff
@@ -18,52 +18,86 @@
...
+  Ranges over enumerated types respect the declaration order of its values:
 
   .. code-block:: chapel
 
-    (1..10) + 5 // 6..15
-    (1..10) - 3 // -2..7
-    (1..) + 1 // 2..
-    (..10) + 1 // ..11
+    enum color {red, orange, yellow, green, blue, indigo, violet};
+    color.orange..color.green;   // orange, yellow, green
```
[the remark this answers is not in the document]
bradcray: That's a good point.  I'm not wild about that specific example since it contains duplicate values and there's some question as to whether or not we want to continue to support that (issue #9786).  Let me see if I can come up with something else.

modules/internal/ChapelRange.chpl
bradcray: I've made this conversion for ranges.  I'm also going to do it for DefaultRectangular and move the multidimensional versions of chpl__intToIdx() over to that module since they're only used there.  Intriguingly, I had to (or arguably, chose to) use different names for the method name than the standalone function because otherwise the calls to the routine in the initializers tried to call the method version which isn't legal.  I think this requires either using a different name or using a fully-qualified call which seems like a mouthful for this context and this (internal) module.  Let me know if you disagree with any of this or have different ideas.

modules/internal/ChapelRange.chpl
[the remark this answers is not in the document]
bradcray: I ended up going with the same color set (since it's used throughout the docs) and gave them a permutation of the values 1..7 indicating how "bright" the color is (though I didn't describe this... it was just my mental model).

bradcray:
@benharsh: If you'd be willing to re-review the delta created by the past four commits and see if you're OK with them w.r.t. your feedback, that'd be great.  I'm going to re-run some testing in the meantime.

modules/internal/ChapelRange.chpl
```diff
@@ -976,27 +1073,34 @@ proc _cast(type t, r: range(?)) where isRangeType(t) {
   // Absolute alignment is not preserved
   // (That is, the alignment shifts along with the range.)
   //
-  inline proc +(r: range(?e, ?b, ?s), i: integral)
+  inline proc +(r: range(?e, ?b, ?s), offset: integral)
   {
-    type resultType = (r._low+i).type;
+    const i = offset: r.intIdxType;
```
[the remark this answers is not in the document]
bradcray: Whoops, I completely forgot.  The unsafe cast is necessary to handle the case of adding a signed integer like '-1' to an unsigned integer type.  We rely on the safe wraparound that takes place in this case whereas using a safeCast complains that -1 can't be cast to an unsigned integer.

But wait... how did this work before?  test/domains/bradc/domainMath/uintDomMath.chpl is a case that exposed this... (will look into it tomorrow)

test/arrays/userAPI/enumArray.chpl
```diff
@@ -0,0 +1,12 @@
+use arrayAPItest;
+
+enum color {red=42, orange=33, yellow=31, green=22, blue=8, indigo=5, violet=3};
```
cassella: Did you intend for the scrambling to result in a strictly descending sequence?

(It still passes when I change a few of the values to make the sequence non-monotonic.)

bradcray: Yeah.  I consider this whitebox testing, where I knew the code well enough to feel confident it would work (and felt that other tests covered other cases well enough that something would go wrong if it broke).

PR #9789, opened 2018-06-09
Fix test errors for new external array set up

lydia-duncan:
One setting didn't accept int64_ts as long long ints, so use a macro instead.
The other was more strict about assigning from a void pointer to an extern
type, so set it in the runtime instead.

### DIFF
--- a/modules/packages/ExternalArray.chpl
+++ b/modules/packages/ExternalArray.chpl
@@ -28,12 +28,13 @@ module ExternalArray {
   extern record chpl_external_array {
     var elts: c_void_ptr;
     var size: uint;
-
-    var freer: c_void_ptr;
   }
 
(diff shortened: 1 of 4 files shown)


lydia-duncan:
@mppf  or @gbtitus - care to do a post-commit review of this, assuming it works?  I know it works on my machine and our normal testing machine.  I don't have access to the machine with the llvm.system failure so I can't double check that (and wb testing is just annoying), so I'm going to merge and see how testing the rest behaves tomorrow

mppf:
I don't have any concerns about this

gbtitus:
This seems ok to me as well.

lydia-duncan:
Thanks, that makes me feel better!

PR #9791, opened 2018-06-10
Update enum range branch testing to point to new version

bradcray:
I'm trying a new approach to supporting enum ranges.  This updates the performance playground to test the new approach.


bradcray:
@ronawho : If you happen to see this today, could you verify that this should test my new `enumRange2` branch while still merging the results in with the previous ones?

PR #9799, opened 2018-06-11
Fix regexp leak

mppf:
This PR adjusts re2-interface.cc to free the per-pthread regexp cache when each pthread is completed. This seems to be the right place to free this memory. 
However, since under qthreads these pthreads are not completed by the time
memory tracking report is printed, it doesn't solve the leak under qthreads. I
think a reasonable approach there would be to adjust the memory tracking to run after all of the pthreads (created by qthreads) are joined.

- [x] full local fifo + regexp testing
- [x] full local testing


mppf:
@ronawho - could you have a look at this regexp runtime .cc change? thanks.

mppf:
> Can you make an issue about the qthreads problem? I think we're calling qthread_finalize() before printing out the mem leaks, so that issue is a little surprising to me (unless qthreads isn't cleaning up)

I'll make an issue. I don't know what the problem is.

> Can/should this be converted to using the CHPL_TLS_ wrappers?

I think that would be reasonable to do. Do you want me to try to do that here?

> If we added explicit regexp init/exit functions, would that provide a way to cleanup the TLS without relying on pthread destruction hooks?

I don't actually understand how to do that simply, since the TLS is per-thread, and the runtime would presumably call regexp init / exit from just 1 thread.

mppf:
> > Can/should this be converted to using the CHPL_TLS_ wrappers?

> I think that would be reasonable to do. Do you want me to try to do that here?

Actually I don't see how the _thread case can call a cleanup function with that API. It will pass the destruction function if it calls pthread_key_create, but if we're using GCC (say), it instead has CHPL_TLS_INIT2 defined to nothing.

ronawho:
>> If we added explicit regexp init/exit functions, would that provide a way to cleanup the TLS without relying on pthread destruction hooks?

>I don't actually understand how to do that simply, since the TLS is per-thread, and the runtime would presumably call regexp init / exit from just 1 thread.

Oh duh. Yeah, good point, nevermind.

>>> Can/should this be converted to using the CHPL_TLS_ wrappers?

>> I think that would be reasonable to do. Do you want me to try to do that here?

>Actually I don't see how the _thread case can call a cleanup function with that API. It will pass the destruction function if it calls pthread_key_create, but if we're using GCC (say), it instead has CHPL_TLS_INIT2 defined to nothing.

Yeah, I wasn't sure about that myself, which is why I was asking. Ok with me to leave this PR as-is then.

PR #9800, opened 2018-06-11
Extend checking of JIRA 203 (transient MPP reservation) to handle more

awallace-cray:
Detect regressions with the following error message in the output,
    "Transient MPP reservation error on confirm",
and mark them with a recognizable error msg,
    "Jira 203 -- Transient MPP reservation error",
for easier triage.

An existing filter was already checking the test output for
    "Transient MPP reservation error on create"
and marking those tests with the same "Jira 203" message.


awallace-cray:
Nah.

Issue #9653, opened 2018-05-29
Convert Random module to use error handing


ronawho:
Convert halts in the Random module to throw'ing functions (where appropriate)

ronawho:
Converting `getNth(n: integral)` and `skipToNth(n: integral)` from halting to throw an `IllegalArgumentError` when n <= 0

We believe error-handling is appropriate here. As a point of comparison Python's (now deprecated)  [random.jumpahead](https://docs.python.org/2/library/random.html#random.jumpahead) raised a ValueError.

Converting halts for illegal seed values in the NPB initializer to be `initHalts()` (placeholder until we have a story for error-handing in initializers.) We believe we eventually want error-handling here (instead of say silently adjusting the seed to something legal like some libraries do): 
 - https://people.sc.fsu.edu/~jburkardt/c_src/randlc/randlc.html 
   - https://people.sc.fsu.edu/~jburkardt/c_src/randlc/randlc.c
   - https://people.sc.fsu.edu/~jburkardt/py_src/randlc/randlc.py

PR #9804, opened 2018-06-11
Convert the Random module over to error-handling

ronawho:
Make skipToNth and getNth throw an IllegalArgumentError if n <= 0. Also convert
halts in the NPB initializer to initHalts, since we want to use error-handling
there eventually.

This closes https://github.com/chapel-lang/chapel/issues/9653

### DIFF
--- a/modules/standard/Random.chpl
+++ b/modules/standard/Random.chpl
@@ -285,28 +285,28 @@ module Random {
 
     /*
       Advances/rewinds the stream to the `n`-th value in the sequence.
-      The first value is with n=1.
+      The first value is with n=1.  n must be > 0, otherwise an
+      IllegalArgumentError is thrown.
 
       :arg n: The position in the stream to skip to.  Must be > 0.
       :type n: `integral`
      */
-
-    proc skipToNth(n: integral) {
+    proc skipToNth(n: integral) throws {
       compilerError("RandomStreamInterface.skipToNth called");
     }
 
     /*
       Advance/rewind the stream to the `n`-th value and return it
-      (advancing the stream by one).  This is equivalent to
+      (advancing the stream by one).  n must be > 0, otherwise an
+      IllegalArgumentError is thrown.  This is equivalent to
       :proc:`skipToNth()` followed by :proc:`getNext()`.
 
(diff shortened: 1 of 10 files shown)


ronawho:
That's an excellent question. I don't have any good ideas there. The only thing that comes to mind is introducing a `throws_transition` keyword or something that would generate a warning instead of an error if the `Error` isn't handled in a non-prototype module.

That doesn't seem worthwhile to me, so I would go with the argument that prototype modules prevent `throws` from being required for most cases.

modules/standard/Random.chpl
```diff
@@ -285,28 +285,28 @@ module Random {
...
 
       :arg n: The position in the stream to skip to.  Must be > 0.
       :type n: `integral`
 
       :returns: The `n`-th value in the random stream as type :type:`eltType`.
-    */
-
-    proc getNth(n: integral): eltType {
+     */
+    proc getNth(n: integral): eltType throws {
```
bradcray: This was intriguing to me:  Is it the case that if one overload of a function throws, they all must?

bradcray:
This looks good to me.  I get where Michael's coming from with his gentler deprecation question, but don't see a good way to do it that isn't more effort than it seems to be worth.

modules/standard/Random.chpl
mppf: That's true, but RandomStreamInterface exists just for documentation purposes (i.e. the other streams don't inherit from it).

mppf:
I don't think we necessarily should do more - I just think it's important to think about deprecation it for a moment in cases like this.

modules/standard/Random.chpl
ronawho: Note that this class isn't actually used for anything (so far as I know). The documentation says:

"""
The :class:`RandomStreamInterface` is included here only for documentation
and does not help with compilation in any way.  In the future, we hope to
turn it into an interface.
"""

So this class is intended to be an abstract base class/interface. In this case I think it's an override, not an overload and so far as I know `throws` is part of the signature so overridden methods must also be declared throws if the base method is.

PR #9824, opened 2018-06-12
Have Qthreads use Chapel's copy of the hwloc topology.

gbtitus:
Add a Qthreads affinity shim called "hwloc_via_chapel", which is
identical to the existing "hwloc" except that it attempts to get the
hwloc topology from the Chapel runtime before getting it from hwloc
directly.  This also adds the new Chapel runtime topology layer function
`chpl_topo_getHwlocTopology()` which retrieves the topology the runtime
has loaded, if any.  In addition, where we used to only load the hwloc
topology in the runtime when `locModel!=flat`, we now also do so when
`tasks==qthreads`.

By use of linker `--wrap` options and hwloc init, load, and destroy
wrapper functions I've confirmed that we load and don't load an hwloc
topology appropriately in various locModel and tasking configs.  I also
used this technique along with running the `performance/elliot/no-op` test
to confirm that we only load the topology once rather than twice in the
Qthreads-based configs, demonstrating that Qthreads is indeed using the
runtime's copy of the topology.  Finally, I ran stream-ep on XC-16 BW28
nodes with and without `QT_INFO=1` using the nightly performance settings
to confirm that Qthreads was computing the expected number of CPUs and
getting the expected performance.

### DIFF
--- a/runtime/include/chpl-topo.h
+++ b/runtime/include/chpl-topo.h
@@ -39,6 +39,11 @@ void chpl_topo_init(void);
 void chpl_topo_exit(void);
 
 //
+// retrieve the hwloc topology (if any) for other packages to use
+//
+void* chpl_topo_getHwlocTopology(void);
+
+//
 // how many NUMA domains are there?
 //
 int chpl_topo_getNumNumaDomains(void);
--- a/runtime/src/topo/hwloc/topo-hwloc.c
+++ b/runtime/src/topo/hwloc/topo-hwloc.c
@@ -50,7 +50,7 @@
 #define _DBG_P(f, ...)
 #endif
 
-static chpl_bool haveTopology;
+static chpl_bool haveTopology = false;
 
 static hwloc_topology_t topology;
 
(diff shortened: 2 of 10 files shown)


gbtitus:
@ronawho requesting your opinion: currently we only enable hwloc in the chplenv scripts when `tasks=qthreads`.  Should we also now do so when `locModel!=flat`?

ronawho:
Quick thoughts (haven't done a full review yet):

I think we should remove the `hwloc_v2_via_chapel` version. I'm pretty sure this one is broken and it's not developed anymore. `binders` was supposed to be the new replacement, but I don't see the point in making it available until we actually want to use it. 

Also, don't forget that the qthreads README should be updated

ronawho:
> requesting your opinion: currently we only enable hwloc in the chplenv scripts when tasks=qthreads. Should we also now do so when locModel!=flat?

Practically speaking `locMode !=flat` will always mean `tasks=qthreads`, so I don't have a strong opinion here.

gbtitus:
I've reworked the commits, remove hwloc_v2_via_chapel and splitting things apart better, and this is ready for re-review now.

gbtitus:
I fixed the Makefile; see what you think now.

ronawho:
Looks good!

Issue #9801, opened 2018-06-11
Should we document methods obtained due to forwarding?


lydia-duncan:
Created based on some discussion in #9764 

It is my opinion that because forwarded methods are available for use by clients of the
wrapper type, they should be documented like other methods defined on the type.  This
belief comes from the view that forwarding is usually syntactic sugar for writing a
wrapper for these methods explicitly yourself, which would make the symbols trivially
documentable.

I also think such methods should not be marked specially to indicate they were implemented
through forwarding - forwarding seems to be an implementation detail (since you could write the
explicit wrapper instead, as I mentioned earlier) that probably should be hidden from the client.

There are a few tricky bits to this:
- How do you turn off documentation while still enabling the method to be forwarded?
  - Seems like "private" could come into play here when declaring that methods are
forwarded, but we don't have that implemented yet and I'm not sure how you could
apply "no doc" to the forwarded symbols independently of whether they had "no doc" applied in the original type.
- What do you do if the field marked as forwarding is fully generic?
  - In this case, since you don't know the type at documentation time, you can't know what methods are provided in this way and so can't generate documentation for them.
  - Does this indicate that maybe we shouldn't document symbols obtained through forwarding at all?  Or is it just special to completely generic forwarded fields?

gbtitus:
Documenting methods obtained via forwarding seems like it has an easy failure mode that can result in mis-describing things.  Suppose Alice documents a forwarding method `f()` and then Bill comes along later and changes the original `f()`.  If Bill doesn't know about Alice's use and Alice doesn't know about Bill's change, Alice's documentation may be left incorrect.  It seems like what we really want is a way to forward the documentation, along with the symbol.  That is, we want chpldoc syntax that would result in the documentation for Alice's `f()` actually being generated from that for the "forwardee" symbol `f()` that Bill changed.

lydia-duncan:
That's a good point - I hadn't thought about how forwarded symbols would get explicit documentation.  This seems like it would fall under the same category as how to mark that a forwarded symbol should or should not be documented, and I'm not sure how to syntactically specify it.

PR #9831, opened 2018-06-12
Avoid documenting compiler generated functions, especially forwarding ones

### LINKED ISSUE #9764: Likely unwanted symbols (with no apparent source) appearing in documentation
I found these strange symbols ([domain.forwarding_expr7__value](https://chapel-lang.org/docs/builtins/ChapelArray.html?highlight=domain#ChapelArray.forwarding_expr7__value), [array.forwarding_expr8__value](https://chapel-lang.org/docs/builtins/ChapelArray.html?highlight=domain#ChapelArray.forwarding_expr8__value) in the Chapel Domains and Arrays documentation.

Oddly, a complete search of the chapel source does not bring up this symbol.
There are likely others.

lydia-duncan:
Resolves #9764 

Prior to this change, when a class or record declared a field as forwarding, we would
generate symbols like `forwarding_expr1_fieldName` in our documentation.  This was
obviously not what we wanted, so this change marks those functions as compiler
generated, and then skips documenting compiler generated symbols.

Adds a test covering the fully generic field case (where we don't expect any
additional methods to be documented for the type).  Also add a future for
documenting methods obtained through forwarding, which hasn't quite been
finalized yet and definitely isn't prioritized (see #9801), but I felt was worth tracking.

Passed a full paratest with futures (and chpldoc built)
Double checked that generating our module documentation removes the forwarding
symbol defined on domains


compiler/AST/symbol.cpp
```diff
@@ -255,7 +255,8 @@ bool Symbol::hasEitherFlag(Flag aflag, Flag bflag) const {
 // Don't generate documentation for this symbol, either because it is private,
 // or because the symbol should not be documented independent of privacy
 bool Symbol::noDocGen() const {
-  return hasFlag(FLAG_NO_DOC) || hasFlag(FLAG_PRIVATE);
+  return hasFlag(FLAG_NO_DOC) || hasFlag(FLAG_PRIVATE) ||
+    hasFlag(FLAG_COMPILER_GENERATED);
```
mppf: I'd probably indent this differently but it'd be OK with me to leave it as is.

test/chpldoc/classes/forwards.doc.future
```diff
@@ -0,0 +1,2 @@
+feature request: documentation for methods obtained through forwarding
+#9801
```
mppf: newline to add

test/chpldoc/classes/forwards.doc.chpl
```diff
@@ -0,0 +1,14 @@
+/* This class should get the forwarded methods from B */
```
mppf: I'd consider making a version of this test where the field f is generic & calling the current output the .good (i.e. not a future).

Issue #9552, opened 2018-05-15
Convert DateTime module to use error handling


benjamin-robbins:
Convert halts in DateTime module to throw'ing functions (where appropriate)

ronawho:
https://github.com/chapel-lang/chapel/pull/9832 removed or converted some halts that are no longer necessary. The harder part of actually converting some cases to error-handling remains.

ronawho:
https://github.com/chapel-lang/chapel/pull/9944 does most of the conversion. There are some remaining halts, but what we do with those depends on https://github.com/chapel-lang/chapel/issues/9941, so I think this is done for now.

PR #9832, opened 2018-06-12
Remove some halts from the DateTime module

ronawho:
Remove or convert some unnecessary halts in Datetime: 
 - convert a halt for gettimeofday to an assert since it should never be able to fail
 - remove some "unreachable" halts that are no longer needed
 - convert some ad-hoc "Abstract base method called" halts to a pureVirtualMethodHalt() wrapper

Related to https://github.com/chapel-lang/chapel/issues/9552


ronawho:
@daviditen mind reviewing? Individual commits have more details, but this should be pretty straightforward

PR #9852, opened 2018-06-12
Update undecorated class warning test for initializers

benharsh:
Add second compilation configuration with slightly different output under initializers.


test/compflags/ferguson/unstable-class.compopts
```diff
@@ -1 +1,2 @@
---warn-unstable
+--warn-unstable # unstable-class.good
```
mppf: Will updating adding --no-force-initializers to this line avoid a failure with --force-initializers?

PR #9855, opened 2018-06-12
Add --break-on-id and --break-on-delete-id to chpldoc

### LINKED ISSUE #9823: Allow --break-on-id in chpldoc compilation
We allow some debugging flags for chpldoc compilation (like --gdb, --lldb) but we don't allow --break-on-id, which would be useful in that mode.

lydia-duncan:
Resolves #9823 

When debugging an issue with `chpldoc` recently, I realized that I could debug
its operations, but couldn't use `--break-on-id` (a developer flag with normal
chpl compilation that I have found quite helpful when I know the id to use)

This change makes it so that `chpldoc` allows `--break-on-id` and
`--break-on-delete-id` to work, by defining them through a macro that is used
both in the normal driver and in the `chpldoc` driver.  I figured
`--break-on-delete-id` was also useful, but that anything that would apply after
(like `--break-on-resolve-id` or `--break-on-codegen-id`) was absolutely not going
to be necessary.

Double checked on a local `--lldb` run for both flags both in `chpldoc` and `chpl`
compilation.  Passed a full paratest (I know there is a help output test but I don't
remember where it lives)

### DIFF
--- a/compiler/main/commonFlags.h
+++ b/compiler/main/commonFlags.h
@@ -43,7 +43,7 @@ void driverSetDevelSettings(const ArgumentDescription* desc, const char* arg_unu
 
 #define DRIVER_ARG_BREAKFLAGS_COMMON \
   {"break-on-id", ' ', NULL, "Break when AST id is created", "I", &breakOnID, "CHPL_BREAK_ON_ID", NULL}, \
- {"break-on-delete-id", ' ', NULL, "Break when AST id is deleted", "I", &breakOnDeleteID, "CHPL_BREAK_ON_DELETE_ID", NULL}
+  {"break-on-delete-id", ' ', NULL, "Break when AST id is deleted", "I", &breakOnDeleteID, "CHPL_BREAK_ON_DELETE_ID", NULL}
 
 #define DRIVER_ARG_DEBUGGERS                                            \
   {"gdb", ' ', NULL, "Run compiler in gdb", "F", &fRungdb, NULL, NULL}, \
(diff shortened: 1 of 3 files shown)


lydia-duncan:
Passed a full paratest.  Seems like we don't have a help output test for chpldoc, probably should do that (coulda sworn Thomas added one...)

lydia-duncan:
Ah, right, those are focused on the user experience and these are developer flags

compiler/main/commonFlags.h
```diff
@@ -41,6 +41,10 @@ void driverSetDevelSettings(const ArgumentDescription* desc, const char* arg_unu
 #define DRIVER_ARG_COPYRIGHT \
   {"copyright", ' ', NULL, "Show copyright", "F", &fPrintCopyright, NULL, NULL}
 
+#define DRIVER_ARG_BREAKFLAGS_COMMON \
```
daviditen: You might consider lining up the two backslashes in this macro like they are in the next macro.  The downside is that it would push both lines past 80 characters instead of just one.

compiler/main/commonFlags.h
```diff
@@ -41,6 +41,10 @@ void driverSetDevelSettings(const ArgumentDescription* desc, const char* arg_unu
...
+  {"break-on-id", ' ', NULL, "Break when AST id is created", "I", &breakOnID, "CHPL_BREAK_ON_ID", NULL}, \
+ {"break-on-delete-id", ' ', NULL, "Break when AST id is deleted", "I", &breakOnDeleteID, "CHPL_BREAK_ON_DELETE_ID", NULL}
```
daviditen: This line is missing a space at the beginning to line up correctly.

daviditen:
I'm guessing that the help tests you were looking for are in `test/compflags/thomasvandoren/chpldoc/`.

lydia-duncan:
Thanks!

PR #9858, opened 2018-06-13
Fix handful of --force-initializers failures from recent changes

benharsh:
This PR fixes a few --force-initializers failure cases that were introduced in recent weeks:
- Skip init/postinit override errors
- Fix field type queries for types with initializers
- Do not let --force-initializers take precedence over types with constructors
- Workaround for case in which AggregateType::defaultInitializer is NULL

Testing:
- [x] local + futures
- [x] partial gasnet
- [x] local --force-initializers ( less than 100 failures )

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -2184,7 +2184,7 @@ bool AggregateType::needsConstructor() const {
     // Defining an initializer means we don't need a default constructor
     return false;
   } else {
-    // The above two branches are only relevant in the recursive version
+    // The above three branches are only relevant in the recursive version
     // of this call, as the outside call site for this function has
     // already ensured that the type which is the entry point has defined
     // neither an initializer nor a constructor.
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -2173,6 +2173,8 @@ bool AggregateType::needsConstructor() const {
   // and library modules
   if (mod && (mod->modTag == MOD_INTERNAL || mod->modTag == MOD_STANDARD))
     return true;
+  else if (initializerStyle == DEFINES_CONSTRUCTOR)
+    return true;
   else if (fUserDefaultInitializers)
     // Don't generate a default constructor when --force-initializers is true,
     // we want to generate a default initializer or fail.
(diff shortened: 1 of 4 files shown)


benharsh:
@mppf : Would you be willing to review the changes to preFold.cpp?
@lydia-duncan : Could you review the rest of the changes?

compiler/AST/AggregateType.cpp
```diff
@@ -2181,9 +2183,6 @@ bool AggregateType::needsConstructor() const {
   if (initializerStyle == DEFINES_INITIALIZER) {
     // Defining an initializer means we don't need a default constructor
     return false;
-  } else if (initializerStyle == DEFINES_CONSTRUCTOR) {
-    // Defining a constructor means we need a default constructor
-    return true;
   } else {
     // The above two branches are only relevant in the recursive version
```
lydia-duncan: You will probably want to update this comment as well

compiler/resolution/preFold.cpp
```diff
@@ -1520,11 +1520,26 @@ static bool isInstantiatedField(Symbol* field) {
   AggregateType* at     = toAggregateType(ts->type);
   bool           retval = false;
 
-  for_formals(formal, at->typeConstructor) {
-    if (strcmp(field->name, formal->name) == 0) {
-      if (formal->hasFlag(FLAG_TYPE_VARIABLE) == true) {
-        retval = true;
-        break;
+  // BHARSH INIT TODO: Sometimes the type constructor is not resolved when
+  // initializers are used. Why?
```
lydia-duncan: I believe we intentionally skip them when the type is concrete, or when we are working on an instantiation (and thus the type constructor on the generic base is the only one resolved), but that information might be out of date

compiler/AST/AggregateType.cpp
```diff
@@ -2181,9 +2183,6 @@ bool AggregateType::needsConstructor() const {
   if (initializerStyle == DEFINES_INITIALIZER) {
     // Defining an initializer means we don't need a default constructor
     return false;
-  } else if (initializerStyle == DEFINES_CONSTRUCTOR) {
-    // Defining a constructor means we need a default constructor
-    return true;
```
lydia-duncan: I'm trying to remember why I did this using this strategy.  I think it was because the original call to needsConstructor handled it on the base type or something to that effect - there might be a duplicate check wrapped around some calls to this function, which you could potentially now remove

PR #9864, opened 2018-06-14
Reduce output for unresolved 'init' calls

benharsh:
This PR introduces a new function, ``trimVisibleCandidates``, to the process of resolving calls. This function accepts list of FnSymbols and populates a list with the most applicable FnSymbols for the given call. This function currently only works for 'init' and '_new' calls, but could conceivably be expanded to support any method.

The core approach is to only accept init functions whose 'this' type is instantiated from the same generic type as the corresponding actual. Regular candidate-filtering will find the best match, if one exists.

If no such match exists, then the most applicable FnSymbols will be display to the user, instead of every single initializer in the program.

As a result, many prediffs can be removed from the testing system.

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -2725,13 +2725,13 @@ void trimVisibleCandidates(CallInfo&       info,
   CallExpr* call = info.call;
 
   bool isInit = false;
-  if (call->numActuals() >= 2 && call->isNamed("init")) {
+  if (call->numActuals() >= 2 && call->isNamedAstr(astrInit)) {
     if (SymExpr* se = toSymExpr(call->get(1))) {
       isInit = se->symbol() == gMethodToken;
     }
   }
 
-  bool isNew = call->numActuals() >= 1 && call->isNamed("_new");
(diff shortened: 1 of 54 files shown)


bradcray:
Now I just need a time machine so I can benefit from this for the past 3 weeks / 9 months of work...

benharsh:
@mppf : would you be willing to review?

compiler/resolution/functionResolution.cpp
```diff
@@ -2719,12 +2719,60 @@ static bool typeUsesForwarding(Type* t);
...
+  CallExpr* call = info.call;
+
+  bool isInit = false;
+  if (call->numActuals() >= 2 && call->isNamed("init")) {
+    if (SymExpr* se = toSymExpr(call->get(1))) {
+      isInit = se->symbol() == gMethodToken;
+    }
+  }
+
+  bool isNew = call->numActuals() >= 1 && call->isNamed("_new");
```
mppf: I wouldn't normally stress about call->isNamed but at this point in the compiler it might be worth avoiding a strcmp  and having e.g. isNamedAstr.

Issue #9835, opened 2018-06-12
Optimize blocking network atomic optimizations


ronawho:
Even for the blocking histogram version of bale we lag behind the reference version by ~20%. See what we can do to optimize blocking network atomics. Some potential ideas: 
 - [x] avoid duplicate `mreg_for_remote_addr()` calls (#9868)
 - [x] specialize`do_nic_amo()` for fetching vs non-fetching to avoid extra branches  (#9873)
 - [x] specialize `do_nic_amo()` for 4 vs 8 byte to avoid extra branches? (#9873)
 - [x] create a specialized `post_fma_and_wait_amo()` that doesn't yield with every attempt (#9876)

ronawho:
Closed by #9868, #9873, and #9876

ronawho:
This improved non-fetching atomics pretty significantly. ~15% improvement for many-to-one-amo,  ra-atomics, and naive bale histogram at 16 nodes. There are [graphs](https://chapel-lang.org/perf/16-node-xc/?startdate=2018/06/12&enddate=2018/06/25&graphs=hpccraatomicsperfgupsn233,remoteamoperformance) for many-to-one-amo and ra-atomics, but we hadn't started running bale yet when these improvements were merged: 

![image](https://user-images.githubusercontent.com/1588337/41865104-fe494a50-7860-11e8-9570-f341fa5783f3.png)

![image](https://user-images.githubusercontent.com/1588337/41865110-01b3d214-7861-11e8-8cdc-04914e8de63f.png)


At higher node counts this seemed to benefit even more. At 256 nodes we saw a 35% improvement for ra-atomics: 

![ra-perf](https://user-images.githubusercontent.com/1588337/41616050-c7c7a55c-73b1-11e8-9be9-93402c3e6ce4.png)

PR #9868, opened 2018-06-14
Avoid a remote memory region lookup for network atomics

ronawho:
Avoid a remote memory region lookup for network atomics

This passes the `remote_mr` to `do_nic_amo()` instead of having `do_nic_amo()`
perform the lookup itself. The remote address must be registered in order for
us to do a network atomic, which means we're already doing a lookup beforehand.
This just saves the result of the `mreg_for_remote_addr()` that we're doing
beforehand and passes it along to `do_nic_amo()`.

Roughly speaking this transforms:

```c
if (mreg_for_remote_addr(obj, loc) == NULL)
  do_fork_amo(...);
else
  do_nic_amo(...);  // was doing mreg_for_remote_addr() itself
```

into:

```c
mem_region_t* remote_mr;
if ((remote_mr = mreg_for_remote_addr(obj, loc)) == NULL)
  do_fork_amo(...);
else
  do_nic_amo(..., remote_mr); // _not_ doing mreg_for_remote_addr()
```

In practice, I don't expect there to be a noticeable performance difference
here because `mreg_for_remote_addr()` caches the most recent lookup and we're
just optimizing out a lookup that immediately follows the first one, so the
value is going to be in the cache. Still the cache lookup isn't free and
there's no reason to do a second lookup, so just avoid it.

Related to https://github.com/chapel-lang/chapel/issues/9835

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1486,7 +1486,7 @@ static void      do_nic_get(void*, c_nodeid_t, mem_region_t*,
                             void*, size_t, mem_region_t*);
 static int       amo_cmd_2_nic_op(fork_amo_cmd_t, int);
 static void      do_nic_amo(void*, void*, c_nodeid_t, void*, size_t,
-                            gni_fma_cmd_type_t, void*);
+                            gni_fma_cmd_type_t, void*, mem_region_t*);
 static void      amo_add_real32_cpu_cmpxchg(void*, void*, void*);
 static void      amo_add_real64_cpu_cmpxchg(void*, void*, void*);
 static void      fork_call_common(int, c_sublocid_t,
@@ -4105,7 +4105,8 @@ size_t do_amo_on_cpu(fork_amo_cmd_t cmd,
     //
     {
       chpl_bool32 my_res;
-      if (mreg_for_local_addr(obj) == NULL) {
+      mem_region_t* mr;
+      if ((mr = mreg_for_local_addr(obj)) == NULL) {
         my_res = atomic_compare_exchange_strong_int_least64_t
                    ((atomic_int_least64_t*) obj,
                     *(int_least64_t*) opnd1,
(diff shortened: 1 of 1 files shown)


gbtitus:
Just a preliminary comment with respect to "almost certainly still cached" in the description block: since the MR caching is per-thread and there's no opportunity for us to switch threads between the two lookups, it's definitely still cached.  This is nevertheless a worthwhile change, since even a lookup that hits the cache isn't free, but we should be clear that that's all we're saving.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6229,8 +6234,8 @@ int chpl_comm_addr_gettable(c_nodeid_t node, void* start, size_t len)
             return;                                                     \
           }                                                             \
                                                                         \
-          if (IS_32_BIT_AMO_ON_GEMINI(_t)                               \
-              || mreg_for_remote_addr(obj, loc) == NULL) {              \
+          remote_mr = mreg_for_remote_addr(obj, loc);                   \
+          if (IS_32_BIT_AMO_ON_GEMINI(_t) || remote_mr == NULL) {       \
```
gbtitus: Truly a nit, and apologies for not catching this in my first quick look, but the old code was written with the `||` the way it was specifically so that for 32-bit AMOs on Gemini we didn't do a remote MR lookup.  (There was no need for that lookup because Gemini doesn't support 32-bit AMOs and thus those have to be done by AMs.)  Of course we could recover that old behavior by recoding as:

```
if (IS_32_BIT_AMO_ON_GEMINI(_t)
    || (remote_mr = mreg_for_remote_addr(obj, loc)) == NULL) {
...
```
XE performance is certainly secondary at this point, but at the same time if we're willing to go to some effort to avoid small overheads on XC we probably ought to do so on other platforms as well, even old ones.

ronawho: Right, I considered noting that in the PR message. I was going to say something like "This optimizes the common/fast path, but it will actually result in an unnecessary remote memory lookup for 32 bit ops on gemini"

I can still note that, but I don't think we care that much because XE isn't that important to us anymore, and most atomics operations are going to be on 64-bit stuff anyways.

gbtitus: I think there are 7 more of these below, where the MR lookup call has now been hoisted above the 32-bit Gemini check that would render it superfluous.

gbtitus: Most AMOs are indeed 64-bit, but absent a pressing need (which textual simplification isn't) I don't see any reason to reduce current XE performance.  And Blue Waters and Titan are still running, so Chapel could end up there.

ronawho: Yeah, good point. I'll make that mod

ronawho:
Passed `release/examples/ runtime/configMatters/` testing on XC and XE. No noticeable performance differences for 16 node XC.

@gbtitus if you get a minute tomorrow, I'd appreciate a quick sanity check.

PR #9862, opened 2018-06-14
Get CPU counts from the runtime topology interface

gbtitus:
Change the authoritative runtime interface for CPU counts from the
"system" layer to the topology layer.  Implement new topology layer
functions that get CPU counts, and change most of the calls to the old
system-based functions so that they go to the new topology-based ones
instead.  (The only remaining calls to the system-based functions are in
the backstop topo=none topology implementation and chplsys itself.)
Rename the old system-based functions, to better indicate where they get
their information and to differentiate them from the new functions.

This resolves #9395.

PR #9869, opened 2018-06-14
Allow for hwloc_get_proc_cpubind() not supported on macOS.

gbtitus:
PR #9862 broke on macOS because hwloc_get_proc_cpubind() returns "not
supported" there.  Allow for that, basically ignoring CPU binding on
that platform.  This will at least get release/examples and the smoke
testing working again.  Afterward I'll look further into whether we
need to support the notion of accessible-vs-not on macOS.


gbtitus:
@ronawho would you take a look at this post-merge?

PR #9873, opened 2018-06-14
Specialize/optimize non-fetching network atomics

ronawho:
This adds a version of `do_nic_amo()` that's optimized for non-fetching
atomics. Compared to the fetching version, `do_nic_amo_nf()` eliminates 5
branches and is small enough to mark it as inline. If the compiler actually
inlines, it should be able to fold away sizeof branches as well.

I don't expect to see any noticeable performance improvements for things like
many-to-one-amo or RA-atomics. They do the same atomic operation in a pretty
tight loop, so all of these branches are probably correctly predicted. Still,
this seems worthwhile and I suspect we'll want to further specialize the
non-fetching case even more in the future.


ronawho:
Passed release/examples/ runtime/configMatters/ testing on XC and XE. No noticeable performance differences for 16 node XC.

@gbtitus if you get a minute tomorrow, I'd appreciate a quick sanity check.

Issue #9669, opened 2018-05-29
Start generating .h files for compiled libraries


lydia-duncan:
This should be reasonably straight-forward, but it is hard to say

lydia-duncan:
Remaining todos:
- [ ] Name should not have "lib" prefix if executable name does
- [ ] insert makefile rule to move header outside of generated code directory (with library)

lydia-duncan:
Ran into issues using the generated ".h" due to all the \_ref_* things we have defined

lydia-duncan:
Michael's suggestion appears to have fixed my problem.  I think I will need to remove a long list of .h files from our testing system, but I am very excited :)

lydia-duncan:
(will need to do something similar for c_ptr_ prefixes)

lydia-duncan:
Interesting, the .h file seems to have no tabs in it.  Not sure why that happened.

PR #9874, opened 2018-06-14
Allow exporting of functions with array arguments during compilation

lydia-duncan:
This change adds support for exporting functions with array arguments
during compilation.

When an exported function with such an argument is detected, create a
new wrapper that takes in an instance of the external array runtime type
instead and calls the conversion function to make a ExternArr instance,
which is then passed to the original function (now no longer exported).
If the original function returns a value, the wrapper will return that as
well.

Because the external arrays are now used in the compiler, make the
ExternalArray module included by default (and remove the use
statements in the original tests now that they are no longer necessary).

My new tests exercised something that hadn't be exercised before,
requiring the addition of the dsiNumIndices method to ExternDom.

Adds several new tests to lock in this functionality, including:
- checking a basic exported function that takes in an array argument
behaves as expected
- checking that having another argument after the array argument does
not cause problems
- checking that having multiple array arguments does not cause problems
- checking that declaring the return type explicitly as void does not
cause problems (the others have no declared return type, and don't return)
- checking that an array of reals is acceptable (other tests were arrays of
ints)
- checking that having another argument which is a c_ptr does not cause
problems
- validating my error message when the array is generic over its element
type
- checking that having another argument before the array argument does
not cause problems
- checking that having another argument of the chpl_external_array type
does not cause problems

Other testing changes:
- I removed the array case from test/functions/export/exportGenericArr, and
renamed it to reflect that it was covering generic arguments in general instead
of arrays.
- I updated test/modules/sungeun/init/printModuleInitOrder and
test/modules/vass/deinit-order-modules to reflect the inclusion of
ExternalArray by default

Tested my export tests with fifo and valgrind, running a full paratest now


lydia-duncan:
@mppf - this is ready for a look when you get back from your day off :)

compiler/passes/normalize.cpp
```diff
@@ -2616,6 +2625,63 @@ static void hack_resolve_types(ArgSymbol* arg) {
...
 }
 
+/************************************* | **************************************
+*                                                                             *
+* We cannot export a function with an array argument directly, due to the     *
+* type being considered generic and due to other languages not understanding  *
+* our array structure, as well as our normal array structure assuming it has  *
+* control over the memory involved.  Instead, build a wrapper for the         *
+* function.  This wrapper will be modified to take in a c_ptr of the right    *
+* type and the size of the array, in addition to its other arguments.         *
```
mppf: I thought the wrapper was using the runtime-defined array struct?

compiler/passes/normalize.cpp
```diff
@@ -2683,6 +2745,70 @@ static bool skipFixup(ArgSymbol* formal, Expr* domExpr, Expr* eltExpr) {
...
+
+      if (eltExpr == NULL) {
+        USR_FATAL(formal, "array argument '%s' in exported function '%s'"
+                  " must specify its type", formal->name, fn->name);
+        continue;
+      }
+
+      // Create a representation of the array argument that is accessible
+      // outside of Chapel (in the form of a pointer and a corresponding size)
+      formal->typeExpr->replace(new BlockStmt(new CallExpr("_type_construct_chpl_external_array")));
```
mppf: This is OK for now if it works but AFAIK a reasonable alternative would be to make chpl_external_array a "well known type" and then put here a SymExpr referring to its TypeSymbol. That seems "cleaner" to be but I understand such things are subjective. But, if it seems better to you too, it's probably worth doing.

compiler/passes/normalize.cpp
```diff
@@ -2683,6 +2745,70 @@ static bool skipFixup(ArgSymbol* formal, Expr* domExpr, Expr* eltExpr) {
...
+
+  INT_ASSERT(callOrigFn);
+
+  for_formals(formal, fn) {
+    if (isArrayFormal(formal)) {
+      BlockStmt*            typeExpr = formal->typeExpr;
+
+      CallExpr*             call     = toCallExpr(typeExpr->body.tail);
+      int                   nArgs    = call->numActuals();
+      Expr*                 eltExpr  = nArgs == 2 ? call->get(2) : NULL;
```
mppf: What is this call we might be working with? Why is argument 2 important? Needs at least some comments here about what it's trying to match.

test/compflags/lydia/library/exportArray/exportFuncWith1Arr1Not.h
```diff
@@ -0,0 +1,6 @@
+#include <stdint.h>
+
+#include "chpl-external-array.h"
+
+void foo(chpl_external_array* x, int64_t otherArg);
```
mppf: I'm surprised by the `*` here. I would have expected the struct to be passed directly.

test/compflags/lydia/library/exportArray/callExported2Arg.test.c
```diff
@@ -0,0 +1,26 @@
...
+
+// Test of calling an exported function that takes an array
+int main(int argc, char* argv[]) {
+  // Initialize the Chapel runtime and standard modules
+  chpl_library_init(argc, argv);
+
+  // Call the function
+  int64_t x[5] = {1, 2, 3, 4, 5};
+  chpl_external_array arrX = chpl_make_external_array_ptr(x, 5);
+  foo(&arrX, 7);
```
mppf: I'm surprised by the `&` here. I would have expected the struct to be passed directly (vs a pointer to it).

test/compflags/lydia/library/exportArray/returnExternArray.chpl
```diff
@@ -1,5 +1,5 @@
-use ExternalArray;
-
+// This is a mock up of what returning an array from an exported function looks
+// like under the covers
```
mppf: isn't it working? why the comment?

test/compflags/lydia/library/exportArray/exportFuncWithArrayArg.h
```diff
@@ -0,0 +1,6 @@
+#include <stdint.h>
+
+#include "chpl-external-array.h"
+
+void foo(chpl_external_array* x);
+void chpl__init_exportFuncWithArrayArg(int64_t ln, int32_t fn);
```
mppf: Is there an issue or something about the desire for the Chapel compiler to produce .h files for the exported stuff?

compiler/passes/normalize.cpp
lydia-duncan: This is what I get for writing comments early, I should be better about updating them

test/compflags/lydia/library/exportArray/exportFuncWith1Arr1Not.h
lydia-duncan: I believe this is due to the compiler seeing chpl_external_array is a record (albeit an extern one) and transforming the argument into a ref at resolution.  I can try to make it not do that for exported functions, but I assumed we wanted to keep that behavior.

compiler/passes/normalize.cpp
mppf: There's no judgement here, just pointing it out for fixing. I figured that was what happened & expect it will happen again (to any / all of us).

test/compflags/lydia/library/exportArray/returnExternArray.chpl
lydia-duncan: This was a test from my initial implementation, to make sure I could use the functions I made the way I thought I could.  I decided to leave the test around, but I could see removing them now with the compiler support as a reasonable course of action.  This test would have to wait for the returning implementation, though

test/compflags/lydia/library/exportArray/exportFuncWithArrayArg.h
lydia-duncan: Yeah, #9669

lydia-duncan: It's going to be real nice when I get to implementing that

compiler/passes/normalize.cpp
lydia-duncan: Thanks :)

test/compflags/lydia/library/exportArray/exportFuncWith1Arr1Not.h
mppf: I don't think the compiler should be doing this for extern types and certainly not for extern types in exported functions. In resolveIntents, concreteIntentForArg has a check for fn having FLAG_EXTERN. What if you add FLAG_EXPORT to that check?

lydia-duncan: Looks like that gets honored all the way until codegen, at which case it gets thrown out.  I'll look into why codegen is doing that

mppf: at a guess, have a look at ArgSymbol::requiresCPtr

lydia-duncan: I tried updating the return check at the end of `shouldChangeArgumentTypeToRef` (where that method is called) similarly to the resolveIntents change.  This resulted in a lot of problems at makeBinary due to I believe some calls defined in the runtime (I see chpl_main_argument and chpl_localeID_t involved, mostly the latter).

The chpl_localeID_t errors go away if I instead update argMustUseCPtr to check if the record is defined as extern, but the chpl_main_argument errors still remain - I'm not as worried when it is just that which gets affected, though, so I'm going to see if it compiles and runs when I update the uses of chpl_main_argument.

Lemme know your thoughts on this

mppf: I think we should be passing around chpl_localeID_t by value where possible so I think we should be passing them by value.

lydia-duncan: Modifying the use of chpl_main_argument, I was able to get my test compiling and running without a segfault.

lydia-duncan: > I think we should be passing around chpl_localeID_t by value where possible too.

So do you think the modification for `shouldChangeArgumentTypeToRef` (keying off of if the function is exported, rather than if the type is an extern record) is the better call?

mppf: Let's leave these changes out of this PR and put them in a new PR. They probably will need discussion from others too, so best to try to separate them.

lydia-duncan: I opened issue #9898 to track this.  I'll send it to the larger group once you've had a chance to look at it

lydia-duncan:
Passed testing again.  @mppf - up for a second look?

PR #9875, opened 2018-06-14
Mason commands from subdirectory

Spartee:
Almost all of the Mason commands were partially re-written in this PR so that they can be called from any place within a mason project. Mason now finds the top level directory by traversing up through the project directory until it encounters a `Mason.toml`. A few workarounds for testing were included that allow the search for a `Mason.toml` to be overwritten. 

Also included with this PR is a slight change to the project structure. A tests directory is included with the `mason new` command as well as a `/target/test` directory to hold test binaries with the `mason build` command.


ben-albrecht:
Do you think we could add a test for this new functionality?

tools/mason/mason.chpl
```diff
@@ -96,27 +96,40 @@ proc main(args: [] string) {
...
-      writeln(command);
-      runCommand(command);
+  try! {
+    const cwd = getEnv("PWD");
+    const projectHome = getTopLvlDirPath(cwd);
+    const toDoc = basename(projectHome);
+    const project = toDoc + '.chpl';
+    if isDir(projectHome + '/src/') {
+      if isFile(projectHome + '/src/' + project) {
+        const command = 'chpldoc ' + projectHome + '/src/' + project + ' -o ' + projectHome + '/doc/';
```
ben-albrecht: It would also be nice to test this functionality if that does not already exist

tools/mason/mason.chpl
```diff
@@ -96,27 +96,40 @@ proc main(args: [] string) {
 
 
 proc masonClean() {
-  runCommand('rm -rf target');
+  try! {
+    const cwd = getEnv("PWD");
+    const projectHome = getTopLvlDirPath(cwd);
```
ben-albrecht: Naming-wise, I find something like `getProjectHome()` easier to remember, as demonstrated by the choice of variable name here :)

tools/mason/MasonUpdate.chpl
```diff
@@ -42,35 +42,55 @@ private var failedChapelVersion : [1..0] string;
...
   }
-  else writeln("Cannot update: no Mason.toml found");
 }
 
 
 /* Writes out the lock file */
 proc genLock(lock: Toml, lf) {
-  var lockFile = open(lf, iomode.cw);
-  var tomlWriter = lockFile.writer();
+  const lockFile = open(lf, iomode.cw);
```
ben-albrecht: Should we also close this `lockFile`?

tools/mason/MasonBuild.chpl
```diff
@@ -63,62 +63,81 @@ private proc checkChplVersion(lockFile : Toml) {
...
 
-    // Install dependencies into $MASON_HOME/src
-    var toParse = open("Mason.lock", iomode.r);
-    var lockFile = parseToml(toParse);
-    checkChplVersion(lockFile);
-
-    if isDir(MASON_HOME) == false {
-      mkdir(MASON_HOME, parents=true);
+    // workaround for testing purposes
+    if isFile("Mason.lock") {
```
ben-albrecht: Could you describe this to me a bit more?

ben-albrecht:
When doing a `mason build` outside of a mason package, I get the error message twice:

```
Mason could not find your configuration file (Mason.toml)
Mason could not find your configuration file (Mason.toml)
```

I assume this is from the implicit `mason update`. Is there a way we could prevent any further commands if the first command fails to find the `Mason.toml`?

tools/mason/MasonUpdate.chpl
```diff
@@ -41,38 +41,56 @@ private var failedChapelVersion : [1..0] string;
 
 /* Finds a Mason.toml file and updates the Mason.lock
    generating one if it doesnt exist */
-proc UpdateLock(args: [] string, tf="Mason.toml", lf="Mason.lock") {
-  if isFile(tf) {
+proc UpdateLock(args: [] string, tf="Mason.toml", lf="Mason.lock", isTest=false) {
```
ben-albrecht: I'd prefer to keep testing logic out of the interface (even though it's not user-facing). I just sent you a patch for the mason tests that I think will eliminate the need for this argument.

tools/mason/MasonRun.chpl
```diff
@@ -24,56 +24,66 @@ use FileSystem;
...
+        masonBuild(args);
+        runCommand(command);
+      }
+      else writeln("Mason could not find your Mason.lock file");
 
-    // Close memory
-    delete tomlFile;
-    toParse.close();
+      // Close memory
+      delete tomlFile;
```
ben-albrecht: Bonus points if you apply the new delete-free memory management here. 

I believe that would involve changing the above assignment to:

```chpl
const tomlFile = new Owned(parseToml(toParse));
```

and then you can remove the `delete`here.

(@mppf can spot check me on that)

tools/mason/MasonBuild.chpl
```diff
@@ -62,63 +64,74 @@ private proc checkChplVersion(lockFile : Toml) {
...
+proc makeTargetFiles(binLoc: string, projectHome: string) {
+
+  if !isDir(projectHome + '/target') {
+    mkdir(projectHome + '/target');
   }
-  if !isDir('target/' + binLoc) {
-    mkdir('target/' + binLoc);
+  if !isDir(projectHome + '/target/' + binLoc) {
+    mkdir(projectHome + '/target/' + binLoc);
+    mkdir(projectHome + '/target/test');
```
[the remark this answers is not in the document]
Spartee: You are correct.

tools/mason/MasonUpdate.chpl
Spartee: applied

tools/mason/mason.chpl
Spartee: this goes in line with something I was going to add to the epic now that we have better error handling which is a more sophisticated error class with better responses such as the one you just mentioned.

PR #9876, opened 2018-06-14
Optimize non-fetching network atomics by yielding less while waiting for ACK

ronawho:
Improve the performance of non-fetching atomics by yielding less frequently
while waiting for a remote ACK.

At the runtime level for a non-fetching network atomics we do something like:

```c
do_nic_amo_nf(void* opnd, c_nodeid_t locale, void* obj) {
  // setup post descriptor
  post_desc.type          = GNI_POST_AMO;
  post_desc.remote_addr   = obj;
  post_desc.first_operand = opnd;
  post_desc.post_id       = &post_done;

  // post to the NIC (initiate the remote operation)
  cdi = post_fma(locale, &post_desc);

  // wait for ACK from remote node
  do {
    chpl_task_yield();                      // yield every iter
    consume_all_outstanding_cq_events(cdi);
  } while(!atomic_load_bool(&post_done))
}
```

Non-fetching network atomics will be extremely fast, and we're really just
waiting for an ACK back. So here we optimize how we wait for that ACK by only
yielding every 64 CQ read attempts instead of after every one. So now we do
something like:

```c
uint64_t iters = 0;
do {
  if ((iters & 0x3F) == 0)                // yield initially, then 1/64 iters
    chpl_task_yield();
  consume_all_outstanding_cq_events(cdi);
  iters++;
} while(!atomic_load_bool(&post_done))
```

Note that we're still yielding initially, but after that we only yield every 64
attempts. I believe this means that a no-op chpl_task_yield takes longer than
we expect and we'll probably want to optimize/investigate that later.

Related to https://github.com/chapel-lang/chapel/issues/9835

This has a pretty significant impact on on non-fetching atomics. This improves
the many-to-one-amo microbenchmark by 15% taking it from from ~210 Mops/s to
~240 Mops/s. This also improves the bale histogram benchmark by ~15%. Lastly,
this improves RA-atomics by 25-35% depending on the node count. At 256 nodes
this takes us from 2 GUPS to 2.7 GUPS:

ra-perf

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1509,6 +1509,8 @@ static chpl_bool reacquire_comm_dom(int);
 static int       post_fma(c_nodeid_t, gni_post_descriptor_t*);
 static void      post_fma_and_wait(c_nodeid_t, gni_post_descriptor_t*,
                                    chpl_bool);
+static void      post_fma_and_wait_amo(c_nodeid_t, gni_post_descriptor_t*,
+                                       chpl_bool);
 #if HAVE_GNI_FMA_CHAIN_TRANSACTIONS
 static int       post_fma_ct(c_nodeid_t*, gni_post_descriptor_t*);
 static void      post_fma_ct_and_wait(c_nodeid_t*, gni_post_descriptor_t*);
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus changes since you reviewed are switching back to yielding every iteration for fetching atomics. I was driving myself crazy trying to tune that without much luck, so I'm just going to punt on that for now. I also changed from yielding every 32 iters to every 64.

Issue #9811, opened 2018-06-11
Support LayoutCS( sorted : bool )


ian-bertolacci:
Per #9810 need to have LayoutCS take a boolean option indicating if the sparse dimension is sorted, and supporting implementation of optional sorting.

bradcray:
Thinking about your param bool loop over on issue #9860 this morning...  while I agree that param loops over bool ranges are reasonable to support, I also found myself wondering whether I was wrong in suggesting that this sorted flag should be a param.  Is it a case where compile-time code specialization is likely to significantly impact performance on critical paths?  Or is it a case where an execution-time check would be sufficient?  (A potential secondary consideration being code size:  If a program uses only sorted or unsorted sparse domains, is the code size significantly smaller?  If it uses both, is the code size significantly bigger?)

ian-bertolacci:
Are you anticipating that people will want to decide in execution-time if they want sorting on?
My intuition says that any additional overhead to index lookup would be bad, but then again it's only one if statement per lookup.

bradcray:
> Are you anticipating that people will want to decide in execution-time if they want sorting on?

Mmmmaybe.  Or, they may want to have the opportunity to turn sorting on and off over the lifetime of a sparse domain, maybe?  I don't feel confident about either of these things, though... was just second-guessing whether my instinct to make it a `param` was overkill.

> My intuition says that any additional overhead to index lookup would be bad, but then again it's only one if statement per lookup.

If index lookup were O(1), I'd agree that we might want to avoid a conditional.  But my guess was that the index lookup case (and others where this param would come into play) already probably have so many conditionals and loops that one more outer-scope conditional would hardly be noticeable.

In no way am I insisting on anything here, though, just exploring the question.

ian-bertolacci:
Lemme get some numbers.

bradcray:
You're also welcome to just file this away as food for thought.  No need to let it stand in the way of progress (it mostly came up for me as I was thinking about ways you might work around the lack of param bool for loops before learning you had a workaround already).

ian-bertolacci:
Closed with #9877

PR #9877, opened 2018-06-14
Layout cs mixed sorted

ian-bertolacci:
Implements #9811


ian-bertolacci:
@bradcray would you like to take a look?

ian-bertolacci:
Should I also include some performance tests?
I have a suite I could put it if we think it would be useful.

bradcray:
I think it never hurts to track performance of features you care about...

modules/layouts/LayoutCS.chpl
```diff
@@ -255,10 +257,21 @@ class CSDom: BaseSparseDomImpl {
...
-    if this.compressRows then
-      ret = binarySearch(idx, col, lo=startIdx(row), hi=stopIdx(row));
-    else
-      ret = binarySearch(idx, row, lo=startIdx(col), hi=stopIdx(col));
+    if this.compressRows {
+      if this.sorted then
+        ret = binarySearch(idx, col, lo=startIdx(row), hi=stopIdx(row));
+      else {
+        ret = linearSearch(idx, col, lo=startIdx(row), hi=stopIdx(row));
+        if !ret[1] then ret[2] = stopIdx(row)+1;
```
bradcray: If linearSearch() returned `hi+1` rather than `Dom.high+1`, would the extra conditional required for linearSearch() over binarySearch() no longer be required?  (for the purpose of symmetry; also because it seems a shame that that routine takes that approach).

ian-bertolacci: Yes that would be nice. I think that would also fit well with how BinarySearch works.

test/sparse/CS/correctness.chpl
```diff
@@ -32,192 +36,323 @@ proc main() {
...
-    csrD += [(1,2), (2,1), (4,5), (5,4)];
-    cscD += [(1,2), (2,1), (4,5), (5,4)];
+    assert(csrsD.first == (0,0));
+    assert(cscsD.first == (0,0));
+    assert(csrsD.last == (0,0));
+    assert(cscsD.last == (0,0));
+    assert(csruD.first == (0,0));
+    assert(cscuD.first == (0,0));
+    assert(csruD.last == (0,0));
+    assert(cscuD.last == (0,0));
```
bradcray: At a glance, it seems like this test would be far more maintainable (and much easier to reason about / ensure correctness for) if it had a generic "test sparse domain" routine that it passed each of the four sparse domains into rather than repeating each operation x the 4 types.  For inspiration, see test/arrays/userAPI/arrayAPItest.chpl and arrayOps2D.chpl (though there's no reason to break it into multiple files necessarily).

bradcray:
The PR's description should include more detail about what is being done in the PR, the approach taken, etc.  See other open PRs by members of the core team for examples (maybe don't consider my current open ones which are abnormally large).

bradcray:
I think I might see why the linear search is so expensive.  I'd collapse these lines in the linearsearch routine:

```chapel
    const r = if Dom.stridable then lo..hi by stride else lo..hi;
    for i in Dom[r] {
```

into simply:

```chapel
    const r = if Dom.stridable then lo..hi by stride else lo..hi;
    for i in r do
```

and see where that gets you.  Or perhaps even:

```chapel
    if Dom.stridable then
     for i in lo..hi by stride do
       ...
    else
      for i in lo..hi do
```

specifically, the creation of a domain slice is pretty heavyweight (and unnecessary in this case, as far as I can see), and for loops over anonymous ranges (as in my second rewrite) tend to generate the cleanest code.

bradcray:
I continue to wonder if we want "unsorted" by default, though the answer probably depends on the performance we get with cg-sparse.chpl for large problem sizes and for sparse mat-mat multiplication once it stops sorting.

bradcray:
@ben-albrecht : Tagging you on this PR because we're discussing changes to LinearSearch for both performance and symmetry with binarySearch.

ian-bertolacci:
I ran my performance test over the weekend, and it still doesnt look good, particularly accessing an element by index (iterating over the values of the array is still fast).

For example, I test with 3 sparse arrays all 10^10 elements size with 1% sparsity (10^8 nonzero elements): one that is 10^7 rows by 10^3 columns (~10 non-zeros per row), one that is 10^5 rows by 10^5 columns (~10^3 non-zeros per row), and one that is 10^3 rows by 10^7 columns (~10^5 non-zeros per row).
The indices being inserted to are random, but the values are inserted into the array in the same order that they inserted to in the domain (I also test with randomized order, but I won't talk about that here).
 
Here are the times in seconds:

NNZ/row | Unsorted | Sorted
-- | -- | --
10 | 0.897782 | 0.776321
1000 | 30.8494 | 3.67239
100000 | 2849.25 | 4.69138

As you can see, the performance for the unsorted case drops like a rock as the number of non-zeros per row increases.
I'm testing some more cases (10^6 by 10^4 and 10^4 by 10^6) right now to fill in the gaps.

bradcray:
OK.  Do we have any data on cg-sparse.chpl or the sparse mat-mat multiply yet?

ian-bertolacci:
Not yet. Lemme spin those up.

ian-bertolacci:
Current runtimes with spots filled in

Total Array Elements | 10,000,000,000.00 |   |   |  
-- | -- | -- | -- | --
Number Non-Zero | 100,000,000.00 | ArrayInsertByIndex |   |  
Rows | Number Non-Zeros/row | Unsorted | Sorted | Slowdown
10,000,000 | 10 | 0.897782 | 0.776321 | 1.15645719
1,000,000 | 100 | 9.55966 | 2.35712 | 4.05565266
100,000 | 1,000 | 30.8494 | 3.67239 | 8.40036053
10,000 | 10,000 | 290.433 | 4.13924 | 70.1657792
1,000 | 100,000 | 2849.25 | 4.69138 | 607.337287

ian-bertolacci:
I finally have numbers on LinearAlgebra and cg. This is with the sortIndices in LinearAlgebra enabled only if the array is sorted.

Test | Sorted | Unsorted | Speedup | Change | Status
-- | -- | -- | -- | -- | --
cg.A | 0.575104 | 1.6721 | 0.343941152 | -1.096996 | Worse
cg.S | 0.085324 | 0.102359 | 0.833575943 | -0.017035 | Worse
cg.W | 0.224095 | 0.430677 | 0.520331942 | -0.206582 | Worse
csrmatmat-1e-1-small | 0.179694 | 0.044292 | 4.057030615 | 0.135402 | Better
csrmatmat-1e-2-small | 0.0044378 | 0.001478 | 3.002571042 | 0.0029598 | Better
csrmatmat-1e-3-small | 0.00013327 | 9.20E-05 | 1.448114745 | 0.00004124 | Better
csrmatmat-1e-4-small | 4.45E-05 | 4.26E-05 | 1.046292599 | 1.971E-06 | Better
csrmatmat-1e-7-big | 0.017638 | 0.015803 | 1.116117193 | 0.001835 | Better
csrmatmat-1e-8-big | 0.010856 | 0.009291 | 1.168442579 | 0.001565 | Better
csrmatmat-1e-9-big | 0.010283 | 0.008663 | 1.187002193 | 0.00162 | Better
m10 | 1.42E-05 | 1.41E-05 | 1.006247781 | 8.8E-08 | Better
m1000 | 0.0005815 | 0.0005764 | 1.008848022 | 5.1E-06 | Better
m10000 | 0.181639 | 0.188282 | 0.964717817 | -0.006643 | Worse
spa-1e-1-small | 0.1374 | 0.137107 | 1.002137017 | 0.000293 | Better
spa-1e-2-small | 0.0329383 | 0.032982 | 0.998675035 | -4.37E-05 | Worse
spa-1e-3-small | 0.0207201 | 0.0206019 | 1.005737335 | 0.0001182 | Better
spa-1e-4-small | 0.0189558 | 0.0189563 | 0.999973624 | -5E-07 | Worse
spa-1e-7-big | 38.4958 | 38.8386 | 0.991173729 | -0.3428 | Worse
spa-1e-9-big | 38.9277 | 38.7827 | 1.00373878 | 0.145 | Better

All cg tests took a hit, but some of the matmul tests did better, especially for the small matrices. SPA seem seems to hover around unchanged.

bradcray:
Do your matmul results avoid the call to the Sort() routine within the matmul routine when sorting isn't necessary?

ian-bertolacci:
yes. I modified that code to be:
```chpl
if C.domain._value.sorted {
  sortIndices(C);
}
```

bradcray:
OK, thanks.  @ben-albrecht, do these results look like they match what you scoped out earlier this year?

bradcray:
I may be missing something, but it looks likely that, if the bulkAdd() interface is used on an unsorted CS domain, we'd still sort the indices.  (I'm looking at the fact that the calls to bulkAdd_prepareInds() don't seem to have changed, nor has its definition).

modules/layouts/LayoutCS.chpl
ian-bertolacci: I've made this it's own issue (#9908).
Also, instead of hi+1, I'm going with hi+stride (current behavior is Dom.high+stride)

ben-albrecht:
> do these results look like they match what you scoped out earlier this year?

Yes, the sparse matrix-matrix multiplication improvements look consistent with what I observed previously.

ben-albrecht:
> yes. I modified that code to be:

```chpl
if C.domain._value.sorted {
  sortIndices(C);
}
```

I think you can drop the `._value` in that call chain now that domains and arrays forward their values:

```chpl
if C.domain.sorted {
  sortIndices(C);
}
```

ian-bertolacci:
@ben-albrecht when I remove `_value` I get the compile error:
```
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1411: In function '_csrmatmatMult':
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1431: error: ambiguous call 'CSDom(2,int(64),domain(2,int(64),false),true,true,false).sorted'
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note: candidates are: _domain.sorted
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note:                 _domain.sorted
```

ben-albrecht:
@ian-bertolacci  - I see.. This seems to be a name conflict with the existing [`domain.sorted`](https://chapel-lang.org/docs/master/builtins/internal/ChapelArray.html#ChapelArray.sorted) iterator.

It might be good design to avoid this conflict by using a different field name e.g. `isSorted`. I believe we ran into something similar in the past, which led to the variable named `dataSorted` elsewhere.

ian-bertolacci:
@ben-albrecht Oh oops. I'll change my sorted to sortedIndices.

ian-bertolacci:
Hm. I'm still having an issue not using `_value` if I change `sorted` (newly`sortedIndices`) to the completely unused symbol `sortedIndices_bertolacci_2018` I still get an error

```
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1411: In function '_csrmatmatMult':
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1430: error: ambiguous call 'CSDom(2,int(64),domain(2,int(64),false),true,true,false).sortedIndices_bertolacci_2018'
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note: candidates are: _domain.sortedIndices_bertolacci_2018
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note:                 _domain.sortedIndices_bertolacci_2018
```

bradcray:
@mppf (or @benharsh): Do either of you guys have insights into Ian's forwarding challenge just above?

benharsh:
Not really, it seems to be a general issue with forwarding param fields. I have opened #9926 to track this bug.

ian-bertolacci:
I've rerun the performance numbers:
Test | Sorted | Unsorted_6.21 | Speedup | Decrease | Status
-- | -- | -- | -- | -- | --
cg.A | 0.575104 | 1.24865 | 0.460580627 | -0.673546 | Worse
cg.S | 0.085324 | 0.09412 | 0.906544836 | -0.008796 | Worse
cg.W | 0.224095 | 0.345449 | 0.648706466 | -0.121354 | Worse
csrmatmat-1e-1-small | 0.179694 | 0.044629 | 4.026395393 | 0.135065 | Better
csrmatmat-1e-2-small | 0.0044378 | 0.0014802 | 2.998108364 | 0.0029576 | Better
csrmatmat-1e-3-small | 0.00013327 | 9.08E-05 | 1.467246504 | 0.00004244 | Better
csrmatmat-1e-4-small | 4.45E-05 | 4.25E-05 | 1.049323974 | 2.094E-06 | Better
csrmatmat-1e-7-big | 0.017638 | 0.016093 | 1.096004474 | 0.001545 | Better
csrmatmat-1e-8-big | 0.010856 | 0.009398 | 1.155139391 | 0.001458 | Better
csrmatmat-1e-9-big | 0.010283 | 0.009035 | 1.138129496 | 0.001248 | Better
m10 | 1.42E-05 | 1.44E-05 | 0.98464638 | -2.21E-07 | Worse
m1000 | 0.0005815 | 0.0005654 | 1.028475416 | 0.0000161 | Better
m10000 | 0.181639 | 0.18332 | 0.990830242 | -0.001681 | No Change
spa-1e-1-small | 0.1374 | 0.139914 | 0.98203182 | -0.002514 | Worse
spa-1e-2-small | 0.0329383 | 0.0342044 | 0.962984294 | -0.0012661 | Worse
spa-1e-3-small | 0.0207201 | 0.0206051 | 1.005581143 | 0.000115 | No Change
spa-1e-4-small | 0.0189558 | 0.0190462 | 0.995253646 | -9.04E-05 | No Change
spa-1e-7-big | 38.4958 | 39.3505 | 0.978279819 | -0.8547 | Worse
spa-1e-9-big | 38.9277 | 40.9261 | 0.951170524 | -1.9984 | Worse

I consider anything with a speedup of 1±0.0119 to be noise (its not a very good noise measure)
This is basically the same performance we were seeing  before.

I think this is probably good to be pulled in.

ben-albrecht:
> we discussed having a config param in the DefaultSparse file to indicate whether the default should be sorted or unsorted such that a user / tester could flip the default for an entire program at compile-time. This seems valuable and unlikely to cause problems since it should only affect performance, not correctness / behavior.

That said, I think we'll want CSR arrays generated by the `LinearAlgebra` library to be unsorted by default for the matrix-multiplication performance benefits.

ian-bertolacci:
I've finally confirmed that all correctness tests pass.

PR #9886, opened 2018-06-15
Allow basic usage of nested types with initializers

### LINKED ISSUE #9353: Support nested classes and records with initializers
As a Chapel user, I want to use nested classes and records in order to hide implementation details and prevent namespace pollution.

Some very basic nested types work today with initializers, but things quickly fall apart if one uses default initializers and/or generics involving outer fields.

Acceptance criteria: The following tests pass with --force-initializers:
- [ ] classes/deitz/nested/test_nested{1,2,3}
- [ ] users/ferguson/nested/classclass
- [ ] users/ferguson/nested/classrecord{1,2,3}
- [ ] users/ferguson/nested/recordclass
- [ ] users/ferguson/nested/recordrecord{1,2,3}
- [ ] studies/uts/*
- [ ] classes/initializers/generics
    - [ ] nested
    - [ ] nested2
    - [ ] nestedInnerGeneric
    - [ ] phase1
        - [ ] nested-function-mods-field{1,2}
        - [ ] nested-function-phase-2-mods-fields
- [ ] release/examples/programs/genericStack
- [ ] release/examples/programs/linkedList
- [ ] classes/bradc/innerGeneric/*
- [ ]   studies/amr/advection/* (5-6 tests)

benharsh:
This PR updates the compiler to allow usage of nested types with initializers, but completely disables using fields/methods from the enclosing type. In other words, 'outer' is gone. In the future it may be possible to allow usage of type/param fields from the enclosing type from within a nested type.

This PR affects existing tests in the following ways:
- some futures are retired
- some tests are rewritten to not use 'outer' in any way
- some tests have new 'compopts' with --no-force-initializers, to allow them to be tested with constructors only

Resolves #9353 

Testing:
- [x] local + futures
- [x] gasnet
- [x] --force-initializers (down to 67 errors)


test/users/ferguson/histogram/reductionex_class.bad
```diff
@@ -1,7 +1,2 @@
-reductionex_class.chpl:44: internal error: CAL0124 chpl version 1.18.0 pre-release (02d6b31f68)
-
-Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
-and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
-please see https://chapel-lang.org/bugs.html for instructions.  In the meantime,
-the filename + line number above may be useful in working around the issue.
-
+reductionex_class.chpl:26: error: Illegal use of identifier 'NBUCKETS' from enclosing type
```
mppf: I do hope that this general pattern (nested type referring to outer type param/type) is allowed in the future. Do we have an issue / story about it? Also I have some concern that fields with the same name as an outer class field will change in behavior once this is implemented, so some of the above patterns would conceivably change. But hopefully they can continue to work.

test/classes/initializers/nested/scoping-mixed.chpl
```diff
@@ -0,0 +1,28 @@
+
+// This test exists to ensure that we do not encounter weird scoping issues
+// when a top-level type has a constructor and a nested type has an
+// initializer. This case did not 'just work' when initializers were enabled
+// for nested types and required compiler changes.
+
+class Node {
+  var x : int;
+}
```
mppf: Does this test need a --no-force-initializers (or maybe always run both with --force-initializers and --no-force-initializers) ?

compiler/resolution/functionResolution.cpp
```diff
@@ -8361,7 +8364,7 @@ static void insertReturnTemps() {
       if (call->list == NULL && isForallRecIterHelper(call))
         continue;
       if (FnSymbol* fn = call->resolvedOrVirtualFunction()) {
-        if (fn->retType != dtVoid) {
+        if (fn->retType != dtVoid && fn->retTag != RET_TYPE) {
```
mppf: Should this code run for fns returning a runtime type?

compiler/passes/normalize.cpp
```diff
@@ -2605,7 +2632,23 @@ static void hack_resolve_types(ArgSymbol* arg) {
...
       // hilde sez: don't we lose information here?
       if (arg->typeExpr->body.length == 1) {
-        Type* type = arg->typeExpr->body.only()->typeInfo();
+        Expr* only = arg->typeExpr->body.only();
+        Type* type = only->typeInfo();
+
+        // The type constructor might be known, but we don't want to remove
+        // the call to it just yet. See 'normalizeCallToTypeConstructor' for
+        // more information on how a SymExpr could be the baseExpr at this
+        // stage.
```
mppf: This feels work-around-y but it's not obvious to me how else it could be written. Probably worth thinking about if there's a "cleaner" way to do it for 5 minutes (or something). If we keep this strategy, it might be worth more clearly labeling it as a workaround. Or, why should this apply only to functions with FLAG_TYPE_CONSTRUCTOR?
Why does the call to the type constructor need to remain? Does this have to do with generic types? Could the comment or the code make the root issue clearer?

test/classes/initializers/nested/scoping-mixed.chpl
benharsh: This test passes successfully with and without --force-initializers.

compiler/resolution/functionResolution.cpp
benharsh: Yes, but in such cases ``fn->retTag`` is ``RET_VALUE``.

compiler/passes/normalize.cpp
benharsh: Minor update: I tried reworking this section to check whether 'type' was generic, but ran into some cryptic issue with tuples. I'm just going to update the comment.

PR #9893, opened 2018-06-15
Fix the error Sam ran into with documenting lifetime checking

### LINKED ISSUE #9887: Support unmanaged keyword in chpldoc
**Problem Summary**
Today the chpldoc tool does not support the unmanaged keyword as a return type and produces an unattractive string in its place. In working on the TOML documentation(issue #9841) I realized this might not be the only return type not supported. 

***Steps to Reproduce***
filename = sample.chpl
``` chapel
 proc parseToml(input: channel) : unmanaged Toml {
    var tomlStr: string;
    input.readstring(tomlStr);
    return parseToml(tomlStr);
}
```
***command***
`chpldoc sample.chpl`

***produces***
proc parseToml(input: channel): AppendExpr.Call09

lydia-duncan:
Resolves #9887 

Since unmanaged and borrowed utilize a primitive each that AstToText was
not expecting, we were getting AppendExpr.Call09 in the output when an
unmanaged or borrowed type was returned, specified as the type of an
argument, or specified as the type of a field.  Additionally, we were outputting
the under-the-covers representation of owned and shared, both when the type
was specified and when it was not.  This change fixes all those issues

Add a test that demonstrated the failure mode for all four lifetime styles, one
that is basically the same but involves generic types, and one for fields and
arguments declared using lifetime checking but no types.  All used to fail, but
now work with this change.


lydia-duncan:
TODOs:
- [ ] handle borrows
- [ ] handle when no explicit type is provided but lifetime checking is specified

lydia-duncan:
@mppf - is there anything I'm missing?

test/chpldoc/globals/lifetimeChecking.doc.chpl
```diff
@@ -0,0 +1,14 @@
...
+var c = new borrowed Foo();
+var d = new shared Foo();
+/* Double check documentation */
+var w = new owned Foo();
+/* Double check documentation */
+var x = new unmanaged Foo();
+/* Double check documentation */
+var y = new borrowed Foo();
+/* Double check documentation */
+var z = new shared Foo();
```
mppf: pesky missing newline

Issue #8242, opened 2018-01-18
Add support for user-defined assignment-based initialization


bradcray:
While converting code to use initializers, I've been musing over the relationship between copy initializers, coercions, and assignments.  For example, start with a record declared like this:

```chapel
record R {
  var x: int;

  // a 0-arg initializer
  proc init() { 
    this.x = 42;
    super.init();
  }

  // a copy initializer
  proc init(r: R) {
    this.x = r.x;
    super.init();
  }

  // an int initializer
  proc init(x: int) {
    this.x = x;
    super.init();
  }
}
```

The presence of the copy initializer permits us to invoke it by writing:

```chapel
var myR2 [: R] = myR;
```

where `myR` and `myR2` are both instances of `R`.  That is, because we're in a variable initialization context, we invoke the copy initializer rather than assignment.  And it also permits us to say the following in phase 1 of an initializer:

```chapel
this.r = r;
```

with a field and a value `r` of type `R`.  That is, because we're in phase 1, we're going to use a copy initializer to implement this assignment rather than assignment.

This has led me to wonder whether this suggests (a) we should also permit automatic invocation of other initializers in single-value cases of other types, such as:

```chapel
var myR3: R = 10;   // should this be equivalent to `= new R(10)`?
```

or:

```chapel
this.r = 11;    // should this be equivalent to `= new R(11)`?
```

or (b) we should use some other way of declaring copy initializers to make it clearer why a procedure named `init` is invoked in one case but not the other, or (c) whether user-defined coercions are really the way to support such cases where integers seem to become `R`s.

Put another way, if copy initializers are special initializers that just happen to be invoked when both (a) the RHS and LHS are of the same type and (b) there's an initializer that accepts a single argument of that type, are we setting up confusion for users who might think initializers would also get invoked when initializing with single-value expressions of other types?

bradcray:
As a more realistic motivation for this issue, I was wondering whether we could add an "`int` initializer" like the one on `R` above to our `atomic int` record in order to support initialization of `atomic int`s as follows:

```chapel
var myAtomicInt = 42;
```

or

```chapel
this.myAtomicInt = 365;
```

even if we were to never add support for directly assigning to atomics.  I.e., could we support initializing them without assigning them.  Was discussing this with @ronawho last night.

bradcray:
Noting belatedly that the previous comment relates to issue #5037.

bradcray:
[...]

PR #9894, opened 2018-06-15
Track alternate variable initialization error under initializers

benharsh:
Previously we would emit an error mentioning 'assignment', but with initializers we will mention an unresolved call to an initializer.

### DIFF
--- a/test/users/ferguson/recordassignstring.compopts
+++ b/test/users/ferguson/recordassignstring.compopts
@@ -1,0 +1,2 @@
+--no-force-initializers # recordassignstring.good
+--force-initializers # recordassignstring.init.good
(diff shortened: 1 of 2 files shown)


benharsh:
@mppf : Do you think this is a reasonable error message?

mppf:
@benharsh I'm fine with this change but the "unresolved init" call seems unnecessarily confusing vs something that mentioned copy-initializing. The compiler knows it's trying to copy-init and this is a concept that Chapel programmers should be familiar with. But, if the proposal in #8242 is adopted, I think this would be the right error message.

benharsh:
Would you prefer this change wait until #8242 is decided upon?

mppf:
It's fine with me either way. If simply updating the .good (as this PR does) helps you, great. If it gives us an excuse to finish deciding #8242, that's also great.

That said, I'd expect us to merge this PR and put completing #8242 on the backlog somewhere.

benharsh:
Previously we would emit an error mentioning 'assignment', but with initializers we will mention an unresolved call to an initializer.

[ r @mppf ]

PR #9903, opened 2018-06-18
Enable chpldoc to print out override proc

mppf:
chpldoc should show `override` when appropriate with method signatures.

- [x] full local testing


mppf:
Wow I don't think I've ever gotten stars before...

lydia-duncan:
It's something Thomas used to do :)

PR #9905, opened 2018-06-18
Fix issue with overloaded private function

### LINKED ISSUE #9793: Module visibility compilation error with overloaded private function
### Summary of Problem
The module `A` has a public method that calls into an overloaded private method. Using an explicit module prefix to call the public method causes an unexpected compilation error because it is trying to incorrectly access the private method instead of the public one. Workarounds: Omit the module prefix; Rename the private function.
```
error: Cannot access 'aproc', 'aproc' is private to 'A'
```

### Steps to Reproduce

**Source Code:**

```chapel
module A {
  proc aproc() {
    writeln("public aproc");
    aproc(1);
  }

  private proc aproc(x: int) {
    writeln("private aproc ", x);
  }
}

module B {
  use A;
  proc main() {
    A.aproc(); // error
    //aproc(); // This works.
  }
}
```

### Configuration Information
`chpl version 1.18.0 pre-release (e770634374)`

lydia-duncan:
Resolves #9793 

Depending on its placement in the module and whether it was accessed using a
module prefix or not, we would complain about calling a private function even
if a public function of the same name were available.  This resolves that issue.

It also adds five tests to lock in the new behavior.
Checks:
- module prefix access
  - private function defined first
  - private function defined last
  - all functions are private
- no module prefix access
  - private function defined last
- ensures that the change does not prevent us from finding symbol resolution
  errors.

Passed a full paratest

### DIFF
--- a/test/visibility/private/functions/symbolResError.chpl
+++ b/test/visibility/private/functions/symbolResError.chpl
@@ -1,0 +1,18 @@
+module A {
+  proc aproc() {
+    writeln("public aproc");
+    aproc(1);
+  }
+
+  private proc aproc(x: int) {
+    writeln(y);
+    writeln("private aproc ", x);
+  }
+}
+
+module B {
+  use A;
+  proc main() {
+    aproc();
+  }
+}
--- a/compiler/passes/ResolveScope.cpp
+++ b/compiler/passes/ResolveScope.cpp
@@ -403,8 +403,11 @@ bool ResolveScope::extend(Symbol* newSym) {
                 newSym->stringLoc());
     }
 
-    // If oldSym is a constructor and newSym is a type, update with the type
-    if (newFn == NULL || newFn->_this == NULL) {
+    // Prefer storage of non-functions over functions,
+    //                   functions over methods,
+    //                   public functions over private functions
+    if (newFn == NULL || (newFn->_this == NULL &&
+                          newFn->hasFlag(FLAG_PRIVATE) == false)) {
       mBindings[name] = newSym;
     }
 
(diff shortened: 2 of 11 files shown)


test/visibility/private/functions/symbolResError.chpl
```diff
@@ -0,0 +1,18 @@
+module A {
+  proc aproc() {
+    writeln("public aproc");
+    aproc(1);
+  }
+
+  private proc aproc(x: int) {
+    writeln(y);
```
bradcray: I'm not understanding why this change would conceivably have changed this behavior?

bradcray:
This looks reasonable to me.  I assume that if all functions are private, we still get the old error message (and that existing tests cover this)?

test/visibility/private/functions/symbolResError.chpl
lydia-duncan: It's mostly paranoia.  I'm fairly convinced that the machinery behind the ResolveScope type goes through all symbols, even if they aren't stored in the map I'm updating, but the control flow was odd enough that I wasn't completely convinced without a test to back it up

lydia-duncan:
> I assume that if all functions are private, we still get the old error message (and that existing tests cover this)?

We do still get the old error message when all overloads are private.  The pre-existing test covers when there isn't overloading at all, I'll add one for overloads when all are private

mppf:
At a quick look this looks reasonable to me but I have deeper questions about why scope resolve thinks it can do anything at all with function calls... or, put another way, how we can arrange for the compiler to do "early resolution". I don't think we should try to answer those questions for this PR though.

PR #9906, opened 2018-06-18
Optimize dsiAccess for new array type

lydia-duncan:
Save a version of the stored pointer within the chpl_external_array type into
a cast version of itself within the ExternArr class so that we don't have to
cast on every dsiAccess call (which will get very expensive over time).
Suggested by Michael.

This meant that I needed to add the ability to cast between c_void_ptrs and
_ddata.

Testing:
- [x] fresh checkout on export tests
- [x] fifo on export tests
- [x] valgrindexe on export tests
- [x] memleaks on export tests
- [ ] standard paratest (in progress after rebase)

### DIFF
--- a/modules/internal/ExternalArray.chpl
+++ b/modules/internal/ExternalArray.chpl
@@ -180,6 +180,7 @@ module ExternalArray {
     const dom;
     
     const _ArrInstance: chpl_external_array;
+    const elts = _ArrInstance.elts: _ddata(eltType);
 
     const _owned: bool;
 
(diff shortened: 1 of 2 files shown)


lydia-duncan:
@mppf - would you review?

modules/internal/ExternalArray.chpl
```diff
@@ -180,6 +180,7 @@ module ExternalArray {
     const dom;
     
     const _ArrInstance: chpl_external_array;
+    const elts = _ArrInstance.elts: _ddata(eltType);
```
mppf: For some reason I expected this to be set in the initializer rather than here. Should work either way, I'm fine with this.

lydia-duncan: elts gets handled via omitted initialization in the initializer :)

PR #9910, opened 2018-06-19
Read HDF5 data into DataFrames

daviditen:
Add a test that reads HDF5 files into columns of a dataframe

The test reads HDF5 files containing data of different types into arrays, then
inserts those arrays into columns in a dataframe.

A possible future direction is to allow reading directly into a dataframe instead
of needing to go through an intermediate array.


test/library/standard/DataFrames/diten/readHDF5Df.good
```diff
@@ -0,0 +1,16 @@
...
+6  0.833333      6   0.833333   
+7  0.857143      7   0.857143   
+8  0.875      8   0.875   
+9  0.888889      9   0.888889   
+10  0.9     10    0.9   
+11 0.909091     11   0.909091   
+12 0.916667     12   0.916667   
+13 0.923077     13   0.923077   
+14 0.928571     14   0.928571   
+15 0.933333     15   0.933333
```
benharsh: Was it intentional that the "real" files have identical content?

daviditen: Yes, it was intentional.  Since they're just examples with no real meaning it's fine for them to have the same data.

daviditen:
Thanks!

PR #9911, opened 2018-06-19
Add deprecation warning/error for extern classes

benharsh:
This PR introduces a compile-time warning if any extern classes are used within the program. Extern classes with constructors will still be allowed.

Also introduced is a compile-time error for any extern classes using initializers. Our initializers implementation has never supported extern classes, so emitting an error seems reasonable given that extern classes are going away entirely.

Various tests have been updated with the appropriate compile-time warning, and a deprecation test has been added to track this removal.

Testing:
- [x] local + futures
- [x] gasnet


bradcray:
This looks good to me.

Issue #9860, opened 2018-06-13
param loop over bool range implicitly casts iterator variable to int unexpectedly


ian-bertolacci:
A user (me) might expect that the following code had no compilation errors.

```chpl
proc foo( param i : bool ) {
  writeln( i );
}

for param i in false..true {
  foo( i );
}
```

However the following error is generated:
```
test.chpl:6: error: unresolved call 'foo(0)'
test.chpl:1: note: candidates are: foo(param i: bool)
```

It appears that the loop iterator `i ` is being implicitly converted to an int, as illustrated by the below code:
```chpl
for param i in false..true {
  writeln( (i.type : string, i) );
}
```

which produces the output:
```
(int(64), 0)
(int(64), 1)
```

```
> chpl --version
chpl version 1.17.1
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)
```

ian-bertolacci:
> "but would anyone ever use them?"
The only reason I am using them is because I cant do `for param i in [true, false] do ....` (unless there is a way to do this and I just don't know about it.

Your coercion to int explanation makes the looping behavior make sense, especially for when I tried `for i in false..true do ...` and got an empty loop.

bradcray:
Oh, I overlooked the `param` part...  it will take additional work to support param loops over ranges of bools or enums (but I can see why that would be a worthy feature...).

bradcray:
Probably obvious, but just in case: a workaround for this today would be to put the body of the loop into a function taking a param bool argument and call it with both true and false.  I..e, rather than:

```chapel
for param b in false..true do
    ...body using b...
```

use:

```
body(false);
body(true);
proc body(param b: bool) {
  ... body using b ...
}
```

This is not to say the request isn't a very reasonable one... just trying to keep you unblocked.

ian-bertolacci:
I found that you can just cast:
```chpl
proc foo( param i : bool ) {
  writeln( i );
}

for param i in false..true {
  foo( i : bool );
}
```

The use case for this was testing LayoutCS across all combinations of sorted and whatever else I was throwing at it.

bradcray:
Oh yeah, that's an even better workaround...

bradcray:
Hi Ian -- I'm going to close this in favor of #9866.

bradcray:
Hi Ian -- I've now implemented this (including the param case) in PR #9916.

PR #9916, opened 2018-06-19
Add support for bool ranges

### LINKED ISSUE #9866: Bool ranges
Now that ranges are starting to support `idxType`s other than primitive `int`s, should we support ranges of `bool`?  And `param` ranges of bools?

Pros
----
* well-defined, rounds out the set of supported types

Cons
-----
* limited utility (although... see issue #9860).

bradcray:
In issue #9860, @ian-bertolacci indicated a desire for bool ranges (where today, a bool range coerces into an int range, much as enum ranges did before).  In working on enum ranges in PR #9781, I wondered whether these would be useful and expected they'd be easy to add given my experience with enum ranges.  This adds support for them and a simple test showing that simple things work.

The compiler change is to support `param` bool ranges that yield booleans rather than uints.

The module changes add overloads for building new bool ranges and for the optimized direct range iterators.

Resolves #9866.


bradcray:
Note that this doesn't add support for param bool ranges which is what Ian really wanted... that'd be a follow-on PR.

bradcray:
OK, now it does support param bool ranges.

Issue #9912, opened 2018-06-19
binarySearch infinite loop if array strided array contains duplicates


ian-bertolacci:
binarySearch enters an infinite loop in the following code:
```chapel
use Search;
var A : [0..by 10 #5 ] int;
A[A.domain.low+0*A.domain.stride] = 1;
A[A.domain.low+1*A.domain.stride] = 2;
A[A.domain.low+2*A.domain.stride] = 2;
A[A.domain.low+3*A.domain.stride] = 3;
A[A.domain.low+4*A.domain.stride] = 4;

writeln( (A, A.domain) );

writeln( "linearSearch" );
writeln( linearSearch( A, 4) );

writeln( "binarySearch" );
writeln( binarySearch( A, 4 ) );
```

However binarySearch terminates when the array is one of the following:
+ `var A : [0..#5] int;`
+ `var A : [0..40 by 10] int;`

Cause is use of `hi=Domain.high` instead off `hi=Domain.alignedHigh`. Fix inbound.

- Output of `chpl --version`: 
```
chpl version 1.17.1.bae434820c
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)
```

ben-albrecht:
Should this be closed now?

Issue #9908, opened 2018-06-19
'Did not find' behavior of linearSearch.


ian-bertolacci:
In changes LayoutCS find behavior (#9877) we (@bradcray, @ben-albrecht ) discussed how things can be slightly simplified by having linearSearch return hi+stride instead of Dom.high+stride. This also makes it more symmetric with binarySearch which returns the location where it should have found the value (which in an unsorted linear search, could logically be the next element after the search range).

PR #9918, opened 2018-06-19
linearSearch and binarySearch updates

ian-bertolacci:
Addresses the issues #9908 and #9912, and changes in the pull request #9877.

In #9877 we discuss removing the domain slicing from linearSearch to increase performance, and in #9908 (and #9877) we discuss linearSearch returning `hi+abs(Dom.stride)` instead of `Dom.high+abs(stride)` to be symmetrical to binarySearch.

In #9912 I noted a bug in binarySearch that can cause an infinite loop with duplicates elements on some strided domains. This was fixed by changing the default input parameter `hi` to be `Dom.alignedHigh` instead of `Dom.high`.

Changes in linearSearch:
+ return hi+abs(stride) instead of Dom.high+abs(stride) (Dom.high is still the default value for hi)
+ removed domain slicing

Changes in binarySearch:
+ changes hi default from Doma.high to Dom.alignedHigh

### DIFF
--- a/modules/packages/Search.chpl
+++ b/modules/packages/Search.chpl
@@ -84,7 +84,7 @@ proc search(Data:[?Dom], val, comparator:?rec=defaultComparator, lo=Dom.low, hi=
    Searches through the array `Data` looking for the value `val` using
    a sequential linear search.  Returns a tuple indicating (1) whether or not
    the value was found and (2) the location of the first occurrence of the
-   value if it was found, or ``Data.domain.high+1`` if it was not found.
+   value if it was found, or ``hi+abs(Dom.stride)`` if it was not found.
 
    :arg Data: The array to search
    :type Data: [] `eltType`
--- a/modules/packages/Search.chpl
+++ b/modules/packages/Search.chpl
@@ -98,31 +98,29 @@ proc search(Data:[?Dom], val, comparator:?rec=defaultComparator, lo=Dom.low, hi=
    :type hi: `Dom.idxType`
 
    :returns: A tuple indicating (1) if the value was found and (2) the location
-      of the value if it was found or ``Data.domain.high+1`` if it was not
+      of the value if it was found or ``hi+abs(Dom.stride)`` if it was not
       found.
    :rtype: (`bool`, `Dom.idxType`)
 
  */
 proc linearSearch(Data:[?Dom], val, comparator:?rec=defaultComparator, lo=Dom.low, hi=Dom.high) {
 
-  chpl_check_comparator(comparator, Data.eltType);
+ chpl_check_comparator(comparator, Data.eltType);
 
-  const stride = if Dom.stridable then abs(Dom.stride) else 1;
-  // Domain slicing is cheap, but avoiding it when possible helps performance
-  if lo == Dom.low && hi == Dom.high {
-    for i in Dom {
-      if chpl_compare(Data[i], val, comparator=comparator) == 0 then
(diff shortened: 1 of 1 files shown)


modules/packages/Search.chpl
```diff
@@ -98,31 +98,29 @@ proc search(Data:[?Dom], val, comparator:?rec=defaultComparator, lo=Dom.low, hi=
...
-    }
-  }
-
-  return (false, Dom.high+stride);
+ chpl_check_comparator(comparator, Data.eltType);
+
+ const stride = if Dom.stridable then abs(Dom.stride) else 1;
+ if Dom.stridable {
+   for i in lo..hi by stride {
+     if chpl_compare( Data[i], val, comparator=comparator ) == 0 then
```
ben-albrecht: *nitpick:* was the stylistic change (extra spaces in `chpl_compare` calls) intentional here?

ian-bertolacci:
Hm. I'm still having an issue not using `_value` if I change `sorted` (newly`sortedIndices`) to the completely unused symbol `sortedIndices_bertolacci_2018` I still get an error

```
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1411: In function '_csrmatmatMult':
$CHPL_HOME/modules/packages/LinearAlgebra.chpl:1430: error: ambiguous call 'CSDom(2,int(64),domain(2,int(64),false),true,true,false).sortedIndices_bertolacci_2018'
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note: candidates are: _domain.sortedIndices_bertolacci_2018
$CHPL_HOME/modules/internal/ChapelArray.chpl:1074: note:                 _domain.sortedIndices_bertolacci_2018
```

modules/packages/Search.chpl
ian-bertolacci: Oh thats one of my styles, I'll change it.

ian-bertolacci:
Tests pass. Should I merge?

modules/packages/Search.chpl
```diff
@@ -159,7 +157,7 @@ proc linearSearch(Data:[?Dom], val, comparator:?rec=defaultComparator, lo=Dom.lo
    :rtype: (`bool`, `Dom.idxType`)
 
  */
-proc binarySearch(Data:[?Dom], val, comparator:?rec=defaultComparator, in lo=Dom.low, in hi=Dom.high) {
+proc binarySearch(Data:[?Dom], val, comparator:?rec=defaultComparator, in lo=Dom.low, in hi=Dom.alignedHigh) {
```
bradcray: I'm guessing that `lo` should be `Dom.alignedLow` as well for symmetry arguments (e.g., if we had 1..100 by 10 align 9, I could imagine that there could be a similar infinite loop searching at the bottom of the range).

And probably ditto for both args of linear search as well (e.g., if someone passed in a domain that was `[1..10 by 2 aligned 2]` would we search 1, 3, 5, 7, 9 rather than 2, 4, 6, 8, 10?

ian-bertolacci:
Ya I think your right.

ian-bertolacci:
I also discovered that it can just flat-out break linear search.
So I've modified *all* lo and hi parameter default values to be alignedLow and alignedHigh.
I'm waiting on more tests, but I expect to have the updated pull request today.

modules/packages/Search.chpl
ben-albrecht: Good catch!

PR #9921, opened 2018-06-19
Fix bug with default initializers and type/param fields with default values

benharsh:
This PR resolves a bug where the compiler could not resolve calls to a default initializer if the type has param/type fields with default values. The problem was that the compiler was generating the 'init' call without using NamedExprs to match the appropriate type/param fields. This encountered errors when the type had non param/type fields between declarations of param/type fields:

```chpl
record R {
  type T = int;
  var x : string;
  param p : real;
}

// Compiler may generate as R.init(int, real) within _defaultOf
var r : R;
```

The solution is to update creation of _defaultOf to use NamedExprs when invoking a default initializer.

Testing:
- [x] local + futures
- [x] gasnet


benharsh:

test/classes/initializers/compilerGenerated/default-param-type-fields.chpl
```diff
@@ -0,0 +1,15 @@
...
+  var foo : string;
+
+  param C = 3;
+
+  var x : C*int;
+}
+
+var r : R(int, false);
+
+writeln(r);
```
lydia-duncan: It would be good to also have a `writeln(r.type: string);` line in this test

PR #9924, opened 2018-06-19
Improvements to lifetime checker

### LINKED ISSUE #9714: Lifetime checking error not caught
### Summary of Problem

The program below should generate an error with `--lifetime-checking`, since the user probably wants `c` to be `owned`, but it is actually `borrowed`, because `c` was declared as `C` instead of `Owned(C)`.

### Steps to Reproduce

**Source Code:**

```chapel
class C { }
var c: C;
c = new owned C();
```

- Output of `chpl --version`: `chpl version 1.18.0 pre-release (51f70f0)`

### LINKED ISSUE #9857: lifetime checking inference overachieves
### Lifetime inference overachieves

The lifetime checking pass infers the lifetime of a symbol to be the minimum of lifetimes assigned to it. But this is running into problems with references, as in:

``` chapel
pragma "safe"
module X {

class MyClass {
  var x:int;
}

var global: borrowed MyClass;

proc setG(arg: borrowed MyClass) {
  ref x = global;
  x = arg; // should be an error
}

proc test() {
  var x = new owned MyClass(1);
  setG(x);
}
test();

}
```

This was added to handle a case like this:
```
var first:MyClass = nil;
while ... {
  if first:object == nil then
     first = e;
}
```

Here it is desirable to infer `first` as having lifetime == `e` rather than infinite lifetime.

It might be possible to resolve this issue by treating avoiding this inference for borrows when the LHS is a reference to something.

mppf:
* Does a better job of removing reference and other compiler-added temporaries 
  for the purpose of the analysis
* Removes several workarounds
* Distinguishes between unknown and infinite lifetimes
* Separate traversal now computes the intrinsic lifetime for each variable. 
  This is the "natural" lifetime of the variable (e.g. the reachability / 
  scope for a record). Doing this separately clarifies the follow on code 
  that uses this concept.
* Several erroneous cases that were previously undetected are now detected.

Resolves #9857
Resolves #9714

Passed full local testing.

### DIFF
--- a/compiler/resolution/lifetime.cpp
+++ b/compiler/resolution/lifetime.cpp
@@ -1484,10 +1484,6 @@ bool EmitLifetimeErrorsVisitor::enterCallExpr(CallExpr* call) {
       bool usedAsRef = lhs->isRef() && rhsExpr->isRef();
       bool usedAsBorrow = isOrContainsBorrowedClass(lhs->type);
 
-      LifetimePair rhsLt = lifetimes->inferredLifetimeForExpr(rhsExpr,
-                                                              usedAsRef,
-                                                              usedAsBorrow);
-
       // track rhs symbol for error messages
       Symbol* rhsSym = NULL;
       if (SymExpr* rhsSe = toSymExpr(rhsExpr)) {
@@ -1495,6 +1491,10 @@ bool EmitLifetimeErrorsVisitor::enterCallExpr(CallExpr* call) {
       }
 
       if (lhs->hasEitherFlag(FLAG_RVV,FLAG_RETARG)) {
+        LifetimePair rhsLt = lifetimes->inferredLifetimeForExpr(rhsExpr,
+                                                                usedAsRef,
+                                                                false);
+
         // Check returns
         if (lhs->isRef() && call->isPrimitive(PRIM_MOVE)) {
           // check returning a reference
@@ -1525,7 +1525,8 @@ bool EmitLifetimeErrorsVisitor::enterCallExpr(CallExpr* call) {
           // (vs own).
 
           // check returning a borrow
-          if (rhsLt.borrowed.infinite || rhsLt.borrowed.unknown) {
+          if (rhsLt.borrowed.infinite || rhsLt.borrowed.unknown /*||
+              rhsLt.borrowed.fromSymbolScope == rhsSym*/) {
             // OK, not an error
           } else {
             if (calledFn &&
@@ -1545,6 +1546,10 @@ bool EmitLifetimeErrorsVisitor::enterCallExpr(CallExpr* call) {
           }
         }
       } else {
+        LifetimePair rhsLt = lifetimes->inferredLifetimeForExpr(rhsExpr,
+                                                                usedAsRef,
+                                                                usedAsBorrow);
+
         // Raise errors for init/assigning from a value with shorter lifetime
         // I.e. insist RHS lifetime is longer than LHS lifetime.
         // I.e. error if RHS lifetime is shorter than LHS lifetime.
--- a/compiler/resolution/lifetime.cpp
+++ b/compiler/resolution/lifetime.cpp
@@ -70,6 +70,8 @@
 
 
    TODO:
+     - **** Separate infinite lifetime from unknown lifetime
+
      - investigate if this can do escape checking for begins
      - decide on how to declare an owning vs borrowed class instance ptr
        - vs a raw one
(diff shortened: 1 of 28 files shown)


mppf:
@vasslitvinov - are you up for having a look at this PR improving the lifetime checker?

compiler/resolution/lifetime.cpp
```diff
@@ -953,109 +1312,136 @@ bool EmitLifetimeErrorsVisitor::enterCallExpr(CallExpr* call) {
   if (call->id == debugLifetimesForId)
     gdbShouldBreakHere();
 
+  // Ignore calls that de-temping allows us to ignore
+  if (lifetimes->callsToIgnore.count(call))
+    return false;
+
   if (isAnalyzedMoveOrAssignment(call)) {
```
vasslitvinov: This function is pretty long. How can we organize the code? Ex.

    checkXXX(call);
    checkYYY(call);
    checkZZZ(call);

?

compiler/resolution/lifetime.cpp
```diff
@@ -1112,7 +1523,7 @@ static bool isLocalVariable(FnSymbol* fn, Symbol* sym) {
   return sym->defPoint->getFunction() == fn && isLocal;
 }
 
-static bool typeHasInfiniteLifetime(Type* type) {
+static bool typeHasInfiniteBorrowLifetime(Type* type) {
```
vasslitvinov: Would be nice to know how "borrow" lifetime differs from "intrinsic" and "inferred" lifetimes.

compiler/resolution/lifetime.cpp
```diff
@@ -1220,6 +1631,19 @@ static bool containsOwnedClass(Type* type) {
   return false;
 }
 
+static bool isOrContainsBorrowedClass(Type* type) {
+  type = type->getValType();
+
+  if (AggregateType* at = toAggregateType(type)) {
+    if (at->isClass())
```
vasslitvinov: If there is a function that tests specifically for a "borrow" type, it should probably be used here instead.

compiler/resolution/lifetime.cpp
```diff
@@ -1220,6 +1631,19 @@ static bool containsOwnedClass(Type* type) {
...
+      return true; // borrow
+    if (isRecord(type))
+      return recordContainsBorrowedClassFields(toAggregateType(type));
```
vasslitvinov: Let us drop "toAggregateType" above.

compiler/resolution/lifetime.cpp
```diff
@@ -37,12 +37,18 @@
...
+       during the one moment it is initialized.
+     * inferred lifetime, based on what that symbol is set to
+       (e.g. for a variable of borrow class type, this is the minimum
+        of the lifetimes stored in it)
 
    The checker emits errors for several cases:
 
-     * reachability of a symbol is greater than its lifetime
+     * intrinsic lifetime of a symbol is greater than its inferred lifetime
      * setting a symbol to another value with a shorter lifetime
```
vasslitvinov: The lifetime terminology is still not clicking for me.
I think you change is a step forward, anyway.

compiler/resolution/lifetime.cpp
```diff
@@ -777,90 +978,229 @@ static bool isRecordInitOrReturn(CallExpr* call, Symbol*& lhs, CallExpr*& initOr
...
+  // visit the same one more than once due to canonicalizing temps
+  LifetimePair lp = unknownLifetimePair();
+  if (lifetimes->intrinsicLifetime.count(sym))
+    lp = lifetimes->intrinsicLifetime[sym];
+
+  if (ArgSymbol* arg = toArgSymbol(sym)) {
+
+    // For arguments, set the lifetime == reachability
+    Lifetime lt = reachabilityLifetimeForSymbol(sym);
+
```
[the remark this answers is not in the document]
mppf: I changed it to `scopeLifetimeForSymbol`

compiler/resolution/lifetime.cpp
mppf: I'll add a comment but each "lifetime" (intrinsic or inferred) contains two components in the analysis - ref lifetime and borrow lifetime. One is for handling "borrow checking" the other one validates refs to locals do not escape.

compiler/resolution/lifetime.cpp
mppf: there isn't a global function for that right now, but I cleaned up this area of the checker implementation in other ways.

mppf:
@vasslitvinov - thanks for your review. I'm planning to merge this once a futures test run completes (and I'm not expecting to see any non-trivial errors there).

Issue #9654, opened 2018-05-29
Convert FileSystem module to use error handling


ronawho:
Convert halts in the FileSystem module to throw'ing functions (where appropriate)

ronawho:
After digging into this, it looks like there aren't any remaining cases in FileSystem where we want to use error handling. Converting the uses of halt/__primitive("chpl_error", ...) to halt wrappers in https://github.com/chapel-lang/chapel/pull/9930 (only remaining halts are for OOM and mismatched zippered iteration lengths)

PR #9930, opened 2018-06-19
Clean up halts in glob() iterators

ronawho:
Previously, the glob iterators used a `__primitive("chpl_error", ...)` if there
was an unhandled error from `chpl_clob()` (a thin wrapper over `glob()`). When
calling `glob()` with no flags like we do, the only possible error codes are
`GLOB_NOMATCH` and `GLOB_NOSPACE`. `GLOB_NOMATCH` is innocuous and will just
result in an empty iterator. `GLOB_NOSPACE` occurs if the `malloc()` inside of
glob fails, so just convert that into our usual "Out of memory allocating ..."
error

I didn't want to update all the callsites of chpl_glob to do this error
checking so I just made a glob_w wrapper that does the checking. That seemed
like a nice cleanup, so I made similar wrappers for the other glob routines to
cleanup a lot of safeCasts at callsites.

This also converts a halt() for bad zippered iteration to use the new
zipLengthHalt() wrapper.

Closes https://github.com/chapel-lang/chapel/issues/9654

### DIFF
--- a/modules/standard/ChapelHaltWrappers.chpl
+++ b/modules/standard/ChapelHaltWrappers.chpl
@@ -66,4 +66,25 @@ module ChapelHaltWrappers {
   proc pureVirtualMethodHalt() {
     halt("pure virtual method called");
   }
+
+
+  //
+  // Halt wrappers for misc runtime time checks (where we expect these checks
+  // to halt even in an error-handling world.)
+  //
+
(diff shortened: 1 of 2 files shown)


ronawho:
This is much easier to review as a split/side-by-side review: https://github.com/chapel-lang/chapel/pull/9930/files?diff=split

bradcray:
Looks good to me, thanks!

Issue #8468, opened 2018-02-13
Support assumed-domain array return types


bradcray:
### Summary of Problem

Today, if you want to return an array of completely unknown size, the main way you have
of doing so is to have the return type be inferred.  But for those who want to say more
about what their procedure is returning (for readability; for checking themselves), it'd
be nice to support an assumed-domain array return type as we do for formal arguments.
I.e., `proc foo(): [] real { ... }` would say "foo() returns an array of reals, but we aren't saying
(maybe can't say) how big / what type of array it is."

[edit: Just as formals support an element-type-less `proc foo(X: [])` format, I think we'd want
to do the same with this feature, though I didn't think of that fast enough to get it into my
future].

### Steps to Reproduce

**Associated Future Test(s):**
[`test/arrays/return/returnArbitraryArray.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/arrays/return/returnArbitraryArray.chpl
          ) #8466

### Configuration Information

- Output of `chpl --version`: `chpl version 1.17.0 pre-release (6b89687)`

bradcray:
Note that this issue is related to #8469.

lydia-duncan:
Does this also mean that recursive functions can't return arrays?

bradcray:
Today I believe it is accurate that recursive functions can't return arrays.  We've had a long-term academic user who's bristled over this for years.

lydia-duncan:
I would like to up the priority on this issue.  It forces me to support inferred returns from exported functions that return an array, especially in the case of an exported function like the following code.  Only supporting explicit return type declarations would have been a nice simplifying case while I got this feature off the ground.

```chapel
export proc foo(size: int) /*: [] int*/ {
  var x: [0..#size] int;
  return x;
}
```

Additionally, it would improve documentation for functions that otherwise have to use inferred returns.

PR #9931, opened 2018-06-19
Add support for returning arrays from exported functions (explicit return type only)

### LINKED ISSUE #9816: Implement support for exporting Chapel arrays part 2b - returning an array
This is similar to #9603 - but aimed at supporting exported functions that return arrays in the compiler.  Once the implementation for part 2a is complete, it should be easier to extend that implementation for functions that return arrays (while ensuring that it can be used with or without taking an array argument).

I will need to write some tests for this mode, too

lydia-duncan:
Resolves #9816 

This provides compiler support for exported functions that return arrays.
Prior to this change, it was difficult to know the corresponding .h file to
write, and difficult to write a .c file that could utilize the returned value.
This change makes a wrapper for exported functions that return arrays,
and in that wrapper takes the returned value and transforms it into
something that is more easily understood by the C code that could call
it (an instance of our runtime-defined array wrapper type).  It only does
this for types that explicitly declared an array return type - if the return
type is inferred, we will not make the transformation (and the caller will
get a segfault because the type passed doesn't match the type specified
in the library implementation).

Due to the new testing, a couple more methods were required on the
ExternDom type, so add them.  Hopefully there aren't too many more of
these required.

Adds a slew of new tests:
- calling an exported function that takes an int argument and returns an
array
- calling an exported function that returns an int array and takes an int
argument
- calling an exported function that returns an int array and takes no
arguments
- calling an exported function that returns an array of reals
- calling an exported function that takes an array argument and returns
an array
- calling an exported function that returns an array whose lower bounds
is not 0 (intended to fail, for now)
- compiling a library with an exported function that returns a
multi-dimensional array (intended to fail, for now)
- compiling a library with an exported function that returns a strided
array (intended to fail, for now)

And some futures that don't work because we don't support
assumed-domain array return types (mostly copies of tests that are
already listed, see #8468; plus a test that exercises compiling a library
with an exported function that returns an associative array, since we must
infer that return type; plus a test to track a valgrind error for creating an
array using the domain of an external array)

TODO:
- [x] ensure that my test files clean up the memory returned to them.
- [x] add .bad files for the .futures

Testing:
- [x] exported arrays fresh checkout (and futures)
- [x] exported arrays with fifo (and futures)
- [x] exported arrays with valgrind (and futures)
- [ ] exported arrays with memleaks?
- [x] full local paratest with futures (in progress)


lydia-duncan:
The valgrind failure is due to not giving an initial value to the array I create from the argument's domain.  This seems to imply that there is something I should do about array initialization that I am not doing.

lydia-duncan:
@mppf - would you review this?  And let me know if you think I should do a memleaks run (we don't run memleaks over .test.c files, so I don't think it will catch anything)

compiler/passes/normalize.cpp
```diff
@@ -2693,13 +2712,12 @@ static void makeExportWrapper(FnSymbol* fn) {
...
-  // pls.
-  if (argsToReplace) {
-    // We have at least one array argument.  Need to make a version of this
-    // function that can be exported
+  if (argsToReplace || returnsArray(fn)) {
+    // Either we have at least one array argument, or we return an array.
+    // Need to make a version of this function that can be exported
     FnSymbol* newFn = fn->copy();
     newFn->addFlag(FLAG_COMPILER_GENERATED);
+    newFn->addFlag(FLAG_LAST_RESORT);
```
mppf: Let's add a comment to say why FLAG_LAST_RESORT is appropriate -- I'm supposing you want Chapel code to call the original and not the export wrapper?

compiler/passes/normalize.cpp
```diff
@@ -1218,6 +1222,21 @@ static void normalizeReturns(FnSymbol* fn) {
   }
 }
 
+static void fixupExportedArrayReturns(FnSymbol* fn) {
+  if (fn->hasFlag(FLAG_EXPORT) && returnsArray(fn)) {
```
mppf: How do you know that this will run only after makeExportWrapper runs?
It'd be bad to do this to the Chapel version before making the export wrapper, right?
Since Chapel code wouldn't be able to use it anymore?

So I'd expect this to check for FLAG_COMPILER_GENERATED and FLAG_EXPORT both, and at least a comment indicating why you know it runs after makeExportWrapper (since at least this relationship is not obvious to me).

mppf:
Regarding memory leaks testing for exported procs I think it'd be reasonable to do but I don't view it as a requirement for this PR to merge. Remember you can always use valgrind to do it if the Chapel leak mechanism isn't available for one reason or another.

compiler/passes/normalize.cpp
lydia-duncan: In the case where the arguments don't differ, we end up with a resolution conflict without this.  We do want to prioritize on calling the original version when we resolve the wrapper's body, too

compiler/passes/normalize.cpp
lydia-duncan: Right, good point.  I'd forgotten all the checks I put in place, I'll add that

lydia-duncan:
> Regarding the tests, I think they would be simpler if you didn't use library mode. export is still useful because you can call Chapel -> C -> Chapel with it. However it's OK for them to use library mode; I just think the result is more complex & less focused on the behavior of exporting procs that return arrays.

I think that's fair.  I mostly use library mode because tagging functions as "exported" doesn't really make sense to me in a Chapel->Chapel world at the moment, since we have `use` statements available and don't allow separate compilation at the moment.  I'm not sure I really understand how to write what you're describing in a way that is less complicated than what I am doing, would you mind providing an example?

mppf:
Sketch of example:
 * chpl file with an export proc bar, and an extern proc foo; calls foo()
 * header file with bar and foo in it
 * c file with foo() defined to call bar()

You compile it all with chpl something.chpl something.h something.c and it should start in Chapel, go to C, and then go back to Chapel.

lydia-duncan:
Oh, I see.  So it's a back-and-forth program.  Interesting

PR #9938, opened 2018-06-20
dsiAlignment information in Cyclic Replicated distributions

ian-bertolacci:
adds to both CyclicDom and ReplicatedDom:
+ dsiAlignment
+ dsiAlignedHigh
+ dsiAlignedLow

Fixes broken test cause by new Search module lo and hi search function defaults.

### DIFF
--- a/modules/dists/CyclicDist.chpl
+++ b/modules/dists/CyclicDist.chpl
@@ -497,6 +497,12 @@ proc CyclicDom.dsiLow return whole.low;
 
 proc CyclicDom.dsiHigh return whole.high;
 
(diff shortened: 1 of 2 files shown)


ben-albrecht:
Do we have an open issue somewhere to document that these should eventually be added to base domains?

ian-bertolacci:
Not yet let me do that right quick.

ian-bertolacci:
Filed under #9942.

ian-bertolacci:
There might be, but there isn't a file that really stands out as having them. These tests should probably be added when #9942 is considered, since we'll (probably) want to cover all distributions to ensure they all have the same behavior.

Issue #9941, opened 2018-06-20
Should timezone awareness be a part of datetime's static type?


ronawho:
In the current DateTime implementation, timezone awareness is determined by whether datetime's tzinfo is nil or not. Operations between mixed datetimes are not allowed, which means that either both datetimes must have timezones or neither can.

In today's implementation this leads to lots of runtime checks for mixed timezone awareness, and mismatches currently lead to halts: https://github.com/chapel-lang/chapel/blob/805996b81270af5e2b3e96cc8d8dfaf8ab668df3/modules/standard/DateTime.chpl#L1301-L1319

In looking to convert DateTime over to error-handling, it seems like a shame to have to make these operators throwing, since that seems burdensome for users. e.g. for something like:

```chpl
use DateTime;

var t1 = datetime.now();
var t2 = datetime.now();
if t1 <= t2 then writeln(t1);
```

Ignoring errors would look something like: 

```chpl
use DateTime;

var t1 = datetime.now();
var t2 = datetime.now();
if (try! t1 <= t2) then writeln(t1); // try! as part of the condition looks weird to me
```

And if you did want to handle errors would look something like:

```chpl
use DateTime;

var t1 = datetime.now();
var t2 = datetime.now();
var cmp: bool;
try {
  cmp = t1 <= t2;
} catch {
  // handle error
}
if cmp then writeln(t1);
```

Mismatched timezones feels like a type error to me. This made me wonder if we should just make timezone awareness be part of the type, which would allow us to perform these checks at compile time.

riftEmber:
In https://github.com/Cray/chapel-private/issues/5170 we decided to do this for 2.0, and implementation is in progress at https://github.com/chapel-lang/chapel/pull/23010.

jeremiah-corrado:
In attempting to implement this change, I'm realizing there may need to be a few additional API changes to the `time` and `dateTime` types to smoothly transition from runtime timezone-awareness to compile time timezone-awareness.

Under the current design, some `time` and `dateTime` factory methods and initializers take nilable `Timezone` arguments, where a `nil` value results in a timezone-unaware (or naiive) value, and a non-nil argument results in a timezone-aware value. This does not mix well with the new design where a `param tzAware: bool` field would designate the timezone awareness of the value.

### Specifics of Problem:

Consider the `replace` interface, which can be used to create a new `time` or `dateTime` that inherits the fields from an existing value, except with some of its values replaced. Example:

```chapel
[...]

PR #9944, opened 2018-06-20
Partially convert DateTime to error-handling

ronawho:
This converts some halts to asserts for internal/private functions, converts
some halts in initializers to initHalts, and actually converts a few halts to
throwing functions. Individual commits have more details but at a high level
this:
 - Converts halts in daysBeforeMonth() and ymdToOrd() to asserts. These are
   private functions and all callsites within DateTime have valid arguments.
 - Converts daysInMonth() to throw an IllegalArgumentError() instead of halting
 - Converts halt() calls in time, date, and timedelta initializers to
   initHalt() calls. These are cases where we want error-handling, but it's
   currently not supported

The remaining uses of halt are for operations between datetimes with mixed
timezone awareness. It's not obvious to me if these should be throwing
routines, or if we should make timezone awareness part of the static type. I've
opened https://github.com/chapel-lang/chapel/issues/9941 to discuss this.

Related to https://github.com/chapel-lang/chapel/issues/9552

### DIFF
--- a/modules/standard/DateTime.chpl
+++ b/modules/standard/DateTime.chpl
@@ -113,6 +113,10 @@ module DateTime {
     var tm_zone:   tm_zoneType; // Timezone abbreviation
   }
 
+  private proc assertInRange(val, low, high) {
+    assert(val >= low && val <= high);
+  }
+
   private proc getLocalTime(t: 2*int) {
     extern type time_t;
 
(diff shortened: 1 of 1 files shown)


modules/standard/DateTime.chpl
```diff
@@ -186,10 +189,12 @@ module DateTime {
...
         year -= 1;
       }
-      preceding -= daysInMonth(year, month);
+      const dim = try! daysInMonth(year, month);
+      preceding -= dim;
     }
     n -= preceding;
-    assert(0 <= n && n < daysInMonth(year, month));
+    const dim = try! daysInMonth(year, month);
+    assertInRange(n, 0, dim);
```
daviditen: This is off by one from the original assert `n < dim` vs. `n <= dim`.

ronawho: Shoot, good catch. Just force pushed a change that makes it `assertInRange(n+1, 1, dim);`

PR #9947, opened 2018-06-21
Quiet testing noise from ill-formed sed command

Spartee:
This PR should make the recently merged sub-directory PR pass testing without any noise thanks to @ronawho

### DIFF
--- a/test/mason/subdir-commands/PREDIFF
+++ b/test/mason/subdir-commands/PREDIFF
@@ -3,6 +3,6 @@
 outfile=$2
 temp=$outfile.temp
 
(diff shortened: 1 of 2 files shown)


ronawho:
`rm -rf ~/.mason && start_test test/mason/subdir-commands/mason-run.chpl` still fails for me. I think you need to change  the `registry` in the first half of the sed command to `mason-registry`

e.g. 
```
sed 's/Initializing mason-registry for mason-registry/Updating mason-registry for mason-registry/'
```

Spartee:
Latest commit should fix it @ronawho

Issue #7249, opened 2017-09-08
except list hides some methods of included types


lydia-duncan:
### Summary of Problem
If two types have a method with the same name, and one of those types is excluded from a scope via the except list of a use statement, this accidentally prevents the included type from utilizing their method with the same name in the scope where the use is defined.  Ugh.

### Steps to Reproduce

**Source Code:**

```chapel
module M {
  class C1 {
    var x: int;

    proc foo() {
      writeln(x);
    }
  }

  class C2 {
    var y: bool;

    proc foo() {
      writeln(y);
    }
  }
}

module OtherMod {
  use M except C2;

  proc main() {
    var c1 = new C1(7);
    // The call below should work, but doesn't due to C2 having a method with the same name
    c1.foo();
    delete c1;
  }
}
```

**Compile command:**
`chpl foo.chpl`

**Associated Future Test(s):**
[`test/visibility/except/multipleTypesSameMethodName.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/visibility/except/multipleTypesSameMethodName.chpl) (#7250)

Affected tests under ``--force-initializers``:
- [ ] visibility/only/typeIncludedAccessField
- [ ] visibility/only/typeIncludedAccessMethod
- [ ] visibility/rename/renameClass
- [ ] visibility/rename/renameRecord

### Configuration Information

- Output of `chpl --version`: 1.16.0 (pre-release)

lydia-duncan:
I know exactly where the fix for this should go, I think.

lydia-duncan:
I was originally going to punt on this, but since initializers all share the same name, I'm realizing this bug means that init and _new will not be found for other types if any type that uses initializers is excluded, so this will need to get fixed sooner rather than later

benharsh:
I'll document a few of the related problems/concerns here:

1) We should have a way to ask a UseStmt if a method call is visible. This requires passing the type name and method name to a UseStmt method.

2) Putting all 'relatedNames' into a list loses information (where they came from). I think it would be better if there was a map from type-name --> set-of-member-names. I'm not sure if that's possible within the current UseStmt/scopeResolve architecture.

3) During type resolution (e.g. resolveNormalCall), we will want to ask whether a method call is visible. If I ask the UseStmt if "T.foo" is visible, but "T" is renamed, the UseStmt will say it is not visible. It appears as though we only store a map from new-name --> old-name, so it's not a simple lookup. Especially if the new names are still floating around.

[...]

PR #9957, opened 2018-06-21
Update except/only to handle 'init' and '_new' specially

benharsh:
This PR updates handling of except/only lists to handle 'init' and '_new' specially. When identifying symbols from a type, 'init' and '_new' will not be added for ``except`` lists and will be added for ``only`` lists. This is only a workaround, and does not address general issues with except/only lists and method names.

Testing:
- [x] local + futures


lydia-duncan:
I think we should open an issue for the general `except`/`only` issue, if there isn't one already (I didn't find one when I went looking earlier)

benharsh:
I believe #7249 is what you were looking for.

lydia-duncan:
I must be getting old, lol, should have remembered I opened one

PR #9962, opened 2018-06-21
Quiet paratesting with EXECENV and SKIPIF

Spartee:
This should fix the error for the sub-directory testing for #9875 that was merged yesterday. Directory needed an EXECENV.


ben-albrecht:
It appears this did not resolve the nightly failure

Issue #9853, opened 2018-06-12
lifetime checking design problems


mppf:
### Assumption that formal arguments have same lifetime

With array.push_back, an element with lifetime == `this` is set to another argument. But, the current analysis assumes that all formal arguments have the same lifetime. That prevents the lifetime checker from detecting errors involving array.push_back such as:

``` chapel
pragma "safe"
module X {

class MyClass {
  var x:int;
}

record BorrowedThing {
  var b: borrowed MyClass;
}

var globalB: BorrowedThing;

proc BorrowedThing.errorSetG(arg: borrowed MyClass) {
  this.b = arg; // this should be an error
}

var globalA: [1..1] borrowed MyClass;

proc errorPushG(arg: borrowed MyClass) {
  globalA.push_back(arg); // this should be an error
}

proc test() {
  var x = new owned MyClass(1);
  globalB.errorSetG(x.borrow());
  errorPushG(x.borrow());
}
test();

}
```

### Writing array.push_back

Were the previous issues resolved, `proc array.push_back(value)` would cause an error in compilation. How can a version of `proc array.push_back(value)` be written that does compile?

Brainstorming possible directions:
* `scope` variables can't be stored in arrays / collections
* analysis assumes each argument has independent lifetime
* lifetime annotation on the arguments indicates `this` should have lifetime <= `value`
* assumption about argument declaration order - lifetime of 1st arg (this) is <= lifetime of 2nd arg (value) .
* assumption that `this` argument has lifetime <= other arguments
* specially mark `borrowed` arguments that can be saved anywhere in the called function
* treat containers differently from other types

### Impact

The impact of adding checking for these cases is that where we previously were checking the way values are returned and assigned, now we will also check that values passed to a function have appropriate lifetime constraints. E.g. 5 arguments passed to a function might have some lifetime constraint among them.

bradcray:
Of the solutions proposed here, this one feels least scary to me:

> * lifetime annotation on the arguments indicates this should have lifetime >= value

Are there lessons to be learned from prior art (e.g., modern C++ or Rust) here?

mppf:
When this is resolved, update the following tests which currently use a workaround:
 - [ ] generic-collection-unspecified-with-type.chpl
 - [ ] generic-collection-unspecified.chpl

mppf:
Generally speaking, a function could take two arguments and it could save borrows from either argument in the other. Or both. E.g.
``` chapel
// supposed this is called with borrows or with records containing borrows
[...]

PR #9963, opened 2018-06-21
Update lifetime checker to get primers to pass

mppf:
This PR resolves problems with lifetime inference with for and forall loop 
iteration variables.

Additionally it marks the `_add` method in DefaultAssociative as unchecked (by 
the lifetime checker). This is to drive towards understanding the current
failures and getting clean testing with --lifetime-checking. A solution to
#9853 should be used here once it exists.

- [x] full local testing


vasslitvinov:
@mppf - how is it that this PR apparently disables checking on index variables, yet `iter-return-first-borrowed-array-element.chpl` behaves as desired? (Does it?)

mppf:
@vasslitvinov - it's a different check in a different section. The workaround only disables the error for setting a reference if that reference is an index variable. Errors for returning or setting a borrow in a bad way are elsewhere.

PR #9964, opened 2018-06-21
Help cray-module correctness tests adapt to new requirement (Cray-internal)

awallace-cray:
This enables Cray-internal testing to continue after one of
Cray's internal test machines was upgraded, and PBS on that
machine no longer supports the old "MPP" style of node selection.

This should have no effect on anyone else.

Thanks, ronawho.


awallace-cray:
Still fixing up the other side to match.

PR #9965, opened 2018-06-21
Simplify version button for user experience; search engine optimization

bradcray:
While thinking and researching about how to have search engines stop pointing to our old versions of documents so much, I realized that a problem that our approach has taken is that all versions of the documentation point to all other versions and therefore no version stands out as being particularly more than anything else.  This happens because of the version button which points to all other versions of a document (when it exists).  But when you think of it, most people using the version button are likely to only want to jump to the latest version or the master version.  Geeks who really want to look at some other old version are probably very willing to hack URLs to get there.

So... this PR simplifies the version button to only include the latest and master versions in the hopes of driving more traffic to the latest version when search engines crawl the page.  While here, it also:

* adds a parenthetical to all old versions saying "(old version)" because it hasn't been clear that the color coding by itself has been sufficient
* puts the latest version above the master version
* changes the wording of the buttons slightly to include the word "version" and to describe the choices
* has the "latest" version point to the top-level documents rather than the version in the 1.17 subdirectory (say).

These changes seem to simplify both the script itself and the user experience.  Hopefully others will agree.

### DIFF
--- a/doc/util/versionButton.php
+++ b/doc/util/versionButton.php
@@ -48,7 +48,6 @@
   overflow: auto;
   box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
   z-index: 1;
-  height: 150px;
   right: 0;
 }
 
@@ -84,7 +83,7 @@ var chplTitle = "<?php echo "$chplTitle";?>";
 
 // Note: assumes second element is most-recent release
 var chplVersions = [
-  "1.18 pre-release",
+  "1.18 (pre-release)",
   "1.17",
   "1.16",
   "1.15",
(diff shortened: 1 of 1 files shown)


bradcray:
@benharsh: Would you review?  Note that you can currently preview the experience live at chapel-lang.org for all versions other than the master version.

bradcray:
Noting belatedly: I could skip adding the 1.17 and 1.18 choices to the menu when those versions are the ones currently being viewed.

cassella:
In the description, you probably mean "... or the master version".

ben-albrecht:
👍 on this change. In the longterm, I think it'd be valuable to re-introduce a way to access older versions beyond URL hacking, but that is certainly more harmful than helpful today.

> In the description, you probably mean "... or the master version".

Should have used `chplspell` @bradcray ! (fixed)

bradcray:
@ben-albrecht : I was thinking of (eventually) adding some sort of "old documentation versions" link, perhaps unified with where we store old versions of the spec (i.e., generalize it to "archived documentation").  This would only provide one link back to each of the old versions, so should be much less harmful than a potential link per documentation page.

PR #9966, opened 2018-06-21
Add mason test command to Mason

Spartee:
This PR includes the `mason test` command which allows users to compile and run tests found in their `/tests` directory which is initialized for them upon project creation. The command compiles the tests and places the binaries in `/target/test`. 

Just like the binaries for source code, running `mason clean` will remove the test binaries.


tools/mason/MasonHelp.chpl
```diff
@@ -139,3 +141,18 @@ proc masonEnvHelp() {
   writeln("Environment variables set by the user will be printed with an");
   writeln("asterisk at the end of the line.");
 }
+
+proc masonTestHelp() {
+  writeln("Run test files located within target/debug/tests");
```
ben-albrecht: I think it'd be nice to stay consistent with `test/` or `tests/` for both the top-level source directory and the build directories.

test/mason/mason-test/PREDIFF
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
ben-albrecht: Assuming we use this `PREDIFF` in all mason tests, can this be a symlink to `../PREDIFF` to reduce future maintenance costs in keeping this file updated? (ditto for other uses of this `PREDIFF`)

test/mason/mason-test/mason-test.skipif
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
ben-albrecht: My previous symlink suggestion applies here as well.

test/mason/mason-test/test/sampletest2.chpl
```diff
@@ -0,0 +1,6 @@
+
+proc test() {
```
ben-albrecht: Can we add usage of the `FizzBuzz` library to demonstrate it is accessible when compiled?  Maybe define a simple function in `FizzBuzz` and call it here.

tools/mason/MasonTest.chpl
```diff
@@ -0,0 +1,99 @@
...
+        }
+        else {
+          if show then writeln("compiled ", test, " successfully");
+          const binCommand = projectHome + '/target/test/' + binary;
+          runCommand(binCommand);
+        }
+      }
+    }
+    else {
+      writeln("No tests were found your Mason.lock");
```
ben-albrecht: I like that this is specific about what the error actually is, but I'm wondering if this detail will matter to the user. Would it be more meaningful to translate it to what it meant for the user, e.g. `"No tests were found in test/"`?

test/mason/mason-test/test/sampletest2.chpl
Spartee: see latest commit, should give enough usage examples of both using a dependency and using your own library in testing.

tools/mason/MasonTest.chpl
```diff
@@ -97,3 +106,24 @@ private proc runTests(show: bool) {
...
+
+    // Add dependencies to project
+    for (_, name, version) in sourceList {
+      var depSrc = "".join(' ',depPath, name, "-", version, '/src/', name, ".chpl");
+      compopts += depSrc;
+    }
+  }
+  return compopts;
+}
+
```
ben-albrecht: Lots of extra newlines here

doc/rst/tools/mason/mason.rst
```diff
@@ -154,6 +159,61 @@ To try out different values at runtime, pass the values for ``number`` to ``maso
    this argument, this command will run the executable over 4 locales.
 
 
+Testing your Project
+====================
+
+Testing a Mason project is easy! Start by adding test files to the /test directory and specifing
```
ben-albrecht: I think it's more conventional to refer to a subdirectory without the leading `/`, e.g. `test/` instead of `/test`.

Also, let's inline code this dir:

``` ``test/`` ```

File: doc/rst/tools/mason/mason.rst [hunk 8]
```diff
@@ -154,6 +159,61 @@ To try out different values at runtime, pass the values for ``number`` to ``maso
...
+    authors = ["Sam Partee <Sam@Partee.com>"]
+    tests = ["sampleTest.chpl"]
+
+    [dependencies]
+    curl = '1.0.0'
+
+
+When the ``mason test [options]`` command is invoked, mason will find and download the nessescary dependencies
+for your project that you listed in your Mason.toml and compile them with your main module found in
+``/src``. For example, after listing ``sampleTest.chpl`` as a test, the project structure would be
```
ben-albrecht: `/src` -> `src/`

doc/rst/tools/mason/mason.rst
ben-albrecht: `specifing` -> `specifying`.

Would you mind running `chplspell` on this file? :)

test/mason/mason-test/Mason.lock
```diff
@@ -0,0 +1,15 @@
+
+[root]
+name = "sampleModule"
+version = "0.1.0"
+chplVersion = "1.18.0..1.18.0"
+tests = ["sampletest.chpl", "sampletest2.chpl", "dependencytest.chpl", "nomodule.chpl"]
```
ben-albrecht: Can you add a test within a subdirectory since we now support that?

PR #9970, opened 2018-06-22
Update a test's .good files now that dsiReallocate() is parallel

bradcray:
Now that dsiReallocate() is parallel, it is adding extra output
to these tests which print out debugging information for forall
loops.  Note that the numa locale model uses a reallocation in
its setup.

I also found myself wondering whether we should skip this test
for numa (are we gaining much?) or should have finer-grained
execution time controls for squashing this debugging output
including only having it on for user code, but this was the
quickest resolution.


bradcray:
@ronawho: As reviewer of the PR that introduced the regression, are you OK with this approach to silencing it?

PR #9975, opened 2018-06-22
Parenthesize "(pre-release)" to match other version formatting changes

bradcray:
As a follow-up to PR #9965, this adds parenthesis to the "pre-release" label for 'master' documentation.


bradcray:
@ben-albrecht: I'd missed that this would also be necessary until after the version button changes hit the website...  I think this should be safe but suspect I won't be sure until it hits the web as well.  Look OK to you, based on your understanding of what our sphinx does?

(I feel pretty confident that the version button will do the right thing since it's using substring searches, but am tagging @benharsh as well just in case).

PR #9976, opened 2018-06-22
Eliminate unnecessary array creation for in-intents

benharsh:
Prior to this PR wrapper-creation during resolution would create an unnecessary instance of the formal's default value if that formal satisfied all of the following conditions:
- formal was for a default-initializer
- formal was for an array field
- array field did not have an explicit type

For example:
```chpl
record R {
  var A = [1, 2, 3];
}
var r = new R();
```

The cause of this extra array was the insertion of ``PRIM_TYPEOF`` at the end of the formal's typeExpr. Wrapper creation identified this typeof and assumed that the formal had a user-defined type associated with it. In order to preserve the runtime type, the compiler would create two instance of the default value, when only one was necessary. One of these arrays would then be leaked. See ``types/records/ferguson/array/array-in-record.chpl`` for an example.

The solution is to not insert PRIM_TYPEOF for the formal's typeExpr, and simply copy the formal's defaultExpr.

This change exposed an additional bug with wrapper creation in which a default value with a ``ref`` kind would not be copied when the call was invoked:

```chpl
var A : [1..10] int;
proc helper() ref {
  return A;
}

proc foo(in X = helper()) {
  X[1] = 100; // modifies A!
}
```

To resolve this bug, wrapper creation is updated to insert a copy of the default value if the result is a reference and the formal does not have any kind of ``ref`` intent. A new test is added to ensure we catch this kind of bug in the future.

Testing:
- [x] local + futures
- [x] gasnet
- [x] --force-initializers (no new failures)
- [x] memleaks
- [x] valgrind

### DIFF
--- a/compiler/AST/AggregateType.cpp
+++ b/compiler/AST/AggregateType.cpp
@@ -1865,36 +1865,6 @@ void AggregateType::fieldToArg(FnSymbol*              fn,
         // Type inference is required if this is a param or variable field
         //
         } else if (defPoint->exprType == NULL && defPoint->init != NULL) {
-          VarSymbol* tmp      = newTemp();
-          BlockStmt* typeExpr = new BlockStmt(new DefExpr(tmp), BLOCK_TYPE);
-
-          // tmp->addFlag(FLAG_INSERT_AUTO_DESTROY);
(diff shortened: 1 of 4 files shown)


benharsh:
@mppf : would you review?

PR #9977, opened 2018-06-22
Add an iterator to read HDF5 files in chunks

daviditen:
This iterator supports reading a dataset from an HDF5 file in chunks
and yields arrays with those chunks.  It currently supports reading
an N-dimensional array into smaller N-dimensional arrays with
in N-dimensional blocks.  The 1-D case is currently handled
separately from the general N-D case, but could potentially be
combined.

Add a test that reads an input file in 8 element chunks.
Add a test that reads a 2D input file in 5x5 element chunks.

A possible next step is to read N-D data into linearized 1-D blocks.


modules/packages/HDF5_HL.chpl
```diff
@@ -3681,6 +3681,149 @@ module HDF5_HL {
       }
     }
 
+    /* Read the dataset named `dset` from the HDF5 file named `filename`.
+       The dataset consists of elements of type `eltType`.  The file will
+       be read in chunks matching the `chunkShape` domain, and yielded as
+       arrays of that size until the end of the dataset is reached.
+
+       Currently, the `chunkShape` domain describing the output arrays and
+       the shape of the data in the file must both be the same rank.
```
vasslitvinov: I suggest adding something like "For example, if the data in the file is 2D, `chunkShape` must be 2-dimensional, too."

Issue #9785, opened 2018-06-09
Should enum->int casts be supported only for enumerated types with explicit values?


bradcray:
In thinking about enums being 0-based vs. 1-based by default this morning when casting them to integers, it occurred to me that a better solution might be to not have enums be anything-based by default.  This would be easy to enforce by only supporting casts to integers for enums that are given values.  Thus, given:

```chapel
enum color {red, green, blue};
enum day {monday=0, tuesday, wednesday};
enum size {small=1, medium, large};
```

a value of type `color` could not be cast to int since no integer values had been given for any of its items, whereas a value of type `day` or `size` could.  The benefits of this choice would be that it would remove a place in the language where we may 0- vs. 1-based decisions on the programmer's behalf while also strengthening the enum abstraction for users who do not want to think of them as having underlying integer values.

(Note that all enums would still have an integral order still which is currently 0-based)

mppf:
My opinion: yes

bradcray:
Consensus here seemed to be that this change makes sense.  It seems to provide the best of both worlds:  Enums that have no relation to integers for those who want them to be a very abstract type and those that can be mapped to integers for those who want more of a C-like capability.

PR #9978, opened 2018-06-22
Make associative domains of enums more standard

### LINKED ISSUE #9787: Should associative domains of enums be empty by default?
Most associative domains are empty by default upon declaration.  At present, associative domains of enums are not, for historical reasons.  Instead, they are fully populated with all of the enums values.  This issue asks whether they too should be empty by default for consistency.

The historical reason why we made associative domains of enums full of all enums by default is as follows:  When enums coerced to integers, an enumerated range or domain like:

```chapel
enum color{ red=1, orange, yellow, green, blue, indigo, violet};
const er = color.red..color.violet
const ed = {color.red..color.violet};
```

would essentially become the integer range `1..7` or domain `{1..7}`.  As a result, arrays over all enums were difficult to express using rectangular domains.  The alternative was to use an associative domain, as these preserved the index type.  But after writing a few of them, we realized the pattern was always "declare the associative domain, populate it with all symbols from the given enum, then use it" so this felt like a common enough case to optimize for, both in terms of the convenience of not having to populate it and for the knowledge that if all values in an enum were used as indices, the implementation could use dense/linear storage under the covers.

Now that we're on the brink of supporting true enum ranges and arrays over the same (e.g., PR #9781), it seems that we should perhaps make things more uniform by using rectangular domains of enums for the dense/regular/fully populated cases and making associative domains of enums more consistent with other associative domains (in starting early and potentially unifying the implementation).

### LINKED ISSUE #9822: Spike: Explore making associative domains of enum empty by default
As a Chapel user, it seems inconsistent to me that most associative domains are empty by default while associative domains of enums are full by default.  While this might support a common pattern, the inconsistency seems more troubling than the inconvenience of populating it myself, and now that I can create arrays over ranges of enums, it also feels less helpful.

Acceptance criteria:
* gain consensus on whether or not to make associative domains of enums empty by default
* understand the implications for existing code and the implementation
* create a series of follow-on tasks to implement the feature if we go ahead with it.

bradcray:
In the past, associative domains of enums have been fully populated because (a) this was the only way to create an array indexed by an enum and (b) it seemed to support a common-case.  Now that PR #9781 has introduced rectangular domains indexed by enums, it doesn't seem to make sense to continue making associative domains of enums different in this respect (see issue #9787).  As a result, this PR takes away special handling for associative domains of enums.

One of the biggest impacts of this change, beyond making the domains empty by default, is that iterating over enum domains/arrays used to be done in a well-defined order before but now results in a random ordering.

Most of this effort involved removing special case code for associative-of-enum cases from the modules.  I also made enums hash based on their order rather than a cast-to-int (given that we're considering removing support for enum->int casts in some cases in issue #9785 ) and fixed a remaining problem where we were using `idxType` instead of `intIdxType`.

I updated the spec to remove special case language for associative domains of enums.

Most of the rest of the changes involved updating tests.  In some cases, I rewrote the test to use a rectangular array of enums if it just wanted an enum->value map.  For tests that were specifically testing enum domains, I sometimes inserted sorted() routines to preserve the output and sometimes updated the output to reflect the new order (and sometimes a bit of both).

resolves #9787 
resolves #9822 

One thing I wonder about specifically w.r.t. this PR is whether we should generate a warning for the 1.18 release when users compile associative domains of enums to let them know that the default value (and behavior) has changed.  On the plus side, it would head off some surprises; on the minus side, it's noisy (e.g., would affect a lot of our tests).


bradcray:
@daviditen (LCALS, diten tests) and @mppf (ferguson tests) have tests that were updated w.r.t. this PR and may want to review them as a result.

mppf:
the change to the test with my name on it looks great

daviditen:
The changes to my tests look good.

Just curious - in several tests (e.g. test/npb/cg/bradc/cg-makea1-small.chpl) you changed things that were initialized with tuple literals to use array literals instead.  Was that strictly necessary, or just a "while I was there" improvement?  I think it's better with array literals either way.

bradcray:
@daviditen : At first I thought it was necessary, but then it turned out to be more of "while I was there" kind of thing when I realized that the tuples did still work.  And I moved away from both when order matters (since we can't count on the ordering of an enumerated domain or array any more.

mppf:
> One thing I wonder about specifically w.r.t. this PR is whether we should generate a warning for the 1.18 release when users compile associative domains of enums to let them know that the default value (and behavior) has changed. On the plus side, it would head off some surprises; on the minus side, it's noisy (e.g., would affect a lot of our tests).

I think we should simply add it to `--warn-unstable`.

spec/Domains.tex
```diff
@@ -385,27 +375,7 @@ \subsubsection{Associative Domain Values}
...
-In addition, fully-populated constant enumerated
-domains are an important case for compiler optimizations, particularly
-if the numeric values of the enumeration are consecutive.
-\end{rationale}
-
-\begin{future}
-We may generally support a \chpl{startPopulated} parameter on associative
-domains, to unify this capability with other values.
-\end{future}
+index set.
```
mppf: I find this sentence (If unspecified the default value...) really hard to read. Maybe it just needs a comma. Could you make it more readable somehow, while you're here?

bradcray:
> I think we should simply add it to `--warn-unstable`.

I wondered about this.  I originally thought of this flag as "Let me know if I'm using things that might change in the future."  Are you also using it for "Let me know about things I'm using that have recently changed"?  (or is my understanding of it otherwise wrong?).  Thanks.

mppf:
> I wondered about this. I originally thought of this flag as "Let me know if I'm using things that might change in the future." Are you also using it for "Let me know about things I'm using that have recently changed"? (or is my understanding of it otherwise wrong?). Thanks.

I'm using it as "Let me know about things I'm using that are changing". I view the delete-free stuff as an ongoing change (that will last probably > 1 release). The first case I used it for was something that warned when somebody returned an iterator (since now it turns into an array, but it didn't before). All of these are more something that's already changed/changing vs. will change (in my mind).

bradcray:
@vasslitvinov: Note that `--warn-unstable` may also be useful for the discussion we were having about the user-defined reduction interface last week.

bradcray:
@mppf: I've added `chpl_warnUnstable` to reflect the `--warn-unstable` flag's setting in the module code and used it to generate a warning when associative domains/arrays of enums are used.  For some reason the warning is getting printed out twice, but I don't think that's my fault, and think 2 warnings are better than 0 (and given that it's off by default...).  I added a test to lock in this behavior.

I also reworded that sentence in the spec (as well as added the comma).

I consider these changes pretty trivial, but would appreciate a post-merge review, if you'd be willing to give it.

mppf:
Later commits look good to me, thanks!

PR #9981, opened 2018-06-22
Better line numbers in errors for fields with default values

benharsh:
Prior to this PR the compiler was reporting incorrect line numbers for errors involving the default value of a field. For example, in the following program the compiler would print the error at line ``5`` instead of the correct line (``3``).
```chpl
/*1*/ pragma "use default init"
/*2*/ record R {
/*3*/   var x : int = 2.0;
/*4*/ }
/*5*/ var r : R;
```

The problem was that the compiler would not use the line number for functions that create default-values, and would walk up the call stack until something more 'appropriate' was found. The compiler rejected the default-value function because it was 'compiler generated'.

This PR adds a new flag, FLAG_LINE_NUMBER_OK, which informs the compiler that the function's line number is OK regardless of whether it is compiler-generated. This new flag is now applied to all default-value functions.

This resolves a --force-initializers failure for the test ``classes/diten/error_default_field_init.chpl``.

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -1539,20 +1539,14 @@ userCall(CallExpr* call) {
   // modules, back up the stack until a call is encountered whose target
   // function is neither.
 
-  CallExpr* cur = call;
-  int i = callStack.n-1;
-  while (cur != NULL && i >= 0 && shouldSkip(cur)) {
-    cur = callStack.v[i];
-    i -= 1;
+  if (shouldSkip(call)) {
+    for (int i = callStack.n-1; i >= 0; i--) {
+      CallExpr* cur = callStack.v[i];
+      if (!shouldSkip(cur))
+        return cur;
+    }
   }
-
-  // If we could not find a suitable call in the callStack, return the original
-  // call.
-  if (shouldSkip(cur)) {
-    return call;
-  } else {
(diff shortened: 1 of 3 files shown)


compiler/resolution/functionResolution.cpp
```diff
@@ -1525,16 +1538,21 @@ userCall(CallExpr* call) {
...
-          callStack.v[i]->getModule()->modTag != MOD_INTERNAL)
-        return callStack.v[i];
-    }
+
+  CallExpr* cur = call;
+  int i = callStack.n-1;
+  while (cur != NULL && i >= 0 && shouldSkip(cur)) {
+    cur = callStack.v[i];
+    i -= 1;
+  }
```
mppf: I've been doing the transformation from the old code to the refactored code in my head and I'm worried there's an off-by-one error in the call stack accesses. I don't think there is but it's harder to follow than the old version. In some ways I'd prefer the structure

``` c++
if (shouldSkip(call)) {
  for (int i = callStack.n-1; i >= 0; i--) {
     CallExpr* cur = callStack.v[i];
     if (!shouldSkip(cur))
       return cur;
  }
}
return call;
```

benharsh:
I'm going to run another local paratest just to be safe.

PR #9989, opened 2018-06-24
Minor tweaks to support upcoming changes

vasslitvinov:
* Store "zippered" in PromotionInfo instead of heuristically recalculating it.

* Allow PRIM_SET_MEMBER et al. to contain field symbols, in addition to
  field names, during pruneResolvedTree().

* Move zipToTuple() out from chapel.ypp.

* Indicate to the compiler that chpl__serialize() updates 'data'. Note 1.

* Updated a comment. Note 2.

Parser files will be committed separately.

Testing:
* [x] linux64 --verify
* [x] gasnet multilocale tests

Optional notes:

Note 1. In chpl__serialize(), the compiler did not realize that 'data'
could be modified. So it could potentially do incorrect copy propagation.
To prevent that, create a clone of chpl__getInPlaceBufferData() that
takes it argument ('data') by ref, instead of const ref.

Note 2. Update a comment in copyPropagation.cpp .
It was initially added in 31a8a95ca47. Since then, things have changed.
Now, if we remove the checks suggested in the comment, we will
overlook those cases where our Symbol gets modified, so it may be
const-propagated incorrectly.
Conversely, the intents for 'this' argument have improved since the comment.
So maybe the intent checks will work instead of the special case.
Remember to check for the ref maybe-const intent, too.

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -199,6 +199,19 @@ static const char* toImmediateString(Expr* expr) {
 }
 
 
+//
+// This is a simple utility function that converts new-style
+// PRIM_ZIP() expressions to old-style _build_tuple() expressions for
+// code that is not ready for the new style yet.
+//
+static void zipToTuple(Expr* zipPrim) {
+  CallExpr* zipExpr = toCallExpr(zipPrim);
+  assert(zipExpr->isPrimitive(PRIM_ZIP));
+  zipExpr->primitive = NULL;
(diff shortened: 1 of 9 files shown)


vasslitvinov:
@benharsh would you review?

I will remove the "vass" assertions  before merging.
They were good for testing, however I do not want them in the production code.

vasslitvinov:
Replaced by #10132

Issue #7777, opened 2017-11-10
Initializing a variable from forall expressions is serial


mppf:
For example:

``` chapel
config const n = 20;

proc notparallel() {
  var x = [i in 1..n] i;
  writeln(x);
}

proc parallel() {
  var x:[1..n] int;

  forall i in 1..n {
    x[i] = i;
  }

  writeln(x);
}

notparallel();
parallel();
```

By setting a breakpoint on chpl_task_addToTaskList I can observe that no tasks are created in `notparallel`.

The issue here is that the initCopy from an iterator to create an array uses a `for` loop to enumerate the yielded elements and add them to the array. To do better, it would need the array's bounds... or possibly a very clever strategy.

bradcray:
This seems somewhat related to #5958 (having sizes of arrays inherit the domain when initialized with a promoted expression) to me.

mppf:
@bradcray - I agree. One way to interpret this issue is indication that lack of a solution to #5958 is both a productivity and a performance issue.

vasslitvinov:
notparallel() now executes in parallel, according to the chpl_task_addToTaskList() check.

Due presumably to #11501.

Closing this as resolved.

bradcray:
Awesome!

Issue #5958, opened 2017-04-06
make promotion and iterators preserve shape/domain


vasslitvinov:
We would like promotion to produce the result of the same shape as the expression being promoted. For example, `A*5` , where `A` is an array, should produce an array of the same domain as `A`.

When multiple expressions are promoted, the shape should be determined by the first expression. For example, `A+B` , where A and B are arrays, should produce an array of the same domain as `A`. This corresponds to the "leader" in a zippered forall loop.

The case that needs design work is when the "leader" is an iterator: #11524.

tmacd9:
What is the domain of C below?

```
config const f1 = true;
config const f2 = false;
var A1: [0..2, 0..2] int;
var A2: [1..3, 1..3] int;
var B1: [2..4, 2..4] int;
var B2: [3..5, 3..5] int;

var C = (if f1 then A1 else A2)  + (if f2 then B1 else B2);
writeln(C);

```

bradcray:
> What is the domain of C below?

In the past, we've proposed that for things that are zipped together (the two conditional expressions in this case), the leader's domain would be taken.  So if f1 was true, it would be A1.domain; if false, A2.domain.

ben-albrecht:
@vasslitvinov - should this be closed?

vasslitvinov:
@ben-albrecht the last paragraph is not implemented, and we do not have a plan how to do it. This prevents me from closing this.

bradcray:
I'm having trouble parsing the last paragraph.  What's a code example that would illustrate it?

bradcray:
(or future if one already exists)

vasslitvinov:
@bradcray I just updated the issue text, except for the first two paragraphs. How much sense is it making now?

vasslitvinov:
The iterator case is now #11524. The rest of this issue has been implemented - closing it.

PR #9994, opened 2018-06-25
Preserve promoted- and forall-expressions' shapes

vasslitvinov:
This PR adds a `_shape_` field to _iteratorRecord created for promoted expressions and forall-expressions and uses it when creating an array out of it.

At the moment this is not a user-facing feature. That is discussed in #10359 .

Addresses a good part of #5958 .
Future work: should a range have a shape? how? See #7777.

### How does it work?

Consider a promoted + on two arrays:

```chpl
var A = B + C;
```

During resolution, this is converted to:

```chpl
    iter +(arg1,arg2) {
      forall (a,b) in zip(arg1,arg2) do
	yield a+b;
    }

    var temp = +(B,C);
    A = chpl_initCopy(temp);
```

where chpl_initCopy() is one the one in ChapelArray.chpl
that takes an _iteratorRecord and creates a 1-D array
out of the things that the corresponding iterator yields.

This PR makes resolution extend the `_iteratorRecord` type of the `temp`
with a `_shape_` field and add the following statement:

    var temp = +(B,C);
    temp._shape_ = B.domain._instance;  <---------
    A = chpl_initCopy(temp);

This is done using the newly-added PRIM_ITERATOR_RECORD_SET_SHAPE
to pass information from one part of resolution to another.

chpl_initCopy() now checks whether the _iteratorRecord argument
has a meaningful shape. If so, it creates an array over that shape
instead of the 1-D array.

While ideally we would like to represent loops and loop expressions
using dedicated AST nodes, this PR achieves shape preservation
for promoted expressions with only small compiler modifications.

### What's in a `_shape_` ?

`_shape_` keeps a pointer to a BaseDom-derived class, instead
of a ref to a `_domain`. It is simpler because the alternative
requires a ref field, which it is not obvious would work. On the
downside, we need to create a `_domain` and set its `_unownend`
to true. We can revisit this when desired.

Looks like the lifetime of the stored class instance should be OK.
That's because the promoted- or forall-expression's lifetime
should be nested in the lifetime of the source expression(s)
that it is computed from.

### Test Changes

* Un-futurized + gave a better name + moved to a better place -- this future requesting to preserve shapes:

    `arrays/tmacd/promotionLosesArrayShape`
    -->
    `arrays/shapes/tmacd-promotion`

* Updated .good to shape-preserved output:

    `distributions/robust/arithmetic/basics/test_promotion3`
    `studies/matrixlib/test_overloading`

* "While there": looks like this test now produced the desired output now. Un-futurize it:

    `arrays/bradc/sliceViaSingleton`

* Switch to IMO-more-elegant forall-expr-based style, for fun:

    `distributions/bharshbarg/stencil/reshape`

* Futurized this test as discussed with Michael and Brad; renamed to match a very similar variant:

    `users/vass/km/array-of-records-crash-1`
    -->
    `users/vass/km/array-of-records-crash-1.inferred-type`

* Pleasantly, remove ugly-looking adjustment that were needed due to shape divergence:

    `types/atomic/sungeun/atomic_vars`

### Miscellaneous

* Moved	resolveExpr() to passes.h to avoid increased header dependence on resolution.h or iterator.h.

* While there, improved the debugging function debugGetTheIteratorFn().

Development history: 62c6da0249..ac271d3c70


bradcray:
Noting the relationship between this PR and issue #5958 (a.k.a., the issue Brad would most like to see fixed).

ben-albrecht:
This PR will be huge for users of the `LinearAlgebra` library, so I'm really happy to see it progress.

While excited for this change, I have one concern about a design decision -- It's unfortunate for an operation such as `A + B` to lose commutability w.r.t. return type, i.e. the return type differs for `A + B` vs. `B + A`. I could see this pattern being a source of confusion/bugs for users. Is this pattern exhibited anywhere else in the language?

That said, I can't come up with any great alternatives, so I can see how we arrived at that decision. Just wanted to express the concern, and see if others had thoughts on it.

bradcray:
> Is this pattern exhibited anywhere else in the language?

I think the answer is likely "everywhere that promotion applies."  Can you offer a sensible counterproposal?  (I can't).

ben-albrecht:
> I think the answer is likely "everywhere that promotion applies." 

I guess I hadn't considered the order-dependent return type as a general feature of promotion. When thinking of it that way, it doesn't seem so bad, since we're consistent with how promotion behaves.

> Can you offer a sensible counterproposal? (I can't).

I cannot. Just wanted to make sure that tradeoff was acknowledged.

bradcray:
> Just wanted to make sure that tradeoff was acknowledged.

Is it a tradeoff when there's no alternative to compare against?  :)

Note that nothing prevents a user (or linear algebra library) from writing their own array `+` overload and defining the semantics as they'd like (e.g., "When adding a Block-distributed array and a Cyclic-distributed array, I will return a Block-Cyclic distributed array!" :) ).

vasslitvinov:
BTW array and domain already support `.shape` method, see #5178.
If we turn _shape_ into a user-facing feature, we will need to give is a different name.

mppf:
AFAIK this addresses issue #7777

compiler/resolution/wrappers.cpp
```diff
@@ -2047,6 +2047,28 @@ static void       buildFastFollowerCheck(bool                  isStatic,
...
+  // If call's result is not used, do not insert.
+  // This happens, for example, during resolveSerializeDeserialize().
+  if (move == NULL) return;
+  INT_ASSERT(move->isPrimitive(PRIM_MOVE));
+  Symbol* irTemp = toSymExpr(move->get(1))->symbol();
+
+  // Which of call's arguments determines the shape?
+  Symbol* shapeSource = NULL;
+  int i = 0;
+  for_actuals(actual, call)
```
mppf: Can I have some { } around the body of this loop, please+

compiler/resolution/resolveFunction.cpp
```diff
@@ -567,6 +567,7 @@ static void insertUnrefForArrayOrTupleReturn(FnSymbol* fn) {
           VarSymbol* tmp       = newTemp(arrayUnrefName, rhsType);
           CallExpr*  initTmp   = new CallExpr(PRIM_MOVE,     tmp, rhs);
           CallExpr*  unrefCall = new CallExpr("chpl__unref", tmp);
+          CallExpr*  shapeSet  = toCallExpr(call->next);
```
mppf: I really don't like relying on AST patterns with elements that happen to be next to each other. What if error handling (say) adds another call in between? Is there some reason you know such things can't happen? Is there an alternative representation or way to find the relevant primitive?

compiler/resolution/lowerIterators.cpp
```diff
@@ -99,9 +99,15 @@ FnSymbol* getTheIteratorFnFromIteratorRec(Type* irType)
 FnSymbol* debugGetTheIteratorFn(Type* type) {
   FnSymbol* result = NULL;
   if (AggregateType* agg = toAggregateType(type)) {
-    if (agg->symbol->hasFlag(FLAG_TUPLE))
-      result = getTheIteratorFn(agg);
-    else if (agg->symbol->hasFlag(FLAG_ITERATOR_CLASS))
+    if (agg->symbol->hasFlag(FLAG_TUPLE)) {
+      // Sometimes the first field of the tuple is its size.
```
mppf: I think that's always true during resolution, isn't it?

compiler/AST/iterator.cpp
```diff
@@ -108,6 +109,63 @@ removeRetSymbolAndUses(FnSymbol* fn) {
...
+  resolveExpr(compute);
+  resolveExpr(set);
+
+  AggregateType* iRecord = toAggregateType(ir->type);
+  INT_ASSERT(iRecord->symbol->hasFlag(FLAG_ITERATOR_RECORD));
+  Symbol* field = iRecord->getField("_shape_", false);
+  if (field == NULL) {
+    field = new VarSymbol("_shape_", value->type);
+    iRecord->fields.insertAtTail(new DefExpr(field));
+    // TODO: what should accessor's return intent be?
```
mppf: `const ref`, right?

mppf: I'm a little surprised you bothered to run build_accessor just to call it. Why not just use PRIM_GET_MEMBER_VALUE say?

compiler/AST/LoopExpr.cpp
```diff
@@ -214,6 +227,33 @@ void lowerLoopExprs(BaseAST* ast) {
...
+  CallExpr* move = toCallExpr(tupleDef->parentExpr);
+  INT_ASSERT(move->isPrimitive(PRIM_MOVE));
+  CallExpr* buildTup = toCallExpr(move->get(2));
+  INT_ASSERT(buildTup->isNamed("_build_tuple"));
+  // The shape comes from the first tuple component.
+  return buildTup->get(1);
+}
+
+// 'feRep', during resolution, is an iterator record for the
+// forall expression. Ensure it will get a shape.
```
mppf: feRep is a call so it can't possibly be an iterator record, which is a type.
In any case I think you could come up with a better variable name. I don't think the abbreviation here is helping.

compiler/AST/LoopExpr.cpp
```diff
@@ -204,6 +211,12 @@ bool LowerLoopExprVisitor::enterLoopExpr(LoopExpr* node) {
     // The iterator expr might be a loop-expr itself, make sure it gets
     // lowered.
     iterExpr->accept(this);
+
+    // Do not preserve the shape if there is a filtering predicate.
+    if (node->forall && noFilter) {
+      normalize(replacement); // for addIterRecShape()
```
mppf: it seems preferable to me for addIterRecShape to deal with whatever form of AST it needs to handle and generate right here instead of adding the new normalize call.

vasslitvinov:
> AFAIK this addresses issue #7777

Turns out it doesn't.

Because the forall expression in the #7777 example uses a range. In this PR, ranges do not get shapes.

That's because here a shape is a domain. For a range, we'd have to create a domain. This PR does not do this. We probably could.

Better yet, how about make the shape of a range be that range? That would be more efficient. We'd need to adjust other parts of the implementation to handle that.

compiler/resolution/resolveFunction.cpp
vasslitvinov: Good point. We could have a hash table from each move into the ret symbol to the corresponding shape thing. (There could be several such moves per function, ex. if the user has multiple return statements.) This way of tracking dependencies would come handy in other cases as well. We are not set up for it right now because such a hash table would need to be updated whenever a function - or its part - is cloned.

Instead, how about I write a little helper that traverses all statements from `call->next` on, until it finds a PRIM_ITERATOR_RECORD_SET_SHAPE ?

compiler/resolution/lowerIterators.cpp
vasslitvinov: Good point. I used to call this helper post-resolution, so this had not been an issue. It becomes an issue when I call whocalls() during resolution, which calls debugGetTheIteratorFn(), which crashes because of the `size` field.

With the above change it doesn't crash any longer.

I will update the comment to clarify: `Through the end of resolution, the first field is the tuple size.`

compiler/AST/iterator.cpp
vasslitvinov: > `// TODO: what should accessor's return intent be?`

If the field stores a class (under BaseDom), as is the case in this PR, it is simpler to return it by value, rather than by const ref, as you suggest. However, it is not a big deal, and we may store something else, not a class, in the future. So I will remove the comment.

vasslitvinov: > I'm a little surprised you bothered to run build_accessor

It is convenient to get `_shape_` from Chapel code. See the module changes in this PR.

To get `_shape_` in Chapel, gotta have an accessor.

compiler/resolution/resolveFunction.cpp
mppf: I'm OK with the helper, yeah

compiler/AST/iterator.cpp
mppf: That seems reasonable, then. Might be worth putting a comment to why we are building the accessor.

compiler/AST/LoopExpr.cpp
mppf: (scratch that, I was missing that LowerLoopExprVisitor runs during normalize)

Issue #9106, opened 2018-04-05
Heap allocation of stack variables is buggy


vasslitvinov:
Background: when needHeapVars()==true, a variable that would normally live on stack needs to be heap-allocated - if it can be referenced on a remote locale. The compiler makes this transformation in makeHeapAllocations() during the 'parallel' pass.

**The bug is**: some variables that need to be heap-allocated are not.

It is unclear how this affects real programs. As of 2018-04-04, the claim that the bug exists is based on Vass's examining the generated C code for a simple test and by examining and tracing through the compiler code in makeHeapAllocations(). needHeapVars()==true happens only in a couple of configurations, which we do not test much. A full paratest of one such multilocale configuration did not show any memory access-related issues.

### Configurations

* One configuration where needHeapVars()==true is:

```
CHPL_COMM=gasnet
CHPL_GASNET_SEGMENT=fast
CHPL_TARGET_ARCH=none
CHPL_TASKS=fifo
```

The full printchplenv dump from the aforementioned paratest is:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: none
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: fast
CHPL_TASKS: fifo
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: none
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none
```

It would be great to set also CHPL_MEM=cstdlib so we can use valgrind. However, while in this configuration everything compiles, running the program gives this "error: Your CHPL_MEM setting doesn't support the registered heap required by your CHPL_COMM setting. You'll need to change one of these configurations."

* Another configuration for needHeapVars()==true is our valgrind testing configuration:

```
CHPL_COMM=none
CHPL_HOST_PLATFORM=linux64
CHPL_MEM=cstdlib
CHPL_TARGET_ARCH=none
CHPL_TASKS=fifo
compile with --no-local
```

This does not expose incorrect remote memory accesses, however.

#### What is going wrong?

[...]

PR #10010, opened 2018-06-26
Do not introduce ref temps

vasslitvinov:
Avoid introducing a ref temp when passing a value to a function by reference.
For example - before this change:

     def ref _ref_tmp_[1705731]:_ref(bool)[648011]
     move( _ref_tmp_[1705731] addr of( rootLocaleInitialized[94602] ) )
     call( fn =[778] _ref_tmp_[1705731] false[19] )

with this change:

     call( fn =[778] rootLocaleInitialized[94602] false[19] )

Ref temps are no longer introduced:
* by insertReferenceTemps() (now removed)
   - it was called from callDestructors.cpp and parallel.cpp
* when passing outer vars into flattened functions in addVarsToActuals()
* for the return-by-ref temp in ReturnByRef::transformMove()

Some ref temps are still inserted elsewhere in the compiler.
Avoiding those is left as future work.

Here are some things that needed adjustment to the new representation:

* removeDeadStringLiteral() is adjusted to the new pattern.

* In inferConst() in inferConstRefs.cpp, in the case
`FnSymbol* fn = call->resolvedFunction()` - check if the arg
is passed to init(), in addition to the exiting check for FLAG_RETARG.
The rest of the changes in inferConstRefs.cpp are code refactoring.
In particular. the functions onlyUsedForRetarg() and onlyUsedForInitializer()
each used to make its own pass over for_SymbolSymExprs. Now they are combined
into onlyUsedForInitOrRetarg(), which does a single pass for_SymbolSymExprs.

* denormalize() needs to distinguish passing by value and passing by ref
based on the formal's ref-ness, not the actual's ref-ness.

* parallel.cpp was using ref-ness of the actual to distinguish between
passing by ref and passing by default intent with ref semantics. It now
uses ref-ness of the type of the formal. Straightening it out is left
as future work.

* Adjusted ref detection in makeHeapAllocations(). Todo: #9106.
While there, factored out needHeapVars() in findBlockRefActuals().

* Adjusted ref detection when creating argument bundles for recursive
iterators in lowerIterators.cpp. Some ref temps are still inserted there.

* I removed a small piece in isDefinedAllPaths() in checkResolved.cpp.
That code was not invoked any longer in any of our tests.

* I removed createRetargTemps() in insertWideReferences.cpp.
Performance spot-check - thanks to BenH - did not show regression.
We may need to come back to this - and adjust it to the new representation -
if there are performance regressions elsewhere.

Detailed development history: 08bee20f37..94677d6447

Testing:

- [x] linux64 --verify
- [x] baseline
- [x] memleaks
- [x] valgrind
- [x] perf playground


mppf:
I'm feeling confused. Looking at the diff, I don't see where the compiler changes to "the new representation" except for in flatten functions. I would have expected changes to normalize and/or function resolution to do what you're describing in the PR message.

vasslitvinov:
>  I don't see where the compiler changes to "the new representation"

@mppf - good question. The key change is in callDestructors.cpp in the first non-trivial commit: https://github.com/chapel-lang/chapel/pull/10010/commits/8ab2d8fabb626a4c1473916601527d6d9584386b

The rest of this PR is to bring the rest of the compiler up to speed with that.

mppf:
Given your response, I'm happy with the general idea / direction.

benharsh:
Have you done any --no-local or 16-node-xc performance testing to ensure that it is OK to remove that portion of insertWideReferences? That part of IWR concerns wide-pointer overhead and not comm-counts. A regression there might not come up in correctness testing.

vasslitvinov:
Not ready to merge yet.

vasslitvinov:
Ready for review.

benharsh:
I think this branch should spend a night or two in the performance playground before being merged.

compiler/optimizations/copyPropagation.cpp
```diff
@@ -326,7 +326,7 @@ static bool isUse(SymExpr* se)
     ArgSymbol* arg = actual_to_formal(se);
 
     if (arg->intent == INTENT_OUT ||
-        arg->intent == INTENT_REF)
+        (arg->intent & INTENT_FLAG_REF))
     {
       return false;
     } else if (arg->intent == INTENT_CONST_REF &&
```
benharsh: Will this branch run anymore, given that the previous one will catch INTENT_CONST_REF?

compiler/optimizations/inferConstRefs.cpp
```diff
@@ -460,77 +460,64 @@ static bool inferConstRef(Symbol* sym) {
...
+{
+  if (form->hasFlag(FLAG_RETARG))
+    return true;
 
-  for_SymbolSymExprs(use, ref) {
-    if (use->parentExpr == defCall) {
-      continue;
-    }
+  // vass - remove this
+  INT_ASSERT(!strcmp(calledFn->cname, "init") == (calledFn->cname == astrInit));
```
benharsh: Should this still be here?

compiler/optimizations/inferConstRefs.cpp
```diff
@@ -460,77 +460,64 @@ static bool inferConstRef(Symbol* sym) {
...
-    if (call->isResolved()) {
-      ArgSymbol* form = actual_to_formal(use);
-      if (form->hasFlag(FLAG_RETARG) == false) {
-        isRetArgOnly = false;
-      }
-    } else {
-      isRetArgOnly = false;
-    }
-  }
+  if (calledFn->cname == astrInit &&
```
benharsh: I think you can use ``calledFn->isInitializer()`` instead.

PR #10014, opened 2018-06-26
Check/track array memory allocation from the comm layer.

### LINKED ISSUE #9497: memory acquired from the comm layer isn't tracked; should be
As a Chapel programmer, I need array and heap extension memory to be tracked, so that I can tell what is allocated, freed, and leaked.  It turns out that this doesn't occur on Cray X* systems when memory is acquired from the comm layer and dynamically registered.  Please fix this.

gbtitus:
When array memory was gotten from the comm layer, we weren't doing the
memhook calls that provide for out-of-memory reporting and memory
tracking.  Add these.  As part of this it turned out that the only
practical way to avoid doing the memhook calls twice in some cases was
to change the fallback calls used when comm-layer allocation isn't done,
so they go directly to the layer implementation interface rather than to
the public memory interface.

While here, I also removed the logic that dealt with localization when
the memory came from the comm layer, because I concluded that there was
no execution path along which the localization call actually occurred.

This resolves #9497.


ronawho:
Add a correctness test with some large arrays (maybe a stream variant) that runs with `--memLeaks` or `--memTrack --memStats` or something to runtime/configMatters?

Issue #9987, opened 2018-06-23
Collection of codes that generate Internal error


BryantLam:
### Summary of Problem
Modules A through J generate some kind of internal error from the compiler. These codes were generated from fuzzing a small set of sample programs.

```
Module  Error     Line
A       NUM0226   for param i in 0.."" {
B       TUP1181   var d1: [0..#LEN]*int(64);
C       EXP0262   proc test(x ..."") { writeln(x); }
D       FOR0392   forall "" in 0..#10 {
E       CAL0124   writeln(t2#(x));
F       NUM0173   var tmp: !3*int;
G       AGG0411   var (a,b,(c,d)) = (1,2,(3,4));
H       PAR0434   for param i in .0..5 {
I       TYP1191   for param i in ''..5 {
J       CAL0078   throw new Error(?);
```

### Steps to Reproduce

**Source Code:**

```chapel
module A {
proc main() {
  for param i in 0.."" {
    writeln(i);
  }
}
}

module B {
config param LEN = 1000;
proc main() {
  var d1: [0..#LEN]*int(64);  
}
}

module C {
proc test(x ..."") { writeln(x); }
proc main() {
  test(1, 2, 3);
}
}

module D {
proc main() {
  forall "" in 0..#10 {
    writeln(123);
  }
}
}

module E {
proc t2(param x) param { return x + 12; }
proc main() {
  var x = 12;
  writeln(t2#(x));
}
}

module F {
proc main() {
  var tmp: !3*int;
  writeln(tmp);
}
}

module G {
  record R {
    var (a,b,(c,d)) = (1,2,(3,4));
  }
}

module H {
proc main() {
  for param i in .0..5 {
    writeln(i);
  }
}
}

module I {
proc main() {
  for param i in ''..5 {
    writeln(i);
  }
}
}

module J {
proc main() throws {
  throw new Error(?);
}
}
```

### Configuration Information
`chpl version 1.18.0 pre-release (b0251ba45e)`

LouisJenkinsCS:
Could I ask how you generated these tests? I'm unfamiliar with fuzzing, but is this a kind of program where you provide the language specification as the input and have it generate random output? Is there a specific program you used?

BryantLam:
[...]

PR #10024, opened 2018-06-26
Add test showing that detuple-style decls in records fail

bradcray:
This is a future test case split off from issue #9987 exhibiting a program
from that issue which seems like it should work yet doesn't.


BryantLam:
I'd be interested in some design opinions on whether R2 should work.
```
record R2 {
  var (a,b,(c,d)) = (1,2,(3,4));
}
```
One valid use case I can think of is to destructure a tuple or object returned from a complex initialization or factory function. For a generic case, record R2's fields' types could be anything and this is one succinct way to write the declaration.

bradcray:
Yeah, there's no reason it shouldn't work.  I wouldn't have filed the future with that expected output if I didn't think it was a valid program.  :)

Issue #9985, opened 2018-06-22
Cannot use outer-scope variable, then shadow-redefine it


BryantLam:
### Summary of Problem
The following code has several test cases in it. The case that breaks is the one that reads the outer-scoped variable first, and then shadows that variable with a scope-local definition.
`11: error: 'a' used before defined (first used here)`

I question whether this behavior should be allowed, but since it is already in the language, it should work consistently. (And even if I didn't want *this specific test case* to work, the error message would not accurately explain what was wrong.)

### Steps to Reproduce

**Source Code:**

```chapel
proc main() {
  var a = 12;
  {
    var b = a;
  }
  {
    var a = 1;
    var b = a;
  }
  {
    var b = a; // <-- error reported here.
    var a = 1;
  }
  {
    var b = a;
    {
      var a = 1;
    }
  }
}
```

### Configuration Information
`chpl version 1.18.0 pre-release (b0251ba45e)`

lydia-duncan:
I believe this is not behaving inconsistently, but agree that the error message could be improved, probably to include the location of the "a" is that being used before defined.

Basically, the scope of a variable is defined to be the block where it is defined and any subscopes.  A symbol defined in the scope of another is always closer to that other symbol than a symbol defined in the outer scope.  We always look in the current scope first, and if we don't find a match there *then* we start looking into higher scopes (and scopes included via use statements).  This allows us to reference functions defined at any part within the scope instead of requiring that the function symbol is defined prior to the call to it.

I'm looking to find a bit of the spec that describes this, but am not finding it quickly.  I'll make another comment when I find it

bradcray:
Hi Bryant -- Lydia's right about this, though in saying that, I acknowledge that we depart from other languages in this respect.  By design, Chapel set out to be a language in which declaration order doesn't matter (e.g., you can call a procedure before it's defined and there's no need for prototypes; module declaration order shouldn't affect semantics or require any kind of "forward"/prototype declaration of the module).  Early on in the project, we decided to be uniform about this principle:  that naming a symbol declared within a given scope would refer to that symbol regardless of whether its declaration came before or after the reference.  An even simpler version of a program that demonstrates this issue is:

```chapel
var a = 1;   // call this a1
{
  var a = a;  // call this new a 'a2'; the initializer here also refers to a2, making this illegal
}
```

If this is something you'd want to advocate for changing (as compared to producing a better error message as Lydia suggests), please file a feature request for it with a rationale for why you think it's an important feature.

[...]

PR #10051, opened 2018-06-27
Have use-before-def error message point to where symbol was defined

### LINKED ISSUE #10034: Use before def errors should also note where the used variable is defined
### Summary of Problem
As seen in #9985, if an outer scoped variable shares a name with a more local scoped variable defined after the use, it can be confusing to tell which symbol is meant in the error message.  If the more local scoped variable is defined much later in the scope than its use-before-def, it would be hard to track it down to rename it or otherwise avoid the error.  It would be nice for this error message to also include the location where the matched variable is defined.

### Steps to Reproduce

**Source Code:**

```chapel
proc main() {
  var a = 12;
  {
    var b = a; // This actually refers to the variable on the following line, but the user thinks they are referring to the one above
    var a = 1;
  }
}
```

**Compile command:**
`chpl foo.chpl`

[`test/visibility/scoping-shadows-error-message.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/visibility/scoping-shadows-error-message.chpl) #1234

### Configuration Information

- Output of `chpl --version`: 1.18.0 (pre-release

bradcray:
In issue #10034 (and #9985 before that), it was noted that the error message for cases that misunderstand Chapel's use-before-def rules in the case of shadowing an outer variable like

```chapel
var a;
{
  var b = a;
  var a = 1;
}
```

were confusing.  This change tries to improve this by having the error message point to the point of definition to clarify that it's the local shadow variable, not the outer variable.

Note that there are other use-before-def errors generated by the compiler, particularly when initializing fields out-of-order that this commit does not improve.  I took a quick look at those cases, but wasn't seeing the way to get to the definition point easily before I timed out.

Resolves #10034.


compiler/passes/normalize.cpp
```diff
@@ -647,8 +647,8 @@ void checkUseBeforeDefs(FnSymbol* fn) {
                 // Only complain one time
                 if (undefined.find(sym) == undefined.end()) {
                   USR_FATAL_CONT(se,
-                                 "'%s' used before defined (first used here)",
-                                 sym->name);
+                                 "'%s' used before defined (defined here: %s:%d)",
+                                 sym->name, sym->defPoint->fname(), sym->defPoint->linenum());
```
vasslitvinov: how about simply sym->defPoint->stringLoc() ?

PR #10058, opened 2018-06-27
Mason: build on last modified

Spartee:
This PR includes the build on last modified feature for Mason. As our compilation time is still pretty slow, this feature allows for mason to decide if your project should be built based on the last modification time of all files within the `src/` directory and the binary found in `target/debug(release)`. Mason will obviously build every time when no binary exists. 

TODO - put `getLastModified` into standard library.

### DIFF
--- a/tools/mason/MasonBuild.chpl
+++ b/tools/mason/MasonBuild.chpl
@@ -123,7 +123,7 @@ proc buildProgram(release: bool, show: bool, force: bool, compopts: [?d] string,
       exit(1);
     }
     else {
-      writeln("Build stopped: No changes to project");
+      writeln("Skipping Build... No changes to project");
     }
   }
   catch e: MasonError {
--- a/tools/mason/MasonBuild.chpl
+++ b/tools/mason/MasonBuild.chpl
@@ -116,6 +116,7 @@ proc buildProgram(release: bool, show: bool, compopts: [?d] string,
         toParse.close();
       }
       else writeln("Cannot build: no Mason.lock found");
+      exit(1);
     }
     else {
       writeln("Build stopped: No changes to project");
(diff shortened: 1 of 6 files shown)


ben-albrecht:
We should have a flag to override this feature, in case the user is editing something that we are not checking like a standard library. Maybe `-f, --force`?

tools/mason/MasonUtils.chpl
```diff
@@ -23,6 +23,7 @@
 use Spawn;
 use FileSystem;
 use TOML;
+use DateTime;
```
ben-albrecht: Are we actually using the `DateTime` module or was that just for debugging?

Spartee: This was for debugging. I will remove it.

tools/mason/MasonBuild.chpl
```diff
@@ -64,54 +68,63 @@ private proc checkChplVersion(lockFile : Toml) {
...
-        writeln("Build Failed");
-        exit(1);
-      }
-      // Close memory
-      toParse.close();
+      else writeln("Cannot build: no Mason.lock found");
+      exit(1);
+    }
+    else {
+      writeln("Build stopped: No changes to project");
```
ben-albrecht: The text "Build stopped: ..." comes off as an error message, like "progress has been halted". 

It may be less alarming to users to change this to something like "Build skipped: ..." or "Skipping build: ...".

Spartee:
fixed latest comment in latest mason test PR [here](https://github.com/Spartee/chapel/blob/4d5c7ce79292bb7aa8e230263f52a4ab83ca8370/tools/mason/MasonUtils.chpl#L48)

PR #10061, opened 2018-06-27
Update --memStats to work properly in multi-locale runs.

gbtitus:
Nightly testing failures following PR 10014 demonstrated that --memStats
is broken for multi-locale configurations.  It prints redundant info
(all nodes' statistics are printed by every node) and would not have
worked with ASLR because it assumed that the memory statistics variable
addresses on remote nodes were the same as on the local node.  Fix all
this.

We can't use the comm layer to communicate memory statistics values from
one node to another, because we've already shut it down at the time we
print the memStats.  (This was what caused the failures with comm=gasnet
in nightly testing.  It was only working with comm=ugni because the comm
layer shutdown there is a bit half-hearted, terminating the polling
thread but leaving memory registered, and the memStats were stored in
registered memory so we didn't need AMs to communicate them.)  So, just
have each node print its own memStats.  Reduce the likelihood of output
corruption by constructing the output in memory and then printing it all
in one call.  The output from different nodes can nevertheless appear in
any order, however, so make it easier to sort by getting rid of the
headers and prefixing the lines with standardized strings.

Finally, make test changes to adapt to the new memStats output layout.


ronawho:
I think we should update the user-facing documentation here: 
https://github.com/chapel-lang/chapel/blob/43d0a414d3bd8b101eebe908ff60098c8f924df3/modules/standard/Memory.chpl#L55-L60

but looks good to me beyond that. I would say something to the effect of "multi-locale execution isn't perfect, but you can do the `grep memStats | sort` trick to get some reasonable output."

Maybe also open up an issue for improving this in the future? (though I certainly don't think it's high priority)

PR #10065, opened 2018-06-27
Allow a pragma to apply to an argument

mppf:
This PR adjusts the parser to allow a pragma to apply to a formal argument.
This will be useful for some lifetime-checking prototype improvements.

- [x] full local testing


lydia-duncan:
You could write a test using `pragma "no doc"` to lock this behavior in.  It would help me understand what you've done to verify the attachment works - I would be okay with a written list of what you checked as well, but I would prefer a test

lydia-duncan:
Another pragma works as well, if a good one comes to mind.

mppf:
@lydia-duncan - see commit 49d61c0 which adds a new test `test/parsing/ferguson/no-codegen-formal.chpl` that exercises the feature. Using "no doc" didn't work because chpldoc doesn't know to check it for arguments.

lydia-duncan:
Okay, I think that works for me.  This seems ready to go, thanks!

PR #10068, opened 2018-06-27
Generate h files in --library mode

### LINKED ISSUE #9669: Start generating .h files for compiled libraries

lydia-duncan:
Resolves #9669

When compiling in --library mode, generate a .h file for the library created,
so that the user doesn't have to write one by hand.  This header file contains
an include of the runtime's header file, as well as declarations for every
exported function in the program (including module initialization functions).
It's name by default is the same as the file containing the main module, but
a different name can be specified using the new flag `--library-header`.

Because every exported function is given a declaration, this file is significantly
longer than what the user would have written themselves.  We have quite a
few exported functions in module code.  It might be nice to have a separate
section for those in user code, so that these declarations are easier to find.
I'm viewing that as a later TODO item, though

For this generated file to be usable, I needed to use `type *` instead of
`c_ptr_type` or `_ref_type` (since C doesn't know what that means without
us including the typedefs and that seemed unnecessary), so this change
also modifies the type output for arguments and return types to do that.

While I was here, I also modified the default name of the generated library
file to start with 'lib' (only when compiling with --library).  To test the interaction
between setting the library name and setting the header file name, I added
support to sub_test to avoid passing `-o` as an argument when the
environment variable `CHPL_TEST_NO_USE_O` is set to true.

With this change, I removed the .h files on all library tests, added the .h's to
the .gitignore file for their directories, and had the tests clean them up as well.

- Testing:
  - [x] fresh checkout testing of the exported arrays directory
  - [x] fifo testing of the exported arrays directory
  - [x] valgrind testing of the exported arrays directory
  - [x] Full std/ testing with futures


compiler/codegen/codegen.cpp
```diff
@@ -1077,6 +1077,47 @@ static void codegen_aggregate_def(AggregateType* ct) {
   ct->symbol->codegenDef();
 }
 
+//
+// Generates a .h file to compliment the library file created using --library
```
cassella: *complement

compiler/codegen/symbol.cpp
```diff
@@ -786,6 +786,22 @@ static Type* getArgSymbolCodegenType(ArgSymbol* arg) {
   return useType;
 }
 
+// Alter the C code type to avoid _ref_, c_ptr_ prefixes and instead use a
+// type that C understands without additional information (e.g. instead of
+// _ref_int64_t, just int64_t *).  But only for exported symbols like the
+// the return type of exported functions, or arguments of those functions.
```
cassella: s/the the/the/

compiler/codegen/codegen.cpp
```diff
@@ -2147,6 +2192,17 @@ void codegen(void) {
                 executableFilename);
     }
     *lastDot = '\0';
+
+  }
+
+  if (fLibraryCompile && libmodeHeadername[0] == '\0') {
+    if (strncmp(executableFilename, "lib", 3) == 0) {
+      strncpy(libmodeHeadername, executableFilename + 3,
+              sizeof(executableFilename) - 3);
```
cassella: What will this do if the user's compiling `lib.chpl`?

lydia-duncan: If they're relying on the default -o settings, probably generate a file named just `.h`.  But in that case you'd get a library named `lib.a` and I think you can't specify that via -l?

lydia-duncan: At least, I don't seem to be able to do so

lydia-duncan:
@mppf  - would you mind reviewing this?

compiler/codegen/codegen.cpp
```diff
@@ -1077,6 +1077,47 @@ static void codegen_aggregate_def(AggregateType* ct) {
...
+      FILE* save_cfile = gGenInfo->cfile;
+
+      gGenInfo->cfile = libhdrfile.fptr;
+
+      //genComment("Generated header file for use with %s",
+      //           executableFilename);
+
+      fprintf(libhdrfile.fptr, "#include \"stdchpl.h\"\n");
+
+      // Maybe need LLVM "extern C header file" stuff?
```
mppf: You would need that if you wanted an extern block to include code callable in the library.

compiler/codegen/codegen.cpp
mppf: I'm really confused about the purpose of this code. I'd think it'd be better to not try to strip off `lib` from `libmymodule.chpl`.

compiler/codegen/symbol.cpp
```diff
@@ -786,6 +786,24 @@ static Type* getArgSymbolCodegenType(ArgSymbol* arg) {
   return useType;
 }
 
+// Alter the C code type to avoid _ref_, c_ptr_ prefixes and instead use a
+// type that C understands without additional information (e.g. instead of
+// _ref_int64_t, just int64_t *).  But only for exported symbols like the
+// return type of exported functions, or arguments of those functions.
+static std::string
+transformTypeForPointer(Type* type, bool exported) {
+  std::string typeName = type->codegen().c;
```
mppf: Shouldn't this be checking for c_ptr with FLAG_DATA_CLASS and for ref with FLAG_REF on type->symbol? The string begins-with stuff is pretty gross.

compiler/codegen/symbol.cpp
```diff
@@ -794,7 +812,9 @@ GenRet ArgSymbol::codegenType() {
   Type* useType = getArgSymbolCodegenType(this);
 
   if( outfile ) {
-    ret.c = useType->codegen().c;
+    std::string argType = transformTypeForPointer(useType,
+                                                  this->defPoint->parentSymbol->hasFlag(FLAG_EXPORT));
```
mppf: What happens if you just pass 'true' here. Does testing pass? I don't think this behavior should depend on FLAG_EXPORT.

compiler/codegen/symbol.cpp
```diff
@@ -1254,7 +1274,10 @@ GenRet FnSymbol::codegenFunctionType(bool forHeader) {
   if( info->cfile ) {
     // Cast to right function type.
     std::string str;
-    str += retType->codegen().c.c_str();
+
+    std::string retString = transformTypeForPointer(retType,
+                                                    hasFlag(FLAG_EXPORT));
```
mppf: I don't think this should depend on FLAG_EXPORT

compiler/codegen/codegen.cpp
lydia-duncan: Cython expects the header file name and the library file name to have a relationship of "moduleName.h" to "libmoduleName.a/.so".  Our generated library file name does not include the lib prefix by default, so generally what I've had to do is explicitly specify it using -o libmoduleName.  I chose to base the name for the header file off the executableFilename so I didn't have to separately compute the name and risk code drift should we decide to change the generated name for the .a, but that means that I have to strip the lib portion back off for the header name.

I could change our generated name in library mode includes the lib prefix (since that seems like what you would expect in that case anyways) and avoid updating the header name when that happens.  Do you think that would be appropriate to do in this PR?

compiler/codegen/symbol.cpp
lydia-duncan: I didn't realize I could use FLAG_DATA_CLASS to detect c_ptrs, thanks!  Is FLAG_REF usable for both the explicit ref types and the qualified ref stuff?

compiler/codegen/symbol.cpp
lydia-duncan: I haven't tried that (wanted to make sure it worked in the export case before expanding), but I will now.

compiler/codegen/codegen.cpp
mppf: > I could change our generated name in library mode includes the lib prefix (since that seems like what you would expect in that case anyways) and avoid updating the header name when that happens. Do you think that would be appropriate to do in this PR?

I think you're talking about the default generated name here, right? Or all the time? Since header mode makes another file, maybe we want to be able to specify the name for it separately as its own flag?

I don't know how much testing impact that does - if it changes a lot of tests separate PR is better.

compiler/codegen/symbol.cpp
mppf: For qualified refs the type here probably ends up as a _ref type (that's what getArgSymbolCodegenType does).

compiler/codegen/codegen.cpp
lydia-duncan: > I think you're talking about the default generated name here, right? Or all the time?

Yeah, the default generated name which can get overridden by `-o`.  I would want to only add a "lib" prefix to it in `--library` compilation.

> Since header mode makes another file, maybe we want to be able to specify the name for it separately as its own flag?

This is certainly not a good reason to avoid it, but won't that mean adding special casing to start_test/sub_test for library compiles so that we don't override the default?  We still explicitly pass `-o` there.

Changing the behavior of `-o` for library mode versus normal compilation doesn't give me pause, since you would only be able to use at most one of the two flags in any given compile (there's no world where you would want to be able to set both in a single compilation, they impact roughly the same thing, I don't particularly see the point of having a separate flag for it).  I would be curious if others felt that was the right choice.

> I don't know how much testing impact that does - if it changes a lot of tests separate PR is better.

I believe every test that it would impact was already impacted by this change (due to preventing git from tracking the header files I am generating, and cleaning up the generated files)

compiler/codegen/symbol.cpp
lydia-duncan: Full local paratest passed.  I'm going to remove the argument from the function

compiler/codegen/symbol.cpp
lydia-duncan: I actually want FLAG_C_PTR_CLASS, FLAG_DATA_CLASS also grabs _ddata and I *don't* want that

lydia-duncan:
Okay, @mppf, I think this is ready for another look

compiler/codegen/codegen.cpp
mppf: Might be worth updating this comment to say "extern block"

compiler/codegen/codegen.cpp
```diff
@@ -2132,21 +2177,58 @@ void codegen(void) {
...
                 executableFilename);
     }
     *lastDot = '\0';
+
+  }
+
+  // If we're in library mode and the executable name was set but the header
+  // name wasn't, use the executable name for the header name as well
+  if (fLibraryCompile && libmodeHeadername[0] == '\0') {
+    strncpy(libmodeHeadername, executableFilename, sizeof(executableFilename));
```
mppf: Can all this computing-library-file-names stuff go into a new function? Just because codegen() is really big doesn't mean we should make it worse ;)

compiler/codegen/symbol.cpp
```diff
@@ -786,6 +786,24 @@ static Type* getArgSymbolCodegenType(ArgSymbol* arg) {
   return useType;
 }
 
+// Alter the C code type to avoid _ref_, c_ptr_ prefixes and instead use a
+// type that C understands without additional information (e.g. instead of
+// _ref_int64_t, just int64_t *).  But only for exported symbols like the
+// return type of exported functions, or arguments of those functions.
```
mppf: I believe this can also apply to DDATA, as long as it isn't wide. Might be worth a TODO comment.

compiler/codegen/symbol.cpp
```diff
@@ -786,6 +786,24 @@ static Type* getArgSymbolCodegenType(ArgSymbol* arg) {
...
+static std::string
+transformTypeForPointer(Type* type) {
+  std::string typeName = type->codegen().c;
+  if (type->symbol->hasFlag(FLAG_REF)) {
+    Type* referenced = type->getValType();
+    return referenced->codegen().c + " *";
+
+  } else if (type->symbol->hasFlag(FLAG_C_PTR_CLASS)) {
+    // Remove the c_ptr_ prefix
+    return typeName.substr(6, std::string::npos) + " *";
```
mppf: If you look at the implementation for getValType(), for a _ref, it's just getting the type of a field with a particular name. I think you should do the same thing here. Something like `toAggregateType(type)->getField("eltType")->type`.

compiler/util/files.cpp
```diff
@@ -690,6 +693,12 @@ void codegen_makefile(fileinfo* mainfile, const char** tmpbinname, bool skip_com
   // factor of 5 or so in time in running the test system, as opposed
   // to specifying BINNAME on the C compiler command line.
 
+  if (fLibraryCompile) {
+    fprintf(makefile.fptr, "HEADERNAME = %s%s\n\n", libmodeHeadername, ".h");
+    tmpheader = astr(tmpDirName, "/", strippedHeadername, ".h");
+    fprintf(makefile.fptr, "TMPHEADERNAME = %s\n", tmpheader);
```
mppf: I'm not sure what the point of giving the header name to the Makefile is anyway.
Or putting it in the /tmp directory and then moving it.
Does this header need to be available when compiling the libsomething.a ?
If the answer is "yes", then this is OK. If the answer is "no", then maybe it should be simplified?

compiler/codegen/codegen.cpp
```diff
@@ -2132,21 +2177,58 @@ void codegen(void) {
       lastSlash++;
     }
 
-    // copy from that slash onwards into the executableFilename,
-    // saving space for a `\0` terminator
-    if (strlen(lastSlash) >= sizeof(executableFilename)) {
-      INT_FATAL("input filename exceeds executable filename buffer size");
+    // "Executable" name should be given a "lib" prefix in library compilation,
+    // and just the main module name in normal compilation.
```
mppf: I was expecting that if somebody ran `chpl --library -o blarg.a blarg.chpl`, then we'd create the file `blarg.a` and not `libblarg.a`. I was expecting `chpl --library blarg.chpl` to make `libblarg.a`, though. How would this code differentiate between these two cases?

compiler/util/files.cpp
lydia-duncan: I talked to Brad about this before doing it: there was a comment that it saved execution time to put everything in the temp directory first then move it out, so I'm following that recommendation and using the same treatment for the header file.

compiler/codegen/codegen.cpp
lydia-duncan: This code only triggers if the executableName hasn't already been set by the user - see [this line](https://github.com/chapel-lang/chapel/pull/10068/files/3ea24c69fa1a595fcd0d209d5780b4738e6a8abf#diff-bb67493ab4303d061942b7df2da31229R2168).  We set executableFilename from the compiler flag directly

lydia-duncan:
TODO: figure out how to regenerate the autocomplete script to avoid the test failure

compiler/codegen/symbol.cpp
lydia-duncan: The eltType field no longer exists on the type at codegen because it was a type field and those get removed at resolution.  It looks like it is just included included in the type name of the instantiation

mppf: Try getDataClassType (I got that by asking "OK, if that's the case, how do we generate the typedefs for a ref?)

lydia-duncan: Refs still have a value field, which isn't just a type.  I'll check the function you recommend, though

lydia-duncan: That worked, good call!

lydia-duncan:
(Thanks @daviditen for the tip on autocompletion)

lydia-duncan:
@mppf - final stamp of approval?

lydia-duncan:
I do not understand 🦆 as a response XD

mppf:
it was a stamp

bradcray:
I just read this:

> We have quite a
> few exported functions in module code. It might be nice to have a separate
> section for those in user code, so that these declarations are easier to find.
> I'm viewing that as a later TODO item, though

Seems like rather than sorting them into separate sections, we should just squash the exported declarations that aren't in user code (since the runtime code that makes use of those calls has already hard-coded their names into its sources, for better or worse).  Although I haven't looked at the code, it seems like this should be a five-minute fix (and an attractive one) in which we'd just check the type of the enclosing module when we're codegen-ing it into the header file (?), squashing those that aren't MOD_USER.

lydia-duncan:
I think that's reasonable.  I took a brief look through the additional functions and none of them struck me as something the user would want to use themselves.

lydia-duncan:
Thinking about this more, we're going to only eliminate the ones in the internal modules, keeping the standard and package exported functions in the output in case a user wants access to our standard or package functions from C or Python

PR #10071, opened 2018-06-27
Update memstatPrint .good files to match the new memStats formatting

ronawho:
PR #10061 changed the formatting of memStats, but we missed changing
memstatPrint as part of that work since this test uses `printMemAllocStats()`
directly instead of using the `--memStats` flag. This just updates the .good
file to match the new formatting.

### DIFF
--- a/test/memory/shannon/memstatPrint.comm-gasnet.good
+++ b/test/memory/shannon/memstatPrint.comm-gasnet.good
@@ -1,40 +1,20 @@
-=================
-Memory Statistics
-==============================================================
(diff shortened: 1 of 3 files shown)


ronawho:
@gbtitus I'm going to merge this to avoid noise in nightly testing

ronawho:
And belatedly I'm realizing that the .good file for all for all of these configs are identical, so we can probably drop the comm-gasnet and no-local specific ones now.

gbtitus:
Apologies for missing this and thanks for fixing it!

Issue #10000, opened 2018-06-25
Misc qthreads shim improvements


ronawho:
There's a bunch of improvements/cleanups to the qthreads shim that'd be nice to get around to

 - [x] Cleanup stack limiting code -- #5749 (done as part of #10254)
 - [x] Fix 0x5F yield frequency in `chpl_sync_lock()` (removed in #10133)
 - [x] Cleanup init/exit code and add comments (done in #10223)
 - [x] Resolve sporadic shutdown hang -- #10074 (done in #10078)
 - [x] Join main pthread to fix leak -- #9819 (done in #10086)
 - [x] Resolve parallel I/O performance issue -- #7748 (done in #10133)
 - [x] Switch `syncvar_t` over to `aligned_t`(done in #10078, #10082)
    - `syncvar_t` still suffers from sync var serialization -- #9651
    - `aligned_t` is what our module code calls for native sync vars, so it's more tested

bradcray:
You won the golden ticket!!!

PR #10073, opened 2018-06-28
Remove qthreads "support" for --blockreport and --taskreport

ronawho:
Qthreads had very experimental support for --blockreport that no longer works
correctly and no support for --taskreport. Instead of trying to get either of
these flags working, just remove the existing code since we don't believe these
features are practically useful anyways.

### DIFF
--- a/runtime/src/tasks/qthreads/tasks-qthreads.c
+++ b/runtime/src/tasks/qthreads/tasks-qthreads.c
@@ -66,8 +66,6 @@
 #include <math.h>
 
 
-//#define SUPPORT_BLOCKREPORT
-//#define SUPPORT_TASKREPORT
 
 #ifdef CHAPEL_PROFILE
 # define PROFILE_INCR(counter,count) do { (void)qthread_incr(&counter,count); } while (0)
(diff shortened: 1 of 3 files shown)


runtime/src/tasks/qthreads/tasks-qthreads.c
```diff
@@ -315,66 +297,18 @@ void chpl_sync_destroyAux(chpl_sync_aux_t *s)
...
-
 static void SIGINT_handler(int sig)
 {
     signal(sig, SIG_IGN);
 
     if (blockreport) {
-#ifdef SUPPORT_BLOCKREPORT
-        report_locked_threads();
-#else
         fprintf(stderr,
```
gbtitus: The `blockreport` and `taskreport` flags will already have been set at the time `chpl_task_init()` is called.  Would it be better to print a message there?  (Especially for `blockreport`, which the user might have set thinking it would enable deadlock detection, as it does for tasks=fifo.)

ronawho: I thought about that too, but I decided to leave it as-is because I'm not actually changing the behavior compared to master (i.e I'll I'm really doing is removing code blocks were ifdeffed out unless somebody went out of their way to modify the source)

ronawho: (And semi-related, I plan to make an issue to see how people feel about removing support for these features completely since I don't think they're very useful and they're pretty broken even under fifo -- https://github.com/chapel-lang/chapel/pull/7255, https://github.com/chapel-lang/chapel/pull/7601)

gbtitus: I'm fine with that, no need for further changes on this one.  And I'll definitely have input to your new issue!

PR #10081, opened 2018-06-28
Add a function to read multiple HDF5 files into a single 1-D array

daviditen:
Add a function `readNamedHDF5FilesInto1DArrayInt` to read all HDF5 files
listed in an argument into a single 1-D array.  This can be used to split
a large file into multiple smaller files to be read in parallel. It assumes
the data files are all the same size, with 2-D arrays of int(64)s.

Split `readAllHDF5Files()` into `readAllHDF5Files()` and
`readAllNamedHDF5Files()` to allow reading a list of files instead of always
reading all files in a directory.

Add a test that reads a 6000x5000 array (into 1-D) from 8 files.


modules/packages/HDF5_HL.chpl
```diff
@@ -3588,6 +3596,40 @@ module HDF5_HL {
...
+       elements are all int(64)s.
+       `fnCols` and `fnRows` refer to the columns and rows that `filenames`
+       should have when converting it back to 2-D.
+     */
+    proc readNamedHDF5FilesInto1DArrayInt(filenames: [] string,
+                                          fnCols: int, fnRows: int,
+                                          dsetName: string) {
+      use BlockDist;
+
+     var filenames2D = reshape(filenames, {1..fnCols, 1..fnRows});
```
lydia-duncan: Weird indentation here

lydia-duncan:
Also, sorry for not getting to the review earlier...

PR #10087, opened 2018-06-28
Mason Test Improvements

Spartee:
This PR adds three new features to the mason test command:
- Parallel compilation and the option via flag to run tests in parallel with `--parallel` flag.
- Test results are abstracted from user view via status codes with the option to view stdout of tests with`--show` flag.
- Tests are found by mason recursively within the `test/` directory with the option for the user to specify them manually in their `Mason.toml`.

### DIFF
--- a/test/mason/mason-test/Mason.toml
+++ b/test/mason/mason-test/Mason.toml
@@ -3,7 +3,12 @@
 name = "sampleModule"
 version = "0.1.0"
 chplVersion = "1.17.0"
-tests = ["sampletest.chpl", "sampletest2.chpl", "dependencytest.chpl", "nomodule.chpl", "testdir/subdirectorytest.chpl"]
+tests = ["sampletest.chpl",
+         "sampletest2.chpl",
+         "dependencytest.chpl",
+         "nomodule.chpl",
+         "testdir/subdirectorytest.chpl",
+         "shallnotpass.chpl"]
 
 [dependencies]
 _MasonTest1 = "0.1.0"
--- a/tools/mason/MasonTest.chpl
+++ b/tools/mason/MasonTest.chpl
@@ -92,9 +92,8 @@ private proc runTests(show: bool, run: bool, parallel: bool) {
 
       var resultDomain: domain(string);
       var testResults: [resultDomain] string;
-      var numPassed = 0;
 
-      forall test in testNames with (+ reduce numPassed) {
+      forall test in testNames {
 
         const testPath = "".join(projectHome, '/test/', test);
         const testName = basename(stripExt(test, ".chpl"));
@@ -118,18 +117,19 @@ private proc runTests(show: bool, run: bool, parallel: bool) {
             }
             else {
               testResults[testName] = "Passed";
-              numPassed += 1;
             }
           }
         }
       }
       if run && !parallel {
         var testBinResults = runTestBinaries(projectHome, testNames, numTests, show);
-        resultDomain = testBinResults[1].domain;
-        testResults = testBinResults[1];
-        numPassed = testBinResults[2];
+        resultDomain = testBinResults.domain;
+        testResults = testBinResults;
       }
-      if run then printTestResults(testResults, numTests, numPassed, show);
+      if run {
+        const numPassed = testResults.count("Passed");
+        printTestResults(testResults, numTests, numPassed, show);
+      }
     }
     else {
       writeln("No tests were found in /test");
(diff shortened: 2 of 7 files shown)


tools/mason/MasonTest.chpl
```diff
@@ -64,25 +69,29 @@ private proc runTests(show: bool, run: bool) {
 
     // parse lockfile
     const toParse = open(projectHome + "/Mason.lock", iomode.r);
-    const lockFile = new Owned(parseToml(toParse));
+    const lockFile = parseToml(toParse);
```
ben-albrecht: Why drop the delete-free `Owned` here?

tools/mason/MasonTest.chpl
```diff
@@ -64,25 +69,29 @@ private proc runTests(show: bool, run: bool) {
...
     const projectPath = "".join(projectHome, "/src/", project, ".chpl");
 
+    if isDir(joinPath(projectHome, "target/test/")) {
+      rmTree(joinPath(projectHome, "target/test/"));
+    }
     // Make target files if they dont exist from a build
     makeTargetFiles("debug", projectHome);
 
+    // get the test names from lockfile or from test directory
+    const testNames = getTests(lockFile, projectHome);
```
ben-albrecht: > Why drop the delete-free Owned here?

Probably because this was causing issues. You can add the `Owned` back in above, and update `lockFile` to `lockFile.borrow()` here to avoid transferring ownership over to `getTests`

tools/mason/MasonTest.chpl
```diff
@@ -99,12 +108,11 @@ private proc runTests(show: bool, run: bool) {
         }
```
ben-albrecht: Just noticed this on line `105` above: I think it'd make sense for the ` if show then writeln(compCommand);` to occur before the compilation, such that the user sees the command before any potential error (and sees the commands in real-time).

test/mason/mason-test/Mason.toml
```diff
@@ -3,7 +3,12 @@
 name = "sampleModule"
 version = "0.1.0"
 chplVersion = "1.17.0"
-tests = ["sampletest.chpl", "sampletest2.chpl", "dependencytest.chpl", "nomodule.chpl", "testdir/subdirectorytest.chpl"]
+tests = ["sampletest.chpl",
+         "sampletest2.chpl",
+         "dependencytest.chpl",
+         "nomodule.chpl",
+         "testdir/subdirectorytest.chpl",
+         "shouldnotpass.chpl"]
```
ben-albrecht: Missed a great opportunity for `shallnotpass.chpl`!

tools/mason/MasonTest.chpl
```diff
@@ -90,28 +90,45 @@ private proc runTests(show: bool, run: bool, parallel: bool) {
...
+
         if compilation != 0 {
           writeln("compilation failed for " + test);
         }
         else {
           if show || !run then writeln("compiled ", test, " successfully");
+          if parallel {
+            var result = runTestBinary(projectHome, testName, show);
+            if result != 0 {
+              testResults[testName] = "Failed";
```
ben-albrecht: FWIW, this is racy, such that if another iteration were to write to the same `testName` element, it would overwrite the previous value. In practice, this should never happen. However, if you wanted to write this safer, you could only modify the domain within this loop, such that the inclusion of a test name in the domain means the test failed, and all the other tests passed.

tools/mason/MasonTest.chpl
```diff
@@ -90,28 +90,45 @@ private proc runTests(show: bool, run: bool, parallel: bool) {
...
+          }
         }
       }
-      if run {
-        runTestBinaries(projectHome, testNames, numTests, show, parallel);
+      if run && !parallel {
+        runTestBinaries(projectHome, testNames, numTests, show);
+      }
+      else if run && parallel {
+        printTestResults(testResults, numTests, numPassed, show);
```
ben-albrecht: Why not pull the `printTestResults` out of `runTestBinaries` so that this is just:

```chpl
else if run then printTestResult(..);
```

tools/mason/MasonTest.chpl
```diff
@@ -116,7 +141,52 @@ private proc runTests(show: bool, run: bool) {
...
+    else {
+      testResults[testName] = "Passed";
+      numPassed +=1;
+    }
+  }
+  return (testResults, numPassed);
+}
+
+
+private proc printTestResults(testResults: [?d] string, numTests: int,
```
ben-albrecht: Idea: Remove `numPassed` everywhere, and just define it within this function:

```chpl
const numPassed = testResult.count('Passed');
```

PR #10090, opened 2018-06-29
Remove unused benchmarks/lulesh/lulesh-sedov15cube.good

cassella:
Remove unused file
test/release/examples/benchmarks/lulesh/lulesh-sedov15cube.good


bradcray:
It looks like the file is still accurate, in that if I do a run:

```
./lulesh --filename=lmeshes/sedov15cube.lmesh
```

the output matches.  I can't recall offhand why I didn't add this test configuration in the suite, but am guessing that it's just because we were running so many configurations already (and this was one of the less interesting ones).

I assume the motivation for removing it is part of an effort to remove stray files that aren't used?

cassella:
Yeah, that's the motivation.

I opened this separately from the PR with the other files since it felt a little different than inadvertently added or inadvertently not removed.

ronawho:
I'm going to close this, but @bradcray feel free to reopen and merge if you think it's worth getting rid of this currently unused file

Issue #10089, opened 2018-06-29
Switch gasnet comm diagnostic to use atomic counters


ronawho:
Today, we use non-atomic counters protected by a `chpl_sync_lock()`/`chpl_sync_unlock()` for comm diagnostics in our gasnet shim. This seems extremely heavyweight/unnecessary, so convert them to atomic counters like we do in ugni.

Currently `chpl_sync_lock()`/`chpl_sync_unlock()` under qthreads are implemented with an atomic spinlock, but I'm looking at transitioning that to `qthread_lock()`/`qthread_unlock()` (as part of #10000 and #7748), so I was looking for cases that use sync lock/unlock to make sure it made sense for them to use a heavier-weight locking scheme, and the gasnet comm diags case struck me as wrong.

PR #10091, opened 2018-06-29
Switch to atomic counters for gasnet's comm diagnostics

ronawho:
Previously, our gasnet shim used non-atomic variables with access protected by
`chpl_sync_lock()`/`chpl_sync_unlock()`. Protecting counters with a lock is
very heavyweight, so just switch to atomic counters, like we were already doing
in ugni.

This makes comm diagnostics much more lightweight under gasnet. This also
cleans up and unifies the comm diagnostics code for ugni and gasnet.

I noticed this while making changes to `chpl_sync_lock()`/`chpl_sync_unlock()`,
which motivated me to look at uses of these routines in the runtime.

This closes https://github.com/chapel-lang/chapel/issues/10089

### DIFF
--- a/runtime/include/chpl-comm-diags.h
+++ b/runtime/include/chpl-comm-diags.h
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2004-2018 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ *
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef _chpl_comm_diags_h_
+#define _chpl_comm_diags_h_
+
+#include "chpl-atomics.h"
+#include "chpl-comm.h"
+
+typedef struct _chpl_atomic_commDiagnostics {
+#define _COMM_DIAGS_DECL_ATOMIC(cdv) atomic_uint_least64_t cdv;
+  CHPL_COMM_DIAGS_VARS_ALL(_COMM_DIAGS_DECL_ATOMIC)
+#undef _COMM_DIAGS_DECL_ATOMIC
+} chpl_atomic_commDiagnostics;
+
+
+static inline
+void chpl_comm_diags_init(chpl_atomic_commDiagnostics* acd) {
+#define _COMM_DIAGS_INIT(cdv) atomic_init_uint_least64_t(&acd->cdv, 0);
+  CHPL_COMM_DIAGS_VARS_ALL(_COMM_DIAGS_INIT);
+#undef _COMM_DIAGS_INIT
+}
+
+static inline
+void chpl_comm_diags_reset(chpl_atomic_commDiagnostics* acd) {
+#define _COMM_DIAGS_RESET(cdv) atomic_store_uint_least64_t(&acd->cdv, 0);
+  CHPL_COMM_DIAGS_VARS_ALL(_COMM_DIAGS_RESET);
+#undef _COMM_DIAGS_RESET
+}
+
+static inline
+void chpl_comm_diags_copy(chpl_commDiagnostics* cd,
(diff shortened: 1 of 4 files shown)


ronawho:
I'm pretty happy with this implementation, but I don't love all the names. In particular: 
 - `chpl_comm_diags_amo_get()` should probably be "transfer" or "copy" instead of "get". 
 - I'm not sure if I like `_amo_` as part of the routine names since I typically think of amo as network atomics, which might give the false impression that these are counters for network atomics. I wanted some way to convey that it's for atomic counters, but wanted to try to keep the names short still.

runtime/include/chpl-comm.h
```diff
@@ -508,6 +512,53 @@ typedef struct _chpl_commDiagnostics {
   uint64_t execute_on_nb;
 } chpl_commDiagnostics;
 
+typedef struct _chpl_atomic_commDiagnostics {
```
gbtitus: I definitely like having the internal counters implemented just once and in a high-performance way, but I'm not fond of that being done in `chpl-comm.h` which is where one would expect to find the interface to the comm layer rather than its exposed internals.  Would you object to putting this in, say, a `chpl-comm-internal.h` file or similar, and `#include`'ing that in the various comm layers, in addition to `chpl-comm.h`?

runtime/src/comm/gasnet/comm-gasnet.c
```diff
@@ -931,14 +926,13 @@ void chpl_comm_post_task_init(void) {
   }
 
   // clear diags
-  memset(&chpl_comm_commDiagnostics, 0, sizeof(chpl_commDiagnostics));
+  chpl_resetCommDiagnosticsHere();
```
gbtitus: `chpl_comm_diags_reset()`?

gbtitus:
Have you considered adding counters for the network atomics calls?  It's an obvious extension.  I think we've avoided it mainly because of the pervasive impact on .good files etc. but that's not actually maybe more of an excuse than a reason, especially now that we don't print zero values.

runtime/include/chpl-comm.h
ronawho: Yeah, I didn't love it either, but the reason I put it here is to have the 2 structs and the copying/init code right next to each other. That way if we change the `chpl_commDiagnostics` "API" to the module code, we'll be more likely to remember to update the atomic API too.

ronawho:
> Have you considered adding counters for the network atomics calls? It's an obvious extension. I think we've avoided it mainly because of the pervasive impact on .good files etc. but that's not actually maybe more of an excuse than a reason, especially now that we don't print zero values.

I hadn't but that's a good idea. That feels separate from what this PR is trying to address, so I'll probably just make a new issue for that.

runtime/include/chpl-comm.h
gbtitus: We could code a generator macro to declare the struct type and members, with the struct tag and member type being parameters.  Then just invoke this one way to build the public type in `chpl-comm.h` and another way to build the private one in `chpl-comm-internal.h`.  (I wouldn't mind the generator being in the public .h file.)

Oh, wait, that won't work.  Codegen via LLVM wouldn't be able to deal with it, for the usual reason.  Sigh.

All right.  Would you mind at least moving the internal struct decl down to the bottom of the file, say, with a header comment saying something like "stuff below here isn't part of the comm interface"?  (Maybe combine this with the generator macro idea if you don't like separating the two structs and their should-correspond-with-each-other member lists.)

ronawho: I don't think the generator is an issue for llvm (otherwise our atomics/network atomics wouldn't work) 

I _think_ the problem is for function like macros

gbtitus: Hmm!  Now I'm confused.  If the LLVM issue was for function-like macros it seems like it would affect ugni's `CHPL_COMM_IMPL_REG_MEM_FREE()`.

runtime/include/chpl-comm-diags.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef _chpl_comm_diags_h_
+#define _chpl_comm_diags_h_
+
+#include "chpl-atomics.h"
+#include "chpl-comm.h"
+
+typedef struct _chpl_atomic_commDiagnostics {
+#define _DIAGS_DECL_ATOMIC(cdv) atomic_uint_least64_t cdv;
```
gbtitus: Oops, I missed this until after I'd approved.  Maybe `_COMM_DIAGS_...`?  (I'm imagining a future in which we do something similar with another layer, like tasking.)

ronawho: Ah, to prevent collisions if `_DIAGS_DECL_*` is already set when we're trying to define it  -- changed

Issue #10097, opened 2018-06-29
analyzeMemLeaksLog is broken on master


ronawho:
Recent changes to `--memStats` reporting in https://github.com/chapel-lang/chapel/pull/10061 has broken [util/devel/analyzeMemLeaksLog](https://github.com/chapel-lang/chapel/blob/master/util/devel/analyzeMemLeaksLog), which is used as part of nightly memtracking, so the graphs show 0 bytes allocated/freed/leaked: https://chapel-lang.org/perf/chapcs/?startdate=2018/05/08&enddate=2018/06/29&suite=memorytracking

For context for others, we changed the memStats formatting because it was incredibly broken for mult-locale. Previously it relied on trying to do gets after the comm layer is shutdown, and it was even more broken for systems with ASLR. Since we can't communicate, we can't coordinate how memory stats for different locales are printed, so each locale has to print it's own memory leaks, which means there's no order to the output. To deal with this we changed the `--memStats` formatting to drop a header/footer section, and append "memStats:" to the beginning of each line so that you can get reasonable output by doing something like `| grep memStats | sort`

For single locale our format for the memstats section used to look like:

```
=================
Memory Statistics
==============================================================
Current Allocated Memory               192
Maximum Simultaneous Allocated Memory  995
Total Allocated Memory                 1409
Total Freed Memory                     1217
==============================================================
```

And it now looks like:

```
memStats: Allocated Now:               192
memStats: Allocation High Water Mark:  995
memStats: Sum of Allocations:         1409
memStats: Sum of Frees:               1217
```

and for multi-locale it used to segfault on systems with ASLR, hang for gasnet-aries, and just happend to still work for ugni because enough of the comm layer was still active for gets to work.

It now looks like:

```
memStats: node 0 Allocated Now:               288
memStats: node 0 Allocation High Water Mark: 1163
memStats: node 0 Sum of Allocations:         1577
memStats: node 0 Sum of Frees:               1289
memStats: node 1 Allocated Now:              0
memStats: node 1 Allocation High Water Mark: 0
memStats: node 1 Sum of Allocations:         0
memStats: node 1 Sum of Frees:               0
```

(where arbitrary interleaving of the output between locale 0 and 1 is possible)

[...]

PR #10101, opened 2018-06-29
Adjust analyzeMemLeaksLog to handle the new memStats format

ronawho:
We adjusted `--memStats` output in #10061 and that broke analyzeMemLeaksLog,
which looks for sentinels and specific strings. In order to get `--memStats`
working with multi-locale we removed the header/footer that analyzeMemLeaksLog
looked for, and we changed some of the wording. This ajdusts analyzeMemLeaksLog
to stop looking for the header/footer and to expect the new wording.

This resolves https://github.com/chapel-lang/chapel/issues/10097

### DIFF
--- a/util/devel/analyzeMemLeaksLog
+++ b/util/devel/analyzeMemLeaksLog
@@ -24,26 +24,16 @@ while (<>) {
         $path = $1;
         $path =~ s/\-/\//g;
         $program = "$path/$program";
-    } elsif (m/Memory Statistics/) {
-        $memstats = 1;
     } elsif (m/Leaked Memory Report/) {
         $memleaks = 1;
-    } elsif ($memstats == 1) {
-        if (m/==============================================================/) {
-            $memstats = 2;
-        }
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus I'm going to merge this since it's causing issues in memleaks testing, but this probably deserves a look from you post-merge

gbtitus:
This looks good and apologies for my having broken things.  (I probably should have searched the repo for the output artifacts of memory statistics instead of just for the execopt and function that produce it.) 

Would it be worthwhile for me to put a little comment mentioning this script, in the runtime code that produces the report?  That might prevent this happening again.

ronawho:
> Would it be worthwhile for me to put a little comment mentioning this script, in the runtime code that produces the report? That might prevent this happening again.

I don't feel strongly about that. If you want to do anything, I would add a test instead of a comment though. Something that runs memleaks testing on a test with a leak and a prediff that runs `util/devel/analyzeMemLeaksLog`

PR #10102, opened 2018-06-29
Preserve loop-expressions through to normalize

benharsh:
This PR stops the immediate lowering of loop-expressions during parsing and
preserves the expression until normalization. The ForallExpr node is renamed to
LoopExpr and repurposed to support this effort. The primary goal is to support
the usage of loop-expressions as default values for fields in types with
initializers. A pleasant result is cleaner AST.

LoopExpr now resides in its own file, which stores its definition and
lowering-logic. The bulk of the code added to LoopExpr.cpp is nearly-identical
to the original lowering in build.cpp.

Now that LoopExprs survive scope resolution, the indices for the
LoopExpr have actually been resolved to SymExprs. Alas, other parts of
the compiler (e.g. destructureIndices, loop-builders) are easier to work
with when using UnresolvedSymExprs as they will create DefExprs in the
right spots. In order to leverage this, replace the indices with
UnresolvedSymExprs.

LoopExprs need to be handled differently depending on whether they are
inside an ArgSymbol. For LoopExprs at module scope or within a function,
they can be lowered in the same manner as they used to: create several
nested functions at the scope of the loop-expr. This doesn't work for
LoopExprs inside ArgSymbols because that pattern is generally not
supported by resolution. There are issues with visible-functions,
scoped-types disappearing, and other bizarre unhandled behavior.

Instead, we will generate a single function at module scope and pass any
required data to that function. The various iterator functions will be
nested within this top-level function in order to share visibility for
those outer variables.

Finally, each function will go through the same post-processing:
- add function to the symbol table
- lower short-circuit operators
- scope resolution
- normalization

The core normalize pass is updated to handle a LoopExpr as a variable's
initialization expression and to lower LoopExprs as part of arbitrary
AST normalization. Lowering of IfExprs and short-circuit operators is
tweaked to return early if the expression is inside a LoopExpr, in favor
of returning to those expressions once the LoopExpr is lowered.

Miscellaneous:
- A 'chpl__initCopy' call is inserted for a default initializer's formal's
  default value if the value is a loop-expr, so that the result is an array
- AST printers make an attempt to distinguish between 'forall' and shorthand
  '[]' loop exprs
- scope resolution module caches are now destroyed after every pass, avoiding
  an issue in which dead AST lives on in the cache
- this PR takes the opportunity to rename the lowered loop-expr functions to
  use a 'chpl__' prefix.

Testing:
- [x] local + futures
- [x] gasnet
- [x] memleaks
- [x] baseline over release/examples
- [x] resolves --force-initializers failures


benharsh:
Down to 8 --force-initializers failures after this PR!

compiler/AST/LoopExpr.cpp
```diff
@@ -0,0 +1,648 @@
...
+// Lower outer-most LoopExprs first, in order to simplify scope-resolution
+// of newly-created functions. If we lower innermost LoopExprs first, and a
+// top-level LoopExpr attempts to scope-resolve its functions later, scope
+// resolution could fail due to nested functions already added to the symbol
+// table.
+//
+bool LowerLoopExprVisitor::enterLoopExpr(LoopExpr* node) {
+  // Don't touch LoopExprs in DefExprs, they should be copied later into
+  // BlockStmts.
+  if (isAlive(node) && node->getStmtExpr() != NULL) {
```
vasslitvinov: For uniformity, let us use `node->inTree()` instead.

compiler/AST/LoopExpr.cpp
```diff
@@ -0,0 +1,648 @@
...
+  FnSymbol* faParent = loopExpr->getFunction();
+  for_vector(SymExpr, se, uses) {
+    Symbol* sym = se->symbol();
+    if (VarSymbol* var = toVarSymbol(sym)) {
+      if (isOuterVar(var, faParent) == true && isGlobal(var) == false) {
+        outerVars.insert(var);
+      }
+    } else if (ArgSymbol* arg = toArgSymbol(sym)) {
+      // There might be nested functions within the loopExpr, so we need to
+      // check for those functions' defPoints instead of the arg's defPoitn.
```
vasslitvinov: defPoitn -> defPoint

compiler/AST/LoopExpr.cpp
```diff
@@ -0,0 +1,648 @@
...
+  }
+
+  return fifn;
+}
+
+//
+// Lifted from 'createTaskFunctions'
+//
+// TODO: build a more general flattening or outer-finding set of functions
+//
```
vasslitvinov: Please explain the role - and the naming (why "fa"?) of `faParent` in a short comment.

compiler/AST/Makefile.share
```diff
@@ -32,8 +32,9 @@ AST_SRCS =                                          \
            flags.cpp                                \
            FnSymbol.cpp                             \
            ForallStmt.cpp                           \
-	   IfExpr.cpp			            \
+           IfExpr.cpp			            \
```
vasslitvinov: Add one more space before the backslash after IfExpr.cpp, so it aligns with the other backslashes? Are there tabs on this line that are not on the other lines?

test/compflags/ferguson/unstable-class.good
```diff
@@ -1,5 +1,3 @@
-unstable-class.chpl:20: warning: undecorated class type MyClass is unstable
-unstable-class.chpl:20: note: use 'unmanaged MyClass' 'owned MyClass', 'borrowed MyClass', or 'shared MyClass'
```
vasslitvinov: Is the above warning gone for a good reason?

compiler/include/LoopExpr.h
```diff
@@ -0,0 +1,66 @@
...
+#define _LOOP_EXPR_H_
+
+class LoopExpr : public Expr {
+public:
+  AList defIndices;    // Flat AList of DefExprs, generated by constructor
+
+  Expr* indices;       // Optional: [Unresolved]SymExpr or CallExpr to _build_tuple
+  Expr* iteratorExpr;  // Expr or CallExpr to _build_tuple in zippered case
+  BlockStmt* expr;     // Loop body
+  Expr* cond;          // Optional: conditional expression (NB: not an IfExpr)
```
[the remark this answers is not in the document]
benharsh: Let's keep it 'cond' and say // filter condition, or NULL if none

compiler/AST/AggregateType.cpp
```diff
@@ -1854,6 +1855,14 @@ void AggregateType::fieldToArg(FnSymbol*              fn,
           arg->addFlag(FLAG_TYPE_VARIABLE);
         }
 
+        if (LoopExpr* fe = toLoopExpr(defPoint->init)) {
+          if (field->isType() == false) {
```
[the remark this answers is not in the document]
benharsh: I have a mild preference for ``foo == false`` over ``!foo``, because sometimes my eyes will too quickly scan past the ``!``. But I don't feel that strongly about it.

compiler/AST/LoopExpr.cpp
benharsh: Ah, I think that was carried over from when the branch only worked for ForallExprs (ForAll). It could be renamed to something else entirely at this point.

test/compflags/ferguson/unstable-class.good
benharsh: It's not gone, just printed in a different order (I think).

PR #10112, opened 2018-06-30
Ignore unmanaged, etc. on records with initializers

benharsh:
The compiler was immediately lowering PRIM_NEW on certain records into an 'init' call, but left the NamedExpr for chpl_manager in place. This would later lead to resolution errors. Instead, remove the NamedExpr when creating the 'init' call.

Resolves a --force-initializers failure for ``classes/delete-free/owned/owned-raw-ingored-record.chpl``

Testing:
- [x] local + futures
- [x] gasnet

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -2261,6 +2261,10 @@ static void normVarTypeInference(DefExpr* defExpr) {
         // Add a call to postinit() if present
         insertPostInit(var, argExpr);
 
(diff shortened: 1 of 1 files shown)


benharsh:
@mppf : would you be willing to review?

benharsh:
@mppf : your PR #10171 resolved the --force-initializers failure this PR was attempting to fix. Do you think this should still be merged?

PR #10114, opened 2018-07-01
Disable enum->int casts for enums without values given to them

### LINKED ISSUE #10026: Disable enum->int casts when enums are lacking associated values
As a Chapel programmer, I don't want my enums to cast to ints if I haven't associated values with them in order to support the concept of a truly abstract enum type while also avoiding assumptions about whether enums use 0- or 1-based numbering by default.

Acceptance criteria:
* enum->int casts are only supported for those enum symbols that have associated integer values

bradcray:
Traditionally, we've given all enums integer values whether or not the user has requested them, starting at an arbitrary value and counting up from there.  At times, users have said that they'd prefer not to have integer values backing enums while others have insisted on their utility for associating (say) bit masks or other integer values with their enums.  This effort tries to support the best of both worlds by only supporting casts between enums and ints for those enums which have values associated with them.  This permits those who want a "pure enum" to do so by not associating any integer values with them while also avoiding a case where Chapel had previously arbitrarily decided between 0- and 1-based numbering.  Note that, as before, enums following one that is given a value will count upwards by 1 (i.e., not every symbol in an enum need be given a specific value).  Note also that any enums preceding the first to be given a value will not have values associated with them (i.e., we don't "count down" from an initial enum that's given a value).

Changes to the implementation

* `buildDefaultFunctions.cpp:` The heart of this PR was to update the `build_enum_cast_function()` routine to:
  * have this routine create an enum->int routine where it previously had not
  * only build enum<->int casts in the event that some of the enum's constants had integer values associated with them
  * support the case where some enums have integer values, but the initial one(s) do(es) not

* `ChapelBase.chpl:` Since we no longer necessarily support enum->int casts, I removed the existing enum->any primitive type cast and replaced it with explicit enum->real and enum->bool casts that go through an integer cast.  That way, if the integer cast isn't supported, these won't work; and if it is, they will.

* `functionResolution.cpp:` In addition, I updated the `ensureEnumTypeResolved()` routine to:
  * handle enums in which some or all of the symbols don't have associated integer values (where before, we'd just start counting from "1")
  * handle an error case where an enum stores both negative values and those greater than max(int), previously handled by `EnumType::sizeAndNormalize()` (removed as noted below)

* `preFold.cpp:` One of the most subtle and annoying changes was that the preFold logic had to be updated now that not every enum symbol can be assumed to have an underlying integer value.  The implication of this is that before, there was a large param-cast-folding conditional that was predicated on `imm` (an immediate value representing the source of the cast) being non-null.  However, for enum->string casts, it's possible to fold these casts even if `imm` is null.  For this reason, I pushed the test of `imm` being non-NULL down into the specific cases that required it while touching up the enum-based casts.

* `type.cpp` / `type.h:` removed some code that was no longer needed:
  * removed the `minConstant` and `maxConstant` symbols from `EnumType`
  * removed the very long and complex `EnumType::sizeAndNormalize()` routine which is no longer needed

* `IO.chpl:` Rewrote binary I/O to avoid relying on casts to integers which are no longer guaranteed to work.  Instead, use the enum's 0-based order in the enumeration.

* `Types.chpl:` 
  * redefined the `enum_mintype()` and `enum_minbits()` functions to do their computations based on the number of constants defined by the enumeration rather than a primitive now that we're using the enums' order for binary I/O.  Added a `chpl_` prefix to these routines since they're not really for end-user use.
  * removed the `numBits()` routine for enums which seems poorly motivated (we don't really know what size they will be in the generated code since we codegen to C enums and its sizes aren't particularly well-defined).

* `ChapelArray.chpl:` Removing support for enum<->int casts and numBits() on enums revealed some places where we were still relying on casts between ints and enums, and other cases where we were still relying on `idxType` rather than `intIdxType` 

* `spec/*.tex:` Updated the spec w.r.t. most recent enum-related changes and introduced notions of "abstract", "concrete", and "semi-concrete" enumerations in order to distinguish between those cases that could or could not be converted to integers.

Other supporting changes
---------------------------
* made a few parts of the compiler not worry so much if it didn't know what an enum's underlying type or value was
* fixed a bug in which the `chpl__enumToOrder()` routine was not properly marked as being generic

New tests
----------
* `test/types/enum/enumCast.chpl` tests enum casts for abstract, concrete, and semi-concrete enums
* `test/types/enum/paramEnumCast.chpl` is similar, but for `param` cases
* `test/types/enum/sizeIsParam.chpl` checks that `.size` on an enum is a `param` after noting that spec didn't say that it was (and fixing it)
* `test/types/enum/nonParamEnum.chpl` was my wondering whether we'd relaxed the assumption that enums had to be param values (answer: not yet, but I filed issue #10115 to ask this question as I think it would not be hard to support)

New test variants
------------------
* `test/types/enum/deitz/test_enum4b.chpl` is a clone of test_enum4.chpl that shows that we no longer have a default initial value
* `test/types/enum/diten/paramEnumOutOfBounds2.chpl` is a clone of paramEnumOutOfBounds.chpl that shows that we get the similar OOB messages for an enum with no values associated with its symbols

Test updates
-------------
* Remove casts that are no longer legal nor necessary:
  * test/users/npadmana/npb-mg/mg.chpl

* Retire future that's now passing:
  * test/types/enum/scope-param-offset-type-2.chpl

* Update .bad for future that's still not passing (trade one internal error for another):
  * test/types/enum/scope-param-offset-concrete.bad

* for a number of tests that were relying on enums having a default numbering starting with 1, I updated the tests to explicitly give them this starting value:
  * test/arrays/bradc/enumIterate.chpl
  * test/classes/bradc/arrayInClass/arrayOfArithInClass.chpl
  * test/classes/bradc/arrayInClass/genericArrayInClass-arith.chpl
  * test/classes/bradc/arrayInClass/genericArrayInClass-otharrs.chpl
  * test/classes/bradc/arrayInClass/genericArrayInClass-real.chpl
  * test/classes/bradc/arrayInClass/genericClassArray-named.chpl
  * test/functions/diten/param_enum_to_int.chpl
  * test/library/packages/ZMQ/req-rep.chpl
  * test/parallel/taskPar/figueroa/taskParallel.chpl
  * test/trivial/bradc/protoast/codegen/enum.chpl
  * test/types/enum/compare.chpl
  * test/types/enum/deitz/test_enum1.chpl
  * test/types/enum/deitz/test_enum1b.chpl
  * test/types/enum/deitz/test_enum2.chpl
  * test/types/enum/deitz/test_enum4.chpl
  * test/types/enum/deitz/test_enum5.chpl
  * test/types/enum/deitz/test_enum_overload.chpl
  * test/types/enum/ldelaney/enum-iterator.chpl
  * test/types/enum/lydia/enumCastInGenericFunc.chpl
  * test/types/enum/lydia/enumIterateInGenericFunc.chpl
  * test/types/enum/primCasts.chpl
  * test/types/enum/sungeun/cast_1.chpl
  * test/types/enum/sungeun/cast_oor_1.chpl
  * test/types/enum/sungeun/cast_oor_2.chpl
  * test/types/enum/typed-offset2.chpl

* Ditto, and also simplified some math by making the enum 0-based rather than 1-based
  * test/studies/shootout/chameneos-redux/hannah/chameneos-redux-cas.chpl
  * test/studies/shootout/chameneos-redux/hannah/chameneos-redux-v1.chpl
  * test/studies/shootout/chameneos-redux/hannah/chameneos-redux.chpl
  * test/studies/shootout/chameneos-redux/lydia/chameneos-redux-lydia.chpl
  * test/studies/shootout/chameneos-redux/lydia/chameneos-redux-waitFor.chpl
  * test/studies/shootout/chameneos-redux/lydia/chameneos-redux-yield.chpl

* Updated test to reflect name and behavior change of this internal routine:
  * test/types/enum/ferguson/enum_mintype_test.chpl

Resolves #10026


bradcray:
@lydia-duncan, @daviditen, and @mppf may want to check the changes to their tests in this PR (those which have your names in the paths).

daviditen:
The changes in my tests look good.

bradcray:
@mppf may be a good reviewer for this PR given the large amount of his code that I've removed / rewritten as a result of it.

mppf:
Just from the PR message, I'm not sure I'm (yet) convinced that always doing I/O with the enum order (rather than an associated integer value should one exist) is the right move. Is there an issue to talk about this? I don't want it to hold up this PR. (The only question is whether or not enums with associated integer values for all the enum constants should use the integer values when doing I/O - but clearly the order is all we have right now for some enums).

compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -879,17 +893,73 @@ static void build_enum_cast_function(EnumType* et) {
...
+      new CondStmt(new CallExpr(PRIM_WHEN),
+                   new BlockStmt(new CallExpr("halt",
+                                              new_StringSymbol(errorString))));
+    whenstmts->insertAtTail(otherwise);
+    fn->insertAtTail(buildSelectStmt(new SymExpr(arg2), whenstmts));
+
+    def = new DefExpr(fn);
+    //
+    // these cast functions need to go in the base module because they
+    // are automatically inserted to handle implicit coercions
```
mppf: Is this comment still accurate? I would have thought they would no longer be used for implicit coercions?

modules/standard/Types.chpl
```diff
@@ -596,30 +594,25 @@ iter type enumerated.these(){
...
-private proc enum_minbits(type t: enumerated) param {
-  return __primitive( "enum min bits", t);
+private proc chpl_enum_minbits(type t: enumerated) param {
+  if t.size <= max(uint(8)) then
+    return 8;
+  if t.size <= max(uint(16)) then
+    return 16;
+  if t.size <= max(uint(32)) then
+    return 32;
+  return 64;
```
mppf: It's going to be a bad day for the compiler when somebody makes an enum with 2**63 constants.

spec/Conversions.tex
```diff
@@ -322,16 +311,22 @@ \subsection{Explicit Enumeration Conversions}
 \index{conversions!explicit!enumeration}
 
 Explicit conversions are allowed from any enumerated type to any
-integer or real type, \chpl{bool}, or \chpl{string}, and vice versa.
+\chpl{string} and vice-versa, and include \chpl{param} conversions.
+For enumerated types that are either concrete or semi-concrete,
```
mppf: might be nice to refer to a section defining concrete and semi-concrete enums

test/types/enum/diten/paramEnumOutOfBounds2.good
```diff
@@ -0,0 +1,2 @@
+paramEnumOutOfBounds2.chpl:3: error: enum cast out of bounds
+paramEnumOutOfBounds2.chpl:3: error: illegal cast from int(64) to enumt
```
mppf: This error message seems a bit misleading. The issue is that the enum is abstract, so it doesn't support casts from integers, rather than that the integral value was too small or too great.

bradcray:
> I'm not sure I'm (yet) convinced that always doing I/O with the enum order (rather than an associated integer value should one exist) is the right move. Is there an issue to talk about this?

I just opened issue #10139 to capture my thinking between the options and support discussion.  Let me know if you think I missed an argument.

compiler/passes/buildDefaultFunctions.cpp
bradcray: Good point.  We seem to run into problems whenever the compiler-provided enum functions aren't at this scope, so I'm updating the comment to avoid the mention of implicit coercions rather than moving where the functions are defined.

modules/standard/Types.chpl
bradcray: I'm not sure how to interpret this.  Are you seeing a bug / limitation that I'm not?

mppf: No, I'm just saying that the compiler theoretically supports it but one will run out of memory and time before it would ever matter.

test/types/enum/diten/paramEnumOutOfBounds2.good
mppf: sure, that sounds fine

Issue #9752, opened 2018-06-06
fix proc f(x) when instantiated with owned


mppf:
As a Chapel programmer, I need there to be a reasonable default for generic functions instantiated with owned or shared actuals.

* According to our design in #8938, passing a `new owned C()` to a generic function with dtAny argument`proc f(x)` will instantiate that function with a borrow. That's not happening now - instead right now it instantiates with a `const ref x:owned C`.
* Should a similar thing be happening for shadow variables for forall statements? What about coforall / cobegin / begin? Discussed in #10142.

mppf:
test test/arrays/deitz/test_block_array_of_ints_assigned_class_field.chpl  fails if using owned due to the forall capturing an owned.

mppf:
I have a branch that does this. It currently has 6 failures.

mppf:
- [ ] What about `proc f(in x)` or `proc f(const ref x)` when instantiated with an `owned` actual?

Issue #8938, opened 2018-03-23
owned & borrowed syntax proposal: new returns borrow


mppf:
Previous: #8797. Next: none.

This is a variation on #8654.

General idea: `MyClass` and `borrowed MyClass` both refer to a borrow, `owned MyClass` refers to an owning pointer. `new` calls can be marked as `owned` or `unmanaged`, but one that is not marked returns the result of `(new owned C(...)).borrow()`.  Additionally, when instantiating generics, Owned/Shared non-type arguments are replaced by Borrows unless the generic explicitly opts in to accepting Owned/Shared. The default intent for Owned/Shared is to perform ownership transfer / sharing (ie `const in`).

A generic argument or field can opt in to accepting owned/shared with the following generic types:
 * `managed` means must be owned/shared
 * `managed?` means it could be owned/shared/borrow/int
 * `owned` means it must be owned
 * `shared` means it must be shared
 * `borrowed` means it must be a class borrow

The general idea driving the special rule for the instantiation of owned/shared in this proposal is that it would be surprising for generic code (especially code that is not written with owned classes in mind) to have ownership transfers by default. However once a generic argument has a type or partially specified type, be it a class type (borrow) or owned or ..., it becomes clear that ownership transfer should or should not occur. So, this special rule need only apply to totally generic arguments (those with no type as in `proc f(x)` or those with a queried type like `proc g(x:?t)`.

Edits to this issue:
 * updated to treat instantiations of generic functions with Owned arguments as borrow type by default, per discussion in #8618.
 * updated to imagine default intent of Owned/Shared is ownership transfer rather than borrow.
 * described type constructor error
 * unsafe -> unmanaged per discussion in #8374
 * special rule for instantiation does not apply to type arguments or compiler-generated default initailzers
 * `borrowed` keyword is added

### Locales Primer Example
``` chapel
class Node {
  var data: real;
  var next: owned Node;
}
var head    = new Node(0); // this is a (new Owned(new Node(0)).borrow()
var current = head;

for i in 1..numLocales-1 do on Locales[i] {
  current.next = new owned Node(i); // forgot 'owned'? error, borrow doesn't coerce to owned
  current      = current.next.borrow();
}
```

### Generic Collection Example
``` chapel
record Collection {
  var element;
  // compiler-generated initializer uses `managed?` on the element argument.
  // compiler-generated type constructor is not subject to the owned->borrow substitution
  // rule since it only consists of type arguments.
[...]

PR #10116, opened 2018-07-02
instantiating dtAny with owned results in borrow

mppf:
The design for owned and borrowed in #8938 includes instantiating functions 
with totally generic arguments such as `proc f(x)` with `borrowed T` assuming 
the actual was `owned T`. This PR makes this change.

Relates to #9752 and #8938.


mppf:
Early versions of the unmanaged support had those but I removed most of them. We have AggregateType -> getUnmanagedClass and several methods that get the Canonical type (but that only is really useful for unmanaged vs AggregateType). For things like _owned, we didn't have a method for it yet.

PR #10129, opened 2018-07-03
Switch perf playground to measure cstdlib atomics

ronawho:
We had performance regressions for several tests the last time we tried to use
cstdlib atomics. That was with gcc 6.2 in September of 2016. We're only on gcc
6.3 for most nightly testing now, so I expect to have similar regressions. Once
those regressions are confirmed I want to check with gcc 8.1 so see if things
have improved.

This also drops the playground down to 1 trial since it's been running pretty
late into the day and generally the extra trials don't really help us.


ronawho:
@dmk42 just fyi

dmk42:
Thanks for doing this!

PR #3154, opened 2016-01-21
Optimize iteration over anonymous low-bounded counted ranges

ronawho:
Similar to how we optimize iteration over anonymous (possibly strided) fully
bounded ranges (added with #1021), but for anonymous low-bounded counted
ranges.

This effectively replaces code like:

```chapel
for i in low..#count {}  (note that `low..#count` is equivalent to `low..low+count-1`)
```

with:

```chapel
for i in chpl_direct_range_iter(low, low+count-1, 1) {}
```

Low-bounded counted ranges are used a lot in our internal iterators (particular
leader and standalone iterators) and are pretty common in user code because
it's easier than writing `low..low+count-1`. Because of this it's worthwhile to
optimize this specific case too. This was also listed in the next steps when
the anonymous range optimization was originally added. Note that this doesn't
add support for fully bounded counted ranges or counted and strided ranges.

This optimization has a pretty nice impact on the generated code but like the
original optimization we don't expect this to have a huge impact on
performance. That said it is important for vectorization and it does eliminate
a few unnecessary function calls.

See deb9f3cf2c48dd7c150b9cf1a2bb64b7514e8c82 for a generated code example.

PR #10130, opened 2018-07-03
Add a flag to disable the anonymous range optimization

ronawho:
Add a flag to disable the anonymous range optimization and turn it off under
--baseline. The anonymous range optimization was added in #1021 and #3154. It
tries to optimize iteration over anonymous ranges so that something like:

```chpl
for i in 1..10 do writeln(i);
```

doesn't actually create a temporary range just for iteration. Instead we omit
the range creation and directly iterate between the bounds of the range.
Previously, this optimization was always on and there was no way to disable it.
In hindsight there always should have been a flag to disable this optimization,
but this has only recently occurred to us.

The original implementation exposed existing leaks (details in #3195) that we
suppressed for a number of tests. Turning this off under baseline allows us to
remove the suppressions and will quiet similar failures that we're currently
seeing under --baseline testing for arrays/ferguson/pushback-no-leak and
distributions/robust/arithmetic/reindexing/simple-alias-mem-use


ronawho:
Passed full std and `--baseline` testing

PR #10131, opened 2018-07-03
Suppress processor-local atomics test for CCE > 8.6.4

dmk42:
CCE versions greater than 8.6.4 have issues with processor-local atomics.  Suppress the affected test for now.

### DIFF
--- a/test/runtime/configMatters/comm/atomics.suppressif
+++ b/test/runtime/configMatters/comm/atomics.suppressif
@@ -1,0 +1,17 @@
+#! /usr/bin/env bash
+#
+# CCE versions greater than 8.6.4 have issues with processor-local atomics,
+# so suppress this case for now.
+#
+eval `$CHPL_HOME/util/printchplenv --make`
+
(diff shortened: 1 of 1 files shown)


ronawho:
Not really important since it's just a suppressif, but for things like this I have a slight preference for using our existing methods for checking compiler vendors/version Something like: 

```python
#!/usr/bin/env python

import os, sys

rel_dir = '../../../../util/chplenv/'
chplenv_dir = os.path.join(os.path.dirname(__file__), rel_dir)
sys.path.insert(0, os.path.abspath(chplenv_dir))

import compiler_utils, chpl_compiler

compiler = chpl_compiler.get('target')
cray = compiler == 'cray-prgenv-cray'
cur_version = compiler_utils.get_compiler_version(compiler)
bad_version = compiler_utils.CompVersion('8.6.4')

print(cray and cur_version > bad_version)
```

dmk42:
Thanks.  I'll keep that in mind for similar situations.  Fortunately, this should be just temporary.

PR #10135, opened 2018-07-03
Fix arg type exprs containing generic

mppf:
* Support generic arguments with types like `arg:Wrapper(SomeGenericType)`
* Support querying homogenous tuples with e.g. `arg:4*?t`
* Move hasGenericDefaults into methods on AggregateType (isGenericWithDefaults(), markAsGenericWithDefaults() )
* Fix a few bugs with handling of generic fields


compiler/passes/normalize.cpp
```diff
@@ -3350,12 +3371,14 @@ static void expandQueryForGenericTypeSpecifier(FnSymbol*  fn,
 }
 
 static TypeSymbol* getTypeForSpecialConstructor(CallExpr* call) {
-  if (call->isNamed("_build_tuple")) {
+  if (call->isNamed("_build_tuple") || call->isNamed("*")) {
```
lydia-duncan: Might be good to have an assert that the call is not PRIM_MULT (in case we change the behavior of "isNamed" later).

compiler/resolution/functionResolution.cpp
```diff
@@ -9956,29 +9973,28 @@ static bool primInitIsIteratorRecord(Type* type) {
...
+      call->replace(typeConsCall);
 
-        retval = (tryResolveCall(typeConsCall) == NULL) ? true : false;
+      retval = (tryResolveCall(typeConsCall) == NULL) ? true : false;
 
-        // Put things back the way they were.
-        typeConsCall->replace(call);
-      }
+      // Put things back the way they were.
+      typeConsCall->replace(call);
```
lydia-duncan: lol, thank you for cleaning up this function, what a weird set up!

Issue #9885, opened 2018-06-15
Run script against data during import


benjamin-robbins:
As a Chapel Developer I want to run a preprocessing function against my data during import so that I can transform my data as needed in line during import rather than as a separate step.

Success Criteria:
Select HDF5 Files for import
Select script for use as preprocssing function
Run function against data as import is happening
Data is transform in array/dataframe

bradcray:
By "script" do you mean a literal external script or some preprocessing computation written in Chapel itself?

daviditen:
My understanding of this task is that we'd provide something like a `HDF5Preprocessor` class with a `preprocess` method that could be extended with `preprocess` overloaded to do whatever you want with the data.  The `preprocess` overload could then call other Chapel routines or call out to external scripts or programs to manipulate the data.

daviditen:
Closed with PR #10138.

PR #10138, opened 2018-07-03
Add a preprocessing capability when reading HDF5 files

daviditen:
Add a class with a 'preprocess(A: [])' method to the HDF5_HL module.  This
class can be inherited with the method overridden and then passed to the
various Chapel HDF5 reading functions to preprocess the input before it is
returned.

Added two example classes that inherit from the new interface class and provide
working 'preprocess()' methods.  One class takes and integer and adds that
to every element in the array.  The other class takes a string that is the
text of a script that reads from stdin and writes to stdout for each element.
It writes the script to a file, then spawns a process calling it.  It passes
the array through it using its stdin and stdout pipes.

Added new copies of existing tests that use the above two example classes.


daviditen:
This is for issue #9885.

test/library/packages/HDF5/HDF5Preprocessors.chpl
```diff
@@ -0,0 +1,62 @@
...
+        //var f = opentmp();
+        //const scriptName = f.realPath();
+
+        const scriptName = "hdf5TempScript.bash";
+        var f = open(scriptName, iomode.cw);
+
+        // write the script to a file
+        {
+          var writer = f.writer();
+          writer.writeln(script);
```
cassella: Would it be worth having a way to give `ScriptPreprocessor` the name of an existing script, instead of the contents?  That way you don't need write access to the current directory, and there's no race between tasks opening (and truncating) and writing the same data to the same file while tasks that got there earlier are trying to execute it.

daviditen:
Thank you!

Issue #10060, opened 2018-06-27
proposal to replace : in where clauses


mppf:
Historically, the Chapel compiler has changed `A:B` in where clauses to mean to query whether A is a subtype of B (where here subtype also includes instantiation, in case B is generic).

This issue proposes changing that behavior:
 * `:` in where clauses would mean a cast, just like in other execution contexts
 * a new included-by-default `proc` will allow the previous query
 * the new `proc` is a function taking two types and returning a param bool
 * the new proc can be called in where clauses and also outside of where clauses

The main question I have is whether conceptually we should refer to such a query as checking if A is a subtype of B or if it should be checking if A is convertible to B (or coercible). According to the [Wikipedia page on subtyping](https://en.wikipedia.org/wiki/Subtyping), there is an idea of an *inclusive* subtype vs a *coercive* one. In Chapel, if B is a parent class to A, then a variable of type A is inherently also a B (this is *inclusive*). But `int(8)` can only convert to `int(64)` as a new value (and so this is *coercive*). But my point here is that we *could* use the terminology of just *subtype* to refer to both situations.

The closest previous work I can find is C++ type traits. They have `is_convertible` (which is the closest to the present question) as well as several other related queries like `is_constructable` and `is_assignable`.

I'm proposing that `isSubtype` be the name of the new function, and that `isSubtype(A,B)` can return `true` in all of these cases:
 * `A` is the same type as `B`
 * `A` is a class type that inherits from `B`
 * `A` is a generic instantiation of `B`
 * `A` is coercible to `B`.

(Relates to #10019).

bradcray:
Since my eyes were recently opened, I'm a huge supporter of ceasing to treat `:` specially in where clauses.  A few notes:

> just like anywhere else

I think this should read "just like in other execution contexts" (since, in declaration contexts, they're less casts and more type specifiers / constraints).

If I were implementing this, I'd see how far we could push things without relying on a new procedure that does what `:` does currently.  E.g., if I found code like `x:integral`, I'd change to `isIntValue(x)` rather than `isSubType(x, integral)`.  In some cases, this will probably mean needing to understand what the function is trying to constrain rather than applying a global query-replace.  My pipe dream would be that no `isSubtype()` predicate would be required.  More realistically, I'd hope that it would only be required in a relatively small number of places which might illuminate what it should mean.

mppf:
[...]

Issue #10019, opened 2018-06-26
remove/deprecate special treatment for `:` in where clauses


mppf:
As brought up in issue #9273, the use of `:` in where clauses to mean "is subtype" vs "cast" is confusing and not documented. This behavior should be removed or deprecated and replaced by a call to some function, e.g. `canCoerce`.

Steps to do that are:
 * reach design consensus (working on that in issue  #10060)
 * update the compiler to warn on `:` used in where clauses with `--warn-unstable`
 * add `canCoerce` or whatever type queries are agreed upon

mppf:
Future work:
- [ ] tidy up remaining uses of isSubtype (array heirarchy ones should call 
   e.g. isDefaultRectangular etc)
- [ ] create deprecation error for : in where (in changeCastInWhere)
- [ ] look into supporting `<=` and other comparisons on types
- [ ] look into behavior changes if isSubtype allows coercions

Issue #9967, opened 2018-06-22
Should casts to borrowed or unmanaged be allowed?


mppf:
Several test and module codes have a workaround to call _to_borrowed or _to_unmanaged on a type or on a value to get the corresponding borrow / unmanaged type.

Should it be allowed to write these as casts, where the RHS is theoretically generic, as in:
``` chapel
  type t = unmanaged MyClass;
  // This sort of thing usually comes up in a generic function
  t:unmanaged; // means use the unmanaged version aka unmanaged MyClass
  t:borrowed; // means use the borrowed version aka borrowed MyClass
```

This specifically relates to #9273.

mppf:
I'm not so much trying to emphasize casting *the type* as much as casting to literally `unmanaged`.

Here is an alternative example:
``` chapel
class MyClass {
 var x: int;
}
class Container {
 var field;
}
proc Container.borrow() {
  return field:borrowed;
}
```

Here the point is that `field:borrowed` is casting to a generic type (in a manner) but I think it has clear meaning & makes such a pattern possible. Right now this (and the type version above) are both written with `_to_borrowed` or `_to_managed` (i.e. using a work-around strategy).

psahabu:
Ah okay, seeing the use case I support this pattern.

PR #10143, opened 2018-07-03
use isSubtype in preference to : in where clauses

mppf:
As a first step towards solving #10060, update compiler and modules and tests
to use `isSubtype` in preference to `:` in where clauses. When possible,
removed where clauses entirely while there. To aid in removing where clauses,
added dtAnyReal, dtAnyImag, dtAnyBool types.

In some places, used the generic version (e.g. `chpl_anyimag`) in preference to
the `imag(?w)` version. The reason for this is that I wanted to avoid the
stamping-out-functions-for-all-types behavior where possible since that
increases the AST size. (see cloneParameterizedPrimitive in normalize.cpp).

Note that isSubtype is already a documented method in Types.chpl.

Relates to #10019.

Future work:
 * tidy up remaining uses of isSubtype (array heirarchy ones should call 
   e.g. isDefaultRectangular etc)
 * create deprecation error for : in where (in changeCastInWhere)
 * look into supporting `<=` and other comparisons on types
 * look into behavior changes if isSubtype allows coercions


compiler/include/type.h
```diff
@@ -379,9 +379,12 @@ TYPE_EXTERN Type*             dtAny;
 TYPE_EXTERN Type*             dtIteratorRecord;
 TYPE_EXTERN Type*             dtIteratorClass;
 TYPE_EXTERN Type*             dtIntegral;
-TYPE_EXTERN Type*             dtAnyComplex;
 TYPE_EXTERN Type*             dtNumeric;
 TYPE_EXTERN Type*             dtAnyEnumerated;
+TYPE_EXTERN Type*             dtAnyBool;
+TYPE_EXTERN Type*             dtAnyComplex;
+TYPE_EXTERN Type*             dtAnyImag;
+TYPE_EXTERN Type*             dtAnyReal;
```
vasslitvinov: Let us move dtAny* up right below dtAny, and dtIterator* after dtNumeric, to match the order in type.cpp .

compiler/passes/ResolveScope.cpp
```diff
@@ -228,7 +228,10 @@ void ResolveScope::addBuiltIns() {
 
   extend(dtAny->symbol);
   extend(dtIntegral->symbol);
```
vasslitvinov: move dtIntegral to right above dtNumeric ?

compiler/passes/cleanup.cpp
```diff
@@ -91,9 +93,10 @@ static void cleanup(ModuleSymbol* module) {
       }
 
     } else if (CallExpr* call = toCallExpr(ast)) {
-      if (call->isNamed("_build_tuple") == true) {
+      if (call->isNamed("_build_tuple"))
         destructureTupleAssignment(call);
-      }
+      if (call->isNamed("isSubtype"))
+        replaceIsSubtypeWithPrimitive(call);
```
vasslitvinov: +1

modules/internal/CPtr.chpl
```diff
@@ -110,77 +110,109 @@ module CPtr {
   inline proc =(ref a:c_ptr, b:_nilType) { __primitive("=", a, c_nil); }
 
   pragma "no doc"
-  inline proc _cast(type t, x) where t:c_ptr && x:_nilType {
+  inline proc _cast(type t:c_ptr, x:_nilType) {
```
vasslitvinov: These are great changes.

Makes me wonder though whether we should rewrite _cast calls into PRIM_CAST like we now do for isSubtype, to avoid this long repetitive list.

modules/internal/CPtr.chpl
```diff
@@ -110,77 +110,109 @@ module CPtr {
...
+  inline proc _cast(type t:int(64), x:c_ptr) where c_intptr != int(64)
+    return __primitive("cast", t, x);
+  pragma "no doc"
+  inline proc _cast(type t:uint(64), x:c_ptr) where c_uintptr != int(64)
+    return __primitive("cast", t, x);
+
+
+
+
+
```
vasslitvinov: Too many newlines above?

modules/internal/ChapelArray.chpl
```diff
@@ -858,7 +856,7 @@ module ChapelArray {
   pragma "use default init"
   record dmap { }
 
-  proc chpl__buildDistType(type t) type where _to_borrowed(t): BaseDist {
+  proc chpl__buildDistType(type t) type where isSubtype(_to_borrowed(t), BaseDist) {
```
vasslitvinov: Ideally I would write this simply as `where t <= BaseDist` .

modules/internal/CString.chpl
```diff
@@ -175,14 +173,14 @@ module CString {
...
-  inline proc _cast(type t, x:real(?w)) where t == c_string {
+  inline proc _cast(type t:c_string, x:chpl_anyreal) {
     return real_to_c_string(x:real(64), false);
   }
 
   //
   // casts from imag
   //
-  inline proc _cast(type t, x:imag(?w)) where t == c_string {
+  inline proc _cast(type t:c_string, x:chpl_anyimag) {
```
vasslitvinov: Please comment briefly in the PR message on `x:imag(?w)` vs. `x:chpl_anyimag` .

test/compflags/lydia/noUseNoinit/ignoreNoInit.chpl
```diff
@@ -1,4 +1,4 @@
-proc _defaultOf(type t) where (isBoolType(t)) {
+proc _defaultOf(type t:bool) {
```
vasslitvinov: Should this be `t:anybool` ?

test/expressions/lydia/noinit/betterNoinit.chpl
```diff
@@ -1,4 +1,4 @@
-proc _defaultOf(type t) where (isBoolType(t)) {
+proc _defaultOf(type t:bool) {
```
vasslitvinov: Same - should this be `anybool` ?

test/functions/lydia/userDefinedDefaultFunction/redefine-bool.chpl
```diff
@@ -1,4 +1,4 @@
-inline proc _defaultOf(type t) param where (isBoolType(t)) return true:t;
+inline proc _defaultOf(type t:bool) param return true:t;
```
vasslitvinov: anybool ?

test/functions/lydia/userDefinedDefaultFunction/redefine-int.chpl
```diff
@@ -1,4 +1,4 @@
-inline proc _defaultOf(type t) param where (isIntegralType(t)) return 5:t;
+inline proc _defaultOf(type t:int) param return 5:t;
```
vasslitvinov: `t:integral` ?

mppf: These tests check that a user-defined _defaultOf can work in various situations. I left them using the specific type to reduce the possible conflicts with other shuffling of _defaultOf in the modules. Either way, I don't view _defaultOf as a user-facing feature so my personal view is that all these tests defining a _defaultOf should just be removed.

vasslitvinov: That's fine, then.

modules/internal/CPtr.chpl
mppf: We'd have to do that in resolution, right, because we don't know the types otherwise?

mppf: That seems like an interesting idea. Parsing : calls as PRIM_CAST would also help with #9967, where the casted-to-type might be generic (so can't be passed to a type fn right now).

Issue #8620, opened 2018-02-27
Add `array.front()` and array.back()`


BryantLam:
Following on from #8601, getting the last element from an array before executing `array.pop_back()` requires using the array's domain (or implicitly knowing what that last index is from the size).
```chpl
writeln(arr[arr.domain.last]);
```

Please consider adding C++ `std::vector`-style convenience functions **`array.front()`** and **`array.back()`** to retrieve the respective elements.

daviditen:
I think the request for these methods sounds good independently from what happens with #8601.

PR #10146, opened 2018-07-03
Add methods array.front() and array.back()

daviditen:
These methods return the first or last element in a 1-D rectangular array.
It is a compiler error to call them on other array types.  It is a runtime
error to call them on an empty array when bounds checking is enabled.

Added tests for these errors and a working case.


daviditen:
This is for issue #8620.

daviditen:
Thanks!

modules/internal/ChapelArray.chpl
```diff
@@ -2795,6 +2795,19 @@ module ChapelArray {
...
 
+    /* Return the last element in the array. The array must be a
+       rectangular 1-D array.
+     */
+    proc back() {
+      if !isRectangularArr(this) || this.rank != 1 then
+        compilerError("back() is only supported on 1D rectangular arrays");
+
+      if boundsChecking && isEmpty() then
+        halt("back called on an empty array");
```
BryantLam: The `halt` on really should be `throw`.

daviditen: PR #10191 just added halt wrappers for this case, so it should call one of them.  Eventually we may want it to be a `throw`, but not quite yet.

Issue #8601, opened 2018-02-26
array.pop_back() does not return its element


BryantLam:
```
# Chapel 1.16
proc pop_back()
  Remove the last element from the array, reducing the size of the domain by one.
```

`array.pop_back()` does not return its element. It should. (Why doesn't it?)

Also, `list.pop_front()` does return an element.

bradcray:
@daviditen : Can you weigh in on this one?  Is it just oversight?

daviditen:
The array-as-vector functionality was modeled heavily after C++ std::vector, and std::vector's pop_back() doesn't return the popped element for [exception safety](https://stackoverflow.com/questions/12600330/pop-back-return-value) reasons.

I haven't dug too deeply into the problem, but at a glance I think that if our copy-initializers are able to throw an error we're in the same boat as C++, but if not it can safely return the popped value.

BryantLam:
Exception safety is a good reason to not have it. --
`list.pop_front` shouldn't return an element in that case.

mppf:
> The array-as-vector functionality was modeled heavily after C++ `std::vector`, and `std::vector`'s `pop_back()` doesn't return the popped element for [exception safety](https://stackoverflow.com/questions/12600330/pop-back-return-value) reasons.

> I haven't dug too deeply into the problem, but at a glance I think that if our copy-initializers are able to throw an error we're in the same boat as C++, but if not it can safely return the popped value.

I disagree about this being a good enough answer here. Certainly it's the historical reason.

But Chapel and C++ have different memory management strategies - in particular while both support ideas like RAII and value types with copy constructors/initializers, the C++ model is more restrictive, significantly.

In particular, in C++ there are two starting-point assumptions that are not true for Chapel:
 1. The language supports value types with self-referential pointers
 2. Types must opt-in to "move" construction

In Chapel, we simply do not expect a record that contains a `c_ptr` to itself to remain a reasonable object. Since Chapel doesn't actually have pointers the way C++ does, and since it doesn't support ref fields, you'd actually have to use a `c_ptr` to even create this pattern. But the language does not support it.

As a result, for a call like this:
``` chapel
   var x = something.pop();
```

there is not any copy-initialization occurring in that line of code. The value is always "moved". (There are exceptions for things like array slices, and that is where I'd characterize the question as not entirely settled). This design is described in [CHIP 13](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/chips/13.rst).

[...]

PR #10148, opened 2018-07-03
Make array.pop_front() and array.pop_back() return the removed element

daviditen:
Make pop_front() and pop_back() return the element that is getting popped.

Add a simple test that prints the return value of both.


daviditen:
This is for issue #8601.

daviditen:
Thanks!

Issue #8561, opened 2018-02-23
How should `enum` values be represented in generated code?


bradcray:
The current Chapel compiler inspects the (implicit or explicit) integer values associated with an enumerated type and makes some attempt to map the symbols in our generated C code to the smallest integer type that can store all of the values, preferring unsigned integers over signed when in doubt.  This can be seen by compiling the following code, saving the C, and inspecting the generated `chpl__header.h` for the enumerated symbols:

```chapel
enum color1 {red, green, blue};
enum color2 {red = 257, green, blue};
enum color3 {red = -10, green, blue};  
enum color4 {red = -257, green, blue};                   
enum color5 {red = 1099511627776, green, blue};

var colorVar1: color1;
var colorVar2: color2;
var colorVar3: color3;
var colorVar4: color4;
var colorVar5: color5;
```

Specifically, the generated C contains the lines:

```c
typedef enum {color1_chpl_red_chpl = UINT8(1), color1_chpl_green_chpl = UINT8(2), color1_chpl_blue_chpl = UINT8(3)} color1_chpl;
typedef enum {color2_chpl_red_chpl = UINT16(257), color2_chpl_green_chpl = UINT16(258), color2_chpl_blue_chpl = UINT16(259)} color2_chpl;
typedef enum {color3_chpl_red_chpl = INT8(-10), color3_chpl_green_chpl = INT8(-9), color3_chpl_blue_chpl = INT8(-8)} color3_chpl;
typedef enum {color4_chpl_red_chpl = INT16(-257), color4_chpl_green_chpl = INT16(-256), color4_chpl_blue_chpl = INT16(-255)} color4_chpl;
typedef enum {color5_chpl_red_chpl = UINT64(1099511627776), color5_chpl_green_chpl = UINT64(1099511627777), color5_chpl_blue_chpl = UINT64(1099511627778)} color5_chpl;
```

showing the attempt to store the values using signed and unsigned integers of various sizes depending on the values.  This issue is designed to capture the tradeoffs of this approach.

@mppf and I both think that he may have implemented this, and in issue #8552 he captures the likely rationale as "I had convinced myself at some point that when doing binary I/O with a value of enum type, it should occupy as many bits as the enum needs to store all of the values. In the context of binary I/O, having enum values default to writing 8-byte values seems to be less likely to be what is wanted than e.g. a 1 byte value, since most enums have only a small collection of small constants."

[...]

Issue #10115, opened 2018-07-02
Should Chapel enums support non-param values?


bradcray:
In the old world, Chapel enums were essentially treated as a collection of associated param ints given the implicit conversions they supported and the fact that every enum symbol had an associated integer value.  In the current world, Chapel enums no longer necessarily have integer values and no longer support implicit conversions.  This makes me wonder whether we could / should relax the requirement that the initializing expressions be param integer values.  I'm not sure what it's buying us anymore.

gbtitus:
Could an enum type define a mix of symbols whose associated values were a mixture of params and non-params?  Or would the presence of at least one non-param associated value essentially make the whole type a non-param one?  I'm trying to think about the implications for users declaring param or non-param variables of such enum types.  Obviously we'd want the compiler to bark if it saw code declaring a param enum var but initializing it with an enum value of that type, whose associated value wasn't param.  I just don't know how hard we want the compiler to have to work to do that.  I guess there's also the question of how hard we want users to have to work to reason about such enum values.

gbtitus:
There's an interaction between this issue and #10139, I think.  Suppose we allow enum associated values to be execution-time values.  Then the following would be legal, presumably:

```chapel
enum epoch { past = 0, present = getCurrentTime():int, future = 999999999999 };
```

`epoch.present` probably won't be the same from run to run.  So if you wanted to be able to write a variable of that enum type in one run of a program and read it back in another, in order for it to make sense when read back the value stored in the file would have to be something that didn't depend on the associated value.

bradcray:
> So if you wanted to be able to write a variable of that enum type in one run of a program and read it back in another, in order for it to make sense when read back the value stored in the file would have to be something that didn't depend on the associated value.

Good point.  Note that other changes to an enum (such as reordering its symbols or changing the param values of some of them, e.g., via a config) could also cause problems in any binary I/O scheme that we're discussing here, but your example has the benefit of not requiring source changes or changes to compilation options to make the point.

gbtitus:
> Note that other changes to an enum (such as reordering its symbols or changing the param values of some of them, e.g., via a config) could also cause problems in any binary I/O scheme that we're discussing here ...

[...]

Issue #9786, opened 2018-06-09
Should enums support duplicate integer values?


bradcray:
At present, we permit multiple symbols within a single `enum` declaration to share the same integer value.  E.g.,

```chapel
enum color = {red=1, green=0, blue=1};
```

This is a bit odd, though it really only causes problems when casting integers back to enums (e.g., should `1:color` generate an error or pick the first match or ...?  This issue asks whether we should continue to support this or prevent it (and if we prevent it, should we prevent it at compile-time which would require all enum values to be `param` values or at execution-time, which would not?)

bradcray:
At present, our execution-time functions that converts enums to strings and enums to orders "break" (or, at least, give surprising results) when an enum contains duplicate integer values with the same value due to the fact that these comparisons become C code that compares values to enums using equality, and thus gets "false positives" when values are shared.  As an example, see the behavior of this test:

```chapel
enum color {red=1, orange=2, yellow=1, green=2, blue=1, indigo=2, violet=1};

writeln(color.red == color.violet);   	   // param evaluation: works           
writeln(color.red != color.violet);   	   // param evaluation: works           
writeln(chpl__enumToOrder(color.violet));  // param evaluation: works           

config const c = color.violet;
writeln(color.red == c);                   // exec-time eval: surprising        
writeln(color.red != c);      	      	   // exec-time eval: surprising        
writeln(chpl__enumToOrder(c));	      	   // exec-time eval: surprising        

writeln(chpl__orderToEnum(6, color));      // uses exec-time enum->string cast: surprising                                                                     
```

This might suggest disallowing enums to have repeated values.  Yet C permits them which gives me pause.  I'd be curious what @dmk42 thinks about this topic.

Or, is there some way we could write our execution-time enum-to-order and enum-to-string routines to make them more bulletproof?  The execution-time code for these routines, essentially looks like:

```c
// given aColor of type enum
if (aColor == red) then return 0;
if (aColor == orange) then return 1;
if (aColor == yellow) then return t;
...
```

and so we hit a false positive as soon as an enum with the shared value is reached.

bradcray:
If we did want to support enums with duplicate integer values and also wanted code like that in the above comment to be supported, we could always represent enums in generated code as their "order" (guaranteed to be unique) rather than using a C enum.

Such an approach would also make representing enums as integers in ranges slightly less weird (though it would likely still add some code cruft relative to traditional int ranges).

dmk42:
[...]

PR #10149, opened 2018-07-03
Get LLVM smoke testing back on track after disabling enum->int casts

bradcray:
Summary: This should fix the llvm smoke-test that I broke...

In PR #10114, I stopped giving each enum an integer value since we're
moving towards supporting abstract enums that don't have underlying
integer values.  However, this broke LLVM code generation which seems
to generate enum values as integers (h/t to DavidK for sleuthing that
out for me).  This PR takes a stab at getting things running again,
and seems to be working, though I haven't taken the time to do a full
LLVM build + test on chapcs...

Specifically, what I've done here is preserve the old logic (which uses the
enums' initializer values when they're available) while having cases
that don't define initializers simply define the integer values to be
the enum's 0-based order.  This should match what the C back-end is
currently doing.

This seems to get testing working again, but seems as though it could
cause problems if/when there are conflicts between a user's enum values
and the ordinal values.  But, I suspect that the C back-end would have the
same issues.

I _suspect_ that the better thing to do would be to always generate enums
as ordinal values as proposed in issue #8561 for both the LLVM and C
back-ends.

### DIFF
--- a/compiler/codegen/type.cpp
+++ b/compiler/codegen/type.cpp
@@ -110,12 +110,16 @@ void EnumType::codegenDef() {
       type = ty->codegen().type;
       info->lvt->addGlobalType(symbol->cname, type);
 
+      int order = 0;
       for_enums(constant, this) {
         //llvm::Constant *initConstant;
 
-        if(constant->init == NULL) INT_FATAL(this, "no constant->init");
-
-        VarSymbol* s = toVarSymbol(toSymExpr(constant->init)->symbol());
+        VarSymbol* s;
+        if (constant->init) {
+          s = toVarSymbol(toSymExpr(constant->init)->symbol());
+        } else {
+          s = new_IntSymbol(order, INT_SIZE_64);
+        }
         INT_ASSERT(s);
         INT_ASSERT(s->immediate);
 
@@ -123,6 +127,7 @@ void EnumType::codegenDef() {
 
         info->lvt->addGlobalValue(constant->sym->cname,
                                   sizedImmediate->codegen());
+        order++;
       }
     }
 #endif
--- a/compiler/codegen/type.cpp
+++ b/compiler/codegen/type.cpp
@@ -110,6 +110,12 @@ void EnumType::codegenDef() {
       type = ty->codegen().type;
       info->lvt->addGlobalType(symbol->cname, type);
 
+      // Convert enums to constants with the user-specified immediate,
+      // sized appropraitely, when it exists.  When it doesn't, give
+      // it the semi-arbitrary 0-based ordinal value (similar to what
+      // the C back-end would do itself).  Note that once some enum
+      // has a non-NULL constant->init, all subsequent ones should as
+      // well.
       int order = 0;
       for_enums(constant, this) {
         //llvm::Constant *initConstant;


bradcray:
@mppf: I don't know how I turned a blind eye toward LLVM codegen in PR #10114 after wondering about it in some enum-based issues this week, but I did and broke smoke testing as a result...  As you'll read above, in this PR, I seem to have gotten testing working again, though I don't think the approach I took is principled (specifically, I wonder if we should/could always codegen enums as their ordinal values in the LLVM back-end, which would be similar to what I've most recently proposed in issue #8561 and remove the chances of duplicate values between the ordinal values and the user values that it seems this patch could lead to).

Anyway, if you could give this a post-merge review and continue this discussion with me, I'd appreciate it.

bradcray:
After typing this and waiting for Travis to finish, I did a quick check to see if I could simply use the ordinal values (and never use the user's initial values), but this didn't work.  :(  This suggests that the LLVM-back-end is still using the user-provided integers for the enums in some cases (and maybe the C back-end is as well since I haven't changed that aspect of it).

Next, I'll try to write a test that exhibits a case where I'd expect this PR to cause problems...

compiler/codegen/type.cpp
```diff
@@ -110,19 +110,24 @@ void EnumType::codegenDef() {
...
+      int order = 0;
       for_enums(constant, this) {
         //llvm::Constant *initConstant;
 
-        if(constant->init == NULL) INT_FATAL(this, "no constant->init");
-
-        VarSymbol* s = toVarSymbol(toSymExpr(constant->init)->symbol());
+        VarSymbol* s;
+        if (constant->init) {
+          s = toVarSymbol(toSymExpr(constant->init)->symbol());
```
dmk42: Would you mind adding a comment that all enum constants after the first explicitly-assigned one will have consant->init set?

mppf:
@bradcray and @dmk42 - thanks for sorting it out. I agree that we should try to push towards enums generating only with ordinal values. One thing I don't understand is why we bother generating enum types differently than integer types. That is, the Chapel compiler knows each enum constant at compile-time, so why can't it just put in an integer value for them? Does the C/LLVM IR even need to know that the enum type != the appropriate integer type to store the ordinals?

bradcray:
At first I thought when you said "enum constant" you meant the logical values that the user can associate with the enums and wrote this:

> the Chapel compiler knows each enum constant at compile-time

It doesn't in the current world in which an enum with no associated constants doesn't result in associated integer values anymore.  And if we relaxed the requirement that enum constants be params (issue #10115) it'd know less.

> so why can't it just put in an integer value for them?

If we continue to permit enums to have duplicate integer values (issue #9786) then replacing enums with integers results in problems due to the potential many-to-one nature.  E.g., mapping an execution-time enum to its corresponding string value causes problems.  This is another advantage of using ordinal representation—it's 1:1 by nature.

Then your last sentence made me think that you might mean "ordinal value" by "enum constant" in which case I think you're right—that the execution time representation of an enum would be no different than some integer value and that the Chapel compiler would take care of resolving and implementing all the operations that were enum-specific.

PR #10166, opened 2018-07-05
Make segfaulting enum future more stable

bradcray:
This test was segfaulting in the compiler, but couldn't be counted on
to do so reliably and in the same way each night.  Instead of trying to
match against the segfault in a .bad, let's skip it for all configurations
other than ones in which we're valgrinding the compiler.


mppf:
yes

bradcray:
Testing:
* verified that it is skipped for vanilla linux64 testing
* verified that if failed to pass for valgrind linux64 testing

Issue #10142, opened 2018-07-03
default intent for shadow variables from owned


mppf:
Based on the design in #8938, the type instantiated for an unconstrained generic formal, when passed an owned actual, is the borrow type:

``` chapel
proc f(x) {
  // x.type is borrowed MyClass
}
f(new owned MyClass());
```

This issue asks if a corresponding behavior should apply to shadow variables. Consider:

``` chapel
var x = new owned MyClass()
forall a in A {
  writeln(x.type:string); // borrowed or owned?
}
```

Current Direction:

Make the shadow variable be `borrowed` by default.

* arguably more consistent with outer-vars-pass-like-blank-function-arguments
* less likely to have perf surprises

Alternative of transferring ownership doesn't make sense as a default coforall/forall

Alternative of 'const ref owned' could work but
    * wouldn't want this dereference for the formal of a generic
    * don't think it's necessary for shadow variables to track changes

By analogy to 'int' or 'unmanaged'
  * int/unmanaged is copied
  * shadow variable won't track changes to outer variable

When is transferring ownership to make a shadow variable useful?
 * not for coforall/forall since only one task gets the ownership & it's racy
 * (could conceivably work with shared but what is the point?)
 * more useful for begin - transfer ownership into a task

Presumably using an intent leaves the type alone:
``` chapel
var x = new owned MyClass()
forall a in A with (const ref x) {
  writeln(x.type:string); // owned
}
```

mppf:
Even if the shadow variable is borrowed by default, what should happen for non-default intents?

We could:
 a) use the intent to indicate ownership transfer or `ref` behavior in these cases
 b) completely ignore the intent (other than intent type) for the purposes of this change

Note that ownership transfer isn't really useful for `forall`/`coforall`/`cobegin` because it's inherently racy. However for `begin` it is useful.

Should this transfer an `owned` thing into the task?
``` chapel
var x = new owned MyClass()
begin (with in x) {
  writeln(x.type:string); // is it owned & destroyed by the task?
}
```

Or should the user instead write this:
``` chapel
var x = new owned MyClass()
begin (with var taskX = x) {
  writeln(taskX.type:string); // is it owned & destroyed by the task?
}
```

mppf:
Task private variables for `begin` don't seem to be working:

``` chapel
class MyClass {
  var x:int;
}

config const n = 4;
proc test() {
  
  {
    var outerOwnedMyClass = new owned MyClass(1);
[...]

PR #10168, opened 2018-07-05
Tasks borrow owned/shared variables by default

mppf:
Following the proposal in #10142, this PR adjusts the
compiler to default to borrowing an `owned` or
`shared` variable that is referred to in a task.


mppf:
- [ ] if we decide to avoid ownership transfer on `in`, cherry-pick afc5fdd34e5824bf6a5de94a5522ef81d728f212

PR #10171, opened 2018-07-05
Fix internal errors related to borrowed and owned

### LINKED ISSUE #10018: delete-free subtask: address internal errors
The following tests have internal errors when the warnings with --warn-unstable are addressed (i.e. when they are converted to use borrowed/unmanaged/etc).

test/classes/ferguson/generic-field/generic-field-unmanaged-borrowed.chpl
test/distributions/bradc/spsBlk/blockSparse2CSR.chpl
test/exercises/boundedBuffer-old/boundedBuffer2.chpl
test/functions/ferguson/query/subquery.chpl
test/functions/generic/genericTypeArgQuery.chpl
test/classes/initializers/compilerGenerated/nested-class-in-record.chpl

This task is to address these issues.

mppf:
* Avoid setting a wrong type when normalizing a managed new (e.g. `new owned C`)
* fix problems with nested type query and borrowed
* adjust specific tests to pass with --warn-unstable

Resolves #10018.

- [x] full local testing

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -2457,21 +2457,32 @@ static bool isNewExpr(Expr* expr) {
 }
 
 static AggregateType* typeForNewExpr(CallExpr* newExpr) {
-  AggregateType* retval = NULL;
 
   if (CallExpr* constructor = toCallExpr(newExpr->get(1))) {
+
+    // Avoid normalize-time type inference for managed new
+    for_actuals(actual, constructor) {
+      if (NamedExpr* ne = toNamedExpr(actual))
+        if (ne->name == astr_chpl_manager)
+          if (SymExpr* se = toSymExpr(ne->actual))
+            if (isTypeSymbol(se->symbol()))
+              return NULL;
+            //if (TypeSymbol* ts = toTypeSymbol(se->symbol()))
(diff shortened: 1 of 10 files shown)


compiler/passes/normalize.cpp
```diff
@@ -3464,6 +3475,17 @@ static void expandQueryForGenericTypeSpecifier(FnSymbol*  fn,
...
+    // Check that whatever it is is a borrow
+    addToWhereClause(fn, formal, new CallExpr(PRIM_IS_SUBTYPE,
+                                              dtBorrowed->symbol, queried));
+    // And then proceed as if the PRIM_TO_BORROWED_CLASS wasn't there
+    // that's because the borrowed class is the 'canonical' class representation
+    // used in the compiler.
+    call = toCallExpr(call->get(1));
+    // And don't try to do anything else if it wasn't a call
+    if (call == NULL)
+      return;
```
lydia-duncan: This makes me a little uneasy.  I don't like setting call to something if we are just going to return, and because of the pointer involvement I am a little worried this changes behavior in a way that could be difficult to follow for someone coming fresh to this piece of code.

mppf: I'll make it separately check isCallExpr and then return if not.

PR #10175, opened 2018-07-05
Build the third-party with position independent code when CHPL_LIMODE=shared

ronawho:


ronawho:
I need to make a similar change for GASNet, though I'm not quite sure how to do that yet.

There are also some less important third-party configs that I should probably update too:
 - [ ] massivethreads
 - [ ] libunwind
 - [ ] libhdfs3?

ronawho:
@PHHargrove and/or @bonachea are there any best practices for building GASNet as position independent?

I saw https://bitbucket.org/berkeleylab/gasnet/pull-requests/36/wip-build-shared-libraries, but it's not obvious to me what the right way to do this is. 

 - configure with `--with-cflags=-fPIC`?
 - something with `MANUAL_DEFINES`?
 - append `-fPIC` to CC/CXX/MPI_CC?
 - something else?

Do I need to define `-DGASNETI_FORCE_PIC=1` or is that for tools only?

bonachea:
@ronawho - GASNet supports configure and build with the `-fPIC` or `-fPIE` ABI flags in `CFLAGS`, `CXXFLAGS` and `MPI_CFLAGS` (assuming the backend compiler(s) support it), and that should “just work” - however it will still generate the GASNet library as a static .a library (containing relocatable objects). This is where GASNet stops, because our system is strongly designed around static linkage for many reasons (most explained in the comments to that declined PR). In particular, although one can convert the position-independent libgasnet*.a to a .so shared library, it’s important to understand the implications when deciding whether one should. 

In particular, the generated library would still be highly dependent on the exact GASNet version and configure-detected details of the system, and is in no way a “generic” GASNet that would be at all binary portable. We’ve previously discussed developing a low-performance binary-portable version of GASNet, but that would entail significant internal development work that goes far beyond simply wrapping things up in a shared library (eg ensuring all APIs at the internal .so library interface are stable across platforms and versions). If the intent is to package the relocatable code into a Chapel shared library built with the same GASNet configure, then the question becomes whether that .so will ever be moved to a system that might invalidate configure-detected system properties (ABI, compiler, conduits, network library .sos, etc).

The option -DGASNETI_FORCE_PIC is only needed if PIC flags were not provided at configure time. So, that should not be relevant to your scenario - the recommended place to provide these flags is `CFLAGS`, `CXXFLAGS` and `MPI_CFLAGS` at configure time.

ronawho:
Right, we have no interest in building .so's for our runtime/third-party either (for the same reasons as you), but we do have a need for an opt-in way to build the .a's as PIC. Our runtime/third-party will still be system dependent .a's though

Thanks for the info!

ronawho:
@bonachea 

Setting `CFLAGS` and `CXXFLAGS` to `-fPIC` (really to `$(SHARED_LIB_CFLAGS)`, our compiler dependent wrapper for getting PIC code) seems to work. However, it seems like `MPI_CFLAGS` overrides any auto-detected defaults.

bonachea:
@ronawho - Correct: CFLAGS and CXXFLAGS default to empty, so you are not overwriting anything by setting those. MPI_CFLAGS default to some autodetected flags (to aid interoperability with the C compiler under the assumption they are the same compiler), so if you want to preserve that intelligence you can instead append the flag to MPI_CC, eg `MPI_CC="${MPI_CC:-mpicc} -fPIC"`

PR #10179, opened 2018-07-06
Add CFLAGS to the massivethreads build

ronawho:
Previously we weren't passing our CLFAGS to massivethreads. Add them here (and
also set CXXFLAGS)

This also updates start_test to avoid throwing `--static` for `--performance`
when massivethreads is being used since is requires several .sos

Related to https://github.com/chapel-lang/chapel/pull/10175


ronawho:
@ben-albrecht quick sanity check on the start_test changes?

Note that I've changed what was:

```py
(chpl_compiler.get('target').startswith("cray-prgenv") and os.getenv("CRAY_MEMKIND", "") != "")
```

to just:
```py
'CRAY_MEMKIND' in os.environ
```

on the basis that an environment variable named `CRAY_MEMKIND` is already enough of a sentinel that we're running on a cray.

ronawho:
>> on the basis that an environment variable named CRAY_MEMKIND is already enough of a sentinel that we're running on a cray.

> I'd prefer to stay consistent with how we determine what kind of system we're running on, if possible. Do you think that's possible here?

I don't understand what this means. The question we need to answer is if we're on a cray linking against memkind. We have to check memkind, and I'm arguing since the environment variable is prefixed with `CRAY_` then we don't need to also check that we're using a cray prgenv

ronawho:
@ben-albrecht convinced me offline to re-add the prgenv check. What we really want to check is whether memkind will be linked in, and if you're using a naked compiler it won't get linked in just because the `cray-memkind` module is loaded, that'll only happen if a prgenv wrapper is loaded

util/start_test
```diff
@@ -688,14 +688,17 @@ def set_up_environment():
...
         # - any locale model other than 'flat' (because we'll need
         #   libnuma which is only available as a .so)
-        if (not (tgt_platform == 'darwin' or
-                 (chpl_compiler.get('target').startswith("cray-prgenv") and
-                  os.getenv("CRAY_MEMKIND", "") != "") or
-                 chpl_locale_model.get() != 'flat')):
+        # - massivethreads (because it depends on various .sos)
+        darwin = tgt_platform == 'darwin'
+        prgenv = compiler_utils.compiler_is_prgenv(chpl_compiler.get('target'))
+        cray_memkind = prgenv and 'CRAY_MEMKIND' in os.environ
```
ben-albrecht: LGTM

ronawho:
@gbtitus I'm going to merge this for use in #10175. The change is simple and the hellos work for massivethreads still, but I'd appreciate a post-commit sanity check

gbtitus:
Looks good.  I decided not to snark on the 214-character line.  :-)

PR #10180, opened 2018-07-06
Fix three bugs in IO.readline(), and add tests

### LINKED ISSUE #10134: Should IO.readline() return false for the last line of a file if it's not newline-terminated?
### Summary of Problem

On success `IO.readline()` returns true and reports the number of
bytes read through `out numRead`.  Except, as noted in #9629/#9637, if
the last line of the file doesn't end with a newline, it returns false
(still reporting the number of bytes read).

This doesn't seem right, and seems a somewhat unintuitive interface to
program to.  Should it return true in this case?

cassella:
- If the last line being read doesn't end with a newline, return true
  anyway.
- Don't store the `EEOF` error from `qio_channel_read_byte()` into the
  `args[]` argument, or count it in `out numRead`.
- Don't use the value of the last byte read for control-flow purposes.
  This fixes the return value if the last byte read was `\0`.

Add tests of these behaviors.

This wraps up and resolves #10134

- [x] full local testing


cassella:
Tested `test/io`, `test/library/standard/IO`, and a few other tests in `test/library/standard` that call readline()

I added the new test in `test/library/standard/IO/readline`, but it could go somewhere in `test/io` if that'd be better.

PR #10186, opened 2018-07-06
Switch perf playground to measure cstdlib atomics with gcc 8.1

ronawho:
We still see the same performance regressions with gcc 6.3 that we saw with gcc
6.2. Swap to 8.1 to see if things are better there.


ronawho:
@dmk42 just fyi again

Here are some of the graphs with regressions: https://chapel-lang.org/perf/chapcs/cstdlib-atomics/?graphs=minimdljsize10time,nbodyvariations

PR #10188, opened 2018-07-06
Fix how we detect if re2 was built with valgrind support

ronawho:
This fixes and cleans up our detection of whether re2 was built with valgrind.
Previously start_test relied `CHPL_RE2_VALGRIND_SUPPORT` (a build time var)
being set at start_test time. This was lame because it could be wrong, and we
don't propagate CHPL_RE2_VALGRIND_SUPPORT in paratest. Further, the checks in
start_test for whether re2 was being used were really wrong. It just checked if
CHPL_REGEXP was in the environment, which meant we preventing valgrind testing
from working when CHPL_REGEXP=none.

This corrects those issues and several others in the code that checks for
valgrind compatibility. Now we:
 - Create a sentinel file in the install dir when re2 was built with valgrind
   support and use that instead of relying on CHPL_RE2_VALGRIND_SUPPORT being
   set at build and test time.
 - Only check for re2 valgrind support if re2 is actually being used!
 - Have paratest check for re2 valgrind support as well
 - Unify valgrind error messages between start_test/paratest
 - Have start_test check the target CHPL_MEM instead of host
 - Fix python style/indentation in start_test checks


ronawho:
This should resolves issues described in https://github.com/chapel-lang/chapel/commit/71b07766eba924122717eb2f6335e7a33cf8fa40#r29599159 -- where the most egregious one is that start_test failed with`CHPL_REGEXP=none` under --valgrind

ronawho:
To test this I ran `start_test release/examples/hello.chpl --valgrind` and `$CHPL_HOME/util/test/paratest.server -dirs release/examples/patterns/ -valgrind` with the following configs:

| tasks    | mem      | re2 support               | result                                     |
| -------- | -------- | ------------------------- | ------------------------------------------ |
| qthreads | jemalloc | re2 w/o valgrind support  | expected tasks!=fifo error                 |
| fifo     | jemalloc | re2 w/o valgrind support  | expected mem!=cstdlib error                |
| fifo     | cstdlib  | re2 w/o valgrind support  | expected re2 not built with valgrind error |
| fifo     | cstdlib  | re2 w/ valgrind support   | works/runs correctly                       |
| fifo     | cstdlib  | regexp=none, no re2 built | works/runs corectly                        |

PR #10197, opened 2018-07-07
fixpath.py: Explicitly handle unset CHPL_HOME or env vars

### LINKED ISSUE #10196: setchplenv.bash interacts poorly with Ubuntu-style empty $MANPATH when $CHPL_HOME is already set
### Summary of Problem

After sourcing `setchplenv.bash`, I can't view non-Chapel manpages on my
Ubuntu system.

My setup configures an empty `$MANPATH`.  When `$MANPATH` is empty,
`man` uses its own mechanism to find the system man pages.

`setchplenv.bash` leaves `$MANPATH` set to
`/home/fortytwo/src/chapel/man: `.  The trailing space keeps `man`
from using its own mechanism, and it can no longer find system man
pages.

### Steps to Reproduce

```
fortytwo@magrathea:~/src/chapel (master)$ echo $MANPATH

fortytwo@magrathea:~/src/chapel (master)$ manpath
/usr/local/man:/usr/local/share/man:/usr/share/man
```
```
fortytwo@magrathea:~/src/chapel (master)$ . util/setchplenv.bash 
Setting CHPL_HOME to /home/fortytwo/src/chapel
Setting CHPL_HOST_PLATFORM to linux64
Updating PATH to include /home/fortytwo/src/chapel/bin/linux64
                     and /home/fortytwo/src/chapel/util
Updating MANPATH to include /home/fortytwo/src/chapel/man
```
```
fortytwo@magrathea:~/src/chapel (master)$ echo ,$MANPATH,
,/home/fortytwo/src/chapel/man: ,
fortytwo@magrathea:~/src/chapel (master)$ manpath
manpath: warning: $MANPATH set, ignoring /etc/manpath.config
/home/fortytwo/src/chapel/man: 
fortytwo@magrathea:~/src/chapel (master)$ man ls
No manual entry for ls
See 'man 7 undocumented' for help when manual pages are not available.
```

If I manually set `$MANPATH` to remove the space, `man` can find
both the system and Chapel man pages.

```
fortytwo@magrathea:~/src/chapel (master)$ export MANPATH=/home/fortytwo/src/chapel/man:
fortytwo@magrathea:~/src/chapel (master)$ manpath
manpath: warning: $MANPATH set, appending /etc/manpath.config
/home/fortytwo/src/chapel/man:/home/fortytwo/src/chapel/man:/usr/local/man:/usr/local/share/man:/usr/share/man
fortytwo@magrathea:~/src/chapel (master)$ man ls | head -1
LS(1)                                        User Commands                                        LS(1)
```

cassella:
Using a default value of ' ' for path when MANPATH was unset led to
changing the user's MANPATH to include a component of ' ', if
CHPL_HOME was set.  Just return the empty path in this case (the path
is already empty, and we're only supposed to be removing things from
it anyway) instead of trying to run the normal path splitting and
filtering in such a way as to end up a no-op.

Resolves #10196

### DIFF
--- a/util/config/fixpath.py
+++ b/util/config/fixpath.py
@@ -17,20 +17,47 @@ def main(env='PATH', delim=':'):
     Removes path components that begin with $CHPL_HOME, to reduce
     $PATH & $MANPATH pollution
 
+    This is used by the setchplenv.* scripts, and may be called in several
+    situations:
+    1. No Chapel environment settings (new shell)
+    2. Same $CHPL_HOME as last time (re-running setchplenv in same dir)
+    3. Different $CHPL_HOME (cd ../other-chapel-dir).
+    4. $CHPL_HOME is set, but path doesn't include an old one.
+       ($CHPL_HOME was hand-set, now setchplenv is run)
+
+    For case 1, just return the existing environment variable.
+
+    For case 2, return the environment variable without the components
+    that begin with $CHPL_HOME.
+
+    For case 3, setchplenv invokes this script before setting the new
+    $CHPL_HOME.  We still have the old $CHPL_HOME set, so we can remove
+    the old $PATH and $MANPATH entries.  The upshot is we do the same thing
+    as in case 2.
+
+    Case 4 should also be the same as case 2, but we won't remove any
+    components since there should be no components starting with $CHPL_HOME.
+    Mentioned only to avoid reintroducing #10196 when this function is modified.
+
     :env: path environment variable ('PATH' or 'MANPATH')
     :delim: path delimiter (':' or ' ')
     :returns: new path with $CHPL_HOME components removed
     """
 
     # Get environment variables, $(MAN)PATH will be always be ':' delimited
-    path = getenv(env, default=' ')
-    chpl_home = getenv('CHPL_HOME', default=' ')
+    path = getenv(env)
+    if path is None:
+        return ''
+
+    chpl_home = getenv('CHPL_HOME')
 
     # Find ':'s that are not escaped
     pattern = r'(?<!\\)\:'
 
     # Split path into list separated by non-escaped ':'s, and sieve chpl_home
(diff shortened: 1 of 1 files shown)


cassella:
I tested that this fixes the problem for me using bash on my system.

I haven't tested any other shells, or any other systems, or any other initial PATH or MANPATH values.

util/config/fixpath.py
```diff
@@ -23,7 +23,7 @@ def main(env='PATH', delim=':'):
     """
 
     # Get environment variables, $(MAN)PATH will be always be ':' delimited
-    path = getenv(env, default=' ')
+    path = getenv(env, default='')
     chpl_home = getenv('CHPL_HOME', default=' ')
```
ben-albrecht: Should this default to `''` as well?

cassella:
Discussed further with @ben-albrecht - It's a little hard to be sure of correctness in all the cases because the script is using special `' '` values (or now `''`), and handling some cases implicitly due to how those values impact the regex match and string replace.  (Discussed some in the linked issue.)  I'm going to try to make it more explicit, and hopefully easier to reason about.

cassella:
I replaced the original commit altogether.  This one explicitly handles the case that the env var is empty.

I wanted to describe what was going on in a comment, but I think I wrote ten times too much.  But I can't think how to describe what's going on any more tersely.

I did some testing of the four cases I mentioned in bash and tcsh.

mppf:
@ben-albrecht - I've been looking at open PRs to see if we can nudge them forward. I think the next step on this one is more review, either from you or somebody else you can nominate? It doesn't seem particularly large or challenging... Can you help nudge it forward?

ben-albrecht:
The code changes look fine to me - the remaining step is to have someone set aside the time to test these changes out on all of the supported shells for the typical use-case, and the edge case this PR sets out to fix.

PR #10198, opened 2018-07-07
Fix up some string manipulation in codegen.cpp setupDefaultFilenames()

ronawho:
Fix misc string copy/manipulation bugs found by gcc 8:
 - fix incorrect size of strncpy "lib" (and just use strcpy since it's a literal)
 - fix incorrect size of strncat (and remove dupe null termination)
 - fix incorrect buffer size of libmodeHeadername strncpy
 - null terminate libmodeHeadername

### DIFF
--- a/compiler/codegen/codegen.cpp
+++ b/compiler/codegen/codegen.cpp
@@ -2191,9 +2191,9 @@ static void setupDefaultFilenames() {
       if (strlen(lastSlash) >= sizeof(executableFilename) - 3) {
         INT_FATAL("input filename exceeds executable filename buffer size");
       }
-      strncpy(executableFilename, "lib", 3);
-      strncat(executableFilename, lastSlash, sizeof(executableFilename)-4);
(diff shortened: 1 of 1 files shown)


ronawho:
Warnings from gcc 8 were: 
```sh
codegen.cpp: In function ‘void setupDefaultFilenames()’:
codegen.cpp:2221:52: error: argument to ‘sizeof’ in ‘char* strncpy(char*, const char*, size_t)’ call is the same expression as the source; did you mean to use the size of the destination? [-Werror=sizeof-pointer-memaccess]
     strncpy(libmodeHeadername, executableFilename, sizeof(executableFilename));
                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
codegen.cpp:2194:14: error: ‘char* strncpy(char*, const char*, size_t)’ output truncated before terminating nul copying 3 bytes from a string of the same length [-Werror=stringop-truncation]
       strncpy(executableFilename, "lib", 3);
```

PR #10199, opened 2018-07-07
Suppress lulesh test on cray-xc ARM platform

awallace-cray:
Suppress lulesh test on cray-xc ARM platform to quiet
nightly test noise from frequent intermittent errors

### DIFF
--- a/test/release/examples/benchmarks/lulesh/lulesh.suppressif
+++ b/test/release/examples/benchmarks/lulesh/lulesh.suppressif
@@ -1,0 +1,11 @@
+#! /usr/bin/env bash
+#
+# This lulesh test does not work well on current aarch64 Cray, so suppress this
+#
+eval `$CHPL_HOME/util/printchplenv --make`
+
+suppress=0
+if [ "$CHPL_MAKE_TARGET_PLATFORM" = cray-xc -a "$CPU" = aarch64 ]; then
+    suppress=1
+fi
+echo $suppress


awallace-cray:
Tested in sandbox.

awallace-cray:
When a suppressed test passes, and this will pass often, you see it in the nightly cron mail under "--- Passing Suppress tests ---".  I'm not sure if that's a bad thing.

awallace-cray:
Hi ronawho, would this work better as a skipif? Thanks.

ronawho:
If it's a sporadic failure, we should not just suppress it -- sporadically "passing" a suppression is no better than sporadically failing.

Beyond that, we stuff the chplenv into skipif/supressif/other files (see util/test/testEnv) so there's no need to run printchplenv yourself. Just check the plain (non `MAKE_` env var). e.g.:

```bash
suppress=0
if [ "$CHPL_TARGET_PLATFORM" = cray-xc -a "$CPU" = aarch64 ]; then
    suppress=1
fi
echo $suppress
```

I'm sure sure what `$CPU` is though, presumably it's set in build scripts, but it'd be nice if there was a more portable way to check for aarch64 that so devs wouldn't have to set CPU if they do testing on aarch64 systems

awallace-cray:
Discarding this in favor of https://github.com/chapel-lang/chapel/pull/10201

PR #10200, opened 2018-07-07
Skip LCALS test on Cray-XC ARM platforms w GNU target compiler

awallace-cray:
LCALS fails on Cray-XC ARM platforms w GNU target compiler, so
add another clause to the existing skipif file. To do this, however,
I have to use the executable form of the skipif file, not the simple
form.


awallace-cray:
Tested on Cray XC, x86_64 and ARM

ronawho:
Let's close this in favor of actually investigating the failure in https://github.com/chapel-lang/chapel/issues/10233

PR #10213, opened 2018-07-09
Add UTF-8 support for the ctype methods

dmk42:
This change adds UTF-8 support for the ctype methods, `isAlpha()` et al.

This change does not attempt to go much beyond the ctype methods.  For example, `split()` will need to be addressed separately.

Tested in UTF-8 and non-UTF-8 environments.  `start_test` is modified to test in a UTF-8 environment now.  To avoid affecting sorting, `LC_COLLATE` is set to `C`.


dmk42:

modules/internal/String.chpl
```diff
@@ -1337,9 +1470,18 @@ module String {
...
+      var codepoint: int(32);
+      var nbytes: c_int;
+      var multibytes = result.buff: c_string;
+      var maxbytes = result.len: ssize_t;
+      if maxbytes > 4 then
+        maxbytes = 4;
+      qio_decode_char_buf(codepoint, nbytes, multibytes, maxbytes);
+      var upCodepoint = codepoint_toUpper(codepoint);
+      if upCodepoint != codepoint {
+        // This assumes that the upper and lower case version of a
```
Kyle-B: Just FYI this doesn't work generally, see http://unicode.org/faq/casemap_charprop.html SpecialCasing.txt linked from there has all the exceptions ( ftp://ftp.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt )

dmk42: Could you elaborate?  I don't understand what doesn't work.

modules/internal/String.chpl
```diff
@@ -430,31 +433,62 @@ module String {
...
+    iter uchars(): int(32) {
+      var localThis: string = this.localize();
+
+      var i = 0;
+      while i < localThis.len {
+        var codepoint: int(32);
+        var nbytes: c_int;
+        var multibytes = (localThis.buff + i): c_string;
+        var maxbytes = (localThis.len - i): ssize_t;
+        if maxbytes > 4 then
```
mppf: Does it help anything to have this conditional here? I would have expected qio_decode_char_buf to handle counts > 4. Does it not?

modules/internal/String.chpl
Kyle-B: Case operations on Unicode strings may cause the length of the string to change, and there is not necessarily a 1:1 mapping of codepoints. Like http://unicode.org/faq/casemap_charprop.html#4 mentions, title casing is a bit special as well, and doing it naively will not always be correct.

modules/internal/String.chpl
```diff
@@ -430,31 +433,62 @@ module String {
...
+      }
+    }
+
     /*
       Index into a string
 
-      :returns: A string with the character at the specified index from
-                `1..string.length`
+      :returns: A string with the multibyte character at the specified index
+                from `1..string.length`
```
mppf: This change makes `i` in this function a *byte offset* which is what I think it should be but I'm not sure if other people want to argue about it?  In any case I think it needs to be clear in the docs comment.

Also, what happens if you index halfway through a multibyte character? Is there a test of that? I'd hope you'd get the replacement character `0xfffd` or maybe an error.

modules/internal/String.chpl
```diff
@@ -1084,19 +1126,29 @@ module String {
...
-          if byte_isUpper(b) {
-            result = false;
+        var locale_result = false;
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: I'd like to leave the maxbytes > 4 business to qio_decode_char_buf

modules/internal/String.chpl
```diff
@@ -1061,16 +1095,24 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if byte_isLower(b) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here

modules/internal/String.chpl
```diff
@@ -1114,12 +1166,20 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if !(byte_isWhitespace(b)) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here

modules/internal/String.chpl
```diff
@@ -1137,12 +1197,20 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if !byte_isAlpha(b) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here

modules/internal/String.chpl
```diff
@@ -1160,12 +1228,20 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if !byte_isDigit(b) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here

modules/internal/String.chpl
```diff
@@ -1183,21 +1259,27 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if !(byte_isAlpha(b) || byte_isDigit(b)) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here

modules/internal/String.chpl
dmk42: Ah, I think I see what you're saying.  One upper case codepoint might map to a sequence of lower case codepoints, for example.  I do not attempt to handle that case.  This change uses the underlying C library, which has no way to handle one-to-many mappings.

modules/internal/String.chpl
```diff
@@ -1208,12 +1290,20 @@ module String {
...
-        for i in 0..#this.len {
-          const char = buff[i];
-          if char <= 0x1f  || char == 0x7f {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: here... Now that I'm mentioning it.. why don't these use the string's codepoint iterator uchars() ?

modules/internal/String.chpl
```diff
@@ -1234,17 +1324,24 @@ module String {
...
-        for i in 0..#this.len {
-          const b = buff[i];
-          if byte_isLower(b) {
+        var i = 0;
+        while i < this.len {
+          var codepoint: int(32);
+          var nbytes: c_int;
+          var multibytes = (this.buff + i): c_string;
+          var maxbytes = (this.len - i): ssize_t;
+          if maxbytes > 4 then
```
mppf: there are a lot of these

modules/internal/String.chpl
```diff
@@ -1268,12 +1366,22 @@ module String {
...
-        if byte_isUpper(b) {
-          // We can just add or subtract 0x20 to change between upper and lower
-          result.buff[i] = b + 0x20;
+      var i = 0;
+      while i < result.len {
+        var codepoint: int(32);
+        var nbytes: c_int;
+        var multibytes = (result.buff + i): c_string;
+        var maxbytes = (result.len - i): ssize_t;
+        if maxbytes > 4 then
```
mppf: still 4

modules/internal/String.chpl
dmk42: Right, checking the code in qio_decode_char_buf, it does handle the larger counts, so I will delete those tests.

modules/internal/String.chpl
dmk42: I'm updating the comment to be more precise.

If you index part way into a multibyte character, it gives you the replacement character.

modules/internal/String.chpl
Kyle-B: Yup, I was just pointing out that this not the correct behavior for Unicode strings. I'd strongly recommend looking into using something like https://github.com/JuliaStrings/utf8proc for the heavy lifting in these functions.

modules/internal/String.chpl
dmk42: Good point.  Most of this code evolved independently of the iterator.  Now that I have them in the same branch together, I can use `uchars()`.  That makes the code easier to read.

Issue #9839, opened 2018-06-12
Convert the Barriers module over to error-handling


ronawho:
Remove halts from the Barriers module, converting them to throw'ing routines where appropriate.

ronawho:
I don't think any of the remaining cases actually want error-handling -- they are either pure virtual functions, an exhaustive select that shouldn't need a halt, or a sanity check on the number of callers

PR #10220, opened 2018-07-09
Remove remaining calls to `halt()` in Barriers.chpl

ronawho:
This just converts halts to halt wrappers where appropriate:
 - use pureVirtualMethodHalt() in abstract base class
 - use exhaustiveSelectHalt() in an exhaustive enum select stmt
 - convert "too many caller" halts to be boundsCheckHalts()

Closes https://github.com/chapel-lang/chapel/issues/9839

### DIFF
--- a/modules/standard/Barriers.chpl
+++ b/modules/standard/Barriers.chpl
@@ -100,7 +100,8 @@ module Barriers {
           }
         }
         otherwise {
-          halt("unknown barrier type");
+          use ChapelHaltWrappers;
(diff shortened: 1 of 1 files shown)


ronawho:
@daviditen the 3rd commit that converts "too many caller" halts into `boundsCheckHalts()` is probably the most questionable part of this PR. I'd be open to turning those into asserts or something, but I don't think it makes sense to convert them over to error-handling, especially since they're racy checks.

Brad ok'd converting them into bounds checks, but you may feel differently

PR #10222, opened 2018-07-09
Introduce custom struct in sys.h to address llvm bug

Spartee:
- created `sys_stat_s`(sys_stat_t), a typedef struct in `sys.h` to hold the values for a stat struct without the macros that caused an error when implementing build on last modified in Mason.
- `stat_to_sys_stat`, a function to convert a stat struct in a sys_stat_s struct
- changes to `sys_stat` were also made so that the struct passed into the function is of type `sys_stat_t`
- A test in the extern folder to test the functionality


test/extern/spartee/sys-stat-extern.good
```diff
@@ -0,0 +1 @@
+1531167971
```
cassella: Git doesn't preserve file modification times, so the last modified time of this file will keep changing.  (For example if you checkout master and then check out this branch again, I bet the test will fail.)

Even if it did, the .good file would need updating every time someone modified the .chpl file.  For example, to add a newline at the end.


You could try using `touch -d` to create a file with a known modification time.  Though I'm not sure that's portable to everywhere Chapel runs.  Or what lengths you'd have to go to to be sure it'll result in the same output no matter what timezone the test is run in.

runtime/include/qio/sys.h
```diff
@@ -65,6 +65,24 @@ typedef struct sys_statfs_s {
...
+  ino_t       st_ino;       /* File serial number. */
+  mode_t      st_mode;      /* Mode of file (see below). */
+  nlink_t     st_nlink;     /* Number of hard links to the file. */
+  uid_t       st_uid;       /* User ID of file. */
+  gid_t       st_gid;       /* Group ID of file.  */
+  dev_t       st_rdev;      /* Device ID (if file is character or block special). */
+  off_t       st_size;      /* File Size in bytes */
+  off_t       st_atimespec; /* Last data access timestamp.  */
+  off_t       st_mtimespec; /* Last data modification timestamp. */
+  off_t       st_ctimespec; /* Last file status change timestamp. */
```
mppf: I was expecting st_atim, st_mtim, st_ctim fields of type `struct timespec` per the newer POSIX: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html

If you have these, we can make paretheses-less methods in Chapel that allow access to st_ctime etc. I also don't know why you named them ending in 'timespec'.

runtime/src/qio/sys.c
```diff
@@ -547,12 +547,34 @@ err_t sys_lseek(fd_t fd, off_t offset, int whence, off_t* offset_out)
   return err_out;
 }
 
-err_t sys_stat(const char* path, struct stat* buf)
+
+void stat_to_sys_stat(const char* path, sys_stat_t* out_buf, struct stat* in_buf)
+{
+  stat(path, in_buf);
+  out_buf -> st_dev = in_buf -> st_dev;
```
mppf: I don't think I've seen any code (in the Chapel codebase) that uses spaces around the `->` like this, so probably better to remove the spaces around the `->`.

runtime/src/qio/sys.c
```diff
@@ -547,12 +547,34 @@ err_t sys_lseek(fd_t fd, off_t offset, int whence, off_t* offset_out)
...
+  out_buf -> st_nlink = in_buf -> st_nlink;
+  out_buf -> st_uid = in_buf -> st_uid;
+  out_buf -> st_gid = in_buf -> st_gid;
+  out_buf -> st_rdev = in_buf -> st_rdev;
+  out_buf -> st_size = in_buf -> st_size;
+  out_buf -> st_blksize = in_buf -> st_blksize;
+  out_buf -> st_blocks = in_buf -> st_blocks;
+  out_buf -> st_mtimespec = in_buf -> st_mtime;
+  out_buf -> st_atimespec = in_buf -> st_atime;
+  out_buf -> st_ctimespec = in_buf -> st_ctime;
```
mppf: I think it'd be nice to use the `struct timespec` versions of these fields when they are available.

test/extern/spartee/sys-stat-extern.chpl
```diff
@@ -0,0 +1,18 @@
...
+  var file_buf: chpl_stat;
+  var file_path = filename.c_str();
+
+  if (sys_stat(file_path, file_buf) == 0) {
+    return file_buf.st_mtimespec;
+    }
+  return -1;
+}
+
+writeln(getLastModified("sys-stat-extern.chpl"));
```
mppf: Add newline at end of file so we don't have the red zero thingy

test/extern/spartee/sys-stat-extern.good
mppf: I agree with @cassella - there'll have to be some other strategy for this to work out. The most obvious one is to just not print the time result. Another direction would be to create a file and then check its last modified time (with some tolerance).

runtime/src/qio/sys.c
```diff
@@ -547,12 +547,51 @@ err_t sys_lseek(fd_t fd, off_t offset, int whence, off_t* offset_out)
...
+  out_buf->st_rdev = in_buf->st_rdev;
+  out_buf->st_size = in_buf->st_size;
+  out_buf->st_blksize = in_buf->st_blksize;
+  out_buf->st_blocks = in_buf->st_blocks;
+
+#if defined(_DARWIN_C_SOURCE)
+  out_buf->st_atim = in_buf->st_atimespec;
+  out_buf->st_mtim = in_buf->st_mtimespec;
+  out_buf->st_ctim = in_buf->st_ctimespec;
+#elif (_POSIX_C_SOURCE >= 200112L)
```
mppf: I'm pretty sure this should be using _POSIX_VERSION

runtime/src/qio/sys.c
```diff
@@ -547,12 +547,51 @@ err_t sys_lseek(fd_t fd, off_t offset, int whence, off_t* offset_out)
...
+  out_buf->st_mode = in_buf->st_mode;
+  out_buf->st_nlink = in_buf->st_nlink;
+  out_buf->st_uid = in_buf->st_uid;
+  out_buf->st_gid = in_buf->st_gid;
+  out_buf->st_rdev = in_buf->st_rdev;
+  out_buf->st_size = in_buf->st_size;
+  out_buf->st_blksize = in_buf->st_blksize;
+  out_buf->st_blocks = in_buf->st_blocks;
+
+#if defined(_DARWIN_C_SOURCE)
```
mppf: Pretty sure this should be #ifdef __APPLE__

runtime/src/qio/sys.c
```diff
@@ -547,12 +547,47 @@ err_t sys_lseek(fd_t fd, off_t offset, int whence, off_t* offset_out)
...
+  out_buf->st_mode = in_buf->st_mode;
+  out_buf->st_nlink = in_buf->st_nlink;
+  out_buf->st_uid = in_buf->st_uid;
+  out_buf->st_gid = in_buf->st_gid;
+  out_buf->st_rdev = in_buf->st_rdev;
+  out_buf->st_size = in_buf->st_size;
+  out_buf->st_blksize = in_buf->st_blksize;
+  out_buf->st_blocks = in_buf->st_blocks;
+
+#if (_POSIX_C_SOURCE > 200112L)
```
mppf: I think this should be _POSIX_VERSION

mppf:
@dmk42 - it seems to me that (almost) every C codebase I've ever seen uses `typedef struct bla { } bla_t` so it's hardly the case that we're the only ones violating this rule.

PR #10226, opened 2018-07-09
Improve resolution of PRIM_SET_MEMBER

vasslitvinov:
* Allow resolveFieldSymbol() to work on an already-resolved
PRIM_SET_MEMBER. While there, refactor the code for clarity.

* Fix removeRandomPrimitive() as follows.
removeRandomPrimitive() removes certain PRIM_SET_MEMBER calls.
Due to #10132, the calls that already included a correct field Symbol
were sheltered from such removal. This PR un-shelters them.


vasslitvinov:
@benharsh would you review?

Issue #10184, opened 2018-07-06
Can't declare generic return type, even if only instantiations passed to it


lydia-duncan:
### Summary of Problem
If you declare a function to have a return type that is generic, the return statement generates a type mismatch, even for instantiations from that generic type.  I think it would be reasonable to support this (and we already do if the return type is not declared at all), but it may require some rejiggering of our generated functions.

### Steps to Reproduce

**Source Code:**

```chapel
class Gen {
  type t;
}

proc retsGen(type t): Gen {
  return new unmanaged Gen(t); // removing the `unmanaged` from this does not change behavior.
}

var x = retsGen(int);
var y = retsGen(real);
writeln(x.type: string);
writeln(y.type: string);
delete x, y;
```

This code gives us the message:
``` 
declaredGenericReturnType.chpl:5: In function 'retsGen':
declaredGenericReturnType.chpl:6: error: type mismatch in assignment from Gen(int(64)) to Gen
```

**Compile command:**
`chpl declaredGenericReturnType.chpl`

**Associated Future Test(s):**
[`test/functions/lydia/declaredGenericReturnType.chpl`](https://github.com/chapel-lang/chapel/blob/master/test/functions/lydia/declaredGenericReturnType.chpl) #10185

### Configuration Information

- Output of `chpl --version`: 1.18.0 (pre-release)

lydia-duncan:
I believe this is the same issue I am encountering after the parsing changes for supporting declared array return types with unspecified domains.  At least, the call where the failure is occurring has the same AST structure.

lydia-duncan:
Discussing with Michael, it seems like the array issue can be better solved a different way (though this is an issue that should get addressed at some point)

PR #10191, opened 2018-07-06
Add halt wrappers for runtime semantics checks

ronawho:
Add halt wrappers for execution time semantic checks that can be disabled at
compilation time with --fast/--no-checks/--no-<check>-checks:

    nilCheckHalt()          for --nil-checks
    boundsCheckHalt()       for --bounds-checks
    formalDomainCheckHalt() for --formal-domain-checks
    localCheckHalt()        for --local-checks
    safeCastCheckHalt()     for --cast-checks
    divByZeroCheckHalt()    for --div-by-zero

PR #10241, opened 2018-07-10
Allow array return types with unspecified domains or element type

### LINKED ISSUE #8468: Support assumed-domain array return types

lydia-duncan:
Resolves #8468 

Prior to this change, an array return type had to be either completely
inferred from the contents of the function, or completely specified
(including both the domain and the element type of the array).  With
this change, declarations such as`proc foo(): [ ] int {`,
`proc foo(): [1..3] {` and `proc foo(): [ ] {` are now supported.

I ran into problems with `proc foo(): [x in ...] {` due to shift/reduce
conflicts at parse time, which caused me to remove support for it
altogether (though I don't think anyone was using it, since we got that
support entirely due to sharing the array type rule with variable declarations)

Note: this doesn't support generic array declared yields just yet, though
it may not be too hard to extend to support that.

Details:
Adds new rules to the parser now that the array declarations supported
by return types are different from both those on formals and those on
ordinary variables.

At normalization, modify our handling of returns when the array is not
fully declared.  This involved adding some checks depending on if the
domain or element type is provided, and removing the return type since
we don't know what to do with it when it is generic (see #10184).  The
checks are implemented in module code.

Adds several new tests of this functionality, and resolves a few futures.
New tests:
- error message when domain returned does not match the declared
domain (execution time, since domains are runtime types)
- error message when element type does not match the declared
element type (compile time)
- a future for allowing coercion on these returns, see #10242 
- allowing both a different domain and a different element type to be
returned if done using a param conditional
- allowing multiple return statements when promotion is involved (and
the domain is generic)
- allowing multiple return statements when the domain is generic
- allowing a declared no element type, no explicit domain
- allowing a return when the domain is specified but the element type is
not

Fixes:
- test/arrays/return/returnArbitraryArray
- test/compflags/lydia/library/exportArray/returningArr/callFuncReturnsArrayTakesArg
- test/compflags/lydia/library/exportArray/returningArr/callFuncReturnsArrayTakesArrayArg
- test/compflags/lydia/library/exportArray/returningArr/callFuncReturnsBadArray
- test/compflags/lydia/library/exportArray/returningArr/returnBadArray-multiRank
- test/compflags/lydia/library/exportArray/returningArr/returnBadArray-strided
- test/compflags/lydia/library/exportArray/returningArr/returnBadArray-unusualIndices

Tested with a full std paratest with futures and a memleaks run over the
test/arrays/return directory


lydia-duncan:
@mppf - this is ready for review, when you get the chance

compiler/parser/chapel.ypp
```diff
@@ -1367,6 +1368,56 @@ opt_init_expr:
...
+    { $$ = new CallExpr("chpl__buildArrayRuntimeType",
+             new CallExpr("chpl__ensureDomainExpr", $2), NULL);
+    }
+| TLSBR TRSBR ret_array_type
+    { $$ = new CallExpr("chpl__buildArrayRuntimeType", gNil, $3); }
+| TLSBR expr_ls TRSBR ret_array_type
+    { $$ = new CallExpr("chpl__buildArrayRuntimeType",
+             new CallExpr("chpl__ensureDomainExpr", $2), $4);
+    }
+| TLSBR expr_ls TIN expr TRSBR type_level_expr
```
mppf: I thought the PR message said this case wasn't working due to parser conflicts?

compiler/passes/normalize.cpp
```diff
@@ -1274,7 +1282,7 @@ static void normalizeYields(FnSymbol* fn) {
       retval->addFlag(FLAG_YVV);
 
       yield->insertBefore(new DefExpr(retval));
-      insertRetMove(fn, retval, yield);
+      insertRetMove(fn, retval, yield, false);
```
mppf: Maybe your PR message should say something about not supporting generic array yield types yet?

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,163 @@ static bool isVoidReturn(CallExpr* call) {
   return retval;
 }
 
-static void insertRetMove(FnSymbol* fn, VarSymbol* retval, CallExpr* ret) {
+static bool hasGenericArrayReturn(FnSymbol* fn) {
+  if (returnsArray(fn)) {
+    BlockStmt* typeExpr = fn->retExprType;
+
+    CallExpr* call = toCallExpr(typeExpr->body.tail);
```
mppf: What is this call supposed to be ? I'd INT_ASSERT it's the right thing or at least put a comment about what you are expecting

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,163 @@ static bool isVoidReturn(CallExpr* call) {
...
+  }
+
+  return false;
+}
+
+static void insertDomainCheck(Expr* actualRet, CallExpr* retVar,
+                              Expr* domExpr) {
+  VarSymbol* unrefTemp = newTemp();
+  retVar->insertBefore(new DefExpr(unrefTemp));
+  CallExpr* makeArray = new CallExpr(PRIM_MOVE, unrefTemp,
```
mppf: we talked about it, so I know what this chpl__unref call is doing, but I believe it would be mysterious to somebody looking at this code in the future. Add a comment?

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,163 @@ static bool isVoidReturn(CallExpr* call) {
...
+  retVar->insertBefore(new DefExpr(cond));
+
+  VarSymbol* checkVar = newTemp();
+  retVar->insertBefore(new DefExpr(checkVar));
+  retVar->insertBefore(new CallExpr(PRIM_MOVE, checkVar, checkDom));
+  retVar->insertBefore(new CallExpr(PRIM_MOVE, cond,
+                                    new CallExpr("_cond_test", checkVar)));
+
+  Symbol* msg = new_StringSymbol("domain mismatch on return");
+  BlockStmt* failureMode = new BlockStmt(new CallExpr("halt",
```
mppf: I think it would simplify this code if you just call a new module routine like `chpl_check_domains_match` or something. That routine should check if bounds checking is on (ignoring the error if it's not on) and should ideally halt using one of the methods Elliot added in #10191 - boundsCheckHalt().

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,163 @@ static bool isVoidReturn(CallExpr* call) {
...
+  cond->addFlag(FLAG_MAYBE_PARAM);
+
+  retVar->insertBefore(new DefExpr(cond));
+
+  VarSymbol* checkVar = newTemp();
+  checkVar->addFlag(FLAG_MAYBE_PARAM);
+  retVar->insertBefore(new DefExpr(checkVar));
+  retVar->insertBefore(new CallExpr(PRIM_MOVE, checkVar, checkEltType));
+  retVar->insertBefore(new CallExpr(PRIM_MOVE, cond,
+                                    new CallExpr("_cond_test", checkVar)));
```
mppf: Once again I think it'd really improve this code to call a modules routine like `chpl_check_array_eltTypes_match` and have that issue the `compilerError` if necessary.

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,163 @@ static bool isVoidReturn(CallExpr* call) {
...
+  if (!noDom) {
+    // Add checks against the declared domain
+    insertDomainCheck(retExpr, ret, domExpr);
+  }
+
+  if (retEltExpr != NULL) {
+    insertElementTypeCheck(retEltExpr, retExpr, ret);
+  }
+
+  // Do something about coercion
```
mppf: Maybe put TODO next to this comment?

test/arrays/return/returnArbitraryBadEltType.good
```diff
@@ -0,0 +1,2 @@
+returnArbitraryBadEltType.chpl:1: In function 'inc':
+returnArbitraryBadEltType.chpl:2: error: type mismatch in assignment from real(64) to int(64)
```
mppf: Since you're making up the error message at this point I think it'd be nice to say something more like `array element type mismatch in return` (and you can still put the types if you want).

compiler/parser/chapel.ypp
lydia-duncan: Nah, this is the case that "works".  The parser barked when the `type_level_expr` part was elided.  That said, I didn't add a test for this, since it was functionality that got grandfathered over from when we were using the same array type stuff as variable declarations

lydia-duncan:
@mppf - I think this is ready for another look.  I'm a little concerned that having the unref call in the helper functions changes the behavior but got around it by having it in the code I insert at compilation.  I guess it makes sense to have it at the same level as the original function, since it impacts what is returned (and we wouldn't insert it otherwise), but it still seems a little odd.

compiler/parser/chapel.ypp
mppf: I'd prefer we either leave this case out of the parser or add a test for it, if you havn't already

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,106 @@ static bool isVoidReturn(CallExpr* call) {
...
+                              Expr* domExpr) {
+  // We are returning an array in this function.  In the case where a
+  // promoted expression is returned, we need to call chpl__unref, which
+  // modifies its argument.  Then we pass that into the check function.
+  // LYDIA NOTE: If the unref call is performed in the check function, we don't
+  // return it correctly from the original function.  This is vaguely
+  // terrifying, but I am not sure what to do about it.
+  VarSymbol* unrefTemp = newTemp();
+  retVar->insertBefore(new DefExpr(unrefTemp));
+  CallExpr* makeArray = new CallExpr(PRIM_MOVE, unrefTemp,
```
mppf: Does this add a memory leak?

compiler/passes/normalize.cpp
```diff
@@ -1293,14 +1301,106 @@ static bool isVoidReturn(CallExpr* call) {
...
+// Validates the declared domain, if it exists.  Also, ensures the value is
+// returned appropriately
+static void insertDomainCheck(Expr* actualRet, CallExpr* retVar,
+                              Expr* domExpr) {
+  // We are returning an array in this function.  In the case where a
+  // promoted expression is returned, we need to call chpl__unref, which
+  // modifies its argument.  Then we pass that into the check function.
+  // LYDIA NOTE: If the unref call is performed in the check function, we don't
+  // return it correctly from the original function.  This is vaguely
+  // terrifying, but I am not sure what to do about it.
```
mppf: I don't understand what the problem is here. Don't return what from the original function? I might be able to help further with more details.

Anyway it's an issue but it doesn't necessarily need to hold up this PR.

lydia-duncan:
Passed a full paratest again, going to add the parser changes and merge

lydia-duncan:
Also, I double checked memory leaks over the test/arrays/return directory (where the original future was)

bradcray:
Thanks for jumping on this one, Lydia!

Issue #5749, opened 2017-03-20
Improve code that limits the size of qthreads memory pools


ronawho:
I setup code to limit the size of qthreads memory pools in #257 and #1649. At the time I thought there was a single global pool for task stacks, but now I believe there's a task-stack pool per worker. 

With that in mind, I'm limiting the max size in #5748 to avoid using outrageous amounts of memory on high cpu count systems (like KNL)

The comments around the limiting code could use some cleanup, and I'd like some verification from the qthreads team that pooling is indeed done on a per-worker basis.

This is just a reminder to cleanup that code and talk to the qthreads team.

gbtitus:
If there really is a pool per Qthreads worker then it seems like we could reduce the pool size quite a bit without hurting performance in the usual case in which we don't oversubscribe the workers with tasks/qthreads.  It seems like enough room for 2 stacks (including guard pages, if appropriate) per pool would be enough, maybe even 1 stack per pool.

Relatedly: we should consider reducing our stack size by the amount needed for guards, if any, plus whatever other overhead is needed for the pooling. We could reduce some of our current waste if the pools fit evenly into the large jemalloc chunk size used when we have a registered heap on hugepages for Cray systems. (Current max-sized 65m pools waste hugepageSize-1m bytes per pool, since the memory is acquired in full hugepages.)

ronawho:
Yeah, I was thinking about that recently too. I was wondering if we could have every worker grab enough room for 2 stacks initially, and maybe have future extensions grab larger sizes. i.e. grab room for 2 stacks on init, and if more are needed later, grab larger chunks. 

Today, I think we'd want room for 2 task stacks -- because we do usually end up with 2 tasks running on "worker 0". A `coforall in 1..here.maxTaskPar` will create maxTaskPar tasks, in addition to the 1 running main, so we'll end up with 2 tasks there.

I'm also not super familiar with how qthreads pooling works. I don't know how they "extend" pools when a pool is out of memory. I'm assuming if the pool is out of memory, they just allocate a new pool, and somehow link them together, but I'm not sure if there is a cost for a "pool miss"

PR #10254, opened 2018-07-10
Reduce hugepage waste from call stacks

ronawho:
Previously, we defaulted to 8 MiB stacks, and 16 MiB hugepages. However, in
addition to 8 MiB for the actual stack qthreads also allocated space for a
runtime data structure `qthread_runtime_data_s`. This meant that 2 stacks
actually used 2 16 MiB hugepages, with 1 mostly wasted.

This lowers the actual stack space by 1 system page, so that we can fit the
entire "stack" inside the 8 MiB envelope. If guard pages are enabled (not
possible with hugepages) we'll use an additional 2 pages.

This also lowers the amount of pooling from ~8 default sized stacks to 2
stacks to limit the amount of memory used for "most" programs. Most of our
benchmarks tend to only create #cores tasks, so on systems that require
registration (and faulting in of physical memory to support the pages) we'd
take up lot of space that often isn't used.

Closes https://github.com/chapel-lang/chapel/issues/5749

### DIFF
--- a/third-party/qthread/Makefile
+++ b/third-party/qthread/Makefile
@@ -67,7 +67,9 @@ CHPL_QTHREAD_CFG_OPTIONS += --enable-oversubscription
 endif
 
 # enable guard pages for stack overflow detection, unless directed not to
+HAVE_GUARD_PAGES = 0
 ifeq (, $(call isTrue, $(CHPL_QTHREAD_NO_GUARD_PAGES)))
+HAVE_GUARD_PAGES = 1
 CHPL_QTHREAD_CFG_OPTIONS += --enable-guard-pages
 endif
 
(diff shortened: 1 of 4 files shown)


ronawho:
TODOs: 
 - [ ] make similar changes for fifo
 - [ ] more testing
 - [ ] probably some code cleanup

third-party/qthread/Makefile
```diff
@@ -62,7 +62,9 @@ CHPL_QTHREAD_CFG_OPTIONS += --enable-oversubscription
 endif
 
 # enable guard pages for stack overflow detection, unless directed not to
+HAVE_GUARD_PAGES = 0
```
gbtitus: This looks backwards to me.

ronawho: How so? Guard pages are disabled in qthreads by default, unless configured with `--enable-guard-pages`

gbtitus: Yeah, sorry, I misread the `ifeq`.  I always initially scan `ifeq (, $(call isTrue, <whatever>))` as meaning "do the if-branch if `<whatever>` is true", when in fact it means the opposite.  Never mind, carry on.

PR #10257, opened 2018-07-10
Build/test scripts should always printchplenv with --all --no-tidy

awallace-cray:


awallace-cray:
I definitely want "--all", to always see CHPL_LLVM. 

"--no-tidy" may add two extra lines to some outputs, but I like the consistency and completeness of always seeing the complete list. Others may disagree. It's optional.

util/start_test
```diff
@@ -1142,7 +1142,7 @@ def print_chapel_environment():
     logger.write()
     logger.write("### Chapel Environment ###")
     try:
-        out = subprocess.check_output(os.path.join(util_dir, "printchplenv"))
+        out = subprocess.check_output(os.path.join(util_dir, "printchplenv"), "--all", "--no-tidy")
```
ben-albrecht: Does anything else consume this output?

awallace-cray: Not that I know of. I didn't change printchplenv's with format options like --simple, figuring those were a little more likely to be used as input somewhere.

awallace-cray:
Did not run developer's paratest, but I ran a whitebox correctness test in sandbox, that uses paratest.

util/start_test
awallace-cray: I looked closer at that testing, and found that line 1145 in start_test needed a fix. 
```
-        out = subprocess.check_output(os.path.join(util_dir, "printchplenv"), "--all", "--no-tidy")
+        out = subprocess.check_output([os.path.join(util_dir, "printchplenv"), "--all", "--no-tidy"])
```

awallace-cray:
Commit 4188f17 is rebased and fixes start_test line 1145

PR #10269, opened 2018-07-11
Add .suppressif files to suppress bounds-checks tests for --fast

daviditen:
These tests are testing bounds checks, so suppress them if bounds checking is
disabled by --fast.

### DIFF
--- a/test/arrays/diten/arrayBackOOB.suppressif
+++ b/test/arrays/diten/arrayBackOOB.suppressif
@@ -1,0 +1,11 @@
+# --fast sets --no-checks
+#
+# --fast sets --no-checks and this test is explicitly checking that one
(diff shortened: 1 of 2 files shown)


lydia-duncan:
Should these be skipped instead of suppressed?  I don't think we expect bounds checking to ever be on for `--fast`

daviditen:
I followed the pattern of several other array OOB tests.  The .suppressif file is copied from `test/arrays/bradc/localSlices/localsliceOOB.suppressif`.  From the description in the .suppressif files themselves:

> There's been debate as to whether tests like this should be suppressed or skipped. For now we suppress with the argument that we're testing that --fast actually turns our checks off.

lydia-duncan:
Okay, cool

PR #10271, opened 2018-07-11
Adjust parameter type to avoid confusing some C compilers

dmk42:
Treat the buffer argument to `qio_encode_char_buf` as a `void *` in Chapel code to avoid confusing some back end compilers about differences between pointers to signed, unsigned, and plain chars.

This quiets some testing noise.


dmk42:
@mppf - Would you please review?

Issue #10002, opened 2018-06-25
Remove remaining halts from the standard modules


ronawho:
Removing the remaining uses of `halt()` from the standard modules, either replacing them with error-handling or halt wrappers where appropriate:

 - [x] remove halts from Barriers.chpl -- #9839 (done in #10220)
 - [x] remove halts from List.chpl -- #9838 (done in #10255)
 - [x] remove halts from Time.chpl -- #9836 (done in #10339)
 - [x] remove halts from Types.chpl -- #9837 (done in #9837)
 
 * [x] remove halts from DynamicIters.chpl (done in #10228 and #10399)
 * [x] remove halts from BigInteger (done in #10272)
 * [x] remove halts from Math.chpl (done in #10391)
 * [x] remove halts from IO.chpl (done in #10346)

PR #10272, opened 2018-07-11
Fix bigint to integral/real casts

ronawho:
For casts to int/uint/real we used to cast to int(64)/uint(64)/real(64) instead
of casting to the type the user specified. This corrects that issue, so we now
cast to the type requested. I also removed calls to safeCast() since these are
regular (non-safe) cast operations.

This also changes the behavior of bigint to uint casts to be the absolute value
truncated to the resultant type instead of halting for negative values. This
matches what the underlying GMP cast does, and is better than halting, but
I opened https://github.com/chapel-lang/chapel/issues/10287 to discuss further.

This also converts a halt in an initializer to an initHalt, since it's a case
where we want error-handling. With that conversation and the removal of the
cast halt, there are no remaining calls to `halt()` in the BigInteger module.

Related to https://github.com/chapel-lang/chapel/issues/10002


ronawho:
Easiest to review the bigint changes if you ignore whitespace

modules/standard/BigInteger.chpl
```diff
@@ -104,6 +104,17 @@ cannot be represented using the GMP scalar type.
...
+Casting from ``bigint`` to ``integral`` and ``real`` types is also
+supported.  Values that are too large for the resultant type are
+truncated.  GMP primitives are used to first cast to platform-specific C
+types, which are then cast to Chapel types.  As a result, casting to
+64-bit types on 32-bit platforms may result in additional truncation.
+Additionally, casting a negative ``bigint`` to a ``uint`` will result in
+the absolute value truncated to fit within the type.::
+
+  var a = new bigint(-1);
+  writeln(a:uint);        // prints "1"
```
mppf: It occurs to me that we could make this print out the same as (-1):uint by doing some bit-complementing for negative numbers. It's unclear to me if this is better though so I might like a 3rd opinion.

ronawho:
> Looks good, but I'd like a 3rd opinion on whether (new bigint(-1)):uint should be the same as (-1):uint or not. I'm certain it shouldn't halt.

pinging @bradcray because of his recent work with GMP for bigint ranges

bradcray:
I don't have a strong opinion about what the negative-bigint-to-uint cast should do.  I find the answer `1` surprising, but arguably pretty much anything is surprising, and I understand the tension between "do what GMP does" and "do what Chapel and C do for [u]ints."  I suppose a reasonable argument for doing what GMP does is "Users of GMP will be less surprised" and if someone really wants the Chapel / C behavior, they could/should cast to `int` first before casting to `uint`?

mppf:
Our bigint is currently a thin wrapper around GMP stuff, but I don't think that's fundamental to the type, so I'd generally prefer we'd do something more similar to the Chapel int->uint cast - on the theory that users of bigint will be more familiar with other parts of Chapel than they are with GMP.

ronawho:
> and if someone really wants the Chapel / C behavior, they could/should cast to int first before casting to uint?

For values that fit inside an int, yes. This all gets screwy if you're converting something less than min(int) to a uint, but in that case I have no idea what you would even want to happen anyways

ronawho:
> I don't have a strong opinion about what the negative-bigint-to-uint cast should do.

I don't really have a strong preference here either. For this I viewed my job as "figure out if we should convert this halt to error-handling", and I think the answer to that is no. Having done that I wanted to get rid of the halt by either converting it to a halt wrapper or by improving the implementation so the halt is no longer needed.

I like this PR because it's simple and just defers to GMP. That said, I'm not opposed to doing extra work so that  bigint->uint matches int->uint, but it's not obvious to me what that code will look like and I imagine we'll have to put some careful thought into boundary cases where truncation will occur. If that's the route we want to go I'd probably change this PR to use a `notImplementedHalt()` wrapper and make a new issue to discuss what bigint->uint casts should do

mppf:
> If that's the route we want to go I'd probably change this PR to use a notImplementedHalt() wrapper and make a new issue to discuss what bigint->uint casts should do

Why don't we merge this PR and also make such an issue?

PR #10278, opened 2018-07-11
Clean up some comments in the HDF5 module

daviditen:
Improve comments for HDF5_Chapel routines.  For example, don't talk about
"output arrays" in comments about a function reading from disk into arrays.
Also, updated a couple variable names to correspond to the comment changes.


daviditen:

PR #10281, opened 2018-07-11
Add compiler error for record of infinite size

### LINKED ISSUE #7714: Better error message for recursive record types
### Summary of Problem
It is illegal to create a record with a field with the same type as itself:
```chpl
record R {
  var x : int;
  var r : R;
}
var r = new R();
```
```
> chpl foo.chpl
foo.chpl:1: error: unable to resolve return type of function '_construct_R'
foo.chpl:1: error: called recursively at this point
```
This is reasonable because the record would require an infinite amount of storage. However, I think this error message could be improved upon.

[`test/types/records/sungeun/recursiveRecord.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/types/records/sungeun/recursiveRecord.chpl
          )

### LINKED ISSUE #10080: Compiler segfault with cycle of records
### Summary of Problem
Compiler segfaults when trying to parse records that are cyclically dependent. There are two variable declarations `x` and `y` in `main` that can or cannot be selectively omitted depending on whether you're using version 1.17.1 or master.

### Steps to Reproduce

**Source Code:**
```chapel
module RecordCycle {
record R1 {
  var a: R3;
}
record R2 {
  var a: R1;
}
record R3 {
  var a: R2;
}
proc main() {
  var x: R3;
  var y: [0..0] R3;
}
} // module RecordCycle
```

### Configuration Information
`chpl version 1.17.1`

benharsh:
This PR adds a compiler error for records of infinite size, like:
```chpl
record R {
  var x : R;
}
```

Due to compiler architecture complications a case involving generic records and type functions has been futurized. This error message only applies to types with initializers since the flow of resolution makes it easier to handle initializers than constructors. The checks for this error are located in what may seem like odd places, but resolution doesn't always give us a chance to properly 'handle' a record before we start doing things with it.

Resolves #7714 
Resolves #10080 

Testing:
- [x] local + futures
- [x] partial gasnet


ian-bertolacci:
Can the error message name one or more of the offending records? That would make debugging easier.

benharsh:
Elsewhere @bradcray suggested text like:
```
Record has infinite size (record 'R' contains field 'x' of type 'R')
Record has infinite size (record 'A' contains field 'b' whose type contains 'A')
```
I would support this change.

vasslitvinov:
I like Brad's suggestion.

To me, the problem is that a recursive record is not well-defined. I was surprised to see the error message talk about infinite size. Maybe instead of "infinite size" the error message could say "recursive", ex.:

```
record 'R' cannot contain a recursive field 'x' of type 'R'
record 'A' cannot contain a recursive field 'b' whose type contains 'A'
```

bradcray:
I have to admit that "infinite size" threw me for a loop initially as well.  In the end, I decided that what I liked about it (compared to other things I could come up with) was that it implied some rationale for _why_ we don't support recursive records rather than just saying "hey, you can't do that."  That said, there's also something attractive about Vass's suggestions in their brevity and clear statement of a rule that's been broken.  I do prefer seeing 'record' uncapitalized as he suggests in any case over my proposal.  I also like that 'A' isn't said redundantly in his suggestions.

I toyed with whether an explanatory note saying "because that would result in a record of infinite size" to Vass's suggestion, but prefer shorter error messages.

I think in the balance, I prefer Vass's suggestions over mine and think we can rely on StackOverflow or the like to help users with the "why" of this error if they don't snap to the problem.

cassella:
Records of Infinite Size?  I don't think they exist.

What happens if a record contains a R.O.I.S.?

```chapel
record A {
  var b: B;
}

record B {
  var b: B;
}

var a = new A();
```

bradcray:
Chuckling at the barely-caught Princess Bride reference (I was late to that party).

Paul: In your case, B would be flagged as an error, so I don't think there's a need to also flag an error for A (it's not poor A's fault that B's author did something dumb).  The mutually recursive case is more interesting because the blame is shared between all records involved.  But let me know if you're seeing something subtle that I'm missing.

vasslitvinov:
Brad, be honest, the real reason you liked Vass's suggestion is because it is essentially your own, with just a minor tweak. Right? :)

W.r.t. the "infinite size" part -- despite all the critiques, it was a great first shot that got us all talking and offering suggestions.

cassella:
@bradcray I was wondering what `recordContainsType(A, B)` would return.

bradcray:
> I was wondering what `recordContainsType(A, B)` would return.

I'm still not following, but maybe @benharsh is, which is probably more important.

benharsh:
Hmm, in the example @cassella wrote my branch encounters a segfault. Looks like B's initializer isn't resolved when I thought it was. I guess I'll have to check each field for an infinite record first before recursing.

benharsh:
Thanks for all the feedback thus far!

I have pushed several commits that update the error message text and correctly emit an error for the case introduced by @cassella . I also added a new error message for a recursive record cycle involving tuples.

benharsh:
Testing came back clean. Does anyone subscribed to this issue object to the new error message text? Would any of you be willing to officially review this PR?

compiler/resolution/functionResolution.cpp
```diff
@@ -8165,6 +8225,9 @@ static void resolveAutoCopies() {
         ts->hasFlag(FLAG_SYNTACTIC_DISTRIBUTION) == false) {
       if (AggregateType* at = toAggregateType(ts->type)) {
         if (isRecord(at) == true) {
+          // If we attempt to resolve auto-copy and co. for an infinite record
+          // we may enter an infinite loop and the compiler will crash.
+          checkForInfiniteRecord(at);
```
vasslitvinov: Please comment in the PR message why checkForInfiniteRecord() is in such odd places like wrapAndCleanUpActuals() and resolveAutoCopies(). Seems like it should occur when you are looking at the record itself.

benharsh: Resolution doesn't always give us a chance to properly 'handle' a record before we attempt to use its fields, so I inserted these checks right before we would otherwise encounter a segfault.

vasslitvinov: Thanks for updating the PR text.

Issue #8771, opened 2018-03-10
How do I resolve non-Mason dependencies


buddha314:
Trying to run `mason build` on https://github.com/buddha314/numsuch/releases/tag/v0.2.2 I get

```
 mason build
Updating mason-registry for local_registry
Updating mason-registry for mason-registry
Compiling NumSuch
In file included from /var/folders/g5/4zwyfk513hd94tjs1fm5mdq00000gn/T//chpl-buddha-55353.deleteme/_main.c:2:
/var/folders/g5/4zwyfk513hd94tjs1fm5mdq00000gn/T//chpl-buddha-55353.deleteme/chpl__header.h:5:10: fatal error: 'cblas.h' file not found
#include "cblas.h"
         ^~~~~~~~~
1 error generated.
make: *** [/var/folders/g5/4zwyfk513hd94tjs1fm5mdq00000gn/T//chpl-buddha-55353.deleteme/NumSuch.tmp] Error 1
error: compiling generated source
Build Failed
```

I'm not sure how to use Mason to tell the compiler my `BLAS_HOME= /usr/local/Cellar/openblas/0.2.20`

buddha314:
I expect the answer to how to handle this will be addressed somewhere in this issue: https://github.com/chapel-lang/chapel/issues/7106 

The notion of `.COMPOPTS` in the testing suite might not be a bad way to go.

buddha314:
I have experimented with `cmake` and [`find_package`](https://cmake.org/cmake/help/v3.0/command/find_package.html) a bit. I can't quite get it to work, but perhaps if we did a mason installer and it installed according to the `cmake` guidelines, it could be cobbled together.  Oh, looks this ticket touches on that: https://github.com/chapel-lang/chapel/issues/7639

ben-albrecht:
This is a planned feature for Mason.

Spartee:
More to come with this in #10260 but this ability has been merged into mason with #10289

PR #10289, opened 2018-07-11
Mason: pkgconfig integration

Spartee:
This PR represents the first time mason users will be able to specify external dependencies in their `Mason.toml` files. Documentation has been added to the Mason documentation, but for a brief overview.
- Mason users can specify an external package under the `[system]` table. Mason will grab the dependency information using the `pkg-config` command and place it in the `Mason.lock`. 
- This allows users who want to use packages such as the LinearAlgebra package to produce builds much faster than before.
- System packages must be downloaded by the user and the package's `.pc` file must be made available to Mason through the `PKG_CONFIG_PATH`.
- External dependencies can now also be used in `mason test` when configured the same as above.
- The `package = "*"` option is also available to users who want Mason to grab the latest version of `package` on their system.
- A `mason system` command was added for users to quickly debug their mason packages with system package dependencies. They are listed below
   - An option for users to print out the `.pc` file of a package has also been added to help users through the build process. This feature aids users who need to add specific compile options that are listed in the .pc file.
   - An option to list all system packages
   - An option to search all installed packages
- Some examples of commands to come include
   - Install pkg-config
   - set PKG_CONFIG_PATH
   - query other ``pkg-config`` variables


tools/mason/PkgConfig.chpl
```diff
@@ -0,0 +1,156 @@
...
+    var pcDir = "".join(getPkgVariable(pkgName, "--variable=pcfiledir")).strip();
+    var pcFile = joinPath(pcDir, pkgName + ".pc");
+    var pc = open(pcFile, iomode.r);
+    for line in pc.lines() {
+      write(line);
+    }
+  }
+  catch e: FileNotFoundError {
+    // start issue about this
+    //throw new MasonError("Package exists but Mason could not find it's .pc file");
```
ben-albrecht: Have you opened an issue for this?

doc/rst/tools/mason/mason.rst
```diff
@@ -311,6 +311,86 @@ If no query is provided, all packages in the registry will be listed.
...
+    
+    [external]
+        [external.pkgconfig]
+        lapack = "3.8.0"
+        openblas = "*"
+
+
+This ``Mason.toml`` is updated to reflect the necessary dependency information for Mason
+to build the package in the ``Mason.lock``. Mason gets external dependency information,
+in this case, from the ``.pc`` files found via the ``PKG_CONFIG_PATH``. The ``*`` means
```
ben-albrecht: Design Q: Should we attempt to support the range syntax for versions in addition to (or instead of) `*`, similar to `chplVersion`? e.g.

```
openblas = "2.3.." # 2.3.0 +
openblas = "4.."   # 4.0.0 +
openblas = "1..3" # 1.0.0 - 3.*.*
```

I'm OK with this being determined beyond this PR, but wanted to mention it.

tools/mason/PkgConfig.chpl
```diff
@@ -0,0 +1,156 @@
...
+   given pkgname and variable */
+proc getPkgVariable(pkgName: string, option: string) {
+
+  var command = " ".join("pkg-config", pkgName, option);
+
+  var lines: [1..0] string;
+  var cmd = command.split();
+  var sub = spawn(cmd, stdout=PIPE);
+  sub.wait();
+
```
ben-albrecht: Perhaps we should check the `sub.exit_status` and throw an error with the `sub.stderr` if it is not `0`

Spartee: Will add this to list of errors im compiling for the conversion to error-handling

tools/mason/PkgConfig.chpl
```diff
@@ -0,0 +1,156 @@
...
+proc getPkgInfo(pkgName: string, version: string) throws {
+
+  var pkgDom: domain(string);
+  var pkgToml: [pkgDom] unmanaged Toml;
+  var pkgInfo: unmanaged Toml = pkgToml;
+
+  if pkgExists(pkgName) {
+    const pcVersion = "".join(getPkgVariable(pkgName, "--modversion")).strip();
+    const libs = "".join(getPkgVariable(pkgName, "--libs")).strip();
+    const include = "".join(getPkgVariable(pkgName, "--variable=includedir")).strip();
```
ben-albrecht: Would `--cflags` be more robust instead instead of specifying the `includedir`? @mppf might have a good idea.

tools/mason/mason.chpl
```diff
@@ -81,6 +82,7 @@ proc main(args: [] string) {
     when 'env' do masonEnv(args);
     when 'doc' do masonDoc(args);
     when 'clean' do masonClean();
+    when 'pkg-pc' do masonPkgPcLookup(args);
```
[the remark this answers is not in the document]
ben-albrecht: Would that be in a follow-up PR then?

test/mason/pkgconfig-tests/openssl.good
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+chplVersion = "CHPL_CUR_FULL..CHPL_CUR_FULL"
+
+[external]
+
+[external.pkgconfig]
+
+[external.pkgconfig.openssl]
+name = "openssl"
+version = "0.9.8zh"
```
ben-albrecht: Hard-coding versions like this into the `.good` files may be fragile to system updates. Could we use a prediff to plug in `openssl version | awk '{print $2}'`  (or the equivalent for other packages) into the version field to make these tests more resilient?

tools/mason/PkgConfig.chpl
mppf: I'd expect we'd want the simplest use of pkg-config possible, where we ask for `--libs` (which really means *linker flags*) and `--cflags` which is the compilation flags (including `-I` arguments).

Is there a particular reason we shouldn't do that?

test/mason/pkgconfig-tests/SKIPIF
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+
+if [[ $HOST == *chapcs* ]] ; then
```
ben-albrecht: If we address the version-independence issue I mentioned below, we may be able to change this constraint to checking if the needed packages are available on the system, e.g.


```bash
# Check if pkg-config is available
command -v pkg-config 2> /dev/null || echo False && exit

# Check if packages are available
pkg-config --exists openssl  || echo False && exit

echo True
```

This may lead to some odd errors in certain testing configurations / platforms, but I think we want to catch those before users do.

mppf: Does Mason give a reasonable error if pkg-config is not found (I didn't look at the code here)?

tools/mason/mason.chpl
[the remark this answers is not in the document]
Spartee: yes the mason external will be introduced in the PR that includes the functionality for `mason external [options]`

tools/mason/PkgConfig.chpl
Spartee: pkg-config assumes a number of things about the system including that if a package's include dir is `/usr/include/` then it prints nothing because that location is assumed by the c compiler. This is a way to ensure that the correct information is always put into the lock file.

doc/rst/tools/mason/mason.rst
Spartee: I would like to see this functionality for external deps not from pkg-config, but for this PR and the purpose of Pkg-config prototyping I think the * is more than enough user functionality.

tools/mason/PkgConfig.chpl
mppf: Why do we need to put /usr/include into the lock file? What if the pkgconfig package specifies another C flag (e.g. -shared etc)

test/mason/pkgconfig-tests/openssl.good
ben-albrecht: Actually, `pkg-config --modversion <package>` would be better than my earlier example, but you probably already figured that out.

tools/mason/PkgConfig.chpl
Spartee: `/usr/include` needs to be included because we can't special case when then path is `/usr/include` during build because if nothing is returned than the toml file is invalid. mason build expects there to be something there and if it is not there then the toml file is invalid and it cant read the toml file anyway. Now we could special case mason update and put a placeholder into the toml file but in my opinion, at that point, we should just put `usr/include` in the lock file.

mppf: If you just need a placeholder, "" could work couldn't it? Or something along those lines?

It seems really weird to me to query pkg-config so that we get a placeholder that our TOML parser can parse. That seems like something we should handle in a different way. It also seems weird to put `/usr/include` into the lock file, since as mentioned above, the C compiler will search there for includes anyway.

test/mason/pkgconfig-tests/SKIPIF
ben-albrecht: Yes

tools/mason/MasonHelp.chpl
```diff
@@ -159,3 +159,20 @@ proc masonTestHelp() {
...
+proc masonSystemHelp() {
+  writeln("Integrate a Mason package with system packages found via pkg-config");
+  writeln();
+  writeln("Usage:");
+  writeln('    mason system [options] [<args>...]');
+  writeln("    mason system [options]");
+  writeln();
+  writeln("Options:");
+  writeln("    pc                          Print a system packages .pc file");
+  writeln("    list                        List all packages available on the system");
```
ben-albrecht: Should this actually be `search` to reflect `mason search`, where an empty `mason search` results in listing all available packages?

tools/mason/MasonHelp.chpl
```diff
@@ -159,3 +159,20 @@ proc masonTestHelp() {
...
 
+proc masonSystemHelp() {
+  writeln("Integrate a Mason package with system packages found via pkg-config");
+  writeln();
+  writeln("Usage:");
+  writeln('    mason system [options] [<args>...]');
+  writeln("    mason system [options]");
+  writeln();
+  writeln("Options:");
+  writeln("    pc                          Print a system packages .pc file");
```
ben-albrecht: `Print a system package's .pc file`

ben-albrecht: We'll want `--help` outputs for each subsubcommand here as well. `mason system pc --help` fails catastrophically, e.g.

doc/rst/usingchapel/prereqs.rst
```diff
@@ -40,6 +40,9 @@ about your environment for using Chapel:
   * If you wish to use chpldoc or Chapel's test system, ``curl`` and
     python-devel (or equivalent packages for your platform) are required.
 
+  * If you wish to use Mason, chapel's package manager, ``pkg-config`` is
```
ben-albrecht: If you wish to use the `mason system` subcommand of Mason, chapel's package manager, ``pkg-config`` is required.

tools/mason/MasonHelp.chpl
```diff
@@ -159,3 +161,45 @@ proc masonTestHelp() {
...
+proc masonSystemHelp() {
+  writeln("Integrate a Mason package with system packages found via pkg-config");
+  writeln();
+  writeln("Usage:");
+  writeln('    mason system [options] [<args>...]');
+  writeln("    mason system [options]");
+  writeln();
+  writeln("Options:");
+  writeln("    pc                          Print a system package's .pc file");
+  writeln("    search                      List all packages available on the system");
```
ben-albrecht: Should this be "Search all packages available on system" ?

Spartee:
Completion of this PR will close #8771

doc/rst/usingchapel/prereqs.rst
```diff
@@ -40,7 +40,7 @@ about your environment for using Chapel:
   * If you wish to use chpldoc or Chapel's test system, ``curl`` and
     python-devel (or equivalent packages for your platform) are required.
 
-  * If you wish to use Mason, chapel's package manager, ``pkg-config`` is
+  * If you wish to use the ``mason system``, chapel's package manager, ``pkg-config`` is
```
ben-albrecht: > If you wish to use the ``mason system``, chapel's package manager,

If you wish to use the ``mason system`` subcommands of Mason, chapel's package manager,

tools/mason/MasonHelp.chpl
```diff
@@ -187,6 +186,8 @@ proc masonSystemSearchHelp() {
   writeln();
   writeln("Options:");
   writeln("    -h, --help                  Display this message");
+  writeln("        --no-desc               Only display package name");
+  writeln("        --hungry                Parse descriptions of package to include more search results");
```
ben-albrecht: Interesting flag name...

tools/mason/MasonSystem.chpl
```diff
@@ -65,38 +62,58 @@ proc pkgConfigExists() throws {
...
-    sub.wait();
-
-    for line in sub.stdout.lines() {
+    for arg in args[3..] {   
+      if arg == '-h' || arg == '--help' {
+        masonSystemSearchHelp();
+        exit(1);
+      }
+      else if arg == "--hungry" {
+        hungry=true;
```
ben-albrecht: inconsistent spacing here

tools/mason/MasonHelp.chpl
Spartee: @ben-albrecht  It's hungry! it consumes more for the search! It made sense to me...other suggestions?

ben-albrecht: Some potentially more intuitive flag names (at the expense of fun), ordered roughly by my preference:

- `--search-desc`
- `--extended`
- `--broad`
- `--advanced`

benharsh: How about ``--[no-]show-desc`` to toggle display of descriptions, and ``--[no-]desc`` to toggle searching of descriptions? FWIW, homebrew uses ``--desc`` as well.

ben-albrecht: > FWIW, homebrew uses --desc as well.

Ah, I did not notice this. Good point. Those would work for me.

PR #10290, opened 2018-07-11
Hoist nested unmanaged types to module scope

### LINKED ISSUE #10284: Problems with nested class types and unmanaged
The following two programs fail to compile:

``` chapel
class Outer {

  var field:unmanaged Inner = nil;

  pragma "use default init"
  class Inner {
    var x:int;
  }

  proc init() {
    field = new unmanaged Inner(1);
  }
}

var x = new Outer();
writeln(x);
```

with `error: 'unmanaged Inner' used before defined`

``` chapel
pragma "use default init"
class Outer {

  var field:unmanaged Inner;

  pragma "use default init"
  class Inner {
    var x:int;
  }
}

var x = new Outer();
writeln(x);
```

with `error: can't omit initialization of field "unmanaged Inner", no type or default value provided`

test/classes/initializers/nested/unmanaged-problem-with-default.chpl
test/classes/initializers/nested/unmanaged-problem-use-before-def.chpl

benharsh:
The defPoint for UnmanagedClassTypes needs to be hoisted to module scope, just like regular AggregateTypes, if it's nested within an AggregateType.

Resolves #10284

Testing:
- [x] local + futures
- [x] --force-initializers

### DIFF
--- a/compiler/passes/flattenClasses.cpp
+++ b/compiler/passes/flattenClasses.cpp
@@ -24,10 +24,11 @@
 
 void flattenClasses() {
   forv_Vec(TypeSymbol, ts, gTypeSymbols) {
-    if (AggregateType* ct = toAggregateType(ts->type)) {
-      if (toAggregateType(ct->symbol->defPoint->parentSymbol->type)) {
-        ModuleSymbol* mod = ct->getModule();
-        DefExpr*      def = ct->symbol->defPoint;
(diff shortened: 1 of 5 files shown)


benharsh:
On this branch there is just one failure for --force-initializers:
``classes/bradc/overloadMethods/virtualIsHalt.chpl``

PR #10293, opened 2018-07-12
Propagate environment variables to ibv-launched programs

dmk42:
The `gasnetrun_ibv` launcher is written in Perl, which mangles binary strings when its UTF-8 support is turned on.  Therefore, GASNet sets all the relevant character-set-support environment variables to a state that prevents UTF-8 support in the launcher.  Unfortunately, this also changes the environment for the Chapel program being launched.  The launcher's `-E` flag to propagate environment variables does not prevent this problem.

This change replicates the pre-launcher character set environment in the command line to the launcher so that the Chapel executable can support UTF-8.

### DIFF
--- a/runtime/src/launch/slurm-gasnetrun_ibv/launch-slurm-gasnetrun_ibv.c
+++ b/runtime/src/launch/slurm-gasnetrun_ibv/launch-slurm-gasnetrun_ibv.c
@@ -158,6 +158,34 @@ static void genNumLocalesOptions(FILE* slurmFile, sbatchVersion sbatch,
   }
 }
 
+static void propagate_environment(FILE *f)
+{
+  char *lang = getenv("LANG");
+  char *lc_all = getenv("LC_ALL");
+  char *lc_collate = getenv("LC_COLLATE");
+
+  // If any of the relevant character set environment variables
+  // are set, replicate the state of all of them.
+  if (lang || lc_all || lc_collate) {
+    fprintf(f, " env");
+
+    // All the undefines have to come first.
+    if (!lang)
+      fprintf(f, " -u LANG");
+    if (!lc_all)
+      fprintf(f, " -u LC_ALL");
+    if (!lc_collate)
+      fprintf(f, " -u LC_COLLATE");
+
+    if (lang)
+      fprintf(f, " LANG=%s", lang);
+    if (lc_all)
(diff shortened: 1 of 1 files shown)


dmk42:
@gbtitus and @ronawho - Would you mind taking a post-commit look at this?

runtime/src/launch/slurm-gasnetrun_ibv/launch-slurm-gasnetrun_ibv.c
```diff
@@ -158,6 +158,34 @@ static void genNumLocalesOptions(FILE* slurmFile, sbatchVersion sbatch,
...
+    // All the undefines have to come first.
+    if (!lang)
+      fprintf(f, " -u LANG");
+    if (!lc_all)
+      fprintf(f, " -u LC_ALL");
+    if (!lc_collate)
+      fprintf(f, " -u LC_COLLATE");
+
+    if (lang)
+      fprintf(f, " LANG=%s", lang);
```
gbtitus: There's not the least chance any of these 3 env var settings will contain a character that's a meta-character to the shell that runs the gasnetrun_obv command, is there?  (I don't think so, but that may mostly reflect the limits of my own knowledge.)

gbtitus:
Looks good!   👍

runtime/src/launch/slurm-gasnetrun_ibv/launch-slurm-gasnetrun_ibv.c
dmk42: I thought of that, and added a commit that put quotes around those characters.  I had to take the quotes back out because metacharacters turn out not to be interpreted here, and it was making the quotes part of the string.

gbtitus: Oh!  So the command must not be interpreted by a shell after all.  A follow-up: there's not the least chance that there will be a white space character within the values of any of these 3, is there?   :-)

dmk42: I guess someone could always define one of the environment variables incorrectly, but all the correct values have no special characters, including no spaces.  If they define them incorrectly, they have worse problems.

Issue #10177, opened 2018-07-06
Compiler error when casting c_void_ptr to c_ptr from a function invoked in a forall nested in a on-statement


LouisJenkinsCS:
**Minimal Working Example**

```chpl
 class A {
   proc fn() : c_ptr(int) {
     return c_nil : c_ptr(int);
   }
 }

 proc main() {
   var a = new A();
   on Locales[0] {
     forall 1..1 {
       var ptr = a.fn();
     }
   }
 }
```

Note that removing the explicit cast `c_nil : c_ptr(int)` shows this error...

```
type mismatch in assignment from c_void_ptr to c_ptr(int(64))
```

[Update] To expose the issue: (a) need to compile with --no-local; 'a' can also be an int or other types, any function taking 'a' as an argument can be used instead of a.fn(). Upcoming test:

    test/parallel/forall/vass/flatten-with-ovar.chpl

However, with the explicit cast, upon inspection from debugging the compiler via gdb, some information that may assist in debugging this issue can be shown below...

```
 {  
   <Expr> =   {  
      <BaseAST> =      {  
         _vptr.BaseAST = 0x7a5140 <vtable for SymExpr+16>,
         astTag = E_SymExpr,
         id = 1125140,
         astloc =         {  
            filename = 0xb4d350 "TupleMWE.chpl",
            lineno = 11
         },
         static tabText = <optimized out>
      },
      parentSymbol = 0xa240090,
      parentExpr = 0x0,
      list = 0x0,
      prev = 0x0,
      next = 0x0
   },
   var = 0x292a7b0,
   symbolSymExprsPrev = 0xa590cc0,
   symbolSymExprsNext = 0x0
}
```

 Note that the `parentExpr` is `nil`, which @bradcray said that this should never happen. As well, the variable `parentSymbol`, which presumably is the variable declaration, has the name `a`...

```
{  
   <BaseAST> =   {  
      _vptr.BaseAST = 0x7acc70 <vtable for ShadowVarSymbol+16>,
      astTag = E_ShadowVarSymbol,
      id = 1125141,
      astloc =      {  
         filename = 0xb4d350 "TupleMWE.chpl",
         lineno = 11
      },
      static tabText = <optimized out>
   },
   qual = QUAL_CONST_VAL,
   type = 0x29299e0,
   flags =   {  
      <std::_Base_bitset<4>> =      {  
         _M_w =         {  
            103079215104,
            0,
            0,
            0
         }
      },
      <No data fields>
[...]

PR #10294, opened 2018-07-12
Support c_void_ptr -> c_ptr assignments and stop treating c_ptr as a class

bradcray:
This PR makes two simple changes:

1) It adds support for assigning from c_void_ptr to c_ptr
2) It has isClassType() return `false` for `c_ptr` types (without this second change, the first resulted in an ambiguity error

It also adds two simple tests to lock in this behavior.

This test fixes the issue reported in issue #10177 though I'm not sure how it's related to the [observed behavior](https://github.com/chapel-lang/chapel/issues/10177#issuecomment-404518430) there in which a SymExpr is missing its parentExpr causing an assertion failure.  I'm going to hold off on closing that issue until we understand the failure mode better to make sure we're not papering over a bigger problem; and to verify that this helps fix things for the larger code which motivated its simplified test case.


compiler/resolution/preFold.cpp
```diff
@@ -456,7 +456,8 @@ static Expr* preFoldPrimOp(CallExpr* call) {
   } else if (call->isPrimitive(PRIM_IS_CLASS_TYPE)) {
     Type* t = call->get(1)->typeInfo();
 
-    if (isClassLike(t) && !t->symbol->hasFlag(FLAG_EXTERN)) {
+    if (isClassLike(t) && !t->symbol->hasFlag(FLAG_EXTERN) &&
+        !t->symbol->hasFlag(FLAG_C_PTR_CLASS)) {
```
mppf: Should it also rule out _ddata (FLAG_DATA_CLASS)?

modules/internal/CPtr.chpl
```diff
@@ -106,6 +106,9 @@ module CPtr {
   pragma "no doc"
   inline proc =(ref a:c_ptr, b:c_ptr) { __primitive("=", a, b); }
 
+  pragma "no doc"
+  inline proc =(ref a: c_ptr, b: c_void_ptr) { __primitive("=", a, b); }
+
```
vasslitvinov: How about going in the other direction, `=(c_void_ptr, c_ptr)` ? Seems it is equally - or even more - valid.

vasslitvinov:
With the disclaimer that I don't use these features, this looks good to me. 
I suggest allowing `=(c_void_ptr, c_ptr)` as well.

modules/internal/CPtr.chpl
mppf: That should already work because we allow coercions from c_ptr to c_void_ptr.

bradcray: Trying a quick test, it seems it does.

vasslitvinov:
@bradcray - it is OK to go ahead and merge this PR before #10177 is fixed.
10177 still crashes - just need `--no-local`.

bradcray:
> #10177 still crashes - just need `--no-local`.

D'oh!

bradcray:
(that said, I was just waiting to see if anybody squawked at the design, less because of the bug... but thanks for pointing out my error).

compiler/resolution/preFold.cpp
bradcray: I'm gonna leave that to a separate PR just to keep this one simple / focused and leverage the testing I've already done.  But I'll check it out.

PR #10295, opened 2018-07-12
lineno improvements

mppf:
@ysahil97 who is working on the Polly+Chapel GSoC project noticed that even
simple functions in the internal/standard modules were getting line number and
file name arguments. That project needs to be able to add a function to the
internal modules that does not get these arguments and that is marked with
LLVM's readnone function attribute.

This PR takes several steps:

* Removes passLineno=true from many primitives where the file name / line
number information is not used in code generation. This helps to prevent
unnecessarily adding file name / line number arguments to internal/standard
module functions when --developer is not set.

* PRIM_CHECK_NIL uses the additional arguments to the primitive for line number
/ filename instead of getting them from the statement. (If we get them from the
statement, why bother having these additional arguments?). Also updated several
comments in primitive code generation to indicate file/line arguments that have
been removed because they were unused.

* allows FLAG_LINE_NUMBER_OK to be set on a function with a pragma and allow
combining it with FLAG_INSERT_LINE_FILE_INFO for testing purposes (in which
case the line/file arguments are not added)

* adds pragma/flag FLAG_LLVM_READNONE to set LLVM's readnone function attribute


compiler/codegen/expr.cpp
```diff
@@ -4258,8 +4254,9 @@ DEFINE_PRIM(PRIM_CHECK_NIL) {
 
     codegenCall("chpl_check_nil",
                 ptr,
-                gGenInfo->lineno,
-                gFilenameLookupCache[gGenInfo->filename]);
+                call->get(2),
+                call->get(3)); /*gGenInfo->lineno,
+                gFilenameLookupCache[gGenInfo->filename]); */
```
benharsh: Did you intend to leave this comment in?

PR #10298, opened 2018-07-12
Allow extern procs to have a computed param cname

### LINKED ISSUE #10053: Allow extern proc renaming to take a param string
We support renaming of extern procs so that a different function name is available to Chapel code. e.g. 

```chpl
extern "foo_in_c" proc foo_in_chapel(): c_int;
writeln(foo_in_chapel()); // will generate a call to foo_in_c
```

Today, the external name must be a string literal, since it's substituted in at parse time. It'd be nice to allow this to be a param string instead. This would allow us to compute external names based on types and other sorts of things, which would be really beneficial for cleaning up things like our Atomics implementation.

Since our runtime support for atomics is in C, we have to embed the type name in the function. So even though Chapel supports generics and overloads, today we create nearly identical records for each type just so that we can call unique C routines. So for each type currently we have something like:

```chpl
extern proc atomic_fetch_add_explicit_int_least64_t(ref obj:atomic_int_least64_t, operand:int(64), order:memory_order):int(64);
inline proc add(value:int(64), order:memory_order = memory_order_seq_cst):void {
  on this do atomic_fetch_add_explicit_int_least64_t(_v, value, order);
}

// repeated ~10 times for the other types
```

We could instead just have one generic record (or really 1 for integral, 1 for real, and 1 for bool) with something like:

```chpl
proc externType type { return _v.type; }
proc plainChplType type {
  if externType == atomic_int_least64_t then return int(64);
  /* other types ... */
}
proc cFuncSuffix() param {
  if externType == atomic_int_least64_t then return "int_least64_t";
  /* other types ... */
}

inline proc add(value:plainChplType, order:memory_order = memory_order_seq_cst): void {
  extern "atomic_fetch_add_explicit_"+cFuncSuffix()  /* param renaming */
    proc atomic_add(ref obj:externType, operand:plainChplType, order:memory_order):plainChplType;
  on this do atomic_add(_v, value, order);
}
```

mppf:
Resolves #10053.

* Parser now handles arbitrary param expressions to compute the 
  cname as in `extern computeCname() proc foo()`
* The identifier form is no longer supported (but it wasn't
   documented, either)
* These param-cname-expressions are stored in the AST as the default 
  expression for a special argument named _chpl_cname
* resolveFormals finds formal arguments with this name, resolves them,
 and then sets the cname to the resolved string value. It then removes
 this formal since it has served its purpose.

I consider marking these formals with INTENT_PARAM but that required more 
code changes to work with them.

- [x] full local testing

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -2171,9 +2171,16 @@ FnSymbol* buildLambda(FnSymbol *fn) {
 }
 
 // Creates a dummy function that accumulates flags & cname
-FnSymbol* buildLinkageFn(Flag externOrExport, const char* cname, Expr* paramCNameExpr) {
+FnSymbol* buildLinkageFn(Flag externOrExport, Expr* paramCNameExpr) {
 
-  if (cname == NULL) cname = "";
+  const char* cname = "";
+  // Look for a string literal we can use
+  if (paramCNameExpr != NULL)
+    if (SymExpr* se = toSymExpr(paramCNameExpr))
+      if (VarSymbol* v = toVarSymbol(se->symbol()))
+        if (v->isImmediate())
+          if (v->immediate->const_kind == CONST_KIND_STRING)
+            cname = v->immediate->v_string;
 
   FnSymbol* ret = new FnSymbol(cname);
 
@@ -2186,7 +2193,8 @@ FnSymbol* buildLinkageFn(Flag externOrExport, const char* cname, Expr* paramCNam
     ret->addFlag(FLAG_EXPORT);
   }
 
-  if (paramCNameExpr) {
+  // Handle non-trivial param names that need to be resolved
+  if (paramCNameExpr && cname[0] == '\0') {
     DefExpr* argDef = buildArgDefExpr(INTENT_BLANK,
                                       astr_chpl_cname,
                                       new SymExpr(dtString->symbol),
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -2170,6 +2170,32 @@ FnSymbol* buildLambda(FnSymbol *fn) {
   return fn;
 }
 
+// Creates a dummy function that accumulates flags & cname
+FnSymbol* buildLinkageFn(Flag externOrExport, const char* cname, Expr* paramCNameExpr) {
+
+  if (cname == NULL) cname = "";
+
+  FnSymbol* ret = new FnSymbol(cname);
+
+  if (externOrExport == FLAG_EXTERN) {
+    ret->addFlag(FLAG_LOCAL_ARGS);
+    ret->addFlag(FLAG_EXTERN);
+  }
+  if (externOrExport == FLAG_EXPORT) {
+    ret->addFlag(FLAG_LOCAL_ARGS);
+    ret->addFlag(FLAG_EXPORT);
+  }
+
+  if (paramCNameExpr) {
+    DefExpr* argDef = buildArgDefExpr(INTENT_BLANK,
+				      astr_chpl_cname,
+				      new SymExpr(dtString->symbol),
+				      paramCNameExpr, NULL);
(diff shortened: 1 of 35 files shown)


mppf:
@ronawho - enjoy :)

mppf:
@lydia-duncan - I'm going to run a final testing run after review completes, but could you have a look at the recent changes (parens not required; chpldoc compatability & test; documentation) ?

spec/Interoperation.tex
```diff
@@ -118,14 +128,24 @@ \subsection{Calling Chapel Functions}
...
-supplied, then it is used verbatim as the exported function symbol.  Otherwise,
-the Chapel name of the procedure is exported.  The rest of the procedure
-declaration is the same as for a non-exported function.  An exported procedure can be
-called from within Chapel as well.  Currently, iterators cannot be exported.
+The rest of the procedure declaration is the same as for a non-exported
+function.  An exported procedure can be called from within Chapel as
+well. Currently, iterators cannot be exported.
+
+As with the \snxt{extern-name} for \chpl{extern} \chpl{proc}, if this
+syntax element is provided, then it is a \chpl{param} \chpl{string}
```
lydia-duncan: "then it is a param string determining the name" -> "then it must be a param string and will be used to determine the name"

test/chpldoc/linkage-spec/exportTest.doc.good
```diff
@@ -72,6 +72,10 @@ exportTest.doc
 
 .. function:: export proc allThreeCommentless2(val: int): int
 
+.. function:: export proc withExternCName()
+
+.. function:: export proc withParamExternCName()
```
lydia-duncan: It might be nice to open an issue for having the C name displayed as well in a nice way, but this is a reasonable update for this PR

compiler/AST/build.cpp
```diff
@@ -2194,7 +2194,8 @@ FnSymbol* buildLinkageFn(Flag externOrExport, Expr* paramCNameExpr) {
   }
 
   // Handle non-trivial param names that need to be resolved
-  if (paramCNameExpr && cname[0] == '\0') {
+  // the check for dtString->symbol avoids this block under chpldoc
+  if (paramCNameExpr && cname[0] == '\0' && dtString->symbol != NULL) {
     DefExpr* argDef = buildArgDefExpr(INTENT_BLANK,
```
lydia-duncan: I think I would prefer this argument to have a special flag or something that chpldoc could key off of, instead of altering the behavior at parse time depending on our compilation mode

mppf: Something is already (very) altered under chpldoc as dtString->symbol == NULL there, so we get assert failures from buildArgDefExpr (before the main chpldoc traversal even runs).

lydia-duncan: Ah, okay.  That's a good justification.  I know we did some modification so that we don't generate documentation for internal symbols, so it makes sense that is happening (but is less than ideal imo)

test/chpldoc/linkage-spec/exportTest.doc.good
mppf: #10387

lydia-duncan: Thanks :)

bradcray:
I was spending some spare cycles working on adding renaming for external variable symbols today and looked at this code for the first time to understand how non-literal strings were handled.  I'm surprised that we took the approach of adding an extra argument to the call and then removing it later (which feels very fragile old-school Chapel compiler mentality to me) rather than storing an `Expr* externCName` in the `[Fn]Symbol` class (say).  It also has the downside that it doesn't extend to extern symbols other than functions.  Was that approach considered and discarded?  (because that's how I'd be inclined to do it for extern variables, say). Thanks.

mppf:
I think storing an Expr* in FnSymbol would be a good improvement and I don't remember discarding that approach.

PR #10299, opened 2018-07-12
Document bigint.divexact

avneet14027:
Test Documentation for proc bigint.divexact(const ref n: bigint, const ref d: bigint)  created.

### DIFF
--- a/modules/standard/BigInteger.chpl
+++ b/modules/standard/BigInteger.chpl
@@ -2218,8 +2218,8 @@ module BigInteger {
 Computes the quotients when corresponding elements of a bigint array *n* are divided by corresponding elements of a bigint array *d*, and stores the quotient in another bigint array preceding the function call for whichever elements of the array are exactly divisible, with remainder zero.
 
 * Arguments:
-	* const ref n: bigint: a bigint array
-	* const ref d: bigint: another bigint array
+    * const ref n: bigint: a bigint array
+    * const ref d: bigint: another bigint array
 
 */
 
--- a/modules/standard/BigInteger.chpl
+++ b/modules/standard/BigInteger.chpl
@@ -2214,14 +2214,10 @@ module BigInteger {
   // divexact
 
 /*
-This function computes the quotient when a bigint variable *n* is divided by a
-bigint variable *d*, i.e. it calculates, ``q = n/d``
-where, *q*: quotient, *n*: dividend, *d*: divisor.
+Computes ``n/d`` and stores the result in ``bigint`` instance.
 
-The result is stored in ``this``, that is, the method receiver. It works faster
-than regular division, when exact division is known to occur.
-It may produce incorrect results,
-when *n* is not divisible by *d*.
+``divexact`` is optimized to handle cases where ``n/d`` results in an integer. 
+When ``n/d`` does not produce an integer, this method may produce incorrect results.
 
 :arg n: a bigint variable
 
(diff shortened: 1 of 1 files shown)


mppf:
@avneet14027 - thanks very much for making this PR!

modules/standard/BigInteger.chpl
```diff
@@ -2212,6 +2212,18 @@ module BigInteger {
 
 
   // divexact
+
+/*
+
+Computes the quotients when corresponding elements of a bigint array *n* are divided by corresponding elements of a bigint array *d*, and stores the quotient in another bigint array preceding the function call for whichever elements of the array are exactly divisible, with remainder zero.
```
mppf: Generally speaking, we keep Chapel module source code to 80 columns (in other words you should add newlines to this to make the lines shorter).

modules/standard/BigInteger.chpl
```diff
@@ -2212,6 +2212,18 @@ module BigInteger {
...
+
+* Arguments:
+    * const ref n: bigint: a bigint array
+    * const ref d: bigint: another bigint array
```
mppf: Since this is a common thing to say in these documentation comments, chpldoc has direct support for it, see https://chapel-lang.org/docs/tools/chpldoc/chpldoc.html#argument-and-return-yield-types

So for example you could write:

   :arg n: a bigint
   :type n: bigint

but these will be redundant with what the documentation already displays in my opinion.

mppf: There is also
```
 :returns: n/d provided that d divides n
```

modules/standard/BigInteger.chpl
mppf: The GMP manual for this function is here: https://gmplib.org/manual/Integer-Division.html

It describes the function in this way:
> Set q to n/d. These functions produce correct results only when it is known in advance that d divides n.
> 
> These routines are much faster than the other division functions, and are the best choice when exact division is known to occur, for example reducing a rational to lowest terms. 

So there are a differences between the new documentation above and the description from GMP:
 * shorter math-y formulation `n/d`
 * the term "bigint array"
 * less emphasis on the function producing incorrect results if n is not divisible by d

I think it'd be better to be closer to the GMP documentation in these regards.

mppf: And to mention that it's expected to be faster than regular division (where it is not assumed that d divides n)

modules/standard/BigInteger.chpl
mppf: It would be important to say that the result is stored in `this` which is also referred to as "the method receiver".

modules/standard/BigInteger.chpl
```diff
@@ -2212,6 +2212,27 @@ module BigInteger {
 
 
   // divexact
+
+/*
+This function computes the quotient when a bigint variable *n* is divided by a
```
ben-albrecht: While this is acceptable, I think it's a little verbose, given how many functions there will be to document. I would suggest spending less effort per each method, and keeping the documentation concise, e.g.

````rst
Computes ``n/d`` and stores result in the``bigint`` instance.

``divexact`` is optimized to handle cases where ``n/d`` results in an integer. 
When ``n/d`` does not produce an integer, this method may produce incorrect results.
````

File: modules/standard/BigInteger.chpl [hunk 4]
```diff
@@ -2212,6 +2212,23 @@ module BigInteger {
...
+:arg n: numerator
+
+:type n: bigint
+
+:arg d: denominator
+
+:type d: bigint
+*/
+
+
```
mppf: I'm about to merge this anyway, but typically we don't leave space between the docs-comment and the function documented.

PR #10302, opened 2018-07-12
Finish setting UTF-8 environment for testing

dmk42:
In #10213 , I set the LANG and LC_COLLATE environment variables in start_test to enable UTF-8 without destroying the sort order.  I forgot one last step, which is to unset LC_ALL, because that overrides the other two.  This change corrects that oversight.

@vasslitvinov was being impacted by the problem that this change fixes.

### DIFF
--- a/util/start_test
+++ b/util/start_test
@@ -627,7 +627,8 @@ def check_environment_with_args():
     if not tgt_platform == "sunos":
         os.environ["LC_COLLATE"] = "C"
         os.environ["LANG"] = "en_US.UTF-8"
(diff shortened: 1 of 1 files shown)


dmk42:
@ben-albrecht - Would you mind reviewing this from a Python point of view?  I'm interested in seeing if I chose the preferred method for deleting an environment variable.

dmk42:
@ben-albrecht - I'll go ahead and merge to fix master, but would still appreciate it if you'd take a look when you get a chance.

PR #10305, opened 2018-07-12
Use pthread mutexes instead of chpl syncs for lock based atomics

ronawho:
`CHPL_ATOMICS=locks` is a low performance, but portable atomics implementation.
Currently, it's only used for pgi where neither the __sync intrinsics nor c11
atomics are supported.

Historically we've used `chpl_sync_lock()/unlock()` to implement these portable
atomics, but this switches to use `pthread_mutex_lock()/unlock()`. The primary
motivation for this is because we use atomics before the tasking layer is
initialized and after it's shutdown, so we can't actually rely on using chpl
syncs. Historically we've gotten away with it because lock/unlock under fifo
used pthread mutexes, which didn't require tasking layer initialization and
until recently (#10133) qthreads lock/unlock was implemented with an atomic
spinwait using qthread_incr which will usually not require qthreads to be
initialized (if inline asm or __sync primitives are used) but that's not
guaranteed so we just lucky.

Now that chpl_sync_lock maps down to qthread_lock (thin wrapper over
qthread_readFE()), the tasking layer must be initialized in order to use
lock/unlock. Since we need to use atomics when the tasking layer isn't
initialized, this just switches the locks based atomic implementation over to
using pthread mutexes. This is safe to do because the critical section can
never yield, so we can't run into situations where we yield to another task
while holding the mutex. This rational (and motivation) is similar to the
switch to a pthread_mutex for the mem tracking table (ae050cd3c0, ffb4c55ac1)

As a side-benefit, pthread mutexes are really optimized (and do hybrid
spin-waiting/blocking) so they get better performance. Here's some results for
a very simple microbenchmark that just does an atomic add under heavy and under
no contention.

```chpl
use Time;

config const iters = 10000000;
config const doSerial = false;

var a: atomic int;

var t: Timer; t.start();
serial doSerial {
  forall 1..iters {
    a.add(1);
  }
}
t.stop();

writeln("a: ", a.read(), " elapsed: ", t.elapsed());
```

| lock version               | serial | parallel |
| -------------------------- | ------ | -------- |
| qthread_incr  (pre #10133) | 0.15s  | 10.1s    |
| chpl sync     (master)     | 1.36s  | 5.77s    |
| pthread mutex (this PR)    | 0.16s  | 0.93s    |

Which shows that pthread mutex is as fast as the atomic spinwait under no
contention and faster than our sync under contention.

### DIFF
--- a/doc/rst/technotes/atomics.rst
+++ b/doc/rst/technotes/atomics.rst
@@ -46,7 +46,7 @@ If compiler support for atomics is available, the atomic operations
 will be mapped down the appropriate compiler intrinsics which often
 map directly to processor atomics.  If intrinsics are not available,
 the atomic implementation defaults to using locks in the form of
-Chapel's sync vars. As a result the locks implementation will be
+Pthread mutexes. As a result the locks implementation will be
 slower than the intrinsic implementation. Since Chapel's atomics
 were modeled after the C11 edition of the C standard, the cstdlib
 implementation is just a wrapper around C standard atomics.  As C11
--- a/doc/rst/usingchapel/chplenv.rst
+++ b/doc/rst/usingchapel/chplenv.rst
@@ -414,8 +414,8 @@ CHPL_ATOMICS
                      standard atomics (from C11)
         intrinsics   implement atomics using target compiler intrinsics
                      (which typically map down to hardware capabilities)
-        locks        implement atomics by using Chapel sync variables to
-                     protect normal operations
+        locks        implement atomics by using mutexes to protect normal
+                     operations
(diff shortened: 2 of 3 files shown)


ronawho:
@gbtitus I'm going to merge this today to quiet testing (and since you gave the general idea an ok), but I'd appreciate a post-commit review when you get a chance

dmk42:
I looked at this out of curiosity.  It looks good to me.

ronawho:
Passed full `CHPL_ATOMICS=locks` paratest

Issue #10307, opened 2018-07-13
Represent enums as ordinal values rather than C enums


bradcray:
Recent work and discussions have led to a proposal to represent enums at execution time using ordinal integer values rather than C enums (or integer values in LLVM that resemble the C enums' values).  This task is meant to capture the work of making that conversion to put us in a saner and more stable world.  Doing so would also permit a number of compiler-generated routines for enums to be simplified+optimized or removed since enum->string or enum->int conversions could be done with a lookup into a dense array of values (with bounds-checking for semi-concrete enums in the enum->int case); and enum<->order conversions would largely no longer be required.

PR #10308, opened 2018-07-13
Optimize enum->int casts for concrete types

### LINKED ISSUE #10303: Performance regressions from #10114
There were some performance regressions for fasta as a result of "disabling enum->int casts for enums without explicit values" (https://github.com/chapel-lang/chapel/pull/10114)

Here's some of the [graphs](https://chapel-lang.org/perf/chapcs/?startdate=2018/06/27&enddate=2018/07/10&graphs=fastashootoutbenchmarkn25000000,fastareduxshootoutbenchmarkn25000000)

image

Fasta uses binary I/O, which may be related.

bradcray:
In removing enum->int casts for abstract enums in PR #10114, I rewrote
casts for other cases to use a select statement, in part for
simplicity, and in part to generate errors for the semi-concrete case.
This turned out to have a major performance impact for fasta which
relies on a lot of enum->int casts.

Here, I'm optimizing the enum->int cast in the case where an enum is
concrete by having it fall back to the old primitive-based
implementation that used to be implemented in ChapelBase.chpl before
PR #10114.  The semi-concrete case is still handled using a switch
statement.

With more effort, other cases could also be similarlt optimized, but I
think the better use of that effort would be to implement issue #10307
and convert all these functions as lookups into C arrays.  So for now, I'm
stopping here in hopes that it restores the performance we expect for
benchmarks we're currently invested in and can prioritize #10307 if
other cases are required.

Resolves #10303.

### DIFF
--- a/compiler/AST/type.cpp
+++ b/compiler/AST/type.cpp
@@ -387,6 +387,20 @@ bool EnumType::isAbstract() {
   return true;
 }
 
+bool EnumType::isConcrete() {
+  // if the first constant has an initializer, it's concrete;
+  // otherwise, it's not.  This loop with a guaranteed return is a
+  // lazy way of getting that first constant.
+  for_enums(constant, this) {
+    if (constant->init) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+  return true;
+}
+
 
 PrimitiveType* EnumType::getIntegerType() {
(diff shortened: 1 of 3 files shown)


ronawho:
This does resolve the fasta regression (tested for the release version)

PR #10310, opened 2018-07-13
Fix compiler crash in #10177

### LINKED ISSUE #10177: Compiler error when casting c_void_ptr to c_ptr from a function invoked in a forall nested in a on-statement

vasslitvinov:
The newly added test from #10177 generates a previously-unexpected AST
which was guarded against with an INT_ASSERT.

This change handles that AST, continuing to guards against other cases
with INT_ASSERT.

Resolves #10177.


bradcray:
Hi @vasslitvinov — Thanks for this quick fix.  I remain confused about why these SymExpr's don't have parentExpr's set, making me wonder whether they should simply be Symbols rather than SymExpr's.  I was wondering if you could provide more help in understanding the reasons for that for future reference.

vasslitvinov:
Our AST is designed so that some Exprs hang directly off a Symbol and so do not have a parentExpr. The most prominent example is FnSymbol::body. Others include ArgSymbols' DefExprs and defaultExprs.

In general, a SymExpr is used instead of a pointer directly to a Symbol in many contexts, such as an actual or the baseExpr of a CallExpr or the label reference in a GotoStmt. This simplifies the enumeration of a Symbol's uses and the substitution of one Symbol for another.

In the case of the outer variable Symbol, we can design the AST either way. Indeed, ShadowVarSymbol::reduceGlobalOp and FnSymbol::_this are Symbols, not SymExprs. Having ShadowVarSymbol::outerVarSE be a SymExpr rather than a Symbol has worked well for me and I do not see a reason to change this.

PR #10312, opened 2018-07-13
Tolerate Cygwin's different multibyte encoding

dmk42:
Cygwin's native library functions perform a different UTF-8 encoding for codepoints larger than 16 bits.  This isn't so surprising because Windows is native UTF-16, which may affect other choices.

This change skips a portion of a test that expects the native library to yield the reference encoding.  It still expects the code built into Chapel to perform the same encoding everywhere.

### DIFF
--- a/test/io/ferguson/ctests/qio_formatted_test.test.c
+++ b/test/io/ferguson/ctests/qio_formatted_test.test.c
@@ -1508,8 +1508,13 @@ int main(int argc, char** argv)
     qbytes_iobuf_size = sizes[i];
 
     if( 0 == strcmp(codeset, "UTF-8") ) {
+#ifndef __CYGWIN__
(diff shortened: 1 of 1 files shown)


dmk42:
@mppf - Would you mind taking a post-commit look at this?

mppf:
This is OK, but it's weird that Cygwin uses a different strategy.

dmk42:
I don't know why this didn't occur to me before.  It turns out that on Cygwin (because it is on Windows), wchar_t is only 16 bits.  When one of our 32-bit codepoints is assigned to a wchar_t, the upper 16 bits are chopped off, so of course it then encodes to a different UTF-8 multibyte sequence.

Cygwin's wint_t is still 32 bits, because it has to be at least as big as an int.

PR #10313, opened 2018-07-13
Fix assorted typos

cassella:
Fix typos in
compiler/AST/LoopExpr.cpp
compiler/AST/baseAst.cpp
compiler/codegen/type.cpp
compiler/passes/normalize.cpp
compiler/resolution/lowerForalls.cpp
compiler/resolution/virtualDispatch.cpp
runtime/src/tasks/qthreads/tasks-qthreads.c

A comment in normalize.cpp correctly identified the test
"owned-raw-ingored-record.chpl".  Rename the test so I can change the
comment.


cassella:
The renamed test passes.

PR #10317, opened 2018-07-13
Update --force-initializers to apply to internal and standard types

benharsh:
The core of this change involves updates to AggregateType's needConstructor()
and wantsDefaultInitializer() methods such that compiler-generated initializers
will be used for internal and standard types when the --force-initializers flag
is thrown.

The rest of the changes are mostly made to sidestep issues with internal types:
- Special handling of initializers for types with FLAG_NO_OBJECT
- Do not generate initializer for _ref type
- Generate a bit-copy at the start of an extern record's default init
- Do not generate initializers for 'object' and tuple types
- Update _ddata, _ref, and other internal types to always use initializers

Testing:
- [x] local + futures
- [x] gasnet
- [x] --force-initializers (1 failure remaining)
- [x] only remaining constructors are in user/test code


compiler/AST/AggregateType.cpp
```diff
@@ -2159,15 +2165,10 @@ bool AggregateType::needsConstructor() const {
...
 
-  // For now, always generate a default constructor for types in the internal
-  // and library modules
-  if (mod && (mod->modTag == MOD_INTERNAL || mod->modTag == MOD_STANDARD))
-    return true;
-  else if (initializerStyle == DEFINES_CONSTRUCTOR)
+  if (initializerStyle == DEFINES_CONSTRUCTOR)
     return true;
-  else if (fUserDefaultInitializers)
+  else if (fUserDefaultInitializers && this != dtObject)
```
lydia-duncan: I think ideally dtObject would be using initializers as well.  What is the issue that is preventing this?

modules/internal/CPtr.chpl
```diff
@@ -70,6 +70,7 @@ module CPtr {
   pragma "no object"
   pragma "no default functions"
   pragma "no wide class"
+  pragma "use default init"
```
lydia-duncan: I'm assuming there is another type with "no object" applied to it that can't use default initializers.  Otherwise, it seems like "no object" shouldn't be special cased as you have earlier.  Could that type be using "no use default init" instead, so that "no object" doesn't have to be special cased and these three types don't need the additional pragma?

compiler/AST/AggregateType.cpp
```diff
@@ -1782,8 +1782,9 @@ void AggregateType::buildConstructor() {
 }
 
 void AggregateType::buildDefaultInitializer() {
-  if (defaultInitializer                       == NULL ||
-      strcmp(defaultInitializer->name, "init") !=    0) {
+  if ((defaultInitializer                       == NULL ||
+      strcmp(defaultInitializer->name, "init") !=    0) &&
+      symbol->hasFlag(FLAG_REF) == false) {
```
lydia-duncan: I'm confused as to why this case was necessary, since the ref class is now marked a "use default init".

compiler/AST/AggregateType.cpp
benharsh: I think dtObject still needs to have a constructor so that user classes with constructors can compile. When constructors are fully deprecated, dtObject will have an initializer.

lydia-duncan: Oh right, good point!

modules/internal/CPtr.chpl
benharsh: Discussed offline, but responding for posterity: The changes involving FLAG_NO_OBJECT avoid dealing with inheritance and 'super' fields, which don't apply to types with that flag.

compiler/AST/AggregateType.cpp
benharsh: The 'use default init' prevents creation of a constructor for _ref, and this part disables creation of an initializer for _ref. The _ref type is so magic/special that I don't think we should try and treat it like a normal type. It's really a primitive type we happen to describe as a "class", for now.

PR #10318, opened 2018-07-13
Add support for --compiler, --linker, and --ldshared to compileline

### LINKED ISSUE #10206: Make compileline support mode that gets compiler/linker used
Right now, you can get both the compiler and the flags used for it with `compileline --compile-cc` (and `compileline --compile-c++` for the linker and flags), but you can't get just the compiler/linker.  It would be useful to get just the compiler/linker, so that we can store it separately for use when compiling Chapel libraries for use in Python.

This would enable #10054 more cleanly.

lydia-duncan:
Resolves #10206

This will allow us to get just the compiler and linking commands when
interoperating with Cython, which is necessary on Crays to ensure we can access
loaded modules like ugni, for instance.

Tested over a full standard paratest (with futures because why not)

### DIFF
--- a/runtime/etc/Makefile.include
+++ b/runtime/etc/Makefile.include
@@ -110,15 +110,24 @@ printcompileline:
 printcxxcompileline:
 	@echo $(CXX) $(CHPL_MAKE_BASE_CXXFLAGS) $(GEN_CXXFLAGS) $(COMP_GEN_CXXFLAGS) $(CHPL_RT_INC_DIR)
 
+printccompiler:
+	@echo $(CC)
+
 printcflags:
 	@echo $(CHPL_MAKE_BASE_CFLAGS) $(GEN_CFLAGS) $(COMP_GEN_CFLAGS)
 
 printcxxflags:
 	@echo $(CHPL_MAKE_BASE_CXXFLAGS) $(GEN_CXXFLAGS) $(COMP_GEN_CXXFLAGS)
 
+printldshared:
+	@echo $(LD) $(LIB_DYNAMIC_FLAG)
+
 printlibraries:
 	@echo $(GEN_LFLAGS) $(COMP_GEN_LFLAGS) -L$(CHPL_RT_LIB_DIR) \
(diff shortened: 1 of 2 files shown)


runtime/etc/Makefile.include
```diff
@@ -110,16 +110,25 @@ printcompileline:
...
 	@echo $(CHPL_MAKE_BASE_CXXFLAGS) $(GEN_CXXFLAGS) $(COMP_GEN_CXXFLAGS)
 
+printldshared:
+	@echo $(LD) $(LIB_DYNAMIC_FLAG)
+
 printlibraries:
 	@echo $(GEN_LFLAGS) $(COMP_GEN_LFLAGS) -L$(CHPL_RT_LIB_DIR) \
 	      -lchpl $(LIBS) -lm $(CHPL_MAKE_THIRD_PARTY_LINK_ARGS) $(CHPL_MAKE_BASE_LFLAGS)
 
+printlinker:
```
ronawho: Change to `printld` or change the `printldshared` above to `printlinkershared`?

util/config/compileline
```diff
@@ -69,10 +72,16 @@ def parseArguments():
...
                       help="Print -D and -I compilation flags to include "
                            "the Chapel runtime")
+    parser.add_option("--ldshared", const="ldshared",
+                      dest="actions", action='append_const',
+                      help="print out the command for ldshared")
     parser.add_option("--libraries", const="libraries",
                       dest="actions", action='append_const',
                       help="print out -L and -l options to link with "
                            "the Chapel runtime")
+    parser.add_option("--linker", const="linker",
```
ronawho: Same comment I made in the makefile, it'd be nice to unify on `linker` or `ld`

lydia-duncan:
Passed testing with the previous change, and I double checked that hello.chpl still built after the latest commit so I'm not going to re-run the full paratest.  If that bites me, I will be sad

mppf:
I have no concerns - looks good to me!

PR #10319, opened 2018-07-13
Fix .bad mismatch from #10300

vasslitvinov:
Run this .future with --verify so it matches .bad.

Skip it in multilocale and some other configurations
because the current failure mode is independent of
most configuration details.


lydia-duncan:
Maybe add a note to the .future file to remove the .skipif once it passes?

PR #10320, opened 2018-07-13
Refresh futures relating to `enum`s

### LINKED ISSUE #9222: Review enum futures
Given the recent changes with enums (e.g., PR #9192) and the hope to finalize them from a Chapel 2.0 sense (issue #9022), this seems like a good time to review enum-related futures and determine whether they are still valid and/or will fall out easily as a result of recent work.  If they are still valid, make sure there is an issue filed against them when appropriate.

bug: enum array field not copied appropriately when record has default initializer
feature request: external enums
unimplemented feature: compile time enum->string conversion
feature request: enumerate iterator similar to python
feature request: enums should be usable like modules
bug: function-scoped enum can't be initialized with global param
bug: function-scoped enum can't be initialized with int-typed function param
bug: function-scoped enum can't be initialized with function param, int-casted
bug: function-scoped enum can't be initialized with function param
feature request: Compiler warning/error if enum not covered by select.
unimplemented feature: allow private enum declarations

bradcray:
In this PR, I'm taking a pass over refreshing any futures which relate to `enum`s to make sure they're up-to-date as I review them.  Happily, the result was that three futures were removed (no longer relevant) and two were retired (working but not detected as such).  I made minor edits to two others.

Removed
----------
* `constructors/assign-param-segfault.chpl` is specific to constructors
  which we're leaving in our wake, so remove the test
* `types/enum/lydia/declaredInClass-constructor.chpl`: ditto
* `functions/diten/param_enum_to_string.chpl` assumed we support enum->string coercions which we've decided not to.  Removed as suggested by @daviditen 

Retired
-------
* `initializers/assign-param-segfault.chpl` works, but didn't match its
  `.good` file because we stopped printing out `param`/`type` fields for
  generic classes at some point in the past and this file didn't have
  a `.bad` file.
* `types/enum/bradc/useEnum.chpl` works but had a bad `.good` file
  and no `.bad` file.

Refreshed
-----------
* `extern/bradc/types/enum.chpl`: updated .future and added a .bad file
* `users/kreider/feature_select.chpl`: updated .future

Resolves #9222.


bradcray:
At the time that I'm deleting this branch, there's a bunch of noise in it due to GitHub being weird today:

screen shot 2018-07-16 at 12 21 08 pm

So far, there's no indication that any of this affected the real "git" itself, though.

vasslitvinov:
That wasn't me I swear. I was out for lunch! :)

Issue #10304, opened 2018-07-12
Codegen compiler performance regression from #10187


ronawho:
There was a ~2x slowdown for codegen as a result of "Refactoring codegenPrimitive, and adding a limit to denormalize" (https://github.com/chapel-lang/chapel/pull/10187)

Here's some [graphs](https://chapel-lang.org/perf/chap01/?startdate=2018/07/06&enddate=2018/07/11&graphs=topaveragecompilationtimeperpass)

![image](https://user-images.githubusercontent.com/1588337/42662809-d83740ee-85e7-11e8-89f1-872d0b41fd1b.png)

Note that there was a huge performance improvement for resolution on the same day as a result of https://github.com/chapel-lang/chapel/pull/10143, but it'd still be nice to see if we can resolve the codegen regression.

I verified the cause of the regression by running test/performance/compiler/bradc/cg-sparse-timecomp.chpl on /tmp of chapcs07, but it should be reproducible anywhere.

bradcray:
In reading about this PR, I was curious about the impact of using function pointers, but had hoped it would be smaller than this.

I found myself wondering whether, having introduced the per-primitive functions, using a switch statement to call into them rather than the function pointers would get us most of the benefit of the cleanup—while the switch statement would still exist, at least most of the code would be moved out of it, resulting in a cleaner result.

@benharsh wondered whether, through counting, we could determine whether there was a small set of primitives that were used the most and if just special-casing those in a switch statement (or otherwise) would get us most of the performance back without requiring the full switch statement.  This seems practical, though also a little unfortunate (when adding a new primitive, one would have to ask which approach to take and maybe be confused about the presence of both).

mppf:
@ronawho - thanks for noticing the problem. Actually simply moving the "register all of the primitive codegen fns" code out of codegenPrimitive resolves the performance issue. So @bradcray's hopes can be restored.

ronawho:
Looks like #10321 resolved this (and codegen is actually a little faster than it was before!):

![image](https://user-images.githubusercontent.com/1588337/42771359-d1d9f7d6-88db-11e8-98d0-f7f67f6692cb.png)

PR #10187, opened 2018-07-06
Refactor codegenPrimitive, add limit to denormalize

mppf:
In issue #9047, the compiler was crashing due to running out of stack 
space.  Investigating revealed that denormalize was causing deeply nested
CallExprs and codegenPrimitive used 13KB of stack space per invocation.

This PR resolves this problem in two ways:
 1. It adjusts the denormalize pass to give up after a certain number of
    denormalized variables are collected.
 2. It replaces the ~2000-line switch statement in codegenPrimitive with 
    small functions and calls to these small functions through function
    pointers.

Implementing the 2nd of these involved several steps:
 * move comments describing primitives in primitive.h to primitive.cpp,
   so that the comments will be in one place and in preparation for the
   next step
 * moved the primitive names to primitive_list.h and enable including
   this file to drive macros similarly to flags_list.h. There are two
   macros that can be defined when using this file - one for primitives
   that exist at code generation and one for other "resolution-time"
   primitives.
 * Used this and other macros to reduce typing to define a static method
   in CallExpr for each primitive that needs code generation
 * Used the primitive_list macro to register the appropriate function
   pointer for each primitive once in codegenPrimitive - along with this
   added this codegen function pointer to the PrimitiveOp in primitive.h
   and added registerPrimitiveCodegen function.
 * moved the code from the switch statement in codegenPrimitive into the
   various functions. The code here is copy-pasted but required minor
   modification since the new functions aren't methods, so some methods
   needed a receiver e.g. `get(1)` became `call->get(1)`.

Resolves #9047
Reviewed by @benharsh - thanks!

- [x] full local testing

PR #10321, opened 2018-07-13
Improve the speed of codegenPrimitive

mppf:
This PR hopefully addresses the compiler performance regression from
PR #10187 noted in issue #10304. It improves the speed of
codegenPrimitive by:

* registering the primitive codegen handlers in another function (called at the
  start of code generation)

While there, it also makes these changes:

* simplified the structure of the function, to consolidate cases and use
  conditionals instead of a switch

* the first conditional is for PRIM_MOVE because that's the most common
  primitive (by far). Having a conditional here is intended to help with branch
  prediction & allow inlining codegenPrimMove if the C++ compiler wants to do
  that. Note though that in my experiments I didn't see any noticeable
  performance difference from this on my machine. Handling PRIM_MOVE specially
  doesn't seem problematic though (since it already has codegenPrimMove to call)
  and it might help on other hardware.

The most common primitives for Hello World are, in order: move, set reference,
cast, check_nil, return. Move occurs about 5x more than set reference.

- [x] full local testing

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -3499,12 +3499,17 @@ GenRet CallExpr::codegen() {
   return ret;
 }
 
+// to define a primitive's code generation method
 #define DEFINE_PRIM(NAME) \
   void CallExpr::codegen ## NAME (CallExpr* call, GenRet &ret)
 
+// to call a primitive's code generation method
+#define CODEGEN_PRIM(NAME, call, ret) \
+  codegen ## NAME (call, ret);
+
 // to call another primitive's DEFINE_PRIM
 #define FORWARD_PRIM(NAME) \
-  codegen ## NAME (call, ret);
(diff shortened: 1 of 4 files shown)


compiler/codegen/expr.cpp
```diff
@@ -4952,53 +4957,49 @@ DEFINE_PRIM(PRIM_LOOKUP_FILENAME) {
     ret = call->codegenBasicPrimitiveExpr();
 }
 
-
-GenRet CallExpr::codegenPrimitive() {
-  SET_LINENO(this);
-
-  static bool codegenPrimitivesRegistered = false;
-  if (!codegenPrimitivesRegistered) {
```
cassella: Was the problem with the original approach just that nothing ever set `codegenPrimitivesRegistered = true`?

mppf: Looks like it - maybe you should have reviewed it :P

PR #10328, opened 2018-07-13
First draft implementation of enumerate() iterator

bradcray:
Implement enumerate() as Thomas proposes in a new package module
called Pythonic.

Move Thomas's test to the library/ directory, updating it to:

* avoid race conditions in the output by writing to an array and then
  printing the array serially,
* test default vs. non-default numbering
* expect 0-based indexing by default

This is linked to issue #10329


e-kayrakli:
Python's `enumerate` yields `(index,item)`. Is there any particular reason this implementation yields `(item,index)`?

bradcray:
Oh does it?  I was basing my implementation on the `.good` file that @thomasvandoren filed for this feature rather than reading the documentation (in spite of linking to the documentation in the related issue and reading enough to see that it started counting at 0).  Ooops.  Thanks for the catch.

bradcray:
> Python's `enumerate` yields `(index,item)`. Is there any particular reason this implementation yields `(item,index)`?

This should be fixed now, thanks for catching the mistake!

mppf:
@bradcray - I'm just trying to help progress on some open PRs. What's the next step on this one?

bradcray:
I'm closing this PR for now because I never have time to work on it (it's not sufficiently high priority) and it wasn't the slam dunk I'd hoped for at the time due to design questions.  I've mentioned it on issue #10329 so that we can find our way back here the next time we want to work on it.

modules/packages/Pythonic.chpl
```diff
@@ -0,0 +1,63 @@
...
+Python programmers transition to Chapel or feel more comfortable in
+it.
+
+*/
+
+module Pythonic {
+  
+  /* Iterates over `iterable`, yielding 2-tuples in which the first
+     element is a value from `iterable` and the second is its
+     corresponding integer in the range ``start..``.  This iterator
```
cassella, line 35: The doc comment also needs updating for the swapped yield order here, and in two more lines below.

modules/packages/Pythonic.chpl
```diff
@@ -0,0 +1,63 @@
...
+  */
+
+  iter enumerate(iterable, start=0) {
+    for (x, i) in zip(iterable, start..) do
+      yield (i,x);
+  }
+
+  pragma "no doc"
+  iter enumerate(param tag: iterKind, iterable, start=1)
+   where tag == iterKind.standalone {
```
cassella, line 59: The standalone iterator should probably have the same default for `start` as the serial one (and the doc comment).

PR #10332, opened 2018-07-13
Stop inserting declarations for certain internal exported functions in the .h

lydia-duncan:
Prior to this change, the generated header file in library mode would contain
declarations for exported functions from our internal modules, as well as the
generated "main" function.  This updates the generation of the header file to
explicitly skip those functions.

Spot checked a couple of the generated header files (since we no longer commit
them to the repository), and passed a full paratest


lydia-duncan:
@bradcray or @mppf - care to review?  I don't intend to merge until Monday so no rush

PR #10343, opened 2018-07-15
Add issue numbers to detupleFieldDecl future

cassella:
Add issue numbers to
test/types/records/tupleFields/detupleFieldDecl.future


bradcray:
This looks fine to me, but doesn't seem like something I'd need to review.

FWIW, I'm not the biggest stickler about putting issue #'s into .future files because (a) it's often hard to predict them when filing the future, (b) we don't have scripts to check them, (c) we have lots of futures that don't have issue #'s in them (even for cases that do have associated issues).  But I know others feel more strongly about putting them in...

PR #10346, opened 2018-07-16
Convert halts in I/O

mppf:
Convert halts in IO.chpl to halt wrapper call, assert, or throw

- [x] full local testing

### DIFF
--- a/modules/standard/IO.chpl
+++ b/modules/standard/IO.chpl
@@ -1530,8 +1530,7 @@ proc _modestring(mode:iomode) {
     when iomode.cw do return _cw;
     when iomode.cwr do return _cwr;
     otherwise {
-      use HaltWrappers;
-      exhaustiveSelectHalt("Invalid iomode");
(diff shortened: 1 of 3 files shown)


modules/standard/IO.chpl
```diff
@@ -1526,7 +1529,10 @@ proc _modestring(mode:iomode) {
     when iomode.rw do return _rw;
     when iomode.cw do return _cw;
     when iomode.cwr do return _cwr;
-    otherwise halt("Invalid mode");
+    otherwise {
+      use HaltWrappers;
+      exhaustiveSelectHalt("Invalid iomode");
```
ronawho: In https://github.com/chapel-lang/chapel/pull/10327 we switched over to fully qualified names, which I think is a nice cleanup. Switch to 
```chpl 
otherwise do HaltWrappers.exhaustiveSelectHalt("Invalid iomode");
```
?

PR #10354, opened 2018-07-16
Convert various tests to use initializers

benharsh:
This PR updates various tests to use explicit initializers instead of constructors. Some futures were resolved by using initializers. Some tests were removed entirely because they tested a bug in constructors, which are going away soon.

Testing:
- [x] local + futures
- [x] gasnet
- [x] --force-initializers


test/classes/initializers/assign-param-segfault.chpl
```diff
@@ -1,19 +0,0 @@
...
-      when Foo.Foo1 do this.baz = Baz.Baz1;
-      when Foo.Foo2 do this.baz = Baz.Baz2;
-      when Foo.Foo3 do this.baz = Baz.Baz3;
-      otherwise this.baz = Baz.Baz0;
-      }
-  }
-}
-var bar = new Bar(Foo.Foo1);
-writeln(bar);
-delete bar;
```
vasslitvinov: Please explain in the PR message why this test is going away.

It does not seem to fall in the category "tested a bug in constructors" (even though it may do).

Ditto `initializeMethod.chpl` below.

test/classes/initializers/generics/varargs-type.chpl
```diff
@@ -0,0 +1,14 @@
...
+  
+  proc init(args... , type T = args.type) {
+    this.T = T;
+    this.args = args;
+    writeln(T:string);
+  }
+}
+
+var c = new C(1, 2, 3);
+delete c;
```
vasslitvinov: I think `c` should be `new unmanaged`, or better yet `new owned` ?

test/distributions/diten/blockCyclicDist.chpl
```diff
@@ -52,8 +52,8 @@ class BlockCyclicDom {
   type idxType;
   const whole: domain(nDims, idxType);
   const dist: unmanaged BlockCyclicDist(idxType);
-  const locDoms: [dist.localeDomain] unmanaged LocBlockCyclicDom(nDims, idxType);
-  proc initialize() {
+  var locDoms: [dist.localeDomain] unmanaged LocBlockCyclicDom(nDims, idxType);
```
vasslitvinov: How about adding a comment like `locDoms is 'var' because it is set in postinit()` ?
Ditto `locArrs` below.

test/functions/iterators/bharshbarg/recurseHeapBug.chpl
```diff
@@ -1,7 +1,7 @@
 
 record R {
   var x : int;
-  proc R() { }
+  proc init() { }
```
vasslitvinov: How about remove init() altogether? While it does not matter for this test, it will spare future readers the question "why is this initializer here?".

test/classes/initializers/generics/varargs-type.future
```diff
@@ -0,0 +1,3 @@
+bug: Unresolved call for initializer that uses type of varargs formal
+
+Based on 'test/functions/nspark/generic-varargs2.chpl'
```
vasslitvinov: Please explain in this .future how this test differs from nspark's and/or what testing value this test adds.

test/types/records/kbrady/return-record.chpl
```diff
@@ -1,63 +0,0 @@
...
-    writeln(z2);
-  }
-
-  /////////////////////////////////
-
-  proc main() {
-    try_baz();
-    try_gob();
-  }
-}
```
vasslitvinov: Maybe it is good to give, in the PR message, a reason for removal for each test individually. Otherwise it keeps me guessing. Thank you.

test/users/thom/arrInClass-intmember.future
```diff
@@ -1,11 +0,0 @@
...
-
-This test gets an "zippering unequal lengths" error message, presumably
-because the default constructor sets the array member up to be a 0
-element array and we then try to assign a 3-element array into it in the
-user constructor.  This semantic future asks whether that is what we
-wish to happen given that the visible user code would seem to be legal
-(if the default "safe" initializations were not quietly happening first).
-
-This seems like yet another one of those "wish we had a way for an array
-member to take its size from the thing being assigned to it" cases...
```
vasslitvinov: Cool to un-futurize this!

Feel free to add a brag to the PR message.

test/users/vass/km/array-of-records-crash-2.chpl
```diff
@@ -5,8 +5,10 @@ const vertex_domain = {1..N_VERTICES} ;
 
 record vertex_struct {
 
+  proc init() { }
+
   // constructor
-  proc vertex_struct(nd: domain(1)) {
+  proc init(nd: domain(1)) {
```
vasslitvinov: Let's remove the "// constructor" comment above. It does not add value.

test/users/vass/km/array-of-records-crash-2.explicit-type.chpl
```diff
@@ -7,8 +7,10 @@ const vertex_domain = {1..N_VERTICES} ;
 
 record vertex_struct {
 
+  proc init() { }
+
   // constructor
-  proc vertex_struct(nd: domain(1)) {
+  proc init(nd: domain(1)) {
```
vasslitvinov: Here, too - let's remove the "// constructor" comment above. It does not add value.

test/classes/initializers/assign-param-segfault.chpl
benharsh: Oops, I think I meant to remove the variant of this test under constructors/ , I'll add it back in.

test/distributions/diten/blockCyclicDist.chpl
benharsh: That seems unnecessary to me, I think readers can easily determine where these fields are set.

test/classes/initializers/generics/varargs-type.future
benharsh: Hmm, not sure what I was thinking... I'll remove the new test.

Issue #10120, opened 2018-07-02
build_configs.py changes (new scripts to build customizable Cray module from source)


awallace-cray:
Develop the modifications to [build_configs.py](https://github.com/chapel-lang/chapel/blob/master/util/build_configs.py) needed to work with the new "Chapel Cray module build script"

This is the 1st item in https://github.com/chapel-lang/chapel/issues/10046#issuecomment-401422989, new scripts to build customizable Cray module from source

The deliverable is a working script suitable for stand-alone use by an outside end-user, as described in https://github.com/chapel-lang/chapel/issues/9551. 

The deliverable is also intended for use in Chapel's  CI automation, ultimately replacing the current Chapel Cray module build scripts used in the Cray-internal Jenkins. However, this task does not include final implementation and testing for that application.

awallace-cray:
2018-07-23 10:00 Status for planning. My only comment is this #10429 (comment). Right now, it seems optional or at least defer-able

PR #10360, opened 2018-07-17
build_configs.py changes (new scripts to build customizable Cray module from source)

awallace-cray:
Issue #10120 build_configs.py changes 
(new scripts to build customizable Cray module from source)

* Move build_configs.py to new folder "build_configs". This
  will help users manage their related working files, like setenv
  scripts and log files.

* Many changes in the way Chapel build config parameters
  are handled in build_configs.py:
  * All known Chapel build config parameters are recognized
    (ie, all "CHPL_XXX" config variables generated by
    "printchplenv --all --no-tidy" as of this date).
    That's about twice as many "build dimensions" as before.
  * build_configs.py no longer tries to:
    - validate build config parameters provided by the user
    - provide meaningful default values for build config
      parameters
    - provide meaningful "help" text describing build config
      parameters.
  * Any or all Chapel build config parameters may have zero
    non-null values assigned. When that happens, the
    corresponding "build dimension" is effectively undefined.
    Ie, that Chapel build config parameter value is left for
    Chapel make to decide.
  * All Chapel build config parameters initially default to a 
    value of "None" (ie, null; empty).
  * On startup, build_configs.py scans the existing CHPL_XXX
    environment variables (if any):
    - Any existing CHPL_XXX env value becomes the default
      value of the corresponding Chapel build parameter, 
      before any commandline options are scanned.
    - If "UNSET" is given as a build parameter value
      (on the the commandline), then the corresponding CHPL_XXX
      env variable will be silently unset (as in bash "unset"),
      for the corresponding Chapel make run.

* Extra make parameters (options or targets) may now be
  added to the end of the build_config.py commandline
  (after all the "-x/--xxxx>" options). If found, these will
  simply be added to the generated Chapel make commands.

* New commandline parameter "-s/--setenv" may specify a bash-
  compatible script file containing arbitrary commands.
  If found, this setenv script will be source'd before each
  Chapel make is run.
  This is typically used to modify the O.S. environment seen
  by the Chapel make. Setenv scripts typically use the CHPL_XXX
  env variables defined through build_config.py, to decide
  what to do.
  Additional env variables with names like "BUILD_CONFIG_xxx"
  are also defined for each Chapel make run. These env vars
  are ignored by Chapel, but may be used by a setenv script;
  for example, to tell whether the setenv script was
  run from within build_configs.py, or not.

* New commandline parameter "-l/--make-log" may redirect the
  stdout/stderr from Chapel make commands to a log file. If no
  make-log file is defined, the Chapel make output is discarded.

* The final exit code of build_configs.py is now just the max
  exit code value from all the Chapel build configs (makes)

* A build_config.py "--dry-run" now runs printchplenv, and the
  results are checked for warnings or errors.
  Setenv scripts, if any, will be honored in dry-runs; however,
  any "--make-log" directives will be ignored.


awallace-cray:
The gitignore file in this subdir may not make sense, and I'm open to just removing it. 
I wanted to allow users to park "working" files for their build_configs projects here (site-specific setenv scripts, temporary log files, etc), and still be able to git clean.  
I momentarily forgot that git clean wipes out anything not checked-in, whether git-ignored or not. 
If you want to put a git clean in your build script, you will have to keep your local working files outside the chapel git workspace that you are building.

awallace-cray:
Closing this in favor of #10429, which includes these commits. There's nothing wrong with merging this PR, then rebasing the newer one. However, if no one else has worked on this one, I just don't see the point.

PR #10363, opened 2018-07-17
Separately traverse point of instantiation in getVisibleFunctions

### LINKED ISSUE #10245: new borrowed and lifetime checker confused by explicit module access
```
chpl --lifetime-checking test/classes/moduleScope/mod-init-borrowed.chpl
```

This test includes code like
``` chapel
proc main() {
  var a   = new borrowed A(10);

  writeln();

  var m1a = new borrowed M1.A(20);
  var m2a = new borrowed M2.A(30);

  writeln();

  var m3a = new M3.A(40);
  var m4a = new M4.A(50);
}
```

Since commit https://github.com/chapel-lang/chapel/commit/785ab658f1423e108e91e61a332529fdfe9d22a0, the compiler has added a BlockStmt around these explicit module accesses in the normalize pass. Since this BlockStmt is marked a normal block, the block's existence impacts the lifetime checker's view of lifetimes. Additionally, it impacts the actual lifetime of variables... the `owned` value created for `new borrowed` is destroyed here.

mppf:
This PR adjusts getVisibleFunctions to separately traverse the
instantiation point instead of including the instantiation point in a
"chain" of visibility.

* Removes insertUseForExplicitModuleCalls
* removes FnSymbol::numPreTupleFormals (it is unused)
* use BLOCK_SCOPELESS in resolveUninsertedCall since these blocks are not
  intended to impact scoping
* replace getVisibilityBlock with two functions - getInstantiationPoint
  and getVisibilityScope. getVisibilityScope does not consider the point
  of instantiation.
* adjust getVisibleFunctions to be more reasonable and more maintainable.
  To that end, the visibilityBlockCache is removed. As far as I could
  see, it did not offer much improvement compile-time and it made
  getVisibleFunctions significantly more complex.

Resolves #10245.

- [x] full local testing


compiler/resolution/visibleFunctions.cpp
```diff
@@ -205,14 +195,61 @@ static BlockStmt* getVisibleFunctions(const char*           name,
...
+    if (block->parentExpr != NULL) {
+      // not a module or function level block
+    } else if (inMod && block == inMod->block) {
+      moduleBlock = true;
+    } else if (inFn != NULL) {
+      // TODO - probably remove this assert
+      INT_ASSERT(block->parentSymbol == inFn ||
+                 isArgSymbol(block->parentSymbol) ||
+                 isShadowVarSymbol(block->parentSymbol));
+      // Could also check
```
vasslitvinov: The comment above sounds like an incomplete sentence. What should it be?

compiler/resolution/visibleFunctions.cpp
```diff
@@ -205,14 +195,61 @@ static BlockStmt* getVisibleFunctions(const char*           name,
...
+      // block that isn't a fn or module block.
+      // But, we could probably optimize by knowing:
+      //   - which blocks are irrelevant
+      //   - which blocks are traversed as outer blocks
       visited.insert(block);
     }
 
     if (VisibleFunctionBlock* vfb = visibleFunctionMap.get(block)) {
-      canSkipThisBlock = false; // cannot skip if this block defines functions
+      // note, defines functions
```
vasslitvinov: Should it be colon, not comma, in the above comment ?
Or maybe convey the point differently?
Ditto a similar comment you inserted below.

compiler/resolution/visibleFunctions.cpp
```diff
@@ -62,6 +62,8 @@ static Map<BlockStmt*, BlockStmt*>            visibilityBlockCache;
```
vasslitvinov: Is visibilityBlockCache still used?

If not, would you add a note to the PR message why we are better off without it.

mppf: It didn't seem to have a significant impact on performance, and I think simpler is better unless we see measurable benefit.

mppf:
@lydia-duncan - FYI, I think this PR should help with some issues with `private`.

lydia-duncan:
Nice, thanks!

PR #10366, opened 2018-07-17
Remove argv from 'chpl_main_argument' in module definition

benharsh:
'argv' is a char**, not a _ddata(string). This would result in a failure in chpl_main_argument's default initializer if denormalization was disabled. Since the field is not used in the module code, I opted to update the type and comment it out in order to indicate that there are not just two fields in the struct.

Testing:
- [x] local + Futures
- [x] baseline

### DIFF
--- a/modules/internal/ChapelUtil.chpl
+++ b/modules/internal/ChapelUtil.chpl
@@ -101,7 +101,7 @@ module ChapelUtil {
   pragma "use default init"
   extern record chpl_main_argument {
     var argc: int(64);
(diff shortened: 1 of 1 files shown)


lydia-duncan:
Do you think this update should be reflected in the library initialization function, since that also takes an argv with a similar purpose/use?

benharsh:
Can you link to the code you're talking about?

lydia-duncan:
I think I probably got confused, sorry!

Issue #10104, opened 2018-06-29
Cryptic compiler error for default value with dynamic owned type


benharsh:
### Summary of Problem
The following program fails to compile with the error message:
```
foo.chpl:16: error: unresolved call 'something_default_bar(_owned(C))'
foo.chpl:16: note: candidates are: something_default_bar(foo: _owned(B))
```

The 'something_default_bar' function is generated by the Chapel compiler and builds the default value for the 'bar' formal.

My understanding is that because the formals use default intent, this program is expected to work.

### Steps to Reproduce

**Source Code:**

```chapel
class B {
  proc method() {
    writeln("B.method");
    return 5;
  }
}
class C : B {
  proc method() {
    writeln("C.method");
    return 100;
  }
}

proc something(foo : owned B, bar = foo.method()) {
  writeln(bar);
}

var c = new owned C();
something(c);
```

benharsh:
@mppf : here's the bug I was talking about on IRC.

This bug causes a failure under --force-initializers for the following test:

``classes/bradc/overloadMethods/virtualIsHalt.chpl``

I don't think it's specific to initializers. Rather, constructors use an old default-value implementation, whereas initializers use the more general implementation like the ``something`` function.

benharsh:
I stand corrected, I think this is entirely specific to owned.

The situation is that the actual is ``_owned(C)`` and the formal is ``_ref(_owned(B))`` with a const-ref intent. The ``_ref`` type is important here.

The first hurdle is that ``canCoerce`` returns false, leading to the 'unresolved call' error. It looks like part of canCoerce is specialized for managed types, and probably needs to use ``getValType`` in a few places.

The second problem is that ``needToAddCoercion`` returns true if the actual is ``T`` and the formal is ``_ref(T)``, but that's easily worked-around. This case applied to some other code, not the default-function.

Finally (as far as I got) there's an issue with casting, resulting in the error:
``value from coercion passed to const ref formal 'foo'``

The AST at this point looks like this:
```
def _formal_tmp_x : _owned(B)
move _formal_tmp_x (call chpl__initCopy(c))
move coerce_tmp (call _cast _ref(_owned(B)) _formal_tmp_x)
move default_arg_bar (call something_default_bar coerce_tmp)
```

And it's an error to pass the result of a cast to a ref formal.


I think what we want is to pass _formal_tmp_x to the default_arg function.

bradcray:
> This bug causes a failure under --force-initializers for the following test:
[...]

PR #10368, opened 2018-07-17
Switch test to `unmanaged` for initializers; file future against challenges with `owned`

bradcray:
`virtualIsHalt.chpl` is one of the few tests that's not working with
`--force-initializers`, due to an issue related to its use of `owned`
and default field values for dynamic  owned types (issue #10104).
Here, I'm changing the test to use `unmanaged` to keep it working
once initializers are the default (since it wasn't designed to exercise
issues with `owned`).

Also adding the code from issue #10104 as a future in order to capture
it as a more succinct example.


bradcray:
@benharsh: Would you review?  Maybe look particularly at the `.good` file for the future to make sure it's as you'd expect?

bradcray:
PS -- I'll update issue #10104 to point to the future after merging.

PR #10371, opened 2018-07-17
Fix variable initialization with qualified type and new expressions

### LINKED ISSUE #8555: Internal error for fully qualified `new` expressions with type specifiers

benharsh:
The simple pattern "var r : A.R = new A.R(1);" was unhandled by normalization and encountered an internal error, but leaving the type expression off was handled successfully. This PR copies the part of normalization that handles type inference from a new-expr and uses the same logic for the problematic pattern.

Resolves #8555

Testing:
- [x] local + futures


compiler/passes/normalize.cpp
```diff
@@ -2401,6 +2401,17 @@ static void normVarTypeWithInit(DefExpr* defExpr) {
     } else {
       Expr*     arg     = toCallExpr(initExpr)->get(1)->remove();
       CallExpr* argExpr = toCallExpr(arg);
+      SymExpr*  modToken = NULL;
+      SymExpr*  modValue = NULL;
+
+      if (argExpr->numActuals() >= 2) {
+        if (SymExpr* se = toSymExpr(argExpr->get(1))) {
+          if (se->symbol() == gModuleToken) {
+            modToken = toSymExpr(argExpr->get(1)->remove());
```
lydia-duncan: I would be tempted to just say `modToken = se->remove();`, since you already have it.  I'd also leave a comment on the modValue assignment reinforcing that the result of get(1) there is different than the value returned earlier (or just swap the order of the assignments so that you get(2) and then remove the first argument).

PR #10372, opened 2018-07-17
Propagate forwarded type instantiation point to scratch fn

### LINKED ISSUE #10358: Nonsensical 'type mismatch' in array assignment
When attempting to assign to an array, sometimes I get some nonsensical 'type mismatch' output from the compiler. This occurs in the `DistributedBag` module as well as in a new work queue I created for the sole purpose of side-stepping this issue. I know that I created the `DistributedBag`, and I also created this new work queue, but I'm kind of stuck due to whatever the issue is. The line of code that causes the issue is [here](https://github.com/chapel-lang/chapel/blob/master/modules/packages/DistributedBag.chpl#L657), which is a simple array assignment. While the `DistributedBag` uses a `c_ptr`, in my work queue I use a Chapel array, so it isn't a problem that is unique to `c_ptr`.

**MWE**

```chpl
use DistributedBag;
record R {}
var bag = new DistBag(R);
bag.add(new R());
```
[TIO](https://tio.run/##S85ILEjN@f@/tDhVwSWzuKQoM6m0JDXFKTHdmqsoNTm/KEUhSKG6lqsssUghKTFdwVYhL7UcrBKoRCNI05oLKKqXmJKiARIP0tDUtP7/HwA)

**Output**

```
$CHPL_HOME/modules/packages/DistributedBag.chpl:657: error: type mismatch in assignment from R to R
```

**Edit:**

Turns out that this bug also effects `DistributedDeque` as well...

**MWE**

```chpl
use DistributedDeque;
record R {}
var deque = new DistDeque(R);
deque.add(new R());
```
[TIO](https://tio.run/##S85ILEjN@f@/tDhVwSWzuKQoM6m0JDXFJbWwNNWaqyg1Ob8oRSFIobqWqyyxSCEFJKxgq5CXWg5WDVamEaRpzQWW0UtMSdEAyQVpaGpa//8PAA)

**Output**

```
$CHPL_HOME/modules/packages/DistributedDeque.chpl:819: error: type mismatch in assignment from R to R
```

mppf:
Resolves #10358 

Resolution for generic forwarded functions are currently added to a "scratch 
function" to be resolved. This was causing problems when these functions used
the point-of-instantiation rule to find relevant functions that are called. 
This PR adjusts the scratch function to have a point-of-instantiation based 
upon the relevant AggregateType (on which forwarding is occurring).

This change makes the pattern of forwarding in cg-sparse-partred.chpl no longer work, since in that case the scratch function would need a different 
instantiation point. In particular, cg-sparse-parted expects that a method only
available at the instantiation of the call (vs at the creation of the type) is
available to be called. I'm looking at adjusting the forwarding implementation 
to not have the scratch function but I think the change in cg-sparse-parted is
minor and acceptable to allow the more common patterns.

- [x] full local testing

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -3068,6 +3068,9 @@ static bool populateForwardingMethods(CallInfo& info) {
       FnSymbol* scratch = new FnSymbol("delegate_scratch_fn");
       scratch->addFlag(FLAG_COMPILER_GENERATED);
 
+      if (at->symbol->instantiationPoint)
+        scratch->instantiationPoint = at->symbol->instantiationPoint;
+
       Expr* where = getInsertPointForTypeFunction(at);
       if (BlockStmt* block = toBlockStmt(where))
         block->insertAtHead(new DefExpr(scratch));
--- a/test/reductions/partial/CSimpl.chpl
+++ b/test/reductions/partial/CSimpl.chpl
@@ -12,7 +12,7 @@ proc plusPRinto(ref RES,DOM,FEXPR) throws {
   defer {
     delete OP;
   }
-  DOM.dist.dsiPartialReduceInto(RES, OP, DOM, FEXPR);
+  DOM.dist._value.dsiPartialReduceInto(RES, OP, DOM, FEXPR);
 }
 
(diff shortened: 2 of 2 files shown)


mppf:
There is one failure in testing -- test/npb/cg/bradc/cg-sparse-partred.chpl

Issue #10260, opened 2018-07-10
Spike: Spack Prototype


Spartee:
If Mason were going to use Spack as the main external dependency manager, how would Mason users specify and use Spack dependencies in their mason packages? Completion of this spike includes a working prototype of a Mason package that uses a Spack dependency.

Spartee:
A working prototype has been built that allows Mason users to download, install and use dependencies through spack. Mason, in this prototype, uses Spack as a backend for dependency management. Other convenience functions are included in the prototype such as compiler installation, help, and find; package search, installation check, and list. 

A change to note is that the `pkg-config` integration in #10289 will be listed under `[system]` and external packages such as Spack packages will be listed under `[external]`. This change is to make the commands, `mason external` and `mason system`, more intuitive for users.

PR #10383, opened 2018-07-18
Mason: Add/Remove

Spartee:
This PR adds the functionality for users to be able to add and remove dependencies from their Mason.toml from the Mason CLI.

The following functionality has been added:
- Add/Remove Mason dependencies
- Add/Remove external dependencies
- Add/Remove system dependencies

This PR is blocked by, at very least, the pkg-config integration(#10289), but adding external dependencies won't do anything until my External support is merged. see #10260


tools/mason/MasonHelp.chpl
```diff
@@ -132,6 +132,24 @@ proc masonSearchHelp() {
...
+proc masonModifyHelp() {
+  writeln("Modify a Mason package's Mason.toml");
+  writeln();
+  writeln("Usage:");
+  writeln("    mason rm [options] package");
+  writeln("    mason add [options] package@version");
+  writeln();
+  writeln("Options:");
+  writeln("    -h, --help                  Display this message");
+  writeln("        --external              Add/Remove dependency from external dependencies");
```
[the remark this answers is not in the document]
Spartee: I think especially with `mason external add <package>` users might confuse the meaning behind the action, and think that the command is also installing the package. I am not opposed to adding this, however. It would be fairly easy to do.

tools/mason/MasonHelp.chpl
```diff
@@ -132,6 +134,24 @@ proc masonSearchHelp() {
   writeln(desc);
 }
 
+proc masonModifyHelp() {
+  writeln("Modify a Mason package's Mason.toml");
+  writeln();
+  writeln("Usage:");
+  writeln("    mason rm [options] package");
+  writeln("    mason add [options] package@version");
```
ben-albrecht: Just noticed this with `mason {add,rm}`, but it's an issue across many commands: `mason <subcommand> --help` outside of a mason project results in:

```
Mason could not find your configuration file (Mason.toml)
```

Users should definitely be able to view help output outside of a mason project.

If this turns out to require a significant change to fix, I'd be OK with documenting this an issue and taking it 
 in separate this PR.

tools/mason/MasonHelp.chpl
```diff
@@ -132,6 +134,24 @@ proc masonSearchHelp() {
...
+  writeln();
+  writeln("Options:");
+  writeln("    -h, --help                  Display this message");
+  writeln("        --external              Add/Remove dependency from external dependencies");
+  writeln("        --system                Add/Remove dependency from system dependencies");  
+  writeln();
+  writeln("Not listing an option will add/remove the dependency from the Mason [dependencies] table");
+  writeln("Versions are not necessary for removing");
```
ben-albrecht: Design: Can `mason add <package>` without a version number just default to the latest version?

ben-albrecht: There's an open design question about whether or not `mason add/rm` should validate packages or not. After some discussion, we decided to not validate packages, therefore `mason add/rm` will not know what the latest version number.

I'd suggest the following edit to the help output:

> Versions are not necessary for removing

```
Versions are necessary for adding dependencies, but not for removing dependencies
Package names and versions are not validated upon adding
```

tools/mason/MasonHelp.chpl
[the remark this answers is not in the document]
ben-albrecht: Let's save it for later discussion

tools/mason/MasonModify.chpl
```diff
@@ -0,0 +1,267 @@
...
+private proc generateToml(toml: Toml, tomlPath: string) {
+  const tomlFile = open(tomlPath, iomode.cw);
+  const tomlWriter = tomlFile.writer();
+  tomlWriter.writeln(toml);
+  tomlWriter.close();
+  tomlFile.close();
+}
+
+private proc checkVersion(version: string) throws {
+  const pattern = compile("([0-9].[0-9].[0-9][a-zA-Z]?)");
```
ben-albrecht: What is the purpose of the `[a-zA-Z]`?

Spartee: Some packages include letters in their version numbers. For instance, the current version of `openSSL` that is installed on my mac is `0.9.8zh`.

ben-albrecht: Gotcha, so this is a general case regex (handling normal, system, and external packages).

PR #10384, opened 2018-07-18
Fix the test minmaxlocscan-shape.chpl

vasslitvinov:
* add missing `zip` keyword
* add .bad


bradcray:
Yes, I think this test may have predated the `zip()` keyword.

PR #10390, opened 2018-07-18
Set Qthreads build options in a more principled way.

### LINKED ISSUE #9797: When building Qthreads, get compilation options from compileline
As a developer, I'd like to get the compiler options for building Qthreads from the `compileline` utility instead of having them hard-coded in the `Makefile`.

We currently hard-code the `-I` options needed when compiling the Qthreads Chapel internal memory interface file `src/alloc/chapel.c`.  It would be better to get those from the `compileline` utility, especially since that can also supply the `-D` options needed to reference the layer-specific .h files for threading and tasking.  We tried this with commit [7ac628ea](https://github.com/chapel-lang/chapel/pull/9736/commits/7ac628ea744cc8688de086af8bb35a9aa466c340) in [PR 9736](https://github.com/chapel-lang/chapel/pull/9736) but had to revert that, because it ended up not working whenever comm=gasnet since `compileline` needed GASNet to be built before it could run properly and the third-party `Makefile` doesn't force GASNet to be built before Qthreads.

The task here is to rearrange the third-party build order so this can work, and reinstate the change to `compileline` in `third-party/qthread/Makefile`.

gbtitus:
We used to set -I options ad-hoc when using the Chapel allocator and/or
the Chapel runtime copy of the hwloc topology.  Here, get them from the
compileline --includes-and-defines output instead.  As part of this, add
a third-party/Makefile dependence to ensure that we build GASNet before
Qthreads when both are in use, because compileline can't produce the
includes and defines information until after GASNet is installed.

This improves and re-applies a change originally made as part of PR
#9736, which was reverted in #9761 because it broke comm=gasnet
builds.

This resolves #9797.

### DIFF
--- a/third-party/qthread/Makefile
+++ b/third-party/qthread/Makefile
@@ -49,7 +49,7 @@ ifeq ($(CFLAGS_NEEDS_RT_INCLUDES), y)
   # When building the Chapel allocator file we need all the Chapel runtime
   # defs and include directories, plus the third-party include dirs other
   # than that for Qthreads itself (since we're building it).  Notably, we
-  # do not want -DOPTIMIZE; if we have that set when building Qthreads it
+  # do not want -DNDEBUG; if we have that set when building Qthreads it
   # seems to cause Chapel programs built with the result to hang during
   # exit when shutting down Qthreads.
   CFLAGS += \
--- a/third-party/qthread/Makefile
+++ b/third-party/qthread/Makefile
@@ -46,14 +46,19 @@ ifeq (, $(call isTrue, $(CHPL_QTHREAD_NO_CHPL_ALLOC)))
 endif
 
 ifeq ($(CFLAGS_NEEDS_RT_INCLUDES), y)
-  CFLAGS += -I$(CHPL_MAKE_HOME)/runtime/include/localeModels/$(CHPL_MAKE_LOCALE_MODEL)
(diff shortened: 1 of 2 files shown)


third-party/qthread/Makefile
```diff
@@ -46,14 +46,19 @@ ifeq (, $(call isTrue, $(CHPL_QTHREAD_NO_CHPL_ALLOC)))
...
-  CFLAGS += -I$(CHPL_MAKE_HOME)/runtime/include
-  ifeq (, $(call isTrue, $(CHPL_QTHREAD_NO_CHPL_ALLOC)))
-    ifeq ($(CHPL_MAKE_MEM),jemalloc)
-      CFLAGS += -I$(JEMALLOC_INCLUDE_DIR)
-    endif
-  endif
+  # When building the Chapel allocator file we need all the Chapel runtime
+  # defs and include directories, plus the third-party include dirs other
+  # than that for Qthreads itself (since we're building it).  Notably, we
+  # do not want -DOPTIMIZE; if we have that set when building Qthreads it
```
ronawho: Think it's DNDEBUG, not DOPTIMIZE

ronawho: https://github.com/chapel-lang/chapel/pull/9736/commits/4bffdb5bc0e814d52bde3d93a72acbda18c706a7 :)

gbtitus: Crud, I meant to fix that at the time!

PR #10392, opened 2018-07-18
Skip this new BigInteger module test directory if we don't have CHPL_GMP

bradcray:
Rookie mistake, which led to the noise on PGI last night.  Apologies for missing it.


ronawho:
Note that for the GMP dir we use GMP.skipif (recursive skipif) which makes it easier to avoid forgetting about gmp skipifs in the future. So for here it would be a BigInteger.skipif one dir higher if you want to go with that. Or alternatively it looks like our existing BigInteger tests are just in the gmp directory if you wanted to move it under there

bradcray:
Oh right, I forget about that feature, thanks.  My preference would be to move other BigInteger tests here to make @ben-albrecht happy rather than vice-versa.

PR #10393, opened 2018-07-18
Implement forwarding at call site instead of with wrapper fns

### LINKED ISSUE #7850: Forwarding array of strings
### Summary of Problem

Forwarding doesn't appear to work on arrays of strings.

### Steps to Reproduce

**Source Code:**

```chapel
// Error case
{
  record Arrays {
    var D: domain(1);
    var A: [D] string;
    forwarding A;
  }

  var keysArrays: [1..4] Arrays;
}
```

```bash
# Compiling
> chpl repro.chpl
src/repro.chpl:3: error: use of '_return_tmp_' before encountering its definition, type unknown
```

For reference, the following code works as expected:

```chapel
// Array element type of 'int' instead of 'string'
{
  record Arrays {
    var D: domain(1);
    var A: [D] int;
    forwarding A;
  }

  var keysArrays: [1..4] Arrays;
}
```

### Configuration Information

- Output of `chpl --version`: `chpl Version 1.17.0 pre-release (b23a4d8)`

### LINKED ISSUE #8449: Record forwarding to a generic class with a sync variable field
When forwarding a class with a `sync` variable field from a record, it will cause the compiler to throw this user error...

```
$CHPL_HOME/modules/internal/ChapelSyncvar.chpl:898: In function 'supportsNativeSyncVar':
$CHPL_HOME/modules/internal/ChapelSyncvar.chpl:899: error: 'param' functions cannot return non-'param' values
$CHPL_HOME/modules/internal/ChapelSyncvar.chpl:88: error: unable to resolve return type
```

Note that the class *must* have a generic type field as well.

```chpl
record ForwardingWrapper {
	type eltType;

	inline proc _value {
	  return new C(eltType);
	}

	forwarding _value;
}


class C {
	type eltType;
	var lock$ : sync bool;
}

var wrapper : ForwardingWrapper(int);
```

[TIO](https://tio.run/##ZY7BCoMwEETPyVfMoQf9BXMU@gWFHksat23okoQ1KiL99lSL9OJpYXbezLiXTcSlCLkoHc5RJiudD8@r2JRIsGiV50Qgzpf1Gq2VD@wDIUl0uI2WB9pcgFAeJCDQhLba/bXR6rMyj3/wjhi9ytqx7Xu0xxY1WgFH9z6hQT8Hh3uM/IO2z7Sva46LKx9ybUr5Ag)

**Configuration Information**

```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none *
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: cstdlib *
CHPL_ATOMICS: intrinsics
CHPL_GMP: none *
CHPL_HWLOC: hwloc
CHPL_REGEXP: none *
CHPL_AUX_FILESYS: none
```

### LINKED ISSUE #8485: forwarding bug with return intent overloads
The forwarding mechanism seems to have a problem with forwarding return intent overloads.

``` chapel
class MyClass {
  var _x:int;
  proc ref accessor() ref { return _x; }
  proc const ref accessor() const ref { return _x; }
}

record R {
  forwarding var ptr:MyClass;
}

proc getx(arg:R) {
  return arg.accessor(); // issue also occurs with arg._x
}

proc test() {
  var r = new R(new MyClass(1));
  var x = getx(r);
  writeln(x);
}

test();
```

The forwarding mechanism probably needs to create forwarding methods for all return-intent overloads in such cases.

### LINKED ISSUE #9601: owned coercion does not work with forwarding
``` chapel
class Lower {
  proc something() {
    writeln("in foo");
  }
}

class Upper {
  forwarding var low : owned Lower = new owned Lower();
}

var u = new owned Upper();

u.something();
```

I would expect this to compile and run, but it doesn't. See also `test/classes/bharshbarg/generic-forwarding.chpl` when modified to use `owned`.

### LINKED ISSUE #9926: Forwarding bug with param fields
### Summary of Problem
The following program fails to compile when I attempt to access a forwarded param field:

```chpl

class Inner {
  param x : int;
  param y : bool;

  var z : if y then x*int else x*real;
}

class C {
  forwarding var i = new Inner(2, true);
}

var c = new C();
writeln(c.x);
```
The output of the compiler is:
```
foo.chpl:14: error: ambiguous call 'C.x'
foo.chpl:10: note: candidates are: C.x
foo.chpl:10: note:                 C.x
```

[classes/forwarding/forward-param.chpl](https://github.com/chapel-lang/chapel/blob/master/test/classes/forwarding/forward-param.chpl)

### LINKED ISSUE #10063: Compiler segfaults when incorrectly forwarding to self
### Summary of Problem
Compiler segfaults when a class forwards to itself. Note that the compilation succeeds if the code in `main` is removed but the self-forwarded class remains (probably because it doesn't get instantiated or simply optimized out).

### Steps to Reproduce

**Source Code:**

```chapel
module ForwardtoSelf {
class A {
  forwarding var data: A;
}
proc main() {
  var a = new A();
  delete a;
}
} // end module ForwardtoSelf
```

### Configuration Information

### LINKED ISSUE #10355: Compiler throws 'ambiguous call' when trying to access generic type from a forwarding record or class
Attempting to obtain the generic type of a forwarded class or record from a forwarding class or record seems to confuse the compiler. 

**MWE**

```chpl
class C {
	type t;
}

record R {
	var instance = new C(int);
	forwarding instance;
}

var r : R;
writeln(r.t : string);
```
[TIO](https://tio.run/##PcsxDoMwDEbhGZ/iH2HpARoxcQNuYAW3RIoMcqxGVdWzB@jQ9el7ceVdcmsxcymY8KHO37vAA32JTOJmC@Yrv9iQtDhrFIxQqZj6pD4E6h6bVbYl6fNPfv/1GO6YA1VLLll7u/kZituJh9DaAQ)

**Output:**

```
.code.tio.chpl:11: error: ambiguous call 'R.t'
.code.tio.chpl:7: note: candidates are: R.t
.code.tio.chpl:7: note:                 R.t
```

In the above, I would expect `r.t` to first search `R` to determine if it has a type `t`, and if not then it will attempt to search the forwarded instance to see if it can find a type `t`.

### LINKED ISSUE #10358: Nonsensical 'type mismatch' in array assignment

mppf:
This PR rewrites the way that forwarding is handled in function
resolution. Instead of creating wrapper-methods for forwarding (with
populateForwardingMethods), it now adjusts the call site to add the
forwarding call (with resolveForwardedCall). This approach allows
coercion as well as return intent overloads to be handled easily and
integrates better with the rest of function resolution.

Resolves #7850
Resolves #10358
Resolves #8449
Resolves #9926
Resolves #10355
Resolves #8485
Resolves #9601 
Resolves #10063


test/classes/forwarding/point-of-insn-issue.chpl
```diff
@@ -0,0 +1,31 @@
+record ForwardingRecord {
+  var instance;
+
+  proc init(type msgType) {
+    instance = new ForwardedClass(msgType);
```
vasslitvinov: Should this be "new owned"? Please check for memory leaks and valgrind.

test/classes/forwarding/forwarding-sync-issue.chpl
```diff
@@ -0,0 +1,17 @@
...
+  forwarding _value;
+}
+
+
+class C {
+  type eltType;
+  var lock$ : sync bool;
+}
+
+var wrapper : ForwardingWrapper(int);
```
vasslitvinov: I do not see what this test is testing. Add a comment about that?

Also, how about some action with wrapper? Like writeln(wrapper.lock$.isFull) ?

mppf:
- [ ] [Error matching compiler output for classes/bharshbarg/generic-forwarding]
- [ ] [Error matching compiler output for classes/forwarding/forwarding-cycle]

bradcray:
This is like the grand slam of PRs!!  Nice to come back from vacation to...

PR #10394, opened 2018-07-18
Remove/update futures related to constructors

benharsh:
This PR removes various futures if they were testing a constructor-specific bug/feature. Other futures had their descriptions updated to mention initializers instead of constructors.

Some other tests, only tangentially related to constructors, were removed in this PR for convenience:
- studies/shootout/pidigits/bradc/pidigits-pretty2.chpl (no longer valuable with advent of bigint)
- users/ferguson/byvalue.chpl (old copy semantics test left behind over the years)
- users/shetag/objRefAsFormArg.chpl (constructor 'feature' not worth testing)
- classes/initializers/phase1/nilDefaultVal.chpl (constructor 'feature' not worth testing)

Testing of affected tests:
- [x] local


test/classes/diten/constructUnion.future
```diff
@@ -1,3 +1,3 @@
-bug: setting a union field in constructor call leaves the field unset
+bug: setting a union field in initializer call leaves the field unset
```
lydia-duncan: I might be tempted to rename the test but leaving it is also reasonable

test/classes/hannah/coercingIntToSyncIntOnConstructor.future
```diff
@@ -1 +1 @@
-bug: default constructor expects a sync int, but when passed an int upon construction, does not perform a coercion from int to sync inc.  
+bug: default initializer expects a sync int, but when passed an int upon initialization, does not perform a coercion from int to sync inc.
```
lydia-duncan: This should be removed, I added a copy to test/classes/initializers/

Issue #10395, opened 2018-07-18
Error-Handling conditional throw expects return value 


Spartee:
## Summary of Problem
When throwing an error within a conditional within a function that returns a value, control errors out because it thinks the throwing branch of the conditional should return a value.

## Steps to Reproduce

**Source Code:**

```chapel
class SampleError : Error {
  var msg:string;
  proc init(msg:string) {
    this.msg = msg;
  }
  proc message() {
    return msg;
  }
}

proc thrower(cond: bool) : bool throws {
  if cond {
    return cond;
  }
  else {
    throw new SampleError("this wont compile");
  }
}

proc caller(cond: bool) {
  try! {
    writeln(thrower(cond));
  }
  catch e: SampleError {
    writeln(e.message());
  }
}

caller(false);
```

## Error
error-handling-repro.chpl:12: In function 'thrower':
error-handling-repro.chpl:12: error: control reaches end of function that returns a value

#### Compile command:
chpl error-handling-repro.chpl

#### Configuration Information
machine info: Darwin C02PG1EFG3QP 16.7.0 Darwin Kernel Version 16.7.0: Fri Apr 27 17:59:46 PDT 2018; root:xnu-3789.73.13~1/RELEASE_X86_64 x86_64
CHPL_HOME: /Users/spartee/chapel *
script location: /Users/spartee/chapel/util
CHPL_HOST_PLATFORM: darwin *
CHPL_HOST_COMPILER: clang
CHPL_TARGET_PLATFORM: darwin
CHPL_TARGET_COMPILER: clang
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2 *
CHPL_LLVM: llvm
CHPL_AUX_FILESYS: none

bradcray:
I'm running into this today as well.  @mppf, do you agree that a throw should be as good as a return when it comes to this error check?

PR #10406, opened 2018-07-18
Convert TOML to error handling

Spartee:
This PR converts the halts in `TOML.chpl` to error handling. The motivation behind this PR was a number of simple mistakes that a user might make in a `Mason.toml` when configuring a Mason project. For this reason the tests were conducted using the `mason update` feature, and the test  were placed in `/test/mason/`. Expect more tests/futures to be added as more users run into uncaught errors that represent the areas of the toml spec we do not support. 

In implementing this PR I ran into #10395 which caused me to unnecessarily modify a few functions to get the error handling to work as expected. This will be changed once #10395 is addressed.


modules/packages/TOML.chpl
```diff
@@ -843,7 +894,8 @@ used to recursively hold tables and respective values
         when fieldDate do return 'datetime';
         when fieldToml do return 'toml';
         otherwise {
-          halt("Unknown type: '", this.tag, "'");
+          throw new TomlError("Unknown type");
+          return "nil";
```
ben-albrecht: I believe this is due to an issue you were opening. It might be worth adding that issue number to the comment here (if it's open already).

Spartee: See PR description

test/mason/mason-update-toml-error/tomls/parseloopUnexpectedChar.good
```diff
@@ -0,0 +1 @@
+Unexpected Token on line 1 -> =
```
ben-albrecht: Looking good. I think it'd be nice to make this error message style consistent with Chapel compiler and runtime errors:  `filename:linenumber: error msg`


e.g.

```
parseloopUnexpectedChar.toml:1: Unexpected Token
```

ben-albrecht: For historians: Turns out we only have access to the file channel within the `TOML` module, the TOML module opts for printing just a line number instead.

bradcray:
@ben-albrecht , @Spartee : The [documentation on TomlError](https://chapel-lang.org/docs/master/modules/packages/TOML.html#TOML.TomlError) looks a little naked? (no English descriptions, just code prototypes).  Was that intended?

PR #10408, opened 2018-07-19
Check LCALS answers using relative differences rather than absolute

dmk42:
LCALS answers were being checked using absolute error instead of relative error.
```
absolute_error = abs(expected - actual)
relative_error = abs(absolute_error / expected)
```
Specifically, an absolute error of 0.17 was the criterion for passing.  This allowed 17% error in some answers, and not even a single-bit error in others.

I switched to checking relative error instead.  Both linux64 and darwin were passing before, so I chose the minimum relative error that would still allow them to pass.

Having done that, I noticed that now, all the platforms in the skipif pass too.  I have deleted them all from the skipif.

ARM passes as well, which was the original reason for investigating this.


dmk42:
@daviditen - would you mind taking a post-commit look at this?

I've tested this on all the most important configurations, but would like to use nightly testing to scour all the dark corners.

daviditen:
Cool, looks good, thanks!

bradcray:
Just curious:  What does the reference version of LCALS do for validation?  Were we doing something different (or: does the reference version also fail in these configurations that were giving us problems?)

daviditen:
The reference version doesn't validate the values.  I added the checks to our version because it was too much of a pain to know if things were going right without them.

PR #10424, opened 2018-07-19
Stop throwing -D_POSIX_C_SOURCE for darwin

ronawho:
This was originally added in 51eab78edac, but there was no clear explanation
given (and this was back when our mac support was hit-or-miss and we didn't
have any nightly testing.) It doesn't seem to hurt anything by not defining it
now, and defining it prevented the Bessel functions added in #10391 from being
accessible.

### DIFF
--- a/make/compiler/Makefile.clang
+++ b/make/compiler/Makefile.clang
@@ -99,8 +99,7 @@ ifeq ($(CHPL_MAKE_PLATFORM), darwin)
 ARCH := $(shell test -x /usr/bin/machine -a `/usr/bin/machine` == ppc970 && echo -arch ppc64)
 RUNTIME_CFLAGS += $(ARCH)
 RUNTIME_CXXFLAGS += $(ARCH)
-# the -D_POSIX_C_SOURCE flag prevents nonstandard functions from polluting the global name space
-GEN_CFLAGS += -D_POSIX_C_SOURCE $(ARCH) -fno-strict-overflow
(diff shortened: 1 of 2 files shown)


ronawho:
@dmk42 I'm going to merge this to quiet testing, but would appreciate a sanity check post-commit

dmk42:
There also seem to be some definitions of _POSIX_C_SOURCE in runtime/include/sys_basic.h, runtime/src/qio/sys.c, runtime/src/qio/sys_xsi_strerror_r.c, and runtime/src/comm/ugni/comm-ugni-launch.c.

It may be that most of those are benign, but the one in runtime/src/qio/sys.c appears to have a typo in it.  It has one less digit than the others.

ronawho:
Those seem to be defining it to a specific value instead of just defining it. It's not obvious to me -- do you want me to change something or look into removing other instances of this, or something?

dmk42:
I mostly just wanted to bring it to your attention in case it mattered, but while you're fixing up the _POSIX_C_SOURCE situation, would you mind checking if you can safely set the one in runtime/src/qio/sys.c to 200112L (instead of 20112L)?  It would be nice to fix that typo while we're at it.

ronawho:
I'm really not that familiar with any of this. I would prefer to merge this to quiet testing and let you make any changes to how we set what posix/whatever standard we want. 

I'm happy to help with testing there, I just couldn't write an intelligent justification for why I'm making the change -- for this PR my argument is basically "it seems to fix testing and dmk ok'd it :)"

PR #10425, opened 2018-07-19
Add modules as a mason dependency

ben-albrecht:
Mason depends on `SysCTypes` being generated, so it should depend on the `modules` target.


ben-albrecht:
@awallace-cray reported `make mason` failing due to `SysCTypes` not being built on a fresh checkout.

PR #10427, opened 2018-07-19
Common DSI Methods Refactor

### LINKED ISSUE #9942: Standard DSI methods should be implemented in base domain
There are a number of DSI methods that do not change in behavior of in across different domain implementation. However, they currently need to be 'reimplemented' in each implementation, which is often trivially done by calling the method on some parent or internal view domain. A better solution would be to have these functions defined at a top level class and then inherited by different domains extending their behavior.
This would prevent them from being accidentally left out (as illustrated by #9938).

Some suggested methods include:
+ size
+ numIndices
+ low
+ high
+ stride
+ alignment
+ first
+ last
+ alignedLow
+ alignedHigh

ian-bertolacci:
Resolves #9942.

These methods have been removed from DefaultRectangularDomain and moved to BaseRectangularDomain:
+ dsiNumIndices
+ dsiLow
+ dsiHigh
+ dsiStride
+ dsiAlignment
+ dsiFirst
+ dsiLast
+ dsiAlignedLow
+ dsiAlignedHigh

They retain the same functionality by invoking `dsiDim( d )` in place of `this.ranges( d )`.

Any class inheriting from BaseRectangularDomain has had these methods removed if they are redundant.

`BaseRectangularDom.dsiDim( d ) : range( idxType, Bounded, stridable )` has been added as a halting-abstract method added to BaseRectangularDomain.

All definitions of `dsiDim( param d )` has also been removed. This required modifying one test (distributions/dm/n.chpl)

No changes were made to any sparse domains, as these kinds of methods were already pushed up to BaseSparseDomImpl.

### DIFF
--- a/modules/dists/BlockCycDist.chpl
+++ b/modules/dists/BlockCycDist.chpl
@@ -1,15 +1,15 @@
 /*
  * Copyright 2004-2018 Cray Inc.
  * Other additional copyright holders may be indicated within.
- * 
+ *
  * The entirety of this work is licensed under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License.
- * 
+ *
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -19,7 +19,7 @@
 
 //
 // BlockCyclic Distribution
-// 
+//
 //      BlockCyclic    BlockCyclicDom     BlockCyclicArr
 //
 //   LocBlockCyclic    LocBlockCyclicDom  LocBlockCyclicArr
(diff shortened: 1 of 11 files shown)


ronawho:
There are some performance regressions from this. Graphs are [here](https://chapel-lang.org/perf/chapcs/dsi_push_up_attempt/?startdate=2018/08/01&enddate=2018/08/30&graphs=arrayvectoroperations,arrayviewcreation,minimdljsize10time,nbodyvariations)

The largest regressions are for the "Array Vector Operations". Here's a simple reproducer for push_back, which is ~2x worse:

```chpl
use Time;

config const n = 50000000;

var A: [1..0] int;

var t: Timer; t.start();
for i in 1..n {
  A.push_back(i);
}
writeln(t.elapsed());
```

That runs in .45s for me on master and .89s with this PR.

I haven't gotten a chance to dig into the source of the regressions yet though.


There are also some correctness failures for some procedures that are missing the override keyword. Tests that are failing: 

```
release/examples/benchmarks/miniMD.miniMD
studies/lulesh/bradc.lulesh-dense
studies/shootout/spectral-norm/sidelnik.spectralnorm_2
```

For these I think you just need to add `override` to any procs that are overriding the methods that have now been pushed into the base class.

mppf:
@ronawho - what should we do with this PR?

I think it's likely that the performance regressions have to do with virtual dispatch in cases where it wasn't occuring before, but that's absolutely just a guess.

ronawho:
I'm not sure what the next steps are here. I guess we need somebody to take this over and fix the correctness and performance issues.

ronawho:
Closing based on performance regressions and relatively low priority of this cleanup. https://github.com/chapel-lang/chapel/issues/9942 will continue to stay open as a reminder that we want this cleanup eventually.

PR #10428, opened 2018-07-19
Fix instantiation point for serialization functions

benharsh:
This was yet another case in which we resolved a function in a temp block-statement, removed that block, and later tried to use the old instantiation point.

Testing:
- [x] local + futures
- [x] gasnet + futures

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -8085,7 +8085,13 @@ static FnSymbol* resolveNormalSerializer(CallExpr* call) {
 
   block->remove();
 
-  return call->resolvedFunction();
+  FnSymbol* ret = call->resolvedFunction();
+
+  if (ret != NULL) {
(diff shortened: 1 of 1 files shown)


mppf:
BTW the other alternative is to make the temp block-statement a scopeless block; then the point of instantiation is set correctly.

mppf:
In fact I think that might be preferable. What do you think?

mppf:
Let's talk about this on PR #10432.

bradcray:
I'm probably missing some context, but I was surprised that this PR didn't resolve or add a test to lock in the new behavior.

mppf:
It resolved failures in multi-locale testing with the following tests:

    optimizations/remoteValueForwarding/serialization/destroyGlobals
    optimizations/remoteValueForwarding/serialization/globals
    optimizations/remoteValueForwarding/serialization/missingDeserializer
    optimizations/remoteValueForwarding/serialization/staticSize

Issue #10119, opened 2018-07-02
Working version of the new "Chapel Cray module packaging script" (new scripts to build customizable Cray module from source)


awallace-cray:
Write a working version of the new "Chapel Cray module packaging script" 

This is the 3rd item in https://github.com/chapel-lang/chapel/issues/10046#issuecomment-401422989, new scripts to build customizable Cray module from source

The deliverable is a working script suitable for stand-alone use by an outside end-user, as described in https://github.com/chapel-lang/chapel/issues/9551. 

The deliverable is also intended for use in Chapel's  CI automation, ultimately replacing the current Chapel Cray module build scripts used in the Cray-internal Jenkins. However, this task does not include final implementation and testing for that application.

awallace-cray:
2018-07-23 10:00 Status, for sprint planning. 
Only as far as reviewing notes, so not much progress. The task is well-understood, at least by me, so thats good. There are a couple non-trivial bits however- generating or supplying files that presently come from some source outside of Chapel.git.

awallace-cray:
2018-08-10 06:30 Status
Ref: https://github.com/awallace-cray/priv/issues/16 
(text search for `Chapel Cray module packaging script`)

Done:
* generate chapel.spec (RPM) w versions
* generate set_default script w versions
* generate release_info file

Well-started, if not fully done
* design: instead of well-structured "setenv" files implementing project-specific details, here we will have project-specific subdirectories little convention or structure for organizing the content within. At least not yet, until we see how two or more different projects shape up.  

To do:
* generate modulefile w versions
* actual RPM build
* fill-in the chapel_rpmbuild.bash "wrapper" script. Right now it only has stubs for important things like accepting commandline parameters
* similar detail completions in the generator scripts
* anything like a README

awallace-cray:
2018-08-14 17:58
As above, except these are working drafts:
* generate modulefile w versions
* actual RPM build

awallace-cray:
2018-08-17 16:37
Working draft scripts are almost complete! 
Only shell scripting left is to replace a few hardcoded lines with appropriate commandline options. 
Only testing still needed is to test with a non-null "version-tag" option (non-null to differentiate a dev- or site-built RPM/installation from an official Chapel release).

awallace-cray:
Earlier today 2018-08-17 we heard from the site, they have started testing setenv-example-3 (without this packaging yet). OK so far.

awallace-cray:
2018-08-22 19:48
[...]

Issue #10109, opened 2018-06-29
Working version of the new "Chapel Cray module build script" (new scripts to build customizable Cray module from source)


awallace-cray:
Write a working version of the new "Chapel Cray module build script"

This is the 2nd item in https://github.com/chapel-lang/chapel/issues/10046#issuecomment-401422989, new scripts to build customizable Cray module from source. See also https://github.com/chapel-lang/chapel/issues/10046#issuecomment-401418133

The deliverable is a working script suitable for stand-alone use by an outside end-user, as described in https://github.com/chapel-lang/chapel/issues/9551. 

The deliverable is also intended for use in Chapel's  CI automation, ultimately replacing the current Chapel Cray module build scripts used in the Cray-internal Jenkins. However, this task does not include final implementation and testing for that application.

awallace-cray:
2018-07-23 10:00 Status for planning. My only comment is this https://github.com/chapel-lang/chapel/pull/10429#issuecomment-406727398

awallace-cray:
2018-07-23 10:00 Status for planning. My only comment is this https://github.com/chapel-lang/chapel/pull/10429#issuecomment-406723321 (re setenv-gh9943.bash).

Issue #9943, opened 2018-06-20
Default COMM causes undefined-reference errors when Chapel is built with non-default COMM


BryantLam:
### Summary of Problem
[Y'all broke my workflow.](https://xkcd.com/1172/)

I do the following when I build Chapel from source:
```
export CHPL_COMM=gasnet
export CHPL_COMM_SUBSTRATE=smp
make -j4
make -j4 mason
```
I am building Chapel for gasnet/smp multi-locale emulation on my single-locale system. This means that my environment is in "single-locale" mode most of the time via the default setting on environment variables `CHPL_COMM` and `CHPL_COMM_SUBSTRATE`. When I need to use multi-locale emulation, I set those environment variables and then immediately recompile the application code.

This worked until 68dbb1dd1693b8ace384b5a117042f6fb5859893 @gbtitus.

Now, when I compile any program with the default `CHPL_COMM` and `CHPL_COMM_SUBSTRATE`, I get the following error:
```
In function `helpSetupLocaleFlat':
_main.c:(.text+0x581a8): undefined reference to `chpl_topo_getNumCPUsPhysical'
_main.c:(.text+0x581f7): undefined reference to `chpl_topo_getNumCPUsPhysical'
_main.c:(.text+0x58246): undefined reference to `chpl_topo_getNumCPUsLogical'
_main.c:(.text+0x58295): undefined reference to `chpl_topo_getNumCPUsLogical'
collect2: error: ld returned 1 exit status
gmake[1]: Error 1
error: compiling generated source
```

If I set `CHPL_COMM=gasnet` and `CHPL_COMM_SUBSTRATE=smp` in my environment, the code compiles and works as expected. (I switch back to single-locale mode by unsetting those environment variables.)

I'm not going to claim that my workflow is the best, so feel free to suggest a more proper approach.

### Configuration Information
`chpl version 1.18.0 pre-release (abbbb6476c)`

ronawho:
It sounds like you haven't rebuilt the runtime for the `CHPL_COMM=none` configuration

Does something like the following resolve the error?: 
 
```sh
# build the runtime/third-party for CHPL_COMM=none
unset CHPL_COMM 
unset CHPL_COMM_SUBSTRATE
make -j4 
make -j4 mason

# build the runtime/third-party for CHPL_COMM=gasnet
export CHPL_COMM=gasnet
export CHPL_COMM_SUBSTRATE=smp
make -j4
make -j4 mason
```

[...]

PR #10429, opened 2018-07-19
New scripts to build custom Cray module from source, part 2.

awallace-cray:
From part 1 (PR #10360, now replaced by this):
---
Task #10120 build_configs.py changes 
(new scripts to build customizable Cray module from source)

* Move build_configs.py to new folder "build_configs". This
  will help users manage their related working files, like setenv
  scripts and log files.

* Many changes in the way Chapel build config parameters
  are handled in build_configs.py:
  * All known Chapel build config parameters are recognized
    (ie, all "CHPL_XXX" config variables generated by
    "printchplenv --all --no-tidy" as of this date).
    That's about twice as many "build dimensions" as before.
  * build_configs.py no longer tries to:
    - validate build config parameters provided by the user
    - provide meaningful default values for build config
      parameters
    - provide meaningful "help" text describing build config
      parameters.
  * Any or all Chapel build config parameters may have zero
    non-null values assigned. When that happens, the
    corresponding "build dimension" is effectively undefined.
    Ie, that Chapel build config parameter value is left for
    Chapel make to decide.
  * All Chapel build config parameters initially default to a 
    value of "None" (ie, null; empty).
  * On startup, build_configs.py scans the existing CHPL_XXX
    environment variables (if any):
    - Any existing CHPL_XXX env value becomes the default
      value of the corresponding Chapel build parameter, 
      before any commandline options are scanned.
    - If "UNSET" is given as a build parameter value
      (on the the commandline), then the corresponding CHPL_XXX
      env variable will be silently unset (as in bash "unset"),
      for the corresponding Chapel make run.

* Extra make parameters (options or targets) may now be
  added to the end of the build_config.py commandline
  (after all the "-x/--xxxx>" options). If found, these will
  simply be added to the generated Chapel make commands.

* New commandline parameter "-s/--setenv" may specify a bash-
  compatible script file containing arbitrary commands.
  If found, this setenv script will be source'd before each
  Chapel make is run.
  This is typically used to modify the O.S. environment seen
  by the Chapel make. Setenv scripts typically use the CHPL_XXX
  env variables defined through build_config.py, to decide
  what to do.
  Additional env variables with names like "BUILD_CONFIG_xxx"
  are also defined for each Chapel make run. These env vars
  are ignored by Chapel, but may be used by a setenv script;
  for example, to tell whether the setenv script was
  run from within build_configs.py, or not.

* New commandline parameter "-l/--make-log" may redirect the
  stdout/stderr from Chapel make commands to a log file. If no
  make-log file is defined, the Chapel make output is discarded.

* The final exit code of build_configs.py is now just the max
  exit code value from all the Chapel build configs (makes)

* A build_config.py "--dry-run" now runs printchplenv, and the
  results are checked for warnings or errors.
  Setenv scripts, if any, will be honored in dry-runs; however,
  any "--make-log" directives will be ignored.

Part 2 (This PR)
---
Task  #10109 New shell script "build_chapel.bash", with a "functions.bash"
for general use by any build_configs-related bash scripts.

build_chapel.bash integrates with build_python.py and the 
existing setenv-example-1 script.

As of this writing, this combination builds a made-up Chapel project on
my MacBook.  Next, I will take that existing setenv-example-1.bash file
and go two directions:
(a) simplify it, and address https://github.com/chapel-lang/chapel/issues/9943. Probably as part of this PR, depending on when things happen. 
(b) add more complexity to another copy of it, to at least start to address a realistic Chapel Cray module. That work will be continued in a new PR for task #10119.


awallace-cray:
setenv-gh9943.bash demonstrates how someone could address the use case described in https://github.com/chapel-lang/chapel/issues/9943, using the modified build_configs.py with an integrated setenv script file.  It yields about the same result as the solution outlined in https://github.com/chapel-lang/chapel/issues/9943#issuecomment-398924164, except now build_configs.py with setenv can handle the entire problem.

awallace-cray:
With today's hands-on experience, I think this build_configs.py should get one more change, but I am going to defer it because it is not very important. 
When you run build_configs with a setenv callback script, the output shows you which Chapel build config parameter values were passed to every child process (source setenv, followed by make, for each Chapel build config it runs.  
However, the setenv script can change things- it can even skip the makes- and these changes are not visible in the standard output.  You can echo anything you want in your setenv callbacks, but that output only shows up in the make_log file (if any)- not the standard output you see from the command line. 
When you do build_configs.py with --dry_run, the make call gets replaced by a printchplenv call, so you could do that - if you thought of it. 
It might be better to just run a subprocess with build_configs (or setenv + build_configs) every time, to get the config info from Chapels make system; then (if not a dry_run), just run a second subprocess to do the make (or setenv + make). That is, just let the setenv callback be run twice for each build config.

awallace-cray:
These mixed case env var names are admittedly lame. 
- https://github.com/awallace-cray/chapel/blob/3fa52583beb9c96ea85c70a0eb5513223c5176d1/util/build_configs/build_configs.py#L428

Should be uppercase names or lowercase names. Or maybe all packed into one env var like a Python dictionary.

util/build_configs/setenv-gh9943.bash
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
```
benharsh: Is there a more descriptive name we could give this file? I don't think 'gh9943' will be useful to most users.

awallace-cray: Arg- I meant to have a README in this directory, but forgot. 
I shall do that later today. While at it, maybe some better file names for example setenv scripts will occur to me. Any suggestions? Thanks again.

PR #10430, opened 2018-07-19
Switch perf playground to measure Ian's DSI refactor

ronawho:


ronawho:
@ian-bertolacci perf results should be up tomorrow at http://chapel-lang.org/perf/chapcs/dsi_push_up_attempt

PR #10432, opened 2018-07-20
Mark removed block scopeless instead of changing instantiation pt

mppf:
I think this is a better way to write the fix in #10428.

- [x] local + futures testing
- [x] gasnet + futures testing


benharsh:
Ah, I wasn't aware of this functionality. Looks good to me!

Issue #8664, opened 2018-03-01
Owned,Shared meta-issue / Owned,Shared bugs


mppf:
This issue tracks Owned/Shared bugs

* #7894
* #8485
* field of Owned type does not work with default constructor / default initializer `classes/ferguson/delete-free/owned-shared-fields2` 
* default = operator for record with Owned field does not use ref RHS, add FLAG_COPY_MUTATES to record

PR #10436, opened 2018-07-20
Owned shared improvements and toposort

mppf:
* shared can be assigned from `nil`
* shared can be assigned from an `owned`
* shared can be initialized from an `owned`
* toposort updating using these features so it can pass --lifetime-checking
* tests added for these features

Note, TopoSortResult ought to store permutationMap as owned,
but for now it uses shared, due to the issue described last in #8664.

- [x] full local testing


ian-bertolacci:
Toposort looks good.

Issue #7857, opened 2017-11-22
"use only" shades class methods


buddha314:
When doing `use BThang only` the class methods are shaded

```chapel
module AThang {
  /*
  athang.chpl
   chpl athang.chpl -M/Users/buddha/sandbox/chapel/usey -o thangs
  */
  use BThang only;

  proc main() {
    writeln("Doin' a thang here...");
    var bthang = new BThang.MahThang();
    writeln(bthang.sayMyName());
  }
}
```

```chapel
/* bthang.chpl */
module BThang {
  class MahThang {
    proc sayMyName() {
      return "Ain't Nuthin but a B Thang, Baaaaaaby...";
    }
  }
}
```

```bash
athang.chpl:7: In function 'main':
athang.chpl:10: error: unresolved call 'MahThang.sayMyName()'
```

The function `sayMyName()` is not recognized as a class function if `only` is used.  Remove `only` and it works fine.

ben-albrecht:
@buddha314 - Thanks for reporting. Looks like a bug to me.

Might be good to get input from @bradcray when he gets back.

lydia-duncan:
This is intentional, though confusing.  `use BThang only;` means none of BThang's symbols are accessible without the module prefix.  You were only able to use MahThang's constructor because you called it via "BThang.MahThang()" - none of MahThang's methods are visible in your current scope.  If you change the line to `use BThang only MahThang;`, the methods will be visible.

buddha314:
If I have the class object, should I not be able to use the methods of that class?  Why let me call the constructor then?

lydia-duncan:
You're able to find the constructor in your current set up because it isn't a method on the type, it's a function that generates the type instance for you and so you can access it through the module prefix.  The type itself isn't really known in your current scope, though.  Try declaring the bthang variable with an explicit type and you'll see what I mean - you can't just say `var bthang: MahThang = new BThang.MahThang();` because the scope doesn't even know about the type MahThang on its own.  Why should it know about the methods or fields on a type whose name it doesn't even know?

You can still pass the instance to some place that can find the methods (like another scope that has used BThang without the limitation provided) so it isn't completely useless to you in that scope, but it is really better to have included the type in your `only` list.

lydia-duncan:
I think it would probably be better for us to prevent functions from returning types that the current scope doesn't understand.  This is kind of related to the questions raised in #6067

buddha314:
[...]

PR #10439, opened 2018-07-20
 Alter semantics of `use ... only;` and `use ... except *;` w.r.t methods/fields

### LINKED ISSUE #9983: "use only" prevents visibility of module deinit
### Steps to Reproduce
`error: unresolved call 'R(int(64)).deinit()` with `use A only`.
If `only` is omitted, the program compiles. Possibly related to #7857.

**Source Code:**

```chapel
module A {
  record R {
    type t;
  }
}

proc main() {
  use A only;
  var r: A.R(int);
}
```

### Configuration Information

lydia-duncan:
Resolves #9983, #7857, and #7249

Basically, replaces the relatedNames vector with a new vector that tracks every
method and field defined in the used module.  This vector is ignored when the
use has an except list and checked when the use has an only list.  This enables
their use if the type on which these fields or methods are defined is somehow
obtained (via qualified access, returning it from an included function, another
module, etc) when the except or only list would otherwise cause the type to be
excluded.

Keep a separate category for special functions, so that we can distinguish between
methods and functions with the same name without eliminating support for
constructors, the _new function, etc.

Adds some new tests covering various combinations of method and field
accesses when a `use M only;` or `use M except *;` statement is present.
Resolves 7 futures, including the bug where excluding one type would impact
other types with methods or fields that shared a name with a method or field
on the excluded type.

Passes all the new tests I wrote, and a full paratest with futures

Note for my reviewer: this is probably easiest to review in Split mode instead
of Unified


lydia-duncan:
@benharsh - would you be up for reviewing this?

bradcray:
This was great to come home from vacation to!

Issue #8199, opened 2018-01-15
When should we set GASNET_MAX_SEGSIZE?


ronawho:
Currently, we (mistakenly) set `GASNET_MAX_SEGSIZE` only when we're _not_ on a cray: 
https://github.com/chapel-lang/chapel/blob/f3613c87d22c75f91e77d72773523868bdf9ef1c/third-party/gasnet/Makefile.setup#L33-L35
https://github.com/chapel-lang/chapel/blob/f3613c87d22c75f91e77d72773523868bdf9ef1c/runtime/src/comm/gasnet/comm-gasnet.c#L764-L777

We used to only set it when we were on a cray, but that condition was accidentally reversed with 1d5e2442d6

It seems weird to base whether we're setting `GASNET_MAX_SEGSIZE` on the system. It seems like we should only set `GASNET_MAX_SEGSIZE` when we're using segment fast/large.

From the gasnet docs:

```
--with-segment-mmap-max=<n>GB - configure maximum segment size.
  Many of the GASNet network conduits obtain memory for the shared segment
  (in FAST and LARGE segment configurations) using mmap().
...
* GASNET_MAX_SEGSIZE - control the upper limit for FAST/LARGE segment size 
```

I think we should adjust our code to set segsize only for segment fast/large. I'd also like to confirm that setting this only impacts performance, and not correctness.

ronawho:
>I think we should adjust our code to set segsize only for segment fast/large. I'd also like to confirm that setting this only impacts performance, and not correctness.

@bonachea and/or @PHHargrove if you get a second, could you weigh in this?

PHHargrove:
The environment variable `GASNET_MAX_SEGSIZE` has no impact at all for segment everything.

For segment fast / large it may limit the maximum size of the GASNet segment.  
I say "may" because other factors may impose a tighter limit.  
However, the available segment size will not exceed the value of this environment variable.  

As to the "only impacts performance, and not correctness" issue:   
So long as you are not choosing an absurdly small value, there is no correctness impact on GASNet.

mppf:
I believe that part of our setting GASNET_MAX_SEGSIZE is so that we can propagate CHPL_RT_MAX_HEAP_SIZE to that variable (so Chapel users don't have to use the GASNet-specific setting). That started in commit 769efae6cd16dfbe5529b2caed4c10b49237ad0e.

There's another part, though; we used to set it on Cray X* systems for some reason. From that commit message:

[...]

PR #10445, opened 2018-07-21
Upgrade to GASNet 1.32.0

ronawho:
This upgrades to GASNet 1.32.0, while keeping our traditional patch to support
our 64-bit ARM cross-configuration file.

1.32.0 includes all the upstream fixes we were pulling in (aries init bug, ibv
perf improvement, slow uncoordinated smp shutdown fix ) as well several other
bug fixes and improvements. Additionally, Windows Subsystem for Linux (WSL)
support has graduated out of BETA status. This release has also changed
GASNET_MAX_SEGSIZE to default to 85% of available memory, so we should be able
to stop setting MAX_SEGSIZE (and only override if CHPL_RT_MAX_HEAP_SIZE is set
-- https://github.com/chapel-lang/chapel/issues/8199)

For the full changelog see: http://gasnet.lbl.gov/dist/ChangeLog


ronawho:
@dmk42 just checking in to verify that we still want/need the aarch cross-configure script

dmk42:
Let's keep it for now.  I imagine our access to a machine for testing it will go away eventually.  That will be a good time to drop it.

PR #10448, opened 2018-07-22
Add join() that takes iterator expressions

### LINKED ISSUE #9372: List comprehension as input to string.join yields wrong output
### Summary of Problem
Wrong output for similar code, possibly due to function promotion of `join` method with list comprehension (Is this what it's called?) as an argument input.
```
_ir__iterator_for_loopexpr2 output: // Wrong?
one two three // Wrong
[domain(1,int(64),false)] string output:
one,two,three
```
Expectation is that both outputs should be `one,two,three`. First expression has confusing/implementation-specific type `_ir__iterator_for_loopexpr2` instead of an array of strings.

### Steps to Reproduce

**Source Code:**

```chapel
var list = ["one", "two", "three"];

writeln(([l in list] l).type:string, " output:");
writeln(",".join([l in list] l));

var x = [l in list] l;
writeln(x.type:string, " output:");
writeln(",".join(x));
```

### Configuration Information

`chpl version 1.18.0 pre-release (296301484f)`

bradcray:
Issue #9372 points out that, for people coming from Python, Chapel's
join() works in strange ways when called with for[all] expressions.  In
the discussion, I postulated that an overload which took an iterator
expression could potentially be provided to address this confusion.
This PR adds such an overload (with no attempts to optimize it for
parallelism or minimizing string operation overheads).  It also adds a
test case capturing the original cases filed on issue #9372.

I did not document the new overload for now because of the fact
that `_iteratorRecord` isn't a user-facing type.  Ultimately, it seems
that we'd want to write this routine to take types that satisfy the
`iterable` interface once we have constrained generics, traits, or the
like.

Resolves #9372.

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -960,6 +960,20 @@ module String {
       return _join(S);
     }
 
+    pragma "no doc"
+    proc join(ir: _iteratorRecord) {
+      var s: string;
+      var first: bool = true;
+      for i in ir {
+        if first then
+          first = false;
+        else
+          s += this;
+        s += i;
+      }
+      return s;
+    }
+
     proc _join(const ref S) : string where isTuple(S) || isArray(S) {
       if S.size == 0 {
         return '';
--- a/test/types/string/methods/join-loop-expr.chpl
+++ b/test/types/string/methods/join-loop-expr.chpl
@@ -1,0 +1,14 @@
+var list = ["one", "two", "three"];
+
+writeln(",".join([l in list] l));
+writeln(",".join(forall l in list do l));
+writeln(",".join(for l in list do l));
+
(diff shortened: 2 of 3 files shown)


modules/internal/String.chpl
```diff
@@ -960,6 +960,20 @@ module String {
       return _join(S);
     }
 
+    pragma "no doc"
+    proc join(ir: _iteratorRecord) {
+      var s: string;
+      var first: bool = true;
+      for i in ir {
+        if first then
```
benharsh: Couldn't you avoid this 'first' stuff by initializing 's' with 'this'?

benharsh: whoops, nevermind.

Issue #10276, opened 2018-07-11
Should the compiler type check all concrete functions?


bradcray:
Today, the Chapel compiler takes a lazy approach to resolution in which only those functions that are called are resolved.  This can sometimes lead to errors lingering in code that isn't called until the day that it first is.  For generic functions, this probably needs to be the way things are (at least in a world without constrained generics — #8629), and I suspect we got into the world we're in today in part because so much of the code we tend to write is generic.  But for completely concrete functions, should we be more aggressive in resolving them to avoid such error cases (as I believe we do for functions that are `export`ed)?

LouisJenkinsCS:
I am in favor of a more aggressive approach, so long as it can be turned on and off via compiler flags. To emulate behavior like this, I have had to create a single unit test that merely includes all other modules, and then have it compile with `--no-dead-code-elimination` which seems to work well enough, but it would be nice if I didn't have to do this. Following from that tidbit, I strongly believe that if this behavior should be disabled by default, then `start_test` should have it enabled by default.

bradcray:
A related question is whether, with this checking, we'd want to do it on unused concrete internal / standard modules by default given that it'd likely add compilation time and "should be safe" (once it's been checked once).  I might suggest not doing it for the internal / standard modules by default, having a flag to enable it, and having the `--devel` flag flip this flag by default.  And then have a smoke test that uses this flag to make sure incorrect concrete functions aren't checked into these modules.

bradcray:
Since filing this, I've been suspecting that getting a simple draft of this working would take five minutes and so put one together here:  https://github.com/bradcray/chapel/tree/resolve-concrete-functions.  As of the time of this writing, it only resolves concrete functions in "user" modules.  Running it on our test suite now to see whether it causes problems or only finds real issues.

bradcray:
While I demonstrated some cases of un-invoked concrete functions being analyzed and generating errors, running on the full test suite resulted in about 10% of the tests failing with an internal error and several extremely long compilations timing out.  Timed out myself today after only spending a few minutes on this, but hope to look at this more upon returning.

bradcray:
Spent a few more minutes on this tonight and came up with https://github.com/bradcray/chapel/tree/resolve-concrete-functions-2 which seems to be working very well (at least, I believe I've resolved the most common internal error(s) I was seeing with the previous attempt).  Didn't have access to parallel testing to try it out, though, and now it's too late for tonight.

bradcray:
[...]

PR #10449, opened 2018-07-22
Resolve concrete user functions

bradcray:
This is a quick attempt at a branch that will resolve concrete user functions as proposed in issue #10276.  I have not yet had a chance to run full testing on it, but spot-checking it, it was looking promising.  Something to finish up after vacation (making it a PR to avoid the potential for duplicated effort).

Even if it were to test cleanly, I think we'd want to add flags to extend this support to standard/internal modules (which will be nontrivial because we have error overloads that are concrete) and also potentially to package modules.

### DIFF
--- a/compiler/resolution/functionResolution.cpp
+++ b/compiler/resolution/functionResolution.cpp
@@ -8019,7 +8019,10 @@ static void resolveExports() {
     }
 
     if (fn->hasFlag(FLAG_EXPORT) ||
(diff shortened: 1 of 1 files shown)


mppf:
I'd propose to not resolve functions marked with FLAG_LAST_RESORT in this way.

bradcray:
Closing this for the time being since it's still WIP, I'm back in town and less worried about someone not knowing I was working on this.  With minimal effort, progress still looks promising.

PR #10452, opened 2018-07-23
Alter ModuleSymbol::getTopLevel* to return std::vectors instead of Vec 

### LINKED ISSUE #10243: Fix unmanaged generic argument types
PR #10171 got us part of the way there but there are problems with that PR that need to be corrected.

The following tests are failing with --force-initializers:
```
test/classes/delete-free/unmanaged/make-unmanaged-in-forall-expr-typed.chpl
test/classes/delete-free/unmanaged/make-unmanaged-in-forall-expr.chpl
test/classes/delete-free/unmanaged/unmanaged-generic2.chpl
test/classes/ferguson/delete-free/unmanaged-generic.chpl
test/functions/deitz/test_promote_generic.chpl
test/studies/amr/advection/amr/AMR_AdvectionCTU_driver.chpl
```

Additionally the checking isn't complete for the borrowed case:
``` chapel
class C {
  var x : int;
  proc init() { }
}

class G {
  type t;
  var x : t;
  proc init(type t) { this.t = t; }
}

proc foo(type t : borrowed G) {
  writeln("G");
}

proc foo(type t : object) {
  writeln("object");
}

foo(borrowed C); // prints "G" !?
foo(borrowed G(int)); // G
```

lydia-duncan:
Resolves #10243 

We've wanted to remove our Vec class for a long time, because reinventing the
wheel is not useful and a maintenance burden.  Such efforts have been on a "as
convenient" basis - I ran across these uses recently and so decided to remove
them while I was here.

This was mostly brainless


compiler/passes/docs.cpp
```diff
@@ -180,25 +181,25 @@ void printModule(std::ofstream *file, ModuleSymbol *mod, unsigned int tabs, std:
   if (!mod->noDocGen()) {
     mod->printDocs(file, tabs, parentName);
 
-    Vec<VarSymbol*> configs = mod->getTopLevelConfigVars();
+    std::vector<VarSymbol*> configs = mod->getTopLevelConfigVars();
     if (fDocsAlphabetize)
-      qsort(configs.v, configs.n, sizeof(configs.v[0]), compareNames);
-    forv_Vec(VarSymbol, var, configs) {
+      std::sort(configs.begin(), configs.end(), compareNames);
```
mppf: I don't think C++ std::sort supports passing a C function pointer. I think it's reasonable to continue to call qsort on the std::vector (the C++ standard promises this is OK). If you want to use std::sort, AFAIK you need to define a function object with a method returning bool - for <.

compiler/passes/docs.cpp
```diff
@@ -180,25 +181,25 @@ void printModule(std::ofstream *file, ModuleSymbol *mod, unsigned int tabs, std:
...
+    for_vector(VarSymbol, var, configs) {
       var->printDocs(file, tabs + 1);
     }
 
-    Vec<VarSymbol*> variables = mod->getTopLevelVariables();
+    std::vector<VarSymbol*> variables = mod->getTopLevelVariables();
     if (fDocsAlphabetize)
-      qsort(variables.v, variables.n, sizeof(variables.v[0]), compareNames);
-    forv_Vec(VarSymbol, var, variables) {
+      std::sort(variables.begin(), variables.end(), compareNames);
```
mppf: here too

compiler/passes/docs.cpp
```diff
@@ -180,25 +181,25 @@ void printModule(std::ofstream *file, ModuleSymbol *mod, unsigned int tabs, std:
...
+    for_vector(VarSymbol, var, variables) {
       var->printDocs(file, tabs + 1);
     }
-    Vec<FnSymbol*> fns = mod->getTopLevelFunctions(fDocsIncludeExterns);
+    std::vector<FnSymbol*> fns = mod->getTopLevelFunctions(fDocsIncludeExterns);
     // If alphabetical option passed, fDocsAlphabetizes the output
     if (fDocsAlphabetize)
-      qsort(fns.v, fns.n, sizeof(fns.v[0]), compareNames);
+      std::sort(fns.begin(), fns.end(), compareNames);
```
mppf: and here

compiler/passes/docs.cpp
```diff
@@ -208,19 +209,19 @@ void printModule(std::ofstream *file, ModuleSymbol *mod, unsigned int tabs, std:
       }
     }
 
-    Vec<AggregateType*> classes = mod->getTopLevelClasses();
+    std::vector<AggregateType*> classes = mod->getTopLevelClasses();
     if (fDocsAlphabetize)
-      qsort(classes.v, classes.n, sizeof(classes.v[0]), compareClasses);
+      std::sort(classes.begin(), classes.end(), compareClasses);
```
mppf: and here

compiler/passes/docs.cpp
```diff
@@ -208,19 +209,19 @@ void printModule(std::ofstream *file, ModuleSymbol *mod, unsigned int tabs, std:
...
-    forv_Vec(AggregateType, cl, classes) {
+    for_vector(AggregateType, cl, classes) {
       printClass(file, cl, tabs + 1);
     }
 
-    Vec<ModuleSymbol*> mods = mod->getTopLevelModules();
+    std::vector<ModuleSymbol*> mods = mod->getTopLevelModules();
     if (fDocsAlphabetize)
-      qsort(mods.v, mods.n, sizeof(mods.v[0]), compareNames);
+      std::sort(mods.begin(), mods.end(), compareNames);
```
mppf: and here

compiler/passes/docs.cpp
lydia-duncan: Hmm, you're right.  Guess I didn't read that carefully enough.  I'll update it, thanks!

lydia-duncan:
Okay, updated.  Second look?

lydia-duncan:
I re-ran after the latest changes, all good :)

PR #10454, opened 2018-07-23
Stop setting GASNET_MAX_SEGSIZE by default

ronawho:
Previously we sometimes set GASNET_MAX_SEGSIZE to 90% of available memory
because the old GASNet default of 2G was too small for us. We were pretty
inconsistent about when we set it, so we opened #8199 to investigate. GASNet
1.32.0 "now defaults to 85% of physical memory, split evenly between co-located
processes on a host" so we shouldn't need to set it except when a user
explicitly sets CHPL_RT_MAX_HEAP_SIZE.

This removes the default setting of MAX_SEGSIZE and chpl_sys_availMemoryBytes,
which was only used to set it.

Closes https://github.com/chapel-lang/chapel/issues/8199


ronawho:
Passed full gasnet udp-everything udp-fast and release/examples for aries-fast

PR #10456, opened 2018-07-23
Clean up LinearAlgebra.Sparse.dot tests

ben-albrecht:
- Corrected a flawed test that depended on promotion flattening to work
  - Pointed out by @vasslitvinov
- Organized existing `dot` tests such that their ordering follows a logical pattern
- Added more comments


ben-albrecht:
Test passes on OS X with `openblas` + `lapack`

PR #10459, opened 2018-07-23
Initializers on by default

benharsh:
While the core change is to simply enable ``--force-initializers`` by default, additional changes were required and desired.

The various compiler changes are made to either work around quirks in the compiler's expectations, or to allow ``--no-force-initializers`` to permit constructors in user code (not internal/standard).

This PR also removes the 'use default init' pragma from the compiler, modules, and test suite. This lead to minor line number changes. The " test/classes/initializers/compilerGenerated/pragmaOnly" directory is not useful when initializers are on by default, and so has been removed entirely.

This PR also retires a dozen or so futures.

The recent shape-preservation PR introduced some failures in 5 tests, which are futurized by this PR in order to make progress on initializers.

Testing:
- [x] local + futures
- [x] gasnet + futures
- [x] partial baseline


compiler/AST/AggregateType.cpp
```diff
@@ -2159,12 +2159,21 @@ bool AggregateType::needsConstructor() const {
...
-  else if (fUserDefaultInitializers && this != dtObject)
+  } else if (fUserDefaultInitializers == false) {
+    // Don't allow constructors for internal/standard types, except for object
+    ModuleSymbol* mod = thisNC->getModule();
+    if (this == dtObject) {
+      return true;
+    } else if (mod->modTag == MOD_INTERNAL || mod->modTag == MOD_STANDARD) {
+      return false;
+    }
+  } else if (fUserDefaultInitializers && this != dtObject) {
```
lydia-duncan: The `fUserDefaultInitializers` check is now unnecessary for this branch, since the new branch eliminates it being set to false.  You could add an assert for it in the branch if you were worried about it

test/arrays/views/initArrFieldWithSlice.bad
```diff
@@ -1,2 +1,2 @@
-initArrFieldWithSlice.chpl:31: error: unresolved call '_new(type C, [domain(1,int(64),false)] real(64))'
```
lydia-duncan: You deserve a medal, this commit must have been obnoxious to make

bradcray:
Woohoo!

PR #10482, opened 2018-07-24
Fix SPA-perf SPA record

ben-albrecht:
This test began failing after the initializers change with the error:

```
Error: unresolved call 'chpl__defaultHash(type int(64))'
```

Changing the `cols` field from generic to concrete (`domain(1)`) resolved this error.


ben-albrecht:
Tested on OS X with `openblas` + `lapack`

Issue #10246, opened 2018-07-10
Refactor HDF5 module


daviditen:
As a Chapel library author I want libraries that wrap existing C libraries to follow a consistent pattern.  The FFTW, LAPACK, MPI, etc. modules have established that the extern interface to an underlying C library should be defined within an inner module to avoid exposing it as much as possible.

Move the extern HDF5 interface into a nested module and update higher level routines as needed to account for this change.

daviditen:
Closed with PR #10347.

PR #10483, opened 2018-07-24
Update HDF5 file_id instances to be type hid_t

ben-albrecht:
After #10246, some instances of `file_id` could be passed to `C_HDF5.H5Fopen()` as an `int(64)`.

Chapel's coercion rules forbid downcasting an `int(64)` to an `int(32)` resulting in an error about calling an internal function (`C_HDF5.H5Fopen()`) to the end-user.

This PR updates instances of `file_id` to be explicitly cast to `C_HDF5.hid_t`.

### DIFF
--- a/modules/packages/HDF5.chpl
+++ b/modules/packages/HDF5.chpl
@@ -3438,7 +3438,7 @@ module HDF5 {
     var files: [BlockSpace] ArrayWrapper(eltType, rank);
     forall (f, name) in zip(files, filenames) {
       var locName = name; // copy this string to be local
-      var file_id = C_HDF5.H5Fopen(locName.c_str(), C_HDF5.H5F_ACC_RDONLY, C_HDF5.H5P_DEFAULT);
+      var file_id = C_HDF5.H5Fopen(locName.c_str(), C_HDF5.H5F_ACC_RDONLY, C_HDF5.H5P_DEFAULT): hid_t;
       var dims: [0..#rank] C_HDF5.hsize_t;
       var dsetRank: c_int;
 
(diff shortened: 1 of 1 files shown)


Spartee:
What version of HDF5 are you using @ben-albrecht? I ran into this issue with 1.8.19, but not with 1.10.1.

daviditen:
Thanks for digging into this @ben-albrecht.

I think @Spartee is right that this is a version issue.  C_HDF5 was created based on HDF5 version 1.10.2, which has:

```
typedef int64_t hid_t;
```

Older versions have:
```
typedef int hid_t;
```

If the version of "hdf5_hl.h" included the extern block at [HDF5.chpl:3338](https://github.com/chapel-lang/chapel/blob/master/modules/packages/HDF5.chpl#L3338) is from one of the older versions, it would be expecting `hid_t` to be a `c_int` within the extern block even though the rest of the module thinks it is an `int(64)`.

I don't know how casting to `C_HDF5.hid_t` (a.k.a. int(64)) would help this - it's already the return type of `C_HDF5.H5Fopen()`, so the cast shouldn't change anything.

ben-albrecht:
> I don't know how casting to `C_HDF5.hid_t` (a.k.a. int(64)) 

In my case `C_HDF5.hid_t` was an `int(32)`. My original solution involved plugging in `int(32)` explicitly, and then I thought `C_HDF5.hid_t` would be more general.

I was on an older version of `HDF5` and have updated to `1.10.1`. I'll give the tests another try at some point to confirm that was the cause of the issue (no llvm build on my machine at the moment).

daviditen:
> In my case C_HDF5.hid_t was an int(32).

That seems odd.  [HDF5.chpl:2854](https://github.com/chapel-lang/chapel/blob/master/modules/packages/HDF5.chpl#L2854) says it should be an `int(64)`.  I'd only expect hid_t to be int(32) within the extern block for old versions of HDF5.

ben-albrecht:
If we assert that the HDF5 version is `1.10+`, then this PR is irrelevant.

PR #10484, opened 2018-07-24
Update llvm.invariant.start emission for records

mppf:
Now that records are using initializers, the code generator
needed to handle a slightly different pattern for initializing
a const record so that it can add llvm.invariant.start.

Additionally the tests of this feature needed updating to match
the initializer call sequence instead of the constructor one.

- [x] full local testing

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -3497,7 +3497,7 @@ GenRet CallExpr::codegen() {
 
       // Handle setting LLVM invariant on const records after
       // they are initialized
-      if (this->isNamedAstr(astrInit)) {
+      if (fn->isInitializer()) {
         if (isUserDefinedRecord(get(1)->typeInfo())) {
           if (SymExpr* initedSe = toSymExpr(get(1))) {
             if (initedSe->symbol()->isConstValWillNotChange()) {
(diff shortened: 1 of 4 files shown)


compiler/codegen/expr.cpp
```diff
@@ -3494,6 +3494,25 @@ GenRet CallExpr::codegen() {
           ret.val = converted;
         }
       }
+
+      // Handle setting LLVM invariant on const records after
+      // they are initialized
+      if (this->isNamedAstr(astrInit)) {
```
ronawho: Can this use `isInitializer()`?

compiler/codegen/expr.cpp
```diff
@@ -3494,6 +3494,25 @@ GenRet CallExpr::codegen() {
...
+        if (isUserDefinedRecord(get(1)->typeInfo())) {
```
ronawho: Why user defined records only?

ronawho: Ah nevermind, I see that this is just a poorly named function now with a TODO to change it

mppf: isUserDefinedRecord doesn't actually mean user defined records - it started out that way but it actually now means "Records for which we manage creation and destruction in the new way". E.g. range and tuples are "user defined records". There is a TODO comment that it should be renamed "isMemoryManagedRecord" or something along those lines.

compiler/codegen/expr.cpp
mppf: as you wish

PR #10486, opened 2018-07-24
Mason: Add skipif for pkgconfig integration

Spartee:
The skipif for the pkg-config integration tests needed a check for CHPL_COMM and CHPL_REGEXP. This PR should quiet all noise on the nightly tests.


test/mason/pkgconfig-tests/SKIPIF
```diff
@@ -1,22 +1,41 @@
 #!/bin/bash
 
 
-skip="False"
+skipZlib="False"
```
ben-albrecht: While we're here, it might make this `SKIPIF` a lot more understandable if we renamed these variables something like:

`has_zlib` or `zlib_available`

Issue #10497, opened 2018-07-24
Error for assignment between arrays of different size disabled by --no-bounds-checks


benharsh:
Consider the following program:

```chpl
var A : [1..4] int = [1, 2, 3, 4, 5];
writeln(A);
```

This program will encounter an error at runtime when compiled with zero flags:
```
error: halt reached - assigning between arrays of different shapes in dimension 1: 4 vs. 5
```

When compiled with ``--no-bounds-checks`` the program will run without error and print:
```
1 2 3 4
```

My primary question is: Should ``--no-bounds-checks`` allow assignment between arrays of different sizes?

Since this error should be relatively cheap to check, it seems like it could be worth allowing regardless of bounds-checking.

-------------------------

The original case I encountered involved loop-exprs. Note that the length of ``A`` and the loop expr are different because of a missing ``#`` operator:
```chpl

class C {
  var x : int;

  proc init(x : int) {
    this.x = x;
    writeln("C.init: ", x);
  }

  proc deinit() { writeln("C.deinit: ", x); }
}

class X {
  var len : int = 3;
  var A : [0..#len] C = [i in 0..len] new C(i);

  proc deinit() {
    delete A;
  }
}

var x = new X();
delete x;
```

When compiled with zero flags this program encounters an error:
```
C.init: 0
C.init: 1
C.init: 2
C.init: 3
foo.chpl:22: error: halt reached - assigning between arrays of different shapes in dimension 1: 3 vs. 4
```

When bounds checks are disabled, the program runs successfully but we fail to delete an allocated ``C``:
```chpl
C.init: 0
C.init: 1
C.init: 2
C.init: 3
C.deinit: 0
C.deinit: 1
C.deinit: 2
```

bradcray:
This seems related to the classic "we don't do good size checking for zippered forall loops" issue (for which I can't find an open issue quickly... maybe it's never been captured as one?) in which we permit:

```chapel
forall (i,j) in zip(1..4, 1..5) do ...
```

without errors (which is to say, I think a fix for the longstanding issue should also fix this issue).

That said, in this specific context, I think I agree with you that we're doing a sufficiently heavyweight operation—and on sufficiently known-size types—that always doing the checking is appropriate to address the potential confusion in this case without waiting for the more general solution.

ben-albrecht:
Similarly, I was surprised by the behavior of this program:

```chpl
const A = [1,2];
const B = [3,4,5];

const AB = A + B;
writeln(AB);
[...]

PR #10499, opened 2018-07-24
 Fix size of loop expr in LCALS

benharsh:
The array field was of size "0..#N" but the loop expr was of size "0..N". We didn't catch this during correctness testing because the test is compiled with ``--no-bounds-checks``, which disables the error message for assignment between arrays of different size.

Testing (of lcals only)
- [x] local
- [x] memleaks


benharsh:
@daviditen : pinging you for a post-merge review given your past ownership of LCALS. I have opened #10497 to discuss whether the array assignment error message should still be emitted even if bounds checks are disabled.

daviditen:
Thanks @benharsh, looks good!

PR #10501, opened 2018-07-24
add back --warn-unstable warnings for arrays, type ctors

mppf:
PR #10460 stopped --warn-unstable from warning on certain common array
and type construction patterns. This PR adds back the error for the
common case (while leaving arbitrary type arguments unchecked).

- [x] full local testing

### DIFF
--- a/compiler/passes/scopeResolve.cpp
+++ b/compiler/passes/scopeResolve.cpp
@@ -880,15 +880,27 @@ static Expr* handleUnstableClassType(SymExpr* se) {
           if (DefExpr* d = toDefExpr(cur))
             inDef = d;
         }
-        // Find outer call, not counting baseExpr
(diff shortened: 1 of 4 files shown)


compiler/passes/scopeResolve.cpp
```diff
@@ -880,15 +880,27 @@ static Expr* handleUnstableClassType(SymExpr* se) {
           if (DefExpr* d = toDefExpr(cur))
             inDef = d;
         }
-        // Find outer call, not counting baseExpr
+        // Find outer call, but don't count:
+        //  * baseExpr
+        //  * calls to types
+        //  * not counting baseExpr
```
benharsh: This comment is a little confusing for the last bullet. My brain parses it as "but don't count not counting baseExpr".

PR #10504, opened 2018-07-24
Allow foralls over non-yielding iterators.

### LINKED ISSUE #10357: Compiler internal error when not yielding from parallel iterator
When not yielding from a parallel iterator, the following error occurs: `internal error: assertion error [ForallStmt.cpp:292]`

**MWE**

```chpl
iter badIterator() {}
iter badIterator(param tag : iterKind) where tag == iterKind.standalone {}

forall x in badIterator() do ;
```
[TIO](https://tio.run/##S85ILEjN@f8/syS1SCEpMcUTSCeW5BdpaCpU13JhiBYkFiXmKpQkpitYKYAkvTPzUjQVyjNSi1LBora2cGG94pLEvJTEnPy8VJBRXGn5RYk5OQoVCpl5aBal5CtY//8PAA) (Note that it doesn't crash on release 1.17)

**Compiler Debugging**

Looking at the actual assertion that it fails, [`fs->inductionVariables().length == 1`](https://github.com/chapel-lang/chapel/blob/master/compiler/AST/ForallStmt.cpp#L292), I ran `gdb` to determine the actual length of `inductionVariables()` which happens to be `fIterVars`...

```
(gdb) print fs->fIterVars
$2 = {head = 0x0, tail = 0x0, parent = 0x289f450, length = 0}
```

Hopefully that can help assist a bit in the resolution of this issue. Since the error is easy enough to get around, its definitely of low priority.

vasslitvinov:
When a parallel iterator does not contain a yield statement,
the corresponding index variable is considered to be of void type.
So it was getting removed during cleanupVoidVarsAndFields().
This change prevents such removal.

Resolves #10357.

### DIFF
--- a/compiler/AST/ForallStmt.cpp
+++ b/compiler/AST/ForallStmt.cpp
@@ -257,6 +257,15 @@ ForallStmt* enclosingForallStmt(Expr* expr) {
   return NULL;
 }
 
+// Is 'expr' the DefExpr of an induction variable in some ForallStmt?
+bool isForallIterVarDef(Expr* expr) {
+  if (expr->list != NULL)
+    if (ForallStmt* pfs = toForallStmt(expr->parentExpr))
+      if (expr->list == &pfs->inductionVariables())
+        return true;
+  return false;
+}
+
 // Is 'expr' an iterable-expression for some ForallStmt?
 bool isForallIterExpr(Expr* expr) {
   if (expr->list != NULL)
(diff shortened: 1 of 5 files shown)


bradcray:
Tagging @daviditen on this since it seems tied to his recent efforts, he was out last week, and it doesn't seem to have been reviewed (?).

daviditen:
I think this looks good.

PR #10506, opened 2018-07-25
Allow the Allinea compiler to compile the Chapel runtime

### LINKED ISSUE #8743: Document Allinea compiler support
As a Chapel user, I want to see documentation on all the back-end compilers that are fully supported by Chapel.

#8742 adds experimental support for the Allinea compiler, ARM's commercial compiler based on Clang.  Support is not complete, so that PR does not document it.  When support is complete, we should add Allinea to `doc/rst/usingchapel/chplenv.rst` and possibly elsewhere.

### LINKED ISSUE #10474: Finish support for Allinea compiler on ARM and document it
As a Chapel developer, I want to be able to use the Allinea compiler as a target C compiler on ARM because diversity in target compilers helps rule out the C back end as a cause of problems (see #10232 ).

#8742 added most of the support for the Allinea compiler.  Unless relevant changes have occurred in the mean time, it should be sufficient to turn off some warning messages that were hindering the build.  Therfore, in the best case, this is trivial.

The result then needs to be documented as mentioned in #8743 .

The reason for doing this right now is that it will help narrow down the cause of the lulesh timeouts on ARM.

dmk42:
#8742 added most of the support for the Allinea compiler (ARM's HPC compiler).  This change adds the final tweak to allow it to work.  It also updates the documentation to include Allinea.

Closes #10474 
Closes #8743

### DIFF
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -139,6 +139,7 @@ Building Chapel for a Cray System from Source
    On a Cray X-series system, ensure that you have one of the following
    Programming Environment modules loaded to specify your target compiler::
 
+       PrgEnv-allinea (ARM only)
        PrgEnv-cray
        PrgEnv-gnu
        PrgEnv-intel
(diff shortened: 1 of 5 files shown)


dmk42:
@ronawho - Would you mind taking a post-commit look at this?

dmk42:
@ronawho - Thanks.  I checked chpl_compiler.py:get() and it did not need to be updated, but the other two did.  I've opened a new PR for that, #10529 .

PR #10509, opened 2018-07-25
Fix testReplWrites

mppf:
This is a follow-on to PR #8079.

chpl_serialReadWriteRectangular is declared in DefaultRectangular and
then ReplicatedDist declares a version of this method with a 'where'
clause. Current resolution rules prefer the more visible function (rather
than the one with a 'where' clause) and PR #8079 made this sense of
"visibility" no longer include point of instantiation (to solve the first
"Hijacking" case from CHIP 20). 

So, after PR #8079, the DefaultRectangular version is more visible to a
call site in ArrayViewSlice. Applying "last resort" to this function
requests the compiler prefer another overload if one is applicable.
Overload sets (or a similar idea) would be a better user-facing way to
solve this problem (see CHIP 20).

- [x] full local testing
- [x] full GASNet testing

### DIFF
--- a/modules/dists/ReplicatedDist.chpl
+++ b/modules/dists/ReplicatedDist.chpl
@@ -539,7 +539,6 @@ proc ReplicatedArr.dsiAccess(indexx) ref {
 
 // Write the array out to the given Writer serially.
 proc ReplicatedArr.dsiSerialWrite(f): void {
-  //  writeln("in dsiSerialWrite() on locale ", here.id);
   localArrs[f.readWriteThisFromLocale().id].arrLocalRep._value.dsiSerialWrite(f);
 }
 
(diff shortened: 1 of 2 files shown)


modules/dists/ReplicatedDist.chpl
```diff
@@ -539,7 +539,6 @@ proc ReplicatedArr.dsiAccess(indexx) ref {
 
 // Write the array out to the given Writer serially.
 proc ReplicatedArr.dsiSerialWrite(f): void {
-  //  writeln("in dsiSerialWrite() on locale ", here.id);
```
benharsh: Leftover debugging comment?

mppf: I think so, but this PR *removes* it.

benharsh: Oops, my mistake!

Issue #10508, opened 2018-07-25
Remaining release/examples memory leaks


ronawho:
Switching to initializers by default (https://github.com/chapel-lang/chapel/pull/10459) resolved memory leaks for SSCA#2, which had been the single largest leak across the entire test suite.

![leaks](https://user-images.githubusercontent.com/1588337/43201760-627de128-8fe7-11e8-94f3-2aeab5f0d12d.png)

Now that the SSCA leak has been resolved we're down to 10 tests and only ~2,000 bytes being leaks for release/examples. That's tantalizingly close to having leak free example tests so I wanted to characterize the remaining leaks: 

User leaks (unmanaged domain maps):

 - [x] hpl.chpl
 - [x] distributions.chpl

Anonymous module-level arrays-of-arrays leak (#5660):

 - [x] isx.chpl
 - [x] meteor.chpl
 - [x] opaque-domains.chpl
 - [x] arrays.chpl
 - [x] decl-with-anon-domain.chpl
 - [x] meteor-fast.chpl

Anonymous module-level array type alias leak (possibly #5660 too):

 - [x] typeAliases.chpl

bradcray:
Just noting that there are [a lot of open issues](https://github.com/chapel-lang/chapel/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+leaks+in%3Atitle) that try to characterize / categorize leaks opened by @benharsh and others which may be of use, or perhaps should be re-evaluated in an initializers-based world/

bradcray:
The remaining cases relate to issue #12732.

PR #10511, opened 2018-07-25
Remove user memory leaks from release/examples tests

ronawho:
Convert hpl.chpl and distributions.chpl to use owned. Delete unmanaged classes
in linkedList.chpl (I tried switching to use shared, but that had leaks still I
think as a result of contains() doing an early break out of an iterator.)

Related to https://github.com/chapel-lang/chapel/issues/10508

### DIFF
--- a/test/release/examples/benchmarks/hpcc/hpl.chpl
+++ b/test/release/examples/benchmarks/hpcc/hpl.chpl
@@ -76,8 +76,8 @@ proc main() {
   //
   const MatVectSpace: domain(2)
     dmapped DimensionalDist2D(targetLocales,
-                              new unmanaged BlockCyclicDim(gridRows, lowIdx=1, blkSize),
-                              new unmanaged BlockCyclicDim(gridCols, lowIdx=1, blkSize))
+                              new owned BlockCyclicDim(gridRows, lowIdx=1, blkSize),
+                              new owned BlockCyclicDim(gridCols, lowIdx=1, blkSize))
                     = {1..n, 1..n+1},
         MatrixSpace = MatVectSpace[.., ..n];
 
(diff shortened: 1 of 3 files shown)


ronawho:
https://github.com/chapel-lang/chapel/compare/master...ronawho:linkedList-shared was my other attempt to switch linkedList.chpl to shared, but I had to get rid of an early return for the `contains()` otherwise there was still a leak

test/release/examples/benchmarks/hpcc/hpl.chpl
```diff
@@ -231,8 +231,8 @@ proc replicateD1(Ab, BD) {
 proc replicateD2(Ab, AD) {
   const replAD = {1..n, 1..blkSize}
     dmapped DimensionalDist2D(targetLocales,
-                              new unmanaged BlockCyclicDim(gridRows, lowIdx=1, blkSize),
-                              new unmanaged ReplicatedDim(gridCols));
+                              new owned BlockCyclicDim(gridRows, lowIdx=1, blkSize),
+                              new owned ReplicatedDim(gridCols));
```
mppf: AFAIK DimensionalDist2D thinks it's using unmanaged for the different dimensions. It ought to use `owned` on them, I'm just surprised this will work without changing something in DimensionalDist2D. Can you double check that it's freeing the DimensionalDist2D instance that frees them, and not just that this function goes out of scope? E.g. write a test that uses this pattern and returns the DimensionalDist2D (or an array / domain over it).

ronawho: Looks like what I had was broken -- switched to pulling the Dim declarations out

ronawho:
Closing for now since these tests fail under `--no-local`.

mppf:
@ronawho - maybe put the list changes in a separate PR?

ronawho:
Yeah, will do

PR #10512, opened 2018-07-25
Fix problem where instantiationPoint might be removed by tryToken

mppf:
@ben-albrecht was seeing a problem where `=` was failing to resolve for an
array of a user-defined record type. The issue was that the instantiationPoint
for a relevant function to find the visible `=` overload was no longer in the
tree. (Unfortunately we havn't yet been able to make a small reproducer).

This PR:
 * adjusts logic finding tryToken conditionals to find nested blocks
   (this was the cause of the original bug)
 * renames --break-on-delete-id to --break-on-remove-id since it breaks
   on remove, not when the AST node is freed ( I used this a lot in
   debugging these problems )
 * adds a 1-argument BlockTag constructor for BlockStmt
 * adjusts several BlockStmts that are intentionally created to be
   removed to be scopeless
 * instead of removing fn's body, adjusted several code paths to replace
   the body stmts, so that the body can be a valid instantiation point
 * adds an assert that instantiationPoint is in the tree in
   getVisibleFunctions to catch this issue earlier next time.
 * adds a backupInstantiationPoint, which is the FnSymbol that originally
   held instantiationPoint. I don't know of another way to resolve
   instantiationPoints missing from replaceEflopiWithForall but I'm sure
   another strategy could be used... But since this issue seems to come
   up a lot, a more robust strategy seemed called for. Made
   instantiationPoint and backupInstantiationPoint private and now use
   instantiationPoint() to read whichever is available and
   setInstantiationPoint() to set them.

### DIFF
--- a/compiler/AST/FnSymbol.cpp
+++ b/compiler/AST/FnSymbol.cpp
@@ -522,6 +522,24 @@ Symbol* FnSymbol::replaceReturnSymbol(Symbol* newRetSymbol, Type* newRetType) {
   return retval;
 }
 
+// Removes all statements from body and adds all statements from block.
+void FnSymbol::replaceBodyStmtsWithStmts(BlockStmt* block) {
+  for_alist(stmt, this->body->body) {
+    stmt->remove();
+  }
+  for_alist(stmt, block->body) {
+    this->body->insertAtTail(stmt->remove());
+  }
+}
+
+// Removes all statements from body and adds expr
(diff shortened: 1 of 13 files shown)


ben-albrecht:
Thanks for the fix @mppf!

Is this merge dependent on my test case being added?

mppf:
I'd like to include the test case, but if it's too hard we can stop trying...

PR #10515, opened 2018-07-25
Address pkg-config nightly failures

Spartee:
This PR changes the message produced by Mason Update upon not finding a mason registry so that moving forward we no longer have to deal with the sed replacements for testing. This PR will address the failures from the pkg-config integration.

### DIFF
--- a/tools/mason/MasonUpdate.chpl
+++ b/tools/mason/MasonUpdate.chpl
@@ -144,7 +144,7 @@ proc updateRegistry(tf: string, args : [] string) {
       const localRegistry = registryHome;
       mkdir(localRegistry, parents=true);
       const cloneRegistry = 'git clone -q ' + registry + ' .';
-      writeln("Initializing mason-registry for ", name);
+      writeln("Updating mason-registry for ", name);
       gitC(localRegistry, cloneRegistry);
     }
   }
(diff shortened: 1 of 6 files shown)


tools/mason/MasonUpdate.chpl
```diff
@@ -42,7 +42,7 @@ private var failedChapelVersion : [1..0] string;
 
 /* Finds a Mason.toml file and updates the Mason.lock
    generating one if it doesnt exist */
-proc UpdateLock(args: [] string, tf="Mason.toml", lf="Mason.lock") {
+proc UpdateLock(args: [] string, tf="Mason.toml", lf="Mason.lock", quiet=false) {
```
ben-albrecht: I'm not a fan of adding special-case arguments to the mason interface for testing purposes.

Long term, it'd probably be better to just support a `--quiet` flag, such that it's included in both the user and internal interface. In that case, `quiet` would be contained in `args`, and we could use it for most of testing.

Until then, I'd continue to use the `sed` commands in our prediff from before.

ronawho: Alternatively we could just unify the message for initializing and updating and get rid of the prediff.

 e.g. always say "Updating" (or something) even if the repo is technically being initialized

ben-albrecht: @ronawho had a good idea: Change the "Initializing mason-registry for mason-registry" to just say "Updating mason-registry for mason-registry" too, so that we can just always expect that string, and drop all the prediffs.

PR #10519, opened 2018-07-25
Get fielderator working on unmanaged classes

mppf:
In converting the test
```
test/param/ferguson/fielderator-user-class.chpl
```
to pass with --warn-unstable, it occurred to me to add a section to it using an unmanaged class type and trying to get fields from it. That caused a bunch of easy-to-fix compilation errors that this PR fixes.

- [x] full local testing


compiler/resolution/preFold.cpp
```diff
@@ -742,12 +739,10 @@ static Expr* preFoldPrimOp(CallExpr* call) {
     }
 
   } else if (call->isPrimitive(PRIM_NUM_FIELDS)) {
-    SymExpr*       se1        = toSymExpr(call->get(1));
-    AggregateType* classType  = toAggregateType(se1->symbol()->type);
+    Type*          t          = canonicalClassType(call->get(1)->getValType());
```
lydia-duncan: It seems a shame to have all these pairs of `Type t = canonicalClassType(...); classType = toAggregateType(t);` pairs.  Should canonicalClassType return an AggregateType?  Or should there be a version that always gives you an AggregateType?

mppf: I think it's fine to have the pairs, in the same way that getValType and toAggregateType are different operations. Nonetheless we could certainly add a canonicalAggregateType or something.

PR #10524, opened 2018-07-25
Generate a helper makefile stub with --library compilation + clean ups

lydia-duncan:
In order to link to a generated Chapel library file, the user has to
include our relevant runtime and third-party headers and library files,
which can be a bit of a pain to write out, even using the handy
`compileline` tool.  This change generates a file `Makefile.chpl_lib`
which can be included in another Makefile.  `Makefile.chpl_lib` contains
variables for the necessary compilation and linking flags, as well as the
compiler and linker settings which were used.  `Makefile.chpl_lib` is only
generated when `--makefile` is used as a compiler flag alongside
`--library`.

While I was working on this, I decided to split the library functionality into
its own file in the codegen directory.

Known limitation: this does not adjust for when the library, header and/or
Makefile live in a different directory from the program linking to it.  I'm
intending to test and adjust for that when I add a generated library
directory (which seemed too big to do in this commit).

Testing this is a bit of a pain.  I chose to write a sample test and Makefile
that used the generated library and `Makefile.chpl_lib` to demonstrate
that the set up worked.  I could have alternatively checked against the
exact contents for the generated Makefile but that would vary on a
system-by-system basis anyways and wouldn't guarantee that it was
*correct* there.


lydia-duncan:
Travis failure is: `library.cpp:100:10: error: ‘std::string’ has no member named ‘pop_back’`.  Only present in C++11

lydia-duncan:
@mppf - care to review?

mppf:
From the PR message ---  Makefile.chpl_lib  -- I'd expect it to be named Makefile.<libraryname> or just Makefile if we're putting it in a <libraryname> directory.

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,141 @@
...
+  }
+}
+
+// Helper function to avoid unnecessary repetition when getting information
+// from compileline
+static std::string getCompilelineOption(std::string option) {
+  std::string fullCommand = "$CHPL_HOME/util/config/compileline --" + option;
+  char buffer[128];
+  std::string res = "";
+  FILE* compilelineCall = popen(fullCommand.c_str(), "r");
```
mppf: don't we already have something like this? Shouldn't this just call `runCommand` from files.cpp ?

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,141 @@
...
+  // TODO: compileline --includes-and-defines adds -I. to the list
+  // automatically.  If the library is in a different directory from the
+  // Makefile that is using this one, that won't be sufficient to find the
+  // header file.  I will handle this with the output directory change
+  std::string cflags = getCompilelineOption("cflags");
+  cflags.erase(cflags.length() - 1); // remove trailing newline
+  std::string includes = getCompilelineOption("includes-and-defines");
+  fprintf(makefile.fptr, "CHPL_CFLAGS = %s %s\n",
+          cflags.c_str(),
+          includes.c_str());
```
mppf: These CFLAGS have a lot of paths relative to CHPL_HOME. It would be nice to shorten them by using CHPL_HOME etc in them. `util/config/replace-paths.py` already does this in a way appropriate for installed Chapel (even with separate bin and share and lib!). You could add an option to compileline to run replace-paths on the output or just run it yourself here in getCompilelineOption. Either way, I think this change will make working with the Makefile much easier:
 * easier to understand as a human
 * more resilient for the Chapel installation path changing

You could supply CHPL_HOME etc as Makefile variables in your generated Makefile. AFAIK the user will still be able to override them.

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+  pclose(compilelineCall);
+  return res;
+}
+
+void codegen_library_makefile() {
+  fileinfo makefile;
+  // TODO: alter location to use generated library directory
+  openCFile(&makefile, "Makefile.chpl_lib", "", false);
```
mppf: As I said in my other comment, if it's not in a --library-dir, I'd expect this Makefile to have the library name in it somewhere.

compiler/main/driver.cpp
```diff
@@ -986,6 +987,7 @@ static ArgumentDescription arg_desc[] = {
  {"localize-global-consts", ' ', NULL, "Enable [disable] optimization of global constants", "n", &fNoGlobalConstOpt, "CHPL_DISABLE_GLOBAL_CONST_OPT", NULL},
  {"local-temp-names", ' ', NULL, "[Don't] Generate locally-unique temp names", "N", &localTempNames, "CHPL_LOCAL_TEMP_NAMES", NULL},
  {"log-deleted-ids-to", ' ', "<filename>", "Log AST id and memory address of each deleted node to the specified file", "P", deletedIdFilename, "CHPL_DELETED_ID_FILENAME", NULL},
+ {"makefile", ' ', NULL, "Generate a makefile to help use the generated library", "F", &fLibraryMakefile, NULL, NULL},
```
mppf: Can we call it --library-makefile please? We generate Makefiles for other purposes and I don't want people to be confused by the flag name.

lydia-duncan:
So for a library `libfoo.a` would you expect it to be named `Makefile.foo` or Makefile.libfoo`?  Either way, I would assume `bar.a` (compiled with `-o bar`) would go to `Makefile.bar`

lydia-duncan:
I'll probably lean towards that naming instead of leaving it just `Makefile` in the generated library, in case the user specifies an already existing directory with a Makefile they intend to use

compiler/include/files.h
```diff
@@ -53,7 +53,8 @@ const char* objectFileForCFile(const char* cfile);
 
 const char* genIntermediateFilename(const char* filename);
 
-void openCFile(fileinfo* fi, const char* name, const char* ext = NULL);
+void openCFile(fileinfo* fi, const char* name, const char* ext = NULL,
```
mppf: I don't think you should modify openCFile in this way. Instead, can you either:
 * call openfile and closefile directly, or
 * add something like openOutputFile / openOtherFile that calls openfile/closefile ?

Also I just noticed that closeCFile does not check the return code from fclose. Could you modify it to call closefile instead of fclose, while you're here?

compiler/util/files.cpp
```diff
@@ -327,13 +328,18 @@ void closefile(fileinfo* thefile) {
...
     fi->filename = astr(name, ".", ext);
   else
     fi->filename = astr(name);
 
-  fi->pathname = genIntermediateFilename(fi->filename);
+  if (useTmpDir) {
+    fi->pathname = genIntermediateFilename(fi->filename);
+  } else {
+    fi->pathname = astr(name);
+  }
```
mppf: I don't think you should change this function to make Makefiles - it's for making C files - see my other comment in the header

lydia-duncan:
> in case the user specifies an already existing directory with a Makefile they intend to use

That train of thought came from thinking that `lib` directories are a pretty well known thing, so I wouldn't expect us to have the same level of control over its contents in the same way that we do over the directory generated with --savec, for instance.

That may be jumping to conclusions though, thoughts?

mppf:
> So for a library libfoo.a would you expect it to be named Makefile.foo or Makefile.libfoo? Either way, I would assume bar.a (compiled with -o bar) would go to Makefile.bar

Either way would be fine with me. It could also be reasonable named Makefile.include.foo or foo.mk (i've seen this convention before too). GASNet makes makefile snippets to be included so might be a reasonable thing to draw precedent/inspiration from.

compiler/include/files.h
lydia-duncan: Sorry, should have made it more clear: I was viewing this change as temporary - I'll make my own version for when I add the library directory and then all the generated library files will use that version instead.

lydia-duncan: But I'm happy to make it now and adjust that new function when I add the library directory functionality

mppf:
I'm having trouble finding any documented convention or best practice for this, but `foo.mk` seems closest to one.
 * it's used as an example in the Makefile docs for including other makefiles
 * I've seen it before for this purpose  (I think GASNet does it this way)

I see GitHub projects of Makefile bits that use either Makefile.something or something.mk conventions.

lydia-duncan:
Ben expressed a preference for `Makefile.foo` or `Makefile.libfoo`

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// Helper function to avoid unnecessary repetition when getting information
+// from compileline
+static std::string getCompilelineOption(std::string option) {
+  std::string fullCommand = "$CHPL_HOME/util/config/compileline --" + option;
+  fullCommand += "> cmd.out.tmp";
+  runCommand(fullCommand);
+  std::string replace = "$CHPL_HOME/util/config/replace-paths.py ";
+  replace += "--fixpath '$(CHPL_HOME)' $CHPL_HOME < cmd.out.tmp";
```
mppf: This might be good enough for this PR, but I'm pretty sure this will have problems with installed Chapel with ./configure --prefix. 

See

https://github.com/chapel-lang/chapel/blob/00dba22bb3e8eebe1372a0851f75b5acd1c940f1/runtime/Makefile.config#L43-L47

Of course you could look at making the script handle such things with --all or something...

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+  std::string res = runCommand(replace);
+  std::string cleanup = "rm cmd.out.tmp";
+  runCommand(cleanup);
```
mppf: 👍

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,196 @@
...
+  if (useTmpDir) {
+    fi->pathname = genIntermediateFilename(fi->filename);
+  } else {
+    fi->pathname = astr(fi->filename);
+  }
+  openfile(fi, "w");
+}
+
+void closeLibraryHelperFile(fileinfo* fi, bool beautifyIt) {
+  closefile(fi->fptr);
```
mppf: I'm not so sure it makes sense to beautify a Makefile?

compiler/main/driver.cpp
```diff
@@ -983,6 +984,7 @@ static ArgumentDescription arg_desc[] = {
  {"ignore-errors-for-pass", ' ', NULL, "[Don't] attempt to ignore errors until the end of the pass in which they occur", "N", &ignore_errors_for_pass, "CHPL_IGNORE_ERRORS_FOR_PASS", NULL},
  {"library", ' ', NULL, "Generate a Chapel library file", "F", &fLibraryCompile, NULL, NULL},
  {"library-header", ' ', "<filename>", "Name generated header file", "P", libmodeHeadername, NULL, NULL},
+ {"library-makefile", ' ', NULL, "Generate a makefile to help use the generated library", "F", &fLibraryMakefile, NULL, NULL},
```
mppf: 👍

compiler/util/files.cpp
```diff
@@ -338,7 +339,7 @@ void openCFile(fileinfo* fi, const char* name, const char* ext) {
 }
 
 void closeCFile(fileinfo* fi, bool beautifyIt) {
-  fclose(fi->fptr);
+  closefile(fi->fptr);
```
mppf: 👍

compiler/codegen/library.cpp
```diff
@@ -0,0 +1,196 @@
...
+  std::string linker = getCompilelineOption("linker");
+  fprintf(makefile.fptr, "CHPL_LINKER = %s\n", linker.c_str());
+
+  std::string linkerShared = getCompilelineOption("linkershared");
+  fprintf(makefile.fptr, "CHPL_LINKERSHARED = %s", linkerShared.c_str());
+
+  closeLibraryHelperFile(&makefile);
+}
+
+const char* getExtension() {
```
mppf: Can we name this getOutputExtension or getLibraryExtension or something? It's a little to vague a name to be used in other (non-library-codegen) files.

compiler/codegen/library.cpp
lydia-duncan: Mostly thinking about switching the header to use it as well.  I can send false in for this case.

compiler/codegen/library.cpp
lydia-duncan: Do we have a nightly testing environment for ./configure?

mppf: no

PR #10529, opened 2018-07-26
Allinea follow-up

dmk42:
After #10506 , add the Allinea compiler to two additional locations.


dmk42:

Issue #7533, opened 2017-10-03
Pool qthread guard pages


ronawho:
Qthreads uses memory pools to avoid concurrent calls to malloc for performance reasons. These pools are used for some internal book keeping structures and for task stacks. While the actual memory for stacks is pooled, guard pages are not.  As a result, there are 4 mprotect() calls per task (2 on allocation, 2 on de-allocation.) This is relatively expensive on normal processors, but on Xeon Phi concurrent mprotect() calls are particularly expensive. I believe this is some combination of the hardware and more likely the OS/software stack.

For more information on this cost see https://github.com/chapel-lang/chapel/pull/2849#issuecomment-155183504 and https://github.com/chapel-lang/chapel/pull/2849#issuecomment-155523708.

Same issue in the qthreads repo: https://github.com/Qthreads/qthreads/issues/45

dmk42:
This issue also affects arm-thunderx (96 cores per locale) and arm-thunderx2 (64 cores per locale) - see #10530 .

On arm-thunderx2, it appears to be more than just a performance problem.  Intermittently, lulesh can hang indefinitely with all but two of its threads in `mprotect()`.  This may indicate a kernel or hardware issue that should be investigated.

ronawho:
We disabled guard pages on thunderx in https://github.com/chapel-lang/chapel/pull/11049 and Xeon Phi is dead. This might still be nice, but incredibly low priority and there's an existing issue against qthreads so I don't see a reason to keep this open.

PR #10530, opened 2018-07-26
Fix timeouts for high core-count systems

### LINKED ISSUE #10475: Spike: Conclude investigation of sporadic lulesh timeouts on ARM
#10232 mapped out most of the conditions under which the timeouts occur.  The following are needed to tie up loose ends:

* Try PrgEnv-allinea for additional assurance that this is not backend-compiler related.
* Try with and without atp loaded. That is implemented with a workaround on ARM.

If Slurm remains the only variable that makes a difference, consider applying #10201 .

dmk42:
In some tests, high core-count systems spend most of their time in the `mprotect()` system call when guard pages are used in qthreads.  The arm-thunderx has 96 cores per locale, and arm-thunderx2 has 64.  This change turns off guard pages for two tests where this makes a big difference.

One remaining item to investigate is that on arm-thunderx2, there appears to be more than just a performance problem.  Intermittently, high contention in the `mprotect()` system call can lead to an application hanging indefinitely, which might be an indication of a kernel bug.

Closes #10475

### DIFF
--- a/test/release/examples/benchmarks/lulesh/lulesh.execenv
+++ b/test/release/examples/benchmarks/lulesh/lulesh.execenv
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Guard pages are expensive on Xeon Phi. See issue #7533
+# Guard pages are expensive on high core-count systems. See issue #7533
 
 import os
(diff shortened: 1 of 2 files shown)


ronawho:
Update the execenv comment to say `# Guard pages are expensive on high core-count systems. See issue #7533`?

And add a comment to that issue as well

PR #10545, opened 2018-07-26
chpl_launchcmd.py should wait a bit for qsub output file to appear

awallace-cray:
Under certain conditions, chpl_launchcmd may throw an exception
(and thus a false test error) because it looks for the qsub output
file before that file actually exists on the file system where
chpl_launchcmd is running. Maybe because chpl_launchcmd is running
on an "elogin" machine, and the NFS or whatever file transport is
slow to deliver the file from the real login node to the elogin
machine. If chpl_launchcmd allowed a little extra time for the
qsub output file to show up, everything might turn out Okay.

This change implements such a scheme with a max extra waiting time
of 30 seconds. If any extra waiting time occurred (ie, the output
file was not available immediately), a log.debug message is issued.

The extra waiting time (if any) is subtracted from the execution
time reported by sub_test as
"launchcmd reports elapsed execution time"

### DIFF
--- a/util/test/chpl_launchcmd.py
+++ b/util/test/chpl_launchcmd.py
@@ -308,31 +308,40 @@ class AbstractJob(object):
                  status(job_id) can raise a ValueError, which can indicate that
                  the job has completed *and* been dequeued. If the output file
                  exists and the job has been dequeued, it is safe to assume it
-                 completed. Otherwise we raise the error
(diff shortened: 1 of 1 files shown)


awallace-cray:
This is tested and solves the problem on kay-elogin2.  We hope the problem will be fixed properly by the system admins, and we won't need to apply this.

awallace-cray:
See PR #10585

PR #10548, opened 2018-07-26
Enable forall-expressions as field defaults

### LINKED ISSUE #10481: Resolve --force-initializers failures due to shape preservation
This task is to resolve failures in the following tests, which are currently futures on master:

```
test/classes/bradc/arrayInClass/arrayDomInClassRecord-illegal.chpl
test/classes/bradc/arrayInClass/arrayDomInClassRecord-simpler.chpl
test/classes/bradc/arrayInClass/arrayDomInClassRecord.chpl
test/classes/bradc/arrayInClass/arrayDomInClassRecord2.chpl
test/users/shetag/inlineArrInitInClass.chpl
```

vasslitvinov:
This PR enbles the compiler to handle forall-expressions
as the default expressions for class/record fields, ex.

   record R { var myField = [idx in ...] ...; }

With #9994, the above initialization is done via the newly-added
overload of chpl__initCopy(ir: _iteratorRecord) that does this:

    forall (r, src) in zip(result, ir) do ....

where `ir` is the serial iterator for a forall expression.

When resolving the forall, the compiler looks for the follower
iterator for `ir` by calling _toFollowerZip(), which eventually calls
PRIM_TO_FOLLOWER. The latter looks for the follower iterator by
resolving the like-named call with tag and followThis actuals
added. The CallExpr is inside the _toFollower function, where the
follower iterator is not visible. So the resolution used to fail.

Cf. prior to #9994, the existing overload of chpl__initCopy(ir: _iteratorRecord)
would have been used, which iterates directly over `ir`:

    for elt in ir {...}

This does not require resolving any calls - the iterator's FnSymbol*
is found directly from the type of `ir`.

Here is the solution for the forall case that is implemented in this PR.

When we are looking for the follower iterator, we know where the
defPoint of the serial iterator is. Because we get to the serial
iterator easily given the type of `ir`.

Given how we generate forall expressions, the follower iterator's
defPoint is adjacent to the serial iterator's. They both are within
the same BlockStmt that is the body of the forall-expression function.

To find the follower, we traverse that BlockStmt looking for an iterator
with the matching name and a "followThis" formal.

Potential alternative solutions:

* When resolving the ForallStmt, we could have the compiler look for
the follower iterator directly instead of calling _toFollower.
For that, it would place the above CallExpr (that used to fail to resolve)
next to the ForallStmt. Maybe the follower iterator will be visible there.

* Investigate how the follower was found prior to #10459 
and see if we can learn from that.

* Instead of traversing the BlockStmt, we could create a CallExpr
as above and place it in that BlockStmt and resolve.
It should find the same iterator less heuristically, however with more work.

* Long-term it would be nice to figure out a way to avoid nesting the
iterator stuff inside the loop-expr wrapper function.

Testing: linux64 --verify; gasnet multilocale tests.

Resolves #10481 .


vasslitvinov:
@benharsh would you review?

PR #10549, opened 2018-07-26
Treat _defaultOf similarly to _new, deinit, etc for qualified access 

### LINKED ISSUE #10547: "use only" prevents visibility through _defaultOf
Related #9983 
Related #7857

### Steps to Reproduce
When using `use A only`, this error is emitted:
```
In function 'main':
error: unresolved call '_defaultOf(type R(int(64)))'
```

If `only` is omitted, the program compiles.

**Source Code:**

```chapel
module A {
  record R {
    type t;
  }
}

proc main() {
  use A only;
  var r: A.R(int);
}
```

### Configuration Information
`chpl version 1.18.0 pre-release (d2735623e9)`

lydia-duncan:
Resolves #10547 

I missed including _defaultOf in the list of functions that should be permitted
when a `use only;` or `use except *;` list allows qualified access.  Rectify
that.

Add a test to lock the fixed behavior in, using the contents of #10547


mppf:
I'm surprised that _defaultOf is even invoked for user-defined records.

lydia-duncan:
I believe it still applies to generic records, just not concrete ones

PR #10554, opened 2018-07-27
Update LCALS tests to compile with --warn-unstable

mppf:
Updates LCALS to use unmanaged to pass with --warn-unstable (except for an
enum-related error). This is a first step -- in the future it would be better
to use `borrowed` `owned` and `shared`.

- [x] full local testing


mppf:
@daviditen - I'm looking for feedback on these updates to LCALS.

I basically just used `unmanaged` everywhere for two reasons:
 1. I didn't want to dig into LCALS too much (since I have a lot of other tests to update).
 2. I didn't want to introduce performance differences, and `unmanaged` should behave the same as the pre-delete-free code would.

However I am open to suggestions for which should be `borrowed` or `owned` or `shared` etc.

Thanks.

PR #10562, opened 2018-07-27
Add deprecation warning for constructors

### LINKED ISSUE #10457: Add deprecation warning for user constructors
Given a program like the following:

```chpl
/*  1 */ class C {
/*  2 */   var x : int;
/*  3 */
/*  4 */   proc C(x : int) {
/*  5 */     this.x = x;
/*  6 */   }
/*  7 */ }
/*  8 */
/*  9 */ proc C.C(s : real) {
/* 10 */   this.x = s:int;
/* 11 */ }
/* 12 */
/* 13 */ var c = new C(1.5);
```

We want the compiler to emit a deprecation warning on lines 4 and 9.

### Initial proposal
Compiler emits the following for each occurrence of a constructor:
``Type '%s' defines a constructor here``

For the first warning only, the compiler will print the following note:
``Constructors have been deprecated as of Chapel 1.18. Please use initializers instead.``

For the program above, the total compiler output would be:
```
foo.chpl:4: note: Constructors have been deprecated as of Chapel 1.18. Please use initializers instead.
foo.chpl:4: warning: Type 'C' defines a constructor here
foo.chpl:9: warning: Type 'C' defines a constructor here
```

benharsh:
This PR adds a deprecation warning for constructors, with the text having been discussed in #10457.

Resolves #10457.

Testing:
- [x] local + futures
- [x] gasnet + futures


bradcray:
Yahoo!

mppf:
Never mind about --warn-special. Still we might consider an on-by-default languages changes warning control flag, or just leave --no-warnings to turn them off.

lydia-duncan:
> I think that the man page needs some help

@benharsh - if I don't hear from you by EOD Thursday, I'm going to take this on so it gets in

benharsh:
If I'm reading this right, there are two potential tasks here:

1) Update flag description (e.g. "Enable [disable] deprecation warnings for constructors")

2) Tie flag to some other meta-flag.

Is that right?

mppf:
@benharsh - mainly I'm hoping for an improved description and man page entry. Tying it to another flag I think probably should wait until after the release, if we do it.

PR #10565, opened 2018-07-27
Use borrowed in Mason classes

mppf:
To pass with --warn-unstable

- [ ] full local testing


Spartee:
LGTM - Thank you for doing this!

PR #10567, opened 2018-07-27
Allow .borrow() on unmanaged or borrowed

### LINKED ISSUE #9501: Should unmanaged and borrowed types have a .borrow() method?
It occurs to me that `shared` and `owned` are nicely symmetric in that you can call the `.borrow()` method to get a borrow out of them. Since borrowing from an unmanaged pointer is also possible (and in fact happens automatically in many cases with coercions) should unmanaged pointers also support a `.borrow()` method to make the operation clearer?

Pros:
 * Allows code borrowing from unmanaged to be written more clearly (I think myVariable.borrow() is clearer than myVariable:MyClass)
 * Allows code borrowing from unmanaged to be written without figuring out or naming the borrow type (e.g. myVariable:MyClass vs myVariable.borrow() ) which can make it easier to write code that is generic for many class types
 * Enables generic code to work with unmanaged and borrowed just as well as it works with owned or shared

Cons
 * Would imply that one cannot make a method on a class type called `.borrow()` since this is reserved for the borrowing operation

mppf:
Implement .borrow() for borrowed/unmanaged classes

This PR adds buildClassBorrowMethod() to run during
buildDefaultFunctions.  In some ways it would be preferable to create a
generic method with `borrowed` as the receiver; however making that work
was not simple/easy.  The new .borrow() method is built for class types.
That has the effect of enabling `.borrow()` on unmanaged because
unmanaged can coerce to owned.  `owned` and `shared` still have their own
implementations of `.borrow()` that should be preferred, but the coercion
to borrow & then call this `.borrow()` would have the same result.

Resolves #9501 (which includes design discussion).

- [x] full local testing

### DIFF
--- a/compiler/passes/buildDefaultFunctions.cpp
+++ b/compiler/passes/buildDefaultFunctions.cpp
@@ -1304,7 +1304,7 @@ static void build_union_assignment_function(AggregateType* ct) {
 }
 
 static void buildClassBorrowMethod(AggregateType* ct) {
-  if (function_exists("borrow", ct, ct))
+  if (function_exists("borrow", dtMethodToken, ct))
     return;
 
   FnSymbol* fn = new FnSymbol("borrow");
--- a/compiler/passes/buildDefaultFunctions.cpp
+++ b/compiler/passes/buildDefaultFunctions.cpp
@@ -40,6 +40,7 @@ static void build_accessors(AggregateType* ct, Symbol* field);
 static void buildDefaultOfFunction(AggregateType* ct);
 
 static void build_union_assignment_function(AggregateType* ct);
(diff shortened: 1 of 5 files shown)


compiler/passes/buildDefaultFunctions.cpp
```diff
@@ -1298,6 +1303,35 @@ static void build_union_assignment_function(AggregateType* ct) {
   normalize(fn);
 }
 
+static void buildClassBorrowMethod(AggregateType* ct) {
+  if (function_exists("borrow", ct, ct))
```
vasslitvinov: The first argument should be the method table, right?
If so, should be dtMethodToken.

PR #10568, opened 2018-07-27
Add warning on deleting a borrow

### LINKED ISSUE #9092: delete-free subtask: error on delete of a borrow
As a Chapel user, I want the compiler to raise an error if I delete a borrowed class instance, as this would probably lead to a double free.

Acceptance Criteria: Test demonstrates an error when a borrow is deleted.

mppf:
Adds a warning if what the compiler considers to be a borrow (including an undecorated class type) is deleted. Adjusts testing so that it will pass.
 * in internal modules, used a cast from the borrow type to unmanaged before deleting
 * in spec tests, just added the warning to the expected output. Updating the spec for delete-free has its own issue: #10631

Resolves  #9092.

- [x] full local testing
- [x] futures testing
- [x] gasnet testing

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -265,7 +265,7 @@ module ChapelArray {
     proc _freePrivatizedClassHelp(pid, original) {
       var prv = chpl_getPrivatizedCopy(object, pid);
       if prv != original then
-        delete prv;
+        delete _to_unmanaged(prv);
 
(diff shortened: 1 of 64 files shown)


mppf:
Passed except for
[Error matching program output for classes/forwarding/forwarding-generic-default]


!

mppf:
@vasslitvinov I've added some testing changes to pass futures and gasnet+futures testing. I plan to merge this next Thursday.

PR #10569, opened 2018-07-28
Minor cleanup+reorg of ForallStmt-related code

vasslitvinov:
* Importantly, move implementForallIntentsNew() to go AFTER
resolveParallelIteratorAndIdxVar(). This sets me up for
merging implementForallIntents1New() and setupShadowVariables(),
which do different aspects of the same task.

* Get rid of implementForallIntentsNew(), instead call
implementForallIntents1New() (with the `1`) and checkForNonIterator()
directly.

* Note that placing checkForNonIterator() requires some care.
(This was the case before as well.) It needs to happen after
the iterator is resolved, and before its yield time is queried.

* Remove the unused field reduceGlobalOp.

* list_ast() now produces prettier output for ForallStmts.

* Move ForallStmt- and UseStmt-specific code from list_ast()
into helper functions.

* Relocate print_indent() and print_on_its_own_line() towards
the top of view.cpp as they are used by other helpers and
have more general applicability.

### DIFF
--- a/compiler/AST/foralls.cpp
+++ b/compiler/AST/foralls.cpp
@@ -372,7 +372,7 @@ bool astUnderFI(const Expr* ast, ForallIntents* fi) {
 //
 //  * find the target parallel iterator (standalone or leader) and resolve it
 //  * issue an error, if neither is found
-//  * handle forall intents, using implementForallIntentsNew()
+//  * handle forall intents, using implementForallIntents1New()
 //  * partly lower by building leader+follow loop(s) as needed
 //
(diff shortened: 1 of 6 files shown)


bradcray:
This seems like a pretty large delta in terms of amount of code changed.  What makes it not review-worthy?  (if for no other reason than to increase the number of people on the team who understand this code).

vasslitvinov:
This is just moving the code around.

I have another cleanup change in the words and will have it reviewed together with this one.

PR #10574, opened 2018-07-30
Unmanaged deletes

mppf:
Fix tests identified by PR #10568 that should be using unmanaged.

- [x] full local testing

### DIFF
--- a/test/exercises/boundedBuffer/solutions/boundedBuffer-atomic.chpl
+++ b/test/exercises/boundedBuffer/solutions/boundedBuffer-atomic.chpl
@@ -26,7 +26,7 @@ config const numProducers = 5,     // the number of producers to create
 
 proc main() {
   // a shared bounded buffer with the requested capacity
-  var buffer = new BoundedBuffer(capacity=bufferSize);
+  var buffer = new unmanaged BoundedBuffer(capacity=bufferSize);
 
   // per-producer/consumer counts of the number of items they processed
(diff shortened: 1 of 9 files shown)


test/exercises/boundedBuffer/solutions/boundedBuffer-atomic.chpl
```diff
@@ -26,7 +26,7 @@ config const numProducers = 5,     // the number of producers to create
 
 proc main() {
   // a shared bounded buffer with the requested capacity
-  var buffer = new BoundedBuffer(capacity=bufferSize);
+  var buffer = new unmanaged BoundedBuffer(capacity=bufferSize);
```
ronawho: I imagine longer term the bounded buffer exercises should use managed classes since they're used in tutorials (at least I think they are)

ronawho: It also seems like the main boundedBuffer.chpl is getting out of sync with the solutions ones -- e.g. this is using an `unmanaged RandomStream` and boundedBuffer.chpl is using an `owned RandomStream`

mppf: Actually AFAIK they're not currently used in tutorials. I'd expect we'd have another look before doing such a tutorial. (But other people might run tutorials off of them...)

Issue #10543, opened 2018-07-26
Force GASNet to build the specified substrate


ronawho:
With the exception of aries/gemini, we let GASNet autodetect available conduits/substrates. This means that if a user explicitly set `CHPL_COMM_SUBSTRATE=mpi` and mpi support wasn't detected the build will still succeed but won't actually work for that configuration. We should configure gasnet with `--enable-mpi` for that case, which would result in a gasnet build failure if mpi support wasn't detected.

i.e. we should throw `--enable-<substrate>` to gasnet for all supported substrates. We might just be able to do `CHPL_GASNET_CFG_OPTIONS += --enable-$(CHPL_MAKE_COMM_SUBSTRATE)`, but I need to verify that there's a 1:1 mapping between our names and the GASNet substrates (and that there's an `--enable` option for each substrate.) If not, we just need to special case for each substrate (and there's not a ton so that wouldn't be too bad)

PR #10576, opened 2018-07-30
Default to ibv on cray-cs systems

ronawho:
This just corrects our default to match what we document. We used to default to
ibv on cray-cs systems, but this was lost when we rewrote the chplenv scripts
in python in 52aacf2485

Noticed while working on https://github.com/chapel-lang/chapel/issues/10543


ronawho:
https://chapel-lang.org/docs/usingchapel/multilocale.html?highlight=ibv#setting-chpl-comm-substrate and the cray platform doc mentions ibv as the default

PR #10579, opened 2018-07-30
Update documentation to note that the mxm substrate is deprecated

ronawho:
GASNet 1.32.0 deprecated mxm in favor of the ibv conduit. From the changelog:

```
* Mellanox ConnectX series HCAs (mxm-conduit)
  - This conduit is now DEPRECATED.
  - Use of ibv-conduit is recommended for users with InfiniBand hardware.
  - Adjust envvar GASNET_PHYSMEM_{MAX,PROBE} handling to match other conduits
```

Noticed while working on https://github.com/chapel-lang/chapel/issues/10543


gbtitus:
Do we need to print a message somewhere, or will GASNet do that?

Issue #10539, opened 2018-07-26
gasnet mpi substrate: fatal error:chpl-mem.h: No such file or directory


aheye:
### Summary of Problem
When building chapel from source on a Cray CS system with a gasnet substrate of mpi a fatal error is thown claiming that chpl-mem.h cannot be found. 

### Steps to Reproduce

```bash
git clone https://github.com/chapel-lang/chapel.git
cd chapel
export CHPL_COMM=gasnet
export CHPL_COMM_SUBSTRATE=mpi
make
```

**Error:**
```
/home/users/aheye/git/chapel/third-party/qthread/qthread-src/src/alloc/chapel.c:18:22: fatal error:chpl-mem.h: No such file or directory
 #include "chpl-mem.h"
                      ^
compilation terminated.
make[7]: *** [alloc/chapel.lo] Error 1
make[6]: *** [all-recursive] Error 1
make[5]: *** [all-recursive] Error 1
make[4]: *** [qthread-build] Error 2
make[3]: *** [/home/users/aheye/git/chapel/third-party/qthread/install/linux64-gnu-unknown-flat-jemalloc-hwloc] Error 2
make[2]: *** [third-party-pkgs] Error 2
make[1]: *** [runtime] Error 2
make: *** [comprt] Error 2
```

### Configuration Information

- Output of `chpl --version`: chpl version 1.18.0 pre-release (d273562)
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
```
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: unknown
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: ibv *
  CHPL_GASNET_SEGMENT: large
CHPL_TASKS: qthreads
CHPL_LAUNCHER: gasnetrun_ibv
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none
```
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16)
[...]

PR #10582, opened 2018-07-30
Force GASNet to build the specified substrate

ronawho:
Previously, we just let GASNet auto-detect the substrate/conduit. This meant
that a user could explicitly build with CHPL_COMM_SUBSTRATE=mpi and even if mpi
wasn't available on the system we wouldn't error, which led to much more
cryptic build or even runtime issues later on.

Now we throw `--enable-$CHPL_COMM_SUBSTRATE` to the GASNet build and if GASNet
can't build the specified substrate it will stop with a clear error message.

Closes https://github.com/chapel-lang/chapel/issues/10543

### DIFF
--- a/third-party/gasnet/Makefile
+++ b/third-party/gasnet/Makefile
@@ -10,6 +10,10 @@ ifneq ($(CHPL_MAKE_COMM_SEGMENT),none)
 CHPL_GASNET_CFG_OPTIONS += --enable-segment-$(CHPL_MAKE_COMM_SEGMENT)
 endif
 
+ifneq ($(CHPL_MAKE_COMM_SUBSTRATE),none)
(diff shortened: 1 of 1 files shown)


ronawho:
I verified that GASnet has an `--enable-<conduit>` configure option for all of our substrates. As a sanity check I also built the following configs:

cray-cs:
 - ibv
 - mpi (with openmpi/gcc loaded)
 - udp

my mac:
 - udp
 - smp
 - mpi (with mpich brew installed)
 - ofi (with libfabric brew installed)

cray-xc
 - aries
 - mpi

I also verified that the cray-cs mpi build failed if the openmpi module wasn't loaded (original issue in https://github.com/chapel-lang/chapel/issues/10539) and I sanity checked that a few other configuration that don't make sense also failed with a clear error message.

PR #10584, opened 2018-07-30
Remove marenostrum documentation, launcher, and chplenv defaults

ronawho:
marenostrum is a dramatically different machine then when we last ran on it
many years ago. Our documentation, launcher, and chplenv defaults are severely
out of date, so just remove them.


bradcray:
Speedy service!!  (You must've badly wanted to pull that already... :) )

PR #10585, opened 2018-07-30
Avoid using spool files for PBSPro

ronawho:
We redirect stdout/stderr directly to the output file for moab/torque systems,
but we stuck with using spool files on PBSPro systems because of slowness
encountered with redirection in the past.

Now we're running into issues on newer PBSPro systems because of spooling, so
try out redirection again. At least in limited testing I'm not seeing any
issues on the older XE PBSPro system that gave us trouble before.

For history see https://github.com/chapel-lang/chapel/pull/4084


awallace-cray:
Makes sense to me.

Issue #10493, opened 2018-07-24
Clean up Atomics.chpl implementation


ronawho:
Currently we have distinct types for `atomic_uint8`, `atomic_uint16`,
`atomic_uint32`, `atomic_uint64`, `atomic_int8`, `atomic_int16`,
`atomic_int32`, and `atomic_int64` where the implementation is identical except
for the name of the C function we call.  Now that we have param string extern
proc renaming (added in #10298) we can just create a single generic type and
compute the extern proc names.

We should be able to clean up `atomic_real32`, and `atomic_real64` too,
possibly folding it into the generic integral and using `compilerError()` for
operations that aren't supported (`or`, `and`, `xor`)

This will make upcoming changes to the Atomic implementation significantly
easier.

As part of this we'll probably need to add some additional and more
comprehensive testing of the atomics API.

Issue #10001, opened 2018-06-25
Spike: Investigate any backwards breaking changes for Atomics


ronawho:
There are some known issues and TODOs for our atomics implementation/API. Specifically: 
 - Our API for compareExchange does not match C/C++. Maybe we should update our API to pass the expected value by reference and have it be updated on failure. 
 - We currently use the C names for the memory order. We should make a Chapel enum that has the names we want, and map them down to the runtime where appropriate (#9763)

We should investigate if there's any other backwards breaking changes we'd like to make. One that comes to mind is whether we should make our memory orders params or keep them as they are.

LouisJenkinsCS:
> We should update our API to pass the expected value by reference and have it be updated on failure.

I hope that this bars the cases where the expected value ends up being passed by wide references. When we're dealing with network atomics here, I'd hate for `x.compareExchange(y,z)` to turn into a GET. Speaking of which, can you even do this with network atomics? If it comes down to getting good performance and compatibility, I'd choose to have an overloaded version of the same method, or one that has a different name, like `compareAndSwap`.

dmk42:
There used to be a Chapel compiler bug that prevented our making the `compareExchange()` argument a ref.  I haven't gotten that far yet, but I'm adding this note as a reminder to myself to check and see if that still exists.

dmk42:
Experiment done.  The Chapel compiler bug I encountered in the past no longer exists.  I have a branch that changes the expected argument to a ref, and it passes the tests.

@gbtitus - Can you address the comment from @LouisJenkinsCS above?  When the NIC does a compareExchange operation that fails, do you get back the previous value without an extra GET?

@ronawho - After I check over my branch to take the answer from @gbtitus into account, we could just make it into a PR and take care of that part of the issue.

I think it would be a good idea to make memory order a param.  A similar experiment should be done with that.

ronawho:
FWIW I have an upcoming PR that will dramatically simplify the Atomics.chpl and NetworkAtomics.chpl modules. It uses the new "param string extern proc renaming" capability added https://github.com/chapel-lang/chapel/issues/10053 / https://github.com/chapel-lang/chapel/pull/10298 to get rid of all the code duplication that resulted from needing unique names for the C routines. We basically end up with a single general atomic_T Chapel record that computes the extern names.

[...]

Issue #9368, opened 2018-04-26
Convert my internal list of TODOs into issues


ronawho:
Here's some TODOs that I've been putting converting into issues. Hopefully opening this issue will be a forcing function to actually get around to that.

Here's most of them:

---
Misc: 
 - [ ] create a simple "no-op" launcher (a hook so we can set env vars)
 - [x] remove block report support (#10127)
 - [ ] operator overloads for atomic ops
 - [ ] stabilize order of  argument bundles
 - [x] possible ra scalability regression (https://github.com/chapel-lang/chapel/issues/9435)
 - [x] verify how much memory is being used for task stacks
 - [x] improve genGraphs speed (#9430)

---
Third-party:
 - [ ] gasnet collab meta-issue
 - [ ] upgrade to jemalloc 5
 - [ ] upgrade to hwloc 2
 - [ ] qthread arm assembly
 - [ ] qthread hwloc 2
 - [ ] qthread faster syncs
 - [ ] fix build with tasks=qthreads hwloc=none

---
Performance:
 - [x] make llvm the default (#9902)
 - [ ] add test for sync vs atomic mutex performance
 - [x] add a mutex/lock library (#9881)
 - [x] make sync barrier reusable (#9734)
 - [x] avoid serialization of sync vars (#9651, #9737)
 - [ ] user-selectable between-locale barrier
 - [ ] corespec
 - [x] avoid dynamic array registration for serial init (#9616)
 - [x] bulk/vector spawn (#9613/#9614)
 - [x] qthreads+OpenMP interference (#9882)
 - [ ] better local barrier implementations
 - [ ] comm/compute overlap?
 - [x] task resetting  (#9618)
 - [ ] LICM improvements
 - [ ] can't inline zippered multi-dimensional iterators
 - [ ] 2D promoted operation performs poorly

---
Benchmark Elegance:
 - [ ] remove difference between ISx release and hand-optimized
 - [ ] reduce diff between elegant and opt (requires task private vars)

---
Benchmarks Perf Optimizations:
 - [x] update ml compopts/execopts to use "scalable numbers" (#9626)
 - [x] LCALS: (#9623)
   - [x] task spawning speed (#9614)
 - [x] Stream: (#9617)
   - [x] arrays on hugepages hurts perf (#7852)
   - [x] scalability for global (#9613)
 - [x] RA: (#9625)
   - [x] add oversub variants (included in RA epic)
 - [x] PRK-stencil: (#9620)
[...]

PR #10588, opened 2018-07-30
Significantly simplify and clean up Atomics.chpl

ronawho:
This cleans up the processor atomic implementation (network atomic
implementation will be cleaned up in another PR)

Previously we had distinct types for `atomic_uint8`, `atomic_uint16`,
`atomic_uint32`, `atomic_uint64`, `atomic_int8`, `atomic_int16`,
`atomic_int32`, `atomic_int64`, `atomic_real32`,`atomic_real64`. The only
reason we had so many types instead of a generic type is because we had to call
to C functions where the names were uniquified by the base type

With #10298, extern procs can now use a computed name so we can just have a
generic atomic type and compute the extern name based on the basetype. We now
just have an  AtomicT type that supports all sizes of real and integer types
and an AtomicBool type.

Nothing has really changed from the user perspective, but this makes upcoming
changes to the Atomics significantly easier.

Closes https://github.com/chapel-lang/chapel/issues/10493


ronawho:
The whole diff is pretty useless. This is easiest to review commit at a time and/or by looking just at the new file.

ronawho:
And no major compilation hit from this. 

On /tmp of chapcs07 with an optimized compiler I did not see any real changes for `time chpl test/runtime/configMatters/comm/atomics-api.chpl` (which is a test that uses the full atomic 
API for every supported type)

| config | time  |
| ------ | ----- |
| master | 8.15s |
| branch | 8.18s |

modules/internal/Atomics.chpl
```diff
@@ -278,148 +101,205 @@ module Atomics {
...
+      when int(64)  do return atomic_int_least64_t;
+
+      when uint(8)  do return atomic_uint_least8_t;
+      when uint(16) do return atomic_uint_least16_t;
+      when uint(32) do return atomic_uint_least32_t;
+      when uint(64) do return atomic_uint_least64_t;
+
+      when real(32) do return atomic__real32;
+      when real(64) do return atomic__real64;
+    }
```
gbtitus: For sure no need for an `otherwise` here, to guard against future mis-coding?

mppf: Pretty sure you'd get a compilation error in that case because the function wouldn't return anything. The `select T` has to be evaluated at compile-time since this is a type function.

modules/internal/Atomics.chpl
```diff
@@ -278,148 +101,205 @@ module Atomics {
...
+  }
+
+  private proc externTString(type T) param {
+    if T == bool then return "bool";
+    if isInt(T)  then return "int_least" + numBits(T):string + "_t";
+    if isUint(T) then return "uint_least" + numBits(T):string + "_t";
+    if isReal(T) then return "_real" + numBits(T);
```
mppf: why do the integer versions cast `numBits(T)` to string but the real version does not? I would have thought the addition would be left-associative, and so the cast wouldn't be necessary in the int case.

modules/internal/Atomics.chpl
```diff
@@ -278,148 +101,205 @@ module Atomics {
...
+      on this do ret = atomic_load(_v, order);
       return ret;
     }
 
     /*
        Stores `value` as the new value.
     */
-    inline proc write(value:bool, order:memory_order = memory_order_seq_cst) {
-      on this do atomic_store_explicit_bool(_v, value, order);
+    inline proc write(value:bool, order:memory_order = memory_order_seq_cst): void {
```
mppf: We've talked about making these memory_order arguments param. Do you think that's a good idea? Do you know if we have an issue about it on the Backlog somewhere?

modules/internal/Atomics.chpl
```diff
@@ -430,1359 +310,319 @@ module Atomics {
...
-    compilerError("Cannot directly assign atomic variables");
-  }
-  inline proc =(ref a:atomic_real32, const ref b:atomic_real32) {
+  // types are records and unless explicitly defined, it will resolve
+  // to the normal record version of the function.  Sigh.
+  inline proc =(ref a:AtomicBool, const ref b:AtomicBool) {
     a.write(b.read());
   }
-  inline proc =(ref a:atomic_real32, b) {
+  inline proc =(ref a:AtomicBool, b) {
```
mppf: I don't understand why we don't support this. Isn't it "obvious" that `myAtomic = 1` will be the same as myAtomic.write(1) ? It seems to me it would be easy to make this `=` overload do that. Might even allow initialization.

modules/internal/Atomics.chpl
ronawho: Right, it will return void -- and these functions are all indirectly guarded by `chpl__processorAtomicType()`, which issues a compileError for unsupported types

modules/internal/Atomics.chpl
ronawho: Whoops, just an omission  -- this explicit cast isn't needed but I found it to be a little bit cleaner

modules/internal/Atomics.chpl
ronawho: Yeah, part of https://github.com/chapel-lang/chapel/issues/10001

modules/internal/Atomics.chpl
gbtitus: Isn't that moot, since it already doesn't return anything if the type doesn't match any of the `when` options?  I was thinking of putting a `compilerError()` in the `otherwise`, the moral equivalent of what is done at https://github.com/chapel-lang/chapel/pull/10588/files#diff-b8fb835d4f7a58ecdc2589f832485f7aR161.  We'll get some kind of compiler error either way, but a `compilerError()` in an `otherwise` allows that to be a more specific one.

modules/internal/Atomics.chpl
ronawho: Supporting atomic initialization and operator overloads are captured elsewhere -- https://github.com/chapel-lang/chapel/issues/5037 and https://github.com/chapel-lang/chapel/issues/8847 (and is also in #9368 as "operator overloads for atomic ops".)

modules/internal/Atomics.chpl
ronawho: Right, we could add compilerErrors to `externT()`, `externTString()`, and a few other routines but they're already guarded by the compilerError in `chpl__processorAtomicType()` and I didn't see the need to duplicate the error.

modules/internal/Atomics.chpl
ronawho: Updated to add the cast

PR #10589, opened 2018-07-30
Create a directory for the generated library files (and allow it to be renamed)

### LINKED ISSUE #10527: Create directory for generated files with library compilation
Once we start generating the files for Python interoperability, the number of files a `--library` compilation can generate will get quickly out of hand.  It would be nice to automatically create a directory where these generated files will live, with the default behavior being a new `lib` directory.

We probably still want to clean up the Cython files used to generate the Python module by default.

lydia-duncan:
Resolves #10527

In discussion, we decided that it would be helpful to put all the generated
files into a single directory, including the generated library file, header file,
and makefile (and later files that we will generate).  This will reduce
unnecessary clutter in the original directory.  The default name for the
generated directory is `lib` but it can be changed via a new flag,
`--library-dir`.

Various adjustments were made so that everything would get put in the
right place and appropriately referenced by test files.  We no longer put
the header file in the temporary directory, instead placing it and the makefile
directly in the generated library directory.

I updated a lot of test files for this change.  The most notable is
test/compflags/library/withMain, where the lib directory name was interfering
with compiling the "lib.chpl" file.  Here, I decided to utilize the new flag to
rename the directory so that they would no longer conflict.

Also updated the autocomplete file.

TODO:
- [ ] update AI workflow for this change


mppf:
I havn't dug in yet, but have a quick question. Does this change the behavior of --library? If so, we'll need to update the technote that documents it.

I would have supposed that the default directory name would be the library name (e.g. `foo` for `libfoo.a`). That just seems more natural to me but I don't have a strong opinion about it. Can you say more why you chose `lib` as the default?

compiler/codegen/library.cpp
```diff
@@ -162,35 +160,31 @@ const char* getLibraryExtension() {
   return "";
 }
 
-void openLibraryHelperFile(fileinfo* fi, const char* name, const char* ext,
-                           bool useTmpDir) {
+void ensureLibDirExists() {
+  if (libDir[0] == '\0') {
+    strncpy(libDir, "lib", sizeof("lib"));
+    libDir[sizeof("lib")] = '\0';
```
mppf: strncpy will add the terminating null - in fact it will initialize all of the memory with nulls.

compiler/main/driver.cpp
```diff
@@ -985,6 +994,7 @@ static ArgumentDescription arg_desc[] = {
  {"ignore-user-errors", ' ', NULL, "[Don't] attempt to ignore user errors", "N", &ignore_user_errors, "CHPL_IGNORE_USER_ERRORS", NULL},
  {"ignore-errors-for-pass", ' ', NULL, "[Don't] attempt to ignore errors until the end of the pass in which they occur", "N", &ignore_errors_for_pass, "CHPL_IGNORE_ERRORS_FOR_PASS", NULL},
  {"library", ' ', NULL, "Generate a Chapel library file", "F", &fLibraryCompile, NULL, NULL},
+ {"library-dir", ' ', "<directory>", "Save generated library helper files in directory", "P", libDir, "CHPL_LIB_SAVE_DIR", verifySaveLibDir},
```
mppf: should --library-dir imply --library?

test/compflags/lydia/library/.gitignore
```diff
@@ -1 +1,3 @@
-testing.h
+lib/testing.h
+lib/hyphenated-name.h
+lib/libtesting.a
```
mppf: that pesky missing newline

lydia-duncan:
> Does this change the behavior of --library? If so, we'll need to update the technote that documents it.

It does.  I will do that!

> I would have supposed that the default directory name would be the library name (e.g. foo for libfoo.a). That just seems more natural to me but I don't have a strong opinion about it. Can you say more why you chose lib as the default?

When I think of "a directory where a library lives", I think of a `lib` directory - this seemed in keeping with that, even though there are additional files included there.  If the user intends to add this directory to their LD_LIBRARY_PATH it doesn't particularly matter what the name is, though.  It was an arbitrary decision, so I'm happy to change it if someone feels strongly about it.

lydia-duncan:
> making --library-dir an optional way to do this

That's an interesting thought.  I'll think about that a little bit more, it's certainly intriguing

compiler/codegen/library.cpp
mppf: I was thinking of `strncpy` but just misreading what you were passing as the length field.

Since this is setting the default, don't we know it won't overflow libDir?

sizeof("lib") is 8 bytes because it's a pointer. Will work, but probably confused.

I think this code should be:
``` c
  if (libDir[0] == '\0') {
    const char* dir = "lib";
    INT_ASSERT(strlen(dir) < sizeof(libDir));
    strcpy(libDir, dir);
  }
```

mppf: Apparently sizeof("lib") is OK; see https://bytes.com/topic/c/answers/212229-sizeof-string-literal ; but I still think it's more reasonable to write the above.

lydia-duncan:
I added some updates to the libraries technote, care to take a look?

lydia-duncan:
I have included them on the mails :) I'll let it sit until tomorrow and let them know they have until then to respond

lydia-duncan:
Brad and Ben have both signed off on this

PR #10592, opened 2018-07-30
gitignore some files that had been sitting around for me

lydia-duncan:
Ignores a generated .good file, the compiler performance directory, and the
generated mason directories.


lydia-duncan:
@ronawho and @Spartee - do these seem like appropriate files/directories to have git ignore?

Spartee:
Looks good to me as well!

PR #10597, opened 2018-07-31
Reduce special-case pruning of shadow variables

vasslitvinov:
During resolution, the compiler converts references to outer variables
in a forall loop body to shadow variables.  The compiler used to
exclude the 'this' variable and variables of atomic types from this
treatment. This was done because the inclusion of such variables
caused issues elsewhere in the compiler that would need to be addressed.

Now the compiler is in a better shape and we do not need these exclusions.

With this change, the compiler detects previously-unnoticed invalid
access to a const field after this.complete() in an initializer.
To avoid switching this field to non-const, instead initialize that field
to a forall-expression prior to this.complete().

I examined manually the AST at the end of resolution to confirm
that a temporary array is not created for the forall-expression.

In the future we may also want to Chapel-ify hg_getCoeffs()
by having it return a forall-expression instead of a temporary array.
I did not do it because the other use of hg_getCoeffs()
would be expressed better as a rank-changing slice.

### DIFF
--- a/compiler/resolution/implementForallIntents.cpp
+++ b/compiler/resolution/implementForallIntents.cpp
@@ -1984,7 +1984,6 @@ static void findOuterVarsNew(ForallStmt* fs, SymbolMap& outer2shadow,
         !sym->hasFlag(FLAG_TEMP)     && // not a temp
         !isFsIndexVar(fs, sym)       && // not fs's index var
         !isFsShadowVar(fs, sym)      && // not fs's intent var
-        !sym->hasFlag(FLAG_ARG_THIS) && // todo: no special case for 'this'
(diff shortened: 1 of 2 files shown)


lydia-duncan:
What testing has been done for this change?

vasslitvinov:
Testing: passes linux64 --verify and gasnet multilocale tests.

lydia-duncan:
I thought you were going to be running linux64 and linux64 --verify, since that's caused issues in the past?

vasslitvinov:
Indeed, it has caused issues in the past. However that was only for futures, when there was a compiler failure. For this PR I did not test futures.

PR #10600, opened 2018-07-31
Retire submitted chameneos entries

benharsh:


benharsh:
@bradcray : we recently discussed NOTEST-ing these tests. Should we use this as an opportunity to do so?

benharsh:
@bradcray : Just added some commits that NOTEST the submitted chameneos, could you review?

bradcray:
This looks fine to me, thanks.  I might have preserved the `.perf*` files as well to snapshot the configurations we were testing when the tests were retired, but would be fine with it going in either way.

Ultimately, we'll want to push the submitted meteor and thread-ring codes in here as well, but I understand that this PR is striving to address last night's perf testing failures.

bradcray:
(BTW, I reviewed just by inspecting the diff, not by grepping the source tree or running testing to see whether anything was missed...).

PR #10601, opened 2018-07-31
Added openssl@1.1 specific malloc/dealloc APIs

### LINKED ISSUE #9794: some crypto module tests don't pass on Ubuntu 18.04
library/packages/Crypto/saru/blowfish/bfTest1.chpl
library/packages/Crypto/saru/hashing/hash1.chpl

I get errors along the lines of:
```
In file included from /tmp/chpl-mppf-17728.deleteme/_main.c:46:0:
/tmp/chpl-mppf-17728.deleteme/Crypto.c: In function 'bfEncrypt_chpl':
/tmp/chpl-mppf-17728.deleteme/Crypto.c:615:16: error: storage size of 'ctx_chpl' isn't known
 EVP_CIPHER_CTX ctx_chpl;
                ^~~~~~~~

```

on Ubuntu 18.04 with GCC 7.3 and openssl 1.1.0g.

It seems that EVP_CIPHER_CTX is defined as an opaque type in ossl_type.h:
``` C
typedef struct evp_cipher_ctx_st EVP_CIPHER_CTX;
```

and this S.O. answer points out that "EVP_CIPHER_CTX was made opaque in OpenSSL 1.1.0" and discusses what to do about it:
https://stackoverflow.com/questions/26345175/correct-way-to-free-allocate-the-context-in-the-openssl

saru95:
[PR by @saru95 and @mppf]

This PR updates Crytpo and the CryptoHandlers .h file to work with both
OpenSSL 1.0.1g and 1.1.

OpenSSL 1.1 has some breaking API changes that make code that works with
1.0 unable to compile. The approach here is to add several tiny functions
to the .h file to handle the difference, based upon a C preprocessor
conditional checking the OpenSSL version.

These compatability types and functions are:
 * CHPL_EVP_MD_CTX
    * in 1.1, this is EVP_MD_CTX*; in 1.0 it is EVP_MD_CTX
 * CHPL_EVP_CIPHER_CTX
    * in 1.1, this is EVP_CIPHER_CTX*, in 1.0 it is EVP_CIPHER_CTX
 * CHPL_EVP_MD_CTX CHPL_EVP_MD_CTX_new, CHPL_EVP_MD_CTX_free
    * create and free a CHPL_EVP_MD_CTX
 * CHPL_EVP_CIPHER_CTX_new, CHPL_EVP_CIPHER_CTX_free
    * create and free a CHPL_EVP_MD_CTX
 * CHPL_EVP_MD_CTX_ptr, CHPL_EVP_CIPHER_CTX_ptr
    * get the ptr (used in other OpenSSL API calls) from the 
      Chapel-specific type
    * e.g. CHPL_EVP_MD_CTX_ptr takes in `CHPL_EVP_MD_CTX* arg`
      and returns an `EVP_MD_CTX*`.
    * in 1.1 it returns *arg (since CHPL_EVP_MD_CTX is already the
      pointer)
    * in 1.0 it returns arg (since CHPL_EVP_MD_CTX refers to a struct)

This PR leaves updating the RSA routines as future work. The RSA routines
aren't currently tested and @saru95 is working on improving them.

Resolves #9794 

Testing with 8ade5eb5c8515950a40de1b889f55e1bf4962db9:
 - [x] skipped on Mac OS X (with no pkg-config)
 - [x] skipped on Mac OS X (with only older OpenSSL)
 - [x] passes on Ubuntu 18.04
 - [x] passes on SLES 12 with libcrypto 1.0.2j


mppf:
I don't think the COMPOPTS files are necessary at all on Ubuntu and in fact it's not great to encode paths like `/usr/local/opt/openssl@1.1/` in the test. Can we detect this path using pkg-config?

Speaking of which, mason has a brand new shiny pkg-config integration. Maybe we can solve this problem by making Crypto a mason package? Or, if we don't want to do that yet, maybe we can ask `mason` to provide the right paths? @Spartee can you advise?

Spartee:
The crypto module could easily be turned into a mason package. If we do not want to do that then you could ask mason for the path information, but it would be much easier(instead of parsing toml) to just use `pkg-config` directly in this case.

Personally, I think Crypto is one of the current packages that should be made into a Mason package.

saru95:
@mppf Looking at pkg-config. The path is `/usr/local/include/openssl/opensslv.h` for macOS.

saru95:
@mppf Could you have a look?

mppf:
@saru95 - it works OK on Ubuntu 18 but on SLES 12 I get errors like this:
```
[Executing compiler /users/mferguson/w/2/bin/linux64/chpl -o buffer4 --cc-warnings `pkg-config --libs --cflags libcrypto` buffer4.chpl < /dev/null]
[Elapsed compilation time for "library/packages/Crypto/saru/utils/buffer4 (compopts: 1)" - 0.010 seconds]
[Executing diff buffer4.good buffer4.1.comp.out.tmp]
1c1
< 00 00 00 00 00 00 00 00 00 00 00
---
> Unrecognized flag: '-D' (use '-h' for help)
```

I think you mean to use chpl --ccflags `pkg-config --cflags libcrypto` --ldflags `pkg-config --libs libcrypto`. Adjusting the COMPOPTS in that manner allows it to work on SLES 12.

Can you make these COMPOPTS files be a symbolic link to one in test/library/packages/Crypto/saru/COMPOPTS, say? Seems like a pain if we have to change them all again.

saru95:
ping @mppf

mppf:
@saru95 - unfortunately not working on Ubuntu 18.04. Now I get

```
[test: library/packages/Crypto/saru/hashing/hash1.chpl]
[Executing compiler /home/mppf/w/15/bin/linux64/chpl -o hash1 --cc-warnings --ccflags `pkg-config --cflags libcrypto` --ldflags `pkg-config --libs libcrypto` hash1.chpl < /dev/null]
[Elapsed compilation time for "library/packages/Crypto/saru/hashing/hash1 (compopts: 1)" - 8.990 seconds]
[Executing diff hash1.good hash1.1.comp.out.tmp]
1,3c1,4
< The quick brown fox jumps over the lazy dog
< SHA1 = 2f d4 e1 c6 7a 2d 28 fc ed 84 9e e1 bb 76 e7 39 1b 93 eb 12
< SHA256 = d7 a8 fb b3 07 d7 80 94 69 ca 9a bc b0 08 2e 4f 8d 56 51 e4 6d 3c db 76 2d 02 d0 bf 37 c9 e5 92
---
> gcc: error: unrecognized command line option ~~~--ldflags~~~; did you mean ~~~--ltrans~~~?
> /home/mppf/w/15/runtime/etc/Makefile.exe:27: recipe for target '/tmp/chpl-mppf-5312.deleteme/hash1.tmp' failed
```

I believe that the issue here is that there were no arguments to --ccflags so the Chapel compiler interpreted the next argument (--ldflags) as the thing to pass to the C compiler.

How about we write a script that handles the empty result?:
```
test/library/packages/Crypto/saru/COMPOPTS 
```

``` bash
#!/bin/bash

PKG=libcrypto
CFLAGS=`pkg-config --cflags $PKG`
LFLAGS=`pkg-config --libs $PKG`

OPTS=""

if [ ! -z "$CFLAGS" ]
then
  OPTS="$OPTS --ccflags \"$CFLAGS\""
fi

if [ ! -z "$LFLAGS" ]
then
  OPTS="$OPTS --ldflags \"$LFLAGS\""
fi

echo $OPTS
```

I've verified that this COMPOPTS works as expected on Ubuntu 18.04 (and then the tests pass). Does it work for you?

saru95:
For some reason, scripts written within `COMPOPTS` are not getting executed on my macOS. Tried with Python as well,

```py
#!/usr/bin/env python

import subprocess

pkg = 'libcrypto'
cflags = subprocess.check_output('pkg-config --cflags %s' % pkg, shell=True).rstrip()
lflags = subprocess.check_output("pkg-config --libs %s" % pkg, shell=True).rstrip()
opts = ''

if cflags:
    opts += '--ccflags %s ' % cflags

if lflags:
    opts += '--ldflags %s ' % lflags

print(opts)
```

Figuring this out.

saru95:
The pushed Python script works for me. I was running into trouble with bash script only to realize that the file wasn't executable and by that time I had already written a python script. lol.

mppf:
@saru95 - looks good on Ubuntu but on SLES 12 now we see this error:

```
[Elapsed compilation time for "library/packages/Crypto/saru/hashing/hash1 (compopts: 1)" - 4.834 seconds]
[Executing diff hash1.good hash1.1.comp.out.tmp]
1,3c1,15
< The quick brown fox jumps over the lazy dog
< SHA1 = 2f d4 e1 c6 7a 2d 28 fc ed 84 9e e1 bb 76 e7 39 1b 93 eb 12
< SHA256 = d7 a8 fb b3 07 d7 80 94 69 ca 9a bc b0 08 2e 4f 8d 56 51 e4 6d 3c db 76 2d 02 d0 bf 37 c9 e5 92
---
> In file included from /tmp/chpl-mppf-33383.deleteme/_main.c:47:0:
> /tmp/chpl-mppf-33383.deleteme/Crypto.c: In function ~~~digestPrimitives_chpl~~~:
> /tmp/chpl-mppf-33383.deleteme/Crypto.c:595:17: error: implicit declaration of function ~~~EVP_MD_CTX_new~~~ [-Werror=implicit-function-declaration]
>  call_tmp_chpl = EVP_MD_CTX_new();
>                  ^~~~~~~~~~~~~~
> /tmp/chpl-mppf-33383.deleteme/Crypto.c:595:15: error: assignment makes pointer from integer without a cast [-Werror=int-conversion]
>  call_tmp_chpl = EVP_MD_CTX_new();
>                ^
> /tmp/chpl-mppf-33383.deleteme/Crypto.c:677:1: error: implicit declaration of function ~~~EVP_MD_CTX_free~~~ [-Werror=implicit-function-declaration]
>  EVP_MD_CTX_free(call_tmp_chpl);
>  ^~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> /users/mppf/w/2/runtime/etc/Makefile.exe:27: recipe for target '/tmp/chpl-mppf-33383.deleteme/hash1.tmp' failed
```

That system reports OpenSSL version 1.0.2j.

Adding this (executable!) skipif file test/library/packages/Crypto/saru/hashing/hash1.skipif
(which is based off of aes.skipif but requires 1.1) resolves the problem for me:
```
#!/usr/bin/env bash

reqMajor=1
reqMinor=1
reqPatch=0

version=`openssl version | sed -e 's/^OpenSSL \([^ ]*\) .*$/\1/'`
major=`echo $version | sed -e 's/\..*$//'`
minor=`echo $version | sed -e 's/.*\.\([^\.]*\)\..*$/\1/'`
patch=`echo $version | sed -e 's/.*\.//'`

# Skip if the OpenSSL version is too old
if [[ $major < $reqMajor ]] ; then
  echo True
elif [[ $minor < $reqMinor ]] ; then
  echo True
elif [[ $patch < $reqPatch ]] ; then
  echo True
else
  echo False
fi
```

You might know of other Crypto tests that should be updated to only run when OpenSSL 1.1 is available.

saru95:
This would be the case in almost all the tests since most of the ciphers use hashing internally. We also can't revert back just the hashing API to 1.0.2 since it's not present in 1.1.

Would it make sense to skip all the tests on a system with 1.0.2?

mppf:
@saru95:

* If you want to support both 1.0 and 1.1, I recommend doing that with #ifdefs in the .h file we already have
* I think it would be reasonable to use a directory skipif that checks for 1.1. This would amount to updating test/library/packages/Crypto/saru.skipif to be something like the script we have above. Additionally since we're using pkg-config, you'd want the skipif to check that pkg-config exists and it can find libcrypto. (In fact it could just do that, and use pkg-config --modversion to check the version).

mppf:
@saru95 - ping; AFAIK this PR just needs a directory-skipif script test/library/packages/Crypto/saru.skipif that checks that pkg-config is available and it can find libcrypto > 1.1.

PR #10602, opened 2018-07-31
Minor mods to mergeSort() with an eye toward performance

### LINKED ISSUE #10502: Sort.mergeSort is "in-place" but temporary O(n) arrays are created
### Summary of Problem
[Documentation for Sort.mergeSort](https://github.com/chapel-lang/chapel/blob/c1acda2bf90d7870ed7c750b73d48dc23f593f5b/modules/packages/Sort.chpl#L584-L593) indicates that the sorting happens in-place, but array-slice copies are made. https://github.com/chapel-lang/chapel/blob/c1acda2bf90d7870ed7c750b73d48dc23f593f5b/modules/packages/Sort.chpl#L609-L610

It should be a quick fix to change these to `ref` and remove the last-line reassignment.

### Configuration Information
`chpl version 1.18.0 pre-release (6e8996b7c9)`

bradcray:
As pointed out in issue #10502, our ancient proof-of-concept mergeSort()
implementation creates array copies of array slices.  Changing those
'var' declarations to be 'ref's instead ought to avoid this, as done here.
While here, I also looked for other instances of slicing and cleaned them
up.

With additional effort, mergeSort() could be written to avoid slicing altogether
by passing low/high bounds into the recursive function separately.

resolves #10502


ronawho:
Did you do performance testing on this? I actually see slightly worse performance with this running on chapcs10 with --M=26 (slightly larger than nightly --M=24):

```
# --M=26 master
08/03/18        1.14366
08/03/18        1.15834
08/03/18        1.12172

# --M=26 mergesort-cleanups
08/03/18        1.4302
08/03/18        1.43149
08/03/18        1.48396
```

bradcray:
Huh... Yeah, on my laptop I was seeing perf improvements.  I'll double check it before merging, though.

ronawho:
Weird, and just watching the output of `top` while this is running with `--M=30` I see more max memory being used (25% vs 20%) which is really confusing. Here's the --M=30 results: 

```
# --M=30 master
08/03/18        18.3698
08/03/18        16.7826
08/03/18        16.8446

# --M=30 mergesort-cleanups
08/03/18        23.6511
08/03/18        23.0715
08/03/18        22.7589
```

modules/packages/Sort.chpl
```diff
@@ -606,15 +606,15 @@ private proc _MergeSort(Data: [?Dom], minlen=16, comparator:?rec=defaultComparat
...
+  ref A1 = Data[lo..mid];
+  ref A2 = Data[mid+1..hi];
   cobegin {
     { _MergeSort(A1, minlen, comparator); }
     { _MergeSort(A2, minlen, comparator); }
   }
 
   // TODO -- This iterator causes unnecessary overhead - we can do without it
-  for (a, _a) in zip(Data[lo..hi], _MergeIterator(A1, A2, comparator=comparator)) do a = _a;
+  for (a, _a) in zip(Data, _MergeIterator(A1, A2, comparator=comparator)) do a = _a;
```
ronawho: Looking at the _MergeIterator iter -- it can't be zippered inlined, so it might be interesting to see what sort of performance we could get by manually inlining it. Or maybe making it non-zippered by doing something like: 

```chpl 
  var i = lo;
  for _a in _MergeIterator(A1, A2, comparator=comparator) {
    Data[i] = _a;
    i += 1;
  }
```

Probably more than you want to look into for now, but I couldn't understand why your change would hurt performance so I was digging around a little.

bradcray:
I no longer think this PR is correct (or at least, don't see why it would be).  See [this comment](https://github.com/chapel-lang/chapel/issues/10502#issuecomment-410405223) for an indication why. Closing it for now.

PR #9796, opened 2018-06-11
Fix several memory leaks

mppf:
* fixes RecordParser to avoid useless cast & optimize it
* add tests for useless string cast leak
* fixes leak with useless cast by removing special cases for initCopy
* while there, removed FLAG_DONOR_FN, which as apparently no longer required
* fixes json-enum leak (described in issue #9809) by moving the removeFlag 
  call to the initializers case
* added tests for initialization leak

Resolves #9809
Resolves #9689

An earlier version passed valgrind testing.
Leaks no more memory than master in test/release/examples/primers.
Passed valgrind testing of test/release/examples/primers.
Passed full local testing.

Reviewed by @benharsh - thanks!

PR #10611, opened 2018-07-31
Fix problem with throwing nested call var init

### LINKED ISSUE #10204: Segmentation Fault when attempting to write to file without file permissions
I noticed that when I attempted to open a file with bad permissions, such as one that is owned by root, it will cause a runtime crash/segmentation fault, a very confusing one at that. Would it be possible to add a runtime check to determine if opening a file has failed, and if so throw a runtime error that is a bit more helpful, such as `open: "Bad file permissions for file %FILENAME%"`.

mppf:
Resolves #10204.

This PR fixes a problem with code like this:
``` chapel
proc foo() throws {
  var r = makeMyRecord().returnThis();
}
```
in the event that makeMyRecord() throws.

The problem was that callDestructors considers any variable for which a DefExpr exists in the current control flow to be subject to deinitialization. But the normalizer was adding call-expr temps and nested calls between the DefExpr for a variable init and the initialization itself. This confused callDestructors.

The fix here is to identify this pattern in the normalizer and to add the call-expr temps before the DefExpr for the variable being initialized in that case. An alternative strategy could be to modify callDestructors to only consider a variable initialized after it is defined *and* it is possibly set.

The change to normalization caused problems with tests such as coercions-covariant.chpl, where a coercion combines with the `in` intent. Adjusted callDestructors to detect the "move" in this case by the FLAG_NO_AUTO_DESTROY and also to remove a destruction when the copy is removed.

This change to normalization caused issues with the previous handling of removing useless casts (added in #9796). To solve that problem, this PR introduces a new function `_removed_cast`. Removed record casts are converted into calls to this function so that the copy behavior is preserved. Along these lines, adjusted two module functions to avoid useless casts (since the code can be simpler without it).

Lastly, while debugging an earlier version of these changes, I noticed that a `shared` `_cast` function was changing reference counts twice (because the default intent for shared is `in`). Fixed this by explicitly marking the default intent (to be clearer to humans) and then stealing the already-incremented reference count from the argument.

- [x] full local testing
- [x] primers leak no more after this PR than they do on master
- [x] primers pass valgrindexe testing

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -1703,10 +1703,11 @@ static void applyGetterTransform(CallExpr* call) {
 static bool shouldInsertCallTemps(CallExpr* call);
 static void evaluateAutoDestroy(CallExpr* call, VarSymbol* tmp);
 static bool moveMakesTypeAlias(CallExpr* call);
+static Expr* getCallTempInsertPoint(Expr* expr);
 
 static void insertCallTemps(CallExpr* call) {
-  if (shouldInsertCallTemps(call) == true) {
-    insertCallTempsWithStmt(call, call->getStmtExpr());
+  if (shouldInsertCallTemps(call)) {
+    insertCallTempsWithStmt(call, getCallTempInsertPoint(call));
   }
 }
 
@@ -1773,6 +1774,27 @@ static bool shouldInsertCallTemps(CallExpr* call) {
     return false;
 
   return true;
+}
+
+// If we're inserting a call-temp for expr, where should we put it?
+static Expr* getCallTempInsertPoint(Expr* expr) {
+  Expr* stmt = expr->getStmtExpr();
+  if (CallExpr* call = toCallExpr(stmt)) {
+    if (call->isPrimitive(PRIM_INIT_VAR)) {
+      SymExpr* se = toSymExpr(call->get(1));
+      INT_ASSERT(se);
+      Symbol* sym = se->symbol();
+      // Put the call temps before the DefExpr for se->symbol()
+      // This allows callDestructors to use a simple strategy
+      // of destroying anything if the DefExpr for it has been
+      // reached. The case to think about is a function returning
+      // a record that throws (or a throwing initializer).
+      if (DefExpr* def = toDefExpr(stmt->prev))
+        if (def->sym == sym)
+          return def;
+    }
+  }
+  return stmt;
 }
 
 static void evaluateAutoDestroy(CallExpr* call, VarSymbol* tmp) {
--- a/compiler/resolution/callDestructors.cpp
+++ b/compiler/resolution/callDestructors.cpp
@@ -749,8 +749,11 @@ void ReturnByRef::transformMove(CallExpr* moveExpr)
 
   // Ignore a CondStmt containing a PRIM_CHECK_ERROR
   // so that we can still detect initCopy after a call that can throw
-  if (isCheckErrorStmt(nextExpr))
-      nextExpr = nextExpr->next;
+  //
+  // Also ignore a DefExpr which might e.g. define a user variable
+  // which is = initCopy(call_tmp).
+  while (nextExpr && (isCheckErrorStmt(nextExpr) || isDefExpr(nextExpr)))
+    nextExpr = nextExpr->next;
 
   CallExpr* copyExpr  = NULL;
 
@@ -826,7 +829,7 @@ void ReturnByRef::transformMove(CallExpr* moveExpr)
     if (rhsFn->hasFlag(FLAG_INIT_COPY_FN) && unaliasFn != NULL) {
       // BHARSH: It seems important that there's a temporary to store the
       // result of the unaliasFn call. Otherwise we'll move into a variable
-      // that has multiplie uses, which seems to cause a variety of problems.
+      // that has multiple uses, which seems to cause a variety of problems.
       //
       // In particular, I noticed that `changeRetToArgAndClone` generates
       // bad AST if I simply did this:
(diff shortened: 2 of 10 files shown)


mppf:
The fix seems to be causing a leak for array slices.

e.g.
``` chapel
var D: domain(1) = {1..10};
var A: [D] real;

proc test() {
  A(1..2) = 1;
}

test();
```

Test failures:
```
arrays/ferguson/from-iterator/skyline.chpl
arrays/ferguson/if-expr/aif-slice.chpl
arrays/ferguson/if-expr/aif-slice-param.chpl
arrays/ferguson/pushback-no-leak.chpl
arrays/ferguson/semantic-examples/0-init-slice-alias.chpl
arrays/ferguson/semantic-examples/0-return-global-slice.chpl
arrays/ferguson/semantic-examples/0-return-global-slice2.chpl
arrays/ferguson/semantic-examples/1.5-return-alias-slice-global.chpl
arrays/ferguson/semantic-examples/1.5-return-slice-global.chpl
arrays/ferguson/semantic-examples/1-return-promoted-expr.chpl
arrays/ferguson/semantic-examples/4-consume-return-slice-local.chpl
arrays/ferguson/semantic-examples/4-domain-array-class-field.chpl
arrays/ferguson/semantic-examples/4-global-slice-alias.chpl
arrays/ferguson/semantic-examples/4-init-array-forexpr2.chpl
arrays/ferguson/semantic-examples/4-init-slice.chpl
arrays/ferguson/semantic-examples/4-pass-slice-inout.chpl
arrays/ferguson/semantic-examples/4-return-tuple.chpl
arrays/ferguson/semantic-examples/4-sparse-subdomain-change-parent.chpl
arrays/ferguson/semantic-examples/9-arr-resize.chpl
arrays/ferguson/semantic-examples/9-other-dom-slice.chpl
arrays/ferguson/semantic-examples/9-yield-tuple-domains.chpl
arrays/ferguson/semantic-examples/9-yield-tuple-domains2.chpl
arrays/shapes/field-defaults.chpl
arrays/shapes/result-domain-follows-source-resize.chpl
compflags/diten/savec.chpl
compflags/lydia/library/exportArray/wrapExternArray.chpl
compflags/lydia/library/exportArray/wrapExternArray-real.chpl
distributions/robust/arithmetic/reindexing/simple-alias-mem-use.chpl
distributions/vass/negative-ref-count-dmap.chpl
expressions/loop-expr/scoping.chpl
functions/iterators/ferguson/no-leak-string-bug.chpl
io/ferguson/stdin-is-directory/stdin-is-directory.chpl
library/standard/Path/prithvi-unnati/commonPath/testCommonPathArray.chpl
library/standard/Path/prithvi-unnati/commonPath/testCommonPathStringArguments.chpl
library/standard/Path/victor-ludorum/testcases_isAbsPath.chpl
library/standard/Spawn/spawn-popen-input-error-long.chpl
library/standard/Spawn/spawn-popen-input-output-error.chpl
library/standard/Spawn/spawn-popen-input-output-error-long.chpl
library/standard/Spawn/spawn-system-shell.chpl
mason/chplVersion/mason-chpl-version-update.chpl
mason/mason-test/mason-test.chpl
mason/masonUpdateTest.chpl
mason/search/mason-search.chpl
memory/figueroa/LeakedMemory3.chpl
memory/figueroa/LeakedMemory5.chpl
memory/figueroa/LeakedMemory6.chpl
memory/sungeun/refCount/arrays.chpl
memory/sungeun/refCount/arraysAndDomains.chpl
memory/sungeun/refCount/arrays-rankchange.chpl
memory/sungeun/refCount/domainMaps.chpl
memory/sungeun/refCount/domains.chpl
parallel/forall/task-private-vars/arrays.chpl
parallel/forall/vass/memleaks2-BlockDist.chpl
patterns/cltools/fileCompile/chplCompile.chpl
performance/compiler/elliot/updateDatFiles.chpl
types/records/ferguson/array/array-in-record2.chpl
types/records/ferguson/array/array-in-record3.chpl
types/records/ferguson/array/array-literal.chpl
types/records/ferguson/array/iterate-array.chpl
types/records/ferguson/array/iterate-array2.chpl
types/records/ferguson/array/iterate-array-break.chpl
types/records/ferguson/array/iterate-array-zip.chpl
types/records/ferguson/array/iterate-array-zip2.chpl
types/records/ferguson/array/tuple-arrays.chpl
types/records/ferguson/array/tuple-unpack-arrays.chpl
types/records/ferguson/return/return-tuple.chpl
types/records/ferguson/return/return-tuple2.chpl
```

mppf:
Closing for now while I figure out what's going on with the leak.

mppf:
Close... 9 failures remain:

[Error matching compiler output for library/standard/DataFrames/psahabu/AddSeries]
[Error matching compiler output for library/standard/DataFrames/psahabu/ArithNoneSeries]
[Error matching compiler output for library/standard/DataFrames/psahabu/MixedOperateSeries]
[Error matching compiler output for library/standard/DataFrames/psahabu/MultiplySeries]
[Error matching compiler output for library/standard/DataFrames/psahabu/SubtractSeries]
[Error cannot locate program output comparison file mason/maso/src/maso.good]
[Error matching program output for mason/pkgconfig-tests/pcTest (execopts: 1)]
[Error matching program output for mason/pkgconfig-tests/pcTest (execopts: 2)]
[Error matching program output for mason/masonUpdateTest (execopts: 2)]

mppf:
[Error cannot locate program output comparison file mason/maso/src/maso.good]
[Error matching program output for mason/pkgconfig-tests/pcTest (execopts: 1)]
[Error matching program output for mason/pkgconfig-tests/pcTest (execopts: 2)]
[Error matching program output for mason/masonUpdateTest (execopts: 2)]

PR #10614, opened 2018-07-31
Mason: Examples

Spartee:
This PR represents a feature for mason users to run their examples within their project directories with the same, specifications, and dependencies as their source code. This PR also allows for users to set special configurations for examples in their mason.toml. See the blurb below from the issue for more details:

Files located in the examples folder of a mason project will now be able to be built and run using `mason build --example <example-name>` followed by `mason run --example <example-name>`. 

Each example will be required to have a main function and the binaries will be placed in `target/examples/`.

A new table in the `Mason.toml` will be added as `examples` where users can put configurations for their examples. Example names will not have to be specified by the user, but they can be under the `[examples]` table.


tools/mason/MasonExample.chpl
```diff
@@ -341,9 +343,11 @@ proc printAvailableExamples(toml: Toml, projectHome: string) {
 // Checks to see if an example, source code, or Mason.toml has been modified
 proc exampleModified(projectHome: string, projectName: string,
                              exampleName: string) {
+  const example = basename(stripExt(exampleName, ".chpl"));
   const examplePath = joinPath(projectHome, "target/example", exampleName);
 
-  if projectModified(projectHome, projectName, "debug") {
+  // check for changes to Mason.toml and src code
```
ben-albrecht: I think we still need to check the source code of the example itself

PR #10604, opened 2018-07-31
Additional SparseBlockDom/Arr dsi methods to support bale-toposort

ian-bertolacci:
Adds or implements the following methods:
+ SparseBlockDom.dsiMember
+ SparseBlockDom.dsiLocalSubdomain()
+ SparseBlockDom.dsiHasSingleLocalSubdomain()
+ SparseBlockArr.dsiLocalSubdomain
+ SparseBlockArr.dsiHasSingleLocalSubdomain()

Resolves #10471.

PR #10623, opened 2018-08-01
Fixes updated sparse_bulk_dist tests for 1 locale conditions

ian-bertolacci:
Updated tests in #10604 did not take into account what happens when testing on only one locale.
This fixes that.


bradcray:
From the description above, I'm not understanding what this change does nor why your previous PR necessitated it.  Why would this test be sensitive to the number of locales?  And why is it more sensitive now than it used to be?

(Also tagging @e-kayrakli since he wrote these tests).

e-kayrakli:
This modifies the parts of the tests that were added in #10604, so I am not familiar from before. 

(Looks like there was a missing check in #10604)

bradcray:
Thanks for the explanations!

Issue #10624, opened 2018-08-01
Compiling Multilocale with MPI and intel compilers


SwetaYamini:
Summary:
A way to override mpicc and mpicxx from environment flags to compile multilocale chapel using MPI conduit and intel compilers (with intel MPI).

Details:
When compiling  Chapel-1.17.1 with the settings

export CHPL_HOST_COMPILER=intel
export CHPL_TARGET_COMPILER=intel
export CHPL_COMM=gasnet
export CHPL_COMM_SUBSTRATE=mpi
export CHPL_LAUNCHER=gasnetrun_mpi

and environment

$ util/printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: intel *
CHPL_TARGET_ARCH: unknown
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: mpi *
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: gasnetrun_mpi *
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none

gives a warning and eventually fails to compile.

configure: WARNING: The compiler families of $CC and $MPI_CC do not match:
      CC:  COMPILER_FAMILY:INTEL	COMPILER_VERSION:1800.20170811
  MPI_CC:  COMPILER_FAMILY:GNU	COMPILER_VERSION:4.9.2
If this is not what you intended, then set $MPI_CC to select a different MPI compiler.
.....
.....
.....
make[3]: *** No rule to make target `XXX/chapel-1.17.1/third-party/gasnet/install/linux64-intel-unknown/substrate-mpi/seg-everything/nodbg/include/mpi-conduit/mpi-par.mak'.  Stop.
make[2]: *** [all.helpme] Error 2
make[1]: *** [runtime] Error 2
make: *** [comprt] Error 2

Following on the warning, and setting MPI_CC=mpiicc successfully compiles and chapel is built. However, it does not compile example program hello4.

 $ bin/linux64/chpl -o hello4 examples/hello4-datapar-dist.chpl 
[...]

PR #10626, opened 2018-08-01
Allow mpiccc and mpicxx values to be overridden

ronawho:
Previously we hardcoded LD=mpicxx for when GASNet required MPI support, but not
all MPI C++ wrappers use that name. e.g. Intel's is named mpiicpc.

This allows users to set MPI_CXX to override the default value (and gasnet
already supports setting MPI_CC to override the default compiler.

It also defaults to MPI_CC=mpiicc and MPI_CXX=mpiicpc for the Intel compiler

Resolves https://github.com/chapel-lang/chapel/issues/10624

### DIFF
--- a/make/compiler/Makefile.mpi-gnu
+++ b/make/compiler/Makefile.mpi-gnu
@@ -19,5 +19,13 @@
 
 include $(CHPL_MAKE_HOME)/make/compiler/Makefile.gnu
 
-CC=mpicc
-CXX=mpicxx
+ifeq (, $(MPI_CC))
+MPI_CC=mpicc
+endif
+
+ifeq (, $(MPI_CXX))
+MPI_CXX=mpicxx
+endif
+
+CC=$(MPI_CC)
+CXX=$(MPI_CXX)
--- a/runtime/etc/Makefile.comm-gasnet
+++ b/runtime/etc/Makefile.comm-gasnet
@@ -58,7 +58,10 @@ LD = $(GASNET_LD)   # GASNet chose C++ linker so stick with it
 else ifeq ($(GASNET_LD),$(GASNET_CC))
 LD = $(GASNET_CXX)  # GASNet chose C, so switch to C++
 else ifeq ($(GASNET_LD_REQUIRES_MPI),1)
-LD = mpicxx         # GASNet chose mpicc, so we'll choose mpicxx
+ifeq (, $(MPI_CXX))
+MPI_CXX = mpicxx    # GASNet defaults to mpicc, so we'll default to mpicxx
+endif
+LD = $(MPI_CXX)
 else
 LD = $(CXX)
 $(warning GASNET_LD unexpectedly set to '$(GASNET_LD)'.  Please file a Chapel bug against this.)
--- a/make/compiler/Makefile.intel
+++ b/make/compiler/Makefile.intel
@@ -25,6 +25,13 @@
 CXX = icpc
 CC = icc
 
+ifeq (, $(MPI_CC))
+MPI_CC = mpiicc
+endif
+ifeq (, $(MPI_CXX))
+MPI_CXX = mpiicpc
+endif
+
 RANLIB = ranlib
 
 


ronawho:
@bradcray mind reviewing you since you did the LD rework previously?

We probably want to document this somewhere, but we don't have an mpi substrate doc already and I figured it was good to get this in sooner rather than later to resolve the user bug. We may also want to consider setting MPI_CC and MPI_CXX to mpiicc and mpiicpc if we detect the intel compiler is being used, but I'm not positive that's a good idea.

bradcray:
I agree that I should review this, but I'm going to start by forwarding you some mails from @PHHargrove from when we last wrestled with this that may be relevant since you're cached up on this and I'm not.

> We may also want to consider setting MPI_CC and MPI_CXX to mpiicc and mpiicpc if we detect the intel compiler is being used, but I'm not positive that's a good idea.

This approach was my initial instinct...  (though I think it's still smart to permit a user to override)

make/compiler/Makefile.mpi-gnu
```diff
@@ -19,5 +19,13 @@
...
+ifeq (, $(MPI_CC))
+MPI_CC=mpicc
+endif
+
+ifeq (, $(MPI_CXX))
+MPI_CXX=mpicxx
+endif
+
+CC=$(MPI_CC)
+CXX=$(MPI_CXX)
```
bradcray: My first thought was "shouldn't we create an mpi-intel equivalent to this and set the relevant MPI_ variables there siilarly?  But then I couldn't find places where this Makefile is still used and wonder if it dates back to our CHPL_COMM=mpi days (in which case it should probably be retired).

ronawho: Ah good point -- this is still used as part of the MPI module that Nikhil added (and we test this config nightly)

runtime/etc/Makefile.comm-gasnet
```diff
@@ -58,7 +58,10 @@ LD = $(GASNET_LD)   # GASNet chose C++ linker so stick with it
 else ifeq ($(GASNET_LD),$(GASNET_CC))
 LD = $(GASNET_CXX)  # GASNet chose C, so switch to C++
 else ifeq ($(GASNET_LD_REQUIRES_MPI),1)
-LD = mpicxx         # GASNet chose mpicc, so we'll choose mpicxx
+ifeq (, $(MPI_CXX))
+MPI_CXX = mpicxx    # GASNet defaults to mpicc, so we'll default to mpicxx
```
bradcray: My intuition would be to replicate similar logic in Makefile.gnu and Makefile.intel (setting the variables to the Intel equivalents) such that the compiler Makefiles are establishing what they expect MPI_CC and MPI_CXX to be, and then this can serve as a fallback for any compiler that doesn't set them.  In saying that, I'm assuming that Makefile.gnu and Makefile.intel have long-since been included before this logic such that that would work.

On second thought, though, I'd probably only add it to Makefile.intel since there's nothing in particular about gnu that sets MPI_CXX to be mpicxx (I think it's more that it's using the default name as well?).  So maybe the logic I'd really expect to see in Makefile.intel is:

```
ifeq (, $(MPI_CC))
MPI_CC=mpiicc
endif
 ifeq (, $(MPI_CXX))
MPI_CXX=mpiicpc
endif
```

which would prevent the default setting of MPI_CXX that you've put here from firing and not require a user override?

make/compiler/Makefile.mpi-gnu
ronawho: So we may want to add a mpi-intel one as well, but unless you object I'll probably leave this change as since I don't think it hurts anything

runtime/etc/Makefile.comm-gasnet
ronawho: Yeah, good idea. Pushed a commit with that change. Tested that it works on they cray-cs I'm using to repro this, and verified that this does not cause any issues with PrgEnv-intel on a cray-xc.

make/compiler/Makefile.mpi-gnu
bradcray: > this is still used as part of the MPI module that Nikhil added

Oh, I see.  So the reason I wasn't finding it by grepping is that the user has to set up for it explicitly.  Makes sense (and I didn't think to git grep for "mpi-gnu" which would've gotten me there...

bradcray:
OK, this looks good to me and I feel about as confident in it as I do any makefile change (which is to say, not much at all).  Now that I understand the role of Makefile.mpi-gnu I think a Makefile.mpi-intel still would ultimately make sense, but think that's pretty separate from this effort since it'd be more like "make the MPI module work with other compilers" and require documentation updates...

I'm also fine with not adding documentation updates for this change since I think most users won't need it (given how long it's taken for anyone to report running into this issue) and that the new behavior is pretty reasonable / intuitive / self-explanatory.

PR #10638, opened 2018-08-01
File a pair of "error message" futures about type mismatches in generic initializations

bradcray:
These tests shouldn't compile, but don't generate clear error messages
that are likely to make sense to end-users.  Digging into them also
seems to suggest that we're still default initializing + assigning
records rather than simply initializing them in-place once.
Specifically, compiling in --devel mode with
--print-callstack-on-error seems to reveal calls to defaultOf() and
the compiler-generated assignment operator which seem surprising.
(Lydia confirms that this is the case for generic records still which
I'd apparently forgotten).

The .good files show what I think a better error message would be; the
.bad shows what they are today for users (for developers they differ
in ways that sometimes more illuminating, sometimes less so).

Also curious: I was surprised that adding the `proc init(type t)`
overload to mismatch2.chpl improved the behavior of mismatch.chpl.  It
doesn't seem like defining this initializer should be required does
it?  (or am I forgetting something?)

### DIFF
--- a/test/types/records/init/generic/mismatch.bad
+++ b/test/types/records/init/generic/mismatch.bad
@@ -1,0 +1,4 @@
+mismatch.chpl:16: error: unresolved call 'A.init(type int(64))'
+mismatch.chpl:5: note: candidates are: A.init(x)
+mismatch.chpl:10: note:                 A.init(a: A)
+mismatch.chpl:1: note:                 A.init(other: A)
--- a/test/types/records/init/generic/mismatch.chpl
+++ b/test/types/records/init/generic/mismatch.chpl
@@ -1,0 +1,17 @@
+record A {
+  type t;
+  var x: t;
+
+  proc init(x) {
+    this.t = x.type;
+    this.x = x;
+  }
+
+  proc init(a:A) { // copy initializer to avoid confusion with the above
+    this.t = a.t;
+    this.x = a.x;
+  }
+}
+
+var a: A(int) = new A(42.0);
+writeln(a);
--- a/test/types/records/init/generic/mismatch.future
+++ b/test/types/records/init/generic/mismatch.future
@@ -1,0 +1,7 @@
+error message: confusing error for mismatched generic type assignment
+
+This test is complaining about an unresolved call to an initializer
+that we're arguably not making.  Compiling in developer mode, it seems
+to be coming from one of our compiler-generated `defaultOf()` calls.
+But why should we be calling `defaultOf()` for a case that has an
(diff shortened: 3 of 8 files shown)


lydia-duncan:
So far as I know, we still use _defaultOf for *generic* records.  We just avoid it for concrete ones

bradcray:
@benharsh: These are the cases I was asking you about on chat.  Before I merge can you take a quick look and see if I'm missing anything / they seem surprising to you?  For reference, the output for devel compilations that I was using to understand them is here:

```
$ chpl --devel mismatch.chpl 
mismatch.chpl:1: error: unresolved call 'A.init(type int(64))'
mismatch.chpl:1: note:   instantiated from _defaultOf
mismatch.chpl:1: note:   instantiated from _defaultOf
mismatch.chpl:5: note: candidates are: A.init(x) [176612]
mismatch.chpl:10: note:                 A.init(a: A) [176642]
mismatch.chpl:1: note:                 init(_mt: _MT, this: A, other: A) [598490]
mismatch.chpl:1: note: unresolved call had id 853887

$ chpl --devel --print-callstack-on-error mismatch2.chpl 
mismatch2.chpl:1: error: type mismatch in assignment from real(64) to int(64)
  mismatch2.chpl:1: =(_arg1: A(int(64)), _arg2: A(real(64))) [854053] [compiler-generated]
  mismatch2.chpl:20: chpl__init_mismatch2() [298362]
mismatch2.chpl:1: note: unresolved call had id 854087
```

where the reference to defaultOf() in the first case and assignment in the second seemed surprising to me.

bradcray:
> So far as I know, we still use _defaultOf for generic records. We just avoid it for concrete ones

Huh...  Any idea what it would take to retire that?  (or why we haven't done so already?)

lydia-duncan:
> >    So far as I know, we still use _defaultOf for generic records. We just avoid it for concrete ones

> Huh... Any idea what it would take to retire that? (or why we haven't done so already?)

We perform the transformation for concrete records during normalization, when we don't know which generic instantiation is used.  Basically, normalization realizes the type you are trying to use and sets the type of the variable being initialized right away (or something like that).  But we don't know what instantiation to use for generic types until function resolution has run.

Basically, it's another "normalization/functionResolution" interaction

benharsh:
I don't think you're missing anything, Brad. The first case is most surprising to me, thought the second case is odd as well.

In anticipation of init= being approved I'm trying to move more things into function resolution, which might help solve these bugs going forward.

bradcray:
Thanks guys.

PR #10641, opened 2018-08-02
Deinitialize modules in code compiled using --library

### LINKED ISSUE #9779: Export module deinit functions, so they can be called by library clients
I'm ~90% certain that we don't automatically call chpl_deinitModules, since it is written in the generated code and I don't see it referenced in the call path for chpl_library_finalize.  We should provide a way to clean up after a .chpl file compiled using --library is done being used.

Interestingly, it seems vaguely more set up for multiple binaries than the module initialization functions do, due to iterating through an assembled list of deinit functions.

I believe we didn't notice this before for two reasons:
1) we weren't initializing the module to begin with
2) our memleaks tracking doesn't run C tests or the files that they compute to use, so we don't have a way to test the leaks from a compiled library.

lydia-duncan:
Resolves #9779 

I realized that we likely weren't deinitializing modules in code compiled
using --library, since we hadn't been initializing the modules to begin
with.  This change allows chpl_library_finalize to clean up any modules
that have been initialized.

This was ridiculously simple - just export chpl_deinitModules and have
chpl_library_finalize call it with an extern declaration.  I added an empty
version of chpl_deinitModules to the minimal modules implementation,
since we don't call the deinit functions in that mode.  This ensures that
we always include something that has generated the code for
chpl_deinitModules, so the operation is safe.

Adds a test to lock in this behavior, that failed prior to the change and now
passes.

Testing:
- [x] full standard paratest with futures (in progress)


lydia-duncan:
@vasslitvinov - you're probably the best person to review this.  I'm running a full paratest now but I don't expect any failures (since a regular program worked and a library mode program worked).  But I would totally understand waiting to check it out until testing has finished :)

lydia-duncan:
Okay, I was wrong, there were failures but they were all due to minimal modules.  There's a note saying that we don't deinit modules in that mode, but I don't see the rationale for it.  What's the reasoning?  I'm fine with adding a degenerate version of the clean up function but it seems a bit weird to me so I figured I would ask first

PR #10643, opened 2018-08-02
Prototype support for "always RVF" types

### LINKED ISSUE #9717: Create a pragma to enable remote value forwarding on immutable and/or forwarding records
Chapel's arrays are internally represented as a 'record-wrapped' structure, `_array`, that can be accessed inside of a `forall`, from any node, without inducing additional communication; this is because it is copied by-value by default instead of by-reference. If you go even further under the hood to understand why this is the case, we can see that it is hard-coded in the [compiler](https://github.com/chapel-lang/chapel/blob/4167b857aab884e635125d81b89dfa25100bbf26/compiler/AST/symbol.cpp#L1868-L1878), with its rationale behind such a design decision seen [here](https://github.com/chapel-lang/chapel/blob/90a5055ea34d79bf19e95f5c0752cfce24b0d967/compiler/passes/parallel.cpp#L1634-L1652). 

I believe that the user should be able to specify how their records should be passed by default, but in particular I believe that this optimization should be made available for developers who wish to exploit privatization similar to how Chapel's arrays do, for performance reasons. A pragma similar to the following would be appreciated...

```chpl
pragma "is record wrapped"
record R {
   var pid : int;
   var instance;
   // ...
   forwarding chpl_getPrivatizedCopy(instance.type, pid);
}
```

bradcray:
This adds a pragma "always RVF" that can be attached to a type in
order to request that the compiler always RVF objects of that type.
This is currently intended as an experimental feature, not for typical
end-users.  Over time, I think we need to think about whether this is
something we'd want to expose in the language, and if so, how to do so
(where one way might be to introduce a notion of 'intents' for
on-clauses and to permit type authors to override the default intent
of `ref` used for all types presently)..

Added a simple test to prove to myself that this feature is working.

Note that, as it stands, use of this feature can result in surprising
differences between --local and --no-local compilations since
r.v.f. doesn't have a role in --local compilations.  The difference
between the .good and .comm-none.good files in my test demonstrates
this difference.  If we were to promote this to a full language
concept, it might suggest that we should always r.v.f. / copy-in such
types across on-clauses to keep behavior as similar as possible
between the two modes.

OTOH, for cases where this feature arguably _ought_ to be used (ones
in which additional copies of the record and modifications to those
copies won't affect the object's semantics -- as with a record
wrapping an invariant class object where the class stores everything
important about the semantic state -- maybe it doesn't matter whether
the copy is made or not (and avoiding the copy would be preferable).
Put another way: The record in my example is not one which should ever
have this pragma attached to it, so maybe the variable output is
completely reasonable and as expected.

One other potential area of surprise is that we don't consider module-scope
symbols for RVF (I believe?), so there's also an asymmetry between how
local and "global" variables are treated w.r.t. this pragma, for better or worse.

I think there are interesting overlaps between this feature and
serialization; and in a recent discussion, BHarsh and I have been
discussing whether this feature + serialization would be a way to have
distributed slices be privatized lazily rather than aggressively as
they are today, optimizing away the overhead of creating them.

resolves #9717.


vasslitvinov:
Can we reuse FLAG_DONT_DISABLE_REMOTE_VALUE_FORWARDING for this?

bradcray:
> Can we reuse FLAG_DONT_DISABLE_REMOTE_VALUE_FORWARDING for this?

I mean, we _could_ in the sense that they apply to different things in different contexts (mine applies to types, this one applies to functions), so there'd be no overlap between their use cases.  But it seems to me that they're saying very different things, so it feels like a minor savings in number of flags for potential future developer confusion.  Mine says "objects of this type should always be RVF'd" whereas the existing flag says "don't let this function conservatively disable RVF because we're asserting that it shouldn't."

PR #10644, opened 2018-08-02
Test for radix sort MSB

avneet14027:
Added a test for radixsort - radixsortMSB.chpl.


test/library/packages/Sort/RadixSort/radixMSB.good
```diff
@@ -0,0 +1,3 @@
+radixMSB.chpl:14: In function 'radix_sort_msd':
```
ben-albrecht: Can we catch this with `--no-warnings` in a `radixMSB.compopts` file?

avneet14027: thanks, I made a  radixMSB.compopts for the same and updated the PR.

test/library/packages/Sort/RadixSort/radixMSB.chpl
```diff
@@ -0,0 +1,100 @@
+use BitOps;
+use Sort;
+use Random;
+
+config param bucketBits=8; 
+
+ module MSBmodule{
```
e-kayrakli: It looks like all your lines starting here have an extra space in the beginning

e-kayrakli: Also, a space before curly would be good. (Same applies to some functions below)

test/library/packages/Sort/RadixSort/radixMSB.chpl
```diff
@@ -0,0 +1,100 @@
...
+
+  proc radixSortMSBWrapper(array:[] int, place: int, startIndex: int, endIndex: int){
```
e-kayrakli: This function is the interface of your module. So, I think it needs some more improvements to make it more like a library function:

- It is true that this is wrapper, but the programmer doesn't need to know that :) This should have the name `radixSortMSB` and the other recursive one can be something like `radixSortMSBHelper` or `__radixSortMSB` etc. I am not fully sure whether there is a settled convention for naming such helpers. Personally, I am not sure even whether `MSB` should be in the name. But I guess that's open for discussion.

- The `place` argument looks like a helper argument similar to the `auxArray`. If I am not mistaken, I cannot see any use for that from the programmer's perspective. So it should be created inside the wrapper, similar to `auxArray`.

- I think whether `startIndex` and `endIndex` arguments should be exposed in the interface can be discussed. However, looking at Chapel's existing sort module, I don't see such arguments. (https://chapel-lang.org/docs/modules/packages/Sort.html). So, probably they should also be created by the wrapper to comply with the existing state-of-the-art.

However, if you wish to do something like this in Chapel, I would use two features of the language, domain queries for array arguments, and default argument values:

```chapel
proc radixSortMSBWrapper(array:[?dom] int, startIndex=dom.low, endIndex=dom.high)
```

`?dom` in the definition of `array` formal sort of creates a variable named `dom` that stores the array's domain. Then, you can use that "variable" to check the low and high indices of the domain and use them as defaults. The programmer can call this function either with only an array argument, or optionally with `startIndex` and/or `endIndex` for partial sorting.

For querying formal arguments check 10.8 of Specs (Someone could point to a better resource than this): https://chapel-lang.org/docs/_downloads/chapelLanguageSpec.pdf
For default arguments: https://chapel-lang.org/docs/primers/procedures.html#details-on-arguments

test/library/packages/Sort/RadixSort/radixMSB.chpl
```diff
@@ -0,0 +1,100 @@
...
+  proc findPlace(array:[] int) {
+    var place: int;
+    var lz: int;
+    var maxEl: int;
+    var numBits: int;
+
+    maxEl = findMaxElement(array);    
+    lz = clz(maxEl);
+    numBits = 64 - lz ;
+    place = ((numBits)>> 3)<<3;
```
e-kayrakli: Either remove the space after `>>` or add spaces around all the operators. Chapel still doesn't have an official style guide, but the ultimate rule is that you should be consistent :)

e-kayrakli:
Thanks for the fixes. They address my concerns in the previous iteration.

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
+ use BitOps;
```
ben-albrecht: If there is anything outside of a module scope in a Chapel program, the compiler will create an implicit module based on the file name. In this case, we are implicitly creating the following module hierarchy:

```chpl
module radixMSBModule {
  use BitOps;
  use Sort;
  use Random;
  config param bucketBits=8;

  module MSBmodule {
    ...
  }
}
```

You will want to put the `use` and `config param` statements inside the `MSBmodule` scope below to avoid this implicit module from being created.

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
...
+ use Sort;
+ use Random;
+ config param bucketBits=8; 
+
+ module MSBmodule {
```
ben-albrecht: Not sure how much it matters, since we intend for this to end up in the `Sort` module eventually, but we generally don't include the name `module` in a module name. 

`Radix`, `MSB` or `RadixMSB` might be a better fit.

test/library/packages/Sort/RadixSort/radixsortMSB.chpl
```diff
@@ -0,0 +1,29 @@
+ use BitOps;
+ use Sort;
+ use Random;
+ use radixMSBModule;
```
ben-albrecht: After addressing the implicit module above, we can change this to just `use <moduleName>` rather than `use <fileName>`

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   // To find maximum element in an array
+   private proc findMaxElement(array:[] int) {
+     return max reduce array;
+   }
+
+   // To find the place value from which msb sorting starts
+   private proc findPlace(array:[] int) {
+     var place: int;
```
ben-albrecht: I would advocate for making this a little more succint, by removing these declarations, and just putting the `var` keyword into the expressions below to make this a little more succinct. Also, I think we can safely rely on type inference rather than explicitly declaring them as `int` in this case, e.g.

```chpl
var maxEl = findMaxElement(array);
...
```

Furthermore, if we don't expect a value to change beyond instantiation, we could assert this by making it `const` instead of `var`, e.g.

```chpl
const maxEl = findMaxElement(array);
...
```

ben-albrecht: There are a few other cases where the above suggestion could be applied (`currentDigit`, `countIndex`, `auxIndex`, `position`)

test/library/packages/Sort/RadixSort/radixsortMSB.chpl
```diff
@@ -0,0 +1,29 @@
...
+ use Time;
+ config const printStats = true;
+
+ proc main() {
+  
+   var array:[1..10000] int;
```
ben-albrecht: It might be nice to make the array size a `config const` so that it's easy to test out different sizes. We may even integrate this into our performance test eventually, with multiple tests timing different problem sizes.

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
...
+  
+   private proc __radixSortMSB(array:[] int, place: int, startIndex: int, endIndex: int, auxArray:[] int) {
+
+     param numBuckets = (1 << bucketBits);
+     var counts:[1..numBuckets+1] int;  
+     var currentDigit: int; 
+     var countIndex: int;
+     var auxIndex: int;
+   
+     //base case, exit condition
```
mppf: In the future this should call a comparison sorter for < 100 or so elements. You could do that now if you like - it should significantly improve performance.

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
...
+   proc radixSortMSB(array:[?dom] int) {
+     var auxArray:[1..array.size] int;
+     var place = findPlace(array);
+     __radixSortMSB(array,place,dom.low,dom.high,auxArray);
+   }
+  
+   private proc __radixSortMSB(array:[] int, place: int, startIndex: int, endIndex: int, auxArray:[] int) {
+
+     param numBuckets = (1 << bucketBits);
+     var counts:[1..numBuckets+1] int;
```
mppf: Let's declare counts to start from 0, so we don't have to add 1 all the time: `0..numBuckets+1`

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
```diff
@@ -0,0 +1,83 @@
...
+     }
+
+     //update counts array
+     counts = + scan counts;
+
+     for i in startIndex..endIndex {
+       countIndex = findPosition(place,array[i], numBuckets);
+       auxIndex = counts[countIndex+1];     
+       auxArray[auxIndex] = array[i];   
+       counts[countIndex+1] -=1;
```
mppf: This still writes elements to auxArray "backwards" since counts[countIndex+1] is subtracted from here. Can we make curOffsets = 0, count[0], count[0] + count[1], ... and then add 1 to curOffsets?

test/library/packages/Sort/RadixSort/radixsortMSB.chpl
```diff
@@ -0,0 +1,29 @@
...
+  
+   var array:[1..10000] int; 
+   fillRandom(array); 
+   var t: Timer;   
+
+   for i in array.domain {
+     array[i] = abs(array[i]);
+   }  
+   
+	 t.start();
```
mppf: this line is just indented funny

test/library/packages/Sort/RadixSort/radixMSBModule.chpl
mppf: If you decide to put this off, make a `// TODO: call comparison sorter for small numbers of elements`.

test/library/packages/Sort/RadixSort/RadixMSB.chpl
```diff
@@ -0,0 +1,93 @@
...
+     }
+   
+     //swap slices of array and aux array   
+     array[startIndex..endIndex] <=> auxArray[1..(endIndex-startIndex)];
+ 
+
+     //recursive call for subparts of array
+     for m in 0..numBuckets-1 {        
+       if m==0 {
+         __radixSortMSB(array, place-bucketBits, startIndex, startIndex + curOffsets[m] -1, auxArray);
```
avneet14027: Reason for extra recursive call at m==0: 
For the purpose of this particular example, consider the **base is 10**:

Input array: 1, 5, 16, 3, 21, 5, 3, 1

**Previous method** - going backwards using counts array: 

Indices:  0 1 2 3 4 5 6 7 8 9
Counts:  6 1 1 0 0 0 0 0 0 0
Updated
Counts:  6 7 8 8 8 8 8 8 8 8
After the elements are put into their respective positions according to 1st MSB: 

Updated
Counts:  0 6 7 8 8 8 8 8 8 8 

The recursive calls here are done using the updated counts.
In the first recursive call of the loop, 
 new start_index = start_index + counts[m] = 1 + 0  = 1
 new end_index = end_index + counts[m+1] = 1 + 0  = 1

Next recursive call,
new start_index = 7
 new end_index = 7

**Current method** - going forward using curOffsets:

Indices:       0 1 2 3 4 5 6 7 8 9
Counts:       6 1 1 0 0 0 0 0 0 0
curOffsets:  0 6 7 8 8 8 8 8 8 8 
After the elements are put into their respective positions according to 1st MSB: 

Updated
curOffsets:   6 7 8 8 8 8 8 8 8 8
The recursive calls here are done using the updated curOffsets.

In the first recursive call of the loop, 
 new start_index = start_index + counts[m] = 1 + 6  = 7
 new end_index = end_index + counts[m+1] = 1 + 0  = 7

One recursive call, indicating the first bucket to be sorted, that is with start_index as 1 and end_index as 6, as indicated in the previous method example above, is missed here, so to cater for this, 2 recursive calls are made at m==0.

test/library/packages/Sort/RadixSort/RadixMSB.chpl
```diff
@@ -0,0 +1,93 @@
...
+     var place = findPlace(array);
+     __radixSortMSB(array,place,dom.low,dom.high,auxArray);
+
+   }
+  
+   private proc __radixSortMSB(array:[] int, place: int, startIndex: int, endIndex: int, auxArray:[] int) {
+
+     // comparison sorting for <=100 elements
+     if(array.size<=100) {
+       insertionSort(array);
```
mppf: Pretty sure this is just supposed to sort array[startIndex..endIndex]

test/GRAPHFILES
```diff
@@ -322,3 +322,5 @@ statements/lydia/moduleVersusFunction.graph
 # suite: Samples (bogus)
 Samples/Performance/foo.graph
 Samples/Performance/cleanCopy/foobar.graph
+# suite: Sample Radix Sort
+library/packages/Sort/RadixSort/radixsortMSB.graph
```
ben-albrecht: Despite not being added to the standard library yet, I think we should add this test to the standard library suite with the other sort performance tests (`# suite: Standard Library`)

test/library/packages/Sort/RadixSort/RadixMSB.chpl
mppf: I can think of two ways to maybe make this look nicer:
 1. Put the extra call before the loop
 2. Make curOffsets array start at -1 of do for m in -1..numBuckets-1

test/library/packages/Sort/RadixSort/radixsortMSB.graph
```diff
@@ -0,0 +1,5 @@
+perfkeys: Time:
```
ben-albrecht: Suggestion to be consistent with existing [sort performance tests](https://chapel-lang.org/perf/chapcs/?graphs=linearithmicsortson224bytesofshuffleddata,quadraticsortson212bytesofshuffleddata):

```
graphtitle: Radix sorts on 2^24 bytes of shuffled data
```

test/library/packages/Sort/RadixSort/radixsortMSB.execopts
```diff
@@ -0,0 +1 @@
+--printStats=false
```
ben-albrecht: Let's use the large problem size as other [sort performance tests](https://chapel-lang.org/perf/chapcs/?graphs=linearithmicsortson224bytesofshuffleddata,quadraticsortson212bytesofshuffleddata) of 2^24 bytes of data, or an array of `2097152` integers, so that our timings are comparable to other sorts in the Sort module.

However, we don't want the correctness test to use this problem size, because we generally want to keep correctness tests quick. To handle this, we can introduce a `radixsortMSB.perfexecopts` with the following content:

```
--printStats=false --size=2097152
```

This will be the execution options used when testing with `start_test --performance`.

test/library/packages/Sort/RadixSort/radixsortMSB.graph
ben-albrecht: If/when we add different data types to the performance tests, we can differentiate the types within the graph by the graph key, e.g. `Radix Sort MSB - int(64)`, `Radix Sort - string`, `Radix Sort MSB - uint(32)`, etc.

test/library/packages/Sort/RadixSort/RadixMSB.chpl
```diff
@@ -0,0 +1,91 @@
+ 
+ module RadixMSB {
```
ben-albrecht: We'll need a `RadixMSB.notest` in order for the test system to recognize that this is not in fact a test.

PR #10646, opened 2018-08-02
Add functionality to reissue compilerErrors

ronawho:
Previously if a compilerError was reissued it was emitted as a warning because
we only had support for reissuing warnings. This adjusts the warning reissuing
code so it can be used to reissue errors as well.

A bunch of .good files were updated for tests using --ignore-errors-for-pass to
correctly show reissued errors as errors instead of warnings.


ronawho:
> If it were me, I would create a small helper function to avoid duplicating the code in resolveNormalCallCompilerWarningStuff(). 


Added a helper function and transitioned to std::map while I was there

PR #10648, opened 2018-08-02
Make max number of mem regions user-settable and raise default.

### LINKED ISSUE #10438: Make ugni MAX_MEM_REGIONS env-var settable
Peeling off from https://github.com/chapel-lang/chapel/issues/6949, since some users have hit the hardcoded 1,024 limit: https://gitter.im/chapel-lang/chapel?at=5b513491f10ffa0b489cd198

gbtitus:
Up until now the maximum number of registered memory regions has been
statically fixed.  Most recently it has been set to 1024.  Here, make
this maximum user-settable via the `CHPL_RT_COMM_UGNI_MAX_MEM_REGIONS`
environment variable, and raise the default to 2048 on XE systems and
4096 on XC systems.  Document this and while doing so, update the
discussion about hugepages and registration to reflect our improved
understanding of Aries NIC behavior.

This resolves #10438.

### DIFF
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -508,13 +508,13 @@ region it registers with the Gemini or Aries NIC.  Roughly speaking there
 are a few memory regions for each tasking layer thread, plus one for each
 array larger than 2 hugepages allocated and registered separately from the
 heap.  By default the comm layer can handle up to 4k (2**12) total memory
-regions, which is plenty under normal circumstances.  In the unlikely
-event a program needs more than this, the following message will be
-printed:
+regions on Cray XC systems or 2k on XE systems, which is plenty under
+normal circumstances.  In the event a program needs more than this, a
+message like the following will be printed:
 
   .. code-block:: sh
 
-    warning: no more registered memory region table entries!
+    warning: no more registered memory region table entries (max is 4096)!
 
 To provide for more registered regions, set the
 ``CHPL_RT_COMM_UGNI_MAX_MEM_REGIONS`` environment variable to a number
--- a/doc/rst/platforms/cray.rst
+++ b/doc/rst/platforms/cray.rst
@@ -373,20 +373,34 @@ To use ugni communications:
    will probably give you satisfactory results.
 
    The Cray network interface chips (NICs) can only address memory that
-   has been registered with them, and there are limits on how many pages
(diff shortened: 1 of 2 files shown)


gbtitus:
It will probably be easiest to review this commit-by-commit.

gbtitus:
I'll make a test.

Yes, with the current coding I would expect the "no more registered memory region table entries" message to always be followed by our out-of-memory message except in fairly restricted circumstances.  Currently we return `NULL` from the comm-layer allocation routine if either we don't have a free table entry (and print the first message) or we can't get the memory.  If that happens on the call from the array allocator we try to get regular memory from the memory layer.  If it happens on the call to extend the heap we print the second message.  But because we don't try the array allocator except for allocations that by definition are larger than the jemalloc chunk size, if the array alloc fails the heap extension certainly will.  So by and large, you get both messages.  The only way you might avoid this is with a fixed-size heap.  We'll still try to do a separate array allocation in that case, and fail, but then we might succeed in getting the memory from the regular heap, if you've set the parameters just right.

I'll make an issue to study alternatives in this case.  We could choose just not to register if we run out of table entries, for example.

ronawho:
Move the test to `test/runtime/configMatters/comm/` so it runs under nightly XC/XE testing?

ronawho:
Oh, and sorry I didn't think of this sooner, but do we want to emit the name of the env var in the warning message?

gbtitus:
I don't think this is something we need to test nightly.  We don't test the SIGBUS stuff nightly, and it's arguably much more important.  And I thought configMatters was for stuff that was supported by multiple comm layers, but potentially in different ways.  (As opposed to stuff that's only supported by one comm layer.)

ronawho:
> I don't think this is something we need to test nightly. 

Seems weird to add a test if we're never going to test it :)

> We don't test the SIGBUS stuff nightly, and it's arguably much more important.

Oh, I didn't realize that -- I think we should move those tests into configMatters too

> And I thought configMatters was for stuff that was supported by multiple comm layers, but potentially in different ways. (As opposed to stuff that's only supported by one comm layer.)

I don't think there's a hard and fast line here -- the directory was originally chosen as a generic-ish name for things we want to test on Crays too. Mosts of the tests that are there are really ugni specific (or at the very least if it weren't for differences in ugni we would never have written the test)

I'd be in favor of moving this test and the sigbus test -- we're not tight on XC testing time, so I'm not worried about an extra couple minutes to run these tests.

PR #10652, opened 2018-08-03
Accept increased barrier comm counts

ronawho:
dba01db239 removed the on-smts from the barrier record wrapper in order to
optimize for network atomics where we don't want the on-stmts, but it did
introduce an extra get for each operation for the processor-atomic case. We
didn't see any performance hit for the processor atomic case and it
significantly helps the network atomics case, so accept the increased comm
counts. Note that dba01db239 effectively reverted 70d59aa843.

Longer term I wonder if we want to remove the record wrapper and go back to
just having the class. That would eliminate the extra gets and I think the main
motivation for the record wrapper was memory management, but with owned/shared
memory management I'm not sure that's necessary now.


ronawho:
@daviditen FYI

ronawho:
@daviditen FYI

dba01db239 removed the on-smts from the barrier record wrapper in order to
optimize for network atomics where we don't want the on-stmts, but it did
introduce an extra get for each operation for the processor-atomic case. We
didn't see any performance hit for the processor atomic case and it
significantly helps the network atomics case, so accept the increased comm
counts. Note that dba01db239 effectively reverted 70d59aa843.

Longer term I wonder if we want to remove the record wrapper and go back to
just having the class. That would eliminate the extra gets and I think the main
motivation for the record wrapper was memory management, but with owned/shared
memory management I'm not sure that's necessary now.

Issue #9741, opened 2018-06-06
can a method be both inline and override ?


mppf:
PR #9724 added a starting point implementation of the `override` keyword (see #8567 for more discussion about the keyword itself). But the starting point implementation does not allow a method to be both `inline` and `override`. Should such methods be allowed?

It's certainly the case that an `override` method can't always be inlined, since the dynamic type will determine which method will be called (so that's not known at compile time). However it could *sometimes* be inlined.

bradcray:
I think of our `inline` as being reasonably imperative (along the lines of "inline this and tell me if you can't (e.g., if it's recursive)") so I'd be inclined to not support this combination until someone comes up with a good motivator for it.

ben-albrecht:
Not a very compelling case, but [`ArrayViewSlice.dsiGetBaseDom()`](https://github.com/chapel-lang/chapel/blob/master/modules/internal/ArrayViewSlice.chpl#L299) is an instance where both `inline` and `override` seem to be needed.

bradcray:
I think this instance of `dsiGetBaseDom()` is a cases that really wants to be overriding a pure virtual method in `BaseArr`.  If we had pure virtual methods, do we imagine that the `override` keyword would be required to fulfill that interface?  (I'm hoping "not", but don't know what other languages do).

PR #10655, opened 2018-08-03
Add `override` to `modules/*` and have compiler enforce for all non-user code

bradcray:
This adds required `override` keywords to the `modules/*` directory with an eye toward the day when `--override-checking` is the default and also changes the compiler to enforce override checking for all non-user code in order to avoid backslides until that day comes.

For `override` methods that were already marked `inline`, I preserved the `inline` keyword in
comments so that if we choose to support the combination someday (issue #9741), we can
find them again easily and consider re-enabling the `inline`.


bradcray:
@mppf: For your consideration

PR #10657, opened 2018-08-03
Changes to support --library-makefile compilation with ./configure set up

lydia-duncan:
Michael pointed out that the prettifying changes for --library-makefile output
(which replaced all exact paths with $(CHPL_HOME) ) wouldn't be sufficient for
when Chapel was installed using ./configure.  This change adds the additional
makefile variables which will help with that.  It also copies the script used
to make these replacements when configuring Chapel - we ran into failures
otherwise

Tested with a full standard paratest
Also validated a fresh `./configure` install when CHPL_HOME was not in my path
(either for the ./configure/make step or for using the configured chpl)

### DIFF
--- a/compiler/codegen/library.cpp
+++ b/compiler/codegen/library.cpp
@@ -80,13 +80,25 @@ static std::string getCompilelineOption(std::string option) {
   std::string fullCommand = "$CHPL_HOME/util/config/compileline --" + option;
   fullCommand += "> cmd.out.tmp";
   runCommand(fullCommand);
+
   std::string replace = "$CHPL_HOME/util/config/replace-paths.py ";
+
+  replace += "--fixpath '$(CHPL_RUNTIME_LIB)' $CHPL_RUNTIME_LIB ";
+  replace += "--fixpath '$(CHPL_RUNTIME_INCL)' $CHPL_RUNTIME_INCL ";
+  replace += "--fixpath '$(CHPL_THIRD_PARTY)' $CHPL_THIRD_PARTY ";
   replace += "--fixpath '$(CHPL_HOME)' $CHPL_HOME < cmd.out.tmp";
 
   std::string res = runCommand(replace);
   std::string cleanup = "rm cmd.out.tmp";
   runCommand(cleanup);
   return res;
+}
+
+static void setupMakeEnvVars(std::string var, fileinfo makefile) {
+  std::string cmd = "echo $";
+  cmd += var;
+  std::string cmdRes = runCommand(cmd);
+  fprintf(makefile.fptr, "%s = %s\n", var.c_str(), cmdRes.c_str());
 }
 
 void codegen_library_makefile() {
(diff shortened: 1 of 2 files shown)


compiler/codegen/library.cpp
```diff
@@ -89,6 +94,13 @@ static std::string getCompilelineOption(std::string option) {
   return res;
 }
 
+static void setupMakeEnvVars(std::string var, fileinfo makefile) {
+  std::string cmd = "echo $";
+  cmd += var;
+  std::string cmdRes = runCommand(cmd);
```
mppf: I'm confused about why you added this code?  What case is it supporting?

lydia-duncan: The values for CHPL_HOME, CHPL_RUNTIME_INCL, etc, weren't necessarily going to be in the make environment unless I put them into the generated makefile.  I had already had that line back when we only added CHPL_HOME to the top of the file, but realized it would get redundant if I just copy+pasted it for the three new variables I was using.  So I made a helper function and passed the environment variable name in instead

lydia-duncan: I had messed up this step due to a typo for CHPL_RUNTIME_INCL, and it didn't get found in the configure environment (even though the replacement in the compileline output was valid)

lydia-duncan: I should probably comment it though, d'oh!

mppf: My question is more about why it runs `echo $SOMEVARIABLE`

lydia-duncan: To get the value of SOMEVARIABLE.  runCommand executes a command via popen, which I thought was just a shell command.  `$CHPL_HOME` on its own in my terminal gives an error message about its value being a directory.  Is there a better strategy to get the value that I have missed?

mppf: If you just want to know the value of an environment variable you can use getenv. The part that is unclear to me is - what $SOMEVARIABLE variables are present here that you're trying to get the value of?

In the PR it looks like you call this function like this:
```
  setupMakeEnvVars("CHPL_RUNTIME_LIB", makefile);
  setupMakeEnvVars("CHPL_RUNTIME_INCL", makefile);
  setupMakeEnvVars("CHPL_THIRD_PARTY", makefile);
  setupMakeEnvVars("CHPL_HOME", makefile);
```

`CHPL_RUNTIME_LIB`, `CHPL_RUNTIME_INCL`, `CHPL_THIRD_PARTY` and `CHPL_HOME` are all variables the compiler knows and uses and has C variables for (in driver.h). So I'm saying, why don't you just pass that?

e.g.
```
  setupMakeEnvVars("CHPL_RUNTIME_LIB", CHPL_RUNTIME_LIB, makefile);
```

(and not runCommand at all).

lydia-duncan: > why don't you just pass that?

Because I didn't know they existed :) I'll update the implementation to do that, thanks for pointing it out!

PR #10661, opened 2018-08-03
Fix another timeout for high core-count systems

dmk42:
#10530 turned off qthreads guard pages on ARM for certain benchmarks that were getting stuck in the `mprotect()` system call.  This change does the same thing for one more benchmark that was missed, test3DLulesh.


dmk42:
@ronawho - Would you please review?

PR #10674, opened 2018-08-06
Simplify normalization of field/var initialization

benharsh:
This PR collapses and merges a bunch of conditionals during normalization that handled field/variable initialization depending on the expression's result type. Since we can't reliably know the type at this stage of compilation, this PR pushes such handling towards resolution.

Also fixes some unknown/untested bugs involving usage of qualified "MOD.TYPE" expressions in new-exprs.

Testing:
- [x] local + futures
- [x] gasnet + futures
- [x] memleaks


test/compflags/bradc/mungeUserIdents/testmunge-extern.chpl
```diff
@@ -4,12 +4,16 @@ extern record bar {
...
 
-extern var baz: bar;
-var b: bar;
-
-foo();
-writeln(baz);
-writeln(baz.foo);
-b.foo = 1.2;
-writeln(b);
+// Use 'main' to ensure that 'bar' is present in the generatec C for this
```
lydia-duncan: Typo on "generatec"

test/compflags/bradc/mungeUserIdents/testmunge-extern.chpl
```diff
@@ -4,12 +4,16 @@ extern record bar {
...
+// module. If it's global, it will be in the header file where this test's
+// prediff will not look.
```
lydia-duncan: I would argue that implies we should add a test case that checks the header file

benharsh: Could that be part of a separate PR (which I will start right away) just to keep this branch cleaner?

lydia-duncan: I'm okay with that

PR #10686, opened 2018-08-07
Implement codepoint index, string length, conversion

### LINKED ISSUE #10476: Add methods to index and find string length in Unicode codepoints
This will be slow, but will provide useful functionality.

Counter to my earlier thinking, indexing can be done with `proc this()` because we can make it generic and supply a different type when we want to index by codepoint.

### LINKED ISSUE #10478: Add a conversion from int(32) Unicode codepoint to UTF-8 string
This will make the `uchars()` iterator more useful.  It returns int(32) codepoints which are currently only used internally by the String module.

dmk42:
This change adds indexing in Unicode codepoint units, string length in codepoints, and conversion from `int(32)` codepoints to strings.

Closes #10476 
Closes #10478


dmk42:
@mppf - Would you mind reviewing this after you get back?

modules/internal/String.chpl
```diff
@@ -151,6 +152,10 @@ module String {
     var shortData : chpl__inPlaceBuffer;
   }
 
+  record codePointIndex {
+    var cpidx     : int;
```
mppf: Is this intended to be user-visible? If so, I'm sure we can do better than the field name `cpidx`.

modules/internal/String.chpl
```diff
@@ -338,15 +343,35 @@ module String {
...
 
+    /*
+      :returns: The number of Unicode codepoints in the string.
+      */
+    proc ulength {
+      var localThis: string = this.localize();
+
+      var i = 0;
+      var n = 0;
+      while i < localThis.len {
```
mppf: Why doesn't this use the yield u-characters iterator?

modules/internal/String.chpl
```diff
@@ -1911,6 +1959,21 @@ module String {
     return s;
   }
 
+  /*
+     :returns: A string with complete multibyte character
+               with the codepoint value `i`.
```
mppf: Maybe "A string storing the Unicode code point `i`". Also consider renaming the argument to `codepoint`.

modules/internal/String.chpl
```diff
@@ -151,6 +152,10 @@ module String {
     var shortData : chpl__inPlaceBuffer;
   }
 
+  record codePointIndex {
+    var cpidx     : int;
+  }
```
mppf: Let's add _cast functions so that myCodePointIndex:int and myInt:codePointIndex will work.

test/types/string/dmk/utf8_misc.chpl
```diff
@@ -0,0 +1,11 @@
+var mystring1: string = "événement";
+var ulen1 = mystring1.ulength;
+
+writeln("Byte length = ", mystring1.length);
+writeln("Codepoint length = ", ulen1);
+
+for i in 1..ulen1 {
+  var cpi: codePointIndex;
+  cpi.cpidx = i;
```
mppf: I'd like us to encourage users to write this as var cpi = i:codePointIndex;

dmk42:
@mppf - Thanks!  This is ready for another look.

PR #10698, opened 2018-08-07
Generate pxd file when compiling with --library-python

### LINKED ISSUE #10442: Add a compilation flag for python interoperability with --library
While we should generate the files necessary to compile a Chapel library into a Python module, these files should not be generated by default - they should only be generated when requested by the user.

Design thought: if we were to support other languages in this way, we would probably want this flag to be able to be specified with other flags (e.g. `chpl --library --python --swift` or something along those lines).  It seems like it would be inconvenient to force a user to perform a separate compile for each style of library they wanted to make

### LINKED ISSUE #10443: Generate .pxd files for Chapel modules compiled with --library and python flag
This is the next file that needs to be generated in order to create a Python module from a Chapel library using Cython.  It should only be generated when the flag defined in #10442 is thrown during library compilation.  We might even want to clean it up afterwards by putting it in a temporary directory like our generated C code (not sure if that will cause problems later, need to experiment).

.pxd files have the format:
```python
cdef extern from "foo.h":
  retType funcName(argList)
  ...

cdef extern from "chpltypes.h":
  ctypedef void* c_fn_ptr

cdef extern from "chpl-init.h":
  void chpl_library_init(int argc, char* argv[])
  void chpl_library_finalize()
```
The function declarations under the library header are basically identical to the contents of the header itself, so should be easily generated.

Test code for this does not need to fully compile (although that is something I will do locally), I can just include it in the .catfiles for now.

lydia-duncan:
Adds support for the compilation flag `--library-python` and the first of the
files it should generate, a .pxd file.

Resolves #10442 
Resolves #10443

After we create the header file, check if we need to create a python module
and make the pxd file.  This file has a known start (which will allow types like
int64_t and uint64_t to be accessed from the .pxd file), then outputs all the
prototypes for the exported functions, then a known end (which includes
the library initialization function and library clean up function).

To support this change, I have added a map of primitive types to the relevant
type they should use in the .pxd file.  While I was there, I also stored the type
for the .pyx file (which is in Python syntax), though it is not used just yet.

Today, we test this change by comparing the .pxd file against what we expect
it to hold, instead of compiling it with Cython.  This is because there are
additional files that are needed for a Cython compile and because we need to
set up a testing mode where Cython is installed.  I have verified that the types
I output work with an actual Cython compile locally.

Updates the autocompletion script for the new flag

Tests cover:
- functions that take and return ints
- functions that take and return bools
- functions that take and return reals
- functions that take and return uints
- functions that take and return c_strings
- functions that take and return complex

Tested over full standard paratest with futures


lydia-duncan:
@benharsh - would you be willing to review this?

compiler/codegen/library.cpp
```diff
@@ -189,3 +196,163 @@ void closeLibraryHelperFile(fileinfo* fi, bool beautifyIt) {
...
+  pythonNames[dtComplex[COMPLEX_SIZE_64]->symbol] =
+              std::make_pair("float complex", "numpy.complex64");
+  pythonNames[dtComplex[COMPLEX_SIZE_128]->symbol] =
+              std::make_pair("double complex", "numpy.complex128");
+
+  /* TODO: // Handle bigint
+  forv_Vec(TypeSymbol, t, gTypeSymbols) {
+    if (
+  }
+  */
```
benharsh: Should this be removed?

compiler/codegen/library.cpp
```diff
@@ -189,3 +196,163 @@ void closeLibraryHelperFile(fileinfo* fi, bool beautifyIt) {
...
+static std::string getPXDTypeName(Type* type) {
+  std::pair<std::string, std::string> tNames = pythonNames[type->symbol];
+  if (tNames.first != "") {
+    return tNames.first;
+  } else {
+    return transformTypeForPointer(type);
+  }
+}
+
+std::string ArgSymbol::getPXDType() {
```
benharsh: I think we have tried to define codegen methods in codegen/symbol.cpp - do you think that's practical in this case?

lydia-duncan: It's certainly doable.  I was leaning towards keeping the related functionality in the same place so it would be easier to find it all, but having all the ArgSymbol functions in the same place for codegen also is reasonable

compiler/codegen/library.cpp
lydia-duncan: I want to come back to bigint, but I don't think it's a high priority.  I'll remove the code bit but leave the TODO

lydia-duncan:
@benharsh - good to go?

Issue #10495, opened 2018-07-24
Add buffered atomics as a package module


ronawho:
Take the prototype work done in #9834 and create a package module that allows
users to opt-in to buffered atomic operations (without us having to formalize
the MCM semantics)

LouisJenkinsCS:
Gotta say, I am glad that Chapel is going more towards aggregation now.

LouisJenkinsCS:
@ronawho Is there any coalescing of the buffered atomics? I.E: Do we combine multiple atomic operations destined for the same atomic? It looks like [here](https://github.com/ronawho/chapel/commit/ccd1d12077e77f9e6950beb34a9c5ee0ff98e9a5#diff-fea662522ed109229b4edf73480758e2R6786) you send them as individual active messages? I thought it was decided that aggregation of active messages are too heavy handed?

In this case, can't you check the table to see if there is a matching active message with the same destination that it can coalesce into? Of course then the issue is determining when the buffer should be flushed (automatically or otherwise; I still am in favor of using the communication thread).

LouisJenkinsCS:
Also, I may be wrong here, but I'd assume that you can coalesce the following kinds of non-blocking active messages...

1. `add_nb`: Should just add the operands together so long as no overflow occurs...
2. `sub_nb`: Should just subtract the operands together so long as no underflow occurs...

Am I missing any?

ronawho:
There is no coalescing and there are __no__ active messages being used. And even current blocking network atomics are much, much, much faster than active messages: https://chapel-lang.org/perf/16-node-xc/?startdate=2018/02/05&enddate=2018/07/24&configs=gnuugniqthreads&graphs=remoteamoperformance,remoteexecuteonperformance

That branch you pointed to uses ugni chained transactions issue multiple FMA operations in a single transaction. We build up a per-thread buffer of 64 operations and issue them as a single ugni operation, which is optimized by the ugni library.

LouisJenkinsCS:
Okay, allow me to rephrase what I said: Would there be a benefit to performing coalescing in the per-thread chained transactions? Does uGNI automatically perform its own coalescing for the transactions of network atomics?

ronawho:
[...]

PR #10702, opened 2018-08-08
Add initial support for buffered atomics

ronawho:
This adds a package module that supports buffered network atomics for
non-fetching operations. It includes `addBuff()`, `subBuff()`, `orBuff()`,
`andBuff()`, and `xorBuff()`. These buffered atomics are not consistent with
regular atomic operations and currently must be synchronized manually with a
call to `flushAtomicBuff()`. e.g.

```chpl
var a: atomic int;
a.addBuff(1);
writeln(a);        // prints 0
flushAtomicBuff();
writeln(a);        // prints 1
```

Under the covers these buffered atomics are implemented with 64 element
thread-local buffers. When full, these buffers are flushed. We use chained
transactions to initiate the network atomics, which offers significantly
increased transaction rate.

For the bale histogram benchmark this improves performance by over 4x.

Currently the code is written like:

```chpl
forall r in rindex do
  A[r].add(1);
```

And with buffered atomics looks something like:

```chpl
use BufferedAtomics;
forall r in rindex do
  A[r].addBuff(1);
flushAtomicBuff();
```

On a 28-core broadwell system the normal atomic implementation gets 80 MB/s per
node and the buffered atomic implementation achieves 360 MB/s per node, which
is on par with the non-blocking UPC implementation.

Using buffered atomics significantly improve ra-atomics as well and we see a
similar ~4.5x speedup. Scalability results with 256M hugepages up to 256 nodes
show below:

ra-perf

Closes https://github.com/chapel-lang/chapel/issues/10495

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -6879,22 +6879,32 @@ void do_nic_amo_nf_buff(void* opnd1, c_nodeid_t locale,
   static __thread c_nodeid_t node_v[MAX_CHAINED_AMO_LEN];
   static __thread int64_t vi = -1;
 
-  static __thread chpl_bool inited = false;
+  // init_status -- (0=first-call, -1=out-of-entries, 1=have-entry)
+  static __thread int init_status = 0;
   static __thread atomic_bool lock;
 
-  if (inited == false) {
+  if (init_status == 1) {
+    // no-op
+  } else if (init_status == -1) {
+    do_nic_amo_nf(opnd1, locale, object, size, cmd, remote_mr);
+    return;
+  } else {
     uint32_t idx;
     while (atomic_exchange_bool(&amo_init_lock, true)) { local_yield(); }
     atomic_init_bool(&lock, false);
     idx = atomic_load_uint_least32_t(&amo_pdc_sz);
     if (idx >= MAX_BUFF_AMO_THREADS) {
+      chpl_warning("Exceeded MAX_BUFF_AMO_THREADS, buffered atomic performance degraded", 0, 0);
+      init_status = -1;
+      atomic_store_bool(&amo_init_lock, false);
       do_nic_amo_nf(opnd1, locale, object, size, cmd, remote_mr);
+      return;
     }
     amo_pdc[idx] = &post_desc;
     amo_node_v[idx] = &node_v[0];
     amo_vi_v[idx] = &vi;
     amo_lock_v[idx] = &lock;
-    inited = true;
+    init_status = 1;
     (void) atomic_fetch_add_uint_least32_t(&amo_pdc_sz, 1);
     atomic_store_bool(&amo_init_lock, false);
   }
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -6550,7 +6550,7 @@ DEFINE_CHPL_COMM_ATOMIC_CMPXCHG(real64, cmpxchg_64, int_least64_t)
         {                                                               \
           mem_region_t* remote_mr;                                      \
           DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
-                   "IFACE chpl_comm_atomic_"#_o"_"#_f                   \
+                   "IFACE chpl_comm_atomic_"#_o"_buff_"#_f              \
                    "(%p, %d, %p)",                                      \
                    opnd, (int) loc, obj);                               \
                                                                         \
@@ -6682,7 +6682,7 @@ DEFINE_CHPL_COMM_ATOMIC_INT_OP(uint64, add, add_i64, uint_least64_t)
         {                                                               \
           mem_region_t* remote_mr;                                      \
           DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
-                   "IFACE chpl_comm_atomic_add_"#_f                     \
+                   "IFACE chpl_comm_atomic_add_buff_"#_f                \
                    "(%p, %d, %p)",                                      \
                    opnd, (int) loc, obj);                               \
                                                                         \
@@ -6774,7 +6774,7 @@ DEFINE_CHPL_COMM_ATOMIC_REAL_OP(real64, add_r64, double)
           _t nopnd = - *(_t*) opnd;                                     \
                                                                         \
           DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
-                   "IFACE chpl_comm_atomic_sub_"#_f                     \
+                   "IFACE chpl_comm_atomic_sub_buff_"#_f                \
                    "(%p, %d, %p)",                                      \
                    opnd, (int) loc, obj);                               \
                                                                         \
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -1507,6 +1507,7 @@ static void      do_nic_amo(void*, void*, c_nodeid_t, void*, size_t,
                             gni_fma_cmd_type_t, void*, mem_region_t*);
 static void      do_nic_amo_nf(void*, c_nodeid_t, void*, size_t,
                                gni_fma_cmd_type_t, mem_region_t*);
+static void      buff_amo_init(void);
 static void      do_nic_amo_nf_buff(void*, c_nodeid_t, void*, size_t,
                                     gni_fma_cmd_type_t, mem_region_t*);
 static void      amo_add_real32_cpu_cmpxchg(void*, void*, void*);
@@ -1991,6 +1992,7 @@ void chpl_comm_post_task_init(void)
   nb_desc_init();
   rf_done_init();
   amo_res_init();
+  buff_amo_init();
 
   //
   // Create all the communication domains, including their GNI NIC
@@ -6837,6 +6839,14 @@ static atomic_bool* amo_lock_v[1024];
 static atomic_bool amo_init_lock = false;
 static atomic_uint_least32_t amo_pdc_sz = 0;
 #endif // HAVE_GNI_FMA_CHAIN_TRANSACTIONS
+
+static
+void buff_amo_init(void) {
+#if HAVE_GNI_FMA_CHAIN_TRANSACTIONS
+  atomic_init_bool(&amo_init_lock, false);
+  atomic_init_uint_least32_t(&amo_pdc_sz, 0);
+#endif // HAVE_GNI_FMA_CHAIN_TRANSACTIONS
+}
 
 void chpl_comm_atomic_buff_flush() {
 #if HAVE_GNI_FMA_CHAIN_TRANSACTIONS
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -6832,10 +6832,11 @@ int amo_cmd_2_nic_op(fork_amo_cmd_t cmd, int fetching)
 }
 
 #if HAVE_GNI_FMA_CHAIN_TRANSACTIONS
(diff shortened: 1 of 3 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -752,11 +752,12 @@ static size_t rdma_threshold = DEFAULT_RDMA_THRESHOLD;
...
 //
-// Maximum number of PUTs in a chained transaction list.  This number
-// was determined empirically (on XC/Aries).  Doing more than this at
-// once didn't seem to improve performance.
+// Maximum number of PUTs/AMOs in a chained transaction list.  This
+// number was determined empirically (on XC/Aries).  Doing more than
+// this at once didn't seem to improve performance.
 //
 #define MAX_CHAINED_PUT_LEN 64
+#define MAX_CHAINED_AMO_LEN 64
```
ronawho: FYI I did tuning on this number and did not just copy the put number. I see max performance with anything above 32, and saw degraded performance with 128 or higher.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6710,6 +6795,123 @@ int amo_cmd_2_nic_op(fork_amo_cmd_t cmd, int fetching)
...
+    inited = true;
+    (void) atomic_fetch_add_uint_least32_t(&amo_pdc_sz, 1);
+    atomic_store_bool(&amo_init_lock, false);
+  }
+
+  if (remote_mr == NULL)
+    CHPL_INTERNAL_ERROR("do_nic_amo(): "
+                        "remote address is not NIC-registered");
+
+  while (atomic_exchange_bool(&lock, true)) { local_yield(); }
```
ronawho: I had been worried about the cost of this lock for each operation, but it ended up being pretty low overhead since in the normal case it's never contended. I see around 365 MB/s  for bale-histo without this lock and 360 MB/s with it.

ronawho: And note that the lock is needed for 2 reasons:
 - so flushing can be done concurrently
 - so we don't run into issues with a second task trying to use the buffer when we're yielding in `post_fma_ct_and_wait`

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6499,6 +6505,38 @@ DEFINE_CHPL_COMM_ATOMIC_CMPXCHG(real64, cmpxchg_64, int_least64_t)
...
                                                                         \
         /*==============================*/                              \
+        void chpl_comm_atomic_##_o##_buff_##_f(void* opnd,              \
+                                               int32_t loc,             \
+                                               void* obj,               \
+                                               int ln, int32_t fn)      \
+        {                                                               \
+          mem_region_t* remote_mr;                                      \
+          DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
+                   "IFACE chpl_comm_atomic_"#_o"_"#_f                   \
```
gbtitus: ... "_buff_"#_f

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6601,6 +6639,37 @@ DEFINE_CHPL_COMM_ATOMIC_INT_OP(uint64, add, add_i64, uint_least64_t)
...
                                                                         \
         /*==============================*/                              \
+        void chpl_comm_atomic_add_buff_##_f(void* opnd,                 \
+                                            int32_t loc,                \
+                                            void* obj,                  \
+                                            int ln, int32_t fn)         \
+        {                                                               \
+          mem_region_t* remote_mr;                                      \
+          DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
+                   "IFACE chpl_comm_atomic_add_"#_f                     \
```
gbtitus: ... `_add_buff_"`

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6658,6 +6727,22 @@ DEFINE_CHPL_COMM_ATOMIC_REAL_OP(real64, add_r64, double)
...
+         /*==============================*/                             \
+        void chpl_comm_atomic_sub_buff_##_f(void* opnd,                 \
+                                            int32_t loc,                \
+                                            void* obj,                  \
+                                            int ln, int32_t fn)         \
+        {                                                               \
+          _t nopnd = - *(_t*) opnd;                                     \
+                                                                        \
+          DBG_P_LP(DBGF_IFACE|DBGF_AMO,                                 \
+                   "IFACE chpl_comm_atomic_sub_"#_f                     \
```
gbtitus: ... `_sub_buff_"`

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6710,6 +6795,123 @@ int amo_cmd_2_nic_op(fork_amo_cmd_t cmd, int fetching)
...
+  static __thread c_nodeid_t node_v[MAX_CHAINED_AMO_LEN];
+  static __thread int64_t vi = -1;
+
+  static __thread chpl_bool inited = false;
+  static __thread atomic_bool lock;
+
+  if (inited == false) {
+    uint32_t idx;
+    while (atomic_exchange_bool(&amo_init_lock, true)) { local_yield(); }
+    atomic_init_bool(&lock, false);
```
gbtitus: If you want to take a teensy bit of overhead out, `lock` can be inited outside the single-threaded region because it's private to this thread.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6710,6 +6795,123 @@ int amo_cmd_2_nic_op(fork_amo_cmd_t cmd, int fetching)
   return nic_cmd;
 }
 
+#if HAVE_GNI_FMA_CHAIN_TRANSACTIONS
+#define MAX_BUFF_AMO_THREADS 1024
```
gbtitus: This aspect of the design troubles me.  These are statically sized and have the potential to get used up during execution.  (The likely way that would happen is probably not long-lived tasking layer pthreads because we won't ever create 1k of those even with tasks=fifo, but rather other pthreads from other programming models or the user, coming in here from non-Chapel code calling Chapel code in a multi-language program.  Once an entry is used here it's never freed, so a program that creates a pthread which calls Chapel code, ends up grabbing an entry here, and then terminates will leave the entry dangling and unusable.)  At the very least we should emit a warning when they do get used up, since we're going to fall off a performance cliff for super-relaxed AMOs at that point.  As an alternative design that doesn't have static sizing, how about defining a thread-private struct containing all the meta-information for each thread's buffered AMOs, and then just chain those together on a (doubly) linked list, under protection of a lock?  Adding the thread-private struct to the list could be done under an `inited` flag in `do_nic_amo_nf_buff()` where you do initialization now.  You'd also need to create a pthread cleanup function (see `othread_cleanup_push()`) to remove the thread's struct from the chain if/when it terminated.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6710,6 +6795,123 @@ int amo_cmd_2_nic_op(fork_amo_cmd_t cmd, int fetching)
...
+  static __thread chpl_bool inited = false;
+  static __thread atomic_bool lock;
+
+  if (inited == false) {
+    uint32_t idx;
+    while (atomic_exchange_bool(&amo_init_lock, true)) { local_yield(); }
+    atomic_init_bool(&lock, false);
+    idx = atomic_load_uint_least32_t(&amo_pdc_sz);
+    if (idx >= MAX_BUFF_AMO_THREADS) {
+      do_nic_amo_nf(opnd1, locale, object, size, cmd, remote_mr);
```
gbtitus: Don't we need to set `inited`, clear the lock, and return after this?  If we continue on we over-index `amo_pdc[]` and the rest (not to mention that the AMO will have already been done by this call and therefore shouldn't be buffered anyway).

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -7679,17 +7881,32 @@ int post_fma_ct(c_nodeid_t* locale_v, gni_post_descriptor_t* post_desc)
...
-      if (post_desc->type == GNI_POST_FMA_PUT)
-        PERFSTATS_ADD(sent_bytes, pdc->length);
-      else
-        PERFSTATS_ADD(rcvd_bytes, pdc->length);
+      for (pdc = post_desc->next_descr, i = 1;
+           pdc != NULL;
+           pdc = pdc->next_descr, i++) {
+        pdc->ep_hndl = cd->remote_eps[locale_v[i]];
+        if (post_desc->type == GNI_POST_FMA_PUT)
+          PERFSTATS_ADD(sent_bytes, pdc->length);
```
gbtitus: I think you can just remove the if-stmt here and do the `PERFSTATS_ADD(sent_bytes, ...)` only.  This was originally here to cover the case where we had chained GETs as well as PUTs, but we don't have chained GETs now and if we add them in the future we can update this.

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -7679,17 +7881,32 @@ int post_fma_ct(c_nodeid_t* locale_v, gni_post_descriptor_t* post_desc)
...
+      }
+    } else
+      if (post_desc->type == GNI_POST_AMO) {
+      gni_ct_amo_post_descriptor_t* pdc;
+      int i;
+       for (pdc = post_desc->next_descr, i = 1;
+           pdc != NULL;
+           pdc = pdc->next_descr, i++) {
+        pdc->ep_hndl = cd->remote_eps[locale_v[i]];
+        if (post_desc->type == GNI_POST_FMA_PUT)
```
gbtitus: This whole if-stmt can be removed.

runtime/src/comm/ugni/comm-ugni.c
ronawho: Oh shoot, yes this is missing the early return and init_lock unlock

runtime/src/comm/ugni/comm-ugni.c
ronawho: Offline we agreed to emit a warning for now, and we can address this later -- practically speaking this is unlikely to bite us in the near future (at least for Chapel-only programs.)

ronawho:
Rebased, fixed the logic for when we run out of table entries (and added a warning), fixed the debug prints, and renamed `atomicBuffFlush()` to `flushAtomicBuff()`

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -7690,15 +7902,24 @@ int post_fma_ct(c_nodeid_t* locale_v, gni_post_descriptor_t* post_desc)
...
+  if (post_desc->type == GNI_POST_FMA_PUT) {
     gni_ct_put_post_descriptor_t* pdc;
     int i;
 
     for (pdc = post_desc->next_descr, i = 1;
          pdc != NULL;
          pdc = pdc->next_descr, i++) {
       pdc->ep_hndl = cd->remote_eps[locale_v[i]];
-      PERFSTATS_ADD_POST(post_desc);
+      PERFSTATS_ADD(sent_bytes, post_desc->length);
```
gbtitus: `pdc->length`?

ronawho:
> Looks like there's no documentation for this yet (?). Can we add it by the release?

Correct, and yes I plan on adding it (it's the first thing listed in https://github.com/chapel-lang/chapel/issues/10743)

ronawho:
>> Looks like there's no documentation for this yet (?). Can we add it by the release?

> Correct, and yes I plan on adding it (it's the first thing listed in #10743)

Initial docs added in https://github.com/chapel-lang/chapel/pull/10900 and are available at https://chapel-lang.org/docs/master/modules/packages/BufferedAtomics.html. They've been reviewed by Greg and Bharsh so far, but I'll probably send them out for general group feedback too

PR #10709, opened 2018-08-08
Fix rcvd_bytes and sent_bytes performance counters in ugni

ronawho:
Previously, we incorrectly counted GNI_POST_AMO as part of rcvd_bytes because
it fell into an else branch. Now we only track PUT/GET FMA/RDMA operations.
This also adds a helper function for that logic partially to simplify call
sites and partially to eliminate branches that aren't needed when perfstats
aren't enabled (although the compiler probably optimized these away anyways)

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -382,6 +382,24 @@ static void perfstats_init(void)
 
 #define PERFSTATS_INIT() perfstats_init();
 
+static inline void perfstats_add_post(gni_post_descriptor_t* post_desc) {
+  switch (post_desc->type) {
+    case GNI_POST_FMA_PUT:
+    case GNI_POST_RDMA_PUT:
+      PERFSTATS_ADD(sent_bytes, post_desc->length);
+      break;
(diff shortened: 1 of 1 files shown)


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -7686,10 +7698,7 @@ int post_fma_ct(c_nodeid_t* locale_v, gni_post_descriptor_t* post_desc)
          pdc != NULL;
          pdc = pdc->next_descr, i++) {
       pdc->ep_hndl = cd->remote_eps[locale_v[i]];
-      if (post_desc->type == GNI_POST_FMA_PUT)
-        PERFSTATS_ADD(sent_bytes, pdc->length);
-      else
-        PERFSTATS_ADD(rcvd_bytes, pdc->length);
+      PERFSTATS_ADD_POST(post_desc);
```
ronawho: Shoot, this is wrong. I'm adding post_desc every time here instead of the current pdc -- will fix in an upcoming PR

ronawho: Fixed in https://github.com/chapel-lang/chapel/pull/10725

PR #10710, opened 2018-08-08
Have library mode flags trigger library mode if it is not already set

### LINKED ISSUE #10593: Make --library-* flags turn on library compilation
We do this already for --log-dir, --log-pass, etc.  It seems reasonable to do so for the various --library flags I am adding

lydia-duncan:
Prior to this change, passing one of ``--library-dir``, ``--library-header``,
``--library-makefile`` or ``--library-python`` without ``--library`` would not
trigger a library compilation and also would not complain that the additional
flag was not set.

Since the flags are kinda long (but need to be to distinguish what operations
will happen), it seemed unfortunate to require both when those flags imply that
``--library`` is necessary.  So, have their presence turn on library compilation
if it is not already turned on.

Adds a test for each of the impacted flags to a new test directory where
``--library`` is not thrown.

Resolves #10593 

Passed full standard paratest with futures


test/compflags/lydia/library/noLibFlag/callDirSet.lastcompopts
```diff
@@ -0,0 +1 @@
+-Lmydir/ -lsetDir `$CHPL_HOME/util/config/compileline --libraries`
```
mppf: pesky missing newline

Issue #10707, opened 2018-08-08
`override` checking complains incorrectly for generic child class


bradcray:
### Summary of Problem

I ran across an erroneous case in override checking while converting tests to use `override` (in the DataFrames.chpl tests specifically).  The behavior was that the compiler would complain whether or not a method had `override`—either that it needed it (which it did) or that the parent class didn't have a matching method.  I think that the behavior relies on the child class being generic and on the argument to the method being non-trivial in some way (`int` didn't seem to break things but `shared class` did).

### Steps to Reproduce

**Associated Future Test(s):**
[`test/functions/resolution/overload-in-generic.chpl`](
           https://github.com/chapel-lang/chapel/blob/master/test/functions/resolution/overload-in-generic.chpl
          ) #10706

### Configuration Information

- Output of `chpl --version`: `1.18.0 pre-release (b84c752)`

bradcray:
@mppf: Tagging you on this to bring it to your attention (no rush for my sake)

PR #10721, opened 2018-08-09
Convert mason and some tests to `override`

bradcray:
This converts mason and some tests in which files are shared between many cases over to use `override` such that they pass (or work better) with `--override-checking`.  Though it only touches 9 files, it makes ~100 tests pass that hadn't before.  Note that the DataFrames updates should make additional tests pass once issue #10707 is closed.

### DIFF
--- a/test/errhandling/ExampleErrors.chpl
+++ b/test/errhandling/ExampleErrors.chpl
@@ -16,7 +16,7 @@ class StringError : Error {
     this.msg = msg;
   }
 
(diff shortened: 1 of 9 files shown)


bradcray:
@mppf: As we discussed, I think this is trivial, but tagging you on it just FYI
@spartee: Tagging you here because of the (minor) change to mason, just as an FYI

Spartee:
LGTM

PR #10722, opened 2018-08-09
Const check tuple args

### LINKED ISSUE #10711: default intent tuple argument modifies call site
``` chapel
proc fun(arg: 3*int) {
  arg(1) = 0;
}

proc main() {
  var tup = (1,3,6);

  fun(tup);

  writeln(tup);
}
```

Should this program output 0 or 1, or be a compilation error?

Right now it outputs 0.

mppf:
Resolves #10711.

This PR adjusts cullOverReferences to make references created by
PRIM_GET_MEMBER to be considered const if the aggregate record/tuple is
const. It corrects several module code errors revealed by the new
checking.

- [x] full local testing

### DIFF
--- a/compiler/resolution/cullOverReferences.cpp
+++ b/compiler/resolution/cullOverReferences.cpp
@@ -664,7 +664,8 @@ void cullOverReferences() {
         SymExpr* aggregateSe = toSymExpr(call->get(1));
         SymExpr* lhsSe = toSymExpr(move->get(1));
         AggregateType* at = toAggregateType(aggregateSe->getValType());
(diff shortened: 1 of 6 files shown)


vasslitvinov:
@mppf - Separate from this PR, seems like capture-2-refs-to-ref.chpl should actually compile and execute, rather than reporting an error as it does according to its .good. What do you think?

vasslitvinov:
@mppf - Separate from this PR, seems like capture-2-refs-to-ref.chpl should actually compile and execute, rather than reporting an error as it does according to its .good. What do you think?

Issue #9879, opened 2018-06-14
simple UDRI with examples


vasslitvinov:
This is a simplification of #9542 .

## Interface

```chpl
record ReduceOp {

  // Create the global op record for reducing values of 'inputType'.
  // Not used when the user provides their own instance.
  // This could be the compiler-generated default constructor,
  // if applicable.
  proc init(type inputType)

  // Create a per-task op given 'parentOp' in the parent task.
  // This could be the compiler-generated copy initializer.
  proc init(parentOp :ReduceOp)

  // Create a per-task accumulation state.
  proc newAccumState() :accumType

  // Accumulate an input value onto the current task's
  // accumulation state 'AS'.
  proc accumulate(ref AS :accumType, input :inputType) :void

  // Combine the parent task's accumulation state 'parentAS'
  // with the child task's accumulation state 'childAS',
  // storing the result back into 'parentAS'.
  // Optional. If not defined, accumulate(parentAS, childAS) is used.
  proc combine(ref parentAS :accumType, childAS :accumType) :void

/* The remaining requirements depend on whether globalAS (the global
   accumulation state) needs to be post-processed into the reduction
   result after all values to be reduced have been processed.

 Case "no generate" : when the final globalAS already IS
      the result of the reduction.

     If so, generate(globalAS) must NOT be defined.
     The requirements below do NOT apply.
     In the case of a reduce intent, the outer variable is used
     for globalAS.

 Case "with generate" : when globalAS needs to be post-processed
      to generate the reduction result.

     If so, generate(globalAS) must be defined.
     The requirements below apply.
*/

  // Generate the reduction result a.k.a output given the global
  // accumulation state 'globalAS', after all values to be reduced
  // have been processed.
  proc generate(globalAS :accumType) :outputType

  // Generate the reduction result in the case of a reduce intent
  // into the outer variable 'outerVar'.
  // Note that semantics of the reduce intent requires
  // that the initial value of the outer variable
  // be reflected in the result of the reduction.
  // Optional when outputType == inputType. If not defined,
  // the following is used:
  //   acumulate(globalAS, output); output = generate(globalAS);
  proc generate(ref output :outputType, globalAS :accumType) :void

}
```

* `inputType` - values being reduced
* `accumType` - per-task accumulation state
* `outputType` - result of the reduction

* "global" - at the outer level - at the same scope as the forall loop or the reduce expression
* "parent", "child" - when a "parent" task creates a "child" sub-task

## Sum reduction

```chpl
[...]

PR #10724, opened 2018-08-09
Cleaner compiler-added shadow variables

vasslitvinov:
This PR makes it easier to change the interface calls
for reduce intents and, more generally, the code to be generated
before/after a forall loop and at task startup/teardown.
The generated C code should remain unaffected.

Background: when a ForallStmt is lowered, it is replaced with copies of
ForallStmt::loopBody() and ShadowVarSymbol::initBlock()/deinitBlock().
When copying, each shadow variable is replaced with its "live"
counterpart. For example, a TFI_TASK_PRIVATE shadow variable
is replaced with the VarSymbol* created for the task-function
FnSymbol being processed. As another example, at the top level
a TFI_REDUCE_OP shadow variable is replaced by the global reduce op.

This change extendes the set of compiler-added ShadowVarSymbols.
The role of each ShadowVarSymbol is clarified, making it easy
to create the SymbolMap to be used when copying the loopBody and
initBlocks/deinitBlocks.

This also changes how the compiler generates the pre/post-ForallStmt
code to set up the global accumulation state (AS) for a reduce intent.
Now it is independent of the initBlock/deinitBlocks. This will be
particularly helpful when we switch to the new UDRI #9879.

The outerVarSym() use is clarified as well:

* For a user-defined shadow variable - TFI_IN, TFI_CONST_IN, TFI_REF, TFI_CONST_REF, TFI_REDUCE - outerVarSym() points to the corresponding user outer Symbol.

* For a TFI_TASK_PRIVATE, outerVarSym() is NULL - this shadow variable does not have an outer variable, by definition.

* For a compiler "parent" helper - TFI_IN_PARENT, TFI_REDUCE_PARENT_AS, TFI_REDUCE_PARENT_OP - outerVarSym() points to the corresponding top-level aka "global" variable.

* For a TFI_REDUCE_OP, outerVarSym() is NULL - it is unused.

Adjust cleanupVoidVarsAndFields() to handle the case where the
accumulation state is void. This happens with the current
implementation of partial reductions, as tested by:
reductions/partial/DR-with-shaped-exprs
reductions/partial/DRtest
npb/cg/bradc/cg-sparse-partred

Move newExtraFormal() and argIntentForForallIntent() closer to their
since place of use; simplify and update them.

While there, change a few functions from extern to 'static' in
implementForallIntents.cpp.

Enact "nicknames" for ShadowVarSymbol* variables within the compiler
For example, "SI" is a ShadowVarSymbol* for an in/const in-intent
shadow variable, "RP" is for a Reduce oP, PRP is for a Parent RP, etc.
The nicknames are documented in the ForallIntentTag enum definition.

Introduce expandShadowVarForall(). It helps set up the new shadow variables
that are added to a ForallStmt within the parallel iterator.
While such a ForallStmt will undergo (later) its own lowering, too,
the difference is that the newly-added shadow variables reflect
forall intents from a different ForallStmt. Forall intents for the
same ForallStmt are handled in expandShadowVarTopLevel().

Testing: linux64, gasnet - multilocale tests, valgrind, memleaks.


compiler/resolution/implementForallIntents.cpp
```diff
@@ -1945,30 +1907,88 @@ static bool needsShadowVar(ForallStmt* fs, BlockStmt* block, Symbol* sym) {
...
   destBlock->insertAtTail(new CallExpr(PRIM_MOVE,
                                                   initTemp, srcExpr));
   insertInitialization(destBlock, destVar, initTemp);
 }
 
-static void insertDeinitialization(ShadowVarSymbol* destVar) {
+// unused
+//static void insertInitialization(ShadowVarSymbol* destVar, Expr* srcExpr) {
+//  insertInitialization(destVar->initBlock(), destVar, srcExpr);
+//}
```
mppf: Can we just remove it?

PR #10726, opened 2018-08-09
AST dumps indicate varargs

mppf:
In debugging issues with varargs I was surprised
that the `--log` output did not identify the `...` argument.
This adjusts it to do so.

Earlier version passed full local testing.


daviditen:
Could you move the printing of "..." down a few lines so it prints after the symbol name and type instead of before?  That way it would match chapel syntax.
```
function foo[176530](arg ... a[176527]:_any[170]?)
```
vs.
```
function foo[176530](arg a[176527]:_any[170]? ...)
```

mppf:
@daviditen - done, look OK now?

daviditen:
Yes, thanks!

PR #10727, opened 2018-08-09
Added polly_array_index call to the DefaultRectangularArr class

ysahil97:
Summary:

- Export the process of index computation in `getDataIndex()` to
  `polly_array_index()`, helping Chapel to communicate the array
  dimension sizes to Polly.

- Used Individual Dimension sizes instead of pre-computed block
  values for index computation. This aligns with Polly's need to
  know the array dimension sizes.

  * Introduced `SizesPerDim` field in `DefaultRectangularArr` class

- Removed the usage of `factoredOffs` in `getDataIndex()` and instead
  subtracted individual offset values from indices.

We have assumed here arrays with unit strides for Chapel-Polly integration.

@mppf Would you like to have a look over this pull request?

- [x] full local testing
Reviewed by @mppf and @benharsh .

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -36,6 +36,39 @@ module DefaultRectangular {
   config param defaultDoRADOpt = true;
   config param defaultDisableLazyRADOpt = false;
   config param earlyShiftData = true;
+  config param usePollyArrayIndex = false;
+
+  // A function which help to compute the final index
+  // to be used for DefaultRectangularArr access. This function
+  // helps Polly to effectively communicate the array dimension
+  // sizes and the index subscripts to Polly.
+  pragma "lineno ok"
+  pragma "llvm readnone"
+  proc polly_array_index(arguments:int ...):int {
+    param rank = (arguments.size - 1) / 2;
+    param blkStart = 2;
+    param blkEnd = 2 + rank - 1;
+    param indStart = blkEnd + 1;
+    param indEnd = indStart + rank - 1;
+    var offset = arguments(1);
+    var blk:rank*int;
+    var ind:rank*int;
+
+    blk(rank) = 1;
+    for param i in 1..(rank-1) by -1 do
+      blk(i) = blk(i+1) * arguments(blkStart+i);
+
+    for param j in 1..rank {
+      ind(j) = arguments(indStart+j-1);
+    }
+
+    var ret:int = offset;
+    for param i in 1..rank {
+      ret += ind(i) * blk(i);
+    }
+
+    return ret;
+  }
 
   class DefaultDist: BaseDist {
     override proc dsiNewRectangularDom(param rank: int, type idxType, param stridable: bool, inds) {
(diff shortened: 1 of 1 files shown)


modules/internal/DefaultRectangular.chpl
```diff
@@ -1124,15 +1159,26 @@ module DefaultRectangular {
         if (rank == 1 && wantShiftedIndex) {
           return chpl__idxToInt(ind(1));
         } else {
-          var sum = 0:intIdxType;
+          //var sum = 0:intIdxType;
```
mppf: Let's make this change controlled by a `config param usePollyArrayIndex = false`. That will leave it off by default (which is good for you, since then you don't have to ensure it is OK for benchmark performance when Polly is not used). We can turn it on for now with `chpl program.chpl -susePollyArrayIndex=true` and in the future when polly integration is enabled.

modules/internal/DefaultRectangular.chpl
```diff
@@ -916,6 +946,7 @@ module DefaultRectangular {
                                            stridable=stridable);
     var off: rank*idxType;
     var blk: rank*chpl__idxTypeToIntIdxType(idxType);
+    var SizesPerDim: rank*chpl__idxTypeToIntIdxType(idxType);
```
mppf: I'd expect it to start with a lower case letter, like `sizesPerDim`. I'd also like to know more about why adding this field is necessary. Is it mostly the same as blk? What if you just accessed dom.dsiDim(dim).length when you needed this information?

modules/internal/DefaultRectangular.chpl
```diff
@@ -37,6 +37,36 @@ module DefaultRectangular {
   config param defaultDisableLazyRADOpt = false;
   config param earlyShiftData = true;
 
+  pragma "lineno ok"
```
mppf: I think we need a comment above this function describing its purpose

modules/internal/DefaultRectangular.chpl
ysahil97: From what I had discovered, polly mainly uses the row-major format for index computation(i.e for Arr[1..4][1..5][1..6], Arr[i][j][k] would correspond to the expression `6*(5*(i) + j) + k`. Since the blk values are precomputed coefficients of the indices in the index expression, it is difficult for Polly to allocate correct array sizes in the case of GPUCodegen, once the blk values are used instead of array dimension sizes.

ysahil97: As far as the last question is concerned, I didn't want to generate extra GEP's and load instructions just for accessing directly from `dom`. Hence, I had precomputed these values in `postinit()` which can be used for further reference.

mppf:
@benharsh - could you have a look at this PR supporting the Chapel+Polly GSoC effort? thanks

benharsh:
I think this is probably OK for now. Eventually there should be tests for different index types (e.g. int(8)).

Michael: do you think it would be worth trying to make ``sizesPerDim`` a ``void`` if the param is false?

mppf:
> Michael: do you think it would be worth trying to make sizesPerDim a void if the param is false?

I'm not sure either way. If it doesn't matter for performance, it would be unnecessary complication. If it does, we'll find out in perf testing. So I'd favor doing that only if we observe it causing an issue.

PR #10731, opened 2018-08-10
Improve override errors for inherited iterators

bradcray:
This does a few things:

1) Updates tests to use `override` for inherited iterator methods when needed
2) Updates compiler-generated methods implementing iterators (e.g., zip1(), advance(), etc.) to inherit `override` from the original iterator when specified
3) Updates generation of `override` warnings/errors to squash them for compiler-generated methods unless `--devel` is on.  An implication of this is that step 2 isn't technically necessary for user compilations; but this still seemed like good practice for consistency and makes --devel compilations cleaner.
4) Adds a utility function FnSymbol::isCompilerGenerated() that answers the question since many compiler-generated functions do not have FLAG_COMPILER_GENERATED (though I've only checked for one additional case here).  This will also help with my "resolve only concrete functions effort" where I found many cases like this where compiler-generated routines were not flagged as such (but had other things one could check for).  As this function evolves we should (a) add more flags to it as we find them and ultimately (b) add FLAG_COMPILER_GENERATED to such cases when we create them.

### DIFF
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -314,7 +314,14 @@ static bool possibleSignatureMatch(FnSymbol* fn, FnSymbol* gn) {
 
 static bool checkOverrides(FnSymbol* fn) {
   ModuleSymbol* parentMod = fn->getModule();
-  return (fOverrideChecking || (parentMod && parentMod->modTag != MOD_USER));
+  // check overrides if any of these are true...
+          // (a) the flag is on and the function is not compiler-generated
+  return ((fOverrideChecking && !fn->isCompilerGenerated()) ||
+          // (b) developer is on (avoids printing developer cases to users)
+          developer == true ||
+          // (c) the function is in the modules/ hierarchy (which we manage
+          //     and want to keep clean)
+          (parentMod && parentMod->modTag != MOD_USER));
 }
 
 static void resolveOverride(FnSymbol* pfn, FnSymbol* cfn) {
@@ -966,8 +973,7 @@ static void checkMethodsOverride() {
             }
           }
 
-          if (okKnown && ok == false &&
-              (developer == true || fn->isUserDefined())) {
+          if (okKnown && ok == false) {
             if (fn->hasFlag(FLAG_OVERRIDE)) {
               USR_FATAL_CONT(fn, "%s.%s override keyword present but no "
                                   "superclass method matches signature "
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -966,7 +966,8 @@ static void checkMethodsOverride() {
             }
           }
 
-          if (okKnown && ok == false) {
+          if (okKnown && ok == false &&
+              (developer == true || fn->isUserDefined())) {
             if (fn->hasFlag(FLAG_OVERRIDE)) {
               USR_FATAL_CONT(fn, "%s.%s override keyword present but no "
                                   "superclass method matches signature "
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -314,13 +314,17 @@ static bool possibleSignatureMatch(FnSymbol* fn, FnSymbol* gn) {
 
 static bool checkOverrides(FnSymbol* fn) {
   ModuleSymbol* parentMod = fn->getModule();
-  // check overrides if any of these are true...
-          // (a) the flag is on and the function is not compiler-generated
-  return ((fOverrideChecking && !fn->isCompilerGenerated()) ||
-          // (b) developer is on (avoids printing developer cases to users)
-          developer == true ||
-          // (c) the function is in the modules/ hierarchy (which we manage
-          //     and want to keep clean)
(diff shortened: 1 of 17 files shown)


bradcray:
@mppf: For your consideration.  It seems likely I'll need you to shepherd this in if you approve.

compiler/resolution/virtualDispatch.cpp
```diff
@@ -966,7 +966,8 @@ static void checkMethodsOverride() {
             }
           }
 
-          if (okKnown && ok == false) {
+          if (okKnown && ok == false &&
+              (developer == true || fn->isUserDefined())) {
```
mppf: This turns on override checking more generally, for developers, right? Are we ready for that? Should we be concerned developers will be surprised by override errors during development?

mppf: I think I understand now - it's filtering out override checking for compiler-generated functions unless --developer. Doesn't turn it on more generally.

mppf: (or maybe in the conditional calling checkOverrides)

bradcray:
@mppf: See the two commits I just pushed to respond to your comments.  I also refreshed the description which bore the signs of being written when I should've been sleeping.  Re-running testing now.

compiler/resolution/virtualDispatch.cpp
```diff
@@ -314,7 +314,14 @@ static bool possibleSignatureMatch(FnSymbol* fn, FnSymbol* gn) {
 
 static bool checkOverrides(FnSymbol* fn) {
   ModuleSymbol* parentMod = fn->getModule();
-  return (fOverrideChecking || (parentMod && parentMod->modTag != MOD_USER));
+  // check overrides if any of these are true...
+          // (a) the flag is on and the function is not compiler-generated
+  return ((fOverrideChecking && !fn->isCompilerGenerated()) ||
+          // (b) developer is on (avoids printing developer cases to users)
+          developer == true ||
```
mppf: this effectively turns on --override-checking when --devel is on. While that might be a good move, it might not be what you are hoping for. Anyway it's not what this PR used to do.

bradcray: Hmmm... you're right.  You'd think I'd be a little better with boolean logic at this point in my career...  So let's see.  The intention is that I was wanting to squash errors for compiler-generated functions when developer was off.  How does this look:

```chapel
  // check overrides for a given function if any of these are true...           
          // (1) the flag is on and either                                      
          //     (a) the function is not compiler-generated or                  
      	  //     (b) --devel is on                                              
      	  //  or...                                                             
  return ((fOverrideChecking &&
           (!fn->isCompilerGenerated() ||
            (fn->isCompilerGenerated() && developer == true)) ||
          // (2) the function is in the modules/ hierarchy (which we manage     
          //     and want to keep clean); oncer override checking is on by      
      	  //     default, we'd likely only check these when --devel is on?      
          (parentMod && parentMod->modTag != MOD_USER));
```

mppf: LGTM

bradcray: Hmm, this new logic causes a single failure for `library/packages/Collection/CollectionCounter (compopts: 2)` (and maybe the old did too, though I thought I got a clean test run last night...?).  It seems that the compiler has expanded the `these()` iterator of the child class into its component methods but not the `these()` iterator of the parent class for whatever reason.  And then since this is a package module, it's doing the checking based on case (2) above.

mppf: IIRC we're not doing dynamic dispatch for iterators at all. I might be wrong (and certainly we ought to eventually). You could certainly try your earlier version to see if moving the condition caused the error.

bradcray: I think it makes sense that moving the conditional caused the behavior change because before we were squashing errors on compiler-generated functions for user compilations, and now we've changed the logic as to whether or not we do checking at all.  So there's effectively a bad overlap with the previous "let's always do checking on the modules/ hierarchy" change.

My intuition is to only check the modules hierarchy if ``--devel`` is on.  This will have the downside of permitting backslides in modules/ in the nightly testing (since ``--devel`` is off there) but I think is more indicative of where we eventually want to be and will take care of this problem.  But I'm also open to different approaches.  I think it's mostly a question at this point of what state I take it too before handing it over to you (where one answer could be that I hand it over now).  I don't intend to merge this today in any case (though you could).

bradcray: Hi @mppf: Looks like I'm not going to find time to work on this more before leaving so am going to let you do with this what you will (where I've given my thoughts just above but am definitely open to others, such as not reporting errors for compiler-generated functions unless devel is on even if they're in the modules/ directory).

I was going to hand off a log to you of the latest test runs, but I think things will change more (for the better) after this and your generic subclass bug fix go in s.t. my current results won't be terribly useful.  I've been generating logs by changing all USR_WARN()s related to overrides into USR_FATAL_CONTs (to fail faster) and by hard-coding the flag to true in driver.cpp.

Overrides conversions have been fun!  Wish I could've gotten further!

mppf:
@bradcray - I created PR #10808 to start with your work & move it forward.

PR #10734, opened 2018-08-10
Fix typos

cassella:
Fix typos in modules/internal/String.chpl and visibleFunctions.cpp

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -2009,7 +2009,7 @@ module String {
     return cpi._cpindex;
   }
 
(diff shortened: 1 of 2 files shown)


modules/internal/String.chpl
```diff
@@ -2009,7 +2009,7 @@ module String {
     return cpi._cpindex;
   }
 
-  // Cast from in to codePointIndex
+  // Cast from int to codePointIndex
```
ben-albrecht: How in the world did you catch this one?

PR #10738, opened 2018-08-10
Fix override checking for generic children of concrete

### LINKED ISSUE #10707: `override` checking complains incorrectly for generic child class

mppf:
Resolves #10707

If the receiver is generic, we were getting override errors no matter
what, and it tried to instantiate with the parent function details, but
that doesn't help since the receiver type can't come from the parent
function - it can only come from the call site.

Fix it by explicitly handling the case in which the receiver type is
generic.

Add tests for this pattern and un-future the test added in #10707.

- [x] full local testing


mppf:
@bradcray - the core of this is about 1 line changed and in code you're modifying in another PR. Have at reviewing it today if you are able? thanks.

bradcray:
This looks good to me, thanks!

PR #10740, opened 2018-08-10
Refactor and improve error messages in InitNormalize

benharsh:
The goal of this PR is to simplify the implementation of error messages in InitNormalize and to check for more errors without halting the compiler. The generated AST after this PR is the same as before.

The first two commits condense the parts of InitNormalize that handle various initialization patterns (concrete vs. generic, inference vs. type-expr) into a new function ``InitNormalize::initializeField``. This new function will be responsible for running error-checking on the generated AST from the various helper functions.

In the following two commits the helper functions are stripped of their error-checking capability (isFieldAccessible, updateFieldsMember) and simply return an expression to be processed. The updateFieldsMember function was also responsible for some AST transformation, which will be handled by the upcoming commits.

Both 'isFieldAccessible' and 'updateFIeldsMember' contained code that basically tried to walk the tree looking for errors. The same was done for the 'processThisUses' error-checking function. The rest of the commits create an AST visitor to replaces the three aforementioned functions. A number of tests have their .good files updated now that the error messages are unified, as before there were arbitrary differences between the three functions.

Testing:
- [x] local + futures
- [x] gasnet + futures


benharsh:
@lydia-duncan : Would you mind reviewing? I would recommend going commit by commit.

PR #10745, opened 2018-08-10
Munge name of polly_array_index() according to number of dimensions

ysahil97:
This compiler modification helps to append the number of dimensions to the name of `polly_array_index()` function. 

This change helps Polly to identify the correct function for its use among a set of functions used for access to arrays of different dimensions.

@mppf, would you like to have a look at this?


compiler/codegen/codegen.cpp
```diff
@@ -161,6 +161,14 @@ static void legalizeName(Symbol* sym) {
   {
     sym->cname = astr("chpl__", sym->cname);
   }
+
+  // Append number of array dimensions to polly_array_index
+  // It helps Polly Optimizer to select the correct function
+  if (strcmp("polly_array_index",sym->cname) == 0){
```
mppf: I think this should use sym->name (vs sym->cname) in the strcmp.

ysahil97: Done

PR #10748, opened 2018-08-10
Add use of lscpu to prediffs for ARM compatibility

dmk42:
Two prediffs that parse `/proc/cpuinfo` get incorrect answers on ARM, where the information in that file is sparse.  Add support for `lscpu` when present, to get more accurate information.

Running `lscpu 2>/dev/null` works regardless of whether `lscpu` is present on the system, without having to give a full path name (in case `lscpu` lives at different locations on different systems).


dmk42:
@gbtitus - Would you please take a post-commit look at this?

gbtitus:
Looks good!  I started to write a note about the `\+` RE not working on MacOS (see #10084) and then realized the logic flow doesn't run `sed` on MacOS.  So all is well.

Issue #10729, opened 2018-08-09
here.runningTasks() nondeterministic


vasslitvinov:
Situation: two forall loops (over a Block domain) next to each other. The second loop queries here.runningTasks() to determine how many tasks to spawn on the current locale.

Issue: those runningTasks() queries may occasionally report a number >1. Even though there are actually no other running tasks at the moment.

This could be a performance problem due to under-utilization of CPU.

We believe this issue caused the failure in gasnet.darwin nightly testing on 7/30/2018 in the test `parallel/forall/task-private-vars/block` .  We are updating this test to pass dataParIgnoreRunningTasks=true so it does not query runningTasks() and so avoids nondeterminism.

A simple reproducer is here. When I run it on 3 locales, it prints a >1 number a couple of times out of 100 runs on a linux64 desktop.

```chpl
// inspired by:
//   test/parallel/forall/task-private-vars/block.chpl

use BlockDist;

config const numMessages = 36;
config const dptpl = 3;
config const ie = false;

const MessageDom = {1..numMessages};
const MessageSpace = MessageDom dmapped Block(MessageDom,
                                              dataParTasksPerLocale = dptpl,
                                              dataParIgnoreRunningTasks = ie);
forall msg in MessageSpace do
  write("."); // whatever
writeln();

coforall loc in Locales do
  on loc do
    if here.runningTasks() > 1 then
      writeln("got issue");
```

vasslitvinov:
@ronawho - good point. I adjusted the test (see the main issue) and am still seeing >1 values.

ronawho:
Noooo, another task counting bug. This looks real, but I don't expect that it impacts real applications. Today we generate on-stmts wrappers as something like:

```c
chpl_taskRunningCntInc();
on_fn(...);
_downEndCount(endcount_chpl, nil);
chpl_taskRunningCntDec();
```

Where the issue is that we decrement the remote end count before decrementing the local running task count so the coforall sync can finish and new remote tasks can be started before the previous ones are completely finished. In practice I think this will only happen for local oversubscription, but I think we can fix this by doing the decrement right before the downEndCount call.

The following program fails within a dozen trials on master:
```chpl
for t in 1..50000 {
  coforall loc in Locales do on loc {
    const rt = here.runningTasks();
    if rt > 1 {
      writeln("Error trial "+t+" on locale "+loc.id+" has "+rt+" tasks");
      exit(1);
    }
  }
}
```

And passes with the following patch:

```diff 
[...]

PR #10750, opened 2018-08-10
Improve our task counting for remote tasks

ronawho:
For code like `coforall loc in Locales do on loc` we used to generate the
on-stmts wrapper functions as something like:

```c
void wrapon_fn() {
  chpl_taskRunningCntInc();     // increment local running task counter
  on_fn(...);                   // actual on function
  _downEndCount(endcount, nil); // decrement remote end count
  chpl_taskRunningCntDec();     // decrement local running task counter
}
```

Where the issue was that we were decrementing the remote end count before
decrementing the local running task count. This allowed the coforall join to
finish and new remote tasks could be started before the previous ones
decremented the running task count. In practice, I think this would only happen
for local oversubscribed testing and not for real distributed applications.

This PR changes things so that we insert the task counting adjustments directly
before and after the actual on fn call (previously we inserted them at the
beginning and epilogue of the wrapper function.) Technically the actual task
will still be running, but it's about to end so for the purposes of task
counting I think it's the right call to decrement the local counter before we
notify the coforall sync that we're done.

We now generate something like:
```c
chpl_taskRunningCntInc();     // increment local running task counter
on_fn(...);                   // actual on function
chpl_taskRunningCntDec();     // decrement local running task counter
_downEndCount(endcount, nil); // decrement remote end count
```

Which allows the included repeated coforall test to pass (I ran 20,000 trials)

Closes https://github.com/chapel-lang/chapel/issues/10729


ronawho:
Passed full gasnet testing, 10,000 trials of the entire parallel/taskCount/runningTaskCount directory, and 50,000 trials of the added repeatedCoforall test.

PR #10752, opened 2018-08-10
Fix DateTime leak and add override intent checking

### LINKED ISSUE #7894: Leaks from DateTime standard module
### Summary of Problem
Usage of the DateTime module can lead to leaking ``TZInfo``-inheriting classes. Existing tests allocate a ``TZInfo`` and pass it to some ``DateTime`` method or initializer. Should ``DateTime`` records be responsible for deleting these classes? What if the record is copied?

The following tests leak:
- [ ] [testDatetimeTZ.chpl](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testDatetimeTZ.chpl)
- [ ] [testTimezone.chpl](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testTimezone.chpl)
- [ ] [testTimezoneConversions.chpl](https://github.com/chapel-lang/chapel/blob/master/test/library/standard/DateTime/testTimezoneConversions.chpl)

mppf:
Resolves #7894.

In investigating #7894, I eventually learned that the memory leak is due
to an inconsistent intent on an overridden method.

The issue is that fromutc had `in` intent in the TZInfo class but was
overridden to use default intent. That confused the compiler, since the
`in` intent now happens at the call site. The compiler ran the copy at
the call site (per the `in` intent) but the virtual method chosen didn't
delete the formal temp, leading to the memory leak.

This PR fixes that problem by making TZInfo.fromutc use the blank intent.

Then, it adds a compiler error for overriding when the intent doesn't
match and adds an ~~~exhausting~~~ exhaustive test of overridden intents.
Note that the error might be that an override isn't found (with
--override-checking) or an error from the new checking, depending on what
the compiler considered to be a "match".

- [x] full local testing

### DIFF
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -330,11 +330,12 @@ static void checkIntentsMatch(FnSymbol* pfn, FnSymbol* cfn) {
     ArgSymbol* ca = cfn->getFormal(i);
 
     if (pa->originalIntent != ca->originalIntent) {
-      USR_FATAL_CONT(cfn, "%s.%s conflicting intent for argument '%s'",
+      USR_FATAL_CONT(cfn, "%s.%s conflicting intent for argument '%s' "
+                          "in overriding method",
                            ct->symbol->name, cfn->name, ca->name);
-      USR_FATAL_CONT(pa, "this function uses %s",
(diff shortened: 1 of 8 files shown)


daviditen:
Sounds good, thanks!

PR #10758, opened 2018-08-12
Preserve the shape of for-expressions

vasslitvinov:
This is the counterpart of #9994 for (serial) for-expressions.

Next step: enable shape preservation for both for- and forall-
expressions when the shape is a range. The challenge there is
that there is no domain object that is already created. So
a new domain needs to be created from the range.

IMPLEMENTATION NOTES

In addition to the _shape_ field added in #9994, each _iteratorRecord
that represents a for-expression also gets the `_fromForExpr_` method.
Which is queried in the shape-ful chpl__initCopy() to determine whether
the being-copied-from _iteratorRecord should be traversed in a for-
vs. forall-loop.

To propagate this within the compiler, PRIM_ITERATOR_RECORD_SET_SHAPE
calls now have a argument telling whether the IR is coming from
a for-expressions.

While there, removed some now-obsolete comments in ChapelArray.chpl.

TEST CHANGES

* knucleotide stores a for-expression in an array 'arr' then sorts it.
This used to work because the array was created in a "shapeless"
manner, i.e. 1-D default-rectangular. Now the array is associative
because freqs.domain - the domain the for-expression iterates over -
is associative. Associative arrays currently cannot be sorted
with quickSort(). So instead iterate over arr.sorted().
The sorted() iterator creates another temp array. This is fine
because 'arr' is very small.

* The submitted knucleotide is now .notest because of the above.

* AverageFiles.chpl started preserving the shape of a **promoted**
expression. (Good.) I am surprised because it was is not the goal
of this PR. In the following simplification of the test, `S` used
to be shapeless, now has the domain D:
```chpl
var image: [D] real;
var S = (image / 45) -  image;
```


modules/internal/ChapelIteratorSupport.chpl
```diff
@@ -120,6 +122,21 @@ module ChapelIteratorSupport {
     else
       return false;
   }
+  pragma "no doc"
+  inline proc chpl_iteratorFromForExpr(ir: _iteratorRecord) param {
+    use Reflection;
+    if canResolveMethod(ir, "_fromForExpr_") then
```
benharsh: Why not always generate "_fromForExpr_" , even if it returns false?

vasslitvinov:
> Why not always generate "fromForExpr" , even if it returns false?

I verified that in some cases we do not invoke setIteratorRecordShape() on an iterator record. So if we want to define fromForExpr on those IRs, we need to find when to do so. I would like to leave that outside the scope of this PR.

One example where this is needed is promotion over ranges, ex.:
```chpl
var A = someScalarFunction(1..4);
```
and probably when promoting over a shape-less iterator, ex. instead of `1..4` above. Affected tests include `functions/deitz/iterators/temps/test_iterator_temp9.chpl` and a few mason tests.

Just in case, here is the change that I tried and that had the above problems:
```diff
diff --git compiler/AST/iterator.cpp compiler/AST/iterator.cpp
--- compiler/AST/iterator.cpp
+++ compiler/AST/iterator.cpp
@@ -117,7 +117,7 @@
 //
 
 // add "proc ir._fromForExpr_ param return true;"
-static void addIteratorFromForExpr(Expr* ref, Symbol* ir) {
+static void addIteratorFromForExpr(Expr* ref, Symbol* ir, bool fromForExpr) {
   SET_LINENO(ref);
   FnSymbol* fn = new FnSymbol("_fromForExpr_");
   fn->addFlag(FLAG_COMPILER_GENERATED);
@@ -134,7 +134,7 @@
 
   fn->insertFormalAtTail(mtArg);
   fn->insertFormalAtTail(irArg);
-  fn->insertAtTail("'return'(%S)", gTrue);
+  fn->insertAtTail("'return'(%S)",fromForExpr ? gTrue : gFalse);
 
   // Ideally, want to put the DefExpr(fn) next to ir->type->symbol->defPoint.
   // Using theProgram->block instead sidesteps a visibility issue
@@ -145,9 +145,9 @@
   resolveFunction(fn);
 }
 
-// return the result of "chpl_iteratorFromForExpr(ir)"
+// return the result of ir._fromForExpr_
 bool checkIteratorFromForExpr(Expr* ref, Symbol* shape) {
-  CallExpr* checkCall = new CallExpr("chpl_iteratorFromForExpr", shape);
+  CallExpr* checkCall = new CallExpr("_fromForExpr_", gMethodToken, shape);
   BlockStmt* holder = new BlockStmt(BLOCK_SCOPELESS);
   holder->insertAtTail(checkCall);
   ref->insertAfter(holder);
@@ -159,7 +159,7 @@
   
   Symbol* checkResult = checkFn->getReturnSymbol();
 
-  // chpl_iteratorFromForExpr() is a param boolean function
+  // _fromForExpr_ should be a param boolean function
   return getSymbolImmediate(checkResult)->bool_value();
 }
 
@@ -194,8 +194,7 @@
     // This sidesteps the visibility issue in the presence of nested
     // LoopExprs. Ex. test/expressions/loop-expr/scoping.chpl
     theProgram->block->insertAtTail(accessor->defPoint->remove());
-    if (fromForExpr)
-      addIteratorFromForExpr(ref, ir);
+    addIteratorFromForExpr(ref, ir, fromForExpr);
   } else {
     INT_ASSERT(field->type == value->type);
   }
diff --git modules/internal/ChapelArray.chpl modules/internal/ChapelArray.chpl
--- modules/internal/ChapelArray.chpl
+++ modules/internal/ChapelArray.chpl
@@ -4192,7 +4192,7 @@
     // in the other case. Ex. users/vass/km/array-of-records-crash-1.*
 
     var result: [shape] iteratorToArrayElementType(ir.type);
-    if chpl_iteratorFromForExpr(ir) {
+    if ir._fromForExpr_ {
       for (r, src) in zip(result, ir) do
         r = src;
     } else {
diff --git modules/internal/ChapelIteratorSupport.chpl modules/internal/ChapelIteratorSupport.chpl
--- modules/internal/ChapelIteratorSupport.chpl
+++ modules/internal/ChapelIteratorSupport.chpl
@@ -123,20 +123,9 @@
       return false;
   }
   pragma "no doc"
-  inline proc chpl_iteratorFromForExpr(ir: _iteratorRecord) param {
-    use Reflection;
-    if canResolveMethod(ir, "_fromForExpr_") then
-      return ir._fromForExpr_;
-    else
-      return false;
-  }
+  proc _domain._fromForExpr_ param return false;
   pragma "no doc"
-  inline proc chpl_iteratorFromForExpr(arg) param {
-    // non-iterator-record cases are always parallel
-    // Todo: what if it is an array or domain whose domain map
-    // that does not provide parallel iterators?
-    return false;
-  }
+  proc _array._fromForExpr_ param return false;
 
   proc _iteratorRecord.writeThis(f) {
     var first: bool = true;
```

PR #10764, opened 2018-08-13
Add forDomain argument to Distribution class constructors

### LINKED ISSUE #10762: Feature Request: Add a default argument for Cyclic's `startIdx` and Block's `boundingBox`
It feels that when the user must specify these arguments for their respective distributions, it makes it impossible to specify an action like: "here is the distribution, please distribute your data over it". To make matters worse, if the user does not know how the data will be distributed, it is possible to create a severe load imbalance.

Let's say that the user specifies that I should allocate data over a `Block` distribution, but does not know the `boundingBox`, what should they use as the argument? No matter what value they choose (barring the extreme cases where they try to allocate some really large out-of-bounds `boundingBox={min(int)..max(int)}` (which won't work since it [overflows](https://tio.run/##S85ILEjN@f@/tDhVwSknPznbJbO4xJqLq7wosyQ1J0@jOjczTyMzr0RTTy83sQLMqtW05ipLLFJIyc9VsFXwyU9OzEkNLkhMTlVIyU0sKEhNgRiET2tiUZGClUI00IRYBaC4Ndw6oISm9f//AA), but it will work if you divide both `min(int)` and `max(int)` by two), or `startIdx=min(int)` for `Cyclic` distribution. I'm all for keeping the ability to specify both `boundingBox` and `startIdx` for the sake of customization, but it should default to a much larger and acceptable default.

What is worse is what happens if the user fails to provide an adequate `boundingBox` or `startIdx`, which is a severe load balance. What happens if the data is distributed over negative indices and they set the `startIdx` at 0 or 1? What if the user lazily gave a `boundingBox={1..1}` due to not knowing what the argument even did and ends up with all data being distributed over a single locale?

I think the user should be able to allocate/create distributed domains with as little thought as possible for the simple use-cases...

```chpl
var cyclicArr : [{1..1000} dmapped Cyclic] int;
var blockArr : [{1..1000} dmapped Block] int;
```

@bradcray Any thoughts on this, as distributed arrays are your specialty?

mppf:
Closes #10762.

This PR adds a version of distribution class initializers accepting a forDomain argument and adjusts the normalizer to pass it.

This allows the long-sought and elusive pattern here to function:
``` chapel
  var D = {0..9};
  var D4 = D dmapped Block;
```

Before this PR, the Block call would have to include a `boundingBox` argument. This PR allows the Block initializer to take it from the domain being `dmapped`.

- [x] full local testing
- [x] full local gasnet testing


mppf:
@benharsh - could you let me know if this general strategy seems OK, before we update all of the distributions? Note that type construction is a bit different - we could try to get a forDomain type constructor working, we could make a separate function to call, or we could make the normalizer detect the type situation and not add the forDomain argument in that case. In any event I think the type situation is less common.

benharsh:
I'd prefer the compiler attempt to invoke an initializer whose first argument is a domain, rather than use a NamedExpr. I would prefer we avoid introducing magic names when possible. This may imply we need to create a second initializer, but perhaps it could simply invoke the existing initializer.

mppf:
@LouisJenkinsCS - no, it doesn't help with new dmap(new Dist) case.

mppf:
@benharsh - it can't simply use the boundingBox field in every case since if that is supplied we don't want to supply it twice. E.g. in `const BlockSpace = Space dmapped Block(boundingBox=Space);`, the new normalizer pattern will call Block(Space, boundingBox=Space) which is nonsense. So I think even Block will need new initializers with forDomain / forDomainType.

mppf:
[Error matching program output for distributions/ferguson/fordomain]
[Error matching compiler output for studies/comd/elegant/arrayOfStructs/CoMD]

mppf:
@benharsh - could you have another look? What do you think about the NamedExpr possibility with the additional information? I'll also want to know if you can run with this next week...

mppf:
I think this is ready to merge and @benharsh will shepherd it next week.

bradcray:
I'd like a chance to catch up with this PR before it's merged.

bradcray:
I'm trying to get my head around this in short order, but not completely successfully.  From what I'm seeing, it appears that this is a language change (in that it involves the compiler), as well as a change to the DSI interface.  Assuming that's right, I feel surprised that there hasn't been any broader design discussion around it (that I can tell) prior to this PR being considered to be mergeable, nor any documentation changes within the PR itself (which is how I was hoping to learn about the changes it was making).

I believe that in past discussions around providing a capability like this, we've considered approaches that wouldn't require any compiler/language changes.  For example, one approach for Block and Cyclic would be to have a default value for the boundingBox / startIdx argument that would serves as a "deferred evaluation" sentinel value (or to implement the  long-desired "query whether a user has provided an argument" feature), and then fill in those pieces when the first domain is created over the distribution.  That approach had the downside of still requiring other `param` / `type` fields to be explicitly specified for the distribution (if/when they didn't have default values), but didn't require language changes.  That is, a user would have to write:

```chapel
const D = {1..n, 1..n} dmapped Block(rank=2);  // where the `rank=` would be optional
const D2 = {myUint8range} dmapped Block(2, uint(8));  // or I could've used `idxType=`
```

So this approach results in better convenience, but also feels a bit more "magic" since information is propagated invisibly in a way that wouldn't be accessible to user-written code in other contexts.  As I alluded to in issue #10762, this also intuitively feels like it breaks (or changes?) order of evaluation to me in a way, since when seeing `Dom dmapped Dist` my intuition is that the `Dom` and `Dist` expressions are each evaluated first, and independently, before the `dmapped` clause as a whole is (similar to other binary operators or method calls).  Yet here, `Dist`'s evaluation seems to depend on `Dom`'s in a way that's not intuitive.

Another approach we've discussed that wouldn't require language changes is simply providing some library helper functions that make certain patterns more convenient, such as:

```chapel
const D = newBlockDom({1..n});
```

In the approach taken here, I'm unclear what happens for cases in which an argument doesn't have a natural default argument (and haven't had the time to experiment yet this week to answer for myself).  E.g., I can imagine a Block-Cyclic implementation that has a heuristic for figuring out a default block size in each dimension, but I can also imagine one that _requires_ the user to specify the block size(s).  I feel pretty confident that this PR would let me write the former; would it let me write the latter?  (essentially, is the `dmapped Block` essentially a shorthand for `default Block()` and can I add additional arguments into that list without breaking things?)  I'm guessing / hoping so, but unsure based on a quick look.

I could imagine eventually coming around to liking this proposal, but these hesitations, combined with the lack of broad discussion around the proposal and the timing w.r.t. the release make me inclined to suggest we discuss further and push the feature into the next release cycle (in whatever form it takes).

mppf:
> In the approach taken here, I'm unclear what happens for cases in which an argument doesn't have a natural default argument (and haven't had the time to experiment yet this week to answer for myself). E.g., I can imagine a Block-Cyclic implementation that has a heuristic for figuring out a default block size in each dimension, but I can also imagine one that requires the user to specify the block size(s). I feel pretty confident that this PR would let me write the former; would it let me write the latter? (essentially, is the dmapped Block essentially a shorthand for default Block() and can I add additional arguments into that list without breaking things?) I'm guessing / hoping so, but unsure based on a quick look.

This PR just adds an argument to some of the initializer calls implementing a `dmapped` to the compiler. So you can still have an initializer with any number of other required arguments.

> As I alluded to in issue #10762, this also intuitively feels like it breaks (or changes?) order of evaluation to me in a way, since when seeing Dom dmapped Dist my intuition is that the Dom and Dist expressions are each evaluated first, and independently, before the dmapped clause as a whole is (similar to other binary operators or method calls). Yet here, Dist's evaluation seems to depend on Dom's in a way that's not intuitive.

That's an interesting point.

> we've considered approaches that wouldn't require any compiler/language changes.

Because of the way that `D dmapped Block(...)` is supported directly by the compiler (as "syntactic sugar") I'm hesitant to believe this problem can be solved without compiler changes.

I'm not convinced that this change indicates a *language change* other than the implication that the DSI interface changes. The order-of-operations bit above is the only language-y thing about it. I think it's highly unlikely to suprise anybody, anyway.

>  Assuming that's right, I feel surprised that there hasn't been any broader design discussion around it (that I can tell) prior to this PR being considered to be mergeable, nor any documentation changes within the PR itself (which is how I was hoping to learn about the changes it was making).

I wasn't completely clear in my comments, but part of @benharsh taking over this PR was making sure you had a chance to comment on it before merging it. Anyway I've marked it as needing more design review and for after the release.

>  For example, one approach for Block and Cyclic would be to have a default value for the boundingBox / startIdx argument that would serves as a "deferred evaluation" sentinel value (or to implement the long-desired "query whether a user has provided an argument" feature), and then fill in those pieces when the first domain is created over the distribution. 

That's an interesting alternative for sure. I'm not so sure I'd call it less of a language change, though.

mppf:
I'm closing this issue due to discussion with @bradcray.
 * there is an order of evaluation concern
 * there is a concern that putting a domain map in a variable would inhibit this logic, which might be surprising

@bradcray prefers a strategy that computes the bounding box, if it's unspecified, based on the first domain attached to the distribution. That seems possible to implement today for Block if desired. (Would race conditions be an issue?).

Anyway, this approach doesn't seem to be the direction to go.

PR #10663, opened 2018-08-03
build_configs: add setenv-example-3.bash (#9551)

awallace-cray:
A realistic Chapel configuration for Cray-XC systems.
Implements https://github.com/chapel-lang/chapel/issues/10461

PR #10765, opened 2018-08-13
build_configs: GBT post-merge review comments (PR #10663)

awallace-cray:
* setenv-example-3 (for Cray) should not use all the CPUs
  At least, not by default
* setenv-example-3 should set most CHPL_* default values
  Especially REGEXP=re2


awallace-cray:
@gbtitus : I'm testing (slowly) but this looks OK so far.

gbtitus:
Cool.  I built with `CHPL_MAKE_MAX_CPU_COUNT=4` set externally to the script and then tested the configs with the Intel and gnu target compilers and comm=none and ugni, and all 4 of those worked.  (I only tested the slurm launcher, not "none".)

awallace-cray:
My tests are good. Merging this.

PR #10768, opened 2018-08-13
Use unmanaged in comd

mppf:
Updates both CoMD implementations to use unmanaged.
Not sure why earlier efforts to migrate code missed these.

- [x] test/studies/comd passes with gasnet and --warn-unstable
- [x] test/studies/comd passes with default config and --warn-unstable

Reviewed-ish by @benharsh - thanks!


mppf:
@benharsh - could you take a quick look at this? I think it doesn't necessarily need review but I'd appreciate a nod in my general direction. Thanks.

PR #10769, opened 2018-08-13
Add the preprocessing capability to function `readNamedHDF5FilesInto1DArrayInt`

daviditen:
This function was missing the preprocessing capability, but didn't need to be.
Add it in.

Add a test that uses this function with a preprocessor.

Fix a small mistake in one of the example preprocessors in
HDF5Preprocessors.chpl.


daviditen:
Passed a full local Linux64 paratest.

ben-albrecht:
> LGTM

Note this test will need to adjust if/when we reduce the `EXAMPLE_HDF5` size

PR #10773, opened 2018-08-14
Two Locale Model updates

### LINKED ISSUE #10747: Querying information from locale obtained from querying locality of array element causes two round-trips each time
For some reason, there are certain situations where querying the locale of an element and obtaining its id will cause two round-trips each time. Querying `locale.id` appears to jump to the `locale` and acquire the locale, and worse yet it appears that the `locale` itself is a wide pointer (which can be queried without communication). See below MWE and try running with 2 locales. @bradcray 

**MWE**

```chpl
use CyclicDist;
use CommDiagnostics;

config const doVerboseComm = false;
config const doCommDiagnostics = true;
config const numElements = 1024 * 1024;

proc setup() {
  if doVerboseComm then startVerboseComm();
  if doCommDiagnostics then startCommDiagnostics();
}

proc tearDown() {
  if doVerboseComm then stopVerboseComm();
  if doCommDiagnostics {
    stopCommDiagnostics();
    writeln("Comm Diagnostics: ", getCommDiagnostics());
    resetCommDiagnostics();
  }
}

const cyclicSpace = {0..#numElements};
const cyclicDom = cyclicSpace dmapped Cyclic(startIdx=0);
var cyclicArr : [cyclicDom] int;

writeln("Querying via domain distribution...");
setup();
  coforall tid in 1..here.maxTaskPar {
    for i in cyclicSpace {
      var locid = cyclicDom.dist.idxToLocale(i).id;
    }
  }
}
tearDown();

writeln("Querying via array reference...");
setup();
  coforall tid in 1..here.maxTaskPar {
    for i in cyclicSpace {
      var locid = cyclicArr[i].locale.id;
    }
  }
}
tearDown();

writeln("Querying via wide pointer...");
setup();
  coforall tid in 1..here.maxTaskPar {
    for i in cyclicSpace {
      var loc = cyclicDom.dist.idxToLocale(i);
      var locID = chpl_nodeFromLocaleID(__primitive("_wide_get_locale", loc));
      assert(cyclicArr[i].locale == Locales[locID]);
    }
  }
}
tearDown();

inline proc getLocale(dom, idx) {
  var loc = dom.dist.idxToLocale(idx);
  var locID = chpl_nodeFromLocaleID(__primitive("_wide_get_locale", loc));
  
  // Handles cases where we get a locale that is allocated on another locale...
  if locID == here.id then return loc;
  else return Locales[locID];
}

writeln("Querying via wide pointer inside function...");
setup();
  coforall tid in 1..here.maxTaskPar {
    for i in cyclicSpace {
      var locid = getLocale(cyclicDom, i).id;
    }
  }
}
tearDown();

writeln("Querying remote object...");
var obj = new object();
setup();
  coforall tid in 1..here.maxTaskPar {
    for i in cyclicSpace {
      var locid = obj.locale.id;
    }
  }
}
tearDown();     
```

**Output**

```
Querying via domain distribution...
Comm Diagnostics: (get = 46137344, execute_on_nb = 1) (get = 46137432)
Querying via array reference...
Comm Diagnostics: (get = 7, execute_on_nb = 1) (get = 7)
Querying via wide pointer...
Comm Diagnostics: (execute_on_nb = 1) (<no communication>)
Querying via wide pointer inside function...
Comm Diagnostics: (get = 46137344, execute_on_nb = 1) (get = 46137432)
Querying remote object...
Comm Diagnostics: (execute_on_nb = 1) (<no communication>)
```

mppf:
* Adds error for -nl >1 and --local
* Update locale.id to get id from wide `this` ptr to avoid communication
* adds tests for both of these

Resolves #10747 

- [x] full local testing
- [x] full gasnet testing
- [x] hellos work with CHPL_LOCALE_MODEL=numa

Thanks to @LouisJenkinsCS for bringing up the issues and suggesting a fix to the locale.id communication.


modules/internal/ChapelLocale.chpl
```diff
@@ -179,10 +179,10 @@ module ChapelLocale {
       :returns: locale number, in the range ``0..numLocales-1``
       :rtype: int
      */
-    proc id : int return chpl_id();  // just the node part
+    proc id : int return chpl_nodeFromLocaleID(__primitive("_wide_get_locale", this));
```
ronawho: @mppf While working on https://github.com/chapel-lang/chapel/pull/11579 I found myself wondering if I could/should use `_wide_get_node` in the network atomic code, which also made me wonder if it would be appropriate here. I'm not very familiar with the differences between` _wide_get_locale` and` _wide_get_node` though so this could be a silly suggestion.

That also led me to https://github.com/chapel-lang/chapel/commit/70aa131bd8, which looks like it might have been an older form of this optimization, but I'm not sure if that optimization still triggers.

mppf: AFAIK _wide_get_node is there for historical reasons and _wide_get_locale should be used in preference to it in order to make explicit the ability to handle sublocales.

PR #10774, opened 2018-08-14
Fix field reuse in generic type's initializer

### LINKED ISSUE #10606: problem with initializer initializing field from this
``` chapel
class MyClass {
  var i:int;
}

class MyGenericClass {
  var mx;
  var i;
}

proc MyGenericClass.init(x) {
  this.mx = x;
  this.i = this.mx.i;
  // this seems to work: this.i = x.i;
}

var x = new unmanaged MyGenericClass(new unmanaged MyClass(1));
writeln(x);

delete x.mx;
delete x;
```

```
x.chpl:10: In initializer:
x.chpl:12: internal error: CAL0124 chpl version 1.18.0 pre-release (91da8ff5d8)

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see https://chapel-lang.org/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.

```

```
x.chpl:10: In initializer:
x.chpl:12: internal error: the type of the actual argument 'this' is generic [callInfo.cpp:124]
```

benharsh:
This PR fixes a bug where the compiler would encounter an internal error when a field was read after it was initialized, but before the type was fully instantiated. The cause was that normalization was transforming the field access into a method call for the paren-less compiler-generated field accessor method. We cannot currently call methods on partially instantiated/initialized types, so the compiler blew up.

The fix is to catch such patterns early and transform them into simple PRIM_GET_MEMBER calls. This PR adds a new error message that is emitted if the user wrote their own field accessor method and attempts to read the field after it is initialized (before this.complete()).

Resolves #10606

Testing:
- [x] local + futures
- [x] gasnet + futures


benharsh:
@lydia-duncan : this is the follow-on to the last PR you reviewed. Willing to review this one as well?

PR #10775, opened 2018-08-14
Make new SomeClass create a borrow and warn

### LINKED ISSUE #9093: lifetime checking: implement default class new behavior
As a Chapel programmer, I want `new MyClass` to do something reasonable so that I can easily express class creation.

Since we are expecting a release where `new` calls allocating classes must be decorated, this does not break those programs that have already been converted.

Acceptance criteria: `new MyClass` does whatever we have decided to do as demonstrated by tests.

### LINKED ISSUE #10719: new is the same as new borrowed
Make `new MyClass()` do the same thing as `new borrowed MyClass()`

mppf:
This PR makes
``` chapel
new MyClass()
```

the same as
``` chapel
new borrowed MyClass()
```

which is the same as
``` chapel
(new owned MyClass()).borrow()
```
except that the first pattern causes a warning (since it's changing behavior).
I expect the warning will be removed after 1.18.

Resolves #10719.
Resolves #9093.

This PR includes a bunch of test and module updates that previous efforts have
missed for one reason or another. It adds a flag `--legacy-new`, that leaves
`new MyClass` returning a borrow, but doesn't automatically delete it (i.e.
like `(new unmanaged MyClass()):borrowed MyClass`) since that's the mode we've
been living in for a while. It uses this flag in the spec tests as a way to get
them passing until these spec sections are addressed.

- [x] full local testing
- [x] full gasnet futures testing

Future Work:

This PR adds an error message that suggests `--warn-unstable` but that's still
not a user-facing flag. Issue #10792 tracks this TODO.


compiler/resolution/functionResolution.cpp
```diff
@@ -5908,11 +5910,17 @@ static void resolveNewSetupManaged(CallExpr* newExpr, Type*& manager) {
...
+        if (isManagedPtrType(type)) {
           manager = type;
-
-        if (isUnmanagedClassType(type))
+        } else if (isUnmanagedClassType(type)) {
           manager = dtUnmanaged;
+        } else if (isClass(type) && isUndecoratedClassNew(newExpr)) {
+          if (fLegacyNew == false && fDefaultUnmanaged == false) {
+            gdbShouldBreakHere();
+            USR_WARN(newExpr, "this new call changes meaning");
```
benharsh: This message doesn't seem helpful to me. Maybe describe what's missing or what the change is?

PR #10778, opened 2018-08-14
build_configs: setenv-example-3 should skip useless runtime build configs

awallace-cray:
build_configs: setenv-example-3 should skip useless runtime build configs

The setenv callbacks now skip these runtime build configs:
    COMM == none && LAUNCHER != none
    COMM != none && LAUNCHER == none

### DIFF
--- a/util/build_configs/setenv-example-3.bash
+++ b/util/build_configs/setenv-example-3.bash
@@ -97,20 +97,25 @@ if [ -z "$BUILD_CONFIGS_CALLBACK" ]; then
     $cwd/build_configs.py -p $dry_run $verbose -s $cwd/$setenv -l "$project.runtime.log" \
         --target-compiler=$compilers --comm=$comms --launcher=$launchers
 
-        # The following 2x2x2 matrix of Chapel runtime configurations are defined by the
-        # arguments passed to build_configs.py, above. All eight runtimes will be built.
         # NOTE: "--target-compiler" values shown above will be discarded by the setenv callback.
 
-        # CHPL_TARGET_COMPILER  CHPL_COMM   CHPL_LAUNCHER
-        # --------------------  ---------   -------------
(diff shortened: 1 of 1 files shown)


awallace-cray:
@gbtitus : Hi Greg, this change was suggested by your email yesterday. 
If you like it, let me know and I will test it. 

Not Urgent, as far as I'm concerned.  
I do not much care if this goes in, or when. 

Thanks

util/build_configs/setenv-example-3.bash
```diff
@@ -97,20 +97,25 @@ if [ -z "$BUILD_CONFIGS_CALLBACK" ]; then
...
-        # cray-prgenv-intel     none        slurm-srun
-        # cray-prgenv-intel     ugni        none
-        # cray-prgenv-intel     ugni        slurm-srun
+        # The following 2x2x2 matrix of Chapel runtime configurations are defined by the
+        # arguments passed to build_configs.py, above.
+
+        # CHPL_TARGET_COMPILER  CHPL_COMM   CHPL_LAUNCHER   Make Chapel?
+        # --------------------  ---------   -------------   -----------
+        # cray-prgenv-gnu       none        none            Yes
+        # cray-prgenv-gnu       none        slurm-srun      No (skip)
```
gbtitus: comm=none with a Chapel launcher is still worthwhile.  (Chapel provides its own launcher so that users don't have to know the details of how to launch programs on their systems, and this is just as valuable for single-node comm=none programs as it is for multi-node ones.)

The only config that doesn't make sense is comm=ugni with launcher=none; that one is known not to work because the ugni comm layer depends on some settings that are done by the Chapel launchers.

awallace-cray:
This built on jupiter and skipped the expected two configs. That's all I tested, but it seems enough.

Issue #10666, opened 2018-08-03
What should the generated Python module be called?


lydia-duncan:
The tutorial for Cython tends to use a different name for the .a/.so/.h/.pxd combo and the .pyx/.py final python module name.  What should we use as the base name for the .pyx/.py files?  How should the user control it?

Options:

1.  Use the name specified via `-o` as the name for the .pyx/.py file and a different, generated name for the .a/.so/.h/.pxd file
  - Rationale: when creating a Python module, it may be that the user doesn't care as much about the C style library
2. Use a different name for the .pxd file, and have the .pyx/.py file match the name of the .a/.so/.h
  - I need to experiment to see if this is actually valid.  I also need to experiment with the .h having a different name from the .a/.so
3. Allow the user to specify the name for the generated python module, and use that name for the .pyx/.py file.  Flag would potentially be named `--library-python-name`
  - If the flag is not provided, the name defaults to `chpl_libraryName` where libraryName is the name used for the .a/.so.

The .pxd file name will match the header name, since the file will probably get hidden from the user.

I lean towards 2 if possible (with 3's strategy as the way to change the name).  If 2 is not possible, then I lean towards just 3

bradcray:
> The tutorial for Cython tends to use a different name for the .a/.so/.h/.pxd combo and the .pyx/.py final python module name.

Is there a reason they can't all share the same name?  I.e., is it for pedagogical purposes of some kind within the tutorial, or is it a requirement of some sort?

lydia-duncan:
It seems to be just a general best practice.  I can use the same name for the .pxd and .pyx/.py files today without issue.  I suspect Cython probably won't disable that support in the future but they may not test it as thoroughly as the other strategy or something, so it might be better to honor it?  I'm not married to following it, but left to my own devices I probably would.

mppf:
I'd favor using the same name if possible. If you're uncertain about how weird/tested that is, maybe you could ask on a Cython chat channel / mailing list?

lydia-duncan:
[Here's the Cython chat thread](https://groups.google.com/forum/#!topic/cython-users/q25rLV1UBRM).  Basically, the names are supposed to match when the code is entirely implemented in the .pyx/.pxd files.  Because the code is implemented in a library file, the .pxd file and .pyx file should have different names.  But the .pyx can match the library.  So strategy 2 seems like the way to go (with a flag to change the name of the generated Python module)

mppf:
[...]

PR #10787, opened 2018-08-15
Generate pyx file in --library-python compilation

### LINKED ISSUE #10662: Generate .pyx file for Chapel library python compilation
This is the next file that needs to be generated in order to create a Python module from a Chapel library using Cython. It should only be generated when the flag defined in #10442 is thrown during library compilation. We might even want to clean it up afterwards by putting it in a temporary directory like our generated C code (not sure if that will cause problems later, need to experiment).

.pyx files have the format:
```cython
from foo cimport chpl_library_init as library_init, chpl_library_finalize as library_finalize, chpl__init_foo
from foo cimport funcName as chpl_funcName

def chpl_setup():
    cdef char** args = ['finalModuleName']
    library_init(1, args)
    chpl__init_foo(1, 1)

def chpl_cleanup():
    library_finalize()

def funcName(argList):
  # translations for specific argument types
  chpl_funcName(newArgList)
```

The translation for certain types will be slightly more complicated than others.

Test code for this does not need to fully compile (although that is something I will do locally), I can just include it in the .catfiles for now.

lydia-duncan:
Resolves #10662 

This generates the next file needed for Cython compilation of Chapel library files into
a Python-accessible module.  .pyx files contain imports from the .pxd file and present
the functions in a form that can be called from Python.  For some types, this means
performing a translation from the Python argument to a corresponding C version - for
instance, the Python 3 bytes type needs to be copied into a C char*.

The .pyx file we generate also unifies the runtime and module initializations into a single
function on the Python side, for the user's convenience.

Following the discussion on #10666, the .pyx file uses the name of the library (minus
the "lib" in the case where we add it) as the name for the .pyx file, changing the name
of the .pxd file to be "chpl_" plus the header name.

Note: this does not do anything special for arrays yet.  The .pxd file needs to export the chpl_external_array definition so that the .pyx file can import and use it to translate some
known Python types.  I'm going to do that in a separate PR, since this one was
feeling a bit large.

Also note: this doesn't add a flag for controlling the name of the generated Python
module.  This PR was getting large so I figured I should separate that part out (though
this PR already includes changing the name of the .pxd file, so maybe this is the right
place for it).

This change is a little easier to review in "split" mode than in "unified" mode, because
I moved the pxd functionality into a separate function, so the diff got muddled.

Testing: I extended the python tests to also validate the .pyx contents.  Once we have
a testing environment with Cython (and I generate the last file necessary), I will
change these tests to actually build and run a python program.  Also, added a test
of the python module name when the library name is specified via -o.

Passed a full paratest


lydia-duncan:
@mppf - would you review?

compiler/codegen/codegen.cpp
```diff
@@ -2161,6 +2161,18 @@ static void setupDefaultFilenames() {
       strncat(executableFilename, lastSlash,
               sizeof(executableFilename)-strlen(executableFilename)-1);
 
+      if (fLibraryPython) {
+        strncpy(pythonModulename, lastSlash, sizeof(pythonModulename)-1);
```
mppf: while lastSlash describes how it was computed, this value *really* needs a different variable name. I'd recommend e.g.
``` c
     const char* lastSlash = strrchr(mainModFilename, '/');
     const char* filename = NULL;
     if (lastSlash == NULL) {
       filename = mainModFilename;
     } else {
       filename = lastSlash + 1;
    }
```

and replacing your uses of lastSlash with filename.

compiler/codegen/library.cpp
```diff
@@ -243,37 +244,143 @@ static void setupPythonTypeMap() {
...
+static void makePYXFile(std::vector<FnSymbol*> functions);
+
 void codegen_library_python(std::vector<FnSymbol*> functions) {
   if (fLibraryCompile && fLibraryPython) {
     setupPythonTypeMap();
 
-    fileinfo pxd = { NULL, NULL, NULL };
+    strcpy(pxdName, "chpl_");
+    strncat(pxdName, libmodeHeadername,
+            sizeof(libmodeHeadername)-strlen(pxdName)-1);
```
mppf: So the code you're writing has a lot of strcpy / strncat.
I tolerate some of these but in general I'd rather look at code that uses `+` on C++ strings (and then calls `.c_str()`) or uses our own `astr` method to concatenate strings.

compiler/codegen/library.cpp
```diff
@@ -243,37 +244,143 @@ static void setupPythonTypeMap() {
...
+    fprintf(pxd.fptr, "from libc.stdint cimport *\n\n");
 
-      for_vector(FnSymbol, fn, functions) {
-        if (!isFunctionToSkip(fn)) {
-          fn->codegenPXD();
+    fprintf(pxd.fptr, "cdef extern from \"%s.h\":\n", libmodeHeadername);
+
+    for_vector(FnSymbol, fn, functions) {
+      if (!isFunctionToSkip(fn)) {
+        fn->codegenPython(true);
```
mppf: Can we call it codegenPythonPXD?

compiler/codegen/library.cpp
```diff
@@ -243,37 +244,143 @@ static void setupPythonTypeMap() {
...
+  // Initialize the included modules (continuation of chpl_setup definition)
+  for_vector(FnSymbol, fn, moduleInits) {
+    fprintf(outfile, "\t%s(1, 1)\n", fn->cname);
+  }
+  fprintf(outfile, "\n");
+
+  // Shut down the runtime and libraries.  chpl_cleanup should get called when
+  // the exported Chapel code is no longer needed
+  fprintf(outfile, "def chpl_cleanup():\n");
+  fprintf(outfile, "\tchpl_library_finalize()\n");
```
mppf: Just curious, but it looks like chpl_library_init doesn't init modules but chpl_library_finalize does deinit modules. That seems odd. Am I misunderstanding?

compiler/codegen/symbol.cpp
```diff
@@ -877,22 +877,42 @@ GenRet ArgSymbol::codegen() {
...
   // TODO: LLVM stuff
 }
 
+// Some Python type instances need to be translated into C level type instances.
+// Generate code to perform that translation when this is the case
+std::string ArgSymbol::getPythonArgTranslation() {
+  Type* t = getArgSymbolCodegenType(this);
+
+  if (t == dtStringC) {
+    std::string res = "\tcdef char* chpl_";
```
mppf: c_string == const char * not char*

compiler/codegen/symbol.cpp
```diff
@@ -877,22 +877,42 @@ GenRet ArgSymbol::codegen() {
 
 // If there is a known .pxd file translation for this type, use that.
 // Otherwise, use the normal cname
-static std::string getPXDTypeName(Type* type) {
+static std::string getPythonTypeName(Type* type, bool pxd) {
```
mppf: I don't really like this bool pxd. Can you make it an enum instead? That makes it easier to understand the call site.

e.g. 

getPythonTypeName(myType, PYTHON_PXD)

compiler/codegen/symbol.cpp
```diff
@@ -877,22 +877,42 @@ GenRet ArgSymbol::codegen() {
...
     return tNames.first;
+  } else if (!pxd && tNames.second != "") {
+    return tNames.second;
   } else {
     return transformTypeForPointer(type);
   }
 }
 
-std::string ArgSymbol::getPXDType() {
+std::string ArgSymbol::getPythonType(bool pxd) {
```
mppf: bool -> enum pls

compiler/codegen/symbol.cpp
```diff
@@ -1652,7 +1672,7 @@ GenRet FnSymbol::codegen() {
 }
 
 // Supports the creation of a python module with --library-python
-void FnSymbol::codegenPXD() {
+void FnSymbol::codegenPython(bool pxd) {
```
mppf: bool -> enum please
also I'd prefer this just be named codegenPythonPXDPYD e.g. but I can live with codegenPython

compiler/codegen/symbol.cpp
```diff
@@ -1716,6 +1740,73 @@ GenRet FnSymbol::codegenPXDType() {
...
+        count++;
+      }
+
+    } // pyx files do not take void as an argument list, just close the parens
+    header += "):\n";
+    funcCall += ")\n";
+    ret.c = header + argTranslate + funcCall + returnStmt;
+
+  } else {
+    // TODO: LLVM stuff
```
mppf: So these LLVM TODOs are getting a bit ridiculous. Supposing, for the sake of argument, that we were using --llvm to build a library and wanted the compiler to generate these Cython interop files. Then info->cfile would be set to the Cython file, right? Same thing with the header file generation. Even if we use --llvm, we still want to emit C declarations for the header, so info->cfile should be set.

compiler/codegen/library.cpp
lydia-duncan: No, that's correct.  I was able to extend chpl_library_finalize to deinit modules because module initialization adds the deinit function to a list of functions to call, so the runtime only has to know that the list will be accurate instead of its contents.  We did discuss how to set up the library initialization functions to all get called by chpl_library_init back in May, but it required more work to implement and I haven't prioritized it yet.

compiler/codegen/symbol.cpp
lydia-duncan: That makes sense.  I remember thinking something along those lines at some point so I'm surprised this one survived, I'll remove it.

compiler/codegen/library.cpp
lydia-duncan: I've been using mostly to deal with the things that get set via compiler flags (mostly because I was following precedent).  I don't expect this to be set via a compiler flag, though, so I will definitely use a C++ string instead!

PR #10791, opened 2018-08-15
Warning for initing local domain from distributed

mppf:
Fixes issue #5657. Newer attempt after PR #5661.

For example, in the following code:
``` chapel
 use BlockDist;
  config const n = 100;
  var myDomain: domain(2) = {1..n, 1..n} dmapped Block({1..n, 1..n});
  for d in myDomain {
    writeln(d.locale.id);
  }
```

the user might think that `myDomain` is distributed, when in fact it is not. The purpose of the warning is to call attention to that issue since it's happened to multiple users over time (and been challenging for them to figure out).

- [x] full local testing
- [x] full local gasnet testing

### DIFF
--- a/compiler/resolution/resolveFunction.cpp
+++ b/compiler/resolution/resolveFunction.cpp
@@ -1808,6 +1808,13 @@ static void insertCasts(BaseAST* ast, FnSymbol* fn, Vec<CallExpr*>& casts) {
                 resolveExpr(moveInit);
                 resolveExpr(assign);
 
+                // Enable error messages assignment between local
+                // and distributed domains. It would be better if this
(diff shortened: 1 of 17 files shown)


mppf:
I think this is ready to merge. @benharsh has agreed to shepherd it next week.

benharsh:
@bradcray : do you want to catch up on this PR before it is merged?

bradcray:
This generally looks good to me.  I probably would've updated the .good files for the tests whose behavior changed rather than rewriting the tests to avoid the warnings (while still having the same behavior).  It feels like it's sweeping something under the rug rather than using pre-existing tests to provide additional coverage for this new feature.

Issue #10520, opened 2018-07-25
Memory leak when pop_front pops last element in array


benharsh:
Consider the following program:
```chpl

class C {
  var x : int;
}

record R {
  var c : C;
  proc init() { }
  proc init(x:int) {
    c = new C(x);
  }
  proc init(other:R) {
    this.c = new C(other.c.x);
  }
  proc deinit() {
    if c != nil then delete c;
  }
}

proc =(ref LHS : R, rhs : R) {
  if LHS.c then delete LHS.c;
  LHS.c = new C(rhs.c.x);
}

proc main() {
  var A = for i in 1..4 do new R(i);

  // A copy of R(4) is leaked
  while A.size > 0 do A.pop_front();
}
```

In this program, a copy of the last element "R(4)" is made during the final pop_front but is never deinitialized.

If I compile with ``-sdebugArrayAsVec`` and run with ``--memLeaks`` I observe the following output:
```
pop_front reallocate: 1..4 => 3..4 (5..4)

=================================================================================================================
Allocated Memory (Bytes)         Number   Size     Total    Description                      Address             
=================================================================================================================
foo.chpl:22                      1        16       16       C                                0x00007fc501d01a10  
=================================================================================================================
```

The second range (3..4) printed by pop_front debugging seems incorrect. This is the range passed to ``dsiReallocate``, but it is non-zero in length despite the new indices 5..4 being non-zero in length. I suspect this means we do not attempt to deinitialize array elements when ``A`` is finally destroyed.

bradcray:
@benharsh: You labeled this as a modules bug...  How confident are you that that's correct (i.e., that it's not a compiler bug?).  Maybe put another way: Is this something that @daviditen could/should fix as the author of `pop_front()` or does it require someone fixing memory management of records in general?  Or do we need someone to dig into it to figure that out?  Thanks.

benharsh:
I think the bug is due to incorrect math in the module code. I'm not 100% sure, but the new ranges for the internal and top-level indices look wrong to me, and the compiler wouldn't have control over those.

I think someone familiar with the ``pop_front`` implementation (@daviditen or whoever) should investigate further and verify.

ian-bertolacci:
I have a program where *all* the listNodes leak:
```
proc foo( N : int ){
  var l : list(int);
  for i in 1..#N do l.push_back( i );

  while l.size > 0 {
    var v = l.pop_front();
  }
}

config const N = 3;

proc main(){
  foo(N);
[...]

PR #10793, opened 2018-08-16
Add a new dsiReallocate that is aware of allocated size and actual array size

daviditen:
When array.pop_front and friends caused a reallocation, they were only paying
attention to the allocated size, which resulted in copies of elements that
were not going to be in the newly resized array.  Make dsiReallocate use both
the allocated bounds and the logical array bounds.  Allocate and free based
on the allocated bounds, but copy elements based on the logical array bounds.

This fixes a leak in the newly added test from issue #10520.


daviditen:
Passed a full local Linux64 paratest.

daviditen:
Thank you!

PR #10795, opened 2018-08-16
Documentation for Mason Packages

Spartee:
### Description
The documentation around Mason packages will need to improve if more users are to adopt the package manager as their methodology of developing chapel programs. This is a step in that direction. 

The documentation is generated via a python script that requests the Mason.toml files found in the [Mason-Registry](https://github.com/chapel-lang/mason-registry/tree/master/Bricks). From there the script writes the packages to the `package-list.rst` which currently resides in the `tools` section of the documentation.

#### Design points
- Where should the `package-list.rst` be located in the chapel documentation?
- Should the script run every night? Once a week? Once a month?


ben-albrecht:
@LouisJenkinsCS - While the features you've requested sound nice, I think they are something that would be pursued in a longer term separate mason packages site, like the [Cargo Crates Website](https://crates.io/crates). I view the effort in this PR as a short-term solution for us to display packages to users without having to make the significant time commitment to designing a new web page.

Of the features you requested, I'd agree with alphabetical sorting and package summaries as being within the scope of this PR. That said, I believe getting package summaries would require pulling package data (unless this is in the manifest file), so I'm not sure that is feasible. @Spartee can double check me on that.

Spartee:
Agreed with @ben-albrecht that most of those feature are what we would want to pursue in a more fully-featured version. 

Alphabetical sorting is entirely possible and will happen in this PR.

We could do summaries pretty easily if they were inside of the `Mason.toml`, however, anything within the packages themselves would require a significant amount of effort. (requests to source libs, parse files, chpldoc files, link generated rst's, etc.)

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+import base64
+import requests
+from pprint import pprint
+import re
+
+# get authentication token for GitHub
+secret = open("secret-token.txt", "r+")
```
ben-albrecht: Are we guaranteed to hit a request limit if we don't have this? I wonder if we should have a way to omit using authentication. Curious on your thoughts, but I don't see it as a necessary change for this first draft.

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
...
+info = secret.readline().split(",")
+user = info[0].strip()
+token = info[1].strip()
+secret.close()
+
+f = open("../rst/tools/mason/package-list.rst", "w+")
```
ben-albrecht: Will running this script more than once without some sort of 'clean' step cause issues?

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
...
+        package = {}
+        package["name"] = brick["name"]
+        package["versions"] = []
+        content = requests.get(brick["url"], auth=(user, token))
+        content = content.json()
+
+        for toml_file in content:
+            download = requests.get(toml_file["download_url"], auth=(user, token))
+            toml = download.text
+            for line in re.findall("(\w+) = ([^\s]+)", toml):
```
ben-albrecht: Seems safer / more robust to rely on the `toml` module to read these values.

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+        # put packages in top lvl package dict
+        packages[package["name"]] = package
+
+
+    # write package-list.rst
+    for name, pack in sorted(packages.items()):
+        if not name.startswith("_"):
+            f.write(name + "\n")
```
ben-albrecht: This might be cleaner as an encapsulated function, especially if we used to `toml` module to create a `toml` object, e.g. `def writePackageInfo(f, toml_object)`.

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+import base64
+import requests
+from pprint import pprint
```
ben-albrecht: I don't see this used anywhere.

doc/util/gen-package-list.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
ben-albrecht: Would be nice to have a top-level doc string for this.

doc/util/gen-package-list.py
Spartee: No, the current one will be overwritten

doc/util/gen-package-list.py
Spartee: Not guarenteed at all but I wanted to put that in there just in case someone ever did get the authentication token for Chapel.

ben-albrecht: OK, so adding an option to skip this could be helpful in the future -- would you mind adding a TODO comment for that?

doc/util/gen-package-list.py
```diff
@@ -1,59 +1,82 @@
...
-    # write package-list.rst
+import toml
+
+"""
+This script requests the GitHub API for the Mason.toml's
+found in the Mason Registry so that a package list can be
+automatically generated.
+
+An authentication token is needed from github in order to
+make more than a few requests ever hour.
```
ben-albrecht: s/ever/every

PR #10797, opened 2018-08-16
Extend --library compilation to respond appropriately to `require` statements

### LINKED ISSUE #9466: 'require' clauses not honored in library compilation
Not sure if this is an "issue", per say, but it's important to note that when linking a library created with the --library flag, if it has `require` statements involving libraries you have to explicitly provide those flags.

For example, if your library code has `use ZMQ;`, you have to link it with `-lzmq`

lydia-duncan:
Prior to this change, users of Chapel compiled libraries involving `require`
statements would have to repeat the information to properly build with
the compiled library.  This change adds #includes for headers specified
via a `require` statement to the generated header file.  In the case
where a Makefile is requested, the generated Makefile will include any
known directories for #included files in the CHPL_CFLAGS variable, and
any known libraries directories and specified `-l` libraries in the
CHPL_LDFLAGS variable.

Resolves #9466

No special treatment is performed for `require "foo.chpl"`,
`require "foo.c"` or `require "foo.o"`, as the Chapel compiler will have
already included the source file when compiling the library (I double checked
this locally by modifying test/extern/bradc/require/usefoobarbaz.chpl).

Note that the header change also supports headers named when compiling
the library directly, e.g. `chpl --library foo.chpl bar.h`

Added two tests, one to lock in the .h behavior and one to lock in the `-l`
behavior.  Modified a test that was specifying a .h on the command line to
no longer hack its header file to #include the .h (since the header file now
gets generated with it)


lydia-duncan:
(no rush on the review, btw)

lydia-duncan:
(I forgot to commit that skipif, whoops)

PR #10811, opened 2018-08-17
Fix memory leak in list.pop_front

ian-bertolacci:
Previously, list.pop_front did not delete the unmanaged listNode instances that it removes from the list, resulting in memory leaking, which this corrects.
(additionally, there was some inconstant formatting which was brought in line with the rest of the file.)

Tests pass on my end.

### DIFF
--- a/modules/standard/List.chpl
+++ b/modules/standard/List.chpl
@@ -184,18 +184,19 @@ record list {
      Remove the first element from the list and return it.
      It is an error to call this function on an empty list.
    */
- proc pop_front():eltType {
-   if boundsChecking && length < 1 {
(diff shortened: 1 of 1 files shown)


lydia-duncan:
This looks reasonable to me.  I'm going to run a paratest and memleaks check, and will wait to merge until Monday because I don't want to check testing this weekend

lydia-duncan:
Memory leaks testing looks like this gets rid of some listNode leaks but not all (the int and BaseDom ones are gone but at least some BaseArr remain).
I think this is good to go, thanks!

ian-bertolacci:
Great thanks!

PR #10813, opened 2018-08-17
Turn off Qthreads guard pages for LCALS on high core-count systems

dmk42:
LCALS appears to suffer from the same Qthreads guard pages issue that was afflicting lulesh and miniMD.  See #10530 and #10661 .  This change copies the execenv from those tests to LCALS.

Turning off guard pages also makes LCALS well over 10% faster.


dmk42:
@gbtitus and @ronawho - FYI

PR #10822, opened 2018-08-20
Close a leak with some visibility tests

lydia-duncan:
User error, I think these had never been cleaned up and so converting them to
use owned is the quickest and most sensible fix

Double checked a fresh checkout


lydia-duncan:
@mppf - The tests that rely on these files never cleaned up the return value of these functions.  Here I'm choosing to have the functions return an owned object instead of modifying the tests to clean up the unmanaged instance.  Do you think that's a reasonable course of action?

mppf:
Sure, looks good to me.

PR #10827, opened 2018-08-20
Make remaining string methods work with UTF-8

### LINKED ISSUE #10479: Make remaining string methods work with UTF-8
The methods that need to be made UTF-8 aware are:
```
split(), startsWith(), endsWith(), find(), rfind()
count(), replace(), join(), strip(), partition()
```
Some of these will definitely require modification.  Some others might not.

dmk42:
This finishes the conversion to UTF-8 based on codepoints by making the string methods not previously converted work with UTF-8.

Testing this change found three preexisting bugs that had to be fixed.
* `wcrtomb()` was being used incorrectly.
* `qio_set_glocale()` was not always being called when needed.
* String literals were being built with malformed hex escapes.

Closes #10479


dmk42:
@mppf - Would you please review?

dmk42:
@mppf - I am having second thoughts about my implementation of the trailing part of `strip()`.  Rather than creating a new copy of the string by iterating over `localThis[start..]`, maybe it would be better to do the work on `localThis` itself and just skip the characters we find that occur before `start`.  What do you think?

modules/internal/String.chpl
```diff
@@ -452,7 +452,16 @@ module String {
     /*
       Iterates over the string Unicode character by Unicode character.
     */
-    iter uchars(): int(32) {
+    inline iter uchars(): int(32) {
+      for (codepoint, i, nbytes) in this.ucharsIndexLen() do
+        yield codepoint;
+    }
```
mppf: does `inline iter` actually do anything?

modules/internal/String.chpl
```diff
@@ -452,7 +452,16 @@ module String {
...
+
+    /*
+      Iterates over the string Unicode character by Unicode character,
+      and includes the byte index and byte length of each character.
+    */
+    inline iter ucharsIndexLen() {
```
mppf: I'm not sure this is a great name for it. I'd consider making this an internal function for the moment.

mppf: It also might be better to have the old iterator for performance reasons. Tuples are pretty fast but aren't free.

modules/internal/String.chpl
```diff
@@ -936,12 +944,12 @@ module String {
                   done = true;
                 // no limit
                 } else {
-                  chunk = localThis[chunkStart..i];
+                  chunk = localThis[chunkStart..i-1];
```
mppf: Why i - 1 here? Wouldn't it need to be subtracting a codepoint size?

modules/internal/String.chpl
```diff
@@ -1099,14 +1107,19 @@ module String {
...
-        label outer for i in 0..#localThis.len by -1 {
-          for j in 0..#localChars.len {
-            if localThis.buff[i] == localChars.buff[j] {
-              end -= 1;
+        // Because we are working with codepoints whose starting byte index
+        // is not initially known, it is faster to work forward, assuming we
+        // are already past the end of the string, and then update the end
+        // point as we are proven wrong.
+        end = 0;
+        label outer for (thisChar, i, nbytes) in localThis[start..].ucharsIndexLen() {
```
mppf: it seems to me ucharsIndexLen should be able to start at a particular byte offset (start).

modules/internal/String.chpl
```diff
@@ -1099,14 +1107,19 @@ module String {
...
+          for removeChar in localChars.uchars() {
+            if thisChar == removeChar {
               continue outer;
             }
           }
-          break;
+          // This was not a character to be removed, so update tentative end.
+          end = start-1 + i + nbytes-1;
```
mppf: I find the two -1s here suspicious

modules/internal/String.chpl
dmk42: Good point.  `linline iter` does nothing.

modules/internal/String.chpl
dmk42: Thanks.  I will make ucharsIndexLen an internal function and restore the old uchars iterator.
I'm not sure there is a good name for ucharsIndexLen.

modules/internal/String.chpl
dmk42: `i` was zero-based and is now one-based, so now we need `i-1`.

modules/internal/String.chpl
dmk42: Yes, that will simplify this method.  Thanks.

modules/internal/String.chpl
dmk42: These are three one-based numbers, so we have to subtract one from two of them to get a one-based result.

modules/internal/String.chpl
dmk42: To clarify, `i` indexes the *last* byte of the *previous* character, which is why we only need to subtract 1 and not the codepoint size.

dmk42:
@mppf - Thanks for the feedback.  This is ready for another look now.

modules/internal/String.chpl
```diff
@@ -467,6 +467,28 @@ module String {
       }
     }
 
+    /*
+      Iterates over the string Unicode character by Unicode character,
+      and includes the byte index and byte length of each character.
+      Skip characters that begin prior to the specified start index.
```
mppf: say index here is in bytes

runtime/src/chpl-init.c
```diff
@@ -141,6 +141,7 @@ void chpl_rt_init(int argc, char* argv[]) {
   // UTF-8 functions (e.g. wcrtomb) work as
   // indicated by the locale environment variables.
   setlocale(LC_CTYPE,"");
+  qio_set_glocale();
```
mppf: I think this used to be here but we moved it into qio functions e.g. qio_channel_scan_string for some reason I can't remember. Anyway I personally think it makes more sense here.

I'm fine with it being here but if it is you should remove the other calls to it.

test/types/string/dmk/utf8_values.chpl
```diff
@@ -0,0 +1,9 @@
+var mystring1 = "événement";
+
+for c in mystring1.uchars() {
+  const str = codePointToString(c);
+  write(c, "\t", str);
+  for i in 0..#str.length do
+    write(" ", str.buff[i]:int);
+  writeln();
+}
```
mppf: You definitely need more testing - especially strip testing left and right sides. Ideally every string fn with unicode support should have a test.

PR #10842, opened 2018-08-21
Add a simple ocean modeling code ported from a python code

daviditen:
This code imports some data from a netCDF file, does some computation
on the data and then writes out the array.

I verified that it gets bitwise equal results for the final array as the
python code it was ported from.  The python version uses matplotlib to
display the result as a nice picture, but for now I just printed out the
final array.

On my system, the python version runs in about 215 seconds.  The chapel
version runs in 0.24 seconds.


daviditen:
Thanks @vasslitvinov!

PR #10847, opened 2018-08-21
Turn override checking on by default

### LINKED ISSUE #10828: turn override-checking on by default
- [ ] test NUMA
- [ ] test GASNet

mppf:
Resolves #10828 

It can be disabled with --no-override-checking.

- [x] full gasnet testing
- [x] full CHPL_LOCALE_MODEL=numa testing

### DIFF
--- a/compiler/main/driver.cpp
+++ b/compiler/main/driver.cpp
@@ -177,7 +177,7 @@ bool fLocal;   // initialized in postLocal()
 bool fIgnoreLocalClasses = false;
 bool fUserDefaultInitializers = true;
 bool fLifetimeChecking = true;
-bool fOverrideChecking = false;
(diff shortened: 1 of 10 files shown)


bradcray:
Woohoo!

Something that I was trying to figure out for myself today:  Should the `--[no-]override-checking` flag remain a developer flag (as it is today) on the argument that typical users shouldn't be turning override checking off ("it's part of the language!"), or should it be made a user flag to help users transition their code over?  (or does it merely help them avoid transitioning their code over?).  I think I concluded that I'm comfortable leaving it a developer / undocumented flag, but wanted to ask the question in case it slipped through the cracks rather than being a conscious decision.

bradcray:
Sounds good.

Issue #10846, opened 2018-08-21
Use NB atomics when chained transactions aren't availble for BufferedAtomics


ronawho:
Currently the buffered atomics module uses chained transactions to perform multiple network atomics as 1 ugni transaction. Chained transactions are a relatively new feature though and we know there are sites uses older CLE releases, so we should fall back to non-blocking atomics instead of just doing blocking atomics like we do today.

PR #10857, opened 2018-08-22
Do NB ops when chained transactions aren't available for buffered atomics

ronawho:
Chained transactions were only added in CLE 5.2UP04, and there are some sites
that are still using CLE versions older than this. Previously, if chained
transactions weren't available we just fell back to using blocking atomics,
which meant there was no performance benefit to using buffered atomics. This
changes things so that we still buffer atomics and when we flush the buffer do
non-blocking operations and then wait for all operations to complete instead of
issuing and waiting for them one at a time.

Performance for bale histo with various atomic implementations:
 -  80 MB/s per node for blocking atomics
 - 200 MB/s per node for buffered atomics using non-blocking transactions
 - 360 MB/s per node for buffered atomics using chained transactions

So performance is roughly 2.5X faster than just using blocking atomics when
chained transactions aren't available, but still behind using chained
transactions.

Closes https://github.com/chapel-lang/chapel/issues/10846


runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6940,15 +6942,50 @@ void do_remote_amo_V(int v_len, void** opnd1_v, c_nodeid_t* locale_v,
...
+    post_desc_v[vi].length          = size_v[vi];
+    post_desc_v[vi].amo_cmd         = cmd_v[vi];
+    post_desc_v[vi].first_operand   = size_v[vi] == 4 ? *(uint32_t*) opnd1_v[vi]:
+                                                        *(uint64_t*) opnd1_v[vi];
+
+    atomic_init_bool(&post_done_v[vi], false);
+    post_desc_v[vi].post_id = (uint64_t) (intptr_t) &post_done_v[vi];
+
+    // initiate the transaction
+    cdi_v[vi] = post_fma(locale_v[vi], &post_desc_v[vi]);
```
gbtitus: You might want to increase the CQ size of the CDs as part of this PR.  Here's the long-ish explanation why:
This will consume a CQ entry per transaction.  The CQ length per comm domain (CD) is fixed, set when the CD is created at init time.  The code that finds a suitable CD for each transaction does check that candidate CDs have space left in their CQs, so to some extent the existing code is resilient against bursts of CQ use caused by doing vectors of NB transactions at a time.  Basically, we'll just use multiple CDs where a series of blocking transactions might have used a single one.  But if many threads happen to do vectors of NB transactions all at once then we might actually run out of total CQ entries, since we only typically have as many CDs as we have worker threads.  If this happened it could cause a live-lock and pause, because the code here doesn't consume any CQ events, and thus free any CQ entries, until it's initiated the entire vector of transactions.  So if we run out of CQ entries across all CDs while all the threads are executing this code, we'll hang when since no thread will be able to find a CD with free CQ entries to post on.

ronawho:
The tests added in https://github.com/chapel-lang/chapel/pull/10868 are failing with this PR as-is. Currently we're storing pointers to operand 1 in an array but the operands being pointed to aren't around when we're actually going to issue the transaction. An upcoming commit will address that but I'm going to sit on this over the weekend and come back to it with fresh eyes (and probably chat with Greg and get another round of review in.)

runtime/src/comm/ugni/comm-ugni.c
```diff
@@ -6863,57 +6863,199 @@ void check_nic_amo(size_t size, void* object, mem_region_t* remote_mr) {
...
+    else
+      pdc[ci-1].next_descr  = &pdc[ci];
+    pdc[ci].next_descr      = NULL;
+    pdc[ci].remote_addr     = (uint64_t) (intptr_t) object_v[vi];
+    pdc[ci].remote_mem_hndl = remote_mr_v[vi]->mdh;
+    pdc[ci].length          = size_v[vi];
+    pdc[ci].amo_cmd         = cmd_v[vi];
+    pdc[ci].first_operand   = opnd1_v[vi];
+  }
+
```
gbtitus: Think you need to do `pdc[ci].next_descr = NULL;` after the loop, to mark the end of the chain.  That's not otherwise guaranteed.

ronawho: That happens a few lines above. We set `pdc[ci].next_descr      = NULL;` and on the lines before that we update the previous pdc's next_descr to point at the current index (that logic was copied from the put_V code)

gbtitus: Oh, crud, never mind.  Sorry.

gbtitus:
👍

PR #10861, opened 2018-08-22
Expand generated Python files to appropriately handle wrapped array types

### LINKED ISSUE #10789: Handle arrays in the generated Python files
The .pxd file needs to export the definition of chpl_external_array, and uses of functions that utilize arrays in the .pyx file need to utilize that type.  The latter translation is similar to what we do for c_strings (but slightly different).

We need to utilize the element type.  This information may be obfuscated for us due to the translation to chpl_external_array at normalization.  Some thought is necessary to determine how to obtain that information again.

lydia-duncan:
Prior to this change, generated Python modules when arrays were involved would
attempt to use the chpl_external_array type name, when the name was not visible
to Python, and the type wasn't as friendly to Python array-like structures.  This
change performs some additional translation for array arguments and return values
(as was originally intended, but waited on due to its complexity) - with it, Python
lists and numpy arrays can have their contents sent to Chapel functions that
take arrays, and Chapel functions called from Python will yield numpy arrays

Resolves #10789

In more detail:
Because the chpl_external_array wrapper loses the element type information
at normalize, I created a map from argument/function symbol to the element
type needed when the wrapper is inserted.  This map is used at code generation
time to ensure Python doesn't give us garbage types.

I moved the generated .pxd file to a constant .pxd file defined in the runtime,
and extended it to export the definitions for the chpl_external_array type as
well as the two functions to make it and the free function.  The .pyx file
imports the functions and uses them.

I altered the handling of argument types when an array is involved - instead
of limiting to a specific Python type, I removed the type so that Python lists
and numpy arrays would be accepted.  If something that is not appropriately
iterable is passed, then the user will get an error.

Additionally, in the .pyx file I now create an instance of chpl_external_array
and populate it with the contents of the Python argument (similar to the
translation for Python bytes to C strings, but more intensive), sending that
instance on to the wrapped function.  In the case where an array is returned,
I create a numpy array of the appropriate type and copy the contents from
the chpl_external_array wrapper, freeing the wrapper afterwards.

Testing changes:
I needed to update the expected output for the other python tests due to the
new lines in the .pxd/.pyx files to support chpl_external_array's use.

I added three new tests: one for integer arrays, one for arrays of reals, and
a future on explicitly taking or returning the chpl_external_array type.  The
last was useful to ensure that I was not accidentally impacting these
explicit instances when I shouldn't but requires more work to create a .pyx
file that can be compiled into a Python module.  I have chosen to wait to
finish that, though I have a mock-up of what that should eventually look like
in the .good file.

Passed a full std/ paratest with futures


compiler/codegen/symbol.cpp
```diff
@@ -1803,6 +1865,32 @@ GenRet FnSymbol::codegenPYXType() {
...
+  // Create the numpy array to return
+  // E.g. cdef numpy.ndarray [C element type, ndim=1] ret =
+  //          numpy.zeros(shape = ret_arr.size, dtype = Python element type)
+  std::string res = "\tcdef numpy.ndarray [" + typeStrCDefs + ", ndim=1] ret";
+  res += " = numpy.zeros(shape = ret_arr.size, dtype = " + typeStr + ")\n";
+
+  // Populate it with the contents we return (which translated C types into
+  // Python types)
+  res += "\tfor i in range(ret_arr.size):\n";
+  res += "\t\tret[i] = (<" + typeStrCDefs + "*>ret_arr.elts)[i]\n";
```
mppf: How many times do we copy these arrays? I thought one of the points of numpy arrays was that they could refer to arrays allocated in C. I think reducing the number of copies on this interface will be important for performance but obviously it can be put off.

compiler/codegen/library.cpp
```diff
@@ -27,6 +27,8 @@
 #include "stlUtil.h"
 #include "stringutil.h"
 
+std::map<Symbol*, TypeSymbol*> elementType;
```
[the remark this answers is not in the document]
lydia-duncan: I'm having it handle return types as well (in which case the mapping is FnSymbol->TypeSymbol), so would `exportedArrayElementType` be okay?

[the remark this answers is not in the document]
mppf: yes

compiler/codegen/symbol.cpp
lydia-duncan: I tried a couple of things prior to this implementation that I had difficulty with.  I intend to return to them at some point, but figured slower support was better than no support.

- I tried experimenting with Cython's [memory views](http://docs.cython.org/en/latest/src/userguide/memoryviews.html) but was having trouble with getting it to play nice due to type stuff.

- I also tried initializing the numpy array's buffer to reference `elts` but hadn't fully wrapped my head around how to store function pointers from C into a Python type

mppf: > I also tried initializing the numpy array's buffer to reference elts but hadn't fully wrapped my head around how to store function pointers from C into a Python type

Since we know how to call a deallocator, can we use a function pointer to the appropriate Python "free" function (which might decrement a reference count, after all it's Python)?

Anyway, probably worth an issue if you don't sort it out soon. I think there are other places with unnecessary copies.

Issue #10838, opened 2018-08-21
spike: `:` in where clauses next steps


mppf:
PR #10143 changed lots of where clauses to not use `:` but there are several next steps towards finishing #10019 remove/deprecate special treatment for `:` in where clauses:

- [x] tidy up remaining uses of isSubtype (array heirarchy ones should call 
   e.g. isDefaultRectangular etc)
- [x] create deprecation error for : in where (in changeCastInWhere)
- [x] look into supporting `<=` and other comparisons on types (see if it is easily feasible)
- [x] look into behavior changes if isSubtype allows coercions (see what tests break if isSubtype allows coercions)

mppf:
Regarding the operators, it could be handled by having resolution specially detect "<=" calls where arguments are types. It's not possible earlier in the compiler since <= could apply to integers (say).

PR #10865, opened 2018-08-23
Adds support for isSubtype checking via operators

mppf:
This PR adjusts the compiler to support `<` `<=` `>` and `>=` operations
on types directly. These are not handled in overloads in ChapelBase (say)
because then it would not be possible to use them with generic types.
This PR modifies resolveCall to call a new function that matches on the
operator name applied to types and replaces it with an appropriate
primitive. While there, made isProperSubtype be treated similarly to
isSubtype. Both of these efforts required adding a new primitive,
`PRIM_IS_PROPER_SUBTYPE`, which is just a variant of `PRIM_IS_SUBTYPE` to
support `<`.

- [x] full local testing

Relates to #10838.


compiler/resolution/functionResolution.cpp
```diff
@@ -2064,6 +2069,55 @@ FnSymbol* tryResolveCall(CallExpr* call) {
...
 }
 
+static bool resolveTypeComparisonCall(CallExpr* call) {
+
+  if (UnresolvedSymExpr* urse = toUnresolvedSymExpr(call->baseExpr)) {
+    if (call->numActuals() == 2) {
+      const char* name = urse->unresolved;
+
+      if (name[0] == '<' || name[0] == '>') {
+
```
vasslitvinov: How about using comparisons like `name == astrSlt` `name == astrSge` etc. following the lead of astrSdot et al.?

Issue #10845, opened 2018-08-21
Add documentation and tests for BufferedAtomics module


ronawho:
Add documentation and correctness tests for the work done in https://github.com/chapel-lang/chapel/pull/10702

PR #10868, opened 2018-08-23
Add some buffered atomic correctness tests

ronawho:
Add a basic API test to ensure that all operations are working as expected and
add a stress test to ensure highly concurrent updates (and concurrent flushing)
work as expected.

Part of https://github.com/chapel-lang/chapel/issues/10845


ronawho:
@benharsh would you mind giving this a quick sanity check?

@gbtitus mind giving this a post-commit review (since you reviewed the original tests and the work this is testing)

gbtitus:
I made a line comment recommending you do 2 ops instead of just 1 in the API test, but I guess the presence of the stress test makes that less necessary.  Otherwise, looks good.

PR #10869, opened 2018-08-23
Generate last file needed for compiling a Python module using Cython 

### LINKED ISSUE #9311: Generate files needed for successful Cython interoperation with Chapel
Today, a Cython run with a Chapel library requires 7 components:
- The library .chpl file
- A .h file with the C declaration of what is exported from that .chpl file
- The generated library itself (.a, .so)
- a .pxd for the library with the `cdef extern` definitions taken from the relevant .h file(s)
- a .pyx to import the library specification into a form Python understands
- a .py file that "cythonize"s the above components
- the python program that uses the library.

Aside: it is likely possible to combine the .pyx and .pxd file into one, but I was having difficulty with that approach.

Anyways, it would be nice to make it so that the user doesn't have to explicitly write all of these files by hand since they are somewhat tricky to get right (and obnoxious, and repetitive, and something that can be determined from the contents of the library .chpl file...).  Ideally, we would simplify it so that the user only has to specify the .chpl file and the python program.

This is definitely a large issue.  I can break it down based on the type of file that needs to be generated and what needs to be supported within it (i.e., types, functions, etc), but estimating each of those components is going to need some breakdown as well.

### LINKED ISSUE #10818: Generate last file needed to compile Python modules from Chapel libraries
The last file needed to create a Python module using Cython is a .py file (typically a setup.py, but if the user is making multiple modules in the same directory, that could run into conflicts so I'm planning to use the same base name as the .pyx file).  This file is of the format:

```python
from distutils.core import setup
from distutils.core import Extension
from Cython.Build import cythonize
import numpy

# grab the base library names from compileline --libraries and strip off the -l
import compileline # note 1
chpl_libraries=[l[len('-l'):] for l in compileline.libraries.split() if l.startswith('-l')]
setup(
    name = 'foo library', # note 2
    ext_modules = cythonize(
        Extension("foo", #note 3
                  include_dirs=[numpy.get_include()],
                  sources=["foo.pyx"], # note4
                  libraries=["foo"] + chpl_libraries) # note 5
        )
    )
```
Where note 1 enables including the static version of the runtime and third-party libraries in the Cython compile, note 2 is a description of the library, note 3 is the name of the python module, note 4 the .pyx file, and note 5 the library file (plus the runtime and third-party libraries mentioned in note 1)

lydia-duncan:
This file handles pulling in the various components needed for a successful
Cython compile - the libraries, the .pyx file, the name to generate, numpy, etc.

Resolves #9311
Resolves #10818

Updates the python tests to include the new file.

Note that this doesn't build the Python module yet, even though all the files
are in place.  That needs to wait until we have a testing environment with
Cython installed.

TODO: merge in python arrays branch changes

Passed full paratest with futures

### DIFF
--- a/compiler/codegen/library.cpp
+++ b/compiler/codegen/library.cpp
@@ -455,6 +455,7 @@ static void makePYFile() {
     // Get the static Chapel runtime and third-party libraries
     fprintf(py.fptr, "chpl_libraries=[");
     bool first = true;
+    // Get the libraries listed in require statements
     for_vector(const char, libName, libFiles) {
       if (first) {
         first = false;
@@ -468,6 +469,8 @@ static void makePYFile() {
     libraries.copy(copyOfLib, libraries.length(), 0);
     int prefixLen = strlen("-l");
     char* curSection = strtok(copyOfLib, " \n");
+    // Get the libraries from compileline --libraries, taking the `name`
+    // portion from all `-lname` parts of that command's output
     while (curSection != NULL) {
       if (strncmp(curSection, "-l", prefixLen) == 0) {
         if (first) {
--- a/compiler/codegen/library.cpp
+++ b/compiler/codegen/library.cpp
@@ -17,6 +17,8 @@
  * limitations under the License.
  */
 
+#include <cstring>
+
 #include "library.h"
 
 #include "FnSymbol.h"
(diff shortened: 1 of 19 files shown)


compiler/codegen/library.cpp
```diff
@@ -421,6 +425,77 @@ static void makePYXSetupFunctions(std::vector<FnSymbol*> moduleInits) {
...
+      } else {
+        fprintf(py.fptr, ", ");
+      }
+      fprintf(py.fptr, "\"%s\"", libName);
+    }
+    std::string libraries = getCompilelineOption("libraries");
+    char copyOfLib[libraries.length() + 1];
+    libraries.copy(copyOfLib, libraries.length(), 0);
+    int prefixLen = strlen("-l");
+    char* curSection = strtok(copyOfLib, " \n");
```
mppf: Let's add a comment saying generally what this loop is trying to achieve

PR #10876, opened 2018-08-23
Add more buffered atomic tests

ronawho:
Add many-to-one and ra-atomic performance tests and beef up the stress test
 - Buffered many-to-one gets 400 M ops/sec compared to 240 for unbuffered
 - Buffered ra-atomics gets .75 GUPS compared to ~.17 GUPS for unbuffered
 - Improve the buffered atomic stress test. Test subBuff and test with
   different values instead of always doing `addBuff(1)`


ronawho:
Easiest to review commit-by-commit (and ignore the ra-atomic clone commit)

ronawho:
> This looks good, but I think you'll want to add the new variant to the Makefile and README as well.

Good catch -- modified them

test/release/examples/benchmarks/hpcc/variants/README
```diff
@@ -11,6 +11,7 @@ annual HPCC competition).
 The main files for these benchmarks are as follows:
 
      stream-promoted.chpl   : global STREAM Triad using promotion
+     ra-atomics.chpl        : a version of RA using buffered atomic vars
```
benharsh: Did you mean "ra-buff-atomics" ?

ronawho: Whoops, also a good catch. Force pushed the correction

gbtitus:
👍

Issue #10833, opened 2018-08-21
Add Jenkins Job for testing Spack Integration


Spartee:
The tests for the Spack integration are found in `CHPL_HOME/test/mason/spack-integration`. These tests need to be run on a nightly job on the CentOS VM(chapel-centos7-02). 

Steps to reproduce this testing manually(might be of benefit):
- clone Chapel repo
- `make`
- `make test-venv`
- `source util/setchplenv.bash`
- `cd $CHPL_HOME/tools/mason && make mason`
- `cd $CHPL_HOME/test/mason/spack-integration && start_test`

All env vars should be taken care of by Mason.

These tests do the following:
- Spack download and bootstrap setup
- Installation of `openBLAS` package through Spack integration
- test compilation of custom`LinearAlgebra` module using downloaded openBLAS

awallace-cray:
Build slave setups

Install gcc-fortran
`sudo yum install gcc-fortran compat-gcc-44-gfortran`

Configure gcc-fortran
* same userid as Chapel tests run
`mason external compiler --edit` 
* detail TBD

PR #10877, opened 2018-08-23
mason-external SKIPIF should not test a specific hostname

awallace-cray:
Instead of testing the hostname of the test machine, use
/etc/os-release to infer if the test machine is running a
CentOS-like Linux.

Throw in tests on CHPL_HOST_PLATFORM and _COMPILER for
good measure.

Related to issue #10833 

Also:
- .gitignore all Mason.lock files from test/mason on down;
- git rm the four Mason.lock files already checked-in.

- The SKIPIF should also make sure mason external has been setup on this machine.


awallace-cray:
@ben-albrecht @spartee 
Not sure how restricted this SKIPIF test must be. This SKIPIF works (ie does not skip) when run on either of our Centos7 VMs. I have not tested this SKIPIF on any other kind of build slave. Comments? 

I would prefer not to have hostnames in SKIPIF tests, but won't let that stop me from putting up the Jenkins job.

awallace-cray:
2018-08-24 06:54 - As above, this is just a bit of volunteer cleanup. Hope I got it right. 
- .gitignore all Mason.lock files from test/mason on down; 
- git rm the four Mason.lock files already checked-in.

awallace-cray:
@spartee 
If and when you get a chance, does this SKIPIF make sense, and is there any reason not to get rid of those Mason.locks? 
Thanks again! 
/Tony

Spartee:
@awallace-cray This looks good to me. SKIPIF makes sense. As long as this skips everything but the VM's which it looks like it does, then I think we are good.

awallace-cray:
I'm going to take that as "approved" -:
Thanks.

PR #10837, opened 2018-08-21
build_configs/cray-internal: Add new scripts to build Chapel Cray module RPM

awallace-cray:
Add new scripts to build a Chapel Cray module RPM with build_configs. 
Implements https://github.com/chapel-lang/chapel/issues/10119

New subdirectory `util/build_scripts/cray-internal` is added. The new scripts in that subdir will build Chapel Cray Module RPM from an existing, pre-built CHPL_HOME directory.  

Two new scripts are added to `util/build_scripts`, to support cray-internal and future subdirectories like it.  

The Chapel Cray-module-specific scripts in the new cray-internal subdir are designed to work with the more generic Chapel package-building scripts in the parent directory build_configs.  The idea is to implement package-specific details in a subdirectory (like cray-internal), while implementing general-purpose things that should apply to most Chapel packages in the parent directory (build_configs). 

Cray-internal may be used as a model for new scripts to build some different kind of Chapel package.  Start by creating a new subdirectory under build_configs; copy and rename the main job script `chapel_package-cray.bash` from cray-internal to the new subdir; edit that file; and go from there, copy-editing additional files from cray-internal as needed.  

This PR also adds a new cmdline option '\-B COMPONENT' to setenv-example-2 and setenv-example-3. This makes the setenvs more flexible and useful to developers.

PR #10897, opened 2018-08-27
build_configs: minor corrections/improvements to setenv-3 (PR #10663

awallace-cray:
These low-impact changes are in response to feedback/experience with external setenv-example-3 user, and should be promoted asap. 

All commits were previously developed (and tested) with PR #10837, and are now moved into this separate PR for transparency:
* For portability, setenv-examples-3 should just accept the default intel version
* build_configs: add CHPL_GMP to build_configs.py dimensions
* build_configs: all printchplenv's should use --anonymize    
  And otherwise use the same options throughout build_configs, for consistency.

### DIFF
--- a/util/build_configs/build_configs.py
+++ b/util/build_configs/build_configs.py
@@ -92,6 +92,7 @@ known_dimensions = [
     ( 'unwind',             'CHPL_UNWIND', ),
     ( 'mem',                'CHPL_MEM', ),
     ( 'atomics',            'CHPL_ATOMICS', ),
+    ( 'gmp',                'CHPL_GMP', ),
     ( 'hwloc',              'CHPL_HWLOC', ),
     ( 'regexp',             'CHPL_REGEXP', ),
     ( 'llvm',               'CHPL_LLVM', ),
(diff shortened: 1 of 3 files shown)


awallace-cray:
2018-08-27 11:49 : created this PR by cherry-picking commits from PR #10837:
git cherry-pick 3806f5e245ae6298fa1a7ac0ef4de325bc111e8c 4bebb4619eb4e205807403c845565214903248d6 7dc367028839acb86536a0a5879937b4a9111e09

awallace-cray:
@gbtitus  
As discussed today. Would you review?  These were tested with #10837, but I am running a new test anyway. Thanks.

util/build_configs/build_configs.py
```diff
@@ -92,6 +92,7 @@ def validate(self):
     ( 'unwind',             'CHPL_UNWIND', ),
     ( 'mem',                'CHPL_MEM', ),
     ( 'atomics',            'CHPL_ATOMICS', ),
+    ( 'gmp',                'CHPL_GMP', ),
```
gbtitus: This is going to greatly lengthen your build time.  Do you really need it, or would some other setting work?  Frankly I'd be inclined to use "none" here, since these are just preliminary builds we're doing to see if the technique of having sites build their own modules even works.

PR #10900, opened 2018-08-27
Add documentation for Buffered Atomics

ronawho:
Closes https://github.com/chapel-lang/chapel/issues/10845

### DIFF
--- a/modules/packages/BufferedAtomics.chpl
+++ b/modules/packages/BufferedAtomics.chpl
@@ -17,6 +17,63 @@
  * limitations under the License.
  */
 
+/*
+   This module provides buffered versions of non-fetching atomic operations for
+   all ``int``, ``uint``, and ``real`` types.  Buffered versions of
+   :proc:`~Atomics.add()`, :proc:`~Atomics.sub()`, :proc:`~Atomics.or()`,
+   :proc:`~Atomics.and()`, and :proc:`~Atomics.xor()` are provided. These
+   variants are internally buffered and the buffers are flushed implicitly when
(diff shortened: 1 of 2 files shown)


ronawho:
Rendered docs available internally at web/~eronagha/buffDocs/modules/packages/BufferedAtomics.html

modules/packages/BufferedAtomics.chpl
```diff
@@ -17,6 +17,63 @@
...
+           a.addBuff(i);                   // buffered atomic add
+
+     flushAtomicBuff();                    // flush any pending operations (required)
+
+
+     const itersSum = iters*(iters+1)/2,   // sum from 1..iters
+           numTasks = numLocales * numTasksPerLocale;
+     assert(a.read() == numTasks * itersSum);
+
+   It's important to be aware that buffered atomic operations are not
```
gbtitus: This could be read as saying that the ops don't happen until the explicit flush is done.  I'd rather not leave that misinterpretation open.  How about "may not be consistent" instead of "are not"?

ronawho: Changed to:

"""
It's important to be aware that buffered atomic operations are not consistent with regular atomic operations and updates may not be visible until the buffers are explicitly flushed
"""

I like the phrasing that buffered atomics aren't consistent with non-buffered atomics, but I agree it's important to note that buffers will be flushed automatically when full. I'm going to merge now, but I'm open to other phrasing (and am planning on having people not familiar with the implementation take a look at the docs too.)

PR #10902, opened 2018-08-27
Make LLVM work with dynamic linking on a Cray XC system

dmk42:
Certain libraries must be explicitly specified for dynamic linking to work when using the PrgEnv compiler/linker wrapper.  This was specified in `Makefile.cray-prgenv` which was not accessible to the LLVM back end.  This change moves the required logic out to a shell script so it can be accessed both directly by a PrgEnv compilation, and indirectly for the benefit of an LLVM compilation.

### DIFF
--- a/runtime/src/comm/ugni/Makefile.share
+++ b/runtime/src/comm/ugni/Makefile.share
@@ -48,7 +48,7 @@ ifeq ($(CHPL_MAKE_COMPILER), clang-included)
 
     # only run gather-cray-prgenv-argumenst.bash once, as an optimization
     ifndef CHPL_UGNI_PRGENV_CFLAGS
-      export CHPL_UGNI_PRGENV_CFLAGS := $(shell $(CHPL_MAKE_HOME)/util/config/gather-cray-prgenv-arguments.bash compile)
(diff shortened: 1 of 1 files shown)


dmk42:
@ronawho - Would you please review this?  It's a compromise between specifying the libraries in a Makefile and specifying them in clangUtil.cpp.

util/config/gather-cray-prgenv-arguments.bash
```diff
@@ -19,6 +24,12 @@ case $1 in
     ;;
 esac
```
ronawho: Update the usage above? (and fail if all args aren't provided?)

ronawho:
This looks correct functionally speaking. I'll admit that I don't love the implementation, though I'm not quite sure what I don't love about it. I think I would prefer that we just look at the values of `CHPL_COMM` and friends in the new bash file, but I don't think those are set from the makefiles, only `CHPL_MAKE_COMM` and friends.

Maybe add some checks to the bash scripts to error out if some args aren't provided or something? I think what I worry about is that there's no checks for missing or invalid arguments.

util/config/gather-cray-prgenv-arguments.bash
dmk42: Yes, thanks for catching that.

dmk42:
I'm not a fan of the situation either, because `Makefile.cray-prgenv` needs to use `CHPL_MAKE_*` and `clangUtil.cpp` needs to use the plain `CHPL_*` variables.  That's why I passed them on the command line.

This is now ready for another look.

dmk42:
Thanks!

PR #10903, opened 2018-08-28
Guard do_remote_amo_nb if chained transactions aren't available

ronawho:
If chained transactions aren't available, don't declare do_remote_amo_nb()
since it wont' be used.


gbtitus:
Splendid!  👍

Issue #10856, opened 2018-08-22
Feature Request: Querying NUMA locality information without 'CHPL_LOCALE_MODEL=numa'


LouisJenkinsCS:
One thing that I would like to see would be the ability to query NUMA information from the `CHPL_LOCALE_MODEL=flat` locale model. Even if we do not give NUMA nodes their own locale, I believe that the user should be able to make NUMA-conscious decisions about the management of their data. One benefit of allowing querying of such information is that it allows for a better privatization strategy, as data can be kept not only for each compute-node, but each NUMA node as well. As a concrete example, the Chapel Aggregation Library (Issue #10386) can obtain significant performance improvements by eliminating false-sharing (issue #10771) and by keeping NUMA-specific destination buffers.

LouisJenkinsCS:
Accidentally hit enter and submitted early.

bradcray:
Can you be more precise about what kinds of NUMA information you'd like to query?

LouisJenkinsCS:
For starters, and this is really in my opinion all that is needed, is to determine the NUMA socket of the thread that the current task is running on. Something like `here.getNUMASocket()`. Something like this should be cached in task local or thread local storage on task creation.

bradcray:
Tagging @ronawho in hopes that he can weigh in on how easy / hard it would be to make such information available.

ronawho:
What are you going to do with that locality information once you have it? I would guess you'd want to be able to spawn a task to that same numa domain, which I think really means you just want to be able to use the numa locale model. IMO the best way to approach this would be to resolve the performance issues in the numa locale model, instead of trying to add something to flat.

Flat uses the nemesis scheduler, which is not numa-aware so you'd really have to hack things to get this sort of information and spawn to a specific numa domain.

Can you give us a concrete example of what you want to be able to do, maybe in pseudocode?

ronawho:
Oh, for that you might just be able to extern to an hwloc routine or `sched_getaffinity` or something. `sched_getaffinity` will give you CPU and not socket. hwloc is probably the more portable and robust solution, but I don't know that API calls you'd want to use off the top of my head.

LouisJenkinsCS:
One issue is the overhead associated with each call, which is why I earned it to cached in thread local variable

ronawho:
I don't know how expensive those calls are, but to experiment you could always write a C routine that caches the value to a variable declared with __thread and then extern that C routine.

LouisJenkinsCS:
[...]

PR #10907, opened 2018-08-28
Fix chpl_topo_getThreadLocality() for hwloc

ronawho:
Previously, it was using `hwloc_set_cpubind`, when it should have been using
`hwloc_get_cpubind`

Noticed while investigating https://github.com/chapel-lang/chapel/issues/10856


ronawho:
This function isn't currently used anywhere, but verified it worked with a simple test:

```chpl
extern proc chpl_topo_getNumNumaDomains(): c_int;
extern proc chpl_topo_getThreadLocality(): chpl_sublocID_t;

writeln(chpl_topo_getNumNumaDomains());
coforall 1..here.maxTaskPar {
  writeln(chpl_topo_getThreadLocality());
}
```

PR #10910, opened 2018-08-28
Sanity-check hugepage-related env vars when CHPL_COMM=ugni.

### LINKED ISSUE #10451: Warn if we detect HUGETLB_NO_RESERVE isn't set
We have launcher hooks for [ugni](https://github.com/chapel-lang/chapel/blob/master/runtime/src/comm/ugni/comm-ugni-launch.c) to set `HUGETLB_NO_RESERVE=yes` (required for numa affinity otherwise hugepages get faulted in allocation time) and `CHPL_JE_MALLOC_CONF="lg_chunk:24"` (which tells jemalloc to use chunk sizes that are a multiple of the hugepage.) lg_chunk:24 is for 16M hugepages and will change for different hugepage sizes.

If the launcher isn't being used or something we should detect that these env vars aren't set in ugni and issue a warning since they can have pretty large correctness implications. 

Semi-related to https://github.com/chapel-lang/chapel/issues/7841

gbtitus:
Print a warning if we don't seem to be using hugepages, because of the
potential performance loss.  Also print warnings if we do seem to be
using hugepages but HUGETLB_NO_RESERVE or CHPL_JE_MALLOC_CONF are
not set, since those are also important for performance and their
absence implies that the Chapel launcher (which always sets them) was
not used.

Note that we should really print an error instead of a warning if
CHPL_JE_MALLOC_CONF is set and specifies a chunk size larger than
the hugepage size, or is not set and jemalloc's default 2 MiB chunks are
larger than the hugepage size.  In those cases we can get misaligned
chunks, which will lead to segfaults or internal errors, so we're better
off just erroring out.  This is work for the future, though, because
getting all the needed information together early enough is going to
require some reorganization which I don't have time for right now.

As a side effect of the current effort I moved jemalloc_conf_ev_name()
from comm-ugni-launch.c to comm-ugni-heap-pages.h and made it an inline
static called chpl_comm_ugni_jemalloc_conf_ev_name(), because now I
needed the name of the jemalloc configuration env var in both the
launcher and the runtime proper.  Really it probably ought to be in a
file related to the jemalloc memory layer, but getting it there for the
launcher ended up looking like more work than I wanted to do just a
couple of days before feature freeze, so that will have to wait.

This closes #10451.


gbtitus:
I put @ronawho as primary, but @bradcray if you want to take a look that would be great also.

bradcray:
Is our test system rich enough to permit us to add tests that these warnings are generated when the environment is "just so"?  Seems like with the `.execenv` files it might be?

gbtitus:
It doesn't look like we can unset environment variables in `.execenv` files.  We can only set them there.  If there was a way to wrap the test execution we could do it, but I don't see that in the testing system or `chpl_launchcmd.py`.  @ronawho?

test/runtime/configMatters/comm/ugni/hugepage-sanity.wrap_real
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env bash
+
```
ronawho: Ooh, this is a clever way to test this!

doc/rst/platforms/cray.rst
```diff
@@ -530,6 +530,38 @@ the number of registered memory regions, so allowing a very high number of
 them may lead to reduced performance.
 
 
+ugni Hugepage-related Internal Warnings
```
ronawho: Drop "Internal"?

doc/rst/platforms/cray.rst
```diff
@@ -530,6 +530,38 @@ the number of registered memory regions, so allowing a very high number of
...
+   Communication performance with ugni is so much better with hugepages
+   that if you do not use them, the runtime will print the following
+   warning when a multilocale program starts:
+
+   .. code-block:: sh
+
+      warning: without hugepages, communication performance will suffer
+
+   You can quiet this warning by giving the ``--quiet`` or ``-q`` option
+   when you run the executable.
```
ronawho: Maybe change this to:
"""
To resolve this you can load a hugepage module with `module load craype-hugepages16M` and recompile, or if you really don't want to use hugepages you can quiet the warning by...
"""

doc/rst/platforms/cray.rst
```diff
@@ -530,6 +530,38 @@ the number of registered memory regions, so allowing a very high number of
...
+   .. code-block:: sh
+
+      warning: HUGETLB_NO_RESERVE should be set to something when using hugepages
+      warning: CHPL_JE_MALLOC_CONF should be set when using hugepages
+
+   As before, you can quiet these warnings by giving ``--quiet`` or
+   ``-q`` when you run.  However, you should be aware that at least the
+   second of these may presage actual trouble in the form of internal
+   errors or segfaults under certain circumstances.
+
```
ronawho: I think it'd be worth a quick note about what these env vars do and what users can set them to if they're not using a launcher.

doc/rst/platforms/cray.rst
gbtitus: Will do.  (You don't have to recompile, though, just rerun.)

doc/rst/platforms/cray.rst
gbtitus: Are you sure?  Not using a launcher is a bit of a hobbyist mode on an Aries-based Cray in the first place.  If you insist on this I'd be inclined to just add one sentence telling them to read the `libhugetlbfs(7)` and `jemalloc(3)` man pages and then the `comm-ugni.c` and `comm-ugni-launch.c` code.

bradcray: While I wish all users would use a launcher on Crays, it seems that they haven't in the past.  At times, this has been because the site has their own specific wrappers around the standard aprun/slurm scripts.  Note that NERSC's official Chapel instructions have users work around the launchers (for reasons that are unclear to me):  http://www.nersc.gov/users/computational-systems/edison/programming/chapel/

gbtitus: I'd forgotten about that.  I can write up something boilerplate.  They really just need to know what to set the env vars to.  They don't need to know the "why" behind the values.

gbtitus:
I updated the documentation.

PR #10916, opened 2018-08-29
Stop make(1) explicitly when the compileline utility fails.

### LINKED ISSUE #10542: bark when compileline fails during Qthreads build
In https://github.com/chapel-lang/chapel/pull/10390 we began using `compileline` to get compilation options for the Qthreads build.  We've seen this fail for a couple of different reasons, either because GASNet hadn't been built, or hadn't been built with the conduit we were expecting.  When this happens currently we get an obscure message, typically a complaint that some runtime include file cannot be found.  We should be more explicit and just say that `compileline` failed.  That will make it easier to track down what's gone wrong.

gbtitus:
The Qthreads build needs some of the runtime build -I and -D options,
which we get with the compileline utility.  But that can fail to run
correctly; the most common reason has been that one of the things it
needs to reference hasn't been built yet.  Such a failure doesn't halt
the make since compileline is run in a make 'shell' function invocation
to set a make variable, not in a target build command.  The make will
eventually fail, due to an error compiling some Qthreads source file
because the -I or -D options it needs aren't there.  But the root cause
is buried much earlier in the log, where the compileline invocation
failed, and the connection between the two isn't at all obvious.

We've solved the instances of this we've seen, by adding prerequisites
reflecting compileline's dependencies, but we may not have seen all of
them and certainly new dependencies lacking corresponding prerequisites
could be created in the future.  So here, force the make to halt right
away when the compileline invocation fails.

This resolves #10542.

### DIFF
--- a/third-party/qthread/Makefile
+++ b/third-party/qthread/Makefile
@@ -52,13 +52,20 @@ ifeq ($(CFLAGS_NEEDS_RT_INCLUDES), y)
   # do not want -DNDEBUG; if we have that set when building Qthreads it
   # seems to cause Chapel programs built with the result to hang during
   # exit when shutting down Qthreads.
-  CFLAGS += \
+  INCS_DEFS := \
     $(shell $(CHPL_MAKE_HOME)/util/config/compileline --includes-and-defines \
+            || echo compilelineFAILURE)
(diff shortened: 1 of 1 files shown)


gbtitus:
I tested this manually, since we don't currently have a situation where it can happen in the normal course of events.

@bradcray, you may be interested in this one as well.

PR #10919, opened 2018-08-29
Refactor PRIM_NEW resolution to enable promotion

### LINKED ISSUE #9772: problem with promoted new for generic class

benharsh:
This PR significantly changes mechanics of PRIM_NEW resolution in order to allow for promoted new-expressions. Prior to this PR, only concrete classes could be used with promoted new-expressions because of the internal "_new" wrapper. Now the cross product of ``[concrete|generic] x [class|record]`` is supported for promoted new-expressions.

The old PRIM_NEW handling was split across four functions based on concrete vs. generic and class vs. record. This implementation used different AST transformations in each of the four cases, making it difficult to maintain a mental model of the transformation. The new implementation uses a single function with shared logic with less differences between the four potential cases.

The _new wrappers are now generated on the fly during resolution, and are generated for both concrete and generic classes (formerly only concrete). A _new wrapper will also be generated for records if the new-expression requires promotion. The 'init' call itself cannot be promoted because it does not return a value.

This PR also significantly simplifies the implementation of "managers" (i.e. owned, etc.) in PRIM_NEW calls on types with initializers. Instead of poking around the AST after the "_new" call is generated, we insert AST as we go. The initializers for owned and shared are also updated in order to enable promotion, which may happen if a 'new owned' expression requires promotion.

Resolves #9772

Testing:
- [ ] full local + futures
- [x] gasnet + futures
- [x] memleaks


benharsh:

modules/internal/OwnedObject.chpl
```diff
@@ -129,12 +129,16 @@ module OwnedObject {
...
 
-      if !isClass(p) then
-        compilerError("Owned only works with classes");
+      this.p = p;
+    }
 
+    proc init(p:?T) where isClass(T) == false && isSubtype(T, _owned) == false  &&
+                    isIterator(p) == false {
+      compilerError("Owned only works with classes");
+      this.t = T;
```
lydia-duncan: Why is there code after the compilerError?  Shouldn't it never be executed?

benharsh: It won't, but the compiler doesn't really do any kind of unreachable code analysis today, so the regular initializer warnings/errors would apply if I didn't initialize the fields.

PR #10924, opened 2018-08-30
Get compiler to actually build the Python module from the library when possible

lydia-duncan:
Extends compiling with `--library-python` to build the Python module
using the generated files.

Details:
Builds the command to use when compiling the Python module based
on the known value of the PYTHONPATH, library and include directory
flags (including those obtained from require statements), and any
additional c flags needed.

Fixes a bug with the representation of c_strings in the .pxd file, and of
numpy types in the .pyx.  Fixes a bug when the generated library .a/.so
is named something that doesn't start with "lib" (on Mac OSX, there's
another issue for static compilation on Linux).

Makes some helper functions for getting the `require` statement libraries
and headers into a single long string.  Splits computing the library name for
use on the command line into a helper as well.  Extends and renames some
of the enum constants used to determine the corresponding Cython/Python
type, moving the numpy adjustment into getPythonTypeName.

Note: this doesn't adjust the compiler and linker used when building on a
Cray, but I do still intend to get to that.

Tests:
- Changes every python module test to use a python file that depends on
the generated module instead of just checking the generated .pxd/.pyx/.py
look the way we expect.
  - This required adding a prediff to eliminate some junk output from the
     Cython compile step.  That's a stop-gap, I'd like to find a way to avoid those
     messages in the first place.
  - It also required skipping these tests when python3, numpy, and cython are
     not installed in the testing environment.  Adding the appropriate testing
     environment will happen in a separate PR

Testing:
- Passed a full std/ paratest with futures (double checked that the python tests are
now skipped in normal settings, and run when their requirements are met)


compiler/codegen/library.cpp
```diff
@@ -483,15 +508,85 @@ static void makePYFile() {
...
   }
   // Don't "beautify", it will remove the tabs
   closeLibraryHelperFile(&py, false);
 }
 
+// Once all the python files have been generated and the .a/.so has been made,
+// make the Python module!
+void codegen_make_python_module() {
+  std::string getOldPythonPath = "echo $PYTHONPATH";
+  std::string pythonPath = runCommand(getOldPythonPath);
```
benharsh: Minor nit: should we use 'getenv' here instead?

compiler/codegen/library.cpp
```diff
@@ -483,15 +508,85 @@ static void makePYFile() {
...
+    pythonPath = CHPL_RUNTIME_INCL;
+  } else {
+    pythonPath.erase(pythonPath.length() - 1);
+    pythonPath += ":";
+    pythonPath += CHPL_RUNTIME_INCL;
+  }
+  pythonPath += "/python";
+
+  std::string getCFlags = "$CHPL_HOME/util/config/compileline --cflags";
+  std::string cFlags = runCommand(getCFlags);
```
benharsh: Should 'getCompilelineOption' be used instead?

lydia-duncan: getCompilelineOption also replaces long paths in the string with $(CHPL_HOME), etc, which aren't valid in a bash command (but are in the generated Makefile, where getCompilelineOption is primarily used).  I could split the portion prior to those cleanups into a separate function, but I was getting tired of rearranging functions so I didn't at the time.

benharsh: OK 👍

PR #10926, opened 2018-08-30
Fix how we determine the size of qthread_runtime_data_s

ronawho:
Previously, we just hardcoded the size to be less than 1 page, but it turns out
that it's actually larger than that on systems without a native swapcontext
(such as ARM.)

Instead of assuming data structure less than a page, add a mechanism to
determine the real size from qthreads and use that to compute how many pages to
reserve.

### DIFF
--- a/doc/rst/usingchapel/executing.rst
+++ b/doc/rst/usingchapel/executing.rst
@@ -290,9 +290,9 @@ automatically when the program or task starts executing, and remains in
 existence until it completes.  The default call stack size is ~8 MiB on
 Linux-based systems, since this is a common value for the process stack
 limit on such systems.  On Cygwin systems the default call stack size is
-~2 MiB. Note that up to 3 system pages of each stack may be reserved for
-use by the tasking layer. 1 page for runtime data structures and up to 2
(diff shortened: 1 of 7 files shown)


doc/rst/usingchapel/executing.rst
```diff
@@ -290,9 +290,9 @@ automatically when the program or task starts executing, and remains in
 existence until it completes.  The default call stack size is ~8 MiB on
 Linux-based systems, since this is a common value for the process stack
 limit on such systems.  On Cygwin systems the default call stack size is
-~2 MiB. Note that up to 3 system pages of each stack may be reserved for
-use by the tasking layer. 1 page for runtime data structures and up to 2
-additional pages if guard pages (--stack-checks) are enabled.
+~2 MiB. Note that up to 4 system pages of each stack may be reserved for
```
gbtitus: Maybe say this in a more generic way: "... some system pages of each stack may be reserved ..." and then "at least 1 page for runtime data structures and up to 2 additional pages if guard pages ...".  That should cover all expected future changes.

PR #10927, opened 2018-08-30
Add a compiler error for using '.type' on something that is already a type

daviditen:
Applying the .type operation to something that is already a type now results
in a compiler error instead of getting ignored in most cases.  This is handled
during preFold for PRIM_TYPEOF.  It needs to be handled during resolution so
that it can apply to uses of '.type' on function calls that return types.

Updated one spot in the modules where .type was used on a type.

Updated test/arrays/ferguson/runtime-type-type.good to expect the error
message instead of silently ignoring the .type operation. Removed the .future
because it now generates a nice error instead of an unresolved call error
about an internal function.

Added a test that tries to use '.type' on some simple types, and another that
uses '.type' on a function call that returns a type.

Update a few tests that used .type on types.

### DIFF
--- a/compiler/resolution/preFold.cpp
+++ b/compiler/resolution/preFold.cpp
@@ -923,9 +923,13 @@ static Expr* preFoldPrimOp(CallExpr* call) {
     call->replace(retval);
 
   } else if (call->isPrimitive(PRIM_TYPEOF)) {
-    Type* type = call->get(1)->getValType();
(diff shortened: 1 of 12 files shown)


daviditen:
Passed a full Linux64 paratest with futures.

daviditen:
Thanks @vasslitvinov!

Issue #10251, opened 2018-07-10
BLAS module cannot find lapacke.h when linking against Intel MKL


BryantLam:
### Summary of Problem
With the caveat that I don't know much about BLAS, LAPACK, or Intel MKL --

Compiling one of the test programs in `$CHPL_HOME/test/library/packages/BLAS` to use Intel MKL results in an error saying it can't find header file "lapacke.h".

Workaround: Symlink `$MKLPATH/lib/intel64/mkl_lapacke.h` to `$HOME/chpl/include/lapacke.h` and include that directory on the compile line.

I imagine the proper fix is to apply whatever logic `isBLAS_MKL` is using on `lapacke.h` as well.

### Steps to Reproduce

**Source Code:**
`$CHPL_HOME/test/library/packages/BLAS/test_blas1.chpl`

**Compile command:**
I modified the Makefile:
- Added `BLASFLAGS = -sisBLAS_MKL=true -I$MKLPATH/include -L$MKLPATH/lib/intel64 -I$HOME/chpl/include -lmkl_rt`
- Commented out LAPACKFLAGS
- Appended $(BLASFLAGS) to CHPLFLAGS

`chpl test_blas1.chpl --ccflags -Wno-enum-conversion -sisBLAS_MKL=true -I$MKLPATH/include -L$MKLPATH/lib/intel64 -I$HOME/chpl/include -lmkl_rt -o test_blas1`

### Configuration Information
`chpl version 1.18.0 pre-release (e5cdbf130e)`
Intel MKL 2017.0.3

### Priority
Not a high priority now that I have a workaround.

ian-bertolacci:
@ben-albrecht Did we have a fix for this?

ben-albrecht:
Yes - I have a fix for this, which adds a simple `-sisLAPACK_MKL` config param to `LAPACK` module. I will attempt to get that in soon.

Note that MKL deviates from the established interface for the low level `LAPACK` routines, so MKL users must rely only on the high level `LAPACKE` routines.

PR #10929, opened 2018-08-30
Enable usage of LinearAlgebra without BLAS/LAPACK

### LINKED ISSUE #8803: Enable LinearAlgebra usage without BLAS/LAPACK dependencies
Currently, using the `LinearAlgebra` library requires compiling with BLAS & LAPACK. This can be burdensome for users to install and find the right flags to compile with, and many of the `LinearAlgebra` features do not rely on these libraries. Ideally, users should be able to compile and run programs using the subset of `LinearAlgebra` that does not rely on BLAS & LAPACK.

There are 2 mechanisms of accomplishing this (that I can think of at the moment):

1. Have a `config param` that specifies the user will be using native implementations, which will fold the `require` statements away and throw compiler errors for functions that require BLAS or LAPACK.

```chpl
// BLAS.chpl
if !noRequire {
  if isBLAS_MKL {
    require "mkl_cblas.h";
  } else {
    require "cblas.h";
  }
}
```

```chpl
// LinearAlgebra.chpl

private proc _csrmatmatMult(A, B) where noRequire {
  compilerError('Dependency error: matrix multiplication requires BLAS');
}
```

2. Put the `use` statements inside of the function bodies for functions that depend on them.

```chpl
private proc _matmatMult(A: [?Adom] ?eltType, B: [?Bdom] eltType) {
  use BLAS;
  ...
  gemm(A, B, C, 1:eltType, 0:eltType);
  return C;
}
```

ben-albrecht:
This PR strives to allow users to use the `LinearAlgebra` module without having `BLAS` or `LAPACK` backends available on their system. Any routines that require `BLAS` or `LAPACK` will either have a Chapel native equivalent that is executed instead, or emits a compiler error that the routine is only available with `BLAS/LAPACK` supports.

Implementation-wise, this means moving all of the `require` statements into the function bodies within the `BLAS` and `LAPACK` module.

Additionally, there are a number of `config` interface changes described below - these changes were made to allow users to opt out of using `BLAS` and/or `LAPACK` and use the native implementations in `LinearAlgebra` if available:

### BLAS

- Deprecated `isBLAS_MKL`
- Two different exclusive flags for controlling which header to include
  - Added "higher-level" `config param blasImpl` with available values `{blas, mkl, none}`
  - Added "lower-level" `config param blasHeader` for overriding default values available in `blasImpl`
    - This flag will override `blasImpl` if both are set.

### LAPACK

- Two different exclusive flags for controlling which header to include
  - Added "higher-level" `config param lapackImpl` with available values `{lapack, mkl, none}`
  - Added "lower-level" `config param lapackHeader` for overriding default values available in `lapackImpl`
    - This flag will override `lapackImpl` if both are set.

### Other changes

- Split up `LinearAlgebra` tests into tests with and without dependencies
- Removed recursive `LinearAlgebra.skipif`, since no-dependency tests can now be run everywhere
- Removed skipif for `LinearAlgebralib.chpl` primer, and added a `compopts` for disabling BLAS/LAPACK.
- Added deprecation test for `isBLAS_MKL` flag
- Made minor documentation improvements for `BLAS`, `LAPACK`, and `LinearAlgebra`
  - Annotated all procedures that have BLAS/LAPACK dependencies
- Simplified Mason-Spack test to a linear algebra program that exercises BLAS and LAPACK, rather than maintain a full copy of the `LinearAlgebra` module. (FYI - @Spartee)

### Testing

- [x] OS X (openblas, lapack) - primer, BLAS, LAPACK, LinearAlgebra tests
- [x] OS X (no dependencies) - primer, LinearAlgebra tests
- [x] OS X - `mason/spack-integration`
- [x] linux64 - full testing
- [x] Cray - primer, BLAS, LAPACK, LinearAlgebra tests

Closes #8803, #10251


benharsh:
Did you look into using the script in 'util/misc/gen-LAPACK/' to generate the additional 'require' stuff?

ben-albrecht:
> Did you look into using the script in 'util/misc/gen-LAPACK/' to generate the additional 'require' stuff?

I did not. I used a simple sed command within vim to add the `require` statements.

Do you think it's important for us to keep the script updated with these changes?

ben-albrecht:
To my surprise, the generated code is not too far off from the current implementation:

https://gist.github.com/ben-albrecht/c72fd501ec2413511fc3708a1a712dc7

I think it would be feasible to update the script. However, taking a glance at the [script](https://github.com/chapel-lang/chapel/blob/master/util/misc/gen-LAPACK/extern-tool/Passes.py), I suspect it will be a nontrivial effort to grok the code and figure out how to make these changes.

Do you think we need to update the script for this PR, i.e. the 1.18.0 release?

daviditen:
What happens if somebody does `-sblasImpl=blas -sblasHeader=mkl_cblas.h`?  The `*Header` params seem like they might add more confusion than they're worth.  Would we ever want `-sblasImpl=blas -sblasHeader=someRandomUnexpectedHeader.h`?

ben-albrecht:
> What happens if somebody does `-sblasImpl=blas -sblasHeader=mkl_cblas.h`? 

Good question. `blasHeader` will override `blasImpl` if it is set. You could consider the flags exclusive. The idea is that a user should never have to set `blasHeader` manually, but the option exists in case a user is trying to use a BLAS implementation that uses a unique header name that we didn't account for.

> The `*Header` params seem like they might add more confusion than they're worth.  Would we ever want `-sblasImpl=blas -sblasHeader=someRandomUnexpectedHeader.h`?

I don't think so. As mentioned earlier, the flags are exclusive, so a combination would never make sense. If we think this is a potential common pitfall, we could add an enum field for `unset` which will allow us to emit a warning when both configs have been set by the user.

We could potentially remove the `blasImpl` flag and require users to specify the header explicitly if they are not using `cblas.h`. However, passing strings to config params is a little cumbersome, and requires the user to know the name of the header rather than just the implementation name, so I'm not a big fan of that being the only interface available:

```bash
# clean enum interface
chpl --blasImpl=mkl program.chpl
# string interface
chpl --blasHeader=\"mkl_cblas.h\" program.chpl
```

daviditen:
> We could potentially remove the blasImpl flag and require users to specify the header explicitly if they are not using cblas.h.

I was meaning to suggest going the other way - only allowing the clean enum interface and not allowing setting the header name because it probably won't work to use some random blas interface we aren't aware of anyway.

benharsh:
> I think it would be feasible to update the script. However, taking a glance at the script, I suspect it will be a nontrivial effort to grok the code and figure out how to make these changes.
> Do you think we need to update the script for this PR, i.e. the 1.18.0 release?

I think it would be OK to leave the script alone for now, but I think it would be good to address post-release.

ben-albrecht:
> I was meaning to suggest going the other way - only allowing the clean enum interface and not allowing setting the header name because it probably won't work to use some random blas interface we aren't aware of anyway.

I see. The netlib-defined BLAS/LAPACK standard interface seems pretty well-established. Anecdotally, I haven't encountered a BLAS implementation that does not work with our module yet. It would be a shame to not support an implementation due to us not having a unique header file name hard-coded into the module. 

That said, I think the header flag will be rarely needed. I wonder if we could make it less prominent in the documentation, such that someone wouldn't find it unless they were specifically trying to find it. For example, we could `pragma "no doc"` it and only mention that it exists in the documentation of the `blasImpl` / `lapackImpl` flags.

ian-bertolacci:
If you'd like, I can make those script modifications.

ben-albrecht:
> If you'd like, I can make those script modifications.

@ian-bertolacci - That would be greatly appreciated. I'll create an issue for this and ping you on it. Thanks!

PR #10931, opened 2018-08-30
Don't sanity check CHPL_JE_MALLOC_CONF with a fixed heap.

gbtitus:
The launcher doesn't set the MALLOC_CONF environment variable when we
have a fixed heap, because we don't need it in that case.  Therefore,
don't sanity-check it in that case either.


gbtitus:
@ronawho, would you take a look post-merge?

PR #10934, opened 2018-08-31
build_configs: improved setenv-3 behavior in response to make errors

awallace-cray:
build_configs.py: printchplenv results are added to the logfile
(after the Chapel make) for each configuration. You do not need
to run another "dry-run" after the build, just to save this info
(though you still may)

setenv-example-3.bash: if the setenv callback wants to skip building
an unwanted Chapel configuration, it should exit immediately instead
of setting-up an environment that will never be used for anything.
Not so much to save time, as to reduce confusing output.

setenv-example-3.bash now skips intel compiler Chapel configs, by
commenting-out one line; this makes the example setenv faster and
more portable. Support for the intel compiler is kept in the script,
to illustrate the method.


awallace-cray:
An easy PR here, just makes some experience-based minor improvements.

PR #10941, opened 2018-08-31
Ensure _RuntimeTypeInfo is initialized

vasslitvinov:
The array created out a for/forall/promoted expression used to
be created out of uninitialized _RuntimeTypeInfo in some cases
when the array's element type is a runtime type.

That is because the _RuntimeTypeInfo requires having at least
one element already available. Whereas the array is created
**before** the expression is executed, so there are no elements yet.

This change create stand-in _RuntimeTypeInfo to use when creating
the array. For example, if elements are domains, the _RuntimeTypeInfo
results in an empty domain. If elements are arrays, the _RuntimeTypeInfo
is for an empty array. And so on, recursively.

When the elements are/contain non-DefaultRectangular arrys/domains,
the compiler issues an error.

This is a workaround for invalid memory accesses,
while waiting for a worked-out design solution.

Testing
- [x] linux64
- [ ] valgrind
- [ ] gasnet multilocale tests
- [ ] linux64 futures


vasslitvinov:
@benharsh would you review?

modules/internal/ChapelIteratorSupport.chpl
```diff
@@ -89,7 +89,98 @@ module ChapelIteratorSupport {
...
+    // The other args are always compile-time only.
+    return chpl__buildDomainRuntimeType(defaultDist, domInst.rank,
+                                        domInst.idxType, domInst.stridable);
+  }
+
+  // Other kinds of arrays/domains are not supported.
+  proc chpl_buildStandInRTT(domInst) type
+  {
+    if domInst.type <= unmanaged BaseDom then
+      compilerError("for/forall/promoted expressions are supported only with Default Rectangular arrays, domains");
```
benharsh: This error message suggests that one cannot use loop/promoted-expressions with non-DefaultRectangular types. Isn't the actual issue that the element type must be DefaultRectangular?

benharsh: I'd also recommend adding a test for this error message.

PR #10942, opened 2018-08-31
Shape preservation with ranges

vasslitvinov:
This continues the work in #9994 and #10758 for shapes that are ranges.

Given open questions about the design used in the above, see #11220,
the range case ultimately falls back to the currently-shapeless route.
As a result, while the shape is preserved, there is no parallelism.
On the upside, the design issue in #11220 is side-stepped by
retaining the existing behavior in that regard, i.e. the runtime type
for the resulting array's element type is taken from the first element.

If we wanted to move this case to the shape-ful implementation,
see a refactoring of the shapeful initCopy() in 151c047.
That refactoring ensures no memory leaks and no double-deletions.
Except the standin domain leaks as before.

Update the tests that depended on the old behavior.
Added more checking to the prototype partial-reduction implementation.
While there, updated a comment.

Testing:
- [x] linux64 -futures
- [x] gasnet - multilocale tests
- [x] valgrind
- [x] memleaks


vasslitvinov:
@benharsh  would you review?

benharsh:
Oops, forgot to mention: I think a new test should be added for those error messages if tests don't exist already.

vasslitvinov:
Still not passing testing. Closing for now.
Will be back to it after the release.

vasslitvinov:
Resuming this work.

vasslitvinov:
@benharsh - I got new changes in 5f1feb1. How do they look to you?

>  I think a new test should be added for those error messages if tests don't exist already.

I think you were talking about the compilerError() additions [here](https://github.com/chapel-lang/chapel/pull/10942/files#diff-feddb5a31fea31a03458bbddb3164972l). How about adding tests for those when this code goes into production?

test/release/examples/benchmarks/shootout/fasta.chpl
```diff
@@ -93,7 +93,7 @@ proc repeatMake(desc, alu, n) {
         s = [i in 0..(r+lineLength)] alu[i % r]: int(8);
 
   for i in 0..n by lineLength {
-    const lo = i % r + 1,
+    const lo = i % r,
```
benharsh: Might want to let Brad know about this change so we can plan for the next re-submission.

PR #10945, opened 2018-09-03
Make performance tests make better use of managed classes

bradcray:
While running release-over-release testing, a number of performance
tests fail on old versions of the compiler due to using 'owned',
'unmanaged', etc.  This PR represents an attempt to (a) use
undecorated 'new's on performance tests in hopes of making them more
backwards compatible and (b) using more principled ownership types to
clean up the implementations and make them more representative of the
type of Chapel we'd like to be writing going forward.

A challenge to this approach is that using `new C()` without a decorator
generates a warning at present which adds a lot of noise when used in
that mode here.  We could handle this by (a) updating the tests to work
around those warnings (though that might break the release-over-release
testing as well), (b) throwing `--no-warnings` for these tests and having
those messages respond to the flag better (I have a prototype for this),
or (c) not throwing that warning by default.

This does not hit all such performance tests.  Some key missing ones
are: LCALS, CoMD which were bigger than I had time to take on in this
pass.

### DIFF
--- a/test/release/examples/benchmarks/hpcc/ptrans.chpl
+++ b/test/release/examples/benchmarks/hpcc/ptrans.chpl
@@ -62,10 +62,10 @@ proc main() {
   // Create Block-Cyclic distributions for both the Matrix and its
   // transpose:
   //
-  const MatrixDist = new unmanaged BlockCyclic(startIdx=(1,1),
+  const MatrixDist = new BlockCyclic(startIdx=(1,1),
                                      blocksize=(rowBlkSize, colBlkSize));
 
-  const TransposeDist = new unmanaged BlockCyclic(startIdx=(1,1),
+  const TransposeDist = new BlockCyclic(startIdx=(1,1),
                                         blocksize=(colBlkSize, rowBlkSize));
 
   //
--- a/test/release/examples/benchmarks/shootout/binarytrees.chpl
+++ b/test/release/examples/benchmarks/shootout/binarytrees.chpl
@@ -20,14 +20,15 @@ proc main() {
   //
   // Create the "stretch" tree, checksum it, print its stats, and free it.
   //
-  const strTree = new unmanaged Tree(strDepth);
-  writeln("stretch tree of depth ", strDepth, "\t check: ", strTree.sum());
-  delete strTree;
+  {
+    const strTree = new Tree(strDepth);
+    writeln("stretch tree of depth ", strDepth, "\t check: ", strTree.sum());
+  }
 
   //
   // Build the long-lived tree.
   //
-  const llTree = new unmanaged Tree(maxDepth);
+  const llTree = new Tree(maxDepth);
 
   //
   // Iterate over the depths in parallel, dynamically assigning them
@@ -39,9 +40,8 @@ proc main() {
     var sum = 0;
 
     for i in 1..iterations {
-      const t = new unmanaged Tree(depth);
+      const t = new Tree(depth);
       sum += t.sum();
-      delete t;
     }
     stats[depth] = (iterations, sum);
   }
@@ -57,7 +57,6 @@ proc main() {
   // Checksum the long-lived tree, print its stats, and free it.
   //
   writeln("long lived tree of depth ", maxDepth, "\t check: ", llTree.sum());
-  delete llTree;
 }
 
 
@@ -65,15 +64,15 @@ proc main() {
 // A simple balanced tree node class
 //
 class Tree {
-  var left, right: unmanaged Tree;
+  var left, right: owned Tree;
 
   //
   // A Tree-building initializer
   //
   proc init(depth) {
     if depth > 0 {
-      left  = new unmanaged Tree(depth-1);
-      right = new unmanaged Tree(depth-1);
+      left  = new owned Tree(depth-1);
+      right = new owned Tree(depth-1);
     }
   }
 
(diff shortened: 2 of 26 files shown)


test/release/examples/benchmarks/hpcc/ptrans.chpl
```diff
@@ -62,10 +62,10 @@ proc main() {
   // Create Block-Cyclic distributions for both the Matrix and its
   // transpose:
   //
-  const MatrixDist = new unmanaged BlockCyclic(startIdx=(1,1),
+  const MatrixDist = new BlockCyclic(startIdx=(1,1),
```
ronawho: ptrans isn't currently leaking (at least according to the nightly single node memleaks report.) That seems a little surprising to me considering these aren't deleted anywhere. Am I missing something obvious here?

bradcray: I was concerned about that too, but have not investigated (yet, at least).  TBH, I don't understand the current status of memory management for distributions, particularly in cases that explicitly allocate the classes like this.  My best guess would be that the value object that wraps the class deletes the object (which also seems problematic).  This  seemed sufficiently orthogonal to this effort that I hadn't looked into it as part of this effort.

test/release/examples/benchmarks/shootout/binarytrees.chpl
```diff
@@ -20,14 +20,15 @@ proc main() {
   //
   // Create the "stretch" tree, checksum it, print its stats, and free it.
   //
-  const strTree = new unmanaged Tree(strDepth);
-  writeln("stretch tree of depth ", strDepth, "\t check: ", strTree.sum());
-  delete strTree;
+  {
```
ronawho: The new scope isn't an obvious win to me over the manual delete, but I'm not opposed if you prefer it. Does the comment above need to be updated? Technically it will still be freed, but there's no longer an explicit line to associate that comment with

test/release/examples/benchmarks/shootout/chameneosredux-fast.chpl
```diff
@@ -35,9 +35,6 @@ proc main() {
 
   print(group1);
   print(group2);
-
-  for c in group1 do delete c;
-  for c in group2 do delete c;
```
ronawho: Should this be removed? Looks like they're still unmanaged above. Oh, maybe missed in the reversion in 3d822dd55ddb389d303719813fabd92262bba6e0?

bradcray: I tried this, but got a deadlock which was surprising to me.  I just sent you and @mppf mail about it in hopes that he/you might have insights.

test/studies/hpcc/STREAMS/elliot/stream-spmd-barrier.compopts
```diff
@@ -1 +1,2 @@
-../../common/probSize.chpl
+../../common/probSize.chpl --no-warnings
+
```
ronawho: Extra line? Probably not a big deal, but I'm not 100% sure that start_test will ignore this line (vs. treat it as a second compopts)

test/release/examples/benchmarks/shootout/binarytrees.chpl
bradcray: I prefer it slightly from the delete-free perspective.  I agree that the text of the comment could be clarified a bit...

test/release/examples/benchmarks/hpcc/ptrans.chpl
ronawho: I had similar confusion and issues in https://github.com/chapel-lang/chapel/pull/10511. Can run this test with --no-local (or gasnet) as a sanity check

test/release/examples/benchmarks/shootout/chameneosredux-fast.chpl
ronawho: Sorry --I meant that it's still using unmanaged, but you removed the deletes (i.e. I think we're leaking now.)

bradcray: Oh, I see.  Huh, could've sworn that I added those back in.  Maybe I forgot to git add...  Thanks!  (my memleaks run hasn't finished yet, but will presumably show this too).

bradcray:
@mppf may want to give this a look from a general delete-free ownership perspective.  @daviditen may want to look at one of his tests that I changed (not that I think it's controversial).  If I merge tonight as hoped (to get testing on it earlier), post-merge looks may still be useful.

test/release/examples/benchmarks/hpcc/ptrans.chpl
bradcray: We were right to be concerned.  While this passed for me on Mac, on linux64, it segfaulted and got a valgrind error.  I ended up backing out this change and opening issue #10993 to capture this oddness.

test/release/examples/benchmarks/shootout/chameneosredux-fast.compopts
```diff
@@ -0,0 +1 @@
+--no-warnings
```
mppf: Maybe this and the other .compopts for chameneous should have newlines at the end...

bradcray: Did this in PR #11010.

Issue #10615, opened 2018-07-31
Proposal for enabling assignment-based initialization: init=


benharsh:
See #8242 for original discussion/inspiration for this proposal.

We want authors of types to permit initialization from certain types in a normal variable declaration, like:

```chpl
var x : MyIntWrapper = 30;
```

But we don't want to simply invoke a regular initializer in this case (e.g. new MyIntWrapper(30)). For example, a "MyVector" type's initializer could accept an integer to represent length, but probably doesn't want to allow initialization from a statement like:

```chpl
var x : MyVector = 20;
```

In this proposal a normal user initializer ("proc init(...)") would only be invoked with a ``new`` expression.

This proposal also would introduce a new language initialization construct: ``init=``. ``init=`` would allow for initialization with or without a ``new`` expression:
```chpl
record R {
  var x : int;
  proc init=(x : int, y = 1) {
    this.x = x * y;
  }
}
var x = new R(10); // init=
var y : R = 10; // init=
var z : R = new R(10, 2); // init=
```

Furthermore, regular ``init`` methods would no longer be considered for copy-initialization:
```chpl
record Q {
  var x : int;
  proc init(x:int) {
    this.x = x;
  }
  proc init(r:Q) {
    writeln("init(Q)");
    this.x = r.x;
  }
  proc init=(r:Q) {
    writeln("init=(Q)");
    this.x = r.x;
  }
}

var A = new Q(10);
var B = A; // prints "init=(Q)"
```

This means that the compiler would no longer emit a warning for completely-generic ``init`` methods. If the user did not write a ``init=`` method, the compiler would generate one for them (like current copy-initializers).

The compiler will will consider ``init`` to be 'more specific' than ``init=`` for the following statements:

```chpl
// using record 'Q' from previous example
var A : Q; // declaration without initialization expression
new Q(10); // new-expr
new Q(someOtherQ); // NB: possibly confusing, may seem like a copy-constructor call
this.init(...); // sibling initializer call
```

We think this proposal would help to eliminate special-case code for wrapper types like 'bigint', and could also enable elegant initialization of atomic variables:
```chpl
var x : atomic int = 5; // x.write(5);
```

### Open Questions

1) Should we allow a completely generic formal for ``init=`` ?

bradcray:
I continue to really like this proposal (barring lurking surprises that we haven't anticipated), but I proposed it, so that's potentially not very interesting.

> Should we allow a completely generic formal for init= ?

[...]

PR #10953, opened 2018-09-04
Don't allow isSubtype/<= between type and value

mppf:
Because normalization of type query arguments relies on
a PRIM_IS_SUBTYPE with a mix of type and values, this PR adds
a new primitive, PRIM_IS_SUBTYPE_ALLOW_VALUES, which has the
old behavior, and uses it for type query arguments and legacy :
in where clauses. In the future, normalization should be updated
and PRIM_IS_SUBTYPE_ALLOW_VALUES should be removed.

Some module code updated to use .type to get the appropriate type.

exprAnalysis doesn't handle PRIM_IS_SUBTYPE anymore, since it
should be removed by the time it runs.


mppf:
@bradcray - yep, I've noticed this before (and tried to point it out on the init= issue). I think it's a good benefit of init=.

mppf:
https://github.com/chapel-lang/chapel/issues/10615#issuecomment-415115934 if you were wondering.

PR #10957, opened 2018-09-04
Make --no-warnings squash the messages printed out for undecorated `new`

bradcray:
While working on converting code over to use new delete-free features, I found that I wanted/needed some way to turn off the messages that the compiler prints for undecorated `new C()` calls.  `--no-warnings` turned off the first line of the message, but not the rest of it since they were essentially stylized print statements.  This PR wraps the entire message in a test to check whether if `--no-warnings` is on to give a means of squashing it.

@mppf and I discussed a few other ways of squashing these such as (a) relying on --warn-unstable or (b) introducing a new compiler flag.  We decided to go with this approach so that it could be used in performance tests in a way that would work with old versions of the compiler.


bradcray:
@mppf: This is what I mentioned on the phone and am running testing on now.  Let me know if this looks blatantly wrong to you.

bradcray:
This passed linux64 testing.

mppf:
I wonder if --warn-special should be involved. I don't know if it's there for older versions of the compiler though.

```
**--[no-]warn-special**

    Enable [disable] all special compiler warnings issued due to syntax and
    other language changes. Currently, these include
    --[no-]warn-domain-literal and --[no-]warn-tuple-iteration.
```

I think "special warnings" is a pretty weird term and would probably prefer to remove `--warn-special` in favor of something more descriptive, but it does already exist...

bradcray:
I'm disinclined to tie this into `--warn-special` for the following reasons:  (a) I wasn't familiar with that flag or what it means off the top of my head; (b) when I'm acting like a user, get a warning, and don't want to see it anymore, I instinctively reach for `--no-warnings` and am disappointed when it doesn't make the warning go away; (c) it seems to be off by default, but I think we want these warnings on by default (?).  But maybe you have a different way of tying it to `--warn-special` in mind that I'm not anticipating here?

mppf:
Right, I misread that it's off-by-default. Ignore me!

bradcray:
If you're not familiar with the `--help-settings` flag, it can help figure out the defaults without reading the source code!

PR #10970, opened 2018-09-04
LinearAlgebra.svd()

### LINKED ISSUE #10043: Implement LinearAlgebra.svd()
This would simply be a matter of wrapping [`LAPACK.*GESDD`](http://www.netlib.org/lapack/lug/node32.html) in `LinearAlgebra`.

This is necessary to get competitive PCA performance.

ben-albrecht:
This PR implements singular value decomposition and adds a test for it.

The interface is fairly simple:

```chpl
// Decompose A such that A = U*s*Vt
var (U, s, Vt) = svd(A);
```

We may want to add additional arguments that allow in-place computation, omitting computation of `U` and `Vt`, and using the more performant `gesdd` LAPACK backend. However, this seems like a good first pass at supporting `svd` in such a way that adding these features won't be backwards breaking.

Resolves #10043


test/library/packages/LinearAlgebra/correctness/testSVD-error.good
```diff
@@ -0,0 +1,3 @@
+uncaught LinearAlgebraError: LinearAlgebra error : SVD computation did not converge. Returned non-zero value for "info":-6
```
ronawho: `Returned non-zero value for "info` -- "info" doesn't mean anything to me in this context and seems more like it's exposing an internal name?  If it's a common name and means something to LA folks, I'm fine with keeping it

ben-albrecht: Yeah, that's a good point. I can translate the value to the meaning and try to give a better error message depending on the value.

PR #10973, opened 2018-09-04
Rename p field in owned/shared to chpl_p

### LINKED ISSUE #10943: Invoking an overloaded method on a managed type causes compiler error
**MWE**

```chpl
class C {
   proc init() {}
   proc p(x) { }
 }

 class D : C {
   proc init() {
     super.init();
   }
   override proc p(x) { }
 }

 var d = new owned D();
 compilerWarning(d.type : string);
 d.p(1);
```

**Output**

```
mwe.chpl:14: warning: _owned(D)
mwe.chpl:15: error: unresolved access of 'D' by '(1)'
```

Note it will work if I make it `new unmanaged D` or `new borrowed D`.

@mppf - When you get back from vacation

mppf:
Resolves #10943.

- [ ] full local testing


modules/internal/OwnedObject.chpl
```diff
@@ -257,14 +257,15 @@ module OwnedObject {
   // Don't print out 'p' when printing an _owned, just print class pointer
```
daviditen: 'p' -> 'chpl_p'

modules/internal/SharedObject.chpl
```diff
@@ -305,13 +307,13 @@ module SharedObject {
   // Don't print out 'p' when printing an Shared, just print class pointer
```
daviditen: 'p' -> 'chpl_p'

PR #10978, opened 2018-09-05
See how converting from 'unmanaged' to 'owned' affects chameneos perf

bradcray:
It's a nice code cleanup to use 'owned'/'borrowed' rather than 'unmanaged'
classes in chameneos.  Checking it in to see how it impacts nightly
performance.

(Note: I wanted to use an undecorated class allocation for the
meetingPlace type, but this caused a memory leak to my surprise.
Using explicit owned or borrowed does not).


bradcray:
@ronawho: Any objections?

ronawho:
FWIW in this context I can't imagine there being any performance difference (though chameneos is weird and sensitive to minor changes.)

It might be worthwhile to add an allocation microbenchmark to check differences in unmanaged vs. owned vs. borrowed vs. shared

PR #10980, opened 2018-09-05
Fix leak for default new with --no-warnings

mppf:
PR #10957 disabled the warning for `new SomeClass` if --no-warnings
is provided. But, it had an error that caused the allocation to leak
in that situation, even though it should be equivalent to
`new borrowed SomeClass`.

This PR fixes that problem and adds a test to catch it if it ever
happens again.

- [x] full local testing


bradcray:
D'oh!  So sorry about this...  I didn't catch that there was a non-warning line at the bottom of that executable (obviously).  This change looks good to me.

PR #10983, opened 2018-09-05
Fix Misc Functions docs to include < etc on types

mppf:
Adds chpldocs for `<` `<=` `>` and `>=` on types and
mentions that these operators are available
in `isSubtype` and `isProperSubtype`.

Note that Sphinx changes will be required for these
operators to be rendered with the right colors in HTML.


ben-albrecht:
I will handle the PR for updating the regex in https://github.com/chapel-lang/sphinxcontrib-chapeldomain  to support syntax-highlighting functions.

PR #10984, opened 2018-09-05
Make -o flag prepend "lib" in --library compilation

### LINKED ISSUE #10923: Static compilation of renamed library file with --library-python unable to find symbols
When compiling a library and Python module using `-o`, if the compilation mode is static we encounter an issue where loading the generated module fails because it can't find chpl_filenameTable.  chpl_filenameTable is an array declared in `chplcgfns.h` and defined in `chpl_compilation_config.c` (which is generated as part of compilation).  Dynamic compilation does not encounter this issue.

TODO:
- [x] check if naming the library something with lib in the front (that isn't the original name) also impacts this

Impacted test: test/compflags/lydia/library/python/checkExplicitLibname.chpl

### LINKED ISSUE #10938: Alter behavior of -o flag in library compilation to prepend lib to final .so/.a
Today, `chpl -o foo --library foo.chpl` will generate `foo.a` or `foo.so` depending on whether you are compiling dynamically or statically.  This seems to work fine for regular C compilation (as long as the exact path to the library is used), and the dynamic version works fine when making a Python module from that library.  However, the static version runs into issues when making a Python module - certain symbols defined in the library itself can't be found (see #10923) when we try to import the generated module.

Since the typical strategy is to name libraries with a `lib` prefix (and `chpl --library foo.chpl` also generates a `libfoo.a` or `libfoo.so`), we'd like to propose changing the behavior of the -o flag to add a `lib` prefix in `--library` compilation, to give the least surprising behavior.

lydia-duncan:
Prior to this change, we would always prepend "lib" to libraries created without
the `-o` flag.  Libraries created with the `-o` flag would use exactly the name
given (plus the relevant extension for static or dynamic code).  This was fine
for using these libraries with C, but Cython expected every library to start
with "lib", especially when compiling statically.

To avoid confusion in --library-python compilation (and follow the unwritten
rules about libraries), we decided to update compilation with `-o` to also
prepend "lib" to the name (if it does not already start with "lib").  This does
mean that the chapel compiler will never create a foo.a/foo.so file but the
user can always rename the generated library after the fact if it is that
important (and they are more likely to be aware of its risks).

Update a few tests that were looking for the "lib"-less version, and resolve the
Cython future now that it works.

Resolves #10938 
Resolves #10923 

Tested on our box where Cython testing runs and passed a full paratest with
futures

### DIFF
--- a/compiler/codegen/codegen.cpp
+++ b/compiler/codegen/codegen.cpp
@@ -2166,9 +2166,8 @@ static void setupDefaultFilenames() {
       if (strlen(filename) >= sizeof(executableFilename) - 3) {
         INT_FATAL("input filename exceeds executable filename buffer size");
       }
-      strcpy(executableFilename, "lib");
-      strncat(executableFilename, filename,
-              sizeof(executableFilename)-strlen(executableFilename)-1);
+      strncpy(executableFilename, filename,
+              sizeof(executableFilename)-1);
 
       if (fLibraryPython && pythonModulename[0] == '\0') {
         strncpy(pythonModulename, filename, sizeof(pythonModulename)-1);
--- a/compiler/codegen/library.cpp
+++ b/compiler/codegen/library.cpp
@@ -117,22 +117,11 @@ static void setupMakeEnvVars(std::string var, const char* value,
 }
 
 static void printMakefileIncludes(fileinfo makefile);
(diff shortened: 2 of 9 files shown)


lydia-duncan:
Thanks!  We discussed it in #10938, which I put before the group on Friday

PR #10985, opened 2018-09-05
Alphabetize expandVars, and add documentation link to Path Manipulations header

lydia-duncan:
I noticed we hadn't added expandVars to the operations categorization quick-link
when looking for a documentation link for the new function.  Rectify that, and
move it so that it is in alphabetical order in the file.

Passed a full paratest with futures, also verified the generated documentation
output

### DIFF
--- a/modules/standard/Path.chpl
+++ b/modules/standard/Path.chpl
@@ -41,6 +41,7 @@
 
    Path Manipulations
    ------------------
+   :proc:`expandVars`
    :proc:`joinPath`
    :proc:`splitPath`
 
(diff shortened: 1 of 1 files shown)


modules/standard/Path.chpl
```diff
@@ -235,6 +236,77 @@ proc dirname(name: string): string {
   return splitPath(name)[1];
 }
 
+/* Expands any environment variables in the path of the form `$<name>` or
+   `${<name>}` into their values.  If <name> does not exist, they are left
+   in place. Returns the path which includes these expansions.
+
+   :arg path: a string representation of a path, which may or may not include
+                   `$<name>` or `${<name>}`.
```
ben-albrecht: double back tics would be more appropriate here.

modules/standard/Path.chpl
```diff
@@ -235,6 +236,77 @@ proc dirname(name: string): string {
   return splitPath(name)[1];
 }
 
+/* Expands any environment variables in the path of the form `$<name>` or
+   `${<name>}` into their values.  If <name> does not exist, they are left
```
ben-albrecht: double back tics on `<name>`

PR #10992, opened 2018-09-06
Task-private variables in the spec

vasslitvinov:
Add the specification of task-private variables.

Given that the with-clause and its syntax \sntx{task-intent-clause}
are defined for task intents i.e. in the Task Parallelism chapter,
update the corresponding pieces there as well.

Introduce the term "shadow variable" as the thing that is referenced
implicitly in the forall loop body.

Add some more index entries.

I put TPVs in their own subsection of the spec because I felt
that the reader will find them more easily that way, than if
if they were defined inside the "Forall Intents" subsection.

Verified that the new example passes under start_test.

### DIFF
--- a/spec/Data_Parallelism.tex
+++ b/spec/Data_Parallelism.tex
@@ -233,8 +233,11 @@ writeln(result);
 \section{Forall Intents}
 \label{Forall_Intents}
 \index{forall intents}
+\index{shadow variables}
 \index{data parallelism!forall intents}
+\index{data parallelism!shadow variables}
 \index{statements!forall@\chpl{forall}!forall intents}
+\index{statements!forall@\chpl{forall}!shadow variables}
 
 If a variable is referenced within the lexical scope of a
 forall statement or expression and is declared outside
@@ -244,11 +247,12 @@ for task-parallel constructs. That is, the variable is considered
 to be passed as an actual argument to
 each task function created by the object or iterator leading
 the execution of the loop. If no tasks are created,
-it is considered to be an actual argument to the leader
+it is considered to be an actual argument to the leader or standalone
 iterator itself. All references to the variable
 within the forall statement or expression implicitly refer
-to the corresponding formal argument of the task function
-or the leader iterator.
+to a \emph{shadow variable}, i.e.
+the corresponding formal argument of the task function
+or the leader/standalone iterator.
 
 Each formal argument of a task function or iterator has the default
 intent by default.  For variables of primitive, enum, and class
@@ -262,8 +266,8 @@ with that intent in the optional \sntx{task-intent-clause}.
 For example, for variables of most types, the \chpl{ref} intent allows
 the body of the forall loop to modify the corresponding original
 variable or to read its updated value after concurrent modifications.
-The \chpl{in} intent is a way to obtain task-private variables
-in a forall loop.
+The \chpl{in} intent is an alternative way to obtain task-private
+variables (\rsec{Task_Private_Variables}).
 A \chpl{reduce} intent can be used to reduce values across iterations
 of a forall or coforall loop.
 Reduce intents are described in the \emph{Reduce Intents} technical note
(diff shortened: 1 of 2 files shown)


spec/Data_Parallelism.tex
```diff
@@ -280,9 +284,110 @@ \section{Forall Intents}
...
+\index{statements!forall@\chpl{forall}!task-private variables}
+\index{statements!forall@\chpl{forall}!shadow variables}
+
+A \emph{task-private variable} declared in a forall loop results
+in a separate shadow variable in each task created by the forall
+loop's parallel iterator, as well as a separate shadow variable
+created at the top level of the parallel iterator itself.
+
+The shadow variable is created at the start of the task or the
+iterator and destroyed at the end of the task/iterator.
```
benharsh: I think you can just say 'task' and not mention iterator here, since we already state in the previous paragraph that forall loops create tasks.

spec/Data_Parallelism.tex
```diff
@@ -280,9 +284,110 @@ \section{Forall Intents}
...
+Within the lexical scope of the forall statement or expression,
+the variable name references the shadow variable in the task
+that executed the current yield statement, or in the iterator
+if the current yield occurs in the iterator outside any task constructs.
```
benharsh: This sentence was confusing to me at first. What if you had a sentence or two dedicated to the concept of a "top-level" shadow variable, after which you can dedicate a sentence to the "or in the iterator" part of the sentence I'm commenting on?

spec/Data_Parallelism.tex
```diff
@@ -280,9 +284,110 @@ \section{Forall Intents}
...
+\end{craychapel}
+
+\begin{chapelexample}{task-private-variable.chpl}
+In the following example, the \chpl{writeln()} statement will observe
+the first shadow variable 4 times: twice each for the yields
+{\tt "before coforall"} and {\tt "after coforall"}.
+An additional shadow variable will be created and observed twice
+for each of the three \chpl{coforall} tasks.
+\begin{chapel}
+var cnt: atomic int;                     // count our shadow variables
```
benharsh: Should we print ``cnt`` at the end?

spec/Data_Parallelism.tex
```diff
@@ -280,9 +284,110 @@ \section{Forall Intents}
...
+shadow var: 0  yield: after coforall
+shadow var: 0  yield: before coforall
+shadow var: 0  yield: before coforall
+shadow var: 1  yield: inside coforall
+shadow var: 1  yield: inside coforall
+shadow var: 2  yield: inside coforall
+shadow var: 2  yield: inside coforall
+shadow var: 3  yield: inside coforall
+shadow var: 3  yield: inside coforall
+\end{chapeloutput}
```
benharsh: Should we include this output in the spec? It could be useful to have a comment matching the integer IDs to lines in the example, like:
```chpl
for 1..2 do
  yield "before coforall"; // shadow var 0
```

spec/Data_Parallelism.tex
vasslitvinov: I don't feel it would illustrate anything.

spec/Data_Parallelism.tex
vasslitvinov: The output does not look great because it needs to be sorted so the first two lines happen to be the last two lines actually printed by the program.

vasslitvinov:
@benharsh - I responded to two of your suggestions and implemented the rest. How is it looking now?

spec/Data_Parallelism.tex
```diff
@@ -280,9 +284,114 @@ \section{Forall Intents}
...
+for each of the three \chpl{coforall} tasks.
+\begin{chapel}
+var cnt: atomic int;                     // count our shadow variables
+record R { var id = cnt.fetchAdd(1); }
+
+iter myIter() { yield ""; }              // serial iterator, unused
+
+iter myIter(param tag) where tag == iterKind.standalone {
+  for 1..2 do
+    yield "before coforall";             // shadow var 0
```
benharsh: Maybe state this is the "top-level" shadow variable, for clarity?

PR #11002, opened 2018-09-06
Document improvements to type query syntax in spec

mppf:
Added a new section on partially generic arguments since
it was getting long.

Moved bit about integral up to generic types section (it was in the wrong
section).

Relates to PR #10135.

### DIFF
--- a/spec/Generics.tex
+++ b/spec/Generics.tex
@@ -196,10 +196,34 @@ writeln(x.type:string);
 \end{chapeloutput}
 \end{chapelexample}
 
+\begin{chapelexample}{query.chpl}
+Type queries can also be used to constrain the types of other function arguments
+and/or the return type.  In this example, the type query on the first argument
+establishes type constraints on the other arguments and also determines the
+return type.
+
+The code
+\begin{chapel}
+writeln(sumOfThree(1,2,3));
+writeln(sumOfThree(4.0,5.0,3.0));
+
+proc sumOfThree(x: ?t, y:t, z:t):t {
+   var sum: t;
+
+   sum = x + y + z;
+   return sum;
+}
+\end{chapel}
+produces the output
+\begin{chapelprintoutput}{}
+6
+12.0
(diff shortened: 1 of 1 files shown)


spec/Generics.tex
```diff
@@ -246,26 +278,56 @@ \subsection{Formal Arguments of Generic Type}
...
+  type indexType;
+  type containerType;
+}
+class Container {
+  type containedType;
 }
 \end{chapel}
-then the function definition
+Here, the function \chpl{f} can only apply when the
+\chpl{c.elementType==real}:
```
vasslitvinov: I suggest moving the above statement **after** the definition of `f`, removing "Here," and changing the earlier "Given the class definitions" with "Given the code:" This brings more code together, which I prefer.

spec/Generics.tex
```diff
@@ -246,26 +278,56 @@ \subsection{Formal Arguments of Generic Type}
...
-proc f(c: C(?t,real)) {
-  // body
+proc f(c: C(real,?t,?u)) {
+  // ...
 }
 \end{chapel}
-specifies that \chpl{f} should be chosen only when \chpl{c.tt==real}.
-\end{example}
+
+And the function \chpl{g} includes a nested generic type and can only apply
```
vasslitvinov: Let us drop "And" . It will still be readable without it, I think.

spec/Generics.tex
```diff
@@ -277,10 +339,33 @@ \subsection{Formal Arguments of Generic Type}
 \end{chapel}
 specifies that \chpl{tuple.size == 2 && tuple(2).type == real}.
 \end{example}
+
+Homogenous tuple arguments of queried type are also supported:
```
vasslitvinov: To me "queried" (in the above line) means there is a "?". The example below does not have that. Would you reconcile?

mppf:
@vasslitvinov - I've tried to take into account all the feedback. How does this version look?

PR #11008, opened 2018-09-06
Convert COMPOPTS to .compopts

bradcray:
Unbeknownst to me, COMPOPTS options are applied to both C and Chapel
tests, so my addition of `--no-warnings` to COMPOPTS here was
inadvertantly being applied to C compilers which (a) wasn't my intention
and (b) broke compilers that don't recognize that flag (like PGI).

Here, I'm converting to .compopts files for each test that needed it.
Happily I didn't need .perfcompopts files because these tests aren't
sensitive to additional warnings being added to them.


bradcray:
@dmk42 : This should resolve the PGI failure I introduced last night.

PR #11010, opened 2018-09-06
Switch from 'unmanaged' chameneos to 'owned' chameneos

bradcray:
Here, I'm switching from 'unmanaged' chameneos to 'owned' in order
to get the automatic deallocations and because we haven't seen
significant performance impacts in other cases that have done so.

I originally tried to make both these cases 'borrowed' incorrectly,
which the compiler didn't catch and led to various problems.  As
a follow up to this PR, I will file issues against those cases.

I also added missing EOF linefeeds to *compopts files as @mppf requested.


bradcray:
@ronawho, @mppf: Would one of you have time to review this?

bradcray:
Wow, you reviewed faster than Travis!

PR #11012, opened 2018-09-06
rename the spec section Where Expressions -> Where Clauses

mppf:
I just noticed the title and found it odd because we are
pretty consistent about calling them where clauses.


bradcray:
I like it.  I didn't check to see if there were other references that would need to be updated.

mppf:
Grep says: It's referred to by CHANGES.md for version 1.15. The spec wouldn't build if there was a cross-reference issue within the document.

bradcray:
Sounds good.  I was thinking of other uses of the term rather than LaTeX cross-references, but it sounds like your grep didn't find any there.

PR #11014, opened 2018-09-06
Add override keyword to spec

mppf:
* linkage-specifier now includes override, extern, export
* note that intents much match in overriden methods
* note that override keyword is now required on overriding methods

Relevant to PRs #10752 #9724.

### DIFF
--- a/spec/Classes.tex
+++ b/spec/Classes.tex
@@ -310,7 +310,7 @@ particular, dynamic dispatch will be used when the method receiver has a
 static type of the base class but refers to an instance of a derived
 class type.
 
-In order to have identical signature, two methods must have the same the
+In order to have identical signatures, two methods must have the same the
 names, intents, types, and order of formal arguments. The return type of
 the overriding method must either be the same as the return type of the base
 class's method or be a subclass of the base class method's return
@@ -319,7 +319,7 @@ type.
 Methods that override a base class method must be marked with the
 \chpl{override} keyword in the \sntx{procedure-kind}. Additionally,
 methods marked with \chpl{override} but for which there is no parent class
-method with an identical signature will cause an error.
+method with an identical signature will result in a compiler error.
 
 \begin{rationale}
   This feature is designed to help avoid cases where class authors
--- a/spec/Classes.tex
+++ b/spec/Classes.tex
@@ -309,10 +309,13 @@ base class's method.  Such a method is a candidate for dynamic
 dispatch in the event that a variable that has the base class type
 references an object that has the derived class type.
 
-The identical signature requires that the names, types, and order of
+The identical signature requires that the names, intents, types, and order of
 the formal arguments be identical. The return type of the overriding
 method must be the same as the return type of the base class's method,
 or must be a subclass of the base class method's return type.
+
+Methods that override a base class method must be marked with the
+\chpl{override} keyword in the \sntx{linkage-specifier}.
 
 Methods without parentheses are not candidates for dynamic dispatch.
 \begin{rationale}
--- a/spec/Classes.tex
+++ b/spec/Classes.tex
@@ -303,19 +303,31 @@ defined at this time.
 \index{dynamic dispatch}
 \index{methods!base class!overriding}
 
-If a method in a derived class is declared with a
-signature identical to that of a method in a base class, then it is said to override the
-base class's method.  Such a method is a candidate for dynamic
-dispatch in the event that a variable that has the base class type
-references an object that has the derived class type.
+If a method in a derived class is declared with a signature identical to
+that of a method in a base class, then it is said to override the base
+class's method. Such methods are considered for dynamic dispatch.  In
+particular, dynamic dispatch will be used when the method receiver has a
+static type of the base class but refers to an instance of a derived
+class type.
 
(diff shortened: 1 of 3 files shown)


mppf:
@bradcray - could you have a look at this spec change for `override`? It seemed easier than I expected, so let me know if I missed something...

spec/Classes.tex
```diff
@@ -309,11 +309,14 @@ \subsection{Overriding Base Class Methods}
 dispatch in the event that a variable that has the base class type
```
bradcray: You've inherited all of this, but since I'm reading it:

"the base class type" -> "a base class type"?  (or "...parent..." or "...ancestor..."?).  I'm on the fence about this.  On one hand, the previous text has restricted our scope from all base classes to a particular one; but on the other hand, this statement applies if the variable has any of the base class types.

Should this refer to "static type" in some way for improved clarity?  ("a variable whose static type is a base class type...")

spec/Classes.tex
```diff
@@ -309,11 +309,14 @@ \subsection{Overriding Base Class Methods}
...
 references an object that has the derived class type.
 
-The identical signature requires that the names, types, and order of
+The identical signature requires that the names, intents, types, and order of
```
bradcray: You've inherited this, but I think this either needs to say:

```Having the identical signature...```
or:
```"The identical signature" ...```

(where I prefer the former) because otherwise it sounds like "The identical signature" is a sentient entity requiring something from you.

I'd probably also say "means that" rather than "requires that" since this is defining things rather than prescribing.  (Or, I'd rewrite the previous paragraph to take more a prescription-based approach (e.g., "If a subclass wants to override a parent class method, that method must have an identical signature....  Having an identical signature requires...").  In contrast, the current text is using a style more like "If the signature happens to be the same..."

File: spec/Classes.tex [hunk 3]
```diff
@@ -309,11 +309,14 @@ \subsection{Overriding Base Class Methods}
...
 the formal arguments be identical. The return type of the overriding
 method must be the same as the return type of the base class's method,
 or must be a subclass of the base class method's return type.
```
bradcray: * drop comma before "or" or add "it" after the "or" (I think I prefer the latter)
* this might also parse more cleanly if "either" was inserted after the first "must" (in which case I'd drop the comma and the second "must"

spec/Classes.tex
```diff
@@ -309,11 +309,14 @@ \subsection{Overriding Base Class Methods}
...
 
+Methods that override a base class method must be marked with the
+\chpl{override} keyword in the \sntx{linkage-specifier}.
```
bradcray: Syntactically speaking, it feels weird to me to call `override` a linkage specifier since it has nothing to do with linking (`inline` arguably doesn't really either in retrospect...).  We could just rename `linkage-specifier` to `routine-kind` / `function-kind` or some other term that's more general.  Or we could have a general `routine-kind` nonterminal that maps down to `linkage-specifier` + `override` (+ `inline`?).

Maybe it's just that the one sentence goes by so quickly, but I wonder whether we should have a sentence or `\rationale` section saying a bit more about why this is the case (e.g., "This feature is designed to help avoid cases where class authors accidentally override a method without knowing it; or fail to override a method that they intended to due to not meeting the identical signature condition.").  

Or, another approach would be to add a sentence saying something like "Methods marked with `override` for which there is no parent class method with an identical signature will cause an error."  to capture the other side of the coin ("Must have it when overriding; can't have it when not overriding").  Though even then a rationale sentence might be valuable.

bradcray:
Those are the comments I have at this time.  Sorry to weigh down what seemed like a simple change.

mppf:
@bradcray - thanks for your comments so far. I believe I've addressed them, so have another look if you would.

spec/Classes.tex
```diff
@@ -303,16 +303,31 @@ \subsection{Overriding Base Class Methods}
...
-method must be the same as the return type of the base class's method,
-or must be a subclass of the base class method's return type.
+If a method in a derived class is declared with a signature identical to
+that of a method in a base class, then it is said to override the base
+class's method. Such methods are considered for dynamic dispatch.  In
+particular, dynamic dispatch will be used when the method receiver has a
+static type of the base class but refers to an instance of a derived
+class type.
+
+In order to have identical signature, two methods must have the same the
```
bradcray: * signature -> signatures?

spec/Classes.tex
```diff
@@ -303,16 +303,31 @@ \subsection{Overriding Base Class Methods}
...
+names, intents, types, and order of formal arguments. The return type of
+the overriding method must either be the same as the return type of the base
+class's method or be a subclass of the base class method's return
+type.
+
+Methods that override a base class method must be marked with the
+\chpl{override} keyword in the \sntx{procedure-kind}. Additionally,
+methods marked with \chpl{override} but for which there is no parent class
+method with an identical signature will cause an error.
```
bradcray: "cause an error" -> "result in a compiler error" (to clarify that it won't just result in an incorrect program run, say).

bradcray:
This is looking really good.  I just had two minor comments and if you agree with them, don't feel the need to re-review.

PR #11015, opened 2018-09-06
File two feature requests for the lifetime checker

bradcray:
This adds two feature requests for tests that don't cause lifetime
checking errors where I think it would be helpful if it did.
See the .future files for details.

On Michael's suggestion, I made these noexec tests since the
execution output would be unpredictable and they shouldn't
execute when things are working anyway.  This also permitted
sensible .bads to be added (currently containing the warnings
for using undecorated news).

### DIFF
--- a/test/classes/delete-free/lifetimes/field-init-escapes.skipif
+++ b/test/classes/delete-free/lifetimes/field-init-escapes.skipif
@@ -1,0 +1,1 @@
+CHPL_TEST_VGRND_EXE != on
--- a/test/classes/delete-free/lifetimes/field-init-escapes.skipif
+++ b/test/classes/delete-free/lifetimes/field-init-escapes.skipif
@@ -1,1 +1,0 @@
-CHPL_TEST_VGRND_EXE != on
--- a/test/classes/delete-free/lifetimes/forall-of-new-borrowed-escapes.bad
+++ b/test/classes/delete-free/lifetimes/forall-of-new-borrowed-escapes.bad
@@ -1,0 +1,11 @@
+forall-of-new-borrowed-escapes.chpl:13: warning: result of new C is now managed by default
+forall-of-new-borrowed-escapes.chpl:13: note: 'new unmanaged C' gives old behavior
+forall-of-new-borrowed-escapes.chpl:13: note: 'new borrowed C' is the new default
+forall-of-new-borrowed-escapes.chpl:13: note: 'new owned C', 'new shared C' also available
+forall-of-new-borrowed-escapes.chpl:13: note: get more help with --warn-unstable
+forall-of-new-borrowed-escapes.chpl:13: In iterator 'chpl__loopexpr_iter2':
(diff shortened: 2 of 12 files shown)


bradcray:
@mppf: Would you have time to give these a quick review when you need a study break?  I'll file the issues afterwards.

test/classes/delete-free/lifetimes/field-init-escapes.skipif
```diff
@@ -0,0 +1 @@
+CHPL_TEST_VGRND_EXE != on
```
mppf: I usually just make such tests .noexec and make the .bad file be empty (as in, it shouldn't have compiled).

bradcray: Oh you're right, that sounds a lot better!  I don't think I've used that pattern (definitely not recently anyway).

bradcray:
@mppf: Want to double-check I didn't screw anything up in doing that juggle?

test/classes/delete-free/lifetimes/forall-of-new-borrowed-escapes.bad
```diff
@@ -0,0 +1,11 @@
...
+forall-of-new-borrowed-escapes.chpl:13: note: 'new unmanaged C' gives old behavior
+forall-of-new-borrowed-escapes.chpl:13: note: 'new borrowed C' is the new default
+forall-of-new-borrowed-escapes.chpl:13: note: 'new owned C', 'new shared C' also available
+forall-of-new-borrowed-escapes.chpl:13: note: get more help with --warn-unstable
+forall-of-new-borrowed-escapes.chpl:13: In iterator 'chpl__loopexpr_iter2':
+forall-of-new-borrowed-escapes.chpl:13: warning: result of new C is now managed by default
+forall-of-new-borrowed-escapes.chpl:13: note: 'new unmanaged C' gives old behavior
+forall-of-new-borrowed-escapes.chpl:13: note: 'new borrowed C' is the new default
+forall-of-new-borrowed-escapes.chpl:13: note: 'new owned C', 'new shared C' also available
+forall-of-new-borrowed-escapes.chpl:13: note: get more help with --warn-unstable
```
mppf: I'd probably write `new borrowed` to avoid these warnings. Some things like `chpl__loopexpr_iter2` might not be stable over time. OK this way though if you prefer, for some reason. I wouldn't add `--no-warnings` to these.

bradcray: Oh, I hadn't read closely enough to notice the mention of the internal name...  That's too bad.  I liked writing them without the `borrowed` best because it seems like a case where people would be most surprised by this ("I wasn't thinking about ownership, so didn't think about these being borrowed") where the warning helps such people somewhat but if it went away before these checks were improved, I could imagine it causing confusion (since it did even for me who was thinking of them as explicitly being borrowed).  I'm inclined to leave them alone for now and if it becomes a maintenance burden before things improve, I'll migrate them to something better.

PR #11017, opened 2018-09-06
Fix performance regression in pop_front/pop_back from PR #10793

daviditen:
PR #10793 fixed a memory leak by tracking both the allocated memory and
the logical array size within dsiReallocate.  It did this by creating an
extra domain from an extra tuple of ranges for the logical size.

This extra domain added some performance overhead when reallocating arrays
for pop_front and pop_back.  Remove the overload of dsiReallocate that was
added in #10793.  Add a new overload (for rank==1 only) that has range
arguments instead of tuples of ranges, and avoids creating the extra domain.

This should fix the performance regression while continuing to avoid the
memory leak.

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -1328,13 +1328,12 @@ module DefaultRectangular {
     }
 
 
-    // Reallocate the array to have space for elements specified by
-    // `allocBounds` while maintaining the elements specified by `arrayBounds`
-    override proc dsiReallocate(allocBounds:rank*range(idxType,
-                                                       BoundedRangeType.bounded,
-                                                       stridable)) {
+    // Reallocate the array to have space for elements specified by `bounds`
+    override proc dsiReallocate(bounds: rank*range(idxType,
+                                                   BoundedRangeType.bounded,
+                                                   stridable)) {
       on this {
-        const allocD = {(...allocBounds)};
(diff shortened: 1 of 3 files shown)


daviditen:
Passed Linux64 paratest.
Verified that the leak fixed by #10793 is still not leaking.

modules/internal/DefaultRectangular.chpl
```diff
@@ -1285,29 +1285,63 @@ module DefaultRectangular {
       }
     }
 
-    override proc dsiReallocate(bounds:rank*range(idxType,BoundedRangeType.bounded,stridable)) {
-      dsiReallocate(bounds, bounds);
+    override proc dsiReallocate(allocBound: range(idxType,
```
ben-albrecht: Why not use the same naming scheme across overloads, e.g. `allocBound` -> `bounds` or vice-versa?

daviditen: I wanted to differentiate the two arguments here - one is about the bound of the allocation, the other is about the bound of the logical array.  I could change `bounds` to `allocBound` if you think it would be worthwhile.  That would require updates in ChapelDistribution.chpl, DefaultRectangular.chpl, BlockDist.chpl, CyclicDist.chpl, StencilDist.chpl, and possibly some tests since they all need to use the same names to be valid overloads.  Think it's worth it?

PR #11033, opened 2018-09-07
Convert a few other shootouts to use managed classes

bradcray:
This converts other shootout performance tests over to use managed
classes.  In the case of Albert Sidelnik's nbody, we can get away with
borrowed classes only resulting in a code that should work with older
versions of the compiler (albeit by leaking).  In the other cases, I
switched from unmanaged to managed pointers to use the language more
as we think we'd like.

TODO:
- [x] memleaks verification
- [x] performance correctness testing

### DIFF
--- a/test/studies/shootout/fannkuch-redux/bharshbarg/fannkuch-redux-coforall.chpl
+++ b/test/studies/shootout/fannkuch-redux/bharshbarg/fannkuch-redux-coforall.chpl
@@ -22,8 +22,8 @@ var maxFlips, checkSums : [0..#NTASKS] int;
 
 var taskCount : atomic int;
 
-var Fanns : [ntasks] unmanaged Fann;
-for f in Fanns do f = new unmanaged Fann();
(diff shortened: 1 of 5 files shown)


bradcray:
Saving this until after the release because none of the codes are contained within the release and only one would enable new release-over-release numbers (the nbody code, but it's not one of our preferred variations).  The other codes all rely on new features.

bradcray:
@benharsh, @lydia-duncan: I think these changes are minor / no-brainers, but am tagging in case you want to review (pre- or post-merge... I'll probably merge tonight since I'm on a shootout tear).

lydia-duncan:
No objections here!

bradcray:
As expected/hoped, no major changes here:

https://chapel-lang.org/perf/chapcs/?startdate=2018/08/31&enddate=2018/09/19&graphs=nbodyvariations,spectralnormstepsizetimes,fannkuchreduxn12,spectralnormshootoutbenchmark

PR #11036, opened 2018-09-07
Update spec for delete-free

mppf:
* replaces type-specifier with type-expression because it can include calls
* updates type-expression to possibly include calls
* replaced constructor-call-expression with new-expression
* migrated several examples to delete-free and initializers
* include 1-length tuples in tuple type expressions
* document different delete-free things owned / unmanged /borrowed / shared
* mentions lifetime checking but leaves details as future work
* removes some mention of multiple inheritance
* describes owned intent and the special no-type instantiation rule.

- [x] spec tests pass with -valgrindexe


vasslitvinov:
Be sure to run -valgrindexe testing on the Class directory of spec tests.

spec/Classes.tex
```diff
@@ -799,48 +923,44 @@ \section{Deleting Class Objects}
...
 \index{deinitializer!classes}
 
-A class author may specify additional actions to be performed before a class object is
-reclaimed, by defining a class deinitializer.  A class deinitializer is a method
-named \chpl{deinit()}.  A class deinitializer takes no arguments (aside from
-the implicit \chpl{this} argument).  If defined, the deinitializer is called each time
-a \chpl{delete} statement is invoked with a valid instance of that class type.  The
-deinitializer is not called if the argument of \chpl{delete} evaluates to \chpl{nil}.
+A class author may create a deinitializer to specify additional actions
+to be performed before memory for instances of that type are reclaimed.
```
vasslitvinov: "before memory for instances of that type are reclaimed" --> "when a class instance is deleted"

I think this way it is more specific.

spec/Generics.tex
```diff
@@ -820,14 +814,30 @@ \subsection{User-Defined Constructors}
...
+    this.x5 = x5;
+  }
 }  // class MyGenericClass
 
-var g1 = new MyGenericClass(11, 111, int, 1);
-var g2 = new MyGenericClass(int, "this is g2", 3.3, 333, 3333,
-                            real, 2, 222, 222.2, 22);
+var g1 = new borrowed MyGenericClass(11, 111, int, 1);
+var g2 = new borrowed MyGenericClass(int, "this is g2", 3.3, 333, 3333,
+                                     real, 2, 222, 222.2, 22);
```
vasslitvinov: Again, if we do not need to add "borrowed", I'd rather not.

spec/Task_Parallelism_and_Synchronization.tex
```diff
@@ -209,10 +209,11 @@ \section{Synchronization Variables}
 }
 \end{chapel}
 \begin{chapelpost}
-var tree: Tree = new Tree(false, new Tree(false, new Tree(true, nil, nil, 1),
-                                                 new Tree(true, nil, nil, 1), 1),
-                                 new Tree(false, new Tree(true, nil, nil, 1),
-                                                 new Tree(true, nil, nil, 1), 1), 1);
+var tree: unmanaged Tree = new unmanaged Tree(false, new unmanaged
+Tree(false, new unmanaged Tree(true, nil, nil, 1),
```
vasslitvinov: A stray newline in the above line?

spec/Classes.tex
```diff
@@ -364,13 +448,69 @@ \subsection{Default Initialization}
...
+  \item \chpl{new shared C()} allocates and initializes the instance that
+    will be deleted when the last \chpl{shared} variable referring to it
+    goes out of scope. Results in something of type \chpl{shared C}.
+
+\end{itemize}
+
+\section{Class Constructors}
+\label{Class_Constructors}
+\index{classes!constructors}
 \index{constructors}
```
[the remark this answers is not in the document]
mppf: I don't want to change that now since Ben H is about to.

mppf:
@vasslitvinov - could you have a look at the recent improvements to this PR? I added some new descriptions of intents. I don't know of anything else significant that is missing from the changes for this PR.

spec/Classes.tex
```diff
@@ -140,15 +166,15 @@ \subsection{Class Lifetime and Borrows}
...
 programming.
 
-When a borrow exists beyond the end of the lifetime of the instance it
-refers to, particularly difficult to debug errors are possible. For this
-reason, the compiler includes a component called the lifetime checker that
-helps to identify cases in which a borrow could exist beyond the lifetime
-of an instance it refers to.
+Errors due to accessing an instance after the end of its lifetime are
+particularly difficult to debug. For this reason, the compiler includes a
+component called the lifetime checker. It identifies some where a
```
vasslitvinov: Please add "cases" between "some" and "where" above.

spec/Classes.tex
```diff
@@ -186,17 +211,39 @@ \subsection{Class Types}
...
     variables referring to the instance go out of scope.
+    See
+    \\ %formatting
+    \mbox{$$ $$ $$ $$ $$} %indent
+    \url{https://chapel-lang.org/docs/builtins/internal/SharedObject.html}.
+
+  \item \chpl{borrowed} refers to a class instance that has a lifetime
+    managed by another variable.
+
+  \item \chpl{unmanaged} the instance must have \chpl{delete} called on it explicitly.
```
vasslitvinov: Let us add "to [free|reclaim|whatever] its memory" (as you have done in other places).

Because, technically, it is still correct not to call `delete`.

spec/Classes.tex
```diff
@@ -314,7 +361,7 @@ \section{Inheritance}
 When inheriting from multiple base classes, only one of the base classes
 may contain fields.  The other classes can only define methods.  Note
 that a class can still be derived from a class that contains fields
-which is itself derived from a class that contains fields.  
+which is itself derived from a class that contains fields.
```
vasslitvinov: Ugh. Would you care to move multiple inheritance to a "future work" section?
This is optional, as it is unrelated to this PR. I could give you a hand with this if you think this should be done.

spec/Classes.tex
```diff
@@ -338,7 +385,7 @@ \subsection{The object Class}
 indirectly.  If no class name appears in the inheritance list, the class derives
 implicitly from \chpl{object}.  Otherwise, a class derives from \chpl{object}
 indirectly through the class or classes it inherits.  A variable of type \chpl{object}
-can hold a reference to an object of any class type. 
+can hold a reference to an object of any class type.
```
vasslitvinov: If we are removing multiple inheritance, we should remove "or classes" in the above.

spec/Classes.tex
```diff
@@ -838,6 +893,28 @@ \subsection{Class Assignment}
 variable of a class type to another, both variables reference the same
 class instance.
 
+Assignment for \chpl{owned} variable transfers ownership and leaves the
```
vasslitvinov: Let's add "an" before "owned" above. That way it will parse smoothly.

spec/Classes.tex
```diff
@@ -838,6 +893,28 @@ \subsection{Class Assignment}
...
+source variable empty. For example:
+
+\begin{chapelexample}{owned-assignment.chpl}
+Supposing that \chpl{C} is a class type:
+\begin{chapelpre}
+class C { }
+\end{chapelpre}
+\begin{chapel}
+var a = new owned C();
```
vasslitvinov: Instead of "Supposing that \chpl{C} is a class type" above, I suggest to add a comment in the `var a .....` line above saying, ex., `// assume C is a class` (or ...... class type). This should make it more organic / connected to the code.

spec/Procedures.tex
```diff
@@ -585,11 +607,44 @@ \subsubsection{The Default Intent}
...
 intent overloads (see \rsec{Return_Intent_Overloads}).
 
+\subsubsection{Default Intent for \chpl{owned} and \chpl{shared}}
+\label{Default_Intent_for_owned_and_shared}
+\index{intents!default for owned}
+\index{intents!default for shared}
+
+The default intent for \chpl{owned} and \chpl{shared} arguments is
+generally \chpl{in}.
+
```
vasslitvinov: Should we indicate that the paragraph below is "an exception" to the paragraph above?

spec/Task_Parallelism_and_Synchronization.tex
```diff
@@ -209,8 +209,7 @@ \section{Synchronization Variables}
 }
 \end{chapel}
 \begin{chapelpost}
-var tree: unmanaged Tree = new unmanaged Tree(false, new unmanaged
-Tree(false, new unmanaged Tree(true, nil, nil, 1),
```
vasslitvinov: Could you please double-check that the next line fits on a page in the resulting PDF? If not, would you split it in a good spot?

mppf: it doesn't show up in the PDF at all. It's part of the post-example-code for the test.

spec/Classes.tex
mppf: I updated the particular 2 multiple inheritance spots. I'd appreciate help updating any other parts this PR doesn't fix, post-merge.

spec/Classes.tex
```diff
@@ -87,35 +84,178 @@ \section{Class Declarations}
...
+\begin{chapelexample}{borrowing.chpl}
+\begin{chapel}
+class C { }
+proc test() {
+  var own = new owned C(); // 'own' manages the memory of the instance
+  var b = own.borrow();    // 'b' refers to the same instance but has no
+                           // impact on the lifetime.
+
+  var bc: borrowed C = b;  // 'bc' stores the result of b.borrow()
+                           // due to coercion from owned C to C
```
cassella: OOC, what's the `owned C` being coerced to `C` here?

PR #11038, opened 2018-09-07
Wordsmith description of class locality in locales primer

bradcray:
While trying, and failing, to remove 'unmanaged' from this primer, I
found myself stumbling over the description of classes and locality,
so took a stab at rewriting it with input from Michael.


bradcray:
@mppf: I continued to wrestle with that paragraph after the last draft I sent you and ended up here.  What do you think?

PR #11040, opened 2018-09-07
Update the library technote for the new state of the feature

### LINKED ISSUE #10316: Update interoperability documentation
This release is seeing a lot of updates to our interoperability stories.  These updates should be documented, but it may be good to wait on that until closer to the release, since the instructions are going to continue to evolve.

There are aspects of the Cython path of interoperability that can be a little tricky.  Some of that will go away, but others should be explicitly stated to avoid user issues.  I'm going to track what needs to be done here so that I don't forget it when I do get to writing the documentation (instead of continually updating the instructions)

lydia-duncan:
Adds new content, cleans up examples and descriptions, and adds some
organizational changes (especially topic headers)

Resolves #10316 

Read through a build of the document and followed all links


doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+============================
+
+Pre-requisites
+--------------
+
+To make use of your library in Python with minimal work, the Chapel compiler
+requires that you have ``python3`` installed in your ``$PATH``, and that your
+``python3`` installation contains ``numpy`` and ``Cython``.
+
+We don't anticipate supporting ``python2`` at this time, sorry.
```
ben-albrecht: (Un)planned features would be better documented in issues rather than here. 

Side note: It bothers me that we only support python2 some places and only python3 here. I think the solution is to support python3 everywhere, but we're in an awkward position right now.

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
 Note that ``compileline --compile-c++`` is also available. It requests a
 compilation command able to compile a C++ program.
 
+Using Your Library in Python
+============================
+
+Pre-requisites
+--------------
+
+To make use of your library in Python with minimal work, the Chapel compiler
```
ben-albrecht:
```
Calling Chapel libraries from python requires the following:

  - ``python3`` installed in your ``$PATH``
  - ``Cython``
  - ``numpy``

```

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+requires that you have ``python3`` installed in your ``$PATH``, and that your
+``python3`` installation contains ``numpy`` and ``Cython``.
+
+We don't anticipate supporting ``python2`` at this time, sorry.
+
+Compiling Your Chapel Library
+-----------------------------
+
+In addition to compiling with ``--library``, it is recommended to compile with
+``--library-python``.  This will create a Python-compatible module in addition
```
ben-albrecht: recommended or necessary?

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+to the normally generated library and header file.
+
+Python Module Name
+------------------
+
+By default, the name of the generated Python module will match the base name
```
ben-albrecht: We use `basename` (one word) elsewhere, so maybe stay consistent with `basename`

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+of the generated library.  In the ``foo.chpl`` example, this would mean the
+Python module will be named ``foo``.
```
ben-albrecht: By default, the name of the generated Python module will match the base name of the generated library, but can be changed independently of the generated library name using the compilation flag ``--library-python-name``:


```
.. code-block::

    # Builds python module as lib/foo.py
    chpl --library --library-python foo.chpl

    # Build python module as lib/bar.py
    chpl --library --library-python --library-python-name=bar foo.chpl
```

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+
+   foo.chpl_setup()
+
+   foo.baz(7) // Call into a library function
+
+   foo.chpl_cleanup()
+
+.. note::
+
+   Currently the ``chpl_cleanup()`` function will also cause the Python program
```
ben-albrecht: The ``chpl_cleanup()`` function will also cause the Python program  to exit. Make sure your
Python functionality is also complete before calling this function.

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+----------------------------
+
+This compilation strategy uses Cython under the covers, generating a
+``chpl_foo.pxd`` file, a ``foo.pyx`` file, and a ``foo.py`` file by default for
+a ``libfoo.a`` / ``libfoo.so``, which are then called using a Cython command
+(this command is rather long due to the need to include the Chapel runtime and
+third-party libraries).  These files are currently left in the same location as
+the generated library - if compilation fails due to generating one or more of
+these files incorrectly, you may be able to modify the file and re-run the
+Cython command yourself.
```
ben-albrecht: Would be nice to have an example of this, if feasible.

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+
+Please report such errors to us using the strategies defined in
```
ben-albrecht: I think we covered this at the top.

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+
+   foo.chpl_cleanup()
+
+.. note::
+
+   Currently the ``chpl_cleanup()`` function will also cause the Python program
+   to exit.  This may change in the future, but for now please ensure your
+   Python functionality is also complete before calling this function.
+
+In Case Something Goes Wrong
```
ben-albrecht: How about we call this something like "Debugging issues" or something

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+Using Your Library in Chapel
+============================
+
 Chapel library files cannot be used from Chapel code.  The library files must
 include the chapel runtime and standard modules for use in a non-Chapel program
 and when the library is linked to a Chapel program this leads to multiple
 definitions of these functions.
 
-As mentioned above, this feature is not very sturdy.  Please refer to
+Other Caveats
```
ben-albrecht: Let's just call this "Caveats"

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+for its own functionality and when two or more libraries are linked to a program
+this leads to multiple definitions of these functions.
+
+
+.. _Exporting Symbols:
+
+Exporting Symbols
+-----------------
+
+At this time, only functions can get exported.  We hope to extend this support
```
ben-albrecht: Only functions can be exported currently.

doc/rst/technotes/libraries.rst
```diff
@@ -97,12 +135,68 @@ Chapel library function calls are finished.  For example:
...
+``chpl__init_<moduleName>``, and you can find its declaration in your generated
+``.h`` file.
+
+.. note::
+
+   It is recommended that you always call the module initialization function
+   before calling any of the exported functions in your library.  You do not
+   need to do this more than once per program.
+
+   We intend to eventually perform the module initialization when
```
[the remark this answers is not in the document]
lydia-duncan: It is already an issue, so I will just remove that explanation, thanks!

doc/rst/technotes/libraries.rst
```diff
@@ -111,23 +205,145 @@ example looks like this:
...
+
+.. _Exporting Symbols:
+
+Exporting Symbols
+-----------------
+
+At this time, only functions can get exported.  We hope to extend this support
+to types and global variables in the future.
+
+Issues?
```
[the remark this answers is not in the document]
lydia-duncan: This is technically its own header instead of a subheader of Caveats.  I figured it would be good to have this section as close to the top-level as possible, so it is easier to find.  I probably should make the other place link to it instead of duplicating information, though.  Thanks!

doc/rst/technotes/libraries.rst
lydia-duncan: Yeah, it would be nice to go on a clean up spree and update everything to use python3.  Maybe this is a conversation for the group meeting about what we need to still support?

doc/rst/technotes/libraries.rst
lydia-duncan: They could write all the Cython steps on their own (or do the work for another strategy) using just the generated library or generated library/header combo.  But that's a lot of work.

doc/rst/technotes/libraries.rst
lydia-duncan: I thought about it, but the example is going to be huge and difficult to read.  Plus, any error they get from Cython will tell them the Cython command used.  Should I point that out?

doc/rst/technotes/libraries.rst
lydia-duncan: Probably not much has changed, though :/

doc/rst/technotes/libraries.rst
ben-albrecht: > but the example is going to be huge and difficult to read

Ah, OK. That's why I said "if feasible" :)

> Plus, any error they get from Cython will tell them the Cython command used. Should I point that out?

Not sure how useful that would be. I'll leave it up to your judgement.

doc/rst/technotes/libraries.rst
lydia-duncan: ``--library`` isn't required when other ``--library-*`` commands are present.  Should I call that out explicitly?

Issue #10947, opened 2018-09-04
Minimize use of unmanaged in release/examples


bradcray:
This epic relates to getting the release/examples directory to the point that it's using delete-free features as we'd like it from a "best practices" perspective.  I believe that in most of the following cases, `unmanaged` is currently being used simply because it was the most expedient way to get the feature in.  But for these codes that are user-facing, it'd be nice to get them to a state where they also use the features to avoid requiring the use of unmanaged.

In some sort of priority-sorted order:

- [ ] shootout (binarytrees and chameneosredux*)
- [ ] primers (distributions and locales)
- [ ] hpcc (hpl and ptrans)
- [ ] users-guide (walkTreeUsingBegins)
- [ ] programs (genericStack, linkedList, and tree)
- [ ] lcals
- [ ] miniMD
- [ ] ssca2

bradcray:
I've got a start on the shootout and hpcc cases in PR #10945 and will also take on the users guide.

@daviditen : Would you have time to look at LCALS in this sprint?

daviditen:
Yes, I can take LCALS.

ronawho:
Close https://github.com/chapel-lang/chapel/issues/10598 in favor of this?

mppf:
Code that actually wants to manage its own memory should use `unmanaged` and if that's what the code is doing we should be OK with that. There might be 1 or 2 of the above cases that fall into this category.

bradcray:
Michael's point is a good one.  I don't mean to imply that no codes should ever use `unmanaged` again, more that we should use what's most appropriate for the code and make sure we're not just using `unmanaged` because it was the quickest way to keep the program working for this release.

mppf:
I poked around a bit and don't see anything else that isn't covered above.

bradcray:
For the 1.18 release, I didn't come up with rewrites of the `programs/` files that I was sufficiently happy with to merge (and found myself mostly wondering if that directory should just be removed.  It's ancient and doesn't necessarily represent code that we care deeply about.

I essentially timed out on SSCA2, but figured it was a sufficiently large / long-lived code that that didn't feel concerning to me (though I'd like to return to it post-release).

mppf:
@bradcray - see PR #11086 moving programs

Issue #10598, opened 2018-07-31
Investigate unmanaged in release/examples


mppf:
Generally speaking we shouldn't be using `unmanaged` or `new borrowed` in release/examples. This issue is documenting the task of `grep`ing for these and trying to resolve them where possible.

Also check for `_to_unmanaged` or `_to_borrowed` since these can generally be replaced with the generic cast.

bradcray:
Oops, I had not seen / thought to look for this prior to opening my issue this morning.   Since mine was an epic and this mentions a few things to check for that I didn't, why don't I put this under the epic.

bradcray:
@mppf: Is your thinking about avoiding `new borrowed C()` that we should be simply writing these as `new C()` or that we should try to write them in some way that doesn't use this pattern at all?

FWIW, I didn't find any hits for `_to_unmanaged` or `_to_borrowed` within release/examples.

mppf:
> FWIW, I didn't find any hits for _to_unmanaged or _to_borrowed within release/examples.

great, less work for the moment.

> @mppf: Is your thinking about avoiding new borrowed C() that we should be simply writing these as new C() or that we should try to write them in some way that doesn't use this pattern at all?

I was imagining that we'd prefer `new owned C()` for these examples.

bradcray:
Looks like the cases that use `new borrowed` at present are:

* benchmarks/shootout/chameneosredux.chpl (which I'm already working on rewriting in PR #10945)
* primers/classes.chpl (which I think is appropriate since it's teaching the concept)
* primers/learnChapelInYMinutes.chpl (which was changed from `new C` to `new borrowed C` in df7be02cdb27d1bf0956e391b2a1634d5353719e)

I've updated issue #10947 to include this last case.

mppf:
I think @bradcray did most of the work on this one - thanks! The epic #10947 can remain open but I think we've done what we hoped to for this sprint, so I'm closing this issue.

PR #11042, opened 2018-09-07
Change uses of 'borrowed' in this primer to 'owned' or ''

bradcray:
As part of issues #10947 and #10598, I looked for uses of `borrowed` in this code and changed them to `owned` for variable declarations and `` for formal arguments.


bradcray:
@mppf: Would you have time to give this one a quick review?  (since your issue was the one that suggested removing `new borrowed` instances)

PR #11043, opened 2018-09-07
Fix valgrind failure in literal-indexing.chpl

### LINKED ISSUE #8156: missing runtime type from empty for expression
The conversion from an iterator (such as with a for expression) to an array does not properly initialize the runtime type for arrays of arrays in the event that the iterator yields 0 times.

For example:

``` chapel
config const n = 0;
var r = 1..n;

proc makeArray(i:int) {
  var A:[1..i] int;
  A[1] = i;
  return A;
}

// AA is an array of arrays
var AA = for i in r do makeArray(i);
writeln(AA);

writeln("AA domain ", AA.domain);
var element:AA.eltType;
writeln("AA (element) eltType domain ", element.domain);
```

Results in `error: attempt to dereference nil` when run with `--n=0` but runs correctly with `--n=1`.

test/arrays/ferguson/from-iterator/for-yields-no-arrays.chpl (in PR #8073)

vasslitvinov:
In the recently-added proc chpl_buildStandInRTT(type arrType: [])
the domain `standinDomain` was getting auto-destroyed before
that function returned. This resulted in access-after-delete in:

    test/arrays/vass/literal-indexing.chpl

To fix that, added pragma "no auto destroy".

As a result, that domain started to leak in:

    test/arrays/ferguson/from-iterator/skyline.chpl
    test/arrays/ferguson/from-iterator/inner-array-resize.chpl

Aside: these are probably the only two non-future tests that build
arrays of arrays using the chpl__initCopy() overload implementing
the shape-less conversion from an _iteratorRecord to an array.
The reason these tests passed valgrind before is that nothing
accessed the prematurely-destroyed `standinDomain`. The runtime
counterpart of the iteratorToArrayElementType()-computed type
was dropped on the floor. Somehow it wasn't stored in the
generic instantiation of the _ddata type of the `data` local
variable. (Maybe that is normal for instantiations of class types.)

To plug this leak, I auto-destroy it manually in that chpl__initCopy.
Since currently we do not have a way to access runtime types from
Chapel code, I added a primitive to do that.

It also required storing the iteratorToArrayElementType()-computed type
in a `type` alias so it can be accessed later. As a side benefit,
I un-futurized and beefed up the case for the case where the iterator
for the _iteratorRecord formal happens to yield no elements at run time:

    test/arrays/ferguson/from-iterator/for-yields-no-arrays.chpl 

This test leaks the `standinDomain` (twice, as it is called twice).
This reflects the bigger issue of not properly memory-managing
the runtime types, which is neither introduced nor addressed
by this PR.

Resolves #8156.

Testing:
- [x] linux64 + futures
- [x] valgrind
- [x] memleaks


compiler/resolution/functionResolution.cpp
```diff
@@ -8263,6 +8263,31 @@ buildRuntimeTypeInfo(FnSymbol* fn) {
...
+           VarSymbol* domTemp = newTemp("domTemp", domField->getValType());
+           call->insertBefore(new DefExpr(domTemp));
+           call->insertBefore("'move'(%S,'.v'(%E,%S))", domTemp,
+                              rttSE->remove(), domField);
+           call->insertBefore(new CallExpr(destroyFn, domTemp));
+          }
+     // Otherwise this call is a no-op. Remove it.
+     call->remove();
+    }
+}
```
mppf: It seems to me that callDestructors would be a better place for this code. That's assuming it can't be in preFold or postFold, (presumably because the runtime types don't exist yet?).

modules/internal/ChapelArray.chpl
```diff
@@ -4292,7 +4297,7 @@ module ChapelArray {
       // returning arrays of arrays but that returned no elements.
       // We need to be able to construct the runtime portion of the type
       // in that event.
-      var A = D.buildArrayWith(data[0].type, data, size:int);
+      var A = D.buildArrayWith(elemType, data, size:int);
```
mppf: comment above this needs updating

compiler/resolution/functionResolution.cpp
mppf: Also this code can't handle runtime types for domains (which store distributions). At least it needs a comment that it would need updating.

PR #11045, opened 2018-09-07
Switch miniMD over to use managed classes

bradcray:
This is not a major change or win, but nice to use the managed
class features.


bradcray:
@benharsh: This should be an easy study break if you have time.

Issue #10215, opened 2018-07-09
should untyped generic instantiated with owned depend on intent?


mppf:
Based on the design in #8938, the type instantiated for an unconstrained generic formal, when passed an owned actual, is the borrow type:

``` chapel
proc f(x) {
  // x.type is borrowed MyClass
}
f(new owned MyClass());
```

This issue asks if a corresponding behavior should apply to other intents:
 * `in` and `const in`
 * `ref` and `const ref`

See also #10142.

mppf:
See also afc5fdd34e5824bf6a5de94a5522ef81d728f212 which removed the sensitivity to intent and passed testing

bradcray:
I'm trying to make sure I'm understanding what you're proposing/asking here.  Is the idea that if I were to write:

```chapel
proc f([const] in x) {
  writeln(x.type:string);
}
var myC = new owned MyClass();
f(myC);
```

we need to decide whether `x` should be a copy-in of a borrow of `myC` vs. a copy-in of `myC` itself (presumably resulting in a new owned class whose scope would be `f()`)?  Thanks.

mppf:
> we need to decide whether `x` should be a copy-in of a borrow of `myC` vs. a copy-in of `myC` itself (presumably resulting in a new owned class whose scope would be `f()`)?  Thanks.

Yes, that's the question. I also have some hope for consistency with task intents but I'm not really sure what to expect there.

There's also the analogous question about `ref` intent.

``` chapel
proc f([const] ref x) {
  writeln(x.type:string);
}
var myC = new owned MyClass();
f(myC);
```

Should it output `borrowed MyClass` or `owned MyClass` ?

mppf:
> somewhat like variable initialization and that we'd borrow in the `var x = myC;` case (right?).

``` chapel
var x = new owned MyClass();
var y = x;
```

y will have type `owned MyClass`. This would create a borrow:

``` chapel
var x = new owned MyClass();
var y:MyClass = x;
```

mppf:
> Do you have arguments for why `borrowed` might make sense in the `ref` cases?

In fact it would probably be a compilation error (can't pass result of coercion by ref). I only wish for the rule about instantiation to be as simple as we can get it.

bradcray:
> In fact it would probably be a compilation error

OK, sounds like having `ref` simply mean "refer to the original [owned] thing" is what we're converging on for this case then?

> `y` will have type `owned MyClass`

[...]

PR #11046, opened 2018-09-07
Fix untyped args passed owned with in intent

mppf:
Change untyped `in` intent to transfer ownership for owned

For example
``` chapel

	class C {}
	var x = new owned C();
	foo(x);

	proc foo(in y) {
	  writeln(y.type:string);
	}
```

now prints `_owned(C)` rather than `C`.

This relates to issue #10215 which is still open but making this change makes the release more reasonable & follows the best choice in that issue.

- [x] full local futures testing (as part of PR #11047)

### DIFF
--- a/test/classes/delete-free/owned/owned-instantiation-types.good
+++ b/test/classes/delete-free/owned/owned-instantiation-types.good
@@ -4,10 +4,10 @@ blank_class MyClass
 const_untyped MyClass
 const_owned _owned(MyClass)
 const_class MyClass
-in_untyped MyClass
+in_untyped _owned(MyClass)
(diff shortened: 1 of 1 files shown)


bradcray:
Assuming clean testing, this looks good, thanks!  For those following along at home, my way of thinking of this being the right behavior is:

* `var x = new owned C()` causes `x` to be `owned C` because that seems most natural
* `var y = x` causes `y` to be `owned C` because the type signatures are identical to the previous
* `in` intent should behave similarly to variable initialization suggesting that `proc foo(in y)` should behave the same as the previous

PR #11048, opened 2018-09-07
Add support for using urxvt with CHPL_COMM_USE_(G|LL)DB

e-kayrakli:
While using CHPL_COMM_USE_(G|LL)DB, a new `xterm` window that runs the relevant debugger is started for each locale. For users who use different terminal emulators and who don't have `xterm` configured properly, it gets unpleasant to work mostly because of cosmetic reasons.

This PR adds ability to assign `urxvt` to `CHPL_COMM_USE_GDB`. Anything other than `urxvt` silently defaults to `xterm` to preserve the current behavior. When both `CHPL_COMM_USE_LLDB` and `CHPL_COMM_USE_GDB` are assigned, `GDB` one takes precedence.

TODO:
- [ ] Test if `CHPL_COMM_USE_GDB=urxvt` performs gracefully if `urxvt` doesn't exist.


e-kayrakli:
Not too much work-in-progress, but I need to run some trivial tests. Needs design review for sure, and code review especially on the string manipulation parts. I always get uneasy dealing with strings in C :(

And can be ignored until someone has nothing important to do :)

ronawho:
> I always get uneasy dealing with strings in C :(

Yeah, me too. I'm not sure if we have an issue for it, but we have a long-standing desire/TODO to script-ify the launchers. For the most part the launchers just run external program or do string manipulation, neither of which I find enjoyable in C.

gbtitus:
I see some minor things that could be coded better, but before working on those there's a design issue I'd like to discuss.  `CHPL_COMM_USE_(G|LL)DB` used to just be checked for existence, that is, if they were set (to anything, it didn't matter) then we ran the corresponding debugger, one instance per locale, in display manager terminal windows.  This change makes its value meaningful.  I don't have a problem with that as such, but what it puts in there is the name of the display manager terminal window program, which seems odd to me.  The terminal program is really just an implementation detail, a place to run the shell that runs the debugger.  If the value of `CHPL_COMM_USE_(G|LL)DB` was meaningful, it would make sense to me as a user if it were the name or path of a debugger program, but here it's the name of something else instead.

All this is background to asking 2 things:
- What is the reason for this change, that is, how did the need arise?  For example, are you using a system that doesn't have `xterm`, so our existing debug support doesn't work?  ("Why" the change is needed is always useful in a PR.)
- Would it be okay if we left `CHPL_COMM_USE_(G|LL)DB` as they are and used a new env var to specify the name of the terminal program, for example `CHPL_COMM_DBG_TERM`?  (Or feel free to propose a different/better name; I'm not totally in love with that one.)

e-kayrakli:
> What is the reason for this change, that is, how did the need arise? For example, are you using a system that doesn't have xterm, so our existing debug support doesn't work? ("Why" the change is needed is always useful in a PR.)

You're right. I added a paragraph in the PR message. I think the message itself will change a little bit further if we are going further with this. For now I kept the rest same.

To answer your question more directly, it was difficult to work with `xterm` for me. Instead of trying to configure it alongside `rxvt-unicode` I just patched the launcher. Arguably, this PR has a very small target audience (me only?). So, if you prefer to keep the code unchanged and reject the PR, I totally understand :)

> Would it be okay if we left CHPL_COMM_USE_(G|LL)DB as they are and used a new env var to specify the name of the terminal program, for example CHPL_COMM_DBG_TERM? (Or feel free to propose a different/better name; I'm not totally in love with that one.)

Oh most definitely. I created the PR just to get the ball rolling. This is what I had in working branch, but what you are proposing is infinitely better.

OTOH, ideally, the spawned terminal emulator should be the same emulator it was rooted in. Only way I know that can let you do that is to check `$TERM` but I think it can be changed by different applications. I observe that kind of behavior on both tmux and GNU screen. So, if you know a way to check current terminal emulator reliably, I believe it is the most principled way to do this.

TLDR: if you are willing to pursue this further, I can revise the PR to make it aligned with your suggestions, if not feel free to close it.

gbtitus:
@e-kayrakli , let's proceed with this.  But can you change to `CHPL_COMM_DBG_TERM` for the env var?  This should actually make the patch simpler, too.

BTW, I don't think using `TERM` works.  That's not the name of the terminal program, it's the name of a (perhaps virtual) terminal type that the program is emulating.  It's really about behavior, not program.  I don't know of a portable way to find your terminal program.  And in some situations, like when you're ssh'ed from an iTerm2 window on a Mac to some Linux system to run a Chapel program there, the program corresponding to the terminal you're "in" doesn't even exist where you're executing.

e-kayrakli:
@gbtitus sounds good! I will ping you once I update the PR.

e-kayrakli:
@gbtitus - last commits should address your feedback on using a different environment variable.

With these changes I don't think it is necessary to test with lldb. But I can still test what happens if `CHPL_COMM_DBG_TERM=urxvt` but `urxvt` is not in `$PATH`

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -97,13 +97,34 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
...
-      const size_t xterm_path_size = PATH_MAX;
-      char *xterm_path = chpl_mem_alloc(xterm_path_size,
+      const size_t term_path_size = PATH_MAX;
+      char *term_path = chpl_mem_alloc(term_path_size,
                                         CHPL_RT_MD_COMMAND_BUFFER, -1, 0);
 
-      if (chpl_run_cmdstr("which xterm", xterm_path, xterm_path_size) > 0) {
-        largv[largc++] = xterm_path;
+      static char cmd[16] = "";
+      sprintf(cmd, "which %s", dbg_term);
```
gbtitus: Need `snprintf(cmd, sizeof(cmd), ...)` here!

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -113,7 +134,10 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
           largv[largc++] = (char *) "--";
         }
       } else {
-        chpl_warning("CHPL_COMM_USE_(G|LL)DB ignored because no xterm", 0, 0);
+        static char err_msg[128] = "";
+        sprintf(err_msg, "CHPL_COMM_USE_(G|LL)DB ignored because no %s",
```
gbtitus: Need `snprintf()` here also.

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -97,13 +97,34 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
...
+      }
+      else if(strcmp(dbg_term, "xterm") != 0  && 
+              strcmp(dbg_term, "urxvt") != 0) {
+        // currently limiting to xterm and urxvt because of lack of
+        // testing/interest. Most other terminal emulators seem to support the
+        // -e flag, but there are some (like tilda) that uses different flags
+        // (-c for tilda). Ideally, there should be a lookup table for those
+        // flags if there is interest in adding support for more terminal
+        // emulators
+        chpl_warning("CHPL_COMM_USE_(G|LL)DB can only be set to xterm or urxvt",
```
gbtitus: This should be `CHPL_COMM_DBG_TERM` instead of `CHPL_COMM_USE_(G|LL)DB`.

runtime/src/launch/amudprun/launch-amudprun.c
```diff
@@ -97,13 +97,34 @@ static char** chpl_launch_create_argv(const char *launch_cmd,
     const char* ev_use_lldb = getenv("CHPL_COMM_USE_LLDB");
 
     if (ev_use_gdb != NULL || ev_use_lldb != NULL) {
+
+      // determine the terminal emulator to use
+      char* dbg_term = getenv("CHPL_COMM_DBG_TERM");
```
gbtitus: When I built this branch with `WARNINGS=1` I got this from gcc, on the assignments at lines 104 and 116:
> launch-amudprun.c:104:18: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]

Changing this declaration to `const char* dbg_term ...` got rid of those errors.  Would you make this change, please?

runtime/src/launch/amudprun/launch-amudprun.c
e-kayrakli: Oops! Fixed it.

runtime/src/launch/amudprun/launch-amudprun.c
e-kayrakli: Done.

Issue #10840, opened 2018-08-21
spike: non-Chapel reproducer for mprotect(2) hang on ARM


gbtitus:
We think we're seeing programs running on ARM compute nodes hang in `mprotect(2)` if they do a lot of calls to that.  This happens with tasks=qthreads with guard pages enabled when a program creates lots of tasks, because Qthreads uses `mprotect()` to create guard pages on each task's stack.  The job for this spike is to write a non-Chapel reproducer for this behavior, which we can hand off to OS folks to study.

gbtitus:
I can reproduce a hang, but interestingly it's not a hang in `mprotect()`.  It's a hang in pthread condition variable processing, which is also implicated in the Chapel hangs because although there we typically see 2 threads in `mprotect()` the rest are in pthread condition variable processing.  Here is the final version of what I was working with: [many-mprotect-3.zip](https://github.com/chapel-lang/chapel/files/2349922/many-mprotect-3.zip).  I ran it within a job started like this:

```
qsub -I -l place=scatter,walltime=0:05:00,select=1:nodetype=mom-aarch64+1:nodetype=TX264 -V
```

I ran it 100 times, like this:

```
for i in $(seq 100) ; do echo ============ ; echo $i ; aprun -cc none -d64 -n1 -N1 -j0 ./many-mprotect-3 -t64 -n1000 -w500000 -p ; done
```

On the 11th run it hung and I used `pcmd kill -ABRT ...` to kill the process and get a core file:

```
============
11
64 threads, 1000 cycles, 500000 usecs condWait per cycle.
starting cycle 0
starting cycle 100
starting cycle 200
[... several minutes pass with no output ...]
Application 13638960 exit signals: Aborted
Application 13638960 resources: utime ~848s, stime ~14729s, Rss ~5788, inblocks ~0, outblocks ~0
```

I got tracebacks from `gdb` for all the threads and processed those to make all the data argument addresses the same so I could compare them.  1 thread was here:

```
[gbt@kay-aarch:tmp] more thr-1.txtp
(gdb) #0  0x0000000020011c58 in __pthread_mutex_lock (mutex=<ADDR>)
    at pthread_mutex_lock.c:81
#1  0x0000000020001d88 in main ()
(gdb)
```

1 thread was here:

```
[gbt@kay-aarch:tmp] more thr-9.txtp
(gdb) #0  0x00000000200125ac in __pthread_mutex_unlock_usercnt (
[...]

PR #11049, opened 2018-09-07
Disable guard pages on thunderx* processors

ronawho:
We've been seeing sporadic hangs in several tests. We still haven't identified
the root cause, but disabling guard pages seems to resolve the hang.

Related to https://github.com/chapel-lang/chapel/issues/10840

### DIFF
--- a/runtime/src/tasks/qthreads/tasks-qthreads.c
+++ b/runtime/src/tasks/qthreads/tasks-qthreads.c
@@ -529,10 +529,10 @@ static chpl_bool setupGuardPages(void) {
 
     // Setup guard pages. Default to enabling guard pages, only disabling them
     // under the following conditions (Precedence high-to-low):
-    // 1) Guard pages disabled at configure time
-    // 2) Guard pages not supported because of huge pages
-    // 3) QT_GUARD_PAGES set to a 'false' value
-    // 4) CHPL_STACK_CHECKS set (--no-stack-checks thrown at compilation time)
+    //  - Guard pages disabled at configure time
+    //  - Guard pages not supported because of huge pages
+    //  - QT_GUARD_PAGES set to a 'false' value
+    //  - CHPL_STACK_CHECKS set (--no-stack-checks thrown at compilation time)
     if (!CHPL_QTHREAD_HAVE_GUARD_PAGES) {
         guardPagesEnabled = false;
     } else if (chpl_getHeapPageSize() != chpl_getSysPageSize()) {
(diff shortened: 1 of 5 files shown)


ronawho:
Passed 1,000 trials of library/packages/LinearAlgebra/performance/SPA-perf with this (hung every ~3-4 w/o)

runtime/src/tasks/qthreads/tasks-qthreads.c
```diff
@@ -523,20 +523,24 @@ static void setupAvailableParallelism(int32_t maxThreads) {
...
+    //  - Guard pages disabled at configure time
+    //  - Guard pages not supported because of huge pages
+    //  - Guard pages not supported by processor
+    //  - QT_GUARD_PAGES set to a 'false' value
+    //  - CHPL_STACK_CHECKS set (--no-stack-checks thrown at compilation time)
     if (!CHPL_QTHREAD_HAVE_GUARD_PAGES) {
         guardPagesEnabled = false;
     } else if (chpl_getHeapPageSize() != chpl_getSysPageSize()) {
         guardPagesEnabled = false;
+    } else if (strncmp(armArch, CHPL_TARGET_ARCH, strlen(armArch)) == 0) {
```
ronawho: I'm also happy to switch this to `strstr("arm-thunderx", CHPL_TARGET_ARCH) != NULL` if you find that to be a little cleaner.

dmk42: Thanks.  I'm not going to ask for a change.

The way would normally do it is to make armArch an array instead of a pointer, and then I can say sizeof(armArch) - 1 and not have to do a strlen.

ronawho:
@gbtitus we're going to disable guard pages on thunderx for now. I know your investigation in https://github.com/chapel-lang/chapel/issues/10840 indicates that there's something more going on, but at the very least this should help eliminate hangs for users and our testing in the short term. This passed 5,000 trials of library/packages/LinearAlgebra/performance/SPA-perf, where it was hanging every ~3 trials before. 

I think there's more to investigate here, but at the very least there seems to be some correlation to guard pages and the hangs, and disabling guard pages seems to help eliminate the hangs.

gbtitus:
Yes, this seems like the right thing to do.  While the #10840 reproducer does imply something wrong in pthreads serialization support rather than in `mprotect()` as such, that was something of a surprising result.  And although we don't see anything wrong with that reproducer (do we?) we certainly do know that turning off guard pages eliminates the hangs, or at least very much reduces their frequency.

Issue #11052, opened 2018-09-07
Write documentation for strings


dmk42:
Currently, strings lack any documentation.  That leaves us with no place to put updates when we change things, like adding UTF-8 support.

We should document the API to make it clear what is available.

ronawho:
https://chapel-lang.org/docs/builtins/String.html ?

dmk42:
Thanks.  I wonder where that came from.  It didn't show up in a grep at all (for things like ASCII, string, isAlpha).

ronawho:
It's just chpldoc from the String module https://github.com/chapel-lang/chapel/blob/master/modules/internal/String.chpl

ronawho:
```
git grep ASCII modules/
modules/internal/String.chpl:    left to do. As of Chapel 1.17, only ASCII strings can be expected to work
modules/internal/String.chpl:    Future work involves support for both ASCII and unicode strings, and
modules/internal/String.chpl:     :returns: A string with the single character with the ASCII value `i`.
...
```

dmk42:
Never mind.  I was totally ignorant about the automatically-generated documentation, which is actually pretty good.

PR #11053, opened 2018-09-07
Remove the requirement that strings be ASCII in the spec

dmk42:
Currently the spec requires strings to be composed of ASCII characters.  This change removes that requirement from the spec.

Note that we lack string documentation in general, so this is the only place we can update to indicate that there is UTF-8 support.  In the long run, we should have some string documentation, as noted in #11052 .


dmk42:
@bradcray - this is trivial but I wanted you to see it.  Would you please review?

bradcray:
Looks good, thanks!

PR #11057, opened 2018-09-08
Update the header of the String documentation

dmk42:
Most of the String documentation has been kept up to date all along, but the beginning needs an update.  This change takes care of that.

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -71,9 +71,11 @@
 
     For testing, we use the following settings.
 
-    export LANG=en_US.UTF-8
-    export LC_COLLATE=C
-    unset LC_ALL
+    ``export LANG=en_US.UTF-8``
+
+    ``export LC_COLLATE=C``
+
+    ``unset LC_ALL``
 
     LANG sets the default character set.  LC_COLLATE overrides it for
     sorting, so that we get consistent results.  Anything in LC_ALL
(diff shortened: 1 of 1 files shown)


dmk42:
@bradcray - Would you mind taking a post-commit look at this?

modules/internal/String.chpl
```diff
@@ -63,12 +63,23 @@
...
+
+    ``export LANG=en_US.UTF-8``
+
+    ``export LC_COLLATE=C``
+
+    ``unset LC_ALL``
+
+    LANG sets the default character set.  LC_COLLATE overrides it for
+    sorting, so that we get consistent results.  Anything in LC_ALL
+    would override everything, so we unset it.
```
mppf: This is OK at saying what the implementation does but I think it should start by saying what users should expect. But I'm not sure we know that yet. E.g. I would want the "normal" Chapel environment to be UTF-8.

dmk42: Yes, this documents the current situation.  It starts by saying that the user selects the character set with the environment variables, and then gives an example based on how we test things.  We could certainly decide to force UTF-8 in the future.

mppf: Here is what I would recommend:

1. Look at the rendered string docs and fix remaining issues. Here are the issues I see:
   * there is a warning about casts from string to numeric types but no introduction that these are supported. Needs at least a 1-sentence description of what they do.
   * `record codePointIndex` is not documented

2. Let's introduce what the general idea is before talking about the current details. E.g. say "Chapel supports multibyte strings and is normally tested with UTF-8." and then go on to say the other stuff in a ..note.

3. We're missing docs about lengths and offsets in unicode strings.

I have created PR #11059 to make these changes. Please have a look.

PR #11059, opened 2018-09-09
Tidy up string documentation

mppf:
This is a follow-on to PR #11057 and cleans up the String chpldoc documentation further.

Reviewed by @dmk42 - thanks!

### DIFF
--- a/modules/internal/String.chpl
+++ b/modules/internal/String.chpl
@@ -36,50 +36,98 @@
 // It might be worth moving them here for documentation - KB Feb 2016
 
 /*
-  The following documentation shows functions and methods used to
-  manipulate and process Chapel strings.
+The following documentation shows functions and methods used to
+manipulate and process Chapel strings.
 
-  Besides the functions below, some other modules provide routines that are
-  useful for working with strings. The :mod:`IO` module provides
-  :proc:`IO.string.format` which creates a string that is the result of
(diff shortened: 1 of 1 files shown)


modules/internal/String.chpl
```diff
@@ -163,6 +211,23 @@ module String {
     var shortData : chpl__inPlaceBuffer;
   }
 
+  /*
+     A value of type :record:`codePointIndex` can be passed to several
+     `string` functions to indicate that the function should operate
+     with units of code points. See for example :proc:`~string.this`.
```
dmk42: The only thing we can do with a `codePointIndex` is pass it to `this()` and cast it back and forth with int.  I don't think we should say several here.

modules/internal/String.chpl
```diff
@@ -36,50 +36,98 @@
...
+
+The string type currently supports code point units as well. In some cases these
+units are indicated by the method name, for example with
+:proc:`~string.ulength`. Other methods, including :proc:`~string.this`, support
+arguments of type :record:`codePointIndex` (instead of `int`) in order to
+request code point units.
+
+.. note::
+
+  Support for graphemes units is not yet implemented.
```
dmk42: Maybe "not implemented at this time."  It isn't clear that we will or should, so saying "not yet" might be misleading.

PR #11064, opened 2018-09-09
Improve LinearAlgebra documentation

ben-albrecht:
This PR improves documentation for `LinearAlgebra` in a number of ways:

- Adds more information about building with and without dependencies
- Removes mentions of promotion flattening from documentation and primer
  - Redirects users of `matPlus` / `matMinus` to use `A + B` / `A - B` now.
- Improves documentation of `svd()`
- Removes documentation of `dot` from top section

### DIFF
--- a/modules/packages/LinearAlgebra.chpl
+++ b/modules/packages/LinearAlgebra.chpl
@@ -66,7 +66,8 @@ headers and libraries available would result in a compilation error.
 
   // example2.chpl
   var A = Matrix([2, 1],
-                 [1, 2], eltType=real);
+                 [1, 2],
+                 eltType=real);
   var (eigenvalues, eigenvectors) = eigvals(A, right=true);
 
 The program above uses :proc:`eigvals`, which depends on :mod:`LAPACK`.
@@ -1109,7 +1110,6 @@ proc cholesky(A: [] ?t, lower = true)
 
  */
 // TODO: example
-// TODO: LaTeX
 proc eigvals(A: [] ?t, param left = false, param right = false)
   where isRealType(t) && A.domain.rank == 2 && usingLAPACK {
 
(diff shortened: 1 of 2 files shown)


modules/packages/LinearAlgebra.chpl
```diff
@@ -17,32 +17,139 @@
...
 
 .. code-block:: bash
 
-  chpl --set blasImpl=none --set lapackImpl=none myProgram.chpl
+  # Building with BLAS dependency
+  chpl -I$PATH_TO_CBLAS_DIR \
+       -L$PATH_TO_BLAS_LIBS -lblas \
+       example3.chpl
+
+.. note::
```
ronawho: I think this is a duplicate note

modules/packages/LinearAlgebra.chpl
```diff
@@ -437,45 +493,45 @@ proc _array.T where isDefaultRectangularArr(this) && this.domain.rank == 2
...
-  compilerWarning('matPlus has been deprecated for _array.plus, ' +
-                  'try: A.plus(B)');
+  compilerWarning('matPlus has been deprecated. ' +
+                  'try: A + B');
   return A.plus(B);
 }
 
-/* Add matrices, maintaining dimensions */
+/* Element-wise addition. Same as ``A + B``. */
 proc _array.plus(A: [?Adom] ?eltType) where isDefaultRectangularArr(A) && isDefaultRectangularArr(this) {
```
ronawho: Seems unfortunate to have `_array` show up in user-facing docs. Not sure if there's anything we can do about that short of another post-processing script.

modules/packages/LinearAlgebra.chpl
```diff
@@ -1153,17 +1210,45 @@ proc eigvals(A: [] ?t, param left = false, param right = false)
...
+  containing the diagonal elements of :math:`\Sigma`, known as the
+  singular values.
+
+  For example:
+
+  .. code-block:: chapel
+
+    var A = Matrix([3, 2,  2],
+                   [2, 3, -2],
+                   eltType=real);
```
ronawho: Can you use:

```chpl
var A = Matrix([3.0, 2.0,  2.0],
               [2.0, 3.0, -2.0])
```

instead of `eltType=real`?

modules/packages/LinearAlgebra.chpl
```diff
@@ -17,32 +17,139 @@
...
+headers and libraries available would result in a compilation error.
+
+**Building programs with dependencies**
+
+.. code-block:: chpl
+
+  // example2.chpl
+  var A = Matrix([2, 1],
+                 [1, 2],
+                 eltType=real);
```
ronawho: Same question about real literals vs eltType=real

modules/packages/LinearAlgebra.chpl
ben-albrecht: Agreed. Probably another post-processing script or `chpldoc` addition.

modules/packages/LinearAlgebra.chpl
ben-albrecht: It's the same note as before, except for `BLAS` instead of `LAPACK`. Maybe I should just add a single note towards the end covering both of them?

modules/packages/LinearAlgebra.chpl
ben-albrecht: Yes, although I was implicitly using this as an opportunity to show what users can do with the `Matrix()` interface. Maybe showing one of the examples without the `eltType` argument would be good as well.

modules/packages/LinearAlgebra.chpl
ronawho: I would merge them into 1 note

modules/packages/LinearAlgebra.chpl
ronawho: Ah, ok. Up to you

Issue #11071, opened 2018-09-10
Develop consistent system for documenting module stability


ben-albrecht:
Many modules include notes about the work-in-progress nature of the module with varying degrees of detail about planned features, changes, and improvements. 

These messages are useful in making users aware of current limitations and future changes, but the inconsistency of their usage and level of detail makes it difficult to know how stable (or unstable) modules actually are.

## Examples

Here are a few excerpts from [1.17 docs](https://chapel-lang.org/docs/modules/packages/ZMQ.html):

List:

> This module is expected to change in the future.

Reflection:

> There are several ways in which this module could be improved upon:
>
>   - the methods here might be better as type methods, so you could use R.numFields() instead of numFields(R).
>   - getField does not yet return a mutable value.

Barriers:

> In the future, we expect to add more language-level support for a "task-team" concept. A task-team will more directly support collective operations such as barriers between the tasks within a team.
>
> The current implementation is designed for correctness, but is not expected to perform well at scale. We expect performance at scale to improve as this barrier implementation is optimized and as the task-team concept is implemented and optimized.

Spawn:

> New in version 1.12: Spawn module added.

DistributedBag:

> This package module is new in 1.16 and may contain bugs. The interface may change. The documentation is being incrementally revised and improved.

MPI:

> This is a prototype of the MPI interface for Chapel. The feature set, configuration support, and documentation are expected to improve over time.

ZMQ:

> Chapel's support for ZeroMQ is a work in progress and may not (yet) support the full functionality of the ZeroMQ C API.

## Exploring Solutions

It would be more informative to have a consistent system of specifying a module's stability. This could be implemented in a variety of ways.

To propose one solution: 

Module stability could be denoted by a numbering or keyword system, with various categories, similar to [NASA's technology readiness levels](https://www.nasa.gov/pdf/458490main_TRL_Definitions.pdf). As a starting point for discussion, maybe we would have 3 categories:

1. development - work-in-progress, not well-tested, very little guaranteed
2. unstable - well-tested, but interface may significantly change
3. stable - well-tested and established. Ready for production code.

With a system like this, we could implement Sphinx directives, such as `:unstable:`, which would insert some consistent boiler-plate note about the stability of the module in the documentation.

ben-albrecht:
A stability index may also be useful at the granularity of functions/methods and data structures.

ben-albrecht:
It might make sense to extend this to compiler and language features as well.

mppf:
[...]

PR #11073, opened 2018-09-10
Improve work-in-progress messages in module documentation

ben-albrecht:
This PR aims to reduce redundant text in WIP documentation notes, remove version numbers referenced in WIP messages, and remove planned feature lists when an issue link is already present.

This PR only updates messages that should be pretty straightforward and not require much discussion. Other messages might be best handled once a best practice is established for these WIP notes (https://github.com/chapel-lang/chapel/issues/11071).

### DIFF
--- a/modules/packages/Sort.chpl
+++ b/modules/packages/Sort.chpl
@@ -274,7 +274,7 @@ proc chpl_check_comparator(comparator, type eltType) {
 /* Basic Functions */
 
 /*
-   General purpose sorting interface. This procedure calls :proc:`quickSort`.
+   General purpose sorting interface.
 
    :arg Data: The array to be sorted
    :type Data: [] `eltType`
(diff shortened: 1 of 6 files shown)


modules/packages/Sort.chpl
```diff
@@ -274,10 +274,7 @@ proc chpl_check_comparator(comparator, type eltType) {
 /* Basic Functions */
 
 /*
-   General purpose sorting interface.
-
-   .. note:: Currently this method calls a sequential :proc:`quickSort`, but
-             this may change the future as other algorithms are implemented.
+   General purpose sorting interface. This procedure calls :proc:`quickSort`.
```
mppf: Why should this say it calls quickSort at all? Anyway if it continues to do so, say "currently calls".

PR #11075, opened 2018-09-10
Convert LCALS to use managed classes

daviditen:
Changed several global arrays from `unmanaged` to `owned`.  Made functions that take those
arrays as arguments use `owned`.  Made functions that take elements of those arrays as arguments
use blank (`borrowed`).

Updated some enum uses in a not often tested path to behave correctly when treated as integers.

Tested with `--memLeaks` for two passes with all valid variants and all kernels enabled.  No leaks were reported.  Tested with `start_test -performance` to match the way nightly runs it.


daviditen:
I haven't done a performance comparison, but I'd be surprised if there was a performance impact in anything with the possible exception of the `ADI` kernel.  The timers are tightly wrapped around the kernel loops and they don't use classes.

ADI uses a class to implement some funky access patterns so it's possible it could be impacted, but I don't think it goes in/out of scope or creates copies during the timed kernel, so I think it should be OK too.

daviditen:
Thanks for reviewing @mppf

PR #11078, opened 2018-09-10
Add initializers to language spec

benharsh:
This PR adds documentation for initializers to the spec. Semantics are primarily described in Classes.tex, and referred back to in other sections. The information in the spec subsumes the initializers technote, so this PR removes the technote.

Also removes documentation for extern classes and the ``outer`` functionality of nested types, which were deprecated due to initializer-related issues.

Testing:
- [x] examples/spec/


spec/Records.tex
```diff
@@ -186,23 +185,18 @@ \subsection{Record Initialization}
 is initialized through a call to the record's default initializer,
 passing no arguments.  The default initializer for a record is defined in
```
mppf: This doesn't seem right to me. The compiler will call whatever record initializer accepts 0 arguments, whether or not it's one the compiler generated.

Maybe

> A variable of a record type declared without an initialization expression is initialized with a call to the initializer for that record that accepts zero arguments. If multiple such initializers exist, an ambiguity error is raised.

spec/Records.tex
```diff
@@ -186,23 +185,18 @@ \subsection{Record Initialization}
...
 the same way as the default initializer for a class
-(\rsec{Default_Initialization}).
+(\rsec{The_Compiler_Generated_Initializer}).
 
-To construct a record as an expression,
+To create a record as an expression,
 i.e. without binding it to a variable, the \chpl{new} operator is
 required.  In this case, storage is allocated and reclaimed as for a record
 variable declaration (\rsec{Record_Storage}), except that the temporary record
 goes out of scope at the end of the enclosing statement.
```
mppf: That's neither what's implemented nor what we rely on today for `owned`. Can it say "the temporary record goes out of scope at the end of the enclosing block"?

spec/Records.tex
```diff
@@ -219,15 +213,12 @@ \subsection{Record Initialization}
 var idCounter = 0;
 record UniqueID {
   var id: int;
-  proc UniqueID() { idCounter += 1; id = idCounter; }
+  proc init() { idCounter += 1; id = idCounter; }
 }
 
 writeln(new UniqueID());  // create and use a record value without a variable
 writeln(new UniqueID());
```
mppf: This feels like a lost opportunity to demonstrate that `var tmp:UniqueID;` will call the `init()`.

mppf:
I wish that Records Common Operations section would start with something like this:

```
\subsection{Copy Initialization of Records}
\label{Copy_Initialization_of_Records}
\index{records!copy initialization}

When a new record variable is created based upon an
existing variable, we say that the new variable is
\emph{copy initialized} from the existing variable. For example:
\begin{chapel}
var x: MyRecordType;
var y = x; // this is called copy initialization
\end{chapel}

We say that y is \emph{copy initialized} from x. Since
x and y both exist after y is initialized and they are
different record variables, they must not be observed
to alias. That is, modifications to a field in x should not
cause the corresponding field in y to change.

Initializing a record variable with the result of a call
returning a record by value simply captures the result
of the call into the variable and does not cause copy
initialization. For example:
\begin{chapel}
proc returnsRecord() {
  var ret: MyRecordType;
  return ret;
}
var z = returnsRecord(); // captures result into z without copy initializing
\end{chapel}

\begin{future}
Specifying further details of copy initialization is future work.
\end{future}
```

spec/Generics.tex
```diff
@@ -858,23 +843,11 @@ \subsection{User-Defined Constructors}
...
-corresponding fields implicitly because these fields do not have initialization
-expressions. The second constructor also initializes implicitly
-the fields \chpl{t5} and \chpl{p5} because they are a type field
-and a parameter field. It does not initialize the fields \chpl{c5}
-and \chpl{v5} because they have initialization expressions, or
-the fields \chpl{x1} and \chpl{x5} because they are not generic fields
-(even though they are of generic types).
+The initializers are required to initialize fields \chpl{t1}, \chpl{p1},
+\chpl{c1}, and \chpl{v1}. Otherwise, field initializations may be omitted
+according to previously-described initializer semantics.
```
mppf: I really dislike this c1-p5 example and wish we would just throw it out. Your call though.

spec/Classes.tex
```diff
@@ -663,73 +600,453 @@ \subsection{User-Defined Constructors}
...
+fieldUsedBeforeInitialized.chpl:5: error: field "x" used before it is initialized
+\end{chapelprintoutput}
+
+\end{chapelexample}
+
+\subsubsection{Omitting Field Initializations}
+\label{Omitting_Field_Initializations}
+\index{classes!initializers!user-defined!omitting}
+
+Some classes have many fields, some of which might have default values. Instead
```
mppf: Maybe reword "Some classes have many fields,  some of which might have default values". Important part is default values, not many fields. 

Maybe "To support XYZ, the language allows ABC" would be a better structure.

spec/Classes.tex
```diff
@@ -663,73 +600,453 @@ \subsection{User-Defined Constructors}
...
+{x = 0.0, y = 0.0, msg = Origin}
+\end{chapeloutput}
+The first initializer initializes the values of fields \chpl{x} and \chpl{y},
+and the compiler inserts initialization for the \chpl{msg} field by using
+its default value. The second initializer initializes the \chpl{msg} field,
+and the compiler inserts initialization for fields \chpl{x} and \chpl{y}
+based on the type of those fields (\rsec{Default_Values_For_Types}).
+\end{chapelexample}
+
+Unchecked, this language feature could lead to confusing behavior for fields
```
mppf: This seems a bit weird to me - "editorializing" or something. Maybe this and the following paragraph could be combined. If that's not enough, you can use a `\chpl{rationale}` section.

spec/Classes.tex
```diff
@@ -340,13 +330,8 @@ \subsection{Nested Classes}
...
 
-A class or record defined within another class is a nested class (or record).
-
-Nested classes or records can refer to fields and methods in the outer class (or
-record) implicitly, or explicitly by means of an \chpl{outer} reference.
-
-A nested class (or record) can be referenced only within
-its immediately enclosing class (or record).
+A class defined within another class or record is a nested class. A nested
+class can be referenced only within its immediately enclosing class or record.
```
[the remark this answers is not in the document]
benharsh: I'm not sure, but I didn't want to get into the task of documenting nested types, I just wanted to remove the 'outer' stuff.

spec/Classes.tex
```diff
@@ -663,73 +600,453 @@ \subsection{User-Defined Constructors}
...
+value looks like initialization, but if written out of order might in fact be
+assignment.
+
+In order to reduce ambiguity and to ensure a well-defined order for
+side-effects, the language requires that all fields be initialized in field
+declaration order. This applies regardless of whether field initializations
+are omitted from the initializer body. If fields are initialized out of order,
+a compile-time error will be issued.
+
+\begin{chapelexample}{fieldsOutOfOrder.chpl}
```
mppf: This entire example seems to me to belong in a rationale section.

spec/Classes.tex
```diff
@@ -663,73 +600,453 @@ \subsection{User-Defined Constructors}
...
+  The compiler could conceivably attempt to analyze methods and functions to
+  determine which fields are used, and selectively allow method calls on
+  partially-initialized class instances. A simpler rule, which is easier to
+  remember and implement, is to forbid method calls on partially-initialized
+  instances.
+\end{rationale}
+
+Once the class is initialized, methods may be called and \chpl{this} may be
+passed to functions. Due to the support for omitted field initialization, there
+is potential for confusion regarding the overall status of initialization.
```
mppf: Here again there seems to be more rationale than I'd personally want, unless it's in a rationale section.

e.g.

> Methods may be called and this may be passed to functions only after the complete method is invoked.
> Rationale: Due to the support for omitted field initialization, there is potential for confusion regarding the overall status of initialization. This confusion is addressed in the design by requiring \chpl{complete} to mark the transition between partially initialized fields and fully initialized.

mppf: I'm not sure where it goes in here, but I keep expecting to see something about the run-time type of `this`.

benharsh: You'll see something about the runtime type in the upcoming section on inheriting classes.

spec/Classes.tex
```diff
@@ -663,73 +600,453 @@ \subsection{User-Defined Constructors}
...
+\subsubsection{Invoking Other Initializers}
+\label{Invoking_Other_Initializers}
+\index{classes!initializers!user-defined!other-initializers}
+
+In order to allow for code-reuse, an initializer may invoke another initializer
+implemented for the same type. Because the invoked initializer must operate on
+completely uninitialized memory, a compile-time error will be issued for field
+initialization before a call to \chpl{init}. Because each initializer
+either explicitly or implicitly invokes the \chpl{complete} method, all fields
+and methods may be used after a call to \chpl{init}.
```
mppf: a call to init -> such a call to init

spec/Classes.tex
```diff
@@ -739,28 +1056,26 @@ \subsection{The Compiler-Generated Constructor}
...
+there are no user-defined initializers for \chpl{C}, so \chpl{new} operators
+will invoke \chpl{C}'s compiler-generated initializer. The \chpl{x} argument
+of the compiler-generated initializer has the default value \chpl{0}.
 The \chpl{y} and \chpl{z} arguments have the default values \chpl{3.14} and
 \chpl{"Hello, World\!"}, respectively.
 
-\chpl{C} instances can be created by calling the compiler-generated constructor as follows:
+\chpl{C} instances can be created by calling the compiler-generated initializer as follows:
 \begin{itemize}
 \item The call \chpl{new C()} is equivalent to \chpl{C(0,3.14,"Hello, World\!")}.
```
mppf: it seems it should say "is equivalent to new C(...)"

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+\index{initializer-inheritance}
+
+User-defined initializers also allow for control over initialization of parent
+classes. All the fields of the parent type must be initialized before any
+fields of the child type, otherwise a compile-time error is issued. This allows
+for parent fields to be used in the definition of child fields. An initializer
+may invoke a parent's initializer using the \chpl{super} keyword.
+
+If the user does not explicitly call the parent's initializer, the compiler
+will insert a call to the parent initializer with zero arguments.
```
mppf: at the start of the initializer, right?

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+initialized.
+
+The second initializer implicitly invokes the parent's initializer with zero
+arguments, and then uses the parent's fields to initialize its own fields.
+\end{chapelexample}
+
+As stated earlier, the compiler will insert a zero-argument call to the
+parent's initializer if the user has not explicitly written one themselves.
+The exception to this rule is if the initializer body invokes another
+initializer. This other initializer will either contain an implicit or explicit
```
mppf: another initializer -> another initializer on the current type (or something)

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+initializer because it explicitly invokes another initializer.
+\end{chapelexample}
+
+\subsubsection{Calling Methods on Parent Classes}
+\label{Calling_Methods_on_Parent_Classes}
+\index{classes!initializers!initializer-inheritance!dynamic-this}
+
+Because parent classes are initialized before child classes, the
+partially-initialized memory can be treated as if it were the parent class
+once super.init() returns. The dynamic type of \chpl{this} is the parent's
```
mppf: \chpl{super.init()}

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+type until the \chpl{complete} method is invoked (except when local fields
+are initialized and used). This means that before the \chpl{complete} method
+is invoked, parent methods may be called on \chpl{this} and \chpl{this} may be
+passed to functions as the parent type.
```
mppf: I thought above we were saying that can't happen... maybe check & add a cross reference from the part discussing complete to here.

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+\end{chapelprintoutput}
+\end{chapelexample}
+
+\subsubsection{The Compiler Generated Initializer for Inheriting Classes}
+\label{The_Compiler_Generated_Initializer_for_Inheriting_Classes}
+\index{classes!initializers!initializer-inheritance!compiler-generated}
+
+The compiler-generated initializer for inheriting classes will have arguments
+with default values and names based on the field declarations in the parent
+class.
```
mppf: Maybe also say that the formals are in order of parent type first?

spec/Classes.tex
```diff
@@ -587,41 +526,43 @@ \section{Class New}
...
+is called, the usual function resolution mechanism (\rsec{Function_Resolution})
+is applied with the exception that an initializer may not be virtually
+dispatched.
+
+A user-defined initializer is responsible for initializing all fields. An
+initializer may omit initialization of fields, but all fields that are
+initialized must be initialized in declaration order.
+
+Initializers for generic classes (\rsec{Generic_Types}) handle \chpl{type} and
+\chpl{param} fields without defaults differently and may need to satisfy
```
lydia-duncan: I think you should mention fields without types or default values here as well.  Otherwise a user might not think to check the generics section if they only had generic var/const fields

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+
+\begin{rationale}
+Without this rule the compiler could insert default initialization for field
+\chpl{y} before \chpl{z} is explicitly initialized. The following statement
+would then be {\em assignment} to field \chpl{y}, despite appearing to be
+initialization. This subtle difference may be confusing and surprising, and
+is avoided by requiring fields to be initialized in field declaration order.
+\end{rationale}
+
+\subsubsection{The complete Method}
```
lydia-duncan: There's a part of me that wonders if this section should be renamed.  Maybe something like "Limitations in Initializers and the complete Method"?  The section covers the need for `complete` but someone reading the initializers section of the spec might not realize they need to pay attention to it, especially if they are unfamiliar with the `complete` method in the first place

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+
+These rules allow all methods and functions to assume that class instances
+have been initialized, provided their value is not \chpl{nil}.
+
+\begin{rationale}
+  The compiler could conceivably attempt to analyze methods and functions to
+  determine which fields are used, and selectively allow method calls on
+  partially-initialized class instances. A simpler rule, which is easier to
+  remember and implement, is to forbid method calls on partially-initialized
+  instances.
```
lydia-duncan: I might end this paragraph with explicitly repeating which strategy is recommended/described by the spec.

spec/Classes.tex
benharsh: "Calling Methods in Initializers" ? That doesn't really cover passing 'this' to functions though...

"Limitations on Instance Usage in Initializers" ? Feels wordy to me

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+
+Methods may be called and \chpl{this} may be passed to functions only after the
+built-in \chpl{complete} method is invoked.
+
+\begin{rationale}
+Due to support for omitted field initialization, there is potential for
+confusion regarding the overall status of initialization. This confusion is
+addressed in the design by requiring \chpl{complete} to explicitly mark the
+transition between partially and fully initialized instances.
+\end{rationale}
```
lydia-duncan: I would move this rationale to the beginning of the section

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+
+This method may not be overridden. If any fields have not been initialized by
+the time the \chpl{complete} method is invoked, they will be considered omitted
+and the compiler will insert initialization statements as described earlier. If
+the user does not invoke the \chpl{complete} method explicitly, the compiler
+will insert a call to \chpl{complete} at the end of the initializer. Note that
+even if the user explicitly initializes every field, the \chpl{complete} method
```
lydia-duncan: I would split this into a separate paragraph.  We can have "note" sections, right?

spec/Classes.tex
benharsh: I felt that it was better to introduce the context for the problem and the solution before rationalizing it.

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+{x = 5.0, y = 5.0, z = 5.0}
+\end{chapeloutput}
+The first initializer must invoke the \chpl{complete} method at the end of
+the if-branch in order to match the state at the end of the else-branch.
+\end{chapelexample}
+
+\subsubsection{Miscellaneous Field Initialization Rules}
+
+Fields may not be initialized within loop statements, parallel statements, or
+on statements.
```
lydia-duncan: Fields can be initialized in `on` statements.  It looks like the comment in test/classes/initializers/phase1/hasOnFieldInit.chpl is out of date (sorry, that's probably my fault)

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
 ({x = 0, y = 3.14, z = Hello, World!}, {x = 2, y = 3.14, z = Hello, World!}, {x = 0, y = 3.14, z = }, {x = 2, y = 3.14, z = }, {x = 0, y = 0.0, z = })
 \end{chapeloutput}
-there are no user-defined constructors for \chpl{C}, so \chpl{new} operators
-will invoke \chpl{C}'s compiler-generated constructor. The \chpl{x} argument
-of the compiler-generated constructor has the default value \chpl{0}.
+there are no user-defined initializers for \chpl{C}, so \chpl{new} expressions
+will invoke \chpl{C}'s compiler-generated initializer. The \chpl{x} argument
+of the compiler-generated initializer has the default value \chpl{0}.
 The \chpl{y} and \chpl{z} arguments have the default values \chpl{3.14} and
 \chpl{"Hello, World\!"}, respectively.
```
lydia-duncan: I feel like these sentences should be "the default value(s) of ", but I might also be crazy . . .

spec/Classes.tex
benharsh: Going with "Limitations on Instance Usage in Initializers" for now.

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+\begin{chapelcompopts}
+--no-warnings
+\end{chapelcompopts}
+\begin{chapeloutput}
+{x = 1.0, y = 1.0, z = 1.0}
+\end{chapeloutput}
+The second initializer leverages the first initializer to initialize all fields
+with the same value. After the \chpl{init} call the type is fully initialized,
+the \chpl{complete} method has been invoked, and so \chpl{this} can be passed
+to the \chpl{writeln} function.
```
lydia-duncan: Assuming the "parent initializer call" section exists (which I suspect it does but I haven't finished reading yet), you might want to call it out and link to it at the end of this section

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+\label{The_postinit_Method}
+\index{classes!initializers!postinit}
+
+The compiler-generated initializer is powerful and flexible, but cannot satisfy
+all initialization patterns desired by users. One way for users to leverage
+the compiler-generated initializer while adding their own functionality is to
+implement a method named \chpl{postinit}. The \chpl{postinit} method may also
+be implemented for types with user-defined initializers.
+
+The compiler will insert a call to a \chpl{postinit} method after the
```
lydia-duncan: I think it should be "*the* postinit method", since there will be only one.

spec/Classes.tex
benharsh: I think "the default value(s)" is OK, at least to my ear.

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+initialized.
+
+The second initializer implicitly invokes the parent's initializer with zero
+arguments, and then uses the parent's fields to initialize its own fields.
+\end{chapelexample}
+
+As stated earlier, the compiler will insert a zero-argument call to the
+parent's initializer if the user has not explicitly written one themselves.
+The exception to this rule is if the initializer body invokes another
+initializer on the current type. This other initializer will either contain an
```
lydia-duncan: Maybe link to the sibling case here?

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+initializer because it explicitly invokes another initializer.
+\end{chapelexample}
+
+\subsubsection{Calling Methods on Parent Classes}
+\label{Calling_Methods_on_Parent_Classes}
+\index{classes!initializers!initializer-inheritance!dynamic-this}
+
+Once \chpl{super.init()} returns, the dynamic type of \chpl{this} is the
+parent's type until the \chpl{complete} method is invoked (except when the
+child's fields are initialized and used). As a result, the parent's methods
```
lydia-duncan: Maybe link to the section defining the "complete" method here?

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+may be called and \chpl{this} may be passed to functions as the parent type.
```
lydia-duncan: Really nitty, but I would say "may be passed to functions as though it were of the parent type."

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+  proc foo() {
+    writeln("Parent.foo");
+  }
+}
+
+class Child : Parent {
+  var z : real;
+
+  proc init(x: real, y: real, z: real) {
+    super.init(x, y); // compiler-generated initializer
```
lydia-duncan: "Parent's compiler-generated initializer" or "compiler-generated initializer for Parent" or something like that

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+\end{chapel}
+\begin{chapelpost}
+writeln(c);
+\end{chapelpost}
+\begin{chapelcompopts}
+--no-warnings
+\end{chapelcompopts}
+Once the parent's initializer is finished, the parent method
+\chpl{foo} may be called. After the \chpl{complete} method is invoked, a call to \chpl{foo}
+resolves to the Child's overriden implementation:
```
lydia-duncan: Is there an override section we can link to here?

benharsh: How about I don't say "overriden" at all and just say "the child's implementation" ?

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+writeln(B);
+writeln(C);
+\end{chapelpost}
+\begin{chapelcompopts}
+--no-warnings
+\end{chapelcompopts}
+\begin{chapeloutput}
+{x = 0.0, y = 0.0, z = 0.0}
+{x = 1.0, y = 2.0, z = 3.0}
+{x = 0.0, y = 10.0, z = 0.0}
```
lydia-duncan: I would either show this output, or include a comment on the declaration showing which fields the numbers in `var B` will go to

spec/Classes.tex
lydia-duncan: But you have to use it in your example, so you might as well link to what it means

spec/Classes.tex
lydia-duncan: But this rationale is also introducing context.  At least, it seems a little out of place here at the bottom of the section

spec/Classes.tex
```diff
@@ -770,6 +1085,314 @@ \subsection{The Compiler-Generated Constructor}
...
+\label{The_postinit_Method_for_Inheriting_Classes}
+\index{classes!initializers!initializer-inheritance!postinit}
+
+The \chpl{postinit} method on inheriting classes allows users to invoke
+the parent's \chpl{postinit} method using the \chpl{super} keyword. If the
+user does not explicitly invoke the parent's \chpl{postinit}, the compiler
+will insert the call at the top of the user's \chpl{postinit} method. If the
+parent type has a \chpl{postinit} method but the inheriting class does not,
+the compiler will generate a \chpl{postinit} method that simply invokes the
+parent's \chpl{postinit} method.
```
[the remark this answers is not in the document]
lydia-duncan: I'd also like to see a link to the original postinit section early on in this paragraph, but I'm not sure about the natural place to put it

spec/Classes.tex
```diff
@@ -739,37 +1063,348 @@ \subsection{The Compiler-Generated Constructor}
...
+  proc postinit() {
+    writeln("Parent.postinit: ", a, ", ", b);
+  }
+}
+
+class Child : Parent {
+  var x, y : real;
+  proc postinit() {
+    // compiler inserts "super.postinit();"
+    writeln("Child.postinit: ", a, ", ", b);
```
lydia-duncan: Did you mean to have this one refer to `x` and `y` instead of the parent's fields?

lydia-duncan: That would also remove the need for the writeln at the end of the example

spec/Domain_Maps.tex
```diff
@@ -144,7 +144,7 @@ \section{Domain Maps for Domain Types}
 
 The following syntactic sugar is provided within the \chpl{dmapped} clause.
 If a \chpl{dmapped} clause starts with the name of a domain map class,
-it is considered to be a constructor expression as if preceded by
+it is considered to be an initializer expression as if preceded by
```
lydia-duncan: Should this be "an initializer *call*" instead of "an initializer *expression*"?  That phrasing just feels weird to me

spec/Classes.tex
```diff
@@ -663,73 +604,456 @@ \subsection{User-Defined Constructors}
...
+{x = 1.0, y = 1.0, r = 0.0, theta = 0.0}
+\end{chapeloutput}
+
+The compiler identifies field \chpl{r} as the latest field in both branches,
+and inserts omitted field initialization statements as needed to ensure that
+fields \chpl{x}, \chpl{y}, and \chpl{r} are all initialized by the end of the
+conditional.
+\end{chapelexample}
+
+Conditionals may also contain calls to sibling and parent initializers,
```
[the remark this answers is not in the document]
benharsh: I will remove the 'sibling' terminology for now.

spec/Generics.tex
```diff
@@ -819,6 +803,7 @@ \subsection{User-Defined Constructors}
     this.p1 = p1;
     this.c1 = c1;
     this.v1 = v1;
+    // compiler inserts initialization for remaining fields
```
lydia-duncan: I recognize you inherited this, but I think technically `c5` and `v5` are not generic.

I'm confused as to why they are suddenly named using 5, but I don't think that's important at all

spec/Domain_Maps.tex
benharsh: "initialization expression" ?

spec/Interoperation.tex
```diff
@@ -308,47 +308,7 @@ \subsubsection{Referring to External Structs Through Pointers}
...
-     int y;
-   } _D, *D;
-\end{chapel}
-where the Chapel compiler would not know about the 'y' field and
-therefore could not refer to it or manipulate it.
-
-If the optional \sntx{external-name} is supplied, then it is used verbatim as
-the exported class symbol.
-
+%INIT TODO: clearly state extern classes are not supported?
```
lydia-duncan: Be sure to call out in your PR message that you also handle the removal of extern classes from the spec

spec/Records.tex
```diff
@@ -161,9 +159,9 @@ \section{Record Variable Declarations}
 
 In the context of a function body, a record variable declaration
 causes storage to be allocated sufficient to store all of the fields in that
-record type.  The record variable is initialized through a call to its
-default initializer.  The default initializer for a record is defined in the
-same way as the default initializer for a class (\rsec{Default_Initialization}).
+record type.  The record variable is initialized with a call to an
+initializer that accepts zero actual arguments. If multiple such initializers
+exist, an ambiguity error is raised.
```
lydia-duncan: I would add an example of a case where ambiguity occurs, to make it obvious

spec/Records.tex
```diff
@@ -219,32 +212,32 @@ \subsection{Record Initialization}
 var idCounter = 0;
 record UniqueID {
   var id: int;
-  proc UniqueID() { idCounter += 1; id = idCounter; }
+  proc init() { idCounter += 1; id = idCounter; }
 }
 
+var firstID : UniqueID(); // invokes zero-argument initializer
```
lydia-duncan: Can we drop the parenthesis?

spec/Records.tex
```diff
@@ -219,32 +212,32 @@ \subsection{Record Initialization}
...
 is assigned to \chpl{timestampCustom}.  It effectively
-initializes \chpl{timestampCustom} via a call to the constructor with desired
+initializes \chpl{timestampCustom} via a call to the initializer with desired
 arguments. The records created with \chpl{new UniqueID()} are discarded after
 they are used.
 \end{chapelexample}
 
-As with classes, the user can provide his own constructors
-(\rsec{User_Defined_Constructors}).  If any user-defined constructors are
+As with classes, the user can provide their own initializers
```
lydia-duncan: ❤️

spec/Records.tex
benharsh: How about I remove the line about ambiguity errors, since the Classes section says it behaves like normal function resolution?

spec/Domain_Maps.tex
lydia-duncan: Works for me!

spec/Records.tex
lydia-duncan: That works for me

spec/Classes.tex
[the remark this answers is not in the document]
benharsh: I ended up linking forward from the introduction of postinit.

spec/Generics.tex
benharsh: Going to leave this for another time, in the interest of the freeze deadline.

spec/Classes.tex
mppf: Pretty sure I had asked Ben to add the rationale section later. On the spec itself, as a style guideline, I think it should say what the language design is and then, in a rationale section, say why we have that design.

Issue #10458, opened 2018-07-23
Write new setenv scripts which reproduce current Chapel Cray RPMs (new scripts to build customizable Cray module from source)


awallace-cray:
Write (and test) new build_configs.py setenv scripts that will reproduce the same Chapel Cray-XC/XE RPMs we currently build for nightly and release, except using the below instead of the currently-used Jenkins/ chapel-code build scripts:
1. https://github.com/chapel-lang/chapel/issues/10120 - build_configs.py changes to work with "Chapel Cray module build script"
2. https://github.com/chapel-lang/chapel/issues/10109 - Working version of "Chapel Cray module build script"
3. https://github.com/chapel-lang/chapel/issues/10119 - Working version of "Chapel Cray module packaging script"

awallace-cray:
FYI, here are the 192 build_configs (aka flavors) in a Cray-XC RPM. This only counts the runtime makes- Chapel compiler and tools like mason are not included.
 
i | COMP | COMM | TASK | LAUNCH | SUB | <ugni> | NET | CPU | AUXFS | LOCALE | make cmd
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
001 | gnu | none | UNSET | none | none | FALSE | aries | UNSET | none | UNSET | compiler
002 | cray | gasnet | UNSET | pbs-aprun | aries | FALSE | aries | UNSET | none | flat | --
003 | cray | gasnet | UNSET | pbs-aprun | aries | FALSE | aries | UNSET | lustre | flat | --
004 | cray | gasnet | UNSET | pbs-aprun | aries | FALSE | aries | UNSET | none | knl | --
005 | cray | gasnet | UNSET | pbs-aprun | aries | FALSE | aries | UNSET | lustre | knl | --
006 | cray | gasnet | UNSET | aprun | aries | FALSE | aries | UNSET | none | flat | --
007 | cray | gasnet | UNSET | aprun | aries | FALSE | aries | UNSET | lustre | flat | --
008 | cray | gasnet | UNSET | aprun | aries | FALSE | aries | UNSET | none | knl | --
009 | cray | gasnet | UNSET | aprun | aries | FALSE | aries | UNSET | lustre | knl | --
010 | cray | gasnet | UNSET | none | aries | FALSE | aries | UNSET | none | flat | --
011 | cray | gasnet | UNSET | none | aries | FALSE | aries | UNSET | lustre | flat | --
[...]

PR #11080, opened 2018-09-10
build_configs: Fix rpmspec bug which had lost the release_info file.

awallace-cray:
I forgot a step the last time I reorganized how the cray-internal
Chapel RPM would be generated, and that oversight caused the `release_info`
file to be silently omitted from the installed Chapel module.

If release_info is missing, then `module help chapel` will not display
the expected information. There isn't even an error.

This change fixes my oversight and restores the release_info file to
the cray-internal Chapel RPM produced by the build_configs/cray-internal
scripts.

See https://github.com/chapel-lang/chapel/issues/10458#issuecomment-418917800


gbtitus:
Looks good to me!  👍

PR #11081, opened 2018-09-10
Document the sorting options for CS added in PR #9877

bradcray:
It seems when the options for turning off sorting for CS were added in PR #9877, we failed to document them.  Adding that documentation here.

### DIFF
--- a/modules/layouts/LayoutCS.chpl
+++ b/modules/layouts/LayoutCS.chpl
@@ -69,7 +69,11 @@ For example:
 
 This domain map is a layout, i.e. it maps all indices to the current locale.
 All elements of a CS-distributed array are stored
-on the locale where the array variable is declared.
+on the locale where the array variable is declared.  By default, the CS
+layout stores sparse indices in sorted order, though this default can
+be changed for a program by compiling with `-sLayoutCSDefaultToSorted=false`, 
+or for a given domain by passing `sortedIndices=false` as an argument to the
+CS() initializer.
 */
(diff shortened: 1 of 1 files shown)


modules/layouts/LayoutCS.chpl
```diff
@@ -69,7 +69,11 @@ For example:
 
 This domain map is a layout, i.e. it maps all indices to the current locale.
 All elements of a CS-distributed array are stored
-on the locale where the array variable is declared.
+on the locale where the array variable is declared.  By default, the CS
+layout stores sparse indices in sorted order.  However, this default can
+be changed for a program by compiling with ``-sLayoutCSDefaultToSorted=false``, 
+or for a specific domain by passing ``sortedIndices=false`` as an argument
+to the ``CS()`` initializer.
```
ben-albrecht: Bonus points for a code-block snippet demonstrating creation of CS arrays with and without sorted indices (so users don't need to parse the full paragraph to learn this).

bradcray: On principle, I'm not adding code blocks to documentation until they are automatically tested (because I believe they'll only get stale otherwise).

ben-albrecht: OK.

ben-albrecht: FWIW, I've opened an issue documenting our desire to add this: https://github.com/chapel-lang/chapel/issues/11090

PR #11084, opened 2018-09-11
shape preservation in the spec

vasslitvinov:
Add the specification of shape preservation

* Improve how promotion is presented.
In particular, move the details of promotion over arrays
from the Arrays chapter to Data Parallelism chapter.
Make arrays reference that subsection so that we are consistent
in defining promotion for:

  - iterators
  - ranges
  - domains
  - arrays

* State that for-, forall- and promoted expressions
are iterators yielding one element at a time.
They become arrays when used to initialize a variable.
Remove Sung's earlier trial paragraph

* Create a table defining the domain of such an array
based on what kind of thing causes promotion.
The table is where promotion is defined.

* Have for- and forall-expressions reference that table
by pointing to the promotion section. If the for/forall expression
has a filtering predicate, then the resulting array has a
1-based one-dimensional domain.

* Move array-promotion sub-section from "Returning Arrays" section
to "Array Arguments" section, as it is fone for domain-promotion.
(The diffs show that the "Returning Arrays" section moved.)

* While there, remove \chpl from \subsubsection
as Latex gets upset otherwise.

### DIFF
--- a/spec/Data_Parallelism.tex
+++ b/spec/Data_Parallelism.tex
@@ -169,14 +169,25 @@ convenience.
 The handling of the outer variables within the forall expression and
 the role of \sntx{task-intent-clause} are defined in \rsec{Forall_Intents}.
 
-\subsection{Execution and Serializability}
-\label{Forall_Expression_Execution_and_Serializability}
+\subsection{Execution}
+\label{Forall_Expression_Execution}
 \index{expressions!forall!semantics}
 
-The forall expression executes a forall loop (\rsec{Forall}),
-evaluates the body expression on each iteration of the loop, and
-returns the resulting values as a collection.  The size and shape of
-that collection are determined by the iteratable-expression.
+The forall expression is an iterator that executes a forall loop (\rsec{Forall}),
+evaluates the body expression on each iteration of the loop,
+and yields each resulting value.
+
+When a forall expression is used to initialize a variable, such as
+%\begin{chapel}
+%var X = forall /*iteratable-expression*/ do /*expression*/;
+%\end{chapel}
+\\ %formatting
+{\tt \mbox{$$ $$ $$ $$ $$} {\bf var} X = {\bf forall} /*iteratable-expression*/ {\bf do} /*expression*/;}
+\\ %formatting
+the variable is an array. The array's domain is defined by the
+\sntx{iterable-expression} following the same rules as for promotion,
+both in the regular case \rsec{Promotion} and in the
+zipper case \rsec{Zipper_Promotion}.
 
 \begin{chapelexample}{forallExpr.chpl}
 The code
@@ -209,6 +220,9 @@ by a forall expression and does not have an
 else clause.  Such an if expression filters the iterations of the
 forall expression.  The iterations for which the condition does not
 hold are not reflected in the result of the forall expression.
+
+When a forall expression with a filtering predicate is captured into
+a variable, the resulting array has a 1-based one-dimensional domain.
 
 \begin{chapelexample}{forallFilter.chpl}
 The following expression returns every other element starting with the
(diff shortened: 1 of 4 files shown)


spec/Data_Parallelism.tex
```diff
@@ -169,14 +169,25 @@ \subsection{Syntax}
...
+evaluates the body expression on each iteration of the loop,
+and yields each resulting value.
+
+When a forall expression is used to initialize a variable, such as
+%\begin{chapel}
+%var X = forall /*iteratable-expression*/ do /*expression*/;
+%\end{chapel}
+\\ %formatting
+{\tt \mbox{$$ $$ $$ $$ $$} {\bf var} X = {\bf forall} /*iteratable-expression*/ {\bf do} /*expression*/;}
+\\ %formatting
```
mppf: I'd probably just make this a complete example, such as
```
\begin{chapel}
var A = forall iterableExpression() do valueToYield();
\end{chapel}
```

the formatting here seems finnicky and unnecessary.

spec/Data_Parallelism.tex
```diff
@@ -169,14 +169,25 @@ \subsection{Syntax}
...
-\label{Forall_Expression_Execution_and_Serializability}
+\subsection{Execution}
+\label{Forall_Expression_Execution}
 \index{expressions!forall!semantics}
 
-The forall expression executes a forall loop (\rsec{Forall}),
-evaluates the body expression on each iteration of the loop, and
-returns the resulting values as a collection.  The size and shape of
-that collection are determined by the iteratable-expression.
+The forall expression is an iterator that executes a forall loop (\rsec{Forall}),
```
mppf: "The forall expression" reads weirdly to me, maybe, "A forall expression is".

spec/Data_Parallelism.tex
```diff
@@ -169,14 +169,25 @@ \subsection{Syntax}
...
+and yields each resulting value.
+
+When a forall expression is used to initialize a variable, such as
+%\begin{chapel}
+%var X = forall /*iteratable-expression*/ do /*expression*/;
+%\end{chapel}
+\\ %formatting
+{\tt \mbox{$$ $$ $$ $$ $$} {\bf var} X = {\bf forall} /*iteratable-expression*/ {\bf do} /*expression*/;}
+\\ %formatting
+the variable is an array. The array's domain is defined by the
```
mppf: the variable is an array -> the variable will be inferred to have array type

spec/Data_Parallelism.tex
```diff
@@ -405,6 +419,32 @@ \section{Promotion}
 is performed with getter methods~(\rsec{Getter_Methods}), field
 access can also be promoted.
 
+If the original function returns a value, the corresponding promoted
+expression is an iterator yielding each computed value in turn.
```
mppf: returns a value ie. this doesn't apply to functions returning a reference? Maybe you mean "if the original function does not return void" or "returns at all"?

spec/Data_Parallelism.tex
```diff
@@ -405,6 +419,32 @@ \section{Promotion}
...
+
+When a promoted expression is used to initialize a variable, such as
+\chpl{var X = A.x;} in the above example, the variable is an array.
```
mppf: the variable is an array -> the variable's type will be inferred to be an array

spec/Data_Parallelism.tex
```diff
@@ -405,6 +419,32 @@ \section{Promotion}
...
+range    &  one-dimensional domain built from that range \\
+iterator &  1-based one-dimensional domain \\
+\hline
+\end{tabular}
+\end{center}
+
+\begin{future}
+If the iterator iterates over an array, a domain or a range,
+we may want the resulting array's domain to be the same as
+if the input expression were that array/domain/range.
```
mppf: I think the real future work here is to allow iterator authors to specify the shape in their iterator. Say that here?

spec/Expressions.tex
```diff
@@ -1514,11 +1514,21 @@ \section{For Expressions}
...
-are determined by the iteratable-expression.
+and yields each resulting value.
+
+When a for expression is used to initialize a variable, such as
+%\begin{chapel}
+%var X = for /*iteratable-expression*/ do /*expression*/;
+%\end{chapel}
+\\ %formatting
+{\tt \mbox{$$ $$ $$ $$ $$} {\bf var} X = {\bf for} /*iteratable-expression*/ {\bf do} /*expression*/;}
+\\ %formatting
```
mppf: same comment about formatting here

PR #11077, opened 2018-09-10
Add documentation for the HDF5 package module

daviditen:
Add some header documentation for the HDF5 module and C_HDF5 submodule.  Point
at the official HDF5 web site to document the extern interface.  Add HDF5 to
the makefile to start generating the documentation for it.

PR #11085, opened 2018-09-11
Allow chpldoc to ignore extern blocks

mppf:
This PR adjusts the parser to allow extern blocks even with CHPL_LLVM=none when running chpldoc. While there, it moves this error handling out of the parser 
and into buildExternBlockStmt.

Even if chpldoc were to document extern blocks, I think it's unlikely it would 
need to parse them. Of course we can revisit this change if there is a
compelling use case.

This PR also re-enables the HDF5 module documentation (see PRs #11077 #11082 
#11083).

- [x] full local testing with CHPL_LLVM=llvm
- [x] full local testing with CHPL_LLVM=none

### DIFF
--- a/compiler/AST/build.cpp
+++ b/compiler/AST/build.cpp
@@ -25,6 +25,7 @@
 #include "CatchStmt.h"
 #include "config.h"
 #include "DeferStmt.h"
+#include "docsDriver.h"
 #include "driver.h"
 #include "files.h"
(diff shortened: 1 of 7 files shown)


mppf:
@lydia-duncan - I've added the test. Maybe @bradcray has an opinion about whether we should try to get it into the release.

PR #11087, opened 2018-09-11
Language Evolution Updates

mppf:
Documents, at a very high level:
 * initializers
 * delete-free changes
 * override

### DIFF
--- a/doc/rst/language/evolution.rst
+++ b/doc/rst/language/evolution.rst
@@ -8,6 +8,231 @@ capture significant language changes that have the possibility of breaking
 existing user codes or code samples from old presentations or papers that
 predated the changes.
 
+Note that the compiler flag ``--warn-unstable`` is available and can be
+useful when migrating programs to the current version of the language.
+The purpose of this flag is to identify  portions of a program that use a
+language feature that has changed meaning.
+
+version 1.18, September 2018
+----------------------------
+
+Version 1.18 includes many language changes that address problems with
+classes.
+
+In summary:
+
+ * constructors are deprecated and replaced with initializers
+   See :ref:`readme-evolution.initializers-replace-constructors`
+ * memory management for class types has changed
+   See :ref:`readme-evolution.class-memory-management`
+ * `override` is now required on overriding methods
+   See :ref:`readme-evolution.mark-overriding`
+
+.. _readme-evolution.initializers-replace-constructors:
+
+initializers replace constructors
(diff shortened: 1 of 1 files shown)


doc/rst/language/evolution.rst
```diff
@@ -8,6 +8,231 @@ capture significant language changes that have the possibility of breaking
...
+   See :ref:`readme-evolution.class-memory-management`
+ * `override` is now required on overriding methods
+   See :ref:`readme-evolution.mark-overriding`
+
+.. _readme-evolution.initializers-replace-constructors:
+
+initializers replace constructors
+*********************************
+
+Code that was contained user-defined constructors will need to be updated
```
benharsh: "that was contained" --> "that contained"

doc/rst/language/evolution.rst
```diff
@@ -8,6 +8,231 @@ capture significant language changes that have the possibility of breaking
...
+    proc init(x: real, y: real) {
+      this.x = x;
+      this.y = y;
+      writeln("In Point.init(x,y)");
+    }
+  }
+  var a:Point;
+  var b = new Point(1.0, 2.0);
+
+
```
benharsh: At a glance, this example just looks like changing the name of the method. I don't think users will assume that's the extent of the change, but they might be left wanting more.

Should we say something about how there are a number of different semantics for initializers, and that the user should read the language spec for more info?

doc/rst/language/evolution.rst
```diff
@@ -8,6 +8,232 @@ capture significant language changes that have the possibility of breaking
...
+  class C {
+    var x: int;
+  }
+  proc main() {
+    var instance: C = new C(1);
+    // above line could be written as
+    //   var instance: borrowed C = new borrowed C(1);
+    // or as
+    //   var instance: borrowed C = (new owned C(1)).borrow();
```
benharsh: I feel like this document exists to provide a rough outline of the changes, so these four comment lines feel like too much information.

doc/rst/language/evolution.rst
```diff
@@ -8,6 +8,232 @@ capture significant language changes that have the possibility of breaking
...
+
+    // instance will now be automatically deleted at the end of this block
+  }
```
benharsh: I think a mention of the "Class New" section of the spec (if that's the right section) would be useful. Up to you.

PR #11098, opened 2018-09-11
Fix some typos in user-facing documentation

ronawho:
Fix some typos in the spec and linear algebra documentation that chplspell found

### DIFF
--- a/modules/internal/ChapelIteratorSupport.chpl
+++ b/modules/internal/ChapelIteratorSupport.chpl
@@ -98,7 +98,7 @@ module ChapelIteratorSupport {
   // The two chpl_buildStandInRTT(type) functions accept, at run time,
   // the _RuntimeTypeInfo for domType/arrType that is **uninitialized**.
   // They returns a fresh domain/array type of the same kind, whose
-  // RTT is **initiaized**. Important: no accessing the uninitialized RTTs.
+  // RTT is **initialized**. Important: no accessing the uninitialized RTTs.
   //
   // It took some acrobatics to get the domain's distribution type,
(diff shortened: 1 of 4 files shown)


cassella:
I tend to play it by ear.

Anything that seems likely to show up in more than a small handful of files, I'll add to the natural laguage dictionary.  Even if it's not "English".  E.g., unconverged or superdiagonal in this batch, or project-specific "words" like uniquifying or atomicness.  Also, people's names, C functions (printf, setaffinity, etc.), assorted command line stuff (libc, lnuma), common variable names (lidx, ridx, bufsize), etc.

Things that are likely only ever going to be in one or two files I'll put in the file-specific dicts.  E.g., part of a git commit hash that's referenced in a comment (eafdfcd) or a very specific variable name (bxor, pivsign).

I've pretty much stopped using the language-specific dictionaries because the same terms that appear in .chpl files appear in the .rst and .tex files describing them.  Though it probably still makes sense to keep latex keywords in their own dictionary.

If I had to start over, I'd probably use the file-specific dictionaries not for things that will probably only be in one or two files, but only for things that shouldn't be in the natural language dictionary since they're plausible typos.  E.g., I see "courant" and "uration" in file-specific dictionaries now.  (for `test/release/examples/benchmarks/lulesh/README`, if you're curious)

I don't think there's really any reason to keep things like "calcforce" or "howpublished" out of the natural language dict, even if they're only used in one place.

cassella:
@ronawho I take it from your question that you're preparing a PR to update the dictionary?

ronawho:
> I take it from your question that you're preparing a PR to update the dictionary?

Nope, I had completely forgotten about this already :)

PR #11099, opened 2018-09-11
Make docs version button work better at release time

bradcray:
The current version of the version button logic for the docs looks
at whether or not the version string in doc/rst/conf.py contains
the text (pre-release) or not, which makes things a bit dicey
right around release time, like now, when we start pulling that
string from the version text.

This change uses a version variable with the version button logic
itself to keep track of which version is the next release and
adds the (pre-release) text there so that it will always be
applied.  This should also reduce the maintenance burden within
doc/rst/conf.py slightly.

### DIFF
--- a/doc/util/versionButton.php
+++ b/doc/util/versionButton.php
@@ -88,13 +88,15 @@ if (pagePath == "") {
 }
 function dropSetup() {
   var currentRelease = "1.17";
+  var nextRelease = "1.18";
   var button = document.getElementById("versionButton");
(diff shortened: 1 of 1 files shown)


bradcray:
@awallace-cray : I think that this should help.  We'll need both of our PRs to see, though.

awallace-cray:
LGTM

bradcray:
Great, merged!  This'll probably make the button look a bit suboptimal until your version number update is merged.

PR #11103, opened 2018-09-12
Fix confusion on my part about IIIT-Hyderabad vs. IIT-Hyderabad

bradcray:
I was surprised that two schools would have such similar acronyms and didn't spend as much time Googling as I should have.


bradcray:
@ysahil97 : See above.

ysahil97:
LGTM :+1:

PR #11105, opened 2018-09-12
Fix some entries that were mis-sorted into standard rather than packages

bradcray:


bradcray:
@ben-albrecht : Looks like there were some other cases as well (who writes this stuff?!?).  Would you double-check this?

CHANGES.md
```diff
@@ -164,6 +161,9 @@ Package Modules
 * replaced string options with enumerations in the `Crypto` module
```
ben-albrecht: nitpick: LinearAlgebra is inconsistently not back-tic'd above. Should be:

```
* made `LinearAlgebra` sparse matrices store indices in unsorted order
```

PR #11108, opened 2018-09-12
Convert builtins/internal URLs to just builtins/

bradcray:
Recent builds of the docs have used builtins/Foo.chpl rather than builtins/internal/Foo.chpl due to being built on Mac OS X and [differences between Linux and BSD](https://jondavidjohn.com/linux-vs-osx-the-cp-command/).  We didn't catch this asymmetry earlier, so had URLs of the latter form in this file where I think we want the former.  Independently of this PR, we're looking into updating the script that builds the online master docs to squash out that internal/ directory.

While here, I also caught a reference to docs/latest/ that needed to be updated to docs/1.18/

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -195,7 +195,7 @@ New Tools / Tool Changes
 Interoperability Improvements
 -----------------------------
 * significantly improved support for calling exported Chapel routines from C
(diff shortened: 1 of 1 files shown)


bradcray:
@ben-albrecht, @ronawho, @awallace-cray:  N.B.

PR #11109, opened 2018-09-12
Remove trailing slashes for consistent BSD/GNU cp

ben-albrecht:
GNU cp does not recognize trailing slashes, so this PR removes them to avoid inconsistent behavior across platforms.

Prior to this PR, we were getting an inconsistency in the docs path:

```
# BSD (good)
chapel/doc/html/builtins/*.chpl

# GNU (bad)
chapel/doc/html/builtins/internal/*.chpl
```

This resulted in an inconsistency between our docs built nightly (built on linux) vs. release docs (built on OS X).


ben-albrecht:
Special thanks to @jondavidjohn for warning other developers about this [back in 2012](https://jondavidjohn.com/linux-vs-osx-the-cp-command/).

PR #11111, opened 2018-09-12
Pre-release prep. Chapel 1.18.0: bump version numbers on master from 1.18 to 1.19

awallace-cray:
NOTE: This is not Chapel release 1.18.0 or 1.19.0. This change prepares the source on
master branch for next development cycle AFTER the upcoming Chapel 1.18.0 release,
currently scheduled for 20 Sep 2018.

The release/1.18 branch has been created, now set up for the next development
cycle.
* Revert "post-release version markings on master"
  (ie, put the "pre-release" strings back in)
* Bump version numbers on master from 1.18 to 1.19

PR #11112, opened 2018-09-12
Improve robustness of doc version menu during release times

bradcray:
During the period when we're branching the release and setting up for the next
one, the version button can get a bit wonky because it needs to track not just
the current version, but potentially two future versions (the one we're about to
release and the one that comes after that).  This updates the logic to support
tagging two releases as pre-releases.

While here, I also made the "pre-release" and "old release" labels use non-breaking
hyphens and spaces (when the pre-release logic used to be in the doc building
logic itself, it did this, but now that we've moved it here, it wasn't).  It looks to me
like this rarely comes into play, but it makes things look nicer when it does.

### DIFF
--- a/doc/util/versionButton.php
+++ b/doc/util/versionButton.php
@@ -89,13 +89,14 @@ if (pagePath == "") {
 function dropSetup() {
   var currentRelease = "1.17";
   var nextRelease = "1.18";
+  var nextNextRelease = "1.19";
   var button = document.getElementById("versionButton");
   // Uses unicode down-pointing triangle
   var arrow = " &#9660;";
   button.innerHTML = "version " + chplTitle + arrow;
 
   // Choose button color
-  if (chplTitle == nextRelease) {
+  if (chplTitle == nextRelease || chplTitle == nextNextRelease) {
     button.innerHTML = "version " + chplTitle + " (pre-release)" + arrow;
     button.classList.add("preRelease");
   } else if (chplTitle == currentRelease) {
--- a/doc/util/versionButton.php
+++ b/doc/util/versionButton.php
@@ -97,12 +97,14 @@ function dropSetup() {
 
   // Choose button color
   if (chplTitle == nextRelease || chplTitle == nextNextRelease) {
-    button.innerHTML = "version " + chplTitle + " (pre-release)" + arrow;
+    # add pre-release label using a non-breaking hyphen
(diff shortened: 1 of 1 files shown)


bradcray:
@awallace-cray : I think this will help the version button on your "bump to 1.19" branch.

doc/util/versionButton.php
```diff
@@ -89,13 +89,14 @@ function toggleDropDown() {
...
+  var nextNextRelease = "1.19";
   var button = document.getElementById("versionButton");
   // Uses unicode down-pointing triangle
   var arrow = " &#9660;";
   button.innerHTML = "version " + chplTitle + arrow;
 
   // Choose button color
-  if (chplTitle == nextRelease) {
+  if (chplTitle == nextRelease || chplTitle == nextNextRelease) {
     button.innerHTML = "version " + chplTitle + " (pre-release)" + arrow;
```
awallace-cray: @bradcray @ben-albrecht 
I remembered this little detail- we are planning to take out that "pre-release" version string from line 64 of conf.py (https://github.com/chapel-lang/chapel/pull/11111/commits/fdfa83bf8fdb0afe0a15924dde30cb077ef3e470), because Brad effectively moved it to versionButton.php (above). 

But look at line 65 in conf.py, right below the former "pre-release" string (https://github.com/chapel-lang/chapel/blob/fdfa83bf8fdb0afe0a15924dde30cb077ef3e470/doc/rst/conf.py#L65)

IIRC, we put in that "non-break hyphen" to make sure the versionButton had no line break- or if it did, it happened where we wanted. 

I wonder if versionButton.php needs something like that?

awallace-cray: The 1.19-preview docs that I built look OK (see `~chapelu/www/1.19.0-pre/master/`, Cray internal).  Button says pre-release w/o a linebreak. At least in my browser. 

BTW, I assume we will bump the "currentRelease"/"nextRelease" numbers (line 90, versionButton.php) on Sept 20.

bradcray: Did this.  @awallace-cray, I'm planning on merging this today because it shouldn't break anything (and if it does, better to know now than later) and will set us up for a smooth transition to 1.19 tomorrow.

bradcray:
@benharsh: I think this is fairly straightforward/trivial, but as the script's original author, would you mind giving this a quick review?

Issue #8968, opened 2018-03-26
Investigate stream promoted perf loss from iterator memory management rework 


ronawho:
There was a non-trivial performance regression for stream-promoted under gn-mpi as a result reworking iterator memory management (https://github.com/chapel-lang/chapel/pull/8073) 

We missed this during our normal performance triage (we don't triage gn-mpi regularly), though it is visible:  https://chapel-lang.org/perf/16-node-xc/?startdate=2018/01/06&enddate=2018/01/25&configs=gnuugniqthreads,gnugasnetmpi&graphs=hpccpromotedstreamperfgbsn5723827200

Since stream is such a simple yet fundamental/core benchmark, it would be good to understand the cause of the performance loss and address it. 

It is worth noting that comm=ugni wasn't impacted, which is curious and may be helpful in tracking down the root cause of the regression (could be extra comm and ugni is fast enough that it doesn't show, or extra allocations and with the registered heap ugni+jemalloc won't return memory to the system so repeated allocations can be faster). It's also worth noting that stream-global wasn't impacted, so this is likely a code-path specific to how we build promotion wrappers.

Note that we discovered these after the fact in the release-over-release timings: 

![stream-prom](https://user-images.githubusercontent.com/1588337/37936575-ef35bd48-310a-11e8-8fea-f3d984f501c2.png)

ronawho:
TODO: look at total memory used and comm counts before/after that PR

ronawho:
I didn't see any obviously bad differences in the generated code before and after this change, and I didn't see any differences in comm-counts or any substantial changes in memory allocations.

If we look into this again, I think we should just incrementally revert changes in this PR to try and identify the root cause. I don't understand the PR all that well, so that's probably something best left to @mppf, but we can figure that out if/when we look at this again.

cassella:
It looks like there have been a few more (unannotated) regressions since the one in January 2018:
https://chapel-lang.org/perf/16-node-xc/?startdate=2018/01/01&enddate=2019/03/21&configs=gnuugniqthreads,gnugasnetmpi&graphs=hpccpromotedstreamperfgbsn5723827200

ronawho:
Good timing -- new release-over-release graphs agree with you:

![stream prom](https://user-images.githubusercontent.com/1588337/54854676-eb785e00-4cb0-11e9-803c-6f8b3261a123.png)

ronawho:
[...]

PR #11113, opened 2018-09-13
chpl_getPrivatizedCopy implemented in Chapel

### LINKED ISSUE #8060: Performance with --llvm lags for prk-stencil with --no-local
For prk-stencil, llvm performs on par with the gnu C backend with `--local`, but performance falls off for `--no-local`, which is what causes 16-node-xc prk-stencil to lag behind with `--llvm`

Here's a small reproducer that is just a stripped down version of prk-stencil (it can probably be stripped down further, but this is a good start):

```chpl
use Time, StencilDist;

config const order: int = 20000;
config param R = 2;

param Wsize = 2*R + 1,
              R1 = R + 1;

proc main() {
  const localDom = {0.. # order, 0.. # order},
   innerLocalDom = localDom.expand(-R);

  const outDist   = new dmap(new Stencil(innerLocalDom));
  const fluffDist = new dmap(new Stencil(innerLocalDom, fluff=(R,R)));

  const Dom = localDom dmapped fluffDist,
   innerDom = innerLocalDom dmapped fluffDist;

  const outputDom = localDom dmapped outDist;

  var input: [Dom] real = 0.0,
      output: [outputDom] real = 0.0;

  var weight: Wsize*(Wsize*(real));

  var timer: Timer; timer.start();
  //forall (i,j) in innerDom with (const in weight) {
  for (i,j) in innerDom {
    var tmpout: real = 0.0;
    for param jj in -R..-1 do tmpout += weight[R1][R1+jj] * input.localAccess[i, j+jj];
    for param jj in 1..R   do tmpout += weight[R1][R1+jj] * input.localAccess[i, j+jj];
    for param ii in -R..-1 do tmpout += weight[R1+ii][R1] * input.localAccess[i+ii, j];
    for param ii in 1..R   do tmpout += weight[R1+ii][R1] * input.localAccess[i+ii, j];
    output.localAccess[i, j] += tmpout;
  }
  writeln(timer.elapsed());
}
```

The generated code is actually pretty similar for `--local` and `--no-local`. The main difference is that the `--no-local` code has an extra dereference to get the `addr` out of the widened input array. 

Here's some performance numbers:

| config                                                         | time  |
| -------------------------------------------------------------- | ----- |
| chpl f.chpl --fast --no-local                                  | 2.05s |
| chpl f.chpl --fast --no-local --target-compiler=clang-included | 5.15s |
| chpl f.chpl --fast --no-local --llvm                           | 5.15s |
| chpl f.chpl --fast --no-local  --ccflags -fno-strict-aliasing  | 6.20s |

The interesting things to note are that clang-included and clang perform as poorly as `--llvm` and that gcc performs poorly with `-fno-strict-aliasing`. I'm hoping that @dmk42's stuct alias work will help here, but we'll have to see when that comes online.

Looking at the generated code you can see that our LICM is ineffective. Reverting 538386e6f7 allows all the meta-data to be hoisted, and clang-included/llvm outperform the C backend. That doesn't help the full stencil-prk, but I would guess there end up being atomics from the RadCache the prevent hoisting or because of the getPrivatizedCopy calls, which we could almost certainly workaround.

mppf:
* Adjusts chpl_getPrivatizedCopy to generate the loads from
  chpl_privateObjects in Chapel code. This allows the Chapel TBAA
  information to be applied and helps some alias analysis problems.
* To make it more understandable, chpl_privateObjects is now an array of
  chpl_privateObject_t (which is just a struct storing `void*`). This
  might also make it more likely to have a different type that other
  loads/stores in a loop (but that depends on the exact pattern of
  loading).
* Remove PRIM_GET_PRIV_CLASS since it is no longer used.

With this PR, the example based on prk-stencil from issue #8060 is
competitive with the C backend with --no-local.

Resolves #8060.

- [x] passed full gasnet testing
- [x] passed 1000 of trials on test/distributions/privatization/ with C backend
- [x] passed 1000 of trials on test/distributions/privatization/ with LLVM backend
- [x] passed full --llvm testing


ronawho:
Nice!

I probably won't get a chance to review until Friday or Monday, but at a high-level -- I have a really hard time reviewing and doing post-mortem performance analysis on PRs like this. To me this is logically 3 only somewhat related PRs -- llvm compiler file moves, LICM cleanup, and the actual improvements to privatization. It'd be really helpful to me if you could split this up. And for a concrete example where this has bitten me in the past, I gave up on https://github.com/chapel-lang/chapel/issues/8968 really easily since I had a hard time narrowing down on the source of performance regressions in such a wide-sweeping change.

mppf:
@ronawho - it is done.

ronawho:
Could you explain more about why `reprivatizeIterators()` isn't needed anymore? I don't understand what this does or why it's no longer necessary. 

Possibly help info -- from the compilerOverview doc:

"""
Reprivatize privatized-object fields (array and domain descriptors may be privatized) in Iterator classes That is, code is inserted when accessing such fields to get the local private copy. This is not necessary with arbitrary classes and records because their fields will never capture privatized objects.  Instead, they will always capture priva- tized IDs which can then be changed into objects at the point they are used. The IDs map to the privatized copy of a class on any given locale.  Iterator classes, on the other hand, are built up by the compiler. If the privatized object (as opposed to the ID) is added to the class, then we may be pointing at a remote object when there is a local object that we could be pointing at.
"""

And the following commits may have more info: 
 - https://github.com/chapel-lang/chapel/commit/117bdaa5b738e57eef41e9de984e4c10a898f608
 - https://github.com/chapel-lang/chapel/commit/6ea8a445a4fc9d33cad0024fe93a0ac087de860b

runtime/include/chpl-privatization.h
```diff
@@ -27,12 +27,19 @@ void chpl_privatization_init(void);
 
 void chpl_newPrivatizedClass(void*, int64_t);
 
+typedef struct chpl_privateObject_s {
+  void* obj;
+} chpl_privateObject_t;
+
 // Implementation is here for performance: getPrivatizedClass can be called
 // frequently, so putting it in a header allows the backend to fully optimize.
```
ronawho: Comment is semi out of date at least the name `getPrivatizedClass`

ronawho:
Can you run a bunch of trials for `test/distributions/privatization` for both backends, and maybe add a test that adds to the privatization array while also reading it to ensure we're not hoisting when we shouldn't be?

compiler/AST/wellknown.cpp
```diff
@@ -35,6 +35,7 @@ AggregateType* dtLocaleID;
 AggregateType* dtMainArgument;
 AggregateType* dtOnBundleRecord;
 AggregateType* dtOwned;
+AggregateType* dtPrivateObject;
```
ronawho: Is this used anywhere?

compiler/AST/primitive.cpp
```diff
@@ -875,6 +874,10 @@ initPrimitive() {
 
   prim_def(PRIM_NEEDS_AUTO_DESTROY, "needs auto destroy", returnInfoBool, false, false);
   prim_def(PRIM_AUTO_DESTROY_RUNTIME_TYPE, "auto destroy runtime type", returnInfoVoid, false, false);
+
+  // Corresponds to LLVM's invariant start
+  // takes in a pointer/reference argument that is the invariant thing
+  prim_def(PRIM_INVARIANT_START, "invariant start", returnInfoVoid, false, false);
```
ronawho: I must be missing something obvious, but where are you marking the privatized load with this primitive?

mppf:
@ronawho:

> Could you explain more about why reprivatizeIterators() isn't needed anymore? I don't understand what this does or why it's no longer necessary.

I added INT_FATAL asserts in the code generation for PRIM_GET_PRIV_CLASS and that's what most but not all of the code in reprivatizeIterators() is adding. These never fired on a full GASNet testing. Additionally I tried testing with an INT_FATAL in reprivatizeIterators() when it identifies something to privatize. Certainly we can do more testing but testing seems to indicate that the code is currently unused. The test case mentioned in those commits, streamCommCheck, still passes without reprivatizeIterators().

mppf:
> And if reprivatizeIterators() isn't needed, is it because of this PR or did you just notice it because of this PR. If it's unrelated to this PR, could we split that out too?

I'm looking at that. I'll try to at least explain the testing steps better there.

compiler/AST/primitive.cpp
mppf: I'm not. This PR includes the primitive as something we can experiment with. Initially I was using it but then I figured out that the existing TBAA rules covered it (at least in the case we're working on here) and the LLVM LICM code doesn't work with llvm.invariant.start on a global variable.

However I'm expecting that this primitive will be useful for other LLVM LICM performance questions - I'm hoping to use it to try hoisting array headers, for example. But it's also not strictly necessary in this PR (or on master at all). (I tend to prefer to put such things on master so they are not lost / redone).

compiler/AST/wellknown.cpp
mppf: No, it once was. I'll remove it.

compiler/AST/checkAST.cpp
```diff
@@ -295,7 +295,6 @@ void checkPrimitives()
      case PRIM_RT_ERROR:
      case PRIM_RT_WARNING:
      case PRIM_NEW_PRIV_CLASS:
```
ronawho: You don't have to do this, but I wonder if we can get rid of `PRIM_NEW_PRIV_CLASS` too and just do extern calls now.

mppf: I thought about doing that but didn't because we might later want compiler analysis to find functions calling this primitive. (Even so, we could look for calls by name...)

ronawho: True, but I think that argument holds for the privatized getter too, and that's now an extern call.

mppf: actually it's not even an extern call.
Anyway we'll leave the issue for another PR.

PR #11122, opened 2018-09-13
Update memcpy intrinsic generation for LLVM 7

mppf:
Allows Chapel to build with LLVM master. The memcpy
intrinsic no longer takes in an alignment argument.

Future work: run llvm-global-to-wide tests nightly (issue #11192).
Passed full llvm testing, gasnet testing.

Reviewed by @dmk42 - thanks!

### DIFF
--- a/compiler/codegen/expr.cpp
+++ b/compiler/codegen/expr.cpp
@@ -2797,6 +2797,21 @@ void codegenCallMemcpy(GenRet dest, GenRet src, GenRet size,
 
     llvm::Function *func = llvm::Intrinsic::getDeclaration(info->module, llvm::Intrinsic::memcpy, types);
     //llvm::FunctionType *fnType = func->getFunctionType();
+
+#if HAVE_LLVM_VER >= 70
+    // LLVM 7 and later: memcpy has no alignment argument
+    llvm::Value* llArgs[4];
+
+    llArgs[0] = convertValueToType(dest.val, types[0], false);
+    llArgs[1] = convertValueToType(src.val, types[1], false);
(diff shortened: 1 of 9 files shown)


dmk42:
Do the files `compiler/util/llvm-global-to-wide/test/*.ll` need to be updated as well?  Some of them use the memcpy intrinsic.

mppf:
@dmk42 - yeah, good call. I'll look at updating those and also getting them to run in the test suite.

compiler/codegen/expr.cpp
```diff
@@ -2797,6 +2797,21 @@ void codegenCallMemcpy(GenRet dest, GenRet src, GenRet size,
...
+
+#if HAVE_LLVM_VER >= 70
+    // LLVM 7 and later: memcpy has no alignment argument
+    llvm::Value* llArgs[4];
+
+    llArgs[0] = convertValueToType(dest.val, types[0], false);
+    llArgs[1] = convertValueToType(src.val, types[1], false);
+    llArgs[2] = convertValueToType(size.val, types[2], false);
+
+    // LLVM memcpy intrinsic has additional arguments alignment, isvolatile
```
vchuravy: This comment is out-of-date

mppf: true, thanks

mppf:
@dmk42 - I can't both update compiler/util/llvm-global-to-wide/test/*.ll and run it in the test suite since the change happens only in LLVM 7 and our test suite runs with LLVM 6. Beyond that, those tests aren't working right now. I'd like to fix that in a separate effort from this PR.

So, is this PR good to go? Thanks.

PR #11127, opened 2018-09-13
Update script to work with new git-based site

### LINKED ISSUE #10337: Update test/studies/shootout/submitted/clbg-diff.py to use GitLab site
Now that CLBG has moved to GitLab-based hosting, we need to update this script (used to make sure our local submitted versions of the benchmark match those on the live site) to keep it working.

bradcray:
This updates the script that we use to ensure that our submitted/
directory's versions of the shootout benchmarks matches the
official website's versions.

While here, I also added a check for any tests that are .futurized or
.notested, since it's likely to indicate that the submitted versions
don't work with the current compiler.

Resolves #10337.

### DIFF
--- a/test/studies/shootout/submitted/clbg-diff.py
+++ b/test/studies/shootout/submitted/clbg-diff.py
@@ -24,9 +24,7 @@ def main():
     print('Updating CLBG repository')
     print('------------------------')
     print()
-    CLBG_repo = ':pserver:anonymous@cvs.debian.org:/cvs/benchmarksgame'
-    CLBG_dir = 'benchmarksgame/bench/'
-    cvs_checkout(CLBG_repo, CLBG_dir)
+    git_checkout()
 
     print()
     print('---------------')
@@ -37,7 +35,7 @@ def main():
     # List of Chapel-implementations
     submitted = glob.glob('*.chpl')
     # List of CLBG-implementations
-    clbg = glob.glob('benchmarksgame/bench/*/*.chapel')
+    clbg = glob.glob('benchmarksgame/*/*.chapel')
     # List of CLBG-implementations without path
     clbgbase = [os.path.basename(c) for c in clbg]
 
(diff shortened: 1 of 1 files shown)


bradcray:
Attn: @ben-albrecht and @ronawho

test/studies/shootout/submitted/clbg-diff.py
```diff
@@ -70,17 +68,37 @@ def main():
...
+def git_checkout():
+    """Grab git source tarball"""
+    url = 'https://salsa.debian.org/benchmarksgame-team/benchmarksgame/raw/master/public/download/'
+    file = 'benchmarksgame-sourcecode.zip'
+    destdir = 'benchmarksgame'
+    stdout, stderr = run_command('rm -f ' + file, True)
+    stdout, stderr = run_command('rm -r ' + destdir, True)
+    stdout, stderr = run_command('wget '+ url + file, True)
+    stdout, stderr = run_command('unzip -d'+ destdir + ' ' + file, True)
+    stdout, stderr = run_command('rm ' + file, True)
```
ben-albrecht: Same thing for `-f` here.

bradcray: In this case, the zipfile is guaranteed to exist if the previous commands were successful, though, right?

ben-albrecht: Yes, but if the `wget` fails, then this generates an additional error. Not a big deal though (especially for a minor developer script).

bradcray: Oh, I was under the impression that the script failed in the presence of errors and figured it would never get this far (at least, that's what I was seeing when I had other commands failing yesterday...  but I'll admit I don't know the first thing about Python failure modes or behavior).

test/studies/shootout/submitted/clbg-diff.py
```diff
@@ -70,17 +68,37 @@ def main():
...
-    """Checkout cvs repository"""
-    cmd = 'cvs -d {0} checkout {1}'.format(r, d)
-    stdout, stderr = run_command(cmd, True)
+def git_checkout():
+    """Grab git source tarball"""
+    url = 'https://salsa.debian.org/benchmarksgame-team/benchmarksgame/raw/master/public/download/'
+    file = 'benchmarksgame-sourcecode.zip'
+    destdir = 'benchmarksgame'
+    stdout, stderr = run_command('rm -f ' + file, True)
+    stdout, stderr = run_command('rm -r ' + destdir, True)
```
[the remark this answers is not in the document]
ronawho: Probably not worth making this change, but alternatively you could use `os.remove()` for files and `shutil.rmtree()` for dirs. There's also a `zipfile` in the standard library and I'm sure there's a way to do a wget or someting equivalent.

PR #11128, opened 2018-09-13
Switch study version of binary trees from owned to unmanaged

bradcray:
This changes my study version of binary trees from owned to unmanaged,
but frees in a distinct step at the end rather than freeing while
we're computing a tree's sum.  By comparing to the previous and
release versions, this should help us distinguish the overhead for
using 'owned' from the overhead for deleting in a separate step
(requiring a re-traversal of the tree).

### DIFF
--- a/test/studies/shootout/binary-trees/binarytrees-blc.chpl
+++ b/test/studies/shootout/binary-trees/binarytrees-blc.chpl
@@ -64,15 +64,15 @@ proc main() {
 // A simple balanced tree node class
 //
 class Tree {
-  var left, right: owned Tree;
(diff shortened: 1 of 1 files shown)


bradcray:
Attn: @ronawho

bradcray:
Surprisingly, this change seems to have closed most of the gap with the free-while-summing unmanaged version, making it seem that `owned` does have more overhead for this memory-intensive benchmark than I had expected.  As a follow-up, it would be worth digging into whether the space required for the record-wrapped type is increasing the space requirements for this benchmark as a whole.

PR #11129, opened 2018-09-13
Upgrade to qthreads 1.14

ronawho:
Upgrade to qthreads 1.14

qthreads 1.13 included some profiling work. qthreads 1.14:
 - removed ROSE and RCRtool support (never used by us, but this gets rid of a
   lot of code in many qthreads files)
 - includes most of the patches we had against 1.12 (with the exception of sync
   var serialization fixes)
    - updated the allocator to use our public API -- 0c3b9de163 / #9736
    - added a hwloc_via_chapel layer -- #9824
    - fixed CHPL_HWLOC=none builds -- 41d9a283548 / #10212
    - added way to get sizeof qthread_runtime_data_s -- 388a6e38435 / #10926
    - fixed a bug in mem pool limitiing code -- de54d106d1 / #10926
 - added a "network tasks" feature flag for tasks (not obvious to me if this
   could eventually be useful for us for hosting our progress engine or
   something)

I don't expect there to be any functional or performance differences from this.

For the full changelogs see:
 - https://github.com/Qthreads/qthreads/releases/1.13
 - https://github.com/Qthreads/qthreads/releases/1.14


ronawho:
@gbtitus mostly fyi, but would appreciate if you could sanity check the TODO list and see if there's anything else there that you can think of

gbtitus:
Maybe add:
 - [ ] Sanity check Qthreads uses Chapel runtime hwloc instance

The easiest way might be to use the linker `--wrap` option to interpose a wrapper for `hwloc_topology_init()` and then make sure that wrapper only got called once.  I should have added such a test when I did [PR 9824](https://github.com/chapel-lang/chapel/pull/9824).  I could add that now ...

For 'OOM from task stacks', would unbounded recursion do it?  Don't we already have a test that does that, or can be configured for recursion depth sufficient to cause it?

ronawho:
> For 'OOM from task stacks', would unbounded recursion do it? Don't we already have a test that does that, or can be configured for recursion depth sufficient to cause it?

I just tested this with:

```chpl
use Time;
coforall 1..10000 do sleep(2);
```

and CHPL_RT_MAX_HEAP_SIZE=2G under ugni with -nl2.

>The easiest way might be to use the linker --wrap option to interpose a wrapper for hwloc_topology_init() and then make sure that wrapper only got called once. I should have added such a test when I did PR 9824. I could add that now

I'm using startup time as a proxy for how many times hwloc_topology_init() is called. I don't think the linker thing is needed, but am happy to verify if you want.

ronawho:
Ran full std testing as well as:
 - [x] Verify that `CHPL_HWLOC=none` works on linux and mac
 - [x] Sanity check stream perf 
 - [x] Sanity check startup time
 - [x] Sanity check OOM from task stacks
 - [x] Sanity check stack size w/ and w/o hugepages
 - [x] Sanity check pool limiting code
 - [x] Sanity check SPMD stream for sync var serialization

gbtitus:
Your 'OOM from task stacks' test seems fine to me.

For counting the number of times `hwloc_topology_init()` is called, startup time as a proxy is okay though it's definitely indirect.  At this point nothing bad would happen if the runtime and Qthreads happened to end up using different topologies somehow.  If and when we do get there it might be nice to have a more direct test that they're sharing as desired, though.

gbtitus:
I took the liberty of making myself a reviewer on this just so I could approve it.   👍

PR #11130, opened 2018-09-13
Update fastas for 1.18 and do some other cleanups

bradcray:
This makes a few changes to our submitted fasta versions for the
purposes of making them compatible with 1.18 and doing some other
cleanups:

* started relying on param int->int(8) coercions
* stopped relying on enum->int coercions
* took advantage of shape preservation fix to retain 0-based indices
  - would be even nicer if we supported shape preservation for ranges
* restored preferred whitespace (removed in last submission)

### DIFF
--- a/test/studies/shootout/submitted/fasta.chpl
+++ b/test/studies/shootout/submitted/fasta.chpl
@@ -24,7 +24,7 @@ use nucleotide;
 //
 // Sequence to be repeated
 //
-const ALU: [0..286] int(8) = [
+const ALU: [0..286] nucleotide = [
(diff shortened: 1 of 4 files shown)


bradcray:
Attn: @benharsh and @ronawho
@vasslitvinov : Noting this as a case where shape preservation for ranges would've avoided the need for using a domain.

bradcray:
No major changes here as expected/hoped:  https://chapel-lang.org/perf/chapcs/?startdate=2018/08/31&enddate=2018/09/19&graphs=submittedfastashootoutbenchmark

PR #4821, opened 2016-11-04
RVF arrays in records

benharsh:
PR #4762 allows us to always RVF record-wrapped symbols. We can also do the same for record-wrapped fields in records that are themselves RVF'd. This resolves some comm-counts that came from arrays packed into a chpl__iterLF tuple.

PR #4814, opened 2016-11-02
Improve RVF by inferring const-refs

benharsh:
Addresses comm-count increases from PR #4745. By inferring references as const-references, RVF knows it can forward them in the absence of synchronization and locality-queries.

The file "inferConstRefs.cpp" was added to store the functions that infer constness in the AST. This analysis can be started by calling "inferConstRefs();", and operates on the entire tree. To infer constness, we look at a ref's uses in the tree. There are a number of cases that attempt to identify modifications to the reference directly, or to follow aliases of that reference and infer constness of those aliases. This analysis can certainly be improved, but this is satisfactory for now. I have attempted to write the analysis in a conservative manner that will throw up its hands in the face of unhandled primitives and AST patterns.

We perform a similar process when determining whether or not the on-statement needs to know about the ref's ".locale" property. Again, we try to follow aliases in simple cases, but for now avoid handling references in aggregate types. This process can be improved in the future, and is intended to be a conservative implementation.

With this change, RVF will no longer attempt to handle refs that are not const-refs. This allows me to remove the "isSafeToDerefArg" function.

As a result of this work, a number of .good files showed reduced comm-counts. Some of the increases from #4745 were resolved, but others remain at large.

PR #4762, opened 2016-10-21
Improve Array Memory Management

mppf:
## Overview

This PR re-implements memory management for arrays, domains, and distributions. The changes mainly fall into these groups:
 * array-related compiler changes:
 * tuple changes
 * other enabling changes in the compiler
 * changes to module code for arrays/domains/distributions
 * other module code changes
 * test changes

## User-facing behavior changes

This PR causes mainly four behavior changes:

 * Arrays are returned by value
 * Arrays, domains, and distributions are no longer kept alive (preventing deletion) by a task using them
 * Arrays are no longer kept alive by slices
 * Tuple behavior is refined, especially for blank-intent tuple arguments

The first of these is a new decision for the language in order to make it more implementable and to make arrays more consistent with records and other types. The second and third are implied by the Lexical Scoping section in the Chapel Evolution document (http://chapel.cray.com/docs/latest/language/evolution.html#lexical-scoping). The last addresses a known design issue in the language and is discussed more below.

Note that while arrays return by value by default, the ref-return-intent or const-ref-return-intent can override that behavior.

In order to make these changes more achievable, it is not currently defined what happens when returning a slice from a function. Currently, such slices return without making a copy - and so it is possible to create core dumps by returning a slice of a local array.

The tuple behavior changes adjust the implementation so that a 1-element tuple behaves similarly to just the element (as far as whether or not a copy occurs). Then the behavior logically extends to tuples with more elements. In particular, creating a tuple containing records and then passing it to a function does not copy the record. But returning such a tuple formal argument or using it to initialize a variable will copy the record. This design has similar goals and ideas to the design presented in CHIP 6 but differs slightly in the details.

## Array-Related Compiler Changes

- removing cases when an array wrapper record is shallow-copied in the compiler. Either the copy constructor should be called (which does a deep copy) or the variable in question should be a reference to the array (which does no copy and does not impact array lifetime). Now autoCopy and initCopy do the same thing for arrays, domains, and distributions. E.g. returning an _array causes the array to be returned by value, but returning a _ref(_array) causes it to be returned by reference.
- adjusts the compiler to treat arrays, domains, and distributions similarly to user records and strings.
- adds a workaround to make `var x => GlobalArray[1..10]` work
- disables most of removeUnnecessaryAutoCopyCalls. Only autoCopy calls for POD types are removed. The rest of this pass should no longer be necessary since we now just do a better job at inserting the minimal number of autoCopies in the first place.
- disables removeWrapRecords
[...]

PR #11134, opened 2018-09-14
remove reprivatizeIterators

mppf:
This PR removes reprivatizeIterators because it is currently unused.
It has two different commits for two different steps:

1. Replace all transformation in reprivatizeIterators with INT_FATAL

- [x] full --no-local testing
- [x] full gasnet testing

2. Remove reprivatizeIterators completely as well as privatized class flag

- [x] full local testing
- [x] full gasnet testing

Performance for test/release/examples/benchmarks/hpcc/stream.chpl is not impacted:

| config                          | master       | remove-reprivatize-iterators |
| ------------------------------- | ------------ | ----------------- |
|  GNU, ugni, 16 node  XC            | 1472 GB/s | 1475 GB/s      |
| same, iterator inlining off`*`         | 169 GB/s   | 169 GB/s  |

`*` --no-inline-iterators --no-optimize-loop-iterators

Reviewed by @ronawho and @benharsh - thanks!

### History

reprivatizeIterators was added in these commits

 * 117bdaa
 * 6ea8a44

these mention the test streamCommCheck

that had the expected number of GETs updated in
 * 89577cc0047fd785885522668fb42957dc6f57eb (regression from qualified refs PR #4745)
 * eef986dbbadea94dfa9ac55e71eff00d384b24b8 (regression from array memory management PR #4762)
 * fdd134d31cdd30e4d04526ea795011557f2510eb (improvement from  Improve RVF by inferring const-refs #4814 )
 * c15ae7ff80492560a91bfd1c07218e0d150fddbc (improvement from  RVF arrays in records #4821 )


benharsh:
I think the reprivatizeIterators function was forgotten about and left in a dead state at some point. The function looks for FLAG_PRIVATIZED_CLASS on fields in an iterator class, but I only see that flag ever applied to a function or a function's formal. I think it's safe to remove.

mppf:
@benharsh - actually I think FLAG_PRIVATIZED_CLASS was getting propagated to type symbols here:

https://github.com/chapel-lang/chapel/blob/1056e73aaec83edf89b42aec95ed8d3c54818920/compiler/resolution/resolveFunction.cpp#L442-L445

However I agree it was probably left dead at some point.

mppf:
@ronawho - streamCommCheck still passes with `./util/start_test test/multilocale/deitz/needMultiLocales/streamCommCheck.chpl -compopts '--no-inline-iterators --no-optimize-loop-iterators'`

PR #11135, opened 2018-09-14
Solve internal error for param loop over improper range

lydia-duncan:
Param for loops were assuming that the lower bound, and upper bound of
a param loop would always be of type int, uint, or bool.  It assumed that the
stride of a param for loop would always be an integer.  Additionally, it assumed
that the bounds would always match, so that if the lower bound was one of the
proper types, the other one would also be.  This assumption was not verified
before trying to work with those types, so when these assumptions were
incorrect, the user would get an internal error instead of a helpful error
message.  This change locks the expected behavior in with error messages for
now.

While here, I noticed that the coercion rules for ranges were not taking
effect.  Using module functions to determine the high and low bounds
allowed our normal coercion rules to take effect without re-implementing their
support in the compiler.

This resolves cases A, H, and I of the internal error issue (#9987)

Adds four new tests to lock in the new error messages:
- the lower bound is of an incorrect type
- the upper bound is of an incorrect type
- the stride is of an incorrect type
- multiple are of an incorrect type


test/types/range/lydia/badRange-string-lower.good
```diff
@@ -0,0 +1,4 @@
+badRange-string-lower.chpl:1: In function 'main':
+badRange-string-lower.chpl:2: error: lower bound of param for loop must be an int, uint, or bool
+badRange-string-lower.chpl:2: error: lower bound and upper bound of param for loop expected to be of the same type
+badRange-string-lower.chpl:2: error: lower bound and stride of param for loop expected to be of the same type
```
bradcray: What would it take to make this less about the param for loop and more about the fact that the range is completely bogus? (i.e., you can't have a range from string to integer).

lydia-duncan: I could change the error message to refer to the range instead, or it would mean solving the problem additionally in a separate spot.  At the time the internal error was occurring, we were in a ParamForLoop AST node and it looks like it doesn't separate the range out into a separate construct until after the fact (if I'm reading the code correctly).  But we technically know it's a range.

lydia-duncan: Can we store param ranges right now?

bradcray: > Can we store param ranges right now?

You're right that we can't and that the current compiler architecture represents this more as a special loop form with explicit low and high bounds rather than a general loop over a param iterator.  But I don't think that's the perspective we want to present to the user.

Put another way: `""..2` would be a problematic range expression in any context, not just param for loops.  I think that flagging the error where you are makes sense given the current code organization, but I think we should complain about the range in a way that's similar to what we do (or would like to do) in other contexts rather than taking about param for loop bounds.  For example, this program:

```chapel
var r = ""..10;
```

generates the (so-so) error message:

```
testit.chpl:1: error: Bounds of 'low..high' must be integers of compatible types.
```

Could we simply print the same thing for these cases?  (or, if we wanted to improve it, maybe "Range bounds must be of compatible types."?) 

(I think the current message is so-so in that it doesn't use the "range" type by name and that ranges now also support enums and bools and may support other types in the future).

lydia-duncan:
Test revealed 6 failures (by running only the ones that failed prior to the last commit):

[Error matching compiler output for param/diten/paramForBool]
- The stride is the default stride (which is an integer).  It seems reasonable to allow stride to be of a different type than the high and low bound (but the code itself assumes they match/are coercible for int and uint bounds, which seems like something I could correct - I just was aiming for a minimal change if possible)

[Error matching compiler output for param/diten/paramForDiffTypes2]
- Uses an int for the low bounds and uint for the high bounds.  The code currently pretends the uint is an int in the case where the low bound is an int (and pretends the int is a uint if the uint is the low bound, which seems like a problem to me)

[Error matching compiler output for param/diten/paramForHuge]
- No explicit type, but looks like it ends up being uint for low bound and int for high bound.  See above

[Error matching compiler output for trivial/deitz/other/bmm]
- Upper bound is cast to a uint, lower bound is not, stride is default.

[Error matching compiler output for trivial/diten/fold_cond_stmt]
- Ditto bmm case

[Error matching compiler output for users/vass/verify-div-ceil-floor-mod-1]
- low and high bound uint, stride is default.

Sounds like I need to adjust the code to allow the stride to be a different type (fair).  Am I wrong to be concerned about the uint/int white lies in the implementation?

test/types/range/lydia/badRange-string-lower.good
lydia-duncan: >  But I don't think that's the perspective we want to present to the user.

That seems reasonable to me, yeah.

> Could we simply print the same thing for these cases? (or, if we wanted to improve it, maybe "Range bounds must be of compatible types."?)

So I should adjust the old error message to match this new one, and use it for my new messages, yes?  Not a problem, just making sure I'm understanding

lydia-duncan:
Of the tests that fail: 3 work as an ordinary range, 1 behaves weirdly (`int(32)..uint(32)` gives `range(int(64))` in paramForDiffType2), and 2 give a different type than the param version does (`range(uint(64))` instead of `range(int(64))` in bmm and fold_cond_stmt).

I can believe paramForDiffType2 being the intended behavior (but don't see how to adjust the param loop code to do it quickly so might argue for making this a future).  I think I can adjust to get the same behavior for the bmm and fold_cond_stmt tests as the ordinary range case

bradcray:
I was looking at these in parallel and think that they should all keep working.  

The main rule these tests rely on is that the bounds of a range should undergo normal coercion and param coercion rules:

* **param/diten/paramForDiffTypes2**: smaller int should coerce to bigger
* **param/diten/paramForHuge**: param int should coerce to uint
* **trivial/deitz/other/bmm**: param int should coerce to uint
* **trivial/diten/fold_cond_stmt**: param int should coerce to uint

The other rule that they rely on is that it's fine to use an integer stride for idxTypes of bool and uint (because you may want negative strides with either one; because no range is likely to require a stride > 2**63), which I think covers these two tests:

* **param/diten/paramForBool**
* **users/vass/verify-div-ceil-floor-mod-1**

One approach that might easily give you all these cases while also unifying the error messages would be to copy the bounds expressions of the param for loop into a normal build_bounded_range() expression as we do for ranges in other contexts, dropping the result on the floor.  Then if we had bad type combinations, that expression would flag the error (?).  This would avoid having to re-implement the various coercion rules in a special case.  The main downside I can think of for this approach is that if we didn't optimize the call away, we'd pay some overhead for building a range we didn't use (we could potentially mitigate this by putting the call within a conditional that never executed, but that still grows the AST size).

Or, you could create duplicates of the chpl_build_bounded_range() routine signatures called chpl_check_param_loop_bounds() that were inline and either empty or had compilerError()s in them.  Then, hopefully, whenever someone implemented a new chpl_build_bounded_range() overload, they could think about whether they should add a similar chpl_check_param_loop_bounds() overload as well.

bradcray:
One other potential approach would be to define Chapel routines that compute and return the param low/high bounds based on the given bounds:

```chpl
proc chpl_compute_low_param_loop_bound(param low: int(?t), param high: int(t)) param {
  return low;
}
proc chpl_compute_high_param_loop_bound(param low: int(?t), param high: int(t)) param {
  return high;
}
```

(and then the same for the uint and bool cases and then an error case).  This would give similar error checking and move all the coercion logic to convert from the given bounds to the param bounds that should be used in the module code.

lydia-duncan:
With that guidance, I was leaning towards:
- changing the `is_int_type(idxType)` check on line 400 to ensure both are ints, allowing the "either is a uint" case to fall into the else branch naturally
- reducing the error cases on if the upper and lower bounds match to only ensure their "is_bool_type" result matches
- removing the "stride same type as bounds" checks entirely (just reread the behavior in the uint/bool branch and it uses an int there already, so I misread that earlier)

It seems less hacky than other AST modifications could be, but I don't have a strong preference between that or defining module code functions (except maybe that this way is less code motion?)

bradcray:
I've only been looking at behavior, not implementation, so don't have a really strong opinion there.  If it were me, I'd be trying to avoid writing code that reproduced existing dispatch/coercion logic which is why I suggested using module code (it also seems much more principled than hacky to me).  At a high-level, I'm suggesting rewriting:

```chapel
for param i in lo .. hi do
```

into:

```chapel
param plo = chpl_compute_low_param_bound(lo, hi);
param phi = chpl_compute_high_param_bound(lo, hi);
for param i in plo..phi do ...
```

where I'm imagining the module routines will take care of the type checking, coercion, etc. ensuring that `plo` and `phi` come out the same type so that the compiler doesn't have to think about them much at all. 

(Really, I'd want to just write `param (plo, phi) = chpl_compute_param_bounds(lo, hi);` but I don't think we have tuples of params at present).

lydia-duncan:
@bradcray - fixed it and it passes testing.  Care to review when you get the chance?

compiler/AST/ParamForLoop.cpp
```diff
@@ -478,6 +479,26 @@ CallExpr* ParamForLoop::foldForResolve()
...
+void ParamForLoop::validateLoop(VarSymbol* lvar,
+                                VarSymbol* hvar,
+                                VarSymbol* svar) {
+  if (!lvar            || !hvar            || !svar)
+    USR_FATAL(this,
+              "param for loop must be defined over a bounded param range");
+
+  if (!lvar->immediate || !hvar->immediate || !svar->immediate)
+    USR_FATAL(this,
+              "param for loop must be defined over a bounded param range");
```
bradcray: Is this message ever triggered within the compiler?  I'd expect that it would always catch in the module code first?  (which seems preferable).  Oh, maybe only if the stride is not a param?

test/types/range/lydia/badRange-string-multi.good
```diff
@@ -0,0 +1,2 @@
+badRange-string-multi.chpl:1: In function 'main':
+badRange-string-multi.chpl:2: error: Range bounds must be of compatible types
```
bradcray: In retrospect, this error message seems a little too bad in this case since the types seem compatible (they're both strings) but just aren't types we can build ranges from.

bradcray:
I only had time to give it a quick glance today and it's looking as I'd expect/hoped.  If you can find someone else to give it a more careful review, that'd be good.

compiler/AST/ParamForLoop.cpp
lydia-duncan: You're right, this will probably only trigger for the stride (or if we accidentally remove the param return intent from those functions, or stop replacing things correctly in the compiler).

test/types/range/lydia/badRange-string-multi.good
lydia-duncan: I'll return to the other suggested message of "Bounds of 'low..high' must be integers of compatible types." (which matches the message in other places, iirc)

lydia-duncan: Err, maybe "Range bounds of 'low..high' ..."?

lydia-duncan: Or I could just add "integers" in "must be of compatible" section (shoulda thought of that first)

modules/internal/ChapelRange.chpl
```diff
@@ -343,6 +343,51 @@ module ChapelRange {
...
+  }
+
+  pragma "last resort"
+  proc chpl_compute_low_param_loop_bound(param low, param high) param {
+    compilerError("Range bounds must be integers of compatible types");
+  }
+
+  pragma "last resort"
+  proc chpl_compute_low_param_loop_bound(low, high) {
+    compilerError("param for loop must be defined over a bounded param range");
```
vasslitvinov: How about one of these options for the error text:

    the bounds of a param for loop must be params
    the range used in a param loop must be a param range

compiler/AST/ParamForLoop.cpp
vasslitvinov: How about switch this to INT_ASSERT ? Because it is not user fault that we made a mistake in the module code or the compiler.

vasslitvinov: Same for the int check below, as long as we do catch it in the module code.

test/types/range/lydia/badRange-string-multi.good
```diff
@@ -0,0 +1,2 @@
+badRange-string-multi.chpl:1: In function 'main':
+badRange-string-multi.chpl:2: error: Range bounds must be integers of compatible types
```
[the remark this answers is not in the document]
bradcray: @lydia-duncan: I can take on the task of rewording this error message (and others that share wordings suggesting ranges only support ints) if it bogs down this PR.  I just ran into it in some other work I was doing and it was arguably mine to get right in the first place.

compiler/AST/ParamForLoop.cpp
lydia-duncan: The stride not being a param is on the user, though.

test/types/range/lydia/badRange-string-multi.good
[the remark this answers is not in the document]
lydia-duncan: That would be helpful, thanks.  I've partially put it off because I was bogged down with the deep dive and slide edits, but I would be lying if the prospect of editing the message again didn't contribute slightly (but I do appreciate Vass trying to get it right).  It sounds like that was the only thing holding this up, so I will go ahead and merge so you can touch both

PR #11138, opened 2018-09-14
Add primitive for llvm.invariant.start

mppf:
Adds a new primitive, PRIM_INVARIANT_START, which
can be used to add hints for LLVM's LICM pass, at least
experimentally. It is only for manual use and
experimentation at this point.

- [x] full local and llvm testing (as part of other PRs)


mppf:
@dmk42 - let me know if you have any comments on this one, thanks

PR #11140, opened 2018-09-14
Use runtime type field in chpl__distributed for sparse domains

benharsh:
The ``chpl__distributed`` implementation for sparse domains was intentionally leaking a sparse domain in order to use its ``parentDom`` field in the type being created by ``chpl__distributed``. This ``parentDom`` is the same as the ``dom`` field in the runtime type ``domainType``.

This PR adds a new primitive, PRIM_GET_RUNTIME_TYPE_FIELD, which will be transformed by the compiler into a field access of the runtime type. This primitive accepts three arguments:

1) Static type variable representing static type of field in runtime type
2) Runtime type variable to be accessed
3) Param-string name of field in runtime type

The first field is used to allow resolution to make progress until runtime types are inserted, at which point the other fields are used to create a field-access expression.

This primitive can be used in ``chpl__distributed`` to avoid creating a temporary variable, and thus resolves a large number of leaks for tests using sparse domains. Some leaks still remain for such tests, but generally the amount leaked is significantly less.

Testing:
- [x] local + futures
- [x] valgrind
- [x] memleaks
- [x] gasnet + futures


benharsh:
@mppf: The current implementation has a standalone forv_Vec over gCallExprs, which felt unpleasant to write. Do you think it would be appropriate to repurpose ``replaceInitPrims`` and handle PRIM_INIT and this new primitive at the same time?

mppf:
> @mppf: The current implementation has a standalone forv_Vec over gCallExprs, which felt unpleasant to write. Do you think it would be appropriate to repurpose replaceInitPrims and handle PRIM_INIT and this new primitive at the same time?

Yes, we should avoid the gCallExprs if possible. I'd be OK with adding it to replaceInitPrims or another similar loop. If you add it to replaceInitPrims, you might consider changing that function name & having it call two methods for each call - replaceInitPrim and replaceRuntimeTypeGetField.

modules/internal/ChapelArray.chpl
```diff
@@ -768,8 +768,11 @@ module ChapelArray {
       // otherwise the parentDom gets destroyed in the sparse case; see
       // sparse/bradc/CSR/sparse.chpl as an example
       //
```
mppf: Looks like you need to update this comment

compiler/resolution/functionResolution.cpp
```diff
@@ -8385,6 +8385,25 @@ static void handleRuntimeTypes()
   replaceValuesWithRuntimeTypes();
   replaceReturnedValuesWithRuntimeTypes();
   insertRuntimeInitTemps();
+
+  forv_Vec(CallExpr, call, gCallExprs) {
```
mppf: Yes, let's move this to an existing for call exprs loop.

modules/internal/ChapelArray.chpl
```diff
@@ -768,8 +768,11 @@ module ChapelArray {
...
-      pragma "no auto destroy" var dom: domainType;
-      return chpl__buildSparseDomainRuntimeType(d, dom._value.parentDom);
+      proc getParentDomType() type {
+        var dom : domainType;
+        return __primitive("static typeof", dom._value.parentDom.type);
+      }
+      return chpl__buildSparseDomainRuntimeType(d, __primitive("get runtime type field", getParentDomType(), domainType, "dom"));
```
mppf: It might be nice to store the result of this primitive in a temporary. Why? Two reasons: 1) make the line shorter and 2) rely less upon how normalize adds a temporary in this particular case.

PR #11144, opened 2018-09-16
Update Dockerfiles for Chapel Release 1.18.0: Release Branch Only

awallace-cray:
When Chapel release 1.18.0 is official, Docker images built from these `dockerfiles/Dockerfile` and `dockerfiles/gasnet/Dockerfile` will be uploaded to https://hub.docker.com/u/chapel/dashboard/.  

The "Full Description" at https://hub.docker.com/r/chapel/chapel/ will be replaced with `dockerfiles/README.md`  

The versions of `dockerfiles/Dockerfile` and `dockerfiles/gasnet/Dockerfile` on master branch will help users to build Docker images from Chapel's current master branch.  The files `dockerfiles/master/Dockerfile and dockerfiles/master/gasnet/Dockerfile` do the same thing.

### DIFF
--- a/util/dockerfiles/Dockerfile
+++ b/util/dockerfiles/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM debian:9.5
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     wget \
@@ -21,12 +21,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     && rm -rf /var/lib/apt/lists/*
 
-ENV CHPL_VERSION master
+ENV CHPL_VERSION 1.18.0
 ENV CHPL_HOME    /opt/chapel/$CHPL_VERSION
 ENV CHPL_GMP     system
 
 RUN mkdir -p /opt/chapel \
-    && wget -q -O - https://github.com/chapel-lang/chapel/archive/$CHPL_VERSION.tar.gz | tar -xzC /opt/chapel --transform 's/chapel-//' \
+    && wget -q -O - https://github.com/chapel-lang/chapel/releases/download/$CHPL_VERSION/chapel-$CHPL_VERSION.tar.gz | tar -xzC /opt/chapel --transform 's/chapel-//' \
     && make -C $CHPL_HOME \
     && make -C $CHPL_HOME chpldoc test-venv mason \
     && make -C $CHPL_HOME cleanall
--- a/util/dockerfiles/README.md
+++ b/util/dockerfiles/README.md
@@ -11,19 +11,17 @@
 
 ## `chapel/chapel:<version>`
 Supported Chapel versions:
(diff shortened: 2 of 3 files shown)


mppf:
This looks OK to me but I'm not particularly familiar with these files.

awallace-cray:
@tjstavenger-pnnl - As noted, the idea is for the master branch version of these Dockerfiles to work like the ones you added under dockerfiles/master, and the release branch version of these same Dockerfiles will show how we built the Docker Hub images for the release.  Thoughts? 
Thanks.

awallace-cray:
Thanks. 

> You'd have the version tagged images in Docker Hub that represent the released software. E.g., chapel/chapel:1.18 for the upcoming release.

chapel/chapel:1.18.0. 
If we had a point release, then chapel/chapel:1.18.1, etc

> Then you'd also have another Docker image that gets periodically built -- daily, weekly, monthly? -- that keeps up with what is in the master branch in GitHub. This would be tagged chapel/chapel:master. 

chapel/chapel-master.  
Because (IMO) the docker image version (ie, optional string following the colon character) should be the Chapel git rev (or build timestamp, or something unique-ish) to identify the source snapshot that was used. With `:latest` always set to point to the most recent. 

If we simply replaced image `chapel/chapel:master` every night, every week, or whatever, it would be harder for consumers to tell the difference.

tjstavenger-pnnl:
I do see the potential confusion with master changing every build, especially when it gets changed out from underneath someone when building. Thought in some ways that's the point since it is trying to represent what is in the master branch.

If you were to go down the CI route for building this image periodically -- could you tag it with the commit hash it was built from & also master? You'd want to clean up old images after a while of doing this. Not sure how you would decide what to delete as any could be used by someone.

Though I have found some forum references to people tagging master branch builds with "master" and redeploying the same tag with an automated CI. So it isn't entirely unique, perhaps not all that common though.

awallace-cray:
Thanks, again- I am not sure of the best way to maintain docker images like this. Another question- whether to build the nightly chapel-master image "FROM" something better than the stock Debian image. Start from the Docker image from the current Chapel release, maybe.

tjstavenger-pnnl:
Another thought for a cleaner tag vs the commit hash, but still with the need to clean up old tags is to create the tag based on the build number in CI. So you'd have major.minor.release.build for the tag name. E.g., 1.18.0.build#

tjstavenger-pnnl:
I haven't found the Docker images they'd be using here, but this sounds similar: https://codefresh.io/docker-tutorial/implementing-git-flow-with-dockers/

Particularly the "Docker Flow" section.

Their org in Docker Hub is here: https://hub.docker.com/u/codefreshio/

tjstavenger-pnnl:
Another possibility would be to create a new image named chapel/chapel-master, have tags named after the builds (1.18.0.build#) and the image's :latest tag pointing to the most recent build. 

That would keep the master / development Docker image separate from the releases, and hopefully removing the master vs latest confusion.

awallace-cray:
The dockerfile images for 1.18.0 are done and will be uploaded tomorrow.

Issue #8991, opened 2018-03-28
explicit calls to init() and postinit()?


cassella:
### Summary of Question

Should it be an error to call ``init()`` and ``postinit()`` explicitly?

Currently it's not.

### Steps to Reproduce

**Source Code:**

```chapel
record R {   
  var x: int;

  proc postinit() {
    writeln("R.postinit() x: ", x);
    x = x + 1;
  }
}

var r = new R();

r.postinit();
r.init();
r.postinit();
```

**Compile command:**

chpl post-postinit.chpl

**Execution command:**

```
fortytwo@magrathea:~/src/chapel (master)$ ./post-postinit -nl 1
R.postinit() x: 0
R.postinit() x: 1
R.postinit() x: 0
```

**Associated Future Test(s):**
None yet

### Configuration Information

```
fortytwo@magrathea:~/src/chapel (master)$ chpl --version
chpl version 1.17.0 pre-release (405f366)
Copyright (c) 2004-2018, Cray Inc.  (See LICENSE file for more details)
```
```
fortytwo@magrathea:~/src/chapel (master)$ printchplenv --anonymize
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native *
CHPL_LOCALE_MODEL: flat
CHPL_COMM: gasnet *
  CHPL_COMM_SUBSTRATE: udp
  CHPL_GASNET_SEGMENT: everything
CHPL_TASKS: qthreads
CHPL_LAUNCHER: amudprun
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_ATOMICS: intrinsics
  CHPL_NETWORK_ATOMICS: none
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_AUX_FILESYS: none
```
```
fortytwo@magrathea:~/src/chapel (master)$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609
```

lydia-duncan:
Yes, it should be an error to make either of these calls outside of a method of the same name (`postinit` should only be called within a `postinit` in the same inheritance hierarchy, ditto for `init`)

bradcray:
[...]

PR #11150, opened 2018-09-17
Add future about calling init/postinit post-initialization

cassella:
Belatedly add future for #8991

### DIFF
--- a/test/types/records/init/init-post-postinit.future
+++ b/test/types/records/init/init-post-postinit.future
@@ -1,0 +1,2 @@
+bug: Explicit calls to .init and .postinit should not be allowed
+#8991
--- a/test/types/records/init/init-post-postinit.bad
+++ b/test/types/records/init/init-post-postinit.bad
@@ -1,0 +1,3 @@
+R.postinit() x: 0
+R.postinit() x: 1
+R.postinit() x: 0
--- a/test/types/records/init/init-post-postinit.chpl
+++ b/test/types/records/init/init-post-postinit.chpl
@@ -1,0 +1,14 @@
+record R {
+  var x: int;
+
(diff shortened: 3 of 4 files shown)


test/types/records/init/init-post-postinit.future
```diff
@@ -0,0 +1,2 @@
+design: Calls to .init and .postinit are allowed post-initialization
```
lydia-duncan: I would call this an error message future, since no one was arguing for allowing it.  So I would change this line to: `error message: Explicit calls to .init and .postinit should not be allowed`

cassella: I'd started with something like that.  But when I checked TestSystem.rst, I saw it describes the "error message" category as for improvements to existing error messages.  (And I didn't read the discussion in the issue as "settled".  (Which is probably why I didn't submit a future for it at the time.))

I can make the change if you'd still like.

bradcray: Paul's right that the "error message" tag is supposed to be about improving the wording/accuracy of an existing error message rather than creating an error message where none exists (which I might call a bug or a feature request depending on how we feel about it).

lydia-duncan: I think I lean towards "bug" in this case, but I don't feel strongly about it.  I prefer either "bug" or "feature request" to "design"

cassella:
I changed the future category, adopted your description of it, and removed some trailing whitespace from the test itself that github hadn't shown.  start_test calls it a clean match against the .bad file.

I force-pushed since this didn't seem like history worth preserving.

lydia-duncan:
Also came back clean for me.  Thanks!

PR #11151, opened 2018-09-17
Improve the handling of outer variables in for/forall-expressions

### LINKED ISSUE #10956: codegen problem with forall-expr over forall-expr
The following code has codegen issue because chpl_gen_main gets an extra formal:

```chapel
var D1 = {1..3};
var D2 = {1..2};
var A = [idx1 in D1] [idx2 in D2] idx1*100 + idx2;
```

test:

    test/arrays/shapes/skyline.chpl

vasslitvinov:
These updates to buildLoopExprFunctions() are motivated by skyline.chpl
and some incorrect behavior observed during debugging.

* Rework isOuterVar(); add isGlobalVar(); ignore `param` and `type` vars.

* Simplify findOuterVars() based on the above.

* Replace outer variables with newly-added formals in the
  serial/leader/follower iterator functions.

* Add outer variables as actuals to the call (see `retCall` / `iterCall`)
  that (later during compilation) becomes a call to the above
  iterator function(s).

* Do the above, as well as replace outer variables with formals
  in the top-level wrapper function (chpl__forexprNN/chpl__forallexprNN),
  uniformly regardless of `insideArgSymbol`.

* Adjust convertIteratorForLoopexpr(), see comments there.

* Update the skyline.chpl future for the updated behavior.

WHILE THERE

* Remove the 'fn' argument in a couple of function so it is easy to see
  when 'fn' is no longer updated.

* Rename 'expr' variables/formals to 'loopBody', for clarity.

* Improve list_ast for zippered ForallStmt.

* Add two "scoping" futures; rename the existing scoping.chpl
  for consistent numbering.

Resolves #10956 .


vasslitvinov:
@benharsh would you review?

Issue #9915, opened 2018-06-19
Optimize post_fma_and_wait()


ronawho:
In https://github.com/chapel-lang/chapel/pull/9876 we optimized `post_fma_and_wait()` for non-fetching atomics by yielding less frequently. We should experiment with other operations to see if they can benefit from similar operations. I imagine small puts will benefit, but I don't expect gets (or large puts/gets) to benefit from this.

PR #11176, opened 2018-09-18
Optimize blocking puts/gets and fetching network AMOs

ronawho:
This is identical to #9876, but for puts/gets and fetching AMOs insead of
non-fetching AMOs. Here we optimize post_fma_and_wait to only yield every 64
iterations while we're waiting for the network ACK/payload. It's important to
yield initially in the case of task oversubscription so we can get comm/compute
overlap. If we're not oversubscribed, we benefit from yielding less frequently
because yielding isn't free and the network ACK will often complete while we're
in the middle of a yield which would increase the total latency. 

This has a pretty significant performance impact on fetching atomics and small
puts/gets. This improves the many-to-one fetchingAMO microbenchmark by ~15%,
improves the bale indexgather code, and improves RA-rmo by 20-35% depending on
the node count. At 256 nodes this takes us from  1.2 GUPS to 1.6 GUPS:

ra-rmo-perf

This brings ra-rmo ahead of the reference (with and without bucketing) at higher node counts:

ra-rmo-perf-ref

Closes https://github.com/chapel-lang/chapel/issues/9915

### DIFF
--- a/runtime/src/comm/ugni/comm-ugni.c
+++ b/runtime/src/comm/ugni/comm-ugni.c
@@ -7706,7 +7706,6 @@ void post_fma_and_wait(c_nodeid_t locale, gni_post_descriptor_t* post_desc,
   int cdi;
   atomic_bool post_done;
   uint64_t iters = 0;
-  uint64_t yield_freq = 0x3F;
 
   atomic_init_bool(&post_done, false);
   post_desc->post_id = (uint64_t) (intptr_t) &post_done;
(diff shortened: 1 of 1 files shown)


ronawho:
@gbtitus the commit history is a little bit interesting here. I tried to do something like:

```c
  uint64_t iters = 0;
  uint64_t yield_freq = 0x3F;
  do {
    if (do_yield && (iters & yield_freq) == 0) {
      local_yield();
      yield_freq >>= 1;
    }
    consume_all_outstanding_cq_events(cdi);
    iters++;
  } while (!atomic_load_explicit_bool(&post_done, memory_order_acquire));
```

basically yield initially, and then yield every 32, 16, 8, 4, 2, and finally every iteration. For some reason this hurt performance, though I suspect it was a side-effect and not because of the yield frequency. If I did something like:

```c
  if (do_yield) { local_yield(); }
  do {
    consume_all_outstanding_cq_events(cdi);
  } while (!atomic_load_explicit_bool(&post_done, memory_order_acquire));
```

I got good performance, but adjusting the backoff one to have a starting yield_freq of `0xFFFF` (yield initially, and then never again) I got worse performance. For that reason I suspect it was more a side-effect rather than an issue with the "algorithm", but just yielding every 64 attempts seems to work well and we only use FMA for transactions less than 4K so I don't think the backoff part is too important.

PR #11191, opened 2018-09-18
Knucleotide for Chapel 1.18

bradcray:
This updates our submitted version of knucleotide to work with 1.18.  The
main difference is that where before a loop expression over an associative
domain was resulting in a 1D array result, now shape preservation is
causing it to be associative, requiring sorting to behave differently
(as Vass has already implemented on master).  While here, I also
incorporated the removal of the parSafe argument which we retired
from master some time ago but never submitted to the site (figuring
it was too minor to bother with previously).  I also updated some comments.

### DIFF
--- a/test/studies/shootout/submitted/knucleotide.chpl
+++ b/test/studies/shootout/submitted/knucleotide.chpl
@@ -53,9 +53,10 @@ proc main(args: [] string) {
 proc writeFreqs(data, param nclSize) {
   const freqs = calculate(data, nclSize);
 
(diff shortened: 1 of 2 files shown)


bradcray:
@benharsh: Just making sure that this is as you'd expect.

bradcray:
No major changes here as expected/hoped:  https://chapel-lang.org/perf/chapcs/?startdate=2018/08/31&enddate=2018/09/19&graphs=submittedknucleotideshootoutbenchmark

PR #11194, opened 2018-09-19
Add error message for user-defined constructors

benharsh:
This PR adds an error message for user-defined constructors and also removes constructors (user-defined or compiler-generated) from existing tests.  Tests that explicitly tested constructor-specific behavior have been removed (e.g. modifying const fields). Compopts files that used "--no-force-initializers" have also been removed or modified to not use that flag.

This PR also takes the opportunity to remove tests using extern classes or the deprecated 'outer' feature for nested types. When possible such tests were modified to avoid using the deprecated functionality.

Testing:
- [x] local + futures
- [x] gasnet + futures


benharsh:
A later PR will begin to remove the constructors implementation from the compiler.

test/classes/constructors/dflt-ctor-generic-uninit-field-var-err-1.good
```diff
@@ -1,2 +0,0 @@
-dflt-ctor-generic-uninit-field-var-err-1.chpl:5: error: unresolved call 'A.init()'
-dflt-ctor-generic-uninit-field-var-err-1.chpl:1: note: candidates are: A.init(v)
```
lydia-duncan: All of the tests up until here (in this commit) have a corresponding initializer test that references them.  It would be good to remove the references since these tests will no longer exist.

test/classes/constructors/generic-uninit-field-constructors.chpl
```diff
@@ -1,142 +0,0 @@
-/////////////////////////////////
```
lydia-duncan: I believe this one also has a test that references it by name

test/classes/dinan/generic_constructor.chpl
```diff
@@ -1,12 +0,0 @@
-class Foo {
```
lydia-duncan: And this one

test/classes/dinan/init_before_userdef_constructor.chpl
```diff
@@ -1,19 +0,0 @@
-proc f(x) {
```
lydia-duncan: And maybe this one?  I might be good to do a greptests for the path name to the tests in these commits, since I may miss a couple

test/classes/dinan/user_def_unique_constructor_const.chpl
```diff
@@ -1,13 +0,0 @@
-class C {
```
lydia-duncan: I think these ones do as well

test/classes/initializers/compilerGenerated/had_constructor.chpl
```diff
@@ -1,20 +0,0 @@
-// This test ensures that --force-initializers doesn't break constructor types
```
lydia-duncan: So far as I recall, all the ones after this one are probably safe from that problem

test/classes/deitz/class-constructor/method_in_constructor.chpl
```diff
@@ -1,10 +0,0 @@
-class C {
```
lydia-duncan: I think this is the last one

PR #11198, opened 2018-09-20
Rename domain.member and range.member to .contains

### LINKED ISSUE #5041: rename domain.member?
The current method name for querying if a value is in a domain is `domain.member()`.

E.g.

    var D:domain(int);

    D += 4;

    assert(D.member(4));
    assert(!D.member(3));

This method name bothers me for English/Programming Style reasons. `member` is a noun and I really want the method name to be a verb in this case.

What should we do?

1) add domain.contains()
   a) and deprecate domain.member()
   b) and keep domain.member() as a synonym
2) enable parser/build to translate `x in someDomain` into this query (#5034)
   a) and deprecate domain.member()
   b) and keep domain.member() as a synonym
3) do nothing

mppf:
* Renames domain.member and range.member to .contains
* Adds deprecation warning versions of these routines
* Updates module code to use .contains
* Updates test code to use .contains

Resolves #5041, which included broader discussion of this direction.

Possible future work: also rename dsiMember to dsiContains.

- [x] full local testing


mppf:
@ben-albrecht - could you look at reviewing this soon? thanks

modules/internal/ChapelArray.chpl
```diff
@@ -1533,16 +1533,32 @@ module ChapelArray {
...
+    /* Return true if this domain contains ``i``. Otherwise return false.
+       For sparse domains, only indices with a value are considered
+       to be contained in the domain.
+     */
+    inline proc contains(i: _value.idxType ...rank) {
+      return contains(i);
+    }
+
+    pragma "no doc"
+    inline proc member(i: rank*_value.idxType) {
```
ben-albrecht: I believe the standard practice for deprecation has been to leave the deprecated feature documented until it is removed. The documentation can be as simple as saying it's deprecated and point to the replacement.

The argument behind this has been "if the feature is usable in Chapel code, it should be documented".

PR #11210, opened 2018-09-20
Adjust llvm-global-to-wide to pass in Debug mode

mppf:
TrackingVH shouldn't be read unless it's set to non-NULL.
To fix that,
 * move generating some fns info was NULL (for these tests)
 * add a bool to indicate if a preservingFn was set

Reviewed by @dmk42 - thanks!

### DIFF
--- a/compiler/llvm/llvm-global-to-wide/RUN_CONFIG
+++ b/compiler/llvm/llvm-global-to-wide/RUN_CONFIG
@@ -12,4 +12,4 @@ ln -s ../../include/llvmVer.h llvmVer.h
 mkdir -p build
 cd build
 export CMAKE_PREFIX_PATH=$CHPL_HOME/third-party/llvm/install/linux64-gnu/
-cmake .. -DLLVM_ROOT=$CHPL_HOME/third-party/llvm/install/linux64-gnu/ -DLLVM_SRC=$CHPL_HOME/third-party/llvm/llvm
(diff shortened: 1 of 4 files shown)


mppf:
@dmk42 - could you have a look at this? Thanks

compiler/llvm/llvm-global-to-wide/RUN_CONFIG
```diff
@@ -12,4 +12,4 @@ ln -s ../../include/llvmVer.h llvmVer.h
 mkdir -p build
 cd build
 export CMAKE_PREFIX_PATH=$CHPL_HOME/third-party/llvm/install/linux64-gnu/
-cmake .. -DLLVM_ROOT=$CHPL_HOME/third-party/llvm/install/linux64-gnu/ -DLLVM_SRC=$CHPL_HOME/third-party/llvm/llvm
+cmake .. -DLLVM_ROOT=$CHPL_HOME/third-party/llvm/install/linux64-gnu/ -DLLVM_SRC=$CHPL_HOME/third-party/llvm/llvm -DCMAKE_BUILD_TYPE=Debug
```
dmk42: This looks like we are always forcing debug mode.  Is that right?

mppf: For this script, yes, but AFAIK this script is only ever used for debugging & testing the llvm-global-to-wide passes.

Issue #5165, opened 2017-01-13
Count operator always returns 1..0 when count is zero


daviditen:
## Bug Report
The range count operator ignores the range's low bound when count is 0, and always returns 1..0.  It would be better if the low bound would be kept so that the count operator can produce empty ranges with a specified low bound.

### Steps to Reproduce

writeln(0..#0); // gets: 1..0, want: 0..-1

### Configuration Information
chpl Version 1.14.0

bradcray:
A corner case to look out for on this issue which may have been related to the current implementation:

    const u = 0:uint;
    writeln(u..#0);

daviditen:
We currently have three tests that run into the issue @bradcray mentions above.

[Error matching program output for distributions/vass/densify-1-self]
[Error matching program output for types/range/hilde/count]
[Error matching program output for users/vass/crash3tuple]

Each of these three tries to build empty ranges starting from 0:uint.

noakesmichael:
1 ..  #0
2 .. #0
3 .. #0

are all "an empty range" and they are all represented internally "as if" 1..0

Of course any bounded range with high < low is an empty range but which
empty range should the count operator use for a count of 0?

And what happens if the low bound is the minimum allowable value for the
range type (e.g. const u = 0:uint)?

daviditen:
This issue is suggesting that:

```
1..#0 == 1..0
2..#0 == 2..1
3..#0 == 3..2
```

The ability to specify and keep the low bound would be useful for example when using `array.push_back()`.  You specify the low bound for a domain/array and use a count of 0, then fill the array with `push_back()`.

The `0:uint..0` and similar cases should probably issue an error.

LouisJenkinsCS:
Question: Is there a reason why this has not been patched? Is it more of a discussion thing, or a lack-of-time thing? 

If it is not intended to be supported for design reasons:  if I want to create a range with possibly zero elements, I should always just compute the index myself to avoid this; since 99% of the time I do not know the if the domain can be empty, I should just go ahead and avoid using the count operator.

bradcray:
[...]

PR #11211, opened 2018-09-20
Minor external array cleanups, suggested by Michael

lydia-duncan:
Add explicit return type to convertToExternalArray, to avoid confusion.

Also permit ExternArray to store empty arrays.  Empty arrays are
represented by the domain {1..0} in Chapel, but 1 is not normally permitted
as a lower bound for an external array.  Since we can't modify the array
bounds after it has been created, it is okay to allow the empty array to be
representable with an ExternArray.

Add a test to lock in the support for empty arrays 


modules/internal/ExternalArray.chpl
```diff
@@ -55,7 +55,8 @@ module ExternalArray {
         halt("there should only be one set of indices, not multiple dimensions");
       }
       var r = inds(1);
-      if (r.low != 0) {
+      // Allow empty arrays (which have the domain 1..0)
```
mppf: 0..-1 is just as reasonable for an empty array.
It seems to me that if you want to check this, you'd check for r.low < r.high.

mppf: I'm not personally sold that these need to support 1..0 bounds, but I'm also OK with it.

modules/internal/ExternalArray.chpl
```diff
@@ -55,7 +55,8 @@ module ExternalArray {
...
+      if (r.low != 0 && !(r.low == 1 && r.high == 0)) {
         halt("external arrays always have a lower bound of 0");
```
mppf: Should it say non-empty external arrays have a lower bound of 0?
Maybe do one more check that we're not relying on the lower bound.

modules/internal/ExternalArray.chpl
lydia-duncan: We were running into this problem because `0..#0` was getting transformed into `1..0` (https://github.com/chapel-lang/chapel/issues/5165 tracks the desire to fix that issue).  I was using `0..#0` to create the array when we know the size, and it is a lot more work for a user to create an external array with their own domain (though if they do, {0..-1} should be accepted without additional work right now).

I don't feel strongly about whether `2..1` or any other empty domain gets supported, but I think I'd rather wait until there was a strong desire to do so.

modules/internal/ExternalArray.chpl
lydia-duncan: I'm a little confused, can you clarify what you're thinking here?

mppf: 1..0 bounds would mean the lower bound could be 0, couldn't it?

lydia-duncan: Yes.  I think the error message change makes sense.  Should have clarified, I'm mostly confused about the "extra check" portion of that comment

mppf: > Maybe do one more check that we're not relying on the lower bound.

Is there code in this module that won't work if r.low == 1 and the array is actually nil?

lydia-duncan: I don't think so.  The rest of the implementation seems to only check on the size of the domain, which is stored separately instead of as a range.  This code doesn't pass the range on, just its size.

It looks like ExternDom's domRange method returns 0..#0/1..0 (which dsiGetIndices and dsiDim(s) relies on) but seems to be used in loops where an empty range would just mean the loop got skipped anyways

lydia-duncan: That said, I don't fully understand the purpose of domRange/dsiGetIndices/dsiDim/dsiDims, so it's possible there is something I'm missing

lydia-duncan: Circling back now that I am on top of things again, do you think there is an additional action item for this PR or is it good to go?  (I know there's a green approval check mark but this discussion made me uncertain)

mppf: I'm OK with it, just wanted you to think about it, which you did

lydia-duncan:
(passed paratest after master merge, merging now)

PR #11212, opened 2018-09-20
Increase commonality in runtime communication diagnostics support.

gbtitus:
Move the definition of the atomic diags counter struct (the one that has
the counters the comm layer implementations update on the fly) and the
runtime communication diagnostics/verbose flag checking and temporary
disablement into common code.  This "temporary disablement" is what we
use to ignore comms done for the specific purpose of turning on or off
comm diags or verbose comm job-wide.

While here, remove some highly detailed internals-oriented verbose comm
output having to do with strided bulk transfer in comm=gasnet.  As far
as I can tell we don't test this and it has the look of something that
should be done as debug output if it's done at all.

### DIFF
--- a/runtime/include/chpl-comm-diags.h
+++ b/runtime/include/chpl-comm-diags.h
@@ -32,13 +32,13 @@ typedef struct _chpl_atomic_commDiagnostics {
 #undef _COMM_DIAGS_DECL_ATOMIC
 } chpl_atomic_commDiagnostics;
 
-chpl_atomic_commDiagnostics chpl_comm_diags;
+chpl_atomic_commDiagnostics chpl_comm_diags_counters;
 atomic_int_least16_t chpl_comm_diags_disable_flag;
 
 static inline
 void chpl_comm_diags_init(void) {
 #define _COMM_DIAGS_INIT(cdv) \
-        atomic_init_uint_least64_t(&chpl_comm_diags.cdv, 0);
+        atomic_init_uint_least64_t(&chpl_comm_diags_counters.cdv, 0);
   CHPL_COMM_DIAGS_VARS_ALL(_COMM_DIAGS_INIT);
 #undef _COMM_DIAGS_INIT
   atomic_init_int_least16_t(&chpl_comm_diags_disable_flag, 0);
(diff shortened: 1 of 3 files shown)


gbtitus:
@ronawho no rush, I'm not going to want to merge this until Monday at the earliest.

runtime/include/chpl-comm-diags.h
```diff
@@ -29,32 +32,68 @@ typedef struct _chpl_atomic_commDiagnostics {
 #undef _COMM_DIAGS_DECL_ATOMIC
 } chpl_atomic_commDiagnostics;
 
+chpl_atomic_commDiagnostics chpl_comm_diags;
```
ronawho: By moving this into a header, won't we have different instance for each .c file that includes it? That's not a problem for ugni/gasnet right now since they're monolithic files, but if we ever split ugni or use this in multiple files in ofi, is this what we want?

gbtitus: That would be true if it there were a prior declaration with the `static` storage class specifier; it would have so-called internal linkage and each declaration would create a unique object.  But since there is no `static` declaration of it anywhere, declaring it without a storage class specifier is the same as declaring it with an explicit `extern`, and it has external linkage: all the declarations name the same object.  See the C standard, section 6.2.2.  But I'll change it to an explicit `extern` declaration to make this more clear.  (The various variable decls in `chpl-comm.h` are this way.)

gbtitus: Hmm, adding the `extern` storage class to the declaration in the .h file actually means that we have to add another declaration without any storage class in one single place somewhere (maybe `chpl-comm.c`, for example).  Without that we get undefined externs at link time, which I hadn't expected from reading the standard.  I'm good with either doing that or leaving off the `extern` in the `.h` file which, will have the same effect.  Do you have a preference?

In the meantime I'm also going to rename `chpl_comm_diags`, perhaps to `chpl_comm_diags_counters`, for clarity because it's too similar to the existing `chpl_comm_diagnostics` flag for my taste.

gbtitus: My preference would be to leave the `extern` storage class specifier off entirely, since then the variables only need to be mentioned in `chpl-comm-diags.h`, nowhere else.  (Switched to plural because whatever we do for the existing `chpl_comm_diags` also has to be done for `chpl_comm_diags_disable_flag`.)

ronawho: Works for me

PR #11226, opened 2018-09-24
Error if use statement lists the used module (or any parent) in its limitation clause

### LINKED ISSUE #10798: use-only-as rename on module results in successful compilation, but unusable symbol
### Summary of Problem
Background: I was trying to figure out whether you could rename a module on a `use` statement. The following code compiled when it should not.

```chapel
module Toolbox {
  var x = 1;
}

proc main() {
  use Toolbox only Toolbox as tb; // This compiles by itself.
  // writeln(tb.x); // Generates compile error: 'tb' undeclared
}
```

### Configuration Information
`chpl version 1.18.0 pre-release (f18926d825)`

lydia-duncan:
Prior to this change, use statements such as `use M only M as bar;` would
resolve, though the symbol could not be found in the scope that was used.
This was because the outer scope of the scope being used was also searched
for symbols and the module name was visible there.

This change adds a check at the end of finding symbols that match the
intended name which eliminates the current module being searched if
it was included in the list (determining whether to return it based on
identity, not naming).

It also extends the error message for unresolved symbols in these limitation
clauses to include the name of the module searched, so that it is more clear.

Resolves #10798

Adds four tests to ensure this works:
- an except list with the module being used
- an only list with the module being used
- an only list where the module being used is renamed
- an only list with the module defining the module being used
  - This case was unlikely to come up in user code but it was easy to prevent
Adds a test to track when the outer module has the same name as the inner
module (and the outer module is not already in scope).  Also adds a future
for the same case except that the outer module is a top level module.

Passed a full standard paratest with futures


compiler/passes/ResolveScope.cpp
```diff
@@ -663,6 +663,18 @@ void ResolveScope::getFields(const char* fieldName,
       INT_ASSERT(false);
     }
   }
+
+  if (symbols.size() > 0) {
```
lydia-duncan: I double checked and this implementation only impacts use statements, not normal scope resolution

test/visibility/except/exceptModuleItself.good
```diff
@@ -0,0 +1,2 @@
+exceptModuleItself.chpl:5: In function 'main':
+exceptModuleItself.chpl:6: error: Bad identifier in 'except' clause, no known 'Toolbox'
```
benharsh: This error message seems odd to me, because by saying 'use Toolbox' we have to know what 'Toolbox' is.

I think it would be better to have a distinct error for this kind of case where the module name cannot be used in an except/only list if it is an operand(?) of the use-statement. Maybe something like:
```
error: Module 'Toolbox' may not be present in an 'except' clause when it present in the module list of the 'use' statement.
```

lydia-duncan: What about `no known 'Toolbox' defined in module 'Toolbox'`, to clarify that we are looking in the module itself?

benharsh: Does the language allow for anything in that module to share the same identifier as the module?

lydia-duncan: This program compiles and runs:
``` chapel
module M {
  module M {
    proc whatev() {
      writeln("whee");
    }
  }
}

M.M.whatev();
```
I think it's intentional behavior, but it might be that we haven't thoroughly thought it through.  At the very least, it allows foo.chpl to contain:
```chapel 
module foo { ... } // Technically foo.foo
proc somethingElse() { ... }
```

benharsh: Ah, OK. Then your recommended change sounds good, and seems like it could replace the existing "Bad identifier..." text as well (if you want). I'd make sure to add a test for multiple modules in the use-statement as well.

lydia-duncan: We don't allow multiple modules in a use statement with an `except` or `only` list (since we use commas to separate both the modules/enums used and the symbols listed in the limitation clause, you can't have multiple modules/enums with these clauses and then it isn't quite clear whether the limitation should apply to just the last one or apply to all of the listed symbols)

lydia-duncan:
Re-running testing after the last .good file update, but I think this is ready for another look if you like.

lydia-duncan:
(passed paratest with futures, except that known failure on master)

lydia-duncan:
Should be.  I'll double check that it does work to be certain (and add a test)

lydia-duncan:
So that program with
```chapel
use M only M;
M.whatev(); // Gives `Symbol 'whatev' undeclared in module 'M'`
```
doesn't work, I believe because we are still only searching the outermost M for the symbol instead of trying to search the inner module.  However, if I wrap the outermost M in another module, the following code will work:
```chapel
module EvenMoreOuter {
  module M {
    module M {
      proc whatev() {
        writeln("whee");
      }
    }
  }
}

use EvenMoreOuter.M only M;
M.whatev(); // Prints `whee`
```

lydia-duncan:
It seems a little unfortunate, but I also think the user should probably use a different name for their module so I'm not as concerned about it.  Would it be okay with you if I file it as a future and issue for the case that doesn't work and move on?

benharsh:
Yeah, filing as a future sounds good.

Issue #4953, opened 2016-12-06
non-inlined iterators can leak memory


mppf:
Non-inlined iterators leak memory if the iteration is broken out of.

See
```
test/types/records/ferguson/leak-futures/iterate-*.chpl
```

Inlined and non-inlined iterators generate pretty different code so the potential solutions are:

1) use a different approach for freeing captured iterator variables in the inlined vs non-inlined iterator cases. Here, for non-inlined iterators, we could default-initialize the iterator class fields and use the = overload to updated them and finally call destructors on them in freeIterator.

2) use a different strategy that works with the advance() function usually freeing the memory. For example, we could generate several code blocks in freeIterators that frees only the variables that we know were created by that point in the iterator's progress (based on the iterator variable tracking the iteration).

I *think* this leak exists for inlined iterators, and so (2) is probably the way to solve it.

mppf:
My current best idea here is to adjust the 'return' from a loop controlled by an iterator to "jump" back in to the iterator at a label just after the last 'yield'. That should allow the iterator to clean up appropriately, and can leave the current autoDestroy business in callDestructors.

mppf:
I think this is also related to leaks in test/errhandling/ferguson/loopexprs-caught.chpl

cassella:
Of the tests and directories mentioned in this issue, the only one leaking today (in my config) is

test/errhandling/parallel/forall-calls-throwing-fn2.chpl

I think what it's leaking are the thrown Errors and their strings; I'm not sure if that's what this issue is about:

```
memStats: Allocated Now:                672
memStats: Allocation High Water Mark: 69224
memStats: Sum of Allocations:         72529
memStats: Sum of Frees:               71857

====================
Leaked Memory Report
==============================================================
Number of leaked allocations
           Total leaked memory (bytes)
                      Description of allocation
==============================================================
7          616        Error
7          56         string copy data
```

PR #11227, opened 2018-09-24
Fix some SparseBlockDist leaks

benharsh:
This PR fixes some memory leaks involving SparseBlockDist.

1) Rewrite SparseBlockDom. dsiLocalSubdomain to avoid using a break-ing zippered loop, which leaked iterator classes.

2) Re-use the global ``defaultDist`` variable instead of creating a new DefaultDist domain map.

3) Manually delete an unmanaged class in a partial reduction test.

These fixes bring the leaks from ~96K down to ~67K.

Testing:
- [x] local + futures
- [x] gasnet + futures
- [x] memleaks

### DIFF
--- a/modules/dists/SparseBlockDist.chpl
+++ b/modules/dists/SparseBlockDist.chpl
@@ -810,16 +810,8 @@ proc SparseBlockDom.dsiHasSingleLocalSubdomain() param return true;
 proc SparseBlockArr.dsiHasSingleLocalSubdomain() param return true;
 
 proc SparseBlockDom.dsiLocalSubdomain() {
-  // TODO -- could be replaced by a privatized myLocDom in BlockDom
-  // as it is with BlockArr
-  var myLocDom:unmanaged LocSparseBlockDom(rank, idxType, stridable, sparseLayoutType) = nil;
-  for (loc, locDom) in zip(dist.targetLocales, locDoms) {
-    if loc == here {
-      myLocDom = locDom;
-      break;
-    }
-  }
-  return myLocDom.mySparseBlock;
+  const (found, targetIdx) = dist.targetLocales.find(here);
+  return locDoms[targetIdx].mySparseBlock;
 }
 
 proc SparseBlockArr.dsiLocalSubdomain() {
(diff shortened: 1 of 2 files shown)


ben-albrecht:
> to avoid using a break-ing zippered loop, which leaked iterator classes.

Should this behavior be documented as an issue, or is this considered a user error?

modules/dists/SparseBlockDist.chpl
```diff
@@ -298,8 +298,9 @@ class LocSparseBlockDom {
   param stridable: bool;
   type sparseLayoutType;
   var parentDom: domain(rank, idxType, stridable);
-  var sparseDist = new unmanaged sparseLayoutType; //unresolved call workaround
-  var mySparseBlock: sparse subdomain(parentDom) dmapped new dmap(sparseDist);
+  var sparseDist = if _to_borrowed(sparseLayoutType) == DefaultDist then defaultDist
+                   else new dmap(new unmanaged sparseLayoutType); //unresolved call workaround
```
ben-albrecht: > `//unresolved call workaround`

I know this comment was here before you touched this, but do you know what this is referring to?

benharsh: No idea :(

benharsh:
The break+zip issue is a known bug.

ben-albrecht:
> The break+zip issue is a known bug.

Is there an issue open for that? Would we ever have reason to revert to the previous implementation after this bug is fixed?

benharsh:
I think #4953 represents the general case for the bug.

I don't think there's any reason to revert to the previous implementation.

ben-albrecht:
> I think #4953 represents the general case for the bug.
>
> I don't think there's any reason to revert to the previous implementation.

Thanks. Wanted to make sure there is a cross-reference, and to include one in the comment if we were planning to revert when the issue resolved.

modules/dists/SparseBlockDist.chpl
e-kayrakli: I tracked this one down to my first PR. Will investigate further soon.

PR #11228, opened 2018-09-24
Destroy the result of a reduction's generate()

benharsh:
This PR simply adds FLAG_INSERT_AUTO_DESTROY to the result of a reduction's generate call, fixing some leaks in the process.

Testing:
- [x] local + futures
- [x] memleaks
- [x] valgrind
- [x] gasnet


benharsh:
Combined with the sparse leaks PR, the leaks are brought down to ~51K.

PR #11230, opened 2018-09-24
File future of failing range membership tests

bradcray:
This looks to be a longstanding issue that I just stumbled into today
where membership tests between certain strided and unstrided range
pairs result in a compiler assertion error (but shouldn't).  I put
together my best attempt at a test that felt fairly exhaustive
w.r.t. different cases, though I can't guarantee that I got the .good
file correct (I'd like to review if the bug fix reveals that I got
something wrong).


bradcray:
Hi @vasslitvinov — This looks like a longstanding bug in some code you wrote.  It appears to me to be low-hanging fruit to fix, but wasn't so low-hanging that I dove into it myself.  Can you take a look?  I can file an issue against this if you'd like (i.e., if you can't put a fix together so quickly that it isn't worth my while).

vasslitvinov:
@bradcray - I got a fix, will test it. Meanwhile:
```chpl
1..10 by -2.member(4..6)
```
should be false because 'this' range does not contain 5.
If you agree, I will update .good.

vasslitvinov:
@bradcray - the fix is here: bd0cbe61bd
Passes linux64+futures paratest.
How about cherry-pick it into this PR ?

bradcray:
@vasslitvinov -- Thanks for the catch:  I fixed the problematic case (and I had it right just a few lines prior!!) but will leave the bug fix itself to you.

Issue #11062, opened 2018-09-09
string slicing by a range of codepoints


mppf:
String support currently allows for indexing into the string with `codePointUnit` but that doesn't work to enable slicing. There is not currently a way to get a string slice with code point units.

``` chapel
var x = "éú";

var firstCodePoint = 1:codePointIndex;
var y = x[1:codePointIndex];
writeln(y);

var z = x[firstCodePoint..firstCodePoint];
writeln(z);
```

I'd expect this program to output
```
é
é
```

(but see also issue #11061 - the first line output is a numeric value currently).

mppf:
I wonder if there should be a more general mechanism to make numbers with units, and then if ranges of united-numbers can be supported.

dmk42:
@bradcray - Thanks for offering to create ranges of codePointIndex records.  This is just a reminder that I'll do the rest once that's available.

bradcray:
@dmk42: Ranges of codePointIndex should now be supported as of #11231.  Let me know if you run into any problems that are obviously on my side of things.

PR #11231, opened 2018-09-24
CodePointIndex range support

bradcray:
This adds support for fully- and partially-bounded codePointIndex ranges and should permit @dmk42 to proceed with his string slicing work (issue #11062 ).  The main work here was adding the comparison and internal helper routines that ranges rely upon near the definition of the codePointIndex types.  I also added a few tests to make sure that they seemed to be working as expected.


bradcray:
Note that while writing tests leading up to this, I was surprised (as @mppf was in issue #11062) that indexing into a string with a codePointIndex returned an `int(32)` rather than something like a single-codepoint string (e.g., writeln("brad"[1:codePointIndex]) gave me 98 rather than "b").

modules/internal/String.chpl
```diff
@@ -233,6 +233,23 @@ module String {
...
+    return int;
+  proc chpl__rangeUnsignedType(type idxType: codePointIndex) type
+    return uint;
+  inline proc chpl__idxToInt(i: codePointIndex)
+    return i:int;
+  inline proc chpl__intToIdx(type idxType: codePointIndex, i: int)
+    return i: codePointIndex;
+  proc chpl__idxTypeToIntIdxType(type idxType: codePointIndex) type
+    return int;
+  pragma "no doc"
```
ronawho: I think all of these above need `no doc` too. They're currently rendered in the documentation.

modules/internal/String.chpl
```diff
@@ -233,6 +233,23 @@ module String {
     var _cpindex  : int;
   }
 
+  // Helper routines in support of being able to use ranges of codePointIndex
+  proc chpl_build_bounded_range(low: codePointIndex, high: codePointIndex)
```
ronawho: Do we want to support partially bounded codepoints so you could do something like `str[2:codePointIndex..]`?

modules/internal/String.chpl
bradcray: Huh... you're right.  I'd really gotten it into my mind that chpldoc skipped over `chpl_` prefixed things unless you were in developer mode (and thought I'd seen evidence of that in some other edits I'd done in other files...).

modules/internal/String.chpl
bradcray: Hmm, probably so.  I thought of this yesterday but apparently wanted to take a nap too badly to follow through on it.

bradcray:
> I'm not sure I'd notice if any needed functions are missing

That's what caused me to resurrect the range userAPI work... (not to suggest it's perfect, but I think it's the best we have in terms of coverage for range work).

> This really leaves me wanting interfaces/concepts.

If only someone cared about what you wanted enough to ask about project priorities.  :D

modules/internal/String.chpl
ronawho: The horrid [fixInternalDocs.sh](https://github.com/chapel-lang/chapel/blob/master/modules/internal/fixInternalDocs.sh) has a `removePrefixFunctions` routine that removes all chpl_ prefixed routines and any operators. It's not used on the string module because we want to document the operator overloads.

bradcray: OK, glad I didn't completely fabricate that impression!

modules/internal/String.chpl
bradcray: Added in 0238a805b243a1408e5bccb7cc9f458f665429d8

modules/internal/String.chpl
bradcray: nodoc'd in 0238a805b243a1408e5bccb7cc9f458f665429d8

PR #11236, opened 2018-09-26
Remove all domain._value uses from LinearAlgebra

ben-albrecht:
In response to #9926 being closed, I removed the temporary `_value` usage that referenced the issue.

I also removed all other instances of `_value` in the `LinearAlgebra` module.


ben-albrecht:
> However, I was unable to remove the `domain._value.type` case without failures.

@benharsh - is this surprising to you? Do you know if there's an open issue I should reference for this? (maybe the general issue #7673?)

ben-albrecht:
@benharsh and I found a solution to the last remaining `._value` and it has been removed.

ben-albrecht:
### Testing

- [x] OS X without dependencies
- [x] Cray testing with dependencies

Issue #9088, opened 2018-04-04
delete-free subtask: develop terminology and specification for lifetime checking


mppf:
As a Chapel user, I need to be able to understand what the lifetime checker is doing in order to reason about the error messages it produces and to write programs that compile without error.

This is not likely to break existing programs.

See #8377 and #9853 the design discussion.

Acceptance Criteria: Terminology is chosen for lifetime checking and it is described in the language specification or similar user-facing documentation.

mppf:
- [ ] adjust spec
- [ ] adjust highlighters

mppf:
We have doc/rst/technotes/lifetimeChecking.rst aka https://chapel-lang.org/docs/technotes/lifetimeChecking.html and #13413 covers the remaining task of updating syntax highlighters.

PR #11238, opened 2018-09-26
Compiler error for dereferencing a nil value

mppf:
Adds compiler error for defereferencing a value that the compiler can prove is nil.

This is implemented as a new per-function data-flow analysis that:
 * tracks references, when what the reference points to is knowable
 * tracks must aliases, when a variable must alias another variable
 * tracks variables that are allocations (and so not nil)
 * tracks variables that are definitely nil

Additionally it adds several flags so that the owned/shared functions can communicate some details to the checking pass about how they operate.

- [x] quickstart testing
- [ ] full local testing


mppf:
@ronawho and @vasslitvinov - I'm hoping you can give feedback on this PR soon, thanks.

ronawho:
High level feedback: 

Cool! 👍 

Just for the record and reference for others. I'm on board with this must-alias control-flow analysis. Doing any sort of may-alias analysis will just lead to false-positives, which I think are very unlikely to be helpful (and that sort of analysis is much more complicated.)

I think this needs some user-facing documentation about what will be caught as well as limitations. Should there be a compiler flag to disable this checking? (in case we get analysis wrong, it would be a shame for this to have no workaround.) Oh, actually maybe it's triggered by `--[no-]lifetime-checking`. Is that what we want, or should there be a separate flag for it?

It'd also be valuable to have some more tests for this feature and maybe include a few that show what we can't track?

I'm not sold on the error message for this. `error: use of nil variable` makes it sound like nil variables are never allowed. Maybe something like `error: invalid use of nil value`? I also wonder if we should try to unify this and the runtime error message for nil-checks

Longer term, I also wonder if we can use this analysis to limit the number of runtime nil-checks we add when `--nil-checks` is enabled.

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+    if (it != aliasMap.end()) {
+      loc = it->second;
+    }
+  }
+
+  if (sym->isRef()) {
+    if (loc.type == MUST_ALIAS_REFVAR) {
+      Symbol* sym = toSymbol(loc.location);
+      INT_ASSERT(sym && !sym->isRef());
+      return sym;
```
vasslitvinov: Let us rename this inner 'sym' variable to, say, 'refdSym'.

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+
+    SymExpr* thisSe = toSymExpr(call->get(1));
+    Symbol* thisSym = thisSe->symbol();
+    // Doesn't raise errors for methods on managed pointer types (e.g. owned)
+    // since:
+    //   * right now owned etc can store nil
+    //   * borrowing from nil owned isn't an error by itself
+    //     but dereferencing that nil borrow would be
+    if (isClassLike(thisSym->getValType())) {
+      // Raise an error if it was definately nil
```
vasslitvinov: spelling: definitely

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+          idxToSym.push_back(sym);
+          INT_ASSERT((int)(idxToSym.size()-1) == index);
+          index++;
+          if (debugging) {
+            printf("Tracking symbol %i %s\n", sym->id, sym->name);
+          }
+        }
+      }
+    }
+  }
```
vasslitvinov: It will be really great to move the above block stmt to a separate function, ex. populateSymToIdx().

That will improve readability of findNilDereferences().

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+              // OK, do nothing
+            } else if (was.type == MUST_ALIAS_IGNORED) {
+              nextIn[sym] = loc;
+            } else {
+              // Conflicting input -> unknown
+              nextIn[sym] = unknownAliasLocation();
+            }
+          }
+        }
+      }
```
vasslitvinov: Again, for readability, write the above as, ex. `AliasMap nextIn = combineIns(bb);`.

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+        if (dst.type < loc.type) {
+          if (debugging) {
+            printAliasEntry("in was", sym, dst);
+            printAliasEntry("in now", sym, loc);
+          }
+          dst = loc;
+          changed = true;
+        }
+      }
+    }
```
vasslitvinov: Ditto the above.

Or, have a single function for combining and propagating, to make the body of the do-while loop very simple.

compiler/resolution/nilChecking.cpp
```diff
@@ -0,0 +1,782 @@
...
+              // if rhs is a ref, just propagate any MUST_ALIAS_REFVAR
+              OUT[lhsSym] = OUT[rhs];
+            else
+              // otherwise, make a new MUST_ALIAS_REFVAR
+              OUT[lhsSym] = refAliasLocation(se->symbol());
+          } else {
+            OUT[lhsSym] = unknownAliasLocation();
+          }
+        }
+      }
```
vasslitvinov: Let's label the above close-curly with the condition that it closes. It looked to me it was `lhsSym->isRef()` , however judging by the "move-ish" comment below that's not the case. So, clarification is needed!

mppf:
@ronawho @vasslitvinov - regarding a flag to enable/disable

> Oh, actually maybe it's triggered by --[no-]lifetime-checking. Is that what we want, or should there be a separate flag for it?

Yeah, it currently uses --lifetime-checking and --no-lifetime-checking. I can't use `--nil-checks` since that controls the runtime flags. It's separate from, but related to, lifetime checking. I'd generally prefer we avoid setting the expectation that every compiler pass will have a flag to turn it off. E.g. do we have a flag to turn of const-ness checking? That is why here I'm trying to group it with lifetime checking. I'm open to creating a flag like `--no-compile-time-nil-checking` but I'd rather leave it with `--lifetime-checking`.

mppf:
@ronawho @vasslitvinov 

> some user-facing documentation about what will be caught as well as limitations.

Note that `--lifetime-checking` doesn't currently have such documentation (although it is "documented" to some degree in the release notes). Additionally it's a developer flag (not a user one). Would it be reasonable to include this in the TODO for user-facing documentation for `--lifetime-checking`, #9088? Or, should we seek to keep these separate in the user's mind (e.g. by creating a new flag)?

vasslitvinov:
Nil-deref checking and lifetime checking are non-overlapping in my mind. So I suggest separate documentation for the two.

Again, I am not excited about creating additional flags. I propose to flag-ify nil-checking compiler pass only when there is user demand.

ronawho:
> Again, I am not excited about creating additional flags. I propose to flag-ify nil-checking compiler pass only when there is user demand.

I worry that the user demand will be "my program doesn't compile because of a bug in your nil checker". If we believe we'll be able to suggest code workarounds or something else, then I'm with not adding a new flag, but otherwise I'd like to see one, even if it's just a developer flag.

compiler/resolution/nilChecking.cpp
mppf: I'd like to leave this one here because it has to do with the fixed-point algorithm's convergence. I moved the others out so now it should fit in a screen though.

mppf:
@vasslitvinov @ronawho - I believe I've addressed your feedback. I'm going to do another testing run, and then is this good to go? thanks.

compiler/resolution/lifetime.cpp
```diff
@@ -274,62 +275,85 @@ void checkLifetimes(void) {
...
 
-    if (debugging) {
-      printf("Visiting function %s id %i\n", fn->name, fn->id);
-      nprint_view(fn);
-      gdbShouldBreakHere();
-    }
+  if (fCompileTimeNilChecking) {
+    // Determine cases where the compiler can prove
+    // a reference-type variable is 'nil'
+    // TODO: enable this for internal modules as well.
```
ronawho: Can you make an issue for this?

doc/rst/technotes/nilChecking.rst
```diff
@@ -0,0 +1,34 @@
...
+Checking for Nil Dereferences
+=============================
+
+The Chapel compiler includes two different mechanisms to check for nil
+dereferences. First, at compile-time, the compiler looks for places where
+it can prove a `nil` will be dereferenced or a method called on `nil`.
+When it can find such a case, it raises an error.
+
+The second mechanism is to check each at each method called on an object
+that the object is not `nil`. This works at run-time using checking code
```
cassella: This sentence feels like it has too many "each"es.  (And maybe note earlier that this second mechanism is at run-time, like the first thing said about the first mechanism was that it's at compile-time?)

doc/rst/technotes/nilChecking.rst
```diff
@@ -0,0 +1,34 @@
...
+ * it does not track may-alias sets
+ * it is not interprocedural
+
+Nonetheless we hope that the additional checking provided by this pass
+is useful.
+
+ * complex control flow
+ * complex alias analysis requirements
+ * interprocedural
+   The analysis cannot de be defeated by  makes conservative assumptions about
```
cassella: This list feels like a rough draft of the previous list?

bradcray:
In getting bogus nil-checking errors, I was excited to find with my tab completion that `--no-nil-checks` was available as a flag, and then disappointed that it didn't didn't actually turn off the checks.  Then I realized that I was using the longstanding execution-time checks flag.  I was about to file a feature request to turn off the compile-time nil checks before seeing on this PR that one was added.  Renaming it `--no-nil-checks-in-compiler` would help developers find it better via tab completion.  The false negative in issue #11813 also made me wonder whether it should be exposed to the user (or simply combined with the existing `--no-nil-checks` flag?).

mppf:
Hmm. The reason I made a separate flag was that I figured that `--no-nil-checks` might be set in order to aid performance, and we wouldn't want to disable the compile-time checking in that setting.

It'd be OK with me to have `--no-compile-time-nil-checks` and `--no-runtime-nil-checks` flags (possibly developer only) and then have `--no-nil-checks` activate them both. How would you feel about such a thing?

I'd be OK with a name to support better tab completion but find `--no-nil-checks-in-compiler` really awkward wording.

bradcray:
> The reason I made a separate flag was that I figured that --no-nil-checks might be set in order to aid performance, and we wouldn't want to disable the compile-time checking in that setting

Oh yeah, that's a good point.  `--fast` and `--no-checks` throw this, and it'd be inappropriate for truly illegal programs to stop getting compile-time checked as a result...

> How would you feel about such a thing?

Only if `--no-checks` only disabled the runtime ones (and I think the runtime checks would therefore need to continue to be a user flag?).  But let me stew on this a bit more... I don't know if it's worth that big of a change.

> really awkward wording.

I was trying to give them a common prefix so that as a developer tab-completing, but I agree it's a bit of a mouthful.  Then again, if its a developer flag, maybe it doesn't matter if it is a mouthful.  (I was really trying to create a common prefix that wasn't either flag in its entirety so that, as a developer, I'd be forced to choose between them.  So maybe I really wanted `--no-nil-checking-in-compiler`?  :)

PR #11242, opened 2018-09-26
Improve comm=ugni fixed heap behavior near registration limit.

### LINKED ISSUE #10700: comm=ugni has poor behavior near the limit of registerable memory
### Summary of Problem

There are some problems with comm=ugni error reporting and related behavior when at or above the  limit of registerable memory.

### Steps to Reproduce

Run anything with a ridiculous registered heap size request, like this:

```shell
[prompt] CHPL_RT_MAX_HEAP_SIZE=-1 ./t -nl 2
warning: Aries TLB cache can cover 256G with 16M pages; with 17179869183G heap,
         cache refills may reduce performance
Hello, world! (from task 0 of 28)
Hello, world! (from task 27 of 28)
Hello, world! (from task 3 of 28)
Hello, world! (from task 2 of 28)
Hello, world! (from task 1 of 28)
<internal>: error: Out of memory allocating "tasking layer unspecified data"
srun: error: nid00108: task 0: Exited with exit code 1
srun: error: nid00109: task 1: Terminated
[prompt]
```

First, the warning is incorrect; it seems to be using the Gemini TLB size of 16k entries rather than the Aries PTE cache size of 512 entries.

Second, the code is written to try to allocate and register the largest heap it can when the request is too big, so we shouldn't be running out of memory here.  (Note that with a fixed size heap requested we shouldn't be using the `HUGETLB_NO_RESERVE` workaround or getting the related `SIGBUS`, so that's not the problem.)

Here's a second problem:

```shell
[prompt] CHPL_RT_MAX_HEAP_SIZE=124g ./t -nl 2
internal error: 1:comm-ugni.c:2369: register_mem_region() failed, error code 2
internal error: 0:comm-ugni.c:2369: register_mem_region() failed, error code 2
srun: error: nid00109: task 1: Exited with exit code 2
srun: error: nid00108: task 0: Exited with exit code 2
[prompt]
```

This was on 128g nodes, so it's not surprising that we can't get 124g.  But we should be making a registration attempt as part of trying to get as much memory as possible, so we shouldn't see the registration failures.  In other words, we should be creating the heap at the largest registerable size we can.

gbtitus:
The ugni comm layer could behave badly in several different ways, if the
user specified a fixed heap size that was either larger than what was
present on the node or could be registered, depending on circumstances.
Here, fix these problems.

Most egregiously, some earlier work of mine seems to have resulted in
our setting `HUGETLB_NO_RESERVE=yes` even for fixed heaps, as long as we
were using hugepages.  This was incorrect; doing so broke the code that
used hugepage allocation failures to maximize a fixed heap size, because
with `HUGETLB_NO_RESERVE=yes` even ridiculously large allocations (for
example, more than the physical memory present) succeed, only to cause
registration failures later when the pages cannot be created.  Thus,
here, only set `HUGETLB_NO_RESERVE` and also `CHPL_JE_MALLOC_CONF` in the
launcher when the heap is dynamically sized.  Also, adjust the sanity
checking in the runtime to only complain about the lack of these
variables when the heap is dynamically sized, and improve the warning
messages it emits so that this is more clear.  Improve the related
documentation.  Finally, make the expected output for the hugepage
sanity test match the new messages.

Disallow negative values in environment variables that are expected to
contain sizes.  Thus things like `CHPL_RT_MAX_HEAP_SIZE=-1` now cause a
warning and are ignored, instead of ending up specifying a gigantic heap
due to type punning.  Obviously one can still specify a too-large
unsigned value for the heap size, but this is still a desirable change.

Improve the message printed when the heap size is larger than can be
registered (Cray XE systems) or larger than desirable for best
performance (Cray XC systems), and add a test for this message.

Finally, speed up the code that tries to find the largest workable fixed
heap size when the user's specified size is too large, while at the same
time improving the result if that starting point was really huge.

This resolves #10700 .


gbtitus:
This will probably be easier to review commit-by-commit.

Issue #11039, opened 2018-09-07
for/forall expression with arrays as elements does not work


vasslitvinov:
The following code currently does not work:

```chpl
var ARR = [222,333];
var BRR = for idx in ARR do [9999];
```

The way it is implemented is that first a to-become-BRR array is created as an `[ARR.domain] [1..0] int`, then each of its elements `elem` is assigned as `elem = [9999]` . Since the domain of each `elem` is `{1..0}`, this assignment fails due to size mismatch.

Cf. in the old shape-less world, each `elem` would be **initialized** from `[9999]` , which succeeds. There, `BRR` is created **after** all its elements have been computed. The corresponding code is:

```chpl
iter ITR() { yield 222; yield 333; }
var BRR = for idx in ITR() do [9999];
// BRR.type == [1..2] [BRR[1].domain] int
```

BTW an open question is what the type `BRR.eltType` should be. It makes sense for it to be the same as `BRR[1].type`, i.e. the type of the first element of the resulting array, if all BRR's elements are of the same type. Which will be always the case if `BRR[1].type` is neither an array nor a domain, i.e. does not have a runtime-type component.

If BRR's elements are arrays, as in the above code examples, it may well be that `BRR[1].domain` is different than `BRR[2].domain`. They may be different objects or (skyline) even have different sizes. Which of those domains should be the domain of `BRR.eltType` ? The current implementation makes that an empty domain, which does not work as discussed above.

Test:

    test/arrays/vass/literal-indexing.chpl

vasslitvinov:
I am thinking that `noinit` on arrays is not exactly what we need for this.

Here is what I think happens when we use `noinit` to create the array to become `BRR` in the main issue text above. To emphasize my concern, consider the case where `ARR.eltType` is a Block-distributed domain. For example:

```chpl
// user code
proc makeArray(elm) return (some newly-created Block-distributed array);
var ARR = [222,333];
var BRR = for elm in ARR do makeArray(elm);
```
```chpl
// potential `noinit`-based implementation
const D = {1..n} dmapped Block({1..n});
type ElmType = [D] real;
var BRR: [ARR.domain] ElmType = noinit;
for (b, elm) in zip(BRR, ARR) do
  b.init(makeArray(elm));
```

My understanding is that `noinit` on an array implicitly invokes `noinit` recursively on each of its elements. In our case, that means creating a Block-distributed array (i.e. of the type ElmType) in each element of `BRR`, initialized with `noinit`.

[...]

PR #11243, opened 2018-09-27
Fix valgrind failure in skyline.chpl

### LINKED ISSUE #11149: valgrind failure in a nested forall expression
The following code:

```chpl
var B = [i1 in {1..n}]
          [i2 in {1..i1}]
            i1*100 + i2;
```

performs read-after-delete, as detected by valgrind.

The cause is that the following function in ChapelIteratorSupport.chpl:

```chpl
  proc iteratorToArrayType(type t:_iteratorRecord) type {
    var A:[1..0] iteratorToArrayType(__primitive("scalar promotion type", t));
    return A.type;
  }
```

deletes `A` including its domain, even though the domain is returned out of it as part of the runtime type object and is accessed later.

Test:
```
arrays/shapes/skyline.chpl
```

BTW I believe only the following tests in our test tree exercise this function:

```
arrays/shapes/skyline.chpl
expressions/loop-expr/scoping-*.chpl       # not sure about scoping-2 and scoping-3
trivial/deitz/statements/nested_fors.chpl
trivial/deitz/statements/nested_fors2.chpl
```

We need to make sure they still pass as they did before.

vasslitvinov:
The valgrind failure was caused by a call to this function
when creating the 'result' array in the shapeful chpl__initCopy():

``` chapel
  proc iteratorToArrayType(type t:_iteratorRecord) type {
    var A:[1..0] iteratorToArrayType(__primitive("scalar promotion type", t));
    return A.type;
  }
```

This function auto-destroys A prior to returning.
That deletes the domain class for A.domain.
However, that class is used as part of the runtime type
that this function returns at run time.
Accessing this runtime type resulted in read-after-delete.

We had not observed this before because we invoked this function
only in the shape-less case where the resulting array DOES have
some elements. In which case the runtime type is unused.
BTW, that happened only in these non-future tests:

```
  expressions/loop-expr/scoping-1.chpl
  trivial/deitz/statements/nested_fors.chpl
  trivial/deitz/statements/nested_fors2.chpl
```

The solution is to rewrite the above function to declare A's domain
explicitly and annotate it with pragma "no auto destroy".
This follows the approach in the existing chpl_buildStandInRTT(arrType).

WHILE THERE

* Create that domain using chpl_buildStandInRTT(), for uniformity
with other cases / wider applicability.

* Return chpl__buildArrayRuntimeType() instead of A.type,
as in existing chpl_buildStandInRTT(arrType).

* Replace the separate iteratorToArrayType() with an overload
of chpl_buildStandInRTT() for _iteratorRecord types. That way it can
apply more generally. This overload creates the type of the array
that this iterator creates when stored in a variable.

* Remove "no doc" pragmas because I think this whole file is "no doc".
Remove "inline" annotations in the cases where I do not think they
change anything.

PRIM_STATIC_FIELD_TYPE

I need to get the type of the _shape_ field of the _iteratorRecord
type passed to the new chpl_buildStandInRTT(), formerly iteratorToArrayType().
Because that's the type that we are creating the standin domain for.
For that, I introduce the new primitive, PRIM_STATIC_FIELD_TYPE.
As with PRIM_STATIC_TYPEOF, it returns the compile-time type.
If that time has a runtime component, it will be uninitialized.

Aside reminder: this is the case where result's element type is itself
an (inner) array. We create 'result' before we have seen any
of its upcoming elements, so we cannot use any of the elements'
types. So we are creating the standin domain to use in result's
element type's runtime type.

I used some trickery to avoid introducing this primitive in #10941.
I did not find a way to avoid it here. The closest solutions were:

* Use PRIM_QUERY. That does not work because the _shape_ field
is neither isInstantiatedField(), nor a type or param field.

* Use PRIM_GET_RUNTIME_TYPE_FIELD. That does not seem applicable,
as _iteratorRecord is not a runtime type.

* Create `var ir: irType`. This does not work because _iteratorRecord
types do not have _defaultOf.

Now that we have this primitive, I used it to replace some of the code
from #10941, i.e. procs instanceType, domInstanceType, arrInstanceType.

UPDATE ON test/arrays/shapes/skyline.chpl

It now completes cleanly, and leaks only the standinDomain, as in
the other cases that create a standinDomain. However, each element
of the resulting array is empty. Even though it (each element) IS
assigned a variable-length array. So it remains a .future #11254.

IN SUMMARY / FUTURE WORK:

This is incremental progress. Still, it does not help an end-user:
* the "stand-in" domains are still leaked
* arrays/shapes/skyline.chpl still does not work right

We may want to improve existing primitives to replace
PRIM_STATIC_FIELD_TYPE with a use of other primitives.

Testing: linux64+futures, valgrind, memleaks.
The newly-added tests leak their standinDomains,
which is a pre-existing issue.

With contributions by @mppf.

Resolves #11149.


vasslitvinov:
@mppf - would you review?

modules/internal/ChapelIteratorSupport.chpl
```diff
@@ -119,8 +105,8 @@ module ChapelIteratorSupport {
     return chpl_buildStandInRTT(instanceObj);
```
mppf: If we keep "static field type", it's probably worth making this case use "static field type".

modules/internal/ChapelIteratorSupport.chpl
```diff
@@ -200,29 +219,37 @@ module ChapelIteratorSupport {
...
   }
-  pragma "no doc"
-  inline proc chpl_iteratorFromForExpr(ir: _iteratorRecord) param {
+
+  // This is the static type of chpl_computeIteratorShape(ir).
+  proc chpl_iteratorShapeStaticTypeOrVoid(type ir: _iteratorRecord) type
+  {
+    use Reflection;
+    if hasField(ir, "_shape_") then
+      return __primitive("static field type", ir, "_shape_");
```
mppf: Could this be PRIM_STATIC_TYPE followed by a PRIM_GET_MEMBER_VALUE ?

modules/internal/ChapelIteratorSupport.chpl
```diff
@@ -75,23 +75,9 @@ module ChapelIteratorSupport {
...
-    return t;
-  }
-  pragma "no doc"
-  proc iteratorToArrayType(type t:_iteratorRecord) type {
-    var A:[1..0] iteratorToArrayType(__primitive("scalar promotion type", t));
-    return A.type;
-  }
-  pragma "no doc"
   proc iteratorToArrayElementType(type t:_iteratorRecord) type {
     return chpl__unref(
```
mppf: is the chpl__unref call still necessary here? I have a feeling it's no longer needed. Could you check?

test/arrays/shapes/skyline.future
```diff
@@ -1,2 +1,2 @@
-bug: valgrind failure
-#11149
+bug: forall-over-forall results in empty array
+#11039
```
mppf: 11039 is certainly related but doesn't seem to describe the current behavior so seems more of a distraction. I'd consider closing that one and making a new one with content pasted from your PR message.

modules/internal/ChapelIteratorSupport.chpl
vasslitvinov: Gives me "illegal return of value where type is expected".

test/arrays/shapes/skyline.future
vasslitvinov: I created #11254 for the new skyline failure mode. However, #11039 still describes the current behavior and needs to be resolved.

modules/internal/ChapelIteratorSupport.chpl
vasslitvinov: The only test that fails with chpl__unref() removed is:

    expressions/loop-expr/zip-arrays.chpl

I am leaving the investigation as future work.

mppf: Could you leave a comment here next to the chpl__unref call to point out that it might not be necessary, but one should check that test first?

PR #11250, opened 2018-09-27
With --gen-ids, emit LLVM IR metadata for Chapel AST ids

### LINKED ISSUE #11161: Store AST IDs in LLVM IR
As a Chapel developer, I would like to be able to map LLVM IR back to AST IDs so that I can more easily debug problems with the compiler. I'm accustomed to being able to do that with the generated C by using the `--gen-ids` flag.

This task is to add a similar mechanism for the LLVM backend. It will start by determining a reasonable way to store this information, where adding LLVM Metadata or debug information are potential directions.

mppf:
Resolves #11161

Updates CallExpr::codegen and FnSymbol::codegenDef.

- [x] full local testing
- [x] full local testing with --llvm --gen-ids

Reviewed by @dmk42 - thanks!


bradcray:
I'm curious whether you checked to see whether adding IDs to the LLVM IR adds significant compilation time/space overhead or not?  I'd still be inclined to have this behavior be always-on by default since it would seem to have much upside if there's no downside (where having them on for code generation has downsides in terms of code size, compilation time, and arguably readability).

mppf:
I don't think it should be on by default right now since I'm expecting we'll only use it for debugging.
I see modest (3% or so) decreases in compilation speed.

Personally, if I'm tracking AST ids, I'll also run --log, so I don't mind also adding --gen-ids. I don't see what benefit it would have to put these details in the LLVM IR when not doing such debugging.

bradcray:
> I don't see what benefit it would have to put these details in the LLVM IR when not doing such debugging.

The benefit I was envisioning was not having to remember to throw the flag when debugging the LLVM IR or to have to quit an active debugging session to re-run with the flag thrown if it was required.  But if it adds 3%, I think that's arguably a reason to leave things as-is.

PR #11255, opened 2018-09-28
Avoid handling overrides where parent has 'where false'

mppf:
This PR helps to solve a problem with override checking that came up when the 
compiler was adjusted to resolve concrete functions.

It adjusts override processing to not consider it an override if the parent 
function has a `where` clause evaluating to false. Such functions cannot be
called so there's no possibility that they'll need to be virtually dispatched.
As a result, child functions matching such a parent function should not be
considered overrides, subject to override errors, or added to the virtual
methods map.

Reviewed by @bradcray.

### DIFF
--- a/compiler/resolution/virtualDispatch.cpp
+++ b/compiler/resolution/virtualDispatch.cpp
@@ -368,7 +368,10 @@ static void checkIntentsMatch(FnSymbol* pfn, FnSymbol* cfn) {
 static void resolveOverride(FnSymbol* pfn, FnSymbol* cfn) {
   resolveSignature(cfn);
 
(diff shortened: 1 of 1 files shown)


bradcray:
This looks good to me!  Just noting for posterity, "where false" here doesn't literally mean "where false" but more generally "where <some param expression evaluating to false>".

bradcray:
Belatedly: Is there a test that would demonstrate the difference between life before and after this PR?  I tried to create one quickly, but failed.

mppf:
@bradcray - I'm not sure it's user-visible. I think we might have created a virtual method call in one case that's not necessary, but that wouldn't change program behavior.

PR #11258, opened 2018-09-28
Better error messages for trying to create nil refs

### LINKED ISSUE #11257: ref to nil causes weird error message
```chapel
ref x = nil;
```
reports
```
error: Cannot set a non-const reference to a const variable.
```

This error message seems like a weird one for this specific case. Please fix.

Are nil references allowed? When can they occur? I am by no means suggesting we should have nil references if they cannot occur.

mppf:
Resolves #11257.

Gives a better error message for programs such as
``` chapel
ref x = nil;
```

Intentionally does not give an error for the `const ref` case such as
``` chapel
const ref y = nil;
```
since this would be allowed for integer literals and `nil` can be considered analogous.

- [x] full local testing

### DIFF
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -2212,12 +2212,6 @@ static void normRefVar(DefExpr* defExpr) {
                    "References must be initialized when they are defined.");
   }
 
-  if (SymExpr* initSym = toSymExpr(init)) {
-    if (initSym->symbol() == gNil) {
-      USR_FATAL_CONT(var, "References must be non-nil");
-    }
-  }
-
   // If this is a const reference to an immediate, we need to insert a temp
   // variable so we can take the address of it, non-const references to an
   // immediate are not allowed.
@@ -2255,8 +2249,12 @@ static void normRefVar(DefExpr* defExpr) {
     }
 
     if (error == true) {
-      USR_FATAL_CONT(sym,
-                     "Cannot set a non-const reference to a const variable.");
+      SymExpr* initSym = toSymExpr(init);
+      if (initSym && initSym->symbol() == gNil)
+        USR_FATAL_CONT(sym, "Cannot create a non-const reference to nil");
+      else
+        USR_FATAL_CONT(sym,
+                       "Cannot set a non-const reference to a const variable.");
     }
   }
 
--- a/compiler/passes/normalize.cpp
+++ b/compiler/passes/normalize.cpp
@@ -2212,6 +2212,12 @@ static void normRefVar(DefExpr* defExpr) {
                    "References must be initialized when they are defined.");
   }
 
+  if (SymExpr* initSym = toSymExpr(init)) {
+    if (initSym->symbol() == gNil) {
+      USR_FATAL_CONT(var, "References must be non-nil");
+    }
+  }
+
   // If this is a const reference to an immediate, we need to insert a temp
   // variable so we can take the address of it, non-const references to an
   // immediate are not allowed.
--- a/test/statements/ferguson/ref-to-nil.chpl
+++ b/test/statements/ferguson/ref-to-nil.chpl
@@ -1,0 +1,2 @@
+const ref x = nil;
+ref y = nil;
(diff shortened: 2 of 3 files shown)


bradcray:
I think that `const ref y = nil;` arguably could be argued to be legal even though I think it's a bit weird / unusual and lead to confusion as in Bryant's issue that spawned this PR (i.e., "is a ref a pointer that I can re-assign?).  Specifically, if you think of `nil` as being a named constant value, `const ref y = nil;` doesn't seem so different than:

```chapel
const zero = 0;
const ref y = zero;
```

OTOH, we don't permit a const ref to refer to a param itself:

```chapel
const ref y = 0;  // illegal
```

so maybe the same should be true for `nil` (do/should we consider `nil` a param value of sorts?).

In any case, I feel a little surprised that we would prefer to specialize this error message in normalize (which we're trying to get away from) over checkResolved.cpp where the other checks for refs-to-params occur (and where the current message is being generated).  Though maybe the problem is that the nil is much harder to find / recognize there?

bradcray:
On second thought, `nil` doesn't quite seem like a `param` value to me in that it has an execution-time representation where `param`s are able to be folded away at compile-time?  So maybe it really is more `const`-like?  In any case, it still seems reasonable to prevent references to them being taken, similar to the `const ref y = 0;` case.

bradcray:
I'm still not convinced the error message in this PR is worded correctly.  The literal interpretation of `ref y = nil;` is not "set this reference (pointer) to nil" but rather "create a reference to nil".  So it seems to me that the error message should be something more like "Cannot create a reference to `nil`" rather than implying that the declaration is trying to set the reference to `nil`.  Such a wording also has the benefit of explaining better what the statement is doing to someone who thinks of it as setting a pointer to nil (where the current one leaves open the possibility that refs could simply be pointers, simply ones that can't store a nil value).

By analogy if for something prevents me from legally saying `ref x = y;` I wouldn't say that you can't assign `x` the value `y`, but rather than you can't create a reference to `y`.

mppf:
> In any case, I feel a little surprised that we would prefer to specialize this error message in normalize (which we're trying to get away from) over checkResolved.cpp where the other checks for refs-to-params occur (and where the current message is being generated).

@bradcray - when I was debugging the compiler to understand Bryant's issue, I found the error we were running into here:

https://github.com/chapel-lang/chapel/blob/e6c2f19ec4fc33485a3830347a59b3e55779ed1b/compiler/passes/normalize.cpp#L2252-L2253

That is the reason I added the checking to normalize.

> On second thought, nil doesn't quite seem like a param value to me in that it has an execution-time representation where params are able to be folded away at compile-time? So maybe it really is more const-like? In any case, it still seems reasonable to prevent references to them being taken, similar to the const ref y = 0; case.

'0' is a param but the value of it is represented also at compile-time. I think it'd be reasonable to view `nil` as a param. That's just not what the compiler does today (rather it uses a bunch of special nil code in the right places).


Regarding the exact wording of the error message, I'm not attached to it in any way.

Brad is recommending we change

> References must be non-nil

to

> Cannot create a reference to nil

and I'd be happy to do that.

Interestingly, you *can* create a reference to a *value holding nil*, e.g.

``` chapel
class MyClass { }
proc foo(ref x: MyClass) {
  writeln(x);
}

var x: MyClass = nil;
foo(x);
```

Nonetheless as I think we're agreeing
``` chapel
ref x = nil;
```

can and should generate some kind of error. The reason I think it should generate an error is that it probably represents some misunderstanding on the programmer's part.

Finally, note that this PR currently emits an error also for the `const ref` case:
``` chapel
const ref x = nil;
```

but this is the case where an error would not be emitted for a param, supposing we consider param integers to be thematically similar to `nil`.

``` chapel
const ref x = 0;
```

Nonetheless I think in the exact case of `const ref x = nil` it's still likely to represent a programmer's misunderstanding and so should be an error. Note that before this PR, it simply failed to compile at code-generation time with an inscrutable error.

bradcray:
Note that the PR message needs to be updated w.r.t. const ref cases before using it as the merge message.

PR #11267, opened 2018-10-01
Enable LLVM assertions for ASSERTS=1 builds

### LINKED ISSUE #11266: nightly LLVM testing should include an asserts-on build
Right now, to save in compile time and library size, asserts are disabled by default for the LLVM libraries built with Chapel. To enable them, one can activate a Debug build with `CHPL_LLVM_DEVELOPER=1` which relates to this Makefile code:

https://github.com/chapel-lang/chapel/blob/b1c96abc3934fb649415b650e81436f2b250c817/third-party/llvm/Makefile#L8-L13

We should at least test "Hello World" in ~~~this Debug LLVM~~~ an assertions-enabled configuration and potentially other tests.  `--fast` might be worthwhile. test/performance/ferguson/ might be a good candidate as it tests --llvm-wide-opt which includes custom LLVM passes.

mppf:
Resolves #11266.

- [x] verified CHPL_DEVELOPER gets an asserts build of LLVM
- [x] verified `make ASSERTS=1` gets an asserts build of LLVM
- [x] verified nightly command runs hellos

### DIFF
--- a/doc/rst/usingchapel/building.rst
+++ b/doc/rst/usingchapel/building.rst
@@ -131,6 +131,7 @@ are described below. Set the value to 1 to enable the feature.
   OPTIMIZE  Enable optimizations (e.g. add -O3 to C compiler).
   PROFILE   Enable profiling support (e.g. add -pg to C compiler).
   WARNINGS  Promote backend C compiler warnings to errors.
+  ASSERTS   Enables correctness assertions in the compiler and runtime.
   ========  =======================================================
 
 
--- a/third-party/llvm/Makefile
+++ b/third-party/llvm/Makefile
@@ -9,7 +9,14 @@ include $(CHPL_MAKE_HOME)/make/Makefile.base
 ifdef CHPL_LLVM_DEVELOPER
 CHPL_LLVM_DEBUG = -DCMAKE_BUILD_TYPE=Debug -DLLVM_OPTIMIZED_TABLEGEN=1
 else
+
+# activate LLVM asserts if it's a WARNINGS=1 build
+ifeq ($(WARNINGS), 1)
+CHPL_LLVM_DEBUG = -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+else
 CHPL_LLVM_DEBUG = -DCMAKE_BUILD_TYPE=Release
+endif
+
 endif
 
 LLVM_SRC_FILE := llvm/CMakeLists.txt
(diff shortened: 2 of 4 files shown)


mppf:
@ronawho - any feedback on this? thanks

ronawho:
`WARNINGS` is documented to `Promote backend C compiler warnings to errors.`, so it seems like a weird thing to trigger asserts off of. i.e. I try to build with `WARNINGS=1` to catch any build-time issues and I'd find it weird if that impacted performance. Oh, but I guess this would only impact compiler performance? If you want to go through with this I think we need to change the documentation for what WARNINGS controls. Either that or we can add a new ASSERTS flag.

Semi-side-note: I'd love to have a configuration that enables asserts/debug stuff for all 3rd party libraries and tests that as part of nightly.

ronawho:
Asserts aren't free. I don't think we want to enable them for optimized builds.

Some third-party libs also have more expensive checks that I think could be valuable to run in a different config. I think it's `--enable-paranoia` in qthreads and I think jemalloc and gasnet have similar "expensive checks"

mppf:
@ronawho - I understand you to mean:
 1. You'd like asserts to be enabled by `make ASSERTS=1` or something along those lines
 2. You'd like these to be generally off in nightly testing
 3. You'd like us to add `ASSERTS=1` in some (possibly new) nightly testing configurations.

Do I understand correctly?

ronawho:
> @ronawho - I understand you to mean:
> 1. You'd like asserts to be enabled by `make ASSERTS=1` or something along those lines
> 2. You'd like these to be generally off in nightly testing
> 3. You'd like us to add `ASSERTS=1` in some (possibly new) nightly testing configurations.
> Do I understand correctly?

Yeah, that's my preference.

mppf:
@ronawho - does this look good?

ronawho:
Add `ASSERTS` documentation to [doc/rst/usingchapel/building.rst](https://github.com/chapel-lang/chapel/blob/master/doc/rst/usingchapel/building.rst#makefile-options) to say something like "Enables correctness assertions in the compiler and runtime"?

Not needed for this PR, but maybe for a new issue (and I can make it after this is merged.) It'd be nice to trigger `-DNDEBUG` in [runtime/make/Makefile.runtime.include](https://github.com/chapel-lang/chapel/blob/master/runtime/make/Makefile.runtime.include#L45) off the new `ASSERTS` flag instead of `OPTIMIZE`. I might also prefer to enable asserts if CHPL_DEVELOPER is set (or maybe if `OPTIMIZE=0`?) And lastly we should pipe this through for other 3rd party packages.

mppf:
@ronawho - I created issue #11286 for the follow-ons

PR #11290, opened 2018-10-03
Add chpl_task_getFixedNumThreads(), and one implementation thereof.

gbtitus:
Add a new tasking layer function that, if the implementation runs tasks
by hosting them on a fixed number of threads, returns the number of such
threads.  Otherwise it returns 0 (zero).  So far only tasks=qthreads
returns a non-zero value for this function.  tasks=massivethreads would
do so as well, I believe, but for now I've left that as future work.
For tasks=fifo this function returns 0.

This will be used in the comm=ofi implementation, and we have discussed
using it in the comm=ugni implementation.

### DIFF
--- a/runtime/src/tasks/qthreads/tasks-qthreads.c
+++ b/runtime/src/tasks/qthreads/tasks-qthreads.c
@@ -1029,9 +1029,8 @@ int32_t chpl_task_getNumBlockedTasks(void)
 // Threads
 
 uint32_t chpl_task_impl_getFixedNumThreads(void) {
-    if (chpl_qthread_done_initializing)
-        return (uint32_t)qthread_num_workers();
-    return 0;
+    assert(chpl_qthread_done_initializing);
+    return (uint32_t)qthread_num_workers();
 }
 
 uint32_t chpl_task_getNumThreads(void)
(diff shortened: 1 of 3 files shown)


runtime/src/tasks/qthreads/tasks-qthreads.c
```diff
@@ -1028,6 +1028,12 @@ int32_t chpl_task_getNumBlockedTasks(void)
 
 // Threads
 
+uint32_t chpl_task_impl_getFixedNumThreads(void) {
+    if (chpl_qthread_done_initializing)
```
ronawho: Is this backwards? Shouldn't we return 0 if qthreads is _not_ initialized?

Also I wonder if we want to support this call if we're not initialized. Should this be an error or a warning instead of a silent "wrong" answer?

gbtitus: Good grief, how embarrassing.  Yes, it's backwards and I'll fix it.

I don't think a warning makes sense.  I'm okay with an internal error if the tasking layer isn't initialized yet.  I actually started out with it defined that way and then realized that to make that work right I'd have to actually implement it in all tasking layers rather than relying on a default implementation for things like fifo, because there's not an implementation-independent way of determining if the tasking layer is initialized yet.  So, being lazy, I just made it also return 0 for "not initialized".

I suppose another solution would be to add a `chpl_task_inited()`, analogous to `chpl_mem_inited()`.  That's actually somewhat more attractive to me.

PR #11297, opened 2018-10-03
Get comm=ofi initialization working.

gbtitus:
There are a couple of lead-up commits here to increase code commonality
across the comm layers and restructure the ofi comm layer, but the
overall effect is to get comm=ofi to the point where it will initialize
for multi-node execution.

The basic setup is, per node, a single receive endpoint for inbound AM
requests and a single scalable transmit endpoint with some number of
transmit contexts.  For tasking layers with a fixed number of threads
(qthreads, e.g.) there are enough transmit contexts for each worker
thread to have a private one for sending AM requests and doing RDMA
transfers, plus a private for the AM handler thread to send responses
on.  The provider defaults to sockets for now, but one can force use of
a particular provider (assuming it can match the requested hints) by
setting `CHPL_RT_COMM_OFI_PROVIDER`.  We'll use either scalable or basic
memory registration, depending on the provider (basic for gni, scalable
for any other provider).

Most of the functional changes here involve a wholesale replacement of
the previously existing init code with stuff brought over more or less
directly from the ofi mockup.  There was some adjustment due to the
environment the mockup code finds itself now running in, for example the
provider name now comes from an environment variable and the number of
transmit contexts now comes (preferably) from the tasking layer thread
count, where both of those came from the command line in the mockup.
The dynamic memory allocation interface is different now, too: we use
the Chapel runtime memory management interface instead of the malloc()
interface.  The logic flow is largely unchanged, however.

I also converted the mockup's VERBOSE interface for expected output into
the comm layer's DBG interface for debug-only output.

There are a few TODOs in here for things I haven't decided yet but that
aren't pressing and/or don't seem to have functional implications.
There are also a number of places where variables and functions are
hidden using the preprocessor.  In most cases these are just to avoid
"defined but not used" compiler errors, for things we won't reference
until the AM and/or RDMA code arrives.  (For now, the AM and RDMA
functions just produce internal errors.)

On 2 nodes of a Cray XC system, `hello.chpl` built with this comm layer
gets through initialization with the sockets provider and then halts as
expected the first time it tries to communicate.  With the gni provider
it gets through initialization up to the memory registration step, where
it fails because there isn't anything to register yet, also as expected.
The shutdown code isn't tested, since the test program always takes the
error exit path and we only shut down the comm layer on the normal path.
I'm not concerned about that, though; shutdown is much simpler than init
and if there are problems in shutdown they should be easy to fix.

### DIFF
--- a/util/printchplenv
+++ b/util/printchplenv
@@ -231,7 +231,7 @@ def filter_overrides(chpl_env):
 def filter_tidy(chpl_env):
     comm = ENV_VALS['CHPL_COMM']
     if chpl_env.name == '  CHPL_COMM_SUBSTRATE':
-        return comm != 'none' and comm != 'ugni'
+        return comm != 'none' and comm != 'ofi' and comm != 'ugni'
     elif chpl_env.name == '  CHPL_GASNET_SEGMENT':
         return comm == 'gasnet'
     elif chpl_env.name == '  CHPL_NETWORK_ATOMICS':
--- a/runtime/include/chpl-tasks.h
+++ b/runtime/include/chpl-tasks.h
@@ -339,10 +339,11 @@ int32_t chpl_task_getNumBlockedTasks(void);
 // Threads
 
 //
-// If the tasking layer runs tasks on a fixed number of threads, and has
-// been initialized, this returns the number of such threads.  Otherwise
-// it returns 0 (zero).  As examples, for CHPL_TASKS=qthreads it returns
-// the number of worker threads, while for CHPL_TASKS=fifo it returns 0.
+// If the tasking layer runs tasks on a fixed number of threads, this
+// returns the number of such threads.  Otherwise it returns 0 (zero).
+// As examples, for CHPL_TASKS=qthreads it returns the number of worker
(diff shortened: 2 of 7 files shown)


gbtitus:
This will be easiest to review commit by commit.  Note that the `chpl_task_getFixedNumThreads()` commits are here also, so I could use that function in the last commit on this branch.

util/printchplenv
```diff
@@ -231,7 +231,7 @@ def filter_overrides(chpl_env):
 def filter_tidy(chpl_env):
     comm = ENV_VALS['CHPL_COMM']
     if chpl_env.name == '  CHPL_COMM_SUBSTRATE':
-        return comm != 'none' and comm != 'ugni'
+        return comm != 'none' and comm != 'ofi' and comm != 'ugni'
```
ronawho: I think this is a little cleaner as `return comm not in ['none', 'ofi', 'ugni']`. Or we could just switch to `comm == 'gasnet'`

gbtitus:
@ronawho I updated printchplenv as you suggested, and I've merged it as you probably saw.  But if you have time to take a closer look at this next week I'd definitely appreciate it.

Issue #11301, opened 2018-10-04
pragma "init copy fn" interferes with error handling


vasslitvinov:
In the following program:

```chpl
iter willthrow() throws {
  throw new Error();
  yield 777;
}

pragma "init copy fn"
proc driver(ARG) {
  writeln("driver start");
  for idx in willthrow() do writeln(idx);
  writeln("driver finish");
  return 555;
}

var BBB = driver(willthrow());
writeln("BBB = ", BBB);
```

The statement `writeln("driver finish")` gets executed, even though this happens after the iterator of the preceding loop throws.

When `pragma "init copy fn"` is removed, that statement is not executed, which is the correct behavior. The statement `writeln(BBB)` is not executed in either case, which is also correct.

bradcray:
Trying to get my head around this one:  We're trying to move away from initCopy() routines in general and, by extension I'd guess, this pragma.  So is the fix here really to rewrite the code using initCopy() to use copy initializers instead?  Tagging @benharsh for opinions.

benharsh:
I'd rather try to implement something using initializers than to try and 'fix' initCopy.

bradcray:
@vasslitvinov: I'm trying to determine what to do with this issue given @benharsh's comment and your support of it.  Should we close it outright?  Is there some internal use of the pragma which is causing problems due to this issue that we should focus on rewriting?

vasslitvinov:
This issue causes incorrect behavior in `proc chpl__initCopy(ir: _iteratorRecord)`.

See #11303, which works around this issue.

While we do not necessarily need to fix this #11301 per se, I am not sure that closing it at the moment is right. I suggest moving it to icebox.

vasslitvinov:
P.S. I think this is the issue we trip over in #11315.

Issue #11300, opened 2018-10-04
preserve the line number when re-throwing an exception


vasslitvinov:
A Chapel user may want to write code that reacts to an exception from a called function then re-throws it so that it is handled higher up in the call stack.

When this re-thrown exception is caught higher up in the call stack, we want it to report the file/line number for the location where it was thrown originally, not where it was re-thrown. Right now the latter happens.

For example:
```chpl
proc willthrow() throws {
  throw new Error();          // original throwing location
}

proc rethrow() throws {
  try {
    willthrow();
  } catch e {
    ... some actions upon an exception: examine it, do cleanup, etc. ...
    throw e;                 // re-throwing location
  }
}

proc main throws {
  rethrow();
}
```
Upon execution, we want the program to report the uncaught exception as originating in willthrow(), not in rethrow().

Test:
```
test/errhandling/errorMessages/preserveLineNoUponRethrow.chpl
```

vasslitvinov:
@mppf fyi

vasslitvinov:
@cassella mentions that other languages use `throw;` without the expression to indicate re-throwing of the error that triggered the current `catch` clause.

@cassella also reminded me that in Java, in my recollection, the file/line number of an error object is set when that object is created, regardless of where it is thrown. In order to record the location of the throw ing statement, a wrapper error would be used, something like:
```
throw new Error(anAlreadyExistingError);
```

dlongnecke-cray:
I have an implementation of this that works, however it does incur a runtime penalty because it uses a flag (that is a field of Error) to prevent file/line IDs from being set more than once.

I am trying to find a compile-time only solution, but it is proving to be a bit more difficult because calls to `chpl_fix_thrown_error` are inserted early on during normalization.

---

Question: Is it worth the extra time I'll need to implement a version with no runtime overhead? Or should I open a PR with the simple version now?

dlongnecke-cray:
> @cassella mentions that other languages use `throw;` without the expression to indicate re-throwing of the error that triggered the current `catch` clause.
> 
> @cassella also reminded me that in Java, in my recollection, the file/line number of an error object is set when that object is created, regardless of where it is thrown. In order to record the location of the throw ing statement, a wrapper error would be used, something like:
> 
> ```
> throw new Error(anAlreadyExistingError);
> ```

If we were to go with the approach used by Java, which ties the file/line of an error to the point of their initialization, then implementing the desired behavior with no runtime overhead would be trivial.

[...]

PR #11303, opened 2018-10-04
Cleanup upon an error in initCopy(_iteratorRecord)

vasslitvinov:
When creating an array using a loop expression and the computation
of a to-be array element throws an exception, do not create the array.
Instead, clean up the to-be array elements that have already been computed.

This came from observing the bug in #11301, where initCopy(_iteratorRecord)
is the function in which pragma "init copy fn" interferes.
I added the pragma "unchecked throws" so that client code does not have
to insert try or throws annotations.

I needed to adjust .good due to #11300. I added a .readme to explain
that the .good is not the ultimately-desired output.

This removes memory leaks in the tests:

    test/errhandling/ferguson/loopexprs-caught.chpl
    test/errhandling/ferguson/loopexprs-uncaught.chpl

While there, removed a stray comment, as suggested by DavidI.

Future work:
* Add similar cleanup to the shape-ful version of initCopy(_iteratorRecord).
* Split the tests loopexprs-caught, loopexprs-uncaught so that all three
  variants on creating an array using a loop expression get tested.
* Test the case where it is the iterator that throws, not the computation
  of the current element.

Testing:
- [ ] linux64 -futures
- [ ] valgrind

### DIFF
--- a/modules/internal/ChapelArray.chpl
+++ b/modules/internal/ChapelArray.chpl
@@ -4212,6 +4212,7 @@ module ChapelArray {
   }
 
   pragma "init copy fn"
+  pragma "unchecked throws"
   proc chpl__initCopy(ir: _iteratorRecord) {
 
     // We'd like to know the yielded type of the record, but we can't
@@ -4235,6 +4236,8 @@ module ChapelArray {
 
     var callAgain: bool;
     var subloc = c_sublocid_none;
+
+   try {
 
     for elt in ir {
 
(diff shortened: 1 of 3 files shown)


modules/internal/ChapelArray.chpl
```diff
@@ -4212,6 +4212,7 @@ module ChapelArray {
   }
 
   pragma "init copy fn"
+  pragma "unchecked throws"
```
mppf: I don't think this function should use unchecked throws.
It should be apparent to the user when they are evaluating an iterator that throws.
E.g.

``` chapel
var A = for i in 1..100 do myThrowingFunction(i);
```

should need to be in a try/catch or in a function that throws.

mppf:
@vasslitvinov - probably should close this PR if progress on it is not going to happen soon. (Or did it already get handled somehow else?)

vasslitvinov:
This change does not seem important enough to pursue at the moment.
Closing this PR for now.

PR #11309, opened 2018-10-05
Minor cleanup in dsiDestroyArr and autoDestroy

vasslitvinov:
* Remove bunch of pragmas and primitives in dsiDestroyArr and
  dsiDestroyDataHelper. These data back to 2016 or earlier.

* Switch to explicit 'ref' intent for the formal for chpl__autoDestroy
  on owned and shared. I need this so I can call chpl__autoDestroy()
  in some cases in my other work. This required an explicit dereference
  of the formal in the chpl__autoDestroy body - otherwise the compiler
  leaves the body empty. I did not investigate this, however I think
  that the compiler sees a pointer and concludes that nothing is needed
  to destroy it.

Tested: linux64, valgrind, memleaks.

Discussed with / suggested by @mppf.

### DIFF
--- a/modules/internal/DefaultRectangular.chpl
+++ b/modules/internal/DefaultRectangular.chpl
@@ -990,22 +990,16 @@ module DefaultRectangular {
     override proc dsiGetBaseDom() return dom;
 
     proc dsiDestroyDataHelper(dd, ddiNumIndices) {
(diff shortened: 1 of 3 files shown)


mppf:
Thanks!

Issue #8057, opened 2017-12-17
internal error: IPE0098


hariharanj4v4:
### Summary of Problem
The error occurs in 'chpl-ipe' when intended to use any module with the 'use' keyword.

### Steps to Reproduce

**Execution command:**
 chpl-ipe

### Configuration Information

- Output of `chpl --version`: 
chpl Version 1.16.0
- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
CHPL_TARGET_PLATFORM: linux64
CHPL_TARGET_COMPILER: gnu
CHPL_TARGET_ARCH: native
CHPL_LOCALE_MODEL: flat
CHPL_COMM: none
CHPL_TASKS: qthreads
CHPL_LAUNCHER: none
CHPL_TIMERS: generic
CHPL_UNWIND: none
CHPL_MEM: jemalloc
CHPL_MAKE: make
CHPL_ATOMICS: intrinsics
CHPL_GMP: gmp
CHPL_HWLOC: hwloc
CHPL_REGEXP: re2
CHPL_WIDE_POINTERS: struct
CHPL_AUX_FILESYS: none

- Back-end compiler and version, e.g. `gcc --version` or `clang --version`: 
gcc (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406

ben-albrecht:
Thanks for reporting this @hariharanj4v4. 

@bradcray's comment on #6760 is also related here:

>In the spirit of full disclosure, I should mention that the ipe has received very little attention over the past several releases, not due to lack of interest, but lack of resources and a need to more fully re-architect our compiler to better support an interpreted/interactive mode. We wish we could spend more time on it, but have been unable to do so recently.

bradcray:
The `use` of modules is outside of the set of features that are documented as being supported by `chpl-ipe` in its initial release:

https://chapel-lang.org/docs/1.16/technotes/chpl-ipe.html#the-ipe-subset-of-chapel

That doesn't mean that an internal error is reasonable, but I'm not surprised that the feature doesn't work...

Since I wrote the comment on issue #6760, interest in putting more effort into `chpl-ipe` has grown considerably, and I'm hoping 2018 will afford us more time work on it.

bradcray:
Support for chpl-ipe has been removed.  We hope to return to an interactive mode for Chapel in the coming years, but have decided to take a different approach than chpl-ipe.

Issue #6760, opened 2017-07-19
Chapel-IPE fails on writeln()


buddha314:
### Steps to Reproduce

Within the chpl-ipe command line, I think this is pretty self-explanatory.
**Source Code:**

```chapel
chpl-ipe



IPE 0.1

 1 > var x = 20;
 2 > writeln('X is');
   IpeProcedure::resolve(SymExpr*, actuals[])
   matchCount = 0


      #<SymExpr       1646 var:  #<VarSymbol     1553 name: NULL.writeln                qual: unknown          type: #<PrimitiveType     255 procedure> depth:  0 offset:  200>>
   0: #<SymExpr       1647 var:  #<VarSymbol     1636 imm:  "X is"                       type: #<AggregateType      18internal error: MIS0491 chpl Version 1.15.0

Internal errors indicate a bug in the Chapel compiler ("It's us, not you"),
and we're sorry for the hassle.  We would appreciate your reporting this bug -- 
please see http://chapel.cray.com/bugs.html for instructions.  In the meantime,
the filename + line number above may be useful in working around the issue.
```

**Compile command:**

**Execution command:**

**Associated Future Test(s):**

### Configuration Information

- Output of `chpl --version`:
```
chpl Version 1.15.0
Copyright (c) 2004-2017, Cray Inc.  (See LICENSE file for more details)
```

- Output of `$CHPL_HOME/util/printchplenv --anonymize`:
- Back-end compiler and version, e.g. `gcc --version` or `clang --version`:
```
gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```

- (For Cray systems only) Output of `module list`:

bradcray:
This is embarrassing.  It looks to me as though this behavior broke since the IPE's documentation was written.  Specifically, between then and now `'foo'` started being interpreted as a Chapel string literal whereas prior to that it had been interpreted as a C string literal.  Today, C string literals can be expressed using `c'foo'` and it appears to me that changing to that form makes this test work as expected.  Still, it'd be nice if this worked with "normal" Chapel strings, and at the very least we should update the documentation.  I've asked @noakesmichael if he could take a look at this issue.

[...]

PR #11310, opened 2018-10-05
Retire chpl-ipe

### LINKED ISSUE #11308: Retire support for chpl-ipe
Though we are still very interested in supporting an interactive mode of programming Chapel, the consensus seems to be that the current 'chpl-ipe' prototype is not the best way of getting there.  To that end, it probably makes sense to retire code, testing, and documentation of chpl-ipe to reduce the set of code that we need to carry forward in working condition.

bradcray:
Work on chpl-ipe has stalled out as we've become increasingly
confident that support for an interpreted environment is likely
to only be achieved via other approaches.  Meanwhile, chpl-ipe's
documentation has become increasingly stale and we're carrying
around code that requires effort to maintain.  For these reasons,
we're retiring support for chpl-ipe and this PR removes the code
that's most obviously related to its support.  Comments in the code
imply that there may be fields in some classes that are IPE-specific
(and could therefore now be removed), but I didn't try to tease
those apart in this effort.

Resolves #11308.


ben-albrecht:
@bradcray - should we retire open `chpl-ipe` issues now that it is retired?

- https://github.com/chapel-lang/chapel/issues/6760
- https://github.com/chapel-lang/chapel/issues/8057

bradcray:
That would be great, thanks!

Issue #11315, opened 2018-10-06
bug: upon a throw, error "non-equal lengths"


vasslitvinov:
The following code is an extract from the test `errhandling/ferguson/loopexprs-caught` where I replaced the range `0..3` with a domain. This replacement simulates what would happen in that test if we switched shape preservation for ranges from using the shape-less initCopy (#10942) to the shape-ful one (added for domains in #9994).

The error I am currently getting in that code is not the desired behavior:
```
error: zippered iterations have non-equal lengths
```

The test (forthcoming in the test system):
```chpl
  proc willthrow() throws {
    throw new Error();
    return 777;
  }

  var array1 = for i in {0..3} do willthrow();  // the error is reported here
```

Issue #11314, opened 2018-10-06
spike: visibility issue in RA/diten


vasslitvinov:
The code at the bottom is a simplification of the current code in `test/studies/hpcc/RA/diten/`, with a range shape 0..#numLocs replaced by a domain shape LocaleSpace. (Any domain can be used instead.) This replacement simulates what would happen in that RA test if we switched shape preservation for ranges from using the shape-less initCopy (#10942) to the shape-ful one (added for domains in #9994).

In that code I am currently getting:
```
In function 'chpl__staticFastFollowCheck':
error: unresolved call 'physicalMemory3([domain(1,int(64),false)] locale)'
```

This appears to be due to incorrect visibility of a compiler-generate function upon a loop expression.

Test :

```chpl
// arrays/shapes/visibility-error.chpl

module m1 {
  use m2;

  proc main() {
    const totalMem = + reduce physicalMemory3(Locales);    // the error is reported here
    var myBuckets = new unmanaged Buckets();
  }

  proc physicalMemory3(arg:locale) return 0;
}

module m2 {
  class Buckets {
    var BucketArray = [idx in LocaleSpace] 777;
  }
}
```

benharsh:
Isn't this about the initCopy for _iteratorRecord? If so, I think it falls into the same bucket as the sync/single/array-slice unwrapping problem.

vasslitvinov:
I have not looked into that question. BenH/Lydia probably have a better understanding.

PR #11319, opened 2018-10-08
Enable parallelism for shape-ful expressions over ranges

vasslitvinov:
This PR moves shape preservation with forall- and promoted expressions
over ranges from the old, serial implementation to the new, parallel
implementation. This new implementation has been in use over arrays and
domains since #9994 and #10758.

The user-facing effects are:

* A compile-time error when building a skyline-ish array, i.e.
an array whose elements are arrays, using a forall- or promoted expression.
This replaces the former incorrect run-time behavior or run-time error
"zippered iterations have non-equal lengths" in these cases.

* A skyline-ish array can still be created using a for-expression.
Which falls back to the old implementation.
Such an array can have elements that are arrays of **different** sizes, ex:
```
    test/arrays/ferguson/from-iterator/skyline.chpl
```
* Expose -- without fixing -- two pre-existing issues:

  - #11314: a strange visibility issue
   Use a workaround - see RA below.

  - #11315: upon a throw, get "error: zippered iterations have non-equal lengths"
   Remove testing that hits this issue - see loopexprs-caught below.

#### TEST CHANGES

* Apply a workaround for #11314 in an RA studies version:
```
    test/studies/hpcc/RA/diten/buckets.chpl
```
* Remove the execopts that result in hitting #11315.
```
    test/errhandling/ferguson/loopexprs-caught.chpl
```
* Update the source to keep building arrays of arrays:
```
    test/arrays/deitz/skyline.chpl
    test/arrays/shapes/skyline.when-empty.chpl
    test/expressions/loop-expr/scoping-1.chpl
    test/trivial/deitz/statements/nested_fors.chpl
```
* Update .bad to reflect the newly-added compile-time error:
```
    test/arrays/shapes/skyline.chpl
    test/arrays/vass/literal-indexing.chpl
    test/expressions/loop-expr/scoping-2.chpl
```
* Add a test for building arrays of domains:
```
    test/domains/vass/arrays-of-domains.chpl
```
* Update .bad because the current/undesirable output that's there
was 2 copies of the same warning, now it is 3 copies:
```
    test/classes/delete-free/lifetimes/forall-of-new-borrowed-escapes.chpl
```
* While there, update .future for:
```
    test/domains/vass/no-dom-field-error.chpl
```
#### TESTING
- [x] linux64 -futures
- [x] gasnet -futures (*)
- [x] valgrind

(*) TODO
- [ ] remove the execopt I unwittingly added to `loopexprs-caught` because under gasnet it works neither with this PR nor on master


vasslitvinov:
Instead see #11501 -- a rebase of this PR.

PR #11320, opened 2018-10-08
Add a common function to snprintf() a size_t or double K/M/G value.

gbtitus:
comm=ugni had a couple of internal functions for `snprintf()`ing a value
of type `size_t` or `double`, in the form `nG`, `nM`, or `nK` based on
whether they're were >= 2\*\*30, >= 2\*\*20, or otherwise.  Here, publicize these
for common use in chpl-utils.h, and use them in the ugni comm layer.  (We
will use them in comm=ofi soon also.)  As part of this, adapt them to do the
right thing for values < 2\*\*10, which was a situation that could not occur in
their original environment.

While here, rename some files:
- include/chpl-align.h -> include/chpl-util.h
- include/chpl_rt_utils_static.h -> include/chpl-tasktable.h
- include/chplexit.h -> include/chpl-exit.h
- src/chplexit.c -> src/chpl-exit.c

`chpl-align.h` is renamed because the new printing functions needed to go
into a source file for things which might be useful in any runtime layer
and surprisingly, we didn't really have a vanilla "utility" module in
the runtime.  So, here we co-opt `chpl-align.h` to form `chpl-util.h`.  The
`chpl_rt_utils_static.h` renaming is because its name seems to imply that it
is a general utility module but actually it just contains the extern
definitions for some task table functions which are currently in the
module code.  `chplexit.c` is renamed because it needed to change anyway
since it included `chpl_rt_utils_static.h` when it didn't need to and
because I've always been bothered that `chpl-init` had a dash and `chplexit`
did not.  And finally,` chplexit.h` is renamed just so that it matches
`chpl-exit.c`.

The renamings caused a number of changes in include statements and the
like, in a ripple effect.


gbtitus:
This will definitely be easiest to review commit-by-commit.

runtime/include/chpl-util.h
```diff
@@ -53,4 +60,44 @@ unsigned char* round_up_to_mask_ptr(unsigned char* p, uintptr_t mask)
 }
 
 
-#endif // _chpl_align_h_
+////////////////////////////////////////
+//
+// snprintf() a K/M/G size
+//
+
+static inline
```
ronawho: Any reason to make this inline?

runtime/include/chpl-util.h
```diff
@@ -53,4 +60,44 @@ unsigned char* round_up_to_mask_ptr(unsigned char* p, uintptr_t mask)
...
+  return snprintf(buf, bufSize, "%zd", val);
+}
+
+static inline
+int chpl_snprintf_f_KMG(char* buf, int bufSize, double val) {
+  const double GiB = (double) (1UL << 30);
+  if (val >= GiB)
+    return snprintf(buf, bufSize, "%.1fG", val / GiB);
+
+  const double MiB = (double) (1UL << 20);
```
ronawho: No longer c90 style decls -- do we still care about avoiding mixed decls and code?

ronawho:
I'm actually not a huge fan of a general purpose "utils" file, because I think it'll end up being a grab bag of unrelated functionality. What would you think of keeping chpl-align and adding a chpl-format or something?

runtime/include/chpl-util.h
ronawho: I agree, but I think mixed decls would make a lot of runtime code easier to read, but we've historically stayed away from it. I just can't remember if we stopped caring about C90.

ronawho: Ah, it looks like we officially stopped caring in https://github.com/chapel-lang/chapel/pull/6224, so +1 for mixed-decls

gbtitus:
After further review I'm closing this PR in favor of a specific one just to add the new functions.

PR #11321, opened 2018-10-08
Resolve some leaks in user code

benharsh:
This PR updates some tests such that they no longer leak (or leak less). The highlights include:
- not break-ing from a loop (leaking iter classes)
- adding dsiDestroy* methods to a non-standard domain map
- more use of ``borrowed`` over ``unmanaged``

Testing:
- [x] local + futures
- [x] gasnet + futures
- [x] memleaks (down to ~25K leaked)

### DIFF
--- a/test/distributions/bradc/assoc/UserMapAssoc.chpl
+++ b/test/distributions/bradc/assoc/UserMapAssoc.chpl
@@ -487,7 +487,7 @@ class UserMapAssocDom: BaseAssociativeDom {
   }
 
   proc setup() {
-    for localeIdx in dist.targetLocDom do
+    coforall localeIdx in dist.targetLocDom do
       on dist.targetLocales(localeIdx) do
         if (locDoms(localeIdx) == nil) then
           locDoms(localeIdx) = new unmanaged LocUserMapAssocDom(idxType=idxType,
@@ -510,7 +510,7 @@ class UserMapAssocDom: BaseAssociativeDom {
   }
 
   proc dsiDestroyDom() {
-    for localeIdx in dist.targetLocDom do
+    coforall localeIdx in dist.targetLocDom do
       on dist.targetLocales(localeIdx) do
         delete locDoms(localeIdx);
   }
--- a/test/distributions/bradc/assoc/UserMapAssoc.chpl
+++ b/test/distributions/bradc/assoc/UserMapAssoc.chpl
@@ -508,6 +508,12 @@ class UserMapAssocDom: BaseAssociativeDom {
   proc dsiReprivatize(other) {
     locDoms = other.locDoms;
   }
+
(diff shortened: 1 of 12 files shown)


test/associative/bharshbarg/arrays/diffDomZip.chpl
```diff
@@ -29,7 +29,6 @@ while !different {
   for (aa, bb) in zip(A,B) do
     if aa != bb {
       different = true;
-      break;
```
cassella: Is removing the break just a workaround, or is there something wrong with breaking from a loop like this?

benharsh: Yeah, it's just a workaround.

benharsh:
@mppf (if you're the right person) : would you mind reviewing the UserMapAssoc changes?
@daviditen : would you mind reviewing the changes under ``test/studies/hpcc/RA/diten/`` ?

test/distributions/bradc/assoc/UserMapAssoc.chpl
```diff
@@ -508,6 +508,12 @@ class UserMapAssocDom: BaseAssociativeDom {
   proc dsiReprivatize(other) {
     locDoms = other.locDoms;
   }
+
+  proc dsiDestroyDom() {
+    for localeIdx in dist.targetLocDom do
```
mppf: Why is this a 'for' while destroyArr uses a coforall?

mppf:
I had a minor question about UserMapAssoc changes but the rest look good to me.

daviditen:
The RA changes look fine to me.

I don't think you should remove the `break` in `diffDomZip.chpl`.  We should fix the underlying problem instead of changing the behavior or tests to work around it.

test/distributions/bradc/assoc/UserMapAssoc.chpl
benharsh: Good point, not sure why I did that...

mppf: The pull of the leak-free side is strong and hard to resist.

mppf: Is it that way in something you based it off of, Block etc?

benharsh: Ah, I think I based it on the loop in UserMapAssocDom.setup. Anyways, I'll update it to use a coforall and re-run tests on that directory.

mppf: I asked because I wondered if whatever you based it on should use coforall.

mppf: Maybe worth a comment if it's not obvious.

Issue #11154, opened 2018-09-17
array aliasing and LLVM performance for elegant tuplesVSarrays


mppf:
See https://chapel-lang.org/perf/chapcs/llvm/?startdate=2018/07/12&graphs=elegancetuplesvsarrays

Elegant tuplesVSarrays is not performing as well as it could with LLVM. This has to do with detecting array aliasing. Here are some related comments:
 * https://github.com/chapel-lang/chapel/issues/5043#issuecomment-322617090
 * #7762

Issue #7762 notes:

> Array wrapper records do not share the underlying arrays unless they are array views. This could be used to detect easy cases.

So, an initial attempt at solving this problem would be:
 * Chapel compiler can detect when arrays may not alias (e.g. two different array variables that are not array views and neither is a reference)
 * in that event, use noalias and alias.scope metadata to mark arrays that cannot possibly alias

bradcray:
It wasn't clear to me from the notes above whether you were keeping in mind that arrays can also alias via scoping.  For example, in the following A and X alias though neither is an array view:

```chapel
var A: [1..10] real;

proc foo(X: []) {
  A = X;  // A and X are the same thing but neither is an array view
}
```

mppf:
@bradcray - yes that's right, but in that event, they compiler's view is that `X` argument is actually a reference to an array wrapper record. Anyway I'll note it above.

mppf:
First, there is a difference in the "performant" version between --llvm and GCC.
This can be resolved by passing `--no-ieee-float` in which case the LLVM version outperforms the GCC version. This is because the --llvm code generator isn't peephole optimizing vaddpd + vmulpd while the GCC version uses vfmadd132pd (i.e. the GCC version is using fused multiply-adds).

mppf:
Indeed, in the "elegant" version, even with --no-ieee-float, the vfmadd132pd instructions have vmovsd instructions doing loads interspersed.

mppf:
Additional evidence that it's an alias analysis problem is that changing the type of subArray from `real` to `real(32)` resolves the performance problem.

mppf:
PRIM_ARRAY_GET returns a pointer to an array element and might be useful here.
The compiler will need to, after inlining (ideally), decide for each PRIM_ARRAY_GET whether it could possibly alias other PRIM_ARRAY_GETs in the same scope.

mppf:
[...]

PR #11322, opened 2018-10-09
Address problems with ffloatOpt in --llvm

mppf:
ffloatOpt can have 3 values but one spot only handled 0,!=0
additionally other compilers fuse multiply-add by default,
so we will do so as well.

I noticed these issues while investigating issue #11154.

Reviewed by @dmk42 - thanks!


mppf:
I'm seeing 1 failure - [Error matching program output for studies/parboil/stencil/stencilSerial]

mppf:
This seemed to have some performance impact:
 * tuplesVSarrays is faster
 * elegance arrayAdd is slower
 *  N-body variations  is slower

PR #11323, opened 2018-10-09
Add common functions to snprintf() nnn[KMG] values.

gbtitus:
comm=ugni had a couple of internal functions for `snprintf()`ing a value
of type `size_t` or `double`, in the form 'nnnG', 'nnnM', or 'nnnK' based on
whether they were >= `2**30`, >= `2**20`, or otherwise.  Here, publicize
these for common use in chpl-format, adjust them to properly handle
values less than `2**10`, and replace use of the originals in the ugni
comm layer.

We will use these in comm=ofi soon also.

### DIFF
--- a/runtime/include/chpl-format.h
+++ b/runtime/include/chpl-format.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2004-2018 Cray Inc.
+ * Other additional copyright holders may be indicated within.
+ * 
+ * The entirety of this work is licensed under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * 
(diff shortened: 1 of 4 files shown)


gbtitus:
The new functions could not be merely `static` directly in the .h file, because when I built with `WARNINGS=1` that led to gcc "defined but not used" warnings which were translated into errors for .c files which included the .h but didn't call at least one of the functions.  Declaring them `static inline` in the .h did resolve that problem, but they won't be called in any performance-sensitive situation where inlining would make a difference and I'm not comfortable using `inline` directives only for the side effect of getting rid of "def'd but not used" errors.  So I've just made these functions traditional non-inlined ones.

Issue #11317, opened 2018-10-08
Combine alias analysis passes


mppf:
The Chapel compiler will soon include 3 alias analysis passes:
 * may-alias analysis in LICM
 * must-alias analysis for nil-checking (PR #11238)
 * interprocedural array alias analysis to hint LLVM (PR todo)

This issue is proposing an effort to combine these 3 alias analysis implementations. The expectation would be that nil-checking (which always runs) would continue to run a faster must-alias analysis, but that LICM and the LLVM hinting could benefit from each other and the must-alias analysis.

The result would be:
 * fewer bugs because we can carefully verify the combined alias analysis results with less testing
 * improved optimization (LICM can make use of the interprocedural analysis)
 * cleaner compiler architecture

mppf:
@ronawho fyi

mppf:
Also I'm pretty sure that LICM alias analysis is not detecting const-ref arguments the way it should be.

PR #11324, opened 2018-10-09
No-alias hints for LLVM

mppf:
This PR adds, just at the beginning of loopInvariantCodeMotion, a new analysis
that determines:

 * case where reference arguments cannot alias each other
 * cases where array header fields or element pointers may not alias each other

It additionally adds support in code generation to generate LLVM alias.scope
and noalias metadata.

Lastly, it adds tests to verify that the compiler is correct in its alias
analysis.
 
The reference argument analysis is an interprocedural analysis following Allen
& Kennedy "Optimizing Compilers for Modern Architectures". It determines cases
where a reference argument may not refer to the same value as another reference
argument by analyzing calls. It considers global variables and when the same
value is passed for multiple arguments.

The array analysis portion starts from the reference argument analysis and
additionally infers that separate array variables, unless they are of array
view type (e.g. slice, reindex, or rank change) cannot alias other separate
array variables. It applies this same reasoning to array references that may
alias. For arrays that cannot alias, it infers that the array header
information and the element pointers cannot alias either.

It uses 2 primitives to create no-alias hints because the compiler associates
no-alias information with particular Symbols. I also considered adding no-alias
information to calls that can create loads or stores. The trouble is, there are
many such calls in the AST.

Right now, the code generation portion of this uses global maps of no-alias
information. These could be updated to be per-function.

This resolves issue #11154.

Future work:
 * combine alias analysis passes (#11317)
 * decide if we want to reverse the pragma "aliasing array" (#11363)

Reviewed by @benharsh and @ronawho - thanks!

- [x] full local testing
- [x] full local --llvm --fast testing


modules/internal/ArrayViewRankChange.chpl
```diff
@@ -436,6 +436,7 @@ module ArrayViewRankChange {
   // other array class implementations, it supports the standard dsi
   // interface.
   //
+  pragma "aliasing array"
```
ronawho: The need for this pragma makes me nervous, though I'm not sure why. @benharsh any objections to needing to mark that array views can create aliases?

modules/internal/ChapelArray.chpl
```diff
@@ -339,13 +339,16 @@ module ChapelArray {
       return new _array(nullPid, value);
   }
 
+  /* This method is commented out because using it would have
+     implications for alias analysis of arrays.
```
ronawho: I don't understand this comment or the implications for commenting this code out. Does this an issue, a todo, something?

modules/internal/ChapelArray.chpl
```diff
@@ -2142,13 +2145,15 @@ module ChapelArray {
...
     pragma "owned"
+    pragma "alias scope from this"
     var _instance; // generic, but an instance of a subclass of BaseArr
     var _unowned:bool;
 
     proc chpl__promotionType() type {
       return _value.eltType;
     }
 
+    pragma "alias scope from this"
```
ronawho: pragmas like this also make me nervous because they are bound to be forgotten about when new methods are added

modules/internal/ArrayViewRankChange.chpl
benharsh: I think this would be OK for now, but I wonder if it would be better to assume that an array *does* alias by default and require that users opt-in for non-aliasing. Technically there's nothing stopping today's users from creating aliasing array distributions, as far as I know.

test/analysis/alias/array-arguments-const-ref.chpl
```diff
@@ -0,0 +1,17 @@
...
+
+proc main() {
+  var A: [1..10] int = 1;
+  var B: [1..10] int = 2;
+
+  var sum = test(A, B);
+  writeln(sum);
+}
+
+proc test(const ref a, const ref b) {
```
ronawho: If there isn't a test for this already, could you add one that has multiple levels of function calls where the outer-most call creates an alias? Something like:

```chpl
test2(A, B); // and a test with test2(A, A)

proc test2(ref a, ref b) {
  test1(a, b);
}

proc test1(ref a, ref b) {
  test(a, b);
}
```

compiler/optimizations/loopInvariantCodeMotion.cpp
```diff
@@ -1210,7 +1258,10 @@ static long licmFn(FnSymbol* fn) {
 
 void loopInvariantCodeMotion(void) {
 
-  if(fNoloopInvariantCodeMotion) {
+  // compute array element alias sets
+  addNoAliasSets();
```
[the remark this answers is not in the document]
mppf: I stopped adding compiler passes because it requires you to do something to the test system each time ;P

But besides that each pass runs a global garbage collection in-between and so the compiler is faster if we have fewer passes.

Happy to change the name.

compiler/optimizations/noAliasSets.cpp
```diff
@@ -0,0 +1,990 @@
+/*
+ * Copyright 2017 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
mppf: Whoops...

modules/internal/ChapelArray.chpl
mppf: This function isn't used right now. If it were used, we'd have an `_array` record for a DefaultRectangular array (say) that referred to the same instance as as another `_array` record variable. The core of the optimization concept here is that this case can't occur.

modules/internal/ChapelArray.chpl
mppf: This pragma isn't strictly necessary but I wanted the no-alias analysis to be able to determine when the result of a dsiAccess call has the same "scope" as the array it comes from.

In the DefaultRectangular example I started with (tuplesVSarrays), this pragma isn't necessary because dsiAccess is inlined.

This pragma is a performance hint.

modules/internal/ArrayViewRankChange.chpl
mppf: It'd be fine with me to reverse the pragma, but the core concept of this change relies on the fact that different _array records don't have instances or elements that alias each other unless they are array views. (Which wasn't always true, but I've been quietly building up to it for a couple years now :)

modules/internal/ChapelArray.chpl
ronawho: Can we just remove the function then?

mppf: Sure

test/analysis/alias/array-arguments-const-ref.chpl
mppf: I added such tests (array-arguments-ref-nested-*.chpl).

mppf:
I'm seeing performance improvements in LLVM perf testing from this PR on these tests:
 * Elegance tuplesVSarrays
 * Serial Array Access vs. Reference Access
 * Array Access vs. Reference Access in forall loop
 * Serial Array Access vs. Reference Access (Multidimensional)

mppf:
This LCALS muladdsub reproducer shows that this PR allows a variant of the LCALS kernel to be much faster:
``` chapel
use Time;

// num_samples: 3000 for "long", 300000 for "medium", 50000000 for "short"
config const num_samples = 50000000; // "short"
// array length: 44217 for "long", 5001 for "medium", 171 for "short"
config const len = 171; // "short"

proc main() {
  var ltimer:Timer;
  var RealArray_1D0: [0..#len] real;
  var RealArray_1D1: [0..#len] real;
  var RealArray_1D2: [0..#len] real;
  var RealArray_1D3: [0..#len] real;
  var RealArray_1D4: [0..#len] real;
  // The stuff from LCALSLoops.chpl loopInit
  initData(RealArray_1D0, 1);
  initData(RealArray_1D1, 2);
  initData(RealArray_1D2, 3);
  initData(RealArray_1D3, 4);
  initData(RealArray_1D4, 5);

  // This portion from RunBRawLoops.chpl runBRawLoops e.g.
  ltimer.start();
  for 0..#num_samples {
    //for i in vectorizeOnly(0..#len) {
    for i in 0..#len {
      RealArray_1D0[i] = RealArray_1D3[i] * RealArray_1D4[i];
      RealArray_1D1[i] = RealArray_1D3[i] + RealArray_1D4[i];
      RealArray_1D2[i] = RealArray_1D3[i] - RealArray_1D4[i];
    }
  }
  ltimer.stop();

  var chksum: real = 0.0;
  // Get this portion from LCALSLoops.chpl loopFinalize 
  updateChksum(chksum, len, RealArray_1D0);
  updateChksum(chksum, len, RealArray_1D1);
  updateChksum(chksum, len, RealArray_1D2);

  // Get the Checksum from LCALSChecksums.chpl
  var expect = 391727437.25186200960888527333736;
  writeln("Got checksum ", chksum, " expected ", expect);
  writeln("Kernel ran in ", ltimer.elapsed(), " seconds");
}

proc initData(ra: [] real, id: int) {
  const factor: real = if id % 2 != 0 then 0.1 else 0.2;
  for (r,j) in zip(ra, 0..) {
    r = factor*(j + 1.1)/(j + 1.12345);
  }
}

proc updateChksum(ref chksum:real, length: int, 
                  ra: [] real, scale_factor: real = 1.0) {
  use LongDouble;

  ref data = ra;
  var len = ra.numElements;

  var tchk: longdouble = chksum;
  for (j, dat) in zip(0..#len, data) {
    tchk += (j+1)*dat*scale_factor;
  }
  chksum = tchk:real;
}
```



|                      | GCC 8.2       | clang 6.0   |  llvm    | llvm + no-alias hints  |
| ------------ | ------------ | ----------- | ------  | --------------------- |
| time              |  3.9 s           |    12 s          | 11.5 s   | 2.1 s                           |

PR #11325, opened 2018-10-09
Add fixed heap support.

gbtitus:
This adds comm=ofi versions of `chpl_comm_impl_regMemHeapInfo()` and
`chpl_comm_impl_regMemHeapPageSize()`, to support a fixed heap when
using the gni provider.  Other providers typically support scalable
memory registration, so a fixed heap isn't needed.  But the limited
support for scalable memory registration in the gni provider makes it
much less work for us to simply require a fixed heap in that case.

The new support includes both allocation and registration.  The fixed
heap is on hugepages if a hugepage module is loaded, and on regular
pages otherwise.  For now this only works with the Aries NIC.  At some
point it will work with the Gemini NIC as well, although for Gemini we
will likely need to make hugepages mandatory instead of optional, due to
limitations in that NIC's on-board TLB.

While here, get rid of the support for multiple memory region sets,
which we inherited from the ofi mockup and were an experimental feature
even there.  Also replace the statically-defined memory regions we
inherited from the mockup with a small (10-entry) table of dynamically
added entries.  This small table should be sufficient since we won't be
supporting full dynamic registration.  In other words, we'll be able to
dynamically add regions to be registered, but only a few of them and
only during initialization, not generally during execution.

As a necessary side effect, this also introduces a tiny interface to
`libhugetlbfs`, so that we can link user programs with that when we have
it and avoid undefined symbols when we don't.

For the most part confirmation that this code is working was done by
observing debug output.  However, `hello.chpl` now runs on 2 Cray XC nodes
with either the sockets or gni providers and reaches the point where it
gets an internal error when trying to do its first communication.  In
other words, we've gotten through registration successfully.  Also, on 2
XC nodes with the gni provider we produce the correct (Aries) warning
for the `runtime/configMatters/comm/ugni/overflow-nic-tlb `test, before
halting due to the lack of actual communication support.

### DIFF
--- a/runtime/src/comm/ofi/comm-ofi-internal.h
+++ b/runtime/src/comm/ofi/comm-ofi-internal.h
@@ -57,6 +57,7 @@ uint64_t chpl_comm_ofi_dbg_level;
 #define DBG_ACK              0x100000UL
 #define DBG_COMMPROGRESS     0x200000UL
 #define DBG_MR              0x1000000UL
+#define DBG_HUGEPAGES       0x8000000UL
 #define DBG_FAB            0x10000000UL
 #define DBG_FABSALL        0x20000000UL
 #define DBG_FABFAIL        0x40000000UL
@@ -149,6 +150,14 @@ void chpl_comm_ofi_oob_init(void);
 void chpl_comm_ofi_oob_fini(void);
 void chpl_comm_ofi_oob_barrier(void);
 void chpl_comm_ofi_oob_allgather(void*, void*, int);
+
+
+//
+// Hugepage interface
+//
+
+void* chpl_comm_ofi_hp_get_huge_pages(size_t);
+size_t chpl_comm_ofi_hp_gethugepagesize(void);
 
 
 //
--- a/runtime/src/comm/ofi/comm-ofi-internal.h
+++ b/runtime/src/comm/ofi/comm-ofi-internal.h
@@ -127,6 +127,15 @@ char* chpl_comm_ofi_dbg_prefix(void);
 
 #define CHK_SYS_CALLOC(p, n) CHK_SYS_CALLOC_SZ(p, n, sizeof(*(p)))
 
+#define CHK_SYS_POSIX_MEMALIGN(p, a, s)                                 \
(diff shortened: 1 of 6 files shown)


runtime/src/comm/ofi/comm-ofi-internal.h
```diff
@@ -126,6 +127,15 @@ char* chpl_comm_ofi_dbg_prefix(void);
 
 #define CHK_SYS_CALLOC(p, n) CHK_SYS_CALLOC_SZ(p, n, sizeof(*(p)))
 
+#define CHK_SYS_POSIX_MEMALIGN(p, a, s)                                 \
+    do {                                                                \
+        if ((posix_memalign(&(p), (a), (s))) != 0) {                    \
```
ronawho: Use `sys_memalign` instead and rename macro?

gbtitus: Will do.  I'm not even sure why I didn't do that in the first place!

gbtitus: It'll fix the CI failure, too.

runtime/src/comm/ofi/comm-ofi.c
```diff
@@ -1073,8 +1288,7 @@ void chpl_comm_wait_nb_some(chpl_comm_nb_handle_t* h, size_t nhandles)
 }
 
 
-int chpl_comm_try_nb_some(chpl_comm_nb_handle_t* h, size_t nhandles)
-{
+int chpl_comm_try_nb_some(chpl_comm_nb_handle_t* h, size_t nhandles) {
```
ronawho: Not important, but I keep wanting to propose that me and you come up with a style guide we like for the runtime and start running clang-format to enforce it.

gbtitus: That would be great!  That particular change is because I lifted this code from comm=ugni where I used an idiosyncratic style for outermost curly braces that I have rued every since.  Here I'm undoing it for comm=ofi so it matches most other runtime code.

